PR #1577, opened 2020-07-29
Js console fixes

MStoykov:


MStoykov:
I think we agreed that we want https://github.com/dop251/goja/issues/162 to be fixed before we do that ?

na--:
Yeah, sorry, I forgot... :sweat_smile:

PR #1583, opened 2020-08-10
Override with env variables the cloud config

MStoykov:
This is how it should've been but ... because of the way this is
configurable ... it is somewhat complicated ...

The alternative to this (at this point) is to have the output know how
to merge the env variables back ... which seems even worse

### DIFF
--- a/cmd/cloud.go
+++ b/cmd/cloud.go
@@ -128,71 +128,32 @@ This will execute the test on the k6 cloud service. Use "k6 login cloud" to auth
 			return err
 		}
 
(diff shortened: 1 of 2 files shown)


imiric:
@MStoykov Could you rebase this on `master`?

MStoykov:
This will have to wait for #1599 as it likely will have conflicts either way

MStoykov:
Given how fragile this code has been in the past and that with https://github.com/loadimpact/k6/issues/1155 we will write completely new version of this handling I am closing this is issue in favor it just getting fixed then.

This basically fixed if you have configured a cloud option through the script, to be able to overwrite it from outside the script. Which is likely is a very rare use case

PR #1584, opened 2020-08-11
Replace dep with Go modules

### LINKED ISSUE #1070: Adopt Go modules
After Go 1.13 comes out, likely sometime next month, we should probably replace [dep](https://github.com/golang/dep/) with the official Go modules:
- https://github.com/golang/go/wiki/Modules
- https://blog.golang.org/modules2019
- https://blog.golang.org/using-go-modules
- https://blog.golang.org/migrating-to-go-modules

imiric:
Closes #1070 

This was prioritized because `dep` has [an issue](https://github.com/golang/dep/issues/1962) importing https://github.com/semihalev/sdns, which we're evaluating for the DNS fixes.

There are some minor differences between `go.mod` and the previous `Gopkg.lock`. The missing packages are indirect and seem fine, and the version differences shouldn't cause problems, but let me know if they should be explicitly overriden.

- In `go.mod` but not in `Gopkg.lock`:
```
github.com/Shopify/toxiproxy v2.1.4+incompatible // indirect  (from github.com/Shopify/sarama)
github.com/google/go-cmp v0.5.1 // indirect                   (from github.com/andybalholm/brotli)
github.com/klauspost/cpuid v1.3.1 // indirect                 (from github.com/klauspost/compress/zstd)
github.com/onsi/ginkgo v1.14.0 // indirect                    (from github.com/manyminds/api2go)
```

- Different versions in go.mod compared to (=>) Gopkg.lock:
```
github.com/stretchr/testify v1.2.2                      => v1.2.1
golang.org/x/crypto v0.0.0-20190308221718-c2843e01d9a2  => c7dcf104e3a7a1417abc0230cb0d5240d764159d
golang.org/x/net v0.0.0-20200520004742-59133d7f0dd7     => 351d144fa1fc0bd934e2408202be0c29f25e35a0
golang.org/x/text v0.3.2                                => v0.3.0
gopkg.in/yaml.v2 v2.3.0                                 => v2.1.1
```


na--:
Something fishy is going on :confused: `+180,579 −110,284 Files changed 818` Are you sure you haven't updated versions or something like that, there shouldn't be so many changes when you're just switching dependency managers and not upgrading the actual dependencies :confused:

.circleci/config.yml
```diff
@@ -12,14 +12,7 @@ jobs:
...
-            curl --fail https://raw.githubusercontent.com/golang/dep/master/install.sh | sh
-            dep version
-            dep status
-            case $(dep ensure 2>&1 | head -n 1) in Warning*) false ;; esac
-            dep check
-
+            go mod verify
 
   lint:
     docker:
```
imiric, line 18: We can remove most of the `GOPATH`/`GO111MODULE` shenanigans now, right?

imiric:
@na-- These were produced by `go mod vendor`/`go mod tidy`. I was also surprised by the amount, but it seems `go mod` is pulling in lots more files than `dep ensure` did.

na--:
I'd be fine if it was just pulling more files, maybe because it wasn't [pruning](https://github.com/loadimpact/k6/blob/0e94653c9954c17b2b1a480b456e23ec120feff2/Gopkg.toml#L84-L87) or tree-shaking as much as dep, but look at the diff... There are differences in old files, which suggests that different versions were pulled. 

I want to update our dependencies soon, but that should be in a separate PR - this one should just encapsulate the switch of dependency manager. Please investigate what's happening, something is fishy.

MStoykov:
From my memories of when I tried to this a year ago ... one of the reason for changing versions is that now `go mod` actually looks at go.mod which tells it that we need version X of given dependancy ... which is not how `dep` works ;)

imiric:
So it's a combination of several things:

- `go mod vendor` pulls in more (unrelated) files than `dep ensure`. See the goja changes even though the version remained the same:
```
> git show b72ffed1 --stat -- vendor/github.com/dop251/goja | grep '^ vendor'
 vendor/github.com/dop251/goja/.gitignore           |    3 +
 vendor/github.com/dop251/goja/.travis.yml          |   17 +
 vendor/github.com/dop251/goja/README.md            |  277 +++++
 vendor/github.com/dop251/goja/ast/README.markdown  | 1068 ++++++++++++++++++++
 vendor/github.com/dop251/goja/file/README.markdown |  110 ++
 vendor/github.com/dop251/goja/token/Makefile       |    2 +
 vendor/github.com/dop251/goja/token/tokenfmt       |  222 ++++
```

- Half of the changes are because many of the `golang.org/x/*` packages were updated.
```
> git show b72ffed1 --stat -- vendor/golang.org/x | grep 'files changed'
 420 files changed, 79116 insertions(+), 105655 deletions(-)
```

For example, `golang.org/x/crypto` was updated from https://github.com/golang/crypto/commit/c7dcf104e3a7a1417abc0230cb0d5240d764159d to https://github.com/golang/crypto/commit/c2843e01d9a2 .

I'm not sure why these were updated, but since we were using very old versions, and these are semi-stdlib packages, we should consider updating them anyway, though we could pin them for now and do that later. Let me know.

- `github.com/golang/protobuf`/`google.golang.org/protobuf` are transitive dependencies that also got updated, and ended up pulling in a bunch of files:
```
> git show b72ffed1 --stat | grep protobuf | wc -l
136
```

na--:
As I said, I'm fine with the READMEs and new files that were previously pruned. I'd strongly prefer if we can leave the changes to actual dependencies in a separate PR, even `golang.org/x` and `protobuf`... Btw, why do we even have protobuf? :sweat_smile:

imiric:
OK, I'll see if I can pin the `golang.org/x` packages, but there's a few of them :sweat_smile:

As for `protobuf`:
```
> go mod why -m google.golang.org/protobuf
# google.golang.org/protobuf
github.com/loadimpact/k6/api/v1
github.com/manyminds/api2go
github.com/manyminds/api2go/routing
github.com/gin-gonic/gin
github.com/gin-gonic/gin/binding
github.com/golang/protobuf/proto
google.golang.org/protobuf/encoding/prototext
```

:disappointed:

imiric:
@na-- I version pinned some of the dependencies and the diff is much cleaner now.

All additions now are just extra unrelated files that `go mod vendor` pulls in, and all changes to `.go` files are deletions, likely because of better tree shaking:
```
> gd master --stat | grep '\.go .*+' | wc -l
0
> gd master --stat | grep '\.go .*-' | wc -l
67
```

imiric:
BTW, should this be fixed?

On this branch:
```
> go run main.go version
k6 v0.27.1 ((devel), go1.14.3, linux/amd64)
```

On `master`:
```
> go run main.go version
k6 v0.27.1 (dev build, go1.14.3, linux/amd64)
```

.circleci/config.yml
```diff
@@ -12,14 +12,7 @@ jobs:
...
             go version
-            export PATH="$GOPATH/bin:$PATH"
-            mkdir -p "$GOPATH/bin"
-            curl --fail https://raw.githubusercontent.com/golang/dep/master/install.sh | sh
-            dep version
-            dep status
-            case $(dep ensure 2>&1 | head -n 1) in Warning*) false ;; esac
-            dep check
-
+            go mod verify
```
na--, line 15: I don't think this does the same job as the previous check: 
- https://golang.org/ref/mod#go-mod-verify
- https://github.com/golang/go/issues/27348

In addition to `go mod verify`, the check should probably be something along the lines of executing `go mod vendor` again and seeing if any of the files have changed? Not sure if `go mod vendor` would always overwrite files though, so maybe even `rm -rf vendor/; go mod vendor; git add vendor; git status`?

imiric: Right, I glossed over this thinking `go mod` would simplify it :smile:

OK, I'll see what I can do.

imiric: @na-- Does 844578c1 cover this? See a failure [here](https://circleci.com/gh/loadimpact/k6/12148). Besides the `grep`, it seems more elegant than `rm`/`git` :)

This was an extreme example since tests wouldn't run either, but it should catch any minor package differences as well.

.circleci/config.yml
imiric: Mostly resolved in 8579e881.

imiric:
Unfortunately, this doesn't fix flaky tests. :laughing:

> Went through the files, the missing ones are because they are `//build +ignore` which in most cases is because those are files to generate other files so ... yeah :D.

Ah, that makes sense, thanks. I just wish there could be a way of configuring `go mod` to ignore certain files from the user's side. :-/

Serious question: what do you think about ignoring some of these at least at the repo level? Something like `/vendor/**/Makefile` or whatever the `.gitignore` syntax is. And even specific files like this 108k test data file: `vendor/github.com/dlclark/regexp2/testoutput1`. I guess we shouldn't blindly ignore all `/vendor/**/*.yml` as those could be functionally significant, but Makefiles and test data seem safe to ignore.

MStoykov:
> Serious question: what do you think about ignoring some of these at least at the repo level?

This will break the `go mod` check ... as long as they don't change I don't think it's that big of an issue ;)

na--:
> > Serious question: what do you think about ignoring some of these at least at the repo level?
> 
> This will break the `go mod` check ... as long as they don't change I don't think it's that big of an issue ;)

I'm not sure it will break the check. Considering that you can't check the integrity of the vendor folder with `go mod` commands yet, the way we'll do it is with running `go mod vendor` and seeing if anything in git has changed. But if the files are in `.gitignore`, then nothing will appear changed, even if you `go mod vendor; git add .` again :tada: 

So, @imiric, my vote is :+1: for ignoring Makefiles, READMEs, `*_test.go` files, and yes, especially single 100k+ test files :sweat_smile:

.circleci/config.yml
na--: hmm this check seem flaky, sorry - why not rely on git to check if the actual file contents in vendor have changed? that way you don't rely on the format of some text file or the output of some command...

and it will work much better with the `.gitignore` idea you shared below

na--: checking if `git status --porcelain` has any output seems good to do the job?

imiric: Yeah, the output and order for the diff might not be stable. OK, I'll try it with `git status`.

imiric: 19d8a1f6 seems like it should do it, let me know. I want to avoid `rm`/`git add` if possible.

imiric:
golangci-lint seems to be failing more often that not these days :disappointed:

I added some `.gitignore` entries and it removed about 50 files. I don't want to remove READMEs and .md files in general, since these are usually helpful when you're working with a dependency and want to quickly check the documentation. `rg`/`grep` picks up examples in the docs, so it's handy to leave them in. `*_test.go` files are always excluded by `go mod` so no need to specify those.

MStoykov:
I am not really certain the *.md files are worth it .. but can we also drop `.png` and `.txt` ?

imiric:
> can we also drop `.png` and `.txt` ?

Not sure, those could be functionally significant :smile:

imiric:
The latest `.gitignore` changes removed about 20 more files, and the additions now are mostly `*.md`, `.gitignore` and `go.mod`.

na--:
@imiric, going through all the files in just in case and I noticed  there are some surprising differences with `vendor/github.com/spf13/cobra` :confused: Any idea why so many files were deleted?

na--:
Ah, nevermind, I think I answered my own question :sweat_smile: They seem like the non-library parts, an app you can use to generate your cobra boilerplate and docs or something like that...

imiric:
Yeah, I think this is just better pruning by `go mod`. We are using the same version: `a114f312e075`.

PR #1586, opened 2020-08-12
Update README.md chocolatey link

dkocich:


na--:
Thanks for making this PR, however the package you linked to is not an official k6 package maintained by us, so we can't just link directly to it from the k6 README. IIRC, in contrast to package managers like homebrew ([example](https://github.com/Homebrew/homebrew-core/pull/57946)) and nix ([another example](https://github.com/NixOS/nixpkgs/pull/94513)), chocolately packages don't even require a code review to be updated. If so, the author of that package, who is not affiliated with us, can release a new version that contains whatever they want, including malware. 

I'm not saying they will, but until we either start maintaining the package on our own, or there's some oversight and governance from the package manager's side, it's a slight security risk to have that package be the one we link to officially.

We plan to improve our [installation documentation page](https://k6.io/docs/getting-started/installation) and start linking to that from our GitHub README: https://github.com/loadimpact/k6-docs/issues/52 . AFAIK, https://chocolatey.org/packages/k6 sits somewhere between 2 and 3, if I have to classify it according to my comment there, though probably closer to 3 :disappointed:

PR #1587, opened 2020-08-12
TestPerVUIterationsRunVariableVU was not race free ...

MStoykov:


lib/executor/per_vu_iterations_test.go
```diff
@@ -125,8 +120,8 @@ func TestPerVUIterationsRunVariableVU(t *testing.T) {
 
 	// The slow VU should complete 16 iterations given these timings,
```
imiric, line 121:
```suggestion
	// The slow VU should complete 15 iterations given these timings,
```

PR #1588, opened 2020-08-12
Es6 update

MStoykov:


MStoykov:
https://github.com/loadimpact/k6/pull/1588/commits/bbf1198665e0a1ed98e7cfd60ae1d28268cf467c is in a separate commit so it's easy to revert it later on

.circleci/config.yml
```diff
@@ -98,7 +98,6 @@ jobs:
             go version
             export GOMAXPROCS=2
             export PATH=$GOPATH/bin:$PATH
-            go test -p 2 -race -timeout 800s ./...
```
na--, line 101: I'd prefer if you comment this whole section with a TODO, not leave `test-prev-golang` but have it do nothing

MStoykov: If I remove it I need to change also the settings of the whole project to not require it for the master branch ...

na--: Ah... Then please comment it out and leave a big visible `TODO`, and also open a `v0.28.0` issue to move k6 to Go 1.15 and re-enable these tests with 1.14 when we do so

MStoykov: https://github.com/loadimpact/k6/issues/1589

Gopkg.toml
```diff
@@ -7,6 +7,10 @@
   branch = "master"
   name = "github.com/GeertJohan/go.rice"
 
+[[constraint]]
+  branch = "master"
+  name = "github.com/dop251/goja"
+
```
imiric, line 13: Weird that goja wasn't already here o.O

MStoykov: Hm ... I could've probably dropped this ... I needed it to specify a specific branch (es6) but now it's no longer required - but I will leave it as it's going to get dropped either way when we move to go modules ;)

imiric: This should probably be pinned to a specific `revision`, otherwise `dep ensure` would upgrade it I think. But yeah, moot point because of the Go modules switch.

PR #1591, opened 2020-08-14
Thread a logrus.Logger through the js package

MStoykov:
This isn't a functional change but it makes certain that all parts of
the js package that aren't for test use provided logger instead of the
global one.

Unfortunately changing to logrus.FieldLogger in lib.State turned out
to require rewriting way too many tests and as it really isn't required,

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -125,7 +125,7 @@ a commandline interface for interacting with it.`,
 			return err
 		}
 
-		r, err := newRunner(src, runType, filesystems, runtimeOptions)
+		r, err := newRunner(logger, src, runType, filesystems, runtimeOptions)
 		if err != nil {
 			return err
 		}
@@ -322,16 +322,16 @@ a commandline interface for interacting with it.`,
 		if conf.SummaryExport.ValueOrZero() != "" {
 			f, err := os.Create(conf.SummaryExport.String)
 			if err != nil {
-				logrus.WithError(err).Error("failed to create summary export file")
+				logger.WithError(err).Error("failed to create summary export file")
 			} else {
 				defer func() {
 					if err := f.Close(); err != nil {
-						logrus.WithError(err).Error("failed to close summary export file")
+						logger.WithError(err).Error("failed to close summary export file")
 					}
 				}()
 				s := ui.NewSummary(conf.SummaryTrendStats)
 				if err := s.SummarizeMetricsJSON(f, data); err != nil {
-					logrus.WithError(err).Error("failed to make summary export file")
+					logger.WithError(err).Error("failed to make summary export file")
 				}
 			}
 		}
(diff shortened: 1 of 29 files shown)


cmd/runtime_options_test.go
```diff
@@ -233,6 +234,7 @@ func testRuntimeOptionsCase(t *testing.T, tc runtimeOptionsTestCase) {
 	fs := afero.NewMemMapFs()
 	require.NoError(t, afero.WriteFile(fs, "/script.js", jsCode.Bytes(), 0644))
 	runner, err := newRunner(
+		logrus.StandardLogger(),
```
na--, line 237: Instead of `logrus.StandardLogger()`, can you use [this](https://github.com/loadimpact/k6/blob/master/lib/testutils/test_output.go#L28-L41) in the tests, with a small `newLogger` helper that uses it as output? Or even better, make a new implementation of [`logrus.FieldLogger`](https://godoc.org/github.com/sirupsen/logrus#Logger) that uses [`testing.TB`](https://golang.org/pkg/testing/#TB)?

This would hopefully greatly reduce the logging noise whenever a test fails

cmd/run.go
```diff
@@ -432,21 +432,21 @@ func init() {
 
 // Creates a new runner.
 func newRunner(
-	src *loader.SourceData, typ string, filesystems map[string]afero.Fs, rtOpts lib.RuntimeOptions,
+	logger *logrus.Logger, src *loader.SourceData, typ string, filesystems map[string]afero.Fs, rtOpts lib.RuntimeOptions,
```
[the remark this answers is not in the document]
MStoykov: Because chaging `lib.State.Logger` to `logrus.FieldLogger` turned out to be a way too big of a change, that I didn't want to waste a day on. And this where it comes from

cmd/runtime_options_test.go
MStoykov: I prefer this to be done in a separate PR ... as this is likely to take a lot of time and half of those changes are in multiple modules and more importantly not just in tests ... am literally blocked and have already made one of my PRs not mergeable (probably) because of those changes.

So I prefer if at later time someone (including me) fixes those across the board, again probably in not just one PR across the whole source base

js/common/util.go
```diff
@@ -32,17 +35,17 @@ func RunString(rt *goja.Runtime, src string) (goja.Value, error) {
 }
 
 // RunES6String Runs an ES6 string in the given runtime. Use this rather than writing ES5 in tests.
-func RunES6String(rt *goja.Runtime, src string) (goja.Value, error) {
+func RunES6String(tb testing.TB, rt *goja.Runtime, src string) (goja.Value, error) {
```
na--, line 38: Ah, we should move `RunES6String()` to `testutils` or some sub-package there... This is going to cause us to compile the whole Go `testing` in our release builds.

lib/state.go
```diff
@@ -45,6 +45,7 @@ type State struct {
 	Options Options
 
 	// Logger. Avoid using the global logger.
+	// TODO change to logrus.FieldLogger when there is time to fix all the tests
 	Logger *logrus.Logger
```
na--, line 49: I think this TODO is already done, right?

js/common/util.go
MStoykov: fixed in 251a88d4

lib/state.go
MStoykov: Unfortunately - no because in a lot of places (in tests) it is used in ways that will require even more code changes ... And when I started doing those it turned out I will need to change way too much code, and as I explained in the original commit message, this is not as important (or at all) for what we currently use it :) 

The thing that I started fixing (and likely not the only problem) is https://github.com/loadimpact/k6/blob/251a88d46ed47d917df877f79909fb7d1134f674/js/modules/k6/http/request_test.go#L293-L294 
We can get the logger the same way it is done in  https://github.com/loadimpact/k6/blob/251a88d46ed47d917df877f79909fb7d1134f674/js/console_test.go#L243-L246 but I decided that doing that will add unnecessary amount of additional work, for no practical benefit really

na-- (approved):
LGTM

PR #1596, opened 2020-08-18
Update regexp2 to latest version

MStoykov:


MStoykov:
The not RE2 changes aren't dependant on us using the latest goja, so reverting back to the old one will not mean we need to revert this as well. But are technically bugfixes and a perf improvement so ... why not update :D

PR #1599, opened 2020-08-20
Tail logs in k6

MStoykov:


cmd/cloud.go
```diff
@@ -308,7 +315,67 @@ func cloudCmdFlagSet() *pflag.FlagSet {
...
 }
 
+//nolint:gochecknoglobals
+var cloudLogsCmd = &cobra.Command{
+	Use:   "logs",
+	Short: "Tail cloud logs from the cloud",
+	Long: `Tail cloud logs from the cloud.
+This will execute the test on the Load Impact cloud service. Use "k6 login cloud" to authenticate.`,
+	Example: `
+        k6 cloud logs 123456`[1:],
```
imiric, line 325: - I'm wondering whether this makes sense as a sub-command. It would seem more intuitive if it was a `--logs` or `--tail-logs` or similar flag for the `k6 cloud` command instead, since it should only toggle the tail behavior.
- Is this example wrong? How can this command execute a new test if it asks for a test ID? Is this meant to run in a separate terminal after `k6 cloud` is run? The asciinema you showed last week did this in a single command, so I'm a bit confused by this documentation.

MStoykov: Sorry, the description is wrong. This will NOT execute the test, this will get the logs of an already executed/executing test in the cloud and just tail them on the screen, the same way the original k6cloudlogs does

imiric: I see. And it's a separate command because we want to enable retrieving logs for past runs as well. Makes sense then, but what do you think about integrating it in `k6 cloud` itself, like you showed in the asciinema? It could be disabled by default and enabled with a flag.

MStoykov: it is currently [enabled by default](https://github.com/loadimpact/k6/pull/1599/files#diff-cf75533afdaed357cb53a3fdfe363a99R251-R256), and there is no flag ... and I would argue it should be disabled by a flag not enabled by it.

imiric (changes requested):
@MStoykov I tested it on prod now and it works great, but I think we should do something about the log output obscuring the progress bar. With `k6 run` the progress bars stay pinned to the bottom and any log output is printed above them. With the current `k6 cloud` the progress bar output is mixed with the logs, and with a lot of events scrolling by, the progress bar is only visible when the logs are quiet. Even scrolling back doesn't show it as the log output overwrites it.

cmd/cloud.go
```diff
@@ -51,19 +53,24 @@ const (
 )
 
 //nolint:gochecknoglobals
-var exitOnRunning = os.Getenv("K6_EXIT_ON_RUNNING") != ""
+var (
+	exitOnRunning   = os.Getenv("K6_EXIT_ON_RUNNING") != ""
+	disableTailLogs = os.Getenv("K6_DISABLE_TAIL_CLOUD_LOGS") != ""
```
imiric, line 58: What do you think about renaming this to `K6_CLOUD_LOGS` with a default value of "1"/true, and disabling it with `K6_CLOUD_LOGS=0`? Having "disable" as part of the name is counter-intuitive to me.

imiric (approved):
LGTM, besides that env var nitpick. Thankfully that progress bar fix was straightforward and I like that both `run` and `cloud` are sharing the same progress bar rendering, even though that UI code could use some cleanup.

cmd/cloud.go
MStoykov: I really didn't want to deal with parsing env variables and cli flags ... in here. I guess you also want the cli flag to be renamed? 
I am really not against this ... it's just that because of the configuration mess this is I prefer not to make anything more complex than .. checking that something is defined, which arguably isn't great. 

I will wait for @na-- review so I do a batch of changes

na--: `K6_DISABLE_TAIL_CLOUD_LOGS` is not a good option name, for multiple _different_ reasons. `TAIL` and `tailing` will probably be very confusing to Windows users, and `DISABLE` breaks the convention we've had with negative option names like `K6_NO_USAGE_REPORT`, `K6_NO_CONNECTION_REUSE`, etc.

Moreover, negative options are confusing, since they force them to deal with double negatives, like `K6_DISABLE_TAIL_CLOUD_LOGS=false` is stupid. So, :+1: :+1: for `K6_CLOUD_LOGS` or something like that

cmd/cloud.go
```diff
@@ -305,10 +327,75 @@ func cloudCmdFlagSet() *pflag.FlagSet {
...
+	// read the comments above for explanation why this is done this way and what are the problems
+	flags.BoolVar(&disableTailLogs, "disable-tailing-logs", disableTailLogs,
+		"disable the automatic tailing of logs when a test is executed in the cloud")
+	flags.Lookup("disable-tailing-logs").DefValue = "false"
+
+	return flags
+}
+
+//nolint:gochecknoglobals
+var cloudLogsCmd = &cobra.Command{
```
na--, line 339: Given the issues we've had, why not delay this the command to get old logs for k6 v0.29.0? As far as I understand, it's not going to work very well now, right? And it's far from essential, especially when we still have `k6cloudlogs` doing essentially the same thing? So, personally, I'd vote for not including this now.

MStoykov: This is a possibility, but likely I will be able to make the changes ... today or tomorrow. 
But I agree that maybe it should be in a separate PR ...

na--: I don't doubt we can make the changes, but why not wait a version for things to be a bit more stable? There's no rush to have this command, the other use case in this PR is much more important...

cmd/cloud.go
MStoykov: I wonder whether to not disable this by default though .... While it is likely to be harder to find this way ... arguably a lot of people (using CIs for example) wouldn't really want this feature enabled by default ?

na--: That's a separate discussion, though yes, I am also not sure if logs should be enabled by default for the `k6 cloud` command :man_shrugging:

MStoykov: My problem with `K6_CLOUD_LOGS` is that this sounds like something that should be in the ... output ... also ... disabling this ... doesn't disable the cloud logs ..., it disables(or enables) the tailing of the clouds logs, the cloud logs will still be send and saved (if any)

na--: I am tentatively of the opinion that `k6 cloud` logs should be enabled by default. This will make `k6 cloud` feel much more similar to `k6 run`, which is good. And it'd be easy to turn off the logs, but a lot of k6 users might not even discover we have cloud logs if we disable them by default - not everyone is reading the k6 release notes...

na--: > My problem with K6_CLOUD_LOGS is that this sounds like something that should be in the ... output ... also ... disabling this ... doesn't disable the cloud logs ..., it disables(or enables) the tailing of the clouds logs, the cloud logs will still be send and saved (if any)

Not sure if I completely understand all your objections ("..." is usually placed at the end of a sentence :wink:), but maybe something like `K6_SHOW_CLOUD_LOGS` then? "Show" is much more understandable than "tail" for non-*nix users and it will more accurately depict what's happening, while also not having the `K6_CLOUD` prefix?

cmd/cloud.go
```diff
@@ -51,19 +53,24 @@ const (
 )
 
 //nolint:gochecknoglobals
-var exitOnRunning = os.Getenv("K6_EXIT_ON_RUNNING") != ""
+var (
+	exitOnRunning = os.Getenv("K6_EXIT_ON_RUNNING") != ""
+	showCloudLogs = os.Getenv("K6_SHOW_CLOUD_LOGS") != "false"
```
na--, line 58: A small wrapper around `strconv.ParseBool` would probably be better? I know it's a PITA to spend time on configuration in its current form, but no need to make it worse than it needs to be

MStoykov: the problem is that .. I will need to move this inside the command as well

na--: erm that's what I meant by a "small wrapper" :sweat_smile: It can be something as simple as `var showCloudLogs = getEnvBool("K6_SHOW_CLOUD_LOGS")`

MStoykov: see 7c6e60e

cmd/cloud.go
```diff
@@ -247,6 +276,14 @@ This will execute the test on the Load Impact cloud service. Use "k6 login cloud
 		var progressErr error
 		ticker := time.NewTicker(time.Millisecond * 2000)
 		shouldExitLoop := false
+		if showCloudLogs {
```
na--, line 279: Can you log a debug message here, something like `Connecting to cloud logs server...`? It would make debugging potential issues easier.

stats/cloud/logs.go
```diff
@@ -0,0 +1,212 @@
...
+	return []string{idFilter, levelFilter}, nil
+}
+
+func (c *Config) getRequest(referenceID string, start time.Duration, limit int) (*url.URL, error) {
+	u, err := url.Parse(c.LogsHost.String)
+	if err != nil {
+		return nil, fmt.Errorf("couldn't parse cloud logs host %w", err)
+	}
+
+	filters, err := parseFilters(referenceID, logrus.InfoLevel.String())
```
na--, line 129: Why are you dealing with strings? Why not just use `logrus.Info`? 

And, more importantly, why are you actually restricting the level to `info` here? I don't think k6 needs to specify a level when it's tailing logs, given that they are limited in the cloud, it should just get whatever was emitted by the cloud. Or, alternatively, this should behave like `k6 run` and show `info` level normally and `debug` if `--verbose` is enabled...

cmd/cloud.go
```diff
@@ -247,6 +276,14 @@ This will execute the test on the Load Impact cloud service. Use "k6 login cloud
...
+			go func() {
+				// TODO replace with another context
+				if err := cloudConfig.StreamLogsToLogger(context.Background(), logger, refID, 0, 100); err != nil {
```
na--, line 282: Why `100`? Why have a limit at all, given that the cloud already does a ton of log rate-limiting :confused: Having holes in the middle of your logs, potentially caused by 2 completely different places seems... unwise :confused: Can't this drop the cloud message for dropped log lines, for example?

stats/cloud/logs.go
MStoykov: carried it over from the k6cloudlogs, dropped it now

cmd/cloud.go
MStoykov: this limits how many logs will be in each message. 

There could still be holes because we couldn't get the logs fast enough. I have also removed this as I don't think it matters all that much, the dropping of the logs (because they can't be pushed to us fast enough) didn't seem to be impacted by it either way when I tested earlier, it mostly had relevance on how fast old logs were returned (where there were problems as well)

stats/cloud/logs.go
```diff
@@ -0,0 +1,193 @@
...
+	Values [][2]string       `json:"values"` // this can be optimized
+}
+
+//easyjson:json
+type msgDroppedEntries struct {
+	Labels    map[string]string `json:"labels"`
+	Timestamp string            `json:"timestamp"`
+}
+
+func getLevelsStr(level string) ([]string, error) {
```
na--, line 55: As far as I can see, you no longer need that function, it's only used in its own test.

MStoykov: this will drop the test coverage ... further :(

stats/cloud/logs.go
```diff
@@ -0,0 +1,176 @@
...
+//easyjson:json
+type msgDroppedEntries struct {
+	Labels    map[string]string `json:"labels"`
+	Timestamp string            `json:"timestamp"`
+}
+
+func (m *msg) Log(logger logrus.FieldLogger) {
+	var level string
+
+	for _, stream := range m.Streams {
```
na--, line 57: Hmm this could potentially cause messages to arrive out-of-order, right? I doubt it's a big issue, if each `msg` encapsulates only a second, since each stream would be a separate instance?

na-- (approved):
LGTM

stats/cloud/logs.go
MStoykov: I thought of this some minutes ago ... as `/query_range` apperantly doesn't return logs in any given order as well. I don't even know if tail does ... my testing - shows that - yes it does seem to come in order, but more investigation is required.

I fear I will need to get all the logs and then to order them and then print them :( But I propose that is a separate PR - this is unlikely to be a problem unless you have a lot of logs and arguably that probably is already pretty bad experience (also likely to lead to dropped logs due to k6 not being able to tail fast enough)

na--:
This doesn't work well in a non-TTY mode :sob: Try running `k6 -v cloud github.com/loadimpact/k6/samples/logging.js 2>&1 | cat` and see... Though maybe that falls under https://github.com/loadimpact/k6/issues/1580 :confused:

stats/cloud/logs.go
na--: :+1: fine for a separate PR, if needed...

MStoykov:
I don't think this PR is the place to fix progressbar issues ;)

na--:
Probably so, but this PR makes it worse, try to run the same example with k6 v0.27.1... So, yeah, the cause is https://github.com/loadimpact/k6/issues/1580, but we definitely need to fix this before v0.28.0 is officially released

MStoykov:
Again ... this is not an issue with this PR ... I am even not of the opinion this needs fixing in v0.28.0 or at least not any more than it needed until now - try running a script which runs for some time with v0.27.1. 
Yes the cloud is worse as it takes longer to run, but on a 30 minute test I would argue the difference isn't there.

na--:
I think you simply didn't run the example I suggested you run above, it isn't that it takes longer, it simply looks nasty... Anyway, I'll try to fix these things in a new PR

MStoykov:
I did run it ... the reason it looks nasty is because of 2 things:
1. the test takes long enough for there to be a couple of progressbar printings (arguably with more lines than required around them), or more accurately because the cloud needs to do stuff it takes a while to start which leads to even more progressbar printings. 
2. because now there are logs printings, this makes the progressbars to printed ... even more
The same behaviour can be seen if ran locally, just the test you proposed ... finishes locally way too fast for progressbars to be printed more than once or twice

image
image

Issue #883, opened 2018-12-19
Configuration issues


na--:
This is meant to be a tracking ("epic") issue that gathers all of the deficiencies and pain points in the current k6 configuration. Individual issues, whenever possible, should likely be tackled in their own separate issues and pull requests, unless at some point we decide it'd just be easier to overhaul everything remaining at once...

I'll start with a description of how the configuration works today AFAIK, and its shortcomings. Since things are quite complicated, I'll likely miss some aspects of it, so we should update the description or add comments in the issue when we notice something else amiss. We also have to decide how we want the configuration to behave in the future, and make the changes so we get from here to there... Maybe not immediately, but the current situation is untenable in the long run.

I'll start with the current state. Generally, users can configure k6 with a mix of 4 different option sources - CLI flags, environment variables, exported script options and global JSON config. They are hierarchical, so as described [here](https://github.com/loadimpact/k6/#script-options-and-execution-control) , CLI flags can override environment variables, which can override script options, which can override JSON config options, which can override default values .

There are various issues and complications with the actual implementation though:
- Adding a new option often necessitates changes in at least 4 different places in the k6 code ([here](https://github.com/loadimpact/k6/blob/d219b19a0353ddc73063ff3c45bc17444f0771c3/lib/options.go#L185-L275), [here](https://github.com/loadimpact/k6/blob/master/lib/options.go#L283-L389), [here](https://github.com/loadimpact/k6/blob/d219b19a0353ddc73063ff3c45bc17444f0771c3/cmd/options.go#L44-L72), [here](https://github.com/loadimpact/k6/blob/d219b19a0353ddc73063ff3c45bc17444f0771c3/cmd/options.go#L75-L98), and where we actually use it...)
- Setting default values for options is very illogical, it mostly happens the CLI flags, like [this](https://github.com/loadimpact/k6/blob/d219b19a0353ddc73063ff3c45bc17444f0771c3/cmd/options.go#L47-L54)... unless it doesn't, like [here](https://github.com/loadimpact/k6/blob/d219b19a0353ddc73063ff3c45bc17444f0771c3/cmd/options.go#L93-L95) or [here](https://github.com/loadimpact/k6/blob/d219b19a0353ddc73063ff3c45bc17444f0771c3/stats/cloud/config.go#L149-L166)...
[...]

PR #1603, opened 2020-08-23
bug/1560 tlsCipherSuites and tlsVersion marshaling

### LINKED ISSUE #1560: tlsCipherSuites and tlsVersion aren't properly marshalled in archives
If we have a script like this:
```
export let options = {
    tlsCipherSuites: [
        "TLS_RSA_WITH_RC4_128_SHA",
        "TLS_ECDHE_ECDSA_WITH_AES_256_CBC_SHA",
    ],
    tlsVersion: {
        min: "tls1.0",
        max: "tls1.2"
    }
};
```
and we archive it with `k6 archive`, we'd get something like this for the options :sob: 
```
    "tlsCipherSuites": [
      5,
      49162
    ],
    "tlsVersion": {
      "min": "",
      "max": ""
    },
```

Another way to test it:
```
$ k6 inspect 'github.com/loadimpact/k6/samples/tlsconfig.js' | jq '.tlsCipherSuites,.tlsVersion' 
[
  5,
  156
]
{
  "min": "ssl3.0",
  "max": "tls1.2"
}

$ k6 archive 'github.com/loadimpact/k6/samples/tlsconfig.js'  && (tar xfO archive.tar metadata.json | jq '.options.tlsCipherSuites,.options.tlsVersion') 
[
  5,
  156
]
{
  "min": "",
  "max": ""
}
```

notice how `tlsVersion` was reasonably ok with `k6 inspect`, there are probably 2 different bugs here...

berndhartzer:
Closes #1560

### DIFF
--- a/lib/options.go
+++ b/lib/options.go
@@ -109,9 +109,7 @@ func (s *TLSCipherSuites) MarshalJSON() ([]byte, error) {
 		}
 	}
 
-	marshaled, _ := json.Marshal(suiteNames)
-
-	return marshaled, nil
+	return json.Marshal(suiteNames)
 }
 
 func (s *TLSCipherSuites) UnmarshalJSON(data []byte) error {
--- a/lib/options.go
+++ b/lib/options.go
@@ -69,8 +69,8 @@ func (v *TLSVersion) UnmarshalJSON(data []byte) error {
 
 // Fields for TLSVersions. Unmarshalling hack.
 type TLSVersionsFields struct {
-	Min TLSVersion `json:"min"` // Minimum allowed version, 0 = any.
-	Max TLSVersion `json:"max"` // Maximum allowed version, 0 = any.
+	Min TLSVersion `json:"min" ignored:"true"` // Minimum allowed version, 0 = any.
+	Max TLSVersion `json:"max" ignored:"true"` // Maximum allowed version, 0 = any.
 }
 
 // Describes a set (min/max) of TLS versions.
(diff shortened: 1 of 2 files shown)


berndhartzer:
> Hi @berndhartzer, thanks for the PR !!! 🎉
> 
> We have fixed the `lint` to not get wrong files (or at least for your case) so please just rebase on top of the current master and force-push in order to get the [fix](https://github.com/loadimpact/k6/commit/4920d5371f01f1ec3fed761f52bc05ed4a9b73a5)

Thanks @MStoykov I've fixed it now 🙂

MStoykov (approved):

lib/options.go
```diff
@@ -96,9 +96,24 @@ func (v *TLSVersions) isTLS13() bool {
...
+		if suiteName, ok := SupportedTLSCipherSuitesToString[id]; ok {
+			suiteNames = append(suiteNames, suiteName)
+		} else {
+			return nil, errors.Errorf("Unknown cipher suite id: %d", id)
+		}
+	}
+
+	marshaled, _ := json.Marshal(suiteNames)
+
+	return marshaled, nil
```
imiric, line 114: Why not just `return json.Marshal(suiteNames)`?

lib/options.go
```diff
@@ -235,7 +250,7 @@ type Options struct {
 
 	// Specify TLS versions and cipher suites, and present client certificates.
 	TLSCipherSuites *TLSCipherSuites `json:"tlsCipherSuites" envconfig:"K6_TLS_CIPHER_SUITES"`
-	TLSVersion      *TLSVersions     `json:"tlsVersion" envconfig:"K6_TLS_VERSION"`
+	TLSVersion      *TLSVersions     `json:"tlsVersion" ignored:"true"`
```
imiric, line 253: FWIW I think we could fix the envconfig issue with a [custom decoder](https://github.com/kelseyhightower/envconfig#custom-decoders) and keep the `K6_TLS_VERSION` env var, but since it's not documented in the k6 docs, and making it work is tricky in general because of #883, I'm OK with ignoring it.

imiric (approved):
LGTM, just a nitpick and a comment.

lib/options.go
berndhartzer: Good question, no reason not to! Will update this now

lib/options.go
na--: We can't keep delaying the envconfig removal, so let's ignore this for now and fix it properly everywhere soon.

na-- (approved):

imiric (approved):
Thanks for your work!

PR #1608, opened 2020-08-26
Fix ws.Set(Timeout/Interval) panicking on float values

MStoykov:
The issue is connected to https://github.com/dop251/goja/pull/190 in the
sense that instead of the float to int conversation to result in either
error or a clamped float value, it results in a 0 int value which panics
the stdlib time package.

Changing the parameter for the timeout/interval to float both makes it
possible to take values such as 0.5, meaning half a ms, and also checks
if the value is positive so we can be certain that it won't panic k6.
If not this results in an exception in the script.

### DIFF
--- a/js/modules/k6/ws/ws.go
+++ b/js/modules/k6/ws/ws.go
@@ -381,7 +381,7 @@ func (s *Socket) trackPong(pingID string) {
 	})
 }
 
-// SetTimeout executes the provided function inside the socket's event loop after at least the provoded
+// SetTimeout executes the provided function inside the socket's event loop after at least the provided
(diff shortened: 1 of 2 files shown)


na-- (approved):

na--:
Added https://github.com/loadimpact/k6/issues/1305#issuecomment-682370864 as a reminder to use floats everywhere, when we unify the duration part of the configuration mess

MStoykov:
For the record, I didn't find another place where we use `int` instead of float

imiric (approved):
👍 though I'm not sure sub-ms values would be useful. :)

imiric (approved):

PR #1610, opened 2020-08-27
preload grafana dashboard with docker-compose setup

jeevananthank:
Today's setup of docker-compose does not preload the dashboard when running the tests locally. This PR solves the issues and detailed description is in the link below.

The grafana dashboard used here is more generic and simple and one can update as per their needs. 

https://community.k6.io/t/preload-grafana-dashbaord-with-docker-compose-setup/874


MStoykov (approved):
LGTM!

I have some reservations about us putting, even more, grafana specific things in the k6 repo ... because even now people ask us questions that are grafana/influxdb related. And to be perfectly honest I don't think we should be the one trying to answer them, and usually, it requires that we go read the grafana/influxdb documentation to figure it out, which arguably should be what, the person who asked the question, have done in the first place.

simskij (approved):
Looks good to me!

na-- (approved):
I haven't actually tried the new default dashboard, but considering we're soon going to move the docker-compose files to a separate example repo (https://github.com/loadimpact/k6/issues/1614), LGTM :sweat_smile:

simskij:
Thank you so much for your contribution, @jeevananthank! 🙏🏼

Issue #738, opened 2018-08-10
DNS round-robin ignored, always uses the same host


stephank:
k6 calls `FetchOne` on the DNS resolver:

https://github.com/loadimpact/k6/blob/d4152d20221c0c466c93fe5cabb8e8f8c885cb5b/lib/netext/dialer.go#L67

But this only ever returns the first entry from the cache:

https://github.com/viki-org/dnscache/blob/c70c1f23c5d84ab39054ec6b418a5ed0fa51cade/dnscache.go#L37

Our HAProxy frontends individually limit requests, and our load test was exceeding those limits, while the entire HAProxy cluster should've been able to handle the load. The error message k6 logged showed it was using the same host for all requests. (Or at least, the failing requests.)

As test and temporary workaround, I've adjusted our script to list the IPs and pick one at random for each VU, and the load test is now fine. (For this, we had to use `--insecure-skip-tls-verify`. This'd also break vhosts, but that wasn't a problem for us. And maybe that can also be worked around by setting `Host` manually.)

na--:
This is very loosely connected to https://github.com/loadimpact/k6/issues/726 and it may be worth it to fix both at the same time

ofauchon:
Hi, 

How would you like these issues to be solved (#726 / #738) ?
I can see multiple solutions: 

- Propose patches for inclusion in github.com/viki-org/dnscache
- Fork github.com/viki-org/dnscache in loadimpact repositories
- Write a new, k6-specific DNS resolver for the project

Thanks;
Olivier.

na--:
I'm not sure what the best solution would be, sorry. I'd start by investigating whether we need dnscache at all, since it would be better if the OS or the go standard library handled DNS caching in a manner that complied with TTL and could handle multiple IPs per domain. 

Regarding github.com/viki-org/dnscache - it seems like it's a very small library that hasn't been updated in over 5 years, so if we have to use some form of dns caching, we should probably replace it with a more fully-featured library (if such a thing exists) or write one ourselves.

na--:
This, together with https://github.com/loadimpact/k6/issues/726 will be fixed once we switch to the new caching DNS resolver that @bookmoons is currently woking on (https://github.com/bookmoons/resolvent/)

brandonh-msft:
any update on this?

na--:
[...]

Issue #1305, opened 2020-01-09
Unify duration configurations


na--:
This seemingly normal script won't work as most people expect:
```js
import http from "k6/http";

export let options = {
    duration: "5s",
    thresholds: {
        "iterations": ["count == 2"],
        "http_reqs": ["count == 2"],
    },
};

export default function () {
    let response = http.get(
        "https://httpbin.test.loadimpact.com/delay/3",
        { timeout: "2s" },
    );
}
```

Because the request timeout doesn't understand stringy durations, `"2s"` would be converted to `2` (because JavaScript :man_facepalming: ), so k6 would understand it as "2 milliseconds". So it's going to start to make a ton of requests instead of just 3...

However this script also doesn't work like how you'd expect either:
```js
import http from "k6/http";

export let options = {
    duration: 5000,
    thresholds: {
        "iterations": ["count == 2"],
        "http_reqs": ["count == 2"],
    },
};

export default function () {
    let response = http.get(
        "https://httpbin.test.loadimpact.com/delay/3",
        { timeout: 2000 },
    );
}
```

k6 runs, doesn't abort with a configuration error or something like that... but it interprets the `5000` value for `duration` as 5000 nanoseconds, i.e. 0.005 milliseconds, so it just immediately aborts... :man_facepalming: This is just a consequence of the way we've wrapped Go's `time.Duration` (which is nanosecond-based) in our [`types.Duration` custom type](https://github.com/loadimpact/k6/blob/7b6b2a684b7f26bbcf1b8549db89614ac398ad54/lib/types/types.go#L87) (that parses things like `"2s"`). Only, when it doesn't get a string, it [falls back](https://github.com/loadimpact/k6/blob/7b6b2a684b7f26bbcf1b8549db89614ac398ad54/lib/types/types.go#L118) to the nanosecond code...

So, I think we should unify and make everything saner... To do that with minimal breaking changes, I propose that we:
- use the same duration type (`types.Duration` / `types.NullDuration`?) everywhere we accept user-specified duration
- modify it so it either accepts valid string durations, or integer durations **in milliseconds**, everything else should be an error

So, the only breaking change is that users won't be able to specify the global options in nanoseconds, which won't be a big loss and I hope nobody has done...

na--:
[...]

PR #1612, opened 2020-08-31
Make DNS resolver configurable

### LINKED ISSUE #726: Unexpected DNS caching
I’d better like to understand K6s behaviour in relation to DNS caching.

I’m trying to run tests against a couple of load balanced K8S clusters. The failover is managed by a traffic manager so once the traffic manager has detected a failure in one of the clusters it will update the DNS records so that traffic only hits the healthy cluster. I’ve tried to reduce the problem, described below.

# Steps to reproduce:

1. Create a performance test to run for 5m against a particular endpoint. E.g., https://gist.github.com/paulgrav/a8b6d77639b0d54774b499c5dfd660c6
2. Ensure that `host.name` points to a CNAME that returns HTML and has a TTL of 60s.
3. Run the test.
4. After one minute, update the CNAME for `host.name` to a domain that doesn’t resolve e.g., `foo.bar.asdf`

# Actual results:

All HTTP calls in the test succeed despite the CNAME record being changed to point to a domain that doesn’t resolve.

# Expected results:

After 60s but before the test ends, there should be some test failures because the HTTP requests fail.

imiric:
This introduces a new composite DNS option (`--dns`, `K6_DNS`, and `dns` in JS options) to allow finer-grained control of k6's DNS resolution.

The sub-options are:
- `ttl`: the cache expiration period per hostname. Note that currently k6 doesn't respect the TTL in the DNS record, as that involves a deeper refactoring of the resolver that's currently in progress.
  Possible values are:
    * `inf`: hosts are cached indefinitely for the duration of `k6 run` (previous default).
    * `0`: the cache is disabled and lookups will be done on each new connection.
    * `3000`, `30s`, `1m`, etc.: cache DNS lookups for the specified period. If no unit is provided, milliseconds are assumed.
  
  Default value: `5m`.

- `select`: strategy to use when selecting an IP if more than one is returned in the response.
  Possible values are:
    * `first`: return the first IP from the response (previous default).
    * `random`: return a random IP from the response.
    * `roundRobin`: rotate the IP returned for each new connection. The current implementation is not stable and could repeat or skip IPs, but this should be good enough in most cases.
  
  Default value: `random`.

- `policy`: IP version preference.
  Possible values are:
    * `preferIPv4`: return an IPv4 address if available, otherwise fall back to IPv6.
    * `preferIPv6`: return an IPv6 address if available, otherwise fall back to IPv4.
    * `onlyIPv4`: only use IPv4 addresses even if an IPv6 address is returned.
    * `onlyIPv6`: only use IPv6 addresses even if an IPv4 address is returned.
    * `any`: no preference, return any address (previous default).

  Default value: `preferIPv4`.

### Usage examples

- CLI: `k6 run --dns='ttl=0,select=roundRobin,policy=onlyIPv4' ...`
- Environment variable: `K6_DNS='ttl=10m,select=first,policy=preferIPv6' k6 run ...`
- JS `options`:
  ```javascript
  export let options = {
    dns: {
      ttl: '1m',
      select: 'random',
      policy: 'any'
    }
  }
  ```

### Breaking changes

As discussed:
- The default `ttl` if unspecified is `5m` instead of the previous `inf` behavior.
- The default `select` if unspecified is `random` instead of the previous `first` behavior.

### Important note

These resolver settings will only be applied on new HTTP/TCP connections. Since DNS lookups are only done before the connection is established, if the test script and server enable HTTP keep-alives, the script might continue to use the established connection to the "old" host even though the DNS record had changed or the TTL expired. I.e. k6 will not actively close the connection or keep track of the TTL in the background, which it arguably shouldn't anyway.

To circumvent this the script will either have to let the connection close by itself after the idle timeout (should we make [`http.Transport.IdleConnTimeout`](https://github.com/golang/go/blob/d277a361231485999cc2b7433e3244e559c7d7da/src/net/http/transport.go#L212) configurable?), or enable the `noConnectionReuse` option to disable keep-alives altogether, which is not ideal...

This PR closes #726 and #738 if used correctly, but is more of a workaround than a proper fix.


lib/options.go
```diff
@@ -44,6 +45,21 @@ const DefaultScenarioName = "default"
 // nolint: gochecknoglobals
 var DefaultSummaryTrendStats = []string{"avg", "min", "med", "max", "p(90)", "p(95)"}
 
+type DNSConfig struct {
+	TTL null.String `json:"ttl"`
+}
+
+func (c *DNSConfig) Decode(value string) error {
```
na--, line 52: hmm instead of a custom `Decode()` function, why not implement the [`TextUnmarshaler` interface](https://golang.org/pkg/encoding/#TextUnmarshaler)?

imiric: You prefer `TextUnmarshaler` because it's more widely used?

Done in 5ef75377, but I also had to implement `UnmarshalJSON` for some reason.

imiric:
@na-- @MStoykov please take a look at the latest changes.

From manual functional tests both TTL and strategy seem to work OK, but I can't resolve the config consolidation failure. From digging into it it seems that `cliConf` is modified by `fileConf` on line 232 here:

https://github.com/loadimpact/k6/blob/5ef753777c5231f4f08d8955ce7432b95db34205/cmd/config.go#L232-L236

So when it's applied again on line 236 it actually applies the modified values, not the original ones that were passed to `getConsolidatedConfig()`.

If I'm not mistaken, to make this work we need a deep copy of the original `cliConf` which can then override everything else, but I didn't want to open that can of worms here as it can impact potentially everything depending on the current behavior.

Anyway, please test this manually yourselves. The easiest way I've found is changing `/etc/hosts` since there's no way to mock this currently (and also why the tests I mentioned in the description are useless).

lib/options.go
```diff
@@ -44,6 +45,79 @@ const DefaultScenarioName = "default"
...
+	Strategy NullDNSStrategy `json:"strategy"`
+}
+
+//go:generate enumer -type=DNSStrategy -transform=kebab -trimprefix DNS -output dns_strategy_gen.go
+type DNSStrategy uint8
+
+type NullDNSStrategy struct {
+	DNSStrategy
+	Valid bool
+}
```
imiric, line 59: This is awful, but I needed a nullable type to work with the `Apply()` mechanism... It seems any config item needs to have this logic for the overriding to work. :-/

lib/netext/dialer.go
```diff
@@ -33,34 +35,96 @@ import (
...
+		r.m.Lock()
+		defer func() {
+			r.roundRobin[host]++
+			r.m.Unlock()
+		}()
+		// TODO: The index approach is not stable and we'll need to track
+		// individual IPs... Do we really want to implement an additional cache
+		// for this, since dnscache does not expose its internal one? If so, we
+		// should scrap dnscache entirely and re-implement it here.
+		return ips[int(r.roundRobin[host])%len(ips)]
```
MStoykov, line 106: Given the size of the dnscache ... code ... I am all for just implementing our own

lib/netext/dialer.go
```diff
@@ -265,3 +321,7 @@ func (c *Conn) Write(b []byte) (int, error) {
 	}
 	return n, err
 }
+
+func init() {
+	rand.Seed(time.Now().UTC().UnixNano())
+}
```
MStoykov, line 327: I would prefer if the resolver has a random generator, and given that it needs to be concurrently called (http.batch) that will be a little more code :( 
So maybe in the next PR

lib/options.go
```diff
@@ -44,6 +45,79 @@ const DefaultScenarioName = "default"
...
+	DNSRoundRobin
+	DNSRandom
+)
+
+// nolint: gochecknoglobals
+var DefaultDNSConfig = DNSConfig{
+	TTL:      null.StringFrom("inf"),
+	Strategy: NullDNSStrategy{DNSFirst, true},
+}
+
```
MStoykov, line 72: I would prefer this to be a getter/generator/construct like ... most other such ones which just returns a new instance.

I do worry a lot with this kind of pointer heavy structures that someone will copy the value but the value will have a pointer inside and then they will just change the value to which the pointer points instead of the pointer. 
I don't think this happens here (or is even possible), but in the future, it might be and those kind of bugs are hard to see usually

lib/options.go
```diff
@@ -44,6 +45,79 @@ const DefaultScenarioName = "default"
...
+	}
+	c.Strategy = DefaultDNSConfig.Strategy
+	if strat, ok := params["strategy"]; ok {
+		if s, err := DNSStrategyString(strat.(string)); err != nil {
+			return err
+		} else {
+			c.Strategy.DNSStrategy = s
+			c.Strategy.Valid = true
+		}
+	}
```
MStoykov, line 101: can we error out on unknown fields as well?

lib/options.go
```diff
@@ -44,6 +45,79 @@ const DefaultScenarioName = "default"
...
+func (s *NullDNSStrategy) UnmarshalJSON(data []byte) error {
+	var str string
+	if err := json.Unmarshal(data, &str); err != nil {
+		return err
+	}
+	var err error
+	if s.DNSStrategy, err = DNSStrategyString(str); err == nil {
+		s.Valid = true
+	}
+	return err
```
MStoykov, line 118: it is usually better to return `nil` directly for [when it will go through a variable just in case it returns an error](https://yourbasic.org/golang/gotcha-why-nil-error-not-equal-nil/) that isn't just the interface `error`

Also, I find the code ... strange, which I noticed it

p.s. Isn't the unmarshalling of this just `string(data)` ? or does it include the quotes around it? I guess it includes the quote ... probably better to just use `json.Unsmarshal` then :thinking:

lib/options.go
```diff
@@ -521,6 +598,20 @@ func (o Options) Apply(opts Options) Options {
 		o.ConsoleOutput = opts.ConsoleOutput
 	}
 
+	// FIXME: This should really be in applyDefault(), but in some tests
+	// o.DNS hasn't been initialized yet...
+	if o.DNS == nil {
+		o.DNS = &DefaultDNSConfig
+	}
```
MStoykov, line 605: so ... here we get the address of the default DNS config and then ... change it(the config, not the address)? This doesn't seem like a good idea and will likely break with tests ... especially parallel ones

lib/options.go
MStoykov: well, we definitely need to be able to put it in the `metadata.json` and back out when making the archive and then running it ;)

core/local/local_test.go
```diff
@@ -974,6 +974,98 @@ func TestExecutionSchedulerIsRunning(t *testing.T) {
...
+				for {
+					select {
+					case err := <-errCh:
+						require.NoError(t, err)
+						return
+					case <-time.After(6 * time.Second):
+						t.Fatal("timed out")
+						return
+					}
+				}
```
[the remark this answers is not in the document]
imiric: Yeah, as I mentioned in the description, these tests are half-baked / useless right now without properly mocking the DNS server. The idea of this and the other `TestVUIntegrationDNS` test was to start the execution scheduler with a valid IP of the httpmultibin server, then update the IP to a failing one and assert on what the behavior should be with the different TTL/strategy options. But since I wasn't able to mock it properly, I left them here purely for reference.

But thanks for mentioning mocking of `Resolver.Dial`. If we bring in dnscache internally then this test would be possible, so I'll see what I can do.

cmd/options.go
```diff
@@ -235,6 +236,18 @@ func getOptions(flags *pflag.FlagSet) (lib.Options, error) {
...
+	if dns, err := flags.GetString("dns"); err != nil {
+		return opts, err
+	} else if dns != "" {
+		if err := opts.DNS.UnmarshalText([]byte(dns)); err != nil {
+			return opts, err
+		}
+		// TODO: This obviously can't work for neither option...
+		opts.DNS.TTL.Valid = flags.Changed("dns")
+		opts.DNS.Strategy.Valid = flags.Changed("dns")
+	}
```
MStoykov, line 249: I think you could try to implement [pflag.Value](https://godoc.org/github.com/spf13/pflag#Value) for example see [here](https://github.com/spf13/pflag/blob/81378bbcd8a1005f72b1e8d7579e5dd7b2d612ab/bytes.go#L11-L34)

lib/options.go
imiric: Resolved by 7578a411 ?

lib/testutils/resolver.go
```diff
@@ -0,0 +1,228 @@
...
+		n, err := s.Read(b)
+		if err != nil {
+			return
+		}
+
+		var msg dnsmessage.Message
+		// FIXME: This returns the error `unpacking Question.Name: segment prefix is reserved`
+		// https://github.com/golang/net/blob/62affa334b73ec65ed44a326519ac12c421905e3/dns/dnsmessage/message.go#L1999
+		if err := msg.Unpack(b[:n]); err != nil {
+			spew.Dump(err)
```
imiric, line 70: @na-- This is the test issue I'm having. I'm not sure why a wrong "segment prefix" is received here. I think it refers to the subnet prefix `/64` or `/128`, but a) the tests should be doing an IPv4 lookup not IPv6, and b) I don't see where this prefix is set in the request or what's actually wrong with the payload. :-/

js/runner.go
```diff
@@ -318,9 +321,50 @@ func (r *Runner) SetOptions(opts lib.Options) error {
...
+	ttl := types.NewNullDuration(0, false)
+	switch ttlS {
+	case "", "inf":
+		// cache indefinitely
+		ttl.Valid = true
+	case "0":
+		// disable cache
+	default:
+		origTTLs := ttlS
+		// Treat unitless values as seconds
```
na--, line 355: Given https://github.com/loadimpact/k6/issues/1305, I'd prefer to have be milliseconds, not seconds. Even though it makes less sense for this particular context, the real UX benefit would be to treat duration configurations the same everywhere in k6, we shouldn't introduce any more special cases

lib/testutils/resolver.go
na--: I'm missing big chunks of the background knowledge, but to be fair, I don't think this test is required for releasing the feature. I can be releases together with the rest of the DNS changes in v0.29.0

cmd/config_consolidation_test.go
```diff
@@ -375,6 +375,70 @@ func getConfigConsolidationTestCases() []configConsolidationTestCase {
...
+			},
+			exp{}, func(t *testing.T, c Config) {
+				assert.Equal(t, &lib.DNSConfig{
+					TTL:      null.StringFrom("30"),
+					Strategy: lib.NullDNSStrategy{DNSStrategy: lib.DNSFirst, Valid: true},
+				}, c.Options.DNS)
+			},
+		},
+		{
+			// FIXME: CLI arg *should* override all
```
na--, line 428: if you have a pointer, for `lib.Options.DNSConfig`, this is going to be nightmarishly difficult to fix with envconfig, so I suggest we just set it to `ignored:"true"` for now, i.e. unable to be configured by environment variables

na--: Hmm yeah, wouldn't having `NullDNSConfig` also solve of having `NullDNSStrategy`? A "higher" DNS setting will simply overwrite all "lower" DNS options with its own or the default.

lib/netext/dialer.go
```diff
@@ -33,34 +35,96 @@ import (
...
-// about a given address.
-type dnsResolver interface {
-	FetchOne(address string) (net.IP, error)
+// DNSResolver is an interface that fetches DNS information about a given host.
+type DNSResolver interface {
+	Fetch(host string) (net.IP, error)
+}
+
+type resolver struct {
+	cache      *dnscache.Resolver
```
na--, line 48: I don't think we should continue to use this library, at all. It's way of handling TTL is... not good: https://github.com/viki-org/dnscache/blob/c70c1f23c5d84ab39054ec6b418a5ed0fa51cade/dnscache.go#L46-L60

js/runner.go
imiric: Hhmm I agree with not introducing special cases, but in this case treating them as milliseconds would be counter-intuitive, especially since we don't log what the interpreted value is (maybe we should?). But OK, will change it.

core/local/local_test.go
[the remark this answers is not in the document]
imiric: The test is fixed in the latest changes, though I still want to get rid of the global var mocking, and the timings might be flaky, but I don't see a way around that.

js/runner_test.go
```diff
@@ -1349,6 +1349,73 @@ func TestVUIntegrationClientCerts(t *testing.T) {
...
+	}{
+		"inf_cache": { // IPs are cached indefinitely
+			lib.Options{},
+		},
+		"no_cache": { // every request does a DNS lookup
+			lib.Options{DNS: &lib.DNSConfig{TTL: null.StringFrom("0")}},
+		},
+		"2s_cache": {
+			lib.Options{DNS: &lib.DNSConfig{TTL: null.StringFrom("2s")}},
+		},
```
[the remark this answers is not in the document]
imiric: I added a test for "1000" in the latest changes, which is now interpreted as 1s.

lib/netext/dialer.go
imiric: In the latest changes I re-implemented the caching with an approach I discussed with Ned, so `viki-org/dnscache` is officially gone 🎉 . It doesn't use a goroutine to update all hosts when the TTL expires globally, and instead keeps track of the last lookup time per host and refreshes when that is over the TTL. There's no cache eviction, but I'm not sure it's needed since it wasn't done before either. Let me know.

lib/options.go
imiric: Not an issue anymore since I switched to using a struct value for `Options.DNSConfig` and not a pointer.

cmd/options.go
imiric: Don't think this is needed anymore, since it seems to work now after switching to a struct value instead of a pointer for `DNSConfig`.

lib/testutils/resolver.go
imiric: Not an issue anymore since I abandoned this whole DNS mock server approach, as discussed.

js/runner.go
imiric: Resolved in the latest changes.

cmd/config_consolidation_test.go
imiric: Yup, I abandoned the pointer approach... It was actually more nightmarish to deal with our insane `Apply()` mechanism than envconfig, but it seems to work now with a struct value, after a frustrating amount of trial and error. 🙈

lib/netext/dialer.go
imiric: `viki-org/dnscache` is gone in the latest changes, let me know how it looks.

imiric:
All major issues with the configuration and tests should now be resolved, and the current test failure is because of the global `netext.LookupIP` override I'm doing. This was the simplest way to mock the stdlib call in tests, as all other attempts I tried involved changing the `netext.Resolver` interface. I don't like this approach as it changes the implementation too much simply for testing purposes, while mocking `Runner.Resolver` is messy, so suggestions are welcome. :) A possible approach could be to pass the `LookupIP` function to `NewResolver()` or another constructor, but that could also be problematic.

lib/options.go
imiric: Thanks about the `err` tip. I removed this unmarshal method in the latest changes, and strategy is now unmarshalled in the `DNSConfig` one.

lib/options.go
```diff
@@ -44,6 +45,86 @@ const DefaultScenarioName = "default"
...
+		return nil
+	}
+	params, err := strvals.Parse(string(text))
+	if err != nil {
+		return err
+	}
+	return c.unmarshalDNSConfig(params)
+}
+
+func (c *DNSConfig) unmarshalDNSConfig(params map[string]interface{}) error {
```
imiric, line 113:
```suggestion
func (c *DNSConfig) unmarshal(params map[string]interface{}) error {
```

lib/options.go
```diff
@@ -44,6 +45,86 @@ const DefaultScenarioName = "default"
...
+	Valid bool
+}
+
+const (
+	DNSFirst DNSStrategy = iota + 1
+	DNSRoundRobin
+	DNSRandom
+)
+
+const DefaultDNSConfigText = "ttl=inf,strategy=first"
```
imiric, line 70: I was lazy about it, but this should probably be returned by `DNSConfig.MarshalText()`, right?

js/runner_test.go
```diff
@@ -1664,3 +1664,50 @@ func TestSystemTags(t *testing.T) {
 		})
 	}
 }
+
+// Test DNS configuration validation and that the correct resolver is initialized.
+func TestResolverConfig(t *testing.T) {
```
imiric, line 1669: Reminder to remove this test, it's not needed anymore.

lib/netext/resolver.go
```diff
@@ -0,0 +1,135 @@
...
+			resolver: res,
+			ttl:      time.Duration(ttl.Duration),
+			cm:       &sync.Mutex{},
+			cache:    make(map[string]cacheRecord),
+		}
+	} else {
+		r = &res
+	}
+
+	return r
```
na--, line 81:
```suggestion
	res := resolver{
		strategy:   strategy,
		rrm:        &sync.Mutex{},
		roundRobin: make(map[string]uint8),
	}
	if !ttl.Valid || ttl.Duration == 0 {
		return &res
	}
	return &cacheResolver{
		resolver: res,
		ttl:      time.Duration(ttl.Duration),
		cm:       &sync.Mutex{},
		cache:    make(map[string]cacheRecord),
	}
```

na-- (commented):
Haven't completely finished looking through the code, will mention anything else as individual comments

lib/netext/resolver.go
```diff
@@ -0,0 +1,135 @@
...
+	r.cm.Lock()
+	defer r.cm.Unlock()
+
+	var ips []net.IP
+	// TODO: Invalidate? When?
+	if d, ok := r.cache[host]; ok && time.Now().UTC().Before(d.lastLookup.Add(r.ttl)) {
+		ips = r.cache[host].ips
+	} else {
+		var err error
+		ips, err = LookupIP(host)
```
na--, line 108: You should unlock the mutex before you do the lookup, since that can take a long time, and only lock it again when you're updating the index.

lib/options.go
```diff
@@ -44,6 +45,86 @@ const DefaultScenarioName = "default"
...
+		return err
+	}
+	return c.unmarshalDNSConfig(params)
+}
+
+func (c *DNSConfig) UnmarshalText(text []byte) error {
+	if string(text) == DefaultDNSConfigText {
+		*c = DefaultDNSConfig()
+		return nil
+	}
```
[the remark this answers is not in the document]
imiric: It wasn't meant as an optimization, but a way to avoid more complex logic for setting default values. I'd argue this shouldn't even be in `UnmarshalText()`, but it was one of those quick workarounds for when the exact default string is used, so the order didn't matter. I'll see if I can get rid of it.

lib/options.go
imiric: Sure, implementing `String()` sounds better, thanks.

lib/netext/resolver.go
imiric: Hhmm yeah, that looks cleaner, and I could just pass `time.Duration` instead of our `types.NullDuration`.

lib/netext/resolver.go
imiric: Good idea, done in 6c4d8fd1.

lib/netext/resolver.go
imiric: Done in cf9bbb2a, thanks.

lib/options.go
```diff
@@ -44,6 +45,86 @@ const DefaultScenarioName = "default"
...
+	DNSFirst DNSStrategy = iota + 1
+	DNSRoundRobin
+	DNSRandom
+)
+
+const DefaultDNSConfigText = "ttl=inf,strategy=first"
+
+func DefaultDNSConfig() DNSConfig {
+	return DNSConfig{
+		TTL:      null.NewString("inf", false),
```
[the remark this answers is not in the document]
imiric: Resolved by 1c2396d0.

lib/options.go
imiric: Resolved by 53c94b92.

lib/options.go
imiric: Yeah, but this isn't really making it worse. We'd likely need 90% of it even if we had a sane config system, as this is just marshalling code for the internal representation, which should handle both JSON and the strvals format.

Anyway, shortened the method name in f06ad67d.

lib/options.go
imiric: I was refactoring `unmarshal()` anyway to see if I can get rid of the `*c = DefaultDNSConfig()` thing, so this was trivial to add. See acdf402b.

lib/netext/dialer.go
imiric: Why would it be a problem with `http.batch()`? I made it internal to `resolver` in f80540fa, and I haven't seen any panics. Let me know if we need something else.

lib/options.go
imiric: This is resolved now.

na--: Because there is no `Decoder` interface in the Go standard library AFAIK. The convention is that if your type can be encoded as text or decoded from text, you can implement the `ecnoding.TextMarshaler` or `ecnoding.TextUnmarshaler` and different things will now know to work with your type. For example, [envconfig](https://github.com/kelseyhightower/envconfig/#supported-struct-field-types) and even JSON: https://golang.org/pkg/encoding/json/#example__textMarshalJSON

lib/options.go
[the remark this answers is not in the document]
imiric: Yeah, so I wasn't able to remove this easily, as it causes `TestConfigConsolidation` to fail 😭 Try it for yourself, comment out this block, run the test and see all the issues... (I have an intense hatred for that specific test 😆, having seen it fail so many times. It's huge and makes it difficult to parse the errors and go to the source.)

To fix it I'd need to manually handle the default values for both fields plus their `Valid` values, which is tricky to get right without messing up some fallback logic. So I'd chalk this one up as another `TODO` for #883. Let me know if you want a comment for it or have another suggestion.

lib/options.go
imiric: Ah right, for some reason I was thinking [`json.Decoder`](https://golang.org/pkg/encoding/json/#Decoder) was an interface. Yeah, better to use an interface in this case for sure.

na-- (commented):
Hmm I don't like the global `var LookupIP = net.LookupIP`, at all. It should be injected as a dependency, I'll try to write some code to see if that would work.

imiric:
@na-- Agreed, we need to get rid of it, but I can't think of a simple approach.

We can't simply replace all of `Resolver` with a mock as that would replace the caching and strategy logic which is precisely what we want to test. :) Something like changing the `NewResolver` initializer to take a `LookupIP` function would also be tricky to mock as `js.Runner` initializes it in a couple of places.

Only thing I can think of is adding something like a `SetLookupIP()` function to the `Resolver` interface and calling that in tests, but it seems messy for just making tests work.

Thanks for the help, I'm trying some things myself.

na--:
I think I have a workable idea, will try to push a PR in a bit

na--:
@imiric, take a look at https://github.com/loadimpact/k6/pull/1627 - not sure this _exactly_ has to be the solution, but it seems reasonably fine as an alternative

js/runner.go
```diff
@@ -318,9 +322,55 @@ func (r *Runner) SetOptions(opts lib.Options) error {
...
+
 	return nil
 }
 
+func parseTTL(ttlS string) (time.Duration, error) {
+	ttl := time.Duration(0)
+	switch ttlS {
+	case "inf":
+		// cache "indefinitely"
+		ttl = time.Duration(math.MaxInt64)
```
na--, line 353: Hmm, this might be a problem :sweat_smile: In the check for cache freshness, you have `d.lastLookup.Add(r.ttl)`. But if `ttl` is already `math.MaxInt64`, then this will wrap around and become a negative number, so `time.Now().UTC()` will never be less than it, so we'll never use the cache. 

I think we should just use 1 year or something like that here, not `math.MaxInt64`

imiric: Indeed, great catch, will change.

lib/netext/resolver.go
```diff
@@ -0,0 +1,140 @@
...
+	ttl   time.Duration
+	cm    *sync.Mutex
+	cache map[string]cacheRecord
+}
+
+// NewResolver returns a new DNS resolver. If ttl is not 0, responses
+// will be cached per host for the specified period. The IP returned from
+// LookupIP() will be selected based on the given strategy.
+func NewResolver(actRes MultiResolver, ttl time.Duration, strategy lib.DNSStrategy) Resolver {
+	r := rand.New(rand.NewSource(time.Now().UnixNano())) // nolint: gosec
```
na--, line 64: This wouldn't be safe to use from multiple goroutines concurrently, see https://golang.org/pkg/math/rand/#NewSource
> Unlike the default Source used by top-level functions, this source is not safe for concurrent use by multiple goroutines.

imiric: So I guess it wasn't panicking because the resolver is used globally?

Anyway, I reused the same mutex for `r.roundRobin` in 3149b4d9. Let me know if this is sufficient.

js/runner.go
imiric: Resolved by a825a9c4.

lib/netext/resolver.go
```diff
@@ -0,0 +1,140 @@
...
+		ips = r.cache[host].ips
+	} else {
+		r.cm.Unlock() // The lookup could take some time, so unlock momentarily.
+		var err error
+		ips, err = r.resolve(host)
+		if err != nil {
+			return nil, err
+		}
+		r.cm.Lock()
+		r.cache[host] = cacheRecord{ips: ips, lastLookup: time.Now().UTC()}
```
na--, line 112: I don't think you need `.UTC()` here
```suggestion
		r.cache[host] = cacheRecord{ips: ips, lastLookup: time.Now()}
```

lib/netext/resolver.go
```diff
@@ -0,0 +1,140 @@
...
+// LookupIP returns a single IP resolved for host, selected by the configured
+// strategy. Results are cached per host and will be refreshed if the last
+// lookup time exceeds the configured TTL (not the TTL returned in the DNS
+// record).
+func (r *cacheResolver) LookupIP(host string) (net.IP, error) {
+	r.cm.Lock()
+
+	var ips []net.IP
+	// TODO: Invalidate? When?
+	if d, ok := r.cache[host]; ok && time.Now().UTC().Before(d.lastLookup.Add(r.ttl)) {
```
na--, line 102:
```suggestion
	if d, ok := r.cache[host]; ok && time.Now().Before(d.lastLookup.Add(r.ttl)) {
```
and here

na-- (commented):
I don't see any other major issues in the code, but a more thorough look and testing is definitely required.

I noticed a potential minor issue with the configuration: `k6 archive 'github.com/loadimpact/k6/samples/http_get.js'  && (tar xfO archive.tar metadata.json | jq '.options.dns')` will result in
```json
{
  "ttl": "5m",
  "strategy": "random"
}
```
Which means that the default values are encoded in the `metadata.json` `options` map not as `null` values, like other unchanged filed (i.e. ones with `Valid == false`) are, but with their actual values. This is a bit problematic, since it will mean re-runs of old cloud tests will execute with the current default DNS options, even if we change them to new defaults in the future, e.g. when we have a proper DNS resolver.

lib/netext/resolver.go
imiric: Sure, resolved by 96def6c0.

lib/netext/resolver.go
imiric: Resolved by 96def6c0.

imiric:
@na-- Yeah, that behavior is a side-effect of setting the default value with envconfig. :-/ I wouldn't say it's envconfig's fault though, as this works as expected with `k6 run`. I'll see if I can set `ignored:"true"` and handle it manually.

MStoykov:
@imiric I would argue it will be better if their a `null` and we don't support env variable for a version or two than having the env variable support but having them in the metadata.json

lib/netext/resolver.go
```diff
@@ -0,0 +1,140 @@
...
+	switch r.strategy {
+	case lib.DNSFirst:
+		return ip
+	case lib.DNSRoundRobin:
+		r.rrm.Lock()
+		// NOTE: This index approach is not stable and might result in returning
+		// repeated or skipped IPs if the records change during a test run.
+		ip = ips[int(r.roundRobin[host])%len(ips)]
+		r.roundRobin[host]++
+		r.rrm.Unlock()
```
MStoykov, line 133: I would argue this map should have a struct with the ... counter and a lock so you only lock one hosts roundRobin ... but given that I think you will need a lock on the outside map ... there might not be a performance improvement at all.

We should probably benchmark it still for the next iteration

MStoykov (commented):
LGTM. I would still prefer it if we make it so that it doesn't save the current default values in the archive before merging it. Again, even if we can't make it be configurable from the env variable for now, that IMO isn't that big of a problem, we can fix it in the next release or a PR after that before the release if it's encapsulated enough

lib/options.go
```diff
@@ -44,6 +46,117 @@ const DefaultScenarioName = "default"
...
+	return c.unmarshal(params)
+}
+
+func (c *DNSConfig) unmarshal(params map[string]interface{}) error {
+	for k, v := range params {
+		switch k {
+		case "strategy":
+			if v == "" {
+				continue
+			}
```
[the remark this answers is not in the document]
imiric: Mainly because of [this test](https://github.com/loadimpact/k6/blob/96def6c0b7dcd90ccb90346b8c5621fd91b0d0eb/lib/options_test.go#L342-L348).

lib/options.go
```diff
@@ -44,6 +46,117 @@ const DefaultScenarioName = "default"
...
+			if s, err := DNSStrategyString(v.(string)); err != nil {
+				return err
+			} else { // nolint: golint
+				c.Strategy.DNSStrategy = s
+				c.Strategy.Valid = true
+			}
```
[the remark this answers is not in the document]
imiric: Since `s` is used in the `else` block, declaring it outside "pollutes" the outer scope, IMO. This is a moot point in this case, but in general I think this is an edge case that `golint` doesn't detect properly. See https://github.com/golang/lint/issues/202 .

But sure, I can switch to that style if you think it's more readable, even though I dislike that it has functional implications in the code and shouldn't be linted...

lib/options.go
```diff
@@ -44,6 +46,117 @@ const DefaultScenarioName = "default"
...
+		case "ttl":
+			ttlv := fmt.Sprintf("%v", v)
+			if ttlv != "" {
```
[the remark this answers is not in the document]
imiric: > maybe you can just check that v is a string ... I am pretty sure it should be

It's not guaranteed actually, since something like `3000` would be interpreted as int, so type casting would be messier.

And see my other comment about not erroring out on empty values.

lib/netext/resolver.go
```diff
@@ -0,0 +1,140 @@
...
+}
+
+func (r *resolver) selectOne(host string, ips []net.IP) net.IP {
+	if len(ips) == 0 {
+		return nil
+	}
+	ip := ips[0]
+	switch r.strategy {
+	case lib.DNSFirst:
+		return ip
```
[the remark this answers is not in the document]
imiric: Yeah, this is wrong... The idea was to use at least some sane default since if `r.strategy` is empty (i.e. it doesn't match any `case`), then we at least return an IP instead of `nil`. This was mostly done to pass some test scenarios where the option isn't properly set.

Since the default was changed to `random`, this should probably fallback to that instead. I'll see if I can get rid of it though, since `r.strategy` should be valid here...

lib/netext/resolver.go
```diff
@@ -0,0 +1,140 @@
...
+	} else {
+		r.cm.Unlock() // The lookup could take some time, so unlock momentarily.
+		var err error
+		ips, err = r.resolve(host)
+		if err != nil {
+			return nil, err
+		}
+		r.cm.Lock()
+		r.cache[host] = cacheRecord{ips: ips, lastLookup: time.Now()}
+	}
```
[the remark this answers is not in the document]
imiric: > This will mean that we will make only 1 lookup request for each host on each ttl expire instead of ... 5k if there 5k VUs making the same request and the dns server isn't really fast

Hhmm but there wouldn't be 5k lookups of the same host with 5k VUs, since the first lookup that returns will write to the cache, and the rest will read it from there, that's the whole point of it :D You can confirm this with Wireshark: there's only one lookup regardless of the number of VUs, since the resolver is used globally.

Unless I'm misunderstanding your proposal, leaving the `CacheRecord` locked while a lookup is in progress would pause all requests that could've just used the cache in the meantime. But agree that we need more tests and benchmarks for this, which I'm inclined to leave for when the custom resolver is added.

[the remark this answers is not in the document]
MStoykov: Maybe your resolve is really fast ... (and maybe most resolves will be really fast and my proposal wouldn't do much). 
But in the current scenario because you unlock during the resolve other VUs could come see that there is nothing in the cache/it has expired and start doing the resolve as well.
My proposal basically moves this so, we unlock the cache manager so other VUs can come and either get another record from it or get blocked on the record's lock that is being currently resolved.
P.s. I have a mistake in the algorithm above that will lead to a dead lock, will fix it in a second

lib/netext/resolver.go
imiric: Yeah, that would be interesting to benchmark, but in general I'd like to avoid more mutexes :D 

I can add some benchmarks if you think they're required for this PR, otherwise I'd rather leave them for the proper resolver PR to avoid delaying the release further.

lib/options.go
```diff
@@ -44,6 +46,117 @@ const DefaultScenarioName = "default"
...
+	return c.unmarshal(params)
+}
+
+// UnmarshalText implements encoding.TextUnmarshaler.
+func (c *DNSConfig) UnmarshalText(text []byte) error {
+	if string(text) == DefaultDNSConfig().String() {
+		*c = DefaultDNSConfig()
+		return nil
+	}
+	params, err := strvals.Parse(string(text))
```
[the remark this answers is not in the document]
imiric: Maybe... But I don't want to introduce any potential parsing issues if `strvals` already handles it. Plus `strvals` is already used in the Kafka and InfluxDB configs, so it's not like we're adding a new dependency with this.

So I'd rather do this once we remove `strvals` with that bespoke version you guys were considering.

lib/netext/resolver.go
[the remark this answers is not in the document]
imiric: Hhmm right, I see. Would you say this is a blocker then?

lib/options.go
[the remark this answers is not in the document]
MStoykov: > any potential parsing issues if strvals already handles it. 

you are hitting the potential parsing issues ... because you have an `interface{}` instead of a string ... but w/e
> Plus strvals is already used in the Kafka and InfluxDB configs, so it's not like we're adding a new dependency with this.

yes ... and using it in one more place makes it that much harder to remove it ;) 

again this is nitpicking, I am not saying this is a required thing ... just pointing out that the usage of `strvals` in this concrete example makes other code necessary that wouldn't have been and arguably adds zero value IMO

lib/options.go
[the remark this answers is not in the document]
MStoykov: I would consider ... that `--dns strategy=,ttl=12` should be an error ;) And the problem you've "fixed" here should be fixed in the UnmarshalJSON instead

lib/netext/resolver.go
MStoykov: No ... the only blocker for me is the fact we write the default values in the metadata.json ... you can ignore everything else until that is fixed :D

MStoykov: Mostly everything else is what I have seen can be improved but arguably half of those will need benchmarking and should be left for another PR if at all

lib/options.go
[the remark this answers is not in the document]
imiric: I would consider that too :) But I can't fix this without breaking or changing the assumptions made in some config tests, and I don't want to introduce other potential issues by doing that.

> the problem you've "fixed" here should be fixed in the UnmarshalJSON instead

Why? `unmarshal()` handles both JSON and text unmarshalling, so it makes sense to keep this consistent. If anything `unmarshal()` shouldn't do any validation and the error should be returned by an eventual `Validate()` method, but I'm not sure how to easily hook that up in the current config state.

So I'd rather leave this as is and fix it once we get to #883.

lib/netext/resolver.go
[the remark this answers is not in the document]
imiric: Resolved by dbf6cad1. An empty `r.strategy` isn't the case anymore...

lib/options.go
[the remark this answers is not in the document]
imiric: Resolved by 8662aa9c.

lib/options.go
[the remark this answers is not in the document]
imiric: Resolved by 6a3a315c, thanks!

lib/options.go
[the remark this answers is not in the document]
imiric: Resolved by 6a3a315c, thanks!

js/runner.go
```diff
@@ -318,9 +321,55 @@ func (r *Runner) SetOptions(opts lib.Options) error {
...
+	r.Resolver = netext.NewResolver(r.ActualResolver, ttl, strategy)
+
 	return nil
 }
 
+func parseTTL(ttlS string) (time.Duration, error) {
+	ttl := time.Duration(0)
+	switch ttlS {
+	case "inf":
+		// cache "indefinitely"
```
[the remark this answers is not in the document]
imiric: 😄 I purposely used "indefinitely" since it implies a duration/time, whereas "infinitely" can be used in other contexts. Even though "inf" is a common constant used for this purpose, and makes sense here. And the quotes are because it's not really infinite/indefinite, just a long duration.

Anyway, sure, if you think it's confusing I'll change it to `"infinitely"`, although at that point the comment is kind of redundant. :)

cmd/options.go
```diff
@@ -91,6 +91,11 @@ func optionFlagSet() *pflag.FlagSet {
 	flags.StringSlice("tag", nil, "add a `tag` to be applied to all samples, as `[name]=[value]`")
 	flags.String("console-output", "", "redirects the console logging to the provided output file")
 	flags.Bool("discard-response-bodies", false, "Read but don't process or save HTTP response bodies")
+	flags.String("dns", lib.DefaultDNSConfig().String(), "DNS configuration. Possible ttl values are: 'inf' "+
+		"for a persistent cache, '0' to disable the cache,\nor a positive duration, e.g. '1s', '1m', etc. "+
+		"Milliseconds are assumed if no unit is provided.\n"+
+		"Possible select values to return a single IP are: 'first', 'random' or 'round-robin'.\n"+
+		"Possible policy values are: 'preferIPv4', 'preferIPv6', 'onlyIPv4', 'onlyIPv6' or 'any'.\n")
```
imiric, line 98: The mixed kebab-case and camelCase for these options is bothering me 😄

Should we rename `policy` to `ip` (short for "IP version preference") and the values to `4`, `6`, `4only`, `6only` and `any`?

lib/options.go
```diff
@@ -44,6 +47,238 @@ const DefaultScenarioName = "default"
...
+}
+
+// MarshalJSON returns the JSON representation of d
+func (d DNSPolicy) MarshalJSON() ([]byte, error) {
+	return json.Marshal(d.String())
+}
+
+// NullDNSPolicy is a nullable wrapper around DNSPolicy, required for the
+// current configuration system.
+type NullDNSPolicy struct {
```
imiric, line 107: Should we move some of these DNS config types to `lib/types`, like Mihail did in 46d7dd0c?

lib/options.go
```diff
@@ -44,6 +47,238 @@ const DefaultScenarioName = "default"
...
+	DNSPolicy
+	Valid bool
+}
+
+// UnmarshalJSON converts JSON data to a valid NullDNSStratey
+func (d *NullDNSPolicy) UnmarshalJSON(data []byte) error {
```
imiric, line 113: These marshalling methods were copied from the `DNSSelect` implementations, just with changed types. Should we DRY it up?

js/runner.go
```diff
@@ -318,9 +323,59 @@ func (r *Runner) SetOptions(opts lib.Options) error {
...
+	dnsSel := opts.DNS.Select.DNSSelect
+	if !dnsSel.IsADNSSelect() {
+		dnsSel = lib.DefaultDNSConfig().Select.DNSSelect
+	}
+	dnsPol := opts.DNS.Policy.DNSPolicy
+	if !dnsPol.IsADNSPolicy() {
+		dnsPol = lib.DefaultDNSConfig().Policy.DNSPolicy
+	}
+	r.Resolver = netext.NewResolver(r.ActualResolver, ttl, dnsSel, dnsPol)
+
```
MStoykov, line 349: nit: 
I would really prefer if this whole logic was in a separate function somewhere ... 
it looks strange as it is ... I do thin `IsADNSPolicy` should be `.Valid` as in null types ?

js/runner.go
```diff
@@ -318,9 +323,59 @@ func (r *Runner) SetOptions(opts lib.Options) error {
...
+		// Treat unitless values as milliseconds
+		if t, err := strconv.ParseFloat(ttlS, 32); err == nil {
+			ttlS = fmt.Sprintf("%.2fms", t)
+		}
+		var err error
+		ttl, err = types.ParseExtendedDuration(ttlS)
+		if ttl < 0 || err != nil {
+			return ttl, fmt.Errorf("invalid DNS TTL: %s", origTTLs)
+		}
+	}
```
MStoykov, line 375: Can we get this as a separate function in `lib.types` ? so we can use it wherever we need number to duration functionality (where the duration is ms if no unit is given)

lib/options.go
MStoykov: Ideally ... but I would argue this can be done in a second PR or at least as the last step, but if you have time - go for it

lib/options.go
MStoykov: Given that drying this up ... currently means much harder to read code I would .. go with "no". 
But maybe try it in another PR and we can see how it goes ?

lib/options.go
```diff
@@ -44,6 +47,238 @@ const DefaultScenarioName = "default"
...
+	}
+}
+
+// String implements fmt.Stringer.
+func (c DNSConfig) String() string {
+	out := make([]string, 0, 2)
+	out = append(out, fmt.Sprintf("ttl=%s", c.TTL.String))
+	out = append(out, fmt.Sprintf("select=%s", c.Select.String()))
+	out = append(out, fmt.Sprintf("policy=%s", c.Policy.String()))
+	return strings.Join(out, ",")
```
MStoykov, line 210: nit: Why ... just not a single fmt.Sprintf ?
Also .. the 2 should be 3 ;)

lib/options.go
```diff
@@ -44,6 +47,238 @@ const DefaultScenarioName = "default"
...
+	return json.Marshal(struct {
+		TTL    null.String   `json:"ttl"`
+		Select NullDNSSelect `json:"select"`
+		Policy NullDNSPolicy `json:"policy"`
+	}{
+		TTL:    c.TTL,
+		Select: c.Select,
+		Policy: c.Policy,
+	})
+}
```
MStoykov, line 224: Why ... just not setting `json:"-"` on Valid ? 
Also ... hm ... I do think you should be using NullDNSConfig so that you don't actually save the default settings in `metadata.json` ?

lib/netext/resolver.go
```diff
@@ -0,0 +1,195 @@
...
+		var err error
+		ips, err = r.resolve(host)
+		if err != nil {
+			return nil, err
+		}
+		ip4, ip6 = groupByVersion(ips)
+		r.cm.Lock()
+		r.cache[host] = cacheRecord{ip4: ip4, ip6: ip6, lastLookup: time.Now()}
+	}
+
```
MStoykov, line 128: As we discussed with @na-- previously the "policy" can be implemented here:
if you have `prefer<version>` if there are IPs from that version *only* they should be saved ... if not only the other ones,
In the case of `only<version>` you only look at .. well that one version
in case of `any` ... you just don't separate them.

This current code might come in handy once we have multiple IPs support as it looks (currently) that each VU will just have 1 IP that will have for it's whole time so you could get 1 VU with ipv4 and another with IPv6 and in the case of `preferIPv4` the second one ... still needs to get IPv6 (otherwise it will just not work ;) ) ... but I would argue that then ... the policy should be changed to an `only<version>` given that the VU in question can't actually make other requests

lib/options.go
imiric: Nah, don't think it's a big deal either, but I'll see if I have some time for it.

And I just noticed the comment typo 🤦

cmd/config_consolidation_test.go
```diff
@@ -375,6 +382,86 @@ func getConfigConsolidationTestCases() []configConsolidationTestCase {
...
+				env: []string{"K6_DNS=ttl=30,select=random,policy=any"},
+				cli: []string{"--dns", "ttl=5"},
+			},
+			exp{},
+			func(t *testing.T, c Config) {
+				assert.Equal(t, lib.DNSConfig{
+					TTL:    null.StringFrom("5"),
+					Select: lib.NullDNSSelect{DNSSelect: lib.DNSRandom, Valid: true},
+					Policy: lib.NullDNSPolicy{DNSPolicy: lib.DNSany, Valid: true},
+				}, c.Options.DNS)
```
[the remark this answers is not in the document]
imiric: Yeah, I can see how it could be confusing. But still think the flexibility outweighs any chance of confusion. It would be useful to e.g. have a base configuration in the script, but then be able to override a single sub-option via the env or CLI, instead of having to redefine everything.

Besides, the only reason `dns` is composite is to avoid adding a bunch of top-level options to our messy config system, but otherwise having them separated with a prefix (e.g. `--dns-ttl`, `--dns-policy`, etc.) would be more preferable IMO. These would also map cleanly to env vars as `K6_DNS_TTL`, etc., if we had a sane config system.

js/runner.go
imiric: Sure, I'll give it a try.

lib/options.go
imiric: 🤦  I find this pattern more readable, but sure, I'll change it, especially to avoid the array. :)

lib/options.go
imiric: It hadn't occurred to me 😄 I'll give it a try, thanks.

lib/netext/resolver.go
imiric: I didn't get the scenario in your second paragraph, but agree that it can be simplified. Will change.

cmd/options.go
imiric: True, but the mixed casing is still annoying...

lib/netext/resolver.go
MStoykov: after https://github.com/loadimpact/k6/pull/1658 (or anything like it) each VU will have an IP that it will use (unless we completely redesign it ... but having multiple IPs have ... problems). So it is possible that you will have VUs that have an ipv4 and VUs which have ipv6 addresses in the same run. But obviously if you have a VU which will use ipv4 .. it should *not* get an ipv6 address even if `preferIPv6` is configured ... 

But this case IMO will be better fixed in v0.30.0

js/runner.go
imiric: Resolved by 3f3f454b.

js/runner.go
imiric: Instead of adding a new function, I added the ms logic to `types.ParseExtendedDuration` in 2dc9f1af, which seems to work fine.

lib/options.go
imiric: Fixed by 401961f3.

lib/options.go
imiric: Resolved by 54290d8c.

Though not sure what you mean with `NullDNSConfig`, as the current `DNSConfig` is essentially that except for the name :) So the default settings aren't saved in `metadata.json` if they're not specified:

```
> go run main.go archive 'github.com/loadimpact/k6/samples/http_get.js' && (tar xfO archive.tar metadata.json | jq '.options.dns')
{
  "ttl": null,
  "select": null,
  "policy": null
}
```

lib/netext/resolver.go
imiric: Let me know if e0480405 resolves this.

The changes in #1658 will indeed need to be reconciled with the DNS `policy` option. There's also the case if `onlyIPv6` is specified, then it doesn't make sense for the VU to use an outgoing IPv4 interface. But yeah, let's leave that for a follow-up PR.

lib/options.go
imiric: Resolved by caf82ab9.

na-- (dismissed):
LGTM! :tada: I am not sure if `local.TestDNSResolver` wouldn't prove to be flaky in the future, but :crossed_fingers:, we'll address it then

lib/netext/resolver.go
```diff
@@ -162,21 +149,26 @@ func (r *resolver) selectOne(host string, ips []net.IP) net.IP {
...
 }
 
-func (r *resolver) applyPolicy(host string, ip4, ip6 []net.IP) (ip net.IP, ips []net.IP) {
-	switch r.policy {
+func (r *resolver) applyPolicy(ips []net.IP) (retIPs []net.IP) {
+	if r.policy == lib.DNSany {
+		return ips
+	}
+	ip4, ip6 := groupByVersion(ips)
+	switch r.policy { //nolint: exhaustive
```
MStoykov, line 157: nit: Can you just have the exhaustive option and ... welll panic if we get to it? This way if we add another one it won't suddenly not be detected

lib/netext/resolver.go
```diff
@@ -162,21 +149,26 @@ func (r *resolver) selectOne(host string, ips []net.IP) net.IP {
...
 	case lib.DNSpreferIPv4:
-		ip = r.selectOne(host, ip4)
-		ips = ip6
+		retIPs = ip6
+		if len(ip4) > 0 {
+			retIPs = ip4
+		}
```
MStoykov, line 162: nit: I find it strange that preferIPv4 first doesn't  set retIPs to ipv4 and then if retIPs is empty to ... actually set it to ipv6

lib/types/types.go
```diff
@@ -95,8 +95,12 @@ func (d Duration) String() string {
 }
 
 // ParseExtendedDuration is a helper function that allows for string duration
-// values containing days.
+// values containing days. If no units are provided, milliseconds are assumed.
```
MStoykov, line 98: This also applies to other configurations such as:
- setup/teardownTimeout
- gracefulStop/rampdown
- timeunit
- duration
- starttime
... more 
I am not saying this is bad idea, I am just pointing out this is breaking change even if it's likely a welcome one
Also ... as far as I can see ... this is not used consistantly so some options will be unmarshalled from JSON to this but not from the cli flags for example `min-iteration-duration` will use time.ParseDuration as this is [what pflag says](https://github.com/spf13/pflag/blob/master/duration.go#L15-L18) or maybe .. not, not certain on whether there is somethign that will check for `UnmarshalText` before that. 

But I would argue this should've been another function in this PR and have this merging be another PR in which this is more tested and looked into
cc @na--

MStoykov (dismissed):
LGTM, apart from my worries that the change to ParseExtentedDuration will break scripts.

I haven't tested it, but will try to get to it today.

lib/types/types.go
na--: hmm I think it's not going to be a big issue and we have to address https://github.com/loadimpact/k6/issues/1305 eventually... @MStoykov keep in mind that this won't actually be used for integer JS/JSON values because of https://github.com/loadimpact/k6/blob/caf82ab98ec3b5dc581519a5269c276ca05a2306/lib/types/types.go#L154-L160
So I'm not sure how much effect this will actually have

na--: but yeah, maybe we shouldn't make half-measures... so I'm fine if we use a separate function this time, with a `TODO` to switch completely to it and address #1305 in another PR

MStoykov: Looking in the code that you just linked ... I am even more of the opinion that this should be a separate functio now ... as this seems to make things ... worse :D

na--: fair enough - I can't think of an example how to break it, but it'd definitely be cleaner, so :+1: from me

cmd/options.go
imiric: OK, I made the `select` options camel case as well, so it's `roundRobin` now, but I don't see that enumer feature you mentioned @na--. `-ignorecase` seems to have been added in [a couple of PRs](https://github.com/alvaroloes/enumer/search?q=ignorecase&type=issues), but they weren't merged. Unless you're talking about a different fork, of which there are several.

We might be able to hack around it, but I'd rather leave it as is, even though I prefer lower/kebab case values.

lib/types/types.go
imiric: Hey, thanks for the heads up, I agree that it's safer to use a different function. I changed it in the recent push, along with some squashing since I had to fix the merge conflicts anyway.

lib/netext/resolver.go
imiric: Sure, fixed in the latest push.

lib/netext/resolver.go
imiric: Fixed in the latest push, though I'm not panicking but just doing the fallback assignment.

cmd/options.go
na--: https://github.com/dmarkham/enumer/pull/22 and https://github.com/dmarkham/enumer/pull/38 seem to have been the PRs and they are merged

na--: but in any case, this can happen in a separate PR, since we use enumer in other places as well

na-- (approved):
LGTM :tada:

MStoykov (approved):

PR #1615, opened 2020-09-02
Various updates

na--:
go commit by commit

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -75,17 +75,21 @@ jobs:
             rm -f *.coverage
             bash <(curl --fail -s https://codecov.io/bash)
 
+  #TODO: when we officially move to Go 1.15, re-enable test-prev-golang with 1.14
 
-  test-prev-golang:
+  test-next-golang:
     docker:
-      - image: circleci/golang:1.13
+      - image: circleci/golang:1.15
     working_directory: /home/circleci/.go_workspace/src/github.com/loadimpact/k6
     steps:
       - checkout
       - run:
-          name: Run tests with previous Go version
+          name: Run tests with next Go version
           command: |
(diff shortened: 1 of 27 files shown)


MStoykov (dismissed):

.circleci/config.yml
```diff
@@ -75,17 +75,21 @@ jobs:
             rm -f *.coverage
             bash <(curl --fail -s https://codecov.io/bash)
 
+  #TODO: when we officially move to Go 1.15, re-enable test-prev-golang with 1.14
 
-  test-prev-golang:
+  test-next-golang:
     docker:
-      - image: circleci/golang:1.13
+      - image: circleci/golang:1.15
```
imiric, line 82: Why remove testing with 1.13 though? Wouldn't it be safer to keep it and add a separate job for 1.15 (or preferably `tip`) if we want to run tests in preparation for an upgrade? Though in that case it should be optional and not fail the entire build if the `test-next-golang` job fails, until we officially support it.

na--: Because the new goja doesn't support Go 1.13 See https://github.com/loadimpact/k6/pull/1588/commits/f56ea02ec90d87f5accd0a44b07f24e4968a5011 and https://github.com/dop251/goja/issues/183

imiric: Ah, that's a shame. Mention that in the commit at least?

na--: I didn't remove Go 1.13 testing in this commit though :confused: that should have been done in https://github.com/loadimpact/k6/commit/f56ea02ec90d87f5accd0a44b07f24e4968a5011...

imiric (approved):

MStoykov (approved):

PR #1616, opened 2020-09-04
Progress bar fixes

na--:
This should fix https://github.com/loadimpact/k6/issues/1580 and a small data race on top


MStoykov (approved):
LGTM, but I haven't tested it ... and likely won't have time before 0.28.0

cmd/ui.go
```diff
@@ -81,16 +81,24 @@ func (w *consoleWriter) Write(p []byte) (n int, err error) {
...
 
 func printBar(bar *pb.ProgressBar) {
+	if quiet {
+		return
+	}
 	end := "\n"
+	// TODO: refactor widthDelta away? make the progressbar rendering a bit more
+	// stateless... basically first render the left and right parts, so we know
+	// how long the longest line is, and how much space we have for the progress
+	widthDelta := -defaultTermWidth
```
imiric, line 91: 👍 Sounds good, it would likely be easier to reason about too.

imiric (approved):
Looks much better from my tests, and nice catch of that data race. 👍

PR #1620, opened 2020-09-07
Just make LogsHost actually be what the name says

MStoykov:


na-- (changes requested):
I'd prefer to adjust the name to be `LogsURL` or something like that, and have it be the complete URL. From my experience, having part of the URL be hardcoded in the source has always led to issues...

na--:
Closing in favor of https://github.com/loadimpact/k6/pull/1625

PR #1625, opened 2020-09-10
Rename K6_CLOUD_LOGS_HOST to K6_CLOUD_LOGS_TAIL_URL

na--:
This replaces https://github.com/loadimpact/k6/pull/1620

### DIFF
--- a/stats/cloud/config.go
+++ b/stats/cloud/config.go
@@ -37,11 +37,11 @@ type Config struct {
 	ProjectID       null.Int    `json:"projectID" envconfig:"K6_CLOUD_PROJECT_ID"`
 	Name            null.String `json:"name" envconfig:"K6_CLOUD_NAME"`
 
-	Host       null.String `json:"host" envconfig:"K6_CLOUD_HOST"`
-	LogsHost   null.String `json:"-" envconfig:"K6_CLOUD_LOGS_HOST"`
-	PushRefID  null.String `json:"pushRefID" envconfig:"K6_CLOUD_PUSH_REF_ID"`
-	WebAppURL  null.String `json:"webAppURL" envconfig:"K6_CLOUD_WEB_APP_URL"`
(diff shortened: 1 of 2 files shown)


stats/cloud/config.go
```diff
@@ -37,11 +37,11 @@ type Config struct {
 	ProjectID       null.Int    `json:"projectID" envconfig:"K6_CLOUD_PROJECT_ID"`
 	Name            null.String `json:"name" envconfig:"K6_CLOUD_NAME"`
 
-	Host       null.String `json:"host" envconfig:"K6_CLOUD_HOST"`
-	LogsHost   null.String `json:"-" envconfig:"K6_CLOUD_LOGS_HOST"`
-	PushRefID  null.String `json:"pushRefID" envconfig:"K6_CLOUD_PUSH_REF_ID"`
-	WebAppURL  null.String `json:"webAppURL" envconfig:"K6_CLOUD_WEB_APP_URL"`
-	NoCompress null.Bool   `json:"noCompress" envconfig:"K6_CLOUD_NO_COMPRESS"`
+	Host        null.String `json:"host" envconfig:"K6_CLOUD_HOST"`
+	LogsTailURL null.String `json:"-" envconfig:"K6_CLOUD_LOGS_TAIL_URL"`
```
imiric, line 41: Didn't we discuss not including "tail" as part of the name? So maybe `CloudLogsURL` and `K6_CLOUD_LOGS_URL`?

na--: We discussed that "tail" shouldn't be a part of the flag name to show logs during `k6 cloud`. But given that we'll have to add another URL to retrieve the logs of old test runs in the `k6 cloud logs`, I needed to differentiate the current websocket URL with the new HTTP URL for that, whenever it's released. And I don't expect any k6 user to actually ever want to change this option, so I don't think it's such an issue.

imiric (approved):

PR #1627, opened 2020-09-11
Try to avoild global variables

na--:

### DIFF
--- a/core/local/local_test.go
+++ b/core/local/local_test.go
@@ -1044,15 +1044,11 @@ func TestDNSResolver(t *testing.T) {
 				ctx, cancel, execScheduler, samples := newTestExecutionScheduler(t, runner, logger, tc.opts)
 				defer cancel()
 
-				mr := testutils.NewMockResolver(nil)
-				defaultLookup := netext.LookupIP
-				netext.LookupIP = mr.LookupIPAll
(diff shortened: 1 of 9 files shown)


core/local/local_test.go
```diff
@@ -1044,15 +1044,11 @@ func TestDNSResolver(t *testing.T) {
 				ctx, cancel, execScheduler, samples := newTestExecutionScheduler(t, runner, logger, tc.opts)
 				defer cancel()
 
-				mr := testutils.NewMockResolver(nil)
-				defaultLookup := netext.LookupIP
-				netext.LookupIP = mr.LookupIPAll
-				defer func() { netext.LookupIP = defaultLookup }()
-
-				mr.Set("myhost", sr("HTTPBIN_IP"))
+				tb.Resolver.Set("myhost", sr("HTTPBIN_IP"))
```
imiric, line 1047: Ah, changing the `httpmultibin`s resolver, that makes sense. 👍

imiric:
I cherry-picked parts of this to finally get a passing build on #1612, so thanks! But the `httpmultibin` change wasn't needed (see my comment above) as I still needed a way to hook into `js.Runner`, which resulted in this hack: 709a984761. That should be removed eventually, but for now it seems like the least intrusive way of doing this.

PR #1628, opened 2020-09-11
Add a Counter.rps sink

na--:
This is an alias to the `Counter.rate` sink we recently re-discovered exists, which has a totally different meaning to `Rate.rate`... :man_facepalming: It's a minor UX improvement, until we can design a better thresholds architecture and fix all of the issues (https://github.com/loadimpact/k6/issues/1443#issuecomment-626957527) properly.

### DIFF
--- a/stats/sink.go
+++ b/stats/sink.go
@@ -56,9 +56,11 @@ func (c *CounterSink) Add(s Sample) {
 func (c *CounterSink) Calc() {}
 
 func (c *CounterSink) Format(t time.Duration) map[string]float64 {
+	rps := c.Value / (float64(t) / float64(time.Second))
 	return map[string]float64{
 		"count": c.Value,
-		"rate":  c.Value / (float64(t) / float64(time.Second)),
+		"rate":  rps,
+		"rps":   rps,
 	}
 }
 
--- a/stats/sink_test.go
+++ b/stats/sink_test.go
@@ -58,7 +58,7 @@ func TestCounterSink(t *testing.T) {
 		for _, s := range samples10 {
 			sink.Add(Sample{Metric: &Metric{}, Value: s, Time: now})
 		}
-		assert.Equal(t, map[string]float64{"count": 145, "rate": 145.0}, sink.Format(1*time.Second))
+		assert.Equal(t, map[string]float64{"count": 145, "rate": 145.0, "rps": 145.0}, sink.Format(1*time.Second))
 	})
 }
 


stats/sink.go
```diff
@@ -56,9 +56,11 @@ func (c *CounterSink) Add(s Sample) {
 func (c *CounterSink) Calc() {}
 
 func (c *CounterSink) Format(t time.Duration) map[string]float64 {
+	rps := c.Value / (float64(t) / float64(time.Second))
```
esquonk, line 59: I am wondering, how often is this calculated, what's the scale of `t`?

esquonk: Because a value of 1 divided by 0.5 seconds will produce `rps == 2` which is not optimal :sweat_smile:

na--: It is executed every 2 seconds currently: https://github.com/loadimpact/k6/blob/4f45b8fd009c846429e84804de1c019055c00a87/core/engine.go#L41
But yeah, that is something we have to fix in the new architecture as well...

stats/sink.go
```diff
@@ -56,9 +56,11 @@ func (c *CounterSink) Add(s Sample) {
...
 	return map[string]float64{
 		"count": c.Value,
-		"rate":  c.Value / (float64(t) / float64(time.Second)),
+		"rate":  rps,
+		"rps":   rps,
```
imiric, line 63: I finally read through all discussion on Slack, and naming this "rps" sounds confusing to me. Like you guys were discussing, RPS implies "request per second", while "rate per second" is meaningless... If anything this should probably be "cps"/"count per second".

I don't want to start a huge discussion about such a nitpick again, but as a user I would find "rps" even more confusing than "rate". At least with "rate" we can document that this means "counts per second", instead of trying to explain what "Counter.rps" means.

na--: I tend to mostly agree with you, but not too strongly, so I'm leaving this open and you, @MStoykov and @esquonk can reach a consensus next week and decide if you should merge this or not.

na--: hmm actually, thinking about this some more, I am withdrawing this PR. Completely ignoring whether `Counter.rps` is easier to grok than `Counter.rate`, adding `Counter.rps` in k6 v0.28.0 when it wasn't present in old k6 versions is a problem and for sure an extra complexity... So, I am again back on the opinion we should simply document `rate` better and fix the whole issue with better names in thresholds v2.

esquonk: If this is an issue for k6 then I guess we'll have to stick with `rate`, although I would prefer to have a different, less-ambiguous name name (`cps`?). Or drop this method completely.

na--: We can't _not_ have rate, given that we've had it and documented it for so long, and having another one, even with the perfect name, will just add extra confusion. We should document its peculiarity, support it everywhere, and figure out a better alternative in the future, when we're not burdened by the limitations we currently have.

esquonk: Yeah, I don't see a "perfect" solution here... let's leave this till Monday, maybe a fresh idea will present itself :smile:

Issue #233, opened 2017-05-24
Clean up old stuff from repo


ragnarlonn:
Shouldn't we remove these:

- /tutorials   (old, obsolete stuff, most of it)
- /samples/pantheontest.js
- ...any other obsolete samples?

- Are there any groups of files in the root that we could move to subdirectories, to make the root less cluttered?

MStoykov:
Also:
- jsdoc.json
- k6.sublime-project
- .gitmodules (to check if it won't create big problems when you checkout very old versions of k6, if not for something else at least to know it :))
- Vagrantfile
- Caddyfile
- API.md - it's old ...

Also we should probably clean .dockerignore and .gitignore

MStoykov:
Probably move the logo.png and demo.svg to a separate folder as well

MStoykov:
./external also looks suspect for being completely outdated

imiric:
While we're on the topic of demo.svg, could we consider replacing it with something less CPU taxing? My laptop fans go into overdrive whenever I browse the repo root on GitHub. :-/

On some browsers more than others, but the performance impact is noticeable on all. And I'm on a fairly decent i7 laptop, so the experience must be worse on less powerful machines.

SVG animations don't seem like an efficient solution for this type of demo. How about switching to https://asciinema.org/ or something similar? I'm not sure if their animation can be embedded in GH's README, but I wouldn't mind this opening in a new window as asciinema's UX is decent, and it's popular among open source projects.

Issue #1614, opened 2020-09-01
Clean up external examples and configs


simskij:
Continuation from #1610 where the topic was brought up regarding whether we should keep external configs and dashboards in the k6 repo or not. Our reasoning is that we'll move it to a separate repo, like `k6-influx-grafana-stack` and remove it from this repo.

The changes proposed in #1610 will be merged here for now and moved to the new repo once it's been setup.

As a part of this, we'll also delete the old `./external` folder to avoid confusion, as that code is, for all intents and purposes, dead.

PR #122, opened 2017-02-16
In-process ES6 compiler, dropping Node dependency!

uppfinnarn:
TESTING REQUESTED!

Replaces the subprocess calls to Babel with an in-process goja VM. Also replaces babel-polyfill with a submodule'd core-js copy.

I'd like to ask everyone who's able to do so to try running their scripts using this branch before merging it, because there may be subtle breakages in the process (notably, there was a [now-fixed goja issue](https://github.com/dop251/goja/issues/15) wherein incorrect code was produced, and I only found out because it choked on k6/metrics.js).

PR #1630, opened 2020-09-13
Root dir cleanup

ragnarlonn:
Fixes most of the stuff mentioned in #233 

It's annoying how many things want to have files in the project root. Can any of these be removed or moved to subdirectories?

- .circleci
- .codecov.yaml
- .editorconfig
- .golangci.yaml
- Dockerfile
- Makefile
- appveyor.yml
- build-release.sh
- docker-compose.yml (there is another in the /packaging subdir)
- gometalinter.json

If there are any external systems (like perhaps CircleCI?) that expect e.g. a YAML config in the project root but which also lets us execute shell commands before they start looking for their config, we could avoid having e.g. .circleci in the root by copying it there as a part of the CI execution.


imiric (changes requested):
Hi @ragnarlonn, thanks for the contribution!

I think there are some blockers for getting this merged:

- The `.gitignore` shouldn't be deleted or moved, as it has some useful global rules to avoid polluting the repo. Same for `.dockerignore` since the `Dockerfile` build context is currently the repo root, which it arguably always should be.
- Removal of things in `external` is currently planned as part of #1614, but only after these files are pruned and migrated to that new repo. Until then I think they should remain here.
- I don't think it's a good idea to move CI related files to locations outside of root just because we want to tidy it up. :) Having to support additional logic of moving things where they should be in CI would just introduce additional failure points for no real benefit. So I vote for keeping these files where each CI expects them.

ragnarlonn:
Hi @imiric !

I understand your points, and I really agree that complicating things (like starting to copy stuff at runtime) is usually a bad idea. I just don't like it when external apps (i.e. Docker, CI stuff) clutter the root dir of a project. It both makes it harder to get an overview of the repo structure *and* it clutters the view for new users who come to our project homepage on Github and want to see the README. We're only going to use more and more of those external apps, so if we accept *some* clutter it's a slippery slope...

This is probably also a bad idea but: it would perhaps be possible to add a couple of `RUN` lines to the Dockerfile that creates .dockerignore when you run the `docker build` command. Then the file exclusion info would exist alongside the build logic, in one single file (Dockerfile).

I checked CircleCI and they don't seem to support config.yml being located anywhere else but /.circleci so I guess we're stuck with that folder for now.

Anyway, I'll put back .dockerignore, .gitignore and external/ then. What about the others in the list?

- codecov.yaml
- .editorconfig
- .golangci.yaml
- Makefile
- appveyor.yml
- build-release.sh
- docker-compose.yml (there is another in the /packaging subdir)
- gometalinter.json

If it is possible to keep some of those outside of the root, we could create e.g. a "devops" folder for them, or something?

imiric:
> I just don't like it when external apps (i.e. Docker, CI stuff) clutter the root dir of a project.

I agree, but a lot of projects force this convention so it's difficult to workaround it. At this point I think any contributor will be used to seeing Docker, CI and build-related files (e.g. `Makefile`, `package.json`, `go.mod` in our case, etc.) in the root dir, and leaving them where people expect them to be also helps with getting up to speed with a project, as you can quickly see what build tool, CI and Docker setup they're using, which would be obscured if we moved everything to some custom directory structure.

> This is probably also a bad idea but: it would perhaps be possible to add a couple of `RUN` lines to the Dockerfile that creates .dockerignore when you run the `docker build` command. Then the file exclusion info would exist alongside the build logic, in one single file (Dockerfile).

I don't think this is possible, since the build context (and thus `.dockerignore`) is evaluated before the `Dockerfile` itself to determine what to package and upload to the server. So changing it in the `Dockerfile` wouldn't matter as the build context would've already been uploaded, which we always want to keep at a minimum to speed up builds and not pollute the image.

That said the `.dockerignore` could probably be cleaned up as `js/node_modules` is not relevant anymore since Node was dropped from the build process in #122. And also `.vagrant` since I don't think anyone uses it anymore 😆

In general I think the criteria of moving things out of root should be: can it be moved safely without requiring special logic to make it work? For custom built processes like the ones using `build-release.sh` that answer is probably "yes" (even though we're planning to remove that file altogether once we switch to GH Actions), but for others like the `Makefile` or the CI files I would say no. I'd vote for removing `.editorconfig` as well as it's not needed for Go projects, although I guess someone working on JS scripts in the repo would find it useful. `gometalinter.json` can also be removed I think.

In theory we could move all of the Docker-related files outside of the root as well, but I think that will break the use case of building directly from the Git repo URL, e.g. `docker build -t imiric/k6-git https://github.com/loadimpact/k6.git`, though there might be workarounds for that, not sure. EDIT: It's apparently [possible to specify a subdirectory as build context](https://stackoverflow.com/a/39194765/96213), but I'm not sure we should deviate from the standard just for the sake of cleanup.

> If it is possible to keep some of those outside of the root, we could create e.g. a "devops" folder for them, or something?

Sure, or maybe "build" or even "external" once we clean it up.

@MStoykov What do you think?

MStoykov:
I think that we should first remove the things that can be easily removed and move what we can outside of the project and then try to fix those. 
Some of those files will be removed/moved once we move to GitHub actions for example.
Also as a whole, I find this to not be a problem :). And I am very much in the [POLA](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) camp and I don't think that making the root directory having 10 directories/files is better than it being obvious/standard how things happen

ragnarlonn:
The reason I find it to be a problem is that I believe the project page on Github is a pretty important user funnel. You could say it's a marketing tool. I believe most of the people who visit our Github page are not going to contribute code, or even build k6. Most of the visitors are likely users, or potential users, who are interested in reading about what k6 is. They want the README information, and right now they have to scroll down a page or two to find it. This isn't excellent UX, and it is sad to see an important entry point for our users getting a bad UX. Especially when we say we're all about UX/DX!

But even if it's only 5% of the visitors who *are* going to build k6 and/or contribute code, they are of course a more important type of user, so I'm definitely not saying "screw them"! I'm just saying it might be smart to strike a balance so the Github home page both looks appealing to regular users and is possible to navigate for contributors. Consider about both target groups :)

na--:
Closing this in favor of https://github.com/loadimpact/k6/pull/1684, where I rebased this PR, fixed a few overzealous removals (we need `.dockerignore` and `.gitignore`, though I cleaned them up a bit), and cleaned up a few other things.

PR #1632, opened 2020-09-15
Marshal/Unmarshal DNS settings in a more consistan way

MStoykov:
this also doesn't write down the default values in the metadata.json
unless they were set specifically.

### DIFF
--- a/lib/options.go
+++ b/lib/options.go
@@ -73,7 +73,7 @@ const (
 	DNSRandom
 )
 
-// UnmarshalJSON converts JSON data to a valid DNSStratey
+// UnmarshalJSON converts JSON data to a valid DNSStrategy
 func (d *DNSStrategy) UnmarshalJSON(data []byte) error {
 	if bytes.Equal(data, []byte(`null`)) {
 		return nil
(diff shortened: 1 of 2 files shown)


lib/options.go
```diff
@@ -79,6 +73,55 @@ const (
 	DNSRandom
 )
 
+// UnmarshalJSON converts JSON data to a valid DNSStratey
```
imiric, line 76:
```suggestion
// UnmarshalJSON converts JSON data to a valid DNSStrategy
```

lib/options.go
```diff
@@ -97,26 +140,27 @@ func (c DNSConfig) String() string {
...
 func (c DNSConfig) MarshalJSON() ([]byte, error) {
-	var s string
-	if c.Strategy.IsADNSStrategy() {
-		s = c.Strategy.String()
-	}
 	return json.Marshal(struct {
-		TTL      string `json:"ttl"`
-		Strategy string `json:"strategy"`
+		TTL      null.String     `json:"ttl"`
+		Strategy NullDNSStrategy `json:"strategy"`
```
imiric, line 145: Aha, this was the crucial part...

imiric (approved):
Thanks!

PR #1570, opened 2020-07-23
Fix wrong name tag for HTTP metrics from redirects

imiric:
Closes #1474

The root cause was reusing the same 'name' tag for all requests processed by a transport. In the case of redirects and digest auth the transport would process more than one request and needs to update the 'name' tag in the same way as 'url' is. For the initial request we need the originally parsed one to be consistent with the credential masking feature (as this gets stripped in the request passed to `RoundTrip()`), and tags for subsequent requests will be dynamically generated.

I'm not sure if this confirms @MStoykov's gut feeling that it will break aggregation somehow, but I'd argue this is the correct approach so we'll have to deal with it. :smile:

I didn't add more tests for this as the current ones are quite thorough.

PR #1633, opened 2020-09-16
Fix sending empty metric tags for failed requests

imiric:
This fixes an issue introduced in #1570 where metrics for failed requests were being sent with missing `url`, `method` and `name` tags, causing requests to show up with "0" tags in the cloud.

### DIFF
--- a/lib/netext/httpext/request_test.go
+++ b/lib/netext/httpext/request_test.go
@@ -181,7 +181,10 @@ func TestMakeRequestTimeout(t *testing.T) {
 	logger := logrus.New()
 	logger.Level = logrus.DebugLevel
 	state := &lib.State{
-		Options:   lib.Options{RunTags: &stats.SampleTags{}},
+		Options: lib.Options{
+			RunTags:    &stats.SampleTags{},
+			SystemTags: &stats.DefaultSystemTagSet,
+		},
 		Transport: srv.Client().Transport,
 		Samples:   samples,
 		Logger:    logger,
@@ -190,7 +193,7 @@ func TestMakeRequestTimeout(t *testing.T) {
 	req, _ := http.NewRequest("GET", srv.URL, nil)
 	var preq = &ParsedHTTPRequest{
 		Req:     req,
-		URL:     &URL{u: req.URL},
+		URL:     &URL{u: req.URL, URL: srv.URL},
 		Body:    new(bytes.Buffer),
 		Timeout: 10 * time.Millisecond,
 	}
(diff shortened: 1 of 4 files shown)


lib/netext/httpext/transport.go
```diff
@@ -130,9 +130,6 @@ func (t *transport) measureAndEmitMetrics(unfReq *unfinishedRequest) *finishedRe
 			}
 		}
 
-		if enabledTags.Has(stats.TagMethod) {
-			tags["method"] = unfReq.request.Method
-		}
```
MStoykov, line 135: I think this needs to stay ... as I am pretty sure if you have `POST -- 301 redirect --> GET` the second request will still be tagged as  `POST` ... I think

imiric: OK, reverted. We should have a test for this, but I don't want to delay the canary further.

MStoykov: Can you start on the test at least, we are not in that much hurry ;)

lib/netext/httpext/request_test.go
```diff
@@ -199,6 +202,20 @@ func TestMakeRequestTimeout(t *testing.T) {
...
+	sampleCont := <-samples
+	allSamples := sampleCont.GetSamples()
+	require.Len(t, allSamples, 8)
+	expTags := map[string]string{
+		"error":      "context deadline exceeded",
+		"error_code": "1000",
+		"status":     "0",
+		"method":     "GET",
+		"url":        srv.URL,
+		"name":       "",
```
MStoykov, line 214: Shouldn't this be equal to `url` ?

imiric: Not according to the `preq.URL.Name != preq.URL.Clean()` check. If I set `Name: srv.URL` on line 196, then the `name` tag is skipped entirely. If I set it to something else, then it does get tagged with the custom value.

Since this is a unit test, I think this is the expected behavior when the `name` tag is not set manually.

lib/netext/httpext/transport.go
imiric: I don't think it's worth delaying the release for this one specific edge case test, but sure, I'll add it.

lib/netext/httpext/request_test.go
MStoykov: Yeah ... but it should have `name` ... and it should be equal to the `url`.
Otherwise, it will ... just not have a `name` at all. I would actually expect that in this case `name` isn't set at all

imiric: Yes, `name` is not set if I set `Name: srv.URL`. Would you prefer that change?

lib/netext/httpext/transport.go
imiric: Resolved by af0f5f0c.

You were right, removing this would've broken the second request's `method` tag. 👍

lib/netext/httpext/request_test.go
MStoykov: I just want every request metric to have `name`, `url` and `method`, just as before ;) 
Currently if the request is with status `0` for some reason and doesn't get to the transport it will not have `name` the same way that this actually fixes this for the other tags (`method` and `url`) it should fix it for `name`

imiric: > I just want every request metric to have name, url and method, just as before ;)

This PR fixes that. The current behavior of an empty `name` tag (note: not missing, but _empty_) is a different issue which already existed even before #1570. You can confirm this by backporting this test change to e.g. `v0.27.0`.

The issue you fixed in #1634 is separate to this fix, but I think it also makes sense. So let's merge #1634, and then this PR?

MStoykov (approved):

PR #1634, opened 2020-09-16
Set url, name and method only once and fix name being empty

MStoykov:


imiric (approved):
This looks like an improvement, but I'm not sure it won't break something else since this is apparently not tested thoroughly.

Issue #1385, opened 2020-04-03
Add TLS keylogger support


MStoykov:
## Feature Description
Currently if you are using https you have only `--http-debug=full` to see what the request/response are ... this is not terrible, but given that it has issues and that with big or binary response/request it is nearly unusable, I would argue that it is better if users can use wireshark(or other tools with similar capabilities).

Luckily for some time now (go1.8) golang has support to dump the tls keys it uses in a [format](https://developer.mozilla.org/en-US/docs/Mozilla/Projects/NSS/Key_Log_Format) that wireshark [understands](https://jimshaver.net/2015/02/11/decrypting-tls-browser-traffic-with-wireshark-the-easy-way/).

So I propose we add a flag to enable logging those keys to a file using the [tls.KeyLogger](https://golang.org/pkg/crypto/tls/#example_Config_keyLogWriter)

A [repo with a basic implementation and explanation](https://github.com/joneskoo/http2-keylog)

jayjayswal:
is this still open? Can i work on this?

MStoykov:
The whole k6 team is working on getting a very big pr by number 1007 (not linking it as it already is linked from thousands of places) to be merged - so we aren't working on anything else up until then.

You can do it, it is somewhat likely though that we won't merge it very fast :( and we might not be able to give you much feedback in the next month or two, but other than that, this shouldn't be that hard (famous last words :D). 

So go for it if you want or need it ;).

You will need to do something like we do for [InsecureSkipTLSVerify](https://github.com/loadimpact/k6/blob/8d1a038d79c06237158fff2dcdc7cf6a3715ccf3/js/runner.go#L163) but make certain that it isn't settable from inside the script `options`, as that will be a requirement for us to merge it :D

jayjayswal:
```
Environment variable: K6_LOG_TLS_KEY
CLI Flag: --log-tls-key 
```
Is this naming convention okay?
Also what I understood from the last two-line, the "LogTLSKey" option won't be available in exported script options(.js file). Correct? If yes, will it be settable from the config.json file?

MStoykov:
Well, we would certainly prefer it not to be, as it will need to be disabled in the cloud either way. 

I personally think this falls under the "this is only for debugging and will be terrible for normal use" category. This definitely will be faster than `--http-debug` but still probably far from cheap and IMO will generate A LOT of data in a normal run - so it being configurable through the script or config file(which is the same) seems like a bad idea either way.

[...]

PR #1636, opened 2020-09-21
Add option for extra certificate authorities.

calavera:
This allows using custom certificates without having to install their CAs in the system pool, and without having to skip their verification.

The configuration is something like this:

```
export let options = {
  tlsExtraCAs: [
      open(./ca1.pem),
      open(./ca2.pem)
  ]
}
```

I'm not completely sure how to add tests for this, tbh. Help very welcome

Signed-off-by: David Calavera <david.calavera@gmail.com>

### DIFF
--- a/js/runner.go
+++ b/js/runner.go
@@ -23,6 +23,7 @@ package js
 import (
 	"context"
 	"crypto/tls"
+	"crypto/x509"
 	"encoding/json"
 	"fmt"
 	"net"
@@ -158,6 +159,21 @@ func (r *Runner) newVU(id int64, samplesOut chan<- stats.SampleContainer) (*VU, 
 		}
 	}
 
+	pool, _ := x509.SystemCertPool()
+	if pool == nil {
+		// SystemCertPool is not available on Windows.
+		// x509.SystemCertPool will always fail in that environment.
+		// The error can be ignored, instead, we initialize a new pool.
+		pool = x509.NewCertPool()
+	}
+	if cas := r.Bundle.Options.TLSExtraCAs; cas != nil {
+		for _, pem := range cas {
+			if !pool.AppendCertsFromPEM([]byte(pem)) {
+				return nil, errors.New("there are no CAs to add to the system pool")
+			}
+		}
+	}
+
 	dialer := &netext.Dialer{
 		Dialer:    r.BaseDialer,
 		Resolver:  r.Resolver,
@@ -172,6 +188,7 @@ func (r *Runner) newVU(id int64, samplesOut chan<- stats.SampleContainer) (*VU, 
 		Certificates:       certs,
 		NameToCertificate:  nameToCert,
 		Renegotiation:      tls.RenegotiateFreelyAsClient,
+		RootCAs:            pool,
 	}
 	transport := &http.Transport{
 		Proxy:               http.ProxyFromEnvironment,
--- a/js/runner.go
+++ b/js/runner.go
@@ -159,21 +159,6 @@ func (r *Runner) newVU(id int64, samplesOut chan<- stats.SampleContainer) (*VU, 
 		}
 	}
 
-	pool, _ := x509.SystemCertPool()
-	if pool == nil {
-		// SystemCertPool is not available on Windows.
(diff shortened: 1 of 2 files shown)


lib/options.go
```diff
@@ -327,6 +327,7 @@ type Options struct {
 	TLSCipherSuites *TLSCipherSuites `json:"tlsCipherSuites" envconfig:"K6_TLS_CIPHER_SUITES"`
 	TLSVersion      *TLSVersions     `json:"tlsVersion" ignored:"true"`
 	TLSAuth         []*TLSAuth       `json:"tlsAuth" envconfig:"K6_TLSAUTH"`
+	TLSExtraCAs     []string         `json:"tlsExtraCAs" envconfig:"K6_TLS_EXTRA_CAS"`
```
MStoykov, line 330: You aren't merging the settings between the `json` and the `env`, please look at how it's done for other configurations and copy-paste them
You also don't support this to be configurable from a cli option, although I don't think that is required

js/runner.go
```diff
@@ -158,6 +159,21 @@ func (r *Runner) newVU(id int64, samplesOut chan<- stats.SampleContainer) (*VU,
 		}
 	}
 
+	pool, _ := x509.SystemCertPool()
+	if pool == nil {
+		// SystemCertPool is not available on Windows.
+		// x509.SystemCertPool will always fail in that environment.
+		// The error can be ignored, instead, we initialize a new pool.
+		pool = x509.NewCertPool()
```
MStoykov, line 167: That will probably need testing ... :sob:

MStoykov (commented):
Hi, we just released v0.28.0 (as you've probably noticed), and will have some cleanup tasks, so we might not be .. very fast in answering ;(/

I have left a couple of comments ... additionally I would have to add that the ~biggest blocker (IMO) is that this won't work ... with `k6 cloud` or `k6 archive` all that well and that seems kind of important to me.~ None of this is true as you are using `open` so it will be part of the archive so that isn't a blocker 

~When I have time I will try to look into a possible solution, but you can work on that if you want as well. The idea is that `k6 archive` (unless bugs) will make an `archive.tar` that you can move to another machine and `k6 run archive.tar` and it will .. work :tm: . So for example in this case that means that we need to actually put the CA roots in the archive and then load them from there ... which might turn out to be harder than expected as I don't think we have a similar problem for another option.
`k6 cloud` is basically `k6 archive`++~

lib/options.go
calavera: good call, I have not looked at that code before, and I need to do it

js/runner.go
calavera: yeah, I've done it in the past, but I'll have to check how tests bootstrap data in this project to do that.

MStoykov:
Hi @calavera , sorry for the long radio silence, are you still willing to work on this, if not we can probably pick it up?

It seems I was *very* wrong in my previous review, and the main blocker(not working with k6 cloud/archive) is actually not a blocker as you are using `open` and that already does what is required :tada: and then just using the contents of the `pem` files instead of their names on disk in the configuration.

This whole PR definitely needs at least 1 unit test though

Also because of the way `options` is saved as JSON in the archive, the whole `pem` file will be inside of it. This also makes it kind of hard to use the environmental variable as you will need to have the value be the contents of the `pem` file, not its name(having the name will require support for the archive/cloud).

I am now wondering if instead of this being in `options` we shouldn't add a new module "k6/tls" and have a way to configure it from there as in(names are illustrative only):

```javascript
import tls from "k6/tls"
tls.addExtraCA(open("./ca.pem"));
```
#1385 might be better off being in there as well.
Although at that point this might be too involved and actually should be left for when we design the new HTTP API? Arguably this could be how you configure it globally and when we are designing the API we will have a way to get a `new tls.Config()` that can be configured as well. 

cc @na-- @imiric

calavera:
> are you still willing to work on this, if not we can probably pick it up?

please, feel free to take over these changes. I'm too busy with other things to be a great driver for them. Don't hesitate to close this PR if you think this would be better implemented in a different way.

PR #1637, opened 2020-09-22
Fix/726 738 dns cache refresh impr

MStoykov:
Things that need to be done:

- [ ] Dialer needs a logger
- [ ] remove no longer needed Resolver type (this would've broken a lot more code so I deprioritized it), or rewrite the new code to be called `Resolver`
- [ ] Tests Tests Tests
- [ ] make the dns cache shared between VUs again


lib/netext/dialer.go
```diff
@@ -44,13 +49,162 @@ type Dialer struct {
...
+		ttl: ttl,
+		dnsCache: DNSCache{
+			v4: &dnsCache{
+				RWMutex: sync.RWMutex{},
+				cache:   make(map[string]*cacheRecord),
+			},
+			v6: &dnsCache{
+				RWMutex: sync.RWMutex{},
+				cache:   make(map[string]*cacheRecord),
+			},
```
imiric, line 92: Why are separate caches needed for v4 and v6?

vendor/github.com/benburkert/dns/message.go
```diff
@@ -0,0 +1,975 @@
...
+	errTooManyQuestions   = errors.New("too many Questions to pack (>65535)")
+	errTooManyAnswers     = errors.New("too many Answers to pack (>65535)")
+	errTooManyAuthorities = errors.New("too many Authorities to pack (>65535)")
+	errTooManyAdditionals = errors.New("too many Additionals to pack (>65535)")
+	errFieldOverflow      = errors.New("value too large for packed field")
+	errUnknownType        = errors.New("unknown resource type")
+)
+
+// Message is a DNS message.
+type Message struct {
```
imiric, line 117: I don't think we should be using this message implementation. It seems to be an old copy of the one [currently used](https://github.com/golang/net/blob/master/dns/dnsmessage/message.go) in stdlib, and we should rely on that one instead. This probably means that we can't use `benburkert/dns` as a library and will need to re-implement what we need ourselves.

imiric (commented):
I ran a few tests on this branch and went over the code briefly, and it doesn't seem to be using the cache at all now(?). I don't see where you're populating `Dialer.dnsCache`, so all my requests hit the DNS server regardless of the config.

I'll review this again once it's more fleshed out. For now only the `dnsmessage` comment seems like a blocker.

lib/netext/dialer.go
MStoykov: We get separate requests for them so it was easier to have them separate

imiric:
Closing as discussed, see #1751.

PR #1639, opened 2020-09-23
Add allowedLabels to remove unwanted labels from loki

MStoykov:


log/loki.go
```diff
@@ -128,15 +163,47 @@ func LokiFromConfigLine(ctx context.Context, fallbackLogger logrus.FieldLogger,
...
+			value = value[:len(value)-1]
+			ended = true
+		}
+		if value == "" {
+			return nil, 0, fmt.Errorf("array item in %s need to not be empty", key)
+		}
+		result = append(result, value)
+	}
+	if !ended {
+		return nil, 0, fmt.Errorf("%s needs to be in the form [value1,value2,...,valueN], but it didn't end", key)
```
imiric, line 203: Couldn't you also check if `value[len(value)-1] != ']'` in the initial check on line 179 and return the same message from there?

log/loki.go
```diff
@@ -128,15 +163,47 @@ func LokiFromConfigLine(ctx context.Context, fallbackLogger logrus.FieldLogger,
...
+	value = value[1:]
+
+	result = append(result, value)
+	for i++; i < len(args) && !ended; i++ {
+		value = args[i]
+		if value == "" {
+			return nil, 0, fmt.Errorf("array item in %s need to not be empty", key)
+		} else if value[len(value)-1] == ']' {
+			value = value[:len(value)-1]
+			ended = true
```
imiric, line 195: This parsing seems convoluted to me... Why do you need to go over `args` at all? Wouldn't stripping the square brackets and doing `strings.Split(value, ",")` accomplish the same thing?

log/loki.go
```diff
@@ -128,15 +163,47 @@ func LokiFromConfigLine(ctx context.Context, fallbackLogger logrus.FieldLogger,
...
+		value = value[:len(value)-1]
+		ended = true
+	}
+	value = value[1:]
+
+	result = append(result, value)
+	for i++; i < len(args) && !ended; i++ {
+		value = args[i]
+		if value == "" {
+			return nil, 0, fmt.Errorf("array item in %s need to not be empty", key)
```
imiric, line 192:
```suggestion
			return nil, 0, fmt.Errorf("array item in %s must not be empty", key)
```

Same for line 198.

log/loki_test.go
```diff
@@ -101,6 +108,77 @@ func TestSyslogFromConfigLine(t *testing.T) {
...
+		args       []string
+		result     []string
+		resultI    int
+		err        bool
+	}{
+		{
+			key:     "test",
+			value:   "[some",
+			i:       2,
+			args:    []string{"else=asa", "e=s", "test=[some", "else]"},
```
imiric, line 124: These tokens don't look right to me, since you're treating the comma separating each value as the same separator for array values, which goes back to the convoluted parsing comment.

So I'd prefer to see `[]string{"else=asa", "e=s", "test=[some,else]"}` here.

log/loki.go
```diff
@@ -269,6 +336,34 @@ func (h *lokiHook) loop() {
 	}
 }
 
+func (h *lokiHook) filterLabels(labels map[string]string, msg string) string {
```
imiric, line 339: Can you add unit tests for this?

log/loki.go
```diff
@@ -269,6 +336,34 @@ func (h *lokiHook) loop() {
...
+	if h.allowedLabels == nil {
+		return msg
+	}
+	var b strings.Builder                  // TODO reuse
+	keys := make([]string, 0, len(labels)) // TODO reuse
```
imiric, line 344: One word `TODO`s aren't very helpful if someone other than you will work on them. :)

Can you add more details about why and how these should be reused?

imiric (changes requested):
Mostly LGTM, besides the parsing and tests comment.

log/loki.go
MStoykov: I can't because I have already `args = strings.Split(line, ",")`  early on .. so there are no more `,` in the values ;).
The alternative to this was to rewrite the whole code to read byte by byte and when it gets to `,` or `[` to change what I do or something like that. This will probably be a great idea when we rewrite it to be used everywhere, but for now, it seems like too much work.

log/loki.go
MStoykov: I would still need this check though as I do need to go through the rest of the args (see https://github.com/loadimpact/k6/pull/1639/files#r495747384) to see when the array ends.

log/loki.go
MStoykov: I don't want to add `how` as that is likely to change, for example I though of `sync.Pool` when I added in the begining but later decided it will probably better to be a struct field. 

Both of those variables (especially the `strings.Builder`, although I don't know if reusing it will help in performance :shrug: ) will be used all the time in case of a need to filter labels and it's very likely that they will have some similar *bounded* size.

log/loki_test.go
MStoykov: see https://github.com/loadimpact/k6/pull/1639/files#r495747384

log/loki.go
imiric: Aha, right. Though a scanner implementation will probably be simpler than this, and shouldn't be too much work. I would prefer if we do it now instead of merging this, but I leave it up to you.

log/loki.go
imiric: Sure, but at least add some more detail than a single word. If you're concerned about performance than mention that as the "why". I also don't think detailing the "how" is helpful as it might distract from the fix, but `TODO`s are easier to address if they're actionable and detailed, as you've explained it here, but maybe more succinctly.

log/loki.go
MStoykov: fixed in 286fa76

log/loki.go
MStoykov: fixed in a93d7db

log/loki.go
MStoykov: fixed in 15fcb34

imiric: Dude, how is that simpler? 😆 It's less hacky, I suppose, but it looks more complex than before. Why not use [`bufio.Scanner`](https://golang.org/pkg/bufio/#Scanner) for this? I was picturing this would be much simpler with that approach.

MStoykov: I tried using [strings.Reader](https://golang.org/pkg/strings/#Reader) ( bufio.Scanner doesn't really help in any way :D). But in the end, I needed to add runes to each other to build a string to get the final keys and values out of one rune at a time (there is no "which index are we at") ... which seemed terrible and still had more issues, so I dropped it for the simplest scanning of one symbol at a time.

I don't think it's simpler but ti let's build on top of it to have `"values with , inside"` and `{dict:values}` and such.

imiric: OK, if you think this is an improvement I'm fine with it, but I'm not sure having a custom parser for this is a good idea.

log/tokenizer_test.go
```diff
@@ -0,0 +1,44 @@
+package log
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestTokenizer(t *testing.T) {
+	tokens, err := tokenize("loki=something,s.e=2231,s=12,12=3,a=[1,2,3],b=[1],s=c")
```
imiric, line 10: I think we need much more tests here, for error cases, etc.

imiric (approved):

log/loki.go
MStoykov: well .. you didn't like the previous less custom ... we both don't like the strvals array syntax of `array[0]=2,array[2]=1` meaning `array=["2","","1"]`... so I don't see another option :)

log/tokenizer_test.go
MStoykov: Most useful cases are in the test of the lokiHook itself. When/if we move it to a separate package for reusage - yes there should be even more test cases (for the record I did test it somewhat manually with this one testcase, just as you see the result is pretty .. big ;D)

Issue #1247, opened 2019-11-15
Improve security and automation of official binary releases


na--:
As pointed out by [this forum post](https://community.k6.io/t/signature-verification-disabled-for-rpm-packages/305), currently our RPM packages are not signed :disappointed: 

I think that during the transition away from AppVeyor/CircleCI (#959), we should also make a substantial improvement (mostly in terms of security) and upgrade in how we build our packages:
- start properly signing them - both the Linux ones and the Windows ones (https://github.com/loadimpact/k6/issues/1034 - which we've also had other issues with the bintray hosting before...)
- maybe investigate if we can also make the bulds reproducible (as [discussed](https://github.com/loadimpact/k6/pull/1245#discussion_r346252753) yesterday, though that might have to wait for Go 1.14: https://github.com/golang/go/issues/35435
- build and deploy the github releases from a CI instead of manually via `build-release.sh`

na--:
Something else we should do for every commit, not just for tagged releases: actually build the rpm/deb/msi/etc. packages and test them (in the appropriate OS containers)... The v0.26.0 build of the linux packages broke: https://circleci.com/gh/loadimpact/k6/8360 :man_facepalming: If we were building these on every commit, or at least once daily, we would have noticed and fixed the issue much sooner.

imiric:
Since the `go-bin-{deb,rpm}` tools are giving us issues and they don't seem to be well maintained (last commit was over 2 years ago), we might want to consider switching to https://github.com/jordansissel/fpm .

I haven't used fpm personally, but have heard good things when used with Python projects. Presumably we should be able to use the directory input option for Go binaries, and would have more output options, including macOS .pkg files, so it's worth looking into.

imiric:
There's another issue with Homebrew packaging on macOS. The current [k6 formula](https://github.com/Homebrew/homebrew-core/blob/0efecb6cd027e6b8791ca2d3b03c21c78c0a7c30/Formula/k6.rb) actually builds the binary, which currently doesn't have any version information or any custom build flags we use for the GitHub releases. It's also a problem if we change our build system, since it currently uses `dep` and in v0.28.0 we switched to Go modules. This type of change hopefully won't happen again, but we shouldn't have to maintain two separate builds.

[...]

Issue #959, opened 2019-03-14
Use Azure Pipelines or GitHub Actions for CI


na--:
We're having occasional resource constraint issues with CircleCI, and AppVeyor (for windows tests and builds) has been consistently terrible for a while... so it seems like it's worth investigating https://azure.microsoft.com/en-us/services/devops/pipelines/

It could give us a single place for tests and builds on Linux, Windows, and macOS (!). They also offer "10 free parallel jobs with unlimited build minutes for all open source projects", which seems much better than what we're currently getting with CircleCI. They also seem to be better integrated with GitHub via [this](https://github.com/marketplace/azure-pipelines/).

na--:
Given the issues we've had with the hosted [golangci-lint](https://golangci.com/) service recently (https://github.com/loadimpact/k6/pull/1197, https://github.com/loadimpact/k6/pull/1204), we should probably transition to running it as a task in our new CI, whichever it ends up being.

na--:
Given that we'd be completely overhauling the CI, we might as well fix as many things that are currently wrong with the CI processes as possible. The latest ones are documented in https://github.com/loadimpact/k6/issues/1247, but a lot of the issues that are linked to this one (listed above) would probably qualify... Basically, I think that instead of faithfully and exactly translating what currently exists to the new CI service, we should pick the good parts of the current setup, but rewrite the bad ones in a better way...

na--:
Something like https://github.com/actions/stale seems necessary when we move to github actions

imiric:
I've been trying to replicate the current CircleCI setup in GH Actions for the last couple of days, and it's about [80% there](https://github.com/imiric/k6/actions), but there are some weird errors.

What works:
- Dependency check and test steps on Linux for current and previous Go versions. I needed to skip one test because of some strange DNS issues (`server misbehaving`?), but other than that they've been very stable, haven't seen any flakiness that plagued CircleCI builds. They run in comparable time, same 2-core VM. So overall, not bad. :+1: 
- Docker building and publishing works fine.

What kind of works:
[...]

Issue #1622, opened 2020-09-08
Extend CI workflow to push docker image to github container registry as well


simskij:
Given that DockerHub has made some changes in its pricing and added some limits to free users, we should consider pushing our docker images to GitHub Packages as well. If nothing else than for having a mirror available if someone hits the limit with DockerHub.

salem84:
Hi, I would like to change circleci pipeline in order to use Github Container Registry instead Github Packages, because as described [here](https://docs.github.com/en/free-pro-team@latest/packages/getting-started-with-github-container-registry/migrating-to-github-container-registry-for-docker-images) GitHub Container Registry supersedes the existing Packages Docker registry and is optimized to support some of the unique needs of containers.
What do you think about it?

na--:
@salem84, please wait a few days - we plant to merge https://github.com/loadimpact/k6/pull/1640 today, and use only GitHub actions for CI. Then pushing the docker images to both the docker hub and the github container registry will be easier. And, as things stand, we'll probably accept such a PR.

salem84:
Sure @na-- 
Yesterday, after I wrote this comment, I have seen other `ci` issues and I have found also #1640 
We will keep in touch and decide after merge!

imiric:
#1689 closes this. Thanks @salem84!

PR #1640, opened 2020-09-23
Replace CircleCI/AppVeyor with GitHub Actions

### LINKED ISSUE #1476: Don't build docker containers if we aren't gonna push them
I doubt we will find anything by constantly building them. But we currently do. 
This adds another 1m and 30s to each build for no reason IMO. 

This can probably wait for us to move to github actions, but maybe it will be fairly easy using the `filters` ?

imiric:
This replaces the CircleCI and AppVeyor setups with GitHub Actions.

It also creates a GH release if a version tag is pushed using the matching `.md` file in the `release notes` directory. You can see how it looks in [this test release](https://github.com/imiric/k6/releases/v0.29.0-test2).

Pending issues:

- Unable to use the [official golangci-lint action](https://github.com/golangci/golangci-lint-action), because there's no way to specify the `--new-from-rev` commit we've been using so far to ignore the hundreds of past linter issues. The action exposes an `only-new-issues` option, but even with it enabled it was picking up a lot of issues.
  We should eventually fix all of these of course, but some cases are not easily fixable, so this might take time. A drawback of not using the official action is that we won't have nice inline errors in the PR, and instead will have to check the `lint` job's output, but this is not a huge problem in practice.

- [Test flakiness on macOS](https://github.com/imiric/k6/runs/1154235795), so we can't enable those builds by default.

Partially addresses #959 and #1247, closes #1476 ?


.github/workflows/all.yml
```diff
@@ -0,0 +1,178 @@
...
+          export GOMAXPROCS=2
+          export PATH=$GOPATH/bin:$PATH
+          echo "mode: set" > coverage.txt
+          for pkg in $(go list ./... | grep -v vendor); do
+              list=$(go list -test -f  '{{ join .Deps  "\n"}}' $pkg | grep github.com/loadimpact/k6 | grep -v vendor || true)
+              if [ -n "$list" ]; then
+                  list=$(echo "$list" | cut -f1 -d ' ' | sort -u | paste -sd, -)
+              fi
+
+              go test -v -race -timeout 800s  --coverpkg="$list" -coverprofile=$(echo $pkg | tr / -).coverage $pkg
```
imiric, line 91: Reminder to disable `-v`.

.github/workflows/all.yml
```diff
@@ -0,0 +1,178 @@
...
+    tags:
+      - v*
+  pull_request:
+
+env:
+  APP_NAME: "k6"
+  DOCKER_IMAGE_ID: ${{ github.repository }}
+  # Force pure Go DNS resolver, see https://golang.org/pkg/net/#hdr-Name_Resolution
+  # Needed for some of our tests (TestErrorCodes).
+  GODEBUG: netdns=go+1
```
imiric, line 18: This can probably be removed if we decide to leave `TestErrorCodes` disabled, although we should figure out a way to make it pass...

MStoykov: Can you link me where it fails for all platforms ?

imiric: I couldn't find a failure on all platforms, so I might be misremembering that, but [here's the Linux one](https://github.com/imiric/k6/runs/928285449).

MStoykov (commented):
Why use fpm instead of what we already used to use? 
I would even argue this should stay in the github actions at least for now - no need to change that part, it has been working okay :D 

As a whole I don't think we want anything else but just moving to github actions without any new things, we can add after that.

Also can you reeanble the old CI so we can have both for ... a couple of weeks ;)

imiric:
> Why use fpm instead of what we already used to use?

A few reasons:
- The [go-bin-deb](https://github.com/mh-cbon/go-bin-deb) and [go-bin-rpm](https://github.com/mh-cbon/go-bin-rpm) projects haven't been updated in 3 years. This normally shouldn't be a problem as the package formats don't change often, but it is a problem when dealing with Go.
- We've already had issues [installing the tools](https://app.circleci.com/pipelines/github/loadimpact/k6/jobs/8360), and more recently [installing the Glide package manager they use](https://app.circleci.com/pipelines/github/loadimpact/k6/788/workflows/ce4d66a2-2f87-403e-bee7-5c000f817cc1/jobs/13192/parallel-runs/0), which is now also abandoned. They would probably install fine as Go modules, but I don't think we should bother.
- fpm is decently maintained, has a much larger user base and seems much more stable. The Ruby kind of sucks, but it installs fine on Ubuntu. It also supports more package formats which could be useful, and it has a nicer UI that doesn't require a JSON file, so we could remove some files from `packaging/`.

.github/workflows/all.yml
```diff
@@ -0,0 +1,177 @@
...
+        GOLANGCI_VERSION: v1.31
+        GO111MODULE: 'on'
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Lint
+        # Why is `go mod vendor` needed here!? If it's not run then
+        # golangci-lint fails because of inconsistent vendoring...
```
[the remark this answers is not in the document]
imiric: Yes, that is the question 😄

I'll run some tests and see if I can get rid of it.

.github/workflows/all.yml
```diff
@@ -0,0 +1,177 @@
...
+            docker run $DOCKER_IMAGE_ID version
+            docker run $DOCKER_IMAGE_ID --help
+            docker run $DOCKER_IMAGE_ID help
+            docker run $DOCKER_IMAGE_ID run --help
+            docker run $DOCKER_IMAGE_ID inspect --help
+            docker run $DOCKER_IMAGE_ID status --help
+            docker run $DOCKER_IMAGE_ID stats --help
+            docker run $DOCKER_IMAGE_ID scale --help
+            docker run $DOCKER_IMAGE_ID pause --help
+            docker run $DOCKER_IMAGE_ID resume --help
```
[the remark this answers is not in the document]
imiric: I guess to test that all commands work? 🤷‍♂️  I don't mind it.

imiric:
I added Windows packaging in the latest push and cleaned things up a bit, so this is mostly done now. I still need to fix that golangci-lint issue, try to make `http.TestErrorCodes` pass, and possibly add a job that runs `brew bump-formula-pr` now that we have Mac support. Since we haven't run tests on macOS even before this, updating the Homebrew package should be safe to automate, but ideally we should look into the flakiness of macOS tests on GHA and enable those too.

Uploading to Bintray is currently commented out as I'm not sure how to test it. Let's discuss that over Slack.

js/modules/k6/http/request_test.go
```diff
@@ -1641,6 +1641,8 @@ func checkErrorCode(t testing.TB, tags *stats.SampleTags, code int, msg string)
 }
 
 func TestErrorCodes(t *testing.T) {
+	// lookup dafsgdhfjg on 127.0.0.53:53: server misbehaving
```
MStoykov, line 1644: can you try to change the domain to `s.com` or something else that will not be existent but looks more like a domain as this seems like github action DNS problem

imiric: Thanks, that worked! 🎉 

I enabled the test in c9b1b9ba.

.github/workflows/all.yml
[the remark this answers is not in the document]
imiric: I added a workaround for this in 4b266969.

.github/workflows/all.yml
imiric: Resolved by c9b1b9ba.

imiric:
I added a job to run `brew` in 1d9cf2eb. It's untested of course, but I confirmed that `brew` is installed and that the passed arguments are correct, so it should work™️.

Pushing to codecov also [seems to work](https://codecov.io/gh/loadimpact/k6/commit/3ecfbb15ca33bd881c319195b3a9139d8eafc047), but I'm not sure what makes it make a PR comment with the coverage report.

The Bintray secrets are now in the repo and the commands should work, but as discussed over Slack, it's difficult to test the jobs as we'd need to create some test package repositories, and I don't want to mess a lot with Bintray, given it's a black box and their support is not great if anything goes wrong. So we could test this with an actual release?

I think the only thing left is the suggestion to keep CircleCI and AppVeyor enabled to run together with GH Actions for a while. I'm not sure if this is a good idea, as stuff like the coverage would be uploaded twice, though I guess we could disable it in CircleCI. @MStoykov any thoughts if this is required?

.github/workflows/all.yml
```diff
@@ -0,0 +1,299 @@
...
+
+  test:
+    strategy:
+      matrix:
+        go-version: [1.14.x, 1.15.x]
+        platform: [ubuntu-latest, windows-latest]
+    runs-on: ${{ matrix.platform }}
+    defaults:
+      run:
+        shell: bash
```
MStoykov, line 67: This can be moved to a global `defaults`

.github/workflows/all.yml
```diff
@@ -0,0 +1,299 @@
...
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: Install Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.14.x
```
MStoykov, line 28: I am wondering if we can define the two versions ... in a variable thing above and just use the variable here so we have 
`current_version`
`previous_version`
and maybe `release_version`

.github/workflows/all.yml
```diff
@@ -0,0 +1,299 @@
...
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Lint
+        run: |
+          pushd .
+          cd $(mktemp -d)
+          go get github.com/golangci/golangci-lint/cmd/golangci-lint@$GOLANGCI_VERSION
+          export PATH="$HOME/go/bin:$PATH"
+          popd
```
MStoykov, line 54: nit: this could've been done with `cd -` ... I think that is fairly standard?

.github/workflows/all.yml
```diff
@@ -0,0 +1,299 @@
...
+    outputs:
+        version: ${{ steps.get_version.outputs.version }}
+    steps:
+      - name: Get version
+        id: get_version
+        run: |
+          VERSION="${GITHUB_REF##*/}"
+          echo "VERSION=${VERSION}"
+          echo "::set-output name=version::${VERSION}"
+
```
MStoykov, line 116: Can you explain to me what this does ?

.github/workflows/all.yml
```diff
@@ -0,0 +1,299 @@
...
+          echo "${{ secrets.DOCKER_PASS }}" | docker login -u "${{ secrets.DOCKER_USER }}" --password-stdin
+          VERSION="${VERSION#v}"
+          echo "VERSION=$VERSION"
+          docker tag "$DOCKER_IMAGE_ID" "$DOCKER_IMAGE_ID:$VERSION"
+          docker push "$DOCKER_IMAGE_ID:$VERSION"
+          # We also want to tag the latest stable version as latest
+          if [[ "$VERSION" != "master" ]] && [[ ! "$VERSION" =~ (RC|rc) ]]; then
+            docker tag "$DOCKER_IMAGE_ID" "$DOCKER_IMAGE_ID:latest"
+            docker push "$DOCKER_IMAGE_ID:latest"
+          fi
```
MStoykov, line 194: we actually want to have `master` tag and `latest` to only be a tag release see https://github.com/loadimpact/k6/pull/1640/files#diff-1d37e48f9ceff6d8030570cd36286a61L124-L131

.github/workflows/all.yml
```diff
@@ -0,0 +1,299 @@
...
+        shell: powershell
+    needs: [deps, lint, test, configure, build]
+    if: startsWith(github.ref, 'refs/tags/v')
+    env:
+      VERSION: ${{ needs.configure.outputs.version }}
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: Install pandoc
+        uses: crazy-max/ghaction-chocolatey@v1
```
MStoykov, line 245: I would prefer if this is also with the commit ... given the security problems otherwise :(. (I also would prefer if that wasn't required but :man_shrugging: )

build-release.sh
```diff
@@ -1,87 +1,102 @@
-#!/usr/bin/env bash
```
MStoykov, line 1: Not certain if that wasn't because it broke in some environments :thinking: ... what is the upside of not using it ?

js/common/bridge_test.go
```diff
@@ -45,12 +45,12 @@ type bridgeTestMethodsType struct{}
 
 func (bridgeTestMethodsType) ExportedFn() {}
 
-//lint:ignore U1000 needed for the actual test to check that it won't be seen
+// nolint: unused
```
MStoykov, line 48: Apperantly there should be no spaces by [golang convention](https://github.com/golangci/golangci-lint/blob/7db400b2d2cf02c4712ad5f74b549c6f0f83ad2c/README.md#nolint)(last line) - this still works but maybe better if we keep to the convention, which I think we do in most other places

build-release.sh
```diff
@@ -1,87 +1,102 @@
...
+    local ALIAS="$1" FMT="$2"
+    local NAME="k6-${VERSION}-${ALIAS}"
+    echo "- Creating ${NAME}.${FMT} package..."
+    case $FMT in
+    deb|rpm)
+        # The go-bin-* tools expect the binary in /tmp/
+        [ ! -r /tmp/k6 ] && cp "dist/${NAME}/k6" /tmp/k6
+        "go-bin-${FMT}" generate --file "packaging/${FMT}.json" -a amd64 \
+            --version "${VERSION#v}" -o "dist/k6-${VERSION}-amd64.${FMT}"
+        ;;
```
MStoykov, line 47: Not certain about that being here (especially without any checks if this will pass). Previously we only needed go to build the binaries for all OSes ... now this will break half way through and also we will upload the deb and rpm files in the release? 

I am not saying that the deb and rpm shouldn't be in the release ... but it seems not all that useful, while being able to build the binaries for everything only running one command was great ;) 

Maybe ... Have a second shell script for building the rpm/debs? And it can error out if `dist/${NAME}/k6` doesn't exist and tell you to run the other one ? or just build it on its own. I am not certain

.github/workflows/all.yml
```diff
@@ -0,0 +1,299 @@
...
+          sudo apt-get install -y fakeroot rpm
+      - name: Build
+        run: |
+          go version
+          ./build-release.sh
+      - name: Upload artifacts
+        uses: actions/upload-artifact@v2
+        with:
+          name: binaries
+          path: dist/
```
MStoykov, line 155: This currently means that if you want one of the files you need to download the whole archive with all binaries to get one ... :( 

My tries at making this better ... let me to [this being the only way](https://github.com/loadimpact/k6cloudlogs/blob/master/.github/workflows/all.yml#L119-L148) ... :( Luckily you only need to write all that code once :D

MStoykov (commented):
I mostly want for them to run in parallel in order to see if something breaks, although arguably with tests this won't be very interesting, so we can probably skip it.

Are all the lint fixes for things that weren't caught in circleci or ?
Either way, I would prefer if we can move most of those ... not CI changes to a different PR and merge that first and rebase this on top ;). But this can be done later on as well or skipped entirely given the kind of changes

.github/workflows/all.yml
imiric: Sure, I guess it could be overriden for Windows jobs.

.github/workflows/all.yml
imiric: Sure, but I'll actually rewrite the installation as a separate step with a custom `working-directory` to avoid this.

.github/workflows/all.yml
imiric: It determines the version used globally, which can then be referenced by all jobs so that they can avoid having to figure out which version they're working with. It also defines a central point for similar configuration we might need globally.

build-release.sh
imiric: The standard Bash location on most Linux systems is `/bin/bash`. It's safer to not rely on `env` if you don't need it, since it avoids the uncertainty of which binary will be used. Otherwise it would run with any `bash` executable in `PATH`, which could not only be incompatible versions but also malicious interpreters. This is unlikely of course, but since this script should only be run in a controlled CI environment, we don't need the cross-platform benefit of using `env`.

js/common/bridge_test.go
imiric: That's good to know, but I've seen a few PRs of ours that actually introduce the space 😆 But OK, will remove it.

build-release.sh
imiric: There's already a second shell script for building the rpm/deb: [`gen-packages.sh`](https://github.com/loadimpact/k6/blob/f85389f4efc04cfe0e8c5840e28a7dfc8c595b8b/packaging/gen-packages.sh). The reason I moved it here is because this script already packages the binaries in .zip and .tar.gz, so it makes sense to keep this contained. I'd also like to get rid of the `packaging/` directory entirely, though now that we're back to using the Go tools I guess that won't be happening, and there are also some Windows package assets there that we can't remove.

> now this will break half way through

Why would it break half way through? We already rely on this directory structure for the other packages and there's no error checking there... If anything, I missed using `$OUT_DIR` instead of `dist` in the `cp` command here, so I'll change that.

As for uploading the deb/rpm to the GitHub release, yeah, that might not be needed. I'd argue it would still be useful for anyone who prefers not using the Bintray repositories, and there's no harm in giving more options. But if you feel strongly about it I could filter them out when creating the GH release.

.github/workflows/all.yml
imiric: Yes, this is unfortunate, but I don't mind getting the whole archive with all binaries. It takes a few seconds anyway, and I much prefer that over cluttering the workflow with each of the 8(!) artifacts that we need (we still need to upload the rpm/deb here even if they're not in the GH release).

.github/workflows/all.yml
imiric: Doesn't this accomplish the same thing? `$VERSION` could be `master`, which will tag and push it by default, and if it's not (i.e. it's a version tag that's not an `RC`), it will also tag and push it as `latest`. I haven't tested and confirmed this yet, but it seems like it will work as you expect it.

I'll let you know once [this WIP PR is merged in my fork](https://github.com/imiric/k6/pull/2), or we can wait until this PR is merged instead.

.github/workflows/all.yml
imiric: Yeah, that would be good, but it doesn't seem to work with the `env` context :( See https://github.com/imiric/k6/actions/runs/280602334 . Apparently this only works for each step's `env`, and not with `matrix` or anything else, so it's quite limited. See [this discussion](https://github.community/t/how-to-use-env-context/16975).

There doesn't seem to be something equivalent to [variables as in Azure Pipelines](https://docs.microsoft.com/en-us/azure/devops/pipelines/process/variables?view=azure-devops&tabs=yaml%2Cbatch), so I think we're stuck with repeating the version everywhere. :-/

js/common/bridge_test.go
imiric: Actually, I reverted the linter fixes entirely in 77088406. It seems these changes were leftover from when I was trying to get the official golangci-lint GH Action to work, and are not needed anymore now that we're installing it manually and passing `--new-from-rev` as we do on CircleCI.

.github/workflows/all.yml
imiric: Sure, fixed in fdf5aebb.

But I don't think we should be concerned about malicious tag changes, that rarely happens. As long as we use a more specific version (so `v1.3.0` instead of `v1`) we shouldn't have any build surprises.

And the drawback of using SHAs is their unfriendliness to humans 😆, as we can't easily tell what a version bump means, unlike with tags and semver.

imiric:
I removed the lint fixes in 77088406. They're not needed so we can leave them for later.

Keeping CircleCI/AppVeyor enabled side-by-side with GHA for a while would be safer, but it would also confuse external contributors and increase the chances for the build to fail, given our flaky test situation... So I'm not sure we need it. Let's wait for Ned's opinion next week, and for him to review this as well.

.github/workflows/all.yml
```diff
@@ -0,0 +1,284 @@
...
+          done
+          grep -h -v "^mode:" *.coverage >> coverage.txt
+          rm -f *.coverage
+          bash <(curl --fail -s https://codecov.io/bash)
+          go tool cover -html=coverage.txt -o coverage.html
+      - name: Upload coverage report
+        uses: actions/upload-artifact@v2
+        with:
+          name: test-coverage-report
+          path: coverage.html
```
imiric, line 96: This might be a problem... Since this job is run in a matrix of Go versions and platforms, `coverage.html` will be generated and uploaded by 4 different jobs. According to [the upload-artifact documentation](https://github.com/actions/upload-artifact#uploading-to-the-same-artifact), the best case scenario is that we don't know from which environment the coverage was generated in, which given that we skip some tests on Windows might be different than if the last upload was from Linux. The worst case could be corruption, if I guess more than one job uploads at once...

And there's also the issue of uploading the coverage to Codecov 4 times, which should only be done once.

Plus, we probably don't need to run with coverage in all environments, right? This would shorten the time for 3/4 jobs, which is helpful to get quicker feedback, and also use less GHA credits, if they start charging us :)

So I'll split this into a job that runs tests with coverage in a single environment (say, Linux with go1.15) and run tests without coverage on all others. Sounds good?

imiric: Hhmm for some reason running tests without coverage is always flaky on Windows, as my last 3 attempts failed. 😞 See https://github.com/imiric/k6/runs/1192519114

It was probably relying on the additional delays introduced by the coverage tracing...

@MStoykov Any ideas to quickly address this? Could we maybe run without `-race` or with `-p=1` on Windows if that would lower the chances of this happening?

MStoykov: we were previously running without `-race` so I guess we can do that as well now :)

MStoykov: codecov has support for .... uh tagging coverage and merging all the tags. But their documentation is currently down ?!? and I am not certain if it was called tags

imiric: Yeah, that could be an option, as it would be good to have coverage from all platforms. I'll look into it, thanks.

And tests are flaky on Windows even without `-race` :( https://github.com/imiric/k6/runs/1192739335

imiric: Running with `-p 1` and `GOMAXPROCS=1` seems to work though: https://github.com/imiric/k6/runs/1192809895

imiric: Here's the build that uploads coverage using [Codecov's flags feature](https://docs.codecov.io/docs/flags): https://github.com/imiric/k6/runs/1193077004

And how it looks in the UI: https://codecov.io/gh/imiric/k6/commit/a384ab1ad3bf283efe86bc1b821f3590fc5bbde0/build

Right now it's running go1.14 jobs without coverage and go.15 jobs with, since I don't think we need coverage for more than one Go version. On Windows I had to run with `GOMAXPROCS=1` and without `-race` in both cases to avoid flakiness.

I'll push the changes to this PR so we can see it here.

MStoykov: Looking [at a file it seems awesome](https://codecov.io/gh/imiric/k6/src/a384ab1ad3bf283efe86bc1b821f3590fc5bbde0/ui/summary.go) ... except that it turns out `windows` count empty lines and on `linux` this apparently doesn't happen

imiric: Yeah, looks like it 🤦  That's a rabbit hole I wouldn't want to get into though... So I'm inclined to leave it as is.

MStoykov: OOH ... this is totally not a problem, I couldn't find a golang issue about this as well and decided to leave it :)

imiric: I'm not sure if this is an issue with Go or with the way we're running the tests on Windows (under MSYS2 right now). A good test would be to run `go test` on PowerShell, but that would involve rewriting the `go list` logic in PS, so... no 😆

imiric:
I was trying to enable tests on macOS with limited concurrency as on Windows, but all runs have [failed](https://github.com/imiric/k6/runs/1219578104) so far, so it doesn't seem possible right now.

.github/workflows/all.yml
```diff
@@ -0,0 +1,319 @@
...
+      - name: Install Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: ${{ matrix.go-version }}
+      - name: Run tests
+        run: |
+          set -x
+          go version
+          export GOMAXPROCS=2
+          args=("-race")
```
na--, line 77: Did you investigate whether `-p 2` (or any other number) is faster? `go test -p` actually does different things than `GOMAXPROCS`.

.github/workflows/all.yml
```diff
@@ -0,0 +1,319 @@
...
+        id: get_version
+        run: |
+          VERSION="${GITHUB_REF##*/}"
+          echo "VERSION=${VERSION}"
+          echo "::set-output name=version::${VERSION}"
+
+  build:
+    runs-on: ubuntu-latest
+    needs: [deps, lint, test]
+    if: startsWith(github.ref, 'refs/tags/v')
```
na--, line 149: I think this is sub-optimal - the binaries and packages should be built on every k6 version, but they should just be published on `v*` tags.

.github/workflows/all.yml
```diff
@@ -0,0 +1,319 @@
...
+      - name: Create release
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          set -x
+          assets=()
+          for asset in ./dist/*; do
+            assets+=("-a" "$asset")
+          done
+          hub release create "${assets[@]}" -m "$VERSION" -m "$(cat ./release\ notes/${VERSION}.md)" "$VERSION"
```
na--, line 245: :heart_eyes:

.github/workflows/all.yml
na--: hmm maybe `-p 2` could also fix the MacOS flakiness? 
```
	-p n
		the number of programs, such as build commands or
		test binaries, that can be run in parallel.
		The default is the number of CPUs available.
```

imiric: Yeah, I was confusing `go test -parallel` (which defaults to `GOMAXPROCS`), with `go test -p`, so I thought it was redundant... Damn these non-POSIX flags!

OK, I'll try running macOS and Windows with `GOMAXPROCS=1`, `-p 1` and no `-race`.

.github/workflows/all.yml
imiric: Hhmm sure, I'll change it.

.github/workflows/all.yml
imiric: No, even with `-p 1` tests on macOS are very time sensitive and [fail consistently](https://github.com/imiric/k6/runs/1225385747). I'll still leave it for Windows as it should make them more reliable.

MStoykov: it seems this is a [known issue](https://github.com/actions/virtual-environments/issues/1336) that isn't getting better :(

na--: :disappointed: the sad fact is that a lot of our tests are too time sensitive, but we can't rewrite them quickly... I still think _some_ MacOS tests are better than none, so maybe we should check if we're running on a mac and `Skip()` some tests, like we do for Windows: https://github.com/loadimpact/k6/search?q=goos+windows

Then we can create an issue like this one for Mac: https://github.com/loadimpact/k6/issues/902

btw I know some of the reasons we skip some tests on Windows were caused specifically by AppVeyor, not Windows itself. So we might unskip some of the currently skipped tests after we merge this! :crossed_fingers:

MStoykov: I propose that we do any kind of CI improvements after we merge this PR, not in this PR ;)

na--: hmm fair enough, enabling partial macos tests and more windows tests is better done afterwards :+1:

.github/workflows/all.yml
imiric: So I like the idea of building binaries for all PRs, as that would make testing them much easier.

But there's currently a couple of minor problems with this:
- The `$VERSION` for these builds is the short Git SHA, which [fails the semantic version check for the RPM package](https://github.com/imiric/k6/runs/1225740444). I can change this to be similar to the `git describe` output, but RPM versions are also [incompatible with semver](https://github.com/semver/semver/issues/145), so we'd have to use tildes instead of dashes as suggested there, e.g. `<latest Git tag>~<commit SHA>`.
- The Windows .msi and Chocolatey packages are built in `publish-windows`, which should probably be done in a separate `build` job. But even then I know that the `.msi` package also has issues with semver, and we'd need a different workaround. You probably didn't mean to imply building Windows packages as well on every PR and on `master`, and since we're building the binary anyway, they're probably not needed.

So maybe to simplify this, what do you think about removing all packaging from `build-release.sh` (essentially reverting most of the changes I've done here...), and only building binaries and uploading them as .tar.gz and .zip on each PR/`master`? Leaving packaging exclusively for when building a `v*` tag, and probably keep using the same `gen-packages.sh` script as we do now.

.github/workflows/all.yml
```diff
@@ -0,0 +1,319 @@
+name: Deps, Lint, Test, Build, Publish
```
na--, line 1: I think we should change this to just `CI`. This looks a bit ghastly :sweat_smile: 
Screenshot_2020-10-08 Replace CircleCI AppVeyor with GitHub Actions by imiric · Pull Request #1640 · loadimpact k6
.

.github/workflows/all.yml
na--: :disappointed: Let's leave this for now then... In the interests of the perfect not being an enemy of the good, we can handle it as a separate PR after this is merged. I'd be much easier to iterate the CI once we no longer have CircleCI and AppVeyor anyway.

build-release.sh
```diff
@@ -1,87 +1,102 @@
...
+    elif command -v shasum > /dev/null; then
+        CHECKSUM_CMD=("shasum" "-a" "256")
+    else
+        echo "ERROR: unable to find a command to compute sha-256 hash"
+        exit 1
+    fi
+
+    echo "--- Generating checksum file..."
+    rm -f "${OUT_DIR}/$CHECKSUM_FILE"
+    (cd "$OUT_DIR" && find * -maxdepth 0 -type f | xargs "${CHECKSUM_CMD[@]}" > "$CHECKSUM_FILE")
```
na--, line 74: Running shellcheck on this file gives me 
```
    (cd "$OUT_DIR" && find * -maxdepth 0 -type f | xargs "${CHECKSUM_CMD[@]}" > "$CHECKSUM_FILE")
                      ^------------------------^ SC2038: Use -print0/-0 or -exec + to allow for non-alphanumeric filenames.
                           ^-- SC2035: Use ./*glob* or -- *glob* so names with dashes won't become options.

For more information:
  https://www.shellcheck.net/wiki/SC2038 -- Use -print0/-0 or -exec + to allo...
  https://www.shellcheck.net/wiki/SC2035 -- Use ./*glob* or -- *glob* so name...
```

.github/workflows/all.yml
```diff
@@ -0,0 +1,319 @@
...
+        uses: actions/checkout@v2
+      - name: Install Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.14.x
+      - name: Install package builders
+        env:
+          GO111MODULE: 'off'
+        run: |
+          gopath="$(go env GOPATH)"
```
na--, line 161:
```suggestion
          export GOPATH="$(go env GOPATH)"
```

.github/workflows/all.yml
```diff
@@ -0,0 +1,319 @@
...
+          go get -d github.com/mh-cbon/go-bin-deb \
+              && (cd "$gopath/src/github.com/mh-cbon/go-bin-deb" \
+              && glide install \
+              && go install)
+          go get -d github.com/mh-cbon/go-bin-rpm \
+              && (cd "$gopath/src/github.com/mh-cbon/go-bin-rpm" \
+              && glide install \
+              && go install)
+          sudo apt-get update -y
+          sudo apt-get install -y fakeroot rpm
```
na--, line 172:
```suggestion
          export GOPATH="$(go env GOPATH)"
          go get github.com/Masterminds/glide
          go get -d github.com/mh-cbon/go-bin-deb \
              && (cd "$GOPATH/src/github.com/mh-cbon/go-bin-deb" \
              && glide install \
              && go install)
          go get -d github.com/mh-cbon/go-bin-rpm \
              && (cd "$GOPATH/src/github.com/mh-cbon/go-bin-rpm" \
              && glide install \
              && go install)
          sudo apt-get update -y
          sudo apt-get install -y fakeroot rpm
```

.github/workflows/all.yml
imiric: It wouldn't be much work to revert some of these changes, upload binaries only on all commits, and keep using `gen-packages.sh` for `v*` tag builds only, though. But OK, it can wait for later.

MStoykov: why do want to *upload* binaries on *all* commits?

.github/workflows/all.yml
imiric: 😄 Sure. What's really annoying is that [you can't (easily) have dependencies between jobs in separate workflows](https://github.community/t/how-do-i-specify-job-dependency-running-in-another-workflow/16482/11)... Ideally, there should be a `publish.yml` workflow that only runs if, say, the `test.yml` and `build.yml` workflows passed. That would've cleaned up this long name and avoided having a huge single `all.yml` workflow.

.github/workflows/all.yml
na--: I'd like to build all the packages, so we know that is not broken when it comes to release.

I can also see _some_ value in saving them and uploading them as artifacts, but that would probably quickly get out of hand and we'd run against some limit if we don't clean them. So, I am not certain if we should do this, probably not. It's not difficult to build k6 yourself from a branch you want to test.

imiric: @MStoykov Because it would make them available for quick testing, as we've recently had to do with the DNS PRs. If we're building them anyway, I don't see a problem with uploading them as build artifacts, unless GH starts charging us or something.

na--: @imiric the storage limits seems super low, so I doubt we'll save all artifacts: https://docs.github.com/en/free-pro-team@latest/github/setting-up-and-managing-billing-and-payments-on-github/about-billing-for-github-actions

.github/workflows/all.yml
imiric: Hhmm I explicitly wanted to avoid messing with the real `GOPATH` as it doesn't need to be exported here and we only need it for changing into the right directories. Why would you want to change it?

build-release.sh
imiric: Thanks, I haven't adopted `shellcheck` yet, but it seems useful. Fixed in bd2b64da.

.github/workflows/all.yml
imiric: Resolved in 8a6c78da.

.github/workflows/all.yml
na--: Well, with `GO111MODULE: 'off'` you explicitly enter in "GOPATH mode", so it makes sense to embrace it :smile:

.github/workflows/all.yml
imiric: Ah, yeah, we'd run out of free storage quickly in that case. OK, then let's leave building/publishing for version tags only for now.

.github/workflows/all.yml
MStoykov: Actually not because this is not called unless there is a tag starting with `v` as it depends on `configure` so actually we will never push `master` tagged docker image ... I think

MStoykov (dismissed):
LGTM, Apart from 2 things that can be fixed in PRs after this one:
1. we won't be pushing `master` docker images ...
2. I do really think that the ergonomics of not having to download a zip with all the executables to just get yours outvalues the fact that someone will need to once copy paste the same code 8 times and change a few symbols in order for the different files to be uploaded separately

na--:
> I do really think that the ergonomics of not having to download a zip with all the executables to just get yours outvalues the fact that someone will need to once copy paste the same code 8 times and change a few symbols in order for the different files to be uploaded separately

@MStoykov, I think you are mistaken and artifacts would be individually uploaded in the GitHub release: https://github.com/loadimpact/k6/blob/aa97043e1fb6ea1d76ed359827623b9f1962973b/.github/workflows/all.yml#L243-L247

They are just zipped to schlep the binaries between the `build` and `publish-github` steps. That said, we need to verify that we are not going to run out of space pretty quickly with the sizable binary artifacts this is going to generate. I see they recently added an option to specify the retention: `https://github.com/actions/upload-artifact/commit/59018c2f85dd0e101b75544aa87f13bb0c94e0b7`

.github/workflows/all.yml
```diff
@@ -0,0 +1,321 @@
...
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          set -x
+          assets=()
+          for asset in ./dist/*; do
+            assets+=("-a" "$asset")
+          done
+          hub release create "${assets[@]}" -m "$VERSION" -m "$(cat ./release\ notes/${VERSION}.md)" "$VERSION"
+      # - name: Upload packages to Bintray
```
na--, line 248: I assume you had these commented out while you were experimenting in your own repo? We can uncomment them now, right?

imiric: These commands actually haven't been tested yet, as we don't have a test repository in Bintray and I don't want to accidentally publish something... I added the secrets in the repo and am fairly confident they'll work, but I'm apprehensive about testing them during an actual release. Any ideas how we could test them? Maybe setup a mock Bintray server and upload there?

na--: hmm... bintray have an open-source tier: https://bintray.com/signup/oss

can't we make another free account like that and add its credentials to your k6 fork, so you can fully test everything, including the artifact upload? after all, your k6 fork is still open-source, so it should count, we'd just be using the account for testing purposes (which should be publicly mentioned in the description, so some k6 users don't stumble on it by mistake from Google or something)

imiric: Hhmm OK, let me see if I can set that up.

MStoykov: I am still fo the opinion that we can test on release ... if it works - great. If it doesn't you will need to the more or less the same thing as you do now to fix it but without half the changing of stuff around to push to another repo and changing credentials and then back.

na--: Yeah, if it's complicated, then you can just uncomment the bintray lines and we'll cross our fingers during the next release :crossed_fingers: :sweat_smile: It'd be ideal if we had a second bintray account and can test everything in a fork though, since that would allow us to more easily iterate on and test the CI scripts for any future modifications. Half of the tediousness of CI scripts comes from the fact that it's difficult to test them...

MStoykov: maybe merge this with uncommented lines and if there is time left before the release (even in the last few days) this can be worked on ?

imiric: It wasn't much work to create a separate Bintray account and identical repos, but [I'm getting a 400 on upload](https://github.com/imiric/k6/runs/1241346992), which could be anything... I'll see if I can quickly troubleshoot this without exposing the secrets, but I'd rather ensure this works now than have to do it during a release.

na--: you can expose things like the API key for your bintray account and then just reset it from bintray, if that will help with the debugging
and yeah, for a 400 status, :+1: for debugging now and not under pressure after a release

imiric: OK, so the 400 was caused because the packages in Bintray needed to be manually created. After doing that, and fixing a `GPG_PASSPHRASE` secret name typo in my repo, the packages uploaded fine in [this build](https://github.com/imiric/k6/actions/runs/302022109), and you can see them [here](https://bintray.com/imiric2). I had to disable all tests as I was getting a flaky failure on previous builds... :disappointed: The flakiness is really a major problem that's wasting a lot of time in day-to-day work, and gives off a bad impression to external contributors.

Anyway, I'll clean things up and squash some commits, and push the final version to this PR in a few minutes.

There's also an `add-path` deprecation warning I just now noticed. Will look into it if it's a simple fix.

imiric:
I fixed the `add-path` deprecation warning, and added `retention-days: 7` to the `upload-artifact` steps.

Let's not forget to disable the webhooks in the repo settings so we can see passing builds, which could be done after this is merged as well.

.github/workflows/all.yml
imiric: Ah true, the `configure` dependency prevents it. It's trivial to fix if we allow `configure` to run on all commits and not just version tags, which seems safe to do. Let me know if you prefer to do this now, otherwise we can fix it later, maybe when we work on #1622.

MStoykov: I am fine with it being later as I don't think docker `master` is all that used, it is more important that `latest` isn't updated unless we actually release, which seems to be the case :) 
cc @na--

MStoykov (approved):

na-- (approved):
LGTM :crossed_fingers:

PR #1644, opened 2020-09-30
Increase http debug logging

repl-david-winiarski:


MStoykov:
Hi @repl-david-winiarski, thanks for this PR :) 
As a first thing that I notice - we do already use another uuid library `github.com/nu7hatch/gouuid` so it will be nice if we don't add a second one :)

repl-david-winiarski:
@MStoykov thanks for pointing that out. I've removed the additional uuid dependency and replaced it with the suggested one.

MStoykov (approved):

imiric (approved):
Thanks for this! 👍

na-- (approved):

PR #1647, opened 2020-10-02
Immediately  shutdown `k6 cloud`  on second sig-c signal

### LINKED ISSUE #1530: Always immediately stop `k6 cloud` on a second Ctrl+C
In `k6 cloud`, we should have something similar to what we're going to add to `k6 run` in v0.27.0: https://github.com/loadimpact/k6/blob/447f892cc2c4073290e25b220307851d14aaae95/cmd/run.go#L279-L294

theerapatcha:
Trap signal and handle second sig-c to shutdown `k6 cloud` immediately

Closes #1530

### DIFF
--- a/cmd/cloud.go
+++ b/cmd/cloud.go
@@ -258,7 +258,7 @@ This will execute the test on the k6 cloud service. Use "k6 login cloud" to auth
 				logger.WithError(err).Error("Stop cloud test error")
 			}
 			shouldExitLoop = true // Exit after the next GetTestProgress call
-		
+
 			sig = <-sigC
 			logger.WithField("sig", sig).Error("Aborting k6 in response to signal")
 			globalCancel()
--- a/cmd/cloud.go
+++ b/cmd/cloud.go
@@ -329,7 +329,6 @@ This will execute the test on the k6 cloud service. Use "k6 login cloud" to auth
 			case <-globalCtx.Done():
 				break runningLoop
 			}
-
 		}
 
 		if testProgress == nil {
--- a/cmd/cloud.go
+++ b/cmd/cloud.go
@@ -199,6 +199,9 @@ This will execute the test on the k6 cloud service. Use "k6 login cloud" to auth
 			name = filepath.Base(filename)
 		}
 
+		globalCtx, globalCancel := context.WithCancel(context.Background())
+		defer globalCancel()
+
 		// Start cloud test run
 		modifyAndPrintBar(progressBar, pb.WithConstProgress(0, "Validating script options"))
 		client := cloud.NewClient(logger, cloudConfig.Token.String, cloudConfig.Host.String, consts.Version)
@@ -226,7 +229,7 @@ This will execute the test on the k6 cloud service. Use "k6 login cloud" to auth
 			pb.WithConstProgress(0, "Initializing the cloud test"),
 		)
 
-		progressCtx, progressCancel := context.WithCancel(context.Background())
+		progressCtx, progressCancel := context.WithCancel(globalCtx)
 		progressBarWG := &sync.WaitGroup{}
 		progressBarWG.Add(1)
 		defer progressBarWG.Wait()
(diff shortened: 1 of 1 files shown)


na-- (commented):
Thanks for this! At a quick glance, it looks very good, but I'd like to test the new code as well before giving it the final :+1:

na-- (approved):
LGTM, tested it as well and everything seems fine, thanks again!

MStoykov (dismissed):

cmd/cloud.go
```diff
@@ -241,10 +244,26 @@ This will execute the test on the k6 cloud service. Use "k6 login cloud" to auth
 			return nil
 		}
 
+		shouldExitLoop := false
```
MStoykov, line 247: Maybe we can directly wait on `globalCtx.Done()` instead of checking this boolean ... which also seems like a race condition

na--: Ah, yeah, I missed this... :sweat_smile: Not really a scary data race, but a data race nonetheless :disappointed: :+1: for directly checking the `globalCtx.Done()` in the `select` below

na-- (changes requested):
Please fix the race condition reported in https://github.com/loadimpact/k6/pull/1647#discussion_r504604946

cmd/cloud.go
theerapatcha: @na-- I fixed the code per the comments above.  If anything more needs to change, please do tell.

cmd/cloud.go
```diff
@@ -303,24 +318,16 @@ This will execute the test on the k6 cloud service. Use "k6 login cloud" to auth
 				if progressErr == nil {
 					if (newTestProgress.RunStatus > lib.RunStatusRunning) ||
 						(exitOnRunning && newTestProgress.RunStatus == lib.RunStatusRunning) {
-						shouldExitLoop = true
+						globalCtx.Done()
```
MStoykov, line 321:
```suggestion
						globalCancel()
```

cmd/cloud.go
```diff
@@ -303,24 +318,16 @@ This will execute the test on the k6 cloud service. Use "k6 login cloud" to auth
 				if progressErr == nil {
 					if (newTestProgress.RunStatus > lib.RunStatusRunning) ||
 						(exitOnRunning && newTestProgress.RunStatus == lib.RunStatusRunning) {
-						shouldExitLoop = true
+						globalCancel()
```
na--, line 321: Hmm this change doesn't seem right, I think you should keep `shouldExitLoop` here. 

Imagine that you have a slow connection or the k6 cloud backend is slow and `client.GetTestProgress(refID)` takes more than 2 seconds (the `ticker` period). Then, on the next iteration of the `for {}` loop, the `select` has an equal chance to pick either branch, and you can end up back here again, rather than breaking out of the loop

na-- (changes requested):

cmd/cloud.go
MStoykov: I think this should just call `ticker.Stop()` ... There ~was~[is a bug](https://github.com/golang/go/issues/37196) where it might send one more tick (if it has alredy ticked basically) but I think we can ignore that or ... drain it with the `select/default` trick

na--: As you say @MStoykov, it can tick one more time if the `client.GetTestProgress(refID)` call took more than 2 seconds. I think cleanest solutions is both the previous logic there (`shouldExitLoop` and breaking the loop as soon as that `if` succeeds) **and** cancelling the context.

MStoykov: Yes @theerapatcha , that is more or less what @na-- wrote, although I agree that it might've not been so clear :(

so you just need to add a `break` on the next line here and we should be ready to merge this :)

theerapatcha: done :)

MStoykov (approved):

na-- (approved):
LGTM, though the old `shouldExitLoop` behavior would have been better than a direct `break` of the loop, since now we won't update the progress (`testProgress = newTestProgress`), but we can fix that later in another PR.

PR #1650, opened 2020-10-04
Add cookie to ws

### LINKED ISSUE #1226: Support passing CookieJar to WebSocket through Params
The ws library is not passing the Cookie jar to the Websocket connection.

The server we have is using the cookie in the request to pull out the necessary variables to establish the websocket

The underlying Go library seems to support it but it is ignored and dropped when establishing the WS.

It would be create to support it.

lcd1232:
Closes #1226.
Add header `Cookie` which allows sending cookies to WebSocket.

### DIFF
--- a/js/modules/k6/ws/ws.go
+++ b/js/modules/k6/ws/ws.go
@@ -28,6 +28,8 @@ import (
 	"io/ioutil"
 	"net"
 	"net/http"
+	"net/http/cookiejar"
+	netURL "net/url"
 	"strconv"
 	"strings"
 	"sync"
@@ -41,6 +43,8 @@ import (
 	"github.com/loadimpact/k6/lib/metrics"
 	"github.com/loadimpact/k6/stats"
 )
+
+const HeaderKeyCookie = "X-Cookie"
 
 // ErrWSInInitContext is returned when websockets are using in the init context
 var ErrWSInInitContext = common.NewInitContextError("using websockets in the init context is not supported")
@@ -104,6 +108,7 @@ func (*WS) Connect(ctx context.Context, url string, args ...goja.Value) (*WSHTTP
 
 	// Leave header to nil by default so we can pass it directly to the Dialer
 	var header http.Header
+	var jar http.CookieJar
 
 	tags := state.CloneTags()
 
@@ -123,6 +128,26 @@ func (*WS) Connect(ctx context.Context, url string, args ...goja.Value) (*WSHTTP
 					continue
 				}
 				for _, key := range headersObj.Keys() {
+					if key == HeaderKeyCookie {
+						req, err := http.NewRequest("", url, nil)
+						if err != nil {
+							return nil, err
+						}
+
+						req.Header.Set("Cookie", headersObj.Get(key).String())
+						cookies := req.Cookies()
+						jar, err = cookiejar.New(nil)
+						if err != nil {
+							return nil, err
+						}
+
+						u, err := netURL.Parse(url)
+						if err != nil {
+							return nil, err
+						}
+
+						jar.SetCookies(u, cookies)
+					}
 					header.Set(key, headersObj.Get(key).String())
 				}
 			case "tags":
@@ -160,6 +185,7 @@ func (*WS) Connect(ctx context.Context, url string, args ...goja.Value) (*WSHTTP
 		NetDialContext:  state.Dialer.DialContext,
 		Proxy:           http.ProxyFromEnvironment,
 		TLSClientConfig: tlsConfig,
+		Jar:             jar,
 	}
 
 	start := time.Now()
--- a/js/modules/k6/ws/ws.go
+++ b/js/modules/k6/ws/ws.go
@@ -44,7 +44,7 @@ import (
 	"github.com/loadimpact/k6/stats"
 )
 
-const HeaderKeyCookie = "X-Cookie"
+const HeaderKeyCookie = "Cookie"
 
 // ErrWSInInitContext is returned when websockets are using in the init context
 var ErrWSInInitContext = common.NewInitContextError("using websockets in the init context is not supported")
--- a/lib/netext/httpext/request.go
+++ b/lib/netext/httpext/request.go
@@ -326,7 +326,6 @@ func MakeRequest(ctx context.Context, preq *ParsedHTTPRequest) (*Response, error
 
 		resp.URL = res.Request.URL.String()
 		resp.Status = res.StatusCode
-		resp.StatusText = res.Status
 		resp.Proto = res.Proto
 
 		if res.TLS != nil {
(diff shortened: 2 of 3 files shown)


js/modules/k6/ws/ws.go
```diff
@@ -123,6 +129,26 @@ func (*WS) Connect(ctx context.Context, url string, args ...goja.Value) (*WSHTTP
 					continue
 				}
 				for _, key := range headersObj.Keys() {
+					if key == HeaderKeyCookie {
+						req, err := http.NewRequest("", url, nil) //nolint:noctx
```
imiric, line 133: Why do you need to create a new request here or mess with the `Cookie` header? Couldn't you create `http.Cookie` directly and pass it to `jar.SetCookies()`?

js/modules/k6/ws/ws.go
```diff
@@ -123,6 +129,26 @@ func (*WS) Connect(ctx context.Context, url string, args ...goja.Value) (*WSHTTP
 					continue
 				}
 				for _, key := range headersObj.Keys() {
+					if key == HeaderKeyCookie {
```
imiric, line 132: HTTP header names are case-insensitive, so using `strings.ToLower()` here would be preferable.

imiric (changes requested):
Thanks for your contribution! I left some minor comments, but we should also have a unit test or two for this change.

js/modules/k6/ws/ws.go
MStoykov: [strings.EqualFold](https://golang.org/pkg/strings/#EqualFold) is ... way faster :) 
```
package main

import (
        "strings"
        "testing"
)

func BenchmarkEq(b *testing.B) {
        for i := 0; i < b.N; i++ {
                _ = strings.EqualFold("Cookie", "cookie")
        }
}

func BenchmarkLower(b *testing.B) {
        for i := 0; i < b.N; i++ {
                _ = strings.ToLower("Cookie") == "cookie"
        }
}
```
```
BenchmarkEq-8           69264416                17.9 ns/op             0 B/op          0 allocs/op
BenchmarkLower-8        21890653                59.3 ns/op             8 B/op          1 allocs/op
```

```

File: js/modules/k6/ws/ws.go (hunk 1 above)
Comment (MStoykov) [reply js/modules/k6/ws/ws.go:133]: On a somewhat related note ... this is not how it works for HTTP requests. 

There is a [default jar](https://github.com/loadimpact/k6/blob/master/lib/state.go#L57) and you can also provide a CookieJar and there is some logic to merge between all of those and Cookie param, which probably should be reused here.

Comment (lcd1232) [reply js/modules/k6/ws/ws.go:133]: > Why do you need to create a new request here or mess with the Cookie header? Couldn't you create http.Cookie directly and pass it to jar.SetCookies()?

Because parsing HTTP cookies is a private function inside go standard library ([proof](https://github.com/golang/go/blob/master/src/net/http/cookie.go#L238)) and I decided not to copy this function.

File: js/modules/k6/ws/ws.go (hunk 2 above)
Comment (lcd1232) [reply js/modules/k6/ws/ws.go]: Fixed it

**Comment by lcd1232** [comment]
@imiric hello. I rewrite it to use cookie object

**Comment by lcd1232** [comment]
@imiric @MStoykov ping

File: js/modules/k6/ws/ws.go [hunk 3]
```diff
@@ -125,6 +139,43 @@ func (*WS) Connect(ctx context.Context, url string, args ...goja.Value) (*WSHTTP
...
+					case reflect.TypeOf(map[string]interface{}{}):
+						cookie := cookieV.ToObject(rt)
+						for _, attr := range cookie.Keys() {
+							switch strings.ToLower(attr) {
+							case "replace":
+								replace = cookie.Get(attr).ToBoolean()
+							case "value":
+								value = cookie.Get(attr).String()
+							}
+						}
```
imiric, line 168: I don't understand why this block is needed. You're parsing a cookie value as an object that can override the same cookie in the jar, but I'm not sure why this is necessary. Can you elaborate on your use case?

js/modules/k6/ws/ws.go
```diff
@@ -104,9 +107,20 @@ func (*WS) Connect(ctx context.Context, url string, args ...goja.Value) (*WSHTTP
...
 
 	tags := state.CloneTags()
 
+	u, err := netURL.Parse(url)
+	if err != nil {
+		return nil, err
+	}
+
+	cookiesMap := make(map[string]http.Cookie)
+	for _, cookie := range state.CookieJar.Cookies(u) {
```
imiric, line 120: Should we write the `cookiesMap` changes back to `state.CookieJar`? cc: @MStoykov 

And I think @MStoykov would like to reuse the cookie logic in the `httpext` package, e.g. [`SetRequestCookies`](https://github.com/loadimpact/k6/blob/4e637f6e57906584a5ae2f90da836802be4e03c9/lib/netext/httpext/request.go#L383), though we should probably agree on the details.

imiric:
Thanks for the changes @lcd1232. Would you mind rebasing this branch on the current `master`, since we switched CI to GitHub Actions?

js/modules/k6/ws/ws.go
MStoykov: Looking at the [websocket docs](https://godoc.org/github.com/gorilla/websocket) it looks like reusing it will be ... harder than I expected ...  read ... not possible unless we rewrite it which won't help at all either way.

Looking at the httpext k6 does write back *only* stuff that [were set in the responses](https://github.com/loadimpact/k6/blob/4e637f6e57906584a5ae2f90da836802be4e03c9/lib/netext/httpext/request.go#L283-L290)

js/modules/k6/ws/ws_test.go
```diff
@@ -316,6 +316,24 @@ func TestSession(t *testing.T) {
 	})
 	assertSessionMetricsEmitted(t, stats.GetBufferedSamples(samples), "", sr("WSBIN_URL/ws-echo"), 101, "")
 
+	t.Run("client_cookie", func(t *testing.T) {
+		_, err := common.RunString(rt, sr(`
+		var params = {
+			cookies: { "Session": "123" },
```
imiric, line 322: We should also have a test for when the value is an object and uses `replace`.

imiric (changes requested):
@lcd1232 Are you still available to wrap up this PR? We'd like to merge it for v0.33.0 as the code LGTM besides the test comment, but it would need rebasing since there have been a lot of changes in the past 7 months (though few related to WS, so there are no conflicts), and you'll need to change the test to use `rt.RunString()` instead of `common.RunString()` since we dropped our wrapper.

If we don't hear from you in the next few days or you're not available anymore, that's OK too, and we can continue and merge it ourselves. You would of course be credited in our release notes ;)

PR #1651, opened 2020-10-06
Prevent loop when babel can transpile but goja can't parse

MStoykov:


na-- (dismissed):
LGTM :man_facepalming: :sweat_smile:

js/compiler/compiler.go
```diff
@@ -81,7 +81,7 @@ func (c *Compiler) Compile(src, filename, pre, post string,
 			if err != nil {
 				return nil, code, err
 			}
-			return c.Compile(code, filename, pre, post, strict, compatMode)
+			return c.Compile(code, filename, pre, post, strict, lib.CompatibilityModeBase)
```
imiric, line 84: This change doesn't make sense to me... Why would we always compile with `base` if the given `compatMode` is `extended`? :-S 

I see there's the `U+2029` edge case, but shouldn't that be specifically handled?

MStoykov: The only use of `compatMode` in this function) is in the encompassing `if`  in order to determine if we should try to run the code through `babel` which should help with stuff like ES6+ syntax. 

The problem is that if babel manages to ~transpile~transform it, but k6 *still* can *not* parse it we will loop. While the concrete example that I got it from ([fuzzing](https://github.com/MStoykov/goja-regexp2-fuzzing)) might be a different bug, we still shouldn't loop. 

Some more background: 

This previously was done with a `bool` that was just always `true` when called from outside but from inside it was `false`. When this was [rewritten by you](https://github.com/loadimpact/k6/pull/1206)  ;),  you changed it to use the compatability mode instead, which I agree with, it just meant that it makes the "intend" of the parameter more obscure as previously it was called `tryBabel` and now it's called `compatMode` ... and well we don't exactly demote the `compatMode` we just have already tried it ... so maybe we should rename the parameter as well?

For the record, nobody noticed this during the PR review(I have commented mostly every other line in this function :facepalm: )  and it took me quite some time to realize what is happening even after I found it out so apparently in MOST cases babel returns pretty okay errors :man_shrugging: and this plus the fact that the way it's called looks pretty okay with the new name makes it kind of hard to see.

na-- (approved):

js/compiler/compiler.go
imiric: Hhmm I see. Then :+1: for renaming the parameter or adding some documentation, as right now it's not clear why we do that.

na-- (approved):

imiric (approved):

PR #1293, opened 2020-01-06
Add option to specify VU network interface

srguglielmo:
This adds a "nic" option (e.g. `k6 run --nic eth0`) which will cause the
Virtual User to use the IP address(es) assigned to the given network
interface. If the NIC has multiple IPs, they will be used by each VU at
random.

Credit for the original code goes to @ofauchon.

Closes GitHub issue loadimpact#476 and original PR loadimpact#757.

PR #1654, opened 2020-10-06
Use IP Range or CIDR for K6 source IP addresses

divfor:
An alternative PR for using IP range option #1293 , no Network Interface required.
Example of run script:
```bash
#!/bin/bash
s=${1:-"ax.js:1000:10m:rps_http"}
read jsfile vus duration casename <<<${s//:/ }
shift
echo -e "testfile[$jsfile] testcase[$casename] vus[$vus] duration[$duration]\noptions[$*]"

set_k6_local_route(){ #$1:net $2:ifname
  net="$1" ifname=${2-"lo"}
  ip r s t local |grep -q "^local $net" || \
  ip -6 r s t local |grep -q "^local $net" || \
  ip route add local "$net" dev $ifname
  ip r s t local |grep "^local $net" || ip -6 r s t local |grep "^local $net"
}

# k6 client ip -- use local type route instead
set_k6_local_route "10.50.0.0/16" "eth0"
set_k6_local_route "fd00:0:1::0/120" "eth0"

# Linux: echo "ulimit -n 1000000" >> ~/.bashrc
# MacOS: sudo launchctl limit maxfiles 65536 1000000
ulimit -n 1000000

# run k6 for ipv4 targets
TESTCASE=$casename VUS=$vus DURATION=$duration k6 run --ip="10.50.0.62-10.50.2.100" --verbose --no-thresholds $* $jsfile
# run k6 for ipv6 targets
sysctl -w net.ipv6.ip_nonlocal_bind=1
TESTCASE=$casename VUS=$vus DURATION=$duration k6 run --ip="fd00:0:188::/120" --verbose --no-thresholds $* $jsfile
sysctl -w net.ipv6.ip_nonlocal_bind=0
```

### DIFF
--- a/cmd/options.go
+++ b/cmd/options.go
@@ -91,6 +91,7 @@ func optionFlagSet() *pflag.FlagSet {
 	flags.StringSlice("tag", nil, "add a `tag` to be applied to all samples, as `[name]=[value]`")
 	flags.String("console-output", "", "redirects the console logging to the provided output file")
 	flags.Bool("discard-response-bodies", false, "Read but don't process or save HTTP response bodies")
+	flags.StringP("ip", "", "", "Client IP Range or CIDR, randomly selected by VU id")
 	return flags
 }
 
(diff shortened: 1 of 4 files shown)


na--:
Hey, @divfor, can we help in some way? If the issue is caused by a wrong commit author, maybe this will help - https://www.git-tower.com/learn/git/faq/change-author-name-email/

PR #1655, opened 2020-10-06
Split StatsD and Datadog config

### LINKED ISSUE #671: Env variable DURATION is being used implicitly
I've bumped into this accidentally:
If global environment variable DURATION is set, it is being used by k6 to parse duration value.
In my case I wanted to execute k6 script with following command:
`k6 run --vus 2 --duration 60s <script_name.js>`
But also I had a global environment variable DURATION=60, and as result my execution kept failing with
```
ERRO[0001] envconfig.Process: assigning K6_DURATION to Duration: converting '60' to type types.NullDuration. details: time: missing unit in duration 60
```
I didn't find any mentions about DURATION in docs:
https://docs.k6.io/docs/running-k6
https://docs.k6.io/docs/environment-variables
https://docs.k6.io/docs/execution-context-variables

imiric:
This addresses the suggestion in [this comment](https://github.com/loadimpact/k6/issues/671#issuecomment-546233883) in order to close #671.

I tested the integration with a local StatsD instance and with Datadog and it works OK.


MStoykov (approved):
LGTM!
I was thinking of just having the parsing of the config to be in the corresponding package and that to be then transcoded to the `common.Config` and mostly no other changes. But I think I figured out there is a problem with that approach back when I left this unfinished and a little more code is not a problem :)

na-- (approved):

PR #1658, opened 2020-10-07
IP Range and CIDR support

divfor:
An alternative PR for using IP range option #1293 , no Network Interface required.
Example of run script:
```bash
#!/bin/bash
s=${1:-"ax.js:1000:10m:rps_http"}
read jsfile vus duration casename <<<${s//:/ }
shift
echo -e "testfile[$jsfile] testcase[$casename] vus[$vus] duration[$duration]\noptions[$*]"

set_k6_local_route(){ #$1:net $2:ifname
  net="$1" ifname=${2-"lo"}
  ip r s t local |grep -q "^local $net" || \
  ip -6 r s t local |grep -q "^local $net" || \
  ip route add local "$net" dev $ifname
  ip r s t local |grep "^local $net" || ip -6 r s t local |grep "^local $net"
}

# k6 client ip -- use local type route instead
set_k6_local_route "10.50.0.0/16" "eth0"
set_k6_local_route "fd00:0:1::0/120" "eth0"

# Linux: echo "ulimit -n 1000000" >> ~/.bashrc
# MacOS: sudo launchctl limit maxfiles 65536 1000000
ulimit -n 1000000

# run k6 for ipv4 targets
TESTCASE=$casename VUS=$vus DURATION=$duration k6 run --ip="10.50.0.62-10.50.2.100" --verbose --no-thresholds $* $jsfile
# run k6 for ipv6 targets
sysctl -w net.ipv6.ip_nonlocal_bind=1
TESTCASE=$casename VUS=$vus DURATION=$duration k6 run --ip="fd00:0:188::/120" --verbose --no-thresholds $* $jsfile
sysctl -w net.ipv6.ip_nonlocal_bind=0
```

### DIFF
--- a/cmd/options.go
+++ b/cmd/options.go
@@ -91,6 +91,7 @@ func optionFlagSet() *pflag.FlagSet {
 	flags.StringSlice("tag", nil, "add a `tag` to be applied to all samples, as `[name]=[value]`")
 	flags.String("console-output", "", "redirects the console logging to the provided output file")
 	flags.Bool("discard-response-bodies", false, "Read but don't process or save HTTP response bodies")
+	flags.StringP("ip", "", "", "Client IP Range or CIDR from which each VU selects one by VU id")
 	return flags
 }
 
@@ -114,6 +115,7 @@ func getOptions(flags *pflag.FlagSet) (lib.Options, error) {
 		MinIterationDuration:  getNullDuration(flags, "min-iteration-duration"),
 		Throw:                 getNullBool(flags, "throw"),
 		DiscardResponseBodies: getNullBool(flags, "discard-response-bodies"),
+		ClientIpRange:         getNullString(flags, "ip"),
 		// Default values for options without CLI flags:
 		// TODO: find a saner and more dev-friendly and error-proof way to handle options
 		SetupTimeout:    types.NullDuration{Duration: types.Duration(60 * time.Second), Valid: false},
--- a/js/iprange.go
+++ b/js/iprange.go
@@ -14,7 +14,6 @@ type SelectMode uint
 const (
 	LoopIncSelectIP SelectMode = iota // increase one by one, no IP is leaking
 	RandomSelectIP                    // random
-	ModLenSelectIP                    // like LoopIncSelect, may leak IP(s) if id is not continous
 )
 
 type IPBlock struct {
@@ -107,15 +106,15 @@ func ipBlockFromCIDR(s string) *IPBlock {
 }
 
 // GetRandomIP return a random IP by seed from an IP block
-func (b IPBlock) GetRandomIP(seed int64) net.IP {
-	r := rand.New(rand.NewSource(seed))
+func (b IPBlock) GetRandomIP(seed uint64) net.IP {
+	r := rand.New(rand.NewSource(int64(seed)))
 	h := r.Uint64()
 	n := r.Uint64()
-	return b.GetModNIndexedIP(h, n)
+	return b.GetRoundRobinIP(h, n)
 }
 
-// GetModNIndexedIP return a IP by indexes from an IP block
-func (b IPBlock) GetModNIndexedIP(hostIndex, netIndex uint64) net.IP {
+// GetRoundRobinIP return a IP by indexes from an IP block
+func (b IPBlock) GetRoundRobinIP(hostIndex, netIndex uint64) net.IP {
 	if ip4 := b.ip.To4(); ip4 != nil {
 		i := b.hostStart + hostIndex%b.hostN
 		return net.IPv4(byte(i>>24), byte(i>>16), byte(i>>8), byte(i))
@@ -169,24 +168,22 @@ func GetPool(ranges string) IPPool {
 	return pool
 }
 
-func (pool IPPool) GetIP(id int64) net.IP {
-	last := len(pool)
-	if last < 1 {
+func (pool IPPool) GetIP(id uint64) net.IP {
+	if len(pool) < 1 {
 		return nil
 	}
-	b := pool[last-1]
-	j := uint64(id) % pool[last-1].weight
-	for i := 0; i < last-1; i++ {
-		if j < pool[i].weight {
-			b = pool[i]
-			break
+	idx := id % pool[len(pool)-1].weight
+	for _, b := range pool {
+		if idx < b.weight {
+			switch b.mode {
+			case RandomSelectIP:
+				return b.GetRandomIP(id)
+			case LoopIncSelectIP:
+				return b.GetRoundRobinIP(id, id)
+			default:
+				return nil
+			}
 		}
-	}
-	switch b.mode {
-	case RandomSelectIP:
-		return b.GetRandomIP(id)
-	case LoopIncSelectIP:
-		return b.GetModNIndexedIP(uint64(id), uint64(id))
 	}
 	return nil
 }
--- a/js/iprange.go
+++ b/js/iprange.go
@@ -1,29 +1,31 @@
 package js
 
 import (
-	"fmt"
-	"strings"
-	"strconv"
-	"net"
-	"time"
-	"math/rand"
 	"encoding/binary"
+	"fmt"
+	"math/rand"
+	"net"
+	"strconv"
+	"strings"
+	"time"
 )
 
 type SelectMode uint
+
 const (
-        LoopIncSelectIP SelectMode = iota // increase one by one, no IP is leaking
-        RandomSelectIP // random
-	ModLenSelectIP // like LoopIncSelect, may leak IP(s) if id is not continous
+	LoopIncSelectIP SelectMode = iota // increase one by one, no IP is leaking
+	RandomSelectIP                    // random
+	ModLenSelectIP                    // like LoopIncSelect, may leak IP(s) if id is not continous
 )
+
 type IPBlock struct {
-        ip        net.IP
-        hostStart uint64
-        netStart  uint64
-        hostN     uint64
-        netN      uint64
-        weight    uint64
-        mode  SelectMode
+	ip        net.IP
+	hostStart uint64
+	netStart  uint64
+	hostN     uint64
+	netN      uint64
+	weight    uint64
+	mode      SelectMode
 }
 type IPPool []IPBlock
 
@@ -59,7 +61,7 @@ func ipBlockFromRange(s string) *IPBlock {
 		fmt.Println("Negative IP range: ", s)
 		return nil
 	}
-	return &IPBlock{ip0, h0, n0, h1-h0+1, n1-n0+1, 1, LoopIncSelectIP}
+	return &IPBlock{ip0, h0, n0, h1 - h0 + 1, n1 - n0 + 1, 1, LoopIncSelectIP}
 }
 
 func ipBlockFromCIDR(s string) *IPBlock {
@@ -79,18 +81,18 @@ func ipBlockFromCIDR(s string) *IPBlock {
 		return nil
 	}
 	ones, bits := pnet.Mask.Size()
-	nz, hz := 0, bits - ones
+	nz, hz := 0, bits-ones
 	if hz > 64 {
-		nz, hz = hz - 64, 64
+		nz, hz = hz-64, 64
 	}
 	// must: 0 <= z <= 64
-	offsetCIDR := func (zbits int, offset uint64) uint64 {
+	offsetCIDR := func(zbits int, offset uint64) uint64 {
 		switch zbits {
 		case 0:
 			return uint64(1)
 		case 64:
 			n := ^uint64(0) - offset
-			if n + 1 > n {
+			if n+1 > n {
 				n = n + 1
 			}
 			return n
@@ -98,9 +100,9 @@ func ipBlockFromCIDR(s string) *IPBlock {
 			return (uint64(1) << zbits) - offset
 		}
 	}
-        hostNum := offsetCIDR(hz, hk-h0)
-        netNum := offsetCIDR(nz, nk-n0)
-        return &IPBlock{ip0, h0, n0, hostNum, netNum, 1, LoopIncSelectIP}
+	hostNum := offsetCIDR(hz, hk-h0)
+	netNum := offsetCIDR(nz, nk-n0)
+	return &IPBlock{ip0, h0, n0, hostNum, netNum, 1, LoopIncSelectIP}
 }
 
 // GetRandomIP return a random IP by seed from an IP block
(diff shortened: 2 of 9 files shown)


js/runner.go
```diff
@@ -164,6 +255,14 @@ func (r *Runner) newVU(id int64, samplesOut chan<- stats.SampleContainer) (*VU,
 		Blacklist: r.Bundle.Options.BlacklistIPs,
 		Hosts:     r.Bundle.Options.Hosts,
 	}
+	if r.Bundle.Options.ClientIpRange.Valid {
+		ippool := GetIpPool(r.Bundle.Options.ClientIpRange.ValueOrZero())
```
divfor, line 259: GetIpPool() should be called once in SetOptions, but I am not familiar with that code area to support it on both CLI and Config file...

js/runner.go
```diff
@@ -113,6 +116,97 @@ func newFromBundle(logger *logrus.Logger, b *Bundle) (*Runner, error) {
...
+                netPos := pool.netStart + netShift
+                hostPos := pool.hostStart + hostShift
+                if hostPos < pool.hostStart {
+                        netPos = netPos + 1
+                }
+                if ip6 := make(net.IP, net.IPv6len); ip6 != nil {
+                        binary.BigEndian.PutUint64(ip6[:8], netPos)
+                        binary.BigEndian.PutUint64(ip6[8:], hostPos)
+                        return ip6
+                }
```
MStoykov, line 205: I don't think I like that :(. 

You might have seen this in the wild and maybe this is "common" practice but given how ipv6 is supposed to be structured I find it very unlikely that you will get a mask that is ... THIS big :D 

And as I've argued before we probably should give IPs to VUs in some orderly way so that if you have 20 IPs and 40 VUs -  each IP will be used by 2 VUs which currently doesn't seem like something that will happen.

MStoykov (changes requested):
LGTM so far!
There are a couple of things that I think must be changed before we can merge this :
1. This shouldn't be configurable from the script or a config file as I don't think this will be portable or all the useful. 
2. Support a single IP. 

And a couple that I would like, but arguably can be changed later and it can be done in a separate PR.
1. support multiple ranges at the same time 
2. getting the IP should not be randomized between the VUs IMO, it should be ordered as that will give better and more easily understandable scalability.

I would recommend you wait for a day or two and rebase on master when we have merged the Github Actiosn PR as well.  

Maybe some more comments around the IP range calculations will be nice too.

Great job and thansk again for this PR :tada:

js/runner.go
```diff
@@ -113,6 +116,97 @@ func newFromBundle(logger *logrus.Logger, b *Bundle) (*Runner, error) {
...
+
+// GetRandomIP return a random-by-seed selected IP from an IP pool
+// IPv4 pool is considered as a uniform distribution from start IP to end IP
+// IPv6 poop is considered as two independent uniform distributions: high 64-bit and low 64-bit
+func (pool *IpPool) GetRandomIP(seed int64) net.IP {
+        var hostShift, netShift uint64
+        r := rand.New(rand.NewSource(seed))
+        if pool.hostRangeNum > 0 {
+                hostShift = r.Uint64() % pool.hostRangeNum
+        }
```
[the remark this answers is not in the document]
divfor: 1. sometimes VUs number is more than IP addrs available so they must share IPs, that is why I use rand() % ip_num.   for 1:1 ordered binding, specify a mode to not using rand() but only using vu_id % ip_num?
2. instead of calculating, unrolling all range IPs to a linear array may require too much memory if the range is very big - especially with CIDR format. 
3. for multiple ranges, how to map VUs to all possible IPs? the later ranges will be less possibility to be picked if append to array list.

js/runner.go
divfor: this is used to test IPv6 only - both of network part and host part should has a range to change. typically, for load balance algorithm like RSS that requires a good hash key including both of network and host bytes. if you don't need this feature, just set a mask >= 64 or a small range.

divfor: from my case, I often use 40K IPs with one Linux machine:)

js/runner.go
```diff
@@ -113,6 +116,97 @@ func newFromBundle(logger *logrus.Logger, b *Bundle) (*Runner, error) {
...
+        for i := 0; i < len(s); i++ {
+                switch s[i] {
+                case '-':
+                        return ipPoolFromRange(s)
+                case '/':
+                        return ipPoolFromCidr(s)
+                }
+        }
+        return nil
+}
```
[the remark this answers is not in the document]
divfor: yes, if use one IP to specify, this code requires --ip=10.10.10.1-10.10.10.1, or remove this option and K6 will pick up interface's primary IP

lib/options.go
```diff
@@ -382,6 +382,9 @@ type Options struct {
 
 	// Redirect console logging to a file
 	ConsoleOutput null.String `json:"-" envconfig:"K6_CONSOLE_OUTPUT"`
+
+	// Specify client IP range or CIDR from which each VU binds to one IP (random seeded by VU id)
+	ClientIpRange null.String `json:"clientIpRange" envconfig:"K6_CLIENT_IP_RANGE"`
```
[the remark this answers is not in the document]
divfor: it is ok for me  -- I copy this but not quite sure what it will do.

js/runner.go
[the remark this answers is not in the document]
divfor: @MStoykov Will VUs be able to change their local IP after initialisation? looks like NewVU() is called only in scale up stage

js/runner.go
divfor: // IPv4: one IP block from start IP to end IP
// IPv6: two independent IP blocks: high 64-bit for net and low 64-bit for hosts
// This is useful if you want to travel some IPv6 nets without traveling 2^64 hosts

MStoykov: As I explained above ... having 2^65 ips for *outgoing* requests ... seems impossible at least for anything that k6 is currently runnig on. Even at 40k IPs ... your OS is probably using more memory just to configure this that k6 will use to unroll them :). 

I might be missing some strange use case and way to do this but as far as I can see ... while you can having full subnet for incoming requests with [anyip](https://blog.widodh.nl/2016/04/anyip-bind-a-whole-subnet-to-your-linux-machine/) this seems to not work for outgoing ones ... which more or less removes the need for k6 to support subnets of the size of /65+, actually a lot less as I would expect the resource usage of a OS to overwhelm thsi pretty fast.

Also this is pretty impractical for k6 ... as you are unlikely to have enough VUs to use even a fraction of the IPs  on a single machine (I would really like it if it was possible to run 2^20 VUs .. but even with 1mb of memory per each this is like 1TB of memory ... just for the VUs which still means that the unrolling of the IPs will be a drop in the ocean of memory used

divfor: local type route is also usable for k6 outgoing requests, so no need to setup IPs to linux kernel -- actually I was using it for a while using serval GB memory.

divfor: here is my K6 using 4k VUs:
image
image

js/runner.go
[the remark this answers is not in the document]
divfor: it can match requirements like:
1. multiple IPs: ip1,ip2,ip3
2. multiple IPs and a range: ip1,ip2,ip3,ip4-ip10|0|6
3. multiple ranges: ip1-ip8,ip10/29

cmd/options.go
```diff
@@ -91,6 +91,7 @@ func optionFlagSet() *pflag.FlagSet {
 	flags.StringSlice("tag", nil, "add a `tag` to be applied to all samples, as `[name]=[value]`")
 	flags.String("console-output", "", "redirects the console logging to the provided output file")
 	flags.Bool("discard-response-bodies", false, "Read but don't process or save HTTP response bodies")
+	flags.StringP("ip", "", "", "Client IP Range or CIDR from which each VU selects one by VU id")
```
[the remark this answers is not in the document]
divfor: now they are all supported

js/runner.go
MStoykov: Can you elaborate on how you actually managed to get requests out of a /65 ipv6 network ... for example

divfor:
maybe I did not talk it clear: 

The weight of each IPBlock is actually the number IPs planned to use:
VU id will loop back to the first IPBlock when total planed IPs number (it is all weights' sum) are picked up, so by default it is sequential for IPBlocks (also for single IPs). 

sequential and random are only subject to inside IPBlock:
when picked an IPBlock, mode will decide which IP will be selected: by default it is LoopIncSelectIP (VU id % range-num), or set to RandomSelectIP for random
Examples could verify from last commit:

1. all single sequential:  1.1.1.1,2.2.2.2,3.3.3.3
2. sequential 3 single IPs and then sequential 9 range IPs: 1.1.1.1,2.2.2.2,3.3.3.1-3.3.3.9
3. sequential 3 single IPs and then sequential 256 cidr IPs: 1.1.1.1,2.2.2.2,3.3.3.0/24
4. sequential 3 single IPs and then sequential 10 cidr IPs: 1.1.1.1,2.2.2.2,3.3.3.0/24|0|10
5. sequential 3 single IPs and then random 10 cidr IPs: 1.1.1.1,2.2.2.2,3.3.3.0/24|1|10

btw, IPv6 has similar usage

js/runner.go
divfor: For example, a /65 IPv6 range format like fd00:1:2:0::0-fd00:1:2:ff::3ff means:
network will be range of 0 ~ 255 and selected by VU_id % IPBlock.netN
host will be range of 0 ~ 1023 and selected by VU_id % IPBlock.hostN

Of cause, if one don't want so strange range method(it is most like case to test load balancer), just give host part for range or give a CIDR mask <= /64

MStoykov: I just now noticed that you have actually explained how to make the setup in the PR description :facepalm: 
I did manage to add /31 network (I only have ipv4 and didn't want to break my network configuring stuff to test this :D ) and it seems to work ... I am still somewhat skeptic that it will be a common practice to use that on load generator machine ... but it's fine to support it

js/iprange.go
```diff
@@ -0,0 +1,192 @@
...
+	"strings"
+)
+
+type SelectMode uint
+
+const (
+	LoopIncSelectIP SelectMode = iota // increase one by one, no IP is leaking
+	RandomSelectIP                    // random
+	ModLenSelectIP                    // like LoopIncSelect, may leak IP(s) if id is not continous
+)
```
MStoykov: This will need to have better user-facing names, even if they are only two see [compatibilityMode](https://github.com/loadimpact/k6/blob/master/lib/runtime_options.go#L30-L40) for how it's done in other places

js/iprange_test.go
```diff
@@ -0,0 +1,201 @@
...
+
+func TestGetIPBlock(t *testing.T) {
+	testdata := map[string]struct {
+		hostCount uint64
+		netCount  uint64
+		ipBlock   string
+	}{
+		"ipv4 range 1": {100, 1, "192.168.0.101-192.168.0.200"},
+		"ipv4 range 2": {101, 1, "192.168.0.100-192.168.0.200"},
+		"ipv4 cidr 1":  {256, 1, "192.168.0.0/24"},
```
MStoykov, line 85: But ... there are 254 actuallu usable addresses ... and the code doesn't return `.0` and `.255` unless they are the only IP ... which I think is the correct behaviour. But than this 256 is wrong

js/iprange.go
```diff
@@ -0,0 +1,189 @@
+package js
```
MStoykov, line 1: This whole file should be in lib/types

MStoykov (commented):
The more I look at the code ... the more it looks like ... you've added the options but didn't implement them 
- weights is always one
- the mode is always loopIncSelect

and the more it seems that we wouldn't have time for most of those.

So at this point in time I am of the opinion that it will be better to scale this down to just single ips and get that inside (no weights, no random) and try to work on ranges and cidr networks after that(which will likely push them off until v0.30.0)

The reason for dropping ranges and (especially) cidrs is that looking at the math required ... I do think this will take a very long time to review and I much more prefer to have at least some multi IP support that we can expand on ... instead of nothing.

But even if you are very determent to have most of the other things ... I don't think weights should be part of this release at all (especially not in the way that you described them) and the mode needs to be between all the IPs ... 

Given the fact that we have 2 weeks to get this merged (I don't want this to be in the last week before the release) and the need for testing of this (which will probably take longer than even I expect) I am going to have to take over if you don't scale this back down, sorry. There just isn't enough time for me to look into everything around how the math (should) work and then try to figure out how to document it on top of all the other things in this PR and the others I have to work on.

I hope you agree to scale this down and work on it, but regardless you will be credited and will be welcome to add whatever is dropped from this PR in the future :)

js/iprange.go
divfor: any suggestion for the names?

js/runner.go
```diff
@@ -164,6 +164,20 @@ func (r *Runner) newVU(id int64, samplesOut chan<- stats.SampleContainer) (*VU,
 		Blacklist: r.Bundle.Options.BlacklistIPs,
 		Hosts:     r.Bundle.Options.Hosts,
 	}
+	if r.Bundle.Options.ClientIpRange.Valid {
+		ippool := GetPool(r.Bundle.Options.ClientIpRange.ValueOrZero())
+		if len(ippool) < 1 {
+			return nil, errors.New("Failed to parse IP Range")
+		}
```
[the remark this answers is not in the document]
divfor: now I move the parsing into newRunner()

js/iprange_test.go
divfor: yes. this just tests for board checking. in practice a user should use  "ipv4 cidr 1":  {254, 1, "192.168.0.01/24|0|254"}

js/iprange.go
divfor: done

divfor:
Actually I have tested and confirmed all have been featured with mixed of range/weight/random.
Okey, let's focus to simply the release notes:  only support single IPs and range/cidr, no weights no random

which one do you prefer to ?
1. keep current commit code, only verify the supported option format
2. I remove current commit code and re-write a simple version and test them, re-submit a PR

MStoykov:
I think it will be better if there is a new PR with ... less commits :)
I recommend that you
1. rebase on the latest master 
2. configure your ide/text editor for golang development as that will automatically run for you gofmt/goimports.
3. run golangci-lint as in ` golangci-lint run --new-from-rev=origin/master` I have this as a pre-push hook in git:
```
#!/bin/sh

# An example hook script to verify what is about to be pushed.  Called by "git
# push" after it has checked the remote status, but before anything has been
# pushed.  If this script exits with a non-zero status nothing will be pushed.
#
# This hook is called with the following parameters:
#
# $1 -- Name of the remote to which the push is being done
# $2 -- URL to which the push is being done
#
# If pushing without using a named remote those arguments will be equal.
#
# Information about the commits which are being pushed is supplied as lines to
# the standard input in the form:
#
#   <local ref> <local sha1> <remote ref> <remote sha1>
#
# This sample shows how to prevent push of commits where the log message starts
# with "WIP" (work in progress).

remote="$1"
url="$2"

z40=0000000000000000000000000000000000000000

while read local_ref local_sha remote_ref remote_sha
do
        if [ "$local_sha" = $z40 ]
        then
                # Handle delete
                :
        else
                if [ "$remote_sha" = $z40 ]
                then
                        # New branch, examine all commits
                        since="origin/master"
                else
                        # Update to existing branch, examine new commits
                        since="$remote_sha"
                fi

        exec golangci-lint run --new-from-rev=$since
        fi
done

exit 0
```

divfor:
okey, let me try to follow up this -- as I am a 3 weeks golang newbie from a QA tester using k6...

divfor:
closed and move #1674 for less commit

PR #1660, opened 2020-10-07
Try to stdlib errors.Unwrap errors as well

MStoykov:


MStoykov:
This is based on changes that I did for https://github.com/loadimpact/k6/pull/1637 but more ... general. Unfortunately, because we use two `errors` libs I need to also "Unwrap" with both of them, but possibly we can remedy this in a future PR

imiric (approved):
LGTM. Could we have a test for this?

lib/netext/httpext/error_codes.go
```diff
@@ -107,7 +108,25 @@ func http2ErrCodeOffset(code http2.ErrCode) errCode {
...
 // errorCodeForError returns the errorCode and a specific error message for given error.
 func errorCodeForError(err error) (errCode, string) {
-	switch e := errors.Cause(err).(type) {
+	inner := stderrors.Unwrap(err)
+	if inner != nil && inner != err {
+		code, resultErr := errorCodeForError(inner)
+		if code != defaultErrorCode {
+			return code, resultErr
+		}
+	}
```
na--, line 117: Is this code actually necessary? Judging by the the [code for `errors.Cause`](https://github.com/pkg/errors/blob/614d223910a179a466c1767a985424175c39b465/errors.go#L272-L288), I don't see a reason to do it like this

na-- (commented):
Also, I know this is tricky to test, but can't we find 1 easily triggered stdlib error where the new Go error wrapping is used and test that?

lib/netext/httpext/error_codes.go
MStoykov: What do you mean ... by is it necessary? If we have an error that has `Unwrap` method and one that has `Cause` ... well they don't work together, which is the whole point of this PR to "Unwrap" the corresponding errors ... correspondingly

na--: You still have the same `Unwrap()` code below, on the result of `errors.Cause(err)`, which [will be a no-op](https://github.com/pkg/errors/blob/614d223910a179a466c1767a985424175c39b465/errors.go#L272-L288) if the error wasn't wrapped with `github.com/pkg/errors`

MStoykov: Yeah ... you are right :facepalm: , I first added this one and then ... figured out that I need the second one as well, but didn't think it through that ... well I wouldn't need the first one

lib/netext/httpext/error_codes.go
```diff
@@ -133,14 +144,16 @@ func errorCodeForError(err error) (errCode, string) {
...
-		if e.Op == "write" {
-			if sErr, ok := e.Err.(*os.SyscallError); ok {
-				switch sErr.Err {
-				case syscall.ECONNRESET:
-					return tcpResetByPeerErrorCode, tcpResetByPeerErrorCodeMsg
-				case syscall.EPIPE:
-					return tcpBrokenPipeErrorCode, tcpBrokenPipeErrorCodeMsg
-				}
+		if sErr, ok := e.Err.(*os.SyscallError); ok {
+			switch sErr.Unwrap() {
```
na--, line 148: I am confused - wouldn't this have been unwrapped in the call above? After all, [`net.OpError` has `Unwrap()`](https://golang.org/pkg/net/#OpError.Unwrap). Maybe that automatic `Unwrap()` above should only happen in the `default:` case of this big `switch`?

MStoykov: Here we have syscall.ECONNRESET inside os.SyscallError inside net.OpError maybe inside something else. 

I think that for this chain specifically everything implements Unwrap now ... but it didn't when this code was added IIRC. I really don't want to put syscall.ECONNRESET in the top switch because:
1. that will mean the top switch will be forever long 
2. the windows fix here won't work or will need even MORE working around it >.>
3. The e.Op will get lost 

Moving the Unwrap to the default seems like a not bad idea ... should we also move the Cause then ? 

I am not really certain what wraps what and where and whether us moving things around will not break something as ... well most of the tests for this are syntethic - I created a specific wrappings of errors that than goes through this code, but as evident by the fact it hasn't caught some errors ... my matroshkas aren't what is genereated in the actual code of the application. So I do prefer to change this less ... if possible :D

MStoykov: Moving Unwrap breaks the TestDNsError as it DNSError inside OpError ... maybe I should just move DNSError under OpError :thinking:

na--: > Moving the Unwrap to the default seems like a not bad idea ... should we also move the Cause then ?

Yes, this is what I meant, and it is going to make the change more minimal

MStoykov: Unwrap in the begining is the same logic as Cause ... just slower and with a return path:
Cause basically says get me the inner error of my Cause error that doesn't implement the causer interface. 
What happens with the Unwrap above is that we Unwrap (and Cause) until we get to an error that no longer implements neither on (or just doesn't have inner error) and than we check it, if it is a concrete error - we return that and if not we go one level above and try that error 

Looking at the document [Unwrap](https://godoc.org/github.com/pkg/errors#Unwrap) was added in a later version to github.com/pkg/errors adn arguably if we upgrade the version we can drop the errors.Cause   :thinking:

MStoykov:
closed in favor of #1951

PR #1663, opened 2020-10-08
Update babel to 7.11.6.

legander:
Update babel from [6.26.0, preset=latest] to [7.11.6, preset=env].

- Improving transpile time quite significantly for large scripts.
- Needs some testing to ensure same level of ES feature support as previous build.


na--:
With the latest changes in goja (native support block-scoped variables, added in https://github.com/dop251/goja/pull/264 and adopted in https://github.com/k6io/k6/pull/1904), we are probably not going to update babel soon, so I'll close this issue... cc @sniku 

In any case, if we want to update babel, we'll definitely need another PR, because:
- we have to take the TC39 tests that we now have into account, we added them [in the k6 CI](https://github.com/k6io/k6/blob/e5d4d82d16c0c86154f017971f4233b43702165d/.github/workflows/tc39.yml#L1), after this PR was submitted (https://github.com/k6io/k6/pull/1747)
- we'll likely not use [go.rice](https://github.com/GeertJohan/go.rice) to embed the file, we want to get rid of that dependency (https://github.com/k6io/k6/issues/1933) and use the [native Go file embedding ](https://golang.org/doc/go1.16#library-embed), since that way we won't have 2 copies of the file in the git repo
- we'd probably want to use the newest Babel version and we might have to adjust the preset

Issue #1601, opened 2020-08-21
Add recover() check on each iteration to guard against goja panics


na--:
We've seen this panic when a k6 instance was stopped (essentially equivalent to `curl -X PATCH --data '{"data":{"type":"status","id":"default","attributes":{"stopped":true}}}' 'http://localhost:6565/v1/status'`):
```
exit status 2

panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x8ecaa0]

goroutine 106 [running]:
github.com/loadimpact/k6/vendor/github.com/dop251/goja.AssertFunction.func1.1(0xc028769b08)
	/home/alpine/go/src/github.com/loadimpact/k6/vendor/github.com/dop251/goja/runtime.go:1489 +0x98
panic(0xd7b6a0, 0x19917f0)
	/usr/local/go/src/runtime/panic.go:969 +0x166
github.com/loadimpact/k6/vendor/github.com/dop251/goja.(*vm).try.func1(0xc006eabee0, 0x0, 0xc0287699e0, 0x0, 0x0, 0x0, 0xc028769a58)
	/home/alpine/go/src/github.com/loadimpact/k6/vendor/github.com/dop251/goja/vm.go:377 +0x3da
panic(0xd7b6a0, 0x19917f0)
	/usr/local/go/src/runtime/panic.go:969 +0x166
github.com/loadimpact/k6/vendor/github.com/dop251/goja.(*vm).run(0xc006eabee0)
	/home/alpine/go/src/github.com/loadimpact/k6/vendor/github.com/dop251/goja/vm.go:289 +0x60
github.com/loadimpact/k6/vendor/github.com/dop251/goja.(*funcObject).Call(0xc00b8d5e00, 0x10ec460, 0x19d6600, 0xc0253f23b0, 0x1, 0x1, 0x203009, 0xc028769990)
	/home/alpine/go/src/github.com/loadimpact/k6/vendor/github.com/dop251/goja/func.go:154 +0x2cc
github.com/loadimpact/k6/vendor/github.com/dop251/goja.AssertFunction.func1.2()
	/home/alpine/go/src/github.com/loadimpact/k6/vendor/github.com/dop251/goja/runtime.go:1494 +0x96
[...]

Issue #1487, opened 2020-06-05
Panic when slicing HTML results


na--:
## Expected Behavior
The following code:
```js
import html from "k6/html";

export default function () {
    var r = html.parseHTML("<html>");
    var o = r.find(".something").slice(0, 4).toArray()
}
```

will result in a panic like:
```
panic: runtime error: slice bounds out of range [:4] with capacity 0 [recovered]
	panic: runtime error: slice bounds out of range [:4] with capacity 0 [recovered]
	panic: runtime error: slice bounds out of range [:4] with capacity 0

goroutine 44 [running]:
github.com/loadimpact/k6/vendor/github.com/dop251/goja.AssertFunction.func1.1(0xc003647c48)
	/home/go/src/github.com/loadimpact/k6/vendor/github.com/dop251/goja/runtime.go:1489 +0x98
panic(0xe05940, 0xc0012ede60)
	/usr/lib/go/src/runtime/panic.go:969 +0x166
github.com/loadimpact/k6/vendor/github.com/dop251/goja.(*vm).try.func1(0xc0018d11e0, 0x0, 0xc003647b20, 0x0, 0x0, 0x0, 0xc003647b98)
	/home/go/src/github.com/loadimpact/k6/vendor/github.com/dop251/goja/vm.go:377 +0x3da
panic(0xe05940, 0xc0012ede60)
	/usr/lib/go/src/runtime/panic.go:969 +0x166
github.com/loadimpact/k6/vendor/github.com/PuerkitoBio/goquery.(*Selection).Slice(...)
	/home/go/src/github.com/loadimpact/k6/vendor/github.com/PuerkitoBio/goquery/array.go:67
github.com/loadimpact/k6/js/modules/k6/html.Selection.Slice(0xc000b66900, 0xc001093f80, 0x0, 0x0, 0x0, 0xc000c67808, 0x1, 0x1, 0xd0f000, 0xc001060270, ...)
	/home/go/src/github.com/loadimpact/k6/js/modules/k6/html/html.go:397 +0x2b9
reflect.callMethod(0xc00181f880, 0xc0036473d8, 0xc0036473c0)
	/usr/lib/go/src/reflect/value.go:710 +0x1ec
reflect.methodValueCall(0x0, 0xc000c67808, 0x1, 0x1, 0x0, 0x0, 0x0, 0x0, 0xc003647630, 0x4d330b, ...)
	/usr/lib/go/src/reflect/asm_amd64.s:35 +0x42
[...]

PR #1665, opened 2020-10-08
Recover from panics caused by goja

2hamed:
Making sure k6 never panics (#1601 and #1487) by adding `defer..recover` blocks. I'm not sure about the performance penalty though.

### DIFF
--- a/lib/executor/constant_vus.go
+++ b/lib/executor/constant_vus.go
@@ -26,13 +26,12 @@ import (
 	"sync"
 	"time"
 
-	"github.com/sirupsen/logrus"
-	"gopkg.in/guregu/null.v3"
-
 	"github.com/loadimpact/k6/lib"
 	"github.com/loadimpact/k6/lib/types"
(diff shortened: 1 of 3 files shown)


na-- (changes requested):
Thanks for making this PR and sorry for the late review!

Unfortunately the current change you propose doesn't encompass all executors. Looking at the code, it might be better to have a single `recover()` block in the lambda returned by the `getIterationRunner()` helper function, which is used in all executors: https://github.com/loadimpact/k6/blob/c22338c57edc61d808c5d0ec4f3b6f2c3b68fcfb/lib/executor/helpers.go#L82

Also, I don't see a reason for changing the import order?

Edit: also, could you please rebase on the latest `master`, we've switched our CI system to GitHub Actions now.

MStoykov:
Thanks for the PR @2hamed , but I wanted to actually get this in v0.29.0 and add some goja stack trace so I did this in #1697, and now that this is merged I need to close this PR, sorry :(.

PR #1532, opened 2020-07-03
Add blocking by hostname patterns (--block-hostname)

krashanoff:
This is an implementation of the `--block-hostname` option as described in #972.

It provides:
* The ability to block tests from accessing particular hostnames that match provided patterns
* A limited wildcard syntax for hostname blocking: wildcards are permitted *only* at the very start of the pattern string, which should consist only of valid letters in any language, numbers, or `.`
* Parameters through the environment variable `K6_BLOCK_HOSTNAMES`, the JSON field `blockHostnames`, and the CLI option `--block-hostname`
* A [new error](https://github.com/krashanoff/k6/blob/feature/block-hostname/lib/netext/httpext/error_codes.go#L52) (code 1111) for request failure due to the hostname being blocked.
* An implementation of a trie to store and search blocked hostname patterns quickly
* Tests for the feature

### Implementation details

The trie is implemented as typename `HostnameTrie` in [lib/options.go](https://github.com/krashanoff/k6/blob/feature/block-hostname/lib/options.go#L195). It is designed to handle internationalized hostnames and the aforementioned limited wildcard syntax.

```go
type HostnameTrie struct {
	children []*HostnameTrie
	r        rune
	terminal bool // end of a valid match
}
```

The only thing notable is that a rune is held at each node instead of a byte to address internationalized hostnames.

The provided interface for the type is:
* `(*HostnameTrie) Insert(s string) error`
    - Insert a new pattern into the HostnameTrie. Returns a non-nil error if the pattern is illegal (e.g. breaks the aforementioned syntax).
* `(*HostnameTrie) Contains(s string) (bool, string)`
    - Check if a hostname or string matches a pattern in the HostnameTrie. Returns whether a match was found, and the matched path if one was found.
* `(*HostnameTrie) UnmarshalJSON([]byte) error`
* `(*HostnameTrie) UnmarshalText([]byte) error`

A pointer to this type is passed to the [Dialer as BlockedHostnames](https://github.com/krashanoff/k6/blob/feature/block-hostname/lib/netext/dialer.go#L45) and to the [Options as the same](https://github.com/krashanoff/k6/blob/feature/block-hostname/lib/options.go#L358).

### golintci warnings

There are two golintci warnings in this PR, being a line that's too long in [cmd/options.go](https://github.com/krashanoff/k6/blob/feature/block-hostname/cmd/options.go#L66) and use of a global variable in [lib/options.go](https://github.com/krashanoff/k6/blob/feature/block-hostname/lib/options.go#L205).

[...]

PR #1666, opened 2020-10-09
Feature/block hostname

MStoykov:
This supersedes #1532


lib/types/hostnametrie.go
```diff
@@ -0,0 +1,162 @@
+package types
```
imiric, line 1: Don't forget to add the copyright header to this file and `hostnametrie_test.go`.

lib/types/hostnametrie_test.go
```diff
@@ -0,0 +1,33 @@
...
+	assert.True(t, matches)
+	assert.Equal(t, "test.k6.io", match)
+	match, matches = trie.Contains("TEST.K6.IO")
+	assert.True(t, matches)
+	assert.Equal(t, "test.k6.io", match)
+	match, matches = trie.Contains("blocked.valId.paTtern")
+	assert.True(t, matches)
+	assert.Equal(t, "*valid.pattern", match)
+	_, matches = trie.Contains("example.test.k6.io")
+	assert.False(t, matches)
```
imiric, line 32: This test can be better structured with a `testCases` array.

lib/netext/dialer.go
```diff
@@ -165,6 +177,11 @@ func (d *Dialer) findRemote(addr string) (*lib.HostAddress, error) {
 
 	remote, err := d.getConfiguredHost(addr, host, port)
 	if err != nil || remote != nil {
+		if err == nil && d.BlockedHostnames != nil && net.ParseIP(host) == nil {
```
imiric, line 180: `net.ParseIP(host)` is already called on line 188. Couldn't you move that before the `d.getConfiguredHost()` call and then you'll know you're dealing with a hostname here?

lib/netext/dialer.go
```diff
@@ -173,6 +190,11 @@ func (d *Dialer) findRemote(addr string) (*lib.HostAddress, error) {
 		return lib.NewHostAddress(ip, port)
 	}
 
+	if d.BlockedHostnames != nil {
+		if match, blocked := d.BlockedHostnames.Contains(host); blocked {
+			return nil, BlockedHostError{hostname: host, match: match}
+		}
+	}
```
imiric, line 197: Why is this block repeated here?

lib/types/hostnametrie.go
```diff
@@ -0,0 +1,162 @@
...
+	return h, nil
+}
+
+// Regex description of hostname pattern to enforce blocks by. Global var
+// to avoid compilation penalty at runtime.
+// based on regex from https://stackoverflow.com/a/106223/5427244
+//nolint:gochecknoglobals,lll
+var legalHostnamePattern *regexp.Regexp = regexp.MustCompile(`^(\*\.?)?((([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\-]*[a-zA-Z0-9])\.)*([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\-]*[A-Za-z0-9]))?$`)
+
+func legalHostname(s string) error {
```
imiric, line 103: Nit: I think `valid*` is a more suitable name here, especially since the returned error contains "invalid". I would also replace "illegal" with "invalid" in the comment on line 111.

lib/netext/dialer.go
na--: I think because the other branch checks against `options.hosts`, though a comment wouldn't be amiss there...

na-- (commented):
I didn't spot anything that @imiric hadn't already commented on, so LGTM besides these comments

lib/netext/dialer.go
MStoykov: The idea was that I didn't want to parse the IP unless I needed to but looking at this code again ... the "speed up" in the case that there is no blockhostname and the host is configured in `options.host` will likely be miniscule and is the only case where there is a speed up so ... I am just going to do it before the options.host logic

lib/types/hostnametrie.go
MStoykov: fixed in 2f729eca

lib/netext/dialer.go
MStoykov: fixed in ce17b17a

lib/netext/dialer.go
MStoykov: fixed in ce17b17a

lib/types/hostnametrie.go
MStoykov: fixed in c4262f92

lib/types/hostnametrie_test.go
MStoykov: fixed in 8321e9b8

lib/types/hostnametrie.go
```diff
@@ -0,0 +1,182 @@
...
+
+func isValidHostnamePattern(s string) error {
+	if len(validHostnamePattern.FindString(s)) != len(s) {
+		return errors.Errorf("invalid hostname pattern %s", s)
+	}
+	return nil
+}
+
+// insert a hostname pattern into the given HostnameTrie. Returns an error
+// if hostname pattern is valid.
```
imiric, line 131:
```suggestion
// if hostname pattern is invalid.
```
😄

na-- (approved):

PR #1669, opened 2020-10-14
CI: Disable -race for 1.14.x tests

MStoykov:


.github/workflows/all.yml
```diff
@@ -74,7 +74,7 @@ jobs:
           set -x
           go version
           export GOMAXPROCS=2
-          args=("-p" "2" "-race")
+          args=("-p" "2")
           # Run with less concurrency on Windows to minimize flakiness.
           if [[ "${{ matrix.platform }}" == windows* ]]; then
             unset args[2]
```
imiric, line 80: You should also remove this line then.

MStoykov: I think we aren't going to go with this change at all ... but will keep the PR open for a few days and add this change if this is what we go with

PR #1671, opened 2020-10-14
Revert the check in TestCancelledRequest

MStoykov:


MStoykov:
Relevant issue https://github.com/loadimpact/k6/issues/1672

lib/netext/httpext/tracer_test.go
```diff
@@ -216,20 +216,21 @@ func TestCancelledRequest(t *testing.T) {
...
 			cancel()
 		}()
 
-		resp, err := srv.Client().Transport.RoundTrip(req)
+		resp, err := srv.Client().Transport.RoundTrip(req) //nolint:bodyclose
 		_ = tracer.Done()
-		assert.Nil(t, resp)
-		assert.Error(t, err)
+		if resp == nil && err == nil {
+			t.Errorf("Expected either a RoundTrip response or error or trail errors but got %#v and %#v", resp, err)
```
na--, line 232: Ok, so this test is never going to fail? :confused:

MStoykov: well, that is what I have been trying to explain ... that this is test is strange ... before I "fixed" it it was doing the exact same thing ... 
As far as I could see you tried to tease out a race condition with this ... and just it so happens that it breaks(IMO) ... strangely and so you wrote it in a way ... that it checks at least that it either errors out or returns an actual response ... which IMO isn't a big check but it's ... something :shrug:

imiric: I'm also puzzled by this test... Could the `rand.Int31n(50)` be causing issues if it returns 0? Maybe we need a minimum sleep period there.

Also the "or trail errors" part should be removed from the error message since it's no longer checking for that.

MStoykov: > Could the rand.Int31n(50) be causing issues if it returns 0

I reverted it back to the rand ... previously I was sleeping 100ms every time - no difference

MStoykov, line 232:
```suggestion
			t.Errorf("Expected either a RoundTrip response or error but got %#v and %#v", resp, err)
```

na-- (approved):
I continue to think something is very fishy with this, but fine, let's leave the investigation for https://github.com/loadimpact/k6/issues/1672

PR #1677, opened 2020-10-20
Use InDelta in TestExternallyControlledRun 

MStoykov:


na-- (approved):

imiric (approved):
We shouldn't have to do this, but OK 🙁

lib/executor/constant_arrival_rate_test.go
```diff
@@ -227,7 +228,7 @@ func TestConstantArrivalRateRunCorrectTiming(t *testing.T) {
 				// check that we got around the amount of VU iterations as we would expect
 				var currentCount int64
 
-				for i := 0; i < 3; i++ {
+				for i := 0; i < seconds; i++ {
```
na--, line 231: Erm, I'm confused - won't `seconds` be `0`, so this loop will never execute? :confused:

na-- (approved):
LGTM

na--:
I added this to the v0.29.0 since it fixes the `context cancelled` bug (https://github.com/loadimpact/k6/issues/1485)

PR #1682, opened 2020-10-22
Add multi source IP support

### LINKED ISSUE #476: Support for Multiple Network Card
Hello,

We would like to use K6 on machine that have multiple network card.  As for now K6 does not support multiple NIC for their VUS.

What will be the best way of doing it?

Thanks!

Alex

MStoykov:
fixes #476

### DIFF
--- a/cmd/options.go
+++ b/cmd/options.go
@@ -93,7 +93,7 @@ func optionFlagSet() *pflag.FlagSet {
 	flags.StringSlice("tag", nil, "add a `tag` to be applied to all samples, as `[name]=[value]`")
 	flags.String("console-output", "", "redirects the console logging to the provided output file")
 	flags.Bool("discard-response-bodies", false, "Read but don't process or save HTTP response bodies")
-	flags.String("client-ips", "", "Client IP Ranges and/or CIDRs from which each VU will be making requests")
+	flags.String("local-ips", "", "Client IP Ranges and/or CIDRs from which each VU will be making requests")
 	flags.String("dns", types.DefaultDNSConfig().String(), "DNS resolver configuration. Possible ttl values are: 'inf' "+
 		"for a persistent cache, '0' to disable the cache,\nor a positive duration, e.g. '1s', '1m', etc. "+
 		"Milliseconds are assumed if no unit is provided.\n"+
@@ -206,12 +206,12 @@ func getOptions(flags *pflag.FlagSet) (lib.Options, error) {
 		}
 	}
 
-	ipRangesString, err := flags.GetString("client-ips")
+	localIpsString, err := flags.GetString("local-ips")
 	if err != nil {
 		return opts, err
 	}
-	if flags.Changed("client-ips") {
-		err = opts.ClientIPRanges.UnmarshalText([]byte(ipRangesString))
+	if flags.Changed("local-ips") {
+		err = opts.LocalIPs.UnmarshalText([]byte(localIpsString))
 		if err != nil {
 			return opts, err
 		}
--- a/cmd/options.go
+++ b/cmd/options.go
@@ -93,6 +93,7 @@ func optionFlagSet() *pflag.FlagSet {
 	flags.StringSlice("tag", nil, "add a `tag` to be applied to all samples, as `[name]=[value]`")
 	flags.String("console-output", "", "redirects the console logging to the provided output file")
 	flags.Bool("discard-response-bodies", false, "Read but don't process or save HTTP response bodies")
+	flags.String("client-ips", "", "Client IP Ranges and/or CIDRs from which each VU will be making requests")
 	flags.String("dns", types.DefaultDNSConfig().String(), "DNS resolver configuration. Possible ttl values are: 'inf' "+
 		"for a persistent cache, '0' to disable the cache,\nor a positive duration, e.g. '1s', '1m', etc. "+
 		"Milliseconds are assumed if no unit is provided.\n"+
@@ -200,6 +201,17 @@ func getOptions(flags *pflag.FlagSet) (lib.Options, error) {
 	}
 	if flags.Changed("block-hostname") {
 		opts.BlockedHostnames, err = types.NewNullHostnameTrie(blockedHostnameStrings)
+		if err != nil {
+			return opts, err
+		}
+	}
+
+	ipRangesString, err := flags.GetString("client-ips")
+	if err != nil {
+		return opts, err
+	}
+	if flags.Changed("client-ips") {
+		err = opts.ClientIPRanges.UnmarshalText([]byte(ipRangesString))
 		if err != nil {
 			return opts, err
 		}
--- a/cmd/options.go
+++ b/cmd/options.go
@@ -93,7 +93,8 @@ func optionFlagSet() *pflag.FlagSet {
 	flags.StringSlice("tag", nil, "add a `tag` to be applied to all samples, as `[name]=[value]`")
 	flags.String("console-output", "", "redirects the console logging to the provided output file")
 	flags.Bool("discard-response-bodies", false, "Read but don't process or save HTTP response bodies")
-	flags.String("local-ips", "", "Client IP Ranges and/or CIDRs from which each VU will be making requests")
+	flags.String("local-ips", "", "Client IP Ranges and/or CIDRs from which each VU will be making requests, "+
+		"e.g. '192.168.220.1,192.168.0.10-192.168.0.25', 'fd:1::0/120', etc.")
 	flags.String("dns", types.DefaultDNSConfig().String(), "DNS resolver configuration. Possible ttl values are: 'inf' "+
 		"for a persistent cache, '0' to disable the cache,\nor a positive duration, e.g. '1s', '1m', etc. "+
 		"Milliseconds are assumed if no unit is provided.\n"+
--- a/js/runner.go
+++ b/js/runner.go
@@ -170,8 +170,8 @@ func (r *Runner) newVU(id int64, samplesOut chan<- stats.SampleContainer) (*VU, 
 		BlockedHostnames: r.Bundle.Options.BlockedHostnames.Trie,
 		Hosts:            r.Bundle.Options.Hosts,
 	}
-	if r.Bundle.Options.ClientIPRanges.Valid {
-		dialer.Dialer.LocalAddr = &net.TCPAddr{IP: r.Bundle.Options.ClientIPRanges.Pool.GetIP(uint64(id - 1))}
+	if r.Bundle.Options.LocalIPs.Valid {
+		dialer.Dialer.LocalAddr = &net.TCPAddr{IP: r.Bundle.Options.LocalIPs.Pool.GetIP(uint64(id - 1))}
 	}
 
 	tlsConfig := &tls.Config{
--- a/js/runner.go
+++ b/js/runner.go
@@ -171,9 +171,9 @@ func (r *Runner) newVU(id int64, samplesOut chan<- stats.SampleContainer) (*VU, 
 		Hosts:            r.Bundle.Options.Hosts,
 	}
 	if r.Bundle.Options.LocalIPs.Valid {
-		ipIndex := uint64(id - 1)
-		if id == 0 {
-			ipIndex = 1
+		var ipIndex uint64
+		if id > 0 {
+			ipIndex = uint64(id - 1)
 		}
 		dialer.Dialer.LocalAddr = &net.TCPAddr{IP: r.Bundle.Options.LocalIPs.Pool.GetIP(ipIndex)}
 	}
--- a/js/runner.go
+++ b/js/runner.go
@@ -171,7 +171,11 @@ func (r *Runner) newVU(id int64, samplesOut chan<- stats.SampleContainer) (*VU, 
 		Hosts:            r.Bundle.Options.Hosts,
 	}
 	if r.Bundle.Options.LocalIPs.Valid {
-		dialer.Dialer.LocalAddr = &net.TCPAddr{IP: r.Bundle.Options.LocalIPs.Pool.GetIP(uint64(id - 1))}
+		ipIndex := uint64(id - 1)
+		if id == 0 {
+			ipIndex = 1
+		}
+		dialer.Dialer.LocalAddr = &net.TCPAddr{IP: r.Bundle.Options.LocalIPs.Pool.GetIP(ipIndex)}
 	}
 
 	tlsConfig := &tls.Config{
--- a/js/runner.go
+++ b/js/runner.go
@@ -170,6 +170,10 @@ func (r *Runner) newVU(id int64, samplesOut chan<- stats.SampleContainer) (*VU, 
 		BlockedHostnames: r.Bundle.Options.BlockedHostnames.Trie,
 		Hosts:            r.Bundle.Options.Hosts,
 	}
+	if r.Bundle.Options.ClientIPRanges.Valid {
+		dialer.Dialer.LocalAddr = &net.TCPAddr{IP: r.Bundle.Options.ClientIPRanges.Pool.GetIP(uint64(id - 1))}
+	}
+
 	tlsConfig := &tls.Config{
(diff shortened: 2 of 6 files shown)


cmd/options.go
```diff
@@ -93,6 +93,7 @@ func optionFlagSet() *pflag.FlagSet {
 	flags.StringSlice("tag", nil, "add a `tag` to be applied to all samples, as `[name]=[value]`")
 	flags.String("console-output", "", "redirects the console logging to the provided output file")
 	flags.Bool("discard-response-bodies", false, "Read but don't process or save HTTP response bodies")
+	flags.String("client-ips", "", "Client IP Ranges and/or CIDRs from which each VU will be making requests")
```
na--, line 96: `--local-ips` or `--local-addr` seems slightly better as a name. This is also what Go uses, see https://golang.org/pkg/net/#IPConn.LocalAddr, https://golang.org/pkg/net/#DialTCP

lib/types/ipblock.go
```diff
@@ -0,0 +1,173 @@
...
+
+func getIPBlock(s string) (*ipBlock, error) {
+	switch {
+	case strings.Contains(s, "-"):
+		return ipBlockFromRange(s)
+	case strings.Contains(s, "/"):
+		return ipBlockFromCIDR(s)
+	default:
+		if net.ParseIP(s) == nil {
+			return nil, fmt.Errorf("%s is not a valid ip, range ip or CIDR", s)
```
na--, line 35:
```suggestion
			return nil, fmt.Errorf("%s is not a valid IP, IP range or CIDR", s)
```

lib/types/ipblock.go
```diff
@@ -0,0 +1,173 @@
...
+	"strings"
+)
+
+// ipBlock represents a continuous segment of IP addresses
+type ipBlock struct {
+	// TODO rename count as it is actually the accumulation of the counts up to and including this
+	// on in the pool
+	// maybe add another field, although technically we need count only to find out when to enter this
+	// block not for anything else
+	firstIP, count *big.Int
```
na--, line 17: This was very confusing, since in the context of `ipBlock` you're using count like a count, but in the context of the `IPPool`, you're completely changing its meaning... I'd suggest you simply have 2 different types, an `ipBlock` with the current defintion and a new `ipPoolBlock` with a `firstIP` and `totalCount` or something like that (`lastIP`?). 

You can still reuse the `count` variable from the `ipBlock`, since you won't need it for anything else, so that you won't loose any performance. But it will be much cleaner semantically to have a different for it.

lib/types/ipblock.go
```diff
@@ -0,0 +1,173 @@
...
+	block.count = new(big.Int)
+	block.ipv6 = ip0.To4() == nil
+	if block.ipv6 {
+		block.firstIP.SetBytes(ip0.To16())
+		block.count.SetBytes(ip1.To16())
+	} else {
+		block.firstIP.SetBytes(ip0.To4())
+		block.count.SetBytes(ip1.To4())
+	}
+	block.count.Sub(block.count, block.firstIP)
```
na--, line 72: Please add a comment that ip0<ip1 is validated elsewhere, I missed the `negative ip range` check on the first reading

na-- (commented):
LGTM, besides the option name and the few minor nitpicks I noted inline.

lib/types/ipblock.go
```diff
@@ -0,0 +1,173 @@
...
+	// that will probably require more math ... including knowing which is the next network and ...
+	// thinking about it - it looks like it's going to be kind of hard or badly defined
+	i := new(big.Int)
+	i.Add(b.firstIP, index)
+	// TODO use big.Int.FillBytes when golang 1.14 is no longer supported
+	return net.IP(i.Bytes())
+}
+
+// NewIPPool return an IPBlock slice with corret weight and mode
+// Possible format is range1[:mode[:weight]][,range2[:mode[:weight]]]
```
imiric, line 116: Outdated comment? Also typos 😄

lib/types/ipblock.go
MStoykov: I am going to wait for @imiric to finish his review as well before starting. But do you mean that you want to make ipBlock and than instead of using it in hte Pool to get it's elements and copy it to ipPoolBlock which has `count` named something different but still with the same value that it has now?

I propose `countWithThisBlock`  or `countUntilThisBlock` which will mean I can drop the if in the `GetIPBig` function

lib/types/ipblock.go
```diff
@@ -0,0 +1,173 @@
+package types
```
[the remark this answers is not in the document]
MStoykov: https://github.com/golangci/golangci-lint/pull/1181 https://github.com/denis-tingajkin/go-header  ? 
cc @na--

lib/types/ipblock.go
na--: > get it's elements and copy it to ipPoolBlock which has count named something different but still with the same value that it has now
yes, exactly that

> I propose countWithThisBlock or countUntilThisBlock
:+1: for both (the second one with changing the implementation slightly, of course)

lib/types/ipblock.go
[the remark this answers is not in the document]
na--: hmm yeah, :+1: for enabling this check in golangci-lint, awesome find!

lib/types/ipblock.go
```diff
@@ -0,0 +1,173 @@
...
+		return ipBlockFromRange(s + "-" + s)
+	}
+}
+
+func ipBlockFromRange(s string) (*ipBlock, error) {
+	ss := strings.SplitN(s, "-", 2)
+	ip0Str, ip1Str := strings.TrimSpace(ss[0]), strings.TrimSpace(ss[1])
+	ip0, ip1 := net.ParseIP(ip0Str), net.ParseIP(ip1Str)
+	if ip0 == nil || ip1 == nil {
+		fmt.Println("Wrong IP range format: ", s)
```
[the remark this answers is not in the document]
na--: huh, completely missed that, but do we need this at all? seems like a forgotten debug statement, we're returning an error below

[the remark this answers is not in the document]
MStoykov: This will be deleted (already did it locally) - the original code didn't return errors it was printing them. 
I apparently missed some :facepalm:  `fmt.Println`

lib/types/ipblock.go
MStoykov: https://github.com/loadimpact/k6/pull/1682/commits/ebd5b470e182c2a868d6b91e0fa999aa12973b20

lib/types/ipblock.go
```diff
@@ -0,0 +1,201 @@
...
+)
+
+// ipBlock represents a continuous segment of IP addresses
+type ipBlock struct {
+	firstIP, count *big.Int
+	ipv6           bool
+}
+
+// ipPoolBlock is almost the same as ipBlock but is put in a IPPool and knows it's start id indest
+// of it's size/count
```
imiric, line 38: Please revise this comment. I'm not sure what `it's start id indest` means, and `it's` should be `its` in this case.

imiric (dismissed):
LGTM, just a comment nitpick.

lib/types/ipblock.go
MStoykov: "ipPoolBlock is similar to ipBlock but instead of knowing its count/size it knows the first index from which it starts in an IPPool" ?

lib/types/ipblock.go
```diff
@@ -0,0 +1,201 @@
...
+		block.count.SetBytes(ip1.To4())
+	}
+	block.count.Sub(block.count, block.firstIP)
+	block.count.Add(block.count, big.NewInt(1))
+
+	return &block
+}
+
+func ipBlockFromCIDR(s string) (*ipBlock, error) {
+	_, pnet, err := net.ParseCIDR(s)
```
divfor, line 102: _, pnet, err := net.ParseCIDR(strings.TrimSpace(s))

lib/types/ipblock.go
```diff
@@ -0,0 +1,201 @@
...
+			startIndex: new(big.Int).Set(pool.count), // this is how many there are until now
+		}
+		pool.count.Add(pool.count, r.count)
+	}
+
+	// The list gets reversed here as later it is searched based on when the index we are looking is
+	// bigger than startIndex but it will be true always for the first block which is with
+	// startIndex 0. This can also be fixed by iterating in reverse but this seems better
+	for i := 0; i < len(pool.list)/2; i++ {
+		pool.list[i], pool.list[len(pool.list)/2-i] = pool.list[len(pool.list)/2-i], pool.list[i]
```
divfor, line 161: a reverse should be:  pool.list[i], pool.list[len(pool.list)-i] = pool.list[len(pool.list)-i], pool.list[i]

js/runner.go
```diff
@@ -170,6 +170,10 @@ func (r *Runner) newVU(id int64, samplesOut chan<- stats.SampleContainer) (*VU,
 		BlockedHostnames: r.Bundle.Options.BlockedHostnames.Trie,
 		Hosts:            r.Bundle.Options.Hosts,
 	}
+	if r.Bundle.Options.LocalIPs.Valid {
+		dialer.Dialer.LocalAddr = &net.TCPAddr{IP: r.Bundle.Options.LocalIPs.Pool.GetIP(uint64(id - 1))}
```
divfor, line 174: From my testing, id will be 0 in the first VU allocation. so a check of id > 0 should be set.

divfor: if r.Bundle.Options.LocalIPs.Valid && id > 0 {

divfor: there is a fact: VU (id==0) will never sends requests... I think this is a bug

lib/types/ipblock.go
MStoykov: I don't think the trimming is needed - I just don't think people should add a random amount of additional space in their argument and then tools to just remove them for them. 

I am actually thinking if I shouldn't drop it from the places I have apparently not dropped it from :thinking: 
cc @na-- @imiric

divfor: at least it should keep as same as ipBlockFromRange() which is using trimming.

js/runner.go
MStoykov: VU with id 0 is generally used for stuff that "don't" do requests but ... `setup` and `teardown` functions can make them and they need to also use one of the provided IPs to make the requests. 

Actually speaking of this ... maybe they should use some "randomish" one but instead the first one? Currently They will use `(2**64 -1) % IPPool.count` which IMO is fine - they use an IP from the provided ones and arguably every IP should be viable, but maybe they should use the first one instead for more consistancy? 
cc @imiric @na--

lib/types/ipblock.go
imiric: I don't have a strong preference either way, but we should keep it consistent, so removing trimming LGTM.

js/runner.go
imiric: I don't think we should add any special handling for 0 ID VUs, but we should fix the `uint64` underflow for the `id=0` case. Picking the first one seems fine to me.

cmd/options.go
```diff
@@ -93,6 +93,7 @@ func optionFlagSet() *pflag.FlagSet {
 	flags.StringSlice("tag", nil, "add a `tag` to be applied to all samples, as `[name]=[value]`")
 	flags.String("console-output", "", "redirects the console logging to the provided output file")
 	flags.Bool("discard-response-bodies", false, "Read but don't process or save HTTP response bodies")
+	flags.String("local-ips", "", "Client IP Ranges and/or CIDRs from which each VU will be making requests")
```
imiric, line 96: Could we add some usage examples here? It would be helpful to showcase how ranges are specified, the comma delimiter, etc.

js/runner.go
```diff
@@ -170,6 +170,14 @@ func (r *Runner) newVU(id int64, samplesOut chan<- stats.SampleContainer) (*VU,
 		BlockedHostnames: r.Bundle.Options.BlockedHostnames.Trie,
 		Hosts:            r.Bundle.Options.Hosts,
 	}
+	if r.Bundle.Options.LocalIPs.Valid {
+		ipIndex := uint64(id - 1)
+		if id == 0 {
+			ipIndex = 1
+		}
```
na--, line 177: This seems incorrect :confused: If I'm reading it right, then we'd get the second IP for `setup()` and `teardown()`, not the first? And it's also a bit confusingly written, this seems simpler:
```suggestion
		var ipIndex uint64
		if id > 0 {
			ipIndex = uint64(id - 1)
		}
```

imiric (approved):
LGTM, just remember to squash.

lib/options.go
```diff
@@ -388,6 +388,9 @@ type Options struct {
 
 	// Redirect console logging to a file
 	ConsoleOutput null.String `json:"-" envconfig:"K6_CONSOLE_OUTPUT"`
+
+	// Specify client IP ranges and/or CIDR from which VUs will make requests
+	LocalIPs types.NullIPPool `json:"-" envconfig:"K6_LOCAL_IPS"`
```
na--, line 393: I am not sure why this shouldn't be configured via the JS `options` and from `config.json`. We can validate so we don't allow it in the cloud, but I don't see to disable it in general.

MStoykov: Why should it be configurable? I find it very unlikely that will be a needed thing for anyone and setting an environment variable won't be a good enough solution as well. 

I am for leaving this for when someone actually asks for it as we can always add it

cmd/options.go
divfor: from my view, take the first IP is not good since the vu id == 0 will take initialising requests - I guess it is DNS resolving?
if skipped, first VU(id == 0) will be using OS interface ip address (OS to decide what interface & its IP address to reach dns servers). this is needed because most of outside dns servers may not have route to responding to k6's --local-ips.

js/runner.go
divfor: from my view, take the first IP is not good since the vu id == 0 will take initialising requests - I guess it is DNS resolving?
if skipped, first VU(id == 0) will be using OS interface ip address (OS to decide what interface & its IP address to reach dns servers). this is needed because most of outside dns servers may not have route to responding to k6's --local-ips.

lib/options.go
na--: > Why should it be configurable?

Because there's no good reason not to be? And, whenever possible, most things in k6 are configurable in 4 different ways - CLI flags, env vars, JS options, JSON options. When that's not the case, it's usually because there is a good reason for it. For example, we need to know the option's value _before_ we execute the init context. Or we think the option is so minor that it'll overcrowd the CLI flags. Or there's a bug in `envconfig`, or something like that.

The more inconsistencies we have with the k6 options, the worse the user experience is. It's not a big deal, but constantly breaking expectations in this way is bound to get annoying.

na-- (approved):
LGTM besides the lack of js/json option support, but we can fix that in the future, along with all of the already existing config inconsistencies :disappointed: 

Regarding future developments in this direction (weights, randomization and such), I was thinking that shoehorning them in a super-complex global option might not be the best way to implement the feature. I had a brief look at the other PR and it was very difficult to understand things. Rather, given that this is essentially a per-VU setting, why can't we have a JS API to control the source IPs for each VU?

Basically, leave the `--local-ips` exactly as it is, without any further complications. Instead, in the future, we can add a new JS API that allow users to control different aspects of the VU, including the local IPs. I don't see a problem with setting the local IP in the init context, or even on every iteration... The `--local-ips` value can be used for validation - if set, allow the JS function to only select IPs from its pool, otherwise allow users to set any local IPs.

More thought obviously has to go into this, but I think it would be better than having increasingly more complex global options. "Pick a random IP out of this pool of IP ranges with different weights" is much more cleanly expressed in a programming language like JS than through some obscure configuration syntax. 

And we actually won't have to add this huge complexity to k6 for the few users like @divfor who might need it. We'd just offer a simple API for setting the source IP. Then every user can have an arbitrary complex scheme for picking random or sequential values however they like. The new HTTP client API should also probably have the ability to set a local IP... And this way, local IPs will work much better in conjunction with scenarios - different scenarios can have different pools of local IPs, which seems potentially useful.

PR #1688, opened 2020-10-26
xk6 plugin integration

### LINKED ISSUE #1353: Plugin support
As a user, I would like to be able to implement custom functionality and protocols and expose it in the JavaScript VM so that I can write the heavy lifting work in Golang but write tests in JS.

## Feature Description
As a protocol developer, I'd like to be able to develop a client that talks to my protocol via the JavaScript VM that k6 runs. To do this, I would prefer to be able to write a plugin that doesn't have to live in the main codebase and isn't the responsibility of the k6 team.

Effectively, a plugin should expose:

- A preflight function (to start any background services it might need)
- A postflight function (to shutdown any background services it might have started)
- A `map[string]interface{}` that gets added to the module mapping so that the plugin's functions can be imported from JavaScript.

## Suggested Solution (optional)
My MVP proposal would be to use Golang's [`plugin`](https://golang.org/pkg/plugin/) package. Despite not supporting Windows in its current state, Windows users could use Docker to run plugins. Moreso, it would not introduce any new dependencies to k6.

Plugins would be loaded by passing a `-plugin` argument when launching k6, which should receive a `.so` file path, and would be initialized (preflight) once, on test start, and cleaned up (postflight) once, on test finish.

A plugin struct could look something like:

```golang
type Plugin struct {
	Name       string
	Preflight  func() error
	Postflight func() error
    Modules    map[string]interface{}
}
```

For each plugin file, k9 would run something to the lines of:

```golang
// Omitting error checking for brevity
for _, path := range plugins {
	p, _ := plugin.Open(path)
	m, _ := p.Lookup("Plugin")
	meta := m.(Plugin)

	// Add plugin modules to JavaScript module registry
	// Add preflight and postflight functions to a callback list
}
```

Tagging @na--, @imiric, and @MStoykov as indicated in Slack.

imiric:
This is a proposal for an external plugin system for k6, inspired by [xcaddy](https://github.com/caddyserver/xcaddy). See also the [xk6 repo](https://github.com/k6io/xk6) and the [SQL plugin](https://github.com/imiric/k6-plugin-sql) adapted from @mostafa's Go plugin version.

Initially plugins are only meant for external JS modules, but support for other plugin types (output, etc.) is likely to be added in the future.

The module support in (x)caddy is much broader in scope than what's currently in this PoC. The main distinction is that Caddy modules can recursively load other modules, and only the "guest"/"child" modules are referred to as "plugins". See [the documentation](https://caddyserver.com/docs/extending-caddy).

## Usage

To build a `k6` binary with the SQL plugin, from the `xk6` repo @ `11baed0` run:

```sh
CGO_ENABLED=1 go run ./cmd/xk6/main.go build feat/1353-xk6-plugins \
  --with github.com/imiric/k6-plugin-sql
```

`CGO_ENABLED=1` is required for the SQL dependencies.

With the following `script.js`:
```javascript
import { check } from 'k6';
import sql from 'k6/x/sql';

export default function () {
  let db = sql.open("sqlite3", "./test.sqlite");
  db.exec(`CREATE TABLE IF NOT EXISTS keyvalues (
        id integer PRIMARY KEY AUTOINCREMENT,
        key varchar NOT NULL,
        value varchar);`);
  db.exec("INSERT INTO keyvalues (key, value) VALUES('plugin-name', 'k6-plugin-sql');");
  let results = sql.query(db, "SELECT * FROM keyvalues;");
  for (const row of results) {
    console.log(`key: ${row.key}, value: ${row.value}`);
  }
  db.close();
}
```

Running `./k6 run script.js` should create a `test.sqlite` DB and show `INFO[0000] key: plugin-name, value: k6-plugin-sql        source=console`.

Closes #1353


modules/modules.go
```diff
@@ -0,0 +1,279 @@
...
+}
+
+// CleanerUpper is implemented by modules which may have side-effects
+// such as opened files, spawned goroutines, or allocated some sort
+// of non-stack state when they were provisioned. This method should
+// deallocate/cleanup those resources to prevent memory leaks. Cleanup
+// should be fast and efficient. Cleanup should work even if Provision
+// returns an error, to allow cleaning up from partial provisionings.
+type CleanerUpper interface {
+	Cleanup() error
```
imiric, line 278: These interfaces are unused right now, but they seem generally useful to have. Should they be hooked up?

modules/modules.go
```diff
@@ -0,0 +1,279 @@
...
+	Validate() error
+}
+
+// CleanerUpper is implemented by modules which may have side-effects
+// such as opened files, spawned goroutines, or allocated some sort
+// of non-stack state when they were provisioned. This method should
+// deallocate/cleanup those resources to prevent memory leaks. Cleanup
+// should be fast and efficient. Cleanup should work even if Provision
+// returns an error, to allow cleaning up from partial provisionings.
+type CleanerUpper interface {
```
MStoykov, line 277: Why `CleanerUpper` instead of `Cleanupper`?

modules/modules.go
```diff
@@ -0,0 +1,279 @@
+// Copyright 2015 Matthew Holt and The Caddy Authors
```
imiric, line 1: What is the appropriate way of handling these headers? This file is a stripped down version of [the original](https://github.com/caddyserver/caddy/blob/b6e96d6f4a55f96ccbb69f112822f0a923942246/modules.go), so do we add the k6 header and leave this one below it, or can we re-license it?

modules/modules.go
imiric: That's the original name 😄  `CleanerUpper` sounds slightly better to me.

MStoykov (commented):
From my initial look at this I think ... that 95% of this is not needed.
I was more thinking of just adding a function that changes the current module map and use that in the plugins. 
I would even argue that it shouldn't check (at least for now) that someone tries to add "totally/not/valid/path" as ... well that is their problem ;) 

This whole thing looks like it tries to be way too complex for what we want from the current js plugins for no added benefit but more code that the "plugins" need to write and it is slightly harder for them to just be copied in k6 once we decided they should be part of it.

Additionally I am pretty sure the [sql module has a k6Module](https://github.com/imiric/k6-plugin-sql/blob/master/sql.go#L23) function in js with this code, which obviously can be fixed in the bridge or through some other means ... but IMO is again .. not needed

imiric:
Thanks @MStoykov. This is already a cut-down version of Caddy plugins, but I left what's currently here as open design questions since we didn't really spec out the feature past "do what xcaddy does" :smile:

You're right that Caddy modules/plugins are much broader in scope than what we need, but I don't think there's much complexity here, it's mostly just namespace handling and getters.

OK, so to summarize:

- Get rid of namespaces, we don't need them? I still feel they're a good idea if we expect this ecosystem to grow, and adding them in later after some plugins are written will be more work than having them in place since v1.

- Get rid of the `Module` interface and requirement for plugins to implement it. Plugins will simply be added to the `js.modules.Index` map in `RegisterModule()`. I totally missed that `k6Module` would be exposed in JS, thanks.


On the topic of naming, should we abandon the "module" naming used by Caddy and use "plugin" instead? Caddy's naming is a bit confusing, but given the additional scope "modules" makes more sense in their case, whereas we could be fine with referring to them as "plugins" altogether.

na--:
Diclaimer: this is a very preliminary comment, since I have only skimmed this PR and taken a glance at xk6. But, from what I can see, I will probably agree with @MStoykov that this is unnecessarily complex.

Take for example the `Provisioner` interface. It probably makes total sense in the context of caddy, but in k6, do we really need it, when users can call arbitrary plugin functions from JS? If a plugin needs some setup, can't it be handled manually, like this:
```js
import plugin from 'k6/x/awesome-plugin';

plugin.provision() // maybe call this in setup() even?
```

Instead of `Validate()`, what is stopping the plugin from doing validation when its module is loaded in the script? And while I can see the rationale behind `CleanerUpper`, again, for the first iteration why can't that be handled as a method call in `teardown()` or something like that.

I think it's still early to tell exactly how the final k6 plugin API should look like, so the more minimal it is right now, the fewer breaking changes we'll have to make in the future. All of these interfaces are optional extra features that we can add in the future, in a backwards-compatible way, right? So we should probably skip them, for now, and add them properly when we see a clear need for them.

> Adopting namespaces would also be useful to distinguish between plugin types once we add support for output and other types of plugins besides JS ones.

Not sure if that is the best way to distinguish between plugin types. Wouldn't interfaces be better? Seems like a much cleaner and type-safe alternative than relying on stringy checks

js/internal/modules/modules.go
```diff
@@ -0,0 +1,52 @@
...
+func GetModule(name string) interface{} {
+	mx.RLock()
+	defer mx.RUnlock()
+	return modules[name]
+}
+
+// RegisterModule registers the given mod as a JavaScript module, available
+// for import from JS scripts by name.
+// This function panics if a module with the same name is already registered.
+func RegisterModule(name string, mod interface{}) {
```
imiric, line 44: Should we avoid the `modules.RegisterModule` stutter and shorten `RegisterModule` to `Register` and `GetModule` to `Get`?

MStoykov: yes

MStoykov:
am I correct in thinking that the move of all the current modules in `internal` was not needed? If I remember correctly the access to internal is for any packages that are the "parent" of the internal and all of its other "children"?

I am not particularly strongly opinionated on whether or not they should be in `internal` - I doubt if anyone will need them, and I doubt that we will need to use them from somewhere else, and they are quite tightly coupled with goja, so reusage for some other project is also unlikely. On the other hand, this makes this change bigger and out of what IMO is needed. 
cc @na--

js/internal/modules/modules.go
imiric: Resolved by 9747b1e2.

imiric:
@MStoykov You're correct, moving all modules to `internal` wasn't necessary, we just need the module map there. I fixed this in the latest push.

js/modules/k6/html/elements_gen.go
```diff
@@ -1,4 +1,4 @@
-// generated by js/modules/k6/html/gen/gen_elements.go directed by js/modules/k6/html/elements.go;  DO NOT EDIT
+// generated by js/internal/modules/k6/html/gen/gen_elements.go directed by js/internal/modules/k6/html/elements.go;  DO NOT EDIT
```
MStoykov, line 1: I think this is now wrong

js/modules/k6/html/gen/gen_elements.go
```diff
@@ -372,7 +372,8 @@ func main() {
 var elemFuncsTemplate = template.Must(template.New("").Funcs(template.FuncMap{
 	"buildStruct": buildStruct,
 	"returnType":  returnType,
-}).Parse(`// generated by js/modules/k6/html/gen/gen_elements.go directed by js/modules/k6/html/elements.go;  DO NOT EDIT
+}).Parse("// generated by js/internal/modules/k6/html/gen/gen_elements.go " +
+	`directed by js/internal/modules/k6/html/elements.go;  DO NOT EDIT
```
MStoykov, line 376: Also now wrong

js/internal/modules/modules.go
```diff
@@ -0,0 +1,52 @@
...
+
+import (
+	"fmt"
+	"sync"
+)
+
+//nolint:gochecknoglobals
+var (
+	modules = make(map[string]interface{})
+	mx      sync.RWMutex
```
MStoykov, line 31: nit: I am not certain having a Lock is truly necessary as the Register happens in init functions and I would argue should not happen at any other time. And the inits are by definition not concurrent so :shrug: 

Don't get me wrong ... it is probably better to have lock in the case that we ever decide to have Register outside the `init` functons. And I doubt the `RLock` significantly slows anything

js/modules/modules.go
```diff
@@ -0,0 +1,46 @@
...
+	return modules.Get(name)
+}
+
+// Register the given mod as an external JavaScript module,
+// available for import from JS scripts with the "k6/x/<name>" import path.
+// This function panics if a module with the same name is already registered.
+func Register(name string, mod interface{}) {
+	if !strings.HasPrefix(name, extPrefix) {
+		name = fmt.Sprintf("%s%s", extPrefix, name)
+	}
```
MStoykov, line 43: I kind of wonder if we shouldn't just either:
1. always prefix it 
2. panic if it isn't prefixed instead of prefixing it.

My reasoning is that if someone writes `Register("k7/x/something")` and then goes to use their module they will be really surprised that it totally doesn't work. While if it panics or the documentation specifically says taht we always prepend `k6/x/` that will be much better IMO

MStoykov (changes requested):
LGTM, but drop the `html` changes that were leftover and .. maybe squash this now in one commit?

js/modules/k6/html/elements_gen.go
imiric: 🤦  Thanks, I missed this. It's fixed in the latest push.

js/modules/k6/html/gen/gen_elements.go
imiric: Fixed in latest push.

js/internal/modules/modules.go
imiric: Good point, I wasn't aware `init()`s weren't concurrent.

I don't feel strongly about it to remove it, but if it's YAGNI then we might as well do so.

@na-- What do you think?

js/modules/modules.go
imiric: Hhmm well the function comment does explain that imports will be in the form `k6/x/<name>`, and this gives the option to register with or without it. I can see how panicking would be helpful though, but not sure if we should force users to specify the prefix. Kind of torn on this one...

mostafa: I think it would be better to just ask for the name and always prepend the prefix ourselves. We can simply mention this approach in the docs and our reasoning would be that it is the intended behavior. Then there's no need for panic.

simskij: Automatically adding prefixes will likely just lead to more confusion.

I'm thinking that manual import path config + Panics would be preferable as it:

1. The error tells the user how to solve it.
2. No hidden behavior you need to know or document. 
3. Import path in the JS file will match exactly what they have written in their plugin.

imiric: OK, @simskij convinced me, `panic` it is. 😄

Will push the change in a bit.

MStoykov: as before I don't think categories are needed for k6. I find the "we have different module registers and functions for different categories" approach much better and easier to document and easy for the people who want add 1 type of plugin to not have to read how it's done for the others.

We can always add a more central plugin system and wire the old calls to it and recommend that people use the new system ... although again ... I highly doubt this day will come

js/internal/modules/modules.go
na--: I don't think it matters much - it shouldn't be needed, but it won't slow down anything either, so :man_shrugging:

na-- (commented):
Hmm why do we need `internal/modules` at all? From what I can see, there's nothing stopping a JS plugin from directly calling `internal/modules.Register("k6/non-x-path")`, right?

Wouldn't it be both simpler _and_ more secure to have something like this in the old `js/modules/index.go` file?
```go

var modules = map[string]interface{}{
	"k6":                k6.New(),
	"k6/crypto":         crypto.New(),
	"k6/crypto/x509":    x509.New(),
	"k6/encoding":       encoding.New(),
	"k6/http":           http.New(),
	"k6/metrics":        metrics.New(),
	"k6/html":           html.New(),
	"k6/ws":             ws.New(),
	"k6/protocols/grpc": grpc.New(),
}
const extPrefix string = "k6/x/"

func Get(name string) interface{} {
	return modules[name]
}

func Register(name string, mod interface{}) {
	if !strings.HasPrefix(name, extPrefix) {
		panic(fmt.Errorf("external module names must be prefixed with '%s', tried to register: %s", extPrefix, name))
	}

	modules[name] = mod
}
```

The built-in k6 modules would be an immutable static map and and you can only add modules starting with `k6/x`, while you can get any module?

na-- (approved):
I didn't know about [this](https://github.com/golang/go/wiki/Modules#can-a-module-depend-on-an-internal-in-another)... :blush: Live and learn, I guess - LGTM!

MStoykov (approved):

PR #1689, opened 2020-10-26
GitHub container registry

salem84:
This PR is linked to #1622 issue and implements Github package publishing in the new workflow.

As described in [Github Docs](https://docs.github.com/en/free-pro-team@latest/packages/getting-started-with-github-container-registry/migrating-to-github-container-registry-for-docker-images) Github Packages will be replaced with Github Container Registry, so I have used it.

In order to launch pipelines correctly CR_PAT secret must be created and it contains a Personal Access Token with `write;packages` scope.

More details can be found on this [link](https://docs.github.com/en/free-pro-team@latest/packages/getting-started-with-github-container-registry/migrating-to-github-container-registry-for-docker-images#authenticating-with-the-container-registry).

I have created tasks similar to Docker publishing, with the same dependencies, in order to execute them in parallel

### DIFF
--- a/.github/workflows/all.yml
+++ b/.github/workflows/all.yml
@@ -136,7 +136,6 @@ jobs:
     runs-on: ubuntu-latest
     outputs:
       version: ${{ steps.get_version.outputs.version }}
-      version_ci: ${{ steps.get_version.outputs.version_ci }}
     steps:
       - name: Get version
         id: get_version
@@ -208,8 +207,6 @@ jobs:
   build-docker:
     runs-on: ubuntu-latest
     needs: [deps, lint, test, test-cov, configure]
-    env:
-      VERSION_CI: ${{ needs.configure.outputs.version_ci }}
     steps:
       - name: Checkout code
         uses: actions/checkout@v2
--- a/.github/workflows/all.yml
+++ b/.github/workflows/all.yml
@@ -135,7 +135,8 @@ jobs:
   configure:
     runs-on: ubuntu-latest
     outputs:
-        version: ${{ steps.get_version.outputs.version }}
+      version: ${{ steps.get_version.outputs.version }}
+      version_ci: ${{ steps.get_version.outputs.version_ci }}
     steps:
       - name: Get version
         id: get_version
@@ -143,6 +144,9 @@ jobs:
           VERSION="${GITHUB_REF##*/}"
           echo "VERSION=${VERSION}"
           echo "::set-output name=version::${VERSION}"
+          VERSION_CI="${VERSION#v}-${GITHUB_SHA::7}"
+          echo "VERSION_CI=${VERSION_CI}"
+          echo "::set-output name=version_ci::${VERSION_CI}"
 
   build:
     runs-on: ubuntu-latest
@@ -204,7 +208,74 @@ jobs:
           path: dist/
           retention-days: 7
 
+  build-docker:
+    runs-on: ubuntu-latest
+    needs: [deps, lint, test, test-cov, configure]
+    env:
+      VERSION_CI: ${{ needs.configure.outputs.version_ci }}
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: Build
+        run: docker build -t $DOCKER_IMAGE_ID .
+      - name: Check
+        run: |
+            docker run $DOCKER_IMAGE_ID version
+            docker run $DOCKER_IMAGE_ID --help
+            docker run $DOCKER_IMAGE_ID help
+            docker run $DOCKER_IMAGE_ID run --help
+            docker run $DOCKER_IMAGE_ID inspect --help
+            docker run $DOCKER_IMAGE_ID status --help
+            docker run $DOCKER_IMAGE_ID stats --help
+            docker run $DOCKER_IMAGE_ID scale --help
+            docker run $DOCKER_IMAGE_ID pause --help
+            docker run $DOCKER_IMAGE_ID resume --help
+      - name: Publish Docker CI
+        run: |
+          echo "DOCKER_IMAGE_ID=$DOCKER_IMAGE_ID"
+          # Log into GHCR
+          echo "${{ secrets.CR_PAT }}" | docker login https://ghcr.io -u ${{ github.actor }} --password-stdin
+          docker tag "$DOCKER_IMAGE_ID" "ghcr.io/$DOCKER_IMAGE_ID/ci:$VERSION_CI"
+          docker push "ghcr.io/$DOCKER_IMAGE_ID/ci:$VERSION_CI"
+
   publish-docker:
+    runs-on: ubuntu-latest
+    needs: [build-docker, configure]
+    if: github.event_name != 'pull_request'
+    env:
+      VERSION: ${{ needs.configure.outputs.version }}
+      VERSION_CI: ${{ needs.configure.outputs.version_ci }}
+    steps:
+      - name: Publish Container Registries
+        run: |
+          echo 'Publish Docker'
+          skopeo copy --src-creds ${{ github.actor }}:${{ secrets.CR_PAT }} \
+                      --dest-creds ${{ secrets.DOCKER_USER }}:${{ secrets.DOCKER_PASS }} \
+                      docker://ghcr.io/$DOCKER_IMAGE_ID/ci:$VERSION_CI \
+                      docker://$DOCKER_IMAGE_ID:$VERSION
+
+          echo 'Publish GHCR'
+          skopeo copy --src-creds ${{ github.actor }}:${{ secrets.CR_PAT }} \
+                      --dest-creds ${{ github.actor }}:${{ secrets.CR_PAT }} \
+                      docker://ghcr.io/$DOCKER_IMAGE_ID/ci:$VERSION_CI \
+                      docker://ghcr.io/$DOCKER_IMAGE_ID:$VERSION
+
+          # We also want to tag the latest stable version as latest
+          if [[ "$VERSION" != "master" ]] && [[ ! "$VERSION" =~ (RC|rc) ]]; then
+            echo 'Publish Docker (latest)'
+            skopeo copy --src-creds ${{ github.actor }}:${{ secrets.CR_PAT }} \
+                        --dest-creds ${{ secrets.DOCKER_USER }}:${{ secrets.DOCKER_PASS }} \
+                        docker://ghcr.io/$DOCKER_IMAGE_ID/ci:$VERSION_CI \
+                        docker://$DOCKER_IMAGE_ID:latest
+
+            echo 'Publish GHCR (latest)'
+            skopeo copy --src-creds ${{ github.actor }}:${{ secrets.CR_PAT }} \
+                        --dest-creds ${{ github.actor }}:${{ secrets.CR_PAT }} \
+                        docker://ghcr.io/$DOCKER_IMAGE_ID/ci:$VERSION_CI \
+                        docker://ghcr.io/$DOCKER_IMAGE_ID:latest
+          fi
+
+  publish-ghcr:
     runs-on: ubuntu-latest
     needs: [deps, lint, test, test-cov, configure]
     env:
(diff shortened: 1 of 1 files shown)


na--:
Thanks for this, however, what is the benefit of building the docker image again? Is there a reason we shouldn't just push the exact same docker image to both the Docker hub and the GitHub container registry?

salem84:
> Thanks for this, however, what is the benefit of building the docker image again? Is there a reason we shouldn't just push the exact same docker image to both the Docker hub and the GitHub container registry?

I was undecided if it was better to reuse it or not.
I rebuilt the image because in this way maybe simpler to keep separated and potentially disable on of this.
Initially I would like splitting all in three jobs: the first build the image without tags, then the second and third step (dependent on first) publish respectively to docker registry and github packages, in order to reuse the same build and parallelize works, but I think it doesn't work.
However I could do a similar process by using only one job composed of three steps...
What do you think about it? In this way pipeline is optimized and potentially we can use variables and conditions to exclude some steps

imiric:
@salem84 I like the idea of splitting the job into `docker-build` with the current `Check` step, and `docker-publish` that uploads the same image to both Docker Hub and ghcr.io. The `docker-build` job should run on all PR commits and merges so that we can confirm image building still works, while `docker-publish` should run only on version tags or `master` commits.

Since uploading to ghcr.io should be faster than to Docker Hub, maybe we could use it as intermediary storage between jobs, otherwise you'd have to move the image manually as an artifact or with some caching action. Thanks for your help with this!

salem84:
@imiric you are welcome, it's a pleasure contribute to k6 project!
Definitely we need an intermediate storage between jobs, so a solution could be:
```
jobs:
   build-docker:
      - build
      - check
      - tag & publish to ghcr.io (maybe useful publish also on PR for testing purpose? or keep same criteria as docker hub? )

  publish-dockerhub
     - pull from ghcr.io (as intermediate storage to share between jobs)
     - tag & publish to docker hub (only on tags / master)
```
Did I get it right your idea?

imiric:
@salem84 Yeah, something like that should work. Though we should use the same publishing criteria for both Docker Hub and GHCR (only on version tags and `master` commits), so the intermediary pushing should use a temporary tag or some private storage that remains hidden from the public releases.

Consider using [skopeo](https://github.com/containers/skopeo) for copying between registries if it makes things easier, though we might just get by with plain Docker.

salem84:
> @salem84 Yeah, something like that should work. Though we should use the same publishing criteria for both Docker Hub and GHCR (only on version tags and `master` commits), so the intermediary pushing should use a temporary tag or some private storage that remains hidden from the public releases.
Ah ok, it's correct! I will analyse this scenario 

> Consider using [skopeo](https://github.com/containers/skopeo) for copying between registries if it makes things easier, though we might just get by with plain Docker.
I'll surely take a look to _skopeo_ ! It's new for me :-)

salem84:
Hi! I have splitted build and publish jobs, as we planned.
In particular, in build job it is used GHCR as "temporary" storage to publish a docker image with a particular naming convention and in publish job image is copied to Docker Hub and GHCR as final image

.github/workflows/all.yml
```diff
@@ -135,14 +135,18 @@ jobs:
   configure:
     runs-on: ubuntu-latest
     outputs:
-        version: ${{ steps.get_version.outputs.version }}
+      version: ${{ steps.get_version.outputs.version }}
+      version_ci: ${{ steps.get_version.outputs.version_ci }}
```
salem84, line 139: VERSION_CI it's used for the temporary image in other jobs

.github/workflows/all.yml
```diff
@@ -204,21 +208,81 @@ jobs:
...
-          echo "REF=${{ github.ref }}"
+          VERSION_CI="${VERSION_CI#v}"
           echo "DOCKER_IMAGE_ID=$DOCKER_IMAGE_ID"
-          # Log into registry
-          echo "${{ secrets.DOCKER_PASS }}" | docker login -u "${{ secrets.DOCKER_USER }}" --password-stdin
+          echo "VERSION_CI=$VERSION_CI"
+          # Log into GHCR
+          echo "${{ secrets.CR_PAT }}" | docker login https://ghcr.io -u ${{ github.actor }} --password-stdin
+          docker tag "$DOCKER_IMAGE_ID" "ghcr.io/$DOCKER_IMAGE_ID/ci:$VERSION_CI"
+          docker push "ghcr.io/$DOCKER_IMAGE_ID/ci:$VERSION_CI"
```
salem84, line 219: In this step image is published separated from "official" images and with VERSION_CI created in configure step

.github/workflows/all.yml
```diff
@@ -135,14 +135,18 @@ jobs:
...
+      version: ${{ steps.get_version.outputs.version }}
+      version_ci: ${{ steps.get_version.outputs.version_ci }}
     steps:
       - name: Get version
         id: get_version
         run: |
           VERSION="${GITHUB_REF##*/}"
           echo "VERSION=${VERSION}"
           echo "::set-output name=version::${VERSION}"
+          VERSION_CI="${VERSION#v}-${GITHUB_SHA::7}"
```
salem84, line 147: VERSION_CI is composed appending short SHA commit id to standard version

.github/workflows/all.yml
```diff
@@ -204,21 +208,45 @@ jobs:
...
-          echo "REF=${{ github.ref }}"
+          VERSION_CI="${VERSION_CI#v}"
           echo "DOCKER_IMAGE_ID=$DOCKER_IMAGE_ID"
-          # Log into registry
-          echo "${{ secrets.DOCKER_PASS }}" | docker login -u "${{ secrets.DOCKER_USER }}" --password-stdin
+          echo "VERSION_CI=$VERSION_CI"
+          # Log into GHCR
+          echo "${{ secrets.CR_PAT }}" | docker login https://ghcr.io -u ${{ github.actor }} --password-stdin
+          docker tag "$DOCKER_IMAGE_ID" "ghcr.io/$DOCKER_IMAGE_ID/ci:$VERSION_CI"
+          docker push "ghcr.io/$DOCKER_IMAGE_ID/ci:$VERSION_CI"
```
salem84, line 219: In this step image is published separated from "official" images and with VERSION_CI created in configure step

imiric: I like using the `/ci` suffix for this, but do we need to push the image if it's not published below? I can see it being useful for testing, but not sure if we need this. @na-- @MStoykov WYT?

.github/workflows/all.yml
```diff
@@ -204,21 +208,45 @@ jobs:
...
+            skopeo copy --src-creds ${{ github.actor }}:${{ secrets.CR_PAT }} --dest-creds ${{ secrets.DOCKER_USER }}:${{ secrets.DOCKER_PASS }} docker://ghcr.io/$DOCKER_IMAGE_ID/ci:$VERSION_CI docker://$DOCKER_IMAGE_ID:latest
+          fi
+      - name: Publish GHCR
+        run: |
+          skopeo copy --src-creds ${{ github.actor }}:${{ secrets.CR_PAT }} --dest-creds ${{ github.actor }}:${{ secrets.CR_PAT }} docker://ghcr.io/$DOCKER_IMAGE_ID/ci:$VERSION_CI docker://ghcr.io/$DOCKER_IMAGE_ID:$VERSION
           # We also want to tag the latest stable version as latest
           if [[ "$VERSION" != "master" ]] && [[ ! "$VERSION" =~ (RC|rc) ]]; then
-            docker tag "$DOCKER_IMAGE_ID" "$DOCKER_IMAGE_ID:latest"
-            docker push "$DOCKER_IMAGE_ID:latest"
+            skopeo copy --src-creds ${{ github.actor }}:${{ secrets.CR_PAT }} --dest-creds ${{ github.actor }}:${{ secrets.CR_PAT }} docker://ghcr.io/$DOCKER_IMAGE_ID/ci:$VERSION_CI docker://ghcr.io/$DOCKER_IMAGE_ID:latest
```
imiric, line 249: Should we join both publishing steps into one, to not repeat the `master`/`RC` logic?

imiric (commented):
Thanks @salem84, this looks good! I left a few questions and nitpicks.

A couple other things:
- Can you rebase on current `master`? It should fix that failing test.
- This works correctly in your fork, right? I noticed that the [`Publish Docker` step failed with 401](https://github.com/salem84/k6/runs/1355565785), but this might be related to the secrets in your repo.
  We'll test this on our side, but just to confirm that it worked as expected for you.

salem84:
> * Can you rebase on current `master`? It should fix that failing test.
Sure

> * This works correctly in your fork, right? I noticed that the [`Publish Docker` step failed with 401](https://github.com/salem84/k6/runs/1355565785), but this might be related to the secrets in your repo.
>   We'll test this on our side, but just to confirm that it worked as expected for you.
Yes it is a problem with docker hub credentials... I have configured my credential on GH secrets, but it obviously receive an error because it is unauthorized with my credentials to publish on `loadimpact/k6` (it's a variable taken from GH repository name)... just to check yesterday I tested locally skopeo and it works... however a test with official credential is needed

.github/workflows/all.yml
```diff
@@ -204,21 +208,45 @@ jobs:
...
-      - name: Publish
-        if: github.event_name != 'pull_request'
+      - name: Publish Docker CI
         run: |
-          echo "REF=${{ github.ref }}"
+          VERSION_CI="${VERSION_CI#v}"
           echo "DOCKER_IMAGE_ID=$DOCKER_IMAGE_ID"
-          # Log into registry
-          echo "${{ secrets.DOCKER_PASS }}" | docker login -u "${{ secrets.DOCKER_USER }}" --password-stdin
+          echo "VERSION_CI=$VERSION_CI"
```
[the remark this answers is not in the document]
salem84: Yes it's redundant

.github/workflows/all.yml
```diff
@@ -204,21 +208,45 @@ jobs:
...
+          # Log into GHCR
+          echo "${{ secrets.CR_PAT }}" | docker login https://ghcr.io -u ${{ github.actor }} --password-stdin
```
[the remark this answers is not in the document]
salem84: Yes with `write:packages` scope. I have used this procedure ([link](https://docs.github.com/en/free-pro-team@latest/packages/getting-started-with-github-container-registry/migrating-to-github-container-registry-for-docker-images#authenticating-with-the-container-registry))
When GHCR will go out from beta, we probably will use Github standard token

.github/workflows/all.yml
```diff
@@ -204,21 +208,45 @@ jobs:
...
+          VERSION_CI="${VERSION_CI#v}"
+          echo "REF=${{ github.ref }}"
+          echo "DOCKER_IMAGE_ID=$DOCKER_IMAGE_ID"
           echo "VERSION=$VERSION"
-          docker tag "$DOCKER_IMAGE_ID" "$DOCKER_IMAGE_ID:$VERSION"
-          docker push "$DOCKER_IMAGE_ID:$VERSION"
+          echo "VERSION_CI=$VERSION_CI"
+      - name: Publish Docker
+        run: |
+          skopeo copy --src-creds ${{ github.actor }}:${{ secrets.CR_PAT }} --dest-creds ${{ secrets.DOCKER_USER }}:${{ secrets.DOCKER_PASS }} docker://ghcr.io/$DOCKER_IMAGE_ID/ci:$VERSION_CI docker://$DOCKER_IMAGE_ID:$VERSION
```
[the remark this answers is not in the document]
salem84: Yes it's included! 
Ok for breaks

.github/workflows/all.yml
salem84: I was uncertain until publishing PR.... I agree with you and will merge it !

imiric (approved):
Thanks for the changes, this LGTM!

Unfortunately, that OCSP test is still broken and should be fixed by a383db48, which will be merged in #1712, but if you don't want to wait you can also cherry-pick the fix here. Sorry for the inconvenience.

.github/workflows/all.yml
na--: Yeah, I don't think we should push _all_ images to github's container registry :confused: I don't see a benefit, it's just going to add noise and take up space...

While it's nice that we are going to build the images for every PR and run a brief sanity check, we should only push the actual images for `master` builds and version tags.

na-- (changes requested):

.github/workflows/all.yml
salem84: Now for each build triggered by:

```
on:
  # Enable manually triggering this workflow via the API or web UI
  workflow_dispatch:
  push:
    branches:
      - master
    tags:
      - v*
  pull_request:
```

 it's pushed on GHCR `/ci`

salem84:
Sorry for the late... I have rebased PR with latest master version.
However build doesn't work because of CR_PAT missing... Moreover I think it's better replace `github.actor` with a variable (==> GH USERNAME) otherwise I think it doesn't work in PR builds

.github/workflows/all.yml
imiric: @salem84 Thanks for the recent changes, but it doesn't look like this issue was addressed. That is, we want to avoid pushing to GHCR on every PR, so this workflow should be triggered only on `master` commits or version tags.

I think just keeping the `if: github.event_name != 'pull_request'` condition on the "Publish Docker CI" step would do it, but maybe @na-- has a better suggestion.

imiric:
Thanks @salem84! I added the `CR_PAT` secret to the k6 repo, but it's created from my personal account, so not sure if it works with this repo. Hopefully we can use `GITHUB_TOKEN` for this soon.

Not pushing images on PRs should be adressed before we merge this.

imiric:
Hi @salem84, are you still available to finish this? Otherwise let us know and we can pick it up if you don't mind.

salem84:
> Hi @salem84, are you still available to finish this? Otherwise let us know and we can pick it up if you don't mind.

Hi @na-- , yes I'm available to finish the PR, but I don't remember what changes must be done... could we do a recap?

na--:
Besides rebasing the PR on the latest k6 `master`, we'd prefer the following changes:
- instead of using `skopeo copy` to copy the docker images between hosts, just push the image to both the docker and github container registries in the same step
- only push the built docker images on `master` commits and version tags, not on every random PR commits

salem84:
> Besides rebasing the PR on the latest k6 `master`, we'd prefer the following changes:
> 
> * instead of using `skopeo copy` to copy the docker images between hosts, just push the image to both the docker and github container registries in the same step
> * only push the built docker images on `master` commits and version tags, not on every random PR commits

Hi @na-- , I must test the process as a whole, however I have done following changes:

- Build and publish are in the same jobs, managed as different steps
- Publish step has a condition on branch:

```
- name: Publish
        if: |
          github.ref == 'refs/heads/master' ||
          startsWith(github.ref, 'refs/tags/v')
```
I suppose we can change this condition with 
```
if: github.event_name != 'pull_request'
```
because pipeline is triggered on master, v tags and PR, it's correct?

- This step publishes to Docker Hub and GithubContainer Registry with the condition for "latest" image version

Tomorrow I will test it, however if I missed something, please feel free to contact me

salem84:
> Besides rebasing the PR on the latest k6 `master`, we'd prefer the following changes:
> 
> * instead of using `skopeo copy` to copy the docker images between hosts, just push the image to both the docker and github container registries in the same step
> * only push the built docker images on `master` commits and version tags, not on every random PR commits

Hi @na-- , I must test the process as a whole, however I have done following changes:

- Build and publish are in the same jobs, managed as different steps
- Publish step has a condition on branch:

```
- name: Publish
        if: |
          github.ref == 'refs/heads/master' ||
          startsWith(github.ref, 'refs/tags/v')
```
I suppose we can change this condition with 
```
if: github.event_name != 'pull_request'
```
because pipeline is triggered on master, v tags and PR, it's correct?

- This step publishes to Docker Hub and GithubContainer Registry with the condition for "latest" image version

Tomorrow I will test it, however if I missed something, please feel free to contact me

salem84:
I have fixed an error and I have tested on my fork.
I have only an issue on DockerHub publish because I have different name between DockerHub and Github username, but this pipeline section is the same as now in main branch

.github/workflows/all.yml
```diff
@@ -226,19 +226,34 @@ jobs:
...
+            docker tag "$DOCKER_IMAGE_ID" "ghcr.io/$DOCKER_IMAGE_ID:latest"
+            docker push "ghcr.io/$DOCKER_IMAGE_ID:latest"
+          fi
+
+          echo "Publish Docker ($DOCKER_IMAGE_ID:$VERSION)"
+
+          # Log into Docker Hub Registry
           echo "${{ secrets.DOCKER_PASS }}" | docker login -u "${{ secrets.DOCKER_USER }}" --password-stdin
-          VERSION="${VERSION#v}"
-          echo "VERSION=$VERSION"
```
imiric, line 237: I think this is still needed for both GHCR and Docker Hub, as it strips the `v` prefix from the version tag, which is how we've been tagging images so far.

imiric (approved):
Thanks for the changes @salem84! This LGTM.

Since we'll have to test this with our internal secrets, I'm OK with merging it and testing it before the next release.

na-- (approved):
Sorry for the huge delay here, I forgot this for v0.31.0 :disappointed:

PR #1690, opened 2020-10-27
Print errors stopping the execution to the stderr

MStoykov:
This was broken after the adding of the loki logger output and should be
better fixed, but a better fix will required either rewrite of the cmd
package or more global variables, so this is preferable for now.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -119,7 +119,13 @@ var RootCmd = &cobra.Command{
 func Execute() {
 	if err := RootCmd.Execute(); err != nil {
 		code := -1
-		var logger logrus.FieldLogger = logrus.StandardLogger()
+
+		var logger logrus.FieldLogger = &logrus.Logger{
+			Out:       os.Stderr,
+			Formatter: new(logrus.TextFormatter),
+			Hooks:     make(logrus.LevelHooks),
+			Level:     logrus.InfoLevel,
+		}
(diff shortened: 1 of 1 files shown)


cmd/root.go
```diff
@@ -119,7 +119,13 @@ var RootCmd = &cobra.Command{
...
 	if err := RootCmd.Execute(); err != nil {
 		code := -1
-		var logger logrus.FieldLogger = logrus.StandardLogger()
+
+		var logger logrus.FieldLogger = &logrus.Logger{
+			Out:       os.Stderr,
+			Formatter: new(logrus.TextFormatter),
+			Hooks:     make(logrus.LevelHooks),
+			Level:     logrus.InfoLevel,
+		}
```
imiric, line 128: This is because the one from `StandardLogger()` has `ExitFunc: os.Exit` which was previously sidestepping the code below?

MStoykov: No, it is because even now configuring the logger is actually changing the `StandardLogger` so the `StandardLogger` here will be the Loki one (for example). But logging to it without waiting for it to push will do nothing. 

Arguably this error should be logged in both loggers IMO, but given that, for it to be "succesfully" logged with anything that won't just write it directly the whole `cmd` package will need to be rewritten (so it waits for a context for example.

imiric (approved):

MStoykov:
Get superseded by #1694

PR #1691, opened 2020-10-28
Cmd refactoring

MStoykov:


MStoykov:
I highly recommend looking at the commits 1 by 1 and with whitespace differences ignored. This way it is way more obvious what actually happens and ... how it *should* have zero impact on functionality. 

I even copy pasted the things from `init` functions in the order that they should be executed .... except for the rootCmd, which also can be done. 

This is the base for more work and I have specifically not tried to fix anything that I don't need to so it is as ... small and hopefully not break anything as possible

cmd/root.go
```diff
@@ -73,60 +73,143 @@ var (
...
+		Level:     logrus.InfoLevel,
+	}
+
+	var fallbackLogger logrus.FieldLogger = &logrus.Logger{
+		Out:       os.Stderr,
+		Formatter: new(logrus.TextFormatter),
+		Hooks:     make(logrus.LevelHooks),
+		Level:     logrus.InfoLevel,
+	}
+	RootCmd := getRootCmd(ctx, logger, fallbackLogger)
```
na--, line 136: No need to have a capitalized variable here anymore, right?

cmd/root.go
```diff
@@ -73,60 +73,143 @@ var (
...
+		},
+	}
+	return RootCmd
 }
 
 // Execute adds all child commands to the root command sets flags appropriately.
 // This is called by main.main(). It only needs to happen once to the rootCmd.
-func Execute() {
+func Execute() { //nolint:funlen
+	ctx := context.Background()
```
na--, line 122: Why not pass this in as a parameter as well?

MStoykov: Only the context? Mostly because this isn't used anywhere else but `main` and decided not to touch more files :P 

I would argue this can be left for when this actually gets some tests that will make that useful :)

cmd/root.go
MStoykov: No, they don't. But I wanted the diff to be as easy to read and reason about as possible, so left it  for later. I can rename them in a separate commit if you want, but I prefer the current ones where the differences in the code IMO stays more clear.

cmd/root.go
```diff
@@ -73,60 +73,142 @@ var (
...
+		"k6",
+		defaultConfigFileName,
+	)
+
+	RootCmd.PersistentFlags().AddFlagSet(rootCmdPersistentFlagSet())
+
+	archiveCmd := getArchiveCmd(logger)
+	RootCmd.AddCommand(archiveCmd)
+	archiveCmd.Flags().SortFlags = false
+	archiveCmd.Flags().AddFlagSet(archiveCmdFlagSet())
```
na--, line 154: hmm instead of building all of these flags here, is there a reason you can't add the flag sets in the respective commands? For example, 
```
	archiveCmd.Flags().SortFlags = false
	archiveCmd.Flags().AddFlagSet(archiveCmdFlagSet())
```
can occur here, right? https://github.com/loadimpact/k6/blob/6475b7aad50b49bcb4b4d9bcc589c3624b4458a2/cmd/archive.go#L100-L101

That way the diff will be even simpler - just moving these commands out of the magical `init()` to the respective `getFooCommand()`

MStoykov: I needed to that for the `convert` one ... but didn't do it for all the others just so that there were less changes as now and before (except for `convert`) the Command is first added and than it's flags are being added. 

AFAIK there is no difference and the `convert` tests pass so :shrug: 

Arguably this needs to happen IMO so I am fine with doing it

na-- (dismissed):

cmd/pause.go
```diff
@@ -26,34 +26,36 @@ import (
...
+			ui.Dump(stdout, status)
+			return nil
+		},
+	}
+	return pauseCmd
 }
 
 func init() {
-	RootCmd.AddCommand(pauseCmd)
 }
```
imiric, line 61: Delete this `init()`?

cmd/cloud.go
```diff
@@ -59,285 +59,286 @@ var (
 	showCloudLogs = true
 )
 
-//nolint:gochecknoglobals
-var cloudCmd = &cobra.Command{
-	Use:   "cloud",
-	Short: "Run a test on the cloud",
-	Long: `Run a test on the cloud.
+//nolint:funlen,gocognit,gocyclo
```
imiric, line 62: It feels like we replaced one bad `nolint` with a few worse ones 😄

Let's work on these soon as well :)

imiric (changes requested):
LGTM, just remove that empty `init()`.

cmd/cloud.go
MStoykov: The only reason those weren't already there was that we haven't touched near the start of the inner function since we have golangci-lint :rofl: .

imiric: Yeah, I know, but it reads like they're new issues, even though we're used to them, so it's good to see these `nolint`s explicitly.

na-- (dismissed):
LGTM, though can you rebase and maybe squash some of the commits? Or are you going to squash the whole PR?

imiric (dismissed):

MStoykov:
@na-- I prefer to squash just the last one in the one it fixes  and leave the rest so it's more obvious as it was here what actually happened instead of it being just one big commit

imiric (approved):

PR #1694, opened 2020-10-29
Make k6 wait for loki to finish pushing before stopping

MStoykov:


MStoykov:
This is dependant on #1691

cmd/root.go
```diff
@@ -177,20 +200,32 @@ func Execute() {
 				fields["hint"] = e.Hint
 			}
 		}
+
 		logger.WithFields(fields).Error(err)
+		if c.loggerIsRemote {
+			fallbackLogger.WithFields(fields).Error(err)
+			cancel()
+			<-c.loggerStopped // TODO have a timeout?
```
imiric, line 208: Not sure if we need a timeout if `^C` can stop this, no?

log/loki.go
```diff
@@ -282,7 +289,7 @@ func (h *lokiHook) loop() {
 		case <-h.ctx.Done():
 			ch := make(chan int64)
 			pushCh <- ch
-			ch <- 0
+			ch <- time.Now().Add(time.Second).UnixNano()
```
imiric, line 292: Since this is the cutoff time, is the second to deal with time sync issues and get some last events? Should it be equal to the push period, so maybe `h.pushPeriod` here instead?

imiric (commented):
LGTM, just some questions and nitpicks.

cmd/root.go
```diff
@@ -126,15 +111,53 @@ func getRootCmd(ctx context.Context, logger *logrus.Logger, fallbackLogger logru
...
+		// added them. The problem is that, besides being very inefficient,
+		// this actually also strips other special characters from the
+		// intended output, like the progressbar formatting ones, which
+		// would otherwise be fine (in a TTY).
+		//
+		// It would be much better if we avoid messing with the output and
+		// instead have a parametrized instance of the color library. It
+		// will return colored output if colors are enabled and simply
+		// return the passed input as-is (i.e. be a noop) if colors are
+		// disabled...
```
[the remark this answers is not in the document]
MStoykov: I just moved this around ... this was written by @na-- a while ago https://github.com/loadimpact/k6/commit/a9cc1309e4fc8317ab105dd474d56740b1f0a197  :wink:  I don't think having this in an issue will help anyone, if you start working on it, maybe make one and we can discuss it but in general I consider this a refactoring and having refactoring issues given the amount of issues we have that IMO we will never actually work on .. so they are basically "ideas" makes the issues hard to use so ... yeah, I don't think this needs to be added to an issue. 

I myself have added issues (mostly refactoring ones) that again are hardly unlikely for us to ever work on ... and even if we did the fact that there is an issue would've not helped in any way.

On related note, I have a branch with 1 more commit where I get EVEN more of the global variables like noColor in a struct and get it passed around, but that will have to wait for v0.30.0 maybe I can try this then

log/loki.go
MStoykov: yeah ... the basic idea is that ... as far as I can see you can call time.Now() in two parallel goroutines ... and you won't exactly get ... sequential times even if in reality they were ... so I added some time to get "hopefully" everything. 

I don't think that `h.pushPeriod` is better ... if anything I am more for `time.Hour` and a comment explaining why it is done at all (this took me 1 hour to figure out why I wasn't getting the last messages even though I was waiting for them and around 10 fmt.Printlns)

cmd/root.go
MStoykov: I guess not, and also the question is ... how big of timeout? I don't want to add one more configuration so :shrug: 
cc @na-- opinion ?

cmd/root.go
```diff
@@ -73,62 +73,181 @@ var (
...
-		if !cmd.Flags().Changed("log-output") {
-			if envLogOutput, ok := os.LookupEnv("K6_LOG_OUTPUT"); ok {
-				logOutput = envLogOutput
-			}
-		}
-		err := setupLoggers(logger, logFmt, logOutput)
-		if err != nil {
-			return err
-		}
+// This is get all for the main/root k6 command struct
```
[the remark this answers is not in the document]
MStoykov: I think I meant to write 
```
This is to get all fields for the main/root k6 command struct
```
but I eat half the words ? I remember that this strut was 4 things in the begining so :shrug:

cmd/root.go
na--: I think this timeout should be handled by each logger individually, so it's easier to configure. And, unless I'm mistaken, the loki logger sort of has that in https://github.com/loadimpact/k6/blob/5091636dbf920aadb51d52d53f7d43fd30049dd6/log/loki.go#L97

That said, it won't hurt to add another hardcoded 5 second max timeout here, with a warning message to the fallback logger, just in case...

na-- (approved):
LGTM, though I'd prefer if we have the global timeout as well... And I'm not sure I like the `rootCommand` struct, but I guess it's better than globals and we're restricted by the structure imposed from cobra, so it's fine for now...

cmd/root.go
```diff
@@ -65,53 +65,38 @@ var configFilePath = os.Getenv("K6_CONFIG") // Overridden by `-c`/`--config` fla
...
+	c.cmd = &cobra.Command{
+		Use:               "k6",
+		Short:             "a next-generation load generator",
+		Long:              BannerColor.Sprintf("\n%s", consts.Banner()),
+		SilenceUsage:      true,
+		SilenceErrors:     true,
+		PersistentPreRunE: c.persistentPreRunE,
 	}
 
 	confDir, err := os.UserConfigDir()
```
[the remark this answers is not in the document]
MStoykov: Hm ... that is a good idea

cmd/root.go
MStoykov: Okay, I am gonna add global timeout

na-- (approved):

imiric (approved):
The 1s delay to get the last messages seems a bit arbitrary, and maybe we should increase it or consider more robust solutions, but we can come back to it if needed, so LGTM.

PR #1695, opened 2020-10-29
Fix --user-agent='' making k6 to not send user agent

MStoykov:


js/modules/k6/http/request_test.go
```diff
@@ -333,6 +333,34 @@ func TestRequestAndBatch(t *testing.T) {
...
 			assert.NoError(t, err)
 		})
+
+		t.Run("Override empty", func(t *testing.T) {
+			_, err := common.RunString(rt, sr(`
+				var res = http.get("HTTPBIN_URL/user-agent", {
+					headers: { "User-Agent": "" },
+				});
+				if (typeof res.json()['User-Agent'] !== 'undefined') {
+					throw new Error("not undefined user agent: " + res.json()['user-agent'])
```
imiric, line 343: I'm not clear what you're checking here... These tests pass even with the previous version.

Object keys are case sensitive, so is the `User-Agent`/`user-agent` intentional or a typo?

MStoykov: :facepalm: I have apparently decided that this gets it from headers ... I have changed to use the other endpoint that actually works as I will expect - fixing the casing of the headers to what

imiric (approved):

PR #1697, opened 2020-10-29
catching of VU panics

MStoykov:


MStoykov:
This supersedes #1665

MStoykov:
For the record [this code](https://github.com/dop251/goja/blob/e21ccf39bfce8045562435deacb82e5ebff5a494/vm.go#L360-L380) resets the stack and other stuff to the previous values if there is a panic (even if the code below panics again) to what it was before a function was called and calling a Callable basically calls this. 

So unless this code panics in the middle the state of the Runtime should be "somewhat" restored to the previous one. Obviously any changes to global variables will stick. 

And if the panic in somehow has gotten the non goja stack but for example some internal to k6 state in a bad state that will lead to problems.  I can't figure out a way to break the goja stack in my first few tries and ... I can't remember anything in k6 that will have such a problem so it seems safe.

js/runner.go
```diff
@@ -621,11 +622,26 @@ func (u *VU) runFn(
 	u.Runtime.Set("__ITER", u.Iteration)
 	u.Iteration++
 
+	defer func() {
+		if r := recover(); r != nil {
+			gojaStack := u.Runtime.CaptureCallStack(20, nil)
+			err = fmt.Errorf("a panic occurred in VU code but was caught, %s", r)
```
imiric, line 628: Nit: the comma feels out of place here. A colon would be more intuitive here when reading the error message as it's referencing the original one.

js/runner_test.go
```diff
@@ -1735,3 +1737,67 @@ func TestSystemTags(t *testing.T) {
...
+			vu.(*ActiveVU).Console.logger = logger.WithField("source", "console")
+			err = vu.RunOnce()
+			require.Error(t, err)
+			assert.Contains(t, err.Error(), "a panic occurred in VU code but was caught, here we panic")
+			entries := hook.Drain()
+			require.Len(t, entries, 1)
+			assert.Equal(t, logrus.ErrorLevel, entries[0].Level)
+			require.Contains(t, entries[0].Message, "panic: here we panic")
+			require.Contains(t, entries[0].Message, "Goja stack:\nfile:///script.js:3:4(12)")
+			require.True(t, strings.HasSuffix(entries[0].Message, "Goja stack:\nfile:///script.js:3:4(12)"))
```
imiric, line 1792: Nit: one of these asserts is superfluous, no? I'd leave just the `HasSuffix` one.

js/runner.go
```diff
@@ -601,9 +603,8 @@ func (u *ActiveVU) RunOnce() error {
 
 func (u *VU) runFn(
 	ctx context.Context, isDefault bool, fn goja.Callable, args ...goja.Value,
-) (goja.Value, bool, time.Duration, error) {
+) (v goja.Value, isFullIteration bool, t time.Duration, err error) {
```
imiric, line 606: I remember you being against named return values in functions shorter than this one :)

Is there a particular requirement for it here or you're OK with it now? I'm fine with it either way, just curious.

MStoykov: I need to overwrite the `err`, there is practically no other way :) unless I make another smaller function just for that :)

MStoykov: Also, I am mostly against naked returns ;) I am perfectly fine with the named returned variables

js/runner_test.go
MStoykov: fixed in d1db643

js/runner.go
MStoykov: fixed in d1db643

imiric (approved):

na-- (approved):
LGTM. Please squash when merging and close https://github.com/loadimpact/k6/issues/1601 and https://github.com/loadimpact/k6/pull/1665

PR #1699, opened 2020-10-30
[fix] update windows installer to latest version.

legander:
Bump windows installer version in readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,7 +68,7 @@ brew install k6
 
 ### Windows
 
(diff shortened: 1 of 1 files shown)


MStoykov:
I do think it's about time we drop the Installation section in the Readme and just link to the docs ?
cc @na-- @imiric

legander:
Yeah maybe thats a better idea 👍  @MStoykov

na-- (approved):
We can fix this now and figure out a better solution later, so please merge this as it is @legander.

PR #1704, opened 2020-11-02
cloud: Update the progress one last time before stopping

MStoykov:


cmd/cloud.go
```diff
@@ -320,7 +320,7 @@ This will execute the test on the k6 cloud service. Use "k6 login cloud" to auth
 						logger.WithError(progressErr).Error("Test progress error")
 					}
 				case <-globalCtx.Done():
-					break runningLoop
+					keepLooping = false
 				}
```
MStoykov, line 324: I am wondering whether this shouldn't be dropped so we can actually get the `RunStatus` upgrade from the first `^C` ?

MStoykov:
closed in favor of #1705

PR #1712, opened 2020-11-06
gRPC and test fixes

na--:
This adds a bunch of tests and changes one details in the gRPC JS API: instead of `grpc.newClient()`, we now have the more JS idiomatic `new grpc.Client()`.

cc @rogchap

### DIFF
--- a/js/init_and_modules_test.go
+++ b/js/init_and_modules_test.go
@@ -1,0 +1,126 @@
+/*
+ *
+ * k6 - a next-generation load testing tool
+ * Copyright (C) 2020 Load Impact
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+package js_test
+
+import (
+	"context"
+	"net/url"
+	"testing"
+	"time"
+
+	"github.com/loadimpact/k6/js"
+	"github.com/loadimpact/k6/js/common"
+	"github.com/loadimpact/k6/js/internal/modules"
+	"github.com/loadimpact/k6/lib"
+	"github.com/loadimpact/k6/lib/testutils"
+	"github.com/loadimpact/k6/loader"
+	"github.com/loadimpact/k6/stats"
+	"github.com/spf13/afero"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"gopkg.in/guregu/null.v3"
+)
+
+type CheckModule struct {
+	t             testing.TB
+	initCtxCalled int
+	vuCtxCalled   int
+}
+
+func (cm *CheckModule) InitCtx(ctx context.Context) {
+	cm.initCtxCalled++
+	assert.NotNil(cm.t, common.GetRuntime(ctx))
+	assert.NotNil(cm.t, common.GetInitEnv(ctx))
+	assert.Nil(cm.t, lib.GetState(ctx))
+}
+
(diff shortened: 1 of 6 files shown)


js/init_and_modules_test.go
```diff
@@ -0,0 +1,110 @@
...
+	assert.Equal(t, checkModule.initCtxCalled, 2)
+	assert.Equal(t, checkModule.vuCtxCalled, 2)
+
+	arc := runner.MakeArchive()
+	assert.Equal(t, checkModule.initCtxCalled, 2) // shouldn't change, we're not executing the init context again
+	assert.Equal(t, checkModule.vuCtxCalled, 2)
+
+	runnerFromArc, err := js.NewFromArchive(logger, arc, rtOptions)
+	require.NoError(t, err)
+	assert.Equal(t, checkModule.initCtxCalled, 3) // chagenes because we need to get the exported functions
```
MStoykov, line 97:
```suggestion
	assert.Equal(t, checkModule.initCtxCalled, 3) // changes because we need to get the exported functions
```

js/init_and_modules_test.go
```diff
@@ -0,0 +1,110 @@
...
+	require.NoError(t, err)
+	assert.Equal(t, checkModule.initCtxCalled, 2)
+	assert.Equal(t, checkModule.vuCtxCalled, 0)
+
+	vuCtx, vuCancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer vuCancel()
+	activationParams := &lib.VUActivationParams{
+		RunContext:         vuCtx,
+		DeactivateCallback: func(lib.InitializedVU) {},
+	}
```
MStoykov, line 81: You need to use separate activationParams between the archive and not archive run to not have a race .. I think

MStoykov (dismissed):
LGTM apart from the race which is the same/very similar to https://github.com/loadimpact/k6/issues/1657 ... and might actually be a bug in the way RunContext is used ... This might need more investigation, but probably after the release.

js/init_and_modules_test.go
na--: hmm :thinking: I'll investigate later or on monday

na--: should be fixed in the last force-push

js/init_and_modules_test.go
na--: Resolved by https://github.com/loadimpact/k6/compare/5abc46540cbfb79737d2cf84dc9d6b5829e71777..18c17d742c0cd89c4d6b1e53bfb7788afe60aed2

js/modules/k6/grpc/client.go
```diff
@@ -70,8 +70,10 @@ type Client struct {
 	conn *grpc.ClientConn
 }
 
-// NewClient creates a new gPRC client to invoke RPC methods.
-func (*GRPC) NewClient(ctxPtr *context.Context) interface{} {
+// XClient represents the Client constructor (e.g. `new grpc.Client()`) and
+// creates a new gPRC client object that can load protobuf definitions, connect
+// to servers and invoke RPC methods.
+func (*GRPC) XClient(ctxPtr *context.Context) interface{} {
```
imiric, line 76: Ah, good to know about the `X` hack 😄

imiric (commented):
LGTM, just minor comments. And CI currently fails on `TestComplicatedFileImportsForGRPC`.

js/init_and_modules_test.go
```diff
@@ -0,0 +1,106 @@
...
+func (cm *CheckModule) VuCtx(ctx context.Context) {
+	cm.vuCtxCalled++
+	assert.NotNil(cm.t, common.GetRuntime(ctx))
+	assert.Nil(cm.t, common.GetInitEnv(ctx))
+	assert.NotNil(cm.t, lib.GetState(ctx))
+}
+
+func TestNewJSRunnerWithCustomModule(t *testing.T) {
+	checkModule := &CheckModule{t: t}
+	modules.Register("k6/check", checkModule)
```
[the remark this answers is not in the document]
na--: We just have to remember to use different packages for registering new k6 modules that help with testing... Given the singleton nature of the module registry, we don't want "pollution"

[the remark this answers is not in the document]
imiric: Yeah, that could be a problem. Should we add some kind of `Deregister` mechanism so that each test can clean up after itself? Since it's in `internal` we don't have to make it public and could only use it in tests.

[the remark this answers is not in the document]
na--: it shouldn't really matter if such tests are in their own package, but sure, an internal de-registering function can be added in the future

imiric (approved):

PR #1714, opened 2020-11-09
Drop some corejs

MStoykov:
The shim.min.js was generating by running
```
grunt "build:es7,es6.promise" "--blacklist=es7.math,es7.array.includes" "--library=off" "--path=custom" "uglify"
```

inside corejs checkedout to tag v2.5.1


MStoykov:
This reduces the mem used for the main.js script from https://github.com/loadimpact/k6/issues/1167#issuecomment-553835092 by 50% and the load start time from 33 seconds to 10 seconds for 5k VUs

MStoykov:
Things that I would've liked more:
1. It would've been better to get all the script files that were webpacked and import them. That would've meant that in the future if we want to drop something we can just stop importing it and delete it ... instead of having to rebuild the whole js file. 
2. This one includes a bit more than what seemed necessary(mostly some set/map methods that are no longer in the standard) ... so it is a little bit on the safer side. But dropping those seemed to give less than anound 3-4% impromvement so :shrug:

imiric (approved):
Great improvement! :tada:

Can we make that `grunt` command part of the `Makefile`? For reference, if nothing else.

MStoykov:
I would still prefer if we can try at 
> 1. It would've been better to get all the script files that were webpacked and import them. That would've meant that in the future if we want to drop something we can just stop importing it and delete it ... instead of having to rebuild the whole js file.

As that IMO is much better ... unfortunately it turned out to be more involved than what I hoped for so I decided to leave it for later :D

MStoykov:
test262 results https://github.com/MStoykov/k6-test262/pull/1

MStoykov:
superseeded by #1732

PR #1718, opened 2020-11-11
Release k6 v0.29.0

na--:
For easier reading of the rendered release notes: https://github.com/loadimpact/k6/blob/release-v0.29.0/release%20notes/v0.29.0.md

### DIFF
--- a/release notes/v0.29.0.md
+++ b/release notes/v0.29.0.md
@@ -1,0 +1,133 @@
+k6 v0.29.0 is here! :tada: It's a feature-packed release with tons of much-requested changes and additions, a lot of them implemented by awesome external contributors! :heart:
+
+As promised in the previous release notes, we're trying to stick to a roughly 8-week release cycle, so you can expect the next k6 version at the start of January 2021, barring any bugfix releases before that.
+
+## New features
+
+### Initial support for gRPC ([#1623](https://github.com/loadimpact/k6/pull/1623))
+
+k6 now supports unary gRPC calls via the new `k6/net/grpc` built-in module. Streaming RPCs are not yet supported and the **JS API is in beta**, so there might be slight breaking changes to the API in future k6 versions, but it's a good start on the road to fully supporting this much-requested protocol!
+
+This is a simple example of how the new module can be used with [grpcb.in](https://grpcb.in/):
+```js
+import grpc from "k6/net/grpc";
+
+let client = new grpc.Client();
+// Download addsvc.proto for https://grpcb.in/, located at:
+// https://raw.githubusercontent.com/moul/pb/master/addsvc/addsvc.proto
+// and put it in the same folder as this script.
+client.load(null, "addsvc.proto");
+
+export default () => {
+    client.connect("grpcb.in:9001", { timeout: "5s" });
+
+    let response = client.invoke("addsvc.Add/Sum", {
+        a: 1,
+        b: 2
+    });
+    console.log(response.message.v); // should print 3
+
+    client.close();
+}
+```
+
+You can find more information and examples how to use k6's new gRPC testing capabilities in [our documentation](https://k6.io/docs/javascript-api/k6-net-grpc).
+
+Huge thanks to @rogchap for adding this feature!
+
+### New options for configuring DNS resolution ([#1612](https://github.com/loadimpact/k6/pull/1612))
+
+You can now control some aspects of how k6 performs DNS resolution! Previously, k6 would have cached DNS responses indefinitely ([#726](https://github.com/loadimpact/k6/issues/726)) and always picked the first resolved IP ([#738](https://github.com/loadimpact/k6/issues/738)) for all connections. This caused issues, especially when load testing services that relied on DNS for load-balancing or auto-scaling.
+
+For technical reasons explored in ([#726](https://github.com/loadimpact/k6/issues/726)), k6 v0.29.0 still doesn't respect the actual TTL value of resolved IPs, that will be fixed in a future k6 version. For now, it simply allows users to specify a global static DNS TTL value and resolution strategy manually. It also has better defaults! Now, by default, the global DNS TTL value is 5 minutes and, if the DNS resolution returned multiple IPs, k6 will pick a random (preferably IPv4) one for each connection.
+
+You can also configure this behavior with the new `--dns` CLI flag, the `K6_DNS` environment variable, or the `dns` script/JSON option. Three DNS resolution options are exposed in this k6 version: `ttl`, `select`, and `policy`.
+
+Possible `ttl` values are :
+- `0`: no caching at all - each request will trigger a new DNS lookup.
+- `inf`: cache any resolved IPs for the duration of the test run _(the old k6 behavior)_.
+- any time duration like `60s`, `5m30s`, `10m`, `2h`, etc.; if no unit is specified (e.g. `ttl=3000`), k6 assumes milliseconds. _The new default value is `5m`._
+
+Possible `select` values are:
+- `first` - always pick the first resolved IP _(the old k6 behavior)_.
+- `random` - pick a random IP for every new connection _(the new default value)_.
+- `roundRobin` - iterate sequentially over the resolved IPs.
+
+Possible `policy` values are:
+- `preferIPv4`: use IPv4 addresses, if available, otherwise fall back to IPv6 _(the new default value)_.
+- `preferIPv6`: use IPv6 addresses, if available, otherwise fall back to IPv4.
+- `onlyIPv4`: only use IPv4 addresses, ignore any IPv6 ones.
+- `onlyIPv6`: only use IPv6 addresses, ignore any IPv4 ones.
+- `any`: no preference, use all addresses _(the old k6 behavior)_.
+
+Here are some configuration examples:
+
+```bash
+k6 run --dns "ttl=inf,select=first,policy=any" script.js # this is the old k6 behavior
+K6_DNS="select=random,ttl=5m,policy=preferIPv4" k6 cloud script.js # new default behavior
+# syntax for the JSON config file or for the exported script `options`:
+echo '{"dns": {"select": "roundRobin", "ttl": "1h33m7s", "policy": "onlyIPv6"}}' > config.json
+k6 run --config "config.json" script.js
+```
+
+### Support for Go extensions ([#1688](https://github.com/loadimpact/k6/pull/1688))
+
+After some discussions ([#1353](https://github.com/loadimpact/k6/issues/1353)) and exploration of different approaches for Go-based k6 extensions, we've settled on adopting something very similar to [caddy's extensions](https://caddyserver.com/docs/extending-caddy). In short, [`xk6`](https://github.com/k6io/xk6) (modeled after [`xcaddy`](https://github.com/caddyserver/xcaddy)) is a small stand-alone tool that will be able to build custom k6 binaries with 3rd party extensions bundled in. The extensions can be simple Git repositories (no central infrastructure needed!) with Go modules. They will be fully compiled, not interpreted, a part of the final custom k6 binary users will be able to build with k6.
+
+xk6 is not yet stable or documented, so any extension authors will struggle until we stabilize and document everything in the coming weeks. The important part is that the k6 changes that would allow xk6 to work were implemented in [#1688](https://github.com/loadimpact/k6/pull/1688), so k6 v0.29.0 is the first version compatible with xk6!
+
+Expect more information soon, but for a brief example, xk6 will work somewhat like this:
+
+```bash
+xk6 build v0.29.0 --with github.com/k6io/xk6-k8s --with github.com/k6io/xk6-sql@v0.1.1
+
+./k6 run some-script-with-sql-and-k8s.js
+```
+
+Thanks, @andremedeiros, for pushing us to add plugins in k6 and for making a valiant attempt to harness Go's poor plugin API! Thank you, @mardukbp, for pointing us towards the xcaddy approach and explaining its benefits!
+
+### Support for setting local IPs, potentially from multiple NICs ([#1682](https://github.com/loadimpact/k6/pull/1682))
+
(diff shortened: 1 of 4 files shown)


release notes/v0.29.0.md
```diff
@@ -0,0 +1,133 @@
...
+- `first` - always pick the first resolved IP _(the old k6 behavior)_.
+- `random` - pick a random IP for every new connection _(the new default value)_.
+- `roundRobin` - iterate sequentially over the resolved IPs.
+
+Possible `policy` values are:
+- `preferIPv4`: use IPv4 addresses, if available, otherwise fall back to IPv6.
+- `preferIPv6`: use IPv6 addresses, if available, otherwise fall back to IPv4 _(the new default value)_.
+- `onlyIPv4`: only use IPv4 addresses, ignore any IPv6 ones.
+- `onlyIPv6`: only use IPv6 addresses, ignore any IPv4 ones.
+- `any`: no preference, use all addresses _(the old k6 behavior)_.
```
imiric, line 61: Do we need to repeat all these options docs here? It seems like clutter for the release notes. I think just having the intro, example and a link to the docs would be enough.

release notes/v0.29.0.md
```diff
@@ -0,0 +1,133 @@
...
+k6 run --config "config.json" script.js
+```
+
+### Support for Go extensions ([#1688](https://github.com/loadimpact/k6/pull/1688))
+
+After some discussions ([#1353](https://github.com/loadimpact/k6/issues/1353)) and exploration of different approaches for Go-based k6 extensions, we've settled on adopting an approach similar to [caddy's](https://caddyserver.com/docs/extending-caddy). In short, [`xk6`](https://github.com/k6io/xk6) (modeled after [`xcaddy`](https://github.com/caddyserver/xcaddy)) is a small stand-alone tool that will allow the building of custom k6 binaries with 3rd party extensions bundled in. The extensions can be simple Git repositories (no central infrastructure needed!) with Go modules. They will be fully compiled, not interpreted, a part of the final custom k6 binary users will be able to build with k6.
+
+xk6 is not yet stable or documented, so any plugin authors will struggle until we stabilize and document everything in the coming weeks. The important part is that the k6 changes that would allow xk6 to work were implemented in [#1688](https://github.com/loadimpact/k6/pull/1688), so k6 v0.29.0 is the first version compatible with xk6!
```
imiric, line 77: In the first paragraph we call them "extensions" and here we call them "plugins". It's even more confusing as in the code they're "modules". We should standardize a single naming convention to avoid confusion, so maybe "plugin" is a better generic term, and internally "modules" makes sense, mainly because they're "JS modules".

release notes/v0.29.0.md
```diff
@@ -0,0 +1,133 @@
...
+
+You can now specify a list of source IPs, IP ranges and CIDRs for `k6 run`, from which VUs will make requests via the new `--local-ips` CLI flag or `K6_LOCAL_IPS` environment variable. The IPs will be sequentially given out to VUs, allowing you to distribute load between different local addresses. This option doesn't change anything on the OS level, so the IPs need to already be configured on the OS level in order for k6 to be able to use them.
+
+The biggest use case for this feature is splitting the network traffic from k6 between multiple network adapters, thus potentially greatly increasing the available network throughput. For example, if you have 2 NICs, you can run k6 with `--local-ips="<IP-from-first-NIC-1>,<IP-from-first-NIC-2>"` to balance the traffic equally between them - half of the VUs will use the first IP and the other half will use the second. This can scale to any number of NICs, and you can repeat some local IPs to give them more traffic. For example, `--local-ips="<IP1>,<IP2>,<IP3>,<IP3>"` will split VUs between 3 different source IPs in a 25%:25%:50% ratio.
```
imiric, line 93:
```suggestion
The biggest use case for this feature is splitting the network traffic from k6 between multiple network adapters, thus potentially greatly increasing the available network throughput. For example, if you have 2 NICs, you can run k6 with `--local-ips="<IP-from-first-NIC>,<IP-from-second-NIC>"` to balance the traffic equally between them - half of the VUs will use the first IP and the other half will use the second. This can scale to any number of NICs, and you can repeat some local IPs to give them more traffic. For example, `--local-ips="<IP1>,<IP2>,<IP3>,<IP3>"` will split VUs between 3 different source IPs in a 25%:25%:50% ratio.
```

?

release notes/v0.29.0.md
```diff
@@ -0,0 +1,133 @@
...
+export let options = {
+  blockHostnames: ["*.k6.io" , "www.example.com"],
+};
+```
+
+Thanks to @krashanoff for implementing this feature!
+
+## UX and enhancements
+
+- HTTP: The [gjson library](https://github.com/tidwall/gjson/) k6 uses for handling the HTTP `Response.json(selector)` behavior was updated, so we now support more [modifiers](https://github.com/tidwall/gjson/blob/master/SYNTAX.md#modifiers) like `@flatten` and multipaths ([#1626](https://github.com/loadimpact/k6/pull/1626)). Thanks, @sondnm!
```
imiric, line 111: We should probably mention this in the k6-docs PR.

release notes/v0.29.0.md
na--: Maybe for most things I'd agree, for example that's how I handled the gRPC section. But considering this is sort of a breaking change, and the release notes are not all that lengthy, I think it's fine.

release notes/v0.29.0.md
na--: hmm the docs already forward to https://github.com/tidwall/gjson#path-syntax for the `.json()` selector option, so I am not sure we need to change anything

release notes/v0.29.0.md
imiric: It still feels like too much detail to have these verbatim from the options page, but OK. :)

na--: it's not quite verbatim, it's actually longer :sweat_smile: there is some extra explanation on top and stresses on the differences with the old k6 versions (see the many "_(the old k6 behavior)_" and "_(the new default value)_" mentions)

release notes/v0.29.0.md
```diff
@@ -0,0 +1,133 @@
...
+export default () => {
+    client.connect("grpcb.in:9001", { timeout: "5s" });
+
+    let response = client.invoke("addsvc.Add/Sum", {
+        a: 1,
+        b: 2
+    })
+    console.log(response.message.v) // should print 3
+
+    client.close()
```
imiric, line 30: Let's be civilized ;)

```suggestion
client.load(null, "addsvc.proto");

export default () => {
    client.connect("grpcb.in:9001", { timeout: "5s" });

    let response = client.invoke("addsvc.Add/Sum", {
        a: 1,
        b: 2
    });
    console.log(response.message.v); // should print 3

    client.close();
```

na--: ah... :+1:  I fixed all of these in the docs PR, but forgot to do it here :sweat_smile:

release notes/v0.29.0.md
```diff
@@ -0,0 +1,133 @@
...
+
+For technical reasons explored in ([#726](https://github.com/loadimpact/k6/issues/726)), k6 v0.29.0 still doesn't respect the actual TTL value of resolved IPs, that will be fixed in a future k6 version. For now, it simply allows users to specify a global static DNS TTL value and resolution strategy manually. It also has better defaults! Now, by default, the global DNS TTL value is 5 minutes and, if the DNS resolution returned multiple IPs, k6 will pick a random (preferably IPv4) one for each connection.
+
+You can also configure this behavior with the new `--dns` CLI flag, the `K6_DNS` environment variable, or the `dns` script/JSON option. Only 3 DNS resolution options are exposed in this k6 version, `ttl`, `select`, and `policy`.
+
+Possible `ttl` values are :
+- `0`: no caching at all - each request will trigger a new DNS lookup.
+- `inf`: cache any resolved IPs for the duration of the test run _(the old k6 behavior)_.
+- any time duration like `60s`, `5m30s`, `10m`, `2h`, etc.; if no unit is specified (e.g. `ttl=3000`), k6 assumes milliseconds. _The new default value is `5s`._
```
MStoykov, line 49:
```suggestion
- any time duration like `60s`, `5m30s`, `10m`, `2h`, etc.; if no unit is specified (e.g. `ttl=3000`), k6 assumes milliseconds. _The new default value is `5m`._
```

release notes/v0.29.0.md
```diff
@@ -0,0 +1,133 @@
...
+
+You can now control some aspects of how k6 performs DNS resolution! Previously, k6 would have cached DNS responses indefinitely ([#726](https://github.com/loadimpact/k6/issues/726)) and always picked the first resolved IP ([#738](https://github.com/loadimpact/k6/issues/738)) for all connections. This caused issues, especially when load testing services that relied on DNS for load-balancing or auto-scaling.
+
+For technical reasons explored in ([#726](https://github.com/loadimpact/k6/issues/726)), k6 v0.29.0 still doesn't respect the actual TTL value of resolved IPs, that will be fixed in a future k6 version. For now, it simply allows users to specify a global static DNS TTL value and resolution strategy manually. It also has better defaults! Now, by default, the global DNS TTL value is 5 minutes and, if the DNS resolution returned multiple IPs, k6 will pick a random (preferably IPv4) one for each connection.
+
+You can also configure this behavior with the new `--dns` CLI flag, the `K6_DNS` environment variable, or the `dns` script/JSON option. Only 3 DNS resolution options are exposed in this k6 version, `ttl`, `select`, and `policy`.
```
MStoykov, line 44: I don't really think "Only 3" is necessary

release notes/v0.29.0.md
```diff
@@ -0,0 +1,133 @@
...
+- any time duration like `60s`, `5m30s`, `10m`, `2h`, etc.; if no unit is specified (e.g. `ttl=3000`), k6 assumes milliseconds. _The new default value is `5s`._
+
+Possible `select` values are:
+- `first` - always pick the first resolved IP _(the old k6 behavior)_.
+- `random` - pick a random IP for every new connection _(the new default value)_.
+- `roundRobin` - iterate sequentially over the resolved IPs.
+
+Possible `policy` values are:
+- `preferIPv4`: use IPv4 addresses, if available, otherwise fall back to IPv6.
+- `preferIPv6`: use IPv6 addresses, if available, otherwise fall back to IPv4 _(the new default value)_.
```
MStoykov, line 58:
```suggestion
- `preferIPv4`: use IPv4 addresses, if available, otherwise fall back to IPv6 _(the new default value)_.
- `preferIPv6`: use IPv6 addresses, if available, otherwise fall back to IPv4.
```

release notes/v0.29.0.md
MStoykov: I don't think this was the old behaviour as the IPs were sorted by having ipv4 so ... in practice ... the default was preferIPv4

MStoykov (approved):

release notes/v0.29.0.md
na--: > IPs were sorted by having ipv4

I don't think that was the case, can you point me to the code?

release notes/v0.29.0.md
na--: Fixed by https://github.com/loadimpact/k6/commit/879d2bf6a58ec4a1ba1d9d1b71b0e0019545bc0d?

release notes/v0.29.0.md
na--: Resolved by https://github.com/loadimpact/k6/commit/879d2bf6a58ec4a1ba1d9d1b71b0e0019545bc0d

release notes/v0.29.0.md
```diff
@@ -0,0 +1,133 @@
...
+- `any`: no preference, use all addresses _(the old k6 behavior)_.
+
+Here are some configuration examples:
+
+```bash
+k6 run --dns "ttl=inf,select=first,policy=any" script.js # this is the old k6 behavior
+K6_DNS="select=random,ttl=5m,policy=preferIPv4" k6 cloud script.js # new default behavior
+# syntax for the JSON config file or for the exported script `options`:
+echo '{"dns": {"select": "roundRobin", "ttl": "1h33m7s", "policy": "onlyIPv6"}}' > config.json
+k6 run --config "config.json" script.js
```
imiric, line 70: Nit: this can be a one-liner:

```bash
k6 run --config <(echo '{"dns": {"select": "roundRobin", "ttl": "1h33m7s", "policy": "onlyIPv6"}}') script.js
```

But maybe it's clearer this way.

imiric (dismissed):
:shipit: :rocket:

release notes/v0.29.0.md
MStoykov: yeah ... I don't think code golfing the examples is good idea :P

imiric: Don't think this qualifies as "code golfing" 😄, it's a well known redirection feature, but sure, it might not be as clear.

MStoykov (dismissed):

MStoykov (approved):

imiric (approved):

release notes/v0.29.0.md
MStoykov: I might be wrong .. it seems it orders it based on https://tools.ietf.org/html/rfc6724 here https://github.com/golang/go/blob/f2e186b87754d3f84a692876501c923eb58f2ee4/src/net/dnsclient_unix.go#L716 ... this does seem to check (as per the RFC) if there is at all a way to open a [udp connection to each IP](https://github.com/golang/go/blob/f2e186b87754d3f84a692876501c923eb58f2ee4/src/net/addrselect.go#L41-L56)  and then as far as I can understand orders the [IPv6 first](https://github.com/golang/go/blob/f2e186b87754d3f84a692876501c923eb58f2ee4/src/net/addrselect_test.go#L14) ... so maybe you are right ... but I am not certain and this is too big of a rabbit hole, but I still think that maybe we should remove the "this was previously the default" statement. 

Maybe we should test it on a machine with an actual working ipv6 address but with v0.28.0 ?

na--: my thinking is that we didn't do any IP address sorting and didn't have any such preference in k6 before 0.29... What the OS or DNS servers or even the Go runtime did is immaterial, for the purposes of k6 it was like black magic and it just got the first IP that was returned. I think it's not unreasonable to claim that `ttl=inf,select=first,policy=any` was the previous DNS behavior, since if you set these options now, you should get exactly the same behavior as k6 v0.28.0, right?

PR #1723, opened 2020-11-16
Fix hostname blocking

na--:
This should close https://github.com/loadimpact/k6/issues/1721 (cc @curvedriver). There were 2 separate bugs... :disappointed: Wildcards didn't catch 0-length matches, and when you blocked both a `domain` and a `sub.domain`, the `HostnameTrie` stopped blocking the `domain` because the `len(t.children) == 0` check was no longer true (wrong assumption that you can't block both a top-level and a sub-domain). This script more succinctly demonstrates the issue, running it with k6 0.29.0 will fail all checks, whereas with this PR they all pass:
```js
import http from 'k6/http';
import { check, group } from "k6";

export const options = {
    blockHostnames: ['*httpbin.org', 'test.k6.io', 'k6.io']
}

export default function getFromBlockedHostname() {
    [
        'https://httpbin.org/get?my_param=my_value',
        'https://k6.io/',
    ].forEach(url =>
        group(url, () => {
            check(http.get(url, { redirects: 0 }), {
                "status is 0": (r) => r.status === 0,
                "error_code is 1111": (r) => r.error_code === 1111,
            });
        }));
}
```

### DIFF
--- a/lib/types/hostnametrie.go
+++ b/lib/types/hostnametrie.go
@@ -163,14 +163,16 @@ func (t *HostnameTrie) childInsert(s string) error {
 func (t *HostnameTrie) Contains(s string) (matchedPattern string, matchFound bool) {
 	s = strings.ToLower(s)
 	if len(s) == 0 {
-		return s, len(t.children) == 0
-	}
-
-	rStr := []rune(s)
-	last := len(rStr) - 1
-	if c, ok := t.children[rStr[last]]; ok {
-		if match, matched := c.Contains(string(rStr[:last])); matched {
-			return match + string(rStr[last]), true
+		if len(t.children) == 0 {
+			return "", true
+		}
+	} else {
+		rStr := []rune(s)
+		last := len(rStr) - 1
+		if c, ok := t.children[rStr[last]]; ok {
+			if match, matched := c.Contains(string(rStr[:last])); matched {
+				return match + string(rStr[last]), true
+			}
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


lib/types/hostnametrie.go
```diff
@@ -135,42 +138,52 @@ func (t *HostnameTrie) insert(s string) error {
...
 
-func (t *HostnameTrie) childInsert(s string) error {
+type trieNode struct {
+	isLeaf   bool
+	children map[rune]*trieNode
+}
+
+func (t *trieNode) insert(s string) error {
 	if len(s) == 0 {
+		t.isLeaf = true
```
MStoykov, line 157: Shouldn't this become ... false if it goes beyond this ?

na--: I don't understand - the next line is `return nil`, so it can't go beyond this :confused:

MStoykov: I meant that if on sequential call to insert on a node that is already a leaf ... we won't unleaf it ... which I now understand is exactly what this is fixing (but I needed to read the whole code ...). 

I am somewhat of the opinion that if there is "*valid.test" and "test.valid,test" we should probably not add the second one to tree ... and/or give out a warning/error that there is a more general glob already. 

But this can be done as an optimization in later PR and I am not particularly opinioanted about it ...

MStoykov (approved):

imiric (approved):
LGTM, nice work! 👍

PR #1725, opened 2020-11-17
Let babel run at most twice on every script

MStoykov:
This is continuation of PR #1651 as it turns out that sometimes babel
doesn't do enough on the first time it runs over some scripts. So let it
run twice.

### DIFF
--- a/js/compiler/compiler.go
+++ b/js/compiler/compiler.go
@@ -71,18 +71,29 @@ func (c *Compiler) Transform(src, filename string) (code string, srcmap *SourceM
 }
 
 // Compile the program in the given CompatibilityMode, wrapping it between pre and post code
-func (c *Compiler) Compile(src, filename, pre, post string,
-	strict bool, compatMode lib.CompatibilityMode) (*goja.Program, string, error) {
+func (c *Compiler) Compile(
+	src, filename, pre, post string, strict bool, compatMode lib.CompatibilityMode,
+) (*goja.Program, string, error) {
+	var babelTransformations int
+	if compatMode == lib.CompatibilityModeExtended {
+		babelTransformations = 2 // this is two because in some cases babel doesn't do enough in 1 iteration
+	}
+	return c.compile(src, filename, pre, post, strict, babelTransformations)
+}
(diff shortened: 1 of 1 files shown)


na-- (changes requested):
As I said in slack, I don't think this is a good workaround for bugs in babel... It might end up being necessary, but at this stage, I'd much prefer if https://github.com/loadimpact/k6/pull/1663 or something like it is the solution. For now, before committing to anything, I'd like us to explore the issue more, as well as have a clearer idea what upgrading babel will mean.

MStoykov:
I am of the opinion that given that this is an issue we(I) introduced if this is not the solution we can revert the original PR and then go on the "see what updating babel" rabbit hole is ... 

We can then again revert this (or the other) fix, when we know what we are going to do, instead of finding out how more and more users are affected by this change. 

I for one ... am AGAINST again updating babel for the reasons of:
1. it makes test slower 
2. it makes smaller scripts .. .slower to transpile
3. fixes two different corner cases which might be fixed by goja improvements ... at some point
4. Given how this is a cornerstone of k6 ... I am of the opinion that will see more problems down the line than not ... for fixing ... two corner case issues, one of which is fixed by the lines in this PR ... 

I am not arguing this is a great engineering solution ... but it is a great product solution ...

na--:
I disagree. I still don't know all of the issues of upgrading babel, so I can't comment on that _yet_. I will apparently have to become familiar, but until then, my opinion is that https://github.com/loadimpact/k6/pull/1651 was a bugfix for a bug in k6. The fact that this k6 bug masked another bug in the babel version we use is immaterial. 

The only reason I can see for reverting https://github.com/loadimpact/k6/pull/1651 is if we want to release a new k6 hotfix version v0.29.1. And we can do that if enough users of k6 v0.29.0 hit a bug with webpacked JS code or whatever. But if it's just a single user, as the case stands now, it doesn't justify us from removing the v0.29.0 bugfix.

MStoykov:
I am closing this as the fix is bad and it is very rare and unlikely issue

PR #1394, opened 2020-04-14
Fix: windows installer

mostafa:
Hi,

I have added the new k6 logo plus its icon and banners to the windows installer. Upon getting the code signing certificate, I'll fix the package signing across all distributions, especially #1034 and #1247.

Also, the current ".ico" file is not used to make a k6 executable, ~so I created a ".syso" file using [rsrc](https://github.com/akavel/rsrc) tool. The `go build` process picks it up and includes it in the final executable file.~ so I used [goversioninfo](https://github.com/josephspurrier/goversioninfo) to generate a resource file to accompany the `go build` process.

I don't expect a new release just for the sake of this small change, so this PR can wait for the next release, or you decide.

Issue #1034, opened 2019-05-30
Sign release binaries and installers for Windows 


na--:
As illustrated by [this recent issue](https://github.com/loadimpact/k6/issues/1029), we might benefit from signing our binary releases with a valid key for Windows.

To be honest, I'm not sure if that would prevent or even decrease future antivirus false-positive results, but even if we just make that "_Do you want this app from an unknown publisher to make changes to your PC_" warning dialog less scary, it's probably worth it. 

Some resources on the topic:
- https://docs.microsoft.com/en-us/windows-hardware/drivers/dashboard/get-a-code-signing-certificate
- https://www.rahulpnath.com/blog/code-signing-in-azure-devops-pipeline/
- https://stackoverflow.com/questions/252226/signing-a-windows-exe-file
- https://docs.microsoft.com/en-us/windows/desktop/SecCrypto/cryptography-tools
- https://docs.microsoft.com/en-us/windows/desktop/SecCrypto/signtool

Potentially connected issues: 
- Use Azure Pipelines for CI https://github.com/loadimpact/k6/issues/959:
  Since it's a Microsoft service, it's worth investigating whether there is some sort of a better way to do binary signing there. I'm somewhat worried about just uploading our future code-signing key to a random CI service, so maybe they have some sort of a key enclave or managed signing or something that adds a bit of security and transparency to the process. 
- Chocolatey installation (this new PR https://github.com/loadimpact/k6/pull/1027 and the [current windows installation instructions](https://github.com/loadimpact/k6#windows))
  Worth investigating if chocolatey, being a package manager, would actually benefit from this whole binary code-signing business. In the Linux world, it's not the binaries that are signed, but the different rpm/deb/apk/pacman/etc. packages...

majkinetor:
> Worth investigating if chocolatey, being a package manager, would actually benefit from this whole binary code-signing business

It wouldn't at this moment AFAIK. Furthermore, Chocolatey Gallery has integration with Virus Total so packages are as safe as it can get because they are not going to be approved if there are more then couple of false positives on entirety of 70+ antivirus engines. 

In the case of latest version there is [one false positive result](https://www.virustotal.com/gui/file/6e294565795c77f4a3da10fa3d4a4a935e2aceb6cf84300f3eb082724a0365b7/detection):

![image](https://user-images.githubusercontent.com/85767/62613358-99577280-b909-11e9-834a-70af9091aeaf.png)
[...]

PR #1727, opened 2020-11-18
Windows binary improvements

imiric:
This enhances the Windows k6 binary by adding an icon and some release metadata like version information, visible in the file's Properties. Additionally, the k6.ico was updated to the new logo in higher resolution, and the MSI installer assets were updated as well.

Before:
2020-11-18-122751_643x654_scrot

After:
2020-11-19-142432_778x648_scrot
2020-11-19-123938_636x500_scrot
2020-11-19-124010_636x509_scrot

This is based on previous work in #1394 ~and the [templating suggestion by @na--](https://github.com/loadimpact/k6/pull/1394#pullrequestreview-464806731).~ We decided to abandon the templating approach in favor of passing metadata via CLI args.

You can inspect the binary if you download the build artifacts from [here](https://github.com/imiric/k6/runs/1424342900).

Pending work:
- I'll add MSI package signing (#1034, #1247) in a follow-up PR.
- As discussed I think we agree to get rid of the `Version` constant and rely only on Git tags, but I'll leave that change for later as well.


packaging/gen-versioninfo.go
```diff
@@ -0,0 +1,55 @@
...
+
+	"github.com/loadimpact/k6/lib/consts"
+)
+
+type version struct {
+	Major, Minor, Patch string
+}
+
+func main() {
+	ver := strings.SplitN(consts.Version, ".", 3)
```
imiric, line 45: It would be better if `consts.Version` is not a real constant, but something like `[]uint16{0, 29, 0}`, but that would involve a lot of other changes, and I think we should get this from Git anyway.

na--:
I skimmed through the changes and at glance they LGTM, but to answer your questions:
  - I'm fine with the .msi signing either being in the same PR or in a separate one, whatever you prefer
  - CLI flags for `goversioninfo` seem slightly preferable - it will make the CI script a bit more verbose, but it'd be much easier to understand what's happening and there would be a lot fewer things that can go wrong
  - not reading the Go `Version` const is probably fine... you should be able to get the git tag version out of some GitHub Actions environment variable, right?

.github/workflows/all.yml
```diff
@@ -171,6 +171,11 @@ jobs:
               && go install)
           sudo apt-get update -y
           sudo apt-get install -y fakeroot rpm
+      - name: Generate Windows binary metadata
+        run: |
+          (cd && go get github.com/josephspurrier/goversioninfo/cmd/goversioninfo)
```
na--, line 176:
```suggestion
          (cd && go get github.com/josephspurrier/goversioninfo/cmd/goversioninfo@1.2.0)
```
we probably should pin the version we install, just in case...

MStoykov (commented):
LGTM, I agree with both of you on all the points :D . 

I would like to remind people that PR numbers are free (until we are close to a power of 2 ;) ), so having (self-contained) changes spread over multiple PRs IMO is preferable ;)

imiric:
> I would like to remind people that PR numbers are free (until we are close to a power of 2 ;) ), so having (self-contained) changes spread over multiple PRs IMO is preferable ;)

True, but MSI package signing is closely related to this, and the less notifications I get and GitHub I have to use, the better :-P (See how I messed up the Assignees and Reviewers in this PR...)

But makes sense, I'll do that in another PR.

.github/workflows/all.yml
imiric: Resolved in latest push.

MStoykov (approved):

na-- (approved):
LGTM

PR #1732, opened 2020-11-19
PoC loading of specific parts of core-js

MStoykov:
This is a very WIP alternative to #1714 that actually loads specific files from it and it is obvious what is loaded in the code instead of needing to know how we build the shim file. 

TODO:
- [ ] Discuss if this is at all useful ... The main benefit is that if tomorrow we decide to drop `es6.promise.js` we just drop it from the [index.js](https://github.com/loadimpact/k6/pull/1732/files#diff-e24e57f081cccef9f422ddff4a40b556ef613c958a617c8397869e45d838185fR126-R170) and then drop it dependencies.
- [ ] Better code - again this is PoC - I intend for there to be a test that does the checking if some files are not needed for example, but also very likely to refactor the whole code ;). Also it will be nice if the source codes are compiled only once.
- [ ] Decide if we want to drop even more corejs? The nice thing here is that you also can see which files are imported as dependencies 
- [ ] Make some script that does the git checkout + copy of the concrete files we need and runs go generate to embed them, without actually adding the files to the repo. This will save some space

Problems:
- This actually generates a bigger binary, and the rice-box.go is bigger as we have slightly more code actually. 
- It is probably slightly slower as it needs to require some stuff, haven't benchmarked it yet.


MStoykov:
Some performance running
```
import http from "k6/http";
import { sleep } from "k6";

export let options = {
    discardResponseBodies: true,
}
export default function() {
    http.get("https://httpbin.test.loadimpact.com/get");
    sleep(Math.random() * 2 + 3);
}
```
with `-vus 5000 --duration 60s main.js -v --no-thresholds --no-summary` 
version       | iterations | mem usage | CPU usage | time to initialize VUs    | run time |
--------------  | ------------ | ---------------- | --------------- | ----------------------------- | ----------- |
6081e6a3   |  64391    | 17066648    | 440%          | ~29.5s                         | 1:54.10  |
stripCoreJs | 64430     | 8201408      | 345%          | ~10.45s                       | 1:39.45  |
ddf9ecd      | 65892     | 6659656      | 386%          | ~10.8s                         | 1:29.55  |
```
The first is ( the current ) master, stripCoreJs is #1714 rebased on it, and the last is the current last commit here.

The numbers move a little bit, but are in the same magnitude.  

I don't know why stripCoreJS uses more memory, maybe I dropped something additional or web packing adds a lot of additional stuff?

simskij:
Yeah, webpack adds some code, but it should be fairly insignificant in this case. I need to take this branch for a drive before having anything useful to say. Will get back to you! Good job nonetheless! 👍

MStoykov:
close in favor of #1772

PR #1734, opened 2020-11-20
Update golang.org/x/crypto and golang.org/x/net

na--:
Turns out we were using very old versions for these libraries and we didn't notice partially because of the `replace` directives. These are now updated to the latest commits on their respective `release-branch.go1.15` branches...


MStoykov (approved):
LGTM :crossed_fingers: :grimacing:

PR #1735, opened 2020-11-20
Corejs rework pat2

MStoykov:
This is just dropping even more stuff I have the separate just to be easier to look at. In the final version it will probably all be squashed together


MStoykov:
close in favor of #1772

PR #1738, opened 2020-11-25
Unify duration configuration values

na--:
This should close https://github.com/loadimpact/k6/issues/1305

### DIFF
--- a/js/modules/k6/ws/ws.go
+++ b/js/modules/k6/ws/ws.go
@@ -390,7 +390,10 @@ func (s *Socket) trackPong(pingID string) {
 // timeout, which is in ms, has elapsed
 func (s *Socket) SetTimeout(fn goja.Callable, timeoutMs float64) error {
 	// Starts a goroutine, blocks once on the timeout and pushes the callable
-	// back to the main loop through the scheduled channel
+	// back to the main loop through the scheduled channel.
+	//
+	// Intentionally not using the generic GetDurationValue() helper, since this
+	// API is meant to use ms, similar to the original SetTimeout() JS API.
 	d := time.Duration(timeoutMs * float64(time.Millisecond))
 	if d <= 0 {
 		return fmt.Errorf("setTimeout requires a >0 timeout parameter, received %.2f", timeoutMs)
@@ -416,7 +419,10 @@ func (s *Socket) SetTimeout(fn goja.Callable, timeoutMs float64) error {
 // in ms
 func (s *Socket) SetInterval(fn goja.Callable, intervalMs float64) error {
 	// Starts a goroutine, blocks forever on the ticker and pushes the callable
-	// back to the main loop through the scheduled channel
+	// back to the main loop through the scheduled channel.
+	//
+	// Intentionally not using the generic GetDurationValue() helper, since this
+	// API is meant to use ms, similar to the original SetInterval() JS API.
 	d := time.Duration(intervalMs * float64(time.Millisecond))
 	if d <= 0 {
 		return fmt.Errorf("setInterval requires a >0 timeout parameter, received %.2f", intervalMs)
--- a/lib/types/types.go
+++ b/lib/types/types.go
@@ -24,6 +24,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"math"
 	"reflect"
 	"strconv"
 	"strings"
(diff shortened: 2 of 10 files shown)


lib/types/types.go
```diff
@@ -233,3 +227,56 @@ func (d NullDuration) ValueOrZero() Duration {
...
+	case time.Duration:
+		return d, nil
+	case string:
+		return ParseExtendedDuration(d)
+	case float32:
+		return time.Duration(float64(d) * float64(time.Millisecond)), nil
+	case float64:
+		return time.Duration(d * float64(time.Millisecond)), nil
+	default:
+		n, err := getInt64(v)
```
MStoykov, line 276: Are any beside time.Duration, string, float64 and int64 ... even possible? Because the previous code didn't handle the rest, and I have the feeling that was because we will never actually get one ?

na--: I added them out of an abundance of caution... I have no idea if some weird webpack minification magic or something like that won't try to encode plain integer values as something that goja will export as uint8 or whatever... Stranger things have happened... :sweat_smile: And given that this helper function is intentionally outside of `js/`, I also think it makes some sense to handle the other Go int native types. This way the function is more complete and can basically convert every possible value we can recognize as a `time.Duration` to that, no need to second-guess.

js/modules/k6/ws/ws.go
```diff
@@ -390,7 +390,10 @@ func (s *Socket) trackPong(pingID string) {
 // timeout, which is in ms, has elapsed
 func (s *Socket) SetTimeout(fn goja.Callable, timeoutMs float64) error {
 	// Starts a goroutine, blocks once on the timeout and pushes the callable
-	// back to the main loop through the scheduled channel
+	// back to the main loop through the scheduled channel.
+	//
+	// Intetionally not using the generic GetDurationValue() helper, since this
```
imiric, line 395:
```suggestion
	// Intentionally not using the generic GetDurationValue() helper, since this
```

js/modules/k6/ws/ws.go
```diff
@@ -416,7 +419,10 @@ func (s *Socket) SetTimeout(fn goja.Callable, timeoutMs float64) error {
 // in ms
 func (s *Socket) SetInterval(fn goja.Callable, intervalMs float64) error {
 	// Starts a goroutine, blocks forever on the ticker and pushes the callable
-	// back to the main loop through the scheduled channel
+	// back to the main loop through the scheduled channel.
+	//
+	// Intetionally not using the generic GetDurationValue() helper, since this
```
imiric, line 424:
```suggestion
	// Intentionally not using the generic GetDurationValue() helper, since this
```

lib/types/types.go
```diff
@@ -233,3 +227,56 @@ func (d NullDuration) ValueOrZero() Duration {
...
+func getInt64(v interface{}) (int64, error) {
+	switch n := v.(type) {
+	case int:
+		return int64(n), nil
+	case int8:
+		return int64(n), nil
+	case int16:
+		return int64(n), nil
+	case int32:
+		return int64(n), nil
```
imiric, line 240: I'm still baffled that multiple cases don't work for type switches 😕:

This could be much more succinct:
```go
	case int, int8, int16, int32:
	  return int64(n), nil
```

But I guess there's a good reason for it.

imiric (dismissed):
LGTM, but have you tested this with old archive files? Hopefully we won't run into any issues with old tests.

lib/types/types.go
na--: You can do that, but then `n` will be of an `interface{}` type, and you can't directly cast that to `int64`... You can do it via the `reflect` package, but that'd be much slower.

na--:
> LGTM, but have you tested this with old archive files? Hopefully we won't run into any issues with old tests.

Hmm good point, I'll see about adding a test. Not sure it's required, since the options in the archive bundles are basically the equivalent of the JSON+JS ones, but I'll dig around and test it manually, if nothing else. In any case, most old tests with duration values without units probably hadn't worked since k6 v0.27.0, since people would have gotten errors like `scenario default has configuration errors: the duration should be at least 1s, but is 1ms`.

na--:
I tested archives manually, and they are not a problem. If someone created an archive with k6 between v0.27.0 and v0.29.0, anything smaller than `duration: 1000000000` (1 billion nanoseconds is 1s) would have resulted in a validation error. And if they did it with a previous version, it would have executed, FWIW... :smile: But in both cases, when the duration value was serialized in the archive, it would have used stringy values, `1s` or `1ms` or whatever. So everything should be fine.

MStoykov (approved):
LGTM. 

This might not actually be breaking anything ... it will just now handle "10s" in http request timeouts?

imiric (approved):

na--:
> This might not actually be breaking anything ... it will just now handle "10s" in http request timeouts?

Basically, yes. And also treat integers ad milliseconds instead of nanoseconds everywhere, which hopefully no-one has ever consciously used.... :sweat_smile: I'll add a `breaking-changes` github label though, to remind us when we write the release notes.

Issue #532, opened 2018-03-14
Shared state/memory between VUs


robingustafsson:
We should evaluate possible ways to share larger blobs of data between VUs. The use case being that one might want to declare (or load from file on disk) a large data structure in the init context that is than used to parameterize data in the test.

Some ideas:
1. Share parts of the init context between VUs (not certain if this is at all doable in a clean way)
2. Create an API that is writable in the init context but only readable when running the test (executing the `export default function()`)

An API for 2) could perhaps be something like:
```js
import share from "k6/share";

// Parse data in to format specifc JS objects (CSV, JSON etc.)
let csvShare = share.load("./data.csv");
let jsonShare = share.load("./data.json");

export default function() {
    // Read data from the shared data blobs
    let row = csvShare.row();
    let randomRow = csvShare.random();
    let itemTitle = jsonShare.q("0.properties.title");
}
```

Globegitter:
Is it possible to have any shared state at the moment? We are interested in sharing a counter between the requests, to know the how many number of requests it took until a certain condition a has been met.

na--:
At the moment you can use a custom `Counter` metric (docs [here](https://docs.k6.io/docs/result-metrics#section-custom-metrics) and [here](https://docs.k6.io/docs/counter-k6metrics)) to track that at the end of the test, but there's no shared counter that you can use and check inside of the script, sorry. Depending on your use case, [thresholds](https://docs.k6.io/docs/thresholds) may be able to do the job, but if not, follow this issue for updates on any new shared state features.

na--:
This was resolved by https://github.com/loadimpact/k6/pull/1739. As described in the [release notes](https://github.com/loadimpact/k6/blob/master/release%20notes/v0.30.0.md#share-memory-between-vus-using-read-only-arrays-1739) (we don't quite have the docs updated yet), k6 now has a way to use a huge read-only array without the need to have a copy in every VU. We could do something similar for a map in the future, if needs be, but that would be a separate issue. And data segmentation (i.e. VUs independently iterating over non-overlapping elements of the array) falls under https://github.com/loadimpact/k6/issues/1539. So, I think this can be closed :tada:

Issue #1020, opened 2019-05-15
Better support for binary data


na--:
We currently don't have a good way to deal with binary data on the JS side yet :disappointed:. We rely on `[]byte` from the Go code, which translates to int arrays in JS, which is far from optimal. These are the places where we currently have binary data that I can think of:
- `open("somefile", "b")`
- HTTP response bodies when the request had `{ responseType: "binary" }` as params
- in `crypto.randomBytes()` (https://github.com/loadimpact/k6/pull/922)
- in the `crypto` hashing and digest functions (https://github.com/loadimpact/k6/pull/952)
- in the new PKI crypto functions (https://github.com/loadimpact/k6/issues/900)
- base64 encoding and decoding

We need a native analogue of the node.js [`Buffer`](https://nodejs.org/api/buffer.html), or better yet - to properly support ES [`ArrayBuffer`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer) objects. We actually have `ArrayBuffer`, the code below compiles and runs as you'd expect:
```js
export default function () {
    var buffer = new ArrayBuffer(16);

    // Create a couple of views
    var view1 = new DataView(buffer);
    var view2 = new DataView(buffer, 12, 4); //from byte 12 for the next 4 bytes
    view1.setInt8(12, 42); // put 42 in slot 12

    console.log(view2.getInt8(0));

    var z = new Int32Array(buffer, 0, 4);
    console.log(z);
};
```
but I'm not sure if this `ArrayBuffer` comes from the bundled core.js or from goja. It [seems](https://github.com/dop251/goja/blob/master/builtin_typedarrays.go) like the `ArrayBuffer` is from goja (though I [wouldn't rely on it](https://github.com/dop251/goja/blob/master/builtin_typedarrays_test.go) very much), but the `DataView` and `Int32Array` seem to be from core.js :confused: 

So we need to investigate exactly what the situation is and what we can use to handle binary data sensibly...

Related issues: https://github.com/loadimpact/k6/issues/856, https://github.com/loadimpact/k6/issues/874, https://github.com/loadimpact/k6/issues/873

na--:
A user found out that the current websocket API lacks support for binary messages: https://community.k6.io/t/converting-audio-file-into-bytes-and-send-over-web-socket/272/

[...]

PR #1739, opened 2020-11-25
shareable array implementation

MStoykov:
This is part of #532

The reason behind this not redefing ... how to open a file or load data
from one is that, as in the majority of the cases after that there is
some amount of ... transformation of the data, so that won't help much.

The original idea was to have a shareable JSON ... but the more concrete
array implementation is faster (in some cases by a lot) and also seems
like a better fix for the first iteration of this, as that will be the
majority of the data type. So it is very likely that the non array parts
will be dropped in favor of doing them later and this getting merged
faster.

There are still more things to be done and depending on whether or not
we want to support iterating (with for-of) it can probably be done
without some of the js code.

### DIFF
--- a/js/initcontext.go
+++ b/js/initcontext.go
@@ -73,7 +73,7 @@ type InitContext struct {
 
 	logger logrus.FieldLogger
 
-	shares     map[string]interface{}
+	shares     map[string]sharedArray
 	sharesLock *sync.Mutex
 }
 
@@ -92,7 +92,7 @@ func NewInitContext(
 		compatibilityMode: compatMode,
 		logger:            logger,
 
-		shares:     make(map[string]interface{}),
+		shares:     make(map[string]sharedArray),
 		sharesLock: new(sync.Mutex),
 	}
 }
@@ -127,87 +127,39 @@ func newBoundInitContext(base *InitContext, ctxPtr *context.Context, rt *goja.Ru
 
 // NewShare ...
 // TODO rename
-func (i *InitContext) NewShare(ctx context.Context, name string, call goja.Callable) goja.Value {
+// TODO constructor?
+func (i *InitContext) NewShare(name string, call goja.Callable) goja.Value {
 	i.sharesLock.Lock()
 	defer i.sharesLock.Unlock()
 	value, ok := i.shares[name]
-	rt := common.GetRuntime(ctx)
 	if !ok { //nolint:nestif
 		gojaValue, err := call(goja.Undefined())
 		if err != nil {
-			common.Throw(rt, err)
+			common.Throw(i.runtime, err)
 		}
-		if gojaValue.ExportType().Kind() == reflect.Slice {
-			// fmt.Println("it's a slice")
-			var tmpArr []interface{}
-			if err = rt.ExportTo(gojaValue, &tmpArr); err != nil {
-				common.Throw(rt, err)
-			}
+		// TODO this can probably be better handled
+		if gojaValue.ExportType().Kind() != reflect.Slice {
+			common.Throw(i.runtime, errors.New("only arrays can be made into shared objects")) // TODO better error
+		}
 
-			arr := make([][]byte, len(tmpArr))
-			for i := range arr {
-				arr[i], err = json.Marshal(tmpArr[i])
-				if err != nil {
-					common.Throw(rt, err)
-				}
-			}
-			value = sharedArray{arr: arr}
-		} else {
-			// fmt.Println("it's an object")
-			value = shared{gojaValue.ToObject(rt)}
+		// TODO this can probably be done better if we just iterate over the internal array, but ...
+		// that might be a bit harder given what currently goja provides
+		var tmpArr []interface{}
+		if err = i.runtime.ExportTo(gojaValue, &tmpArr); err != nil {
+			common.Throw(i.runtime, err)
 		}
+
+		arr := make([][]byte, len(tmpArr))
+		for index := range arr {
+			arr[index], err = json.Marshal(tmpArr[index])
+			if err != nil {
+				common.Throw(i.runtime, err)
+			}
+		}
+		value = sharedArray{arr: arr}
 		i.shares[name] = value
 	}
-	switch value.(type) {
-	case sharedArray:
-		// fmt.Println("wrapping array/slice")
-		// TODO cache this
-		cal, err := rt.RunString(`(function(val) {
-	var arrayHandler = {
-		get: function(target, property, receiver) {
-			// console.log("accessing ", property)
-			switch (property){
-			case "length":
-				return target.length()
-			case Symbol.iterator:
-				return function() {return target.iterator()}
-			/*
-			return function(){
-
-				var index = 0;
-				return {
-					"next": function() {
-						if (index >= target.length()) {
-							return {done: true}
-						}
-						var result = {value:target.get(index)};
-						index++;
-						return result;
-					}
-				}
-			}
-			*/
-			}
-			return target.get(property);
-		}
-	};
-	return new Proxy(val, arrayHandler)
-})`)
-		if err != nil {
-			common.Throw(rt, err)
-		}
-		call, _ := goja.AssertFunction(cal)
-		wrapped, err := call(goja.Undefined(), i.runtime.ToValue(common.Bind(i.runtime, value, i.ctxPtr)))
-		if err != nil {
-			common.Throw(rt, err)
-		}
-
-		return wrapped
-	case shared:
-		// TODO
-	}
-
-	return i.runtime.ToValue(common.Bind(i.runtime, value, i.ctxPtr))
+	return value.wrap(i.ctxPtr, i.runtime)
 }
 
 // Require is called when a module/file needs to be loaded by a script
@@ -259,7 +211,6 @@ func (i *InitContext) requireFile(name string) (goja.Value, error) {
 
 		if pgm.pgm == nil {
 			// Load the sources; the loader takes care of remote loading, etc.
-			// TODO: don't use the Global logger
 			data, err := loader.Load(i.logger, i.filesystems, fileURL, name)
 			if err != nil {
 				return goja.Undefined(), err
--- a/js/initcontext.go
+++ b/js/initcontext.go
@@ -22,10 +22,13 @@ package js
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"net/url"
 	"path/filepath"
+	"reflect"
 	"strings"
+	"sync"
 
 	"github.com/dop251/goja"
 	"github.com/pkg/errors"
@@ -69,6 +72,9 @@ type InitContext struct {
 	compatibilityMode lib.CompatibilityMode
 
 	logger logrus.FieldLogger
+
+	shares     map[string]interface{}
+	sharesLock *sync.Mutex
 }
 
 // NewInitContext creates a new initcontext with the provided arguments
@@ -85,6 +91,9 @@ func NewInitContext(
 		programs:          make(map[string]programWithSource),
 		compatibilityMode: compatMode,
 		logger:            logger,
+
+		shares:     make(map[string]interface{}),
+		sharesLock: new(sync.Mutex),
 	}
 }
 
@@ -110,7 +119,95 @@ func newBoundInitContext(base *InitContext, ctxPtr *context.Context, rt *goja.Ru
 		programs:          programs,
 		compatibilityMode: base.compatibilityMode,
 		logger:            base.logger,
+
+		shares:     base.shares,
+		sharesLock: base.sharesLock,
 	}
+}
+
+// NewShare ...
+// TODO rename
+func (i *InitContext) NewShare(ctx context.Context, name string, call goja.Callable) goja.Value {
+	i.sharesLock.Lock()
+	defer i.sharesLock.Unlock()
+	value, ok := i.shares[name]
+	rt := common.GetRuntime(ctx)
+	if !ok { //nolint:nestif
+		gojaValue, err := call(goja.Undefined())
+		if err != nil {
+			common.Throw(rt, err)
+		}
+		if gojaValue.ExportType().Kind() == reflect.Slice {
+			// fmt.Println("it's a slice")
+			var tmpArr []interface{}
+			if err = rt.ExportTo(gojaValue, &tmpArr); err != nil {
+				common.Throw(rt, err)
+			}
+
+			arr := make([][]byte, len(tmpArr))
+			for i := range arr {
+				arr[i], err = json.Marshal(tmpArr[i])
+				if err != nil {
+					common.Throw(rt, err)
+				}
+			}
+			value = sharedArray{arr: arr}
+		} else {
+			// fmt.Println("it's an object")
+			value = shared{gojaValue.ToObject(rt)}
+		}
+		i.shares[name] = value
+	}
+	switch value.(type) {
+	case sharedArray:
+		// fmt.Println("wrapping array/slice")
+		// TODO cache this
+		cal, err := rt.RunString(`(function(val) {
+	var arrayHandler = {
+		get: function(target, property, receiver) {
+			// console.log("accessing ", property)
+			switch (property){
+			case "length":
+				return target.length()
+			case Symbol.iterator:
+				return function() {return target.iterator()}
+			/*
+			return function(){
+
+				var index = 0;
+				return {
+					"next": function() {
+						if (index >= target.length()) {
+							return {done: true}
+						}
+						var result = {value:target.get(index)};
+						index++;
+						return result;
+					}
+				}
+			}
+			*/
+			}
+			return target.get(property);
+		}
+	};
+	return new Proxy(val, arrayHandler)
+})`)
+		if err != nil {
+			common.Throw(rt, err)
+		}
+		call, _ := goja.AssertFunction(cal)
+		wrapped, err := call(goja.Undefined(), i.runtime.ToValue(common.Bind(i.runtime, value, i.ctxPtr)))
+		if err != nil {
+			common.Throw(rt, err)
+		}
+
+		return wrapped
+	case shared:
+		// TODO
+	}
+
+	return i.runtime.ToValue(common.Bind(i.runtime, value, i.ctxPtr))
 }
 
 // Require is called when a module/file needs to be loaded by a script
--- a/js/initcontext.go
+++ b/js/initcontext.go
@@ -125,10 +125,10 @@ func newBoundInitContext(base *InitContext, ctxPtr *context.Context, rt *goja.Ru
 	}
 }
 
-// NewShare ...
-// TODO rename
-// TODO constructor?
-func (i *InitContext) NewShare(name string, call goja.Callable) goja.Value {
+// XShare is a constructor returning a shareable read-only array (general objects coming)
+// indentified by the name and having their contents be whatever the call returns - again only
+// arrays currently supported.
+func (i *InitContext) XShare(name string, call goja.Callable) goja.Value {
 	i.sharesLock.Lock()
 	defer i.sharesLock.Unlock()
 	value, ok := i.shares[name]
--- a/js/initcontext.go
+++ b/js/initcontext.go
@@ -22,13 +22,10 @@ package js
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
 	"net/url"
 	"path/filepath"
-	"reflect"
 	"strings"
-	"sync"
 
 	"github.com/dop251/goja"
 	"github.com/pkg/errors"
@@ -73,8 +70,7 @@ type InitContext struct {
 
 	logger logrus.FieldLogger
 
-	shares     map[string]sharedArray
-	sharesLock *sync.Mutex
+	sharedObjects *common.SharedObjects
 }
 
 // NewInitContext creates a new initcontext with the provided arguments
@@ -91,9 +87,7 @@ func NewInitContext(
 		programs:          make(map[string]programWithSource),
 		compatibilityMode: compatMode,
 		logger:            logger,
-
-		shares:     make(map[string]sharedArray),
-		sharesLock: new(sync.Mutex),
+		sharedObjects:     common.NewSharedObjects(),
 	}
 }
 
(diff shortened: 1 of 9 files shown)


js/initcontext.go
```diff
@@ -110,7 +119,95 @@ func newBoundInitContext(base *InitContext, ctxPtr *context.Context, rt *goja.Ru
...
+		// TODO cache this
+		cal, err := rt.RunString(`(function(val) {
+	var arrayHandler = {
+		get: function(target, property, receiver) {
+			// console.log("accessing ", property)
+			switch (property){
+			case "length":
+				return target.length()
+			case Symbol.iterator:
+				return function() {return target.iterator()}
```
MStoykov, line 173: This unfortunately can't be done through [goja.ProxyTrapConifg](https://pkg.go.dev/github.com/dop251/goja#ProxyTrapConfig) So if we want to have support for [for-of](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of) We need to use js proxy (we can use both, but I don't think that will help with anything :(

na--: Hmm why can't we use `goja.ProxyTrapConifg`? Because of `Symbol.iterator`? If so, can't we just get the goja definition of the iterator symbol from the runtime and compare against that in the `Get` property of the `goja.ProxyTrapConifg`?

MStoykov: Yes, 
I am gonna copy you the relevant part(which is most of it) of documentation that I linked ;) 

> Note that the Proxy may not have Symbol properties when using this as a handler because property keys are passed as strings. get() and getOwnPropertyDescriptor() for Symbol properties will always return undefined; has() and deleteProperty() for Symbol properties will always return false; set() and defineProperty() for Symbol properties will throw a TypeError. If you need Symbol properties implement the handler in JavaScript.

MStoykov: This will probably take some breaking of the API to be fixed and also the standard Symbols will need to be exported by goja .. which they currently aren't. 

I doubt this will actually make much of a performance difference to be honest ... But it will be kind of nice to not have to execute a script to wrap it I guess

MStoykov: Although i guess it is worth an issue to ask why the name wasn't a goja.Value to begin with instead of string

na--: We can't access the symbol from the Go side, but can't we get it from the JS side by just getting `Symbol.iterator` from the runtime, even now?

na--: Ah, nevermind, somehow missed the first of your 3 comments...

imiric (commented):
Nice work! It mostly LGTM, save for that wrapping script that uses `Proxy`. It would be great if that could be simplified/avoided, but I'm not familiar with goja enough to know if that would eventually be feasible with `ProxyTrapConfig` or with some other approach.

Strong :+1: for supporting `for...of`, and I think we should also support shareable objects even if it has worse performance than arrays. That will likely be easier to use in a lot of cases, but am also not sure how difficult it is to implement or if we'd be OK with postponing it for a second iteration of this, though probably yes. Maybe @sniku can chime in on this.

js/share_test.go
```diff
@@ -0,0 +1,152 @@
...
+)
+
+func TestInitContextNewShareable(t *testing.T) {
+	data := `'use strict';
+function generateArray() {
+    console.log("once");
+    var n = 50;
+    var arr = new Array(n);
+    for (var i = 0 ; i <n; i++) {
+        arr[i] = "something" +i;
```
imiric, line 43: Would the current implementation be usable for binary data, via `ArrayBuffer`, `TypedArray`, etc.?

While a plain array is definitely useful, I think a big use case will be sharing binary data as well. If this is currently possible, a test that showcases it would be helpful.

imiric (approved):
LGTM, just a few nitpicks. I guess it's OK if we leave shareable objects for later.

js/share.go
```diff
@@ -0,0 +1,110 @@
...
+
+	"github.com/dop251/goja"
+	"github.com/loadimpact/k6/js/common"
+)
+
+// TODO rename
+// TODO check how it works with setup data
+// TODO check how it works if logged
+type sharedArray struct {
+	arr [][]byte
```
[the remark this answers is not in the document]
MStoykov: I would expect goja will try to do some additional processing if the type I give it a slice ... which we don't want. And also this makes the golang syntax ... easier IMO ;)

js/share_test.go
```diff
@@ -0,0 +1,152 @@
...
+		i++;
+	}
+
+	if (s.something != undefined) {
+		throw "s.something should've been undefined but was " + s.something;
+	}
+
+	try {
+		s.something = 21
+		throw "the previous line should've errored s.something = 21"
```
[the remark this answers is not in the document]
MStoykov: I can fix that :P

js/share_test.go
MStoykov: Do you mean like the top array to be ArrayBuffer? I guess it will ... kind of work ... maybe, but will be pretty bad IMO. We should just try to get [SharedArrayBuffer](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SharedArrayBuffer) at some point.

If you ask for if it can be inside the array ... I guess - no, as I have to jsonify the data and then unjsonify it ...

Given how unusable ArrayBuffer is currently in k6 - I would argue this will be fixed when we work on it if it;s possible

imiric: Right, it would make sense to support this eventually once #1020 is fixed. And even `SharedArrayBuffer`, I wasn't aware of it 👍

js/initcontext.go
```diff
@@ -110,7 +119,47 @@ func newBoundInitContext(base *InitContext, ctxPtr *context.Context, rt *goja.Ru
...
+
+		shares:     base.shares,
+		sharesLock: base.sharesLock,
+	}
+}
+
+// NewShare ...
+// TODO rename
+// TODO constructor?
+func (i *InitContext) NewShare(name string, call goja.Callable) goja.Value {
```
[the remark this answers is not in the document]
MStoykov: I don't think it should be in a separate module, the same way `open` isn't. 
On the other hand this can be used inside the VU code .... it won't be performant, probably, but there really is no problem ... technically.
Also currently the error is a little bad if you try to use it as ... it is basically undefined, which is somewhat confusing, but so is `open`

js/share.go
```diff
@@ -0,0 +1,112 @@
...
+	"encoding/json"
+
+	"github.com/dop251/goja"
+	"github.com/loadimpact/k6/js/common"
+)
+
+// TODO rename
+// TODO check how it works with setup data
+// TODO check how it works if logged
+type sharedArray struct {
```
MStoykov, line 37: logging it will do:
```
      TypeError: Could not convert &{Object 0xc000dfa4b0 0xc0008a4750 false map[] [] 0 0 <nil>} to primitive
        at github.com/loadimpact/k6/js/common.Bind.func1 (native)
```
But I don't know ... what should be logged so :shrug: 
And passing it by setup ... doesn't work .. but it also doesn't panic - it passes some Object that seems ... empty - this can probably be fixed and made to work, but it will be harder with the cloud/distributed in mind.

js/initcontext.go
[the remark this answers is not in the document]
imiric: OK, it's fine as is too.

js/share.go
imiric: Having it work from `setup()` would be good, but I don't think it's required for this PoC.

That logging error looks gnarly though... We could log the first and last N elements stringified, if that's easily doable.

js/initcontext.go
```diff
@@ -110,7 +119,47 @@ func newBoundInitContext(base *InitContext, ctxPtr *context.Context, rt *goja.Ru
...
+
+		shares:     base.shares,
+		sharesLock: base.sharesLock,
+	}
+}
+
+// XShare is a constructor returning a shareable read-only array (general objects coming)
+// indentified by the name and having their contents be whatever the call returns - again only
+// arrays currently supported.
+func (i *InitContext) XShare(name string, call goja.Callable) goja.Value {
```
na--, line 131: As I mentioned in slack, I'd really like not to have new k6-specific superglobal names like `open()`, `__VU`, etc. Instead, this should be accessible via `k6/data/SharedArray` (so, `import {SharedArray} from "k6/data";`) or something like that. The `k6/data` package will probably be reused by https://github.com/loadimpact/k6/issues/1539 for other things like segmented iterators and maybe XML/JSON/etc. streaming parsers and segmentation (though those probably will be sub-packages)

js/initcontext.go
```diff
@@ -110,7 +119,47 @@ func newBoundInitContext(base *InitContext, ctxPtr *context.Context, rt *goja.Ru
...
+		// TODO this can probably be done better if we just iterate over the internal array, but ...
+		// that might be a bit harder given what currently goja provides
+		var tmpArr []interface{}
+		if err = i.runtime.ExportTo(gojaValue, &tmpArr); err != nil {
+			common.Throw(i.runtime, err)
+		}
+
+		arr := make([][]byte, len(tmpArr))
+		for index := range arr {
+			arr[index], err = json.Marshal(tmpArr[index])
```
na--, line 154: This is good enough for the first version, but I wonder if we couldn't figure out something better. For example, gob (or some other binary encoding/decoding) is probably faster. 

Or maybe no encoding and decoding is necessary at all, and we can share the raw Go objects and use [`Object.freeze()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/freeze) to make it read-only and prevent data races? goja [seems to support it](https://github.com/dop251/goja/blob/6060b0671c09a09fbe8a33e29dfdbae33f85a640/builtin_object.go#L261-L287), and it even seems to work:
```js
export default function () {
    const obj = [1, 2, 3];
    obj[1] = 22;
    Object.freeze(obj); // comment this for no error
    console.log(obj);
    obj[2] = 33; // this throws a TypeError if frozen
    console.log(obj);
}
```

Even if it's not perfect and there's some obscure way to modify the object, we'll document that the shared data is read-only, so it'd be fine.

js/initcontext.go
na--: btw now it should be easy to move this outside of the `InitContext`, you can just have the shares registry in the `InitEnvironment` that was added in https://github.com/loadimpact/k6/pull/1706

js/initcontext.go
```diff
@@ -69,6 +72,9 @@ type InitContext struct {
 	compatibilityMode lib.CompatibilityMode
 
 	logger logrus.FieldLogger
+
+	shares     map[string]sharedArray
+	sharesLock *sync.Mutex
```
na--, line 77: Somewhat connected to https://github.com/loadimpact/k6/pull/1739#discussion_r540121267, maybe this should be more generic. Basically, a way VUs can share all sorts of objects between them in a `sync.Once` fashion, with each object being referenced by its string name/path. So, basically:
- move this to https://github.com/loadimpact/k6/blob/master/js/common/initenv.go#L33 and leave it private
- use `sync.Map` instead of this custom one
- and have generic `GetOrCreateShare(name string, createCallback func() interface{}) interface{}` method 

In the `SharedArray` constructor you can easily type assert that the value you get is a sharedArray, but extensions would be easily able to use it for other purposes as well, e.g. shared atomic counters or whatever.

js/initcontext.go
MStoykov: My problem is that this freezes only the top object not any object that is property of the top object. Also, goja does sometimes have an internal state for even simple types that change without any locking or stuff ... because JS is single-threaded by definition. 

So IMO this will break horribly, but was one of the initial things I thought about ... also this means that I get an object from 1 VM and put it in another, without any kind of checking that it doesn't for example have functions ... that modify the original VMs properties or something ... as a whole I think while this approach will probably be "faster", unless goja has support for this it will probably be way too much work to actually properly check that things won't blow up, and any change to goja will likely necessitate refactoring. 

Maybe gob will be a better fit - maybe we should try it for a second iteration ... I mostly went directly for JSON as we already do this for the setupData ... and also ... JS(ON) ;)

js/initcontext.go
MStoykov: earlier it was map[string]interface{} so ... I can use it for the sharedObject ... but I decided to drop it and leave it to be extended when we actually need to ... So I am again for just leaving it until we actually need it, and at that point whoever does the change can better gauge if that is still better

js/initcontext.go
```diff
@@ -110,7 +119,47 @@ func newBoundInitContext(base *InitContext, ctxPtr *context.Context, rt *goja.Ru
...
+}
+
+// XShare is a constructor returning a shareable read-only array (general objects coming)
+// indentified by the name and having their contents be whatever the call returns - again only
+// arrays currently supported.
+func (i *InitContext) XShare(name string, call goja.Callable) goja.Value {
+	i.sharesLock.Lock()
+	defer i.sharesLock.Unlock()
+	value, ok := i.shares[name]
+	if !ok { //nolint:nestif
```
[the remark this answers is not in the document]
MStoykov: predominantly because earlier the code at the end was way longer ... but the nolint is no longer necessary actually .. so maybe the linter checking the unneeded nolints ... has a bug :scream:

js/initcontext.go
na--: > until we actually need it

I gave you an example of how it can be useful - extensions would be able to share state between al VUs, for example as a more generic version of https://github.com/MStoykov/xk6-counter

And given that it doesn't really cost you anything to switch it, I don't see a reason not to. And sync.Map probably is much better for this use case anyway:
>  The Map type is optimized for two common use cases: (1) when the entry for a given key is only ever written once but read many times, as in caches that only grow, or (2) when multiple goroutines read, write, and overwrite entries for disjoint sets of keys. In these two cases, use of a Map may significantly reduce lock contention compared to a Go map paired with a separate Mutex or RWMutex.

js/initcontext.go
na--: hmm didn't realize `Object.freeze()` only touches the first layer, but even that's easy to work around. Copying the public domain `deepFreeze()` from [here](https://stackoverflow.com/questions/34776846/how-to-freeze-nested-objects-in-javascript) (or, originally, [here](https://github.com/substack/deep-freeze/blob/master/index.js)) seems to work well enough:

```js
function deepFreeze(o) {
    Object.freeze(o);
    if (o === undefined) {
        return o;
    }

    Object.getOwnPropertyNames(o).forEach(function (prop) {
        if (o[prop] !== null
            && (typeof o[prop] === "object" || typeof o[prop] === "function")
            && !Object.isFrozen(o[prop])) {
            deepFreeze(o[prop]);
        }
    });

    return o;
};

export default function () {
    const obj = [1, [1, 2], { c: 3, d: 4 }];
    deepFreeze(obj); // comment this for no error
    console.log(JSON.stringify(obj));
    obj[1].push(3); // this throws a TypeError if frozen
    obj[2].f = 5; // and this also
    console.log(JSON.stringify(obj));
}
```

And we already have a JS wrapper anyway, why not go all the way? :sweat_smile:

MStoykov: > Also, goja does sometimes have an internal state for even simple types that change without any locking or stuff ... because JS is single-threaded by definition.

So I don't think this is a good idea - we can explore it in a second iteration ;)

na--: > Also, goja does sometimes have an internal state for even simple types that change without any locking or stuff ... because JS is single-threaded by definition.

Sorry, I didn't address this properly. I know about the internal state, when you pass a goja `Runtime` some Go value, it wraps it in its own internal types. And that should be fine, since it doesn't touch the original value, as far as I know. At most it'll make a copy of a string to convert it to utf-16 or something like that...

So I think it's worth it to try using `Object.freeze()` (or, rather, `deepFreeze()`) even now, in the first iteration. It's going to be fine, even if it doesn't work at 100% and some JS guru can get around it and still modify something. They are doing it on their own risk of a data race... Besides, it's not like we haven't had issues with transforming objects to JSON and back before - we've had different key names and Go strangeness with that as well. And we do it for `setup()` because we have to ship the data between instances in the cloud. Here the requirements are much different.

js/initcontext.go
MStoykov: see e94190c as the comment says ... using sync.Map isn't really useful :(

na--: :+1: yeah, makes sense when it can't accept a callback... :disappointed: and we can't even use a `RWMutex` because of that behavior... oh well :man_shrugging:

js/initcontext.go
```diff
@@ -69,6 +69,8 @@ type InitContext struct {
 	compatibilityMode lib.CompatibilityMode
 
 	logger logrus.FieldLogger
+
+	sharedObjects *common.SharedObjects
```
na--, line 73: I don't think this needs to be here. You should be able to just create it via `common.NewSharedObjects()` in `js/bundle.go`, in the `initenv := &common.InitEnvironment{...}`. We need to eventually remove the needlessly special `InitContext` struct anyway, no need to add things to it that it doesn't need.

na--: Ah, nevermind, ignore me, that would create a copy per VU... :disappointed:

js/initcontext.go
na--: I don't completely understand everything Dmitry says, but it seems like this won't be feasible... :disappointed: 

One benefit of recursively freezing objects instead of copying them (i.e. marshaling and unmarshaling), besides improved performance, would have been the better UX. With the current solution, each runtime will get its own copy of the array item, and would seeming be able to make changes in it. But if it accesses the same array item the next time, the changes it made on the previous iteration will be gone, since it'd receive a fresh copy of the data... Not the end of the world, but unless we document it very well, someone is liable to spend a bit of time pulling their hair. So, maybe it's worth recursively `Object.freeze()` things even when we're copying the data on the Go side?

MStoykov:
Some [benchmarks](https://gist.github.com/MStoykov/ea62b0d16dc21261779f46db0fe8bd53) with the latest code

js/share_test.go
```diff
@@ -0,0 +1,173 @@
...
+		}
+	}
+
+	try {
+		s[1]= "21";
+
+		throw "the previous line should've errored";
+	} catch(e) {
+		if (!e.toString().includes("Host object field 1 cannot be made configurable")) {
+			throw "wrong error " + e.toString();
```
imiric, line 128: I would prefer to see separate Go testcases for each scenario, rather than this large JS script. It would be easier to run them individually and also to modify or add new ones.

js/share_test.go
```diff
@@ -0,0 +1,173 @@
+/*
```
imiric, line 1: This file should be in `js/modules/k6/data`.

imiric (changes requested):
Mostly LGTM, just some minor suggestions.

Also:

- I'm out of my depths with the `Object.freeze` discussion, but shouldn't we take [Dmitry's advice](https://github.com/dop251/goja/issues/232#issuecomment-745354746) and use the Go API rather than wrap it in JS? It would probably complicate things a bit, but it should be worth it if it considerably increases performance.

- The benchmark in that gist shows the share test actually using more memory than the no-share test(?). I can understand the CPU increase, but isn't the whole point of this to reduce memory usage?

js/share_test.go
MStoykov: I tried this early when I moved it to a package, but I can't really run this whole test and I mostly care about the part that it is accessible by other VUs when you just run a test ... which is really hard to do in a js/modules test

js/share_test.go
MStoykov: Hm ... maybe, it used to be fairly short. On the other hand that will increase the amount of time the test takes(goja initialization isn't free even without the babel and corejs), while changes to this test seem ... unlikely unless we add functionality and I don't think we will be adding all that much as I doubt a lot of users will want additional Array like functionality.  


I am even starting to question if enabling for-of was worth it ...

MStoykov:
@imiric :
* I thought about this some more, and unless I drop the for-of part ... I still need to go through javascript code, at which point I am pretty sure the difference he is showing will be even less noticeable. The benchmark I posted was mostly to compare the current solution to the "native" performance(of just a JS array) and see if there is enough of a difference, and I don't think there is. You can read my long [comment](https://github.com/dop251/goja/issues/232#issuecomment-745210887) on the current performance characteristics and my opinion on whether optimizing it .. matters.  I still think that it is better to have this as is and come back instead of continuing to tinker with it up until we need to release v0.30.0.

* the benchmark I linked was about access times, not memory ... I specifically don't load the whole array in the first case as If I did it will take around 90-100mb per VU as I explained [here](https://github.com/dop251/goja/issues/232#issuecomment-745210887) and I don't have that kind of memory. Sorry for the misunderstanding - the non shared test has 10 elements the shared one has 100k elements in the array ;)

MStoykov:
An interesting use case for this that I just thought of so I am just writing here - we can move it to issue if this gets merged and we think there is merit:
It should be possible to generate a shareable object in `setup` using the response from a request (and not only) and then being able to use it from each VU. This will need some additional work ... due to the way this works in the cloud.

An alternative (that will likely be better) is to make it possible for `setup` to return SharedArray and for it to ... keep being such, which again will need some work for cloud/distributed execution. This will also probably need to work if it returns an object/array that has SharedArray(s) in it.

js/share_test.go
imiric: Hhmm right, you need `getSimpleRunner`, etc. OK, it's not a big deal.

js/share_test.go
imiric: I would still prefer legibility/usability over a slightly longer run time, but it's fine if you don't think it's worth it.

I saw your comment about the 100-200x performance penalty with the iterator, and now I'm not so sure if we should keep it either 😄 I would assume access by index will be more common, and looping by index would work as well, but some users might prefer the `for-of` notation and consider the performance loss acceptable(?). We need documentation to explain this either way, so I'm leaning towards leaving it in as is.

js/modules/k6/data/data.go
```diff
@@ -0,0 +1,93 @@
...
+	if err != nil {
+		common.Throw(rt, err)
+	}
+	obj := gojaValue.ToObject(rt)
+	if obj.ClassName() != "Array" {
+		common.Throw(rt, errors.New("only arrays can be made into SharedArray")) // TODO better error
+	}
+	arr := make([]string, obj.Get("length").ToInteger())
+
+	cal, err := rt.RunString(`(function(input, output) {
```
na--, line 79: Can you explain (and maybe add as a code comment here) why we're encoding the data to JSON on the JS side? `JSON.stringify()` does something magical that `json.Marshal()` doesn't? Or does this save us from utf8<->utf16 conversions?

na-- (dismissed):
I haven't played with this since mid-December, but the code LGTM in general. 

Having the test in the same module will be better, but if that's annoying to do because of import loops or something like that, maybe you can move just the small `try/catch` tests in there, as discrete unit tests for error handling and maybe a small one for sanity check. But leave the big test where it is, as an integration test of sorts?

js/modules/k6/data/data.go
MStoykov: The commit comment explains it :P. Using JSON.parse was the easiest way to get a goja object thing that can then be frozen as otherwise `json.Decode` gets you a golang map which ... just isn't a goja object. You can make it into a value but it still the golang map underneath and doesn't let itself be frozen (the goja author actually commented on that). 
So at that point, I had the question - do I use two different libraries to do the encode/decode or not, and went with "not" ;).

They have some ... interesting [behavior differences around stuff like functions and other strange types](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify).

But IMO it is much better to be as JS like as possible - explaining that what we do is JSON.parse/JSON.stringify a value instead of json.Encode/Decode in golang and that is why your strange value doesn't work will be again easier for users. Also possibly make it possible for them to make it work.

Another benefit is this way I iterate only once here instead of twice, or once but use string indexes and `Object.Get` :scream:.

na--: Please add a code comment with that, since I doubt anyone will read the commit description in the future and might try to "optimize" it

MStoykov: That's the problem .. people don't read git commits :sob:

na-- (approved):

imiric (approved):
LGTM! Noticed a couple of typos, but it's not a big deal.

The `goheader` linter failures are because they're new files? 😕

MStoykov:
> The goheader linter failures are because they're new files? 

No idea ... I had some golangci-lint's cache problems locally, but they seemed to go away. Maybe we are hitting something similar in github actions ...

PR #1742, opened 2020-11-28
Fix break line after password input (#1703)

### LINKED ISSUE #1703: There is no new line after `password` prompt for login command
The screenshot shows it very clearly :). 
image

I expect that the reading of the password just swallows the newline as well so it doesn't get printed ?

paroar:
Closes #1703

### DIFF
--- a/ui/form_fields.go
+++ b/ui/form_fields.go
@@ -22,6 +22,7 @@ package ui
 
 import (
 	"bufio"
+	"fmt"
 	"io"
 	"os"
 	"strings"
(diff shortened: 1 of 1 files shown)


na-- (changes requested):
Thanks for making this PR, however the `PasswordField` doesn't seem to be the right place to make this change... 

See how the `ui.Form` works, the `Field` objects don't write things, only the `Form` does: https://github.com/loadimpact/k6/blob/40979fd751999415e6f00de16ce4fedd22a838aa/ui/form.go#L30-L48

And it has an `io.Writer` injected, so we certainly should try to avoid using a global `fmt.Print()` call and use that writer instead (e.g. use `fmt.Fprint(w, "\n")` instead).

PR #1744, opened 2020-11-29
Dont skip host header

### LINKED ISSUE #1740: Http response.request.headers are missing "host" value
If you set a value for "host" in the header of a http.post request, that value gets stripped from the response-objects .request.header property. Minor issue but I was asked to report it.

## Environment
k6 v0.28.0 ((devel), go1.14.9, windows/amd64)

```
import http from "k6/http";

export default function main() {
    let r = http.post("https://localhost", "body", { headers: { "foo": "bar", "host": "https://some.domain.com" } });
    console.log(JSON.stringify(r.request.headers));
}

```

noelzubin:
Closes #1740

### DIFF
--- a/js/modules/k6/http/request.go
+++ b/js/modules/k6/http/request.go
@@ -286,12 +286,10 @@ func (h *HTTP) parseRequest(
 				}
 				for _, key := range headers.Keys() {
 					str := headers.Get(key).String()
-					switch strings.ToLower(key) {
-					case "host":
+					if strings.ToLower(key) == "host" {
 						result.Req.Host = str
-					default:
-						result.Req.Header.Set(key, str)
 					}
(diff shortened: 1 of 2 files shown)


na-- (changes requested):
Thanks for making this PR, @noelzubin. Can you please also add some tests? Specifically:
- that the `host` header is accessible from `response.request.headers` (the problem exposed in https://github.com/loadimpact/k6/issues/1740)
- how k6 will handle situations in which the `host` header is different from the URL that is being accessed
- (optional) how this interacts with the [global k6 option `hosts`](https://k6.io/docs/using-k6/options#hosts)

na-- (approved):

MStoykov (approved):

PR #1746, opened 2020-11-30
Sign Windows release binary and installer

### LINKED ISSUE #1034: Sign release binaries and installers for Windows

imiric:
Closes #1034

I tested this in my fork and you can download the signed MSI from [here](https://github.com/imiric/k6/actions/runs/385130865). As explained in [this comment](https://github.com/loadimpact/k6/issues/1034#issuecomment-734268724), I chose to not use any 3rd-party extensions to avoid any security issues. As noted there, this won't remove the "Windows protected your PC" warning, but should eventually if we reach a certain "trust" level with Microsoft, which is gained by a lot of people choosing to run the app anyway.

Two secrets were added to the GitHub repo for this: `WIN_SIGN_CERT` and `WIN_SIGN_PASS`. Information about both can be found in the password manager.

Note that currently this only signs the binary in the MSI package and the MSI itself that are uploaded to Bintray. The binary that is uploaded to GitHub and available as release assets is not signed, as that would be more convoluted to setup, since `signtool.exe` is Windows-only and the build stage happens on Linux. I suppose we could overwrite the assets from Windows, but I didn't test this.

### DIFF
--- a/.github/workflows/all.yml
+++ b/.github/workflows/all.yml
@@ -311,7 +311,7 @@ jobs:
         run: echo "${env:ProgramFiles(x86)}\Windows Kits\10\bin\x64" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
       - name: Convert base64 certificate to PFX
         run: |
-          $bytes = [Convert]::FromBase64String("${{ secrets.CERTIFICATE }}")
+          $bytes = [Convert]::FromBase64String("${{ secrets.WIN_SIGN_CERT }}")
           [IO.File]::WriteAllBytes("k6.pfx", $bytes)
       - name: Sign Windows binary
         run: signtool sign /f k6.pfx /p "${{ secrets.WIN_SIGN_PASS }}" /tr "http://timestamp.digicert.com" /td sha256 /fd sha256 "packaging\k6.exe"
--- a/.github/workflows/all.yml
+++ b/.github/workflows/all.yml
@@ -308,7 +308,7 @@ jobs:
           move .\packaging\k6-$env:VERSION-win64\k6.exe .\packaging\
           rmdir .\packaging\k6-$env:VERSION-win64\
       - name: Add signtool to PATH
-        run: echo "C:\Program Files (x86)\Windows Kits\10\bin\10.0.17763.0\x86" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
+        run: echo "${env:ProgramFiles(x86)}\Windows Kits\10\bin\x64" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
       - name: Convert base64 certificate to PFX
         run: |
           $bytes = [Convert]::FromBase64String("${{ secrets.CERTIFICATE }}")
--- a/.github/workflows/all.yml
+++ b/.github/workflows/all.yml
@@ -324,6 +324,8 @@ jobs:
           light.exe -ext WixUIExtension k6.wixobj
       - name: Sign MSI package
         run: signtool sign /f k6.pfx /p "${{ secrets.WIN_SIGN_PASS }}" /tr "http://timestamp.digicert.com" /td sha256 /fd sha256 "packaging\k6.msi"
+      - name: Cleanup signing artifacts
+        run: del k6.pfx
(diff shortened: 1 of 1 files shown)


.github/workflows/all.yml
```diff
@@ -307,13 +307,23 @@ jobs:
           Expand-Archive -Path ".\dist\k6-$env:VERSION-win64.zip" -DestinationPath .\packaging\
           move .\packaging\k6-$env:VERSION-win64\k6.exe .\packaging\
           rmdir .\packaging\k6-$env:VERSION-win64\
+      - name: Add signtool to PATH
+        run: echo "C:\Program Files (x86)\Windows Kits\10\bin\10.0.17763.0\x86" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
```
imiric, line 311: This static path is concerning if it ever changes, but I haven't found a way to determine it dynamically. :-/

[This GitHub Action](https://github.com/dlemstra/code-sign-action/blob/master/index.ts#L15) also hardcodes it.

MStoykov (dismissed):

na-- (changes requested):
Interesting that signtool is an x86 program, and we're using it to sign an x64 executable... :confused: :man_shrugging: 

In any case, I have two requests... Can you please delete `k6.pfx` at the end of the task, just in case our VM is not properly cleaned up between builds? And also rename the `CERTIFICATE` secret to something more descriptive, `WIN_SIGN_CERTIFICATE` or something like that? It's not unlikely that we'll want to add other certificates in the CI process in the future.

.github/workflows/all.yml
imiric: I managed to get a more robust `${env:ProgramFiles(x86)}\Windows Kits\10\bin\x64` working (4353e64b), and also use the x64 executable @na--.

imiric:
@na-- I changed the secret to `WIN_SIGN_CERT` and added removal of the `k6.pfx`.

I haven't tested a full build yet, but I can do that tomorrow, or we can merge this as is since I'll test a full deployment before the next release anyway.

na-- (approved):
LGTM

MStoykov (approved):

na--:
Hmm @imiric I think that we should also upload the signed .msi package to the github releases as well... As a separate PR of course, but if we have the .msi in there as well, it will probably make it easier to implement the `/latest` link in k6.io that would redirect to it

imiric:
@na-- Sure, makes sense. Can you create a new issue for that, since merging this will close #1034? Or we can keep that open and mention the upload. I'll likely do that while testing the full deployment, and fix the checksum issue as well...

na--:
I'll create a new one, merge this and close the new one :shipit: :sweat_smile:

PR #1747, opened 2020-11-30
Add tc39/test262 tests in k6

MStoykov:
https://github.com/tc39/test262/blob/main/INTERPRETING.md Should probably be used as reference ... about what the test metadata (and not only) means


.github/workflows/all.yml
```diff
@@ -83,6 +83,28 @@ jobs:
           fi
           go test "${args[@]}" -timeout 800s ./...
 
+  tc39:
+    runs-on: ubuntu-latest
+    steps:
```
imiric, line 88: Do we want to run this on every commit? We might be OK with just running them on each release (version tag), especially if they're flaky, and we should probably exclude it from the requirements for a successful build.

MStoykov: 1. there are currently no requirements (I think)
2. they aren't really flaky ... just they sometimes take too much time - predomenantly when I increase them or decide to try to enable parallel execution ... OR maybe they take too much memory :man_shrugging: 
3. In the end we might move them to not being run on eveyr commit :D, but currently they are in development :P

.github/workflows/all.yml
```diff
@@ -83,6 +83,28 @@ jobs:
...
+          go-version: 1.15.x
+      - name: Run tests
+        run: |
+          set -x
+          cd js/tc39
+          mkdir -p TestTC39/test262
+          cd TestTC39/test262
+          git init
+          git remote add origin https://github.com/tc39/test262.git
+          git fetch origin --depth=1 72154b17fc99a26e79b2586960f059360d4ce43d
```
na--, line 103: What is so special about this commit? And if you don't want to use `master` because of potential future flakyness, can we have this commit ID as an environment variable, so it'd be a bit more visible and easier to update.

.github/workflows/all.yml
na--: We can move this to its own job, and have it run only when files in `js` were changed with https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestpaths

That said, if it takes less than 10 minutes, this is probably good enough...

js/tc39/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+This is obviously a modified version of [the code in the goja
+repo](https://github.com/dop251/goja/blob/master/tc39_test.go)
+
+
+# Reasons for recording breaking_test_errors.json
+
+Unfortunately k6 doesn't pass all the test that are currently defined as "interesting" and probably
+won't even more. Goja decided to just not run the ones that it knows it fails currently, but this
+means that if they stop fail, someone needs to go re-enable that test. This also means that if the
```
na--, line 31:
```suggestion
means that if they stop failing, someone needs to go re-enable these tests. This also means that, if the
```

js/tc39/tc39_test.go
```diff
@@ -0,0 +1,562 @@
...
+		} else if strings.HasSuffix(file.Name(), ".js") && !strings.HasSuffix(file.Name(), "_FIXTURE.js") {
+			ctx.runTest(newName, func(t *testing.T) {
+				ctx.runTC39File(newName, t)
+			})
+		}
+	}
+}
+
+func TestTC39(t *testing.T) {
+	if testing.Short() {
```
na--, line 520: I had no idea this existed... And I'm not sure I even like it, given the existence of build flags. but oh well... :man_shrugging: On the other hand, I :heart: the skip message below, and how we can optionally include the tc39 tests in our normal golang tests, or not!

na-- (dismissed):
I don't have time to dive into the nitty-gritty details (e.g. `tc39TestCtx.runTC39Test()` and how exactly the TC39 test suite is structured), but overall I quite like how you've structured the tests on the k6 side :heart: 

@MStoykov, what do you think about exporting the `ctx.errors` to another JSON file, and saving it as a build artifact in github actions, not just printing it to `stdout`? That way it'd be easier to diff the changes when there are issues, and it shouldn't be too difficult to do.

.github/workflows/all.yml
MStoykov: This was the latest commit at the time I started :D

MStoykov:
> @MStoykov, what do you think about exporting the ctx.errors to another JSON file, and saving it as a build artifact in github actions, not just printing it to stdout? That way it'd be easier to diff the changes when there are issues, and it shouldn't be too difficult to do.

My current way of doing it is to `echo {} > breaking_test_errors.json` and then get the json out of the output ... I would like to dump what the `breaking_test_errors.json` should be so the test pass ... but this will require some more monkeying with the ctx.errors as it currently only records what isn't in ctx.expectedErrors ... and I will need some way to merge them together - probably not hard, but I really prefer not to have to run the test 2-10 times again to check out that I did right. 

I also do intend on coming back to enable/disable more tests and hopefully reorganize it so we can have proper concurrent execution ... and probably some memory usage reduction, so this will just be ... one more thing to fix ... next year :P

imiric (changes requested):
Fantastic work @MStoykov!!! 👏 

I saw some commented out code in places, but there's probably a good reason you left it there, so it's fine.

I would echo [Ned's suggestion](https://github.com/loadimpact/k6/pull/1747#discussion_r539254768) to only run this if files in `js` change, or even on `master` commits and version tags only. We don't really need it for every PR, even if it takes less than 10 minutes.

js/tc39/checkout.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+sha=72154b17fc99a26e79b2586960f059360d4ce43d # this is just the commit it was last tested with
+mkdir -p ./TestTC39/test262
+cd ./TestTC39/test262
```
imiric, line 4: Can you add `/js/tc39/TestTC39` to `.gitignore`?

MStoykov:
The workflows change seems to work https://github.com/loadimpact/k6/pull/1764 
It will run on every master push and tags and on PRs that changes anything in the `js` directory

imiric (approved):
Just squash/rebase.

PR #1749, opened 2020-11-30
Fix break line after password input (loadimpact#1703)

### LINKED ISSUE #1703: There is no new line after `password` prompt for login command

paroar:
Sorry, I deleted the branch instead of update the PR and now is duplicated. Anyway thanks for the pointers before.

Closes #1703

### DIFF
--- a/ui/form.go
+++ b/ui/form.go
@@ -65,9 +65,13 @@ func (f Form) Run(r io.Reader, w io.Writer) (map[string]interface{}, error) {
 
 			color.Set(color.FgCyan)
 			s, err := field.GetContents(r)
-			if displayLabel == "Password" {
-				fmt.Fprint(w, "\n")
+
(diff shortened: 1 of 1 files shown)


ui/form.go
```diff
@@ -65,6 +65,9 @@ func (f Form) Run(r io.Reader, w io.Writer) (map[string]interface{}, error) {
 
 			color.Set(color.FgCyan)
 			s, err := field.GetContents(r)
+			if displayLabel == "Password" {
```
na--, line 68: This would be better off as a [type assertion](https://tour.golang.org/methods/15), ensuring that `field` is of type `PasswordField`.

na-- (approved):

imiric (approved):
Thanks for this!

PR #1754, opened 2020-12-03
Encode arrays correctly in query params.

noelzubin:
Closes https://github.com/loadimpact/k6/issues/1713


na-- (commented):
Thanks for making this PR. I guess picking to reuse the same key for all slice values (instead of `key[]` or `key.<N>` or any other convention) is better than the current behavior (https://github.com/loadimpact/k6/issues/1713, https://github.com/loadimpact/k6/issues/878)...

Can you please add some tests in your PR? And also you probably don't need reflection, you should be able to get by with a simple [type assertion](https://tour.golang.org/methods/15) that `v` is `[]interface{}`?

noelzubin:
@na-- I dont think go allows type assertion on []interface{}, Although it allows type assertions on []string though.

na--:
Ah, yeah, fair enough... :man_facepalming: Though the rest of my request still stands - please add some tests, and also you probably don't need to check for `reflect.Array` in the `switch`, which should take care of the lint error as well.

imiric (changes requested):
@noelzubin Thanks for the PR. Are you available to finish it? We'd like to have this merged for v0.33.0. I think it would just need a test and maybe disabling the `exhaustive` linter if we can't avoid the reflection.

If we don't hear from you by sometime next week or if you're not available, that's OK too, and we can proceed to do the leftover changes.

imiric:
Fixed in #2060.

Thanks for your contribution @noelzubin! We'll make sure to thank you in the release notes :)

PR #1756, opened 2020-12-06
Prevent panic when slicing end is higher than nodes array size

asettouf:
Hi, 

first time contributing on this project (which I find lovely), so let me know if there is anything I can make to improve this PR.

Solves https://github.com/loadimpact/k6/issues/1487

### DIFF
--- a/js/modules/k6/html/html.go
+++ b/js/modules/k6/html/html.go
@@ -398,7 +398,10 @@ func (s Selection) Map(v goja.Value) []string {
 }
 
 func (s Selection) Slice(start int, def ...int) Selection {
-	if len(def) > 0 {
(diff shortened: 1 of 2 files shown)


na-- (approved):
Thanks!

asettouf:
@na-- I'm just a bit surprised I'm decreasing the coverage. Should I add more tests for _html.go_, on the `Slice` function?

imiric (approved):
LGTM, thanks for the fix!

I would've preferred a more unit-focused test to avoid testing the `js.Runner` for this, but it's not a big deal.

As for the Codecov report, don't worry about it. It's frequently wrong and we want to replace it with something better, but haven't gotten around to it. 😞

PR #1757, opened 2020-12-06
feature/adding application insights output plugin

polatengin:
Allow using Application Insights as an output target, so you can use `--output appinsights` to fork k6 outputs to Application Insights.


stats/appinsights/config.go
```diff
@@ -0,0 +1,67 @@
...
+
+// ParseArg takes an arg string and converts it to a config
+func ParseArg(arg string) (Config, error) {
+	c := Config{}
+
+	pairs := strings.Split(arg, ",")
+	for _, pair := range pairs {
+		r := strings.SplitN(pair, "=", 2)
+		if len(r) != 2 {
+			return c, fmt.Errorf("couldn't parse %q as argument for csv output", arg)
```
rguthriemsft, line 56: This error message looks incorrect, should it be "for appinsights output" instead of csv output.

stats/appinsights/config.go
```diff
@@ -0,0 +1,67 @@
...
+		}
+		switch r[0] {
+		case "instrumentation_key":
+			c.InstrumentationKey = null.StringFrom(r[1])
+		default:
+			return c, fmt.Errorf("unknown key %q as argument for csv output", r[0])
```
rguthriemsft, line 62: same as above.

rguthriemsft (changes requested):
Need to add an appinsights_test.go file with test for the collector.

mostafa:
Hey @polatengin,

thanks for the output and the PR :pray:. I've done acceptance testing and it works fine. Yet, there are three things you might do before the PR is mergeable, which should be decided by the k6 Team (not me), considering the roadmap and other related things:

1. Please consider rebasing from the latest master.
2. @rguthriemsft left good reviews :pray:. Please consider addressing them.
3. Upon testing, I found that the namespace sent to App Insights is `azure.applicationinsights`. Is it possible to change it to `k6` instead? I am not sure if it is configurable via the Go library you've used. Yet, since I am adding a new integration for Azure Monitor on our Cloud service ([Cloud APM](https://k6.io/docs/cloud/integrations/cloud-apm)), I can set a custom namespace while sending custom metrics to Azure Monitor. So I suspect it might be configurable. I mean the `Metric Namespace` in the screenshot below.
image

MStoykov:
Hi @polatengin , thanks for this PR :bow:. 

k6 has had support for output extensions for some time now. As of today it also has a [template repo for output extensions](https://github.com/grafana/xk6-output-template). This plus all the other extension outputs should be enough for people to extract this code in an extension and use it without it being in k6 core.

Given that I am closing this PR.

PR #1766, opened 2020-12-12
fix: flooring of progress when rate is low

### LINKED ISSUE #1753: When iterations per second is less than 1, it floors to 0 in the progress output
If you set up a scenario where the executors runs less than one iteration per second, the value is floored to zero while in progress. However, it reports the correct amount in the end-of-test summary.

## Environment
- MacOS Big Sur
- k6 v0.29.0

## Expected Behavior

```
checkEveryFiveSeconds   [=========>----------------------------] 0/1 VUs  0m17.1s/1m0s  0.20 iters/s
```

or

```
checkEveryFiveSeconds   [=========>----------------------------] 0/1 VUs  0m17.1s/1m0s  >0 iters/s
```

Given end-of-test summary

```
  data_received........: 0 B 0 B/s
  data_sent............: 0 B 0 B/s
  iteration_duration...: avg=154.19µs min=83.97µs med=151.48µs max=282.84µs p(90)=205.18µs p(95)=244.01µs
  iterations...........: 8   0.205413/s
  vus..................: 1   min=1 max=1
  vus_max..............: 1   min=1 max=1
```

## Actual Behavior

```
checkEveryFiveSeconds   [=========>----------------------------] 0/1 VUs  0m17.1s/1m0s  0 iters/s
```

Minimal repro:

```js
export let options = {
  scenarios: {
    checkEveryFiveSeconds: {
      executor: 'constant-arrival-rate',
      rate: 1,
      timeUnit: '5s',
      duration: '1m',
      preAllocatedVUs: 1,
    },
  },
};

export default function () {
  console.log(new Date());
}
```

DarkAEther:
Fixes flooring of the rate when precision is 0 and the rate is low (between 0 and 1).

Closes #1753 

Any feedback and suggestions are welcome!

### DIFF
--- a/ui/pb/helpers.go
+++ b/ui/pb/helpers.go
@@ -49,8 +49,20 @@ func GetFixedLengthIntFormat(maxValue int64) (formatStr string) {
 // precision + how many zeros it will be padded on the left with) in the
 // returned string corresponds to the number of digits in the supplied maxValue
 // and the desired precision.
+// if the precision is 0 and the maxValue is between 0 and 1, the number of decimal
+// places required to show the first non-zero digit is calculated and that is set 
+// to the precision.
 func GetFixedLengthFloatFormat(maxValue float64, precision uint) (formatStr string) {
 	resLen := 1
+	var decimalPlacesCount uint = 0
+	if precision == 0 && maxValue < 1 && maxValue > 0 {
+		value := maxValue
+		for value < 1 {
+			value *= 10
+			decimalPlacesCount += 1
+		}
+		precision = decimalPlacesCount
+	}
 	if maxValue < 0 {
 		maxValue = -maxValue
 		resLen++


asettouf:
Would be nice to have at least a test case associated with it. Maybe in `constant_arrival_rate_test.go` for unit testing?

DarkAEther:
> Would be nice to have at least a test case associated with it. Maybe in `constant_arrival_rate_test.go` for unit testing?

Thank you for the suggestion! I'll start working on it.

DarkAEther:
Hey again! I looked at the code in `constant_arrival_rate_test.go`. However, there don't appear to be any existing tests in there testing the printed rate (unless I missed something).

Would be great if I could get some help with adding a test case for this.

Thanks!

imiric:
@DarkAEther Thanks for the PR, and yeah, testing this as is would be difficult since this value is only exposed when rendered in the UI.

I'm wondering if this logic is generic enough for all use cases where `pb.GetFixedLengthFloatFormat()` is used. What do you think @MStoykov?

If so, we could make it part of that function, otherwise you could extract this into a standalone function in `constant_arrival_rate.go` and just unit test that, avoiding the need to test `ConstantArrivalRate.Run()`.

MStoykov:
This also seems to be broken in the ramping arrival rate as well ... 

I wonder if it won't be easier and more user friendly for the format to be something like "1/5s" (read as "1 per 5 seconds") instead of getting floats involved ... as "0.2/s" seems ... bad IMO ... and it might be okay here, but the ramping arrival rate will change this so it will be something like "0.1112/s"  then "0.1113/s" and so on ... instead it should probably go from something like "0/20s" to "1/20s"  or similar ... Where always the "/<number>s" is just what is configured and the number in front is the only thing that changes

DarkAEther:
> This also seems to be broken in the ramping arrival rate as well ...
> 
> I wonder if it won't be easier and more user friendly for the format to be something like "1/5s" (read as "1 per 5 seconds") instead of getting floats involved ... as "0.2/s" seems ... bad IMO ... and it might be okay here, but the ramping arrival rate will change this so it will be something like "0.1112/s" then "0.1113/s" and so on ... instead it should probably go from something like "0/20s" to "1/20s" or similar ... Where always the "/s" is just what is configured and the number in front is the only thing that changes

Hey! Sorry for the late response. This sounds like an interesting idea. This definitely seems like it would be more readable especially for smaller fractions. If it sounds reasonable and in scope, I could try out this change.

PR #1768, opened 2020-12-14
Summary callback changes

na--:
This is a WIP attempt at fixing a bunch of issues (fixes https://github.com/loadimpact/k6/issues/1316, fixes https://github.com/loadimpact/k6/issues/1611, closes https://github.com/loadimpact/k6/issues/817, closes https://github.com/loadimpact/k6/issues/647, closes https://github.com/loadimpact/k6/issues/351) and making possible the implementation of a bunch of other ones (https://github.com/loadimpact/k6/issues/1315, https://github.com/loadimpact/k6/issues/1675, https://github.com/loadimpact/k6/issues/1319)...

The way we do that is by supporting a new `handleSummary()` exported function that allows users to customize the test summary (or summaries) however the want. The final goal is something like this:
```js
import http from 'k6/http';
import { JUnit, TextSummary } from 'https://jslib.k6.io/k6-summary/1.0.0/index.js' // or something like that
// ...
export function handleSummary(data) {
    // save everything to files and send it to a server when the test finishes or gets aborted
    http.post('https://example.com/testResults', JSON.stringify(data));

    return {
        'stdout': TextSummary(data),
        './results/junit.xml': JUnit(data),
        './results/summary.json': JSON.stringify(data),
    }
}
```

With the current state of the PR, `handleSummary()` is supported and its results are saved to stderr/stdout/files. I've also heavily refactored the `--summary-export` (implemented it through the new system in a backwards compatible way) and fixed a bunch of `--sumary-trend-stats` bugs and group/check order. But we don't have a JS equivalent of the old summary (`TextSummary()` in the example above) yet or `JUnit()`. For now, I've exposed the old Go code that generated the test summary as a second parameter to `handleSummary()`, but I'll remove that in the final version. So, to test this, you might use something like this:
```js
export function handleSummary(data, oldCallback) {
    console.log('handleSummary() was called');
    return {
        'stdout': oldCallback(),
        'raw-data.json': JSON.stringify(data, null, 4),
    };
}
```

The following things remain to be done:
- replace the old text summary in Go with one a JS function
- write a JUnit example (https://github.com/loadimpact/k6/issues/1366, https://github.com/loadimpact/k6/issues/1120) and publish it and the old JSON/text summary methods to jslib
- make a k6-docs PR
- go through the old k6 issues and see if there are other ones that should be closed or updated now that we have this

## Breaking changes and other bugfixes
- `--no-summary` now disables even `--summary-export` 
- the `value` for `Rate` metrics in the old `--summary-export` JSON file was was always `0`, regardless of the `pass/(pass+fail)` ratio


na--:
Here is a more complete example that also demonstrates the fixed summary columns even in the old backwards-compatible `--summary-export` handling, and the fact that the wrappers can work in `--compatibility-mode=base`: https://gist.github.com/na--/647cbafe46abd1a0af37402577e5d251

sniku:
I'm working on creating an end-of-test summary for functional/integration tests. 
This will take some time, so I'll just post comments here as I work through it. I don't expect any comments back since we have holidays! :smile: 

The example posted above is hitting a bug/mistake in the `summaryWrapperLamnbdaCode`. The result is:
```
ERRO[0001] handleSummary() failed with error "TypeError: Value is not an object: undefined", falling back to the default summary  source=console
```

`oldCallback` should probably be passed to `exportedSummaryCallback` as a second argument.

https://github.com/loadimpact/k6/pull/1768/files#diff-a0d09d808d4fa4f068a2bb81b9fa180f596f46439d040f840687d559404b5cd9R99

I can fix/work around it, so it's not a problem.

stats/stats.go
```diff
@@ -69,8 +69,17 @@ var ErrInvalidValueType = errors.New("invalid value type")
 // A MetricType specifies the type of a metric.
 type MetricType int
 
-// MarshalJSON serializes a MetricType as a human readable string.
+// MarshalJSON serializes a ValueType as a human readable string.
 func (t MetricType) MarshalJSON() ([]byte, error) {
```
imiric, line 73: Copy/paste error?

stats/stats.go
```diff
@@ -86,7 +95,7 @@ func (t MetricType) MarshalJSON() ([]byte, error) {
 }
 
 // UnmarshalJSON deserializes a MetricType from a string representation.
```
imiric, line 97:
```suggestion
// UnmarshalText deserializes a MetricType from a string representation.
```

Weird that `golangci-lint` missed this... :-/

stats/stats.go
```diff
@@ -123,6 +132,15 @@ type ValueType int
 
 // MarshalJSON serializes a ValueType as a human readable string.
```
imiric, line 133:
```suggestion
// MarshalJSON serializes a ValueType to a JSON string.
```

todo.txt
```diff
@@ -0,0 +1,12 @@
+todo:
+
+remove all of the old UI summary code, replacing it with JS
```
imiric, line 3: Is this strictly needed? Couldn't we just reuse the current Go implementation and call it from JS?

If nothing else it would save us time to get this released in v0.30.0.

imiric (changes requested):
Nice work! :clap:

The code LGTM, besides the minor comments. I'll take another look next year ;)

Just some notes:
- Strong :thumbsup: for splitting this into separate commits/PRs. It would've been easier to review like that as well. :sweat_smile:
- I like the `GetResolversForTrendColumns()` refactor, despite of those `Sink` pain points. :thumbsup:

cmd/runtime_options.go
```diff
@@ -95,6 +107,32 @@ func getRuntimeOptions(flags *pflag.FlagSet, environment map[string]string) (lib
...
+			opts.NoSummary = null.BoolFrom(val)
+		}
+	}
+
+	if !opts.SummaryExport.Valid { // If not explicitly set via CLI flags, look for an environment variable
+		if envVar, ok := environment["K6_SUMMARY_EXPORT"]; ok {
+			opts.SummaryExport = null.StringFrom(envVar)
+		}
+	}
+
```
MStoykov, line 135: See https://github.com/loadimpact/k6/pull/1599/files#r482818305

todo.txt
na--: Unfortunately we need it... :disappointed: Otherwise there would be no way for the users to specify `handleSummary` and keep the current stdout text summary and output some other format to a file, for example. We can't keep the current hack of passing the Go function as a second argument, unless we want to maintain that indefinitely... :disappointed:

MStoykov:
I have 2 concept/UX/big issues: 
1. I think I've already expressed this - but I don't think returning an object is good, IMO it's probably better to provide some API, through one of the arguments, not an import, on how to write to stdout/files so something like:
```javascript
import http from 'k6/http';
import { JUnit, TextSummary } from 'https://jslib.k6.io/k6-summary/1.0.0/index.js' // or something like that
// ...
export function handleSummary(data, api)  { // Worst name ever I know :)
    // save everything to files and send it to a server when the test finishes or gets aborted
    http.post('https://example.com/testResults', JSON.stringify(data));
    api.writeToStdout(TextSummary(data));
    // or TextSummary(api.stdout, data);
    var f = api.openFile("./results/junit.xml");
    f.write(JUnit(data));
    // or JUnit(f, data); 
}
```

This will (without generators that are currently not supported, last I checked) let us write big responses without being memory hungry. 
All of those objects/functions should have very similar syntax - probably just `write` which appends data, we can add `seek` to the file ones at a later point if there is sufficient demand.

I haven't looked into it, but I expect there is a somewhat common "interface" for this in the JS world that can be used.

2. I feel like there will be a lot of cases where the test will be executed and the function will throw an exception because the test wasn't called correctly and the most common case IMO will be because the `handleSummary` will expect that it has `p(23)` but the user didn't' call it with the correct `--summary-trend-stats`, so it will blow up after an hour of testing.  

I think it is worth trying to figure out if we can actually have a way that this is prevented. My first thought (which works great for the summary-trend-stats) is to completely change the API to something like (all names of stuff are bad and only for illustrative purposes):

```
import summary from "k6/summary";

File: stats/stats.go (hunk 1 above)
Comment (na--) [reply stats/stats.go:73]: Resolved by https://github.com/loadimpact/k6/pull/1768/commits/970122cd9733bc25a1630fc2072ba1257a2e1407

File: stats/stats.go (hunk 2 above)
Comment (na--) [reply stats/stats.go]: Resolved by https://github.com/loadimpact/k6/pull/1768/commits/970122cd9733bc25a1630fc2072ba1257a2e1407

File: stats/stats.go (hunk 3 above)
Comment (na--) [reply stats/stats.go]: Resolved by https://github.com/loadimpact/k6/pull/1768/commits/970122cd9733bc25a1630fc2072ba1257a2e1407

File: js/runner.go [hunk 6]
```diff
@@ -312,6 +307,77 @@ func (r *Runner) IsExecutable(name string) bool {
...
+		return nil, err
+	}
+
+	handleSummaryFn := goja.Undefined()
+	if exported := vu.Runtime.Get("exports").ToObject(vu.Runtime); exported != nil {
+		fn := exported.Get(consts.HandleSummaryFn)
+		if _, ok := goja.AssertFunction(fn); ok {
+			handleSummaryFn = fn
+		} else if fn != nil && !goja.IsUndefined(fn) && !goja.IsNull(fn) {
+			return nil, fmt.Errorf("exports.%s should be a function or undefined", consts.HandleSummaryFn)
```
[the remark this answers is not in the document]
na--: Resolved by https://github.com/loadimpact/k6/pull/1768/commits/970122cd9733bc25a1630fc2072ba1257a2e1407

na--:
@mstoykov, I think we've previously discussed most of these ideas in the private issue, but to rehash things:

> I don't think returning an object is good, IMO it's probably better to provide some API, through one of the arguments, not an import, on how to write to stdout/files

If you recall, having some sort of a file writing API was port of my original proposal for this functionality. I later retracted this, because that API would have been a deep and bikesheddy rabbit hole which didn't (and still doesn't) seem worth getting into. I didn't want to have to figure out a nice JS API for file opening/closing, permissions, appending, binary data, etc. This feature is big enough without having to deal with that from scratch as well...

The currently returned `{filepath: contents}` map is basically a simple MVP version that would likely satisfy 95+% of users. Memory usage is unlikely to be a concern, even if someone decides to generate fancy PDF or HTML reports. And if this ever proves insufficient, we can always add a file writing API at a later point, as a separate PR. The current implementation is [20 lines](https://github.com/loadimpact/k6/blob/970122cd9733bc25a1630fc2072ba1257a2e1407/cmd/run.go#L436-L459) of simple and testable code, so it won't be a big overhead to maintain.

Besides, any file writing API will probably be better off as something generic, and better of being polished as an xk6 extension for a while, not directly in core. 

cc @sniku @imiric what are your thoughts here? I am not opposed to a proper API, but that would mean delaying the feature again and not getting it in v0.30 for sure.

> I feel like there will be a lot of cases where the test will be executed and the function will throw an exception because the test wasn't called correctly...

True, that's a risk, but you can say the same thing about `teardown()`. If you saw, I plan to add some exception handling for errors in the JS `handleSummary()` wrapper, so we will at least show the default summary if an error is thrown by user code. 

I don't think the UX of passing the required percentiles (`["p(23)"]`) and other data is very good. Even the toy example looks strange, but a real-life example would be even more awkward. And having percentile APIs instead of plain data wouldn't necessarily be less error prone.

That said, I like the approach of relying on an imported module and calling a method on it (`summary.register()`) more than relying on exported magic functions with specific names... I am not sure it makes sense to `register()` multiple handlers for the end-of-test summary, though we might want to instead make this a more generic event-handling system. Something like:

```js
import events from "k6/events";

events.registerHandler('endOfTestSummary', someHandler); // or maybe addEventListener :D 
events.registerHandler('midTestSummary', someOtherHandler);
events.registerHandler('vuInit', perVUInit); 
events.registerHandler('abortedByThreshold', whatever); 
// ...
```

But again, this seems like it can grow in complexity and needs careful design because of the k6 multi-JS-runtime arhitecture. It also seems like something that can be added in the future, in a backwards compatible way. So, for now, starting with an exported `handleSummary` function doesn't seem like a bad idea and doesn't seem like it will introduce any overhead. If anything, most of the overhead of this feature comes from the huge amount of old text summary and `--summary-export` cruft. Once we wrap those in `handleSummary()`, it should be reasonably easy to work with in the future.

cmd/runtime_options.go
na--: I fixed this in https://github.com/loadimpact/k6/pull/1788/commits/f12d7b72463e6fe1b3f6a516944dc1daec2cee8f

imiric:
I like the flexibility of @MStoykov's proposal, but also feel that it would cause a delay of v0.30.0 to implement it. Since we can add a proper API in a backwards compatible way in the future (new users simply won't have to return an object), my vote is to merge this as is and discuss and design the API proposal in a subsequent issue/PR.

I also like the event-based registration idea for summary handlers, but that also feels out of scope for this MVP.

js/runner.go
```diff
@@ -312,6 +307,77 @@ func (r *Runner) IsExecutable(name string) bool {
...
+
+	vu, err := r.newVU(0, out)
+	if err != nil {
+		return nil, err
+	}
+
+	handleSummaryFn := goja.Undefined()
+	if exported := vu.Runtime.Get("exports").ToObject(vu.Runtime); exported != nil {
+		fn := exported.Get(consts.HandleSummaryFn)
+		if _, ok := goja.AssertFunction(fn); ok {
```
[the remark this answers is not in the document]
na--: Sorry, just realized I've forgotten to respond here. I am not sure if what you suggest will work, since we're creating a new VU here and `vu.Runtime.Get("exports")` would be different than `r.Bundle.exports`. Moreover, this way we'll raise an error if `handleSummary` was exported but it wasn't a function.

MStoykov:
I tried (successfully I would say) port https://github.com/Mattihew/k6-to-junit to this new change [here](https://gist.github.com/MStoykov/5e052293cc1bbdd11284a2cc5a90a194)

cmd/run.go
```diff
@@ -290,36 +291,18 @@ a commandline interface for interacting with it.`,
...
-					defer func() {
-						if err := f.Close(); err != nil {
-							logger.WithError(err).Error("failed to close summary export file")
-						}
-					}()
-					s := ui.NewSummary(conf.SummaryTrendStats)
-					if err := s.SummarizeMetricsJSON(f, data); err != nil {
-						logger.WithError(err).Error("failed to make summary export file")
-					}
+					logger.WithError(err).Error("failed to handle the end-of-test summary")
```
MStoykov, line 305: I guess that in a sequential release we will fallback to a default representation or the dumping the `lib.Summary` as JSON to the disk?

ui/summary.go
```diff
@@ -309,11 +212,11 @@ func (s *Summary) summarizeMetrics(w io.Writer, indent string, t time.Duration,
 
 				resolver := s.trendValueResolvers[tc]
 
-				switch v := resolver(sink).(type) {
-				case float64:
+				v := resolver(sink)
+				if tc != "count" { // sigh
```
MStoykov, line 216: nitpick ... but why are you checking for the reverse here instead of for the equal one? ;)

cmd/run.go
na--: not sure what you mean by "a sequential release", but we'll get an error here only if something goes majorly wrong... Both `initRunner.HandleSummary()` and `handleSummaryResult()` try to continue working even if there are errors. `handleSummaryResult()` tries to save all files, even if some of them fail (because of wrong names, lack of permissions, etc.). And `initRunner.HandleSummary()` has internal `try / catch` in the JS wrapper code: https://github.com/loadimpact/k6/blob/0838813db6edead0f6faa3cdaadf3b5eb070a1ab/js/summary.go#L97-L104

Though I should add a test for that...

ui/summary.go
na--: :man_shrugging: probably just because that was the case order in the old `switch`, but yeah, doesn't makes sense, so I'll  change it

na--: in the JS variant of this though, no need to further mess with the Go code

js/runner.go
```diff
@@ -312,6 +307,77 @@ func (r *Runner) IsExecutable(name string) bool {
...
+			handleSummaryFn = fn
+		} else if fn != nil && !goja.IsUndefined(fn) && !goja.IsNull(fn) {
```
MStoykov, line 332: I checked and `Get` returns `nil` if it doesn't exist. Did you intend for `exports.handleSummary = undefined;` (or `null`) to be a valid thing that won't raise an error or were you over zealous in the checks? 

I am not particularly certain it matters either way, to be honest. If someone has defined `handleSummary` it is probably better to use `delete exports.handleSummary`, on the other hand, I don't think we will be hitting it. So the question is mostly questing what was the idea behind it, not changing it one way or another

js/runner.go
```diff
@@ -312,6 +307,77 @@ func (r *Runner) IsExecutable(name string) bool {
...
+
+	wrapperArgs := []goja.Value{
+		handleSummaryFn,
+		vu.Runtime.ToValue(r.Bundle.RuntimeOptions.SummaryExport.String),
+		vu.Runtime.ToValue(summaryDataForJS),
+		vu.Runtime.ToValue(getOldTextSummaryFunc(summary, r.Bundle.Options)), // TODO: remove
+	}
+	rawResult, _, _, err := vu.runFn(ctx, false, handleSummaryWrapper, wrapperArgs...)
+
+	// TODO: refactor the whole JS runner to avoid copy-pasting these complicated bits...
```
MStoykov, line 363: I am pretty sure you can just use `runPart` for most of the above. 
You will need to keep the check for whether `exports.handleSummary` is a function, but everything else seems to be done by `runPart` either way ... I think ...

MStoykov: Although I guess you will need to move https://github.com/loadimpact/k6/blob/0838813db6edead0f6faa3cdaadf3b5eb070a1ab/js/summary.go#L95-L118 to golang code after the `runPart`

MStoykov (approved):
LGTM. 

I had some more comments, but even the `runPart` thing seems likely to be more involved with little actual benefit, so I guess it can be even be left for a later PR or not done at all.

cmd/run.go
MStoykov: yeah, I missed this code at the time ...

This is probably my biggest issue reviewing the PR, the code is all over the place, although I don't know how to fix that if it's at all possible given what the code tries to do, so I don't have some concrete feedback. I hope that when we rewrite even more of the whole metric and threshold (which specifically are also over the place) this might get more streamlined

imiric (dismissed):
LGTM, nice work!

js/runner.go
na--: > were you over zealous in the checks?

Likely that, and it makes sense to leave only the `nil` check :+1:

js/runner.go
na--: Yeah, I initially tried using `runPart`, but it was an even bigger mess. We can figure out some way to refactor the whole `js.Runner` internally in the future to reduce the boilerplate, but for now I don't want to mess with it more if I don't need to... :sweat_smile:

MStoykov (approved):

imiric (approved):

Issue #1321, opened 2020-01-28
Allow whitelisting of metrics and sub-metrics


na--:
k6 measures a lot of things, and it can and can generate quite a lot of data points for a variety of different metrics. Not everything we measure is important to everyone, and we want to measure even more things - https://github.com/loadimpact/k6/issues/877, https://github.com/loadimpact/k6/issues/880, https://github.com/loadimpact/k6/issues/888, https://github.com/loadimpact/k6/issues/921, https://github.com/loadimpact/k6/issues/1311, and all of the non-HTTP protocols we want to add...

Unfortunately, currently there's no way for a user to tell k6 that they're not interested in some of the things it measures (https://github.com/loadimpact/k6/issues/570). That's a problem, because measuring and reporting useless things adds a significant overhead both to k6 (https://github.com/loadimpact/k6/issues/764), and to any custom outputs that users pipe the metrics data to (https://github.com/loadimpact/k6/issues/1060).

On top of that, some of the current default metrics just add clutter and confusion to the end-of-test summary report (https://github.com/loadimpact/k6/issues/1319), without a lot of benefit to most users. All of the `http_req_*` "minor" metrics probably shouldn't be enabled by default, instead we should probably have only `http_req_duration`, which is `sending + waiting + receiving`. Maybe we should also expose the [composite metric](https://github.com/loadimpact/k6/blob/9af6e756546192ddda0bfc10d25e6fd9b8e98e76/lib/netext/httpext/tracer.go#L41-L42) for the connection establishment, but I doubt most people would care about `http_req_tls_handshaking` individually, most of the time. `group_duration` is also somewhat useless when you have a variety of different groups and don't use an external metrics output... Unfortunately, we currently can't remove any of these metrics _by default_, since we don't have a mechanism for users to enable them if they want/need them...

There's also no way besides thresholds for users to force k6 to calculate and display information about sub-metrics, i.e. specific subsets of the data points of the main metrics. This point was [recently raised](https://community.k6.io/t/show-tag-data-in-output-or-summary-json-without-threshold/392) by a user in the community forum, and unfortunately, thresholds is the only way to tell k6 to keep track of sub-metrics right now :disappointed:

So, I think I've listed all of the connected issues I can remember... My proposed solution, as suggested by the issue title, is to allow users to whitelist metrics and sub-metrics. k6 should have a default whitelist, but we should allow users to set their own. Here is a suggestion for the simplest possible approach we can offer, though I think this might not be the best solution:
```js
[...]

Issue #877, opened 2018-12-18
Emit an errors metric and have a default error rate script abort threshold 


na--:
It would be very useful if we resurrect the currently unused [`errors` metric](https://github.com/loadimpact/k6/blob/d219b19a0353ddc73063ff3c45bc17444f0771c3/lib/metrics/metrics.go#L35) and also add a default threshold that would abort the script if the error rate exceeds a certain configurable value.

This issue is a summary of an idea that has been discussed in various other issues and PRs before:
- https://github.com/loadimpact/k6/issues/735
- https://github.com/loadimpact/k6/issues/688#issuecomment-402426764
- https://github.com/loadimpact/k6/pull/741#issuecomment-412809068
- https://github.com/loadimpact/k6/issues/818

na--:
All exceptions during an iteration will add to the `errors` metric, but we probably should offer users more flexibility by extending the [`fail()` function](https://docs.k6.io/docs/fail-err) to support custom error tags. Something like this: `fail("error message", { tags: {action: "customer-add", cause: "cannot-login" } }`. 

And ideally, we should improve the UX of `fail()` errors. Currently, running this script:
```js
import { fail } from "k6";

export default function () {
   fail("test");
}
```
will produce and error like this:
```
ERRO[0000] GoError: test
    at native
    at script.js:4:18(4) 
```

whereas something like `Fail "test" at script.js:4:18(4)` or `script error ...` or `user error` would be much better - normal users don't need to know k6 is written in Go... If there's an use case, we might even add a way to specify extra data in `fail()` besides the metric tags and error message, though I'm not sure what a good use case would be for that.

lukecusolito:
@na-- the referenced issue #735 was closed in favor of this one. Will this issue cover the following?

When a check fails or fail() is called, a failure exit code is returned. 
This would enable us to fail a build pipeline in the event the check criteria is not met without setting a threshold.

na--:
@lukecusolito, sorry for the late response, I somehow missed your question until I just now looked at this issue for another reason :disappointed: 

To answer your question, a single check or a `fail()` likely won't abort the whole script. Rather, if a certain percentage of iterations are aborted with an error (either `fail()` or a thrown JS error), k6 would abort the whole script. 

[...]

PR #1769, opened 2020-12-14
[WIP] Add InterruptedIterations metric

imiric:
These are partial changes for #877, but I'm creating a draft PR to clear up a few things and get some feedback. It's probably easier to view this commit by commit.

Currently only scenarios 1 and 2 from [this comment](https://github.com/loadimpact/k6/issues/877#issuecomment-673328354) are working (`cause: error` and `cause: fail`). `cause: interrupted` is commented out as it needs a different approach, and I'm not sure how to handle the last `^C` scenario (`cause: signal` would make sense for that).

I settled on using a `Rate` metric for `interrupted_iterations`, but opted to hack around actually emitting the `0` value in order to avoid an explosion of metric data, as this would be a duplicate of `iterations`, and given #1321 is still open we shouldn't exacerbate this problem. See e2b88ab4.


imiric:
Some UI questions:

`fail('test')` and `throw 'test'` currently look like this:

```
ERRO[0001] fail: test at github.com/loadimpact/k6/js/common.Bind.func1 (native)  executor=constant-vus scenario=default
ERRO[0001] test at file:///tmp/test-877-fail.js:5:11(8)  executor=constant-vus scenario=default
```

and the metric in the summary like this:

```
interrupted_iterations...: 57.14% ✓ 40  ✗ 30
```

1. I would like to make the `fail` case also show the correct file and line number. I haven't dug into `bridge.go` yet and not sure if it's possible with goja, but is this something that we want?
2. The way `Rate` metrics are currently rendered in the summary makes the `✓` and `✗` numbers kind of ambiguous. Does `✓` refer to the "successful"/"complete" iterations or how many were interrupted? And same for `✗`, it could be interpreted both ways.
  Currently `✓` is the `Sink.Trues` value, so these are the interrupted iterations, but not sure how we could make this clearer, besides using a `Counter`.

lib/executor/helpers.go
```diff
@@ -84,26 +87,56 @@ func getIterationRunner(
...
 
 		select {
 		case <-ctx.Done():
-			// Don't log errors or emit iterations metrics from cancelled iterations
+			// TODO: Do this elsewhere? It might be racy with
+			// Engine.processMetrics() as the channel could be closed at this point.
+			// TODO: Use enum?
+			// TODO: How to distinguish interruptions from signal (^C)?
+			// tags["cause"] = "duration"
+			// state.Samples <- stats.Sample{
```
imiric, line 104: This is where we would handle the 3rd scenario of executors interrupting the iteration, and while it worked in manual tests and `TestMetricsEmission`, it would cause other tests to hang (presumably because they were setting a `nil` channel), and I'm not sure if it wouldn't be racy, so I commented it out for now.

Is this the right place/approach, or would we have to receive two contexts here or something more complicated?

imiric: Digging through the context chains we're using, this shouldn't be racy since the `Samples` channel is closed only when the `globalCtx` is done, which happens after all metrics processing is finished, so it should be usable even if this VU "run context" is done. Yet some of our tests use the same context for both, e.g.:

https://github.com/loadimpact/k6/blob/420dd4161c280bdf0398af5e58f73c939057695a/api/v1/status_routes_test.go#L108

This explains why it works in real-world tests but not in the test suite. I'll see if I can fix the tests then.

core/engine.go
```diff
@@ -288,6 +288,14 @@ func (e *Engine) processMetrics(globalCtx context.Context, processMetricsAfterRu
 			if !e.NoThresholds {
 				e.processThresholds()
 			}
+			if iim, ok := e.Metrics[metrics.InterruptedIterations.Name]; ok {
```
imiric, line 291: This map lookup is racy according to the current `TestSentReceivedMetrics` failure. So I guess we'll need to rethink or abandon this hack...

MStoykov: I am pretty sure this breaks any output as you are only emitting the `1` to them ...

js/common/bridge_test.go
```diff
@@ -348,7 +348,7 @@ func TestBind(t *testing.T) {
 		}},
 		{"Error", bridgeTestErrorType{}, func(t *testing.T, obj interface{}, rt *goja.Runtime) {
 			_, err := RunString(rt, `obj.error()`)
-			assert.Contains(t, err.Error(), "GoError: error")
+			assert.Contains(t, err.Error(), "error")
```
MStoykov, line 351: I think this can be moved to another PR that we can merge today and make this ... more focused ;)

lib/executor/helpers.go
```diff
@@ -84,26 +87,53 @@ func getIterationRunner(
...
+				tags := map[string]string{"cause": "error"}
+
+				// TODO: investigate context cancelled errors
+				switch e := err.(type) {
+				case *goja.Exception:
+					if er, ok := e.Value().Export().(lib.IterationInterruptedError); ok {
+						tags["cause"] = er.Cause()
+					}
+					logger.Error(err.Error())
+				case fmt.Stringer:
```
MStoykov, line 121: This literally was worded as `fmt.Stringer` because I didn't want to add one more place where lib and lib/executor specifically depend on goja. So I think it is better if we don't actually start using it and also you need to fix the fact that now exceptions don't have "source=stacktrace" ;) 

I do think now is a good time to actually add and error type (under lib), probably just an interface "Exception", that is implemented in the `js` package and wraps goja.Exception so that we don't use this directly.

core/engine.go
imiric: I guess it depends on how this metric is agreggated. If their system allows it they could factor it in with the `iterations` metric and get a percentage that way, which is kind of what we're doing here for the summary.

It makes no sense to me to emit the `0` value of this metric considering it's equivalent to `iterations`, and it would add a considerable amount of output data and affect the performance. A 5s/5VU test with an empty `default` and no sleep does 2115298 iterations on my machine and generates a 1.5 GB JSON file when emitting the `0` value. The same test without it manages to do 3203002 iterations and generate a 1.9 GB JSON file, but this is because it's doing ~50% more iterations.

Considering we already have complaints about outputting too much data and interest in #1321, I think we should rather work on #1321 before we consider adding another default metric.

Though we should probably rethink duplicating `iterations`...

How about instead of a new metric, we simply reused `iterations` and added an `interrupted: "<cause>"` tag to it? If the tag is omitted it's a complete iteration and if it's set then it was interrupted with the specific cause. WYT?

js/common/bridge_test.go
imiric: Sure, will do.

lib/executor/helpers.go
imiric: Hhmm sure, I'll give that a try, thanks.

core/engine.go
MStoykov: I agree with prioritizing #1321, but that will actually probably require quite the refactoring, IMO even for an MVP. 
The reusage of the iterations was discussed (I think) and we decided against it based on the fact the primary usage for this will be to check that a percentage of the iterations aren't actually interrupted, which with the current thresholds isn't possible if the metric isn't rate... so :(. 

 I think that maybe we should just make another PR with the changes that are in this PR (the test fixes and the lib.Exception) and merge that to both have them and make this PR easier to rebase and then work on #1321 so we can merge this then.

imiric: > The reusage of the iterations was discussed (I think) and we decided against it based on the fact the primary usage for this will be to check that a percentage of the iterations aren't actually interrupted, which with the current thresholds isn't possible if the metric isn't rate

Then maybe that should be a feature? I'll see what I can do.

js/common/bridge_test.go
imiric: See #1775.

lib/executor/helpers.go
imiric: I managed to remove the goja dependency in d52efc49, but I'm not happy with how it turned out...

I wanted to treat all errors and exceptions consistently, avoid logging in `runFn()` and delegate that to `getIterationRunner()`, but this took an embarassing amount of trial/error and I'm still not sure all errors will be handled correctly. :-/

Some notes:
- I didn't think making `lib.Exception` an interface was needed, as it's generic enough to have a single implementation.
- This needs a lot more tests to ensure errors are rendered properly.
- The `source=stacktrace` logger field doesn't make sense to me. If anything the source should be the same as the "cause" value, as "stacktrace" isn't really a "source".
- Should all errors output the full stack trace? I was trying to keep it consistent with `master`, but maybe errors from `fail()` should only output a short single-line trace with the last stack frame.

na--:
We didn't merge this as it was because we realized we needed https://github.com/grafana/k6/issues/1321 (or something like it) for this PR to not introduce performance problems. At this point it has so many merge conflicts that it'd be easier to start from scratch, so I'll close it to reduce noise.

PR #1772, opened 2020-12-15
Drop some amount of corejs also the file in git tree

MStoykov:
Reasons for doing it this way:
1. this lets us easily see what is actually included in the fina core-js
   build by looking at k6-shim.js. The format there is probably going to
   change as it is copied from the core-js repo's shim.js, but the idea
   will stay
2. There is no need to have the shim.min.js in the git tree as it is
   already in lib/rice-box.go. Maybe after the golang embed proposal is
   available we can switch around, but this will likely take more then a
   year before we don't need to support a version of golang without it
3. go generate is awesome, but in this case it has way too many things
   that need to happen which makes it IMO a bad fit. This plus the fact
   that this will not be changed regularly means it is probably better
   this way.


MStoykov:
Some performance running
```
import http from "k6/http";
import { sleep } from "k6";

export let options = {
    discardResponseBodies: true,
}
export default function() {
    http.get("https://httpbin.test.loadimpact.com/get");
    sleep(Math.random() * 2 + 3);
}
```
with `-vus 500 --duration 60s main.js -v --no-thresholds --no-summary` 
version       | iterations | mem usage | CPU usage | time to initialize VUs    | run time |
--------------  | ------------ | ---------------- | --------------- | ----------------------------- | ----------- |
g15e15e59   |  7464    | 1424300    | 45%          | ~5.8s                         | 1:10.90  |
g0e6c0aba  | 7472     |  599880      | 24%          | ~3.2s                       | 1:07.19 |
g0e6c0aba + base      | 7473     | 279292      | 18%          | ~0.9s                         | 1:05.81  |


The first is ( the current ) master, g0e6c0aba is this PR and `+base` is with compatility-mode=base and the same script rewritten to be es5.

MStoykov:
Comment on the still available core-js modules:
* es7.array.flat-map - this seems fine and probably something that can be added to goja by us
* es7.array.flatten - same as above + it is now called `flat` and `flatten` doesn't exist see ://github.com/tc39/proposal-flatMap 
* es7.string.trim-left, trim-right are synonims (from AnnexB(b for browser)) to trimStart,trimEnd - again can be added to goja (even easier) and will probably be welcomed as they are used by babel7 for some reason :facepalm: 
* es7.string.match-all - can be added and probably won't be all that hard, but we will need some benchmark tests ;) 

* es7.object.get-own-property-descriptors - the variant without `s` is in goja, so adding this shouldn't be too hard.
* es7.object.values, es7.object.entries - no idea but seems easy to be added 
* es7.object.define-getter, es7.object.define-setter, es7.object.lookup-getter,es7.object.lookup-setter - apparently deprecated, so maybe not even worth implementing :shrug: 
* es7.reflect.define-metadata and friends are seem to be https://github.com/rbuckton/reflect-metadata which isn't even a proposal AFAIK ... https://medium.com/jspoint/introduction-to-reflect-metadata-package-and-its-ecmascript-proposal-8798405d7d88  So maybe dropping them is a good idea as well

js/bundle_test.go
```diff
@@ -114,12 +114,18 @@ func TestNewBundle(t *testing.T) {
 		}
 	})
 	t.Run("CompatibilityMode", func(t *testing.T) {
+		compatibilityTestSrc := `
+		module.exports.default = function() {};
+		if ([[1,2],[3,4]].flatten()[4]  == 4) {
+			throw Error("Array.flatten doesn't work")
```
imiric, line 120: This would never fail, no? It would always be undefined, even if `flatten()` works, unless you meant `[3] === 4`.

But this would be better checked with `if (typeof Array.prototype.flatten !== 'function') ...`.

Also maybe we should use a different object to test this, since [`flatten` was renamed to `flat` in corejs v3](https://github.com/zloirock/core-js/issues/405), and it would avoid us having to update this when we update to v3.

js/lib/build.sh
```diff
@@ -0,0 +1,19 @@
...
+sha="a3616d74c0b9b9409d580337e04c32da6e3877e4" # v2.5.1
+mkdir -p ./corejs-repo
+cd ./corejs-repo
+git init
+git remote add origin https://github.com/zloirock/core-js.git
+git fetch origin --depth=1 "${sha}"
+git reset --hard "${sha}"
+cp ../k6-shim.js shim.js
+cp ../corejs-build.sh corejs-build.sh
+docker run --user "$UID:$GID" --rm -ti  -v $(pwd):/opt/g --entrypoint /opt/g/corejs-build.sh --workdir /opt/g node
```
imiric, line 12: It would be safer to quote, and also use `id` in case the env vars aren't defined:

```suggestion
docker run --user "$(id -u):$(id -g)" --rm -ti  -v "$(pwd):/opt/g" --entrypoint /opt/g/corejs-build.sh --workdir /opt/g node
```

But since we're using Docker for this, could we do everything there, including the `git` and `rice` commands? I don't have `rice` installed locally, so it would be good if all of this is contained. :)

I would also suggest using Bash and `set -euo pipefail`, but I think you're avoiding that to do the cleanup, which is fine, but you could also do it in Bash with `trap`. Not a big deal though if you don't want to complicate it.

js/lib/corejs-build.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+
+npm i .
```
imiric, line 3: This is failing for me when I run `build.sh`. `grunt` fails to install because of:

```
npm ERR! code ERESOLVE
npm ERR! ERESOLVE unable to resolve dependency tree
npm ERR!
npm ERR! While resolving: core-js@2.5.1
npm ERR! Found: grunt@1.0.4
npm ERR! node_modules/grunt
npm ERR!   dev grunt@"1.0.x" from the root project
npm ERR!
npm ERR! Could not resolve dependency:
npm ERR! peer grunt@"~0.4" from grunt-livescript@0.6.0
npm ERR! node_modules/grunt-livescript
npm ERR!   dev grunt-livescript@"0.6.x" from the root project
npm ERR!
npm ERR! Fix the upstream dependency conflict, or retry
npm ERR! this command with --force, or --legacy-peer-deps
npm ERR! to accept an incorrect (and potentially broken) dependency resolution.
```

`npm i --force` works, but maybe there's a better workaround.

imiric (changes requested):
LGTM, just some minor suggestions.

I like that we can easily see and change what's included in corejs 👍

The idea is that we would run `build.sh` manually whenever we want to upgrade, right?

imiric:
Also, that's an impressive performance gain! 👏  🎉 

How did you know what to remove and what to keep? Can we assure that we're not removing something a script was relying on?

MStoykov:
@imiric mostly the discussion here https://github.com/dop251/goja/pull/180 and talking with @simskij ... a long time ago. Plus obviously going through the things in the list from core-js v2.5.1 and looking at them, the test262 also helps ;)  

Some of the drops might be used but also aren't standard at all ... just core-js used to (and probably still does) stuff that are still in the proposal process :facepalm: - see the comments in `k6-shim.js` for it

js/lib/corejs-build.sh
MStoykov: Hm ... that seems to be because I don't specify a version for node and it works with mine (14.x something) while it doesn't work with 15.4.0

js/lib/build.sh
MStoykov: I prefer not to complicate it ... predominantly because I don't think anyone will run this again more than once or twice, so spending a lot of time making it awesome seems unnecessary when in reality whoever runs it (probably to drop even more stuff in the future) will run it once and then we will either never touch it again or completely drop it. 

This very much depends on what the discussion about https://github.com/loadimpact/k6/pull/1772#issuecomment-745417009 revolves around ... if we drop the metadata stuff, the rest is standard and likely fairly easy to implement, so I will vote for taking a week to implement parts of it and get them in goja and then repeat until all of them are in goja and then just drop the whole core-js polyfill :tada: . But I also expect that even if this happens it won't be in the next release or soon at all

js/lib/corejs-build.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+
+npm i --force .
+modules="$(cat shim.js | grep ^require  | sed "s|require('\./modules/||g"  | sed "s/');//g" | tr "\n" ",")"
```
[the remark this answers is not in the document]
MStoykov: I am not going to this for now, as if we remove more from the `shim.js` or change its format, I will need to redo it.
For example, if we are finally left with 5-10 items I am probably just going to hard them inside the script instead of having a file with an explanation on what is what and why ;)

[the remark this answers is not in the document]
imiric: OK, though I don't see a reason why not to simplify this anyway.

js/bundle_test.go
simskij: > But this would be better checked with `if (typeof Array.prototype.flatten !== 'function') ....`

We should have both of these. First check if it's actually a function, and fast fail if it's no, then whether the behavior is at least somewhat along the lines of what one could expect.

> Also maybe we should use a different object to test this, since flatten was renamed to flat in corejs v3, and it would avoid us having to update this when we update to v3.

`Array.prototype.flat` is the actual name according to the specification. We should probably use that name as well.

js/lib/corejs-build.sh
simskij: 14.x is still the LTS, so as long as it works with that, I wouldn't bother. Usually, `current`, contains weird kinks until it reaches lts.

simskij (approved):
Once @imiric's comments have been resolved. :)

js/lib/k6-shim.js
```diff
@@ -0,0 +1,198 @@
...
+//require('./modules/es6.reflect.has');
+//require('./modules/es6.reflect.is-extensible');
+//require('./modules/es6.reflect.own-keys');
+//require('./modules/es6.reflect.prevent-extensions');
+//require('./modules/es6.reflect.set');
+//require('./modules/es6.reflect.set-prototype-of');
+//require('./modules/es7.array.includes'); // in goja
+require('./modules/es7.array.flat-map');
+// this is now called flat, so maybe drop it https://github.com/tc39/proposal-flatMap 
+require('./modules/es7.array.flatten');
```
simskij, line 141: We should support `Array.prototype.flat` though, although I think the use case might be somewhat limited

simskij:
> es7.array.flatten - same as above + it is now called flat and flatten doesn't exist see https://github.com/tc39/proposal-flatMap

Yeah, we should drop this, once it's in goja. Couldn't find any mention of it in https://github.com/dop251/goja/pull/180

> * es7.reflect.define-metadata and friends are seem to be [rbuckton/reflect-metadata](https://github.com/rbuckton/reflect-metadata) which isn't even a proposal AFAIK ... [medium.com/jspoint/introduction-to-reflect-metadata-package-and-its-ecmascript-proposal-8798405d7d88](https://medium.com/jspoint/introduction-to-reflect-metadata-package-and-its-ecmascript-proposal-8798405d7d88)  So maybe dropping them is a good idea as well

Yeah, if we ever get into any issues because of this, we can add it back or publish an article on how to polyfill it. My bet is that it won't ever happen. 

> es7.object.define-getter, es7.object.define-setter, es7.object.lookup-getter,es7.object.lookup-setter - apparently deprecated, so maybe not even worth implementing 🤷

Agreed. 

So, to sum it up, with som quite minor additions to goja, we should be able to drop the corejs shim layer altogether. 🎉

MStoykov:
As @simskij eluded we have agreed to drop the last two points of https://github.com/loadimpact/k6/pull/1772#issuecomment-745417009 :
*  the metadata one, because it is not in the spec, or even in the process of getting in. Our implementation is 4 years old so it likely has changed, but also is very unlikely to be used by anyone inside k6
* the define/lookup-getter/setter because they have been deprecated for around a decade from what I could find and again are unlikely to be used. 
Both of those can be polyfilled by people who need them in the cases where they do, so even if some users have used them for some reason, there is an easy fix for them.

Everything else has now been implemented in [goja](https://github.com/dop251/goja) and as such we are dropping the whole corejs in #1824, which should land in v0.31.0. 

And as such, I am closing this PR as it superseded.

PR #1775, opened 2020-12-17
Treat errors from Throw() as plain goja values

imiric:
This avoids outputting "GoError" as mentioned in https://github.com/loadimpact/k6/issues/877#issuecomment-484396949 .

I split it off from #1769 to avoid the clutter in that PR.

### DIFF
--- a/js/common/bridge_test.go
+++ b/js/common/bridge_test.go
@@ -348,7 +348,7 @@ func TestBind(t *testing.T) {
 		}},
 		{"Error", bridgeTestErrorType{}, func(t *testing.T, obj interface{}, rt *goja.Runtime) {
 			_, err := rt.RunString(`obj.error()`)
-			assert.Contains(t, err.Error(), "GoError: error")
+			assert.Contains(t, err.Error(), "error")
 		}},
 		{"JSValue", bridgeTestJSValueType{}, func(t *testing.T, obj interface{}, rt *goja.Runtime) {
 			v, err := rt.RunString(`obj.func(1234)`)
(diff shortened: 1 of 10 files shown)


na--:
Hmm I'm not sure this is the way to fix the issue. Even if we don't use goja's `GoError`, we probably should use a custom error wrapper type of our own, so we can then type assert by it.

imiric:
@na-- I tried porting `lib.Exception` added in #1769 and panicking with it in `Throw()`, but it messes up the logged stack trace when `k6.fail()` is used, and we'd have to resort to manually building the stack trace as done in https://github.com/loadimpact/k6/blob/ccffbcaaddd7f2d472cdced0da847fde2140ef88/js/runner.go#L627-L637

This approach doesn't produce the same trace as what goja produces internally. Before (as done in #1769):

```
ERRO[0000] fail: failed
        at github.com/loadimpact/k6/js/common.Bind.func1 (native)
        at another (file:///home/ivan/Documents/Load%20Impact/tests/test-877-fail.js:6:13(4))
        at myfail (file:///home/ivan/Documents/Load%20Impact/tests/test-877-fail.js:14:10(2))
        at file:///home/ivan/Documents/Load%20Impact/tests/test-877-fail.js:28:11(6)  executor=shared-iterations scenario=default
```

After (with the ported `lib.Exception`):

```
ERRO[0000] failed
file:///home/ivan/Documents/Load%20Impact/tests/test-877-fail.js:18:1(26)
```

Where did you consider adding a type assertion for it? I also think an internal error type would make sense here, but not if it messes up the current logging.

na--:
Sorry for the 6-month late response... :sweat_smile: I _think_ we discussed this in Slack, but still should have said something here... :man_facepalming: 

Anyway, prompted by https://github.com/k6io/k6/issues/2049, I just tried your fix (replacing `panic(rt.NewGoError(err))` with `panic(rt.ToValue(err))` in `js.common.Throw()` on top of https://github.com/k6io/k6/pull/2046 and it actually seems to work like a charm! Stack traces are preserved and `GoError` disappears, so we should probably release this even in k6 v0.33.0?

na-- (approved):
LGTM. We can either merge this first and I will rebase https://github.com/k6io/k6/pull/2046, or the other way around, :man_shrugging: will be simpler, probably about the same.

MStoykov (approved):

PR #1776, opened 2020-12-23
Add support for ArrayBuffer request bodies

imiric:
This is part of #1020, just for supporting `ArrayBuffer` request bodies. It works fine for submitting binary data from `open()`, e.g.:

```
var img = open('image.png', 'b');

exports.default = function() {
  var arr = new Uint8Array(img);
  http.post('http://127.0.0.1:9000/postbin', arr.buffer,
            { headers: { 'Content-Type': 'image/png' }});
}
```

It should probably have much more tests, but let me know what other scenarios we should cover. All the other typed arrays?

I wasn't able to get it to work with submitting the typed array itself instead of the `ArrayBuffer`, i.e. submitting `arr` instead of `arr.buffer` above (as mentioned [here](https://github.com/loadimpact/k6/issues/1020#issuecomment-727911302)). Goja's `Export()` on it returns an empty `map[string]interface{}{}`, so I couldn't get the `ArrayBuffer` from it. We should probably raise a warning in that case, unless we can work around it.

### DIFF
--- a/js/modules/k6/http/request.go
+++ b/js/modules/k6/http/request.go
@@ -222,6 +222,8 @@ func (h *HTTP) parseRequest(
 			if err := handleObjectBody(newData); err != nil {
 				return nil, err
 			}
+		case goja.ArrayBuffer:
+			result.Body = bytes.NewBuffer(data.Bytes())
 		case map[string]interface{}:
 			if err := handleObjectBody(data); err != nil {
 				return nil, err
--- a/js/modules/k6/http/request_test.go
+++ b/js/modules/k6/http/request_test.go
@@ -1416,6 +1416,45 @@ func TestRequestAndBatch(t *testing.T) {
 	})
 }
 
+func TestRequestArrayBufferBody(t *testing.T) {
+	t.Parallel()
+	tb, _, _, rt, _ := newRuntime(t) //nolint: dogsled
+	defer tb.Cleanup()
+	sr := tb.Replacer.Replace
+
+	tb.Mux.HandleFunc("/post-arraybuffer", http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		require.Equal(t, "POST", r.Method)
+		var in bytes.Buffer
+		_, err := io.Copy(&in, r.Body)
+		require.NoError(t, err)
+		_, err = w.Write(in.Bytes())
+		require.NoError(t, err)
+	}))
+
+	testCases := []struct {
+		arr, expected string
+	}{
+		{"Uint8Array", "104,101,108,108,111"},
+		{"Uint16Array", "104,0,101,0,108,0,108,0,111,0"},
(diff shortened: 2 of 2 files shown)


MStoykov (changes requested):
LGTM!

I think @na-- and I [agreed upon just supporting ArrayBuffer](https://github.com/loadimpact/k6/issues/1020#issuecomment-727914267) for now. The Support for Int8Array(for example) seemed a little funky and at least for an MVP it is fine especially if we can error out to the user correctly.

Unfortunately, as you have noticed if someone provides `Int8Array` it enters the `map[string]interface{}` case with 0 keys so changing the error in the `default` won't be sufficient. IMO there is no case where this is what the user wanted - so we can error on that one as well (possibly in the `handleObjectBody`)?

The `map[string]goja.Value` is only used from the `SubmitForm` code it seems? Is there a case where we could want to submit a form with no data? I think not, I think it needs at least some info inside of it or am I wrong?

imiric:
@MStoykov I don't think it would be safe to error out in that case, as we have no way to determine whether the user passed a typed array or an empty object (that I can see). If anything it should probably be logged as a warning, but given that we can't distinguish between those scenarios I'm inclined to leave it as is.

@na-- What do you think?

na--:
Hmm it'd have been nice if we could prevent footguns like that, but if there doesn't seem to be a good way to differentiate the typed arrays from other potential arguments, I'm not sure if there's anything we can do about it :man_shrugging: 

@MStoykov, goja [seems to have](https://github.com/dop251/goja/blob/27b0a7dc4c7fab42694a39c77da455a586da9173/typedarrays.go#L847-L855) custom `export()` and `exportType()` methods for `arrayBufferObject`, but not for [`typedArrayObject`](https://github.com/dop251/goja/blob/27b0a7dc4c7fab42694a39c77da455a586da9173/typedarrays.go#L68). Maybe that's why they get exported as `map[string]interface{}`, some default export for all JS objects? If so, it might be worth to make a small goja PR that adds these methods, which would then allow us to differentiate and work with these types in Go?

Other than that, the code of the PR mostly LGTM. I'd only advise moving the test to its own standalone function, I don't see why we have to cram it into the already huge composite `http` test we have.

MStoykov:
>  @MStoykov, goja seems to have custom export() and exportType() methods for arrayBufferObject, but not for typedArrayObject. Maybe that's why they get exported as map[string]interface{}, some default export for all JS objects? If so, it might be worth to make a small goja PR that adds these methods, which would then allow us to differentiate and work with these types in Go?

What do you propose they are exported to? ArrayBuffer?

MStoykov (approved):

na--:
> What do you propose they are exported to? ArrayBuffer?

No, their constituent types, see [this](https://github.com/dop251/goja/blob/27b0a7dc4c7fab42694a39c77da455a586da9173/typedarrays.go#L58-L66)

MStoykov (approved):

na-- (approved):

PR #1782, opened 2021-01-06
Update goja to latest master

MStoykov:


js/lib/lib.go
```diff
@@ -37,11 +38,17 @@ var (
...
-			true,
-		)
+			parser.WithDisableSourceMaps)
+		if err != nil {
+			panic(err)
+		}
+		coreJs, err = goja.CompileAST(program, false)
+		if err != nil {
+			panic(err)
+		}
```
imiric, line 51: Why was this done? To disable sourcemap support and improve performance?

In any case it should be in a separate commit from the goja update that explains it.

imiric (changes requested):

js/lib/lib.go
MStoykov: Unfortunately goja fixing sourcemaps support (which was apperantly a little broken) made it so now if there are sourcemap mappings it errors out. Given that we don't support sourcemaps I decided that it is best to disable it. 

Supporting them will need more work especially around archives and ... I hoped that given #1772  and the now initiative to just put the rest in goja, that I will reduce somewhat the amount of ... work needed.

tl;dr this is needed in order for the update to compile corejs at all and I disabled it elsewhere because it won't work with `k6 archive` but it also didn't work until now.

MStoykov: I added an issue https://github.com/loadimpact/k6/issues/1789 as I forgot at the time

imiric: OK, thanks for explaining. It still belongs in a separate commit with that explanation ;)

MStoykov: I might not have stressed this enough - the tests don't pass if I don't change this as the core-js shim has sourceMap tag/line in it

imiric: That's fine, I'm only arguing for including it as a separate commit on top the the goja update one, so we can document why it was done, instead of it being mixed with the goja changes which I usually gloss over.

MStoykov: I can add it to the message,which I should've probably done :facepalm: , but I don't like it being in a separate commit

na--: :+1: for a commit message and a single commit, if the tests will be broken without the change. That said, are we sure that we didn't have source maps before? The line numbers of errors weren't that far off the mark, despite us having core.js and babel transformation... IIRC, they were only off by 1 because of https://github.com/loadimpact/k6/blob/ccffbcaaddd7f2d472cdced0da847fde2140ef88/js/initcontext.go#L198-L202

:confused:

MStoykov: I am pretty sure because they were [disabled LONG time ago in babel](https://github.com/loadimpact/k6/commit/31819e6ffe966fc247f58d3e717e41037068747a) and the reason we are not seeing bigger differences is because [babel tries](https://github.com/loadimpact/k6/blob/ef717e1af2a92d35d39a3d815a69b6ab6b8c65e1/js/compiler/compiler.go#L45) :).

MStoykov: it *tries* because it works terrible for the first 10-15 lines if I remember correctly

na--: ok then, :+1: for this and we can figure out proper source maps in the future in https://github.com/loadimpact/k6/issues/1789, after we get rid of core.js

vendor/github.com/dop251/goja/README.md
```diff
@@ -212,34 +212,8 @@ There are two standard mappers: [TagFieldNameMapper](https://godoc.org/github.co
...
-```
-
-Then it can be used in JS as follows:
-
-```js
-var o = new MyObject(arg);
-var o1 = MyObject(arg); // same thing
-o instanceof MyObject && o1 instanceof MyObject; // true
-```
+In order to implement a constructor function in Go use `func (goja.ConstructorCall) *goja.Object`.
```
na--, line 215: Hmm how does this affect our `XConstructor` magic?

MStoykov: In no way ... as [we are not using it](https://github.com/loadimpact/k6/blob/master/js/common/bridge.go#L258-L263)

But also this seems as only a documentation chang

na--: But apparently there's a native way to do this in goja which we don't use. Should we open an issue to investigate this, given that this is the code above our code to do the same :sweat_smile: https://github.com/loadimpact/k6/blob/ccffbcaaddd7f2d472cdced0da847fde2140ef88/js/common/bridge.go#L257

MStoykov: I don't think I will come to this until we decide to rewrite the bridge :D ... so if you think you will come to this refactoring work before that ... or think someone else will do it - open an issue ;)

na-- (approved):
LGTM. I see `Symbol` is now exported, which I remember being a blocker for something?

MStoykov:
It is only part of the blocker https://github.com/dop251/goja/issues/227

imiric (approved):

PR #1783, opened 2021-01-06
Add git-lfs to dockerfile

HoneyryderChuck:
It's fairly common to store load testing images in more efficient storage like git lfs. This wouldn't increase the size of the intermediate container by much, as well.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 FROM golang:1.15-alpine as builder
 WORKDIR $GOPATH/src/github.com/loadimpact/k6
 ADD . .
-RUN apk --no-cache add git
+RUN apk --no-cache add git git-lfs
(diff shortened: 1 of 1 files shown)


na-- (changes requested):
Hmm this is not going to add `git-lfs` to the `loadimpact/k6` Docker image. Our `Dockerfile` is a [multi-stage one](https://docs.docker.com/develop/develop-images/multistage-build/), and this change will only apply for the first `builder` stage, not the final image. And we don't need `git-lfs` in the `builder` stage, only plain git.

That said, I don't think it's a good idea to add either `git` or `git-lfs` to the `loadimpact/k6` image either, so I'll reject this PR, sorry. Most people probably won't need them, so they'll just bloat the image... The people who need them can easily add them in their own Docker image, based on the official k6 one, like this:
```dockerfile
FROM loadimpact/k6
RUN apk --no-cache add git git-lfs
```

HoneyryderChuck:
this was meant for a gitlab docker executor situation, where we'd just script around the base image. but fair enough.

PR #1785, opened 2021-01-08
Add go-header linter to check the license header

MStoykov:


.license-template
```diff
@@ -0,0 +1,17 @@
+*
```
MStoykov, line 1: yes this needs to not have space before it, while all the others need it and yes we need the `*`

.golangci.yml
```diff
@@ -47,6 +47,13 @@ linters-settings:
   funlen:
     lines: 80
     statements: 60
+  goheader:
+    template-path: ".license-template"
+    values:
+      const:
+        year: 2021
+      regexp:
+        year-range: (\d\d\d\d|{{year}})
```
MStoykov, line 56: this probably needs a better place and is here so we don't need to update all the files that have a year before 2021/2020. 
After https://github.com/denis-tingaikin/go-header/issues/2#issue-552648515 is fixed we can(but might still be better to not do it) just use the default `year` value(that was added later ... then what we use in golangci-lint) and let it fix it every new year :)

na-- (approved):

.license-template
imiric: Ah, nice catch. I tried to setup go-header a few months ago and it kept failing for some reason, but this was probably it.

PR #1793, opened 2021-01-12
temporarily disable goheader linter

MStoykov:


MStoykov:
It seems like this flaky in the github actions and from my playing around it is something with the golangci-lint version as even locally it is flaky with v1.31.0 but with v1.35.2 it seems to work a lot better.

We can alternatively update the version and comb through the new linters and disable the ones we don't want, but maybe we should do that after v0.30.0 is released?

na-- (approved):
LGTM My vote is disable it for now and update golangci-lint after v0.30.0

imiric (approved):

PR #1794, opened 2021-01-12
Set Docker workdir to a writeable path

imiric:
Inspired by [this question](https://stackoverflow.com/q/65678315/96213) and similar we've had on the forum.

This might break some use cases, but it's safer to have this set in the final image than not, and it can be easily overriden.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,4 +10,5 @@ RUN apk add --no-cache ca-certificates && \
 COPY --from=builder /go/bin/k6 /usr/bin/k6
 
 USER 12345
+WORKDIR /home/k6
 ENTRYPOINT ["k6"]


MStoykov:
I am fine with this change ... although I wonder if it will then not move to 

"where is my file that k6 generated? I told it to put in the local directory"

MStoykov (approved):

na-- (approved):
LGTM, though I'll tag it with the new `breaking change` label, so we mention it in the release notes

imiric:
@MStoykov Those users are probably running the image as root then, and should be aware that they can override this with `--workdir`. But I suspect those will be a minority, and based on these questions so far, more people are being tripped up by this default behavior.

PR #1796, opened 2021-01-13
We actually need to disable sourceMaps when we compile them as well

MStoykov:


na-- (dismissed):
LGTM, why not update goja as well?

MStoykov:
I wanted to also enable some tc39 tests with a new version and do it in another PR. But I already started running the tests so I can add to this one if you want to :)

na--:
However you prefer to do it, but :+1: for a goja update today from me

na--:
btw I have no idea how source maps work precisely, but would this change be too difficult to test?

MStoykov:
It literally cares about a comment at the end of a source file
```javascript
//# sourceMappingURL=ajv.min.js.map
```

na-- (approved):

imiric (approved):
LGTM, but this is breaking change, no? Sourcemaps were supported before?

MStoykov:
if they at any point worked in any capacity it wasn't intentional ;) 
Given the issues listed (and now fixed) in https://github.com/dop251/goja/issues/235  I would expect  that they never did for us

Issue #1571, opened 2020-07-23
Multipart request with file array does not work as expected


mattiaforc:
Hi everyone,
I'm struggling with `multipart/form-data` when making requests with arrays of file as content. 
My backend accepts a multipart request with two keys: `files` and `info`. The former is an array of files (with MIME type `application/pdf`) and the latter is just a JSON object with some metadata (which is negligible for this situation). 
The problem is that I can not create a multipart request with multiple files, because it sends just one file.

## Environment
- k6 version: k6 v0.27.0 (2020-07-14T12:19:13+0000/v0.27.0-0-g6fa889d0, go1.14.4, linux/amd64)
- OS and version: Ubuntu 18.04.4 LTS, Kernel: Linux 5.4.0-42-generic, Architecture: x86-64

## Expected Behavior

Just like in Postman requests, I expect to be able to define multiple files under the same key `files`, which are sent and then received by the backend as an array of files. 
![image](https://user-images.githubusercontent.com/11019007/88288354-f5223380-ccf3-11ea-8508-25e46724670d.png)

This is also possible in JS (according to [MDN web docs](https://developer.mozilla.org/en-US/docs/Web/API/FormData/append), see also [this SO question](https://stackoverflow.com/questions/12989442/uploading-multiple-files-using-formdata)):
> The `append()` method of the `FormData` interface appends a new value onto an existing key inside a `FormData` object, or adds the key if it does not already exist. The difference between `FormData.set` and `append()` is that if the specified key already exists, `FormData.set` will overwrite all existing values with the new one, whereas `append()` will append the new value onto the end of the existing set of values.

## Actual Behavior

My guess is that this is the most correct code (note that the file is read during `init` phase, and used later on in the rest of the snippet) 
```js
// init
const file1 = open('loadTestFile1.pdf', 'b')
const file2 = open('loadTestFile2.pdf', 'b')

// actual VU function
let files = [
    http.file(file1, "loadTestFile1.pdf", "application/pdf"), 
    http.file(file2, "loadTestFile2.pdf", "application/pdf")
]

let info = http.file(JSON.stringify({
      "irrelevant": 1234
}), "info", "application/json")

let payloadData = { files, info }

let res = http.post(url, payloadData, { headers: { 'Authorization': 'Basic value' } })
```
which results in a single larger `files` object. This is a problem not only because it creates only one object but also due to my backend file size limit (10MB) being exceeded (note that the limit is *per-file*, so I have to be able to send two 8MB files).

[...]

Issue #843, opened 2018-11-14
K6 Does not respect the boundary value placed in a Content-Type header


Ardesco:
If you specify a Context-Type header for a multipart form with a specific boundary, it is ignored:

```
let parameters = {
        headers: {
            "Authorization": "someAuthToken",
            "Content-Type": "multipart/form-data; boundary=--myBoundary",
            "Accept-Encoding": "br, gzip, deflate"
        }
    };
```

When looking at the logged request, you get:

```
POST /some/url HTTP/1.1
Host: api.feat1.dev.imaginecurve.com
User-Agent: k6/0.22.1 (https://k6.io/);
Content-Length: 507
Accept-Encoding: br, gzip, deflate
Authorization: someAuthToken
Content-Type: multipart/form-data; boundary=--myBoundary
Cookie: device_view=full

--87154a0502b4f0b523bb4fc2571b3d87d741f6e7ca73baaad7f8ab5eb74a
Content-Disposition: form-data; name="foo"

foo content
--87154a0502b4f0b523bb4fc2571b3d87d741f6e7ca73baaad7f8ab5eb74a
Content-Disposition: form-data; name="bar"

bar content
--87154a0502b4f0b523bb4fc2571b3d87d741f6e7ca73baaad7f8ab5eb74a
Content-Disposition: form-data; name="_dummyFile"; filename="placeholder"
Content-Type: application/octet-stream


--87154a0502b4f0b523bb4fc2571b3d87d741f6e7ca73baaad7f8ab5eb74a--
```

Note that the boundary separator is not the one specified in the Content-Type.  This results in clients being unable to parse the payload because it can't find the specified boundary.

I would suggest this is linked to #747

na--:
Thanks, you're right, we should fix this at the same time we fix https://github.com/loadimpact/k6/issues/747, since both issues are very connected!

IsabelaPastorini:
In the case that I **MUST** change manually the 'Content-type' and then add a random boundary value doesn't work. My scenario is:

- I receive from a flowContext my header params and the content-type is set automatically to 'application/json'
- Since I'm testing an endpoint that requires a 'multipart/form-data' I have to manually change the header content-type and added a random boundary value.
```
let binPdf = open("../bin-pdf.pdf", "b");
param.headers['Content-Type'] = "multipart/form-data;boundary=random-boundary-value";
        var data = {
            file: http.file(binPdf, "bin-pdf.pdf", "application/pdf") 
        };
[...]

Issue #747, opened 2018-08-20
Better handling of multipart/form-data requests


na--:
A user asked a [question](https://k6io.slack.com/archives/C3E688J8Z/p1534745637000100) about `multipart/form-data` requests in the slack chat and I found out that users cannot directly make `multipart/form-data` POST requests. The following script produces an invalid request:
```js
import http from 'k6/http';

export default function () {
   let resp = http.post(
       "https://httpbingo.org/post",
       { username: "testuser", password: "testpass" },
       { headers: {"Content-Type": "multipart/form-data"} }
   );
   console.log(resp.body);
};
```
It seems like it sends `multipart/form-data` as the `Content-Type` header, but sends the actual POST data like the request was a normal `application/x-www-form-urlencoded` HTTP request.

We only correctly use send `multipart/form-data` requests when there's a file involved, so as I mentioned in the slack chat, this works as an ugly workaround:
```js
import http from 'k6/http';

export default function () {
   let resp = http.post(
       "https://httpbingo.org/post",
       { username: "testuser", password: "testpass", _dummyFile: http.file("", "placeholder") },
   );
   console.log(resp.json().headers["Content-Type"]);
};
```
It would output something like `multipart/form-data; boundary=some_random_boundary...`. It seems like k6 automatically switches to `multipart/form-data` content-type when there are files involved, so when we try to send a dummy file as well as the username and password it works. And of course users can also manually build the HTTP request body, but it's a huge PITA:

```js
import http from 'k6/http';

export default function () {
    let resp = http.post(
        "https://httpbingo.org/post",
        "--my-awesome-handcrafted-boundary\r\nContent-Disposition: form-data; name=\"username\"\r\n\r\ntestuser\r\n--my-awesome-handcrafted-boundary\r\nContent-Disposition: form-data; name=\"password\"\r\n\r\ntestpass\r\n--my-awesome-handcrafted-boundary--\r\n",
        { headers: { "Content-Type": "multipart/form-data; boundary=my-awesome-handcrafted-boundary" } }
    );
    let respJson = resp.json();
    console.log(respJson.headers["Content-Type"]);
    console.log(respJson.form.username);
    console.log(respJson.form.password);
};
```

Also, when I decided to see how the HAR converter deals with `multipart/form-data`, I saw that it just [skips them](https://github.com/loadimpact/k6/blob/5ed5fb05ad734957128939d9fcd61a977f39c503/converter/har/converter.go#L120-L123), so we should fix that as well...

[...]

PR #1800, opened 2021-01-14
Expand ArrayBuffer support

imiric:
This addresses the backwards compatible changes of #1020, and showcases a workaround for better handling of multipart requests (ref. #747, #843, #1571).

### DIFF
--- a/js/initcontext.go
+++ b/js/initcontext.go
@@ -204,7 +204,8 @@ func (i *InitContext) compileImport(src, filename string) (*goja.Program, error)
 	return pgm, err
 }
 
-// Open implements open() in the init context and will read and return the contents of a file
+// Open implements open() in the init context and will read and return the contents of a file.
+// If the second argument is "b" it returns the data as a binary array, otherwise as a string.
 func (i *InitContext) Open(ctx context.Context, filename string, args ...string) (goja.Value, error) {
 	if lib.GetState(ctx) != nil {
 		return nil, errors.New(openCantBeUsedOutsideInitContextMsg)
--- a/js/modules/k6/crypto/crypto.go
+++ b/js/modules/k6/crypto/crypto.go
@@ -56,6 +56,7 @@ func New() *Crypto {
 	return &Crypto{}
 }
 
+// RandomBytes returns random data of the given size.
 func (*Crypto) RandomBytes(ctx context.Context, size int) []byte {
 	if size < 1 {
 		common.Throw(common.GetRuntime(ctx), errors.New("invalid size"))
@@ -68,60 +69,70 @@ func (*Crypto) RandomBytes(ctx context.Context, size int) []byte {
 	return bytes
 }
 
+// Md4 returns the MD4 hash of input in the given encoding.
 func (c *Crypto) Md4(ctx context.Context, input interface{}, outputEncoding string) interface{} {
 	hasher := c.CreateHash(ctx, "md4")
 	hasher.Update(input)
 	return hasher.Digest(outputEncoding)
 }
 
+// Md5 returns the MD5 hash of input in the given encoding.
 func (c *Crypto) Md5(ctx context.Context, input interface{}, outputEncoding string) interface{} {
 	hasher := c.CreateHash(ctx, "md5")
 	hasher.Update(input)
 	return hasher.Digest(outputEncoding)
 }
 
+// Sha1 returns the SHA1 hash of input in the given encoding.
 func (c *Crypto) Sha1(ctx context.Context, input interface{}, outputEncoding string) interface{} {
 	hasher := c.CreateHash(ctx, "sha1")
 	hasher.Update(input)
 	return hasher.Digest(outputEncoding)
 }
 
+// Sha256 returns the SHA256 hash of input in the given encoding.
 func (c *Crypto) Sha256(ctx context.Context, input interface{}, outputEncoding string) interface{} {
 	hasher := c.CreateHash(ctx, "sha256")
 	hasher.Update(input)
 	return hasher.Digest(outputEncoding)
 }
 
+// Sha384 returns the SHA384 hash of input in the given encoding.
 func (c *Crypto) Sha384(ctx context.Context, input interface{}, outputEncoding string) interface{} {
 	hasher := c.CreateHash(ctx, "sha384")
 	hasher.Update(input)
 	return hasher.Digest(outputEncoding)
 }
 
+// Sha512 returns the SHA512 hash of input in the given encoding.
 func (c *Crypto) Sha512(ctx context.Context, input interface{}, outputEncoding string) interface{} {
 	hasher := c.CreateHash(ctx, "sha512")
 	hasher.Update(input)
 	return hasher.Digest(outputEncoding)
 }
 
+// Sha512_224 returns the SHA512/224 hash of input in the given encoding.
 func (c *Crypto) Sha512_224(ctx context.Context, input interface{}, outputEncoding string) interface{} {
 	hasher := c.CreateHash(ctx, "sha512_224")
 	hasher.Update(input)
 	return hasher.Digest(outputEncoding)
 }
 
+// Sha512_256 returns the SHA512/256 hash of input in the given encoding.
 func (c *Crypto) Sha512_256(ctx context.Context, input interface{}, outputEncoding string) interface{} {
 	hasher := c.CreateHash(ctx, "sha512_256")
 	hasher.Update(input)
 	return hasher.Digest(outputEncoding)
 }
 
+// Ripemd160 returns the RIPEMD160 hash of input in the given encoding.
 func (c *Crypto) Ripemd160(ctx context.Context, input interface{}, outputEncoding string) interface{} {
 	hasher := c.CreateHash(ctx, "ripemd160")
 	hasher.Update(input)
 	return hasher.Digest(outputEncoding)
 }
 
+// CreateHash returns a Hasher instance that uses the given algorithm.
 func (*Crypto) CreateHash(ctx context.Context, algorithm string) *Hasher {
 	hasher := Hasher{}
 	hasher.ctx = ctx
@@ -162,6 +173,7 @@ func (hasher *Hasher) Update(input interface{}) {
 	}
 }
 
+// Digest returns the hash value in the given encoding.
 func (hasher *Hasher) Digest(outputEncoding string) interface{} {
 	sum := hasher.hash.Sum(nil)
 
@@ -235,6 +247,8 @@ func (c Crypto) CreateHMAC(ctx context.Context, algorithm string, key interface{
 	return &hasher
 }
 
+// Hmac returns a new HMAC hash of input using the given algorithm and key
+// in the given encoding.
 func (c *Crypto) Hmac(
 	ctx context.Context, algorithm string, key, input interface{}, outputEncoding string,
 ) interface{} {
--- a/js/modules/k6/crypto/crypto.go
+++ b/js/modules/k6/crypto/crypto.go
@@ -68,55 +68,55 @@ func (*Crypto) RandomBytes(ctx context.Context, size int) []byte {
 	return bytes
 }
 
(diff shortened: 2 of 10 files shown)


js/initcontext.go
```diff
@@ -238,7 +240,8 @@ func (i *InitContext) Open(ctx context.Context, filename string, args ...string)
 	}
 
 	if len(args) > 0 && args[0] == "b" {
-		return i.runtime.ToValue(data), nil
+		b := i.runtime.NewArrayBuffer(data)
```
na--, line 243: hmm while returning the JS equivalent to `[]byte` from `open()` is not ideal, it was somewhat usable. So I am not sure if we should break it, since some scripts probably depend on the current strange behavior. Why not just add another flag value for an `ArrayBuffer` response?

na--:
> Currently `lib/netext/httpext/` doesn't depend on goja, but more importantly I don't see where we could initialize a throwaway Runtime for this purpose, since doing that in `readResponseBody` or `MakeRequest` would be too expensive. We'd have to pre-initialize it in the context or `lib.State`. :-/

Technically, there's a JS wrapper around the `httpext.Response`, though it's a very thin one at the moment: https://github.com/loadimpact/k6/blob/c00a85bfea21d748f0f9d11fdbc326087f34c411/js/modules/k6/http/response.go#L36-L44

And here (and in `batch()`) you should be able to access the Runtime with `common.GetRuntime(ctx)`: https://github.com/loadimpact/k6/blob/c00a85bfea21d748f0f9d11fdbc326087f34c411/js/modules/k6/http/request.go#L116

But yeah, `ArrayBuffer` as the `Response.body` should wait for v0.31.0, it's too complicated a change at this stage. I am even hesitating if we should merge this PR or wait for it as well... @MStoykov what do you think?

> `Uint8Array(res.body)`

This is a viable workaround, but we have to benchmark it... I think it would probably involve copying the whole body, so not great about CPU and memory usage. Better than nothing, but I hope we could find something more efficient.

js/initcontext.go
imiric: Yeah, I also wasn't sure if we should break it.

> Why not just add another flag value for an `ArrayBuffer` response?

:-/ Sure, maybe `ab`?

imiric:
@na-- Thanks, I'll take a look at accessing the Runtime that way.

As for delaying this PR, give me the rest of the day to address these two issues. If we can't get it merged today or early tomorrow, or @MStoykov finds other issues, I'm OK with postponing it.

js/initcontext.go
imiric: Done, see c593025f.

imiric:
I added the response body `ArrayBuffer` support somewhat in a hurry, especially the tests, so let me know if it can be improved.

lib/netext/httpext/response.go
```diff
@@ -35,7 +35,7 @@ import (
 // ResponseType is used in the request to specify how the response body should be treated
 // The conversion and validation methods are auto-generated with https://github.com/alvaroloes/enumer:
 //nolint: lll
-//go:generate enumer -type=ResponseType -transform=snake -json -text -trimprefix ResponseType -output response_type_gen.go
+//go:generate enumer -type=ResponseType -transform=title-lower -json -text -trimprefix ResponseType -output response_type_gen.go
```
imiric, line 38: This depends on [this enumer fork](https://github.com/dmarkham/enumer) to generate `arrayBuffer` properly, so check the `*_gen.go` files changes. The older [alvaroloes/enumer](https://github.com/alvaroloes/enumer) seems unmaintained.

I included the changed `compression_type_gen.go` as well, since I ran `go generate ./lib/netext/httpext/`, but I could include just the `response_type_gen.go` changes if you prefer.

MStoykov:
I am fairly hard for this being post v0.30.0

I am also of the opinion that we should:
1. start with things that take as arguments `[]byte` to make them work with `ArrayBuffer` instead of start with things that return `[]byte` to return `ArrayBuffer`.
2. Possibly break all the current things returning `[]byte` to return `ArrayBuffer` after the previous step is done...

As far as I am concern having 2 binary formats, one of which(`[]byte`) is terrible for usage and not JS-y, is way worse than the  (unlikely) breaking of some user scripts(that will also be fixed by a 1 lines). 

Also ... I have barely seen anyone use the `binary` "open" or HTTP responses, and likely they are used just to push them back into something(HTTP, crypto, encoding), on this I am pretty certain because you barely can do anything else with it either way. So as long as those work with `ArrayBuffer` this will be seamless for people, without us needing to explain that ... "yeah you should just use ArrayBuffer, more or less always, but by default, we shoot you in the foot so we don't break a script we haven't seen that likely was broken and hacky to begin with"

imiric:
@na-- @MStoykov I added the k6/crypto and k6/encoding changes to support ArrayBuffer, so I think that covers everything mentioned in #1020. I added a couple of new tests, but mostly edited existing ones to deal with ArrayBuffer properly. Let me know if more tests are missing. `ToBytes()` should probably be unit tested, etc.

We should discuss the breaking changes here, if having `open()` always return `ArrayBuffer` is worth it, etc. I agree with Mihail that we'd probably be fine with that for most use cases (e.g. that pass the return value directly to `http.post()`, etc.), but it will break scripts that did some manual work on `[]byte`, and will be annoying to fix. We probably should anyway, with a big disclaimer in the release notes and docs?

na--:
I haven't looked at the latest code yet, but have some comments that I hope are not too uninformed...

> it will break scripts that did some manual work on []byte, and will be annoying to fix.

I am coming around to `open(filename, 'b')` returning an `ArrayBuffer`, i.e. having a breaking change for k6 v0.31.0. Mostly because the fix would probably be quite easy, just wrapping the `open()` call in `new Uint8Array()` and then using `.buffer` in the HTTP request call. Or at least that's what I think... 

On another topic, as you know I've gone through most of the old k6 issues, and here is a list of ones related to binary handling or request bodies that I found: https://github.com/loadimpact/k6/issues/747, https://github.com/loadimpact/k6/issues/843, https://github.com/loadimpact/k6/issues/1375, https://github.com/loadimpact/k6/issues/1382, https://github.com/loadimpact/k6/issues/1571, https://github.com/loadimpact/k6/issues/1770, https://github.com/loadimpact/k6/issues/1798

I didn't spend a ton of time investigating each issue, but I think that after this PR and https://github.com/loadimpact/jslib.k6.io/pull/21 are merged, we should go through all of them with a fine-toothed comb. If an issue could be solved by the `FormData` JS builder, we should post a comment on how to do it and then close the issue. I also made a note that we should add some proper `FormData` examples in the docs: https://github.com/loadimpact/k6-docs/issues/203

MStoykov:
> I am coming around to open(filename, 'b') returning an ArrayBuffer, i.e. having a breaking change for k6 v0.31.0. Mostly because the fix would probably be quite easy, just wrapping the open() call in new Uint8Array() and then using .buffer in the HTTP request call. Or at least that's what I think...

you don't need to wrap it or anything - we should be taking `ArrayBuffer` everywhere we are currently taking `[]byte` so if your code is 
```
var b = open("something.dat", "b");
export default function() {
   http.post(url, b);
}
```

That will just work if we change to just returning `ArrayBuffer`. 
Which was my whole proposal:
1. wherever we are taking `[]byte` take `Arraybuffer` (at least make it possible to take `ArrayBuffer` as well)
2. wherever we are returning `[]byte`, just return `ArrayBuffer`

as long as users just pushed `[]byte` around it will just work - no breaking change.

The breaking change comes from the fact that you can take the `length` of a `[]byte` as well as modify individual elements with `b[1] = 12` or just get them, but you can't change the length of the `[]byte` either up or down without it becoming `[]interface{}` which then can't be used anywhere :sob: . ArrayBuffer on the other side has `byteLength` and you need a view like `Uint8Array` to actually edit the values. 

Given my experiments with `[]byte` trying to fix this before we introduced taking `ArrayBuffer` as an HTTP body ... you can't do much with the above things, and I am certain that whoever hacked something with it will be glad to use ArrayBuffer instead.

My main concern actually is that `length` is likely to be used in checks and such and will just now ... not work and possibly it will just fail checks without anything. 
But I am still learning (even not heavily) toward just biting the bullet and breaking this, so we don't have 2 binary data supports, where the default is useless and not JS-y.

 I agree on the issues, I also would prefer if we go through the documentation of the current "magical" ways to use formdata and document that they are not great and possibly should be abandoned for the FormData polyfill. After some testing of the polyfill of course :)

na--:
@MStoykov, I know, I was responding specifically to this sentence by @imiric :wink: 
> it will break scripts that did some manual work on []byte, and will be annoying to fix.

js/modules/k6/crypto/crypto.go
```diff
@@ -56,16 +58,19 @@ func New() *Crypto {
...
 	if err != nil {
-		common.Throw(common.GetRuntime(ctx), err)
+		common.Throw(rt, err)
 	}
-	return bytes
+	ab := rt.NewArrayBuffer(bytes)
+	return &ab
 }
 
 func (c *Crypto) Md4(ctx context.Context, input []byte, outputEncoding string) interface{} {
```
MStoykov, line 76: This and every other shorthand function also needs to take `ArrayBuffer`

MStoykov:
Also a relevant problem to some of the changes https://github.com/loadimpact/k6/issues/1770#issuecomment-745133170

If we are taking `[]byte` and the user provides us with a `string` depending on whether it is [asciiString](https://github.com/dop251/goja/blob/33737909d64ebd785cb4a59435242701b153bde6/string_ascii.go#L15) or [unicodeString](https://github.com/dop251/goja/blob/33737909d64ebd785cb4a59435242701b153bde6/string_unicode.go#L20) we are then doing crypto on either utf8 or utf16 strings ... which as you can guess have totally different outputs. 

So we might need to change this to always go to utf8 for now and add encoding argument, which will be more in line with what [nodejs does](https://nodejs.org/api/crypto.html#crypto_hash_update_data_inputencoding)

MStoykov (changes requested):
I am of the opinion that we should drop the changes where we add a way to return `ArrayBuffer` And keep the rest where an `ArrayBuffer` is now an appropriate argument for stuff.

This way we can see and test better if we can completely drop returning `[]byte` for returning `ArrayBuffer` instead in it's place. This IMO will require some writing of scripts, possibly going through issues connected with it and trying them.

I myself went down the rabbit whole of what encoding is a string if we take it as `[]byte` vs as `string` in argument which ate too much time. 

So as I am of the opinion that taking ArrayBuffer is more important and "probably" backwards compatible it should be done first, and then we can figure out what we do next from there

imiric:
@MStoykov Makes sense, I'll revert some of the changes then.

imiric:
Reverting got messy, so I reworked the commits and added some tests. Let me know.

js/modules/k6/crypto/crypto.go
imiric: Resolved in latest push.

MStoykov (approved):
LGTM, I would want to play with it more, but given https://github.com/loadimpact/k6/issues/1770#issuecomment-770876684 my main concern is gone so I am fine with this getting merged and fixing it more later

imiric:
@MStoykov @na-- What do you think about changing `b64decode` (https://github.com/loadimpact/k6/pull/1800#issuecomment-769955641)? Right now you can `b64encode` an `ArrayBuffer`, but you can't get it back from `b64decode`, which seems rather limiting. Same goes for `[]byte`, though I guess if no one has asked for it, the current API is fine?

MStoykov:
I think we should have an issue where we discuss those breaking changes.

I am of the opinion it is the correct thing for them to return `ArrayBuffer`, and breaking it will possibly help users more than not. But maybe we should see some user scripts that use it and whether at least some of them will keep working even if we change them. I haven't come across many, as the ones that have been reported *don't* work for this exact reason so :man_shrugging: 

Do you want to try and make an issue for this with some code showing usage that will:
1. keep working
2. break because we now don't return a string
3. will now work(this should be the easiest part as there is code that isn't working currently exactly because of this ;) )

na-- (approved):
LGTM

PR #1803, opened 2021-01-15
Cleanup and handleSummaryResult() tests

na--:
There were a few other things I wondered if I should fix in the current Go summary code, but at this point in time I'll probably just fix them in the upcoming JS replacement... For example, this code: https://github.com/loadimpact/k6/blob/c00a85bfea21d748f0f9d11fdbc326087f34c411/ui/summary.go#L91 

Should be
```go
if !inEscSeq && !inLongEscSeq {
	n++
}
```
i.e. that `StrWidth()` function is completely broken... :man_facepalming: But this turns not to matter one bit, since almost all of the errors are by the same margin and mostly cancel out... :rofl: Not to mention that the Unicode normalization it tries to do might actually be worse than not doing anything at all: https://github.com/loadimpact/k6/blob/c00a85bfea21d748f0f9d11fdbc326087f34c411/ui/summary.go#L64-L65

See [this MDN article](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/normalize) for a good explanation of the various normalization modes, and why `NFC` or `NFKC` (or even doing nothing) might be better for our use case than `NFKD`. On the other hand, not even they are sufficient, so maybe it doesn't really matter and everything in life is pointless: https://hsivonen.fi/string-length/ :sob: :sweat_smile: :rofl: 

Anyway, I think this is the final summary-related PR before v0.30.0.

### DIFF
--- a/js/summary.go
+++ b/js/summary.go
@@ -36,10 +36,12 @@ import (
 // TODO: move this to a separate JS file and use go.rice to embed it
 const summaryWrapperLambdaCode = `
 (function() {
-	var forEach = function(obj, callback) {
+	var forEach = function (obj, callback) {
 		for (var key in obj) {
 			if (obj.hasOwnProperty(key)) {
-				callback(key, obj[key]);
+				if (callback(key, obj[key])) {
+					break;
+				}
 			}
 		}
 	}
@@ -88,9 +90,7 @@ const summaryWrapperLambdaCode = `
 		return JSON.stringify(results, null, 4);
 	};
 
-	var oldTextSummary = function(data) {
-		// TODO: implement something like the current end of test summary
-	};
+	// TODO: bundle the text summary generation from jslib and get rid of oldCallback
 
 	return function(exportedSummaryCallback, jsonSummaryPath, data, oldCallback) {
 		var result = {};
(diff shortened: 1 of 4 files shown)


MStoykov (approved):
LGTM!

 I opted to not read the linked stuff ... for now :P

na--:
You are very wise... :sweat_smile: ugh, I'll have to remember how to fix the windows fs tests now...

js/summary.go
```diff
@@ -36,10 +36,12 @@ import (
...
 const summaryWrapperLambdaCode = `
 (function() {
-	var forEach = function(obj, callback) {
+	var forEach = function (obj, callback) {
 		for (var key in obj) {
 			if (obj.hasOwnProperty(key)) {
-				callback(key, obj[key]);
+				if (callback(key, obj[key])) {
+					break;
+				}
```
imiric, line 44: Why was this added? I don't see any of the callbacks returning `true`.

Would it make more sense to `catch` an error to trigger this `break` instead?

na--: It was added for the text summary generation code. It currently lives in https://gist.github.com/na--/eaf6540fae765499cbfddce1a052583f, but I'll clean it up and push it to jslib.k6.io soon, and it will be bundled in the k6 v0.31.0 cycle, so the `forEach` changes will soon be used. And anyway, returning a result seems cleaner than throwing an error - for one thing, I wouldn't want to catch other actual errors by mistake.

imiric (approved):

imiric (approved):

PR #1809, opened 2021-01-19
Release k6 v0.30.0

na--:
Reviewing the release notes probably is easiest by looking at them rendered: https://github.com/loadimpact/k6/blob/release-v0.30.0/release%20notes/v0.30.0.md

The current planned release date for v0.30.0 is tomorrow. Before that, I plan to make the jslib PR with the v0.0.1 JS summary functions, and hopefully also at least a WIP docs PR (help appreciated here)...

### DIFF
--- a/release notes/v0.30.0.md
+++ b/release notes/v0.30.0.md
@@ -1,0 +1,124 @@
+k6 v0.30.0 is here! :tada: It was a bit of a slow after-holiday release, but it still packs a few major new features and improvements that users have been requesting for a long time!
+
+## New features
+
+### Shared memory read-only arrays between VUs ([#1739](https://github.com/loadimpact/k6/pull/1739))
+
+k6 has long had an issue with the handling of big data files with test fixtures. For example, if you have a huge `users.json` file with test users for your application:
+```js
+[
+  {"username": "user1", "password": "password1", "other": "some-long-data-....-1"},
+  {"username": "user2", "password": "password2", "other": "some-long-data-....-2"},
+  // ... ~1 million more users or more... :D
+  {"username": "user999999", "password": "password999999", "other": "some-long-data-....-999999"}
+]
+```
+
+If you just use `JSON.parse(open('users.json'))` in your script, then every VU will have a copy of the whole huge data set. Every VU in k6 is a separate JavaScript runtime, so there wasn't a thread-safe way to share data between them. Until now, that is!
+
+We've added a new built-in `SharedArray` object in the new `k6/data` module that allows VUs to share read-only data:
+```js
+import { SharedArray } from 'k6/data';
+import { sleep } from 'k6';
+import http from 'k6/http';
+
+let users = new SharedArray('someName', function () {
+    // This function will be called only once, in the first init context
+    // execution. Every other VU will just get a memory-safe read-only reference
+    // to the already loaded data.
+    console.log('Loading users.json, this happens only once...');
+    // You are not restricted to JSON, you can do anything - parse a CSV or XML
+    // file, generate random data, etc. - as long as you return an array.
+    return JSON.parse(open('users.json'));
+});
+
+export let options = { vus: 10, duration: '30s' };
+export default function () {
+    let randomUserID = Math.floor(Math.random() * users.length);
+    let user = users[randomUserID]; // alternatively, we can also use __VU and/or __ITER
+    console.log(`VU ${__VU} is running iteration ${__ITER} with user ${user.username}...`);
+    http.post('https://httpbin.test.k6.io/post', JSON.stringify(user));
+    sleep(Math.random() * 2); // or, better yet, use arrival-rate
+}
+```
+
+Notice how `Loading users.json` is logged only once, but each VU uses the `users` variable like a normal JS array. The data is read only once and we have just a single copy of the huge array in memory! Behind the scenes, k6 uses a JS [`Proxy`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy) to transparently copy only the row each VU requests in `users[randomUserID]` to it. This on-demand copying is a bit inefficient, but it's still leagues better than having a copy of the huge array in every VU! And you can avoid the copying in every iteration by pinning the data used by every VU and having `let user = users[randomUserID]` in the [init context](https://k6.io/docs/using-k6/test-life-cycle)!
+
+And yes, you can have multiple `SharedArray` objects in the same script, just make sure to give them unique names - this is what `someName` in the script above was for. Because VUs are independent JS runtimes, we need some way to differentiate between the different shared memory objects, so we require them to have unique names. These names are also the IDs that any [xk6](https://github.com/k6io/xk6) extensions would need to use to access them.
+
+This works both locally and in the cloud. We advise everyone who deals with large data files to wrap them in a `SharedArray` and give this new feature a try. The required script changes should be minimal, while the memory usage should be significantly lower. Hopefully, we can finally consider one of the biggest blockers k6 users have had for a long time solved! :tada:
+
+### Support a `handleSummary()` callback at the end of the test ([#1768](https://github.com/loadimpact/k6/pull/1768))
+
+You can now `export` a function called `handleSummary()` and k6 will call it at the end of the test run, after even `teardown()`. `handleSummary()` will be called with a JS object containing the same information that is used to generate the end-of-test summary and `--summary-export`, and allows users to completely customize how the end-of-test summary looks like.
+
+Besides customizing the end-of-test CLI summary (if `handleSummary()` is exported, k6 will not print the default), you can also transform the summary data to various machine or human-readable formats and save it to files. This allows the creation of JS helper functions that generate JSON, CSV, XML (JUnit/xUnit/etc.), HTML, etc. files from the summary data. Even binary formats like PDF are not out of reach, potentially, with an appropriate JS library that works in k6! You can also send the generated reports to a remote server by making an HTTP request with them (or using any of the other protocols k6 already supports)! Here's a simple example:
+
+```js
+import http from 'k6/http';
+import k6exmaple from 'https://raw.githubusercontent.com/loadimpact/k6/master/samples/thresholds_readme_example.js';
+export default k6exmaple; // use some predefined example to generate some data
+export const options = { vus: 5, iterations: 10 };
+
+// These are still very much WIP and untested, but you can use them as is or write your own!
+import { jUnit, textSummary } from 'https://jslib.k6.io/k6-summary/0.0.1/index.js';
+
+export function handleSummary(data) {
+    console.log('Preparing the end-of-test summary...');
+
+    // Send the results to some remote server or trigger a hook
+    let resp = http.post('https://httpbin.test.k6.io/anything', JSON.stringify(data));
+    if (resp.status != 200) {
+        console.error('Could not send summary, got status ' + resp.status);
+    }
+
+    return {
+        'stdout': textSummary(data, { indent: ' ', enableColors: true}), // Show the text summary to stdout...
+        'junit.xml': jUnit(data), // but also transform it and save it as a JUnit XML...
+        'summary.json': JSON.stringify(data), // and a JSON with all the details...
+        // And any other JS transformation of the data you can think of,
+        // you can write your own JS helpers to transform the summary data however you like!
+    }
+}
+```
+
+k6 expects `handleSummary()` to return a `{key1: value1, key2: value2, ...}` map. The values can be string or `ArrayBuffer` and represent the generated summary report contents. The keys should be strings and determine where the contents will be displayed or saved: `stdout` for [standard output](https://en.wikipedia.org/wiki/Standard_streams), `stderr` for standard error, or a path to a file on the system (which will be overwritten).
+
+The format of the `data` parameter is similar but not identical to the data format of `--summary-export`. The format of `--summary-export` remains unchanged, for backwards compatibility, but the data format for this new k6 feature was made more extensible and had some of the ambiguities and issues from the previous format fixed. We can't cover the new format in the release notes, though you can easily see what it contains by using `return { 'stdout': JSON.stringify(data)};` in `handleSummary()`! :smile:
+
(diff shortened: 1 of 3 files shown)


release notes/v0.30.0.md
```diff
@@ -0,0 +1,123 @@
...
+```js
+[
+  {"username": "user1", "password": "password1", "other": "some-long-data-....-1"},
+  {"username": "user2", "password": "password2", "other": "some-long-data-....-2"},
+  // ... ~1 million more users or more... :D
+  {"username": "user999999", "password": "password999999", "other": "some-long-data-....-999999"}
+]
+```
+
+If you just use `JSON.parse(open('users.json'))` in your script, then every VU will have a copy of the whole huge data set. Every VU in k6 is a separate JavaScript runtime, so there wasn't a thread-safe way to share data between them. Until now, that is!
```
MStoykov, line 17: Technically there wasn't "a way" to share data ... even non-thread-safe-ly;P . So maybe just drop the `thread-safe` part?

release notes/v0.30.0.md
```diff
@@ -0,0 +1,123 @@
...
+We've added a new built-in `SharedArray` object in the new `k6/data` module that allows VUs to share read-only data:
+```js
+import { SharedArray } from 'k6/data';
+import { sleep } from 'k6';
+import http from 'k6/http';
+
+let users = new SharedArray('someName', function () {
+    // This function will be called only once, in the first init context
+    // execution. Every other VU will just get a memory-safe read-only reference
+    // to the already opened data.
```
MStoykov, line 28:
```suggestion
    // to the already returned data.
```

release notes/v0.30.0.md
```diff
@@ -0,0 +1,123 @@
...
+    console.log('Loading users.json, this happens only once...');
+    // You are not restricted to JSON, you can do anything - parse a CSV or XML
+    // file, generate random data, etc. - as long as you return an array.
+    return JSON.parse(open('users.json'));
+});
+
+export let options = { vus: 10, duration: '30s' };
+export default function () {
+    let randomUserID = Math.floor(Math.random() * users.length);
+    let user = users[randomUserID]; // alternativaly, we can also use __VU and/or __ITER
```
imiric, line 38:
```suggestion
    let user = users[randomUserID]; // alternatively, we can also use __VU and/or __ITER
```

release notes/v0.30.0.md
```diff
@@ -0,0 +1,123 @@
...
+    console.log(`VU ${__VU} is running iteration ${__ITER} with user ${user.username}...`);
+    http.post('https://httpbin.test.k6.io/post', JSON.stringify(user));
+    sleep(Math.random() * 2); // or, better yet, use arribal-rate
```
MStoykov, line 41:
```suggestion
    sleep(Math.random() * 2); // or, better yet, use arrival-rate
```
also ... maybe just use arrival rate ?

imiric, line 41:
```suggestion
    sleep(Math.random() * 2); // or, better yet, use arrival-rate
```

release notes/v0.30.0.md
```diff
@@ -0,0 +1,123 @@
...
+}
+```
+
+Notice how `Loading users.json` is logged only once, but each VU uses the `users` variable like a normal JS array. The data is read only once and we have just a single copy of the huge array in memory! Behind the scenes, k6 uses a JS [`Proxy`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy) to transparently copy only the row each VU requests in `users[randomUserID]` to it. This on-demand copying is a bit inefficient, but it's still leagues better than having a copy of the huge array in every VU! And you can avoid the copying in every iteration by pinning the data used by every VU and having `let user = users[randomUserID]` in the [init context](https://k6.io/docs/using-k6/test-life-cycle)!
```
MStoykov, line 45: Can we get an ascinema recording?

release notes/v0.30.0.md
```diff
@@ -0,0 +1,123 @@
...
+
+And yes, you can have multiple `SharedArray` objects in the same script, just make sure to give them unique names - this is what `someName` in the script above was for. Because VUs are independent JS runtimes, we need some way to differentiate between the different shared memory objects, so we require them to have unique names. These names are also the IDs that any [xk6](https://github.com/k6io/xk6) extensions would need to use to access them.
```
imiric, line 47: The last sentence is a bit confusing. Why mention xk6 here at all?

release notes/v0.30.0.md
```diff
@@ -0,0 +1,123 @@
...
+
+This works both locally and in the cloud. We advise everyone that deals with huge data files to wrap them in a `SharedArray` and give this new feature a try. The required script changes should be minimal, while the benefits should be significant. Hopefully, we can finally consider one of the biggest blockers k6 users have had for a long time solved! :tada:
```
imiric, line 49:
```suggestion
This works both locally and in the cloud. We advise everyone who deals with large data files to wrap them in a `SharedArray` and give this new feature a try. The required script changes should be minimal, while the benefits should be significant. Hopefully, we can finally consider one of the biggest blockers k6 users have had for a long time solved! :tada:
```

MStoykov, line 49: I vote for having the required changes (with a few examples) spelled out in the docs and that being linked from here. This will at least give us something to link to whenever people ask what "wrap them in a `SharedArray`" means.

release notes/v0.30.0.md
imiric, line 45: Should we link to that memory/CPU benchmark that @MStoykov did here? It would be good to give some numbers to give an idea of the performance improvement.

release notes/v0.30.0.md
```diff
@@ -0,0 +1,123 @@
...
+
+### Support a `handleSummary()` callback at the end of the test ([#1768](https://github.com/loadimpact/k6/pull/1768))
+
+You can now `export` a function called `handleSummary()` and k6 will call it at the end of the test run, after even `teardown()`. `handleSummary()` will be called with a JS object containing the same information that is used to generate the end-of-test summary and `--summary-export`, and allows users to completely customize how the end-of-test summary looks like.
+
+Besides customizing the end-of-test CLI summary (if `handleSummary()` is exported, k6 will not print the default), you can also generate and save as files various machine-readable data formats with the same data (JSON, CVS, JUnit/xUnit/etc. XML, etc.), as well as human-readable formats like HTML. Even binary formats like PDF are not out of reach, potentially, with an appropriate JS library that works in k6! You can also send the generated reports to a remote server by making an HTTP request with them (or using any of the other protocols k6 already supports)!
```
imiric, line 55:
```suggestion
Besides customizing the end-of-test CLI summary (if `handleSummary()` is exported, k6 will not print the default), you can also generate and save as files various machine-readable data formats with the same data (JSON, CSV, JUnit/xUnit/etc. XML, etc.), as well as human-readable formats like HTML. Even binary formats like PDF are not out of reach, potentially, with an appropriate JS library that works in k6! You can also send the generated reports to a remote server by making an HTTP request with them (or using any of the other protocols k6 already supports)!
```

MStoykov, line 55:
```suggestion
Besides customizing the end-of-test CLI summary (if `handleSummary()` is exported, k6 will not print the default), you can also generate and save to files various machine-readable data formats with the same data (JSON, CVS, JUnit/xUnit/etc. XML, etc.), as well as human-readable formats like HTML. Even binary formats like PDF are not out of reach, potentially, with an appropriate JS library that works in k6! You can also send the generated reports to a remote server by making an HTTP request with them (or using any of the other protocols k6 already supports)!
```
This just seems more natural to me ... 


But also the whole thing reads like there is built-in way to export to all of those. I think it should be reworded so that confusion is not possible. 

I think something like: "you can also generate and save data to files in both text and binary form. This does allow the creation of js helpers to generate <formats> that aren't provided by k6 natively but here is a list of community-made ones." (and the last one is obviously a link to the documentation with just the most basic of junit one that I've hacked together).

release notes/v0.30.0.md
```diff
@@ -0,0 +1,123 @@
...
+
+k6 expects `handleSummary()` to return a `{key1: value1, key2: value2, ...}` map. The values can be string or `ArrayBuffer` and represent the generated summary report contents. The keys should be strings and determine where the contents will be displayed or saved: `stdout` for [standard output](https://en.wikipedia.org/wiki/Standard_streams), `stderr` for standard error, or a path to a file on the system (which will be overwritten). Here's a simple example:
```
MStoykov, line 57: Maybe the above explanation of the possible uses and especially the listing of the formats should be after this example. Although then it will be harder for someone just skimming this to notice that they can now maybe generate their custom JSON report directly in k6. 

So maybe someone with better English than mine should try to restructure it even more :thinking:

imiric (changes requested):
Nice work @na--! 👏  Just spotted some minor things.

MStoykov (commented):
LGTM!. Mostly problems with some of the wording in the handleSummary section, and asking for some more links to relevant documentation that we(especially me) should've probably already started writing ... :(

release notes/v0.30.0.md
na--: hmm not sure it's a good idea to do that in the release notes

release notes/v0.30.0.md
na--: :+1: for that, we should include it in the docs PR after we release

sniku, line 49: I think it makes sense to be explicit here. I'm suggesting a change from `benefits` to `memory decrease` 

```suggestion
This works both locally and in the cloud. We advise everyone that deals with huge data files to wrap them in a `SharedArray` and give this new feature a try. The required script changes should be minimal, while the memory usage should be significantly lower. Hopefully, we can finally consider one of the biggest blockers k6 users have had for a long time solved! :tada:
```

release notes/v0.30.0.md
na--: There was, with the setup data, until we saw the data races and fixed it :smile: I'd prefer to leave thread-safe, to emphasize that a special solution was actually needed

release notes/v0.30.0.md
na--: not sure your suggestion makes it clearer, how about "already loaded data"?

release notes/v0.30.0.md
na--: that was my initial idea, but it makes the script longer and takes a bit of focus away from the actual feature

release notes/v0.30.0.md
na--: can you point me towards the benchmark? I can't seem to find it

release notes/v0.30.0.md
na--: To give some more exposure to xk6 and for SEO :rofl: But also because this feature can be useful to any xk6 plugin authors - both for accessing `SharedArrays`, but also for implementing something similar to it, see https://github.com/loadimpact/k6/blob/45fff81b0b3af8d174c960450d16f5c4431ef9a4/js/common/initenv.go#L82-L84

release notes/v0.30.0.md
na--: Resolved by https://github.com/loadimpact/k6/pull/1809/commits/78875f72a82841ecb179455f9c891c87526a5424

release notes/v0.30.0.md
na--: Resolved by https://github.com/loadimpact/k6/pull/1809/commits/78875f72a82841ecb179455f9c891c87526a5424

release notes/v0.30.0.md
na--: Resolved by https://github.com/loadimpact/k6/pull/1809/commits/78875f72a82841ecb179455f9c891c87526a5424

release notes/v0.30.0.md
na--: Resolved by https://github.com/loadimpact/k6/pull/1809/commits/78875f72a82841ecb179455f9c891c87526a5424

release notes/v0.30.0.md
na--: Resolved by https://github.com/loadimpact/k6/pull/1809/commits/9b61b1bce833dfbbd61f2606860ca7f9831f6c42?

release notes/v0.30.0.md
na--: resolved by https://github.com/loadimpact/k6/pull/1809/commits/78875f72a82841ecb179455f9c891c87526a5424?

release notes/v0.30.0.md
na--: hmm this paragraph could be moved after the example, but I'd prefer the previous paragraph to stay before

release notes/v0.30.0.md
na--: Resolved by https://github.com/loadimpact/k6/pull/1809/commits/78875f72a82841ecb179455f9c891c87526a5424

release notes/v0.30.0.md
na--: see if https://github.com/loadimpact/k6/pull/1809/commits/80ca8a30809819556f409fdf55e8b6a06fbbb306 is better?

release notes/v0.30.0.md
na--: resolved by https://github.com/loadimpact/k6/pull/1809/commits/78875f72a82841ecb179455f9c891c87526a5424?

release notes/v0.30.0.md
imiric: Found it in #1739: https://gist.github.com/MStoykov/ea62b0d16dc21261779f46db0fe8bd53

release notes/v0.30.0.md
imiric: Hhmm seems out of place to me, since we'd have to expand on what you mean, and that's too much detail for the release notes. But OK.

release notes/v0.30.0.md
na--: hmm a bit contrived and not showing much benefit - I'll see if I can repeat it with a bigger array later, but in any case, not sure if we can give any benchmark results that are representative, since every case would be different.

imiric: OK, I would've preferred to see even percentage ranges, but you're right that it will depend on the use case.

imiric (dismissed):
:shipit:

MStoykov (approved):

MStoykov (approved):

PR #1822, opened 2021-01-26
Disable some babel plugins that are now part of goja or are not useful

MStoykov:
Unfortunatelly babel doesn't let you disable plugins from a preset, so
we need to specify every plugin that we want with it corresponding
configuration.


js/compiler/compiler.go
```diff
@@ -37,7 +37,39 @@ import (
...
+			// all the other module plugins are just dropped
+			[]interface{}{"transform-es2015-modules-commonjs", map[string]interface{}{"loose": false}},
+			// "transform-regenerator", // Doesn't really work unless regeneratorRuntime is also added
+
+			// es2016 https://github.com/babel/babel/blob/v6.26.0/packages/babel-preset-es2016/src/index.js
+			"transform-exponentiation-operator",
+
+			// es2017 https://github.com/babel/babel/blob/v6.26.0/packages/babel-preset-es2017/src/index.js
+			// "syntax-trailing-function-commas", // in goja
+			// "transform-async-to-generator", // Doesn't really work unless regeneratorRuntime is also added
```
imiric, line 71: Should we consider including this along with [transform-runtime](https://babeljs.io/docs/en/babel-plugin-transform-runtime) then? It would add a lot of value to support generators and this poor-man's `async`, unless they would add a lot to the bundle and explode the memory requirements.

MStoykov: I think we discussed this internally and we came to the conclusion that it is better to break all the async/await stuff that is halfway working and letting people do it on their own through the [template-es6](https://github.com/k6io/template-es6) or similar.

I removed the generator tests specifically because we were failing practically all of them because it was needing `regeneratorRuntime`. And I am heavily against adding `regeneratorRuntime` and reenabling plugins, possibly bringing back the polyfill for Promise. It will likely not work great, again all of those tools are not written for k6 but for nodejs/browsers. Goja does implement Ecmascript faithfully, but there are still possibilities that those tools require something special in one of the cases to work. 

Testing this will likely show whether I am correct or not, but IMO "not async" async is not worth the time investment and explanation on why it is not working as someone would expect. And then the possibility of breaking changes once we actually start having real async.

imiric (dismissed):
I didn't really dig into each enabled plugin and their configuration, but as long as the TC39 tests pass I think we can be confident that we're not breaking anything.

I like that they're now explicitly listed instead of being part of some obscure preset. Nice work! :clap:

MStoykov:
> I didn't really dig into each enabled plugin and their configuration, 

The configuration is copied from the presets, which are linked above each section of plugins. So it should be exactly the same but without some of them.

> I like that they're now explicitly listed 

Yeah, this also means that in the future removing one will be just commenting a line and rerunning the tests. And let's hope we can remove all of them :champagne:

na-- (approved):

PR #1824, opened 2021-01-26
Drop corejs

MStoykov:
Everything but:
- Promises that aren't really working
- define/lookup getter/setter which are AnnexB
- other stuff that aren't standard
is now in goja, so there is no need for corejs

Things that will break:
- The workaround explained in this [comment](https://github.com/loadimpact/k6/issues/779#issuecomment-674311032) and on [stackoverflow](https://stackoverflow.com/a/65849645/5427244) although this breaks also because of #1822 
- [define/lookup getter/setter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/__defineGetter__) which are depricated. defineProperty is since ES5.1 at least, so it is very unlikely any code is using `__defineGetter__` they were also added to the Ecmascript later in AnnexB so :man_shrugging: 
- From the rest almost everything is AnnexB stuff completely unrelated to k6 such as `bold` to make a bold HTML tag.
- or completely unsupported stuff that was at some point a proposal

The one thing that now going over this [shim.js](https://github.com/loadimpact/k6/blob/affa94232383673bda7ca42fb32b172f13541769/js/lib/k6-shim.js) from the previous PR, which might give us trouble is that `flatten` was renamed to `flat` somewhere along the proposal process. But corejs had it with the old name as that happened after that version. I think we shouldn't do anything and see if this actually breaks for users as this is 1 line change and will align with what is in browsers for example where `flatten` isn't defined.


imiric (commented):
This is _YUGE_! :clap:

Did you run some benchmarks before/after? It would be good to know the difference and mention it in the release notes.

MStoykov:
> Did you run some benchmarks before/after

I didn't have time for that as well ...

imiric (approved):
I'm OK with this, just two questions:

- There are a lot of new errors in `breaking_test_errors.json`. Are we OK with them? I can't really tell whether they are important or not.

- Maybe for a separate PR, but could you document how `breaking_test_errors.json` is generated exactly? I followed the `js/tc39/README.md`, emptied it with `{}` and ran the tests, but no JSON was generated, though I didn't wait for it to finish. Do you parse the `out.log` and convert it to JSON? Exact commands would be helpful here.

MStoykov:
* all the new errors are `define/lookup-getter/setter`s, which we decided to dro, and some `Promise` ones as again the tc39 tests aren't the best-tagged tests ever :( 

* It is in the output. I run it with `-v` (not certain it's required :thinking: ) and write it to a file. Then I open it and search for `FAIL` - the JSON will be just before it. It takes 2-3 seconds to delete the file to only contain the JSON in vim and then I move it on the `breaking_test_errors.json`. 

Ideally, this will be automated (somewhat) one day,  but currently, if you don't clean the file it will not work properly and I don't want to spend an afternoon fixing that, given that we will likely stop touching this that much after a few more changes. If anything I am more interested in optimizing it to be kind of "faster" but that will take longer and given that it will naturally become faster, the less babel is needed, I am more for trying to

na-- (approved):
LGTM. Added a `breaking change` tag, since we probably should at least mention this as a minor breaking change, despite the fact nobody will probably notice it. Please update the PR description with what you think the note in the release notes should be. Ideally now, while everything is still fresh in your mind.

MStoykov:
@na-- I've updated the description

PR #1837, opened 2021-02-02
Disable some more tc39 tests

MStoykov:
This is mostly done for
1. speed
2. less tests we monitor, when the tests aren't important or relevant
   currently


na-- (commented):
Hmm instead of disabling the stable sorting tests, wouldn't it be very easy to add stable sorting in goja? Instead of `sort.Sort()` in [here](https://github.com/dop251/goja/blob/f5884268f0c0a0eb94423dcf7c2eda14907375e6/builtin_array.go#L355) use [`sort.Stable()`](https://golang.org/pkg/sort/#Stable)? I guess in most JS runtimes, the sorting is stable by default: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort#browser_compatibility

MStoykov:
Given that the stable sort tests take a solid minute together, I am kind of for just disabling them :P But if you want you can make a PR to goja ;P

na-- (commented):
If Go 1.16 fixes some of the unicode issues, why not leave the tests?

MStoykov:
it fixes *2* tests out of all of them ... and they test can you use a *specific* [unicode](https://github.com/tc39/test262/blob/main/test/language/identifiers/start-unicode-13.0.0.js) symbol (with [escape](https://github.com/tc39/test262/blob/main/test/language/identifiers/start-unicode-13.0.0-escaped.js)) at the beginning of an identifier. 

Also, it means that it will break with go1.15 but work with 1.16.

na-- (approved):

imiric (approved):
Seems safe :crossed_fingers:

Though really @MStoykov you don't need to keep optimizing this. The current 9m runtime is good enough.

PR #1839, opened 2021-02-03
Speeding up tc39 tests through using a pool of Babels

MStoykov:
PoC pool of babels so transformation is less bottlenecked on them

this speeds up tc39 tests

It also has the nice benefit of dropping the babels once they are no
longer needed if you just run k6.

But a different approach is probably going to be needed as the js package
(for example) with -race takes upwards of 3x as much time.


imiric (changes requested):
Is the performance hit for regular tests because the `babel` object is short-lived? According to the `sync.Pool` docs:

> On the other hand, a free list maintained as part of a short-lived object is not a suitable use for a Pool, since the overhead does not amortize well in that scenario. It is more efficient to have such objects implement their own free list.

In any case, I vote :-1: for this as is, since the main test suite taking ~15m will be painful every time, whereas the current ~11m runtime of the TC39 suite is perfectly fine, especially since it runs only on `js/*` changes.

js/compiler/compiler.go
```diff
@@ -79,28 +79,45 @@ var (
...
+	babel  *babel
 }
 
 // New returns a new Compiler
 func New(logger logrus.FieldLogger) *Compiler {
 	return &Compiler{logger: logger}
 }
 
+// InitializeBabel specifically initializes a separate (non-global) instance of babel for this compiler specifically
+// the returned error is only if babel couldn't be parsed/run which should never be possible
```
imiric, line 100:
```suggestion
// InitializeBabel initializes a separate (non-global) instance of babel specifically for this Compiler.
// An error is returned only if babel itself couldn't be parsed/run which should never be possible.
```

js/compiler/compiler.go
```diff
@@ -148,34 +165,42 @@ type babel struct {
...
-
-		this := vm.Get("Babel")
-		bObj := this.ToObject(vm)
-		globalBabel = &babel{vm: vm, this: this}
-		if err = vm.ExportTo(bObj.Get("transform"), &globalBabel.transform); err != nil {
+	onceBabelCode.Do(func() {
+		globalBabelCode, err = goja.Compile("<eval>", babelSrc, false)
+		if err != nil { // TODO cache this?
 			return
 		}
```
imiric, line 178: Why would we need to cache it? Actually, is the `err` check and return here needed at all?

js/compiler/compiler.go
```diff
@@ -148,34 +165,42 @@ type babel struct {
...
-		if _, err = vm.RunString(babelSrc); err != nil {
-			return
-		}
-
-		this := vm.Get("Babel")
-		bObj := this.ToObject(vm)
-		globalBabel = &babel{vm: vm, this: this}
-		if err = vm.ExportTo(bObj.Get("transform"), &globalBabel.transform); err != nil {
+	onceBabelCode.Do(func() {
+		globalBabelCode, err = goja.Compile("<eval>", babelSrc, false)
```
imiric, line 175: Why `<eval>` as `name`?

js/compiler/compiler.go
```diff
@@ -79,28 +79,45 @@ var (
...
 // Transform the given code into ES5
 func (c *Compiler) Transform(src, filename string) (code string, srcmap *SourceMap, err error) {
-	var b *babel
-	if b, err = newBabel(); err != nil {
-		return
+	if c.babel == nil {
+		onceBabel.Do(func() {
+			globalBabel, err = newBabel()
+		})
+		c.babel = globalBabel
```
imiric, line 115: It should never happen, but should we still check and return `err` here?

imiric (commented):
LGTM, just some minor things.

Could you include some before/after comparisons in the PR description?

js/compiler/compiler.go
MStoykov: hm ... I thought I did at least on the first ...uuh .. return.

js/compiler/compiler.go
MStoykov: just what it's used in other places ... we can call it "babel.js" I guess

js/compiler/compiler.go
MStoykov: in practice ... it should never be an error, I am almost fine with panicking as if babel doesn't compile ... well there is nothing we can do anymore ;)

js/compiler/compiler.go
MStoykov: I went with a really long descriptive name just in case

js/compiler/compiler.go
MStoykov: fixed it

MStoykov:
I will squash to 3 commits:
1. disable "async"
2. the speedup through pool
3. enable windows parallelism 

when we agree it looks okay to merge ;)

MStoykov:
I don't think I will be doing anything else for now apart from rewriting commits once I get some :+1: 

Here is a [list of all the newly skipped](https://gist.github.com/MStoykov/33ee684ae1fb0eb67432b5f510f35dca) tests with explanation on how it was generated (in the comment)

js/compiler/compiler.go
imiric: I'm not sure you answered my question :sweat_smile: If Babel doesn't compile then this would only fail once, so what's the point of caching it and of the `TODO`? And since we check this `err` a few lines below this, I don't see the point of this additional check and `return` from this lambda.

MStoykov: OKay .. the whole `if` is ... kind of not needed as this whole thing is inside `once.Do`.

But the idea behind the caching is that k6 will execute this `once.Do` .. once and only the first call will actually set the `err` for it's execution, all other executions will block on the `once.Do` until finished and will return a `nil` `globalBabelCode`

imiric: Ah, I see, so as a way to prevent blocking other goroutines? I would at least explain that in the `TODO`, as it's unclear why right now. Though given this is such an unlikely edge case I would just remove the check and `TODO`.

js/compiler/compiler.go
imiric: OK, I'm not really sure why it's important though :sweat_smile:

imiric (commented):
This LGTM, but I'm wondering if the `Compiler` pooling could've been done in the `tc39` package instead. I'm always wary of changes that are only useful for tests, especially in a core package like `compiler`.

Also nitpick, but I'd prefer deleting lines instead of commenting them out, unless they're temporary and you're sure we might need them back.

And I'd really appreciate a before/after run time comparison in the description :)

MStoykov:
I deleted the commented lines and did unexport one of the methods that doesn't need to be exported anymore, along rebasing stuff to three commits.

I did the compilerpool in tc39 but it will actually be useful in both some `js` test and `k6/http` where there are a bunch of es6 tests that are predominantly tc39 transpiling :(. But I would update them at a later point

Also .... literally all the current caching in that package even now is for testing as the normal k6 can only compile/transpile one file at a time and could've just initialized one babel of it's own ... but in tests it makes a big difference we don't initialize babel for each time we need it (the goja compiler is a lot less of a problem ;) )

js/compiler/compiler.go
```diff
@@ -204,3 +227,40 @@ func (b *babel) Transform(logger logrus.FieldLogger, src, filename string) (stri
 	}
 	return code, &srcMap, err
 }
+
+// Pool is a pool of compilers so it can be used easier in parallel tests as they have their own babel.
+type Pool struct {
```
na--, line 232: Hmm if we're not using this internally in k6, and we probably won't ever need it, why don't we implement it in `js/compiler/compiler_test.go`? TC39 tests (and `js/http` tests in the future) will still be able to import it and use it, but it won't "pollute" k6 release code until we actually need it.

MStoykov: It is literally 20 lines of code that doesn't add any dependency and having `js/http` depend on `tc39` seems ... like not so great idea ... 

But if you both want it so much I guess I am gonna move it :shrug:

na--: > having `js/http` depend on `tc39` seems ... like not so great idea ...

You didn't read my suggestion correctly - I don't want to move it in `tc39`, I want it to stay in the exact same package it currently is, just in a `_test.go` file. This way, other test packages will still be able to import it without any changes, but it won't be part of the normal k6 release code. And it's not that it will increase the binary size, I just don't want to have something in the codebase that is meant just for tests, that gets confusing very quickly, especially for newcomers.

MStoykov: > I want it to stay in the exact same package it currently is, just in a _test.go file. 

and 

> This way, other test packages will still be able to import it without any changes,

are contradicting AFAIK - you can *not* import things from `_test.go` files of one package in another `_test.go` files as ... the original package is still not compiled with the tests ;) . see https://stackoverflow.com/questions/31794141/can-i-create-shared-test-utilities

MStoykov: so even if I move it to tc39 ... it will still need to be in a *non* _test.go file which IMO will be even more confusing.

Also to reiterate the *whole* codebase around caching babel in this packages is about tests specifically as it makes no sense in the case of a running k6 where there will be only 1 compiler and compiling and transpiling source codes one after the other

na--: Ah, yeah... :face_with_head_bandage: I thought this wasn't a problem because I remember that the libraries were in `_test.go` file in the cloud integration tests, and while that's true, we don't have multiple modules there :man_facepalming: :sweat_smile:

na--: Fair enough, objections withdrawn :sweat_smile:

na-- (approved):

imiric (approved):
LGTM. Looking at the tests from other PRs, this speeds up the TC39 tests by 2 minutes (8 vs 6m), and the others don't seem to be impacted and have similar run times. So the description should at least be updated to remove "all other tests suffer".

Nice work! :clap:

MStoykov:
OH ... I keep forgetting this. 
The best time I had previously to this was 8m 30s and the droppage of some of the test got another 20s. But I also sometimes get 10minutes+. After all changes I am consistently under 4minutes and usually closer to 3minutes 10-20s

Issue #946, opened 2019-03-11
Refactor http.Response to have httpext.Response embeded 


MStoykov:
After https://github.com/loadimpact/k6/pull/907  and https://github.com/dop251/goja/commit/64be363d126e1a249e54a93cd47e475b6bb2e08b we can get `httpext.Response` to be an anonymous struct in `http.Response` as this will also give us back the ability to cache the whole json as a goja object

namit-chandwani:
Hey @MStoykov, I'd like to work on this issue. 
What would be a good starting point for resolving this?

MStoykov:
hi @namit-chandwani , kind of unfortunately I think this was fixed by https://github.com/grafana/k6/pull/1259 and https://github.com/grafana/k6/pull/1841 but wasn't closed :(.

PR #1841, opened 2021-02-04
More ArrayBuffer support

### LINKED ISSUE #1020: Better support for binary data

imiric:
These are the leftover changes to close #1020. Most of them break backwards compatibility, as discussed in the issue.

It might be easier to review by commit.

### DIFF
--- a/js/initcontext.go
+++ b/js/initcontext.go
@@ -207,8 +207,9 @@ func (i *InitContext) compileImport(src, filename string) (*goja.Program, error)
 	return pgm, err
 }
 
-// Open implements open() in the init context and will read and return the contents of a file.
-// If the second argument is "b" it returns the data as a binary array, otherwise as a string.
+// Open implements open() in the init context and will read and return the
+// contents of a file. If the second argument is "b" it returns an ArrayBuffer
+// instance, otherwise a string representation.
 func (i *InitContext) Open(ctx context.Context, filename string, args ...string) (goja.Value, error) {
 	if lib.GetState(ctx) != nil {
 		return nil, errors.New(openCantBeUsedOutsideInitContextMsg)
@@ -242,7 +243,8 @@ func (i *InitContext) Open(ctx context.Context, filename string, args ...string)
 	}
 
 	if len(args) > 0 && args[0] == "b" {
-		return i.runtime.ToValue(data), nil
+		ab := i.runtime.NewArrayBuffer(data)
+		return i.runtime.ToValue(&ab), nil
 	}
 	return i.runtime.ToValue(string(data)), nil
 }
--- a/js/modules/k6/http/response.go
+++ b/js/modules/k6/http/response.go
@@ -23,6 +23,7 @@ package http
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"net/url"
 	"strings"
@@ -66,7 +67,7 @@ func processResponse(ctx context.Context, resp *httpext.Response, respType httpe
 }
 
 func (h *HTTP) responseFromHttpext(resp *httpext.Response) *Response {
-	return &Response{Response: resp, h: h}
+	return &Response{Response: resp, h: h, cachedJSON: nil, validatedJSON: false}
 }
 
 // HTML returns the body as an html.Selection
@@ -116,7 +117,8 @@ func (res *Response) JSON(selector ...string) goja.Value {
 		}
 
 		if err := json.Unmarshal(body, &v); err != nil {
-			if syntaxError, ok := err.(*json.SyntaxError); ok {
+			var syntaxError *json.SyntaxError
+			if errors.As(err, &syntaxError) {
 				err = checkErrorInJSON(body, int(syntaxError.Offset), err)
 			}
 			common.Throw(rt, err)
--- a/js/modules/k6/http/response.go
+++ b/js/modules/k6/http/response.go
@@ -22,11 +22,13 @@ package http
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"net/url"
 	"strings"
 
 	"github.com/dop251/goja"
+	"github.com/tidwall/gjson"
 
 	"github.com/loadimpact/k6/js/common"
 	"github.com/loadimpact/k6/js/modules/k6/html"
@@ -37,6 +39,20 @@ import (
 type Response struct {
 	*httpext.Response `js:"-"`
 	h                 *HTTP
+
+	cachedJSON    interface{}
+	validatedJSON bool
+}
+
+type jsonError struct {
+	line      int
+	character int
+	err       error
+}
+
+func (j jsonError) Error() string {
+	errMessage := "cannot parse json due to an error at line"
+	return fmt.Sprintf("%s %d, character %d , error: %v", errMessage, j.line, j.character, j.err)
 }
 
 // processResponse stores the body as an ArrayBuffer if indicated by
@@ -51,18 +67,6 @@ func processResponse(ctx context.Context, resp *httpext.Response, respType httpe
 
 func (h *HTTP) responseFromHttpext(resp *httpext.Response) *Response {
 	return &Response{Response: resp, h: h}
-}
-
-// JSON parses the body of a response as json and returns it to the goja VM
-func (res *Response) JSON(selector ...string) goja.Value {
-	v, err := res.Response.JSON(selector...)
-	if err != nil {
-		common.Throw(common.GetRuntime(res.GetCtx()), err)
-	}
-	if v == nil {
-		return goja.Undefined()
-	}
-	return common.GetRuntime(res.GetCtx()).ToValue(v)
 }
 
 // HTML returns the body as an html.Selection
@@ -81,6 +85,69 @@ func (res *Response) HTML(selector ...string) html.Selection {
 		sel = sel.Find(selector[0])
 	}
 	return sel
+}
+
+// JSON parses the body of a response as JSON and returns it to the goja VM.
+func (res *Response) JSON(selector ...string) goja.Value {
+	rt := common.GetRuntime(res.GetCtx())
+	hasSelector := len(selector) > 0
+	if res.cachedJSON == nil || hasSelector { //nolint:nestif
+		var v interface{}
+
+		body, err := common.ToBytes(res.Body)
+		if err != nil {
+			common.Throw(rt, err)
+		}
+
+		if hasSelector {
+			if !res.validatedJSON {
+				if !gjson.ValidBytes(body) {
+					return goja.Undefined()
+				}
+				res.validatedJSON = true
+			}
+
+			result := gjson.GetBytes(body, selector[0])
+
+			if !result.Exists() {
+				return goja.Undefined()
+			}
+			return rt.ToValue(result.Value())
+		}
+
+		if err := json.Unmarshal(body, &v); err != nil {
+			if syntaxError, ok := err.(*json.SyntaxError); ok {
+				err = checkErrorInJSON(body, int(syntaxError.Offset), err)
+			}
+			common.Throw(rt, err)
+		}
+		res.validatedJSON = true
+		res.cachedJSON = v
+	}
+
+	return rt.ToValue(res.cachedJSON)
+}
+
+func checkErrorInJSON(input []byte, offset int, err error) error {
+	lf := '\n'
+	str := string(input)
+
+	// Humans tend to count from 1.
+	line := 1
+	character := 0
+
+	for i, b := range str {
+		if b == lf {
+			line++
+			character = 0
+		}
+		character++
+		if i == offset {
+			break
+		}
+	}
+
+	return jsonError{line: line, character: character, err: err}
 }
 
 // SubmitForm parses the body as an html looking for a from and then submitting it
--- a/js/modules/k6/http/response.go
+++ b/js/modules/k6/http/response.go
@@ -71,12 +71,12 @@ func (h *HTTP) responseFromHttpext(resp *httpext.Response) *Response {
 
 // HTML returns the body as an html.Selection
 func (res *Response) HTML(selector ...string) html.Selection {
-	body, err := common.ToBytes(res.Body)
+	body, err := common.ToString(res.Body)
 	if err != nil {
 		common.Throw(common.GetRuntime(res.GetCtx()), err)
 	}
 
-	sel, err := html.HTML{}.ParseHTML(res.GetCtx(), string(body))
+	sel, err := html.HTML{}.ParseHTML(res.GetCtx(), body)
 	if err != nil {
 		common.Throw(common.GetRuntime(res.GetCtx()), err)
 	}
--- a/js/modules/k6/http/response.go
+++ b/js/modules/k6/http/response.go
@@ -21,6 +21,7 @@
 package http
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"net/url"
@@ -37,6 +38,16 @@ import (
 type Response struct {
 	*httpext.Response `js:"-"`
 	h                 *HTTP
+}
+
+// processResponse stores the body as an ArrayBuffer if indicated by
+// respType. This is done here instead of in httpext.readResponseBody to avoid
+// a reverse dependency on js/common or goja.
+func processResponse(ctx context.Context, resp *httpext.Response, respType httpext.ResponseType) {
+	if respType == httpext.ResponseTypeBinary {
+		rt := common.GetRuntime(ctx)
+		resp.Body = rt.NewArrayBuffer(resp.Body.([]byte))
+	}
 }
 
 func (h *HTTP) responseFromHttpext(resp *httpext.Response) *Response {
--- a/js/modules/k6/http/response.go
+++ b/js/modules/k6/http/response.go
@@ -22,7 +22,6 @@ package http
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"net/url"
 	"strings"
@@ -68,17 +67,12 @@ func (res *Response) JSON(selector ...string) goja.Value {
 
 // HTML returns the body as an html.Selection
 func (res *Response) HTML(selector ...string) html.Selection {
-	var body string
-	switch b := res.Body.(type) {
-	case []byte:
-		body = string(b)
-	case string:
-		body = b
-	default:
-		common.Throw(common.GetRuntime(res.GetCtx()), errors.New("invalid response type"))
+	body, err := common.ToBytes(res.Body)
+	if err != nil {
+		common.Throw(common.GetRuntime(res.GetCtx()), err)
 	}
 
-	sel, err := html.HTML{}.ParseHTML(res.GetCtx(), body)
+	sel, err := html.HTML{}.ParseHTML(res.GetCtx(), string(body))
 	if err != nil {
 		common.Throw(common.GetRuntime(res.GetCtx()), err)
 	}
--- a/js/modules/k6/ws/ws.go
+++ b/js/modules/k6/ws/ws.go
@@ -330,28 +330,24 @@ func (s *Socket) handleEvent(event string, args ...goja.Value) {
 	}
 }
 
-// Send writes the given string or ArrayBuffer message to the connection.
-func (s *Socket) Send(message interface{}) {
-	rt := common.GetRuntime(s.ctx)
-
-	var (
-		msgType int
-		msg     []byte
-	)
-	switch m := message.(type) {
-	case string:
-		msgType = websocket.TextMessage
-		msg = []byte(m)
-	case goja.ArrayBuffer:
-		msgType = websocket.BinaryMessage
-		msg = m.Bytes()
-	default:
-		err := fmt.Errorf("unsupported message type: %T, expected string or ArrayBuffer", message)
-		common.Throw(common.GetRuntime(s.ctx), err)
+// Send writes the given string message to the connection.
+func (s *Socket) Send(message string) {
+	if err := s.conn.WriteMessage(websocket.TextMessage, []byte(message)); err != nil {
+		s.handleEvent("error", common.GetRuntime(s.ctx).ToValue(err))
 	}
 
-	if err := s.conn.WriteMessage(msgType, msg); err != nil {
-		s.handleEvent("error", rt.ToValue(err))
+	stats.PushIfNotDone(s.ctx, s.samplesOutput, stats.Sample{
+		Metric: metrics.WSMessagesSent,
+		Time:   time.Now(),
+		Tags:   s.sampleTags,
+		Value:  1,
+	})
+}
+
+// SendBinary writes the given ArrayBuffer message to the connection.
+func (s *Socket) SendBinary(message *goja.ArrayBuffer) {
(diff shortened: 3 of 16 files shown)


lib/netext/httpext/response.go
```diff
@@ -136,6 +137,8 @@ func (res *Response) JSON(selector ...string) (interface{}, error) {
 			body = b
 		case string:
 			body = []byte(b)
+		case goja.ArrayBuffer:
+			body = b.Bytes()
```
MStoykov, line 141: I am definitely against `httpext` depending on `goja` ;). I also don't think that supporting neither `[]byte` or `goja.ArrayBuffer` makes a lot of sense here, so I am for just dropping both.

MStoykov (changes requested):
Seems good to me - I will need to try using it probably. 

Can you drop the support for `[]byte` everywhere as well so that we can see if something breaks when we no longer support it.

lib/netext/httpext/response.go
imiric: True, I didn't want to refactor this to remove the dependency, and also think it doesn't make sense to support this use case.

That said, dropping `[]byte` would be yet another breaking change (we have a test for this use case: `TestNoResponseBodyMangling`), and practically I'm not sure convoluting the code just to remove the dependency is worth it.

@na-- Do you agree with dropping both `[]byte` and `goja.ArrayBuffer` from here?

na--: Hmm I completely agree with @MStoykov that we shouldn't depend on goja in httpext, but I am not sure we can drop `[]byte` from here - what would happen if the user has `responseType: binary` and tries to use `.json()`? :confused: 

Rather, maybe we should just move the whole .JSON method in there? https://github.com/loadimpact/k6/blob/a379c3787372ce2a4c70cb2049fc43be7af798a2/js/modules/k6/http/response.go#L47

MStoykov: >  `responseType: binary` and tries to use `.json()` 

Why would they do that? 

I would argue that this just doesn't seem like a case that we need to support or something that anyone would've done. So between, making this PR more complicated and moving stuff, and just .. dropping a switch, I much prefer the second one, given that I don't think it has ever been exercised

na--: > Why would they do that?

Why not? It will actually be more performant since goja won't do the utf8->utf16 conversion... And why would wanting the response type to be binary preclude you from accessing it as a JSON? 

I am fine with making minor breaking changes that are useful for the vast majority of users, but as far as I can see, the reason for this proposed breaking change is "it will be slightly easier to keep the k6 code a bit cleaner"... Not good enough of a reason in my book, sorry...

If you dislike the move of the `.json()` method back to the JS struct, here's an alternative that has some other performance benefits. Always use `string` or `[]byte` in `httpext.Response.Body`, and use a `Proxy` in the JS `Response` wrapper for accessing it. That proxy will return a `goja.ArrayBuffer` if the response type is `binary`, but it will also lazily return string if the response type is text. This will prevent a bunch of unnecessary utf8->utf16 conversions, right?

MStoykov: > Why not? It will actually be more performant since goja won't do the utf8->utf16 conversion... And why would wanting the response type to be binary preclude you from accessing it as a JSON?

I don't think a single user would know that and would have done it. In the same way that I don't think that if a user will use `.json()` they would've before that set `responseType: binary`. So for me, this seems like a "breaking change" that will not be hit by anyone. 

Given that and the fact that the benefit from this being supported is based on users doing stuff and will require some amount of additional code that isn't really needed, I am for dropping this now and either having a PR to reimplement it later, maybe as part of #946. 

Or better (IMO) have the `response.body` field be a [property accessor](https://pkg.go.dev/github.com/dop251/goja#Object.DefineAccessorProperty) that only transforms to `string` or `ArrayBuffer` when required and if that isn't called we can optimize to not actually do any transformation and just work on `[]byte` (or however we save it). 

But my argument is that this is very unlikely to have been used and this PR has enough things that can break without any amount of additional complicated or not complicated refactoring for features (I  argue) that aren't used.

imiric: I agree with Mihail that it's unlikely someone would deliberately use `responseType: 'binary'` and then called `res.json()`, but they might've done it inadvertently so dropping support for it now would likely break some scripts.

That said, since Mihail suggested we drop `[]byte` from everywhere where `ArrayBuffer` is used now, my vote is to drop it from here as well, and test this thoroughly with existing scripts (canary) and consider bringing it back with either the Proxy or property accessor approach if needed.

na--: I still don't see any benefits for users mentioned as a reason for the breaking change, and I see easy workaround that would prevent the breaking change, with performance side-benefits on top, so my position is still that we shouldn't make a breaking change... :slightly_smiling_face: 

> Or better (IMO) have the `response.body` field be a [property accessor](https://pkg.go.dev/github.com/dop251/goja#Object.DefineAccessorProperty) that only transforms to `string` or `ArrayBuffer` when required and if that isn't called we can optimize to not actually do any transformation and just work on `[]byte` (or however we save it).

This seems awesome and _much_ easier to do compared to a `Proxy`, so what's the drawback of just doing that?

imiric: The breaking change wouldn't have benefits to users, of course. The reason is more to remove a feature that didn't make sense before and to avoid any complicated refactors.

But I also like the property accessor approach and am currently exploring it, so will push a change soon-ish.

imiric: So I added the dynamic `body` property via the `DefineAccessorProperty` method but it presents some challenges which I don't think are worth overcoming.

The way to set it up involves embedding `goja.Object` in the `httpext.Response` wrapper:

```go
type Response struct {
	*httpext.Response `js:"-"`
	*goja.Object      `js:"-"`
}
```

The problem with this is that it essentially overrides all `httpext.Response` fields, so properties like `status`, `cookies`, etc. return `undefined`. To fix it we would have to use `Proxy` or `DynamicObject` instead, which would involve some reflection for things like `Keys()` or `OwnKeys()`.

We would also need to add caching for the dynamic `body` property since there's no point in re-evaluating it given that the `responseType` remains static.

Maybe I'm not implementing this properly and we wouldn't need to jump through all these hoops, so please let me know if there's a simpler way. But if this is the only way then I'd have to push back as the added complexity of goja-fying `Response` doesn't seem worth it.

MStoykov: an "easier" way (that I haven't tested) will be to 
```
val := rt.ToValue(responseObject)
obj := val.ToObject(rt) // which should return a goja object
obj.DefineAccessorProperty ... // do it here, I expect you will not be able to overwrite names 
```

If that doesn't work I guess you can make an Object and set its prototype to the Response object ... which though is probably wrong in some other aspect. 

FYI source for the [objectGoReflect](https://github.com/dop251/goja/blob/ed7b3dc62eb92d6cf9e1782b3a2595e21c851819/object_goreflect.go#L79)

imiric: Thanks, but the first approach of defining the accessor on `Response` just doesn't call the getter so it doesn't work.

The second one with the prototype does work, and fields like `status` and `cookies` are accessible, but it has other issues. See https://github.com/imiric/k6/commit/017270f8c501ec940abceff3267392c28ad1528a

For one, we'd need to deal with `*goja.Object` internally instead of `*Response`, which is a PITA.

But more importantly, it doesn't passthrough the methods on `Response`, so `.json()` or `.html()` don't work. Calling `Object.keys()` also returns just `["body"]`.

So it seems like `Proxy` or `DynamicObject` would be the way to go, but I'm not sure we should.

Also `DefineDataProperty` seems like a better fit in this case since it avoids any caching requirements.

MStoykov: at 

https://github.com/imiric/k6/blob/017270f8c501ec940abceff3267392c28ad1528a/js/modules/k6/http/response.go#L56-L58

You need to provide a pointer to the http.Response as the JSON method is defined on the pointer receiver ;)

imiric: Thanks, that takes care of it then, but I'm still not a fan of `*goja.Object` being passed around everywhere. :-/

Anyway, I'll clean this up and push it here.

MStoykov: I did try to prevent us from having to refactor all of this for a "use-case" that doesn't exist, but :man_shrugging:

imiric: OK, see the changes for this in 412cbd10, but frankly, it's awful...

- Turns out we can't use `DefineDataProperty` since response bodies for `http.batch()` requests are lazily evaluated, so I had to switch back to `DefineAccessorProperty`.

- I considered adding a cache for `body`, but that would mean doubling the memory usage (it would be stored both as `[]byte` on `httpext.Response.Body` and as `string` or `ArrayBuffer` in the cache). I guess without it we technically still get the memory penalty, and also the CPU overhead of needlessly calling `getBody()` everytime it's accessed. Or... we could just go back to storing it once in `httpext.Response.Body` as `interface{}`.

- `Object.keys(resp)` still returns only `["body"]`. For this to work I think we'd need to switch to `Proxy` or `DynamicObject`, which has its own challenges as I commented above.

- `*goja.Object` is passed around internally instead of `*Response`.

The only things I like about this approach is that it avoids the repeated type assertions of `httpext.Response.Body` we had everywhere, and the silly `processResponse` workaround.

But let me know if anything here is an improvement or could be salvaged. I'm still leaning towards the original suggestion by Mihail to drop both `[]byte` and `ArrayBuffer` support from `res.json()` (and `res.html()`).

MStoykov: > Object.keys(resp) still returns only ["body"]

This actually breaks a much more important (IMO) functionality of being able to do:
```javascript
console.log(JSON.stringify(resp, null, "  "));
````
 and getting good debugging information of what all the response fields are ... 

I am of the opinion that one day ... we will need to rewrite this to actually use `DynamicObject`or a better (IMO) Object that we add Define*Property on them, depending on how it turns out) so we can stop doing needless transformations (like this one(It will probably need to have some nonenumerable Symbol property that is the original object so we can get it back).  But that will (as you have noticed) require quite a lot of reworking and likely changes throughout the codebase. Which is one of the reasons I haven't even tried to fix the triple copying of the data that is being uploaded (another problem that will be helped by something like that).

So I am also voting for just dropping the again ... almost certainly totally unused functionality of calling `json()`/`html()` on responses that are binary.

**Comment by na--** [comment]
Set the milestone to v0.32.0. Also, a reminder that we discussed websocket binary support, which might be a part of this PR or another similar one.

**Comment by imiric** [comment]
OK, so I reverted the WIP changes discussed here, and pushed the agreed breaking changes for v0.32.0.

Some notes:
- I moved `Response.JSON()` to the JS `k6/http` module, where it arguably should've been to begin with, so also supporting `ArrayBuffer` bodies does not introduce a dependency on `httpext` anymore, and we can use `common.ToBytes()` to convert all supported types. So there are no breaking changes for `Response.HTML()` or `Response.JSON()` anymore.
- `b64decode()` now returns `ArrayBuffer` by default, and the previous default `string` is returned when specifying an additional `"s"` (format) argument.
- `ArrayBuffer` support in `k6/ws` is still pending, and I intend to push it in this PR as well, unless you think it's large enough already, let me know.

File: lib/netext/httpext/response.go (hunk 1 above)
Comment (imiric) [reply lib/netext/httpext/response.go]: Resolving this as `httpext` no longer depends on `goja` (see 6a9ea790).

File: js/initcontext.go [hunk 2]
```diff
@@ -239,7 +240,8 @@ func (i *InitContext) Open(ctx context.Context, filename string, args ...string)
 	}
 
 	if len(args) > 0 && args[0] == "b" {
-		return i.runtime.ToValue(data), nil
+		b := i.runtime.NewArrayBuffer(data)
+		return &b, nil
```
MStoykov, line 244: Reading https://pkg.go.dev/github.com/dop251/goja#ArrayBuffer I would argue you should be calling ToValue on the value and just returning `goja.Value` instead of `interface{}` above. 

This should be exactly what happens either way but IMO it is better for it to be explicit.

js/modules/k6/http/response.go
```diff
@@ -68,17 +67,12 @@ func (res *Response) JSON(selector ...string) goja.Value {
...
-		body = b
-	default:
-		common.Throw(common.GetRuntime(res.GetCtx()), errors.New("invalid response type"))
+	body, err := common.ToBytes(res.Body)
+	if err != nil {
+		common.Throw(common.GetRuntime(res.GetCtx()), err)
 	}
 
-	sel, err := html.HTML{}.ParseHTML(res.GetCtx(), body)
+	sel, err := html.HTML{}.ParseHTML(res.GetCtx(), string(body))
```
MStoykov, line 79: This seems to me that will go string>[]byte-> string in the more common case IMO.

Maybe `ToString` ?

imiric: Nice catch ;) I didn't think it was important to optimize, but sure, we could add a `ToString` function.

js/initcontext.go
imiric: Resolved by 46fffe7e.

js/modules/k6/http/response.go
imiric: Resolved by e84daccb.

js/modules/k6/ws/ws.go
```diff
@@ -292,7 +292,8 @@ func (*WS) Connect(ctx context.Context, url string, args ...goja.Value) (*WSHTTP
 				Tags:   socket.sampleTags,
 				Value:  1,
 			})
-			socket.handleEvent("message", rt.ToValue(string(readData)))
+			ab := rt.NewArrayBuffer(readData)
+			socket.handleEvent("message", rt.ToValue(string(readData)), rt.ToValue(&ab))
```
imiric, line 296: Would this be prohibitively expensive memory wise?

I experimented with overriding the first argument to be a `String` object with an additional `buffer` property, but that doesn't expose the string primitive properties (`.length`, etc.) so it would be a breaking change.

Passing it as a second argument is backwards compatible, and I'm expecting it to be garbage collected if it's not used, but haven't run any benchmarks. One optimization we could do is to only pass it if the `message` handler function expects a second argument, but it would be awkward to do and would require additional refactoring.

MStoykov: I think it will be better to just have `binaryMessage` event or something like that instead of ... this ;) 

This is not how it's done in other tools but given the current way it is done in k6 I think this is better and then we can fix it properly in the next implementation when it is more async

js/modules/k6/ws/ws.go
```diff
@@ -334,13 +335,27 @@ func (s *Socket) handleEvent(event string, args ...goja.Value) {
...
+		msg = []byte(m)
+	case goja.ArrayBuffer:
+		msgType = websocket.BinaryMessage
+		msg = m.Bytes()
+	default:
+		err := fmt.Errorf("unsupported message type: %T, expected string or ArrayBuffer", message)
+		common.Throw(common.GetRuntime(s.ctx), err)
+	}
+
+	if err := s.conn.WriteMessage(msgType, msg); err != nil {
```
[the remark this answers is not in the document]
imiric: I disagree that it's cleaner, but here you go: b80813c3.

Expanding the API to now include another method that does the same thing for a different data type is hardly a good idea. It might've worked fine for that user and it's an isolated solution, but it's another method we'll have to eventually deprecate.

@na-- WDYT?

js/modules/k6/ws/ws.go
imiric: Sure, see e44fb6d2.

But I also think this is a rough bandaid that adds more work for users once we move to a better implementation and deprecate it.

js/modules/k6/ws/ws.go
```diff
@@ -293,6 +293,10 @@ func (*WS) Connect(ctx context.Context, url string, args ...goja.Value) (*WSHTTP
 				Value:  1,
 			})
 			socket.handleEvent("message", rt.ToValue(string(readData)))
+			if _, ok := socket.eventHandlers["binaryMessage"]; ok {
+				ab := rt.NewArrayBuffer(readData)
+				socket.handleEvent("binaryMessage", rt.ToValue(&ab))
+			}
```
MStoykov, line 299: No matter whether it is 1 or 2 events I do think that we need to check the messagetype and call either the correct method or if we choose to have only `message` events change the type of the value to ArrayBuffer or string based on it.

This will require more work in `readPump` which I think should probably be better handled in a separate PR , possibly even in a following release so ... I am mostly for dropping the `ws` changes for now 

WDYT @na-- @imiric

imiric: You're right, we should only call the `binaryMessage` handler if the `messageType == BinaryMessage`, I overlooked that.

It shouldn't be much work to refactor, I'll look into it today. Don't think we should delay this because of that change.

imiric: @MStoykov Let me know if 5e29bdfb is what you had in mind. I'll fix the new linter errors, and probably rebase on `master`.

js/modules/k6/ws/ws.go
```diff
@@ -360,9 +360,15 @@ func (s *Socket) Send(message string) {
...
-		s.handleEvent("error", common.GetRuntime(s.ctx).ToValue(err))
+func (s *Socket) SendBinary(message goja.Value) {
+	msg := message.Export()
+	if ab, ok := msg.(goja.ArrayBuffer); ok {
+		if err := s.conn.WriteMessage(websocket.BinaryMessage, ab.Bytes()); err != nil {
+			s.handleEvent("error", common.GetRuntime(s.ctx).ToValue(err))
+		}
+	} else {
+		common.Throw(common.GetRuntime(s.ctx),
+			fmt.Errorf("expected ArrayBuffer as argument, received: %T", msg))
```
MStoykov, line 371: Ideally, this should be the type of `message` but that will require a lot more code and practically can have a case where we can get one (if it's nil/undefined :thinking: )

I did try it and 
```golang
message.ToObject(common.GetRuntime(s.ctx)).ClassName()
```
seems to work okay and blows up with TypeError if `message` is null/undefined which is more or less what we want so ... maybe we can just add it ? But I would prefer if more testing is done that it won't panic the whole VU in some strange case

MStoykov (approved):

js/modules/k6/ws/ws.go
imiric: I'm not sure if we should bother with that. Sure, it would be good to log the JS type instead, but not everything can be converted to an object. For example, `socket.sendBinary(1)` throws `expected ArrayBuffer as argument, received: string` for some reason, which is more confusing than `expected ArrayBuffer as argument, received: int64`. And we'd have to handle the `null`/`undefined` case as well.

I don't see any way in the Go API to call `typeof` so we'd have to `RunString()` it. Do you think all this is worth it?

imiric: OK, should be resolved by 339af11c. I think I was exhaustive in the error tests, but let me know if I missed anything.

Also I rebased on `master` and cleaned up the fixup commit, but didn't want to squash the changes for your suggestions yet, in case @na-- disagrees with it.

MStoykov (approved):
LGTM :+1: 

Again I am fine with us having 1 `Send` and 1 event for binary and text messages I just think given the current API this will be confusing and unexpected for current scripts. When we rewrite the API with an event loop we can make it better.

My main nitpick is your usage of intermediate variables in places where it is not needed :rofl: , but I don't think it's so bad and didn't want to leave the same comment in 10 places

imiric:
> My main nitpick is your usage of intermediate variables in places where it is not needed

Please point them out :smile: I only spotted `errMessage`, which in my defense I copied as is from `httpext`, but I think other usages are when they were reused.

js/initcontext.go
```diff
@@ -239,7 +240,8 @@ func (i *InitContext) Open(ctx context.Context, filename string, args ...string)
 	}
 
 	if len(args) > 0 && args[0] == "b" {
-		return i.runtime.ToValue(data), nil
+		ab := i.runtime.NewArrayBuffer(data)
```
MStoykov, line 243: here @imiric , there is no point in having `ab` ;) 

This is done in another 5 places probably

imiric: Hhmm how would I get the pointer to it otherwise? Parenthesis wouldn't work in this case.

Though I did also spot `msg := message.Export()` which is not needed.

MStoykov: I also forget that you can't directly get the pointer from the function call :facepalm: 

On the other hand it seems like ... you don't need to? I mean do you really need to give a pointer?

imiric: I err'd on the assumption that it's better to pass around pointers to this data than incur copying penalties, but considering the [internal implementation is a pointer](https://github.com/dop251/goja/blob/1b11a6af686d7af5badfdd6588364537db72df09/typedarrays.go#L37) I guess it would be safe to return the struct itself.

Do you think this is worthwhile changing?

MStoykov: as my original comment suggests, I don't think it is a requirement.

In general, this means that it will now need to dereference 2 pointers in order to get to the actual data which is objectively worse, but not enough to matter in the grand scheme of things

na-- (approved):
LGTM

imiric:
Force-pushed because of `master` rebase.

na-- (approved):

PR #1842, opened 2021-02-05
Drop shorthand property plugin from babel

MStoykov:
As explained in this [comment](https://github.com/dop251/goja/pull/252#issuecomment-776582858) the breaking tests are about:

1. AnnexB handling of `__proto__` property where the "name" of `__proto__` should not be `__proto__`
2. Some keywords should not be possible property names -  unlikely to be important
3. escaping `import` in `import.meta` should not be allowed - unlikely to be important
4. `arguments` and `caller` should not be available to methods define with the `shorthand` syntax


js/tc39/breaking_test_errors.json
```diff
@@ -2145,6 +2139,7 @@
...
   "test/language/expressions/import.meta/same-object-returned.js-strict:true": "test/language/expressions/import.meta/same-object-returned.js: SyntaxError: test/language/expressions/import.meta/same-object-returned.js: Unexpected token (28:8)\n  26 | ---*/\n  27 | \n\u003e 28 | var a = import.meta;\n     |         ^\n  29 | var b = function() { return import.meta; }();\n  30 | \n  31 | assert.sameValue(import.meta, a, at \u003ceval\u003e:2:28542(114)",
+  "test/language/expressions/import.meta/syntax/escape-sequence-import.js-strict:true": "test/language/expressions/import.meta/syntax/escape-sequence-import.js: error is not an object (Test262: This statement should not be evaluated.)",
```
MStoykov, line 2142: testing whether `import` in `import.meta` can be escaped and work - it shouldn't , but it does - I don't see that as a ... problem :)

js/tc39/breaking_test_errors.json
```diff
@@ -2217,7 +2212,8 @@
...
   "test/language/expressions/object/__proto__-duplicate.js-strict:true": "test/language/expressions/object/__proto__-duplicate.js: error is not an object (Test262: This statement should not be evaluated.)",
-  "test/language/expressions/object/__proto__-permitted-dup-shorthand.js-strict:true": "test/language/expressions/object/__proto__-permitted-dup-shorthand.js: SyntaxError: test/language/expressions/object/__proto__-permitted-dup-shorthand.js: Redefinition of __proto__ property (22:2)\n  20 | var obj = {\n  21 |   __proto__,\n\u003e 22 |   __proto__,\n     |   ^\n  23 | };\n  24 | \n  25 | assert(obj.hasOwnProperty(\"__proto__\")); at \u003ceval\u003e:2:28542(114)",
+  "test/language/expressions/object/__proto__-fn-name.js-strict:true": "test/language/expressions/object/__proto__-fn-name.js: Test262Error: Expected true but got false at harness/sta.js:22:9(49)",
+  "test/language/expressions/object/__proto__-permitted-dup-shorthand.js-strict:true": "test/language/expressions/object/__proto__-permitted-dup-shorthand.js: TypeError: Object prototype may only be an Object or null: 2 at test/language/expressions/object/__proto__-permitted-dup-shorthand.js:21:12(9)",
```
MStoykov, line 2216: AnnexB handling of `__proto__` property where the name of the property is wrong (it is now "__proto__" but it shouldn't) - of no relevance IMO 

also a lot of the other AnnexB tests about `__proto__` are failing, so support for this seems ... not great

js/tc39/breaking_test_errors.json
```diff
@@ -2284,19 +2277,19 @@
...
   "test/language/expressions/object/method-definition/fn-name-fn.js-strict:true": "test/language/expressions/object/method-definition/fn-name-fn.js: TypeError: Cannot read property 'enumerable' of undefined at verifyNotEnumerable (harness/propertyHelper.js:212:11(8))",
+  "test/language/expressions/object/method-definition/forbidden-ext/b1/meth-forbidden-ext-direct-access-prop-arguments.js-strict:false": "test/language/expressions/object/method-definition/forbidden-ext/b1/meth-forbidden-ext-direct-access-prop-arguments.js: Test262Error: Expected a TypeError to be thrown but no exception was thrown at all at harness/sta.js:22:9(49)",
+  "test/language/expressions/object/method-definition/forbidden-ext/b1/meth-forbidden-ext-direct-access-prop-caller.js-strict:false": "test/language/expressions/object/method-definition/forbidden-ext/b1/meth-forbidden-ext-direct-access-prop-caller.js: Test262Error: Expected a TypeError to be thrown but no exception was thrown at all at harness/sta.js:22:9(49)",
```
MStoykov, line 2281: methods should not have access to `arguments` and `caller` - seems insignificant for k6.

js/tc39/breaking_test_errors.json
```diff
@@ -2463,6 +2453,7 @@
   "test/language/import/escaped-as-import-specifier.js-strict:true": "test/language/import/escaped-as-import-specifier.js: unexpected error type (ReferenceError), expected (SyntaxError)",
   "test/language/import/escaped-as-namespace-import.js-strict:true": "test/language/import/escaped-as-namespace-import.js: unexpected error type (ReferenceError), expected (SyntaxError)",
   "test/language/import/escaped-from.js-strict:true": "test/language/import/escaped-from.js: unexpected error type (ReferenceError), expected (SyntaxError)",
+  "test/language/keywords/ident-ref-this.js-strict:true": "test/language/keywords/ident-ref-this.js: error is not an object (Test262: This statement should not be evaluated.)",
```
MStoykov, line 2456: `this` should not be usable as an identifier, this is the case for some other keywords as well

js/tc39/breaking_test_errors.json
```diff
@@ -2697,6 +2688,9 @@
...
   "test/language/module-code/privatename-valid-no-earlyerr.js-strict:true": "test/language/module-code/privatename-valid-no-earlyerr.js: SyntaxError: test/language/module-code/privatename-valid-no-earlyerr.js: Unexpected character '#' (33:2)\n  31 | \n  32 | class outer {\n\u003e 33 |   #x = 42;\n     |   ^\n  34 | \n  35 |   f() {\n  36 |     var self = this; at \u003ceval\u003e:2:28542(114)",
   "test/language/reserved-words/await-module.js-strict:true": "test/language/reserved-words/await-module.js: error is not an object (Test262: This statement should not be evaluated.)",
+  "test/language/reserved-words/ident-reference-false.js-strict:true": "test/language/reserved-words/ident-reference-false.js: error is not an object (Test262: This statement should not be evaluated.)",
+  "test/language/reserved-words/ident-reference-null.js-strict:true": "test/language/reserved-words/ident-reference-null.js: error is not an object (Test262: This statement should not be evaluated.)",
+  "test/language/reserved-words/ident-reference-true.js-strict:true": "test/language/reserved-words/ident-reference-true.js: error is not an object (Test262: This statement should not be evaluated.)",
```
MStoykov, line 2693: `falce`, `null` and `true` should not be usable as identifiers

js/tc39/breaking_test_errors.json
na--: Hmm I don't understand this - [`arguments`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/arguments) is not not recommended, but it's not deprecated yet either, and we don't support ["rest parameters"](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/rest_parameters) yet, right? What am I missing?

js/tc39/breaking_test_errors.json
na--: Is this a goja bug then? And same question for `this`?

na-- (approved):
LGTM, though I don't understand some things I've added as inline comments

js/tc39/breaking_test_errors.json
MStoykov: The [tests](https://github.com/tc39/test262/blob/main/test/language/expressions/object/method-definition/forbidden-ext/b1/meth-forbidden-ext-direct-access-prop-caller.js) test that *methods* do *not* have `arguments` or `caller`.  So in this case they now do ... as goja does not have methods it has only functions ;)

js/tc39/breaking_test_errors.json
MStoykov: seems so, some checking is not done anymore in order to allow ... some other keywords to be used as identifiers in some cases namely when they are escaped as properties of objects, see https://github.com/dop251/goja/commit/13aee8e3cc3dad5437c2dea65c534d4bf2afb11e

PR #1845, opened 2021-02-08
Add `global` as in nodejs for backwards compatibility

MStoykov:
this is only being added to the "extented" compatibility-mode as it was
being provided by core-js before it was dropped.


imiric (dismissed):
Can you add a small test for it?

imiric (approved):

na-- (approved):

PR #1848, opened 2021-02-08
Rewrite SharedArray to use goja.DynamicArray

### LINKED ISSUE #1820: SharedArray doesn't work with forEach
The following code:
```
import { SharedArray } from 'k6/data'
var csvData = new SharedArray("another data name", function() {
    return ["one", "two"];
});
export default function() {
    csvData.forEach((item) => console.log(item));
}
```

Gets 
```
 TypeError: Value is not an object: undefined
```
which needs to be investigated

MStoykov:
This should make it considerably faster, but probably more importantly, a
lot easier to reason about and also supporting all operations supported
by a normal Array in JavaScript.

From my quick testing it seems that now the difference even for 10
elements is within 5% CPU difference, which IMO is even within the
margin of error.

fixes #1820

### DIFF
--- a/js/modules/k6/data/share.go
+++ b/js/modules/k6/data/share.go
@@ -22,6 +22,7 @@ package data
 
 import (
 	"github.com/dop251/goja"
+	"github.com/loadimpact/k6/js/common"
 )
 
 // TODO fix it not working really well with setupData or just make it more broken
@@ -66,9 +67,12 @@ func (s wrappedSharedArray) Get(index int) goja.Value {
 	}
 	val, err := s.parse(goja.Undefined(), s.rt.ToValue(s.arr[index]))
 	if err != nil {
-		panic(err)
+		common.Throw(s.rt, err)
 	}
-	s.deepFreeze(s.rt, val)
+	err = s.deepFreeze(s.rt, val)
+	if err != nil {
+		common.Throw(s.rt, err)
+	}
 
 	return val
 }
@@ -77,14 +81,14 @@ func (s wrappedSharedArray) Len() int {
 	return len(s.arr)
 }
 
-func (s wrappedSharedArray) deepFreeze(rt *goja.Runtime, val goja.Value) {
+func (s wrappedSharedArray) deepFreeze(rt *goja.Runtime, val goja.Value) error {
 	_, err := s.freeze(goja.Undefined(), val)
 	if err != nil {
-		panic(s.rt.NewTypeError(err))
+		return err
 	}
 
 	if goja.IsUndefined(val) {
-		return
+		return nil
 	}
 
 	o := val.ToObject(rt)
@@ -93,11 +97,14 @@ func (s wrappedSharedArray) deepFreeze(rt *goja.Runtime, val goja.Value) {
 		if prop != nil {
 			frozen, err := s.isFrozen(goja.Undefined(), prop)
 			if err != nil {
-				panic(s.rt.NewTypeError(err))
+				return err
 			}
 			if !frozen.ToBoolean() {
-				s.deepFreeze(rt, prop)
+				if err = s.deepFreeze(rt, prop); err != nil {
+					return err
+				}
 			}
 		}
 	}
+	return nil
 }
--- a/js/modules/k6/data/share.go
+++ b/js/modules/k6/data/share.go
@@ -82,24 +82,28 @@ func (s wrappedSharedArray) Len() int {
 }
 
 func (s wrappedSharedArray) deepFreeze(rt *goja.Runtime, val goja.Value) error {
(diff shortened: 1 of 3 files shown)


js/modules/k6/data/share.go
```diff
@@ -33,103 +31,80 @@ type sharedArray struct {
...
 }
 
-func (s sharedArray) Get(index int) (interface{}, error) {
-	if index < 0 || index >= len(s.arr) {
-		return goja.Undefined(), nil
-	}
-
-	// we specifically use JSON.parse to get the json to an object inside as otherwise we won't be
-	// able to freeze it as goja doesn't let us unless it is a pure goja object and this is the
-	// easiest way to get one.
```
imiric, line 57: I'm still wondering why we stringify the array elements and then parse them back on `Get()`. Attempting to freeze a non-object [would simply return it as is](https://github.com/dop251/goja/blob/f5884268f0c0a0eb94423dcf7c2eda14907375e6/builtin_object.go#L296-L298), so the stringify/parse dance doesn't seem necessary and is probably a lot of overhead.

js/modules/k6/data/share.go
```diff
@@ -33,103 +31,80 @@ type sharedArray struct {
...
 	}
-	sai.index++
-	var tmp interface{}
-	if err := json.Unmarshal(sai.a.arr[sai.index], &tmp); err != nil {
-		return goja.Undefined(), err
+	val, err := s.parse(goja.Undefined(), s.rt.ToValue(s.arr[index]))
+	if err != nil {
+		common.Throw(s.rt, err)
+	}
+	err = s.deepFreeze(s.rt, val)
```
imiric, line 72: Could we `deepFreeze` all elements once upfront (in e.g. `getShareArrayFromCall()`), and avoid having to do this everytime `Get()` is called? I understand it's needed because it's parsed so it's a new object everytime, but if we can avoid having to stringify/parse then we could just return the same frozen object.

imiric (commented):
Looks much better without the `RunString` wrappers :+1:

Just left a few comments I probably should've made in the previous PR. :)

js/modules/k6/data/share.go
MStoykov: This was more or less explored in https://github.com/dop251/goja/issues/232 and the answer is that ... while for extremely simple values such as String and numbers it might be feasible ... for even Objects, it is likely that sharing them between goja Runtimes will lead to race conditions *even* if frozen. 

Again the EcmaScript standard is meant for *single-threaded* operations ... everything in the standard (except the few Atomic and *Shared* stuff inside of it) is meant for only 1 thing operating on objects from it. So even if theoretically today goja could support it, tomorrow due to change that might break. 

Additionally, if we don't serialize and then deserialize, we will need to actually check that things such as lambdas(which can reference outside objects inside of them) are not present. Proxies will also be problematic, as well as probably property methods and so on ... All of those things are not an issue as long as we serialize to JSON and back, because goja follows the EcmaScript standard and it actually specifically says that things like that don't get serialized :tada: .

Again the "overhead" is minimal and IMO the possible complications far outweigh the amount of additional reasoning and code that will be needed for this to work. Also again ... telling people that their data will be serialized to JSON and back is a thing that they can reason about as well, whatever else we come up with ... likely won't be so easy to reason about and test with available tools.

js/modules/k6/data/share.go
MStoykov: see https://github.com/loadimpact/k6/pull/1848#discussion_r572818058

js/modules/k6/data/share.go
```diff
@@ -33,103 +31,80 @@ type sharedArray struct {
...
+func (s wrappedSharedArray) deepFreeze(rt *goja.Runtime, val goja.Value) error {
+	_, err := s.freeze(goja.Undefined(), val)
+	if err != nil {
+		return err
+	}
 
-		return o;
-	};
+	if goja.IsUndefined(val) {
+		return nil
```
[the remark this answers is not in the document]
MStoykov: I just rewrote the JS snippet that was before in golang + goja.  I don't think it matters ... the snippet was [provided](https://github.com/loadimpact/k6/pull/1739#discussion_r540159964) by @na-- and the original code doesn't even have the check ... so I wonder if I shouldn't just drop it?

js/modules/k6/data/share.go
imiric: Thanks for explaining, that makes sense. I think I glossed over the initial discussion in that issue. It might be good to link to it in a comment as a future reference for why this is done.

imiric (approved):

na-- (approved):
LGTM, and yeah, much better than before! :heart: the new way of calling JS functions from inside of Go!

PR #1853, opened 2021-02-11
Fix throw not being respected on Malformed URL

### LINKED ISSUE #1502: `throw` option not respected with malformed URLs
This script demonstrates the issue:

```js
import { sleep } from "k6";
import http from "k6/http";

export const options = {
    iterations: 10, // just to demonstrate how the sleep() is never reached
    //throw: true, // it's false by default, uncomment to test the different behavior
};

export default function () {
    // this will throw an exception if `throw` is true
    let r = http.get("https://some-wrong-domain.k6.io");
    console.log(r.error_code);

    // this will always throw an exception in the current k6
    r = http.get("https:// test.k6.io");

    // these will never be reached
    console.log(r.error_code);
    sleep(1);
}
```

And this somewhat connected issue should probably be fixed at the same time: https://github.com/loadimpact/k6/issues/1199

gchaincl:
Closes #1502

Notes:
* Not sure if the verification should happen on Request()
* The way throw is disabled on the test feels clunky, it might be a better approach

### DIFF
--- a/js/modules/k6/http/request.go
+++ b/js/modules/k6/http/request.go
@@ -90,7 +90,10 @@ func (h *HTTP) Options(ctx context.Context, url goja.Value, args ...goja.Value) 
 // taking goja.Values as arguments
 func (h *HTTP) Request(ctx context.Context, method string, url goja.Value, args ...goja.Value) (*Response, error) {
 	state := lib.GetState(ctx)
-	throw := state.Options.Throw.Bool
+	var throw bool
+	if state != nil {
+		throw = state.Options.Throw.Bool
+	}
 	u, err := ToURL(url)
 	if err != nil {
 		if throw {
--- a/js/modules/k6/http/request.go
+++ b/js/modules/k6/http/request.go
@@ -89,9 +89,17 @@ func (h *HTTP) Options(ctx context.Context, url goja.Value, args ...goja.Value) 
 // Request makes an http request of the provided `method` and returns a corresponding response by
 // taking goja.Values as arguments
 func (h *HTTP) Request(ctx context.Context, method string, url goja.Value, args ...goja.Value) (*Response, error) {
+	state := lib.GetState(ctx)
+	throw := state.Options.Throw.Bool
 	u, err := ToURL(url)
 	if err != nil {
-		return nil, err
+		if throw {
+			return nil, err
+		}
+		state.Logger.WithField("error", err).Warn("Request Failed")
+		return &Response{
+			Response: &httpext.Response{},
+		}, nil
 	}
 
 	var body interface{}
--- a/js/modules/k6/http/request_test.go
+++ b/js/modules/k6/http/request_test.go
@@ -601,6 +601,35 @@ func TestRequestAndBatch(t *testing.T) {
 			}
 		})
 	})
+	t.Run("MalformedURL", func(t *testing.T) {
+		js := `
+			http.request("GET", "https:// test.k6.io");
+		`
+		t.Run("throw=true", func(t *testing.T) {
+			_, err := rt.RunString(js)
+			require.Error(t, err)
+		})
+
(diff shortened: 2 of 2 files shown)


js/modules/k6/http/request.go
```diff
@@ -89,9 +89,20 @@ func (h *HTTP) Options(ctx context.Context, url goja.Value, args ...goja.Value)
 // Request makes an http request of the provided `method` and returns a corresponding response by
 // taking goja.Values as arguments
 func (h *HTTP) Request(ctx context.Context, method string, url goja.Value, args ...goja.Value) (*Response, error) {
+	state := lib.GetState(ctx)
+	var throw bool
+	if state != nil {
+		throw = state.Options.Throw.Bool
+	}
```
na--, line 96: `state` should be non-`nil` here, though unfortunately that is checked only in `parseRequest()`, which is called further down below... 
https://github.com/loadimpact/k6/blob/a379c3787372ce2a4c70cb2049fc43be7af798a2/js/modules/k6/http/request.go#L107
https://github.com/loadimpact/k6/blob/a379c3787372ce2a4c70cb2049fc43be7af798a2/js/modules/k6/http/request.go#L121-L128

Given that we must implement the same `throw` check in `http.batch()` (i.e. fix the similar code [here](https://github.com/loadimpact/k6/blob/a379c3787372ce2a4c70cb2049fc43be7af798a2/js/modules/k6/http/request.go#L476-L479) and [here](https://github.com/loadimpact/k6/blob/a379c3787372ce2a4c70cb2049fc43be7af798a2/js/modules/k6/http/request.go#L491-L493) :disappointed: ), maybe the best way would be delaying the parsing of the URL and doing it in `parseRequest()`, so we only change one place?

na-- (changes requested):
Thanks for making this PR, it mostly looks good and the test especially is excellent! Unfortunately, as I've mentioned in one of the inline comments, we need to also have the same logic in [`http.batch()` calls](https://k6.io/docs/javascript-api/k6-http/batch-requests). So, you probably need to refactor things a bit and make the validation in some sort of a helper and/or fit it `parseRequest()`?

gchaincl:
Thanks for reviewing @na-- , I agree that we might need some refactoring, will have a look and see how to make it work for batching too.

js/modules/k6/http/request.go
MStoykov: I am for even moving further and in the `httpext` package:
1. otherwise error codes will need to become exported (and probably moved)
2. we already need to then stringify the URL back ... which is wasteful, given that we already have it as a string

This might have other implications as I haven't gone around the code to look for places where this will not be easier

MStoykov: I think https://github.com/loadimpact/k6/blob/a379c3787372ce2a4c70cb2049fc43be7af798a2/js/modules/k6/http/http_url.go#L49 will make moving all of the URL parsing in [httpext.MakeRequest](https://github.com/loadimpact/k6/blob/a379c3787372ce2a4c70cb2049fc43be7af798a2/lib/netext/httpext/request.go#L174) ... impossible :(

gchaincl: Since both, `Request` and `Batch` call `parseRequest()` I agree with @na--'s suggestion on moving the URL validation there. we could replace `reqURL httpext.URL` by `reqURL gojaValue`.

imiric:
@gchaincl Thanks for the PR. Are you available to refactor it as discussed above? We'd like to have it merged for v0.33.0.

If we don't hear from you by sometime next week or if you're not available--that's OK too--we will proceed to do the leftover changes.

imiric:
@gchaincl See the changes for this in #2045.

Thanks again for your contribution and we'll make sure to thank you in the release notes. :)

Issue #1828, opened 2021-01-28
Proposal for adding a new standard  `http_req_failures` metric 


sniku:
I'm opening a new issue instead of building on top of https://github.com/loadimpact/k6/issues/1311 to start the discussion with a  clean slate. I believe this proposal addresses most if not all issues brought up in the original proposal. 

Before discussing the implementation details and effort required to build this feature, let's discuss why we want this feature at all. 

1. Users want to know if their http requests fail or not, without adding boilerplate code.
2. Users want to see the http timing metrics for successful requests. 
3. Users want to see the ratio between successful and failed requests (0.1% failures is often acceptable)
4. Some users may want to know the absolute number of failed requests (15 requests failed)

## Basic requirements
This basic script must:
1. show the number of failed requests,
2. show response time for successful requests,
3. test must exit with non-0 exit code because the threshold is crossed.

```js
import { sleep } from 'k6'
import http from 'k6/http'

export let options = {
  thresholds: {
    // test fails if more than 10% of HTTP requests fail. 
    // default http_success_hook is used to determine successful status codes.
    http_reqs_failure: ['rate < 0.1'],
  }
};

export default function() {
 let response_success  = http.get("https://httpbin.test.k6.io/status/200");
 let response_failure1 = http.get("https://httpbin.test.k6.io/status/503");
 let response_failure2 = http.get("https://httpbin.test.k6.io/status/503");
}
```

## Discussion about the metric type for `http_reqs_failure`

There are two possible metric types for the new `http_reqs_failure`. Rate or Counter.
Both types have their advantages, and it's not entirely clear which one is better for this use case.

Advantages of Rate:
 - ability to easily configure thresholds `http_reqs_failure: rate<0.1`
 - shows rate... 

Advantages of Counter:
 - shows failures per second
 - shows the total number of failed requests (might be useful to some users)
 - is consistent with `http_reqs` metric. 

The end-of-test summary for this test should look similar to this:

## Output when using Counter metric
```sh
http_reqs..................: 3      3.076683/s
http_reqs_failure..........: 2      2.036683/s
http_reqs_success..........: 1      1.036683/s
```

## Output when using Rate metric

```sh
http_reqs..................: 3      3.134432/s
http_reqs_failure..........: 33.33% ✓ 1 ✗ 2
http_reqs_success..........: 66.66% ✓ 2 ✗ 1
```

Neither `Rate` nor `Counter` covers all possible use-cases. I think `Rate` is preferable over `Counter`. 

[...]

Issue #1716, opened 2020-11-09
Being able to edit metrics emitted by requests before that is done


mstoykov:
Currently, if you make an http request it will emit a bunch of metrics and a user can add additional tags before they make the request. What isn't possible is to modify the metrics after the request is finished based on something from the response (for example). 

Reported use case:

https://community.k6.io/t/track-transmitted-data-per-url-with-http-batch/1054 
Another user has a machine identifier in the response and would like to tag requests with the machines they come from (and they are not exactly 1-to-1 relationship to an IP)  

## Feature Description
Have a way to edit metrics emitted by requests before they are begin emitted but after the request  have finished.

## Suggested Solution (optional)
The proposed solution is one more parameter to [Params](https://k6.io/docs/javascript-api/k6-http/params)

```
http.request("GET", "https://test.k6.io", null, {
  preMetricEmission: function(response, metricObject) {
    metricObject["somethingCool"] = response.json()["somethingCool"];
  }
}
```

With the above code `preMetricEmission`(another proposed name was `postRequest`) must be called after the request is done but before we emit the metrics (after they are calculated though). The provided objects are http.Response (possibly with some things missing?) and metricObject should be a json representation of the metrics and maybe have some additional fields/methods. 

This also immediately gives a way of dropping unwanted http metrics by just removing them from the metricObject. This also means that you can edit the tags for each of the metrics which means that k6 should probably make it impossible to remove tags that are required by another part of k6. 

Another thing against this is that ... well the metrics won't be emitted until the function returns, also what should be the behavior when the function throws an exception? 

Obviously this also will be called multiple times if there is a redirect which will also help with this [user case](https://community.k6.io/t/custom-trend-including-all-redirects/301). 

The good part is that apart from those this will be fairly straightforward to be added in the current codebase ... probably :D

na--:
> The good part is that apart from those this will be fairly straightforward to be added in the current codebase ... probably :D

I am not so sure. Given that the callback will need to be executed in the same JS runtime as the other user code, one big potential issue are `http.batch()` calls, where we'd need to add some locking so that only a single callback is ever called at one time.

Also, currently the metric emission happens in here: https://github.com/loadimpact/k6/blob/012f1aab2444677cd4a7398e612a534ba51d3f3d/lib/netext/httpext/transport.go#L169

[...]

PR #1856, opened 2021-02-15
Add a way to mark http requests as failed or not(passed)

MStoykov:
This is done through running a callback on every request before emitting
the metrics. Currently only a built-in metric looking at good statuses
is possible, but a possibility for future JS based callbacks is left
open.

The implementation specifically makes it hard to figure out anything
about the returned callback from JS and tries not to change any other
code so it makes it easier for future implementation, but instead tries
to do the bare minimum without imposing any limitations on the future
work.

Additionally because it turned out to be easy, setting the callback to
null will make the http library to neither tag requests as passed nor
emit the new `http_req_failed` metric, essentially giving users a way to
go back to the previous behaviour.

The cloud output is also not changed as the `http_req_li` already is
aggregated based on tags so if an `http_req_li` is received that has tag
`passed:true` then the whole `http_req_li` is about requests that have
"passed".

part of #1828

### DIFF
--- a/js/internal/modules/modules.go
+++ b/js/internal/modules/modules.go
@@ -35,7 +35,18 @@ var (
 func Get(name string) interface{} {
 	mx.RLock()
 	defer mx.RUnlock()
-	return modules[name]
+	mod := modules[name]
+	if i, ok := mod.(HasModuleInstancePerVU); ok {
+		return i.NewModuleInstancePerVU()
+	}
+	return mod
+}
+
+// HasModuleInstancePerVU should be implemented by all native Golang modules that
+// would require per-VU state. k6 will call their NewModuleInstancePerVU() methods
+// every time a VU imports the module and use its result as the returned object.
+type HasModuleInstancePerVU interface {
+	NewModuleInstancePerVU() interface{}
 }
 
 // Register the given mod as a JavaScript module, available
--- a/js/modules/k6/http/http.go
+++ b/js/modules/k6/http/http.go
@@ -30,7 +30,7 @@ import (
 )
 
 func init() {
-	modules.Register("k6/http", New())
+	modules.Register("k6/http", new(GlobalHTTP))
 }
 
 const (
@@ -45,6 +45,36 @@ const (
 
 // ErrJarForbiddenInInitContext is used when a cookie jar was made in the init context
 var ErrJarForbiddenInInitContext = common.NewInitContextError("Making cookie jars in the init context is not supported")
+
+// GlobalHTTP is a global HTTP module for a k6 instance/test run
+type GlobalHTTP struct{}
+
+var _ modules.HasModuleInstancePerVU = new(GlobalHTTP)
+
+// NewModuleInstancePerVU returns an HTTP instance for each VU
+func (g *GlobalHTTP) NewModuleInstancePerVU() interface{} { // this here needs to return interface{}
+	return &HTTP{ // change the below fields to be not writable or not fields
+		SSL_3_0:                            netext.SSL_3_0,
+		TLS_1_0:                            netext.TLS_1_0,
+		TLS_1_1:                            netext.TLS_1_1,
+		TLS_1_2:                            netext.TLS_1_2,
+		TLS_1_3:                            netext.TLS_1_3,
+		OCSP_STATUS_GOOD:                   netext.OCSP_STATUS_GOOD,
+		OCSP_STATUS_REVOKED:                netext.OCSP_STATUS_REVOKED,
+		OCSP_STATUS_SERVER_FAILED:          netext.OCSP_STATUS_SERVER_FAILED,
+		OCSP_STATUS_UNKNOWN:                netext.OCSP_STATUS_UNKNOWN,
+		OCSP_REASON_UNSPECIFIED:            netext.OCSP_REASON_UNSPECIFIED,
+		OCSP_REASON_KEY_COMPROMISE:         netext.OCSP_REASON_KEY_COMPROMISE,
+		OCSP_REASON_CA_COMPROMISE:          netext.OCSP_REASON_CA_COMPROMISE,
+		OCSP_REASON_AFFILIATION_CHANGED:    netext.OCSP_REASON_AFFILIATION_CHANGED,
+		OCSP_REASON_SUPERSEDED:             netext.OCSP_REASON_SUPERSEDED,
+		OCSP_REASON_CESSATION_OF_OPERATION: netext.OCSP_REASON_CESSATION_OF_OPERATION,
+		OCSP_REASON_CERTIFICATE_HOLD:       netext.OCSP_REASON_CERTIFICATE_HOLD,
+		OCSP_REASON_REMOVE_FROM_CRL:        netext.OCSP_REASON_REMOVE_FROM_CRL,
+		OCSP_REASON_PRIVILEGE_WITHDRAWN:    netext.OCSP_REASON_PRIVILEGE_WITHDRAWN,
+		OCSP_REASON_AA_COMPROMISE:          netext.OCSP_REASON_AA_COMPROMISE,
+	}
+}
 
 //nolint: golint
 type HTTP struct {
@@ -67,31 +97,6 @@ type HTTP struct {
 	OCSP_REASON_REMOVE_FROM_CRL        string `js:"OCSP_REASON_REMOVE_FROM_CRL"`
 	OCSP_REASON_PRIVILEGE_WITHDRAWN    string `js:"OCSP_REASON_PRIVILEGE_WITHDRAWN"`
 	OCSP_REASON_AA_COMPROMISE          string `js:"OCSP_REASON_AA_COMPROMISE"`
-}
-
-func New() *HTTP {
-	//TODO: move this as an anonymous struct somewhere...
-	return &HTTP{
-		SSL_3_0:                            netext.SSL_3_0,
-		TLS_1_0:                            netext.TLS_1_0,
-		TLS_1_1:                            netext.TLS_1_1,
-		TLS_1_2:                            netext.TLS_1_2,
-		TLS_1_3:                            netext.TLS_1_3,
-		OCSP_STATUS_GOOD:                   netext.OCSP_STATUS_GOOD,
-		OCSP_STATUS_REVOKED:                netext.OCSP_STATUS_REVOKED,
-		OCSP_STATUS_SERVER_FAILED:          netext.OCSP_STATUS_SERVER_FAILED,
-		OCSP_STATUS_UNKNOWN:                netext.OCSP_STATUS_UNKNOWN,
-		OCSP_REASON_UNSPECIFIED:            netext.OCSP_REASON_UNSPECIFIED,
-		OCSP_REASON_KEY_COMPROMISE:         netext.OCSP_REASON_KEY_COMPROMISE,
-		OCSP_REASON_CA_COMPROMISE:          netext.OCSP_REASON_CA_COMPROMISE,
-		OCSP_REASON_AFFILIATION_CHANGED:    netext.OCSP_REASON_AFFILIATION_CHANGED,
-		OCSP_REASON_SUPERSEDED:             netext.OCSP_REASON_SUPERSEDED,
-		OCSP_REASON_CESSATION_OF_OPERATION: netext.OCSP_REASON_CESSATION_OF_OPERATION,
-		OCSP_REASON_CERTIFICATE_HOLD:       netext.OCSP_REASON_CERTIFICATE_HOLD,
-		OCSP_REASON_REMOVE_FROM_CRL:        netext.OCSP_REASON_REMOVE_FROM_CRL,
-		OCSP_REASON_PRIVILEGE_WITHDRAWN:    netext.OCSP_REASON_PRIVILEGE_WITHDRAWN,
-		OCSP_REASON_AA_COMPROMISE:          netext.OCSP_REASON_AA_COMPROMISE,
-	}
 }
 
 func (*HTTP) XCookieJar(ctx *context.Context) *HTTPCookieJar {
--- a/js/modules/k6/http/http.go
+++ b/js/modules/k6/http/http.go
@@ -73,6 +73,8 @@ func (g *GlobalHTTP) NewModuleInstancePerVU() interface{} { // this here needs t
 		OCSP_REASON_REMOVE_FROM_CRL:        netext.OCSP_REASON_REMOVE_FROM_CRL,
 		OCSP_REASON_PRIVILEGE_WITHDRAWN:    netext.OCSP_REASON_PRIVILEGE_WITHDRAWN,
 		OCSP_REASON_AA_COMPROMISE:          netext.OCSP_REASON_AA_COMPROMISE,
+
+		responseCallback: defaultExpectedStatuses.match,
 	}
 }
 
@@ -97,6 +99,8 @@ type HTTP struct {
 	OCSP_REASON_REMOVE_FROM_CRL        string `js:"OCSP_REASON_REMOVE_FROM_CRL"`
 	OCSP_REASON_PRIVILEGE_WITHDRAWN    string `js:"OCSP_REASON_PRIVILEGE_WITHDRAWN"`
 	OCSP_REASON_AA_COMPROMISE          string `js:"OCSP_REASON_AA_COMPROMISE"`
+
+	responseCallback func(int) bool
 }
 
 func (*HTTP) XCookieJar(ctx *context.Context) *HTTPCookieJar {
--- a/js/modules/k6/http/response_callback.go
+++ b/js/modules/k6/http/response_callback.go
@@ -1,0 +1,117 @@
+/*
+ *
+ * k6 - a next-generation load testing tool
+ * Copyright (C) 2021 Load Impact
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+package http
+
+import (
+	"context"
+	"errors"
+	"fmt"
+
+	"github.com/dop251/goja"
+	"github.com/loadimpact/k6/js/common"
+)
+
+//nolint:gochecknoglobals
+var defaultExpectedStatuses = expectedStatuses{
+	minmax: [][2]int{{200, 399}},
+}
+
+// expectedStatuses is specifically totally unexported so it can't be used for anything else but
+// SetResponseCallback and nothing can be done from the js side to modify it or make an instance of
+// it except using ExpectedStatuses
+type expectedStatuses struct {
+	minmax [][2]int
+	exact  []int
+}
+
+func (e expectedStatuses) match(status int) bool {
+	for _, v := range e.exact {
+		if v == status {
+			return true
+		}
+	}
+
+	for _, v := range e.minmax {
+		if v[0] <= status && status <= v[1] {
+			return true
+		}
+	}
+	return false
+}
+
+// ExpectedStatuses returns expectedStatuses object based on the provided arguments.
+// The arguments must be either integers or object of `{min: <integer>, max: <integer>}`
+// kind. The "integer"ness is checked by the Number.isInteger.
+func (*HTTP) ExpectedStatuses(ctx context.Context, args ...goja.Value) *expectedStatuses { //nolint: golint
+	rt := common.GetRuntime(ctx)
+
+	if len(args) == 0 {
+		common.Throw(rt, errors.New("no arguments"))
+	}
+	var result expectedStatuses
+
+	jsIsInt, _ := goja.AssertFunction(rt.GlobalObject().Get("Number").ToObject(rt).Get("isInteger"))
+	isInt := func(a goja.Value) bool {
+		v, err := jsIsInt(goja.Undefined(), a)
+		return err == nil && v.ToBoolean()
+	}
+
+	errMsg := "argument number %d to expectedStatuses was neither an integer nor an object like {min:100, max:329}"
+	for i, arg := range args {
+		o := arg.ToObject(rt)
+		if o == nil {
+			common.Throw(rt, fmt.Errorf(errMsg, i+1))
+		}
+
+		if isInt(arg) {
+			result.exact = append(result.exact, int(o.ToInteger()))
+		} else {
+			min := o.Get("min")
+			max := o.Get("max")
+			if min == nil || max == nil {
+				common.Throw(rt, fmt.Errorf(errMsg, i+1))
+			}
+			if !(isInt(min) && isInt(max)) {
+				common.Throw(rt, fmt.Errorf("both min and max need to be integers for argument number %d", i+1))
+			}
+
+			result.minmax = append(result.minmax, [2]int{int(min.ToInteger()), int(max.ToInteger())})
+		}
+	}
+	return &result
+}
+
+// SetResponseCallback sets the responseCallback to the value provided. Supported values are
+// expectedStatuses object or a `null` which means that metrics shouldn't be tagged as failed and
+// `http_req_failed` should not be emitted - the behaviour previous to this
+func (h *HTTP) SetResponseCallback(ctx context.Context, val goja.Value) {
+	if val != nil && !goja.IsNull(val) {
+		// This is done this way as ExportTo exports functions to empty structs without an error
+		if es, ok := val.Export().(*expectedStatuses); ok {
+			h.responseCallback = es.match
+		} else {
+			//nolint:golint
+			common.Throw(common.GetRuntime(ctx), fmt.Errorf("unsupported argument, expected http.expectedStatuses"))
+		}
+	} else {
+		h.responseCallback = nil
+	}
+}
--- a/js/modules/k6/http/response_callback_test.go
+++ b/js/modules/k6/http/response_callback_test.go
@@ -1,0 +1,562 @@
+/*
+ *
+ * k6 - a next-generation load testing tool
+ * Copyright (C) 2021 Load Impact
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+package http
+
+import (
+	"context"
+	"fmt"
+	"sort"
+	"testing"
+
+	"github.com/dop251/goja"
+	"github.com/loadimpact/k6/js/common"
+	"github.com/loadimpact/k6/lib"
+	"github.com/loadimpact/k6/lib/metrics"
+	"github.com/loadimpact/k6/stats"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestExpectedStatuses(t *testing.T) {
+	t.Parallel()
+	rt := goja.New()
+	rt.SetFieldNameMapper(common.FieldNameMapper{})
+	ctx := context.Background()
+
+	ctx = common.WithRuntime(ctx, rt)
+	rt.Set("http", common.Bind(rt, new(GlobalHTTP).NewModuleInstancePerVU(), &ctx))
+	cases := map[string]struct {
+		code, err string
+		expected  expectedStatuses
+	}{
+		"good example": {
+			expected: expectedStatuses{exact: []int{200, 300}, minmax: [][2]int{{200, 300}}},
+			code:     `(http.expectedStatuses(200, 300, {min: 200, max:300}))`,
+		},
+
+		"strange example": {
+			expected: expectedStatuses{exact: []int{200, 300}, minmax: [][2]int{{200, 300}}},
+			code:     `(http.expectedStatuses(200, 300, {min: 200, max:300, other: "attribute"}))`,
+		},
+
+		"string status code": {
+			code: `(http.expectedStatuses(200, "300", {min: 200, max:300}))`,
+			err:  "argument number 2 to expectedStatuses was neither an integer nor an object like {min:100, max:329}",
+		},
+
+		"string max status code": {
+			code: `(http.expectedStatuses(200, 300, {min: 200, max:"300"}))`,
+			err:  "both min and max need to be integers for argument number 3",
+		},
+		"float status code": {
+			err:  "argument number 2 to expectedStatuses was neither an integer nor an object like {min:100, max:329}",
+			code: `(http.expectedStatuses(200, 300.5, {min: 200, max:300}))`,
+		},
+
+		"float max status code": {
+			err:  "both min and max need to be integers for argument number 3",
+			code: `(http.expectedStatuses(200, 300, {min: 200, max:300.5}))`,
+		},
+		"no arguments": {
+			code: `(http.expectedStatuses())`,
+			err:  "no arguments",
+		},
+	}
+
+	for name, testCase := range cases {
+		name, testCase := name, testCase
+		t.Run(name, func(t *testing.T) {
+			val, err := rt.RunString(testCase.code)
+			if testCase.err == "" {
+				require.NoError(t, err)
+				got := new(expectedStatuses)
+				err = rt.ExportTo(val, &got)
+				require.NoError(t, err)
+				require.Equal(t, testCase.expected, *got)
+				return // the t.Run
+			}
+
+			require.Error(t, err)
+			exc := err.(*goja.Exception)
+			require.Contains(t, exc.Error(), testCase.err)
+		})
+	}
+}
+
+type expectedSample struct {
+	tags    map[string]string
+	metrics []*stats.Metric
+}
+
+func TestResponseCallbackInAction(t *testing.T) {
+	t.Parallel()
+	tb, _, samples, rt, ctx := newRuntime(t)
+	defer tb.Cleanup()
+	sr := tb.Replacer.Replace
+	httpModule := new(GlobalHTTP).NewModuleInstancePerVU().(*HTTP)
+	rt.Set("http", common.Bind(rt, httpModule, ctx))
+
+	HTTPMetricsWithoutFailed := []*stats.Metric{
+		metrics.HTTPReqs,
+		metrics.HTTPReqBlocked,
+		metrics.HTTPReqConnecting,
+		metrics.HTTPReqDuration,
+		metrics.HTTPReqReceiving,
+		metrics.HTTPReqWaiting,
+		metrics.HTTPReqSending,
+		metrics.HTTPReqTLSHandshaking,
+	}
+
+	allHTTPMetrics := append(HTTPMetricsWithoutFailed, metrics.HTTPReqFailed)
+
+	testCases := map[string]struct {
+		code            string
+		expectedSamples []expectedSample
+	}{
+		"basic": {
+			code: `http.request("GET", "HTTPBIN_URL/redirect/1");`,
+			expectedSamples: []expectedSample{
+				{
+					tags: map[string]string{
+						"method":            "GET",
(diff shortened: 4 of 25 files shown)


lib/netext/httpext/transport.go
```diff
@@ -164,8 +168,33 @@ func (t *transport) measureAndEmitMetrics(unfReq *unfinishedRequest) *finishedRe
...
+	var failed float64
+	if t.responseCallback != nil {
+		var statusCode int
+		if unfReq.response != nil {
+			statusCode = unfReq.response.StatusCode
+		}
+		passed := t.responseCallback(statusCode)
+		if enabledTags.Has(stats.TagPassed) {
+			if passed {
+				failed = 0
```
imiric, line 180: Currently the metric value depends on whether the `passed` tag is enabled or not. These should be decoupled so that the metric emits the correct value regardless.

imiric (changes requested):
The panic and tag/metric coupling are blockers to me, and the rest is up for discussion.

js/modules/k6/http/response_callback.go
```diff
@@ -0,0 +1,110 @@
...
+	o := a.ToObject(rt)
+	return o != nil && o.ClassName() == "Number"
+}
+
+// SetResponseCallback ..
+func (h HTTP) SetResponseCallback(ctx context.Context, es *expectedStatuses) {
+	if es != nil {
+		lib.GetState(ctx).HTTPResponseCallback = es.match
+	} else {
+		lib.GetState(ctx).HTTPResponseCallback = nil
```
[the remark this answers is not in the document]
MStoykov: > This panics with nil pointer when run from the init context

:facepalm: 

I guess this needs #1858 :thinking:

js/modules/k6/http/response_callback.go
```diff
@@ -0,0 +1,110 @@
...
+	return &result
+}
+
+func checkNumber(a goja.Value, rt *goja.Runtime) bool {
+	o := a.ToObject(rt)
+	return o != nil && o.ClassName() == "Number"
+}
+
+// SetResponseCallback ..
+func (h HTTP) SetResponseCallback(ctx context.Context, es *expectedStatuses) {
```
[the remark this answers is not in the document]
MStoykov: The whole point is that in the future it will be able to be a callback, we just didn't want to design all the arguments and how it will behave currently (which is #1716 ). 

So we have only a single way to get the current "callback" ... unfortunately it seems that `function() {}` gets exported to `*expectedStatus`   without a problem in this case :thinking: ... which I expected will be an error (it is an error with for example just giving `200`)

lib/netext/httpext/transport.go
MStoykov: yeah .. I guess I should move this,  2 lines up, outside the check for the tag :thinking:

js/modules/k6/http/response_callback_test.go
```diff
@@ -0,0 +1,307 @@
...
+		"string status code": {
+			code: `(http.expectedStatuses(200, "300", {min: 200, max:300}))`,
+			err:  "argument number 2 to expectedStatuses was neither an integer nor a an object like {min:100, max:329}",
+		},
+
+		"string max status code": {
+			code: `(http.expectedStatuses(200, 300, {min: 200, max:"300"}))`,
+			err:  "both min and max need to be number for argument number 3",
+		},
+		"float status code": { // TODO probably should not work
```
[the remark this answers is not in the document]
MStoykov: I guess I can call [`Number.isInteger`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isInteger) :thinking:

js/modules/k6/http/response_callback.go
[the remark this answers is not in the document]
MStoykov: I think though that it is fairly important for people to be able to call 
```javascript
http.setResponseCallback(http.expectedStatuses(200));
export default function() {}
```
Instead of having to do it in the default function, either creating a new "callback" each time or having to add `__ITER == 0` check  ... so maybe I should try to implement #1858 specifically for this `HTTP` ... now ?

cc @na-- @imiric @sniku

js/modules/k6/http/response_callback.go
```diff
@@ -0,0 +1,111 @@
...
+			result.minmax = append(result.minmax, [2]int{int(min.ToInteger()), int(max.ToInteger())})
+		}
+	}
+	return &result
+}
+
+func checkNumber(a goja.Value, rt *goja.Runtime) bool {
+	c, _ := goja.AssertFunction(rt.GlobalObject().Get("Number").ToObject(rt).Get("isInteger"))
+	v, err := c(goja.Undefined(), a)
+	return err == nil && v.ToBoolean()
```
MStoykov, line 101: Technically goja should be allowing users to overwrite both `Number` and `isInteger` so this has the potential to `panic`, but to be honest I am not certain guarding for this will be all that more useful than just panicking.

 And unfortunately, I am not certain how to get this in any other way that will be more ... safe.

js/modules/k6/http/response_callback.go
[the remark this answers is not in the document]
sniku: I agree that it's important to be able to set the callback outside of the default function. Setting callback inside of the default function feels very wrong.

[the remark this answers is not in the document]
MStoykov: @sniku it does have the nice benefit of being able to say that the next 3 requests expect status code `401` for example and then go back to expecting 200-399

js/modules/k6/http/response_callback_test.go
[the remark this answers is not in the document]
MStoykov: done in dac7f05

lib/netext/httpext/transport.go
MStoykov: fixed in 17e5cc2

js/modules/k6/http/response_callback.go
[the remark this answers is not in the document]
na--: I still haven't looked at the code of this PR properly, will try to do so later today, so I don't know the technical difficulties yet. But from a UX perspective, being able to set the response callback in in both places (init and VU context) is best, but only slightly better than being able to set it only in the init context. Being able to set it in the init context is much more important than being able to set it in the VU context, because you can overwrite it for individual requests if you need to.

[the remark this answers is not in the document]
MStoykov: fixed in 7e0b787

js/internal/modules/modules.go
```diff
@@ -35,7 +35,13 @@ var (
 func Get(name string) interface{} {
 	mx.RLock()
 	defer mx.RUnlock()
-	return modules[name]
+	mod := modules[name]
+	if i, ok := mod.(interface {
+		NewGlobalModule() interface{ NewModuleInstance() interface{} }
+	}); ok {
```
imiric, line 41: This is a bit funky, but it's to avoid an import from `internal`, right? A comment would be nice to have.

js/modules/k6/http/http.go
```diff
@@ -46,6 +46,44 @@ const (
 // ErrJarForbiddenInInitContext is used when a cookie jar was made in the init context
 var ErrJarForbiddenInInitContext = common.NewInitContextError("Making cookie jars in the init context is not supported")
 
+// RootModule is only used to be registered and then to create a global (per k6 instance) module instance
+type RootModule struct{}
+
+// NewGlobalModule return a new GlobalHTTP
```
imiric, line 52:
```suggestion
// NewGlobalModule returns a new GlobalHTTP
```

js/modules/k6/http/response_callback.go
```diff
@@ -0,0 +1,110 @@
...
+
+	for _, v := range e.minmax { // binary search
+		if v[0] <= status && status <= v[1] {
+			return true
+		}
+	}
+	return false
+}
+
+// ExpectedStatuses is ...
```
imiric, line 57: Reminder to fill this out :)

js/modules/k6/http/response_callback.go
```diff
@@ -0,0 +1,110 @@
...
+	return &result
+}
+
+func checkNumber(a goja.Value, rt *goja.Runtime) bool {
+	c, _ := goja.AssertFunction(rt.GlobalObject().Get("Number").ToObject(rt).Get("isInteger"))
+	v, err := c(goja.Undefined(), a)
+	return err == nil && v.ToBoolean()
+}
+
+// SetResponseCallback ..
```
imiric, line 98: Same here.

imiric (commented):
LGTM, just some missing documentation.

js/internal/modules/modules.go
MStoykov: It was mostly because I didn't want to create a named interface in some package.
But if the approach is approved I am probably going to stuff it in `common` as I prefer the longer names (not just `New`) and this has potential for writing it wrong. 

I would argue some more "integration" tests will fix that, but it will be nicer to not need those for just checking that you provided the correct `interface`

MStoykov: I can also move just the `HTTP` changes to a separate commit or PR they are kind of small really

na--: I'd definitely prefer named interfaces. Also, why the 2-level design here? What is the benefit of that compared to having just a single interface with `NewModuleInstance()`?

js/modules/k6/http/http.go
```diff
@@ -67,10 +105,14 @@ type HTTP struct {
 	OCSP_REASON_REMOVE_FROM_CRL        string `js:"OCSP_REASON_REMOVE_FROM_CRL"`
 	OCSP_REASON_PRIVILEGE_WITHDRAWN    string `js:"OCSP_REASON_PRIVILEGE_WITHDRAWN"`
 	OCSP_REASON_AA_COMPROMISE          string `js:"OCSP_REASON_AA_COMPROMISE"`
+
+	responseCallback func(int) bool
 }
 
+// New ...
+// TODO deprecate this method
```
na--, line 113: Is there a reason for not using this in `NewModuleInstance()`? Or for not deleting it now, if you don't want to use it?

samples/http_2.js
```diff
@@ -1,9 +1,16 @@
 import http from "k6/http";
 import { check } from "k6";
 
+if (__VU == 1) {
+    http.setResponseCallback(http.expectedStatuses(200))
+} else {
+    http.setResponseCallback(http.expectedStatuses(300))
+}
 export default function () {
+    console.log(JSON.stringify(http));
```
na--, line 10: This seems like some debug code that snuck in

js/modules/k6/http/response_callback.go
```diff
@@ -0,0 +1,118 @@
...
+// expectedStatuses is specifically totally unexported so it can't be used for anything else but
+// SetResponseCallback and nothing can be done from the js side to modify it or make an instance of
+// it except using ExpectedStatuses
+type expectedStatuses struct {
+	minmax [][2]int
+	exact  []int // this can be done with the above and vice versa
+}
+
+func (e expectedStatuses) match(status int) bool {
+	for _, v := range e.exact { // binary search
```
na--, line 46: Are this comment and the one below `TODO`s for the future? :smile:

js/modules/k6/http/response_callback.go
```diff
@@ -0,0 +1,118 @@
...
+	if len(args) == 0 {
+		common.Throw(rt, errors.New("no arguments"))
+	}
+	var result expectedStatuses
+
+	for i, arg := range args {
+		o := arg.ToObject(rt)
+		if o == nil {
+			//nolint:lll
+			common.Throw(rt, fmt.Errorf("argument number %d to expectedStatuses was neither an integer nor an object like {min:100, max:329}", i+1))
```
na--, line 75:
```suggestion
			common.Throw(rt, fmt.Errorf(
				"argument number %d to expectedStatuses was neither an integer nor an object like {min:100, max:329}", i+1,
			))
```

js/modules/k6/http/response_callback.go
```diff
@@ -0,0 +1,118 @@
...
+		}
+	}
+	return &result
+}
+
+func checkNumber(a goja.Value, rt *goja.Runtime) bool {
+	c, _ := goja.AssertFunction(rt.GlobalObject().Get("Number").ToObject(rt).Get("isInteger"))
+	v, err := c(goja.Undefined(), a)
+	return err == nil && v.ToBoolean()
+}
```
na--, line 101: A bunch of minor nitpicks - no need for `nolint`, no need to get the `Number.isInteger` for every call, slightly confusing error message, etc.

```suggestion
func (*HTTP) ExpectedStatuses(ctx context.Context, args ...goja.Value) *expectedStatuses { //nolint: golint
	rt := common.GetRuntime(ctx)

	if len(args) == 0 {
		common.Throw(rt, errors.New("no arguments"))
	}
	var result expectedStatuses

	jsIsInt, _ := goja.AssertFunction(rt.GlobalObject().Get("Number").ToObject(rt).Get("isInteger"))
	isInt := func(a goja.Value) bool {
		v, err := jsIsInt(goja.Undefined(), a)
		return err == nil && v.ToBoolean()
	}

	errMsg := "argument number %d to expectedStatuses was neither an integer nor an object like {min:100, max:329}"
	for i, arg := range args {
		o := arg.ToObject(rt)
		if o == nil {
			common.Throw(rt, fmt.Errorf(errMsg, i+1))
		}

		if isInt(arg) {
			result.exact = append(result.exact, int(o.ToInteger()))
		} else {
			min := o.Get("min")
			max := o.Get("max")
			if min == nil || max == nil {
				common.Throw(rt, fmt.Errorf(errMsg, i+1))
			}
			if !(isInt(min) && isInt(max)) {
				common.Throw(rt, fmt.Errorf("both min and max need to be integers for argument number %d", i+1))
			}

			result.minmax = append(result.minmax, [2]int{int(min.ToInteger()), int(max.ToInteger())})
		}
	}
	return &result
}
```

lib/netext/httpext/request.go
```diff
@@ -269,8 +270,24 @@ func MakeRequest(ctx context.Context, preq *ParsedHTTPRequest) (*Response, error
...
+				return status == 401
+			}
+		}
 		transport = digestTransport{originalTransport: transport}
 	} else if preq.Auth == "ntlm" {
+		if tracerTransport.responseCallback != nil {
+			originalResponseCallback := tracerTransport.responseCallback
+			tracerTransport.responseCallback = func(status int) bool {
+				tracerTransport.responseCallback = originalResponseCallback
+				return status == 401
```
na--, line 288: Hmm I'm not sure this is correct. I think it should be something like `status == 401 || originalResponseCallback(status)`, considering that NTLM is a connection-level authentication IIRC. So, it's possible for the first request to actually pass through completely fine without any authentication, if it was reusing a connection that was previously authenticated.

Something similar will be true for digest auth eventually, when we get around to https://github.com/loadimpact/k6/issues/800, but as it is should be fine for now, I think...

lib/netext/httpext/transport.go
```diff
@@ -164,8 +168,33 @@ func (t *transport) measureAndEmitMetrics(unfReq *unfinishedRequest) *finishedRe
...
+		var statusCode int
+		if unfReq.response != nil {
+			statusCode = unfReq.response.StatusCode
+		}
+		expected := t.responseCallback(statusCode)
+		if expected {
+			failed = 0
+		} else {
+			failed = 1
+		}
```
na--, line 182:
```suggestion
		if !expected {
			failed = 1
		}
```

lib/netext/httpext/transport.go
```diff
@@ -164,8 +168,33 @@ func (t *transport) measureAndEmitMetrics(unfReq *unfinishedRequest) *finishedRe
 			tags["ip"] = ip
 		}
 	}
+	var failed float64
```
na--, line 171: I read your comments in the PR description about the cloud output, that you can derive the state of the HTTP request by the `expected_response` tag. And while that's true, I think we should limit the magic and also send `metrics.HTTPReqFailed.Name: 0/1` in the map here, for exactly the same reason we send `http_reqs`: https://github.com/loadimpact/k6/blob/09ea784d59611d46b071b205cb2ba2d1521a8199/stats/cloud/data.go#L115-L130

After all, users would likely be setting thresholds on this metrics, we should make it possible to do so in the cloud as well.

js/modules/k6/http/response_callback_test.go
```diff
@@ -0,0 +1,444 @@
...
+						"url":    sr("HTTPBIN_URL/get"),
+						"name":   sr("HTTPBIN_URL/get"),
+						"status": "200",
+						"group":  "",
+						"proto":  "HTTP/1.1",
+					},
+					metrics: HTTPMetricsWithoutFailed,
+				},
+			},
+		},
```
na--, line 280: Please add a `http.batch()` test as well. There shouldn't be any concurrency issues right now, given how restricted the callbacks are (still :heart: how you managed to find that workaround!), but I'd like us to have such a test now, before we add any more functionality here.

na-- (commented):
Besides the relatively minor issues I've noted as inline comments, overall this LGTM! We're going to have our work cut out for us, when we start to build on top of this with more generic callbacks, but for now this is more than good enough.

js/modules/k6/http/response_callback.go
MStoykov: more or less ... :P I decided against optimizing it before I get some code reviews and ... have forgotten.
I would also argue that it will require checking that there are no repeated status codes or ranges that are interleaving and so on ... which will just be even *more* tests for stuff that aren't essential all that much, but will be nice in the final version.

samples/http_2.js
MStoykov: yeah ... I apparently once didn't hit `n` when `git add -p`-ing :facepalm:

js/modules/k6/http/http.go
MStoykov: predominantly going through all the code and changing ... in case we decided that this was not going to be the solution ... also initially I thought there are like 50 places where it is used, but it's a lot less actually ... So I am probably going to drop it.

js/internal/modules/modules.go
MStoykov: Basically what I said in this [comment](https://github.com/loadimpact/k6/issues/1858#issuecomment-779891465)

Without the two levels, any "global" variables for each VU for a given module will need to be part of the one struct (or global variables). 

This makes less of sense for our production code as there we only have 1 instance of those, but for tests I would argue it will be nice to have multiple global module states without the need to clean them between tests ;). 

I am actually currently trying to move this to a separate PR with named interfaces and everything but the "global module instance per test" actually needs even more support in the modules registration framework as it technically needs to be made of instance per test and then that one instance to be used to create per VU instances of the module. That is currently not used by anything but the `data/SharedArray` can use that to not actually have to use `context` to store its data ;). 

I am pretty sure we can make it work without one of the levels(by just skipping the "root" struct) but I think this makes it more explicit and let us (if needed) have a shared state for all the global instances of a module - I can't currently figure out what that will be used for (neither in production code nor in the k6 tests) but :shrug: I prefer not to have to refactor this again just because we figure out a case.

MStoykov: Yeah ... I also think I prefer to try to make that PR after this is merged ... either as is (in this parts of the code) or with small changes, as it turns out it will need to be slightly bigger and will probably also include slight changes to other modules in order to verify that what  we provide as interfaces is sufficient.

For example while trying to do it I figure out that we can technically give some vuState (unlike the lib.State) to each module instance - that can possibly have the `context` inside of it. That context needs to change between calls to the `iterations` so we can't just give it a `context`, so I am not particularly certain it is a great idea but might be what we use to get the context around instead of the current bridge :shrug:

js/modules/k6/http/response_callback.go
MStoykov: the `nolint` is there because I am returning an unexported struct, changing to an empty interface works but I prefer the help of the compiler than of the linter so ;)

lib/netext/httpext/request.go
MStoykov: hm ... I guess you are right, will change it, but I have no way of testing it so :shrug:

lib/netext/httpext/transport.go
MStoykov: :facepalm:

js/modules/k6/http/response_callback.go
na--: I didn't even see that one :sweat_smile: I meant `//nolint:lll`

js/modules/k6/http/response_callback.go
na--: Yeah, I doubt a binary search would help here, unless someone decides to list all possible status codes and not use ranges. Just remove the comment for now

js/internal/modules/modules.go
na--: I am very confused, to be honest :sweat_smile: I don't see how you's be able to have both per-module and per-VU state with the current code, given that you currently _always_ make **both** a new "global" module and a new module instance for every VU that imports the module: https://github.com/loadimpact/k6/blob/71a349072b43f59ec716c1c72ee16d0a9c4c8fba/js/internal/modules/modules.go#L42

Besides, 
>  I can't currently figure out what that will be used for (neither in production code nor in the k6 tests) but :shrug:

Weren't you the one advocating for the [YAGNI principle](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) before? :smile: If you want to merge this as it is, please remove one level of interfaces here and maybe add it again in the proper module. I also think that we might just add a separate interface in the future, if we ever need both per-module and per-VU storage in a module, but we can discuss that in the other PR you want to make.

MStoykov: The point is that I want to make 
- per vu modules(NewModuleInstance) for this PR as that is what we need
- per test modules (NewGlobalModule) for things such as the `data/SharedArray`. This is currently not supported ... as well I haven't implemented it ;). Arguably it can be used for something in the HTTP module (or any other modules as well)
- the root module (naming stuff is hard :sob: ) is basically there so we can have something to generate the GlobalModule.

 Arguably it can just be a function that returns it not an interface (although as we know from [net/http](https://golang.org/pkg/net/http/#HandleFunc) those are isomorphic ;) ) which I guess will be ... easier to write by someone developing a module? But I also think that having 2 interfaces and three "types" RootModule, GlobalModule, ModuleInstance (have I mentioned naming stuff is hard) is easier to reason about :man_shrugging: 
- 
My current plan is that the modules registry will get the RootModule(whether a function or struct/interface) and then to get a module instance out you will need to instantiate the module *registry* which will call `NewGlobalModule` once and cache it for each module that is imported and call NewModuleInstance on that GlobalModule each time that is initialized. This (the module registry initialization) is IMO mostly useful for testing as otherwise if you decide to use the module registry in tests you will get 1 *shared* GlobalModule for all the test in a single *golang* package.

I guess I can drop some of it .. but if I am going to remake it, either way, remaking it in this PR seems pointless to me ;) We are not making this a public interface yet, or documentation it ...

lib/netext/httpext/transport.go
MStoykov: AFAIK the cloud has very custom logic on parsing those, which means that just adding it here *will* not help. There will still need to be changes to the code that handles it. 

Arguably it is probably better to add it, but I decided against it for two reasons:
1. it adds nothing and now there are counter, trend and Rate metrics here with absolutely no distinction between them that isn't "magical" - the cloud just knows.
2. unlike all the other metrics HTTPReqFailed can be disabled. Even if we revert this functionality this is something that we intend on readding with #1321 ... From that perspective maybe having it here when it is enabled and emitted is better as it will be future proof for when it probably shouldn't be here, and the cloud should not make it out of the tag :shrug: 

I am not hard against it by far ... it just seems like something that we will just add so it's added ... not like something that will be needed

js/internal/modules/modules.go
na--: :+1: :man_shrugging: :sweat_smile: You are right, something like a global module instance would be better place for the `SharedArray` values than the current `lib.State` BS (we probably want to remove even more things from there!), and per-VU storage would be useful for things like NTLM and digest authentication caches.

lib/netext/httpext/transport.go
na--: It's about consistency... And, for that matter, the `expected_response` tag can be disabled, so `http_req_failed` needs to be in the `http_req_li_all` map.

js/modules/k6/http/response_callback_test.go
MStoykov: db01bba

stats/units.go
```diff
@@ -37,3 +37,11 @@ func D(d time.Duration) float64 {
 func ToD(d float64) time.Duration {
 	return time.Duration(d * float64(timeUnit))
 }
+
+// B formats a boolena for emission
```
na--, line 41:
```suggestion
// B formats a boolean value for emission.
```

lib/netext/httpext/request.go
```diff
@@ -269,8 +270,25 @@ func MakeRequest(ctx context.Context, preq *ParsedHTTPRequest) (*Response, error
 	}
 
 	if preq.Auth == "digest" {
+		// In both digest and NTLM it is expected that the first response will be 401
+		// this does mean that a non 401 response will be marked as failed/unexpected
```
na--, line 274:
```suggestion
		// Until digest authentication is refactored, the first response will always 
		// be a 401 error, so we expect that.
```

lib/netext/httpext/request.go
```diff
@@ -269,8 +270,25 @@ func MakeRequest(ctx context.Context, preq *ParsedHTTPRequest) (*Response, error
...
+		if tracerTransport.responseCallback != nil {
+			originalResponseCallback := tracerTransport.responseCallback
+			tracerTransport.responseCallback = func(status int) bool {
+				tracerTransport.responseCallback = originalResponseCallback
+				return status == 401
+			}
+		}
 		transport = digestTransport{originalTransport: transport}
 	} else if preq.Auth == "ntlm" {
+		if tracerTransport.responseCallback != nil {
```
na--, line 284:
```suggestion
		// The first response of NTLM auth may be a 401 error.
		if tracerTransport.responseCallback != nil {
```

lib/netext/httpext/transport.go
```diff
@@ -164,8 +168,35 @@ func (t *transport) measureAndEmitMetrics(unfReq *unfinishedRequest) *finishedRe
...
-	trail.SaveSamples(stats.IntoSampleTags(&tags))
+	finalTags := stats.IntoSampleTags(&tags)
+	trail.SaveSamples(finalTags)
+	if t.responseCallback != nil {
+		trail.Failed.SetValid(true)
+		if failed == 1 {
+			trail.Failed.Bool = true
+		}
+		trail.Samples = append(trail.Samples,
+			stats.Sample{
```
na--, line 195: I worry a bit that this will cause a slice expansion every time, given that we have expected response and response callbacks enabled by default. So it may be worth it to initialize the `trail.Samples` slice with a cap +1 in `SaveSamples()`?

stats/cloud/data.go
```diff
@@ -110,6 +110,29 @@ type SampleDataMap struct {
 // NewSampleFromTrail just creates a ready-to-send Sample instance
 // directly from a httpext.Trail.
 func NewSampleFromTrail(trail *httpext.Trail) *Sample {
+	if trail.Failed.Valid {
+		return &Sample{
```
na--, line 114: Instead of copy-pasting the whole `&Sample{}` definition based on `trail.Failed`, can you please only change the `Values` map?

na-- (changes requested):
LGTM in general, left some more nitpicks as inline comments.

stats/cloud/data.go
MStoykov: I specifically did this so I don't expand every time as you pointed out in https://github.com/loadimpact/k6/pull/1856#discussion_r584496342

lib/netext/httpext/transport.go
MStoykov: yeah ... I wondered about that, and whether to do with a ... boolean/int flag saying to add 1/the number :thinking:

stats/cloud/data.go
na--: Yeah, I get that, but you can only copy the `Values` map init, not the whole `Sample` struct

js/internal/modules/modules.go
```diff
@@ -35,7 +35,17 @@ var (
...
+	mod := modules[name]
+	if i, ok := mod.(PerTestModule); ok {
+		return i.NewVUModule()
+	}
+	return mod
+}
+
+// PerTestModule is a simple interface representing a per test module object that can be used to
+// make a per VU instance of the Module
+type PerTestModule interface {
```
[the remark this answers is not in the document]
MStoykov: The idea is that this interface should be the per-test instance and it generates a per VU module. Which it currently does. I do agree it is currently mostly per instance one, but making it truly per test will take more refactoring of the modules so I am not making this in this PR. So I would argue the name is pretty accurate both in what it can be used (in production code) and in what it will stay in the future(even though then there will be a one more level).

The idea is that I will actually continue with #1858  and add the code so that all of that is implemented, but doing it in this PR seems unnecessary and this is all in the internal module, so nobody can use it beside us (mostly xk6 I guess ;)).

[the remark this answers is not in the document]
na--: I get the ideas and I still maintain this current name is misleading. My suggestions above aren't great, but are definitely less confusing. And if not them, how about `HasPerVUModules`?

btw do we really need a name for an interface that has per-instance modules? Wouldn't that be covered by just having a different constructor? For example, this code: https://github.com/loadimpact/k6/blob/82cab8b693963659d5f0cfb4cbf785c792577a0f/js/modules/k6/data/data.go#L36-L38

can be something like this:
```go
modules.Register("k6/data", &data{
    sharedObjects: common.NewSharedObjects(),
    // ... whatever other per-instance state we want
}) 
```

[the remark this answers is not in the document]
MStoykov: I am for just leaving it for the PR I have already started and will probably publish next week or so. 

The answer the other constructor variant:
yes ... but only if you don't want for you to be able to juts `import SharedArray from "k6/data"` in the js packages without needing to work with `data` (which will need to be exported) from the `k6/data` module directly. 

So as always , yes we can remove some abstraction, but we have to to then know how to make it  work everywhere. And I would argue this will be even more useful for `xk6` extensions as ... I technically could've implemented this with one more `SharedArray` hack like this https://github.com/loadimpact/k6/blob/82cab8b693963659d5f0cfb4cbf785c792577a0f/js/common/initenv.go#L70-L73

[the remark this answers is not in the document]
na--: > I am for just leaving it for the PR I have already started and will probably publish next week or so.

:confused: By then v0.31.0 will be released, so why should this have a misleading name until then when the fix is just to rename it?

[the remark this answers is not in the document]
MStoykov: but your renames are ... more inaccurate ... 
p.s. this is why I wanted to leave it unnamed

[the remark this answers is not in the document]
MStoykov: cc @imiric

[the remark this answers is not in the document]
MStoykov: Again the current interface is meant for a *per test* (which currently is per instance mostly, but there is no difference in production code so :shrug: ) that has a way to create a *per vu* module. 

So if anything the name should `PerInstanceModule` not `PerTestModule`, but it is definitely not *per VU* ...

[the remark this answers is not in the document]
na--: > Again the current interface is meant for a per test (...) that has a way to create a per vu module.

Can I have a per-test module that doesn't implement `PerTestModule`? Of course. Then `PerTestModule` is not a good name for the interface, is it?

Leaving interfaces unnamed is just postponing the problem where we apparently don't agree on naming things, which probably means we don't think of this in the same terms, which probably means it's a good idea to discuss them now and reach a consensus...

> but it is definitely not per VU
That's why I suggested `HasPerVUModules` as a name of the interface. It signals that the parent object, which is used in `modules.Register()`, has a unique instance of the module unique in every VU. Here's another suggestion:
```go
// HasModuleInstancePerVU should be implemented by all native Golang modules that 
// would require per-VU state. k6 will call their NewModuleInstance() methods
// every time a VU imports the module and use its result as the returned object.
type HasModuleInstancePerVU interface {
    NewModuleInstance() interface{}
}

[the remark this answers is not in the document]
Comment (MStoykov) [reply js/internal/modules/modules.go:47]: the point of this code is to make the current responseCallback possible without:
1. one more global but actually per module shared state store
2. not rewriting all the modules to do it.

So it is only checking for a specific case and that will be changed and *all* modules will just implement the interface as that takes 5 lines and removes the wondering of whetether or not your module is correct or not. And then I would expect that we will just make it required for xk6 modules as well after some time of just logging that the should be implementing the new interface ... 

For the purposes of actually merging this code .. .I am fine with w/e I just don't think that a temporary name for a temporary interface in a an *internal* module, matters enough for a design committee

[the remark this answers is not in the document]
Comment (na--) [reply js/internal/modules/modules.go:47]: It matters because I am the one reviewing the code and the current name doesn't make sense to me. The fact that we still disagree about this means that either the name is bad, or I am completely missing something (which, unless I am a complete idiot, still means something in the code can be better and more obvious). 

In any case, even an _internal_ name has to make sense and describe accurately the thing it's naming... I _think_ I understand how, what and why and  you're trying to achieve here, but I might be wrong, so I'll take a pause and re-read through this whole discussion and https://github.com/loadimpact/k6/issues/1858 later today.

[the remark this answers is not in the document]
Comment (imiric) [reply js/internal/modules/modules.go:47]: I agree that the current `PerTestModule` name is slightly confusing. Considering that the function is called `NewVUModule()`, `HasPerVUModule` makes sense to me.

In any case, it's an internal interface we can always rename later, so don't think we should spend a lot of time on this.

File: stats/cloud/data.go [hunk 25]
```diff
@@ -173,6 +193,25 @@ func (sdagg *SampleDataAggregatedHTTPReqs) CalcAverages() {
 	sdagg.Values.Receiving.Calc(count)
 }
 
+// AggregatedRate is an aggregation of a Rate metric
+type AggregatedRate struct {
+	Count   float64 `json:"count"`
+	NzCount float64 `json:"nz_count"`
```
imiric, line 199: What does `Nz` stand for in this context?

imiric (dismissed):

js/internal/modules/modules.go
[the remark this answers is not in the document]
na--: I've laid out my arguments in https://github.com/loadimpact/k6/issues/1858#issuecomment-788867806 how I think this should be evolved in k6 v0.32.0 and why. Having thought about this a lot more and re-read the discussions here and in the issue, I still maintain that this is a better name than the current interface:
```go
type HasModuleInstancePerVU interface {
    NewModuleInstance() interface{}
}
```
Maybe even this, if we want to be super-descriptive:
```go
type HasModuleInstancePerVU interface {
}
```

lib/netext/httpext/tracer.go
```diff
@@ -61,17 +63,17 @@ type Trail struct {
 // SaveSamples populates the Trail's sample slice so they're accesible via GetSamples()
 func (tr *Trail) SaveSamples(tags *stats.SampleTags) {
 	tr.Tags = tags
-	tr.Samples = []stats.Sample{
+	tr.Samples = make([]stats.Sample, 0, 9) // this is with 1 more for a possible HTTPReqFailed
+	tr.Samples = append(tr.Samples, []stats.Sample{
```
na--, line 67: I don't remember the benchmarks you posted in slack, but does making this a fixed-length array instead of a slice help? 
```suggestion
	tr.Samples = append(tr.Samples, [...]stats.Sample{
```

na-- (commented):
LGTM besides the interface name.

lib/netext/httpext/tracer.go
MStoykov: It gives 
```
./tracer.go:67:21: cannot use [...]stats.Sample{...} (type [8]stats.Sample) as type []stats.Sample in append
```
if I add `[:]`
```
./tracer.go:76:3: invalid operation [8]stats.Sample{...}[:] (slice of unaddressable value)
```
I doubt that will make any difference either way ... as I am pretty sure there are enough optimizations specifically for this case ... so it likely doesn't even allocate a slice, to begin with.

The memprofile shows that ... yes SaveSamples is the main (40%) memory used generator, and the only thing that allocates on the heap is the `make`.

As I said in slack if I add 1 more argument to say the slice needs to be bigger the difference is
```
name                                         old time/op    new time/op    delta
MeasureAndEmitMetrics/no_responseCallback-8    6.82µs ± 2%    6.70µs ± 3%    ~     (p=0.310 n=5+5)
MeasureAndEmitMetrics/responseCallback-8       7.16µs ± 3%    6.97µs ± 3%    ~     (p=0.095 n=5+5)
name                                         old alloc/op   new alloc/op   delta
MeasureAndEmitMetrics/no_responseCallback-8    1.10kB ± 0%    1.03kB ± 0%  -5.84%  (p=0.008 n=5+5)
MeasureAndEmitMetrics/responseCallback-8       1.10kB ± 0%    1.10kB ± 0%    ~     (all equal)
name                                         old allocs/op  new allocs/op  delta
MeasureAndEmitMetrics/no_responseCallback-8      9.00 ± 0%      9.00 ± 0%    ~     (all equal)
MeasureAndEmitMetrics/responseCallback-8         9.00 ± 0%      9.00 ± 0%    ~     (all equal)
```
regardless of whether with a single boolean or with an integer to be added to the length. I can add this as it does seem helpful ... but not by much

stats/cloud/data.go
MStoykov: done

na-- (approved):

imiric (approved):

PR #1857, opened 2021-02-15
Move common cloud APIs to a separate package

na--:
This makes the actual cloud output leaner, leaving it concerned only with metric emission to the cloud. I'll probably split the config in future PRs, since a big part of the config deals only with the cloud output, and also to hopefully make https://github.com/loadimpact/k6/issues/1155 easier.


cloudapi/config.go
```diff
@@ -239,3 +240,25 @@ func (c Config) Apply(cfg Config) Config {
 	}
 	return c
 }
+
+// MergeFromExternal merges three fields from json in a loadimact key of the provided external map
```
MStoykov, line 244:
```suggestion
// MergeFromExternal merges three fields from json in a loadimpact key of the provided external map
```

MStoykov (dismissed):
LGTM! 

I hope we don't get some strange breakage of the cloud config again :crossed_fingers:

imiric (commented):
I like the split, but have a problem with `cloudapi` being a new root-level package. Couldn't it be moved to `/stats/cloud/api` instead?

na--:
@imiric I don't think that's a good idea - `/stats/cloud` (future `/output/cloud` or something like that) still has all of the code that has to deal with the cloud output and metric aggregation, while `cloudapi` has precisely everything else - configuration, test creation and query APIs, log streaming logic, etc. So it makes no sense to try and force the non-output stuff as a sub-package of the output.

Whether a top-level package is ideal is a separate discussion... Unfortunately, we don't seem to have any other suitable place for it besides `/lib`, and we want to avoid that, so it ended up being a top-level package...

imiric (approved):

PR #1861, opened 2021-02-17
Run babel on compilation errors as well as on parsing ones

MStoykov:
Unfortunately parsing checks syntax, not whether what the syntax did was
actually ... supported(sometimes).

As can be seen this fixes a lot of cases where this is a problem.

This is actually probably going to be an even bigger problem in the
future as goja gets more features in it which will let the parsing
to pass even though some feature actually isn't supported but the syntax
is now "valid".

Good example (which was how this was kind of found) is that destructuring
looks a lot like an object with shorthand properties, but this is only
noticable once the code is compiled, not while parsing. So in the future
when goja for example has support for all the rest of the syntax (like
`let`) it will be harder to catch.

Even now code such as `let [x] = [2]` which is desctructuring doesn't
return an error on the parsing stage but instead in the compilation.
This example might be a bug in goja, but I would argue we can have
others.


js/tc39/breaking_test_errors.json
```diff
@@ -3186,6 +3104,7 @@
   "test/language/literals/regexp/S7.8.5_A1.4_T2.js-strict:true": "test/language/literals/regexp/S7.8.5_A1.4_T2.js: Test262Error: Code unit: d800 Expected SameValue(«\\\\\\ud800», «\\�») to be true at harness/sta.js:22:9(49)",
   "test/language/literals/regexp/S7.8.5_A2.1_T2.js-strict:true": "test/language/literals/regexp/S7.8.5_A2.1_T2.js: Test262Error: Code unit: d800 Expected SameValue(«nnnn\\ud800», «nnnn�») to be true at harness/sta.js:22:9(49)",
   "test/language/literals/regexp/S7.8.5_A2.4_T2.js-strict:true": "test/language/literals/regexp/S7.8.5_A2.4_T2.js: Test262Error: Code unit: d800 Expected SameValue(«a\\\\\\ud800», «a\\�») to be true at harness/sta.js:22:9(49)",
+  "test/language/literals/regexp/early-err-flags-unicode-escape.js-strict:true": "test/language/literals/regexp/early-err-flags-unicode-escape.js: error is not an object (Test262: This statement should not be evaluated.)",
```
MStoykov, line 3107: This is https://github.com/babel/babel/issues/6691 and practically ... not a problem for us.

js/tc39/breaking_test_errors.json
```diff
@@ -4724,16 +4643,15 @@
   "test/language/statements/for-in/head-const-bound-names-dup.js-strict:true": "test/language/statements/for-in/head-const-bound-names-dup.js: unexpected error type (TypeError), expected (SyntaxError)",
   "test/language/statements/for-in/head-const-bound-names-fordecl-tdz.js-strict:true": "test/language/statements/for-in/head-const-bound-names-fordecl-tdz.js: Test262Error: Expected a ReferenceError to be thrown but no exception was thrown at all at harness/sta.js:22:9(49)",
   "test/language/statements/for-in/head-const-bound-names-in-stmt.js-strict:true": "test/language/statements/for-in/head-const-bound-names-in-stmt.js: error is not an object (Test262: This statement should not be evaluated.)",
+  "test/language/statements/for-in/head-let-bound-names-dup.js-strict:true": "test/language/statements/for-in/head-let-bound-names-dup.js: unexpected error type (TypeError), expected (SyntaxError)",
```
MStoykov, line 4646: This actually now returns the "correcter":
```
ERRO[0000] TypeError: file:///home/mstoykov/work/k6_copy/js/tc39/d.js: Duplicate declaration "x"
> 1 | for (let [x, x] in {}) {}
    |              ^
  2 |  at <eval>:2:28542(114)
```
instead of 
```
SyntaxError: Unexpected strict mode reserved word at file:///home/mstoykov/work/k6_copy/js/tc39/d.js:1:6
```
The error comes from babel transformation so :shrug:

js/tc39/breaking_test_errors.json
```diff
@@ -4907,9 +4808,9 @@
   "test/language/statements/for-of/head-const-bound-names-in-stmt.js-strict:true": "test/language/statements/for-of/head-const-bound-names-in-stmt.js: error is not an object (Test262: This statement should not be evaluated.)",
   "test/language/statements/for-of/head-decl-no-expr.js-strict:true": "test/language/statements/for-of/head-decl-no-expr.js: error is not an object (Test262: This statement should not be evaluated.)",
   "test/language/statements/for-of/head-expr-no-expr.js-strict:true": "test/language/statements/for-of/head-expr-no-expr.js: error is not an object (Test262: This statement should not be evaluated.)",
+  "test/language/statements/for-of/head-let-bound-names-dup.js-strict:true": "test/language/statements/for-of/head-let-bound-names-dup.js: unexpected error type (TypeError), expected (SyntaxError)",
```
MStoykov, line 4811: https://github.com/loadimpact/k6/pull/1861/files#r577493679

na-- (approved):
Even if you don't refactor and remove the duplication, :+1: LGTM as it is, since it's clear what is happening. Though maybe add a comment with the explanation _why_ we do it from the PR description?

js/compiler/compiler.go
```diff
@@ -119,6 +119,39 @@ func (c *Compiler) Compile(src, filename, pre, post string,
...
+	// when goja for example has support for all the rest of the syntax (like
+	// let) it will be harder to catch.
+	//
+	// Even now code such as `let [x] = [2]` which is desctructuring doesn't
+	// return an error on the parsing stage but instead in the compilation.
+	// This example might be a bug in goja, but I would argue we can have
+	// others.
+	//
+	// So because of this if there is an error during compilation it is still worth it to transform
+	// and try again just in case it fixes it, which it does for some cases.
```
na--, line 143: I didn't mean copy the whole PR description...
```suggestion
	// Parsing only checks the syntax, not whether what the syntax expresses
	// is actually supported (sometimes).
	// 
	// For example, destructuring looks a lot like an object with shorthand
	// properties, but this is only noticeable once the code is compiled, not 
	// while parsing. Even now code such as `let [x] = [2]` doesn't return an
	// error on the parsing stage but instead in the compilation.
	//
	// So, because of this, if there is an error during compilation, it still might 
	// be worth it to transform the code and try again.
```

na-- (changes requested):

imiric (dismissed):
I'd prefer if we could get rid of this "try Babel if parsing/compiling failed" approach, but glad to see it got rid of so many errors. Could we avoid the duplication though?

MStoykov:
> I'd prefer if we could get rid of this "try Babel if parsing/compiling failed" approach, 

That is dependant on goja getting support for more or less everything babel currently provides us ... which will take quite some time, especially given that it is mostly 1 person working on it in their spare time. My contributions are unlikely to make a dent in that unless we prioritize it and that is actually unlikely as well as there is not so much benefit to that as for example working on things that babel doesn't provide us with (generators, Promise, async/await being the interesting ones).

> Could we avoid the duplication though?

Looking at the code ... it's unlikely any refactoring will make it clearer and this code is rarely touched ... so hopefully whoever do touch it will notice the same block a comment apart :)

imiric (approved):

na-- (approved):

Issue #1320, opened 2020-01-28
Improve execution information in scripts


na--:
Just realized that, with execution segments (#997), we can easily make the `__VU` constants much more useful when we're executing scripts in the cloud or in the future native k6 distributed execution. 

Currently, the `__VU` variable will start from `1` in each instance, so for multiple instances there would be duplicate values. With execution segments, we can very easily make each instance starts its VU numbers from the exact VU id that it should, so that each `__VU` value is globally unique, regardless of how many machines we run the load test on. And while I think this should be the default, I can sort of see a use case where we'd like to use the local machine sequential number, not the global one, so we should probably expose both...

This ties neatly into another topic we should look into, probably after merging #1007 - exposing more execution information to user scripts. Though, instead of magic `__WHATEVER` constants, I think we should do it by exposing a nice JS API that queries these things from the k6 engine/execution scheduler/executors/etc. This would change the model from a push-based one (i.e. k6 having to update all `__WHATEVER` variables every time they're changed), to a pull based one (a function would query the already existing data in a thread-safe way), which is much easier to support and way, way more efficient (i.e. basically no performance overhead if users don't care for the information).

So, something like this:
```js
import { getLocalVUNumber, getGlobalVUNumber, getStartTime } from "k6/execution";

//

vuNum := getGlobalVUNumber()

```

:arrow_up: is definitely NOT the final API, since it probably makes sense to expose the logical execution-related entities (i.e. "Instance", "VU", "Iteration", "Executor") in some way instead of just having global functions, I'm just using it as an illustration...

We probably should expose things like:
- local and global VU numbers
- iteration numbers:
  - have a method to duplicate `__ITER`, i.e. "which iteration IN this VU is currently running"
  - but since VUs can be reused, some other interesting numbers would be "iteration number for the VU in the current executor"
  - for the arrival-rate executors, in particular, it might also be useful to expose (and we can easily do it) the global number of that iteration across all instances
- when did the script execution begin and/or what's the total execution time up until now
- when did the current executor start and how long has it been running
- which executor is currently responsible for running the code 
- for executors with `stages`, which stage are we currently at, maybe even the "progress" percent?
- execution segment related things (though that probably should be a separate API that deals with data partitioning/streming/etc. ?)
- for the current instance, the number of active/initialized VUs and the number of complete/interrupted iterations up until now (can't think of a use case for this, besides maybe something in `teardown()`, but we can easily expose it, since we keep track of it for the CLI interface)

[...]

Issue #1539, opened 2020-07-08
Data segmentation API framework


na--:
We now have support for partitioning work (i.e. VUs and iterations) between multiple k6 instances, via the `executionSegment` and `executionSegmentSequence` options, originally described in https://github.com/loadimpact/k6/issues/997 and subsequently evolved in https://github.com/loadimpact/k6/pull/1007. In the end, we had to implement [striping](https://github.com/loadimpact/k6/issues/997#issuecomment-493397996) even in the initial version (thus, the need for `executionSegmentSequence` :disappointed:), because some executors like `ramping-vus` and the `arrival-rate` ones needed it for optimal performance.

The good news from that extra effort though, is that we now have most of the things we need to tackle data segmentation/partitioning between multiple k6 instances, without any external runtime scheduling between them. We need to refactor and improve some things in the initial implementation, like https://github.com/loadimpact/k6/issues/1499, https://github.com/loadimpact/k6/issues/1427, and https://github.com/loadimpact/k6/issues/1386, but the rough building blocks are already here...  :tada:  I'm making this issue as a place to discuss this effort, so that I can close https://github.com/loadimpact/k6/issues/997, given that most of it is done.  

Of course, we don't need to start implementing this right away. We "just" have to figure out how its JS API and options should look like... :sweat_smile: This will allow us to start implementing things like a streaming data support (https://github.com/loadimpact/k6/issues/592), shared read-only memory (https://github.com/loadimpact/k6/issues/532), CSV API (https://github.com/loadimpact/k6/issues/1021), a JSONPath API (https://github.com/loadimpact/k6/issues/992), and an XML parsing API with XPath support. I don't think binary data handling (https://github.com/loadimpact/k6/issues/1020) is going to be affected by this, but it probably deserves some thought as well.

If we have a clear idea how the data segmentation, we can start implementing the issues above without the complicated data segmentation in their first versions, knowing we'd be able to add it at a later point, hopefully without having to completely refactor everything again. Currently, I think we can split the process like this:
1. Figure out how how data segmentation should look like (this issue)
2. Start implementing initial versions, without segmentation, of a streaming/shared data, CSV API, JSONPath, etc.
[...]

Issue #1001, opened 2019-04-17
Provide an API to exit a test


ppcano:
The [k6 API ](https://docs.k6.io/docs) does not provide an API to exit a load test.  The [fail](https://docs.k6.io/docs/fail-err) does not cover this case.

A common workaround is to configure a [threshold that aborts the test](https://docs.k6.io/docs/thresholds#section-aborting-a-test-when-a-threshold-is-crossed) when the threshold is crossed.

```js
import { sleep } from "k6";
import http from "k6/http";
import { Counter } from "k6/metrics";

export let options = {
    stages: [
        { duration: "30s", target: 50 }
    ],
    thresholds: {
        "myThresholdMetric": [{threshold: "count<1", abortOnFail: true}]
    }
};

const myThresholdMetric = new Counter("myThresholdMetric");

export default function() {
    http.get("https://test.loadimpact.com/");
    if (__VU === 1 && __ITER === 1) { 
        myThresholdMetric.add(1);
    }
    sleep(3);
};
```

But this or other workarounds have a few inconveniences or limitations:

1. The [Threshold](https://docs.k6.io/docs/thresholds) API works well for evaluating complex conditions but the API is not intuitive or simple when you want just to check a simple condition. 

```js
    if (condition) { 
        exit();
    }
```
2. It cannot `exit` the load test on a non-error mode. 

```js
exit(0);
```

3. It cannot provide different `exit` codes. Related issue: [Different exit codes for different errors](https://github.com/loadimpact/k6/issues/870)

```js
exit(5);
```

4. it cannot provide user message

```js
exit(10, “service was not available”);
```

5. The API should also work on the `setup/teardown` methods.

```js
export function teardown(time) {
    if (condition) { 
        exit(1);
    }
}
```

The `exit` API is a suggestion, but in any case, I think it could be a good addition if k6 provides an API to support these cases during local and cloud execution.  Some users have already asked for a similar utility.

na--:
This is a partial duplicate of an [older issue](https://github.com/loadimpact/k6/issues/939), but I'll close the old one and leave this one open, since this is more detailed and somewhat easier to find. Just a couple of comments on the proposed API:
- I prefer `abort()` as the name of the new function, as proposed in the other issue. It's makes it more clear that we're aborting the whole test. `exit()` also has connotations of immediate termination, while k6 actually has to perform some housekeeping when it aborts tests, like sending latest metrics to collectors, etc.
[...]

PR #1863, opened 2021-02-17
Core changes for `k6/x/execution` JS module

imiric:
These are core changes needed for the [`k6/x/execution` JS module](https://github.com/k6io/xk6-execution).

Initially the module was part of core as well, but we decided to extract it into an extension to be able to quickly iterate on the API. See the JS usage and example in the extension repo.

Part of #1320


js/runner_test.go
```diff
@@ -2001,3 +2001,98 @@ func TestComplicatedFileImportsForGRPC(t *testing.T) {
...
+		`, "scenario information can only be returned from an exported function"},
+		{"test_ok", `
+		var exec = require('k6/execution');
+
+		exports.default = function() {
+			var ts = exec.getTestStats();
+			if (ts.vus_active !== 1) throw new Error('unexpected vus_active: '+ts.vus_active);
+			if (ts.vus_max !== 0) throw new Error('unexpected vus_max: '+ts.vus_max);
+			if (ts.iterations_completed !== 0) throw new Error('unexpected iterations_completed: '+ts.iterations_completed);
+			if (ts.iterations_interrupted !== 0) throw new Error('unexpected iterations_interrupted: '+ts.iterations_interrupted);
```
imiric, line 2049: These checks are kind of pointless... Should I move these tests to some higher-level ones that involve actual executors?

js/modules/k6/execution/execution.go
```diff
@@ -0,0 +1,94 @@
...
+		return nil, errors.New("scenario information can only be returned from an exported function")
+	}
+
+	progress, _ := ss.ProgressFn()
+	out := map[string]interface{}{
+		"name":      ss.Name,
+		"executor":  ss.Executor,
+		"startTime": ss.StartTime,
+		"progress":  progress,
+	}
```
MStoykov, line 71: I think it will be better if `progress` calls the method when accessed (or is a function) because:
1. this won't call `progressFn` unless the user actually needs it instead of needing the `name` for example
2. will allow users to monitor the progress without calling `getScenarioStats` constantly

js/modules/k6/execution/execution.go
```diff
@@ -0,0 +1,94 @@
...
+
+	return out, nil
+}
+
+// GetTestStats returns global test information.
+func (e *Execution) GetTestStats(ctx context.Context) (map[string]interface{}, error) {
+	es := lib.GetExecutionState(ctx)
+	if es == nil {
+		return nil, errors.New("test information can only be returned from an exported function")
```
MStoykov, line 80: The message is ... not worded very well and not consistent with the wording in other places where we talk about the `init context` .

MStoykov (commented):
Haven't taken a deeper look or tried the code, but some comments:
1. IMO we should allow getting the info in all cases where any info is available
2. The things that call functions should probably call the functions on the spot (without any caching) or be functions in the first place
3. Given that a lot of the `*State` objects seem "static" can we have 1 of those and just give it everywhere
4. given `3.` (and even without it) can you check and make certain that a user script can't change any of the values?

imiric:
@MStoykov:

> IMO we should allow getting the info in all cases where any info is available

Are you referring to being able to call this from init? Most of this information doesn't make sense outside the context of a running VU/scenario. Only `vus_max` seems relevant in init and we could get it from the config instead of `*ExecutionState`, but besides complicating things slightly it would be weird to have to document which information is available and which isn't, instead of just disallowing it entirely. If we go with your 2nd suggestion below, it might make sense, so I'll think about it some more. :)

> The things that call functions should probably call the functions on the spot (without any caching) or be functions in the first place

Hhmm I'm not sure I prefer the API to be something like `getTestStats().duration()` or complicating it with `DynamicObject` like we discussed in #1320. Returning an already evaluated object is easier to understand (users don't expect proxies) and these are small objects so the performance benefit would likely be negligible.

I could give it a shot if you feel strongly about it, though would prefer the `DynamicObject` approach in that case.

> Given that a lot of the `*State` objects seem "static" can we have 1 of those and just give it everywhere

Hhmm you mean about `*ScenarioState` specifically and how it's set in the context each time a VU is activated? True, it would be better if it was set on the context without being passed around in `VUActivationParams`. I'll look into it, thanks.

But `*State` and `*ExecutionState` are only set once.

> given `3.` (and even without it) can you check and make certain that a user script can't change any of the values?

Sure. With dumb / already evaluated objects this shouldn't matter, but I'll confirm after your suggested changes.

js/modules/k6/execution/execution.go
imiric: Yeah, it would be a slight performance benefit to evaluate it on access. I'll look into doing this with `DynamicObject` then.

js/modules/k6/execution/execution.go
imiric: True, I'll make it consistent.

js/modules/k6/execution/execution.go
MStoykov: I would say that a normal Object with `defineProperty` will be fine, but a `DynamicObject` will do the job as well ... just a little too much IMO

na--:
I have only skimmed the code changes, so I haven't fully grokked the PR yet, but I have a more generic comment here. I think the MVP version of https://github.com/loadimpact/k6/issues/1320 shouldn't be exposing information users can already mostly get. It should be exposing things that they cannot currently get in any way. Primarily because it would require some refactoring to the executors and may change how the rest of the feature is implemented.

From what I can see, `exec.getVUStats()` would return exactly `__VU` and `__ITER`, right? While it is _nice_ to have a better UX than the current super-global "constants" for that information, it's not _necessary_ to do so initially, users manage just fine without it. `exec.getScenarioStats()` is new information, so better, but even now you can get most of that information by just having different `exec` function (or via environment variables in each scenario). The rest of the information from `exec.getScenarioStats()` and the whole `exec.getTestStats()` is novel and previously inaccessible information, true, but probably wouldn't actually be useful to users from inside of the script... 

I think the MVP of this feature should have these two pieces of information, with everything else being a nice-to-have bonus:
- _in the current scenario_, what is the ID number of the VU currently running the iteration
- _in the current scenario_, what is the sequential number of the iteration for this VU that is currently being executed
- for `shared-iterations`, `constant-arrival-rate` and `variable-arrival-rate` scenarios, a way to get the sequential number of the current iteration among _all_ of the iterations for the current scenario, across all VUs and even instances.

The first 2 are the equivalent of `__VU` and `__ITER`, but only in terms of a single scenario. The last one is a very useful feature for data sharding and it should be easy to calculate with the execution segments. But all of these are things are information that's completely unavailable to user scripts right now. So, yeah, I may be wrong, but I think we should start with the "hard" (i.e. unknown) parts of this issue first, and implement the "easy" (i.e. known) parts last, as a polish to everything.

imiric:
@na-- OK, so to clarify, that would involve adding a few additional executor-specific counters, right? Except for the last global across instances values which could be calculated given the execution segment?

na--:
> OK, so to clarify, that would involve adding a few additional executor-specific counters, right?

Yeah, basically a local VU number for the scenario (basically immutable after the VU activation in the scenario) and a local iteration number for the specific VU in the scenario (mutable, but the only lock contention here would be between whatever increments it and this new `exec` method used to retrieve it, so pretty light). Still, as I said, those things require some refactoring of the executors... :disappointed:

imiric:
So it's unfortunately safe to say that this won't make the v0.31.0 release. :disappointed:

I think the functionality described [here](https://github.com/loadimpact/k6/pull/1863#issuecomment-782069940) is implemented now, but it will need thorough testing and actually making the tests pass consistently.

Please give the code an overview, the commits should be fairly isolated, and test it on your machine to ensure it's at least functionally OK.

What's pending:

- Fixing the flaky tests. I think I managed to find a timing sweet spot for the `*GlobalIters` tests (they passed for more than 30m with [this script](https://dave.cheney.net/2013/06/19/stress-test-your-go-packages)), but it's clear from the Windows failures that the others are still flaky. I'm not sure how to refactor them to avoid this, so any ideas are welcome.

- Addressing @MStoykov's comments above, which I started doing but ran into issues with sharing the object across goja runtimes.

- Addressing any additional feedback.

lib/executor/helpers.go
```diff
@@ -108,6 +108,7 @@ func getIterationRunner(
 
 			// TODO: move emission of end-of-iteration metrics here?
 			executionState.AddFullIterations(1)
+			incrIter()
```
imiric, line 111: We're currently only tracking completed/full iterations. Would tracking interrupted iterations too be interesting to users? It would essentially double the amount of counters for everything... :-/

js/modules/k6/execution/execution.go
imiric: Resolved by c034057e?

js/modules/k6/execution/execution.go
imiric: Resolved by 327a370a.

imiric:
@MStoykov @na-- Could you take another look and let me know if there's anything pending here?

From what we discussed, allowing the use of this module in the init context is the major thing still missing. After taking another look, I still think it would require substantial refactoring to achieve, and most of the information just doesn't make sense in the init context.

The big (only?) use case that does make sense is retrieving the max test duration before the test starts, as mentioned on the [forum](https://community.k6.io/t/get-total-test-duration/1567/2) and [here](https://github.com/k6io/k6/issues/1320#issuecomment-812463418). The problem with this is that the init context is evaluated before the `ExecutionScheduler`/`ExecutionState` is created.
This is a chicken/egg situation since we need to have access to the consolidated options, including the ones exported in JS, to be able to return the correct duration, but these aren't consolidated until after the init context is evaluated for the first time (see `cmd/run.go`).
I considered adding `lib.Options` to `common.InitEnvironment` to precalculate the max duration/VUs and avoid depending on `ExecutionScheduler`, but we run into the same chicken/egg thing, since options aren't consolidated when `InitEnvironment` is first initialized.

A possible approach would be to offer a pure JS API to calculate and return just the max duration, but that would duplicate a lot of the same logic we have in Go, and it doesn't seem worth it considering it can be done with the current implementation from the `default` function, and once we have #1001 to abort the test, it would fulfill the requirement of the person on the forum.

lib/execution_segment.go
```diff
@@ -730,3 +731,96 @@ func (et *ExecutionTuple) GetNewExecutionTupleFromValue(value int64) (*Execution
...
+	if s.scaled == 1 { // we are the first need to go to the 0th element which means we need to remove the start
+		s.unscaled -= s.start + 1 // this could've been just settign to 0
+	} else { // not at the first element - need to get the previously added offset so
+		s.unscaled -= s.offsets[int(s.scaled-2)%len(s.offsets)] // slice's index start 0 our start at 1
+	}
+	s.scaled--
+}
+
+// GoTo sets the scaled index to its biggest value for which the corresponding
+// unscaled index is is smaller or equal to value.
```
MStoykov, line 775:
```suggestion
// unscaled index is smaller or equal to value.
```

lib/execution_segment.go
```diff
@@ -730,3 +731,96 @@ func (et *ExecutionTuple) GetNewExecutionTupleFromValue(value int64) (*Execution
 		SegmentIndex: newIndex,
 	}, nil
 }
+
+// SegmentedIndex ...
```
MStoykov, line 735: maybe write something here ;)

imiric: :sweat_smile: I was hoping one of you guys could come up with a useful description for this, I wasn't sure how to document it.

lib/executor/constant_arrival_rate.go
```diff
@@ -196,9 +199,24 @@ func (car *ConstantArrivalRate) Init(ctx context.Context) error {
...
+	car.segIdx = lib.NewSegmentedIndex(start, lcd, offsets)
+
 	return err
 }
 
+// incrGlobalIter increments the global iteration count for this executor,
+// taking into account the configured execution segment.
+func (car *ConstantArrivalRate) incrGlobalIter() {
+	car.segIdx.Next()
+	atomic.StoreUint64(car.globalIter, uint64(car.segIdx.GetUnscaled()))
```
MStoykov, line 212: This is logically racy as there is nothing preventing 2 goroutines to call `car.segIdx.Next()` and then `atomic.StoreUint64` and the one who got the "first" values from `segIdx.Next` to store it second, effectively having the previous value in `car.globalIter`.

Same for the `shared-iterations`

js/runner_test.go
```diff
@@ -2006,3 +2006,101 @@ func TestComplicatedFileImportsForGRPC(t *testing.T) {
...
 	}
 }
+
+func TestExecutionStats(t *testing.T) {
+	t.Parallel()
+
+	testCases := []struct {
+		name, script, expErr string
+	}{
+		{"vu_ok", `
```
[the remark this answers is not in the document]
imiric: Not sure I understand the drawbacks, but I don't like using maps for test cases because of unordered test runs. I know it shouldn't matter and tests shouldn't rely on the order, but as long as they don't I prefer knowing the order they start in.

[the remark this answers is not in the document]
MStoykov: > I don't like using maps for test cases because of unordered test runs.

I am not saying to use maps but use `{name: "vu_ok", script: "some javascript"}`

>I know it shouldn't matter and tests shouldn't rely on the order, but as long as they don't I prefer knowing the order they start in.

Also https://go-review.googlesource.com/c/go/+/310033/ and .. if it doesn't matter it's better to be random so at least it continues to not matter ;)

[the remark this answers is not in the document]
MStoykov: Also `t.Parallel` in practice should mean that they are already somewhat randomized, the proposed `-shuffle` shuffles top-level tests only from the looks of it ;)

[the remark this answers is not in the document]
imiric: > I am not saying to use maps but use `{name: "vu_ok", script: "some javascript"}`

Ah, I see what you mean now :) Sure, I'll change it.

lib/executor/constant_arrival_rate.go
imiric: Let me know if it's as simple as 8973a841.

I wanted to remove the use of `atomic`, but we still need it because `globalIter` needs to remain a pointer since `incrGlobalIter` is passed to the VU `State` which is where it's being called (this is part of the refactor to fix how `globalIter` works).

js/runner_test.go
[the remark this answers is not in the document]
imiric: Resolved by 8b24539a.

imiric:
@MStoykov Let me know if the recent refactor, specifically 69e67506, addresses your feedback about how scenario iterations work. They're now being incremented and set in the VU `State`, which keeps them stable within a single iteration. I also moved the regular VU iteration increment to the same place, and now `__ITER` is no longer set for `setup()` and `teardown()`.

Can you please confirm with the test script you posted on Slack?

There are still a couple of tests pending (test ideas are welcome!) that I'll add later today, and the existing ones are possibly flaky (especially `TestSharedIterationsGlobalIters`) but let me know if there's something glaringly wrong with the current state.

lib/state.go
```diff
@@ -67,8 +67,22 @@ type State struct {
...
+	IncrScIterGlobal     func() int64
+	scenarioVUIter       map[string]int64
+	scIter, scIterGlobal int64
+}
+
+// Init initializes some private state fields.
+func (s *State) Init() {
+	s.scenarioVUID = make(map[string]uint64)
+	s.scenarioVUIter = make(map[string]int64)
+	s.iteration, s.scIter, s.scIterGlobal = -1, -1, -1
```
imiric, line 85: This is... nasty. I'd much prefer it if we could keep these as `uint64`, but since the increment needs to happen before `runFn()`, this seemed like the path of least resistance that doesn't involve a lot of additional checks.

Any better ideas?

lib/state.go
```diff
@@ -79,3 +95,67 @@ func (s *State) CloneTags() map[string]string {
...
+	} else {
+		s.scenarioVUIter[s.ScenarioName] = 0
+	}
+	if s.IncrScIter != nil {
+		s.scIter = s.IncrScIter()
+	}
+	if s.IncrScIterGlobal != nil {
+		s.scIterGlobal = s.IncrScIterGlobal()
+	}
+}
```
MStoykov, line 136: This is also logically racy as there is nothing stopping a goroutine going to `IncrScIter` and then before it gets to IncrScIterGlobal 20 others to get to it which leads to this 
```

INFO[0006] 8 9 6 1                                       source=console
INFO[0006] 10 11 5 1                                     source=console
INFO[0006] 12 13 2 1                                     source=console
INFO[0006] 14 15 5 2                                     source=console
INFO[0006] 18 19 5 3                                     source=console
INFO[0006] 17 18 11 0                                    source=console
INFO[0006] 11 12 1 1                                     source=console
INFO[0006] 16 17 4 1                                     source=console
...
```
from 
```
    var stats = execution.getScenarioStats()
    if (stats.iterationGlobal != stats.iteration) {
    console.log(stats.iterationGlobal, stats.iteration, __VU, __ITER)
    }
```

MStoykov: This seems like a bad idea idea either way as it lets anything having access to `lib.State` to update the iteration (which was also possible before to be honest, but I still don't like it)
IMO :
1.  `s.iteration` should be part of  InitializedVU 
2. scIter and globalIter be part of ActivatedVU 

And both to somehow get queried from there, but to be update from their respected places *when* an iteration gets started.

Given that the whole iterations things are mostly a roundabout way to do #1539 and that something like [xk6-segment](https://github.com/MStoykov/xk6-segment) will be way more useful (not requiring to have only 1 item per iteration) I am fine with all the iteration counters to be dropped at this point as they seem to be problematic and we can get  1 PR without any of them and then you to do another with what we can do as long as both you and @na-- agree.

MStoykov (commented):
I tried to do partly the proposed changes, but hit some problems :( so I don't know how reasonable it is.

Also I noticed that a linter was complaining that `SharedIterations.Run` is copying a lock/mutex due to the method being with a non pointer receiver. So if we are going to have lock/mutexes embedded we need all methods (which will be using it but probably better to just be *all*) to be with pointer receivers.

I haven't really looked at most of the other code as those problems seemed big enough to me to look into :(

lib/state.go
imiric: Hhmm you're right. I don't know what I was thinking with the `State.scIterMx` mutex, it wouldn't work across VUs :man_facepalming:

I see your point about incrementing iterations from `State` being problematic, since it's part of the context and widely shared. Moving the fields to `InitializedVU`/`ActivatedVU` does make sense.

I'll look into fixing this, but I wouldn't want to drop anything at this point, since we've delayed this feature long enough. Let's discuss it later this week.

lib/state.go
```diff
@@ -67,8 +68,23 @@ type State struct {
...
+	IncrScIter           func() int64
+	IncrScIterGlobal     func() int64
+	scIterMx             sync.RWMutex
+	scenarioVUIter       map[string]int64
+	scIter, scIterGlobal int64
+}
+
+// Init initializes some private state fields.
+func (s *State) Init() {
+	s.scenarioVUID = make(map[string]uint64)
```
MStoykov, line 85: This seems to me like it can be not a map ;)

as in reality it gets set whenever you activate the VU and then we get it through using the scenarioName so it can just as well be a single uint64

lib/state.go
```diff
@@ -67,8 +68,23 @@ type State struct {
...
+	s.scenarioVUIter = make(map[string]int64)
```
MStoykov, line 86: I see why this is map ... I do think it is needed in the current way things are done, but I hope you can drop it with the proposed refactoring as I truly dislike it :sweat_smile:

lib/state.go
MStoykov: I am mostly proposing that we scale this back to without the iterations so we can merge it more easily and then you to make another PR ... with just the more ... involved iteration changes

lib/state.go
MStoykov: I guess they can be `^0` and adding 1 will overflow them ? does that have some problems 
https://play.golang.org/p/71mrOn9ATxg
you need the cast as otherwise the magical constant magic in go doesn't translate it to uint64 before `^` and instead it things it is the int `-1` :facepalm:

js/modules/k6/execution/execution.go
```diff
@@ -0,0 +1,170 @@
...
+	if rt == nil {
+		return nil, errors.New("goja runtime is nil in context")
+	}
+
+	stats := map[string]interface{}{
+		// XXX: For consistency, should this be startTime instead, or startTime
+		// in ScenarioStats be converted to duration?
+		"duration": func() goja.Value {
+			return rt.ToValue(es.GetCurrentTestRunDuration().String())
+		},
```
MStoykov, line 129: I think this might be better off as `progress` although I also would expect people would want the max duration of the test which might be put here as `duration` ;)

MStoykov (commented):
I left some comments while doing a second pass on mostly not the iteration stuff:
LGTM but I need to either see it without the iteration changes or with the fixed iterations changes for better comments :(, but I don't see some big problems with it :+1:

lib/state.go
imiric: @MStoykov Please take a look at 9cc9de4b. There are some wonky things about it still, like the setting of lambdas on `lib.State` that are passed via `VUActivationParams`, but I think this sync issue should be fixed now.

We can only retrieve iterations via `State` now, which I needed to expose to `k6/execution`, but the increment is done internally in `js.ActiveVU`.

imiric:
@MStoykov

> Also I noticed that a linter was complaining that SharedIterations.Run is copying a lock/mutex due to the method being with a non pointer receiver.

Which linter? I haven't noticed this. I think there would be failures in the `*GlobalIters` tests if the `iterMx` mutex is not locking properly (i.e. each VU got a different mutex instance). Since `getNextGlobalIter` is defined on a pointer, I would assume the same instance is being used. I'll confirm this tomorrow, but you're saying that all `Run` methods should be defined on a pointer receiver?

MStoykov:
@imiric 
```
$ golangci-lint run lib/executor | grep copylocks
lib/executor/shared_iterations.go:208:10: copylocks: Run passes lock by value: go.k6.io/k6/lib/executor.SharedIterations contains sync.Mutex (govet)
```

I am saying that now that there is a `sync.Mutex` probably all methods need to be with pointer receivers as otherwise it might be very confusing in the future when someone uses it ... and it doesn't work in the `Run` method as it is on a non-pointer receiver and it gets copied ;)

js/runner.go
```diff
@@ -743,6 +771,31 @@ func (u *VU) runFn(
...
+		u.scenarioIter[u.scenarioName] = 0
+	}
+	if u.getNextScLocalIter != nil {
+		u.scIterLocal = u.getNextScLocalIter()
+	}
+	if u.getNextScGlobalIter != nil {
+		u.scIterGlobal = u.getNextScGlobalIter()
+	}
+}
+
```
MStoykov, line 798: Why do we not have GlobalIter for some executors?

I do think whenever we have local one we can (and should) have a global one and all executors do have 
1. way to make SegmetnedIndex
2. make iterations somehow 
which just means it needs to call `next` and get the scaled and unscaled values of it ... preferably atomically as I di in [xk6-segment](https://github.com/MStoykov/xk6-segment/blob/8df18d74cb838a7c16167afec6feabb65ae93af0/pkg/segment/segment.go#L112-L122). Even in golang you don't need the additional result type ... it is only useful for javascript.

lib/state.go
imiric: I'm not sure it can though. We set it only once if it's not defined for a particular scenario, and `js.VU` is never destroyed so these IDs can persist after being shared among scenarios. There's no need to overwrite it on each VU activation, as the VU ID should remain static throughout the test run.

lib/state.go
imiric: Why? :smile: I think it makes sense to persist this across VU activations for each scenario, otherwise it would get messy with VU sharing between scenarios.

If this is the same comment as [the previous one](https://github.com/k6io/k6/pull/1863#discussion_r643780977), let's resolve this and discuss it there.

lib/state.go
imiric: Hhmm could be, I didn't think of that :smile: :+1:

Is relying on int overflow a valid pattern in Go?

js/modules/k6/execution/execution.go
imiric: Hhmm so what's your suggestion? :smile:

Rename this to `progress` and convert it to float, and add a (max)`duration` property?

js/runner.go
imiric: > Why do we not have GlobalIter for some executors?

I think I asked the same question to @na-- since [he mentioned these 3 explicitly in the specs](https://github.com/k6io/k6/pull/1863#issuecomment-782069940), and IIRC he doesn't think it would be useful for the other executors. I do agree that it would make things neater if we could just define this on `BaseExecutor`, but honestly I don't mind the current state much.

I don't quite understand what would be the benefit of changing `next`. The only difference I can see in the `xk6-segment` implementation is that we would avoid the additional read lock by not having to call `GetUnscaled()`. But semantically it makes more sense to me for `Next()` to only increment and to have a separate getter. It's also more flexible to use and I would have to change how it's used in `ramping-vus`.

lib/state.go
imiric: Resolved by 6fc76f85.

BTW @MStoykov is there a reason you used `^(uint64)(0)` instead of `^uint64(0)`?

lib/state.go
```diff
@@ -67,8 +67,20 @@ type State struct {
...
+	// Returns the iteration number of this VU in the current scenario.
+	GetScenarioVUIter func() uint64
+	// Returns the iteration number across all VUs in the current scenario
+	// unique to this single k6 instance.
+	// TODO: Maybe this doesn't belong here but in ScenarioState?
+	GetScenarioLocalVUIter func() uint64
+	// Returns the iteration number across all VUs in the current scenario
+	// unique globally across k6 instances (taking into account execution
+	// segments).
+	GetScenarioGlobalVUIter func() uint64
```
imiric, line 83: These 2 aren't relevant to a single VU/State, but I added them here since it was simpler to call them from the same place where we increment the other iterations. They would probably be better off as part of `ScenarioState` instead, but that would imply exposing a `ScenarioState.IncrIteration()` method, which we wanted to avoid for the VU `State`, so it's the same situation.

This doesn't bother me too much, but let me know if it should be changed.

imiric (commented):
@MStoykov I fixed the `copylocks` warning in 4a57d75e, but I changed `sync.Mutex` to a pointer instead of `Run`, to keep the change localized to this issue.

js/runner.go
MStoykov: I would like to get @na-- to review this PR before I continue commenting after this PR. I think I am pulling this PR in a different direction than what you think it should be and would like to get a third opinion if I or you are right on the different issues or if ... we are both wrong :rofl: 

Regarding this ... I think you keep slapping synchronization on top of things trying to fix them not synchronizing , while I am trying to see if we can do with the amount of synchronization we have currently. Some thoughts:

- SegmentedIndex doesn't need to be thread-safe for `ramping-vus`, so it means that its performance will suffer (somewhat) for something it doesn't need
- The places where SegementedIndex needs to be thread-safe are the same places where it needs to be threadsafe from `next`/`prev`/`GoTo` to `GetScaled` and `GetUnscaled` (might've gotten a name wrong, I ain't looking at the code at the moment).  For me, this means that there should be `SyncSegmentedIndex`(better name exist) that returns the `scaled`, `unscaled` in the cases where it needs to. Even if we only care about `next` we can probably just have an even smaller wrapper for now.
- Again you already lock for `next` and then ... you lock once more around it (through additional means) to get the 2 counters you care about, so combined with the below I think you can have it all done at once with *1* synchronization mechanism wrapping a nonsynchronized SegmentedIndex which can still be used in the cases where synchronization ... isn't required.
- I see no reason why we shouldn't have iteration counters for all executors ... we might want them only for specific ones, but there is nothing wrong (and it definitely will be useful) for even the externally-controlled one to have them so they can do the same tricks the `shared-iterations` can do. And given that the current implementation clearly will be simpler if all of them have it instead of only some - I am for all of them having it.

na--: > > Why do we not have GlobalIter for some executors?
> 
> I think I asked the same question to @na-- since [he mentioned these 3 explicitly in the specs](https://github.com/k6io/k6/pull/1863#issuecomment-782069940), and IIRC he doesn't think it would be useful for the other executors.

I am currently reviewing this PR and still have a lot of things I need to understand in it, but if you're referencing this sentence of mine:
> for `shared-iterations`, `constant-arrival-rate` and `variable-arrival-rate` scenarios, a way to get the sequential number of the current iteration among _all_ of the iterations for the current scenario, across all VUs and even instances.

Then the reason for listing only these 3 executors is simple - they are the only ones where we can actually _know_ the sequential number of an iteration globally, among all VUs and instances, without any synchronization. This is because the iteration number is the thing we segment on, not VUs, as is the case with the rest of the executors.

na--: I am still reviewing this (getting easily distracted and finding it a bit difficult to make headway :disappointed: ), but yeah, as @MStoykov mentioned above, something in the synchronization looks a bit wrong or unnecessary.

I think his first point is about the fact that `SegmentedIndex.Next()` has locking: https://github.com/k6io/k6/blob/25ee787380f7917ca33115711612a6a6d724f29f/lib/execution_segment.go#L750-L758
But we also lock where it's used, for example: https://github.com/k6io/k6/blob/25ee787380f7917ca33115711612a6a6d724f29f/lib/executor/shared_iterations.go#L193-L197

If `Next()` (and the other traversal methods) returned the new scaled and unscaled values, you wouldn't need to lock twice and ` GetScaled()` and `GetUnscaled()` will also be obsolete, I think?

MStoykov: yes, that is what I meant.

imiric: OK, I'll try refactoring this then.

imiric: Take a look at e9feff31. I decided to keep returning a struct instead of multiple values, to simplify future expansion.

imiric: The other comment here is about adding global iterations to all executors. @na-- WDYT?

lib/executor/constant_arrival_rate.go
```diff
@@ -196,9 +197,22 @@ func (car *ConstantArrivalRate) Init(ctx context.Context) error {
...
 	return err
 }
 
+// getNextGlobalIter advances and returns the next global iteration number for
+// this executor, taking into account the configured execution segment.
+// Unlike the local iteration number returned by getNextLocalIter(), this
+// iteration number will be unique across k6 instances.
+func (car *ConstantArrivalRate) getNextGlobalIter() uint64 {
+	res := car.segIdx.Next()
+	// iterations are 0-based
```
MStoykov, line 212:
```suggestion
	// iterations are indexed from 0
```
nit: this sounds better to me

js/runner.go
MStoykov: https://github.com/k6io/k6/pull/2057 what I actually meant ... I think've missed initially that you have the local iteration as a counter that you atomically increment. 

The commit doesn't fix any comments and the names are not great, but I think that + the commit message should be good enough explanation

lib/execution_segment.go
imiric: @na-- Care to contribute a useful description for this? :smile:

na--: How about this?
```suggestion
// SegmentedIndex is an iterator that returns both the scaled and the unscaled 
// sequential values according to the given ExecutionTuple. It is not thread-safe,
// concurrent access has to be externally synchronized.
```

lib/execution_segment.go
```diff
@@ -730,3 +730,79 @@ func (et *ExecutionTuple) GetNewExecutionTupleFromValue(value int64) (*Execution
...
+type SegmentedIndex struct {
+	start, lcd       int64
+	offsets          []int64
+	scaled, unscaled int64 // for both the first element(vu) is 1 not 0
+}
+
+// NewSegmentedIndex returns a pointer to a new SegmentedIndex instance,
+// given a starting index, LCD and offsets as returned by GetStripedOffsets().
+func NewSegmentedIndex(start, lcd int64, offsets []int64) *SegmentedIndex {
```
na--, line 743: btw, writing the comment suggestion above, we can probably make this constructor accept the `ExecutionTuple` directly. As far as I can see, this is the only way it's always going to be called, so we can just have it call `et.GetStripedOffsets()` internally?

imiric: Done in ed07691a.

core/local/local_test.go
```diff
@@ -1387,3 +1388,281 @@ func TestNewExecutionSchedulerHasWork(t *testing.T) {
...
+					vus: 2,
+					duration: '1s',
+					startTime: '2s',
+					gracefulStop: '0s',
+			    },
+		    },
+		};
+
+		export function cvus() {
+			const stats = Object.assign({scenario: 'cvus', globalVUID: __VU}, exec.getVUStats());
```
MStoykov, line 1423: by globalVUID do you mean "global for the instance" or for the test (in case of execution segments usage). As it is definitely not the latter and making it will be a breaking change I don't think is worth it.

On the other hand we have no way to get the latter ("global VUI id for the whole test") which is something that is a particular problem for cloud tests.

We can probably add it similar to how global Iteration was added, but maybe in a separate PR?

imiric: Maybe "global" was the wrong name for it, but it's `__VU`, so global for the test and instance, since no execution segments are used. It's just a label to ensure VUs are being reused in the test, but I can change it to something else if you want.

I don't think we discussed global VU IDs like you mention, so yeah, let's add it to a separate issue and PR.

js/runner.go
```diff
@@ -540,6 +542,10 @@ type VU struct {
 	setupData goja.Value
 
 	state *lib.State
+	// ID of this VU in each scenario
+	scenarioID map[string]uint64
+	// count of iterations executed by this VU in each scenario
+	scenarioIter map[string]uint64
```
na--, line 548: Are these maps just because of the `ramping-vus` executor?

core/local/local_test.go
na--: Should be fairly simple to add, so I'd prefer now if possible

na--: hmm actually it won't be all that simple, when you consider dynamically allocated VUs for the arrival-rate executors :disappointed: and if we currently don't calculate the total max VUs in every k6 instance, regardless of the segment, it might have performance implications if we just add this now... so probably a separate PR will be better :disappointed:

js/runner.go
imiric: Yeah, I thought it would be simpler to just keep the count for all scenarios in cases where VUs are released and reused, though it would be more efficient to just use `uint64` here instead. Mihail made the same comment above. I'll see if I can change it.

core/local/local_test.go
MStoykov: > performance implications if we just add this now.

We do now calculate through segment indexes which number an iteration is globally ... so I doubt calculating that for a VU will be noticeable

na--: Tto calculate which number a VU is, you first need to get the total max number of possible VUs (i.e. without an execution segments), which may be a very expensive operation for some `ramping-vus` configs. Whereas the global iteration counter is just a segmented index with the current execution segment, for which we've already done the calculations :man_shrugging:

MStoykov: why? 

as far as I am concern it is literally when you make a VU you have segmented index ... which gives it's local (unscaled) and global VU (scaled) number and that's it.

I don't care what is the max VUs it is the responsibility of the code making the VUs to make certain we don't go above that

na--: :man_facepalming: yeah, you are completely right, I had a stupid idea in my head about how this should be done, probably influenced by some cloud-related implementation details :sweat_smile: the simple segmented index will be perfectly adequate way to implement this :+1:

lib/executor/base_executor.go
```diff
@@ -66,6 +83,12 @@ func (bs BaseExecutor) GetConfig() lib.ExecutorConfig {
 	return bs.config
 }
 
+// getNextLocalVUID increments and returns the next VU ID that's specific for
+// this executor (i.e. not global like __VU).
+func (bs BaseExecutor) getNextLocalVUID() uint64 {
+	return atomic.AddUint64(bs.VUIDLocal, 1)
+}
```
imiric, line 90: Is this still needed? Right now we have 3 VU ID values: local to a single scenario (this one), global to a single instance (`__VU`),  and global across instances (added in #2064). I gathered this one from [this comment](https://github.com/k6io/k6/pull/1863#issuecomment-782069940) ("_in the current scenario_, what is the ID number of the VU currently running the iteration"), but I'm wondering how useful that is in practice...

MStoykov: I feel like the use case for this VU ID is if you have *multiple* scenarios and need to segment something based on the VUs between them. 

I would argue in most cases the *per scenario* iteration will be sufficient and probably all other cases are some kind of memory optimization that likely is better handled by SharedArray and co.

So I am slightly :+1: for dropping this particular counter, we can always put it back in the future. But I am by no means blocking the PR based on that,

imiric: Honestly, I would prefer it if we switched to GUIDs instead of incremental IDs, and just have those instead of 3 different types. It would break certain use cases like "run this only on the _first_ VU", but IMO those are workarounds for other features we need like events/hooks. GUIDs would work in both local and distributed contexts, and we would avoid all these counters.

It would be a huge breaking change, but something to think about for v1+?

na--: hmm there's nothing stopping you from having GUIDs even now, just have this in the init context:
```js
const vuID = getMyFancyUUID();
```

Integer IDs, especially when we can have them globally unique and mostly sequential so easily, have other benefits that UUIDs don't have, especially when it comes to ad-hoc partitioning of data or work :man_shrugging: so I don't really see a downside in having them here

cmd/cloud.go
```diff
@@ -251,14 +251,16 @@ This will execute the test on the k6 cloud service. Use "k6 login cloud" to auth
...
-			}()
+			progressBarWG := &sync.WaitGroup{}
+			if !quiet {
+				progressBarWG.Add(1)
+				defer progressBarWG.Wait()
+				go func() {
+					showProgress(progressCtx, conf, []*pb.ProgressBar{progressBar}, logger)
+					progressBarWG.Done()
+				}()
+			}
```
MStoykov, line 263: The problem is that:
1. `--quiet` just makes it way more likely (and possible at all) to see the race, as writing to stdout is *synchronous* so it basically makes the synchronization for us .... or just slows k6 down enough that it doesn't happen :man_shrugging: 
2. The problem actually (as somewhat, but not very well(IMO) shown by the stack trace) is that [GetProgress](https://github.com/k6io/k6/blob/master/lib/executor/base_executor.go#L75) is defined on the * non-pointer receiver* which means that when it's invoked it will *copy* the struct effectively *read*ing all of its fields. At the same time the `Init` *write*s to one of the fields which makes the race listed above.

So we should probably have all the `BaseExecutor` methods be on the pointer receiver as it won't copy the struct.

imiric: Resolved by f9e0ab65. Thanks! :clap:

js/runner.go
imiric: Actually, taking a look at this again, I'm not sure if changing it would be better. The memory needed for these maps is negligible (a few entries at most?), but we need some state to know if a VU ID was previously assigned by that same executor:
  https://github.com/k6io/k6/blob/f9e0ab6550c21477dbdfcb6cd49be8d755a47f11/js/runner.go#L620-L622

This is what drives the `BaseExecutor.VUIDLocal` counter (which could probably be unexported), as otherwise `GetNextScVUID()` would have to be called from elsewhere. Maybe in each executor after `GetPlannedVU()`:
  https://github.com/k6io/k6/blob/f9e0ab6550c21477dbdfcb6cd49be8d755a47f11/lib/executor/shared_iterations.go#L277

But that would mean several implementations of it, when using a map is simpler and we wouldn't need to think about this for new executors that might reuse VUs.

Let me know if there's another reason both you and @MStoykov dislike this approach and well, any suggestions. :)

lib/executor/shared_iterations.go
```diff
@@ -232,15 +234,26 @@ func (si SharedIterations) Run(parentCtx context.Context, out chan<- stats.Sampl
 	regDurationDone := regDurationCtx.Done()
 	runIteration := getIterationRunner(si.executionState, si.logger)
 
+	maxDurationCtx = lib.WithScenarioState(maxDurationCtx, &lib.ScenarioState{
```
na--, line 237: I don't like how we attach this only to `maxDurationCtx`. Even if this works now, I think this can become a subtle source of bugs in the future, see https://play.golang.org/p/xAEuY_FOOC3

In essence, right now this is the diagram of the contexts currently looks:
```
parentCtx -> maxDurationCtx -> regDurationCtx
                           \-> maxDurationCtxWithScenarioState -> vuCtx
```

This is not necessarily a problem, even before the `vuCtx` was branched as a "sibling" to the `regDurationCtx`, but if we can move the scenario state attachment before the `maxDurationCtx` and `regDurationCtx` it seems to me that it will much safer. Something like this:
```
parentCtx -> parentCtxWithScenarioState -> maxDurationCtx -> regDurationCtx
                                                         \-> vuCtx
```

imiric: Hhmm would something like [d5ffc781](https://github.com/imiric/k6/commit/d5ffc78183602b87c679aad0854ab2c454efa853) be an improvement?

The problem with attaching it to `parentCtx` is that we need to pass the `progressFn` function to `ScenarioState`. Since this is a lambda created after `getDurationContexts()` is called, I had to split the initialization of `ScenarioState` like this, which is not nice.

A better approach might be to expand the `lib.Executor` interface and refactor all executors to have a standalone `GetProgress()` method (renaming the current one to `GetProgressBar()`), and then attach `ScenarioState` before `Run()` is called:
https://github.com/k6io/k6/blob/f9e0ab6550c21477dbdfcb6cd49be8d755a47f11/core/local/local.go#L329

But that would involve a much larger refactor to avoid what is a hypothetical source of bugs, so I don't want to start that here.

na--: Yeah, https://github.com/imiric/k6/commit/d5ffc78183602b87c679aad0854ab2c454efa853 has other issues that might also be a similarly subtle source of future bugs :sweat_smile: And I don't really have a better idea how to not have them without a big executor refactor like the one you propose ... :disappointed:  We could probably have some of this and `getDurationContexts()` wrapped in a new helper function, but the progress will still have to be "dangling" and attached separately.

In any case, we can ignore this for now, but I'll add a new issue to maybe refactor this in the future.

na--: Created https://github.com/k6io/k6/issues/2068, resolving this thread so it doesn't clutter the review.

js/modules/k6/execution/execution.go
```diff
@@ -0,0 +1,169 @@
...
+	if rt == nil {
+		return nil, errors.New("goja runtime is nil in context")
+	}
+
+	stats := map[string]interface{}{
+		"id":         vuState.VUID,
+		"idGlobal":   vuState.VUIDGlobal,
+		"idScenario": vuState.VUIDScenario,
+		"iteration":  vuState.Iteration,
+		"iterationScenario": func() goja.Value {
```
na--, line 58:
```suggestion
		"idInScenario": vuState.VUIDScenario,
		"iteration":  vuState.Iteration,
		"iterationInScenario": func() goja.Value {
```

This sounds just a bit better to me, but it's a minor nitpick. @MStoykov @imiric WDYT?

js/modules/k6/execution/execution.go
na--: I think `duration` is fine (and we might add `maxDuration` in the future, not now). Though I'd like it to be a float number in milliseconds instead of a string, so JS people don't have to parse stringy times.

js/modules/k6/execution/execution.go
imiric: Sure, I'm fine with that. Just that `idGlobal` and `iterationGlobal` in `getScenarioStats()` would slightly stick out, and renaming them to something like `idInTest` would be a bit more confusing. We'll have to document each of these properly anyway, so having them shorter and following the same naming convention would be better UX. We'd also document them in the TypeScript definitions, so the name itself doesn't matter much.

lib/executor/base_executor.go
```diff
@@ -38,53 +40,74 @@ import (
...
 // GetConfig returns the configuration with which this executor was launched.
-func (bs BaseExecutor) GetConfig() lib.ExecutorConfig {
+func (bs *BaseExecutor) GetConfig() lib.ExecutorConfig {
 	return bs.config
 }
 
+// getNextLocalVUID increments and returns the next VU ID that's specific for
+// this executor (i.e. not global like __VU).
+func (bs *BaseExecutor) getNextLocalVUID() uint64 {
+	return atomic.AddUint64(bs.VUIDLocal, 1)
```
na--, line 89: Shouldn't this also be segmented? Right now with it being per-instance, I am not sure how useful it will be?

js/modules/k6/execution/execution.go
```diff
@@ -0,0 +1,169 @@
...
+	if vuState.GetScenarioGlobalVUIter != nil {
+		iterGlobal = vuState.GetScenarioGlobalVUIter()
+	} else {
+		iterGlobal = goja.Null()
+	}
+
+	stats := map[string]interface{}{
+		"name":      ss.Name,
+		"executor":  ss.Executor,
+		"startTime": ss.StartTime,
```
na--, line 94: This probably needs a type conversion, either to a proper string or to some sort of a unix timestamp (with milliseconds). Right now, it's Go's `time.Time` type and behaves very strangely in JS. For example:
```js
export default function () {
    let scenarioStats = execution.getScenarioStats();
    console.log(scenarioStats.startTime);
    console.log(JSON.stringify(scenarioStats.startTime));
    console.log(typeof scenarioStats.startTime);
    console.log(scenarioStats.startTime instanceof Date);
    console.log(typeof scenarioStats.startTime.getMonth);
    console.log(Object.keys(scenarioStats.startTime));
}
```

will result in this:
```
2021-06-17 16:06:17.00787938 +0300 EEST m=+0.488634128
"2021-06-17T16:06:17.00787938+03:00"
object
false
undefined
add,addDate,after,appendFormat,before,clock,date,day,equal,format,gobEncode,hour,iSOWeek,in,isZero,local,location,marshalBinary,marshalJSON,marshalText,minute,month,nanosecond,round,second,string,sub,truncate,uTC,unix,unixNano,weekday,year,yearDay,zone
```

lib/executor/base_executor.go
imiric: :-/ We brought up actually removing this counter in [this thread](https://github.com/k6io/k6/pull/1863#discussion_r651867771). I'm not sure how helpful it is already, but even less so if we add another global variant of it. I think we already have a lot of possibly confusing values and adding anything new without knowing if users will actually need it only forces us to support it on the off chance someone is depending on it. The module is already large enough, so I'd rather leave this as is or remove it and add it later if needed.

na--: Ah, I replied in that previous thread only to your latest message, without reading the start... :sweat_smile: :+1: for removing this counter and potentially adding it back in the future. I could argue why a segmented per-scenario VU index in would be useful, but we can certainly add it later.

na--: Hmm actually, a segmented index here would be most useful for things like `constant-vus` and `ramping-vus` and maybe `per-vu-iterations`, where the VU number is the leading thing, not the iteration number, so that's what people would likely want to segment on :man_shrugging: 

So, yeah, we can remove it for now, so we can ship the PR more quickly, but I'm almost certain we're going to have to put a segmented per-scenario VU index back fairly soon :sweat_smile: Especially since we still don't have a per-scenario VU init function (https://github.com/k6io/k6/issues/785) where something like this could be generated easily by users.

js/modules/k6/execution/execution.go
imiric: Yeah, I ran into this when I was writing the tests, see https://github.com/dop251/goja/issues/170.

I'm now converting it in Go, see 80671dfa.

lib/executor/base_executor.go
imiric: OK, I removed it in 670b7126 :disappointed: FYI @MStoykov

na--:
> List of problems that the original issue intended to solve:
> ....

@MStoykov, the original issue was just a catch-all for all of the somewhat similar "lack of information" problems that users experienced. In my mind, this PR was never meant to address _all_ of them, just to be at least an MVP for the trickiest and/or most useful ones first. After we merge this, I'm going to create a new issue with the remaining unsolved problems and we'll chip away at them in future PRs. It has grown to more than an MVP, and it addresses some things which are not necessarily problems or needs for anyone, so I am not completely opposed to pruning a few more things from it, but probably not everything you suggested...

> If we keep this, we either need to rename the fields to better show what they are or the method to `getInstanceStats`/`getTestIntanceStats`

Yeah, I was going to bring this up as well, the `getTestStats()` names and/or `getTestStats()` itself probably need a rename. We can discuss this and all of the names in Slack. We probably should have a consistent and logical naming scheme that makes clear where everything falls under (whole test (global) vs per instance, in scenario or out of it, etc.)

> `duration` - is the time the test has ran so far as a string, probably better to be float. 

I agree, already brought this up here in https://github.com/k6io/k6/pull/1863#discussion_r653507876

> progress - shows how far the executor has come as a float between 0 and 1. where 1 is completed. This is based on the max duration and is in some ways `time.now() - startTime/scenarioDuration` (this one isn't provided). No idea what that is useful for, and given that adding `scenarioDuration` will probably be way easier I wonder if we shouldn't add that instead

It sometimes depends on the number of iterations, it's not always the duration. I am not sure if users would be able to use it in a script, but might be useful for someone running with `--quiet` and wanting to log some progress occasionally :man_shrugging: 

> On this topic I am still of the opinion (now even more so) that the API is bad and it should look like :
> 
> 
> ```
> import exec from "k6/execution"; // exec might be changed
> 
> export default function() {
> 	exec.scenario.iteration // instead of  exec.getSecnarioStats().iterationGlobal
> 	exec.vu.id // instead of exec.getVUStats().idGlobal
> }
> ```

I disagree, I don't see what the problem with getter functions is, if anything, they add clarity to me. Consider that `k6/execution` will have more things in it, and they won't all be simply supplying information. For example, `abortTest()` (https://github.com/k6io/k6/pull/1920) should probably go in there. This seems quite nice to me:
```js
import execution from 'k6/execution';

export default function () {
    if (execution.getScenarioStats().iterationGlobal > 10000) {
        execution.abortTest();
    }
}
````

File: js/modules/k6/execution/execution.go (hunk 25 above)
Comment (MStoykov) [reply js/modules/k6/execution/execution.go]: hm, given the fact Date doesn't have timezone and I am not really certain what anyone will use the Date Object instead of a the seconds since UTC epoch :thinking: I am somewhat for returning it as a number to be honest

File: js/modules/k6/execution/execution.go (hunk 13 above)
Comment (imiric) [reply js/modules/k6/execution/execution.go]: Resolved by 7f420594.

File: js/modules/k6/execution/execution.go (hunk 25 above)
Comment (imiric) [reply js/modules/k6/execution/execution.go]: You're right, fixed in f32a4d05.

**Comment by imiric** [comment]
@MStoykov @na-- [I extracted the JS extension](https://github.com/k6io/xk6-execution/pull/1) and cleaned this up a bit. Let me know if there are any leftover blockers here.

Do we need more tests? Most of them were written in JS and copied to the extension repo as is, but will need some tweaking to make them run there. So I'm wondering if we should rewrite some in Go to cover some scenarios here.

**Comment by na--** [comment]
I haven't looked at the new version of the PR yet, but regarding more tests, I think we probably don't need to add or convert any more tests now, unless it's something very quick... :man_shrugging:  All of the old k6 tests should verify that we haven't broken any old k6 behavior with the refactoring (:crossed_fingers:), and we always intend to bring the extension back into k6, so it will come with all of the tests then,

File: lib/testutils/minirunner/minirunner.go [hunk 26]
```diff
@@ -124,34 +131,58 @@ func (r *MiniRunner) HandleSummary(ctx context.Context, s *lib.Summary) (map[str
...
 }
 
 // ActiveVU holds a VU and its activation parameters
 type ActiveVU struct {
 	*VU
 	*lib.VUActivationParams
 	busy chan struct{}
+
+	scenarioName              string
+	getNexIterations          func() (uint64, uint64)
```
MStoykov, line 150: nit:typo 

```suggestion
	getNextIterations          func() (uint64, uint64)
```

MStoykov (approved):
LGTM!

I still think something should be dropped (progress in particular), but that can be discussed along the API in the extension

lib/executor/ramping_arrival_rate.go
```diff
@@ -175,11 +175,23 @@ func (varc RampingArrivalRateConfig) HasWork(et *lib.ExecutionTuple) bool {
...
+// Init values needed for the execution
+func (varr *RampingArrivalRate) Init(ctx context.Context) error {
+	// err should always be nil, because Init() won't be called for executors
+	// with no work, as determined by their config's HasWork() method.
+	et, err := varr.BaseExecutor.executionState.ExecutionTuple.GetNewExecutionTupleFromValue(varr.config.MaxVUs.Int64)
+	varr.et = et
+	varr.iterSegIndex = lib.NewSegmentedIndex(et)
+
+	return err //nolint: wrapcheck
+}
```
na--, line 193: Hmm this `Init()` makes sense, I think, but I wonder why we didn't have this re-segmentation before? Was that a bug we have to mention in the release notes separately?

lib/testutils/minirunner/minirunner.go
imiric: Resolved in a recent force-push.

lib/executor/constant_vus.go
```diff
@@ -176,15 +176,26 @@ func (clv ConstantVUs) Run(parentCtx context.Context, out chan<- stats.SampleCon
...
+		StartTime:  startTime,
+		ProgressFn: progressFn,
+	})
+
 	returnVU := func(u lib.InitializedVU) {
 		clv.executionState.ReturnVU(u, true)
 		activeVUs.Done()
 	}
+
+	// Channel for synchronizing scenario-specific iteration increments
```
na--, line 191: This comment seems to be for an earlier version of the PR?

na-- (approved):
LGTM besides the comment. And I am still not sure if this change doesn't have any side-effects - it looks correct, but as I mentioned in the comment, does that mean we had a minor bug before? :man_shrugging:

lib/executor/constant_vus.go
imiric: :facepalm: Merge gone awry... It was also in `per_vu_iterations.go`. Fixed in latest push.

Some of the intermediate commits are definitely broken, but :shrug:

lib/executor/ramping_arrival_rate.go
imiric: Not sure, it seems to segment properly to me on `master`, but maybe I'm not reproducing it.

If it was a minor bug we should mention it, sure.

MStoykov: Apparently, I didn't do part of https://github.com/k6io/k6/issues/1308 :sweat: for this executor.

With 
```javascript
export let options = {
        scenarios: {
                contacts: {
                        startRate: 5,
                        timeUnit: '1s',
                        preAllocatedVUs: 2,
                        maxVUs: 3,
                        stages:[ { target: 5, duration: "3s"}],
                },
        },
};
export default function () {
    console.log(__VU, __ITER);
}
```
(Notice the `maxVUs: 3` and preallocatedVUs: '2', it important and they both can't be proportionally split given a segment)
You can clearly see that with master and `--execution-segment '0:1/2'` it does  7 iterations (there re in total 15 = 3s * 5 iter/s).
With the changes from this branch it does 10 or 2/3 of the iterations as when you scale based on the maxVUs it gets 2 maxVUs out of 3. 
So ... good thing we caught and we should mention it. I doubt anyone actually got bitten by this or even if all that many people experienced it unknowingly, but it's definitely still a bug

na--: @imiric can you please add a link to or a copy of :arrow_up: in the PR description, so when we link it in the v0.33.0 release notes, it's clear that it also fixes it?

na-- (approved):
:shipit:

MStoykov (approved):
LGTM :clap:

PR #1867, opened 2021-02-19
update goja to let-const branch

MStoykov:
testing https://github.com/dop251/goja/tree/let-const


na--:
This seems like it should be closed in favor of https://github.com/loadimpact/k6/pull/1904?

Issue #1858, opened 2021-02-15
Internal module instances per VU


MStoykov:
Currently, internal modules such as `http` have a single instance of them for the whole k6 instance. This is not how it works for not internal modules - imported js files, which get an instance *per* VU. 

This will make it possible for per VU settings (of a single module) such as the callback from #1856 to actually not be in the [lib.State](https://github.com/loadimpact/k6/blob/11811171a3c416daf17f7f0ad182ca30aa9990bb/lib/state.go#L43). This is also probably true for the [RPSLimit](https://github.com/loadimpact/k6/blob/11811171a3c416daf17f7f0ad182ca30aa9990bb/lib/state.go#L61) and [Transport](https://github.com/loadimpact/k6/blob/11811171a3c416daf17f7f0ad182ca30aa9990bb/lib/state.go#L55) especially if in the future we have *multiple* transports. 

Having both per VU and global store would've also let stuff like #1739 to be implemented only by a single module without other changes in other parts of the code.

This is not to say that `lib.State` should be removed as there are definitely things that need to be in a more *shared* location between modules such as the tls.Config.

Currently all, but [HTTP](https://github.com/loadimpact/k6/blob/11811171a3c416daf17f7f0ad182ca30aa9990bb/js/modules/k6/http/http.go#L50) and [gRPC](https://github.com/loadimpact/k6/blob/11811171a3c416daf17f7f0ad182ca30aa9990bb/js/modules/k6/grpc/grpc.go#L34) use empty structs (both the exceptions use it for constants) and one of the easiest ways to add a way to have this will be to [instead of taking an instance](https://github.com/loadimpact/k6/blob/11811171a3c416daf17f7f0ad182ca30aa9990bb/js/internal/modules/modules.go#L44) to take the `New` method most of those call to get an instance. 

This though is a breaking change especially with `xk6` in mind. Given that, and the fact that it will also be nice if there is a way to have a "shared" global state per/in module without the usage of global variables such as in [xk6-counter](https://github.com/MStoykov/xk6-counter/blob/bbfeabd9222254eceff36a9c1d8d9f8119c2e512/counter.go#L16) maybe it will be better to keep the current behaviour, but being able to add a method `New` (name up for debate, maybe Instantiate) that will then return a new copy of another object that will be the actual module instance: 

So for HTTP it will be something like: 
```golang 
type HTTP struct{
   // global fields for the whole module that can be accessed by each instance + some syncing
}
type HTTPinstance {
[...]

Issue #1802, opened 2021-01-15
Drop common.Bind for goja native support


MStoykov:
After https://github.com/dop251/goja/commit/06e99480d22cd2c4361b2d7edbd67f2030b28ecb the scratched out part of mine comment https://github.com/dop251/goja/issues/243#issuecomment-759547730 is now possible :tada: . 

Whether we want to do is another question. 

Summary of what currently [common.Bind](https://github.com/loadimpact/k6/blob/c00a85bfea21d748f0f9d11fdbc326087f34c411/js/common/bridge.go#L137) seems to be doing(probably missed something):
It's predominant roll is to make certain that for methods on the struct provided that look like:
```golang
func (s someStruct) something(ctx context.Context, another parameter int) (val, err) 
```

1. `ctx` will be set to the one provided as the argument to Bind 
2. If the name starts with 'X' it will wrap it in a function call (removing the X from the name) so that it can be used as a constructor - this should probably be using [goja.ConstructorCall](https://pkg.go.dev/github.com/dop251/goja#ConstructorCall) 
3. returning a map[string]interface{} with everything that was an exported property of the original object as name=>wrapper. Where sometimes the wrapper does nothing :). 

The last point means that things such [as using an interface in the arguments of a function](https://community.k6.io/t/pass-value-from-js-to-go-with-interface-argument/1253), but `common.Bind`-ing the "to be" argument will lead to it not answering the interface{}. The [code posted in the community forum](https://community.k6.io/t/pass-value-from-js-to-go-with-interface-argument/1253) showcases this a lot better, then I can put it in words :rofl: .

Other problems with the current code include that it is kind of ... hard to read due to the sheer amount of reflect in it. It also having support for `*context.Context`(the pointer part), which seems completely unnecessary. 

Also needing to be called in order to have access to the context, is something that we will probably need to document for [xk6]*(https://github.com/k6io/xk6) and explain it. 

Proposed change:
1. add the context.Context as a global object to each goja.Runtime, or something else so we can get it from the goja.Runtime
2. change all the calls to use 
```golang
func(goja.FunctionCall, goja.Runtime) goja.Value
```
or for constructors - drop the 'X' and change to
```golang
func(goja.ConstructorCall, goja.Runtime) goja.Value
```
3. drop the common.Bind at all 

This will mean that instead of having 
```golang
func(ctx.Context, number int64) {
// do something with number
}
```
we will have something like
```golang
[...]

PR #1869, opened 2021-02-23
Output refactoring

na--:
This completely fixes https://github.com/loadimpact/k6/issues/917, fixes https://github.com/loadimpact/k6/issues/1075, and fixes https://github.com/loadimpact/k6/issues/1606. It also mostly fixes https://github.com/loadimpact/k6/issues/1075. I think we can close it now and do the transition of the remaining Collectors to Outputs in separate PRs, since they should be perfectly usable with the simple `Output` adapter that wraps them for now. 

This PR already includes the new `json` output and an adapter for every other `Collector`-type one. The current plan is to also move the `cloud` and `csv` outputs to the new `Output` interface for k6 v0.31.0, while `influxdb`, `statsd`, `datadog`, and `kafka` are left with the old `Collector` interface and moved in k6 v0.32.0. The major bugs (https://github.com/loadimpact/k6/issues/1075 and https://github.com/loadimpact/k6/issues/1606) should be fixed for all outputs in v0.31.0 regardless, we're just leaving the semi-mechanical translation for later. We can use the additional time to also revisit some other issues (e.g. https://github.com/loadimpact/k6/issues/1801, https://github.com/loadimpact/k6/issues/1052, https://github.com/loadimpact/k6/issues/737 and many more).


output/helpers.go
```diff
@@ -0,0 +1,125 @@
...
+// Stop waits for the periodic flusher flush one last time and exit. You can
+// safely call Stop() multiple times from different goroutines.
+func (pf *PeriodicFlusher) Stop() {
+	select {
+	case <-pf.stop:
+		// Aldready stopped
+	default:
+		close(pf.stop)
+	}
+	<-pf.stopped
```
MStoykov, line 106: I think this is still racy as it should be possible for two goroutines to get between the `<-pf.stop`  and `close(pf.stop)`. Not certain if it matters enough, but I would argue this needs a lock ... I doubt that there will be much congestion on it ;)

na--: > it should be possible for two goroutines to get between the `<-pf.stop` and `close(pf.stop)`. 

I don't understand what you mean here, sorry :confused:

na--: Ah, ok, I got it, we could theoretically close the channel more than once...

na--: I'll fix it with a `sync.Once` when I write the tests

cmd/outputs.go
```diff
@@ -0,0 +1,275 @@
...
+}
+
+func (ca *collectorAdapter) SetRunStatus(latestStatus lib.RunStatus) {
+	ca.collector.SetRunStatus(latestStatus)
+}
+
+// Stop implements the new output interface.
+func (ca *collectorAdapter) Stop() error {
+	ca.runCtxCancel()
+	time.Sleep(2 * time.Second) // shrug, better than nothing...
```
MStoykov, line 271: :rofl:

MStoykov: I mean you can wait for the `Run` from line 256 to end ... that will be better ;)

output/json/json.go
```diff
@@ -0,0 +1,152 @@
...
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+package json
+
+import (
+	"compress/gzip"
+	stdlibjson "encoding/json"
```
MStoykov, line 25: I think this isn't necessary, but I am fine with it being that way ;)

output/helpers_test.go
```diff
@@ -0,0 +1,118 @@
...
+	buffer.AddMetricSamples([]stats.SampleContainer{single})
+	assert.Equal(t, []stats.SampleContainer{single}, buffer.GetBufferedSamples())
+	assert.Equal(t, cap(buffer.buffer), 3)
+	assert.Empty(t, buffer.GetBufferedSamples())
+}
+
+//nolint:gosec
+func TestSampleBufferConcurrently(t *testing.T) {
+	t.Parallel()
+	rand.Seed(time.Now().UnixNano())
```
MStoykov, line 66: From the perspective that this might get other tests and if one of the random values was the actual reason behind this failed I would prefer if:
1. we don't use the global but a local instance
2. we t.Log the random seed
I think that you can just also generate a chan int64 with 
```golang
var signal :+ make(chan struct{})
defer close(signal)
var ch := make(chan int64, 50)
go func() {
for {
   select {
   case ch<- rand.Int63n(1000):
   case <- signal:
     return
   }
}()
```
So you don't have to wrap the rand.Source  so it is concurrently safe.

(Code above was written in this text fields so it will probably not work :P )

MStoykov (commented):
LGTM. I will need to try it out some as trying to figure out if we've broken a config without trying to use it ... is too complicated :P

cmd/outputs.go
na--: True, I'll do that :+1:

output/json/json.go
na--: I had some issues when I imported it in `cmd/`, though that might have been a gopls issue instead of a Go one :man_shrugging:

output/helpers_test.go
na--: I'll remove rand altogether, it doesn't really bring any benefit... In fact, the minimum synchronization it requires (be it a channel or the stdlib's locks) might prevent us from having a concurrency issue surface, which is the point of the test.

cmd/outputs.go
na--: Fixed by https://github.com/loadimpact/k6/pull/1869/commits/e5afaa3393905be4e6d81ca53da838426d0a0df5

output/helpers_test.go
na--: Went half-way, made the randomness predictable and took it out of the goroutines: https://github.com/loadimpact/k6/pull/1869/commits/cbe173bdefbe64241a64826d8c14c7c8297cca9c

output/helpers.go
na--: Resolved by https://github.com/loadimpact/k6/pull/1869/commits/19e623f855ba46dc5989341ca13492ce2b51a70c

na--:
Hmm actually with the changes from https://github.com/loadimpact/k6/pull/1869/commits/e5afaa3393905be4e6d81ca53da838426d0a0df5 that @MStoykov suggested, I think even https://github.com/loadimpact/k6/issues/1423 and https://github.com/loadimpact/k6/issues/1606 are fixed... :tada:  I'll add some tests for the JSON output and remove the `oldjson` one and run some manual tests and this PR should be ready for merging. I'd still like to transition at least the `csv` and `cloud` Collectors to the new `Output` interface before k6 v0.31.0, but we can leave the other ones for after.

output/helpers.go
```diff
@@ -0,0 +1,125 @@
...
+
+// Stop waits for the periodic flusher flush one last time and exit. You can
+// safely call Stop() multiple times from different goroutines, you just can't
+// call it from inside of the flushing function.
+func (pf *PeriodicFlusher) Stop() {
+	pf.once.Do(func() {
+		close(pf.stop)
+	})
+	<-pf.stopped
+}
```
MStoykov, line 106: nit: you can technically send `struct{}` on `pf.stop` and then close it in the `run` and wait on it here in the `once.Do`. 

This should be 1 less channel and 1 less check on closed channel ... 

not important though ;)

na--: I could have optimized it in other ways as well, but this seems good enough and much more understandable, so I don't plan on changing it if there aren't any bugs :man_shrugging:

MStoykov:
Did some runs with `-o json` and `-o cloud` - no problems and it did pickup some cloud settings.
So it LGTM to be merged.

MStoykov (approved):

output/extensions/registry.go
```diff
@@ -0,0 +1,64 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+package extensions
```
imiric, line 21: I'm not sure if this should be called "extensions" or be in a separate package. Like I mentioned in another comment, I don't think we should have a distinction between built-in and external outputs, since they all should implement the `output.Output` interface and are treated the same internally.

Also being in `output/extensions` makes it seem like these are extensions _for_ outputs, not additional outputs that are loaded externally.

output/json/wrapper.go
```diff
@@ -26,38 +26,40 @@ import (
...
 		Tags:  sample.Tags,
 	}
 }
 
-func WrapSample(sample *stats.Sample) *Envelope {
-	if sample == nil {
-		return nil
-	}
-	return &Envelope{
+// WrapSample is a used to package a metric sample in a way that's nice to
```
imiric, line 52:
```suggestion
// WrapSample is used to package a metric sample in a way that's nice to
```

output/json/wrapper.go
```diff
@@ -26,38 +26,40 @@ import (
...
+// export to JSON.
+func WrapSample(sample stats.Sample) Envelope {
```
imiric, line 54: Maybe unrelated to this PR, but it would be good to move this file to a common location for all outputs (`output/json.go` or `stats/json.go`?), since it seems generic enough, and it's already used by the Kafka output, so it's weird that it would depend on the JSON output package directly.

cmd/ui.go
```diff
@@ -111,31 +112,25 @@ func modifyAndPrintBar(bar *pb.ProgressBar, options ...pb.ProgressBarOption) {
...
+	default:
+		for _, out := range outputs {
+			outputDescriptions = append(outputDescriptions, out.Description())
 		}
-		fprintf(stdout, "     output: %s%s\n", ui.ValueColor.Sprint(out), ui.ExtraColor.Sprint(link))
-	} else {
-		fprintf(stdout, "     output: %s\n", ui.ValueColor.Sprint(output))
 	}
+
+	fprintf(stdout, "     output: %s\n", ui.ValueColor.Sprint(strings.Join(outputDescriptions, ";")))
```
imiric, line 133: Nit: maybe add a space after the semi-colon and use a comma instead. Here's what it currently looks like:

```
output: json(test.json);influxdb (http://127.0.0.1:8086)
```

I guess a space between `json` and `(test.json)` would be good too.

imiric (commented):
Mostly LGTM. I left a few comments we can discuss, and the only blocker for me would be the way built-in outputs are "registered" and that whole built-in/external distinction.

Some general notes:

- I like the `output.Params` approach and every output plucking what it needs from it.

- The config consolidation in each output feels like a future wart, but maybe we can't improve that before #883.

- :+1: for abandoning the "collector" naming and standardizing everything on "output".

- The `PeriodicFlusher` is a neat abstraction to avoid every output re-implementing it. :+1:

core/engine.go
```diff
@@ -109,10 +111,47 @@ func NewEngine(
...
+// StopOutputs stops all configured outputs.
+func (e *Engine) StopOutputs() {
+	e.stopOutputs(len(e.outputs))
+}
+
+func (e *Engine) stopOutputs(upToID int) {
+	e.logger.Debugf("Stopping %d outputs...", upToID)
+	for i := 0; i < upToID; i++ {
+		if err := e.outputs[i].Stop(); err != nil {
+			e.logger.WithError(err).Errorf("Stopping output %d failed", i)
```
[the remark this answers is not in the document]
na--: The problem here is that I don't have a `Name()`, I have a `Description()`, which wouldn't be suitable here :disappointed: ... To improve the error message, I either have to change the `output.Output` interface and force every output to expose a `Name()`, or I'd have to add a wrapper struct for outputs in the `Engine` that contains its type (or whole `Params`). And I can't just use a map, since users can have multiple outputs with the same type.

cloudapi/config.go
```diff
@@ -262,3 +264,24 @@ func MergeFromExternal(external map[string]json.RawMessage, conf *Config) error
 	}
 	return nil
 }
+
+// GetConsolidatedConfig combines the default config values with the JSON config
+// values and environment variables and returns the final result.
+func GetConsolidatedConfig(jsonRawConf json.RawMessage, env map[string]string) (Config, error) {
```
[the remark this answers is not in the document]
na--: > this consolidation should happen before they're initialized

It does :confused: This is precisely what this bit of code does: https://github.com/loadimpact/k6/blob/7e1bc5dcab910f0d197a63e843e5591de4c60423/cmd/outputs.go#L68-L72

And this isn't new technical debt, it was already part of k6 and #883, it was just squirreled away in `cmd/` in a much, much worse way, see these snippets from `master`:
https://github.com/loadimpact/k6/blob/51790fc8df37113ac20b586b28051e21672dab16/cmd/config.go#L68
https://github.com/loadimpact/k6/blob/51790fc8df37113ac20b586b28051e21672dab16/cmd/config.go#L96
https://github.com/loadimpact/k6/blob/51790fc8df37113ac20b586b28051e21672dab16/cmd/config.go#L174
https://github.com/loadimpact/k6/blob/51790fc8df37113ac20b586b28051e21672dab16/cmd/config.go#L196
https://github.com/loadimpact/k6/blob/51790fc8df37113ac20b586b28051e21672dab16/cmd/collectors.go#L90-L97

Though I just saw that I missed `config.Name = null.StringFrom(arg)` from that last code snippet, so I'll fix that... :sweat_smile: 

But yeah, we have to fix this configuration mess - the current state is definitely not the end goal, it is just slightly more sane than before, #883 is still very much required.

cmd/outputs.go
```diff
@@ -0,0 +1,274 @@
...
+			return nil, fmt.Errorf("could not create the '%s' output: %w", outputType, err)
+		}
+		result = append(result, output)
+	}
+
+	return result, nil
+}
+
+func parseOutputArgument(s string) (t, arg string) {
+	parts := strings.SplitN(s, "=", 2)
```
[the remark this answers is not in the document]
na--: This is parsing of the `--out` / `K6_OUT` option (e.g. `--out influxdb="https://whatever"`). The `Output` interface has no bearing here, we are parsing the `--out` precisely to value to figure out which `Output` _implementation_ we have to initialize. The only way to improve this further would be to have a custom type for `Config.Out`: https://github.com/loadimpact/k6/blob/7e1bc5dcab910f0d197a63e843e5591de4c60423/cmd/config.go#L55

But that would be better handled in a separate PR, if we ever want to do it.

output/json/wrapper.go
na--: Hmm I am not sure any other outputs would ever use the function, but we can easily move it to `/output` in the future, if that ever proves necessary. Or not, I don't really see the problem with the current explicit dependency on `output/json` from the `kafka` output. Isn't it actually a good thing, it doesn't cost us anything, but it somewhat shows the `kafka` output is encoding the metric samples it sends the same way as the `json` output?

cloudapi/config.go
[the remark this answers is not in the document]
imiric: > It does :confused: This is precisely what this bit of code does

I know. My point was that each output package shouldn't handle raw configuration at all, and should just receive the consolidated values, which should be done much earlier and only once / in a single location.

But yeah, maybe something for #883 then. At least it's consistently bad now :smile:

cmd/outputs.go
```diff
@@ -0,0 +1,274 @@
...
+	"github.com/loadimpact/k6/stats/kafka"
+	"github.com/loadimpact/k6/stats/statsd"
+)
+
+// TODO: move this to an output sub-module after we get rid of the old collectors?
+//nolint: funlen
+func getAllOutputConstructors() (map[string]func(output.Params) (output.Output, error), error) {
+	// Start with the built-in outputs
+	result := map[string]func(output.Params) (output.Output, error){
+		"json": json.New,
```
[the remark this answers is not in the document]
na--: I started that way, but then I refactored it the way it currently is, for several different reasons, and I'd advocate maybe refactoring the JS extensions as well in the future. Here are the reasons:
1. This allows us to print much better error messages instead of panicking when there are name conflicts, see [below](https://github.com/loadimpact/k6/blob/7e1bc5dcab910f0d197a63e843e5591de4c60423/cmd/outputs.go#L129).
2. It allows us to very easily distinguish between built-in outputs and extension, which might be very useful if we want to send the latter in the usage reports.
3. Package `init()` functions are a bit of an antipattern (see [this](https://leighmcculloch.com/posts/tool-go-check-no-globals-no-inits/), for example), and in this case it's completely unnecessary to have them for built-in modules.
4. This doesn't look more elegant than a static map, to me it looks ugly as hell, I'd normally consider such imports code smells in a Go program :sweat_smile: :  https://github.com/loadimpact/k6/blob/51790fc8df37113ac20b586b28051e21672dab16/js/modules.go#L23-L34

You have a valid argument for easier extension->core merges, but I think this is a minor concern. Cleaning up the extension would take 1 minute of just removing the `init()` from it and adding the same constructor that was referenced in it to this map instead

core/engine.go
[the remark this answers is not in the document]
imiric: > I either have to change the `output.Output` interface and force every output to expose a `Name()`

:+1: That sounds OK to me. `Description()` is too "loose" of an interface and outputs can return whatever they want there. The description should rather be assembled from different output information so that we can control how it's rendered. Maybe even `Version()` should be added, and make it no-op for built-in outputs.

Anyway, it's a minor thing, not a big deal.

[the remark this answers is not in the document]
na--: hmm :-1: for `Version()`... It would be something that only extensions would use, and it would add unnecessary complexity for writing them. Only xk6 cares about the versions for now, and if we ever start caring, I'd rather figure out some way to extract their versions from [`runtime/debug.BuildInfo`](https://golang.org/pkg/runtime/debug/#BuildInfo) than forcing extension authors to deal with built-time linker flags or bumping an in-code `Version` constant on every release of their extension.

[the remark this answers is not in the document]
na--: about `Name()`, cc @MStoykov what do you think? I'd personally rather have a wrapper struct around `output.Output` than forcing every output to have one extra helper method.

cloudapi/config.go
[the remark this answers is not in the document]
na--: >  each output package shouldn't handle raw configuration at all,

We can debate this when we start dealing with #883, but if I correctly understand what you want, I am not sure I agree. I don't think having a central configuration clearing house is wise - why should we have one huge module that deals with every type of configuration? Why not have each component expose a unified interface for dealing with raw configuration sources and returning errors, but leave the particular configuration logic internal for every module? 

Besides, this will never work with extensions, there is no way for k6 to know how they are configured, so the only way is to to pass the "raw" values to the extension and return any resulting validation values, since it's the only thing that knows its own config.

Btw I think that the new outputs would be ideal testing grounds for potential #883 solutions. They are now self-contained, and each one receives a raw JSON chunk, a CLI argument, and a map of environment variables in their constructor. So we can start converting their configuration parsing and validation to something saner (that avoids `Apply()` and `envconfig` and all of the other issues) one by one.

core/engine.go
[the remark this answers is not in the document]
MStoykov: yeah I also think it's better if we can just use the name we already have ... otherwise I would argue that we should get the name from `Name()` everywhere.

I just realized is that this makes it ... harder to reason about if you decide to have multiple outputs of the same type .... is this even supported now?

I am not certain that was supported even before that, and I am not certain is all that needed, but seems like something that can be useful especially if outputs can filter metrics on their own. So you basically can (for example) shard based on something between multiple influxdb instances ?

output/extensions/registry.go
na--: See comment above (https://github.com/loadimpact/k6/pull/1869#discussion_r582765131), I think there is value in distinguishing between built-in and extension modules, even if they all implement the same interface.

Regarding the name, I am not sure it's the best, but this doesn't seem all that confusing: 
```go
extensions.Register("outputName", OutputConstructor)
```

I guess I can move this to the parent module (`/output`) and rename it a bit, like this:
```go
output.RegisterExtension("outputName", OutputConstructor)
```

cc @MStoykov thoughts?

cmd/outputs.go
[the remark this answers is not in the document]
imiric: I agree, but those are arguments for where `Register()` is being called, which could just as well happen in a way that avoids these anti-patterns. Where do you envision `Register()` being called from for external outputs now?

My main issue is there being a distinction between internal and external outputs, but if you think there are benefits to that, I'm OK with it.

core/engine.go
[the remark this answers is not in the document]
na--: > otherwise I would argue that we should get the name from Name() everywhere.

We can't, since `Name()` would be a method of an output instance, and we can't instantiate the output without giving it some valid parameters first. That's why we have the output type as a string in https://github.com/loadimpact/k6/blob/7e1bc5dcab910f0d197a63e843e5591de4c60423/output/extensions/registry.go#L56

The only way around that is to have 2 interfaces, the current `Output` and a new `OutputConstructor` which has a `Name() string` property and a `New(Params) Output` method. I don't mind that refactoring, but it would require every output to have these two interfaces, and likely an empty struct with the second one that is passed to the `Register()` function.

> I just realized is that this makes it ... harder to reason about if you decide to have multiple outputs of the same type .... is this even supported now?

Yes, it was supported even before this PR... The main problem with it is that JSON and environment variable configuration would be shared between all outputs, because they are based on the type names. Only the CLI flag config could be distinct. I've explored this issue and potential solution a bit before (https://github.com/loadimpact/k6/issues/587#issuecomment-416178562 and in #883), but I don't think it's a very urgent problem to solve.

cmd/outputs.go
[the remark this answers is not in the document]
imiric: :facepalm: You're right, I must've been looking at the wrong file. It's OK as is.

cmd/outputs.go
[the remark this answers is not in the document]
na--: > Where do you envision Register() being called from for external outputs now?

For extensions it would have to be in `init()` again, that unfortunately can't be easily changed due to the xk6 architecture. It's the simplest way to go about it, and is good enough for extensions. We just don't _need_ to do the same in k6, we have perfect knowledge of the built-in outputs, we can even write a simple `//go:generate` helper to generate the list for us, if the number of built-in outputs ever grows substantially.

cmd/login_cloud.go
```diff
@@ -58,22 +59,33 @@ This will set the default token used when just "k6 run -o cloud" is passed.`,
...
+			currentJSONConfigRaw := currentDiskConf.Collectors["cloud"]
+			if currentJSONConfigRaw != nil {
+				// We only want to modify this config, see comment below
+				if jsonerr := json.Unmarshal(currentJSONConfigRaw, &currentJSONConfig); jsonerr != nil {
+					return jsonerr
+				}
+			}
+
+			// We want to use this fully consolidated config for things like
+			// host addrseses, so users can overwrite them with env vars.
```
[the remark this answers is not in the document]
na--: Resolved by https://github.com/loadimpact/k6/commit/ac4a3f98155cd370e0e8be895f971aee2efc440f

output/json/wrapper.go
na--: Resolved by https://github.com/loadimpact/k6/commit/ac4a3f98155cd370e0e8be895f971aee2efc440f

cmd/ui.go
na--: Resolved by https://github.com/loadimpact/k6/commit/ac4a3f98155cd370e0e8be895f971aee2efc440f

cloudapi/config.go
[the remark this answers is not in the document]
na--: Fixed the cloud regression in https://github.com/loadimpact/k6/pull/1869/commits/626694d1dcf1bc62001a9e270675f390d415b3bb

core/engine.go
```diff
@@ -109,10 +111,47 @@ func NewEngine(
 	return e, nil
 }
 
+// StartOutputs spins up all configured outputs, giving the thresholds to any
+// that can accept them. And if some output fails, stop the already started
+// ones. This may take some time, since some outpts make initial network
```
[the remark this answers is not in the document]
na--: Resolved by https://github.com/loadimpact/k6/commit/ac4a3f98155cd370e0e8be895f971aee2efc440f

output/extensions/registry.go
MStoykov: I am finer with this being called `outputs.Register` and possibly having an `internal` packages as with the modules if we want to distinguish them. 

Although apparently @na-- is against us using the same method for both internal an external modules/outputs :man_shrugging:

na--: So you want me to change the module name from `output` to `outputs` and move `Register` there? I don't mind the second, but I'm against the rename to `outputs`

Regarding having a distinction between built-in and extension outputs, feel free to disagree in https://github.com/loadimpact/k6/pull/1869#discussion_r582765131

cmd/outputs.go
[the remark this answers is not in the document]
MStoykov: I am mostly for everything to be done the same way (in this case in `init()`) so we don't by accident break something and then for it to work for our "special" internal ones but not for the extensions. 

Introducing externally developed extensions in the base k6 by removing `init` is not as interesting IMO but it also ... kind of prevents us from introducing module/outputs by still letting them be outside the `k6` repo(and be able to be imported as external) and by just importing them and registering as "internal". Or more accurately while we can import them and register them as internal, the fact that they have `init` basically means that they will be registered as non 'internal' as well, which I don't like ;).

I think I brought this up back when xk6 was designed but there is nothing stopping us from requiring that `parh.to/extension/module/name` to have a function `NewModule() interface{}` and for xk6 to generate `additional_extensions.go` file in the root of the `k6` project that just calls the correct `Register` function based on something ... for example the name or the returned type. This even means that a single "extension" can both register a js module and an output if we go the "name" variant. 

The above suggestion can still be made and combined with #1802 and/or #1858 both of which will likely require some changes to the already written extensions. 

With the above we can completely remove the `init` from anywhere but `additional_extensions.go` 
* or just have it instead of adding additional files to add to change `main.go` for example 
* or for that matter just generate static maps in the corresponding files as in [nedomi](https://github.com/ironsmile/nedomi/tree/master/handler)

output/extensions/registry.go
MStoykov: `output` is fine ... just `outputs.Register` was more like English so this is what I wrote, but I agree that the package is better of being called `output`

core/engine.go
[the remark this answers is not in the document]
MStoykov: :+1:

cmd/outputs.go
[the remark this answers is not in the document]
na--: > we don't by accident break something and then for it to work for our "special" internal ones but not for the extensions.

I doubt we'll ever touch the extension registration after we have it set up once, and it already uses the same interfaces as internal modules, so this is mostly an theoretical concern. In any case, the proper way to ensure extension compatibility would be to to have a CI task that actually uses k6. I created https://github.com/loadimpact/k6/issues/1873

> This even means that a single "extension" can both register a js module and an output if we go the "name" variant.

I am confused, this can easily be done now :confused: A single extension repo can register however many extension types and modules it wants to. Right now xk6 essentially does a `import _ whatever` and that's all it does, I don't think it cares one bit for what every extension does. So I don't see a benefit for the extra complexity.

output/extensions/registry.go
na--: Is this better: https://github.com/loadimpact/k6/pull/1869/commits/5e41f5e8af0116ee3d5cbb048b0a1dde992e2a83 ?

MStoykov (approved):

output/json/wrapper.go
imiric: I slightly disagree, but it's not a big deal.

imiric (approved):
I still have some minor reservations about some of the design decisions here, but you've explained the reasons behind them, so no need to keep discussing them.

Issue #1681, opened 2020-10-22
Improve errors and warnings


na--:
This started as a discussion with @sniku that a lot of k6 users probably don't realize that, even if 100% of their HTTP requests fail, `k6 run their-script.js` will exit with a 0 status (i.e. "all ok") if they have not defined some thresholds. So, he suggested that we can add a warning at the end of a `k6 run`, when we show the summary, if no thresholds are defined. He even proposed to use the summary data to suggest some appropriate `http_req_duration` threshold value for them, if it makes sense. Basically have a customized warning+hint combo if we have the data (e.g. more than 100 HTTP requests) and have a default generic "no thresholds" warning for every other scenario.

All of these things seem fine and an improvement in UX, but my biggest problem is that some percentage of users (more than 10%, I think) will purposefully not want to set any thresholds and this warning might get annoying. For example, someone doing simple performance measurements or exploratory testing, or people who are just writing their script and testing it with 1VU, not the SUT. Or, on the other end, people who just use k6 to generate load and are using some other tool to monitor their SUT. Or people using external outputs. So, I think we need some way for users to suppress the warning.

Also, there are other warnings/errors we log that users might want to suppress. The most obvious example is the HTTP request failure, which you can currently only turn off by enabling `throw`... And if we have the possibility to disable warnings, we probably would also want to use them for other things as well! Some ideas:
- when we have shared memory (https://github.com/loadimpact/k6/issues/532) or streaming data (https://github.com/loadimpact/k6/issues/592), we should probably start emitting warnings when users `open()` large files.
- depends on when the new HTTP API lands, but a lot of people don't know about `discardResponseBodies` and `responseType`, so we might nudge them with a warning if they have large HTTP responses or something like that
- suggest thresholds for other things, like failed or interrupted iterations
- etc. other good practices that aren't followed

So, I think we should:
- Give each error and warning a unique SEO-friendly ID, e.g. `k6NoThresholdsDefined`, `k6OpenLargeFile`, etc.
- Allow users to explicitly disable specific warning/error messages with a new k6 option, e.g. `options.ignoredWarnings = ['k6NoThresholdsDefined']` (any ideas for a better name?).
- This option will probably also be re-used for cloud-only warnings/errors/smart-result hints, so we should simply build a set of ignored string IDs without validating it on the k6 side.
[...]

PR #1876, opened 2021-03-02
Basic check that metric.Add doesn't let NaNs through

MStoykov:


js/modules/k6/metrics/metrics_test.go
```diff
@@ -43,13 +43,19 @@ func TestMetrics(t *testing.T) {
...
-		"False": {`false`, 0.0},
+		"Float":     {JS: `2.5`, Float: 2.5},
+		"Int":       {JS: `5`, Float: 5.0},
+		"True":      {JS: `true`, Float: 1.0},
+		"False":     {JS: `false`, Float: 0.0},
+		"null":      {JS: `null`, isError: true},
+		"undefined": {JS: `undefined`, isError: true},
+		"NaN":       {JS: `NaN`, isError: true},
+		"string":    {JS: `"string"`, isError: true},
+		"string 5":  {JS: `"5.3"`, Float: 5.3},
```
MStoykov, line 58: This also fails with adding another metric object, but it looks like it will require some rewriting so I have skipped it for now

js/modules/k6/metrics/metrics.go
```diff
@@ -52,14 +53,17 @@ type Metric struct {
 }
 
 // ErrMetricsAddInInitContext is error returned when adding to metric is done in the init context
-var ErrMetricsAddInInitContext = common.NewInitContextError("Adding to metrics in the init context is not supported")
+var (
+	ErrMetricsAddInInitContext = common.NewInitContextError("Adding to metrics in the init context is not supported")
```
sniku, line 57: Raising exceptions that interrupt the execution makes the tool more fragile. 
Errors are expected when doing stress testing, so I think raising exceptions is not a good default because it will condition users to wrap code in `try/catch`. 
I think k6 should be resilient to errors, as I argued in  https://github.com/loadimpact/k6/issues/1435

MStoykov: Given that:

1. this will mean that they will just be getting spammed by log messages until #1681, without a way to stop them
2. this is practically only reported by you, so I would argue it is fairly rare? :thinking: and as such IMO an actual "exception" :wink: :wink:   

I think this should throw an exception with a stacktrace instead of just logging  `Adding to metrics in the init context is not supported` and letting the user guess where that is and what should they do about it. 

I guess we can add the stacktrace to the warning but that IMO will be even noisier.

PR #1877, opened 2021-03-02
Get the env for the VU back from the __ENV variable

### LINKED ISSUE #1722: __ENV doesn't keep changes between init context and VU context in VU code
```
if (typeof __ENV.something == "undefined")  {
 __ENV.something= "test";
}
export default function () {
    console.log(__ENV.something);
}
```
This code should print "test" and it did up to https://github.com/loadimpact/k6/commit/e1a5c0af001977b6a10945791f5b16288937d391 and if the golang version is go1.13.8 but not if it is 1.15.3 (or 1.14.6) 
image
image

This is fairly common pattern to fix your __ENV variables in cases where you didn't provide them but you use them through the script.

MStoykov:
fixes #1722

### DIFF
--- a/js/bundle.go
+++ b/js/bundle.go
@@ -246,7 +246,7 @@ func (b *Bundle) Instantiate(logger logrus.FieldLogger, vuID int64) (bi *BundleI
 		Runtime: rt,
 		Context: ctxPtr,
 		exports: make(map[string]goja.Callable),
-		env:     b.RuntimeOptions.Env,
+		env:     rt.Get("__ENV").Export().(map[string]string),
 	}
 
 	// Grab any exported functions that could be executed. These were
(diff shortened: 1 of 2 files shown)


js/bundle.go
```diff
@@ -246,7 +246,7 @@ func (b *Bundle) Instantiate(logger logrus.FieldLogger, vuID int64) (bi *BundleI
 		Runtime: rt,
 		Context: ctxPtr,
 		exports: make(map[string]goja.Callable),
-		env:     b.RuntimeOptions.Env,
+		env:     rt.Get("__ENV").Export().(map[string]string),
```
MStoykov, line 249: It will be better if we don't have to cache it back and again ... but for that I will need to get it somehow out of `b.instantiate`

which will be a lot bigger change IMO.

Given that we set it *always* there is no chance for this to panic unless someone changes what `__ENV` is ... :(  Which should probably not be possible, to begin with.

That will probably require making it nonconfigurable/writable and then only changing the object - again a lot bigger change, but arguably way more accurate.

Maybe I should just check that it is `map[string]string` and set it instead of `b.RuntimeOptions.Env` for this hotfix?

PR #1885, opened 2021-03-05
More tests

na--:
This should close https://github.com/loadimpact/k6/issues/1873 and adds the missing test from https://github.com/loadimpact/k6/pull/1882. This can also be considered a somewhat tentative first step in https://github.com/loadimpact/k6/issues/342

### DIFF
--- a/.github/workflows/xk6.yml
+++ b/.github/workflows/xk6.yml
@@ -15,8 +15,8 @@ jobs:
   test-xk6:
     strategy:
       matrix:
-        go-version: [1.14.x, 1.15.x]
-        platform: [ubuntu-latest, windows-latest]
+        go-version: [1.15.x]
+        platform: [ubuntu-latest, windows-latest, macos-latest]
     runs-on: ubuntu-latest
     steps:
       - name: Checkout code
--- a/.github/workflows/xk6.yml
+++ b/.github/workflows/xk6.yml
@@ -1,0 +1,57 @@
+name: xk6
+on:
+  # Enable manually triggering this workflow via the API or web UI
+  workflow_dispatch:
+  push:
+    branches:
+      - master
+  pull_request:
+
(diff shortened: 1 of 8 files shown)


.github/workflows/xk6.yml
```diff
@@ -0,0 +1,57 @@
...
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  test-xk6:
+    strategy:
+      matrix:
+        go-version: [1.14.x, 1.15.x]
+        platform: [ubuntu-latest, windows-latest]
```
MStoykov, line 19: not certain it matters all that much to test with the previous version as well, but maybe test it with macos ?

MStoykov (dismissed):
LGTM

Maybe move the test extensions in the `samples` ?

.github/workflows/xk6.yml
na--: hmm given https://github.com/k6io/xk6/issues/9, we should probably test it with the newer Go versions instead, but let's switch after we fix that 1.16 issue and release k6 v0.31.0

na--: but yeah, :+1: for macos, I'll add it

na--:
I thought about `samples`, but these extensions aren't a good example for potential extension authors... so I'd prefer it is very clear that they are for testing purposes only

.github/workflows/xk6.yml
na--: Resolved by https://github.com/loadimpact/k6/pull/1885/commits/6dfbcc30d09fc53d9428a6f81e81a2b61a62466e

MStoykov (approved):

imiric (approved):
LGTM

And we want to run this on every commit, right? I suppose it's fine, and it runs quick enough.

PR #1888, opened 2021-03-08
Fix metrics data race in the Engine test run finalization

na--:
There was a race between these two branches of the `select`: https://github.com/loadimpact/k6/blob/f1413e601b021d19f06d4d72ee14714a46344162/core/engine.go#L330-L339

I didn't add a test, since apparently the xk6 one I added in https://github.com/loadimpact/k6/pull/1885 already covers it - yay for integration tests... :tada:  :sweat_smile: This should close https://github.com/loadimpact/k6/issues/1887

### DIFF
--- a/core/engine.go
+++ b/core/engine.go
@@ -328,6 +328,15 @@ func (e *Engine) processMetrics(globalCtx context.Context, processMetricsAfterRu
 		case <-ticker.C:
 			processSamples()
 		case <-processMetricsAfterRun:
+		getCachedMetrics:
+			for {
+				select {
+				case sc := <-e.Samples:
+					sampleContainers = append(sampleContainers, sc)
+				default:
+					break getCachedMetrics
+				}
+			}
 			e.logger.Debug("Processing metrics and thresholds after the test run has ended...")
 			processSamples()
 			if !e.runtimeOptions.NoThresholds.Bool {


core/engine.go
```diff
@@ -328,6 +328,15 @@ func (e *Engine) processMetrics(globalCtx context.Context, processMetricsAfterRu
...
 		case <-processMetricsAfterRun:
+		getCachedMetrics:
+			for {
+				select {
+				case sc := <-e.Samples:
+					sampleContainers = append(sampleContainers, sc)
+				default:
+					break getCachedMetrics
+				}
+			}
```
MStoykov, line 339: From the looks of things, this should be called after all the VUs have stopped ... so at least it should not be possible for there to be an endless loop as something keeps adding samples.

Obviously, this won't prevent some extension or badly written internal code to just keep sending metrics without checking the context in a separate goroutine, but I don't think that is such a big problem.

I somewhat would prefer for there to be some timeout to this still, just in case we are wrong or something changes in the future. But I think this can wait

MStoykov (approved):

imiric (approved):
I can't think of a more elegant way to flush `e.Samples`, though I agree with Mihail that a timeout would be good to have.

core/engine.go
na--: I am not convinced a timeout is a good idea here... It shouldn't be needed, and we don't have timeouts for other things that process metrics and might be overwhelmed, so this probably shouldn't have one either. That said, I added this issue and mentioned the potential problem in it: https://github.com/loadimpact/k6/issues/1889

but for now, unless we observe some issues stemming from this, I'm for not touching it :sweat_smile:

Issue #994, opened 2019-04-09
Provide decent stack traces in setup code


neerolyte:
When an Error is thrown in setup it'd be nice if an exception was thrown, currently we just get the current function, file and line (which often seems to be wrong?). E.g:

```
$ cat foo.js
import bar from "./bar.js";
export function setup() {
        bar();
};
export default function() {
};
$ cat bar.js 
export default function () {
        baz();
}
function baz() {
        throw new Error("baz");
}
$ docker run -it --rm -v "$PWD:/src" -w /src loadimpact/k6 run foo.js

          /\      |‾‾|  /‾‾/  /‾/   
     /\  /  \     |  |_/  /  / /    
    /  \/    \    |      |  /  ‾‾\  
   /          \   |  |‾\  \ | (_) | 
  / __________ \  |__|  \__\ \___/ .io

  execution: local
     output: -
     script: foo.js

    duration: -,  iterations: 1
         vus: 1, max: 1

ERRO[0000] Engine error                                  error="setup: Error: baz at baz (/src/bar.js:6:8(4))"
ERRO[0000] Engine Error 
```

cuonglm:
This seems to be the same underlying issue with https://github.com/loadimpact/k6/issues/702.

> currently we just get the current function, file and line (which often seems to be wrong?)

The error message is display, `baz at baz` comes from `throw new Error("baz");`.

MStoykov:
The good stacktrace is now produced by this lines https://github.com/loadimpact/k6/blob/8f4797d2761560fefb3e01c1bf440ee6a1c55548/lib/executor/helpers.go#L98-L107 Which basically check "badly" that something is goja.Exception.

I would argue that so that we don't get goja in `lib/...` even more we have another error type that the runner wrap ~errors~ exceptions in before returning them as otherwise we just hope that whatever we get and has String is actually goja.Exception ... https://github.com/dop251/goja/blob/e21ccf39bfce/runtime.go#L301

PR #1895, opened 2021-03-10
quick fix for #994

MStoykov:


MStoykov:
The more correct way to do this will be to wrap the [goja.Exception](https://pkg.go.dev/github.com/dop251/goja?utm_source=godoc#Exception) in some lib.Exception interface that we check for both at this place and at the https://github.com/loadimpact/k6/blob/ef717e1af2a92d35d39a3d815a69b6ab6b8c65e1/lib/executor/helpers.go#L99-L103 ? Or maybe I should just move this wrapping I do in this PR in the js/runner and wrap all Exceptions like that ?

imiric:
I'm not sure what the proper fix for this should be, but the current change shows the same error with the example in #994 as reported there, and as shown on `master`. Isn't it the idea that it should show a full stack trace instead?

MStoykov:
> I'm not sure what the proper fix for this should be, but the current change shows the same error with the example in #994 as reported there, and as shown on master. Isn't it the idea that it should show a full stack trace instead?

I was fixing it for cases as https://community.k6.io/t/goerror-invalid-url-value-nil/1491/8 which seemed like it will work, but apparently the code I changed in 782d775 wraps the cases coming from setup/teardown ;(

MStoykov:
superseded by #1971

PR #1897, opened 2021-03-10
Drop the SSLv3 support

### LINKED ISSUE #1083: Drop or deprecate SSLv3 support
[go 1.13 will deprecate SSLv3 support and it will be removed in go1.14](https://github.com/golang/go/commit/a6a7b148f874b32a34e833893971b471cd9cdeb7)

Should we do the same?

codebien:
SSLv3 has been removed from the supported version list of the crypto protocols.

Closes #1083


na-- (approved):
Thanks!

MStoykov (approved):

PR #1903, opened 2021-03-14
Auto marshal to JSON

### LINKED ISSUE #878: Automatically marshal JS-object HTTP request bodies when content-type is application/json
If we have this example k6 script:
```js
import http from "k6/http";

export default function () {
    let data = {
        arr: [1, 2, 3, "a"],
        str: "test",
        obj: { a: "b", c: 4 },
    };
    let resp = http.post("https://httpbin.org/post", data, { headers: { "content-type": "application/json" } });
    console.log(resp.json().data);
}
```
Most people will expect it to print something like `{"arr":[1,2,3,"a"],"str":"test","obj":{"a":"b","c":4}}`, or to print a warning, or even fail with an error.

Instead, k6 will silently do a stupid thing and actually assume that the content-type is `application/x-www-form-urlencoded`, and thus try to urlencode the supplied object. Even that may have been fine, if there was a standard in for URL-encoding nested objects. But as far as I can see, there isn't an RFC for this, and and [only PHP supports string keys](https://stackoverflow.com/questions/30261513/is-there-a-spec-for-nested-urlencoded-params) in maps anyway. In the above example, k6 will actually print `arr=%5B1+2+3+a%5D&obj=map%5Ba%3Ab+c%3A4%5D&str=test`. Basically converts JS arrays and objects to strings and then just tacks them on, which is pretty much useless to everyone. To get the expected result, users have to manually pass `JSON.stringify(data)` instead of simply `data`.

So my suggestion is to specifically check if a request's body isn't a string or a binary payload and if its `content-type` is `application/json` and just automatically marshal the request body to JSON.

codebien:
Auto encode to JSON the body when the Content-Type header is set to `application/json` 

Closes #878

### DIFF
--- a/js/modules/k6/http/request.go
+++ b/js/modules/k6/http/request.go
@@ -23,6 +23,7 @@ package http
 import (
 	"bytes"
 	"context"
+	"encoding/json"
 	"fmt"
 	"mime/multipart"
 	"net/http"
@@ -289,7 +290,9 @@ func (h *HTTP) parseRequest(
 				bodyQuery.Set(k, formatFormVal(v))
 			}
 			result.Body = bytes.NewBufferString(bodyQuery.Encode())
-			result.Req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
+			if result.Req.Header.Get("Content-Type") == "" {
+				result.Req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
+			}
 			return nil
 		}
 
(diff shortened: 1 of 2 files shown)


na-- (changes requested):
Sorry, as I explained in https://github.com/loadimpact/k6/issues/878#issuecomment-799367830, we're probably not going to merge this as it is. We don't want to automatically marshal objects to JSON based on a request header.

That said, I skimmed the code and I like the refactoring of removing the `formatFormVal` and `handleObjectBody` lambdas out of the method and into their own methods, so we might merge just that, without any logic changes.

PR #1904, opened 2021-03-15
update goja to latest master

MStoykov:


MStoykov:
From my investigation, the new failures around `test/language/arguments-object/mapped/nonconfigurable-descriptors-basic.js` and co. are about goja not doing something properly in non-strict mode with `arguments`. 

It does work perfectly fine in strict mode and as previously the `let` meant that babel was used which *adds* "use strict" at the beginning of the script it was ... working. 

Given that this is highly unlikely to be relevant to any actual test script and that  k6 also *always* runs in strict mode as well ... I don't think this is a problem.

I will try to figure out if I can propose a fix for some more time and if not will just make an issue in the goja repo

MStoykov:
Made a PR https://github.com/dop251/goja/pull/266, again I don't think we need to wait on this

js/runner_test.go
```diff
@@ -1819,7 +1819,9 @@ func TestVUPanic(t *testing.T) {
 			require.Len(t, entries, 1)
 			assert.Equal(t, logrus.ErrorLevel, entries[0].Level)
 			require.True(t, strings.HasPrefix(entries[0].Message, "panic: here we panic"))
-			require.True(t, strings.HasSuffix(entries[0].Message, "Goja stack:\nfile:///script.js:3:4(12)"))
+			// broken since goja@f3cfc97811c0b4d8337902c3e42fb2371ba1d524 see
+			// https://github.com/dop251/goja/issues/179#issuecomment-783572020
+			// require.True(t, strings.HasSuffix(entries[0].Message, "Goja stack:\nfile:///script.js:3:4(12)"))
```
imiric, line 1824: Are we OK with this? Is this a breaking change then? Should we test with the re-throwing workaround suggested in that comment?

MStoykov: I made an issue about discussing that https://github.com/loadimpact/k6/issues/1906. 

As I mention there I don't think that we have ever actually used this functionality really and the "workaround" is both error prone IMO and will need a lot of work. 

Regardless I am of the opinion that we can have it not work for some time and either fix it a later PR or just leave it until we actually need it again, and figure out what to do then.

imiric: OK. Hopefully not a lot of users were relying on this, but like you mentioned, it wasn't very useful to begin with.

imiric (approved):

imiric (approved):

js/runner_test.go
MStoykov: At least the current behaviour was *not* helpful as combined with the lack of anything but the last frame and the  fact that even that line number is usually wrong I spent an hour or two debugging the wrong place before deciding to ignore the "goja stacktrace" and just go with logic and the golang stacktrace

imiric (approved):

na-- (approved):
LGTM :crossed_fingers: :sweat_smile:

PR #1911, opened 2021-03-17
Add GetJSModules support from #1858

MStoykov:
This allows integration tests to have separate global module instance of
all internal js modules.

For this to be supported for extension an additional type/interface will
be required a bit more added code


MStoykov:
at this point I think that `k6/js/internal/modules` can be dropped and all of it to be moved to `k6/js/modules`

js/modules/k6/data/data.go
```diff
@@ -23,21 +23,50 @@ package data
...
+}
+
+func (s *sharedArrays) get(rt *goja.Runtime, name string, call goja.Callable) sharedArray {
+	s.mu.RLock()
+	array, ok := s.data[name]
+	s.mu.RUnlock()
+	if !ok {
+		s.mu.Lock()
+		array, ok = s.data[name]
+		if !ok {
```
imiric, line 50: Why check for this twice? If it doesn't exist the first time couldn't you just lock and add it?

Also I'm not sure why you need the `func()` and `defer` below. Why not just call `Unlock()` after `s.data[name] = array`?

js/internal/modules/modules.go
```diff
@@ -61,3 +61,42 @@ func Register(name string, mod interface{}) {
...
+			result[name] = perInstance
+		} else {
+			result[name] = HasModuleInstancePerVUDummyWrapper{Module: module}
+		}
+	}
+	return result
+}
+
+// HasModuleInstancePerVUDummyWrapper is a wrapper to be used around an module that doesn't support
+// HasModuleIntancePerVU, but needs to be used in places where it is required.
```
imiric, line 92:
```suggestion
// HasModuleInstancePerVU, but needs to be used in places where it is required.
```

js/modules/k6/data/data.go
MStoykov: First; I am not certain the RWLock is needed at all as I doubt it speeds up anything.

The second check is because the first check only happens with the read lock and it needs to be unlocked in order to get the write lock to add it. So it is possible for multiple VUs to get to that point and one to writelock, execute the callback, go to write it, unlock the write lock and then the second VU to come by and do the same all over again effectively executing the callback twice. Which is exactly what we *don't* want to happen.

The defer is needed as there could be a panic/goja exception in the callback code and we don't want to lock the whole `sharedArrays` forever just because of that.

imiric (dismissed):

js/internal/modules/modules.go
```diff
@@ -61,3 +61,42 @@ func Register(name string, mod interface{}) {
 	}
 	modules[name] = mod
 }
+
+// GetJSModules returns a map of all js modules
+func GetJSModules() map[string]HasModuleInstancePerVU {
```
na--, line 66: I don't understand why we need to wrap all modules artificially in a `HasModuleInstancePerVU` implementation, when they explicitly don't have a different instance for each VU :confused: In fact, I think this adds confusion without actually adding any benefits... 

The only "benefit" I can see is that is that we don't use `map[string]interface{}` here, but rather hide the "nasty" `interface{}` behind a useless `HasModuleInstancePerVUDummyWrapper`, which seems pointless.

js/initcontext.go
```diff
@@ -140,11 +140,11 @@ func (i *InitContext) Require(arg string) goja.Value {
...
 
 func (i *InitContext) requireModule(name string) (goja.Value, error) {
-	mod := modules.Get(name)
-	if mod == nil {
+	mod, ok := i.modules[name]
+	if !ok {
 		return nil, errors.Errorf("unknown module: %s", name)
 	}
-	return i.runtime.ToValue(common.Bind(i.runtime, mod, i.ctxPtr)), nil
+	return i.runtime.ToValue(common.Bind(i.runtime, mod.NewModuleInstancePerVU(), i.ctxPtr)), nil
```
na--, line 147: Connected to my other comment, what's the problem of keeping this logic here?

```suggestion
	if i, ok := mod.(HasModuleInstancePerVU); ok {
		mod = i.NewModuleInstancePerVU()
	}

	return i.runtime.ToValue(common.Bind(i.runtime, mod, i.ctxPtr)), nil
```

js/internal/modules/modules.go
MStoykov: My reason was that GetJSModules can be used by other things beside the current use in `requireModule` and in all of those cases this will need to be checked for each module. 

I also expected that such a thing will happen while I was doing the rest of the code changes, but at least at the time of writing apparently I didn't need to use it directly (or at least nowhere where this changes broke the code was this necessary, and I haven't gone through all the tests to see if a good use of this can be found there). 

Given that I am fine with reverting to the less types variant

na--: If it happens, we can always make a short helper function that handles it, so I'd definitely prefer if we don't have a mandatory `HasModuleInstancePerVUDummyWrapper` when we ain't gonna need it :smile:

na-- (changes requested):
I saw that you requested another review, has something changed and you're now against doing [this](https://github.com/k6io/k6/pull/1911/files#r606212218) change? 

Also, can you please rebase on top of the latest `master`, we've had a lot of changes since this PR

js/internal/modules/modules.go
```diff
@@ -61,3 +61,25 @@ func Register(name string, mod interface{}) {
 	}
```
MStoykov, line 61: The check above here is done only against the modules that are *registered* through the *Register* call. 

While the current usage is such they are required to have `x/` in the public `modules` module, it still isn't *technically* correct.

So should I:
1. do nothing
2. drop the internal `modules` package and move everything to the public one. This way I will check it there and hope that nobody will add `x/k6` standard lib.
3. Actually check against the result of `GetJSModules`

imiric: Hhmm considering we only made the package internal to begin with was to distinguish between internal and external module registration, and internal modules don't call `Register()` anymore, I vote for the 2nd approach, and to do any additional checks in the public package.

js/init_and_modules_test.go
```diff
@@ -60,6 +60,7 @@ func (cm *CheckModule) VuCtx(ctx context.Context) {
 }
 
 func TestNewJSRunnerWithCustomModule(t *testing.T) {
+	t.Skip()
```
imiric, line 63: Does this still have to be skipped? If so, let's add a reason for it.

na--: IIRC, this was because the test broke when it was ran with `--count 2`, since it tried to register the same module twice. @MStoykov, can we simply register it once with a `sync.Once()`? Or, alternatively, make the test package to be `js` instead of `js_test` and directly mess with the `modules` map in the `InitContext` of the `Bundle`, instead of using `modules.Register()` : https://github.com/k6io/k6/blob/8939092c6c521543eb727202d900ddef07c0511b/js/initcontext.go#L73

na-- (commented):
Mostly LGTM, besides the minor nitpicks mentioned inline and another more global one. 

I think @simskij mentioned it somewhere, but we should probably move the `HasModuleInstancePerVU` interface out of the `internal` package, so that extensions can also explicitly check that they implement it. Not a deal breaker, we can do that a few versions from now and keep it in `internal` in v0.32.0, they can still implement it by just having the appropriate method. But if we don't intend to make any more changes, we probably can make it public even now.

js/init_and_modules_test.go
MStoykov: How about I have it have a random-ish name ? ... the only problem is that with `count=10000` it will probably fill the map ... but :shrug:

na--: Eh, :+1: that would work, though I'd prefer you have an `atomic`-ally incremented name instead of a random one, just so we don't have another slightly flaky test that fails once in a blue moon :sweat_smile:

js/modules/k6/http/http.go
```diff
@@ -46,11 +41,14 @@ const (
...
 
+// New returns a new global module instance
+func New() *GlobalHTTP {
+	return &GlobalHTTP{}
+}
+
 // GlobalHTTP is a global HTTP module for a k6 instance/test run
 type GlobalHTTP struct{}
 
-var _ modules.HasModuleInstancePerVU = new(GlobalHTTP)
```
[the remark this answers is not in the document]
MStoykov: well that will mean there will be an import loop ;)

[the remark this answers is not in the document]
MStoykov: should I put in the `modules` package ?

MStoykov:
should I:
- [ ] move everything to the not *internal* modules package 
- [ ] just have it check in the modules package that some js modules are `HasModuleInstancePerVU` as to not make cycles ? I can move the definition somewhere else but this seems pointless for the internal packages? Alternatively I can just move the `HasModuleInstancePerVU` in the public `modules` package and import it from there ... that should not have any cycles I think :thinking:

js/internal/modules/modules.go
na--: Not sure I grok everything fully, but yeah, 2nd approach seems best :+1:, we don't need an `internal` module anymore.

na-- (approved):
LGTM, though yeah, it seems like we don't really need the `internal/modules` package anymore

na-- (approved):

imiric (approved):

PR #1916, opened 2021-03-18
Add package repository scripts, run in CI (staging PoC)

imiric:
This is part of the effort to [abandon Bintray](https://jfrog.com/blog/into-the-sunset-bintray-jcenter-gocenter-and-chartcenter/) in favor of a self-hosted package repository solution. The current approach involves pushing the packages and repository metadata to an S3 bucket, which is served from a CDN (CloudFront).

You can explore the current test bucket at: https://dl.staging.k6.io/, and test it on Debian/etc. with:

```shell
curl -fsSL https://dl.staging.k6.io/key.gpg | apt-key add -
echo "deb https://dl.staging.k6.io/deb stable main" > /etc/apt/sources.list.d/k6.list
apt-get update && apt-get install k6
```

... and Fedora/etc.:
```shell
cat > /etc/yum.repos.d/k6.repo <<EOF
[k6]
name=k6
baseurl=https://dl.staging.k6.io/rpm/x86_64
enabled=1
gpgcheck=1
gpgkey=https://dl.staging.k6.io/key.gpg
EOF
yum install k6
```

Pending:
- [x] ~Create infrastructure with Terraform~
- [ ] ~Create NuGet/Chocolatey repo~ As discussed, we don't need to move the `.nupkg` building and publishing since the Bintray repository isn't being used for the [Chocolatey package](https://github.com/majkinetor/au-packages/tree/master/k6). We will however need to add this when we takeover the maintenance of the package.
- [x] ~Add `k6-latest-amd64.msi`. I wonder if we can do this with some S3 redirects, otherwise just duplicating the file would work as well.~ Redirects are aggressively cached by CloudFront and are not an option, so the latest file is duplicated.
- [x] ~Publish the `k6io/packager` image to GHCR to speed up the `publish-packages` step and avoid rebuilding it everytime.~ A separate CI job was added to publish the image manually and on a weekly schedule (to ensure building keeps working).

Part of #1247


MStoykov:
> I'm interested if the yum upgrade workflow works correctly, since I had to run yum clean expire-cache after a release for it to pickup the new version. So feedback from Fedora users is welcome :)

The default `metadata_expire` is 6 hours apparently. [man yum.conf](https://linux.die.net/man/5/yum.conf) say that it means that it would not try to update metadata that isn't at least 6 hours old. So it seems like this is likely working as expected.

MStoykov (commented):
Seems okay to me. 

Maybe we should merge this earlier in the cycle and build the repo for the current version and update the docs? 

I don't think any of the TODOs are a must, and the two I am most interested in is the `debsign` and the md5sum checks which arguably can be done with s3cmd which we need either way so I don't know why it wasn't just used instead of using `curl` ?

imiric:
I'm wondering if we should bother with signing .deb packages with `debsigs` (`debsign` actually signs .changes and .dcs files) or `dpkg-sig`. See https://blog.packagecloud.io/eng/2014/10/28/howto-gpg-sign-verify-deb-packages-apt-repositories/.

Verification is not enabled by default and the user needs to manually enable it. Even then it's not built into the `apt-get install` workflow, so they'd either have to check it manually with `debsig-verify` or enable it globally and risk "causing dpkg to reject installation of unsigned Debian files", which can just cause more problems than it's worth.

So it's probably better to stick with the current approach done in Bintray to just have a separate .asc file and the user can verify it manually with `gpg` if they want to.

RPMs are properly signed and verified automatically (if `gpgcheck` is enabled for the repo), so `yum` does something right... :smile: 

Re: md5sum checks, I'd also prefer if we just downloaded it with `s3cmd` and let it check automatically. I used `curl` because we discussed avoiding egress S3 costs, but considering we'll do releases every couple of months (or a bit more frequent for patch releases), downloading ~200MB(?) each time doesn't seem so bad. If we setup the automatic deletion of older versions, which we probably should, then these should be fixed costs known upfront. So yeah, I think I'll just use `s3cmd` for that.

MStoykov:
Yeah .. I remember that deb and signing packages was something ... strange ... But I think they check the md5sums and and md5sums were signed ? Is this the `asc`? Either way we can do this later.

If moving to `s3cmd` is easy, I am for it doing it now, even if we end up needing to move to curl later.

lskillen:
Have you considered something like @cloudsmith-io which does all of this for you? :-) (I work there, happy to help!)

imiric:
@lskillen Thanks! We considered some hosted services, including Cloudsmith, but ultimately decided it was best to roll our own and maintain control over the release process and package distribution. We don't want to repeat the migration issues we're dealing with now with Bintray, and setting up and maintaining this infrastructure ourselves is not an insurmountable amount of work.

lskillen:
I understand, completely. It's unfortunate to be in a situation that requires such reactivity. 🤕 We'll not be going anywhere, so if you need help with that in the future we'll be there for you. If you did feel nervous about lock-in later, we offer custom domains to put the control back in your court. All the best of luck with the self-host! 😁👍

packaging/bin/create-rpm-repo.sh
```diff
@@ -0,0 +1,74 @@
...
+sync_to_s3() {
+  log "Syncing to S3 ..."
+  s3cmd sync --delete-removed "${REPODIR}/" "s3://${S3PATH}/"
+
+  # Disable cache for repo metadata and index files, so that new releases will be
+  # available immediately.
+  s3cmd --recursive modify --add-header="Cache-Control:no-cache, max-age=0" \
+    "s3://${S3PATH}/x86_64/repodata"
+  s3cmd modify --recursive --exclude='*' --include='index.html' \
+    --add-header='Cache-Control:no-cache, max-age=0' "s3://${S3PATH}/"
```
MStoykov, line 42: Does this work on all `index.html`?

MStoykov (commented):
Looks okay to me

I would prefer if we `--cf-invalidate` the "index/packages" files instead of just them not having any caching as they are likely to be hitted all the time and given that we update usually once every 2 months I would argue it is probably more beneficial if they have 1hour caching then no caching
- again dropping old files IMO is low priority and doing it based on number of packages is IMO wrong (as we can have a situation where we need to release multiple fix releases). So IMO it will be better to not do it for now


I also wonder if it won't be better if all the "samely" named function have a prefix/suffix as just reading it it sounds like they will do the same thing but they are all specific (more or less0

packaging/bin/create-msi-repo.sh
```diff
@@ -0,0 +1,54 @@
...
+# Number of packages to keep, older packages will be deleted.
+RETAIN_PKG_COUNT=25
+
+log() {
+    echo "$(date -Iseconds) $*"
+}
+
+delete_old_pkgs() {
+  find "$1" -name '*.msi' -type f | sort -r | tail -n "+$((RETAIN_PKG_COUNT+1))" | xargs -r rm -v
+}
```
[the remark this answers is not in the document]
imiric: Sure, we could drop it for now. My main motivation was that if we don't have some cleanup in place the time and costs will keep increasing for every release. s3cmd does preserve `mtime` and we could filter by that. But OK, I'll remove this for now and we can add cleanup later.

packaging/bin/create-rpm-repo.sh
imiric: Yes, on all `index.html` within the RPM repo.

packaging/bin/create-msi-repo.sh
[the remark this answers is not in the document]
MStoykov: I am fine with keeping it, just with `mtime` instead of just a count ... But also I don't think it matters that much so :shrug: 

WDYT @na-- , you wanted to keep packages forever

packaging/bin/create-deb-repo.sh
```diff
@@ -0,0 +1,109 @@
...
+  s3cmd sync --delete-removed "${REPODIR}/" "s3://${S3PATH}/"
+
+  # Disable cache for repo metadata and index files, so that new releases will be
+  # available immediately.
+  s3cmd modify --add-header="Cache-Control:no-cache, max-age=0" \
+    "s3://${S3PATH}/dists/stable/"{Release,Release.gpg,InRelease}
+  s3cmd modify --add-header="Cache-Control:no-cache, max-age=0" \
+    "s3://${S3PATH}/dists/stable/main/binary-amd64"/Packages{,.gz,.bz2}
+  s3cmd modify --recursive --exclude='*' --include='index.html' \
+    --add-header='Cache-Control:no-cache, max-age=0' "s3://${S3PATH}/"
```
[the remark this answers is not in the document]
imiric: So this is not that simple :( Passing `--cf-invalidate` instead actually does nothing and I still get hits from CF. It seems this option requires some additional IAM permissions (`cloudfront:CreateInvalidation` at the very least), but there's no s3cmd documentation about it, I can't see anything useful with `-v --debug` (I would at least expect it to fail...), and I can't find anything in the repo issues, except old issues claiming that it doesn't work (536, 790).

s3cmd seemed like the simplest tool to use that's still decently maintained, but I could switch to the official `aws-cli` for this which has a standalone `aws cloudfront` command. For now I'll keep the `Cache-Control` header, even though it means these files will never be cached, which should be fine since they're small index and metadata files.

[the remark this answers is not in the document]
MStoykov: :sob:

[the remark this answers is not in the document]
MStoykov: I think we can rethink,refactor this after a month or two if it turns out that it is a problem or is costing too much money so ... just drop it for now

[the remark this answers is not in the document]
imiric: So, good news... I switched to awscli instead of s3cmd, and CF invalidation works with `aws cloudfront`. I still have some work to do to add the permission in the IAM policy with Terraform (currently I just added it manually), and to test the latest MSI with S3 redirects again since I wasn't able to get it working with `s3cmd` and ended up duplicating the file, but this seems much better than setting `no-cache` for these files. I'll push the changes here once I test them thoroughly in my fork.

Thanks for the suggestion!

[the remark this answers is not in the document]
imiric: See the latest changes with the awscli switch. We still need to set a short 60s `max-age` for the index files to avoid browsers caching them, and I did the same for the repo metadata files just in case. But now they're properly invalidated in CloudFront.

Also I couldn't get the S3 redirect to work for the latest MSI package, as it would always be cached by CloudFront for some reason. So I settled on just copying the actual latest file instead.

packaging/bin/create-deb-repo.sh
```diff
@@ -0,0 +1,116 @@
...
+    --exclude='*.deb' --exclude='*.asc' --exclude='*.html' \
+    --cache-control='max-age=60,must-revalidate' \
+    --metadata-directive=REPLACE \
+    "s3://${S3PATH}" "s3://${S3PATH}"
+  # Set it separately for HTML files to set the correct Content-Type.
+  aws s3 cp --no-progress --recursive \
+    --exclude='*' --include='*.html' \
+    --content-type='text/html' \
+    --cache-control='max-age=60,must-revalidate' \
+    --metadata-directive=REPLACE \
```
MStoykov, line 55: what does `Note that if the object is copied over in parts,` from the documentation of this option means ?

I can't find an explanation so I wonder if this is just something that `aws s3 cp` won't do on it's own? or is it something that has a flag that is just badly documented and me searching ... fails

packaging/bin/entrypoint.sh
```diff
@@ -0,0 +1,55 @@
...
+# Generate and sync the main index.html
+(cd "$pkgdir" && generate_index.py)
+aws s3 cp --no-progress --cache-control='max-age=60,must-revalidate' \
+  "${pkgdir}/index.html" "s3://${s3bucket}/index.html"
+# Also sync the GPG key
+aws s3 cp --no-progress "${pkgdir}/key.gpg" "s3://${s3bucket}/key.gpg"
+
+# Invalidate CloudFront cache for index files, repo metadata and the latest MSI
+# package redirect.
+# TODO: Maybe just invalidate the entire bucket (/*), since CF charges per
```
MStoykov, line 45: seems like 1k free paths are included https://aws.amazon.com/about-aws/whats-new/2010/08/31/cloudfront-adds-invalidation-feature/

So it is unlikely IMO for us to go *over* that so it is probably better if we cache better ;)

MStoykov (approved):
LGTM! As long as you say it works I am fine with it, but I ain't testing this before you release it :P

packaging/bin/create-deb-repo.sh
imiric: Yeah, the documentation is a bit confusing... I think this is mostly relevant when doing multipart uploads, but since we're copying from the same bucket and just replacing the metadata, I don't think it matters. I did have to specify `content-type` manually because otherwise everything gets `application/octet-stream` :-/

This was more intuitive with `s3cmd` which had a `modify` command, so hopefully this copying over doesn't incur additional costs.

packaging/bin/entrypoint.sh
imiric: Yeah, 1k paths/mo is generous and we probably won't cross it, so this is fine.

packaging/bin/create-msi-repo.sh
```diff
@@ -0,0 +1,66 @@
...
+    --content-type='application/x-msi' \
+    --metadata-directive=REPLACE \
+    "s3://${S3PATH}/k6-latest-amd64.msi" "s3://${S3PATH}/k6-latest-amd64.msi"
+}
+
+mkdir -p "$REPODIR"
+
+# Download existing packages
+# For MSI packages this is only done to be able to generate the index.html correctly.
+# Should we fake it and create empty files that have the same timestamp and size as the original ones?
```
na--, line 50: Nah, the cost of doing this once every 2 months or so is negligible, especially when you're deleting old packages... Though, if we ever figure it out, we might save some bandwidth by completely not hosting the .msi packages on `dl.k6.io` and just redirecting to the github release binaries :sweat_smile:

packaging/bin/create-msi-repo.sh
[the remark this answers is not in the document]
na--: I don't have strong opinion either way. The data hoarder in me balks at deleting anything, but my main problem with keeping them around forever is that it would increase release times, not so much the costs, so I understand why we have to delete them. And I see you already use `mtime`, which I somewhat prefer, so :+1: for resolving this?

na-- (dismissed):
LGTM :tada: This is awesome to see! :heart:

packaging/bin/create-msi-repo.sh
MStoykov: hmmm, I do wonder if we should not actually drop the whole msi hosting and truly just use github ?

na--: We'd still need a way to redirect people to the latest msi release (and maybe to the other plain zipped binaries we have), which judging by the "CloudFront caches redirects aggressively and I wasn't able to invalidate it" comment below, probably won't be easy to do if we just host redirects to them on dl.k6.io... 

The current solution is good enough, and has a nice side-benefit of having a folder with the old installations at hand, we should leave it be, I think...

packaging/bin/generate_index.py
```diff
@@ -0,0 +1,484 @@
...
+            </g>
+        </g>
+
+        <!-- k6 Logo -->
+        <g id="logo">
```
na--, line 307: Hmm another minor nitpick - can we make this a link to `/`, so that clicking the logo brings you back to the root folder?

packaging/bin/create-msi-repo.sh
imiric: Yeah, this would be good, though instead of dropping it entirely I'd prefer to have S3 redirects to GH like Ned mentioned, so that we can point users to a central location for all packages.

The caching issue wouldn't be a problem for links to specific versions, as they could remain static. But the link to a latest version would be an issue as it needs to be updated, though we could workaround it if we started publishing an MSI without a version in its filename, e.g. `k6-amd64.msi`. That way the latest link could also be static and redirect to `https://github.com/k6io/k6/releases/latest/download/k6-amd64.msi`.

packaging/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+LABEL maintainer="k6 Developers <developers@k6.io>"
+
+ENV DEBIAN_FRONTEND=noninteractive
+
+RUN apt-get update -y && \
+    apt-get install -y apt-utils createrepo curl git gnupg2 python3 unzip
+
+ARG AWSCLI_VERSION=2.1.35
+
+RUN curl -fSsL -o "awscliv2.zip" \
```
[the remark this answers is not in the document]
imiric: Good idea, see b5c3736f.

packaging/bin/generate_index.py
imiric: Sure, see 90da6346.

na-- (dismissed):

MStoykov (approved):

packaging/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+LABEL maintainer="k6 Developers <developers@k6.io>"
+
+ENV DEBIAN_FRONTEND=noninteractive
+
+RUN apt-get update -y && \
+    apt-get install -y apt-utils createrepo curl git gnupg2 python3 unzip
+
+COPY ./awscli-key.gpg .
+
+ARG AWSCLI_VERSION=2.1.35
```
na--, line 12: Minor nitpick, but maybe this could be an environment variable in `.github/workflows/packager.yml`, instead of hardcoded in the `Dockerfile`? So we can easily bump both it and `VERSION` at the same time.

And the `Dockerfile` default behavior, if it's not specified, could be to install the latest version, which according to https://docs.aws.amazon.com/cli/latest/userguide/install-cliv2-linux.html#cliv2-linux-install is always conveniently located at `https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip` (i.e. same URL without the `-version` before the `.zip`)

na-- (dismissed):

na-- (approved):

packaging/Dockerfile
imiric: Sure, see dec4712e.

Technically this wasn't hardcoded as build args can be changed at build time, so this was just the default, but I made it fallback on the latest version if unspecified and exposed it to be set via an env var.

MStoykov (approved):

PR #1919, opened 2021-03-19
Warn on unknown options from script

### LINKED ISSUE #1790: Warning on unknown Options
It is common for users to mistakenly mistype or even misplace configuration options. 

For [scenarios](https://k6.io/docs/using-k6/scenarios) we specifically don't allow [unknown fields](https://github.com/loadimpact/k6/blob/ef717e1af2a92d35d39a3d815a69b6ab6b8c65e1/lib/helpers.go#L33-L46) but this was skipped for the general `options` and friends because that would've been a breaking change.

But maybe we can try to parse without enabling unknown fields and warn the user and then continue with them allowed? So no breaking change but the user still got some message that what they are doing is probably not what they want to.  
 
 In the future  (half a year from releasing it) we can probably make it an error, as given how k6 works with the `options` I have some doubts this is even possible to use an unknown field for something.

MStoykov:
fix #1790

### DIFF
--- a/js/bundle.go
+++ b/js/bundle.go
@@ -21,6 +21,7 @@
 package js
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"net/url"
@@ -99,7 +100,7 @@ func NewBundle(
 		return nil, err
 	}
 
-	err = bundle.getExports(rt, true)
+	err = bundle.getExports(logger, rt, true)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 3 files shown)


MStoykov:
I specifically don't use [lib.StrictJSONUnmarshall](https://github.com/loadimpact/k6/blob/ef717e1af2a92d35d39a3d815a69b6ab6b8c65e1/lib/helpers.go#L33-L46) as I worried the UseNumber may break something with the `ext` and I really don't want to revert/bugfix it for the extra copy of that code ;).

Apart from this lib.Options has a couple of fields that use `json.Unmarshal` all of the TLS* something all of which should probably just use lib.StrictJSONUnsmarshal

imiric (approved):
:+1:

na-- (approved):

PR #1920, opened 2021-03-22
[WIP]  Add abortTest() helper function

### LINKED ISSUE #1001: Provide an API to exit a test

gernest:
Closes #1001
This adds abortTest() helper function to the k6 module. This function when
called inside a script it will

- stop the whole test run and k6 will exit with 107 status code
- stops immediately the VU that called it  and no more iterations are started
- make sure the teardown() is called
- the engine run status is 7 (RunStatusAbortedScriptError)

`(*goja.Runtime).Interrupt` is used for halting script execution and capturing
stack traces for better error message of what is happening with the script.

We introduce InterruptError which is used with `(*goja.Runtime).Interrupt` to
identify interrupts emitted by abortTest(). This way we use special handling of
this type.

Example script is

```js
import { abortTest, sleep } from 'k6';

export default function () {
    // We abort the test on second iteration
    if (__ITER == 1) {
        abortTest();
    }
    sleep(1);
}

export function teardown() {
    console.log('This function will be called even when we abort script');
}
```

abortTest() can be called in both default and setup functions, however you can't use it
in the init context

The following script will fail with the error
```
ERRO[0000] Using abortTest() in the init context is not supported at (...path to the script )init.js:13:43(34)
```

```js
import {
    abortTest
} from 'k6';

abortTest();

export function setup() {
}

export default function () {
    // ... some test logic ...
    console.log('mayday, mayday');
}
```

You can customize the reason for abortTest() by passing values to the function

```js
abortTest("Exceeded expectations");
```

Will emit `"Exceeded expectations"` on the logs

Missing 
 - [ ] Tests

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -45,6 +45,7 @@ import (
 	"github.com/loadimpact/k6/core"
 	"github.com/loadimpact/k6/core/local"
 	"github.com/loadimpact/k6/js"
+	"github.com/loadimpact/k6/js/common"
 	"github.com/loadimpact/k6/lib"
 	"github.com/loadimpact/k6/lib/consts"
 	"github.com/loadimpact/k6/loader"
@@ -63,6 +64,7 @@ const (
 	invalidConfigErrorCode       = 104
 	externalAbortErrorCode       = 105
 	cannotStartRESTAPIErrorCode  = 106
+	abortedByScriptErrorCode     = 107
 )
 
 // TODO: fix this, global variables are not very testable...
@@ -345,6 +347,8 @@ func getExitCodeFromEngine(err error) ExitCode {
 		default:
 			return ExitCode{error: err, Code: genericTimeoutErrorCode}
 		}
+	case *common.InterruptError:
+		return ExitCode{error: errors.New("Engine error"), Code: abortedByScriptErrorCode, Hint: e.Reason}
 	default:
 		//nolint:golint
 		return ExitCode{error: errors.New("Engine error"), Code: genericEngineErrorCode, Hint: err.Error()}
--- a/js/common/interupt_error.go
+++ b/js/common/interupt_error.go
@@ -1,0 +1,25 @@
+package common
+
+type InterruptError struct {
+	Reason string
+}
+
+func (i InterruptError) Error() string {
+	return i.Reason
+}
+
+var AbortTest = &InterruptError{
+	Reason: "abortTest() was called in a script",
+}
+
+var AbortTestInitContext = &InterruptError{
+	Reason: "Using abortTest() in the init context is not supported",
+}
+
+func IsInteruptError(err error) bool {
+	if err == nil {
+		return false
+	}
+	_, ok := err.(*InterruptError)
+	return ok
+}
--- a/js/modules/k6/k6.go
+++ b/js/modules/k6/k6.go
@@ -116,6 +116,24 @@ func (*K6) Group(ctx context.Context, name string, fn goja.Callable) (goja.Value
 	return ret, err
 }
 
+func (*K6) AbortTest(ctx context.Context, extra ...goja.Value) {
+	state := lib.GetState(ctx)
+	rt := common.GetRuntime(ctx)
+	if state == nil {
+		rt.Interrupt(common.AbortTestInitContext)
+		return
+	}
+	e := *common.AbortTest
+	if len(extra) > 0 {
+		var m string
+		for _, v := range extra {
+			m += v.String()
+		}
+		e.Reason = m
+	}
+	rt.Interrupt(&e)
+}
+
 func (*K6) Check(ctx context.Context, arg0, checks goja.Value, extras ...goja.Value) (bool, error) {
 	state := lib.GetState(ctx)
 	if state == nil {
--- a/lib/executor/helpers.go
+++ b/lib/executor/helpers.go
@@ -28,6 +28,7 @@ import (
 
 	"github.com/sirupsen/logrus"
 
+	"github.com/loadimpact/k6/js/common"
 	"github.com/loadimpact/k6/lib"
 	"github.com/loadimpact/k6/lib/types"
 	"github.com/loadimpact/k6/ui/pb"
@@ -74,6 +75,58 @@ func validateStages(stages []Stage) []error {
 	return errors
 }
 
+// cancelKey is the key used to store the cancel function for the context of an
+// executor. This is a work around to avoid excessive changes for the abolity of
+// nested functions to cancel the passed context
+type cancelKey struct{}
+
+type cancelExec struct {
+	cancel context.CancelFunc
+	reason error
+}
+
+// Context returns context.Context that can be cancelled by calling
+// CancelExecutorContext. Use this to initialize context that will be passed to
+// executors.
+//
+// This allows executors to globally halt any executions that uses this context.
+// Example use case is when a script calls abortTest()
+func Context(ctx context.Context) context.Context {
+	ctx, cancel := context.WithCancel(ctx)
+	return context.WithValue(ctx, cancelKey{}, &cancelExec{cancel: cancel})
+}
+
+// CancelExecutorContext cancels executor context found in ctx, ctx can be a
+// child of a context that was created with Context function.
+func CancelExecutorContext(ctx context.Context, err error) {
+	if x := ctx.Value(cancelKey{}); x != nil {
+		v := x.(*cancelExec)
+		v.reason = err
+		v.cancel()
+	}
+}
+
+// CancelReason returns a reason the executor context was cancelled. This will
+// return nil if ctx is not an executor context(ctx or any of its parents was
+// never created by Context function).
+func CancelReason(ctx context.Context) error {
+	if x := ctx.Value(cancelKey{}); x != nil {
+		v := x.(*cancelExec)
+		return v.reason
+	}
+	return nil
+}
+
+func handleInterupt(ctx context.Context, err error) bool {
+	if err != nil {
+		if common.IsInteruptError(err) {
+			CancelExecutorContext(ctx, err)
+			return false
+		}
+	}
+	return false
+}
+
 // getIterationRunner is a helper function that returns an iteration executor
 // closure. It takes care of updating the execution state statistics and
 // warning messages. And returns whether a full iteration was finished or not
@@ -96,6 +149,10 @@ func getIterationRunner(
 			return false
 		default:
 			if err != nil {
+				if handleInterupt(ctx, err) {
+					executionState.AddInterruptedIterations(1)
+					return false
+				}
 				if s, ok := err.(fmt.Stringer); ok {
 					// TODO better detection for stack traces
 					// TODO don't count this as a full iteration?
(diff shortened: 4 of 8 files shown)


na-- (changes requested):
Thanks! I haven't done a full code review yet, I've only skimmed the changes and played around with the resulting binary a bit, but in general, your approach looks good! The only big issue I noticed is that these changes break the  `--linger` functionality. 

`k6 run --linger script.js` should still linger when `abortTest()` is used, i.e. the k6 _process_ shouldn't exit directly after the test run has stopped. This behavior should remain the same as it was with the old way of aborting the test run (thresholds with `abortOnFail: true`).

gernest:
@na-- thanks for the quick and good feedback. I had missed the `--linger` flag, but the docs were informative so I was able to address the flag use case.

When we run scripts that use `abortTest()` with  `--linger`  ,  basically every thing described on the PR message above stays the same except we don't immediately exit with `107` code .

Winding down of the engine  will be observed like with thresholds and waiting for `CTRL+C` which will  then allow us to exit with `107`.

Relevant commit c07af18

lib/executor/helpers.go
```diff
@@ -74,6 +75,60 @@ func validateStages(stages []Stage) []error {
...
+// This allows executors to globally halt any executions that uses this context.
+// Example use case is when a script calls abortTest()
+func Context(ctx context.Context) context.Context {
+	ctx, cancel := context.WithCancel(ctx)
+	return context.WithValue(ctx, cancelKey{}, &cancelExec{cancel: cancel})
+}
+
+// CancelExecutorContext cancels executor context found in ctx, ctx can be a
+// child of a context that was created with Context function.
+func CancelExecutorContext(ctx context.Context, err error) {
```
imiric, line 101: This can be unexported, right?

js/modules/k6/k6.go
```diff
@@ -116,6 +116,26 @@ func (*K6) Group(ctx context.Context, name string, fn goja.Callable) (goja.Value
...
 
+// AbortTest exposes abortTest function in the k6 module. When called ti will
+// interrupt the active goja runtime passed with ctx.
+func (*K6) AbortTest(ctx context.Context, extra ...goja.Value) {
+	state := lib.GetState(ctx)
+	rt := common.GetRuntime(ctx)
+	if state == nil {
+		rt.Interrupt(&common.InterruptError{Reason: common.AbortTestInitContext})
+		return
+	}
```
MStoykov, line 127: as mentioned above, I think you should just drop this lines

lib/executor/helpers.go
```diff
@@ -28,6 +28,7 @@ import (
 
 	"github.com/sirupsen/logrus"
 
+	"github.com/loadimpact/k6/js/common"
```
MStoykov, line 31: you should not be importing `k6/js/*` anything from anywhere in `lib`. 

This has many reasons but in general we don't want to tie the `lib` functionality to a specific runner even if we only currently have `js` one and it is unlikely we will ever support others(and there are many other problems in the way of it)

lib/executor/helpers.go
```diff
@@ -96,6 +151,10 @@ func getIterationRunner(
 			return false
 		default:
 			if err != nil {
+				if handleInterupt(ctx, err) {
+					executionState.AddInterruptedIterations(1)
+					return false
+				}
```
MStoykov, line 157: Looking at the code nothing uses the return value. This seems to not matter as the contexts are getting checked constantly, but it is probably a good idea to figure out if we need to use it somehow

cc @imiric @na--

imiric (commented):
This looks very good to me, nice work! I just left some minor comments and nitpicks.

Also there's a check next to "Check teardown is called when a script aborts", but I don't see that test and I agree that it should be added.

MStoykov (changes requested):
Seems good in general :tada: 

- definitely, need to move the IterruptedError definition out of `js/common` and probably ... `lib/errors`, something else ?
- not so certain that this isn't an abuse of contexts and that I wouldn't prefer a more direct approach, but looking at the code around this ... I think that will require some pretty big refactoring and probably will make things actually worse than better, so I am fine with it

cmd/run.go
```diff
@@ -276,8 +279,16 @@ a commandline interface for interacting with it.`,
...
 			initBar.Modify(pb.WithConstProgress(0, "Starting test..."))
+			var interrupt error
 			if err := engineRun(); err != nil {
-				return getExitCodeFromEngine(err)
+				if !conf.Linger.Bool {
+					return getExitCodeFromEngine(err)
+				}
+				if !common.IsInteruptError(err) {
+					return getExitCodeFromEngine(err)
+				}
```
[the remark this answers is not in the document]
gernest: thanks , nice catch

js/modules/k6/k6.go
```diff
@@ -116,6 +116,26 @@ func (*K6) Group(ctx context.Context, name string, fn goja.Callable) (goja.Value
 	return ret, err
 }
 
+// AbortTest exposes abortTest function in the k6 module. When called ti will
+// interrupt the active goja runtime passed with ctx.
+func (*K6) AbortTest(ctx context.Context, extra ...goja.Value) {
```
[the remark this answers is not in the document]
gernest: done 48a0ed6f

js/common/interupt_error.go
```diff
@@ -0,0 +1,26 @@
+package common
```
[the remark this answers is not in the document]
gernest: done a7d66f94

lib/executor/helpers.go
gernest: Yep! initially the cancellable context was initialised independently in each executor, that is no longer the case though.

done e58e3eb5

js/common/interupt_error.go
```diff
@@ -0,0 +1,26 @@
...
+	return i.Reason
+}
+
+// AbortTest is a reason emitted when a test script calls abortTest() without arguments
+const AbortTest = "abortTest() was called in a script"
+
+// AbortTestInitContext is a reason  emitted when a test script calls abortTest() without arguments
+// in the init context
+const AbortTestInitContext = "Using abortTest() in the init context is not supported"
+
```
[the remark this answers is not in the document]
gernest: nice catch! 🤦  lol!

done eaa9ea97

js/modules/k6/k6.go
gernest: done eaa9ea97

lib/executor/helpers.go
gernest: I moved `InterruptError` to `lib/errors`. 

done 75f8e87b

cmd/run.go
```diff
@@ -63,11 +64,13 @@ const (
 	invalidConfigErrorCode       = 104
 	externalAbortErrorCode       = 105
 	cannotStartRESTAPIErrorCode  = 106
+	abortedByScriptErrorCode     = 107
 )
 
 // TODO: fix this, global variables are not very testable...
 //nolint:gochecknoglobals
 var runType = os.Getenv("K6_TYPE")
+var errScriptInterupted = errors.New("script execution  was aborted")
```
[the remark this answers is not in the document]
gernest: done f2c9911fbce0edc3dfac41bbcb6a24c0a111bdc7

js/common/interupt_error.go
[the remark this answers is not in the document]
imiric: Nit: maybe make it `(C) 2021` :wink:

(I wish the `goheader` linter would work properly and do this for us...)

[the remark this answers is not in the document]
gernest: done 921f35a1

gernest:
@na-- @imiric @MStoykov thanks a lot for the amazing feedback, I have addressed all review comments (I think) PTAL

> Also there's a check next to "Check teardown is called when a script aborts", but I don't see that test and I agree that it should be added.

Thanks for catching this one, I added the test in b5b8441

na-- (approved):
LGTM! The only minor nitpick I have is that instead of directly type casting errors, you should probably use the recent [`errors.Is()`](https://golang.org/pkg/errors/#Is) and [`errors.As()`](https://golang.org/pkg/errors/#As) stdlib functions. 

Other than that, everything looks great! The changes ended up more complex than I expected, but can't think of a way to simplify them without significant refactoring and everything is still understandable, so this is going to be it! :tada:

lib/executor/helpers.go
na--: Yeah, we might make a small PR to refactor this, but it might be needed eventually (e.g. in https://github.com/k6io/k6/issues/1250 or https://github.com/k6io/k6/issues/877) and is so small that it's probably fine as it is.

imiric (approved):
LGTM, well done! Just left a tiny nitpick, feel free to disregard.

Also, please rebase on top of the latest `master` and cleanup the commits a bit before this is merged. Not necessarily squashing to a single commit, but whatever makes sense for the changes to be atomic.

rushby:
Hi all,

Just been looking into this functionality would be great if this can get merged in soon!

na--:
@rushby, we wanted to merge this for k6 v0.33.0, but we weren't able because it's blocked by some other changes we needed to make in k6 (e.g. https://github.com/k6io/k6/pull/1863) and in our cloud, so it can work both in single instance and in distributed tests. We plan to merge it for k6 v0.34.0, slated to be released sometime in the end of August/early September.

imiric:
Closing this in favor of #2093.

We changed the API slightly but you'll get full credits @gernest. Thanks!

PR #1925, opened 2021-03-25
Fix DOCKER_IMAGE_ID and add GHCR_IMAGE_ID after repo move

MStoykov:


MStoykov:
I wonder if I shouldn't make the `GHCR_IMAGE_ID` the default name and the `DOCKER_IMAGE_ID` only for when pushing to docker?

na-- (approved):

.github/workflows/all.yml
```diff
@@ -237,13 +238,13 @@ jobs:
 
           # Log into GitHub Container Registry
           echo "${{ secrets.CR_PAT }}" | docker login https://ghcr.io -u ${{ github.actor }} --password-stdin
-          docker tag "$DOCKER_IMAGE_ID" "ghcr.io/$DOCKER_IMAGE_ID:$VERSION"
-          docker push "ghcr.io/$DOCKER_IMAGE_ID:$VERSION"
+          docker tag "$DOCKER_IMAGE_ID" "ghcr.io/$GHCR_IMAGE_ID:$VERSION"
```
imiric, line 241: Shouldn't this be:

```suggestion
          docker tag "$GHCR_IMAGE_ID" "ghcr.io/$GHCR_IMAGE_ID:$VERSION"
```

?

That is, we want the GHCR image to be accessible via `k6io/k6` and the Docker Hub one via `loadimpact/k6` (until we move the Docker Hub org as well).

imiric (approved):
> > I wonder if I shouldn't make the `GHCR_IMAGE_ID` the default name and the `DOCKER_IMAGE_ID` only for when pushing to docker?

I think it's fine as is and we'll address this detail once we move the Docker Hub org.

imiric (approved):

PR #1926, opened 2021-03-25
update golang.org/x/* dependencies to internal-branch.go1.16-vendor

MStoykov:
This is mostly done so we catch up with the `golang.org/x` libraries for
the golang 1.16 release.

The majority of the LOC changes are unicode 13 related as it was added
in golang1.16 and some syscall constants that were added.


imiric (commented):
That's a lot of new LOCs...

Can you document why we're doing this in the commit and the PR description? To resolve some HTTP2 issue with Go 1.16?

MStoykov:
The majority of the LOC is Unicode 13 tables and there is a smaller amount of syscall constants that were added.

I am mostly updating this as we previously never updated the `golang.org/x/net`(/http2) and that lead to problems and the `x` libraries are practically embed in the golang stdlib in some cases and are possibly the most test libraries out there, also fairly heavy used in all other dependencies.

I am not aware of any particular issue that this is fixing, but given their near stdlib nature, these are probably the libraries I am not going to try and figure out if we need to update them or not, just update them ;).

I am planning to make a table for the rest of our dependencies and categorize them based on whether we *should* update them or if whatever version we currently have is perfectly fine and even if new ones have some improvements that wouldn't matter(based on what we use it not actually going through the changelog as that would take forever). And probably other characteristics, but I think the `golang.org/x` are exempt from that.

imiric:
OK, thanks, that makes sense. Could you still put a summary of that in the commit message, for future reference?

na-- (approved):
Though yes, please add PR descriptions.

imiric (approved):

na-- (approved):

Issue #1933, opened 2021-03-29
k6 dependencies update policy discussion


MStoykov:
edit:  updated table in [comment](https://github.com/k6io/k6/issues/1933#issuecomment-838173237) to preserve history better

I make this issue to discuss whether we will be updating each dependency and on what ... schedule.

I would argue that while some dependencies are super important and later versions likely fix bugs and performance problems and it is worth our time updating them, others are either ones that we want to drop or are used in a single place for a single task and because of some specificity there is no point updating it regularly or at all.

In none of the cases, this means that a given library MUST be updated blindly or MUST NOT be updated if there turns out to be a good reason. Just that updating some libraries probably has more relevance than others and given our limited resources taking the time to update all of them constantly and fixing problems is probably not ... wanted.

Additionally, while I would prefer if some scripts are added to make this easier, I am definitely against this breaking the CI or preventing a release.

### Goja

[Goja](https://github.com/dop251/goja) is probably one of the bigger exceptions. It is what we use to run javascript and as any improvement can turn out to have a considerable either positive or negative impact on k6 we practically have to update it constantly. So even if something that we don't particularly care about is added/fixed we probably should update to it at least once each release cycle. In reality, there has been something useful more or less every month so we have been updating it constantly.

### golang.org/x/*

While those libraries are not part of the golang stdlib they either get embed in it in different forms or are the natural extension. They also happen to have a branch for each release so my proposal is to just update to the new branch each time we update the golang version so we follow it. Or even just at the beginning of each release just in case.

Given their ubiquity in general I don't expect it will be breaking constantly. If anything [not updating x/net has turned](https://github.com/k6io/k6/pull/1734) out to be problematic albeit because we were at an extreme not updating it ...

### Others:

The list below is what it was in go.mod and I am listing the `indirect` just for ease of checking against the go.mod. 

Post factum the categories are :
- never - we either don't care about any updates as we use it for something extremely specific and  updating it won't help. Or the more common case it is something we want to drop.
- every release - things that we more or less need to update as we are very dependant on it 
- 1-2 a year - things that we should be updating, but it is likely with minimal benefits and just doing it all the time has more chances of us running into problems then something else.
 

| name | description | thoughts | proposed update period|
|-----|---------|---|----|
| github.com/Azure/go-ntlmssp | NTLM library | I am for updating it, but given our problems with testing it, I think that should wait | ~never for now~ 1-2 a year as we still need to update it |
[...]

PR #1940, opened 2021-03-31
Update other dependencies

MStoykov:
This updates the majority of the k6 dependencies that we intended to update from #1933.

The biggest size increases are from yaml.v3 addition, compression algorithms, and testify additions.


MStoykov:
The only remaining dependencies that should be updated at all from https://github.com/k6io/k6/issues/1933 are:
- `github.com/DataDog/datadog-go` there were a lot of changes including new dependencies for support of windows pipes and some aggregation code. Given that it still doesn't support tcp and we might want to add some support for the aggregation I decided I should look at it more closely and open an issue to discuss what we should do.
- `github.com/fatih/color`, `github.com/mattn/go-colorable` and `github.com/mattn/go-isatty` - there are a lot of changes here including a lot of fixes for windows (supposedly). I did test it under Linux somewhat, but decided that given the amount of changes a lot more testing under windows will be needed and decided that would mean it should be in a separate PR
- `github.com/spf13/cobra` and `github.com/spf13/pflag` - again a lot more changes including more support for context and some windows related changes. So it is actually a combination of the two above - I want to look at the changes in more detail and it likely will require more testing under windows. Also did update color libraries somewhat which means that it should probably happen after we update those ;)

There were no updates for:
- `github.com/nu7hatch/gouuid` 
- `github.com/mccutchen/go-httpbin`

na-- (changes requested):
This PR adds `gopkg.in/yaml.v3`, while keeping `gopkg.in/yaml.v2`. I think we can drop the latter with a minor change in k6:
```
$ go mod why 'gopkg.in/yaml.v3'
# gopkg.in/yaml.v3
github.com/loadimpact/k6/lib/testutils/httpmultibin
github.com/stretchr/testify/assert
gopkg.in/yaml.v3
$ go mod why 'gopkg.in/yaml.v2'
# gopkg.in/yaml.v2
github.com/loadimpact/k6/ui
gopkg.in/yaml.v2
```

MStoykov:
Unfortunately - not @na-- , as it's also used in `gin` and another `yaml` library which used by `helm` of which we use `strvals`

Also even if we could I prefer to do it in a separate PR, where I will drop the direct uses by k6

na-- (approved):
https://github.com/golang/go/issues/41305 ... :disappointed:

imiric (approved):
:crossed_fingers: A lot of new LOCs, but it seems unavoidable.

Nice job reviewing each update and detailing it in separate commits! :clap:

Issue #911, opened 2019-01-29
Remove dependencies for the API that are not necessary


MStoykov:
This probably should be done when we rewrite for #550. 
Removing the api directory and then removing dependencies through `dep ensure` removed 3mb from `vendor`. And we definitely don't need `negroni` or `api2go` for 5 api calls

PR #1943, opened 2021-04-02
Drop github.com/urfave/negroni as a dependency

MStoykov:
The only "breaking change" is that if now k6 panics during answering a
REST API request it will close the connection instead of returning
`500`. This can also be done easily I just don't think it's particularly
necessary.

This is part of #911


na-- (approved):
:tada: :crossed_fingers:

PR #1947, opened 2021-04-05
Drop dependency on the package github.com/manyminds/api2go

MStoykov:
This still depends on the module and the package
github.com/manyminds/api2go/jsonapi , but with this it also drops a lot
of completely unused dependencies that were just pulled because we were
using a struct definition.

This also let us drop github.com/julienschmidt/httprouter which was
otherwise referenced from the now dropped github.com/manyminds/api2go/routing


na-- (approved):
This is awesome! :hearts: :tada: :confetti_ball:

na-- (approved):

PR #1949, opened 2021-04-05
Drop dependency on github.com/julienschmidt/httprouter

MStoykov:


api/v1/routes.go
```diff
@@ -22,27 +22,80 @@ package v1
...
+			rw.WriteHeader(http.StatusMethodNotAllowed)
+			return
+		}
 
-	router.POST("/v1/teardown", HandleRunTeardown)
+		handleRunTeardown(rw, r)
+	})
 
-	return router
+	return mux
```
MStoykov, line 100: I really do wonder if I shouldn't just make one big switch with `string.HasPrefix` and friends?

na--: :+1: it might be a bit better, but even like this it's a great improvement compared to before, so you decide :man_shrugging:

na-- (approved):

api/v1/routes.go
MStoykov: I started doing it and looks worse IMO so I am going to just leave it

PR #1950, opened 2021-04-05
Update golangci-lint to v1.39.0

na--:
Might be a good idea to go through the complete list of current linter issues and try to prune certain classes of the most common errors we have. For example, disabling certain linters in tests, or in go-generated files, `gofumpt`-ing the whole codebase, etc.


na--:
When a PR that doesn't touch any code has a -0.08% code coverage difference, it's obvious something is iffy with codecov...

MStoykov (approved):

PR #1951, opened 2021-04-05
Drop `github.com/pkg/errors` and fix `error_code` detection

na--:
In the spirit of spring cleaning and https://github.com/k6io/k6/issues/1933 :sweat_smile: 

This also makes some of the changes that https://github.com/k6io/k6/pull/1660 did, though in a slightly different way, so we should probably try to port the rest of them.


lib/netext/httpext/error_codes.go
```diff
@@ -109,7 +109,13 @@ func http2ErrCodeOffset(code http2.ErrCode) errCode {
 
 // errorCodeForError returns the errorCode and a specific error message for given error.
 func errorCodeForError(err error) (errCode, string) {
-	switch e := errors.Cause(err).(type) {
+	// We explicitly check for `Unwrap()` in the default switch branch, but
+	// checking for the concrete error types first gives us the opportunity to
+	// also directly detect high-level errors, if we need to, even if they wrap
+	// a low level error inside.
```
MStoykov, line 115: I would prefer if you port the tests as well before we merge this.

The logic in #1660 is the reverse: we first go for the most inner error and then upwards while here we go from the outer one inner. 
The current code also in practice just goes for the inner one, *only*, unless it is url.Error in which case it goes for the inner error (this probably is not needed?!?). 

Either way the tests will show whether at least for those cases this keeps working and they are perfectly fine in a separate commit in this PR

MStoykov (changes requested):
LGTM, but I would prefer to get the tests from #1660 ported, before we merge this

lib/netext/httpext/error_codes.go
na--: :+1: I'll port the tests as well as some of your other changes from that PR. 

My logic with the current code is that, if you want to, you can always drill down and get a wrapped error, even before reaching the `default`, as with `url.Error`. But it doesn't _force_ you to do so, you can catch top-level errors even if they wrap another one, if that's what's necessary to return the proper error code.

na--: I ported your tests and some of the fixes from https://github.com/k6io/k6/pull/1660, then fixed the one broken test with t he new logic. Finally, I fixed some lint issues and and refactored the whole `errorCodeForError()` function to be a bit more readable by splitting apart the `*net.OpError` detection in a new function `errorCodeForNetOpError()`. 

Please read commit by commit, and if you're satisfied everything usable from https://github.com/k6io/k6/pull/1660 has been ported, please close that PR.

MStoykov (dismissed):
LGTM 

But the test added in 9fac9cac breaks until d14d24d6

na--:
Yes, I know, I deliberately didn't fix it as a part of the cherry-pick. I'll probably squash this PR when merging it in `master`, but I'll leave the commits as they are here in the PR, so the iterative issue fixing and refactoring process is clear to anyone that wants to debug it in the future, but we won't have any issues with `git bisect` or something like that.

imiric (dismissed):
LGTM, I like the more descriptive error messages. Just left a couple of tiny nitpicks.

lib/netext/httpext/error_codes_syscall_windows.go
```diff
@@ -0,0 +1,16 @@
+package httpext
```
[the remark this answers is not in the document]
na--: Resolved by https://github.com/k6io/k6/pull/1951/commits/bd835279d8b7131829678c4acc8326c820584840

js/modules/k6/html/html.go
```diff
@@ -143,8 +143,8 @@ func (s Selection) adjacentUntil(until func(string) *goquery.Selection,
 		}
 	}
 
-	errmsg := fmt.Sprintf("Invalid argument: Cannot use a %T as a selector", def[0].Export())
-	panic(s.rt.NewGoError(errors.New(errmsg)))
+	err := fmt.Errorf("invalid argument, cannot use a %T as a selector", def[0].Export())
```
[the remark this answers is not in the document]
na--: Resolved by https://github.com/k6io/k6/pull/1951/commits/bd835279d8b7131829678c4acc8326c820584840

MStoykov (dismissed):

MStoykov (approved):

imiric (approved):

PR #1952, opened 2021-04-06
Maitenance update of github.com/dop251/goja to latest master

MStoykov:
This also pulled golang.org/x/text update to v0.3.6 from v0.3.4 which is
mostly bugfixes and some new functionality we don't care about.


MStoykov:
I wonder if we shouldn't replace the `x/text` library though as goja seems to be updating this whenever it has a release, which might be a problem

go.sum
```diff
@@ -66,8 +67,9 @@ github.com/dgryski/go-sip13 v0.0.0-20181026042036-e10d5fee7954/go.mod h1:vAd38F8
 github.com/dlclark/regexp2 v1.4.1-0.20201116162257-a2a8dda75c91 h1:Izz0+t1Z5nI16/II7vuEo/nHjodOg0p7+OiDpjX5t1E=
 github.com/dlclark/regexp2 v1.4.1-0.20201116162257-a2a8dda75c91/go.mod h1:2pZnwuY/m+8K6iRw6wQdMtk+rH5tNGR1i55kozfMjCc=
 github.com/dop251/goja v0.0.0-20210216182323-60bc6ebb9fc1/go.mod h1:Mw6PkjjMXWbTj+nnj4s3QPXq1jaT0s5pC0iFD4+BOAA=
-github.com/dop251/goja v0.0.0-20210317175251-bb14c2267b76 h1:da/WKkWtBnwAC/DPGZQ7W2pKVEHGEyFU1u40L73DJN8=
-github.com/dop251/goja v0.0.0-20210317175251-bb14c2267b76/go.mod h1:Mw6PkjjMXWbTj+nnj4s3QPXq1jaT0s5pC0iFD4+BOAA=
+github.com/dop251/goja v0.0.0-20210402105715-96fa0aa6d91e h1:MjlxDU2bcHGr4NxuZ1HQQv3fMmMaz3cKhoRf1W1Dl9o=
+github.com/dop251/goja v0.0.0-20210402105715-96fa0aa6d91e/go.mod h1:R9ET47fwRVRPZnOGvHxxhuZcbrMCuiqOz3Rlrh4KSnk=
+github.com/dop251/goja_nodejs v0.0.0-20210225215109-d91c329300e7/go.mod h1:hn7BA7c8pLvoGndExHudxTDKZ84Pyvv+90pbBjbTz0Y=
```
na--, line 72: I wonder why this didn't show before, considering [this](https://github.com/dop251/goja/blame/33737909d64ebd785cb4a59435242701b153bde6/goja/main.go#L17-L18) is some very old goja code :confused: Maybe some Go 1.16 changes in go modules? :man_shrugging: In any case, since we don't use goja as a program but as a library, it doesn't really matter to us, just more `go.sum` noise :disappointed:

na-- (approved):
Hmm the only place we use `golang.org/x/text` in k6 actually is here: https://github.com/k6io/k6/blob/3515db5886ff1eb0811ab35666ccb70c905b8374/ui/summary.go#L31

Which will hopefully disappear soon, when https://github.com/k6io/k6/issues/1806 is done. So we probably won't care when goja updates is, given that it's the one thing that actually uses it :man_shrugging:

imiric (approved):

PR #1961, opened 2021-04-12
Create codeql-analysis.yml

jfcg:
Add Github CodeQL security checks for Go.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,22 @@
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ master ]
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout repo
+      uses: actions/checkout@v2
+
+    - name: Initialize CodeQL
+      uses: github/codeql-action/init@v1
+      with:
+        languages: go
+
+    - name: CodeQL Analysis
+      uses: github/codeql-action/analyze@v1


jfcg:
Btw in my fork under `/security/code-scanning` I see:
k6

You can see merge scan results [here](https://github.com/k6io/k6/security/code-scanning?query=ref%3Arefs%2Fpull%2F1961%2Fmerge+tool%3ACodeQL)

na--:
@jfcg, thanks for submitting this PR and sorry for the delay in responding! :disappointed: We are still unsure if we'd like to integrate CodeQL, since we already have extensive linter coverage for our Go code via [golangci-lint](https://golangci-lint.run/). Besides, all of the results from this scan I can see are false-positives in our JS examples, for example:
https://github.com/k6io/k6/blob/47a0b493cce0020d445a47edbca79c25ab5a550f/cmd/testdata/example.js#L81
https://github.com/k6io/k6/blob/47a0b493cce0020d445a47edbca79c25ab5a550f/samples/graphql.js#L4

So, before we can merge this, it will have to be configured to check only Go code, not JavaScript... Though the 2 `Incomplete regular expression for hostnames` issues in our Go code that were caught are valid and golangci-lint doesn't seem to be able to catch them, so there is some value in this service :tada:

jfcg:
simplified the codeql file, includes only Go checks

na-- (commented):
Thanks! Why did you remove the `pull_request` build trigger?

jfcg:
> Thanks! Why did you remove the `pull_request` build trigger?

Analysis takes some time and most projects I have contributed CodeQL patches/configurations preferred only on pushes. This is the leanest codeql configuration, but of course if you prefer they can also be added.

na-- (approved):
Sorry for the delay here, I forgot to circle back and merge this :disappointed:

MStoykov (approved):

PR #1963, opened 2021-04-13
Move statsd to output

MStoykov:
Also, deprecates the datadog one where instead now the
statsd output should be used with configuring
`K6_STATSD_ENABLE_TAGS=true` .

The tag format k6 support is supported by (at least) Datadog, 
New Relic and statsd v0.9.0.  

Additionally, a new `K6_STATSD_TAG_BLOCKLIST` was added analogous to the
`K6_DATADOG_TAG_BLACKLIST`.


MStoykov:
Possible TODOs:
- [ ] Support cli arg like `-o statsd=,tagFormat=newrelic`
- [ ] Change the default to `statsdv2`? statsd [v0.9.0](https://github.com/statsd/statsd/releases/tag/v0.9.0) last August 
- [ ] Research and make an issue listing other tag formats so we know what we don't support and possibly try to support them at a later point


I decided against updating or changing the stated library at this time as I really didn't want to feature creep this and it took long enough to actually test even basically that the configuration works

na--:
Wait... so statsd v0.9.0 now supports tags, and their format is the same as DataDog and NewRelic? :confused: And older versions of statsd would have just silently discarded the tags if we had sent them? So what is the problem if we just start always sending tags and get rid of the `tagFormat` complexity?

MStoykov:
> Wait... so statsd v0.9.0 now supports tags, 

yes

> and their format is the same as DataDog and NewRelic? 

yes (as they are the same format) as well as the graphite ... syntax 
https://github.com/statsd/statsd/pull/697/files#diff-6dc9992921c93c2b519d0811350eed1b8cc133d31cb9317dadd09da80f8c21ddR382-R417
as can be seen in this test the difference is where the tags are on the line.

Unfortunately our datadog/statsd library support only one of the formats (the datadog/newrelic/statsdv2 one, not the graphite one) so ... we can't add `graphite` to the list :(. 

We can only change the part of how a key,value pair is represented not where on the line it is ... so we can change [them not being `key:value`](https://github.com/k6io/k6/pull/1963/files#diff-be89c7f9837b9c6108b2fcf5e30b5cf67940c1988d3ea39da9593dcb169ac1feR57-R64) but `key=value` which is not very useful as I haven't found anything that supports it with `=`, but it is what it is :shrug: . This seems to mostly be so that the datadog lib doesn't take `[][2]string` or `map[string]string` but `[]string` instead as tags, but I don't want to go dig down this.

> And older versions of statsd would have just silently discarded the tags if we had sent them?

Unfortunately ... not from my testing. This [PR](https://github.com/statsd/statsd/pull/696) seems to indicate this was the case but my testing with v0.8.6 (the last one before v0.9.0) shows that even if it did do something it doesn't do enough of it with any settings I tried (which already is a problem). 

Now due to the way the [backend interface](https://github.com/statsd/statsd/blob/master/docs/backend_interface.md) is done and how the code is structured the `packet` is [emitted](https://github.com/statsd/statsd/blob/0c241dbbb3d19e13f61a4626df190fc9c0e1c27a/stats.js#L236) before the check for [bad line](https://github.com/statsd/statsd/blob/0c241dbbb3d19e13f61a4626df190fc9c0e1c27a/stats.js#L273) so in theory, a backend can support them :shrug:  but in reality, both the `console` and the `graphite` one that are included in the statsd by default *do not* listen for the `packet` event so ... I would argue that doesn't help us.

For the record statsd as a project is ... kind of *dead* for *years* , there are mostly very small updates once in a while with the tag support linked above being like the biggest one in years. 

So given all of the above, I don't want to try to figure out if I can make it work somehow and then write an incredibly wrong explanation on how others might be able to make it work with <v0.9.0...  

My idea with enabling statsdv2 by default is that v0.9.0 is not all that new at this point so ... hopefully, anyone using pure statsd still has updated to it at this point  in time :shrug: 

We can also leave this so for some versions and then change it in 1-2 versions ?

cmd/outputs.go
```diff
@@ -58,28 +57,11 @@ func getAllOutputConstructors() (map[string]func(output.Params) (output.Output,
...
-			if err != nil {
-				return nil, err
-			}
-			datadogc, err := datadog.New(params.Logger, conf)
-			if err != nil {
-				return nil, err
-			}
-			return newCollectorAdapter(params, datadogc), nil
+			params.Logger.Warn("The datadog output is deprecated, and will be removed in a future k6 version. " +
+				"Please use the statsd output with tagFormat=dadog instead.")
```
na--, line 63:
```suggestion
				"Please use the statsd output with tagFormat=datadog instead.")
```

output/statsd/output.go
```diff
@@ -0,0 +1,188 @@
...
+
+	logger logrus.FieldLogger
+	client *statsd.Client
+}
+
+func (o *Output) dispatch(entry stats.Sample) error {
+	var tagList []string
+	switch o.config.TagFormat.t {
+	case tagFormatDatadog, tagFormatNewrelic, tagFormatStatsdv2:
+		tagList = processTags(o.config.TagBlocklist, entry.Tags.CloneTags())
```
imiric, line 70: I kind of see @na--'s comment about this: if they all use the same format, why not just drop this or leave only `tagFormatStatsdv2`?

Though I guess it would be useful to have to handle any inconsistencies and eventually support the graphite format as well.

imiric (commented):
I'll have to read more about statsd changes and tag formats, but LGTM! :thumbsup:

output/statsd/output.go
MStoykov: statsd prior to v0.9.0 *does* not support the tag format so if someone is still using it and updates k6 it will stop working for them. Also arguably `graphite` and possibly other pieces of software that support the non-tag statsd format but not the "datadog" one will also possibly *stop* working.

So IMO we need at least `none` and one more whether we have 3 names for the same thing makes very little difference in the actual code (as you can see), so I prefer to have less confusion and just name them accordingly.

MStoykov: statsd prior to v0.9.0 *does* not support the tag format so if someone is still using it and updates k6 it will stop working for them. Also arguably `graphite` and possibly other pieces of software that support the non-tag statsd format but not the "datadog" one will also possibly *stop* working.

So IMO we need at least `none` and one more whether we have 3 names for the same thing makes very little difference in the actual code (as you can see), so I prefer to have less confusion and just name them accordingly.

simskij:
I don't have a strong opinion either way, but my personal preference would be to have `tagFormat` renamed to just `format` and have it default to statsv2 without the user having to specify it. We'd only have one additional format at this point, and that would be statsv1, where we would omit the tags.

If the v1 format, or Etsy format I guess, now has support for tags as well, we could just call it "tagless", "graphite" or "simple". The point is that I think we should make the defaults as sane as possible, and given that both Datadog and NR supports tags, I would vote for v2 being a sensible default.

output/statsd/output.go
imiric: I guess I don't see the point in exposing different vendor formats if they're all essentially the same and could be grouped under a generic `statsdv2` format. Seems cosmetic for no good reason. Similar to what [@simskij mentioned](https://github.com/k6io/k6/pull/1963#issuecomment-821144078).

MStoykov:
I am fine with:
- renaming to `format`
- having only `statsdv1` and `statstdv2` 
- and having `statsdv2` being the default

I am a lot less fine having no way to *not* send tags. 

I do prefer if we have separate names even if they mean the same as that improves the findability of the option, but I agree that documentation should be enough and if we ever need we can add them.

@simskij graphite (or I guess carbon) seems to have a different format for ingestion and now supports tags ... or at least graphite itself supports tags, this seems to have been the reason the tags were added to stated v0.9.0.  But I digress ... I guess we can have `tagless` but then the question is what do we call the other one? as statsdv2 is technically what New Relic called its [integration](https://docs.newrelic.com/docs/integrations/host-integrations/host-integrations-list/statsd-monitoring-integration-version-2/) not the actual format :facepalm: . So in reality it has no "official" or semi-official name. 

So I am not really certain that `statsdv1` and `statsdv2` are all that great either ... technically statsd does not have v1.0.0 even ;)

simskij:
Then maybe `statsdv2` and `tagless` (or similar) for now, and then we can add graphite/carbon as well in the future if there is ever enough interest from the community?

MStoykov:
so "statsd" tag formats I could find:

- graphite : `fo/o;tag1=val1:250|c`  https://github.com/statsd/statsd/pull/697/files#diff-6dc9992921c93c2b519d0811350eed1b8cc133d31cb9317dadd09da80f8c21ddR385  ([here](https://github.com/cloudfoundry/statsd-injector/issues/5#issuecomment-467497494) it's called "influxdb" :shrug: 
- dogstatsd/newrelic `fo/o:250|c|#tag1:val1` https://github.com/statsd/statsd/pull/697/files#diff-6dc9992921c93c2b519d0811350eed1b8cc133d31cb9317dadd09da80f8c21ddR404
- signalfx `statsd.[foo=bar,dim=val]test:1|g"`  https://docs.signalfx.com/en/latest/integrations/agent/monitors/collectd-statsd.html#adding-dimensions-to-statsd-metrics

I am starting to come to the conclusion to just support "stasdv2"/datadog/newrelic format and drop option to support other for this release while at the same making an "extension" repo with the support to disable tags (between now and the release) and leave it at that.

na-- (dismissed):
LGTM, though I haven't tested this

na-- (approved):

output/statsd/output_test.go
```diff
@@ -0,0 +1,137 @@
...
+			"pushInterval": "%s",
+			"tagBlocklist": ["tag1", "tag2"],
+			"enableTags": true
+		}`, addr.String, namespace.String, bufferSize.Int64, pushInterval.Duration.String())),
+			})
+	}, func(t *testing.T, containers []stats.SampleContainer, expectedOutput, output string) {
+		outputLines := strings.Split(output, "\n")
+		expectedOutputLines := strings.Split(expectedOutput, "\n")
+		for i, container := range containers {
+			for j, sample := range container.GetSamples() {
```
imiric, line 86: Can you add a `require.Len()` assertion for both `containers` and `container.GetSamples()`? Otherwise if they're empty none of the checks below will run and the test will silently pass.

output/statsd/datadog.go
```diff
@@ -108,34 +74,47 @@ func NewConfig() Config {
...
 		// TODO: get rid of envconfig and actually use the env parameter...
 		return result, err
 	}
-	result = result.Apply(envConfig)
+	result = result.Apply(envdatadogConfig)
 
 	return result, nil
 }
+
+// NewDatadog creates a new statsd connector client with datadog tag format
```
imiric, line 99: This comment is now outdated since there's no "tag format".

imiric (commented):
LGTM, just a minor test thing and comment nitpick. I like that the format was dropped and now only an env var is needed to handle tags.

Did you test this with Datadog or New Relic? An integration test with a local statsd would be good to have at least, since we can't easily test against DD/NR, but it could be added later.

MStoykov:
Given that we just use a library that outputs the datadog format and is made by datadog ... I don't intend on having an integration test. New relic worked before so, unless they decide to break stuff, we haven't changed that particular part at all. 

I do test with statsd v0.9.0 and v0.8.6 and it has a way to show the messages before ingesting them (`dumpMessages:true`) and just checked that v0.31.1 with `-o datadog` and the new version with `-o datadog` or `-o statsd` and `K6_STATSD_ENABLE_TAGS=true` produce same/similar format lines. At the end of the day this PR does *not* change how we output the format or anything like that so even that is mostly testing the options configuration.

I also used the `console` backend which just prints a "summary" which is pretty small if you don't have tags and is pretty obvious when there are tags.

imiric (approved):

hynd:
Just to add another datapoint, AWS Cloudwatch supports the DD/NR tag format:
https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch-Agent-custom-metrics-statsd.html

PR #1965, opened 2021-04-14
Make possible for the cloud output to stop the engine

### LINKED ISSUE #1880: Stop `k6 run --out cloud` test when it gets aborted from the web app
This _should_ be relatively easy now that https://github.com/loadimpact/k6/pull/1869 is done. We might give outputs another optional interface, like `WithThresholds` and `WithRunStatusUpdates`, basically the missing `TODO` in the middle here :sweat_smile: https://github.com/loadimpact/k6/blob/a5d884155825b0b21406b5b25a508dc229830b49/output/types.go#L76-L90

Then the engine can pass every output that supports it (probably only the `cloud` for now) a reference to `engine.Stop()`: https://github.com/loadimpact/k6/blob/a5d884155825b0b21406b5b25a508dc229830b49/core/engine.go#L358-L363

This issue overrides https://github.com/loadimpact/k6/issues/1160

MStoykov:
This does add an interface that can be implemented by other outputs as well.

fixes #1880

TODOs:
- [ ] Figure out a good name. I kept changing what I think this should be called through the writing of this and am not certain I like any of the names much ... I would argue that the best one is `StopExecutionOnOutputSignal` as it makes it more accurate that we will over stop if the output couldn't be started, and that this is a signal from the output to *stop* not just that there was some error. Unfortunately, it's a bit of a mouthful
- [ ] Figure out how to write an integration test of this ... I have tested it manually and it works (and takes 3 minutes to test) but it will be nice to have a full integration test about this.
- [ ] Print which output signaled to close 

Related to the last point,  I decided to have an error be "returned" not just to call some function is so that we can *always* print it in some good known format instead of it being a responsibility of the output.

### DIFF
--- a/cloudapi/config_test.go
+++ b/cloudapi/config_test.go
@@ -48,6 +48,7 @@ func TestConfigApply(t *testing.T) {
 		PushRefID:                       null.NewString("PushRefID", true),
 		WebAppURL:                       null.NewString("foo", true),
 		NoCompress:                      null.NewBool(true, true),
+		StopOnError:                     null.NewBool(true, true),
 		MaxMetricSamplesPerPackage:      null.NewInt(2, true),
 		MetricPushInterval:              types.NewNullDuration(1*time.Second, true),
 		MetricPushConcurrency:           null.NewInt(3, true),
--- a/core/engine.go
+++ b/core/engine.go
@@ -138,8 +138,8 @@ func (e *Engine) StartOutputs() error {
 			thresholdOut.SetThresholds(e.thresholds)
 		}
 
-		if stopOut, ok := out.(output.WithStopRunWithError); ok {
-			stopOut.SetStopRunWithErrorFunc(
+		if stopOut, ok := out.(output.WithSetStopEngineRunWithError); ok {
+			stopOut.SetStopEngineRunWithError(
 				func(err error) {
 					e.logger.WithError(err).Error("Received error to stop from output")
 					// TODO don't stop if configured
--- a/core/engine.go
+++ b/core/engine.go
@@ -138,8 +138,8 @@ func (e *Engine) StartOutputs() error {
 			thresholdOut.SetThresholds(e.thresholds)
 		}
 
-		if stopOut, ok := out.(output.WithSetStopEngineRunWithError); ok {
-			stopOut.SetStopEngineRunWithError(
+		if stopOut, ok := out.(output.WithTestRunStop); ok {
+			stopOut.SetTestRunStopCallback(
 				func(err error) {
 					e.logger.WithError(err).Error("Received error to stop from output")
 					// TODO don't stop if configured
--- a/core/engine.go
+++ b/core/engine.go
@@ -142,7 +142,6 @@ func (e *Engine) StartOutputs() error {
 			stopOut.SetTestRunStopCallback(
 				func(err error) {
 					e.logger.WithError(err).Error("Received error to stop from output")
-					// TODO don't stop if configured
 					e.Stop()
 				})
 		}
--- a/core/engine.go
+++ b/core/engine.go
@@ -138,6 +138,15 @@ func (e *Engine) StartOutputs() error {
 			thresholdOut.SetThresholds(e.thresholds)
 		}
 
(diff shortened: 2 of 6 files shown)


na--:
I'll get back to you on the names later, but please make this a cloud output option, not a runtime option.

core/engine.go
```diff
@@ -138,6 +138,15 @@ func (e *Engine) StartOutputs() error {
 			thresholdOut.SetThresholds(e.thresholds)
 		}
 
+		if thresholdOut, ok := out.(output.WithStopRunWithError); ok {
+			thresholdOut.SetStopRunWithErrorFunc(
```
imiric, line 142: It has nothing to do with thresholds, right?
```suggestion
		if stopOut, ok := out.(output.WithStopRunWithError); ok {
			stopOut.SetStopRunWithErrorFunc(
```

output/types.go
```diff
@@ -80,8 +80,11 @@ type WithThresholds interface {
 	SetThresholds(map[string]stats.Thresholds)
 }
 
-// TODO: add some way for outputs to report mid-test errors and potentially
-// abort the whole test run
+// WithStopRunWithError TODO
```
imiric, line 83: Let's document this now.

imiric (commented):
LGTM, just some minor comments.

Re: the name, `StopRunWithError` sounds better to me than `StopExecutionOnOutputSignal`, as "signal" has the Unix signal connotation. Though I would add `Engine` to it since that's what happens, so `StopEngineRunWithError`?

And I agree with the need for an integration test, but it could be done in a follow up PR.

imiric (dismissed):

cloudapi/config_test.go
```diff
@@ -48,6 +48,7 @@ func TestConfigApply(t *testing.T) {
 		PushRefID:                       null.NewString("PushRefID", true),
 		WebAppURL:                       null.NewString("foo", true),
 		NoCompress:                      null.NewBool(true, true),
+		StopOnError:                     null.NewBool(true, true),
```
na--, line 51: I think this should be `false` by default. First, because it would be a minor breaking change. 

Second, because in most situations, people wouldn't run `k6 run --out cloud` with `--no-summary` or `--no-thresholds`, so even if the cloud starts rejecting their metrics mid-test, they'd still get some value out of the results by examining the local summary output.

core/engine.go
```diff
@@ -138,6 +138,15 @@ func (e *Engine) StartOutputs() error {
 			thresholdOut.SetThresholds(e.thresholds)
 		}
 
+		if stopOut, ok := out.(output.WithSetStopEngineRunWithError); ok {
+			stopOut.SetStopEngineRunWithError(
+				func(err error) {
+					e.logger.WithError(err).Error("Received error to stop from output")
+					// TODO don't stop if configured
```
na--, line 145: This `TODO` seems obsolete

output/types.go
```diff
@@ -80,8 +80,12 @@ type WithThresholds interface {
...
 }
 
-// TODO: add some way for outputs to report mid-test errors and potentially
-// abort the whole test run
+// WithSetStopEngineRunWithError is an output that requires to have function which will be called with an error from
+// the output which should trigger the Engine to stop
+type WithSetStopEngineRunWithError interface {
+	Output
+	SetStopEngineRunWithError(func(error))
+}
```
na--, line 88:
```suggestion
// WithTestRunStop is an output that can stop the Engine mid-test, interrupting 
// the whole test run execution if some internal condition occurs, completely
// independently from the thresholds. It requires a callback function which 
// expects an error and triggers the Engine to stop.
type WithTestRunStop interface {
	Output
	SetStopTestRunCallback(func(error))
}
```

cloudapi/config_test.go
MStoykov: This is just the test

na--: :man_facepalming: sorry :sweat_smile:

na-- (dismissed):

imiric (approved):

na-- (approved):

PR #1968, opened 2021-04-15
Support non-TLS HTTP/2 (i.e. H2C)

RoySunnySean007:
Http2.0 non-tls patch
Pls help to check if we did sth wrong, thx!

(edited by @na-- - this is the relevant issue: https://github.com/k6io/k6/issues/970)

### DIFF
--- a/js/runner.go
+++ b/js/runner.go
@@ -197,6 +197,11 @@ func (r *Runner) newVU(id int64, samplesOut chan<- stats.SampleContainer) (*VU, 
 		MaxIdleConnsPerHost: int(r.Bundle.Options.BatchPerHost.Int64),
 	}
 	_ = http2.ConfigureTransport(transport)
+	http2Transport, err := http2.ConfigureTransports(transport)
+	http2Transport.AllowHTTP = true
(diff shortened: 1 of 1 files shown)


js/runner.go
```diff
@@ -197,6 +197,11 @@ func (r *Runner) newVU(id int64, samplesOut chan<- stats.SampleContainer) (*VU,
 		MaxIdleConnsPerHost: int(r.Bundle.Options.BatchPerHost.Int64),
 	}
 	_ = http2.ConfigureTransport(transport)
+	http2Transport, err := http2.ConfigureTransports(transport)
```
na--, line 200: You shouldn't call both `http2.ConfigureTransport()` and `http2.ConfigureTransports()`, they do the same thing.

```suggestion
	http2Transport, err := http2.ConfigureTransports(transport)
```

Also, you may have to pass the `http2Transport` as the `VU.Transport` below, though I'm not completely sure about that... :man_shrugging:

na-- (changes requested):
As I [explained in the issue](https://github.com/k6io/k6/issues/970#issuecomment-820336657), we won't be able to merge this because it likely breaks k6 for every other use case...

PR #1971, opened 2021-04-16
Print stacktrace on exception in init context or setup/teardown

### LINKED ISSUE #994: Provide decent stack traces in setup code

MStoykov:
also exit with error code 107 for both cases.

fixes #994

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -47,6 +47,7 @@ import (
 	"github.com/loadimpact/k6/js"
 	"github.com/loadimpact/k6/lib"
 	"github.com/loadimpact/k6/lib/consts"
+	"github.com/loadimpact/k6/lib/types"
 	"github.com/loadimpact/k6/loader"
 	"github.com/loadimpact/k6/ui/pb"
 )
@@ -63,6 +64,7 @@ const (
 	invalidConfigErrorCode       = 104
 	externalAbortErrorCode       = 105
 	cannotStartRESTAPIErrorCode  = 106
+	scriptExceptionErrorCode     = 107
 )
 
 // TODO: fix this, global variables are not very testable...
@@ -124,6 +126,10 @@ a commandline interface for interacting with it.`,
 
 			initRunner, err := newRunner(logger, src, runType, filesystems, runtimeOptions)
 			if err != nil {
+				var serr types.ScriptException
+				if errors.As(err, &serr) {
+					return ExitCode{error: serr, Code: scriptExceptionErrorCode, Hint: "script exception"}
+				}
 				return err
 			}
 
@@ -345,6 +351,11 @@ func getExitCodeFromEngine(err error) ExitCode {
 		default:
 			return ExitCode{error: err, Code: genericTimeoutErrorCode}
 		}
+	}
+
+	var serr types.ScriptException
+	if errors.As(err, &serr) {
+		return ExitCode{error: serr, Code: scriptExceptionErrorCode, Hint: "script exception"}
 	}
 
 	return ExitCode{error: errors.New("engine error"), Code: genericEngineErrorCode, Hint: err.Error()}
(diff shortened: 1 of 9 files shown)


cmd/run.go
```diff
@@ -63,6 +64,7 @@ const (
 	invalidConfigErrorCode       = 104
 	externalAbortErrorCode       = 105
 	cannotStartRESTAPIErrorCode  = 106
+	scriptExceptionErrorCode     = 107
```
na--, line 67: This should probably be 108, since we already use 107 in https://github.com/k6io/k6/pull/1920 (we probably should also move these to the `consts` package eventually)

js/initcontext_test.go
```diff
@@ -131,7 +131,7 @@ func TestInitContextRequire(t *testing.T) {
 			fs := afero.NewMemMapFs()
 			assert.NoError(t, afero.WriteFile(fs, "/file.js", []byte(`throw new Error("aaaa")`), 0o755))
 			_, err := getSimpleBundle(t, "/script.js", `import "/file.js"; export default function() {}`, fs)
-			assert.EqualError(t, err, "Error: aaaa at file:///file.js:2:7(4)")
+			assert.EqualError(t, err, "Error: aaaa\n\tat file:///file.js:2:7(4)\n\tat reflect.methodValueCall (native)\n\tat file:///script.js:1:117(14)\n")
```
na--, line 134: This `reflect.methodValueCall (native)` is not a great improvement to not having a stack trace at all :sweat_smile: Can you please also add a test that has a stacktrace consisting of purely JS functions?

na-- (commented):
Mostly LGTM, though I'd like to play around with it a bit before it's merged. 

Should we try to release this in k6 v0.32.0?

lib/types/types.go
```diff
@@ -280,3 +280,10 @@ func GetDurationValue(v interface{}) (time.Duration, error) {
 		return time.Duration(n) * time.Millisecond, nil
 	}
 }
+
+// ScriptException is an interface to be used to represent and error as a result of script exception.
```
imiric, line 284:
```suggestion
// ScriptException is an interface used to represent an error as a result of a script exception.
```

js/initcontext_test.go
MStoykov: This is technically `script.js`-> `reflect.methodValueCall`->`file.js`

cmd/run.go
MStoykov: I don't think it really matters who gets `107` and who gets `108`. 

Whoever is merged second will have a conflict and will need to resolve it.

Given that IMO this is more likely to be merged now I prefer to no add a "hole" in the error codes ;)

MStoykov:
I specifically didn't add it as v0.32.0, because I also wonder, it's pretty small and the worst case is ... somehow it breaks some stacktrace case?!? 

On the other hand,today is the cut off date so ;)

cmd/run.go
na--: fair enough :+1:

imiric (commented):
LGTM, and works fine from what I could test.

Agree with Ned that a more complex test is needed. Maybe just replicating the example from #994 with `setup()` and `teardown()`?

:+1: for releasing this in v0.32.0.

imiric (approved):

na-- (approved):

PR #1973, opened 2021-04-16
Return better message on missing file with spaces

### LINKED ISSUE #1972: inconsistent error message for file names with/without spaces
Non-existent script files containing " " (space) character in the name are handled differently from non-existent files without spaces. :man_shrugging: This is easier to understand by looking at the screenshot.

Script files with spaces that exist work as expected. 

## Environment

- k6 version: `k6 v0.31.1 (2021-03-17T13:23:23+0000/e9d8349, go1.15.8, linux/amd64)`
- OS and version: Linux / debian testing (5.10.0-3-amd64)

## Expected Behavior
I expected the same error message to be printed in both cases.

## Actual Behavior
See screenshot above.

`k6 run file\ that\ doesnt\ exist\ and\ has\ spaces.js`

MStoykov:
fixes #1972

### DIFF
--- a/loader/readsource.go
+++ b/loader/readsource.go
@@ -67,6 +67,11 @@ func ReadSource(
 	pwdURL := &url.URL{Scheme: "file", Path: filepath.ToSlash(filepath.Clean(pwd)) + "/"}
 	srcURL, err := Resolve(pwdURL, filepath.ToSlash(src))
 	if err != nil {
+		var noSchemeError noSchemeRemoteModuleResolutionError
+		if errors.As(err, &noSchemeError) {
+			// TODO maybe try to wrap the original error here as well, without butchering the message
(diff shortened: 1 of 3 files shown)


loader/readsource.go
```diff
@@ -67,6 +67,11 @@ func ReadSource(
 	pwdURL := &url.URL{Scheme: "file", Path: filepath.ToSlash(filepath.Clean(pwd)) + "/"}
 	srcURL, err := Resolve(pwdURL, filepath.ToSlash(src))
 	if err != nil {
+		var noSchemeError noSchemeRemoteModuleResolutionError
+		if errors.As(err, &noSchemeError) {
+			// TODO maybe try to wrap the original error here as well, without butchering the message
```
MStoykov, line 72: I copied this from the other place this is made, below, but looking at it ... I think it is done?

na-- (approved):
LGTM, this seems to have slipped through the cracks as well :disappointed:

imiric (approved):

Issue #342, opened 2017-10-19
More robust integration tests


StephenRadachy:
Integration tests should be added that leverage the existing sample files - so we're not just checking that the cli can write to stdout.

liclac:
Some integration tests are ran as a part of the `js` test suite, but I'd definitely welcome this. It'd be best done as a target in the Makefile, I think.

PR #1975, opened 2021-04-16
Fix color handling, generate the summary with JS, improve CI

na--:
Read commit by commit. This should be merged after we release k6 v0.32.0, I'm just polishing it and pushing it, so it's easy to merge after the release.

This PR should fix https://github.com/k6io/k6/issues/1806 and close https://github.com/k6io/k6/issues/1806, as well as make several other minor improvements, mostly about various dependencies (https://github.com/k6io/k6/issues/1933).

It also introduces a minor breaking change, though I'm not sure if it's even worth noting in the release notes... k6 will no longer highlight the output of several sub-commands: `pause`, `resume`, `scale`, `stats`, and `status`.

Finally, because I wanted to use the new [file embedding capabilities](https://golang.org/doc/go1.16#embed) in Go 1.16, I bumped the minimum required version to that. I switched the xk6 tests to run with the latest stable Go version and the tip. For the main test suite, I temporarily disabled the 1.15 tests, but also added testing with the tip (which is non-blocking, so it won't prevent a PR merge or a release). This shouldn't be a problem, since we will only release this in k6 v0.33.0, ~2.5 months from now. But it gave me the opportunity to also drop the go.rice dependency :tada:


cmd/status.go
```diff
@@ -46,8 +45,8 @@ func getStatusCmd(ctx context.Context) *cobra.Command {
 			if err != nil {
 				return err
 			}
-			ui.Dump(stdout, status)
-			return nil
+
+			return yamlPrint(stdout, status)
```
MStoykov, line 49: Maybe this is a good case for wrapping errors as "could not marshal YAML:" will be quite strange error for someone calling `k6 pause` or co.

MStoykov (commented):
Seems fine on first read, but we need to test this a lot ... especially on windows

cmd/status.go
na--: Yeah, though to be fair, the new state isn't any worse than the previous behavior where it would have printed just the raw YAML marshaling error, without even the `could not marshal YAML:` in front of it :sweat_smile: but yeah, I may wrap these errors locally as well...

na--:
I converted this back to draft because I want to make a few more changes, prompted by https://github.com/k6io/jslib.k6.io/pull/32

First, I think I should rename the `uiState` sent to `handleSummary()` to the more generic `state`, and have it include things like the total test run duration. It is currently in the summary, just not exposed in the JS object directly: https://github.com/k6io/k6/blob/47a0b493cce0020d445a47edbca79c25ab5a550f/lib/runner.go#L110

Also, `AreColorsDisabled` (i.e. `NoColors`) should probably not be in `data.state` and should instead be moved to the `data.options`.

js/summary.go
```diff
@@ -164,6 +180,12 @@ func summarizeMetricsToObject(data *lib.Summary, options lib.Options) map[string
 		// TODO: improve when we can easily export all option values, including defaults?
 		"summaryTrendStats": options.SummaryTrendStats,
 		"summaryTimeUnit":   options.SummaryTimeUnit.String,
+		"noColor":           data.NoColor, // TODO: move to the (runtime) options
+	}
+	m["state"] = map[string]interface{}{
+		"isStdOutTTY":     data.UIState.IsStdOutTTY,
+		"isStdErrTTY":     data.UIState.IsStdErrTTY,
+		"testRunDuration": int64(data.TestRunDuration.Round(time.Second) / time.Second),
```
imiric, line 188: Should we use millisecond precision here instead?

js/summary.go
```diff
@@ -21,21 +21,31 @@
...
-	"go.k6.io/k6/ui"
 )
 
-// TODO: move this to a separate JS file and use go.rice to embed it
-const summaryWrapperLambdaCode = `
+// Copied from https://github.com/k6io/jslib.k6.io/tree/master/lib/k6-summary
+//go:embed summary.js
+var jslibSummaryCode string //nolint:gochecknoglobals
+
+// TODO: move this or parts of it to a separate JS file and embed it like above
```
imiric, line 40: :+1: but why not do this now? :) Reading JS in a Go file is cumbersome, which is fine for tests and small pieces of code, but this would be better in a standalone JS file.

js/summary.go
```diff
@@ -21,21 +21,31 @@
...
 	"github.com/dop251/goja"
 	"go.k6.io/k6/js/common"
 	"go.k6.io/k6/lib"
 	"go.k6.io/k6/stats"
-	"go.k6.io/k6/ui"
 )
 
-// TODO: move this to a separate JS file and use go.rice to embed it
-const summaryWrapperLambdaCode = `
+// Copied from https://github.com/k6io/jslib.k6.io/tree/master/lib/k6-summary
```
imiric, line 36: This means we can deprecate the jslib, right?

cmd/cloud.go
```diff
@@ -330,7 +333,8 @@ This will execute the test on the k6 cloud service. Use "k6 login cloud" to auth
 				return ExitCode{error: errors.New("Test progress error"), Code: cloudFailedToGetProgressErrorCode}
 			}
 
-			fprintf(stdout, "     test status: %s\n", ui.ValueColor.Sprint(testProgress.RunStatusText))
+			valueColor := getColor(noColor || !stdoutTTY, color.FgCyan)
```
MStoykov, line 336: Given how now you repeat `color.FgCyan` a lot I would argue it should probably be a constant somewhere ... probably named `valueColorColor` :P

js/summary.go
```diff
@@ -168,8 +168,9 @@ func summarizeMetricsToObject(data *lib.Summary, options lib.Options) map[string
 		"noColor":           data.NoColor, // TODO: move to the (runtime) options
 	}
 	m["state"] = map[string]interface{}{
-		"isStdOutTTY": data.UIState.IsStdOutTTY,
-		"isStdErrTTY": data.UIState.IsStdErrTTY,
+		"isStdOutTTY":     data.UIState.IsStdOutTTY,
+		"isStdErrTTY":     data.UIState.IsStdErrTTY,
+		"testRunDuration": int64(data.TestRunDuration.Round(time.Second) / time.Second),
```
MStoykov, line 188: maybe provide it as is or in milliseconds? I don't think it will be used much but if someone has 300ms test run, having It reported as taking 0s is only *technically* correct ;)

MStoykov (approved):
LGTM, I had a couple of nitpicks, but if it works .... I am fine with merging it and my 5 minutes of testing didn't show any problems so :+1:

js/summary.go
imiric: Same as https://github.com/k6io/k6/pull/1975#discussion_r633544162

js/summary.go
na--: :+1: I'll fix this, and I think I'll even rename it to `testRunDurationMs` and make it a float for maximum precision

cmd/ui.go
```diff
@@ -81,6 +83,26 @@ func (w *consoleWriter) Write(p []byte) (n int, err error) {
 	return origLen, err
 }
 
+// getColor() returns the requested color, or an uncolored object, depending on
+// the value of noColor. The explicit EnableColor() and DisableColor() are
+// needed because the library checks os.Stdout itself otherwise...
+func getColor(noColor bool, attributes ...color.Attribute) *color.Color {
```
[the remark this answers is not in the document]
na--: This is very difficult to do currently with all of these globals, I think it falls under https://github.com/k6io/k6/issues/342 and https://github.com/loadimpact/k6/issues/883. The end goal is to almost completely remove the globals, i.e. have something that passes `os.Environ()` and the CLI flags as parameters, which will allow us to have sane tests as well.

.github/workflows/all.yml
```diff
@@ -84,6 +85,32 @@ jobs:
...
+          echo "$HOME/go/bin" >> "$GITHUB_PATH"
+          echo "$HOME/sdk/gotip/bin" >> "$GITHUB_PATH"
+      - name: Run tests
+        run: |
+          set -x
+          which go
+          go version
+          export GOMAXPROCS=2
+          args=("-p" "2" "-race")
+          go test "${args[@]}" -timeout 800s ./...
```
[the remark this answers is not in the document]
na--: Hmm yeah, I thought we had this, will add it to the CI configs :+1:

js/summary.go
na--: :+1: yeah, no reason not to do it now, will do so

js/summary.go
na--: No? jslib has a ton of other libraries. Besides, having `textSummary()` in there is very helpful for users that want to define their own `handleSummary()` function, e.g. to emit a json or a junit file, but they also want to continue getting the text summary at the end of the test. They can just `import { textSummary } from 'https://jslib.k6.io/lib/k6-summary/0.0.1/index.js'` and use that in `handleSummary()`'s result.

cmd/cloud.go
na--: It's already a constant :stuck_out_tongue: It's used in like 5 places across 2 modules, and not always for values (e.g. the banner), so I don't want to define a constant for that.

cmd/ui.go
```diff
@@ -81,6 +83,26 @@ func (w *consoleWriter) Write(p []byte) (n int, err error) {
 	return origLen, err
 }
 
+// getColor() returns the requested color, or an uncolored object, depending on
```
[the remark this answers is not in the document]
na--: Resolved by the latest [force-push](https://github.com/k6io/k6/compare/25cd5281f4b120f1a179b569483c9a4278b7b248..c7e8a7b34824aa8619886c96129dbbbdd4d6c6b6)

js/summary.go
na--: Resolved by the latest [force-push](https://github.com/k6io/k6/compare/25cd5281f4b120f1a179b569483c9a4278b7b248..c7e8a7b34824aa8619886c96129dbbbdd4d6c6b6)

MStoykov (approved):

js/summary.go
na--: Resolved by https://github.com/k6io/k6/pull/1975/commits/6f6bfac43f69123c697c51740bf24899199609e8

.github/workflows/all.yml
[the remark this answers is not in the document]
na--: Hmm we also need [`fail-fast: false`](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast)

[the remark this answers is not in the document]
na--: Fixed by https://github.com/k6io/k6/pull/1975/commits/6778f1267f9ff248284d96e10cb23876b5ec818a?

MStoykov (approved):

cmd/ui.go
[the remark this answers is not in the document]
imiric: I meant high-level as in user acceptance tests, similar to the xk6 ones we have now, running `k6` itself in a shell, checking the output, etc.

But that requires coming up with the framework and CI pipeline, so it's fine if we do it later as part of #342.

js/summary.go
imiric: I didn't mean deprecating jslib.k6.io altogether :smile:, just this lib.

I wasn't aware there were differences between the version committed here and the one on jslib. Though since we already bundle it in the binary, couldn't we expose it as a built-in JS module and avoid the remote loading? :thinking:

na--: There aren't any differences, except we don't bundle the JUnit code (we'll probably split it apart in jslib as well soon). And we can potentially expose it as a JS module, if we figure out a nice module name, but that will be a separate issue/PR

js/runner.go
```diff
@@ -344,7 +345,8 @@ func (r *Runner) HandleSummary(ctx context.Context, summary *lib.Summary) (map[s
 	}()
 	*vu.Context = ctx
 
-	handleSummaryWrapperRaw, err := vu.Runtime.RunString(summaryWrapperLambdaCode)
+	wrapper := strings.Replace(summaryWrapperLambdaCode, "/*JSLIB_SUMMARY_CODE*/", jslibSummaryCode, 1)
```
imiric, line 348: I'm probably missing it, but what's the reason for having a wrapper and doing this replace? Since both `summaryWrapperLambdaCode` and `jslibSummaryCode` are static, couldn't we just bundle a single JS file?

stats/stats_test.go
```diff
@@ -29,120 +29,6 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-func TestMetricHumanizeValue(t *testing.T) {
```
imiric, line 32: These were very important tests IMO that we lose in the conversion to JS. We should probably add some for the JS implementation.

imiric (approved):
LGTM, just a couple of minor notes.

js/runner.go
na--: We probably could, but this insulates us from future changes in the jslib summary code, and it also keeps the scope a bit cleaner :man_shrugging:

stats/stats_test.go
na--: I replicated (and IIRC even expanded a bit) these tests in https://github.com/k6io/jslib.k6.io/blob/master/tests/summary.js when I was writing the JS summary generation code.

js/runner.go
na--: Ah, sorry, misread your question. I don't want to bundle a single JS file since I want it to be easy to update the `summary.js` when the one in jslib is updated

stats/stats_test.go
imiric: Ah, neat, I forgot about those. I suppose it's fine then, though running them as part of the k6 tests would give us more confidence, as they're run more frequently and use the latest k6 version. But this goes back to the point of eventually deprecating the jslib version, which we can consider later.

na--: In k6 we have somewhat higher-level tests in https://github.com/k6io/k6/blob/master/js/summary_test.go - they are like integration tests for the JS summary code, so I'd rather expand those than to write more low-level ones.

PR #1976, opened 2021-04-19
Pin Codecov bash uploader version and check its hash

imiric:
After the [recent breach](https://about.codecov.io/security-update/) this seems like the least we should do for now.

### DIFF
--- a/.github/workflows/all.yml
+++ b/.github/workflows/all.yml
@@ -90,8 +90,6 @@ jobs:
         go-version: [1.16.x]
         platform: [ubuntu-latest, windows-latest]
     runs-on: ${{ matrix.platform }}
-    env:
-      CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
     steps:
       - name: Checkout code
         uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/all.yml
```diff
@@ -122,9 +120,14 @@ jobs:
           grep -h -v "^mode:" *.coverage >> coverage.txt
           rm -f *.coverage
       - name: Upload coverage to Codecov
+        env:
+          CODECOV_BASH_VERSION: 1.0.1
+          CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
         run: |
+          curl -fsSL --remote-name-all "https://raw.githubusercontent.com/codecov/codecov-bash/${CODECOV_BASH_VERSION}/{codecov,SHA512SUM}"
```
na--, line 127: I think `{codecov,SHA512SUM}` should be outside of the quotes to expand properly. 

But in any case, I think this validation is wrong - if someone can replace `https://raw.githubusercontent.com/codecov/codecov-bash/1.0.1/codecov`, they can replace `https://raw.githubusercontent.com/codecov/codecov-bash/1.0.1/SHA512SUM` :sweat_smile: Instead, we should copy-paste that SHA value in our CI and statically compare against it, not against a remotely computed one.

imiric: You're right... See 2a960212.

> I think `{codecov,SHA512SUM}` should be outside of the quotes to expand properly.

Also right, but apparently `curl` itself supports this notation :tada:, so it worked before too :smile: In any case, not needed anymore since we're only downloading the script itself now.

na-- (approved):

MStoykov (approved):

PR #1978, opened 2021-04-19
Update installation instructions in README

imiric:
Let's wait until we're ready to announce this before merging.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,6 +104,27 @@ sudo dnf install k6    # use yum install --nogpgcheck k6 for older distros (e.g.
 Note that the `gnupg2` package is required for signature verification.
 
 
+#### Migrating from Bintray
+
+The Bintray repositories will be [shutdown after May 1st, 2021](https://jfrog.com/blog/into-the-sunset-bintray-jcenter-gocenter-and-chartcenter/). If you previously added them you will have to add our repositories following the instructions above and should delete the Bintray ones.
+
+For Debian-based distributions, you can run:
+
+```bash
+sudo sed -i '/dl\.bintray\.com\/loadimpact\/deb/d' /etc/apt/sources.list
+sudo apt-key del 379CE192D401AB61
+sudo apt-get update
+```
+
+Or delete the repository file if you added it to `/etc/apt/sources.list.d/`.
+
+And for rpm-based ones, delete the repository file in `/etc/yum.repos.d/`. If you followed the [official installation instructions](https://k6.io/docs/getting-started/installation/#red-hat-centos), this should be:
+
+```bash
+sudo rm /etc/yum.repos.d/bintray-loadimpact-rpm.repo
+```
+
+
 ### Docker
 
 ```bash
--- a/README.md
+++ b/README.md
@@ -68,28 +68,41 @@ brew install k6
 
 ### Windows
 
-You can manually download and install the [official `.msi` installation package](https://dl.bintray.com/loadimpact/windows/k6-v0.31.1-amd64.msi) or, if you use the [chocolatey package manager](https://chocolatey.org/), follow [these instructions](https://bintray.com/repo/buildSettings?repoPath=%2Floadimpact%2Fchoco) to set up the k6 repository.
+If you use the [Chocolatey package manager](https://chocolatey.org/) you can install the unofficial k6 package with:
+
+```
+choco install k6
+```
+
+Otherwise you can manually download and install the [latest official `.msi` package](https://dl.k6.io/msi/k6-latest-amd64.msi).
 
 ### Linux
 
-**Notice: Because [Bintray is being shutdown](https://jfrog.com/blog/into-the-sunset-bintray-jcenter-gocenter-and-chartcenter/) we are going to start self-hosting our packages soon, before k6 v0.32.0. This means you will have to re-install them, since the old .rpm and .deb repos will stop working.**
-
-For Debian-based Linux distributions, you can install k6 from the private deb repo like this:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -68,29 +68,34 @@ brew install k6
...
 
-For Debian-based Linux distributions, you can install k6 from the private deb repo like this:
+### Linux
+
+For Debian-based Linux distributions like Ubuntu, you can install k6 from the private deb repo like this:
 
 ```bash
-sudo apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 379CE192D401AB61
-echo "deb https://dl.bintray.com/loadimpact/deb stable main" | sudo tee -a /etc/apt/sources.list
+curl -s https://dl.k6.io/key.gpg | sudo apt-key add -
```
imiric, line 84: I don't think we should bother uploading the key to any keyservers, as they're usually slow and flaky. However, we lose any potential web of trust benefits, and we're not specifying a key ID here, which could lead to users installing the wrong key if `dl.k6.io` is compromised.

Any thoughts on a middle ground solution?

README.md
```diff
@@ -68,29 +68,34 @@ brew install k6
...
+echo "deb https://dl.k6.io/deb stable main" | sudo tee /etc/apt/sources.list.d/k6.list
```
imiric, line 85: Should we include any information for users who already added the Bintray repo? I'll test if it's a smooth migration in that case, but we should probably mention it.

na--: Yes, we should probably mention something about the migration (and any manual steps that are required), both here and in the docs.

README.md
na--: Hmm I don't have a problem with uploading the key to a keyserver or two, though I'm not sure if we need to request users to explicitly install it? Doesn't `apt-get install k6` ask users to trust our key the first time they try to install it? If so, we can just add a comment like this:
> apt-get will ask you to trust our key <keyid>, confirm it or download it from https://dl.k6.io/key.gpg and import it

imiric: > Doesn't `apt-get install k6` ask users to trust our key the first time they try to install it?

No, apt is not that smart :sweat:, and fails at `apt-get update` if the key doesn't exist:

```
Err:1 https://dl.k6.io/deb stable InRelease
  The following signatures couldn't be verified because the public key is not available: NO_PUBKEY C780D0BDB1A69C86
```

But I added the key to `keyserver.ubuntu.com` and `pgp.mit.edu`, so I'm using `--recv-keys` in the latest push, with a fallback to `curl | apt-key add` if that fails.

README.md
imiric: Let me know if 90623b17 is clear enough. I'll add something similar to the docs.

From my tests there's no problem with having both the Bintray and our repos added at the same time, though installation will default to Bintray, I think because it was added first. Once it starts failing, it falls back to any alternative repos that carry the same package, so it's mostly a smooth process, for Debian/apt at least. On CentOS any `yum` operation fails if the Bintray repo fails, but once it's removed then it picks up the package from `dl.k6.io`.

MStoykov (approved):

README.md
```diff
@@ -68,29 +68,67 @@ brew install k6
 
 ### Windows
 
-You can manually download and install the [official `.msi` installation package](https://dl.bintray.com/loadimpact/windows/k6-v0.31.1-amd64.msi) or, if you use the [chocolatey package manager](https://chocolatey.org/), follow [these instructions](https://bintray.com/repo/buildSettings?repoPath=%2Floadimpact%2Fchoco) to set up the k6 repository.
+If you use the [Chocolatey package manager](https://chocolatey.org/):
```
na--, line 71: Since we haven't taken over this package yet, we should add a disclaimer here that it's not an official package.

README.md
```diff
@@ -68,29 +68,67 @@ brew install k6
...
-wget https://bintray.com/loadimpact/rpm/rpm -O bintray-loadimpact-rpm.repo
-sudo mv bintray-loadimpact-rpm.repo /etc/yum.repos.d/
-sudo dnf install k6   # use yum instead of dnf for older distros
+sudo dnf install https://dl.k6.io/rpm/repo.rpm   # use yum instead of dnf for older distros
+sudo dnf install k6
+```
+
+Note that the `gnupg2` package is required for signature verification.
+
+CentOS versions older than 8 don't support the PGP V4 signature we use, so you'll need to disable the verification by installing k6 with:
```
na--, line 106: This may be a bit too prominent here, maybe add a single line comment above, similar to `use yum instead of dnf for older distros`:
```
sudo dnf install https://dl.k6.io/rpm/repo.rpm   # use yum instead of dnf for older distros
sudo dnf install k6 # use yum install --nogpgcheck k6 for older distros (e.g. CentOS 7) without PGP V4 signature support
```

na--:
LGTM besides these minor issues I noted inline :tada: I'll now play around with the various OS packages and repos

README.md
imiric: Sure, see 7e74e888. Not a real disclaimer, just a mention of "unofficial", which I think is good enough since we emphasize "official" for the MSI.

Though I think either way should be fine since we'll be taking it over in a few weeks anyway.

README.md
imiric: Sure, see 7e74e888. It's a bit cluttered, but better than mentioning it separately I guess.

na-- (approved):
LGTM, though please don't merge this quite yet.

MStoykov (approved):

PR #1980, opened 2021-04-19
Fix paused script not being marked as ended if stopped before resumed

MStoykov:


na-- (approved):

imiric (approved):
LGTM. A test would be nice to have if you have time.

PR #1982, opened 2021-04-22
Drop Collector interface and helpers

MStoykov:


cmd/outputs.go
```diff
@@ -37,13 +36,11 @@ import (
 	"github.com/loadimpact/k6/output/influxdb"
 	"github.com/loadimpact/k6/output/json"
 	"github.com/loadimpact/k6/output/statsd"
-	"github.com/loadimpact/k6/stats"
 
 	"github.com/k6io/xk6-output-kafka/pkg/kafka"
 )
 
 // TODO: move this to an output sub-module after we get rid of the old collectors?
```
MStoykov, line 43: I couldn't figure out a good name for the module so I left it for now. Suggestions are welcome.

For the record, the reason we can't have it in `output` is that this way anything that is using any of the helpers defined in `output` will import it, which is all the outputs and then for the internal ones we import them in the same module - making an import cycle

na-- (approved):

cmd/outputs.go
na--: I guess it can be in a sub-package of `output`, but that doesn't seem very nice. I guess it can remain here until we actually _need_ to move it, and when we do, we'll figure out where... :sweat_smile:

imiric: It would be slightly cleaner somewhere under `output`, but it doesn't bother me leaving it as is, since it's only called from here. So yeah, let's leave it for later if needed.

imiric (approved):

PR #1984, opened 2021-04-22
Explicitly disable GPG check for RPM repo metadata

imiric:
After some investigation, we decided to disable this explicitly to avoid it failing for users who have `repo_gpgcheck` enabled globally.

Our RPM packages are signed, but we're not planning to sign repo metadata since enabling this check is known to [cause](https://bugzilla.redhat.com/show_bug.cgi?id=1768206) [issues](https://bugzilla.redhat.com/show_bug.cgi?id=1636743) and it's not enabled even for main Fedora repositories.

### DIFF
--- a/packaging/k6-rpm-repo.spec
+++ b/packaging/k6-rpm-repo.spec
@@ -1,5 +1,5 @@
 Name:           k6-rpm
-Version:        0.0.1
+Version:        0.0.2
 Release:        1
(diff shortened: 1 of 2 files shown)


MStoykov:
can you also bump the version in the spec ?

MStoykov (approved):

Issue #1930, opened 2021-03-28
Support to use gRPC Server Reflection


hojongs:
## Feature Description
gRPC Server reflection allow gRPC client to gRPC call without proto file.
https://github.com/grpc/grpc/blob/master/doc/server-reflection.md

ghz, is another gRPC load testing tool, supports this feature.
https://ghz.sh/docs/options#--proto

I saw this issue https://github.com/k6io/k6/issues/441#issuecomment-686464551. in the issue, gRPC Server reflection feature is mentioned.
So I looked k6 issues&PRs about this feature but there isn't.
Do you have a plan for detail schedule for it?

Indeed, k6 is using this library https://github.com/jhump/protoreflect and it seems already to contain server reflection feature. https://github.com/jhump/protoreflect#grpc-server-reflection

## Suggested Solution (optional)

I'm not sure but how about using server reflection feature in `jhump/protoreflect` ?
Maybe, Can this API be used? https://pkg.go.dev/github.com/jhump/protoreflect/grpcreflect#NewClient

na--:
IIRC, we didn't add server reflection initially because we tried to make the gRPC pull request an MVP of sorts, adding the bare minimum of features to try and minimize any breaking changes that would result. We didn't have [xk6](https://github.com/k6io/xk6) as an experimental proving ground yet, so it was directly merged into the k6 core with a minimal amount of iteration :sweat_smile: 

But yes, server reflection is something we'd like to support. We don't plan on working on it ourselves soon, but we'll probably accept a good PR if someone submits it. 

If you're interested in working on this, please share your implementation ideas here, before opening a PR. I haven't looked into it, but this should probably be an optional feature, maybe a parameter to the gRPC `Client` constructor? :man_shrugging:

joshcarp:
Hey,
I've implemented a change that solves this issue.
I've implemented a `grpc.Client.reflect` method that is similar to the `grpc.Client.connect` method. 
A couple of things:
- I didn't use `github.com/jhump/protoreflect` because it uses the old proto api, and the only thing that we need is to get the file descriptor protos, after that the processing is the same as `load` method.

juwatanabe:
Would love to have seen this functionality in the latest dot release. Any reason why this wasn't included in v0.33?

na--:
[...]

Issue #1846, opened 2021-02-08
gRPC performing at nearly 50% of HTTP equivalent


thecodejunkie:
Hi,

I was chatting with @simskij, on the Gophers Slack about some performance observations I was having when using k6 to compare gRPC and HTTP endpoints (in the same service), and he suggested I post here as well to get some insights from the performance wizards :)

Both endpoints have a similar setup (for testing purpose) and for gRPC I get about 32k req/s and for HTTP I am seeing about 55k req/s. I was expecting gRPC to be a bit higher, but I might be wrong?

I run my service and k6 tests on the same machine, a 2.2Ghz 6 core i7 MBP with 32GB RAM. Below are the test results

```sh
          /\      |‾‾| /‾‾/   /‾‾/
     /\  /  \     |  |/  /   /  /
    /  \/    \    |     (   /   ‾‾\
   /          \   |  |\  \ |  (‾)  |
  / __________ \  |__| \__\ \_____/ .io

  execution: local
     script: http.js
     output: -

  scenarios: (100.00%) 1 scenario, 50 max VUs, 40s max duration (incl. graceful stop):
           * default: 50 looping VUs for 10s (gracefulStop: 30s)


running (10.0s), 00/50 VUs, 552787 complete and 0 interrupted iterations
default ✓ [======================================] 50 VUs  10s

     ✓ is status 200

     checks.....................: 100.00% ✓ 552787 ✗ 0
     data_received..............: 81 MB   8.1 MB/s
     data_sent..................: 44 MB   4.4 MB/s
     http_req_blocked...........: avg=3.09µs   min=0s      med=2µs      max=35.3ms  p(90)=3µs    p(95)=3µs
     http_req_connecting........: avg=88ns     min=0s      med=0s       max=1.58ms  p(90)=0s     p(95)=0s
     http_req_duration..........: avg=772.04µs min=60µs    med=510µs    max=75.5ms  p(90)=1.6ms  p(95)=2.23ms
     http_req_receiving.........: avg=31.77µs  min=8µs     med=17µs     max=74.58ms p(90)=30µs   p(95)=36µs
     http_req_sending...........: avg=13.45µs  min=4µs     med=8µs      max=37.86ms p(90)=15µs   p(95)=18µs
     http_req_tls_handshaking...: avg=0s       min=0s      med=0s       max=0s      p(90)=0s     p(95)=0s
[...]

Issue #785, opened 2018-09-28
Per-VU init lifecycle function


na--:
As mentioned in the comments of https://github.com/loadimpact/k6/issues/784, there isn't a very good way for users to implement per-VU initialization, since we don't support making HTTP and websocket calls in the init script phase (for good reasons, also mentioned in that issue).

The two current workarounds are:
1) Using the `__ITER` [execution context variable](https://docs.k6.io/docs/execution-context-variables) and initializing things in the first iteration of each VU. The problem with this is that the longer first iteration could then potentially skew the `iteration_duration` metric by quite a lot and it would also eat up the script execution `duration` for initialization purposes. 
2) Using `setup()` to initialize things for all VUs, return the data in an array, and then use the `__VU` [execution context variable](https://docs.k6.io/docs/execution-context-variables) so each VU uses its own data. That's a bit awkward, but it won't skew the `iteration_duration` metric. Another potential drawback is that in the distributed execution context, all requests would be made from a single IP address, which isn't ideal if we want the init code to do login or sign-up or something like that.

One potential solution to this issue is to add a new per-VU lifecycle function called `init(setupData)` / `initVU(setupData)` / `setupVU(setupData)` or something like that. It will be called once in each VU, after the global `setup()` is done, so it will receive the data that `setup()` returned. It will also be called before the actual load test (looping execution of the `default` function) starts, so it won't be counted as part of the script execution `duration` or `stages`.

That per-VU function will be executed in the actual VU runtimes, unlike `setup()` and `teardown()`, so it can save any state as global variables in the VU for later access by the `default` function. And having it as a separate function won't interfere with the actual init phase when we get the script options, won't skew the `iteration_duration` metric for the first iteration and won't take up some of the configured script execution duration. We can even tag any resulting metrics appropriately, like we do with the ones from `setup()` and `teardown()` now.

Details to be considered:
 - Errors (uncaught exceptions) in one of the per-VU functions should probably abort the whole script execution. 
- Since all those init functions have to finish before we can start the test, we should probably have a separate timeout for them, like we have with `setupTimeout` and `teardownTimeout`.
- Since we'd execute this new function when initializing the VUs, for hundreds of VUs we probably don't want to have a fully sequential or fully-parallel initialization, so some sort of configurable rate-limiting would be useful. We can probably move and reuse the [`SlotLimiter`](https://github.com/loadimpact/k6/blob/master/js/modules/k6/http/limiter.go) we use in `http.batch()` requests or use something like it.

[...]

PR #1987, opened 2021-04-24
Add grpc reflection functionality to grpc client

joshcarp:
* Add gRPC reflection via the `grpc.Client().reflect` function.
* Use the reflection API to retrieve method descriptors from service
* Factor out `convertToMethodInfo` to convert file descriptors into `[]MethodInfo` and `protoreflect.MethodDescriptor`. This is used in both `Load` and `Reflect` 
* Add `samples/grpc_reflection.js` example of how to use this new function

Help needed:
- I don't know how to test this properly, any advice would be appreciated

This PR closes: #1930


na-- (commented):
Thanks for making this PR and sorry for the delay in reviewing!

I am not sure if having a `client.reflect()` method in the init context is the right approach here. Generally, we try to avoid network calls in the init context... On the other hand, this is somewhat similar to `client.load()`, so I am not sure if we can't make an exception this time :man_shrugging: 

Other than this API design issue, I skimmed the rest of the code and it looks good. As I mentioned on slack, we've already closed the k6 v0.32.0 release window, so this can only be merged after we release it around May 10th. So we'll discuss the API design internally (or in the issue) and get back to you around then.

joshcarp:
> I am not sure if having a client.reflect() method in the init context is the right approach here. Generally, we try to avoid network calls in the init context...

Yeah I was trying to avoid that but the only other option I could come up with was to do it in the VU loop, but then that'll cause extra load on the API side because the reflection api will be getting called every loop.

js/modules/k6/grpc/reflect.go
```diff
@@ -0,0 +1,41 @@
...
+		},
+		Options: lib.Options{UserAgent: null.String{NullString: sql.NullString{
+			String: "k6",
+			Valid:  true,
+		}}, HTTPDebug: null.String{}},
+		Dialer: netext.NewDialer(
+			net.Dialer{
+				Timeout:   2 * time.Second,
+				KeepAlive: 10 * time.Second,
+			}, netext.NewResolver(net.LookupIP, 0, types.DNSfirst, types.DNSpreferIPv4)),
```
imiric, line 39: Hhmm shouldn't all of this inherit from the global settings (`state.Options`)? The `UserAgent` for sure should be consistent with the one set for HTTP and [eventually WS :sweat_smile:](https://github.com/k6io/k6/issues/2036), and the Dialer/TLS/DNS settings also seem out of place here, since these should be configurable and not static.

Hopefully this wouldn't be a big problem and we can fix it before #883.

samples/grpc_reflection.js
```diff
@@ -0,0 +1,21 @@
+import grpc from 'k6/net/grpc';
+import { check } from "k6";
+
+let client = new grpc.Client();
+client.reflect("127.0.0.1:10000", { plaintext: true })
```
imiric, line 5: I agree with @na-- that having this in the init context is problematic. Keep in mind that the init context is evaluated at least 4 times (for getting the `options`, for `setup()`, `teardown()` and `handleSummary()` even if these aren't defined in the script), and then for every created VU. So it's wasteful to have the gRPC connection/reflection happen here multiple times, and it could impact VU initialization time.

To me it would make sense to do this in `setup()`, but since we can't easily share data from there it wouldn't be feasible for distributed execution. Maybe we should prioritize having a per-VU init function (#785), though even in that case it would be wasteful for each VU to do the reflection, and the service definitions (whether loaded statically or reflected) should be shared among VUs anyway. So it seems like we need some kind of `SharedArray` and per-VU init to make this happen, which is not trivial to add.

And from an API perspective, I think reflection should be an implementation detail, so also agree with @na-- that it would make sense for this to be an [optional parameter to the `Client` constructor](https://github.com/k6io/k6/issues/1930#issuecomment-808890461), instead of a separate method that scripts have to call.

Could it be added as an argument to `client.connect()` instead, in which case we could cache the definitions internally and avoid the init context problems? :thinking:

imiric (changes requested):
Thanks a lot for your contribution @joshcarp! Are you still available to work on this PR? We'd like to try to get it merged for the upcoming v0.33.0 release (planned for end of June), but after taking a deeper look I'm not sure we'd have time to fix the leftover things here.

As for tests, we should probably add reflection support to the [test server](https://github.com/k6io/k6/blob/da5eebc638d9917cedfe98e232a7eed4cb69a157/lib/testutils/httpmultibin/httpmultibin.go#L284) and test against that.

samples/grpc_reflection.js
imiric: That could work as well, but I'd prefer if we avoided the use of `__ITER` for what should be built-in functionality, so I'm for handling that internally in `connect()` if possible. There's no reason to have scripts manually call `getProtoDefinitions()` unless we plan to expose them in JS for some reason.

joshcarp:
Hey, 
Yeah I'll be getting around to this in the next week or so

samples/grpc_reflection.js
na--: btw since it seems like the constructor doesn't have any parameters currently, while `connect()` does, and since it seems like the very rough consensus is that this should happen during `connect()`, should the option be a `connect()` parameter?

MStoykov: I am not very versed in gRPC, but from what I can see:
1. in order to have the reflection we need to connect to the server so it is only natural that it should happen with the `connect` method or after it
2.  if there is a way to detect that the server can give us the proto files I think k6 should just get the, And then have a way to tell it *not* to with an argument on the `connect` method. If it isn't possible I have no opinion on whether it should be an argument or a function call after the connect :shrug: 
3. We can definitely cache *across-VUs* as well as *per-vu* with the same/similar method used in SharedArray or the [xk6-segment extension](https://github.com/MStoykov/xk6-segment/blob/8df18d74cb838a7c16167afec6feabb65ae93af0/pkg/segment/segment.go#L39-L70) ... I do consider this trivial as well ;) . Although probably there should be a way to tell if you don't want to :shrug: maybe in a separate PR. This should remove the whole `__ITER==0` thing and we can probably use the server address as the key (from what I understand)

na--: cross-VU sharing of protobuf definitions is definitely something out of scope for this PR

js/modules/k6/grpc/client.go
```diff
@@ -114,6 +116,51 @@ func walkFileDescriptors(seen map[string]struct{}, fd *desc.FileDescriptor) []*d
 	return fds
 }
 
+// Reflect will use the grpc reflection api to make the file descriptors available to request.
+func (c *Client) Reflect(addr string, params map[string]interface{}) ([]MethodInfo, error) {
```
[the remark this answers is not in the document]
joshcarp: Moved to reflect.go

js/modules/k6/grpc/reflect.go
joshcarp: Now it's in the VU context and this has been removed

samples/grpc_reflection.js
joshcarp: I've changed it so that the `client.reflect` function is called at the end of the `connect` function only if `{reflect: true}` is passed in. 

I have used a `sync.Once` to ensure that it is only called once per client

imiric (commented):
Hi @joshcarp, thanks for picking this back up. The latest changes LGTM. We'd like to get this merged for the upcoming v0.34.0 release, slated for end of August.

A few points besides the minor comments:

- The performance is surprisingly poor when testing against a local `samples/grpc_server` with `samples/grpc_reflection.js`:

    ```
    grpc_req_duration....: avg=538.05ms min=75.52ms med=440ms    max=959.33ms p(90)=907.96ms p(95)=933.64ms
    iteration_duration...: avg=549.21ms min=76.89ms med=474.45ms max=960.83ms p(90)=914.34ms p(95)=937.59ms
    ```

  It's not much better with `samples/grpc.js`, even without `console.log()`-ing the message. So the server might be unoptimized, but I'm wondering if you know the reason for this.

  I suppose it's related to how we connect and disconnect on each iteration. Ideally we could keep the connection open across iterations, but that might be a much bigger change. And this is the way WebSockets work now, so it's probably acceptable as is.

- I see you registered the reflection on the `httpmultibin` server, but I don't se any tests for it. Do you think some could be added alongside the existing ones in `client_test.go`?

- There are some linter errors that should be fixed, and we'll look into the xk6 test failures, which are probably related to the recent k6io -> grafana org change.

joshcarp:
Hey
> I see you registered the reflection on the httpmultibin server, but I don't se any tests for it. Do you think some could be added alongside the existing ones in client_test.go?

Yeah, I started testing it although ran into some problems because the grpc test cases seem to all use the same state which means it's difficult to test both `load` and `reflect` 

> There are some linter errors that should be fixed, and we'll look into the xk6 test failures, which are probably related to the recent k6io -> grafana org change.

Apologies for that 

> The performance is surprisingly poor when testing against a local samples/grpc_server with samples/grpc_reflection.js:

investigating

js/modules/k6/grpc/reflect.go
```diff
@@ -0,0 +1,56 @@
+package grpc
```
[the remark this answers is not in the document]
joshcarp: reflect has been moved below connect

samples/grpc_server/main.go
```diff
@@ -138,7 +138,7 @@ func (s *routeGuideServer) RouteChat(stream RouteGuide_RouteChatServer) error {
 			return nil
 		}
 		if err != nil {
-			return err
+ 			return err
```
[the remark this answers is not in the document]
joshcarp: just gofumpt'ed it

joshcarp:
Hey @imiric 
I've added changes, one thing that was required was to refactor how the grpc tests work because the current method uses the same state between tests, so it was difficult to test both `load` and reflect in the same test harness. I've now put it in table tests that run in a completely fresh context for every test. I have tested:
- Reflection on an api that doesn't have reflection enabled: `ReflectUnregistered`
- Reflect on a server that does `Reflect`
- Reflect and invoke on a method that doesn't exist `ReflectInvokeNoExist`
- Reflect and invoke on a method that does exist `ReflectInvoke`

joshcarp:
Performance issues seem to be an issue in k6 already: https://github.com/grafana/k6/issues/1846
I haven't dug too deep into it, I'll try to do some benchmarks between reflect and loading directly from source files

js/modules/k6/grpc/client_test.go
```diff
@@ -69,16 +71,558 @@ func assertMetricEmitted(t *testing.T, metric *stats.Metric, sampleContainers []
...
+			val, err := rt.RunString(tb.Replacer.Replace(test.initString.code))
+			assertReponse(t, test.initString, err, val, tb, samples)
+			ctx = lib.WithState(ctx, state)
+			val, err = rt.RunString(tb.Replacer.Replace(test.vuString.code))
+			assertReponse(t, test.vuString, err, val, tb, samples)
+		})
+	}
+}
 
+func assertReponse(t *testing.T, r codeBlock, err error, val goja.Value, tb *httpmultibin.HTTPMultiBin, samples chan stats.SampleContainer) {
```
imiric, line 604:
```suggestion
func assertResponse(t *testing.T, r codeBlock, err error, val goja.Value, tb *httpmultibin.HTTPMultiBin, samples chan stats.SampleContainer) {
```

js/modules/k6/grpc/client_test.go
```diff
@@ -69,16 +71,558 @@ func assertMetricEmitted(t *testing.T, metric *stats.Metric, sampleContainers []
...
+			},
+		},
+		{
+			name: "LoadNotFound",
+			initString: codeBlock{
+				code: `
+			var client = new grpc.Client();
+			client.load([], "./does_not_exist.proto");`,
+				// (rogchap) this is a bit of a hack as windows reports a different system error than unix
+				err: "no such file or directory|The system cannot find the file specified",
```
imiric, line 107: I'd make this explicit with an `if runtime.GOOS == "windows" { ... } else { ... }` check instead. And we can avoid the regex ;)

js/modules/k6/grpc/client_test.go
```diff
@@ -69,16 +71,558 @@ func assertMetricEmitted(t *testing.T, metric *stats.Metric, sampleContainers []
 	assert.True(t, seenMetric, "url %s didn't emit %s", url, metric.Name)
 }
 
+// Run represents an execution of a k6 script.
+type codeBlock struct {
```
imiric, line 75: Outdated comment?

imiric (commented):
Thanks for the quick changes @rogchap!

Nice work on the tests too. I'd prefer smaller tests for individual methods (`TestClientLoad`, `TestClientConnect`, etc.), but this is great as well.

This LGTM for merging, but let's wait for someone else from the k6 team to also review it, possibly sometime next week.

> Performance issues seem to be an issue in k6 already: #1846

Yeah, I remember that issue, but the performance in this test is orders of magnitude worse. In any case, it's also that way on `master` with `samples/grpc.js`, so it's not related to these changes, but probably to the test server itself.

js/modules/k6/grpc/client.go
```diff
@@ -287,10 +299,65 @@ func (c *Client) Connect(ctxPtr *context.Context, addr string, params map[string
...
 	}
-
+	if reflect {
+		var err error
+		c.sync.Do(func() {
+			err = c.reflect(ctxPtr)
+		})
+		if err != nil {
+			return false, fmt.Errorf("error invoking reflect API: %w", err)
+		}
```
MStoykov, line 309: The `err` will only be populated on the first run which means that if you run multiple iterations and the first one fails reflect the API you will get 
```
ERRO[0000] error invoking reflect API: rpc error: code = Unimplemented desc = unknown service grpc.reflection.v1alpha.ServerReflection
        at go.k6.io/k6/js/common.Bind.func1 (native)
        at file:///home/mstoykov/work/k6io/k6/samples/grpc_reflection.js:7:64(9)  executor=shared-iterations scenario=default source=stacktrace
ERRO[0000] method "/main.RouteGuide/GetFeature" not found in file descriptors
        at go.k6.io/k6/js/common.Bind.func1 (native)
        at file:///home/mstoykov/work/k6io/k6/samples/grpc_reflection.js:9:13(21)  executor=shared-iterations scenario=default source=stacktrace
ERRO[0000] method "/main.RouteGuide/GetFeature" not found in file descriptors
        at go.k6.io/k6/js/common.Bind.func1 (native)
        at file:///home/mstoykov/work/k6io/k6/samples/grpc_reflection.js:9:13(21)  executor=shared-iterations scenario=default source=stacktrace
ERRO[0000] method "/main.RouteGuide/GetFeature" not found in file descriptors
        at go.k6.io/k6/js/common.Bind.func1 (native)
        at file:///home/mstoykov/work/k6io/k6/samples/grpc_reflection.js:9:13(21)  executor=shared-iterations scenario=default source=stacktrace
ERRO[0000] method "/main.RouteGuide/GetFeature" not found in file descriptors
        at go.k6.io/k6/js/common.Bind.func1 (native)
        at file:///home/mstoykov/work/k6io/k6/samples/grpc_reflection.js:9:13(21)  executor=shared-iterations scenario=default source=stacktrace
```

as you can see the much more prevalent message is that k6 couldn't find file descriptors.

This also means that if the connection fails for any reason on the first request you practically bricked this connection.

Additional problems:
- if the server reloads to change the definition and the user decided to use reflection so that ... works, it won't
- The connection can technically connect to multiple servers over many iterations, but this will prevent it to load different definitions for them.

I currently am of the opinion that it might be better off to *not* cache the definition and keep reloading them on each iteration just so that we don't tie our hands now. On the other hand, IMO this whole grpc API will be thrown out at some point when we start implementing everything it should support, so ... might as well have caching now :shrug: .

From my experiments. 100 VUs with caching it does 2.3k without caching it does a bit above 1k iter/s and with the `__ITER == 0` hack around 10k.

We can obviously keep adding flags, but again IMO the whole grpc module will need a rewrite and redoing of the API at some point so I am not certain how much adding flags for corner cases as the mentioned above is ... reasonable.

MStoykov (changes requested):
Thanks for the PR, great job.

Please do not delete license files and rebase on top of `master` which will also fix the currently failing CI for you. (you can also squash everything, although we will do that either way in the end :shrug: )

As I mentioned in one of the comments I am not certain caching (or more accurately not re-reflecting on each connection) is a good idea, especially given all the things that this break currently. On the other hand the API currently isn't all that useful and as I mention I doubt we won't redo it as a whole at some point in the future so ... maybe have caching for now ... and hope people won't have problems with the fact it does break some workflows that reflection enables :shrug: 

@imiric the abysmal performance you are noticing is because the current grpc test server has a bunch of sleeps in it to be more "realistic" which probably should be dropped. My numbers below are from *after* I've dropped them.

js/modules/k6/grpc/client.go
joshcarp: >if the server reloads to change the definition and the user decided to use reflection so that ... works, it won't

is this a problem? grpc reflection isn't really a dynamic thing that is expected to change definitions ever, at least in all the time that I've used it

joshcarp:
Hey, 
I think this PR was a good exercise on how to implement grpc server reflection but considering that there are massive performance issues with it It might be best to not merge this until the grpc module has been refactored to support such a feature.

I think that merging it might give people the (possibly) incorrect idea that the grpc functionality is reliable and ready to use, which it is not. 

If I get time I'll run more benchmarks against a non-grpc server and possibly using a non k6 tool to triage grpc issues.
Stuff that needs investigation:
- https://github.com/grafana/k6/issues/1846 
    - is there an issue with the k6 implementation of grpc in k6? or is grpc slower in this case. If so;
    - is there an issue with the example grpc server?
    - what is the impact of grpc server reflection on this baseline performance?

na--:
> what is the impact of grpc server reflection on this baseline performance?

Hmm it shouldn't be a big problem if the initial `connect()` method is slower when reflection is enabled, that is somewhat expected... :confused: So I don't think it's a blocker for this PR.

Also, in general, we shouldn't compare k6 gRPC performance to k6 HTTP performance, that would be comparing apples to oranges. If we're worried about k6's gRPC implementation being slow for some reason, we should compare it to other gRPC implementations. But, again, that shouldn't be a big concern here, since the performance of the `connect()` method is generally not critical, and since we expect it to be slower with reflection enabled.

PR #1988, opened 2021-04-26
Do actually unlock SharedArray if write raced

MStoykov:
previous to this the code (which I have no idea why I added the
function, it probably was moved into a separate function later) would
*not* unlock the lock if 2 writes raced on Write lock after as the
second writer would see that it was already written but not enter the
if to defer the unlock.

### DIFF
--- a/js/modules/k6/data/data.go
+++ b/js/modules/k6/data/data.go
@@ -46,13 +46,11 @@ func (s *sharedArrays) get(rt *goja.Runtime, name string, call goja.Callable) sh
 	s.mu.RUnlock()
 	if !ok {
 		s.mu.Lock()
+		defer s.mu.Unlock()
 		array, ok = s.data[name]
 		if !ok {
-			func() { // this is done for the defer below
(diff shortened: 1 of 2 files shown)


MStoykov:
This is kind of hard to hit as in general there will be 1 writer in the initial VU and then the VUs will just read it ... but if you write code as in
```javascript
if (__VU == 0) { // workaround to collect all files for the cloud execution
    for(let i=0; i<dataFiles.length; i++){
        open(dataFiles[i]);
    }
} else {
    csvData = new SharedArray("data", function(){return papaparse.parse(open(dataFiles[__VU%dataFiles.length]), { header: true }).data});
}
```

This can race ... even then on my 8 core laptop it happened only once out of *many* runs, but one look at the code, and I was :facepalm: -ing ....

na-- (approved):
:sweat_smile: LGTM

imiric (approved):
LGTM

I found this strange when I first saw it, but also missed that `Unlock` wouldn't be called. :-/

PR #1990, opened 2021-04-27
Set cloud status to aborted by script on setup/teardown script errors

MStoykov:
Previous to this it would've been aborted by system


na-- (approved):

imiric (approved):
A test would be good to have, but not a big deal.

PR #1995, opened 2021-04-28
Tests refactorings, mostly `js` parallelization

MStoykov:


lib/testutils/httpmultibin/httpmultibin.go
```diff
@@ -368,12 +367,14 @@ func NewHTTPMultiBin(t testing.TB) *HTTPMultiBin {
 		Dialer:          dialer,
 		HTTPTransport:   transport,
 		Context:         ctx,
-		Cleanup: func() {
```
na--, line 371: I think you should leave this as it was, and make the cleanup like this:
```go
t.Cleanup(func() {
    result.Cleanup()
}()
```

This will allow tests to modify the cleanup order, if they need to, by overwriting the `tb.Cleanup` lambda.

na-- (approved):
LGTM except the minor nitpick I mentioned inline

lib/testutils/httpmultibin/httpmultibin.go
MStoykov: Given that this hasn't been required so far I prefer to be added when it's needed

js/bundle_test.go
```diff
@@ -729,32 +768,51 @@ func TestOpen(t *testing.T) {
...
-	}
-	logger := testutils.NewLogger(t)
-
-	bi, err := b.Instantiate(logger, 0)
-	if !assert.NoError(t, err) {
-		return
-	}
+		if !assert.NoError(t, err) {
+			return
+		}
```
na--, line 784: This pattern also repeats a lot below
```suggestion
		require.NoError(t, err)
```

js/bundle_test.go
```diff
@@ -763,6 +821,24 @@ func TestBundleInstantiate(t *testing.T) {
...
 	t.Run("Options", func(t *testing.T) {
+		t.Parallel()
+		b, err := getSimpleBundle(t, "/script.js", `
+		export let options = {
+			vus: 5,
+			teardownTimeout: '1s',
+		};
+		let val = true;
+		export default function() { return val; }
+	`)
```
na--, line 832:
```suggestion
			export let options = {
				vus: 5,
				teardownTimeout: '1s',
			};
			let val = true;
			export default function() { return val; }
		`)
```

na-- (dismissed):
LGTM besides the minor nitpicks I mentioned inline. We can fix this pattern in a separate PR:
```
if !assert.NoError(t, err) {
	return
}
```
since I see it occurs quite often... :disappointed:

js/bundle_test.go
```diff
@@ -669,14 +706,16 @@ func TestOpen(t *testing.T) {
...
-		fs = afero.NewReadOnlyFs(fs)
 		t.Run(name, func(t *testing.T) {
+			t.Parallel()
 			for _, tCase := range testCases {
 				tCase := tCase
 
 				testFunc := func(t *testing.T) {
+					t.Parallel()
+					fs, prefix, cleanUp := fsInit()
+					defer cleanUp()
```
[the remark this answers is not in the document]
MStoykov: I might try to do one more pass in a later PR to move all `defer cleanup` kind of things in tests to just use `t.Cleanup`

na-- (approved):

imiric (approved):

PR #2000, opened 2021-04-29
[WIP] ARM64 tests and builds, drop 32-bit builds

na--:
Closes https://github.com/k6io/k6/issues/1979, closes https://github.com/k6io/k6/issues/1999, closes https://github.com/k6io/k6/issues/842

There are still some flaky tests that need fixing though :disappointed:

### DIFF
--- a/.github/workflows/arm-tests.yml
+++ b/.github/workflows/arm-tests.yml
@@ -1,0 +1,26 @@
+name: ARM tests
+on:
+  push:
+    branches:
+      - master
+      - ci-changes # TODO: remove
+
+jobs:
+  test:
+    runs-on: [self-hosted, Linux, ARM64]
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: Install Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.x
+      - name: Run tests
+        run: |
+          set -x
+          go version
+          go env
+          export GOMAXPROCS=2
+          export CGO_ENABLED=1
+          go test -p 2 -race -timeout 800s ./...
+
--- a/build-release.sh
+++ b/build-release.sh
@@ -68,16 +68,16 @@ echo "--- Building Release: ${VERSION}"
 
 mkdir -p "$OUT_DIR"
 
-build mac     ""   GOOS=darwin  GOARCH=amd64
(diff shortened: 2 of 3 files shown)


MStoykov:
I propose we just drop the 32bit ones for this release and add arm ones for the next one ;)

na--:
Hmm given that the vast, vast majority of tests are fine, why not just add a `t.Skip()` for ARM architectures on the failing one, add an issue to fix it, and release this? We can't keep postponing this, especially when the tests on ARM seem less problematic than on Windows :sweat_smile:

MStoykov:
well, then do that ;)

.github/workflows/arm-tests.yml
```diff
@@ -0,0 +1,26 @@
+name: ARM tests
+on:
+  push:
+    branches:
+      - master
+      - ci-changes # TODO: remove
+
+jobs:
+  test:
+    runs-on: [self-hosted, Linux, ARM64]
```
imiric, line 10: I'm not finding a clear answer to this in the GHA docs, but where is `self-hosted` configured? Or does our ARM runner poll recent GHA builds and kicks off when it encounters a `self-hosted` one?

build-release.sh
```diff
@@ -68,16 +68,16 @@ echo "--- Building Release: ${VERSION}"
...
-
-package linux32 tgz
-package linux64 tgz
-package linux64 rpm
-package linux64 deb
-package mac     zip
-package win32   zip
-package win64   zip
+build osx-amd64     ""   GOOS=darwin  GOARCH=amd64
+build osx-arm64     ""   GOOS=darwin  GOARCH=arm64
```
imiric, line 72: There's no such thing as "OS X" anymore, so this should probably be:

```suggestion
build macos-amd64     ""   GOOS=darwin  GOARCH=amd64
build macos-arm64     ""   GOOS=darwin  GOARCH=arm64
```

And same for `package` below.

imiric (commented):
Mostly LGTM save for the `s/osx/macos/` comment.

Also, I think rebasing on `master` after https://github.com/k6io/k6/pull/2014 is merged should fix those xk6 errors.

MStoykov (approved):

imiric (approved):
:shipit:

PR #2002, opened 2021-05-01
Update go version at go.mod

inosato:
According to https://github.com/k6io/k6/pull/1922, k6 seems to support go v1.15 and v1.16 now.
But go.mod is still defined as go 1.14.
So I change it to 1.15.
If I change it to 1.16 , it may not work with v1.15.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/loadimpact/k6
 
-go 1.14
+go 1.15
 
 require (
 	github.com/Azure/go-ntlmssp v0.0.0-20200615164410-66371956d46c


imiric:
Hi, thanks for proposing this.

What is the reason to update this? AFAIK it specifies the minimum Go version supported and we don't have a reason to bump this yet unless we want to stop supporting building with 1.14.

I'll refrain from merging this since most of the team is currently on vacation, but we'll get back to you sometime next week.

inosato:
@imiric 
Hi, thank you for your response.

> What is the reason to update this? 

I'm not finding a problem with using go 1.14.
But probably 1.14 is no longer supported.

* Release policy of go ( https://golang.org/doc/devel/release.html#policy )
  > Each major Go release is supported until there are two newer major releases. 


The newest version is 1.16, so `two newer major releases` are 1.16 and 1.15.
I think bump 1.15 is better.


> we'll get back to you sometime next week.

OK, it is not urgent.

na--:
In k6 we usually adopt a similar strategy to Go in trying to support the last two st able Go versions by testing with them:
https://github.com/k6io/k6/blob/af1e032ebf949cde092d4a2eddb78528ea0a5864/.github/workflows/all.yml#L60-L63
https://github.com/k6io/k6/blob/af1e032ebf949cde092d4a2eddb78528ea0a5864/.github/workflows/all.yml#L87-L90

That said, unless we have a specific reason for bumping the minimum Go version in `go.mod`, I don't think we should do it. It will just needlessly break the builds of people with older Go versions when they might have a reason to build k6 with them (e.g. building for an architecture that was dropped in newer versions), despite k6 still being able to work.

https://github.com/k6io/k6/pull/1975 is an example of a specific reason we're probably going to require at least Go 1.16 soon, starting with k6 v0.33.0 - we need Go 1.16 for `//go:embed`. Thanks for this PR, but I'll close it, given that this PR I linked already bumps it to 1.16 for a specific reason.

PR #2005, opened 2021-05-04
Fix log message in statsd output

vishalkuo:
This addresses https://github.com/k6io/k6/issues/2004 

Per the issue, I'm happy to discuss and implement longer term solutions for log flags but I figured this is a fast and non-controversial solution.


na-- (approved):
Thanks!

MStoykov (approved):

PR #1886, opened 2021-03-07
output/json: Set and flush thresholds

codebien:
Added the ability to set and flush the thresholds implementing the WithThresholds interface as required from the Engine.

The SetThresholds implementation is the same as the cloud version. If you like the idea, I could also refactor out it to centralize the logic.

Closes #1052

PR #2007, opened 2021-05-06
Fix threshold encoding when marshalling to JSON

imiric:
This fixes an encoding issue for thresholds that are marshalled to JSON, recently added to the JSON output in #1886.

[`json.Marshal()`](https://golang.org/pkg/encoding/json/#Marshal) escapes HTML-sensitive characters by default, so `<` is converted to `\u003c`, `>` to `\u003e`, etc.

We missed it in our tests since they also relied on this behavior which obscures the wrong encoding, but you can see it if you checkout af1e032e and run `go run main.go run -o json=- ./samples/thresholds.js`. The `http_req_duration` metric will be rendered as:

```
{"type":"Metric","data":{"name":"http_req_duration","type":"trend","contains":"time","tainted":null,"thresholds":["p(95)\u003c500"],"submetrics":null,"sub":{"name":"","parent":"","suffix":"","tags":null}},"metric":"http_req_duration"}
```

Whereas on this branch it's rendered properly:

```
{"type":"Metric","data":{"name":"http_req_duration","type":"trend","contains":"time","tainted":null,"thresholds":["p(95)<500"],"submetrics":null,"sub":{"name":"","parent":"","suffix":"","tags":null}},"metric":"http_req_duration"}
```

We should probably merge this before releasing v0.32.0.

### DIFF
--- a/output/json/json_test.go
+++ b/output/json/json_test.go
@@ -47,7 +47,7 @@ func getValidator(t *testing.T, expected []string) func(io.Reader) {
 				t.Errorf("Read unexpected line number %d, expected only %d entries", i, len(expected))
 				continue
 			}
-			assert.JSONEq(t, expected[i-1], string(s.Bytes()))
+			assert.Equal(t, expected[i-1], string(s.Bytes()))
 		}
 		assert.NoError(t, s.Err())
 		assert.Equal(t, len(expected), i)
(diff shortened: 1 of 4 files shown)


stats/thresholds.go
```diff
@@ -213,7 +216,21 @@ func (ts Thresholds) MarshalJSON() ([]byte, error) {
...
 		configs[i].AbortGracePeriod = t.AbortGracePeriod
 	}
-	return json.Marshal(configs)
+
+	return MarshalJSONWithoutHTMLEscape(configs)
+}
+
+// MarshalJSONWithoutHTMLEscape marshals t to JSON without escaping characters
+// for safe use in HTML.
+func MarshalJSONWithoutHTMLEscape(t interface{}) ([]byte, error) {
```
imiric, line 225: I would've liked to move this to a helpers file in `lib`, and also reuse it here: https://github.com/k6io/k6/blob/af1e032ebf949cde092d4a2eddb78528ea0a5864/lib/archive.go#L384-L395

But unfortunately that creates an import cycle (`stats`->`lib`->`stats`), so I left it here.

stats/thresholds.go
```diff
@@ -213,7 +216,21 @@ func (ts Thresholds) MarshalJSON() ([]byte, error) {
...
+	buffer := &bytes.Buffer{}
+	encoder := json.NewEncoder(buffer)
+	encoder.SetEscapeHTML(false)
+	err := encoder.Encode(t)
+	bytes := buffer.Bytes()
+	// Remove the newline appended by Encode() :-/
+	// See https://github.com/golang/go/issues/37083
+	return bytes[:len(bytes)-1], err
```
imiric, line 233: This is unfortunate, but I don't see a way around it. :disappointed:

output/json/json_test.go
```diff
@@ -47,7 +47,7 @@ func getValidator(t *testing.T, expected []string) func(io.Reader) {
 				t.Errorf("Read unexpected line number %d, expected only %d entries", i, len(expected))
 				continue
 			}
-			assert.JSONEq(t, expected[i-1], string(s.Bytes()))
+			assert.Equal(t, expected[i-1], string(s.Bytes()))
```
imiric, line 50: `assert.JSONEq()` uses `json.Unmarshal()` and was hiding the encoding issue. So an equality check seems safer.

na--: Hmm but `Equal` may flakily fail if the order of encoding between the two differs, which it might when there are Go maps (e.g. `tags`) involved.

MStoykov: go maps are sorted before they are encoded with the stdlib json. This is actually one of the "*optimizations*" in easyjson ;)

imiric: Ah, that explains it then. I added another tag and [stress tested](https://dave.cheney.net/2013/06/19/stress-test-your-go-packages) `output/json` for a few minutes and didn't see a failure.

Besides, even if that wasn't the case we'd need to stabilize them for the tests, since `assert.JSONEq` is unusable because of the encoding issue.

MStoykov (approved):

PR #2015, opened 2021-05-11
Update brotli to v1.0.2

MStoykov:
mostly because of https://github.com/andybalholm/brotli/issues/22


na-- (approved):
LGTM, though please move the second half of the title in the description field :wink: :smile:

PR #2017, opened 2021-05-12
Make the windows version info only for windows

MStoykov:
Previously this would've been embedded in *all* binaries as go didn't
know that it shouldn't. This looks to have had no impact, but with the
addition of macos + arm64 binaries we found it as that platform doesn't
support embedding of the specific generate file.


na-- (approved):

imiric (approved):
Clean solution! I was expecting to have to do some wrangling inside `build-release.sh` :sweat_smile:

PR #2021, opened 2021-05-14
Fix a mistake in the v0.32.0 release notes

na--:
Not sure how we missed this :sweat_smile: :man_facepalming:

### DIFF
--- a/release notes/v0.32.0.md
+++ b/release notes/v0.32.0.md
@@ -84,7 +84,7 @@ We will now publish binary arm64 releases for macOS and Linux for new k6 release
 
 ## Internals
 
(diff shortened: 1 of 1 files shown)


na--:
We should just get rid of Codecov, with PRs like this it's obvious how completely wrong it is :confused:

MStoykov (approved):

MStoykov:
> We should just get rid of Codecov, with PRs like this it's obvious how completely wrong it is confused

I can't currently open to see which 1 line is now not being covered, but it is very possible that between some runs we don't run all lines. I am still fine with dropping codecov, but if we are dropping it I prefer to replace it in not this release cycle as 1 release  cycle of mostly CI changes in enough for me :P

na--:
Hmm seems codecodev has updated the comment since, it used to be way more wrong. Click on the `edited` field in the top, see the initial version and how completely wrong it was.

MStoykov:
It seems like it updated with the `windows` specific tests as well ... at least from that edit

Issue #2024, opened 2021-05-14
Websockets should tag connection errors better


MStoykov:
Currently, if an error (like timeout) happens during the creation of the websocket k6 will not emit `status=0` or tags `error` and `error_code` as will the HTTP module, but will just no emit the `status` tag, call the `error` handle if defined and end continue executing javascript.

This will require for the `error_code` code to be made exported and possibly moved to a different module, which will impact #2008, so we might want to merge that first.

neha13rb1998:
Can you please explain how to capture websocket requests errors in k6?

na--:
@neha13rb1998, you can define an `error` handler, see https://k6.io/docs/javascript-api/k6-ws/socket/socket-on-event-callback

neha13rb1998:
@na-- This won't store the errors in database. I want to capture websocket errors and store them in a database. This data will be used to visualize on a dashboard. How can I achieve this?

na--:
@neha13rb1998, you can define a new custom metric (e.g. a `Counter`) and `add()` to it in the `error` websocket handler, see:
- https://k6.io/docs/using-k6/metrics/#custom-metrics
- https://k6.io/docs/javascript-api/k6-metrics/counter/

neha13rb1998:
@na-- I defined the custom metric but metric is not generated. 
Where am I going wrong?
Below is my code:
import ws from 'k6/ws';
import { check } from 'k6';
import { Rate } from 'k6/metrics';
import { Counter } from 'k6/metrics';

// A custom metric to track failure rates
var failureRate = new Rate("check_failure_rate");
var errorCounter = new Counter("error_rate");

// Options
export let options = {

    //stages for spike test
    stages: [
  
      { duration: '30s', target: 1000 }, // below normal load
      { duration: '1m', target: 1000 },
      { duration: '30s', target: 5000 }, // spike to 5000 users
      { duration: '3m', target: 5000 }, // stay at 5000 for 3 minutes
      { duration: '30s', target: 1000 }, // scale down. Recovery stage.
      { duration: '3m', target: 1000 },
      { duration: '30s', target: 0 },
  
    ],
    
      noVUConnectionReuse: false, //specifying whether k6 should reuse TCP connections between iterations of a VU
      rps: 50, //The maximum number of requests to make per second, in total across all VUs
[...]

PR #2025, opened 2021-05-15
Fix error codes for http2, x509 and tls errors 

MStoykov:
It seems http2, x509 and tls errors are not pointer errors, after all.

I mostly found this through the fact that even after the last patches it
still didn't match them, but now there are finally tests for all of
them, so hopefully this will be a permanent fix.

### DIFF
--- a/lib/netext/httpext/error_codes.go
+++ b/lib/netext/httpext/error_codes.go
@@ -57,6 +57,7 @@ const (
 	tcpBrokenPipeErrorCode   errCode = 1201
 	netUnknownErrnoErrorCode errCode = 1202
 	tcpDialErrorCode         errCode = 1210
+	tcpDialTimeoutErrorCode  errCode = 1211
 	tcpDialRefusedErrorCode  errCode = 1212
 	tcpDialUnknownErrnoCode  errCode = 1213
 	tcpResetByPeerErrorCode  errCode = 1220
@@ -86,6 +87,7 @@ const (
 
 const (
 	tcpResetByPeerErrorCodeMsg  = "%s: connection reset by peer"
+	tcpDialTimeoutErrorCodeMsg  = "dial: i/o timeout"
 	tcpDialRefusedErrorCodeMsg  = "dial: connection refused"
 	tcpBrokenPipeErrorCodeMsg   = "%s: broken pipe"
 	netUnknownErrnoErrorCodeMsg = "%s: unknown errno `%d` on %s with message `%s`"
--- a/lib/netext/httpext/error_codes.go
+++ b/lib/netext/httpext/error_codes.go
@@ -187,22 +187,22 @@ func errorCodeForError(err error) (errCode, string) {
 		return blackListedIPErrorCode, blackListedIPErrorCodeMsg
 	case netext.BlockedHostError:
 		return blockedHostnameErrorCode, blockedHostnameErrorMsg
-	case *http2.GoAwayError:
+	case http2.GoAwayError:
 		return unknownHTTP2GoAwayErrorCode + http2ErrCodeOffset(e.ErrCode),
 			fmt.Sprintf(http2GoAwayErrorCodeMsg, e.ErrCode)
-	case *http2.StreamError:
+	case http2.StreamError:
 		return unknownHTTP2StreamErrorCode + http2ErrCodeOffset(e.Code),
 			fmt.Sprintf(http2StreamErrorCodeMsg, e.Code)
-	case *http2.ConnectionError:
-		return unknownHTTP2ConnectionErrorCode + http2ErrCodeOffset(http2.ErrCode(*e)),
-			fmt.Sprintf(http2ConnectionErrorCodeMsg, http2.ErrCode(*e))
+	case http2.ConnectionError:
+		return unknownHTTP2ConnectionErrorCode + http2ErrCodeOffset(http2.ErrCode(e)),
+			fmt.Sprintf(http2ConnectionErrorCodeMsg, http2.ErrCode(e))
 	case *net.OpError:
 		return errorCodeForNetOpError(e)
-	case *x509.UnknownAuthorityError:
+	case x509.UnknownAuthorityError:
 		return x509UnknownAuthorityErrorCode, x509UnknownAuthority
-	case *x509.HostnameError:
+	case x509.HostnameError:
 		return x509HostnameErrorCode, x509HostnameErrorCodeMsg
-	case *tls.RecordHeaderError:
+	case tls.RecordHeaderError:
 		return defaultTLSErrorCode, err.Error()
 	case *url.Error:
 		return errorCodeForError(e.Err)
--- a/lib/netext/httpext/error_codes.go
+++ b/lib/netext/httpext/error_codes.go
@@ -62,7 +62,8 @@ const (
 	tcpDialUnknownErrnoCode  errCode = 1213
 	tcpResetByPeerErrorCode  errCode = 1220
 	// TLS errors
-	defaultTLSErrorCode           errCode = 1300
+	defaultTLSErrorCode           errCode = 1300 //nolint:deadcode,varcheck // this is here to save the number
+	tlsHeaderErrorCode            errCode = 1301
 	x509UnknownAuthorityErrorCode errCode = 1310
 	x509HostnameErrorCode         errCode = 1311
 
@@ -205,7 +206,7 @@ func errorCodeForError(err error) (errCode, string) {
 	case x509.HostnameError:
 		return x509HostnameErrorCode, x509HostnameErrorCodeMsg
 	case tls.RecordHeaderError:
-		return defaultTLSErrorCode, err.Error()
+		return tlsHeaderErrorCode, err.Error()
 	case *url.Error:
 		return errorCodeForError(e.Err)
 	default:
--- a/lib/netext/httpext/error_codes_test.go
+++ b/lib/netext/httpext/error_codes_test.go
@@ -21,15 +21,18 @@
 package httpext
 
 import (
+	"context"
 	"crypto/tls"
-	"crypto/x509"
 	"errors"
 	"fmt"
+	"io/ioutil"
 	"net"
 	"net/http"
+	"net/http/httptest"
 	"net/url"
 	"os"
 	"runtime"
+	"strings"
 	"syscall"
 	"testing"
 	"time"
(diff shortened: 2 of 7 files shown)


na-- (dismissed):

lib/netext/httpext/error_codes_test.go
```diff
@@ -218,3 +219,76 @@ func TestDnsResolve(t *testing.T) {
...
+
+	res, err := client.Get(tb.Replacer.Replace("HTTP2BIN_URL/tsr")) //nolint:noctx
+	require.NotNil(t, res)
+	require.NoError(t, err)
+	_, err = ioutil.ReadAll(res.Body)
+	_ = res.Body.Close()
+	require.Error(t, err)
+
+	code, msg := errorCodeForError(err)
+	assert.Equal(t, unknownHTTP2StreamErrorCode+errCode(http2.ErrCodeInternal)+1, code)
```
imiric, line 248: This summing of error codes seems very brittle, and I just saw we do the same in `http2ErrorCodeOffset()`. :-/ Why is this done instead of having a specific error code we can assert on?

lib/testutils/httpmultibin/httpmultibin.go
```diff
@@ -165,10 +164,10 @@ func getEncodedHandler(t testing.TB, compressionType httpext.CompressionType) ht
 		)
 
 		switch compressionType {
-		case httpext.CompressionTypeBr:
+		case "br":
 			encw = brotli.NewWriter(rw)
 			encoding = "br"
-		case httpext.CompressionTypeZstd:
+		case "zstd":
```
imiric, line 170: Is this part of the http2 related changes? If not I'd prefer seeing it in a separate commit.

Also, you could just drop the `encoding` var and pass `compressionType` as is (or rename it to `encoding`).

imiric (commented):
I'm concerned that we're replicating a scenario in tests that matches our expectations but doesn't reproduce what happens in production. Could we add integration tests for this to ensure it's fixed?

lib/testutils/httpmultibin/httpmultibin.go
MStoykov: this was creating an import cycle so I needed to drop it... I even think it would just be part of the `http` test that does this, but wasn't into trying to tackle the spaghetti that is the `http` module tests

lib/netext/httpext/error_codes_test.go
MStoykov: there are a bunch of error codes defined in the http2 specification and some of them can be returned for stream, some for connection errors so it was easier to just have 20 codes between those and make the codes on the spot instead of defining like 14 error codes per http2*Error

MStoykov:
> Could we add integration tests for this to ensure it's fixed? 

By integration tests do you mean using the js `http` module? We could, but IMO it makes very little sense. What I've added here are 3(1 http2+ 2 x509) tests for the things I could reproduce and moving it to the `http` module will not help with that. I would argue these *are* integration tests even if not between the `http` module and the error codes code, and the problem with adding more is that I will need to find a way to reproduce this error, which has always been the problem. 

We can move them (or add additional) tests in `http` module but we will probably need to make the `error_codes` exported and in another package which is already kind of planned for #2024, but I would argue the more important part is having ways to reproduce the errors

lib/netext/httpext/error_codes_test.go
imiric: OK, I guess it's fine since they're constants, but it's very confusing to read. Maybe some comments would help.

lib/netext/httpext/error_codes.go
```diff
@@ -191,22 +191,22 @@ func errorCodeForError(err error) (errCode, string) {
...
 		return errorCodeForNetOpError(e)
-	case *x509.UnknownAuthorityError:
+	case x509.UnknownAuthorityError:
 		return x509UnknownAuthorityErrorCode, x509UnknownAuthority
-	case *x509.HostnameError:
+	case x509.HostnameError:
 		return x509HostnameErrorCode, x509HostnameErrorCodeMsg
-	case *tls.RecordHeaderError:
+	case tls.RecordHeaderError:
 		return defaultTLSErrorCode, err.Error()
```
MStoykov, line 210: I do think we should probably add TLSHeaderErrorCode ...

lib/testutils/httpmultibin/httpmultibin.go
```diff
@@ -253,15 +252,29 @@ func (*GRPCStub) HalfDuplexCall(grpctest.TestService_HalfDuplexCallServer) error
 }
 
 // NewHTTPMultiBin returns a fully configured and running HTTPMultiBin
-func NewHTTPMultiBin(t testing.TB) *HTTPMultiBin {
+//nolint:funlen
+func NewHTTPMultiBin(t testing.TB, opts ...MultiBinOptWrap) *HTTPMultiBin {
+	// TODO instead of having opts maybe make the whole mutlibin more composable, as there is no point for most of the
+	// servers inthe majority of the tests
```
imiric, line 258:
```suggestion
	// TODO instead of having opts maybe make the whole multibin more composable, as there is no point for most of the
	// servers in the majority of the tests
```

lib/testutils/httpmultibin/httpmultibin.go
```diff
@@ -375,3 +391,19 @@ func NewHTTPMultiBin(t testing.TB) *HTTPMultiBin {
...
 	return result
 }
+
+// MultiBinOptWrap is a way to configure a new multibin
+type MultiBinOptWrap struct {
+	Key   MultiBinOpt
+	Value interface{}
+}
+
+// MultiBinOpt is an option key enum typefor mutlbin configuration
```
imiric, line 401:
```suggestion
// MultiBinOpt is an option key enum typefor multibin configuration
```

lib/testutils/httpmultibin/httpmultibin.go
```diff
@@ -253,15 +252,29 @@ func (*GRPCStub) HalfDuplexCall(grpctest.TestService_HalfDuplexCallServer) error
...
-func NewHTTPMultiBin(t testing.TB) *HTTPMultiBin {
+//nolint:funlen
+func NewHTTPMultiBin(t testing.TB, opts ...MultiBinOptWrap) *HTTPMultiBin {
+	// TODO instead of having opts maybe make the whole mutlibin more composable, as there is no point for most of the
+	// servers inthe majority of the tests
+	var customConnContext func(ctx context.Context, c net.Conn) context.Context
+	for _, opt := range opts {
+		switch opt.Key {
+		case NullOpt:
+			t.Fatalf("unsupported multibin option %d, the NullOpt should not be used", opt.Key)
```
imiric, line 263: What's the purpose of defining/checking for it then?

imiric (dismissed):
LGTM, nice work on the tests! :clap:

lib/testutils/httpmultibin/httpmultibin.go
MStoykov: Literally a linter told me I don't check all of them and it is nice to have a more specific message for why it failed. On the other hand, maybe I should add that this will happen also if `Key` wasn't set as it will default to 0 :thinking:

imiric: OK, but I mean why even define `NullOpt` at all then? It's not used anywhere, and you could start `ConnContextFuncOpt` at `iota + 1` if you don't want to use `0`. Though it's a minor thing, I don't mind it much, just wondering.

imiric (approved):

lib/testutils/httpmultibin/httpmultibin.go
MStoykov: hm .... I guess given that this is not user-facing code it might be easier if I just drop it :thinking: in general I prefer to have it for user-facing enums so we can see when it isn't set and provide an error.

lib/netext/httpext/error_codes.go
na--: Given that `defaultTLSErrorCode` doesn't seem to be used for anything else, I am not sure it's necessary :man_shrugging:

lib/netext/httpext/request_test.go
```diff
@@ -357,3 +359,116 @@ func TestTrailFailed(t *testing.T) {
...
+
+	res, err := MakeRequest(ctx, preq)
+	require.NoError(t, err)
+	assert.NotNil(t, res)
+	assert.Len(t, samples, 1)
+	sampleCont := <-samples
+	allSamples := sampleCont.GetSamples()
+	require.Len(t, allSamples, 9)
+	expTags := map[string]string{
+		"error":             "Request timeout",
```
na--, line 463: rebase on the latest `master` and fix the capitalization? otherwise it's going to fail when you merge the PR

na-- (changes requested):
I don't think the whole `MultiBinOptWrap` is necessary. If you need custom low-level options in a certain test, simply spin up a custom `httptest.NewUnstartedServer()`, it's only a few lines of code. If you need it in multiple tests, write another small helper around it, instead of making `HTTPMultiBin` into more of a monster than it already is... We risk it becoming sentient and eating us soon :sweat_smile:

lib/netext/httpext/error_codes.go
MStoykov: well this is also the only tls error I am matching ... as it is the only one I get, but arguably it isn't the only one that can be so it taking the "default" one seems like a bad idea ... ;)

MStoykov:
> I don't think the whole MultiBinOptWrap is necessary. If you need custom low-level options in a certain test, simply spin up a custom httptest.NewUnstartedServer(), it's only a few lines of code. If you need it in multiple tests, write another small helper around it, instead of making HTTPMultiBin into more of a monster than it already is... We risk it becoming sentient and eating us soon sweat_smile

I am fine by this ... but in that case I don't think we have more than a few tests that actually use more then what `httptest.NewUnstartedServer` can provide in "few lines"(in my case the lines were close to 20 before the lines in the actual tests as I need tls configuration).

na-- (approved):

MStoykov:
```
        	Error Trace:	error_codes_test.go:306
        	Error:      	Not equal: 
        	            	expected: "read tcp 127.0.0.1:51368->127.0.0.1:51367: wsarecv: An existing connection was forcibly closed by the remote host."
        	            	actual  : "read: connection reset by peer"
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1 +1 @@
        	            	-read tcp 127.0.0.1:51368->127.0.0.1:51367: wsarecv: An existing connection was forcibly closed by the remote host.
        	            	+read: connection reset by peer
        	Test:       	TestDefaultTLSError
```
        	
this happens ever 1 out of 3-4 runs :( ... but I don't know with what I could've broken it

imiric (dismissed):
LGTM, nice job fixing that test :tada:

na-- (dismissed):

na-- (approved):

imiric (approved):

PR #2026, opened 2021-05-18
Dependency updates

MStoykov:


modtools_frozen.yml
```diff
@@ -0,0 +1,60 @@
...
+  minVersion: v0.0.0-20201204225414-ed752295db88
+  validUntil: 2029-08-04T17:00:51+03:00
+- path: google.golang.org/genproto
+  minVersion: v0.0.0-20200903010400-9bfcb5116336
+  validUntil: 2029-08-04T17:01:00+03:00
+- path: google.golang.org/protobuf
+  minVersion: v1.25.1-0.20200805231151-a709e31e5d12
+  validUntil: 2029-08-04T17:01:13+03:00
+- path: github.com/golang/protobuf
+  minVersion: v1.4.2
```
na--, line 56: Why are these frozen?

na-- (approved):
LGTM, though I'm not sure we should adopt modtools as it is. I think it should only care about direct dependencies and we should leave the indirect dependency versions alone, they can be managed by whatever direct dependency imported them in the first place :man_shrugging:

modtools_frozen.yml
MStoykov: They are dependency of the `grpc` and I think we should update them when `grpc` updates them instead of whenever they get an update

MStoykov:
@na-- I definitely am not for adding `modtools` to a CI. It does (after you freeze some stuff) help check if there are updates instead of running 20-30 commands to see that. 

I guess you can do `go get -u .` but that updates *everything* and I at least prefer not to update some of the dependencies we still have. 

So I am for keeping the `modtools` yml file but running `modtools` manually ... at least for now, maybe it will turn out to be too cumbersome to keep updating the file if we keep getting dependencies we need to not update :man_shrugging:

imiric (approved):
I'm also not a fan of seeing yet another file in the repo root, but modtools sounds like a good idea. It should probably be a feature of `go mod`, but oh well :man_shrugging:

na-- (approved):
I didn't imagine we'll include modtools in CI. As I said, my biggest problem is that it can't filter out indirect dependencies when I ran it locally :man_shrugging: Not a big deal, just not very useful and has a lot of noise, so maybe some other tool is better? https://github.com/psampaz/go-mod-outdated ?

MStoykov:
In order to use it ... without it also reporting a bunch of dependencies we don't  want to I needed to do:
```
$ go list -u -m -mod=readonly -json all | go-mod-outdated -update -direct  | grep -E -v '(gopkg.in/guregu/null.v3|google.golang.org/protobuf|github.com/manyminds/api2go|github.com/kubernetes/helm|github.com/influxdata/influxdb1-client|github.com/golang/protobuf|github.com/DataDog/datadog-go|github.com/spf13/afero|github.com/mitchellh/mapstructure)'
+----------------------------------------+---------------------------------------+------------------------------------+--------+------------------+
|                 MODULE                 |                VERSION                |            NEW VERSION             | DIRECT | VALID TIMESTAMPS |
+----------------------------------------+---------------------------------------+------------------------------------+--------+------------------+
| github.com/fatih/color                 | v1.10.0                               | v1.11.0                            | true   | true             |
| github.com/klauspost/compress          | v1.11.13                              | v1.12.2                            | true   | true             |
| github.com/tidwall/gjson               | v1.7.4                                | v1.8.0                             | true   | true             |
| golang.org/x/crypto                    | v0.0.0-20201016220609-9e8e0b390897    | v0.0.0-20210513164829-c07d793c2f9a | true   | true             |
| golang.org/x/net                       | v0.0.0-20201209123823-ac852fbbde11    | v0.0.0-20210510120150-4163338589ed | true   | true             |
| google.golang.org/grpc                 | v1.36.1                               | v1.37.1                            | true   | true             |
+----------------------------------------+---------------------------------------+------------------------------------+--------+------------------+
```

Still no way to configure that `x/net` (for examle) needs to be from a particular branch :( 

I guess this can be a single bash script as well ?

na--:
I remember we had better tooling for dependencies before Go modules... :disappointed: :sweat_smile: 

:shipit:

PR #2029, opened 2021-05-19
Add t.Parallel to more tests in core and core/local packages

MStoykov:


imiric (approved):
I wish there was a way to make `t.Parallel()` apply for subtests as well to avoid the boilerplate...

PR #2038, opened 2021-05-21
Arrival rate fixes

MStoykov:


imiric (approved):
LGTM, nicely done!

lib/executor/ramping_arrival_rate.go
```diff
@@ -517,15 +518,18 @@ func (p *activeVUPool) Running() uint64 {
...
 		defer p.wg.Done()
 
+		close(ch)
 		for range p.iterations {
 			atomic.AddUint64(&p.running, uint64(1))
 			runfn(ctx, avu)
 			atomic.AddUint64(&p.running, ^uint64(0))
 		}
 	}()
+	<-ch
```
na--, line 532: I am confused by this :confused: I read your explanation in https://github.com/k6io/k6/pull/2038/commits/5df0e6b0032c122bebbe537280c8731d4c89273c, but if this extra bit of waiting helps all that much, then we probably have a bigger conceptual problem, since this is not very different from `sleep(time.Millisecond)`. Can you point me to what exactly failed when this wasn't present?

MStoykov: This is basically waiting for the goroutine to have *started* and then it tells you it has Added a VU. 

What fails are all the usual TestConstantArrivalRateRunCorrectTiming as shown in https://github.com/k6io/k6/runs/2640090549?check_suite_focus=true, which at least in my testing happens as we :
1. add all the VUs
2. try to start an iteration
3. none of the VUs has already gotten to the line `for range p.iterations` as it probably didn't even initialize the goroutine yet.

This is particularly easy to reproduce with `db89d213` and `go test -count 1 -cpu 1 -race -run TestConstantArrivalRateRunCorrectTiming`, specifically the `-cpu 1` as with `-cpu 8` I in practice can't reproduce it - probably as there is a free slot for that goroutine to always start immediately or at least fast enough.


This as we now no longer have a "semapthore" (using the channel `activeVUs`) but instead have workers. The semaphore was initialized *synchronously* while the workers are *asynchronous* for the most part - we just say "start this goroutine" and with this change at least wait for the "start the goroutine" part instead of hoping the go will schedule it fast enough. Which is totally different from what `time.Sleep(time....)` will do

na--: My point was that if the CPU contention was so heavy, it's possible that a goroutine wouldn't be allowed to run for a long while, even after it was started, so it might not get to run the `for` loop after it closes the channel :man_shrugging: 

I get how this is different from a `Sleep`, but as far as I understand, it's not actually locking anything that raced before... So if it failed before, this doesn't _guarantee_ it will work afterwards? I guess it would just greatly reduce the changes of a test failing, which might be good enough, though a better solution would be to improve the test, with https://github.com/k6io/k6/issues/1986 or something else... 

So yeah, I don't mind this code, I just didn't understand what it solved.

na-- (approved):
LGTM :crossed_fingers:

PR #2039, opened 2021-05-21
t.Parallel lib/execution

MStoykov:


imiric (approved):
LGTM, I like the extracted `handleVUTest` struct, it avoids the boilerplate and makes the tests more succinct :+1:

Though why not use it also for `TestVUHandleRace`, `TestVUHandleStartStopRace` and `BenchmarkVUHandleIterations`?

na-- (approved):

PR #2041, opened 2021-05-25
Don't panic on binaryResponses where the body wasn't set

MStoykov:
This panic was still caught and didn't abort the whole k6 just the
current iteration.


imiric (approved):
Thanks!

MStoykov:
The test failures should be fixed by #2038

na-- (approved):

PR #2045, opened 2021-05-31
Fix respect `throw` option with invalid URLs

### LINKED ISSUE #1502: `throw` option not respected with malformed URLs

imiric:
This is a continuation of #1853, with some changes as discussed in that PR.

Closes #1502


js/modules/k6/http/request.go
```diff
@@ -120,19 +115,28 @@ func (h *HTTP) Request(ctx context.Context, method string, url goja.Value, args
...
 		return nil, ErrHTTPForbiddenInInitContext
 	}
 
+	u, err := httpext.ToURL(reqURL.Export())
+	if err != nil {
+		if state.Options.Throw.Bool {
+			return nil, err
+		}
+		state.Logger.WithField("error", err).Warn("Request Failed")
+		return nil, nil
```
imiric, line 132: This `nil, nil` return is sketchy, but it would need defining another error for the `http` module since we can't check for the specific `K6Error` error code, as they're not exported. So I'd prefer fixing this later after the error codes refactor.

lib/netext/httpext/batch.go
```diff
@@ -57,7 +57,11 @@ func MakeBatchRequests(
 	result := make(chan error, reqCount)
 	perHostLimiter := lib.NewMultiSlotLimiter(perHostLimit)
 
-	makeRequest := func(req BatchParsedHTTPRequest) {
+	makeRequest := func(req *BatchParsedHTTPRequest) {
+		if req == nil {
+			result <- nil
+			return
```
imiric, line 63: We might want to optimize this so that a goroutine isn't created at all for nil `req`s, but that would involve some shuffling of values around to ensure they match the index of the request as specified in the script, so it was easier to just exit early.

js/modules/k6/http/request.go
```diff
@@ -104,8 +99,8 @@ func (h *HTTP) Request(ctx context.Context, method string, url goja.Value, args
 		params = args[1]
 	}
 
-	req, err := h.parseRequest(ctx, method, u, body, params)
-	if err != nil {
+	req, err := h.parseRequest(ctx, method, url, body, params)
+	if err != nil || req == nil {
 		return nil, err
```
imiric, line 104: One difference here with how this was done in #1853 is that the returned `Response` will be `undefined` in JS.

To me this makes more sense since if the request failed during parsing and we never made the call, we shouldn't create a `Response` object for it either.

lib/netext/httpext/error_codes.go
```diff
@@ -233,7 +235,10 @@ func NewK6Error(code errCode, msg string, originalErr error) K6Error {
 
 // Error implements the `error` interface, so K6Errors are normal Go errors.
 func (k6Err K6Error) Error() string {
-	return k6Err.Message
+	if k6Err.OriginalError == nil {
+		return k6Err.Message
+	}
+	return fmt.Sprintf("%s: %s", k6Err.Message, k6Err.OriginalError)
```
imiric, line 241: The [Windows failure](https://github.com/k6io/k6/pull/2045/checks?check_run_id=2711529014) seems to be because of this change.

Is this a wrong pattern after Go 1.13? To me it would make sense to render the original error as well, so I'm inclined to just update the [Windows error](https://github.com/k6io/k6/blob/2692aa9ce84cc9866953ce897b96767c089b4d6c/core/local/local_test.go#L1039).

js/modules/k6/http/request.go
MStoykov: The problem is that now 
```
var resp = http.get("http:// my bad url.com")
resp.status
```
Will be an exception of the type `TypeError: Cannot read property 'status' of undefined` as will anything else touching the response. 
This also will be the first time we actually have `null` returned for the response when we don't actually also throw an exception (return error).

SO at least return an "empty" Response ?

imiric: Yeah, that would be a breaking change. Though like I mentioned [here](https://github.com/k6io/k6/pull/2045#discussion_r642532999), this makes more sense to me since there was no request made, regardless of the value of `throw`. Which is not the same as if we actually made the request and should return some response.

But yeah, you're probably right that an empty `Response` as done in #1853 would be safer.

lib/netext/httpext/error_codes.go
na--: I think you shouldn't change this. Half of the purpose of `K6Error` is for us to be able to set a specific user-friendly error message, while at the same time keeping the original error accessible. If you want the original error message to be part of the new error message, you can always include it in the constructor, like you do in `ToURL`.

imiric: OK, I think this is a bit clunky, but resolved by 283d4bdd.

js/modules/k6/http/request.go
imiric: No longer an issue after f2665be0.

lib/netext/httpext/batch.go
imiric: No longer an issue after f2665be0.

js/modules/k6/http/request.go
imiric: Resolved by f2665be0.

imiric:
@na-- @MStoykov This is ready for review now.

I did things a bit differently than discussed in [#1853](https://github.com/k6io/k6/pull/1853#discussion_r575075757). See my rationale in f2665be0.

js/modules/k6/http/request.go
```diff
@@ -120,19 +124,24 @@ func (h *HTTP) Request(ctx context.Context, method string, url goja.Value, args
...
-	ctx context.Context, method string, reqURL httpext.URL, body interface{}, params goja.Value,
+	ctx context.Context, method string, reqURL goja.Value, body interface{}, params goja.Value,
 ) (*httpext.ParsedHTTPRequest, error) {
 	rt := common.GetRuntime(ctx)
 	state := lib.GetState(ctx)
 	if state == nil {
 		return nil, ErrHTTPForbiddenInInitContext
 	}
 
+	u, err := httpext.ToURL(reqURL.Export())
```
na--, line 135: Hmm I think `goja.Runtime.ToValue()` is a relatively expensive operation. We'd probably be better off if `reqURL` was of an `interface{}` type here, and we had a check like this here:


```suggestion
	if urlJSValue, ok := reqURL.(goja.Value); ok {
    	reqURL = urlJSValue.Export()
	}
	u, err := httpext.ToURL(reqURL)
```

This way we won't have to covert the url from a JS value to a Go one, and back to a JS value, and back to a go value.

js/modules/k6/http/request.go
```diff
@@ -520,13 +528,8 @@ func (h *HTTP) parseBatchRequest(
...
 			params = rt.ToValue(p)
 		}
-
-	default:
-		// Handling of "http://example.com/" or http.url`http://example.com/{$id}`
-		reqURL, err = ToURL(data)
-		if err != nil {
-			return nil, err
-		}
+	case string, httpext.URL:
```
MStoykov, line 531: Can you keep the comment from the previous `default` clause ?

js/modules/k6/http/request.go
```diff
@@ -444,7 +454,11 @@ func (h *HTTP) Batch(ctx context.Context, reqsV goja.Value) (goja.Value, error)
 	}
 
 	if err != nil {
-		return nil, err
+		if state.Options.Throw.Bool {
+			return nil, err
+		}
+		state.Logger.WithField("error", err).Warn("A batch request failed")
+		return rt.ToValue(results), nil
```
MStoykov, line 461: Reading [ToValue](https://pkg.go.dev/github.com/dop251/goja?utm_source=godoc#Runtime.ToValue) documentation has the following paragraph(near the very end, a few pages down):
```
Any other Go function is wrapped so that the arguments are automatically converted into the required Go types and the return value is converted to a JavaScript value (using this method). If conversion is not possible, a TypeError is thrown. 
```

Which to me means that most of our usage of `rt.ToValue` is probably not necessary, including this one.

na--: @MStoykov WDYT about https://github.com/k6io/k6/pull/2045#discussion_r646379357?

MStoykov: Actually looking at `common.Bind` where we actually return `func(call goja.FunctionCall) goja.Value` and have to `ToValue` we apparently call it regardless ... so again even less needed. But Goja would've called it, either way, it's just that in this case we wrap this particular function call ... instead of goja doing it :facepalm: 
https://github.com/k6io/k6/blob/da5eebc638d9917cedfe98e232a7eed4cb69a157/js/common/bridge.go#L246-L253

Also the part about multiple return values will not work ... and you can very easily make k6 panic if you write the wrong go code like `func SomeThingThatGetsBinded() (string, string)` ....

js/modules/k6/http/request.go
MStoykov: Probably true .... but given this `common.Bind` code (among other):
https://github.com/k6io/k6/blob/da5eebc638d9917cedfe98e232a7eed4cb69a157/js/common/bridge.go#L231-L236
I am not certain ... how much of a saving this will be :shrug:

js/modules/k6/http/request.go
imiric: Hhmm but the comment doesn't make sense anymore since it was for the `ToURL()` call, which is now done in `parseRequest()`.

Though would it be better to leave this as `default` and just pass anything to `rt.ToValue()`?

js/modules/k6/http/request.go
imiric: Resolved by 39a0ed9f.

js/modules/k6/http/request.go
imiric: Thanks for confirming this. I've seen both usages used, but it's good to know we don't have to convert it beforehand.

Resolved by 6b8cd309.

js/modules/k6/http/request.go
```diff
@@ -104,9 +104,13 @@ func (h *HTTP) Request(ctx context.Context, method string, url goja.Value, args
...
 
-	req, err := h.parseRequest(ctx, method, u, body, params)
+	req, err := h.parseRequest(ctx, method, url, body, params)
 	if err != nil {
-		return nil, err
+		if state.Options.Throw.Bool {
+			return nil, err
+		}
+		state.Logger.WithField("error", err).Warn("Request Failed")
+		return &Response{Response: &httpext.Response{}}, nil
```
MStoykov, line 113: I am for this to be created from some function and to include *more* stuff as currently `response.json()` ends up as a panic instead of simple(r) exception, this will likely be true for everything. 
Also I would argue that this should still have `error` and `error_code` set accordingly to what the problem was so you can detect what actually got wrong in the code.

imiric: :-/ I still think it's strange to work with a `Response` object if a request was never made. IMO the correct behavior in this case would be to simply throw the parse exception and let the user handle it with `try`/`except` as usual, in which case they'd have access to the error object. But since the Throw option is so old and we don't promote `try`/`except` at all (it's nowhere in our docs :disappointed:) we've backed ourselves into this corner where we have to return a fake-valid `Response` object. Would it be too late to bring up the discussion of deprecating Throw?

I'll see how I can improve things so that at least there are no panics, but this feels way off.

na--: Yeah... :disappointed: This is definitely one of the things that we have to improve in the new HTTP API... Not sure exactly how we should improve it, considering the context and requirements of load testing, but it has so many corner cases and footguns... :disappointed:

na-- (approved):
LGTM, or at least didn't spot any issues in the code, but I haven't actually tested it for other things like `https://github.com/k6io/k6/pull/2045#discussion_r646554675`

js/modules/k6/http/request.go
MStoykov: For me, this is a requirement for the merging of this PR

js/modules/k6/http/request.go
```diff
@@ -444,7 +475,11 @@ func (h *HTTP) Batch(ctx context.Context, reqsV goja.Value) (goja.Value, error)
 	}
 
 	if err != nil {
-		return nil, err
+		if state.Options.Throw.Bool {
+			return nil, err
+		}
+		state.Logger.WithField("error", err).Warn("A batch request failed")
```
imiric, line 481: Are we fine with this returning a different message than for a single request failure (`Request Failed`)? I thought it would be useful for troubleshooting to know it was a batch request, but maybe this should have the same message for consistency?

na-- (approved):

na-- (approved):
LGTM, though again, I haven't tested this yet, I'll do it after we merge in `master`, together with everything else for the next release

MStoykov (approved):

PR #2046, opened 2021-06-02
Error refactoring and exit code fixes

na--:
This PR does a few things:
- refactors (and I think simplifies) a bunch of error handling and exit code determination
- fixes the bug of k6 returning an exit code of `103` for script errors in `VU>0` initialization, while preserving the helpful hints
- (breaking change) changes one `k6 cloud` exit code from 99 to 97, to make it distinct from the already existing `99` exit code for `k6 run` (the k6 cloud code doesn't always mean that the thresholds have failed)

### DIFF
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -171,7 +171,7 @@ func readEnvConfig() (Config, error) {
 // TODO: add better validation, more explicit default values and improve consistency between formats
 // TODO: accumulate all errors and differentiate between the layers?
 func getConsolidatedConfig(fs afero.Fs, cliConf Config, runner lib.Runner) (conf Config, err error) {
-	// TODO: use errext.WithExitCode(err, exitcodes.InvalidConfig) where it makes sense?
+	// TODO: use errext.WithExitCodeIfNone(err, exitcodes.InvalidConfig) where it makes sense?
 
 	fileConf, _, err := readDiskConfig(fs)
 	if err != nil {
@@ -232,7 +232,7 @@ func deriveAndValidateConfig(conf Config, isExecutable func(string) bool) (resul
 	if err == nil {
 		err = validateConfig(result, isExecutable)
 	}
-	return result, errext.WithExitCode(err, exitcodes.InvalidConfig)
+	return result, errext.WithExitCodeIfNone(err, exitcodes.InvalidConfig)
 }
 
 func validateConfig(conf Config, isExecutable func(string) bool) error {
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -33,6 +33,8 @@ import (
 	"github.com/spf13/pflag"
 	"gopkg.in/guregu/null.v3"
 
+	"go.k6.io/k6/errext"
+	"go.k6.io/k6/errext/exitcodes"
 	"go.k6.io/k6/lib"
 	"go.k6.io/k6/lib/executor"
 	"go.k6.io/k6/lib/types"
@@ -169,6 +171,8 @@ func readEnvConfig() (Config, error) {
 // TODO: add better validation, more explicit default values and improve consistency between formats
 // TODO: accumulate all errors and differentiate between the layers?
 func getConsolidatedConfig(fs afero.Fs, cliConf Config, runner lib.Runner) (conf Config, err error) {
+	// TODO: use errext.WithExitCode(err, exitcodes.InvalidConfig) where it makes sense?
+
 	fileConf, _, err := readDiskConfig(fs)
 	if err != nil {
 		return conf, err
@@ -225,10 +229,10 @@ func applyDefault(conf Config) Config {
 func deriveAndValidateConfig(conf Config, isExecutable func(string) bool) (result Config, err error) {
 	result = conf
 	result.Options, err = executor.DeriveScenariosFromShortcuts(conf.Options)
-	if err != nil {
-		return result, err
+	if err == nil {
+		err = validateConfig(result, isExecutable)
 	}
-	return result, validateConfig(result, isExecutable)
+	return result, errext.WithExitCode(err, exitcodes.InvalidConfig)
 }
 
 func validateConfig(conf Config, isExecutable func(string) bool) error {
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -22,6 +22,7 @@ package cmd
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -38,6 +39,7 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 
+	"go.k6.io/k6/errext"
 	"go.k6.io/k6/lib/consts"
 	"go.k6.io/k6/log"
 )
@@ -178,23 +180,37 @@ func Execute() {
 	)
 
 	if err := c.cmd.Execute(); err != nil {
-		fields := logrus.Fields{}
-		code := -1
-		if e, ok := err.(ExitCode); ok {
-			code = e.Code
-			if e.Hint != "" {
-				fields["hint"] = e.Hint
-			}
+		exitCode := -1
+		var ecerr errext.HasExitCode
+		if errors.As(err, &ecerr) {
+			exitCode = int(ecerr.ExitCode())
 		}
 
-		logger.WithFields(fields).Error(err)
+		hint := ""
+		var herr errext.HasHint
+		if errors.As(err, &herr) {
+			hint = herr.Hint()
+		}
+
+		errText := err.Error()
+		var xerr errext.Exception
+		if errors.As(err, &xerr) {
+			errText = xerr.StackTrace()
+		}
+
+		fields := logrus.Fields{}
+		if hint != "" {
+			fields["hint"] = hint
+		}
+
+		logger.WithFields(fields).Error(errText)
 		if c.loggerIsRemote {
-			fallbackLogger.WithFields(fields).Error(err)
+			fallbackLogger.WithFields(fields).Error(errText)
 			cancel()
 			c.waitRemoteLogger()
 		}
 
-		os.Exit(code)
+		os.Exit(exitCode) //nolint:gocritic
 	}
 
 	cancel()
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -186,12 +186,6 @@ func Execute() {
 			exitCode = int(ecerr.ExitCode())
 		}
 
-		hint := ""
-		var herr errext.HasHint
-		if errors.As(err, &herr) {
-			hint = herr.Hint()
-		}
-
 		errText := err.Error()
 		var xerr errext.Exception
 		if errors.As(err, &xerr) {
@@ -199,8 +193,9 @@ func Execute() {
 		}
 
 		fields := logrus.Fields{}
-		if hint != "" {
-			fields["hint"] = hint
+		var herr errext.HasHint
+		if errors.As(err, &herr) {
+			fields["hint"] = herr.Hint()
 		}
 
 		logger.WithFields(fields).Error(errText)
--- a/errext/exit_code.go
+++ b/errext/exit_code.go
@@ -1,0 +1,66 @@
+/*
+ *
+ * k6 - a next-generation load testing tool
+ * Copyright (C) 2021 Load Impact
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+package errext
+
+import "errors"
+
+// ExitCode is the code with which the application should exit if this error
+// bubbles up to the top of the scope. Values should be between 0 and 125:
+// https://unix.stackexchange.com/questions/418784/what-is-the-min-and-max-values-of-exit-codes-in-linux
+type ExitCode uint8
+
+// HasExitCode is a wrapper around an error with an attached exit code.
+type HasExitCode interface {
+	error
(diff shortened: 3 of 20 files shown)


errext/exit_code.go
```diff
@@ -0,0 +1,66 @@
...
+func WithExitCode(err error, exitCode ExitCode) error {
+	if err == nil {
+		// No error, do nothing
+		return nil
+	}
+	var ecerr HasExitCode
+	if errors.As(err, &ecerr) {
+		// The given error already has an exit code, do nothing
+		return err
+	}
```
MStoykov, line 49: I think I would never expect this to happen.

Is there a situation where that is a problem?

js/runner.go
```diff
@@ -753,3 +755,11 @@ func (s *scriptException) StackTrace() string {
 func (s *scriptException) Unwrap() error {
 	return s.inner
 }
+
+func (s *scriptException) Hint() string {
+	return "script exception"
+}
+
+func (s *scriptException) ExitCode() errext.ExitCode {
+	return exitcodes.ScriptException
```
imiric, line 764: Like I mentioned on Slack, it seems out of place to have non-`cmd` packages define what exit codes to use. I would prefer returning `js`-specific errors here, and mapping them to exit codes in `cmd`. If a huge `switch` is unavoidable, I'd still prefer having that in a single place than handling exit codes in internal packages.

js/timeout_error.go
```diff
@@ -18,44 +18,41 @@
...
 	"time"
 
+	"go.k6.io/k6/errext"
+	"go.k6.io/k6/errext/exitcodes"
 	"go.k6.io/k6/lib/consts"
 )
 
-// TimeoutError is used when somethings timeouts
-type TimeoutError struct {
+// timeoutError is used when somethings times out.
```
imiric, line 32:
```suggestion
// timeoutError is used when some operation times out.
```

js/timeout_error.go
```diff
@@ -18,44 +18,41 @@
  *
  */
 
-package lib
+package js
```
imiric, line 21: This is great since it was only used in `js`, but could we rename the file to `errors.go`? If we follow the convention that every package should return its own set of errors, naming it `timeout_error.go` is not open for expansion and someone might define errors elsewhere.

imiric (commented):
This is cleaner than before, and I like the `With*` and interfaces approach, but I'd prefer it if `errext` wasn't a top-level package, but maybe `cmd/errors`. There's really no reason an internal package should ever import it, and having it globally available suggests otherwise.

errext/exit_code.go
na--: I use this property in https://github.com/k6io/k6/blob/1d741a854b1b18cd3af99f397d857375a630648c/cmd/run.go#L248-L252

The idea is that, since you're wrapping errors, you generally want to keep the most specific exit code possible. So having the `WithExitCode` not overwrite an already existing attached error code is very helpful - in the above code, if we already had a an exit code (e.g. a script error) we'll exit with its code (e.g. `107`), but if we don't, we'd still attach a the more generic one there.

na--: And, of course, you can always negate that with a custom error type (or another helper), if some other use case requires masking of a previously attached exit code. But this makes sense to me as the default behavior.

js/timeout_error.go
na--: didn't realize `InitContextError` even existed :face_with_head_bandage: :+1: what you suggest sounds good to me

cmd/config.go
```diff
@@ -225,10 +229,10 @@ func applyDefault(conf Config) Config {
 func deriveAndValidateConfig(conf Config, isExecutable func(string) bool) (result Config, err error) {
 	result = conf
 	result.Options, err = executor.DeriveScenariosFromShortcuts(conf.Options)
-	if err != nil {
-		return result, err
+	if err == nil {
+		err = validateConfig(result, isExecutable)
 	}
-	return result, validateConfig(result, isExecutable)
+	return result, errext.WithExitCode(err, exitcodes.InvalidConfig)
```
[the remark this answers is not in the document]
na--: hmm this seemed very idiomatic to me, I think a lot of error wrappers in Go work like this, but I might be wrong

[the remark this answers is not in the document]
na--: @MStoykov WDYT?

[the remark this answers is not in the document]
MStoykov: I am fine with wrapping nil errors to return nil errors ... I still feel it's a bit off, but at least I don't think there is ever a case where I would want to wrap a *nil* error so whenever I try to wrap one it should return `nil` instead which is what currently happens

js/runner.go
na--: Hmm I agree this isn't the cleanest solution, but I don't see what we're sacrificing (besides "purity") with this approach, and it seems better than the alternatives :man_shrugging: 

The problem with having a huge `switch`, besides the ugliness of the `switch` itself, is that we'd have to make all of the error types exported, so we can match against them in `cmd`, i.e. everything will actually be much more tightly coupled, even if we ignore the ugly `switch`. In this way, certain local error types would have to know about "exit codes", but it still seems better than the alternative. Besides, with a `switch`, we'd always have to remember to update it every time we add a new error code, while this way errors are localized where they're emitted.

Originally, I thought about having the constants in `errext/exitcodes` be simple `iota` automatically-generated ones, and then to have a huge `map[errext.ExitCode]int` in `cmd/` with the actual `int` exit codes, but that just seems unnecessary and would mean that every Go program that wants to get the error codes will have to import `cmd`.

cmd/config.go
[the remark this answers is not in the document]
imiric: OK, I'm fine with it then. :)

js/runner.go
imiric: Yeah, the coupling is a problem, though we could just check against known interfaces defined in `cmd` and extract details about the error that would determine the status code it maps to. It's a bit of a roundabout way of doing things, but IMO cleaner than everything importing a global `errext` package.

But I'd have to play around with this to know for sure, so if you think it's unfeasible I'm OK with keeping it as is.

> would mean that every Go program that wants to get the error codes will have to import `cmd`

That makes sense to me since these are CLI (`cmd`) exit codes. If `cmd` is too much to import, would a subpackage like `cmd/errors` be better?

na--: btw part of the idea for `errext` was that it would allow us to chip away at some of the `lib/` mess. Making a new global package for just errors might seem a bit much, but we do quite a lot of error handling in k6, so it's not too outlandish.

`errext` (and its sub-packages) wont't import any other k6 package, so I think they can have a somewhat clean-ish structure:
- `/errext/`: interface types and small helpers
- `/errext/exitcodes` - only constants - I can understand why having them in `cmd` makes sense, but having them and nothing else in this package is also very clean (regardless of whether we have a big `switch` in cmd that references them, or whether we stick with my proposal in this PR where every error knows its code)
- `/errext/errorcodes` - this seems like a suitable place for the `error_code` detection, once we take it out of `lib/netext/httpext` so we can use it for other things like web sockets or gRPC

js/timeout_error.go
na--: Unfortunately, we can't do a lot about `InitContextError`. It is essentially used as the equivalent of `throw new Error()`, only from the Go side. We can't un-export it, since it's used by multiple different packages. And we can't move it to the `js/` package from `js/common/`, since that causes an import loop :disappointed: 

As things currently are in this PR, the `InitContextErrors` are detected as script errors, they bubble up from goja and have an appropriate stacktrace. For example, when using a `group()` in the init context:
```
ERRO[0000] GoError: Using group() in the init context is not supported
	at go.k6.io/k6/js/common.Bind.func1 (native)
	at file:///path/to/script.js:28:19(59)  hint="error while initializing VU #2 (script exception)"
```

The biggest problem IMO is that `GoError` and `at go.k6.io/k6/js/common.Bind.func1 (native)` are very confusing, but I'll create a separate issue about that.

na--: Created a new issue in https://github.com/k6io/k6/issues/2049

errext/exitcodes/codes.go
```diff
@@ -0,0 +1,39 @@
...
+// Package exitcodes contains the constants representing possible k6 exit error codes.
+//nolint: golint
+package exitcodes
+
+import "go.k6.io/k6/errext"
+
+const (
+	CloudTestRunFailed       errext.ExitCode = 97 // This used to be 99 before k6 v0.33.0
+	CloudFailedToGetProgress errext.ExitCode = 98
+	ThresholdHaveFailed      errext.ExitCode = 99
```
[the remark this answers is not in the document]
na--: Resolved by the cleanup [force-push](https://github.com/k6io/k6/compare/1d741a854b1b18cd3af99f397d857375a630648c..8771485908bf0be49e0855788f5534c0b96a12d3).

js/timeout_error.go
na--: Resolved by the cleanup [force-push](https://github.com/k6io/k6/compare/1d741a854b1b18cd3af99f397d857375a630648c..8771485908bf0be49e0855788f5534c0b96a12d3).

na--:
@imiric I tried to shuffle things around, but adding any more things to `cmd` just made things messier and more error-prone. Even just moving the exit codes in there made things more difficult to test... :disappointed: I may be missing some clever trick on how to have them tidy, so do you want to make a PoC PR on top of this one or `master` with how you think this should be handled?

js/timeout_error.go
imiric: Ah, that's a shame. We can leave it as is then, it's not a huge problem.

imiric:
@na-- I doubt I'll be able to make progress if you haven't :smile: I'll take a look, but let's not delay merging this because of it.

js/timeout_error.go
```diff
@@ -66,3 +63,16 @@ func (t TimeoutError) Hint() string {
 	}
 	return hint
 }
+
+// ExitCode returns the coresponding exit code value to the place.
+func (t timeoutError) ExitCode() errext.ExitCode {
+	// TODO: add handleSummary()
```
imiric, line 69: Why leave this for later?

imiric (commented):
@na-- I some trivial changes in #2050, but otherwise this LGTM.

js/timeout_error.go
na--: because I didn't want to add any new exit codes in this PR, and to be honest, I'm not sure this deserves its own separate exit code, `genericTimeout` is probably good enough :man_shrugging: We only have 128 exit codes, we should be somewhat conservative with them

imiric (approved):

errext/exit_code.go
MStoykov: I am totally not sold on the "This is already wrapped in error fo the "same type" so I will silently not wrap it for you" idea. IMO w/e code will exit with the code should check whether it was wrapped multiple times and decide whether it should exit with the most inner one or not. Or for the code that is wrapping to check if it has ErrorCode and either wrap it again ... or not.

I am win with `WithExitCodeIfNotOneAlready` being added

errext/hint.go
```diff
@@ -0,0 +1,63 @@
...
+func WithHint(err error, hint string) error {
+	if err == nil {
+		// No error, do nothing
+		return nil
+	}
+	var oldhint HasHint
+	if errors.As(err, &oldhint) {
+		// The given error already had a hint, wrap it
+		hint = hint + " (" + oldhint.Hint() + ")"
+	}
```
MStoykov, line 46: I would prefer if this is done in the `Hint` method

js/errors.go
```diff
@@ -0,0 +1,109 @@
...
+	"go.k6.io/k6/lib/consts"
+)
+
+// scriptException is used for errors that resulted from a script exception and
+// contain a stack trace that lead to them.
+type scriptException struct {
+	inner *goja.Exception
+}
+
+var _ errext.Exception = &scriptException{}
```
MStoykov, line 40: You should probably check for the `HasHint` and `HasErrorCode` while you are at it.

MStoykov: although to be honest I am not certain if I wouldn't have preferred there to be `newScriptExceiption()` which calls 

`WithHint(WithExitCode(scriptException{inner: ex}...))` 

On the other hand my biggest problem is that errext.ScriptException implementations can have different exitcodes ... which again is the whole problem with the exit codes being codified by whatever returns them instead of what will exit with them ...

MStoykov (commented):
I am not going to "request changes", but I am still not on board with the whoever returns errors to say what the exit code should be. 

I have left some other comments which mostly resolve around problems that I think raise from that.

errext/hint.go
na--: yes, that will be much cleaner :+1:

errext/exit_code.go
na--: hmm fair enough... I am not sure if I should just rename this to `WithExitCodeIfNotOneAlready` / `WithExitCodeIfNone`, since we don't really have a use case for something that always attaches an exit code yet, in most places it either doesn't matter (root errors deep in the code that first attach an exit code) or we don't want to overwrite it

alternatively, I can rip it out and only check for existing exit codes in a few places, I'll see how that looks like

js/errors.go
na--: Decided not to wrap this in `WithHint(WithExitCode())`, it didn't really improve things in terms of readability or clarity

errext/exit_code.go
na--: Is this ok? https://github.com/k6io/k6/pull/2046/commits/d8e41ec4a5b7394a192dc5410ebf5e647cab242d

cmd/root.go
```diff
@@ -178,23 +180,37 @@ func Execute() {
...
+		if errors.As(err, &ecerr) {
+			exitCode = int(ecerr.ExitCode())
+		}
+
+		hint := ""
+		var herr errext.HasHint
+		if errors.As(err, &herr) {
+			hint = herr.Hint()
+		}
+
```
MStoykov, line 194: I think this can be move down to the place we actually set it to `fields`

na--: Resolved by https://github.com/k6io/k6/pull/2046/commits/bbd254196f60e752f37b8d9a2eedd70f9f617e20

na--: actually, wait, this can be simplified more

MStoykov: yeah I meant ... replacing the `if hint != '"` :) ... but probably should've worded it better

na--: Resolved by https://github.com/k6io/k6/pull/2046/commits/a7b85eba6b495e370f699ef855adada36bdb7eb3

PR #2047, opened 2021-06-03
Fix panic on missing files in cdnjs links

### LINKED ISSUE #2037: panic when importing modules from cdnjs
From https://community.k6.io/t/out-of-the-blue-k6-is-throwing-panic/1762

This script causes a panic:
```js
import faker from "cdnjs.com/libraries/Faker";

export default function () {
}
```

MStoykov:
fixes #2037

### DIFF
--- a/loader/cdnjs.go
+++ b/loader/cdnjs.go
@@ -73,6 +73,11 @@ func cdnjs(logger logrus.FieldLogger, path string, parts []string) (string, erro
 			if ver.Version != version {
 				continue
 			}
+			if len(ver.Files) == 0 {
+				return "",
+					fmt.Errorf("cdnjs: no files for version %s of %s, this is a problem with the library or cdnjs not k6",
+						version, path)
+			}
 			backupFilename = ver.Files[0]
 			for _, file := range ver.Files {
 				if file == filename {


loader/loader.go
```diff
@@ -242,12 +242,14 @@ func Load(
 }
 
 func resolveUsingLoaders(logger logrus.FieldLogger, name string) (*url.URL, error) {
-	_, loader, loaderArgs := pickLoader(name)
+	loaderName, loader, loaderArgs := pickLoader(name)
 	if loader != nil {
 		urlString, err := loader(logger, name, loaderArgs)
 		if err != nil {
+			logger.Warnf("Using magic urls for `%s` is deprated. Please use normal urls", loaderName)
```
imiric, line 249:
```suggestion
			logger.Warnf("Using magic URLs for `%s` is deprecated. Please use normal URLs", loaderName)
```

Though I'm wondering if people will understand what we mean by "magic URLs". We don't use that name anywhere, I think only internally.

loader/loader.go
```diff
@@ -242,12 +242,14 @@ func Load(
...
 			return nil, err
 		}
+		logger.Warnf("Using magic urls for `%s` is deprated. Please just use the full url `%s`", loaderName, urlString)
```
imiric, line 252:
```suggestion
		logger.Warnf("Using magic URLs for `%s` is deprecated. Please use the full URL instead: `%s`", loaderName, urlString)
```

imiric (commented):
:+1:

@MStoykov Have you considered setting up a typo checker/linter locally? :smile:

na--:
I thought the consensus was that we'd only deprecate the cdnjs "magic" loader? :confused: Also, please reference the issue this closes.

Btw I also need to find a good spell checker that works well with comments and strings at least, though ideally also with CamelCase variable names... :sweat_smile: Any recommendations @imiric?

MStoykov:
> I thought the consensus was that we'd only deprecate the cdnjs "magic" loader?

I have always been of the opinion that we will deprecate the whole *magic* not just one of the two parts. 

>  Also, please reference the issue this closes.

It is referenced in the commit https://github.com/k6io/k6/pull/2047/commits/425c2d83408b711417c3fe8729826d23dd0d613f

na-- (changes requested):
Please don't do that :pray: :smile: Every time you rebase the commit, the issue will have a separate mention. Moreover, when I open the PR, I see an empty description and have to go and hunt down the issue myself. The issue and PR should be linked directly.

Regarding the deprecation - I am fully on board with deprecating cdnjs, very few people use this and it's problematic (makes an extra network request). I am somewhat opposed to deprecating the github loader though. It's convenient (I use it myself pretty often), doesn't make extra requests and we've given a ton of examples with it in the docs, readme and forum... 

Deprecating the whole "magic" loader logic also doesn't seem needed, or indeed, even possible right now. We'd need to support running .tar archive bundles with these magic URLs for quite a while, so I don't see how we can rip out that support anytime soon. And the feature of being able to add custom `import` prefixes with dynamic Go logic behind them might be useful for other use cases in the future, e.g. xk6 extensions :man_shrugging:

MStoykov:
>  Regarding the deprecation - I am fully on board with deprecating cdnjs, very few people use this and it's problematic (makes an extra network request). I am somewhat opposed to deprecating the github loader though. It's convenient (I use it myself pretty often), doesn't make extra requests and we've given a ton of examples with it in the docs, readme and forum... 

> makes an extra network requests

Who cares given that we do a ton of other network requests :shrug: 

> It's convenient 

it is even *less* of a URL then the cdnjs one is ... if you put the cdnjs one in the browser you at least go to a page with the library, putting the github one will get you a 404. This goes the other way as well ... you actually need to know how to construct it ...

 > Deprecating the whole "magic" loader logic also doesn't seem needed, or indeed, even possible right now. 

deprecating it for both will definitely be *less* work as I won't need to keep one of them around.

And if we don't start we would never get it done ... so I am for deprecating *both* or *none* I see no point in having to explain why we deprecated a thing halfway through if I at least won't get to drop the *problematic* part of the code ... not the 20 lines of matching a json >.> 

Also I am sorry but I will continue to link to issues in the *commits* that *fix* the issue as that is what actually fixes the issue not the PR with possibly 20 commits in it.

imiric:
> Btw I also need to find a good spell checker that works well with comments and strings at least, though ideally also with CamelCase variable names...

:man_shrugging: I don't use one myself, just a bit OCD about it :sweat_smile:

I think aspell is the standard, but not sure how well it works with source code. From a quick search [misspell](https://github.com/client9/misspell) seems popular, and it's included in golangci-lint, we just need to enable it. Though these kinds of issues are better caught locally in your editor or in a pre-commit hook instead of later in CI, so I'd enable it in my editor.

> Please don't do that :pray: :smile: Every time you rebase the commit, the issue will have a separate mention.

I agree that's annoying, but it's a GitHub annoyance, not Git's. I think both commits *and* PRs should mention the issue number. The commit reference is very useful when viewing the commit in isolation.

Re: the deprecation, I agree with @MStoykov that it doesn't make much sense to deprecate just cdnjs and not the feature altogether. Since at least Ned finds it useful :smile: then I don't think we should deprecate any of it and let it be. It doesn't add a ton of complexity and it's an old feature we have no urgency to get rid of.

MStoykov:
I would also like to point out that using the github one seems less likely to be widely used than the cdnjs one given that the second one hosts all kinds of js libraries IMO.

Neither one (unless you use cdnjs with versions which ... again not a point in girthub's favor) have any guarantees that running the same script 5 minutes apart will get you the same file - both get w/e is latest.

So I am also for not deprecating anything if we are not going to deprecate both.

>  It doesn't add a ton of complexity and it's an old feature we have no urgency to get rid of. 

:rofl: https://github.com/k6io/k6/issues/1089 exist because https://github.com/k6io/k6/pull/1059 was trying to do something seemingly very simple, but between (old) archive support, magic URLs and cli arguments being both schemeless and `./`-less it was ... quite the nightmare to get it to a working state (it also had 2+ fixing PRs on top). So ... there is quite a lot of complexity, but we are unlikely to drop it for at least a year going forward(at least for archive runs) as we need to support it for clients at least until most of them upgrade to a newer version.

na-- (approved):

imiric (approved):
A test for this would be nice, but would require some refactoring :sweat_smile:

PR #2050, opened 2021-06-07
Minor tweaks to #2046

imiric:
Just renaming and moving of `scriptException`.

I also tried to move parts of `errext` to `cmd`, but as @na-- found it ends up being messier being split out of `errext`, and there's not much benefit from it.


na-- (approved):
LGTM :+1:

Issue #1923, opened 2021-03-24
HTTP Cookies merge fix proposal


MStoykov:
## Current state

Cookies in k6 can be sourced from 3 places:
1. the cookie jar used. Which is either the default VU jar or provided with the [params.jar](https://k6.io/docs/javascript-api/k6-http/params)
```javascript
    {
        let defaultJar = http.cookieJar();
        defaultJar.set("https://httpbin.test.k6.io/cookies", "key", "default value");
        let resp = http.get("https://httpbin.test.k6.io/cookies");
        console.log(resp.body);
    }

    {
        let jar = new http.CookieJar();
        jar.set("https://httpbin.test.k6.io/cookies", "key", "not default value");
        let resp = http.get("https://httpbin.test.k6.io/cookies", {jar: jar}); // Binks
        console.log(resp.body);
    }
```
Will produce 
```
INFO[0001] {
  "cookies": {
    "key": "default value"
  }
}  source=console
INFO[0001] {
  "cookies": {
    "key": "not default value"
  }
}  source=console
```
2. using params.cookies:
```javascript
        let resp = http.get("https://httpbin.test.k6.io/cookies", {cookies: {val: "key"}});
        console.log(resp.body);
```
produces
```
INFO[0001] {
  "cookies": {
    "val": "key"
  }
}       source=console
```
3. Just setting the `Cookie` header manually:
```javascript
        let resp = http.get("https://httpbin.test.k6.io/cookies", {headers: {Cookie: "key=headervalue"}});
        console.log(resp.body);
```
produces
```
INFO[0001] {
  "cookies": {
    "key": "headervalue"
  }
}  source=console
```

The current behaviour is as follows:
1. get the active jar (either from the VU or from params.jar).
2. add all cookies from `params.cookies` to the Cookie header through using `[Request.AddCookie](https://golang.org/pkg/net/http/#Request.AddCookie)`, which effectively appends it to the header `Cookie`.
3. For each cookie for the domain in the jar use [Request.AddCookie](https://golang.org/pkg/net/http/#Request.AddCookie) as well. This is skipped for cookies with key that was added by `params.cookies` that had `replace:true`. 

So if you have something like:
```javascript
    let jar = new http.CookieJar();
    jar.set("https://httpbin.test.k6.io/cookies", "key", "jar.value");
    let resp = http.get("https://httpbin.test.k6.io/cookies", {
        jar: jar, // Binks
[...]

PR #2052, opened 2021-06-08
Add cookie support to WebSockets

### LINKED ISSUE #1226: Support passing CookieJar to WebSocket through Params

imiric:
This is a continuation of #1650, with some minor cleanup.

Closes #1226

### DIFF
--- a/js/modules/k6/ws/ws_test.go
+++ b/js/modules/k6/ws/ws_test.go
@@ -164,7 +164,7 @@ func TestSession(t *testing.T) {
 				socket.send("test")
 			})
 			socket.on("message", function (data){
-				if (!data=="test") {
+				if (!(data=="test")) {
 					throw new Error ("echo'd data doesn't match our message!");
 				}
 				socket.close()
@@ -314,6 +314,24 @@ func TestSession(t *testing.T) {
 		assert.NoError(t, err)
 	})
 	assertSessionMetricsEmitted(t, stats.GetBufferedSamples(samples), "", sr("WSBIN_URL/ws-echo"), 101, "")
+
+	t.Run("client_cookie", func(t *testing.T) {
+		_, err := common.RunString(rt, sr(`
+		var params = {
+			cookies: { "Session": "123" },
+		};
+		var res = ws.connect("WSBIN_URL/ws-echo-cookie", params, function(socket){
+			socket.on("message", function (data){
+				if (!(data == "Session=123")) {
+					throw new Error ("echo'd data doesn't match our message!");
+				}
+				socket.close();
+			});
+		});
+		`))
+		assert.NoError(t, err)
+		assertSessionMetricsEmitted(t, stats.GetBufferedSamples(samples), "", sr("WSBIN_URL/ws-echo-cookie"), 101, "")
+	})
 
 	serverCloseTests := []struct {
 		name     string
--- a/js/modules/k6/ws/ws_test.go
+++ b/js/modules/k6/ws/ws_test.go
@@ -316,6 +316,7 @@ func TestSession(t *testing.T) {
 	assertSessionMetricsEmitted(t, stats.GetBufferedSamples(samples), "", sr("WSBIN_URL/ws-echo"), 101, "")
 
 	t.Run("client_cookie", func(t *testing.T) {
+		t.Parallel()
 		_, err := rt.RunString(sr(`
 		var params = {
 			cookies: { "Session": "123" },
--- a/js/modules/k6/ws/ws_test.go
+++ b/js/modules/k6/ws/ws_test.go
@@ -316,7 +316,7 @@ func TestSession(t *testing.T) {
 	assertSessionMetricsEmitted(t, stats.GetBufferedSamples(samples), "", sr("WSBIN_URL/ws-echo"), 101, "")
 
 	t.Run("client_cookie", func(t *testing.T) {
-		_, err := common.RunString(rt, sr(`
+		_, err := rt.RunString(sr(`
 		var params = {
 			cookies: { "Session": "123" },
 		};
--- a/js/modules/k6/ws/ws.go
+++ b/js/modules/k6/ws/ws.go
@@ -112,7 +112,6 @@ func (*WS) Connect(ctx context.Context, url string, args ...goja.Value) (*WSHTTP
 
 	// Leave header to nil by default so we can pass it directly to the Dialer
 	var header http.Header
-	var jar http.CookieJar
 
 	tags := state.CloneTags()
 
@@ -203,6 +202,8 @@ func (*WS) Connect(ctx context.Context, url string, args ...goja.Value) (*WSHTTP
 		v := cookie
 		cookies = append(cookies, &v)
 	}
+
+	var jar http.CookieJar
 	if len(cookies) > 0 {
 		jar, err = cookiejar.New(nil)
 		if err != nil {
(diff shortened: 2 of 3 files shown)


js/modules/k6/ws/ws_test.go
```diff
@@ -315,6 +315,25 @@ func TestSession(t *testing.T) {
 	})
 	assertSessionMetricsEmitted(t, stats.GetBufferedSamples(samples), "", sr("WSBIN_URL/ws-echo"), 101, "")
 
+	t.Run("client_cookie", func(t *testing.T) {
```
imiric, line 318: I tried adding a [test for `replace`](https://github.com/k6io/k6/pull/1650#discussion_r636233070), but it seems that cookies in the global cookie jar (`http.cookieJar()`) are not propagated to `ws.connect()`. I.e. this script doesn't send the cookie:

```js
import http from 'k6/http';
import ws from 'k6/ws';

export default function () {
  const jar = http.cookieJar();
  jar.set("ws://127.0.0.1:8080/", "Session", "123");
  var res = ws.connect("ws://127.0.0.1:8080/", null, function(socket){
    socket.on("message", function (data){
      console.log(data);
      socket.close();
    });
  });
}
```

So my question is: should it? Based on [this comment](https://github.com/k6io/k6/issues/1226#issuecomment-552424017) probably not as it would be a breaking change. But should we then have an equivalent `ws.cookieJar()`?

MStoykov: I think it should and I also think that is the *better* first change.

Unfortunately the web socket library we use takes *only* jars and as adding both a cookie and using the jar should probably not add the cookie to the jar you will need to:
1. create a new jar to copy the values from the provide one (or the default one) and add the cookie to that
2.  set any cookies set by the websocket one back in the original jar

Given that ... I am more for not having a way to specify a cookie but to
1. have a way to specify a jar
2. use the default one by default ;) 

And then whoever wants to specify a Cookie can just add it to the jar and pass the jar or continue setting the Header ... both of which arguably aren't that much harder especially as I would expect the bigger problem will be that you would've made an HTTP request it would set you the cookie in the jar, but then that won't be used by the websocket. And the someone less likely variant of websocket HTTP call setting a cookie that isn';t set on the default jar

imiric: @MStoykov I'm a bit confused by the changes you want :sweat_smile: Can you summarize them as commit messages? :)

FWIW `gorilla/websocket` does support setting a `Cookie` header and not just via a jar. See [Dialer.DialContext](https://pkg.go.dev/github.com/gorilla/websocket#Dialer.DialContext). There's [an issue about the header overriding the jar](https://github.com/gorilla/websocket/issues/597), but IMO this seems like desired behavior.

So the way I see it to handle this "properly" and achieve feature-parity with how it works for `http` we should:

- Add a per-VU WS-only cookie jar (`ws.cookieJar()`) from which we will copy to the `Dialer.Jar` for each new connection.

- Keep this `params` approach to override cookies per connection, maybe without using `Dialer.Jar` but simply setting it as a header. We should also read it from a manual header passed via `params`, i.e. `{ headers: { 'Cookie': 'Session=123' }}`.

- Allow passing a standalone jar via `params` (`new ws.CookieJar()`) that will be set to `Dialer.Jar` after merging any cookies set via `params` and in the global jar. We should document how overriding works to avoid surprises.

I would've preferred if all this could work with `http.cookieJar()`, since the only reason we need a WS approach is to avoid breaking existing scripts, but hopefully we can unify this when we refactor WS and after the new HTTP API.

WDYT? cc: @na--

MStoykov: > FWIW gorilla/websocket does support setting a Cookie header and not just via a jar. See Dialer.DialContext. There's an issue about the header overriding the jar, but IMO this seems like desired behavior.

As mentioned in this [comment](https://github.com/gorilla/websocket/pull/599#issuecomment-689163241) from the "fix" this seems like the *correct* behaviour ;).  And I agree with that.

>So the way I see it to handle this "properly" and achieve feature-parity with how it works for http we should: 

This is IMO the wrong thing to try to do given #1923 which literally spells out how not ... good and confusing the cookie works for HTTP. Also it was provoked by me trying to work on this exact PR/issue.

IMO this should :
1. Use the default cookiejar from lib.state by default.
2. be able to take `Params.jar` which is http.Cookiejar .. .we can probably move it and keep it's old name as well ... although a bit finicky.
3. `Params.headers.Cookie` to overwrite any of the above (so no change I think)
4. `Params.Cookies` should *not* be supported, at least for now. (look at how the merging works for HTTP and you will see it as PITA, and likely is ... completely not useful as I will try to explain below.)

Reasons:
1. I would expect that if I do an HTTP request and I then make a websocket one they will use the same jars. I would argue this will remove the reason for most of the people requesting this, after all the reports are about authenticating, usually with a session which likely got received through and HTTP request.
2. Having a way to specify another jar lets them set cookies in a more uniform way even if they don't use HTTP request to get the cookie
3. The header Cookie and how the order in it works is not very well standardized, so having us merge stuff into it "somehow" just makes things confusing IMO and requires reading way too much documentation from the user to learn how it works (also we will need to write it). IMO if we have a way to get the cookies for a given URL out of the cookie jar (which we do), have a way to modify those values and then seriliaze them to a Cookie value so people can do stuff like
```
var jar  = ... // cookiejar
var cookies = jar.cookiesForURL(myurl);
cookies[mycoolCookie]="specialValue";
var headers = {Cookie: serializeToCookieValue(cookies)};
```
will remove the whole need for having `Cookie` in the `params`. It might be 2 lines more or something like that, but given how much cleaner it will be and less confusing IMO that should've been what was done instead of the current merging of cookies between params.jar and params.cookies that is happening ...

> I would've preferred if all this could work with http.cookieJar(), since the only reason we need a WS approach is to avoid breaking existing scripts, but hopefully we can unify this when we refactor WS and after the new HTTP API.

I see no reason why not do it now ... or just not do these changes at all for now ;) .

This has been the case forever and even now users can always just set the `Cookie` header. The main benefit of providing a `jar` IMO is that it makes things less confusing when you need to get a cookie from HTTP request and give it to the websocket one (which is still HTTP first) and , secondly, for when a websocket connection set cookies ... which currently needs to be parsed by the user somehow. 

The only useful thing about  `params.Cookie` currently is:
1. it lets you set a Cookie for every redirect after that ... which probably is a bug now that I write it 
2. It lets you to *maybe* overwrite a cookie set in the `jar` given that those two match
3. It lets you merge easier with `jar` cookies
4. It lets you set cookies easier then setting the header.

The problems with those are:
1. this IMO is probably not what anyone expects ?!?
2. it only works if you set the `replace:true` part which makes the syntax horrible and I doubt there are many people who even know about this or use it. (it also means that it will change if the jar suddenly gets the same cookie for some reason)
3. this is probably the most useful one, but even this one can probably be done easier and IMO more consistently with some js code
4. great ... unless look at `2.` at which point it doesn't. 

The whole problem IMO comes from the fact that understanding how the 3 ways interact and interact with redirects makes the whole thing way too confusing to the point that even the (technically small) improvements that it does in the case `2-4` likely get removed by the fact that anyone who needs to set cookies specifically as the usual cookie jar support doesn't work will also hit one of the many problems listed above in some respect.

Issue #1836, opened 2021-02-02
teardown makes the max value of iteration_duration not correct


heww:
## Environment

- k6 version: v0.30.0
- OS and version: Ubuntu 18.04
- Docker version and image, if applicable:

## Expected Behavior

teardown should not affect the max value of the iteration_duration

## Actual Behavior

the max value of the iteration_duration includes the running duration of the teardown

## Steps to Reproduce the Problem

```shell
cat <<EOF | k6 run -u 100 -i 100 -
import { sleep } from 'k6'

export let options = {
    setupTimeout: '70s',
    teardownTimeout: '130s',
    noUsageReport: true,
}

export function setup() {
    sleep(60)
}

export default function () {
    sleep(10)
}

export function teardown() {
    sleep(120)
}
EOF
```

```shell

          /\      |‾‾| /‾‾/   /‾‾/   
     /\  /  \     |  |/  /   /  /    
    /  \/    \    |     (   /   ‾‾\  
   /          \   |  |\  \ |  (‾)  | 
  / __________ \  |__| \__\ \_____/ .io

  execution: local
     script: -
     output: -

  scenarios: (100.00%) 1 scenario, 100 max VUs, 10m30s max duration (incl. graceful stop):
           * default: 100 iterations shared among 100 VUs (maxDuration: 10m0s, gracefulStop: 30s)


running (03m10.1s), 000/100 VUs, 100 complete and 0 interrupted iterations
default ✓ [======================================] 100 VUs  00m10.0s/10m0s  100/100 shared iters

     █ setup

     █ teardown

     data_received........: 0 B 0 B/s
     data_sent............: 0 B 0 B/s
     iteration_duration...: avg=11.56s min=10s med=10s max=2m0s p(90)=10s p(95)=10s
     iterations...........: 100 0.526129/s
     vus..................: 0   min=0   max=100
     vus_max..............: 100 min=100 max=100
```

The max value of the iteration_duration should be 10s, not 2m.

na--:
Thanks for reporting this, though it's an already known issue, see https://github.com/loadimpact/k6/issues/891, https://github.com/loadimpact/k6/issues/1508, and https://github.com/loadimpact/k6/issues/1321. The linked forum posts in https://github.com/loadimpact/k6-docs/issues/205 offer a workaround, until we get around to actually writing the docs and properly fixing the issue...

[...]

PR #2057, opened 2021-06-10
Use a single segmented index for local and global iterations

MStoykov:


lib/executor/base_executor.go
```diff
@@ -37,33 +37,37 @@ import (
...
+	executionState    *lib.ExecutionState
+	VUIDLocal         *uint64 // counter for assigning executor-specific VU IDs
+	iterationSegIndex *lib.SegmentedIndex
+	logger            *logrus.Entry
+	progress          *pb.ProgressBar
 }
 
 // NewBaseExecutor returns an initialized BaseExecutor
 func NewBaseExecutor(config lib.ExecutorConfig, es *lib.ExecutionState, logger *logrus.Entry) *BaseExecutor {
 	// Start at max uint64 so that the first iteration can be 0
```
imiric, line 50:
```suggestion
```

lib/executor/base_executor.go
```diff
@@ -37,33 +37,37 @@ import (
...
+		logger:            logger,
+		iterationSegIndex: segIdx,
 		progress: pb.New(
 			pb.WithLeft(config.GetName),
 			pb.WithLogger(logger),
 		),
 	}
 }
 
+func (bs *BaseExecutor) NextIterationCounters() (uint64, uint64) {
```
imiric, line 66: The linter caught it, but let's document what each return value means.

Also, this can be unexported?

lib/executor/ramping_arrival_rate.go
```diff
@@ -189,21 +188,11 @@ func (varr *RampingArrivalRate) Init(ctx context.Context) error {
 	et, err := varr.BaseExecutor.executionState.ExecutionTuple.GetNewExecutionTupleFromValue(varr.config.MaxVUs.Int64)
 	varr.et = et
 	start, offsets, lcd := et.GetStripedOffsets()
-	varr.segIdx = lib.NewSegmentedIndex(start, lcd, offsets)
+	varr.iterationSegIndex = lib.NewSegmentedIndex(start, lcd, offsets)
```
imiric, line 191: Hhmm do we still need to redefine it here? Couldn't we use the instance created in `NewBaseExecutor()`, in which case we can get rid of this `Init()` entirely?

imiric (commented):
Thanks @MStoykov, this is indeed simpler! :clap:

I thought about using a single function for both iterations, but didn't consider using `SegmentedIndex` for it. I suppose that has a slight overhead since before it was a single `*uint64` and `atomic` and now it's more complex, but since we're doing it for all executors it shouldn't matter.

Just left some minor comments, but LGTM. :+1:

lib/executor/ramping_arrival_rate.go
MStoykov: I do think it is needed as this is based on the execution tuple created from the MaxVUs not just based on the execution segment alone.

If I remember correctly this meant that for some corner cases there will be ... wrong values returned otherwise:
execution segment 0:1/2,1/2:1 with 3 VUs will mean that 1 of the instances will have 2 VUs and there other 1 VU so if we are doing arrival rate of `3` the iterations should be 
```
1st instance: 1, 3, 4,6, ... 
2nd instance: 2,5,8 ... 
```
but if you just base it on the original execution segments they will be 
```
1st instance: 1,3,5,7 ...
2nd instance: 2,4,6,8 ...
```
It will still do the same amount of iteration just in the case that there needed to be 12(divides by 2 and 3 ;) ) iterations at some point there will be ... holes:
```
1st instance (has to do 2/3 of 12 so 8): 1,3,5,7,9,11,13,15
2nd instance (has to do 1/3 of 12 so 4):,2,4,6,8
```

P.s. I wrote this quickly so ... likely some of my math is slightly wrong and I also index from 1 ;)

na--: Yeah, because of differences like that, I'd prefer if we don't have `NextIterationCounters()` in `BaseExecutor` and instead have it as a helper. Maybe something like `GetISegmentedCounterIterator(start, lcd, offsets)` that returns essentially the `next()` closure?

na-- (commented):
This LGTM in principle, though as I said in the inline comment, I think we shouldn't add to `BaseExecutor`. I also haven't looked too closely at the details, we can fix any of my nitpicks in https://github.com/k6io/k6/pull/1863, since there we can see the full diff with `master`.

js/init_and_modules_test.go
```diff
@@ -28,16 +28,16 @@ import (
...
 	"go.k6.io/k6/js/common"
 	"go.k6.io/k6/js/modules"
 	"go.k6.io/k6/lib"
 	"go.k6.io/k6/lib/testutils"
 	"go.k6.io/k6/loader"
 	"go.k6.io/k6/stats"
-	"github.com/spf13/afero"
-	"github.com/stretchr/testify/assert"
-	"github.com/stretchr/testify/require"
 	"gopkg.in/guregu/null.v3"
```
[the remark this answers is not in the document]
MStoykov: no I haven't ... and I am against it as the whole idea of gofmt (and goimports respectively) is for them to work the same for everybody and to preferably have 0 knobs ... 

Even if I decide to go through the trouble of doing this is everyone who contributes will likely not do it so we will need to ask them to ... fix it, which IMO is a waste of everybodies time, as again you probably shouldn't go look at 20 lines of imports and care how they are ordered. 

Don't get me wrong if goimprots decides to default to first putting the names of the module you are in :+1: , but me telling it that it should do it is a totally different thing.

lib/executor/base_executor.go
MStoykov: resolved in cfd3f59

lib/executor/ramping_arrival_rate.go
na--: I think we should probably have 2 different types - a simple `SegmentedIndex` without any locking or synchronization, and a `SynchronizedSegmentedIndex` (or some better name...) that has locking and atomic operations.

MStoykov: I don't think a second type is needed, as we have 1 place we use it and it's only for 1 method it is literally going to be 
```
func nextIteartionsCounters() (uint64, uint64){
   someMutex.Lock()
   defer someMutex.Unlock()
   segmentedIndex.next()
   return segmentedIndex.scaled, segmentedIndex.unscaled
}
```

I see no reason to make a whole new type for that .. .but it definitely is better so I am fine with that as well :)

na--: btw I had this pending as a comment in the main PR, but probably worth mentioning here: `SegmentedIndexResult` seems unnecessary. We are never going to have more than the scaled and unscaled indexes in this type, so a tuple (i.e. 2 return values) seems a bit more idiomatic in Go. 

If we ever add more things on top of `SegmentedIndex`, we are likely to do it as another wrapper, not by changing the original (synchronized or unsynchromized), so we can add a struct then, if we need to.

js/init_and_modules_test.go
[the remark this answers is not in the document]
imiric: > I am against it as the whole idea of gofmt (and goimports respectively) is for them to work the same for everybody and to preferably have 0 knobs

Yeah, but they don't :smile:

I don't think we should insist on this for external contributors, but we at least can have that configured so that whenever we change a file it's updated. It's a minor thing, but increases readability of imports substantially IMO.

js/runner.go
```diff
@@ -782,24 +770,14 @@ func (u *ActiveVU) incrIteration() {
...
 		u.scenarioIter[u.scenarioName]++
 	} else {
 		u.scenarioIter[u.scenarioName] = 0
 	}
-	if u.getNextScLocalIter != nil {
-		u.scIterLocal = u.getNextScLocalIter()
-	}
-	if u.getNextScGlobalIter != nil {
-		u.scIterGlobal = u.getNextScGlobalIter()
+	// TODO remove this
```
imiric, line 778: I don't think comments like this are helpful without indicating _why_. ;)

lib/executor/base_executor.go
imiric: I'd expand more on that, but I can change it in #1836.

js/init_and_modules_test.go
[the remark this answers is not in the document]
MStoykov: > I don't think we should insist on this for external contributors, but we at least can have that configured so that whenever we change a file it's updated. 

And apparently, if it's not configured and someone contributes it will move them ... so this IMO should just not be a thing

imiric (approved):
LGTM :clap:

PR #2058, opened 2021-06-10
Added GetName method for metrics

### LINKED ISSUE #2053: Add Custom metrics getName() method
## Feature Description

Currently, the custom metrics objects (like Counter) cannot return its own name.

It can be very useful to have it though. The reason is that I keep track of metrics in Go extensions as well, and return all of them in a map. I need to know the metric’s name to know which goes with which.

## Suggested Solution (optional)

Adding a `getName()` method to each [Metric types](https://k6.io/docs/using-k6/metrics/#metric-types). Note: They currently only have one method each, which is `add(<value>)`, but they each have the `name` parameter set in their constructor.

olimpias:
Closes #2053

### DIFF
--- a/samples/custom_metrics.js
+++ b/samples/custom_metrics.js
@@ -26,16 +26,24 @@ export default function () {
     let res = http.get("http://httpbin.org/");
     let passed = check(res, { "status is 200": (r) => r.status === 200 });
 
+
     // Add one for number of requests
     myCounter.add(1);
 
-    // Set max response time seen
+	console.log(myCounter.getName(), " is config ready")
+
+	// Set max response time seen
     maxResponseTime = Math.max(maxResponseTime, res.timings.duration);
     myGauge.add(maxResponseTime);
+
+	console.log(myGauge.getName(), " is config ready")
 
     // Add check success or failure to keep track of rate
     myRate.add(passed);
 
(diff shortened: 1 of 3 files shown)


samples/custom_metrics.js
```diff
@@ -26,16 +26,24 @@ export default function () {
     let res = http.get("http://httpbin.org/");
     let passed = check(res, { "status is 200": (r) => r.status === 200 });
 
+
     // Add one for number of requests
     myCounter.add(1);
 
-    // Set max response time seen
+	console.log(myCounter.getName(), " is config ready")
```
na--, line 33: The indentation looks wrong here, uses a tab instead of the spaces used in the rest of the file. Besides, we probably don't need to demonstrate that every custom metric has this new method, a single example will probably be enough.

na-- (commented):
Thanks, mostly LGTM, except the wrong indentation and too many examples I noted inline.

na-- (dismissed):
LGTM, thanks! Forgot to enable running the CI, will do so now

na-- (approved):

MStoykov (approved):

imiric (approved):
Thanks!

olimpias:
Thank you for the reviews 🙌

PR #2060, opened 2021-06-11
Fix array encoding in form data requests

### LINKED ISSUE #1713: samples html_form_post.js fails
K6 http.post method isn't correctly encoding a list of values assigned to a key in an object.

```javascript
const form_data = {
    comment: "Hello world!",
    topping: [
        'onion',
        'bacon',
        'cheese'
    ]
};
```
The simplest way to see this is to run the sample at 
https://github.com/loadimpact/k6/blob/master/samples/html_form_post.js and note that it fails.

The list of toppings should be passed as multiple values for the same key, but it is instead having the values stringified and passed as one value.

With this issue, it's difficult to pass multiple select check boxes with the same name or multi-selections in a select.

## Environment
- k6 version: k6 v0.28.0 (2020-09-24T14:33:59+0000/v0.28.0-0-gdee9c4ce, go1.14.9, linux/amd64)
- OS and version: Both windows 10 and Linux K6
- Docker version and image, if applicable:

## Expected Behavior

The topping values should be encoded as
topping=onion&topping=bacon&topping=cheese
and the sample should pass all checks.

## Actual Behavior

Test case fails because the topping values are encoded as (and URI encoded)
topping=[onion bacon cheese]

1. Copy sample from https://github.com/loadimpact/k6/blob/master/samples/html_form_post.js locally or clone the git repository
2. k6 run samples/html_form_post.js
3. Observe test failure outcome

imiric:
This is a continuation of #1754, with an added check for the correct encoding and disabling of the `exhaustive` linter, since it doesn't seem like we can avoid the reflection.

Closes #1713

### DIFF
--- a/js/modules/k6/http/request.go
+++ b/js/modules/k6/http/request.go
@@ -177,7 +177,7 @@ func (h *HTTP) parseRequest(
 		if !requestContainsFile(data) {
 			bodyQuery := make(url.Values, len(data))
 			for k, v := range data {
-				switch reflect.TypeOf(v).Kind() {
+				switch reflect.TypeOf(v).Kind() { //nolint:exhaustive
 				// handle json arrays in params
 				case reflect.Slice, reflect.Array:
 					s := reflect.ValueOf(v)
--- a/js/modules/k6/http/request.go
+++ b/js/modules/k6/http/request.go
@@ -177,16 +177,13 @@ func (h *HTTP) parseRequest(
 		if !requestContainsFile(data) {
 			bodyQuery := make(url.Values, len(data))
 			for k, v := range data {
-				switch reflect.TypeOf(v).Kind() { //nolint:exhaustive
-				// handle json arrays in params
-				case reflect.Slice, reflect.Array:
-					s := reflect.ValueOf(v)
-					for i := 0; i < s.Len(); i++ {
-						bodyQuery.Add(k, formatFormVal(s.Index(i)))
+				if arr, ok := v.([]interface{}); ok {
+					for _, el := range arr {
+						bodyQuery.Add(k, formatFormVal(el))
 					}
-				default:
-					bodyQuery.Set(k, formatFormVal(v))
+					continue
 				}
+				bodyQuery.Set(k, formatFormVal(v))
 			}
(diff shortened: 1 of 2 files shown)


na-- (dismissed):

js/modules/k6/http/request.go
```diff
@@ -177,7 +177,16 @@ func (h *HTTP) parseRequest(
...
 			bodyQuery := make(url.Values, len(data))
 			for k, v := range data {
-				bodyQuery.Set(k, formatFormVal(v))
+				switch reflect.TypeOf(v).Kind() { //nolint:exhaustive
+				// handle json arrays in params
+				case reflect.Slice, reflect.Array:
+					s := reflect.ValueOf(v)
+					for i := 0; i < s.Len(); i++ {
+						bodyQuery.Add(k, formatFormVal(s.Index(i)))
+					}
```
MStoykov, line 186: Using a simple type switch and matching just `[]interface{}` works just as well. I have no idea how to make a js code that will give me a go array and apparently the test below doesn't do it ... so I am kind of for ditching the `reflect` ;)

na--: that was my objection in the original PR, but then I got this response and didn't verify for myself: https://github.com/k6io/k6/pull/1754#issuecomment-738773727

MStoykov: Well I did write the code .. and we do assert to `[]interface{}` in the batch parsing code:
https://github.com/k6io/k6/blob/b9a29eb7fc75b5d903f0c9e3a307ed80d06dacc5/js/modules/k6/http/request.go#L468-L474

imiric: Yeah, I assumed that comment to be true and didn't try a type assertion. Seems to work fine, see 0d48122b.

na-- (approved):

MStoykov (approved):
LGTM!

I would've used a `range` as it seems more idiomatic to me but this is fine as well

MStoykov (approved):

imiric:
Apparently 2 approvals are required if a previous approval was dismissed? :-/

na-- (approved):

PR #2062, opened 2021-06-14
Tc39 tests "fixes"

MStoykov:


MStoykov:
please read commit by commit, both changes were provoked mostly by my testing of goja's restructuring branch and testing a PR for goja which leads to a lot of changes in the json file .. which hopefully this two changes will bring down.

imiric (approved):
I'm not very familiar with these tests, but the code and your rationale LGTM.

na-- (approved):
LGTM. Losing the line numbers might make the diagnosing of some errors trickier, but we rarely try to diagnose specific errors here, and we can always display the line numbers if we wanted to, so I don't mind :man_shrugging:

MStoykov:
> Losing the line numbers might make the diagnosing of some errors trickier, but we rarely try to diagnose specific errors here, and we can always display the line numbers if we wanted to, so I don't mind man_shrugging

as the one person who does this and in general, does the updates, it, in general, involves cleaning JSON with the errors and running tests one by one, and routinely editing and running the script with `k6` directly just to be able to find out a bit more about what's going on, so the 1 line comment that I might need to add so that I can see the (sometimes wrong) line number is truly the smallest problem :rofl: , also in general it's obvious from the error message and the 10 lines long scripts so yeah :D. 

But it by far makes the diffs of the breaking error json a lot more readable

PR #2064, opened 2021-06-15
Add global VU IDs, expose in k6/execution

imiric:
I did this in a bit of a hurry, but from what I understood from [this discussion](https://github.com/k6io/k6/pull/1863#discussion_r651586477) we don't need to take into account `MaxVUs`, so it was relatively straightforward. I spent more time on the test and trying to fix a data race that's still an issue in #1863.

Resolves https://github.com/k6io/k6/pull/1863#discussion_r650981134


lib/executor/execution_test.go
```diff
@@ -39,24 +40,62 @@ import (
...
+			assert.Equal(t, uint64(expGlobal), idg)
+
+			idl, idg = es.GetUniqueVUIdentifiers()
+			assert.Equal(t, uint64(2), idl)
+			expGlobal += offsets[0]
+			assert.Equal(t, uint64(expGlobal), idg)
+
+			idl, idg = es.GetUniqueVUIdentifiers()
+			assert.Equal(t, uint64(3), idl)
+			expGlobal += offsets[0]
```
MStoykov, line 78: Shouldn't this be `offsets[1]` ? 

Also I feel like this is basically testing that it uses execution segments underneed, which doesn't seem that useful.

A concrete example with concrete numbers at least means that it requires that "given this segmented execution parameters we will be getting this VU ids". In the case of executors this was mostly done because 
1. there is a lot more logic *based* on the execution segments
2. sometimes some of the numbers are *not* based on the original execution segments provided as configuration, but based on something on top of them for the concrete scenario and executor. 

What I am trying to say is that this IMO doesn't test anything new, apart from "GetUniqueVUIdentifiers uses segemetned index underneed" which seems like an implementation detail to me(albeit the one that makes this possible at all)

MStoykov (approved):
LGTM, apart from the comment

lib/executor/execution_test.go
imiric: `offsets` has a single element for these execution segments, and I picked them more or less deliberately so that I wouldn't have to cycle through multiple offsets :sweat_smile:

You're right that this is more of a white-box test now, but we have plenty of those around... Arguably the previous test was also white-box, it's just that the implementation was to always increment by 1, so it was simpler to check.

I wanted to keep this same test style so that I wouldn't have to build and assert against `expectedIDs`, as it would be unwieldy, and I like the current random approach.

How would you rewrite it or test this differently?

na-- (approved):

lib/executor/execution_test.go
MStoykov: 1. I probably will just *not* test it.
2.  Or just expect that the IDs will go up
2+. Check that with all the different segments of a given sequence I will *always* get different global VU ids between the segments. Possibly that there are no holes as well

`2.` in my opinion test the important property of the whole thing which is that we will have *different* global VU IDs between the instances, not what their exact numbers will be.

MStoykov: *but* IMO we can do this later as we need to finish up the main PR so ... merge this as is :P

PR #2071, opened 2021-06-23
Add metrics registry

### LINKED ISSUE #1832: Common metrics registry
Currently, k6 [metrics](https://k6.io/docs/using-k6/metrics) (both built-in ones and [custom ones](https://k6.io/docs/using-k6/metrics#custom-metrics)), as well as any measurements that are made for them, are very decoupled from the places that consume them in the [`Engine`](https://github.com/loadimpact/k6/blob/46d53d99deba98add23079df93579408fa7ea094/core/engine.go#L250) (i.e. [thresholds](https://k6.io/docs/using-k6/thresholds) and the [end-of-test summary and outputs](https://k6.io/docs/getting-started/results-output)). At a first glance, this might seem like a good architectural choice, but the reality is that instead of loosely coupled, they are very tightly coupled at a distance. This brings far more issues than it solves, if it even solves any... :disappointed:

With the current architecture, k6 doesn't "know" all of the built-in metrics that actually exist. They are defined as global variables in [`lib/metrics/metrics.go`](https://github.com/loadimpact/k6/blob/46d53d99deba98add23079df93579408fa7ea094/lib/metrics/metrics.go#L27-L67), but they are unconnected to anything else. And while there's a separate issue just to refactor that ugliness away (https://github.com/loadimpact/k6/issues/572), not knowing the set of built-in metrics when the [init context](https://k6.io/docs/using-k6/test-life-cycle) is executed and [custom metrics](https://k6.io/docs/javascript-api/k6-metrics) can be defined makes it possible for users to define custom metrics with the same names as the built-in ones, but with different [metric and value types](https://github.com/loadimpact/k6/blob/46d53d99deba98add23079df93579408fa7ea094/stats/stats.go#L48-L61).

Compounding the issue, custom metrics themselves are quite problematic. There isn't absolutely any validation (https://github.com/loadimpact/k6/issues/1435), so custom metrics from different VUs (or even the same VU) can redefine both built-in metrics and other custom metrics with different metric and value types! Currently, this is not going to throw an exception in k6:
```js
import { Rate, Trend, Gauge } from 'k6/metrics'

let metric1 = new Rate('same_metric_name');
let metric2 = new Trend('same_metric_name', true);
let metric3 = new Trend('same_metric_name', false);
let metric4 = new Gauge('http_req_duration'); // same name as built-in metric, but completely different type! 
```

The [metrics constructor](https://github.com/loadimpact/k6/blob/46d53d99deba98add23079df93579408fa7ea094/js/modules/k6/metrics/metrics.go#L57-L74) doesn't really do validation or create anything in a centralized place. It just returns something like a [metric "template"](https://github.com/loadimpact/k6/blob/46d53d99deba98add23079df93579408fa7ea094/js/modules/k6/metrics/metrics.go#L50-L52) that is [attached to every metric `Sample`](https://github.com/loadimpact/k6/blob/46d53d99deba98add23079df93579408fa7ea094/js/modules/k6/metrics/metrics.go#L94-L95) later generated by the JS `.add()` method. So we don't really know the custom metrics and their types when the `Engine` starts running the thresholds. It only realizes a metric exists when it "sees" a sample from it for the first time, and copies whatever "template" it first sees for the metric name :sob: https://github.com/loadimpact/k6/blob/46d53d99deba98add23079df93579408fa7ea094/core/engine.go#L406-L413

So, this will not only *not* produce an error:
```js
import http from 'k6/http';
import { Gauge } from 'k6/metrics'

let blah = new Gauge('http_req_duration');

export default function () {
    blah.add(100); // mess up everything
    http.get("https://httpbin.test.k6.io/");
    http.get("https://httpbin.test.k6.io/");
}
```
It will produce messed up summary like this:
```
...
     http_req_connecting........: avg=70ms     min=0s       med=70ms     max=140.01ms p(90)=126.01ms p(95)=133.01ms
     http_req_duration..........: 204.468661 min=100 max=410.000517
     http_req_receiving.........: avg=213.77µs min=196.04µs med=213.77µs max=231.49µs p(90)=227.95µs p(95)=229.72µs
...
```

The current architecture is blocking all of these issues and probably others as well:
- Custom metric validation - https://github.com/loadimpact/k6/issues/1435
- Metric and sub-metric blocking and tracking - https://github.com/loadimpact/k6/issues/1321
- Validation of thresholds - https://github.com/loadimpact/k6/issues/1443 (specifically the "aggregation method validation" section) and https://github.com/loadimpact/k6/issues/961#issuecomment-557482834
- Nasty bug where thresholds are not triggered if a metric hasn't been emitted - https://github.com/loadimpact/k6/issues/1346
- Idea for immutable tag sets - https://github.com/loadimpact/k6/issues/1831

I've opened this issue as the first prerequisite step for solving all of the ones listed above. I think the way to do that would be to have something like a single central "metrics registry" in k6. A list (or, rather, a map) of metric names with their types - both the built-in ones and all custom ones.

Any attempts at re-defining built-in or previously defined custom metrics should check the metric and value types of the new metric. If they don't match the already existing metric, an exception should be raised. And if they do match, instead of creating a new object, we should just return a `*Metric` pointer to the already existing metric object. To reduce undefined behavior even further, and to save us from having to ensure thread-safety, we should probably restrict the creation of custom metrics only to the [initial init context](https://github.com/loadimpact/k6/blob/46d53d99deba98add23079df93579408fa7ea094/cmd/run.go#L124) (from which we get the exported script `options`).

MStoykov:
Add a basic metric registry and define metrics as builtin for the registry instead of having globals.

This also fixes problems around custom metrics "redefining" metric with the same name but with different type.
fixes #1832


cmd/run.go
```diff
@@ -197,6 +196,16 @@ a commandline interface for interacting with it.`,
...
 
+			registry := stats.NewRegistry(engine.Samples)
+			builtInMetrics := metrics.RegisterBuiltinMetrics(registry)
+			globalCtx = stats.WithRegistry(globalCtx, registry)
+			globalCtx = metrics.WithBuiltinMetrics(globalCtx, builtInMetrics)
+
+			lingerCtx, lingerCancel := context.WithCancel(globalCtx)
+			defer lingerCancel()
+			runCtx, runCancel := context.WithCancel(lingerCtx)
+			defer runCancel()
```
MStoykov, line 207: This hacks were needed as I need to have the builtinMetrics in both `setup`, `teardown`, `handleSummary`  and all the outputs, so it needs to be as early as possible, but I also need the engine's Samples as I didn't want to move that as well

cmd/run.go
```diff
@@ -442,3 +451,53 @@ func handleSummaryResult(fs afero.Fs, stdOut, stdErr io.Writer, result map[strin
 
 	return consolidateErrorMessage(errs, "Could not save some summary information:")
 }
+
+// StartOutputs spins up all configured outputs, giving the thresholds to any
+// that can accept them. And if some output fails, stop the already started
+// ones. This may take some time, since some outputs make initial network
+// requests to set up whatever remote services are going to listen to them.
+func StartOutputs(
```
MStoykov, line 459: I literally otherwise needed to thread too much stuff through the engine, and decided that moving it out will be much easier.

THis probably should be method on some object having `outputs` and `logger` as fields ... but I didn't want to make additional changes (this will be a theme in this comments and the whole WIP PR :P )

js/modules/k6/k6_test.go
```diff
@@ -136,32 +138,42 @@ func TestGroup(t *testing.T) {
 	})
 }
 
-func TestCheck(t *testing.T) {
+func checkTestRuntime(t testing.TB, ctxs ...*context.Context) (
```
MStoykov, line 141: Like 90% of the changes here are so the test is better structured .... I probably can pull some of those out in a separate commit, if we aren't going to go with this design or if the time allows

js/runner.go
```diff
@@ -756,8 +757,50 @@ func (u *VU) runFn(
 	if u.Runner.Bundle.Options.NoVUConnectionReuse.Bool {
 		u.Transport.CloseIdleConnections()
 	}
+	// TODO move this to a function or something
+	builtinMetrics := metrics.GetBuiltInMetrics(ctx)
```
MStoykov, line 761: This also was the easier way then to just keep adding to the Trail's "constuctor"

stats/registry.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+	return m
+}
+
+// PushContainer emits a contained through the registry's sink
+// TODO drop this somehow
+func (r *Registry) PushContainer(ctx context.Context, container SampleContainer) {
+	PushIfNotDone(ctx, r.sink, container)
+}
+
```
MStoykov, line 78: A lot of the actual design problems come from the fact we have containers and even more because of custom containers such as HTTPTrail.

I don't think that just removing the later is feasible at all without so much rework that it's totally a different PR, even if we wanted to, so some workaround will need to be found. 

The predominant problem is that while I can make certain that the only way the type Sample can be emitted is through the registry, by just not having a different way to emit a metric through the registry but through the metric it generates (see Sample's Push method which, probably should be an Emit on the Metric :thinking: ), this more or less doesn't work with SampleContainers and especially with HTTPTrail as that one is specifically useful in the cloud output, where instead of looking at it's Samples we just look at it ... properly typed fields. 

I do think we should postpone fixing that for a later PR,hopefully still in the v0.33.0 cycle. I have hopes that if we introduce separate Metric/Sample types for Gauge, Counter, Rate and Trend that will make some this problems go away :shrug:

MStoykov (commented):
The main thing I want to know from this PR is are we fine with how we need to now thread builintMetrics through the code. I can revert the registry changes if we don't think we want anything like that or if we want a more full PR, but the builtinMetrics threading took a lot of time and fixes and will need at least a bit more test fixes which will make it span more or less the whole codebase, so I prefer if I do it early and as a single PR.

lib/executor/ramping_arrival_rate.go
```diff
@@ -479,10 +480,7 @@ func (varr RampingArrivalRate) Run(parentCtx context.Context, out chan<- stats.S
 		// Since there aren't any free VUs available, consider this iteration
 		// dropped - we aren't going to try to recover it, but
 
-		stats.PushIfNotDone(parentCtx, out, stats.Sample{
-			Value: 1, Metric: metrics.DroppedIterations,
-			Tags: metricTags, Time: time.Now(),
-		})
+		droppedIterationMetric.Emit(parentCtx, time.Now(), metricTags, 1)
```
MStoykov, line 483: This seems a lot better to me than the above or the Sample.Push I show in the `constant-arrival-rate`
https://github.com/k6io/k6/pull/2071/files#diff-2b3a85b9dccbd5050f5755479947c5237340bf7335dba7f83d654c55548df710R349-R352

core/engine.go
```diff
@@ -147,6 +148,10 @@ func (e *Engine) StartOutputs() error {
 				})
 		}
 
+		if builtinMetricOut, ok := out.(output.WithBuiltinMetrics); ok {
+			builtinMetricOut.SetBuiltinMetrics(e.builtInMetrics)
```
na--, line 152: This seems unnecessary - you already do it in `StartOutputs()` in `cmd/run.go`

js/modules/k6/k6_test.go
na--: separate PR/commit sounds good :+1:

js/runner.go
na--: As I mentioned in another comment, it should be relatively easy to thread the metric registry and the built-in metrics struct to the VU `State`, to which you already have direct access here.

stats/registry.go
```diff
@@ -0,0 +1,91 @@
...
+	"context"
+	"fmt"
+	"sync"
+)
+
+// Registry is what can create metrics
+// TODO maybe rename the whole package to metrics so we have `metrics.Registry` ?
+type Registry struct {
+	metrics map[string]*Metric
+	sink    chan<- SampleContainer
```
na--, line 33: I think the registry shouldn't have this `sink`. It's not a clear separation of concerns - the registry should be concerned with registering metrics and potentially time series later on. The actual processing and pushing of metric samples around should be handled outside of it.

stats/stats.go
```diff
@@ -334,6 +334,10 @@ type Sample struct {
 	Value  float64
 }
 
+func (s Sample) Push(ctx context.Context) {
```
na--, line 337: This also seems bad - too much mixing of concerns. I'd rather not have self-propelled metric samples :sweat_smile: They should be simple data points and something else should push them around.

na-- (changes requested):
In general, I like the metrics registry and the `BuiltInMetrics struct` approach, but I am not a huge fan of how they are passed around via context (I don't think we need it, we can simply pass them to where they are required) and I don't like the self-propelled metric samples.

lib/metrics/metrics.go
```diff
@@ -21,48 +21,133 @@
...
 
 import (
+	"context"
+
 	"go.k6.io/k6/stats"
 )
 
-// TODO: refactor this, using non thread-safe global variables seems like a bad idea for various reasons...
+const (
+	VUsName               = "vus"
```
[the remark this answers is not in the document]
MStoykov: This for example means that the cloud output will have to have access to the registry in order to make half the changes that are currently there which made .... a lot more code changes that were only needed for this specific case so I do prefer the constants to that.

stats/registry.go
MStoykov: How do you prevent some code to just make a new registry and register a metric in it and then emit it?

cmd/run.go
MStoykov: for starters ... lib.State is not around in the init context

core/engine.go
MStoykov: oops this whole function should be dropped ;)

stats/registry.go
na--: Hmm, I see your point, but at the same time, I am not sure what you describe is actually a problem we should guard against :sweat_smile: A simple metric registry, even when it doesn't have a sink, would be enough to guard against scripts that register conflicting custom metrics from JS, either between themselves or with built-in ones, which is what we want. We control the `k6/metrics` API and it will only ever use the single registry we initialize.

I'd say that guarding against malicious xk6 extensions sending garbage is out of scope for this PR. They can run arbitrary Go code and have a 1001 other ways of messing everything, including way more important things than metrics. Even in this PR, they can just send whatever they want directly to the output, or over https://github.com/k6io/k6/blob/7da10ac81128940e0828d9aca2e955f4e046c742/lib/state.go#L64

cmd/run.go
na--: Fair enough, but you have this: https://github.com/k6io/k6/blob/7da10ac81128940e0828d9aca2e955f4e046c742/js/common/initenv.go#L31-L33

It's actually even better, since that way you can ensure that metrics can be registered only in the init env :man_shrugging:

lib/metrics/metrics.go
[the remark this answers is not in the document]
na--: > This for example means that the cloud output will have to have access to the registry

Why? Wouldn't it just need access to the `BuiltInMetrics`, which it can easily get through https://github.com/k6io/k6/blob/7da10ac81128940e0828d9aca2e955f4e046c742/output/types.go#L100-L104

js/modules/k6/k6_test.go
MStoykov: https://github.com/k6io/k6/pull/2094

lib/executor/ramping_arrival_rate.go
na--: Yeah, I saw this, and it's slightly better than `Sample.Push()`, but I still think we're not separating concerns properly

lib/metrics/metrics.go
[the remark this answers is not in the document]
na--: > no good reason to not have the names of the metrics as constants

Not sure if it's a good reason, but my reason is that, when we have this metrics registry, there _shouldn't_ really be a need to compare metric names. And the fact that there is such a need is a bit of a code smell, something we probably need to refactor further in the future, potentially in another PR.

When every metric has a single object, you should be able to just compare these object pointers to each other, or use them as map indexes, and that would be enough (and more efficient than using strings). IMO metric names are an implementation detail, the majority of the k6 code shouldn't know them or care about them. The cloud output is a bit different, for now, because it repackages the samples in `http_req_li_all` and `http_iter_li_all`, but it should still be easily implementable by passing it the `BuiltInMetrics`.

What are the other places where metric names are used? As far as I can see, they are only required in the cloud output:
```
(git)-[metricsRegistry] % rg 'metrics\.[\w]+Name'
output/cloud/output_test.go
232:		Metric: metrics.VUsName,
650:		Metric: metrics.HTTPReqDurationName,

output/cloud/output.go
380:				metrics.DataSentName:     float64(sc.BytesWritten),
381:				metrics.DataReceivedName: float64(sc.BytesRead),
385:				values[metrics.IterationDurationName] = stats.D(sc.EndTime.Sub(sc.StartTime))
386:				values[metrics.IterationsName] = 1

output/cloud/data.go
119:	values[metrics.HTTPReqsName] = 1
120:	values[metrics.HTTPReqDurationName] = stats.D(trail.Duration)
121:	values[metrics.HTTPReqBlockedName] = stats.D(trail.Blocked)
122:	values[metrics.HTTPReqConnectingName] = stats.D(trail.Connecting)
123:	values[metrics.HTTPReqTLSHandshakingName] = stats.D(trail.TLSHandshaking)
124:	values[metrics.HTTPReqSendingName] = stats.D(trail.Sending)
125:	values[metrics.HTTPReqWaitingName] = stats.D(trail.Waiting)
126:	values[metrics.HTTPReqReceivingName] = stats.D(trail.Receiving)
128:		values[metrics.HTTPReqFailedName] = stats.B(trail.Failed.Bool)

output/cloud/data_test.go
53:				Metric: metrics.VUsName,
71:						metrics.DataSentName:          1234.5,
72:						metrics.DataReceivedName:      6789.1,
73:						metrics.IterationDurationName: stats.D(10 * time.Second),
```

imiric (commented):
Don't have much to add besides what @na-- mentioned:

- Avoid embedding in the context if possible.
- `Sample.Push()` and `Registry.PushContainer()` feel slightly out of place. The registry should handle metric registration only and metric pushing should be done separately IMO.
- Having another method for outputs to implement (`SetBuiltinMetrics`) seems unnecessary since most outputs shouldn't care about this, and if needed they should be able to retrieve it with e.g. `metrics.GetBuiltInMetrics()` (without a context). But I think [you're removing this](https://github.com/k6io/k6/pull/2071/files#r668700631), so :thumbsup:.
- Renaming the `stats` package to `metrics` to have `metrics.Registry` also SGTM.

cmd/run.go
MStoykov: I will drop this and make a separate PR if we want to move them out? 
WDYT @na-- @imiric @codebien

MStoykov (commented):
I really prefer to keep the metric names as constants, I am pretty sure I have a builtinMetrics in the vicinity for most of them(especially in the tests) but I still feel that having a constant for what is effectively a constant is probably much better than not having one.

I can now move `Registry` to `metrics` but renaming `stats` to metrics will be an even bigger change IMO that won't really help anything IMO.

For the record this changes can and probably will be done in following PRs, after all we will need to at least:
1. drop js support for thresholds which I would expect will have other side-effects
2. try to make tags more ... performant which likely will be an even bigger change that will require a lot of moving of stuff around that I could not possibly anticipate currently
3. I do want to have GaugeMetric and TrendMetric and possibly GaugeSample and TrendSample and so on, all of which will probably as the above require *a lot* of changes possibly moving them between packages and adding interfaces and such.
4. emitting only certain metrics and sub metrics will likely require big refactors as well

 which all means that doing more renames will likely just mean we will need to rename them again in the future ;).

MStoykov:
I wonder if `NewRegistry` should not returned a registry with the builtinMetrics already added and as a field on it? Or maybe just have the fields for each one separately :thinking: .

I am not particularly fond of this idea myself as I was wondering if builtinMetrics should not be populated by their respective modules although that will likely have problems with script that first register a metric with the name `http_req_duration` and then import/require it (I think imports will be put on top but I think `require` can be anywhere). :thinking:

output/types.go
```diff
@@ -95,3 +96,9 @@ type WithRunStatusUpdates interface {
 	Output
 	SetRunStatus(latestStatus lib.RunStatus)
 }
+
+// WithBuiltinMetrics means the output can receive the builtin metrics.
+type WithBuiltinMetrics interface {
+	Output
+	SetBuiltinMetrics(builtinMetrics *metrics.BuiltinMetrics)
+}
```
imiric, line 104: I'm not sure this is a pattern we want to use for many things, just like we want to avoid embedding everything in `Context`. In this case since `InitEnvironment` is already embedded in the `Context`, and we could potentially access the built-in metrics from `Registry`, what do you think about changing the `Output` interface and simply passing the context from the `Engine` to each output in `Start(ctx)`?

imiric (commented):
LGTM overall, just some questions and suggestions.

I'm wondering if we can avoid having to initialize `Registry` and `BuiltinMetrics` everywhere and just use the instance from `InitEnvironment`. In that case it would make sense to [embed the builtin metrics on `Registry`](https://github.com/grafana/k6/pull/2071#issuecomment-880603333), and we can access it from where it's needed with `common.GetInitEnv()`. It's still relying on `Context`, but this is one of the few valid use cases IMO.

I would also try to get rid of the `BuiltinMetrics` struct and access them from `Registry.metrics` with `Registry.GetMetric(name string)`, where `name` is one of the constants defined in `metrics.go`. But this might become unwieldy for places like `Trail.SaveSamples`, so maybe not.

stats/stats.go
```diff
@@ -446,6 +450,16 @@ type Metric struct {
 	Sink       Sink         `json:"-"`
 }
 
+// Emit a sample for the metric at the given time, with the provided tags and values
+func (m *Metric) Emit(t time.Time, tags *SampleTags, value float64) Sample {
```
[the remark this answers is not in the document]
MStoykov: Maybe just `Sample` instead of `ToSample` as it is not converting it ;) it's sampling the metric which I would argue is what Emit also means

output/types.go
MStoykov: InitEnvironment is embeded in VUs not in the Engine, so this will be quite a big change and also we do not provide the Output with a context to begin with

MStoykov:
InitEnvironment should only be available in the init scope/context and we emit metrics (what we need builtinMetrics) always in places outside of that so ... we will need to give builtinmetrics access through something else in all cases.

Calling a function to get the metric we want to emit on each emotion even if that is just map access is likely going to be way too slow ... especially for something like emitting metrics.

js/runner.go
```diff
@@ -756,8 +768,49 @@ func (u *VU) runFn(
 	if u.Runner.Bundle.Options.NoVUConnectionReuse.Bool {
 		u.Transport.CloseIdleConnections()
 	}
+	// TODO move this to a function or something
+	bytesWritten, bytesRead := u.Dialer.GetBytes()
```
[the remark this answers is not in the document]
MStoykov: 6c0ee44

lib/executor/constant_vus.go
```diff
@@ -142,7 +143,9 @@ var _ lib.Executor = &ConstantVUs{}
 
 // Run constantly loops through as many iterations as possible on a fixed number
 // of VUs for the specified duration.
-func (clv ConstantVUs) Run(parentCtx context.Context, out chan<- stats.SampleContainer) (err error) {
+func (clv ConstantVUs) Run(
+	parentCtx context.Context, out chan<- stats.SampleContainer, builtinMetrics *metrics.BuiltinMetrics,
```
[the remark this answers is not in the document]
MStoykov: 4b945b5

output/types.go
imiric: Hhmm right, it's not the same context, and the `InitEnvironment` is embedded in `Bundle.instantiate()`, so the `Engine` doesn't have access to it.

> also we do not provide the Output with a context to begin with

I know, I was suggesting we should change `Output.Start()` to take one. That seems cleaner than adding another `With*` method to outputs, but not if it requires a lot of changes. So it's fine as is.

stats/stats.go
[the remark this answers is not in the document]
imiric: To me "emit" means to dispatch, send, etc., which this doesn't do. "Sampling the metric" sounds like a process, so `ToSample` sounds better to me, but `Sample` is fine as well.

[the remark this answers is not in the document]
MStoykov: b8b8f44d

cmd/run.go
```diff
@@ -386,12 +389,13 @@ func runCmdFlagSet() *pflag.FlagSet {
 // Creates a new runner.
 func newRunner(
 	logger *logrus.Logger, src *loader.SourceData, typ string, filesystems map[string]afero.Fs, rtOpts lib.RuntimeOptions,
+	builtinMetrics *metrics.BuiltinMetrics, registry *metrics.Registry,
```
na--, line 392: This is getting a bit crowded :sweat_smile: Not saying we should change it now, just making a note. IIRC we wanted to refactor and rename it anyway, so it's not a big deal, we should just do that sooner.

core/local/local.go
```diff
@@ -337,7 +339,10 @@ func (e *ExecutionScheduler) runExecutor(
 
 // Run the ExecutionScheduler, funneling all generated metric samples through the supplied
 // out channel.
-func (e *ExecutionScheduler) Run(globalCtx, runCtx context.Context, engineOut chan<- stats.SampleContainer) error {
+//nolint:cyclop
+func (e *ExecutionScheduler) Run(
+	globalCtx, runCtx context.Context, engineOut chan<- stats.SampleContainer, builtinMetrics *metrics.BuiltinMetrics,
```
na--, line 344: Why do we pass the `BuiltinMetrics` here, instead of in the `ExecutionScheduler` constructor? :confused: Wouldn't it be clearer to pass it to the `NewExecutor()` constructor, e.g. through the `ExecutionState`? https://github.com/grafana/k6/blob/0628db0553c1938a52fcde01a83326207748f32f/lib/executors.go#L106

output/cloud/data.go
```diff
@@ -116,16 +116,16 @@ func NewSampleFromTrail(trail *httpext.Trail) *Sample {
...
 	values := make(map[string]float64, length)
-	values[metrics.HTTPReqs.Name] = 1
-	values[metrics.HTTPReqDuration.Name] = stats.D(trail.Duration)
-	values[metrics.HTTPReqBlocked.Name] = stats.D(trail.Blocked)
-	values[metrics.HTTPReqConnecting.Name] = stats.D(trail.Connecting)
-	values[metrics.HTTPReqTLSHandshaking.Name] = stats.D(trail.TLSHandshaking)
-	values[metrics.HTTPReqSending.Name] = stats.D(trail.Sending)
-	values[metrics.HTTPReqWaiting.Name] = stats.D(trail.Waiting)
-	values[metrics.HTTPReqReceiving.Name] = stats.D(trail.Receiving)
+	values[metrics.HTTPReqsName] = 1
```
na--, line 119: You can have access to the built-in metrics here (through `WithBuiltinMetrics`), right? So why not use the `Name` properties from them and avoid the constants?

output/cloud/data_test.go
```diff
@@ -67,9 +68,9 @@ func TestSampleMarshaling(t *testing.T) {
 					Time: toMicroSecond(now),
 					Tags: stats.IntoSampleTags(&map[string]string{"test": "mest"}),
 					Values: map[string]float64{
-						metrics.DataSent.Name:          1234.5,
-						metrics.DataReceived.Name:      6789.1,
-						metrics.IterationDuration.Name: stats.D(10 * time.Second),
+						metrics.DataSentName:          1234.5,
```
na--, line 71: Same here, you literally have the `builtinMetrics` struct in the test above

lib/metrics/metrics.go
```diff
@@ -24,45 +24,113 @@ import (
...
-var (
-	// Engine-emitted.
-	VUs               = stats.New("vus", stats.Gauge)
-	VUsMax            = stats.New("vus_max", stats.Gauge)
-	Iterations        = stats.New("iterations", stats.Counter)
-	IterationDuration = stats.New("iteration_duration", stats.Trend, stats.Time)
-	DroppedIterations = stats.New("dropped_iterations", stats.Counter)
-	Errors            = stats.New("errors", stats.Counter)
+const (
+	VUsName               = "vus" //nolint:golint
```
na--, line 28: I think I've mentioned this before, but I don't see a good reason for these to be public, or even consts. They can be simple strings in `RegisterBuiltinMetrics()`, because they shouldn't have to be directly referenced anywhere... They should only be used once, in the definition, and never needed as strings anywhere else again.

na-- (approved):
Mostly LGTM beyond some minor nitpicks I mentioned inline. Also, please fix your PR description, the current one looks like it was a leftover from the WIP days of the pull request :sweat_smile:

output/cloud/data.go
MStoykov: Because unfortunately in some places I don't have a context in this tests, so after I tried for an hour to add it everywhere and failed I decided there is no point int making it harder for myself. There are way too many places where using builtingMetrics in the tests is way too hard for the (IMO) no benefit of not having the names as consts that I prefer to skip adding a bunch of lines. Like right here this is a function it doesn't have access to anything and it's tests don't have access to anything and before I stopped adding stuff I was at 100 lines+ of changes just to  change that and have working tests.

Again I am pretty sure a lot of this code will be changes *again* and *again* while we do the remaining metric changes, so making it "better" in practically any way that adds more code or is harder seems like wasted time for me

na--: Hmm, yeah, I can see how it will be tricky in some of the standalone tests. That said, I'd probably prefer if we use naked strings in these tests instead, since we're only only verifying that some metric exists or something like that... :man_shrugging: 

Having the names as exported consts or not doesn't functionally matter, but the current way is more misleading than helpful IMO. It makes it seem OK to use the constants in normal non-test code, like here, when an explicit dependency through `WithBuiltinMetrics` will be easy and much cleaner. Still, as I mentioned, this is a relatively minor nitpick, and we can easily fix it later if we decide to, so I'm not very insistent on changing it now.

lib/metrics/registry_test.go
```diff
@@ -0,0 +1,53 @@
+package metrics
```
imiric, line 1: We really should add a linter for this :smile: Maybe [go-header](https://github.com/denis-tingaikin/go-header) works now?

```suggestion
/*
 *
 * k6 - a next-generation load testing tool
 * Copyright (C) 2021 Load Impact
 *
 * This program is free software: you can redistribute it and/or modify
 * it under the terms of the GNU Affero General Public License as
 * published by the Free Software Foundation, either version 3 of the
 * License, or (at your option) any later version.
 *
 * This program is distributed in the hope that it will be useful,
 * but WITHOUT ANY WARRANTY; without even the implied warranty of
 * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 * GNU Affero General Public License for more details.
 *
 * You should have received a copy of the GNU Affero General Public License
 * along with this program.  If not, see <http://www.gnu.org/licenses/>.
 *
 */

package metrics
```

imiric (approved):
LGTM. Not particularly fond of all the passing of `*BuiltinMetrics` and `*Registry` everywhere, and all the references kept around, but everything else looks clean :+1:

yorugac (approved):

lib/executor/constant_arrival_rate.go
```diff
@@ -346,7 +349,7 @@ func (car ConstantArrivalRate) Run(parentCtx context.Context, out chan<- stats.S
 			// dropped - we aren't going to try to recover it, but
 
 			stats.PushIfNotDone(parentCtx, out, stats.Sample{
-				Value: 1, Metric: metrics.DroppedIterations,
+				Value: 1, Metric: droppedIterationMetric,
```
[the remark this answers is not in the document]
MStoykov: I remember I did it so it's faster, but I don't remember if it was previously a function call or did I decide the dereference will be slow enough :thinking: I can revert it, but I don't think it matters much as again we are likely going to redo every call to `PushIfNotDone` at some point during this rewrite so I prefer not to rebase again just for this :(

lib/metrics/registry_test.go
MStoykov: uff

MStoykov: unfortunately it seems like there has been no improvement on the areas that were previously problematic from a quick look :(

MStoykov: I found other files with it missing so I prefer to close it in another PR with all of them

codebien (approved):

PR #2072, opened 2021-06-24
Basic showcase of variant 1 for #1802

MStoykov:
This will need more work if we want to help users more and the returned
object will likely need some help as well(which possibly can be done by
the importing code).


MStoykov:

PR #2073, opened 2021-06-24
Basic implementation of variant 2 from #1802

MStoykov:
In this case we give the Module a function to get the context and it
needs to that whenever it needs it instead of just caching it's value
... Especially between function call or something like that

### DIFF
--- a/js/modules/k6/metrics/metrics.go
+++ b/js/modules/k6/metrics/metrics.go
@@ -103,16 +103,23 @@ func (m Metric) GetName() string {
 	return m.metric.Name
 }
 
-type MetricsModule struct {
-	getContext func() context.Context
-}
+type (
+	RootMetricsModule struct{}
+	MetricsModule     struct {
+		getContext func() context.Context
+	}
+)
 
 func (m *MetricsModule) WithContext(getContext func() context.Context) {
 	m.getContext = getContext
 }
 
-func New() *MetricsModule {
+func (*RootMetricsModule) NewModuleInstancePerVU() interface{} {
 	return &MetricsModule{}
+}
+
+func New() *RootMetricsModule {
+	return &RootMetricsModule{}
 }
 
 // This is not possible after common.Bind as it wraps the object and doesn't return the original one.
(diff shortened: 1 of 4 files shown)


js/modules/k6/metrics/metrics.go
```diff
@@ -96,24 +103,59 @@ func (m Metric) GetName() string {
...
-	return newMetric(ctx, name, stats.Counter, isTime)
+// This is not possible after common.Bind as it wraps the object and doesn't return the original one.
+func (m *MetricsModule) ReturnMetricType(metric Metric) string {
+	return metric.metric.Type.String()
 }
 
-func (*Metrics) XGauge(ctx *context.Context, name string, isTime ...bool) (interface{}, error) {
-	return newMetric(ctx, name, stats.Gauge, isTime)
+// Counter ... // NOTE we still need to use goja.ConstructorCall  somewhere to have access to the
+func (m *MetricsModule) XCounter(call goja.ConstructorCall, rt *goja.Runtime) *goja.Object {
```
MStoykov, line 131: Having X infront keeps the next letter uppercase ;)

js/modules/k6/metrics/metrics.go
```diff
@@ -96,24 +103,59 @@ func (m Metric) GetName() string {
...
+		getContext func() context.Context
+	}
+)
+
+func (m *MetricsModule) WithContext(getContext func() context.Context) {
+	m.getContext = getContext
+}
+
+func (*RootMetricsModule) NewModuleInstancePerVU() interface{} {
+	return &MetricsModule{}
```
MStoykov, line 118: This is somewhat problematic as this :
1. exports withContext to the js which lets you do:
```
withContext(null)
```
bricking the module :facepalm: , this likely will only get worse with more `WIth` methods as we add
2. doesn't have a way  to make a difference between `default` import and not default import :
```
import something from "somewhere"; // this actually imports whatever the *default* import is as `something`
import * as something from "somewhere"; // this imports all the exported things from the module as fields on the object `something` (in a hand wavy way)
```

We currently can't differentiate between those, and I don't know of this will be useful, but might be required if/when goja has import/export support

We can implement a thing even now though as the babel transformation [makes this possible](https://babeljs.io/docs/en/babel-plugin-transform-modules-commonjs#babel) 

WDYT @na-- @imiric @codebien

MStoykov:

PR #2075, opened 2021-06-24
Basic implementation of variant 3 from #1802

MStoykov:
This goes all the way and tries (unfortunately not very well, I will try
again) to make the user embed the ModuleInstance it gets in the return
module so that it always has access to the Context and w/e else we
decide to add to it.

I also decided to force some esm along it (this is not required) to
test out some ideas.


MStoykov:

Issue #1089, opened 2019-07-24
Refactor loading and archiving or scripts post #1059


MStoykov:
While #1059 will do a lot of good for loading of scripts and their storage in archives:
1. We will correctly load remote scripts from archives instead of making http requests
2. You can now use full URLs with scheme  like `file://C:/path/to/somewhere.js` or `https://example.com/test.js` 
3. Fixing a lot of issues - see the PR 

During the development of #1059 multiple times I found out that I need to make more changes or add more problems to be solved in the future. The current PR is more or less the minimal amount of code that fixes problems without introducing new ones (mostly :(). Along the way two things became apparent: 
1. The whole `loader` package should be rewritten in a way that is more encapsulated. @na-- gave a good explanation at what that will probably look like in a [comment](https://github.com/loadimpact/k6/pull/1059#discussion_r303844608)
2. Putting the (relevant parts of a ) user filesystem into an archive and having it run on another system is *hard* and there are a lot of problems ... especially with windows.

For this I add this new issue where all current problems will be listed and a solution to them will be drafted, possibly with some problems being marked as non issues:
1. Currently and especially after #1059 you will be able to have multiple names for the same file and #1059 for various reasons doesn't fix all of them. This means that if you import a `utils.js` file with for example `example.com/utils.js` and `https://example.com/utils.js` it will be loaded and ran twice. The problems is even bigger if that package in hosted on github/cdnjs and someone uses the shortcut urls as that will be loaded again ...
2. Case insensitivity of (parts of) URLs and some filesystems: By [rfc3986](https://tools.ietf.org/html/rfc3986#section-6.2.2.1) both the scheme and host part of a URI are case insensitive. While that was [reported](https://github.com/golang/go/issues/3913) only the scheme is now lowered cased by default in golang. Which means that currently `https://example.com/utils.js` and `https://eXample.com/utils.js` are different files for k6 and will be loaded and executed separately.
Additionally windows is in general case insenstive, although you can [make it case sensitive](https://www.howtogeek.com/354220/how-to-enable-case-sensitive-folders-on-windows-10/) and I was told that macos's can be case insensitive, and I won't be surprised if you can make the same for one of the many filesystems that linux uses. So ... we will have to even check whether the filesystem is or is not case insensitive for the particular path we are using it and than ... do something ?!? 
[...]

PR #2078, opened 2021-06-25
Warn users on importing dependencies as absolute paths

MStoykov:
part of #1089

### DIFF
--- a/js/initcontext.go
+++ b/js/initcontext.go
@@ -26,6 +26,7 @@ import (
 	"fmt"
 	"net/url"
 	"path/filepath"
+	"runtime"
 	"strings"
 
 	"github.com/dop251/goja"
@@ -153,11 +154,6 @@ func (i *InitContext) requireModule(name string) (goja.Value, error) {
 func (i *InitContext) requireFile(name string) (goja.Value, error) {
 	// Resolve the file path, push the target directory as pwd to make relative imports work.
 	pwd := i.pwd
-	if filepath.IsAbs(name) {
-		i.logger.Warnf("you import '%s' with an absolute path, this won't be cross platform and likely will not work if"+
-			" you move the script between machines. Also it might not work on k6 cloud if it doesn't have `file://` in front",
-			name)
-	}
 	fileURL, err := loader.Resolve(pwd, name)
 	if err != nil {
 		return nil, err
@@ -166,6 +162,11 @@ func (i *InitContext) requireFile(name string) (goja.Value, error) {
 	// First, check if we have a cached program already.
 	pgm, ok := i.programs[fileURL.String()]
 	if !ok || pgm.module == nil {
+		if filepath.IsAbs(name) && runtime.GOOS == "windows" {
+			i.logger.Warnf("you import '%s' with an absolute path, this won't be cross platform and likely will not work if"+
+				" you move the script between machines. Also it might not work on k6 cloud if it doesn't have `file://` in front",
+				name)
+		}
 		i.pwd = loader.Dir(fileURL)
 		defer func() { i.pwd = pwd }()
 		exports := i.runtime.NewObject()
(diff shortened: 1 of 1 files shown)


js/initcontext.go
```diff
@@ -153,6 +153,11 @@ func (i *InitContext) requireModule(name string) (goja.Value, error) {
 func (i *InitContext) requireFile(name string) (goja.Value, error) {
 	// Resolve the file path, push the target directory as pwd to make relative imports work.
 	pwd := i.pwd
+	if filepath.IsAbs(name) {
+		i.logger.Warnf("you import '%s' with an absolute path, this won't be cross platform and likely will not work if"+
+			" you move the script between machines. Also it might not work on k6 cloud if it doesn't have `file://` in front",
```
codebien, line 158:
```suggestion
		i.logger.Warnf("you import '%s' with an absolute path, this won't be cross platform and likely will not work if"+
			" you move the script between machines or in cloud. You should use `file://` in front to have the full cross platform and cloud finally supported",
```
I would make "the suggestion" more explicit

codebien (changes requested):

na-- (changes requested):
:-1:, this is going to trigger for *nix file paths, which are not problematic. Also, the warning will be shown multiple times, once for every VU init.

MStoykov:
>  this is going to trigger for Linux file paths,

They still are not 
1. cross platform
2. cross machine in most cases, even if `k6 archive` + `k6 run archive.tar` on different machines will work (hopefully)

na--:
See my other point about the warning being shown once per VU, I submitted my review prematurely and edited the message. Also, if you want this warning to be for all absolute paths (which I don't), it should also be shown for `file://` imports, which it isn't.

js/initcontext.go
MStoykov: The thing is that "full cross platform" isn't true, as just moving the script between machines will still not work. It will work if you archive it and run the archive, but not if you than unarchive it on a different machine.

MStoykov: The reality of it is that the correct way for something to be cross-platform and cross-machine (as the OS is not the only problem) is to *not* use absolute paths at all.

So I really do prefer if that is the sentiment we sent more than "this won't work in the cloud if you are on Windows and use absolute paths without file://"

cc @na-- @imiric

js/initcontext.go
```diff
@@ -161,6 +162,11 @@ func (i *InitContext) requireFile(name string) (goja.Value, error) {
 	// First, check if we have a cached program already.
 	pgm, ok := i.programs[fileURL.String()]
 	if !ok || pgm.module == nil {
+		if filepath.IsAbs(name) && runtime.GOOS == "windows" {
+			i.logger.Warnf("you import '%s' with an absolute path, this won't be cross platform and likely will not work if"+
+				" you move the script between machines. Also it might not work on k6 cloud if it doesn't have `file://` in front",
```
na--, line 167:
```suggestion
			i.logger.Warnf("'%s' was imported with an absolute path - this won't be cross-platform and won't work if you move the script"+
				" between machines or run it with `k6 cloud`; if absolute paths are required, import them with the `file://` schema",
```

na-- (commented):
LGTM in general, made an inline comment with a suggestion for a better warning message

js/initcontext.go
MStoykov: the `file://` will only work for `k6 cloud` or if the do `k6 archive` and then run the archive on another machine. While `if absolute paths are required, import them with the file:// schema` sounds like that will fix everything.

na--: Hmm it doesn't to me, and I even change `might not work` from the first sentence to `won't work`. But if you want, change the end to `import them with the file:// schema for slightly better compatibility` :man_shrugging:

codebien (dismissed):

na-- (approved):

codebien (approved):

PR #2081, opened 2021-07-01
Remove deprecated outputs datadog and kafka

MStoykov:


cmd/outputs.go
```diff
@@ -48,16 +47,14 @@ func getAllOutputConstructors() (map[string]func(output.Params) (output.Output,
 		"cloud":    cloud.New,
 		"influxdb": influxdb.New,
 		"kafka": func(params output.Params) (output.Output, error) {
-			params.Logger.Warn("The kafka output is deprecated, and will be removed in a future k6 version. " +
+			return nil, errors.New("The kafka output is removed. " +
 				"Please use the new xk6 kafka output extension instead. " +
 				"It can be found at https://github.com/k6io/xk6-output-kafka.")
```
na--, line 52:
```suggestion
			return nil, errors.New("the kafka output was deprecated in k6 v0.32.0 and removed in k6 v0.34.0, " +
				"please use the new xk6 kafka output extension instead - https://github.com/k6io/xk6-output-kafka")
```

cmd/outputs.go
```diff
@@ -48,16 +47,14 @@ func getAllOutputConstructors() (map[string]func(output.Params) (output.Output,
...
-			return kafka.New(params)
 		},
 		"statsd": statsd.New,
 		"datadog": func(params output.Params) (output.Output, error) {
-			params.Logger.Warn("The datadog output is deprecated, and will be removed in a future k6 version. " +
+			return nil, errors.New("the datadog output is removed. " +
 				"Please use the statsd output with env variable K6_STATSD_ENABLE_TAGS=true instead.")
```
na--, line 57:
```suggestion
			return nil, errors.New("the datadog output was deprecated in k6 v0.32.0 and removed in k6 v0.34.0, " +
				"please use the statsd output with env. variable K6_STATSD_ENABLE_TAGS=true instead")
```

go.mod
```diff
@@ -10,11 +10,12 @@ require (
 	github.com/andybalholm/brotli v1.0.2
 	github.com/dop251/goja v0.0.0-20210427212725-462d53687b0d
 	github.com/fatih/color v1.11.0
+	github.com/gedex/inflector v0.0.0-20170307190818-16278e9db813 // indirect
```
na--, line 13: Why do we have additions to `go.mod` when we're removing a dependency? :confused:

MStoykov: This now specifies the version that was required by the removed dependency. 
As go always get the lowest possible version, when you remove a dependency that was requiring the *highest* possible version it will now require something different, which is likely not what you want, so it needs to now require that specific version in go.mod in order for the dependency version to not change

na--: So right now (and probably before), the versions of these indirect dependencies the k6 vendors are not the ones that were required by the actual direct dependencies' `go.mod` files, right? If so, it's probably not a big deal, but we should probably clean this up now or when we update the direct dependencies of k6? We don't have a good reason for the current state, right?

na-- (approved):

go.mod
MStoykov: the versions of the `inderect` dependencies were probably all added as case like
The general case I think is, we have:
> k6 -> A -> C(v1.2.0) 
> k6 -> B -> C(v1.3.0)

we dropped dependency `B`, and if this isn't added we will now get `C` `v1.2.0` as that is the highest version required.

So go mod adds an indirect dependency to `C` `v1.3.0`, so we don't change the version of something we didn't intend to. 
If at a later point we update `A` and it requires `C` `v1.4.0` it will be that version that will be used, and from my quick test it will also remove the `indirect` dependency in go.mod(as it's no longer changing what versions we will get).
 If we drop `A` (in this case that would be `api2go` ;) ) we will probably also drop the `indirect` dependencies as well.
 
 So we can clean whenever and change the dependency version, which seems like a way to get something broken. Or just when updating/dropping dependency it will fix itself

MStoykov: example: do `go get google.golang.org/grpc` and see how it drops the indirect dependency on `xerrors`

imiric (approved):
LGTM, love to see the negative diff stat. :)

I would've preferred to leave this for the end of the v0.34.0 cycle to keep `master` working until right before release, but it's not a big deal.

MStoykov:
I do prefer to do it now, so we can actually update kafka extension output to depend on version of k6 without that k6 depending on the kafka extension output

na--:
Yeah, dropping them ASAP is probably going to be cleaner, and xk6 users shouldn't be affected, since by default they use the `latest` k6, right? So... :shipit:

PR #2082, opened 2021-07-02
Source map support

MStoykov:


MStoykov:
This basically fixes https://github.com/k6io/k6/issues/1789 , although there is still more work to be done especially given that this has zero automatic testing. Also, I haven't done any archive testing.

na-- (changes requested):
Haven't reviewed the code yet, but you seem to have committed a whole k6 binary in one of the commits, so please clean this up and force-push: https://github.com/k6io/k6/blob/aacadb7e7a69a7b2b828df3ac787d018641c227b/js/compiler/k6

Noticed this when I `git pulled` the latest changes and it took a while:
```
remote: Total 116 (delta 78), reused 110 (delta 75), pack-reused 0
Receiving objects: 100% (116/116), 15.37 MiB | 1.41 MiB/s, done.
```
:joy:

MStoykov:
This is now ready for review :tada: Although definitely headed for 0.35.0 not 0.34.0  

It  seems to finally be working very well for everything that I have tried and doesn't seem to have all that big of performance implications - on a script taking 2m20s and 1.7GB it was taking more or less the same time (it fluctuates even without it) and 100mb more (again it fluctuates this is the biggest I have seen). 

Notes:
- in case we need to be able to disable and reenable it I think (by idea from @na-- ) that we can just use the compatibilityMode possibly with 2 more values to either disable or enable the sourcemaps depending on how it turns out
- it used to have an additional flag which is why some of the additional changes were made. The whole compilerOptions can be removed but maybe it's time to make something like that given how many arguments it's taking now :(
- obviously, we will have to wait for the source map fixes to be merged

MStoykov:
of course I have now found that one of my even bigger scripts (24k from the cloud request builder) which takes 1min and 1.7gb(the other script was only 10k but with some posts) now takes 20 seconds more and even more importantly 5.5gb of memory o.O I have no idea why this is happening, but I guess we need to look into it some more :(

js/bundle.go
```diff
@@ -132,7 +137,13 @@ func NewBundleFromArchive(
...
 
 	c := compiler.New(logger)
-	pgm, _, err := c.Compile(string(arc.Data), arc.FilenameURL.String(), "", "", true, compatMode)
+	c.COpts = compiler.Options{
+		Strict:            true,
+		CompatibilityMode: compatMode,
+		SourceMapEnabled:  true,
+		SourceMapLoader:   generateSourceMapLoader(logger, arc.Filesystems),
+	}
+	pgm, _, err := c.Compile(string(arc.Data), arc.FilenameURL.String(), true, c.COpts)
```
MStoykov, line 146: This can probably be reverted, but still seems kind of nicer to me so I am leaving it for now

js/compiler/compiler_test.go
```diff
@@ -71,13 +72,26 @@ func TestTransform(t *testing.T) {
...
+		c.COpts.SourceMapEnabled = true
+		src, _, err := c.Transform("()=> true", "test.js", nil)
+		assert.NoError(t, err)
+		assert.Equal(t, `"use strict";
+
+() => true;
+//# sourceMappingURL=k6://internal-should-not-leak/file.map`, src)
+		// assert.Equal(t, 3, srcmap.Version)
+		// assert.Equal(t, "test.js", srcmap.File)
+		// assert.Equal(t, "aAAA,qBAAK,IAAL", srcmap.Mappings)
```
codebien, line 86: Should we enable them?

codebien (changes requested):
Thanks for working on it :clap:, I left a few comments just to "start the discussion". I expect I need to play a bit more with the generated Source Maps.

I expect this will be an important feature, it would be appreciated if we could maintain the commit history atomic and with detailed commit messages I think it could help future readers. 

>now takes 20 seconds more and even more importantly 5.5gb of memory

I expect the JSON marshal/unmarshal could have an important impact with big source maps.

js/bundle.go
MStoykov: yeah, I mean that currently, this doesn't let you wrap the code in w/e you want. If it's dependency it will wrap it in the `function(module, exports ...` wrapper if not it won't. And all of that with 1 boolean instead of 2 strings. 

This was never used for anything apart from that outside of tests so :shrug: it seems like a good change to me

js/compiler/compiler.go
```diff
@@ -86,10 +87,13 @@ var (
 	globalBabel        *babel        // nolint:gochecknoglobals
 )
 
+const sourceMapURLFromBabel = "k6://internal-should-not-leak/file.map"
```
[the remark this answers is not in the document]
MStoykov: This is literally a constant that babel sets in the file which tells goja (in this case) where the sourcemap is. Goja asks k6 to get this and we return the sourcemap babel generated. It's just a particular *very* unlikely value for a real sourcemap URL ;)

js/compiler/compiler.go
```diff
@@ -119,48 +123,81 @@ func (c *Compiler) Transform(src, filename string) (code string, srcmap []byte,
...
+	code, srcmap, err = c.babel.transformImpl(c.logger, src, filename, c.COpts.SourceMapEnabled, inputSrcMap)
+	// fmt.Println(code)
 	return
 }
 
+// Options are options to the compiler
+type Options struct { // TODO maybe have the fields an exported and use the functional options pattern
+	CompatibilityMode lib.CompatibilityMode
+	SourceMapEnabled  bool
+	// TODO maybe move only this in the compiler itself and leave ht rest as parameters to the Compile
```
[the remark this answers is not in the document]
MStoykov: I guess it depends of we want to do it *or* not. I kind of prefer the options struct as it seems more ... structured ( :wink: :wink: ). If we want to keep it I will drop the TODO, otherwise I will need to DO it.

js/compiler/compiler.go
```diff
@@ -119,48 +123,81 @@ func (c *Compiler) Transform(src, filename string) (code string, srcmap []byte,
...
-	code := pre + src + post
-	ast, err := parser.ParseFile(nil, filename, code, 0, parser.WithDisableSourceMaps)
-	if err != nil {
-		if compatMode == lib.CompatibilityModeExtended {
-			code, _, err = c.Transform(src, filename)
+func (c *Compiler) Compile(src, filename string, main bool, cOpts Options) (*goja.Program, string, error) {
+	return c.compileImpl(src, filename, main, cOpts, nil)
+}
+
+//nolint:cyclop
```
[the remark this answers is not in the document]
MStoykov: I will try to split something out ... but I don't know if that will be any better. I do think I can drop the second try at transforming as I don't think with the current babel plugins there are any cases(also I dropped the comment :facepalm: )

js/compiler/compiler_test.go
MStoykov: This is leftover (as are the rest of the similar commented checks) from when k6 parsed the sourcemap. But in reality we don't need to as k6 expects unparsed so we have no need to parse it into a struct and then check what the values are IMO

js/compiler/compiler.go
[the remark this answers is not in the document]
yorugac: :+1: for options struct
Also, it seems `Compile` is called only in 3 places, in bundle.js and in initcontext.js: it doesn't look as if anything would need an overwrite of compiler options, etc. So I think it can stay simple. Unless there is an expectation of that changing..

js/compiler/compiler.go
```diff
@@ -218,7 +283,24 @@ func (b *babel) Transform(logger logrus.FieldLogger, src, filename string) (stri
...
 		return code, nil, err
 	}
-	return code, nil, err
+	if !sourceMapsEnabled {
+		return code, nil, nil
+	}
+
+	// this is to make goja try to load a sourcemap.
+	// it is specifically a special url as it should never leak outside of this code
+	// additionally the alternative support from babel is to embed *the whole* sourcemap at the end
```
yorugac, line 292:
```suggestion
	// it is a special url as it should never leak outside of this code
	// additionally the alternative support from babel is to embed *the whole* sourcemap at the end
```

yorugac (commented):
If I understood correctly, this PR has some pending design decisions yet so I just added some comments and a question, please see.

ATM, it seems rather straight-forward in terms of implementation changes even with my lacking knowledge of sourcemaps spec :sweat_smile:

And one more thing: `TestSourceMapsExternal` shows one example of externally generated sourcemap but it seems like there's more than one external tool that can do that. Are there any expectations of the range of those tools working: are all of them going to work or only some of them, etc. ?

js/compiler/compiler.go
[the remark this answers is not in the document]
codebien: :+1: same for me for keeping it and removing the comment

codebien (changes requested):
Few additional comments

MStoykov:
> I expect this will be an important feature, it would be appreciated if we could maintain the commit history atomic and with detailed commit messages I think it could help future readers. 

Tried to keep the more important parts :) 

Also, I might move the modules updates to a separate PR if this takes more time.

> And one more thing: TestSourceMapsExternal shows one example of externally generated sourcemap but it seems like there's more than one external tool that can do that. Are there any expectations of the range of those tools working: are all of them going to work or only some of them, etc. ? 

The specification for this has been like that for nearly 8 years (not that many updates have been done either way) so hopefully everything at this point generates “standardized”(it's not really a standard) source maps, and it will work.

I definitely though expect that it will break, both goja and the source map library might have bugs, and [we definitely](https://github.com/grafana/k6/pull/2082/files#diff-b81bc95f1f1a81f26e9a79e1f64bf9cbde72a4824a05681905aa7c728544d294R260-R262) need to have some more fixes. 

But I would expect this will work for most cases and the case we care the most where we have source file, and it goes through our babel and then we no longer have accurate lines :sob:

js/compiler/compiler.go
```diff
@@ -194,16 +244,53 @@ func newBabel() (*babel, error) {
...
+		// TODO (kind of alternatively) drop the newline in the "commonjs" wrapping and have only the first line wrong
+		// except everything being off by one
+		return sourceMap, nil
+	}
+	if str, ok := mappings.(string); ok {
+		// ';' is the separator between lines so just adding 1 will make all mappings be for the line after which they were
+		// originally
+		m["mappings"] = ";" + str
+	} else {
+		// we have mappings but it's not a string - this is some kind of erro
```
codebien, line 266:
```suggestion
		// we have mappings but it's not a string - this is some kind of error
```

codebien (changes requested):
Thanks for the polishing :pray:

https://github.com/grafana/k6/pull/2082#issuecomment-902726202

What is the status of it? Did you get more details?

js/compiler/compiler_test.go
```diff
@@ -67,104 +67,133 @@ func TestTransform(t *testing.T) {
...
-				// This is a string with U+2029 Paragraph separator in it
-				// the important part is that goja won't parse it but babel will transform it but still
-				// goja won't be able to parse the result it is actually "\<U+2029>"
-				_, _, err := c.Compile(string([]byte{0x22, 0x5c, 0xe2, 0x80, 0xa9, 0x22}), "script.js", "", "", true, lib.CompatibilityModeExtended)
-				assert.IsType(t, parser.ErrorList{}, err)
-				assert.Contains(t, err.Error(), ` Unexpected token ILLEGAL`)
-			}()
+	t.Run("Invalid for goja but not babel", func(t *testing.T) {
+		t.Parallel()
+		t.Skip("Find something else that breaks this as this was fixed in goja :(")
```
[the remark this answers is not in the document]
MStoykov: I guess this is becoming harder and harder to find and likely will be, so I will drop it :+1:

codebien (dismissed):
Seems we have done here! :clap:

js/compiler/compiler.go
```diff
@@ -119,48 +124,92 @@ func (c *Compiler) Transform(src, filename string) (code string, srcmap []byte,
...
+type Options struct {
+	CompatibilityMode lib.CompatibilityMode
+	SourceMapLoader   func(string) ([]byte, error)
+	Strict            bool
+}
+
+// compilationState is helper struct to keep the state of a compilation
+type compilationState struct {
+	// set when we couldn't load external source map so we can try parsing without loading it
+	couldntLoadSourceMap bool
```
olegbespalov, line 141: I'm wondering if we can avoid this property. For example, using the wrapped errors and checks like `errors.Is`. It seems simpler and more go idiomatic, wdyt?

js/compiler/compiler.go
```diff
@@ -194,16 +243,57 @@ func newBabel() (*babel, error) {
 	return result, err
 }
 
-// Transform the given code into ES5, while synchronizing to ensure only a single
+func increaseMappingsByOne(sourceMap []byte) ([]byte, error) {
```
na--, line 246: minor nitpick, but this function probably deserves a comment for _why_ it exists and does what it does

js/compiler/compiler.go
```diff
@@ -119,48 +124,92 @@ func (c *Compiler) Transform(src, filename string) (code string, srcmap []byte,
...
+	return c.srcMap, err
+}
+
+func (c *Compiler) compileImpl(
+	src, filename string, main bool, cOpts Options, srcMap []byte,
+) (*goja.Program, string, error) {
+	code := src
+	state := compilationState{srcMap: srcMap, compiler: c, main: main}
+	if !main {
+		code = "(function(module, exports){\n" + code + "\n})\n"
```
na--, line 185: doesn't this affect the source map? :confused:

js/compiler/compiler.go
```diff
@@ -194,16 +243,57 @@ func newBabel() (*babel, error) {
...
+		// except everything being off by one
+		return sourceMap, nil
+	}
+	if str, ok := mappings.(string); ok {
+		// ';' is the separator between lines so just adding 1 will make all mappings be for the line after which they were
+		// originally
+		m["mappings"] = ";" + str
+	} else {
+		// we have mappings but it's not a string - this is some kind of error
+		return nil, errors.New(`missing "mappings" in sourcemap`)
```
na--, line 266: I don't completely understand what's going on here and _why_ this function is necessary, but from what I can see, if we hit this error, k6 will actually abort the script, right? `couldntLoadSourceMap` won't be set and sourcemaps ignored, we'll just exit with an error?

js/compiler/compiler.go
MStoykov: The problem @olegbespalov is that the error doesn't get propaged through goja and fixing that will take quite a lot refactoring IIRC, so I opted for *not* trying to do that first.

js/compiler/compiler.go
MStoykov: this is what https://github.com/grafana/k6/pull/2082/files#r782919998 fixes ;)

js/compiler/compiler.go
MStoykov: yes if we get somehow a sourcemap that does have "mappings" - arguably the thing that a sourcemap is for (most other fields aren't really necessary always) and it happens to not be a string we will error out. I guess we can disable the sourcemap in that case as well and log it :shrug:

js/compiler/compiler.go
na--: that isn't obvious from the code or the comments :man_shrugging:

js/compiler/compiler.go
na--: :+1: a warning seems like the better approach than to completely fail

MStoykov: 44c2849   should fix it and I will try to write a test for this

js/compiler/compiler.go
MStoykov: is  44c2849  sufficient?

js/compiler/compiler.go
```diff
@@ -243,7 +243,9 @@ func newBabel() (*babel, error) {
 	return result, err
 }
 
-func increaseMappingsByOne(sourceMap []byte) ([]byte, error) {
+// this function increases the lines in the sourcemap by line so that it fixes the case where we need to wrap an
+// required file in a function to support/emulate commonjs
+func (c *compilationState) increaseMappingsByOne(sourceMap []byte) ([]byte, error) {
```
codebien, line 248:
```suggestion
// increaseMappingsByOne increases the lines in the sourcemap by line so that it fixes the case where we need to wrap a
// required file in a function to support/emulate commonjs
func (c *compilationState) increaseMappingsByOne(sourceMap []byte) ([]byte, error) {
```

codebien (changes requested):

na-- (dismissed):
LGTM. Things are a bit convoluted, but they should be simplified when we get rid of Babel :crossed_fingers:

codebien (dismissed):

js/compiler/compiler.go
olegbespalov: :cry: I see, thanks for clarifying :+1:

na-- (approved):
:crossed_fingers:

js/compiler/compiler.go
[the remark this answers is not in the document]
yorugac: Partially out of curiosity: what is the worst that can happen if someone does use this internal URL deliberately for external sourcemap?

yorugac (approved):
It seems to be more readable than before :+1: though it might be just multiple re-reads effect :slightly_smiling_face: I have some remarks on comments mostly, please see.

MStoykov:
A problem I found around doing some final testing and especially after https://github.com/grafana/k6/pull/2082#discussion_r782928602:

If the sourcemap is actually broken (in some obvious way) or just can't be parsed as JSON in general this will now abort the test and at least in some cases we will have hard time noticing it as you can have: 

```javascript
var o={d:(e,r)=>{for(var t in r)o.o(r,t)&&!o.o(e,t)&&Object.defineProperty(e,t,{enumerable:!0,get:r[t]})},o:(o,e)=>Object.prototype.hasOwnProperty.call(o,e)},e={};o.d(e,{Z:()=>r});const r=()=>{!function(o){throw"cool is cool"}()};var t=e.Z;export{t as default};
//# sourceMappingURL=data:application/json;charset=utf-8;base64,eyJ2ZXJzaW9uIjozLCJzb3VyY2VzIjpbIndlYnBhY2s6Ly8vd2VicGFjay9ib290c3RyYXAiLCJ3ZWJwYWNrOi8vL3dlYnBhY2svcnVudGltZS9kZWZpbmUgcHJvcGVydHkgZ2V0dGVycyIsIndlYnBhY2s6Ly8vd2VicGFjay9ydW50aW1lL2hhc093blByb3BlcnR5IHNob3J0aGFuZCIsIndlYnBhY2s6Ly8vLi90ZXN0MS50cyJdLCJuYW1lcyI6WyJfX3dlYnBhY2tfcmVxdWlyZV9fIiwiZXhwb3J0cyIsImRlZmluaXRpb24iLCJrZXkiLCJvIiwiT2JqZWN0IiwiZGVmaW5lUHJvcGVydHkiLCJlbnVtZXJhYmxlIiwiZ2V0Iiwib2JqIiwicHJvcCIsInByb3RvdHlwZSIsImhhc093blByb3BlcnR5IiwiY2FsbCIsInMiLCJjb29sVGhyb3ciXSwibWFwcGluZ3MiOiJBQUNBLElBQUlBLEVBQXNCLENDQTFCLEVBQXdCLENBQUNDLEVBQVNDLEtBQ2pDLElBQUksSUFBSUMsS0FBT0QsRUFDWEYsRUFBb0JJLEVBQUVGLEVBQVlDLEtBQVNILEVBQW9CSSxFQUFFSCxFQUFTRSxJQUM1RUUsT0FBT0MsZUFBZUwsRUFBU0UsRUFBSyxDQUFFSSxZQUFZLEVBQU1DLElBQUtOLEVBQVdDLE1DSjNFLEVBQXdCLENBQUNNLEVBQUtDLElBQVVMLE9BQU9NLFVBQVVDLGVBQWVDLEtBQUtKLEVBQUtDLEksc0JDR2xGLGNBSEEsU0FBbUJJLEdBQ2YsS0FBTSxlQUdOQyxJIiwiZmlsZSI6InRlc3QxLmpzIiwic291cmNlc0NvbnRlbnQiOlsiLy8gVGhlIHJlcXVpcmUgc2NvcGVcbnZhciBfX3dlYnBhY2tfcmVxdWlyZV9fID0ge307XG5cbiIsIi8vIGRlZmluZSBnZXR0ZXIgZnVuY3Rpb25zIGZvciBoYXJtb255IGV4cG9ydHNcbl9fd2VicGFja19yZXF1aXJlX18uZCA9IChleHBvcnRzLCBkZWZpbml0aW9uKSA9PiB7XG5cdGZvcih2YXIga2V5IGluIGRlZmluaXRpb24pIHtcblx0XHRpZihfX3dlYnBhY2tfcmVxdWlyZV9fLm8oZGVmaW5pdGlvbiwga2V5KSAmJiAhX193ZWJwYWNrX3JlcXVpcmVfXy5vKGV4cG9ydHMsIGtleSkpIHtcblx0XHRcdE9iamVjdC5kZWZpbmVQcm9wZXJ0eShleHBvcnRzLCBrZXksIHsgZW51bWVyYWJsZTogdHJ1ZSwgZ2V0OiBkZWZpbml0aW9uW2tleV0gfSk7XG5cdFx0fVxuXHR9XG59OyIsIl9fd2VicGFja19yZXF1aXJlX18ubyA9IChvYmosIHByb3ApID0+IChPYmplY3QucHJvdG90eXBlLmhhc093blByb3BlcnR5LmNhbGwob2JqLCBwcm9wKSkiLCJmdW5jdGlvbiBjb29sVGhyb3coczogc3RyaW5nKSB7XG4gICAgdGhyb3cgXCJjb29sIFwiKyBzXG59XG5leHBvcnQgZGVmYXVsdCAoKSA9PiB7XG4gICAgY29vbFRocm93KFwiaXMgY29vbFwiKVxufTtcbiJdLCJzb3VyY2VSb290IjoiIn0=
```
A file with embeded sourcemap, if you now corrupt this (delete some symbols in the base64 encoded string) The error we get back is something like(using spew-go):
```
(*goja.Exception)(0xc001e4b920)(SyntaxError: file:///home/mstoykov/work/k6io/k6/test1.js: invalid character '9' after object key at <internal/k6/compiler/lib/babel.min.js>:2:28536(109))
```
In this particular case as this code *still* goes through babel and this is what it returns. if it didn't we will get
```
(parser.ErrorList) (len=1 cap=1) file:///home/mstoykov/work/k6io/k6/test1.js: Line 3:1 Could not load source map: illegal base64 data at input byte 1724
```
Which at least is somewhat matchable with a string :shrug: . 

I don't think this is a particularly likely occurrence *and* will become less of a problem as time goes on and hopefully we use less babel. 

We can also always *on error with sourcemap enabled* retry with it disabled, but this goes and more in trying to handle random corner cases that I doubt will be hit by anyone. And obviously we can always revert this or add the necessary changes or add a flag to enable/disable source map if the need arises.

js/compiler/compiler.go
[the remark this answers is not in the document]
MStoykov: The code returning the sourcemap to goja matches first on it and returns an internal field that get's set when babel transforms. If babel hasn't ran the field is nil - so nothing, if it has it would've set this url.

js/compiler/compiler.go
```diff
@@ -119,48 +124,92 @@ func (c *Compiler) Transform(src, filename string) (code string, srcmap []byte,
...
+	CompatibilityMode lib.CompatibilityMode
+	SourceMapLoader   func(string) ([]byte, error)
+	Strict            bool
+}
+
+// compilationState is helper struct to keep the state of a compilation
+type compilationState struct {
+	// set when we couldn't load external source map so we can try parsing without loading it
+	couldntLoadSourceMap bool
+	// srcMap is the current full sourceMap that has been generated read so far
```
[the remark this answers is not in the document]
MStoykov: it's an `or` - it's either read from disk or generated by babel

js/compiler/compiler.go
```diff
@@ -194,16 +243,61 @@ func newBabel() (*babel, error) {
...
+	var err error
+	m := make(map[string]interface{})
+	if err = json.Unmarshal(sourceMap, &m); err != nil {
+		return nil, err
+	}
+	mappings, ok := m["mappings"]
+	if !ok {
+		// no mappings, no idea what this will do, but just return it as technically we can have sourcemap with sections
+		// TODO implement incrementing of `offset` in the sections?
+		// TODO (kind of alternatively) drop the newline in the "commonjs" wrapping and have only the first line wrong
```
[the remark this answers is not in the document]
MStoykov: > // no mappings, no idea what this will do, but just return it as technically we can have sourcemap with sections  
 		
and 

> // TODO implement incrementing of `offset` in the sections? 

are connected, sections is part of an *extension* of the sourcemap [specification](https://sourcemaps.info/spec.html#h.n05z8dfyl3yh). 

> // TODO (kind of alternatively) drop the newline in the "commonjs" wrapping and have only the first line wrong

Has to do with just removing the whole function here. 

> but it's not clear what kind of problems the current approach here can result in ?

Given that we don't do anything to the sourcemap we don't fix the off by one error commonjs wrapping adds - see the newly added documentation of the method.

js/compiler/compiler.go
```diff
@@ -119,48 +124,92 @@ func (c *Compiler) Transform(src, filename string) (code string, srcmap []byte,
...
+	return c.compileImpl(src, filename, main, cOpts, nil)
+}
+
+// here we take the srcMap as pointer so we can change in case that we have an original srcMap that will be loaded, but
+// the code still needs to go through babel. In this way we can give the correct inputSourceMap to babel which hopefully
+// will generate a sourcemap which goes from the final source file lines to the original ones that were transpiled
+// outside of k6
+// TODO have some kind of "compilationContext" to save this instead and modify and move that
+// This won't be needed if we weren't going through babel so removing babel would kind of remove the need for that
+func (c *compilationState) sourceMapLoader(path string) ([]byte, error) {
```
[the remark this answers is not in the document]
MStoykov: update in 51b7df6

js/compiler/compiler.go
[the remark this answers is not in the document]
MStoykov: Added some more explanations in 51b7df6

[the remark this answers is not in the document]
yorugac: Okay, I think I got it. It's not about "fixing smth" but more about potential enhancements. Thanks for the update :+1:

na-- (approved):

na-- (approved):

codebien (approved):

PR #2085, opened 2021-07-05
lib/options: BlacklistIPs json encoding/decoding

codebien:
Support JSON encoding for the `BlacklistIPs` option.

### DIFF
--- a/lib/options.go
+++ b/lib/options.go
@@ -28,10 +28,9 @@ import (
 	"reflect"
 	"strconv"
 
-	"gopkg.in/guregu/null.v3"
-
 	"go.k6.io/k6/lib/types"
 	"go.k6.io/k6/stats"
+	"gopkg.in/guregu/null.v3"
 )
 
 // DefaultScenarioName is used as the default key/ID of the scenario config entries
@@ -178,8 +177,12 @@ func (ipnet *IPNet) UnmarshalText(b []byte) error {
 	}
 
 	*ipnet = *newIPNet
-
 	return nil
+}
+
+// MarshalText encodes the IPNet representation using CIDR notation.
+func (ipnet *IPNet) MarshalText() ([]byte, error) {
+	return []byte(ipnet.String()), nil
 }
 
 // HostAddress stores information about IP and port
--- a/lib/options_test.go
+++ b/lib/options_test.go
@@ -307,15 +307,28 @@ func TestOptions(t *testing.T) {
 		opts := Options{}.Apply(Options{
 			BlacklistIPs: []*IPNet{{
 				IPNet: net.IPNet{
-					IP:   net.IPv4zero,
-					Mask: net.CIDRMask(1, 1),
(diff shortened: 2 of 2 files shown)


lib/options.go
```diff
@@ -178,10 +178,20 @@ func (ipnet *IPNet) UnmarshalText(b []byte) error {
...
 
+// MarshalJSON returns the JSON representation of IPNet using CIDR notation.
+func (ipnet *IPNet) MarshalJSON() ([]byte, error) {
+	return []byte(fmt.Sprintf("%q", ipnet.String())), nil
+}
+
+// UnmarshalJSON decodes a JSON representation of IPNet using CIDR notation.
+func (ipnet *IPNet) UnmarshalJSON(b []byte) error {
+	// Remove quotes and fallback to unmarshal text
+	return ipnet.UnmarshalText([]byte(strings.Trim(string(b), `"`)))
```
MStoykov, line 192: There is [bytes.Trim](https://golang.org/pkg/bytes/#Trim).

lib/options.go
```diff
@@ -178,10 +178,20 @@ func (ipnet *IPNet) UnmarshalText(b []byte) error {
...
+}
+
```
MStoykov, line 194: I think this is fine, but in reality, all you need is MarshalText as it seems that we fall back to that and if you just replace all of that with MarshalText that just return `[]byte(ipnet.String())` it seems still to work.

MStoykov (commented):
I am not certain why the only MarshalText works fine, but if it does and it's simpler it's probably better then adding 2 methods ;)

codebien:
>I am not certain why the only MarshalText works fine, but if it does and it's simpler it's probably better then adding 2 methods ;)

@MStoykov I agree, some tests I did confuse me. The problem is just in the encoding part and we can fall back on `TextMarshaler`.

MStoykov (dismissed):

lib/options.go
```diff
@@ -572,6 +575,11 @@ func (o Options) Validate() []error {
 					o.ExecutionSegment, o.ExecutionSegmentSequence))
 		}
 	}
+	for _, ip := range o.BlacklistIPs {
+		if ip.String() == "<nil>" {
+			errors = append(errors, fmt.Errorf("blacklisted IP %q has an invalid CIDR", ip.IP.String()))
+		}
```
na--, line 581: This is a very strange check, why do you compare against the string? And the error message doesn't really follow logically from the check. And in the check you're comparing `ip`, but in the error message you're referencing `ip.IP` :confused: 

So this is weird in multiple different ways, and if needs to be this way for some reason, that reason should be mentioned in a code comment here.

na--: And probably in the PR description, since this seems like a different fix than just being able to marshal the `blacklistIPs` in the JSON properly? :confused:

na-- (approved):

MStoykov (approved):

PR #2090, opened 2021-07-13
cloudapi/logs: Keep connection reliable

### LINKED ISSUE #1966: k6 cloud gets log error on long running tests
Since #1599 `k6 cloud` by default tail logs from the cloud effectively trying to simulate even more how `k6 run` UX. In the cloud currently, we use [loki](https://grafana.com/oss/loki/) and it turns out that it has a maximum time for tailing `tail_max_duration` with a [default value](https://grafana.com/docs/loki/latest/configuration/#querier_config) of `1h`. 

This in practice means that after 1hour k6 will get an error and stop tailing logs. Nothing else will break as the feature was added specifically in a way so that it doesn't interfere with the normal execution. 

Given the above and the fact that we could have network problems, it is probably better if k6 can notice error during tailing and try to reconnect (a few times). This will be fairly straightforward and easy the harder part will be to actually not miss or duplicate messages so it will likely still result in some amount of not "cloud logs" in the form of "we are reconnecting to tail logs, this might result in either some repeated or missed messages reason=error something". It probably is even a good idea to reconnect every 50 minutes or something like that so we don't even receive the error in the first place.

Obviously all of the above should have some amount of configuration: `retries`, `time between retries`, `reconnect every X minutes`. 

Originally reported in https://community.k6.io/t/tail-errors-after-k6-cloud-execution/1619

codebien:
It adds some capabilities to reconnect to our log tail service.

Some questions/notes:
* Re-generating `cloudapi` package the `_easyjson` file gets some changes
* Should we support the options from the json configuration? The already existing option doesn't so I replicated it.
* I added `gomock` as a dependency to support mocking some interfaces for unit testing. I find it generally very useful but we are not using it so I expect some cons with k6 that I'm not considering.
* The current abstraction is still tight with the current `gorilla/websocket`, if I'm in the right way with this PR, we could consider designing a better abstraction from the real streaming implementation.
* The current logic closes the connection before re-dial this generates to get some logs like this:
   ```
   INFO[0064] we are reconnecting to tail logs, this might result in either some repeated or missed messages
   ERRO[0065] couldn't unmarshal a message from the cloud:   error=EOF
   ERRO[0065] couldn't unmarshal a message from the cloud:   error=EOF
  ```
  should we improve the UX and avoid this message or log something more specific?

Closes #1966

### DIFF
--- a/cloudapi/logs.go
+++ b/cloudapi/logs.go
@@ -23,9 +23,13 @@ package cloudapi
 import (
 	"context"
 	"fmt"
+	"math"
+	"math/rand"
 	"net/http"
 	"net/url"
 	"strconv"
+	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/gorilla/websocket"
@@ -53,8 +57,11 @@ type msgDroppedEntries struct {
 	Timestamp string            `json:"timestamp"`
 }
 
-func (m *msg) Log(logger logrus.FieldLogger) {
+// Log writes the Streams and Dropped Entries to the passed logger.
+// It returns the most recent timestamp seen overall messages.
+func (m *msg) Log(logger logrus.FieldLogger) int64 {
 	var level string
+	var ts int64
 
 	for _, stream := range m.Streams {
 		fields := labelsToLogrusFields(stream.Stream)
@@ -64,8 +71,8 @@ func (m *msg) Log(logger logrus.FieldLogger) {
 		}
 
 		for _, value := range stream.Values {
-			nsec, _ := strconv.Atoi(value[0])
-			e := logger.WithFields(fields).WithTime(time.Unix(0, int64(nsec)))
+			nsec, _ := strconv.ParseInt(value[0], 10, 64)
+			e := logger.WithFields(fields).WithTime(time.Unix(0, nsec))
 			lvl, err := logrus.ParseLevel(level)
 			if err != nil {
 				e.Info(value[1])
@@ -73,13 +80,24 @@ func (m *msg) Log(logger logrus.FieldLogger) {
 			} else {
 				e.Log(lvl, value[1])
 			}
+
+			// find the latest seen message
+			if nsec > ts {
+				ts = nsec
+			}
 		}
 	}
 
 	for _, dropped := range m.DroppedEntries {
-		nsec, _ := strconv.Atoi(dropped.Timestamp)
-		logger.WithFields(labelsToLogrusFields(dropped.Labels)).WithTime(time.Unix(0, int64(nsec))).Warn("dropped")
+		nsec, _ := strconv.ParseInt(dropped.Timestamp, 10, 64)
+		logger.WithFields(labelsToLogrusFields(dropped.Labels)).WithTime(time.Unix(0, nsec)).Warn("dropped")
+
+		if nsec > ts {
+			ts = nsec
+		}
 	}
+
+	return ts
 }
 
 func labelsToLogrusFields(labels map[string]string) logrus.Fields {
@@ -92,42 +110,47 @@ func labelsToLogrusFields(labels map[string]string) logrus.Fields {
 	return fields
 }
 
-func (c *Config) getRequest(referenceID string, start time.Duration) (*url.URL, error) {
+func (c *Config) logtailConn(ctx context.Context, referenceID string, since time.Time) (*websocket.Conn, error) {
 	u, err := url.Parse(c.LogsTailURL.String)
 	if err != nil {
 		return nil, fmt.Errorf("couldn't parse cloud logs host %w", err)
 	}
 
-	u.RawQuery = fmt.Sprintf(`query={test_run_id="%s"}&start=%d`,
-		referenceID,
-		time.Now().Add(-start).UnixNano(),
-	)
+	u.RawQuery = fmt.Sprintf(`query={test_run_id="%s"}&start=%d`, referenceID, since.UnixNano())
 
-	return u, nil
+	headers := make(http.Header)
+	headers.Add("Sec-WebSocket-Protocol", "token="+c.Token.String)
+
+	var conn *websocket.Conn
+	err = retry(sleeperFunc(time.Sleep), 3, 5*time.Second, 2*time.Minute, func() (err error) {
+		// We don't need to close the http body or use it for anything until we want to actually log
+		// what the server returned as body when it errors out
+		conn, _, err = websocket.DefaultDialer.DialContext(ctx, u.String(), headers) //nolint:bodyclose
+		return err
+	})
+	if err != nil {
+		return nil, err
+	}
+	return conn, nil
 }
 
 // StreamLogsToLogger streams the logs for the configured test to the provided logger until ctx is
 // Done or an error occurs.
 func (c *Config) StreamLogsToLogger(
-	ctx context.Context, logger logrus.FieldLogger, referenceID string, start time.Duration,
+	ctx context.Context, logger logrus.FieldLogger, referenceID string, tailFrom time.Duration,
 ) error {
-	u, err := c.getRequest(referenceID, start)
-	if err != nil {
-		return err
-	}
+	var mconn sync.Mutex
 
-	headers := make(http.Header)
-	headers.Add("Sec-WebSocket-Protocol", "token="+c.Token.String)
-
-	// We don't need to close the http body or use it for anything until we want to actually log
-	// what the server returned as body when it errors out
-	conn, _, err := websocket.DefaultDialer.DialContext(ctx, u.String(), headers) //nolint:bodyclose
+	conn, err := c.logtailConn(ctx, referenceID, time.Now().Add(-tailFrom))
 	if err != nil {
 		return err
 	}
 
 	go func() {
 		<-ctx.Done()
+
+		mconn.Lock()
+		defer mconn.Unlock()
 
 		_ = conn.WriteControl(
 			websocket.CloseMessage,
@@ -138,9 +161,9 @@ func (c *Config) StreamLogsToLogger(
 	}()
 
 	msgBuffer := make(chan []byte, 10)
-
 	defer close(msgBuffer)
 
+	var mostRecent int64
 	go func() {
 		for message := range msgBuffer {
 			var m msg
@@ -150,8 +173,8 @@ func (c *Config) StreamLogsToLogger(
 
 				continue
 			}
-
-			m.Log(logger)
+			ts := m.Log(logger)
+			atomic.StoreInt64(&mostRecent, ts)
 		}
 	}()
 
@@ -164,9 +187,34 @@ func (c *Config) StreamLogsToLogger(
 		}
 
 		if err != nil {
-			logger.WithError(err).Warn("error reading a message from the cloud")
+			logger.WithError(err).Warn("error reading a log message from the cloud, trying to establish a fresh connection with the logs service...") //nolint:lll
 
-			return err
+			var since time.Time
+			if ts := atomic.LoadInt64(&mostRecent); ts > 0 {
+				// add 1ns for avoid possible repetition
+				since = time.Unix(0, ts).Add(time.Nanosecond)
+			} else {
+				since = time.Now()
+			}
+
+			// TODO: avoid the "logical" race condition
+			// The case explained:
+			// * The msgBuffer consumer is slow
+			// * ReadMessage is fast and adds at least one more message in the buffer
+			// * An error is got in the meantime and the re-dialing procedure is tried
+			// * Then the latest timestamp used will not be the real latest received
+			// * because it is still waiting to be processed.
+			// In the case the connection will be restored then the first message will be a duplicate.
+			newconn, errd := c.logtailConn(ctx, referenceID, since)
+			if errd != nil {
+				// return the main error
+				return err
+			}
+
+			mconn.Lock()
+			conn = newconn
+			mconn.Unlock()
+			continue
 		}
 
 		select {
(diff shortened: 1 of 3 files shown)


cloudapi/logs.go
```diff
@@ -176,3 +176,186 @@ func (c *Config) StreamLogsToLogger(
...
+		websocket.FormatCloseMessage(websocket.CloseGoingAway, "closing"),
+		time.Now().Add(time.Second))
+
+	wsc.m.Lock()
+	_ = wsc.conn.Close()
+	wsc.m.Unlock()
+	return nil
+}
+
+// retry retries a to exeucute a provided function until it isn't succeeded
```
MStoykov, line 347:
```suggestion
// retry retries a to execute a provided function until it doesn't succeeded
```

MStoykov (commented):
I haven't looked over all the code but I have two notes from my fast going through it:
1. I think this would've been better handled by just handling an error on reading a message, instead of doing all of this additional changes that while great for a general usecase are both overkill and ... not enough as 
2. the URL we connect to has the parameter `start` which is from when loki should start returning logs. You reconnect to the same URL ... from what I can see which means that it will start sending logs from the very beginning of the test. This is both not necessary and heavy for loki and likely will not work as loki will only return only 1k+ old logs (I don't remember the exact value) and if I remember correctly it will return the *oldest* ones first ... which is totally not what we need.

So IMO this should've been an additional wrapping around for-loop reading messages that just on the correct error reconnects ... in the `if err !=nil` block but changing `start` to the last timestamp it has seen last.

cloudapi/logs.go
```diff
@@ -176,3 +176,186 @@ func (c *Config) StreamLogsToLogger(
...
+type dialConfig struct {
+	url           string
+	headers       http.Header
+	retryAttempts int64
+	retryInterval time.Duration
+	refreshAfter  time.Duration
+}
+
+// OnRedialing sets a function to invoke
+// when the a new connection refresh is on going.
```
imiric, line 253:
```suggestion
// OnRedialing sets a function that is called before an attempt is made to
// establish a new WS connection.
```

cloudapi/logs.go
```diff
@@ -176,3 +176,186 @@ func (c *Config) StreamLogsToLogger(
...
+
+	wsc.m.Lock()
+	_ = wsc.conn.Close()
+	wsc.m.Unlock()
+	return nil
+}
+
+// retry retries a to exeucute a provided function until it isn't succeeded
+// or maximum number of attempts is hit. It waits the specificed interval
+// between the latest iteration and the next retry.
```
imiric, line 349:
```suggestion
// retry executes a provided function until it returns an error or the
// maximum number of attempts is reached. It waits the specified interval
// between each retry.
```

cloudapi/logs.go
```diff
@@ -176,3 +176,186 @@ func (c *Config) StreamLogsToLogger(
...
+func retry(attempts uint, interval time.Duration, do func() error) (err error) {
+	for i := 0; i < int(attempts); i++ {
+		if i > 0 {
+			time.Sleep(interval)
+		}
+		err = do()
```
imiric, line 355: Could we do exponential backoff here to avoid flooding the backend with connection attempts?

cloudapi/logs.go
```diff
@@ -176,3 +176,186 @@ func (c *Config) StreamLogsToLogger(
...
+func (wsc *wsconn) keepAlive(ctx context.Context, refreshAfter time.Duration) {
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-wsc.afterfn(refreshAfter):
+			wsc.onRedialing()
+			// error skipped to let it fail on reading
+			// or retry a new dial on the next timer firing
+			_ = wsc.dial(ctx)
```
imiric, line 298: I'm wondering if we should handle keep alive with a read deadline instead, see [this comment](https://github.com/gorilla/websocket/issues/649#issuecomment-718186107). It seems like the suggested approach and a good idea to set it regardless to avoid connections hanging, or having to depend on `context` for it.

imiric (commented):
Just some minor comments.

`gomock` looks useful, but it pulls in a lot of dependencies, so I'd prefer it if we can avoid it and test with `httpmultibin` instead.

codebien:
Ready for a new review, simplified the solution and added the exponential backoff. I added them in two separate fixups so we can revert or change independently.

If the current code will be accepted then I will RS and remove the useless commits about `gomock`.

cloudapi/cloudapi_easyjson.go
```diff
@@ -17,7 +17,7 @@ var (
 	_ easyjson.Marshaler
 )
 
-func easyjsonC9ac27e5DecodeGithubComLoadimpactK6Cloudapi(in *jlexer.Lexer, out *msgStreams) {
```
MStoykov, line 20: This seems like a change due to the module rename and likely will be in other places.

Can you please make a separate PR updating it for ... everything and base this on it.

codebien: https://github.com/grafana/k6/pull/2097

cloudapi/logs.go
```diff
@@ -111,18 +113,26 @@ func (c *Config) getRequest(referenceID string, start time.Duration) (*url.URL,
...
-	u, err := c.getRequest(referenceID, start)
-	if err != nil {
-		return err
-	}
+	var conn *websocket.Conn
 
-	headers := make(http.Header)
-	headers.Add("Sec-WebSocket-Protocol", "token="+c.Token.String)
+	dial := func() error {
+		u, err := c.getRequest(referenceID, start)
```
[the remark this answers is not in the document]
codebien: `getRequest` is using `start` as a relative point from `Now`, in this way, when the dial function is invoked all the previous logs should be skipped. We are passing `0` from the cmd so, currently, the final query it's matching `Now`. 

https://github.com/grafana/k6/blob/master/cloudapi/logs.go#L101-L104

cloudapi/logs.go
```diff
@@ -176,3 +191,29 @@ func (c *Config) StreamLogsToLogger(
 		}
 	}
 }
+
+// retry retries to execute a provided function until it isn't succeeded
```
imiric, line 195:
```suggestion
// retry retries to execute a provided function until it isn't successful
```

cloudapi/logs.go
```diff
@@ -176,3 +191,29 @@ func (c *Config) StreamLogsToLogger(
...
+// or maximum number of attempts is hit. It waits the specified interval
+// between the latest iteration and the next retry.
+// Interval is used as the base to compute an exponential backoff until,
```
imiric, line 198: Incomplete sentence?

cloudapi/logs.go
```diff
@@ -111,18 +113,26 @@ func (c *Config) getRequest(referenceID string, start time.Duration) (*url.URL,
...
+			// We don't need to close the http body or use it for anything until we want to actually log
+			// what the server returned as body when it errors out
+			conn, _, err = websocket.DefaultDialer.DialContext(ctx, u.String(), headers) //nolint:bodyclose
+			return err
+		})
+	}
+
+	if err := dial(); err != nil {
 		return err
 	}
```
imiric, line 137: I would move everything until here to a `getConnection()` function.

imiric (commented):
Mostly LGTM. I'll take another pass once it's rebased and ready for merging.

Some notes:

- You're planning to remove `gomock` and add tests using `httpmultibin`? :+1:
- Nit and maybe unrelated to this PR, but `Config.getRequest()` parses the same `LogsTailURL` on each call:
https://github.com/grafana/k6/blob/4c5931af8f7b8645b5f6e527e897eeeaeef176d8/cloudapi/logs.go#L97-L101
  We could probably do this once and store the `url.URL`, though nevermind if it's difficult because of #883. :sweat_smile:

cloudapi/logs.go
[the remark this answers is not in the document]
MStoykov: You are correct. Sorry, I was confused by the name of the parameter (which is also my fault given that I named it ;) ).

Given that though it means that we won't even try to get the messages between when we disconnect and when we finally reconnect. 
So we either:
1. Rename the parameter to be better named and less confusing and open an issue to fix the above later
2. Fix this now so that `start` is the actual parameter `start` to the loki [endpoint](https://grafana.com/docs/loki/latest/api/#get-lokiapiv1tail)  which is the time from which loki should start returning logs

cloudapi/logs.go
```diff
@@ -166,6 +173,18 @@ func (c *Config) StreamLogsToLogger(
 		if err != nil {
 			logger.WithError(err).Warn("error reading a message from the cloud")
 
+			// try to restore the stream establishing a new connection
+			logger.Warn("trying to establish a fresh connection with the tail logs, this might result in either some repeated or missed messages") //nolint:lll
```
codebien, line 177: I think we could remove or rephrase it, because with the current solution `some repeated or missed messages` shouldn't really happen.

na--: why not combine both warnings? something like this
```go
logger.WithError(err).Warn("error reading a log message from the cloud, trying to establish a fresh connection with the logs service...")
```

cloudapi/logs.go
```diff
@@ -166,6 +173,18 @@ func (c *Config) StreamLogsToLogger(
...
+
+			newconn, errd := c.logtailConn(ctx, referenceID, time.Now())
```
na--, line 179: given how time can drift due to network latency and be different between the client and the server, we probably should use the time of the last received log message (`+1ms` to avoid repetition) here. We can only fall back to `time.Now()` if we haven't received any log messages so far

codebien:
Updated, I'm not so happy with the integration tests' complexity. I will try to polish for simplifying them when the core solution will be approved.

cloudapi/logs.go
```diff
@@ -47,6 +51,23 @@ type msgStreams struct {
...
+		if err != nil {
+			return
+		}
+		if unix > ts {
+			ts = unix
+		}
+	}
+	return
+}
+
```
MStoykov, line 70: Loki should always (per definition) retunr the logs in *order* so in our case if I remember correctly the last message is always *the* last message and as such with the latest timestamp

cloudapi/logs.go
```diff
@@ -150,8 +176,16 @@ func (c *Config) StreamLogsToLogger(
...
-
 			m.Log(logger)
+
+			var ts int64
+			for _, stream := range m.Streams {
+				sts := stream.LatestTimestamp()
+				if sts > ts {
+					ts = sts
+				}
+			}
```
MStoykov, line 187: I would argue you technically need to look ad the droppedEntries as well, but I don't think it matters all that much as that basically means we got *a lot of logs* and we loki couldn't send them all (or we couldn't read fast enough) and so even if we look at them and choose a different timestamp we will still probably be dropping a lot of logs, so I think just checking the Streams is fine for now

cloudapi/logs.go
```diff
@@ -164,8 +198,17 @@ func (c *Config) StreamLogsToLogger(
 		}
 
 		if err != nil {
-			logger.WithError(err).Warn("error reading a message from the cloud")
+			logger.WithError(err).Warn("error reading a log message from the cloud, trying to establish a fresh connection with the logs service...") //nolint:lll
 
+			newconn, errd := c.logtailConn(ctx, referenceID, latest.TimeOrNow())
```
MStoykov, line 203: I would argue that now the whole readMessage, send on channel, set latest timestamp based on it is racy (logically if not programmatically given the inclusion of atomics now).

I doubt it will make a difference where we get messages fast enough that we can't log them until we get the next one and then a third one is an error, so maybe not do anything, but add a comment saying that this *non* exact correctness is on purpose?

imiric (commented):
LGTM, just a minor comment.

Nice work on the tests, `cloudapi/logs.go` is 90% covered :thumbsup: Not sure how we could simplify them, it's readable as is I think.

cloudapi/logs.go
codebien: >For any single log stream, logs must always be sent in increasing time order. If a log is received with a timestamp older than the most recent log received for that stream, that log will be dropped.

Apparently, you're right. In the case, the client (k6) is sending it out of order then loki will drop it. So we can safely consider the last item in the slice as the latest.

cloudapi/logs.go
codebien: Should I consider adding a comment regarding it?

MStoykov: yeah, let's add a comment

cloudapi/logs.go
MStoykov: Not only that but tail specifically returns them ordered as otherwise logging them will be pretty bad ;) .


 Not that we currently interleave log messages from different streams when we logged them on the screen .... as I've just noticed :facepalm:

codebien:
Updated with the suggested request changes

cloudapi/logs.go
codebien: I added the comment but I would like to have a 100% correct solution but I didn't yet find one that can maintain the code simple as-is. If you have suggestions they are welcome.

cloudapi/logs.go
```diff
@@ -164,8 +199,25 @@ func (c *Config) StreamLogsToLogger(
...
+			if errd == nil {
+				mconn.Lock()
+				conn = newconn
+				mconn.Unlock()
+				continue
+			}
 
+			// return the main error
 			return err
```
MStoykov, line 221: nit: I would've done `errd!=nil` return `err` and than skip the `continue` but also keep the good case with smaller identation

cloudapi/logs.go
```diff
@@ -176,3 +228,69 @@ func (c *Config) StreamLogsToLogger(
...
+		t = time.Unix(0, ts).Add(1 * time.Millisecond)
+	}
+	return
+}
+
+// Set sets the tracked timestamp value.
+func (tst *timestampTrack) Set(ts int64) {
+	if ts < 1 {
+		return
+	}
```
MStoykov, line 254: I kind of feel this isn't needed ... if we are logging you have a `ts` bigger than 1 ;)

cloudapi/logs.go
```diff
@@ -176,3 +228,69 @@ func (c *Config) StreamLogsToLogger(
...
+}
+
+// TimeOrNow returns as Time the latest tracked value plus 1ms
+// or Now as the default value.
+func (tst *timestampTrack) TimeOrNow() (t time.Time) {
+	t = time.Now()
+	if ts := atomic.LoadInt64(&tst.ts); ts > 0 {
+		// add 1ms for avoid possible repetition
+		t = time.Unix(0, ts).Add(1 * time.Millisecond)
+	}
```
MStoykov, line 246: nit: I feel like this will be better of inline in the place it's used.

Currently, the whole timestampTrack is a 1 int64 that we atomically change under the hood and here you have some business logic on top. Given that it is literally set in 1 place and read in 1 place I would prefer if the business logic is more inlined and it's just that int64 that we atomically read and write, instead of it being something that I now need to go read that `TImeOrNow` (which I guess should be `ValueOrTImeNow`?) also adds 1 ms in one the cases.

p.s. also probably better to add 1 nanosecond just in case :rofl: . BUt I still feel this is better of in the place we are using it.

MStoykov (dismissed):
I have some nitpicks that can be fixed later, but it looks ready either way, good job :clap: .

I kind of wonder if the 0, 5s,25s,2m isn't a bit too much of a backing off, but hopefully nobody will be hitting this ... that hard ;) 

I still have to test it locally, somehow though so if anybody has some good suggestions , I am welcoming them ;)

cloudapi/logs.go
codebien: >then skip the continue

I reversed the check having the classic `!= nil` but I'm not sure I got how you would change to avoid the continue.

codebien:
New changes are ready for review.

MStoykov (dismissed):

na-- (dismissed):
The code LGTM, though I haven't tested this in practice, plan to do it after we merge in `master`.

na-- (dismissed):

MStoykov (approved):

na-- (approved):

Issue #2091, opened 2021-07-13
Support NO_COLOR


MStoykov:
There apparently is now somewhat of a "standard" way to disable colors for terminal applications https://no-color.org/, it appears to be fairly popular given the list of things supporting it and fatih/color added [support](https://github.com/fatih/color/pull/137), which is how I found out. 
## Feature Description
```
NO_COLOR= k6 run script.js
```

should not output any color codes
## Suggested Solution (optional)
Unfortunately, the above fatih/color update isn't enough as we check ourselves whether we should output color or not. Including `--no-color` flag which likely just needs to get the env variable source configured ?

PR #2092, opened 2021-07-13
Update libs

MStoykov:


na-- (commented):
LGTM, though can you also update the `golang.org/x/` repos, so that @codebien doesn't have to in https://github.com/k6io/k6/pull/2090? Ideally, we shouldn't mix dependency updates with feature PRs

MStoykov:
I did update them to the versions that are in the current 1.16 vendored branch. There are changes that are for go 1.17 that I haven't update to.

MStoykov:
an by a "did update" I mean I did run the commands to do and as nothing happened I checked and there are *no changes* in the `internal-branch.go1.16-vendor` branches for neither x/crypto nor x/net

na-- (dismissed):
Huh, I guess `github.com/golang/mock@v1.6.0` depend on the `master` versions then? :confused:

MStoykov:
it depends on x/tools which [does](https://cs.opensource.google/go/x/tools/+/master:go.mod)

imiric (dismissed):
LGTM, just a couple of questions:

- Why these specific libs, or are these all that are outdated?
- Did you use `modtools` for this?

MStoykov:
@imiric

* those were the ones outdated which either had something useful updated and didn't require more work or update x/net and co to their latest master (which is still unreleased for go 1.17). But maybe I should update the remaining 2-3 that had very small inconsequential updates as well just to remove them from the list of ones that have updates :thinking: . I would've made additional PRs for the more problematic updates such as `cobra` 
* I used `modtools check` but not `modtools update` as it will update everything at the same time and I want to do it one by one ;)

vendor/github.com/fatih/color/color.go
```diff
@@ -108,7 +116,14 @@ const (
...
 // New returns a newly created color object.
 func New(value ...Attribute) *Color {
-	c := &Color{params: make([]Attribute, 0)}
+	c := &Color{
+		params: make([]Attribute, 0),
+	}
+
+	if noColorExists() {
+		c.noColor = boolPtr(true)
+	}
```
na--, line 125: Honestly, the fact that they touch `os` and the fact that they don't really give you a choice whether you want to respect `NO_COLOR` or not is more than a bit shitty... I get that it makes the library API slightly easier to use, and it "just works" for the majority of users, but it makes testing things that use this library (like k6) such a PITA... :sob: 

At this point, I'd happily remove it and replace it with something else, since we moved the end-of-test summary to JS, we use it very sparingly:
```
rg -g '!vendor' -g '!go.mod' -g '!go.sum' 'fatih/color|color\.'
ui/form.go
27:	"github.com/fatih/color"
50:		if _, err := fmt.Fprintln(w, color.BlueString(f.Banner)+"\n"); err != nil {
60:				displayLabel += " " + color.New(color.Faint, color.FgCyan).Sprint("["+extra+"]")
66:			color.Set(color.FgCyan)
73:			color.Unset()
80:				if _, printErr := fmt.Fprintln(w, color.RedString("- "+err.Error())); printErr != nil {

ui/pb/progressbar.go
28:	"github.com/fatih/color"
34:	colorFaint   = color.New(color.Faint)
35:	statusColors = map[Status]*color.Color{
36:		Interrupted: color.New(color.FgRed),
37:		Done:        color.New(color.FgGreen),

cmd/cloud.go
37:	"github.com/fatih/color"
333:			valueColor := getColor(noColor || !stdoutTTY, color.FgCyan)

cmd/ui.go
35:	"github.com/fatih/color"
89:func getColor(noColor bool, attributes ...color.Attribute) *color.Color {
91:		c := color.New()
96:	c := color.New(attributes...)
102:	c := getColor(noColor, color.FgCyan)
140:	valueColor := getColor(noColor, color.FgCyan)

cmd/login_cloud.go
29:	"github.com/fatih/color"
145:				valueColor := getColor(noColor || !stdoutTTY, color.FgCyan)
```

Not now, but the next time we do something with colors in k6...

na-- (approved):

vendor/github.com/fatih/color/color.go
MStoykov: I did test and it definitely doesn't work ... hence #2091 was made ;) So this update has no consequences for us ... it appears. I am pretty sure it's because of https://github.com/k6io/k6/blob/04c8b1fcf9a7e8628c1200603d433b404df241ba/cmd/ui.go#L86-L99

na--: Glad to see my workaround continues to work around the issue :tada:  :sweat_smile: True, we won't support `NO_COLOR` until we explicitly add support for it, but I'd still prefer that approach :man_shrugging:

na-- (approved):

imiric (approved):

PR #2093, opened 2021-07-13
Add JS function to abort test

### LINKED ISSUE #1001: Provide an API to exit a test

imiric:
These are the changes from #1920, squashed and rebased on `master`, while changing the function from `k6.abortTest()` to the `k6/execution` module as `test.abort()`. Additionally a new `ExecutionStatusInterrupted` was added, a separate exit code (108), and a few other fixes and minor changes. See the commits for details.

Closes #1001


cmd/run.go
```diff
@@ -270,8 +275,18 @@ a commandline interface for interacting with it.`,
...
+			err = engineRun()
+			if err != nil {
+				if common.IsInterruptError(err) {
+					interrupt = err
+				}
+				if !conf.Linger.Bool {
+					if interrupt == nil {
+						return errext.WithExitCodeIfNone(err, exitcodes.GenericEngine)
+					}
+					return errext.WithExitCodeIfNone(interrupt, exitcodes.ScriptException)
```
imiric, line 288: The current behavior (and in #1920) is to exit the script immediately without printing the test summary. Shouldn't we do a clean exit and print the test summary instead? This is already the behavior when `--linger` is used, so it would be consistent.

imiric:
Lint errors are from https://github.com/k6io/k6/pull/2073, so ignore :)

And there are currently no tests for this, since they were all at the JS level and were moved to the extension (mostly). Should we add some lower level tests here?

js/common/interrupt_error.go
```diff
@@ -0,0 +1,44 @@
...
+
+// InterruptError is an error that halts engine execution
+type InterruptError struct {
+	Reason string
+}
+
+func (i InterruptError) Error() string {
+	return i.Reason
+}
+
```
MStoykov, line 33: This will probably be better off as a pointer receiver because otherwise you somewhere might return non pointer to it and it will *still* count as error but will not pass the `IsInterruptError` below. While in order to "break" it with pointer receiver you will need to return pointer to a pointer to InterruptError which ... seems like a bad idea in the first place 

[playground link trying to showcase the problem](https://play.golang.org/p/JUgsHN_hpdp)

MStoykov (commented):
I didn't understand why you need to use my WIP PRs to test it? 

Apart from that and the error comment it seems okay to me

js/common/interrupt_error.go
imiric: Good point, thanks for explaining. Will fix.

imiric:
> I didn't understand why you need to use my WIP PRs to test it?

Because `exec.test.abort()` is based on https://github.com/grafana/xk6-execution/pull/2, which depends on #2073, so I needed both your changes and these core changes to test the extension. Your commits just don't need to be merged here, so I'll remove them once this is approved, and then you can rebase on `master` and I'll point the extension PR to your branch instead.

js/common/interrupt_error.go
imiric: Resolved in recent push.

cmd/run.go
MStoykov: Hm ... I guess it's better if we are consistent and it seems to just drop this return ? (and refactor the if's into one I guess)

MStoykov (approved):
LGTM apart from the possible change to showing the summary

na-- (approved):
LGTM in principle, but I didn't review this very deeply, mostly relied on my memories from reviewing the old PR :sweat_smile:

cmd/run.go
imiric: I'm weary of doing changes in `cmd`, so hopefully 2381b70c didn't break anything. Let me know.

The only slight nitpick would be that the error is shown _after_ the summary and not before, but I think we can live with that.

cmd/run.go
```diff
@@ -331,16 +319,24 @@ a commandline interface for interacting with it.`,
...
 			engineWait()
 			logger.Debug("Everything has finished, exiting k6!")
-			if interrupt != nil {
-				return errext.WithExitCodeIfNone(interrupt, exitcodes.ScriptException)
+
+			if runErr != nil {
+				if common.IsInterruptError(runErr) {
+					return errext.WithExitCodeIfNone(runErr, exitcodes.ScriptException)
+				}
+				return errext.WithExitCodeIfNone(runErr, exitcodes.GenericEngine)
```
MStoykov, line 332: this also changes that we will show the summary on *non* InterruptError which probably is not what we want

imiric: No? I did that intentionally as I don't see why we shouldn't show the summary in all cases.

MStoykov: I see some merit to showing the summary in more cases but I think this is kind of a bigger and different change so I would prefer to have it outside of this PR 

cc @na--

imiric: I removed this in the latest force push, so now only interrupt errors should show the summary.

js/modules/k6/execution/execution.go
```diff
@@ -71,8 +72,9 @@ func (*RootModule) NewModuleInstance(m modules.InstanceCore) modules.Instance {
 			common.Throw(rt, err)
 		}
 	}
-	defProp("scenario", mi.newScenarioInfo)
 	defProp("instance", mi.newInstanceInfo)
+	defProp("scenario", mi.newScenarioInfo)
+	defProp("test", mi.newTestInfo)
```
imiric, line 77: I exposed this as `exec.test.abort()` as it was before the `test`/`instance` rename. I think it makes sense to have this under the global "test" property, since it will abort the entire test and not just an instance in the cloud.

MStoykov:
This needs to be rebased

cmd/run_test.go
```diff
@@ -125,3 +132,48 @@ func TestHandleSummaryResultError(t *testing.T) {
 	assertEqual(t, "file summary 1", files[filePath1])
 	assertEqual(t, "file summary 2", files[filePath2])
 }
+
+func TestAbortTest(t *testing.T) { //nolint: tparallel
```
MStoykov, line 136: Is this `nolint` required :thinking:

cmd/run_test.go
```diff
@@ -125,3 +132,48 @@ func TestHandleSummaryResultError(t *testing.T) {
...
+	t.Parallel()
+
+	t.Run("Check status code is 107", func(t *testing.T) { //nolint: paralleltest
+		ctx, cancel := context.WithCancel(context.Background())
+		defer cancel()
+		logger := testutils.NewLogger(t)
```
MStoykov, line 142: Does this touch some global and that is why it can't be paralleled?

MStoykov (approved):
LGTM! I haven't retested it but I remember previously `--linger` stopped it from exiting so if you can just retest it I think it's fine

cmd/run_test.go
imiric: Yes, it seems to be an issue with `cmd.Flags()`.

MStoykov (approved):

yorugac (dismissed):
LGTM

A note about docs: it might be good to warn users about what they pass to `abort` so that they don't see smth like this:
```
test aborted: [object Object]
```
:sweat_smile:

cmd/run.go
```diff
@@ -323,6 +334,9 @@ a commandline interface for interacting with it.`,
 			logger.Debug("Waiting for engine processes to finish...")
 			engineWait()
 			logger.Debug("Everything has finished, exiting k6!")
+			if interrupt != nil {
+				return errext.WithExitCodeIfNone(interrupt, exitcodes.ScriptException)
```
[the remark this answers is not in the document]
imiric: To clarify: you're suggesting to use a different name for exit code `107` or a different exit code for when the test is aborted by the script? Changing the name is fine, but changing the exit code will require also changing k6agent to handle it properly, and I don't see why we should since `107` is only used for this case when the test is interrupted.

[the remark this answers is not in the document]
MStoykov: I am proposing that  `AbortedByScriptErrorCode` (name subject to change) is added and have it have a *new* exit code so we can differentiate between some using `abort()` and there is an exception in the init context

core/local/local.go
```diff
@@ -346,7 +347,14 @@ func (e *ExecutionScheduler) Run(
 	executorsCount := len(e.executors)
 	logger := e.logger.WithField("phase", "local-execution-scheduler-run")
 	e.initProgress.Modify(pb.WithConstLeft("Run"))
-	defer e.state.MarkEnded()
+	var interrupted bool
+	defer func() {
+		if interrupted {
+			e.state.SetExecutionStatus(lib.ExecutionStatusInterrupted)
+		} else {
+			e.state.MarkEnded()
```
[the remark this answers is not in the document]
imiric: Done in recent push.

core/local/local.go
```diff
@@ -414,7 +428,10 @@ func (e *ExecutionScheduler) Run(
 			return err
 		}
 	}
-
+	if err := executor.CancelReason(execCtx); err != nil {
```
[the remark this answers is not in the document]
imiric: Sure, done in recent push.

js/modules/k6/execution/execution.go
```diff
@@ -170,6 +172,31 @@ func (mi *ModuleInstance) newInstanceInfo() (*goja.Object, error) {
 	return newInfoObj(rt, ti)
 }
 
+// newTestInfo returns a goja.Object with property accessors to retrieve
+// information and control execution of the overall test run.
+func (mi *ModuleInstance) newTestInfo() (*goja.Object, error) {
+	ctx := mi.GetContext()
+	rt := common.GetRuntime(ctx)
+	if rt == nil {
+		return nil, errors.New("goja runtime is nil in context")
```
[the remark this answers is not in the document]
imiric: `exec.test.abort()` does work in the init context, see `TestInitContextForbidden/abortTest` (maybe the wrong parent test for this). It will just exit with code 255 instead of 107 right now. Should we fix that here or in a separate PR?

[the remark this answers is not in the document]
MStoykov: Why does this check exist? I am pretty sure there is no place in which this is possible?

cmd/run.go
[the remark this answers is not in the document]
imiric: I settled on `ScriptAborted` (108) in d46a401a, and simplified some things in `cmd/run.go`. Let me know.

js/modules/k6/execution/execution.go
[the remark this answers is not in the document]
imiric: @MStoykov I suppose extra precaution against a panic :smile: But if it's not possible, I removed it in fb517cef.

imiric:
CLA Assistant is bugged again :disappointed: I had already approved it and did it again earlier today on `https://cla-assistant.io/grafana/k6?pullRequest=2093`, but it's still stuck on pending. :-/ I guess it's not a merge requirement :man_shrugging:

BTW, WDYT about renaming `InterruptError` to `AbortedError` and similarly in all other cases? The original PR used "interrupt" everywhere, but in some places we use "abort" and the main API is `test.abort()`, so I think it makes sense to standardize it in the code as well.

Other than that this is ready for review again. @na-- @MStoykov

I'm currently also adding integration tests for this to hopefully get to the bottom of why it doesn't seem to work for cloud tests in production.

cmd/run_test.go
MStoykov: I think you can just add `t.Parallel` to the ones below? Or just split them in two separate tests?

MStoykov (approved):
LGTM

cmd/run_test.go
imiric: Sorry, missed this originally.

To answer your first question: yes, it's needed, otherwise `tparallel` fails with `TestAbortTest's subtests should call t.Parallel`. And both `tparallel` and `paralleltest` are enabled by default... It's confusing, I know :disappointed: Maybe we should disable `tparallel` globally?

We can't add `t.Parallel()` to either test because of several data races... And I prefer having them as subtests in this case. :)

imiric:
@MStoykov @na-- Take a look at the fix in 0a75c41b. Previously for test `abort_initerr.js` we used to exit with code 255, and with `abort_initvu.js` with 103. This ensures we exit with the new 108 code in all cases.

cmd/run_test.go
```diff
@@ -125,3 +134,85 @@ func TestHandleSummaryResultError(t *testing.T) {
...
+		{
+			testFilename: "abort_initerr.js",
+		},
+		{
+			testFilename: "abort_initvu.js",
+		},
+		{
+			testFilename: "abort_teardown.js",
+			expLogOutput: "Calling teardown function after test.abort()",
+		},
```
MStoykov, line 156: nitpick: for these really small testfiles I kind of feel like moving them to files just adds extra steps for whoever looks at the code. AFAIK the reason for this to be the case for other tests is that the testdata is quite ... big and hard to put in a string in the code, none of this is true for this couple of lines here.

core/local/local.go
```diff
@@ -346,7 +348,13 @@ func (e *ExecutionScheduler) Run(
 	executorsCount := len(e.executors)
 	logger := e.logger.WithField("phase", "local-execution-scheduler-run")
 	e.initProgress.Modify(pb.WithConstLeft("Run"))
-	defer e.state.MarkEnded()
+	var interrupted bool
+	defer func() {
+		e.state.MarkEnded()
+		if interrupted {
+			e.state.SetExecutionStatus(lib.ExecutionStatusInterrupted)
+		}
```
MStoykov, line 356: can we get a comment that we on purpose first hit the `ended` and then the `interrupted` state

MStoykov (approved):
LGTM to me, I have some nitpicks, but otherwise :clap:

cmd/run_test.go
imiric: I agree, but these tests test the `runCmd`, so I can't see how we could turn the JS into strings... Writing to `os.Stdin` maybe? :confused:

core/local/local.go
imiric: :man_shrugging: Not sure, doesn't the code speak for itself?

imiric:
@MStoykov Thanks for yet another review of this :sweat_smile:

Can you please comment on my renaming suggestion [here](https://github.com/grafana/k6/pull/2093#issuecomment-977803462)?

And you probably saw it, but do you think the small refactor in 986bbf2e is safe? I tested it manually with a few scripts, and it set the expected `RunStatus`, but I'm not very confident that it won't mess up something else. Should I add some tests for it?

MStoykov (commented):
The rewrite seems fine, I am not certain I like it more, looking at it again as I don't see anywhere where you set `rs` to one value and then set it to another later on (apart from the fact it has a default value that is being used in only 1 of the cases :thinking: ) - so as a nitpick, I guess I prefer it to not use a defer and overwriting a variable, but it's safe AFAIK 

`Aborted` is way too general, but also you do call `abort` ... :thinking: . Maybe we should rename it not be `abort()` but something else? `interrupt()` doesn't sound better at all. 
`AbortedByAbortError` is abysmal. I kind of feel like `InterruptedError` was still bad as it sounds like something outside interrupted the test, but `Aborted` is just so general that I feel that it needs to be `AbortedByScript`

na-- (changes requested):
Mostly LGTM, with only the problem we discussed in Slack that I think needs to be fixed. The new `ExecutionStatusInterrupted` status and `ScriptAborted` exit code are fine, but I think we shouldn't also have a new `RunStatus`...

Rather, `RunStatusAbortedScriptError` shouldn't be renamed and we should reuse `RunStatusAbortedUser` for this, at least for now. We can add a new RunStatus in the future, if needs be.

na--:
Forgot to mention my thoughts on the method name. For me, `execution.test.abort()` seems good enough, or at least I can't think of anything better :man_shrugging:

na-- (approved):
LGTM :crossed_fingers:

codebien:
@imiric considering all the evolutions, I would prefer to have a cleaned commit history (e.g. the first commit c2fe0140fa8efc725cb4e219f851960441fe59ef is very detailed so it's important IMHO but it's not up to date). Ideally, this PR should be rebased for avoiding the squash before being merged. If we are not confident about doing it, can you please update the PR's description or add a comment with the summary with introduced features/changes?

It would be helpful in the case we will need to recheck/share this PR to have a central place instead to re-read the description, comments, commits and code.

imiric:
@codebien I wanted to keep the original commit mostly intact, as it was authored by gernest. The changes on top of it are mostly atomic new features and shouldn't be squashed, but I can rebase it on current `master` and clean it up a bit. I'll update the PR description as well.

na-- (approved):

codebien (approved):
Should we add  for this?

imiric:
5 months later and almost 9 months since #1920, we're finally merging this! :sweat_smile: It's a Christmas miracle! :smile::christmas_tree:

Thanks for the reviews everyone! I'll try to add some documentation for this next week.

na--:
He signed the old CLA in https://github.com/grafana/k6/pull/1920#issuecomment-803851948, which is basically identical to the current one for k6 (the new one only expands to cover other xk6 projects), so it's fine to merge as it is.

jackgoodacre-crypto:
Hey all, 

Have found my way onto this PR as a result of looking for the ability to abort k6 execution with non-zero exit code. I appreciate the efforts required to develop these changes, so thank you for building on top of suggestions/requests coming from users.

I was just wondering, although this functionality is merged to master I don't believe it is available in an official version yet right? If not, are there currently any estimations as to when this may occur?

Thanks again

PR #2097, opened 2021-07-20
Full code generation

codebien:
Regenerated the full codebase. 

Using:
* enumer: github.com/dmarkham/enumer 1.5.2
* easyjson: github.com/mailru/easyjson 0.7.7


imiric (approved):

MStoykov (commented):
Not really happy with all the new "lower case" code ... I am not even certain that I want case-insensitivity of options :thinking:

codebien:
Re-generated the code using https://github.com/alvaroloes/enumer

MStoykov (dismissed):

imiric (approved):

na-- (approved):

na-- (changes requested):
Hmm actually, we should probably add a CI check for this :thinking: It can be something like this: https://github.com/grafana/k6/blob/a33b054344416f32a5034732be1da06018a682bc/.github/workflows/all.yml#L31-L35

Even part of the same CI job :man_shrugging: Basically, install the easyjson and enumer tools, run `go generate ./...` and verify that not a single file has changed

codebien:
Added a new step to the `deps` check for code generation

codebien:
The previous version seems didn't support untracked, the latest fixed.

.github/workflows/all.yml
```diff
@@ -33,6 +33,12 @@ jobs:
             go version
             test -z "$(go mod vendor && git status --porcelain)"
             go mod verify
+      - name: Check code generation
+        run: |
+          go install github.com/alvaroloes/enumer@v1.1.2
+          go install github.com/mailru/easyjson/easyjson@v0.7.7
+          go generate ./...
+          exit $(git status --porcelain 2>/dev/null | wc -l)
```
na--, line 41:
```suggestion
          exit $(git status --porcelain 2>/dev/null | wc -c)
```

If you have single line without a `\n` at the end, `wc -l` will result in `0`, try `echo -n foo | wc -l`

na--: btw what was the problem with `test -z`, doesn't it do basically the same thing? :confused:

codebien: I was researching a solution for removing `test -z` how @MStoykov requested but in the end, I lost the focus. Yep, I will bring back the `test -z` to have consistency.

MStoykov: I think this is because I asked "do we need test -z?" but I asked that as I was left with the impressions `test -z` tests whether the returned exit status is `0` which it definitely doesn't :facepalm:  and even if it did `git status --porcelain` won't exit with non zero status it will just not print some stuff on the output.

 original [comment](https://github.com/grafana/k6/commit/da37f41e438b16353c381912bb8c91fa2b75b0ee#r54025705) ... which I think is on a commit :facepalm:  oh ... how I love github and its emails ...

codebien:
The `test -z` is restored

imiric (approved):

na-- (approved):

PR #2099, opened 2021-07-24
Fix cloud config merging

MStoykov:
Previously after the usual consolidation of metrics where json config,
script options, cli, env order is followed the `ext.loadimpact` configs
(from within the script options) would take precedence over anything
else.

Now it has the same precedence as the script config.

This likely as always has additional sideeffects ... for example the
K6CLOUD_TOKEN to K6_CLOUD_TOKEN fix is being done after all of this
which should probably be dropped either way.

### DIFF
--- a/cloudapi/config_test.go
+++ b/cloudapi/config_test.go
@@ -44,7 +44,6 @@ func TestConfigApply(t *testing.T) {
 
 	full := Config{
 		Token:                           null.NewString("Token", true),
-		DeprecatedToken:                 null.NewString("DeprecatedToken", true),
 		ProjectID:                       null.NewInt(1, true),
 		Name:                            null.NewString("Name", true),
 		Host:                            null.NewString("Host", true),
--- a/cloudapi/config_test.go
+++ b/cloudapi/config_test.go
@@ -20,10 +20,13 @@
 package cloudapi
 
 import (
+	"encoding/json"
+	"os"
 	"testing"
 	"time"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	"gopkg.in/guregu/null.v3"
 
 	"go.k6.io/k6/lib/types"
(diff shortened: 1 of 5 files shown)


na-- (approved):
LGTM in principle, but we need to test all of the components very, **very** carefully, with various combinations of env/JS/JSON options, since the `cmd` parts don't really have any good unit tests:
 - `k6 cloud script.js` 
 - `k6 cloud bundle.tar`
 - `k6 run --out cloud script.js`
 - `k6 run --out cloud bundle.tar`
 - `k6 login cloud`

MStoykov:
I tested the following 

- [x]  k6 login cloud (with and without config)

The following have been tested with having `name` and a non standard field in the `ext.loadimpact`( that we just expect to keep being in the metadata.json), and either resetting or not resetting the `name` through env variables. I
also changed the token in the middle of the tests and reran them with overwriting and not overwriting the token.

- [x] k6 run -o cloud script.js - works for everything
- [x] k6 run -o cloud archive.tar - works for everything overwriting stuff accordingly in the cloud. 
- [ ] k6 cloud script.js - works, but doesn't overwrite stuff in the `metadata.json` for values coming from `env`. But the updated values are being used for the actual execution from the cloud so ... kind of works :shrug: .
- [ ] k6 cloud archive.tar - has the same problem, but unless you overwrite with `K6_CLOUD_TOKEN` it won't even work as it will use the one in the archive(this last part is actually working correctly, and was not working previously).

Both of the problems with `k6 cloud` are present in old versions so I am somewhat for merging this and making a separate fix for it.

MStoykov:
https://github.com/grafana/k6/pull/2099/commits/d535634339bc3fff247bb76f84da3ea69ba16bd3 seems to fix the `k6 cloud` issues I noticed completely so ... :crossed_fingers:

imiric (approved):
LGTM. Basic `cloud` and `run -o cloud` tests worked fine, but I was able to reproduce the issue in Slack after renewing the API token, so the real test will be with a canary.

cloudapi/config_test.go
```diff
@@ -70,3 +72,20 @@ func TestConfigApply(t *testing.T) {
 	assert.Equal(t, full, empty.Apply(full))
 	assert.Equal(t, full, defaults.Apply(full))
 }
+
+func TestConfigConsolidation(t *testing.T) { //nolint:paralleltest
```
codebien, line 76:
```suggestion
func TestGetConsolidatedConfig(t *testing.T) { //nolint:paralleltest
```

cmd/cloud.go
```diff
@@ -164,20 +155,26 @@ This will execute the test on the k6 cloud service. Use "k6 login cloud" to auth
...
 
-			if err = cloudapi.MergeFromExternal(arc.Options.External, &cloudConfig); err != nil {
+			// Cloud config
+			cloudConfig, err := cloudapi.GetConsolidatedConfig(
+				derivedConf.Collectors["cloud"], osEnvironment, "", arc.Options.External)
+			if err != nil {
 				return err
 			}
+			if !cloudConfig.Token.Valid {
+				return errors.New("Not logged in, please use `k6 login cloud`.") //nolint:golint,revive,stylecheck
```
codebien, line 165: Is expected to have a capital letter?

codebien (changes requested):
Just tiny things, basic examples that I know work for me

cmd/cloud.go
MStoykov: It's a message that gets shown to the user, so it should probably be a full sentence. 

Also, this is a problem across the codebase so if you want to change that, please open a separate issue ;)

codebien (approved):

na-- (approved):
LGTM :crossed_fingers:

PR #2101, opened 2021-07-28
Redo how modules are given access to context.Context and co.

MStoykov:
merged the metricRegistry to try to fix some other issues


js/common/bridge.go
```diff
@@ -280,3 +280,72 @@ func Bind(rt *goja.Runtime, v interface{}, ctxPtr *context.Context) map[string]i
...
+		result[i] = T.Method(i).Name
+	}
+
+	return result
+}
+
+// ModuleInstanceCore is something that will be provided to modules and they need to embed it in ModuleInstance
+type ModuleInstanceCore interface {
+	// we can add other methods here
+	// sealing field will help probably with pointing users that they just need to embed this in the
```
imiric, line 306: The first line is superfluous :), the second incomplete.

js/modules/modules.go
```diff
@@ -69,6 +70,11 @@ type HasModuleInstancePerVU interface {
 	NewModuleInstancePerVU() interface{}
 }
 
+// IsModuleV2 ... TODO better name
+type IsModuleV2 interface { // TODO rename?
```
imiric, line 74: Just `Module`? Or do we want to keep supporting `common.Bind()`?

imiric (commented):
LGTM, don't have any major comments. I can make this work for `k6/x/execution`.

codebien (changes requested):
Not big catches, I would like just fix the linter and rephrase the commit message of e69a552c02292e41134b841d722c4460bdf3f9d1

js/modules/modules.go
MStoykov: I would prefer to keep supporting common.BInd for at least 1 more release ... especially as otherwise this will mean we will need to rewrite all modules ... now, which will probably be a bit too much work for the remaining ... 2 weeks ? or something like that

js/common/bridge.go
```diff
@@ -280,3 +280,72 @@ func Bind(rt *goja.Runtime, v interface{}, ctxPtr *context.Context) map[string]i
...
+		field := typ.Field(i)
+		if field.Type == reflect.TypeOf(&mic).Elem() {
+			continue
+		}
+		name := FieldName(typ, field)
+		if name != "" {
+			exports[name] = val.Field(i).Interface()
+		}
+	}
+	return Exports{Default: exports, Others: exports}
```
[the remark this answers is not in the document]
MStoykov: Currently(using common.Bind), this is the behaviour so this is just backwards compatible.
Default and others is the difference between:
- `import metrics from "k6/metrics"` - here whatever is default will be `metrics`
- `import * as metrics from "k6/metrics"` - here all things in Others will be `metrics`
With common.BInd there is just no way to differentiate between them, now there is and if the user wants they can defined `Default` as being `Counter` for example (in the case of metrics there aren't good examples :rofl: ) so that 
```
import theDefaultOfMetrics, { Counter} from "k6/metrics"
var c = new theDefaultOfMetrics("coolcounter") // the same as 
var c = new Counter("coolcounter") // this
```
(I did actually test it and it works). 
For some other modules this will likely make more sense, and is actually a functionality of ESM that we should probably support, now, instead of when we start needing it ;).

You can probably use it in the `k6/execution` so you can only have a `default` import ? :thinking: 

But either way `GenerateExports` tries to be just the most close thing to `common.Bind`s behaviour currently.

MStoykov:
closing in favour of #2108

Issue #2088, opened 2021-07-12
Pass setup data object into handleSummary callback


mschmidt-onecause:
## Feature Description
The setup data object that is returned by the setup callback is made available for use in VU functions and the teardown callback, but is not available in handleSummary.  In our case, the use case is to be able to call an API during setup that returns an ID, and then use that ID to call another API with the test summary results.

na--:
Thanks for opening this issue! This makes perfect sense and should be something relatively easy to add, I should have probably added it from the start... :sweat_smile: Would you like to try submitting a PR for it?

SamuelJohnson01997:
Hi. If this issue is not taken up yet, I’d like to take it up and try submitting a PR..

na--:
@SamuelJohnson01997, go for it! :+1:

PR #2103, opened 2021-07-31
Pass setup data object into handleSummary callback

SamuelJohnson01997:
Issue: #2088

Passing the setup data into handleSummary callback in the same way how it is in Teardown.

### DIFF
--- a/js/summary.go
+++ b/js/summary.go
@@ -25,7 +25,7 @@ import (
 	"fmt"
 	"io"
 	"time"
-
+	"encoding/json"
 	"github.com/dop251/goja"
 	"go.k6.io/k6/js/common"
 	"go.k6.io/k6/lib"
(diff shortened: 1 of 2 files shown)


imiric (changes requested):
Thanks for this @SamuelJohnson01997!

I think it looks good, but would like to see some tests for it before merging.

SamuelJohnson01997:
Sure. On it.

SamuelJohnson01997:
Have added the test for the same. Kindly have a look :)

js/summary.go
```diff
@@ -117,6 +117,17 @@ func summarizeMetricsToObject(data *lib.Summary, options lib.Options) map[string
 	}
 	m["metrics"] = metricsData
 
+	var setupDataI interface{}
+	if setupData != nil {
+		if err := json.Unmarshal(setupData, &setupDataI); err != nil {
+			return nil
```
na--, line 123: hmm I don't like that we're silently ignoring the error here, _and_ returning nothing at the same time :confused: this error is very, _very_ unlikely to ever happen, but if it does, we should still at least return `m` instead of `nil` here

SamuelJohnson01997: Makes sense. Can I just make it return m?

na--: Yep, that is what I meant. Maybe add a `// TODO: log the error` as well, though since it probably won't ever occur, it's not a big deal.

PR #2108, opened 2021-08-05
common.bind replacement

MStoykov:


MStoykov:
This is #2101 but not based on the metrics registry changes as it seems we won't merge them fast enough but #2093 is important and should land with the full k6/execution for this release.

Also I will likely squash everything when merging this once it's approved

imiric (commented):
LGTM, no major comments. :+1:

How the module struct needs to be defined now is a bit convoluted, and `GenerateExports()` is somewhat hairy, but nowhere near the mess of `common.Bind()` :sweat_smile:

js/modules/modules.go
```diff
@@ -69,31 +65,108 @@ type HasModuleInstancePerVU interface {
 	NewModuleInstancePerVU() interface{}
 }
 
+// IsModuleV2 ... TODO better name
+type IsModuleV2 interface {
```
imiric, line 69: I think this name is fine as it will probably be temporary until we remove `common.Bind()` in the next few weeks.

Let's just fix the comment.

imiric (approved):

js/modules/modules.go
```diff
@@ -69,31 +65,108 @@ type HasModuleInstancePerVU interface {
...
+	NewModuleInstance(InstanceCore) Instance
```
imiric, line 70: Does this make `NewModuleInstancePerVU()` obsolete? Essentially every V2 module will return a new instance per VU?

js/initcontext.go
```diff
@@ -140,14 +150,52 @@ func (i *InitContext) Require(arg string) goja.Value {
...
+
+func toESModuleExports(exp modules.Exports) map[string]interface{} {
+	result := make(map[string]interface{}, len(exp.Named)+2)
+
+	for k, v := range exp.Named {
+		result[k] = v
+	}
+	// Maybe check that those weren't set
+	result["default"] = exp.Default
+	result["__esModule"] = true // this so babel works with
```
imiric, line 182: Could you expand on this comment? Why is this required exactly?

It shouldn't be necessary for ES5 (`--compatibility-mode=base`), so it feels hacky.

js/modules/modules.go
```diff
@@ -69,31 +65,108 @@ type HasModuleInstancePerVU interface {
...
+	// sealing field will help probably with pointing users that they just need to embed this in their Instance
+	// implementations
+}
+
+// Exports is representation of ESM exports of a module
+type Exports struct {
+	// Default is what will be the `default` export of a module
+	Default interface{}
+	// Named is the named exports of a module
+	Named map[string]interface{}
```
imiric, line 130: @MStoykov I'm having an issue with exports for `k6/x/execution`, see [here](https://github.com/grafana/xk6-execution/pull/2/files#diff-1dc74514874c10c685d1ac9d2fe9f738853388a531c3b277c223e9561baad3bfR315).

With `--compatibility-mode=extended` and ES6's `import` everything works as expected. But with `--compatibility-mode=base` and ES5's `require()`, `default` is not exported... by default, and I need to manually reference it. I'm not defining `Named` since the module doesn't export any methods and everything should work via properties of the proxy.

I'm not sure if this is acceptable when working with ES5, or if the way we handle exports here should take that into account. Maybe `Named` should also be `interface{}` and we'd need more of `common.Bind`'s reflection to make it work?

js/modules/modules.go
MStoykov: yes, but we still use it, and somebody else, might too, so I prefer to deprecate it at a later point and remove at an even later one

js/modules/modules.go
MStoykov: Okay some explanation on commonjs vs ESM: 

In the case of commonjs, `modules.exports` (or just `exports`) is what is returned when you do `require("modulename")` and ... that is that. The major magic is that if you set something to `modules.exports`, the alias `exports` will magically be set to the same value as it's somekind of an alias. Other than that `require` acts exactly as a function would. And this was somewhat what we just did before that ... we just returned one big object which was just returned back to the script.

ESM on the other hand is ... well not at all the same. Not only is there no object to set values to, it's all special syntax and it has this concept of default and not default(named) exports, which also have a separate way of importing.

It just so happens that babel will try to guess based on the syntax and what it gets from `require` whether it has to use `default` or `named` exports somewhere and also when it sees `export something` it will set `modules.exports` accordingly including with the `__esModule` field. So the code in this PR just emulates what babel does when it exports on its own as much as possible and allows there to be a difference between what is default and not default export.

I am not really certain what a good solution will look like especially as I have mentioned numerous times that compatibility-mode=base is now less and less useful and hopefully one day when we have import/export syntax natively we can make it even less relevant.

I did have some worries about the usage of compatibility-mode=base myself, but as I used to set `Default` and `Named` to the same thing for `k6/metrics` it works the same as before.

I guess the easiest solution is to if `Named` is nil to just put `default` as the whole return and not set `__esModule`. (and maybe the reverse if `Default` is nil :shrug: ). This will later need to change significantly again once we know what is `require` and what is `import ... from ...`.  

I did wonder briefly if I should make some Proxy/DynamicObject which if it gets `__esModule` accessed change its form so it can act as both ... but as it wasn't needed I decided against it ... but I guess it will work :shrug: .

I think the Named check is probably going to be all we need though :thinking: , can you try it @imiric ?

imiric: Thanks for the explanation. We should mention some of that in the comment next to `__esModule` :wink:

You're right, that seems like the simplest fix for now, see da1fb43c. Works fine for `k6/x/execution` :+1:

na-- (dismissed):
LGTM in general, though I left a minor inline nitpick and :+1:-ed some of @imiric's comments

MStoykov:
@imiric @na-- does caae962 fixes all the comments you wanted ?

imiric (approved):

na-- (approved):
LGTM, or at least I can't think of a way to make this better...

codebien (approved):

Issue #1730, opened 2020-11-19
Switch to the InlfuxDB 2 Go client library


na--:
Prompted by [this](https://community.k6.io/t/does-k6-support-influxdb-2-0/1115) community forum question, I noticed that we're currently using the old InlfuxDB library (notice `influxdb1-client`): https://github.com/loadimpact/k6/blob/6081e6a3ff488990aab3fa349a84521cd495a1b3/stats/influxdb/collector.go#L30

https://github.com/influxdata/influxdb-client-go appears to be the new version, from its readme:
> This repository contains the reference Go client for InfluxDB 2.
> Note: Use this client library with InfluxDB 2.x and InfluxDB 1.8+ (see details). For connecting to InfluxDB 1.7 or earlier instances, use the influxdb1-go client library.

InfluxDB v1.8.0 seems to have been [released on 2020-04-13](https://docs.influxdata.com/influxdb/v1.8/about_the_project/releasenotes-changelog/), so it might be a bit too early to switch to the new Go library, considering that InfluxDB 2 should be somewhat [backwards compatible](https://docs.influxdata.com/influxdb/v2.0/reference/api/influxdb-1x/), but in a few months it will probably be fine.

MStoykov:
Looking at the [new](https://docs.influxdata.com/influxdb/v2.0/api/#operation/PostWrite) and [old](https://docs.influxdata.com/influxdb/v2.0/reference/api/influxdb-1x/write/) it looks like there seems to be very little reason to upgrade, unless someone has multiple orgs ... 

We could also drop the dependency and just do the requests on our own, we do 2 requests that are pure HTTP and will let us try to optimize the marshaling or anything else depending on our types. And also we won't wait on the dependency to [add basic functionality](https://github.com/influxdata/influxdb1-client/issues/19) .. that then I forgot to use it :facepalm:

imiric:
From my tests with v2 a few months ago we weren't able to write to it, though I forget the exact reason... The line protocol format seems to have remained backwards compatible, but there are some conceptual differences with org support and more robust auth in v2 that make it incompatible with v1 clients. Plus the switch from InfluxQL to Flux which shouldn't matter for us, but it hints that they're fundamentally different.

benc-uk:
I'd live to see InfluxDB v2 support as well. It's a big upgrade, and has the data explorer & dashboard built-in so no need for Grafana. It would make consuming k6.io data much easier

R0m3rCh:
[...]

PR #2110, opened 2021-08-05
output/influxdb: v2 support

codebien:
After many attempts and benchmarks with the async API, the solution with the current sample buffer is performing better.

This is the comparison with the current benchmarks:
```
name                        old time/op  new time/op  delta
Influxdb1Second-16           2.00s ± 0%   2.00s ± 0%    ~     (p=0.562 n=11+11)
Influxdb2Second-16           4.00s ± 0%   4.00s ± 0%  +0.02%  (p=0.002 n=10+11)
Influxdb100Milliseconds-16   401µs ± 2%   370µs ± 2%  -7.83%  (p=0.000 n=9+11)
```

and this with the highest rate:
```
name         old time/op  new time/op  delta
Influxdb-16  31.6µs ±10%  22.1µs ± 6%  -29.99%  (p=0.000 n=11+10)
```


output/influxdb/config.go
```diff
@@ -38,15 +38,15 @@ import (
 type Config struct {
 	// Connection.
 	Addr             null.String        `json:"addr" envconfig:"K6_INFLUXDB_ADDR"`
-	Username         null.String        `json:"username,omitempty" envconfig:"K6_INFLUXDB_USERNAME"`
-	Password         null.String        `json:"password,omitempty" envconfig:"K6_INFLUXDB_PASSWORD"`
```
codebien, line 42: I totally removed them but I think could be a good idea to maintain and make them compatible following the influx client's doc suggestion WDYT?.

>Use the form username:password for an authentication token. Example: my-user:my-password. Use an empty string ("") if the server doesn't require authentication.

arukiidou:
related issue
https://github.com/grafana/k6/issues/1730

output/influxdb/output.go
```diff
@@ -196,11 +203,46 @@ func (o *Output) Stop() error {
...
+		// the bucket already exists
+		return nil
+	}
+
+	// TODO: It has the bucket name in the error message. Can we do a better check?
+	if err.Error() != fmt.Sprintf("bucket '%s' not found", o.Config.Bucket.String) {
+		return err
+	}
+
+	// create a bucket with the default (infinite) retention policy
```
MStoykov, line 231: the `infinite` part seems bad to me

output/influxdb/output.go
```diff
@@ -209,17 +251,12 @@ func (o *Output) flushMetrics() {
...
-		o.logger.WithError(err).Error("Couldn't create batch from samples")
-		return
-	}
+	batch := o.batchFromSamples(samples)
+	o.logger.WithField("points", len(batch)).Debug("Writing...")
 
-	o.logger.WithField("points", len(batch.Points())).Debug("Writing...")
 	startTime := time.Now()
-	if err := o.Client.Write(batch); err != nil {
+	if err := o.pointWriter.WritePoint(context.Background(), batch...); err != nil {
```
MStoykov, line 258: Have you tried using the Async writer but without first getting all the batches as a whole but just in `batchFromSamples` without creating a whole slice and so on?

MStoykov (commented):
I wonder what happens if somebody tries to login while having saved log credentials ... I guess they will be overwritten and lost so they will not be able to go back(without relogging) ?

It seems to me from the code that it will work with username and password if found currently? 

I am also really impressed that we need almost 100k of dependencies to do this :facepalm:

codebien:
>I wonder what happens if somebody tries to login while having saved log credentials ... I guess they will be overwritten and lost so they will not be able to go back(without relogging)?

@MStoykov Yes, you're right. We could read the config and add a warning asking for a confirmation, in the case, the user has an already saved configuration. WDYT?

>It seems to me from the code that it will work with username and password if found currently?

Yes, if they are found then the config merges them in a `Token`. So, potentially, the user could abort the previously suggested confirmation and continue with the existing credentials and getting [the warning](https://github.com/grafana/k6/blob/b4251baa3a58580194054e8d44d93c32ee390e84/output/influxdb/output.go#L91) during the test run.

>I am also really impressed that we need almost 100k of dependencies to do this :facepalm:

Unfortunately, this is the cost to pay for it, otherwise, we have to consider alternatives if we consider this size relevant. /cc @na--

output/influxdb/output.go
codebien: Would you add it as a configurable option?

output/influxdb/output.go
codebien: I tried without using the SampleBuffer, I was pushing directly the received sample to the async writer. In the middle, there was only the tags conversion.

output/influxdb/output.go
MStoykov: we used to have it as a configuration as well as the precision of the measurements and w/e "persistence" is in this context so we might want to actually read them and support them :shrug:

MStoykov:
Lol I just checked and this adds like 3mb to the binary (or almost 10%) 
```
35140   k6-master
38880   k6-influxdbv2
35096   k6-34a7743f
```
34a7743f is the commit this branch is ... branching from

output/influxdb/config.go
```diff
@@ -161,20 +155,6 @@ func ParseURL(text string) (Config, error) {
 			default:
 				return c, fmt.Errorf("insecure must be true or false, not %s", vs[0])
 			}
-		case "payload_size":
```
MStoykov, line 164: hm I guess you fixed #2137  through dropping it @codebien :rofl:

codebien: It was used for the UDP client that has been dropped in the v2.

MStoykov: I guess this is another breaking change, can you please make a list of the breaking changes in the main PR message ?

codebien:
I miss one feature to be totally full retro compatible. The creation of the database with 1.8. The compatibility query API just accepts a Flux query, which changes my plan to execute an influx query with the CREATE DATABASE instruction.
If we want to continue supporting it then we have to add a dedicated HTTP request executing this equivalent `curl` request: 

```
curl -i -XPOST http://localhost:8086/query --data-urlencode "q=CREATE DATABASE k6"
```
Also, how should we identify that the user is using the 1.8 version?
* We could query the [health](https://pkg.go.dev/github.com/influxdata/influxdb-client-go/v2#Client.Health) endpoint that returns a [version](https://pkg.go.dev/github.com/influxdata/influxdb-client-go/v2@v2.5.0/domain#HealthCheck) value.
* We could detect it matching the returned error message trying to call `/buckets`, the client returns an `Unexpected status code 404` message in the case we are executing it on 1.8. But some logic with dependency from a message returned would be very fragile in terms of code.

 /cc @na @MStoykov @imiric

output/influxdb/util.go
```diff
@@ -21,42 +21,29 @@
...
-	client "github.com/influxdata/influxdb1-client/v2"
+	client "github.com/influxdata/influxdb-client-go/v2"
+	influxdblog "github.com/influxdata/influxdb-client-go/v2/log"
 	"gopkg.in/guregu/null.v3"
 )
 
+func init() {
+	// disable the internal influxdb log
+	influxdblog.Log = nil
+}
```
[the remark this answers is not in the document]
codebien: The main problem is with tests that are invoked with `t.Parallel` so when multiple outputs are created then we would get a race.

output/influxdb/output.go
```diff
@@ -196,11 +218,89 @@ func (o *Output) Stop() error {
...
+		return err
+	}
+
+	return nil
+}
+
+// createDatabase18 create a database using the v1 API for the old 1.8+ InfluxDB server versions.
+func (o *Output) createDatabase18(ctx context.Context) error {
+	// it uses the Bucket value following the compatibility mode
+	db := o.Config.Bucket.String
```
codebien, line 265: :warning: I'm not sure about it, but I think we don't have other options. To support both database and bucket we have to normalize them to Bucket.

Database: `http://localhost:8086/mydb?precision=1s`
Bucket: `http://localhost:8086/mybucket?retention=7d`

BrynCooke:
I had a quick go with this PR in combination with Telegraf and get the following error message: `the server healthcheck has failed`
I may have configured things incorrectly, but but it's worth mentioning that Telegraf has a separate health output plugin https://github.com/influxdata/telegraf/tree/master/plugins/outputs/health

BrynCooke:
After experimenting this morning, it looks like Telegraf will return 404 for when calling `Client.Health` or `Client.Ready`. 
Deleting: https://github.com/grafana/k6/pull/2110/files#diff-a3d0c9f25457484db69ff5e116813b7abc22151fdce42d093813546b8932ac97R191-R204 enabled me to use this PR with Telegraf.

codebien:
Updated the `health` proposal skipping the auto-creation in case of error. That means auto-creation doesn''t work with Telegraf so this is a breaking change. :confused:

output/influxdb/config.go
```diff
@@ -141,40 +143,40 @@ func ParseURL(text string) (Config, error) {
...
 		pass, _ := u.User.Password()
 		c.Password = null.StringFrom(pass)
 	}
+	if bucket := strings.TrimPrefix(u.Path, "/"); bucket != "" {
+		c.Bucket = null.StringFrom(bucket)
+	}
 	for k, vs := range u.Query() {
 		switch k {
-		case "insecure":
+		case "insecureSkipTLSVerify":
```
MStoykov, line 156: Please list this option rename as a breaking change ... and maybe we should support it for a verison or two and print a warning?

MStoykov (commented):
Apart from my general opinion that having outputs in core should be only when it's necessary and the fact that his adds 3mb to the binary I am okay with the curren state. 

Additional Notes (in no particular order):
1. We should test telegraf creation of the db/bucket and see if it works. On the same not maybe if one of the creation errors out and we don't have an idea which version it is ... we should try the other with some amount of warnings? Or maybe we should not be doing this at all? After all from what I've noticed by default, we might create a bucket with infinite retention which is likely not what a user wants (us eating their diskspace with 2 years old data). 
2. The current benchmark is pretty bad ... and unfortunately, it is me who has written it. As far as I can see I have tried to figure out a way to simulate influxdb being slow on the ingest ... but in doing so I am basically testing that pushing in parallel is better ... which it is, because the benchmark is literally written so that is the end result. From that perspective it might be better if you want to try to go the asynchronous route to test against a *real* influxdb and ... in all cases delete if not rewrite the benchmark. 
At the end of the day, we care how the performance is against a real influxdb (or telegraf I guess) not against something that acts in a certain way I decided upon. In the same vain having the benchmark running a real influxdb is IMO outside the scope of k6 core project, and likely will take way too much time to develop(and support), but if someone has ideas on how to benchmark this better, I am open. I guess we can have something that reads as fast as possible and see if the asynchronous one at least keeps up :shrug: and then hope influxdbv2 is fast. But even on that, I think the current benchmark has teh `for b.N` loop in the wrong place, so maybe starting from scratch is better.

codebien:
>that having outputs in core should be only when it's necessary

As we said, this will be moved probably in a dedicated extension.

>Or maybe we should not be doing this at all?

This is what the current code is implementing now, just for maintaining logic around this problem a bit simpler.

>We should test telegraf creation of the db/bucket and see if it works.

I tested it and the bucket creation doesn't work. Did you expect something different?

>In the same vain having the benchmark running a real influxdb is IMO outside the scope of k6 core project, and likely will take way too much time to develop(and support),

When we will integrate this into the core then I think we should have a series of integration tests. In this way, we could have a good reliability. Instead, if we will maintain it as an extension maybe we could be more fault-tolerant avoiding them considering that we could react faster to eventual bugs. 

>I guess we can have something that reads as fast as possible and see if the asynchronous one at least keeps up

The highest rate reported in the description is done in this way. I can push the refactored bench having only one test with the highest rate.

PR #2114, opened 2021-08-11
Don't show deprecated outputs in the list of available ones

MStoykov:


na-- (commented):
Weren't we dropping kafka and datadog in v0.34.0?

MStoykov:
we have dropped them but the way we say what outputs we support is by going through the map in the code and getting the keys. 
But because we want to still print a message for kafka and datadog they are in that map just so they print the deprecation and removal notice and then error out.

na-- (approved):
:man_facepalming: :coffee: :+1:

imiric (approved):

PR #2122, opened 2021-08-23
github/ci: macos in tests matrix

codebien:


codebien:
Going to close, still too many flaky tests with macOS.

PR #2140, opened 2021-09-17
Fix inconsistency in env variables names

yorugac:
Very small fix :slightly_smiling_face:

### DIFF
--- a/cmd/config_consolidation_test.go
+++ b/cmd/config_consolidation_test.go
@@ -80,6 +80,19 @@ func verifyConstLoopingVUs(vus null.Int, duration time.Duration) func(t *testing
 	}
 }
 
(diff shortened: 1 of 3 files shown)


codebien (dismissed):
LGTM

imiric (approved):
Surprised nobody has mentioned this before :sweat_smile: We should probably mark it as a breaking change in the release notes.

na-- (approved):
LGTM, though please also add a test in https://github.com/grafana/k6/blob/master/cmd/config_consolidation_test.go
Our config is woefully under-tested, so we might as well start with simple things like this.

imiric (approved):

codebien (approved):

MStoykov (approved):

PR #2150, opened 2021-10-01
Support camelCase CSV output argument for file_name and save_interval.

JosephWoodward:
This change adds support for camelCased versions of file_name and save_interval CSV output arguments,
writing a warning to the logs informing users the older snake cased argument will be deprecated in time.

### DIFF
--- a/output/csv/config.go
+++ b/output/csv/config.go
@@ -29,6 +29,7 @@ import (
 	"gopkg.in/guregu/null.v3"
 
 	"github.com/kelseyhightower/envconfig"
+	"github.com/sirupsen/logrus"
 	"go.k6.io/k6/lib/types"
 )
 
@@ -59,7 +60,7 @@ func (c Config) Apply(cfg Config) Config {
 }
 
 // ParseArg takes an arg string and converts it to a config
-func ParseArg(arg string) (Config, error) {
+func ParseArg(arg string, logger *logrus.Logger) (Config, error) {
 	c := Config{}
 
 	if !strings.Contains(arg, "=") {
@@ -76,11 +77,17 @@ func ParseArg(arg string) (Config, error) {
 		}
 		switch r[0] {
 		case "save_interval":
+			logger.Warnf("CSV output argument '%s' will soon be deprecated, please use 'saveInterval' instead.", r[0])
+			fallthrough
+		case "saveInterval":
 			err := c.SaveInterval.UnmarshalText([]byte(r[1]))
 			if err != nil {
 				return c, err
 			}
 		case "file_name":
+			logger.Warnf("CSV output argument '%s' will soon be deprecated, please use 'fileName' instead.", r[0])
+			fallthrough
+		case "fileName":
 			c.FileName = null.StringFrom(r[1])
 		default:
 			return c, fmt.Errorf("unknown key %q as argument for csv output", r[0])
@@ -92,7 +99,7 @@ func ParseArg(arg string) (Config, error) {
 
 // GetConsolidatedConfig combines {default config values + JSON config +
 // environment vars + arg config values}, and returns the final result.
-func GetConsolidatedConfig(jsonRawConf json.RawMessage, env map[string]string, arg string) (Config, error) {
+func GetConsolidatedConfig(jsonRawConf json.RawMessage, env map[string]string, arg string, logger *logrus.Logger) (Config, error) {
 	result := NewConfig()
 	if jsonRawConf != nil {
 		jsonConf := Config{}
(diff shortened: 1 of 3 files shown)


output/csv/config.go
```diff
@@ -59,7 +60,7 @@ func (c Config) Apply(cfg Config) Config {
 }
 
 // ParseArg takes an arg string and converts it to a config
-func ParseArg(arg string) (Config, error) {
+func ParseArg(arg string, logger *logrus.Logger) (Config, error) {
```
JosephWoodward, line 63: Do you have any thoughts on how we could avoid such a breaking change to the package's pubic API? Other options I was considering was:

1. Use `Config` to indicate a soon to be deprecated argument was used, and write the log message further up the call stack where the logger is initialised (in `output.go`)
2. Move the checking of the soon to be deprecated arguments to `output.go`

imiric: Hi, thanks for the PR!

Hhmm yeah, this is not ideal, but I think I prefer leaving it as is more than the alternatives, as currently it's clear that the warning happens where the configuration is assembled, and doing that validation outside of `config.go` would muddy things a bit.

Other than that a package like [`warnings`](https://pkg.go.dev/gopkg.in/warnings.v0) might help, though it's not very idiomatic and we wouldn't want to add another dependency because of this.

So I'd say to leave it as is, unless someone from @grafana/k6-devs has a better idea.

JosephWoodward: Okay, thanks @imiric 

> Hhmm yeah, this is not ideal, but I think I prefer leaving it as is more than the alternatives, as currently it's clear that the warning happens where the configuration is assembled, and doing that validation outside of config.go would muddy things a bit.

Yeah, I agree - this way keeps the logic local. Interested to see what @grafana/k6-devs think.

output/csv/config.go
```diff
@@ -92,7 +99,8 @@ func ParseArg(arg string) (Config, error) {
 
 // GetConsolidatedConfig combines {default config values + JSON config +
 // environment vars + arg config values}, and returns the final result.
-func GetConsolidatedConfig(jsonRawConf json.RawMessage, env map[string]string, arg string) (Config, error) {
+//nolint: lll
```
JosephWoodward, line 102: Passing the additional logger parameter causes the line length linting error, but given we're only passing the logger until people have migrated to the camel case version I felt the no lint comment would be sufficient for the time being.

output/csv/config.go
yorugac: Perhaps, it makes sense to add this case (logging during config parsing) to #883 as "nice-to-have" to figure out how best to approach this if it comes up again?

JosephWoodward: @yorugac Good idea, I've just added this as a nice to have to that issue. :+1:

output/csv/config.go
MStoykov: I in general prefer to split the `func <name>(<arguments>) (<results>){` on three lines as in 
```
func <name>(
    <arguments>,
)(<results>){ 
```
as it keeps it more readable especially when lines start to wrap.

output/csv/config_test.go
```diff
@@ -91,23 +94,37 @@ func TestParseArg(t *testing.T) {
...
+	testLog := logrus.New()
+	testLog.SetOutput(testutils.NewTestOutput(t))
+
 	for arg, testCase := range cases {
 		arg := arg
 		testCase := testCase
 
 		t.Run(arg, func(t *testing.T) {
-			config, err := ParseArg(arg)
+			config, err := ParseArg(arg, testLog)
```
MStoykov, line 127: This could've also tested that the warning is being logged

output/csv/config.go
```diff
@@ -76,11 +77,17 @@ func ParseArg(arg string) (Config, error) {
 		}
 		switch r[0] {
 		case "save_interval":
+			logger.Warnf("CSV output argument '%s' will soon be deprecated, please use 'saveInterval' instead.", r[0])
```
MStoykov, line 80:
```suggestion
			logger.Warnf("CSV output argument '%s' is deprecated, please use 'saveInterval' instead.", r[0])
```
It is deprecated, we just still support it. As I don't know if we will ever remove it (maybe when we redo configuration) I don't think it's needed to also add "we will remove it in v0.XX.0" or something like that that but it's possible I guess

JosephWoodward:
@MStoykov Thank you for the comments, I will get these sorted this evening.

JosephWoodward:
@MStoykov This is ready for a re-review when you can please. I've added an assertion over the expected log messages and level, and updated your other comments too.

inancgumus (approved):
LGTM

MStoykov (approved):

JosephWoodward:
@FloorD That'd be great! Big fan of what everyone is doing at K6/Grafana so certainly wouldn't say no!

PR #2151, opened 2021-10-01
bugfix#2036 Websocket connections ignore userAgent setting

cooliscool:
#### Changes I have made:
Initializing `header` variable beforehand to apply the default User-Agent string (example - "k6/0.27.0 (https://k6.io/)") , OR to apply the User-Agent supplied through `Options` if it exists.

### DIFF
--- a/js/modules/k6/ws/ws.go
+++ b/js/modules/k6/ws/ws.go
@@ -108,10 +108,10 @@ func (*WS) Connect(ctx context.Context, url string, args ...goja.Value) (*WSHTTP
 		return nil, errors.New("last argument to ws.connect must be a function")
 	}
 
-	// Leave header to nil by default so we can pass it directly to the Dialer
-	var header http.Header
-
+	header := make(http.Header)
 	tags := state.CloneTags()
+
+	header.Set("User-Agent", state.Options.UserAgent.String)
 
 	// Parse the optional second argument (params)
 	if !goja.IsUndefined(paramsV) && !goja.IsNull(paramsV) {
@@ -119,7 +119,6 @@ func (*WS) Connect(ctx context.Context, url string, args ...goja.Value) (*WSHTTP
 		for _, k := range params.Keys() {
 			switch k {
 			case "headers":
-				header = http.Header{}
 				headersV := params.Get(k)
 				if goja.IsUndefined(headersV) || goja.IsNull(headersV) {
 					continue


js/modules/k6/ws/ws.go
```diff
@@ -108,18 +108,17 @@ func (*WS) Connect(ctx context.Context, url string, args ...goja.Value) (*WSHTTP
 		return nil, errors.New("last argument to ws.connect must be a function")
 	}
 
-	// Leave header to nil by default so we can pass it directly to the Dialer
-	var header http.Header
-
+	header := make(http.Header)
 	tags := state.CloneTags()
 
+	header.Set("User-Agent", state.Options.UserAgent.String)
```
codebien, line 114: Can you move this before `CloneTags`, please? I would prefer to not have it in the middle between initialization and assignment.

codebien (changes requested):
Hi @cooliscool,
thanks for your contribution! :tada:

Can you also add a small test for asserting the expected logic with the header assigned, please?

js/modules/k6/ws/ws.go
cooliscool: Sure.! Working on that.

MStoykov (commented):
LGTM but I would prefer we have a test(or to expand one) to actually check that it works so we don't break it later

cooliscool:
@codebien @MStoykov 

#### Explanation for the changes made: 
I couldn't find a way to retrieve headers of a websocket request that got fired, from within JS environment. ( the actual JS WebSockets API doesn't even have an option to set custom header. [refer this stackoverflow question](https://stackoverflow.com/questions/4361173/http-headers-in-websockets-client-api)  )
So, I've modified the echo server in `httpmultibin.go` to send back a `Echo-User-Agent` header in the response if the `User-Agent` is set by client. In future this could be expanded add tests for any custom headers sent through k6.

Also kindly let me know if there's a better way to implement this test. I will be happy to make it better.

cooliscool:
I can't figure out why lint is breaking with the message 
`js/modules/k6/ws/ws_test.go:322:2  paralleltest  Function TestSession has missing the call to method parallel in the test run`

here: 
1. The same pattern as other sub tests in `TestSession` is followed in the newly added `useragent` test.
2. Tests are running fine. 
3. Putting in a `t.Parallel()` within `t.Run()` causes the tests to fail.

lib/testutils/httpmultibin/httpmultibin.go
```diff
@@ -108,7 +108,14 @@ type jsonBody struct {
...
 	return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
-		conn, err := (&websocket.Upgrader{}).Upgrade(w, req, w.Header())
+
+		// Echo back User-Agent header if it exists
+		responseHeaders := w.Header().Clone()
+		if ua := req.Header.Get("User-Agent"); ua != "" {
+			responseHeaders.Add("Echo-User-Agent", req.Header.Get("User-Agent"))
+		}
+
+		conn, err := (&websocket.Upgrader{}).Upgrade(w, req, responseHeaders)
```
MStoykov, line 118: Can you please instead of changing this make a custom handler just for you test and add it just for your test as it's done in https://github.com/grafana/k6/blob/f76bfa5a34ba9af7e0165cb7016167ffed7dcb91/js/modules/k6/http/tls_test.go#L40-L43 .

You can skip most of the rest of the code after the `Upgrade` ... I think :thinking:

cooliscool: thanks for the code example.. !

cooliscool:
> Apart from moving the changes from httpmultibin to the test that uses them I have no other requests and I think after that we can merge this tada
> 
> Thanks a lot for all the work bow

Thank you!
I have removed modifications made to `httpmultibin`, and moved test specific code to `TestUserAgent`.

PR #2153, opened 2021-10-04
Fix go get xk6 github action

inancgumus:
+ `go get` is now only being used for adding dependencies to a Go module
+ this fix uses `go install` instead to install the xk6 binary (instead of `go get`)

### DIFF
--- a/.github/workflows/xk6.yml
+++ b/.github/workflows/xk6.yml
@@ -50,7 +50,7 @@ jobs:
           cd .github/workflows/xk6-tests
           # Temporary hack to ignore the v0.33.0 tag change
           export GONOSUMDB="go.k6.io/k6"
-          go get go.k6.io/xk6/cmd/xk6@master
+          go install go.k6.io/xk6/cmd/xk6@latest
           if [ "${{ github.event_name }}" == "pull_request" -a \
                "${{ github.event.pull_request.head.repo.full_name }}" != "${{ github.repository }}" ]; then
             export XK6_K6_REPO="github.com/${{ github.event.pull_request.head.repo.full_name }}"
--- a/.github/workflows/xk6.yml
+++ b/.github/workflows/xk6.yml
@@ -50,7 +50,7 @@ jobs:
           cd .github/workflows/xk6-tests
           # Temporary hack to ignore the v0.33.0 tag change
           export GONOSUMDB="go.k6.io/k6"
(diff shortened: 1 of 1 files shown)


MStoykov (changes requested):
Can you please change the `go get` in the `all.yml` as well

.github/workflows/xk6.yml
```diff
@@ -50,7 +50,7 @@ jobs:
           cd .github/workflows/xk6-tests
           # Temporary hack to ignore the v0.33.0 tag change
           export GONOSUMDB="go.k6.io/k6"
-          go get go.k6.io/xk6/cmd/xk6@master
+          go install go.k6.io/xk6/cmd/xk6@latest
```
imiric, line 53: I think we want to keep `master` here, to test even with unreleased versions.

inancgumus:
> Can you please change the `go get` in the `all.yml` as well

* glide has a problem with `go install`, see: https://github.com/Masterminds/glide/issues/1082

Do we need to keep glide anymore? It's because the following now works:
* `go install github.com/mh-cbon/go-bin-deb@master`
* `go install github.com/mh-cbon/go-bin-rpm@master`
Instead of:
```
go get -d github.com/mh-cbon/go-bin-deb \
              && (cd "$gopath/src/github.com/mh-cbon/go-bin-deb" \
              && glide install \
              && go install)
          go get -d github.com/mh-cbon/go-bin-rpm \
              && (cd "$gopath/src/github.com/mh-cbon/go-bin-rpm" \
              && glide install \
              && go install)
```

@MStoykov @imiric

yorugac:
Just a note, I think this PR should fix the current failure in Github actions for master branch at `test-xk6 (tip, macos-latest)` job.

inancgumus:
> test-xk6 (tip, macos-latest)

I couldn't find that branch?

yorugac:
`test-xk6` is a job name, it just failed on master branch here: https://github.com/grafana/k6/runs/3788611719

Also, a suggestion: how about adding `fail-fast: false` to the `test-xk6` as well? As is done for `test-prev` and `test-current-cov` in `all.yml`

MStoykov (approved):
Given that this currently breaks CI and the others will not be relevant until we move to 1.18 to release I am for merging this as is and @inancgumus please open at least a very basic issue to actually change the rest so we don't forget it

imiric (approved):

PR #2124, opened 2021-08-29
Refactor ramping VUs executor

inancgumus:
Refactored Ramping VUs executor—_a little bit_. This is the first step as I'm learning about the system. I hope this refactoring makes sense.

PR #2155, opened 2021-10-04
Refactor ramping VUs executor

inancgumus:
`RampingVUs.Run` method is complex and this PR aims to refactor it.

**Improvements:**

* Makes the `Run` method easier to read and understand.
* Makes it explicit which Goroutines are being fired in the `Run` method.
* Separates responsibilities to other parts like:
  * `rampingVUsRunState` and its methods.
  * Moves `trackProgress` Goroutine from the `makeProgressFn` to the `Run` method.
  * Makes `populateVUHandles` to only handle initializing VU handles (_not firing up a GR_)
  * Uses two strategy functions for code reusability and clarity (_`handleVUs` and `handleRemainingVUs` use them_).
* Makes the `handleVUs` algorithm easier to understand and manage.
* Adds a `handleRemainingVUs` test with explanations  to make newcomers' and new contributors' job easier

_Note: The other PR (#2124) is convoluted and somehow creates problems locally and with Github. I'd like to keep working on it on this separate branch._

### DIFF
--- a/lib/executor/ramping_vus.go
+++ b/lib/executor/ramping_vus.go
@@ -533,7 +533,6 @@ func (vlv *RampingVUs) Run(ctx context.Context, _ chan<- stats.SampleContainer, 
 
 	runState := &rampingVUsRunState{
 		executor:       vlv,
-		wg:             sync.WaitGroup{},
 		vuHandles:      make([]*vuHandle, maxVUs),
 		maxVUs:         maxVUs,
 		activeVUsCount: new(int64),
--- a/lib/executor/ramping_vus.go
+++ b/lib/executor/ramping_vus.go
@@ -493,154 +493,185 @@ var _ lib.Executor = &RampingVUs{}
 
 // Run constantly loops through as many iterations as possible on a variable
 // number of VUs for the specified stages.
-//
-// TODO: split up? since this does a ton of things, unfortunately I can't think
-// of a less complex way to implement it (besides the old "increment by 100ms
-// and see what happens)... :/ so maybe see how it can be split?
-// nolint:funlen,gocognit,cyclop
-func (vlv RampingVUs) Run(
-	parentCtx context.Context, out chan<- stats.SampleContainer, _ *metrics.BuiltinMetrics,
-) (err error) {
-	rawExecutionSteps := vlv.config.getRawExecutionSteps(vlv.executionState.ExecutionTuple, true)
-	regularDuration, isFinal := lib.GetEndOffset(rawExecutionSteps)
-	if !isFinal {
-		return fmt.Errorf("%s expected raw end offset at %s to be final", vlv.config.GetName(), regularDuration)
+func (vlv RampingVUs) Run(ctx context.Context, _ chan<- stats.SampleContainer, _ *metrics.BuiltinMetrics) error {
+	rawSteps := vlv.config.getRawExecutionSteps(vlv.executionState.ExecutionTuple, true)
+	regDur, finalRaw := lib.GetEndOffset(rawSteps)
+	if !finalRaw {
+		return fmt.Errorf("%s expected raw end offset at %s to be final", vlv.config.GetName(), regDur)
 	}
-
-	gracefulExecutionSteps := vlv.config.GetExecutionRequirements(vlv.executionState.ExecutionTuple)
-	maxDuration, isFinal := lib.GetEndOffset(gracefulExecutionSteps)
-	if !isFinal {
-		return fmt.Errorf("%s expected graceful end offset at %s to be final", vlv.config.GetName(), maxDuration)
+	gracefulSteps := vlv.config.GetExecutionRequirements(vlv.executionState.ExecutionTuple)
+	maxDur, finalGraceful := lib.GetEndOffset(gracefulSteps)
+	if !finalGraceful {
+		return fmt.Errorf("%s expected graceful end offset at %s to be final", vlv.config.GetName(), maxDur)
 	}
-	maxVUs := lib.GetMaxPlannedVUs(gracefulExecutionSteps)
-	gracefulStop := maxDuration - regularDuration
-
-	startTime, maxDurationCtx, regDurationCtx, cancel := getDurationContexts(parentCtx, regularDuration, gracefulStop)
+	startMaxVUs := lib.GetMaxPlannedVUs(gracefulSteps)
+	startTime, maxDurCtx, regDurCtx, cancel := getDurationContexts(ctx, regDur, maxDur-regDur)
 	defer cancel()
 
-	activeVUs := &sync.WaitGroup{}
-	defer activeVUs.Wait()
-
-	// Make sure the log and the progress bar have accurate information
 	vlv.logger.WithFields(logrus.Fields{
-		"type": vlv.config.GetType(), "startVUs": vlv.config.GetStartVUs(vlv.executionState.ExecutionTuple), "maxVUs": maxVUs,
-		"duration": regularDuration, "numStages": len(vlv.config.Stages),
-	},
-	).Debug("Starting executor run...")
+		"type":      vlv.config.GetType(),
+		"startVUs":  vlv.config.GetStartVUs(vlv.executionState.ExecutionTuple),
+		"maxVUs":    startMaxVUs,
+		"duration":  regDur,
+		"numStages": len(vlv.config.Stages),
+	}).Debug("Starting executor run...")
 
-	activeVUsCount := new(int64)
-	vusFmt := pb.GetFixedLengthIntFormat(int64(maxVUs))
-	regularDurationStr := pb.GetFixedLengthDuration(regularDuration, regularDuration)
-	progressFn := func() (float64, []string) {
-		spent := time.Since(startTime)
-		currentlyActiveVUs := atomic.LoadInt64(activeVUsCount)
-		vus := fmt.Sprintf(vusFmt+"/"+vusFmt+" VUs", currentlyActiveVUs, maxVUs)
-		if spent > regularDuration {
-			return 1, []string{vus, regularDuration.String()}
-		}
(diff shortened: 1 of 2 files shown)


lib/executor/ramping_vus.go
```diff
@@ -486,174 +486,235 @@ func (vlvc RampingVUsConfig) HasWork(et *lib.ExecutionTuple) bool {
...
+			return
 		}
-	}()
+		handleNewMaxAllowedVUs(s)
+	}
+}
 
-	return nil
+func (rs rampingVUsRunState) maxAllowedVUsHandlerStrategy() func(lib.ExecutionStep) {
+	var cur uint64
```
yorugac, line 683: Please add the comment here, smth like:
```
// cur is a current maximum of allowed VUs
```

lib/executor/ramping_vus.go
```diff
@@ -486,174 +486,235 @@ func (vlvc RampingVUsConfig) HasWork(et *lib.ExecutionTuple) bool {
...
+		pv := graceful.PlannedVUs
+		for ; pv < cur; cur-- {
+			rs.vuHandles[cur-1].hardStop()
+		}
+		cur = pv
+	}
+}
+
+func (rs rampingVUsRunState) scheduledVUsHandlerStrategy() func(lib.ExecutionStep) {
+	var cur uint64
```
yorugac, line 694: Similarly, please add a comment:
```
// cur is a currently scheduled number of VUs
```
Rationale: it's arguable but IMHO, this part became harder to read than before despite better overall structure. Esp. given the same name of `cur` variable in both functions. So simple comments should help.

lib/executor/ramping_vus.go
```diff
@@ -486,174 +486,235 @@ func (vlvc RampingVUsConfig) HasWork(et *lib.ExecutionTuple) bool {
...
+	return func(raw lib.ExecutionStep) {
+		pv := raw.PlannedVUs
+		for ; cur < pv; cur++ {
+			_ = rs.vuHandles[cur].start() // TODO: handle the error
+		}
+		for ; pv < cur; cur-- {
+			rs.vuHandles[cur-1].gracefulStop()
+		}
+		cur = pv
```
yorugac, line 703: Is this assignment necessary here? It seems no matter what, by the end of these two loops `cur = pv` should be true.

lib/executor/ramping_vus.go
```diff
@@ -486,174 +486,235 @@ func (vlvc RampingVUsConfig) HasWork(et *lib.ExecutionTuple) bool {
...
-	for i := uint64(0); i < maxVUs; i++ {
-		vuHandle := newStoppedVUHandle(
-			maxDurationCtx, getVU, returnVU, vlv.nextIterationCounters,
-			&vlv.config.BaseConfig, vlv.logger.WithField("vuNum", i))
-		go vuHandle.runLoopsIfPossible(runIteration)
-		vuHandles[i] = vuHandle
+	defer runState.wg.Wait()
+	runState.populateVUHandles(maxDurationCtx, cancel)
+	for i := uint64(0); i < runState.maxVUs; i++ {
+		go runState.vuHandles[i].runLoopsIfPossible(runState.runIteration)
```
yorugac, line 557: Here 1 loop turned into 2 loops after refactoring :joy: Is it possible to put all the `vuHandle` initialization into one `runState`'s method with 1 loop?

yorugac (changes requested):
All seems fine overall, just small suggestions mostly.

lib/executor/ramping_vus.go
```diff
@@ -493,154 +493,184 @@ var _ lib.Executor = &RampingVUs{}
...
-// nolint:funlen,gocognit,cyclop
-func (vlv RampingVUs) Run(
-	parentCtx context.Context, out chan<- stats.SampleContainer, _ *metrics.BuiltinMetrics,
-) (err error) {
-	rawExecutionSteps := vlv.config.getRawExecutionSteps(vlv.executionState.ExecutionTuple, true)
-	regularDuration, isFinal := lib.GetEndOffset(rawExecutionSteps)
-	if !isFinal {
-		return fmt.Errorf("%s expected raw end offset at %s to be final", vlv.config.GetName(), regularDuration)
+func (vlv RampingVUs) Run(ctx context.Context, _ chan<- stats.SampleContainer, _ *metrics.BuiltinMetrics) error {
+	rawSteps := vlv.config.getRawExecutionSteps(vlv.executionState.ExecutionTuple, true)
```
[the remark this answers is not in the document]
inancgumus: @na-- WDYT about my previous idea?

lib/executor/ramping_vus.go
inancgumus: Good catch ;)

lib/executor/ramping_vus.go
inancgumus: @yorugac

Some of this was because of a previous review request.

I like to keep it this way. Making the loop in the `Run` method here allows us to see all the goroutines we're launching and easier to see what's going on. You can see that the `Run` method launches all the necessary goroutines, and we can read it like a script.

lib/executor/ramping_vus.go
```diff
@@ -486,174 +486,234 @@ func (vlvc RampingVUsConfig) HasWork(et *lib.ExecutionTuple) bool {
...
+	vuHandles      []*vuHandle // handles for manipulating and tracking all of the VUs
+	maxVUs         uint64      // the scaled number of initially configured MaxVUs
+	activeVUsCount *int64      // the current number of active VUs, used only for the progress display
+	started        time.Time
+	wg             *sync.WaitGroup
+
+	runIteration func(context.Context, lib.ActiveVU) bool // a helper closure function that runs a single iteration
+}
+
+func (rs rampingVUsRunState) makeProgressFn(regular time.Duration) (progressFn func() (float64, []string)) {
```
MStoykov, line 591: I am of the opinion this (and all other methods) should be pointer receivers. Looking at the [Go Code Reviews Comments](https://github.com/golang/go/wiki/CodeReviewComments#receiver-type) section on this:
- this definitely seems a bit big
- has a sync struct in it - I actually noticed that you decided to make sync.WaitGroup a pointer and because in general I (and it seems from a quick grep most of k6 and literally all the vendored code) prefer to have it as a non-pointer field and copy the struct. I don't really have a ... particular reason for that I just think it makes it obvious that we make changes to the struct and locking/unlocking is a change.
- you also edit another field (`activeVUsCount`) although again using the fact that it's a pointer and you are updating it with atomics and I think that it should stay a pointer as that makes it obvious (to me) that we are using atomics on it. 

I am not so certain about the `executor` field but I think it can be a non pointer as we will only copy it once and then we just use it's fields (most of which are naturally pointers), but I don't think this matters all that much, so we can keep it a pointer

lib/executor/ramping_vus.go
MStoykov: Looking at this now I also feel that this will likely be better to just be part of `populateVUHandles` - it already hides a bunch of stuff, you can just rename it to `loopVUHandlesIfPossible` ;) 

What was the original concern as I can't find it?

lib/executor/ramping_vus.go
```diff
@@ -486,174 +486,234 @@ func (vlvc RampingVUsConfig) HasWork(et *lib.ExecutionTuple) bool {
...
-	var currentScheduledVUs, currentMaxAllowedVUs uint64
+	var (
+		handleNewMaxAllowedVUs = runState.maxAllowedVUsHandlerStrategy()
+		handleNewScheduledVUs  = runState.scheduledVUsHandlerStrategy()
+	)
+	handledGracefulSteps := runState.handleVUs(
+		ctx,
+		handleNewMaxAllowedVUs,
+		handleNewScheduledVUs,
+	)
```
MStoykov, line 568: It's not obvious to me that `handleVUs` goes through the steps and that this is likely where this goroutine will spend most of it's time. 

Maybe a rename to `goOverSteps` or simple `run` would be more understandable

lib/executor/ramping_vus.go
inancgumus: `populateVUHandles` only does what it says, and it does a single thing: populating VU handles. The `Run` method is responsible for orchestrating them.

What do you mean by the original concern?

MStoykov: > What do you mean by the original concern? 

you wrote "Some of this was because of previous review requests." so I am asking what was the concern in the original review requests as I can't find them.

inancgumus: Ah, got it now, thanks. Multiple people have reviewed this PR many many times, so I don't remember, sorry :(

lib/executor/ramping_vus.go
inancgumus: See 1eebf1785deb7c0b4b3174538adb1fde92702670

lib/executor/ramping_vus.go
inancgumus: Agreed. See 1eebf1785deb7c0b4b3174538adb1fde92702670

lib/executor/ramping_vus.go
yorugac: I couldn't find mention of it in my email history either.

I still think that it should be 1 loop / 1 method :) I.e. the readability won't be that much worse while adding loops like this makes at least me automatically wonder about "what if" of potential performance losses. How about adding a comment in `Run` mentioning that this call starts goroutines?

inancgumus: Here you go 😅 95fb52d5c1639014c77a028403f5152513ebff8e

yorugac (approved):
The ramping-vus are structured much better now :smile: Great job, Inanc! :star:

na-- (approved):
Looks awesome! :heart: :shipit:

inancgumus:
At last, thanks! 😅

PR #2156, opened 2021-10-05
Upgrade to Go 1.17

b5710546232:
# Description of the changes
- Upgrade go version to  1.17

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module go.k6.io/k6
 
-go 1.17
+go 1.16
 
 require (
 	github.com/Azure/go-ntlmssp v0.0.0-20200615164410-66371956d46c
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module go.k6.io/k6
 
-go 1.16
+go 1.17
 
 require (
 	github.com/Azure/go-ntlmssp v0.0.0-20200615164410-66371956d46c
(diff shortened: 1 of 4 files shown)


na-- (changes requested):
Thanks for this! Can you also re-enable `test-prev` now and make that CI task use Go 1.16, now that we'll use 1.17 for everything else: https://github.com/grafana/k6/blob/e0d8d9336f77fd8b4dc50c840ac7aedfac43574b/.github/workflows/all.yml#L69-L72


 We had [disabled it previously because we required 1.16 for `go:embed`](https://github.com/grafana/k6/pull/1975/commits/c03a01b01e9fc83ec3bd5683ec35446be13f061a).

na-- (dismissed):
Thanks!

go.mod
```diff
@@ -1,6 +1,6 @@
 module go.k6.io/k6
 
-go 1.16
+go 1.17
```
na--, line 3:
```suggestion
go 1.16
```

hmm maybe don't update this quite yet? :sweat_smile:

b5710546232: Ok, i will fix it soon. :)

MStoykov (approved):
LGTM and thanks for the change :bow:

MStoykov:
@b5710546232 very sorry about that, but it appears some months ago I decided to try to enable parallel testing under windows and as the test kept passing I decided it now works :tada:. 

But now looking at this it's obvious that I did the change in the disabled `test-prev` :facepalm: so if you can please `git revert 4c3571a75` and push again ... very sorry.

na-- (approved):

MStoykov (approved):

PR #2159, opened 2021-10-05
Update dependencies

MStoykov:
This became a lot bigger than anticipated ... Especially the `x/` updates are quite big :crossed_fingers: 
There is slight update on all, but in most cases it is "nothing interesting to us" or "small optimizations here and there" except the `x/*` ones where I skipped that as the changelog is quite big and only in commit messages ;(

Skipped upgrades:
- cobra - there is nothing we care about IMO it updates proto/grpc libraries :facepalm: 
- color/istty and co. - there are some changes between the repositories and it seems there will be more releases soon - next few days
- goquery - adds new functionality that we will need to update `k6/html` so decided to skip it for now, maybe an issue needs to be opened
- goja - there is an issue with some of our current `k6/crypto/x509` code that is being looked after in https://github.com/dop251/goja/issues/333


na-- (changes requested):
LGTM, with a minor nitpick. It's nice to see that golang.org/x/crypto/ssh/terminal has finally been replaced with golang.org/x/term: https://github.com/grafana/k6/blob/5c25ef3e63eded90ac14180a046756c6199f022a/vendor/golang.org/x/crypto/ssh/terminal/terminal.go#L8-L23

We should probably replace our own usage of the original package and drop the module entirely with this update, it seems fairly straightforward to do: https://github.com/grafana/k6/search?q=golang.org%2Fx%2Fcrypto%2Fssh%2Fterminal

yorugac (dismissed):

yorugac:
@MStoykov regarding this:

> goquery - adds new functionality that we will need to update k6/html so decided to skip it for now, maybe an issue needs to be opened

what functionality is it?

MStoykov:
@yorugac 
> 2021-06-14 (v1.7.0) : Add Single and SingleMatcher functions to optimize first-match selection (thanks @gdollardollar).

na-- (approved):

yorugac (approved):

PR #2160, opened 2021-10-05
Add grpc reflection (WIP)

inancgumus:
This PR is an initial barebone of the ongoing work of merging #1987 (Closed). It's not yet ready for review.


MStoykov:
@inancgumus , the current code there leads to metric emission for the reflect call. This, unfortunately:
1. happens when the *whole* connection gets closed,  so you will only see it if you have multiple iterations. If you have single iteration for example the emission of the metric happens *after* we cancel the main context (at least in my experiments and IIRC) so at least to me it seemed to be connected to the context getting canceled.
2. More importantly at least @na-- didn't want this to be emitted at all and I am kind of with him
3. it also happens without any of the tags ... because they aren't set yet - which is also how I found it as `getTags` was extended to return empty tags in case if they were missing.
So my original proposal was to:
1. create another context that you cancel after the reflect call is done ... hopefully making it emit a metric at that time. Because you canceled the context also it should then not actually push the metric.
2. if we decide we want this emitted we should probably set some *real* tags for it, and if not I think you can set tags on the context here instead of changing `getTags`.
3. maybe try to find out another way (without cancelling the context) to close the connection

MStoykov:
But you can add another context for the reflect "client" which seems to work from my quick test - it definitely didn't emit a metric as the context it will use for the `PushIfNotDone` call is the one that we cancel. I have no idea if this doesn't have some side effects we don't want though so still some investigation needed :shrug: 
```
diff --git a/js/modules/k6/grpc/client.go b/js/modules/k6/grpc/client.go
index e183a420..41950eb6 100644
--- a/js/modules/k6/grpc/client.go
+++ b/js/modules/k6/grpc/client.go
@@ -310,7 +310,9 @@ func (c *Client) Connect(ctxPtr *context.Context, addr string, params map[string
 // It is called in the connect function the first time the Client.Connect function is called.
 func (c *Client) reflect(ctxPtr *context.Context) error {
        client := reflectpb.NewServerReflectionClient(c.conn)
-       methodClient, err := client.ServerReflectionInfo(*ctxPtr)
+       ctx, cancel := context.WithCancel(*ctxPtr)
+       defer cancel()
+       methodClient, err := client.ServerReflectionInfo(ctx)
        if err != nil {
                return fmt.Errorf("can't get server info: %w", err)
        }
```

inancgumus:
@MStoykov Do you mean the third `grpc_req_duration` emission in the second command's output below?

This is for `grpc.js` with the last k6 release:

```
$ k6 run -i 2 --out csv=out.csv samples/grpc.js && cat out.csv
metric_name,timestamp,metric_value,check,error,error_code,expected_response,group,method,name,proto,scenario,service,status,subproto,tls_version,url,extra_tags
grpc_req_duration,1634127292,675.738591,,,,,,GetFeature,/main.RouteGuide/GetFeature,,default,main.RouteGuide,0,,,127.0.0.1:10000/main.RouteGuide/GetFeature,
checks,1634127292,1.000000,status is OK,,,,,,,,default,,,,,,
data_sent,1634127292,221.000000,,,,,,,,,default,,,,,,
data_received,1634127292,196.000000,,,,,,,,,default,,,,,,
iteration_duration,1634127292,677.243352,,,,,,,,,default,,,,,,
iterations,1634127292,1.000000,,,,,,,,,default,,,,,,
grpc_req_duration,1634127292,32.253398,,,,,,GetFeature,/main.RouteGuide/GetFeature,,default,main.RouteGuide,0,,,127.0.0.1:10000/main.RouteGuide/GetFeature,
checks,1634127292,1.000000,status is OK,,,,,,,,default,,,,,,
data_sent,1634127292,221.000000,,,,,,,,,default,,,,,,
data_received,1634127292,196.000000,,,,,,,,,default,,,,,,
iteration_duration,1634127292,33.295279,,,,,,,,,default,,,,,,
iterations,1634127292,1.000000,,,,,,,,,default,,,,,,
```

This is for `grpc_reflect.js`:

```
$ go run . run -i 2 --out csv=out.csv samples/grpc_reflect.js && cat out.csv
metric_name,timestamp,metric_value,check,error,error_code,expected_response,group,method,name,proto,scenario,service,status,subproto,tls_version,url,extra_tags
grpc_req_duration,1634127349,680.474636,,,,,,GetFeature,/main.RouteGuide/GetFeature,,default,main.RouteGuide,0,,,127.0.0.1:10000/main.RouteGuide/GetFeature,
checks,1634127349,1.000000,status is OK,,,,,,,,default,,,,,,
data_sent,1634127349,536.000000,,,,,,,,,default,,,,,,
data_received,1634127349,3000.000000,,,,,,,,,default,,,,,,
iteration_duration,1634127349,683.887176,,,,,,,,,default,,,,,,
iterations,1634127349,1.000000,,,,,,,,,default,,,,,,
#
# this?
#
grpc_req_duration,1634127349,683.086818,,,,,,,,,,,1,,,,
vus,1634127350,1.000000,,,,,,,,,,,,,,,
vus_max,1634127350,1.000000,,,,,,,,,,,,,,,
grpc_req_duration,1634127350,864.075159,,,,,,GetFeature,/main.RouteGuide/GetFeature,,default,main.RouteGuide,0,,,127.0.0.1:10000/main.RouteGuide/GetFeature,
checks,1634127350,1.000000,status is OK,,,,,,,,default,,,,,,
data_sent,1634127350,535.000000,,,,,,,,,default,,,,,,
data_received,1634127350,3000.000000,,,,,,,,,default,,,,,,
iteration_duration,1634127350,865.910412,,,,,,,,,default,,,,,,
iterations,1634127350,1.000000,,,,,,,,,default,,,,,,
```

inancgumus:
> @inancgumus I'm not a fan of using the context to pass values like that or relying on some internal state, and would prefer the context cancellation approach if possible.
> 

@imiric Agreed. I didn't like it as well. It was emitting a metric because it was using a different context than the one the connection uses.

Fix: https://github.com/grafana/k6/pull/2160/commits/927673d342e891dd4d39738bca220099cb73fca5

Issue #611, opened 2018-05-02
Support compression in websockets


wallenZhao2046:
I have used K6 in websocket performance test. While the content from server is compressed, and I got "�0E�e�!�56���j0&��v!�{�:�9�^�" printed in console. I used nodejs code to handle the same message from server. I got the right text.

So my question is how to decode compressed text in K6?

my K6 script is
```
socket.on('message', function (data) {
      console.log(typeof (data));
      console.log(data.length)
      if (data.length < 200) {
        // I got "�0E�e�!�56���j0&��v!�{�:�9�^�" in console
        console.log(data);
      }
      // I tried to decode it got "incorrect header check"
      let text = pako.inflate(data, { to: 'string' }); 
    }
```

If I used the following JS script, I got the correct text to inflate it to plain text.

```
ws.on('message', function (data) {

        console.log('-------- begin -------');
        // I got <Buffer 1f 8b 08 00 00 00 00 00 00 00 2d 8b 4b 0a 80 20 14 45 f7 72 c7 21 f9 1b e4 6e 34 1f 14 12 49 5e 47 d1 de 33 68 7a 3e 37 f6 8c 80 c4 b5 b7 4c 4c 68 3d in console
        console.log(data);
        console.log('-------- end -------');

        let text = pako.inflate(data, {
            to: 'string'
        });
        // msg is " msg: {"id":"btcusdt","subbed":"market.btcusdt.depth.step0","ts":1525243319443,"status":"ok"} "
        console.log('msg: ' + text);
    })
```

na--:
I already answered in the [stackoverflow question](https://stackoverflow.com/questions/50128949/k6-k6-websocket-how-to-decode-compressed-text-in-k6/50129281#50129281), but I'll copy it here for posterity :smile: 

> As I mentioned in the [previous] [GitHub issue](https://github.com/loadimpact/k6/issues/594), it seems like pako's browserify version may have some issues when you specify `{ to: 'string' }` in the options. Instead you should be able to do something like this:
> ```
> let binText = pako.inflate(data);
> let strText = String.fromCharCode.apply(null, binText);
> console.log(strText);
> ```

[...]

PR #2162, opened 2021-10-06
support compression in websockets #611

cooliscool:
#611

#### Here's my research on compression in Websockets : 

permessage-deflate compression is defined in RFC7692 and gorilla/websockets has implemented this as an experimental feature. 
[read this here](https://pkg.go.dev/github.com/gorilla/websocket#hdr-Compression_EXPERIMENTAL)

permessage-deflate will work only if both client and server supports this extension - and successfully negotiates it through `Sec-Websocket-Extensions` header.

If `EnableCompression` is set to true, gorilla/websocket will automatically try to negotiate compression with peer. 
```
var upgrader = websocket.Upgrader{
    EnableCompression: true,
}
```
Upon successful negotiation any message received in compressed form will be automatically decompressed by gorilla/websocket. So compression/decompression is taken care by gorilla/websocket if its supported.

Also, use of compression is experimental and may result in decreased performance.

the official RFC7692 has only defined deflate algorithm for compression in the specification as of now ( deflate from permessage-deflate). In future, say both a client and server wants to support algorithm 'foo' then this has to be negotiated through extension 'permessage-foo'. Few newer ones being discussed by IETF are 'permessage-bzip2', 'permessage-lz4', 'permessage-snappy'. These have not yet made it into the official RFCs. I'm not sure if any browsers are having experimental support for additional permessage algorithms. 
More about in [this stackoverflow question](https://stackoverflow.com/questions/19298651/how-does-websocket-compress-messages#answer-19300336)

I'd like to propose a PR for enabling `EnableCompression: true` to make compression work wherever it's supported.

### DIFF
--- a/js/modules/k6/ws/ws.go
+++ b/js/modules/k6/ws/ws.go
@@ -163,9 +163,10 @@ func (*WS) Connect(ctx context.Context, url string, args ...goja.Value) (*WSHTTP
 		HandshakeTimeout: time.Second * 60, // TODO configurable
 		// Pass a custom net.DialContext function to websocket.Dialer that will substitute
 		// the underlying net.Conn with our own tracked netext.Conn
-		NetDialContext:  state.Dialer.DialContext,
-		Proxy:           http.ProxyFromEnvironment,
-		TLSClientConfig: tlsConfig,
+		NetDialContext:    state.Dialer.DialContext,
+		Proxy:             http.ProxyFromEnvironment,
+		TLSClientConfig:   tlsConfig,
+		EnableCompression: true,
 	}
 
 	start := time.Now()


codebien:
HI @cooliscool, thanks for your contribution :pray: 

As you noted, the feature is still marked as experimental and may result in decreased performance, we would prefer to not add it by default. Add a `compression` option to the [connect param](https://k6.io/docs/javascript-api/k6-ws/connect-url-params-callback) object can avoid unexpected use-cases for users. We already have the same thing for the [HTTP params](https://k6.io/docs/javascript-api/k6-http/params).

It would be useful also to add benchmarks with compression enabled and not.

cooliscool:
Thanks for the review @codebien 
Adding a `compression` option to `connect params` feels like the right way to do this. 
I'm working on this. 👍

cooliscool:
#### Added benchmarks - with and without compression

#### Here's what I did : 
ws handler will send a message of 1KB to the client after Upgrade, then close the websocket 

benchmarked this for the two scenarios - with and without compression

from the benchmark, time taken for a single operation is more - when compression is enabled - which could be explained by the overhead caused by encoding and decoding. (the advantages in network latency due to compression may not be captured here properly - because there's no real network,  everthing is in memory right?)


PS: I had to separate the two benchmarks into different functions because of the increased cyclop complexity.

Issue #1152, opened 2019-09-11
Create multi-message WebSockets test


imiric:
While working on fixing the flaky `TestSentReceivedMetrics` (PR #1149) we noticed that all our current WS tests only test sending one message and disconnecting.

This issue is to add one or more tests where more than one message is passed, which will require a change in [`websocketEchoHandler()`](https://github.com/loadimpact/k6/blob/1f913f80a7267b7474dfeb96768a6065cfe620b0/lib/testutils/httpmultibin/httpmultibin.go#L103-L121) to avoid disconnecting after the first message.

Since the WS handler can impact test flakiness, make sure to test this change thoroughly.

cooliscool:
Hi 👋 I'm interested to take this up.

PR #2184, opened 2021-10-13
More multi message tests for WebSockets

cooliscool:
In relation to #1152 
#### Explanation : 

Didn't modify the ws-echo handler in  httpmultibin, instead I extended httpmultibin with a tb.Mux.HandleFunc(...)

Here are the few tests I could add: 
1. **send_receive_multiple_ws**
    send and receive messages 'test1', 'test2', 'test3' in sequence then close.
2. **send_receive_multiple_wss**
    send and receive messages 'test1', and then 'test2', then close on a wss connection
3. **send_receive_text_binary**
    I thought of this scenario where multiple kind of messages (text and binary) are exchanged over the same socket. So, here's a test that sends a text message and a binary message in sequence.

Other than that, I'm a bit concerned about whether running these tests parallely using same runtime will cause a race condition, for test to fail. (randomly, only if some race succeeds)

And, please let me know if any of these tests needs to be modified, or if you'd like me to add more scenarios, or remove.

ALso, I've added a funtion `assertMetricEmittedCount()` for asserting the count of `ws_msgs_sent` and `ws_msgs_received` collected. I made the number of messages sent '3' in send_receive_multiple_ws because of unparam linter complaining about passing same value to assertMetricEmittedCount() always :/ but that extra param `count` could be useful later.

### DIFF
--- a/js/modules/k6/ws/ws_test.go
+++ b/js/modules/k6/ws/ws_test.go
@@ -383,7 +383,7 @@ func TestSession(t *testing.T) {
 			var msg2 = "test2"
 			var msg3 = "test3"
 			var allMsgsRecvd = false
-			var res = ws.connect("WSBIN_URL/ws-echo-multi", function(socket){
+			var res = ws.connect("WSBIN_URL/ws-echo-multi", (socket) => {
 				socket.on("open", () => {
 					socket.send(msg1)
 				})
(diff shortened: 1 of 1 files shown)


cooliscool:
sorry. closed this by mistake.

let me summarise the suggested changes here : 

1. replace instances of `assertMetricEmitted()` with `assertMetricEmittedCount()` and count = 1 to reduce the no. of helper functions. 

2. changing the `ws-echo` `HTTPMultiBin` handler to support multiple messages. instead of using a test specific handler. the gotcha here would be that existing tests depending on `ws-echo` shouldn't break. 

Its doable without making huge changes in the existing tests i believe.

imiric:
@cooliscool Don't worry about the `HTTPMultiBin` change, it's probably better that this uses an isolated handler. And we can move it later if we need it elsewhere.

PR #2190, opened 2021-10-19
influxdb: Concurrent PeriodicFlusher

### LINKED ISSUE #2185: influxdb: Concurrent flush operations for PeriodicFlusher
The current `PeriodicFlusher` helper flushes samples synchronously and it delegates the eventual concurrency to the specific Output. This is something required because the out-of-order sometimes is not accepted (e.g. writing files). This approach has the downside for the flusher to make it very dependent on the output's rate achieved.

Instead, there are cases (e.g InfluxDB), where out-of-order is not relevant so we can add a safe concurrent flush for them and speed up things when few flush requests will get a bad response time.

### Suggested implementation

Add a new field to the `PeriodicFlusher` struct for explicit the required concurrency limit, then spawn in `run` method the relative amount of goroutines.

A dedicated `NewAsyncPeriodicFlusher(limit int)` must be implemented to avoid a breaking change for the current API.

codebien:
* Restored the concurrency support for the InfluxDB output.
* Small optimization to avoid the flush process with an empty SampleContainer slice.

Closes #2185

### DIFF
--- a/output/influxdb/config.go
+++ b/output/influxdb/config.go
@@ -56,11 +56,20 @@ type Config struct {
 // NewConfig creates a new InfluxDB output config with some default values.
 func NewConfig() Config {
 	c := Config{
-		Addr:             null.NewString("http://localhost:8086", false),
-		DB:               null.NewString("k6", false),
-		TagsAsFields:     []string{"vu", "iter", "url"},
-		ConcurrentWrites: null.NewInt(10, false),
-		PushInterval:     types.NewNullDuration(time.Second, false),
+		Addr:         null.NewString("http://localhost:8086", false),
+		DB:           null.NewString("k6", false),
+		TagsAsFields: []string{"vu", "iter", "url"},
+		PushInterval: types.NewNullDuration(time.Second, false),
+
+		// The minimum value of pow(2, N) for handling a stressful situation
+		// with the default push interval set to 1s.
+		// Concurrency is not expected for the normal use-case,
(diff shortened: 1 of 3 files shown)


inancgumus:
On a side-note, can you also change the names of the file from `helpers.go` (vague), and put types to their related files `periodic_flusher.go` (for the `PeriodicFlusher` type and `sample_buffer.go` (for the `SampleBuffer` type).

na--:
Regarding the file name, in general I agree with @inancgumus, but in this specific case, I think `helpers.go` may be the more practical choice. Consider the use case of someone new to k6 and its outputs that wants to write an extension. 

If they look at https://pkg.go.dev/go.k6.io/k6/output, the file name doesn't matter, they mostly see the types, the filenames are at the very bottom (or if you click to see the implementation of something).

But if they open https://github.com/grafana/k6/tree/master/output, they see 3 very simple files that should be almost self-explanatory:
- `output/extensions.go` - things to do with output extensions
- `output/types.go` - some type definitions related to outputs
- `output/helpers.go` - helpers to write outputs 

Contrast this to something like `output/periodic_flusher.go` - nobody will know what this beast is and why it's needed until they actually look at the code

yorugac:
Just a suggestion: once the logic of spawning in `NewAsyncPeriodicFlusher` is finalized, it might make sense to add the test checking that AsyncPeriodicFlusher cannot spawn more than `concurrency` goroutines.

Issue #796, opened 2018-10-05
Tag metrics from different stages


na--:
Reading through [this issue](https://github.com/loadimpact/k6/issues/204), I liked the idea of automatically tagging the metrics from different stages, I think it could be useful when analyzing the results later. 

Also, instead (or in addition) of those system tags, we might want to add the option of users to add their own `key=value` tags to the stages, like they currently can do whole test runs or with individual metrics. Though this option would probably have to be part of the new configuration mechanism that would be required for the [arrival-rate based executor](https://github.com/loadimpact/k6/issues/550)

na--:
Cross-posting to note another use case for this feature, related to setting thresholds only for specific stages: https://github.com/loadimpact/k6/issues/1136#issuecomment-557488687

joseray:
@ppcano this is already implemented?

na--:
@joseray, no, it's not yet implemented. You can watch this issue, since we'll close it when we implemet the feature.

na--:
This issue would be useful, though I think it's best if we focus our energies on other similar ones that are more flexible. If we had https://github.com/loadimpact/k6/issues/1320, i.e. be able to get the number of the current stage for `ramping-vus` and `ramping-arrival-rate`, we'd be able to manually tag metrics we care about or use a [`group()`](https://k6.io/docs/using-k6/tags-and-groups).

And, depending on how we choose to implement https://github.com/loadimpact/k6/issues/884, https://github.com/loadimpact/k6/issues/1724 and https://github.com/loadimpact/k6/issues/1321, we'd get much more flexible tagging as well... I'm not going to close this issue, but I'll tag it as `evaluation needed`, since at this point it seems unlikely we'll implement it as is. And even if we implement it (a name for each stage seems a good idea to have in the config from a UX perspective), the implementation is likely to just be a shortcut based on the code for the issues I listed above.

robingustafsson:
Just leaving a note here that I spoke with a customer yd that wanted to be able to set thresholds on a specific stage (the steady-state after a ramp-up stage) and asked if this was possible. If this functionality existed it would be possible to also set up thresholds limited to a certain stage, such as a steady-state stage.

codebien:
Few updates for this issue:
* #2172 is going to be merged in the next few days that will introduce the ability to set a custom tag from the JS code using the `js/execution` API. Still not something automatic as requested from this issue but it helps for the same final goal.

```js
exec.vu.tags['stage'] = 'ramp-up'
```

[...]

PR #2199, opened 2021-10-25
Scenario's stage number

codebien:
Adding scenario's stage id information.
Export the property `exec.scenario.stage.id` that returns a zero-based value tracking for the current running stage.

An example using it for tracking the stage as a tag:
```js
import http from 'k6/http';

export const options = {
  stages: [
    { duration: '1m', target: 10 },
    { duration: '1m', target: 20 },
    { duration: '1m', target: 0 },
  ],
};

export default function () {
  exec.vu.tags["stage"] = 'stage#${exec.scenario.stage.id}'
  http.get('http://test.k6.io/');
}
```

### DIFF
--- a/js/modules/k6/execution/execution.go
+++ b/js/modules/k6/execution/execution.go
@@ -130,6 +130,21 @@ func (mi *ModuleInstance) newScenarioInfo() (*goja.Object, error) {
 		"iterationInTest": func() interface{} {
 			return vuState.GetScenarioGlobalVUIter()
 		},
+		"stage": func() interface{} {
+			stage, err := getScenarioState().CurrentStage()
+			if err != nil {
+				common.Throw(rt, err)
+			}
+			si := map[string]func() interface{}{
+				"number": func() interface{} { return stage.Index },
+				"name":   func() interface{} { return stage.Name },
+			}
+			obj, err := newInfoObj(rt, si)
+			if err != nil {
+				common.Throw(rt, err)
+			}
+			return obj
+		},
 	}
 
 	return newInfoObj(rt, si)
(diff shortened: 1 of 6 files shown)


MStoykov:
@na-- I am very much against `exec.scenario.stage.number==exec.scenario.stage.number` being able to return `false`.

I do think this will be *way* more confusing and add bugs than if `exec.scenario.stage.number` is just a constant.

All of the above other usecases, which I doubt are all that interesting for most users - I would expect most just want to know "vaguely" whether this iteration started in a given stage, after all as you've eluded that will change and it can change also over the course of a request, and we are not even doing asynchronous code yet. But I digress ... all of the above other usecases can be handled through js code, which is already possible as @codebien has shown. And as I have argued time and time again, adding functionality that can be implemented outside of k6 means that we still have to document it(which we have to do for the js code as well), but now we also can't change it, while we can always release another jslib showing how to use something else but as v0.2.0 ;) 

If anything in order to make it easier I am more for `exec.scenario.stages` being a "copy" of what the `stages` in the configuration for the given scenario are. This is unlikely to have to change and will be helpful for implementing the above use case as well as anything else.

Also again, can we discuss API changes with examples in *the issues* at least along implementing them if not before that?

na--:
I moved the milestone for this to v0.36.0. I disagree with you and think something like this is probably the best solution, but let's discuss it in https://github.com/grafana/k6/issues/2215 after we ship v0.35.0...

codebien:
This isn't in the current plan, we could reconsider this in the future but at the moment is not going to be merged. For the current plan check: https://github.com/grafana/k6/issues/796#issuecomment-1094897787

Issue #2203, opened 2021-10-27
Switch to NullDuration.TimeDuration helper


codebien:
Call `types.NullDuration.TimeDuration` helper instead convert it directly with `time.Duration(var)` a NullDuration option. Here an example:
https://github.com/grafana/k6/blob/a66aa5842c8a510f8b607a21eb22b7c33d176340/lib/executor/base_config.go#L103-L105

It should be replaced with `return bc.StartTime.TimeDuration()`

PR #2206, opened 2021-10-28
Use TimeDuration helper for NullDuration types

9a24f0:
This PR is an attempt to resolve GH-2203.
Any kind of feedback are welcome 🎉

### DIFF
--- a/cmd/config_consolidation_test.go
+++ b/cmd/config_consolidation_test.go
@@ -241,7 +241,7 @@ func getConfigConsolidationTestCases() []configConsolidationTestCase {
 		{opts{cli: []string{"-u", "1", "-i", "6", "-d", "10s"}}, exp{}, func(t *testing.T, c Config) {
 			verifySharedIters(I(1), I(6))(t, c)
 			sharedIterConfig := c.Scenarios[lib.DefaultScenarioName].(executor.SharedIterationsConfig)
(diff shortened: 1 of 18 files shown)


FloorD:
Hi @9a24f0 👋 we would love to send you a little k6 swag pack as per "thank you" for contributing to k6 this hacktoberfest. Please send your address details to floor[at]grafana.com and I'll take care it's shipped asap.

lib/types/types.go
```diff
@@ -230,7 +230,7 @@ func (d NullDuration) ValueOrZero() Duration {
 
 // TimeDuration returns a NullDuration's value as a stdlib Duration.
 func (d NullDuration) TimeDuration() time.Duration {
-	return time.Duration(d.Duration)
+	return d.TimeDuration()
```
MStoykov, line 233: This should be reverted as this is just calling the same function, effectively making an endless recursion

MStoykov (changes requested):

lib/types/types.go
9a24f0: Thank you so much! This is reverted in 047c231

codebien (approved):
LGTM, thanks

MStoykov (approved):
LGTM, thanks for making the changes @9a24f0 :bow:

na-- (approved):
Thanks!

PR #2209, opened 2021-10-29
cloudapi/logtail: Raise flush wait time for tests

codebien:
I don't see a way for making the `RestoreConnFromLatestMessage` test more stable using `HTTPMultiBin`. Considering that we already skipped the option to mock the WebSocket connection during the original PR, I just raised the timeout hoping it would be enough for the CI.


yorugac (approved):
My understanding is that this change is a "let's try this to see if it'd help a little", right? So :+1:

na-- (approved):
:crossed_fingers:

codebien:
>"let's try this to see if it'd help a little"

I'm not happy to say but yes, that's it.

Issue #1043, opened 2019-06-07
Support dumping of crypto keys to an nss formated key log file


MStoykov:
[NSS key log files](https://developer.mozilla.org/en-US/docs/Mozilla/Projects/NSS/Key_Log_Format) are more or less the only correct way that I have found to debug real HTTPS traffic. Given that HTTP2 is requiring TLS for all browsers (atm) I would guess practically 100% of the traffic generated by k6 is HTTPS one.

 Golang does [support it since 1.8](https://pkg.go.dev/crypto/tls#example-Config-KeyLogWriter), a thing that I just now realised as I have previously looked at the [original issue](https://github.com/golang/go/issues/13057). 

This will require a some way of specifying a file to dump to ... I propose that we add a cli flag and env variable and name it as Firefox this, but with k6 in front - `K6_SSLKEYLOGFILE`. 

Arguably this is [more involved](https://daniel.haxx.se/blog/2018/01/15/inspect-curls-tls-traffic/) than the `http-debug` but also has the benefits of using wireshark to debug web traffic which IMO is the more powerful option, and I would guess a familiar one for people who have needed it. 

This also ... automatically negates any shortcomings of the golang stdlib on reporting what it actually sends (#774 #986  #1042) and just looks at the actual data send through the wire.

PR #2210, opened 2021-10-29
feature: add KeyLogWriter runtime option and envVar

chtnnh:
@codebien if you could have a look that would be great!

Reference #1043

### DIFF
--- a/js/runner.go
+++ b/js/runner.go
@@ -31,6 +31,7 @@ import (
 	"net"
 	"net/http"
 	"net/http/cookiejar"
+	"os"
 	"runtime/debug"
 	"strconv"
 	"strings"
@@ -200,6 +201,15 @@ func (r *Runner) newVU(idLocal, idGlobal uint64, samplesOut chan<- stats.SampleC
 		Certificates:       certs,
 		NameToCertificate:  nameToCert,
 		Renegotiation:      tls.RenegotiateFreelyAsClient,
+	}
+	if keyWriter := r.Bundle.RuntimeOptions.KeyWriter; keyWriter.Valid {
+		f, err := os.OpenFile(keyWriter.String, os.O_WRONLY|os.O_CREATE|os.O_APPEND, 0600)
+		if err != nil {
+			return nil, err
+		}
+		defer f.Close()
+		tlsConfig.KeyLogWriter = f
+
 	}
 	transport := &http.Transport{
 		Proxy:               http.ProxyFromEnvironment,
--- a/lib/runtime_options.go
+++ b/lib/runtime_options.go
@@ -58,6 +58,7 @@ type RuntimeOptions struct {
 	NoThresholds  null.Bool   `json:"noThresholds"`
 	NoSummary     null.Bool   `json:"noSummary"`
 	SummaryExport null.String `json:"summaryExport"`
+	KeyWriter     null.String `json:"keyWriter"`
 }
 
 // ValidateCompatibilityMode checks if the provided val is a valid compatibility mode
--- a/cmd/runtime_options.go
+++ b/cmd/runtime_options.go
@@ -130,6 +130,12 @@ func getRuntimeOptions(flags *pflag.FlagSet, environment map[string]string) (lib
 		}
 	}
 
+	if envVar, ok := environment["K6_SSLKEYLOGFILE"]; ok {
+		if !opts.KeyWriter.Valid {
+			opts.KeyWriter = null.StringFrom(envVar)
+		}
+	}
+
 	if opts.IncludeSystemEnvVars.Bool { // If enabled, gather the actual system environment variables
 		opts.Env = environment
 	}


codebien:
Hi @chtnnh, thanks for your contribution! Can you fix the linter and add a test, please?

js/runner.go
```diff
@@ -201,6 +202,15 @@ func (r *Runner) newVU(idLocal, idGlobal uint64, samplesOut chan<- stats.SampleC
 		NameToCertificate:  nameToCert,
 		Renegotiation:      tls.RenegotiateFreelyAsClient,
 	}
+	if keyWriter := r.Bundle.RuntimeOptions.KeyWriter; keyWriter.Valid {
+		f, err := os.OpenFile(keyWriter.String, os.O_WRONLY|os.O_CREATE|os.O_APPEND, 0600)
+		if err != nil {
+			return nil, err
+		}
+		defer f.Close()
```
MStoykov, line 210: This will close the file before anything has had time to write to it. 

We either need to close it at some other time (there is currently no VU cleanup state, and I would argue that will be kind of hard to implement), or we can just not close it with a comment to actually figure out how to close it later on. Arguably in the fast majority of cases this file will stay open until just before k6 gets killed so :shrug:

lib/runtime_options.go
```diff
@@ -58,6 +58,7 @@ type RuntimeOptions struct {
 	NoThresholds  null.Bool   `json:"noThresholds"`
 	NoSummary     null.Bool   `json:"noSummary"`
 	SummaryExport null.String `json:"summaryExport"`
+	KeyWriter     null.String `json:"keyWriter"`
```
MStoykov, line 61:
```suggestion
	KeyWriter     null.String `json:"-"`
```

I don't think this is used anywhere, but still, it shouldn't be configurable from any place that is JSON ... or maybe it works with `config.json`, we need to test that :thinking: 
cc @na-- WDYT, should this even be configurable from there?

FloorD:
Hey @chtnnh 👋 we would love to send you a little k6 swag pack as per "thank you" for contributing to k6 this hacktoberfest. Please send your address details to floor[at]grafana.com and I'll take care it's shipped asap.

chtnnh:
@codebien @MStoykov could you give me a day to review your comments and submit another commit? In the middle of midterms at university :/

@FloorD thank you so much! Sent!

chtnnh:
Hey @codebien @MStoykov @na-- 

Just went through the comments, and I am unclear as to how to proceed. :/

Could I please get some pointers as to what direction I should take? Thanks!

js/runner.go
codebien: @MStoykov @chtnnh  I think we could open the file from the `run` command then inject it in the `js.Runner` populating a field in the case the option is available. 

```go
type Runner struct {
    NSSKeyLogWriter io.Writer
....
}
```
It should help also for creating a unit test.

lib/runtime_options.go
codebien: I would not encourage it so I think it should be better to not have it. But let's wait for the @na--'s opinion.

js/runner.go
MStoykov: While it will be nice to open this once and use it throughout, I think doing it in the `run` command will require way too many `lib.Runner` (the interface) changes that I am not particularly certain about and will just be one more thing this interface does (and it already is pretty huge IMO).

But this(the opening of the file) can definitely be moved to the creation of the js.Runner where we first get the RuntimeOptions and keep it as a field. But at least currently, looking at the code around all of this, there is no particular moment the `js.Runner` has no `Stop` method or some way to emulate it that will make the rest of k6 to wait for it to close the file after it is no longer necessary.

Alternatively, we have a context when we activate a VU and a [known place](https://github.com/grafana/k6/blob/b8164eed5413732e94f0f334b9a873146bd0f58b/js/runner.go#L658) where it(vu) no longer runs, but we haven't yet returned it, so the rest of the system will wait for us to ... for example close the file. Opening and closing a file on each activation though might not be great :shrug:. Also this means that we are using 1 file descriptor per VU just for that ... again not great.

We can probably combo the two, so that each time a VU is activated it will create the file and then use it and if it's the last one to be deactivated it will close it :shrug: , but IMO this goes in the same (if not bigger) rabbit hole as the adding of more methods to the `lib.Runner` interface and IMO is completely out of scope. 

I am also not certain that just because we are not running code, some (retry) logic in net/http (or something else as this should be used by all tls connections) will not just try to make a new connection and try to write to an already closed key log file.

So I am mostly for this to be moved in the `New*` methods returning the runner and having it in one place and just having a TODO+issue that we need to close this file. From my experiments, it works great even without closing it and the 1 more open file forever is probably the least problem around the runner ;) 

WDYT @na-- @imiric

MStoykov:
Hi @chtnnh, sorry for the really slow response. We did have some internal discussions at it seems most other contributors want the file to be closed. This unfortunately likely will mean a lot more code :(. 

We are currently near the end of the development cycle for v0.37.0 and will stop merging PRs ... soon. There have been some changes (and more to come) that will make the "when should I close the keylog file?" question easier to answer. 

What I am trying to say is that I am going to try to work on (or at least look into) to make your required changes as easy and small as possible. Unfortunately this will have to wait for the next release and might not even land there, depending on priorities.

But I haven't forgotten you :). Thanks for the contribution :bow:

js/runner.go
kkriegkxs: @MStoykov @na-- 
To echo my comment from my (duplicate as it turns out) solution: I think you may be overthinking this quite a bit :) just think of a use case: it is **only to debug** the comm with a few users at most - never to run at full load, because doing so will compromise the test infra. A few extra file handles on K6' side is nothing compared to the CAP portion (inspect streams, store/dump session) - IMO the only appropriate way to implement large scale capture, is by using an appropriately sized squid/nginx reverse proxy cluster with logging and post-processing.

MStoykov:
Thanks @chtnnh for this PR :bow:  

We have figured out a way to implement this in a better way, so I am closing this in favor of https://github.com/grafana/k6/pull/2487

PR #2216, opened 2021-11-02
Make ChangePathFs compatible with new afero.Fs without updating

MStoykov:
Given that we won't be using this method I don't think we need to have
actual real implementation for it which will require us to update
aforo.Fs, so I just returned an error.

This is predominantly motivated by the fact that an extension having a
dependency on a newer afero.Fs ( through another dependency for example)
will not compile as the interface afero.Fs was extended in v1.5.0.

### DIFF
--- a/lib/fsext/changepathfs.go
+++ b/lib/fsext/changepathfs.go
@@ -21,6 +21,7 @@
 package fsext
 
 import (
+	"errors"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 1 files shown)


na-- (approved):
hmm not ideal, but yeah, we don't require this and I doubt anyone is using k6 as a library to get `ChangePathFs`, :+1: :sweat_smile:

codebien (approved):

imiric (approved):
LGTM

Maybe before merging @dgzlopes can confirm this works with his extension by checking out this branch and setting `XK6_K6_REPO`?

MStoykov:
@imiric I already did, but I can wait for @dgzlopes to reconfirm as well :)

na--:
:shipit:, it's obvious this will be a problem sooner or later, even without the specific extension, simply because afero decided to make a breaking change without releasing a new major version... :man_facepalming:

PR #2220, opened 2021-11-02
Fix mixup between getting the length of a string and its runes

MStoykov:


na-- (approved):

log/loki.go
```diff
@@ -504,11 +504,12 @@ func (pushMsg *lokiPushMessage) WriteTo(w io.Writer) (n int64, err error) {
 			write(nanoseconds[:])
 			write([]byte(`",`))
 			if len([]rune(v.msg)) > pushMsg.maxSize {
-				difference := int64(len(v.msg) - pushMsg.maxSize)
+				msgRunes := []rune(v.msg)
```
na--, line 507: Actually, same suggestion here as in the other PR, move this before the `if` and then you can check for `len(msgRunes)` to avoid doing the conversion twice

na-- (changes requested):

na-- (approved):

yorugac (approved):

Issue #936, opened 2019-03-03
Any way to force to use HTTP/1.1 protocol


spaiz:
Just looked over the docs and didn't find any way to force the k6 not to switch to HTTP 2

I must do it cause we have a clients that won't use HTTP 2, so when I test I want to test the same protocol.

Tnx.

na--:
Unfortunately there's currently no way to force k6 to use HTTP/1.1 if the remote server supports HTTP/2 and the connection is over HTTPS. I think that's mostly a consequence of the way the Go standard library's HTTP client operates - it also [doesn't allow](https://github.com/golang/go/blob/95d4e6158b4199e1eee957e2c8c934d2cb86c35e/src/net/http/request.go#L132-L134) us to specify the HTTP version you want to use for a particular request.

That said, I think we can potentially slightly ameliorate this in k6 by adding the option to disable HTTP/2 globally, or at maybe at a per-VU/per-`group()` basis. Not sure how much refactoring would be needed for the more granular approaches, but the way we can do this is by modifying the `http.Transport` we use to make HTTP requests in k6. If we don't execute this line:
https://github.com/loadimpact/k6/blob/0b1b25949256544226cbe3641ad8024c06ef3420/js/runner.go#L171
k6 will be unable to make HTTP/2 requests. The trouble comes from the fact that this is executed once, in the beginning of initializing each VU, so it'd be somewhat difficult to get more granular than a global "don't allow any HTTP/2 requests" option...

spaiz:
Thanks for the answer.

Disallowing http/2 usage globally should be good enough. Cannot think about situation when I would like to test both protocols in the same benchmark session.

Just found in the go's [doc](https://github.com/golang/go/blob/95d4e6158b4199e1eee957e2c8c934d2cb86c35e/src/net/http/doc.go#L88):

> Starting with Go 1.6, the http package has transparent support for the
HTTP/2 protocol when using HTTPS. Programs that must disable HTTP/2
can do so by setting Transport.TLSNextProto (for clients) or
Server.TLSNextProto (for servers) to a non-nil, empty
map. Alternatively, the following GODEBUG environment variables are
currently supported:

	GODEBUG=http2client=0  # disable HTTP/2 client support
	GODEBUG=http2server=0  # disable HTTP/2 server support
	GODEBUG=http2debug=1   # enable verbose HTTP/2 debug logs
	GODEBUG=http2debug=2   # ... even more verbose, with frame dumps


[...]

PR #2222, opened 2021-11-03
feature/Force HTTP1 functionality

sjordhani22:
Greetings, 

We as a team of college students at WPI have been recently using K6 for some of our load-testing needs. In K6’s current implementation, it is not possible to force requests to be sent over the HTTP/1 protocol. We realized this is something that we would like to be able to do with K6 and while doing some research on the current issues, it seems this is something that others have requested as well (https://github.com/grafana/k6/issues/936). This PR attempts to address this issue by allowing users to specify a “--force-http1” flag on the command line, setting “ForceHTTP1” to true inside the options object in a script, or setting it as an environment variable. If enabled, this flag will send all HTTP requests using HTTP/1.1 - independent of the HTTP version supported by the endpoint. While a global option like this is not ideal, the code does allow people to select which protocol to use, is not very intrusive and can be easily removed in the future, and allows users to run their tests with HTTP/1.1 until the new HTTP implementation is complete. We are aware that you all want to see a new HTTP implementation, however, we believe that this change is small, straightforward, solves a major use case, and will allow for users to get tests in HTTP/1 running today. 

Additionally, we don’t mind updating the documentation to introduce this new feature, we would just need to know where to modify the documentation. 

Co-authored-by: Ashwin Pai 53979314+Ashpai24@users.noreply.github.com
Co-authored-by: Saniya Syeda 60491333+ssyeda422@users.noreply.github.com

### DIFF
--- a/js/runner.go
+++ b/js/runner.go
@@ -31,6 +31,7 @@ import (
 	"net"
 	"net/http"
 	"net/http/cookiejar"
+	"os"
 	"runtime/debug"
 	"strconv"
 	"strings"
@@ -211,8 +212,8 @@ func (r *Runner) newVU(idLocal, idGlobal uint64, samplesOut chan<- stats.SampleC
 		MaxIdleConnsPerHost: int(r.Bundle.Options.BatchPerHost.Int64),
 	}
 
-	if r.Bundle.Options.ForceHTTP1.Bool {
-		transport.TLSNextProto = map[string]func(string, *tls.Conn) http.RoundTripper{} //send req over h1 protocol
+	if ForceHTTP1() {
+		transport.TLSNextProto = map[string]func(string, *tls.Conn) http.RoundTripper{} //send over h1 protocol
 	} else {
 		_ = http2.ConfigureTransport(transport) //send over h2 protocol
 	}
@@ -265,6 +266,23 @@ func (r *Runner) newVU(idLocal, idGlobal uint64, samplesOut chan<- stats.SampleC
 	})
 
 	return vu, nil
+}
+
+// Will check if force http1 env variable has been set in order to force requests to be sent over h1
+// This feature is temporary until #936 is resolved
+func ForceHTTP1() bool {
+	godebug := os.Getenv("GODEBUG")
+	if godebug == "" {
+		return false
+	}
+	variables := strings.SplitAfter(godebug, ",")
+
+	for _, v := range variables {
+		if strings.Trim(v, ",") == "http2client=0" {
+			return true
+		}
+	}
+	return false
 }
 
 // Setup runs the setup function if there is one and sets the setupData to the returned value
--- a/js/runner.go
+++ b/js/runner.go
@@ -212,8 +212,8 @@ func (r *Runner) newVU(idLocal, idGlobal uint64, samplesOut chan<- stats.SampleC
 		MaxIdleConnsPerHost: int(r.Bundle.Options.BatchPerHost.Int64),
 	}
 
-	if ForceHTTP1() {
-		transport.TLSNextProto = map[string]func(string, *tls.Conn) http.RoundTripper{} //send over h1 protocol
+	if forceHTTP1() {
+		transport.TLSNextProto = make(map[string]func(string, *tls.Conn) http.RoundTripper) //send over h1 protocol
 	} else {
 		_ = http2.ConfigureTransport(transport) //send over h2 protocol
 	}
@@ -268,9 +268,9 @@ func (r *Runner) newVU(idLocal, idGlobal uint64, samplesOut chan<- stats.SampleC
 	return vu, nil
 }
 
-// Will check if force http1 env variable has been set in order to force requests to be sent over h1
-// This feature is temporary until #936 is resolved
-func ForceHTTP1() bool {
+// forceHTTP1 checks if force http1 env variable has been set in order to force requests to be sent over h1
+// TODO: This feature is temporary until #936 is resolved
+func forceHTTP1() bool {
 	godebug := os.Getenv("GODEBUG")
 	if godebug == "" {
 		return false
--- a/js/runner.go
+++ b/js/runner.go
@@ -210,8 +210,13 @@ func (r *Runner) newVU(idLocal, idGlobal uint64, samplesOut chan<- stats.SampleC
 		MaxIdleConns:        int(r.Bundle.Options.Batch.Int64),
 		MaxIdleConnsPerHost: int(r.Bundle.Options.BatchPerHost.Int64),
 	}
-	_ = http2.ConfigureTransport(transport)
 
+    if r.Bundle.Options.ForceHTTP1.Bool {
+	    transport.TLSNextProto = map[string]func(string, *tls.Conn) http.RoundTripper{} //send req over h1 protocol
+	} else {
+	    _ = http2.ConfigureTransport(transport) //send over h2 protocol
+	}
+	
 	cookieJar, err := cookiejar.New(nil)
 	if err != nil {
 		return nil, err
--- a/js/runner.go
+++ b/js/runner.go
@@ -211,12 +211,12 @@ func (r *Runner) newVU(idLocal, idGlobal uint64, samplesOut chan<- stats.SampleC
 		MaxIdleConnsPerHost: int(r.Bundle.Options.BatchPerHost.Int64),
 	}
 
-    if r.Bundle.Options.ForceHTTP1.Bool {
-	    transport.TLSNextProto = map[string]func(string, *tls.Conn) http.RoundTripper{} //send req over h1 protocol
+	if r.Bundle.Options.ForceHTTP1.Bool {
+		transport.TLSNextProto = map[string]func(string, *tls.Conn) http.RoundTripper{} //send req over h1 protocol
 	} else {
-	    _ = http2.ConfigureTransport(transport) //send over h2 protocol
+		_ = http2.ConfigureTransport(transport) //send over h2 protocol
 	}
-	
+
 	cookieJar, err := cookiejar.New(nil)
 	if err != nil {
 		return nil, err
--- a/js/runner.go
+++ b/js/runner.go
@@ -213,9 +213,9 @@ func (r *Runner) newVU(idLocal, idGlobal uint64, samplesOut chan<- stats.SampleC
 	}
 
 	if forceHTTP1() {
-		transport.TLSNextProto = make(map[string]func(string, *tls.Conn) http.RoundTripper) //send over h1 protocol
+		transport.TLSNextProto = make(map[string]func(string, *tls.Conn) http.RoundTripper) // send over h1 protocol
 	} else {
-		_ = http2.ConfigureTransport(transport) //send over h2 protocol
+		_ = http2.ConfigureTransport(transport) // send over h2 protocol
 	}
 
 	cookieJar, err := cookiejar.New(nil)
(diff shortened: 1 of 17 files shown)


na-- (changes requested):
Thanks for making this! Unfortunately, while the PR looks good technically, we can't merge it as it is... :disappointed: 
> While a global option like this is not ideal, the code does allow people to select which protocol to use, is not very intrusive and can be easily removed in the future, and allows users to run their tests with HTTP/1.1 until the new HTTP implementation is complete.

That's the crux of the problem - once we add such a global new option, however small, it can almost never be removed afterwards. Removing it will be a major breaking change, so we want to avoid adding a new option that is known from the start will be temporary... We are slowly chipping away on the issues around a better HTTP API, but we're not quite three yet :disappointed: 

As an alternative to your proposal, I can suggest bringing back support for `GODEBUG` (see bottom of [this section](https://pkg.go.dev/net/http#pkg-overview)). As I mentioned in https://github.com/grafana/k6/issues/936#issuecomment-470509886, because of the way we've implemented it with [golang.org/x/net/http2](https://pkg.go.dev/golang.org/x/net/http2#ConfigureTransport), k6 doesn't currently respect `GODEBUG=http2client=0`. Since that doesn't involve any long-term compatibility guarantees, we would probably be OK with a pull requests that adds support for it.

sjordhani22:
Thank you so much for getting back to us, @na-- ! We were wondering if we can get a little bit more clarity on: 

"...k6 doesn't currently respect GODEBUG=http2client=0. Since that doesn't involve any long-term compatibility guarantees, we would probably be OK with a pull requests that adds support for it."

If we remove the global option, and instead check for whether the environment variable is set [here](https://github.com/grafana/k6/pull/2222/files#diff-0f675fa3e61b737c161f4231ff48d85831738d5ec7aea1223315c16c994b2e44R214-R218), would this be something you would consider merging? 

Some pseudocode showing how we are imagining it: 

```
// psuedocode

disable_h1 := os.getEnv("GODEBUG")
if disable_h1 == "http2client=0" 
   // use h1
else
   /// configure_transport
end
```

na--:
Yeah, I think that would be ok to merge as a _temporary_ workaround, until we have a proper fix for this. It's far from ideal (e.g. I don't like the fact that we'll depend on the `os` package and global state in that part of the code), but it should be better than adding another explicit global k6 option :man_shrugging: 

Also, keep in mind that `GODEBUG` can have other things in it as well, so it's probably better to parse it in a separate function, with a comment that it's temporary until https://github.com/grafana/k6/issues/936 is resolved.

sjordhani22:
Sorry for the late response @na-- ! We made the changes you had requested, please let us know if this looks good or if there is anything else we must change. Thanks!

js/runner.go
```diff
@@ -262,6 +268,23 @@ func (r *Runner) newVU(idLocal, idGlobal uint64, samplesOut chan<- stats.SampleC
 	return vu, nil
 }
 
+// Will check if force http1 env variable has been set in order to force requests to be sent over h1
+// This feature is temporary until #936 is resolved
+func ForceHTTP1() bool {
```
codebien, line 273:
```suggestion
// forceHTTP1 checks if force http1 env variable has been set in order to force requests to be sent over h1
// TODO: This feature is temporary until #936 is resolved
func forceHTTP1() bool {
```
I think we can avoid exporting this function. As the linter is suggesting the comment must start with the function's name.

js/runner.go
```diff
@@ -210,7 +211,12 @@ func (r *Runner) newVU(idLocal, idGlobal uint64, samplesOut chan<- stats.SampleC
 		MaxIdleConns:        int(r.Bundle.Options.Batch.Int64),
 		MaxIdleConnsPerHost: int(r.Bundle.Options.BatchPerHost.Int64),
 	}
-	_ = http2.ConfigureTransport(transport)
+
+	if ForceHTTP1() {
+		transport.TLSNextProto = map[string]func(string, *tls.Conn) http.RoundTripper{} //send over h1 protocol
```
codebien, line 216:
```suggestion
		transport.TLSNextProto = make(map[string]func(string, *tls.Conn) http.RoundTripper) //send over h1 protocol
```
I hope in this way it can be a bit more readable

codebien (changes requested):
Hi @sjordhani22,
thanks for your contribution!

Added some small things in the comments, also can you add a test, please?

sjordhani22:
Thanks for the feedback, @codebien! I wrote two test cases that set the GODEBUG variable, initiate a request, and then check to see that the intended protocol was used. In the tests I send requests to `https://k6.io/`, please let me know if this is alright. I also made the changes you requested.

sjordhani22:
Greetings! 

I was taking a look at the lint errors that came back from the review and I made changes to fix those. One of the issues however, mentioned that the new tests that were written were not run in parallel. I had originally tried running the tests in parallel but it seems like when I did that, the force h1 feature was not working and the tests would fail. I think it has to do with the setting of the environment variable in the test. Any ideas on this? Thanks!

codebien (changes requested):
Hi @sjordhani22,
sorry for the late response, I think it's fine to set a `nolint` line for excluding the parallel check for this specific test. Generally, you would resolve a problem like this by having the env var injected but we decided to not have a new option so, for now, I think this is the best we can do. 

Apparently, the two tests are very similar in terms of code, I think we could improve them by creating a table test and injecting the params required instead to have one dedicated for each case. This should have the benefit of having only one `//nolint` directive.

js/runner_test.go
```diff
@@ -2074,113 +2074,91 @@ func TestMinIterationDurationIsCancellable(t *testing.T) {
...
+			script: `var k6 = require("k6");
+			var check = k6.check;
+			var fail = k6.fail;
+			var http = require("k6/http");;
+			exports.default = function() {
+				var res = http.get("https://k6.io/");
+				if (
+					!check(res, {
+					'checking to see if status was 200': (res) => res.status === 200,
+					'checking to see if protocol was H2': (res) => res.proto === 'HTTP/2.0'
```
codebien, line 2116:
```suggestion
					'checking to see if protocol was H2': (res) => res.proto === '%s'
```
I don't think we need to have the script repeated, you can just inject the expected protocol here as a parameter and move this as code as part of the run code.

js/runner_test.go
```diff
@@ -2074,113 +2074,91 @@ func TestMinIterationDurationIsCancellable(t *testing.T) {
...
-	})
-	require.NoError(t, err)
+// nolint:paralleltest
+func TestForceHTTP1Feature(t *testing.T) {
 
-	registry := metrics.NewRegistry()
-	builtinMetrics := metrics.RegisterBuiltinMetrics(registry)
-	r2, err := NewFromArchive(testutils.NewLogger(t), r1.MakeArchive(), lib.RuntimeOptions{}, builtinMetrics, registry)
-	require.NoError(t, err)
+	var cases = map[string]struct {
```
codebien, line 2080:
```suggestion
	cases := map[string]struct {
```

codebien (changes requested):
Hey @sjordhani22,
some small improvements, I think we are pretty close. Thanks for your effort. :pray:

sjordhani22:
@codebien, thanks for the feedback! Made those changes as well. 

Thanks!

codebien (dismissed):
LGTM :clap: 

Let's wait for a 2nd review by @na--, @oleiade I remember you were working on `js/runner` tests maybe you want to review this PR before we get it merged.

js/runner_test.go
```diff
@@ -2074,6 +2074,80 @@ func TestMinIterationDurationIsCancellable(t *testing.T) {
...
+					'checking to see if status was 200': (res) => res.status === 200,
+					'checking to see if protocol was H2': (res) => res.proto === '%s'
+					})
+				) {
+					fail('test failed')
+				}
+			}`, tc.protocol))
+			require.NoError(t, err)
+
+			err = r1.SetOptions(lib.Options{
```
oleiade, line 2123: As I asked myself "why?" when seeing it, I think it could be worth adding a comment here documenting why we set the `InsecureSkipTLSVerify` here?

js/runner_test.go
```diff
@@ -2074,6 +2074,80 @@ func TestMinIterationDurationIsCancellable(t *testing.T) {
...
+				require.NoError(t, err)
+			}()
+			assert.Equal(t, tc.expectedForceH1Result, forceHTTP1())
+
+			r1, err := getSimpleRunner(t, "/script.js", fmt.Sprintf(`var k6 = require("k6");
+			var check = k6.check;
+			var fail = k6.fail;
+			var http = require("k6/http");;
+			exports.default = function() {
+				var res = http.get("https://k6.io/");
```
oleiade, line 2111: Just a tiny nitpick, but I would find it best if we were using `https://test-api.k6.io/` here instead.

oleiade (approved):
LGTM 👍🏻 

A couple nitpicks that you can safely ignore if you prefer. Great work! 🙏🏻

js/runner_test.go
codebien: I agree, I forgot to comment about it, I would do one more step using https://pkg.go.dev/github.com/grafana/k6@v0.27.1/lib/testutils/httpmultibin so we can just use a local network.

sjordhani22: Greetings @oleiade & @codebien, 

I originally wanted to use the httpmultibin in the test cases as I noticed that other test cases were doing the same. However, when I use the httpmultibin endpoints and `Replacer.Replace(....)`, the requests seem to fail and I get an i/o timeout error (as seen in the screenshot). Any ideas why this may be happening? 

Screen Shot 2021-11-30 at 9 50 10 AM

MStoykov: @sjordhani22 I would guess you didn't configure the runner to use the mutlibin. Please post your code so we can diagnose more specifically the reason, but I would guess you didn't [overwrite hosts](https://github.com/grafana/k6/blob/master/js/runner_test.go#L1180) ?

sjordhani22: @MStoykov Thanks for the help, that was the issue!

sjordhani22:
Made those additional changes requested. Not sure why GitHub UI says I "dismissed stale reviews" after a push. 

Thanks!

na-- (approved):

na--:
Thanks, LGTM now. The stale review dismissal is because we've configured the GitHub repo so that every commit you make in the PR automatically dismisses the old PR reviews. This way we have to review the latest code before it's merged in `master`.

I see that one of the commit authors, @ssyeda422, hasn't signed the CLA: https://github.com/grafana/k6/pull/2222#issuecomment-959033106

That also needs to happen, or the their commit needs to be rewritten and the PR squashed, before we can merge the PR

MStoykov (approved):

ssyeda422:
Hi, sorry about that! Totally forgot to resign. I did that just now, let me know if it’s all set.

Thank you!

mstoykov:
Hi @marcelmfa I have opened https://github.com/grafana/k6-docs/issues/664 to document this, but you need to set `GODEBUG` env variable to `http2client=0` so `GODEBUG=http2client=0 k6 run script.js` for LInux/MacOs for windows it's different depending on what you use so see this [blog](http://www.dowdandassociates.com/blog/content/howto-set-an-environment-variable-in-windows-command-line-and-registry/) for example.

phizatmonster:
> 

hello @mstoykov and others, i'm facing the same question as @marcelmfa on how to run k6 with the GODEBUG env var, as my team is using windows OS.  i looked thru the blog link u posted, but i'm still unsure what the equivalent command to `GODEBUG=http2client=0 k6 run script.js` would be for windows cli.  thanks

PR #2223, opened 2021-11-03
Drop map structure

MStoykov:


js/compiler/compiler.go
```diff
@@ -218,15 +217,7 @@ func (b *babel) Transform(logger logrus.FieldLogger, src, filename string) (stri
...
-	var rawMap map[string]interface{}
-	if err = b.vm.ExportTo(vO.Get("map"), &rawMap); err != nil {
-		return code, nil, err
-	}
-	var srcMap SourceMap
-	if err = mapstructure.Decode(rawMap, &srcMap); err != nil {
-		return code, &srcMap, err
-	}
-	return code, &srcMap, err
+	return code, nil, err
```
na--, line 220: I don't think you now return anything other than `nil` as the second result here and in `Compiler.Transform()` :confused:

MStoykov: yeah, but I didn't want to remove it everywhere just to re-add it in #2082

na-- (commented):
:tada: Awesome, LGTM besides the minor nitpick I mentioned inline

js/compiler/compiler.go
na--: hmm fair enough, but at least put in a comment with a `TODO` and a link to the other PR before this line, now it looks very weird

na-- (approved):

inancgumus (approved):

MStoykov:
this btw drops around 120kb from the final binary

PR #2225, opened 2021-11-05
Update x/net with fixes from go1.17.3

MStoykov:


na-- (approved):
Those extensive HTTP/2 changes are a bit scary so :crossed_fingers:

codebien (approved):

PR #2226, opened 2021-11-08
Refactor `k6/http` and `k6/html` to the new JS `Module` API

na--:
I thought I needed to refactor both `k6/html` and `k6/http` together, since the latter depends on the former, but turns out that wasn't quite the case and the panic was caused by something else... :sweat_smile: Still, it doesn't hurt and the `k6/html` changes are relatively minor...

I like how this looks like for now, and I'll probably make another PR after this to move some things out of the [`js.Runner.newVu()`](https://github.com/grafana/k6/blob/dae70006ca77767e456acafbdda284d064dbd882/js/runner.go#L149) and `lib.State` and into the `k6/http` module, for example:
- `TLSConfig`, `Transport` and `CookieJar` probably should go in the `Client` and be initialized in a constructor, which will use the global `lib.Options` as the default options for its initialization
- `BPool` can either be in the `Client` or `ModuleInstance`, depends
- `RPSLimit` should probably be in the `RootModule`

As a prerequisite to these though, other built-in modules (and xk6 extensions) should be able to essentially do the equivalent of `require('k6/http')` internally, i.e. get the `ModuleInstance` for their VU. And to do that without Go import loops, we probably have to move some of the logic from [`js/InitContext`](https://github.com/grafana/k6/blob/dae70006ca77767e456acafbdda284d064dbd882/js/initcontext.go#L133) to [`js/common.InitEnvironment`](https://github.com/grafana/k6/blob/dae70006ca77767e456acafbdda284d064dbd882/js/common/initenv.go#L34)

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -38,6 +38,8 @@ linters-settings:
     min-confidence: 0
   gocyclo:
     min-complexity: 25
+  cyclop: #TODO: see https://github.com/grafana/k6/issues/2294, leave only one of these?
+    max-complexity: 25
   maligned:
     suggest-new: true
   dupl:
--- a/js/modules/k6/html/html.go
+++ b/js/modules/k6/html/html.go
@@ -21,7 +21,6 @@
 package html
 
 import (
-	"context"
 	"errors"
 	"fmt"
 	"strings"
@@ -31,20 +30,62 @@ import (
 	gohtml "golang.org/x/net/html"
 
 	"go.k6.io/k6/js/common"
+	"go.k6.io/k6/js/modules"
 )
 
-type HTML struct{}
+// RootModule is the global module object type. It is instantiated once per test
+// run and will be used to create k6/html module instances for each VU.
+type RootModule struct{}
 
-func New() *HTML {
-	return &HTML{}
+// ModuleInstance represents an instance of the HTML module for every VU.
+type ModuleInstance struct {
+	vu         modules.VU
+	rootModule *RootModule
+	exports    *goja.Object
 }
 
-func (HTML) ParseHTML(ctx context.Context, src string) (Selection, error) {
+var (
+	_ modules.Module   = &RootModule{}
+	_ modules.Instance = &ModuleInstance{}
+)
+
+// New returns a pointer to a new HTML RootModule.
+func New() *RootModule {
+	return &RootModule{}
+}
+
+// Exports returns the JS values this module exports.
+func (mi *ModuleInstance) Exports() modules.Exports {
+	return modules.Exports{
+		Default: mi.exports,
+	}
+}
+
+// NewModuleInstance returns an HTML module instance for each VU.
+func (r *RootModule) NewModuleInstance(vu modules.VU) modules.Instance {
+	rt := vu.Runtime()
+	mi := &ModuleInstance{
+		vu:         vu,
+		rootModule: r,
+		exports:    rt.NewObject(),
+	}
+	if err := mi.exports.Set("parseHTML", mi.parseHTML); err != nil {
+		common.Throw(rt, err)
+	}
+	return mi
+}
+
+func (mi *ModuleInstance) parseHTML(src string) (Selection, error) {
+	return ParseHTML(mi.vu.Runtime(), src)
+}
+
+// ParseHTML parses the provided HTML source into a Selection object.
+func ParseHTML(rt *goja.Runtime, src string) (Selection, error) {
 	doc, err := goquery.NewDocumentFromReader(strings.NewReader(src))
 	if err != nil {
 		return Selection{}, err
 	}
-	return Selection{rt: common.GetRuntime(ctx), sel: doc.Selection}, nil
+	return Selection{rt: rt, sel: doc.Selection}, nil
 }
 
 type Selection struct {
--- a/js/modules/k6/http/cookiejar.go
+++ b/js/modules/k6/http/cookiejar.go
@@ -21,7 +21,6 @@
 package http
 
 import (
-	"context"
 	"fmt"
 	"net/http"
 	"net/http/cookiejar"
@@ -30,27 +29,23 @@ import (
 	"time"
 
 	"github.com/dop251/goja"
-
 	"go.k6.io/k6/js/common"
 )
 
-// HTTPCookieJar is cookiejar.Jar wrapper to be used in js scripts
-type HTTPCookieJar struct {
-	// js is to make it not be accessible from inside goja/js, the json is because it's used if we return it from setup
-	Jar *cookiejar.Jar `js:"-" json:"-"`
-	ctx *context.Context
-}
+// ErrJarForbiddenInInitContext is used when a cookie jar was made in the init context
+// TODO: unexport this? there's no reason for this to be exported
+var ErrJarForbiddenInInitContext = common.NewInitContextError("Making cookie jars in the init context is not supported")
 
-func newCookieJar(ctxPtr *context.Context) *HTTPCookieJar {
-	jar, err := cookiejar.New(nil)
-	if err != nil {
-		common.Throw(common.GetRuntime(*ctxPtr), err)
-	}
-	return &HTTPCookieJar{jar, ctxPtr}
+// CookieJar is cookiejar.Jar wrapper to be used in js scripts
+type CookieJar struct {
+	moduleInstance *ModuleInstance
+	// js is to make it not be accessible from inside goja/js, the json is
+	// for when it is returned from setup().
+	Jar *cookiejar.Jar `js:"-" json:"-"`
 }
 
 // CookiesForURL return the cookies for a given url as a map of key and values
-func (j HTTPCookieJar) CookiesForURL(url string) map[string][]string {
+func (j CookieJar) CookiesForURL(url string) map[string][]string {
 	u, err := neturl.Parse(url)
 	if err != nil {
 		panic(err)
@@ -65,8 +60,8 @@ func (j HTTPCookieJar) CookiesForURL(url string) map[string][]string {
 }
 
 // Set sets a cookie for a particular url with the given name value and additional opts
-func (j HTTPCookieJar) Set(url, name, value string, opts goja.Value) (bool, error) {
-	rt := common.GetRuntime(*j.ctx)
+func (j CookieJar) Set(url, name, value string, opts goja.Value) (bool, error) {
+	rt := j.moduleInstance.vu.Runtime()
 
 	u, err := neturl.Parse(url)
 	if err != nil {
--- a/js/modules/k6/http/file.go
+++ b/js/modules/k6/http/file.go
@@ -21,7 +21,6 @@
 package http
 
 import (
-	"context"
 	"fmt"
 	"strings"
 	"time"
@@ -42,8 +41,8 @@ func escapeQuotes(s string) string {
 	return quoteEscaper.Replace(s)
 }
 
-// File returns a FileData parameter
-func (h *HTTP) File(ctx context.Context, data interface{}, args ...string) FileData {
+// File returns a FileData object.
+func (mi *ModuleInstance) file(data interface{}, args ...string) FileData {
 	// supply valid default if filename and content-type are not specified
 	fname, ct := fmt.Sprintf("%d", time.Now().UnixNano()), "application/octet-stream"
 
@@ -57,7 +56,7 @@ func (h *HTTP) File(ctx context.Context, data interface{}, args ...string) FileD
 
 	dt, err := common.ToBytes(data)
 	if err != nil {
-		common.Throw(common.GetRuntime(ctx), err)
+		common.Throw(mi.vu.Runtime(), err)
 	}
 
 	return FileData{
--- a/js/modules/k6/http/http.go
+++ b/js/modules/k6/http/http.go
@@ -21,101 +21,159 @@
 package http
 
 import (
-	"context"
+	"net/http"
+	"net/http/cookiejar"
 
+	"github.com/dop251/goja"
 	"go.k6.io/k6/js/common"
-	"go.k6.io/k6/lib"
+	"go.k6.io/k6/js/modules"
 	"go.k6.io/k6/lib/netext"
 	"go.k6.io/k6/lib/netext/httpext"
 )
 
-const (
-	HTTP_METHOD_GET     = "GET"
-	HTTP_METHOD_POST    = "POST"
-	HTTP_METHOD_PUT     = "PUT"
-	HTTP_METHOD_DELETE  = "DELETE"
-	HTTP_METHOD_HEAD    = "HEAD"
-	HTTP_METHOD_PATCH   = "PATCH"
-	HTTP_METHOD_OPTIONS = "OPTIONS"
+// RootModule is the global module object type. It is instantiated once per test
+// run and will be used to create HTTP module instances for each VU.
+//
+// TODO: add sync.Once for all of the deprecation warnings we might want to do
+// for the old k6/http APIs here, so they are shown only once in a test run.
+type RootModule struct{}
+
+// ModuleInstance represents an instance of the HTTP module for every VU.
+type ModuleInstance struct {
+	vu            modules.VU
+	rootModule    *RootModule
+	defaultClient *Client
+	exports       *goja.Object
+}
+
+var (
+	_ modules.Module   = &RootModule{}
+	_ modules.Instance = &ModuleInstance{}
 )
 
-// ErrJarForbiddenInInitContext is used when a cookie jar was made in the init context
-var ErrJarForbiddenInInitContext = common.NewInitContextError("Making cookie jars in the init context is not supported")
-
-// New returns a new global module instance
-func New() *GlobalHTTP {
-	return &GlobalHTTP{}
+// New returns a pointer to a new HTTP RootModule.
+func New() *RootModule {
+	return &RootModule{}
 }
 
-// GlobalHTTP is a global HTTP module for a k6 instance/test run
-type GlobalHTTP struct{}
+// NewModuleInstance returns an HTTP module instance for each VU.
+func (r *RootModule) NewModuleInstance(vu modules.VU) modules.Instance {
+	rt := vu.Runtime()
+	mi := &ModuleInstance{
+		vu:         vu,
+		rootModule: r,
+		exports:    rt.NewObject(),
+	}
+	mi.defineConstants()
 
-// NewModuleInstancePerVU returns an HTTP instance for each VU
-func (g *GlobalHTTP) NewModuleInstancePerVU() interface{} { // this here needs to return interface{}
-	return &HTTP{ // change the below fields to be not writable or not fields
-		TLS_1_0:                            netext.TLS_1_0,
-		TLS_1_1:                            netext.TLS_1_1,
-		TLS_1_2:                            netext.TLS_1_2,
-		TLS_1_3:                            netext.TLS_1_3,
-		OCSP_STATUS_GOOD:                   netext.OCSP_STATUS_GOOD,
-		OCSP_STATUS_REVOKED:                netext.OCSP_STATUS_REVOKED,
-		OCSP_STATUS_SERVER_FAILED:          netext.OCSP_STATUS_SERVER_FAILED,
-		OCSP_STATUS_UNKNOWN:                netext.OCSP_STATUS_UNKNOWN,
-		OCSP_REASON_UNSPECIFIED:            netext.OCSP_REASON_UNSPECIFIED,
-		OCSP_REASON_KEY_COMPROMISE:         netext.OCSP_REASON_KEY_COMPROMISE,
-		OCSP_REASON_CA_COMPROMISE:          netext.OCSP_REASON_CA_COMPROMISE,
-		OCSP_REASON_AFFILIATION_CHANGED:    netext.OCSP_REASON_AFFILIATION_CHANGED,
-		OCSP_REASON_SUPERSEDED:             netext.OCSP_REASON_SUPERSEDED,
-		OCSP_REASON_CESSATION_OF_OPERATION: netext.OCSP_REASON_CESSATION_OF_OPERATION,
-		OCSP_REASON_CERTIFICATE_HOLD:       netext.OCSP_REASON_CERTIFICATE_HOLD,
-		OCSP_REASON_REMOVE_FROM_CRL:        netext.OCSP_REASON_REMOVE_FROM_CRL,
-		OCSP_REASON_PRIVILEGE_WITHDRAWN:    netext.OCSP_REASON_PRIVILEGE_WITHDRAWN,
-		OCSP_REASON_AA_COMPROMISE:          netext.OCSP_REASON_AA_COMPROMISE,
-
+	mi.defaultClient = &Client{
+		// TODO: configure this from lib.Options and get rid of some of the
+		// things in the VU State struct that should be here. See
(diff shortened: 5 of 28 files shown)


na--:
I'll fix the linter errors and squash the commits later, but the code should be reasonably ready for review now.

js/modules/k6/http/http.go
```diff
@@ -127,3 +178,11 @@ func (*HTTP) URL(parts []string, pieces ...string) (httpext.URL, error) {
 	}
 	return httpext.NewURL(urlstr, name)
 }
+
+// Client represents a stand-alone HTTP client.
+//
+// TODO: move to its own file
+type Client struct {
+	moduleInstance   *ModuleInstance
```
MStoykov, line 186: This is a bit long maybe just `mi`

js/modules/k6/http/tls_test.go
```diff
@@ -1,3 +1,4 @@
+//go:build go1.12
```
MStoykov, line 1: we can just drop this IMO ;)

js/modules/k6/http/http.go
oleiade: Considering @na--'s point, I also prefer `moduleInstance`, and would avoid abbreviating for clarity's sake.

codebien: Could be just `module` be an option? Like `c.module.vu.State()`

na--: Hmm :thinking: It would be a bit misleading, since we technically have the `Root` module that makes module instances per VU, but with the new architecture that shouldn't really be embedded in anything besides the module instances themselves, so it would probably be fine :man_shrugging:

js/modules/k6/http/http.go
```diff
@@ -21,101 +21,152 @@
...
+//
+// TODO: add sync.Once for all of the deprecation warnings we might want to do
+// for the old k6/http APIs here, so they are shown only once in a test run.
+type RootModule struct{}
+
+// ModuleInstance represents an instance of the HTTP module for every VU.
+type ModuleInstance struct {
+	vu            modules.VU
+	rootModule    *RootModule
+	defaultClient *Client
```
codebien, line 45:
```suggestion
	client *Client
```
Do we expect to have a dedicated field for non-default? Otherwise I think we can remove it.

codebien (commented):
LGTM just some small suggestions, not too enthusiastic about the `ModuleInstance` <-> `Client` cycle dependency but I think we can't do better for now.

js/modules/k6/http/request.go
```diff
@@ -105,36 +70,49 @@ func (h *HTTP) Request(ctx context.Context, method string, url goja.Value, args
...
+// processResponse stores the body as an ArrayBuffer if indicated by
+// respType. This is done here instead of in httpext.readResponseBody to avoid
+// a reverse dependency on js/common or goja.
+func (c *Client) processResponse(resp *httpext.Response, respType httpext.ResponseType) {
+	if respType == httpext.ResponseTypeBinary && resp.Body != nil {
+		resp.Body = c.moduleInstance.vu.Runtime().NewArrayBuffer(resp.Body.([]byte))
+	}
+}
+
+func (c *Client) responseFromHttpext(resp *httpext.Response) *Response {
```
oleiade, line 105: renaming this with the HTTP acronym in capital letter would make it more readable IMO: `responseFromHTTPext`. If I remember correctly, it's also considered best practice to keep acronyms in capital letters.

oleiade (approved):
LGTM :+1: A couple of nitpicks that you should feel free to ignore. 

A general remark about a pattern I've observed in other places in the code: I often see private and public functions/methods/types appearing in a seamingly unordered fashion. I was always told, and exposed to the idea that especially in Go (as naming has an impact on the privateness/publicness of the code), it's important to have types/functions/methods appear in order of importance: public first, then private. Most essential/important first, less important/implementation details last. In my experience, being intentional about it makes a codebase much easier to understand, navigate and maintain. 

What do you folks think?

js/modules/k6/html/html.go
```diff
@@ -31,20 +30,59 @@ import (
...
+		common.Throw(rt, err)
+	}
+	return mi
+}
+
+func (mi *ModuleInstance) parseHTML(src string) (Selection, error) {
+	return ParseHTML(mi.vu.Runtime(), src)
+}
+
+func ParseHTML(rt *goja.Runtime, src string) (Selection, error) {
```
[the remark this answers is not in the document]
MStoykov: it's used in the `k6/http` code https://github.com/grafana/k6/pull/2226/files#diff-2ce2e12f5858af9c47fdd2835d26fe286c6424ebad263de29887b4df2cb97b5aR73 . Previous to that an empty `HTML` struct was created to call the function on it

js/modules/k6/http/http.go
MStoykov: I still kind of feel that `mi` will likely be pretty common occurrence as a name across internal modules that it will not be ... too hard to know what it is, you already use it , just not for the field ;) 

But we can always change it later

js/modules/k6/http/http.go
```diff
@@ -21,101 +21,152 @@
...
+	mustExport := func(name string, value interface{}) {
+		if err := mi.exports.Set(name, value); err != nil {
+			common.Throw(rt, err)
+		}
+	}
+
+	mustExport("url", mi.URL)
+	mustExport("CookieJar", mi.newCookieJar)
+	mustExport("cookieJar", mi.getVUCookieJar)
+	mustExport("file", mi.File) // TODO: deprecate or refactor?
```
MStoykov, line 85: Does File need to stay exported?

js/modules/k6/http/http.go
```diff
@@ -21,101 +21,152 @@
...
-	OCSP_REASON_SUPERSEDED             string `js:"OCSP_REASON_SUPERSEDED"`
-	OCSP_REASON_CESSATION_OF_OPERATION string `js:"OCSP_REASON_CESSATION_OF_OPERATION"`
-	OCSP_REASON_CERTIFICATE_HOLD       string `js:"OCSP_REASON_CERTIFICATE_HOLD"`
-	OCSP_REASON_REMOVE_FROM_CRL        string `js:"OCSP_REASON_REMOVE_FROM_CRL"`
-	OCSP_REASON_PRIVILEGE_WITHDRAWN    string `js:"OCSP_REASON_PRIVILEGE_WITHDRAWN"`
-	OCSP_REASON_AA_COMPROMISE          string `js:"OCSP_REASON_AA_COMPROMISE"`
+func (mi *ModuleInstance) Exports() modules.Exports {
+	return modules.Exports{
+		Default: mi.exports,
+		// TODO: add new HTTP APIs like Client, Request, etc. as named exports?
```
MStoykov, line 116: see https://github.com/grafana/k6/issues/2258 as it was more or less written for this module :rofl:

js/modules/k6/http/http.go
```diff
@@ -21,101 +21,152 @@
...
+// TODO: deprecate these?
+func (mi *ModuleInstance) defineConstants() {
+	rt := mi.vu.Runtime()
+	mustAddProp := func(name, val string) {
+		err := mi.exports.DefineDataProperty(
+			name, rt.ToValue(val), goja.FLAG_FALSE, goja.FLAG_FALSE, goja.FLAG_TRUE,
+		)
+		if err != nil {
+			common.Throw(rt, err)
+		}
```
MStoykov, line 129: I feel like this is the one use case for using `goja.Object` as the exports - being able to make things not act just as a simple map, as in this case not being able to overwrite this constants.

The question is should you not be able to overwrite this but be able to overwrite the functions ?
Should we return something different that is more robust in handling such cases 

See https://github.com/grafana/k6/pull/2241#issuecomment-977836690 (And above comments) for discussion on why I prefer `map[string]interface{}` for most cases

MStoykov (approved):
LGTM in general, I left some comments I do think we should discuss what we want to do with Exports more in general soon so that we can align all internal modules within this release

lib/state.go
```diff
@@ -56,7 +56,7 @@ type State struct {
 	// Networking equipment.
 	Transport http.RoundTripper
 	Dialer    DialContexter
-	CookieJar *cookiejar.Jar
+	CookieJar *cookiejar.Jar // TODO: move this to the k6/http ModuleInstance?
```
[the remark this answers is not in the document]
na--: True, but I'd like to handle this by being able to get the `ModuleInstance` of one built-in module (e.g. `k6/http` in this case) from another module (e.g. `k6/ws`). That way we wouldn't need to keep expanding the `lib.State` to contain any possible VU state. Instead, every module will contain the per-VU state in its own `ModuleInstance` and other modules will use that.

js/modules/k6/http/request_test.go
```diff
@@ -129,9 +129,9 @@ func assertRequestMetricsEmittedSingle(t *testing.T, sampleContainer stats.Sampl
 	}
 }
 
-func newRuntime(
-	t testing.TB,
-) (*httpmultibin.HTTPMultiBin, *lib.State, chan stats.SampleContainer, *goja.Runtime, *context.Context) {
+func newRuntime(t testing.TB) (
+	*httpmultibin.HTTPMultiBin, *lib.State, chan stats.SampleContainer, *goja.Runtime, *ModuleInstance,
+) {
```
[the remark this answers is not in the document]
na--: Yeah, that would be much better, but I'd like to do it in another PR... This one is so short mostly because I was able to keep the same number of results to `newRuntime()` and most places just didn't use the context and had `_` in there :sweat_smile:

samples/http_get.js
```diff
@@ -1,5 +1,5 @@
 import http from 'k6/http';
 
 export default function () {
-  const response = http.get("https://test-api.k6.io/");
+  http.get('https://test-api.k6.io/');
```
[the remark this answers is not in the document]
na--: probably forgot to restore the file to the original condition after testing with it :sweat_smile: but now that it's here, I think it should stay, it's stupid to assign to a variable and not use it afterwards

js/modules/k6/http/http.go
na--: Yes, we expect to be able to create non-default HTTP clients with different options and transports in the near (:crossed_fingers:) future. So you'd be able do to something like this:
```js
import http from 'k6/http';
let myAwesomeTransport = new http.Transport({forceHTTP1: true}) // or something like this
let myAwesomeClient = new http.Client(myAwesomeTransport);
```

js/modules/k6/http/request.go
```diff
@@ -48,49 +46,16 @@ var ErrHTTPForbiddenInInitContext = common.NewInitContextError("Making http requ
...
-// Del makes an HTTP DELETE and returns a corresponding response by taking goja.Values as arguments
-func (h *HTTP) Del(ctx context.Context, url goja.Value, args ...goja.Value) (*Response, error) {
-	return h.Request(ctx, HTTP_METHOD_DELETE, url, args...)
-}
-
-// Options makes an HTTP OPTIONS request and returns a corresponding response by taking goja.Values as arguments
-func (h *HTTP) Options(ctx context.Context, url goja.Value, args ...goja.Value) (*Response, error) {
-	return h.Request(ctx, HTTP_METHOD_OPTIONS, url, args...)
+func (c *Client) getMethodClosure(method string) func(url goja.Value, args ...goja.Value) (*Response, error) {
+	return func(url goja.Value, args ...goja.Value) (*Response, error) {
```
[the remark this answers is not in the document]
na--: Even though it's a single line, I wouldn't like to repeat it several times: https://github.com/grafana/k6/blob/c6da5bc56531cedf2a8f694bf3121e96a7cad4b5/js/modules/k6/http/http.go#L94-L99

and adding a lambda to that already noisy function also doesn't seem like a good idea, so this seems like the least bad way to go about it :sweat_smile:

js/modules/k6/http/http.go
na--: I started by reading that issue, but I am still not sure exactly what you mean, sorry :disappointed: Do you mean I should just return everything as named exports here and rely on this? https://github.com/grafana/k6/blob/5c14dae129a5eb6ed309c94a194370878b7b08e5/js/initcontext.go#L180-L182

js/modules/k6/http/http.go
na--: Hmm :thinking: I can understand the `map[string]interface{}` preference, but goja will just convert that to an `Object` anyway. And I like being able to have constants be constant :sweat_smile: 

Do you know hot this should work with proper ES modules? If you have `export const something`, does that actually make it constant in the code that imports it?

js/modules/k6/http/http.go
na--: Resolved by https://github.com/grafana/k6/pull/2226/commits/d87b4d6721568af0d3d6a61ec6272d0ae3f7ccbe

js/modules/k6/http/request.go
na--: Resolved by https://github.com/grafana/k6/pull/2226/commits/d87b4d6721568af0d3d6a61ec6272d0ae3f7ccbe

js/modules/k6/http/tls_test.go
na--: Resolved by https://github.com/grafana/k6/pull/2226/commits/d87b4d6721568af0d3d6a61ec6272d0ae3f7ccbe

js/modules/k6/http/http.go
```diff
@@ -21,101 +21,153 @@
...
+	mustAddProp("OCSP_REASON_CERTIFICATE_HOLD", netext.OCSP_REASON_CERTIFICATE_HOLD)
+	mustAddProp("OCSP_REASON_REMOVE_FROM_CRL", netext.OCSP_REASON_REMOVE_FROM_CRL)
+	mustAddProp("OCSP_REASON_PRIVILEGE_WITHDRAWN", netext.OCSP_REASON_PRIVILEGE_WITHDRAWN)
+	mustAddProp("OCSP_REASON_AA_COMPROMISE", netext.OCSP_REASON_AA_COMPROMISE)
 }
 
-// XCookieJar creates a new cookie jar object.
-func (*HTTP) XCookieJar(ctx *context.Context) *HTTPCookieJar {
-	return newCookieJar(ctx)
+func (mi *ModuleInstance) newCookieJar(call goja.ConstructorCall, rt *goja.Runtime) *goja.Object {
```
codebien, line 152:
```suggestion
func (mi *ModuleInstance) newCookieJar(call goja.ConstructorCall) *goja.Object {
```
:thinking: Is this expected to be a constructor? Is this form accepted, I'm not seeing it in the doc.

js/modules/k6/http/http.go
codebien: What I meant is that from my understanding `ModuleInstance` will have only a unique field as `client` so we could avoid to explicit `default` because that should be clear and it will be wrapped behind the exported methods `get/post/head etc...`. Btw it's just a preference, not really required.

lib/state.go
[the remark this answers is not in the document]
codebien: I think you should add your comment to the code, it should give more context to the TODO

codebien (changes requested):
Thanks for the effort in doing it, I mostly omitted to review the part related to the `exports` because I would wait for updates from @MStoykov experiments before providing any input there.

The rest is mostly asking for clarifications about TODOs.

js/modules/k6/http/http.go
MStoykov: maybe [this](https://pkg.go.dev/github.com/dop251/goja#readme-native-constructors) will help or do you mean in the `k6/http` [doc](https://k6.io/docs/javascript-api/k6-http/cookiejar) ?

js/modules/k6/http/http.go
MStoykov: > If you have export const something, does that actually make it constant in the code that imports it? 

yes

lib/netext/httpext/request_test.go
```diff
@@ -207,7 +211,7 @@ func TestResponseStatus(t *testing.T) {
 					ResponseType: ResponseTypeNone,
 				}
 
-				res, err := MakeRequest(ctx, preq)
+				res, err := MakeRequest(context.Background(), state, preq)
```
MStoykov, line 214: can this be any context that will be canceled/timeouted in some time

js/modules/k6/http/http_test.go
```diff
@@ -46,6 +79,7 @@ func TestTagURL(t *testing.T) {
 	for expr, data := range testdata {
 		expr, data := expr, data
 		t.Run("expr="+expr, func(t *testing.T) {
+			t.Parallel()
```
MStoykov, line 82: How is this not panicking when it's using the same runtime for all 5 cases?

MStoykov (commented):
LGTM apart from the tests changes. I would expect that there will be some trouble with the tests either way so :shrug:

js/modules/k6/http/http.go
codebien: ok, so `goja` has two ways for defining a constructor:

```
func(ConstructorCall, *Runtime) *Object
```
or
```
func(ConstructorCall) *Object
```
@MStoykov Which one should we prefer in this case where both could work? Is the first one better in terms of performance? I need to check it in the meanwhile but maybe you already know the response.

MStoykov: I don't think there is any performance difference, the first one just gives you access to the `Runtime` which we already have though, so no need to use that one :)

codebien: In the second version we should use the `mi.vu.Runtime()` (to be get from the context) call for resolving the runtime so it means one more call, for that reason I was guessing that the first (already used) version would be faster.

.golangci.yml
```diff
@@ -38,6 +38,8 @@ linters-settings:
     min-confidence: 0
   gocyclo:
     min-complexity: 25
+  cyclop: #TODO: leave only one of these?
```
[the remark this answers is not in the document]
na--: https://github.com/grafana/k6/issues/2294

js/modules/k6/http/file.go
```diff
@@ -42,8 +41,10 @@ func escapeQuotes(s string) string {
 	return quoteEscaper.Replace(s)
 }
 
-// File returns a FileData parameter
-func (h *HTTP) File(ctx context.Context, data interface{}, args ...string) FileData {
+// File returns a FileData object.
+//
+// TODO: deprecate?
```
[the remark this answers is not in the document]
na--: This was actually a mistake, we can't deprecate this, so I'll remove the comment :disappointed:

js/modules/k6/http/http.go
```diff
@@ -21,101 +21,153 @@
...
+// Exports returns the JS values this module exports.
+func (mi *ModuleInstance) Exports() modules.Exports {
+	return modules.Exports{
+		Default: mi.exports,
+		// TODO: add new HTTP APIs like Client, Request, etc. as named exports?
+	}
+}
 
-	responseCallback func(int) bool
+// TODO: deprecate these?
```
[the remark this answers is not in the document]
na--: Same with this

js/modules/k6/http/file.go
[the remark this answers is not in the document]
MStoykov: we can deprecate it ... we probably can't remove it ;)

js/modules/k6/http/http.go
MStoykov: in that sense - yes @codebien probably using the first one is a bit faster, but IMO as we use the `mi.vu` to get everything else and it already has the runtime it makes more sense to just use that. 

>(to be get from the context) 

this is currently an implementation detail that will change, just the theoretical slowness of that probably doesn't matter for most cases while the fix for that will require quite a bit of reworking of `js` module ... last I looked at it.

So in retrospect you are correct, but IMO it's better to just get the runtime from `mi.vu` everywhere including in constructors where it's possible for goja to give it to us.

js/modules/k6/http/http_test.go
na--: Weird, I don't know, but I'll fix it

js/modules/k6/http/cookiejar.go
```diff
@@ -30,27 +29,22 @@ import (
...
 )
 
-// HTTPCookieJar is cookiejar.Jar wrapper to be used in js scripts
-type HTTPCookieJar struct {
-	// js is to make it not be accessible from inside goja/js, the json is because it's used if we return it from setup
-	Jar *cookiejar.Jar `js:"-" json:"-"`
-	ctx *context.Context
-}
+// ErrJarForbiddenInInitContext is used when a cookie jar was made in the init context
+var ErrJarForbiddenInInitContext = common.NewInitContextError("Making cookie jars in the init context is not supported")
```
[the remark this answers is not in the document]
codebien: ok, then it would be ideal to add a TODO comment for it

lib/netext/httpext/request_test.go
na--: Resolved by the [latest force push](https://github.com/grafana/k6/compare/f072e4d180fb41c3686dd0c44a6ae8fbd0783193..18919f3f0f95645ace7c0a7ba1f71423978509e8)

js/modules/k6/http/http_test.go
na--: Resolved by the [latest force push](https://github.com/grafana/k6/compare/f072e4d180fb41c3686dd0c44a6ae8fbd0783193..18919f3f0f95645ace7c0a7ba1f71423978509e8)

js/modules/k6/http/http.go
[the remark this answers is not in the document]
na--: Resolved by the [latest force push](https://github.com/grafana/k6/compare/f072e4d180fb41c3686dd0c44a6ae8fbd0783193..18919f3f0f95645ace7c0a7ba1f71423978509e8)

js/modules/k6/http/file.go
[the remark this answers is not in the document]
na--: Resolved by the [latest force push](https://github.com/grafana/k6/compare/f072e4d180fb41c3686dd0c44a6ae8fbd0783193..18919f3f0f95645ace7c0a7ba1f71423978509e8)

js/modules/k6/http/cookiejar.go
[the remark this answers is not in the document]
na--: Agree, added in https://github.com/grafana/k6/compare/18919f3f0f95645ace7c0a7ba1f71423978509e8..8a475aba3b2f096509f7d0f41752f0a02738f7ef

MStoykov (approved):

js/modules/k6/http/http.go
na--: Hmm do you also want me to change this? https://github.com/grafana/k6/blob/8a475aba3b2f096509f7d0f41752f0a02738f7ef/js/modules/k6/http/http.go#L165

na--: Assumed yes, so fixed both in https://github.com/grafana/k6/compare/8a475aba3b2f096509f7d0f41752f0a02738f7ef..2d45769d19a3fa08c9b1aa0a0340c8a4f63f81a0 :sweat_smile:

codebien (approved):
LGTM (ofc with green CI), thanks for doing it :pray:

MStoykov (dismissed):

oleiade (dismissed):
Just went through the PR again. Realized I forgot to approve it in the first place. Sorry for that 🙏🏻 Great work 🤝 🎉

MStoykov (approved):

codebien (approved):

PR #2229, opened 2021-11-08
Fix tests after min TLS version increase in golang tip

MStoykov:


imiric (approved):

imiric:
@MStoykov Just maybe explain in the commit and the PR description why we're doing this. ;)

yorugac (approved):

PR #2232, opened 2021-11-09
js/ws: modulev2 migration

codebien:
`js/ws` module migration implementing the ModuleV2 API.

The `Connect` method requires some refactoring to remove the linter issues (too many statements and cyclomatic complexity). I tried to fix it directly here but from my exploration, it requires too many changes compared to this, so I would prefer to have a separated PR for refactoring instead unique one.

### DIFF
--- a/js/modules/k6/ws/ws.go
+++ b/js/modules/k6/ws/ws.go
@@ -37,16 +37,52 @@ import (
 	"github.com/gorilla/websocket"
 
 	"go.k6.io/k6/js/common"
+	"go.k6.io/k6/js/modules"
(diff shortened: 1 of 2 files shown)


codebien:
We need to fix `./samples/websocket.js` but in the meanwhile the `ws` version of https://test-api.k6.io can be used.

imiric (approved):
LGTM :+1:

Agree with refactoring `Connect` in a separate PR. The `ws` package is long overdue for some cleanup.

MStoykov (approved):

oleiade (approved):
LGTM 👍🏻 🐻

PR #2239, opened 2021-11-15
js/crypto: ModuleV2 migration

codebien:
Migrated `k6/crypto` to the `modules.Module` (aka modules.ModuleV2) interface.

### DIFF
--- a/js/modules/k6/crypto/crypto.go
+++ b/js/modules/k6/crypto/crypto.go
@@ -21,7 +21,6 @@
 package crypto
 
 import (
-	"context"
 	"crypto/hmac"
 	"crypto/md5"
 	"crypto/rand"
@@ -31,6 +30,7 @@ import (
 	"encoding/base64"
 	"encoding/hex"
 	"errors"
+	"fmt"
 	"hash"
 
 	"golang.org/x/crypto/md4"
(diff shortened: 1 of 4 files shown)


js/modules/k6/crypto/crypto.go
```diff
@@ -68,114 +115,157 @@ func (*Crypto) RandomBytes(ctx context.Context, size int) *goja.ArrayBuffer {
 	return &ab
 }
 
-// Md4 returns the MD4 hash of input in the given encoding.
-func (c *Crypto) Md4(ctx context.Context, input interface{}, outputEncoding string) interface{} {
-	hasher := c.CreateHash(ctx, "md4")
+// MD4 returns the MD4 hash of input in the given encoding.
+func (c *Crypto) MD4(input interface{}, outputEncoding string) interface{} {
```
MStoykov, line 119: You could just as well make them unexported at this point. I am not particularly certain they should be used by some other go module :shrug:

js/modules/k6/crypto/crypto.go
```diff
@@ -39,23 +39,70 @@ import (
...
+	mustExport("md5", mi.MD5)
+	mustExport("randomBytes", mi.RandomBytes)
+	mustExport("ripemd160", mi.Ripemd160)
+	mustExport("sha1", mi.SHA1)
+	mustExport("sha256", mi.SHA256)
+	mustExport("sha384", mi.SHA384)
+	mustExport("sha512", mi.SHA512)
+	mustExport("sha512_224", mi.SHA512224)
+	mustExport("sha512_256", mi.SHA512256)
+	mustExport("hexEncode", mi.HexEncode)
```
MStoykov, line 93: I kind of feel like this will be better of just as `map[string]interface{}` in the `Exports`. Is there anything, in particular, we are winning by doing it this way?

js/modules/k6/crypto/x509/x509.go
```diff
@@ -86,47 +131,42 @@ type PublicKey struct {
...
 }
 
-// GetIssuer extracts certificate issuer
-func (X509) GetIssuer(ctx context.Context, encoded []byte) Issuer {
+// Issuer extracts certificate issuer
+func (mi X509) Issuer(encoded []byte) Issuer {
 	parsed, err := parseCertificate(encoded)
 	if err != nil {
-		throw(ctx, err)
+		common.Throw(mi.vu.Runtime(), err)
```
MStoykov, line 160: I kind of prefer to return `error` instead, it does the same thing, but it still IMO looks better, so we might think about it.

MStoykov (approved):
LGTM, I have left some comments, but none of those are blocking IMO

oleiade (dismissed):
LGTM 👍🏻

js/modules/k6/crypto/crypto.go
```diff
@@ -278,6 +281,6 @@ func (hasher *Hasher) Digest(outputEncoding string) (interface{}, error) {
 		return &ab, nil
 
 	default:
-		return nil, fmt.Errorf("Invalid output encoding: %s", outputEncoding)
+		return nil, fmt.Errorf("invalid output encoding: %s", outputEncoding)
```
codebien, line 284: This change makes the linter happy, is the capital here required?

MStoykov: The linter is unhappy because errors should not be with a capital letter with the idea that they will be composed and not just a thing that will be returned to the user.  But we are in the *second* situation - this will be the message to the user that they will see in this particular case as an exception and in that case IMO as well as @na-- (from all the discussions we have had around other cases like that) it is better for it to be with a capital letter.

I do actually feel that it's time we disable this particular lint error ?

codebien: Is not possible to delegate some capital operation in a centralized place before returning it to the user? In that way, we could just exec something like `toUpper(s[0])` and preserve the Go way of composing error messages.

MStoykov (approved):

MStoykov (approved):

oleiade (approved):

PR #2240, opened 2021-11-16
Update tc39 tests

MStoykov:
This will end up being squashed in 1 commit it mostly exists as many so that `breaking_test_errors.json` can be seen changing

Here is a [list](https://gist.github.com/MStoykov/0ba38c3f23b7b095c631e5c121a973cb) of all the "new" Skipped tests. New here is defined as from after I updated tc39 (first commit) to the latest one where I disable a bunch of the ones that have no chance to run. 
If you see a test about something you have no idea what it is ... it likely is still a proposal ;)


oleiade (approved):
LGTM 👍🏻

js/tc39/tc39_test.go
```diff
@@ -51,6 +51,45 @@ var (
...
+		"numeric-separator-literal",
+		"Object.fromEntries",
+		"Object.hasOwn",
+		"__getter__",
+		"__setter__",
+		"ShadowRealm",
+		"SharedArrayBuffer",
+		"error-cause",
+		"resizable-arraybuffer", // stage 3 as of 2021 https://github.com/tc39/proposal-resizablearraybuffer
+
```
codebien, line 88: Out of curiosity, is this newline relevant for separating something? If yes, I would add a comment as you did before for `goja`.

codebien (commented):
LGTM, just a comment.

js/tc39/tc39_test.go
MStoykov: just to separate them, the above list is from goja, while the below is stuff I added (goja doesn't run those through different means)

codebien (approved):

PR #2241, opened 2021-11-16
js/encoding: ModuleV2 migration

codebien:
Migrated `k6/encding` to the `modules.Module` (aka `modules.ModuleV2`) interface.

### DIFF
--- a/js/modules/k6/encoding/encoding.go
+++ b/js/modules/k6/encoding/encoding.go
@@ -21,24 +21,55 @@
 package encoding
 
 import (
-	"context"
 	"encoding/base64"
 
 	"go.k6.io/k6/js/common"
+	"go.k6.io/k6/js/modules"
 )
 
-type Encoding struct{}
+type (
+	// RootModule is the global module instance that will create module
+	// instances for each VU.
+	RootModule struct{}
 
-func New() *Encoding {
-	return &Encoding{}
+	// Encoding represents an instance of the encoding module.
+	Encoding struct {
+		vu modules.VU
+	}
+)
+
+var (
+	_ modules.Module   = &RootModule{}
+	_ modules.Instance = &Encoding{}
+)
+
+// New returns a pointer to a new RootModule instance.
+func New() *RootModule {
+	return &RootModule{}
 }
 
-// B64encode returns the base64 encoding of input as a string.
+// NewModuleInstance implements the modules.Module interface to return
+// a new instance for each VU.
+func (*RootModule) NewModuleInstance(vu modules.VU) modules.Instance {
+	return &Encoding{vu: vu}
+}
+
+// Exports returns the exports of the encoding module.
+func (e *Encoding) Exports() modules.Exports {
+	return modules.Exports{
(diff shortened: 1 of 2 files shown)


MStoykov (approved):
LGTM, I still (as in the other PRs) am not certain if we:
-  shouldn't stop exporting stuff that no longer need to be exported 
- just return `map[string]interface{}` instead of creating a goja.Object that is basically that.

yorugac (dismissed):

yorugac:
Not sure about exports but regarding this:

> - just return map[string]interface{} instead of creating a goja.Object that is basically that.

I think it's a matter of readability too: `goja.Object` is a "clear" entity on its own, but `map[string]interface{}` could be anything.

oleiade:
I agree with @yorugac here. I also think explicit is better than implicit on the front of types.

oleiade (dismissed):
LGTM :+1:

codebien:
>I think it's a matter of readability too: goja.Object is a "clear" entity on its own, but map[string]interface{} could be anything.

This is true for the specific line, but I think that the map exported directly from the `Exports` method adds some important benefits, check the latest fixup:
* We have removed the `obj` field from the instance, so the instance is the object and not a wrapper of another object. It should be clearer in terms of architecture.
* Export method from `Exports` should be better in terms of readability instead of Export them from the `NewInstanceModule`.

Not visible from the fixup but from other PRs:
* It's not required to add the `mustExport` method

MStoykov:
> I think it's a matter of readability too: goja.Object is a "clear" entity on its own, but map[string]interface{} could be anything. 

goja.Object can be *anything* ;) `map[string]interface{}` can only be a map ;). 

So the argument for me goes the other direction in two different ways:
1. `goja.Object` can truly be anything you can make it act as w/e you want which lets you w/e you want which also means that whoever reads this code now needs to actually look at whether you do something funky - which is what i went to look for, like I expected that it was `goja.Object` so that someone can't change an import, but that wasn't it we just made a goja.Object and then used it for the same thing that map is used for
2. the way the `map[string]interface{}` can be written directly in the Export makes it IMO a lot easier to understand what it's doing and in case like this where there is nothing special it's the better option IMO

Around #2258 we might decided to make this all more ... robust and with more specific types, maybe having something like nodejs' [module.syncBuiltinESMExports()](https://nodejs.org/api/module.html#modulesyncbuiltinesmexports), but given that this PR just moves from the old `common.Bind` to the new `modules.Module` based approach and in the old one this *is* `map[string]interface{}` if anything it's probably better to not more changes than necessary

yorugac (dismissed):

yorugac:
For the sake of this discussion I just wanted to note of possibility of future renamings of those types but it seems @MStoykov already did that :slightly_smiling_face: I.e. there're pros and cons as mentioned and I also think this question should be revisited separately, like in #2258 as it simply has larger scope than this PR.

MStoykov (approved):

yorugac (approved):

PR #2242, opened 2021-11-16
js/k6: ModuleV2 migration

codebien:
Migrated `k6` to the `modules.Module` (aka `modules.ModuleV2`) interface.

### DIFF
--- a/js/modules/k6/k6.go
+++ b/js/modules/k6/k6.go
@@ -22,7 +22,6 @@
 package k6
 
 import (
-	"context"
 	"errors"
 	"math/rand"
 	"sync/atomic"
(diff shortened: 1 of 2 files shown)


js/modules/k6/k6.go
```diff
@@ -31,22 +30,64 @@ import (
...
-var ErrCheckInInitContext = common.NewInitContextError("Using check() in the init context is not supported")
+	return mi
+}
 
-// New returns a new module Struct.
-func New() *K6 {
-	return &K6{}
+// Exports returns the exports of the k6 module.
+func (mi *K6) Exports() modules.Exports {
+	return modules.Exports{Default: mi.obj}
```
MStoykov, line 90: I would argue this should set only `Named` and then you can also drop the `obj` and have the map inlined here similar to 
https://github.com/grafana/k6/blob/3a604bf408ec278f71da2f577c76edc7778aece2/js/modules/k6/metrics/metrics.go#L162-L171

After all I would expect that users will use that as `import { sleep } from "k6";sleep()` not as `import k6 from "k6"; k6.sleep()`

js/modules/k6/k6_test.go
```diff
@@ -182,7 +225,16 @@ func checkTestRuntime(t testing.TB, ctxs ...*context.Context) (
 	ctx = common.WithRuntime(ctx, rt)
 	ctx = lib.WithState(ctx, state)
 	state.BuiltinMetrics = metrics.RegisterBuiltinMetrics(metrics.NewRegistry())
-	require.NoError(t, rt.Set("k6", common.Bind(rt, New(), &ctx)))
+	m, ok := New().NewModuleInstance(
+		&modulestest.VU{
+			RuntimeField: rt,
+			InitEnvField: &common.InitEnvironment{},
+			CtxField:     common.WithRuntime(ctx, rt),
```
yorugac, line 232: Here `common.WithRuntime` is called twice: at L225 and at L232. Shouldn't it be twice?

yorugac (changes requested):

MStoykov (approved):

PR #2243, opened 2021-11-16
js/data: ModuleV2 migration

codebien:
Migrated `k6/data` to the `modules.Module` (aka `modules.ModuleV2`) interface.

### DIFF
--- a/js/modules/k6/data/data.go
+++ b/js/modules/k6/data/data.go
@@ -21,23 +21,87 @@
 package data
 
 import (
-	"context"
 	"errors"
 	"strconv"
 	"sync"
 
 	"github.com/dop251/goja"
 	"go.k6.io/k6/js/common"
-	"go.k6.io/k6/lib"
+	"go.k6.io/k6/js/modules"
 )
 
-type data struct {
-	shared sharedArrays
+type (
+	// RootModule is the global module instance that will create module
+	// instances for each VU.
+	RootModule struct {
+		shared sharedArrays
+	}
+
+	// Data represents an instance of the data module.
+	Data struct {
+		vu     modules.VU
+		shared *sharedArrays
+	}
+
+	sharedArrays struct {
+		data map[string]sharedArray
+		mu   sync.RWMutex
+	}
+)
+
+var (
+	_ modules.Module   = &RootModule{}
+	_ modules.Instance = &Data{}
+)
+
+// New returns a pointer to a new RootModule instance.
+func New() *RootModule {
+	return &RootModule{
+		shared: sharedArrays{
+			data: make(map[string]sharedArray),
+		},
+	}
 }
 
-type sharedArrays struct {
-	data map[string]sharedArray
-	mu   sync.RWMutex
+// NewModuleInstance implements the modules.Module interface to return
+// a new instance for each VU.
+func (rm *RootModule) NewModuleInstance(vu modules.VU) modules.Instance {
+	return &Data{
+		vu:     vu,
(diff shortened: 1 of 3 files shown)


js/modules/k6/data/data.go
```diff
@@ -21,23 +21,87 @@
...
+	return modules.Exports{
+		Named: map[string]interface{}{
+			"SharedArray": d.XSharedArray,
+		},
+	}
+}
+
+// XSharedArray is a constructor returning a shareable read-only array
+// indentified by the name and having their contents be whatever the call returns
+func (d *Data) XSharedArray(call goja.ConstructorCall) *goja.Object {
```
MStoykov, line 86: this nolonger needs the X

MStoykov (approved):
Minor nitpick

js/modules/k6/data/share_test.go
```diff
@@ -43,22 +45,33 @@ var array = new data.SharedArray("shared",function() {
...
+		&modulestest.VU{
+			RuntimeField: rt,
+			InitEnvField: &common.InitEnvironment{},
+			CtxField:     common.WithRuntime(context.Background(), rt),
+			StateField:   nil,
+		},
+	).(*Data)
+	if !ok {
+		return rt, fmt.Errorf("not a Data module instance")
+	}
```
yorugac, line 62: Perhaps, not even a request change, but a curiosity question: why didn't you use `require.True(t, ok)` here?

yorugac (commented):
All seems fine :slightly_smiling_face: There is just 1 "nitpick" question.

js/modules/k6/data/share_test.go
codebien: assert|require with concurrency could generate a race so I tend to avoid it for helper functions if it's possible (if the func return error). Otherwise, I would prefer a classic `require.True` as you suggested.

js/modules/k6/data/data.go
```diff
@@ -21,23 +21,87 @@
...
+// Exports returns the exports of the data module.
+func (d *Data) Exports() modules.Exports {
+	return modules.Exports{
+		Named: map[string]interface{}{
+			"SharedArray": d.sharedArray,
+		},
+	}
+}
+
+// XSharedArray is a constructor returning a shareable read-only array
```
codebien, line 84:
```suggestion
// sharedArray is a constructor returning a shareable read-only array
```

js/modules/k6/data/share_test.go
MStoykov: I think what you are talking about @codebien is that `t.FailNow` and co. are [supposed](https://pkg.go.dev/testing#B.FailNow) to be called *only* from the goroutine that is running the test, not by anyone that is spawned by it. 

We do ... not do that quite a lot it just so happens that in most cases the failure never happens :smile:  
I think it's perfectly fine to leave the `error` instead of using `require.True` and co., but also in this case I would expect it's perfectly fine to use it as well

codebien: >that in most cases the failure never happens
Sure, but in case of changes for refactoring, you could get the race.

>but also in this case I would expect it's perfectly fine to use it as well

Yep, but I don't know how the helper could be used in the next test we will write. For this reason, I think it makes sense just for helpers shared across multiple tests. Ofc, in a single test I would prefer to use directly the assert function.

MStoykov (approved):

yorugac (approved):

PR #2245, opened 2021-11-18
Support websocket-connect error handler in user scripts

CCludts:
Hi there!

I'm writing a websocket loadtest with k6. The websocket server returns 503 errors when it's overloaded or when transient network errors occur. This is intended and fine, our production client retries these errors.

However I couldn't find a way to distinguish 3xx, 4xx, 5xx or problems with the websocket handshake. All of these scenarios result in 'websocket: bad handshake' error, which is returned by Gorilla-websocket.

Gorilla however does return more information. This pull request adds an extra (optional) argument to ws.connect, to handle websocket-connect errors.
In practice this can be used to report the different error codes, and also allow retrying of transient errors.

here's my example code:
(it's typescript, but the compiler emits simple javascript)

```
interface WSHttpResponse {
  url: string,
  status: number,
  headers: { [key: string]: string[] },
  body: ArrayBuffer,
  error: string,
}

ws.connect(wsEndpoint, params, sock => {
  // connect worked, setup socket as usual
  },
  //@ts-ignore  ws-connect error handler. @types/k6 package still need to be updated
  (err: WSHttpResponse) => {
    console.log("ERROR: " + JSON.stringify(err));
  }
);
```

example output:

```
ERROR {"url":"ws://localhost:10555","status":503,"headers":{"Date":"Thu, 18 Nov 2021 17:36:39 GMT","Server":"envoy"},"body":"something went wrong","error":""}
```

update: I simplified the code by using the existing WSHttpResponse instead of a new type

### DIFF
--- a/js/modules/k6/ws/ws.go
+++ b/js/modules/k6/ws/ws.go
@@ -77,13 +77,6 @@ type message struct {
 	data  []byte
 }
 
-type wsErr struct {
-	Error   string
-	Headers map[string][]string
-	Code    int
-	Body    []byte
-}
-
 const writeWait = 10 * time.Second
 
 func New() *WS {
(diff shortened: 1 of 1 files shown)


MStoykov:
Hi @CCludts , thanks for this PR. 

Unfortunately, I think that this is better handled by just actually not throwing unless `throw` is true. As explained in https://github.com/grafana/k6/issues/2247 this is  bug and if we fix that you will not need to provide a callback ;)

Issue #1443, opened 2020-05-11
Remove the JS runtime from threshold calculations


sniku:
This issue aims to bring 80% of the benefits of #1441 while doing only 20% of the work. 

The ideal future state of thresholds is defined in #1441 . 
This issue addresses only the most important problem of thresholds - removing the js runtime. This issue doesn't introduce the new concepts from the original specification. 

## Motivation

This issue addresses the following issues (two first issues from the original spec).
1. Thresholds in k6 are executed in JavaScript runtime while thresholds in k6 cloud are parsed and executed in Python. This doesn't work for non-trivial cases.
2. Thresholds can't be validated.

## Quick overview of the new thresholds

```javascript
import { http_req_duration } from "k6/metrics/http";
import { Counter } from "k6/metrics";

let successfulLogins = new Counter("successful_logins");

export let options = {
  thresholds: {  // thresholds are still defined in an object {}
    'http_req_duration{staticAssets:true}': [
      {threshold: "mean < 500"}  // the threshold string is not JS. 
    ],
    'successful_logins': ["count > 1"]
  }
}
```

Yes, it's the old format, but the threshold expression is no longer executed in a JS runtime. Instead, it's parsed and executed in go.

## Parsing
(copied from the original spec).

To provide consistency between local and cloud execution, threshold parsing should happen only in k6.
When the test is executed in the cloud, k6 should parse and provide the agreed-upon format to the k6-cloud backend.

The proposed format is
```javascript
let thresholds = [{
  metric_name: "http_req_duration|others",
  metric_type: "trend|gauge|counter|rate",
  metric_value_contains: "time|default",  // isTime=true
  metric_initial_value: 0, // 
  filters: [
    {"left-hand": "status", "operator": ">", "right-hand": "200", "right-hand-type": "int"}, // not supported by this proposal.
    {"left-hand": "staticAssets", "operator": "==", "right-hand": "yes", "right-hand-type": "numeric"},
  ],
  conditions: [
    {"left-hand": "mean", "operator": "<", "right-hand": "200", "right-hand-type": "string"},
  ],
  name: "http_req_duration(status>200, staticAssets=yes)", // auto-generate when not provided. 
}];
```

## Validation 

Thresholds should be validated and abort the execution in case of validation errors.

### `metric_name` validation. 
The metric name must be one of:
1. Built-in k6 metric (`http_req_duration`, etc)
2. user-defined metric (`successful_logins`, etc).

[...]

Issue #1441, opened 2020-05-08
WIP: New thresholds


sniku:
Note: This issue is WIP. Please don't comment until the WIP flag is removed.

# New Threshold specification

# Motivation

There are several problems with the current Threshold implementation:

1. Thresholds in k6 are executed in JavaScript runtime while thresholds in `k6 cloud` are parsed and executed in Python. This doesn't work for non-trivial cases.
2. Thresholds can't be validated. 
3. Threshold DSL format is limited in filtering. Example `http_req_duration{status:200,status:201}`
4. The UX of specifying non-trivial Thresholds is poor. Example `group_duration{group:::publicCrocodiles}`. 

# Quick overview of the new format

```javascript
import { Threshold } from "k6";
import { http_req_duration } from "k6/metrics/http";
import { Counter } from "k6/metrics";

let successfulLogins = new Counter("successful_logins");

export let options = {
  thresholds: [
    new Threshold(http_req_duration, {
      filter: [
        'staticAssets == "true"',
        // "status >= 200", // future extension. Not supported by this proposal.
      ],
      conditions: ["mean < 500", "p(99) < 300"],
      abortOnFail: false,
      name: "My SLA for the CDN"
    }),
    new Threshold(successfulLogins, {
      conditions: ["count > 1"],
    }),

  ]
}
```

# Improvements

The new format will bring the following benefits:
1. Consistency between local execution and cloud execution.
2. Ability to validate thresholds.
3. Developer-experience - the new format is easier to write and understand.
4. Ability to specify more than one threshold on the same metric (http_req_duration)
5. Ability to filter metric data in a more flexible way (create sub-metrics).
6. Performance. k6 won't need to calculate values for `aggregation methods` that are not used. (no need to calculate `avg` when threshold only uses `min`) 

### Drawbacks of the new format

1. Thresholds can no longer be a free-form javascript code.
2. ?

### backwards compatibility
To provide backwards compatibility, the new k6 thresholds are defined in an array. The old thresholds are using an object {}.

Old: 
```javascript
export let options = {
  thresholds: {} // object
};
```

New
```javascript
export let options = {
  thresholds: [] // array
};
```

#### Parsing
To provide consistency between local and cloud execution, threshold parsing should happen only in k6.
When the test is executed in the cloud, k6 should parse and provide the agreed-upon format to the k6-cloud backend.

The proposed format is
```javascript
let thresholds = [{
  metric_name: "http_req_duration|others",
[...]

Issue #1346, opened 2020-03-06
Metrics for which no data was recorded are not displayed or evaluated for Thresholds


sniku:
When a custom metric is defined, but no data for that metric is recorded, it's not displayed in the results, and thresholds are not properly evaluated.

 It's easiest to show it on an example.

Consider this script.

```JavaScript
import { Counter } from 'k6/metrics';
import { sleep } from 'k6';
import http from 'k6/http';

let allErrors = new Counter('error_counter');

export let options = {
  thresholds: {
    error_counter: [
      'count < 10', // 10 of fewer errors are tolerated
    ]
  }
}

export default function () {
  let auth_resp = http.post('https://test-api.k6.io/auth/token/login/', {username: 'test-user', 'password': 'supersecure'});

  if (auth_resp.status >= 400){
    allErrors.add(1, { errorType: 'authError' }); // tagged value creates submetric (useful for making thresholds specific)
  }

  sleep(1);
}
```

If the test run is successful, the `allErrors` will contain 0 errors. In that case metric as well as the threshold indicator won't be present in the output. 

![image](https://user-images.githubusercontent.com/442646/76085707-dc1e1d80-5fb2-11ea-8652-18f03873b77e.png)

I expected both to be present, just like in this case, when 1 error was recorded:

![image](https://user-images.githubusercontent.com/442646/76085845-1982ab00-5fb3-11ea-8606-1e5eeac2e23a.png)

The most problematic manifestation of this bug is when threshold should fail, but doesn't because it's not evaluated. 

Example:
```JavaScript
import { Counter } from 'k6/metrics';
import { sleep } from 'k6';
import http from 'k6/http';

let allErrors = new Counter('error_counter');

export let options = {
  thresholds: {
    error_counter: [
      'count < 10' // This threshold should fail, but it doesn't
    ]
  }
}

export default function () {
  sleep(1);
}
```

na--:
A duplicate of https://github.com/loadimpact/k6/issues/1053, but I'll leave this open, since it's better.

na--:
FWIW, there's an easy workaround for this issue, when it comes to `Counter` metrics at least: https://community.k6.io/t/looking-for-max-execution-time-option/613/3?u=ned

na--:
https://github.com/loadimpact/k6/issues/1832 (or something like it) is a prerequisite for the current issue.

Issue #1253, opened 2019-11-25
summaryTrendStats option doesn't affect the k6 REST API results


na--:
Running k6 with custom trend columns like this: `k6 run --linger --summary-trend-stats="min,med,avg,count,max,p(85),p(95),p(99),p(99.9),p(99.999)" github.com/loadimpact/k6/samples/thresholds_readme_example.js`

And then calling the REST API `/v1/metrics` method: `curl -s 'http://localhost:6565/v1/metrics' | jq '.data[] | select(.id == "group_duration")'`

I'd expect to see the percentiles I configured in `summaryTrendStats`, but instead I see something like this:
```json
{
  "type": "metrics",
  "id": "group_duration",
  "attributes": {
    "type": "trend",
    "contains": "time",
    "tainted": null,
    "sample": {
      "avg": 128.2877491095891,
      "max": 146.327711,
      "med": 127.666555,
      "min": 123.76435,
      "p(90)": 131.6244158,
      "p(95)": 133.2135961
    }
  }
}

```

PR #2251, opened 2021-11-19
Fix/1443 remove the JS runtime from threshold calculations

oleiade:
This Pull Request addresses the minimum scope described in #1443 and replaces the current thresholds' expression evaluation implementation. It drops the reliance on the [Goja JS runtime](https://pkg.go.dev/github.com/dop251/goja#Runtime), in favor of a pure-go based implementation. With this PR, thresholds are parsed, and evaluated directly in Go, and all JS runtime-related threshold code has been dropped. It sticks to the existing format, and does not introduce any of the design and format changes proposed in #1443 and #1441.

Design notes regarding the PR's content:
  - The introduction of a minimalistic parser combinators library, taken and adapted from [Jeffhail benthos' bloblang implementation](https://github.com/Jeffail/benthos/). The design choice of combinators was motivated by their ease of use, of extension, and general maintainability and performance. Looking at #1443 and #1441 it also appeared like the best choice when it comes to further improvements we might want to make. The choice to extract the code from Bloblang was motivated by the fact that this was definitely the cleanest and idiomatic implementation I could find; but it was unfortunately (to my knowledge) not offered as a dedicated  library module. I also brought a few improvements in the process, by modifying and extending the existing API to fit K6's use-case.
  - I put the library's code in `pkg`. I was not entirely sure what the best place would be for code that's generic enough to be a good independent library, but not big enough to make it into a dedicated repo. `pkg` is often used as the place where projects put their public reusable code. I'm not hung onto it at all. Please propose an alternative if you prefer to put it in another place 👀 
  - I added a sink map to the `Thresholds` struct, and pass it around to children `Threshold` instances when they need to be evaluated. It's the simplest thing I could think of, but with my limited understanding of the current design of K6, I might have missed another way of doing it.

Let me know what you think 👀 🐻

### DIFF
--- a/stats/thresholds.go
+++ b/stats/thresholds.go
@@ -17,54 +17,35 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  *
  */
-
 package stats
 
 import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"strings"
 	"time"
-
-	"github.com/dop251/goja"
 
 	"go.k6.io/k6/lib/types"
 )
-
-const jsEnvSrc = `
-function p(pct) {
-	return __sink__.P(pct/100.0);
-};
-`
-
-var jsEnv *goja.Program
-
-func init() {
-	pgm, err := goja.Compile("__env__", jsEnvSrc, true)
-	if err != nil {
-		panic(err)
-	}
-	jsEnv = pgm
-}
 
 // Threshold is a representation of a single threshold for a single metric
 type Threshold struct {
 	// Source is the text based source of the threshold
 	Source string
-	// LastFailed is a makrer if the last testing of this threshold failed
+	// LastFailed is a marker if the last testing of this threshold failed
 	LastFailed bool
 	// AbortOnFail marks if a given threshold fails that the whole test should be aborted
 	AbortOnFail bool
 	// AbortGracePeriod is a the minimum amount of time a test should be running before a failing
 	// this threshold will abort the test
 	AbortGracePeriod types.NullDuration
-
-	pgm *goja.Program
-	rt  *goja.Runtime
+	// parsed is the threshold expression parsed from the Source
+	parsed *thresholdExpression
 }
 
-func newThreshold(src string, newThreshold *goja.Runtime, abortOnFail bool, gracePeriod types.NullDuration) (*Threshold, error) {
-	pgm, err := goja.Compile("__threshold__", src, true)
+func newThreshold(src string, abortOnFail bool, gracePeriod types.NullDuration) (*Threshold, error) {
+	parsedExpression, err := parseThresholdExpression(src)
 	if err != nil {
 		return nil, err
 	}
@@ -73,23 +54,57 @@ func newThreshold(src string, newThreshold *goja.Runtime, abortOnFail bool, grac
 		Source:           src,
 		AbortOnFail:      abortOnFail,
 		AbortGracePeriod: gracePeriod,
-		pgm:              pgm,
-		rt:               newThreshold,
+		parsed:           parsedExpression,
 	}, nil
 }
 
-func (t Threshold) runNoTaint() (bool, error) {
-	v, err := t.rt.RunProgram(t.pgm)
-	if err != nil {
-		return false, err
+func (t *Threshold) runNoTaint(sinks map[string]float64) (bool, error) {
+	// Extract the sink value for the aggregation method used in the threshold
+	// expression
+	lhs, ok := sinks[t.parsed.AggregationMethod]
+	if !ok {
+		return false, fmt.Errorf("unable to apply threshold %s over metrics; reason: "+
+			"no metric supporting the %s aggregation method found",
+			t.Source,
+			t.parsed.AggregationMethod)
 	}
-	return v.ToBoolean(), nil
+
+	// Apply the threshold expression operator to the left and
+	// right hand side values
+	var passes bool
+	switch t.parsed.Operator {
+	case ">":
+		passes = lhs > t.parsed.Value
+	case ">=":
+		passes = lhs >= t.parsed.Value
+	case "<=":
+		passes = lhs <= t.parsed.Value
+	case "<":
+		passes = lhs < t.parsed.Value
+	case "==", "===":
+		// Considering a sink always maps to float64 values,
+		// strictly equal is equivalent to loosely equal
+		passes = lhs == t.parsed.Value
+	case "!=":
+		passes = lhs != t.parsed.Value
+	default:
+		// The parseThresholdExpression function should ensure that no invalid
+		// operator gets through, but let's protect our future selves anyhow.
+		return false, fmt.Errorf("unable to apply threshold %s over metrics; "+
+			"reason: %s is an invalid operator",
+			t.Source,
+			t.parsed.Operator,
+		)
+	}
+
+	// Perform the actual threshold verification
+	return passes, nil
 }
 
-func (t *Threshold) run() (bool, error) {
-	b, err := t.runNoTaint()
-	t.LastFailed = !b
-	return b, err
+func (t *Threshold) run(sinks map[string]float64) (bool, error) {
+	passes, err := t.runNoTaint(sinks)
+	t.LastFailed = !passes
+	return passes, err
 }
 
 type thresholdConfig struct {
@@ -98,11 +113,11 @@ type thresholdConfig struct {
 	AbortGracePeriod types.NullDuration `json:"delayAbortEval"`
 }
 
-//used internally for JSON marshalling
+// used internally for JSON marshalling
 type rawThresholdConfig thresholdConfig
 
 func (tc *thresholdConfig) UnmarshalJSON(data []byte) error {
-	//shortcircuit unmarshalling for simple string format
+	// shortcircuit unmarshalling for simple string format
 	if err := json.Unmarshal(data, &tc.Threshold); err == nil {
 		return nil
 	}
@@ -122,9 +137,9 @@ func (tc thresholdConfig) MarshalJSON() ([]byte, error) {
 
 // Thresholds is the combination of all Thresholds for a given metric
 type Thresholds struct {
-	Runtime    *goja.Runtime
 	Thresholds []*Threshold
 	Abort      bool
+	sinked     map[string]float64
 }
 
 // NewThresholds returns Thresholds objects representing the provided source strings
@@ -138,60 +153,88 @@ func NewThresholds(sources []string) (Thresholds, error) {
 }
 
 func newThresholdsWithConfig(configs []thresholdConfig) (Thresholds, error) {
-	rt := goja.New()
-	if _, err := rt.RunProgram(jsEnv); err != nil {
-		return Thresholds{}, fmt.Errorf("threshold builtin error: %w", err)
-	}
+	thresholds := make([]*Threshold, len(configs))
+	sinked := make(map[string]float64)
 
-	ts := make([]*Threshold, len(configs))
 	for i, config := range configs {
-		t, err := newThreshold(config.Threshold, rt, config.AbortOnFail, config.AbortGracePeriod)
+		t, err := newThreshold(config.Threshold, config.AbortOnFail, config.AbortGracePeriod)
 		if err != nil {
 			return Thresholds{}, fmt.Errorf("threshold %d error: %w", i, err)
 		}
-		ts[i] = t
+		thresholds[i] = t
 	}
 
-	return Thresholds{rt, ts, false}, nil
+	return Thresholds{thresholds, false, sinked}, nil
 }
 
-func (ts *Thresholds) updateVM(sink Sink, t time.Duration) error {
-	ts.Runtime.Set("__sink__", sink)
-	f := sink.Format(t)
-	for k, v := range f {
-		ts.Runtime.Set(k, v)
-	}
-	return nil
-}
-
-func (ts *Thresholds) runAll(t time.Duration) (bool, error) {
-	succ := true
-	for i, th := range ts.Thresholds {
-		b, err := th.run()
+func (ts *Thresholds) runAll(timeSpentInTest time.Duration) (bool, error) {
+	succeeded := true
+	for i, threshold := range ts.Thresholds {
+		b, err := threshold.run(ts.sinked)
 		if err != nil {
 			return false, fmt.Errorf("threshold %d run error: %w", i, err)
 		}
-		if !b {
-			succ = false
 
-			if ts.Abort || !th.AbortOnFail {
+		if !b {
+			succeeded = false
+
+			if ts.Abort || !threshold.AbortOnFail {
 				continue
 			}
 
-			ts.Abort = !th.AbortGracePeriod.Valid ||
-				th.AbortGracePeriod.Duration < types.Duration(t)
+			ts.Abort = !threshold.AbortGracePeriod.Valid ||
+				threshold.AbortGracePeriod.Duration < types.Duration(timeSpentInTest)
 		}
 	}
-	return succ, nil
+
+	return succeeded, nil
 }
 
 // Run processes all the thresholds with the provided Sink at the provided time and returns if any
 // of them fails
-func (ts *Thresholds) Run(sink Sink, t time.Duration) (bool, error) {
-	if err := ts.updateVM(sink, t); err != nil {
-		return false, err
+func (ts *Thresholds) Run(sink Sink, duration time.Duration) (bool, error) {
+	// Initialize the sinks store
+	ts.sinked = make(map[string]float64)
+
+	// FIXME: Remove this comment as soon as the stats.Sink does not expose Format anymore.
+	//
+	// As of December 2021, this block reproduces the behavior of the
+	// stats.Sink.Format behavior. As we intend to try to get away from it,
+	// we instead implement the behavior directly here.
+	//
+	// For more details, see https://github.com/grafana/k6/issues/2320
+	switch sinkImpl := sink.(type) {
+	case *CounterSink:
+		ts.sinked["count"] = sinkImpl.Value
+		ts.sinked["rate"] = sinkImpl.Value / (float64(duration) / float64(time.Second))
+	case *GaugeSink:
+		ts.sinked["value"] = sinkImpl.Value
+	case *TrendSink:
+		ts.sinked["min"] = sinkImpl.Min
+		ts.sinked["max"] = sinkImpl.Max
+		ts.sinked["avg"] = sinkImpl.Avg
+		ts.sinked["med"] = sinkImpl.Med
+
+		// Parse the percentile thresholds and insert them in
+		// the sinks mapping.
+		for _, threshold := range ts.Thresholds {
+			if !strings.HasPrefix(threshold.parsed.AggregationMethod, "p(") {
+				continue
+			}
+
+			ts.sinked[threshold.parsed.AggregationMethod] = sinkImpl.P(threshold.parsed.AggregationValue.Float64 / 100)
+		}
+	case *RateSink:
+		ts.sinked["rate"] = float64(sinkImpl.Trues) / float64(sinkImpl.Total)
+	case DummySink:
+		for k, v := range sinkImpl {
+			ts.sinked[k] = v
+		}
+	default:
+		return false, fmt.Errorf("unable to run Thresholds; reason: unknown sink type")
 	}
-	return ts.runAll(t)
+
+	return ts.runAll(duration)
 }
 
 // UnmarshalJSON is implementation of json.Unmarshaler
--- a/stats/thresholds_parser.go
+++ b/stats/thresholds_parser.go
@@ -1,0 +1,208 @@
+/*
+ *
+ * k6 - a next-generation load testing tool
+ * Copyright (C) 2021 Load Impact
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+package stats
+
+import (
+	"fmt"
+	"strconv"
+	"strings"
+
+	"gopkg.in/guregu/null.v3"
+)
+
+// thresholdExpression holds the parsed result of a threshold expression,
+// as described in: https://k6.io/docs/using-k6/thresholds/#threshold-syntax
+type thresholdExpression struct {
+	// AggregationMethod holds the aggregation method parsed
+	// from the threshold expression. Possible values are described
+	// by `aggregationMethodTokens`.
+	AggregationMethod string
+
+	// AggregationValue will hold the aggregation method's pivot value
+	// in the event it is a percentile. For instance: an expression of the form p(99.9) < 200,
+	// would result in AggregationValue to be set to 99.9.
+	AggregationValue null.Float
+
+	// Operator holds the operator parsed from the threshold expression.
+	// Possible values are described by `operatorTokens`.
+	Operator string
+
+	// Value holds the value parsed from the threshold expression.
+	Value float64
+}
+
+// parseThresholdAssertion parses a threshold condition expression,
+// as defined in a JS script (for instance p(95)<1000), into a thresholdExpression
+// instance.
+//
+// It is expected to be of the form: `aggregation_method operator value`.
+// As defined by the following BNF:
+// ```
+// assertion           -> aggregation_method whitespace* operator whitespace* float
+// aggregation_method  -> trend | rate | gauge | counter
+// counter             -> "count" | "rate"
+// gauge               -> "value"
+// rate                -> "rate"
+// trend               -> "avg" | "min" | "max" | "med" | percentile
+// percentile          -> "p(" float ")"
+// operator            -> ">" | ">=" | "<=" | "<" | "==" | "===" | "!="
+// float               -> digit+ ("." digit+)?
+// digit               -> "0" | "1" | "2" | "3" | "4" | "5" | "6" | "7" | "8" | "9"
+// whitespace          -> " "
+// ```
+func parseThresholdExpression(input string) (*thresholdExpression, error) {
+	// Scanning makes no assumption on the underlying values, and only
+	// checks that the expression has the right format.
+	method, operator, value, err := scanThresholdExpression(input)
+	if err != nil {
+		return nil, fmt.Errorf("failed parsing threshold expression; reason: %w", err)
+	}
+
+	parsedMethod, parsedMethodValue, err := parseThresholdAggregationMethod(method)
+	if err != nil {
+		return nil, fmt.Errorf("failed parsing threshold expression's left hand side; reason: %w", err)
+	}
+
+	parsedValue, err := strconv.ParseFloat(value, 64)
+	if err != nil {
+		return nil, fmt.Errorf("failed parsing threshold expresion's right hand side; reason: %w", err)
+	}
+
+	condition := &thresholdExpression{
+		AggregationMethod: parsedMethod,
+		AggregationValue:  parsedMethodValue,
+		Operator:          operator,
+		Value:             parsedValue,
+	}
+
+	return condition, nil
+}
+
+// Define accepted threshold expression operators tokens
+const (
+	tokenLessEqual     = "<="
+	tokenLess          = "<"
+	tokenGreaterEqual  = ">="
+	tokenGreater       = ">"
+	tokenStrictlyEqual = "==="
+	tokenLooselyEqual  = "=="
+	tokenBangEqual     = "!="
+)
+
+// operatorTokens defines the list of operator-related tokens
+// used in threshold expressions parsing.
+//
+// It is meant to be used during the scan of threshold expressions.
+// Although declared as a `var`, being an array, it is effectively
+// immutable and can be considered constant.
+//
+// Note that because scanning uses a substring parser, and will match
+// the smallest common substring, the actual slice order matters.
+// Longer tokens with symbols in common with shorter ones must appear
+// first in the list in order to be effectively matched.
+var operatorTokens = [7]string{ // nolint:gochecknoglobals
+	tokenLessEqual,
+	tokenLess,
+	tokenGreaterEqual,
+	tokenGreater,
+	tokenStrictlyEqual,
+	tokenLooselyEqual,
+	tokenBangEqual,
+}
+
+// scanThresholdExpression scans a threshold condition expression of the
+// form: `aggregation_method operator value`. An invalid or unknown operator
+// will produce an error. However, no assertions regarding
+// either the left-hand side aggregation method nor the right-hand
+// side value will be made: they will be returned as is, only trimmed from
+// their spaces.
+func scanThresholdExpression(input string) (string, string, string, error) {
+	for _, op := range operatorTokens {
+		substrings := strings.SplitN(input, op, 2)
+		if len(substrings) == 2 {
+			return strings.TrimSpace(substrings[0]), op, strings.TrimSpace(substrings[1]), nil
+		}
+	}
+
+	return "", "", "", fmt.Errorf("malformed threshold expression")
+}
+
+// Define accepted threshold expression aggregation tokens
+// Percentile token `p(..)` is accepted too but handled separately.
+const (
+	tokenValue      = "value"
+	tokenCount      = "count"
+	tokenRate       = "rate"
+	tokenAvg        = "avg"
+	tokenMin        = "min"
+	tokenMed        = "med"
+	tokenMax        = "max"
+	tokenPercentile = "p"
+)
+
+// aggregationMethodTokens defines the list of aggregation method
+// used in the parsing of threshold expressions.
+//
+// It is meant to be used during the parsing of threshold expressions.
+// Although declared as a `var`, being an array, it is effectively
+// immutable and can be considered constant.
+var aggregationMethodTokens = [8]string{ // nolint:gochecknoglobals
+	tokenValue,
+	tokenCount,
+	tokenRate,
+	tokenAvg,
+	tokenMin,
+	tokenMed,
+	tokenMax,
+	tokenPercentile,
+}
+
+// parseThresholdMethod will parse a threshold condition expression's method.
+// It assumes the provided input argument is already trimmed and cleaned up.
+// If it encounters a percentile method, it will parse it and verify it
+// boils down to an expression of the form: `p(float64)`, but will return
+// it verbatim, as a string.
+func parseThresholdAggregationMethod(input string) (string, null.Float, error) {
+	// Is the input one of the methods keywords?
+	for _, m := range aggregationMethodTokens {
+		// Percentile expressions being of the form p(value),
+		// they won't be matched here.
+		if m == input {
+			return input, null.Float{}, nil
+		}
+	}
+
+	// Otherwise, attempt to parse a percentile expression
+	if strings.HasPrefix(input, tokenPercentile+"(") && strings.HasSuffix(input, ")") {
+		aggregationValue, err := strconv.ParseFloat(trimDelimited("p(", input, ")"), 64)
+		if err != nil {
+			return "", null.Float{}, fmt.Errorf("malformed percentile value; reason: %w", err)
+		}
+
+		return input, null.FloatFrom(aggregationValue), nil
+	}
+
+	return "", null.Float{}, fmt.Errorf("failed parsing method from expression")
+}
+
+func trimDelimited(prefix, input, suffix string) string {
+	return strings.TrimSuffix(strings.TrimPrefix(input, prefix), suffix)
+}
--- a/stats/thresholds_parser_test.go
+++ b/stats/thresholds_parser_test.go
@@ -1,0 +1,361 @@
+/*
+ *
+ * k6 - a next-generation load testing tool
+ * Copyright (C) 2021 Load Impact
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+package stats
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"gopkg.in/guregu/null.v3"
+)
+
+func TestParseThresholdExpression(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name           string
+		input          string
+		wantExpression *thresholdExpression
+		wantErr        bool
+	}{
+		{
+			name:           "unknown expression's operator fails",
+			input:          "count!20",
+			wantExpression: nil,
+			wantErr:        true,
+		},
+		{
+			name:           "unknown expression's method fails",
+			input:          "foo>20",
+			wantExpression: nil,
+			wantErr:        true,
+		},
+		{
+			name:           "non numerical expression's value fails",
+			input:          "count>abc",
+			wantExpression: nil,
+			wantErr:        true,
+		},
+		{
+			name:           "valid threshold expression syntax",
+			input:          "count>20",
+			wantExpression: &thresholdExpression{AggregationMethod: "count", Operator: ">", Value: 20},
+			wantErr:        false,
+		},
+	}
+	for _, testCase := range tests {
+		testCase := testCase
+
+		t.Run(testCase.name, func(t *testing.T) {
+			t.Parallel()
+
+			gotExpression, gotErr := parseThresholdExpression(testCase.input)
+
+			assert.Equal(t,
+				testCase.wantErr,
+				gotErr != nil,
+				"parseThresholdExpression() error = %v, wantErr %v", gotErr, testCase.wantErr,
+			)
+
+			assert.Equal(t,
+				testCase.wantExpression,
+				gotExpression,
+				"parseThresholdExpression() gotExpression = %v, want %v", gotExpression, testCase.wantExpression,
+			)
+		})
+	}
+}
+
+func BenchmarkParseThresholdExpression(b *testing.B) {
+	for i := 0; i < b.N; i++ {
+		parseThresholdExpression("count>20") // nolint
+	}
+}
+
+func TestParseThresholdAggregationMethod(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name            string
+		input           string
+		wantMethod      string
+		wantMethodValue null.Float
+		wantErr         bool
+	}{
+		{
+			name:            "count method is parsed",
+			input:           "count",
+			wantMethod:      "count",
+			wantMethodValue: null.Float{},
+			wantErr:         false,
+		},
+		{
+			name:            "rate method is parsed",
+			input:           "rate",
+			wantMethod:      "rate",
+			wantMethodValue: null.Float{},
+			wantErr:         false,
+		},
+		{
+			name:            "value method is parsed",
+			input:           "value",
+			wantMethod:      "value",
+			wantMethodValue: null.Float{},
+			wantErr:         false,
+		},
+		{
+			name:            "avg method is parsed",
+			input:           "avg",
+			wantMethod:      "avg",
+			wantMethodValue: null.Float{},
+			wantErr:         false,
+		},
+		{
+			name:            "min method is parsed",
+			input:           "min",
+			wantMethod:      "min",
+			wantMethodValue: null.Float{},
+			wantErr:         false,
+		},
+		{
+			name:            "max method is parsed",
+			input:           "max",
+			wantMethod:      "max",
+			wantMethodValue: null.Float{},
+			wantErr:         false,
+		},
+		{
+			name:            "med method is parsed",
+			input:           "med",
+			wantMethod:      "med",
+			wantMethodValue: null.Float{},
+			wantErr:         false,
+		},
+		{
+			name:            "percentile method with integer value is parsed",
+			input:           "p(99)",
+			wantMethod:      "p(99)",
+			wantMethodValue: null.FloatFrom(99),
+			wantErr:         false,
+		},
+		{
+			name:            "percentile method with floating point value is parsed",
+			input:           "p(99.9)",
+			wantMethod:      "p(99.9)",
+			wantMethodValue: null.FloatFrom(99.9),
+			wantErr:         false,
+		},
+		{
+			name:            "parsing invalid method fails",
+			input:           "foo",
+			wantMethod:      "",
+			wantMethodValue: null.Float{},
+			wantErr:         true,
+		},
+		{
+			name:            "parsing empty percentile expression fails",
+			input:           "p()",
+			wantMethod:      "",
+			wantMethodValue: null.Float{},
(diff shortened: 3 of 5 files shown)


oleiade:
I'd be really interested in your thoughts on how I approached testing @inancgumus; I'm all ears for any recommendations on that topic specifically 👂🏻 🙏🏻

MStoykov:
I haven't gone through the code, yet, but I would expect expressions like `1+1=2` and so on to possibly be valid. But also from UX perspective, it will be better if we require that something from the metric is used and if not to error that the threshold isn't based on the metric. 

Given that, for the purposes of the tests, you can rewrite (as you seem to have done) in a way that you get the test to pass -where it requires that the threshold fails just to write a threshold that will fail when evaluated but still based on the metric.

yorugac:
@oleiade I've only started digging into this, hopefully will finish tomorrow :sweat_smile: But I have a preliminary question: the work you reference in the PR description (Jeffhail benthos' bloblang) -- should it be referenced in the code as well? I imagine it depends on how much you adapted it...

Also, CI failure shows:
```
--- FAIL: TestClient (0.01s)
    --- FAIL: TestClient/ReflectUnregistered (0.03s)
        client_test.go:650: 
            	Error Trace:	client_test.go:650
            	            				client_test.go:687
            	Error:      	"can't list services: can't send request: EOF at go.k6.io/k6/js/common.Bind.func1 (native)" does not contain "rpc error: code = Unimplemented desc = unknown service grpc.reflection.v1alpha.ServerReflection"
            	Test:       	TestClient/ReflectUnregistered
FAIL
```
This doesn't look like a flaky one... Did it happen before? I can't repeat it on the current master :confused:

oleiade:
@yorugac Sorry for the huge chunk of code I've just dropped 🙏🏻 I prefer to make smaller iterative PRs in general, but I found it hard to avoid in this case. You're right about mentioning Jeff Hail's work somehow in the code. I wanted to find out what would be the best way to do that, and it somehow slipped out of my mind. I've adapted and extended the API, and modified some implementation here and there, but a big part of the code is still verbatim. I would like to add a copy of his copyright license at the top of the file (as is the MIT license etiquette, I think?); and mention that the base code comes from bloblang's implementation. Does that sound like a good approach.

Regarding the failing test, as I haven't touched the code since yesterday, and only rebased my PR on master, I can only assume that either the failing test comes from master, or that it impacted my code's behavior somehow. I'll look into it, thanks for pointing it out! 👍🏻

oleiade:
@yorugac I've modified the `combinators` package description to mention that most of the code comes from jeffhail's work. I also added their License to the top of the file, as the MIT requests. Do you find it sufficient?

yorugac (changes requested):
Thank you for the clarification on licenses, @oleiade ! It is now clear enough to me what happens with the code adaptation in this PR :+1: But I'm not sure about how it _should_ be done in k6 code base. I.e. should there be 2 full license headers as is currently done in `combinators.go`: it takes a lot of space and IIRC, there was some talk about dropping all of those headers? The only precedent in k6 that I know of ATM is the tc39 files and those are left without a header.
@na--, @MStoykov, could you please add your opinion on what should be done in this case with license / authorship?

On the review part. `combinators` pkg looks quite nice and `Result` is reminding me of some other languages :smile: I'm not sure about its name though: maybe rename it to just `parser`? I currently assume "combinators" came from the original code but it doesn't make much sense in k6 context IMHO. I don't have a strong opinion about `pkg` folder except there is only `lib` at the moment: are we moving `lib` -> `pkg`? It doesn't make sense to me to have 2 folders without understanding the difference between them.

Overall, I like how this change shapes up a lot! Theo, please see the comments: there are some suggestions about renamings and some requests about logic and clarifications. And one more thing I've been wondering about: should some benchmarking be done here?

stats/thresholds.go
```diff
@@ -122,9 +213,9 @@ func (tc thresholdConfig) MarshalJSON() ([]byte, error) {
 
 // Thresholds is the combination of all Thresholds for a given metric
 type Thresholds struct {
-	Runtime    *goja.Runtime
 	Thresholds []*Threshold
 	Abort      bool
+	Sinked     map[string]float64
```
yorugac, line 218: One more thing: since `Sinked` is not used anywhere else, I suggest making it `sinked`.

oleiade:
@yorugac Thanks for the review! I'll look into your comments and address them when I can. 

Regarding the naming, I don't think `parser` would make a lot of sense considering that parser combinators are actually a "generic" parsing technique/toolbox you build parsers on top of: as opposed to a recursive descent or LRLA parser for instance, which you tailor specifically to the format you're trying to parse. I also kept it separate in the spirit of a) being able to reuse it to extend the threshold's parsing in the future (to fit with the initial proposal this PR partly answers to) and b) making it possible for a potential third-party repo to use it directly from our repo by adding `github.com/grafana/k6/pkg/combinators` to their dependencies. Initially I considered naming it `crocombine` and putting in a separate repo...

Regarding benchmarking, I've thought about it, without never actually getting to implementing it. Now that you ask about it, I'll add some :)

stats/thresholds_parser.go
```diff
@@ -0,0 +1,230 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+package stats
+
+import (
+	"fmt"
+
+	c "go.k6.io/k6/pkg/combinators"
```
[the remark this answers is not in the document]
oleiade: I agree with you, this is a left over. Explicit is better than implicit 👍🏻

stats/thresholds_parser.go
```diff
@@ -0,0 +1,230 @@
...
+			return res
+		}
+
+		parsed, ok := res.Payload.([]interface{})
+		if !ok {
+			err := fmt.Errorf("failed parsing percentile expression; " +
+				"reason: converting ParsePercentile() parser result's payload to []interface{} failed",
+			)
+			res.Err = c.NewFatalError(input, err, "percentile")
+		}
```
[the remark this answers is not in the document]
oleiade: That's an excellent idea, I'll do that 💯

stats/thresholds_parser.go
```diff
@@ -0,0 +1,230 @@
...
+
+		percentile := fmt.Sprintf("%s%g%s", parsed[0].(string), parsed[1].(float64), parsed[2].(string))
```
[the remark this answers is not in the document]
oleiade: That's a very good point, I do not check that the percentile is with bounds `[0, 100]` at the moment. Going to add that + a test for it.

On the front of precision, the `%g` formatter in the `Sprintf` call ensures to limit the decimal precision to the minimum necessary. For instance, `99.0` would be represented as `99`, but `99.1` would remain `99.1` while `99.10` would be truncated to `99.1`.

stats/thresholds.go
```diff
@@ -26,70 +25,162 @@ import (
...
+	// The Sequence combinator will return a slice of interface{}
+	// instances. Up to us to decide what we want to cast them down
+	// to.
+	// Considering our expression format, the parser should return a slice
+	// of size 3 to us: aggregation_method operator sink_value. The type system
+	// ensures us it should be the case too, but let's protect our future selves anyhow.
+	var ok bool
+	parsed, ok := result.Payload.([]interface{})
+	if !ok || len(parsed) != 3 {
+		return nil, fmt.Errorf("parsing threshold condition %s failed; reason: malformed expression", expression)
```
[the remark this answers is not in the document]
oleiade: You're right! I'll make it more explicit 👍🏻

stats/thresholds.go
```diff
@@ -26,70 +25,162 @@ import (
...
+			"reason: unable to cast parsed operator to string",
+		)
+	}
+
+	value, ok := parsed[2].(float64)
+	if !ok {
+		return nil, fmt.Errorf("the threshold expression parser failed; " +
+			"reason: unable to cast parsed value to underlying type (float64)",
+		)
+	}
```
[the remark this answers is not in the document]
oleiade: In this case, the threshold and parser combinators libraries should protect us from this. These checks are really here for protecting our future selves, if we were to make some breaking changes in either the threshold parser or combinators code, and to make the linter agree with the casts.

stats/thresholds.go
```diff
@@ -26,70 +25,162 @@ import (
...
+		passes = lhs >= t.parsed.Value
+	case "<=":
+		passes = lhs <= t.parsed.Value
+	case "<":
+		passes = lhs < t.parsed.Value
+	case "==":
+		passes = lhs == t.parsed.Value
+	case "===":
+		// Considering a sink always maps to float64 values,
+		// strictly equal is equivalent to loosely equal
```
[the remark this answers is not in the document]
oleiade: In this PR I try to fit as close as possible to the existing implementation, without changing the threshold expression format. As they were written in JS before, `===` is valid there, although it makes no sense in Go, which has a static type system. So supporting `===` is really there for compatibility reason, but as we a predefined type under the hood `==` and `===` are therefore equivalent.

pkg/combinators/combinators_test.go
```diff
@@ -0,0 +1,568 @@
...
+* This program is distributed in the hope that it will be useful,
+* but WITHOUT ANY WARRANTY; without even the implied warranty of
+* MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+* GNU Affero General Public License for more details.
+*
+* You should have received a copy of the GNU Affero General Public License
+* along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*
+ */
+package combinators
```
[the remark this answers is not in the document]
oleiade: I think you're right, it's probably fine to have it only once where the package description lives 👍🏻

pkg/combinators/error.go
```diff
@@ -0,0 +1,88 @@
...
+* but WITHOUT ANY WARRANTY; without even the implied warranty of
+* MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+* GNU Affero General Public License for more details.
+*
+* You should have received a copy of the GNU Affero General Public License
+* along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*
+ */
+
+package combinators
```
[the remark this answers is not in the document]
oleiade: I think you're right, it's probably fine to have it only once where the package description lives 👍🏻

stats/thresholds.go
```diff
@@ -26,70 +25,162 @@ import (
...
+// aggregation_method  -> trend | rate | gauge | counter
+// counter             -> "count" | "sum" | "rate"
+// gauge               -> "last" | "min" | "max" | "value"
+// rate                -> "rate"
+// trend               -> "min" | "mean" | "avg" | "max" | percentile
+// percentile          -> "p(" float ")"
+// operator            -> ">" | ">=" | "<=" | "<" | "==" | "===" | "!="
+// float               -> digit+ (. digit+)?
+// digit               -> "0" | "1" | "2" | "3" | "4" | "5" | "6" | "7" | "8" | "9"
+// whitespace          -> space | tab
```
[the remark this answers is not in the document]
oleiade: That's a good question. It's probably one of the blind spots of the BNF. The parser as I wrote it discards `\n`, `\r`, `\t`, and `\r\n`. I added really as sentinels, to be sure that even if programmers were to add them to their expression, they wouldn't impact the execution. The goal of a BNF being to be as explicit as possible on the format, I would actually add a definition for newline, as being `newline -> "\n" | "\r\n"`, and add it to assertion as optional. Does that address your question? 🐻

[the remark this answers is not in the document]
oleiade: Here's an updated BNF to explicit what I meant:

```
assertion           -> aggregation_method whitespace* operator whitespace* float newline*
aggregation_method  -> trend | rate | gauge | counter
counter             -> "count" | "sum" | "rate"
gauge               -> "last" | "min" | "max" | "value"
rate                -> "rate"
trend               -> "min" | "mean" | "avg" | "max" | percentile
percentile          -> "p(" float ")"
operator            -> ">" | ">=" | "<=" | "<" | "==" | "===" | "!="
float               -> digit+ (. digit+)?
digit               -> "0" | "1" | "2" | "3" | "4" | "5" | "6" | "7" | "8" | "9"
whitespace          -> space | tab
tab                 -> "\t"
space               -> " "
newline		 -> linefeed | crlf
crlf 		-> carriage_return linefeed
linefeed		-> "\n"
carriage_return	-> "\r"
```

stats/thresholds_parser.go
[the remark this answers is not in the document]
yorugac: Ah, I'm sorry, I should have been more verbose in this comment about precision :sweat_smile: I meant not the precision in this particular line but in all steps of thresholds processing together: string -> float64 -> string. I.e. it's worth to consider if there can be any loss / misrepresentation at some point and if it can, IMO, it should be either fixed or documented. I can't say that I see the problem with conversions right now but I'd have run some additional tests for edge cases, etc., to be sure. Just a suggestion :slightly_smiling_face: 

However, this does work with current `master`:
```
http_req_failed: ['rate<123e5'],
http_req_duration: ['p(34)<123e-5'],
```
But I don't think this would work in this PR because exponent is not processed in the value. To be clear, I'm not saying this _must_ work... e.g. was anyone actually using this form of floats in their scripts? Maybe it can be dropped? But this is probably something to decide / investigate more.

stats/thresholds.go
```diff
@@ -26,70 +25,185 @@ import (
...
-		return false, err
+func (t *Threshold) runNoTaint(sinks map[string]float64) (bool, error) {
+	// Extract the sink value for the aggregation method used in the threshold
+	// expression
+	lhs, ok := sinks[t.parsed.AggregationMethod]
+	if !ok {
+		return false, fmt.Errorf("unable to apply threshold %s over metrics; reason: "+
+			"no metric supporting the %s aggregation method found",
+			t.parsed.AggregationMethod,
+			t.parsed.AggregationMethod)
```
yorugac, line 68:
```suggestion
		return false, fmt.Errorf("unable to apply threshold %s over metrics; reason: "+
			"no metric supporting the %s aggregation method found",
			t.Source,
			t.parsed.AggregationMethod)
```

stats/thresholds.go
```diff
@@ -138,60 +252,54 @@ func NewThresholds(sources []string) (Thresholds, error) {
...
 
 // Run processes all the thresholds with the provided Sink at the provided time and returns if any
 // of them fails
-func (ts *Thresholds) Run(sink Sink, t time.Duration) (bool, error) {
-	if err := ts.updateVM(sink, t); err != nil {
-		return false, err
+func (ts *Thresholds) Run(sink Sink, duration time.Duration) (bool, error) {
+	// Update the sinks store
+	ts.sinked = make(map[string]float64)
+	f := sink.Format(duration)
```
yorugac, line 297: Looking at this with fresh eyes, I don't think custom percentiles are supported right now, are they? So here, `Format` in case of `Trend` will only calculate `P(90)` and `P(95)` by default. But what if user specifies something else? Like I just wrote in another comment:
```
http_req_duration: ['p(34)<123e-5']
````
It seems to me the old version was doing this in JS, with `jsEnvSrc` substitute...

Fixing this would likely require a cast to `TrendSink` as `P()` is not a part of `Sink` interface :slightly_frowning_face: I had to cast this in PRW extension too.

File: stats/thresholds.go (hunk 7 above)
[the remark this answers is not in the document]
Comment (yorugac) [reply stats/thresholds.go:87]: Got it, that's clear, thanks! But now I wonder if some people might expect a different behavior when using `==` and `===` in thresholds. I.e. isn't it more straight-forward to have just 1 operator without any ambiguity? :thinking:
I think this point could use input (or clarifications if it was already discussed in the past) from others :slightly_smiling_face:

File: stats/thresholds.go (hunk 10 above)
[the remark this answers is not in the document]
Comment (yorugac) [reply stats/thresholds.go:134]: Yes, I missed that yesterday: currently tabs are allowed inside threshold expression and newline is allowed in the end as you described.
So this is again more a question of design: _should_ there be tabs or newlines there? What's the use case that someone would need those chars in threshold expression? I was more thinking in line of simplifying it as much as possible and maybe help people avoid some mistakes in writing tests. I guess it can be seen both ways :thinking: 

And I agree that BNF is a very nice way to be clear on expectations in this case :+1:

File: stats/thresholds.go (hunk 12 above)
Comment (oleiade) [reply stats/thresholds.go:297]: That's a good question! I must admit, I focused on what ended up in the sinks, and not so much on how it ended in there. I'm going to run a bunch of tests, see what happens, and I'll let you know. Not only that, but I might ask you for help if things go south 🙏🏻

File: stats/thresholds.go (hunk 10 above)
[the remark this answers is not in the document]
Comment (oleiade) [reply stats/thresholds.go:134]: That's a good question. A decision needs to be made here. As with the rest of this PR, my approach has been to get as close as possible to what is documented and already implemented (with my probably limited view of it still). I would tend to think that having the parser discard spaces and tabs is a nice to have, as it makes it permissive to "not perfect" input. I have no opinion on the newlines. Furthermore, I don't think I have the experience/knowledge of the test scripts and how they're used in the wild yet to be an able to decide, so I'll let you folks decide :)

**Comment by MStoykov** [comment]
Having had (not so) fast of a look through the changes it looks pretty nice, but I kind of feel like the whole combinators based parser is a bit of overkill:
- while there are some ideas of extending the thresholds, I very much doubt this will happen soon, or that it will use the same syntax. Any change to this (or any future) threshold definition syntax needs to also be done in the k6 cloud and that in general means that we will likely need quite a big redesign. My expectation is that we will completely redesign the system after the metric system redesign and likely whatever is currently being done will *only* be kept for backwards compatibility
- reusing someone else's code like this will likely require it to not be licensed as theirs and not have a copyright by us. This I guess is more of a legal problem but does also (at least for me) tip the balance to not using it if possible.
- I wrote [this](https://go.dev/play/p/OxDzfJnYMuE) in 15 minutes, in general, because of some of the other comments here. IMO it is some error handling away from being production-ready(so the remaining 80% ;) ). Also happens to be one screen long and at least to me *a lot* more understandable than the 1000+ lines of code and tests for the parser you wrote.  The main place that I would expect your approach will win (apart from extensibility, although even that is dependant on the change) is error handling - I would expect your parser will give nicer errors, but even that can be .. improved. And given the simplicity of it all I really don't see why we would want 1000 lines + of parser :( 

I will go through the rest of the comments and comment on those, but just wanted to start the discussion on that as this will probably cut in half the amount of code added by this PR 

cc @na--

File: stats/thresholds.go (hunk 7 above)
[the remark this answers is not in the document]
Comment (MStoykov) [reply stats/thresholds.go:87]: I think it's perfectly fine for `===` to be equal ( :stuck_out_tongue_winking_eye: ) to `==` and it's important to keep this as compatible with what is in the cloud and this is supported there.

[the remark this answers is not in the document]
Comment (MStoykov) [reply stats/thresholds.go:87]: also I would prefer for this and the previous case to be merged ;)

File: stats/thresholds.go (hunk 10 above)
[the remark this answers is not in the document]
Comment (MStoykov) [reply stats/thresholds.go:134]: if the cloud supports tabs we should support them as well, if not - not

File: stats/thresholds.go (hunk 12 above)
Comment (MStoykov) [reply stats/thresholds.go:297]: yes it needs to support `p(<random float between 0 and 100>)`, and yes previously that was done through the "reexporting" of the Trend's [P method](https://github.com/grafana/k6/blob/14079775b9bad0b8b2f503a7fa1e787fbc5cba36/stats/sink.go#L113-L131)

Comment (oleiade) [reply stats/thresholds.go:297]: With my recent changes, the parser supports and verifies that supplied values are floats in the range `[0, 100]` in either integer or decimal notation. It doesn't support the *exponent* notation of floating numbers, though. Should it?

**Comment by oleiade** [comment]
Hey @MStoykov thanks for your input 🙏🏻 

- Regarding the scope of the changes. I started implementing this, trying to follow what's described in #1443 and #1441 as close as possible. Hence, the initial choice to start with a more fully-featured parsing approach. It's only later that it was clarified to me that we didn't actually intend to implement the scope described in these issues (?), and at that point I already had used the combinators.
- Regarding the licensing, I followed what the MIT License requires: "The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.". But indeed, I don't know what repercussions this leads to when it comes to the compatibility with our own license. PS: I could also republish the code in my own name if relevant.
- Regarding the usage of combinators and of a dedicated parsing library all in all, and if we keep in mind that I wrote that initially with a larger scope in mind, you're right, a smaller parser could be written, and I'm open to do that. But I'd like to point out some upsides of the approach I took:
- its flexibility (you can build any kind of parsers using that technique, without having to rewrite a whole new parser every time, it's really more of a framework than anything else).
- If you put the combinators toolkit aside, the parsing code (threshold_parser) is actually quite small.
- error management; which I think is critical in terms of UX (this parser tells you which part of the expression is wrong, why, and at what character).

All in all, I must say I'm still a bit confused by the scope and the plan regarding thresholds, as I've heard "different bell sounds" 🇫🇷. So I'll trust you folks on what's the best approach here:
- Would you prefer I drop the combinators approach and replace it with a custom smaller parser indeed?
- Does the BNF I defined works? I've aligned with what's in our documentation, but If relevant, who should I talk to, to align it with the cloud's?

File: stats/thresholds.go (hunk 10 above)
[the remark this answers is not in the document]
Comment (oleiade) [reply stats/thresholds.go:134]: Looking at the cloud's parser, they drop spaces, but do not seem to touch at tabs, nor newlines actually: I assume they produce an error in that case.

See https://github.com/grafana/k6-cloud-lib/blob/develop/k6_lib/utils/thresholds.py#L66

Thus I will make the parser *not* ignore tabs nor new lines.

**Comment by na--** [comment]
_(Disclaimer: I haven't reviewed the whole PR yet, just skimmed it... my current answer is mostly about the latest discussion :arrow_up: about the custom parser)_

TLDR version: I generally agree with @MStoykov.

Here's my rationalle... We don't know how the k6 thresholds of the future will look like... [#1441](https://github.com/grafana/k6/issues/1441) is one suggestion for one aspect of them, there are various improvement ideas in [#1136](https://github.com/grafana/k6/issues/1136) and some of its linked issues and in [this comment](https://github.com/grafana/k6/issues/1443#issuecomment-626957527) (probably others as well), [#1831](https://github.com/grafana/k6/issues/1831) will almost certainly also affect how we do things...

So, we don't know _how_ thresholds 2.0 will look like, but it's almost certainly going to be different than the current thresholds. Our current syntax is just too limited to support a lot of the things we want. Though, even after we have better thresholds and deprecate the current syntax, we probably won't ever stop supporting it... It's such a major feature that all old k6 scripts should continue to work. We'll just silently (or with a warning) support the old format, internally translating it to the new feature (since its capabilities will be a superset of the current ones), but we probably won't ever develop it or add new features to it.

And that's the crucial reason why I think we should go with simplicity here. There's no need to develop a fancy custom parser for something that is likely going to be deprecated soon. The current format should be parsable in a few dozen lines even without any regexes, as @MStoykov's demo shows. We also don't need perfect UX, the current UX is so bad that almost anything will be a significant improvement :disappointed: Just take a look at the examples in https://github.com/grafana/k6/issues/961... So errors like `"foo" is an invalid threshold expression` are more than good enough, for now.

Again, we don't know how parsing of the future thresholds will look like. We might go with our own custom parser, and then a fancy fully-featured one will be great. But we might decide to adopt PromQL or some other widely-used format for them and use a ready library for parsing it. So, no need to over-engineer anything here, just to get rid of goja and have basic validation of the threshold syntax...

**Comment by oleiade** [comment]
Thanks for the context clarification @na--. Makes sense. I'll strip down the implementation in accordance then.

However, something that the review up to this stage outlined is that there were different perceptions as to what the accepted format should be. At the moment, any expression that's valid JS is accepted (although I think we do not document it as such). Do we target the same level of compatibility to not break anything, or is it okay to support the absolute minimum even if that's effectively a breaking change in the user-facing interface?

Can we agree on a target specification for the expressions to make our(my) lives easier? Incorporating my understanding of the current documented state, and the outcome of the discussions that happened in this PR, I would expect this to not break anything for users:

```
assertion           -> aggregation_method whitespace* operator whitespace* float
aggregation_method  -> trend | rate | gauge | counter
counter             -> "count" | "sum" | "rate"
gauge               -> "last" | "min" | "max" | "value"
rate                -> "rate"
trend               -> "min" | "mean" | "avg" | "max" | percentile
percentile          -> "p(" float ")"
operator            -> ">" | ">=" | "<=" | "<" | "==" | "===" | "!="
float               -> digit+ (. digit+ | "e" -? digit+)?
digit               -> "0" | "1" | "2" | "3" | "4" | "5" | "6" | "7" | "8" | "9"
whitespace          -> space
space               -> " "
```

**Comment by na--** [comment]
We have to double check exactly what the cloud supports, but here are some things that I noticed:
- Gauge metrics don't have `last`, `min` and `max` in k6, [only `value`](https://github.com/grafana/k6/blob/703fd5723f51f5fb7e24e87361f920870d2e3564/stats/sink.go#L85)
- similarly, Counter [doesn't have `sum`](https://github.com/grafana/k6/blob/703fd5723f51f5fb7e24e87361f920870d2e3564/stats/sink.go#L60-L61)
- Trend [has `med`](https://github.com/grafana/k6/blob/703fd5723f51f5fb7e24e87361f920870d2e3564/stats/sink.go#L156)
- `aggregation_method` is actually just the metric type, it's not part of the parsing here, instead we need look up the metric type from the new `Registry` based on the metric name
- we probably don't need to support `E` notation for float numbers? 
- we can probably support more white-space options beyond just space, depending on the cloud :thinking: 

For the first 3 points, if the cloud already supports these "operators", we can probably also easily implement them, see https://github.com/grafana/k6/issues/1312

**Comment by oleiade** [comment]
I've pushed a dumbed-down version of the parser as requested. 

- **regarding spaces**: the cloud drops any amount of whitespaces. As far as I can tell, it does not manage tabs, thus I haven't added them to the specification, nor support it in the parser.
- **regarding the exponential notation**: I've dropped any explicit support for it and rely on Go's [`strconv.ParseFloat`](https://pkg.go.dev/strconv#ParseFloat) method exclusively. My understanding is that `ParseFloat` does not support the exponential notation. 

It seems I copy-pasted an outdated definition indeed. Sorry for that. Here's the up to date one (incorporating your comments):

```
// assertion           -> aggregation_method  whitespace* operator whitespace* float
// aggregation_method  -> trend | rate | gauge | counter
// counter             -> "count" | "rate"
// gauge               -> "value"
// rate                -> "rate"
// trend               -> "avg" | "min" | "max" | "med" | percentile
// percentile          -> "p(" float ")"
// operator            -> ">" | ">=" | "<=" | "<" | "==" | "===" | "!="
// float               -> digit+ ("." digit+)?
// digit               -> "0" | "1" | "2" | "3" | "4" | "5" | "6" | "7" | "8" | "9"
// whitespace          -> " "
```

File: stats/thresholds_parser.go [hunk 13]
```diff
@@ -0,0 +1,145 @@
...
+	// Is the input one of the methods keywords?
+	for _, m := range methods {
+		if m == input {
+			return input, nil
+		}
+	}
+
+	// Otherwise, attempt to parse a percentile expression
+	if strings.HasPrefix(input, "p(") && strings.HasSuffix(input, ")") {
+		trimmed := strings.TrimSuffix(strings.TrimPrefix(input, "p("), ")")
```
MStoykov, line 135: Nit: I would've written this as `input[2:len(input)-1]` possibly replacing 2 and 1 with `len("p(")` 

Not particularly certain my shorter version is more readable as it's shorter, or your longer version is better with it being more specific/expressive  

I wonder if some helper function
```
func TrimSides(input, prefix, suffix string) (result string, didTrim bool) 
```
Will not be beneficial, I think I am doing this in other places as well.

But regardless, just mentioning it, it's not a requirement to be fixed.

stats/thresholds_parser.go
```diff
@@ -0,0 +1,145 @@
...
+		_, err := strconv.ParseFloat(trimmed, 64)
```
MStoykov, line 136: This will parse floats in the form 3.1415927E+02 I am technically for us *parsing* this, but I remember there was some discussion, so I am mentioning it

stats/thresholds.go
MStoykov: It seems to me that this code still doesn't support percentiles?!?

I would also, looking at this code and the other places where `sink.Format` is used, I kind of want to remove this method. It's in most cases used as a way to get info about values for the given sink, but then we also cast the sink to its concrete type in order to get even more info from it. 

Given that there are 4 sink types, I feel like a switch in all cases (as the fields might not be the same) might be better :thinking: .

Especially as it turns out that the [CounterSink](https://github.com/grafana/k6/blob/bd221f75725dbf7b62590829eea9f49029d1508a/stats/sink.go#L61) does actually return a `rate` but [the summary code overwrites it](https://github.com/grafana/k6/blob/bd221f75725dbf7b62590829eea9f49029d1508a/js/summary.go#L57-L61) possibly with the same value :facepalm:  

And in the REST API where it is used "correctly" we have a [minor bug](https://github.com/grafana/k6/issues/1253) about Trends and also doesn't show the same stuff as in the summary :facepalm: .

Fixing all of this out of the scope for this PR, but I kind of feel it might be better to *not* depend on what `Format` returns in the thresholds

MStoykov (changes requested):
LGTM :clap: , but percentiles definitely need fixing ;)

I have left other nitpicks, but more tests with percentiles will be appreciated ;)

stats/thresholds_parser.go
oleiade: My personal preference is to go with a more functional style and try to avoid "static" indices in code as much as I can, as they involve the overhead of thinking about what they mean.

I've added a private `trimDelimited(prefix, input, suffix string) string` helper function to hide the logic and allow a much clearer `trimDelimited("p(", input, ")")` call.

stats/thresholds_parser.go
```diff
@@ -0,0 +1,145 @@
...
+	return "", "", "", fmt.Errorf("malformed threshold expression")
+}
+
+// parseThresholdMethod will parse a threshold condition expression's method.
+// It assumes the provided input argument is already trimmed and cleaned up.
+// If it encounters a percentile method, it will parse it and verify it
+// boils down to an expression of the form: `p(float64)`, but will return
+// it verbatim, as a string.
+func parseThresholdAggregationMethod(input string) (string, error) {
+	methods := []string{
```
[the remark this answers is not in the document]
oleiade: Indeed! I've defined the tokens as constant strings, and the methods as global array `[7]string` instead 👍🏻

stats/thresholds_parser.go
```diff
@@ -0,0 +1,145 @@
...
+}
+
+// scanThresholdExpression scans a threshold condition expression of the
+// form: `aggregation_method operator value`. An invalid or unknown operator
+// will produce an error. However, no assertions regarding
+// either the left-hand side aggregation method nor the right-hand
+// side value will be made: they will be returned as is, only trimmed from
+// their spaces.
+func scanThresholdExpression(input string) (string, string, string, error) {
+	operators := []string{
```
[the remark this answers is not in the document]
oleiade: Indeed! I've defined the operator tokens as constant strings, the operators as a global `[7]string` array, and documented that the operator order matters 👍🏻

stats/thresholds_parser.go
oleiade: You're right 👍🏻 If no blocking objections are made, I would keep it that way, as it relies on a standard library method. Otherwise, I would specifically look for the `e+` sequence and make the parser fail accordingly.

stats/thresholds.go
```diff
@@ -138,60 +154,101 @@ func NewThresholds(sources []string) (Thresholds, error) {
...
+		ts.sinked["rate"] = sinkImpl.Value / (float64(duration) / float64(time.Second))
+	case *GaugeSink:
+		ts.sinked["value"] = sinkImpl.Value
+	case *TrendSink:
+		ts.sinked["min"] = sinkImpl.Min
+		ts.sinked["max"] = sinkImpl.Max
+		ts.sinked["avg"] = sinkImpl.Avg
+		ts.sinked["med"] = sinkImpl.Med
+		ts.sinked["p(90)"] = sinkImpl.P(0.90)
+		ts.sinked["p(95)"] = sinkImpl.P(0.95)
```
MStoykov, line 220: This shouldn't be needed if the below part will add the *one* percentile it needs

This seems like a good place for some optimization (not in this PR, this was the case before as well), but here we do generate a whole map with a bunch of values. But in most cases only 1 of those will be used at all.

This is *low prio* though IMO, only the percentiles calculations are "heavy" and as long as you drop this 2 precalculated which might not be needed, the below code will calculate just the ones it needs.

Also given that in this case it will calculate the percentile ones even if they are used in multiple thresholds, it might even not be beneficial to JIT calculate them if we also don't cache them :thinking: 

But I digress – just drop those two lines 

```suggestion
```

oleiade: Yeah, I had doubts about this too, but put them back as they are "default" percentiles, and wanted to make no assumptions to whether they would be present or not. Glad you clarified it, I'll drop them back indeed! 👍🏻

stats/thresholds_parser_test.go
```diff
@@ -0,0 +1,144 @@
...
+ */
+
+package stats
+
+import (
+	"reflect"
+	"testing"
+)
+
+func Test_parseThresholdExpression(t *testing.T) {
```
MStoykov, line 28: Is there a particular reason you decided to go with `Test_xxx` instead of what seems to be the standard `TestXxx` naming that we've been using through k6? 

I am not really particularly for one or the other, but I prefer if there is one that we stick to instead of it being different everywhere ;)

stats/thresholds_parser_test.go
```diff
@@ -0,0 +1,144 @@
...
+	tests := []struct {
+		name          string
+		input         string
+		wantCondition *thresholdExpression
+		wantErr       bool
+	}{
+		{"unknown expression's operator fails", "count!20", nil, true},
+		{"unknown expression's method fails", "foo>20", nil, true},
+		{"non numerical expression's value fails", "count>abc", nil, true},
+		{"valid threshold expression syntax", "count>20", &thresholdExpression{AggregationMethod: "count", Operator: ">", Value: 20}, false},
```
MStoykov, line 40: nit: I in general do not like "keyless" struct literals, both because they are harder to read and because adding one filed means you need to change all of them, even if most are fine with their zero value.

I do think that in this case (especially with some single digit tests) it's fine and maybe even better for readability. If any other reviewer also wants them to be with keys, perhaps we should change it though :shrug: 

The other cases in this file are getting more and more in the territory of me wanting them to not be keyless.

stats/thresholds_parser_test.go
```diff
@@ -0,0 +1,144 @@
...
+			t.Parallel()
+
+			gotCondition, err := parseThresholdExpression(testCase.input)
+			if (err != nil) != testCase.wantErr {
+				t.Errorf("parse() error = %v, wantErr %v", err, testCase.wantErr)
+				return
+			}
+			if !reflect.DeepEqual(gotCondition, testCase.wantCondition) {
+				t.Errorf("parse() = %v, want %v", gotCondition, testCase.wantCondition)
+			}
```
MStoykov, line 55: Is there any reason why you are not using [testify](https://pkg.go.dev/github.com/stretchr/testify) we are after all using it through the codebase?

stats/thresholds_test.go
```diff
@@ -25,99 +25,250 @@ import (
...
+			args{map[string]float64{"rate": 0.00001}},
+			false,
+			false,
+		},
+		{
+			"valid expression over non-existing sink",
+			fields{"rate>0.01", false, false, types.NullDurationFrom(2 * time.Second), &thresholdExpression{tokenRate, tokenGreater, 0.01}},
+			args{map[string]float64{"med": 27.2}},
+			false,
+			true,
```
MStoykov, line 144: I definitely want to know what the fifth element here is without having to go up to the definition of the type.

MStoykov (changes requested):
Looks much better IMO :clap:. I

 have some comments around testing, some more nitpicky others less nitpicky but it seems like most of them aren't a single case, and I opted for not pointing out all the cases before we discuss them.

stats/thresholds_parser_test.go
oleiade: Good catch!

We had this discussion internally a while ago, and really decided that test names shouldn't have an underscore separator, indeed.  I think these tests predate that discussion (if not, I'll blame VSCode's snippets instead).

stats/thresholds_parser_test.go
oleiade: I was unsure about the established best practice. Pretty sure I've seen both styles throughout the codebase, but my memory could also be failing me. In general, and in the spirit of "explicit is better than implicit", I agree with you.

I'll add the struct keys 😸

stats/thresholds_parser_test.go
oleiade: Once again, I'll blame VSCode. The Go extension's `generate unit test for function` command does use the standard package only; and that's the one I use to write all the boilerplate for me. I'll modify the code to use `testify` instead, and find out if I can customize the snippet command to avoid that in the future 👍🏻

stats/thresholds_test.go
oleiade: Agreed 👍🏻

oleiade:
@MStoykov Thanks a lot for your review. I have addressed your comments. Let me know if you have other comments, recommendations or questions 👍🏻 

Ping @na-- @yorugac The PR has been rewritten for the most part, and is now in a reviewable state again. Please take a look 👀 whenever you've got a moment 😸

stats/thresholds_test.go
```diff
@@ -25,99 +25,274 @@ import (
...
+	tests := []struct {
+		name    string
+		fields  fields
+		args    args
+		wantOk  bool
+		wantErr bool
+	}{
+		{
+			name:    "valid expression using the > operator over passing threshold",
+			fields:  fields{"rate>0.01", false, false, types.NullDurationFrom(2 * time.Second), &thresholdExpression{tokenRate, tokenGreater, 0.01}},
```
MStoykov, line 85: Here `fields` has a bunch of ... fields (:stuck_out_tongue: ) three of which are the same in every case. They also coincide with the fields of Threshold.

From the looks of it - the one that is actually used is also `parsed` which IMO can just be part of the top test case struct. 
Also see https://github.com/grafana/k6/pull/2251/files#r756909900 although from my understandign because this is `runNoTaint` that isn't really interesting parameter. Still (maybe an additional test) to check that it doesn't matter will bea good idea :shrug: .

stats/thresholds_test.go
```diff
@@ -25,99 +25,274 @@ import (
...
+			args:    args{map[string]float64{"rate": 1}},
```
MStoykov, line 86: Seems to me like you can just make `args` `map[string]float64` and skip some types and wrapping in the process

stats/thresholds.go
```diff
@@ -138,60 +154,99 @@ func NewThresholds(sources []string) (Thresholds, error) {
...
-		ts.Runtime.Set(k, v)
-	}
-	return nil
-}
-
-func (ts *Thresholds) runAll(t time.Duration) (bool, error) {
-	succ := true
-	for i, th := range ts.Thresholds {
-		b, err := th.run()
+func (ts *Thresholds) runAll(duration time.Duration) (bool, error) {
```
MStoykov, line 171: I would argue `duration` doesn't sound correct as this is the ["current test run duration"](https://github.com/grafana/k6/blob/c93adb3732ddb182ade693191bfe56ad0cfc1e96/core/engine.go#L449-L459). You might call it the duration so far, but I would argue `timeInTest` will be slightly better as "current" might be a bit misleading given that it's (at this point) somewhere in the past ;)

stats/thresholds_test.go
oleiade: For fields, VSCode snippets are at it again, you're right they're quite redundant. Going to address that indeed🙏🏻

stats/thresholds.go
oleiade: Thanks for the recommendation 👍🏻 I've renamed the parameter to `timeSpentInTest`, and it does make the code itself much clearer indeed 😸

stats/thresholds_test.go
oleiade: Agreed 👍🏻

stats/thresholds_test.go
```diff
@@ -22,102 +22,284 @@ package stats
...
+}
+
+func TestThreshold_runNoTaint(t *testing.T) {
+	t.Parallel()
+
+	type fields struct {
+		Source           string
+		LastFailed       bool
+		AbortOnFail      bool
+		AbortGracePeriod types.NullDuration
```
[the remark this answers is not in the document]
oleiade: Indeed, and as a result of @MStoykov's recommendation, I added a specific test for it 😸

yorugac (changes requested):
I think overall it's all clear and good :tada: My request changes are very minor but there are also a couple of more involved questions too, please see.

Also, out of curiosity: I ran benchmarks added in here and they perform nicely, but is there a way to compare it easily to the old version?

stats/thresholds_parser_test.go
```diff
@@ -0,0 +1,333 @@
...
+			wantMethod:   "foo",
+			wantOperator: "<=",
+			wantValue:    "bar",
+			wantErr:      false,
+		},
+		{
+			name:         "expression with unknown operator fails",
+			input:        "foo!bar",
+			wantMethod:   "",
+			wantOperator: "", wantValue: "",
```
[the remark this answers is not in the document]
MStoykov: Hm, I would argue they can just be left ... out and they will be empty ;)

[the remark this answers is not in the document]
oleiade: I prefer to keep them for it to be explicit and document that they won't be set in this context 😸

stats/thresholds.go
```diff
@@ -138,60 +154,99 @@ func NewThresholds(sources []string) (Thresholds, error) {
...
+
+			// Percentile value has already been parsed as a float64, and should be valid.
+			// The error check is only here to protect our future selves.
+			percentile, err := strconv.ParseFloat(trimDelimited("p(", threshold.parsed.AggregationMethod, ")"), 64)
+			if err != nil {
+				return false, fmt.Errorf(
+					"unable to parse percentile value as 64-bit floating point number; "+
+						"reason: %w", err,
+				)
+			}
```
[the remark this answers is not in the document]
oleiade: The idea was to make a tradeoff. Considering that percentile are a "special-case", as they hold an actual value, and are the only ones doing so, I preferred parsing them as a "validation that the expression makes sense", and extract their value only later "just in time", as they're actually used. That way, I avoided having a dedicated nullable value in the `thresholdExpression`.

I can also envision a version of this where we introduce a `tokenPercentile`, and a `null.Float` (if it exists) in the `thresholdExpression`. We would then in the context of this version iterate over the thresholds, and if the aggregation method is a percentile, then we verify the value is actually set and add it to the sinks. Would you prefer it that way? 😸 

I truly have no preference. I can see on one side, we pre-mash the work, while the other avoids adding abstraction. Let  me know 🐻

stats/thresholds.go
oleiade: I have opened a dedicated issue: #2320. Please don't hesitate to add information you find missing @MStoykov. Resolving this discussion; but let me know if I've missed something that still needs addressing 🙏🏻

stats/thresholds.go
```diff
@@ -138,60 +154,99 @@ func NewThresholds(sources []string) (Thresholds, error) {
...
 // of them fails
-func (ts *Thresholds) Run(sink Sink, t time.Duration) (bool, error) {
-	if err := ts.updateVM(sink, t); err != nil {
-		return false, err
+func (ts *Thresholds) Run(sink Sink, duration time.Duration) (bool, error) {
+	// Initialize the sinks store
+	ts.sinked = make(map[string]float64)
+
+	// FIXME: Remove this comment as soon as the stats.Sink does not expose Format anymore,
+	// see: https://github.com/grafana/k6/pull/2251#discussion_r769807203
```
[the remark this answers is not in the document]
oleiade: I have opened a dedicated issue: #2320, and updated the comment accordingly 👍🏻

oleiade:
Note that I will make sure to squash all "Update", and "fixup!" commits before merging 👍🏻

stats/thresholds.go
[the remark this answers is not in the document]
yorugac: Hm, looking at it with fresh eyes, yes, you're right, it can be seen either way... I think my first "instinct" is to minimize number of possible errors and parsing which would be with an additional field in struct. It also reduces number of lines :smile: But depending on how thresholds evolve in the future, that field might become obsolete / to be moved, etc.
So yes, I have a slight preference towards the `tokenPercentile` field but not a very strong one, in case you change your mind and would like to keep the current version :slightly_smiling_face:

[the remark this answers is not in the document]
oleiade: I've modified the implemented design accordingly 👍🏻. The `thresholdExpression` type now has an `AggregationValue` field that's a `null.Float`, and the parser defines a `tokenPercentile` constant too. Let me know if it looks good on your side 😸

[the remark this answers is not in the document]
yorugac: Yes, I think it "reads" well and easy! Thank you :slightly_smiling_face:

stats/thresholds.go
```diff
@@ -73,23 +54,57 @@ func newThreshold(src string, newThreshold *goja.Runtime, abortOnFail bool, grac
...
-func (t Threshold) runNoTaint() (bool, error) {
-	v, err := t.rt.RunProgram(t.pgm)
-	if err != nil {
-		return false, err
+func (t *Threshold) runNoTaint(sinks map[string]float64) (bool, error) {
+	// Extract the sink value for the aggregation method used in the threshold
+	// expression
+	lhs, ok := sinks[t.parsed.AggregationMethod]
+	if !ok {
+		return false, fmt.Errorf("unable to apply threshold %s over metrics; reason: "+
```
na--, line 66: The next step here, in another PR, would be to tie the threshold validation with the metrics in our metrics registry. This validation is awesome, much better than what existed (or rather, what didn't exist :sweat_smile:) before, but we can do it right after the initial execution of the init context. Right now, this script:
```js
import { sleep } from 'k6';
import { Counter } from 'k6/metrics';


const myCounter = new Counter('my_counter');

export const options = {
    vus: 5,
    iterations: 30,
    thresholds: {
        my_counter: ['p(95)<200'], // error, Counters do not have p()
    },
};


export default function () {
    console.log(`[t=${(new Date()) - exec.scenario.startTime}ms] VU{${exec.vu.idInTest}} ran scenario iteration ${exec.scenario.iterationInTest}`);
    myCounter.add(1);
    sleep(1);
}
```

will emit something like this:
```
INFO[0000] [t=0ms] VU{3} ran scenario iteration 0        source=console
INFO[0000] [t=0ms] VU{5} ran scenario iteration 2        source=console
INFO[0000] [t=0ms] VU{2} ran scenario iteration 1        source=console
INFO[0000] [t=0ms] VU{1} ran scenario iteration 3        source=console
INFO[0000] [t=0ms] VU{4} ran scenario iteration 4        source=console
INFO[0001] [t=1001ms] VU{2} ran scenario iteration 7     source=console
INFO[0001] [t=1001ms] VU{4} ran scenario iteration 5     source=console
INFO[0001] [t=1001ms] VU{3} ran scenario iteration 9     source=console
INFO[0001] [t=1001ms] VU{5} ran scenario iteration 6     source=console
INFO[0001] [t=1001ms] VU{1} ran scenario iteration 8     source=console
ERRO[0002] Threshold error                               component=engine error="threshold 0 run error: unable to apply threshold p(95)<200 over metrics; reason: no metric supporting the p(95) aggregation method found" m=my_counter
```

but right now there is nothing preventing us from validating the thresholds _before_ we actually start running the script, `registry` should have all of the metrics (custom or built-in) and their types right after the last line here: https://github.com/grafana/k6/blob/23575eb63c948afc34dbe58a5e24d803305b4803/cmd/run.go#L113-L115

na--: Fixing https://github.com/grafana/k6/issues/1346 can probably be done at the same time as :arrow_up: , but I'm not finding an already open issue for it. @oleiade, do you want to create one or should I?

na-- (approved):
LGTM, so much nicer than the JS hack :heart: :sweat_smile: it also unlocks us to implement so many other fixes and improvements like https://github.com/grafana/k6/pull/2251/files#r782875298, https://github.com/grafana/k6/issues/1346 and threshold improvements in general :tada:

stats/thresholds.go
na--: And, thinking about it, https://github.com/grafana/k6/issues/2320 will also be easier when we don't have to do any validation in `Thresholds.Run()` :thinking: :tada: Each threshold can be "pre-compiled" by just turning it into a Go lambda or something like that. Once we do the validation once after the init context execution, that a threshold impression is valid for the metric type it is defined for, we shouldn't need any more validation afterwards :tada:

MStoykov (approved):

stats/thresholds.go
oleiade: Very good points indeed! I've created #2330 to keep track of it 🎉

oleiade:
We experienced issue with this feature in the larger, internal, context of K6, and decided to revert it, and delay its release to `v0.37.0`. See #1443 for more details.

PR #2255, opened 2021-11-19
Added progression variable to the ramping arrival rate

### LINKED ISSUE #2252: [Feature Request]:  Add stepped-arrival-rate
### Feature Description

In addition to the ramping-arrival-rate 
one of the most common features in loadtesting is stepping. 
For example when you know that your system must hold up to 1000RPS. 
You want to test it step by step in the following manner:
100 rps for 5 minutes
up to 300 rps for 5 minutes
up to 600 rps for 10 minutes
up to 1000 rps for 30 minutes.

Right now it's possible to configure such steps with the following scenarios
```
cenarios: {
    contacts: {
      startRate: 10,
      timeUnit: '1s',
      preAllocatedVUs: 50,
      maxVUs: 1000,
      stages: [
         {target: 100, duration: '1s' },
        {target: 100, duration: '5m' },
         {target: 300, duration: '1s' },
        {target: 300, duration: '5m' },
         {target: 600, duration: '1s' },
        {target: 600, duration: '10m' },
         {target: 1000, duration: '1s' },
        {target: 1000, duration: '30m' },

      ],
    },
```

For me right now it's looks like woraround. 

My suggestion is to add "stepped-arrival-rate":
Where target will mean not the final target of this stage but the RPS which we should hold during the 'duration'.

With such configuration, the scenarios will looks like

```
scenarios: {
    contacts: {
      executor: 'stepped-arrival-rate',
      startRate: 10,
      timeUnit: '1s',
      preAllocatedVUs: 50,
      maxVUs: 1000,
      stages: [
        {target: 100, duration: '5m' },
        {target: 300, duration: '5m' },
        {target: 600, duration: '10m' },
        {target: 1000, duration: '30m' },
      ],
    },
```
Which is more clear to read and understand in my opinion

LaserPhaser:
Added an ability to control how the arrival rate will be added
Linear - default
instant - optional

Closes #2252


lib/executor/helpers.go
```diff
@@ -72,6 +72,16 @@ func validateStages(stages []Stage) []error {
 		} else if s.Target.Int64 < 0 {
 			errors = append(errors, fmt.Errorf("the target for stage %d shouldn't be negative", stageNum))
 		}
+
+		switch s.Progression.String {
```
oleiade, line 76: Considering `s.Progression` is defined as a `null.String` I would follow the same procedure as for the other field, and at least append an error if `s.Target.Valid` is `false.

lib/executor/helpers.go
```diff
@@ -72,6 +72,16 @@ func validateStages(stages []Stage) []error {
...
+		case "linear",
+			"constant",
```
oleiade, line 78: Did you mean "instant" instead? If not, could you clarify your reasoning? My assumption is that a `constant` interpolation would either mean a linear one (as in, "a constant rate of change"), or, no movement at all, in which case, this usage is already addressed by the existing feature-set of options (although it might feel "hacky": `stages: [{duration: '1m', target: 10}, {duration: '10m', target: 10}`, would, from my understanding, perform a "constant" rate of change).

lib/executor/ramping_vus.go
```diff
@@ -54,6 +54,7 @@ func init() {
 type Stage struct {
 	Duration types.NullDuration `json:"duration"`
 	Target   null.Int           `json:"target"` // TODO: maybe rename this to endVUs? something else?
+	Progression null.String     `json:"progression"`
```
oleiade, line 57: I would prefer to name it `Interpolation` but as the naming is already being discussed in [#2256](https://github.com/grafana/k6/issues/2256), I'm fine with leaving it at that for now :+1:

lib/executor/ramping_arrival_rate.go
```diff
@@ -271,24 +271,34 @@ func (varc RampingArrivalRateConfig) cal(et *lib.ExecutionTuple, ch chan<- time.
...
-			endCount += dur * ((to-from)/2 + from)
-			for ; i <= endCount; i += float64(next()) {
-				// TODO: try to twist this in a way to be able to get i (the only changing part)
-				// somewhere where it is less in the middle of the equation
-				x := (from*dur - noNegativeSqrt(dur*(from*from*dur+2*(i-doneSoFar)*(to-from)))) / (from - to)
-
-				ch <- time.Duration(x) + stageStart
-			}
-		} else {
+		switch progression := stage.Progression.String; progression {
```
oleiade, line 274: This looks like a good refactor opportunity: to have the interpolation functions being dedicated either functions or methods called by the various switch cases. One good hint from my point of view, is that the `constant` branch is basically a duplicate of the non-ramping behavior of the `linear` one.

LaserPhaser:
@oleiade Thank you for reply, 
It's basically my fault that I did not close this PR. 
I've already updated the K6 documentations and added the example on how to create stepped-arrival rate. 
https://k6.io/docs/examples/instant-load-increase/

Thank you!

PR #2262, opened 2021-11-24
Disable linting errors with capital letter

MStoykov:
provoked by https://github.com/grafana/k6/pull/2239#discussion_r755353180

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -30,6 +30,8 @@ issues:
    #TODO: remove this exclusion after https://github.com/golang/lint/issues/381
    # is fixed by https://github.com/golang/lint/pull/380 or something else
    - "should have a package comment, unless it's in another file for this package"
+   # most k6 errors are actually used as exceptions for users
+   - "error strings should not be capitalized or end with punctuation or a newline"
 
 linters-settings:
   govet:


codebien:
Continuing discussion from https://github.com/grafana/k6/pull/2239#discussion_r755896919, if we merge this we could get something like: `This is the wrap error: This is the original error`, are we fine with it?

MStoykov:
@codebien we could, but the solutions to this IMO are:
1. always to have this be lower case - somewhat worse UX for users, fine, but we already have around 10 places where this isn't true
2. on a case by case to add `//nolint: golint`  and probably `stylecheck` as well - which in practice does the same but also disables all other things they will warn us about and need to be added to each occurance
3. as you've said do this somewhere else, but in the case of goja that someone else is in goja or to wrap *all* things returning error to goja (or panicking through common.Throw) in  something - completely out of the question IMO
4. sometimes having errors that aren't for users and have capital letters and or punctuation - seems pretty w/e to me ;)

oleiade:
I have strong opinions on not having capital letter, nor ending period when it comes to logging. Assuming the user-facing logging happens only in the packages under `cmd/`, would disabling this linting rule only for these packages be a good compromise? Another idea would be to have a dedicated logger, with dedicated formatting rules for user-facing logging (which could also not involve the `[ERROR]` prefix for instance): such as capitalizing the first character, and adding a period at the end if not present (in line with what @codebien proposes)?

codebien:
https://github.com/grafana/k6/blob/14079775b9bad0b8b2f503a7fa1e787fbc5cba36/cmd/root.go#L297-L307
I tried a simple test, seems something like the following could work for these formatters:
```go
	default:
		c.logger.SetFormatter(&log.CapitalizeFormatter{
			Formatter: &logrus.TextFormatter{ForceColors: loggerForceColors, DisableColors: noColor},
		})
		c.logger.Debug("Logger format: TEXT")
	}

```
where the wrapper should be similar to:
```go
type CapitalizeFormatter struct {
	Formatter logrus.Formatter
}

func (f *CapitalizeFormatter) Format(e *logrus.Entry) ([]byte, error) {
	if e.Level == logrus.ErrorLevel && len(e.Message) > 0 {
		e.Message = strings.TrimPrefix(e.Message, "GoError: ")
		e.Message = strings.ToUpper(string(e.Message[0])) + e.Message[1:]
	}
	return f.Formatter.Format(e)
}

```

WDYT?

oleiade:
I like it, that works for me 👍🏻

na--:
Soft :-1: - it doesn't seem like a good idea to have a special log formatter just for this, seems like there is a ton of potential for unintended side-effects...

Besides, I remember we discussed that maybe we shouldn't actually use the logger to output any errors that occur. So wouldn't it be easier to refactor this code: https://github.com/grafana/k6/blob/a8a0c29e8250fbe8772a0c4c2e2767ebcf18d8b9/cmd/root.go#L182-L201

As it's obvious from that code, we already have a few interfaces that allow us to attach exit codes, hints and stacktraces to specific errors. We can add another custom error type that is meant for user-facing error messages (i.e. basically exceptions) and which can be capitalized? :man_shrugging:

MStoykov:
I am also very much against changing the message in the logger for whatever reasons. 

Also, this was mostly in discussion with *exceptions* being thrown in js. Having them having different messages if you check them in code and if you let them interrupt the execution seems like a pretty bad idea to me 

I will close this as apparently there doesn't seem to be a consensus and the whole point of me trying to stop the linter from doing this is so we don't have to constantly ignore it or make messages tailored for it. This doesn't seem to be that big of a problem on second look especially compared to the apparent disagreement in this thread :) 

Even there is a solution it obviously won't be this one

PR #2263, opened 2021-11-24
Drop mentions of core.js in options

MStoykov:


cmd/runtime_options.go
```diff
@@ -60,9 +60,9 @@ func runtimeOptionFlagSet(includeSysEnv bool) *pflag.FlagSet {
 	flags.Bool("include-system-env-vars", includeSysEnv, "pass the real system environment variables to the runtime")
 	flags.String("compatibility-mode", "extended",
 		`JavaScript compiler compatibility mode, "extended" or "base"
-base: pure Golang JS VM supporting ES5.1+
-extended: base + Babel with ES2015 preset + core.js v2,
-          slower and memory consuming but with greater JS support
+base: pure goja - Golang JS VM supporting ES5.1+
+extended: base + Babel with parts of ES2015 preset
```
MStoykov, line 64: I wonder if we want to just list classes and import/export here?

cmd/runtime_options.go
```diff
@@ -60,9 +60,9 @@ func runtimeOptionFlagSet(includeSysEnv bool) *pflag.FlagSet {
...
+		  slower to compile in case the script uses unsupported by base syntax
```
imiric, line 65: I preferred the original "slower and memory consuming but with greater JS support" because we should mention it uses more memory, and "in case" makes it sound that "slower to compile" is optional, while that's always the case with `extended`, and it's the default anyway.

cmd/runtime_options.go
imiric: It would be too much information for this help output, so no, but we could mention it in the docs.

cmd/runtime_options.go
MStoykov: it isn't really "more memory" consuming ... in most cases. Unless you have a very large script that when it gets transpiled babel actually takes a lot of memory *and* at the same time you run with low enough number of VUs that this memory is then not used, it isn't really more memory consuming ;). The memory consumption was predominantly because core.js was adding 2mb+ to each VU.
and it's not *always* the case, if you run it against a script that is just supported by goja it won't even parse the babel, much less execute it.

imiric: OK, clear, thanks. Then I would just reword it to "slower to compile in case the script uses syntax unsupported by base".

imiric (approved):

oleiade (approved):

Issue #1357, opened 2020-03-11
Flaky tests


na--:
Just a list of tests we have noticed are flaky in CI runs, so we don't forget to fix/refactor them before merging #1007 in `master`:

`TestVariableLoopingVUsRun` - https://ci.appveyor.com/project/robingustafsson/k6/builds/31408611

na--:
`TestVariableLoopingVUsRampDownNoWobble` - https://circleci.com/gh/loadimpact/k6/9110

imiric:
I refactored these two tests and they seem more stable, but while stress testing `lib/executor`, I've also seen these fail occasionally:

- `TestPerVUIterationsRunVariableVU`
- `TestConstantArrivalRateRunCorrectRate`
- `TestExternallyControlledRun`

Will look into them next before creating a PR for this.

MStoykov:
they will probably get changed with the striping algorithm and rewrite so maybe just wait until then ?

imiric:
Another one :disappointed:: `core.TestMetricsEmission`: https://circleci.com/gh/loadimpact/k6/9243

na--:
Another `TestConstantArrivalRateRunCorrectRate` failure: https://circleci.com/gh/loadimpact/k6/9464

na--:
`TestExternallyControlledRun` flakiness: https://circleci.com/gh/loadimpact/k6/9530

na--:
Food for thought: https://dmitryfrank.com/articles/mocking_time_in_go (from https://news.ycombinator.com/item?id=23292215)

imiric:
I'm exploring that article and the `time` mocking packages, but one thing concerns me, mentioned in https://github.com/golang/go/issues/8869:

> The benbjohnson clock package attempts to make Ticker and Timer calls (as well as their
> related AfterFunc and Sleep methods) work against a fake time that can be set and
> updated in tests. _However, it relies on some micro-sleeps and runtime.Gosched calls that
> are obviously going to be flaky._

It would be a substantial effort to convert the entire codebase to use this `Clock` interface. Is that really worth it if it doesn't guarantee deterministic tests?

Also see https://github.com/grafana/metrictank/pull/342, where they abandon it altogether.

The article author uses a [fork](https://github.com/dimonomid/clock) where this might not be an issue(?), but not sure if this can be fixed outside of the stdlib.
[...]

PR #2269, opened 2021-11-29
WIP Fix/1986 mock time in test

oleiade:
This is a draft Pull Request of my attempt to mock time in test #1986 (and to hopefully get to solve #1357 in the process?). I've focused mostly on executors in `lib/executor` as they seemed a good first candidate.

In its current state, the PR adds a dependency to [benbjohnson/clock](https://github.com/benbjohnson/clock), and integrates its `Clock` interface in the `BaseConfig` of the executors. As a default, it ensures the "standard" clock is used (the one that uses the standard library's `time` package under the hood), but it also allows us to use a `clock.Mock` instance  that lets us control the passage of time.

While I was able to integrate the interface in the executors, and replace our usages of `time` with the added `clock` in the code. I've been struggling to get tests to pass using that technique: especially finding out where and when to make the time pass in the tests code. 

If you have ideas, recommendations, or want/can offer some help, this PR is here for that :)

PS: this PR's branch is based upon #2268, and you can safely ignore any commit starting with "Refactor", to focus on commits starting with "WIP" instead.

### DIFF
--- a/core/engine.go
+++ b/core/engine.go
@@ -27,6 +27,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/benbjohnson/clock"
 	"github.com/sirupsen/logrus"
 	"gopkg.in/guregu/null.v3"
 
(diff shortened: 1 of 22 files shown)


oleiade:
Closed. See #1986 for more details.

PR #2270, opened 2021-11-30
Drop strvals use strings.Split

MStoykov:


MStoykov:
This only drops like 24kb from the binary but obviously a lot more in dependency code

na-- (approved):
Huh, awesome, didn't realize we had just one place using it left! :tada: :partying_face:

MStoykov:
I was pretty sure I dropped it along the `kafka` output, but apparently not :shrug:

MStoykov:
@imiric if you can take a look as well as I see there aren't all that many tests for this :(

na--:
@MStoykov, that's what I initially thought, but the test are just in another place, see https://github.com/grafana/k6/blob/a8a0c29e8250fbe8772a0c4c2e2767ebcf18d8b9/cmd/config_consolidation_test.go#L433-L511

codebien (approved):
LGTM

lib/types/dns.go
```diff
@@ -213,33 +213,37 @@ func (c *DNSConfig) UnmarshalText(text []byte) error {
...
-	if err != nil {
-		return err
+	values := strings.Split(string(text), ",")
+	params := make(map[string]string, len(values))
+	for _, value := range values {
+		args := strings.SplitN(value, "=", 2)
+		if len(args) != 2 {
+			return fmt.Errorf("no value for key %s", value)
+		}
+		params[args[0]] = args[1]
```
imiric, line 223: I'd rather see this as a separate helper function, but we can move it later.

There's also a similar [`lib.SplitKV()`](https://github.com/grafana/k6/blob/5c14dae129a5eb6ed309c94a194370878b7b08e5/lib/util.go#L27-L34).

MStoykov: I was initially going to use https://github.com/grafana/k6/blob/master/log/tokenizer.go#L87 , but then I decided that is too much for 6 lines.

SplitKV also just makes a value empty if there is only a key, which I don't think is ... a good idea :thinking:

Issue #1931, opened 2021-03-28
Shared-memory read-only files


na--:
Prompted by https://community.k6.io/t/ways-to-transfer-files-in-tests/1552

After https://github.com/k6io/k6/pull/1841, can we add an optional argument (e.g. `r`) to [`open()`](https://k6.io/docs/javascript-api/init-context/open-filepath-mode) that would allow us to have only a single copy in memory. Like a [`SharedArray`](https://k6.io/docs/javascript-api/k6-data/sharedarray), but for `string`/`ArrayBuffer` data?

And unlike `SharedArray()`, we probably don't even need to have a `name`, we can use the absolute path to the file as the cross-VU sharing key.

As a bonus, this `sync.Once`-like behavior will also fix https://github.com/k6io/k6/issues/1771, right?

MStoykov:
I have to add that the current way bodies are handled does at least 1 additional copy to send the body and 1 to keep it as `response.request.body` which means that while the above is probably great, not fixing those issues first will not have as big of an effect as probably everybody expects.

na--:
Hmm didn't we fix some of these things with https://github.com/k6io/k6/pull/1841 or the previous moves to `ArrayBuffer`? And if not, aren't these things fixable with enough `Proxy` magic?

Assuming one such shared binary file is just a wrapped `[]byte` on the Go side, with `Proxy` and freezing on the JS side to keep it immutable, we should be able to pass it around without any copies, right?

na--:
Regarding `Response.request.body`, we should be able to just reference the original `ArrayBuffer` or `string` JS variable in `Response.request.body`, right? That will prevent any unnecessary copying and will also work for the `SharedBlob` object (or whatever the thing we implement in this issue is going to be called). 

The only time `Response.request.body` should be different from the request body is when k6 marshals it, i.e. the user has passed a plain JS object that k6 internally encoded into a `application/x-www-form-urlencoded` or a `multipart/form-data` content-type string.

anton-makarov-photobox:
@na-- Hi. You mention that you open the file with `SharedArray`. Do you have an example of it? 

```
 k6:
    image: loadimpact/k6:latest
    networks:
      - performance_test_network
    volumes:
      - ./dist:/scripts
      - ./assets:/assets
    ports:
      - "6565:6565"
    environment:
      - K6_OUT=influxdb=http://influxdb:8086/k6
```

[...]

PR #2271, opened 2021-11-30
Make netext httpext Response's Request a pointer

oleiade:
A tiny PR making the `Request` attribute of `netext/httpext/Response` struct a pointer. This should avoid copies of said Request objects (as they can be rather big) to be made every time we instantiate a Response. This is in line with Go's standard library.

This is a tiny change, which, as far as I can tell, does not break tests. I've also verified that it does not break the JSON serialization of requests (pointers can be tricksters on that front). 

ref #1931

### DIFF
--- a/lib/netext/httpext/request.go
+++ b/lib/netext/httpext/request.go
@@ -249,7 +249,7 @@ func MakeRequest(ctx context.Context, preq *ParsedHTTPRequest) (*Response, error
 		transport = ntlmssp.Negotiator{RoundTripper: transport}
 	}
 
-	resp := &Response{ctx: ctx, URL: preq.URL.URL, Request: *respReq}
+	resp := &Response{ctx: ctx, URL: preq.URL.URL, Request: respReq}
 	client := http.Client{
 		Transport: transport,
 		CheckRedirect: func(req *http.Request, via []*http.Request) error {
--- a/lib/netext/httpext/response.go
+++ b/lib/netext/httpext/response.go
@@ -91,7 +91,7 @@ type Response struct {
 	OCSP           netext.OCSP              `json:"ocsp"`
 	Error          string                   `json:"error"`
 	ErrorCode      int                      `json:"error_code"`
-	Request        Request                  `json:"request"`
+	Request        *Request                 `json:"request"`
 }
 
 // NewResponse returns an empty Response instance.


codebien (commented):
This mostly exists for `response.request.body` but is it useful? Could be the same think Go is doing an option?

>// Request's Body is nil (having already been consumed).

[from](https://pkg.go.dev/net/http#Response):
```
	// Request is the request that was sent to obtain this Response.
	// Request's Body is nil (having already been consumed).
	// This is only populated for Client requests.
	Request *Request
```

yorugac (commented):
ATM, I'm not sure if this change will solve the problems mentioned in #1931 but it seems harmless enough. I have a question though: how does it correlate with the module's PR https://github.com/grafana/k6/pull/2226/ ?

oleiade:
@yorugac This tiny refactor is the result of both reading about #1931 and discussions I've had with @MStoykov. It doesn't claim in any way to be directly related/fixing #1931. Hence me only refering it to said issue, and not marking it as closing, to indicate it popped out in the same context. My bad if there was any confusion on that front :smile_cat:. I'm not sure I see correlation with #2226, or at least not ones that keep from merging this. Do you have anything specific in mind?

@codebien I am not entirely sure I understood your point, could you elaborate? Regardless of the larger context, I only meant to post a tiny improvement here, in the spirit of the [broken window theory](https://blog.codinghorror.com/the-broken-window-theory/): not necessarily to make a big structural change. If there are larger implications, I would prefer to open and address them as a separate issue if that's okay with you :smile_cat:

PS: Sourcegraph shows the Response's request is actually used in a few places accross the standard library.

Cheers :bear:

na-- (approved):

yorugac:
@oleiade I simply meant if there can be a conflict between this PR and #2226, as they're changing the same code: I didn't really look at the #2226 since I'm not among the reviewers there so wished to clarify that point. Anyway, I suppose @na--'s approval can be interpreted as an indirect answer being "no conflicts" :slightly_smiling_face:

yorugac (approved):

na--:
> Anyway, I suppose @na--'s approval can be interpreted as an indirect answer being "no conflicts" slightly_smiling_face

Rather "I need to rebase anyway, so I can fix any conflicts that arise" :sweat_smile:

MStoykov:
I don't really this helps all that much (and now I understand what you meant "doesn't use a pointer" when we're talking about this). The problem I was talking about with you about "the response having a copy of the request"(I think that is what I said) isn't so much it's a *copy* it's that it has the whole request with all of its different fields and in particular having the [Body of the *request* as a *string*](https://github.com/grafana/k6/blob/86e32e3e035a19432cdc4f70e7e2d3874fb6e035/lib/netext/httpext/request.go#L56). 

I do think this is *better* though so :+1: ;) 

p.s. @codebien I think you are conflating http.Request (from the standard library) and httpext.Request (which is discussed here) - the second doesn't even have a copy of the first (which might be an interesting change :thinking: )

MStoykov (approved):

PR #2273, opened 2021-11-30
Data Segment API

codebien:
This PR adds the primitives for accessing the execution segment info useful for data partitioning. It mostly adds the [x/segment](https://github.com/MStoykov/xk6-segment) extension to the core. 

This PR still lacks in terms of testing coverage, it's more a PoC PR for getting an agreement about the API and the use cases it should cover.

A basic example about how it's expected to be used:

```js
// the essential part exported from the `samples/segmented-index.js` sample 
import { SharedSegmentedIndex } from 'k6/segment';

export default function () {
    let iterator = new SharedSegmentedIndex('myarr');
    let index = iterator.next();

    const reqBody = JSON.stringify(data[index.unscaled]);
    var res = http.post('https://httpbin.test.k6.io/anything', reqBody, params);
    check(res, { 'status 200': (r) => r.status === 200 });

    console.log(`Something: ${res.json().json.something}`)
}
```
running it with the following command:
```sh
k6 run -i 5 --execution-segment "0:1/2" --execution-segment-sequence "0,1/2,1" ./samples/segmented-index.js
```
the expected console output:
```
INFO[0001] Something: something else 1                   source=console
INFO[0001] Something: something else 3                   source=console
INFO[0001] Something: something else 5                   source=console
INFO[0001] Something: something else 7                   source=console
INFO[0002] Something: something else 9                   source=console
```

### Questions

* Does it require a dedicated module? I think it could be directly part of `k6/data`.
* Do we expect the users will use both scaled and unscaled?
* `GoTo` is exported, how do we expect to use it?

### DIFF
--- a/js/initcontext.go
+++ b/js/initcontext.go
@@ -46,6 +46,7 @@ import (
 	"go.k6.io/k6/js/modules/k6/html"
 	"go.k6.io/k6/js/modules/k6/http"
 	"go.k6.io/k6/js/modules/k6/metrics"
+	"go.k6.io/k6/js/modules/k6/segment"
 	"go.k6.io/k6/js/modules/k6/ws"
 	"go.k6.io/k6/lib"
 	"go.k6.io/k6/loader"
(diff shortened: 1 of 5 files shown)


oleiade:
In the spirit of putting the review in context: I'd be keen to understand what's the use case/user scenario this addresses? Is there an issue giving some context opened somewhere? Cheers :smile:

codebien:
>In the spirit of putting the review in context

Yep, sorry, I missed it in the description. It's a first iteration for https://github.com/grafana/k6/issues/1539

codebien:
Let's move back to Draft because after a discussion with @MStoykov we identified that this requires more discussion and API definition in the issue before starting reviewing it.

codebien:
At the moment, this PR doesn't fit our expectations, so I'm closing it.

PR #2274, opened 2021-12-01
Add TimeFormat option for output/csv with 'unix' (default) and 'rfc3399' options

rpocklin:
Simple PR to increase time unit granularity with CSV file output to be more in-line with JSON file output.

**file.csv**

Before:
`http_req_receiving,1638331259,...`

After:
`http_req_receiving,1638331421334748,...`

### DIFF
--- a/output/csv/consts.go
+++ b/output/csv/consts.go
@@ -1,24 +1,3 @@
-/*
- *
- * k6 - a next-generation load testing tool
- * Copyright (C) 2016 Load Impact
- *
- * This program is free software: you can redistribute it and/or modify
- * it under the terms of the GNU Affero General Public License as
- * published by the Free Software Foundation, either version 3 of the
- * License, or (at your option) any later version.
- *
- * This program is distributed in the hope that it will be useful,
- * but WITHOUT ANY WARRANTY; without even the implied warranty of
- * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
- * GNU Affero General Public License for more details.
- *
- * You should have received a copy of the GNU Affero General Public License
- * along with this program.  If not, see <http://www.gnu.org/licenses/>.
- *
- */
-
-// Package csv custom enum types used in config
 package csv
 
 // TimeFormat custom enum type
--- a/output/csv/consts.go
+++ b/output/csv/consts.go
@@ -1,0 +1,40 @@
+/*
+ *
+ * k6 - a next-generation load testing tool
+ * Copyright (C) 2016 Load Impact
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+// Package csv custom enum types used in config
+package csv
+
+// TimeFormat custom enum type
+type TimeFormat string
+
+// valid defined values for TimeFormat
+const (
+	Unix    TimeFormat = "unix"
+	RFC3399 TimeFormat = "rfc3399"
+)
+
+// IsValid validates TimeFormat
+func (timeFormat TimeFormat) IsValid() bool {
+	switch timeFormat {
+	case Unix, RFC3399:
+		return true
+	}
+	return false
+}
--- a/output/csv/output.go
+++ b/output/csv/output.go
@@ -54,6 +54,7 @@ type Output struct {
 	ignoredTags  []string
 	row          []string
 	saveInterval time.Duration
+	timeFormat   TimeFormat
 }
 
 // New Creates new instance of CSV output
@@ -87,6 +88,7 @@ func newOutput(params output.Params) (*Output, error) {
 
 	saveInterval := config.SaveInterval.TimeDuration()
 	fname := config.FileName.String
+	timeFormat := config.TimeFormat
 
 	if fname == "" || fname == "-" {
 		stdoutWriter := csv.NewWriter(os.Stdout)
@@ -97,6 +99,7 @@ func newOutput(params output.Params) (*Output, error) {
 			csvWriter:    stdoutWriter,
 			row:          make([]string, 3+len(resTags)+1),
 			saveInterval: saveInterval,
+			timeFormat:   timeFormat,
 			closeFn:      func() error { return nil },
 			logger:       logger,
 			params:       params,
@@ -114,6 +117,7 @@ func newOutput(params output.Params) (*Output, error) {
 		ignoredTags:  ignoredTags,
 		row:          make([]string, 3+len(resTags)+1),
 		saveInterval: saveInterval,
+		timeFormat:   timeFormat,
 		logger:       logger,
 		params:       params,
 	}
@@ -182,7 +186,7 @@ func (o *Output) flushMetrics() {
 		for _, sc := range samples {
 			for _, sample := range sc.GetSamples() {
 				sample := sample
-				row := SampleToRow(&sample, o.resTags, o.ignoredTags, o.row)
+				row := SampleToRow(&sample, o.resTags, o.ignoredTags, o.row, o.timeFormat)
 				err := o.csvWriter.Write(row)
 				if err != nil {
 					o.logger.WithField("filename", o.fname).Error("CSV: Error writing to file")
@@ -200,9 +204,17 @@ func MakeHeader(tags []string) []string {
 }
 
 // SampleToRow converts sample into array of strings
-func SampleToRow(sample *metrics.Sample, resTags []string, ignoredTags []string, row []string) []string {
+func SampleToRow(sample *metrics.Sample, resTags []string, ignoredTags []string, row []string,
+	timeFormat TimeFormat,
+) []string {
 	row[0] = sample.Metric.Name
-	row[1] = fmt.Sprintf("%d", sample.Time.Unix())
+
+	if timeFormat == RFC3399 {
+		row[1] = fmt.Sprint(sample.Time.Format(time.RFC3339))
+	} else {
+		row[1] = fmt.Sprintf("%d", sample.Time.Unix())
+	}
+
 	row[2] = fmt.Sprintf("%f", sample.Value)
 	sampleTags := sample.Tags.CloneTags()
 
--- a/output/csv/output.go
+++ b/output/csv/output.go
@@ -27,6 +27,7 @@ import (
 	"fmt"
 	"os"
 	"sort"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
@@ -88,7 +89,6 @@ func newOutput(params output.Params) (*Output, error) {
 
 	saveInterval := config.SaveInterval.TimeDuration()
 	fname := config.FileName.String
-	timeFormat := config.TimeFormat
 
 	if fname == "" || fname == "-" {
 		stdoutWriter := csv.NewWriter(os.Stdout)
@@ -99,7 +99,7 @@ func newOutput(params output.Params) (*Output, error) {
 			csvWriter:    stdoutWriter,
 			row:          make([]string, 3+len(resTags)+1),
 			saveInterval: saveInterval,
-			timeFormat:   timeFormat,
+			timeFormat:   config.TimeFormat,
 			closeFn:      func() error { return nil },
 			logger:       logger,
 			params:       params,
@@ -117,7 +117,7 @@ func newOutput(params output.Params) (*Output, error) {
 		ignoredTags:  ignoredTags,
 		row:          make([]string, 3+len(resTags)+1),
 		saveInterval: saveInterval,
-		timeFormat:   timeFormat,
+		timeFormat:   config.TimeFormat,
 		logger:       logger,
 		params:       params,
 	}
@@ -210,9 +210,9 @@ func SampleToRow(sample *metrics.Sample, resTags []string, ignoredTags []string,
 	row[0] = sample.Metric.Name
 
 	if timeFormat == RFC3399 {
-		row[1] = fmt.Sprint(sample.Time.Format(time.RFC3339))
+		row[1] = sample.Time.Format(time.RFC3339)
 	} else {
-		row[1] = fmt.Sprintf("%d", sample.Time.Unix())
+		row[1] = strconv.FormatInt(sample.Time.Unix(), 10)
 	}
 
 	row[2] = fmt.Sprintf("%f", sample.Value)
--- a/output/csv/config.go
+++ b/output/csv/config.go
@@ -38,6 +38,7 @@ type Config struct {
 	// Samples.
 	FileName     null.String        `json:"file_name" envconfig:"K6_CSV_FILENAME"`
 	SaveInterval types.NullDuration `json:"save_interval" envconfig:"K6_CSV_SAVE_INTERVAL"`
+	TimeFormat   TimeFormat         `json:"time_format" envconfig:"K6_CSV_TIME_FORMAT"`
 }
 
 // NewConfig creates a new Config instance with default values for some fields.
@@ -45,6 +46,7 @@ func NewConfig() Config {
 	return Config{
 		FileName:     null.StringFrom("file.csv"),
 		SaveInterval: types.NullDurationFrom(1 * time.Second),
+		TimeFormat:   Unix,
 	}
 }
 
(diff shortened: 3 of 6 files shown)


na-- (changes requested):
Ah, that's a good catch and it probably should have been the default. Unfortunately, now that the `csv` output has been released for quite a few versions already, we can't simply make a breaking change like this... :disappointed: It will probably disrupt the workflows of everyone that uses it.

Please add a new option to the output that controls this behavior, with 1-second Unix timestamps being the default and microsecond ones having to be explicitly enabled.

codebien:
>Please add a new option to the output that controls this behavior, with 1-second Unix timestamps being the default and microsecond ones having to be explicitly enabled.

@na-- Considering that Go has Unix{Milli | Micro | Nano} I guess we can expand the set of accepted values?

na--:
milliseconds and microseconds sure, nanoseconds are probably not very useful :man_shrugging:

codebien (commented):
I was thinking if it could be useful to create a type `Precision` that should wrap this concept, Duration doesn't cover it very well and defining a new set of strings anytime could be not the best, it could be added during #883 implementation.

yorugac:
Slightly connected to the topic of configurable precision. Values of all builtin durations are all scaled up to milliseconds right now:
https://github.com/grafana/k6/blob/ee8d9bdf1fc8cd1ac600cb3304dc7a7062d770ca/stats/units.go#L27-L33
IIRC, summary relies on floats to scale down to microseconds, i.e. 0.5ms = 500µs, etc.
Similarly, custom metrics seem to assume milliseconds as values by default (when `isTime = true`). I guess for most cases it is fine as is but if there'll be a new option, it might make sense to extend it to values too.

rpocklin:
Hi,

From what i'm hearing it would be good to be able to specify a time granularity option (eg. `--granularity msec | milli | micro | nano`) which changes the timestamp granularity consistently (eg. in summary report, CSV and JSON outputs).

This is a bit beyond this current PR and my current Go expertise but would really like to see this feature if possible.

I am open to working on this if you can show me a good place for the CLI options and how to pass them into the output parsers.

na--:
I'm sorry if I wasn't clear, but I don't think we should touch the global options for this change. Instead, this should be just an option for the CSV output, and only this file needs to be changed to add it: https://github.com/grafana/k6/blob/master/output/csv/config.go

rpocklin:
@na-- I have updated the PR making CSV output flag **useISO8601** (which is opt-in) and sets the time granularity to micro seconds (same as JSON formatter).  This does not affect any metric values mentioned by @yorugac but that could be done separately as a separate PR.

I have included unit tests, also to use this feature simply run k6 with the following command-line argument:
`--out csv=fileName=results.csv,useISO8601=true`

The default behaviour is unchanged as requested, so it will be backwards compatible.

rpocklin:
> I'm sorry if I wasn't clear, but I don't think we should touch the global options for this change. Instead, this should be just an option for the CSV output, and only this file needs to be changed to add it: https://github.com/grafana/k6/blob/master/output/csv/config.go

@na-- I have updated the PR making CSV output flag useISO8601 (which is opt-in) and sets the time granularity to micro seconds (same as JSON formatter). This does not affect any metric values mentioned by @yorugac but that could be done separately as a separate PR.

I have included unit tests, also to use this feature simply run k6 with the following command-line argument:
--out csv=fileName=results.csv,useISO8601=true

The default behaviour is unchanged as requested, so it will be backwards compatible.

MStoykov:
My two cents: It might be worth taking into consideration that iso8601 timestamps are [a lot](https://github.com/grafana/k6/pull/2436#issuecomment-1068822372) more cpu intensive to marshal than pure integers. 

I would argue that given that in most cases those files will be read by computers not humans, the performance likely should take precedence and we should just be dumping ms/ns as integers instead of anything else.

na--:
:+1: at the very least, these should be separate options :thinking: One can be called `timeUnit` or `granularity` or something like that, with `time.Duration` values, the default being `1s`. Then you can have a separate `timeFormat` option with values like `timestamp` (the default), `iso8601`, etc.

rpocklin:
> My two cents: It might be worth taking into consideration that iso8601 timestamps are [a lot](https://github.com/grafana/k6/pull/2436#issuecomment-1068822372) more cpu intensive to marshal than pure integers.
> 
> I would argue that given that in most cases those files will be read by computers not humans, the performance likely should take precedence and we should just be dumping ms/ns as integers instead of anything else.

The original goal of this PR was to allow these result logs to be analysed under second-level granularity, this PR (when running k6 with the new flag) simply aligns the CSV time output format with JSON, which uses ISO 8601 time format by default and micro-second time granularity.

rpocklin:
> From what i'm hearing it would be good to be able to specify a time granularity option (eg. `--granularity msec | milli | micro | nano`) which changes the timestamp granularity consistently (eg. in summary report, CSV and JSON outputs).

> 👍 at the very least, these should be separate options 🤔 One can be called `timeUnit` or `granularity` or something like that, with `time.Duration` values, the default being `1s`. Then you can have a separate `timeFormat` option with values like `timestamp` (the default), `iso8601`, etc.

The main goal of this PR is to align the output formats between CSV and JSON, in both time granularity and date format.  They could be separated for more customisation and happy to do another PR for that in the future but right now i'm looking at a sensible step forward to solve our immediate issue in parsing these logs consistently between CSV and JSON file formats and visualising in graphs - second-level time granularity isn't acceptable.

At the end of the day the flag is opt-in (as you requested it to be a non-breaking change) and this makes it consistent with the JSON output format (which is what anyone using this library would expect, and a good step forward for the library).

I'd suggest any flag for time precision or format should be considered as a global option so you don't have to implement different flags across the output writers, which you recommended against earlier.

na--:
Sorry for the delay in responding, @rpocklin. I understand and mostly agree with your desire to unify the CSV and JSON outputs, it makes a lot of sense, but I disagree with the way you propose to go about doing it.

I'd still prefer to add 2 separate options to the csv output:
* `timePrecision` - `1s` (current value and future default), `1ms`, etc.
* `timeFormat` - `timestamp` (current value and future default), `iso8601`, etc.

I am fine with merging a PR that adds both of these options, as well as merging a PR that only adds one of them. However, I don't think we should merge a PR for a single new option that conflates the two. That will be harder to maintain and reason about in the long term.

rpocklin:
> Sorry for the delay in responding, @rpocklin. I understand and mostly agree with your desire to unify the CSV and JSON outputs, it makes a lot of sense, but I disagree with the way you propose to go about doing it.
> 
> I'd still prefer to add 2 separate options to the csv output:
> 
> * `timePrecision` - `1s` (current value and future default), `1ms`, etc.
> * `timeFormat` - `timestamp` (current value and future default), `iso8601`, etc.
> 
> I am fine with merging a PR that adds both of these options, as well as merging a PR that only adds one of them. However, I don't think we should merge a PR for a single new option that conflates the two. That will be harder to maintain and reason about in the long term.

Thanks @na-- I have re-worked this PR based on this approach, such that now there is a CSV output string option called `timeFormat = [ 'unix' | 'rfc3399' ]` and I have updated the tests as well.

I'd appreciate feedback on the custom enum type `TimeFormat` (found in [https://github.com/rpocklin/k6/blob/improve-time-unit-granularity-for-csv-output/output/csv/consts.go](here)).

Once compiled, you can run with this flag set as follows:
```
./k6 --out csv=fileName=results.csv,timeFormat=rfc3399 script.js
```

As requested, if you run it without `timeFormat` defined, or with it set to `timeFormat=unix` you will get the default (current) behaviour.

output/csv/output.go
```diff
@@ -200,9 +204,17 @@ func MakeHeader(tags []string) []string {
...
 // SampleToRow converts sample into array of strings
-func SampleToRow(sample *metrics.Sample, resTags []string, ignoredTags []string, row []string) []string {
+func SampleToRow(sample *metrics.Sample, resTags []string, ignoredTags []string, row []string,
+	timeFormat TimeFormat,
+) []string {
 	row[0] = sample.Metric.Name
-	row[1] = fmt.Sprintf("%d", sample.Time.Unix())
+
+	if timeFormat == RFC3399 {
+		row[1] = fmt.Sprint(sample.Time.Format(time.RFC3339))
```
codebien, line 213:
```suggestion
		row[1] = sample.Time.Format(time.RFC3339)
```
Format returns already a string

output/csv/output.go
```diff
@@ -200,9 +204,17 @@ func MakeHeader(tags []string) []string {
...
+	} else {
+		row[1] = fmt.Sprintf("%d", sample.Time.Unix())
```
codebien, line 215:
```suggestion
		row[1] = strconv.Itoa(sample.Time.Unix())
```

codebien (changes requested):
Hi @rpocklin,
sorry for the late response. Thanks for your effort! :pray: 

It seems you have pushed two unexpected files (`go-bin-rpm-amd64.deb` and `1`), can you remove them, please?

codebien:
Hi @rpocklin, we would like to get your contribution merged during the current cycle so we will have it in the next release. Do you plan to address the request changes? We will do our best to support you as fast as we can.

dub-rp:
Yes, thanks for the nudge, I will finalise it in the next few days.

rpocklin:
> Hi @rpocklin, we would like to get your contribution merged during the current cycle so we will have it in the next release. Do you plan to address the request changes? We will do our best to support you as fast as we can.

Thanks for the nudge, i'll get this done in the next day or two :)

output/csv/output.go
rpocklin: I tried this and the compiler complains `Itoa` expects `int` whereas `sample.Time.Unix()` is `int64`
```
cannot use sample.Time.Unix() (value of type int64) as int value in argument to strconv.Itoacompiler[IncompatibleAssign](https://pkg.go.dev/golang.org/x/tools/internal/typesinternal?utm_source%3Dgopls#IncompatibleAssign)
```

output/csv/output.go
rpocklin: Thanks, will make this change.

rpocklin:
> Hi @rpocklin, sorry for the late response. Thanks for your effort! 🙏
> 
> It seems you have pushed two unexpected files (`go-bin-rpm-amd64.deb` and `1`), can you remove them, please?

Yep will remove these!

output/csv/consts.go
```diff
@@ -0,0 +1,40 @@
...
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
```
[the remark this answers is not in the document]
rpocklin: Will remove thanks.

output/csv/output.go
```diff
@@ -200,9 +204,17 @@ func MakeHeader(tags []string) []string {
...
+	}
```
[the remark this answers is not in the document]
rpocklin: I'm open to a suggestion, but keep in mind these both take different variables in the method arguments.

output/csv/consts.go
```diff
@@ -0,0 +1,40 @@
...
+)
+
+// IsValid validates TimeFormat
+func (timeFormat TimeFormat) IsValid() bool {
+	switch timeFormat {
+	case Unix, RFC3399:
+		return true
+	}
+	return false
+}
```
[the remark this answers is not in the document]
rpocklin: This seemed to be idiomatic for Go, and allows for a clean extension of `TimeFormat` types in the future.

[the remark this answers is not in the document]
codebien: If a timeFormat exists it must be valid, for that I prefer to maintain this logic on the parsing side so we can keep it centralized. @mstoykov opinion?

I think we should move this check in the output constructor so we can check the consolidated value so we can validate also values from json and env.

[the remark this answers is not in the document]
rpocklin: If you can make a suggestion that might help illustrate what you mean?

output/csv/output.go
```diff
@@ -97,6 +100,7 @@ func newOutput(params output.Params) (*Output, error) {
 			csvWriter:    stdoutWriter,
 			row:          make([]string, 3+len(resTags)+1),
 			saveInterval: saveInterval,
+			timeFormat:   timeFormat,
```
codebien, line 103:
```suggestion
			timeFormat: config.TimeFormat,
```
It seems we can avoid allocating the `timeFormat` variable

output/csv/output.go
```diff
@@ -87,6 +89,7 @@ func newOutput(params output.Params) (*Output, error) {
 
 	saveInterval := config.SaveInterval.TimeDuration()
 	fname := config.FileName.String
+	timeFormat := config.TimeFormat
```
codebien, line 92:
```suggestion
	if config.TimeFormat != Unix && config.TimeFormat != RFC3339 {
	    return fmt.Errorf("the timeFormat provided is not supported")
	}
	
```

output/csv/consts.go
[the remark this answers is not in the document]
codebien: https://github.com/grafana/k6/pull/2274#discussion_r885497705

output/csv/output.go
dub-rp: Won't this continue to grow into a long and unmaintainable `if` statement as we add more time formats?  

This was the reason I made an `isValid()` method which isolates the validity test from the parser itself, and keeps the validation and type definitions within the same file `output/csv/consts.go`.  It's also easier to unit test as-is.

output/csv/output.go
rpocklin: Yep, nice tip will remove.

output/csv/output.go
rpocklin: Won't this continue to grow into a long and unmaintainable if statement as we add more time formats?

This was the reason I made an isValid() method which isolates the validity test from the parser itself, and keeps the validation and type definitions within the same file output/csv/consts.go. It's also easier to unit test as-is.

output/csv/output.go
```diff
@@ -114,6 +117,7 @@ func newOutput(params output.Params) (*Output, error) {
 		ignoredTags:  ignoredTags,
 		row:          make([]string, 3+len(resTags)+1),
 		saveInterval: saveInterval,
+		timeFormat:   timeFormat,
```
mstoykov, line 120: This lines gives an error on compilation:

> output/csv/output.go:120:17: undefined: timeFormat

I guess it should be 
```suggestion
		timeFormat:   config.TimeFormat,
```

but I also notice that github actions isn't running tests which is more worrying. 

Can you rebase on top of the current master and I hope that this will fix it :crossed_fingers:

output/csv/consts.go
```diff
@@ -0,0 +1,20 @@
+// Package csv custom enum types used in config
```
mstoykov, line 1: This is comment for the whole *package*, but you are basically documenting the file. 

You likely did this as you were told by the linter, but it will be better if you add as separate `doc.go` file and just add some very basic doc such has `Package csv implements an output writing metrics in csv format`

output/csv/consts.go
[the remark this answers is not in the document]
mstoykov: For enums in general we use [enumer](https://github.com/alvaroloes/enumer) as for example with 
https://github.com/grafana/k6/blob/1679c6843c7a5578b25aaba228d98476926467c2/lib/runtime_options.go#L32-L33 

Which generates https://github.com/grafana/k6/blob/1679c6843c7a5578b25aaba228d98476926467c2/lib/compatibility_mode_gen.go and is then used 
https://github.com/grafana/k6/blob/1679c6843c7a5578b25aaba228d98476926467c2/lib/runtime_options.go#L71-L78

mstoykov (changes requested):

output/csv/consts.go
rpocklin: Thanks for the suggestion, done.

output/csv/consts.go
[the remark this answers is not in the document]
rpocklin: > For enums in general we use [enumer](https://github.com/alvaroloes/enumer) as for example with
> 
> https://github.com/grafana/k6/blob/1679c6843c7a5578b25aaba228d98476926467c2/lib/runtime_options.go#L32-L33
> 
> Which generates https://github.com/grafana/k6/blob/1679c6843c7a5578b25aaba228d98476926467c2/lib/compatibility_mode_gen.go and is then used
> 
> https://github.com/grafana/k6/blob/1679c6843c7a5578b25aaba228d98476926467c2/lib/runtime_options.go#L71-L78

@mstoykov I spent a long time learning `enumer` and how it works for Go.  I ended up with the follow error:
`enumer: can't handle non-integer constant type TimeFormat`

From what I can understand, this would only work on integer-based enums, whereas in this case we need to preserve the string literals as they are passed in and compared from the user (via the command line).

Let me know if this is still possible with your suggestion, otherwise i'm at a loss how to implement what you are suggesting.

mstoykov:
Just a note here that I did some fixes in https://github.com/grafana/k6/pull/2560 and noticed that we haven't noticed that the rfc is with *three* `3` not with two and two `9`. 

This has been fixed, but if you are using it locally @rpocklin, just a heads-up that the option is also named correctly now.

Thanks again for the working on this :bow:

Issue #2276, opened 2021-12-01
What about VUs = 0 ?


msaf1980:
I need to testing several scenarios: for example short and long duration queries

Number of VU's configured via env vars.

I need to disable of some scenario execution. But if set to vus to 0, get error `the number of VUs should be more than 0`

What about to allow of vus = 0 with simple disable scenario execution ?

```
let USERS_1H = getenv.getInt(`${__ENV.USERS_1H}`, 10);
let USERS_1D = getenv.getInt(`${__ENV.USERS_1D}`, 0);

export const options = {
    scenarios: {
        render_1h_offset_0: {
            executor: 'constant-vus',
            exec: 'api_render',
            vus: USERS_1H,
            duration: DURATION,
            gracefulStop: '10s',
            env: { GROUP: 'render_1h' },
        },
        render_1h_offset_7d: {
            executor: 'constant-vus',
            exec: 'api_render',
            vus: USERS_1D,
            duration: DURATION,
            gracefulStop: '10s',
            env: { GROUP: 'render_1d' },
        },
}
```

MStoykov:
@msaf1980, thanks for the issue and PR 

Unfortunately, IMO the check here is correct - the idea behind it is that it's unlikely that what you wanted is to make a configuration for a `constant-vus` and have it have *no* vus (negative ones are even a bigger red flag). Disabling a scenario in this way seems like an interesting approach but I would argue that ultimately it will be more confusing for anybody reading the script. I would propose that you just add or remove parts of the scenario depending on the configuration you provide:

```javascript
let USERS_1H = getenv.getInt(`${__ENV.USERS_1H}`, 10);
let USERS_1D = getenv.getInt(`${__ENV.USERS_1D}`, 0);
let scenarios = {};
if (USERS_1H > 0) {
    scenarios["render_1h_offset_0"] = {
            executor: 'constant-vus',
            exec: 'api_render',
            vus: USERS_1H,
            duration: DURATION,
            gracefulStop: '10s',
            env: { GROUP: 'render_1h' },
      }
}
if (USERS_1D > 0) {
     scenarios["render_1h_offset_7d"] = {
            executor: 'constant-vus',
            exec: 'api_render',
            vus: USERS_1D,
            duration: DURATION,
            gracefulStop: '10s',
            env: { GROUP: 'render_1d' },
        }
}
export const options = {
    scenarios: scenarios
}
``` 
(the code was written right here, so expect a type somewhere - it's for illustrative purposes ;) ).

[...]

PR #2277, opened 2021-12-01
allow to set vus = 0

msaf1980:
Simple PR for https://github.com/grafana/k6/issues/2276


MStoykov:
I am closing this with the same arguments as #2276  , but even if we wanted to support `vus=0` to mean disable an executor, I don't see any reason for this to support negative numbers as well

Issue #1048, opened 2019-06-12
Refactor/Rethink loading/compiling/transpiling of scripts specifically from archives


MStoykov:
During the creation of #1040 it turned out that the way we load, (transpile), compile and possibly execute script files is not ideal (read buggy and unintuitive). 

As mentioned in the PR if you run a script directly (not from an archive) you will first load, (transpile), compile and execute the main script and than all it's dependencies as they are required. This all happens before we actually start any VU to actually load test. This is mostly for the purposes of evaluating the `options` (as well as erroring out early when for example the "default function" is not defined). 
This also happens when you run `k6 archive script.js` in order for all required files to be imported in the archive. 
It is important that we actually execute the script as .. there is no other reliable way to find what files will actually be `open()`ed and `require`d. 
Once you run an archive though, previous to #1040, we won't load, compile/transpile and execute each and every script before we actually start running VUs. K6 just starts running VUs and they do that as they are required, which previous to 3506ee1, would've (IMO) wrongly shared the source of between them in a shared map, the only reason for this to not be a race condition is that currently VUs are initialized sequentially so no two VUs will try to load their scripts while another hasn't finished loading. 

The current fix was to just make it behave the same way it behaves when ran locally. But there is no real reason for this as during `k6 archive` all the relevant checks are ran and `options` is evaluated and saved in the archive.

We could've possibly just compiled everything that is script in the archive, but this need testing. 

I would much rather prefer if the whole [bundle](https://github.com/loadimpact/k6/blob/607b63c4400c8b6852c7f51a0a6151f0bb8ec6b7/js/bundle.go) is rewritten(or practically the whole code) in such a way that we thread safely load, (transpile) and compile scripts as they are required, in such a way that multiple VUs can be initialized at the same time. This is probably not going to be that hard but will require some changes possibly rippling through most of the `js` module itself and arguably the [archive](https://github.com/loadimpact/k6/blob/607b63c4400c8b6852c7f51a0a6151f0bb8ec6b7/lib/archive.go). 

Maybe sometime in the future we can skip the first running of the scripts not in a VU and just initialize one VU and than use the options from it for all other ...

yorugac:
This issue is a bit outdated ATM, but I've looked briefly into possibility of this refactoring as part of work in #2279 and here are some initial considerations. They are likely incomplete as they were done from perspective of totally unrelated issue and without up-to-date info on the current goals here but should be useful for future reference.

[...]

PR #2279, opened 2021-12-01
Extend k6 inspect command with additional flag

yorugac:
This PR adds additional output to `k6 inspect` with `--execution-requirements` CLI option. When switched on, it triggers calculation of execution requirements, substitutes config with the derived one and adds `maxVUs` and `totalDuration` fields to the output.

Signature of `getConsolidatedConfig` was modified: now it doesn't depend on `lib.Runner` but only on `lib.Options`.

Some minor refactoring in `cmd/` was added on the way, but not too much as that'd require additional work and likely a separate PR.

This is part of the work for implementing https://github.com/grafana/k6-operator/issues/9

### DIFF
--- a/cmd/inspect.go
+++ b/cmd/inspect.go
@@ -27,15 +27,19 @@ import (
 	"os"
 
 	"github.com/sirupsen/logrus"
+	"github.com/spf13/afero"
 	"github.com/spf13/cobra"
 
+	"go.k6.io/k6/core/local"
 	"go.k6.io/k6/js"
 	"go.k6.io/k6/lib"
 	"go.k6.io/k6/lib/metrics"
-	"go.k6.io/k6/loader"
+	"go.k6.io/k6/lib/types"
 )
 
-func getInspectCmd(logger logrus.FieldLogger) *cobra.Command {
+func getInspectCmd(logger *logrus.Logger) *cobra.Command {
+	var addExecReqs bool
+
 	// inspectCmd represents the inspect command
 	inspectCmd := &cobra.Command{
 		Use:   "inspect [file]",
@@ -43,19 +47,9 @@ func getInspectCmd(logger logrus.FieldLogger) *cobra.Command {
 		Long:  `Inspect a script or archive.`,
 		Args:  cobra.ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
-			pwd, err := os.Getwd()
+			src, filesystems, err := readSource(args[0], logger)
 			if err != nil {
 				return err
-			}
-			filesystems := loader.CreateFilesystems()
-			src, err := loader.ReadSource(logger, args[0], pwd, filesystems, os.Stdin)
-			if err != nil {
-				return err
-			}
-
-			typ := runType
-			if typ == "" {
-				typ = detectType(src.Data)
 			}
 
 			runtimeOptions, err := getRuntimeOptions(cmd.Flags(), buildEnvMap(os.Environ()))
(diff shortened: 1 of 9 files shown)


cmd/inspect.go
```diff
@@ -27,80 +27,123 @@ import (
...
+			}
+			if err != nil {
+				return err
+			}
+
+			// ATM, output can take 2 forms: standard (equal to lib.Options struct) and extended, with additional fields.
+			inspectOutput := interface{}(b.Options)
+
+			if addExecReqs {
+				inspectOutput, err = addExecRequirements(b, builtinMetrics, registry, logger)
```
codebien, line 84: Maybe I'm missing something but could we reuse the `newRunner` function from `cmd/run.go` and inject the runner here? 

https://github.com/grafana/k6/blob/5c14dae129a5eb6ed309c94a194370878b7b08e5/cmd/run.go#L390

cmd/inspect.go
```diff
@@ -27,80 +27,123 @@ import (
...
 				if err != nil {
 					return err
```
codebien, line 86: A returned error at this point, would be clear enough or should we add a small detail wrapping it?

codebien (changes requested):
I added some comments, if there is an open issue/PR from the operator where this should be used can you link it, please? It would be helpful to see the global picture :pray:

yorugac:
Thanks @codebien! I've added relevant link to the description

cmd/inspect.go
yorugac: `k6 run` returns those errors as they are so this is a ~ copy-paste of that behavior :thinking: Clearly, it worked so far :slightly_smiling_face: I think if someone encounters one of these errors in practice and complains about it, it'd be another matter...

cmd/inspect.go
```diff
@@ -27,80 +27,123 @@ import (
...
+	builtinMetrics *metrics.BuiltinMetrics, registry *metrics.Registry,
+	logger *logrus.Logger) (interface{}, error) {
+	runner, err := js.NewFromBundle(logger, b, builtinMetrics, registry)
+	if err != nil {
+		return nil, err
+	}
+
+	conf, err := getConsolidatedConfig(afero.NewOsFs(), Config{}, runner)
+	if err != nil {
+		return nil, err
```
MStoykov, line 121: Bundle is actually a type that I and @na-- have mostly wanted to remove in the past, so exposing it more seems like a bad idea. Refactoring everything needed is probably going to be way too much work, but I would argue you that given that you already got the Bundle, and that it actually *has* the options and that `getConsolidatedConfig` just gets the Options from the runner, which gets it from the Bundle inside .... You can just rewrite getConsolidatedConfig to be taking the Options and get the Options directly from the runner, skipping making `NewFromBundle` public.  

Even more importantly the one thing that `get

yorugac: I suppose you meant this https://github.com/grafana/k6/issues/1048 ?

MStoykov: I think that issue (as most very old issues) is now very outdated, but in parts - yes.

yorugac: @MStoykov  I've looked into this: trying to refactor in "minimal fashion" (without touching `js/` stuff) gets stuck on the point of `Runner.IsExecutable` → which needs `Bundle.exports` so again, the same problem with private field of `Bundle`. Trying to refactor this at the level of `Bundle` is definitely too out-of-scope for this PR: I think it should be done separately. 
For now, I'll add some notes on that to the #1048 issue: it can be continued in a separate PR or be left alone until next prioritization.

I added a small refactoring around `getConsolidatedConfig`: it can truly be used with only options, without `Runner`: I think it's relatively small to be present in this PR and might simplify next stages of refactorings too.

MStoykov: :( 

Hm looking at this again, maybe just instead of using NewBundle and then NewFromBundle, you can just go with creating the runner the same `run` will do. The additional work `NewFromBundle` does is ... minimal and should not have any side effects. But will make the cmd<->js dependency smaller. 

WDYT ?

yorugac: I wouldn't say dependency would be "smaller".. meaning, even prior to this PR `inspect` was the only one in cmd using `Bundle` directly which makes sense in its case. From this follows that probably any major refactoring in `js` will impact `inspect` first.

Of course, it can be re-written as you describe but TBH, I think all options here are equally bad :joy: -- because that major refactoring is needed. I slightly prefer the current version because I don't see any harm from public `NewFromBundle` at this point while this version more explicitly shows the use cases that future refactoring should take into account. So it's kind of a way to document too.

MStoykov (approved):

codebien (approved):
LGTM :+1:

codebien (approved):

MStoykov (approved):

PR #2280, opened 2021-12-02
gomod: update the file format to v1.17

codebien:
`go mod tidy -go=1.17` migration

The command creates only this block and doesn't re-arrange the already existing indirect directives.

```go
require (
       github.com/andybalholm/cascadia v1.1.0 // indirect
       github.com/davecgh/go-spew v1.1.1 // indirect
       github.com/dlclark/regexp2 v1.4.1-0.20201116162257-a2a8dda75c91 // indirect
       github.com/go-sourcemap/sourcemap v2.1.3+incompatible // indirect
       github.com/inconshreveable/mousetrap v1.0.0 // indirect
       github.com/josharian/intern v1.0.0 // indirect
       github.com/onsi/gomega v1.10.1 // indirect
       github.com/tidwall/match v1.1.1 // indirect
)
```
so I manually deleted the indirect rows and re-executed `go mod tidy` to get all of them in the 2nd block.


MStoykov:
This changes dependency versions. 

I don't think this is a good idea, especially as this downgrades some stuff. We currently have some versions which work for us and we have some empirical evidence that they work okay, and downgrading them doesn't do anything for us AFAIK. 

Given this, I would recommend copying the `//indirect` dependency versions form the top `require` block to the second one so we don't change any versions ;).

Also ... why are we doing this?

codebien:
>This changes dependency versions.

Yep, I maintained them to discuss if we have to drop all or some could be interesting for us.

>Also ... why are we doing this?

This is the new `gomod` file format so I expect we will have to migrate at some point and it helps to main stuff clear/easy to understand. IMHO.

MStoykov (changes requested):
I am not against moving to the new format, but definitely don't change any dependency versions while doing so

go.sum
```diff
@@ -108,9 +108,8 @@ github.com/google/go-cmp v0.2.0/go.mod h1:oXzfMopK8JAjlY9xF4vHSVASa0yLyX7SntLO5a
 github.com/google/go-cmp v0.3.0/go.mod h1:8QqcDgzrUqlUb/G2PQTWiueGozuR1884gddMywk6iLU=
 github.com/google/go-cmp v0.3.1/go.mod h1:8QqcDgzrUqlUb/G2PQTWiueGozuR1884gddMywk6iLU=
 github.com/google/go-cmp v0.4.0/go.mod h1:v8dTdLbMG2kIc/vJvl+f65V22dbkXbowE6jgT/gNBxE=
+github.com/google/go-cmp v0.5.0 h1:/QaMHBdZ26BB3SSst0Iwl10Epc+xhTquomWX0oZEB6w=
```
MStoykov, line 111: this is a change - it seems to have no implications but still I would prefer if we specify v0.5.1 it seems to have some changes to it, so I have no idea why the `vendor` folder doesn't change more :shrug:

MStoykov (changes requested):
This looks a lot better I still want go-cmp to be the exact same version ;)

go.sum
codebien: `go mod tidy` removes it. As you can note the `go-cmp` has been removed from the indirect list by `tidy`. The `vendor` folder doesn't contain indirect packages.

yorugac: v0.5.1 seems like a small diff from v0.5.0... This is weird but, where did v0.5.1 originally come from? It seems like protobuf v1.4.3 specifies go-cmp v0.4.0 ([src](https://github.com/golang/protobuf/blob/v1.4.3/go.mod)) while grpc v1.41.0 specifies go-cmp v0.5.0 ([src](https://github.com/grpc/grpc-go/blob/v1.41.0/go.mod)) :confused: Was there anything else importing go-cmp with v0.5.1?

codebien: I guess the unique way I can explain it is that we add it manually for any reason. I'm thinking if it isn't an attempt to overwrite the `go-cmp` version with the fixes from `0.5.1`.

yorugac: Looking at go.sum:
```
9d9cea457 (Roger Chapman    2020-10-21 19:42:32 +1100 107) github.com/google/go-cmp v0.2.0/go.mod h1:oXzfMopK8JAjlY9xF4vHSVASa0yLyX7SntLO5aqRK0M=
9d9cea457 (Roger Chapman    2020-10-21 19:42:32 +1100 108) github.com/google/go-cmp v0.3.0/go.mod h1:8QqcDgzrUqlUb/G2PQTWiueGozuR1884gddMywk6iLU=
9d9cea457 (Roger Chapman    2020-10-21 19:42:32 +1100 109) github.com/google/go-cmp v0.3.1/go.mod h1:8QqcDgzrUqlUb/G2PQTWiueGozuR1884gddMywk6iLU=
9d9cea457 (Roger Chapman    2020-10-21 19:42:32 +1100 110) github.com/google/go-cmp v0.4.0/go.mod h1:v8dTdLbMG2kIc/vJvl+f65V22dbkXbowE6jgT/gNBxE=
8e9367331 (Mihail Stoykov   2021-03-29 18:48:29 +0300 111) github.com/google/go-cmp v0.5.0/go.mod h1:v8dTdLbMG2kIc/vJvl+f65V22dbkXbowE6jgT/gNBxE=
83fcc2303 (Ivan Mirić       2020-08-14 10:32:54 +0200 112) github.com/google/go-cmp v0.5.1 h1:JFrFEBb2xKufg6XkJsJr+WbKb4FQlURi5RUcBveYu9k=
83fcc2303 (Ivan Mirić       2020-08-14 10:32:54 +0200 113) github.com/google/go-cmp v0.5.1/go.mod h1:v8dTdLbMG2kIc/vJvl+f65V22dbkXbowE6jgT/gNBxE=
```
The dates are great here :smile: But it appears that the oldest commit, by Ivan, is actually the initial migration to modules. And Mihail's commit in March was update of grpc to v1.36.1. v0.5.1 looks now like a remnant of the migration...
If it doesn't break anything then we're probably "safe" not to fight with `go mod tidy` result and downgrade go-cmp a bit. Alternatively, this PR can be postponed until next dependency update. WDYT?

MStoykov: It seems that this went up and down a couple of times including around moving kafka output out as an extension and dropping it. 
Also it seems like it isn't *at all* in the `vendor` directory so that solves the mystery I was having.

I also did try to do it manually and it *worked* for me and no amount of `go mod tidy`  and or `go mod vendor` reverted my manual changes :shrug:

MStoykov (approved):

na-- (approved):

PR #2281, opened 2021-12-02
statsd output: allow append to name some tags values

msaf1980:
Allow to append to statsd metric names values of some tag. May be used for separate statistic for group.

### DIFF
--- a/output/statsd/config.go
+++ b/output/statsd/config.go
@@ -38,6 +38,7 @@ type config struct {
 	Namespace    null.String        `json:"namespace,omitempty" envconfig:"K6_STATSD_NAMESPACE"`
 	PushInterval types.NullDuration `json:"pushInterval,omitempty" envconfig:"K6_STATSD_PUSH_INTERVAL"`
 	TagBlocklist stats.TagSet       `json:"tagBlocklist,omitempty" envconfig:"K6_STATSD_TAG_BLOCKLIST"`
+	TagAppend    []string           `json:"tagAppend,omitempty" envconfig:"K6_STATSD_TAG_APPEND"`
 	EnableTags   null.Bool          `json:"enableTags,omitempty" envconfig:"K6_STATSD_ENABLE_TAGS"`
 }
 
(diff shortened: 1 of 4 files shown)


codebien (commented):
Hi @msaf1980,
can you clarify the end goals and the use-cases about why you want to do it, please? Is something common supported in `statsd` ecosystem or from other software?

msaf1980:
> Hi @msaf1980, can you clarify the end goals and the use-cases about why you want to do it, please? Is something common supported in `statsd` ecosystem or from other software?

I have several group of queries. And need to separate metrics for them. Tags in DataDog format not useful for me (we used tags in different format with patched statsite).

For example. separation with label tag

https://github.com/msaf1980/xk6-carbonapi/blob/statsd/examples/carbonapi.js#L249

Resulting metric in StatsD will be in format (when k6 is running with env var ```K6_STATSD_TAG_APPEND='label'```)

```k6.graphite.staging.http_req_recv_bytes.render_1h_offset_0```

codebien:
Hi @msaf1980,
thanks for your contribution. :pray: 

Your use case sounds legit but we prefer to introduce features to [k6's StatsD output](https://k6.io/docs/results-visualization/statsd) that are widely adopted by different platforms that are implementing StatsD and our team is concerned that this case seems too specific for [statsite](https://github.com/statsite/statsite). The builtin k6 output for StatsD already supports tags and according to our current information, that seems the best experience we can provide to support most of the use cases. Additionally, [statsite](https://github.com/statsite/statsite) looks abandoned since 2019 and issues with the support for its format can be considered a legacy issue and not a common use case. Instead, it is likely a good idea to create a dedicated `xk6-output-statsite` extension for your use case: in other words, k6 team would appreciate it if you were to consider creating such an extension so that we can support such cases via extension and officially list it on the [Extensions](https://k6.io/docs/extensions) page.

We would also consider re-opening the topic / PR in the future if we encounter new similar use cases.

PR #2282, opened 2021-12-02
update iter/s precision from 0 to 2

m3hm3t:
This Pull Request addresses issue 1753.

If you set up a scenario where the executors run less than one iteration per second, the value is floored to zero while in progress.

By this PR, the behavior changes like at the below.

From:
image

To:
image

### DIFF
--- a/lib/executor/constant_arrival_rate.go
+++ b/lib/executor/constant_arrival_rate.go
@@ -249,7 +249,7 @@ func (car ConstantArrivalRate) Run(
 
 	vusFmt := pb.GetFixedLengthIntFormat(maxVUs)
 	progIters := fmt.Sprintf(
-		pb.GetFixedLengthFloatFormat(arrivalRatePerSec, 0)+" iters/s", arrivalRatePerSec)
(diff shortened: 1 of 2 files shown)


MStoykov:
Hi @m3hm3t thanks for the PR, unfortunately, this was already kind of tried in https://github.com/grafana/k6/pull/1766. 

And as I mention [there](https://github.com/grafana/k6/pull/1766#issuecomment-746019781) it looks really strange with `ramping-arrival-rate` and big timeUnit. I don't think we have agreed on my suggestion, but maybe you can experiment with it? @DarkAEther seems to have not had time (which is completely understandable) do to so, so it's fine if somebody else does :).

m3hm3t:
Thanks for the comment @MStoykov. I will look into it.

codebien (changes requested):
Setting the status as request changes

mstoykov:
After some internal discussion we have agreed that having this is probably better than nothing. Also some people have pointed out that changing the timeUnit shown will likely be more confusing than anything else.

mstoykov (approved):
LGTM! thanks for the PR and sorry for the really long delay :(

oleiade (approved):
👍🏻 🦕

PR #2284, opened 2021-12-03
Prevent goroutines getting blocked in js tests

MStoykov:
The predominant reason was the use of context.Background() for a context, which leads to a goroutine waiting on it ending, to block indefinitely.

The other case were gRPC tests which did *not* close their connections.

### DIFF
--- a/js/console_test.go
+++ b/js/console_test.go
@@ -48,7 +48,10 @@ func TestConsoleContext(t *testing.T) {
 	rt := goja.New()
 	rt.SetFieldNameMapper(common.FieldNameMapper{})
 
-	ctxPtr := new(context.Context)
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	ctxPtr := &ctx
+
 	logger, hook := logtest.NewNullLogger()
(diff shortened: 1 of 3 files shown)


MStoykov:
These were fined around https://github.com/grafana/k6/pull/2228#issuecomment-984663379 and were mostly caught by paniccing when `context.Done()` return nil at https://github.com/grafana/k6/blob/5c14dae129a5eb6ed309c94a194370878b7b08e5/js/runner.go#L652 - I am not really certain if it won't be good idea to check this in the production code as well :thinking:

na-- (approved):

olegbespalov (approved):

codebien (approved):
> - I am not really certain if it won't be good idea to check this in the production code as well thinking

If this would happen in production then I expect to see a lot of stuff stuck then it should be easy to catch, so I'm not sure. Considering the numbers of VUs we have, I see it relevant if there is a case where it could happen just for some VUs. But I don't see/know that case.

MStoykov:
@codebien I doubt we have a bug in the production code, I was wondering about guarding again the same regression in test code, but this also didn't catch some variants and no other test broke in the whole codebase so :shrug: 

Let's leave it for now

PR #2285, opened 2021-12-04
Added support for local files to `--log-output`

### LINKED ISSUE #2249: Add support for local files to `--log-output`
### Feature Description

[`--log-output`](https://k6.io/docs/using-k6/options/#log-output) currently supports only `none` (disabled), `stdout`, `stderr` and [`loki`](https://grafana.com/oss/loki/). 

Supporting local files should be simple and will be beneficial for debugging, or in CI environments, and it can also be used by log-forwarding tools to ship the logs pretty much anywhere. Combined with the [`logformat` ](https://k6.io/docs/using-k6/options/#logformat) option, this feature should be pretty versatile.

Somewhat connected issue: https://github.com/grafana/k6/issues/1939

alyakimenko:
Added file hook for logger.

Examples of usage:
```bash
k6 run --log-output file=./k6.log --logformat json ./examples/stages.js
```
With a defined log level:
```bash
k6 run --log-output file=./k6.log,level=info --logformat json ./examples/stages.js
```

Closes #2249

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -246,7 +246,7 @@ func (c *rootCommand) rootCmdPersistentFlagSet() *pflag.FlagSet {
 	flags.BoolVarP(&c.commandFlags.quiet, "quiet", "q", false, "disable progress updates")
 	flags.BoolVar(&c.commandFlags.noColor, "no-color", false, "disable colored output")
 	flags.StringVar(&c.logOutput, "log-output", "stderr",
-		"change the output for k6 logs, possible values are stderr,stdout,none,loki[=host:port]")
+		"change the output for k6 logs, possible values are stderr,stdout,none,loki[=host:port],file[=./path.fileformat]")
 	flags.StringVar(&c.logFmt, "logformat", "", "log output format") // TODO rename to log-format and warn on old usage
 	flags.StringVarP(&c.commandFlags.address, "address", "a", "localhost:6565", "address for the api server")
 
@@ -279,27 +279,37 @@ func (c *rootCommand) setupLoggers() (<-chan struct{}, error) {
 	}
 
 	loggerForceColors := false // disable color by default
-	switch c.logOutput {
-	case "stderr":
+	switch line := c.logOutput; {
+	case line == "stderr":
 		loggerForceColors = !c.commandFlags.noColor && c.commandFlags.stderrTTY
 		c.logger.SetOutput(c.commandFlags.stderr)
-	case "stdout":
+	case line == "stdout":
 		loggerForceColors = !c.commandFlags.noColor && c.commandFlags.stdoutTTY
 		c.logger.SetOutput(c.commandFlags.stdout)
-	case "none":
+	case line == "none":
 		c.logger.SetOutput(ioutil.Discard)
-	default:
-		if !strings.HasPrefix(c.logOutput, "loki") {
-			return nil, fmt.Errorf("unsupported log output `%s`", c.logOutput)
-		}
+
+	case strings.HasPrefix(line, "loki"):
 		ch = make(chan struct{})
-		hook, err := log.LokiFromConfigLine(c.ctx, c.fallbackLogger, c.logOutput, ch)
+		hook, err := log.LokiFromConfigLine(c.ctx, c.fallbackLogger, line, ch)
 		if err != nil {
 			return nil, err
 		}
 		c.logger.AddHook(hook)
 		c.logger.SetOutput(ioutil.Discard) // don't output to anywhere else
 		c.logFmt = "raw"
+
+	case strings.HasPrefix(line, "file"):
+		hook, err := log.FileHookFromConfigLine(c.ctx, c.fallbackLogger, line)
+		if err != nil {
+			return nil, err
+		}
+
+		c.logger.AddHook(hook)
+		c.logger.SetOutput(ioutil.Discard)
+
+	default:
+		return nil, fmt.Errorf("unsupported log output `%s`", line)
 	}
 
 	switch c.logFmt {
--- a/log/file.go
+++ b/log/file.go
@@ -1,0 +1,160 @@
+/*
+ *
+ * k6 - a next-generation load testing tool
+ * Copyright (C) 2020 Load Impact
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+// Package log implements various logrus hooks.
+package log
+
+import (
+	"bufio"
+	"context"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/sirupsen/logrus"
+)
+
+// fileHookBufferSize is a default size for the fileHook's loglines channel.
+const fileHookBufferSize = 100
+
+// fileHook is a hook to handle writing to local files.
+type fileHook struct {
+	fallbackLogger logrus.FieldLogger
+	loglines       chan []byte
+	path           string
+	w              io.WriteCloser
+	bw             *bufio.Writer
+	levels         []logrus.Level
+}
+
+// FileHookFromConfigLine returns new fileHook hook.
+func FileHookFromConfigLine(
+	ctx context.Context, fallbackLogger logrus.FieldLogger, line string,
+) (logrus.Hook, error) {
+	hook := &fileHook{
+		fallbackLogger: fallbackLogger,
+		levels:         logrus.AllLevels,
+	}
+
+	parts := strings.SplitN(line, "=", 2)
+	if parts[0] != "file" {
+		return nil, fmt.Errorf("logfile configuration should be in the form `file=path-to-local-file` but is `%s`", line)
+	}
+
+	if err := hook.parseArgs(line); err != nil {
+		return nil, err
+	}
+
+	if err := hook.openFile(); err != nil {
+		return nil, err
+	}
+
+	hook.loglines = hook.loop(ctx)
+
+	return hook, nil
+}
+
+func (h *fileHook) parseArgs(line string) error {
+	tokens, err := tokenize(line)
+	if err != nil {
+		return fmt.Errorf("error while parsing logfile configuration %w", err)
+	}
+
+	for _, token := range tokens {
+		switch token.key {
+		case "file":
+			if token.value == "" {
+				return fmt.Errorf("filepath must not be empty")
+			}
+			h.path = token.value
+		case "level":
+			h.levels, err = parseLevels(token.value)
+			if err != nil {
+				return err
+			}
+		default:
+			return fmt.Errorf("unknown logfile config key %s", token.key)
+		}
+	}
+
+	return nil
+}
+
+// openFile opens logfile and initializes writers.
+func (h *fileHook) openFile() error {
+	if _, err := os.Stat(filepath.Dir(h.path)); os.IsNotExist(err) {
+		return fmt.Errorf("provided directory '%s' does not exist", filepath.Dir(h.path))
+	}
+
+	file, err := os.OpenFile(h.path, os.O_WRONLY|os.O_APPEND|os.O_CREATE, 0o600)
+	if err != nil {
+		return fmt.Errorf("failed to open logfile %s: %w", h.path, err)
+	}
+
+	h.w = file
+	h.bw = bufio.NewWriter(file)
+
+	return nil
+}
+
+func (h *fileHook) loop(ctx context.Context) chan []byte {
+	loglines := make(chan []byte, fileHookBufferSize)
+
+	go func() {
+		for {
+			select {
+			case entry := <-loglines:
+				if _, err := h.bw.Write(entry); err != nil {
+					h.fallbackLogger.Errorf("failed to write a log message to a logfile: %w", err)
+				}
+			case <-ctx.Done():
+				if err := h.bw.Flush(); err != nil {
+					h.fallbackLogger.Errorf("failed to flush buffer: %w", err)
+				}
+
+				if err := h.w.Close(); err != nil {
+					h.fallbackLogger.Errorf("failed to close logfile: %w", err)
+				}
+
+				return
+			}
+		}
+	}()
+
+	return loglines
+}
+
+// Fire writes the log file to defined path.
+func (h *fileHook) Fire(entry *logrus.Entry) error {
+	message, err := entry.Bytes()
+	if err != nil {
+		return fmt.Errorf("failed to get a log entry bytes: %w", err)
+	}
+
+	h.loglines <- message
+	return nil
+}
+
+// Levels returns configured log levels.
+func (h *fileHook) Levels() []logrus.Level {
+	return h.levels
+}
--- a/log/file_test.go
+++ b/log/file_test.go
@@ -1,0 +1,168 @@
+/*
+ *
+ * k6 - a next-generation load testing tool
+ * Copyright (C) 2020 Load Impact
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+package log
+
+import (
+	"bufio"
+	"bytes"
+	"context"
+	"fmt"
+	"io"
+	"os"
+	"testing"
+	"time"
+
+	"github.com/sirupsen/logrus"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+type nopCloser struct {
+	io.Writer
+	closed chan struct{}
+}
+
+func (nc *nopCloser) Close() error {
+	nc.closed <- struct{}{}
+	return nil
+}
+
+func TestFileHookFromConfigLine(t *testing.T) {
+	t.Parallel()
+
+	tests := [...]struct {
+		line       string
+		err        bool
+		errMessage string
+		res        fileHook
+	}{
+		{
+			line: "file",
+			err:  true,
+			res: fileHook{
+				levels: logrus.AllLevels,
+			},
+		},
+		{
+			line: fmt.Sprintf("file=%s/k6.log,level=info", os.TempDir()),
+			err:  false,
+			res: fileHook{
+				path:   fmt.Sprintf("%s/k6.log", os.TempDir()),
+				levels: logrus.AllLevels[:5],
+			},
+		},
+		{
+			line: "file=./",
+			err:  true,
+		},
+		{
+			line: "file=/a/c/",
+			err:  true,
+		},
+		{
+			line:       "file=,level=info",
+			err:        true,
+			errMessage: "filepath must not be empty",
+		},
+		{
+			line: "file=/tmp/k6.log,level=tea",
+			err:  true,
+		},
+		{
+			line: "file=/tmp/k6.log,unknown",
+			err:  true,
+		},
+		{
+			line: "file=/tmp/k6.log,level=",
+			err:  true,
+		},
+		{
+			line: "file=/tmp/k6.log,level=,",
+			err:  true,
+		},
+		{
+			line:       "file=/tmp/k6.log,unknown=something",
+			err:        true,
+			errMessage: "unknown logfile config key unknown",
+		},
+		{
+			line:       "unknown=something",
+			err:        true,
+			errMessage: "logfile configuration should be in the form `file=path-to-local-file` but is `unknown=something`",
+		},
+	}
+
+	for _, test := range tests {
+		test := test
+		t.Run(test.line, func(t *testing.T) {
+			t.Parallel()
+
+			res, err := FileHookFromConfigLine(context.Background(), logrus.New(), test.line)
+
+			if test.err {
+				require.Error(t, err)
+
+				if test.errMessage != "" {
+					require.Equal(t, test.errMessage, err.Error())
+				}
+
+				return
+			}
+
+			require.NoError(t, err)
+			assert.NotNil(t, res.(*fileHook).w)
+		})
+	}
+}
+
+func TestFileHookFire(t *testing.T) {
+	t.Parallel()
+
+	var buffer bytes.Buffer
+	nc := &nopCloser{
+		Writer: &buffer,
+		closed: make(chan struct{}),
+	}
+
+	hook := &fileHook{
+		loglines: make(chan []byte),
+		w:        nc,
+		bw:       bufio.NewWriter(nc),
+		levels:   logrus.AllLevels,
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
+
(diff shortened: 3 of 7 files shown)


log/file.go
```diff
@@ -0,0 +1,145 @@
...
+func (h *fileHook) write(entry *logrus.Entry) (err error) {
+	dir := filepath.Dir(h.path)
+	if err = os.MkdirAll(dir, os.ModePerm); err != nil {
+		return fmt.Errorf("failed to create dir for logfile: %s %w", h.path, err)
+	}
+
+	file, err := os.OpenFile(h.path, os.O_WRONLY|os.O_APPEND|os.O_CREATE, 0o600)
+	if err != nil {
+		return fmt.Errorf("failed to open logfile: %s %w", h.path, err)
+	}
```
codebien, line 122: Doing in this way will be slow. `FileHookFromConfigLine` should create and open the file once, it should accept a context and listen to `Done` for closing the file. The lines should be flushed asynchronously mostly as the `loki` hook is already doing.

codebien (changes requested):
Hey @alyakimenko,
thanks for your contribution :pray: :tada: 

Some request changes in the comments. Also, did you try to use the logrus' [builtin writer](https://pkg.go.dev/github.com/sirupsen/logrus@v1.8.1/hooks/writer)?

log/file.go
alyakimenko: I completely agree with you. I'll refactor it.

alyakimenko: I made it more asynchronous. Was trying to keep it as simple as possible.

log/file.go
```diff
@@ -0,0 +1,166 @@
...
+
+// fileHook is a hook to handle writing to local files.
+type fileHook struct {
+	ctx            context.Context
+	fallbackLogger logrus.FieldLogger
+	ch             chan []byte
+	stopped        chan struct{}
+
+	path   string
+	file   *os.File
```
codebien, line 48: I would
```suggestion
	w   io.WriteCloser
```
It will allow us to create better-controlled test cases. e.g. we should add something like:

```go
buf := bytes.NewBuffer()
fileHook = {
w: buf,
}

f.Fire(entry)
cancel()
assert.Equal(t, "the content of the message", buf.String())
```

codebien (changes requested):
Thanks for these changes, this version looks very close.

log/tokenizer.go
```diff
@@ -36,7 +36,7 @@ type tokenizer struct {
 func (t *tokenizer) readKey() (string, error) {
 	start := t.i
 	for ; t.i < len(t.s); t.i++ {
-		if t.s[t.i] == '=' {
+		if t.s[t.i] == '=' && t.i != len(t.s)-1 {
```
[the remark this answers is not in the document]
alyakimenko: Added. There was an `index out of range` error, without the `t.i != len(t.s)-1` check. In that case when we pass something like `k6 run loki=` or `k6 run file=`

log/file.go
alyakimenko: Replaced with `io.WriteCloser` and added test

log/file_test.go
```diff
@@ -0,0 +1,170 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+type bytesBufferWithClose struct {
+	buffer *bytes.Buffer
+}
+
+func (bb *bytesBufferWithClose) Write(p []byte) (int, error) {
+	return bb.buffer.Write(p)
+}
```
codebien, line 41:
```suggestion
type nopCloser struct {
	Writer io.Writer
}

func (nc *nopCloser) Close() error {return nil}
```
The `Write` method is not required with an embedded struct. You can check the `NopCloser` method from the stdlib https://cs.opensource.google/go/go/+/refs/tags/go1.17.5:src/io/io.go;l=616-620

log/file_test.go
```diff
@@ -0,0 +1,170 @@
...
+
+func (bb *bytesBufferWithClose) Close() error {
+	return nil
+}
+
+func Test_FileHookFromConfigLine(t *testing.T) {
```
codebien, line 47:
```suggestion
func TestFileHookFromConfigLine(t *testing.T) {
```

log/file_test.go
```diff
@@ -0,0 +1,170 @@
...
+				}
+
+				return
+			}
+
+			require.NoError(t, err)
+			test.res.fallbackLogger = res.(*fileHook).fallbackLogger
+			test.res.ch = res.(*fileHook).ch
+			test.res.w = res.(*fileHook).w
+			require.Equal(t, &test.res, res)
```
codebien, line 134:
```suggestion
			assert.NotNil(t, res.(*fileHook).w)
```
I don't think we require to assert all the internals, the method it's mostly acting as a constructor so just check that the writer has been initialized and any error has been returned.

log/file_test.go
```diff
@@ -0,0 +1,170 @@
...
+			require.Equal(t, test.res.levels, res.(*fileHook).Levels())
+		})
+	}
+}
+
+func Test_fileHookFire(t *testing.T) {
```
codebien, line 140:
```suggestion
func TestFileHookFire(t *testing.T) {
```

log/file_test.go
```diff
@@ -0,0 +1,170 @@
...
+
+	logger := logrus.New()
+	logger.AddHook(hook)
+	logger.SetOutput(io.Discard)
+
+	logger.Info("example log line")
+
+	time.Sleep(time.Millisecond * 20)
+
+	require.Equal(t, strings.Contains(buffer.buffer.String(), "example log line"), true)
```
codebien, line 169:
```suggestion
	assert.Contains(t, buffer.buffer.String(), "example log line"))
```

log/file_test.go
```diff
@@ -0,0 +1,170 @@
...
+
+	go hook.loop(ctx)
+
+	logger := logrus.New()
+	logger.AddHook(hook)
+	logger.SetOutput(io.Discard)
+
+	logger.Info("example log line")
+
+	time.Sleep(time.Millisecond * 20)
```
codebien, line 167:
```suggestion
cancel()
```
I would prefer to make the concurrency deterministic and not rely on the `time`.

log/tokenizer_test.go
```diff
@@ -61,4 +62,7 @@ func TestTokenizer(t *testing.T) {
 		},
 	}, tokens)
 	assert.NoError(t, err)
+
+	_, err = tokenize("empty=")
+	require.Equal(t, "key `empty=` with no value", err.Error())
```
codebien, line 67:
```suggestion
	assert.EqualError(t, err, "key `empty=` with no value")
```

codebien (changes requested):

log/file_test.go
alyakimenko: We got a race condition here. I consider using `sync.WaitGroup` within a helper struct, something like this
```go
type wgCloser struct {
	io.Writer
	wg sync.WaitGroup
}

func (nc *wgCloser) Close() error {
	nc.wg.Done()
	return nil
}

func TestFileHookFire(t *testing.T) {
	t.Parallel()

	var buffer bytes.Buffer
	wc := &wgCloser{
		Writer: &buffer,
		wg:     sync.WaitGroup{},
	}

	hook := &fileHook{
		ch:     make(chan []byte),
		w:      wc,
		levels: logrus.AllLevels,
	}

	ctx, cancel := context.WithCancel(context.Background())

	wc.wg.Add(1)
	go hook.loop(ctx)

	logger := logrus.New()
	logger.AddHook(hook)
	logger.SetOutput(io.Discard)

	logger.WithContext(ctx).Info("example log line")

	cancel()
	wc.wg.Wait()

	assert.Contains(t, buffer.String(), "example log line")
}
```
What do you think?

codebien: waiting for channel closing could simplify a bit the code

```go
cancel()
<-hook.ch // wait for ch closing
```

codebien (changes requested):
Directly commented in the previous suggestion https://github.com/grafana/k6/pull/2285#discussion_r770796762

codebien (commented):
LGTM :clap:, let's wait for @MStoykov review

codebien:
@alyakimenko in the meanwhile you could consider rebasing the git history to simplify the next review.

log/tokenizer.go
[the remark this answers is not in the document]
MStoykov: Why do you want to support `k6 run file=` ? What would that do?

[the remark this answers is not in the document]
alyakimenko: The check is for the case when a single empty key is provided (like `k6 run --log-output=file=`).
Without the check the case causes `index out of range` panic.
With the check it prints `key %s with no value` error message.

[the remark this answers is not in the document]
MStoykov: ahhh I was left with the impression this was just adding the support to have it be empty :facepalm:

cmd/root.go
```diff
@@ -271,27 +271,43 @@ func (c *rootCommand) setupLoggers() (<-chan struct{}, error) {
...
 		}
 		c.logger.AddHook(hook)
 		c.logger.SetOutput(ioutil.Discard) // don't output to anywhere else
 		c.logFmt = "raw"
+
+	case strings.HasPrefix(line, "file"):
+		if c.fallbackLogger == nil {
+			return nil, fmt.Errorf("fallback logger is nil")
+		}
+
```
MStoykov, line 300: Was this added because in some test this was true? 

I think this is currently not possible to happen in the real code, so I am against there being an error for it ;) (and if it can happen here it needs to be added to the `loki case as well`

log/file.go
```diff
@@ -0,0 +1,157 @@
...
+}
+
+// FileHookFromConfigLine returns new fileHook hook.
+func FileHookFromConfigLine(
+	ctx context.Context, fallbackLogger logrus.FieldLogger, line string,
+) (logrus.Hook, error) {
+	hook := &fileHook{
+		fallbackLogger: fallbackLogger,
+		ch:             make(chan []byte),
+		path:           defaultLogfilePath,
```
MStoykov, line 53: I don't think this needs a defaultLogFilePath especially as it will just log somewhere on hte system. Additionally the current one is Posix specific and likely will have problems on windows.

I do propose that for a first release we :
1. don't have a default one
2. if we decide to have one, it to actually be in the current directory, but I think this needs to be discussed more so let's open an issue instead ?

log/file.go
```diff
@@ -0,0 +1,157 @@
...
+		levels:         logrus.AllLevels,
+	}
+
+	if fallbackLogger != nil {
+		hook.fallbackLogger = fallbackLogger
+	}
+
```
MStoykov, line 60: again afaik there is no case (outside of some tests I guess) where fallbackLogger will be nil so if it's *nil* it's a much bigger issue and IMO worthy of a panic.

log/file.go
```diff
@@ -0,0 +1,157 @@
...
+	return nil
+}
+
+// createOpenFile opens logfile by provided path,
+// it will create a file if it doesn't exist.
+func (h *fileHook) createOpenFile() error {
+	dir := filepath.Dir(h.path)
+	if err := os.MkdirAll(dir, os.ModePerm); err != nil {
+		return fmt.Errorf("failed to create dir for logfile %s: %w", h.path, err)
+	}
```
MStoykov, line 113: I kind of feel like we shouldn't just be creating directories for users if they don't exist, I am much more for erroring out and letting the user create the directory with the permissions they want.

log/file.go
```diff
@@ -0,0 +1,157 @@
...
+	levels         []logrus.Level
+}
+
+// FileHookFromConfigLine returns new fileHook hook.
+func FileHookFromConfigLine(
+	ctx context.Context, fallbackLogger logrus.FieldLogger, line string,
+) (logrus.Hook, error) {
+	hook := &fileHook{
+		fallbackLogger: fallbackLogger,
+		ch:             make(chan []byte),
```
MStoykov, line 52: Maybe this should be buffered. As it stands we won't block the first Log that gets fired but the one after that will block on the channel(given that it happens fast enough).  IMO we either drop the channel or we have some buffering so at least for some burst it won't block at all. I don't have a particular value, but likely something like 100 will be enough, adding it as a configuration option might be a good idea but can also be left for a following PR and just have it as a constant for now.

WDYT.

log/file.go
```diff
@@ -0,0 +1,157 @@
...
+	if err := os.MkdirAll(dir, os.ModePerm); err != nil {
+		return fmt.Errorf("failed to create dir for logfile %s: %w", h.path, err)
+	}
+
+	file, err := os.OpenFile(h.path, os.O_WRONLY|os.O_APPEND|os.O_CREATE, 0o600)
+	if err != nil {
+		return fmt.Errorf("failed to open logfile %s: %w", h.path, err)
+	}
+
+	h.w = file
```
MStoykov, line 120: Maybe we should have this buffered? I remember that there were [some benchmarks](https://www.instana.com/blog/practical-golang-benchmarks/#file-i-o) showing considerable speedup

MStoykov (changes requested):
LGTM in general:clap: :tada: ! But I have some considerations around defaults, for which you will find comments

log/file.go
```diff
@@ -0,0 +1,157 @@
...
+	return nil
+}
+
+func (h *fileHook) loop(ctx context.Context) {
```
oleiade, line 124: A nice pattern I discovered when reading the Concurrency in Go book, is to "confine" concurrency: confining concurrency to a given scope in order to avoid data races. Following this pattern, we would refactor this code to look like this:

```go
func loop(ctx context.Context) chan []byte {
	ch := make(chan []byte)

	go func() {
		defer close(ch)

		for {
			select {
			case entry := <-h.ch:
				if _, err := h.w.Write(entry); err != nil {
					h.fallbackLogger.Errorf("failed to write a log message to a logfile: %w", err)
				}
			case <-ctx.Done():
				if err := h.w.Close(); err != nil {
					h.fallbackLogger.Errorf("failed to close logfile: %w", err)
				}

				return
			}
		}
	}()
	
	return ch
}
```

And in the constructor we would then do something along the lines of:
```go
hook.ch = hook.loop(ctx)
```

I need to explicit that this is in no way a request. When seeing this code, and having to make a bit of mental gymnastic to figure out who declared/used/closed the `ch`, I thought it might be worth mentioning this pattern here. Let me know what you think :smile_cat:

oleiade (commented):
LGTM :champagne: 

Great work overall! I've left a bunch of recommendations and proposals. Let me know what you think. I don't expect to have further comments after that : bullettrain_front:

cmd/root.go
codebien: @MStoykov This was mostly suggested by me. It should make easier the detection of eventual nil logger in any possible future change. It would be difficult to detect an issue just by testing and/or running some examples because it would be used only in case of an error of the main writer. And in the case, the panic would happen we are losing the main error generated by the writer. That is my thought about why I requested to have it. WDYT?

log/file.go
codebien: I agree with your proposal

log/file.go
```diff
@@ -0,0 +1,157 @@
...
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/sirupsen/logrus"
+)
+
+// defaultLogfilePath is a default logfile path.
+const defaultLogfilePath = "/tmp/k6run.log"
```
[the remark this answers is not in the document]
codebien: https://github.com/grafana/k6/pull/2285#discussion_r779506050 I'm for the plan suggested there but this can be a valid alternative to discuss in the issue

log/file.go
```diff
@@ -0,0 +1,157 @@
...
+			return fmt.Errorf("unknown logfile config key %s", token.key)
+		}
+	}
+
+	return nil
+}
+
+// createOpenFile opens logfile by provided path,
+// it will create a file if it doesn't exist.
+func (h *fileHook) createOpenFile() error {
```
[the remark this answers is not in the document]
codebien: >I find the name of this function quite unconventional

IMHO, it would make sense in the case it would be an exported method so a public API for this struct. But it's something internal for wrapping the `OR` logic. If we delegate the input again then better to have directly the entire logic in the caller, otherwise we have just one more layer of logic where we could generate a failure.

Maybe we can improve the name? Just a suggestion, I don't have one better at the moment.

```suggestion
func (h *fileHook) openOrCreateFile() error {
```

log/file.go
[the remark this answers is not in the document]
oleiade: I'll leave it up to you @alyakimenko and @MStoykov to figure out what makes most sense 😸

log/file.go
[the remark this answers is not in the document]
oleiade: I've given this a bit more thought and, given the choice, I do think I'd prefer to get rid of this helper function all in all.

If we were to keep the function, I find that the naming is not the issue. Whether you open, or create the file ends up in the same underlying standard library call, and system call. It's always technically opening a file, with, in one scenario the instruction to not fail if it doesn't exist, but have the OS create it instead. But I'm really, REALLY, nitpicking on this, and I really don't want it to be a blocker! (from my point of view, it's not)

log/file.go
```diff
@@ -0,0 +1,157 @@
...
+	dir := filepath.Dir(h.path)
+	if err := os.MkdirAll(dir, os.ModePerm); err != nil {
+		return fmt.Errorf("failed to create dir for logfile %s: %w", h.path, err)
+	}
+
+	file, err := os.OpenFile(h.path, os.O_WRONLY|os.O_APPEND|os.O_CREATE, 0o600)
```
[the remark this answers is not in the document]
oleiade: Indeed, thanks for the clarification. I've learned something today 🎉

log/file.go
alyakimenko: I got rid of the default log file path

log/file.go
alyakimenko: I replaced the mkdir logic with a check for the existence of a directory.

log/file.go
alyakimenko: I made it buffered and defined `fileHookDataChDefaultSize` constant with a value of 100.

log/file.go
alyakimenko: I made it buffered. But I kept the underlying `io.WriteCloser` in order to have the ability to close a file on exit.

@codebien I also returned the [`loggerStopped` channel](https://github.com/grafana/k6/pull/2285#discussion_r767666922) to properly flush when work is done.

log/file.go
alyakimenko: That's an interesting pattern. Thanks for the recommendation!
I applied it 🙂

log/file_test.go
```diff
@@ -0,0 +1,161 @@
...
+	logger := logrus.New()
+	logger.AddHook(hook)
+	logger.SetOutput(io.Discard)
+
+	logger.Info("example log line")
+
+	cancel()
+	<-hook.dataCh
+
+	assert.Contains(t, buffer.String(), "example log line")
```
MStoykov, line 162: This test now fails. This (IMO) is because we now have a buffered channel between logger.Info -> ... -> fileHook.Fire and the actual place it's logged (the loop). 

With a buffered channel (as long as there is space) the send (in `Fire`) can just put it in the channel and continue. With an unbuffered channel (as before) the send needs to wait for the actual receive on the channel to happen, before it can continue. 

The latter will always manage to enter the `select` and write the message, the first won't. I don't think we care about losing the last few messages in some cases, as in the real case that *should* not be possible. But we still need this test to pass, so please add a small sleep and do `-race -count 10` a few times locally to be sure it's enough. I would wager anything above 1-10ms should be sufficient

alyakimenko: With the timeout, it works as expected. Thanks!
There are a few linter errors:
```
log/file.go:21:1    stylecheck  ST1000: at least one file in a package should have a package comment
log/levels.go:21:1  stylecheck  ST1000: at least one file in a package should have a package comment
```
Should I worry about them?

log/file_test.go
```diff
@@ -0,0 +1,166 @@
...
+	logger := logrus.New()
+	logger.AddHook(hook)
+	logger.SetOutput(io.Discard)
+
+	logger.Info("example log line")
+
+	time.Sleep(10 * time.Millisecond)
+
+	cancel()
+	<-hook.loggerStoppedCh
```
codebien, line 163:
```suggestion
	cancel()
	<-nc.closed
```
I would prefer to remove the time for asserting concurrency, I think we can benefit from the Close method of the WriteCloser in this case.

```
type nopCloser struct {
	io.Writer
	closed chan struct{}
}

func (nc *nopCloser) Close() error {
	nc.closed <- struct{}{}
	return nil
}
```

log/file.go
```diff
@@ -0,0 +1,164 @@
...
+)
+
+// fileHookDataChDefaultSize is a default size for the fileHook's data channel.
+const fileHookDataChDefaultSize = 100
+
+// fileHook is a hook to handle writing to local files.
+type fileHook struct {
+	fallbackLogger  logrus.FieldLogger
+	dataCh          chan []byte
+	loggerStoppedCh chan struct{}
```
codebien, line 42:
```suggestion
```
It seems used only for testing so I think we can remove it and use the alternative proposed below

codebien (changes requested):
Hi @alyakimenko, thanks for the improvements. I spot few suggestions, also the linter should be fixed :pray:

log/file.go
alyakimenko: Yeah, that's right. Thanks!
I removed the channel.

log/file_test.go
alyakimenko: This is a good suggestion. I applied it :)

log/file.go
```diff
@@ -0,0 +1,161 @@
...
+	"github.com/sirupsen/logrus"
+)
+
+// fileHookDataChDefaultSize is a default size for the fileHook's data channel.
+const fileHookDataChDefaultSize = 100
+
+// fileHook is a hook to handle writing to local files.
+type fileHook struct {
+	fallbackLogger logrus.FieldLogger
+	dataCh         chan []byte
```
codebien, line 41:
```suggestion
	loglines         chan []byte
```
I think `dataCh` doesn't mean much, we can be more specific.

log/file.go
```diff
@@ -0,0 +1,161 @@
...
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/sirupsen/logrus"
+)
+
+// fileHookDataChDefaultSize is a default size for the fileHook's data channel.
+const fileHookDataChDefaultSize = 100
```
codebien, line 36:
```suggestion
// fileHookBufferSize is a default size for the fileHook's data channel.
const fileHookBufferSize = 100
```

codebien (changes requested):
I'm sorry few more things

log/file.go
alyakimenko: Renamed

log/file.go
alyakimenko: Renamed :)

MStoykov (commented):
Hi @alyakimenko, due to some refactorings you now will need to rebase this PR on master, sorry for the inconvience :(

alyakimenko:
@MStoykov Hello! I've finished rebasing

MStoykov:
@alyakimenko I think you didn't pull the upstream changes before rebasing :(

alyakimenko:
@MStoykov Sorry, it's my bad. I fixed it.

log/file_test.go
alyakimenko: Looks like, It didn't help :(

alyakimenko: How would you like it if I brought the timeout back?

codebien: Yep, I'm sorry this was my fault, the problem is with the Write sync order. You should restore the timeout. :pensive:

log/file_test.go
codebien: @alyakimenko Yes, please. The linter must be fixed for getting the PR merged.

log/file_test.go
alyakimenko: I restored the timeout.

log/file_test.go
alyakimenko: @codebien
`Package log implements various logrus hooks.`
Do you find the description suitable for the package?

codebien: @alyakimenko it's what it does so it's fine for me, but other reviewers could have different opinions

codebien (dismissed):
LGTM

log/file.go
```diff
@@ -0,0 +1,162 @@
...
+	h.bw = bufio.NewWriter(file)
+
+	return nil
+}
+
+func (h *fileHook) loop(ctx context.Context) chan []byte {
+	loglines := make(chan []byte, fileHookBufferSize)
+
+	go func() {
+		defer close(loglines)
```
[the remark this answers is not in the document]
alyakimenko: Yeah, that makes sense. Thank you. I removed the `defer`.

MStoykov (approved):

codebien (approved):
@alyakimenko going to merge :tada:  thanks for your effort :pray:

PR #2301, opened 2021-12-16
js/grpc: Tests for multi services in same proto file

### LINKED ISSUE #2295: grpc: Unit test the multi services fix
https://github.com/grafana/k6/pull/2265 introduced a fix for a proto file with multi-services defined. The new addition is not covered by unit tests. The suggestion is to add one for covering the added logic.

It would be nice to address also:
* https://github.com/grafana/k6/pull/2265#discussion_r767889043
* https://github.com/grafana/k6/issues/2297

codebien:
Closes #2295

### DIFF
--- a/js/modules/k6/grpc/client.go
+++ b/js/modules/k6/grpc/client.go
@@ -342,7 +342,7 @@ type fileDescriptorLookupKey struct {
 }
 
 func resolveServiceFileDescriptors(
-	mc reflectpb.ServerReflection_ServerReflectionInfoClient,
+	mc sendReceiver,
 	res *reflectpb.ListServiceResponse,
 ) (*descriptorpb.FileDescriptorSet, error) {
 	services := res.GetService()
@@ -384,10 +384,18 @@ func resolveServiceFileDescriptors(
 	return fdset, nil
 }
 
+// sendReceiver is a smaller interface for decoupling
+// from reflectpb.ServerReflection_ServerReflectionInfoClient
+// that has the dependency from grpc.ClientStream.
+type sendReceiver interface {
+	Send(*reflectpb.ServerReflectionRequest) error
+	Recv() (*reflectpb.ServerReflectionResponse, error)
+}
+
 // sendReceive sends a request to a reflection client and,
 // receives a response.
 func sendReceive(
-	client reflectpb.ServerReflection_ServerReflectionInfoClient,
+	client sendReceiver,
 	req *reflectpb.ServerReflectionRequest,
 ) (*reflectpb.ServerReflectionResponse, error) {
 	if err := client.Send(req); err != nil {
--- a/js/modules/k6/grpc/client_test.go
+++ b/js/modules/k6/grpc/client_test.go
@@ -24,13 +24,18 @@ import (
 	"bytes"
 	"context"
 	"errors"
+	"fmt"
 	"net/url"
 	"os"
 	"runtime"
 	"strings"
+	"sync/atomic"
 	"testing"
 
 	"google.golang.org/grpc/reflection"
+	reflectpb "google.golang.org/grpc/reflection/grpc_reflection_v1alpha"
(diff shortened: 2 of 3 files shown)


js/modules/k6/grpc/client_test.go
```diff
@@ -761,3 +766,101 @@ func TestDebugStat(t *testing.T) {
...
+				lsr  = &reflectpb.ListServiceResponse{}
+				mock = &getServiceFileDescriptorMock{}
+			)
+			for i, service := range tt.services {
+				// if only one package is defined then
+				// the package is the same for every service
+				pkg := tt.pkgs[0]
+				if len(tt.pkgs) > 1 {
+					pkg = tt.pkgs[i]
+				}
```
yorugac, line 818: I think this implies that the following would work just as well:
```suggestion
				// if only one package is defined then
				// the package is the same for every service
				pkg := tt.pkgs[0]
```
Just in case, I've already checked :slightly_smiling_face: The test successfully passes so you can leave only the 1st element in `pkgs`. From that follows that you can simplify `pkgs` field down to one `string`. This is true for all test cases here, in `TestResolveFileDescriptors`.

But at the moment I'm not fully clear that this is the intended logic: i.e. if there _can be_ a case with multiple services and multiple packages, it must be reflected in the test cases in such a way that it would fail without multiple packages.
Could you clarify the logic / expectations / test cases here?

js/modules/k6/grpc/client.go
```diff
@@ -384,10 +384,18 @@ func resolveServiceFileDescriptors(
 	return fdset, nil
 }
 
+// sendReceiver is a smaller interface for decoupling
+// from reflectpb.ServerReflection_ServerReflectionInfoClient
+// that has the dependency from grpc.ClientStream.
+type sendReceiver interface {
+	Send(*reflectpb.ServerReflectionRequest) error
+	Recv() (*reflectpb.ServerReflectionResponse, error)
+}
```
yorugac, line 393: This is basically a mock for `ServerReflection_ServerReflectionInfoClient` for tests, right? Could you please update the comment to reflect that specific point, about mock & tests? Right now it sounds a tad convoluted (and therefore, meaning might be lost in the future) :sweat_smile:

yorugac (changes requested):
Hi @codebien in general all clear but I have doubts about correct logic of this test -- please see the comments for details :slightly_smiling_face:

codebien (commented):
Auto spot, this is missing https://github.com/grafana/k6/pull/2265#discussion_r767889043

js/modules/k6/grpc/client_test.go
codebien: Thanks for catching it, I pushed a fixup for making the test fail without that condition. Let me know if it is clear or something is missing :+1:

js/modules/k6/grpc/client.go
codebien: Yes, mock is the concrete action in this case, but generally I don't think it's a good thing have a so big dependency with no reason. Tried to improve the comment in the fixup.

yorugac (approved):
Thanks for the update! LGTM :+1:

js/modules/k6/grpc/client_test.go
```diff
@@ -761,3 +766,101 @@ func TestDebugStat(t *testing.T) {
...
+}
+
+type getServiceFileDescriptorMock struct {
+	nreqs int64
+	pkgs  []string
+	names []string
+}
+
+func (m *getServiceFileDescriptorMock) Send(req *reflectpb.ServerReflectionRequest) error {
+	return nil
```
MStoykov, line 841: Maybe add a TODO to actually check what is sent if it has anything interesting to be tested around it

MStoykov (approved):

MStoykov (approved):

yorugac (approved):

PR #2304, opened 2021-12-17
refactor: JSON:API structures for the RestAPI, part 1

olegbespalov:
### What this PR does

Minimizes the dependencies that are used in the k6. Replaces `api2go/jsonapi` with a few envelop structures JSON API.

Deprecates the methods that will be removed soon.

Introduces the `client.callAPI` method that doesn't rely on the "magic" of marshaling/unmarshaling and expects the prepared structures on input/output.

### Scope

This PR introduces additional structures that are structured the way how the [JSON:API](https://jsonapi.org/) requires. The actual removal of the `api2go/jsonapi` is in #2313.

### Expected feedback and Open Questions

- **Question** Is it okay to deprecate the `client.Call` method like that?
- **Feedback needed** Are we okay with such simplifying of the code?

### Related PRs

#2313


olegbespalov:
@MStoykov @na-- @codebien The PR is marked as WIP, but I'd like to get your opinions since it's my first PR

codebien (changes requested):
@olegbespalov What is missing to consider it still in progress? Please, add it to the description. Can you add also more details in the description or in the commit messages about what you did and your decisions, please? It would be helpful for reviewers. :pray:

codebien:
I forgot to add this part:

I would prefer to not mention the issue it's going to be closed in the commit messages otherwise GitHub will add noise to the issue (as you can see). The form `Closes #issueID` in the description or the commit message should be better to avoid it and it will make the PR linked also.

oleiade:
Grüß Dich @olegbespalov,

Could you document, in the Pull Request description, how you approach running and interacting with the API locally? I've started looking into your work, but I must admit that being still new to K6, I'm not sure how I can run the API locally in order to talk to it and introspect what comes out of it :)

Vielen Dank 🙏🏻

olegbespalov:
Hey @oleiade! Sure, but I'll do this with this comment if you don't mind.

The K6's API is automatically running with the execution of the k6, so the easiest way to try it, run the k6 in the paused sated with the following command:

```
k6 run --paused samples/es6sample.js
```

Once you do that, you'll be able to curl the API using the `http://127.0.0.1:6565`

for instance:
```
curl -s http://127.0.0.1:6565/v1/metrics | jq        
{
  "data": [
    {
      "type": "metrics",
      "id": "vus",
      "attributes": {
        "type": "gauge",
        "contains": "default",
        "tainted": null,
        "sample": {
          "value": 0
        }
      }
    },
    {
      "type": "metrics",
      "id": "vus_max",
      "attributes": {
        "type": "gauge",
        "contains": "default",
        "tainted": null,
        "sample": {
          "value": 1
        }
      }
    }
  ]
}
```

For sure, during the long run, it's possible to observe the changes in the responses.

oleiade:
Perfect, thanks a lot @olegbespalov 🙏🏻

oleiade (commented):
LGTM :+1: 

Well done @olegbespalov!

A few comments:
- The JSON:API specification has a strict policy when it comes to [content negociation](https://jsonapi.org/format/#content-negotiation). Namely, the client **MUST** send all JSON:API data with the header `Content-Type: application/vnd.api+json`, and the server should basically refuse anything that does not have this header. I acknowledge that this might be a breaking change, considering the `api/v1` does not seem to follow this strictly either. I also acknowledge that considering this is more of a control-oriented API, it might not be so important. But I wanted to make sure I mention to be sure that whatever we (or don't) on that front is explicit :smile:
- In general, I've always heard (and followed) the idiom of ordering the content of Go files by publicness (public symbols should appear first, at the top of the file), and importance (symbols that are important to the user/understanding should appear first, and implementation details last). I've noticed that private symbols were often appearing first throughout the code, and wondered if it was a conscious explicit decision, or would you be open to reordering things?

Hope that's helpful! My understanding is that the branch is still a work in progress, so let me know when I should come back and smash the approve button +1:

api/v1/client/client.go
```diff
@@ -99,7 +98,8 @@ func (c *Client) Call(ctx context.Context, method string, rel *url.URL, body, ou
 		case string:
 			bodyData = []byte(val)
 		default:
-			bodyData, err = jsonapi.Marshal(body)
+			// TODO: not sure if it safe, should we introduce another Call ?
```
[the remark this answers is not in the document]
olegbespalov: Sure! Sorry, my bad, I wasn't clear enough, and it seems that I should bring this into the description since it's one of my concerns about this PR.

Let me fix it quickly!

olegbespalov:
Hey @oleiade !

Thanks for the review & your valid input :relaxed: 

> The JSON:API specification has a strict policy when it comes to content negotiation.

Good catch! However, I'm not sure how strict we are in the following specifications :thinking: I mean, maybe we follow the structure of the JSON:API, but don't implement the specification. Let me try to find the answer to that question :+1: 

Regarding the ordering thing. To be honest, there was no conscious, explicit decision about this. The conclusion that I was following was to use the private first, but some of the structures become public with the latest changes. 

The reasons behind your request sound reasonable. But on the other side, they are style things, and I believe if it's possible, we should catch it with the linter :robot: Wdyt?

Anyway, I'll do investigations & update the PR.

api/v1/metric_jsonapi.go
```diff
@@ -0,0 +1,82 @@
...
+func newMetricEnvelope(m *stats.Metric, t time.Duration) metricJSONAPI {
+	return metricJSONAPI{
+		Data: newMetricData(m, t),
+	}
+}
+
+func newMetricsJSONAPI(list map[string]*stats.Metric, t time.Duration) MetricsJSONAPI {
+	metrics := make([]metricData, 0, len(list))
+
+	// TODO: lock ?
```
codebien, line 53: I guess it could be a problem, is the pointer required?

api/v1/metric_jsonapi.go
```diff
@@ -0,0 +1,82 @@
...
+)
+
+type metricJSONAPI struct {
+	Data metricData `json:"data"`
+}
+
+// MetricsJSONAPI is JSON API envelop for metrics
+type MetricsJSONAPI struct {
+	Data []metricData `json:"data"`
+}
```
codebien, line 36: Why `Metrics` is exported and `metric` not?

codebien (changes requested):
The current approach seems fine to me, the unique issue I can see is that we are using Marshal and Unmarshal relying on the same struct in the tests. I would prefer to see somewhere tests having the input in a more controlled way. Like passing a pre-defined `[]byte` string. So we should avoid introducing hidden bugs generated by a bad struct mapping that it should be essential for this refactor.

api/v1/metric_jsonapi.go
olegbespalov: A `MetricsJSONAPI` is used outside the package, and `metricJSONAPI` is not. So I decided to keep it unexported to limit usages.

api/v1/metric_jsonapi.go
olegbespalov: I believe here the pointers aren't required, but in general, it's `engine.Metrics`. And I see the lock there, but it used only for the unexported methods of the engine :thinking:

olegbespalov: Fixed with the latest commit

api/v1/status_routes_test.go
```diff
@@ -80,16 +86,43 @@ func TestGetStatus(t *testing.T) {
...
 
 func TestPatchStatus(t *testing.T) {
+	t.Parallel()
+
 	testdata := map[string]struct {
-		StatusCode int
-		Status     Status
+		StatusCode     int
+		ExpectedStatus Status
+		Payload        []byte
```
codebien, line 94:
```suggestion
		ExpectedStatusCode     int
		ExpectedStatus Status
		Payload        []byte
```
I would prefer to change it in ExpectedStatusCode to be consistent with `ExpectedStatus`

api/v1/status_routes_test.go
```diff
@@ -103,8 +136,10 @@ func TestPatchStatus(t *testing.T) {
 	registry := metrics.NewRegistry()
 	builtinMetrics := metrics.RegisterBuiltinMetrics(registry)
 
-	for name, indata := range testdata {
+	for name, inData := range testdata {
```
codebien, line 139: inData remember to me `inputData` but they are also `expectedData` so I would prefer a simple `td` as abbreviation of a generic `testdata`

api/v1/metric_routes.go
```diff
@@ -36,7 +36,10 @@ func handleGetMetrics(rw http.ResponseWriter, r *http.Request) {
 		t = engine.ExecutionScheduler.GetState().GetCurrentTestRunDuration()
 	}
 
+	engine.MetricsLock.Lock()
 	metrics := newMetricsJSONAPI(engine.Metrics, t)
+	engine.MetricsLock.Unlock()
```
codebien, line 41: Checking the code it makes sense but it's strange that here we didn't see a data race before. Do we know the usage of this endpoint? We should check that we are not impacting the engine performances.

codebien (changes requested):
Some few comments

api/v1/client/client.go
```diff
@@ -72,15 +72,52 @@ func WithHTTPClient(httpClient *http.Client) Option {
...
+// WithLogger sets the specified logger to the client.
 func WithLogger(logger *logrus.Entry) Option {
 	return Option(func(c *Client) {
 		c.logger = logger
 	})
 }
 
+// CallAPI executes the desired REST API request.
+// it's expected that the body and out are the structs that follows the JSON:API
+func (c *Client) CallAPI(ctx context.Context, method string, rel *url.URL, body, out interface{}) (err error) {
```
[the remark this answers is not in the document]
olegbespalov: yes, there are some usages in the other projects that imports k6

api/v1/status_routes_test.go
```diff
@@ -118,29 +153,26 @@ func TestPatchStatus(t *testing.T) {
...
 			}
-			if indata.Status.VUs.Valid {
-				assert.Equal(t, indata.Status.VUs, status.VUs)
+			if inData.ExpectedStatus.VUs.Valid {
+				assert.Equal(t, inData.ExpectedStatus.VUs, status.VUs)
 			}
-			if indata.Status.VUsMax.Valid {
-				assert.Equal(t, indata.Status.VUsMax, status.VUsMax)
+			if inData.ExpectedStatus.VUsMax.Valid {
+				assert.Equal(t, inData.ExpectedStatus.VUsMax, status.VUsMax)
```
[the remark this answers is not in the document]
olegbespalov: Not really, the thing is that in that case, the expected will be invalid false, but the one from the response will be valid false.

```
expected: null.Bool{NullBool:sql.NullBool{Bool:false, Valid:false}}
actual  : null.Bool{NullBool:sql.NullBool{Bool:false, Valid:true}}
```

So I believe this check is fine

api/v1/metric_routes.go
olegbespalov: On top of my head, I have no answer about the usage of this endpoint, but I guess that this endpoint is just a nice feature.

Usually, external outputs are used for the metric collection.

api/v1/status_routes_test.go
olegbespalov: `td` is too short for the scope, but renamed the variable to the `testCase` since the `testData` is holding the table tests.

api/v1/client/client.go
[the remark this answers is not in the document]
codebien: Do we plan to unexport in https://github.com/grafana/k6/pull/2313?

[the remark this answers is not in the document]
olegbespalov: No, there is no plan to make this non-exported. If we do so, the components that rely on this logic should repeat the call method.

Instead of this, we can think about extracting the client code to the separated package, and there this can be internal, but I believe we should do that after the refactoring of the API.

api/v1/metric_routes.go
```diff
@@ -37,12 +36,11 @@ func handleGetMetrics(rw http.ResponseWriter, r *http.Request) {
 		t = engine.ExecutionScheduler.GetState().GetCurrentTestRunDuration()
 	}
 
-	metrics := make([]Metric, 0)
-	for _, m := range engine.Metrics {
-		metrics = append(metrics, NewMetric(m, t))
-	}
+	engine.MetricsLock.Lock()
+	metrics := newMetricsJSONAPI(engine.Metrics, t)
+	engine.MetricsLock.Unlock()
```
MStoykov, line 41: we had a race ?

MStoykov (dismissed):
LGTM! thanks, @olegbespalov,  for doing all of this "manual" marshalling and unmarshalling. I don't really want to double-check with the `jsonapi` if all of this is exactly as it should be. But at least using k6 v0.35.0 I could talk to this version and everything seemed fine, so hopefully it's all correct :crossed_fingers: .

oleiade (approved):
LGTM (again 😸) 👍🏻

api/v1/metric_routes.go
codebien: Yes, in the case that the API is hit during the test

codebien (dismissed):
LGTM

MStoykov (approved):

codebien (approved):

PR #2313, opened 2021-12-28
refactor: removal of the manyminds/api2go, part 2

olegbespalov:
### What this PR does

It is a second PR related to the cleaning of `api2go/jsonapi` dependency. Once the client's usage is removed from the other projects, we can merge this.

### Scope

The scope of this PR is to remove the `api2go/jsonapi` from the k6.

### Related PRs

#2304 

Closes: #911


codebien (commented):
When do we expect to merge this? Will it be merged after the `v0.36.0` release?

olegbespalov:
@codebien, to be honest, I don't have a sure answer, probably yes, since even the first PR in a row is under the changes. 

But, why?

codebien:
We already added the deprecate warning in the first PR and this PR is removing it, so I didn't get why if we would merge it in the same release.

olegbespalov:
We can merge it in the same release, but only once we're sure that no other components are using the `client.Call`

codebien:
I don't see the value to merge the part 1 without this part. I think we should just have a unique PR and maintain it blocked until all the calls to the `Call` method haven't been removed. But let's wait for more opinions.

olegbespalov:
> I don't see the value to merge the part 1 without this part. I think we should just have a unique PR and maintain it blocked until all the calls to the Call method haven't been removed. But let's wait for more opinions.

The main thing why it split into two parts is because it's a breaking change. It's usually a good practice to make such changes. Once the first part is merged, the other components have time to be migrated without a rush. Once the other components are fully migrated to the new interface, PR can be merged.

oleiade (dismissed):

MStoykov (dismissed):
LGTM. I wonder if we should let there be a version with the deprecations and then removing it ... :thinking: I kind of doubt there are many users and I doubt they will even notice if it doesn't get removed so :shrug:

api/v1/group.go
```diff
@@ -84,63 +82,6 @@ func NewGroup(g *lib.Group, parent *Group) *Group {
...
-// SetID sets a group ID
-// Deprecated: use instead g.ID directly
-// This method will be removed with the one of the PRs of (https://github.com/grafana/k6/issues/911)
-func (g *Group) SetID(v string) error {
-	g.ID = v
-	return nil
-}
-
-// GetReferences returns the slice of jsonapi.References
-// Deprecated: use instead g.Groups properties
```
MStoykov, line 103: If we decide to keep the deprecation for a version it seems to me like `// Deprecated` needs to be the last comment and separated with a newline from the rest in order for it to work ... but I might be wrong.

oleiade (dismissed):

codebien (approved):
LGTM

MStoykov (dismissed):

oleiade (approved):

MStoykov (approved):

codebien (approved):

Issue #1079, opened 2019-07-12
Fork afero


MStoykov:
During #1059 it became obvious [afero](https://github.com/spf13/afero) has some problems specifically with it's [MemMapFs](https://github.com/spf13/afero/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+memmapfs), but also missing features in [CacheOnReadFs](https://github.com/loadimpact/k6/blob/6b08d619bc6b65352bf4238966be37729f962e81/lib/fsext/cacheonread.go), problems with [Walk](https://github.com/loadimpact/k6/blob/6b08d619bc6b65352bf4238966be37729f962e81/lib/fsext/walk.go)(which are due to the way MemMapFs works in some ways) and possibly many more.

Given that the cleanest (and I would argue very good) solution is for MemMapFs to be rewritten in a way that it fixes all it's bugs and is in general not hacky, I propose that we fork afero and use the [go modules](https://github.com/loadimpact/k6/issues/1070) [replace](https://github.com/golang/go/wiki/Modules#when-should-i-use-the-replace-directive) functionality. We should still try to merge all the fixes in the original repo, but unfortunately it looks pretty dead to me so it will probably take a while ... 

This is definitely not *high* priority, as the current workarounds work, but it will still make the experience of using MemMapFs and afero much less error-prone and with less hacks.

na--:
I think the issue should be called "Fork or replace afero", since it might be better if we use another library, instead of trying to fix all of the issues with afero, and then potentially inherit the maintenance burden of the forked library... For example, it might be worth investigating https://github.com/C2FO/vfs, since it seems much more active and it's supposedly inspired by afero. Or maybe https://github.com/twpayne/go-vfs, since it supposedly has afero compatibility. Or maybe this: https://godoc.org/golang.org/x/tools/godoc/vfs - this seems semi-official :smile: 

But yeah, if we end up sticking with afero and forking it, go modules (https://github.com/loadimpact/k6/issues/1070) definitely seems like a prerequisite...

na--:
To update this, it's probably worth waiting for https://tip.golang.org/pkg/io/fs/ and seeing what gets built on top of it

na--:
Go 1.16 was released:
- https://golang.org/doc/go1.16#fs
- https://benjamincongdon.me/blog/2021/01/21/A-Tour-of-Go-116s-iofs-package/

yorugac:
[...]

PR #2314, opened 2022-01-04
fix: Implementing a limitation for the open()

### LINKED ISSUE #1771: open() seems to not always return the full data if that wasn't loaded in __VU =0
The following code:
```javascript
if (__VU >0) {
JSON.parse(open("./arr.json"));
}
```
Where arr.json is sufficiently big JSON seems to fail sometimes because not the whole JSON gets loaded. 

This is, technically, a  bad script, but maybe it should fail in another way that is more explanatory of what is happening or not at all

olegbespalov:
Implementing a limitation for the open() that limits opening files
by the list of the files that were opened during the initialization
step (__VU == 0).

For example, a code like:

```js
if (__VU >0) {
   JSON.parse(open("./arr.json"));
}
```

Should return an error.

Closes #1771

### DIFF
--- a/js/bundle.go
+++ b/js/bundle.go
@@ -334,6 +334,10 @@ func (b *Bundle) instantiate(logger logrus.FieldLogger, rt *goja.Runtime, init *
 	unbindInit()
 	*init.ctxPtr = nil
 
+	if vuID == 0 {
+		init.allowOnlyOpenedFiles()
+	}
+
 	rt.SetRandSource(common.NewRandSource())
 
 	return nil
--- a/js/initcontext.go
+++ b/js/initcontext.go
@@ -319,7 +319,7 @@ func readFile(fileSystem afero.Fs, filename string) (data []byte, err error) {
 			// loading different files per VU is not supported, so all files should are going
 			// to be used inside the scenario should be opened during the init step (without any conditions)
 			err = fmt.Errorf(
-				"open() can't be used under with files that weren't opened during initialization (__VU==0), path: %q",
+				"open() can't be used with files that weren't previously opened during initialization (__VU==0), path: %q",
 				filename,
 			)
 		}
@@ -339,12 +339,12 @@ func readFile(fileSystem afero.Fs, filename string) (data []byte, err error) {
 func (i *InitContext) allowOnlyOpenedFiles() {
 	fs := i.filesystems["file"]
 
-	alreadyOpenedFS, ok := fs.(fsext.OnlyOpenedEnabler)
+	alreadyOpenedFS, ok := fs.(fsext.OnlyCachedEnabler)
 	if !ok {
 		return
 	}
 
-	alreadyOpenedFS.AllowOnlyOpened()
+	alreadyOpenedFS.AllowOnlyCached()
 }
 
 func getInternalJSModules() map[string]interface{} {
--- a/js/initcontext.go
+++ b/js/initcontext.go
@@ -313,7 +313,18 @@ func (i *InitContext) Open(ctx context.Context, filename string, args ...string)
 	return i.runtime.ToValue(string(data)), nil
 }
 
-func readFile(fileSystem afero.Fs, filename string) ([]byte, error) {
+func readFile(fileSystem afero.Fs, filename string) (data []byte, err error) {
+	defer func() {
+		if errors.Is(err, fsext.ErrPathNeverRequestedBefore) {
+			// loading different files per VU is not supported, so all files should are going
+			// to be used inside the scenario should be opened during the init step (without any conditions)
+			err = fmt.Errorf(
+				"open() can't be used under with files that weren't opened during initialization (__VU==0), path: %q",
+				filename,
+			)
+		}
+	}()
+
 	// Workaround for https://github.com/spf13/afero/issues/201
 	if isDir, err := afero.IsDir(fileSystem, filename); err != nil {
 		return nil, err
@@ -321,18 +332,7 @@ func readFile(fileSystem afero.Fs, filename string) ([]byte, error) {
 		return nil, fmt.Errorf("open() can't be used with directories, path: %q", filename)
 	}
 
-	data, err := afero.ReadFile(fileSystem, filename)
-	if err == nil {
-		return data, nil
-	}
-
-	// loading different files per VU is not supported, so all files should are going
-	// to be used inside the scenario should be opened during the init step (without any conditions)
-	if errors.Is(err, fsext.ErrFileNeverOpenedBefore) {
-		return nil, fmt.Errorf("open() can't be used under the conditions, path: %q", filename)
-	}
-
-	return nil, err
+	return afero.ReadFile(fileSystem, filename)
 }
 
 // allowOnlyOpenedFiles enables seen only files
--- a/js/initcontext.go
+++ b/js/initcontext.go
@@ -48,6 +48,7 @@ import (
 	"go.k6.io/k6/js/modules/k6/metrics"
 	"go.k6.io/k6/js/modules/k6/ws"
 	"go.k6.io/k6/lib"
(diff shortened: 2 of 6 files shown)


js/initcontext.go
```diff
@@ -317,6 +313,40 @@ func (i *InitContext) Open(ctx context.Context, filename string, args ...string)
 	return i.runtime.ToValue(string(data)), nil
 }
 
+func readFile(fileSystem afero.Fs, filename string) ([]byte, error) {
+	// Workaround for https://github.com/spf13/afero/issues/201
+	if isDir, err := afero.IsDir(fileSystem, filename); err != nil {
```
MStoykov, line 318: Because of this the following code:
```
if (__VU == 1) {
   open("./somenonexistingfile.js")
}
```

errors out with "no such file or directory" instead of the message below. This can be fixed later, as at least I don't know of a good fix without having to reimplement `fs.Stat` as well and we want to [move away](#1079 ) from afero either way. On the other hand that doesn't seem like that big of a change so :shrug:

MStoykov (changes requested):
LGTM, nicely done :clap: ! 

I have left some nitpick comments, but mostly I want the message fixed ... possibly when the file is missing as well :). 
I have tested it and it seems to work great with archives as well ;)

js/initcontext.go
olegbespalov: That's fine. It's a tiny change :+1:

olegbespalov:
Thanks, @MStoykov :bow: 

Addressed all comments :+1:

lib/fsext/cacheonread.go
```diff
@@ -21,27 +21,95 @@
...
+	"errors"
+	"os"
+	"sync"
 	"time"
 
 	"github.com/spf13/afero"
 )
 
+// ErrPathNeverRequestedBefore represent an error when path never opened/requested before
+var ErrPathNeverRequestedBefore = errors.New("path nevere requested before")
```
MStoykov, line 33:
```suggestion
var ErrPathNeverRequestedBefore = errors.New("path never requested before")
```

js/initcontext.go
```diff
@@ -317,6 +313,40 @@ func (i *InitContext) Open(ctx context.Context, filename string, args ...string)
...
+	defer func() {
+		if errors.Is(err, fsext.ErrPathNeverRequestedBefore) {
+			// loading different files per VU is not supported, so all files should are going
+			// to be used inside the scenario should be opened during the init step (without any conditions)
+			err = fmt.Errorf(
+				"open() can't be used under with files that weren't opened during initialization (__VU==0), path: %q",
+				filename,
+			)
+		}
+	}()
```
MStoykov, line 326: Note(not actionable): Technically, this error will either happen on `afero.IsDir` or not at all. But it seems kind of strange to check it only there and also when/if we ever stop checking that as we move to [io/fs](#1079) it will be a bit confusing why it wasn't previously checked.

I am noting this for future reference only, don't do anything about it just based on that.

js/initcontext.go
```diff
@@ -317,6 +313,40 @@ func (i *InitContext) Open(ctx context.Context, filename string, args ...string)
...
+func (i *InitContext) allowOnlyOpenedFiles() {
+	fs := i.filesystems["file"]
+
+	alreadyOpenedFS, ok := fs.(fsext.OnlyOpenedEnabler)
+	if !ok {
+		return
+	}
+
+	alreadyOpenedFS.AllowOnlyOpened()
+}
```
MStoykov, line 348: Note(not actionable): This seems a bit fragile, as we can at any point  change what the top fs for `file` is and then this suddenly won't work. This is *unlikely* as this code is very rarely changed and will probably be spotted.

The test below also doesn't test with `loader.CreateFilesystems` which is what is used in reality. An integration test with it might be better, but I see no test actually has ever used it so :shrug: maybe there are some problems I can't remember at the moment. Either way this is fine for now, hopefully we will be moving to io/fs soon and that will lead to some bigger changes and hopefully more tests.

MStoykov (approved):
LGTM :clap: I have 2 more comments with notes, but I wouldn't argue you should work on them now especially if the plan is to move to io/fs in the next release, which will in practice redo all of this either way.

yorugac (changes requested):
LGTM overall :+1: but I have a renaming request and some smaller cleanup remarks.

One more thing. To improve test coverage it might also make sense to add tests for successful cases (i.e. with `if (__VU==0)` and with absent `if` condition)?

js/runner_test.go
```diff
@@ -1157,6 +1158,55 @@ func TestVUIntegrationOpenFunctionErrorWhenSneaky(t *testing.T) {
...
+func TestVUDoesNotOpenUnderConditions(t *testing.T) {
+	t.Parallel()
+
+	baseFS := afero.NewMemMapFs()
+	data := `
+			if (__VU > 0) {
+				data = open("/home/somebody/test.json");
+			}
+			exports.default = function(data) {
+				console.log("hey")
```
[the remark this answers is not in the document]
MStoykov:
```suggestion
			if (__VU > 0) {
				var data = open("/home/somebody/test.json");
			}
			exports.default = function() {
				console.log("hey")
```

js/runner_test.go
```diff
@@ -1157,6 +1158,55 @@ func TestVUIntegrationOpenFunctionErrorWhenSneaky(t *testing.T) {
...
+func TestVUDoesNonExistingPathnUnderConditions(t *testing.T) {
+	t.Parallel()
+
+	baseFS := afero.NewMemMapFs()
+	data := `
+			if (__VU == 1) {
+				data = open("/home/nobody");
+			}
+			exports.default = function(data) {
+				console.log("hey")
```
MStoykov, line 1195:
```suggestion
				let data = open("/home/nobody");
			}
			exports.default = function() {
				console.log("hey")
```

js/runner_test.go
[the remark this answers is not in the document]
yorugac: Yes, that's what I meant.

js/initcontext.go
olegbespalov: :thinking: not sure if I get it. The error happens `afero.IsDir` and `afero.ReadFile`. Even if we move to the `io/fs` still should happen on the equivalent of the `afero.ReadFile`, no?

MStoykov: When we move – yes, but when we move to io/fs,  `afero.IsDir` will *hopefully* be just dropped ;). 

What I am trying to say is that if `afero.IsDir` did not return `fsext.ErrPathNeverRequestedBefore` than `afero.Readfile` will not return it either ... or we have bugs I guess or some strange race condition.

I wondered for a bit if we can't just add `afero.IsDir` check in the already reimplemented `Open` , but I am not certain we always use the CacheOnReadFs and IIRC we don't so :shrug:

MStoykov (approved):

yorugac (dismissed):
Thanks for the renaming! LGTM :+1:

MStoykov (approved):

yorugac (approved):

js/bundle.go
```diff
@@ -334,6 +334,10 @@ func (b *Bundle) instantiate(logger logrus.FieldLogger, rt *goja.Runtime, init *
 	unbindInit()
 	*init.ctxPtr = nil
 
+	if vuID == 0 {
+		init.allowOnlyOpenedFiles()
+	}
```
na--, line 339: I am probably be missing something, but this seems wrong? :confused: Isn't the whole point that VUs > 0 shouldn't be able to open non-cached files?

olegbespalov: Yes, the point is that VUs > 0 shouldn't be able to open non-cached files.

This line notifies the FS that VU#0 did the first run, and the list of "known" files list can be frozen.

Issue #2318, opened 2022-01-06
Log unhandled Promise rejections


MStoykov:
### Feature Description

Currently if a promise gets rejected and there is no `reject` handle this will *not* be logged or noticed in any way, for example:
```javascript
export default function() {
        new Promise((resolve, reject) => {
                setTimeout(resolve, 1000);
        }).then(()=> console.log("yey"));
}
```
will not print anything at all
but if you write
```javascript
export default function() {
        new Promise((resolve, reject) => {
                setTimeout(resolve, 1000);
        }).then(()=> console.log("yey")).catch((e)=>console.log("oh", e));
}
```
you will get `oh ReferenceError: setTimeout is not defined`

### Suggested Solution (optional)

The ECMAScript standard has a defined "callback" to handle this called [HostPromiseRejectionTracker](https://262.ecma-international.org/12.0/#sec-host-promise-rejection-tracker) as noted there it gets called with the promise and one of *two* operations:
1. "reject" - a Promise is rejected *and* it doesn't have rejection handler at this point.
2. "handle" - a Promise that previously got called with the "reject" handler just got a rejection handler added. 

Here the idea is that the host implementation (this is k6 ;) ) will hopefully have some logic that if it gets a "reject" it won't immediately  just log it but wait for sometime, until some event?, both or a third option, before logging if a corresponding "handle" hasn't also arrived. 

I found this out while working on #2228, and I quickly found out that a really good implementation is ... going to take some time. So instead I [just logged it each time](https://github.com/grafana/k6/pull/2228/files#diff-663d689366c7fca665dcbb0e54a8401fbcd724f3ec4bce700c306a82926da9b9R313-R336) I got "reject" and did nothing on "handle". 

This unfortunately means that at least for the moment, before an event loop and some API from k6 to return promises which will rejected/resolved through it, this will happen for all create Promises. As in the above example, the Promise "body" gets executed *before* `new Promise` returns so by the time `.then` or `.catch` is called on it, it's already too late - it was already rejected.

Now the current main usage is to polyfill async/await and as I noted in https://github.com/grafana/k6/issues/779#issuecomment-1006418756 this isn't really ... supported. Given that and the fact that otherwise it gets a bit too hard to debug even that I am kind of for the simplest solution, possibly even added to `v0.36.0` and then with event loop PR(or after/before it) a more complicated and full implementation can be though of.

Implementation ideas/problems:
[...]

Issue #779, opened 2018-09-25
How to use imported async functions in test script? 


kkeranen:
It seem that [async code is not supported in k6](https://github.com/loadimpact/k6/issues/194),  so how imported async methods should be used? I demonstrate the problem by two examples.

For context initialization I would simply like to code something like:

```
async () => {
  await login();
}
```

The reason being that login procedure is quite complex and for that we have some utility functions that are asynchronous. I could re-write the login functions in synchronous way, but that would be mostly implementing some thing twice because testing framework (k6) doesn't support it. So it doesn't sound such a good idea.

Also, in the actual test code (default function) I would like to call our microservice via HTTP API wrapper, not HTTP directly, such as `await serviceClient.createTree(/*some parameters*/);`. That would be more realistic, a bit closer to end-to-end testing. Again, duplicating the code and making it synchronous because k6 doesn't support async code seems a bad idea. Are there any better workarounds?

micsjo:
On the second topic, API wrappers. 

If you have them properly specified you can use tools such as Swagger Codegen to generate a javascript client wrapper for it. If that's how you actually want to build your tests it works fine.

kkeranen:
Thank you @micsjo. Do you have a link to that "API wrappers" documentation? I couldn't find it.

na--:
k6 doesn't currently support promises and `async`/`await` calls, or even simpler things like `setTimeout()`, mostly due to the lack of a global event loop in each VU. That's something we're planning to eventually fix, both because it would improve code reuse and general JS support, but mostly so that we can implement different async APIs that would allow users to user a broader array of JS libraries in k6 tests. 

As briefly described [here](https://github.com/loadimpact/k6#script-execution), the current synchronous k6 execution model mostly works because k6 runs multiple JS [goja](https://github.com/dop251/goja) runtimes in parallel. The current implementation is simpler and it's sufficient for most load testing purposes, but it locks us out of the broader JS ecosystem of tools and libraries, and makes code reuse somewhat difficult as you've found out. As mentioned [here](https://github.com/loadimpact/k6/issues/746#issuecomment-414271374), we have a localized mini-event-loop in the case of websocket connections, but it's very bounded and much more manageable than a global one.

[...]

PR #2319, opened 2022-01-06
Add PromiseRejectionHandler

MStoykov:
This just logs every time a promise gets rejected but has not reject
handler (yet).

The message is basically the one firefox gives.

The simplest variant of #2318

### DIFF
--- a/js/bundle.go
+++ b/js/bundle.go
@@ -307,6 +307,28 @@ func (b *Bundle) instantiate(logger logrus.FieldLogger, rt *goja.Runtime, init *
 	}
 	rt.Set("__ENV", env)
 	rt.Set("__VU", vuID)
+	rt.SetPromiseRejectionTracker(func(p *goja.Promise, op goja.PromiseRejectionOperation) {
+		// TODO this only handles the case where a promise get's rejected without having a reject handler added.
+		// But the other case where for an already rejected promise (one that this call has already handled)
+		// there is now a reject handler attached is not handled.
+		//
+		// More complete implementation might need to wait for some time, the end of an iteration,
+		// some other event before logging a "rejection" that never got "handled"
+		//
+		// Read Notes on https://tc39.es/ecma262/#sec-host-promise-rejection-tracker
+		// see issue https://github.com/grafana/k6/issues/2318 for more info
+		if op == goja.PromiseRejectionReject {
+			frames := rt.CaptureCallStack(20, nil)
+			buf := &bytes.Buffer{}
+			for _, frame := range frames {
+				frame.Write(buf)
+				buf.WriteRune('\n')
+			}
+
+			logger.Warnf("Uncaught (in Promise) %s\n %s", p.Result(), buf.String())
+		}
+	})
+
 	rt.Set("console", common.Bind(rt, newConsole(logger), init.ctxPtr))
 
 	if init.compatibilityMode == lib.CompatibilityModeExtended {
--- a/js/bundle_test.go
+++ b/js/bundle_test.go
@@ -976,3 +976,33 @@ func TestBundleMakeArchive(t *testing.T) {
 		})
 	}
 }
+
+func TestPromiseRejectionHandler(t *testing.T) {
+	t.Parallel()
+	logger := logrus.New()
+	logger.SetLevel(logrus.InfoLevel)
+	logger.Out = ioutil.Discard
+	hook := testutils.SimpleLogrusHook{
+		HookedLevels: []logrus.Level{logrus.WarnLevel, logrus.InfoLevel, logrus.ErrorLevel, logrus.FatalLevel, logrus.PanicLevel},
+	}
+	logger.AddHook(&hook)
+	_, err := getSimpleBundle(t, "/script.js", `
+		export let options = {
+			vus: 5,
+			teardownTimeout: '1s',
+		};
+		let val = true;
+		Promise.reject("here").catch((l) => console.log("caught", l));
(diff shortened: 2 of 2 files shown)


js/bundle.go
```diff
@@ -307,6 +307,28 @@ func (b *Bundle) instantiate(logger logrus.FieldLogger, rt *goja.Runtime, init *
 	}
 	rt.Set("__ENV", env)
 	rt.Set("__VU", vuID)
+	rt.SetPromiseRejectionTracker(func(p *goja.Promise, op goja.PromiseRejectionOperation) {
+		// TODO this only handles the case where a promise get's rejected without having a reject handler added.
```
na--, line 311:
```suggestion
		// TODO this only handles the case where a promise gets rejected without having a reject handler added.
```
Also, something doesn't seem grammatically correct in the bottom 2 lines, but I don't understand exactly what you're trying to say, so I can't offer a correction :confused:

MStoykov: from https://tc39.es/ecma262/#sec-host-promise-rejection-tracker  's Note 1

> Note 1
> 
> HostPromiseRejectionTracker is called in two scenarios:
> 
>   * When a promise is rejected without any handlers, it is called with its operation argument set to "reject".
>   * When a handler is added to a rejected promise for the first time, it is called with its operation argument set to "handle".

We handle only the first one 

the rest of the note is 
> A typical implementation of HostPromiseRejectionTracker might try to notify developers of unhandled rejections, while also being careful to notify them if such previous notifications are later invalidated by new handlers being attached.

this only implements up to the the `,` as ... IMO and IME implementations actually just don't notify you if a promise gets rejected without a handler but then later gets handlers ... for some definition of "later". 

```javascript
var p = somethingReturningAPromise()
// here it has *no* handlers 
// if here it happens to either fulfill or get rejected nothing happens, including if 
// `somethingReturningAPromise` gets an exception  
// if it's rejected though SetPromiseRejectionTracker will be called with `op` equal to "reject"

p.then((a)=>console.log("succes", a)
//^ will add a handler for when the promise is fulfilled *and* when it becomes 
// "fulfilled" it will queue the argument (the arrow function)
// 
// it will *also* queue it if it has already fulfilled,
// even if it has fulfiled before and that already queued something.
// so if we repeat this line *and* the promise is already fulfilled 
// it will get it's argument to `then` queued twice for example

p.catch((bad)=> console.log("promise go rejected")) 
// ^ this will now both call SetPromiseRejectionTracker with "handled"
// and queue the argument to catch *if* the promise was already rejected.
// the same applies as for `.then` if called multiple times and the promise
//  is already rejected it will queue multiple times *by specification* 
```
so for code such as 
```javascript
var p = promiseSomething();
// here p gets rejected somehow :shrug:
// SetPromiseRejectionTracker gets fired and we get a log
p.then((something)=>console.log("cool")
// ^ doesn't queue the body as the promise *is rejected*
// if this is the end, it's really hard to find what goes on as you
//  just made a promise set "then" and if we don't have 
// SetPromiseRejectionTracker, nothing would've even been printed

// if you add
p.catch((b)=>console.log(b)); 
// this will now call SetPromiseRejectionTracker with `op` 
// equal to "handled" and actually queue the provided (arrow) function to be executed
```
With our current code it will execute this code, logging that unhandled exception (in promise) was caught and then when it finished `(b)=>console.log(b)` will be executed, where `b` is the value of whatever was the promise rejected with (possibly an exception). If there were multiple calls for `catch` and they were made *after* the promise was rejected all of them will be queued.  

the really problematic variants are stuff like the ones in https://github.com/grafana/k6/issues/779#issuecomment-1006417364 explained in https://github.com/grafana/k6/issues/779#issuecomment-1006420200

```javascript

const timeout = (ms) => { // this technically is async but it doesn't really matter
  return new Promise((resolve, reject) => {
     // don't ask me why the function is resolve, but the term is fulfilled
    setTimeout(resolve, ms);
    unknwonidentifier.someproperty // some error/exception here !!!!!!!!!!!!!!!!!!!!!!!!! <--- 
  });
};

timeout(12).then(dosomethingCool)
```
will *not* without `SetPromiseRejectionTracker` log you anything as the exception happens inside the Promise and there is no *catch* defined.

p.s. queue here is an internal to *goja* queue that gets ran through when the stack is empty, and before it returns to the host(k6) for example from calls to `RunString`. This also is part of the specification.

js/bundle.go
```diff
@@ -307,6 +307,28 @@ func (b *Bundle) instantiate(logger logrus.FieldLogger, rt *goja.Runtime, init *
...
+		// But the other case where for an already rejected promise (one that this call has already handled)
+		// there is now a reject handler attached is not handled.
+		//
+		// More complete implementation might need to wait for some time, the end of an iteration,
+		// some other event before logging a "rejection" that never got "handled"
```
codebien, line 316: >waiting some time or end of iteration seems like a good idea but it probably needs an "OOM" prevention to not wait on ... **1milltion** unhandled rejections or something like that.

So, is this a real order of magnitude in the case we would track/registry all the "not yet handled" rejection? Seems so high to me but I expect to be my fault in terms of JavaScript world knowledge.

codebien (commented):
Why would we merge a partial solution? Do we expect that it's fixing the most common use-case where this issue is faced?

MStoykov:
In the end this will be part of #2228 with https://github.com/grafana/k6/pull/2228/commits/295ffc09abf938a5588b6286404b819e98cdbbd6

PR #2322, opened 2022-01-10
Close the underlying ws connections after 50 millseconds

mstoykov:
Previous to this the event loop can be blocked on slow operation
-read/write, and not actually see that the context has been canceled.

With this change the underlying connection will be closed 5 milliseconds
after the context is which circumvents the whole event loop. This also
means that the other side will likely not get CloseMessage and that
`close` or `error` handlers will not be called. The latter of which
would've likely not ran either way given that the goja VM would've also
probably be interrupted.

### DIFF
--- a/js/modules/k6/ws/ws.go
+++ b/js/modules/k6/ws/ws.go
@@ -283,9 +283,6 @@ func (mi *WS) Connect(url string, args ...goja.Value) (*WSHTTPResponse, error) {
 	})
 
 	if connErr != nil {
-		// Pass the error to the user script before exiting immediately
-		socket.handleEvent("error", rt.ToValue(connErr))
-
 		return nil, connErr
 	}
 
--- a/js/modules/k6/ws/ws.go
+++ b/js/modules/k6/ws/ws.go
@@ -289,6 +289,36 @@ func (mi *WS) Connect(url string, args ...goja.Value) (*WSHTTPResponse, error) {
 		return nil, connErr
 	}
 
+	// we do it here as below we can panic, which translates to an exception in js code
+	defer func() {
+		socket.Close() // just in case
+		end := time.Now()
+		sessionDuration := stats.D(end.Sub(start))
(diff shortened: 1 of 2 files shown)


js/modules/k6/ws/ws_test.go
```diff
@@ -1317,3 +1320,89 @@ func TestCookieJar(t *testing.T) {
...
+			t.Fatalf("/ws-echo-someheader cannot upgrade request: %v", err)
+		}
+		_, r, err := conn.NextReader() // this is slightly faster than `conn.ReadMessage`
+		if err != nil {
+			ch <- err
+			close(ch)
+			return
+		}
+		n, _ := io.Copy(io.Discard, r)
+		if n == 9000 { // this is exactly how big it should be
```
olegbespalov, line 1341: This condition looks a bit brittle. Should it be exactly 9000?

js/modules/k6/ws/ws.go
```diff
@@ -283,12 +283,39 @@ func (mi *WS) Connect(url string, args ...goja.Value) (*WSHTTPResponse, error) {
...
+	}()
+
+	go func() {
+		// Wait for the context to be over and close the underlying connection directly
+		// This will *not* send CloseMessage or trigger any `close` events but prevents a slow event handler to take forever
+		select {
+		case <-ctx.Done():
+			t := time.NewTimer(time.Millisecond * 50)
+			select {
+			case <-t.C: // wait 5 milliseconds in case the event loop below isn't blocked
```
olegbespalov, line 310: Do we need this comment? Is it five milliseconds?

codebien (commented):
Ping me when this is ready for review

mstoykov:
This is a hacky fix for a very rare situation. 

And given that we will likely replace the current API with something else, for example https://github.com/grafana/k6/issues/2425 it doesn't make much sense to persue this

PR #2323, opened 2022-01-10
chore: relax varnamelen linter

olegbespalov:
Exclude most common short variables from the `varnamelen` linter.

Other possible (count > 5) short variables that can be added to the whitelist:

```bash
$ golangci-lint run | grep "is too short for the scope of its usage" | cut -d' ' -f4 | sort | uniq -c | sort -rn
     50 'tb'
     36 'r1'
     34 'b'
     33 's'
     32 't'
     27 'm'
     27 'et'
     26 'c'
     22 'sr'
     20 'v'
     16 'ok'
     16 'es'
     15 'ts'
     15 'f'
     13 'e'
     12 'k'
     10 'u'
      9 'cm'
      8 'o'
      8 'h'
      7 'mi'
      7 'l'
      6 'p'
      6 'g'
      5 'j'
      5 'd'
```

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -43,6 +43,22 @@ linters-settings:
   funlen:
     lines: 80
     statements: 60
+  varnamelen:
+    ignore-type-assert-ok: true
+    ignore-map-index-ok: true
+    ignore-chan-recv-ok: true
+    ignore-names:
+      - i
+      - wg
+      - fn
+      - ch
+      - fs
+      - tc
+      - r
+      - w
+      - rw
+      - vu
+      - rt
   goheader:
     template-path: ".license-template"
     values:


na--:
I think we can just disable this whole linter... :thinking: Code reviews would catch most of these issues, and if they don't, they probably weren't issues :sweat_smile:

olegbespalov:
>Code reviews would catch most of these issues

@na-- if we automate this, it saves time for both parties, the code owner and reviewer. Isn't it better?

MStoykov:
I agree with @na--. @olegbespalov while in theory it might help, the false positives are so *many* that I doubt there are more than 10-20% where *I* will be like "this probably truly will be better with longer variable names". I certainly can't from the top of my head come up with a case where I was "please rename `r` to `reader` as the scope it's used it's too big". But I definitely have been "can we split this in multiple functions/methods" which will do the same thing and has little to do with variable names being short.

olegbespalov:
@MStoykov, thanks for the input.

>while in theory it might help, the false positives are so many that I doubt there are more than 10-20% where I will be like "this probably truly will be better with longer variable names".

Well, this PR is exactly about reducing the noise. And if something appears later, it seems that we have the chance to adjust the settings again, don't we? Even 10-20% sounds nice, but I bet it will be more keeping in mind the current state of the codebase.

>I certainly can't from the top of my head come up with a case where I was "please rename r to reader as the scope it's used it's too big". But I definitely have been "can we split this in multiple functions/methods" which will do the same thing and has little to do with variable names being short.

Well, the r to the reader is a commonly used case. That's why it's in the list of the exceptions (like many others `wg`, `ch` and so on). On the flip side in our codebase is a list of the 86 lines (attaching it to the end of the message) with the shorties that potentially can create difficulties for the reader.

Yes, you are correct that it's an issue of the same kind as the function length. But I'd argue that if the subject of the function length is resolved, that issue is also gone. It depends on the settings. For example, we now have 80 for the lines and 60 for the statements. The other important thing that we should remember is that the ideal state when all functions will be tiny & lovely probably isn't the near future. The splitting into the smaller pieces of the function can be potentially buggy, that's why probably the devs tend to avoid this. On the other side, the definitive variable names are pretty-safe changes. 

And to keep the point, even the statements like `"can we split this in multiple functions/methods" which will do the same thing and has little to do with variable names being short."` should come not from us, but the linter. We should use our time way more efficiently.

```bash
$ golangci-lint run | grep "is too short for the scope of its usage" | cut -d' ' -f4 | sort | uniq -c | sort -rn
     50 'tb'
     36 'r1'
     36 'm'
     34 's'
     34 'b'
     31 't'
     27 'et'
     26 'c'
     22 'sr'
     20 'v'
     16 'es'
     15 'f'
     14 'ts'
     13 'k'
     13 'e'
     10 'u'
      9 'cm'
      8 'o'
      8 'h'
      7 'mi'
      7 'l'
      6 'n'
      6 'g'
      5 'p'
      5 'j'
      5 'd'
      4 'to'
      4 'th'
      4 'le'
      3 'vm'
      3 'tt'
      3 'ip'
      3 'fd'
      3 'dt'
      3 'bi'
      2 'x'
      2 'ti'
      2 't1'
      2 't0'
      2 'sc'
      2 'r2'
      2 'mt'
      2 'mr'
      2 'md'
      2 'ln'
      2 'js'
      2 'id'
      2 'g1'
      2 'ds'
      2 'ct'
      2 'b1'
      1 'zw'
      1 'vt'
      1 'vi'
      1 've'
      1 'tu'
      1 'tr'
      1 'tm'
      1 't3'
      1 't2'
      1 'T'
      1 'ss'
      1 'sm'
      1 'si'
      1 'sd'
      1 'rc'
      1 'pr'
      1 'pi'
      1 'pf'
      1 'oc'
      1 'nv'
      1 'mp'
      1 'mc'
      1 'lr'
      1 'll'
      1 'lf'
      1 'ir'
      1 'I'
      1 'hk'
      1 'g2'
      1 'g0'
      1 'ex'
      1 'cl'
      1 'ci'
      1 'cb'
      1 'a'
```

MStoykov:
@olegbespalov please make a PR to disable the linter and open an issue to re-enable it. Currently this is blocking PRs that we have already approved but needed to be rebase - I think this is indication enough of how much this linter is helpful, but a discussion can still be had, just not while it's blocking every PR :)

olegbespalov:
>Currently this is blocking PRs that we have already approved but needed to be rebase - I think this is indication enough of how much this linter is helpful

I believe it's just an indicator that ideally such things should be merged lastly or at the beginning of the cycle.

PR #2329, opened 2022-01-12
chore: fix clarity

olegbespalov:
Provide clarity

### DIFF
--- a/js/bundle.go
+++ b/js/bundle.go
@@ -334,15 +334,13 @@ func (b *Bundle) instantiate(logger logrus.FieldLogger, rt *goja.Runtime, init *
 	unbindInit()
 	*init.ctxPtr = nil
 
-	if isInitialInstantiation(vuID) {
+	// If we've already initialized the original VU init context, forbid
+	// any subsequent VUs to open new files
+	if vuID == 0 {
 		init.allowOnlyOpenedFiles()
 	}
 
 	rt.SetRandSource(common.NewRandSource())
 
 	return nil
-}
-
-func isInitialInstantiation(vu uint64) bool {
-	return vu == 0
 }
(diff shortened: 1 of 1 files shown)


na-- (changes requested):
I think a comment would be better than a whole new function, something like
> If we've already initialized the original VU init context, forbid any subsequent VUs to open new files

js/bundle.go
```diff
@@ -334,11 +334,15 @@ func (b *Bundle) instantiate(logger logrus.FieldLogger, rt *goja.Runtime, init *
 	unbindInit()
 	*init.ctxPtr = nil
 
-	if vuID == 0 {
+	if isInitialInstantiation(vuID) {
```
MStoykov, line 337: if you move this code to the top of this function it will still pass and break all openings of files ;). This is because the new function checks that we are in the initial VU not that we have initialized. 

It just so happens that it is in part of the code that it has already initialized it, and we check that we are in the first VU so we don't call it for each one. 

Given the current code around this. we can just as well call `allowOnlyOpenedFiles` on each VU I guess, but maybe preferable a comment explaining that we have initialized, and so we no longer want to be able to open new ones in other VUs.

olegbespalov: Yes, that's true, but is there any reason we should move it? :thinking: 

The current approach isn't ideal, to be honest, I don't like it either, but honestly believe that calling `allowOnlyOpenedFiles` on each VU user will make this logic more complicated than just an explanation with the comment.

MStoykov: > Yes, that's true, but is there any reason we should move it? thinking

Sorry I didn't mean that you *should* move it, just that given the these three lines you would think that if you move it to the top it will *not* run, but it will ;)

yorugac (approved):

na-- (approved):

MStoykov (approved):

PR #2331, opened 2022-01-12
Fix all linter errors in loader package

MStoykov:
This removes 67 linter errors, does not (seem) to introduce any races in
the tests and it also seems to not improve their runtime as well :(

The changes were done with the idea of "as little change as possible" so
that there should be no bugs introduced.


codebien (approved):
LGTM

oleiade (approved):
Sehr Gut 👍🏻

PR #2337, opened 2022-01-14
Use envconfig fork to abstract os package

MStoykov:
Previous to this the envconfig package directly called os.LookupEnv.
With this it also takes a function that can replace this call giving us
the ability to not use for example map[string]string for the backing
instead of having to touch the global environment in tests.


MStoykov:
I didn't use `replace` and instead changed the module path in the fork because:
1. I preferred it as it looks cleaner
2. but also i figured out later that otherwise this will break xk6 as `replace` clauses are only taken from the final module - if one of our dependency has `raplace` that won't have any effect, we will need to add the same `replace` in order for it to work.

Luckily, the changes for the most part are pretty small ... so that isn't really a problem.

The two other problems with envconfig seem to be more hairy so at least for now I don't plan on fixing them, but I am taking PRs :P

MStoykov:
[This](https://github.com/MStoykov/envconfig/commit/765c6d8c76f1929fab47961fa009e13a3d07ab23) is the actual envconfig change for which it was forked. It's backwards compatible *on purpose*, but we can change that if we decide in the future. The only other change is [renaming the module](https://github.com/MStoykov/envconfig/commit/e8a89633b9af4b2d26b1ad155e3c0a6cba2878ed).

cmd/config_consolidation_test.go
```diff
@@ -582,7 +579,9 @@ func runTestCase(
 		testCase.options.fs = afero.NewMemMapFs() // create an empty FS if it wasn't supplied
 	}
 
-	consolidatedConfig, err := getConsolidatedConfig(testCase.options.fs, cliConf, runnerOpts)
+	consolidatedConfig, err := getConsolidatedConfig(testCase.options.fs, cliConf, runnerOpts,
+		// TODO: just makes testcase.options.env in map[string]string
```
na--, line 583:
```suggestion
		// TODO: just make testcase.options.env in map[string]string
```

na-- (approved):
LGTM! One step closer... :relaxed:

yorugac:
@MStoykov test-tip failed here: https://github.com/grafana/k6/runs/4938790631 ? That's not the flaky one it seems :confused:

MStoykov:
> That's not the flaky one it seems 

It is, it just fails really rarely.

yorugac (approved):

PR #2339, opened 2022-01-17
Add source map samples

MStoykov:


oleiade:
Nice! This is sure going to help with #2340. Could you give me a bit of context to the context in K6 in which this would be useful? My understanding of SourceMaps is that they're essentially a mechanism allowing to map from minified code, to its original source, correct?

oleiade (approved):

MStoykov:
>  mechanism allowing to map from minified code,

if you replace "minified" with "transformed" you are there. 

For k6 specifically as we still use babel internally we do transform code that supposedly k6 "supports", which is what the issue_1804 is showing. The other one is showing an embeded source map(The last line in test1.js) generated by tsc, babel and webpack. It also happens that the particular example *still* needs the k6 internal babel. So that sourcemap does actually go through the internal babel and it returns a "combined" map that gets you back (somewhat) to the original sourcecode, which is test1.ts.

yorugac (approved):

Issue #2296, opened 2021-12-14
a Roadmap to target higher ES standards? ES6+ ES7+ ES8+ or ES2020 ES2021+ ?


tx0c:
### Feature Description

the [main feature of k6](https://github.com/grafana/k6#features) still says ES6, but ES6 has existed long time, some features are not cool enough at the end of 2021, and we have been using many newer JS features for years, like this [`optional chaining operator (?.)`](https://node.green/#ES2020-features-optional-chaining-operator-----) with babel or natively since 2020;

wish can write tests in this way, checking the response header should have `'content-type'` and should string `includes` `'application/json'`

```js
  check(res, {
    'header has content-type application/json': r => r.headers['content-type'] ?. includes('application/json')
  });
```

instead of writing it the old way:

```js
  check(res, {
    'header has content-type application/json': r => r.headers['content-type'] && r.headers['content-type'].indexOf('application/json') > -1
  });
```

wonder what JS engine is builtin this `k6` tool?  can that bump version to target a higher version of JS standard?

many things like `URL`, `URLSearchParams` have been builtin Node.js for years;

and have to import it is not cool
https://k6.io/docs/examples/urls-with-query-parameters/

### Suggested Solution (optional)

many things like `URL`, `URLSearchParams` have been builtin Node.js for years;

and have to import it is not cool
https://k6.io/docs/examples/urls-with-query-parameters/

na--:
We'd like to have all of that! Unfortunately it's not easy to keep up with the speed of the JS ecosystem... :sweat_smile:

I'll try to provide some background of how JS execution actually happens in k6. After all, k6 itself is written in Go, so we can't just immediately get all of the new features from V8 or node.js... Instead, k6 uses [goja](https://github.com/dop251/goja) to parse and execute scripts, a JavaScript runtime written purely in Go. It is not as fast as other more optimized runtimes and it doesn't support all of the latest JS features, but there are a lot of reasons for us to use it despite these shortcomings!

[...]

PR #2345, opened 2022-01-18
Disable sourcemaps support in babel if input is big

MStoykov:
This prevents babel from using up a lot of memory (3.5x as much as
before source maps) on really big inputs.

### DIFF
--- a/js/compiler/compiler.go
+++ b/js/compiler/compiler.go
@@ -149,7 +149,7 @@ func (c *Compiler) Transform(src, filename string, inputSrcMap []byte) (code str
 	if sourceMapEnabled && len(src) > maxSrcLenForBabelSourceMap {
 		sourceMapEnabled = false
 		c.logger.Warnf("the source for `%s` needs to go through babel but is over %d bytes. "+
-			"For performance reasons sourcemaps support will be disabled for this particular file.",
+			"For performance reasons source map support will be disabled for this particular file.",
 			filename, maxSrcLenForBabelSourceMap)
 	}
 
--- a/js/compiler/compiler.go
+++ b/js/compiler/compiler.go
@@ -24,7 +24,6 @@ import (
 	_ "embed" // we need this for embedding Babel
 	"encoding/json"
 	"errors"
-	"fmt"
 	"os"
 	"strconv"
 	"sync"
@@ -36,25 +35,6 @@ import (
 
 	"go.k6.io/k6/lib"
 )
-
-//nolint:gochecknoglobals
-var maxSrcLenForBabelSourceMap = 250 * 1024
-
-const maxSrcLenForBabelSourceMapVarName = "K6_DEBUG_SOURCEMAP_FILESIZE_LIMIT"
-
-// TODO: drop this code and everything it's connected to when babel is dropped
-func init() {
-	v := os.Getenv(maxSrcLenForBabelSourceMapVarName)
-	if len(v) > 0 {
-		i, err := strconv.Atoi(v)
-		if err != nil {
-			//nolint:forbidigo
-			fmt.Printf("Tried to parse %q from %s as integer but couldn't %s\n",
-				v, maxSrcLenForBabelSourceMapVarName, err)
-		}
-		maxSrcLenForBabelSourceMap = i
-	}
-}
 
 //go:embed lib/babel.min.js
 var babelSrc string //nolint:gochecknoglobals
@@ -103,6 +83,9 @@ var (
 		"highlightCode": false,
 	}
 
+	maxSrcLenForBabelSourceMap     = 250 * 1024 //nolint:gochecknoglobals
+	maxSrcLenForBabelSourceMapOnce sync.Once    //nolint:gochecknoglobals
+
 	onceBabelCode      sync.Once     // nolint:gochecknoglobals
 	globalBabelCode    *goja.Program // nolint:gochecknoglobals
 	globalBabelCodeErr error         // nolint:gochecknoglobals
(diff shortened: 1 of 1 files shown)


js/compiler/compiler.go
```diff
@@ -34,6 +34,8 @@ import (
 	"go.k6.io/k6/lib"
 )
 
+const maxSrcLenForBabelSourceMap = 500 * 1024 // 500kb
```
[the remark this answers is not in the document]
MStoykov: Unfortunately, if you go through the code you will see that `onceBabel.Do` is called only if the compiler doesn't have its own babel set – it's used to set the global babel. 

While this is currently true for the case we care about, I prefer to not have to find out this get broken at somepoint just to use the logger for a thing, we want to drop in a bit as well.

js/compiler/compiler.go
```diff
@@ -34,6 +37,25 @@ import (
 	"go.k6.io/k6/lib"
 )
 
+//nolint:gochecknoglobals
+var maxSrcLenForBabelSourceMap = 250 * 1024
+
+const maxSrcLenForBabelSourceMapVarName = "K6_DEBUG_SOURCEMAP_FILESIZE_LIMIT"
+
+//  drop this code and everything it's connected to when goja is dropped
```
na--, line 45:
```suggestion
// TODO: drop this code and everything it's connected to when babel is dropped
```

js/compiler/compiler.go
```diff
@@ -34,6 +37,25 @@ import (
...
+func init() {
+	v := os.Getenv(maxSrcLenForBabelSourceMapVarName)
+	if len(v) > 0 {
+		i, err := strconv.Atoi(v)
+		if err != nil {
+			//nolint:forbidigo
+			fmt.Printf("Tried to parse %q from %s as integer but couldn't %s\n",
```
na--, line 52: At least use `log` here, since we redirect its output to our own logger in https://github.com/grafana/k6/blob/13b2372fd98fb3fefe103a1d22f212164dff9107/cmd/root.go#L138

MStoykov: doesn't really matter as this happens in hte `init` before that line is called

MStoykov: I can add one more `sync.Once` at this point :shrug:

na--: :+1: for either another `sync.Once` instead of `init()`, or for just silently ignoring any parse error

js/compiler/compiler.go
```diff
@@ -124,7 +132,28 @@ func (c *Compiler) Transform(src, filename string, inputSrcMap []byte) (code str
...
+					v, maxSrcLenForBabelSourceMapVarName, err)
+				return
+			}
+			maxSrcLenForBabelSourceMap = i
+		}
+	})
+	if sourceMapEnabled && len(src) > maxSrcLenForBabelSourceMap {
+		sourceMapEnabled = false
+		c.logger.Warnf("the source for `%s` needs to go through babel but is over %d bytes. "+
+			"For performance reasons sourcemaps support will be disabled for this particular file.",
```
codebien, line 152: Should we mention `K6_DEBUG_SOURCEMAP_FILESIZE_LIMIT`?

js/compiler/compiler.go
```diff
@@ -124,7 +132,28 @@ func (c *Compiler) Transform(src, filename string, inputSrcMap []byte) (code str
...
+				c.logger.Warnf("Tried to parse %q from %s as integer but couldn't %s\n",
+					v, maxSrcLenForBabelSourceMapVarName, err)
+				return
+			}
+			maxSrcLenForBabelSourceMap = i
+		}
+	})
+	if sourceMapEnabled && len(src) > maxSrcLenForBabelSourceMap {
+		sourceMapEnabled = false
+		c.logger.Warnf("the source for `%s` needs to go through babel but is over %d bytes. "+
```
codebien, line 151:
```suggestion
		c.logger.Warnf("The source for `%s` needs to go through babel but is over %d bytes. "+
```

js/compiler/compiler.go
```diff
@@ -124,7 +132,28 @@ func (c *Compiler) Transform(src, filename string, inputSrcMap []byte) (code str
 		return
 	}
 
-	code, srcMap, err = c.babel.transformImpl(c.logger, src, filename, c.Options.SourceMapLoader != nil, inputSrcMap)
+	sourceMapEnabled := c.Options.SourceMapLoader != nil
+	maxSrcLenForBabelSourceMapOnce.Do(func() {
+		// TODO: drop this code and everything it's connected to when babel is dropped
```
codebien, line 137: Do we have an issue to link?

js/compiler/compiler.go
codebien, line 152:
```suggestion
			"For performance reasons source map support will be disabled for this particular file.",
```
for consistency with other logs

codebien (changes requested):

js/compiler/compiler.go
na--: https://github.com/grafana/k6/issues/2296#issuecomment-993356717 maybe? :thinking:

codebien: mmm...if we don't have one yet, I'm for creating one dedicated with the steps we plan for dropping babel. For example, it would be useful to have somewhere a link to https://github.com/dop251/goja/issues/348 and this potential new issue could be a good candidate, IMHO.

olegbespalov (approved):
LGTM :+1:

codebien (approved):
LGTM

js/compiler/compiler.go
na--: I think we shouldn't mention this, for a couple of reasons:
- `K6_DEBUG_SOURCEMAP_FILESIZE_LIMIT=whatever k6 cloud` won't work
- we intend to get rid of this option eventually

So we shouldn't really document this. It's a temporary option until we get rid of Babel. We can always mention it and the caveats if someone asks in the forum.

PR #2346, opened 2022-01-18
Sort struct fields

codebien:
Enabled the `fieldalignment` analyzer for `golanci` and executed the same command with the `-fix` flag for applying automatically all the suggested fixes for the current detected issues.

I still have to review the changes and profiling the new codebase to check that we are not generating any type of regression.


cloudapi/config.go
```diff
@@ -33,128 +33,28 @@ import (
...
-	// won't make much sense... But this would allow us to select tighter sample groups for
-	// aggregation if we want.
-	AggregationOutlierIqrRadius null.Float `json:"aggregationOutlierIqrRadius" envconfig:"K6_CLOUD_AGGREGATION_OUTLIER_IQR_RADIUS"`
-
-	// Connection or request times with how many IQRs below Q1 to consier as non-aggregatable outliers.
-	AggregationOutlierIqrCoefLower null.Float `json:"aggregationOutlierIqrCoefLower" envconfig:"K6_CLOUD_AGGREGATION_OUTLIER_IQR_COEF_LOWER"`
-
-	// Connection or request times with how many IQRs above Q3 to consier as non-aggregatable outliers.
-	AggregationOutlierIqrCoefUpper null.Float `json:"aggregationOutlierIqrCoefUpper" envconfig:"K6_CLOUD_AGGREGATION_OUTLIER_IQR_COEF_UPPER"`
+	PushRefID                       null.String        `json:"pushRefID" envconfig:"K6_CLOUD_PUSH_REF_ID"`
```
MStoykov, line 36: Dropping the comments here is a big no ;)

codebien: I see the same problem in several changes, I need to restore them. Hopefully, they aren't a problem for the real size.

na--: Also, in general, I am not sure we should optimize these config fields for size, that probably won't help a lot with the actual k6 performance... Rather, we should optimize them for human readability, grouping related fields together...

codebien: This is why I would get a benchmark before do any progress here. Because, if we don't get any improvement then we can just close this PR without any addition.

output/json/json_test.go
```diff
@@ -47,7 +47,7 @@ func getValidator(t *testing.T, expected []string) func(io.Reader) {
 				t.Errorf("Read unexpected line number %d, expected only %d entries", i, len(expected))
 				continue
 			}
-			assert.Equal(t, expected[i-1], string(s.Bytes()))
+			assert.JSONEq(t, expected[i-1], string(s.Bytes()))
```
codebien, line 50: It checks just if the object contains the same keys and values but not the order as before. If you think the order here is relevant then we can restore it.

olegbespalov (approved):
:crossed_fingers:

na-- (changes requested):
Sorry, but I think most of these are unnecessary and counter-productive. For example, `commandFlags` (which I rename to `globalState` in my most recent PRs). 

It's always passed as a pointer, a slight inefficiency of the struct field alignment doesn't matter one bit, since that variable basically has a single copy... But it matters quite a lot that `stdout` is close to both `outMutex` and `stdoutTTY` for comprehension and for preventing errors. So, :-1: for that specific change.

Same for `ExecutionScheduler`, `Bundle` ,`ActiveVU` and a bunch of others. The ease of comprehension by grouping related properties close to each other is orders of magnitude more important than saving a few bytes. Basically, the only place where this isn't the case is where we have potentially millions of objects, e.g. with metric samples and possible HTTP responses, etc.

MStoykov:
I more or less agree with @na-- .

I am fine with a lot of the reordering but I guess they will need to be made by hand in order to not break comments in a lot of the cases. 

Also as mentioned in most places it makes no sense to safe a few bytes as we have very few copies of that struct, *but* the order of the arguments matters inside the struct for readability. 

So maybe:
1. Try only the changes to the metrics/sample types  or outputs ones (I see the cloud has changes) and benchmark those. If there is an improvement - let's try to merge those. (There are some changers since then so that might be harder). 
2. If there is any improvement, we can try to enable the linter and you can go case by case and move some of the other types that aren't loosing readability to the more compact variant.
3. We can nolint the rest. 

I would expect `1.` to have some pretty big and easy to see benefit ... if not it's likely 2. and 3. will take too much time making this not very good idea to begin with. But again I do expect we will see significant improvement for the cases where it matters. And if that is the case we can nolint the places where it wouldn't and where we want to keep them more readable.

codebien:
Based on the previous comment and the current conflict status, I will do the new work to a new PR.

PR #2350, opened 2022-01-24
Update goja

MStoykov:


js/modules/k6/ws/ws.go
```diff
@@ -123,24 +123,19 @@ func (mi *WS) Exports() modules.Exports {
...
-	var callableV, paramsV goja.Value
-	switch len(args) {
-	case 2:
-		paramsV = args[0]
-		callableV = args[1]
-	case 1:
-		paramsV = goja.Undefined()
-		callableV = args[0]
-	default:
+	if callableV == nil {
```
codebien, line 134:
```suggestion
    // The params argument is optional
	if callableV == nil {
```
I'm not sure the next time I will be able to understand **why** we are doing it. So, I would prefer to maintain the comment explaining it.

codebien (changes requested):
Just a small suggestion

olegbespalov:
LGTM :+1:

olegbespalov (dismissed):

js/modules/k6/ws/ws.go
MStoykov: the reason for this changes was fixed upstream so I am dropping them

olegbespalov (approved):

codebien (approved):
LGTM

PR #2353, opened 2022-01-26
Cmd no globals

MStoykov:
edit: go commit by commit otherwise it's very hard to follow what happens IMO
edit2: I have optimized for *small* change so the code review can be done faster and also to not break anything. The final code is still pretty terrible for many ... many reasons, but fixing those is IMO outside the scope of this PR. 

edit3: this was split so that the smaller changes are now in #2357  and #2358. And this PR is based on #2358 as it kind of depends on it to run the tests in parallel

### DIFF
--- a/cmd/config_consolidation_test.go
+++ b/cmd/config_consolidation_test.go
@@ -136,6 +136,8 @@ func mostFlagSets() []flagSetInit {
 	result := []flagSetInit{}
 	for i, fsi := range []func(globalFlags *commandFlags) *pflag.FlagSet{runCmdFlagSet, archiveCmdFlagSet, cloudCmdFlagSet} {
 		i, fsi := i, fsi // go...
+		// TODO: this still uses os.GetEnv which needs to be removed
+		// before/along adding tests for those fields
 		root := newRootCommand(context.Background(), nil, nil)
 		result = append(result, func() (*pflag.FlagSet, *commandFlags) {
 			flags := pflag.NewFlagSet(fmt.Sprintf("superContrivedFlags_%d", i), pflag.ContinueOnError)
--- a/cmd/config_consolidation_test.go
+++ b/cmd/config_consolidation_test.go
@@ -172,11 +172,6 @@ type opts struct {
 	// actually will change those variables to their default values :| In our
 	// case, this happens only some of the time, for global variables that
 	// are configurable only via CLI flags, but not environment variables.
-	//
-	// For the rest, their default value is their current value, since that
-	// has been set from the environment variable. That has a bunch of other
-	// issues on its own, and the func() doesn't help at all, and we need to
-	// use the resetStickyGlobalVars() hack on top of that...
 	cliFlagSetInits []flagSetInit
 }
 
@@ -623,10 +618,8 @@ func runTestCase(
 	}
 }
 
-//nolint:paralleltest // see comments in test
 func TestConfigConsolidation(t *testing.T) {
-	// This test and its subtests shouldn't be ran in parallel, since they unfortunately have
-	// to mess with shared global objects (variables, ... santa?)
+	t.Parallel()
 
 	for tcNum, testCase := range getConfigConsolidationTestCases() {
 		tcNum, testCase := tcNum, testCase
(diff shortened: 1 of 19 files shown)


olegbespalov:
A PR looks good :+1: 

However, I have a small readability improvement point. What if we use a more explicit name instead of the `fl`?

At some point, we (and other people) are going to read it again and ideally, we should limit the number of the shortcuts like that.

And sometimes it looks not consistent, example:
```go
outputs, err := createOutputs(conf.Out, src, conf, runtimeOptions, executionPlan, osEnvironment, logger, fl)
```

MStoykov:
> However, I have a small readability improvement point. What if we use a more explicit name instead of the fl?

as long as you come up with something. I guess part of the review process will be coming up with names about `commandFlags` and `fl` as both are ... bad.

Also I might break this in 3 PRs :
1. the first commit here
2. the last commit here without the paralization 
3. everything else - which IMO has the most to be discussed about 

WDYT?

cmd/root.go
```diff
@@ -45,35 +45,58 @@ import (
...
+	isDumbTerm := os.Getenv("TERM") == "dumb"
+	stdoutTTY := !isDumbTerm && (isatty.IsTerminal(os.Stdout.Fd()) || isatty.IsCygwinTerminal(os.Stdout.Fd()))
+	stderrTTY := !isDumbTerm && (isatty.IsTerminal(os.Stderr.Fd()) || isatty.IsCygwinTerminal(os.Stderr.Fd()))
+	outMutex := &sync.Mutex{}
+	return &commandFlags{
+		defaultConfigFilePath: defaultConfigFilePath,  // Updated with the user's config folder in the init() function below
+		configFilePath:        os.Getenv("K6_CONFIG"), // Overridden by `-c`/`--config` flag!
+		exitOnRunning:         os.Getenv("K6_EXIT_ON_RUNNING") != "",
+		showCloudLogs:         true,
+		runType:               os.Getenv("K6_TYPE"),
```
MStoykov, line 91: All of those os.Getenv should go, but as currently this are the only places where they are and there are zero tests covering this I decided against introducing even *more* changes.

cmd/root.go
```diff
@@ -45,35 +45,58 @@ import (
...
-var configFilePath = os.Getenv("K6_CONFIG") // Overridden by `-c`/`--config` flag!
-
-//nolint:gochecknoglobals
-var (
-	// TODO: have environment variables for configuring these? hopefully after we move away from global vars though...
-	quiet   bool
-	noColor bool
-	address string
-)
+// TODO better name - there are other commadn flags these are just ... non lib.Options ones :shrug:
```
codebien, line 53:
```suggestion
// TODO better name - there are other command flags these are just ... non lib.Options ones :shrug:
```

codebien (approved):
an amount of code that I hope at some point would be rid by croconf, just spot one tiny typo

MStoykov:
I've update the name and have split this in 3 PRs so this is only about the flags

olegbespalov (dismissed):
Sorry for the delay, PR looks good to me! :muscle:

MStoykov:
> Sorry for the delay

Don't worry, I actually will be waiting for @na-- to give a :+1: as well as the person who probably is most familiar with all the things that can go wrong in this code

cmd/config_consolidation_test.go
```diff
@@ -131,13 +134,14 @@ func mostFlagSets() []flagSetInit {
 	// getConsolidatedConfig() is used, but they also have differences in their CLI flags :/
 	// sigh... compromises...
 	result := []flagSetInit{}
-	for i, fsi := range []flagSetInit{runCmdFlagSet, archiveCmdFlagSet, cloudCmdFlagSet} {
+	for i, fsi := range []func(globalFlags *commandFlags) *pflag.FlagSet{runCmdFlagSet, archiveCmdFlagSet, cloudCmdFlagSet} {
 		i, fsi := i, fsi // go...
-		result = append(result, func() *pflag.FlagSet {
+		root := newRootCommand(context.Background(), nil, nil)
```
na--, line 139: hmm this will still do `os.GetEnv()` for some things :confused: so maybe add a short comment in this test file with an explanation that those can't be fully tested yet and a `TODO` to test them in the future, when they can?

MStoykov:
```suggestion
                // TODO: this still uses os.GetEnv which needs to be removed 
                // before/along adding tests for those fields 
		root := newRootCommand(context.Background(), nil, nil)
```
?

na-- (dismissed):
LGTM, besides the minor nitpick for the comment! :tada: This is definitely not the final state these should be in, but a great step in the right direction! :heart:

cmd/config_consolidation_test.go
na--: Seems good :+1:

na--: Erm I think you messed up the indentation and used spaces instead of tabs

MStoykov: :facepalm:

codebien (approved):

na-- (approved):

PR #2355, opened 2022-01-27
Try to load and parse the sourcemap before returning it

MStoykov:
This should hopefully fix most of the cases where a source map that
is tried to be loaded but turns out to be gibberish doesn't stop the
whole execution of k6.

edit: this also fixes the error message about source map not being loaded, when goja also has other errors. Previous it would tell you stuff such as 
```
WARN[0000] Couldn't load source map for file:///home/mstoykov/work/k6io/k6/l.js  error="file:///home/mstoykov/work/k6io/k6/l.js: Line 1:1 Unexpected reserved word (and 1 more errors)"
```
when for example the source map wasn't loaded properly but also babel needs to be invoked in order to transpile import/export for example.

### DIFF
--- a/js/compiler/compiler.go
+++ b/js/compiler/compiler.go
@@ -157,7 +157,7 @@ func (c *Compiler) Transform(src, filename string, inputSrcMap []byte) (code str
 
 	// check that babel will likely be able to parse the inputSrcMap
 	if sourceMapEnabled && len(inputSrcMap) != 0 {
-		if err := verifySourceMapForBabel(inputSrcMap); err != nil {
+		if err = verifySourceMapForBabel(inputSrcMap); err != nil {
 			sourceMapEnabled = false
 			inputSrcMap = nil
 			c.logger.WithError(err).Warnf(
@@ -421,25 +421,19 @@ func verifySourceMapForBabel(srcMap []byte) error {
 	if err != nil {
 		return fmt.Errorf("source map is not valid json: %w", err)
 	}
-	if v, ok := m["version"]; !ok {
+	// there are no checks on it's value in babel
+	// we technically only support v3 though
+	if _, ok := m["version"]; !ok {
 		return fmt.Errorf("source map missing required 'version' field")
-	} else {
-		// there are no checks on it's value in babel
-		// we technically only support v3 though
-		_ = v
 	}
 
-	// This actually gets checked by the go implementation
-	if v, ok := m["mappings"]; !ok {
+	// This actually gets checked by the go implementation so it's not really necessary
+	if _, ok := m["mappings"]; !ok {
 		return fmt.Errorf("source map missing required 'mappings' field")
-	} else {
-		_ = v
 	}
-	if v, ok := m["sources"]; !ok {
+	// the go implementation checks the value even if it doesn't require it exists
+	if _, ok := m["sources"]; !ok {
 		return fmt.Errorf("source map missing required 'sources' field")
-	} else {
-		// the go implementation checks the value even if it doesn't require it exists
-		_ = v
 	}
 	return nil
 }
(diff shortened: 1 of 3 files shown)


js/compiler/compiler_test.go
```diff
@@ -203,5 +203,58 @@ func TestCorruptSourceMap(t *testing.T) {
...
 
-	require.Contains(t, msg, `Could not load source map: missing \"mappings\" in sourcemap`)
+	require.Contains(t, msg, `Couldn't load source map for somefile`)
+	require.Contains(t, msg, `json: cannot unmarshal number into Go struct field v3.mappings of type string`)
+}
+
+func TestCorruptSourceMapOnlyForBabel(t *testing.T) {
+	t.Parallel()
+	// this a valid source map for the go implementation but babel doesn't like it
+	corruptSourceMap := []byte(`{"mappings": ";"}`)
```
MStoykov, line 213: Fixing this will mean that we will be running babel possibly twice on every error it returns. 
I find it somewhat unlikely that if we have a json with `mappings` that it will not have the other  source map fields. Also hopefully we well really even less on babel soon.

js/compiler/compiler_test.go
```diff
@@ -203,5 +203,58 @@ func TestCorruptSourceMap(t *testing.T) {
...
+	_, _, err := compiler.Compile("class s {};\n//# sourceMappingURL=somefile", "somefile", false)
+	// Ideally we will figure out a way to not error in this case
+	require.Error(t, err)
+	require.Equal(t, err.Error(), `Error: somefile: "version" is a required argument. at <internal/k6/compiler/lib/babel.min.js>:2:28536(109)`)
+}
+
+func TestMinimalSourceMap(t *testing.T) {
+	t.Parallel()
+	// this is the minimal sourcemap valid for both go and babel implementations
+	corruptSourceMap := []byte(`{"version": 3, "mappings": ";", "sources": []}`)
```
codebien, line 240:
```suggestion
	corruptSourceMap := []byte(`{"version":3,"mappings":";","sources":[]}`)
```
one more step in minimalism :)

codebien (commented):
LGTM, just one tiny thing

js/compiler/compiler_test.go
MStoykov: already did it in 82a519b

js/compiler/compiler_test.go
MStoykov: 82a519b hopefully catches all the cases :crossed_fingers:

js/compiler/compiler.go
```diff
@@ -395,3 +413,27 @@ func (c *Pool) Get() *Compiler {
 func (c *Pool) Put(co *Compiler) {
 	c.c <- co
 }
+
+func verifySourceMapForBabel(srcMap []byte) error {
+	// this function exists to do what babel checks in sourcemap before we give it to it.
+	m := make(map[string]interface{})
```
na--, line 419:
```suggestion
	m := make(map[string]json.RawMessage)
```
we don't really care about the values, so this should be more efficient, right/

js/compiler/compiler.go
```diff
@@ -153,6 +155,16 @@ func (c *Compiler) Transform(src, filename string, inputSrcMap []byte) (code str
...
 	}
 
+	// check that babel will likely be able to parse the inputSrcMap
+	if sourceMapEnabled && len(inputSrcMap) != 0 {
+		if err = verifySourceMapForBabel(inputSrcMap); err != nil {
+			sourceMapEnabled = false
+			inputSrcMap = nil
+			c.logger.WithError(err).Warnf(
+				"The source for `%s` needs to go through babel, but it's source map will"+
+					" not be accepted by babel, so it gets disabled", filename)
```
na--, line 165:
```suggestion
				"The source for `%s` needs to be transpiled by Babel, but its source map will"+
					" not be accepted by Babel, so it was disabled", filename)
```

js/compiler/compiler.go
MStoykov: :shrug: dobut the performance of this will be interesting compared to the performance of parsing it by babel, or the source file by babel. But I guess it won't hurt :+1:

codebien (approved):
LGTM, remind to rebase or squash before merging

na-- (approved):

Issue #2330, opened 2022-01-12
Parse and validate thresholds before starting the execution


oleiade:
## Problem statement

We would like to parse and validate threshold expressions at the end of the *init context*'s evaluation, in order to have invalid thresholds make K6 fail early.

### Context

As of #2251 (addressing #1443), thresholds are parsed and executed in Go, as opposed to prior to that, when the operation was done in JS. Still, threshold evaluation happens during the running phase of K6. During the review of #2251, @na-- [mentioned](https://github.com/grafana/k6/pull/2251/files#r782875298) that as a follow-up, we probably should move the parsing and validation of thresholds right after the initial evaluation of the *init context*.  That way, if thresholds expression don't parse, or parse into a case we know to be an error (see Scope), we may stop right at the end of the initialization phase and have K6 fail early.

As of today, parsing a threshold expression is done using the `stats.parseThresholdExpression` function. Parsing is made in `stats.newThreshold`. Executing the threshold, and comparing using it against the actual numbers we've collected (`sinks`) is done in `stats.Threshold.runNoTaint`.

### Scope

The goal would then be for our changes to halt the execution of K6, right after the evaluation of the *init context*, in the following scenarios.

### Invalid threshold expression

```javascript
export const options = {
    vus: 5,
    iterations: 30,
    thresholds: {
         http_req_failed: ['rave<0.01'],  // "rave" is not a valid aggregation method
    },
};
```

### Invalid aggregation method for metric

```javascript
export const options = {
    vus: 5,
    iterations: 30,
    thresholds: {
        my_counter: ['p(95)<200'], // error, Counters do not have p()
    },
};
```

### Threshold on non-existing metric

```javascript
export const options = {
    vus: 5,
    iterations: 30,
    thresholds: {
        iDoNotExist: ['p(95)<200'], // error, the metric it is applied to doesn't exist
    },
};
```

## Requirements

- When an invalid threshold is found, `k6` should exit with the code `104` [InvalidConfig](https://github.com/grafana/k6/blob/master/errext/exitcodes/codes.go#L35)

## Proposed change

Although it's still unclear to me at this point exactly what the concrete solution should look like, the rationale would be to find the right place in the evaluation of the init context (somewhere where the metrics registry is available), to use `stats.parseThresholdExpression`.

## Open questions

While we believe this change's scope shouldn't be very big, and don't expect complications, the following questions remain were raised or remain open until implementation:
[...]

PR #2356, opened 2022-01-27
Parse and validate threshold during init

oleiade:
## What this PR does

This Pull Request addresses the scope described in #2330. Namely, it reintroduces the changes from #1443/#2251, and builds upon them to ensure that thresholds are parsed and validated before the execution starts.

### What it does

Following this Pull Request:
- k6 parses thresholds during [bundling](https://github.com/grafana/k6/blob/master/cmd/run.go#L115). Parsing makes sure that the thresholds defined in the options make sense from a syntactic point of view. 
- k6 validates thresholds during [configuration validation](https://github.com/grafana/k6/blob/master/cmd/run.go#L131). There, it makes sure that thresholds apply to existing metrics, and that the aggregation method they hold is supported by the metric they apply to.
- If any of the conditions described above were not met, k6 would return immediately with an exit code `104 - Invalid Config`.

### What it does not

It was discussed in the context of the #2330 issue that k6 should also make sure to handle metrics properly when in an archiving/cloud execution context. Some changes were [proposed](https://github.com/grafana/k6/issues/2330#issuecomment-1013066839) to how k6 handles the `metadata.json` file.
In order to keep this Pull Request reviewable and avoid scope creep, I decided to address that issue separately. If you have arguments towards doing otherwise: shoot 🚀.

## Notable changes

The most notable change I can think of is that `deriveAndValidateConfig` now takes a `metrics.Registry` as input. Another important one is that threshold parsing code now declares an explicit sentinel error type `ErrThresholdParsing` which the caller code depends on in order to decide which exit code it should use.

## Review guide

- This PR reverts the revert of the #2251 merge (that's a mouthful). Thus, please ignore the first commit of this branch, as it was already reviewed and accepted as part of the initial PR.
- 🔥 I would also recommend using the commits view of GitHub, as opposed to Files Changed, in order to be able to avoid the overhead of reviewing threshold parsing code again 😉 
- I did my best to arrange and order the commits logically. I've also included extensive commits messages, which should help you through them:
  1. First: The very first brings back Go-based thresholds parsing.
  2. Second: Some changes to the public API of thresholds parsing has been made so that certain symbols could be accessed from outside the `stats` package.
  3. Third: Contains the code implementing thresholds validation against metrics.
  4. Fourth: Integrates thresholds validation in the affected commands.

## Dependency

ref #2330
ref #1443
ref #2251

Please let me know if any more information is needed on your side, I'll happily add it 🙇🏻

### DIFF
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -38,6 +38,7 @@ import (
 	"go.k6.io/k6/errext/exitcodes"
 	"go.k6.io/k6/lib"
 	"go.k6.io/k6/lib/executor"
+	"go.k6.io/k6/lib/metrics"
 	"go.k6.io/k6/lib/types"
 	"go.k6.io/k6/stats"
 )
@@ -235,17 +236,21 @@ func applyDefault(conf Config) Config {
 }
 
 func deriveAndValidateConfig(
-	conf Config, isExecutable func(string) bool, logger logrus.FieldLogger,
+	conf Config,
+	registry *metrics.Registry,
+	isExecutable func(string) bool,
+	logger logrus.FieldLogger,
 ) (result Config, err error) {
 	result = conf
 	result.Options, err = executor.DeriveScenariosFromShortcuts(conf.Options, logger)
 	if err == nil {
-		err = validateConfig(result, isExecutable)
+		err = validateConfig(result, registry, isExecutable)
 	}
+
 	return result, errext.WithExitCodeIfNone(err, exitcodes.InvalidConfig)
 }
 
-func validateConfig(conf Config, isExecutable func(string) bool) error {
+func validateConfig(conf Config, registry *metrics.Registry, isExecutable func(string) bool) error {
 	errList := conf.Validate()
 
 	for _, ec := range conf.Scenarios {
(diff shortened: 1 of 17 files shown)


cmd/config.go
```diff
@@ -274,3 +275,97 @@ func validateScenarioConfig(conf lib.ExecutorConfig, isExecutable func(string) b
...
+		// The defined threshold applies to a non-existing metrics
+		if !registry.Has(name) {
+			err := fmt.Errorf("invalid threshold defined on %s; reason: no metric named %s found", name, name)
+			return errext.WithExitCodeIfNone(err, exitcodes.InvalidConfig)
+		}
+
+		// Get the metric from the registry
+		metric := registry.Get(name)
+
+		switch metric.Type {
```
oleiade, line 311: I would take any idea on how to improve here. I wanted to return thorough, exhaustive, user-friendly error messages, but it cascaded quickly into LONG multiline statements. I've had ideas, but none really satisfying: feedback welcome 🙏🏻

stats/thresholds.go
```diff
@@ -73,23 +53,67 @@ func newThreshold(src string, newThreshold *goja.Runtime, abortOnFail bool, grac
...
-	v, err := t.rt.RunProgram(t.pgm)
-	if err != nil {
-		return false, err
+func (t *Threshold) runNoTaint(sinks map[string]float64) (bool, error) {
+	// Because aggregation method can either be a static keyword ("count", "rate", etc...),
+	// or a parametric expression ("p(somefloatingpointvalue)"), we need to handle this
+	// case specifically. If we encounter the percentile aggregation method token,
+	// we recompute the whole "p(value)" expression in order to look for it in the
+	// sinks.
+	sinkKey := t.Parsed.AggregationMethod
```
olegbespalov, line 66: What do you think about moving the logic (and the comment) of building the sink key to the `ThresholdExpression`, and then calling it `sinkKey := t.Parsed.SinkKey()` or `sinkKey := t.Parsed.Code()`.

The reason is that this calculation of the sink key is dependent exclusively on the state of `ThresholdExpression` and it also makes the current method a bit smaller. And that can be also re-used later in the:

```go
key := fmt.Sprintf("p(%v)", threshold.Parsed.AggregationValue)
```

olegbespalov (commented):
In general, PR looks good to me, I left one small comment.

Special thanks for changing the short variable names to the more readable :heart:

stats/thresholds.go
oleiade: Great idea! On it 🦖

olegbespalov (approved):
:+1:

cmd/config.go
```diff
@@ -254,9 +259,78 @@ func validateConfig(conf Config, isExecutable func(string) bool) error {
...
+	// Note that the reason for passing it as a pointer in the first place is
+	// because it holds a Mutex, which effectively forbids passing it by value.
+	if conf.Thresholds != nil && len(conf.Thresholds) > 0 && registry == nil {
+		err := fmt.Errorf(
+			"unable to validate thresholds configuration; " +
+				"reason: provided registry is nil",
+		)
+		errList = append(errList, err)
+		return consolidateErrorMessage(errList, "there were problems while validating the specified script configuration: ")
+	}
```
MStoykov, line 272: nit: this seems like defensive programming to me - there is no case where `registry` should be `nil` to begin with IMO.

cmd/config.go
```diff
@@ -254,9 +259,78 @@ func validateConfig(conf Config, isExecutable func(string) bool) error {
...
+
+	for thresholdName, thresholds := range conf.Thresholds {
+		// Fetch metric matching the threshold's name
+		metric, ok := registry.Get(thresholdName)
+		if !ok {
+			// The defined threshold applies to a non-existing metrics
+			err := fmt.Errorf("invalid threshold defined on %s; reason: no metric named %s found", thresholdName, thresholdName)
+			return errext.WithExitCodeIfNone(err, exitcodes.InvalidConfig)
```
MStoykov, line 280: This (and the below) should just add to the list so we get all errors in one go

cmd/config.go
```diff
@@ -254,9 +259,78 @@ func validateConfig(conf Config, isExecutable func(string) bool) error {
...
+		supportedMethods = []string{stats.TokenValue}
+	case stats.Rate:
+		supportedMethods = []string{stats.TokenRate}
+	case stats.Trend:
+		supportedMethods = []string{
+			stats.TokenAvg,
+			stats.TokenMin,
+			stats.TokenMax,
+			stats.TokenMed,
+			stats.TokenPercentile,
```
MStoykov, line 315: this doesn't really check if the percentile is valid

MStoykov (changes requested):
As far as I can see (and test) you are still parsing the thresholds way too early so if you have a threshold that is just not with valid syntax to begin with it will abort even earlier (and irregardless of `--no-thresholds`). this is waht https://github.com/grafana/k6/commit/bea617ab1eed45f15ef06ee30effc7e8db488174 did very ... bluntly basically and I think somethign like that is required even now. 

I see that in the issue this is listed as "undecided" but I have some memory that this is more or less required to merge the whole PR (especially for the cloud use case)

WDYT @na--

cmd/config.go
```diff
@@ -254,9 +259,78 @@ func validateConfig(conf Config, isExecutable func(string) bool) error {
...
+				"reason: provided registry is nil",
+		)
+		errList = append(errList, err)
+		return consolidateErrorMessage(errList, "there were problems while validating the specified script configuration: ")
+	}
+
+	for thresholdName, thresholds := range conf.Thresholds {
+		// Fetch metric matching the threshold's name
+		metric, ok := registry.Get(thresholdName)
+		if !ok {
```
MStoykov, line 277: This will also break [this](https://github.com/grafana/k6/issues/2390#issuecomment-1041407562) example even though in practice all the thresholds are *valid*, they just should be evaluated differently from how they are currently. 

The problem here is that `thresholdName` includes the tags while we can only check the *name* of the treshold not the name+ the tags. This is the error:

> ERRO[0000] invalid threshold defined on http_reqs{key:""}; reason: no metric named http_reqs{key:""} found


p.s. The thresholds also are evaluated wrongly with this PR as well - a tag with empty value in threshold is always "matched" even if it doesn't have the tag at all

cmd/config.go
oleiade: I agree, this is defensive programming; however, I would like to keep it that way. I'm happy to discuss how we could improve this, and it's probably my C programming habits kicking in, but the pointer should be checked. It's defensive in the sense that I don't want to risk a potential segfault, or security issue to go through to production because my future self did something stupid and somehow ended up passing nil in there :)

cmd/config.go
oleiade: I'm uncertain if I understand why, could you elaborate?

In the current state of the PR, one could argue the validation is done in two phases: parsing, which happens during bundling as of this PR's state (expression format validation), and actual validation which happens in config validation: "does the metric to you apply that expression to supports the operation?".

My initial interpretation of your comment is "this does not verify that for a given 'p({value})' percentile expression, the metric it applies to actually holds samples for the selected percentile's `{value}`" would that be correct?

cmd/config.go
oleiade: Thanks for pointing the issue out. Do you have pointers or a proposal to make this work, or improve it?

MStoykov: :shrug: I guess something a kin to https://github.com/grafana/k6/blob/737dfa7ced6e4f94dbb3d6f2c2837085e783ac09/core/engine.go#L106-L114

cmd/config.go
MStoykov: I guess you are correct, I probably commented on this *before* I made the general comment that currently parsing of a threshold will still fail the test.

MStoykov:
@oleiade, sorry :bow: . This was shared internally around the time we did revert the previous PR. I though you have seen it, but apperantly not ;( .

oleiade:
The scope and implementation has evolved quite a bit, to the point where it was more convinient to start over. Closing in favor of #2463

PR #2365, opened 2022-02-01
js/grpc: modules.Module migration

codebien:
Migrated `k6/grpc` to the `modules.Module` (aka `modules.ModuleV2`) interface.


js/modules/k6/grpc/client.go
```diff
@@ -249,7 +243,7 @@ func (c *Client) Connect(ctxPtr *context.Context, addr string, params map[string
 	// returns. We only need to close the channel to un-block in a non-error scenario;
 	// otherwise it can be GCd without closing as we return on an error on the channel.
 	errc := make(chan error, 1)
-	go func() {
+	go func(state *lib.State) {
```
MStoykov, line 246: nit: not particularly certain this adds anything -  both the `Client` and it's vu (`c.vu`) get accessed all the time so making it explicit for `state` seems strange. Is there something in particular this fixes?

js/modules/k6/grpc/client_test.go
```diff
@@ -687,7 +693,8 @@ func TestClient(t *testing.T) {
 			val, err := replace(tt.initString.code)
 			assertResponse(t, tt.initString, err, val, ts)
 
-			ctx = lib.WithState(ctx, ts.vuState)
+			mvu.CtxField = lib.WithState(ctx, ts.vuState)
```
MStoykov, line 696: It should still work without this line, I think :thinking:

MStoykov (dismissed):
LGTM :tada: . I did leave some non essential nit picks

js/modules/k6/grpc/client.go
codebien: Yep, the linter

js/modules/k6/grpc/client_test.go
codebien: You're right, done

js/modules/k6/grpc/client.go
MStoykov: do you mean 
```js/modules/k6/grpc/client.go:211:2: variable 'state' is only used in the if-statement (js/modules/k6/grpc/client.go:212:2); consider using short syntax (ifshort)
```

where actually ifshort is *wrong* ?

MStoykov: which I think is https://github.com/esimonov/ifshort/issues/18

codebien: yep, that is the case, so I would prefer to stay safe without using a `nolint` directive and refactor when the state will be used outside. I can add a comment to make it visible and clear. WDYT?

codebien: mmm...I think we could refactor a bit and remove several problems with one shot, trying

codebien: Done, applied a simple refactor so also the other nolint directives have been removed

js/modules/k6/grpc/client.go
```diff
@@ -212,15 +206,73 @@ func (t transportCreds) ClientHandshake(ctx context.Context,
...
+	}
+	if reflect {
+		err := c.reflect(ctx)
+		if err != nil {
+			errc <- err
+			return
+		}
+	}
+
+	errc <- nil
```
olegbespalov, line 265: Do you think we can reduce nesting level and make code simpler? E.g.:

```suggestion
	if !reflect {
		errc <- nil

		return
	}

	errc <- c.reflect(ctx)
```

js/modules/k6/grpc/client.go
```diff
@@ -212,15 +206,73 @@ func (t transportCreds) ClientHandshake(ctx context.Context,
...
+}
+
 // Connect is a block dial to the gRPC server at the given address (host:port)
-// nolint:funlen,cyclop
-func (c *Client) Connect(ctxPtr *context.Context, addr string, params map[string]interface{}) (bool, error) {
-	state := lib.GetState(*ctxPtr)
+func (c *Client) Connect(addr string, params map[string]interface{}) (bool, error) {
+	state := c.vu.State()
```
olegbespalov, line 270: I noticed that there are several places (here and other parts of there code) where we use this check:
```go
state := c.vu.State()
if state == nil {
	return false, errConnectInInitContext
}
```

How about adding a simple method like `modules.IsInitialContext(vu VU) bool` I believe that improves the readability of the code, and what is also essential make code more maintainable since we consolidate the logic that identifies that the particular place is dependent during what phase it is executing?

olegbespalov (commented):
Thanks! The PR looks good, I left a few suggestions that can improve the readability.

js/modules/k6/grpc/client.go
```diff
@@ -243,65 +295,18 @@ func (c *Client) Connect(ctxPtr *context.Context, addr string, params map[string
...
-
-			// (rogchap) we create a wrapper for transport credentials so that we can report
-			// on any TLS errors.
-			creds := transportCreds{
-				credentials.NewTLS(tlsCfg),
-				errc,
-			}
-			opts = append(opts, grpc.WithTransportCredentials(creds))
-		}
+	defer close(errc)
```
MStoykov, line 307: It seems to me like you can move the whole `errc` and just return an `err` from `c.dial` while wrapping all the goroutine and channel code inside of it.

js/modules/k6/grpc/client.go
```diff
@@ -212,15 +205,46 @@ func (t transportCreds) ClientHandshake(ctx context.Context,
...
+	if !reflect {
+		return nil
+	}
+
+	err = c.reflect(ctx)
+	if err != nil {
+		return err
+	}
+
+	return err
```
olegbespalov:
```suggestion
	return c.reflect(ctx)
```
Maybe also possible to rename a `reflect` method to `extractError` or something more human-friendly :thinking:

js/modules/k6/grpc/client.go
```diff
@@ -212,15 +205,46 @@ func (t transportCreds) ClientHandshake(ctx context.Context,
...
-	state := lib.GetState(*ctxPtr)
+func (c *Client) Connect(addr string, params map[string]interface{}) (bool, error) {
+	state := c.vu.State()
 	if state == nil {
 		return false, errConnectInInitContext
 	}
-	isPlaintext, reflect, timeout := false, false, 60*time.Second
 
+	isPlaintext, reflect, timeout := false, false, 60*time.Second
 	for k, v := range params {
```
olegbespalov, line 228: What if we move this to a `parseParams` method, it reduces the function length a bit and looks safe, wdyt?

codebien:
I did some refactor inspired by your suggestions. WDYT?

js/modules/k6/grpc/client.go
codebien: Why do you want rename reflect to extractError? `reflect` is related to the grpc reflection feature in this case

js/modules/k6/grpc/client.go
codebien: I like the idea in terms of readability, but in this way, we can't reuse the `state` variable so for most of the methods we are doing 2x on the call for resolving the `State` that comes from the slow context https://github.com/grafana/k6/blob/0dda328d9a7f563607dae4fe89ff8c5138f8dc09/js/initcontext.go#L169-L171.

so I'm feeling a bit conflicted here without having some profiling data

MStoykov: @olegbespalov please make a separate issue about this. 
@codebien in #2228 I in practice stop getting the state from a context, which is a thing that needs to happen for *everything* at some point either way.

But I do think this should be made ina *separate* PR and probably *across* the codebase, so please open an issue @olegbespalov so we can discuss the API for example and whether we want to do this for more stuff

js/modules/k6/grpc/client.go
olegbespalov: Ah, true, missed that.

In general, wanted to rename since reflect too general, but we can keep it as is

olegbespalov (dismissed):
PR looks good to me, great work! :+1:

MStoykov (commented):
Hm I think we can't remove the goroutine because of https://github.com/grafana/k6/pull/1623#discussion_r498553737 also ... I really prefer to not change the code *all* that much in the same PR that already changes a lot of stuff. 

If there are some refactoring ideas, please move them to a separate PR, so we can have smaller and easier to review PRs instead of just doing a lot of those in one :pray:

codebien:
>Hm I think we can't remove the goroutine because of https://github.com/grafana/k6/pull/1623#discussion_r498553737 also

We should be able to remove it as documented in the commit's description c93b8b09e4c1bcea94468341d9ec76c334cbbeb5 

From my understanding, we did it in that way just for getting the TLS error that is covered by `grpc.WithReturnConnectionError`, I added a test for it in the same commit that fails without using the option. So, I would like to maintain it.

I can move in another PR the stats handler refactor and the reorder. WDYT?

js/modules/k6/grpc/client.go
olegbespalov: Just to keep the connection, opened an issue https://github.com/grafana/k6/issues/2368

MStoykov:
This was added in https://github.com/grpc/grpc-go/pull/3430 which was released *before* the PR that added grpc to k6(and before the comment I linked to). I have no idea if this was just missed or is buggy and just doens't cover all cases :shrug: .

I guess if @rogchap might have some idea, as well as @na--. I have found only one issue https://github.com/grpc/grpc-go/issues/4822, which seems not relevant, but there are others where it seems like[ TLS errors still seem to not work in some cases](https://github.com/grpc/grpc-go/issues/4163). Unfortunately I don't really understand enough to know if those matter or not for our case :(.  Your test for example seems to handle what the issue is talking about so :shrug: . 

If we can go back to the previous code and have this in a separate PR I think it will be better. We also (AFAIK) don't have that many gRPC users so an introduced bug here will likely stay there for while before we can fix it. 

So if it isn't that much work I prefer if this gets split , sorry :( 

p.s. thansk for the commit message, and sorry for skip reading it, but unfortunately I am not particularly certain this catches all the cases the previous one did :(

codebien:
@MStoykov @olegbespalov I restored the entire codebase without the Connect/dial split. I let only the change for the dedicated method for the parameters' parse.

olegbespalov (approved):

MStoykov (approved):

MStoykov:
```
2022/02/03 14:35:21 http: TLS handshake error from 127.0.0.1:50993: EOF
--- FAIL: TestClient (0.00s)
    --- FAIL: TestClient/BadTLS (0.11s)
        client_test.go:675: 
            	Error Trace:	client_test.go:675
            	            				client_test.go:717
            	Error:      	"GoError: context deadline exceeded at reflect.methodValueCall (native)" does not contain "certificate signed by unknown authority"
            	Test:       	TestClient/BadTLS
FAIL
FAIL	go.k6.io/k6/js/modules/k6/grpc	0.397s
```
seems to be because of the test

js/modules/k6/grpc/client_test.go
```diff
@@ -669,11 +689,16 @@ func TestClient(t *testing.T) {
 
 			ts := setup(t)
 
-			// create and attach the GRPC module to goja
 			ctx := common.WithRuntime(context.Background(), ts.rt)
```
MStoykov, line 692: You should actually be able to remove this and maybe have a cancelable context and cancel it :thinking:

MStoykov (approved):

olegbespalov (approved):

Issue #644, opened 2018-05-23
stack overflow while receiving a encrypt response body


gushengyuan:
The k6 process get a stack overflow fatal error while processing a encrypted response body.
In my case, the response body is encrypted in AES/ECB/PCKS5Padding 128bits, I see the respone in wireshark,

`
INFO[0001] 0baa840038a54b78b9eff8d38cdf4865
runtime: goroutine stack exceeds 1000000000-byte limit==============] 1s / 1s
fatal error: stack overflow

runtime stack:
runtime.throw(0xc9296b, 0xe)
        /usr/local/Cellar/go/1.9.3/libexec/src/runtime/panic.go:605 +0x9c
runtime.newstack(0x0)
        /usr/local/Cellar/go/1.9.3/libexec/src/runtime/stack.go:1050 +0x6ef
runtime.morestack()
        /usr/local/Cellar/go/1.9.3/libexec/src/runtime/asm_amd64.s:415 +0x8e

goroutine 30 [running]:
runtime.mapaccess1_faststr(0xb9e080, 0xc0440f4420, 0xc0468c2379, 0xf, 0xc0a8ff9438)
        /usr/local/Cellar/go/1.9.3/libexec/src/runtime/hashmap_fast.go:208 +0x448 fp=0xc0a8ff9370 sp=0xc0a8ff9368 pc=0x40bc48
github.com/loadimpact/k6/vendor/github.com/dop251/goja.(*objectPropIter).next(0xc0ff67bb00, 0x0, 0x0, 0x0, 0x0, 0x0, 0x5166830)
        /Users/robin/Projects/go/src/github.com/loadimpact/k6/vendor/github.com/dop251/goja/object.go:587 +0xdc fp=0xc0a8ff9440 sp=0xc0a8ff9370 pc=0x823a5c
github.com/loadimpact/k6/vendor/github.com/dop251/goja.(*objectPropIter).(github.com/loadimpact/k6/vendor/github.com/dop251/goja.next)-fm(0x0, 0x0, 0x0, 0x0, 0x0, 0x0)
        /Users/robin/Projects/go/src/github.com/loadimpact/k6/vendor/github.com/dop251/goja/object.go:589 +0x98 fp=0xc0a8ff94b0 sp=0xc0a8ff9440 pc=0x86c6a8
[...]

PR #2375, opened 2022-02-04
js/console: Marshal by default native JS objects

### LINKED ISSUE #1146: `console.log` could show JSON content by default
During debugging when working with JSON format, it is common to use the `console.log` to see the response body. 

```js
    let body = JSON.parse(res.body);
    console.log(body);
```

k6 will output `[object Object]`.

You can use `JSON.stringify()` if you want to see the content:

```js
    let body = JSON.parse(res.body);
    console.log(JSON.stringify(body));
```

It could be good if `k6` prints by default t the JSON content as browsers and node do.

codebien:
Closes #1146

:warning: BREAKING CHANGE: After this PR will be merged, all `js/console` methods like `console.log` and `console.info`, in the case a JavaScript Object (or an Array) is passed as an argument then it is converted into a string encoded as a JSON object.

**Objects**

```js
export default function () {
  var user = [{}]

  console.log("the user", user) 
  // before: the user [object Object]
  // after: the user {"name":"Bob"}
}
```

**Arrays**
```js
export default function () {
  var user = {name: "Bob"}

  console.log([1, 2, "mytest", {user: "Bob"}]) 
  // before: 1,2,mytest,[object Object]
  // after: [1,2,"mytest",{"user":"Bob"}]
}
```

The previous way can be restored by directly parsing the object to a string using the `String` function.

```js
export default function () {
  var user = {name: "Bob"}
  console.log(String(user)) 
}
```

### DIFF
--- a/js/console.go
+++ b/js/console.go
@@ -21,7 +21,9 @@
 package js
 
 import (
+	"encoding/json"
 	"os"
+	"reflect"
 	"strings"
 
 	"github.com/dop251/goja"
@@ -52,17 +54,16 @@ func newFileConsole(filepath string, formatter logrus.Formatter) (*console, erro
 	return &console{l}, nil
 }
 
-func (c console) log(level logrus.Level, msgobj goja.Value, args ...goja.Value) {
-	msg := msgobj.String()
-	if len(args) > 0 {
-		strs := make([]string, 1+len(args))
-		strs[0] = msg
-		for i, v := range args {
-			strs[i+1] = v.String()
+func (c console) log(level logrus.Level, args ...goja.Value) {
+	var strs strings.Builder
+	for i := 0; i < len(args); i++ {
+		if i > 0 {
+			strs.WriteString(" ")
 		}
-
-		msg = strings.Join(strs, " ")
+		strs.WriteString(c.valueString(args[i]))
 	}
+	msg := strs.String()
+
 	switch level { //nolint:exhaustive
 	case logrus.DebugLevel:
 		c.logger.Debug(msg)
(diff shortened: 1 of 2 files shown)


js/console.go
```diff
@@ -62,7 +64,20 @@ func (c console) log(ctx *context.Context, level logrus.Level, msgobj goja.Value
 		}
 	}
 
-	msg := msgobj.String()
+	var msg string
+	switch msgobj.ExportType().Kind() { //nolint:exhaustive
```
olegbespalov, line 68: Are we planning to have more cases for the switch? If not, maybe it's better to extract this to a small method where use `if`s is possible?

MStoykov:
@codebien does this still happen with your code https://github.com/grafana/k6/issues/644#issuecomment-767508338

oleiade (changes requested):

codebien:
I applied some changes, @oleiade you're cases should be supported now.

I dropped the `msg` argument, it didn't seem necessary, let me know if you see a case to maintain it.
```go
msg goja.Value, args ...goja.Value)
```

>@codebien does this still happen with your code https://github.com/grafana/k6/issues/644#issuecomment-767508338

@MStoykov Do you mean the case of a circular object printed as `[object Object]`? `JSON.stringify` doesn't support that case. In the case of browsers and runtimes the console seems to have a dedicated spec for formatting the outputs.

js/console.go
```diff
@@ -75,22 +80,40 @@ func (c console) log(level logrus.Level, msgobj goja.Value, args ...goja.Value)
...
+		return v.String()
+	}
+
+	kind := exptype.Kind()
+	if kind != reflect.Map && kind != reflect.Slice {
+		return v.String()
+	}
+
+	b, err := json.Marshal(v.Export())
+	if err != nil {
```
olegbespalov, line 115: Should we do a warning log here, maybe, or it's too much? :thinking:

olegbespalov (approved):
:+1: LGTM left only one non-blocking comment

oleiade (approved):
🎉 👍🏻 🌮

js/console.go
codebien: Do you have any suggestions about the warning?

js/console.go
```diff
@@ -75,22 +80,40 @@ func (c console) log(level logrus.Level, msgobj goja.Value, args ...goja.Value)
...
+	if exptype == nil {
+		return v.String()
+	}
+
+	kind := exptype.Kind()
+	if kind != reflect.Map && kind != reflect.Slice {
+		return v.String()
+	}
+
+	b, err := json.Marshal(v.Export())
```
MStoykov, line 114: Is the `Export` needed here?

MStoykov (approved):
LGTM. 

Can you please explain how this is breaking change in the PR description. As well as point out how to get the previously reported values when possible.

js/console.go
codebien: Not really, I removed it

codebien:
@MStoykov @oleiade @olegbespalov I updated the description with changes and added a check in case of an empty argument list (e.g `console.log()`).

oleiade (dismissed):
👍🏻

js/console.go
olegbespalov: Nothing special, my concern was that if we do not log anything it hard to realize that something wrong with marshaling, but looking into it right now it seems like not the big deal

olegbespalov (approved):
:+1:

codebien:
The output after the latest change

```text
---script.js
export default function () {
	console.log()
	console.error()
	console.log("")
	console.log("the last one")
}

---output
INFO[0000]                                               source=console
ERRO[0000]                                               source=console
INFO[0000]                                               source=console
INFO[0000] the last one                                  source=console
```

MStoykov (approved):

olegbespalov (approved):
:muscle:

Issue #2344, opened 2022-01-18
Deprecate common.Bind() and relatives


codebien:
## Problem statement

Now that `k6` has a new and cleaner API for JS modules and most of the internal modules had been migrated we can consider deprecating the [common.Bind()](https://github.com/grafana/k6/blob/13b2372fd98fb3fefe103a1d22f212164dff9107/js/common/bridge.go#L137) method for having a unique and consistent way of extending the k6's JS API.

### v0.36.0

- [x] Mention that we plan to deprecate them in the release notes

### v0.37.0

- [x] Submit and merge a PR for https://github.com/grafana/k6/issues/2343
- [ ] Add a deprecation comment and relative log warning for `common.Bind()`
- [ ] Add a deprecation comment for connected APIs 

Below a list of the connected methods to deprecate:
```sh
$ rg '\bWith.*context.Context' -g '!vendor'
api/common/context.go
33:func WithEngine(ctx context.Context, engine *core.Engine) context.Context {

js/common/context.go
37:func WithRuntime(ctx context.Context, rt *goja.Runtime) context.Context {
51:func WithInitEnv(ctx context.Context, initEnv *InitEnvironment) context.Context {

lib/context.go
34:func WithState(ctx context.Context, state *State) context.Context {
48:func WithExecutionState(ctx context.Context, s *ExecutionState) context.Context {
62:func WithScenarioState(ctx context.Context, s *ScenarioState) context.Context {
```
From an internal discussion:
>Removing `WithExecutionState` and `WithScenarioState` might not be needed or will need more methods on the `modules.VU` interface

### v0.38.0 or v0.39.0
- [ ] #2385

PR #2376, opened 2022-02-04
js/modules: Remove common utils dependency

codebien:
Something was missed from the module migration to the new `modules.Module` API, where the Runtime and the State are now directly accessible without getting them from the context.

This is required for #2344

### DIFF
--- a/js/modules/k6/k6_test.go
+++ b/js/modules/k6/k6_test.go
@@ -23,7 +23,6 @@ package k6
 import (
 	"context"
 	"fmt"
-	"runtime"
 	"testing"
 	"time"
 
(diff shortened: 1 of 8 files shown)


MStoykov (dismissed):

MStoykov:
Can you also cleanup the tests as they still use `WithRuntime`

na-- (dismissed):

codebien:
I extended the changes to all the tests in `js/k6/modules`

js/modules/k6/k6_test.go
```diff
@@ -111,12 +111,11 @@ func TestSleep(t *testing.T) {
 			assert.NoError(t, err)
 			dch <- endTime.Sub(startTime)
 		}()
-		runtime.Gosched()
+
 		time.Sleep(1 * time.Second)
-		runtime.Gosched()
```
na--, line 116: I wonder why these were here? :confused: Seems they were here from almost the start (https://github.com/grafana/k6/commit/274eef82cbafef77a9eb8461e7a3d920bfd875e7), but they don't seem necessary, so :+1: for removing them

na-- (approved):
:tada:

MStoykov (approved):

olegbespalov (approved):

PR #2381, opened 2022-02-07
Remove calls to context-based utils

codebien:
Removed all not strictly required calls to the context-based utils functions. The `lib.GetState` required a refactor for the executors' tests and the relative used minirunner Fn.

This is expected to be the latest refactor PR before opening the PR for writing the deprecation comments.

### DIFF
--- a/js/runner_test.go
+++ b/js/runner_test.go
@@ -555,6 +555,14 @@ func TestVURunContext(t *testing.T) {
 			vu.Runtime.Set("fn", func() {
 				fnCalled = true
 
+				// Assert to maintain the retro-compatibility
+				// with the old context-based getters
+				assert.Equal(t, vu.Runtime, common.GetRuntime(*vu.Context), "incorrect runtime in context")
+				assert.Nil(t, common.GetInitEnv(*vu.Context)) // shouldn't get this in the vu context
+
+				require.NotNil(t, vu.moduleVUImpl.Runtime())
+				require.Nil(t, vu.moduleVUImpl.InitEnv())
+
 				state := vu.moduleVUImpl.State()
 				require.NotNil(t, state)
 				assert.Equal(t, null.IntFrom(10), state.Options.VUs)
--- a/js/runner_test.go
+++ b/js/runner_test.go
@@ -534,7 +534,7 @@ func TestVURunContext(t *testing.T) {
 	r1, err := getSimpleRunner(t, "/script.js", `
 		exports.options = { vus: 10 };
 		exports.default = function() { fn(); }
-		`)
+	`)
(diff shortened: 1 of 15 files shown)


js/runner_test.go
```diff
@@ -555,17 +555,13 @@ func TestVURunContext(t *testing.T) {
 			vu.Runtime.Set("fn", func() {
 				fnCalled = true
 
-				assert.Equal(t, vu.Runtime, common.GetRuntime(*vu.Context), "incorrect runtime in context")
-				assert.Nil(t, common.GetInitEnv(*vu.Context)) // shouldn't get this in the vu context
```
codebien, line 559: I'm not sure if it does make sense to maintain these (or equivalent) assertions. WDYT?

lib/testutils/minirunner/minirunner.go
```diff
@@ -39,7 +39,7 @@ var (
 // using a real JS runtime, it allows us to directly specify the options and
 // functions with Go code.
 type MiniRunner struct {
-	Fn              func(ctx context.Context, out chan<- stats.SampleContainer) error
+	Fn              func(ctx context.Context, state *lib.State, out chan<- stats.SampleContainer) error
```
na--, line 42: Hmm, I am starting to think that now, after we've refactored the JS module interface, we should also refactor `MiniRunner` to more closely resemble it  :thinking: For example, instead of passing a bunch of params, have a struct that resembles this? https://github.com/grafana/k6/blob/cd017d67db048ba85f5781dc50be13992698e666/js/modules/modules.go#L104-L117

@MStoykov, WDYT?

js/runner_test.go
na--: the semantics are still the same, right, just accesses differently, so why not test them, in at least a few places?

lib/testutils/minirunner/minirunner.go
MStoykov: Yeah, that will likely help. Although I don't see it using most of those, but still probably a good idea :+1:

MStoykov (dismissed):

olegbespalov (dismissed):

lib/testutils/minirunner/minirunner.go
codebien: https://github.com/grafana/k6/issues/2382

codebien:
The latest added asserts have caught that the init env unset was missing. I think we forgot to add it in the context's migration PR.

MStoykov (approved):

olegbespalov (approved):

na-- (approved):

PR #2384, opened 2022-02-10
Deprecate context-based utils

codebien:
Deprecating the context-based utils functions. They are anticipating the next step where we could remove all of them from the `k6` public API.

Closes https://github.com/grafana/k6/issues/2344

### DIFF
--- a/js/common/bridge.go
+++ b/js/common/bridge.go
@@ -134,6 +134,9 @@ func BindToGlobal(rt *goja.Runtime, data map[string]interface{}) func() {
 }
 
 // Bind the provided value v to the provided runtime
+//
+// Deprecated: JS modules can implement the modules.VU interface for getting the context,
+// goja runtime and the VU State, so the goja.Runtime.Set method can be used for data binding.
 func Bind(rt *goja.Runtime, v interface{}, ctxPtr *context.Context) map[string]interface{} {
 	exports := make(map[string]interface{})
 
--- a/js/common/context.go
+++ b/js/common/context.go
@@ -26,6 +26,9 @@ import (
 	"github.com/dop251/goja"
 )
 
+// TODO: https://github.com/grafana/k6/issues/2385
+// Rid all the context-based utils functions
+
 type ctxKey int
 
 const (
@@ -34,11 +37,15 @@ const (
 )
 
 // WithRuntime attaches the given goja runtime to the context.
+//
+// Deprecated: Implement the modules.VU interface for sharing the Runtime.
 func WithRuntime(ctx context.Context, rt *goja.Runtime) context.Context {
 	return context.WithValue(ctx, ctxKeyRuntime, rt)
 }
 
 // GetRuntime retrieves the attached goja runtime from the given context.
+//
+// Deprecated: Use modules.VU for get the Runtime.
 func GetRuntime(ctx context.Context) *goja.Runtime {
 	v := ctx.Value(ctxKeyRuntime)
 	if v == nil {
@@ -48,11 +55,15 @@ func GetRuntime(ctx context.Context) *goja.Runtime {
 }
 
 // WithInitEnv attaches the given init environment to the context.
+//
+// Deprecated: Implement the modules.VU interface for sharing the init environment.
 func WithInitEnv(ctx context.Context, initEnv *InitEnvironment) context.Context {
 	return context.WithValue(ctx, ctxKeyInitEnv, initEnv)
 }
 
 // GetInitEnv retrieves the attached init environment struct from the given context.
+//
+// Deprecated: Use modules.VU for get the init environment.
 func GetInitEnv(ctx context.Context) *InitEnvironment {
 	v := ctx.Value(ctxKeyInitEnv)
 	if v == nil {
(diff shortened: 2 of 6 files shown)


js/common/bridge.go
```diff
@@ -134,6 +134,9 @@ func BindToGlobal(rt *goja.Runtime, data map[string]interface{}) func() {
 }
 
 // Bind the provided value v to the provided runtime
+//
+// Deprecated: JS modules can implement the modules.VU interface for getting the context,
+// goja runtime and the VU State, so the goja.Runtime.Set method can be used for data binding.
 func Bind(rt *goja.Runtime, v interface{}, ctxPtr *context.Context) map[string]interface{} {
```
codebien, line 140: Should we log from `common.Bind` considering that we still use it (init context)?

lib/context.go
```diff
@@ -30,18 +32,31 @@ const (
...
 	if v == nil {
 		return nil
 	}
-	return v.(*State)
+	state := v.(*State)
+	if state == nil {
+		return nil
+	}
+	state.Logger.Warn("GetState has been deprecated and it could be removed in the next versions." +
+		"Please, consider to replace it using the new modules.VU API.")
```
codebien, line 58: Should we use a `sync.Once` for each method for avoiding printing the warning multiple times?

js/common/context.go
```diff
@@ -48,12 +73,20 @@ func GetRuntime(ctx context.Context) *goja.Runtime {
...
+// Deprecated: Implement the modules.VU interface for sharing the init environment.
 func WithInitEnv(ctx context.Context, initEnv *InitEnvironment) context.Context {
 	return context.WithValue(ctx, ctxKeyInitEnv, initEnv)
 }
 
 // GetInitEnv retrieves the attached init environment struct from the given context.
+//
+// Deprecated: Use modules.VU for get the init environment.
 func GetInitEnv(ctx context.Context) *InitEnvironment {
+	logger := logger(ctx)
```
codebien, line 86: should we check and skip in case of `logger == nil`? It would be a bug but maybe avoid us to hotfix it

js/common/bridge.go
na--: We can even emit a nice warning here, since we have the module name, something like `Module 'k6/x/whaever' is using deprecated APIs that will be removed in k6 v0.38.0, see https://k6.io/docs/extensions/new-api-info for more details on how to update it.` :man_shrugging:

lib/context.go
na--: Probably a good idea :thinking:

js/bundle.go
```diff
@@ -329,7 +329,8 @@ func (b *Bundle) instantiate(logger logrus.FieldLogger, rt *goja.Runtime, init *
 		Registry:    b.registry,
 	}
 	init.moduleVUImpl.initEnv = initenv
-	ctx := common.WithInitEnv(context.Background(), initenv)
+	ctx := common.WithLogger(context.Background(), logger)
```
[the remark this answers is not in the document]
MStoykov: Yeah, also Initeenv [already had a logger](https://github.com/grafana/k6/blob/77ff732867c254421d07f8c68e0bc53731b0b8c3/js/common/initenv.go#L35) ;)

olegbespalov (dismissed):
LGTM :+1:

js/initcontext.go
```diff
@@ -214,6 +214,10 @@ func (i *InitContext) requireModule(name string) (goja.Value, error) {
 		mod = perInstance.NewModuleInstancePerVU()
 	}
 
+	i.logger.Warnf(`Module '%s' is using deprecated APIs that will be removed in k6 v0.38.0,`+
+		` for more details on how to update it see`+
+		` https://k6.io/docs/extensions/guides/create-an-extension/#advanced-javascript-extension`, name)
+
 	return i.moduleVUImpl.runtime.ToValue(common.Bind(i.moduleVUImpl.runtime, mod, i.moduleVUImpl.ctxPtr)), nil
```
MStoykov, line 221: I am pretty sure this will be logged for *each* VU

js/initcontext.go
```diff
@@ -214,6 +214,10 @@ func (i *InitContext) requireModule(name string) (goja.Value, error) {
 		mod = perInstance.NewModuleInstancePerVU()
 	}
```
MStoykov, line 215: We probably should drop this lines and the `modules.HasModuleInstancePerVU` interface :thinking:

MStoykov (changes requested):
I am pretty sure that the log warning will be printed per *each* VU(but haven't tested). It's probably good idea if it's only once though ;).

 Although on the other hand,it will probably drive the point home that this requires changes if it fills the screen :rofl:

MStoykov (dismissed):
This will print only once even if multiple modules don't use the new API. But I see this as corner case that isn't worth the complexity ;)

codebien:
>This will print only once even if multiple modules don't use the new API. But I see this as corner case that isn't worth the complexity ;)

Yep, it was my motivation for printing on every call to require. But based on your previous comment I changed my opinion, the user can always fix run by run and it maintains the code simpler without adding and injecting dependencies, in the end, it will be there for only one release.

MStoykov (approved):

na--:
I haven't tested it with an extension though, just read the code

PR #2387, opened 2022-02-15
Prevent panic on not providing second argument to check

MStoykov:


js/modules/k6/k6.go
```diff
@@ -158,6 +159,9 @@ func (mi *K6) Check(arg0, checks goja.Value, extras ...goja.Value) (bool, error)
 	if state == nil {
 		return false, ErrCheckInInitContext
 	}
+	if checks == nil {
+		return false, fmt.Errorf("No checks provided to `check`")
```
olegbespalov, line 163: It seems like there is nothing that we can format :wink:

MStoykov: fixed in [3be0fd4](https://github.com/grafana/k6/pull/2387/commits/3be0fd4d976396490f870ae7de1f87d760b87296)

js/modules/k6/k6.go
```diff
@@ -158,6 +158,9 @@ func (mi *K6) Check(arg0, checks goja.Value, extras ...goja.Value) (bool, error)
 	if state == nil {
 		return false, ErrCheckInInitContext
 	}
+	if checks == nil {
+		return false, errors.New("No checks provided to `check`") //nolint:stylecheck
```
olegbespalov, line 162: Sorry again :disappointed:  

But do we really need the `nolint` here? In the end, the error looks like this:
```
GoError: No checks provided to `check` at go.k6.io/k6/js/modules/k6.(*K6).Check-fm (native)
```
So it seems like it doesn't matter if the error message starts from the capital, but having no `nolint` code looks more attractive, right?

na-- (commented):
LGTM, though I agree with @olegbespalov's comment.

na-- (approved):

olegbespalov (approved):
:+1:

PR #2391, opened 2022-02-18
Add TLS Issuer to Response

zv0n:
I've added TLS Issuer information to the Response object. In our organization we create Kubernetes Ingresses dynamically for certain pods and we would like to test whether the Ingress has a correct SSL certificate rather than a placeholder one.

The Issuer is extracted from the first peer certificate as per docs:

> The first element is the leaf certificate that the connection is verified against.

I have chosen to use the first organization of the certificate as the issuer, this seems like a reasonable value to me, but I'm open to changing that as I don't know too much about X590 certificates.

### DIFF
--- a/lib/netext/httpext/response.go
+++ b/lib/netext/httpext/response.go
@@ -85,6 +85,7 @@ type Response struct {
 	Timings        ResponseTimings          `json:"timings"`
 	TLSVersion     string                   `json:"tls_version"`
 	TLSCipherSuite string                   `json:"tls_cipher_suite"`
+	TLSIssuer      string                   `json:"tls_issuer"`
 	OCSP           netext.OCSP              `json:"ocsp"`
 	Error          string                   `json:"error"`
 	ErrorCode      int                      `json:"error_code"`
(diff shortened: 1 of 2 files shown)


MStoykov:
Hi @zv0n, Thanks for the PR !  :tada: 

Unfortunately I feel like this is a very specific "patch" for your specific use case. 

I would prefer if we are letting the user as much of the available information as possible without making some assumptions such as "the first peer is fine". Additionally, this adds one more field to the already many fields on the http.Response. And the problem there is that for *each* request, we always have to populate and return those to js VM, and it possibly needs to do work on them, and most users never use *any* of them.

While one more field is not an issue, adding all the fields around the TLS connection is likely going to be. 

Can you please open an issue with *your* particular use case and what you need to get it done. Don't worry about proposing a solution, although an "ideal" script example of how you want to use it is also welcome.

The solution likely will include a lot more work than just adding a bunch of fields and might not even come for some time, for which I am sorry :(. But adding one more field all the time is just not very sustainable. 

I am going to close this as I don't see as merging this particular PR, but we would really like to accommodate your use case, even if it takes more time.

Issue #1302, opened 2020-01-08
Refactor the module structure for the new executors


na--:
This discussion is a nice starting point for what's currently suboptimal: https://github.com/loadimpact/k6/pull/1007#discussion_r332447323

I also noticed that the `lib` package depends on the `js/compiler` package... `go list -f '{{ join .Imports "\n" }}'  github.com/loadimpact/k6/lib | sort` produces:
```
...
github.com/loadimpact/k6/js/compiler
github.com/loadimpact/k6/lib/fsext
github.com/loadimpact/k6/lib/types
github.com/loadimpact/k6/loader
github.com/loadimpact/k6/stats
github.com/loadimpact/k6/ui/pb
...
```
it's not related to the new executors, but this shouldn't be the case

na--:
Some of the tests of the API (and probably other modules as well) are currently structured in such a way that they import the `js` package: https://github.com/loadimpact/k6/blob/298cf6bfab04d64392db34872240aca0f96c3e78/api/v1/setup_teardown_routes_test.go#L37

This would cause any library that also imports them to include things like goja and core.js, which is probably not what we want... :man_facepalming: We can instead use different package names for these tests, for example `v1_test`.

MStoykov:
hm ... I think that it probably won't require the `js` package as it is only in the test, but I could be wrong ...

oleiade:
We have recently discussed this issue in our backlog grooming session. We agreed that if the specific issue reported here: the `lib` package importing the `js` module, we should close.

I just verified on `master` and it seems to be the case indeed. Closing this. Feel free to reopen in the future if you believe this issue is still relevant regardless.

PR #2392, opened 2022-02-21
Fix wrong omitempty for options.scenarios

na--:
With the current code, running this script:
```js
export default function () {
    console.log(JSON.stringify(options, null, 4));
}
```

Will produce something like this:
```js
{
    "scenarios,omitempty": {},
    "summaryTrendStats": [ "avg", "min", "med", "max", "p(90)", "p(95)"],
    "systemTags": ["check","error","error_code","expected_response","group","method","name","proto","scenario","service","status","subproto","tls_version","url"]
}
```

The actual issue is caused by the `lib.Options.ForEachSpecified()` hack: https://github.com/grafana/k6/blob/737dfa7ced6e4f94dbb3d6f2c2837085e783ac09/js/bundle.go#L290
where it doesn't actually properly parse the `json` struct tag and returns its whole value: https://github.com/grafana/k6/blob/737dfa7ced6e4f94dbb3d6f2c2837085e783ac09/lib/options.go#L615-L618

However, getting rid of that whole mess is something that should probably be done as a part of #883. Similarly, not exporting the default values for `summaryTrendStats` and `systemTags` will also be more complicated and out of scope.

This PR seems like a good idea because it seems to fix the issue without any side-effects (:crossed_fingers:), it's simple and it doesn't make sense for `Scenarios` to be the only `lib.Options` property with `omitempty`...

### DIFF
--- a/lib/options.go
+++ b/lib/options.go
@@ -292,7 +292,7 @@ type Options struct {
 	// We should support specifying execution segments via environment
 	// variables, but we currently can't, because envconfig has this nasty bug
 	// (among others): https://github.com/kelseyhightower/envconfig/issues/113
-	Scenarios                ScenarioConfigs           `json:"scenarios,omitempty" ignored:"true"`
+	Scenarios                ScenarioConfigs           `json:"scenarios" ignored:"true"`
 	ExecutionSegment         *ExecutionSegment         `json:"executionSegment" ignored:"true"`
 	ExecutionSegmentSequence *ExecutionSegmentSequence `json:"executionSegmentSequence" ignored:"true"`
 


olegbespalov (approved):
:crossed_fingers:

MStoykov (approved):
It seems to have the correct value in the archive, so I hope this is fine :crossed_fingers:

MStoykov (approved):

codebien (commented):
@na-- The test seems very fragile, could we add the current fixed case there, just for having at least one field? It should also assert the generated string.
https://github.com/grafana/k6/blob/f60fd2580cf992304834a8351e4d3454786e3a61/lib/options_test.go#L372-L378 
This is more for the future when we will work on the 883, mostly for reducing the probability of breaking stuff.

olegbespalov (approved):

na--:
@codebien, I tried, but it's quite tricky to test scenario configs in `lib/options_test.go` because of import loops with `lib/executor`, thus my desire to fix them properly in https://github.com/grafana/k6/issues/1302 and https://github.com/grafana/k6/issues/883 :disappointed: And asserting the generated JSON of an empty `Options` will either be tricky or would require asserting a bug... It should not actually contain `summaryTrendStats` and `systemTags`, but it currently does does (see https://github.com/grafana/k6/pull/2392#issue-1145804231) :disappointed: ... 

I feel like a test here will be a half-baked measure that may hurt us more than it will help us when it comes to properly fixing the whole mess... :disappointed: I may be quite mistaken though, or have misunderstood you, so please open a PR if you think we can add a sensible test. Maybe something in `cmd/` can work well, similar to `config_consolidation_test.go`? :thinking:

codebien (approved):

PR #2397, opened 2022-02-24
Use golangci-lint action in github

MStoykov:


.github/workflows/all.yml
```diff
@@ -43,27 +43,17 @@ jobs:
...
-          GOLANGCI_LINT_VERSION=$(head -n 1 "${GITHUB_WORKSPACE}/.golangci.yml" | tr -d '# ')
-          go install "github.com/golangci/golangci-lint/cmd/golangci-lint@$GOLANGCI_LINT_VERSION"
-      - name: Run linters
-        run: |
-          BASEREV=$(git merge-base HEAD origin/master)
-          echo "Base revision: $BASEREV"
-          golangci-lint run --out-format=tab --new-from-rev "$BASEREV" ./...
+          # Required: the version of golangci-lint is required and must be specified without patch version: we always use the latest patch version.
+          version: v1.43
+          only-new-issues: true
```
imiric, line 56: Ah, glad to see this works now. The reason we didn't use it before was that `only-new-issues` didn't work the same as `--new-from-rev` so it always showed old issues.

.github/workflows/all.yml
```diff
@@ -43,27 +43,17 @@ jobs:
...
-        run: |
-          GOLANGCI_LINT_VERSION=$(head -n 1 "${GITHUB_WORKSPACE}/.golangci.yml" | tr -d '# ')
-          go install "github.com/golangci/golangci-lint/cmd/golangci-lint@$GOLANGCI_LINT_VERSION"
-      - name: Run linters
-        run: |
-          BASEREV=$(git merge-base HEAD origin/master)
-          echo "Base revision: $BASEREV"
-          golangci-lint run --out-format=tab --new-from-rev "$BASEREV" ./...
+          # Required: the version of golangci-lint is required and must be specified without patch version: we always use the latest patch version.
+          version: v1.43
```
olegbespalov, line 55: Awesome stuff :+1: 

But we still can keep the version in a single place using `::set-output` (https://stackoverflow.com/a/65616499)

MStoykov: done in [4119c40](https://github.com/grafana/k6/pull/2397/commits/4119c40d75147a5b314ae07bd935e089037b926d)

na-- (approved):

Issue #2395, opened 2022-02-22
On-demand copy for Response.Request.Body


codebien:
As part of https://github.com/grafana/k6/issues/2311 investigation and as previously reported in https://github.com/grafana/k6/issues/1931#issuecomment-809177427 the `Response.Request.Body` field as a `string` type requires executing a dedicated copy for each request.
This problem creates noise for better profiling of #2311 and it should be required from any solution applied to #2311.

https://github.com/grafana/k6/blob/737dfa7ced6e4f94dbb3d6f2c2837085e783ac09/lib/netext/httpext/request.go#L52-L56

### Suggestion

The current string field is really required only when the JS property is accessed directly. The proposal is to switch to `[]byte` and generate the string only on-demand when the property is accessed from JS. It should allow reusing the original slice used on the HTTP parsing.

The copy should be executed in the Get function for a `goja.Proxy` object. The Proxy wraps the `k6/http.Response` object.

### Example

```js
import http from "k6/http";
import { sleep } from "k6";

const binFile = open("1MB.bin", "b");

export const options = {
        systemTags: [],
	iterations: 1000,
	vus: 200,
};

export default function () {
    const data = {
	field: Math.random(),
	file: http.file(binFile, "testdata.bin"),
    };

    const res = http.post("http://test:8080/upload", data);
    sleep(1);
}
```
![profiling](https://user-images.githubusercontent.com/2103732/155115875-7ef6765d-3607-46da-8d45-b63ed0f87750.png)

na--:
@MStoykov, this is about the copy of the **request** body that is saved in `Response.request.body`, not the response body. So `io.ReadCloser` has nothing to do with it and some sort of an on-demand lazy copy seems like it should be easy to do.

Moreover, this probably is actually a bug, not just very inefficient: https://github.com/grafana/k6/blob/737dfa7ced6e4f94dbb3d6f2c2837085e783ac09/lib/netext/httpext/request.go#L145

For binary bodies, this will likely mangle them... So, I am fine with trying to use a proxy. I am also fine with making a minor breaking change and actually make the property return a lazy-loaded `ArrayBuffer` since it would basically be a bugfix :man_shrugging: Or have it purely on the JS side and reference literally the exact same variable that was passed to `http.request()`, so no copying takes place.

codebien:
> Or have it purely on the JS side and reference literally the exact same variable that was passed to http.request(), so no copying takes place.

[...]

PR #2398, opened 2022-02-25
k6/http: Use a proxy for the Response.Request

codebien:
This PR removes the all-time copy for the Response.Request.Body, it's a PoC trying to provide the benefit without generating any breaking change.

Open questions:

* Return an ArrayBuffer?
* Eventloop data-safe still to check

Closes https://github.com/grafana/k6/issues/2395


codebien:
Here for a long-time, probably too much, and it is now outdated. We may consider reopening it, but let's close it for now.

PR #2399, opened 2022-02-25
Event loop testing

MStoykov:
The PR mostly tries to make possible to test event loop APIs in the module themselves.

~The whole `k6/events` will be dropped before being merged.~

It is now moved in a not registered module that gets registered only in a `core/local/eventloop_test.go`


oleiade (approved):
👍🏻

na-- (changes requested):
Please don't put these in `/core/local`, that folder is about to disappear and this will just make my task of fixing the tests harder :sweat_smile: 

I suggest we either put the integration tests in a top-level `/tests/` package, or in something like `/cmd/integration_tests/` (and also move [the current `/cmd/integration_test.go`](https://github.com/grafana/k6/blob/master/cmd/integration_test.go) file there).

That way, you also don't need to have `init()` functions for `modules.Register()`, you can just directly call that in he test.

MStoykov:
I can move them to `cmd/integration_tests/eventloop/eventloop_test.go` but remaking them will require me to pull some stuff from `cmd` so I can either:
1. move it to a different folder all together I guess `tests`
2. leave it there for now and when somebody works on integration tests improvement they can refactor them a bit 
WDYT ?

na--:
I think it's fine if `cmd/integration_tests/<something>` imports `cmd/`, that is a fairly usual Go practice where sub-packages import types in the main package or something like that... So I am fine with either option, though maybe every so slightly in favor of a top-level `/tests` or `/integration_tests` folder, since that will be a bit more discoverable :man_shrugging:

cmd/integration_tests/eventloop/eventloop_test.go
```diff
@@ -23,7 +30,7 @@ func eventLoopTest(t *testing.T, script []byte, testHandle func(context.Context,
 	logHook := &testutils.SimpleLogrusHook{HookedLevels: []logrus.Level{logrus.InfoLevel, logrus.WarnLevel, logrus.ErrorLevel}}
 	logger.AddHook(logHook)
 
-	script = []byte(`import {setTimeout} from "k6/experimental";
+	script = []byte(`import {setTimeout} from "k6/x/events";
   ` + string(script))
 	registry := metrics.NewRegistry()
 	builtinMetrics := metrics.RegisterBuiltinMetrics(registry)
```
na--, line 36: This and `js.New()` is unnecessary and make this into not-really-an-integration test :sweat_smile: 

If you make `newGlobalTestState()` and `newRootCommand()` exported in `cmd/`, you can use them here and write full integration tests that are actually shorter to write, since they won't need all of this setup, won't need the `init()` code, etc. From what I can see, you are mostly looking at log lines, which is super easy to do with that.

na--:
> I am kind of for letting it be as it's now (see [e3e94f0](https://github.com/grafana/k6/pull/2399/commits/e3e94f00eb48e84fcae73054ad35d177d0f7be92)) and fixing it later when this structure emerges.

On first glance, I thought that this was a good idea and gave it a :+1: , but having thought about it some more, I now disagree...

We have a nice `cmd/` package that can be tested now, so we can't use that as an procrastination excuse anymore... So I don't think we should keep postponing writing good tests with the hope that someone will fix them in the future. Let's decide on this integration test structure now and have this be the first PR that introduces and uses it.

cmd/integration_tests/eventloop/eventloop_test.go
MStoykov: I will need to move `newGlobalTestState` in separate package so that we don't pollute the global scope with the  `testing` package.

na--: :thinking: hmm yeah, maybe to this root `tests/` package? get it completely out of `cmd/` :man_shrugging:

MStoykov:
I started working on this @na-- and I need to reexport half the `cmd` package just to get it to work ... and I expect that I would tehn also need to move most of the tests out of `cmd` or export even more stuff. As moving GlobalStateTest out means that it needs to export `cmd` to get the GlobalState (which also makes anything in it exported), but then some tests in `cmd` use `GlobalStateTest. 

So yeah I am *firmly* against feature creaping this PR to this extend. Especially as I still have to export more (currently rightfully) not exported types just to get it started ...

na--:
Fair enough, I created https://github.com/grafana/k6/issues/2459 to track this issue. If you have any comments about it from your efforts today, please write them in there.

js/modulestest/testmodules/events/events.go
```diff
@@ -0,0 +1,157 @@
+// Package events is implementing setInterval setTimeout and co. Not to be used mostly for testing purposes
+package events
```
na--, line 2: Minor nitpick, but does this package even need to be under `/js`? Why not under `/cmd/integration_tests/eventloop/`?

```suggestion
// Package events implements setInterval, setTimeout and co. Not to be used, mostly for testing purposes
package events
```

na-- (dismissed):

olegbespalov (approved):

js/modulestest/testmodules/events/events.go
MStoykov: The original idea was that I will somehow keep the whole `js` integration tests in ... `js` but this means that under `js` I need to import *at least* `./core/local` (or what it is currently there). Which just doesn't seemed like a good idea - so I moved the tests outside of `js` (or didn't move them in it more accurately) but the test extension/module was left there. 

And now currently this package does test itself with a unit tests, using the changes in this PR to make it easier to test evetnt loop related functionaility in the module itself. So it still seems like a good idea to keep it there. But I can move it I guess :shrug:

na--: It's not a big deal, but if don't actually plan to expose this to scripts, it seems better to keep it next to the test that actually uses it. It's own unit tests will swill work there, and nothing uses it here, so it feels tidier :man_shrugging:

MStoykov:
I rebased the whole PR in 3 commits - no actual changes from before that

na-- (approved):

olegbespalov (approved):

PR #2400, opened 2022-02-25
Fix/1443 remove js runtime from threshold calculation

oleiade:
This Pull Request does the following:
* brings back the reverted threshold parsing.
* adds support for the `--no-thresholds` option of the `run` command
* introduces a `Thresholds.Parse` method. Calls it as part of the `run` command
* modifies tests in `thresholds_test.go` which assumed `UnmarshalJSON` would parse thresholds

This PR acts as a tentative to address the minimum scope necessary to merge Thresholds parsing safely to master. It will need to lead to further refactors and improvements both in k6 itself, and in our internal infrastructure.

Reviewers can safely ignore the merge commit, as it was already approved in https://github.com/grafana/k6/pull/2251. Reviewers should focus on the last commit only.

### DIFF
--- a/cmd/archive.go
+++ b/cmd/archive.go
@@ -28,6 +28,8 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 
+	"go.k6.io/k6/errext"
+	"go.k6.io/k6/errext/exitcodes"
 	"go.k6.io/k6/lib/metrics"
 )
 
@@ -73,6 +75,18 @@ An archive is a fully self-contained test run, and can be executed identically e
 			)
 			if err != nil {
 				return err
+			}
+
+			// Parse the thresholds, only if the --no-threshold flag is not set.
+			// If parsing the threshold expressions failed, consider it as an
+			// invalid configuration error.
+			if !runtimeOptions.NoThresholds.Bool {
+				for _, thresholds := range conf.Options.Thresholds {
+					err = thresholds.Parse()
+					if err != nil {
+						return errext.WithExitCodeIfNone(err, exitcodes.InvalidConfig)
+					}
+				}
 			}
 
 			_, err = deriveAndValidateConfig(conf, r.IsExecutable, logger)
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -125,6 +125,18 @@ a commandline interface for interacting with it.`,
 				return err
 			}
 
+			// Parse the thresholds, only if the --no-threshold flag is not set.
+			// If parsing the threshold expressions failed, consider it as an
+			// invalid configuration error.
+			if !runtimeOptions.NoThresholds.Bool {
+				for _, thresholds := range conf.Options.Thresholds {
+					err = thresholds.Parse()
+					if err != nil {
+						return errext.WithExitCodeIfNone(err, exitcodes.InvalidConfig)
+					}
+				}
+			}
+
 			conf, err = deriveAndValidateConfig(conf, initRunner.IsExecutable, logger)
 			if err != nil {
 				return err
--- a/stats/thresholds.go
+++ b/stats/thresholds.go
@@ -17,54 +17,35 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  *
  */
-
 package stats
 
 import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"strings"
 	"time"
-
-	"github.com/dop251/goja"
 
 	"go.k6.io/k6/lib/types"
 )
-
-const jsEnvSrc = `
-function p(pct) {
-	return __sink__.P(pct/100.0);
-};
-`
-
-var jsEnv *goja.Program
-
-func init() {
-	pgm, err := goja.Compile("__env__", jsEnvSrc, true)
-	if err != nil {
-		panic(err)
-	}
-	jsEnv = pgm
-}
 
 // Threshold is a representation of a single threshold for a single metric
 type Threshold struct {
 	// Source is the text based source of the threshold
 	Source string
-	// LastFailed is a makrer if the last testing of this threshold failed
+	// LastFailed is a marker if the last testing of this threshold failed
 	LastFailed bool
 	// AbortOnFail marks if a given threshold fails that the whole test should be aborted
 	AbortOnFail bool
 	// AbortGracePeriod is a the minimum amount of time a test should be running before a failing
 	// this threshold will abort the test
 	AbortGracePeriod types.NullDuration
-
-	pgm *goja.Program
-	rt  *goja.Runtime
+	// parsed is the threshold expression parsed from the Source
+	parsed *thresholdExpression
 }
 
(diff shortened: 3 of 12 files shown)


stats/thresholds_test.go
```diff
@@ -56,9 +54,8 @@ func TestNewThreshold_InvalidThresholdConditionExpression(t *testing.T) {
 	abortOnFail := false
 	gracePeriod := types.NullDurationFrom(2 * time.Second)
 
-	gotThreshold, err := newThreshold(src, abortOnFail, gracePeriod)
+	gotThreshold := newThreshold(src, abortOnFail, gracePeriod)
 
-	assert.Error(t, err, "instantiating a threshold with an invalid expression should fail")
```
MStoykov: Can you instead of removing this, call Parse and check the error from it?

oleiade: To clarify, do you mean calling `Parse` in the test here, or calling it in `UnmarshalJSON`?

PS: calling it inside the test would be equivalent to test `Parse` itself. But I want to make sure :)

MStoykov: I meant calling it here on its own, so you can just get the same error you got before. 

This also is true for all other places – just everywhere call `Parse` after the thresholds are created and check the error

oleiade: Alright, will do 👍🏻

stats/thresholds.go
```diff
@@ -44,18 +44,18 @@ type Threshold struct {
 	parsed *thresholdExpression
 }
 
-func newThreshold(src string, abortOnFail bool, gracePeriod types.NullDuration) (*Threshold, error) {
+func newThreshold(src string, abortOnFail bool, gracePeriod types.NullDuration) *Threshold {
 	parsedExpression, err := parseThresholdExpression(src)
 	if err != nil {
-		return nil, err
+		return nil
 	}
```
MStoykov, line 51: You need to remove this lines otherwise this will return a nil an then you will get nullPointerException

MStoykov (commented):
Apart from: 
1. the one thing that leads to nil pointer exceptions that I did actually fix in my patch
2.  the tests not calling `Parse` -which probably would've caught `1`

I think this  okay :+1:

stats/thresholds.go
oleiade: Good catch Indeed 👀  I actually ran into the problem while writing test for `Parse` 👍🏻

stats/thresholds_test.go
oleiade: I actually had forgotten to remove `parseThresholdExpression` call from `newThreshold`. It's now done. You did in your patch too. It makes sense.

Following that, I don't think it makes much sense to call `Parse` in the `newThreshold` (singular) tests anymore, as it would only test the input we give in the test is parsable, and would not assert anything about `newThreshold` itself anymore. Would you agree?

oleiade:
As no one had committed to a review yet, I've force-pushed. Sorry if that caused any disturbance. I have removed the dependency of `newThreshold` (**singular**) to `parseThresholdExpression`. I have added tests for the `Thresholds.Parse` (**plural**) method.

stats/thresholds_test.go
MStoykov: On a similar note the current test failures are because you don't Parse it in some cases where that is needed as it will then be used. 

Given that this in practice just restores to what hte test were before I feel that is much better either way.

cmd/run.go
```diff
@@ -125,6 +125,16 @@ a commandline interface for interacting with it.`,
...
+			// Parse the thresholds, only if the --no-threshold flag is not set.
+			if !runtimeOptions.NoThresholds.Bool {
+				for _, thresholds := range conf.Options.Thresholds {
+					err := thresholds.Parse()
+					if err != nil {
+						return errext.WithExitCodeIfNone(err, exitcodes.InvalidConfig)
+					}
+				}
+			}
+
```
MStoykov, line 137: Unfortunately I think we need to copy this to `cmd/cloud.go` and `cmd/archive.go` :thinking: Looking at the code the alternative is to add it in `getConsolidatedConfig`, which arguably will make this more testable :thinking:

stats/thresholds_test.go
oleiade: Indeed, I caught that in the process. Thanks for pointing it out anyway 😸 I've added parsing steps to the tests.

cmd/run.go
oleiade: I've added the check to `cmd/cloud.go` and `cmd/archive.go`. I've added tests to verify the behavior is respected in the context of the `run` command. Looking into doing the same for `cloud` and `archive`. 

I chose not to put the code in `getConsolidatedConfig` as the actual check is really a side effect, one that does not have any impact on the config struct `getConsolidateConfig` produces (as far as I can tell). Moreover, I do think that validation before running will justify refactoring this somehow in the close future anyway.

cmd/archive.go
```diff
@@ -75,6 +77,18 @@ An archive is a fully self-contained test run, and can be executed identically e
 				return err
 			}
 
+			// Parse the thresholds, only if the --no-threshold flag is not set.
+			// If parsing the threshold expressions failed, consider it as an
+			// invalid configuration error.
+			if !runtimeOptions.NoThresholds.Bool {
+				for _, thresholds := range conf.Options.Thresholds {
```
olegbespalov, line 84: What if we move this loop to the `conf.options.ValidateTreshholds` ? :thinking:

stats/thresholds_parser.go
```diff
@@ -0,0 +1,214 @@
...
+	for _, m := range aggregationMethodTokens {
+		// Percentile expressions being of the form p(value),
+		// they won't be matched here.
+		if m == input {
+			return input, null.Float{}, nil
+		}
+	}
+
+	// Otherwise, attempt to parse a percentile expression
+	if strings.HasPrefix(input, tokenPercentile+"(") && strings.HasSuffix(input, ")") {
```
olegbespalov, line 200: If we change the logical condition to, we can benefit from the early return with the error that strictly follows from the check, like:
```
if !strings.HasPrefix(input, tokenPercentile+"(") || !strings.HasSuffix(input, ")") {
     return "", null.Float{}, fmt.Errorf("failed parsing method from expression")
}
```

And this change also reduces the nesting of the ifs, which improves readability.

WDYT?

cmd/archive.go
MStoykov: I think that is probably a small enough change to be done, but given that we are trying to get it merged before the release - I vote for leaving this for the next PR in the next release that will either way add a bunch of Validation.

MStoykov (approved):
LGTM - I am going to make some more manual tests but hopefully we don't need to change anything to get this in v0.37.0 and make other non necessary fixes for v0.38.0

olegbespalov (approved):
Okay, keeping my comments for the following PRs :+1:

oleiade:
Thanks everyone, for your input 🎉 
Merging 🦖

Issue #2401, opened 2022-02-28
Issue with HEAD requests in 0.36.0?


grantyoung:
### Brief summary

After upgrading to `0.36.0` we noticed that in one of our tests it started failing with no successful requests. Uniquely this was the only test that was sending HEAD requests (`http.head`).

Digging further we identified our nginx server was rejecting all requests vaguely with 400 errors. After switching back to `0.35.0` the issue was no longer happening suggesting this is an issue on the k6 side.

### k6 version

0.36.0

### OS

Ubuntu 18.04, Mac OS 12.2.1

### Steps to reproduce the problem

- Create a test sending HEAD requests, possibly against a standard nginx server
- Notice that they fail with 400 errors

### Expected behaviour

HEAD requests are successful

### Actual behaviour

HEAD requests are being rejected with 400 errors

MStoykov:
Hi @grantyoung I think this as in v0.36.0 we did refactor this code and it seems that we made `http.head` take `body` as second argument as all other helpers except `http.get` :thinking: 

Can you confirm that adding a `null` body fixes this for you?

na--:
Sorry @grantyoung, this seems to be a bug I introduced with https://github.com/grafana/k6/pull/2226 that the tests didn't catch as a regression :disappointed: We plan to release v0.37.0 in a couple of weeks and we'll fix it in there.

grantyoung:
No problem at all @na--, thanks for the update.

PR #2402, opened 2022-02-28
Revert k6/http.head to not taking body as second argument

MStoykov:
This has been the case forever until v0.36.0.

### DIFF
--- a/js/modules/k6/http/http.go
+++ b/js/modules/k6/http/http.go
@@ -92,7 +92,10 @@ func (r *RootModule) NewModuleInstance(vu modules.VU) modules.Instance {
 		args = append([]goja.Value{goja.Undefined()}, args...) // sigh
 		return mi.defaultClient.Request(http.MethodGet, url, args...)
 	})
-	mustExport("head", mi.defaultClient.getMethodClosure(http.MethodHead))
+	mustExport("head", func(url goja.Value, args ...goja.Value) (*Response, error) {
+		args = append([]goja.Value{goja.Undefined()}, args...) // sigh
+		return mi.defaultClient.Request(http.MethodHead, url, args...)
+	})
 	mustExport("post", mi.defaultClient.getMethodClosure(http.MethodPost))
 	mustExport("put", mi.defaultClient.getMethodClosure(http.MethodPut))
 	mustExport("patch", mi.defaultClient.getMethodClosure(http.MethodPatch))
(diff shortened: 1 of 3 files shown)


na--:
Ah, I probably messed it up with https://github.com/grafana/k6/pull/2226 :disappointed: Can you also add a test so we don't have the same regression again?

Also, I wonder if we shouldn't add a "known bugs" sections to the release notes and edit the v0.36.0 ones? This will be fixed in v0.37.0 in a couple of weeks, so probably no need to make a v0.36.1 point release, but it might still be useful to have it until then

codebien (approved):
>Also, I wonder if we shouldn't add a "known bugs" sections to the release notes and edit the v0.36.0 ones?

+1 for this, if we are not going to release a patch.

na--:
@MStoykov, can you add something like this to the bottom of `/release notes/v0.36.0.md`? 

> # Known Bugs
> 
> [#2226](https://github.com/grafana/k6/pull/2226) introduced an unintended breaking change to `http.head()`. The signature in k6 v0.35.0 was `http.head(url, [params])` and was inadvertently changed to `http.head(url, [body], [params])` in v0.36.0. That change will be reverted in k6 v0.37.0, but until then, we suggest users use the stable `http.requset('HEAD', url, null, params)` API for HTTP HEAD requests that need to specify custom [parameters](https://k6.io/docs/javascript-api/k6-http/params). Thanks, @[grantyoung](https://github.com/grantyoung), for reporting the problem ([#2401](https://github.com/grafana/k6/issues/2401))!

We can edit the https://github.com/grafana/k6/releases/tag/v0.36.0 manually when we agree on the message

js/modules/k6/http/http.go
```diff
@@ -92,7 +92,10 @@ func (r *RootModule) NewModuleInstance(vu modules.VU) modules.Instance {
 		args = append([]goja.Value{goja.Undefined()}, args...) // sigh
 		return mi.defaultClient.Request(http.MethodGet, url, args...)
 	})
-	mustExport("head", mi.defaultClient.getMethodClosure(http.MethodHead))
+	mustExport("head", func(url goja.Value, args ...goja.Value) (*Response, error) {
+		args = append([]goja.Value{goja.Undefined()}, args...) // sigh
```
codebien, line 96:
```suggestion
        // It avoid creating a dedicated method reusing the common Request method
        // setting the body of the request to null.
		args = append([]goja.Value{goja.Undefined()}, args...)
```
Sorry for seeing it just now, but if we want a comment there then I think we should use the opportunity for explaining why we are doing it.

oleiade:
+1 to adding a "Known bugs" indeed. Moreover, as @codebien mentioned, we could probably release a "patch" release too?

oleiade (approved):
👍🏻 

Consider my comments as non-blocking.

na-- (approved):

PR #2403, opened 2022-02-28
CI: move deps and lint jobs only on pull requests

MStoykov:
This was provoked as the new use of golangci-lint action fails on master
with all the lint errors there. But in reality it's probably not needed
to run those tasks on master - they are mostly about new code


olegbespalov (approved):

na-- (changes requested):
You need to remove `lint` from the `needs:` lists in all.yml

na-- (approved):

olegbespalov (approved):

PR #2407, opened 2022-03-05
http3 support, implemented with github.com/lucas-clemente/quic-go

daodao2007:
When running, cmd  -3 opens http3

Add a flag to cmd to support http3
Will save a __bhttp3 variable in js rt
If there is a __bhttp3 variable, use the http3 request in MakeRequest 

See the code for more details


MStoykov:
Hi @daodao2007, thanks for the PR :)

We are currently at the end of a release cycle and should release v0.37.0 next week. This also means that this will likely get fewer eyes right now. 


But I have a few, quick and unfortunately (IMO) completely blocking this PR from being merged, observations:
1. Using `--http3` in practice means that users can't use none http3 servers. IMO this should be handled similar to how the HTTP/1 and HTTP/2 are now - if available k6 should just use the highest version available. Arguably this should be configurable, so you can say you don't want a *specific* one but unless you do it, it should just go with w/e is highest (and supposedly best ;) ).
2. I did run this locally and there are no timings. It just seems like the use quic implementation does *not* support [httptrace](https://pkg.go.dev/net/http/httptrace). While the first point was more of UX thing - this in practice makes this change mostly useless as k6 is predominately a performance tool, and if we are not measuring how long a request takes ... what are we even doing :(. I have no idea if another library implements it or if there is something we need to configure. 

Again I don't know how to fix neither of those, but also I see no way we are merging without them :( . 

Can you please look into them and figure out if they are actually fixable and what will be needed?

daodao2007:
> 
OK, I'll see again

PR #2408, opened 2022-03-05
Close archives after writing them and fix test to not add untracked files to the repo

na--:
Running `go test` with the current `master` leaves a dangling `cmd/archive.tar` file that "dirties" the repo. 

The proper fix would be to not use the OS filesystem at all, by threading an `afero.Fs` through everything, but that should probably wait for after v0.37.0 is released. I'll check it out and make another PR if it's quick.

### DIFF
--- a/cmd/archive_test.go
+++ b/cmd/archive_test.go
@@ -40,10 +40,12 @@ func TestArchiveThresholds(t *testing.T) {
 		t.Run(testCase.name, func(t *testing.T) {
 			t.Parallel()
 
+			tmpPath := filepath.Join(t.TempDir(), "archive.tar")
+
 			cmd := getArchiveCmd(testutils.NewLogger(t), newCommandFlags())
 			filename, err := filepath.Abs(testCase.testFilename)
 			require.NoError(t, err)
-			args := []string{filename}
+			args := []string{filename, "--archive-out", tmpPath}
 			if testCase.noThresholds {
 				args = append(args, "--no-thresholds")
 			}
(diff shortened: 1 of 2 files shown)


cmd/archive_test.go
```diff
@@ -40,10 +41,17 @@ func TestArchiveThresholds(t *testing.T) {
 		t.Run(testCase.name, func(t *testing.T) {
 			t.Parallel()
 
+			tmpDir, err := os.MkdirTemp("", "k6archivetest")
+			require.NoError(t, err)
+			defer func() {
+				require.NoError(t, os.RemoveAll(tmpDir))
+			}()
+			tmpPath := filepath.Join(tmpDir, "archive.tar")
+
```
MStoykov, line 50: maybe use [t.TempDir](https://pkg.go.dev/testing#T.TempDir)?

olegbespalov: Join @MStoykov here `t.TempDir` seems cleaner

na--: :+1: very good point, forgot that existed :sweat_smile: Fixed by the [latest force push](https://github.com/grafana/k6/compare/42ff1bd3e62408450d7f61fb7df9eaca7aa023de..c1e305197a49aa38472db3fe8c4ba6289d9c852d)

na--:
btw I just noticed this naive comment here 
> I'll check it out and make another PR if it's quick.

it wasn't... :face_with_head_bandage:  :rofl:

MStoykov (approved):

MStoykov:
> it wasn't... face_with_head_bandage rofl

Yeah I also looked at it when I was doing the previous round of `cmd` refactorings :sweat:

oleiade (approved):
Nice one! 🦖 🎉

PR #2409, opened 2022-03-05
Simplify execution requirements calculation in `k6 inspect`

na--:
I tried to read https://github.com/grafana/k6/issues/1048 and honestly got lost in more than a few places... However, I am pretty sure this simplification is safe from reading through the code and thinking logically about it. There is no reason to create a runner and an execution scheduler to get the execution requirements a script. It should be enough to derive the `scenarios` and just use their `GetFullExecutionRequirements()` method, that's exactly what the execution scheduler does, after all... :sweat_smile:  https://github.com/grafana/k6/blob/1e28a3e1c8655d2118c823113e7a90ce44736cfe/core/local/local.go#L69

Anyway, tagged this as v0.38.0 since it doesn't really matter, but we should probably not introduce changes like this right before the release, even if they should not really change anything :sweat_smile: :crossed_fingers:

### DIFF
--- a/cmd/inspect.go
+++ b/cmd/inspect.go
@@ -30,7 +30,6 @@ import (
 	"github.com/spf13/afero"
 	"github.com/spf13/cobra"
 
-	"go.k6.io/k6/core/local"
 	"go.k6.io/k6/js"
 	"go.k6.io/k6/lib"
 	"go.k6.io/k6/lib/metrics"
@@ -57,7 +56,6 @@ func getInspectCmd(logger *logrus.Logger, globalFlags *commandFlags) *cobra.Comm
 				return err
 			}
 			registry := metrics.NewRegistry()
-			builtinMetrics := metrics.RegisterBuiltinMetrics(registry)
 
 			var b *js.Bundle
 			typ := globalFlags.runType
@@ -85,7 +83,7 @@ func getInspectCmd(logger *logrus.Logger, globalFlags *commandFlags) *cobra.Comm
 			inspectOutput := interface{}(b.Options)
 
 			if addExecReqs {
-				inspectOutput, err = addExecRequirements(b, builtinMetrics, registry, logger, globalFlags)
+				inspectOutput, err = addExecRequirements(b, logger, globalFlags)
 				if err != nil {
 					return err
 				}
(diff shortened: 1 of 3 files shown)


cmd/inspect.go
```diff
@@ -112,37 +110,24 @@ func getInspectCmd(logger *logrus.Logger, globalFlags *commandFlags) *cobra.Comm
...
-	logger *logrus.Logger, globalFlags *commandFlags) (interface{}, error) {
-	// TODO: after #1048 issue, consider rewriting this without a Runner:
-	// just creating ExecutionPlan directly from validated options
-
-	runner, err := js.NewFromBundle(logger, b, builtinMetrics, registry)
-	if err != nil {
-		return nil, err
-	}
-
+func addExecRequirements(b *js.Bundle, logger *logrus.Logger, globalFlags *commandFlags) (interface{}, error) {
```
oleiade, line 113: I would suggest taking the opportunity to document the function. Moreover, maybe we could find a more suitable name? In general, something is **add**ed **to** something, but in this case, the function _produces_ something instead?

oleiade: Also I've spotted an inconsistency in the naming of arguments: `b` is shorthand for its type:`Bundle`, but `logger` is the full name of its type: `Logger`. In this case, I would plead for renaming `b` to `bundle` for clarity, but I'm not hung up on either one or the other solution.

oleiade (approved):
👍🏻 besides a few non-blocking documentation-related coments 😸

cmd/inspect.go
na--: Good points :+1: The one about `bundle` was already addressed in latter PR by just not needing it anymore :sweat_smile: https://github.com/grafana/k6/blob/691be6b68ef829a0ccf514220d91d06befb2bc84/cmd/inspect.go#L79

About the rest, for purely practical purposes, I'd prefer to also fix them in a separate commit/PR and leave this PR for only the actual code simplification.

codebien (approved):
LGTM, just a minor request

js/runner.go
```diff
@@ -347,9 +347,10 @@ func (r *Runner) GetOptions() lib.Options {
 
 // IsExecutable returns whether the given name is an exported and
 // executable function in the script.
+//
+// TODO: completely remove this?
```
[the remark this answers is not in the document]
na--: Actually, I am not sure if we can remove this without a lot more refactoring :thinking: `Bundle.IsExecutable` is not available everywhere as a substitute... :thinking:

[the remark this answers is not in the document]
na--: I dug a little more here and realized that the need for the public method in `Bundle` disappears because of the other simplifications I did in https://github.com/grafana/k6/pull/2412 with `loadedTest` :tada: So I ended up reverting this part of the change there, see https://github.com/grafana/k6/pull/2412/commits/c3fec322bb00515b7c4643365a3e2f274414bdba

I think it's fine to merge this as it is and remove it in the other PR, given that it allows us to merge these PRs one by one instead of all at once.

yorugac:
@na-- post-factum just in case :slightly_smiling_face: 
I think this is a nice neat change :+1: And together with other `cmd` PRs yes, looks like a first step to #1048

PR #2410, opened 2022-03-05
Remove most direct access to the `os` package, support `NO_COLOR` and `K6_NO_COLOR`

na--:
This wasn't quite part of the plan, but several hours of yak shaving produced some nice results :sweat_smile: Anyway, `cmd/` looks much nicer and it's now finally safe to test with basically full internal integration tests :tada: 

That said, there is a very high chance of bugs and errors, since we don't yet have full test coverage of `cmd/`... :disappointed: So a careful review and PRs that add more tests are going to be quite welcome :bow: Ideally, we should try to merge this at the beginning of the v0.38.0 cycle, so we also have the maximum time to safely hit any bugs during our manual tests and work...

And I think there were actually some (potential) bugs that were fixed by this refactor. For example, these loggers: https://github.com/grafana/k6/blob/1e28a3e1c8655d2118c823113e7a90ce44736cfe/cmd/root.go#L163-L175
Did not actually use the wrapped `stdout` that is protected by a mutex: https://github.com/grafana/k6/blob/1e28a3e1c8655d2118c823113e7a90ce44736cfe/cmd/root.go#L97

With the last commit, this PR also closes https://github.com/grafana/k6/issues/2091 :tada:

### DIFF
--- a/cmd/archive.go
+++ b/cmd/archive.go
@@ -21,10 +21,6 @@
 package cmd
 
 import (
-	"os"
-
-	"github.com/sirupsen/logrus"
-	"github.com/spf13/afero"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 
@@ -33,7 +29,8 @@ import (
 	"go.k6.io/k6/lib/metrics"
 )
 
-func getArchiveCmd(logger *logrus.Logger, globalFlags *commandFlags) *cobra.Command {
+func getArchiveCmd(globalState *globalState) *cobra.Command { // nolint: funlen
+	archiveOut := "archive.tar"
 	// archiveCmd represents the archive command
 	archiveCmd := &cobra.Command{
 		Use:   "archive",
@@ -49,19 +46,22 @@ An archive is a fully self-contained test run, and can be executed identically e
   k6 run myarchive.tar`[1:],
 		Args: cobra.ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
-			src, filesystems, err := readSource(args[0], logger)
+			src, filesystems, err := readSource(globalState, args[0])
 			if err != nil {
 				return err
 			}
 
-			runtimeOptions, err := getRuntimeOptions(cmd.Flags(), buildEnvMap(os.Environ()))
+			runtimeOptions, err := getRuntimeOptions(cmd.Flags(), globalState.envVars)
 			if err != nil {
 				return err
 			}
 
 			registry := metrics.NewRegistry()
 			builtinMetrics := metrics.RegisterBuiltinMetrics(registry)
-			r, err := newRunner(logger, src, globalFlags.runType, filesystems, runtimeOptions, builtinMetrics, registry)
+			r, err := newRunner(
+				globalState.logger, src, globalState.flags.runType,
+				filesystems, runtimeOptions, builtinMetrics, registry,
+			)
 			if err != nil {
 				return err
 			}
@@ -70,9 +70,7 @@ An archive is a fully self-contained test run, and can be executed identically e
 			if err != nil {
 				return err
 			}
-			conf, err := getConsolidatedConfig(
-				afero.NewOsFs(), Config{Options: cliOpts}, r.GetOptions(), buildEnvMap(os.Environ()), globalFlags,
-			)
+			conf, err := getConsolidatedConfig(globalState, Config{Options: cliOpts}, r.GetOptions())
 			if err != nil {
 				return err
 			}
@@ -89,7 +87,7 @@ An archive is a fully self-contained test run, and can be executed identically e
 				}
 			}
 
-			_, err = deriveAndValidateConfig(conf, r.IsExecutable, logger)
+			_, err = deriveAndValidateConfig(conf, r.IsExecutable, globalState.logger)
 			if err != nil {
 				return err
 			}
@@ -101,7 +99,7 @@ An archive is a fully self-contained test run, and can be executed identically e
 
 			// Archive.
 			arc := r.MakeArchive()
-			f, err := os.Create(globalFlags.archiveOut)
+			f, err := globalState.fs.Create(archiveOut)
 			if err != nil {
 				return err
 			}
@@ -115,16 +113,16 @@ An archive is a fully self-contained test run, and can be executed identically e
 	}
 
 	archiveCmd.Flags().SortFlags = false
-	archiveCmd.Flags().AddFlagSet(archiveCmdFlagSet(globalFlags))
+	archiveCmd.Flags().AddFlagSet(archiveCmdFlagSet(&archiveOut))
 
 	return archiveCmd
 }
 
-func archiveCmdFlagSet(globalFlags *commandFlags) *pflag.FlagSet {
+func archiveCmdFlagSet(archiveOut *string) *pflag.FlagSet {
 	flags := pflag.NewFlagSet("", pflag.ContinueOnError)
 	flags.SortFlags = false
 	flags.AddFlagSet(optionFlagSet())
 	flags.AddFlagSet(runtimeOptionFlagSet(false))
-	flags.StringVarP(&globalFlags.archiveOut, "archive-out", "O", globalFlags.archiveOut, "archive output filename")
+	flags.StringVarP(archiveOut, "archive-out", "O", *archiveOut, "archive output filename")
 	return flags
 }
--- a/cmd/cloud.go
+++ b/cmd/cloud.go
@@ -27,7 +27,6 @@ import (
 	"errors"
 	"fmt"
 	"os"
-	"os/signal"
 	"path/filepath"
 	"strconv"
 	"sync"
@@ -35,8 +34,6 @@ import (
 	"time"
 
 	"github.com/fatih/color"
-	"github.com/sirupsen/logrus"
-	"github.com/spf13/afero"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 
@@ -50,7 +47,10 @@ import (
 )
 
 //nolint:funlen,gocognit,gocyclo,cyclop
-func getCloudCmd(ctx context.Context, logger *logrus.Logger, globalFlags *commandFlags) *cobra.Command {
+func getCloudCmd(globalState *globalState) *cobra.Command {
+	showCloudLogs := true
+	exitOnRunning := false
+
 	cloudCmd := &cobra.Command{
 		Use:   "cloud",
 		Short: "Run a test on the cloud",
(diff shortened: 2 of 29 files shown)


na--:
Forgot that this PR has a minor breaking change. I renamed the `--logformat` CLI flag to `--log-format`, to more closely match `--log-output`. The old flag still works, it just emits a deprecation warning.

Also, I noticed that in [the docs](https://k6.io/docs/using-k6/options/#logformat), we advertise that the option can be configured through `K6_LOGFORMAT`, but as far as I can see, that has never been the case before now, only the CLI flag was usable. Anyway, the now we have support for env vars, though the name is `K6_LOG_FORMAT`, again to match `K6_LOG_OUTPUT`.

na--:
With the last commit I had to fix a minor issue where an error in the CLI flags (the parsing of which happens before `setupLoggers()`) will result in an ugly log message with the format we use for non-TTY terminals, even if a user is using a TTY. As a nice side-effect, I also added a few lines of code to close https://github.com/grafana/k6/issues/2091 :sweat_smile:

oleiade:
Experimenting with this PR, and using both the `K6_NO_COLOR=` and `NO_COLOR=` statements, the colors don't seem to be disabled :-/

I might have missed something though 😸 
Capture d’écran 2022-03-07 à 15 03 41

cmd/root.go
```diff
@@ -50,96 +52,180 @@ const (
...
-		logrus.WithError(err).Warn("could not get config directory")
-		confDir = ".config"
-	}
-	defaultConfigFilePath := filepath.Join(
-		confDir,
-		"loadimpact",
-		"k6",
-		defaultConfigFileName,
-	)
+type globalState struct {
```
oleiade, line 66: Although private, could benefit from a documentation comment detailing its reason to be, in which context its meant to be used etc...

na--:
Fixed support for `NO_COLOR` and `K6_NO_COLOR` with https://github.com/grafana/k6/pull/2410/commits/5e14510c1ce5143e4818f4fe64744f3f22910b6c :man_facepalming: I can't wait to finally fix #883 once and for all. This PR is a nice step in that direction, but as it's obvious from this commit, it's still soooo easy to make minor mistakes that silently break something :disappointed:

cmd/root.go
na--: Duplicate commit?

oleiade:
It works 🎉 

Capture d’écran 2022-03-07 à 15 49 58

cmd/root.go
oleiade: Nope 😸 just copy pasted it as it applies too :)

na--: Resolved by https://github.com/grafana/k6/pull/2410/commits/55cd43ddcd30401a5f656604561a2204761d659d

cmd/root.go
```diff
@@ -50,96 +52,180 @@ const (
...
-	showCloudLogs         bool
-	runType               string
-	archiveOut            string
-	quiet                 bool
-	noColor               bool
-	address               string
-	outMutex              *sync.Mutex
-	stdoutTTY, stderrTTY  bool
-	stdout, stderr        *consoleWriter
+type globalFlags struct {
```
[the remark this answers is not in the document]
na--: Resolved by https://github.com/grafana/k6/pull/2410/commits/55cd43ddcd30401a5f656604561a2204761d659d

cmd/version.go
```diff
@@ -21,21 +21,19 @@
...
-func getVersionCmd() *cobra.Command {
+func getVersionCmd(globalState *globalState) *cobra.Command {
 	// versionCmd represents the version command.
 	versionCmd := &cobra.Command{
 		Use:   "version",
 		Short: "Show application version",
 		Long:  `Show the application version and exit.`,
 		Run: func(_ *cobra.Command, _ []string) {
-			fmt.Println("k6 v" + consts.FullVersion()) //nolint:forbidigo // we probably shouldn't do that though
+			fprintf(globalState.stdOut, "k6 v%s", consts.FullVersion())
```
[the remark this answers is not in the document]
na--: Resolved by https://github.com/grafana/k6/pull/2410/commits/298bcd9a88d3115e93ccf4e235c8ab1ea65a85e7

Will rebase and squash it in the first commit later, after the PR is :+1: and ready to merge, so I don't mess up the other PRs that depend on it

codebien:
>and it's not that I see the changes in these PRs as fragile

I meant which part of the code we think it's more exposed to be broken in the future if we wouldn't add tests for it. I didn't mean that the current changes are fragile.

Btw, I will add some on top of #2412, as you suggested, if I will identify something useful by reviewing and testing the PRs.

codebien (approved):
LGTM

na--:
I'm honestly not sure which part would benefit the most :sweat_smile: Maybe we can look at the code coverage by just running the existing `cmd` tests in https://github.com/grafana/k6/pull/2412 and go from there? :thinking:

oleiade (dismissed):
🎉 👍🏻

cmd/cloud.go
```diff
@@ -345,27 +357,20 @@ This will execute the test on the k6 cloud service. Use "k6 login cloud" to auth
 		},
 	}
 	cloudCmd.Flags().SortFlags = false
-	cloudCmd.Flags().AddFlagSet(cloudCmdFlagSet(globalFlags))
+	cloudCmd.Flags().AddFlagSet(cloudCmdFlagSet(&showCloudLogs, &exitOnRunning))
 	return cloudCmd
 }
 
-func cloudCmdFlagSet(globalFlags *commandFlags) *pflag.FlagSet {
+func cloudCmdFlagSet(showCloudLogs, exitOnRunning *bool) *pflag.FlagSet {
```
olegbespalov, line 364: It seems like we don't need pointers for these booleans, don't we? :thinking:

cmd/archive.go
```diff
@@ -115,16 +113,16 @@ An archive is a fully self-contained test run, and can be executed identically e
...
 
 	archiveCmd.Flags().SortFlags = false
-	archiveCmd.Flags().AddFlagSet(archiveCmdFlagSet(globalFlags))
+	archiveCmd.Flags().AddFlagSet(archiveCmdFlagSet(&archiveOut))
 
 	return archiveCmd
 }
 
-func archiveCmdFlagSet(globalFlags *commandFlags) *pflag.FlagSet {
+func archiveCmdFlagSet(archiveOut *string) *pflag.FlagSet {
```
olegbespalov, line 121: It seems like we don't need a pointer for this string, don't we? :thinking:

cmd/cloud.go
```diff
@@ -60,56 +60,71 @@ This will execute the test on the k6 cloud service. Use "k6 login cloud" to auth
...
 				if err != nil {
 					return fmt.Errorf("parsing K6_SHOW_CLOUD_LOGS returned an error: %w", err)
 				}
 				if !cmd.Flags().Changed("show-logs") {
-					globalFlags.showCloudLogs = showCloudLogsValue
+					showCloudLogs = showCloudLogsValue
 				}
 			}
+
+			if exitOnRunningEnv, ok := globalState.envVars["K6_EXIT_ON_RUNNING"]; ok {
```
olegbespalov, line 79: These two blocks look similar, it can be a small helper

olegbespalov:
General comment, what is the idea behind `globalState`? Is it a temporary thing, because to be honest, it seems scary since we are coupeling many of the structs with the global state? :thinking:

cmd/cloud.go
na--: they are similar, but it's just 2 blocks, and we plan to rewrite this whole config mess with croconf (hopefully soon)... this repetition can be extra incentive

cmd/archive.go
na--: we need it here, `flags.StringVarP()` expects a pointer, but I got completely rid of the parameter in https://github.com/grafana/k6/pull/2418 anyway :sweat_smile:

cmd/cloud.go
na--: same as the other comment, we need pointers here, but the arguments are removed completely in one of the follow-up PRs https://github.com/grafana/k6/pull/2418 because of some further refactoring

na--:
> General comment, what is the idea behind `globalState`? Is it a temporary thing, because to be honest, it seems scary since we are coupeling many of the structs with the global state? :thinking:

It is scary, but the question is "is is scarier than everything directly touching the `os` package"? And to that, I'd answer with a very emphatic "no" :sweat_smile:

I have some ideas for further improvements here, but so far this seems to work out very well and unlock very nice integration testing, see https://github.com/grafana/k6/pull/2412 for the follow-up :man_shrugging:

olegbespalov:
>is is scarier than everything directly touching the os package

My concern is that `globalState` aggregates not only the functionality of the os package interactions but also many other things.

na--:
> My concern is that `globalState` aggregates not only the functionality of the os package interactions but also many other things.

https://github.com/grafana/k6/blob/f82b364a41aaf1a5731b8d44f77601fbb4eab634/cmd/root.go#L80-L100

@olegbespalov, "many other things" seems like a bit of an exaggeration :confused: Out of these 16 things, 11 are `os` proxies directly. The `defaultFlags` and `flags` directly derived from `os` values and `defaultFlags` will hopefully disappear once we have croconf. The remaining 3 are in `ctx`, `logger` and `fallbackLogger` and are there purely for convenience. We'll have to pass them everywhere we pass `globalState` anyway, so why not include them in the struct?

na--:
@olegbespalov, any other comments or more specific suggestions?

cmd/root.go
```diff
@@ -50,96 +53,209 @@ const (
...
 
+	logger         *logrus.Logger
+	fallbackLogger logrus.FieldLogger
+}
+
+// Ideally, this should be the only function in the whole codebase where we use
+// global variables and functions from the os package. Anywhere else, things
+// like os.Stdout, os.Stderr, os.Stdin, os.Getenv(), etc. should be removed and
+// the respective properties of globalState used instead.
+func newGlobalState(ctx context.Context) *globalState {
```
MStoykov, line 108: Any particular reason we don't want to just provide `envs map[string]string` to this so we can drop `os.Get...` here as well? I guess we will still need access to os.Stdout and co , but this might be worth abstracting either way :shrug:

cmd/root.go
```diff
@@ -50,96 +53,209 @@ const (
...
+	stdOut := &consoleWriter{os.Stdout, colorable.NewColorable(os.Stdout), stdoutTTY, outMutex, nil}
+	stdErr := &consoleWriter{os.Stderr, colorable.NewColorable(os.Stderr), stderrTTY, outMutex, nil}
+
+	logger := &logrus.Logger{
+		Out: stdErr,
+		Formatter: &logrus.TextFormatter{
+			ForceColors: stderrTTY,
+			// This is a hack just for this env var, since we need the logger
+			// before we can parse all of the env vars...
+			DisableColors: !stderrTTY || os.Getenv("NO_COLOR") != "" || os.Getenv("K6_NO_COLOR") != "",
```
MStoykov, line 122: Doesn't this need ot check that `NO_COLOR` is set at all ?

MStoykov (dismissed):
LGTM as much as it can as that code is always hairy :grimacing:. 

I would've preferred more new tests for what previously wasn't testable.

cmd/root.go
na--: :thinking: Maybe, but I am somewhat reluctant to do it that way, doesn't seem super safe or practical... It's as easy to set `NO_COLOR=true` as it is to set `NO_COLOR=`, right? But in the reverse case, if you had a global `NO_COLOR=true` variable, it's not as easy to unset it. If we accepted empty strings values, you'd need to do this:
```sh
unset NO_COLOR
k6 run script.js
```

whereas with the current code, you can do this:
```
NO_COLOR="" k6 run script.js
```

cmd/cloud.go
```diff
@@ -345,27 +357,20 @@ This will execute the test on the k6 cloud service. Use "k6 login cloud" to auth
...
-	// We also need to explicitly set the default value for the usage message here, so setting
-	// K6_EXIT_ON_RUNNING=true won't affect the usage message
-	flags.Lookup("exit-on-running").DefValue = "false"
-
-	// read the comments above for explanation why this is done this way and what are the problems
-	flags.BoolVar(&globalFlags.showCloudLogs, "show-logs", globalFlags.showCloudLogs,
+	// TODO: Figure out a better way to handle the CLI flags
+	flags.BoolVar(exitOnRunning, "exit-on-running", *exitOnRunning,
+		"exits when test reaches the running status")
+	flags.BoolVar(showCloudLogs, "show-logs", *showCloudLogs,
```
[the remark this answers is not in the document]
na--: I have this fixed in the follow-up PRs, https://github.com/grafana/k6/pull/2412 and https://github.com/grafana/k6/pull/2418 specifically, see:
https://github.com/grafana/k6/blob/f82b364a41aaf1a5731b8d44f77601fbb4eab634/cmd/cloud.go#L47-L53
https://github.com/grafana/k6/blob/f82b364a41aaf1a5731b8d44f77601fbb4eab634/cmd/cloud.go#L322-L332

cmd/root.go
```diff
@@ -50,96 +53,209 @@ const (
...
+	loginCmd := getLoginCmd()
+	loginCmd.AddCommand(
+		getLoginCloudCommand(gs),
+		getLoginInfluxDBCommand(gs),
+	)
+	rootCmd.AddCommand(
+		getArchiveCmd(gs), getCloudCmd(gs), getConvertCmd(gs), getInspectCmd(gs),
+		loginCmd, getPauseCmd(gs), getResumeCmd(gs), getScaleCmd(gs), getRunCmd(gs),
+		getStatsCmd(gs), getStatusCmd(gs), getVersionCmd(gs),
+	)
```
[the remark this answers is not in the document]
na--: I also have this fixed in the follow-up PRs, https://github.com/grafana/k6/pull/2418 specifically, see:
https://github.com/grafana/k6/blob/f82b364a41aaf1a5731b8d44f77601fbb4eab634/cmd/root.go#L237-L245

cmd/root.go
na--: I could, but I kind of like that this is the only place in the code that touches the `os` package... passing the env vars here won't really have any effect besides a slightly cleaner internals. If you dislike using `os.Getenv` in `DisableColors`, I'd rather move up the this line higher in the function than pass it from outside;
```go
envVars := buildEnvMap(os.Environ())
```
but I don't think it matters :man_shrugging:

na--:
@MStoykov, re: more tests, it's still not super easy to write them in this PR, but check out https://github.com/grafana/k6/pull/2412 - it makes it much easier and adds several integration tests

cmd/root.go
MStoykov: Maybe,  but we both do it like that in the other place you check this *and* (probably more importantly) this is how `NO_COLOR` is defined to work "if it exists" not depending on it's value. 

Given that whoever will be using k6 is likely to set  `NO_COLOR` globally in order to disable color in everything, it will be kind of bad of us to not actually do the same thing as the other programs supporting https://no-color.org/ :shrug: . 


I really doubt there is a practical case where somebody having disabled colors *globally* will want to specifically enable them for us. 

Also, this is how it actually works because you handle `NO_COLOR` in another place as well. This only matters for early errors like 
image
Also note the format change :thinking:

cmd/root.go
MStoykov: I guess we can do this later if needed. 

Moving the remaining `os` things will in practice move most of the code from this function to w/e calls it, likely not improving anything, but the opposite

cmd/root.go
na--: I looked at some of the projects listed on https://no-color.org/ and they do exactly what I do here, check if NO_COLOR actually has a value. See [this](https://github.com/muesli/termenv/blob/0899c83a5e78cfd47f3077b860dc4a94f54ce55b/termenv.go#L77-L84) and [this](https://github.com/xyproto/textoutput/blob/dec1ab3255b039faa9afd059be44afa9026936e4/textoutput.go#L37) for example. So I don't think I should change it, I think the "spec" is just poorly written :man_shrugging: 

Regarding the format change, I think it's is just something `logurs.TextFormatter` does :thinking: In any case, if we decide to fix it, I think that should be a separate PR, this behavior is the same for the one we have for non-TTY terminals right now and maybe both need to be fixed :man_shrugging:

MStoykov: >  Command-line software which adds ANSI color to its output by default should check for the presence of a NO_COLOR environment variable that, when present (regardless of its value), prevents the addition of ANSI color.

MStoykov: Also, *again* this PR *already* disables colors if NO_COLOR is set to an empty value 


image
Just that happens a bit later, so it *doesn't* get disabled if k6 errors out early as I show in the previous screenshot. See also @oleiade [comment](https://github.com/grafana/k6/pull/2410#issuecomment-1060770737). 

This happens in [getFlags](https://github.com/grafana/k6/pull/2410/files#diff-ab967ab1a2f3a1b769106eeb7bfe892ef0e81d1d27811fa15be08e6749feee1fR187-R191)

MStoykov: As I mention I kind of prefer the `NO_COLOR= command` disabling the colors as I said. WDYT @oleiade @codebien @olegbespalov 

I am fine with this being removed and no actually add `NO_COLOR` if we do not agree on how it should behave

na--: hmm 3rd options: `NO_COLOR` will work without a value, but `K6_NO_COLOR` requires one? :sweat_smile:

na--: Resolved by the latest [force-push](https://github.com/grafana/k6/compare/ca4cc570c4a9de37022244dbae704e4fb3def926..d49634c76b9cabc355b79c167581534c85e753b7)?

MStoykov (approved):

cmd/root.go
oleiade: I'm not entirely sure if I understood what the different visions in this discussion are.

Thus, reading the https://no-color.org specs, and judging by how my favorite shell (fish), albeit a bit unconventional, handles this case, my preferred option is to have this behavior (which, I believe, is what this PR does as of now?):

```
# Colorful output
k6 run script.js

# No color output (set for the execution of this command only, the next invocation without the `NO_COLOR=` statement would be colorful)
NO_COLOR= k6 run script.js

# No color output (set for the execution of this command only, the next invocation without the `NO_COLOR=` statement would be colorful)
K6_NO_COLOR= k6 run script.js

# persistent no color (fish syntax)
set -x NO_COLOR
k6 run script.js  # no colors
k6 run script.js  # still no colors
set -e NO_COLOR
k6 run script.js # colorful
```

**N.B** interestingly, I noticed that fish shell has a different behavior (in appearance), compared to bash, when it comes to empty env variable:
```fish
> set -x NO_COLOR
> env | grep NO_COLOR
NO_COLOR=
```

oleiade (approved):
🎉 🍾 🍪

cmd/root.go
MStoykov: > # No color output (set for the execution of this command only, the next invocation without the `NO_COLOR=` statement would be colorful)
> K6_NO_COLOR= k6 run script.js

This will now be colorful with the current changes. But `K6_NO_COLOR=literallyanything` will including `K6_NO_COLOR=false`

Issue #1342, opened 2020-03-03
Test suites / execute multiple scripts with k6


na--:
I'm creating this issue as a place to discuss possible ways for k6 to execute multiple distinct scripts in a single run. This is prompted by https://community.k6.io/t/running-10-test-cases-out-of-100/448, and other similar previous user inquiries.

Something like this can currently be achieved by a simple shell script that sequentially calls `k6 run "$script_path"` for all files in a folder, but that workaround is not very convenient without an external output like InfluxDB. It also doesn't work well with `k6 cloud`, and isn't efficient in general (each separate `k6 run` has to re-initialize its VUs). [GitLab’s performance testing tool](https://about.gitlab.com/blog/2020/02/18/how-were-building-up-performance-testing-of-gitlab/) is probably also worth investigating...

I also have a vague idea how we can implement something like "test suites" natively in k6 after #1007 is done, and do it relatively easily (:tm: :sweat_smile: ). To get 80% of the way there, we just need to implement one minor (sort-of-already-planned) improvement on the currently planned #1007 feature set :tada:. Instead of specifying `startTime: "30s"` for an executor, users should be able to specify `startAfter: "someOtherExecutorName"`. This will be very easy to implement (minor changes in the config validation, [here](https://github.com/loadimpact/k6/blob/9cb4ed0356752da48d8a036bc178f3b8ab87c8d1/lib/executors.go#L261) and [here](https://github.com/loadimpact/k6/blob/9cb4ed0356752da48d8a036bc178f3b8ab87c8d1/core/local/local.go#L292)). 

This will add a lot of convenience in general, so it's probably worth doing just for that. But it will also probably satisfy the "test suite" requirements of the majority of our users, albeit with some manual configuration work (which we can ameliorate with some JS helper functions).

[...]

PR #2412, opened 2022-03-06
More `cmd/` refactoring to simplify test loading and support integration tests

na--:
More yak shaving, but after this, some parts of `cmd/` are almost normal looking :sweat_smile: :tada: I particularly like how clean `cmd/archive.go` became :tada: https://github.com/grafana/k6/blob/147e3d1beaf99aef837a3b30142c5728c0a6d17a/cmd/archive.go#L44-L72

Closes https://github.com/grafana/k6/issues/342

### DIFF
--- a/cmd/cloud.go
+++ b/cmd/cloud.go
@@ -30,7 +30,6 @@ import (
 	"path/filepath"
 	"strconv"
 	"sync"
-	"syscall"
 	"time"
 
 	"github.com/fatih/color"
@@ -190,13 +189,10 @@ This will execute the test on the k6 cloud service. Use "k6 login cloud" to auth
 			}
 
 			// Trap Interrupts, SIGINTs and SIGTERMs.
-			sigC := make(chan os.Signal, 2)
-			globalState.signalNotify(sigC, os.Interrupt, syscall.SIGINT, syscall.SIGTERM)
-			defer globalState.signalStop(sigC)
-			go func() {
-				sig := <-sigC
+			gracefulStop := func(sig os.Signal) {
 				logger.WithField("sig", sig).Print("Stopping cloud test run in response to signal...")
-				// Do this in a separate goroutine so that if it blocks the second signal can stop the execution
+				// Do this in a separate goroutine so that if it blocks, the
+				// second signal can still abort the process execution.
 				go func() {
 					stopErr := client.StopCloudTestRun(refID)
 					if stopErr != nil {
@@ -206,11 +202,12 @@ This will execute the test on the k6 cloud service. Use "k6 login cloud" to auth
 					}
 					globalCancel()
 				}()
-
-				sig = <-sigC
+			}
+			hardStop := func(sig os.Signal) {
 				logger.WithField("sig", sig).Error("Aborting k6 in response to signal, we won't wait for the test to end.")
-				os.Exit(int(exitcodes.ExternalAbort))
-			}()
+			}
+			stopSignalHandling := handleTestAbortSignals(globalState, gracefulStop, hardStop)
+			defer stopSignalHandling()
 
 			et, err := lib.NewExecutionTuple(test.derivedConfig.ExecutionSegment, test.derivedConfig.ExecutionSegmentSequence)
 			if err != nil {
--- a/cmd/cloud.go
+++ b/cmd/cloud.go
@@ -42,7 +42,6 @@ import (
 	"go.k6.io/k6/errext/exitcodes"
 	"go.k6.io/k6/lib"
 	"go.k6.io/k6/lib/consts"
-	"go.k6.io/k6/lib/metrics"
 	"go.k6.io/k6/ui/pb"
 )
 
@@ -91,56 +90,23 @@ This will execute the test on the k6 cloud service. Use "k6 login cloud" to auth
 		RunE: func(cmd *cobra.Command, args []string) error {
 			printBanner(globalState)
 
-			logger := globalState.logger
 			progressBar := pb.New(
 				pb.WithConstLeft("Init"),
-				pb.WithConstProgress(0, "Parsing script"),
+				pb.WithConstProgress(0, "Loading test script..."),
 			)
 			printBar(globalState, progressBar)
 
-			// Runner
-			filename := args[0]
-			src, filesystems, err := readSource(globalState, filename)
+			test, err := loadTest(globalState, cmd, args, getPartialConfig)
 			if err != nil {
 				return err
 			}
 
-			runtimeOptions, err := getRuntimeOptions(cmd.Flags(), globalState.envVars)
-			if err != nil {
-				return err
-			}
-
-			modifyAndPrintBar(globalState, progressBar, pb.WithConstProgress(0, "Getting script options"))
-			registry := metrics.NewRegistry()
-			builtinMetrics := metrics.RegisterBuiltinMetrics(registry)
-			r, err := newRunner(logger, src, globalState.flags.runType, filesystems, runtimeOptions, builtinMetrics, registry)
-			if err != nil {
-				return err
-			}
-
-			modifyAndPrintBar(globalState, progressBar, pb.WithConstProgress(0, "Consolidating options"))
-			cliOpts, err := getOptions(cmd.Flags())
-			if err != nil {
-				return err
-			}
-			conf, err := getConsolidatedConfig(globalState, Config{Options: cliOpts}, r.GetOptions())
-			if err != nil {
-				return err
-			}
-
-			// Parse the thresholds, only if the --no-threshold flag is not set.
-			// If parsing the threshold expressions failed, consider it as an
-			// invalid configuration error.
-			if !runtimeOptions.NoThresholds.Bool {
-				for _, thresholds := range conf.Options.Thresholds {
-					err = thresholds.Parse()
-					if err != nil {
-						return errext.WithExitCodeIfNone(err, exitcodes.InvalidConfig)
-					}
-				}
-			}
-
-			derivedConf, err := deriveAndValidateConfig(conf, r.IsExecutable, logger)
+			// It's important to NOT set the derived options back to the runner
+			// here, only the consolidated ones. Otherwise, if the script used
+			// an execution shortcut option (e.g. `iterations` or `duration`),
+			// we will have multiple conflicting execution options since the
+			// derivation will set `scenarios` as well.
+			err = test.initRunner.SetOptions(test.consolidatedConfig.Options)
 			if err != nil {
 				return err
 			}
@@ -149,13 +115,9 @@ This will execute the test on the k6 cloud service. Use "k6 login cloud" to auth
 			// TODO: validate for externally controlled executor (i.e. executors that aren't distributable)
 			// TODO: move those validations to a separate function and reuse validateConfig()?
 
-			err = r.SetOptions(conf.Options)
-			if err != nil {
-				return err
-			}
+			modifyAndPrintBar(globalState, progressBar, pb.WithConstProgress(0, "Building the archive..."))
+			arc := test.initRunner.MakeArchive()
 
-			modifyAndPrintBar(globalState, progressBar, pb.WithConstProgress(0, "Building the archive"))
-			arc := r.MakeArchive()
 			// TODO: Fix this
 			// We reuse cloud.Config for parsing options.ext.loadimpact, but this probably shouldn't be
 			// done, as the idea of options.ext is that they are extensible without touching k6. But in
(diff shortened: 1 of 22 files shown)


na--:
With the latest commit here, we can now write dead-simple **and safe** integration tests :tada: See this for an example: https://github.com/grafana/k6/blob/441a8f0749f232cf7b1cff6a07e99c876555c5f6/cmd/integration_test.go#L24-L37

So between this and the fact that we have xk6 integration tests in our CI, I think it's pretty safe to close https://github.com/grafana/k6/issues/342 once this PR is merged :tada:

na--:
And yeah, I consider these to be integration tests because this is what `main.go` does now:
https://github.com/grafana/k6/blob/441a8f0749f232cf7b1cff6a07e99c876555c5f6/main.go#L23-L29
https://github.com/grafana/k6/blob/441a8f0749f232cf7b1cff6a07e99c876555c5f6/cmd/root.go#L297-L301

And this is what an integration test in `cmd/` can do:
https://github.com/grafana/k6/blob/441a8f0749f232cf7b1cff6a07e99c876555c5f6/cmd/integration_test.go#L27-L30

The two are basically the same, only the integration test has a dedicated environment (so we can run it safely in parallel) and also complete control that environment :tada:

cmd/test_load.go
```diff
@@ -0,0 +1,199 @@
+package cmd
```
oleiade, line 1: While reading this code I noticed there are a lot of Homonyms, which made it quite confusing for me. The main helper function is called `loadTest`, which I intuitively expected to: perform a load test. 

My understanding is that what it does is actually load the configuration, state, and perform setups that are prerequisites to perform a test. Thus, **I would suggest a different verb than *load* in order to avoid confusion?**
Maybe `prepare`, or `configure`, or `assemble`: `prepareTest`, `configureTest`, or `assembleTest`? We could even prepend those verbs to the existing name: `prepareLoadTest`, `configureLoadTest`, or `assembleLoadTest`? what do you think?

**The term `test` is also equivoqual**, because it's so polysemic. Personally, when I read it, I immediately think of a unit/integration test, but used in the context of a load testing software, it also binds to the idea of running a load test. In this specific context, the best idea I can come up with, paired with the idea of dropping the term `load` to avoid homonymy would be to use something like `run` or `testrun`, or `loadtestrun` instead (something along those lines, to explicit its nature). What do you think?

Another suggestion I have is to use this opportunity to document every symbols we introduce here properly. Even if they are private, I think we would benefit from having doc for most of them (at least types and functions). That would make our future selves and future team members lives easy. Even just two sentences to explain what the function/symbol does, and how/where it's meant to be used (its reason to be).

cmd/test_load.go
```diff
@@ -0,0 +1,199 @@
...
+func loadTest(
+	gs *globalState, cmd *cobra.Command, args []string,
+	// supply this if you want the test config consolidated and validated
+	cliConfigGetter func(flags *pflag.FlagSet) (Config, error), // TODO: obviate
+) (*loadedTest, error) {
+	if len(args) < 1 {
+		return nil, fmt.Errorf("k6 needs at least one argument to load the test")
+	}
+
+	testPath := args[0]
```
oleiade, line 48: In the spirit of my comment about naming, I would argue that this could be called `sourcePath`, as it's only used for logging and in the context of `readSource`.

cmd/test_load.go
```diff
@@ -0,0 +1,199 @@
...
+	"go.k6.io/k6/loader"
+)
+
+const (
+	testTypeJS      = "js"
+	testTypeArchive = "archive"
+)
+
+type loadedTest struct {
+	testPath        string // contains the raw string the user supplied
```
oleiade, line 25: Considering this really the path to a resource that can either be a test script or an archive, what would you think of naming it `sourcePath` instead?

cmd/test_load.go
```diff
@@ -0,0 +1,199 @@
...
+	"go.k6.io/k6/lib/metrics"
+	"go.k6.io/k6/loader"
+)
+
+const (
+	testTypeJS      = "js"
+	testTypeArchive = "archive"
+)
+
+type loadedTest struct {
```
oleiade, line 24: My top-level package comment about naming would probably apply to this struct too 😄

cmd/test_load.go
```diff
@@ -0,0 +1,199 @@
...
+	if cliConfigGetter != nil {
+		if err := test.consolidateDeriveAndValidateConfig(gs, cmd, cliConfigGetter); err != nil {
+			return nil, err
+		}
+	}
+
+	return test, nil
+}
+
+func (lt *loadedTest) initializeFirstRunner(gs *globalState) error {
```
oleiade, line 88: The name of this function led to me to an immediate load of questions: "why do we load a **first** runner", "how is the **first** runner different from the other runners**. Which leads me to suggest to detail that in the documentation of the function 😸

cmd/test_load.go
```diff
@@ -0,0 +1,199 @@
...
+	consolidatedConfig, err := getConsolidatedConfig(gs, cliConfig, lt.initRunner.GetOptions())
+	if err != nil {
+		return err
+	}
+
+	gs.logger.Debug("Parsing thresholds and validating config...")
+	// Parse the thresholds, only if the --no-threshold flag is not set.
+	// If parsing the threshold expressions failed, consider it as an
+	// invalid configuration error.
+	if !lt.runtimeOptions.NoThresholds.Bool {
```
oleiade, line 181: What I like about this PR, is that it makes access the whole configuration, and especially `runtimeOptions` much easier without having to modify a bunch of APIs. Kudos for that 👍

cmd/runtime_options.go
```diff
@@ -86,11 +88,13 @@ func getRuntimeOptions(flags *pflag.FlagSet, environment map[string]string) (lib
...
 
-	if envVar, ok := environment["K6_COMPATIBILITY_MODE"]; ok {
+	if envVar, ok := environment["K6_TYPE"]; ok && !opts.TestType.Valid {
 		// Only override if not explicitly set via the CLI flag
-		if !opts.CompatibilityMode.Valid {
-			opts.CompatibilityMode = null.StringFrom(envVar)
-		}
+		opts.TestType = null.StringFrom(envVar)
+	}
+	if envVar, ok := environment["K6_COMPATIBILITY_MODE"]; ok && !opts.CompatibilityMode.Valid {
```
oleiade, line 95: Out of curiosity, what is k6's "compatibility mode"?

oleiade (changes requested):
Thumbs up from me :+1: :tada: Kudos for those changes, I do believe they will make our lives easier indeed!

I do think it would be beneficial to find better names for some of the APIs introduced in this PR; see my comments smiley 🐱

cmd/runtime_options.go
MStoykov: https://k6.io/docs/using-k6/javascript-compatibility-mode does this answer your question?

oleiade: It does! Thanks @MStoykov !

cmd/test_load.go
na--: :thinking: you raise some good points, but as always, naming things well is going to be hard... :sweat_smile: 

`load` can indeed be confusing... :disappointed: Even though, AFAIK, we don't use any variations of the term "load testing" or "load test" internally anywhere in the k6 codebase, we use it in our marketing website and documentation, so contributors could be confused.

On the other hand, `configure`, `assemble`, etc. don't reflect the reality of what is happening under the hood of that function, so they can be confusing in that way. So my suggestion is to rename this function to `loadAndConfigureTest()`, which both accurately reflects what's going on, and doesn't allow any conflation with "load testing", it's very clearly about "loading a test". WDYT?

Regarding the term "test", I think it's unavoidable and already pretty well defined in the context of k6. k6 is a tool that runs tests. And execution `k6 run test.js` is a "testrun". And with future support for test suites (https://github.com/grafana/k6/issues/1342, something that was one of the goal of my recent PRs to unlock), k6 will be able to run multiple tests with a single `k6 run`, e.g. `k6 run test1.js test2.js`. So the term will be even more clearly defined. 

While there might be a small confusion for a new developer between a k6 "test" and a Go test of k6 itself that lives in a `_test.go` file, that seems like the lesser evil and something that will be clear after the first few minutes of looking at the code.

cmd/test_load.go
na--: `rootSourcePath` or `sourceRootPath` will probably be even better, considering a script can `import` other files :thinking:

cmd/test_load.go
na--: Hmm yeah, though the bigger problem is that `Runner` shouldn't even be called `Runner` :sweat_smile: It's a terrible name and something we've wanted to rename for a long, **long** time, see https://github.com/grafana/k6/blame/3db46be4b1f4591a3ae85b3d4fd863bb94e37165/lib/runner.go#L38

A VUFactory will be a better name, but still quite terrible, since a `Runner` does other things besides creating VUs :sob: So it first actually needs a complete refactor/rewrite, probably to split apart its responsibilities in 2 or 3 different clear and well-defined concepts and interfaces, which we can then name nicely... That is a task for another time though, refactoring the similarly unclear `Engine` was enough for this series of PRs :sweat_smile: Until that future time, I chose to keep the current confusing nomenclature about `Runner` and just sprinkle some TODOs https://github.com/grafana/k6/blob/207ef83e62b7ac097b6f024e9ca5a8ff56d93ba0/cmd/test_load.go#L31

I will add a comment to this method though, to explain what is happening in it and why :+1:

cmd/test_load.go
oleiade: All in on `loadAndConfigureTest` 👍🏻 
Fair enough about the usage of "test" 👍🏻 🙇🏻

oleiade (dismissed):
Besides my few comments on naming, looking good! The yak is shaved, and we're equipped to make ourselves a nice pull-over for the next winter. Kudos 🎉

cmd/test_load.go
na--: Resolved by https://github.com/grafana/k6/pull/2412/commits/c0bc707e030326ae2038db0940071b27493487f7

cmd/test_load.go
na--: Resolved by https://github.com/grafana/k6/pull/2412/commits/6ffdb958f5e13bdc77fa7ef317af29c7be9b55a0

cmd/test_load.go
na--: Resolved by https://github.com/grafana/k6/pull/2412/commits/6ffdb958f5e13bdc77fa7ef317af29c7be9b55a0

oleiade (approved):

cmd/test_load.go
```diff
@@ -0,0 +1,199 @@
...
+)
+
+type loadedTest struct {
+	testPath        string // contains the raw string the user supplied
+	source          *loader.SourceData
+	fileSystems     map[string]afero.Fs
+	runtimeOptions  lib.RuntimeOptions
+	metricsRegistry *metrics.Registry
+	builtInMetrics  *metrics.BuiltinMetrics
+	initRunner      lib.Runner // TODO: rename to something more appropriate
```
MStoykov, line 33: Just `runner` I mean it seems to be the *only* runner either way.

cmd/cloud.go
```diff
@@ -206,11 +202,12 @@ This will execute the test on the k6 cloud service. Use "k6 login cloud" to auth
 					}
 					globalCancel()
 				}()
-
-				sig = <-sigC
+			}
+			hardStop := func(sig os.Signal) {
 				logger.WithField("sig", sig).Error("Aborting k6 in response to signal, we won't wait for the test to end.")
-				os.Exit(int(exitcodes.ExternalAbort))
-			}()
```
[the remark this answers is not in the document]
MStoykov: Or maybe just have `os.Exit` in both hardStops :shrug: I kind of feel like this DRY-ing of the code isn't really worth it

cmd/test_load.go
na--: Maybe, but I'd prefer to wait with the final refactor and rename before messing with the names more...

cmd/cloud.go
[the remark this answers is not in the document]
na--: :+1: for `gracefulStop` and `onHardStop`
Regarding the DRY, keep in mind that this will be used at least 2 more times, in `k6 coordinate` and `k6 agent`

cmd/common.go
```diff
@@ -87,3 +90,36 @@ func printToStdout(gs *globalState, s string) {
...
+			firstHandler(sig)
+		case <-done:
+			return
+		}
+
+		select {
+		case sig := <-sigC:
+			if secondHandler != nil {
+				secondHandler(sig)
+			}
```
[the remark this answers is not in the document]
na--: Yes, I think so, if `os.Exit()` is here. And I kind of like it to be here, to softly force the user to acknowledge that a second Ctrl+C should _always_ abort the execution almost immediately, not be something optional.

cmd/cloud.go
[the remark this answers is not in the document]
na--: (for expediency, I will address this in the next PR)

Issue #2413, opened 2022-03-07
`--log-output=file` does not always flush all logs to the file after k6 is done


na--:
We don't have the same "done" channel for `file` log outputs that we do for `loki`: https://github.com/grafana/k6/blob/1e28a3e1c8655d2118c823113e7a90ce44736cfe/cmd/root.go#L292-L309

So this code is not guaranteed to run: https://github.com/grafana/k6/blob/1e28a3e1c8655d2118c823113e7a90ce44736cfe/log/file.go#L129-L136

I found this while writing some integration tests so I have a bugfix that I will submit shortly.

PR #2414, opened 2022-03-07
Wait for file log to drain before k6 exits

na--:
This closes https://github.com/grafana/k6/issues/2413, a minor deficiency that slipped by in https://github.com/grafana/k6/pull/2285 

The integration test for this is unfortunately will have to wait until https://github.com/grafana/k6/pull/2412 is merged. That is in fact how I found the issue, tried to write an integration for sending logs to a file and it failed because the file didn't contain the expected text at the end of the test :sweat_smile: So we'll have to merge this in v0.37.0 without a test :disappointed:

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -290,7 +290,7 @@ func (c *rootCommand) setupLoggers() (<-chan struct{}, error) {
 		c.logger.SetOutput(ioutil.Discard)
 
 	case strings.HasPrefix(line, "loki"):
-		ch = make(chan struct{})
+		ch = make(chan struct{}) // TODO: refactor, get it from the constructor
 		hook, err := log.LokiFromConfigLine(c.ctx, c.fallbackLogger, line, ch)
 		if err != nil {
 			return nil, err
@@ -300,7 +300,8 @@ func (c *rootCommand) setupLoggers() (<-chan struct{}, error) {
 		c.logFmt = "raw"
 
 	case strings.HasPrefix(line, "file"):
-		hook, err := log.FileHookFromConfigLine(c.ctx, c.fallbackLogger, line)
+		ch = make(chan struct{}) // TODO: refactor, get it from the constructor
+		hook, err := log.FileHookFromConfigLine(c.ctx, c.fallbackLogger, line, ch)
 		if err != nil {
 			return nil, err
 		}
--- a/log/file.go
+++ b/log/file.go
@@ -44,15 +44,19 @@ type fileHook struct {
 	w              io.WriteCloser
 	bw             *bufio.Writer
 	levels         []logrus.Level
+	done           chan struct{}
 }
 
 // FileHookFromConfigLine returns new fileHook hook.
 func FileHookFromConfigLine(
-	ctx context.Context, fallbackLogger logrus.FieldLogger, line string,
+	ctx context.Context, fallbackLogger logrus.FieldLogger, line string, done chan struct{},
 ) (logrus.Hook, error) {
+	// TODO: fix this so it works correctly with relative paths from the CWD
+
 	hook := &fileHook{
 		fallbackLogger: fallbackLogger,
 		levels:         logrus.AllLevels,
+		done:           done,
 	}
 
 	parts := strings.SplitN(line, "=", 2)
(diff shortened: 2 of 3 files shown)


log/file.go
```diff
@@ -44,15 +44,19 @@ type fileHook struct {
...
 	levels         []logrus.Level
+	done           chan struct{}
 }
 
 // FileHookFromConfigLine returns new fileHook hook.
 func FileHookFromConfigLine(
-	ctx context.Context, fallbackLogger logrus.FieldLogger, line string,
+	ctx context.Context, fallbackLogger logrus.FieldLogger, line string, done chan struct{},
 ) (logrus.Hook, error) {
+	// TODO: fix this so it works correctly with relative paths from the CWD
```
MStoykov, line 54: what do you mean by this?

log/file.go
```diff
@@ -135,6 +139,7 @@ func (h *fileHook) loop(ctx context.Context) chan []byte {
 					h.fallbackLogger.Errorf("failed to close logfile: %w", err)
 				}
 
+				close(h.done)
```
MStoykov, line 142: I think this should be in a defer, WDYT?

cmd/root.go
```diff
@@ -300,7 +300,8 @@ func (c *rootCommand) setupLoggers() (<-chan struct{}, error) {
 		c.logFmt = "raw"
 
 	case strings.HasPrefix(line, "file"):
-		hook, err := log.FileHookFromConfigLine(c.ctx, c.fallbackLogger, line)
+		ch = make(chan struct{})
```
olegbespalov, line 303: Don't get it, where and how do we use it? :thinking:

log/file.go
```diff
@@ -44,15 +44,19 @@ type fileHook struct {
 	w              io.WriteCloser
 	bw             *bufio.Writer
 	levels         []logrus.Level
+	done           chan struct{}
 }
 
 // FileHookFromConfigLine returns new fileHook hook.
 func FileHookFromConfigLine(
-	ctx context.Context, fallbackLogger logrus.FieldLogger, line string,
+	ctx context.Context, fallbackLogger logrus.FieldLogger, line string, done chan struct{},
```
olegbespalov, line 52: This done channel looks more like the internal state. Do you think we should accept it as input?

Can a signature be like?

```go
FileHookFromConfigLine(ctx context.Context, fallbackLogger logrus.FieldLogger, line string) (logrus.Hook, chan struct{},error)`
```

Wdyt?

olegbespalov: Ideally, even with the doneCh `<-chan sruct{},` but I guess it is outside of the scope of this PR

na--: hmm yeah, that is slightly better :+1:

log/file.go
na--: Noticed that we don't explicitly check if we got a relative path here and instead just open exactly what we were given: https://github.com/grafana/k6/blob/1e28a3e1c8655d2118c823113e7a90ce44736cfe/log/file.go#L108

This works, for now, since we're directly using the `os` package and it will properly handle relative paths based on the current working directory. However, after we switch to `afero.Fs` here, we will need to explicitly account for it.

log/file.go
na--: actually, I think I'll refactor both this and `LokiFromConfigLine` in a separate PR, so I'll leave it as it is for now

log/file.go
na--: resolved with the latest force push

olegbespalov (approved):
:+1:

log/file.go
MStoykov: can you expand this comment there with that info :)

MStoykov (approved):

log/file.go
na--: No need, I plan to fix it with a small commit in https://github.com/grafana/k6/pull/2412 :sweat_smile:

na--: FWIW, this is the fix and the test to verify it: https://github.com/grafana/k6/pull/2412/commits/7574bd2c2488a7f6a3bc804a8f12aede47383b17

Issue #2390, opened 2022-02-15
Custom metric threshold calculation using wrong statistics


timterrill:
### Brief summary

When experimenting with the differences between custom metric thresholds and thresholds on tagged requests, I discovered that the custom metric thresholds were not using the custom metric data, they were using the overall statistics from the entire test (http_req_duration).  

This caused thresholds for the custom metric to fail when the tagged requests passed.  In my test example I have 2 requests, each of which has a unique custom metric and a unique tag:
- homepage request:
  - custom metric = trend_homepage
  - request tag = name:tag_homepage
- pi request
  - custom metric = trend_pi
  - request tag = name:tag_pi

 The metrics for the tag and custom metric should be the same, therefore the thresholds for each request should both pass or fail.

### k6 version

k6 v0.34.1 ((devel), go1.17, darwin/amd64)

### OS

macOS 12.1

### Steps to reproduce the problem

To reproduce this issue, 

1. Download https://gist.github.com/timterrill/e91c36a579916b697cb7af061bbcc432
2. Run with the following command:

$> k6 run k6-threshold-bug.js

NOTE: Depending on the actual response times of the 2 endpoints, the threshold may not actually fail, but repeated runs should reproduce it.

### Expected behaviour

The output I expect from this run is as follows (I modified the 2 rows for `{ trend_XXXX }` to be what I expected):

...
```
   ✓ http_req_duration..............: count=217 min=18.19ms avg=21.26ms  p(98)=26.92ms  p(99.9)=31.21ms  max=31.69ms
       { expected_response:true }...: count=217 min=18.19ms avg=21.26ms  p(98)=26.92ms  p(99.9)=31.21ms  max=31.69ms
     ✓ { name:tag_homepage }........: count=214 min=18.19ms avg=21.25ms  p(98)=26.96ms  p(99.9)=31.22ms  max=31.69ms
     ✓ { name:tag_pi }..............: count=3   min=20.92ms avg=21.41ms  p(98)=22.09ms  p(99.9)=22.12ms  max=22.12ms
     ✓ { trend_homepage }...........: count=214 min=18.19ms avg=21.25ms  p(98)=26.96ms  p(99.9)=31.22ms  max=31.69ms
     ✓ { trend_pi }.................: count=3   min=20.92ms avg=21.41ms  p(98)=22.09ms  p(99.9)=22.12ms  max=22.12ms
```
...
```
[...]

Issue #1889, opened 2021-03-08
Remove the `Engine` and split metrics/VUs handling


na--:
The [current `Engine`](https://github.com/loadimpact/k6/blob/2036baeaa83d76ce17aae95fef7e78e3ef24c735/core/engine.go#L46) vaguely resembles a Rube Goldberg machine... It has slowly been getting better, for example https://github.com/loadimpact/k6/pull/1869 improved it slightly by mostly decoupling the outputs from the `Engine`, but it still has quite a lot of moving parts that can be streamlined and improved. And as https://github.com/loadimpact/k6/issues/1887, it still has some gotchas and potentially some other hidden bugs or data races....

I'm opening this issue as an epic where we can gather Engine issues and ideas for fixing them. Here are some specific problems that currently exist, though there are quite a few other TODOs sprinkled in the code already:
- the `engineWait` and `engineRun` callbacks are more than a bit strange, but they were the easiest way to handle the `--linger` peculiarities: https://github.com/loadimpact/k6/blob/2036baeaa83d76ce17aae95fef7e78e3ef24c735/cmd/run.go#L255
- stopping and starting outputs doesn't really need to be in the Engine anymore, so it probably shouldn't be: https://github.com/loadimpact/k6/blob/2036baeaa83d76ce17aae95fef7e78e3ef24c735/core/engine.go#L133-L134
- the fix for https://github.com/loadimpact/k6/issues/1887 is a bit sketchy and it might be a good idea to add a timeout for it (https://github.com/loadimpact/k6/pull/1888#discussion_r589249165), or maybe completely refactor the way metrics are processed, so we don't need such hacks. For example, maybe decouple metric ingestion and metric processing, sort of how outputs do it?

Needless to say, but this refactoring should definitely be done after https://github.com/loadimpact/k6/issues/1832, and maybe even some of its related issues, given that they will probably substantially change the requirements.

na--:
I've been thinking on some related issues to this problem recently and I think I was wrong :arrow_up:, especially with this:
> stopping and starting outputs doesn't really need to be in the Engine anymore, so it probably shouldn't be

I think it might actually be the other way around :sweat_smile: Instead of refactoring metrics handling or outputs out of the `Engine`, we should probably refactor the test execution out of it :bulb: It is actually the thing that doesn't really fit all that well. Most of it is handled by the `ExecutionScheduler` anyway, the `Engine` just gets in the way. 

[...]

PR #2426, opened 2022-03-09
Use the registry for (sub-)metric validation and move data crunching out of the `Engine`

na--:
This PR is the first part of https://github.com/grafana/k6/issues/1889. It aims to prepare the `Engine` for complete annihilation :sweat_smile: I want to get completely rid of it and use its constituent parts separately instead, basically to almost completely decouple the test execution from the metrics handling. 

However, the `Engine` has tons of tests that would have to be re-written. So this PR is an attempt to softly split the `Engine` by moving things around and defining and implementing its replacement components, while still keeping them under the same `Engine` facade. This way we can have a reasonably high assurance we haven't broken anything, since all of the old tests still work with minimal adjustment. Hopefully, this will allow us to merge things gradually, not have to do it all at once.

Given the extensive refactoring here, there were a few opportunities to fix some bugs and side-issues along the way:
- commit 'Refactor the Engine to actually use the metrics registry` (https://github.com/grafana/k6/commit/adc4d9f6683bb2252df7fe5375981fae221116b9) implements a lot of the requirements of https://github.com/grafana/k6/issues/2330 when it comes to validating that thresholds are defined on valid metrics and sub-metrics
- commit `Fix the bug of thresholds not working for unused metrics` (https://github.com/grafana/k6/commit/e5d8c325f07d96186b576b837ff181bfbf15315b) fixes https://github.com/grafana/k6/issues/1346
- commit `Fix submetric matching bug when nonexistent keys are specified` (https://github.com/grafana/k6/commit/a3138b13eec8d34414df17d6214962aa6329a75b) fixes https://github.com/grafana/k6/issues/2390
- commit `Move the Engine data crunching logic in a new component under metrics` (https://github.com/grafana/k6/commit/2476cfc8d2640adbadf2f81fff8ec7c3adf53bd8) goes a long way towards closing https://github.com/grafana/k6/issues/1889, since now the `MetricsEngine` gets its metrics like just another output, almost no special handling when it comes to it :tada: 

Overall, I am not sure if it will be easier to review this commit by commit or all at once. The commits are logically separated, but a lot of the same code is changed in multiple commits because of that :man_shrugging:


na--:
In the CI for https://github.com/grafana/k6/issues/2428, which is built on top of this one and just renames things, there was this data race that needs to be tracked down and fixed (most likely) here:
```
==================
WARNING: DATA RACE
Read at 0x00c00006c388 by goroutine 178:
  go.k6.io/k6/metrics.(*TrendSink).Calc()
      /home/runner/work/k6/k6/metrics/sink.go:134 +0x47
  go.k6.io/k6/metrics.(*TrendSink).P()
      /home/runner/work/k6/k6/metrics/sink.go:124 +0xc9
  go.k6.io/k6/metrics.TestGetResolversForTrendColumnsCalculation.func1()
      /home/runner/work/k6/k6/metrics/sample_test.go:155 +0x207
  testing.tRunner()
      /opt/hostedtoolcache/go/1.16.14/x64/src/testing/testing.go:1203 +0x202

Previous write at 0x00c00006c388 by goroutine 173:
  go.k6.io/k6/metrics.(*TrendSink).Calc()
      /home/runner/work/k6/k6/metrics/sink.go:139 +0xb2
  go.k6.io/k6/metrics.(*TrendSink).P()
      /home/runner/work/k6/k6/metrics/sink.go:124 +0xc9
  go.k6.io/k6/metrics.TestGetResolversForTrendColumnsCalculation.func1()
      /home/runner/work/k6/k6/metrics/sample_test.go:155 +0x207
  testing.tRunner()
      /opt/hostedtoolcache/go/1.16.14/x64/src/testing/testing.go:1203 +0x202

Goroutine 178 (running) created at:
  testing.(*T).Run()
      /opt/hostedtoolcache/go/1.16.14/x64/src/testing/testing.go:1248 +0x5d7
  go.k6.io/k6/metrics.TestGetResolversForTrendColumnsCalculation()
      /home/runner/work/k6/k6/metrics/sample_test.go:148 +0x336
  testing.tRunner()
      /opt/hostedtoolcache/go/1.16.14/x64/src/testing/testing.go:1203 +0x202

Goroutine 173 (running) created at:
  testing.(*T).Run()
      /opt/hostedtoolcache/go/1.16.14/x64/src/testing/testing.go:1248 +0x5d7
  go.k6.io/k6/metrics.TestGetResolversForTrendColumnsCalculation()
      /home/runner/work/k6/k6/metrics/sample_test.go:148 +0x336
  testing.tRunner()
      /opt/hostedtoolcache/go/1.16.14/x64/src/testing/testing.go:1203 +0x202
==================
==================
WARNING: DATA RACE
Read at 0x00c000098718 by goroutine 178:
  go.k6.io/k6/metrics.(*TrendSink).P()
      /home/runner/work/k6/k6/metrics/sink.go:127 +0x1e7
  go.k6.io/k6/metrics.TestGetResolversForTrendColumnsCalculation.func1()
      /home/runner/work/k6/k6/metrics/sample_test.go:155 +0x207
  testing.tRunner()
      /opt/hostedtoolcache/go/1.16.14/x64/src/testing/testing.go:1203 +0x202

Previous write at 0x00c000098718 by goroutine 173:
  sort.Float64Slice.Swap()
      /opt/hostedtoolcache/go/1.16.14/x64/src/sort/sort.go:305 +0x10d
  sort.(*Float64Slice).Swap()
      <autogenerated>:1 +0x2e
  sort.medianOfThree()
      /opt/hostedtoolcache/go/1.16.14/x64/src/sort/sort.go:94 +0xcb
  sort.doPivot()
      /opt/hostedtoolcache/go/1.16.14/x64/src/sort/sort.go:116 +0x676
  sort.quickSort()
      /opt/hostedtoolcache/go/1.16.14/x64/src/sort/sort.go:203 +0xa5
  sort.Sort()
      /opt/hostedtoolcache/go/1.16.14/x64/src/sort/sort.go:231 +0x87
  sort.Float64s()
      /opt/hostedtoolcache/go/1.16.14/x64/src/sort/sort.go:332 +0xa4
  go.k6.io/k6/metrics.(*TrendSink).Calc()
      /home/runner/work/k6/k6/metrics/sink.go:138 +0x57
  go.k6.io/k6/metrics.(*TrendSink).P()
      /home/runner/work/k6/k6/metrics/sink.go:124 +0xc9
  go.k6.io/k6/metrics.TestGetResolversForTrendColumnsCalculation.func1()
      /home/runner/work/k6/k6/metrics/sample_test.go:155 +0x207
  testing.tRunner()
      /opt/hostedtoolcache/go/1.16.14/x64/src/testing/testing.go:1203 +0x202

Goroutine 178 (running) created at:
  testing.(*T).Run()
      /opt/hostedtoolcache/go/1.16.14/x64/src/testing/testing.go:1248 +0x5d7
  go.k6.io/k6/metrics.TestGetResolversForTrendColumnsCalculation()
      /home/runner/work/k6/k6/metrics/sample_test.go:148 +0x336
  testing.tRunner()
      /opt/hostedtoolcache/go/1.16.14/x64/src/testing/testing.go:1203 +0x202

Goroutine 173 (running) created at:
  testing.(*T).Run()
      /opt/hostedtoolcache/go/1.16.14/x64/src/testing/testing.go:1248 +0x5d7
  go.k6.io/k6/metrics.TestGetResolversForTrendColumnsCalculation()
      /home/runner/work/k6/k6/metrics/sample_test.go:148 +0x336
  testing.tRunner()
      /opt/hostedtoolcache/go/1.16.14/x64/src/testing/testing.go:1203 +0x202
==================
--- FAIL: TestGetResolversForTrendColumnsCalculation (0.00s)
    --- FAIL: TestGetResolversForTrendColumnsCalculation/p(99.9) (0.00s)
    --- FAIL: TestGetResolversForTrendColumnsCalculation/p(99.999) (0.00s)
        testing.go:1102: race detected during execution of test
--- FAIL: TestNew (0.00s)
--- FAIL: TestThresholdRun (0.00s)
    --- FAIL: TestThresholdRun/true (0.00s)
FAIL
FAIL	go.k6.io/k6/metrics	0.106s
```

core/engine.go
```diff
@@ -92,42 +93,78 @@ func NewEngine(
...
-		for _, name := range []string{
-			"http_req_duration{expected_response:true}",
-		} {
-			if _, ok := e.thresholds[name]; ok {
-				continue
+func (e *Engine) initSubMetricsAndThresholds() error {
+	for metricName, thresholds := range e.options.Thresholds {
+		metric, err := e.getOrInitPotentialSubmetric(metricName)
+
+		if e.runtimeOptions.NoThresholds.Bool {
```
MStoykov: Shouldn't this skip the whole parsing either way?

core/engine.go
```diff
@@ -436,25 +473,25 @@ func (e *Engine) processSamplesForMetrics(sampleContainers []stats.SampleContain
...
-			if !ok {
-				m = stats.New(sample.Metric.Name, sample.Metric.Type, sample.Metric.Contains)
-				m.Thresholds = e.thresholds[m.Name]
-				m.Submetrics = e.submetrics[m.Name]
+			m := sample.Metric // this should have come from the Registry, no need to look it up
+			if !m.Observed {
+				// But we need to add it here, so we can show data in the
+				// end-of-test summary for this metric
 				e.Metrics[m.Name] = m
+				m.Observed = true
```
MStoykov: I really dislike this new property on the Metric type. I would really prefer if we go back to the map. 

I do think Metric already has too many responsibilities and now adding it knowing it was "observed" seems contra productive. 
we even still add it to the map so ... :shrug: Why not just get it out of it.

na--: I really dislike it as well, I want to completely split apart the current `Metric` into at least two parts:
- The first part, which should be the actual `Metric`, will be just the `Name`, `Type` and Contains` - that should be the only thing that the `Registry` actually controls and cares about
- A second completely separate `struct` should be the `Sink` and `Observed` and probably the threshold and submetric things. That will only be a concern of the `MetricsRegistry`. I don't think they should even be exported... In any case, the mapping from a `*Metric` to that second struct will happen only in the `MetricsRegistry`.

Unfortunately, until we have that separation, I kind of had to put `Observed` in the `Metric` struct, since that's where the `Sink` already was :disappointed: Doing otherwise without even more refactoring would cause us to have a map lookup for every sample of every metric that k6 generates... :disappointed: I intended on splitting this when I work on the time series refactoring.

na--: btw this is what I mean by this TODO: https://github.com/grafana/k6/blob/f06d6f0b673ebc9884995478a4f8d5d7ac03a923/stats/stats.go#L449-L450

core/engine.go
na--: Hmm I don't think so, given that thresholds are basically the only way to currently create sub-metrics for the end-of-test summary, and we can have `--no-thresholds` without also having `--no-summary`. So, if `--no-thresholds` are enabled, I'm following the previously established logic that we shouldn't fail the test for wrongly configured thresholds, but I'm still trying to parse them for sub-metrics, just emit any errors as warnings.

core/engine.go
MStoykov: > e a map lookup for every sample of every metric that k6 generates..

Didn't we have the same up to this point as well?

na--: Yes, we did, but we don't actually need to have that, so I guess you can consider it an optimization? :sweat_smile: : 

I probably dislike the extra item in `Metric` even more than you, but with `Sink` already in there, this actually seems a better (temporary and transient) architecture to me, since all of this state is now kept in `Metric` and will be easy to refactor into a different struct all at the same time. Previously, the `Observed` state was implicitly tracked by that map in the `Engine`, while all of the other state was in every `Metric`...

If you want, I am fine with implementing my `TODO` above now, before we merge this PR, though I'd prefer to do it in a separate PR after this https://github.com/grafana/k6/pull/2442 PR from @oleiade is also merged. It will leave the `metrics.Registry` as the only source of truth for metrics, so we'll be easily able to assign them sequential IDs and use that in the `MetricsEngine` without the need for a map.

metrics/engine/engine.go
```diff
@@ -0,0 +1,174 @@
...
+
+	t := me.executionState.GetCurrentTestRunDuration()
+
+	for _, m := range me.metricsWithThresholds {
+		if len(m.Thresholds.Thresholds) == 0 {
+			continue
+		}
+		m.Tainted = null.BoolFrom(false)
+
+		me.logger.WithField("m", m.Name).Debug("running thresholds")
```
olegbespalov, line 157:
```suggestion
		me.logger.WithField("metric_name", m.Name).Debug("running thresholds")
```
If we don't use `m` before in logs, it's better to avoid starting. The human-friendly names help during the investigations and cost nothing

metrics/engine/engine.go
```diff
@@ -0,0 +1,174 @@
...
+			continue
+		}
+		if !succ {
+			me.logger.WithField("m", m.Name).Debug("Thresholds failed")
+			m.Tainted = null.BoolFrom(true)
+			thresholdsTainted = true
+			if m.Thresholds.Abort {
+				shouldAbort = true
+			}
+		}
```
olegbespalov, line 170:
```suggestion
		if succ {
		   continue
		}
		
		me.logger.WithField("m", m.Name).Debug("Thresholds failed")
		m.Tainted = null.BoolFrom(true)
		thresholdsTainted = true
		if m.Thresholds.Abort {
			shouldAbort = true
		}		
```
:see_no_evil: sorry, I really don't like nested `if`s

metrics/engine/engine.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+
+	return nil
+}
+
+// ProcessThresholds processes all of the thresholds.
+//
+// TODO: refactor, make private, optimize
+func (me *MetricsEngine) ProcessThresholds() (thresholdsTainted, shouldAbort bool) {
```
olegbespalov, line 145: Maybe `EvaluateThresholds` :thinking:  ?

core/engine.go
```diff
@@ -89,89 +84,32 @@ func NewEngine(
...
-		// even if they don't have any metric samples during the test run
-		metric.Observed = true
-		e.Metrics[metric.Name] = metric
-		if metric.Sub != nil {
-			metric.Sub.Metric.Observed = true
-			e.Metrics[metric.Sub.Metric.Name] = metric.Sub.Metric
-		}
+	if !(rtOpts.NoSummary.Bool && rtOpts.NoThresholds.Bool) {
+		e.ingester = me.GetIngester()
+		outputs = append(outputs, e.ingester)
```
codebien, line 102: Is it not possible to have this in the same place where we set the other outputs?

cmd/run.go
```diff
@@ -151,17 +153,20 @@ func (c *cmdRun) run(cmd *cobra.Command, args []string) error {
...
-	err = outputManager.StartOutputs()
+	// TODO: re-enable the code below
+	/*
+		outputManager := output.NewManager(outputs, logger, func(err error) {
+			if err != nil {
+				logger.WithError(err).Error("Received error to stop from output")
+			}
+			runCancel()
+		})
+	*/
```
codebien, line 164: When would you re-enable it? I guess when the engine will be totally removed but not sure, so can you make it clear, please? :pray:

codebien (commented):
Some minor things

core/engine.go
```diff
@@ -57,23 +58,23 @@ type Engine struct {
...
+	options        lib.Options
 	runtimeOptions lib.RuntimeOptions
 	outputs        []output.Output
 
 	logger   *logrus.Entry
 	stopOnce sync.Once
 	stopChan chan struct{}
 
-	Metrics     map[string]*stats.Metric
+	Metrics     map[string]*stats.Metric // TODO: refactor, this doesn't need to be a map
```
oleiade: The To-do could benefit from mentioning what it could be instead 😸

core/engine.go
```diff
@@ -92,42 +93,78 @@ func NewEngine(
...
+			return nil, err
 		}
+	}
+
+	return e, nil
+}
 
-		parent, sm := stats.NewSubmetric(name)
-		e.submetrics[parent] = append(e.submetrics[parent], sm)
+func (e *Engine) getOrInitPotentialSubmetric(name string) (*stats.Metric, error) {
```
oleiade: I would suggest removing the "Potential" part of the name, and explicit the behavior of the function in its documentation instead.

core/engine.go
```diff
@@ -92,42 +93,78 @@ func NewEngine(
...
+
+	metric := e.registry.Get(nameParts[0])
+	if metric == nil {
+		return nil, fmt.Errorf("metric '%s' does not exist in the script", nameParts[0])
+	}
+	if len(nameParts) == 1 { // no sub-metric
+		return metric, nil
+	}
+
+	if nameParts[1][len(nameParts[1])-1] != '}' {
```
oleiade: I suggest binding a `submetricDefinition` (or something along those lines) variable to the `nameParts[1]` value to ease readability.

core/engine.go
```diff
@@ -473,7 +510,7 @@ func (e *Engine) processSamples(sampleContainers []stats.SampleContainer) {
 
 	// TODO: run this and the below code in goroutines?
```
oleiade: Could we do that in this PR as we're at it?

oleiade (dismissed):
LGTM 🎉 

A bunch of minor, non-blocking comments on my side. I do agree with the brought up argument about `metrics.Observed`; I dislike it too, but I understand the arguments in (temporary favor of it, and can live with it for now.

metrics/package.go
```diff
@@ -0,0 +1,10 @@
+// Package metrics contains various k6 components that deal with metrics and
+// thresholds.
+package metrics
+
+// TODO: move most things from the stats/ package here
+
+// TODO: maybe even move the outputs to a sub-folder here? it may be worth it to
+// do a new Output v2 implementation that uses channels and is more usable and
+// easier to write? this way the old extensions can still work for a while, with
+// an adapter and a deprecation notice
```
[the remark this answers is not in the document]
na--: I kind of already also have it the issue though, see the end of https://github.com/grafana/k6/issues/2430#issue-1165215219

core/engine.go
na--: It may be, but I think it will actually be worse to do it that way. 

he [other outputs receive a whole bunch of things](https://github.com/grafana/k6/blob/f06d6f0b673ebc9884995478a4f8d5d7ac03a923/cmd/outputs.go#L80-L130) this one doesn't need, while this one needs the `MetricsEngine`, something we create a few lines above here (or in `cmd/run.go` in the follow up PR without an Engine). Besides, we add this output based on completely different criteria only some of the time, based on RuntimeOptions values...

core/engine.go
na--: It's slightly better in the follow-up PR: https://github.com/grafana/k6/blob/60237cc7c3c15e93df91f4160dd4fd6ac3765a6c/metrics/engine/engine.go#L42-L48

core/engine.go
na--: I'd prefer to do this as a separate PR, hopefully along the lines of https://github.com/grafana/k6/issues/2430

oleiade: Fair enough! 😸

metrics/engine/engine.go
oleiade: I would also rename `Thresholds.Run` to `Thresholds.Evaluate` for consitency then, WDYT?

metrics/engine/engine.go
```diff
@@ -0,0 +1,174 @@
...
+func (me *MetricsEngine) GetIngester() output.Output {
+	return &outputIngester{
+		logger:        me.logger.WithField("component", "metrics-engine-ingester"),
+		metricsEngine: me,
+	}
+}
+
+func (me *MetricsEngine) getOrInitPotentialSubmetric(name string) (*stats.Metric, error) {
+	// TODO: replace with strings.Cut after Go 1.18
+	nameParts := strings.SplitN(name, "{", 2)
```
oleiade, line 76: I ended up doing something somewhat similar to extract metric/submetric name for thresholds validation. I ended up using `strings.FieldFunc`, which, I believe, could lead to somewhat clearer code here?

```go
delimiterFn := func(c rune) bool {
    return c == '{' || c == '}' || c == ':'
}

fmt.Println("Fields are: %q", strings.FieldsFunc("metric_name{foo:bar}", f))
```

The resulting length is either 1 (metric), or an odd value >=3. No need to match the closing `}` for instance. You could also drop the matching of `:` in your case, I believe.

Hope that's helpful 🙇🏻

na--: Wouldn't that also accept `metric_name}foo{bar:` as valid? :confused:

oleiade: That's a fair point. I need to look into it. I'll add a test for it in https://github.com/grafana/k6/pull/2463. Commit https://github.com/grafana/k6/pull/2463/commits/e4e51f0a4207154ba3d42137cc283ddd9c382222 introduces a `ParseMetricName` function.

core/engine.go
na--: resolved by https://github.com/grafana/k6/pull/2426/commits/d7bc04b33223b6c00f944c6d57b82bdba4bef52f

core/engine.go
na--: resolved by https://github.com/grafana/k6/pull/2426/commits/d7bc04b33223b6c00f944c6d57b82bdba4bef52f

core/local/local.go
```diff
@@ -51,6 +51,10 @@ type ExecutionScheduler struct {
 	maxDuration     time.Duration // cached value derived from the execution plan
 	maxPossibleVUs  uint64        // cached value derived from the execution plan
 	state           *lib.ExecutionState
+
+	// TODO: remove these when we don't have separate Init() and Run() methods
+	// and can use a context + a WaitGroup (or something like that)
+	stopVusEmission, vusEmissionStopped chan struct{}
```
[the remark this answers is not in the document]
na--: resolved by https://github.com/grafana/k6/pull/2426/commits/d7bc04b33223b6c00f944c6d57b82bdba4bef52f

metrics/engine/engine.go
na--: resolved by https://github.com/grafana/k6/pull/2426/commits/d7bc04b33223b6c00f944c6d57b82bdba4bef52f

metrics/engine/engine.go
na--: resolved by https://github.com/grafana/k6/pull/2426/commits/d7bc04b33223b6c00f944c6d57b82bdba4bef52f

metrics/engine/engine.go
na--: resolved by https://github.com/grafana/k6/pull/2426/commits/d7bc04b33223b6c00f944c6d57b82bdba4bef52f

oleiade (approved):
🍾

PR #2429, opened 2022-03-09
Expose the Source of the NullHostnameTrie

olegbespalov:
## Why?

This change is exposing the source of the `NullHostnameTrie`, otherwise, the only way of getting the access to the underlying array is to use Marshaling, like:

```go
func extractHostnames(list k6types.NullHostnameTrie) ([]string, error) {
	if !list.Valid {
		return []string{}, nil
	}

	tmp, err := list.MarshalJSON()
	if err != nil {
		return []string{}, err
	}

	var extracted []string
	err = json.Unmarshal(tmp, &extracted)
	if err != nil {
		return []string{}, err
	}

	return extracted, nil
}
```

which isn't efficient

### DIFF
--- a/lib/types/hostnametrie.go
+++ b/lib/types/hostnametrie.go
@@ -70,6 +70,15 @@ func (d *NullHostnameTrie) UnmarshalJSON(data []byte) error {
 	return nil
 }
 
+// Source return source hostnames that were used diring the construction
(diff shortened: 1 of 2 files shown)


MStoykov:
Is there a particular reason we need to?

MStoykov (approved):

codebien (approved):

PR #2436, opened 2022-03-11
Output json optimizations

MStoykov:
Mostly provoked by this [community forum post](https://community.k6.io/t/k6-is-memory-hungry-when-using-several-modules/3000/33). 

Moving to using easyjson gave some improvement, but then I noticed we are spending a bunch of time in compression so decided to try out a different gzip implementation.

Combined improvement is twice better on both a CPU and allocations
```
name            old time/op    new time/op    delta
FlushMetrics-8    37.2ms ± 6%    18.2ms ± 6%  -50.99%  (p=0.000 n=10+9)

name            old alloc/op   new alloc/op   delta
FlushMetrics-8    2.44MB ± 1%    2.52MB ± 0%   +3.28%  (p=0.000 n=10+10)

name            old allocs/op  new allocs/op  delta
FlushMetrics-8     47.3k ± 1%     20.1k ± 0%  -57.59%  (p=0.000 n=10+10)
```

Breaking change: the tag keys are now *not* sorted in the output. This is due to no longer using the stdlib marshalling which does this.


MStoykov:
Some of those can likely be used in the output/cloud as well as they don't seem that complicated to apply

na-- (approved):
LGTM, though please add a PR description with a short description and a benchmark comparison for the cumulative speedup of all commits

MStoykov:
@na-- updated

MStoykov:
Note: if we change the time to be int64 instead of `time.Time` we can get even better performance
Compared to the optimized version here:
```
name            old time/op    new time/op    delta
FlushMetrics-8    18.2ms ± 6%    11.3ms ±12%  -37.96%  (p=0.000 n=9+10)

name            old alloc/op   new alloc/op   delta
FlushMetrics-8    2.52MB ± 0%    1.85MB ± 0%  -26.64%  (p=0.000 n=10+9)

name            old allocs/op  new allocs/op  delta
FlushMetrics-8     20.1k ± 0%      7.6k ± 0%  -62.31%  (p=0.000 n=10+10)
```

Compared to the original:
```
name            old time/op    new time/op    delta
FlushMetrics-8    37.2ms ± 6%    11.3ms ±12%  -69.59%  (p=0.000 n=10+10)

name            old alloc/op   new alloc/op   delta
FlushMetrics-8    2.44MB ± 1%    1.85MB ± 0%  -24.23%  (p=0.000 n=10+9)

name            old allocs/op  new allocs/op  delta
FlushMetrics-8     47.3k ± 1%      7.6k ± 0%  -84.01%  (p=0.000 n=10+10)
```

Unfortunately I can't figure out a not terrible way to support both so we can be backwards compatible :( 

Also, of note is that I remembered there is a slight breaking change - the tag keys are now not sorted in the JSON output.

oleiade (approved):

PR #2416, opened 2022-03-07
Updating Code of Conduct and Github urls

javaducky:

PR #2437, opened 2022-03-11
chore: update CONTRIBUTING.md

olegbespalov:
#What?

This PR is updating the `CONTRIBUTING.md`

Also, WDYT if we add the Makefile's target with the tooling installation?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -23,16 +23,16 @@ Contributing code
 
 If you'd like to contribute code to k6, this is the basic procedure. Make sure to follow the [style guide](#style-guide) described below.
 
-1. Find an issue you'd like to fix. If there is none already, or you'd like to add a feature, please open one and we can talk about how to do it.  Out of respect for your time, please start a discussion regarding any bigger contributions either in a GitHub Issue, in the community forums or in the `#contributors` channel of the k6 slack **before** you get started on the implementation.
+1. Find an issue you'd like to fix. If there is none already, or you'd like to add a feature, please open one, and we can talk about how to do it.  Out of respect for your time, please start a discussion regarding any bigger contributions either in a GitHub Issue, in the community forums or in the `#contributors` channel of the k6 slack **before** you get started on the implementation.
   
    
-   Remember, there's more to software development than code; if it's not properly planned, stuff gets messy real fast. 
+   Remember, there's more to software development than code; if it's not properly planned, stuff gets messy real fast.
 
 2. Create a fork and open a feature branch - `feature/my-cool-feature` is the classic way to name these, but it really doesn't matter.
 
 3. Create a pull request!
 
-4. Sign the [Contributor License Agreement](https://cla-assistant.io/loadimpact/k6) (the process is integrated with the pull request flow through cla-assistant.io)
+4. Sign the [Contributor License Agreement](https://cla-assistant.io/grafana/k6) (the process is integrated with the pull request flow through cla-assistant.io).
 
 5. We will discuss implementation details until everyone is happy, then a maintainer will merge it.
 
@@ -42,14 +42,17 @@ Development setup
 To get a basic development environment for Go and k6 up and running, first make sure you have **[Git](https://git-scm.com/downloads)** and **[Go](https://golang.org/doc/install)** (1.10 or newer) installed and working properly.
 
 Once that's done, you can get the k6 source into your Go workspace (`$GOPATH/src`) by running:
+
 ```bash
 go get go.k6.io/k6
 ```
+
 This will also build a `k6` binary and put it in `$GOPATH/bin`.
 
 **Building from source**:
 
 Standing in the repo root (`$GOPATH/src/go.k6.io/k6`) you can build a k6 binary from source by running:
+
 ```bash
 cd $GOPATH/src/go.k6.io/k6
 go build
@@ -58,15 +61,25 @@ go build
 **Running the linter**:
 
 We make use of the [golangci-lint](https://github.com/golangci/golangci-lint) tool to lint the code in CI. To run it locally, first install it:
+
 ```bash
 go get -u github.com/golangci/golangci-lint/cmd/golangci-lint
 ```
+
 then run:
-```
-golangci-lint run --out-format=tab --new-from-rev master ./...
+
+```bash
+make lint
 ```
 
-If you've added new dependencies you might also want to check and make sure all dependencies exists in `vendor/` folder by running:
+You can also run the linter inside the docker container, which will benefit from the version of the linter being the same as it will be in CI.
+
+```bash
+make ci-like-lint
+```
+
+If you've added new dependencies (and just in case [we have a policy about it](Dependencies.md)) you might also want to check and make sure all dependencies exists in `vendor/` folder by running:
+
 ```bash
 go get -u github.com/FiloSottile/vendorcheck
 vendorcheck ./...
@@ -75,21 +88,25 @@ vendorcheck ./...
 **Running the test suite**:
 
 To exercise the entire test suite:
+
 ```bash
-go test -race ./...
+make test
 ```
 
 To run the tests of a specific package:
+
 ```bash
 go test -race go.k6.io/k6/core
 ```
 
 To run just a specific test case use `-run` and pass in a regex that matches the name of the test:
+
 ```bash
 go test -race ./... -run ^TestEngineRun$
 ```
 
 Combining the two above we can run a specific test case in a specific package:
+
 ```bash
 go test -race go.k6.io/k6/core -run ^TestEngineRun$
 ```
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -19,12 +19,12 @@ ci-like-lint :
 lint :
 	golangci-lint run --out-format=tab --new-from-rev master ./...
 
-tests :
+test :
```
codebien, line 22:
```suggestion
test:
```
Is the space expected? I see we have it on every command

codebien (changes requested):
just one simple request

CONTRIBUTING.md
```diff
@@ -58,15 +61,25 @@ go build
...
+```bash
+make ci-like-lint
 ```
 
-If you've added new dependencies you might also want to check and make sure all dependencies exists in `vendor/` folder by running:
+If you've added new dependencies (and just in case [we have a policy about it](Dependencies.md)) you might also want to check and make sure all dependencies exists in `vendor/` folder by running:
+
 ```bash
 go get -u github.com/FiloSottile/vendorcheck
 vendorcheck ./...
```
MStoykov, line 85: lol this can probably go away ... It has been deprecated by the author and if they will be doing something they can just as well do `go mod vendor` which will even fix it for them.

CONTRIBUTING.md
```diff
@@ -23,16 +23,16 @@ Contributing code
...
    
-   Remember, there's more to software development than code; if it's not properly planned, stuff gets messy real fast. 
+   Remember, there's more to software development than code; if it's not properly planned, stuff gets messy real fast.
 
 2. Create a fork and open a feature branch - `feature/my-cool-feature` is the classic way to name these, but it really doesn't matter.
 
 3. Create a pull request!
 
-4. Sign the [Contributor License Agreement](https://cla-assistant.io/loadimpact/k6) (the process is integrated with the pull request flow through cla-assistant.io)
+4. Sign the [Contributor License Agreement](https://cla-assistant.io/grafana/k6) (the process is integrated with the pull request flow through cla-assistant.io).
```
javaducky, line 35: I thought about changing this the other day with some of the other changes, but not sure this is what we want just yet. They do seem to cover different shared repositories.

CONTRIBUTING.md
```diff
@@ -142,6 +159,6 @@ If you're writing a longer block of text to a terminal, wrap it at 80 characters
 
 **License**
 
-If you make a new source file, you must copy the license preamble from an existing file to the top of it. We can't merge a PR with unlicensed source files. We also can't merge PRs unless all authors have signed the [Contributor License Agreement](https://cla-assistant.io/loadimpact/k6).
+If you make a new source file, you must copy the license preamble from an existing file to the top of it. We can't merge a PR with unlicensed source files. We also can't merge PRs unless all authors have signed the [Contributor License Agreement](https://cla-assistant.io/grafana/k6).
```
javaducky, line 162: Same concern with regard to changing the particular CLA.

CONTRIBUTING.md
olegbespalov: But check [this example](https://github.com/grafana/k6/pull/2416#issuecomment-1061077979), it seems like the new one is what the contributor should sign, no? :thinking:

Makefile
olegbespalov: :+1:

CONTRIBUTING.md
javaducky: It definitely feels like that's what it should be, I guess I just had hesitation initially. 🤷

javaducky (approved):

MStoykov (changes requested):
Can we replace all `slack` mentions with the community forum?

CONTRIBUTING.md
```diff
@@ -42,14 +42,17 @@ Development setup
 To get a basic development environment for Go and k6 up and running, first make sure you have **[Git](https://git-scm.com/downloads)** and **[Go](https://golang.org/doc/install)** (1.10 or newer) installed and working properly.
 
 Once that's done, you can get the k6 source into your Go workspace (`$GOPATH/src`) by running:
+
 ```bash
 go get go.k6.io/k6
```
MStoykov, line 47: let's propose that they use `git clone`

CONTRIBUTING.md
```diff
@@ -42,14 +42,17 @@ Development setup
 To get a basic development environment for Go and k6 up and running, first make sure you have **[Git](https://git-scm.com/downloads)** and **[Go](https://golang.org/doc/install)** (1.10 or newer) installed and working properly.
```
MStoykov, line 42:
```suggestion
To get a basic development environment for Go and k6 up and running, first make sure you have **[Git](https://git-scm.com/downloads)** and **[Go](https://golang.org/doc/install)** (1.17 or newer) installed and working properly.
```

CONTRIBUTING.md
```diff
@@ -42,14 +42,17 @@ Development setup
...
+
 ```bash
 go get go.k6.io/k6
 ```
+
 This will also build a `k6` binary and put it in `$GOPATH/bin`.
 
 **Building from source**:
 
 Standing in the repo root (`$GOPATH/src/go.k6.io/k6`) you can build a k6 binary from source by running:
```
MStoykov, line 54: Yeah ... I think we need to rewrite this whole section

CONTRIBUTING.md
```diff
@@ -58,42 +61,53 @@ go build
...
+
 ```bash
 go test -race go.k6.io/k6/core
 ```
 
 To run just a specific test case use `-run` and pass in a regex that matches the name of the test:
+
 ```bash
 go test -race ./... -run ^TestEngineRun$
 ```
```
MStoykov, line 99: I am not particularly certain we need to explain how to run go tests - if the developer isn't familiar with go to know this, I don't think we are the people to teach them that.

javaducky: While I agree that we shouldn't be expected to _teach_ potential contributors how to run tests in Go, we _should_ however lower any barriers for a contributor to add to our project. Maybe instead, we point contributors to make use of the `Makefile` target as a "things to do before submitting a PR." WDYT?

I plan to work up similar for "supported extensions", so this change can be tabled for now and be revisited.

CONTRIBUTING.md
```diff
@@ -58,42 +61,53 @@ go build
...
+make ci-like-lint
 ```
 
 **Running the test suite**:
 
 To exercise the entire test suite:
+
 ```bash
-go test -race ./...
+make test
```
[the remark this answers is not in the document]
olegbespalov: We had a conversation the other day with @imiric, and he provided me a comment which seems reasonable:

> it's a test process, not a "tests" artifact; same as "lint," instead of "linters"

Does that make sense? :)

CONTRIBUTING.md
javaducky: Disregard...later convo made me realize there is already reference to `make test` in another section. 🙈

javaducky (approved):

CONTRIBUTING.md
```diff
@@ -1,135 +1,107 @@
...
 
-We make use of the [golangci-lint](https://github.com/golangci/golangci-lint) tool to lint the code in CI. To run it locally, first install it:
-```bash
-go get -u github.com/golangci/golangci-lint/cmd/golangci-lint
-```
-then run:
-```
-golangci-lint run --out-format=tab --new-from-rev master ./...
-```
+We make use of the [golangci-lint](https://github.com/golangci/golangci-lint) tool to lint the code in CI. To run it locally, first [install it](https://golangci-lint.run/usage/install/#local-installation), then run:
```
MStoykov, line 50: Given the fact that the version here is pretty important maybe we can specify to get it from the first line of `.golangci.yaml` ?

MStoykov: Maybe even preface it that they can just use the docker version as I would expect most users will read top-to-bottom, while doing the instructions. 

Maybe this can be rewritten to show they can skip that if they just use docker. 

WDYT @javaducky

CONTRIBUTING.md
```diff
@@ -139,9 +111,8 @@ If you're writing a longer block of text to a terminal, wrap it at 80 characters
    Endpoint: /v1/status
 ```
 
+### License
 
-**License**
-
-If you make a new source file, you must copy the license preamble from an existing file to the top of it. We can't merge a PR with unlicensed source files. We also can't merge PRs unless all authors have signed the [Contributor License Agreement](https://cla-assistant.io/loadimpact/k6).
+If you make a new source file, you must copy the license preamble from an existing file to the top of it. We can't merge a PR with unlicensed source files. We also can't merge PRs unless all authors have signed the [Contributor License Agreement](https://cla-assistant.io/grafana/k6).
```
MStoykov, line 116: This is no longer needed so I think we can drop the whole section

MStoykov (dismissed):
LGTM, I have listed some more changes, but am fine with merging it as is and making them later.

CONTRIBUTING.md
olegbespalov: :facepalm: true!

CONTRIBUTING.md
olegbespalov: Yes, that's for sure. Let's reference here the line with the version.

I'm starting to think that it would be nice to have a tiny container where we can keep pre-installed tools with the desired versions and use it for the running things like `golangci-lint run --fix` or re-generate easy son, protobuf and so-on.

What do you think?

MStoykov: I think that most people will not need anything but `golangci-lint` most of the time. This means that if we put more stuff in that image it will just be bigger for no benefit, most of the time. 

I am fine with there being another one for `go generate ./...` calling, but let's do that in a separate PR. Also, this happens *fairly* rare, but I guess it's a good idea to have the exact versions somewhere.

MStoykov (approved):

codebien (approved):

oleiade (approved):

javaducky (approved):

PR #2442, opened 2022-03-14
Remove `stats.New`

oleiade:
This PR removes the `stats.New` method from the `stats` package. This change is intended to facilitate the move of the `stats` package code in a top-level `metrics` code. It was triggered and requested, as preparation step to https://github.com/grafana/k6/pull/2433 and https://github.com/grafana/k6/pull/2426.

Note that this PR might be taking certain shortcut, assuming the API will be readjusted later by the mentioned PRs.

### Question marks

I took the initiative to use the `registry.NewMetric` in places where `stats.New` used to be. I could see two main alternatives: 
* using `registry.MustNewMetric`, but considering we use `require.NoError` everywhere else, I picked `registry.NewMetric` instead
* instantiating `Metric` instances “manually” everywhere it's possible. However, considering we want to explicit that metrics should be created through a registry, I judged it more relevant.

### Dependencies

This should be merged before https://github.com/grafana/k6/pull/2433 and https://github.com/grafana/k6/pull/2426

### DIFF
--- a/metrics/registry.go
+++ b/metrics/registry.go
@@ -61,7 +61,7 @@ func (r *Registry) NewMetric(name string, typ stats.MetricType, t ...stats.Value
 	oldMetric, ok := r.metrics[name]
 
 	if !ok {
-		m := stats.New(name, typ, t...)
+		m := newMetric(name, typ, t...)
 		r.metrics[name] = m
 		return m, nil
 	}
@@ -90,4 +90,30 @@ func (r *Registry) MustNewMetric(name string, typ stats.MetricType, t ...stats.V
 // Get() will return a nil value.
 func (r *Registry) Get(name string) *stats.Metric {
 	return r.metrics[name]
+}
+
+func newMetric(name string, mt stats.MetricType, vt ...stats.ValueType) *stats.Metric {
+	valueType := stats.Default
+	if len(vt) > 0 {
+		valueType = vt[0]
+	}
+	var sink stats.Sink
+	switch mt {
+	case stats.Counter:
+		sink = &stats.CounterSink{}
+	case stats.Gauge:
+		sink = &stats.GaugeSink{}
+	case stats.Trend:
+		sink = &stats.TrendSink{}
+	case stats.Rate:
+		sink = &stats.RateSink{}
+	default:
+		return nil
+	}
+	return &stats.Metric{
+		Name:     name,
+		Type:     mt,
+		Contains: valueType,
+		Sink:     sink,
+	}
 }
--- a/stats/stats.go
+++ b/stats/stats.go
@@ -466,13 +466,15 @@ func (m *Metric) Sample(t time.Time, tags *SampleTags, value float64) Sample {
 	}
 }
 
-func New(name string, typ MetricType, t ...ValueType) *Metric {
-	vt := Default
-	if len(t) > 0 {
-		vt = t[0]
+// newMetric instantiates a new Metric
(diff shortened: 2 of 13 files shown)


lib/metrics/registry.go
```diff
@@ -61,19 +61,39 @@ func (r *Registry) NewMetric(name string, typ stats.MetricType, t ...stats.Value
 	oldMetric, ok := r.metrics[name]
 
 	if !ok {
-		m := stats.New(name, typ, t...)
+		valueType := stats.Default
```
olegbespalov, line 64: It seems like a small method (like it was before). Do you think we can benefit from moving this into the other constructor? :thinking:

oleiade: I ended up renaming `stats.New` to `stats.newMetric`, keeping it private, to make things simpler 👍🏻

oleiade:
Rebased this on #2426. This now depends on said PR.

A notable change made is that we now use the `myMetric.AddSubmetric` API to handle tagged metrics. This was made possible by #2426 changes.

The failing tests are inherited from #2426 and the PR should be green once those issues are addressed in said PR.

metrics/registry.go
```diff
@@ -91,3 +91,29 @@ func (r *Registry) MustNewMetric(name string, typ stats.MetricType, t ...stats.V
 func (r *Registry) Get(name string) *stats.Metric {
 	return r.metrics[name]
 }
+
+func newMetric(name string, mt stats.MetricType, vt ...stats.ValueType) *stats.Metric {
```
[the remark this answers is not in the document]
oleiade: Hey @olegbespalov,

I should have given the whole rationale, indeed. It is indeed the same as `stats.New`, just renamed and made private. There are two main reasons I decided to take that route:
- Creating a metric involves selecting and instantiating a `Sink` (interface type). To create the proper concrete `Sink` type for a metric, we need to make a type switch on the metric type. Keeping the constructor as a helper, helps to obfuscate that from the end-user, and avoids creating a helper that would return an interface type (`Sink`).
- It avoids expanding this construction logic in business logic code. Note that this constructor will be confined to the `stats` package, and later to the `metrics` package, and is really a helper.

Please let me know if you have any concerns. I'm also open to improvement propositions 😸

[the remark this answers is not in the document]
olegbespalov: Hey @oleiade !

My primary rationale for pointing out was to keep the single place for the logic. I see a risk of somebody changing one place but forgetting to change the other as a higher than the bad practice of exposing the interface that is an internal one.

But (TM)

> Note that this constructor will be confined to the stats package, and later to the metrics package, and is really a helper.

:arrow_up: That's probably all that I wanted to hear :)

olegbespalov (dismissed):

stats/stats.go
```diff
@@ -466,13 +466,15 @@ func (m *Metric) Sample(t time.Time, tags *SampleTags, value float64) Sample {
 	}
 }
 
-func New(name string, typ MetricType, t ...ValueType) *Metric {
-	vt := Default
-	if len(t) > 0 {
-		vt = t[0]
+// newMetric instantiates a new Metric
+func newMetric(name string, mt MetricType, vt ...ValueType) *Metric {
```
na--, line 470: I am guessing this will get deleted when things are moved to `metrics/`?

na-- (approved):

olegbespalov (approved):

PR #2443, opened 2022-03-14
Update `golang.org/x/*` dependencies to the latest

olegbespalov:
Update the `golang/x` dependencies to the latest

Main issue: #2445


codebien (approved):

MStoykov:
Can you please redo the PR with running `go mod tidy` before `go mod vendor` on each commit instead of just once in the end.

 In this way each commit will be self-sufficient.

MStoykov:
Why is `x/crypto` and `x/net` updated twice ?

olegbespalov:
@MStoykov It was before we decided to go with updates from the `master,` decided to keep to commits to have a chance to discard the latest (from master) if the decision turn to opposite

codebien (approved):

MStoykov (approved):

Issue #2445, opened 2022-03-15
Dependecies update after the `v0.37.0` release


olegbespalov:
Based on the https://github.com/grafana/k6/blob/master/Dependencies.md

This issue collects the list of the dependencies that will be updated after the release:

```
modtools check --direct-only
Some dependencies are out-of-date. Please upgrade by running 'modtools update' or the following commands:

(no new version since the last release) go get github.com/PuerkitoBio/goquery@v1.8.0
(no new version since the last release) go get github.com/andybalholm/brotli@v1.0.4
(+) go get github.com/dop251/goja@v0.0.0-20220214123719-b09a6bfa842f
(no new version since the last release) go get github.com/fatih/color@v1.13.0
(Maybe update?) go get github.com/gorilla/websocket@v1.5.0
(Maybe update? Many fixes) go get github.com/jhump/protoreflect@v1.12.0
(Maybe update?) go get github.com/klauspost/compress@v1.15.1
(no new version since the last release) go get github.com/mattn/go-colorable@v0.1.12
(no new version since the last release) go get github.com/mattn/go-isatty@v0.0.14
(+) go get github.com/spf13/cobra@v1.4.0
(no new version since the last release) go get github.com/tidwall/gjson@v1.14.0
(+) go get golang.org/x/crypto@v0.0.0-20220315160706-3147a52a75dd
(+) go get golang.org/x/net@v0.0.0-20220225172249-27dd8689420f
(+) go get golang.org/x/term@v0.0.0-20210927222741-03fcf44c2211
(+) go get golang.org/x/time@v0.0.0-20220224211638-0e9765cccd65
(Maybe update?) go get google.golang.org/grpc@v1.45.0
```

- [x] Goja: https://github.com/grafana/k6/pull/2441
- [x] `golang/x/*`: https://github.com/grafana/k6/pull/2443
- [x] `github.com/spf13/cobra`: https://github.com/grafana/k6/pull/2446
- [x] Other dependencies part 1 https://github.com/grafana/k6/pull/2449
- [x] Other dependencies part 2 https://github.com/grafana/k6/pull/2452

MStoykov:
@olegbespalov you can definitely update all the ones that I did skip the previous few times. Some of those are now 1year+ without updates so as long as they aren't huge it's probably good idea to just update them. 

This also means that we won't be reminded that we haven't and have to go look up if there is anything we want from them.

PR #2452, opened 2022-03-17
Other dependencies p.2

olegbespalov:
From what I checked the changes have nothing really interesting for us, but on the other side there are bugfixes, and at some point, we probably might want to use the future versions, so it's good to keep diff smaller.

* Maintenance update of fatih/color to `v1.13.0` from `v1.12.0`
* Maintenance update of klauspost/compress to `v1.15.1` from `v1.13.6`
* Maintenance update of mattn/go-colorable to `v0.1.12` from `v0.1.9`

Main issue: #2445


MStoykov (approved):

Dependencies.md
```diff
@@ -13,10 +13,10 @@ This is mostly so that we keep up with dependencies for which we want new featur
 For example some dependencies that were in the middle of big updates were skipped as they didn't add anything of value at the time, but were being updated day after day seemingly fixing bugs added the previous day. As in that case that wasn't a dependency we wanted to be on top off - we skipped it.
 
 Through the rest of the development cycle dependencies shouldn't be updated unless:
+
 1. required for development
-2. a particular bug fix/feature is really important
-	
-	
+2. a particular bug fix/feature is really important.
+
```
MStoykov, line 19: It's really preferable to not have any other but the necessary changes when a dependency is updated.

Did you make this change by accident? If so - I am fine with moving it to another commit in this PR.

olegbespalov: yes, `license/cla` stuck again (and it seems like I'm missing the access to re-run the webhook). I can forse push again with removing this change :see_no_evil:

codebien: I still see changes on `Dependencies.md`

MStoykov: Yeah, you are currently removing a line, @olegbespalov . Not a big problem, but still preferable if we don't do it in the commit updating a dependancy

olegbespalov: I've reverted it, using checkout now, should be no changes

MStoykov (approved):

codebien (approved):

PR #2453, opened 2022-03-22
Treat go panics as an interrupt error

### LINKED ISSUE #2354: panics in k6 extension only skip an iteration instead of failing the test or iteration
### Brief summary

Currently when a k6 extension panics, the k6 runtime "catches" the panic, simply skips the iteration and logs an error message like: 
```
"time="2022-01-26T14:22:12-05:00" level=error msg="a panic occurred in VU code but was caught: record on line 1373: wrong number of fields" executor=constant-vus scenario=default"
```
There is no reasonable way to programmatically catch or report extension panics in k6 and apart from the log message they are essentially like a "silent" error.
For example if a load test has 10 iterations and all of them panic, k6 still considers the load test as a whole as successful.

### k6 version

v0.35.0

### OS

cross platform

_No response_

create an extension method that instantly panics and call this method from a JS script

### Expected behaviour

any (combination?) of these:
- A panic bubbles up to the k6 JS layer as JS error and can be handled or ignored there
- A panic stops the iteration but makes the iteration count as an "error" somewhere
- A panic fails the k6 process with a non-zero exit code immediately

### Actual behaviour

the k6 runtime "catches" the panic, simply skips the iteration and logs an error message like "time="2022-01-26T14:22:12-05:00" level=error msg="a panic occurred in VU code but was caught: record on line 1373: wrong number of fields" executor=constant-vus scenario=default". Otherwise it continues "as if" the iteration or loadtest was successful.

olegbespalov:
# What?

We change the behavior and the go `panic`'s will be treated as the run interruption

# Why?

Otherwise it is hard to detect panics in the code

Closes #2354

### DIFF
--- a/cmd/panic_integration_test.go
+++ b/cmd/panic_integration_test.go
@@ -1,0 +1,98 @@
+package cmd
+
+// TODO: convert this into the integration tests, once https://github.com/grafana/k6/issues/2459 will be done
+
+import (
+	"path/filepath"
+	"testing"
+
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/afero"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"go.k6.io/k6/errext/exitcodes"
+	"go.k6.io/k6/js/modules"
+	"go.k6.io/k6/lib/testutils"
+)
+
+// alarmist is a mock module that do a panic
+type alarmist struct {
+	vu modules.VU
+}
+
+var _ modules.Module = &alarmist{}
+
+func (a *alarmist) NewModuleInstance(vu modules.VU) modules.Instance {
+	return &alarmist{
+		vu: vu,
+	}
+}
+
+func (a *alarmist) Exports() modules.Exports {
+	return modules.Exports{
+		Named: map[string]interface{}{
+			"panic": a.panic,
+		},
+	}
+}
+
+func (a *alarmist) panic(s string) {
+	panic(s)
+}
+
+func init() {
+	modules.Register("k6/x/alarmist", new(alarmist))
+}
+
+func TestRunScriptPanicsErrorsAndAbort(t *testing.T) {
+	t.Parallel()
+
+	testCases := []struct {
+		caseName, testScript, expectedLogMessage string
+	}{
+		{
+			caseName: "panic in the VU context",
+			testScript: `
+			import { panic } from 'k6/x/alarmist';
+
+			export default function() {
+				panic('hey')
+			}
+			`,
+			expectedLogMessage: "a panic occurred in VU code: hey",
+		},
+		{
+			caseName: "panic in the init context",
+			testScript: `
+			import { panic } from 'k6/x/alarmist';
+
+			panic('hey')
+			export default function() {
+				console.log('lorem ipsum');
+			}
+			`,
+			expectedLogMessage: "a panic occurred in the init context code: hey",
+		},
+	}
+
+	for _, tc := range testCases {
+		tc := tc
+		name := tc.caseName
+
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+
+			testFilename := "script.js"
+			testState := newGlobalTestState(t)
+			require.NoError(t, afero.WriteFile(testState.fs, filepath.Join(testState.cwd, testFilename), []byte(tc.testScript), 0o644))
+			testState.args = []string{"k6", "run", testFilename}
+
+			testState.expectedExitCode = int(exitcodes.ScriptAborted)
+			newRootCommand(testState.globalState).execute()
+
+			logs := testState.loggerHook.Drain()
+
+			assert.True(t, testutils.LogContains(logs, logrus.ErrorLevel, tc.expectedLogMessage))
+		})
+	}
+}
--- a/cmd/panic_integration_test.go
+++ b/cmd/panic_integration_test.go
@@ -56,7 +56,8 @@ func TestRunScriptPanicsErrorsAndAbort(t *testing.T) {
 			import { panic } from 'k6/x/alarmist';
 
 			export default function() {
-				panic('hey')
+				panic('hey');
+				console.log('lorem ipsum');
 			}
 			`,
 			expectedLogMessage: "a panic occurred in VU code: hey",
@@ -66,7 +67,7 @@ func TestRunScriptPanicsErrorsAndAbort(t *testing.T) {
 			testScript: `
 			import { panic } from 'k6/x/alarmist';
 
-			panic('hey')
+			panic('hey');
 			export default function() {
 				console.log('lorem ipsum');
 			}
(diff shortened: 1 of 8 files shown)


MStoykov (approved):
LGTM! But can you try to make an "integration" test that will test that this will abort a whole test :thinking: ?

na-- (changes requested):
LGTM as well, but also agree with @MStoykov's desire for an integration test. With the recent `cmd/` changes, it should be relatively easy to create an integration test that injects a mock JS extension that just calls `panic()` and then start a test that imports it and calls it. 

Actually, it will probably be even better to have 2 integration tests, one that panics in the init context and one that panics in the VU context.

olegbespalov:
@MStoykov @na-- is c5419e53 implementing your request, or it should be something different? :thinking:

olegbespalov:
@MStoykov @na-- is c5419e53 implementing your request, or it should be something different? :thinking:

cmd/run_test.go
```diff
@@ -212,3 +213,83 @@ func TestRunScriptErrorsAndAbort(t *testing.T) {
...
+}
+
+func (a *alarmist) panic(s string) {
+	panic(s)
+}
+
+func TestRunScriptPanicsErrorsAndAbort(t *testing.T) {
+	t.Parallel()
+
+	modules.Register("k6/x/alarmist", new(alarmist))
```
olegbespalov, line 245: :question: to a reviwers, is there a way to do `Unregister`, should we implement something like that? :thinking: 

e.g. use case:
```golang
defer func (){
  modules.Register("k6/x/alarmist")
}()
```

olegbespalov, line 245: :question: to a reviwers, is there a way to do `Unregister`, should we implement something like that? :thinking: 

e.g. use case:
```golang
defer func (){
  modules.Register("k6/x/alarmist")
}()
```

MStoykov: The only use case is tests which make this less interesting. In #2399 I am circumventing this by having a test in a package. Or at least have the tests for a given module in a separate package. So that the module is only registered for those tests. 
(FYI package tests are compiled separately from each other so that is the only place that a `_test.go` file will be executed)

There was some [discussion](https://github.com/grafana/k6/issues/1858#issuecomment-787818358) long time ago about having non global registries and so on, but we ended up with the current situation. 

I guess we can work on the `js/runner` to maybe be able to give it a list of modules :shrug: but that likely will look even worse.

so :shrug:

olegbespalov: >(FYI package tests are compiled separately from each other so that is the only place that a _test.go file will be executed)

That's what I know, but even inside the package, it's changing of the global state, which I'd like to clean up after, but for sure, it's not a big deal.

But in general, is your suggestion to move this into a separate package? :thinking:

MStoykov: > That's what I know, but even inside the package, it's changing of the global state, which I'd like to clean up after, but for sure, it's not a big deal. 

Yeah and my whole idea in that package is that this tests will be alone in a package - so will not affect anything. And then at the end the process testing just ends, so there is nothing to cleanup :tada: .

> But in general, is your suggestion to move this into a separate package? thinking 

Yes and more *importanly* have the registering in a `init` function so it's called only once. This way in case somebody (I) runs the test with `-count 2` they will not get a failure on the second call to `Register` as it will be called only once ;)

na--: Another way instead of `init()` would be to write a small helper that first calls `modules.GetJSModules()` and only calls `modules.Register()` if the module wasn't previously registered. Not sure which is simpler :man_shrugging:

cmd/panic_integration_test.go
```diff
@@ -0,0 +1,98 @@
...
+			testState := newGlobalTestState(t)
+			require.NoError(t, afero.WriteFile(testState.fs, filepath.Join(testState.cwd, testFilename), []byte(tc.testScript), 0o644))
+			testState.args = []string{"k6", "run", testFilename}
+
+			testState.expectedExitCode = int(exitcodes.ScriptAborted)
+			newRootCommand(testState.globalState).execute()
+
+			logs := testState.loggerHook.Drain()
+
+			assert.True(t, testutils.LogContains(logs, logrus.ErrorLevel, tc.expectedLogMessage))
```
na--, line 95: Just to be extra thorough, can you please add `console.log('lorem ipsum')` to the first test as well, after the panic, and also verify here that the logs does not contain that string. Right now the test checks the exit code and the expected log message, but it doesn't actually verify that the k6 execution doesn't somehow continue.

na-- (approved):
LGTM except that minor nitpick

cmd/panic_integration_test.go
olegbespalov: :+1: addressed in 72e14c2

js/bundle.go
```diff
@@ -298,7 +299,24 @@ func (b *Bundle) Instantiate(
...
+			// the logger we use see
+			// https://github.com/sirupsen/logrus/issues/1028
+			// https://github.com/sirupsen/logrus/issues/993
+			b := new(bytes.Buffer)
+			for _, s := range gojaStack {
+				s.Write(b)
+			}
+			logger.Error("panic: ", r, "\n", string(debug.Stack()), "\nGoja stack:\n", b.String())
+		}
+	}()
```
MStoykov, line 318: nit: Can we put this *just* before the `eventLoop.start` as otherwise this will catch any panic up until that point as well

js/modules/k6/execution/execution_test.go
```diff
@@ -186,7 +186,7 @@ func TestVUTags(t *testing.T) {
 	})
 }
 
-func TestAbortTest(t *testing.T) { //nolint: tparallel
+func TestAbortTest(t *testing.T) { //nolint:tparallel
```
MStoykov, line 189: nit: I kind of prefer to not have unrelated changes in unrelated files like that ;)

js/bundle.go
MStoykov: Maybe also movign this to a helper function will be ... helpful :thinking:

MStoykov (dismissed):
LGTM, I have some nitpicks but we can do them later I guess

cmd/run_test.go
olegbespalov: I moved the module registration to the `init` since it's hard to move the test to the separated package without  https://github.com/grafana/k6/issues/2459. I'm suggesting to merge this as is, and I'll create a separated PR with moving in the next cycle

js/modules/k6/execution/execution_test.go
olegbespalov: It sounds like a personal preference :thinking: Let me try to provide my two cents on why I'm doing this and continue doing it.

It's a common/best practice to leave the place cleaner than before. For example, this is a comment for a machine (linter) comment, and it doesn't require space in between. The same things I usually do when spotting any grammar issues. Such changes itself is super-minor and don't deserve a separated PR.

Does that make sense? Will it help if I'll do this in a separate commit?

js/bundle.go
olegbespalov: Good point, a consolidation sounds about right :+1:

olegbespalov:
@na-- @MStoykov sorry guys, please have a look again :bow:

na-- (approved):
:heart:

MStoykov (approved):

PR #2455, opened 2022-03-23
Changes error messages. "should" --> "must", "can"

MattDodsonEnglish:
In a few cases, the CLI application stops running and returns an error, stating that some parameter "should" or "shouldn't" have a certain value.

The word "should," though, generally indicates _strong advice_. In these cases, the application stops running, so it makes more sense to indicate that that the error is stating _a requirement_. For requirements or impossibilities, the helping verbs "must" or "can't" would be more appropriate.

Believe it or not, there's already an RFC that covers almost exactly this topic.
https://datatracker.ietf.org/doc/html/rfc2119#section-4

### DIFF
--- a/lib/netext/httpext/response_type_gen.go
+++ b/lib/netext/httpext/response_type_gen.go
@@ -60,7 +60,7 @@ func (i ResponseType) MarshalJSON() ([]byte, error) {
 func (i *ResponseType) UnmarshalJSON(data []byte) error {
 	var s string
 	if err := json.Unmarshal(data, &s); err != nil {
-		return fmt.Errorf("ResponseType must be a string, got %s", data)
+		return fmt.Errorf("ResponseType should be a string, got %s", data)
 	}
 
 	var err error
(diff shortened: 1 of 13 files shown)


oleiade (dismissed):
👍🏻 📖

lib/netext/httpext/response_type_gen.go
```diff
@@ -60,7 +60,7 @@ func (i ResponseType) MarshalJSON() ([]byte, error) {
 func (i *ResponseType) UnmarshalJSON(data []byte) error {
 	var s string
 	if err := json.Unmarshal(data, &s); err != nil {
-		return fmt.Errorf("ResponseType should be a string, got %s", data)
```
MStoykov, line 63: This is generated so you shouldn't edit it manually

MStoykov (changes requested):
Seems okay, thanks :bow: ! but there is one case where you edit a generated file that needs to be reverted :(. I guess we can try to make a PR upstream to change the message but that will take a while

lib/netext/httpext/response_type_gen.go
MattDodsonEnglish: Thanks, I reverted the change.

MStoykov (approved):

oleiade (approved):

PR #2456, opened 2022-03-24
Fix panic due to error in HasWork

MStoykov:
The panic was happening in the `Init` but it was due to the fact that
HasWork wrongfully returned True when it should've returned false.


MStoykov:
While this is fixing the particular case that we know of, there are other places where `Scale` Is used. 

I have trouble remembering why we decided to leave both given that segment.Scale has known problems with splitting work :thinking:

MStoykov:
I can't find any discussion about the topic :( but here is a [PR](https://github.com/grafana/k6/pull/2457) (based on this one) that removes the Segment.Scale and not test fail so :shrug:  Maybe it's good idea to have only one implementation ...

na-- (approved):
LGTM :crossed_fingers:

olegbespalov (approved):
LGTM

PR #2463, opened 2022-03-25
Validate thresholds at init

oleiade:
## What this PR does

We merged Go-based thresholds parsing in `v0.37.0`. This Pull Request builds upon that and implements thresholds validation at init. As of this PR, unless the `--no-thresholds` flag is passed, thresholds will be checked before the execution of k6 commands. Specifically, if a threshold applies to a non-existing metric, or if it uses an aggregation method that's not supported by the former, k6 will immediately produce an error; the execution won't even start.

This addresses #2330 to the extent of the internally agreed upon scope.


metrics/metric.go
```diff
@@ -121,3 +123,61 @@ func (m *Metric) AddSubmetric(keyValues string) (*Submetric, error) {
...
+// return value indicates whether or not the submetric was found.
+func (m *Metric) FindSubmetric(keyValues string) (int, bool) {
+	for idx, sm := range m.Submetrics {
+		if sm.Name == m.Name+"{"+keyValues+"}" {
+			return idx, true
+		}
+	}
+
+	return -1, false
+}
```
MStoykov, line 144: This seems to not be used anywhere?

cmd/testdata/thresholds/non_existing_submetric.js
```diff
@@ -0,0 +1,13 @@
+export const options = {
+	thresholds: {
+		// http_reqs being a builtin metric has no foo:bar
+		// tag definition. As such k6 should detect it and
+		"http_reqs{foo:bar}": ["count>0"],
```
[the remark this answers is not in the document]
MStoykov: Actually it seems like this test file isn't called :thinking:

metrics/metric.go
oleiade: Indeed. It's an artifact of a previous implementation. Cleaning that up 🧹

cmd/testdata/thresholds/non_existing_submetric.js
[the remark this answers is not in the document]
oleiade: Indeed! I realized that in the process of fixing the previous test failures. As we don't have static analysis, we can't know before hand which tags where defined, thus submetrics are only fully available at the end of execution.

I forgot to remove this test too indeed it seems! Cleaning that up 👍🏻

MStoykov:
The PR descriptions still talks about submetric checking

> Specifically, if a threshold applies to a non-existing metric, **or submetric (including groups)**, or if it uses an aggregation method that's not supported by the former, k6 will immediately produce an error; the execution won't even start.

cmd/archive_test.go
```diff
@@ -32,6 +32,36 @@ func TestArchiveThresholds(t *testing.T) {
...
+			noThresholds: true,
+			testFilename: "testdata/thresholds/non_existing_metric.js",
+			wantErr:      false,
+		},
+		{
+			name:         "run should succeed on a threshold applied to a non existing submetric with the --no-thresholds flag set",
+			noThresholds: true,
+			testFilename: "testdata/thresholds/non_existing_metric.js",
+			wantErr:      false,
+		},
```
MStoykov, line 52:
```suggestion
```

This is just repeat of the previous but talks about submetrics

cmd/run_test.go
```diff
@@ -167,6 +167,22 @@ func TestRunScriptErrorsAndAbort(t *testing.T) {
...
+		{
+			testFilename: "thresholds/non_existing_metric.js",
+			name:         "run should succeed on a threshold applied to a non existing metric with the --no-thresholds flag set",
+			extraArgs:    []string{"--no-thresholds"},
+		},
+		{
+			testFilename: "thresholds/non_existing_metric.js",
+			name:         "run should succeed on a threshold applied to a non existing submetric with the --no-thresholds flag set",
+			extraArgs:    []string{"--no-thresholds"},
+		},
```
MStoykov, line 185:
```suggestion
```

oleiade:
@MStoykov Good catch 👍🏻 Edited out 🙇🏻

cmd/testdata/thresholds/non_existing_metric.js
```diff
@@ -0,0 +1,13 @@
+export const options = {
+	thresholds: {
+		// non existing is neither registered, nor a builtin metric.
+		// k6 should catch that.
+		"non existing": ["rate>0"],
```
MStoykov, line 5:
```suggestion
		"non_existing": ["rate>0"],
```
nit: I kind of prefer if this doesn't have a space in it - IIRC we don't enforce any character set on metrics, but we might want in the future and spaces might not be in that set then.

metrics/metric_test.go
```diff
@@ -66,3 +66,104 @@ func TestAddSubmetric(t *testing.T) {
...
+			wantTags:             nil,
+			wantErr:              true,
+		},
+		{
+			name:                 "metric name with mixed valid and invalid tag definitions",
+			metricNameExpression: "test_metric{abc:123,easyas:}",
+			wantMetricName:       "",
+			wantTags:             nil,
+			wantErr:              true,
+		},
```
MStoykov, line 136: Can you add a bunch more cases with even more misformed metrics such as `test_metric{` and such

https://github.com/grafana/k6/pull/2426#discussion_r834502077 has some other examples.

Also, you can skip setting `wantMetricName`, `withTags`, `wantErr`, when it's their "nil" values - it will save some space and I don't think it makes it any less readable - I would argue it does the opposite

oleiade:
For your information: I've updated the code of the `ParseMetricName` function to cater for the cases where tags definition curly braces are either unmatched, or in the wrong order 👍🏻

metrics/thresholds.go
```diff
@@ -244,6 +248,58 @@ func (ts *Thresholds) Parse() error {
...
+	metric := r.Get(parsedMetricName)
+	if metric == nil {
+		err := fmt.Errorf("%w defined on %s; reason: no metric name %q found", ErrInvalidThreshold, metricName, metricName)
+		return errext.WithExitCodeIfNone(err, exitcodes.InvalidConfig)
+	}
+
+	for _, threshold := range ts.Thresholds {
+		// Return a digestable error if we attempt to validate a threshold
+		// that hasn't been parsed yet.
+		if threshold.parsed == nil {
```
MStoykov, line 279: Is there particular reason why we shouldn't parse it here?

MStoykov:
As a general nitpick: I think you are adding why too many `assert`/`require` custom messages which quickly add up to the size of each test, but ... I would argue most of them (definitely not all) have very little to add to the readability of the test. And the rare cases the test fail I would argue will require someone to go read the code either way. 

And in that later case, each assert/require not being on 4 lines will be more beneficial IMO.

metrics/metric_test.go
MStoykov: I think you've addressed this while I went through the rest of the code

metrics/thresholds.go
oleiade: You mean that we should parse it if we find it to be nil? I must admit that I didn't think of doing that, but that's a good idea, will do so 👍🏻

MStoykov: There might be some reason not to do it here- I haven't looked at how this gets called. But if we can that might make using this function easier

metrics/metric_test.go
oleiade: I've removed the `wantMetric`, `withTags` and `wantErr` where relevant.
I've also added the test case for https://github.com/grafana/k6/pull/2426#discussion_r834502077.

metrics/thresholds.go
oleiade: I've given some thought to it, and I believe it's okay to do it. `Parse` and `Validate` are two different thresholds, but if one calls `Validate` with a threshold unparsed, it will be a good sentinel to try to parse it anyway. The updated code also makes sure to return the resulting parse error 👍🏻

metrics/thresholds_parser.go
```diff
@@ -49,6 +49,28 @@ type thresholdExpression struct {
...
+// sink the threshold expression should be applied to.
+//
+// Because a threshold expression's aggregation method can either be
+// a static keyword ("count", "rate", etc...), or a parametric
+// expression ("p(somefloatingpointvalue)"), we need to handle this
+// case specifically. If we encounter the percentile aggregation method token,
+// we recompute the whole "p(value)" expression in order to look for it in the
+// sinks.
+func (te *thresholdExpression) SinkKey() string {
+	//
```
olegbespalov, line 65: :eyes:

metrics/metric.go
```diff
@@ -121,3 +124,62 @@ func (m *Metric) AddSubmetric(keyValues string) (*Submetric, error) {
...
+
+	parserFn := func(c rune) bool {
+		return c == '{' || c == '}'
+	}
+
+	// Split the metric_name{tag_key:tag_value,...} expression
+	// into two "metric_name" and "tag_key:tag_value,..." strings.
+	parts := strings.FieldsFunc(name, parserFn)
+	if len(parts) == 0 || len(parts) > 2 {
+		return "", nil, fmt.Errorf("%w", ErrMetricNameParsing)
```
olegbespalov, line 167: It seems like we don't need error wrapping here :man_shrugging:

olegbespalov (dismissed):
Looks good :+1:

metrics/thresholds_test.go
```diff
@@ -334,6 +337,219 @@ func TestThresholdsParse(t *testing.T) {
 	})
 }
 
+func TestThresholdsValidate(t *testing.T) {
+	t.Parallel()
+
+	t.Run("validating thresholds applied to a non-existing metric fails", func(t *testing.T) {
```
codebien, line 343:
```suggestion
	t.Run("FailNotExistingMetric", func(t *testing.T) {
```
All the other words seem redundant to me, IMHO

Also, this should generate something like the following line that doesn't seem consistent 
`TestThresholdsValidate/validating_thresholds_applied_to_a_non-existing_metric_fails`

metrics/thresholds_test.go
```diff
@@ -334,6 +337,219 @@ func TestThresholdsParse(t *testing.T) {
...
+		parseErr := ts.Parse()
+		require.NoError(t, parseErr)
+
+		var wantErr errext.HasExitCode
+
+		// Collect the result of the parsing operation
+		gotErr := ts.Validate("non-existing", testRegistry)
+
+		assert.Error(t, gotErr, "Validate should fail when applied to a non-existing metric")
+		assert.ErrorIs(t, gotErr, ErrInvalidThreshold, "Validate error chain should contain an ErrInvalidThreshold error")
```
codebien, line 365:
```suggestion
		assert.ErrorIs(t, gotErr, ErrInvalidThreshold)
```
This form seems already clear to me

codebien (changes requested):
Thanks for the great comments, and for making errors with high dev experience for k6 users. :pray: 

I left some comments, some of them could be applied in multiple places but I didn't spot all of them because I expect some could require some discussions before.

metrics/metric_type.go
```diff
@@ -84,3 +84,39 @@ func (t MetricType) String() string {
...
+			tokenAvg,
+			tokenMin,
+			tokenMax,
+			tokenMed,
+			tokenPercentile,
+		}
+	default:
+		// unreachable!
+		return nil
+	}
```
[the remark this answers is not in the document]
oleiade: Good idea! Rust even has a macro for that actually `unreachable!()` 🎉

metrics/thresholds_parser.go
```diff
@@ -49,6 +49,28 @@ type thresholdExpression struct {
...
+	sinkKey := te.AggregationMethod
+	if te.AggregationMethod == tokenPercentile {
+		sinkKey = fmt.Sprintf("%s(%g)", tokenPercentile, te.AggregationValue.Float64)
+	}
+
+	return sinkKey
```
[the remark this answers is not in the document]
oleiade: Fair enough 🙇🏻

metrics/thresholds_test.go
oleiade: I removed the dash that created the inconsistency, but I prefer to keep the name as explicit as possible (especially for my future self) 👍🏻

metrics/thresholds.go
```diff
@@ -244,6 +248,63 @@ func (ts *Thresholds) Parse() error {
...
+			thresholdExpression, err := parseThresholdExpression(threshold.Source)
+			if err != nil {
+				return fmt.Errorf("unable to validate threshold %q on metric %s; reason: "+
+					"parsing threshold failed %w", threshold.Source, metricName, err)
+			}
+
+			threshold.parsed = thresholdExpression
+		}
```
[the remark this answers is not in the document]
oleiade: I would prefer to keep it as is right now. I have ideas to improve this in a future, separate PR during the next cooldown period 👍🏻 Resolving for now, but you are welcome to unresolve it if you have a strong opinion on this 😸

metrics/metric_test.go
```diff
@@ -66,3 +66,117 @@ func TestAddSubmetric(t *testing.T) {
 		})
 	}
 }
+
+func TestParseMetricName(t *testing.T) {
+	// TODO: verify we get an ErrMetricNameParsing
```
codebien, line 71: This TODO seems already addressed, is it?

codebien (approved):
LGTM :clap:  Just one minor thing

olegbespalov (dismissed):

MStoykov (dismissed):
LGTM :tada: :clap:

olegbespalov (approved):

MStoykov (approved):

codebien (approved):

PR #2465, opened 2022-03-29
gh: dev release workflow

codebien:
Added a workflow for testing the new windows certificate, it's a copy without the uploading phases of the current release process, it builds the request branch, creates the platform binaries and it signs the Windows version.

### DIFF
--- a/.github/workflows/dev-release.yml
+++ b/.github/workflows/dev-release.yml
@@ -57,8 +57,8 @@ jobs:
           goversioninfo -64 \
             -platform-specific=true \
             -charset="1200" \
-            -company="Load Impact AB" \
-            -copyright="© Load Impact AB. Licensed under AGPL." \
+            -company="Grafana Labs AB" \
+            -copyright="© Grafana Labs AB. Licensed under AGPL." \
             -description="A modern load testing tool, using Go and JavaScript" \
             -icon=packaging/k6.ico \
             -internal-name="k6" \
--- a/.github/workflows/dev-release.yml
+++ b/.github/workflows/dev-release.yml
@@ -1,0 +1,144 @@
+name: Dev Build
+on:
+  workflow_dispatch:
+    inputs:
+      k6_version:
+        description: 'The version of the release, it must use the semantic versioning format with the v prefix. It is a development release so it is suggested to append a build metadata (e.g. v0.38.0-dev).'
+        required: true
+      k6_branch_or_commit:
+        description: 'k6 branch, tag or commit'
+        default: 'master'
+        required: true
+      go_version:
+        description: 'Go version for building binaries'
+        default: '1.17.x'
+        required: true
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    env:
+      VERSION: ${{ github.event.inputs.k6_version }}
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.inputs.k6_branch_or_commit }}
+      - name: Install Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: ${{ github.event.inputs.go_version }}
+      - name: Install package builders
+        env:
+          GO111MODULE: 'off'
+        run: |
+          gopath="$(go env GOPATH)"
+          go get github.com/Masterminds/glide
+          go get -d github.com/mh-cbon/go-bin-deb \
+              && (cd "$gopath/src/github.com/mh-cbon/go-bin-deb" \
+              && glide install \
+              && go install)
+          go get -d github.com/mh-cbon/go-bin-rpm \
+              && (cd "$gopath/src/github.com/mh-cbon/go-bin-rpm" \
+              && glide install \
+              && go install)
+          sudo apt-get update -y
+          sudo apt-get install -y fakeroot rpm
+      - name: Generate Windows binary metadata
+        run: |
+          (cd && GO111MODULE=on go get github.com/josephspurrier/goversioninfo/cmd/goversioninfo@v1.2.0)
+          IFS=. read -a version <<< "$(echo $VERSION | sed 's:[^0-9\.]::g')"
+          # Need a blank versioninfo.json for the CLI overrides to work.
+          echo '{}' > versioninfo.json
+          goversioninfo -64 \
+            -platform-specific=true \
+            -charset="1200" \
+            -company="Load Impact AB" \
+            -copyright="© Load Impact AB. Licensed under AGPL." \
+            -description="A modern load testing tool, using Go and JavaScript" \
+            -icon=packaging/k6.ico \
+            -internal-name="k6" \
+            -original-name="k6.exe" \
+            -product-name="k6" \
+            -translation="0x0409" \
+            -ver-major="${version[0]}" \
+            -ver-minor="${version[1]}" \
+            -ver-patch="${version[2]}" \
+            -product-version="${VERSION#v}"
+      - name: Build
+        run: |
+          go version
+          ./build-release.sh "dist" ${VERSION}
+      - name: Upload artifacts
+        uses: actions/upload-artifact@v2
+        with:
+          name: binaries
+          path: dist/
+          retention-days: 1
+
+  package-windows:
+    runs-on: windows-2019
+    defaults:
+      run:
+        shell: powershell
+    needs: [build]
+    env:
+      VERSION: ${{ github.event.inputs.k6_version }}
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.inputs.k6_branch_or_commit }}
+      - name: Install pandoc
+        uses: crazy-max/ghaction-chocolatey@b6061d587628735be315d74358228b83a7dba9a7
+        with:
+          args: install -y pandoc
+      - name: Install wix tools
+        run: |
+          curl -O wix311-binaries.zip https://github.com/wixtoolset/wix3/releases/download/wix3112rtm/wix311-binaries.zip
+          Expand-Archive -Path .\wix311-binaries.zip -DestinationPath .\wix311\
+          echo "$pwd\wix311" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
+      - name: Download binaries
+        uses: actions/download-artifact@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/dev-release.yml
```diff
@@ -0,0 +1,168 @@
...
+        run: |
+          (cd && GO111MODULE=on go get github.com/josephspurrier/goversioninfo/cmd/goversioninfo@v1.2.0)
+          IFS=. read -a version <<< "$(echo $VERSION | sed 's:[^0-9\.]::g')"
+          # Need a blank versioninfo.json for the CLI overrides to work.
+          echo '{}' > versioninfo.json
+          goversioninfo -64 \
+            -platform-specific=true \
+            -charset="1200" \
+            -company="Grafana Labs AB" \
+            -copyright="© Grafana Labs AB. Licensed under AGPL." \
```
na--, line 90: This is not the correct legal name, I think

.github/workflows/dev-release.yml
```diff
@@ -0,0 +1,168 @@
...
+            -icon=packaging/k6.ico \
+            -internal-name="k6" \
+            -original-name="k6.exe" \
+            -product-name="k6" \
+            -translation="0x0409" \
+            -product-version="${VERSION#v}"
+      - name: Build
+        run: |
+          go version
+          ./build-dev-release.sh
```
na--, line 101: This file doesn't exist :confused: You forgot to rename `build-release.sh`?

codebien: :face_with_spiral_eyes: this is an error copying from my fork. :man_facepalming:

build-release.sh
```diff
@@ -36,14 +36,14 @@ build() {
...
-    local NAME="k6-${VERSION}-${ALIAS}"
+    local NAME="k6-dev-${VERSION}-${ALIAS}"
     echo "- Creating ${NAME}.${FMT} package..."
     case $FMT in
     deb|rpm)
         # The go-bin-* tools expect the binary in /tmp/
         [ ! -r /tmp/k6 ] && cp "${OUT_DIR}/${NAME}/k6" /tmp/k6
         "go-bin-${FMT}" generate --file "packaging/${FMT}.json" -a amd64 \
-            --version "${VERSION#v}" -o "${OUT_DIR}/k6-${VERSION}-linux-amd64.${FMT}"
+            --version "0.38.0" -o "${OUT_DIR}/k6-${VERSION}-linux-amd64.${FMT}"
```
na--, line 46: Why this change?

.github/workflows/dev-release.yml
```diff
@@ -0,0 +1,168 @@
...
+on:
+  # Enable manually triggering this workflow via the API or web UI
+  workflow_dispatch:
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  test:
```
na--, line 11: I think you can safely remove `test` from here :thinking: No point in having it just waste time, right?

build-release.sh
na--: hmm how about `vX.Y.Z-something`?

build-release.sh
```diff
@@ -5,8 +5,11 @@ set -eEuo pipefail
 eval "$(go env)"
 
 OUT_DIR="${1-dist}"
-# To override the latest git tag as the version, pass something else as the second arg.
-VERSION=${2:-$(git describe --tags --always --dirty)}
+
+if [ -z "$VERSION" ]; then
+    # To override the latest git tag as the version, pass something else as the second arg.
+    VERSION=${2:-$(git describe --tags --always --dirty)}
+fi
```
codebien, line 12: This seems something old, is it still required? The `VERSION` value should be always set from the `configure` job.

codebien, line 12: This seems something old, is it still required? The VERSION value should be always set from the `configure` job and in any case, the tags are not available, the checkout action doesn't fetch them.

.github/workflows/dev-release.yml
codebien: I will change this in a dedicated PR when I will have the right name.

build-release.sh
imiric: I think it's still valuable if we ever want to run this outside of CI, e.g. locally for some reason.

.github/workflows/dev-release.yml
```diff
@@ -0,0 +1,144 @@
...
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: Install pandoc
+        uses: crazy-max/ghaction-chocolatey@b6061d587628735be315d74358228b83a7dba9a7
+        with:
+          args: install -y pandoc
+      - name: Install wix tools
+        run: |
+          curl -O wix311-binaries.zip https://github.com/wixtoolset/wix3/releases/download/wix3112rtm/wix311-binaries.zip
```
imiric, line 102:
```suggestion
          curl -Lso wix311-binaries.zip https://github.com/wixtoolset/wix3/releases/download/wix3112rtm/wix311-binaries.zip
```

We recently had to fix this in https://github.com/grafana/xk6-browser/commit/cbe11c4231a92f32d61d0d8a375d6e7595722fe7, since GitHub added a redirect and the download was failing without `-L`.

build-release.sh
codebien: Ok, then I think we can keep it with the `if` condition.

imiric: Sure, though the `if` is not strictly necessary, since the parameter substitution serves the same purpose if the version is passed as an argument. But it's fine.

imiric (dismissed):
LGTM, but wait for another approval :)

codebien:
@imiric Apparently you moved to `pwsh` before https://github.com/grafana/xk6-browser/commit/36adf8cc32cbd43f6ece520e1aede6db1849e06c, `powershell` seems not to support this config. I can consider migrating entirely to `pwsh` but I would prefer to do it in another PR. I'm going to roll back.

imiric:
Hhmm PowerShell Core worked fine for our scripts, and we could remove the UTF-8 option for the GH env var. What issue are you running into?

codebien:
```
Run curl -Lso wix311-binaries.zip https://github.com/wixtoolset/wix3/releases/download/wix3112rtm/wix311-binaries.zip
Invoke-WebRequest : A parameter cannot be found that matches parameter name 'Lso'.
At D:\a\_temp\0c5f23b0-f239-4e54-baae-0d261b5d675a.ps1:2 char:6
+ curl -Lso wix311-binaries.zip https://github.com/wixtoolset/wix3/rele ...
+      ~~~~
    + CategoryInfo          : InvalidArgument: (:) [Invoke-WebRequest], ParentContainsErrorRecordException
    + FullyQualifiedErrorId : NamedParameterNotFound,Microsoft.PowerShell.Commands.InvokeWebRequestCommand
```
As reported in private, it requires Powershell Core so I would like to migrate to it, but I would prefer to do it in a dedicated PR.

build-release.sh
na--: I am confused by this, `${2:-$(git describe --tags --always --dirty)}` means take the second positional argument or use the result of `git describe ...` if that was empty. Why should we also support an external `VERSION` env var, we are only using positional arguments everywhere else in the script :confused: 

Soft :-1: for this, unless we decide to refactor the whole script, but that should happen in some other PR

.github/workflows/dev-release.yml
```diff
@@ -0,0 +1,168 @@
...
+            export GOMAXPROCS=1
+          fi
+          go test "${args[@]}" -timeout 800s ./...
+
+  configure:
+    runs-on: ubuntu-latest
+    outputs:
+      version: ${{ steps.get_version.outputs.version }}
+    steps:
+      - name: Get version
```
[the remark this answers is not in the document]
na--: I see the `workflow_dispatch`, so do we need this job here? Can't you directly use `${{ github.event.inputs.k6_version }}` in the jobs below?

.github/workflows/dev-release.yml
```diff
@@ -0,0 +1,144 @@
...
+            -icon=packaging/k6.ico \
+            -internal-name="k6" \
+            -original-name="k6.exe" \
+            -product-name="k6" \
+            -translation="0x0409" \
+            -product-version="${VERSION#v}"
+      - name: Build
+        run: |
+          go version
+          ./build-release.sh
```
na--, line 77: Won't this work? It will remove the need to set `VERSION` here and support it in the built-release script

```suggestion
          ./build-release.sh "dist/" "${{ github.event.inputs.k6_version }}"
```

codebien:
I applied the suggestions and re-tested them, I also added the branch input so we can build eventual PRs not yet merged.

.github/workflows/dev-release.yml
```diff
@@ -0,0 +1,135 @@
...
+      VERSION: ${{ github.event.inputs.k6_version }}
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.inputs.k6_branch_or_commit }}
+      - name: Install Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.17.x
```
na--, line 30: hmm it might be nice if this is a parameter too!

na-- (dismissed):

na-- (approved):

.github/workflows/dev-release.yml
```diff
@@ -0,0 +1,141 @@
...
+            -charset="1200" \
+            -company="Grafana Labs AB" \
+            -copyright="© Grafana Labs AB. Licensed under AGPL." \
+            -description="A modern load testing tool, using Go and JavaScript" \
+            -icon=packaging/k6.ico \
+            -internal-name="k6" \
+            -original-name="k6.exe" \
+            -product-name="k6" \
+            -translation="0x0409" \
+            -product-version="${VERSION#v}"
```
imiric, line 68: I think we need the `-ver-major`, `-ver-minor`, `-ver-patch` flags here, in addition to `-product-version`. Otherwise the version won't be shown properly in the Properties AFAIR. [That's what we do in the main workflow](https://github.com/grafana/k6/blob/59a8883676edf4e28edb39561c33ebd941e05944/.github/workflows/all.yml#L191-L194).

If these are not needed, then the array created on line 54 is not needed either.

imiric (changes requested):

imiric (approved):
:+1:

na-- (approved):

PR #2467, opened 2022-03-31
Lint k6/html

MStoykov:


MStoykov:
This brings the number of lint errors way down, but not to zero. I purposefully skipped over any "exported * needs comment" as that will take *quite* a while in this package.
Without them the errors are a handful and maybe we should `nolint` some more of them. 
```
$ golangci-lint run --out-format=tab ./ |grep -v "exported: exported "
element.go:123:2   revive      var-naming: var formId should be formID
element.go:246:18  revive      var-naming: method Id should be ID
html.go:312:20     revive      var-naming: method Html should be HTML
html.go:240:1      revive      exported: comment on exported method Selection.PrevUntil should be of the form "PrevUntil ..."
html.go:492:1      revive      exported: comment on exported method Selection.Data should be of the form "Data ..."
util.go:68:6       revive      var-naming: func getHtmlAttr should be getHTMLAttr
elements.go:426:2  revive      var-naming: var listId should be listID
elements.go:531:2  revive      var-naming: var formId should be formID
element.go:416:2   exhaustive  missing cases in switch of type html.NodeType: ErrorNode, RawNode
element.go:438:2   exhaustive  missing cases in switch of type html.NodeType: DoctypeNode, DocumentNode, ElementNode, ErrorNode, RawNode
```

Especially the ones asking to rename methods likely will not be possible at all as they might actually break the API, although this needs checking as well.

oleiade (approved):
👍🏻

olegbespalov (approved):

PR #2469, opened 2022-04-01
Lint lib

MStoykov:


MStoykov:
This cuts the lint errors from 519 to 157 - a lot of those are also package level documentation.

 But I would prefer to fix any other lint errors another time as the other kinds take more work :(

na-- (dismissed):
LGTM
It'd be really nice if the `testing` package has a `t.RunParallel()` function for sub-tests, so we can avoid this `t.Parallel()` line everywhere...

MStoykov:
Sorry @na-- I needed to fix one more lint and found out we have a bunch of not used helper functions ... and then badssl broke :sob:

na-- (approved):

oleiade (approved):

PR #2472, opened 2022-04-02
Make redirects without Location still error in go-tip

MStoykov:
This is probably temporary, but it's probably better to keep this
behaviour for a while, or at least until a version of go with this
change is released.

### DIFF
--- a/lib/netext/httpext/request.go
+++ b/lib/netext/httpext/request.go
@@ -293,7 +293,7 @@ func MakeRequest(ctx context.Context, state *lib.State, preq *ParsedHTTPRequest)
 	// make it an error to not have Location header on a redirect
 	if res != nil && resErr == nil {
 		switch res.StatusCode {
-		case 301, 302, 303:
+		case http.StatusMovedPermanently, http.StatusFound, http.StatusSeeOther:
 			if res.Header.Get("Location") == "" {
 				resErr = NewK6Error(defaultErrorCode, fmt.Sprintf("%d response missing Location header", res.StatusCode), nil)
 			}
--- a/lib/netext/httpext/request.go
+++ b/lib/netext/httpext/request.go
@@ -286,6 +286,19 @@ func MakeRequest(ctx context.Context, state *lib.State, preq *ParsedHTTPRequest)
 	mreq := preq.Req.WithContext(reqCtx)
 	res, resErr := client.Do(mreq)
 
+	// due to https://github.com/golang/go/commit/a41763539c7ad09a22720a517a28e6018ca4db0f
+	// redirects with missing Location are no longer errors. But they were up to that point
+	// (possibly realeased in go 1.19).
+	// In order to not have a change in behaviour depending on the version compiled we still
+	// make it an error to not have Location header on a redirect
+	if res != nil && resErr == nil {
+		switch res.StatusCode {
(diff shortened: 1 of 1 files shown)


lib/netext/httpext/request.go
```diff
@@ -286,6 +286,19 @@ func MakeRequest(ctx context.Context, state *lib.State, preq *ParsedHTTPRequest)
...
 	res, resErr := client.Do(mreq)
 
+	// due to https://github.com/golang/go/commit/a41763539c7ad09a22720a517a28e6018ca4db0f
+	// redirects with missing Location are no longer errors. But they were up to that point
+	// (possibly realeased in go 1.19).
+	// In order to not have a change in behaviour depending on the version compiled we still
+	// make it an error to not have Location header on a redirect
+	if res != nil && resErr == nil {
+		switch res.StatusCode {
+		case 301, 302, 303:
```
olegbespalov, line 296: I'd suggest using the `http` constants instead of numbers here:
`http.StatusMovedPermanently`, `http.StatusFound` and `http.StatusSeeOther`

That will be more human-friendly

na-- (changes requested):
I don't think this is the best solution, since it would require us to go back in the future and remove this code. Instead, I propose that this test: https://github.com/grafana/k6/blob/9fefa4ab68749a54f7dbf48eaf209a83bc817184/js/modules/k6/http/request_test.go#L2026-L2031

Should be gated behind a gotip / Go 1.18 build flag so it only runs on the the new versions of Go that have the bug fixed.

MStoykov:
This only happens on `gotip` so not in `go1.18` and (IMO) it's unlikely to be back ported. This means that it will first (possibly) appear in 1.19. 

At that point (as I mentioned privately) it will mean that compiling with `1.19` or `1.18` will have different behavior on this (arguably) rare case. 

For me, it seems a lot more stable to keep the same behavior across versions. We can always decide to break this and have the correct behavior once a version of go that will support it is released. 

Given the fact that nobody has complained to us about this behavior it means that even if it was hit it was not a problem. Leaving this dependent on the go version and making our test *just* pass, while having different behaviors has a chance (be it small) that it will break somebody script. Even more importantly, as we probably will revert this at some future point when there are more go version released that support it, it will break based on which version of go is used to compile. This might mean that if they compile locally it will work or break.

na--:
Again, a bugfix can be a breaking change, and in this case it's a minor one, but that's beside the point. The fact that in this case it's dependent on a new Go version is irrelevant. 

As a slightly better alternative to fixing the test and having it run only on Go 1.19, I suggest we just remove it. Its behavior and the RFC compliance is [already tested in the Go tests](https://github.com/golang/go/commit/a41763539c7ad09a22720a517a28e6018ca4db0f#diff-f12c66d39f4770beee6b1076c40e115cdaf3858873027689693f2410a1d1c819R536), so there is no need to repeat it here.

lib/netext/httpext/request.go
oleiade: I would concur with this comment. Why is it resolved @olegbespalov ?

olegbespalov: @oleiade it's resolved because the code was adjusted according to suggestion

MStoykov:
Closed in favor of #2473

PR #2481, opened 2022-04-06
+ TLS key logging to enable Wireshark SSL support

kkriegkxs:
**Related issue**: https://github.com/grafana/k6/issues/2482

**Problem:**
Add SSL decryption support for CAP e.g. Wireshark in the same fashion as Chrome & Firefox do:
if env SSLKEYLOGFILE is set, output keys to it

**Implementation:**
Extend tlsConfig to add KeyLogWriter (https://pkg.go.dev/crypto/tls#Config.KeyLogWriter) if a corresponding debug flag is set

### DIFF
--- a/cmd/options.go
+++ b/cmd/options.go
@@ -99,6 +99,8 @@ func optionFlagSet() *pflag.FlagSet {
 		"Milliseconds are assumed if no unit is provided.\n"+
 		"Possible select values to return a single IP are: 'first', 'random' or 'roundRobin'.\n"+
 		"Possible policy values are: 'preferIPv4', 'preferIPv6', 'onlyIPv4', 'onlyIPv6' or 'any'.\n")
+	flags.Bool("sslkeylog", false, "If enabled, output TLS secrets to env:SSLKEYLOGFILE")
+
 	return flags
 }
 
@@ -129,6 +131,8 @@ func getOptions(flags *pflag.FlagSet) (lib.Options, error) {
 		TeardownTimeout: types.NullDuration{Duration: types.Duration(60 * time.Second), Valid: false},
 
 		MetricSamplesBufferSize: null.NewInt(1000, false),
+
+		SslKeyLog: getNullBool(flags, "sslkeylog"),
 	}
 
 	// Using Changed() because GetStringSlice() doesn't differentiate between empty and no value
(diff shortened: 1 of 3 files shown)


kkriegkxs:
Hey guys!
K6 is awesome, so I figured might as well contribute to the community. 

The logic is similar to the way the browsers handle it: is there is an env var SSLKEYLOGFILE is set, log keys using the standard go implementation in tls.Config.KeyLogWriter, however only if (new) Option SslKeyLog is set (regardless of SSLKEYLOGFILE's value)

MStoykov:
Hi thanks for the PR, but firstly there are two other PRs before that the latest of which is https://github.com/grafana/k6/pull/2210. In which you can see that the k6 team would prefer if the file is closed, which doesn't seem to be addressed here either. Which isn't surprising as that will be kind of hard currently, due to lack of signalling that k6 will be stopped instead of just a scenario ending.

Additionally you recreate tlsconfig in this one which seems unnecessary.

kkriegkxs:
@MStoykov the file handle persist for the entire duration of the connection and IMO is irrelevant from the performance hit perspective as key logging should only be enabled for debugging and hence low volume (if not single user) - while k6 team is not wrong to want it done properly, it shouldn't matter much.
I chose to recreate the object in it's entirety when flag enabled to underscore the possibility of other additional debug changes - which may be excessive but is explicit (and has no performance implications in this context). This can be simplified.

I've looked the other PRs - regardless of which is chosen, I recommend mirroring the Mozilla & Chrome implementation logic: SSLKEYLOGFILE for the file as opposed to K6 destination - it is an all around better way because sniffers like Wireshark use (by convention) a single key log file and it would be a paint to constantly reconfigure it. Consequently you only want K6 to log the keys if it has a corresponding option enabled - not always: you can and often will have a case when you want browsers to log keys but not K6 (and WARN when you do).

kkriegkxs:
Also please remember that if you want to implement large scale capture, you'd need a separate reverse proxy **cluster** (squid/nginx etc.) with request logging or something to that tune because otherwise you will negatively impact the load generator, and an extra few files handles per VU of key logging will be the least of your concerns (think of CAP inspecting the streams and then either dumping them or keeping them in mem).  
Otherwise in a low volume debug situation, the extra file handles are irrelevant and increasing the code complexity just to do it right in 0.1% out of 99.9% of the time is detrimental.  
IMO :)

MStoykov:
Thank for openning this @kkriegkxs :bow: . 

While I don't disagree completely with some of your points, we would prefer to have better solution when possible. 

I have opened https://github.com/grafana/k6/pull/2487 which seems to have fixed most of our problems with the previous implementation.
 It seems likely that it will ship in the next release :tada: , so I am closing this PR, thanks again.

kkriegkxs:
Looked at your change - good stuff! Happy this is going to be implemented

PR #2484, opened 2022-04-07
lib/netext/grpcext: Split the gRPC module

### LINKED ISSUE #2315: Provide an interface to allow customized gRPC options for the gRPC client
### Feature Description

The current connect [interface](https://github.com/grafana/k6/blob/master/js/modules/k6/grpc/client.go#L217) in the gRPC client uses a pre-defined gRPC options. This limits the capability of the gRPC client in that the dial options can't be configured in a customized gRPC connectivity environment. For an environment with more complex connectivity configurations such as with gRPC look-aside loadbalancer or authentications, an interface with a customized gRPC options is needed. The users can leverage the current gRPC client by creating a k6 extension to pass in the customized gRPC options.

_No response_

_No response_

codebien:
The split of the gRPC business logic from the `js/k6/grpc` module to a `netext` package.
It enables a simpler process for extending the networking+metric features for gRPC without the k6/js dependencies.

An example of how the API of this package can be used for creating gRPC extensions: [codebien/xk6-grpc-example](https://github.com/codebien/xk6-grpc-example) 

Closes #2315


lib/netext/grpcext/conn.go
```diff
@@ -0,0 +1,305 @@
...
+
+// ReflectionClient returns a reflection client based on the current connection.
+func (c *Conn) ReflectionClient() (*ReflectionClient, error) {
+	return &ReflectionClient{Conn: c.raw}, nil
+}
+
+// Invoke executes a unary gRPC request.
+//
+// TODO: should we support grpc.CallOption?
+func (c *Conn) Invoke(ctx context.Context, method string, md metadata.MD, req Request) (*Response, error) {
```
na--, line 90: > should we support grpc.CallOption?

Yes, I think this might be a good idea, given that they are present in [`grpc.ClientConn.Invoke()`](https://pkg.go.dev/google.golang.org/grpc#ClientConn.Invoke). Can you think of a reason why we shouldn't support them now, even if we don't use them yet?

Also, do we need a `method` parameter here when we have a `MethodDescriptor` in the request?

lib/netext/grpcext/conn.go
```diff
@@ -0,0 +1,305 @@
...
+		Indent:    "  ",
+	}
+	b, err := marshaler.Marshal(msg)
+	if err != nil {
+		return ""
+	}
+	return string(b)
+}
+
+type ctxKeyTags struct{}
```
na--, line 288: you haven't deleted these from the js package, it seems: https://github.com/grafana/k6/blob/grpcext/js/modules/k6/grpc/tags.go

js/modules/k6/grpc/client.go
```diff
@@ -353,49 +253,22 @@ func (c *Client) dial(
 	reflect bool,
 	options ...grpc.DialOption,
 ) error {
```
na--, line 255: This `Client.dial()` code doesn't seem it has anything to do with JavaScript, what is the reason to not move it to `grpcext`?

codebien: It opens the connection then it routes through reflection if requested from the parameter, I think this way of connecting it's a property of the js module. :thinking:  It could be a good idea to change the name of the method that probably it's not accurate after the refactor because the real dialling part is now delegated to `grpcext`. This method instead is mostly the core part of the `Connect` method.

lib/netext/grpcext/conn.go
codebien: Unfortunately, yes. The `FullName` of the method descriptor returns the full name domain as a dot notation. Instead, Invoke wants the `package.serviceName/methodName` format. I don't know if the simple replace of the last dot with a slash could be fine. Apparently, also the [protoc generator](https://github.com/golang/protobuf/blob/5d5e8c018a13017f9d5b8bf4fad64aaa42a87308/protoc-gen-go/grpc/grpc.go#L312) generates the url instead of use+replace the method descriptor's full name.

lib/netext/grpcext/conn.go
```diff
@@ -0,0 +1,313 @@
...
+	md metadata.MD,
+	req Request,
+	opts ...grpc.CallOption) (*Response, error) {
+	if method == "" {
+		return nil, fmt.Errorf("method is required")
+	}
+	if req.MethodDescriptor == nil {
+		return nil, fmt.Errorf("request method descriptor is required")
+	}
+	if len(req.Message) == 0 {
```
olegbespalov, line 100: :thinking: should a message always be presented?

lib/netext/grpcext/conn.go
```diff
@@ -0,0 +1,313 @@
...
+
+	protov1 "github.com/golang/protobuf/proto" //nolint:staticcheck,nolintlint // this is the old v1 version
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/metadata"
+	grpcstats "google.golang.org/grpc/stats"
+	"google.golang.org/grpc/status"
+	"google.golang.org/protobuf/encoding/protojson"
+	"google.golang.org/protobuf/encoding/prototext"
+	"google.golang.org/protobuf/proto"
```
cygnrh, line 24: Do you consider using `github.com/gogo/protobuf/proto` which has faster mashalling/unmarshalling?

lib/netext/grpcext/conn.go
```diff
@@ -0,0 +1,313 @@
...
+
+// Dial establish a gRPC connection.
+func Dial(ctx context.Context, addr string, options ...grpc.DialOption) (*Conn, error) {
+	conn, err := grpc.DialContext(ctx, addr, options...)
+	if err != nil {
+		return nil, err
+	}
+	return &Conn{
+		raw: conn,
+	}, nil
```
cygnrh, line 79: how about including options in this struct so we don't have to pass options again for Invoke?

lib/netext/grpcext/conn.go
codebien: Hi @cygnrh, thanks for your review.

Unfortunately, we would prefer to not include it, we are very meticulous in adding new dependencies to `k6`, you can read some motivations in [Dependencies.md](https://github.com/grafana/k6/blob/master/Dependencies.md).
We would prefer to fix allowing the marshaler and unmarshaler to be injected, it would allow us to remove also the tightly coupled JSON dependency. I opened an issue for it https://github.com/grafana/k6/issues/2497

If you have relevant observations for the grpc module, it would be helpful to get your feedback in a new issue or in  https://github.com/grafana/k6/issues/1846.

lib/netext/grpcext/conn.go
codebien: I think [WithDefaultCallOptions](https://pkg.go.dev/google.golang.org/grpc#WithDefaultCallOptions) covers this case. Could it work for you?

lib/netext/grpcext/conn.go
codebien: Yes, at least a `{}` must be there. The change in https://github.com/grafana/k6/pull/2484/commits/8b64a1414a1939fceef051447911c21f039831bf should make it visible. This is related to the `protojson` dependency, where a valid JSON must be used.

lib/netext/grpcext/conn.go
cygnrh: I'm a bit confused that options ...grpc.DialOption is passed in from `Dial` and `Invoke`. If options is part of the Conn struct, `Invoke` can use it as a receiver function.

codebien: Would you add a slice DialOption field?
```go
type Conn struct {
        Options []grpc.DialOption

	raw clientConnCloser
}
```
and calling the `Conn.Invoke` method, would you like to not pass the Call Options?

```go
options := []grpc.DialOption{
   .....
}
conn, err := Dial(ctx, addr, options...)
err = conn.Invoke(ctx, url, md)
```

If yes, then the issue is mainly related to the different types involved. The eventual `Conn.Options` couldn't be used on the underhood call `grpc.Invoke` because the Invoke method wants the CallOption type.

Please, correct me if I misunderstood.

js/modules/k6/grpc/client.go
olegbespalov: But it seems like the vast majority of the logic related to the reflection is still can be moved to a `grpcext`. 

I belive we can even make a private a `ReflectionClient`. What js nodule needs are the `*descriptorpb.FileDescriptorSet`, so the `grpcext.Conn` can has this method. What do your think?

olegbespalov: And by the number of the logic, the current `Client.dial` is more like `Client.reflect` than dial :sweat_smile:

lib/netext/grpcext/conn.go
MStoykov: I guess a User can just use https://pkg.go.dev/google.golang.org/grpc#WithDefaultCallOptions as well

MStoykov (approved):
LGTM and IMO we can continue iterating in the next release on any of the currently commented stuff

olegbespalov (approved):

lib/netext/grpcext/conn.go
cygnrh: I see. They are different types of options. That makes sense.

lib/netext/grpcext/conn.go
cygnrh: We haven't observed performance issue since we only did the initial test. If we do see anything as k6 gRPC being more widely adopted, I'll post in the thread https://github.com/grafana/k6/issues/1846.

Issue #2432, opened 2022-03-10
Hooks for extensions - e.g. graceful shutdown, per VU/iteration, etc.


imiric:
### Feature Description

Currently k6 extensions are abruptly terminated in some scenarios, e.g. when pressing `^C`, as we've found out in https://github.com/grafana/xk6-browser/pull/268#discussion_r821905588. This doesn't leave any time for extensions to cleanup and do a clean shutdown, which in xk6-browser's case would involve closing the browser connection, though in general k6 should ensure that any pending event loop operations are processed as well. For example, it would be good DX to allow the Promise rejection handlers to run, but this is up for debate whether `^C` should interrupt those as well.

### Suggested Solution

One or more of these might be a good idea:

- Passing a cancelable context to extensions via the `moduleVUImpl` [here](https://github.com/grafana/k6/blob/fb320f6df8a1694477d96d2f57d982fae1c3d2d1/js/runner.go#L153). This context could then be cancelled first whenever k6 starts shutting down (whether that's abruptly with `^C` or otherwise).

- Allow extensions to specify a callback for "deregistration"/cleanup, either via the `modules.Register()` method, or another `moduleVUImpl` method. k6 would then run this first before shutdown. A drawback with this might be that callbacks might run an arbitrary amount of time, but we can always timeout on it.

- Maybe alternatively to the point above, pass a "shutdown" channel to extensions that extensions can use to signal that they're done with cleanup. This might be more flexible and simpler to timeout on.

- If we do decide to have a timeout, it should probably use a sane default value but be user configurable.

### Already existing or connected issues / PRs

- https://github.com/grafana/xk6-browser/pull/268

MStoykov:
I feel like maybe some kind of full blown event system that possibly is also available to js scripts will be better solution. Some links:
- https://doc.deno.land/deno/stable/~/Event
- https://doc.deno.land/deno/stable/~/ErrorEvent
- https://doc.deno.land/deno/stable/~/addEventListener
- https://doc.deno.land/deno/stable/~/dispatchEvent
- https://doc.deno.land/deno/stable/~/MessageEvent
- https://nodejs.org/dist/latest-v12.x/docs/api/events.html#events_passing_arguments_and_this_to_listeners
- https://developer.mozilla.org/en-US/docs/Web/Events

mstoykov:
For your particular case I would expect that you can workaround, as long as you don't need anything to be executed on the event loop - js code usually:

[...]

PR #2485, opened 2022-04-07
SSLKEYLOGFILE support

### LINKED ISSUE #1043: Support dumping of crypto keys to an nss formated key log file

MStoykov:
fixes #1043 

During some of the previous discussions we really wanted to close the file at the end of the test. Unfortunately there isn't really a good way for this currently and arguably it will take a lot of time so the solutions (AFAIK) are:
1. don't close the file - in this case we can just open it in `js.Runner` and give it to all `VUs`. The file will be closed when the process exits, possibly not flushing a line ... I guess (although os.File is not buffered by default). 
2. Wait for a rewrite to get us a signal for "k6 is stopping" to close the file. This arguably is connected with way too many other life-time issues in k6 and is unlikely to be easy to fix. 
3. Move this outside of the `js` package and push it in. This in practice will mean more arguments, but may let us do it in `cmd/run.go` for example :shrug:, Which does know when a k6 will stop executing
4. Do some "reference counting" on activation and deactivation of VUs and open/close the file when appropriate - what this PR does.

### DIFF
--- a/cmd/runtime_options.go
+++ b/cmd/runtime_options.go
@@ -117,6 +117,12 @@ func getRuntimeOptions(flags *pflag.FlagSet, environment map[string]string) (lib
 		}
 	}
 
+	if envVar, ok := environment["SSLKEYLOGFILE"]; ok {
+		if !opts.KeyWriter.Valid {
+			opts.KeyWriter = null.StringFrom(envVar)
+		}
+	}
+
 	if opts.IncludeSystemEnvVars.Bool { // If enabled, gather the actual system environment variables
 		opts.Env = environment
 	}
--- a/js/keylogger.go
+++ b/js/keylogger.go
@@ -1,0 +1,17 @@
+package js
+
+import (
+	"os"
+	"sync"
+)
+
+type keyloggerWriter struct {
+	f *os.File
+	m *sync.Mutex
+}
+
+func (k *keyloggerWriter) Write(b []byte) (int, error) {
+	k.m.Lock()
+	defer k.m.Unlock()
+	return k.f.Write(b)
+}
--- a/js/runner.go
+++ b/js/runner.go
@@ -80,8 +80,8 @@ type Runner struct {
 	console   *console
 	setupData []byte
 
-	keylogger      *os.File
-	keyloggermutex *sync.Mutex
+	keylogger      *keyloggerWriter
+	keyloggerMutex *sync.Mutex
 	keyloggerUsers int
 }
 
@@ -136,7 +136,7 @@ func NewFromBundle(
 		ActualResolver: net.LookupIP,
 		builtinMetrics: builtinMetrics,
 		registry:       registry,
-		keyloggermutex: new(sync.Mutex),
+		keyloggerMutex: new(sync.Mutex),
 	}
 
 	err = r.SetOptions(r.Bundle.Options)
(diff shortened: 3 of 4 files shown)


na-- (changes requested):
I agree with the options you have given and I can't think of any new ones. However, my preference for a solution is option number 3, to move this outside of the `js` package and pass a single synchronized `io.Writer` to all. With a potential option to rewrite it and get it out of `cmd/` when 2. (probably from https://github.com/grafana/k6/issues/2432?) is available.

I don't think your current approach of using `js.Runner.getKeylogger()` and `js.Runner.freeKeylogger()` at VU activation/deactivation is a very good one, and the complex synchronization you're having to do is one of the main reasons I think so. Another big reason why I don't think this should be in the `js/` package at all is I'd prefer not to have `os` calls that interact with the real file system there, at all, even if it was used only in a single place. 

for me, this seems like it should be done only once and a synchronized `io.Writer` and passed to every VU externally :thinking_face: so, maybe somewhere in the VU init code, as I mentioned above, or maybe even in the `loadAndConfigureTest()` function in `cmd`

na-- (changes requested):
> So you want in the future it to be in the js ?

Probably I'd still prefer to avoid that, if possible :thinking: A lot of the current TLS and network setup that's currently in `js/` probably should be in `netext` / `httpext` anyway, but that's likely going to be a separate refactoring from https://github.com/grafana/k6/issues/2432 :disappointed: 

> Unfortunately I can't fix that with using afero.FS~ as it's purposfully read only~ it actually seems to not be so this might be away?!?

The second part of this sentence doesn't parse, but yes, I'd prefer to open a file for writing only outside of `js/`. Using the `afero` FS instead of the OS FS inside is also probably OK, though that might have its own issues and require further refactoring because of the cache-on-read and other layers we have :man_shrugging: 

> 5. open the file in cmd/run.go and put it in runtimeOptions (or just the Option it currently uses there).

Hmm :thinking: Something like this might actually work out nicely! I wouldn't want to pass the file handle directly in `lib.RuntimeOptions`, that would seem weird. However, we can refactor the code a bit more and introduce a new `RuntimeState` struct (for example) that contains both the current `RuntimeOptions` and some some extra state like this file handle. We can then pass this broader struct to the `js` code (and other places) in place of the current `RuntimeOptions`.

na--:
Thinking about it some more, I think there is already need for such a `RuntimeState` struct, or whatever we decide to call it... Besides this file handle, it can contain a bunch of other state that gets initialized in `cmd/run.go`, but then is passed as individual parameters:
- the `RuntimeOptions`, as mentioned, but also the regular `Options` / `Config` struct!
- the metrics `Registry` and `BuiltinMetrics`
- the logger
- the `afero` filesystems
- potentially some of the things we currently attach to the contexts, though I am not so sure about these

MStoykov:
is this https://github.com/grafana/k6/pull/2487 what you have in mind @na--? It needs more work - I will need to basically only have one API, but didn't want to change all the tests ... and will have to add tests as well.

na--:
I answered in that PR, but for clarity, yes, something like https://github.com/grafana/k6/pull/2487 seems like a better approach to me :+1:

MStoykov:
Closing in favor of #2487

PR #2486, opened 2022-04-07
cmd/version: Removed unexpected printline

codebien:


MStoykov (dismissed):
LGTM! I would've preferd if you just concat stromgs three times though instead of getting `fmt` but this might be slightly more readable.

na-- (commented):
Sorry, I fixed this locally the other day and then forgot to push the PR :man_facepalming: 

I am confused though, what is the test actually testing? It almost repeats the actual implementation, so I am not sure if it's actually a good test :confused:

codebien:
>What is the test actually testing?

@na-- The goal is to match the expected string as close as possible, for avoiding any possible breaking change in the future. This is system/version dependent so it isn't possible to check `assert.Equal(t, exp, v)`. If you have a better suggestion, I'm happy to implement it.

na--:
> The goal is to match the expected string as close as possible

This is my problem exactly. Your test almost contains an alternate re-implementation of the [original code](https://github.com/grafana/k6/blob/ece4633407ba2e88f82aa2781a183fd2a4278920/lib/consts/consts.go#L39-L48), which is generally not a great practice... :confused: It will probably be better to ensure the output contains the constituent parts (i.e. `consts.Version`, `runtime.GOOS`, etc.) and doesn't contain a newline, but not fix in what order they appear with the `Sprintf()` call :man_shrugging: 

Also, it doesn't test that `k6 version` actually works, it just tests `getCmdVersion()` directly. It is very easy to make it into a full integration tests, which will check more things, like this: https://github.com/grafana/k6/blob/ece4633407ba2e88f82aa2781a183fd2a4278920/cmd/integration_test.go#L122-L123

codebien:
> It will probably be better to ensure the output contains the constituent parts

@na-- generally I agree but in this case, the string is the feature, IMHO. For example, we are checking _and doesn't contain a newline_ but it wouldn't catch any other character added by mistake in the string.

In this case, for me, the main problem is the fact of not having a decoupled version so I can't mock the various envs variables. Maybe when we will add the extensions to the version we could consider having a struct injected or as part of the `globalState` making the string logic generation system independent and easier to assert.

I moved the test as an integration test so it checks more stuff.

na-- (approved):
LGTM now

We can adjust the test again, as well as refactor the version generating code, when we want to start including bundled extensions. And this might be a bit bikesheddy, but I still think that tests shouldn't be so tightly coupled to the actual implementation that they are basically re-implementing the original code and comparing the results of both :man_shrugging:

PR #2487, opened 2022-04-07
Add RuntimeState and keylogger to it

### LINKED ISSUE #1043: Support dumping of crypto keys to an nss formated key log file

MStoykov:
Based on https://github.com/grafana/k6/pull/2485#issuecomment-1091650420 
fixes #1043

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -268,6 +268,11 @@ func (c *cmdRun) run(cmd *cobra.Command, args []string) error {
 	logger.Debug("Waiting for engine processes to finish...")
 	engineWait()
 	logger.Debug("Everything has finished, exiting k6!")
+	if test.keywriter != nil {
+		if err := test.keywriter.Close(); err != nil {
+			logger.WithError(err).Warn("Error while closing the SSLKEYLOGFILE")
+		}
+	}
 	if interrupt != nil {
 		return interrupt
 	}
--- a/cmd/runtime_options.go
+++ b/cmd/runtime_options.go
@@ -117,6 +117,12 @@ func getRuntimeOptions(flags *pflag.FlagSet, environment map[string]string) (lib
 		}
 	}
 
+	if envVar, ok := environment["SSLKEYLOGFILE"]; ok {
+		if !opts.KeyWriter.Valid {
+			opts.KeyWriter = null.StringFrom(envVar)
+		}
+	}
+
 	if opts.IncludeSystemEnvVars.Bool { // If enabled, gather the actual system environment variables
 		opts.Env = environment
 	}
--- a/cmd/test_load.go
+++ b/cmd/test_load.go
@@ -4,6 +4,10 @@ import (
 	"archive/tar"
 	"bytes"
 	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"sync"
 
 	"github.com/spf13/afero"
 	"github.com/spf13/cobra"
@@ -25,12 +29,15 @@ const (
 // and configured k6 test.
 type loadedTest struct {
 	sourceRootPath  string // contains the raw string the user supplied
+	pwd             string
 	source          *loader.SourceData
+	fs              afero.Fs
 	fileSystems     map[string]afero.Fs
 	runtimeOptions  lib.RuntimeOptions
 	metricsRegistry *metrics.Registry
 	builtInMetrics  *metrics.BuiltinMetrics
 	initRunner      lib.Runner // TODO: rename to something more appropriate
+	keywriter       io.Closer
 
 	// Only set if cliConfigGetter is supplied to loadAndConfigureTest() or if
 	// consolidateDeriveAndValidateConfig() is manually called.
@@ -49,7 +56,7 @@ func loadAndConfigureTest(
 
 	sourceRootPath := args[0]
 	gs.logger.Debugf("Resolving and reading test '%s'...", sourceRootPath)
-	src, fileSystems, err := readSource(gs, sourceRootPath)
+	src, fileSystems, pwd, err := readSource(gs, sourceRootPath)
 	if err != nil {
 		return nil, err
 	}
@@ -67,8 +74,10 @@ func loadAndConfigureTest(
 
 	registry := metrics.NewRegistry()
 	test := &loadedTest{
+		pwd:             pwd,
 		sourceRootPath:  sourceRootPath,
 		source:          src,
+		fs:              gs.fs,
 		fileSystems:     fileSystems,
 		runtimeOptions:  runtimeOptions,
 		metricsRegistry: registry,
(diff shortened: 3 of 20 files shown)


na-- (commented):
:+1: Something like this LGTM. I think you're not covering the `case testTypeArchive` branch in `initializeFirstRunner()`, so some further refactoring might be needed, but I like the general approach

cmd/test_load.go
```diff
@@ -100,12 +104,25 @@ func (lt *loadedTest) initializeFirstRunner(gs *globalState) error {
 		testType = detectTestType(lt.source.Data)
 	}
 
+	state := &lib.RuntimeState{
+		Logger:         gs.logger,
+		RuntimeOptions: lt.runtimeOptions,
+		BuiltinMetrics: lt.builtInMetrics,
+		Registry:       lt.metricsRegistry,
+	}
+	if lt.runtimeOptions.KeyWriter.Valid {
```
na--, line 113:
```suggestion
	if lt.runtimeOptions.KeyWriter.Valid {
		logger.Debugf("SSLKEYLOGFILE was specified, logging TLS connection keys to '%s'...", lt.runtimeOptions.KeyWriter.String)
```

I wonder if this shouldn't be a warning message even :thinking:

cmd/run.go
```diff
@@ -268,6 +268,11 @@ func (c *cmdRun) run(cmd *cobra.Command, args []string) error {
 	logger.Debug("Waiting for engine processes to finish...")
 	engineWait()
 	logger.Debug("Everything has finished, exiting k6!")
+	if test.keywriter != nil {
+		if err := test.keywriter.Close(); err != nil {
+			logger.WithError(err).Warn("error while closing the SSLKEYLOGFILE")
```
na--, line 273:
```suggestion
			logger.WithError(err).Warn("Error while closing the SSLKEYLOGFILE")
```
I think the convention is that log messages are capitalized, errors are not.

cmd/test_load.go
MStoykov: Hmm maybe - this is from some of the previous PRs - haven't really looked at most of this messages :thinking:

cmd/runtime_options.go
```diff
@@ -117,6 +117,12 @@ func getRuntimeOptions(flags *pflag.FlagSet, environment map[string]string) (lib
 		}
 	}
 
+	if envVar, ok := environment["SSLKEYLOGFILE"]; ok {
```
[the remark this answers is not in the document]
MStoykov: I don't see why we should have two for this as well. neithe firefox, chrome or curl decided to prefix theirs or add ones with a prefix. We can also always do it later.

MStoykov:
afero is striking again with it's strange bugs :facepalm: I literally couldn't make `OpenFile` create a file with `fs.OsFs` with no version I tried and I htink that MemMapFs got fixed in one the very last releases :(. 

Moving tlsconfig out of js.Runner seemed like way too big of a yak shave to start on it - given the amount of refactoring I already needed to do to make this possible at all -so for now i just skip tls checking.

cmd/test_load.go
MStoykov: And yeah - I will make it a warning

cmd/test_load.go
```diff
@@ -202,3 +224,31 @@ func (lt *loadedTest) consolidateDeriveAndValidateConfig(
...
+}
+
+func (cw *syncWriter) Write(b []byte) (int, error) {
+	cw.m.Lock()
+	defer cw.m.Unlock()
+	return cw.w.Write(b)
+}
+
+func (lt *loadedTest) createFileForAppend(filename string) (io.WriteCloser, error) {
+	fs := lt.fileSystems["file"]
```
na--, line 240: I think this is a bug, you should use the `globalState.fs` that is passed to `loadAndConfigureTest()` here, not `lt.fileSystems`. Right now, `SSLKEYLOGFILE=./test.log k6 cloud script.js` will include `test.log` in the archive bundle

cmd/test_load.go
```diff
@@ -202,3 +224,31 @@ func (lt *loadedTest) consolidateDeriveAndValidateConfig(
...
+	if ok, err := afero.Exists(fs, filename); !ok {
+		// This whole parts is because afero has problems creating files with OpenFile.
```
na--, line 242:
```suggestion
		// This whole part is because afero has problems creating files with OpenFile.
```

na-- (commented):
LGTM besides https://github.com/grafana/k6/pull/2487#pullrequestreview-936203862

cmd/test_load.go
MStoykov: This seems to be because I was using `filesystems` ... hopefully this will fix the windows failing tests as well :crossed_fingers:

cmd/test_load.go
MStoykov: fixed in [47c41d7](https://github.com/grafana/k6/pull/2487/commits/47c41d7977978c0330333854006dd0d2453ad7b2)

MStoykov:
I also made this small unrelated change [b710000](https://github.com/grafana/k6/pull/2487/commits/b710000dbf79b159b1a81804ec6004dc00a92e36) ... kind of wonder if I should make the `require` an `assert` as currently when something fails - there are no logs .. Or maybe by default it can also dump at least the `stderr` :shrug:

na-- (approved):
LGTM! :tada: :heart:

oleiade (approved):
Tested and approved on Windows! Great job, and super useful indeed! 🪟 🎉

kkriegkxs:
Good stuff! Thanks @MStoykov 
Also glad you went with the standard env var ;)

Issue #2450, opened 2022-03-17
k6/execution: Export consolidated options


codebien:
As a branch of https://github.com/grafana/k6/issues/2259 issue and as a useful utility for having a more consistent behaviour for #796 and some other issues, we want export a `test.options` object as part of the `k6/execution` API.

The common exported config object `export const options = { ... }` has some lack, it hasn't the capability to get accurate values within the VU's code like the following example:

```js
export default function() {
  console.log(options.tags.mytag); // it prints undefined
}
```
The _why_ is extensively explained by [this comment](https://github.com/grafana/k6/issues/2278#issuecomment-984564468):

>The issue is that options get overwritten after they get merged with everything else (cli flags, env and so on). This overwriting though for some reason is really complicated and in this particular case options.tags is just set to its internal type that is a wrapper around a map, not just a map. As such it does not have properties such as mytag.

### Proposal

```js
import 'exec' from 'k6/execution';

export default function() {
  console.log(exec.test.options.tags.mytag);
}
```
* The exec code should export and coverts in a nice read-only `goja.Object` the available consolidated and derived config from the `lib.State.Options`.
* A deprecation warning should be added when a setting operation on the `options` object is executed. (Maybe also on getting?)

PR #2492, opened 2022-04-12
Restore `options.scenarios`

codebien:
Using as an example the following script, it prints an unexpected `{}` object using `>v0.37.0`. This change has been introduced by https://github.com/grafana/k6/pull/2392.

```javascript

export const options = {
	scenarios: {
		contacts: {
			iterations: 1,
		},
	},
};

export default function() {
	console.log(JSON.stringify(options.scenarios))
}
```

In v0.36.0 and previous versions, the `scenarios` key wasn't set during the `js.Bundle.Instantiate` because the key used for it was `scenarios,omitempty`.

The current change is an hack for restoring the previous behaviour where the expected output is:

```javascript
INFO[0000] {"contacts":{"executor":"shared-iterations","iterations":1}}  source=console
``` 

### Questions

* Could we rid the entire `b.Options.ForEachSpecified("json",` logic when https://github.com/grafana/k6/issues/2450 will be implemented. If yes, probably we should directly implement 2450 and not merge this.
* Do we have an alternative to make this change less const-string dependent?

### DIFF
--- a/js/bundle.go
+++ b/js/bundle.go
@@ -289,6 +289,13 @@ func (b *Bundle) Instantiate(
 		jsOptionsObj = jsOptions.ToObject(rt)
 	}
 	b.Options.ForEachSpecified("json", func(key string, val interface{}) {
+		// goja.ToValue generates an empty Object `{}` for the scenarios's value.
+		// This skips the set preserving the original `options.scenarios` value.
+		//
+		// TODO: rid this part when #883 will be implemented
+		if key == "scenarios" {
+			return
+		}
 		if err := jsOptionsObj.Set(key, val); err != nil {
 			instErr = err
 		}


MStoykov:
This unfortunately means that we won't set the `scenarios` in the script at all, so if they are not entirely configured from within they will have wrong values ... 

It seems to me like the `Set` doesn't work correctly and it seems to be some goja issue - probably due to https://github.com/dop251/goja/issues/377 (the type definition).

js/bundle.go
```diff
@@ -289,6 +289,13 @@ func (b *Bundle) Instantiate(
 		jsOptionsObj = jsOptions.ToObject(rt)
 	}
 	b.Options.ForEachSpecified("json", func(key string, val interface{}) {
+		// goja.ToValue generates an empty Object `{}` for the scenarios's value.
+		// This skips the set preserving the original `options.scenarios` value.
+		//
+		// TODO: rid this part when #883 will be implemented
+		if key == "scenarios" {
+			return
```
MStoykov, line 297:
```suggestion
			val = (map[string]lib.ExecutorConfig)(val.(lib.ScenarioConfigs))
```

This seems like a better fix.
We basically remove the type "around" map and give goja a map which then is set correctly.

This definitely needs a bunch of tests though

MStoykov: Probably a better variant will be for this to have some kind of interface `ToValue` or something like that. Then we can check that `val` implements it and call the method to get the actual value - the body in this will be this same line. 

But I feel that this might be too much work especially as this will likely be the only "exception" so maybe leave that as a  comment and hope that we will never have to implement it :crossed_fingers:

codebien:
Considering the effort required for this, we would directly introduce the final and better solution. I opened an issue for tracking and future updates. https://github.com/grafana/k6/issues/2495

js/bundle.go
codebien: Unfortunately, this doesn't work really well because it includes the BaseConfig that the executor's config contains. The ideal approach would be to use a mapper like ToValue to return the correct object.

PR #2493, opened 2022-04-13
js/k6/exec: test.options object

codebien:
This PR adds the capability for fetching the consolidated and derived options for tests using the `k6/execution` API. It adds a `test.options` object exporting all the potential set properties.

A simple example of the expected API:

```javascript

export const options = {
  vus: 10,
  duration : '30s'
};

export default function() {
  console.log(exec.test.options.scenarios.default.vus) // 10
}
```

### Open questions

1. ~We are defining the default values for some objects/arrays as `null`. An example is [Thresholds](https://k6.io/docs/using-k6/options/#thresholds).
Which value do we expect to get if we access properties like Thresholds? `undefined` or `null` or `{}` (empty object)?~ We will use the same value returned from the JSON marshal, at the moment, the k6 options' values are Nullable so the default value will be `null`.

```javascript

export const options = {
  vus: 10,
  duration : '30s'
};

export default function() {
  console.log(exec.test.options.thresholds) // what do you expect to get?
}
```

2. This point from the original issue is missing:
>A deprecation warning should be added when a set operation on the options object is executed. (Maybe also on getting?)

~~In case of `const`, it seems not possible by spec to set back an eventual wrap/proxy for warning the access. Do we have an alternative?~~ We used a classic JS Object but with the Object.freeze function applied. 

I would define these cases before polishing and addressing edge cases.

Closes https://github.com/grafana/k6/issues/2450

### DIFF
--- a/js/common/frozen_object.go
+++ b/js/common/frozen_object.go
@@ -1,0 +1,69 @@
+package common
+
+import (
+	"fmt"
+
+	"github.com/dop251/goja"
+)
+
+// FreezeObject replicates the JavaScript Object.freeze function.
+func FreezeObject(rt *goja.Runtime, obj goja.Value) error {
+	global := rt.GlobalObject().Get("Object").ToObject(rt)
+	freeze, ok := goja.AssertFunction(global.Get("freeze"))
+	if !ok {
+		panic("failed to get the Object.freeze function from the runtime")
+	}
+	isFrozen, ok := goja.AssertFunction(global.Get("isFrozen"))
+	if !ok {
+		panic("failed to get the Object.isFrozen function from the runtime")
+	}
+	fobj := &freezing{
+		global:   global,
+		rt:       rt,
+		freeze:   freeze,
+		isFrozen: isFrozen,
+	}
+	return fobj.deepFreeze(obj)
+}
+
+type freezing struct {
+	rt       *goja.Runtime
+	global   goja.Value
+	freeze   goja.Callable
+	isFrozen goja.Callable
+}
+
+func (f *freezing) deepFreeze(val goja.Value) error {
+	if val != nil && goja.IsNull(val) {
+		return nil
+	}
+
+	_, err := f.freeze(goja.Undefined(), val)
+	if err != nil {
+		return fmt.Errorf("object freeze failed: %w", err)
+	}
+
+	o := val.ToObject(f.rt)
+	if o == nil {
+		return nil
+	}
+
+	for _, key := range o.Keys() {
+		prop := o.Get(key)
+		if prop == nil {
+			continue
+		}
+		frozen, err := f.isFrozen(goja.Undefined(), prop)
+		if err != nil {
+			return err
+		}
+		if frozen.ToBoolean() { // prevent cycles
+			continue
+		}
+		if err = f.deepFreeze(prop); err != nil {
+			return fmt.Errorf("deep freezing the property %s failed: %w", key, err)
+		}
+	}
+
+	return nil
+}
--- a/js/modules/k6/execution/execution.go
+++ b/js/modules/k6/execution/execution.go
@@ -21,6 +21,7 @@
 package execution
 
 import (
+	"encoding/json"
 	"errors"
 	"fmt"
 	"reflect"
@@ -167,6 +168,9 @@ func (mi *ModuleInstance) newInstanceInfo() (*goja.Object, error) {
 // newTestInfo returns a goja.Object with property accessors to retrieve
 // information and control execution of the overall test run.
 func (mi *ModuleInstance) newTestInfo() (*goja.Object, error) {
+	// the cache of goja.Object in the optimal parsed form
+	// for the consolidated and derived lib.Options
+	var optionsObject *goja.Object
 	rt := mi.vu.Runtime()
 	ti := map[string]func() interface{}{
 		// stop the test run
@@ -178,6 +182,16 @@ func (mi *ModuleInstance) newTestInfo() (*goja.Object, error) {
 				}
 				rt.Interrupt(&common.InterruptError{Reason: reason})
 			}
+		},
+		"options": func() interface{} {
+			if optionsObject == nil {
+				opts, err := optionsAsObject(rt, mi.vu.State().Options)
+				if err != nil {
+					common.Throw(rt, err)
+				}
+				optionsObject = opts
+			}
+			return optionsObject
 		},
 	}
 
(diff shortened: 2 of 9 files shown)


js/modules/k6/execution/execution.go
```diff
@@ -227,6 +241,47 @@ func newInfoObj(rt *goja.Runtime, props map[string]func() interface{}) (*goja.Ob
...
 
+// optionsAsObject maps the lib.Options struct that contains the consolidated
+// and deriverd options configuration in a goja.Object.
+func optionsAsObject(rt *goja.Runtime, options lib.Options) (*goja.Object, error) {
+	// TODO: refactor removing the json part
+	// and replace it with a static data mapping,
+	// potentially using a generator.
+	b, err := json.Marshal(options)
+	if err != nil {
+		return nil, fmt.Errorf("failed to encode the lib.Options as json: %w", err)
```
MStoykov, line 252: I kind of prefer the `json.Marshal` as it guarantees that we decouple the actual options and what we return.

MStoykov (commented):
Have you thought about freezing the object that is returned so that users will get an error if they try to modify something?

codebien:
>Have you thought about freezing the object that is returned so that users will get an error if they try to modify something?

@MStoykov it would add a breaking change, if we have to do it I think could be good to give to the users at least one release before doing it, so they would have the time to migrate. WDYT?

MStoykov:
I meant the `exec.test.options` ... which a completely new API - so no breaking changes ... or am I missing something?

codebien:
Oh, you're right. :man_facepalming:  I thought you meant the exported `options` object.

js/modules/k6/execution/execution.go
```diff
@@ -227,6 +241,48 @@ func newInfoObj(rt *goja.Runtime, props map[string]func() interface{}) (*goja.Ob
...
+		}
+	}
+
+	for kprop, vprop := range props {
+		mustSetReadOnlyProperty(kprop, vprop)
+	}
+
+	mustSetReadOnlyProperty("consoleOutput", options.ConsoleOutput)
+
+	// TODO: implement the MarshalText function
```
olegbespalov, line 280:
```suggestion
	// TODO: expose after implementing a `NullIPPool.MarshalText` function
```
I believe that it nice to mention where we have to implement the `MarshalText`

And maybe just implement it in a separate commit? :smile:

olegbespalov (commented):
Overall looks good, left a small comment :+1:

js/modules/k6/execution/execution.go
codebien: The latest commit includes the NullIPPool MarshalText

js/modules/k6/execution/execution.go
codebien: I removed the TODO

codebien:
>Have you thought about freezing the object that is returned so that users will get an error if they try to modify something?

The `goja.Object` doesn't provide a freezing method from Go or at least I didn't find it, so I added the properties to the JS object as read-only.

I also added an integration test for monitoring the default values. Unfortunately, they will be `null` even in the case where we have set a value. This happens because we are coding nullable types as `valid: false` when we set them for default values.

Here, an example:
https://github.com/grafana/k6/blob/57e994242f10122bccfbda229c4b29499a42457b/lib/executor/ramping_vus.go#L68-L74

Of course, they will be set when the user explicitly set them or the config is derived.

MStoykov:
>  doesn't provide a freezing method from Go 

It doesn't, but I managed to make it work for the original SharedArray implementation https://github.com/grafana/k6/pull/1739/files

codebien:
Added the object freeze:

```js
import exec from 'k6/execution'

export default function() {
	console.log(Object.isFrozen(exec.test.options))
	console.log(Object.isFrozen(exec.test.options.paused))
	console.log(Object.isFrozen(exec.test.options.httpDebug))
	console.log(Object.isFrozen(exec.test.options.dns))
	console.log(Object.isFrozen(exec.test.options.scenarios))
	console.log(Object.isFrozen(exec.test.options.scenarios.default))
	console.log(Object.isFrozen(exec.test.options.scenarios.default.iterations))
}
```

```js
INFO[0000] true                                          source=console
INFO[0000] true                                          source=console
INFO[0000] true                                          source=console
INFO[0000] true                                          source=console
INFO[0000] true                                          source=console
INFO[0000] true                                          source=console
INFO[0000] true                                          source=console
```

js/common/frozen_object.go
```diff
@@ -0,0 +1,61 @@
...
+	"fmt"
+
+	"github.com/dop251/goja"
+)
+
+// FreezeObject replicates the JavaScript Object.freeze function.
+func FreezeObject(rt *goja.Runtime, obj goja.Value) error {
+	global := rt.GlobalObject().Get("Object")
+	freeze, _ := goja.AssertFunction(global.ToObject(rt).Get("freeze"))
+	isFrozen, _ := goja.AssertFunction(rt.GlobalObject().Get("Object").ToObject(rt).Get("isFrozen"))
```
olegbespalov, line 13:
```suggestion
	isFrozen, _ := goja.AssertFunction(global.ToObject(rt).Get("isFrozen"))
```

js/common/frozen_object.go
```diff
@@ -0,0 +1,61 @@
...
+import (
+	"fmt"
+
+	"github.com/dop251/goja"
+)
+
+// FreezeObject replicates the JavaScript Object.freeze function.
+func FreezeObject(rt *goja.Runtime, obj goja.Value) error {
+	global := rt.GlobalObject().Get("Object")
+	freeze, _ := goja.AssertFunction(global.ToObject(rt).Get("freeze"))
```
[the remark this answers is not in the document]
olegbespalov: But is it the same as the `object freeze failed`? Ideally, we should really rarely use `panic`, no?

But again, I started this comment from the `Not sure if that is the case`, so it was more a discussion point, you can even drop this comment if you think that this is never-ever possible :)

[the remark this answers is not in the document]
codebien: I added the check for the `ok` value

js/common/frozen_object.go
```diff
@@ -0,0 +1,69 @@
...
+
+	"github.com/dop251/goja"
+)
+
+// FreezeObject replicates the JavaScript Object.freeze function.
+func FreezeObject(rt *goja.Runtime, obj goja.Value) error {
+	global := rt.GlobalObject().Get("Object").ToObject(rt)
+	freeze, ok := goja.AssertFunction(global.Get("freeze"))
+	if !ok {
+		panic(fmt.Errorf("failed to get the Object.freeze function from the runtime"))
```
olegbespalov, line 14: :see_no_evil: It's a 100% nitpick, but we don't need here formatting (`fmt.Errorf`) and maybe even an error when panicing

codebien: You're right. I'm sorry, this is a very basic error on my side :man_facepalming:

olegbespalov (dismissed):
:muscle:

js/common/frozen_object.go
olegbespalov: No worries at all, it's not a big deal :hugs:

js/common/frozen_object.go
```diff
@@ -0,0 +1,69 @@
+package common
+
+import (
+	"fmt"
+
+	"github.com/dop251/goja"
+)
+
+// FreezeObject replicates the JavaScript Object.freeze function.
+func FreezeObject(rt *goja.Runtime, obj goja.Value) error {
```
oleiade, line 10: This is AWESOME! 👍🏻

codebien:
>Regarding the default values of thresholds, I would prefer it to be an empty object {}, but if I console.log(options.thresholds) as of today, it prints undefined which makes sense. I can see both work, so up to you :sauropod:

@oleiade I updated the description with the decision applied for that part, sorry for not doing it before. Reporting here also :arrow_down: :

>We will use the same value returned from the JSON marshal, at the moment, the k6 options' values are Nullable so the default value will be `null`.

It is the same value also [reported in the docs](https://k6.io/docs/using-k6/options/#thresholds) at the moment as the default value so it shouldn't change the current UX.

MStoykov (approved):
LGTM

codebien:
I rebased the fixups and applied the @MStoykov's suggestion

MStoykov (approved):

olegbespalov (approved):

codebien:
This PR will implement the doc: https://github.com/grafana/k6-docs/pull/629

PR #2496, opened 2022-04-14
k6-dev-tool container

olegbespalov:
# What?

This PR introduces the possibility to build a `k6-dev-tools` container that contributors/maintainers can use during the development. The idea is to keep "locked" the versions of the tools there and use this for the common operations like running the linter, applying the auto-fixes, code generation, and so on (the origin from there https://github.com/grafana/k6/pull/2437#discussion_r850159605).

How to use it?

Build it with
```bash
make make build-k6-dev-tools
```

Then run either: 

```bash
make lint
```
or
```bash
make fix
```
or
```bash
make generate
```

# Why?

It improves the developer's UX, and I believe should help external maintainers.


javaducky:
Just a side comment about the `Makefile`. WDYT about adding a `make help` target which lists the targets with descriptions. And example is [here](https://github.com/grafana/xk6/blob/9001905aa9172b94b0cfd3ff75274d933eeeed9c/Makefile#L5-L13).

This would allow folks to see the available targets as the following:
```
$ make help
Usage: make <OPTIONS> ... <TARGETS>

Available targets are:

  help           Prints a list of available build targets.
  clean          Removes any previously created build artifacts.
  build          Builds the 'xk6' binary.
  build-docker
  format         Applies Go formatting to code.
  test           Executes any unit tests.
  vendor         Pulls source for external dependencies.

Targets run by default are: clean, format, test, and build
```

javaducky:
Not sure why the linting via Docker is such a significant time difference... 1.5m versus 11s.

```
~/Develop/k6  feat/docker-dev ✔                                                                                                                                2h31m  
▶ time make lint
make lint  0.06s user 0.06s system 0% cpu 1:30.46 total

~/Develop/k6  feat/docker-dev ✔                                                                                                                                2h32m  
▶ time golangci-lint run --fix --new-from-rev master ./...
golangci-lint run --fix --new-from-rev master ./...  15.86s user 1.31s system 151% cpu 11.336 total
```

olegbespalov:
@javaducky yeah, `make help` is a nice thing, added

>Not sure why the linting via Docker is such a significant time difference... 1.5m versus 11s.

I guess it's related to the cache (the previous implementation has it). I return now should be better (after the first run).

Makefile
```diff
@@ -1,30 +1,73 @@
...
+format:
+	$(call run_k6_tools,gofumpt -w .)
 
-ci-like-lint :
-	@docker run --rm -t -v $(shell pwd):/app \
-		-v $(TMPDIR)/golangci-cache-$(GOLANGCI_LINT_VERSION):/golangci-cache \
-		--env "GOLANGCI_LINT_CACHE=/golangci-cache" \
-		-w /app golangci/golangci-lint:$(GOLANGCI_LINT_VERSION) \
-		make lint
+## fix: Runs golangci-lint with the verson that used inside the CI.
```
javaducky, line 44: Copy/paste inheritance 😉.  Should substitute "fix" with "lint" to match the target name.

olegbespalov: :facepalm:

javaducky:
LGTM if there are no objections from others with the switch to Docker tooling. 👍

codebien (approved):
LGTM

na-- (dismissed):

Dockerfile.dev
```diff
@@ -0,0 +1,19 @@
+FROM golang:1.17
```
MStoykov, line 1: This image is 1.31GB ... and cleaning the cache doesn't help.

 It seems that it is due to the go install which has the stdlib precomipiled a bunch.

Do we really need it once the tools below are build? Can we maybe copy it to a different, smaller image. All the tools below are under 100mb.

olegbespalov: Well, to be honest, we don't need all tools installed, but to be honest, the size of the tools container is not so important. It's not the image that pulls in the CI during the deployment, right? 

I'd say that the more important is the maintainability. So fewer instructions it has is better :man_shrugging: 

Not sure that here applies the same approach as we do in the main image since, for example, for the lining, we need the binary and go installed. But let me try to iterate on this. Maybe we can decrease it :thinking:

olegbespalov: So, migrated to the alpine (it's also consistent with the main `Dockerfile`), the size is decreased but still significant. So maybe we can iterate on it in the following PRs? In any case, I hope that this will be a team's effort so it can be constantly improved. WDYT?

```bash
docker image ls | grep k6-dev-tools
k6-dev-tools                      latest          0787216b5245   About an hour ago   888MB
```

na-- (changes requested):
Actually, I am not so sure if this is as beneficial as I initially thought :thinking: I imagine all of us have (and any prospective k6 developer will have) Go installed on our systems, so I don't see the benefit of lugging a 800+MB docker image around just to run lint and `go generate`... I'd definitely prefer to install the correct `enumer`, `golangci-lint`, etc. versions directly on my machine with `go install tool@version`...

So :-1: for packaging this in docker. :+1: for having a `make` target that installs all of the correct versions.

olegbespalov:
@na-- I see your point, but the idea is to help all maintainers/contributors (maybe even the external are the first audience). I thought that docker, in that case, was the best approach. 

If the only size is the case, let me try to iterate on that. I believe there are still possible improvements without decreasing Dockerfile's maintainability.

PR #2503, opened 2022-04-21
Set tlsAuth.Password as a nullable string

codebien:
Optional options should be nullable using a `guregu/null.`-like type for supporting consistently the `exec.test.options`.
Moved the `tlsAuth.password` option to null.String


oleiade (approved):
Looking good 👍🏻 👀

na-- (approved):

PR #2504, opened 2022-04-23
Update instructions for building from source

JamieEdge:
Updated the instructions for building from source to reflect the following:

* Git should not be required where modules are proxied via `proxy.golang.org`, which is the default behaviour of Go.
* Go installation instructions are now at `go.dev` instead of `golang.org`.
* Go 1.17 is the minimum required version for k6 as specified in `go.mod`.
* `go install` requires a version to be specified.
* `go install` does not clone sources to `$GOPATH/src`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,11 +116,9 @@ In case there isn't an official package for your operating system or architectur
 
 ### Build from source
 
-k6 is written in Go, so it's just a single statically-linked executable and very easy to build and distribute. To build from source you need **[Git](https://git-scm.com/downloads)** and **[Go](https://golang.org/doc/install)** (1.16 or newer). Follow these instructions:
+k6 is written in Go, so it's just a single statically-linked executable and very easy to build and distribute. To build from source you need **[Go](https://go.dev/doc/install)** (1.17 or newer). Follow these instructions:
 
-- Run `go install go.k6.io/k6` which will:
-  - git clone the repo and put the source in `$GOPATH/src/go.k6.io/k6`
-  - build a `k6` binary and put it in `$GOPATH/bin`
+- Run `go install go.k6.io/k6@latest` which will build a `k6` binary and put it in `$GOPATH/bin`.
 - Make sure you have `$GOPATH/bin` in your `PATH` (or copy the `k6` binary somewhere in your `PATH`), so you are able to run k6 from any location.
 - Tada, you can now run k6 using `k6 run script.js`
 
--- a/README.md
+++ b/README.md
@@ -118,7 +118,7 @@ In case there isn't an official package for your operating system or architectur
 
 k6 is written in Go, so it's just a single statically-linked executable and very easy to build and distribute. To build from source you need **[Go](https://go.dev/doc/install)** (1.17 or newer). Follow these instructions:
 
-- Run `go install go.k6.io/k6@latest` which will build a `k6` binary and put it in `$GOPATH/bin`.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -116,11 +116,9 @@ In case there isn't an official package for your operating system or architectur
 
 ### Build from source
 
-k6 is written in Go, so it's just a single statically-linked executable and very easy to build and distribute. To build from source you need **[Git](https://git-scm.com/downloads)** and **[Go](https://golang.org/doc/install)** (1.16 or newer). Follow these instructions:
+k6 is written in Go, so it's just a single statically-linked executable and very easy to build and distribute. To build from source you need **[Go](https://go.dev/doc/install)** (1.17 or newer). Follow these instructions:
 
-- Run `go install go.k6.io/k6` which will:
-  - git clone the repo and put the source in `$GOPATH/src/go.k6.io/k6`
-  - build a `k6` binary and put it in `$GOPATH/bin`
+- Run `go install go.k6.io/k6@latest` which will build a `k6` binary and put it in `$GOPATH/bin`.
```
oleiade, line 121:
```suggestion
- Run `go install go.k6.io/k6@latest` which will build a `k6` binary and put it in the `$GOBIN` folder (which defaults to `$GOPATH/bin`).
```

To be more accurate, the[ Go documentation](https://go.dev/ref/mod#go-install) states that the binary will placed in the `GOBIN` directory(which defaults to `$GOPATH/bin` or `$HOME/go/bin` if the `GOPATH` environment variable is not set).

oleiade: On another note, as detailed in the [Go documentation](https://go.dev/ref/mod#go-install), explicitly specifying `@latest` will override the `go.mod` if present. In the context of building k6 from source, I think I would prefer the documented `go install` command to not specify a version specifically (and to not override the `go.mod`) as the main scenario.

oleiade (changes requested):
Thanks a lot for your contribution, and this well-deserved refresher of our README 👏🏻  I've left some comments, which, I believe, would improve the build from source instructions even further. Cheers 🎉

README.md
JamieEdge: > To be more accurate, the[ Go documentation](https://go.dev/ref/mod#go-install) states that the binary will placed in the `GOBIN` directory(which defaults to `$GOPATH/bin` or `$HOME/go/bin` if the `GOPATH` environment variable is not set).

Good point. I use `GOBIN` myself but wasn't sure if this was too much information when originally updating the instructions. I've added the suggested changes now.

JamieEdge: > On another note, as detailed in the [Go documentation](https://go.dev/ref/mod#go-install), explicitly specifying `@latest` will override the `go.mod` if present. In the context of building k6 from source, I think I would prefer the documented `go install` command to not specify a version specifically (and to not override the `go.mod`) as the main scenario.

I understand that it is fairly common for those who use Go frequently to install tools written in Go using `go install ...@latest` outside of a module. Personally, this is how I install all of the tools I use when working on Go projects.

Would it make sense for the instructions to be split into 'Installing using Go' and 'Building from Source'? The former section would have only the `go install ...@latest` command, whereas the latter would explain using `go build .` within a cloned repository.

I think that using `go install` without a version specified would only be of use if you were inside of a module which added k6 as a dependency. This scenario seems quite rare given that k6 is a command line tool and wouldn't be compiled into projects that it tests.

When installing from source, it could be argued that fetching k6 and all of its dependencies via the default module proxy is preferable over cloning the repository and running a local build (e.g. version information is embedded into the binary).

na-- (approved):

Issue #2520, opened 2022-05-06
Thresholds over checks results in `NaN`


oleiade:
### Brief summary

We discovered a regression in how k6 versions >= `v0.38.0` handled thresholds. It was brought to us that following[ the technique consisting in defining thresholds over checks](https://k6.io/docs/using-k6/thresholds/#failing-a-load-test-using-checks) would result in a `NaN` result as described in the `Actual Behavior` section.

### k6 version

v0.38.1

### OS

Linux

### Steps to reproduce the problem

```
import { check, sleep } from 'k6';
import http from 'k6/http';
​
export const options = {
  scenarios: {
    iWillFail: {
      exec: 'iWillFail',
      executor: 'constant-vus',
      startTime: '2s',
      vus: 1,
      duration: '30s',
    },
  },
  thresholds: {
    'checks{type:read}': [{ threshold: 'rate>0.9', abortOnFail: true }],
  },
};
​
export function iWillFail() {
  let res = http.get(`https://test-api.k6.io/`);
  check(res, {
    'read status is 200': (r) => r.status === 200,
  }, { type: 'read' });
​
  sleep(1);
}
```

### Expected behaviour

We would expect the `checks{type:read}` output value to be a number, and not `NaN`.

### Actual behaviour

```
➜  tempo git:(main) ✗ ./k6 version
k6 v0.38.1 (2022-05-06T08:56:23+0000/44cc995, go1.17.9, linux/amd64)
➜  tempo git:(main) ✗ ./k6 run fails.js 
          /\      |‾‾| /‾‾/   /‾‾/   
     /\  /  \     |  |/  /   /  /    
    /  \/    \    |     (   /   ‾‾\  
   /          \   |  |\  \ |  (‾)  | 
  / __________ \  |__| \__\ \_____/ .io
  execution: local
     script: fails.js
     output: -
  scenarios: (100.00%) 1 scenario, 1 max VUs, 1m2s max duration (incl. graceful stop):
           * iWillFail: 1 looping VUs for 30s (exec: iWillFail, startTime: 2s, gracefulStop: 30s)
running (0m02.0s), 0/1 VUs, 0 complete and 0 interrupted iterations
iWillFail ✗ [--------------------------------------] 1 VUs  00.0s/30s
WARN[0002] No script iterations finished, consider making the test duration longer 
     ✗ { type:read }...: NaN% ✓ 0   ✗ 0  
     vus...............: 0    min=0 max=0
[...]

PR #2521, opened 2022-05-06
Fix threshold over rate metric resulting in `NaN` result

oleiade:
This Pull Request attempts to address #2520. It was discovered internally by the team that thresholds applied to checks (rate metrics) would immediately fail, and return a `NaN` result.

I've found out that the `NaN` was the result of a division by zero. I assumed it happens because the threshold was run before a single check sample was recorded (I might be wrong there though). I've thus introduced a check in `Thresholds.Run` to perform insert the sink's "rate" value computation only if the right-hand side of the division is not zero. 

Besides, the `Threshold.runNoTaint` function used to make a verification that a sink existed, before actually evaluating the threshold expression. This PR drops this verification, considering the thresholds are now verified at startup. Because of the previously mentioned issue, it would have led to a situation where we know the metric exists, we know the threshold is valid for this metric, but we fail with an error because at a certain point of the execution, our internal storage doesn't have data for it yet. 

I'm not a 100% confident this PR doesn't have unintended side-effects. I have tried it which a bunch of scripts of mine who try to exhibit as much of possible thresholds use cases, and none seem to be broken by this change. But I'd take any more insights if you have. The point of a hotfix being, after all, to fix stuff, not potentially introduce more problems 🤞🏻 

**Learning point**: good to know that `int64(0) / int64(0)` leads to a panic, but `float64(0) / float64(0)` leads to `NaN`.

### DIFF
--- a/core/engine.go
+++ b/core/engine.go
@@ -166,6 +166,11 @@ func (e *Engine) Init(globalCtx, runCtx context.Context) (run func() error, wait
 // test run status when it ends. It returns a function that can be used after
 // the provided context is called, to wait for the complete winding down of all
 // started goroutines.
+//
+// Because the background process is not aware of the execution's state, `processMetricsAfterRun`
+// will be used to signal that the test run is finished, no more metric samples will be produced,
+// and that the remaining metrics samples in the pipeline should be processed as the background
+// process is about to exit.
 func (e *Engine) startBackgroundProcesses(
 	globalCtx, runCtx context.Context, runResult <-chan error, runSubCancel func(), processMetricsAfterRun chan struct{},
 ) (wait func()) {
@@ -226,7 +231,7 @@ func (e *Engine) startBackgroundProcesses(
 			for {
 				select {
 				case <-ticker.C:
-					thresholdsTainted, shouldAbort := e.MetricsEngine.EvaluateThresholds()
+					thresholdsTainted, shouldAbort := e.MetricsEngine.EvaluateThresholds(true)
 					e.thresholdsTaintedLock.Lock()
 					e.thresholdsTainted = thresholdsTainted
 					e.thresholdsTaintedLock.Unlock()
@@ -244,6 +249,13 @@ func (e *Engine) startBackgroundProcesses(
 	return processes.Wait
 }
 
+// processMetrics process the execution's metrics samples as they are collected.
+// The processing of samples happens at a fixed rate defined by the `collectRate`
+// constant.
+//
+// The `processMetricsAfterRun` channel argument is used by the caller to signal
+// that the test run is finished, no more metric samples will be produced, and that
+// the metrics samples remaining in the pipeline should be should be processed.
 func (e *Engine) processMetrics(globalCtx context.Context, processMetricsAfterRun chan struct{}) {
 	sampleContainers := []metrics.SampleContainer{}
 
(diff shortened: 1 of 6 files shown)


na-- (dismissed):
Hmm I am not sure I fully grok what's going on here. 

From what I understand, this problem didn't appear before, in k6 v0.37.0 and below, because we basically only started running the threshold calculations for a metric once k6 saw a sample for the metric, is that right? And this PR kind of reverts that behavior, for `Rate` metrics at least? 

Because `0/0` is NaN, and we can't even pretend it's `0`, since that will be invalid as well? So it's best to just not run the threshold until there is some data. 

If I've understood things correctly, LGTM, though I don't know some of the internals here very well now.

codebien (approved):
The code seems fine to me

>Learning point: good to know that int64(0) / int64(0) leads to a panic, but float64(0) / float64(0) leads to NaN.

This is strange, they should be the same result, right? Is a type wrapped by `goja`? :thinking:

oleiade:
@codebien it's actually a behavior you can observe in pure Go: https://go.dev/play/p/BVaocakjKXD

oleiade:
@mstoykov opinions, and ideas to improve https://github.com/grafana/k6/pull/2521/commits/0116c5d82aba978378d4c36c50036a20e3b0dc5f ?

core/engine.go
```diff
@@ -299,7 +311,7 @@ func (e *Engine) processMetrics(globalCtx context.Context, processMetricsAfterRu
 			if !e.runtimeOptions.NoThresholds.Bool {
 				// Ensure the ingester flushes any buffered metrics
 				_ = e.ingester.Stop()
-				thresholdsTainted, _ := e.MetricsEngine.EvaluateThresholds()
+				thresholdsTainted, _ := e.MetricsEngine.EvaluateThresholds(true)
```
mstoykov, line 314: I think you continued my mistake with the boolean logic :rofl:. 

We don't should ignore empty sinks everywhere except here. Also my `IsEmpty` code is all wrong ... so you basically need to reverse to logic (or check for `== 0` instead of `>0` ).

 ```suggestion
				thresholdsTainted, _ := e.MetricsEngine.EvaluateThresholds(false)
```

File: core/engine.go [hunk 2]
```diff
@@ -226,7 +231,7 @@ func (e *Engine) startBackgroundProcesses(
 			for {
 				select {
 				case <-ticker.C:
-					thresholdsTainted, shouldAbort := e.MetricsEngine.EvaluateThresholds()
+					thresholdsTainted, shouldAbort := e.MetricsEngine.EvaluateThresholds(false)
```
mstoykov, line 234:
```suggestion
					thresholdsTainted, shouldAbort := e.MetricsEngine.EvaluateThresholds(true)
```

mstoykov (changes requested):
I think you continued my faulty boolean logic :sweat_smile: 

I've left what I think needs to change in order for the code to be correct. 

It will also be nice if we have tests for this as currently there are none :(.

oleiade:
@mstoykov After some more head-scratching and trying to figure out why the engine tests were failing, I think I've found out the issue. My boolean logic was actually correct, I just inverted the one you did in your initial patch. Your logic was additive: `Evalutate` or `EvaluateAll`, and I made it subtractive `Evaluate(` (all is the default), or `Evaluate(without)`. I realize now that I'd rather follow the positive/additive logic though, as it's the main path, and is simpler to understand. Going to modify the flag accordingly 👍🏻

oleiade:
I intend to add integration tests to cover the scenarios which triggered this specific issue separately in #2525.

oleiade:
@mstoykov could I get a thumbs up please? 🙇🏻

mstoykov (approved):

PR #2526, opened 2022-05-11
release v0.38.2

oleiade:
These are the release version bump and release notes for `v0.38.2` which addresses further issues we've discovered (thanks @efdknittlfrank thresholds in `v0.38.0` and `v0.38.1`.

### DIFF
--- a/release notes/v0.38.2.md
+++ b/release notes/v0.38.2.md
@@ -2,7 +2,7 @@ k6 v0.38.2 is a patch release containing a couple of bugfixes!
 
 ### Threshold over sub-metrics without samples would result in `NaN` ([#2520](https://github.com/grafana/k6/issues/2520))
 
-There was a bug in thresholds applied to sub-metrics set to `abortOnFail` that would lead to their result being `NaN` rather than a numerical value. The following script, for instance:
+There was a bug in thresholds applied to sub-metrics set to `abortOnFail`: leading k6 to evaluate thresholds that would have likely aborted before they had a chance of passing (because no samples for the given metric were recorded yet). This bug would have led to such thresholds' results value to be `NaN` rather than a numerical value. The following script, for instance:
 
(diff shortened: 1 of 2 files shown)


olegbespalov (approved):
LGTM :+1:

release notes/v0.38.2.md
```diff
@@ -0,0 +1,94 @@
+k6 v0.38.2 is a patch release containing a couple of bugfixes!
+
+### Threshold over sub-metrics without samples would result in `NaN` ([#2520](https://github.com/grafana/k6/issues/2520))
+
+There was a bug in thresholds applied to sub-metrics set to `abortOnFail` that would lead to their result being `NaN` rather than a numerical value. The following script, for instance:
```
mstoykov, line 5: I would argue the predominant problem was that we evaluate thresholds which will likely abort the test before they have any chance of actually passing such as `rate > 0.9` when no samples for the given metric have been emitted. 

The `NaN` was a specific case and it's mostly relevant as teh fix in v0.38.0 was supposed to make certain we will evaluate thresholds for which no metrics are emitted ... at least *once*. 

I would argue we should probably print a warning about such thresholds - once with no metrics as just because they pass doesn't mean that everything works correctly - maybe the user missconfigured something. But that is definitely an issue for v0.39.0 and beyond

mstoykov (dismissed):
I do think the main issue should be reworded not around the `NaN` case but the general one where we will evaluate metrics way too early. But I am approving this in case we want/need to release as this can be fixed later

release notes/v0.38.2.md
oleiade: Fair point! I would still keep the emanation of the issue as being the `NaN` appearance as it might be the error that drives users to look for a solution, but I'll add the underlying reasoning then 👍🏻

PR #2530, opened 2022-05-12
Fix printing wrong message on remote resolution

mstoykov:
Previously we will always print the really long message that says we
tried everything. But in reality we only do this if we have something
that:
* is not local (start with ./)
* is not prefixed with `https://`
* is not a magical github or cdnjs URL (which entirely uses the Opaque
  URL property)

But the if changed here basically checks that either we don't have
scheme *or* that it's Opaque. But in reality if it has scheme it never
has Opaque. And Opaque always has no scheme - so all cases enter the if.

The correct is to enter *only* when both are true as in that case it's
not magical and it doesn't have scheme so the schem

### DIFF
--- a/loader/loader.go
+++ b/loader/loader.go
@@ -236,7 +236,7 @@ func Load(
 			return result, nil
 		}
 
-		if moduleSpecifier.Scheme == "" || moduleSpecifier.Opaque == "" {
+		if moduleSpecifier.Scheme == "" && moduleSpecifier.Opaque == "" {
 			// we have an error and we did remote module resolution without a scheme
 			// let's write the coolest error message to try to help the lost soul who got to here
 			return nil, noSchemeRemoteModuleResolutionError{err: err, moduleSpecifier: originalModuleSpecifier}


olegbespalov (approved):

codebien:
>And Opaque always has no scheme - so all cases enter the if.

@mstoykov Apparently, the message mentions the `nodejs` case, where for example seems `node:fs` would be resolved with a scheme and opaque.

Example url resolution with Go:
```go
u, err := url.Parse("node:fs")
fmt.Println(u.Scheme) // node
fmt.Println(u.Opaque) // fs
```

Is the logic before the `if` changing it making the scheme empty?

It seems the case is not covered by the tests, can we add it? It should help with _documenting_ the case.

mstoykov:
The message predates nodejs introducing the `node` scheme and is still printed in those cases ;) 

We don't really support anything but `https` and `file` as a scheme either way. But in this particular case it just sidestep this and gets in a totally different [part of the code returning the same error](https://github.com/grafana/k6/blob/5f16f6d1feab6e3d9b6f3f7c856a8a7b1781c5a7/loader/loader.go#L125-L127). 

Given that this still works I really don't want to touch this more as this code is convoluted enough. And until we decide to deprecate at least the `https://` prefixing of everything it will continue to be so. If we decide to completely drop it (and especially if we also drop the special "magical" urls) I guess we will be able to in practice delete 80+% of the code that deals with those.

Even this fix in practice just makes a fairly rare case a bit less confusing, but if it required any more than this change I would probably just drop it.

codebien (approved):

PR #2532, opened 2022-05-12
Try to respect zero target stages at the end

### LINKED ISSUE #2531: Ramping vus/stages drops 0 target stages at the end.
### Brief summary

Related to #2527. 

If you have stages at the end with target 0, they will be dropped
```
export let options = {
  stages: [
    {duration: "1s", target 2},
    {duration: "1s", target 0},
    {duration: "1h", target 0},
  ]
}
```
will run for only 2s instead of 1h and 2s. 

The usefulness of this behaviour is questionable but it's somewhat strange that we have stages (even with target 0) and we do not take them into account. Arguably this should at least print a warning or something.

The main usecase I can find is like in the case of #2527 trying to make certain that `vus` and `vus_max` gets zeroed out before the test exits.

### k6 version

v0.38.2 and all before AFAIK

### OS

all

_No response_

Run a script which has more than one sequential stage with target 0 at the end like
```
export let options = {
  stages: [
    {duration: "1s", target 2},
    {duration: "1s", target 0},
    {duration: "1h", target 0},
  ]
}
```

### Expected behaviour

The test runs *at least* for the sum of all the stages durations

### Actual behaviour

It stops after the first stage with target 0 at the end

mstoykov:
Previous to this they were just ignored. Now ramping-vus will just not
do anything for that time.

closes #2531

### DIFF
--- a/lib/executor/ramping_vus.go
+++ b/lib/executor/ramping_vus.go
@@ -458,7 +458,7 @@ func (vlvc RampingVUsConfig) GetExecutionRequirements(et *lib.ExecutionTuple) []
 	}
 
 	// add one step for the end of the gracefulStop
-	if steps[len(steps)-1].TimeOffset != executorEndOffset {
+	if steps[len(steps)-1].PlannedVUs != 0 || steps[len(steps)-1].TimeOffset != executorEndOffset {
 		steps = append(steps, lib.ExecutionStep{TimeOffset: executorEndOffset, PlannedVUs: 0})
 	}
 
--- a/lib/executor/ramping_vus.go
+++ b/lib/executor/ramping_vus.go
@@ -243,8 +243,7 @@ func (vlvc RampingVUsConfig) getRawExecutionSteps(et *lib.ExecutionTuple, zeroEn
 		fromVUs = stageEndVUs
 	}
 
-	if zeroEnd && steps[len(steps)-1].PlannedVUs != 0 {
-		// If the last PlannedVUs value wasn't 0, add a last step with 0
+	if zeroEnd {
 		steps = append(steps, lib.ExecutionStep{TimeOffset: timeTillEnd, PlannedVUs: 0})
 	}
 	return steps
@@ -458,8 +457,8 @@ func (vlvc RampingVUsConfig) GetExecutionRequirements(et *lib.ExecutionTuple) []
 		steps = vlvc.reserveVUsForGracefulRampDowns(steps, executorEndOffset)
 	}
 
-	// If the last PlannedVUs value wasn't 0, add a last step with 0
-	if steps[len(steps)-1].PlannedVUs != 0 {
+	// add one step for the end of the gracefulStop
+	if steps[len(steps)-1].TimeOffset != executorEndOffset {
 		steps = append(steps, lib.ExecutionStep{TimeOffset: executorEndOffset, PlannedVUs: 0})
 	}
 
--- a/lib/executor/ramping_vus_test.go
+++ b/lib/executor/ramping_vus_test.go
@@ -560,6 +560,8 @@ func TestRampingVUsConfigExecutionPlanExampleOneThird(t *testing.T) {
 		{TimeOffset: 16 * time.Second, PlannedVUs: 1},
 		{TimeOffset: 20 * time.Second, PlannedVUs: 0},
 	}
+	expRawStepsZeroEnd := expRawStepsNoZeroEnd // no need to copy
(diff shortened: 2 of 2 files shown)


oleiade (dismissed):
LGTM 👍🏻 Playing with it locally, it seems to work as expected indeed 🤝

lib/executor/ramping_vus.go
```diff
@@ -458,8 +457,8 @@ func (vlvc RampingVUsConfig) GetExecutionRequirements(et *lib.ExecutionTuple) []
 		steps = vlvc.reserveVUsForGracefulRampDowns(steps, executorEndOffset)
 	}
 
-	// If the last PlannedVUs value wasn't 0, add a last step with 0
-	if steps[len(steps)-1].PlannedVUs != 0 {
+	// add one step for the end of the gracefulStop
+	if steps[len(steps)-1].TimeOffset != executorEndOffset {
```
na--, line 461: Hmm I am not sure this check is correct now :thinking: Shouldn't we always add this last step with 0 VUs here? That is, I think we should remove the `if` and just `append()` the step always :thinking:

na--: To see how this fails, try running this valid script:
```js
import http from 'k6/http';
import { sleep } from 'k6';

export const options = {
    scenarios: {
        test: {
            executor: 'ramping-vus',
            startVUs: 0,
            stages: [
                { duration: '2s', target: 2 },
            ],
            gracefulRampDown: '0s',
            gracefulStop: '0s',
        },
    },
};

export default function () {
    console.log(`[t=${(new Date()) - exec.scenario.startTime}ms] VU#${exec.vu.idInTest} ran scenario iteration ${exec.scenario.iterationInTest} (${exec.vu.iterationInScenario} for VU)`);
    sleep(1);
}
```

You'll get this with your branch:
```
ERRO[0000] Executor error                                error="test expected graceful end offset at 2s to be final" executor=test startTime=0s type=ramping-vus
ERRO[0000] test expected graceful end offset at 2s to be final 
exit status 103
```

that is, this check gets triggered: https://github.com/grafana/k6/blob/27f2ead359dac24b4131dac4b57c95c058f06e48/lib/executor/ramping_vus.go#L516

na--: Mind you, while I am pretty sure this code is wrong, I am far from sure my proposal is the correct one... :sweat_smile:

mstoykov: I am pretty sure that removing this `if` broke something else :rofl: I just don't remember what currently. But I did just try to add more steps with the motto of "it can't be worse" and it turned out it was breaking stuff ... similar to the breakage you are showing :(

na--: Quite possible... :disappointed: I'll try to debug this some more tomorrow, it's taking me a bit of time to grok what exactly was the logic behind everything here... :sweat_smile:

na--: @mstoykov, can you try to remember what gets broken when the `if` got removed, since I just tried it and the tests seem to pass fine :confused: 

Alternatively, I think we can use the following `if` instead:
```suggestion
	if steps[len(steps)-1].PlannedVUs != 0 || steps[len(steps)-1].TimeOffset != executorEndOffset {
```

The tests also pass with it too and the logic is that if one of the two conditions for the final step is not satisfied, we add a new final step after it the expected time offset and 0 value.

na--: So yeah, my change suggestion would be to fix the `if` in the way :arrow_up: and to add a `ramping-vus` scenario with this config to the tests:
```js
{
    // ...
    stages: [
        { duration: '2s', target: 2 },
    ],
    gracefulRampDown: '0s',
    gracefulStop: '0s',
}
```

mstoykov: Can't find anything wrong with removing it as a whole as well. Maybe I tested with some other change  as well or am misremembering

na--: :+1: 

Still, I think I slightly prefer having the `if` there, just with both the `PlannedVUs` and `TimeOffset` conditions. Without the `if`, the last step could be repeated twice, which _shouldn't_ be a problem, but it also shouldn't be necessary :man_shrugging:

mstoykov: I will add a test case for the script you posted above and expand the `if`

na-- (approved):
:crossed_fingers:

oleiade (approved):
🤝

na--:
Reopened by https://github.com/grafana/k6/pull/2575#issuecomment-1160065925

Issue #2535, opened 2022-05-13
Make certain `./lib` doesn't require goja


mstoykov:
./lib should not depend on goja. I thought we fixed that with the metrics and thresholds rewrite but it turns out we did add one more dependency a year ago in the form of https://github.com/grafana/k6/blob/5c357289fcc7a717a9b39990bd975ea2784b42bc/lib/executor/helpers.go#L128 and then another one around the grpc refactor https://github.com/grafana/k6/blob/5c357289fcc7a717a9b39990bd975ea2784b42bc/lib/netext/grpcext/conn.go#L58

Both of those should be removed and we should probably figure out some way to make certain we don't import the js package from lib and ... other places

codebien:
It could be helpful to implement https://github.com/OpenPeeDeeP/depguard or something similar to avoid the same mistake in the future.

na--:
Closed by https://github.com/grafana/k6/pull/2809 (which was actually merged in https://github.com/grafana/k6/pull/2826)

PR #2536, opened 2022-05-13
Move InterruptError in errext and other refactorings

mstoykov:
This is mostly done so that ./lib/executor doesn't need to import
./js/common and goja

part of #2535


olegbespalov (approved):

codebien (approved):
LGTM :bow:

PR #2543, opened 2022-05-17
Replace badssl.com in tests

### LINKED ISSUE #2302: Remove usage of `badssl.com` in k6 tests
Currently, we are making external requests to `badssl.com` to test some specific SSL errors and compatibility with old versions [1](https://github.com/grafana/k6/blob/922c963f46466ba89deb506115a23b7bc7d2c822/js/runner_test.go#L1046) [2](https://github.com/grafana/k6/blob/922c963f46466ba89deb506115a23b7bc7d2c822/js/runner_test.go#L1046) [3](https://github.com/grafana/k6/blob/3bcdabbf84b4232c69233ce67ec2d5adf8a4a57e/js/modules/k6/http/request_test.go#L2313-L2324). 

This probably need to be looked over and seen if they are still *all* needed, there seems to be some repetition at first glance, and tried to use a local HTTPS server, which should be possible for all of them. 

There is also a [sample test](https://github.com/grafana/k6/blob/a33b054344416f32a5034732be1da06018a682bc/samples/tlsconfig.js) which probably can stay as it is, I am just mentioning it for completeness.

The idea with not using an external site is both to make the tests more resilient to outside change, but also to make them faster, as making network requests outside of localhost is definitely slower. I would also recommend on not just adding all of this functionality to the httpmultibin but instead just specific server or even endpoints (not certain if possible) for the cases we care about. 

This was extracted as the last part of https://github.com/grafana/k6/issues/537, before closing it.

mstoykov:
fixes #2302

### DIFF
--- a/js/modules/k6/http/request_test.go
+++ b/js/modules/k6/http/request_test.go
@@ -21,7 +21,6 @@ import (
 	"net/http/cookiejar"
 	"net/http/httptest"
 	"net/url"
-	"runtime"
 	"strconv"
 	"strings"
 	"testing"
@@ -2348,9 +2347,6 @@ func TestErrorsWithDecompression(t *testing.T) {
 }
 
 func TestRequestAndBatchTLS(t *testing.T) {
-	if runtime.GOOS == "windows" {
-		t.Skip()
-	}
 	t.Parallel()
 
 	t.Run("cert_expired", func(t *testing.T) {
--- a/js/modules/k6/http/request_test.go
+++ b/js/modules/k6/http/request_test.go
@@ -2356,7 +2356,7 @@ func TestRequestAndBatchTLS(t *testing.T) {
 	t.Run("cert_expired", func(t *testing.T) {
 		t.Parallel()
 		_, state, _, rt, _ := newRuntime(t)
-		cert, key := GenerateTLSCertificate(t, "expired.com", time.Now().Add(-time.Hour), 0)
+		cert, key := GenerateTLSCertificate(t, "expired.localhost", time.Now().Add(-time.Hour), 0)
 		s, client := GetTestServerWithCertificate(t, cert, key)
 		go func() {
 			_ = s.Config.Serve(s.Listener)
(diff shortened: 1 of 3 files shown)


na--:
https://github.com/grafana/k6/issues/2578 is somewhat connected

mstoykov:
Not really - the problem there really is the fact that the go stdlib doesn't returned typed errors with 1.18 on mac os and apperantly using the system CA roots. Not that the test fail as badssl isn't there or does something wrong.

mstoykov:
More or less all of this is really fragile - changing more or less anything breaks one or more of the tests. 

Unfortunately we once again have a failing CI due ot badssl renewal problems, so even this more fragile code is better. It also happens to be fairly decoupled from any other tests so it should not fail just because we change something for another test :crossed_fingers: . 

If this is approved I will open an issue to fix this in the future :tm:

js/modules/k6/http/request_test.go
```diff
@@ -2439,3 +2530,120 @@ func TestBinaryResponseWithStatus0(t *testing.T) {
 	`)
 	require.NoError(t, err)
 }
+
+func GenerateTLSCertificate(t *testing.T, host string, notBefore time.Time, validFor time.Duration) ([]byte, []byte) {
```
na--, line 2534: hmm can you add these helper functions to the `go.k6.io/k6/lib/testutils/httpmultibin` or `go.k6.io/k6/lib/testutils` package, then you won't need to copy them in 2 places almost verbatim?

mstoykov: They are different so - no we can't :)

na--: Yeah, but I compared them and the differences seemed minor enough to parameterize? In any case, not a deal breaker, it can be fixed later or not at all - anything is better than relying on badssl :sweat_smile:

na-- (dismissed):
LGTM besides the minor nitpick I mentioned inline. I only skimmed some parts of this PR, so I haven't double-checked if the broken crypto is broken exactly as we want it to :sweat_smile: But I don't think it matters all that much, anything is better than having these tests randomly break every few weeks...

na-- (dismissed):

olegbespalov (approved):
:+1: great work!

A minor nitpick, maybe is a suggestion to not use domain names like `mybadssl.com` or `tlsv10.com`, but explicitly name them `bad-ssl-stub.localhost` or whatever. 

Something that indicates that this is a local/stub/fake domain without digging up.

na-- (dismissed):

na-- (approved):

olegbespalov (approved):
:muscle:

PR #2552, opened 2022-06-02
Update github.com/tidwall/gjson to 1.14.1 from 1.14.0

mstoykov:
The update is an optimizations by making a custom AppendJSONString
instead of using encoding/json. This is 2x faster for this particular
thing.


olegbespalov (approved):

codebien (approved):
LGTM, @mstoykov the CLA is stuck

PR #2554, opened 2022-06-02
Update gopkg.in/yaml.v3 to v3.0.1 from pre v3.0.0

mstoykov:
This has a bunch of small updates which are unlikely to matter for the
case k6 uses this.
The only reason to update is CVE-2022-28948.

Which is only applicable for k6 when it's running its tc39 tests.

So again no problem for our users.


olegbespalov (approved):

codebien:
@mstoykov It seems you need to rebase, GitHub says 3 commits.

mstoykov:
@codebien it says that but then when you `rebase and merge` it works so ... :shrug:

PR #2558, opened 2022-06-08
exec.vu.tags: Handle null or undefined values

codebien:
Setting a property of `vu.tags` as  `null` or `undefined`, like the following example, generates a panic. 

```javascript

export default function () {
	exec.vu.tags['any'] = null; // or undefined
}
```

- The new logic discards the set operation for `null` and `undefined` values and it prints a warning, instead, in case the `--throw` option is set, then it panics.
- In the case of the `--throw` option set, it's raised a panic passing a TypeError object.

### DIFF
--- a/js/modules/k6/execution/execution.go
+++ b/js/modules/k6/execution/execution.go
@@ -325,7 +325,11 @@ func (o *tagsDynamicObject) Get(key string) goja.Value {
 // In any other case, if the Throw option is set then an error is raised
 // otherwise just a Warning is written.
 func (o *tagsDynamicObject) Set(key string, val goja.Value) bool {
-	switch val.ExportType().Kind() { //nolint:exhaustive
+	kind := reflect.Invalid
+	if typ := val.ExportType(); typ != nil {
+		kind = typ.Kind()
+	}
+	switch kind {
 	case
 		reflect.String,
 		reflect.Bool,
@@ -335,12 +339,12 @@ func (o *tagsDynamicObject) Set(key string, val goja.Value) bool {
 		o.State.Tags.Set(key, val.String())
 		return true
 	default:
-		err := fmt.Errorf("only String, Boolean and Number types are accepted as a Tag value")
+		reason := "only String, Boolean and Number types are accepted as a Tag value"
 		if o.State.Options.Throw.Bool {
-			common.Throw(o.Runtime, err)
+			common.Throw(o.Runtime, fmt.Errorf(reason))
 			return false
 		}
-		o.State.Logger.Warnf("the execution.vu.tags.Set('%s') operation has been discarded because %s", key, err.Error())
+		o.State.Logger.Warnf("the execution.vu.tags.Set('%s') operation has been discarded because %s", key, reason)
 		return false
 	}
 }
--- a/js/modules/k6/execution/execution.go
+++ b/js/modules/k6/execution/execution.go
@@ -341,8 +341,7 @@ func (o *tagsDynamicObject) Set(key string, val goja.Value) bool {
 	default:
 		reason := "only String, Boolean and Number types are accepted as a Tag value"
 		if o.State.Options.Throw.Bool {
-			common.Throw(o.Runtime, fmt.Errorf(reason))
-			return false
+			panic(o.Runtime.NewTypeError(reason))
 		}
 		o.State.Logger.Warnf("the execution.vu.tags.Set('%s') operation has been discarded because %s", key, reason)
 		return false
--- a/js/modules/k6/execution/execution_test.go
+++ b/js/modules/k6/execution/execution_test.go
@@ -142,7 +142,7 @@ func TestVUTags(t *testing.T) {
 			assert.Equal(t, "vu101", val.String())
 		})
 
-		t.Run("DiscardWrongTypeRaisingError", func(t *testing.T) {
+		t.Run("DiscardWrongTypeAndRaisingError", func(t *testing.T) {
 			t.Parallel()
 
 			tenv := setupTagsExecEnv(t)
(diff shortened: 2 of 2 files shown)


js/modules/k6/execution/execution.go
```diff
@@ -325,6 +325,10 @@ func (o *tagsDynamicObject) Get(key string) goja.Value {
 // In any other case, if the Throw option is set then an error is raised
 // otherwise just a Warning is written.
 func (o *tagsDynamicObject) Set(key string, val goja.Value) bool {
+	if goja.IsUndefined(val) || goja.IsNull(val) {
+		o.State.Tags.Set(key, "")
+		return true
+	}
 	switch val.ExportType().Kind() { //nolint:exhaustive
```
olegbespalov, line 332: I'm wondering if we should set the tags as the empty case that way :thinking: especially since we have a "strict mode" option  (`o.State.Options.Throw.Bool`)

Maybe we should use something like this :thinking:   
```suggestion
	kind := reflect.Invalid
	if val != nil {
		kind = val.ExportType().Kind()
	}

	switch kind { //nolint:exhaustive
```

WDYT?

js/modules/k6/execution/execution_test.go
```diff
@@ -111,10 +111,12 @@ func TestVUTags(t *testing.T) {
 				v   interface{}
 				exp string
 			}{
-				"string": {v: `"tag1"`, exp: "tag1"},
-				"bool":   {v: true, exp: "true"},
-				"int":    {v: 101, exp: "101"},
-				"float":  {v: 3.14, exp: "3.14"},
+				"undefined": {v: `null`, exp: ""},
+				"null":      {v: `undefined`, exp: ""},
```
mstoykov, line 115:
```suggestion
				"null":           {v: `null`, exp: ""},
				"undefined": {v: `undefined`, exp: ""},
```
Not that we use the map key anywhere ... but it seems to be the "name" of the test in all other cases ;)

js/modules/k6/execution/execution.go
olegbespalov: > I saw the logic we have if we do the same from the options I changed because it wouldn't be consistent

I'm not sure that I know what this logic is :disappointed:

codebien: I mean the example reported in the issue's description:

```javascript

export const options = {
	tags: {
		undef: undefined,
		nul: null,
	}
};

export default function () {
	console.log('' === exec.vu.tags.nul) // true
	console.log('' === exec.vu.tags.undef) // true
}
```

mstoykov: Not really certain what the behaviour needs to be :(. It setting those to `''` seems bad. IMO it should either:
1. throw an exception as it just being pointless (maybe depending on `options.throw`)
2. actually set it to `null` and `undefined`. And return those on `Get` and (true on `Has`). But I am not really certain what that will be used for and ... it likely will need to stringify in an output to `"null"`/`"undefined"` or `''` all of which seem bad.
So I guess I am leaning towards `1` :shrug: 

I was going to argue this should delete a tag .. but `delete` exists so it seems like this isn't needed

oleiade: I would be inclined to go for @mstoykov's 1 too. It could probably throw a JS `TypeError`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypeError?

codebien: @na-- do you want to do the same if we set from the exported options or just from the execution API?

mstoykov (dismissed):

olegbespalov (dismissed):
:+1:

js/modules/k6/execution/execution.go
codebien: >It could probably throw a JS TypeError: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypeError?

@oleiade if `--throw` is set then we use `common.Throw` that accepts just a Go error type. `goja.NewTypeError` returns a `*goja.Object` so I guess we can't use it.

mstoykov: You can just `panic` with the `TypeError` directly ?

olegbespalov: Fix for the panic, that panics? :see_no_evil:

mstoykov: What do you think `common.Throw` does ?
https://github.com/grafana/k6/blob/27f2ead359dac24b4131dac4b57c95c058f06e48/js/common/util.go#L35-L41

codebien: Damn, I also tried it. It's the habit of not using different types from err and string. :facepalm:

js/modules/k6/execution/execution_test.go
```diff
@@ -142,21 +142,24 @@ func TestVUTags(t *testing.T) {
...
+				`[1, 3, 5]`,             // array
+				`{f1: "value1", f2: 4}`, // object
+			}
 
-			// object
-			_, err = tenv.Runtime.RunString(`exec.vu.tags["custom-tag"] = {f1: "value1", f2: 4}`)
-			require.Contains(t, err.Error(), "only String, Boolean and Number")
+			for _, val := range cases {
+				_, err := tenv.Runtime.RunString(`exec.vu.tags["custom-tag"] = ` + val)
+				require.Contains(t, err.Error(), "TypeError:")
```
mstoykov, line 160:
```suggestion
				_, err := tenv.Runtime.RunString(`exec.vu.tags["custom-tag"] = ` + val)
				require.Error(t, err)
				require.Contains(t, err.Error(), "TypeError:")
```

otherwise if this doesn't produce an error it will panic the tests

olegbespalov (approved):

mstoykov (approved):

PR #2563, opened 2022-06-13
PoC ESM support

mstoykov:
This is not a final PR it is here mostly to discuss whether the changes presented here are satisfying our needs. 

This PR will need to be merged *after* the goja PR (to be done) with the ESM support is merged. 

TODO:
- [ ] A lot more tests especially with multiple commonjs and not commonjs modules
- [ ] Support go modules that are Cyclic Module Records and them requesting (at least) other go modules. This is needed for k6/ws -> k6/http.CookieJar
- [ ] Support import.meta and dynamic import - the later likely will not be enabled by default in k6 at least for now. 
- [ ]  Cleaning up and fixing `require` and `open` as they need GetActiveScriptOrModule from the specification to be written correctly ;(.


js/bundle.go
```diff
@@ -58,18 +60,109 @@ type Bundle struct {
...
+	return bi.ModuleInstance.GetBindingValue(unistring.NewFromString(name))
+}
+
+type moduleCacheElement struct {
+	err error
+	m   goja.ModuleRecord
+}
+
+// TODO use the first argument
+func (b *Bundle) resolveModule(ref interface{}, specifier string) (goja.ModuleRecord, error) {
```
codebien, line 91: It's a very long method and a bit difficult to follow/review, can we split it a bit?

codebien (commented):
Some basic comments for starting the discussion

mstoykov:
Superseeded by #3456

PR #2564, opened 2022-06-13
Use require a lot more than assert in js tests

mstoykov:
The js tests had a bunch of places where they were using `assert` instead of `require`.

Mostly cases where the test should've not continued such as expecting error or no error. In some places `if assert....` was used which also is completely not needed.

I have kept most usage of `assert` where they will report multiple (and even singular failures sometimes), but I would argue in most cases just using `require` for everything is likely enough. 

There are also a few places where I've added `requires` that were previously missing.

Finally, there are few places that could've been left as `assert` but I decided to go for `require` in order for there to be no `assert.Error` for example. As I would argue this should always be with require.

Some of those lead to *fewer* reported failures as previously an `if assert ...` was followed by more code that could run even if that assert is a failure. But I would argue in all of those cases that is unlikely, and even if that happens this particular failures will likely not help debug the original problem.

### DIFF
--- a/js/console_test.go
+++ b/js/console_test.go
@@ -393,7 +393,7 @@ func TestFileConsole(t *testing.T) {
 							entryStr, err := entry.String()
 							require.NoError(t, err)
 
-							f, err = os.Open(logFilename)
+							f, err = os.Open(logFilename) //nolint:gosec
 							require.NoError(t, err)
 
 							fileContent, err := ioutil.ReadAll(f)
--- a/js/console_test.go
+++ b/js/console_test.go
@@ -32,16 +32,16 @@ func TestConsoleContext(t *testing.T) {
 	_ = rt.Set("console", &console{logger})
 
 	_, err := rt.RunString(`console.log("a")`)
-	assert.NoError(t, err)
-	if entry := hook.LastEntry(); assert.NotNil(t, entry) {
-		assert.Equal(t, "a", entry.Message)
-	}
+	require.NoError(t, err)
+	entry := hook.LastEntry()
+	require.NotNil(t, entry)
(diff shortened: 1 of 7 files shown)


mstoykov:
This PR will be squashed upon merging. I have separate commits as that was easier to work with locally. 

You can review them all at the same time, but I would recommend disabling whitespace diffing ;)

js/console_test.go
```diff
@@ -371,40 +371,39 @@ func TestFileConsole(t *testing.T) {
...
-								entryStr, err := entry.String()
-								assert.NoError(t, err)
+							// Test if what we logged to the hook is the same as what we logged
+							// to the file.
+							entryStr, err := entry.String()
+							require.NoError(t, err)
 
-								f, err := os.Open(logFilename)
-								assert.NoError(t, err)
+							f, err = os.Open(logFilename) //nolint:gosec
```
na--, line 396: This is weird :confused: `getSimpleRunner()` uses `afero.NewMemMapFs()` by default, so why (and how) are we polluting the real filesystem here? :confused:

na-- (approved):

js/console_test.go
mstoykov: We don't use afero in the console code at all - I guess this is a thing we can open a issue about :thinking: 

https://github.com/grafana/k6/blob/aa064148a11fa70defc300b513a5a20f7bf21868/js/console.go#L44

codebien (approved):
LGTM

js/console_test.go
mstoykov: https://github.com/grafana/k6/issues/2565

PR #1430, opened 2020-05-01
fix setting js options on the wrong object

MStoykov:
previos to this we were getting a possibly unexported variable "options"
and setting values to it. But the exported options could not have a
local variable with the same name ...

This also fixes a test to specifically need this to work.

PR #2566, opened 2022-06-13
Add to showcase how currently options is accessible everywhere

mstoykov:
This is also true for anything defined in the main module exported or
not. It is just much easier to showcase with options particularly.

This IMO should never have been the case and should be fixed.

Additionally ESM in practice will "fix" this as modules all have their
own scopes. So it makes sense to maybe break this before we add ESM
support

### DIFF
--- a/js/module_loading_test.go
+++ b/js/module_loading_test.go
@@ -25,6 +25,7 @@ import (
 	"context"
 	"os"
 	"testing"
+	"time"
 
 	"github.com/spf13/afero"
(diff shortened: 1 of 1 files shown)


mstoykov:
Partially previously discussed in https://github.com/grafana/k6/pull/1430

mstoykov:
closing this as I already opened a [PR](https://github.com/grafana/k6/pull/2571) with the fix in which we can discuss the issue showcased here

Issue #2545, opened 2022-05-23
Using `k6/execution` in setup leads to a panic


mstoykov:
### Brief summary

The following scripts panics
```javascript
import { scenario } from 'k6/execution';

export function setup() {
  console.log(scenario.iterationInTest)
}

export default () => { }
```

### k6 version

v0.34.0 forward

### OS

all

### Steps to reproduce the problem

run 
```javascript
import { scenario } from 'k6/execution';

export function setup() {
  console.log(scenario.iterationInTest)
}

export default () => { }
```

### Expected behaviour

No idea - this likely should be an exception as `setup` isn't part of any specific scenario so accessing this specifically should definitely not return a meaningful result. But a panic is not the correct result as well. 

### Actual behaviour

getting a panic, specifficallt the below one with the current master/v0.38.0

```
ERRO[0000] panic: runtime error: invalid memory address or nil pointer dereference
goroutine 1 [running]:
runtime/debug.Stack()
        runtime/debug/stack.go:24 +0x65
go.k6.io/k6/js/common.RunWithPanicCatching.func1()
        go.k6.io/k6/js/common/util.go:101 +0x19a
panic({0xfb3800, 0x1d6a8e0})
        runtime/panic.go:1038 +0x215
github.com/dop251/goja.AssertFunction.func1.1()
        github.com/dop251/goja@v0.0.0-20220405120441-9037c2b61cbf/runtime.go:2300 +0x6e
panic({0xfb3800, 0x1d6a8e0})
        runtime/panic.go:1038 +0x215
github.com/dop251/goja.(*vm).try.func1()
        github.com/dop251/goja@v0.0.0-20220405120441-9037c2b61cbf/vm.go:537 +0x635
panic({0xfb3800, 0x1d6a8e0})
        runtime/panic.go:1038 +0x215
go.k6.io/k6/js/modules/k6/execution.(*ModuleInstance).newScenarioInfo.func7()
        go.k6.io/k6/js/modules/k6/execution/execution.go:132 +0x60
reflect.Value.call({0xf58fa0, 0xc001c249b0, 0x49af3c}, {0x1103ff3, 0x4}, {0x1dbb168, 0x0, 0x106d9c0})
        reflect/value.go:557 +0x845
reflect.Value.Call({0xf58fa0, 0xc001c249b0, 0xc00258afc0}, {0x1dbb168, 0x0, 0x0})
        reflect/value.go:340 +0xc5
[...]

Issue #2260, opened 2021-11-23
`k6/execution`: add more information about the current VU


na--:
This is a continuation of some of the things mentioned in https://github.com/grafana/k6/issues/1320 that relate to information about the current VU, but which were not implemented in the original `k6/execution` API that shipped in k6 v0.34.0.

We already offer quite a lot of information through [`k6/execution.vu`](https://k6.io/docs/javascript-api/k6-execution/#vu), but we don't have these:
- the current group name / path - [`group()`](https://k6.io/docs/javascript-api/k6/group-name-fn/) calls can be nested arbitrarily deeply, and when we couple them with some code reuse functions, it might be useful to be able to programmatically get what is the current group. This might become even more relevant if we enhance what `group()` can do by implementing https://github.com/grafana/k6/issues/884 or something similar...
- `idInScenario` - the ID of the current VU in the current scenario. Since VUs are initialized concurrently after #1007, and since VUs can be reused across scenarios, the current `vu.idInInstance` and `vu.idInTest` identifiers are quite jumbled and unsuitable for anything besides partitioning data. This local ID for every scenario could be useful, especially in executors like [`ramping-vus`](https://k6.io/docs/using-k6/scenarios/executors/ramping-vus/) where VUs are very ordered and lower-numbered VUs will always be started first and stopped last.

na--:
A way to atomically query the current `BytesWritten` and `BytesRead` (i.e. the values that are emitted as `data_sent` and `data_received` at the end of the iteration) also makes some sense to be added to the `execution.vu` capabilities: https://github.com/grafana/k6/issues/2283#issuecomment-985419488 (with https://github.com/grafana/k6/issues/1250 maybe being implemented at the same time :man_shrugging:)

amommersteeg-d2l:
Hello I am interested in the `idInScenario` feature suggested above, has there been any progress? 

For our use case we have a test that requires unique data read from an array, we were using the `vu.idInInstance` as the index for that array. But when this test runs with other tests as a scenario the `vu.idInInstance` can start to be larger than the length of the array.

na--:
@amommersteeg-d2l, there hasn't been any progress here because this issue hasn't been very highly prioritized so far. Thanks for sharing your use case, that helps us to prioritize more accurately and understand what we should build. Please also upvote (:+1:) the issue itself.

[...]

Issue #2259, opened 2021-11-23
`k6/execution`: add more information about the current test run


na--:
This is a continuation of some of the things mentioned in https://github.com/grafana/k6/issues/1320 that relate to information about the currently running test, but which were not implemented in the original `k6/execution` API that shipped in k6 v0.34.0.

https://github.com/grafana/k6/pull/2093 adds a new `k6/execution.test.abort()` method, but we currently don't expose any of the information about the current test run in it. Some potentially useful things we could add expose there:
- the start time of the test run
- the total calculated test run duration (from the execution plan)
- the final consolidated script `options`? this seems cleaner than relying on the fact that we'll update the exported script `options`, and we've had a bunch of issues with that (see https://github.com/grafana/k6/pull/1430)...
- other things like the whole runtime options? or at leasat parts of them - an alternative interface for accessing the environment variables will be much nicer than `__ENV`... 
- the current working directory and file name - this should return the original ones when running archive bundle, but...
- ... there should be a way to tell if we're executing a JS file or a .tar archive bundle (which will be helpful as a proxy to determine if we're running the script locally or in the cloud)
- _(somewhat unlikely)_ the execution segment sequence, though that is somewhat dependent on any new APIs from https://github.com/grafana/k6/issues/1539 and the actual instance execution segment should probably be in `k6/execution.instance`...

Some of these things will probably be useful in the init context as well (e.g. the CWD and env vars), so parts of `k6/execution.test` should definitely be accessible from the init context. Some parts, like the consolidated `options` or total test run duration, can't be available in the first init context execution (because it is the thing that actually gets the exported `options`...), but would be quite useful in subsequent VU initializations...

imiric:
Just a note so that we don't make the same mistake from #1320: let's avoid suggesting features for the `k6/execution` API until we're sure that users actually want them. This is fine for cases where it actually solves an existing issue or makes a previously cumbersome workflow easier, but each addition is a change that possibly complicates things and something we'll have to maintain, so the leaner we can keep this API, the better.

na--:
If you look at the old issue, a lot (most?) of these actually came from user requests. But yeah, in general, I agree and this is not a very high priority issue as of now. I'd appreciate any users to comment with their specific use cases here, that will help with us prioritizing it more highly.

tom-miseur:
[...]

PR #2567, opened 2022-06-13
fix: proper error exec.scenario.* in not init context

olegbespalov:
# What?

Always check if `scenarioState` is available for the `execution.scenario`, otherwise erroring.

# Why?

It makes no sense to use this `execution.scenario` not inside the VU context.

Closes: #2545

### DIFF
--- a/js/modules/k6/execution/execution.go
+++ b/js/modules/k6/execution/execution.go
@@ -90,18 +90,20 @@ func (mi *ModuleInstance) Exports() modules.Exports {
 	return modules.Exports{Default: mi.obj}
 }
 
+var errRunInInitContext = errors.New("getting scenario information in the init context is not supported")
+
 // newScenarioInfo returns a goja.Object with property accessors to retrieve
 // information about the scenario the current VU is running in.
 func (mi *ModuleInstance) newScenarioInfo() (*goja.Object, error) {
 	rt := mi.vu.Runtime()
 	vuState := mi.vu.State()
 	if vuState == nil {
-		return nil, errors.New("getting scenario information in the init context is not supported")
+		return nil, errRunInInitContext
 	}
 	getScenarioState := func() *lib.ScenarioState {
 		ss := lib.GetScenarioState(mi.vu.Context())
 		if ss == nil {
-			common.Throw(rt, errors.New("getting scenario information in the init context is not supported"))
+			common.Throw(rt, errRunInInitContext)
 		}
 		return ss
 	}
@@ -126,9 +128,17 @@ func (mi *ModuleInstance) newScenarioInfo() (*goja.Object, error) {
 			return p
 		},
 		"iterationInInstance": func() interface{} {
+			if vuState.GetScenarioLocalVUIter == nil {
+				common.Throw(rt, errRunInInitContext)
+			}
+
 			return vuState.GetScenarioLocalVUIter()
 		},
 		"iterationInTest": func() interface{} {
+			if vuState.GetScenarioGlobalVUIter == nil {
+				common.Throw(rt, errRunInInitContext)
+			}
+
 			return vuState.GetScenarioGlobalVUIter()
 		},
 	}
--- a/js/modules/k6/execution/execution.go
+++ b/js/modules/k6/execution/execution.go
@@ -90,7 +90,7 @@ func (mi *ModuleInstance) Exports() modules.Exports {
 	return modules.Exports{Default: mi.obj}
 }
 
-var errRunInInitContext = errors.New("getting scenario information outside of vu code is not supported")
+var errRunInInitContext = errors.New("getting scenario information outside of the VU context is not supported")
 
 // newScenarioInfo returns a goja.Object with property accessors to retrieve
 // information about the scenario the current VU is running in.
--- a/js/modules/k6/execution/execution.go
+++ b/js/modules/k6/execution/execution.go
@@ -90,7 +90,7 @@ func (mi *ModuleInstance) Exports() modules.Exports {
 	return modules.Exports{Default: mi.obj}
 }
 
-var errRunInInitContext = errors.New("getting scenario information in the init context is not supported")
+var errRunInInitContext = errors.New("getting scenario information outside of vu code is not supported")
 
 // newScenarioInfo returns a goja.Object with property accessors to retrieve
 // information about the scenario the current VU is running in.
--- a/js/modules/k6/execution/execution_test.go
+++ b/js/modules/k6/execution/execution_test.go
@@ -418,3 +418,32 @@ func TestOptionsTestSetPropertyDenied(t *testing.T) {
 	require.NoError(t, err)
 	assert.Equal(t, true, rt.ToValue(paused).ToBoolean())
 }
+
+func TestScenarioNoAvailableInInitContext(t *testing.T) {
+	t.Parallel()
+
+	rt := goja.New()
+	m, ok := New().NewModuleInstance(
+		&modulestest.VU{
+			RuntimeField: rt,
+			InitEnvField: &common.InitEnvironment{},
+			CtxField:     context.Background(),
+			StateField: &lib.State{
+				Options: lib.Options{
+					Paused: null.BoolFrom(true),
+				},
+			},
+		},
+	).(*ModuleInstance)
+	require.True(t, ok)
+	require.NoError(t, rt.Set("exec", m.Exports().Default))
+
+	scenarioExportedProps := []string{"name", "executor", "startTime", "progress", "iterationInInstance", "iterationInTest"}
+
+	for _, code := range scenarioExportedProps {
+		prop := fmt.Sprintf("exec.scenario.%s", code)
+		_, err := rt.RunString(prop)
+		require.Error(t, err)
+		require.ErrorContains(t, err, "getting scenario information in the init context is not supported")
+	}
+}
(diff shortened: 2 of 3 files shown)


na-- (commented):
I am not sure throwing an exception is the right thing to do here :thinking: We are pretending that the things in `k6/execution` are properties, and this doesn't seem very idiomatic. Usually, when some object property in JS is unavailable, it is just not there, so attempting to access it will result in an `undefined` result and not an exception, right?

olegbespalov:
Hey @na-- , thanks for the input!

> We are pretending that the things in k6/execution are properties, and this doesn't seem very idiomatic. 

Not sure if I had a chance to learn it from somewhere :shrug: 

Moreover it a bit contradicts the existing logic, like if you try to access the properties: "name", "executor", "startTime", "progress" you get the error, not the `undefined` or default value.

The change is also just extending the same behavior to the  "iterationInInstance", "iterationInTest" properties.

So, I'm a bit confused :thinking:

mstoykov:
I don't think that throwing exception on property access is unidiomatic. Move over in this particular case the idea is mostly that `exec.scenario` is "undefined" and you are trying to access a property of it - which will lead to an exception either way.


From a UX perspective I would also argue that it will be a lot easier to debug when `exec.scenaro.name` throws an exception instead of returning you a `""` or `null` or something.

mstoykov (dismissed):
LGTM

na--:
> > We are pretending that the things in k6/execution are properties, and this doesn't seem very idiomatic.
> 
> Not sure if I had a chance to learn it from somewhere shrug

I just meant that the API is `exec.scenario.name`, not something like `exec.scenario.getName()`. So the elements of `k6/execution` are properties, not methods (with the exception of `test.abort()`) and they should behave like normal JS properties, if possible.

> Moreover it a bit contradicts the existing logic, like if you try to access the properties: "name", "executor", "startTime", "progress" you get the error, not the `undefined` or default value.
> 
> The change is also just extending the same behavior to the "iterationInInstance", "iterationInTest" properties.
> 
> So, I'm a bit confused thinking

Yeah, very good points :thinking: But instead of making this work like them, I kind of thing we should also fix them. I don't see a reason why this should actually fail:
```js

console.log(JSON.stringify(exec, null, 4));

export default function () {
    console.log(JSON.stringify(exec, null, 4));
}
```

it should just print the execution properties that were available in each context, something like this:
```js
// Init context
{
    "instance": {
        "iterationsCompleted": 0,
        "iterationsInterrupted": 0,
        "vusActive": 0,
        "vusInitialized": 0,
        "currentTestRunDuration": 0
    },
    "test": {
        "options": { /* ... */ }
        // abort() is not printed by JSON.stringify(), but it's there
    },
    "vu": {
        "idInInstance": 0,
        "idInTest": 0
    }
} 

// VU context
{
    "instance": {
        "iterationsCompleted": 0,
        "iterationsInterrupted": 0,
        "vusActive": 1,
        "vusInitialized": 1,
        "currentTestRunDuration": 0.385693
    },
    "scenario": {
        "name": "default",
        "executor": "per-vu-iterations",
        "startTime": 1655191933532,
        "progress": 0,
        "iterationInInstance": 0,
        "iterationInTest": 0
    },
    "test": {
        "options": { /* ... */ }
        // abort is not printed by JSON.stringify(), but it's there
    },
    "vu": {
        "idInInstance": 1,
        "idInTest": 1,
        "iterationInInstance": 0,
        "iterationInScenario": 0,
        "tags": {
            "group": "",
            "scenario": "default"
        }
    }
} 
```

...or something like that :man_shrugging: Take a look at some of the ideas in https://github.com/grafana/k6/issues/2259 and https://github.com/grafana/k6/issues/2260, things like the CWD or a better way to access `__ENV` should be available in the init context.

olegbespalov:
@na-- 

> But instead of making this work like them, I kind of thing we should also fix them.

```js

console.log(JSON.stringify(exec, null, 4));

export default function () {
    console.log(JSON.stringify(exec, null, 4));
}
```

It's an excellent example, but the question do you think this is a real-world example?

I mean, as @mstoykov said, the UX aspect is essential. So basically, if we apply your suggestion instead of the precise error about using accessing some certain data, we start getting the property `undefined` in the example below.

```
import { scenario } from 'k6/execution';

export function setup() {
  console.log(scenario.iterationInTest)
}

export default () => { }
```

Is it worth it? :thinking:

mstoykov:
If we want this in v0.39.0 - I would argue keeping the current behaviour of throwing an exception is the only way forward. Otherwise, the changes required are way too big and will likely need a lot more testing and discussion.

I propose that we merge this as is, and open an issue to discuss potential changes. This change in practice only prevents a rare panic in a case where the API should not be used either way :shrug: .

na--:
> If we want this in v0.39.0 - I would argue keeping the current behaviour of throwing an exception is the only way forward. Otherwise, the changes required are way too big and will likely need a lot more testing and discussion.
>
> I propose that we merge this as is, and open an issue to discuss potential changes. This change in practice only prevents a rare panic in a case where the API should not be used either way shrug .

Agreed :+1: Relaxing this constraint ("throw an exception when accessing an invalid property") in the future won't even be a breaking change, if we want to eventually do that... :thinking: So no need to figure it out now.

> > We probably need some actual JS developers to chime in here
> 
> I'd argue smile With all my respect to JS developers, the target audience is load-test users, no?

Disagree about this though. Consistent and predictable APIs are useful for any user, long-time JS developers or no :smile:

na--:
To get back to this specific fix, I notice that it actually changes the "laziness" properties of `exec.scenario`. Before `getScenarioState()` was called on every property access, now it's called only once, and I am not sure that doesn't have any implications :thinking:

allansson:
> > We probably need some actual JS developers to chime in here
> 
> I'd argue 😄 With all my respect to JS developers, the target audience is load-test users, no?

Isn't the whole point using JS to get more people into load testing by using a language that a lot of people already know? Otherwise we could've rolled our own language, used XML or whatever. 

Either way, we _are_ using JS and with every language comes a set of expectations. If I'm writing an api in Go, I don't use snake_case and I know that my methods should return errors as the second element in a tuple and that it should either be the error or `nil`. It would be very frustrating if the arguments were swapped, because it subverts my expectation of what should happen.

People learning JS just to use k6 are probably gonna go to some generic JS tutorial to learn the fundamentals, not some k6 tutorial, so their expectations are most likely being set by the more general JS community. And even if they _are_ learning it from k6 tutorials there is a value in staying idiomatic because their knowledge will be transferable to general JS development.

We don't want k6 to be the PHP of JS-runtimes, right? 😅

So I'm with @na--  on this one. I can't remember a single case where an error was thrown because I was trying to _read_ a property. Throwing while setting a property would be fine. Assignment is a side-effect and you can expect some validation to occur. But reading in general is safe (as long as you are checking for undefined, of course).

IMO, the idiomatic way of preventing access at the global scope would've been to, instead of exporting an object, export a function from "k6/execution" either returning the `execution`-object or throwing an error if the function was called in an invalid context (e.g. the global scope).

olegbespalov:
Hey @allansson 

Thanks for the input. All that you said is valid, with one critical remark, and the @w1kman well defines this remark. 

> But I have no problem with an error in case it was thrown to stop me from "CLEARLY doing something wrong" (rather than spending time trying to figure out why undefined was returned). That would also require the error message to be super helpful.

It saves the user time, which means it's a better UX of the k6 tool.

If you are saying that it's wrong to throw an error from the property, well, that probably means that it shouldn't be property (which it is right now) :shrug: But I believe our users will be happier if they know sooner if the things that they are trying to do is not the right path.

codebien:
Could we apply the usual `--throw` logic? Raise an error if `--throw` is set and print a warning (or just return `undefined`) in the opposite?

olegbespalov:
> To get back to this specific fix, I notice that it actually changes the "laziness" properties of exec.scenario. Before getScenarioState() was called on every property access, now it's called only once, and I am not sure that doesn't have any implications thinking

So I addressed your concern in reverted the change. If we still want just fix the panic in the current release, this PR is only about fixing the panic, nothing more. 

>Could we apply the usual --throw logic? Raise an error if --throw is set and print a warning (or just return undefined) in the opposite?

We can, but it makes sense if it will be consistent across all API (or at least k6/execution), so that's why the next paragraph.

Keeping in mind all other input I have a strong feeling that any other change deserves a separate issue where we can discuss the API first and make a decision that we can use as the standard across all our APIs. The decision that respects the JS API and also has a good UX.

mstoykov (approved):

js/modules/k6/execution/execution.go
```diff
@@ -90,18 +90,20 @@ func (mi *ModuleInstance) Exports() modules.Exports {
 	return modules.Exports{Default: mi.obj}
 }
 
+var errRunInInitContext = errors.New("getting scenario information outside of vu code is not supported")
```
na--, line 93: :thinking: I still don't quite like it, but I think this is a bit better:

```suggestion
var errRunInInitContext = errors.New("getting scenario information outside of the VU context is not supported")
```

for sure `VU` should be capitalized

na-- (dismissed):
LGTM except the minor nitpick with the error message

js/modules/k6/execution/execution.go
olegbespalov: :facepalm:

na-- (approved):

na--:
For anyone who was part of the broader discussion above, or who stumbles on it in the future, @olegbespalov opened this issue where we can continue it: https://github.com/grafana/k6/issues/2574

PR #2573, opened 2022-06-15
Fix max duration calculations with multiple 0 vus stages at end

mstoykov:
In b725f03b we fixed the calculation of the stages, unfortunately we
didn't notice that the max duration for the test was wrongly calculated.

This also tries to fix cases where graceful* will extend the test more
then needed. Previously the code did not take into account that multiple
0 VUs stages can mean that the VUs are long ago stopped before
gracefulStop (for example) needs to be taken into account.

### DIFF
--- a/lib/executor/ramping_vus.go
+++ b/lib/executor/ramping_vus.go
@@ -209,8 +209,10 @@ func (vlvc RampingVUsConfig) getRawExecutionSteps(et *lib.ExecutionTuple, zeroEn
 
 		stageVUDiff := stageEndVUs - fromVUs
 		if stageVUDiff == 0 {
+			// skip those as this doesn't change anything
 			if i == len(vlvc.Stages)-1 {
-				// add the last step always
+				// don't skip the last step as we won't add another one after it
+				// unlike in any of the other cases
 				steps = append(steps, lib.ExecutionStep{TimeOffset: timeTillEnd, PlannedVUs: uint64(scaled)})
 			}
 			continue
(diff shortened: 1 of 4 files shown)


lib/executor/ramping_vus.go
```diff
@@ -202,13 +202,17 @@ func (vlvc RampingVUsConfig) getRawExecutionSteps(et *lib.ExecutionTuple, zeroEn
...
-	for _, stage := range vlvc.Stages {
+	for i, stage := range vlvc.Stages {
 		stageEndVUs := stage.Target.Int64
 		stageDuration := stage.Duration.TimeDuration()
 		timeTillEnd += stageDuration
 
 		stageVUDiff := stageEndVUs - fromVUs
 		if stageVUDiff == 0 {
+			if i == len(vlvc.Stages)-1 {
+				// add the last step always
```
olegbespalov, line 213: Out of curiosity, why only the last? Could we leave a comment about why we always add?

lib/executor/ramping_vus.go
```diff
@@ -426,6 +430,14 @@ func (vlvc RampingVUsConfig) reserveVUsForGracefulRampDowns( //nolint:funlen
 		})
 		lastPlannedVUs = rawStep.PlannedVUs
 	}
+	/*
```
olegbespalov, line 433: If we don't need it, why not drop it? :thinking:

lib/executor/ramping_vus.go
mstoykov: [dbb6f50](https://github.com/grafana/k6/pull/2573/commits/dbb6f50740a7532da7dac087236a122fde8e5511)  ?

olegbespalov (approved):
:crossed_fingers:

oleiade (approved):

PR #2575, opened 2022-06-20
Revert "Try to respect zero target stages at the end (#2532)"

mstoykov:
This reverts commit b725f03be5873b9b0d933b94a83924ff7c1f15e4.


mstoykov:
This reopens #2532 . It turned out there are more problems to be solved such as #2573 and this will require more time to be looked into, but v0.39.0 is around the corner. 

Given that this has been the case for nearly 2 years and how it hasn't really be a problem, we are reverting it until we have time to look into it a lot more.

olegbespalov (approved):

PR #2577, opened 2022-06-26
Fix contributing docs test command

Llandy3d:
This fixes that the `CONTRIBUTING.md` docs specify the command to run tests as `make test` while defined as `tests` in the makefile.


olegbespalov (approved):
@Llandy3d, thanks for fixing it! :+1:

na-- (approved):
Thanks!

Issue #2544, opened 2022-05-19
Integrating TaskQueue in k6


mstoykov:
While working on #2425 turned out that there is a very common pattern of having multiple events happening, needing some code to be ran on the event loop, possibly (and usually) before the previous ran has managed to complete. 

Usually this is done by calling `RegisterCallback` while on the event loop and keeping it as field on a struct while running *off* the event loop (in a separate goroutine). And then using that callback to run on the event loop and if needed to call `RegisterCallback` again to "restore" it and be able to use it again. 

This works great if you have 1 event for each time you need to something off the event loop. For example a (fairly) basic http request of the event loop will likely have only 1 thing that will make it run on event loop again - the request finishing. In this way each `http.asyncRequest` will call `RegisterCallback` before spinning off a goroutine and (for example) returning promise to the js. 

With web sockets there is very good chance that you will get a second event (receiving a message for example) wanting to run something on the event loop before the previous one has had time to run. This will be true as well for a more complex HTTP API having events for stuff such as "tls handshak start", "tls handshake started", "sended headers", "received headers", etc.

This is more or less what happened and is what https://github.com/MStoykov/k6-taskqueue-lib solves. 

It gets RegisterCallback, the whole method not just its result,  and will call it to get a callback. IT also exposes a [Queue](https://pkg.go.dev/github.com/mstoykov/k6-taskqueue-lib@v0.1.0/taskqueue#TaskQueue.Queue) method to queue concurrently, that will wait for the currently running one(if any) to finish, refresh the RegisterCallback and queue the ones that have been queued. 

The only strange part is that again the event loop needs to know that it can end, so there is also a [Close](https://pkg.go.dev/github.com/mstoykov/k6-taskqueue-lib@v0.1.0/taskqueue#TaskQueue.Close).

### Proposal 

Given that this seems like it will be really common pattern for protocols such as tcp, udp, websockets, http server-push, gRPC, just to name a few it makes sense that this gets better integration in k6.

Adding exactly the same API seems a bit strange as you have `RegisterCallback` and then this and likely most users will opt to use this especially as it makes it a lot less likely they will make a mistake. I even somewhat wonder if we should let RegisterCallback long-term. 

Some considerations include the fact that the ECMAScript specification talks about [jobs](https://262.ecma-international.org/12.0/#sec-jobs) and the web application api talks about task queues in it's [event loop section](https://www.w3.org/TR/2016/WD-html52-20161206/webappapis.html#event-loops).

[...]

PR #2582, opened 2022-06-30
Panic on RegisterCallback result being used more than once

mstoykov:
This should help with actually finding problems with code using it.

I have already seen multiple times code which seems okay but in fact
will call the returned callback twice. This effectively will lead either
to deadlock of the eventloop or earlier exit from a loop.

Given that it's never actually correct to that, panic seems okay.

We also make certain that if the user recovers from the panic we will
not have locked the eventloop either way,

In the future better APIs for most use cases should probably reduce the
need for people to use RegisterCallback and consequently worrying about
this.

### DIFF
--- a/js/eventloop/eventloop.go
+++ b/js/eventloop/eventloop.go
@@ -61,11 +61,17 @@ func (e *EventLoop) wakeup() {
 // This function *must* be called from within running on the event loop, but its result can be called from anywhere.
 func (e *EventLoop) RegisterCallback() func(func() error) {
 	e.lock.Lock()
+	var callbackCalled bool
 	e.registeredCallbacks++
 	e.lock.Unlock()
 
 	return func(f func() error) {
 		e.lock.Lock()
+		if callbackCalled { // this is protected by the lock on the event loop
+			e.lock.Unlock() // let not lock up the whole event loop, somebody could recover from the panic
+			panic("RegisterCallback called twice")
+		}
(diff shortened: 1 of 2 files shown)


js/eventloop/eventloop.go
```diff
@@ -61,11 +61,17 @@ func (e *EventLoop) wakeup() {
 // This function *must* be called from within running on the event loop, but its result can be called from anywhere.
 func (e *EventLoop) RegisterCallback() func(func() error) {
 	e.lock.Lock()
+	var done bool
```
na--, line 64: hmm `done` is misleading here, something like this seems better:

```suggestion
	var callbackRegistered bool
```

na-- (dismissed):
LGTM in general, besides the naming nitpick. And I agree, the real solution would be less footgun-prone API like https://github.com/grafana/k6/issues/2544 :sweat_smile:

js/eventloop/eventloop.go
mstoykov: `callbackRegisted` is also not right - it is registered once this method returns. Maybe `called`

na--: :+1: for `called` or `callbackWasCalled`, if we want to be more descriptive

na-- (approved):

imiric (approved):
Makes sense :+1:

PR #2581, opened 2022-06-29
k6 v0.39.0 release notes

na--:

PR #2583, opened 2022-06-30
Add a deprecation warning for thresholds on tags we'll make non-indexable

na--:
This is connected to the future deprecations that are described at the end of the release notes (https://github.com/grafana/k6/pull/2581/files). The earlier we mention them, the better.

### DIFF
--- a/cmd/integration_test.go
+++ b/cmd/integration_test.go
@@ -282,6 +282,42 @@ func TestSSLKEYLOGFILE(t *testing.T) {
 	require.Regexp(t, "^CLIENT_[A-Z_]+ [0-9a-f]+ [0-9a-f]+\n", string(sslloglines))
 }
 
+func TestThresholdDeprecationWarnings(t *testing.T) {
+	t.Parallel()
+
+	// TODO: adjust this test after we actually make url, error, iter and vu non-indexable
+
+	ts := newGlobalTestState(t)
+	ts.args = []string{"k6", "run", "--system-tags", "url,error,vu,iter", "-"}
+	ts.stdIn = bytes.NewReader([]byte(`
+		export const options = {
+			thresholds: {
+				'http_req_duration{url:https://test.k6.io}': ['p(95)<500', 'p(99)<1000'],
+				'http_req_duration{error:foo}': ['p(99)<1000'],
+				'iterations{vu:1,iter:0}': ['count == 1'],
+			},
+		};
+
+		export default function () { }`,
+	))
+
+	newRootCommand(ts.globalState).execute()
+
+	logs := ts.loggerHook.Drain()
+	assert.True(t, testutils.LogContains(logs, logrus.WarnLevel,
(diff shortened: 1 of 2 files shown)


cmd/integration_test.go
```diff
@@ -282,6 +282,42 @@ func TestSSLKEYLOGFILE(t *testing.T) {
...
+		"Thresholds that are based on the 'error' metric tag will be deprecated in the future. Please use the 'error_code'",
+	))
+	assert.True(t, testutils.LogContains(logs, logrus.WarnLevel,
+		"Thresholds like 'iterations{vu:1,iter:0}' that are based on the 'vu' metric will be deprecated",
+	))
+	assert.True(t, testutils.LogContains(logs, logrus.WarnLevel,
+		"Thresholds like 'iterations{vu:1,iter:0}' that are based on the 'iter' metric will be deprecated",
+	))
+}
+
```
codebien, line 320: Can we make all of them the same in the common part? At least for the first part.

```
"The 'http_req_duration{url:https://test.k6.io}' threshold is defined based on the 'bad_tag' metric tag that will be deprecated in the next releases."
```
And where available
```
"Please use the 'the_right_tag'".
```

codebien (commented):
Just a reflection:

Are we sure about deprecating some specific tags for the threshold operation? What I expect we will do is specifically disabling the pre-announced system tags as indexed tags so the threshold will be implicitly impacted. But I could potentially re-enable and set a threshold on it.

```js
export default function() {
    exec.scenario.tags['iter'] = {
       value: exec.iterationInTest,
       index: true,
    }
} 
```

It's probably an unlikely edge case, but are we sure the threshold is the right context?
In the case we will add the unindexed tags then I expect we will document very well what is an indexed or not indexed tag and the effects they will have on the rest of the k6 ecosystem.

mstoykov:
> Are we sure about deprecating some specific tags for the threshold operation? What I expect we will do is specifically disabling the pre-announced system tags as indexed tags so the threshold will be implicitly impacted. But I could potentially re-enable and set a threshold on it.

Given the fact that we have:
1. outright disabled this tags forever
2. have custom code to work around the fact that they are high cardinality bu available as tags
3. have custom options more or less specifically added for them to most outputs to remove or change them to not tags


I would argue this is a long time coming. 


Yes users can add `my_tag` and give it values from 1 to ♾ and that is fine 🐕🔥. But k6 by default having an (easy) way to completely destroy it performance seems like a *really* bad idea. 

As I said in private I am more for completely removing this tags and bringing them at some point as *non-indexable* tags/ fields/ metadata - w/e we will call it.

But this PR isn't about this it is about the fact that using those tags likely will not work for thresholds in the future irregardless of what we exactly decide to do.

mstoykov (dismissed):

cmd/integration_test.go
na--: the problem is that we won't be deprecating the tag, per se, we'll be changing it... :thinking: I'll try to reword them in a unified way somehow

codebien (approved):

mstoykov (approved):

olegbespalov (approved):

PR #2590, opened 2022-07-05
Bump direct dependencies

codebien:
Updated all the direct dependencies reported by `modtools`.

`ntlm` introduced a breaking change and there are some ongoing discussions, in any case, it seems on our side that k6 doesn't have the dependency from the change so it should be still fine to upgrade.

Check for details: https://github.com/Azure/go-ntlmssp/pull/34


olegbespalov (approved):
:crossed_fingers:

mstoykov (approved):

na-- (approved):

oleiade (approved):
🦖

PR #2593, opened 2022-07-07
Clean context usage after common.Bind is removed

mstoykov:


js/initcontext.go
```diff
@@ -160,9 +160,7 @@ type moduleVUImpl struct {
 }
 
 func newModuleVUImpl() *moduleVUImpl {
```
na--, line 162: This seems a bit useless now... Are we ever going to have something here? 

Looking at it a bit more, I think we can refactor this method to drop the `vuImpl *moduleVUImpl` altogether and just create it there: https://github.com/grafana/k6/blob/4fdfea41b366f641e9d8a75675896a54ab167656/js/bundle.go#L255-L258

Right now we just have a bunch of noise when we always pass the pointer to the empty struct from ` newModuleVUImpl()`, see: 
https://github.com/grafana/k6/blob/cccf2b35417fa6e46c1007deea753cce6c50dc49/js/runner.go#L155-L156
https://github.com/grafana/k6/blob/cccf2b35417fa6e46c1007deea753cce6c50dc49/js/bundle_test.go#L484
https://github.com/grafana/k6/blob/cccf2b35417fa6e46c1007deea753cce6c50dc49/js/bundle_test.go#L577
etc.

mstoykov: Managed to get it to work with a lot less changes than it looked like it will need through the magic of ... embeded structs :tada:

na-- (approved):
:crossed_fingers:

Issue #2580, opened 2022-06-29
Time Series data model


codebien:
We want to introduce the time series concept to k6 for getting the benefit of efficiently identifying the combination between a metric and the relative tags.

A summary of the benefits it brings across the platform:

* It allows executing faster lookups and comparisons by a single value, the ID (hash) of the series.
* It reduces the storage by allocating just one time a possible combination (the time series) and referencing it in all other places. 
* It allows aggregating data by time series.
* It enables easier detection of high cardinality issues just by counting the entire unique data set of generated time series.
* It enables time series based outputs to simplify the integration's logic (e.g. Prometheus).

#### Metrics and Series

* Metric (name, type)
* Time Series (metric, tags)
* Sample (timestamp, value, time series ref)
* Tag - (aka [k6/1831](https://github.com/grafana/k6/issues/1831))

```go
// This is the current k6 Metric polished from the Thresholds dependencies.
type Metric struct {
	name     string
	type     MetricType
	contains ValueType
}

type TagSet atlas.Node

type TimeSeries struct {
        Metric *TimeSeries
        Tags *TagSet
}

type Sample struct {
        TimeSeries *TimeSeries
        Timestamp uint64
        Value float64
}
```
#### Sink

The sinks are implemented by metric types and they keep values up to date by time series and/or aggregated views:

* Counter: a monotonic increasing value
* Gauge: the latest value
* Rate: the percentage of non-zero values across the total of the events 
* Trend as SparseHistogram: counters per dynamic ranges (buckets) - https://grafana.com/blog/2021/11/03/how-sparse-histograms-can-improve-efficiency-precision-and-mergeability-in-prometheus-tsdb

## Storage

It will store the root of Atlas in the `metrics.Registry` and it will provide a method for get it for branching out a new TagSet.

```go
type Registry struct {
        ...
	tagSetRoot *atlas.Node
}

func (r *Registry) RootTagSet() *TagSet {
	return (*TagSet)(r.tagSetRoot)
}
```

## Samples generation

The current sampling process is controlled by the `metrics.PushIfNotDone` method. All the actual callers should resolve the time series from the storage before pushing a new Sample, or in the event, no one is found the storage will create and insert a new one.
It requires the dependency from the time series database for all the callers (e.g. executors, JavaScript modules). Most of them already have the `metrics.Registry` dependency.

## metrics.Ingester

[...]

Issue #2584, opened 2022-07-04
Support non-indexable high-cardinality metric tags / metadata


na--:
Currently, the k6 [system metric tags](https://k6.io/docs/using-k6/tags-and-groups/#system-tags) are a mix of high-cardinality ones (`url`, `error`, `vu` and `iter`) and relatively low-cardinality ones (everything else). This has already been a problem multiple times and has forced us to adopt some hacks to work around the issues, e.g. these workarounds in the built-in outputs: [1](https://github.com/grafana/k6/blob/e7c2aeecfeda9b0e49b95cdd4a6afb3a416c5432/output/influxdb/config.go#L60), [2](https://github.com/grafana/k6/blob/e7c2aeecfeda9b0e49b95cdd4a6afb3a416c5432/output/statsd/config.go#L85), [3](https://github.com/grafana/k6/blob/e7c2aeecfeda9b0e49b95cdd4a6afb3a416c5432/output/cloud/output.go#L331-L337)). On top of that, even the current high-cardinality tags are going be a big problem for the upcoming time series changes (https://github.com/grafana/k6/issues/2580), but we'll need to add even more and higher cardinality tags for things like tracing (https://github.com/grafana/k6/issues/2128), to store the trace and span IDs. Finally, k6 has no information whether [user-defined custom tags](https://k6.io/docs/using-k6/tags-and-groups/#user-defined-tags) have high or low cardinality. 

So, because of all of these reasons, we'll need to have some way to differentiate between two distinct types of metric sample tags/metadata:
1. low-cardinality tags that can be used for indexing... so, in the time series mentioned above, in [sub-metric thresholds](https://k6.io/docs/using-k6/thresholds/#thresholds-on-tags), in certain outputs that do aggregation based on tags, etc. 
2. and tags with high cardinality, containing metadata that's often unique for every measurement / data point

Different outputs will be free to treat these high-cardinality tags/metadata however they want - they can discard it (probably the behavior of outputs like statsd, datadod, prometheus, etc. that work with time series), sample it (e.g. to export some tracing information), transform it (e.g. in `influxdb` this will correspond to data point `fields`) or output it directly like they do to tags (e.g. `json`, `csv` and other similar outputs can just write it to the disk, since we don't do any indexing or aggregation, just dump raw data). This will finally allow us to deprecate the hacks we currently have.

The exact API changes on how we'll add this for custom tags in k6 are still under consideration, but the currently favored approach is to extend the tags API to support both strings and objects, something like this:
[...]

PR #2594, opened 2022-07-11
Refactor metric samples to use Atlas for tags and have time series

### LINKED ISSUE #1831: Put metric samples in time series (immutable tag sets) on creation
Currently, k6 builds the set of [tags](https://k6.io/docs/using-k6/tags-and-groups) for every metric measurement it emits in a very ad-hoc manner and as a `map[string]string`. For example, building the tags set for the HTTP metrics happens [here](https://github.com/loadimpact/k6/blob/46d53d99deba98add23079df93579408fa7ea094/lib/netext/httpext/request.go#L232-L243) and [here](https://github.com/loadimpact/k6/blob/46d53d99deba98add23079df93579408fa7ea094/lib/netext/httpext/transport.go#L93-L168). It is a gradual accretion of key-value pairs, based on all of these things:
- the global test run [`tags` option](https://k6.io/docs/using-k6/options#tags)
- the [current `group` tag from the `lib.State`](https://github.com/loadimpact/k6/blob/46d53d99deba98add23079df93579408fa7ea094/js/modules/k6/k6.go#L93-L102), set there by [`group()`](https://k6.io/docs/javascript-api/k6/group-name-fn)
- the `tags` parameter for the specific request [`Params`](https://k6.io/docs/javascript-api/k6-http/params)
- the observed system tag values from the request and response data, modulated by the [`systemTags` global option](https://k6.io/docs/using-k6/options#system-tags)

This is not too terrible and it mostly works, but it's not very efficient either... One big issue is that all of these string maps put a lot of GC pressure on k6 - we generate unique tag maps for every metric sample, and discard them as soon as we've crunched the data for the end-of-test summary and thresholds [in the `Engine`](https://github.com/loadimpact/k6/blob/46d53d99deba98add23079df93579408fa7ea094/core/engine.go#L250) and/or we've sent it to whatever external output is configured. Another performance problem is that we can't efficiently aggregate the metrics back, since that involves heavy string comparisons, both in the outputs and for any sub-metrics we have.

And finally, we don't really know the total number of unique metric tag sets we have in a test run. Load tests should generally have a small number of such, for easier data analysis, that's why we have things like the `name` tag and [URL grouping](https://k6.io/docs/using-k6/http-requests#url-grouping). It would be very good UX if k6 could warn users if they exceed a certain number of unique tag sets in their script, since that almost certainly is a mistake.

To solve all of these issues, I think we can probably use some graph-based data structure to keep track of the various tag-sets that exist. The process of building the tag set for a specific metric sample would be walking through the graph and adding nodes when new tag keys or values are observed that weren't present in the graph. At the end of that tag accretion process, instead of a string map, the metric sample would contain a pointer to an immutable node from the graph. And every other metric sample with the same tags will have the exact same pointer. So, comparing metrics (for aggregation in outputs or sub-metric partitioning) should become as easy as comparing the pointers of their tag set objects.

The exact data structure we want to use and how we'll ensure thread-safety (given that tag sets can be generated concurrently by multiple VUs) is something that still needs to be determined... Some sort of an [STM](https://en.wikipedia.org/wiki/Software_transactional_memory) approach seems better than adding locks to every graph node, but I'm not sure how realistic that would be for Go... There seem to exist some libraries for it (e.g. https://github.com/anacrolix/stm), but it needs a lot of further evaluation.

Also, it's not certain how much we can optimize the [`Contains()`](https://github.com/loadimpact/k6/blob/46d53d99deba98add23079df93579408fa7ea094/stats/stats.go#L228-L243) check if a tag set is a subset of another tag set, used for [sub-metric updates](https://github.com/loadimpact/k6/blob/46d53d99deba98add23079df93579408fa7ea094/core/engine.go#L416-L418). It won't be slower, given that the graph nodes will probably contain the full tag sets, but it might not be possible to make it faster, if the alternative is a BFS of the graph. This depends solely on the data structure and algorithms we pick.

Alternatives to graphs should also be explored. We might be able to achieve most of the same benefits by using clever hashing of the tag set values. Or even having a global set with values being the concatenated `key=value` pairs, so that each metrics sample tag sets is a subset of the global one. For sure, before we start anything here, we should see how other projects (e.g. Prometheus) deal with similar issues in an efficient way...

Finally, it's not clear if we should try to enforce an upper limit to the number of unique tag sets. Personally, I think not - we should emit a very stern warning of some number is exceeded, say 1000 unique tag set combinations, but not try anything more fancy than that (like falling back to stringmap based tag sets).

codebien:
Integrated https://github.com/mstoykov/atlas as the internal data structure for `SampleTags`. Atlas is a Btree implementation based on `[2]string` array for the nodes (the stored tags pairs). It uses pointers most of the time for comparisons so it's very fast and read locks can be avoided.

It allows allocating a specific Tag pair just once. The reduction in terms of memory from my basic benchmarks is very important. For example, the benchmark on a[ similar function](https://github.com/grafana/k6/blob/bfa415c5fdfa6084f49b02877540cf6c05e23605/output/json/json_test.go#L57-L92) reduces the overall allocated memory from 600MB to 87MB (note: this is a very extreme and limit case, the average is not so high).

Also, the time series concept has been introduced as defined from the Data model and Storage parts in #2580.

Closes #1831 #2580


metrics/sample.go
```diff
@@ -282,25 +283,24 @@ func (st *SampleTags) CloneTags() map[string]string {
...
-		res[k] = v
-	}
-	return res
+	return st.n.Path()
 }
 
+// rootTagNode is the global Node.
+// TODO: move to a not global place if consensus is achieved
+var rootTagNode = atlas.New() //nolint:gochecknoglobals
+
```
codebien, line 292: Is the `metrics.Registry` again a good place to store the root?

metrics/sample.go
```diff
@@ -179,7 +180,7 @@ func parsePercentile(stat string) (float64, error) {
 // All methods should not panic, even if they are called on a nil pointer.
 //easyjson:skip
 type SampleTags struct {
-	tags map[string]string
+	n    *atlas.Node
```
oleiade, line 183: I would recommend to keep this attribute named `tags` for clarity. I personally find it hard to follow code when anything other than a local variable is named with a single letter. Adding a documentation comment could be useful to explicit what to expect from an `*atlas.Node` too?

oleiade (commented):
I'm aligned with @na-- remarks. Also feeling somewhat uncomfortable with merging code using a "POC" library in `master`.

mstoykov:
> Hmm I'd definitely want to see more benchmarks and tests before we merge something like this

We are definittely not just merging this. I would even expect that we would rewrite most of the code to actually not have a global root, but have it in the metrics.Registry before we even think of merging. 

I would also expect at least some of the code to actually use the root node directly (the http path definitely) instead of `NewSampleTags` as that will also be a lot more performant then creating a map and then making it into an atlas node.

> Considering we'll have just a single copy of the tags per TimeSeries


How do we build and have only one copy and keep it only one copy? From the code that I have seen so far - we always build a new slice of tags and then we hash it and go get it from the registry. But we still made a new copy - we just threw it away in the mean time. 

> Besides, I have just skimmed the atlas code, but I am not sure it works like how we'd need it to work. A tag set with keys {a: b, c: d} should be equivalent to a tag set with the same keys in a different order, {c: d, a: b}, which I don't think is the case now. 

It should, but you can test it and open an issue if it doesn't ;)  


To be honest I would expect the reduction in memory to be negligable for most real uses case, but the reduction in GC time in those same cases should be significant once we are using this directly building from [lib.State#Tags](https://pkg.go.dev/go.k6.io/k6/lib#State) instead of going from the root node up for each set.

One of the points of this structure is that it "memorizes" how it was constructed and optimizes the same route for the next time. So as long as the code that generates it takes the same route it will be faster the following times. But building it from a map has more chances of it being randomized. 

I still would like to see if even like that it makes a difference and in which direction if you just have a fairly basic test. I would expect thsi will be slightly ... worse, before we stop generating the whole map and instead use the atlas.Node for that as well

codebien:
@na-- @mstoykov @oleiade I pushed a new iteration with a basic Atlas integration. Please, take a look and let me know your opinions. I expect several points of discussion so I didn't complete the entire migration, in any case, the `js/http` package should work as expected.

metrics/tagindex.go
```diff
@@ -0,0 +1,132 @@
...
+		mutex: sync.RWMutex{},
+		locks: false,
+	}
+}
+
+// NewTagMap creates a TagMap,
+// if a not-nil map is passed then it will be used as the internal map
+// otherwise a new one will be created.
+//
+// TODO: this should be removed from the code asap
```
mstoykov, line 36: Mark is [deprecated](https://rakyll.org/deprecated/)

js/modules/k6/metrics/metrics.go
```diff
@@ -128,14 +128,19 @@ func (m Metric) add(v goja.Value, addTags ...map[string]string) (bool, error) {
 		return raiseNan()
 	}
 
-	tags := state.CloneTags()
+	tags := state.Tags.BranchOut()
 	for _, ts := range addTags {
 		for k, v := range ts {
-			tags[k] = v
+			tags.AddTag(k, v)
```
mstoykov, line 134: Nit: this will be bad with a *large* amount of `addTags` which arguablt is *really* rare -  have only seen being used with a single tag ever.

A potential fix here will be to get the original js object which has the keys ordered the way they were provided and use that order to only walk the same path in the atlas.

Again I have only seen this being used with 1 tag and ... I would not expect anything noticeable before we are above at least 5-6 tags, probably more like 10+  both of which seem really unlikely to me

js/modules/k6/ws/ws.go
```diff
@@ -247,17 +247,18 @@ func (mi *WS) Connect(url string, args ...goja.Value) (*WSHTTPResponse, error) {
 
 	if state.Options.SystemTags.Has(metrics.TagIP) && conn.RemoteAddr() != nil {
 		if ip, _, err := net.SplitHostPort(conn.RemoteAddr().String()); err == nil {
-			tags["ip"] = ip
+			// TODO: make as a not indexable
```
mstoykov, line 250: I guess @na-- needs to weight in on what the API will be for that and whether this current API is good

mstoykov: Also maybe have the same comment for the `url` as well?!?

lib/netext/grpcext/conn.go
```diff
@@ -196,25 +197,28 @@ func (statsHandler) TagRPC(ctx context.Context, _ *grpcstats.RPCTagInfo) context
 func (h statsHandler) HandleRPC(ctx context.Context, stat grpcstats.RPCStats) {
 	state := h.vu.State()
 	tags := getTags(ctx)
+	if tags == nil {
+		// FIXME: this is an RPC for reflection
+		// find how to handle it properly
```
mstoykov, line 202: as far as I remember - we just emitted with empty tags for the reflect https://github.com/grafana/k6/pull/2160#pullrequestreview-784595048 (and all the other discussion in that PR)

lib/netext/httpext/request.go
```diff
@@ -187,17 +187,17 @@ func MakeRequest(ctx context.Context, state *lib.State, preq *ParsedHTTPRequest)
 		}
 	}
 
-	tags := state.CloneTags()
+	tags := state.Tags.BranchOut()
 	// Override any global tags with request-specific ones.
 	for k, v := range preq.Tags {
-		tags[k] = v
+		tags.AddTag(k, v)
```
mstoykov, line 193: nit: same nit as in the case in k6/metrics - this potentially can be really bad with a lot of tags and them being walked randomly.

metrics/tagindex.go
```diff
@@ -0,0 +1,132 @@
...
+// FIXME: naming! we now have TagMap, SampleTags, SystemTag, TagSet
+// so the naming decision and maybe some sort of uniquification could be good.
+//
+// TagSet should be the best but is already used, TagIndex could be a valid alternative?
+
+// TagMap represents a set of tags.
+type TagMap struct {
+	mutex sync.RWMutex
+	tags  *atlas.Node
+	locks bool
```
mstoykov, line 18: Can we get some comment explaining why this exists?

vendor/github.com/mstoykov/atlas/atlas.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+	// TODO: https://github.com/mstoykov/atlas/issues/2
+	// apparently this is faster than if n.linkKey == sub.linkKey
+	if n.linkKey[0] == sub.linkKey[0] && n.linkKey[1] == sub.linkKey[1] {
+		return n.prev.Contains(sub.prev)
+	}
+	return n.prev.Contains(sub)
+}
+
+func (n *Node) Len() int {
```
mstoykov, line 96: Something is wrong here as this is define twice

metrics/tagindex.go
codebien: It's a wrapper for the tag set concept around Atlas, essentially hiding the `n = n.AddLink()` logic behind an `AddTag` method, From/To SampleTags and the concurrency handling when required. It is the same as SampleTags but mutable.

The concurrency feature it's mostly required for removing the TagMap from [the lib](https://github.com/grafana/k6/blob/521a2823afd0018cb8d9f8a2931c18dd63ad0010/lib/state.go#L97-L101) package and use a unique struct.

So a comment could be like the following:

```suggestion
	// TagMap represents a ordered set of tags, it's expected to be initialized once
	// with an empty or a root set of tags and use BranchOut for generating from it all the new tag set,
	// in this way, TagMap can optimize and reduce as much as possible the memory allocation.
        // By default a TagMap is not concurrent safe, check the `SetSafeConcurrency` method to get it.
        type TagMap struct {
	  mutex sync.RWMutex
	  tags  *atlas.Node
	  locks bool
```

na--: I am not sure which is better, this property or separate types :thinking:  Let me see if I understand things correctly:
1. we don't have to account for any concurrency when building the atlas `Node` that has `--tags` (the de-facto root of the test run)
2. we also don't have any concurrency when setting the `Node` that corresponds to the initial `scenario.tags` in every VU
3. we need thread safety when dealing with VU tags during execution, both because `lib.State.Tags` can be changed from `k6/execution`, and because we can have extra metric tags everywhere (custom metrics, HTTP requests, etc.), however:
    1. we only need to ensure thread safety when we access the VU tags, once we branch out of them, we don't need any locking
    2. the actual building of the tags (e.g. for a single HTTP request) is sequential, tag by tag
   
Am I missing something? Because, to me, it seems like we need only very little locking in the few places that read and write the current VU tags in `lib.State`? I think everything else besides that can just depend on atlas' internal locking? :thinking:

cmd/options.go
```diff
@@ -235,18 +235,19 @@ func getOptions(flags *pflag.FlagSet) (lib.Options, error) {
...
-			var name, value string
-			name, value, err = parseTagNameValue(s)
-			if err != nil {
-				return opts, fmt.Errorf("error parsing tag '%s': %w", s, err)
-			}
-			parsedRunTags[name] = value
+	// Atlas initialization where the root node is created.
+	// All the new tagsets must branch out from the root,
+	// so the underhood structure can minimize the memory footprint allocating the nodes only once.
+	tagsRootNode := metrics.RootTagMap()
```
na--, line 241: I think having this dynamic atlas-based thing in `lib.Options` might be a mistake, or at least an over-complication we don't need :thinking: We only need the option validation and string values here, a `map[string]string` (or some sorted wrapper around one) should be plenty good enough for the Options, we just need to validate the values at this point... The atlas integration and its root node seems like it should be stored somewhere outside of the Options, in some sort of a test state? :thinking: 

So, on Thursday, when I started reviewing this, I though this place was the already existing `lib.RuntimeState` that was initialized in `cmd/`, `loadedTest.initializeFirstRunner()`, where we also initialize the `metrics.Registry` and the built-in metrics: 
https://github.com/grafana/k6/blob/f2684ad90fb7cfc8e02cfa776d303e3c9df10f83/lib/runtime_state.go#L10-L18
https://github.com/grafana/k6/blob/f2684ad90fb7cfc8e02cfa776d303e3c9df10f83/cmd/test_load.go#L112-L117

That gets propagated in the `js.Runner` (though it probably needs a bit of refactoring to actually save the whole object instead of deconstruct it): https://github.com/grafana/k6/blob/f2684ad90fb7cfc8e02cfa776d303e3c9df10f83/js/runner.go#L109

Unfortunately, this gets initialized before we have the consolidated and derived options, and we need that to get the final value of `--tags` :sob: https://github.com/grafana/k6/blob/f2684ad90fb7cfc8e02cfa776d303e3c9df10f83/cmd/test_load.go#L94

I originally thought "Oh, #883 strikes again :sob: ", but then reconsidered :sweat_smile: This isn't a matter of configuration, precisely, it's a matter of some state that we can't propagate to the rest of the test because of our poor code structure :disappointed: So... I kind of went down that rabbit hole and fixed the problem, and will submit a refactoring PR shortly (just fixing some final tests).

metrics/tagindex.go
codebien: >Am I missing something? Because, to me, it seems like we need only very little locking in the few places that read and write the current VU tags in lib.State? I think everything else besides that can just depend on atlas' internal locking?

Your description is right, there aren't missing points :+1: 

My rationale behind it: I thought it could increase the complexity of our project structure to have one more struct doing the same thing (+ locks). I found moving the locks on `metrics.TagMap` and removing the `lib.TagMap` a convenient simplification of our Tag representations. I thought also that it could be useful in the future, now that we have the event loop and concurrent operations within iterations (but I don't have a concrete use-case to point for that).

na--: That's the thing though - you might have concurrent operations, you can kind of have them even now (e.g. with `http.batch()` or websockets), but when we are measuring something and building the tag set, the only part where we need to do locking is getting the current VU's tags. After we get them, the building of the tag set for the current measurement is actually sequential. 

For example, we have already done the HTTP request, and we have already measured it, now we are just transfering these observations and measurements to some other object, but we do this perfectly sequentially and without any need for locking (besides the atlas internals, just because that's how it works). That's why we don't have any locking here now: https://github.com/grafana/k6/blob/f2684ad90fb7cfc8e02cfa776d303e3c9df10f83/lib/netext/httpext/transport.go#L95

cmd/options.go
codebien: >We just need to validate and store the values at this point, not start building the root atlas Node... The atlas integration and its root Node seems like it should be stored somewhere outside of the Options, in some sort of a test state?

I already have this version in local with a different approach, sorry if I haven't yet pushed it but the last transition (`output/cloud/...`) takes time.

Instead of using the `RuntimeState`, I added a method on the Registry for getting the root TagMap. I did it in this way so it wasn't required to fix more tests initializing a new dependency. For example, [in this method ](https://github.com/grafana/k6/blob/f2684ad90fb7cfc8e02cfa776d303e3c9df10f83/metrics/engine/engine.go#L89-L92) we need the root for initializing a new `SampleTags` for the Submetric and the Registry is already available in the metrics engine.
We can consider having a dedicated field in the RuntimeState if we prefer.

na--: hmm, why do we need the root atlas Node to create a sub-metric? :confused:

codebien: We have a [SampleTags initialization](https://github.com/grafana/k6/blob/f2684ad90fb7cfc8e02cfa776d303e3c9df10f83/metrics/metric.go#L102-L108) in AddSubmetric.

na--: Ah, I see, since you're using [`Node.Contains(Node)`](https://github.com/mstoykov/atlas/blob/23cc720eb639c314be5df4442d8567969c63fc06/atlas.go#L101) to see if a time series' tags contain the sub-metric tags, they need to have a common root. For that, I guess storing it in the `metrics.Registry` is great :+1:  :+1: 

But I'd suggest against storing the `Node` that results from applying the `--tags` there. Let me publish my refactoring PR today and you'll see why I think there is a better place.

na--: So, this is my proposal: https://github.com/grafana/k6/pull/2627 _(most of the diff is just fixes in the tests, the actual code changes are fairly minor)_

While the base `atlas.New()` root `Node` can sit in the `metrics.Registry` (though it can also be in the `TestPreInitState` :man_shrugging: ), the Node that is the result of `--tags` can be saved in the `TestRunState` object. The executors will then be able to build on top of that and add the `scenario` tag, as well as any `tags` that were explicitly specified in the scenario config. I think all of this won't need any `SampleTags` or locking, just simple atlas nodes in `TestRunState` and the scenarios, and simple string:string map for `lib.Options` (or a simple sorted wrapper without any locking).

metrics/tagindex.go
codebien: >but when we are measuring something and building the tag set, the only part where we need to do locking is getting the current VU's tags. After we get them, the building of the tag set for the current measurement is actually sequential.

This is how the `metrics.TagMap` implementation works, the default behaviour doesn't use locks, it isn't concurrent safe. `SetConcurrentSafe` has to be invoked if safety is required and in the current implementation is invoked **only** for the `lib.State.Tags`.

codebien: If this is creating confusion we can split them, but one will be the replica of the other + locks.

na--: > If this is creating confusion we can split them, but one will be the replica of the other + locks.

I'd prefer that approach - it's a fairly small data structure. 

I think the type with locking will only be needed in a single place, in the [VU's `lib.State.Tags`](https://github.com/grafana/k6/blob/1e8468292ab5eb0d2333b1278b44e81c5cbef679/lib/state.go#L76), right? And besides, it won't be exactly the same type - keep in mind we won't just need to lock the current atlas `Node` in the VU when we want to build and emit a `Sample`, we'll also need to lock the unindexable tags (https://github.com/grafana/k6/issues/2584) that are currently set in the VU's `State`. In my now hopelessly conflicted and obsolete branch, I didn't have atlas, but I basically continued having `lib.State.Tags` to be its own type with internal locking, just extended its `Clone` method.

cmd/options.go
mstoykov: While I agree with @na-- on how the atlas node/tags should be handled and kept (I personally only thing you need the non read-only AddTag in places where you are emitting metrics) I don't see why we need to refactor such a huge part of k6 to do it. Especially given that it's the summer and a lot of us will be on/off quite a bit and there are at least 3 big changes in the work which will definitely need to be rebased on that quite big PR ...

na--: I didn't intend the PR to become so big in terms of LoC, but that is just because of the tests... The actual changes in the non-test code are fairly few and on the small side, basically no logical changes anywhere, just moving things around.

codebien:
Updates:
* Restored `lib.State.Tags` and renamed `metrics.TagMap` to `metrics.TagSet`
* `metrics.TagSet` is now always lock free
* `lib.RunTags` from SampleTags to `map[string]string`
* Root set in PreInitState
* Resolved RunTags set in `TestState`

Note: as you can see from the failing xk6 test, we are breaking the extensions removing the ability to create a `SampleTags` from a map.

lib/test_state.go
```diff
@@ -14,20 +15,47 @@ type TestPreInitState struct {
...
+	// FIXME: remove when we have a stable version
+	// so we don't have to fix the tests multiple times
+	if s.RootTagSet == nil {
+		s.RootTagSet = metrics.NewTagSet(nil)
+	}
+
+	tags := s.RootTagSet.BranchOut()
+	if len(opts.RunTags) < 1 {
+		return tags
+	}
```
na--, line 35: hmm, I noticed a couple of things here that can maybe be improved, though I am not very sure about that :thinking: 
1. `RootTagSet` might be better off in the `Registry`, as I think you originally said it's going to? be :sweat_smile:  `TestPreInitState` is nice, but you kind of have to initialize its properties in every test :disappointed: Whereas if you add it to `metrics.NewRegistry()`, that will be automatic everyhere, and we'll be even more sure that we have just a single root `Node`. As an extra precaution to ensure that, we can make the root atlas `Node` unexported and have a `GetTagsRoot() *metrics.TagSet` method that returns a new `*metrics.TagSet` wrapper object with the same root node. That way we will have a bulletproof way to ensure we always have exactly one single immutable root.
2. Since `RunTags()` doesn't use all of the `lib.Options` properties, it probably doesn't need to have that as its argument, just a `map[string]string`.
3. Connected to both of the previous points, `RunTags()` also doesn't seem like it should be a method of the `TestPreInitState` :thinking: It can just be a method of `metrics.TagSet`, something like `func (tg *TagSet) BranchOutWith(tags map[string]string) *TagSet { ... }`?

metrics/tags.go
```diff
@@ -5,8 +5,83 @@ import (
...
+	tags *atlas.Node
+}
+
+// NewTagSet creates a TagSet, if a not-nil map is passed
+// then it adds all the pairs to the set, otherwise an empty set is created.
+// Under the hood it initializes an Atlas root node.
+// It should be only used in a centralized place and
+// all the new TagSet should branch from it for getting
+// the optimal performances.
+func NewTagSet(m map[string]string) *TagSet {
```
na--, line 23: Connected to my other comment (https://github.com/grafana/k6/pull/2594/files#r936302703), I kind of think this function shouldn't exist at all, or if it does exist, it should definitely be unexported and used privately only in a few places in the metrics package :thinking: 

Otherwise, if it exists in its current state, it's going to be misused, since people might not realize that comparing tag sets will fail if they don't share the same root. Even it's current comment is a bit misleading, it says "all the new TagSet should branch from it for getting the optimal performances." when it should probably say "all the new TagSets should branch from it or everything will be broken" :sweat_smile:

js/modules/k6/ws/ws.go
na--: Forgot to reply here. The basic idea is this:
- we'll continue storing these non-indexable tags as a plain Go `map[string]string` in `lib.State.Tags`, i.e. a new field in the `lib.TagMap` struct currently called `metadata`, for historical reasons... :sweat_smile: 
- branching out will just copy it, though it should be `nil` most of the time, unless someone explicitly added a non-indexable tag through `vu.tags` in `k6/execution`
- per-testrun `--tags` and per-scenario `tags` are, by definition, indexable, since they will apply to all of the generated metrics, i.e. have super-low cardinality :sweat_smile: so they are always going to be kept just in atlas, non-indexable tags only come to play on the VU level
- we know which system tags will be non-indexable (url, error, vu, iter) so, if they are enabled, just add their values in that map instead of in atlas
- we also add any per-measurement (i.e. per HTTP request, per `add()`, etc.) custom non-indexable metrics to the map instead of the atlas node

After this PR stabilizes a bit, I'll try to rebase my changes on top of it and make another PR to show what I mean.

lib/test_state.go
codebien: >func (tg *TagSet) BranchOutWith(tags map[string]string) *TagSet { ... }

:thinking: Are you sure adding this method doesn't lead to a bad memory usage where we allocate maps when they are not a requirement and iterate using the range pseudo-random order?

I think having only `BranchOut` +  iteration with `AddTag` will force us to do it in an optimal way.

codebien: Maybe an alternative could be to add `BranchOutWith` to the `registry` instead to add it to the `TagSet`.

codebien: Check the latest version

output/cloud/output.go
```diff
@@ -383,24 +384,33 @@ func (out *Output) AddMetricSamples(sampleContainers []metrics.SampleContainer)
...
+				}
+
 				newSamples = append(newSamples, &Sample{
 					Type:   DataTypeSingle,
 					Metric: sample.Metric.Name,
 					Data: &SampleDataSingle{
 						Type:  sample.Metric.Type,
 						Time:  toMicroSecond(sample.Time),
-						Tags:  sample.Tags,
+						Tags:  encodedTags,
```
mstoykov, line 413: Is this needed - I am pretty sure this will be done automatically ? Or does it not get done automatically?

js/modules/k6/metrics/metrics.go
codebien: Fixed in https://github.com/grafana/k6/pull/2632 and rebased here.

output/cloud/output.go
codebien: I've changed the type from `SampleTags` to `json.RawMessage` otherwise we can't use the unmarshal function, we have to unmarshal to a temporary type and manually resolve using the Atlas root.

lib/netext/grpcext/conn.go
codebien: https://github.com/grafana/k6/pull/2634

metrics/registry.go
```diff
@@ -89,3 +99,32 @@ func (r *Registry) MustNewMetric(name string, typ MetricType, t ...ValueType) *M
...
 }
+
+// BranchTagSetRoot creates a new TagSet starting from the root.
+func (r *Registry) BranchTagSetRoot() *TagSet {
+	return &TagSet{tags: r.tagSetRoot}
+}
+
+// BranchTagSetRootWith creates a new TagSet starting from the root and
+// adds all the key-value pairs provided with the map to the new TagSet.
+func (r *Registry) BranchTagSetRootWith(m map[string]string) *TagSet {
```
na--, line 110: I have a few of concerns about these names:
1. "TagSetRoot" is unfortunately very easy to misread as `Branch_Tag_SetRoot`, i.e. set the root of the branch tag (whatever that means), not as `Branch_TagSet_Root`... So I suggest replacing `TagSet` with just `Tags`, or changing the order from `TagSetRoot` to `RootTagSet`
2. Branch is most often a noun, not a verb, so "BranchTag..." is a bit confusing. The `BranchOut()` method in `TagSet` is much clearer :+1:

Finally, I am not sure the second method, `BranchTagSetRootWith()` (however it's named), should be in the `Registry`, it isn't in any way specific to the root, right? Why not have it just in the `metrics.TagSet`?

Here are my concrete suggestions. Rename the `BranchTagSetRoot()` method to `RootTagSet()` and move the second method to `TagSet` and rename it to `BranchOutWith()`
```suggestion
// RootTagSet creates and returns a new TagSet, starting from the common global root.
func (r *Registry) RootTagSet() *TagSet {
	return &TagSet{tags: r.tagsRoot}
}

// BranchOutWith returns a new TagSet by branching out from the receiver
// and adding the key-value pairs provided with the map to the new TagSet.
func (tg *TagSet) BranchOutWith(m map[string]string) *TagSet {
```

It will make some one-liners in the tests from this PR and https://github.com/grafana/k6/pull/2633 a bit longer, but I'd argue it will also make them more readable. For example, these lines:
```go
piState.Registry.BranchTagSetRootWith(options.RunTags)

piState.Registry.BranchTagSetRootWith(map[string]string{"a": "1"}).SampleTags(),
```

will become these:
```go
piState.Registry.RootTagSet().BranchOutWith(options.RunTags)

piState.Registry.RootTagSet().BranchOutWith(map[string]string{"a": "1"}).SampleTags(),
```

lib/netext/grpcext/conn.go
codebien: #2634 has been merged so this has been resolved.

metrics/tagindex.go
codebien: This is same as https://github.com/grafana/k6/pull/2594#discussion_r936304976 and it will be fixed with https://github.com/grafana/k6/pull/2633

output/cloud/output.go
mstoykov: Unmarshalling the clout output metrics I think is only done in tests?

metrics/registry.go
na--: hmm or maybe even `TagsRoot()`:
```go
piState.Registry.TagsRoot().BranchOutWith(options.RunTags)

piState.Registry.TagsRoot().BranchOutWith(map[string]string{"a": "1"}).SampleTags(),
```
:thinking:

mstoykov: `TagsRoot` seems nicer

codebien: >Here are my concrete suggestions. Rename the BranchTagSetRoot() method to RootTagSet()

It is fine to me to have a better name but I think the `BranchOutTagsRoot()` or in a different order is better, because it makes it visible that it is already branched out and you don't need to branch after. Because with `RootTagSet()` it seems legit to think I have to do `RootTagSet().BranchOut()`.

```go
piState.Registry.RootTagSet().BranchOutWith(options.RunTags)
```

Also, for the same reason, we are allocating `metrics.TagSet` twice for `RootTagSet()` and for `BranchOutWith(options.RunTags)`. We could resolve this by storing a centralised `metrics.TagSet` instead of the `atlas.Node` but we said we prefer the node for safety.

>Finally, I am not sure the second method, BranchTagSetRootWith() (however it's named), should be in the Registry, it isn't in any way specific to the root, right? Why not have it just in the metrics.TagSet?

>and move the second method to TagSet and rename it to BranchOutWith()

:-1: I think this will increase the risk we use TagSet in a bad form. I think it's root specific logic because the good case where we should use a `BranchOutWith` is from the root. For the rest, we should design the code in `branch := TagSet.BranchOut(); branch.AddTag(k,v)`. In this way, it forces us to think about to have a consistent order and reusing the order from the passed structure (e.g like we did for the js modules reusing the `goja.Object`).

metrics/sample.go
```diff
@@ -179,7 +180,7 @@ func parsePercentile(stat string) (float64, error) {
 // All methods should not panic, even if they are called on a nil pointer.
 //easyjson:skip
 type SampleTags struct {
-	tags map[string]string
+	tags *atlas.Node
 	json []byte
 }
```
mstoykov, line 185: This enforces an additional allocation, for the SampleTags, object even if `tags` is "reused". 

There are currently two uses for this type:
1. caching `json`
2. Helps with the creation so that you cna use `AddTag` on the same object.

Arguably some encapsulation of `atlas.Node` although that likely isn't so intersting.

I would argue `2` is less of an issue and it can be solved with a type that can be allocated on the stack and then to retunr the `atlas.Node` that is then send away. How well that will work is to be seen - the k6 escape analysis might not work great on this case.

I would argue hte `json` caching likely is ... uh at best overrated. Most tags are on single samples either way so the cache won't be hit. (The cloud output outputs all 8 http metric sampels with 1 tag). 

This also totally doesn't take advantage of the fact that the vast majority of hte `atlas.Node` will be the same and consequently the `json` it produce will be as well.

Given That I kind of prefer if we drop that and move it to a `map[*atlas.Node][]byte` per each output that needs it. We can even have it as a helper type I guess.

We get:
1. less allocation per each sample tags we need - arguably what we wanted
2. an even more optimized tag marshalling to json or w/e actually as we can use the same thing for the csv marshal

mstoykov: Actually ... we don't even use this for the creation of the node in the first place ...

So yeah it seems quite bad at thsi point. 

Arguably it's keeping some backwards compatibility ... but I am pretty sure we still breaking any usage of this type anywhere either way.

metrics/registry.go
na--: Ah, very good points :+1: :man_facepalming: I am not too worried about the performance implications of branching twice from the root, since that will only happen a few times in the actual k6 run, but you are right that it's a bad example. And, while you are right that the only good case for `BranchOutWith` is from the root, I kind of still don't think this method should be in the Registry :thinking: 

So, new suggestion :sweat_smile: 
```suggestion
// TagsRoot creates and returns a new TagSet, starting from the common global root.
func (r *Registry) TagsRoot() *TagSet {
	return &TagSet{tags: r.tagsRoot}
}

// SortAndAddTags sorts the given tags by their keys and adds them to the 
// current tag set one by one, without branching out. This is generally 
// discouraged and sequential usage of the AddTag() method should be 
// preferred and used whenever possible. 
// 
// The only place this method should be used is if we already have a 
// map[string]string of tags, e.g. with the test-wide --tags from the root,
// with scenario.tags, with custom per-reqeust tags from a user, etc.
// Then it's more efficient to sort their keys before we add them. If we don't,
// go map iteration happens in pseudo-random order and this will generate
// a lot of useless dead-end atlas Nodes.
func (tg *TagSet) SortAndAddTags(m map[string]string) *TagSet {
```

metrics/tags.go
```diff
@@ -5,8 +5,83 @@ import (
...
+	return tmcopy
+}
+
+// SampleTags creates a SampleTags using the current tag set.
+func (tg *TagSet) SampleTags() *SampleTags {
+	st := &SampleTags{}
+	st.tags = tg.tags
+	return st
+}
+
```
mstoykov, line 84: Connected to https://github.com/grafana/k6/pull/2594/files#r937514286

Can't we have it retunr the atlas.Node and a "constructor" to basically take that :thinking: 

If we decide to drop SampleTags (or at least make it be just *atlas.Node I guess) that will be sufficent and might actually be easier to work with :thinking: 

The `branchFromRoot.....` will actually be `metrics.TagSetFromSample(registry.RootTagsNode())`

metrics/sample.go
na--: > I would argue the json caching likely is ... uh at best overrated. 

Yeah, once this TODO is implemented, which would be super easy to do, marshaling an atlas node to JSON should be super easy: https://github.com/grafana/k6/blob/44b62ba4144079459815abc100fceb833f022532/metrics/sample.go#L245-L251

However, I suggest we can keep the `SampleTags`, for now. I am fine with getting rid of the JSON caching, but I am using them for the unindexable tags. Remember, a metric sample will have the atlas Node (for the indexed tags) and a `map[string]string` for the high-cardinality unindexed ones. It's easiest to keep `SampleTags` as a place to put both, otherwise, I will also need to change the other types like `ConnectedSampleContainer` and `ConnectedSamples` here.

mstoykov: Hmmm maybe it makes more sense to have the unindexable "tags" in a separate struct either way :shrug:

na--: Why? :confused: It makes perfect sense for me to reuse `SampleTags`:
 - it will contain all of the "tags" for a "sample", both the indexed and the unindexed ones, I can't think of a better name
 - it will wrap the TagSet and `map[string]string` and either prevent their modification after a Sample altogether, or at least heavily discourage it

I mean, the type itself will be substantially changed - I will need to adjust some of the methods and add new ones, and also need to change how outputs work with it for unindexable tags. But I'll need a custom type and I can't think of a better name for it, so why not just reuse and change `SampleTags`? :sweat_smile: It will also narrow the scope of the things @codebien needs to fix today, since I will take care of it next week.

na--:
@codebien, @mstoykov and I had a call to discuss some of the important details in this PR and here's what we decided:
1. We won't have a `metrics.TagSet` with an internally mutating `atlas.Node`, instead we'll have an immutable `TagSet` that is redefining `atlas.Node` like this:
	```go

	type TagSet atlas.Node

	func (tg *TagSet) Add(name, value string) *TagSet {
		return (*TagSet)(((*atlas.Node)(tg)).AddLink(name, value))
	}

	func (tg *TagSet) Delete(name string) *TagSet {
		return (*TagSet)(tg.Delete(name))
	}
	
	// ... + wrappers for other atlas.Node methods that return a *Node or have unsuitable names
	// ... + JSON methods from SampleTags
	```
2. We'll completely get rid of `metrics.SampleTags`, the JSON marshaling/unmarshaling methods will be added to `metrics.TagSet` and the `metrics.Sample` struct will look somewhat like this (after [non-indexable tags](https://github.com/grafana/k6/issues/2584)):
	```go
	type Sample struct {
		Metric   *Metric
		Time     time.Time
		Tags     *TagSet
		Metadata map[string]string // could be nil
		Value    float64
	}
	```
3. Naming things in the `metrics.Registry` will be simpler because they are now immutable, e.g.:
	```go
	type Registry struct {
		metrics map[string]*Metric
		l       sync.RWMutex

		rootTagSet *atlas.Node
	}

	func NewRegistry() *Registry {
		return &Registry{
			metrics: make(map[string]*Metric),
			// All the new TagSts must branch out from this root, otherwise
			// comparing them and using their Equals() method won't work correctly.
			rootTagSet: atlas.New(),
		}
	}

	// ...

	func (r *Registry) RootTagSet() *TagSet {
		return (*TagSet)(r.rootTagSet)
	}
	```
4. If we have a lot of tags being created somewhere, e.g. when [emitting the HTTP metrics](https://github.com/grafana/k6/blob/8f981bb3b1e1e996675187615aca5f6480a5a600/lib/netext/httpext/transport.go#L98-L201), we can easily create accumulator-like helpers to help with that task. Whether we export them or make them global will be decided on a case-by-case basis. For example, there is probably some benefit of having a helper with the system tags in the above situation:
	```go
	tags := NewSystemTagAccumulator(t.tags, t.state.Options.SystemTags)
	// ...
	tags.AddSystemTag(metrics.TagMethod, unfReq.request.Method)
	// ...
	tags.AddSystemTag(metrics.TagStatus, unfReq.response.StatusCode)
	// ...
	```

@mstoykov, @codebien, am I missing anything?

metrics/registry.go
na--: We had a video call to resolve this and the notes are in https://github.com/grafana/k6/pull/2594#issuecomment-1205273413

metrics/tags.go
na--: We had a video call to resolve this and the notes are in https://github.com/grafana/k6/pull/2594#issuecomment-1205273413

metrics/sample.go
na--: We had a video call to resolve this and the notes are in https://github.com/grafana/k6/pull/2594#issuecomment-1205273413

na--:
As an addendum to the previous note, we realized that we can make a tiny change to `metrics.Sample` and group the `Metric` and `Tags` fields into their own simple `TimeSeries` struct and use that in `Sample`, like this:
```go
type TimeSeries struct {
    Metric *Metric
    Tags *TagSet
}

type Sample struct {
	TimeSeries
	Time     time.Time
	Metadata map[string]string // could be nil
	Value    float64
}
```

This doesn't change anything significant and the `TimeSeries` type is a plain naked struct without any methods. However, it will allow us to go a long way towards addressing https://github.com/grafana/k6/issues/2580. Because of the way comparison works in go, you'd be able to use `==` to compare the `TimeSeries` values of 2 metric Samples, or use the them as map indexes. From https://go.dev/ref/spec#Comparison_operators
>  Pointer values are comparable. Two pointer values are equal if they point to the same variable or if both have value nil. Pointers to distinct zero-size variables may or may not be equal.
> ...
>  Struct values are comparable if all their fields are comparable. Two struct values are equal if their corresponding non-blank fields are equal. 

We might still have a central registry for TimeSeries in the future (e.g. to be able to count exactly how many unique ones we have), but if we do this small change now, it's likely we won't need to make another breaking change to `metrics.Sample` in the future (:crossed_fingers: :sweat_smile:).

na--:
FYI we almost inadvertently introduced a VU activation bug with the migration to atlas. Changes to `vu.tags` (from `k6/execution`) shouldn't persist between scenarios, and they don't in `master`, we get our activation base tags from the main test's `runTags`: https://github.com/grafana/k6/blob/5b70e70b657750234651a337c1b736c177926855/js/runner.go#L613
but we changed it to get it from the `VU.State`: https://github.com/grafana/k6/blob/6fbde0ab9a35dd6520b83acbfd482477438e2147/js/runner.go#L616
and I perpetuated the problem in https://github.com/grafana/k6/pull/2638 with https://github.com/grafana/k6/blob/e1a641a1aa0c82d1d2f8db77611ccf4ba9bf301b/js/runner.go#L638-L639

Noticed this when I was doing the unindexable tags and I've fixed it and added a test and an explicit comment, see the [latest force-pushes](https://github.com/grafana/k6/compare/fb6411c2a7b4b42c9dd378cd19a1bdb9f1095cbb..1c8b7e9f758fcaecf6cb60a2e6938fcfd07c34b8) (since I chose to fix the older commit that perpetuated it): https://github.com/grafana/k6/blob/1c8b7e9f758fcaecf6cb60a2e6938fcfd07c34b8/js/runner.go#L618-L621

metrics/tags.go
```diff
@@ -3,12 +3,162 @@ package metrics
...
+// With returns another TagSet object that contains the combination of the
+// current receiver tags and the name=value tag from its parameters.
+//
+// It doesn't modify the receiver, it will either return an already existing
+// TagSet with these tags, if it exists, or create a new TagSet with them and
+// return it.
+//
+// If a tag with the specified name already exists in the set, it will be
+// overwritten with the new value in the returned set.
+func (ts *TagSet) With(name, value string) *TagSet {
```
oleiade, line 38: nitpick, feel free to ignore 🐻 

When looking at it in a context where it's used, I found `With` and `Without` to be kind of abstract. Although I see where it comes from, as those are immutable methods, and we need to carry on the meaning that a new instance will be created out of it,  I wondered if `Including`, and `Excluding` could be more explicit and lead to "sentence-like" when taken in their calling context:
- `myTag := myRegistry.TagSet().Including("a", "1")` instead of `myRegistry.RootTagSet().With("a", "1")`
- `tags := currentTags.Excluding("key")`

metrics/tags.go
```diff
@@ -3,12 +3,162 @@ package metrics
...
+// Without returns another TagSet object that contains all of the tags from the
+// existing TagSet except the one with the given key.
+//
+// It doesn't modify the receiver, it will either return an already existing
+// TagSet with these tags, if it exists, or create a new TagSet with them and
+// return it.
+//
+// If a tag with the specified name doesn't exist in the set, it will return the
+// receiver.
+func (ts *TagSet) Without(name string) *TagSet {
```
oleiade, line 51: Same nitpick as for `With` 🐻

oleiade (dismissed):
TLDR LGTM 🎉 

Kudos to you @codebien and @na-- for the great work. It's all coming together pretty nicely, and I really appreciated that there was an obvious effort made on keeping the code small, simple and well documented 👏🏻 👍🏻 

I have a bunch of comments on naming, and documentation here and there, but nothing blocking.

I leave this review without feeling 100% sure I thoroughly understand how `Registry.rootTagSet` and samples `Timeseries` interact under the hood, but it's something that could and should be clarified later.

metrics/tags.go
```diff
@@ -3,12 +3,162 @@ package metrics
...
+}
+
+// Map returns a {key: value} string map with all of the tags in the set.
+func (ts *TagSet) Map() map[string]string {
+	return ((*atlas.Node)(ts)).Path()
+}
+
+// WithTagsFromMap sorts the given tags by their keys and adds them to the
+// current tag set one by one, without branching out. This is generally
+// discouraged and sequential usage of the AddTag() method should be preferred
```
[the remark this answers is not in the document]
na--: ah, I apparently forgot to update the comment when `TagSet.AddTag()` was renamed to `TagSet.With(), will fix it :+1:

metrics/tags.go
na--: Hmm, I think I personally slightly prefer `With()` and `Without()`, but I am fine with either, so I can rename if everyone else prefers `Including()` and `Excluding()` :thinking: 

I prefer `With()` and `Without()` because they are shorter and because I somewhat associate `Including` and `Excluding` with filtering more than I associate them with sets. But again, both will work, so @mstoykov @codebien @olegbespalov , please comment if you'd like me to rename them.

na--: And another reason, consider the following pieces of code:
```go
tags := tagSet().With("foo", "bar").With("foo", "baz")
```
and
```go
tags := tagSet().Including("foo", "bar").Including("foo", "baz")
``` 

In the first example, while it's not clear that the second `With()` statement will overwrite the first (which it will), it's not suggesting the opposite either. But the second example might mislead a reader to think that the final `tags` value will include both `foo` tags, which it won't.

metrics/tags.go
```diff
@@ -3,12 +3,162 @@ package metrics
...
+}
+
+// UnmarshalJSON WILL ALWAYS RETURN AN ERROR because a TagSet needs to be
+// started from a common atlas root. This function exists to prevent any
+// automatic reflection-based attempts at unmarshaling.
+func (ts *TagSet) UnmarshalJSON([]byte) error {
+	return errors.New("metrics.TagSet cannot be directly unmarshalled from JSON")
+}
+
+var _ interface {
```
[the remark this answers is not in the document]
na--: Resolved by the [latest force-pushes](https://github.com/grafana/k6/compare/28bfc28a17b4c9b7a6051201ffcda4ce3d3d8b16..d3db2988e0db43a96c4a78fb5d96018639c2ba02) to the last commit

metrics/tags.go
```diff
@@ -3,12 +3,162 @@ package metrics
...
+}
+
+// WithTagsFromMap sorts the given tags by their keys and adds them to the
+// current tag set one by one, without branching out. This is generally
+// discouraged and sequential usage of the AddTag() method should be preferred
+// and used whenever possible.
+//
+// The only place this method should be used is if we already have a
+// map[string]string of tags, e.g. with the test-wide --tags from the root, with
+// scenario.tags, with custom per-reqeust tags from a user, etc. Then it's more
```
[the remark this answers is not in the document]
na--: Resolved by the [latest force-pushes](https://github.com/grafana/k6/compare/28bfc28a17b4c9b7a6051201ffcda4ce3d3d8b16..d3db2988e0db43a96c4a78fb5d96018639c2ba02) to the last commit

metrics/tags.go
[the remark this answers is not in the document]
na--: Resolved by the [latest force-pushes](https://github.com/grafana/k6/compare/28bfc28a17b4c9b7a6051201ffcda4ce3d3d8b16..d3db2988e0db43a96c4a78fb5d96018639c2ba02) to the last commit

metrics/tags.go
mstoykov: Counter suggestion: `Plus`/`Minus` 

`tags := tags.Plus("foo" , "bar")` 

`tags := tagSet().Plus("foo", "bar").Plus("foo", "baz")`

same number of letters but I would say the same probem as `Including` - not really communicating that the same tags will be overwritten.

I am not certain that `With` does this any better though :shrug:. 

I did mention before that `With` and `Without` seem very similar while scanning code thouhg so :shrug: .

Also `And`/`Not` might be fine ... although we lean really into the `Set` part at this point:

`tags := tags.And("foo", "bar").And("foo", "baz").Not("bar")`

Not really certain it  communicates anything better ... but even less symbols :P

na--: hmm `With()` doesn't communicate straight addition / accumulation, which both `Including` and `Plus` do. In my mind, I read `tags := rootTagSet().With("scenario", "foo")` as "tags is the root tag set **with** a `scenario` tag equal to `foo`". Whether the root tag set had such a tag or not doesn't matter, "with" suggest a singular tag value. Whereas `Plus`, `Including` and `And` kind of suggest both will be kept. 

The only other word that I can think of, similar to `With` is `Containing`, but it's longer, it doesn't flow as well when I read it in my head and it doesn't have a nice opposite word :sweat_smile:

lib/netext/httpext/tracer_test.go
```diff
@@ -129,7 +130,7 @@ func TestTracer(t *testing.T) { //nolint:tparallel
 				time.Sleep(traceDelay)
 			}
 			trail := tracer.Done()
-			trail.SaveSamples(builtinMetrics, metrics.IntoSampleTags(&map[string]string{"tag": "value"}))
+			trail.SaveSamples(builtinMetrics, registry.RootTagSet().WithTagsFromMap(map[string]string{"tag": "value"}))
```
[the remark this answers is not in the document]
na--: ah, this was because of the mass search-and-replace :sweat_smile: will fix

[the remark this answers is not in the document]
na--: Resolved by the latest [force-push](https://github.com/grafana/k6/compare/d3db2988e0db43a96c4a78fb5d96018639c2ba02..6dc8bdc9b75362f1d15b97943a65406e8c5e88d2)

metrics/tags.go
oleiade: I kinda like the `New*` approach you propose @olegbespalov, it does make the copy and the immutability explicit. The only concern I would have is that the `New` pattern in Go is pretty much established as a way to instantiate an object from scratch, rather than as an already existing's instance method ❓

na--: Yeah, `NewWith()`, `NewWithout()`, etc. would be both more verbose and somewhat confusing with regular constructors :thinking: After all, the majority of time `With()` won't return a new object, but an already existing atlas `Node`... I am not completely against it, but I don't think it's an improvement.

@mstoykov, @oleiade, can you restate what specifically you find objectionable about 
```go
func (ts *TagSet) With(name, value string) *TagSet { /* ... */ }
func (ts *TagSet) Without(name string) *TagSet { /* ... */ }
```

Give some example where their usage is confusing or something like that? I am somewhat at a loss exactly what their problems are that we are trying to solve :sweat_smile: Having re-written the PR with them, the code seems perfectly readable and understandable to me, but I might be missing something. And in the interests of avoiding bikeshedding, I'd like to reach some consensus here soon :bow:

olegbespalov: For sure, from scratch (an `existing` as an argument) and `New*` is a standalone method, but I provided it just as my consideration (FYI); 

In any case for code below is more readable with the existing chain approach instead than with `New*`:

Good: 
```
x.With("tag", "value").With("tag2", "value").With("tag2", "value")
```

vs.

Bad:
```
New(New(New(x, "tag", "value"), "tag2", "value"), "tag3", "value")
```

mstoykov: :-1: for `New*` for the reasons @na-- gave more or less. 

@na-- My original issue was `With` and `Without` being way too similar when just scrolling through. Nobody else has mentioned this and I very much doubt that we will use this *everywhere* so it's unlikely to be a problem.

More or less my suggestions were about trying to figure out a slightly better pair of names, but I also think `With`/`Without` are currently the best so far. Especially as the others usually have way bigger downsides. 

I personally also prefer if we leave this discussion and if there is such a big problem later on we can rename (and leave `With`/`Without` for a version as deprecated). I would expect that if we would want to rename them we will have more concrete problems and arguments.

na--: Alright, then it seems like there is at least a tentative majority (if not an outright consensus) that `With()` and `Without()` are good enough, or at least the least bad alternative... :sweat_smile:  I'll resolve this conversation for now then. 

And yeah, we can always rename this in the future, if we find some big issues or figure out something substantially better.

metrics/tags.go
na--: Resolving for the same reason as https://github.com/grafana/k6/pull/2594#discussion_r946880613

metrics/tags.go
codebien: I like `With`, it is explicit enough for the purpose, IMHO. I would like to not have a similar word for the negative form so we could mix the proposals. 

So we could have `With` and `Drop` - or `Remove` or `Delete` ( eventually `Del` in short).

codebien (commented):
Thanks for the effort to push more steps :bow:  I have a few minor comments.

lib/vu_state.go
```diff
@@ -73,68 +73,39 @@ type State struct {
...
-}
-
-// TagMap is a safe-concurrent Tags lookup.
-type TagMap struct {
-	m     map[string]string
+// VUStateTags wraps the current VU's tags and ensures a thread-safe way to
+// access and modify them exists. This is necessary because the VU tags and
+// metadata can be modified from the JS scripts via the `vu.tags` API in the
+// `k6/execution` built-in module.
+type VUStateTags struct {
```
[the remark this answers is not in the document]
na--: Yeah, I kind of want to rename `lib.State` either to `lib.VUState`, or to get it out of the `lib` package altogether and make it clear in some way that it's the VU state. :+1: I'll add a TODO

metrics/sample.go
```diff
@@ -2,21 +2,26 @@ package metrics
...
-	Value  float64
+	Tags   *TagSet
+}
+
+// A Sample is a single metric measurement.
+type Sample struct {
+	TimeSeries
+	Time     time.Time
+	Value    float64
+	Metadata map[string]string // could be nil
```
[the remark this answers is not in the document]
na--: hmm I am fine with removing the map, but I'd prefer to leave the TODOs until the unindexable tags PR

lib/vu_state.go
```diff
@@ -73,68 +73,39 @@ type State struct {
...
-	tg.mutex.RLock()
-	defer tg.mutex.RUnlock()
-	v, ok := tg.m[k]
-	return v, ok
-}
-
-// Len returns the number of the set keys.
-func (tg *TagMap) Len() int {
+// GetCurrentValues returns the value of the VU tags in a thread-safe way.
+func (tg *VUStateTags) GetCurrentValues() *metrics.TagSet {
```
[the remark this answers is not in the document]
na--: Hmm, I am maybe fine with getting rid of `Get`, but I kind of like `Current`, since it implies that the value can change, and so implies that there is some kind of locking... :thinking: But even the `Get` part kind of makes sense in the context of this object that is basically there to ensure thread safety :thinking: It maybe makes a bit more sense in the context of the unindexable tags PR (which I really hope to publish the WIP version today :crossed_fingers:), where I have this `TagsAndMeta` struct and the methods look like this:
```go
// metrics.TagsAndMeta
type TagsAndMeta struct {
	Tags     *TagSet
	Metadata map[string]string // could be nil
}

// GetCurrentValues returns the current value of the VU tags and a copy of the
// metadata (if any) in a thread-safe way.
func (tg *VUStateTags) GetCurrentValues() metrics.TagsAndMeta {
	tg.mutex.RLock()
	defer tg.mutex.RUnlock()
	return tg.tagsAndMeta.Clone()
}

// Modify allows the thread-safe modification of the current VU tags and metadata.
func (tg *VUStateTags) Modify(callback func(tagsAndMeta *metrics.TagsAndMeta)) {
	tg.mutex.Lock()
	defer tg.mutex.Unlock()
	callback(tg.tagsAndMeta)
}
```

In this context, when one of the methods returns a pointer and the other does not, I think even `Get` has value.

output/cloud/output.go
```diff
@@ -452,82 +441,84 @@ func (out *Output) aggregateHTTPTrails(waitPeriod time.Duration) {
...
+			aggrData.CalcAverages()
+
+			if aggrData.Count > 0 {
+				/*
+					out.logger.WithFields(logrus.Fields{
+						"http_samples": aggrData.Count,
+						"ratio":        fmt.Sprintf("%.2f", float64(aggrData.Count)/float64(trailCount)),
+						"t":            time.Since(start),
+					}).Debug("Aggregated HTTP metrics")
+				//*/
```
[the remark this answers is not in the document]
na--: I am not sure where this code is buried, maybe in a single commit that got reverted? Can't find it in the final diff, sorry :confused:

metrics/sample.go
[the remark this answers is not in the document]
codebien: I'm fine with the TODOs :+1:

output/cloud/output.go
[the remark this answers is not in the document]
codebien: I see it here: https://github.com/grafana/k6/blob/atlas/output/cloud/output.go#L510-L516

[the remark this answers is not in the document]
mstoykov: This is a really old comment for debugging purposes. I think it has been around since the beginning of the cloud output aggregation.

codebien:
I added also my opinion regarding With/Without naming here https://github.com/grafana/k6/pull/2594#discussion_r947724265

output/cloud/output.go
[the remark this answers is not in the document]
na--: Yeah, it is in `master` too: https://github.com/grafana/k6/blob/master/output/cloud/output.go#L518-L524
I guess it showed on your diff because the indentation changed and not on mine because I've disabled that in github diffs

metrics/tags.go
mstoykov: I think I like `Drop` the most :thinking: and it seems I finally have someone who also things they are too similar :tada:. 

But again I do think we can change this a bit later again. We certainly do have time until this is released and we will likely add and change a bunch of stuff, so not certain if renaming now is that beneficial unless the majority of the time decides it's really better.

As at this point it still seems like different people have different ideas on how this should look and `With`/`Without` seems to still be the okay+ for most people. So ... uh please write your opinion if you disagree :P

metrics/sample.go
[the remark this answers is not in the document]
na--: Resolved by the [latest force-push](https://github.com/grafana/k6/compare/92dbf477b8e7a102f38da3aad849912899f5a9cc..727391c38dd2cbb2b571e024032910b14f3fd183)

metrics/tags.go
codebien: I disagree, we are using `Without` in only one place, so it seems quite quick to change. We have some time before we are going to merge this so I think it's better to change it now if we have an agreement on one of the previous.

Instead, if we will continue to have different opinions let's keep it as-is.

na--: I am for keeping them as they are. You can't mix them, because `With()` and `Without()` have a different number of arguments. And besides, after https://github.com/grafana/k6/pull/2654 is merged, almost nobody should directly use `Without()` outside of tests, it will probably only be used in these 2 places in the code:
https://github.com/grafana/k6/blob/6ed8668af8eba6669e79d141f8e3ec942fa8e367/metrics/tags.go#L185-L186
https://github.com/grafana/k6/blob/6ed8668af8eba6669e79d141f8e3ec942fa8e367/metrics/tags.go#L214-L216

output/cloud/output.go
```diff
@@ -363,24 +364,33 @@ func (out *Output) AddMetricSamples(sampleContainers []metrics.SampleContainer)
 				values[metrics.IterationsName] = 1
 			}
 
+			encodedTags, err := easyjson.Marshal(sc.GetTags())
+			if err != nil {
+				out.logger.Error("Encoding tags failed", err)
```
olegbespalov, line 369: A small one otherwise, I believe formatting will be not so nice

```suggestion
				out.logger.WithError(err).Error("Encoding tags failed")
```

output/cloud/output.go
```diff
@@ -363,24 +364,33 @@ func (out *Output) AddMetricSamples(sampleContainers []metrics.SampleContainer)
...
-					Tags:   sc.GetTags(),
+					Tags:   encodedTags,
 					Values: values,
 				},
 			})
 		default:
 			for _, sample := range sampleContainer.GetSamples() {
+				encodedTags, err := easyjson.Marshal(sample.Tags)
+				if err != nil {
+					out.logger.Error("Encoding tags failed", err)
```
olegbespalov, line 384: Same as above 

```suggestion
					out.logger.WithError(err).Error("Encoding tags failed")
```

output/cloud/output.go
olegbespalov, line 516: If we don't need it (since no TODO or NOTICE here), let's clean it!

```suggestion	
```

codebien:
@olegbespalov I added your suggestions

go.mod
```diff
@@ -14,14 +14,15 @@ require (
 	github.com/gorilla/websocket v1.5.0
 	github.com/grafana/xk6-redis v0.1.1
 	github.com/grafana/xk6-timers v0.1.2
-	github.com/grafana/xk6-websockets v0.1.3
+	github.com/grafana/xk6-websockets v0.1.4-0.20220909124839-39d27946bc0c
```
olegbespalov, line 17: Maybe create a tag in `github.com/grafana/xk6-websockets`? :thinking:

output/cloud/output.go
olegbespalov: I see no changes :cry:

go.mod
codebien: It doesn't sound optimal to me to tag an under-development branch. I expect you want it for documenting better, is it right?

Is a comment a good alternative?

```suggestion
	github.com/grafana/xk6-websockets v0.1.4-0.20220909124839-39d27946bc0c  // v0.1.3 plus time series migration
```

output/cloud/output.go
codebien: @olegbespalov I missed one, you're right, they are two lines in two different places. It should be fixed now.

go.mod
olegbespalov: Ah no, I think it's fine even without :)

olegbespalov (dismissed):

na-- (dismissed):

oleiade (approved):

olegbespalov (dismissed):

mstoykov:
@imiric there is a merge conflict

codebien:
At this point, I consider squashing to merge so I just merged the current master branch here to avoid any possibility to break in some way this branch rebasing.

olegbespalov (approved):

go.mod
mstoykov, line 17: Is this how you are going to leave it ? And then update the extenions and then k6 again?
Or shall we remove it(the extension) for this PR and then update it and readd.

Again in the first case as we will squash this PR the commit that the extension uses will be removed after merge.

js/modules/k6/ws/ws.go
mstoykov: I guess we should resolve this @na--?

mstoykov: And also remove the comment?

lib/test_state_test.go
```diff
@@ -0,0 +1 @@
+package lib
```
mstoykov, line 1: maybe don't add this file?

lib/vu_state.go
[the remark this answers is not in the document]
mstoykov: I dislike the current name because it is long and IMO adds very little info. 

I kind of feel like `Values` will be misinterpreted as well. But my suggestion of `Copy` also seems easy to misinterpret so :shrug: 

I am kind of for just leaving it as is - we are going to come back to this code in the future again either way so  we can discuss the better name then.

go.mod
mstoykov: If we *remove* it from the source as well it will build

mstoykov (approved):
LGTM! 

I have left comments - but none of them are blocking

codebien:
As a reminder, otherwise, the `xk6-websockets` extension will be broken:
- from now rebasing is denied for this PR
- for merging this PR use the classic Merge feature for keep the commit history

go.mod
```diff
@@ -43,6 +43,8 @@ require (
 	gopkg.in/yaml.v3 v3.0.1
 )
 
+require github.com/grafana/xk6-websockets v0.1.4-0.20221004122443-2565a84e0662
```
mstoykov, line 46: Can you tag v0.1.4 and redo *this* commit in particular?

mstoykov (approved):

na-- (approved):
:tada: :crossed_fingers: :pray: :sweat_smile:

Issue #903, opened 2019-01-21
Run windows tests with `-race`


MStoykov:
Because of a [recent issue](https://github.com/golang/go/issues/27089) `-race` needs CGO which needs gcc under windows :( . 
We should still be able to run our test with `-race` even if that mean that we need to install gcc.

PR #2595, opened 2022-07-11
#903: added vet flag to resolve race issue

vanshaj:
As per the last comment in issue : https://github.com/golang/go/issues/27089 
I have turned off the vet flag and "make tests" worked fine on my windows machine.

Kindly review


olegbespalov:
Hi @vanshaj !

Thanks for your PR.

Could you please provide more details about the race issue? Maybe an output? :thinking: 

Cheers!

vanshaj:
> Hi @vanshaj !
> 
> Thanks for your PR.
> 
> Could you please provide more details about the race issue? Maybe an output? 🤔
> 
> Cheers!

yess, Kindly have a look at below image

image

olegbespalov:
In #903, I believe @mstoykov meant, first of all, the CI, but we don't run `make test` there. And it seems like in [one of the last CI](https://github.com/grafana/k6/pull/2591) refactorings. We fixed it [with the workaroubnd](https://github.com/grafana/k6/pull/2591/files#diff-faff1af3d8ff408964a57b2e475f69a6b7c7b71c9978cccc8f471798caac2c88R33-R38).

However, I believe it's also essential to have a `make test` target fixed too (maybe at some point, both the local & CI will run the same command). So let's try to finalize this PR :+1:

Makefile
```diff
@@ -1,30 +1,30 @@
-GOLANGCI_LINT_VERSION = $(shell head -n 1 .golangci.yml | tr -d '\# ')
```
olegbespalov, line 1: I see that, for some reason, the change affected whole formatting. Is it possible to keep only the actual change(s)?

Makefile
```diff
@@ -1,30 +1,30 @@
...
+		-v $(TMPDIR)/golangci-cache-$(GOLANGCI_LINT_VERSION):/golangci-cache \
+		--env "GOLANGCI_LINT_CACHE=/golangci-cache" \
+		-w /app golangci/golangci-lint:$(GOLANGCI_LINT_VERSION) \
+		make lint
+
+lint :
+	golangci-lint run --out-format=tab --new-from-rev master ./...
+
+tests :
+	go test -race -timeout 210s ./... -vet=off
```
olegbespalov, line 23: Just a small note here that this affects runs for all operation systems which isn't good, but I guess in that case it's fine as long as we have the linter altogether (`make check`)

olegbespalov (changes requested):

Makefile
vanshaj: Sure I will make the formatting correct.

Makefile
vanshaj: @olegbespalov Do you want me to add some check related to underlying OS?

Makefile
olegbespalov: Hi @vanshaj 

It seems like the whole file formatting is still affected. It shows that 30 lines were removed and 30 added:
image

Makefile
olegbespalov: No, I guess in the current case, it will be over. Just share the opinion that in some cases we can do that.

olegbespalov:
Hey @oleiade. Could you please share your thoughts about the PR? Thanks.

oleiade:
Once again, thanks for your contribution @vanshaj, we really appreciate and value that you took the time and put in the effort to propose an improvement to k6.

Unfortunately, we discussed it internally, and experimented around this issue internally, and came to the conclusion we would not merge this PR in `master`. The main reason being, we cannot reproduce the issue you reported, and as a result, we can't risk introducing a change that might harm other users (including ourselves) in ways we cannot foresee for a potential benefit we cannot observe.

I hope this makes sense, and that you understand. Please don't hesitate to contribute again in the future 🙇🏻

PR #2596, opened 2022-07-12
output/json: Metric data race

codebien:
:warning: [**Breaking Change**] The `tainted` field has been removed from the flushed fields for a metric type.

The previous implementation [makes a copy](https://github.com/grafana/k6/blob/bfa415c5fdfa6084f49b02877540cf6c05e23605/output/json/json_easyjson.go#L226) of the Metric struct where it generates a data race on the Observed field that [is written](https://github.com/grafana/k6/blob/bfa415c5fdfa6084f49b02877540cf6c05e23605/metrics/engine/engine.go#L133)
from the metrics.Engine.

The current implementation reduces the fields copied resolving the simplest race on Observed. The Observed field is not useful for the JSON output.

`metrics.Thresholds` shouldn't be racy because the output [mostly read "immutable"](https://github.com/grafana/k6/blob/bfa415c5fdfa6084f49b02877540cf6c05e23605/metrics/thresholds.go#L126-L133) values.

The performance seems not impacted, it is aligned with the results from https://github.com/grafana/k6/commit/a5146cf7099cfc5b4ec8938c667c507ce373aff9.

```
BenchmarkFlushMetrics-8   	      61	  16767247 ns/op	 2526610 B/op	   20067 allocs/op
```

### DIFF
--- a/output/json/json.go
+++ b/output/json/json.go
@@ -183,13 +183,6 @@ func (o *Output) handleMetric(m *metrics.Metric, jw *jwriter.Writer) {
 		wrapped.Data.Thresholds = ts
 	}
 
-	// TODO: refactor after the other refactors
-	// in the metrics area will be completed.
-	//
-	// This parts can be racy because
-	// they are controlled and written from the metrics.Engine.
-	wrapped.Data.Tainted = m.Tainted
-
 	wrapped.MarshalEasyJSON(jw)
 	jw.RawByte('\n')
 }
--- a/output/json/json.go
+++ b/output/json/json.go
@@ -28,9 +28,7 @@ import (
 	"time"
 
 	"github.com/klauspost/compress/gzip"
-
 	"github.com/mailru/easyjson/jwriter"
-
 	"github.com/sirupsen/logrus"
 
 	"go.k6.io/k6/metrics"
@@ -52,7 +50,7 @@ type Output struct {
 	out         io.Writer
 	closeFn     func() error
 	seenMetrics map[string]struct{}
-	thresholds  map[string][]*metrics.Threshold
+	thresholds  map[string]metrics.Thresholds
 }
 
 // New returns a new JSON output.
@@ -132,11 +130,13 @@ func (o *Output) Stop() error {
 
 // SetThresholds receives the thresholds before the output is Start()-ed.
 func (o *Output) SetThresholds(thresholds map[string]metrics.Thresholds) {
-	ths := make(map[string][]*metrics.Threshold)
-	for name, t := range thresholds {
-		ths[name] = append(ths[name], t.Thresholds...)
+	if len(thresholds) == 0 {
+		return
 	}
-	o.thresholds = ths
+	o.thresholds = make(map[string]metrics.Thresholds, len(thresholds))
+	for name, t := range thresholds {
+		o.thresholds[name] = t
+	}
 }
 
 func (o *Output) flushMetrics() {
@@ -149,7 +149,6 @@ func (o *Output) flushMetrics() {
 		count += len(samples)
 		for _, sample := range samples {
 			sample := sample
-			sample.Metric.Thresholds.Thresholds = o.thresholds[sample.Metric.Name]
 			o.handleMetric(sample.Metric, jw)
 			wrapSample(sample).MarshalEasyJSON(jw)
 			jw.RawByte('\n')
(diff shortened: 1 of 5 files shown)


output/json/json.go
```diff
@@ -171,6 +170,29 @@ func (o *Output) handleMetric(m *metrics.Metric, jw *jwriter.Writer) {
...
+
+	if ts, ok := o.thresholds[m.Name]; ok {
+		wrapped.Data.Thresholds = ts
+	}
+
+	// TODO: refactor after the other refactors
+	// in the metrics area will be completed.
+	//
+	// This parts can be racy because
+	// they are controlled and written from the metrics.Engine.
```
mstoykov, line 190: Seems to me that Submetrics isn't really racy as they get initialized early (after some of the previous refactoring). 

And Tainted is in practice always going to be null at this time as thresholds get ran on a timer, and it's unlikely that timer has ran at the first time the JSON output will be outputting.  So maybe we should just drop `Tainted` as a slight breaking change. From what i can see again it would've always been null, so it seems not all that relevant :shrug:

output/json/json.go
```diff
@@ -171,6 +170,29 @@ func (o *Output) handleMetric(m *metrics.Metric, jw *jwriter.Writer) {
...
+	wrapped.Data.Tainted = m.Tainted
+	if len(m.Submetrics) > 0 {
+		wrapped.Data.Submetrics = m.Submetrics
+	}
```
mstoykov, line 194:
```suggestion
	wrapped.Data.Submetrics = m.Submetrics
```

I see no need for an extra `if`

output/json/json.go
```diff
@@ -171,6 +170,29 @@ func (o *Output) handleMetric(m *metrics.Metric, jw *jwriter.Writer) {
 	}
 	o.seenMetrics[m.Name] = struct{}{}
 
-	wrapMetric(m).MarshalEasyJSON(jw)
+	//nolint:exhaustruct
```
mstoykov, line 173: I think this is a linter added after the version we currently use so I would prefer if you don't add `nolint` for it but maybe downgrade your local version (I just did that yesterday). 

Preferably this should be fixed in the tool ... but I guess we will just update the version we use ... soon :tm:

output/json/json.go
```diff
@@ -132,11 +130,13 @@ func (o *Output) Stop() error {
 
 // SetThresholds receives the thresholds before the output is Start()-ed.
 func (o *Output) SetThresholds(thresholds map[string]metrics.Thresholds) {
-	ths := make(map[string][]*metrics.Threshold)
+	if thresholds == nil || len(thresholds) < 1 {
```
mstoykov, line 133:
```suggestion
	if len(thresholds) == 0 {
```

is enough, but I wonder if we even need it :thinking: does `make` panic on length being 0 ?

mstoykov (approved):
LGTM in general. 

I have some comments where we can remove some stuff, but other than that I am okay with this change.

mstoykov (approved):

output/json/json.go
codebien: How the output is implemented at the moment it doesn't make sense to be printed. We are just flushing the metrics the first time and we skip all the other times. So it seems reasonable to me to remove it because the value will be updated probably after the first time.

@na-- WDYT? Is it the right moment for this? Do we have future plans for the JSON output where `tainted` could be more valuable?

codebien: For the record, consensus to drop `tainted` achieved privately

mstoykov (approved):

output/json/json.go
```diff
@@ -171,6 +170,19 @@ func (o *Output) handleMetric(m *metrics.Metric, jw *jwriter.Writer) {
...
 
-	wrapMetric(m).MarshalEasyJSON(jw)
+	wrapped := metricEnvelope{
+		Type:   "Metric",
+		Metric: m.Name,
+	}
+	wrapped.Data.Name = m.Name
+	wrapped.Data.Type = m.Type
+	wrapped.Data.Contains = m.Contains
+	wrapped.Data.Submetrics = m.Submetrics
```
na--, line 180: hmm I think this may be race-y as well :thinking: can you add a sub-metric to the test?

mstoykov: It really shouldn't be as all the submetrics are *now* initialized early on in the starting of k6 as I mention in https://github.com/grafana/k6/pull/2596#discussion_r919808764 

https://github.com/grafana/k6/blob/7cdc0b18226a37f8cb90cacb57f81232d2298158/metrics/engine/engine.go#L73-L89

na--: fair enough :+1:

na--: was about to resolve this, but adding submetrics to the test still is a good idea though :sweat_smile: just to ensure we output whatever we expect (and we don't break that in a few versions)

codebien: Added

na-- (approved):

PR #2602, opened 2022-07-14
Modules test runtime

mstoykov:


mstoykov:
This is (likely) only the first iteration. 

I would prefer if more people try to use it in their extensions and see how it behaves for them.

js/modulestest/runtime.go
```diff
@@ -0,0 +1,51 @@
+// Package modulestest contains helpers to test js modules
+package modulestest
```
na--, line 2: hmm this seems like it should be in the `testutils` package (https://github.com/grafana/k6/tree/master/lib/testutils) :thinking: on the other hand, I wouldn't want for even a `lib` sub-package to depend on `js/`... :confused: so I wonder if we shouldn't move `testutils` to a top-level package and have this as a sub-package of it?

mstoykov: Maybe but given the breaking change that will be (for both packages) this seems like a thing that might be nice but will have fairly low impact.

I kind of like the fact that we have two diffrent test helper packages that are more under the place where they should be used/the package they should help with. Intead of just having a global "all test helpers go here" kind of variant.

oleiade (commented):
Testing this branch against my ongoing testing of [xk6-redis](https://github.com/grafana/xk6-redis) and getting back to you 🤝

oleiade (approved):
No red flags after integrating it with xk6-redis 👍🏻 

I stumbled into some issues with the way the context is initialized and canceled by the test runtime itself, and not being able to provide it from the outside if I recall correctly. But nothing blocking, and that wouldn't be changeable in the future 🚀

js/modulestest/modulestest.go
```diff
@@ -64,3 +66,9 @@ func (m *VU) Runtime() *goja.Runtime {
 func (m *VU) RegisterCallback() func(f func() error) {
 	return m.RegisterCallbackField()
 }
+
+func (m *VU) checkIntegrity() {
+	if m.InitEnvField != nil && m.StateField != nil {
+		panic("this is bug in your tests: you can't have both InitEnvField and StateField set at the same time")
```
na--, line 72:
```suggestion
		panic("there is a bug in your tests: you can't have both InitEnvField and StateField set at the same time")
```

js/modulestest/runtime.go
```diff
@@ -0,0 +1,51 @@
...
+
+// InitContext will set the InitContext of the Runtime
+func (r *Runtime) InitContext(i *common.InitEnvironment) {
+	r.VU.InitEnvField = i
+}
+
+// MoveToVUContext will set the state and nil the InitEnv just as a real VU
+func (r *Runtime) MoveToVUContext(state *lib.State) {
+	r.VU.InitEnvField = nil
+	r.VU.StateField = state
```
na--, line 50: hmm the `Field` suffix of all of these struct fields kind of annoys me now, but it's a minor nitpick :sweat_smile:

mstoykov:
@oleiade both of the fields(context and the cancel function) are [exposed](https://github.com/grafana/k6/pull/2602/files#diff-707687f99c6d6cc033b321838452724390a8d8849fe0da2cd0679845050cb117R26-R35) and I even specifically use the cancel function in a way to use it even if it is changed in some other place.

So you can definitely change it - I think I even use this in one of the tests in k6 core

js/modulestest/runtime.go
mstoykov: I have to give them name and I can't remove the `Field` as that is the name of the method defined on the interface ;0

mstoykov: Also, at least in theory this should only be used when defining the test, and then you should just use `vu.State()` to get back the `lib.State`. And in the majority of the test it seems like you barely need to touch those after figuring out some default test setup for the particular package tests.

na--: yeah, I know, but still unreasonably annoyed by them, and I don't have any better suggestions...  :sweat_smile:

js/modulestest/modulestest.go
```diff
@@ -64,3 +66,9 @@ func (m *VU) Runtime() *goja.Runtime {
 func (m *VU) RegisterCallback() func(f func() error) {
 	return m.RegisterCallbackField()
 }
+
+func (m *VU) checkIntegrity() {
+	if m.InitEnvField != nil && m.StateField != nil {
+		panic("there is a bug in your tests: you can't have both InitEnvField and StateField set at the same time")
```
codebien, line 72:
```suggestion
		panic("there is a bug in the test: InitEnvField and StateField are not allowed at the same time")
```

js/modulestest/runtime.go
```diff
@@ -0,0 +1,51 @@
...
+		EventLoop:     eventloop,
+		CancelContext: cancel,
+	}
+	// let's cancel again in case it has changed
+	t.Cleanup(func() { result.CancelContext() })
+	return result
+}
+
+// InitContext will set the InitContext of the Runtime
+func (r *Runtime) InitContext(i *common.InitEnvironment) {
```
codebien, line 43:
```suggestion
func (r *Runtime) SetInitContext(i *common.InitEnvironment) {
```

js/modulestest/runtime.go
```diff
@@ -0,0 +1,51 @@
...
+	r.VU.InitEnvField = i
+}
+
+// MoveToVUContext will set the state and nil the InitEnv just as a real VU
+func (r *Runtime) MoveToVUContext(state *lib.State) {
```
codebien, line 48: Is this a simplified version of the activate operation? If yes, then I would prefer some sort of parallelism with the real VU in terms of naming.

mstoykov: That doesn't sound like a bad idea ... but I am not certain that someone who isn't aware of the particulars of the k6 internals will be more okay with `Activate`. 

Arguably that name exists more because we are trying to hide the internals of the js.Runner outside of it. But here we are basically testing it/around it/in it. So it seems more useful IMO to name this something that the user (writing the test) will know they need. And most users are aware that they need the `*lib.State` in order to execute most of their code.

Maybe if decide to have overarching naming of all the phases we can use those - but that is way bigger task then this PR :) 


WDYT @na-- @oleiade

js/modulestest/runtime.go
mstoykov: Will wait on that to do it with https://github.com/grafana/k6/pull/2602#discussion_r928558087 if we decide on anything

js/modulestest/runtime.go
oleiade: I have no strong opinions on that matter. TL;DR: I'm rather for keeping it named `MoveToVUContext`. 

When trying out the integration in xk6-redis, I found `MoveToVUContext` quite intuitive, as making the runtime enter the VU context, or the init context, are basically the two only operations I was interested in.

I think `Activate` would have been confusing for me, but I haven't used much that part of the k6 internals either, so I might have a colder cache than you folks on that front. I would have judged `Activate` taken out of the k6 context as to broad, "activate, what? why?", whereas "move (what: the runtime) into a VU context" feels self-sufficient (although the word context is so polysemic in Go that it can lead to confusion.

oleiade:
For the reference, and in hope of feeding into the discussions, here's an example of how I used the `Runtime` type in my experimentation with xk6-redis. It assumes the `Runtime` and `VU` types are predefined.

```go
// Wrapping the Runtime in a dedicated type to avoid having to inject
// the redis module, and define the state for the VU context everytime
type fancyTestSetup struct {
	*Runtime
	tb testing.TB
}

func newFancyTestSetup(tb testing.TB) *fancyTestSetup {
	testRuntime := NewRuntime(tb)

	m, ok := New().NewModuleInstance(testRuntime.VU).(*ModuleInstance)
	require.True(tb, ok)

	err := testRuntime.VU.Runtime().Set("Client", m.Exports().Named["Client"])
	require.NoError(tb, err)

	return &fancyTestSetup{
		Runtime: testRuntime,
		tb:      tb,
	}
}

func (fts *fancyTestSetup) VUContextState() *lib.State {
	tb := httpmultibin.NewHTTPMultiBin(fts.tb)

	rt := goja.New()
	rt.SetFieldNameMapper(common.FieldNameMapper{})

	root, err := lib.NewGroup("", nil)
	if err != nil {
		fts.tb.Fatal(err)
	}

	samples := make(chan metrics.SampleContainer, 1000)

	return &lib.State{
		Group:  root,
		Dialer: tb.Dialer,
		Options: lib.Options{
			SystemTags: metrics.NewSystemTagSet(
				metrics.TagURL,
				metrics.TagProto,
				metrics.TagStatus,
				metrics.TagSubproto,
			),
			UserAgent: null.StringFrom("TestUserAgent"),
		},
		Samples:        samples,
		TLSConfig:      tb.TLSClientConfig,
		BuiltinMetrics: metrics.RegisterBuiltinMetrics(metrics.NewRegistry()),
		Tags:           lib.NewTagMap(nil),
	}
}

// Here we use the `MoveToVUContext`
func TestClientSet(t *testing.T) {
	t.Parallel()

	ts := newFancyTestSetup(t)
	ts.MoveToVUContext(ts.VUContextState())

	rs := RunT(t)
	rs.RegisterCommandHandler("SET", func(c *Connection, args []string) {
		if len(args) <= 2 && len(args) > 4 {
			c.WriteError(errors.New("ERR unexpected number of arguments for 'GET' command"))
			return
		}

		switch args[0] {
		case "existing_key", "non_existing_key":
			c.WriteOK()
		case "expires":
			if len(args) != 4 && args[2] != "EX" && args[3] != "0" {
				c.WriteError(errors.New("ERR unexpected number of arguments for 'SET' command"))
			}
			c.WriteOK()
		}
	})

	gotScriptErr := ts.EventLoop.Start(func() error {
		_, err := ts.VU.Runtime().RunString(fmt.Sprintf(`
			const redis = new Client({
				addrs: new Array("%s"),
			});

			redis.set("existing_key", "new_value")
				.then(res => { if (res !== "OK") { throw 'unexpected value for set result: ' + res } })
				.then(() => redis.set("non_existing_key", "some_value"))
				.then(res => { if (res !== "OK") { throw 'unexpected value for set result: ' + res } })
				.then(() => redis.set("expires", "expired", 10))
				.then(res => { if (res !== "OK") { throw 'unexpected value for set result: ' + res } })
				.then(() => redis.set("unsupported_type", new Array("unsupported")))
				.then(
					res => { throw 'expected to fail setting unsupported type' },
					err => {}
				)
		`, rs.Addr()))

		return err
	})

	assert.NoError(t, gotScriptErr)
	assert.Equal(t, 3, rs.HandledCommandsCount())
	assert.Equal(t, [][]string{
		{"SET", "existing_key", "new_value"},
		{"SET", "non_existing_key", "some_value"},
		{"SET", "expires", "expired", "ex", "10"},
	}, rs.GotCommands())
}

// Here we set the Runtime to run as in the init context
func TestClientCommandsInInitContext(t *testing.T) {
	t.Parallel()

	testCases := []struct {
		name      string
		statement string
	}{
		{
			name:      "spop should fail when used in the init context",
			statement: "redis.spop('shouldfail')",
		},
		{
			name:      "sendCommand should fail when used in the init context",
			statement: "redis.sendCommand('GET', 'shouldfail')",
		},
	}

	for _, tc := range testCases {
		tc := tc

		t.Run(tc.name, func(t *testing.T) {
			t.Parallel()

			fts := newFancyTestSetup(t)
			fts.InitContext(&common.InitEnvironment{})

			gotScriptErr := fts.EventLoop.Start(func() error {
				_, err := fts.VU.Runtime().RunString(fmt.Sprintf(`
				const redis = new Client({
					addrs: new Array("unreachable:42424"),
				});
	
				%s.then(res => { throw 'expected to fail when called in the init context' })
			`, tc.statement))

				return err
			})

			assert.Error(t, gotScriptErr)
		})
	}
}
```

mstoykov:
@oleiade  In `VUContextState` you should use the test runtime goja directly - it already has the FieldNameMapper set as well ;). From the looks of it you also do not use it at all apart from that ;).

oleiade:
Very good point indeed, it's relica, of my current way of doing it 🙇🏻

js/modulestest/runtime.go
codebien: >When trying out the integration in xk6-redis, I found MoveToVUContext quite intuitive

Out of curiosity, why VUContext is so intuitive for you? We don't mention in any doc the concept about `VUContext` and also the code doesn't export an API mentioning it. I guess is a concept that has [some historical reasons](https://github.com/grafana/k6/commit/a0b0eb12c994fec1fd83c565b2fcb443361166b1) because the unique place where we really mention is here: https://k6.io/docs/javascript-api/init-context

oleiade: I guess it's because from a purely programming perspective, it's named exactly after what I wanted to perform: "Prepare a runtime that's set up to reproduce the context of a VU executing a script". Probably auto-complete factors in: if I type `myRuntime.` and look at the auto-completion, `MoveToVUContext` lets me know somewhat clearly what's going to happen, whereas `Activate`, doesn't answer fundamental questions such as "what is implied by activate? What does it activate?", it's quite vague, and the expectation is unclear. But arguably, when programming (as in writing tests), I don't use the documentation, I only rely on the code's documentation, and the auto-completion mostly.

js/modulestest/runtime.go
codebien: I think the consensus there is to keep as-is so I think we can proceed to make a decision here.

mstoykov: I kind of wonder if I shouldn't just initialize the InitEnv with some very basic values and then let the users directly customize it before calling `MoveToVUContext` :thinking: That will also make some of the usage easier :thinking: 

WDYT @oleiade

oleiade: I find `SetInitContext` somewhat more intuitive, as without it my future self would have to remember that I need to set a `common.InitEnvironment` in the `r.VU.InitEnvField`. Consider it a helper/documentation method 👍🏻

mstoykov: What I meant @oleiade is that `r.VU.InitEnvField` *will* be set to some structure with some fields - most of them just initialized to not be nil (I can't really have a default `filesystems` for every test). And then you will only set the fields that you need (or the whole thing if you need to I guess)

mstoykov:
I did rebase this and squashed some of the fixes as well as rename InitContext to SetInitContext

codebien (approved):

js/modulestest/runtime.go
mstoykov: I made a PR with the removal of SetInitContext and having a default one that seems okay for ... all our current tests https://github.com/grafana/k6/pull/2621

js/modulestest/runtime.go
mstoykov: Resolving this as it seems as there is consensus

oleiade (approved):

codebien (approved):

na-- (approved):
:heart:

Issue #2587, opened 2022-07-05
Preventing storing environment variables' values in archive


olegbespalov:
# What?

Add a new flag to a `k6 archive` that will prevent storing environment variables (passed with the `--env`) in the archive's `metadata.json`.

# Why?

Sometimes scripts can use the environment variables in the init context and it's not possible to produce an archive without passing them, however storing the actual values of the environment variables can be harmful from the security perspective.

mstoykov:
To me it seems like dropping *all* env variables will be worse experience. A lot of people do use env variables to configure/scale their scripts apart from providing it with secrets/credentials.

IMO it will be easier if we provide a way to specify which env variables to be removed. Listing every one separately might be way too much work so maybe having a way to specify a `glob` pattern(s) that if matched will remove an env variable? 

Something like `k6 archive --remove-env SECRET* --remove-env ANOTHER_ONE script.js`?

 WDYT?

olegbespalov:
> To me it seems like dropping all env variables will be worse experience. A lot of people do use env variables to configure/scale their scripts apart from providing it with secrets/credentials.

@mstoykov I guess if the people use the variables, they can simply not use the new flag (and pass the secrets whenever they run the archive). 

On the other side, the case that covers this PR just excludes from storing everything that passed (only `--env` variables) to the `k6 archive`.

mstoykov:
> I guess if the people use the variables, they can simply not use the new flag (and pass the secrets whenever they run the archive). 

They could, but arguably the idea is to make it useful for more people. 

And I would argue being able to do `k6 cloud --remove-env SECRET* -r VU_COUNT=1000 script.js` is fairly important. And because in practice if you want to not save the credentials ... you don't have a choice really. You either will use this new flag and have it, but will have to set everything through the cloud UI . Which I would argue will not work for this case. Or you just put all the env variables in.

PR #2605, opened 2022-07-19
Introduce a new flag to the k6 archive sub-command

olegbespalov:
# What?

This PR introduces a new flag to the `k6 archive` command. If this flag is provided we exclude environment variables provided with the `--env` from the script's archive (`metadata.json`).

# Why?

There are some cases when environment variables are required for archiving, but it's safer to not store them.

Closes: #2587

### DIFF
--- a/cmd/archive.go
+++ b/cmd/archive.go
@@ -9,7 +9,8 @@ import (
 type cmdArchive struct {
 	gs *globalState
 
-	archiveOut string
+	archiveOut     string
+	excludeEnvVars bool
 }
 
 func (c *cmdArchive) run(cmd *cobra.Command, args []string) error {
@@ -35,6 +36,12 @@ func (c *cmdArchive) run(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	if c.excludeEnvVars {
+		c.gs.logger.Debug("environment variables will be excluded from the archive")
+
+		arc.Env = nil
+	}
+
 	err = arc.Write(f)
 	if cerr := f.Close(); err == nil && cerr != nil {
 		err = cerr
@@ -48,6 +55,14 @@ func (c *cmdArchive) flagSet() *pflag.FlagSet {
 	flags.AddFlagSet(optionFlagSet())
 	flags.AddFlagSet(runtimeOptionFlagSet(false))
 	flags.StringVarP(&c.archiveOut, "archive-out", "O", c.archiveOut, "archive output filename")
+	flags.BoolVarP(
+		&c.excludeEnvVars,
+		"exclude-env-vars",
+		"",
+		false,
+		"do not embed any environment variables (either from --env or the actual environment) in the archive metadata",
+	)
+
 	return flags
 }
 
--- a/cmd/archive_test.go
+++ b/cmd/archive_test.go
@@ -1,7 +1,13 @@
 package cmd
 
 import (
+	"archive/tar"
+	"bytes"
+	"encoding/json"
+	"errors"
+	"io"
 	"io/ioutil"
+	"os"
 	"path/filepath"
 	"testing"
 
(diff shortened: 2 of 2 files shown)


cmd/archive.go
```diff
@@ -55,6 +56,12 @@ func (c *cmdArchive) run(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	if c.skipEnv {
+		c.gs.logger.Info("environment variables will be excluded from the archive")
```
na--, line 60: Considering it's something that manually has to be triggered by a CLI flag, I am not sure we need to noisily log it

```suggestion
		c.gs.logger.Debug("environment variables will be excluded from the archive")
```

cmd/archive.go
```diff
@@ -55,6 +56,12 @@ func (c *cmdArchive) run(cmd *cobra.Command, args []string) error {
...
+
+		arc.WipeEnv()
```
na--, line 62: Considering that `arc.Env` is public, why do we need a new method? :confused:

cmd/archive.go
```diff
@@ -68,6 +75,8 @@ func (c *cmdArchive) flagSet() *pflag.FlagSet {
 	flags.AddFlagSet(optionFlagSet())
 	flags.AddFlagSet(runtimeOptionFlagSet(false))
 	flags.StringVarP(&c.archiveOut, "archive-out", "O", c.archiveOut, "archive output filename")
+	flags.BoolVarP(&c.skipEnv, "skip-env", "", false, "exclude from the archive environment variables passed with `--env`")
```
na--, line 78: hmm I am not sure this is the best name :thinking: though I am not sure the alternatives I can think of are any better :disappointed:  `--without-env` or maybe `--exclude-env`?

```suggestion
	flags.BoolVarP(&c.excludeEnv, "exclude-env", "", false, "do not embed any environment variables (either from --env or the actual environment) in the archive metadata")
```

since these env vars will be available in the script init phase, they just won't be embedded in the archive, it's best for the flag name to also indicate that, right?

cmd/archive.go
olegbespalov: Yes, `arc.Env` is public, and we can modify it without the method. Still, using the method, we also consolidate (and document) the modification's purpose, making it more maintainable for the future, WDYT?

cmd/archive.go
olegbespalov: >or the actual environment

But as far as I know, we never include actual environment variable values into the archives, right? :thinking:

cmd/archive.go
na--: I think it's fine to directly manipulate the public property. If we ever decide to make it private in the future, this code won't compile and we can make a method then. But as long as the property is private, and we only wipe it here, I don't see a need for a method.

cmd/archive.go
na--: we can, if the user executes `k6 archive --include-system-env-vars=true script.js`, see https://k6.io/docs/using-k6/k6-options/reference#include-system-env-vars

olegbespalov: True :+1:

cmd/archive.go
olegbespalov: Okay, I'll change it

cmd/archive.go
```diff
@@ -35,6 +36,12 @@ func (c *cmdArchive) run(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	if c.excludeEnv {
+		c.gs.logger.Debug("environment variables will be excluded from the archive")
+
+		arc.Env = map[string]string{}
```
na--, line 42: :thinking: I think this should be `nil`, not an empty map, so it is encoded as `null` in the `metadata.json`. That way we'll be able to differentiate (if we ever need to) whether an archive really had no environment variables when created (`"env": {}`) or if they were intentionally excluded (`"env": null`).

olegbespalov: See no issues with that, I can adjust it

na-- (approved):

mstoykov (approved):

codebien (dismissed):

codebien (changes requested):
Apparently you added an `archive.tar` file

codebien (approved):

mstoykov (approved):

codebien (approved):

na-- (approved):

PR #2610, opened 2022-07-21
Update goja

mstoykov:


mstoykov:
See https://github.com/dop251/goja/issues/410 for found issues.

IMO none of them are particularly problematic. The eval ones might be nice to be fixed - but likely not used by anyone either way or so I hope :crossed_fingers:

core/engine_test.go
```diff
@@ -967,7 +967,7 @@ func TestVuInitException(t *testing.T) {
 
 	var exception errext.Exception
 	require.ErrorAs(t, err, &exception)
-	assert.Equal(t, "Error: oops in 2\n\tat file:///script.js:10:9(30)\n\tat native\n", err.Error())
+	assert.Equal(t, "Error: oops in 2\n\tat file:///script.js:10:9(29)\n\tat native\n", err.Error())
```
olegbespalov, line 970: Could it be `assert.ErrorContains` or something similar? :thinking:

mstoykov: This is from 1.7.1 which I think we just update to. And I would prefer not to do more than the needed changes in update like that. 

But yeah - we can start replacing those calls. But it probably will be nicer if it's done through the whole codebase in one commit :shrug:

olegbespalov (approved):
:crossed_fingers:

core/engine_test.go
mstoykov: Actually is seems like it should be [`EqualError`](https://pkg.go.dev/github.com/stretchr/testify/require#EqualError) in this case

olegbespalov: Whatever works best, I mean something that we shouldn't adjust with the updates :shrug:

codebien: Maybe the others but this specific code is testing the exception so I think it's better to validate also the returned number.

na-- (dismissed):
:heart: :tada:

na-- (approved):

codebien (approved):

PR #2615, opened 2022-07-22
Improve the comments for `EventLoop.RegisterCallback()`

na--:
~I will rename all of the uses of `EventLoop.RegisterCallback()` in the code as well, once we agree on the name.~

I have scaled this back to just improving the comments and adding a `TODO` with a rename suggestion for the future (e.g. during https://github.com/grafana/k6/issues/2544)

### DIFF
--- a/js/eventloop/eventloop.go
+++ b/js/eventloop/eventloop.go
@@ -71,19 +71,19 @@ func (e *EventLoop) RegisterCallback() func(func() error) { // TODO: remove
 // script iteration.
 //
 // ReservePendingCallback() *must* be called from the main runtime thread, but
-// its result scheduleOnMainThread() is thread-safe and can be called from any
-// goroutine. scheduleOnMainThread() ensures that its callback parameter is
-// added to the VU runtime's tasks queue, to be executed on the main runtime
-// thread eventually, when the VU is done with the other tasks before it. Unless
-// the whole event loop has been stopped, invoking scheduleOnMainThread() will
-// queue its argument and "wake up" the loop (if it was idle, but not stopped).
+// its result queueCallback() is thread-safe and can be called from any
+// goroutine. queueCallback() ensures that its callback parameter is added to
+// the VU runtime's tasks queue, to be executed on the main runtime thread
+// eventually, when the VU is done with the other tasks before it. Unless the
+// whole event loop has been stopped, invoking queueCallback() will queue its
+// argument and "wake up" the loop (if it was idle, but not stopped).
 //
 // Keep in mind that once you call ReservePendingCallback(), you *must* also
-// call scheduleOnMainThread() exactly once, even if don't actually need to run
-// any code on the main thread. If that's the case, you can pass an empty no-op
+// call queueCallback() exactly once, even if don't actually need to run any
+// code on the main thread. If that's the case, you can pass an empty no-op
 // callback to it, but you must call it! The event loop will wait for the
-// scheduleOnMainThread() invocation and the k6 iteration won't finish and will
-// be stuck until the VU itself has been stopped (e.g. because the whole test or
+// queueCallback() invocation and the k6 iteration won't finish and will be
+// stuck until the VU itself has been stopped (e.g. because the whole test or
 // scenario has ended). Any error returned by any callback on the main thread
 // will abort the current iteration and no further event loop callbacks will be
 // executed this iteration.
@@ -91,7 +91,7 @@ func (e *EventLoop) RegisterCallback() func(func() error) { // TODO: remove
 // A common pattern for async work is something like this:
 //
 //    func doAsyncWork(vu modules.VU) *goja.Promise {
-//        scheduleOnMainThread := vu.ReservePendingCallback()
+//        queueCallback := vu.ReservePendingCallback()
 //        p, resolve, reject := vu.Runtime().NewPromise()
 //
 //        // Do the actual async work in a new independent goroutine, but make
@@ -100,7 +100,7 @@ func (e *EventLoop) RegisterCallback() func(func() error) { // TODO: remove
 //            // Also make sure to abort early if the context is cancelled, so
 //            // the VU is not stuck when the scenario ends or Ctrl+C is used:
 //            result, err := doTheActualAsyncWork(vu.Context())
-//            scheduleOnMainThread(func() error {
+//            queueCallback(func() error {
 //                if err != nil {
 //                    reject(err)
 //                } else {
@@ -117,7 +117,7 @@ func (e *EventLoop) RegisterCallback() func(func() error) { // TODO: remove
 // is immediately returned and the main thread resumes execution. It also
 // ensures that the Promise resolution happens safely back on the main thread
 // once the async work is done, as required by goja and all other JS runtimes.
-func (e *EventLoop) ReservePendingCallback() (scheduleOnMainThread func(callback func() error)) {
+func (e *EventLoop) ReservePendingCallback() (queueCallback func(func() error)) {
 	e.lock.Lock()
 	var callbackCalled bool
 	e.registeredCallbacks++
--- a/js/eventloop/eventloop.go
+++ b/js/eventloop/eventloop.go
@@ -89,9 +89,9 @@ func (e *EventLoop) RegisterCallback() func(func() error) { // TODO: remove
 //
 // A common pattern for async work is something like this:
 //
-//    func doAsyncWork() *goja.Promise {
+//    func doAsyncWork(vu modules.VU) *goja.Promise {
 //        addToMainThreadQueue := vu.RegisterPendingCallback()
-//        p, resolve, reject := runtime.NewPromise()
+//        p, resolve, reject := vu.Runtime().NewPromise()
 //
 //        // Do the actual async work in a new independent goroutine, but make
 //        // sure that the Promise resolution is done on the main thread:
--- a/js/eventloop/eventloop.go
+++ b/js/eventloop/eventloop.go
@@ -66,7 +66,7 @@ func (e *EventLoop) RegisterCallback() func(func() error) { // TODO: remove
 // RegisterPendingCallback signals to the event loop that you are going to do
 // some asynchronous work off the main thread and that you may need to execute
 // some code back on the main thread when you are done. So, once you call this
-// method, the even loop will wait for you to finish and give it the callback it
+// method, the event loop will wait for you to finish and give it the callback it
 // needs to run back on the main thread before it can end the whole iteration.
 //
 // RegisterPendingCallback() *must* be called from the main runtime thread, but
@@ -85,7 +85,7 @@ func (e *EventLoop) RegisterCallback() func(func() error) { // TODO: remove
 // be stuck until the VU itself has been stopped (e.g. because the whole test or
 // scenario has ended).
 //
-// A common patten for async work is something like this:
+// A common pattern for async work is something like this:
 //
 //    func doAsyncWork() *goja.Promise {
 //        addToMainThreadQueue := vu.RegisterPendingCallback()
--- a/js/eventloop/eventloop.go
+++ b/js/eventloop/eventloop.go
@@ -55,9 +55,11 @@ func (e *EventLoop) wakeup() {
 	}
 }
 
+// RegisterCallback is deprecated, see RegisterPendingCallback.
+//
 // Deprecated: due to the confusing name, we renamed this method to
-// RegisterPendingCallback
-func (e *EventLoop) RegisterCallback() func(func() error) {
+// RegisterPendingCallback.
+func (e *EventLoop) RegisterCallback() func(func() error) { // TODO: remove
 	return e.RegisterPendingCallback()
 }
 
--- a/js/eventloop/eventloop.go
+++ b/js/eventloop/eventloop.go
@@ -71,18 +71,18 @@ func (e *EventLoop) RegisterCallback() func(func() error) { // TODO: remove
 // script iteration.
 //
 // ReservePendingCallback() *must* be called from the main runtime thread, but
-// its result queueCallback() is thread-safe and can be called from any
-// goroutine. queueCallback() ensures that its callback parameter is added to
+// its result enqueueCallback() is thread-safe and can be called from any
+// goroutine. enqueueCallback() ensures that its callback parameter is added to
 // the VU runtime's tasks queue, to be executed on the main runtime thread
 // eventually, when the VU is done with the other tasks before it. Unless the
-// whole event loop has been stopped, invoking queueCallback() will queue its
+// whole event loop has been stopped, invoking enqueueCallback() will queue its
 // argument and "wake up" the loop (if it was idle, but not stopped).
 //
 // Keep in mind that once you call ReservePendingCallback(), you *must* also
-// call queueCallback() exactly once, even if don't actually need to run any
+// call enqueueCallback() exactly once, even if don't actually need to run any
 // code on the main thread. If that's the case, you can pass an empty no-op
 // callback to it, but you must call it! The event loop will wait for the
-// queueCallback() invocation and the k6 iteration won't finish and will be
+// enqueueCallback() invocation and the k6 iteration won't finish and will be
 // stuck until the VU itself has been stopped (e.g. because the whole test or
 // scenario has ended). Any error returned by any callback on the main thread
 // will abort the current iteration and no further event loop callbacks will be
@@ -91,7 +91,7 @@ func (e *EventLoop) RegisterCallback() func(func() error) { // TODO: remove
 // A common pattern for async work is something like this:
 //
 //    func doAsyncWork(vu modules.VU) *goja.Promise {
-//        queueCallback := vu.ReservePendingCallback()
+//        enqueueCallback := vu.ReservePendingCallback()
 //        p, resolve, reject := vu.Runtime().NewPromise()
 //
 //        // Do the actual async work in a new independent goroutine, but make
@@ -100,7 +100,7 @@ func (e *EventLoop) RegisterCallback() func(func() error) { // TODO: remove
 //            // Also make sure to abort early if the context is cancelled, so
 //            // the VU is not stuck when the scenario ends or Ctrl+C is used:
 //            result, err := doTheActualAsyncWork(vu.Context())
-//            queueCallback(func() error {
+//            enqueueCallback(func() error {
 //                if err != nil {
 //                    reject(err)
 //                } else {
@@ -117,7 +117,7 @@ func (e *EventLoop) RegisterCallback() func(func() error) { // TODO: remove
 // is immediately returned and the main thread resumes execution. It also
 // ensures that the Promise resolution happens safely back on the main thread
 // once the async work is done, as required by goja and all other JS runtimes.
-func (e *EventLoop) ReservePendingCallback() (queueCallback func(func() error)) {
+func (e *EventLoop) ReservePendingCallback() (enqueueCallback func(func() error)) {
 	e.lock.Lock()
 	var callbackCalled bool
 	e.registeredCallbacks++
--- a/js/eventloop/eventloop.go
+++ b/js/eventloop/eventloop.go
@@ -55,33 +55,34 @@ func (e *EventLoop) wakeup() {
 	}
 }
 
-// RegisterCallback is deprecated, see RegisterPendingCallback.
+// RegisterCallback is deprecated, see ReservePendingCallback.
 //
 // Deprecated: due to the confusing name, we renamed this method to
-// RegisterPendingCallback.
+// ReservePendingCallback.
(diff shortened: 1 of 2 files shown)


js/eventloop/eventloop.go
```diff
@@ -55,11 +55,60 @@ func (e *EventLoop) wakeup() {
...
+//
+// Keep in mind that once you call RegisterPendingCallback(), you *must* also
+// call addToMainThreadQueue() exactly once, even if don't actually need to run
+// any code on the main thread. If that's the case, you can pass an empty no-op
+// callback to it, but you must call it! The event loop will wait for the
+// addToMainThreadQueue() invocation and the k6 iteration won't finish and will
+// be stuck until the VU itself has been stopped (e.g. because the whole test or
+// scenario has ended).
+//
+// A common patten for async work is something like this:
```
inancgumus, line 88:
```suggestion
// A common pattern for async work is something like this:
```

js/eventloop/eventloop.go
```diff
@@ -55,11 +55,60 @@ func (e *EventLoop) wakeup() {
...
+// Deprecated: due to the confusing name, we renamed this method to
+// RegisterPendingCallback
 func (e *EventLoop) RegisterCallback() func(func() error) {
+	return e.RegisterPendingCallback()
+}
+
+// RegisterPendingCallback signals to the event loop that you are going to do
+// some asynchronous work off the main thread and that you may need to execute
+// some code back on the main thread when you are done. So, once you call this
+// method, the even loop will wait for you to finish and give it the callback it
```
inancgumus, line 69:
```suggestion
// method, the event loop will wait for you to finish and give it the callback it
```

inancgumus (approved):
The explanation in the comment is clearer and easy to understand now. Nice 👏 

I should add that I'm on the side of choosing simple names. Your mileage may vary :) I find the naming verbose (I know it's on purpose) and slightly less understandable (that's just me, don't look at me!). However, after reading the revised comment, I would probably be less confused and would have gotten the job done!

Here's my suggestion:

```go
func (e *EventLoop) PendingCallback() (once func(run func() error))
// or, just:
func (e *EventLoop) Callback() (once func(run func() error))
```

na--:
Definitely :-1: for just calling it `Callback`, that is definitely on the more confusing side :sweat_smile: `RegisterPendingCallback()` doesn't directly register the callback, that's true, but it actually alters the behavior of the event loop by registering that there is a callback actually pending :thinking: Just using `PendingCallback` is probably OK to me - it's clear enough, especially with the expanded comment, but I think `RegisterPendingCallback()` is just a tad better. 

Simple names are great when they are not misleading, and in this case there is a lot of subtle but crucial details that users of the method need to pay attention to, that is why we also have https://github.com/grafana/k6/issues/2544 :disappointed:

mstoykov:
`ReserveCallback` ?

na--:
Hmm :thinking: I think `Reserve` is better than `Register`, but `Pending` still caries useful information, so how about `ReservePendingCallback()`?

na--:
And maybe rename `addToMainThreadQueue` to `scheduleOnMainThread`, for this final signature:
```go
 func (e *EventLoop) ReservePendingCallback() (scheduleOnMainThread func(callback func() error)) {
```

:+1: or :-1:

js/eventloop/eventloop.go
```diff
@@ -55,11 +55,62 @@ func (e *EventLoop) wakeup() {
...
+// callback to it, but you must call it! The event loop will wait for the
+// addToMainThreadQueue() invocation and the k6 iteration won't finish and will
+// be stuck until the VU itself has been stopped (e.g. because the whole test or
+// scenario has ended).
+//
+// A common pattern for async work is something like this:
+//
+//    func doAsyncWork() *goja.Promise {
+//        addToMainThreadQueue := vu.RegisterPendingCallback()
+//        p, resolve, reject := runtime.NewPromise()
```
mstoykov, line 92: Maybe mention here that the goja documentation specifically says `resolve`/`reject` needs to *not* be run in parallel with the goja VM running.

js/eventloop/eventloop.go
```diff
@@ -55,11 +55,62 @@ func (e *EventLoop) wakeup() {
...
+//
+//        go func() {
+//            result, err := doTheActualAsyncWork()
```
mstoykov, line 95:
```suggestion
//            result, err := doTheActualAsyncWork() // the actual work here
```

js/eventloop/eventloop.go
```diff
@@ -55,11 +55,62 @@ func (e *EventLoop) wakeup() {
...
+//            addToMainThreadQueue(func() error {
+//                if err != nil {
+//                    reject(err)
+//                } else {
+//                    resolve(result)
+//                }
+//                return nil
```
mstoykov, line 102: Just now noticed that it isn't documented what happens when an error is returned :facepalm:

js/eventloop/eventloop.go
na--: isn't the function name enough? why repeat it as a comment?

js/eventloop/eventloop.go
na--: I can add that to the comment, what do you think I should say?

js/eventloop/eventloop.go
na--: I already do, see the comment below
```
// ... It also
// ensures that the Promise resolution happens safely back on the main thread
// once the async work is done, as required by goja and all other JS runtimes.
```

js/eventloop/eventloop.go
mstoykov: I would argue it will help with the explanation and drawing the attention to that specific line.

js/eventloop/eventloop.go
mstoykov: I guess just "Any error returned will abort the current iteration. And no further event loop callback will be executed."

mstoykov (commented):
I guess if we are going to be expanding the documentation further:
It's worth also nothing that in the case the event loop is aborted (because of an error being returned or the time running out or the user hitting ^C) the context will be Done. And it's kind of a good idea for the async code to take this into a consideration as well. 

So in this example if it was possible to abort that async work with providing it the context - that should'vbe been done. Maybe we are waiting on channel in that work - add a case with `vu.Context().Done()` and so on. 

Otherwise, as mentioned above that will block the event loop actually exiting until that happens. 

This likely will have *some* changes to it if/when we go with some kind of [event system](https://github.com/grafana/k6/issues/2432#issuecomment-1065218518) but likely most of this will still be true.

js/eventloop/eventloop.go
mstoykov: Hm ... I kind of prefer in example comments more or at least some. 

Also, you might want to run this through go1.19's `go fmt` as there are [some changes in that around comments](https://tip.golang.org/doc/go1.19) (scroll to "Doc comments"). It will be nice if at least for some of the comments we don't need to fix them in a few months  ;) .

js/eventloop/eventloop.go
na--: Resolved by https://github.com/grafana/k6/pull/2615/commits/23efb9583bfd5280ecac3fcbd333e7a713d752ed

js/eventloop/eventloop.go
na--: The first part is resolved by Resolved by https://github.com/grafana/k6/pull/2615/commits/23efb9583bfd5280ecac3fcbd333e7a713d752ed, but I don't want to mess with 1.19 on my machine right now. When it comes out, I imagine we'll have other comments that are problematic, we can fix all of them at once.

js/eventloop/eventloop.go
na--: Resolved by https://github.com/grafana/k6/pull/2615/commits/23efb9583bfd5280ecac3fcbd333e7a713d752ed

na--:
Since context usage can differ a lot, I added it to the example like this: https://github.com/grafana/k6/pull/2615/commits/b1973dadcab3c51b4158ea76067dac7513229a44

mstoykov (commented):
LGTM, but with two caveats
1. I would prefer to just update the documentation unless we agree that this name will stick. Otherwise we will just continue to add new names. I am definitely the worst person to say if this name is better or good enough so I will defer to others.
2. Arguably the more user facing documentation is on the [modules.VU method](https://github.com/grafana/k6/blob/521a2823afd0018cb8d9f8a2931c18dd63ad0010/js/modules/modules.go#L98-L111). It might also be good to link from the one to the others' documentation instead of copying and pasting it in two places :thinking:

na--:
I will adjust the comment and the name of the VU method (and where its used in the code) once everyone gives me a :+1: for the current signature, so I don't have to do it more than once

js/eventloop/eventloop.go
```diff
@@ -55,11 +55,69 @@ func (e *EventLoop) wakeup() {
...
+//        }()
+//
+//        return p
+//    }
+//
+// This ensures that the actual work happens asynchronously, while the Promise
+// is immediately returned and the main thread resumes execution. It also
+// ensures that the Promise resolution happens safely back on the main thread
+// once the async work is done, as required by goja and all other JS runtimes.
+func (e *EventLoop) ReservePendingCallback() (scheduleOnMainThread func(callback func() error)) {
```
olegbespalov, line 120: To be honest, the new name doesn't look better :man_shrugging: 

Maybe, or something similar.
```suggestion
func (e *EventLoop) RegisterCallbackQuota() (releaseQuota func(callback func() error)) {
```

Anyway, whatever the name will be at the end, please don't forget to fix the text in a panic below where now it stays `RegisterCallback called twice`.

na--: I'll fix all `RegisterCallback` mentions except the deprecated method when we reach consensus about the new name. I have a question about your name suggestion though - why ["quota"](https://www.merriam-webster.com/dictionary/quota)? :confused:

olegbespalov: @na-- personally, for me ["quota"](https://dictionary.cambridge.org/dictionary/english/quota) it represents what exactly this method is about:

> a fixed limit on the amount of something that someone is allowed to have or is expected to do

But as I said, I'm not considering myself the best person to define the new name. The only thing is that the current suggestion is still not ideal since, from the name of the method, it's hard to understand what the method is about.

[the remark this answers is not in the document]
codebien: And the alternative, `asyncslot.RunOnMain(func() error {` or `asyncslot.RunTask(func() error{`

na--: even by the "a fixed, limited amount or number that is officially allowed" definition, quota definitely doesn't represent what is going on here - we're dealing with neither a "fixed" nor a "limited" number of slots, nor is there any restriction to officially "allow" anything :confused: calling this function reserves you a single execution slot, but you can call it an unlimited number of times, there is no upper limit set by the VU, only the scenario/test duration limits it

so hard :-1: for "quota" from me, it adds more confusion than clarity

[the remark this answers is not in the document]
na--: Hmm, separate types are probably going to help with clarity a bit, but I don't want to grow the scope of this beyond a rename right now... :thinking: maybe something to consider in https://github.com/grafana/k6/issues/2544?

> And the relative alternatives, `asyncslot.RunOnMain(func() error {` or `asyncslot.RunTask(func() error{`

I see a couple of problems in this suggestion:
- `asyncslot` is wrong because the slot we are reserving is actually for synchronous work :sweat_smile: we are doing the asynchronous work in a separate thread (or 10, or 100) ourselves, and then we are reserving a slot for synchronous work on the main thread so we can call some callbacks or resolve promises or whatever
- `RunOnMain` and `RunTask` are misleading because they won't actually run whatever callback you give them immediately, they will schedule it to run in the future by adding it to the queue of pending task to be executed on the main thread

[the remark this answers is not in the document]
codebien: >Hmm, separate types are probably going to help with clarity a bit, but I don't want to grow the scope of this beyond a rename right now... thinking maybe something to consider in https://github.com/grafana/k6/issues/2544?

IMHO, that would be the way how we should proceed, but we are deprecating+adding methods here, so adding now a method for eventually breaking it again in the future doesn't sound good to me. So, should we just document here and discuss all the rest in https://github.com/grafana/k6/issues/2544 or in the dedicated PR?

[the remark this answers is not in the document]
na--: I doubt we'll ever remove `RegisterCallback` completely, `TaskQueue` depends on it after all :thinking: Right, @mstoykov? So we maybe can un-export it in the future, but it's still worth figuring out a non-confusing name for it.

olegbespalov: As I said, I'm fine with that. But on the other note please count this comment as :-1: for the suggested `ReservePendingCallback`.

olegbespalov: And to be more constructive, I'm joining this suggestion https://github.com/grafana/k6/pull/2615#issuecomment-1192798652

na--: Can you explain why? 

I already replaced `Register` with `Reserve`, since I also thought that was slightly betterm but as I said in https://github.com/grafana/k6/pull/2615#issuecomment-1193653699, `Pending` also caries useful information, even more so than `Reserve` or `Register`... It's probably the most essential part of the whole name, if we go by its [definitions](https://dictionary.cambridge.org/dictionary/english/pending): "about to happen or waiting to happen", "used to say that one thing must wait until another thing happens", "waiting to happen", "while waiting for something to happen or be decided", i.e. exactly what we want to communicate here...

olegbespalov: Sure. Simply because I find the "pending" part is overthinking. From `Reserve` it's already clear that callback won't be executed immediately, so it already carries the information.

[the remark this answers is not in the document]
mstoykov: I guess we can remove it entirely, but I doubt we will ever do it. So I am also for at least documenting it better. Renaming it again is dependent on people agreeing it's a big enough of an improvement to warrant us going through the trouble of deprecating and rewriting everything that uses it :( 

I very much doubt we will remove it though as it's a pretty good building block which is probably how any of the future "porcelain" on top will need to be implemented either way. 

Having the callbacks typed requires that we have a type somewhere and as this is uses primarily from the `js/modules` sub-package - it makes sense to be there ... but on the other hand it makes sense to be in the `js/eventloop`. I did not go this way as I didn't want to import two packages everywhere to use the API and at least at the time to me it didn't really seem to improve the situation all that much :shrug: . Again this is probably better decided by not the person who has designed this and knows how it works ;).

mstoykov, line 120:
```suggestion
func (e *EventLoop) ReservePendingCallback() (queueCallback func(callback func() error)) {
```

`MainThread` seems:
1. strange   as it's a goroutine and we have multiple ones per each VU. Now arguably `schelduleOnMainThread := vu.ReservePendingCallback()` looks somewhat okay ... but:
2. arguably is overly long without too much additional info ... IMO - which again probably isn't the most important one in this particular topic :rofl: 

`schedule` IMO have more of a connotation we will "schedule it at some particular time". Also we do queue it in a queue ;)

na--: :+1: :+1: for `queueCallback`, will change it

na--: Resolved by https://github.com/grafana/k6/pull/2615/commits/8da5202aee2e0d0afd4a8f82fc5c0f01d2b8a339

javaducky (commented):
`ReservePendingCallback` makes sense, given the documentation in the comment. 👍 

My only comment with the remainder of the function signature is using `queueCallback` as the name. I feel like _that_ could be `callback` or `pendedCallback`...as a user of the API, I'm not concerned with how these pended callbacks are stored, i.e. in a queue, stack, or anything else.

na--:
hmm @javaducky, "queue" is used as a verb in "queueCallback()", i.e. as a shorthand for "addThisCallbackToTheMainThreadTaskQueue()" :sweat_smile: so maybe "enqueueCallback()" will be clearer? :thinking:

javaducky:
Ah. I guess it threw me off because `ReservePendingCallback` returns `queueCallback` function which wraps the callback to be queued 😅.

mstoykov:
We can also name it `addCallback` :shrug:

na--:
Went with`enqueueCallback`

js/eventloop/eventloop.go
```diff
@@ -55,11 +55,69 @@ func (e *EventLoop) wakeup() {
...
+//
+// Keep in mind that once you call ReservePendingCallback(), you *must* also
+// call enqueueCallback() exactly once, even if don't actually need to run any
+// code on the main thread. If that's the case, you can pass an empty no-op
+// callback to it, but you must call it! The event loop will wait for the
+// enqueueCallback() invocation and the k6 iteration won't finish and will be
+// stuck until the VU itself has been stopped (e.g. because the whole test or
+// scenario has ended). Any error returned by any callback on the main thread
+// will abort the current iteration and no further event loop callbacks will be
+// executed this iteration.
```
codebien, line 89:
```suggestion
// executed in the same iteration.
```

codebien (approved):

codebien (approved):

js/eventloop/eventloop.go
```diff
@@ -55,11 +55,69 @@ func (e *EventLoop) wakeup() {
...
-func (e *EventLoop) RegisterCallback() func(func() error) {
+// RegisterCallback is deprecated, see ReservePendingCallback.
+//
+// Deprecated: due to the confusing name, we renamed this method to
+// ReservePendingCallback.
+func (e *EventLoop) RegisterCallback() func(func() error) { // TODO: remove
+	return e.ReservePendingCallback()
+}
+
+// ReservePendingCallback signals to the event loop that you are going to do
```
mstoykov, line 66: I have some problems with the word `pending` as to me it means that there already is a callback but it's waiting on something. 

I kind of prefer `future` although [future](https://en.wikipedia.org/wiki/Futures_and_promises) also means a "reverse" promise so ... uh not certain it's all that better.

From googling on synonyms of future i get:
- later - ReserveLaterCallback :grimacing: 
- skipping a bunch of terrible ones
- upcoming - ReserveUpcomingCallback seems a bit better
- planned - ReservePlannedCallback - seems even better to me ... I even can see how `PlanACallback` can be a good name
- prospective - not really 
- expected - ReserveExpectedCallback is not the worse, but definitely not my favorite.

na--: For me, pending is the word that describes exactly what is going on here, as I said in [this comment](https://github.com/grafana/k6/pull/2615#discussion_r928728759). Its definitions are "about to happen or waiting to happen", "used to say that one thing must wait until another thing happens", "waiting to happen", "while waiting for something to happen or be decided", exactly what we're trying to communicate here, right?

oleiade: Adding a bit of entropy: `ReserveScheduledCallback`?

oleiade (approved):
I find this change satisfactory as is. Feel free to re-request my review if this unfolds further 👍🏻

javaducky (approved):

na--:
We didn't have as broad of a consensus as I'd have liked for the new name, so I decided to revert it (for now) and merge only the improved comments. We can consider renaming it (potentially also as a breaking Go API change) when we tackle https://github.com/grafana/k6/issues/2544

js/modules/modules.go
```diff
@@ -95,20 +95,14 @@ type VU interface {
...
-	// possibly letting you call RegisterCallback again.
-	// Calling the result can be done at any time. The event loop will block until that happens, (if it doesn't
-	// have something else to run) so in the event of an iteration end or abort (for example due to an exception),
-	// It is the module responsibility to monitor the context and abort on it being done.
-	// This still means that the returned function here *needs* to be called to signal that the module
-	// has aborted the operation and will not do anything more, not doing so will block k6.
+	// RegisterCallback lets a JS module declare that it wants to run a function
+	// on the event loop *at a later point in time*. See the documentation for
+	// `EventLoop.RegisterCallback()` in the `k6/js/eventloop` Go module for
+	// the very important details on its usage and restrictions.
```
mstoykov, line 101:
```suggestion
	// RegisterCallback lets a JS module declare that it wants to run a function
	// on the event loop *at a later point in time*. See the documentation for
	// [*go.k6.io/k6/js/eventloop/EventLoop.RegisterCallback]  for
	// the very important details on its usage and restrictions.
```

Not certain how well this works but apperntly this is the [syntax for doc links](https://go.dev/doc/comment) (need to search for "doc links" as there are no anchors :( ) It seems we might drop everything before `eventloop` too :thinking: but we need to probably test this.

mstoykov: Uff .. thsi won't work for this as this is a method on an interface >.>

mstoykov (approved):

oleiade (approved):

olegbespalov (approved):

Issue #2247, opened 2021-11-18
WebSocket connection does not respect the `throw` option


na--:
### k6 version

all

### OS

all

### Expected Behaviour

With this code:
```js
import ws from 'k6/ws';

export let options = {
    throw: false,
};

export default function () {
    try {
        const resp = ws.connect('wss://test-api.k6.io/whatever-non-ws-url', null, function (socket) {
            socket.on('open', function () {
                console.log('connected');
                socket.close();
            });

            socket.on('error', function (e) {
                console.error('connection error occurred: ', e.error());
            });

            socket.on('close', function () {
                console.log('disconnected');
            });
        });
        console.log(`ended normally, response: \n${JSON.stringify(resp, null, 4)}`);
    }
    catch (e) {
        console.error(`ended with exception, error: ${e}`);
    }
}
```

I expect the `ws.connect()` call to return normally and emit a warning log message, and the `Response` to contain an appropriate error and error code. This is what happens with HTTP requests (which the websocket starts with) when `throw` is `false`.

Related issue: https://github.com/grafana/k6/issues/2024

### Actual Behaviour

An exception will be thrown and `ended with exception, error: websocket: bad handshake` logged.

### Steps to Reproduce the Problem

Run the code above.

na--:
This was fixed by https://github.com/grafana/k6/pull/2616

PR #2616, opened 2022-07-25
fix: fix ws not support throw option

fatelei:
# What is the pr for ?
Fix issue #2247 

# Test

## With `throw = false`
image

## With `throw = true`
image

### DIFF
--- a/js/modules/k6/ws/ws.go
+++ b/js/modules/k6/ws/ws.go
@@ -289,9 +289,7 @@ func (mi *WS) Connect(url string, args ...goja.Value) (*WSHTTPResponse, error) {
 		if state.Options.Throw.Bool {
 			return nil, connErr
 		}
-		if state.Logger != nil {
-			state.Logger.WithField("error", connErr).Warn("Ws Connection Failed")
-		}
+		state.Logger.WithField("error", connErr).Warn("Ws Connection Failed")
 		return &WSHTTPResponse{
 			Error: connErr.Error(),
 		}, nil
--- a/js/modules/k6/ws/ws.go
+++ b/js/modules/k6/ws/ws.go
@@ -289,7 +289,9 @@ func (mi *WS) Connect(url string, args ...goja.Value) (*WSHTTPResponse, error) {
 		if state.Options.Throw.Bool {
 			return nil, connErr
 		} else {
-			state.Logger.WithField("error", connErr).Warn("Ws Connection Failed")
+			if state.Logger != nil {
+				state.Logger.WithField("error", connErr).Warn("Ws Connection Failed")
+			}
 			return nil, nil
 		}
 	}
--- a/js/modules/k6/ws/ws.go
+++ b/js/modules/k6/ws/ws.go
@@ -292,7 +292,7 @@ func (mi *WS) Connect(url string, args ...goja.Value) (*WSHTTPResponse, error) {
 			if state.Logger != nil {
 				state.Logger.WithField("error", connErr).Warn("Ws Connection Failed")
 			}
-			return nil, nil
+			return &WSHTTPResponse{}, nil
 		}
 	}
 
--- a/js/modules/k6/ws/ws.go
+++ b/js/modules/k6/ws/ws.go
@@ -288,12 +288,13 @@ func (mi *WS) Connect(url string, args ...goja.Value) (*WSHTTPResponse, error) {
 		socket.handleEvent("error", rt.ToValue(connErr))
 		if state.Options.Throw.Bool {
 			return nil, connErr
-		} else {
-			if state.Logger != nil {
-				state.Logger.WithField("error", connErr).Warn("Ws Connection Failed")
-			}
-			return &WSHTTPResponse{}, nil
 		}
+		if state.Logger != nil {
+			state.Logger.WithField("error", connErr).Warn("Ws Connection Failed")
+		}
+		return &WSHTTPResponse{
+			Error: connErr.Error(),
+		}, nil
 	}
 
 	// Run the user-provided set up function
--- a/js/modules/k6/ws/ws.go
+++ b/js/modules/k6/ws/ws.go
@@ -286,8 +286,12 @@ func (mi *WS) Connect(url string, args ...goja.Value) (*WSHTTPResponse, error) {
 	if connErr != nil {
 		// Pass the error to the user script before exiting immediately
 		socket.handleEvent("error", rt.ToValue(connErr))
-
-		return nil, connErr
+		if state.Options.Throw.Bool {
+			return nil, connErr
+		} else {
+			state.Logger.WithField("error", connErr).Warn("Ws Connection Failed")
+			return nil, nil
+		}
 	}
 
 	// Run the user-provided set up function
--- a/js/modules/k6/ws/ws.go
+++ b/js/modules/k6/ws/ws.go
@@ -289,7 +289,7 @@ func (mi *WS) Connect(url string, args ...goja.Value) (*WSHTTPResponse, error) {
 		if state.Options.Throw.Bool {
 			return nil, connErr
 		}
-		state.Logger.WithField("error", connErr).Warn("Ws Connection Failed")
+		state.Logger.WithError(connErr).Warn("Attempt to establish a WebSocket connection failed")
 		return &WSHTTPResponse{
 			Error: connErr.Error(),
 		}, nil
--- a/js/modules/k6/ws/ws_test.go
+++ b/js/modules/k6/ws/ws_test.go
@@ -1346,9 +1346,9 @@ func TestWSConnectWithThrowErrorOption(t *testing.T) {
 			});
 		});
 		`)
+		require.NoError(t, err)
 		entries := logHook.Drain()
 		require.Len(t, entries, 1)
 		assert.Contains(t, entries[0].Message, "Attempt to establish a WebSocket connection failed")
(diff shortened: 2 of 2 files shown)


codebien:
Hi @fatelei, thanks for your contributon :pray:

Can you add a test coverage, please?

In any case, we plan to merge soon a new WebSocket API, this would be a breaking change. @mstoykov Does it make sense at this point considering the upcoming API?

fatelei:
> Hi @fatelei, thanks for your contributon 🙏
> 
> Can you add some test coverage, please?
> 
> In any case, we plan to merge soon a new WebSocket API, this would be a breaking change. @mstoykov Does it make sense at this point considering the upcoming API?

unit test has been added

js/modules/k6/ws/ws.go
```diff
@@ -286,8 +286,14 @@ func (mi *WS) Connect(url string, args ...goja.Value) (*WSHTTPResponse, error) {
...
 		socket.handleEvent("error", rt.ToValue(connErr))
-
-		return nil, connErr
+		if state.Options.Throw.Bool {
+			return nil, connErr
+		} else {
+			if state.Logger != nil {
+				state.Logger.WithField("error", connErr).Warn("Ws Connection Failed")
+			}
+			return &WSHTTPResponse{}, nil
```
mstoykov, line 295: Part of the issue is to also return a response containing the error so that the javascript code can do something based on that.

This kinds of threads on https://github.com/grafana/k6/issues/2024 as well and adding an `error_code` and `error` will require (as mentioned in that issue) for the error code to be moved somewhere and exported. This likely is a bit too much work so I am okay with not being part of this PR. 

But we definitely want the response to at least have an `error` field.

js/modules/k6/ws/ws_test.go
```diff
@@ -1288,3 +1289,40 @@ func TestCookieJar(t *testing.T) {
...
+		RuntimeField: rt,
+		StateField:   state,
+	})
+	require.NoError(t, rt.Set("ws", m.Exports().Default))
+	_, err = rt.RunString(`
+		var res = ws.connect("INVALID", function(socket){
+			socket.on("open", function() {
+				socket.close();
+			});
+		});
```
mstoykov, line 1325: You should likely check that the response is non `null` and has an `error` in it.

mstoykov (changes requested):
Thanks for the PR @fatelei :bow: 
I have left some comments - you will also need to fix your linters

@codebien as I mention the fuller fix for this will require for error codes to be moved around and is arguably a bigger change. Likely one too touchy and big for an outside contributor :(. 

I don't think we will be removing this API any time soon so any fixes for it are welcome as long as they aren't overly complex and not actually all the useful. We will definitely prefer to push the newer API. But the old one is still used by people so we will need to support at least until we have actually released a stable version of the new one IMO.

codebien (changes requested):

js/modules/k6/ws/ws_test.go
```diff
@@ -1288,3 +1289,58 @@ func TestCookieJar(t *testing.T) {
...
+		if (res && res.error) {
+			throw new Error(res.error);
+		}
+		`)
+		assert.Error(t, err)
```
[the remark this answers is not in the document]
fatelei: ok

js/modules/k6/ws/ws.go
```diff
@@ -286,8 +286,15 @@ func (mi *WS) Connect(url string, args ...goja.Value) (*WSHTTPResponse, error) {
...
 		// Pass the error to the user script before exiting immediately
 		socket.handleEvent("error", rt.ToValue(connErr))
-
-		return nil, connErr
+		if state.Options.Throw.Bool {
+			return nil, connErr
+		}
+		if state.Logger != nil {
+			state.Logger.WithField("error", connErr).Warn("Ws Connection Failed")
+		}
```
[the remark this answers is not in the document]
codebien: This is only partially resolved, WithError and a clearer message should be used.

js/modules/k6/ws/ws_test.go
```diff
@@ -1288,3 +1291,64 @@ func TestCookieJar(t *testing.T) {
...
+		var res = ws.connect("INVALID", function(socket){
+			socket.on("open", function() {
+				socket.close();
+			});
+		});
+		`)
+		entries := logHook.Drain()
+		require.Len(t, entries, 1)
+		assert.Contains(t, entries[0].Message, "Ws Connection Failed")
+		assert.NoError(t, err)
```
codebien, line 1352:
```suggestion
		require.NoError(t, err)
		entries := logHook.Drain()
		require.Len(t, entries, 1)
		assert.Contains(t, entries[0].Message, "Ws Connection Failed")
```

js/modules/k6/ws/ws_test.go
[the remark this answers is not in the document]
codebien: Why did we remove the part of the check?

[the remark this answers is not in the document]
fatelei: misunderstand

codebien:
@fatelei you can run in local `make lint` so you don't have to wait for a CI run.

codebien (dismissed):
LGTM

js/modules/k6/ws/ws_test.go
```diff
@@ -1288,3 +1291,70 @@ func TestCookieJar(t *testing.T) {
...
+				socket.close();
+			});
+		});
+		`)
+		assert.Error(t, err)
+	})
+
+	t.Run("ThrowDisabled", func(t *testing.T) {
+		t.Parallel()
+		state.Options.Throw = null.BoolFrom(false)
```
mstoykov, line 1344: This will `race` with the other setting in the previous test as you are setting the same object.'

I would recommend just using `newTestState` and creating a new one in each of the subtests

mstoykov (changes requested):

js/modules/k6/ws/ws_test.go
```diff
@@ -1288,3 +1291,48 @@ func TestCookieJar(t *testing.T) {
...
+
+	logHook := &testutils.SimpleLogrusHook{HookedLevels: []logrus.Level{logrus.WarnLevel}}
+	testLog := logrus.New()
+	testLog.AddHook(logHook)
+	testLog.SetOutput(io.Discard)
+
+	t.Run("ThrowEnabled", func(t *testing.T) {
+		t.Parallel()
+		ts := newTestState(t)
+		ts.state.Logger = testLog
```
mstoykov, line 1306: I don't think this is a problem - but this is still shared between the two and you seem to *not* use it. So maybe it shouldn't be set at all :shrug:

js/modules/k6/ws/ws_test.go
```diff
@@ -655,6 +657,7 @@ func TestErrors(t *testing.T) {
 		Dialer: tb.Dialer,
 		Options: lib.Options{
 			SystemTags: &metrics.DefaultSystemTagSet,
+			Throw:      null.BoolFrom(true),
```
mstoykov, line 660: Can you also set this to `true` in the `newTestState` as I would expect a lot of the other tests depend on this behaviour to report failures. So while they might not be failing now. If we break them - it might be needed for this to be set in order for them to be reported correctly.

mstoykov (changes requested):

js/modules/k6/ws/ws_test.go
fatelei: sounds great

js/modules/k6/ws/ws_test.go
fatelei: i only set `Logger` with throw option is false.

fatelei: if without `Logger`, when throw option is false, `ws` can not log warn message

js/modules/k6/ws/ws_test.go
```diff
@@ -1288,3 +1292,46 @@ func TestCookieJar(t *testing.T) {
...
 	assertSessionMetricsEmitted(t, metrics.GetBufferedSamples(ts.samples), "", sr("WSBIN_URL/ws-echo-someheader"), statusProtocolSwitch, "")
 }
+
+func TestWSConnectWithThrowErrorOption(t *testing.T) {
+	t.Parallel()
+
+	logHook := &testutils.SimpleLogrusHook{HookedLevels: []logrus.Level{logrus.WarnLevel}}
+	testLog := logrus.New()
+	testLog.AddHook(logHook)
+	testLog.SetOutput(io.Discard)
```
mstoykov, line 1302: I would argue you can move this in the `ThrowDisabled` as well. And arguably just splie this 1 test with 2 sub tests in two tests without any subTests :thinking:

fatelei: how about add two tests:
- TestWSConnectEnableThrowErrorOption: maybe not need ?
- TestWSConnectDisableThrowErrorOption

mstoykov: Sounds good. I think we can keep both and have them next to each other as contrasting tests.

You might also want to check that there is no log in the case where we throw :thinking:

mstoykov (approved):
LGTM! Thanks for getting this through :bow:

codebien (approved):
Thanks @fatelei for the effort :pray:

mstoykov:
I have marked this as "breaking change" as the default value for `throw` is `false` - so by default we will now not throw an exception while we previously would've.

PR #2621, opened 2022-07-28
Drop SetInitContext and instead have an a okay default value

mstoykov:


codebien (approved):

oleiade (approved):
That works for me 👍🏻 🙇🏻

PR #2624, opened 2022-07-31
#2565: Removed os packages dependency on console and used afero package

### LINKED ISSUE #2565: `console` uses `os` package directly
[js/console.go](https://github.com/grafana/k6/blob/aa064148a11fa70defc300b513a5a20f7bf21868/js/console.go#L44) uses directly the `os` package while almost everything uses an abstraction such as `afero`

We should be using afero across everything (until #1079) so this needs to change.

This likely can be done similarly to #2487

vanshaj:
1. Removed dependency of os package on console
2. Removed dependency of os package on console_test
3. Now using afero package in console and console_test

Closes #2565

### DIFF
--- a/js/console.go
+++ b/js/console.go
@@ -27,6 +27,7 @@ import (
 
 	"github.com/dop251/goja"
 	"github.com/sirupsen/logrus"
+	"github.com/spf13/afero"
 )
 
 // console represents a JS console implemented as a logrus.Logger.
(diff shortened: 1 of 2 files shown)


na-- (changes requested):
Thanks for making this PR, however it doesn't really solve the issue that https://github.com/grafana/k6/issues/2565 aims to solve. 

It doesn't really matter if `newFileConsole()` uses the `os` package directly or uses `afero.NewOsFs()`, it is basically the same thing. What needs to to happen is for it to accept an [`afero.Fs`](https://pkg.go.dev/github.com/spf13/afero#Fs) argument, so we can actually test it. In tests, we'll be able to pass it (or the code that uses it) an [`afero.NewMemMapFs()`](https://pkg.go.dev/github.com/spf13/afero#NewMemMapFs). In the actual k6 execution, we will be able to give it the real OS filesystem (i.e. `afero.NewOsFs()`), retrieved here and then threaded throughout the k6 codebase: https://github.com/grafana/k6/blob/48f66204988c316e845b8812dafc2cf596b53738/cmd/root.go#L134-L136

The end goal is that if we only access the real OS filesystem (and other OS attributes) through that one single `globalState` object in `cmd/root.go`, if we thread these throughout k6 and don't touch the real `os` package, then we will be able to write pretty good end-to-end integration tests of the whole of k6 by just mocking this `globalState` with test data, like this: 
https://github.com/grafana/k6/blob/48f66204988c316e845b8812dafc2cf596b53738/cmd/root_test.go#L30-L87
https://github.com/grafana/k6/blob/48f66204988c316e845b8812dafc2cf596b53738/cmd/integration_test.go#L104-L481

PR #2625, opened 2022-08-01
Fix broken tests and add an extra one

na--:
While doing a fairly hefty refactoring (https://github.com/grafana/k6/pull/2594#discussion_r934219137), I noticed that a couple of the currently existing tests are invalid or broken... :disappointed: I'll fix them commit by commit here before making another PR with the actual refactoring.


lib/executor/ramping_arrival_rate_test.go
```diff
@@ -713,7 +713,7 @@ func TestRampingArrivalRateGlobalIters(t *testing.T) {
 		BaseConfig:      BaseConfig{GracefulStop: types.NullDurationFrom(100 * time.Millisecond)},
 		TimeUnit:        types.NullDurationFrom(950 * time.Millisecond),
 		StartRate:       null.IntFrom(0),
-		PreAllocatedVUs: null.IntFrom(2),
+		PreAllocatedVUs: null.IntFrom(4),
```
mstoykov, line 716: Is this needed ? we should be spreading over the maxVUs either way :thinking:

na--: it is needed, otherwise the last segment will get 0 pre-initialized VUs and 1 max VUs, and will drop all of its iterations while it's trying to initialize that VU.

cmd/integration_test.go
```diff
@@ -389,3 +389,50 @@ func TestSubMetricThresholdNoData(t *testing.T) {
...
+	script := []byte(tb.Replacer.Replace(`
+		import http from "k6/http";
+		import { check } from "k6";
+		import { Counter } from "k6/metrics";
+
+		let statusCheck = { "status is 200": (r) => r.status === 200 }
+		let myCounter = new Counter("setup_teardown");
+
+		export let options = {
+			iterations: 5,
```
mstoykov, line 406: I guess we have a test that "fails" in the same way this is checking it won't fail?

lib/executor/ramping_arrival_rate_test.go
na--: As you say, the (re-)segmentation is based on MaxVUs, so nothing changes in the actual iteration numbers, but we actually also need a VU to run the planned iterations :sweat_smile:

cmd/integration_test.go
na--: Sorry, I don't understand what you mean

mstoykov (approved):

cmd/integration_test.go
mstoykov: This test does check that if we do something it *works*. We sometimes (probably should do it more) write the reverse test where we check that the reverse of what should work ... does not work. 

having read the test in more detail now ... I am not certain what that will be in this case though as we seem to not be checking that the threshold parsing works (which is what I though it did reading your commit message). But instead it seems to be about teardown and setup emitting metrics that can be used in thresholds.

So I am not certain how we can check that this test doesn't just pass as we do not actually check anything in it.  

I am also less certain that moving this to `cmd/` makes a difference :shrug:

na--: I see. The integration tests automatically check if the exit code is 0 by default, so if the thresholds had failed, it would catch it. If I change the `options.thresholds`, the test fails, when before it didn't.

Notice also the checkmarks in the end-of-test summary checks at the end, e.g. `✓ http_reqs......................: 7`. That's another guarantee that the thresholds were actually correctly parsed and evaluated that I added to the test.

na--: hmm though you are right, we don't have an integration test that checks if some thresholds have failed :thinking: I'll make a simple one in a bit

oleiade (dismissed):

cmd/integration_test.go
na--: resolved by https://github.com/grafana/k6/pull/2625/commits/ad33cbbc95e3f61af5c1a6b971156883c4ab6939

mstoykov (approved):

mstoykov (approved):

codebien (approved):
LGTM

oleiade (approved):

codebien:
@na-- A test failed but it seems not connected :thinking:

na--:
I re-ran it, since I plan to actually merge this with a merge commit, so I don't want any red crosses in the `master` history :sweat_smile:

PR #2627, opened 2022-08-01
Create distinct test state objects for the pre-init and run phases and thread them everywhere

na--:
While all of the commits here should be self-contained, a lot of them change the same lines of code multiple times, so it might be easier to just review the final diff.

This refactoring was prompted by desire in https://github.com/grafana/k6/pull/2594#discussion_r934219137 to be able to construct some object out of the final (i.e. consolidated, validated and derived) script `Options` and make it accessible to the various k6 components (e.g. executors, JS APIs, etc.) while the test is running. It turns out that it was basically impossible, and previous things that needed to be threaded through the codebase like that were done in an ad-hoc manner, mostly by adding extra function arguments wherever they were needed :scream: 

The only exception to that was the `lib.RuntimeState`, which was added to https://github.com/grafana/k6/pull/2487. Unfortunately, it only could contain things that were available before the initial init context execution and the creation of the (badly named) `Runner` :disappointed: Moreover, on a second look, the name `RuntimeState` was a bit misleading to me, since it clashes a bit with the goja JS "runtime" we have.

So, the TLDR version is that I decided go down this rabbit hole and rename `RuntimeState` to `TestPreInitState` and also introduce a new `TestRunState` struct that builds on top of, and to thread it everywhere I could :sweat_smile: We went from this: https://github.com/grafana/k6/blob/1e8468292ab5eb0d2333b1278b44e81c5cbef679/lib/runtime_state.go#L10-L18
To this: https://github.com/grafana/k6/blob/69a0b771db29f71176525675465262c2bfbdc133/lib/test_state.go#L10-L34

And from this:
https://github.com/grafana/k6/blob/1e8468292ab5eb0d2333b1278b44e81c5cbef679/cmd/run.go#L92
https://github.com/grafana/k6/blob/1e8468292ab5eb0d2333b1278b44e81c5cbef679/cmd/run.go#L128-L131
https://github.com/grafana/k6/blob/1e8468292ab5eb0d2333b1278b44e81c5cbef679/core/engine.go#L92

To this:
https://github.com/grafana/k6/blob/69a0b771db29f71176525675465262c2bfbdc133/cmd/run.go#L93
https://github.com/grafana/k6/blob/69a0b771db29f71176525675465262c2bfbdc133/cmd/run.go#L129
https://github.com/grafana/k6/blob/69a0b771db29f71176525675465262c2bfbdc133/core/engine.go#L89

**Besides the simplified invocations, the biggest benefit is that we can now easily add things to the `TestPreInitState` and `TestRunState` structs and have them available everywhere** :tada: Most of the rest of the changes here were just me fixing the endless tests that broke from the change APIs :sob:


lib/test_state.go
```diff
@@ -0,0 +1,34 @@
...
+	BuiltinMetrics *metrics.BuiltinMetrics
+	KeyLogger      io.WriteCloser
+
+	// TODO: replace with logrus.FieldLogger when all of the tests can be fixed
+	Logger *logrus.Logger
+}
+
+// TestRunState contains the pre-init state as well as all of the state and
+// options that are necessary for actually running the test.
+type TestRunState struct {
```
codebien, line 24:
```suggestion
type TestRun struct {
```
Do we really need the word `State`? Every struct is a state of something in the end.

na--: hmm though the `Runner` and `Options` are not exactly "state"... :sweat_smile: :man_shrugging: 

:+1: from me for renaming it to just `TestRun` - I am fine with either name, what do the rest of the @grafana/k6-devs think?

js/bundle_test.go
```diff
@@ -50,31 +50,44 @@ import (
...
+	var logger *logrus.Logger
 	for _, o := range opts {
 		switch opt := o.(type) {
 		case afero.Fs:
 			fs = opt
 		case lib.RuntimeOptions:
-			rtOpts = opt
-		case logrus.FieldLogger:
+			rtOpts = &opt
+		case *logrus.Logger:
```
mstoykov, line 79: Not certain if this doesn't break something ... maybe this `opts` should have a `default` case :thinking:

na--: Resolved by https://github.com/grafana/k6/pull/2627/commits/fe9f70ab769b0caeb31b07b86d95dd2b372ef436

cmd/test_load.go
```diff
@@ -192,59 +171,103 @@ func detectTestType(data []byte) string {
...
 }
 
-func (cw *syncWriter) Write(b []byte) (int, error) {
+func (cw *syncWriteCloser) Write(b []byte) (int, error) {
 	cw.m.Lock()
 	defer cw.m.Unlock()
 	return cw.w.Write(b)
 }
+
+func (cw *syncWriteCloser) Close() error {
```
mstoykov, line 269: nit: The idea behind this not having a `Close` method was that it makes no sense for something in the tests to be able to just close it.

core/local/local.go
```diff
@@ -105,9 +103,9 @@ func NewExecutionScheduler(
 	}
 
 	return &ExecutionScheduler{
-		runner:  runner,
-		logger:  logger,
-		options: options,
+		runner:       trs.Runner,
```
mstoykov, line 106: Why leave this one? is it used a lot?

mstoykov (approved):
LGTM in general :clap: !

 I have left some nitpick comments but nothing blocking. 

I do like how some of the signatures are now smaller and we have pre and post init state structs. 

But I am a bit worried that now that we provide a whole struct with k fields and not just the 2 things that will be used from it. This might lead to strange and hard to diagnose issues where in some test or the prod  code we forgot to set a field. Hopefully this will be caught by tests when the change is made.

core/local/local.go
na--: probably just forgot to remove it, will do so now, along with `options`

na--: Resolved by https://github.com/grafana/k6/pull/2627/commits/ec45155e8a195a1dd3e97f2a154768125d33574a

codebien (dismissed):

mstoykov (approved):

cmd/test_load.go
na--: ah, makes perfect sense, will restore the old behavior

na--: Resolved by https://github.com/grafana/k6/pull/2627/commits/a75766b398ef3ed87628362e22f715a50d70a525

lib/test_state.go
```diff
@@ -0,0 +1,34 @@
...
+import (
+	"io"
+
+	"github.com/sirupsen/logrus"
+	"go.k6.io/k6/metrics"
+)
+
+// TestPreInitState contains all of the state that can be gathered and built
+// before the test run is initialized.
+type TestPreInitState struct {
```
oleiade, line 12: I regret that the name is somewhat implementation specific (`PreInit` refers to "happens before `ExecutionScheduler.Init` is called right?), and would rather have it being something more specific to the commands themselves: we're trying to get k6 to perform some action, and this is the state we start in, before initializing any k6 components. Maybe `TestInitialState`, and documenting the context would do?

 ```suggestion
// TestRunInitialState contains the state gathered and consolidated before any k6 components were initialized.
//
// This state is useful because of A, B, C.
// This state is used in scenarios X, Y, Z.
// Use this state when you want to achieve "foo".
type TestRunInitialState struct {
    // RuntimeOptions holds the "runtime" configuration we want to pass into the test script (no-thresholds, etc...)
    RuntimeOptions RuntimeOptions
    
    // etc... 
}
```

I also noticed, as you mentioned in another comment, that we have somewhat of a mix of state and configuration in there too 👍🏻

lib/test_state.go
oleiade: Personally, I find `TestRunState` more explicit. It's somewhat consistent with the rest of our naming, where we already have things like `ExecutionState` and `RuntimeOptions`, with a clear distinction between `State` (something is running, and this is where its knobs are at), and `Options` (whenever you need to decide what behavior to observe, or which state to go in, look at those values).

lib/test_state.go
na--: Pre-init in the sense of https://k6.io/docs/using-k6/test-life-cycle, so it's just implementation specific. We have documented the init phase as the initial execution k6 does of the global scope, to get the exported script `options` and functions and such. So this struct basically contains most of the state and options we have gathered before we can execute this init phase.

oleiade (approved):
Besides a question about naming/documentation, LGTM 🚀 🦕

lib/test_state.go
na--: FWIW, I am not super happy with either `TestPreInitState` or `TestRunState` :sweat_smile: The nice thing is that, if we can figure out better names in the future, that will be an easy refactor to do, since their contents and method signatures won't change :tada:

codebien (approved):

PR #2631, opened 2022-08-03
lib/options: Migrated RunTags to map[string]string type

codebien:
It removes the dependency from SampleTags. In this way, it facilitates the transition to a more efficient SampleTags.

Related to: https://github.com/grafana/k6/pull/2594

### DIFF
--- a/core/engine_test.go
+++ b/core/engine_test.go
@@ -690,8 +690,14 @@ func TestRunTags(t *testing.T) {
 	t.Parallel()
 	tb := httpmultibin.NewHTTPMultiBin(t)
 
-	runTagsMap := map[string]string{"foo": "bar", "test": "mest", "over": "written"}
-	runTags := metrics.NewSampleTags(runTagsMap)
+	expectedRunTags := map[string]string{"foo": "bar", "test": "mest", "over": "written"}
+
+	// it copies the map so in the case the runner will overwrite
+	// some run tags' values it doesn't affect the assertion.
+	runTags := make(map[string]string)
+	for k, v := range expectedRunTags {
+		runTags[k] = v
+	}
 
 	script := []byte(tb.Replacer.Replace(`
 		import http from "k6/http";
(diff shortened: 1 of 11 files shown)


js/runner.go
```diff
@@ -861,3 +861,11 @@ func (s *scriptException) Hint() string {
 func (s *scriptException) ExitCode() exitcodes.ExitCode {
 	return exitcodes.ScriptException
 }
+
+func cloneRunTags(runTags map[string]string) map[string]string {
```
na--, line 865: this has nothing runTags specific, it might be better to just call it `copyStringMap(source map[string]string) map[string]string`?

core/engine_test.go
```diff
@@ -691,7 +691,7 @@ func TestRunTags(t *testing.T) {
 	tb := httpmultibin.NewHTTPMultiBin(t)
 
 	runTagsMap := map[string]string{"foo": "bar", "test": "mest", "over": "written"}
-	runTags := metrics.NewSampleTags(runTagsMap)
+	runTags := metrics.NewSampleTags(runTagsMap).CloneTags()
```
na--, line 694: why is this done? won't `runTagsMap` be exactly equivalent to `runTags`?

na-- (approved):
LGTM in general, just these 2 minor nitpicks

core/engine_test.go
codebien: In the case, there is a bug and the runner overwrites the run tags then the map would be changed and the assertion will pass, but it shouldn't.

I have updated it a bit the code and added a comment, so it would be clearer.

core/engine_test.go
```diff
@@ -690,8 +690,14 @@ func TestRunTags(t *testing.T) {
...
 	tb := httpmultibin.NewHTTPMultiBin(t)
 
-	runTagsMap := map[string]string{"foo": "bar", "test": "mest", "over": "written"}
-	runTags := metrics.NewSampleTags(runTagsMap)
+	runTags := map[string]string{"foo": "bar", "test": "mest", "over": "written"}
+
+	// it copies the map so in the case the runner will overwrite
+	// some run tags' values it doesn't affect the assertion.
+	expectedRunTags := make(map[string]string)
+	for k, v := range expectedRunTags {
```
na--, line 698:
```suggestion
	for k, v := range runTags {
```

I am pretty sure this should be `runTags`, but not why the test didn't fail? :sweat_smile:

mstoykov: Because the line below goes over `expectedRunTags` and checks that it's key,values are in the final tags - which when it's empty is always true

codebien: I have reversed them so it should be safer now

na-- (approved):

mstoykov (approved):

PR #2632, opened 2022-08-03
Keep the provided tags with a consistent order

codebien:
The new TagSet will be not optimal in the case the same set is provided with a different order. Let's keep them sorted as provided by the JavaScript Object.

Related to: https://github.com/grafana/k6/pull/2594#discussion_r929894515

[:warning: BREAKING CHANGE] It was possible to pass multiple tag sets to [add metric](https://k6.io/docs/javascript-api/k6-metrics/counter/counter-add), it was an error and it has been fixed in this PR.

### DIFF
--- a/js/modules/k6/metrics/metrics.go
+++ b/js/modules/k6/metrics/metrics.go
@@ -92,7 +92,7 @@ func limitValue(v string) string {
 	}, string(omitMsg))
 }
 
-func (m Metric) add(v goja.Value, addTags ...map[string]string) (bool, error) {
+func (m Metric) add(v goja.Value, addTags goja.Value) (bool, error) {
 	state := m.vu.State()
 	if state == nil {
 		return false, ErrMetricsAddInInitContext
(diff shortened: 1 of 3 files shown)


na-- (approved):
This conflicts with my local changes, but so many things do that it doesn't matter at this point :rofl: I'll just rebase once everything with atlas when the branch stabilizes, or maybe just copy-paste the relevant parts, that might be easier :sweat_smile:

js/modules/k6/metrics/metrics.go
```diff
@@ -129,9 +129,10 @@ func (m Metric) add(v goja.Value, addTags ...map[string]string) (bool, error) {
 	}
 
 	tags := state.CloneTags()
-	for _, ts := range addTags {
```
mstoykov, line 132: I guess this technically is a brekaing change as we previouls supported multiple maps .... but that seems like a bad idea to begin with - just maybe we should put in the release notes.

mstoykov (approved):

js/modules/k6/metrics/metrics.go
codebien: I updated the PR description

PR #2636, opened 2022-08-04
SECURITY.md

codebien:


oleiade (approved):
Glad to see this happen 👍🏻 

Do we have plans to also expose it as a `security.txt` file at the root of `https://k6.io` too as per good security practice?

na-- (approved):

na-- (changes requested):
Do not merge, we have a CI issue in master, we first have to merge https://github.com/grafana/k6/pull/2639 or something like it.

PR #2637, opened 2022-08-06
fix(js/runner): set RPSLimit if value is above zero. Closes #2613

### LINKED ISSUE #2613: rps=0 value causes a GoError
### Brief summary

Setting the above parameter in the options object causes a GoError to be thrown during the run. Tested this on both local runs and k6 Cloud, the behaviour is the same.

### k6 version

k6 v0.38.3

### OS

macOS 12.4

_No response_

setting an option `rps` in the options object to a value of 0

`export let options = {
    rps: 0,
    "stages": [
        {
            "duration": "5s",
            "target": 15
        },
        {
            "duration": "2m",
            "target": 15
        },
        {
            "duration": "5s",
            "target": 0
        }`
        
causes the following exeption:

`ERRO[0042] GoError: rate: Wait(n=1) exceeds limiter's burst 0
Run    [at go.k6.io/k6/js/modules/k6/http.(*RootModule).NewModuleInstance.func2 (native)
        at file:///Users/nobody/Documents/GitHub/Demo-scripts/SimpleDemo.js:34:12(5)
        at native
 executor=ramping-vus scenario=default  instance_id=1 lz="amazon:ie:dublin" source=stacktrace test_run_id=1429887
` 

### Expected behaviour

It would be expected either a validation/config error to be thrown out or that the `rps` limit is disabled/unlimited.

### Actual behaviour

the test runs successfully, but it throws out the errors constantly in the console:

`ERRO[0042] GoError: rate: Wait(n=1) exceeds limiter's burst 0
Run    [at go.k6.io/k6/js/modules/k6/http.(*RootModule).NewModuleInstance.func2 (native)
        at file:///Users/nobody/Documents/GitHub/Demo-scripts/SimpleDemo.js:34:12(5)
        at native
 executor=ramping-vus scenario=default  instance_id=1 lz="amazon:ie:dublin" source=stacktrace test_run_id=1429887
`

tbourrely:
Hello 🙂,

Here is my contribution regarding #2613 

Is this what you had in mind ?

### DIFF
--- a/js/runner.go
+++ b/js/runner.go
@@ -435,7 +435,7 @@ func (r *Runner) HandleSummary(ctx context.Context, summary *lib.Summary) (map[s
 func (r *Runner) SetOptions(opts lib.Options) error {
 	r.Bundle.Options = opts
 	r.RPSLimit = nil
-	if rps := opts.RPS; rps.Valid {
+	if rps := opts.RPS; rps.Valid && rps.Int64 > 0 {
 		r.RPSLimit = rate.NewLimiter(rate.Limit(rps.Int64), 1)
 	}
 
--- a/js/runner_test.go
+++ b/js/runner_test.go
@@ -43,6 +43,7 @@ import (
 	"github.com/spf13/afero"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"golang.org/x/time/rate"
 	"google.golang.org/grpc/test/grpc_testing"
 	"gopkg.in/guregu/null.v3"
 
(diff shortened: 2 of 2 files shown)


js/runner_test.go
```diff
@@ -151,6 +152,53 @@ func TestRunnerOptions(t *testing.T) {
...
+		variant := variant
+
+		t.Run(variant.name, func(t *testing.T) {
+			t.Parallel()
+
+			r, err := getSimpleRunner(t, "/script.js", `exports.default = function() {};`)
+			require.NoError(t, err)
+			err = r.SetOptions(variant.options)
+			require.NoError(t, err)
+			assert.Equal(t, variant.limiter, r.RPSLimit)
```
na--, line 197: hmm I am kind of surprised this works correctly for the `non-zero int` case :confused: I am not sure if testify is actually comparing the struct contents behind the pointers or what :thinking: 

in any case, it might be good enough to just compare if the limiter is `nil` or not `nil`

na-- (approved):
thanks!

js/runner_test.go
mstoykov: From the [docs](https://pkg.go.dev/github.com/stretchr/testify/require#Equal)

> Pointer variable equality is determined based on the equality of the referenced values (as opposed to the memory addresses). Function equality cannot be determined and will always fail. 

So this seems like it should work and apperantly it does.

js/runner.go
```diff
@@ -435,7 +435,7 @@ func (r *Runner) HandleSummary(ctx context.Context, summary *lib.Summary) (map[s
 func (r *Runner) SetOptions(opts lib.Options) error {
 	r.Bundle.Options = opts
 	r.RPSLimit = nil
-	if rps := opts.RPS; rps.Valid {
+	if rps := opts.RPS; rps.Valid && rps.Int64 > 0 {
```
mstoykov, line 438: I do wonder if it won't be better to log a warning/info that the rps is <=0 and consequently won't be applied :thinking:

mstoykov (approved):
LGTM ! thanks for the contribution :bow:

js/runner.go
na--: As I said in https://github.com/grafana/k6/issues/2613#issuecomment-1192501598, I don't consider zero or negative values as problematic, or worthy of a warning. I think it's usually clear enough when you give `0` or a negative number in such limit-related options that it means "unlimited".

mstoykov: :shrug: I would argue we need to warn on `rps` usage so that people just don't use it. But this is separate discussion ;). 

I guess you are write there will be no confusion, and I doubt there are that many users either way :shrug:

oleiade (approved):
LGTM 👍🏻 Thank you so much for your contribution 🙇🏻

PR #2638, opened 2022-08-08
Simplify the atlas integration and have a `TimeSeries` for every `metrics.Sample`

na--:
The two commits here implement the two decisions as described in https://github.com/grafana/k6/pull/2594#issuecomment-1205273413 and https://github.com/grafana/k6/pull/2594#issuecomment-1205359198, with some minor modifications. 

For example, instead of an `TagSet.Add()` method, I decided to name it `TagSet.With()`, which seems more well-suited to the immutable nature of the operation and less error-prone. I also refactored the mutable container of tags in every VU's `State` to have just 2 thread-safe operations, `GetCurrentValues() *metrics.TagSet` and a callback-based `Modify(callback func(currentTags *metrics.TagSet) (newTags *metrics.TagSet))`. I think the results look pretty nice: https://github.com/grafana/k6/blob/45223dffbbef51663faa1dd0d6db49e33684bd25/js/runner.go#L638-L654

(I also want to rename `SortAndAddTags()`, but I don't have a good idea for a better name :disappointed:)

### DIFF
--- a/.github/workflows/xk6-tests/xk6-js-test/jstest.go
+++ b/.github/workflows/xk6-tests/xk6-js-test/jstest.go
@@ -58,11 +58,10 @@ func (j *JSTest) Foo(arg float64) (bool, error) {
 
 	ctx := j.vu.Context()
 
-	tags := state.CloneTags()
-	tags["foo"] = "bar"
+	tags := state.Tags.GetCurrentValues().With("foo", "bar")
 	metrics.PushIfNotDone(ctx, state.Samples, metrics.Sample{
 		Time:   time.Now(),
-		Metric: j.foos, Tags: metrics.IntoSampleTags(&tags),
+		Metric: j.foos, Tags: tags,
 		Value: arg,
 	})
 
--- a/.github/workflows/xk6-tests/xk6-js-test/jstest.go
+++ b/.github/workflows/xk6-tests/xk6-js-test/jstest.go
@@ -60,9 +60,9 @@ func (j *JSTest) Foo(arg float64) (bool, error) {
 
 	tags := state.Tags.GetCurrentValues().With("foo", "bar")
 	metrics.PushIfNotDone(ctx, state.Samples, metrics.Sample{
-		Time:   time.Now(),
-		Metric: j.foos, Tags: tags,
-		Value: arg,
+		Time:       time.Now(),
+		TimeSeries: metrics.TimeSeries{Metric: j.foos, Tags: tags},
+		Value:      arg,
 	})
 
 	return true, nil
--- a/core/engine_test.go
+++ b/core/engine_test.go
@@ -80,7 +80,7 @@ func getTestRunState(
 		TestPreInitState: piState,
 		Options:          options,
 		Runner:           runner,
-		RunTags:          piState.Registry.BranchTagSetRootWith(options.RunTags),
+		RunTags:          piState.Registry.RootTagSet().SortAndAddTags(options.RunTags),
 	}
 }
 
@@ -302,7 +302,7 @@ func TestEngine_processSamples(t *testing.T) {
 				out <- metrics.Sample{
 					Metric: metric,
 					Value:  1.25,
-					Tags:   piState.Registry.BranchTagSetRootWith(map[string]string{"a": "1"}).SampleTags(),
+					Tags:   piState.Registry.RootTagSet().SortAndAddTags(map[string]string{"a": "1"}),
 				}
 				close(done)
 				return nil
@@ -342,7 +342,7 @@ func TestEngine_processSamples(t *testing.T) {
 				out <- metrics.Sample{
 					Metric: metric,
 					Value:  1.25,
-					Tags:   piState.Registry.BranchTagSetRootWith(map[string]string{"a": "1", "b": "2"}).SampleTags(),
+					Tags:   piState.Registry.RootTagSet().SortAndAddTags(map[string]string{"a": "1", "b": "2"}),
 				}
 				close(done)
 				return nil
@@ -368,7 +368,7 @@ func TestEngine_processSamples(t *testing.T) {
 
 		assert.Len(t, test.engine.MetricsEngine.ObservedMetrics, 2)
 		sms := test.engine.MetricsEngine.ObservedMetrics["my_metric{a:1}"]
-		assert.EqualValues(t, map[string]string{"a": "1"}, sms.Sub.Tags.CloneTags())
+		assert.EqualValues(t, map[string]string{"a": "1"}, sms.Sub.Tags.Map())
 
 		assert.IsType(t, &metrics.GaugeSink{}, test.engine.MetricsEngine.ObservedMetrics["my_metric"].Sink)
 		assert.IsType(t, &metrics.GaugeSink{}, test.engine.MetricsEngine.ObservedMetrics["my_metric{a:1}"].Sink)
@@ -398,7 +398,7 @@ func TestEngineThresholdsWillAbort(t *testing.T) {
 			out <- metrics.Sample{
 				Metric: metric,
 				Value:  1.25,
-				Tags:   piState.Registry.BranchTagSetRootWith(map[string]string{"a": "1"}).SampleTags(),
+				Tags:   piState.Registry.RootTagSet().SortAndAddTags(map[string]string{"a": "1"}),
 			}
 			close(done)
 			return nil
@@ -444,7 +444,7 @@ func TestEngineAbortedByThresholds(t *testing.T) {
 			out <- metrics.Sample{
 				Metric: metric,
 				Value:  1.25,
-				Tags:   piState.Registry.BranchTagSetRootWith(map[string]string{"a": "1"}).SampleTags(),
+				Tags:   piState.Registry.RootTagSet().SortAndAddTags(map[string]string{"a": "1"}),
 			}
 			<-ctx.Done()
 			close(done)
@@ -522,22 +522,20 @@ func TestEngine_processThresholds(t *testing.T) {
 				t, nil, nil, nil, lib.Options{Thresholds: thresholds}, piState,
 			)
 
-			tag1 := piState.Registry.BranchTagSetRoot()
-			tag1.AddTag("a", "1")
-			tag2 := piState.Registry.BranchTagSetRoot()
-			tag2.AddTag("b", "1")
+			tag1 := piState.Registry.RootTagSet().With("a", "1")
+			tag2 := piState.Registry.RootTagSet().With("b", "1")
 
 			test.engine.OutputManager.AddMetricSamples(
 				[]metrics.SampleContainer{
 					metrics.Sample{
 						Metric: gaugeMetric,
 						Value:  1.25,
-						Tags:   tag1.SampleTags(),
+						Tags:   tag1,
 					},
 					metrics.Sample{
 						Metric: counterMetric,
 						Value:  2,
-						Tags:   tag2.SampleTags(),
+						Tags:   tag2,
 					},
 				},
 			)
--- a/core/engine_test.go
+++ b/core/engine_test.go
@@ -187,10 +187,17 @@ func TestEngineRun(t *testing.T) {
 
 		runner := &minirunner.MiniRunner{
 			Fn: func(ctx context.Context, _ *lib.State, out chan<- metrics.SampleContainer) error {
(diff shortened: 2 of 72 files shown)


.github/workflows/xk6-tests/xk6-js-test/jstest.go
```diff
@@ -58,12 +58,11 @@ func (j *JSTest) Foo(arg float64) (bool, error) {
...
 
-	tags := state.CloneTags()
-	tags["foo"] = "bar"
+	tags := state.Tags.GetCurrentValues().With("foo", "bar")
 	metrics.PushIfNotDone(ctx, state.Samples, metrics.Sample{
-		Time:   time.Now(),
-		Metric: j.foos, Tags: metrics.IntoSampleTags(&tags),
-		Value: arg,
+		Time:       time.Now(),
+		TimeSeries: {Metric: j.foos, Tags: tags},
```
mstoykov, line 64:
```suggestion
		TimeSeries: metrics.TimeSeries{Metric: j.foos, Tags: tags},
```

na--: Yeah, I am just making a 3rd commit to fix the xk6 tests, forgot about them when developing locally :sweat_smile:

core/engine_test.go
```diff
@@ -187,10 +187,17 @@ func TestEngineRun(t *testing.T) {
 
 		runner := &minirunner.MiniRunner{
 			Fn: func(ctx context.Context, _ *lib.State, out chan<- metrics.SampleContainer) error {
-				metrics.PushIfNotDone(ctx, out, metrics.Sample{Metric: testMetric, Time: time.Now(), Value: 1})
+				metrics.PushIfNotDone(ctx, out, metrics.Sample{
+					TimeSeries: metrics.TimeSeries{Metric: testMetric, Tags: piState.Registry.RootTagSet()},
```
mstoykov, line 191: Given how much this need to be written ... maybera "constructor"?

na--: I thought about it, but it will become one of these constructors with 5 different parameters, where a struct initialization is just more readable 99% of the time. Moreover, when I add `metadata map[string]string`, which might be `nil`, I will need to add it as another parameter and edit all of the tests to add `nil` everywhere once again... So I decided to stick with this.

mstoykov: Maybe even for the whole sample as I see that we in some places have forgotten Time or Value in the past :facepalm:

metrics/tags.go
```diff
@@ -3,70 +3,143 @@ package metrics
...
+// discouraged and sequential usage of the AddTag() method should be preferred
+// and used whenever possible.
+//
+// The only place this method should be used is if we already have a
+// map[string]string of tags, e.g. with the test-wide --tags from the root, with
+// scenario.tags, with custom per-reqeust tags from a user, etc. Then it's more
+// efficient to sort their keys before we add them. If we don't, go map
+// iteration happens in pseudo-random order and this will generate a lot of
+// useless dead-end atlas Nodes.
+func (ts *TagSet) SortAndAddTags(m map[string]string) *TagSet {
```
na--, line 69: Does anyone have any better ideas for this name? `WithSortedTags()` is my only one, but it's a bit misleading

core/engine_test.go
na--: this is what I meant - a constructor for the whole sample will be super cumbersome :disappointed: especially with the upcoming `metadata`

.github/workflows/xk6-tests/xk6-js-test/jstest.go
na--: Resolved by https://github.com/grafana/k6/compare/7644b1c7dc4b05d84574ac315741dd43bc66cb2e..e1a641a1aa0c82d1d2f8db77611ccf4ba9bf301b

lib/netext/grpcext/conn.go
```diff
@@ -196,37 +196,36 @@ func (statsHandler) TagRPC(ctx context.Context, _ *grpcstats.RPCTagInfo) context
...
+	if stateRPC == nil {
+		// TODO: investigate this more, there has to be a way to fix it :/
+		stateRPC = &rpcState{tags: state.Tags.GetCurrentValues()}
 	}
 
 	switch s := stat.(type) {
 	case *grpcstats.OutHeader:
+		// TODO: figure out something better, e.g. via TagConn() or TagRPC()?
 		if state.Options.SystemTags.Has(metrics.TagIP) && s.RemoteAddr != nil {
+			stateRPC.ip = s.RemoteAddr.String()
```
mstoykov, line 212: Is this used?

lib/vu_state.go
```diff
@@ -93,65 +93,39 @@ type State struct {
...
-	tg.tags.Delete(k)
-}
-
-// Clone returns a map with the entire set of items.
-func (tg *TagMap) Clone() map[string]string {
-	tg.mutex.RLock()
-	defer tg.mutex.RUnlock()
-
-	return tg.tags.Map()
+	tg.tags = callback(tg.tags)
```
mstoykov, line 130: Not certain this is better than having `Set` and `Delete` The only place where this is used not for this (in the vu Activate) the lib.state should not be used by anything. So it will be safe to just not do it like this ;)

metrics/tags.go
```diff
@@ -3,70 +3,143 @@ package metrics
...
 }
 
-// TagSetFromSampleTags creates a TagSet starting
-// from the set defined by SampleTags.
-// Adding a tag on the new set doesn't impact the SampleTags.
-func TagSetFromSampleTags(st *SampleTags) *TagSet {
-	tm := &TagSet{}
-	tm.tags = st.tags
-	return tm
+// Without returns another TagSet, either an already existing one or a newly
```
mstoykov, line 29: While I really like both names, I feel like  they are so close that I will personally misread `Without` and `With` when going over code quickly. Does anyone else thinks this is a problem? Arguably it should be fairly obvious which one was intended to use, and so unlikely that you will actually not know what happens. 

nit: Also you can word both documentaiton in similar way :)

metrics/tags.go
mstoykov: Maybe just `AddTags` - we don't have a not sort variant and arguably this is an implementation detail. 

We can still document it with explanation why. But arguably users should not care and in general the provided function should do the "write thing" even if their name doesn't spell it out.

na--: Then it should be `WithTags()`? :thinking:

lib/vu_state.go
na--: `Set()` and `Delete()` and getting the current values are not atomic, we might have some weird indeterminate states with them, even some lost tags that get overwritten... This locks the `State.Tags` for the whole modification operation - you get the current state and emit the new state, no chance of a race.

lib/netext/grpcext/conn.go
na--: No, but my original plan was to actually add the `ip` tag in the `case *grpcstats.End:` below, that's why I added it :man_facepalming: This whole thing probably needs to be refactored, it seems very fragile, and that `stateRPC == nil` check is very suspect... :confused: In some other PR though, for now I'll remove the `stateRPC.ip`

metrics/tags.go
na--: :+1: for making the comments similar, :-1: for changing the names. They are dissimilar enough, and they accept a different number of parameters, so I don't think anyone will confuse them

lib/vu_state.go
mstoykov: Thats all nice and true ... but where does this matter? AFAIK all the cases we have are 
1. we have to set 1 thing irregardless of anything in the map
2. we have to get 1 thing irregardless of anything in the map
3. the initialization of this ... which doesn't need any of this as it's before there can be any race conditions. 

So for me this just forces you to use lambdas .. .and I know how much you love lambdas but that isn't a reason :P

metrics/tags.go
mstoykov: I guess to be similar ot the other names :+1: ... I do wonder though if someone won't decide to use that instead of `With` even when they don't need to :( 

How many places do we have where we use that? Can we maybe move it only there ... or at least move it out of the type - I think it will be better as a helper function and will make it more obvious that it should probably not be used. 

I am just a bit worried that somebody will have to add a tag and see `WithTags` and skip over `With`. Or have 2 tags and decide that instead of just calling `With` twice they will go with `WithTags` - effectively giving them self a speed bump.

So I kind of prefer if it's hard to find and definitely a thing you might read the documentation on.

core/engine_test.go
mstoykov: My original comment was about only `TimeSeries` 

I guess you won't be putting the `metadata` there so it might still be nicer :thinking:

lib/vu_state.go
```diff
@@ -93,65 +93,39 @@ type State struct {
...
 		mutex: sync.RWMutex{},
+		tags:  tags,
+		// metadata is intentionally nil by default
 	}
 }
 
-// BranchOut creates a TagSet starting from the state of the current Set.
-func (tg *TagMap) BranchOut() *metrics.TagSet {
+// GetCurrentValues returns the value of the VU tags in a thread-safe way.
+func (tg *VUStateTags) GetCurrentValues() *metrics.TagSet {
```
mstoykov, line 120: Maybe just call it `TagSet`  ?

lib/vu_state.go
na--: There are a few reasons why I don't want to change this:
- it might not matter anywhere _now_, but it might matter in the future, I don't know... this lambda is so easy to use that I don't consider it a problem worth changing and potentially shooting ourselves in the foot later
- it's more inline with the immutable nature of the atlas Nodes and `metrics.TagSet` - you disliked the mutable `TagSet` that was originally part of https://github.com/grafana/k6/pull/2594, why do you want a mutable
- this paradigm works much better with the unindexable tags, where through `k6/execution.vu.tags` a user actually might actually transform an indexed tag (stored in atlas) into an unindexed one (stored in the string metadata map), this probably should be atomic to avoid any issues

lib/vu_state.go
na--: :-1: for a couple of reasons:
- it is already obvious from the signature it will return a TagSet, so that will just duplicate it
- having "Current" in the name communicates that the value is mutable and might change

lib/netext/grpcext/conn.go
na--: Resolved by https://github.com/grafana/k6/pull/2638/commits/eec1bc9afb712ba699034f6d27a5eacb56e21017

metrics/tags.go
na--: As you can see from https://github.com/grafana/k6/pull/2638/commits/eec1bc9afb712ba699034f6d27a5eacb56e21017, it is used in quite a few places, though mostly in the tests... Only in a few places in the normal k6 flow, though fairly crucial ones. I think it's very well documented and we know to check for its usage during code reviews, so it should be fine

Resolved by https://github.com/grafana/k6/pull/2638/commits/eec1bc9afb712ba699034f6d27a5eacb56e21017?

metrics/tags.go
na--: Resolved by https://github.com/grafana/k6/pull/2638/commits/eec1bc9afb712ba699034f6d27a5eacb56e21017

metrics/tags.go
mstoykov: >  we know to check for its usage during code reviews, so it should be fine


Unfortunately we are not the only people who will use it ... I am a lot less worried about us, but every extension developer who decides to emit a metric and add a tag will need to figure this out. And if we have two methods that do the same thing on the type I am worried they will use the "wrong" one.

mstoykov: Also, now that you renamed it ... shouldn't it be `WithMap` or `WithTagsFromMap` ? Because `With` does add a tag as well ... that is the whole point.

mstoykov (approved):
I am okay with most of the changes - the ones that I am not okay with can be  changed/renamed later and if/when  someone else brings them forward again. 

I doubt even when we merge this we will never change most of this again, so getting everything blocked on this seems pointless

metrics/tags.go
na--: hmm :thinking: good point, :+1: for `WithTagsFromMap()`

na--: Resolved by the latest [force push](https://github.com/grafana/k6/compare/eec1bc9afb712ba699034f6d27a5eacb56e21017..bcc550324c84eb125e74870945ff33d9ce1314f8)

Issue #2623, opened 2022-07-29
Templating issue in latest k6 release


varghesevv:
### Brief summary

javascript templating with backtick symbol is not working on latest k6 releases. It is working up to release "grafana/k6:0.39.0-24-g0127a508", not anything after that

referenced the example from this url - https://k6.io/docs/test-types/load-testing

### k6 version

0.39.0-25-ge2316a28

### OS

linux ubuntu

### Docker version and image (if applicable)

grafana/k6:0.39.0-25-ge2316a28

### Steps to reproduce the problem

working fine with this command
```docker run -v ${PWD}:/test grafana/k6:0.39.0-24-g0127a508 run /test/test-1.js```

not working with this
```docker run -v ${PWD}:/test grafana/k6:latest run /test/test-1.js``` or
```docker run -v ${PWD}:/test grafana/k6:0.39.0-25-ge2316a28 run /test/test-1.js```
```
/* conf-test-1.js */
export const options = {
  stages: [
    { duration: '5s', target: 2 } 
  ],
  thresholds: {
    'http_req_duration': ['p(99)<1500']
  },
};

export function get_url(){
  return `${BASE_URL}/auth/token/login/`;
}


/* test-1.js */

import http from 'k6/http';
import { check, group, sleep } from 'k6';

import { get_url } from './conf-test-1.js'
export { options } from './conf-test-1.js'

const BASE_URL = 'https://test-api.k6.io';
const USERNAME = 'TestUser';
const PASSWORD = 'SuperCroc2020';

export default () => {
  const loginRes = http.post(get_url(), {
    username: USERNAME,
    password: PASSWORD,
  });

  const authHeaders = {
    headers: {
      Authorization: `Bearer ${loginRes.json('access')}`,
    },
  };

  const myObjects = http.get(`${BASE_URL}/my/crocodiles/`, authHeaders).json();
  check(myObjects, { 'retrieved crocodiles': (obj) => obj.length > 0 });

  sleep(1);
};


### Expected behaviour

Should work with javascript templates

### Actual behaviour

```time="2022-07-29T09:55:54Z" level=error msg="ReferenceError: BASE_URL is not defined\n\tat get_url (file:///test/conf-test-1.js:13:12(1))\n\tat file:///test/test-1.js:12:29(8)\n\tat native\n" executor=ramping-vus scenario=default source=stacktrace```

mstoykov:
[...]

PR #2639, opened 2022-08-08
Fix the pushing of docker images in the CI build process

na--:
Because of a mistake I made in https://github.com/grafana/k6/pull/2591, we were pushing `master` builds as stable `latest` and `vX.Y-Z` docker images, which was not the intention :disappointed: 

So now we have a bunch `master` builds here https://hub.docker.com/r/grafana/k6/tags and https://hub.docker.com/r/loadimpact/k6/tags :disappointed:  And I guess this is also why the user from https://github.com/grafana/k6/issues/2623 encountered the issue :thinking:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -153,33 +153,52 @@ jobs:
             docker run $DOCKER_IMAGE_ID pause --help
             docker run $DOCKER_IMAGE_ID resume --help
 
-      - name: Publish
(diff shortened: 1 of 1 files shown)


imiric (approved):
LGTM :crossed_fingers:

mstoykov (approved):

oleiade (approved):

PR #2645, opened 2022-08-15
Update goja

mstoykov:


js/modules/k6/html/html_test.go
```diff
@@ -822,7 +822,7 @@ func TestParseHTML(t *testing.T) {
 			v, err := rt.RunString(`doc.find("body").children().get(1)`)
 			if assert.NoError(t, err) {
 				elem, _ := v.Export().(Element)
-				assert.Contains(t, elem.InnerHTML(), "Lorem ipsum dolor sit amet")
```
na--, line 825: hmm how sure are we that this won't affect anything on the JS side? :thinking: I guess they get some string type and goja magically makes it work for JS, regardless of the internals?

mstoykov: the problem here was that the `goja.Value` that is now returned isn't just a wrapped `string` which it previously was, for this particular case - if it had ~unicode~ non-ascii it wouldn't have worked before as well.

oleiade (approved):

na-- (approved):

Issue #2650, opened 2022-08-18
--include-system-env-vars doesn't check for the invalid environment variables


olegbespalov:
### Brief summary

Currently, when using the `--include-system-env-vars` flag, it's possible to also include the environment variables with invalid names like `LOREM-IPSUM.` 

Ideally, the same validation criteria should be applied as those used for the `--env`. The only exception probably is instead of the erroring, we should print a warning that something like `environment variable LOREM-IPSUM` is invalid and will be ignored.

### k6 version

v0.39.0

### OS

linux

### Steps to reproduce the problem

```
env -i 'LOREM-ISPUM=hello' ./k6 archive --include-system-env-vars script.js
```

### Expected behaviour

LOREM-ISPUM will be excluded since it's invalid

### Actual behaviour

It stays in the archive's metadata

olegbespalov:
Also maybe as part of this issue, we can document grafana/k6-docs that we aren't accept invalid environment variables

mstoykov:
It seems a bit strange to me to enfornce what "valid" environment variable name is if we got it from the environment.

How exactly will we get such a variable and if we get it why would *we* argue it's invalid if we got it from the environment. From a quick test that doesn't seem to be possible in my local bash version btw. 


And if it is possible `k6 run` will be spewing warning with the changes from #2652 .

As far as I can see the only benefit from this change is that if somebody has shell that doesn't prevent the usage of invalid names, we will prevent them from using them in the script through `__ENV`. 

But in that case maybe just prevent `__ENV` from accessing variables with invalid names?

na--:
IIRC we have observed this problem with some weird CI environment that allowed for non-POSIX env vars... Vanilla Windows, Linux and *nix are thankfully saner. I can imagine they have all sorts of weird edge cases in custom CI pipelines...

mstoykov:
Okay then with the current approach we will *potentially* be throwing warnings in CI and it is very likely and nobody will even be able to (sanely) write scripts that will use it as it will be impossible to run them anywhere else but in those CIs.

na--:
Could be :thinking: But it's still a better alternative than incorporating these broken variables in the .tar archive and then being unable to run them in the cloud when someone runs `k6 cloud --include-system-env-vars=true script.js` with such a broken environment. At least there will be a warning.

[...]

PR #2652, opened 2022-08-18
Detecting and filtering invalid environment names

olegbespalov:
# What?

Align the `--include-system-env-vars` with the `--env` in terms of the name validation.

In case of detecting an invalid name issue, the user will be warned.

```
 env -i 'LOREM-ISPUM=hello' IPSUM=lorem ./k6 run --include-system-env-vars samples/http_2.js
```

As the result, archive metadata should contain only `IPSUM=lorem`.

# Why?

Closes: #2650

### DIFF
--- a/cmd/runtime_options.go
+++ b/cmd/runtime_options.go
@@ -5,6 +5,7 @@ import (
 	"regexp"
 	"strconv"
 
+	"github.com/sirupsen/logrus"
 	"github.com/spf13/pflag"
 	"gopkg.in/guregu/null.v3"
 
@@ -55,7 +56,11 @@ func saveBoolFromEnv(env map[string]string, varName string, placeholder *null.Bo
 	return nil
 }
 
-func getRuntimeOptions(flags *pflag.FlagSet, environment map[string]string) (lib.RuntimeOptions, error) {
+func getRuntimeOptions(
+	logger *logrus.Logger,
+	flags *pflag.FlagSet,
+	environment map[string]string,
+) (lib.RuntimeOptions, error) {
 	// TODO: refactor with composable helpers as a part of #883, to reduce copy-paste
 	// TODO: get these options out of the JSON config file as well?
 	opts := lib.RuntimeOptions{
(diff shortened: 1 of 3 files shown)


cmd/runtime_options_test.go
```diff
@@ -113,11 +118,12 @@ func TestRuntimeOptions(t *testing.T) {
 	runtimeOptionsTestCases := map[string]runtimeOptionsTestCase{
 		"empty env": {
 			useSysEnv: true,
+			systemEnv: map[string]string{},
 			// everything else is empty
 			expRTOpts: lib.RuntimeOptions{
 				IncludeSystemEnvVars: null.NewBool(true, false),
 				CompatibilityMode:    defaultCompatMode,
-				Env:                  nil,
+				Env:                  map[string]string{},
```
olegbespalov, line 126: I hope that this change is acceptable. I can't imagine that sys vars can be nil, and is it essential (having `nil` there)? :thinking:

cmd/runtime_options.go
```diff
@@ -104,7 +109,15 @@ func getRuntimeOptions(flags *pflag.FlagSet, environment map[string]string) (lib
 	}
 
 	if opts.IncludeSystemEnvVars.Bool { // If enabled, gather the actual system environment variables
-		opts.Env = environment
+		for k, v := range environment {
+			if !userEnvVarName.MatchString(k) {
+				logger.Warnf("invalid system environment variable name '%s'", k)
```
na--, line 114:
```suggestion
				logger.Warnf("The environment variable '%s' has an invalid name and was ignored and not included in the test", k)
```

oleiade (dismissed):
Same comment as @na-- on the error message, but besides that no comment. Looks good to me 👍🏻 🦖

na-- (approved):

mstoykov (approved):
LGTM, hopefully nobody will have some strange env variables :shrug: 

Can we get test for this new change (I don't see anything) and maybe for the `--env` checks if there isn't one yet.

oleiade (approved):

olegbespalov:
@mstoykov 

>Can we get test for this new change (I don't see anything) and maybe for the --env checks if there isn't one yet.

Completely missed that, my bad :facepalm: 

Fixed in the latest commit (test for the `--env` we have).

mstoykov (approved):

na-- (approved):

Issue #2653, opened 2022-08-18
Always throw exception deliberately for invalid `vu.tags` values in `k6/execution`


na--:
I found a minor bug in k6 v0.39.0 and the current `master`, but after some consideration, it is kind of a good thing :sweat_smile: 

If you have this script:

```js
import exec from 'k6/execution';

export let options = {
    throw: false, // the default
};

export default function () {
    exec.vu.tags["custom-tag"] = [123];
    console.log(exec.vu.tags);
}
```

After https://github.com/grafana/k6/pull/2558, we'd expect that it will just emit a warning and ignore the value, right? Unfortunately, because [we return false here](https://github.com/grafana/k6/blob/b89a3aaa6bff07907140e6b31880971aeef3570b/js/modules/k6/execution/execution.go#L337), what ends up happening is actually this:
```
WARN[0000] the execution.vu.tags.Set('custom-tag') operation has been discarded because only String, Boolean and Number types are accepted as a Tag value 
ERRO[0000] TypeError: 'Set' on a dynamic object returned false
	at file:///script.js:8:34(9)
	at native  executor=per-vu-iterations scenario=default source=stacktrace
```

The expected warning is there, but then also an exception despite `--throw` being false :sweat_smile: 

And while the fix to just a warning is easy enough, we can just return `true` in `tagsDynamicObject.Set()`, I am kind of re-evaluating our decision to rely on `--throw` here :thinking_face: The issue that the old PR fixed was that k6 was panicking when someone tried to set an invalid tag value, but I think we tried to overcompensate too much in the other direction when we tried to reduce the default behavior (when `--throw` is not enabled, which very few people do) to just a warning...

Put another way, there doesn't seem to be a reason to not always throwing an exception when we get an invalid tag value. And it won't even be a breaking change, since that is what ended up being done in v0.39.0 anyway, despite our best efforts :sweat_smile: FWIW, we didn't catch the issue because of https://github.com/grafana/k6/issues/2651

So, my position is that we shouldn't do anything for v0.40.0, since we already throw an exception (albeit, with an extra warning message), but that I should remove this `--throw` check in my PR for https://github.com/grafana/k6/issues/2584, so we can clean this up for k6 v0.41.0 and deliberately and always throw an exception when we get such invalid values. These script errors could potentially invalidate the whole test run, so it's best to make them very visible.

PR #2654, opened 2022-08-18
Support non-indexable high-cardinality metric tags/metadata

na--:
This is built on top of https://github.com/grafana/k6/pull/2594 and should close https://github.com/grafana/k6/issues/2584 and fix https://github.com/grafana/k6/issues/2653 when it's done. 

One change from the original issue description has been the decision to have an indexable `url` tag, but to always have its value be equal to the `name` tag. See this for the details and the reason why: https://github.com/grafana/k6/issues/2584#issuecomment-1219618111

One of the main use cases for this support of high-cardinality metric tags/metadata is to allow k6 to support traces and to export the trace_id and span_id values it generated to something external like Tempo and Jaeger. The goal is to eventually have this with the new k6 HTTP API (https://github.com/grafana/k6/issues/2584). However, this PR will even allow us to implement tracing purely in JavaScript, e.g. in the [httpx library](https://k6.io/docs/javascript-api/jslib/httpx/) from [jslib.k6.io](https://jslib.k6.io/) (https://github.com/grafana/k6-jslib-httpx/issues/9). Here is how something like that might look like:
```js
import { Httpx } from 'https://jslib.k6.io/httpx/1.2.3/index.js';

export default function () {
    const client = new Httpx({
        baseURL: 'https://httpbin.test.k6.io',
        headers: {
            'Content-Type': 'application/json',
        },
        timeout: '10s',
        tracing: true,
    });

    client.post(`/anything`, { name: 'Bert' }); // per-request trace and span IDs get generated

    exec.vu.tags.trace_id = {
        value: somehowGenerateTraceID(),
        index: false
    }

    // These would be different spans but part of the same trace
    client.post(`/anything`, { val: 'foo' }); 
    client.post(`/anything`, { val: 'bar' }); 
    client.post(`/anything`, { val: 'baz' }); 

    exec.vu.tags.span_id = {
        value: somehowGenerateSpanID(),
        index: false
    }

    // All of the concurrent batch requests will be part of the same span
    // maybe we're doing some sort of a map-reduce logic or something like that that we consider a single transaction
    client.batch( /* ... a bunch of requests ... */ )

    // Clean up, since we don't want the same trace between iterations (at least in this example).
    delete exec.vu.tags.span_id 
    delete exec.vu.tags.trace_id
}
```


metrics/system_tag.go
```diff
@@ -7,61 +7,70 @@ import (
...
 	TagExpectedResponse
 
 	// System tags not enabled by default.
-	TagIter
-	TagVU
+	TagIter // non-indexable
+	TagVU   // non-indexable
 	TagOCSPStatus
-	TagIP
+	TagIP // non-indexable
```
na--, line 40: Hmm, actually, I am not sure that IP should be non-indexable :thinking: it is not enabled by default, and generally it shouldn't be too high cardinality, right? Even a geo-distributed CDN will probably only have dozens of unique IPs in a load test :thinking:

codebien: I think we can start as a non-indexable and maybe change in the future. I think the reverse could be more complicated. In any case, the user can force to have it indexed if it's required, right?

js/common/tags.go
```diff
@@ -0,0 +1,79 @@
+package common
```
codebien, line 1:
```suggestion
package tagging
```
I think we could create a dedicated package `tags` or `tagging` for it, so we could use like `tags.ApplyUserDefined`

codebien (commented):
Happy to see it :clap:

First iteration for me with a few comments.

metrics/sample.go
```diff
@@ -19,9 +19,9 @@ type TimeSeries struct {
 // A Sample is a single metric measurement.
 type Sample struct {
 	TimeSeries
-	Time  time.Time
-	Value float64
-	// TODO: add Metadata map[string]string, which could be nil
+	Time     time.Time
+	Value    float64
+	Metadata map[string]string // could be nil
```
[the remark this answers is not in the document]
na--: Yeah, more comments and explanation here are definitely a good idea! :+1:

js/common/tags.go
na--: hmm :thinking: maybe, though this is something very specific for the JS packages and it's just a couple of helper functions :thinking: calling it `tags` or `tagging` might occupy a valuable package name in the future :sweat_smile:

metrics/system_tag.go
na--: No, you currently can't force a system tag to be indexed. It would probably be fairly simple to make that configurable, but I am not sure we should, it might lead to more issues than it solves... :thinking:

na--:
Actually, besides addressing PR comments and maybe a few more tests, I am not sure if anything else is needed in this PR here :thinking: So I'll remove the draft status of this PR, it's officially ready for reviews :tada:

lib/netext/httpext/request.go
```diff
@@ -187,13 +181,18 @@ func MakeRequest(ctx context.Context, state *lib.State, preq *ParsedHTTPRequest)
...
 
-	tracerTransport := newTransport(ctx, state, tags, preq.ResponseCallback)
+	tracerTransport := newTransport(ctx, state, &preq.TagsAndMeta, preq.ResponseCallback)
 	var transport http.RoundTripper = tracerTransport
 
 	if state.Options.HTTPDebug.String != "" {
 		// Combine tags with common log fields
 		combinedLogFields := map[string]interface{}{"source": "http-debug", "vu": state.VUID, "iter": state.Iteration}
-		for k, v := range tags.Map() {
+		for k, v := range preq.TagsAndMeta.Tags.Map() {
```
olegbespalov, line 190: Maybe create a small method `TagsAndMeta.All() map[string]` to consolidate the access/management inside the `TagsAndMeta` struct and expose less outside? WDYT?

js/common/tags.go
mstoykov: Maybe than move it to the `common` package?

js/modules/k6/execution/execution.go
```diff
@@ -215,14 +215,19 @@ func (mi *ModuleInstance) newVUInfo() (*goja.Object, error) {
...
-	// (stored in the TagSet) or false if not (stored in the metadata string
-	// map) or undefined if the tag doesn't exist?
-	// tagsObj.SetPrototype(...)
+	// (stored in the TagSet), or false if not (stored in the metadata string
+	// map), or undefined if the tag doesn't exist.
+	proto := rt.NewObject()
+	if err := proto.Set("isIndexed", tags.isIndexed); err != nil {
+		return o, err
+	}
+	if err := tagsObj.SetPrototype(proto); err != nil {
```
mstoykov, line 226: Is there a particular reason to be on the Prototype ? Not that it's wrong I am just asking if you shouldn't just do `o.Set` instead :shrug:

js/common/tags.go
na--: It already is in the `common` package? :confused:

metrics/system_tag.go
```diff
@@ -7,61 +7,70 @@ import (
...
 )
 
 // DefaultSystemTagSet includes all of the system tags emitted with metrics by default.
 // Other tags that are not enabled by default include: iter, vu, ocsp_status, ip
 //nolint:gochecknoglobals
-var DefaultSystemTagSet = TagProto | TagSubproto | TagStatus | TagMethod | TagURL | TagName | TagGroup |
-	TagCheck | TagError | TagErrorCode | TagTLSVersion | TagScenario | TagService | TagExpectedResponse
+var DefaultSystemTagSet = SystemTagSet(
+	TagProto | TagSubproto | TagStatus | TagMethod | TagURL | TagRawURL | TagName | TagGroup |
```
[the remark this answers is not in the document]
na--: Hmm good point :thinking: I think I included it by default in an attempt to minimize the breaking change - we might be changing the `url`, but people would still have access to the `raw_url` :thinking:

lib/netext/httpext/transport.go
```diff
@@ -80,68 +80,64 @@ func (t *transport) measureAndEmitMetrics(unfReq *unfinishedRequest) *finishedRe
...
+		// However, if the user set the `name` tag value somehow, we will use
+		// whatever they set as the value of the `url` tags too, to prevent
+		// high-cardinality values in the indexed tags.
+		ctm.SetSystemTagOrMetaIfEnabled(enabledTags, metrics.TagURL, nameTagValue)
+
+		// However, if the actual clean-URL value would have been different than
+		// the `name` value, we will also attach it as the `raw_url` non-indexed
+		// tag (i.e. add it to the Sample.Metadata).
+		if nameTagValue != cleanURL {
+			ctm.SetSystemTagOrMetaIfEnabled(enabledTags, metrics.TagRawURL, cleanURL)
```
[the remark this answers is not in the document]
na--: :+1: for points 2. and 3., I will fix them.

I am not so sure about point 1. though :thinking: We don't always set all possible tags on all metric samples. For example, we only set the `error` and `error_code` tags if there was actually an error with the request. So I think it makes some sense to only set the `raw_url` tag when the actual URL of the HTTP request is going to be different from the value of the `url` tag, i.e. when there is URL grouping with `name`. 

> which means that it will be hard to use for the people who might want to use it directly

Why? Users can't set a threshold on it, and `Response.url` will be the actual URL. So it's mostly going to be exposed for things like the `json` and `csv` outputs and potentially for exporting tracing data, where it doesn't seem like a problem. Besides, a user can manually set the `raw_url` tag on requests that won't automatically have it assigned, if they really want to.

So, I am leaning towards :-1: for point 1. :thinking: @grafana/k6-devs, what do you think?

js/modules/k6/execution/execution.go
na--: A couple of reasons. If I try, because of the `goja.DynamicObject` validation logic for `Set()`, I will get this error :sweat_smile: 
```
TypeError: setting 'isIndexed' metric tag failed; only String, Boolean and Number types, or objects with the 'value' and 'index' properties, are accepted as a metric tag values
```

But even if I work around this, I think it's better if `Object.keys(execution.vu.tags)` doesn't return `isIndexed`

mstoykov: Hmm I would argue this needs to go in `tagsDynamicObject#Get` and you should check `isIndexed` and return  ToValue(o.isIndexed)` isntead of adding a whole new prototyupe

na--: Hm OK, I can probably do that, but it feels like a hack and I don't understand what the problem with the prototype is? :confused:

js/modules/k6/execution/execution.go
```diff
@@ -297,72 +307,81 @@ func optionsAsObject(rt *goja.Runtime, options lib.Options) (*goja.Object, error
...
+
+func (o *tagsDynamicObject) isIndexed(key string) goja.Value {
+	tcv := o.state.Tags.GetCurrentValues()
+	if _, ok := tcv.Tags.Get(key); ok {
+		return o.runtime.ToValue(true)
+	}
+	if _, ok := tcv.Metadata[key]; ok {
+		return o.runtime.ToValue(false)
+	}
+	return goja.Undefined()
```
mstoykov, line 322: Ahhh the famous three valued boolean :tada: 


Not certain this won't be more confusing then not - I did expect that this will tell me if `url` is indexed but it gave me `undefined` which was super confusing for me. I guess with some documentation it will be easier. 

And I guess we should have another API that answeres teh question "is this tag globally indexable or not".

js/modules/k6/execution/execution.go
mstoykov: > it feels like a hack 

Well the whole DynamicObject is built on this hack ;)

> I don't understand what the problem with the prototype is? 

You can definitely do it, but:
1. you don't need to as there is an easier way
2. you should  use [DefinedAccessorProperty](https://pkg.go.dev/github.com/dop251/goja#Object.DefineAccessorProperty) and set `enumerable` to goja.FLAG_False as otherwise `for (let l in exec.vu.tags)` will still catch it (even though it is won't come in `Object.keys`)
3. I would prefer that when we start setting prototypes we do actually create *one* prototype object and set it across althe instances of that particular "type". In this case this is likely an overkill, but so is using a prototype ;)

js/modules/k6/execution/execution.go
na--: Yeah... This API only "knows" about the VU-level tags, which aren't all that many. I even hesitated adding the API, since pretty much all such non-indexed tags will be set by the users, so they probably don't really need a way to query them...

> And I guess we should have another API that answeres teh question "is this tag globally indexable or not".

That may be a... complicated question to answer :sweat_smile: Theoretically, someone could disable an indexed system tag with `--system-tags` and then manually supply tag values with the same name... They will probably shoot themselves in the foot doing that, but they could, and I have no idea what such an API should return then... :sweat_smile:

js/modules/k6/execution/execution.go
na--: hmm I tried to do this, but now I get this error :sweat_smile: 
```
TypeError: Dynamic objects do not support accessor properties
```

mstoykov: I meant `2` as in if you are using prototype - you should use `Define...` instead of `Set`

na--: Ah, I see... :man_facepalming: :+1: I will try to make a single prototype per VU/runtime and use that then.

lib/netext/httpext/transport.go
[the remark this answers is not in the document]
codebien: An alternative suggestion. Could it be an option to not include `raw_url` in the core? The user has the option to add it manually so we could apply it in this way for now. The tracing library could do the same. We could provide a js helper in utils if it's really required. WDYT?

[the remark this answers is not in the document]
na--: hmm, implementing this purely in JS is going to be tricky, e.g. if you have a `name` tag to some request that has redirects :thinking:

js/common/tags.go
na--: Ping @mstoykov - can you please explain what you meant here?

metrics/system_tag.go
na--: FYI my intention is to roll back this change and leave the IP address as an indexable (i.e. normal) tag

js/common/tags.go
mstoykov: Oh ... I meant that I am terrible at reading code before i finish my second :coffee: ... and was of the impression that it is in folder `tags` :facepalm: 

tl;dr-ignore me

metrics/system_tag.go
```diff
@@ -7,61 +7,70 @@ import (
...
 	TagSubproto
 	TagStatus
 	TagMethod
 	TagURL
+	TagRawURL // non-indexable
 	TagName
 	TagGroup
 	TagCheck
-	TagError
+	TagError // non-indexable
```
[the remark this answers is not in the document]
na--: I tried this, but it actually makes it more difficult to read :disappointed: You end up with 4 groups (enabled/disabled * indexable/non-indexable). The current way of splitting by enabled/disabled and commenting the non-indexable is more readable.

lib/netext/httpext/request.go
na--: Tried this, but I'd prefer to leave things as they are. In most places, this helper method will just lead to more allocations :disappointed: Here is, but also in these other places: https://github.com/grafana/k6/blob/d3d0eb8af4b14a169143126a0f839f1f6833b153/output/csv/output.go#L238-L247
https://github.com/grafana/k6/blob/d3d0eb8af4b14a169143126a0f839f1f6833b153/output/json/wrapper.go#L46
https://github.com/grafana/k6/blob/d3d0eb8af4b14a169143126a0f839f1f6833b153/js/modules/k6/execution/execution.go#L379-L386

As far as I can see, there is no single place where such a method won't cause unnecessary extra allocations. And since it can be so easily misused, I'd prefer to not add it

metrics/system_tag.go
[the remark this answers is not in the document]
na--: Connected to my points in https://github.com/grafana/k6/pull/2654#discussion_r971639135, I kind of think it should stay included by default, but I am open to change my mind. But let's please use the other thread, to keep things in a single place.

js/modules/k6/execution/execution.go
na--: Any extra comments here or can this be resolved?

lib/netext/httpext/transport.go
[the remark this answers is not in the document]
mstoykov: I am currently for dropping it from this PR and add it later. 

I am still of the opinion that this likely will not be used by 99% of the users. 

And even less so for the people who want to name their redirects. 

WDYT? dropp `raw_url` open PR for later?

metrics/system_tag.go
na--: Resolved by [this force-pushed fixup](https://github.com/grafana/k6/compare/d3d0eb8af4b14a169143126a0f839f1f6833b153..c8bfdd7f722411ed6846fe41d626ab07f618712f) of the original commit.

metrics/sample.go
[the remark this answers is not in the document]
na--: Resolved by [this force-pushed fixup](https://github.com/grafana/k6/compare/d3d0eb8af4b14a169143126a0f839f1f6833b153..c8bfdd7f722411ed6846fe41d626ab07f618712f) of the original commit.

js/modules/k6/execution/execution.go
na--: Resolved by [this force-pushed fixup](https://github.com/grafana/k6/compare/c8bfdd7f722411ed6846fe41d626ab07f618712f..7c90538db8b799df11022500cb81d91c4e223dc1) to the second commit. 

I don't particularly like how now I have to check if the `state` is `nil` everywhere, but overall I agree it's better to have a single prototype (and even `tags` object) per VU :bow: Besides, I think there aren't any actual technical obstacles for the `exec.vu.tags` (and other `exec.vu` properties) to work even in the init context, with some minor refactoring in the future... :thinking: I think we don't currently do it, because the `Tags`, `VUID` and `VUIDGlobal` properties of the VU are being kept in the `lib.State` struct, but that doesn't really need to be the case :thinking: These can be made available even in the init context, I think, so we may choose to relax this constraint in the future.

lib/netext/httpext/transport.go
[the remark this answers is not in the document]
na--: to get back to your original points 2. and 3. - I tried to fix them, but the resulting code is a bit nasty, and doesn't really optimize much. Right now, this allocation:
```go
cleanURL := URL{u: unfReq.request.URL, URL: unfReq.request.URL.String()}.Clean()
```

just allocates the `URL` object directly on the stack, and that URL object is very tiny: https://github.com/grafana/k6/blob/7c90538db8b799df11022500cb81d91c4e223dc1/lib/netext/httpext/url.go#L9-L14

And, if the actual URL doesn't have credentials in it, the `URL.Clean()` method doesn't allocate anything new, it just returns one of the existing object properties: https://github.com/grafana/k6/blob/7c90538db8b799df11022500cb81d91c4e223dc1/lib/netext/httpext/url.go#L34-L40 

So, we don't gain anything significant and the cost is a more difficult to follow code, so I'd prefer to leave points 2. and 3. as they are too.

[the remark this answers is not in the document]
mstoykov: > if we drop raw_url, then the breaking change we are doing to url will be much bigger IMO.

welll .. I guess we disagree on this as well as on other things.

Irregardless of whether `raw_url` is present or not, the `url` breaking change is a breaking change. To fix it users will need to:
1. care about this - which is the other part that I think you are *extremely* overestimating how many users (if any) care. We have never had this as funcitionality in the cloud. I don't remember this being a problem. Not to speak that any system that needs to intake this data will need to be able to ingest it - which is not really possible with any of the current outputs for most RPSs.
2. They will still need to actually change to using `raw_url` instead of `url`. Which might be a lot more involved.
3. In this PR all but 2 outputs - csv and json, does *not* support outputing the metadata. So it still "much bigger" breaking change for them.

Given this, for me `raw_url` is fixing the problem half-arsedly for some percentage of all users and for users that IMO just don't exist. 

If users want to know their original `url` they likely won't be using `name` in the first place ;). 

Also with the current imlpementation of `1` it means that some requests will have to be queried (for simplicity in grphana) for `url` and some for `raw_url` to get their url. 

But again IMO this helps noone and not in a way that will even matter without even *more* changes.

So I propose we drop it and discuss it after the release if we even should provide it for users. Or wait for users to come and ask about it.

output/cloud/output.go
```diff
@@ -145,13 +145,13 @@ func newOutput(params output.Params) (*Output, error) {
 // validateRequiredSystemTags checks if all required tags are present.
 func validateRequiredSystemTags(scriptTags *metrics.SystemTagSet) error {
 	missingRequiredTags := []string{}
-	requiredTags := metrics.TagName |
+	requiredTags := metrics.SystemTagSet(metrics.TagURL | metrics.TagName |
```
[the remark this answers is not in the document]
na--: hmm :thinking: I don't remember a reason, so this is most likely a mistake, will investigate and remove it.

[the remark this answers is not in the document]
na--: resolved by the [latest force-push](https://github.com/grafana/k6/compare/7c90538db8b799df11022500cb81d91c4e223dc1..6ba6ff20da1b1bd48f5f680b610795852cf5ce02), along with another issue I forgot to fix in the PR (since it was discussed outside of it :disappointed: )

output/cloud/output.go
```diff
@@ -145,13 +140,13 @@ func newOutput(params output.Params) (*Output, error) {
 // validateRequiredSystemTags checks if all required tags are present.
 func validateRequiredSystemTags(scriptTags *metrics.SystemTagSet) error {
 	missingRequiredTags := []string{}
-	requiredTags := metrics.TagName |
+	requiredTags := metrics.SystemTagSet(metrics.TagName |
 		metrics.TagMethod |
 		metrics.TagStatus |
 		metrics.TagError |
```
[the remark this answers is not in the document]
na--: Should be resolved by https://github.com/grafana/k6/pull/2654/commits/cb191cc8994036bb1b953bd20417585d4492bb10, unless someone has objections to making the `error` tag indexed again?

lib/netext/httpext/transport.go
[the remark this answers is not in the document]
na--: Yeah, I guess we disagree :slightly_smiling_face: In my mind, it is very simple, and `raw_url` satisfies a very important use case with very little cost. 

With the current k6 version, if you have some outlier request that times out or just takes too long, you can find its actual URL, even if you've specified a `name` tag. Yes, you'd need to also export your granular non-aggregated results to either a .json, or a .csv output, or stream them to InfluxDB, but you can do it. 

This is the capability I am trying to preserve with `raw_url`. And, if we don't have `raw_url`, users will literally have no way to diagnose such an issue, unless they write a wrapper around `k6/http` that manually looks at `Response.timings` and logs something... :confused: 

I haven't added support for non-indexed tags as fields in the `influxdb` output just because that output is basically on life-support... But if that will help you accept this PR, I can easily add that too.

Having `raw_url` will also be helpful in other things. For example, once we have tracing or some output with exemplars and/or sampling of metric outliers, we should probably include the `raw_url` (and other non-indexed tags) there too, since they may help tremendously with the debugging and we won't have concerns about their high cardinality.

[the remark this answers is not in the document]
mstoykov: So your argument is that we need this because if someone: 
1. uses `name` tag
2. has a particular request with a particular url that is hidden by `name` but is very problematic - let's say takes 10x longer or returns some error. 
3. has an output that supports this *and* the capacity to actually ingest the data. 

We need to have a separate non indexable tags that will be on by default for everybody? 

IMO in this particular case the user can just use the workaround @codebien provided. They still need to do extra work to figure this out - so why should we make everybody have `raw_url`.


I am still fo the opinion that my original `1` is very strange UX decision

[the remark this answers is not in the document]
na--: > so why should we make everybody have raw_url.

we don't make anyone have anything :smile: I want us to provide the `raw_url` value by default, which might be of use to a subset of users that are diagnosing an issue with their systems :wink: and I'd argue it has a very negligible cost

> I am still fo the opinion that my original 1 is very strange UX decision

should we always send the `error` and `error_code` tags then, even on successful requests? :thinking: or the TLS version for non-HTTP requests? :sweat_smile: there are a lot of precedents of k6 not sending all possible tags for every metric, and I don't see us changing that

js/modules/k6/execution/execution.go
mstoykov: I meant just the prototype ... but  I also was under the wrong impression that the `tags` were created once ... so I guess this is an improvement as well, so I will resolve it.

lib/netext/httpext/transport.go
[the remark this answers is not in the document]
mstoykov: > I'd argue it has a very negligible cos

We will always allocate a map and a string even if most users never use them (or have a way to use them). And this happens specifically in the case where the majority of the users of `name` will be doing it in order to get *better* performance, and because the concrete url is not important ;). 

Again IMO the users you are some very small minority and there is a workaround for their use case. 

We can also always add this *later* if I am wrong and there is a big majority of users needing it. But as time has shown we rarely if ever remove stuff.

> should we always send the error and error_code tags then, even on successful requests? thinking or the TLS version for non-HTTP requests? sweat_smile there are a lot of precedents of k6 not sending all possible tags for every metric, and I don't see us changing that

No - an empty `error` when you are querying for an `error` will tell you that there is no error. The same as having an *empty* error. Same for `error_code` or any of the other suggestions.

Querying for `raw_url` and getting empty doesn't mean that hte `url` is empty - it means the `url` is the same as the `name`.  This will now need to be documented and if somebody doesn't read it and just sees "ooh tehre is an raw_url in my data" and uses it - can very easily later miss all the requests that do *not* have `raw_url`. 

Arguably all of the above mean that this needs to be discussed. And I would prefer to do it outside of this PR instead of blocking the PR on this particular issue. So this is why I propose to just drop it and have the discussion outside of the PR. Might even happen within this release :shrug:

metrics/tags_test.go
```diff
@@ -109,6 +109,8 @@ func TestTagSetContains(t *testing.T) {
 	assert.False(t, st.Contains(outer))
 }
 
+// TODO: Add TagsAndMeta tests
+
```
codebien, line 113: This sounds relevant to be addressed, spot as a reminder

js/modules/k6/execution/execution.go
```diff
@@ -297,72 +302,71 @@ func optionsAsObject(rt *goja.Runtime, options lib.Options) (*goja.Object, error
...
-		return nil
+	tcv := o.state.Tags.GetCurrentValues()
+	if tag, ok := tcv.Tags.Get(key); ok {
+		return o.runtime.ToValue(tag)
 	}
-	return o.Runtime.ToValue(tag)
+	if metadatum, ok := tcv.Metadata[key]; ok {
+		return o.runtime.ToValue(metadatum)
+	}
+	return nil
```
codebien, line 318: This have to be removed, spot as a reminder

na--:
For anyone watching, after some internal discussions, we are likely not going to merge this PR as it is. 

We are cannibalizing it for parts and trying to merge the parts nobody objects to (most of the Go internals and refactoring, and the `Metadata` map in `metrics.Sample`) in k6 v0.41.0, while leaving the rest (any JS API changes that are not purely refactoring) for further consideration in the future. See https://github.com/grafana/k6/issues/2584#issuecomment-1269775631 for more details.

codebien (commented):
I rebased the code with the new changes from `master`. There is a [backup branch](https://github.com/grafana/k6/tree/backup-non-indexable-tags) in case you would check the original version.

I also removed the previously added `metrics.TagRawURL` after an internal discussion. We will open a dedicated issue for it.
There is also a small change in [the latest fixup](https://github.com/grafana/k6/pull/2654/commits/cf7dd6a6ee971f038a2a063732ddf5dcf611b7c6) from the original version.

When the code will be ready to be merged we will have to execute again [the update of the experimental module](https://github.com/grafana/k6/blob/master/js/modules/k6/experimental/README.md#upgrading).

na--:
> I rebased the code with the new changes from master. There is a [backup branch](https://github.com/grafana/k6/tree/backup-non-indexable-tags) in case you would check the original version.

Hmm :thinking:  it would have probably been simpler and safer to keep the old branch in this PR and just close it (without merging) and open a new PR with the new version. Because, when you have a PR with a branch, you can safely delete and then restore the branch without any fears. But if the free-standing [`backup-non-indexable-tags`](https://github.com/grafana/k6/tree/backup-non-indexable-tags) branch is deleted, it won't be recoverable, so it will just clutter the [branches list](https://github.com/grafana/k6/branches).

> I also removed the previously added `metrics.TagRawURL` after an internal discussion. We will open a dedicated issue for it.
> There is also a small change in [the latest fixup](https://github.com/grafana/k6/pull/2654/commits/cf7dd6a6ee971f038a2a063732ddf5dcf611b7c6) from the original version.

The reason the original code was like that was to optimize it slightly. In your code, `strconv.Itoa()` and `strconv.FormatBool()` will always be called, even when they are not needed (e.g. when the tags are disabled). It's a very minor optimization, since these tags are enabled by default, so the readability improvements are probably more important.

js/modules/k6/execution/execution_test.go
```diff
@@ -101,6 +101,8 @@ func TestVUTagsSetSuccessAccetedTypes(t *testing.T) {
 	}
 }
 
+// TODO: add tests for vu.tags helper method that returns whether the tag is indexed or not
```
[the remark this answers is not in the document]
na--: I think you rebased something incorrectly. This TODO was removed in the original branch and actually replaced with the tests for `exec.vu.tags.isIndexed()`, see: https://github.com/grafana/k6/commit/4ca7f0460d2ebce852856e2f2390b7195bcdb67d#diff-72a43f529f0963da646f852d3894f7f76b29d7ae1ca96fa6d7025b6ec542e794R106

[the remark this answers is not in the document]
na--: And since we are dropping the JS APIs, for now, it's not needed in the new branch.

output/json/wrapper.go
```diff
@@ -1,34 +1,78 @@
...
-		Time  time.Time       `json:"time"`
-		Value float64         `json:"value"`
-		Tags  *metrics.TagSet `json:"tags"`
-	} `json:"data"`
 	Metric string `json:"metric"`
+	Type   string `json:"type"`
+	Data   struct {
+		Time  time.Time           `json:"time"`
+		Value float64             `json:"value"`
+		Tags  tagsAndMetaEnvelope `json:"tags"`
```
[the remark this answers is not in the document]
na--: We can't add them all as tags, since you can now have a tag and a metadata value with the same key. Previously that was impossible, now it can be done. This was one of my (albeit, more minor) arguments why the original proposal for non-indexable tags was better.

Now we don't have a choice, tags and metadata need to be separate in all outputs, unless somehow there is no problem to have multiple `key=value` elements with the same key

js/modules/k6/execution/execution.go
```diff
@@ -314,13 +317,8 @@ func (o *tagsDynamicObject) Get(key string) goja.Value {
...
 	var err error
-	o.state.Tags.Modify(func(tags *metrics.TagSet) *metrics.TagSet {
-		newTags, applyErr := common.ApplyCustomUserTag(o.runtime, tags, key, val)
-		if applyErr != nil {
-			err = applyErr
-			return tags
-		}
-		return newTags
+	o.state.Tags.Modify(func(tagsAndMeta *metrics.TagsAndMeta) {
+		err = common.ApplyCustomUserTag(o.runtime, tagsAndMeta, key, val)
```
olegbespalov, line 321: just a nitpick, if we panic here, we can keep the scopes cleaner:

like:
```
o.state.Tags.Modify(func(tagsAndMeta *metrics.TagsAndMeta) {
		if err := common.ApplyCustomUserTag(o.runtime, tagsAndMeta, key, val); err != nil {
            panic(o.runtime.NewTypeError(err.Error()))
        }
}

return true
```

lib/netext/httpext/transport.go
```diff
@@ -80,75 +80,51 @@ func (t *transport) measureAndEmitMetrics(unfReq *unfinishedRequest) *finishedRe
 		trail:             trail,
 	}
 
-	tags := t.tags
+	ctm := t.tagsAndMeta.Clone()
```
olegbespalov, line 83: I believe the context of the usage of this variable is big enough to give it a full name like `clonedTagsAndMetadata` (or just a `clonedTags`) or even better to call it `sampleTags` (which I guess it is) since it can be essential to know downstream.

metrics/tags.go
```diff
@@ -156,6 +156,82 @@ var _ interface {
...
+func (tm *TagsAndMeta) Delete(key string) {
+	tm.Tags = tm.Tags.Without(key)
+	if tm.Metadata != nil {
+		delete(tm.Metadata, key)
+	}
+}
+
+// Clone makes a complete copy of the current TagsAndMeta object and returns it.
+func (tm TagsAndMeta) Clone() TagsAndMeta {
+	res := TagsAndMeta{Tags: tm.Tags}
```
olegbespalov, line 224: Just double-checking since Atlas struct has pointers; it's a safe way to cone it?

codebien:
@olegbespalov can you please comment on the new PR?

olegbespalov:
@codebien, just for my understanding, will this PR be closed? Shouldn't I check them anymore? :thinking:

codebien:
Yes, we are going to rollback this PR  to the previous history and close then move in two more defined PRs.

lib/netext/httpext/transport.go
olegbespalov: Moved into #2726

na--:
Closing this for the reasons stated in https://github.com/grafana/k6/pull/2654#issuecomment-1269779558 and https://github.com/grafana/k6/issues/2584#issuecomment-1269775631

In short, we merged the internal parts of this PR in other PRs (https://github.com/grafana/k6/pull/2716, https://github.com/grafana/k6/pull/2726, https://github.com/grafana/k6/pull/2727) and we will continue iterating on a better solution for the JS APIs that expose the high-cardinality `metrics.Sample.Metadata` for user scripts.

PR #2660, opened 2022-08-22
feat(local): load system cert pool. Resolves #2364

### LINKED ISSUE #2364: Extra TLS Latency from Loading Root Certificates
### Brief summary

The first time root certificates are loaded in OSX it takes a long time to parse them (> 100ms) and this time will appear as part of `http_req_tls_handshaking`. The certificates can be forced to load by running: `x509.SystemCertPool()` somewhere in the code before benchmarking.

### k6 version

k6 v0.36.0 ((devel), go1.17.6, darwin/amd64)

### OS

macOS 10.15.7

_No response_

```
import http from 'k6/http';

export default function () {

  const response = http.get('https://www.google.com');

}
```

```
k6 run -i 1 test.js
```

```
     http_req_tls_handshaking.......: avg=138.26ms min=138.26ms med=138.26ms max=138.26ms p(90)=138.26ms p(95)=138.26ms
```

this should take around <30ms on my machine.

```
k6 run -i 2 --no-connection-reuse test.js
```

```
     http_req_tls_handshaking.......: avg=80.18ms  min=12.71ms med=80.18ms  max=147.65ms p(90)=134.15ms p(95)=140.9ms
```

we can see when running two requests the first one took 147ms and second one took 12.71ms.

### Expected behaviour

The first HTTPS request shouldn't stall because of client side certificate parsing of the root certificates.

### Actual behaviour

The first HTTPS request stalls because of client side certificate parsing of the root certificates.

tbourrely:
Hey 🙂 

This is what I understood of the #2364 issue.
I'd like some input on how to test this programmatically. I don't see much change when running k6 with or without the line.

Regards,

### DIFF
--- a/cmd/test_load.go
+++ b/cmd/test_load.go
@@ -3,12 +3,14 @@ package cmd
 import (
 	"archive/tar"
 	"bytes"
+	"crypto/x509"
 	"fmt"
 	"io"
 	"os"
 	"path/filepath"
 	"sync"
 
+	"github.com/sirupsen/logrus"
 	"github.com/spf13/afero"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
@@ -240,6 +242,13 @@ func loadAndConfigureTest(
 	return test.consolidateDeriveAndValidateConfig(gs, cmd, cliConfigGetter)
 }
 
+// loadSystemCertPool attempts to load system certificates.
+func loadSystemCertPool(logger *logrus.Logger) {
+	if _, err := x509.SystemCertPool(); err != nil {
+		logger.WithError(err).Warning("Unable to load system cert pool")
+	}
+}
+
 func (lct *loadedAndConfiguredTest) buildTestRunState(
 	configToReinject lib.Options,
 ) (*lib.TestRunState, error) {
@@ -247,6 +256,10 @@ func (lct *loadedAndConfiguredTest) buildTestRunState(
 	if err := lct.initRunner.SetOptions(configToReinject); err != nil {
 		return nil, err
 	}
+
+	// it pre-loads system certificates to avoid doing it on the first TLS request.
+	// This is done async to avoid blocking the rest of the loading process as it will not stop if it fails.
+	go loadSystemCertPool(lct.preInitState.Logger)
 
 	return &lib.TestRunState{
 		TestPreInitState: lct.preInitState,
--- a/cmd/tls_integration_test.go
+++ b/cmd/tls_integration_test.go
@@ -1,0 +1,74 @@
+package cmd
+
+// TODO: convert this into the integration tests, once https://github.com/grafana/k6/issues/2459 will be done
+
+import (
+	"bytes"
+	"encoding/json"
+	"testing"
+
+	"github.com/spf13/afero"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"go.k6.io/k6/lib/testutils/httpmultibin"
+)
+
+type metric struct {
+	Avg float64 `json:"avg"`
+	Min float64 `json:"min"`
+	P95 float64 `json:"p(95)"`
+}
+
+type summaryMetrics struct {
+	HTTPReqTLSHandshaking metric `json:"http_req_tls_handshaking"`
+}
+
+type summaryMetricsWrapper struct {
+	Metrics summaryMetrics `json:"metrics"`
+}
+
+func TestTLSLoadSystemCertificates(t *testing.T) {
+	t.Parallel()
+
+	summaryFilename := "summary.json"
+
+	tb := httpmultibin.NewHTTPMultiBin(t)
+	testState := newGlobalTestState(t)
+	testState.args = []string{"k6", "run", "--no-connection-reuse", "--summary-export", summaryFilename, "-i", "2", "-"}
+	testState.stdIn = bytes.NewReader([]byte(tb.Replacer.Replace(`
+		import http from "k6/http"
+		export const options = {
+		hosts: {
+			"HTTPSBIN_DOMAIN": "HTTPSBIN_IP",
+		},
+		insecureSkipTLSVerify: true,
+		}
+
+		export default () => {
+		http.get("HTTPSBIN_URL/get");
+		}
+  `)))
+	newRootCommand(testState.globalState).execute()
+
+	content, err := afero.ReadFile(testState.fs, summaryFilename)
+	require.NoError(t, err)
+
+	var summaryMetricsWrapperContent summaryMetricsWrapper
+	err = json.Unmarshal(content, &summaryMetricsWrapperContent)
+	require.NoError(t, err)
+
+	delta := 10.0
+	assert.InDelta(
+		t,
+		summaryMetricsWrapperContent.Metrics.HTTPReqTLSHandshaking.Min,
+		summaryMetricsWrapperContent.Metrics.HTTPReqTLSHandshaking.Avg,
+		delta,
+	)
+
+	assert.InDelta(
+		t,
+		summaryMetricsWrapperContent.Metrics.HTTPReqTLSHandshaking.Min,
+		summaryMetricsWrapperContent.Metrics.HTTPReqTLSHandshaking.P95,
+		delta,
+	)
+}


olegbespalov:
Hey @tbourrely !

Thanks for the contribution, I've just wanted to say that currently, the team is busy with a new release, but we'll get back to you straight after that!

Cheers!

codebien:
Hi @tbourrely,
thanks for your contribution.

I think a better place for this code could be closer to the TLS configuration initialisation:
https://github.com/grafana/k6/blob/e09bb87277865d668586429eee97158fbbfc58e5/js/runner.go#L169-L177

You can test checking the p95 and avg metrics are equal (with some tolerance) to the min metric. An example for testing the Runner code with TLS can be found here:
https://github.com/grafana/k6/blob/e09bb87277865d668586429eee97158fbbfc58e5/js/runner_test.go#L958

@olegbespalov and @mstoykov WDYT? @tbourrely wait for more opinions from the other reviewers before submitting changes.

mstoykov:
It likely is better near to the tls configurations, but it also might be better to be inside of [sync.Once](https://pkg.go.dev/sync#Once) as in practice we are trying to do this *once* so that we preload it before the actual test starts :thinking:. 

It might even be better to just do it in a goroutine earlier, maybe while making a new Runner :thinking: this way it can go in the background while we continue with the other operations, hopefully it will be done before we need it and if not the program will still work :shrug: 

This likely needs to be looked into further, but all of those sound as an improvement. 

But I definitely would want this to be tested as I do remember that this function had some problems on specific OSes

mstoykov:
Actually yes if you look into the test failure you will see how it fails with
> crypto/x509: system root pool is not available on Windows

I don't know though why it only happens for the 1.17.x :shrug:

codebien:
Now you remember me that it has a strong dependency from the OS I would prefer to have it in the cmd package, where the `buildTestRunState` method seems to me a good candidate.
https://github.com/grafana/k6/blob/93b3a309944aec2cc8284e115334c015eb76e708/cmd/test_load.go#L243-L258

In this way, it makes more sense to have an integration test because it has the tight dependency on the OS.

@mstoykov WDYT?

mstoykov:
I am fine with it being in `cmd` - not so certain where exactly.

But if we pull it earlier and earlier I am definitely for it to be in a separate goroutine, and preferably to *not* abort the test even if it fails, but instead report a warning :shrug:. Although I guess after the windows fix it won't be returning an error :shrug: 

WDYT @na-- @olegbespalov

codebien (changes requested):
Hey @tbourrely,
it seems we have a consensus on the request changes. Do they sound clear to you or do you need more input?

tbourrely:
Hey 😊

Thanks for your inputs. I will have a look at it as soon as I can and comeback to you if some things are still unclear.

tbourrely:
Hey @codebien @mstoykov @olegbespalov can you confirm that the changes corresponds to what you had in mind ?
if so, I will move forward with the integration test 🙂

cmd/test_load.go
```diff
@@ -240,14 +241,23 @@ func loadAndConfigureTest(
...
 func (lct *loadedAndConfiguredTest) buildTestRunState(
+	gs *globalState,
 	configToReinject lib.Options,
 ) (*lib.TestRunState, error) {
 	// This might be the full derived or just the consodlidated options
 	if err := lct.initRunner.SetOptions(configToReinject); err != nil {
 		return nil, err
 	}
 
+	go loadSystemCertPool(gs)
```
codebien, line 259:
```suggestion
    // it pre-loads the system certificates so it can be avoided doing it on the first TLS request.
    // It loads in async so we can avoid blocking the rest of the loading process,
    // because the test will not be stopped if it fails.
	go loadSystemCertPool(gs.logger)
```
I guess something like this explaining why could be helpful. Feel free to add more context if it isn't clear enough.

cmd/test_load.go
```diff
@@ -240,14 +241,23 @@ func loadAndConfigureTest(
 	return test.consolidateDeriveAndValidateConfig(gs, cmd, cliConfigGetter)
 }
 
+func loadSystemCertPool(gs *globalState) {
```
codebien, line 244:
```suggestion
func loadSystemCertPool(gs *logrus.FieldLogger) {
```
We can keep the dependency smaller, it isn't required to pass the entire state.

cmd/test_load.go
```diff
@@ -240,14 +241,23 @@ func loadAndConfigureTest(
...
+	if _, err := x509.SystemCertPool(); err != nil {
+		gs.logger.WithError(err).Warning("Unable to load system cert pool")
+	}
+}
+
 func (lct *loadedAndConfiguredTest) buildTestRunState(
+	gs *globalState,
```
codebien, line 251:
```suggestion
```
`loadadedAndConfiguredTest` contains the `TestPreInitState` where the logger is available, so If we use just the logger then we can avoid adding this dependency.

codebien (changes requested):
>can you confirm that the changes corresponds to what you had in mind ?

@tbourrely yes, it sounds fine to me. I added some comments.

codebien:
Hi @tbourrely,
this seems ready to be reviewed again. Can you rebase the branch and resolve the conflict, please? We would like to include this at the beginning of the next cycle.

tbourrely:
Hey 👋 

I was not super convinced of my test but I'll gladly rebase and put it to review. 

Sorry for the delay !

tbourrely:
Rebased & squashed 👍

cmd/test_load.go
```diff
@@ -248,6 +257,12 @@ func (lct *loadedAndConfiguredTest) buildTestRunState(
 		return nil, err
 	}
 
+	// it pre-loads system certificates to avoid doing it on the first TLS request.
+	// This is done async to avoid blocking the rest of the loading process as it will not stop if it fails.
+	go loadSystemCertPool(lct.preInitState.Logger)
+
+	// TODO: init atlas root node, etc.
+
```
codebien, line 265:
```suggestion
```
It seems we don't need these lines

cmd/tls_integration_test.go
```diff
@@ -0,0 +1,67 @@
...
+		export default function() {
+			let response = http.get("https://test-api.k6.io");
+		};
+	`
+
+	testFilename := "script.js"
+	summaryFilename := "summary.json"
+	testState := newGlobalTestState(t)
+	require.NoError(t, afero.WriteFile(testState.fs, filepath.Join(testState.cwd, testFilename), []byte(script), 0o644))
+	testState.args = []string{"k6", "run", testFilename, "--summary-export", summaryFilename, "-i", "2"}
```
codebien, line 47:
```suggestion
	testState.args = []string{"k6", "run", "--no-connection-reuse", "--summary-export", summaryFilename, "-i", "2", testFilename}
```
You're missing the option to re-init the connection, this is the reason why you get one request with a 0 value. If you change it, you will see the min value increase as expected.

codebien (changes requested):

cmd/tls_integration_test.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+func TestTLSLoadSystemCertificates(t *testing.T) {
+	t.Parallel()
+
+	script := `
+		import http from "k6/http";
+
+		export default function() {
+			let response = http.get("https://test-api.k6.io");
```
olegbespalov, line 39: I'm wondering if it's possible to avoid the network/internet call here :thinking:

tbourrely: Do you mean creating an https server locally to run the test against ?

olegbespalov: Yeah, for most of our tests, we use a local HTTP Server. It prevents us from the fails when the network fails. 

But I guess it's not the case with this test, so nevermind :relaxed:

codebien: We can do something inspired by what we have here https://github.com/grafana/k6/blob/8a74171c8b43f48fb53078e0bcfa7851828124bc/cmd/integration_test.go#L268-L284

@tbourrely feel free to resolve it or we can do it internally after we merged this PR. I unresolve it so we can remember to do it after we have merged if you haven't done it before.

cmd/tls_integration_test.go
```diff
@@ -0,0 +1,71 @@
+//go:build !windows
+// +build !windows
```
codebien, line 2:
```suggestion
```
Did you find issues on Windows?

cmd/tls_integration_test.go
```diff
@@ -0,0 +1,71 @@
...
+
+package cmd
+
+// TODO: convert this into the integration tests, once https://github.com/grafana/k6/issues/2459 will be done
+
```
codebien, line 7:
```suggestion
```
I think we can remove this part, it is already an integration test. I think we can move it directly as a file of the `cmd/integration_tests` package.

codebien (changes requested):
LGTM, just two nitpicks. I would like to move the file into the `cmd/integration_tests` package.

cmd/tls_integration_test.go
codebien: This is not true, sorry, my mistake.

cmd/tls_integration_test.go
tbourrely: > Actually yes if you look into the test failure you will see how it fails with
> 
> > crypto/x509: system root pool is not available on Windows
> 
> I don't know though why it only happens for the 1.17.x 🤷
> 
> edit: seems to have been fixed with [golang/go@3544082](https://github.com/golang/go/commit/3544082f75fd3d2df7af237ed9aef3ddd499ab9c) ... so this might be blocked on us dropping 1.17.x support - very likely for the next release when we move to 1.19.x as the primary version we release with

I thought some windows tests were running against 1.17.x golang version, my bad !

cmd/tls_integration_test.go
tbourrely: done 😉

codebien (approved):
Using HTTPMultiBin loads the TLS server so it invokes before the `SystemCertPool` func, this is the reason why the test is not failing without the patch. I will address this issue with a new PR, I think this version is fine and we can merge it.

olegbespalov (approved):
Thanks, @tbourrely! :bow:

PR #2664, opened 2022-08-26
Golang update

manishprajapat180998:


arpit-razorpay (approved):
Approved.

na--:
This seems like it includes way more than a golang update.... And it seems like a PR from an old branch of k6, since we are already using Go 1.18 in `master`: https://github.com/grafana/k6/blob/e09bb87277865d668586429eee97158fbbfc58e5/Dockerfile#L1

So I'll close this as I closed your other duplicate PR. For what it's worth, we will not update to Go 1.19 until at least several weeks from now, after we release k6 v0.40.0 and after the first couple of Go 1.19.x bugfix releases come out.

PR #2679, opened 2022-09-12
Added max call (send and receive) size params to the gRPC client.

### LINKED ISSUE #2666: Support for configuring maximum call receive message size
### Feature Description

Currently there is no way to change maximum call receive message size for grpc client.
I use this snippet to create a client, but there is no param to configure max message size
```
const client = new grpc.Client();
client.load(['protobuf'], 'vormir.proto');
client.connect('hz-mihirn01.nferx.com:50051', {
  plaintext: true,
  timeout: '1m',
  // need to add max message size here
});
```

```
const client = new grpc.Client();
client.load(['protobuf'], 'vormir.proto');
client.connect('hz-mihirn01.nferx.com:50051', {
  plaintext: true,
  timeout: '1m',
  // add new param
  max_receive_size: 100000,
});
```

_No response_

ariasmn:
This should close #2666.

### DIFF
--- a/js/modules/k6/grpc/client.go
+++ b/js/modules/k6/grpc/client.go
@@ -108,6 +108,14 @@ func (c *Client) Connect(addr string, params map[string]interface{}) (bool, erro
 	ctx, cancel := context.WithTimeout(c.vu.Context(), p.Timeout)
 	defer cancel()
 
+	if p.MaxReceiveSize != 0 {
+		opts = append(opts, grpc.WithDefaultCallOptions(grpc.MaxCallRecvMsgSize(int(p.MaxReceiveSize))))
+	}
+
+	if p.MaxSendSize != 0 {
+		opts = append(opts, grpc.WithDefaultCallOptions(grpc.MaxCallSendMsgSize(int(p.MaxSendSize))))
+	}
+
 	c.addr = addr
 	c.conn, err = grpcext.Dial(ctx, addr, opts...)
 	if err != nil {
@@ -339,6 +347,8 @@ type connectParams struct {
 	IsPlaintext           bool
 	UseReflectionProtocol bool
 	Timeout               time.Duration
+	MaxReceiveSize        int64
+	MaxSendSize           int64
 }
 
 func (c *Client) parseConnectParams(raw map[string]interface{}) (connectParams, error) {
@@ -346,6 +356,8 @@ func (c *Client) parseConnectParams(raw map[string]interface{}) (connectParams, 
 		IsPlaintext:           false,
 		UseReflectionProtocol: false,
 		Timeout:               time.Minute,
+		MaxReceiveSize:        0,
+		MaxSendSize:           0,
 	}
 	for k, v := range raw {
 		switch k {
@@ -366,6 +378,18 @@ func (c *Client) parseConnectParams(raw map[string]interface{}) (connectParams, 
 			params.UseReflectionProtocol, ok = v.(bool)
 			if !ok {
 				return params, fmt.Errorf("invalid reflect value: '%#v', it needs to be boolean", v)
+			}
+		case "maxReceiveSize":
+			var ok bool
+			params.MaxReceiveSize, ok = v.(int64)
+			if !ok {
+				return params, fmt.Errorf("invalid maxReceiveSize value: '%#v', it needs to be an int", v)
+			}
+		case "maxSendSize":
+			var ok bool
+			params.MaxSendSize, ok = v.(int64)
+			if !ok {
+				return params, fmt.Errorf("invalid maxSendSize value: '%#v', it needs to be an int", v)
 			}
 
 		default:


js/modules/k6/grpc/client.go
```diff
@@ -339,13 +347,17 @@ type connectParams struct {
 	IsPlaintext           bool
 	UseReflectionProtocol bool
 	Timeout               time.Duration
+	MaxReceiveSize        int64
```
ariasmn, line 350: Had to use `int64` instead of `int` since it seems like the string is parsed as `int64` and didn't want to mess with it. If it's not ok please let me know.

js/modules/k6/grpc/client.go
```diff
@@ -108,6 +108,14 @@ func (c *Client) Connect(addr string, params map[string]interface{}) (bool, erro
 	ctx, cancel := context.WithTimeout(c.vu.Context(), p.Timeout)
 	defer cancel()
 
+	if p.MaxReceiveSize != 0 {
+		opts = append(opts, grpc.WithDefaultCallOptions(grpc.MaxCallRecvMsgSize(int(p.MaxReceiveSize))))
+	}
+
+	if p.MaxSendSize != 0 {
```
codebien, line 115:
```suggestion
	if p.MaxSendSize > 0 {
```
Why do we accept negative numbers here? Are they good for some reason?

codebien (changes requested):
Hi @ariasmn,
thanks for your contribution :bow: :clap:

I added just one minor comment, the missing part is the test coverage. Can you add a test for these cases, please?

js/modules/k6/grpc/client.go
ariasmn: Not that I know of, but didn't think too much about it since the gRPC client allow those negative values.
I can add a check for a positive int when parsing the params if you think it's the best.

ariasmn:
> Hi @ariasmn, thanks for your contribution bow clap
> 
> I added just one minor comment, the missing part is the test coverage. Can you add a test for these cases, please? However, the general idea is good +1

Sure, I will work on them and I will update the PR when I have them ready!
Thanks for the feedback.

ariasmn:
@codebien Just modified the PR with the requested changes and the addition of tests.

Don't know if I should be testing more than I did (throwing the correct error when using invalid values and that the connection is made successfully when using correct values)

Sorry if I'm missing something... Just let me know and I'll try to fix it (with some guidance in that case please :smile:)

codebien:
>throwing the correct error when using invalid values and that the connection is made successfully when using correct values

@ariasmn As you suggested, it is a required part otherwise we can't guarantee that the code is working as expected without manually testing all the time.

Unfortunately, the current gRPC test suite is more like an integration test so we need to assert the expected value doing a request to `httpmultibin` (as the other cases are doing). Otherwise, the optimal case would be to just check that we have set the expected option and just rely for the rest on the Go gRPC library test suite.

codebien:
@ariasmn I think in this case the real proof is testing if the limit is hit, the "normal" case where the limit is respected I expect is already covered from the other tests, because I expect the gRPC client has a default value set (to be verified).

ariasmn:
@codebien I see. 

> 



> @ariasmn I think in this case the real proof is testing if the limit is hit, the "normal" case where the limit is respected I expect is already covered from the other tests, because I expect the gRPC client has a default value set (to be verified).

The gRPC client does have a default value set, as you can see [here](https://github.com/grpc/grpc-go/blob/a077b9468383c645cffbb02b58df82cda8bdc629/clientconn.go#L97)

So, If I understood correctly, I should:

- Add the integration tests to check if the limit is hit, both in the send and receive cases.
- Remove the tests from my [last commit](https://github.com/grafana/k6/pull/2679/commits/3cd1dbff5781b4708e8204223e6fc657dbe6b564), because those are redundant.

Right?

In any case, I appreciate some guidance on how to do the mentioned tests. I have been looking at the current test cases and trying to work on them without any success sadly.

Sorry for the trouble and the time wasted, maybe the issue is getting a bit harder given my Golang skills :sweat:

ariasmn:
So I think I got it!

First, I removed the redundant tests as per your suggestion.
To test if the limit is hit, what I did was invoke the testing client with the new params set to 1 (min value) and let the test server send a response. Then, if a response code is the [RESOURCE_EXHAUSTED](https://grpc.github.io/grpc/core/md_doc_statuscodes.html) one, I throw the error and check if it's the one expected.

For me this totally makes sense, but maybe I'm missing something. Let me know in that case.

Again, thank you so much for your time!

codebien:
Hi @ariasmn, thanks for you effort, the solution is fine to me :clap: Apparently the code is not respecting the go format. You can run the linter locally as described in the [CONTRIBUTING](https://github.com/grafana/k6/blob/master/CONTRIBUTING.md#running-the-linter) doc.

ariasmn:
@codebien Ops, my bad, was tired and forgot to run `make lint`

Everything should be ok now!

js/modules/k6/grpc/client.go
```diff
@@ -367,6 +379,24 @@ func (c *Client) parseConnectParams(raw map[string]interface{}) (connectParams,
 			if !ok {
 				return params, fmt.Errorf("invalid reflect value: '%#v', it needs to be boolean", v)
 			}
+		case "maxReceiveSize":
+			var ok bool
+			params.MaxReceiveSize, ok = v.(int64)
+			if !ok {
+				return params, fmt.Errorf("invalid maxReceiveSize value: '%#v', it needs to be an int", v)
```
codebien, line 386:
```suggestion
				return params, fmt.Errorf("invalid maxReceiveSize value: '%#v', it needs to be an integer", v)
```

codebien (commented):
LGTM, just one minor comment. Let's wait for @olegbespalov's review.

js/modules/k6/grpc/client.go
ariasmn: Just changed every "int" for "integer" (actually feels better when reading the error).

If any more changes are needed after the next review, I will do them in a separate commit.

olegbespalov (approved):
Looks good to me! :+1: @ariasmn, thanks for the contribution!

codebien (approved):

ariasmn:
Thank you guys for the help and the amazing project!

Issue #1741, opened 2020-11-27
Display any extensions bundled by xk6 in the pre-execution output


simskij:
## Feature Description
With the introduction of xk6, it is likely that we'll start to get support questions or bug reports where users are running bespoke versions of k6. From a "k6 core" point of view, the course of action would/should likely be to ask the user to try again with an official build.

That aside, if we ever want to support any extension, or make life easier for extension developers, we need some way to tell by execution logs how to replicate their builds.

## Suggested Solution
Add a new row in the pre-execution output of k6 cli. This would only be visible when k6 detects something in the 'k6/x' namespace, and should then list extensions and their versions. Mockup below:

```
          /\      |‾‾| /‾‾/   /‾‾/   
     /\  /  \     |  |/  /   /  /    
    /  \/    \    |     (   /   ‾‾\  
   /          \   |  |\  \ |  (‾)  | 
  / __________ \  |__| \__\ \_____/ .io
  execution: local
     script: test.js
 extensions: github.com/dgzlopes/k6-extension-notifications
             github.com/dgzlopes/k6-extension-zmq
     output: -
```

With versions displayed:
```
          /\      |‾‾| /‾‾/   /‾‾/   
     /\  /  \     |  |/  /   /  /    
    /  \/    \    |     (   /   ‾‾\  
   /          \   |  |\  \ |  (‾)  | 
  / __________ \  |__| \__\ \_____/ .io
  execution: local
     script: test.js
 extensions: github.com/dgzlopes/k6-extension-notifications@v0.2.0
             github.com/dgzlopes/k6-extension-zmq@v0.0.0-20180725130230-947c36da315
     output: -
```

Given that we've been discussing other kinds of extensions than js extensions in the future, we might also need some way to detect those. One way of doing this would be to add a tripwire to `modules.Register` instead. Personally, I'd argue that we can leave that as a future improvement and focus on the js case for now.

na--:
We might also investigate a way to display the included extensions and their version in the output of `k6 version`, so we have a one-stop command we can ask users to perform, which will give us all the information we want (k6 version, Go version, OS, extensions).

imiric:
Hey, can we prioritize this for an upcoming cycle?

[...]

PR #2686, opened 2022-09-19
Added information about built-in modules for Version command

amelkikh:
Hi!
Please consider to review my PR.
Additional information about built-in modules seems to be helpful when compiled tool shared with teammates.

Have a good day!

### DIFF
--- a/cmd/version.go
+++ b/cmd/version.go
@@ -4,18 +4,38 @@ import (
 	"fmt"
 
 	"github.com/spf13/cobra"
-
+	"go.k6.io/k6/js/modules"
 	"go.k6.io/k6/lib/consts"
 )
 
 func getCmdVersion(globalState *globalState) *cobra.Command {
 	// versionCmd represents the version command.
-	return &cobra.Command{
+	var showModules bool
+
+	versionCmd := &cobra.Command{
 		Use:   "version",
 		Short: "Show application version",
 		Long:  `Show the application version and exit.`,
 		Run: func(_ *cobra.Command, _ []string) {
 			printToStdout(globalState, fmt.Sprintf("k6 v%s\n", consts.FullVersion()))
+
+			if showModules {
+				mods := modules.GetJSModules()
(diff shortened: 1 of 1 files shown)


codebien (changes requested):
Hi @amelkikh,
thanks for the effort in doing it :bow: :clap: 

I expect you need this to get information about the extensions included in a binary built with `xk6`. If that is the reason then we share the same feeling. We opened an issue in the past for it #1741.
Ideally, we would prefer to have a broader solution to support any extension (javascript module or output).

I try to summarise the idea for our vision.
When an extension is registered using the [output.RegisterExtension](https://pkg.go.dev/go.k6.io/k6/output#RegisterExtension) / [modules.Register](https://pkg.go.dev/go.k6.io/k6/js/modules#Register) methods then the internal logic should resolve the [PkgPath](https://pkg.go.dev/reflect#Type) using the Go [reflect](https://pkg.go.dev/reflect) package to get the module path of the given object and save it in a (shared) slice. In this way, the slice of resolved extensions can be used for getting the list when the `k6 version` command is invoked and it can be combined with the metadata returned from [debug.BuildInfo](https://pkg.go.dev/debug/buildinfo@go1.19.1#Read) for returning a more detailed list as previewed in the description of #1741.

Let me know if you need more information.

Note: If you are going to submit the requested changes then remember to sign the CLA, please. We will need it signed to eventually approve and merge the PR.

amelkikh:
Hi, @codebien!

Thanks for the detailed explanation. I share this approach with you. 
It will take me some time to figure out how to make a sign and dive into how your project works as I don't have much experience yet.

codebien:
Hi @amelkikh,
did you hit some sort of issues implementing the suggestion? Can I help in some way? Thanks.

codebien:
Hi @amelkikh,
thanks for your commitment and for letting us know. I hope we will see you as a contributor again in the future.

PR #2687, opened 2022-09-20
log: add more tokenizer tests

nsmith5:
Adds a couple more tests to the tokenizer in `log` package. Also formats the tests a bit to make it easy to add more if desired.

### DIFF
--- a/log/tokenizer_test.go
+++ b/log/tokenizer_test.go
@@ -62,9 +62,6 @@ func TestTokenizerInvalid(t *testing.T) {
 			t.Parallel()
 
 			_, err := tokenize(test.input)
-			if err == nil {
-				t.Fatal("expected an error but got none")
-			}
 			require.EqualError(t, err, test.errorMsg)
 		})
 	}
--- a/log/tokenizer_test.go
+++ b/log/tokenizer_test.go
@@ -4,44 +4,68 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func TestTokenizer(t *testing.T) {
-	tokens, err := tokenize("loki=something,s.e=2231,s=12,12=3,a=[1,2,3],b=[1],s=c")
-	assert.Equal(t, []token{
-		{
-			key:   "loki",
-			value: "something",
-		},
-		{
-			key:   "s.e",
-			value: "2231",
-		},
-		{
-			key:   "s",
-			value: "12",
-		},
-		{
-			key:   "12",
-			value: "3",
-		},
-		{
-			key:    "a",
-			value:  "1,2,3",
-			inside: '[',
-		},
-		{
-			key:    "b",
-			value:  "1",
-			inside: '[',
-		},
-		{
-			key:   "s",
-			value: "c",
-		},
-	}, tokens)
-	assert.NoError(t, err)
+	t.Parallel()
 
-	_, err = tokenize("empty=")
-	assert.EqualError(t, err, "key `empty=` with no value")
+	input := "loki=something,s.e=2231,s=12,12=3,a=[1,2,3],b=[1],s=c"
+	tokens, err := tokenize(input)
+	require.NoError(t, err)
+
+	expected := []token{
+		{key: "loki", value: "something"},
+		{key: "s.e", value: "2231"},
+		{key: "s", value: "12"},
+		{key: "12", value: "3"},
+		{key: "a", value: "1,2,3", inside: '['},
+		{key: "b", value: "1", inside: '['},
+		{key: "s", value: "c"},
+	}
+	assert.Equal(t, expected, tokens)
+}
+
+func TestTokenizerInvalid(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name     string
+		input    string
+		errorMsg string
+	}{
+		{
+			name:     "empty value",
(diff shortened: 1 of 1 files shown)


log/tokenizer_test.go
```diff
@@ -7,41 +7,63 @@ import (
...
+			wantTokens: []token{
+				{key: "loki", value: "something"},
+				{key: "s.e", value: "2231"},
+				{key: "s", value: "12"},
+				{key: "12", value: "3"},
+				{key: "a", value: "1,2,3", inside: '['},
+				{key: "b", value: "1", inside: '['},
+				{key: "s", value: "c"},
+			},
+			wantErrorMsg: "",
```
mstoykov, line 28:
```suggestion
```

nitpick: there is no need for this line - it will get this line by default ;)

mstoykov (approved):
Thanks for the PR :bow: 

It LGTM in general, but there are lint issues, that you will need to fix.

Also I left a small nitpick.

p.s. Is there a particular reason you are looking at this tests?

nsmith5:
👋🏻 Hey @mstoykov thanks for the review. No particular reason for these tests. I've been using and really enjoying the k6 project so just having a look through the code to understand how its working. Wanted to get a sense of what contributing was like so picked something simple like adding some unit tests :D

log/tokenizer_test.go
```diff
@@ -7,41 +7,64 @@ import (
...
+				if err != nil {
+					t.Fatal("didn't expect error")
+				}
+				assert.Equal(t, test.wantTokens, gotTokens)
+			} else {
+				if err == nil {
+					t.Fatal("expected an error but got none")
+				}
+				assert.EqualError(t, err, test.wantErrorMsg)
+			}
```
codebien, line 67: I think we could simplify this logic if we divide the test in two different functions TestTokenizer and TestTokenizerInvalid / Error.

In this way, we can simplify this `if-else` logic and remove the `name` field, because the name will be pretty obvious from the function name and from the expected error message. @mstoykov WDYT?

codebien (changes requested):
Now I see

log/tokenizer_test.go
nsmith5: Sure happy to split it up into two tests for clarity. I like having the name field because it sets the name of the sub-test with `t.Run(test.name, ...)`. When a particular test fails you can find the test case data quickly because of the test name is even more specific (e.g `TestTokenInvalid/empty_value_should_error` failed quickly gets you the test case and the particular case that failed).

nsmith5: Broken up into two tests. Happy to drop the names too if you want. Limme know :)

log/tokenizer_test.go
```diff
@@ -7,41 +7,78 @@ import (
...
+			t.Parallel()
+
+			gotTokens, err := tokenize(test.input)
+			if err != nil {
+				t.Fatal("didn't expect error")
+			}
+			assert.Equal(t, test.wantTokens, gotTokens)
+		})
+	}
+}
```
codebien, line 42: It doesn't seems we need a table test here so I would prefer a less-nested solution. 

```go
func TestTokenizer(t *testing.T) {
	t.Parallel()

	input := "loki=something,s.e=2231,s=12,12=3,a=[1,2,3],b=[1],s=c"
	tokens, err := tokenize(input)
	require.NoError(t, err)

	expected := []token{
		{key: "loki", value: "something"},
		{key: "s.e", value: "2231"},
		{key: "s", value: "12"},
		{key: "12", value: "3"},
		{key: "a", value: "1,2,3", inside: '['},
		{key: "b", value: "1", inside: '['},
		{key: "s", value: "c"},
	}
	assert.Equal(t, expected, tokens)
}
```

log/tokenizer_test.go
```diff
@@ -7,41 +7,78 @@ import (
...
+		t.Run(test.name, func(t *testing.T) {
+			t.Parallel()
 
-	_, err = tokenize("empty=")
-	assert.EqualError(t, err, "key `empty=` with no value")
+			_, err := tokenize(test.input)
+			if err == nil {
+				t.Fatal("expected an error but got none")
+			}
+			assert.EqualError(t, err, test.errorMsg)
```
codebien, line 81:
```suggestion
			require.EqualError(t, err, test.errorMsg)
```

codebien (changes requested):

log/tokenizer_test.go
codebien: The assert library will print the expected error m,essage so it should be easy to catch. In any case, having it it's fine so we can keep it :+1:

mstoykov (approved):
LGTM :+1:

log/tokenizer_test.go
```diff
@@ -4,44 +4,68 @@ import (
...
+		test := test
+		t.Run(test.name, func(t *testing.T) {
+			t.Parallel()
 
-	_, err = tokenize("empty=")
-	assert.EqualError(t, err, "key `empty=` with no value")
+			_, err := tokenize(test.input)
+			if err == nil {
+				t.Fatal("expected an error but got none")
+			}
```
codebien, line 67:
```suggestion
```
Having the `require` condition this is not required anymore

codebien (changes requested):

mstoykov (approved):

codebien (approved):

Issue #2688, opened 2022-09-21
The status API return json without set content-type


wingyplus:
### Brief summary

When requesting to status API (both GET and PATCH). The API returns JSON without set content-type. It cause some client that decode body base on content-type not working correctly. Here is an log from my HTTP client (I use Elixir tesla in this case):

```
20:50:16.239 [info] GET http://localhost:6565/v1/status -> 200 (4.971 ms)
 
20:50:16.239 [debug] 
>>> REQUEST >>>
(no query)
content-type: application/json

(no body)

<<< RESPONSE <<<
date: Wed, 21 Sep 2022 13:50:16 GMT
content-length: 149
content-type: text/plain; charset=utf-8

{"data":{"type":"status","id":"default","attributes":{"status":7,"paused":false,"vus":0,"vus-max":0,"stopped":false,"running":true,"tainted":false}}}
```

### k6 version

k6 v0.40.0 ((devel), go1.19.1, darwin/arm64)

### OS

macOS 12.5 (Monterey)

### Steps to reproduce the problem

* Starting by run `k6 run <script>`.
* Then request to `http://localhost:6565/v1/status`. both GET and PATCH

### Expected behaviour

The API should response with content-type `application/json`

### Actual behaviour

The API response with content-type `text/plain; charset=utf-8`

PR #2689, opened 2022-09-21
Fix /v1/status doesn't set json content-type

wingyplus:
Currently, the api doesn't set `Content-Type` header which cause some http client that decode data base on `Content-Type` not working correctly. Fixes by set Content-Type header to /v1/status api (both GET and PATCH).

Fixes: #2688


na-- (approved):
LGTM, thanks! :bow:

codebien (approved):
Thanks for your contribution :bow:

Issue #2676, opened 2022-09-08
ci: VERSION_DETAILS does not contain the latest tag


codebien:
During the build the new tag hasn't been picked from the [git describe command](https://github.com/grafana/k6/blob/dcbe2f9c4fef8dbfc5671e8b8de94a74146728af/build-release.sh#L13), in this way the `k6 version` prints an older tag. It isn't a critical bug because all the important information like the date or the commit hash are correct.

The issue could be related to the GitHub [checkout action](https://github.com/grafana/k6/blob/7f3131bce3d0b80636fc0eb0c89497037ba412e1/.github/workflows/all.yml#L163-L166) that doesn't fetch for a strange reason the latest created tag.

```
k6 version
k6 v0.40.0 (2022-09-08T09:06:02+0000/v0.39.0-92-gdcbe2f9c, go1.18.6, linux/amd64)
```

### Current

`v0.39.0-92-gdcbe2f9c`

### Expected

`v0.40.0-0-gdcbe2f9c`

na--:
Potentially relevant issues: https://github.com/actions/checkout/issues/882, https://github.com/actions/checkout/issues/701

na--:
Decided to dig this a little deeper and we can actually see all of the `git` actions that were executed by the v0.40.0 build :tada: And this seems like the reason for the problem to me: https://github.com/grafana/k6/runs/8245641202?check_suite_focus=true#step:2:620

In case that log disappears, I'll post the raw logs and a screenshot here as well: 
![logs](https://user-images.githubusercontent.com/3336773/189352253-e1ffb50f-6918-4d16-a791-6b65125fdc77.png)
[v0.40.0-build.log](https://github.com/grafana/k6/files/9535450/v0.40.0-build.log)

mstoykov:
Was this same command executed in the previous releases? Or did something change in the github-action, maybe outside the changes in our workflow?

olegbespalov:
I can guess that it's broken since, for the version details, we don't pass the `--tags` (https://git-scm.com/docs/git-describe#Documentation/git-describe.txt---tags) like we do for the Version.

na--:
:man_facepalming: Yeah, that is probably it...

na--:
@olegbespalov tried it in a private repo and the lack of `--tags` turned out to be the problem indeed.

PR #2691, opened 2022-09-22
Fix VERSION_DETAILS does not contain the latest tag

olegbespalov:
# What?

Add a `--tags` flag to fetch all tags for the `git describe`.

# Why?

Without that flag k6's version details gets a previous flag.

Closes: #2676

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,7 @@ FROM golang:1.18-alpine as builder
 WORKDIR $GOPATH/src/go.k6.io/k6
 ADD . .
 RUN apk --no-cache add git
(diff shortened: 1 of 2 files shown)


codebien (approved):

na-- (changes requested):
It seems like the same problem is also present in the Dockerfile: https://github.com/grafana/k6/blob/979378a08104f0f58a6aaf011a4c1ac764112437/Dockerfile#L5

na-- (approved):
:crossed_fingers:

codebien (approved):

mstoykov (approved):

PR #2693, opened 2022-09-26
Dont lock the event loop once a promise rejection is left unhandled

mstoykov:
Previously when a promise is rejected but *not* handled, the event loop would end the current iteration and then let a new one start.

Unfortuantely it also didn't clear the failure so the next time it will again *fail* the iteration as it still thinks a promise has been rejected without a handler.

Heavily related to https://github.com/grafana/xk6-browser/issues/510


imiric (approved):
Thanks for spotting this and the fix! :bow::tada:

I spent way too long troubleshooting the xk6-browser behavior, and it turned out to be a one-line k6 fix... :face_exhaling::facepalm:

oleiade (approved):
I might have run into this without realizing in the past 👀 
LGTM 👍🏻

Issue #2700, opened 2022-09-29
k6/net/grpc: load a protoset rather than individual proto files


jklipp:
### Feature Description

We use buf.build along with the buf schema registry to pull in some well-known protos (googleapis, envoyproxy/protoc-gen-validate) rather than vendor them in our own repo.  As such, it's a bit awkward to try to provide these protos to the grpc `Client.load` function as our proto dependencies may not be around locally on disk.  

### Suggested Solution (optional)

It's pretty common for grpc tooling to accept a protoset (a proto-serialized set of protos) rather than individual proto files and import paths, and would be nice of there was an additional `Client.loadProtoset` or similar functionality on the k6 grpc interface.

This looks pretty straightforwad, would k6 be interested in a PR?

bourquep:
@jklipp Have you been able to load protosets from the Buf schema registry with k6?

PR #2702, opened 2022-09-29
k6/net/grpc: initialize grpc from a protoset

jklipp:
This adds a LoadProtoset function to parse a protoset file to initialize
a grpc client.

In some cases it's easier to produce and consume a single protoset file
than multiple protos and import paths.

Closes https://github.com/grafana/k6/issues/2700.


jklipp:
Adjusted tests to use a single service with a dependent proto, such that the client shouldn't load unless it's able to see both of those protos in the protoset.  Multiple services was causing random ordering issues that led to not matching the expected order of MethodInfos in tests (possibly due to putting the FileDescriptorProtos into a map in NewFiles on FileOptions?).

mstoykov (approved):
LGTM! 

Thanks a lot for the excellent PR :tada:

mstoykov:
I added `documentattion-needed` - it will be nice if we can at least open an issue on k6-docs when we merge this

oleiade (approved):
LGTM 👍🏻 

Sorry my late review, I completely missed the notification 🤦🏻

PR #2703, opened 2022-09-30
atlas: url tag value always equal to the name tag

codebien:
This change was introduced in https://github.com/grafana/k6/pull/2654 but we decided to split it into a dedicated PR. The quote from the original PR and from the issue describe the reasons.

>One change from the original issue description has been the decision to have an indexable url tag, but to always have its value be equal to the name tag. See this for the details and the reason why: https://github.com/grafana/k6/issues/2584#issuecomment-1219618111

### DIFF
--- a/js/modules/k6/http/request_test.go
+++ b/js/modules/k6/http/request_test.go
@@ -47,11 +47,7 @@ import (
 )
 
 // TODO replace this with the Single version
-func assertRequestMetricsEmitted(t *testing.T, sampleContainers []metrics.SampleContainer, method, url, name string, status int, group string) {
-	if name == "" {
-		name = url
-	}
-
+func assertRequestMetricsEmitted(t *testing.T, sampleContainers []metrics.SampleContainer, method, url string, status int, group string) {
 	seenDuration := false
 	seenBlocked := false
 	seenConnecting := false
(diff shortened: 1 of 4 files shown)


js/modules/k6/http/request_test.go
```diff
@@ -83,17 +80,16 @@ func assertRequestMetricsEmitted(t *testing.T, sampleContainers []metrics.Sample
...
-	assert.True(t, seenSending, "url %s didn't emit Sending", url)
-	assert.True(t, seenWaiting, "url %s didn't emit Waiting", url)
-	assert.True(t, seenReceiving, "url %s didn't emit Receiving", url)
+	assert.True(t, seenDuration, "url %s didn't emit Duration", nameAndURL)
+	assert.True(t, seenBlocked, "url %s didn't emit Blocked", nameAndURL)
+	assert.True(t, seenConnecting, "url %s didn't emit Connecting", nameAndURL)
+	assert.True(t, seenTLSHandshaking, "url %s didn't emit TLSHandshaking", nameAndURL)
+	assert.True(t, seenSending, "url %s didn't emit Sending", nameAndURL)
+	assert.True(t, seenWaiting, "url %s didn't emit Waiting", nameAndURL)
+	assert.True(t, seenReceiving, "url %s didn't emit Receiving", nameAndURL)
```
olegbespalov, line 92: Looking for the changes, it seems like in most places, it doesn't really matter that variable names `nameAndURL` moreover, for the checks like above it even confuses :thinking: 

Maybe we stick with the just `url` variable name and comment on the assertion 

```
// URL equals name, because https://github.com/grafana/k6/issues/2584#issuecomment-1219618111
assert.Equal(t, tags["url"], tags["name"])
```

Or is any better shorter explanation, WDYT?

codebien: I addressed it in the latest force-pushed commit, let me know

olegbespalov (approved):
:muscle:

codebien (commented):
Apparently there is a data race :unamused:

NikolaiNikolaevichBoyko (approved):

PR #2706, opened 2022-10-01
Put PGP into a code block

nontw:
For easy copying.  It makes it stand out more than plain text too.


olegbespalov (approved):
@nontw thanks for fixing it :relaxed:

oleiade (approved):

PR #2708, opened 2022-10-03
Update x dependencies

mstoykov:


mstoykov:
The commit messages contain some relevant changes.

It seems like there are a bunch of http2 changes in the `x/net` - which might help both with performance but also likely with some rare cases of status 0.

na-- (approved):

codebien (approved):

PR #2709, opened 2022-10-03
Update dependencies

mstoykov:


codebien (dismissed):

olegbespalov (approved):

mstoykov:
I update `tidwall/pretty` in the end as not only `converter/har` was using it but also `tidwall/gjson` and the update is minimal

codebien (approved):

Issue #2711, opened 2022-10-04
Websockets should return HTTP error codes


brietaylor:
### Feature Description

Given the example code from https://k6.io/docs/using-k6/protocols/websockets/

```js
import ws from 'k6/ws';
import { check } from 'k6';

export default function () {
  const url = 'ws://echo.websocket.org';
  const params = { tags: { my_tag: 'hello' } };

  const res = ws.connect(url, params, function (socket) {
    socket.on('open', () => console.log('connected'));
    socket.on('message', (data) => console.log('Message received: ', data));
    socket.on('close', () => console.log('disconnected'));
  });

  check(res, { 'status is 101': (r) => r && r.status === 101 });
}
```

I'd like to extend this code with some custom behaviour for different status codes.  For example:

```js
if (res.status == 429) {
  console.log("Slow down");
  sleep(3); // backoff
} else {
  // error handling
}
```

However, if you try to run this code against a server that returns 429, `res.status` gets set to 0.  And there is no apparent workaround.

I think this is probably a bug and not a feature, given that the examples on the website explicitly check `res.status`.  Although I couldn't find any documentation indicated whether this was expected behaviour or not.

### Suggested Solution (optional)

I think this is a short change in `modules/k6/ws/ws.go`.  I've written a brief PR, which I'll link here.

### Already existing or connected issues / PRs (optional)

https://github.com/grafana/k6/pull/2712

mstoykov:
closed with #2712

PR #2712, opened 2022-10-04
Return HTTP response to user for failed connections.

brietaylor:
Allow users to process the error response, for example to implement and test backoff after receiving a 429 Slow Down.

I've also removed the log message for failed connections.  In my own load tests this is swamping out other log messages.

Issue link: https://github.com/grafana/k6/issues/2711

### DIFF
--- a/js/modules/k6/ws/ws_test.go
+++ b/js/modules/k6/ws/ws_test.go
@@ -729,6 +729,21 @@ func TestErrors(t *testing.T) {
 	})
 }
 
+func TestWrongStatusCode(t *testing.T) {
+	t.Parallel()
+	test := newTestState(t)
+	tb := httpmultibin.NewHTTPMultiBin(t)
+	sr := tb.Replacer.Replace
+	test.VU.StateField.Options.Throw = null.BoolFrom(false)
+	_, err := test.VU.Runtime().RunString(sr(`
+	var res = ws.connect("WSBIN_URL/status/404", function(socket){});
+	if (res.status != 404) {
+		throw new Error ("no status code set for invalid response");
+	}
+	`))
+	assert.NoError(t, err)
+}
+
 func TestSystemTags(t *testing.T) {
 	t.Parallel()
 	testedSystemTags := []string{"status", "subproto", "url", "ip"}
--- a/js/modules/k6/ws/ws_test.go
+++ b/js/modules/k6/ws/ws_test.go
@@ -1235,6 +1235,5 @@ func TestWSConnectDisableThrowErrorOption(t *testing.T) {
 		`)
 	assert.NoError(t, err)
 	entries := logHook.Drain()
-	require.Len(t, entries, 1)
-	assert.Contains(t, entries[0].Message, "Attempt to establish a WebSocket connection failed")
+	require.Len(t, entries, 0)
 }
--- a/js/modules/k6/ws/ws_test.go
+++ b/js/modules/k6/ws/ws_test.go
@@ -729,7 +729,7 @@ func TestErrors(t *testing.T) {
 	})
 }
 
-func TestWrongStatusCode(t *testing.T) {
+func TestConnectWrongStatusCode(t *testing.T) {
 	t.Parallel()
 	test := newTestState(t)
 	tb := httpmultibin.NewHTTPMultiBin(t)
@@ -1248,7 +1248,7 @@ func TestWSConnectDisableThrowErrorOption(t *testing.T) {
 			throw new Error("res.error is expected to be not null");
 		}
 		`)
-	assert.NoError(t, err)
+	require.NoError(t, err)
 	entries := logHook.Drain()
-	require.Len(t, entries, 0)
+	assert.Empty(t, entries)
 }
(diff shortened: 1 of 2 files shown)


mstoykov (changes requested):
Hi @brietaylor, thanks for the PR!

The change LGTM in general, but it lacks tests and breaks tests that expect this line to be logged.

I am not certain if I don't want the logging to stay (at least maybe in the other case). I would expect a lot of people are just not checking `ws.connect`  response at all :thinking: 

But I don't know I guess it will be really noisy. 

p.s. do you know of the [now experimental](https://github.com/grafana/k6/blob/master/release%20notes/v0.40.0.md#experimental-modules-2630-and-2656) `k6/experimental/websockets`

brietaylor:
Thanks for your comments!  I'm sorry for the delays, we've just been really busy over here.  I'll fix the tests.

Also I did not know about the experimental websockets!  Thanks for the link, I'll be sure to check them out in the future.

brietaylor:
I don't think I can run the full test suite locally on my laptop, I get a bunch of failures even on `master`.  (I might need some scaffolding?)

Anyways, the tests for `k6/js/modules/k6/ws` are now passing locally for me, I hope this will be enough!

brietaylor:
Hm, I'm not sure what the linter wants here?  I tried to match the style of the other tests, I don't think I'm comfortable to make bigger changes to the test framework (eg to parallelize it)

js/modules/k6/ws/ws_test.go
```diff
@@ -633,6 +633,17 @@ func TestErrors(t *testing.T) {
 		require.NoError(t, err)
 		assertSessionMetricsEmitted(t, metrics.GetBufferedSamples(test.samples), "", sr("WSBIN_URL/ws-close"), statusProtocolSwitch, "")
 	})
+
+	t.Run("wrong status code", func(t *testing.T) {
+		test.VU.StateField.Options.Throw = null.BoolFrom(false)
```
codebien, line 638:
```suggestion
       t.Parallel()
		test.VU.StateField.Options.Throw = null.BoolFrom(false)
```

codebien (changes requested):
Hi @brietaylor,
you need to call `t.Parallel()` in the sub-test.

js/modules/k6/ws/ws_test.go
mstoykov: Only this will lead to races due ot how the test is written.

I have opened a [PR](https://github.com/grafana/k6/pull/2732) fixing the tests and other lint issues. 

This has some changes of being merged for v0.41.0 (as it has practically no functional changes). But unfortunately this likely won't make it as we are now in the freeze period for code changes for v0.41.0. 

I would still recommend that once we merge the lint fixes, you rebase on top of master and push here. 

You will be able to copy paste one of the other use cases to call `t.Parallel()` and the rest of the test setup and copy your logic there. 

Sorry for the inconvience :(

mstoykov:
Hi @brietaylor as I mentioned in this [comment](https://github.com/grafana/k6/pull/2712#discussion_r996922558) I have fixed the linter issues and that PR is now merged so you can rebase on top of it. 

I would recommend making a completely new `Test<Something>` instead of adding this as a subcase somewhere.

Sorry for the confusion :(

brietaylor:
Ok, I can rebase this, but fair warning we have other things going on so it will take some time.

brietaylor:
@mstoykov Done with the rebase.  I also moved the test into its own function as you requested. Could you please review again?

Thanks,

Brie

js/modules/k6/ws/ws_test.go
```diff
@@ -729,6 +729,21 @@ func TestErrors(t *testing.T) {
 	})
 }
 
+func TestWrongStatusCode(t *testing.T) {
```
codebien, line 732:
```suggestion
func TestConnectWrongStatusCode(t *testing.T) {
```

js/modules/k6/ws/ws_test.go
```diff
@@ -1235,6 +1250,5 @@ func TestWSConnectDisableThrowErrorOption(t *testing.T) {
 		`)
 	assert.NoError(t, err)
 	entries := logHook.Drain()
-	require.Len(t, entries, 1)
-	assert.Contains(t, entries[0].Message, "Attempt to establish a WebSocket connection failed")
+	require.Len(t, entries, 0)
```
codebien, line 1253:
```suggestion
	assert.Empty(t, entries)
```

js/modules/k6/ws/ws_test.go
```diff
@@ -1235,6 +1250,5 @@ func TestWSConnectDisableThrowErrorOption(t *testing.T) {
 		`)
 	assert.NoError(t, err)
```
codebien, line 1251:
```suggestion
	require.NoError(t, err)
```

codebien (changes requested):
Hi @brietaylor,
thanks for your effort :bow:

LGTM, I added a few minor comments.

mstoykov:
I have marked it as breaking change as it *does* change what is returned and removes a warning log message.

mstoykov (approved):
LGTM! :bow: 

Thanks for all the work, and sorry we had some set backs ;(

codebien (approved):

PR #2713, opened 2022-10-05
k6/execution: Support non-indexable in vu.tags 

codebien:
Moved from https://github.com/grafana/k6/pull/2654 in a dedicated PR the JS API proposal for `non-indexable-tags`.

This is still under discussion.


codebien:
We have https://github.com/grafana/k6/pull/2654 as a backup so we don't really need of this PR.

PR #2716, opened 2022-10-06
js: ApplyCustomUserTags

codebien:
Refactoring of the logic for applying user-defined tags in a centralized function.
It also includes the change for always raising an exception in case of a denied type is provided. https://github.com/grafana/k6/issues/2653

Closes https://github.com/grafana/k6/issues/2653

### DIFF
--- a/js/common/tags.go
+++ b/js/common/tags.go
@@ -1,0 +1,57 @@
+package common
+
+import (
+	"fmt"
+	"reflect"
+
+	"github.com/dop251/goja"
+	"go.k6.io/k6/metrics"
+)
+
+// ApplyCustomUserTags modifies the given metrics.TagSet object with the
+// user specified custom tags.
+// It expects to receive the `keyValues` object in the `{key1: value1, key2:
+// value2, ...}` format.
+func ApplyCustomUserTags(rt *goja.Runtime, tags *metrics.TagSet, keyValues goja.Value) (*metrics.TagSet, error) {
+	if keyValues == nil || goja.IsNull(keyValues) || goja.IsUndefined(keyValues) {
+		return tags, nil
+	}
+
+	keyValuesObj := keyValues.ToObject(rt)
+
+	newTags := tags
+	var err error
+	for _, key := range keyValuesObj.Keys() {
+		if newTags, err = ApplyCustomUserTag(rt, newTags, key, keyValuesObj.Get(key)); err != nil {
+			return nil, err
+		}
+	}
+
+	return newTags, nil
+}
+
+// ApplyCustomUserTag modifies the given metrics.TagSet object with the
+// given custom tag and its value.
+func ApplyCustomUserTag(rt *goja.Runtime, tags *metrics.TagSet, key string, val goja.Value) (*metrics.TagSet, error) {
+	kind := reflect.Invalid
+	if typ := val.ExportType(); typ != nil {
+		kind = typ.Kind()
+	}
+
+	switch kind {
+	case
+		reflect.String,
+		reflect.Bool,
+		reflect.Int64,
+		reflect.Float64:
+
+		return tags.With(key, val.String()), nil
+
+	default:
+		return nil, fmt.Errorf(
+			"invalid value for metric tag '%s': "+
+				"only String, Boolean and Number types are accepted as a metric tag values",
+			key,
+		)
+	}
+}
--- a/js/modules/k6/execution/execution.go
+++ b/js/modules/k6/execution/execution.go
@@ -4,7 +4,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"reflect"
 	"time"
 
 	"github.com/dop251/goja"
@@ -214,8 +213,8 @@ func (mi *ModuleInstance) newVUInfo() (*goja.Object, error) {
 	}
 
 	err = o.Set("tags", rt.NewDynamicObject(&tagsDynamicObject{
-		Runtime: rt,
-		State:   vuState,
+		runtime: rt,
+		state:   vuState,
 	}))
 	return o, err
 }
@@ -297,72 +296,65 @@ func optionsAsObject(rt *goja.Runtime, options lib.Options) (*goja.Object, error
 }
 
 type tagsDynamicObject struct {
-	Runtime *goja.Runtime
-	State   *lib.State
+	runtime *goja.Runtime
+	state   *lib.State
 }
 
 // Get a property value for the key. May return nil if the property does not exist.
 func (o *tagsDynamicObject) Get(key string) goja.Value {
-	tag, ok := o.State.Tags.GetCurrentValues().Get(key)
-	if !ok {
-		return nil
+	tcv := o.state.Tags.GetCurrentValues()
+	if tag, ok := tcv.Get(key); ok {
+		return o.runtime.ToValue(tag)
 	}
-	return o.Runtime.ToValue(tag)
+	return nil
 }
 
-// Set a property value for the key. It returns true if succeed.
-// String, Boolean and Number types are implicitly converted
-// to the goja's relative string representation.
-// In any other case, if the Throw option is set then an error is raised
-// otherwise just a Warning is written.
+// Set a property value for the key. It returns true if succeed. String, Boolean
+// and Number types are implicitly converted to the goja's relative string
+// representation. An exception is raised in case a denied type is provided.
 func (o *tagsDynamicObject) Set(key string, val goja.Value) bool {
-	kind := reflect.Invalid
-	if typ := val.ExportType(); typ != nil {
-		kind = typ.Kind()
-	}
-	switch kind {
-	case
-		reflect.String,
-		reflect.Bool,
-		reflect.Int64,
-		reflect.Float64:
-
-		o.State.Tags.Modify(func(currentTags *metrics.TagSet) *metrics.TagSet {
-			return currentTags.With(key, val.String())
-		})
-		return true
-	default:
-		reason := "only String, Boolean and Number types are accepted as a Tag value"
-		if o.State.Options.Throw.Bool {
-			panic(o.Runtime.NewTypeError(reason))
+	var err error
+	o.state.Tags.Modify(func(tags *metrics.TagSet) *metrics.TagSet {
+		newTags, applyErr := common.ApplyCustomUserTag(o.runtime, tags, key, val)
+		if applyErr != nil {
+			err = applyErr
+			return tags
 		}
-		o.State.Logger.Warnf("the execution.vu.tags.Set('%s') operation has been discarded because %s", key, reason)
-		return false
+		return newTags
+	})
+	if err == nil {
+		return true
 	}
+	panic(o.runtime.NewTypeError(err.Error()))
 }
 
 // Has returns true if the property exists.
 func (o *tagsDynamicObject) Has(key string) bool {
-	_, ok := o.State.Tags.GetCurrentValues().Get(key)
-	return ok
+	ctv := o.state.Tags.GetCurrentValues()
+	if _, ok := ctv.Get(key); ok {
+		return true
+	}
+	return false
 }
 
-// Delete deletes the property for the key. It returns true on success (note, that includes missing property).
+// Delete deletes the property for the key. It returns true on success (note,
+// that includes missing property).
 func (o *tagsDynamicObject) Delete(key string) bool {
-	o.State.Tags.Modify(func(currentTags *metrics.TagSet) *metrics.TagSet {
-		return currentTags.Without(key)
+	o.state.Tags.Modify(func(tags *metrics.TagSet) *metrics.TagSet {
+		return tags.Without(key)
 	})
+
 	return true
 }
 
-// Keys returns a slice with all existing property keys. The order is not deterministic.
+// Keys returns a slice with all existing property keys. The order is not
+// deterministic.
 func (o *tagsDynamicObject) Keys() []string {
-	tags := o.State.Tags.GetCurrentValues().Map()
-	if len(tags) < 1 {
-		return nil
-	}
-	keys := make([]string, 0, len(tags))
-	for k := range tags {
+	ctv := o.state.Tags.GetCurrentValues()
+
+	tagsMap := ctv.Map()
+	keys := make([]string, 0, len(tagsMap))
+	for k := range tagsMap {
 		keys = append(keys, k)
 	}
 	return keys
--- a/js/modules/k6/execution/execution_test.go
+++ b/js/modules/k6/execution/execution_test.go
@@ -34,177 +34,119 @@ func setupTagsExecEnv(t *testing.T) *modulestest.Runtime {
 	return testRuntime
 }
 
-func TestVUTags(t *testing.T) {
+func TestVUTagsGet(t *testing.T) {
 	t.Parallel()
 
-	t.Run("Get", func(t *testing.T) {
-		t.Parallel()
+	tenv := setupTagsExecEnv(t)
+	tenv.MoveToVUContext(&lib.State{
+		Tags: lib.NewVUStateTags(metrics.NewRegistry().RootTagSet().With("vu", "42")),
+	})
+	tag, err := tenv.VU.Runtime().RunString(`exec.vu.tags["vu"]`)
+	require.NoError(t, err)
+	assert.Equal(t, "42", tag.String())
 
-		tenv := setupTagsExecEnv(t)
-		tenv.MoveToVUContext(&lib.State{
-			Tags: lib.NewVUStateTags(
-				metrics.NewRegistry().RootTagSet().WithTagsFromMap(map[string]string{"vu": "42"})),
-		})
-		tag, err := tenv.VU.Runtime().RunString(`exec.vu.tags["vu"]`)
-		require.NoError(t, err)
-		assert.Equal(t, "42", tag.String())
+	// not found
+	tag, err = tenv.VU.Runtime().RunString(`exec.vu.tags["not-existing-tag"]`)
+	require.NoError(t, err)
+	assert.Equal(t, "undefined", tag.String())
+}
 
-		// not found
-		tag, err = tenv.VU.Runtime().RunString(`exec.vu.tags["not-existing-tag"]`)
-		require.NoError(t, err)
-		assert.Equal(t, "undefined", tag.String())
+func TestVUTagsJSONEncoding(t *testing.T) {
+	t.Parallel()
+
+	tenv := setupTagsExecEnv(t)
+	tenv.MoveToVUContext(&lib.State{
+		Options: lib.Options{
+			SystemTags: metrics.NewSystemTagSet(metrics.TagVU),
+		},
+		Tags: lib.NewVUStateTags(metrics.NewRegistry().RootTagSet().With("vu", "42")),
+	})
+	tenv.VU.State().Tags.Modify(func(tags *metrics.TagSet) *metrics.TagSet {
+		return tags.With("custom-tag", "mytag1")
 	})
 
-	t.Run("JSONEncoding", func(t *testing.T) {
-		t.Parallel()
+	encoded, err := tenv.VU.Runtime().RunString(`JSON.stringify(exec.vu.tags)`)
+	require.NoError(t, err)
+	assert.JSONEq(t, `{"vu":"42","custom-tag":"mytag1"}`, encoded.String())
+}
 
-		tenv := setupTagsExecEnv(t)
-		tenv.MoveToVUContext(&lib.State{
-			Options: lib.Options{
-				SystemTags: metrics.NewSystemTagSet(metrics.TagVU),
-			},
-			Tags: lib.NewVUStateTags(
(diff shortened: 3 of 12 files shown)


js/modules/k6/execution/execution.go
```diff
@@ -297,72 +296,67 @@ func optionsAsObject(rt *goja.Runtime, options lib.Options) (*goja.Object, error
...
-// Set a property value for the key. It returns true if succeed.
-// String, Boolean and Number types are implicitly converted
-// to the goja's relative string representation.
-// In any other case, if the Throw option is set then an error is raised
-// otherwise just a Warning is written.
+// Set a property value for the key. It returns true if succeed. String, Boolean
+// and Number types are implicitly converted to the goja's relative string
+// representation. Objects are checked and accepted if they contain the index
+// and value keys. In any other case, if the Throw option is set then an error
+// is raised otherwise just a Warning is written.
```
na--, line 316: This comment about Throw is invalid, see https://github.com/grafana/k6/issues/2653
I think this PR copied the parts of https://github.com/grafana/k6/pull/2654 that fixed the issue? Though I see that even in my PR I forgot to adjust the comment :sweat_smile:

lib/netext/httpext/request.go
```diff
@@ -167,33 +167,20 @@ func MakeRequest(ctx context.Context, state *lib.State, preq *ParsedHTTPRequest)
...
-
 	// Only set the name system tag if the user didn't explicitly set it beforehand,
 	// and the Name was generated from a tagged template string (via http.url).
-	if _, ok := tags.Get("name"); !ok && state.Options.SystemTags.Has(metrics.TagName) &&
-		preq.URL.Name != "" && preq.URL.Name != preq.URL.Clean() {
-		tags = tags.With("name", preq.URL.Name)
-	}
-
-	// Check rate limit *after* we've prepared a request; no need to wait with that part.
-	if rpsLimit := state.RPSLimit; rpsLimit != nil {
```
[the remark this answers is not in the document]
na--: and btw the fact that no tests failed probably means we are lacking tests for this feature :disappointed:

js/modules/k6/grpc/client_test.go
```diff
@@ -656,106 +644,6 @@ func TestClient(t *testing.T) {
 				`,
 			},
 		},
-		{
-			name: "MaxReceiveSizeBadParam",
```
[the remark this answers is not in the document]
codebien: I think I did a mistake in resolve a conflict in the original PR, going to fix

lib/netext/httpext/request.go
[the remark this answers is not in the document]
codebien: Good catch, going to add one

js/modules/k6/execution/execution.go
codebien: Yeah, I would split the fix in one more commit, going to do soon

js/modules/k6/ws/ws.go
```diff
@@ -155,17 +155,11 @@ func (mi *WS) Connect(url string, args ...goja.Value) (*WSHTTPResponse, error) {
...
-				}
-				tagObj := tagsV.ToObject(rt)
-				if tagObj == nil {
-					continue
-				}
-				for _, key := range tagObj.Keys() {
-					tags = tags.With(key, tagObj.Get(key).String())
+				newTags, err := common.ApplyCustomUserTags(rt, tags, params.Get(k))
+				if err != nil {
+					return nil, err
```
olegbespalov, line 160: Not sure what the end user-faced error looks like, but if it's generic, I'd also suggest explicitly mentioning that this is coming from the WebSockets module. And basically it's relevant for any place like HTTP, groups, or whatever.

My point if we explicitly say "where" the issue is to the users, they will be happy :relaxed:

lib/netext/httpext/request_test.go
```diff
@@ -481,3 +488,65 @@ func TestMakeRequestTimeoutInTheBegining(t *testing.T) {
...
+			select {
+			case <-samples:
+			case <-ctx.Done():
+				return
+			}
+		}
+	}()
+
+	logger := logrus.New()
+	logger.Level = logrus.DebugLevel
```
olegbespalov, line 515: Please also use `logger.Out = io.Discard`

js/common/tags.go
```diff
@@ -0,0 +1,57 @@
...
+
+	"github.com/dop251/goja"
+	"go.k6.io/k6/metrics"
+)
+
+// ApplyCustomUserTags modifies the given metrics.TagSet object with the
+// user specified custom tags.
+// It expects to receive the `keyValues` object in the `{key1: value1, key2:
+// value2, ...}` format.
+func ApplyCustomUserTags(rt *goja.Runtime, tags *metrics.TagSet, keyValues goja.Value) (*metrics.TagSet, error) {
```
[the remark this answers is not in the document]
codebien: Yeah good spot, I don't think it's quite relevant. The function sets a tag from a JavaScript value, I guess the **custom user** part would enforce that, but having them in the `js` package probably should be enough.

@na-- coined the name so proxying the question to him if it was a different reason.

js/common/tags.go
```diff
@@ -0,0 +1,57 @@
...
+		reflect.String,
+		reflect.Bool,
+		reflect.Int64,
+		reflect.Float64:
+
+		return tags.With(key, val.String()), nil
+
+	default:
+		return nil, fmt.Errorf(
+			"setting '%s' metric tag failed;"+
```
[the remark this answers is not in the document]
codebien: Why do you want to do that? This will be hit by the users, I think we should keep a user-friendly experience. A code's function name reported in the error I don't helps. I'm happy to improve the message if you think is not clear yet enough.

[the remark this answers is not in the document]
olegbespalov: Nope, I didn't mean that we should mention the function name in error what I meant was that the error message ideally should match the function name, which helps maintainers/contributors. 

Like since the error message is saying `"setting '%s' metric tag failed;"` it will be good if the method names something like `SetTag` instead of the applying. Or if the `ApplyTag` will be chosen then the error message should be `"...applying tag failed..."`. So the point is that they should be consistent.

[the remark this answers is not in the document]
codebien: Gotcha, I agree! I wait for the final decision about the naming then I will change this in a consistent way.

js/modules/k6/execution/execution_test.go
```diff
@@ -34,177 +34,121 @@ func setupTagsExecEnv(t *testing.T) *modulestest.Runtime {
...
+	cases := []string{
+		"null",
+		"undefined",
+		"[]",
+		"{}",
+		`[1, 3, 5]`,
+		`{f1: "value1", f2: 4}`,
+		`{"foo": "bar"}`,
+		`{"value": "foo"}`, // need index as well
+		`{"index": "foo"}`, // need value as well
```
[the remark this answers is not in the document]
codebien: No, it shouldn't and this is some not well-migrated part. I'm going to re-arrange it better.

js/common/tags.go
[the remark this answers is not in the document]
na--: A bunch of things in the `js` package (almost exclusively, in fact :sweat_smile:) also emit the system metric tags, so I think it makes sense to keep the current name. After all, we mostly do a bunch of validation here, which we don't do for non-user-specified tags.

js/common/tags.go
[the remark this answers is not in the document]
na--: Since users would usually see this error in another context, e.g. when making an http request with `http.get(url, tags: {foo: somethingInvalid})`, I'd suggest something like this:
```suggestion
			"invalid value for metric tag '%s': "+
```

js/modules/k6/execution/execution.go
```diff
@@ -297,72 +296,66 @@ func optionsAsObject(rt *goja.Runtime, options lib.Options) (*goja.Object, error
...
 }
 
-// Set a property value for the key. It returns true if succeed.
-// String, Boolean and Number types are implicitly converted
-// to the goja's relative string representation.
-// In any other case, if the Throw option is set then an error is raised
-// otherwise just a Warning is written.
+// Set a property value for the key. It returns true if succeed. String, Boolean
+// and Number types are implicitly converted to the goja's relative string
+// representation. Objects are checked and accepted if they contain the index
```
na--, line 314: Please also add a mention in the PR description that it fixes https://github.com/grafana/k6/issues/2653

js/modules/k6/metrics/metrics.go
```diff
@@ -109,12 +109,11 @@ func (m Metric) add(v goja.Value, addTags goja.Value) (bool, error) {
...
 
 	tags := state.Tags.GetCurrentValues()
-	if addTags != nil {
-		tagsobj := addTags.ToObject(m.vu.Runtime())
-		for _, key := range tagsobj.Keys() {
-			tags = tags.With(key, tagsobj.Get(key).String())
-		}
+	newTags, err := common.ApplyCustomUserTags(m.vu.Runtime(), tags, addTags)
+	if err != nil {
+		return false, fmt.Errorf("failed to apply the defined tags: %w", err)
```
na--, line 114: this error wrapping is more confusing to the user than helpful... they don't care that we internally "apply" tags, they are just passing a JS object with `{key: value}` tags. So, there either shouldn't be any error wrapping here, or it should be something specific for where the error originates, like this:
```suggestion
		return false, fmt.Errorf("invalid custom metric tags: %w", err)
```
or maybe even add the `Metric`'s name in the error? :thinking:

js/modules/k6/http/request.go
```diff
@@ -307,19 +308,11 @@ func (c *Client) parseRequest(
...
-				if tagObj == nil {
-					continue
-				}
-				tagKeys := tagObj.Keys()
-				result.Tags = make([][2]string, 0, len(tagKeys))
-				for _, key := range tagKeys {
-					result.Tags = append(result.Tags, [2]string{key, tagObj.Get(key).String()})
+				newTags, err := common.ApplyCustomUserTags(rt, result.Tags, params.Get(k))
+				if err != nil {
+					return nil, fmt.Errorf("failed to apply the defined tags: %w", err)
```
na--, line 313: Same as https://github.com/grafana/k6/pull/2716/files#r990395132, "apply" shouldn't be part of the error because the users are not "applying" anything. Either no wrapping or something like this:
```suggestion
					return nil, fmt.Errorf("invalid HTTP request metric tags: %w", err)
```

js/modules/k6/ws/ws.go
```diff
@@ -155,17 +155,11 @@ func (mi *WS) Connect(url string, args ...goja.Value) (*WSHTTPResponse, error) {
...
-				}
-				tagObj := tagsV.ToObject(rt)
-				if tagObj == nil {
-					continue
-				}
-				for _, key := range tagObj.Keys() {
-					tags = tags.With(key, tagObj.Get(key).String())
+				newTags, err := common.ApplyCustomUserTags(rt, tags, params.Get(k))
+				if err != nil {
+					return nil, fmt.Errorf("failed to apply the defined tags: %w", err)
```
na--, line 160: same as the other comments

na-- (dismissed):
LGTM except the minor issues I left as inline comments

js/modules/k6/metrics/metrics.go
codebien:
```suggestion
		return false, fmt.Errorf("'%s' custom metric defines invalid tags: %w", m.metric.Name, err)
```
Changed with this

js/modules/k6/grpc/client.go
```diff
@@ -84,7 +84,7 @@ func (c *Client) Connect(addr string, params map[string]interface{}) (bool, erro
 
 	p, err := c.parseConnectParams(params)
 	if err != nil {
-		return false, err
+		return false, fmt.Errorf("Connect request defines invalid parameters: %w", err)
```
olegbespalov, line 87: Maybe be more precise, since Connect is a generic one? WDYT?
```suggestion
		return false, fmt.Errorf("grpc.Connect has invalid parameters: %w", err)
```

js/modules/k6/grpc/client.go
```diff
@@ -161,9 +161,9 @@ func (c *Client) Invoke(
 		return nil, fmt.Errorf("method %q not found in file descriptors", method)
 	}
 
-	p, err := c.parseParams(params)
+	p, err := c.parseInvokeParams(params)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("Invoke request defines invalid parameters: %w", err)
```
olegbespalov, line 166: Same
```suggestion
		return nil, fmt.Errorf("grpc.Invoke has invalid parameters: %w", err)
```

js/modules/k6/ws/ws.go
```diff
@@ -155,17 +155,11 @@ func (mi *WS) Connect(url string, args ...goja.Value) (*WSHTTPResponse, error) {
...
-				}
-				tagObj := tagsV.ToObject(rt)
-				if tagObj == nil {
-					continue
-				}
-				for _, key := range tagObj.Keys() {
-					tags = tags.With(key, tagObj.Get(key).String())
+				newTags, err := common.ApplyCustomUserTags(rt, tags, params.Get(k))
+				if err != nil {
+					return nil, fmt.Errorf("invalid WebSocket request metric tags: %w", err)
```
olegbespalov, line 160:
```suggestion
					return nil, fmt.Errorf("ws.Connect has metric tags: %w", err)
```

olegbespalov (dismissed):
LGTM in general, left a few non-blocking comments :+1:

js/modules/k6/execution/execution.go
```diff
@@ -297,72 +296,66 @@ func optionsAsObject(rt *goja.Runtime, options lib.Options) (*goja.Object, error
...
+	o.state.Tags.Modify(func(tags *metrics.TagSet) *metrics.TagSet {
+		return tags.Without(key)
 	})
+
 	return true
 }
 
-// Keys returns a slice with all existing property keys. The order is not deterministic.
+// Keys returns a slice with all existing property keys. The order is not
+// deterministic.
```
codebien, line 352: >The order is not deterministic.

This is not true anymore, going to remove

js/modules/k6/grpc/client.go
```diff
@@ -117,7 +117,7 @@ func (c *Client) Connect(addr string, params map[string]interface{}) (bool, erro
 
 	p, err := c.parseConnectParams(params)
 	if err != nil {
-		return false, err
+		return false, fmt.Errorf("grpc.Connect has invalid parameters: %w", err)
```
na--, line 120:
```suggestion
		return false, fmt.Errorf("invalid grpc.Connect() parameters: %w", err)
```

js/modules/k6/grpc/client.go
```diff
@@ -215,29 +215,21 @@ func (c *Client) Invoke(
...
-		tags = tags.With("service", parts[0])
+		p.Tags = p.Tags.With("service", parts[0])
 	}
 	if state.Options.SystemTags.Has(metrics.TagMethod) {
-		tags = tags.With("method", parts[1])
-	}
-
-	// Only set the name system tag if the user didn't explicitly set it beforehand
-	if _, ok := tags.Get("name"); !ok && state.Options.SystemTags.Has(metrics.TagName) {
-		tags = tags.With("name", method)
```
na--, line 234: Hmm where is this done now? he `name` tag should be set to be equal to the `url` tag, if the user didn't explicitly specify it.

js/modules/k6/metrics/metrics.go
```diff
@@ -109,12 +109,11 @@ func (m Metric) add(v goja.Value, addTags goja.Value) (bool, error) {
...
 
 	tags := state.Tags.GetCurrentValues()
-	if addTags != nil {
-		tagsobj := addTags.ToObject(m.vu.Runtime())
-		for _, key := range tagsobj.Keys() {
-			tags = tags.With(key, tagsobj.Get(key).String())
-		}
+	newTags, err := common.ApplyCustomUserTags(m.vu.Runtime(), tags, addTags)
+	if err != nil {
+		return false, fmt.Errorf("'%s' custom metric defines invalid tags: %w", m.metric.Name, err)
```
na--, line 114:
```suggestion
		return false, fmt.Errorf("cannot add tags for the '%s' custom metric: %w", m.metric.Name, err)
```

na--:
LGTM except the error message nitpicks and the gRPC `name` tag issue.

js/modules/k6/grpc/client.go
codebien: mmm...good catch, my fault. And apparently, we are not covering with tests this part.

codebien:
@olegbespalov do you agree with the latest suggestions?

js/modules/k6/grpc/client.go
na--: :disappointed: yeah...

olegbespalov:
yeah, the most important they're pointing to exact place, so I'm happy :relaxed:

js/modules/k6/grpc/client.go
codebien: Adding the test I hit the issue that the URL is not set equal to the Name and I guess that is what we expect after https://github.com/grafana/k6/pull/2703 so I would prefer to add it in a dedicated bugfix PR.

olegbespalov (approved):
LGTM, with one nit (non-blocking) :+1:

js/modules/k6/execution/execution.go
```diff
@@ -297,72 +296,65 @@ func optionsAsObject(rt *goja.Runtime, options lib.Options) (*goja.Object, error
...
-		return true
-	default:
-		reason := "only String, Boolean and Number types are accepted as a Tag value"
-		if o.State.Options.Throw.Bool {
-			panic(o.Runtime.NewTypeError(reason))
+	var err error
+	o.state.Tags.Modify(func(tags *metrics.TagSet) *metrics.TagSet {
+		newTags, applyErr := common.ApplyCustomUserTag(o.runtime, tags, key, val)
+		if applyErr != nil {
+			err = applyErr
```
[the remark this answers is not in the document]
olegbespalov: I see :+1: anyway it's was just a non-blocking nit :sweat_smile:

js/modules/k6/grpc/client.go
na--: Hmm IIRC the `url` tag in gRPC has a much lower cardinality than in HTTP requests, it is just the `server/method_name`, right? So I am not sure we need to have the same "`url` tag should always be equal to the `name` tag" logic in the gRPC module that we had to add to the `k6/http` module. gRPC doesn't have any GET parameters, which is the big problem with HTTP.

js/modules/k6/grpc/client.go
```diff
@@ -117,7 +117,7 @@ func (c *Client) Connect(addr string, params map[string]interface{}) (bool, erro
 
 	p, err := c.parseConnectParams(params)
 	if err != nil {
-		return false, err
+		return false, fmt.Errorf("invalid grpc.Connect() parameters: %w", err)
```
na--, line 120:
```suggestion
		return false, fmt.Errorf("invalid grpc.connect() parameters: %w", err)
```

The actual JS method name is with lowercase, see https://k6.io/docs/javascript-api/k6-net-grpc/client/client-connect

js/modules/k6/grpc/client.go
```diff
@@ -194,9 +194,9 @@ func (c *Client) Invoke(
 		return nil, fmt.Errorf("method %q not found in file descriptors", method)
 	}
 
-	p, err := c.parseParams(params)
+	p, err := c.parseInvokeParams(params)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("invalid grpc.Invoke() parameters: %w", err)
```
na--, line 199:
```suggestion
		return nil, fmt.Errorf("invalid grpc.invoke() parameters: %w", err)
```

https://k6.io/docs/javascript-api/k6-net-grpc/client/client-invoke

js/modules/k6/ws/ws.go
```diff
@@ -155,17 +155,11 @@ func (mi *WS) Connect(url string, args ...goja.Value) (*WSHTTPResponse, error) {
...
-				}
-				tagObj := tagsV.ToObject(rt)
-				if tagObj == nil {
-					continue
-				}
-				for _, key := range tagObj.Keys() {
-					tags = tags.With(key, tagObj.Get(key).String())
+				newTags, err := common.ApplyCustomUserTags(rt, tags, params.Get(k))
+				if err != nil {
+					return nil, fmt.Errorf("invalid ws.Connect() metric tags: %w", err)
```
na--, line 160:
```suggestion
					return nil, fmt.Errorf("invalid ws.connect() metric tags: %w", err)
```

https://k6.io/docs/javascript-api/k6-ws/connect/

na--:
LGTM except the wrong method capitalization in error messages

js/modules/k6/grpc/client.go
codebien: Added an issue for addressing it with a specific discussion in a issue + PR. https://github.com/grafana/k6/issues/2722

na-- (approved):

olegbespalov (approved):

PR #2720, opened 2022-10-10
Adjust level of the warning for an invalid system variables

olegbespalov:
# What?

Change the log level from a warning to debug for the message about the invalid system variables to improve UX on Windows machines, where by default exist variables like 'CommonProgramFiles(x86)' or 'ProgramFiles(x86)'.

Finding the message when k6 verbose mode is enabled is still possible.

image

# Why?

It's not a good UX when we show this by default to all windows users.

### DIFF
--- a/cmd/runtime_options.go
+++ b/cmd/runtime_options.go
@@ -111,7 +111,7 @@ func getRuntimeOptions(
 	if opts.IncludeSystemEnvVars.Bool { // If enabled, gather the actual system environment variables
 		for k, v := range environment {
 			if !userEnvVarName.MatchString(k) {
-				logger.Warnf("invalid system environment variable name '%s'", k)
+				logger.Debugf("invalid system environment variable name '%s', will be ignored", k)
 
 				continue
 			}


cmd/runtime_options.go
```diff
@@ -111,7 +111,7 @@ func getRuntimeOptions(
 	if opts.IncludeSystemEnvVars.Bool { // If enabled, gather the actual system environment variables
 		for k, v := range environment {
 			if !userEnvVarName.MatchString(k) {
-				logger.Warnf("invalid system environment variable name '%s'", k)
+				logger.Debugf("invalid system environment variable name '%s', will be ignored", k)
```
codebien, line 114:
```suggestion
				logger.Debugf("Invalid system environment variable name '%s', will be ignored", k)
```

codebien (changes requested):
LGTM, just a nitpick

codebien (approved):

na--:
Hmm if these environment variables are always present on Windows machines, then wouldn't that make them valid by definition? :sweat_smile: Maybe we need to relax our constraints instead of adjusting the warning message severity?

olegbespalov:
But is there a case when we users need them inside the k6? :thinking:

imiric (approved):
Thanks, this is better UX :+1:

But I'm wondering two things:
- Why are only these ones reported? My env on Windows 10 has many more vars defined that are not mentioned.
- What's with the empty name one (`''`)? :thinking:

olegbespalov:
> Why are only these ones reported? My env on Windows 10 has many more vars defined that are not mentioned.

These have invalid names [from the k6 perspective](https://github.com/grafana/k6/blob/c6f57c6e22e6966e526643d2d0944e48df0ea36c/cmd/runtime_options.go#L19), do others that you mention also have similar names?

The one with "name" `''` is a weird one, yeah. And briefly looking into the list of the environment variables in the WM machine I see nothing that can cause it :shrug:

imiric:
> These have invalid names [from the k6 perspective](https://github.com/grafana/k6/blob/c6f57c6e22e6966e526643d2d0944e48df0ea36c/cmd/runtime_options.go#L19)

Ah, I see. Then I agree with @na--. If these are logged because of an arbitrary validation regex, we might want to expand the regex to include these symbols, since these are always present on Windows AFAICT.

But I'm still not clear what a "valid" env var name is from the k6 perspective, and why we wouldn't just pass all the ones that are set. But I wasn't part of the discussion in #2652 and what it aimed to fix, so no need to open that discussion here. Maybe we can discuss it on Slack.

olegbespalov:
@imiric, there was no discussion in #2652 about the pattern, the pattern, and maybe some decision about why it so made a long time ago.

I'm fine to extend the pattern. However, it's good to know the reason (`Why?`) we're going to do that. Is there any use case where these environment variables are helpful inside the k6?

na--:
The question of whether they are helpful or not is secondary.

Apparently environment variables with names that contain parentheses exist in the wild. They seem perfectly fine and are even specified by default in Windows. So, the k6 validation is probably too strict and needs to be relaxed.

The [original validation](https://github.com/grafana/k6/pull/495/files#diff-4af4cfe74652bd08dab633185b71ed80f8669e5206014cceb22961033ffd6f87R78-R80) was added by me, in [my first ever k6 PR](https://github.com/grafana/k6/pull/495), in fact :sweat_smile: And I don't remember a specific reason why I was so restrictive, so unless someone can provide one, I don't see a reason not to relax the constraint. k6 scripts don't care what the environment variable names are, they can be anything that's a valid JSON/JS object index :sweat_smile:

PR #2726, opened 2022-10-13
metrics.Sample.Metadata: Data structure for high cardinality tags

codebien:
This is built on top of https://github.com/grafana/k6/pull/2654 and https://github.com/grafana/k6/pull/2594 with a more limited scope and it closes https://github.com/grafana/k6/issues/2584.

One of the main use cases for this support of high-cardinality metric tags/metadata is to allow k6 to support traces and to export the `trace_id` and `span_id` values it generated to something external like Tempo and Jaeger. The goal is to eventually have this with the new k6 HTTP API. However, this PR does not add a direct JavaScript API so the direct use of this feature will be only available from an extension.

One change from the original issue description has been the decision to not have a `raw_url` non-indexable url metadata and have a separate issue for tracking feedback and ideas.


js/modules/k6/execution/execution.go
```diff
@@ -314,13 +317,8 @@ func (o *tagsDynamicObject) Get(key string) goja.Value {
...
 	var err error
-	o.state.Tags.Modify(func(tags *metrics.TagSet) *metrics.TagSet {
-		newTags, applyErr := common.ApplyCustomUserTag(o.runtime, tags, key, val)
-		if applyErr != nil {
-			err = applyErr
-			return tags
-		}
-		return newTags
+	o.state.Tags.Modify(func(tagsAndMeta *metrics.TagsAndMeta) {
+		err = common.ApplyCustomUserTag(o.runtime, tagsAndMeta, key, val)
```
olegbespalov, line 321: just a nitpick, if we panic here, we can keep the scopes cleaner:

like:
```
o.state.Tags.Modify(func(tagsAndMeta *metrics.TagsAndMeta) {
		if err := common.ApplyCustomUserTag(o.runtime, tagsAndMeta, key, val); err != nil {
            panic(o.runtime.NewTypeError(err.Error()))
        }
}

return true
```

lib/netext/httpext/transport.go
```diff
@@ -80,75 +80,51 @@ func (t *transport) measureAndEmitMetrics(unfReq *unfinishedRequest) *finishedRe
 		trail:             trail,
 	}
 
-	tags := t.tags
+	ctm := t.tagsAndMeta.Clone()
```
olegbespalov, line 83: I believe the context of the usage of this variable is big enough to give it a full name like `clonedTagsAndMetadata` (or just a `clonedTags`, or even better to call it `sampleTags` which I guess they are) since it can be essential to know downstream.

metrics/tags.go
```diff
@@ -156,6 +156,82 @@ var _ interface {
...
+func (tm *TagsAndMeta) Delete(key string) {
+	tm.Tags = tm.Tags.Without(key)
+	if tm.Metadata != nil {
+		delete(tm.Metadata, key)
+	}
+}
+
+// Clone makes a complete copy of the current TagsAndMeta object and returns it.
+func (tm TagsAndMeta) Clone() TagsAndMeta {
+	res := TagsAndMeta{Tags: tm.Tags}
```
olegbespalov, line 224: Just double-checking since Atlas struct has pointers; it's a safe way to cone it?

lib/netext/httpext/transport.go
na--: IIRC, `ctm` is shorthand for `currentTagsAndMetadata`, so we can just use the `tagsAndMeta` here. 

Cloning is not the important bit here, this basically equivalent to `vu.State().Tags.GetCurrentValues()` (without the locking). We are just making a copy of the base `TagsAndMeta` object that is in `transport` because we might emit more than one sample (e.g. due to redirects). And the `transport` needs a custom copy of the tags+meta because its values can be modified from the script, e.g. `http.get(url, {tags: {foo: 'bar'}}).

metrics/tags.go
na--: Yes, it has pointers precisely so it can be cloned safely like this. An `*atlas.Node` pointer is essentially a way to represent a static `key: value` string map. If two `*atlas.Node` pointers with the same root are equal, then they contain the same map values.

This is why it is vitally important for all `TagSet` values in a k6 test run to always start from the same atlas root node: https://github.com/grafana/k6/blob/4ae7d1e8db25b60433c77b7fde18f458b444e168/metrics/registry.go#L23-L25

metrics/engine/engine.go
```diff
@@ -87,23 +87,19 @@ func (me *MetricsEngine) getThresholdMetricOrSubmetric(name string) (*metrics.Me
...
-		me.logger.Warnf("Thresholds like '%s', based on the high-cardinality 'error' metric tag, "+
-			"are deprecated and will not be supported in future k6 releases. "+
-			"To prevent breaking changes and reduce bugs, use the 'error_code' metric tag instead", name,
-		)
-	}
 	if _, ok := sm.Tags.Get("vu"); ok {
-		me.logger.Warnf("Thresholds like '%s', based on the high-cardinality 'vu' metric tag, "+
-			"are deprecated and will not be supported in future k6 releases.", name,
+		me.logger.Warnf(
+			"The high-cardinality 'vu' metric tag was made non-indexable in k6 v0.41.0, so thresholds"+
```
mstoykov, line 92: nit; maybe change this message to not use `non-indexable` - I am not certain we should use `metadata` yet :shrug: 


maybe non-indexable is correct enough :shrug:

.github/workflows/xk6-tests/xk6-js-test/jstest.go
```diff
@@ -58,7 +58,7 @@ func (j *JSTest) Foo(arg float64) (bool, error) {
 
 	ctx := j.vu.Context()
 
-	tags := state.Tags.GetCurrentValues().With("foo", "bar")
+	tags := state.Tags.GetCurrentValues().Tags.With("foo", "bar")
```
mstoykov, line 61: this seems bad .... Maybe just rename `Tags` to `TagsAndMetadata` for now?

I feel like it will be better if they are separate fields in `state` but this might take longer to do and I can see benefits for both, so I am for renaming *and* opening an issue to discuss the internal API and maybe make changes for the next release.

mstoykov (approved):
LGTM!

I did a really fast pass over it, so might've have missed something. 

Some things that we might want to fix (later):
1. It is a bit inconsistant between using metadata and non-indexable. Both in the code and in the messaging to the user. Some of those likely can be mass changed now.
2. I find `state.Tags` (and if we rename it to `state.TagsAndMetadata`) a bit ... strange. As it is combined in the state, but then we edit the separate parts and put it separately in the final samples.
3. Emitting samples with the current tags never was ... nice as you needed to get the tags from one place add yours and then set the whole thing in the samples. I guess now that you *need* to get tags object from somewhere it will guide users to actuallyt get it from the lib.State. But it still clunky IMO. This is even more clunky with metadata - I would expect in most cases code will jsut *not* set metadata. Arguably having it combined in `state.TagsAndMetadata` guides to this as well :shrug: . I just find the whole thing not nice and not intuitive and not so hard to get wrong. 

All of those though should likely be done in separate PRs ... likely *after* we have used the new code for a bit. And have discussed the js API for editing metadata/non-indexable nontags

js/modules/k6/execution/execution.go
na--: I am not sure if the lambda won't be shown in the JS stracktrace if we panic directly from it? :thinking:

metrics/engine/engine.go
codebien: I don't think replacing non-indexable with metadata is better. If we can find something clearer I'm ok with it, otherwise, I would prefer to let as-is. I haven't a better suggestion at the moment.

.github/workflows/xk6-tests/xk6-js-test/jstest.go
na--: One of the big reasons for `TagsAndMetadata` to be packaged together like this was because, in the original PR, we treated them as 2 flavors of tags on the user-facing side, but internally we were saving them in completely different data structures. 

We needed to lock them at the same time and we needed to remove corresponding tags when adding metadata and vice-versa: 
https://github.com/grafana/k6/blob/cd1e21d128251cd0c58c805acb18a62223be0537/metrics/tags.go#L176-L180
https://github.com/grafana/k6/blob/cd1e21d128251cd0c58c805acb18a62223be0537/metrics/tags.go#L185-L186

Now that it's most likely we won't need to have differently-flavored (indexed / non-indexed) tags, a lot of this feels unnecessary :thinking: I guess the common locking and helper methods are still good to have, as well as the internal "accumulator" for the final atlas `Tags` value, and we don't have a lot of time to rewrite this now... :disappointed: But :+1: for renaming `state.Tags` to `state.TagsAndMeta`

codebien: Considering we are going to open an issue I'm for not renaming now, and including the refactor as the first thing in the next release. @na-- WDYT?

metrics/tags.go
```diff
@@ -156,6 +156,82 @@ var _ interface {
...
+	Metadata map[string]string // could be nil
+}
+
+// SetTag adds the given key=value tag to the Tags and removes it from the
+// Metadata, if it was present there.
+func (tm *TagsAndMeta) SetTag(key, value string) {
+	tm.Tags = tm.Tags.With(key, value)
+	if tm.Metadata != nil {
+		delete(tm.Metadata, key)
+	}
```
na--, line 180: This (along with the comment above) probably needs to be removed. It doesn't make sense if we treat tags and metadata as completely different entities and not 2 flavors of the same thing.

metrics/tags.go
```diff
@@ -156,6 +156,82 @@ var _ interface {
...
+}
+
+// SetMetadata adds the given key=value datum to the Metadata and removes it
+// from the Tags, if it was present there.
+func (tm *TagsAndMeta) SetMetadata(key, value string) {
+	tm.Tags = tm.Tags.Without(key)
```
na--, line 186: Same as https://github.com/grafana/k6/pull/2726/files#r994451121 for this line (and the comment above).

metrics/tags.go
```diff
@@ -156,6 +156,82 @@ var _ interface {
...
+	}
+}
+
+// Delete removes the given key from both the tags and the metadata.
+func (tm *TagsAndMeta) Delete(key string) {
+	tm.Tags = tm.Tags.Without(key)
+	if tm.Metadata != nil {
+		delete(tm.Metadata, key)
+	}
+}
```
na--, line 220: This probably also needs to be split into 2 different methods - `DeleteTag` and `DeleteMetadata` :thinking:

.github/workflows/xk6-tests/xk6-js-test/jstest.go
na--: :+1: I am also fine with tackling this in next release, with maybe just a TODO in `lib.State` for now

js/modules/k6/execution/execution.go
codebien: Added in this fixup because it seems fine to me https://github.com/grafana/k6/pull/2726/commits/2428c114bbe91aff3e1685e6f47f1453aa561949

lib/netext/httpext/transport.go
codebien: https://github.com/grafana/k6/pull/2726/commits/2428c114bbe91aff3e1685e6f47f1453aa561949

lib/netext/httpext/request.go
```diff
@@ -181,13 +181,18 @@ func MakeRequest(ctx context.Context, state *lib.State, preq *ParsedHTTPRequest)
...
 	if state.Options.HTTPDebug.String != "" {
 		// Combine tags with common log fields
 		combinedLogFields := map[string]interface{}{"source": "http-debug", "vu": state.VUID, "iter": state.Iteration}
-		for k, v := range preq.Tags.Map() {
+		for k, v := range preq.TagsAndMeta.Tags.Map() {
+			if _, present := combinedLogFields[k]; !present {
+				combinedLogFields[k] = v
+			}
+		}
+		for k, v := range preq.TagsAndMeta.Metadata {
```
na--, line 195: Just in case, we should probably reverse the order here - first add the Metadata entries and then the Tags. That way, if there is a duplicate key, the tag value will be shown (which is probably more important :thinking:

na-- (approved):
Thank you! :bow:

metrics/tags_test.go
```diff
@@ -109,6 +109,8 @@ func TestTagSetContains(t *testing.T) {
 	assert.False(t, st.Contains(outer))
 }
 
+// TODO: Add TagsAndMeta tests
```
olegbespalov, line 112: :eyes:  should we add them with this PR?

olegbespalov (approved):
In general, LGTM :+1: :muscle: Left a small non-blocking comment.

lib/netext/httpext/request.go
codebien: Should we split them into two sub-json-objects? I think could be relevant to see both, especially in debugging phases.

```json
{
  // ...
  "iter": 10,
  "tags": {"key1":"val1",...},
  "meta": {"meta1":"val1"}
}
```

.github/workflows/xk6-tests/xk6-js-test/jstest.go
codebien: Added https://github.com/grafana/k6/issues/2733 for tracking this

lib/netext/httpext/request.go
na--: I don't think this is possible, and even if it is, it won't look nice with logrus. Since it's such a minor concern and only matters for `--http-debug`, I think we'd be fine with the current solution, just flipping the tags.

mstoykov (dismissed):

na-- (dismissed):

codebien:
@olegbespalov @na-- I added the tests and reversed the order of inserting tagsAndMeta for the logs.
If you will approve I will go through the process for pinning the right experimental version for xk6-websocket.

na-- (approved):

mstoykov (dismissed):

mstoykov (approved):

olegbespalov (approved):

na-- (approved):

PR #2738, opened 2022-10-18
Fixes for promise rejection reporting

mstoykov:
- fix a panic on `Promise.reject()` - rejection with `undefined` reason.
- include the actual rejection instead of only it's stack in case the stack exists.

### DIFF
--- a/js/eventloop/eventloop.go
+++ b/js/eventloop/eventloop.go
@@ -176,7 +176,6 @@ func (e *EventLoop) Start(firstCallback func() error) error {
 		// This will get a random unhandled rejection instead of the first one, for example.
 		// But that seems to be the case in other tools as well so it seems to not be that big of a problem.
 		for promise := range e.pendingPromiseRejections {
-			// TODO maybe throw the whole promise up and get make a better message outside of the event loop
 			value := promise.Result()
 			if !goja.IsNull(value) && !goja.IsUndefined(value) {
 				if o := value.ToObject(e.vu.Runtime()); o != nil {
--- a/js/eventloop/eventloop.go
+++ b/js/eventloop/eventloop.go
@@ -178,10 +178,12 @@ func (e *EventLoop) Start(firstCallback func() error) error {
 		for promise := range e.pendingPromiseRejections {
 			// TODO maybe throw the whole promise up and get make a better message outside of the event loop
 			value := promise.Result()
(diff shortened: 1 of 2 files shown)


na-- (dismissed):

js/eventloop/eventloop_test.go
```diff
@@ -143,3 +144,72 @@ func TestEventLoopPanicOnDoubleCallback(t *testing.T) {
...
+func TestEventLoopRejectString(t *testing.T) {
+	t.Parallel()
+	vu := &modulestest.VU{RuntimeField: goja.New()}
+	loop := eventloop.New(vu)
+	err := loop.Start(func() error {
+		_, err := vu.Runtime().RunString("Promise.reject('some string')")
+		return err
+	})
+	loop.WaitOnRegistered()
+	require.EqualError(t, err, "Uncaught (in promise) some string")
```
imiric, line 169: This reads slightly wrong to me. `Uncaught (in promise)` implies that an error was thrown in the promise and not handled, not that the promise was rejected with a specific value (or `undefined` as in the above test). The tests below do make sense, though.

mstoykov: This is the wording of literally every other JS running thing I have looked at: firefox, chrome, deno, node.  Everything uses this in my experience/experiments

imiric: OK then. Still sounds confusing to me :smile:

js/eventloop/eventloop.go
```diff
@@ -178,10 +178,12 @@ func (e *EventLoop) Start(firstCallback func() error) error {
 		for promise := range e.pendingPromiseRejections {
 			// TODO maybe throw the whole promise up and get make a better message outside of the event loop
```
imiric, line 179: Can this TODO be deleted now?

imiric (dismissed):
I defer to others for the final approval, but this LGTM.

olegbespalov (dismissed):

js/eventloop/eventloop.go
mstoykov:
```suggestion
```

Yeah I don't think this will be possible either way :thinking:

olegbespalov (approved):

imiric (approved):

Issue #2049, opened 2021-06-07
Improve UX of native Go errors


na--:
This error: https://github.com/k6io/k6/blob/da5eebc638d9917cedfe98e232a7eed4cb69a157/js/modules/k6/k6.go#L40-L41
https://github.com/k6io/k6/blob/da5eebc638d9917cedfe98e232a7eed4cb69a157/js/modules/k6/k6.go#L70-L74

And others like it, when returned, produce a stacktrace that might be very confusing to users. Before https://github.com/k6io/k6/pull/1971 and https://github.com/k6io/k6/pull/2046, it used to be something like this:

```
ERRO[0000] engine error                                  hint="error while initializing VU #5: 'GoError: Using group() in the init context is not supported\n\tat go.k6.io/k6/js/common.Bind.func1 (native)\n\tat file:///path/to/script.js:28:19(59)\n'"
```

After them, it will be something like this:
```
ERRO[0000] GoError: Using group() in the init context is not supported
	at go.k6.io/k6/js/common.Bind.func1 (native)
	at file:///path/to/script.js:28:19(59)  hint="error while initializing VU #2 (script exception)"
```

But in both cases, `GoError` and `at go.k6.io/k6/js/common.Bind.func1 (native)` are probably quite confusing to a lot of k6 users. They feature implementation details (like the fact that k6 is written in Go :sweat_smile:) that don't help in the issue diagnosis at all.

https://github.com/k6io/k6/issues/1802 might get rid of the extra stacktrace line, but we might have to do some text trimming in `js.scriptException` if we want to get rid of `GoError`, since I think that's something that goja attaches?

na--:
@MStoykov reminded me that @imiric had actually opened a PR that partially solved this issue, https://github.com/k6io/k6/pull/1775 :sweat_smile: We should probably dust up that PR and merge it (or something similar) in the next k6 release...

na--:
Moved the milestone of the PR to v0.33.0, since it's easy to rebase and port and seems to completely solve the `GoError` part of the issue! :tada: I'd still leave this issue open for the `at go.k6.io/k6/js/common.Bind.func1 (native)` part.

olegbespalov:
Closing since it's no longer relevant:

```
ERRO[0000] GoError: Using group() in the init context is not supported
        at go.k6.io/k6/js/modules/k6.(*K6).Group-fm (native)
        at file:///path/to/script.js:4:6(36)  hint="script exception"
```

PR #2739, opened 2022-10-18
Make common.Throw add stacktraces the same as returning an error

mstoykov:


mstoykov:
This in practice reverts https://github.com/grafana/k6/pull/1775 as it turns out that with this change I can't get the stack trace to print it in the promise rejections.

js/common/util_test.go
```diff
@@ -11,6 +11,7 @@ import (
 
 func TestThrow(t *testing.T) {
 	t.Parallel()
+	t.Skip("Until we remvoe GoError")
```
codebien, line 14:
```suggestion
	t.Skip("Until we remove GoError")
```

Should we link https://github.com/grafana/k6/issues/2049?

codebien (changes requested):

js/common/util_test.go
na--: I don't understand why the test is skipped, but to be honest, I am not sure precisely what it's testing :sweat_smile: Especially the second half, which has a minor bug - `fn1 is invalid` should probably be `fn2 is invalid`

mstoykov: It is testing that the reported error will *not* have `GoError: ` in front

mstoykov: > Should we link https://github.com/grafana/k6/issues/2049?

I am not personally all that big of a fan of back referencing, but if you want to :shrug:

codebien: TBH, now @na-- mentioned it, I think would be better to just refactor and re-enable this test, in the end, it is also testing that the `Throw` function is working as expected and we still have it.

I think we could have two cases:
- one in case of `*goja.Exception`
- and one in the else path

mstoykov: [0503fcb](https://github.com/grafana/k6/pull/2739/commits/0503fcb06d2d80efed3f03bb073fcb4f323ae8f9)

na-- (approved):

codebien (approved):

PR #2743, opened 2022-10-19
Remove byte array from redundant error message

MattDodsonEnglish:
According to the MDN docs, the JavaScript world prefers the term ArrayBuffer and considers "byte array" synonymous.

> [ArrayBuffer] is an array of bytes, often referred to in other languages as a "byte array".
>
> - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer

Having both terms in the same message suggests they are different concepts.

### DIFF
--- a/js/common/util.go
+++ b/js/common/util.go
@@ -32,7 +32,7 @@ func GetReader(data interface{}) (io.Reader, error) {
 	case goja.ArrayBuffer:
 		return bytes.NewBuffer(r.Bytes()), nil
 	default:
-		return nil, fmt.Errorf("invalid type %T, it needs to be a string, byte array or an ArrayBuffer", data)
+		return nil, fmt.Errorf("invalid type %T, it needs to be a string or ArrayBuffer", data)
 	}
 }
 


MattDodsonEnglish:
This PR is also a way of skipping a discussion: if the assumption in this PR is incorrect, I'll probably learn the reason why when it gets closed :-) .

na--:
There was a time when goja (and so, k6) didn't support `ArrayBuffer`, and when "binary data" literally meant a normal JS array with 0-255 integer values :sob: And I think we still support that format in a bunch of places, for backwards compatibility. In the very function you are editing, in fact, you can see `case []byte:` and `case goja.ArrayBuffer:` are 2 different cases.

So, the current error text is unfortunately correct :sweat_smile:

na--:
Not sure :thinking: I don't see a problem with the code in `GetReader()` staying the same, just in case, but we should probably adjust the docs and this error message to discourage these "byte array" monstrosities :thinking:

na-- (approved):

mstoykov (approved):

olegbespalov (approved):

Issue #2724, opened 2022-10-12
Support wildcards in options.hosts


tom-miseur:
### Feature Description

It would be useful if it were possible to use wildcards in `options.hosts` in the same way we can with [blockHostnames](https://k6.io/docs/using-k6/k6-options/reference/#block-hostnames).

So instead of:

```javascript
export const options = {
  hosts: {
    'subdomain1.k6.io': '1.2.3.4',
    'subdomain2.k6.io': '1.2.3.4',
  },
}
```

One could specify any subdomain:

```javascript
export const options = {
  hosts: {
    '*.k6.io': '1.2.3.4',
  },
}
```

eugercek:
May I pick this up?

na--:
@eugercek, sure, though please make sure to add plenty of tests and ideally also some benchmarks.

eugercek:
I have some questions to proceed:
1. I found the Trie implementation from #1532 and used the same approach to handle wildcards. But there's one difference `options.hosts` need to store a (`IP` or `P:Port`), so I created a new type. Since we use, don't use 1.18 can't use generics to create a generic type. Is it ok to introduce a new type? The new type: https://github.com/eugercek/k6/blob/wildcard-host/lib/types/addresstrie.go#L35 
  - There are two types:
    - `matchTrieNode` is a trie node with leaf indication as a value( not empty string).
    - `AddressTrie` wrapper around `matchTrieNode`.

2. Currently, `hosts` doesn't do hostname validation should I add it?

3. What are your opinions on moving to iterative trie implementation rather than the current recursive approach?

4. Because of circular dependency, I can't import `go.k6.io/k6/lib` from the `types` package. Ideally, using `types.NewAddressTrie()` with a `lib.Hostaddress` argument would be better, but not sure how to proceed.
  - Currently, I get it as a string and will handle it in `Dialer`
  - Maybe can use `net.TCPAddr`

na--:
Closed by https://github.com/grafana/k6/pull/2747

tom-miseur:
Thank you very much for your contribution, @eugercek!

PR #2747, opened 2022-10-25
Add Wildcard to `Hosts`

eugercek:
Implements #2724

The current implementation works and passes all the tests. But I'm not sure about some of the parts, we can discuss them.

- Should we check `Hosts` keys and values before execution? Currently, it fails at runtime.
- `matchTrieNode.match` is a mess 😞 I didn't want to merge `if`s because IMO this is cleaner.
- I couldn't use `HostAddress` because of circular dependency, used `net.TCPAddr` if we move `HostAddress` to types, we can use that instead of `net.TCPAddr`
- I couldn't find a better way than using a Trie. Maybe regex checking on all map elements concurrently can be faster, have no clue though 🤔


codebien:
Hi @eugercek,
thanks for the effort in doing this :clap: :bow: 

>Should we check Hosts keys and values before execution? Currently, it fails at runtime.

Which check would you suggest? Can you point to me an example, please?

> I couldn't use HostAddress because of circular dependency, used net.TCPAddr if we move HostAddress to types, we can use that instead of net.TCPAddr

I think it would be a good idea.

>I couldn't find a better way than using a Trie. Maybe regex checking on all map elements concurrently can be faster, have no clue though thinking

As requested in the issue, can you include a [benchmark comparison](https://pkg.go.dev/golang.org/x/perf/cmd/benchstat) between the old and the current version, please?

>Since we don't use Go 1.18 couldn't use generics to abstract trieNode to use it as matchTrieNode

We are building k6 binaries using Go 1.19 so generics are already supported. Anyway, generics aren't the unique way of doing abstraction in Go. Which part of `matchTrieNode` is missing in `trieNode` for re-using it or that is not possible to integrate?

Generally speaking, we should share the code between `BlockedHostnames` and `Hosts` as much as we can.

eugercek:
> thanks for the effort in doing this 👏🏻  🙇🏻 

My pleasure 😊 

> Which check would you suggest? Can you point to me an example, please?

We can check if the key is a valid hostname or hostname:port and if the value is valid IP or IP:port. We do this check in `HostnameTrie` https://github.com/grafana/k6/blob/8a74171c8b43f48fb53078e0bcfa7851828124bc/lib/types/hostnametrie.go#L112-L128

> I think it would be a good idea.

Moving it.

> As requested in the issue, can you include a [benchmark comparison](https://pkg.go.dev/golang.org/x/perf/cmd/benchstat) between the old and the current version, please?

Sorry forgot that. Adding it.

>  Which part of matchTrieNode is missing in trieNode for re-using it or that is not possible to integrate?

I noticed that I could use use  `source` as a hash map to look up key values, I was comparing iterative and recursive trie implementation and totally forget the original idea,😞. I want to provide iterative vs recursive benchmarks too if it's OK.
(The old problem)
In `Hosts` we need to store the value (IP or IP:port) for the matching trie node. I choose to put the value inside the trie node. I thought the best way is to create a generic value type `[Value T]` and use bool in `HostnameTrie` and using string in `AddressTrie`. So the problem was `isLeaf bool` should be `value string`.

eugercek:
Hi @codebien, it's ready for your review :slightly_smiling_face:

## Trie Implementation

I used `trieNode` and iterative implementation. Iterative's `contains()` can't be elegant. The recursive one is much simpler. Though `insert()` is simpler in iterative IMO. Also, I moved the trie implementation to a different file and added unit tests.

## Benchmark

I couldn't run benchmarks in the master because of a bug, which I fixed in [b97e14e](https://github.com/grafana/k6/pull/2747/commits/b97e14ee1f801fa0330fde60894b2a407ee55464), that's the only difference between the real master and the master I used.


Also, maybe we can add a benchmark test to GitHub actions just to make sure it's correct atm. A simple go test -v ./... -run=^$ -bench=. -benchtime=1x` last 1 minute on my machine.

Benchmark command
```sh
git checkout master && go test -v ./... -run=^$ -bench=. -benchmem  > old.txt
git checkout wildcard-host && go test -v ./... -run=^$ -bench=. -benchmem  > new.txt 
benchstat old.txt new.txt > diff.txt
```

Benchmark values  [diff.txt](https://github.com/grafana/k6/files/9881321/diff.txt)

codebien:
Hi @eugercek,
I fixed the benchmark error in https://github.com/grafana/k6/pull/2750. Can you rebase so we can avoid that change here, please?

Regarding the benchmark, I would suggest to not running the entire suite but adding a filter `-bench=BenchmarkTheUsefulBench` for just running the required bench for this PR.

codebien:
It would be also helpful if you could rebase and squash the current commit history and move to something with dedicated commits for:

- Movement for the current files without changes to the new package
- Add the new required changes
- Migration for all the occurrences
- The new additional logic for Hosts

But before doing something like this wait for @mstoykov's review because maybe he would ask for a different general idea.

mstoykov:
Yeah - it will be nicer if at least the movement of the HostAddress is done separately. I personally even think you can open a PR for that on it's own and base this one on top of that. 

On first pass I have a bunch of small questions here and there, but it seems most of them are related to the trie implementation that is in practice the same as before :thinking: 

> Since we don't use Go 1.18 couldn't use generics 

We definitely can bump the go version - we do already only build with 1.18+. But I am not certain that it will make that much of a difference. 

Can you also elaborate on the envconfig issue you are hitting? THere are a couple of comments in the code about it

eugercek:
Hi @mstoykov 

## Trie Implementation
Changes:
- Added unit tests, I think these can be a great resource to learn the implementation.
- Removed `trieNode.insert()`'s `error` returning.

I didn't change any logic in`trieNode` implementation only simplified a function, `trieNode.insert()` never returns an error I changed that. The current implementation is recursive. I implemented an iterative version but removed that later because I don't think `match()` is readable in the iterative version but `insert()` is much simpler in the iterative IMO.

## Generic
Sorry for the generic noise, at first I thought I need to use generics but it's not a must, the current implementation has no problem.


## envconfig

Even though there's no environment variable for the `hosts` option it creates the  `NullAddressTrie.Trie` AFAIU it should not create the `NulAdressTrie.Trie`. That's why I used a "constructor" `NewAddressTrie()` with `Valid` field set to `true` like in `NullHostnameTrie`.

I'll do the following if I understood correctly:
- Create another PR for the changes related to the previous code (added unit test, moving `trie` implementation to a different file, removing the error message). But I'm waiting for your comment on the recursive vs iterative implementation topic.
- Rebase this PR to a more readable set of commits.

Also, I couldn't get which benchmarks I should run, or if should I create new ones. Can you elaborate on this?

codebien:
>- Create another PR for the changes related to the previous code (added unit test, moving trie implementation to a different file, removing the error message).

This would be great, yes.

>But I'm waiting for your comment on the recursive vs iterative implementation topic.

If you change the current implementation, I suggest two things:
- Create benchmarks for `insert` and `contains` methods.
- Apply the changes only after and in a separate commit so we can clearly compare the two versions.

So, ideally, I expect 3 commits:
- only the movements
- benchmarks addition
- changes from recursive to iterative

>Also, I couldn't get which benchmarks I should run, or if should I create new ones. Can you elaborate on this?

The original benchmark request is for creating a comparison between the current way (`Hosts map[string]*HostAddress`) with the new one based on `types.NullAddressTrie`. In this way, we can identify if we are introducing relevant performance degradation.

eugercek:
Hi @codebien 

I listed what I'll do, I'll update this message when I progress

- I'll create the PR and remove its feature from this PR
  - Since you're OK with the review (with benchmark) I wanna think a little more about the iterative trie implementation on the new PR
  - Will add benchmark for the `contains` and `insert`
 - Adding the latter benchmark on this PR and squashing the commits the form you wanted

eugercek:
I'm rebasing this on master and squashing commits to a readable way. 

Also added Hosts benchmark.

eugercek:
Benchmark result is sad 😔 

```
name           old time/op    new time/op    delta
DialerHosts-8     211ns ± 1%     213ns ± 1%  +0.71%  (p=0.000 n=30+30)

name           old alloc/op   new alloc/op   delta
DialerHosts-8      160B ± 0%      160B ± 0%    ~     (all equal)

name           old allocs/op  new allocs/op  delta
DialerHosts-8      5.00 ± 0%      5.00 ± 0%    ~     (all equal)

```

Also you can see https://github.com/grafana/k6/pull/2747#issuecomment-1293788622 for all benchmarks

Benchmark script

```
bench(){
  echo "started: $2 benchmark"
  git checkout "$1"  2> /dev/null
  go test -run=^$ -bench=BenchmarkDialerHosts -benchmem -count 30 > "$2"
}

cd lib/netext

bench 51a4d001 "original"
bench cd82d0d9 "addresstrie"

benchstat original addresstrie > "bs_addresstrie"
```

lib/types/types.go
```diff
@@ -212,3 +213,81 @@ func GetDurationValue(v interface{}) (time.Duration, error) {
 		return time.Duration(n) * time.Millisecond, nil
 	}
 }
+
+// HostAddress stores information about IP and port
```
oleiade, line 217: As we're at it, I recommend moving this type to a dedicated `net.go` or `ip.go` file instead. What do you think?

lib/types/types.go
```diff
@@ -212,3 +213,81 @@ func GetDurationValue(v interface{}) (time.Duration, error) {
...
+}
+
+// UnmarshalText implements the encoding.TextUnmarshaler interface.
+// The IP address is expected in a form accepted by ParseIP.
+func (h *HostAddress) UnmarshalText(text []byte) error {
+	if len(text) == 0 {
+		return &net.ParseError{Type: "IP address", Text: "<nil>"}
+	}
+
+	ip, port, err := splitHostPort(text)
```
oleiade, line 264: Unless the dedicated function exists for a specific reason, I would recommend using the existing standard `[net.SplitHostPort]](https://pkg.go.dev/net#SplitHostPort)` function instead 👍🏻

lib/types/addresstrie.go
```diff
@@ -0,0 +1,120 @@
...
+		}
+	}
+
+	n.Trie = NewAddressTrie(source)
+	n.Valid = true
+	return nil
+}
+
+// AddressTrie is wrapper around trieNode to integrate with net.TCPAddr
+type AddressTrie struct {
```
oleiade, line 79: Same comment as on the source `HostAddress` type. I would prefer to use the `Host` nomenclature instead of `Address`. Same comment for the overarching `NullAddressTrie` 🙇🏻

lib/types/addresstrie.go
```diff
@@ -0,0 +1,120 @@
...
+	n      *trieNode
```
oleiade, line 80: Maybe not worth doing in this PR, but as we use the trie in multiple places now, it would be convenient to move its implementation to a dedicated `trie.go` file.

lib/types/addresstrie.go
```diff
@@ -0,0 +1,120 @@
...
+		}
+		jsonMap[k] = right
+	}
+
+	return json.Marshal(jsonMap)
+}
+
+// UnmarshalJSON converts JSON to NullAddressTrie
+func (n *NullAddressTrie) UnmarshalJSON(data []byte) error {
+	if bytes.Equal(data, []byte("null")) {
```
oleiade, line 47: I would prefer it if we avoided using static strings as such in the code. They can be confusing as one would hardly know if they're supposed to be a "standardized" value or an arbitrary one. In this case, it would be worth declaring a constant string `null` or something along those lines.

lib/types/addresstrie_test.go
```diff
@@ -0,0 +1,144 @@
...
+		"simple-ipv6.io:443":          {IP: net.ParseIP("aa::bb"), Port: 8443},
+		"with-port-ipv6.io:80":        {IP: net.ParseIP("cc::dd"), Port: 80},
+		"with-port-ipv6.io:443":       {IP: net.ParseIP("cc::dd"), Port: 8443},
+		"*.wildcard-ipv6.io":          {IP: net.ParseIP("ee::ff")},
+		"specific.wildcard-ipv6.io":   {IP: net.ParseIP("ee:11::ff")},
+		"*wildcard-2-ipv6.io":         {IP: net.ParseIP("aa::aa")},
+		"specific.wildcard-2-ipv6.io": {IP: net.ParseIP("a1::a1")},
+	})
+
+	tcs := []struct {
```
oleiade, line 37: I found this test table very hard to process and comprehend. The reason for it is that it depends on this big mapping that was set beforehand. Cognitively I can't keep all of it loaded in my head as I read the code, which means that later on, as a maintainer it would take me a lot of time to chase an issue using this test suite, or modify it.

I would recommend to either add a `name`, or `description` field to the test table, in which, for each case, the actual thing that is tested and its expected outcome are described. _e.g_ "using a wildcard over a domain with two subdomains defined should succeed with 2 results"

Or split and regroup the different categories of scenarios into dedicated sub-tests.

What do you think? 🙇🏻

lib/types/addresstrie_test.go
```diff
@@ -0,0 +1,144 @@
...
+}
+
+func TestNullAddressTrieJSON(t *testing.T) {
+	t.Parallel()
+
+	tcs := []struct {
+		t       NullAddressTrie
+		marshal string
+	}{
+		{t: NullAddressTrie{}, marshal: "null"},
```
oleiade, line 97: Same comment on static strings as before 👍🏻

lib/types/addresstrie_test.go
```diff
@@ -0,0 +1,144 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestAddressTrie(t *testing.T) {
+	t.Parallel()
+
+	at := NewAddressTrie(map[string]HostAddress{
```
oleiade, line 15: Quick uninformed question, do we want to support compound wildcards `*.hello.*`. Does it even make sense? I'm not sure. If this is nonsensical or has not added value, please disregard this 👍🏻

oleiade:
A question from my side, as I just joined the review. I'm a bit unclear as to what we're benchmarking and what is the underlying objective?

eugercek:
Hi @oleiade thanks for your detailed review and compliment 🙏🏻 

On the benchmark @codebien requested it. I don't have any idea how the information is used  😅 But we're benchmarking previous `Hosts` implementation and current one. Previous one was O(1) hashmap lookup, current was is O(n) for n length hostname.

lib/types/types.go
```diff
@@ -212,3 +213,81 @@ func GetDurationValue(v interface{}) (time.Duration, error) {
 		return time.Duration(n) * time.Millisecond, nil
 	}
 }
+
+// HostAddress stores information about IP and port
+// for a host.
+type HostAddress net.TCPAddr
```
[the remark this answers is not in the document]
eugercek: Totally agree 👍🏻 I thought it was codebase naming convention and didn't touch it. Changing that to `Host`as you wish

lib/types/types.go
eugercek: `net.ip` sounds good to me, but I have no knowledge on the codebase beside this issue 🙇🏻 Your opinions are more important for me, which one we should choose?

lib/types/types.go
```diff
@@ -212,3 +213,81 @@ func GetDurationValue(v interface{}) (time.Duration, error) {
...
+
+// NewHostAddress creates a pointer to a new address with an IP object.
+func NewHostAddress(ip net.IP, portString string) (*HostAddress, error) {
```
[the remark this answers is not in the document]
eugercek: Totally agree 👍🏻 changing it.

oleiade:
Hey @eugercek 

@codebien brought me up to speed. From my perspective, and I think @codebien agrees, judging by the numbers the benchmark exhibits, we're good to go with the new trie API that your PR introduces. We judged a 1% performance difference acceptable if we're trading it for convenience and maintainability 👍🏻

lib/types/addresstrie.go
eugercek: Actually we did it in #2777 🙇🏻 Have you seen that, I rebased this PR on #2747

lib/types/addresstrie.go
eugercek: Purging the `Address` suffix everywhere 🔫

lib/types/types.go
```diff
@@ -212,3 +213,81 @@ func GetDurationValue(v interface{}) (time.Duration, error) {
...
+	var port int
+	if portString != "" {
+		var err error
+		if port, err = strconv.Atoi(portString); err != nil {
+			return nil, err
+		}
+	}
+
+	return &HostAddress{
+		IP:   ip,
```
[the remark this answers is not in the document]
eugercek: It was not mine 🙇🏻, I only moved this around because of circular dependency. But I totally agree again, changing it as you requested.

lib/types/addresstrie.go
```diff
@@ -0,0 +1,120 @@
...
+	for k := range at.source {
+		at.insert(k)
+	}
+
+	return at
+}
+
+func (t *AddressTrie) insert(s string) {
+	s = strings.ToLower(s) // domains are not case-sensitive
+	t.n.insert(s)
```
imiric, line 102: We should probably validate that `s` is a valid hostname pattern, no? Like we do for [`HostnameTrie.insert()`](https://github.com/grafana/k6/blob/28c06296e5f6e92544500a5b69d80f32db0dd9c9/lib/types/hostnametrie.go#L113-L120).

lib/types/addresstrie.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+// Match returns matched address can return three types of value
```
imiric, line 105:
```suggestion
// Match returns the host matching s, where the value can be one of:
```

imiric (changes requested):
Great work again @eugercek! :clap: Everything is very comprehensible. :+1:

I just left some minor comments, but they're not really blockers.

Apologies for the delay and back and forth with reviewing this. We definitely would like to make it part of the upcoming v0.42.0 release, if you have time to wrap up the minor changes we raised here by early next week. :pray:

lib/types/addresstrie_test.go
imiric: Ah, one use case I think a second asterisk as suffix would be convenient is for ports. You can already specify a port along with the hostname, e.g. `'test.k6.io:443': '1.2.3.4:8443'`. So maybe someone would find it useful to override only specific ports, like `'*.k6.io:24*': '1.2.3.4:8443'`. But I think this would be very niche and complicate the implementation, so let's defer to YAGNI for the second asterisk.

eugercek:
I'll push your requested changes today. Thanks for your reviews.

lib/types/addresstrie.go
eugercek: I agree with the `Hosts` over other names we mentioned.

lib/types/types.go
eugercek: Yep it does the below.

```go
host, port, err := net.SplitHostPort(string(text))
if err != nil {
	// This error means that there is no port.
	// Make host the full text.
	host = string(text)
}
```

`net.SplitHostPort` returns `missing port in address` error when no port is given, but we accept `domain` and `domain:port` forms.

lib/types/addresstrie.go
eugercek: Also, I think, `BlockHostnames` is better than `Hostnames` which is very similar to `Hosts` also it's akin to the [k6 option name](https://k6.io/docs/using-k6/k6-options/reference/#block-hostnames). I can change that if you wish.

lib/types/addresstrie_test.go
eugercek: Done

lib/types/addresstrie_test.go
eugercek: I agree with you, every case had a reason but I can not understand their reason at a first glance, it's a poorly written test 😞 . But I remember why I put it like this, it's very hard to categorize this _perfectly_ and that's why I passed that in the first place. Refactoring this 🤓 
-

lib/types/addresstrie_test.go
eugercek: IIRC in #1532, only one wildcard is chosen. that's why I implemented it like that. YAGNI is my best friend 😆

lib/types/addresstrie.go
eugercek: Make sense, but we can not use the previous regex, since this one also has ports. Simple `(:[0-9]+)?` append would be enough.

lib/types/addresstrie.go
eugercek: Sorry for my bad engrish 😞, fixed

lib/types/types.go
eugercek: Moved to `net.go` just, can move later if you wish.

eugercek:
I added all resolved and unresolved changes, waiting for your review  🙇🏻

One criticism from me, new names `Hosts` (previosuly `HostAddressTrie`) and `Host` (`HostAdress`) can be confusing.

lib/types/types.go
oleiade: I see! Indeed, and it doesn't return a dedicated error value either 😞 Makes sense then, thanks a lot for the clarification 🙇🏻

lib/types/types.go
oleiade: Perfect, I would have prefered `net.go` indeed 👍🏻 🙇🏻

lib/types/addresstrie.go
oleiade: My bad then, thanks for the clarification 👍🏻 🙇🏻

lib/types/addresstrie_test.go
oleiade: Fair enough, thanks a lot for the pointers folks. If everyone's okay with it, let's disregard my question indeed, and reevaluate it a later point if we do end up needing it then 👍🏻 🙇🏻

oleiade:
@eugercek Thanks for your work refactoring this 👍🏻 Regarding the naming, while we acknowledge that it might turn out confusing, I believe @imiric and I will turn out rather aligned on this and would prefer to proceed with the `Hosts` and `Host` naming for the reason we mentioned previously 😄

lib/types/addresstrie_test.go
oleiade: Thanks for doing this 🙇🏻

lib/types/addresstrie.go
imiric: I think the plural is enough of a distinguishing feature between `Host` and `Hosts`, so I'm OK with it.

I wouldn't rename `HostnameTrie` to `BlockHostnames` since there's nothing "blocking" about it. It could just as well be used as an allow list, so `Hostnames` is a generic term for the container. But I wouldn't rename it in this PR, let's leave it as `HostnameTrie` for now.

lib/types/addresstrie.go
imiric: Ah, cheers for that, the new regex looks good :+1:

imiric (dismissed):
Thanks for the recent changes and your work on this PR @eugercek :bow: Outstanding work! :clap:

Just noticed some tiny nitpicks, but no real blockers from me.

lib/types/addresstrie_test.go
eugercek: My pleasure ❤️

lib/types/hosts_test.go
```diff
@@ -0,0 +1,248 @@
...
+)
+
+type HostTestCase struct {
+	hostname, expVal, desc string
+}
+
+func TestHosts(t *testing.T) {
+	t.Parallel()
+
+	at, err := NewHosts(map[string]Host{
```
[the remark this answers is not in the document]
eugercek: Done

lib/types/hosts.go
```diff
@@ -0,0 +1,157 @@
...
+	return nil
+}
+
+// Hosts is wrapper around trieNode to integrate with net.TCPAddr
+type Hosts struct {
+	n      *trieNode
+	source map[string]Host
+}
+
+// NewHosts returns new Address Trie from given addresses
```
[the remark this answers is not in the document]
eugercek: Done

oleiade (approved):
Thanks a lot for your work great work, much appreciated 🙇🏻

imiric (approved):
Thank you again! :bow: :shipit:

PR #2754, opened 2022-11-02
add extension version

HarrisChu:
I write a k6 extension and build with xk6, but I cannot know the extension version via command line.

if extension module implement `VersionGetter` interface, could display via version command.

```
package k6plugin

import (
	"github.com/vesoft-inc/k6-plugin/pkg/aggcsv"
	"github.com/vesoft-inc/k6-plugin/pkg/nebulagraph"
	"go.k6.io/k6/js/modules"
	"go.k6.io/k6/output"
)

func init() {
	modules.Register("k6/x/nebulagraph", nebulagraph.NewNebulaGraph())
	output.RegisterExtension("aggcsv", func(p output.Params) (output.Output, error) {
		return aggcsv.New(p)
	})
}
```

```
> ./k6 version
k6 v0.40.0 ((devel), go1.18.5, linux/amd64)
extension k6/x/nebulagraph v1.0.2
```

### DIFF
--- a/cmd/ui.go
+++ b/cmd/ui.go
@@ -17,6 +17,7 @@ import (
 
 	"gopkg.in/yaml.v3"
 
+	"go.k6.io/k6/js/modules"
 	"go.k6.io/k6/lib"
 	"go.k6.io/k6/lib/consts"
 	"go.k6.io/k6/output"
@@ -81,7 +82,13 @@ func getColor(noColor bool, attributes ...color.Attribute) *color.Color {
 
 func getBanner(noColor bool) string {
 	c := getColor(noColor, color.FgCyan)
-	return c.Sprint(consts.Banner())
+	banner := make([]string, 0, len(modules.GetJSModuleVersions())+1)
+
+	banner = append(banner, consts.Banner())
+	for pkg, version := range modules.GetJSModuleVersions() {
+		banner = append(banner, fmt.Sprintf("extension  %s %s", pkg, version))
+	}
+	return c.Sprint(strings.Join(banner, "\n"))
 }
 
 func printBanner(gs *globalState) {
--- a/cmd/version.go
+++ b/cmd/version.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/spf13/cobra"
 
+	"go.k6.io/k6/js/modules"
 	"go.k6.io/k6/lib/consts"
 )
 
@@ -16,6 +17,9 @@ func getCmdVersion(globalState *globalState) *cobra.Command {
 		Long:  `Show the application version and exit.`,
 		Run: func(_ *cobra.Command, _ []string) {
 			printToStdout(globalState, fmt.Sprintf("k6 v%s\n", consts.FullVersion()))
+			for path, version := range modules.GetJSModuleVersions() {
+				printToStdout(globalState, fmt.Sprintf("extension %s %s\n", path, version))
+			}
 		},
 	}
 }
--- a/js/modules/modules.go
+++ b/js/modules/modules.go
@@ -3,6 +3,8 @@ package modules
 import (
 	"context"
 	"fmt"
+	"reflect"
+	"runtime/debug"
 	"strings"
 	"sync"
 
@@ -15,8 +17,9 @@ const extPrefix string = "k6/x/"
 
 //nolint:gochecknoglobals
 var (
-	modules = make(map[string]interface{})
-	mx      sync.RWMutex
+	modules        = make(map[string]interface{})
+	moduleVersions = make(map[string]string)
+	mx             sync.RWMutex
 )
 
 // Register the given mod as an external JavaScript module that can be imported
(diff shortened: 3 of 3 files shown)


codebien:
Hey @HarrisChu,
thanks for your contribution :bow:

Please, check [this previous comment](https://github.com/grafana/k6/pull/2686#pullrequestreview-1113579243) where we shared our vision regarding this feature.

HarrisChu:
> 

perfect for me.
so I should close this pr?

codebien:
If you would like to contribute then you can keep this open and address it as a request change.

HarrisChu:
I cannot test the code as `debug.ReadBuildInfo` not populated for test
just test it with my extension.

image

cmd/version.go
```diff
@@ -16,6 +17,9 @@ func getCmdVersion(globalState *globalState) *cobra.Command {
 		Long:  `Show the application version and exit.`,
 		Run: func(_ *cobra.Command, _ []string) {
 			printToStdout(globalState, fmt.Sprintf("k6 v%s\n", consts.FullVersion()))
+			for path, version := range modules.GetJSModuleVersions() {
+				printToStdout(globalState, fmt.Sprintf("extension %s %s\n", path, version))
+			}
```
imiric, line 22: k6 supports both JS and output extensions. So we should make sure that both are shown in this output (same goes for the `k6 run` output).

You can test by building a binary with an output extension, e.g.:
```
XK6_K6_REPO=github.com/HarrisChu/k6 xk6 build master \
  --with github.com/vesoft-inc/k6-plugin \
  --with github.com/grafana/xk6-output-influxdb
```

It should be a matter of doing something similar as in `js/modules/modules.go` and calling `getPackageVersion(mod)` from [`output.RegisterExtension()`](https://github.com/grafana/k6/blob/e14f14d416b0af4e495e6f565a2a9787a73ade54/output/extensions.go#L27). You'll probably want to move `getPackageVersion()` somewhere where it's accessible by both (maybe `lib/extensions.go`?).

js/modules/modules.go
```diff
@@ -15,8 +17,9 @@ const extPrefix string = "k6/x/"
 
 //nolint:gochecknoglobals
 var (
-	modules = make(map[string]interface{})
-	mx      sync.RWMutex
+	modules        = make(map[string]interface{})
+	moduleVersions = make(map[string]string)
```
imiric, line 21: Would it make sense to reuse the `modules` map, and have its values be a small struct like:

```go
type module struct {
	mod     interface{}
	version string
}
```

This way `getPackageVersion()` would only return the version, as its name implies, and not also set (another) global map value.

This will complicate things slightly, as you'll need to change wherever `GetJSModules()` is called (or change `GetJSModules()` itself to continue to return just the modules without the version; though I think the first approach is slightly better), but it should avoid the repetition, and be a bit cleaner in the end. It would also avoid the need for a separate `GetJSModuleVersions()` function.

js/modules/modules.go
```diff
@@ -34,6 +37,38 @@ func Register(name string, mod interface{}) {
 		panic(fmt.Sprintf("module already registered: %s", name))
 	}
 	modules[name] = mod
+	getPackageVersion(mod)
+}
+
+func getPackageVersion(mod interface{}) {
```
imiric, line 43:
```suggestion
func getModuleVersion(mod interface{}) {
```

I think this is slightly clearer, as we are getting the Go _module_ version after all (and the map is called `moduleVersions`).

imiric (changes requested):
Thanks for this PR @HarrisChu! :bow:

We've had #1741 in our backlog for a long time, so it's great that it's finally being implemented. As you've found, it's an important feature for k6 extension maintainers.

I do have a couple of requirements before merging this. Namely making the `k6 run` output more pleasing, and also showing the versions of any output extensions. The others are suggestions that would be nice to have, but not blockers.

Since testing this is a bit tricky, we might want to have an E2E test instead. Currently we have an [`xk6` CI workflow](https://github.com/grafana/k6/blob/e14f14d416b0af4e495e6f565a2a9787a73ade54/.github/workflows/xk6.yml), so it might make sense to add an assertion or two about the actual output of both `k6 run` and `k6 version` there. Unless someone has a better idea to test this...

cmd/ui.go
```diff
@@ -81,7 +82,13 @@ func getColor(noColor bool, attributes ...color.Attribute) *color.Color {
...
 func getBanner(noColor bool) string {
 	c := getColor(noColor, color.FgCyan)
-	return c.Sprint(consts.Banner())
+	banner := make([]string, 0, len(modules.GetJSModuleVersions())+1)
+
+	banner = append(banner, consts.Banner())
+	for pkg, version := range modules.GetJSModuleVersions() {
+		banner = append(banner, fmt.Sprintf("extension  %s %s", pkg, version))
+	}
+	return c.Sprint(strings.Join(banner, "\n"))
```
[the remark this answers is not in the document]
HarrisChu: ok

js/modules/modules.go
HarrisChu: looks good

js/modules/modules.go
HarrisChu: good catch.

btw, If a package has two modules, it just has one package in map.
how about display the module name? e.g.

```
 extensions: github.com/dgzlopes/k6-extension-notifications k6/x/ext1 v1.0.0
                     github.com/dgzlopes/k6-extension-notifications k6/x/ext2 v1.0.0
```

imiric:
Hi @HarrisChu, are you available to finish the work on the pending items? We'd like to release this in the upcoming v0.42.0, so this should be merged in the next week or two.

If not, that's OK as well, and someone from our side would pick it up. You'd still get credit in the release notes, of course :slightly_smiling_face:

oleiade (changes requested):
I align with the comments and recommendations of @imiric 🙏🏻 
Really good work though, looking forward to see those comments addressed 👍🏻

imiric:
Closing this since it was superseded by #2805. Thanks again for your contribution @HarrisChu! :bow:

HarrisChu:
> 

sorry, I had some emergencies...
glad it's already in v0.42, I would update our plugins.

imiric:
@HarrisChu No problem.

It didn't make it in v0.42.0, but it will be released in v0.43.0 a few weeks from now.

Issue #2755, opened 2022-11-02
Thresholds over Trend metrics fail as the `med` sink's value remains zero 


oleiade:
### Brief summary

As reported by a user on the [OSS support forum](https://community.k6.io/t/bug-with-k6-threshold-validation/5038), there seems to be an issue with Thresholds applying to the median value of Trend metrics. Both core and custom metrics are affected.

Setting a threshold on a Trend metric's value, such as `http_req_duration`, will always fail, regardless of the expression used.

### k6 version

>= v0.40.0

### OS

Docker + Linux

### Docker version and image (if applicable)

loadimpact/k6:latest

### Steps to reproduce the problem

Here is an example script demonstrating the issue. It depends on having [Lhotse](https://github.com/oleiade/lhotse) running locally to simulate the desired performance of a target HTTP server. 

```javascript
import { check } from "k6";
import http from "k6/http";
import { Trend } from "k6/metrics";

export let options = {
  thresholds: {
    http_req_duration: ["med>39"],
    custom_trend: ["med>39"],
  },

  scenarios: {
    issueHappensWithCoreTrendMetric: {
      executor: "constant-arrival-rate",
      exec: "exampleFunction",
      rate: 1,
      timeUnit: "30s",
      duration: "30s",
      preAllocatedVUs: 2,
      exec: "issueHappensWithCoreTrendMetric",
    },
    issueHappensWithCustomTrendMetric: {
      executor: "constant-arrival-rate",
      exec: "exampleFunction",
      rate: 1,
      timeUnit: "30s",
      duration: "30s",
      preAllocatedVUs: 2,
      exec: "issueHappensWithCustomTrendMetric",
    },
  },
};

export function issueHappensWithCoreTrendMetric() {
  // Server produces requests that will take between 70, and 80ms
  // to complete. Hence, the resulting http_req_duration metric
  // should have a median value well above the 39ms threshold.
  const res = http.get("http://localhost:3434/latency/70ms-80ms", {}, {});

  check(res, {
    "is status 200": (r) => r.status === 200,
  });
}

const customTrend = new Trend("custom_trend");

export function issueHappensWithCustomTrendMetric() {
  // Add values to the custom trend metric so that its median
  // value is well above the 39ms threshold.
  customTrend.add(70);
  customTrend.add(80);
  customTrend.add(90);
}
```

[...]

Issue #2751, opened 2022-10-31
Run benchmarks on master


codebien:
#2750 fixed a benchmark error. It could prevent similar issues if we run the entire benchmark suite when we merge on `master` a new feature.

The suggestion is to add a new check that runs all benchmarks. Additionally, with minor priority, it would be useful if we could have a generated diff and catch relevant performance changes.

PR #2759, opened 2022-11-08
Fix 2755 and ensure evaulation of thresholds over trend metrics median 

oleiade:
This PR addresses #2755. 

Turns out the Trend Sink median value was never calculated in time for thresholds to be able to evaluate them. The reason for that is that the median value can be quite expensive to compute, as it relies most notably on sorting the sink's array of values. As a result, its calculation happens in a dedicated `Calc` method. It appears we forgot to call that method in the `Thresholds.Run` method during our latest refactoring of the thresholds system.

This PR ensures the `Calc` method is called just in time, so that the costly computation only happens when Thresholds are effectively evaluated. It also adds a couple of tests to protect us from regressions in the future. Finally, it adds some documentation and comments to the code, so that next time we touch that part of the code, there is some background information available to the contributor.

Thanks a lot to @imiric for his collaboration in tracking and getting this issue resolved 🙇🏻 🎉 

Let me know what you think 👍🏻

### DIFF
--- a/metrics/sink.go
+++ b/metrics/sink.go
@@ -74,7 +74,6 @@ type TrendSink struct {
 	Count    uint64
 	Min, Max float64
 	Sum, Avg float64
-	Med      float64
 }
 
 // IsEmpty indicates whether the TrendSink is empty.
--- a/metrics/sink.go
+++ b/metrics/sink.go
@@ -17,7 +17,6 @@ var (
 
 type Sink interface {
 	Add(s Sample)                              // Add a sample to the sink.
-	Calc()                                     // Make final calculations.
 	Format(t time.Duration) map[string]float64 // Data for thresholds.
 	IsEmpty() bool                             // Check if the Sink is empty.
 }
@@ -36,8 +35,6 @@ func (c *CounterSink) Add(s Sample) {
 
 // IsEmpty indicates whether the CounterSink is empty.
 func (c *CounterSink) IsEmpty() bool { return c.First.IsZero() }
-
-func (c *CounterSink) Calc() {}
 
 func (c *CounterSink) Format(t time.Duration) map[string]float64 {
 	return map[string]float64{
@@ -66,15 +63,13 @@ func (g *GaugeSink) Add(s Sample) {
 	}
 }
 
-func (g *GaugeSink) Calc() {}
-
 func (g *GaugeSink) Format(t time.Duration) map[string]float64 {
 	return map[string]float64{"value": g.Value}
 }
 
 type TrendSink struct {
-	Values  []float64
-	jumbled bool
+	Values []float64
+	sorted bool
 
 	Count    uint64
 	Min, Max float64
@@ -87,7 +82,7 @@ func (t *TrendSink) IsEmpty() bool { return t.Count == 0 }
 
 func (t *TrendSink) Add(s Sample) {
 	t.Values = append(t.Values, s.Value)
-	t.jumbled = true
+	t.sorted = false
 	t.Count += 1
 	t.Sum += s.Value
 	t.Avg = t.Sum / float64(t.Count)
@@ -108,10 +103,14 @@ func (t *TrendSink) P(pct float64) float64 {
 	case 1:
 		return t.Values[0]
 	default:
+		if !t.sorted {
+			sort.Float64s(t.Values)
+			t.sorted = true
+		}
+
 		// If percentile falls on a value in Values slice, we return that value.
 		// If percentile does not fall on a value in Values slice, we calculate (linear interpolation)
 		// the value that would fall at percentile, given the values above and below that percentile.
-		t.Calc()
 		i := pct * (float64(t.Count) - 1.0)
 		j := t.Values[int(math.Floor(i))]
 		k := t.Values[int(math.Ceil(i))]
@@ -120,30 +119,13 @@ func (t *TrendSink) P(pct float64) float64 {
 	}
 }
 
-func (t *TrendSink) Calc() {
-	if !t.jumbled {
-		return
-	}
-
-	sort.Float64s(t.Values)
-	t.jumbled = false
-
-	// The median of an even number of values is the average of the middle two.
-	if (t.Count & 0x01) == 0 {
-		t.Med = (t.Values[(t.Count/2)-1] + t.Values[(t.Count/2)]) / 2
-	} else {
-		t.Med = t.Values[t.Count/2]
-	}
-}
-
 func (t *TrendSink) Format(tt time.Duration) map[string]float64 {
-	t.Calc()
 	// TODO: respect the summaryTrendStats for REST API
 	return map[string]float64{
 		"min":   t.Min,
 		"max":   t.Max,
 		"avg":   t.Avg,
-		"med":   t.Med,
+		"med":   t.P(0.5),
 		"p(90)": t.P(0.90),
 		"p(95)": t.P(0.95),
 	}
@@ -164,8 +146,6 @@ func (r *RateSink) Add(s Sample) {
 	}
 }
 
-func (r RateSink) Calc() {}
-
 func (r RateSink) Format(t time.Duration) map[string]float64 {
 	var rate float64
 	if r.Total > 0 {
@@ -183,8 +163,6 @@ func (d DummySink) IsEmpty() bool { return len(d) == 0 }
 func (d DummySink) Add(s Sample) {
 	panic(errors.New("you can't add samples to a dummy sink"))
 }
-
-func (d DummySink) Calc() {}
 
 func (d DummySink) Format(t time.Duration) map[string]float64 {
 	return map[string]float64(d)
--- a/metrics/sink_test.go
+++ b/metrics/sink_test.go
@@ -28,12 +28,6 @@ func TestCounterSink(t *testing.T) {
 			assert.Equal(t, now, sink.First)
 		})
 	})
-	t.Run("calc", func(t *testing.T) {
-		sink := CounterSink{}
-		sink.Calc()
-		assert.Equal(t, 0.0, sink.Value)
-		assert.Equal(t, time.Time{}, sink.First)
-	})
 	t.Run("format", func(t *testing.T) {
 		sink := CounterSink{}
 		for _, s := range samples10 {
@@ -66,14 +60,6 @@ func TestGaugeSink(t *testing.T) {
 			assert.Equal(t, 10.0, sink.Max)
 		})
 	})
-	t.Run("calc", func(t *testing.T) {
-		sink := GaugeSink{}
-		sink.Calc()
-		assert.Equal(t, 0.0, sink.Value)
-		assert.Equal(t, 0.0, sink.Min)
-		assert.Equal(t, false, sink.minSet)
-		assert.Equal(t, 0.0, sink.Max)
-	})
 	t.Run("format", func(t *testing.T) {
 		sink := GaugeSink{}
 		for _, s := range samples6 {
@@ -84,7 +70,6 @@ func TestGaugeSink(t *testing.T) {
 }
 
 func TestTrendSink(t *testing.T) {
-	unsortedSamples5 := []float64{0.0, 5.0, 10.0, 3.0, 1.0}
 	unsortedSamples10 := []float64{0.0, 100.0, 30.0, 80.0, 70.0, 60.0, 50.0, 40.0, 90.0, 20.0}
 
 	t.Run("add", func(t *testing.T) {
@@ -92,11 +77,10 @@ func TestTrendSink(t *testing.T) {
 			sink := TrendSink{}
 			sink.Add(Sample{TimeSeries: TimeSeries{Metric: &Metric{}}, Value: 7.0})
 			assert.Equal(t, uint64(1), sink.Count)
-			assert.Equal(t, true, sink.jumbled)
+			assert.Equal(t, false, sink.sorted)
 			assert.Equal(t, 7.0, sink.Min)
 			assert.Equal(t, 7.0, sink.Max)
 			assert.Equal(t, 7.0, sink.Avg)
-			assert.Equal(t, 0.0, sink.Med) // calculated in Calc()
 		})
 		t.Run("values", func(t *testing.T) {
 			sink := TrendSink{}
@@ -104,40 +88,7 @@ func TestTrendSink(t *testing.T) {
 				sink.Add(Sample{TimeSeries: TimeSeries{Metric: &Metric{}}, Value: s})
 			}
 			assert.Equal(t, uint64(len(unsortedSamples10)), sink.Count)
-			assert.Equal(t, true, sink.jumbled)
-			assert.Equal(t, 0.0, sink.Min)
-			assert.Equal(t, 100.0, sink.Max)
-			assert.Equal(t, 54.0, sink.Avg)
-			assert.Equal(t, 0.0, sink.Med) // calculated in Calc()
-		})
-	})
-	t.Run("calc", func(t *testing.T) {
-		t.Run("no values", func(t *testing.T) {
-			sink := TrendSink{}
-			sink.Calc()
-			assert.Equal(t, uint64(0), sink.Count)
-			assert.Equal(t, false, sink.jumbled)
-			assert.Equal(t, 0.0, sink.Med)
-		})
-		t.Run("odd number of samples median", func(t *testing.T) {
-			sink := TrendSink{}
-			for _, s := range unsortedSamples5 {
-				sink.Add(Sample{TimeSeries: TimeSeries{Metric: &Metric{}}, Value: s})
-			}
-			sink.Calc()
-			assert.Equal(t, uint64(len(unsortedSamples5)), sink.Count)
-			assert.Equal(t, false, sink.jumbled)
-			assert.Equal(t, 3.0, sink.Med)
-		})
-		t.Run("sorted", func(t *testing.T) {
-			sink := TrendSink{}
-			for _, s := range unsortedSamples10 {
-				sink.Add(Sample{TimeSeries: TimeSeries{Metric: &Metric{}}, Value: s})
-			}
-			sink.Calc()
-			assert.Equal(t, uint64(len(unsortedSamples10)), sink.Count)
-			assert.Equal(t, false, sink.jumbled)
-			assert.Equal(t, 55.0, sink.Med)
+			assert.Equal(t, false, sink.sorted)
 			assert.Equal(t, 0.0, sink.Min)
 			assert.Equal(t, 100.0, sink.Max)
 			assert.Equal(t, 54.0, sink.Avg)
@@ -163,11 +114,13 @@ func TestTrendSink(t *testing.T) {
 			sink := TrendSink{}
 			sink.Add(Sample{TimeSeries: TimeSeries{Metric: &Metric{}}, Value: 5.0})
 			sink.Add(Sample{TimeSeries: TimeSeries{Metric: &Metric{}}, Value: 10.0})
+			assert.Equal(t, false, sink.sorted)
 			assert.Equal(t, 5.0, sink.P(0.0))
 			assert.Equal(t, 7.5, sink.P(0.5))
 			assert.Equal(t, 5+(10-5)*0.95, sink.P(0.95))
 			assert.Equal(t, 5+(10-5)*0.99, sink.P(0.99))
 			assert.Equal(t, 10.0, sink.P(1.0))
+			assert.Equal(t, true, sink.sorted)
 		})
 		t.Run("more than 2", func(t *testing.T) {
 			sink := TrendSink{}
@@ -179,6 +132,7 @@ func TestTrendSink(t *testing.T) {
 			assert.InDelta(t, 95.5, sink.P(0.95), tolerance)
 			assert.InDelta(t, 99.1, sink.P(0.99), tolerance)
 			assert.InDelta(t, 100.0, sink.P(1.0), tolerance)
+			assert.Equal(t, true, sink.sorted)
 		})
 	})
 	t.Run("format", func(t *testing.T) {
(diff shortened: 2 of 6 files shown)


metrics/thresholds_test.go
```diff
@@ -638,12 +638,21 @@ func TestThresholdsRunAll(t *testing.T) {
...
+func TestThresholdsRun(t *testing.T) {
 	t.Parallel()
 
+	// The median is calculated by the `TrendSink.Calc` method.
+	// If values are not added to the sink via its `Add` method,
+	// the sinks `jumbled` flag will remain false.
+	filledTrendSink := &TrendSink{}
+	filledTrendSink.Add(Sample{Time: time.Now(), Value: 80})
+	filledTrendSink.Add(Sample{Time: time.Now(), Value: 80})
+	filledTrendSink.Add(Sample{Time: time.Now(), Value: 80})
```
olegbespalov, line 650: Having all values the same looks like one of the possible test cases, but I believe it's better to have a case where they are diverse. WDYT?

metrics/thresholds_test.go
```diff
@@ -652,34 +661,86 @@ func TestThresholds_Run(t *testing.T) {
...
+		},
 	}
 	for _, testCase := range tests {
 		testCase := testCase
 		t.Run(testCase.name, func(t *testing.T) {
 			t.Parallel()
 
-			thresholds := NewThresholds([]string{"p(95)<2000"})
-			gotParseErr := thresholds.Parse()
+			// thresholds := NewThresholds([]string{"p(95)<2000"})
```
olegbespalov, line 739: If we don't need it, let's remove it.

metrics/thresholds.go
```diff
@@ -188,6 +188,11 @@ func (ts *Thresholds) Run(sink Sink, duration time.Duration) (bool, error) {
 		ts.sinked["min"] = sinkImpl.Min
 		ts.sinked["max"] = sinkImpl.Max
 		ts.sinked["avg"] = sinkImpl.Avg
+
+		// Compute the median value of the trend sink
+		// We do this just in time, as it's needed, because it
+		// requires sorting the values, which computanionally expensive.
+		sinkImpl.Calc()
```
olegbespalov, line 195: I'm missing here a bit of the context, but for me, it sounds like without the `sinkImpl.Calc()` it's not safe to use the `sinkImpl.Med` which means that we can have a small getter in the `sinkImpl` which combines two operations and access to the `Med` should be done only by this getter. That way any other place where we potentially can read the value we don't forget to apply the calculations.

metrics/thresholds.go
```diff
@@ -188,6 +188,11 @@ func (ts *Thresholds) Run(sink Sink, duration time.Duration) (bool, error) {
...
 		ts.sinked["med"] = sinkImpl.Med
```
mstoykov, line 196: Looking at this my immediate reacto is to: 
1. just use `p(50)` here
2. drop `Med` field
3. drop `Calc` from everywhere
4. (optional) prioritize making `med` (and maybe all the rest of the threshold exposed fields) just in time calculated - so instead of you having a structure we always populate with some different fields. Have the threshold told us what it needs and then return that specific field, one at a time. 

WDYT?

metrics/thresholds_test.go
oleiade: Fair enough 👍🏻 I'll switch to `80`, `90`, `100` 😄

metrics/thresholds.go
oleiade: I totally agree, we reached the same conclusion with @imiric. We opened a dedicated issue to address refactoring the issues around `Calc` #2760.

From your perspective would you rather these changes be addressed in this specific PR, or would you be okay with merging this as is, and addressing the refactoring in a separate PR? (I'm fine with both)

mstoykov: I would find 1-3 to be better done in this PR then in a following one.

4. is better of in a separate PR ... likely one with more tests and maybe even a benchmark

oleiade: Very good, let's do that 🤝

oleiade:
For reference, I've went ahead and wrote some benchmarks to compare the performance of `TrendSink.P()` and `TrendSink.Calc()` in computing the median. As far as I can tell, there is no impactful performance difference, supporting the idea to just drop `Calc` all in all.

```go
func BenchmarkTrendSink(b *testing.B) {
	// Produce a fixed random values set for the benchmark.
	values := make([]float64, 0, 10001)
	for i := 0; i < 10000; i++ {
		values = append(values, rand.Float64()*100)
	}

	b.Run("P", func(b *testing.B) {
		// Prepare a sample size table to control the
		// variation of the number of samples for each
		// benchmark.
		var table = []struct {
			percentile float64
			sinkSize   int
		}{
			// 50th percentile
			// Note that computing the 50th percentile is equivalent
			// to computing the median.
			{percentile: 0.5, sinkSize: 10},
			{percentile: 0.5, sinkSize: 100},
			{percentile: 0.5, sinkSize: 1000},
			{percentile: 0.5, sinkSize: 10000},

			// 90th percentile
			{percentile: 0.9, sinkSize: 10},
			{percentile: 0.9, sinkSize: 100},
			{percentile: 0.9, sinkSize: 1000},
			{percentile: 0.9, sinkSize: 10000},

			// 95th percentile
			{percentile: 0.95, sinkSize: 10},
			{percentile: 0.95, sinkSize: 100},
			{percentile: 0.95, sinkSize: 1000},
			{percentile: 0.95, sinkSize: 10000},

			// 99th percentile
			{percentile: 0.99, sinkSize: 10},
			{percentile: 0.99, sinkSize: 100},
			{percentile: 0.99, sinkSize: 1000},
			{percentile: 0.99, sinkSize: 10000},
		}

		for _, v := range table {
			b.Run(fmt.Sprintf("%d_with_%d_values", int(v.percentile*100), v.sinkSize), func(b *testing.B) {
				sink := TrendSink{Values: values[0:v.sinkSize]}

				b.ResetTimer()
				for i := 0; i < b.N; i++ {
					sink.P(v.percentile)
				}
			})
		}
	})

	b.Run("Calc", func(b *testing.B) {
		// Note that we mix odd and even numbers as
		// they correspond to different code paths
		// in the algorithm.
		var table = []struct {
			sinkSize int
		}{
			{sinkSize: 10},
			{sinkSize: 11},
			{sinkSize: 100},
			{sinkSize: 101},
			{sinkSize: 1000},
			{sinkSize: 1001},
			{sinkSize: 10000},
			{sinkSize: 10001},
		}

		for _, v := range table {
			b.Run(fmt.Sprintf("with_%d_values", v.sinkSize), func(b *testing.B) {
				sink := TrendSink{Values: values[0:v.sinkSize]}
				sink.Count = uint64(v.sinkSize)
				sink.jumbled = true

				b.ResetTimer()
				for i := 0; i < b.N; i++ {
					sink.Calc()
				}
			})
		}
	})
}
```

Produces the following output:
```
Running tool: /usr/bin/go test -benchmem -run=^$ -bench ^BenchmarkTrendSink$ go.k6.io/k6/metrics

goos: linux
goarch: amd64
pkg: go.k6.io/k6/metrics
cpu: AMD Ryzen 9 3900XT 12-Core Processor           
BenchmarkTrendSink/P/50_with_10_values-4         	578926669	         2.071 ns/op	       0 B/op	       0 allocs/op
BenchmarkTrendSink/P/50_with_100_values-4        	563760801	         2.096 ns/op	       0 B/op	       0 allocs/op
BenchmarkTrendSink/P/50_with_1000_values-4       	581052195	         2.042 ns/op	       0 B/op	       0 allocs/op
BenchmarkTrendSink/P/50_with_10000_values-4      	582762024	         2.136 ns/op	       0 B/op	       0 allocs/op
BenchmarkTrendSink/P/90_with_10_values-4         	581312148	         2.055 ns/op	       0 B/op	       0 allocs/op
BenchmarkTrendSink/P/90_with_100_values-4        	571656776	         2.129 ns/op	       0 B/op	       0 allocs/op
BenchmarkTrendSink/P/90_with_1000_values-4       	548049884	         2.265 ns/op	       0 B/op	       0 allocs/op
BenchmarkTrendSink/P/90_with_10000_values-4      	559638206	         2.201 ns/op	       0 B/op	       0 allocs/op
BenchmarkTrendSink/P/95_with_10_values-4         	455316678	         2.369 ns/op	       0 B/op	       0 allocs/op
BenchmarkTrendSink/P/95_with_100_values-4        	568306954	         2.085 ns/op	       0 B/op	       0 allocs/op
BenchmarkTrendSink/P/95_with_1000_values-4       	531825253	         2.182 ns/op	       0 B/op	       0 allocs/op
BenchmarkTrendSink/P/95_with_10000_values-4      	520142770	         2.333 ns/op	       0 B/op	       0 allocs/op
BenchmarkTrendSink/P/99_with_10_values-4         	526888444	         2.382 ns/op	       0 B/op	       0 allocs/op
BenchmarkTrendSink/P/99_with_100_values-4        	571957029	         2.164 ns/op	       0 B/op	       0 allocs/op
BenchmarkTrendSink/P/99_with_1000_values-4       	503542969	         2.083 ns/op	       0 B/op	       0 allocs/op
BenchmarkTrendSink/P/99_with_10000_values-4      	583143890	         2.102 ns/op	       0 B/op	       0 allocs/op
BenchmarkTrendSink/Calc/with_10_values-4         	615380493	         1.964 ns/op	       0 B/op	       0 allocs/op
BenchmarkTrendSink/Calc/with_11_values-4         	593478692	         1.984 ns/op	       0 B/op	       0 allocs/op
BenchmarkTrendSink/Calc/with_100_values-4        	573625743	         2.008 ns/op	       0 B/op	       0 allocs/op
BenchmarkTrendSink/Calc/with_101_values-4        	611975035	         2.024 ns/op	       0 B/op	       0 allocs/op
BenchmarkTrendSink/Calc/with_1000_values-4       	616264000	         1.977 ns/op	       0 B/op	       0 allocs/op
BenchmarkTrendSink/Calc/with_1001_values-4       	559834198	         2.048 ns/op	       0 B/op	       0 allocs/op
BenchmarkTrendSink/Calc/with_10000_values-4      	584059414	         1.999 ns/op	       0 B/op	       0 allocs/op
BenchmarkTrendSink/Calc/with_10001_values-4      	612993572	         2.027 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	go.k6.io/k6/metrics	36.063s
```

metrics/thresholds.go
olegbespalov: I'm fine with the suggestions that @mstoykov did and also fine if the proper refactoring will be done in a separated PR

oleiade:
Heads-up, I've implemented the requested changes 👍🏻

metrics/thresholds.go
olegbespalov: Hey @oleiade, Sorry, but I see no changes :cry: probably push is missing

oleiade: Sorry @olegbespalov I forgot I had made the changes in a follow-up branch. My bad 🤦🏻

metrics/sink.go
```diff
@@ -108,10 +111,13 @@ func (t *TrendSink) P(pct float64) float64 {
 	case 1:
 		return t.Values[0]
 	default:
+		if !t.sorted {
+			sort.Float64s(t.Values)
```
mstoykov, line 115:
```suggestion
			sort.Float64s(t.Values)
			t.sorted = true
```

oleiade: Indeed 🤦🏻 Cheers! 🙇🏻

metrics/sink_test.go
```diff
@@ -203,6 +176,61 @@ func TestTrendSink(t *testing.T) {
...
+				sink := TrendSink{Values: values[0:v.sinkSize]}
+
+				b.ResetTimer()
+				for i := 0; i < b.N; i++ {
+					sink.P(v.percentile)
+				}
+			})
+		}
+	})
+}
```
mstoykov, line 232: Not certain what this benchmarks as after the first `sink.P` the sink will be "sorted" (after my other proposed change) and as such it will be super fast  .... 

This are also quite a big number of cases in the table that will make running the benchmarks take quite a while, and I am not certain why we would want to do it for this particular inputs.

Also Trend sizes of < 10k are likely not interesting at all. 

I am for this benchmark basically being dropped as I see it as just taking more time in running the benchmarks. And I can tell you from experience that it already takes too long if you want to have a `-count 5` for any statistical significance.

Arguably we should even go and clean some of the other ones that *likely* aren't all that interesting before or with #2751

metrics/thresholds.go
```diff
@@ -188,7 +188,12 @@ func (ts *Thresholds) Run(sink Sink, duration time.Duration) (bool, error) {
 		ts.sinked["min"] = sinkImpl.Min
 		ts.sinked["max"] = sinkImpl.Max
 		ts.sinked["avg"] = sinkImpl.Avg
-		ts.sinked["med"] = sinkImpl.Med
+
+		// Computing the trend's median value is equivalent
+		// to computing its 50th percentile value.
+		// As this can turn out to be a rather computationally
+		// expensive operation, we do this just in time.
```
mstoykov, line 195:
```suggestion
```

I don't feel like this comment adds much - if you are unaware that the 50th percentile is the median, this comment doesn't tell you why this is true anymore then the code. 

I also have problems with the "just in time" part as I would argue it would've been "just in time" if we did just before *actually* using it, not just before maybe needing it ;)

metrics/thresholds_test.go
```diff
@@ -638,12 +638,21 @@ func TestThresholdsRunAll(t *testing.T) {
 	}
 }
 
-func TestThresholds_Run(t *testing.T) {
+func TestThresholdsRun(t *testing.T) {
 	t.Parallel()
 
+	// The median is calculated by the `TrendSink.Calc` method.
+	// If values are not added to the sink via its `Add` method,
+	// the sinks `jumbled` flag will remain false.
```
mstoykov, line 646:
```suggestion
```

metrics/sink_test.go
```diff
@@ -68,7 +69,6 @@ func TestGaugeSink(t *testing.T) {
 	})
 	t.Run("calc", func(t *testing.T) {
 		sink := GaugeSink{}
-		sink.Calc()
```
olegbespalov, line 71: It seems to me like the whole test run (`t.Run("calc", func(t *testing.T)`) here (and for other metric types) is no longer needed, should we clean up them since it seems like it's the best moment to do that?

metrics/sink_test.go
```diff
@@ -84,64 +84,37 @@ func TestGaugeSink(t *testing.T) {
...
-			assert.Equal(t, true, sink.jumbled)
+			assert.Equal(t, false, sink.sorted)
 			assert.Equal(t, 7.0, sink.Min)
 			assert.Equal(t, 7.0, sink.Max)
 			assert.Equal(t, 7.0, sink.Avg)
-			assert.Equal(t, 0.0, sink.Med) // calculated in Calc()
+
+			// The median value needs to be explicitly calculated
+			// using the the `TrendSink.Calc` method.
+			assert.Equal(t, 0.0, sink.Med)
```
mstoykov, line 101:
```suggestion
```

We should drop the whole field IMO  as we no longer write to it

metrics/sink.go
```diff
@@ -85,9 +80,17 @@ type TrendSink struct {
...
 func (t *TrendSink) IsEmpty() bool { return t.Count == 0 }
 
+// Add adds a sample to the TrendSink.
+//
+// Note that this method does not compute the median value. The
+// operation is delegated to the Calc() method, which is called
+// automatically when the sink is formatted, or when thresholds
+// are checked. The reason is that calculating the median value
+// is an expensive operation, and we don't want to do it for
+// every sample.
```
mstoykov, line 90:
```suggestion
```

This is no longer true and with the removing of the `Med` field will also be unobservable

mstoykov (dismissed):

metrics/sink_test.go
oleiade: Fair enough, I kept it as a relic of illustrating the performance difference between `Calc` and `P`, but you're right, it doesn't add much value outside of this context 👍🏻

metrics/sink_test.go
oleiade: Good catch 👀 I missed those 🙇🏻

metrics/thresholds_test.go
```diff
@@ -652,34 +658,86 @@ func TestThresholds_Run(t *testing.T) {
...
+		},
 	}
 	for _, testCase := range tests {
 		testCase := testCase
 		t.Run(testCase.name, func(t *testing.T) {
 			t.Parallel()
 
-			thresholds := NewThresholds([]string{"p(95)<2000"})
-			gotParseErr := thresholds.Parse()
+			// thresholds := NewThresholds([]string{"p(95)<2000"})
```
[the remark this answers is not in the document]
oleiade: Rebase mistake, good catch, thanks :)

metrics/thresholds_test.go
olegbespalov: It seems like this was affected by rebase to :cry:

oleiade: Arf 😞 I need to get better at this 🙇🏻

olegbespalov (approved):

na-- (dismissed):
:heart_eyes: :tada:

mstoykov (commented):
Can we drop `Med` on `TrendSink` as well?

na-- (changes requested):
Do not use atomics for `sorted`. See that there is also no locking for `Values`, which is arguably even more important. These `Sink` implementation are not meant to be thread-safe, so if there are races, the problem lies in the code that uses them.

oleiade:
@na-- I thought so too, but thought of trying it to track the race condition appearing in the CI but that I couldn't reproduce locally. Sorry for the noise.

I've benchmarked things again, and having a `sorted` flag has no noticeable impact on performance anyways 👍🏻

na--:
What was the data race? Can you paste it here, since I can't see it in any of the previous CI runs in https://github.com/grafana/k6/actions :confused:

oleiade:
@na-- Here's an example: https://github.com/grafana/k6/actions/runs/3428932294/jobs/5713872207

I've confirmed it was due to setting the `TrendSink.sorted` to true (pre-atomic) here: https://github.com/grafana/k6/pull/2759/files#diff-878c4547f8d16497e0b27c511b2d1f65663a93422548d060af4abe1d77274ebcR108.

Benchmark, and test scripts I've written show basically no difference in performance whether or not we store the `sorted` state or not. Thus I'll just drop it all in all, unless of you folks has an objection against it? 🤔

metrics/thresholds_test.go
```diff
@@ -638,12 +638,18 @@ func TestThresholdsRunAll(t *testing.T) {
...
 }
 
-func TestThresholds_Run(t *testing.T) {
+func TestThresholdsRun(t *testing.T) {
 	t.Parallel()
 
+	filledTrendSink := &TrendSink{}
+	filledTrendSink.Add(Sample{Time: time.Now(), Value: 70})
+	filledTrendSink.Add(Sample{Time: time.Now(), Value: 80})
+	filledTrendSink.Add(Sample{Time: time.Now(), Value: 90})
```
mstoykov, line 647: This needs to be in the `t.Run` below so that you don't wokr on the same Trend - this is where the race comes from

oleiade: Thanks 🙇🏻

metrics/sink.go
```diff
@@ -66,28 +63,23 @@ func (g *GaugeSink) Add(s Sample) {
...
 	Sum, Avg float64
-	Med      float64
 }
 
 // IsEmpty indicates whether the TrendSink is empty.
 func (t *TrendSink) IsEmpty() bool { return t.Count == 0 }
 
 func (t *TrendSink) Add(s Sample) {
 	t.Values = append(t.Values, s.Value)
-	t.jumbled = true
```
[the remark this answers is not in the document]
oleiade: This made me reconsider the way I benchmarked it, and on a very large number of values in the sink, it starts showing a noticeable difference: I was wrong, thanks for pointing it out.

Also, `jumbled` is called `sorted` now. While reviewing the code `jumbled` meant literally nothing to me, thus I renamed it to `sorted` and inverted the logic.

metrics/sink_test.go
```diff
@@ -84,60 +70,23 @@ func TestGaugeSink(t *testing.T) {
...
 		})
 		t.Run("values", func(t *testing.T) {
 			sink := TrendSink{}
 			for _, s := range unsortedSamples10 {
 				sink.Add(Sample{TimeSeries: TimeSeries{Metric: &Metric{}}, Value: s})
 			}
 			assert.Equal(t, uint64(len(unsortedSamples10)), sink.Count)
-			assert.Equal(t, true, sink.jumbled)
-			assert.Equal(t, 0.0, sink.Min)
-			assert.Equal(t, 100.0, sink.Max)
```
olegbespalov, line 109: The cases `Min`,`Max`, `Avg` on the unsorted array look important. Am I missing a reason why we remove them? :thinking: 

```golang
assert.Equal(t, 0.0, sink.Min)
assert.Equal(t, 100.0, sink.Max)
assert.Equal(t, 54.0, sink.Avg)
```

oleiade: There just have been too many changes overall for me to efficiently keep track of, and I do 3 things at the same time, I lost track or messed it up in a rebase, again 🤦🏻 thanks for pointing it out.

oleiade: Those specific lines have shifted down, I guess the GitHub interface is confusing: https://github.com/grafana/k6/pull/2759/files/179c2dd47ea83b8f6563e83019108f1b898035c8#diff-f37293b2fc466b351eccf12f450b485548f0ca27048f28327602c752f8b2db24R90

But I guess it's also true that with removing `jumbled` I lost a bit track along the way, I'll add a test back to make sure  `sorted` and not sorted are tested 👍🏻

oleiade: I've brought back the initial test on the `sorted` state, thanks for bringing that up 👍🏻 

Plus, I've added a couple of others to assert that the `Add` and `P` operations do modify it 🚀

olegbespalov: Sorry, my bad :cry: Thanks! :bow:

olegbespalov (approved):

PR #2767, opened 2022-11-11
Fix emission of `ws_session_duration` when setup throws an error

### LINKED ISSUE #2734: k6/ws: connect doesn't emit wssessionduration on exception in setup
### Brief summary

The `k6/ws` module's 'connect` doesn't emit `WSSessionDuraiton` if there is exception the first callback it has.

Given that at that point k6 will emit `ws_sessions` (and `ws_connecting) (since 0.27.1 at least) - it seems reasonable to also emit `ws_session_duration` 

This was noticed while rewriting test for lint issues and it turn out [one of them](https://github.com/grafana/k6/pull/2732#discussion_r996756991) was checking for this, but was getting false possitive due to reusage of test setup. 

### k6 version

v0.40.0

### OS

linux

_No response_

run 
```
import ws from 'k6/ws';

export default function() {
  let url = `wss://test-api.k6.io/ws/crocochat/test/`;

  let res = ws.connect(url, function(socket) { throw "oops" });
}
```

Obviously the real code will be probably something that fails with an exception then actually just throwing within ;)

### Expected behaviour

Emit all the three ws metrics even with very low valuse

### Actual behaviour

emitting 2 of the 3 metrics

imiric:
Closes #2734

This only fixes the issue, but see #2768 for some related refactoring of `WS.Connect()`.

### DIFF
--- a/js/modules/k6/ws/ws.go
+++ b/js/modules/k6/ws/ws.go
@@ -273,6 +273,21 @@ func (mi *WS) Connect(url string, args ...goja.Value) (*HTTPResponse, error) {
 		Time: start,
 	})
 
+	defer func() {
+		end := time.Now()
+		sessionDuration := metrics.D(end.Sub(start))
+
+		metrics.PushIfNotDone(ctx, state.Samples, metrics.Sample{
+			TimeSeries: metrics.TimeSeries{
+				Metric: socket.builtinMetrics.WSSessionDuration,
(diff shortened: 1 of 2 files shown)


codebien (approved):
LGTM, thanks :bow:

oleiade (approved):
👏🏻

PR #2768, opened 2022-11-11
Refactor `WS.Connect()`

imiric:
This refactors parts of the `WS.Connect()` method to extract argument parsing and establishing of the WS connection into separate functions.

It's built on top of #2767, so we should review/merge that first.

### DIFF
--- a/js/modules/k6/ws/ws.go
+++ b/js/modules/k6/ws/ws.go
@@ -10,6 +10,7 @@ import (
 	"io"
 	"net"
 	"net/http"
+	"net/http/cookiejar"
 	"strconv"
 	"strings"
 	"sync"
@@ -21,6 +22,7 @@ import (
 	"go.k6.io/k6/js/common"
 	"go.k6.io/k6/js/modules"
 	httpModule "go.k6.io/k6/js/modules/k6/http"
+	"go.k6.io/k6/lib"
 	"go.k6.io/k6/metrics"
 )
 
@@ -97,6 +99,14 @@ type message struct {
 	data  []byte
 }
 
+type wsConnectArgs struct {
+	setupFn           goja.Callable
+	headers           http.Header
+	enableCompression bool
+	cookieJar         *cookiejar.Jar
+	tagsAndMeta       *metrics.TagsAndMeta
+}
+
 const writeWait = 10 * time.Second
 
 // Exports returns the exports of the ws module.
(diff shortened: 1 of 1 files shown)


js/modules/k6/ws/ws.go
```diff
@@ -399,6 +242,72 @@ func (mi *WS) Connect(url string, args ...goja.Value) (*HTTPResponse, error) {
...
+
+	connStart := time.Now()
+	conn, httpResponse, dialErr := wsd.DialContext(ctx, url, args.headers)
+	connEnd := time.Now()
+
+	if state.Options.SystemTags.Has(metrics.TagIP) && conn.RemoteAddr() != nil {
+		if ip, _, err := net.SplitHostPort(conn.RemoteAddr().String()); err == nil {
+			args.tagsAndMeta.SetSystemTagOrMeta(metrics.TagIP, ip)
+		}
+	}
```
codebien, line 279: I see we have this same code across http/grpc/ws should we create a helper function and reuse it?

js/modules/k6/ws/ws.go
```diff
@@ -667,3 +620,87 @@ func wrapHTTPResponse(httpResponse *http.Response) (*HTTPResponse, error) {
 
 	return &wsResponse, nil
 }
+
+//nolint:gocognit
```
codebien, line 624: What should we do for removing it? Would adding a default case in the switch statement be enough?

codebien (commented):
LGTM, some minor comments

olegbespalov:
I've skimmed and it looks good :+1: One thing that I guess is important to mention is that as I understood @mstoykov a long-term plan is that we're going to deprecate the current `k6/ws` in favor of the new `xk6-websockets` (which currently is an experimental module), so not sure if we need to invest any effort for the refactoring old module :shrug:

olegbespalov (dismissed):

imiric:
Good point @olegbespalov :+1: I knew this would be deprecated, probably sooner than later, but since I was already touching it in #2734, and this always bothered me, I thought a quick refactor wouldn't hurt. :)

js/modules/k6/ws/ws.go
imiric: :thinking: Probably. But that would expand the scope of this refactor, and I didn't want to spend too much time on it.

Maybe create an issue for it?

js/modules/k6/ws/ws.go
imiric: Doesn't seem so. We would likely need to extract some of it into helper functions, but IMO this function is readable enough, and it's only slightly over the linter limit (33 > 30). At a certain point, there are diminishing returns trying to appease a linter, so I wouldn't bother with it. Especially since we plan to deprecate this module anyway. :man_shrugging:

codebien: LGTM

codebien (dismissed):

oleiade (dismissed):
👏🏻

imiric:
Sorry, I had to force-push to squash @codebien's suggestion.

codebien (approved):

olegbespalov (approved):

PR #2771, opened 2022-11-14
Disable containedctx

olegbespalov:
# What?

Disabling the `containedctx`

# Why?

It make no sense for this repo

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -81,4 +81,5 @@ linters:
   - decorder
   - nonamedreturns
   - nosnakecase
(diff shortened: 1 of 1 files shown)


oleiade (approved):
I disagree with the statement that this linter makes no sense in this repo. It's a recommended practice to not store contexts in structs: https://go.dev/blog/context-and-structs. I acknowledge we do it, and that we prefer to not refactor the code doing, probably for good reason, thus, this gets my 👍🏻  😄

PR #2777, opened 2022-11-17
Move to Iterative Trie implementation (`trieNode`)

eugercek:
Hi @codebien 👋🏻 here is the aforementioned PR

What have changed:

- Moved `trieNode` to separate file
- New unit tests for trie implementation
- New benchmarks for the `trieNode.insert()` and `trieNode.contains()` methods
- Moved from recursive to iterative implementation

Here is a simple script that I used to create the benchmarks on my machine.

```bash
bench(){
  echo "started: $2 benchmark"
  git checkout "$1"  2> /dev/null
  go test -run=^$ -bench=BenchmarkTrie -benchmem -count 30 > "$2"
}

cd lib/types

bench 2f14099  "original"
bench e11e0679 "remove_error"
bench 08933201 "iterative"

benchstat original remove_error > "bs_remove_error"
benchstat original iterative > "bs_iterative"
```

# Benchmarks

My machine was idle.

```
goos: darwin
goarch: amd64
pkg: go.k6.io/k6/lib/types
cpu: Intel(R) Core(TM) i5-8257U CPU @ 1.40GHz
```

##  `bs_iterative`

```
name            old time/op    new time/op    delta
TrieInsert-8      13.3µs ± 1%     6.3µs ± 1%  -52.78%  (p=0.000 n=29+30)
TrieContains-8    11.7µs ± 2%     3.1µs ± 1%  -73.01%  (p=0.000 n=30+27)

name            old alloc/op   new alloc/op   delta
TrieInsert-8      7.46kB ± 0%    7.46kB ± 0%     ~     (all equal)
TrieContains-8    1.02kB ± 0%    0.42kB ± 0%  -58.27%  (p=0.000 n=30+30)

name            old allocs/op  new allocs/op  delta
TrieInsert-8         130 ± 0%       130 ± 0%     ~     (all equal)
TrieContains-8      90.0 ± 0%      24.0 ± 0%  -73.33%  (p=0.000 n=30+30)
```

##  `bs_remove_error`

```
name          old time/op    new time/op    delta
TrieInsert-8    14.0µs ± 0%    13.3µs ± 1%   ~     (p=0.106 n=1+28)

name          old alloc/op   new alloc/op   delta
TrieInsert-8    7.46kB ± 0%    7.46kB ± 0%   ~     (all equal)

name          old allocs/op  new allocs/op  delta
TrieInsert-8       130 ± 0%       130 ± 0%   ~     (all equal)
```

### DIFF
--- a/lib/types/trie.go
+++ b/lib/types/trie.go
@@ -8,46 +8,69 @@ type trieNode struct {
 }
 
 func (t *trieNode) insert(s string) {
-	if len(s) == 0 {
-		t.isLeaf = true
-		return
-	}
+	runes := []rune(s)
 
-	// mask creation of the trie by initializing the root here
 	if t.children == nil {
-		t.children = make(map[rune]*trieNode)
+		t.children = map[rune]*trieNode{}
 	}
 
-	rStr := []rune(s) // need to iterate by runes for intl' names
-	last := len(rStr) - 1
-	if c, ok := t.children[rStr[last]]; ok {
-		c.insert(string(rStr[:last]))
-		return
+	ptr := t
+	for i := len(runes) - 1; i >= 0; i-- {
+		r := runes[i]
+		c, ok := ptr.children[r]
+
+		if !ok {
+			ptr.children[r] = &trieNode{children: map[rune]*trieNode{}}
+			c = ptr.children[r]
+		}
+
+		ptr = c
 	}
 
-	t.children[rStr[last]] = &trieNode{children: make(map[rune]*trieNode)}
-	t.children[rStr[last]].insert(string(rStr[:last]))
+	ptr.isLeaf = true
 }
 
-func (t *trieNode) contains(s string) (matchedPattern string, matchFound bool) {
-	s = strings.ToLower(s)
-	if len(s) == 0 {
-		if t.isLeaf {
-			return "", true
+func (t *trieNode) contains(s string) (string, bool) {
+	rs := []rune(strings.ToLower(s))
+
+	builder, wMatch := strings.Builder{}, ""
+	found := true
+
+	ptr := t
+	for i := len(rs) - 1; i >= 0; i-- {
+		child, ok := ptr.children[rs[i]]
+		_, wOk := ptr.children['*']
+
+		if wOk {
+			wMatch = builder.String() + string('*')
 		}
-	} else {
-		rStr := []rune(s)
-		last := len(rStr) - 1
-		if c, ok := t.children[rStr[last]]; ok {
-			if match, matched := c.contains(string(rStr[:last])); matched {
-				return match + string(rStr[last]), true
-			}
+
+		if !ok {
+			found = false
+			break
 		}
+
+		builder.WriteRune(rs[i])
+		ptr = child
 	}
 
-	if _, wild := t.children['*']; wild {
-		return "*", true
+	if found && ptr.isLeaf {
+		return reverse(builder.String()), true
 	}
 
-	return "", false
+	if _, ok := ptr.children['*']; ok {
+		builder.WriteRune('*')
+		return reverse(builder.String()), true
+	}
+
+	return reverse(wMatch), wMatch != ""
+}
+
+func reverse(s string) string {
+	rs := []rune(s)
+	for i, j := 0, len(rs)-1; i < len(rs)/2; i, j = i+1, j-1 {
+		rs[i], rs[j] = rs[j], rs[i]
+	}
+
+	return string(rs)
 }
--- a/lib/types/trie.go
+++ b/lib/types/trie.go
@@ -7,10 +7,10 @@ type trieNode struct {
 	children map[rune]*trieNode
 }
 
-func (t *trieNode) insert(s string) error {
+func (t *trieNode) insert(s string) {
 	if len(s) == 0 {
 		t.isLeaf = true
-		return nil
+		return
 	}
 
 	// mask creation of the trie by initializing the root here
@@ -21,11 +21,12 @@ func (t *trieNode) insert(s string) error {
 	rStr := []rune(s) // need to iterate by runes for intl' names
 	last := len(rStr) - 1
 	if c, ok := t.children[rStr[last]]; ok {
-		return c.insert(string(rStr[:last]))
+		c.insert(string(rStr[:last]))
+		return
 	}
 
 	t.children[rStr[last]] = &trieNode{children: make(map[rune]*trieNode)}
-	return t.children[rStr[last]].insert(string(rStr[:last]))
+	t.children[rStr[last]].insert(string(rStr[:last]))
 }
 
 func (t *trieNode) contains(s string) (matchedPattern string, matchFound bool) {
--- a/lib/types/trie.go
+++ b/lib/types/trie.go
@@ -1,0 +1,52 @@
+package types
+
+import "strings"
+
+type trieNode struct {
+	isLeaf   bool
+	children map[rune]*trieNode
+}
+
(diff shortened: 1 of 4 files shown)


lib/types/trie_test.go
```diff
@@ -77,3 +77,35 @@ func TestTrie(t *testing.T) {
...
+	root := trieNode{}
+	arr := []string{
+		"k6.io", "*.sub.k6.io", "specific.sub.k6.io",
+		"grafana.com", "*.sub.sub.grafana.com", "test.sub.sub.grafana.com",
+	}
+
+	for _, v := range arr {
+		root.insert(v)
+	}
+
```
codebien, line 111: `b.ResetTimer()` should be added here

codebien (changes requested):
Hi @eugercek,
thanks again for your effort around this contribution.

I quickly skimmed the code, I will back to you later or on Monday with a better review.

In addition to the comments below,  I think we should 1aa445136973e1d60c526507993778941fffcac0 commit into 7ab4e56aa3cab960889d0a3f99f148b0f4607d91. You can do it using the fixup during a rebase.

lib/types/trie_test.go
eugercek: Added in both

eugercek:
IMO `trieNode` should be generic enough because it's a DS, it should be usable in other part of the codebase. Currently we only use in `HostnameTrie` and `AddresTrie` (not in master a separate PR). Because hostname is not case sensitive we lowercased it in `trieNode.contains()` too, last commit fixes that.

imiric (commented):
Thanks for this @eugercek! :clap:

The changes LGTM, besides the minor linter issues.

I haven't reviewed the full changes in #2747 yet, but is the move to an iterative implementation done because of performance or stack limitation reasons when adding wildcard support for `hosts`? I find the recursive version simpler to read, and wondering whether this optimization is worth the slight increase in complexity for our use case.

eugercek:
Hi @imiric this PR was solely optional, I thought iterative one can be simpler and more performant. I think `insert` is simpler but I agree `contains` is more complex in iterative implementation. It's up to you guys.

Fixed the linter errors, sorry for that.

Also thanks for all of your kind words, it's my pleasure to ,trying to, contribute to k6.

imiric (approved):
Makes sense @eugercek. I doubt a typical k6 user would run into performance problems with the previous implementation, but it is a nice improvement, so I don't see why we shouldn't merge this.

I think we can ignore the ST1000 linter error, and the failed tests, which are known to be flaky.

Thanks for your work! :bow:

lib/types/trie.go
```diff
@@ -8,46 +8,69 @@ type trieNode struct {
...
 	}
 
-	rStr := []rune(s) // need to iterate by runes for intl' names
-	last := len(rStr) - 1
-	if c, ok := t.children[rStr[last]]; ok {
-		c.insert(string(rStr[:last]))
-		return
+	ptr := t
+	for i := len(runes) - 1; i >= 0; i-- {
+		r := runes[i]
```
codebien, line 19: This allocation can be removed and `runes[i]` could be reused directly.

lib/types/trie.go
```diff
@@ -8,46 +8,69 @@ type trieNode struct {
...
-		return "*", true
+	if _, ok := ptr.children['*']; ok {
+		builder.WriteRune('*')
+		return reverse(builder.String()), true
+	}
+
+	return reverse(wMatch), wMatch != ""
+}
+
+func reverse(s string) string {
```
codebien, line 69: s/reverse/reverseString/

lib/types/trie.go
```diff
@@ -8,46 +8,69 @@ type trieNode struct {
...
+	builder, wMatch := strings.Builder{}, ""
+	found := true
+
+	ptr := t
+	for i := len(rs) - 1; i >= 0; i-- {
+		child, ok := ptr.children[rs[i]]
+		_, wOk := ptr.children['*']
+
+		if wOk {
+			wMatch = builder.String() + string('*')
```
codebien, line 45:
```go
if _, wOk := ptr.children['*']; wOk {
			wMatch = builder.String() + string('*')
```

lib/types/trie.go
```diff
@@ -8,46 +8,69 @@ type trieNode struct {
...
+	ptr.isLeaf = true
 }
 
-func (t *trieNode) contains(s string) (matchedPattern string, matchFound bool) {
-	s = strings.ToLower(s)
-	if len(s) == 0 {
-		if t.isLeaf {
-			return "", true
+func (t *trieNode) contains(s string) (string, bool) {
+	rs := []rune(strings.ToLower(s))
```
codebien: You could conside to use `unicode.ToLower` directly in the `for` below

codebien (changes requested):
LGTM, thanks @eugercek :bow:

Some minor request changes.

lib/types/trie.go
eugercek: Changed

lib/types/trie.go
eugercek: Done

lib/types/trie.go
eugercek: Done

lib/types/trie.go
eugercek: Have you seen [c14be6b](https://github.com/grafana/k6/pull/2777/commits/c14be6b8ea70900d132f7ad87841c1424114fe6f)

> IMO trieNode should be generic enough because it's a DS, it should be usable in other part of the codebase. Currently we only use in HostnameTrie and AddresTrie (not in master a separate PR). Because hostname is not case sensitive we lowercased it in trieNode.contains() too, last commit fixes that.

Which way should we proceed?

eugercek:
I'll rebase/fixup after the last change.

lib/types/trie.go
codebien: I missed it, thanks for pointing me there. I agree it is better to keep the data structure as generic as possible.

codebien (changes requested):
@eugercek LGTM, can you fix the linter, please?

eugercek:
Only ` ST1000: at least one file in a package should have a package comment ` was producing error, added a little comment.

lib/types/trie_test.go
```diff
@@ -0,0 +1,120 @@
...
+			val, ok := root.contains(tc.query)
+
+			require.Equal(t, tc.found, ok)
+			require.Equal(t, tc.expVal, val)
+		})
+	}
+}
+
+func TestReverse(t *testing.T) {
```
imiric, line 67:
```suggestion
func TestReverseString(t *testing.T) {
```

lib/types/types.go
```diff
@@ -1,3 +1,6 @@
+// Package types contains types used in the codebase
+// Most of the types have Null prefix like gopkg.in/guregu/null.v3
+// Most of the types have UnmarshalJSON and MarshalJSON methods
```
imiric, line 3:
```suggestion
// Most of the types have a Null prefix like gopkg.in/guregu/null.v3
// and UnmarshalJSON and MarshalJSON methods.
```

imiric (approved):
Just a couple of nitpicks, LGTM otherwise. Thanks again!

lib/types/trie_test.go
eugercek: Fixed

lib/types/types.go
eugercek: Fixed

imiric (approved):

codebien (approved):
LGTM, thanks @eugercek :tada:

Issue #2674, opened 2022-09-01
ESM discussion: `open` and `require` are relative to?


mstoykov:
## Background:
Currently, `require` and `open` are both relative to the current *root* module/script that is being executed. This is best explain with an example:

If you have. 

`main.js` that imports 'folder/B.js' 
```javascript
// this is main.js
import { a } from "./folder/B.js";
var somefile = a(); // this will open `./data.json`
export default() {}
```

```javascript
export function a () {
   return open("./data.json");
}
a() // this will open `./folder/data.json`
```

The same goes for `require`. 

## Problem/question:

`import()` (the dynamic import)  is defined as being relative to the file it's being written in as in the above case if you replace `open` with `import` it will in *both* instances open `./folder/data.json`. 

The same is technically true for how `require` is defined :( , from https://wiki.commonjs.org/wiki/Website:Index/specs/modules/1.0

>  Relative identifiers are resolved relative to the identifier of the module in which "require" is written and called.

Arguably both `require` and `open` will be almost always be used in the top context of the module/script, so it wouldn't matter - they will just always be relative to the file they are written in.

So the questions are two:
1. should `require` be fixed to be relative to the module it is in ... basically the same as `import()` and as it's described?
2. should `open` be the same just to not be confusing?

From implementation perspective making it all the same will be way easier in ESM as getting the root module/script that is currently running isn't a thing goja currently supports and likely will require some hacks.

mstoykov:
Closing in favor of #3857

PR #2782, opened 2022-11-23
Add tests around how require and open treat relative paths

mstoykov:
Updates #2674

### DIFF
--- a/js/path_resolution_test.go
+++ b/js/path_resolution_test.go
@@ -1,0 +1,143 @@
+package js
+
+import (
+	"testing"
+
+	"github.com/spf13/afero"
+	"github.com/stretchr/testify/require"
+	"go.k6.io/k6/lib/testutils"
+)
+
+// This whole file is about tests around https://github.com/grafana/k6/issues/2674
+
+func TestOpenPathResolution(t *testing.T) {
+	t.Parallel()
+	t.Run("simple", func(t *testing.T) {
+		t.Parallel()
+		fs := afero.NewMemMapFs()
+		require.NoError(t, afero.WriteFile(fs, "/path/to/data.txt", []byte("data file"), 0o644))
+		data := `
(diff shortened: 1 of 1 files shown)


js/path_resolution_test.go
```diff
@@ -0,0 +1,143 @@
...
+
+	t.Run("intermediate", func(t *testing.T) {
+		t.Parallel()
+		fs := afero.NewMemMapFs()
+		require.NoError(t, afero.WriteFile(fs, "/path/to/data.txt", []byte("data file"), 0o644))
+		require.NoError(t, afero.WriteFile(fs, "/path/another/script/script.js", []byte(`
+            module.exports = open("../../to/data.txt");
+        `), 0o644))
+		data := `
+        let data = require("./../../../another/script/script.js")
```
oleiade, line 41: non-blocking nitpick 🙇🏻  I found it hard to process what the test demonstrates with all the `..` symbols, if we could reduce it to the smallest amount that still demonstrates the issue (by for instance simplifying the paths to the files for instance) I would find it beneficial. That's applying to the whole file 👍🏻

oleiade (approved):
👍🏻 🙇🏻

js/path_resolution_test.go
mstoykov: The whole tests is about how `require` and `open` work with relative paths and what are they *relative* to. 

I did start with somewhat simpler examples but I fast realized that if all my paths are `../otherDir/data.js` it will work always as it files will be on the same fs "depth". So to avoid that I started making them more .... depthly so I can be certain that if I am off by a depth it will give me an error :) 

I might try to get this somewhat nicer when/if we change what either one is relative to.

oleiade: Fair enough, if it does make more sense this way, then it's perfectly fine by me 👍🏻 🙇🏻

PR #2784, opened 2022-11-29
Experimental Prometheus remote-write Output

codebien:
:warning: This is not yet ready for being merged because there is some working progress on the extension. However, the direct changes in this PR are pretty small, and they can be reviewed.

WIP tasks:
- https://github.com/grafana/xk6-output-prometheus-remote/pull/67
- https://github.com/grafana/xk6-output-prometheus-remote/pull/64
- Respect the naming convention for Counter, Rate (and Gauges?) - Trend already implements it
- https://github.com/grafana/xk6-output-prometheus-remote/issues/68


cmd/outputs.go
```diff
@@ -32,6 +34,9 @@ func getAllOutputConstructors() (map[string]func(output.Params) (output.Output,
 				"please use the statsd output with env. variable K6_STATSD_ENABLE_TAGS=true instead")
 		},
 		"csv": csv.New,
+		"experimental-prometheus-remote-write": func(params output.Params) (output.Output, error) {
```
javaducky, line 37: Wow...this is going to have a lot of typos in usage 😅. Should this be abbreviated as "xp-prometheus" possibly?

vendor/github.com/grafana/xk6-output-prometheus-remote/pkg/remotewrite/config.go
```diff
@@ -0,0 +1,274 @@
...
+
+	if b, err := getEnvBool(env, "K6_PROMETHEUS_INSECURE_SKIP_TLS_VERIFY"); err != nil {
+		return c, err
+	} else {
+		if b.Valid {
+			c.InsecureSkipTLSVerify = b
+		}
+	}
+
+	if user, userDefined := env["K6_PROMETHEUS_USERNAME"]; userDefined {
```
javaducky, line 199: Breaking change! This was previously `K6_PROMETHEUS_USER`.

javaducky (dismissed):
LGTM! I published an image to Dockerhub and used it with my k6-operator demo to publish metrics using authentication against my Grafana Cloud instance...worked perfectly!

The only hiccup was related to the renaming of the `K6_PROMETHEUS_USER` variable to `K6_PROMETHEUS_USERNAME`.

imiric (dismissed):
LGTM, besides what Paul mentioned. There's not much else here, besides the many, many dependencies which I'm not reviewing :sweat_smile:, and the Prometheus extension itself, which we can review and fix any issues with in its repo.

Great work @codebien! :clap:

cmd/outputs.go
ppcano: I was not very convinced with `experimental-xxx`. Like much more this proposal.

na--: I am very much against dropping the `experimental` part of the name, `xp` has absolutely no inherent connection to `experimental`. And `experimental` is temporary anyway, it will be dropped in a few k6 releases, so it's not a problem.

I don't mind shortening the rest of the name, e.g. `experimental-prometheus-rw` that will later become just `prometheus-rw`. But I am on the fence, since even that might cause some confusion.

javaducky: Any longer, I believe Remote Write is on by default for Prometheus instances, so I feel like we could omit that distinction and then settle on `experimental-prometheus`.

na--: It might be on by default, but there is a substantial difference in how k6 would operate in a pull-based and a push-based Prometheus setup. It's not clear we'll ever implement the pull-based approach, but if we do, I imagine its options might be quite different and it will be a separate output type :thinking:

codebien: The default way for Prometheus is still the pull-based model. The remote write has to be enabled by a specific flag and the current Output doesn't work without the receiver enabled.

Also, important to mention that Prometheus Remote Write is a protocol and not a specific implementation, the Prometheus server is one of the available implementations. Does only `prometheus` make sense if we flush to another service (e.g. Mimir)? `K6_PROMETHEUS_REMOTE_URL=http://mimir:9090/api/v1/write k6 run -o prometheus`.

The unique alternative I can think is `experimental-promrw`. Prom is a common abbreviation for Prometheus.

In the case we confirm `experimental-prometheus-remote-write` as the output's name, is the current prefix `K6_PROMETHEUS_` fine or it would be better to have `K6_PROMETHEUS_REMOTE_WRITE_`?

javaducky: Sorry, I should have said that I believe remote write is "enabled by default." I realize the pull-based model is still the primary, or default, method. I didn't think that newer versions require the feature flag any longer, but I very well could be wrong.

Anyway, I only wanted to point out that the lengthy name for the flag would be prone to error. I'm fine with it either way.

javaducky: > In the case we confirm `experimental-prometheus-remote-write` as the output's name, is the current prefix `K6_PROMETHEUS_` fine or it would be better to have `K6_PROMETHEUS_REMOTE_WRITE_`?

As much as I hate to say it, we _should_ be consistent :(

codebien: The team decided to move to the compromise `experimental-prometheus-rw`.

imiric (approved):
:rocket:

codebien:
Added a simple integration test

imiric (approved):
Thanks for the test!

na-- (approved):

Issue #763, opened 2018-09-12
Use HDR histograms for calculating percentiles in thresholds and summary stats


na--:
Currently the [`Trend`](https://docs.k6.io/docs/trend-k6metrics)-based threshold checks and end-of-test summary stats rely on saving all of the relevant metric values in-memory. This is basically a large memory leak that can negatively affect long-running and/or HTTP-heavy load tests, as [reported](https://k6io.slack.com/archives/C3E688J8Z/p1535727394000100) by this user on slack.

For the moment the best solution for solving those issues without any loss of functionality or and only a very tiny loss of precision appears to be [HdrHistogram](http://hdrhistogram.github.io/HdrHistogram/). Here's an excerpt from the description in the above website:
> A Histogram that supports recording and analyzing sampled data value counts across a configurable integer value range with configurable value precision within the range. Value precision is expressed as the number of significant digits in the value recording, and provides control over value quantization behavior across the value range and the subsequent value resolution at any given level.
> [ ... ]
> HDR Histogram is designed for recoding histograms of value measurements in latency and performance sensitive applications. [ ... ] The HDR Histogram maintains a fixed cost in both space and time. A Histogram's memory footprint is constant, with no allocation operations involved in recording data values or in iterating through them. The memory footprint is fixed regardless of the number of data value samples recorded, and depends solely on the dynamic range and precision chosen. The amount of work involved in recording a sample is constant, and directly computes storage index locations such that no iteration or searching is ever involved in recording data values.

[This](https://github.com/HdrHistogram/HdrHistogram) is the original Java library by Gil Tene. He also has some great talks ([this](https://www.infoq.com/presentations/latency-response-time) and its [earlier](https://www.infoq.com/presentations/latency-pitfalls) version for example) that explain some of the common pitfalls when people measure latency and why he built HdrHistogram. They're also a very strong argument why we should prioritize the [arrival-rate based VU executor](https://github.com/loadimpact/k6/issues/550)... :smile: 

[This](https://github.com/codahale/hdrhistogram) is an MIT-licensed Go implementation of HdrHistogram, though it seems to be dead - archived repo with no recent commits and unresolved issues and PRs. So we may need to fork that repo and maintain it or and re-implement the algorithm ourselves.

[...]

PR #2785, opened 2022-11-29
Fix code comment

codebien:
~Add the Sum stat into the resolvers map. It is helpful for reusing the same function from the Remote Write output. https://github.com/grafana/xk6-output-prometheus-remote/pull/64~

Just fix a comment

### DIFF
--- a/metrics/value_type.go
+++ b/metrics/value_type.go
@@ -5,7 +5,7 @@ import "errors"
 // Possible values for ValueType.
 const (
 	Default = ValueType(iota) // Values are presented as-is
-	Time                      // Values are timestamps (nanoseconds)
+	Time                      // Values are time durations (milliseconds)
 	Data                      // Values are data amounts (bytes)
 )
 


metrics/value_type.go
```diff
@@ -5,7 +5,7 @@ import "errors"
 // Possible values for ValueType.
 const (
 	Default = ValueType(iota) // Values are presented as-is
-	Time                      // Values are timestamps (nanoseconds)
+	Time                      // Values are timestamps (milliseconds)
```
na--, line 8:
```suggestion
	Time                      // Values are durations (milliseconds)
```

na-- (commented):
Hmm :thinking: if I am reading this correctly: https://github.com/grafana/k6/blob/616e5e2f24ba41995f4d8bf9e1fab8f89285fb44/cmd/options.go#L191-L200

Then changing this function will also allow users to specify `k6 run --summary-trend-stats min,med,max,sum`, which is not something we want, right? :thinking: 

Why do you need this change?

na--:
I am not sure how the sum of a `Trend`'s values would be useful in any way for the end-of-test summary? Maybe it could be of use in some weird `handleSummary()` use case? :thinking: Even so, AFAIK, nobody has requested something like this so far.

Also, the `sum` is easy to get with our current `TrendSink` implementation, since it doesn't require any calculation - we update it on every value `add()` call. However, if we adopt HDR histograms (https://github.com/grafana/k6/issues/763), calculating the sum would [require some computation](https://github.com/openhistogram/circonusllhist/blob/c78ce013c914a89b519b0e9e184e62589cb727a9/circonusllhist.go#L771-L784) and even so, it will be just an approximation.

So, given the dubious value to users and the potential complication for HDR histograms, I think copying some code to the extension (with a `TODO`) would be preferable to hastily updating the values in this k6 function.

codebien:
Ok, then I think I can do something hacky for excluding the `sum` from the validation and still reuse the k6 function. Thanks for your opinion :bow:

na--:
The fix to the `metrics.Time` comments were good though :sweat_smile:

oleiade (approved):

Issue #2459, opened 2022-03-24
Move integration tests to a dedicated package


na--:
https://github.com/grafana/k6/pull/2412 and its preceding PRs refactored some cruft in `cmd/` and allowed us to safely and easily write full integration tests of k6. However, these tests are limited to `cmd/`, since the `globalTestState` and `globalState` types, as well as the helper functions and other types they depend on are not exported: https://github.com/grafana/k6/blob/6d7d732d8798b30904f5a75d779f93aaf7770450/cmd/root_test.go#L18-L30

As evidenced by https://github.com/grafana/k6/pull/2399#issuecomment-1077753466, it would be much better to further refactor `cmd/`, to export and maybe even move these types out of there into a new package or packages, so we can have a dedicated top-level `tests/` or `integration_tests/` package where we could organize these tests nicely.

olegbespalov:
Are we going to tackle this in the [v0.39.0](https://github.com/grafana/k6/milestone/23)? :thinking: 

I'm asking because if yes, I'd like to merge the https://github.com/grafana/k6/milestone/23 as it is (in [v0.38.0](https://github.com/grafana/k6/milestone/22)), and move the tests to a separate module in the next cycle.

na--:
Yeah, I think we should leave it for the next release, possibly start this cooldown. I think it's relatively urgent that we do it because a whole bunch of not-quite-ok integration tests will start piling up until we do, and so I added the `high prio` tag... But it's not so urgent that we need to drop things to get it done in this release, we already have plenty of integration test goodness for this one, even if it's not quite as user-friendly as I'd like :sweat_smile:

imiric:
@na-- I started looking into this, and have a couple of questions:

- Since the goal is to export `globalTestState` and related types, and move them to a separate package, this means we'd also need to export [`cmd.globalState`](https://github.com/grafana/k6/blob/24f9871266fd59d4079929c7c6d95454461049b5/cmd/root.go#L60) (since `testGlobalState` embeds it).

  But [this comment](https://github.com/grafana/k6/blob/24f9871266fd59d4079929c7c6d95454461049b5/cmd/root.go#L52-L55) mentions that we want to _prevent_ direct access to some of these fields from the rest of the codebase. So how do you envision this working?

- What other non-`cmd` integration tests do you have in mind that we should add?

  Currently these test types are only useful for `cmd` integration tests, which look fine to me being in `cmd/`.

[...]

PR #2787, opened 2022-11-30
Refactor console UI handling outside of the `cmd` package

imiric:
This is a refactor I started while working on #2459 (which I'm still working on based on this branch), and evolved into moving ~all~ most functionality that handles interacting with the console outside of the `cmd` package.

See the individual commits for details. Apologies for the large PR :relaxed: While I could make separate PRs for each commit, I think it makes more sense reviewed in the final state. Most of the changes are mechanical file moves, so it's mostly a noisy diff.

This involves:

- Creation of a new `ui/console` package and `Console` struct that encapsulates synchronized writing to stdout/stderr, and all previous `print*` functions.
  This decouples the `cmd` package from any console UI related behavior, which can be accessed via `globalState.console`.
  Another benefit is that handling of color output is mostly done in a single place, and we avoid the numerous `isTTY` and `noColor` checks that were in many places in `cmd` before.
- Consolidating most console related functionality into subpackages of `ui/console`. So `ui.Form` is now in `ui/console/form`, and `ui/pb` in `ui/console/pb`.
  This makes it clear that these packages are console-only, while freeing up the `ui` namespace to potentially add other UIs to k6 (resurrect the old web UI, maybe? :wink:).

There are some leftover things I'd like to address eventually, though. Primarily, [the coupling of executors with progress bars](https://github.com/grafana/k6/blob/564197a60714b52c571ae31edc839bc2cf762e4c/lib/executor/base_executor.go#L37-L40). It doesn't make sense for these low-level components to be interacting with a UI element directly. Even though `pb` is effectively standalone and decoupled from any OS dependencies, the way the rendering is done is very much specific to a console output. Instead, there should be an intermediate `Progress` type that executors can return, and `pb` can read. I didn't get into it here, as it would expand the scope even more, but I can create an issue for it if there's agreement that it should be done. Let me know.

Note that this **isn't** part of the v0.42.0 cycle, so it can be reviewed during the cooldown phase or later.


olegbespalov (commented):
Did the first round of review, generally looks good.

A bunch of the comments TODO cross-platform check :question:  me, are we going to do that before merging, right?

ui/console/console.go
```diff
@@ -0,0 +1,238 @@
...
+	termWidth := defaultTermWidth
+	if c.IsTTY {
+		tw, _, err := term.GetSize(int(c.Stdout.Fd()))
+		if !(tw > 0) || err != nil {
+			return termWidth, err
+		}
+		termWidth = tw
+	}
+
+	return termWidth, nil
```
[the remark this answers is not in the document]
imiric: Thanks, done in 22982999a.

ui/console/console.go
```diff
@@ -0,0 +1,238 @@
...
+	w.mutex.Unlock()
+
+	if err != nil && n < origLen {
+		return n, err
+	}
+	return origLen, err
+}
+
+// getColor returns the requested color with the given attributes.
+func getColor(attributes ...color.Attribute) *color.Color {
```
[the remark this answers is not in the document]
imiric: Right. Technically it's a constructor, so I renamed it to `newColor` in 22982999a.

cmd/cloud.go
```diff
@@ -64,13 +63,13 @@ func (c *cmdCloud) preRun(cmd *cobra.Command, args []string) error {
 // TODO: split apart some more
 //nolint:funlen,gocognit,cyclop
 func (c *cmdCloud) run(cmd *cobra.Command, args []string) error {
-	printBanner(c.gs)
+	c.gs.console.Printf("\n%s\n\n", c.gs.console.Banner())
```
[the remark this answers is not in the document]
imiric: I overlooked it, that's why :sweat_smile: Nice catch!

Fixed in 22982999a.

I'm against encapsulating the check, since it's more readable to have the logic that decides whether to do something (print the banner in this case) outside of any functions, unless the check is overly verbose or is used in many places. Previously, `printBanner()` did that check internally, but now it's clearer that the banner is only printed if `quiet` wasn't enabled.

In general, I'm not a fan of the pattern of calling a function that returns right away depending on some check, especially if that check can be done by the caller. Functions should do what it says on the tin; nothing more and nothing less. :slightly_smiling_face:

We could always have another `maybePrintBanner()` or `printBannerUnlessQuiet()` function, but I think that's a bit silly considering it only saves us 2 lines, and it's called from 2 places.

Notice that I also didn't encapsulate the actual printing of the banner, since calling `console.Printf()` directly is also clearer.

imiric:
Thanks for the review @olegbespalov!

> A bunch of the comments TODO cross-platform check :question: me, are we going to do that before merging, right?

That's all cruft from the existing implementation, which I'm not planning to touch here :sweat_smile: The state of our UI, how we print progressbars, etc., is still a mess, but refactoring that and addressing all those TODOs is a much larger and trickier task that what was done here.

ui/console/console.go
```diff
@@ -0,0 +1,238 @@
...
+func New(
+	stdout, stderr OSFileW, stdin OSFileR,
+	colorize bool, termType string,
+	signalNotify func(chan<- os.Signal, ...os.Signal),
+	signalStop func(chan<- os.Signal),
+) *Console {
+	outMx := &sync.Mutex{}
+	outCW := newConsoleWriter(stdout, outMx, termType)
+	errCW := newConsoleWriter(stderr, outMx, termType)
+	isTTY := outCW.isTTY && errCW.isTTY
```
imiric, line 43: Note that this mostly assumes that both stdout and stderr will either be a TTY or not together. I.e. that there won't be a case where stdout is a TTY, but stderr isn't. Previously we were checking this separately.

Is this a safe assumption to make? In that case, I might just pull out the check from `newConsoleWriter()` and do it only once here, for either stdout or stderr, and remove the `isTTY` field from `consoleWriter`.

WDYT?

cmd/cloud.go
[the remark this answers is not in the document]
olegbespalov: >I'm against encapsulating the check, since it's more readable to have the logic that decides whether to do something 

I see where you are coming from, but generally can't agree :sweat_smile: 

Looking at all usages of the Banner right now, we see everywhere the if statement, and considering it, that likely there will be no place that won't use the same logic (`maybePrintBanner`) having these `if` are unnecessary details. Moreover, it can lead to errors because the maintainer/contributor can easily forget to put in the additional check before placing the banner.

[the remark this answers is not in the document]
imiric: I prefer favoring clarity, even at the expense of verbosity or duplication, so I don't see the external check as an unnecessary detail. The only reason I forgot to add it in this case is because I removed the `printBanner()` function, and forgot to update this usage of it. Since it only happens in 2 places, I doubt this would happen again.

But since you feel strongly about it, I added a `maybePrintBanner` function to `cmd` in 6ca990ae3. I also noticed the same issue with printing the `cloud` progress bar, so also added `maybePrintBar`. Let me know if this addresses your concern.

I tested the behavior of `k6 cloud`, with and without `--quiet`, and it now matches the behavior on `master-next`.

olegbespalov (approved):
:+1:

oleiade (approved):
LGTM 👍🏻

cmd/integration_test.go
```diff
@@ -1090,13 +1079,7 @@ func TestAbortedByScriptInitError(t *testing.T) {
 	)
 	newRootCommand(ts.globalState).execute()
 
-	// FIXME: remove this locking after VU initialization accepts a context and
-	// is properly synchronized: currently when a test is aborted during the
-	// init phase, some logs might be emitted after the above command returns...
-	// see: https://github.com/grafana/k6/issues/2790
-	ts.outMutex.Lock()
```
na--, line 1097: Ah, I should have reviewed the PR :disappointed: This should not have been removed here - the fact that there was a data race was actually quite valuable! We want to get data races in tests when we don't expect unsynchronized access, this is super useful.

cmd/root_test.go
```diff
@@ -64,18 +66,70 @@ func TestMain(m *testing.M) {
...
+	stdOut, stdErr bufferStringer
 	loggerHook     *testutils.SimpleLogrusHook
 
 	cwd string
 
 	expectedExitCode int
 }
 
+// A thread-safe buffer implementation.
+type safeBuffer struct {
```
na--, line 88: As I wrote in the other comment, I think this is wrong. We don't want a safe and synchronized buffer here, we want the exact opposite. We want tests to fail with a data race when we have unsynchronized access to the stdout, e.g. if multiple things are concurrently writing there. Or, as it happened in the recent integration tests, if trying to read it after an integration test was supposedly done resulted in a data race because some goroutine wasn't done and was still trying to write. 

These are super valuable revelations we are **masking**, not actually fixing, by using locking here.

imiric: As mentioned on Slack, this synchronizes writing with reading, which is only useful for tests. I.e. we never need to inspect the contents of stdout in the real world, so this is unique for tests.

I don't think we're masking anything with this. We we're manually locking stdout before so we can read from it, and this simply pushes it down the stack so it happens transparently for the test.

na--: The fact is that whenever a test wants to inspect the contents of `stdout` or `stderr` when it knows they might be written to (i.e. in the middle of the test), it can manually lock them. But when it wants to inspect them after the test is supposed to be finished, as is the case for the majority of assertions in the integration tests, it's useful to access them directly, without locking. 

That way, if there is a data race, we will be able to see it, and it will most likely be a great signal that we have an actual bug. That is, we want the test to fail! For example, if some forgotten goroutine that is still active and logs stuff after the test is supposed to be finished, as was the case for the `FIXME` comment you deleted. Discovering that data race prompted me to open https://github.com/grafana/k6/issues/2790. And this bug was not fixed in this PR, so you shouldn't have deleted the `FIXME`, it was just masked by this `safeBuffer`. The bug was actually fixed by https://github.com/grafana/k6/pull/2800, which [deleted the `stdout` locking and `FIXME` comment](https://github.com/grafana/k6/pull/2800/files#diff-76b37580ac2865dc962ced55720a7e30bcb354318fee75d8610dd13a92ba2149L1093).

imiric: OK, understood. I saw the fix in #2800, but didn't think it was a big deal to handle it this way.

How do you want to proceed? Another reason I added `safeBuffer` is because the stdout mutex is no longer exposed to `GlobalState` after the changes here, so tests wouldn't have access to it. I also prefer not having to deal with this locking in the many tests here that inspect stdout while the test is running.

I won't have time to look into this today, but feel free to create a PR if you have a good idea for handling this.

na--: I've started working on a PR, but it's likely going to be ready after the holidays. 

> many tests here that inspect stdout while the test is running.

Are there really that many tests? There are 5 `outMutex.Lock()` calls that are removed here, one of which is the `FIXME` under my other comment, another is fixed by another PR, one seems unnecessary (in `TestPrometheusRemoteWriteOutput()`) and the last 2 are combined in a helper function by https://github.com/grafana/k6/pull/2800/commits/d5124d12bbefc028f8e8ff5481abfc492eb05dbc

So, by my reading, after we fix the actual bugs, there will be 1 or 2 necessary lockings the stdout mutex in the integration tests, used only in the few places where we actually care about the stdout contents mid-test.

Issue #2790, opened 2022-12-01
Propagate the real context when initializing VUs


na--:
Currently, VU initialization doesn't receive a context. The last time we propagate the context is here:
https://github.com/grafana/k6/blob/564197a60714b52c571ae31edc839bc2cf762e4c/core/local/local.go#L172-L173

But then we don't pass it to the layers below:
https://github.com/grafana/k6/blob/564197a60714b52c571ae31edc839bc2cf762e4c/core/local/local.go#L182
https://github.com/grafana/k6/blob/564197a60714b52c571ae31edc839bc2cf762e4c/core/local/local.go#L143
https://github.com/grafana/k6/blob/564197a60714b52c571ae31edc839bc2cf762e4c/js/runner.go#L115-L126
https://github.com/grafana/k6/blob/564197a60714b52c571ae31edc839bc2cf762e4c/js/bundle.go#L242-L247

And, at the end, we end up using `context.Background()` :disappointed: 
https://github.com/grafana/k6/blob/564197a60714b52c571ae31edc839bc2cf762e4c/js/bundle.go#L329

This is not ideal because we don't have any way of aborting the VU initialization if the user hits Ctrl+C or if one of the VUs fails during the concurrent VU initialization and we need to abort the initialization of the rest of the VUs and wait for that to finish quickly (see [here](https://github.com/grafana/k6/blob/564197a60714b52c571ae31edc839bc2cf762e4c/core/local/local.go#L172-L203) and [here](https://github.com/grafana/k6/blob/564197a60714b52c571ae31edc839bc2cf762e4c/core/local/local.go#L286-L293)).

Currently, we just rely on the fact that we'll use `os.Exit()` to stop k6, however that assumption breaks in a few cases:
- in integration tests, where these dangling VU initializations can emit logs and cause data races, after the test was supposed to be finished (https://github.com/grafana/k6/pull/2789) 
- in test suites (https://github.com/loadimpact/k6/issues/1342), where the fact that one test fails may not necessarily mean that the k6 process should exit
- when the [`--linger` option](https://k6.io/docs/using-k6/k6-options/reference/#linger) is used, we might want to have the k6 process linger even if there was an init error, same as we'd do if the test was aborted by threshold or `execution.test.abort()`

In general, the current behavior breaks the very reasonable expectation that once the `k6 run` command has finished and returned, all of its child goroutines have also finished and returned before it...

PR #457, opened 2018-01-14
Setup & Teardown

liclac:
```js
export function setup() {
    return {v: 1};
}

export default function(data) {
    console.log(JSON.stringify(data));
}

export function teardown(data) {
    if (data.v != 1) {
        throw new Error("incorrect data: " + JSON.stringify(data));
    }
}
```

Both `setup()` and `teardown()` are ran in full VU contexts, and can thus do things like network requests just like you can in `default()`.

Pass `k6 run --no-setup --no-teardown ...` to skip running them; for clustered execution, etc.

Closes #194

Issue #2562, opened 2022-06-11
Wasm support. Also js —> wasm


gedw99:
### Feature Description

It would be cool to also support wasm for those that don’t want to us js. 

There is also the ability to transpire js to wasm now. The suborbital project got it working and others are now using the package to do it .

### Suggested Solution (optional)

https://github.com/suborbital/javy

gedw99:
Debugging js and wasm would be related to this. 

Wazero are working partially on wasm debugging . Wazero is pure golang so no CGO :) 

goja has debugging support . Not sure if that’s exposed in k6 main branch yet ?

na--:
Previously discussed here: https://github.com/grafana/k6/issues/751#issuecomment-596525404

See some of my comments in that issue, they apply as much for any future WebAssembly support in k6 as they do for potentially supporting Go or some other language. Admittedly, WASM will probably be simpler than some of the others, but it will still not going to be easy to do or maintain in the long-run.

codefromthecrypt:
You may also have noticed language compiler docs aren't well written or written for backend stuff. wazero is starting to compile our notes here https://wazero.io/languages/

We also cut our first beta tag (v1.0.0-beta.1) and also opened a [gophers slack](https://gophers.slack.com/) `#wazero` channel for support, updates and conversation! Note: You may need an [invite](https://invite.slack.golangbridge.org/) to join gophers.

Hope to see you around!

mstoykov:
After some internal discussion and evaluation, the team doesn't think this will be a thing we are unlikely to go ahead with adding WASM support anytime soon.

As such I will close this issue and potentially open another one in the future if the situation changes.

PR #2791, opened 2022-12-02
Start to run `setup()` and `teardown()` only if they were defined

na--:
Previous to this PR, k6 would initialize a transient VU and try to execute `setup()` and `teardown()` even if we knew they were not defined and exported by the script. That VU was created and then immediately discarded afterwards. 

It was probably done like that because, when `setup()` and `teardown()` were implemented (https://github.com/grafana/k6/pull/457) in k6 v0.20.0, we didn't keep track of what JS functions were exported from the script. We only started doing that in k6 v0.27.0, when we added support for multiple scenarios, each with a potentially different `exec` function. We just didn't go back and improve the `setup()` and `teardown()` efficiency afterwards and they continued to use the "try to run it and see if it works" approach.

While it was inefficient, it didn't use to be a big problem before. However, now that we want VU initialization to be interruptible by cancelling a context (https://github.com/grafana/k6/issues/2790), it became a problem for fine-grained unit tests. Because the [`setupTimeout` and `teardownTimeout` default values are defined in the `cmd/` package](https://github.com/grafana/k6/blob/70fa2b8d2884c86976f18e43f7074f1a7a8f645c/cmd/config.go#L217-L222), tests in sub-packages like `js/` and `core/` saw their values as `0` :sob: So, they created contexts with `0` timeouts, which immediately expired and the transient VUs for `setup()` and `teardown()` couldn't even be initialized without an error. Our config mess (#883) strikes again...

I thought about fixing the config somehow, but actually not running the `setup()` and `teardown()` functions if they were not defined seems both simpler and cleaner. Any low-level test that wants to use them can just manually define their timeouts, which most do and is usually a good idea anyway (the default 60s value is usually too much for tests).

### DIFF
--- a/js/runner.go
+++ b/js/runner.go
@@ -274,6 +274,14 @@ func forceHTTP1() bool {
 
 // Setup runs the setup function if there is one and sets the setupData to the returned value
 func (r *Runner) Setup(ctx context.Context, out chan<- metrics.SampleContainer) error {
+	if !r.IsExecutable(consts.SetupFn) {
+		// do not init a new transient VU or execute setup() if it wasn't
+		// actually defined and exported in the script
+		r.preInitState.Logger.Debugf("%s() is not defined or not exported, skipping!", consts.SetupFn)
+		return nil
+	}
+	r.preInitState.Logger.Debugf("Running %s()...", consts.SetupFn)
+
 	setupCtx, setupCancel := context.WithTimeout(ctx, r.getTimeoutFor(consts.SetupFn))
 	defer setupCancel()
 
@@ -307,6 +315,14 @@ func (r *Runner) SetSetupData(data []byte) {
 
 // Teardown runs the teardown function if there is one.
 func (r *Runner) Teardown(ctx context.Context, out chan<- metrics.SampleContainer) error {
+	if !r.IsExecutable(consts.TeardownFn) {
+		// do not init a new transient VU or execute teardown() if it wasn't
+		// actually defined and exported in the script
+		r.preInitState.Logger.Debugf("%s() is not defined or not exported, skipping!", consts.TeardownFn)
+		return nil
+	}
+	r.preInitState.Logger.Debugf("Running %s()...", consts.TeardownFn)
+
 	teardownCtx, teardownCancel := context.WithTimeout(ctx, r.getTimeoutFor(consts.TeardownFn))
 	defer teardownCancel()
 
--- a/cmd/integration_test.go
+++ b/cmd/integration_test.go
@@ -108,7 +108,7 @@ func TestStdoutAndStderrAreEmptyWithQuietAndLogsForwarded(t *testing.T) {
 	// Instead it should be in the log file
 	logContents, err := afero.ReadFile(ts.fs, logFilePath)
 	require.NoError(t, err)
-	assert.Equal(t, "init\ninit\ninit\nfoo\ninit\n", string(logContents)) //nolint:dupword
+	assert.Equal(t, "init\ninit\nfoo\n", string(logContents)) //nolint:dupword
 }
 
 func TestRelativeLogPathWithSetupAndTeardown(t *testing.T) {
--- a/cmd/integration_test.go
+++ b/cmd/integration_test.go
@@ -29,16 +29,6 @@ import (
 	"go.k6.io/k6/lib/testutils/httpmultibin"
 )
 
-const (
-	noopDefaultFunc   = `export default function() {};`
-	fooLogDefaultFunc = `export default function() { console.log('foo'); };`
-	noopHandleSummary = `
-		export function handleSummary(data) {
-			return {}; // silence the end of test summary
-		};
-	`
-)
-
 func TestVersion(t *testing.T) {
 	t.Parallel()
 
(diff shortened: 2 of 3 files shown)


na--:
The flaky test is fixed in https://github.com/grafana/k6/pull/2792, will rebase after that gets merged.

olegbespalov (dismissed):

js/runner.go
```diff
@@ -274,6 +274,13 @@ func forceHTTP1() bool {
 
 // Setup runs the setup function if there is one and sets the setupData to the returned value
 func (r *Runner) Setup(ctx context.Context, out chan<- metrics.SampleContainer) error {
+	if !r.IsExecutable(consts.SetupFn) {
+		// do not init a new transient VU or execute setup() if it wasn't
+		// actually defined and exported in the script
+		r.preInitState.Logger.Debugf("%s() is not defined or not exported, skipping...", consts.SetupFn)
```
imiric, line 280: Like I mentioned on Slack, I'd rather move this check to where `Runner.Setup()` is called. There's no point in calling a function that will no-op, if the same check could be done by the caller. Besides being straightforward to read, we wouldn't end up with this confusing log:

```
DEBU[0000] Running setup()                               phase=local-execution-scheduler-run
DEBU[0000] setup() is not defined or not exported, skipping...
```

If `setup()` is not exported or defined, what are we running...?

Additionally, does it make sense to [change the execution status and progressbar](https://github.com/grafana/k6/blob/56fcad04e33db6359322f57534be474c861b3932/core/local/local.go#L424-L425) if we don't end up running it?

Also, wouldn't returning `nil` here be a minor breaking change for the HTTP API? [We run `Setup()` on `POST /v1/setup`](https://github.com/grafana/k6/blob/56fcad04e33db6359322f57534be474c861b3932/api/v1/setup_teardown_routes.go#L78), and previously we would return a 500 error if `setup` is not found, whereas now it returns a 200 without actually running it. It's questionable whether 500 is a good fit (a 4xx code might be more suitable), but a 2xx definitely wouldn't make sense if we didn't run anything. I don't think this would actually break anything, but it's one more reason to not do this here.

So I would definitely prefer that this check was added [here](https://github.com/grafana/k6/blob/56fcad04e33db6359322f57534be474c861b3932/core/local/local.go#L422) instead. If we want to keep this log, then it should be done from there as well. Though I wonder if we need it, since we also don't log if `setup()` is not run because of `--no-setup`.

na--: > Also, wouldn't returning `nil` here be a minor breaking change for the HTTP API? [We run `Setup()` on `POST /v1/setup`](https://github.com/grafana/k6/blob/56fcad04e33db6359322f57534be474c861b3932/api/v1/setup_teardown_routes.go#L78), and previously we would return a 500 error if `setup` is not found, whereas now it returns a 200 without actually running it.

Did you check that? Because it shouldn't have returned 500 if there was no `setup()` before, `runPart()` should have returned `goja.Undefined`: https://github.com/grafana/k6/blob/70fa2b8d2884c86976f18e43f7074f1a7a8f645c/js/runner.go#L499-L502
And `js.Runner.Setup()` should have just returned `nil` for an error: https://github.com/grafana/k6/blob/70fa2b8d2884c86976f18e43f7074f1a7a8f645c/js/runner.go#L284-L288

500 error in the `POST /v1/setup` when no there wasn't as `setup()` function would be a bug.

na--: To respond to the rest of the comment, I have two problems with checking for this in the `ExecutionScheduler`. The first one is that it will make an already long and complex method even longer and more complex.

The second and bigger problem is that, conceptually, it's not any of the `ExecutionScheduler`'s concern how a "setup" call is implemented. The `lib.Runner` interface, flawed as it is, doesn't expose that information, it just has this method (notice the comment): https://github.com/grafana/k6/blob/70fa2b8d2884c86976f18e43f7074f1a7a8f645c/lib/runner.go#L59-L60

The `js.Runner` implementation expects the script to has an exported method called `setup()` that behaves in a specific way. The `testutils.MiniRunner` implementation of the same interface, however, just has this: https://github.com/grafana/k6/blob/70fa2b8d2884c86976f18e43f7074f1a7a8f645c/lib/testutils/minirunner/minirunner.go#L23

Another future implementation (e.g. lua, https://github.com/grafana/k6/issues/751, https://github.com/grafana/k6/issues/2562, whatever) could have a completely different way of handling "setup".

So, my suggestion would actually be to move this log to the JS runner, so there are no conflicting logs: https://github.com/grafana/k6/blob/70fa2b8d2884c86976f18e43f7074f1a7a8f645c/core/local/local.go#L423

na--: The [latest force-push](https://github.com/grafana/k6/compare/876e8174d8f4de119de69656f2604316e54c07f5..10c7581aca2d1255ad8efeba1670d0f3fe70068f) cleans up the conflicting logging.

imiric: Ah, I misread the `runPart()` handling of this. :facepalm: And no, I didn't actually test it. You're right, it wouldn't change. :+1:

I don't quite see that this additional check would add much complexity to `ExecutionScheduler.Run()`, but see your point about it being conceptually specific to the JS implementation, however far away those alternative implementations might be. I suppose that would be the argument for changing the execution status and progress bar as well(?).

Yeah, the logging makes more sense in `Setup()` then.

imiric (approved):

olegbespalov (approved):

PR #2795, opened 2022-12-04
Increase and improve integration tests

na--:
I've tried to make the commits in this PR very self-contained, so it might be easier to review them one by one. However, given that the only changes are in tests, it shouldn't matter too much either way.

The PR copies (with improvements) some of the more useful `core.Engine` tests, in preparation for https://github.com/grafana/k6/issues/1889. It also makes some minor fixes and improvements to the integration test framework and adds rudimentary support (and a few tests that) for verifying metrics sent to the JSON output, for extra assurance. It doesn't have any functional changes, so it should be safe to merge as soon as it's reviewed.

### DIFF
--- a/core/engine_test.go
+++ b/core/engine_test.go
@@ -1231,6 +1231,7 @@ func TestEngineRunsTeardownEvenAfterTestRunIsAborted(t *testing.T) {
 	assert.Equal(t, 1.0, count)
 }
 
+// TODO: delete, functionality duplicated in cmd/integration_test.go
 func TestActiveVUsCount(t *testing.T) {
 	t.Parallel()
 
--- a/core/engine_test.go
+++ b/core/engine_test.go
@@ -1108,6 +1108,7 @@ func TestMetricsEmission(t *testing.T) {
 	}
 }
 
+// TODO: delete, functionality duplicated in cmd/integration_test.go
 func TestMinIterationDurationInSetupTeardownStage(t *testing.T) {
 	t.Parallel()
 	setupScript := `
--- a/core/engine_test.go
+++ b/core/engine_test.go
@@ -1188,6 +1188,7 @@ func TestMinIterationDurationInSetupTeardownStage(t *testing.T) {
 	}
 }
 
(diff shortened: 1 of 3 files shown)


core/engine_test.go
```diff
@@ -720,6 +720,7 @@ func TestSentReceivedMetrics(t *testing.T) {
 	})
 }
 
+// TODO: delete, functionality duplicated in cmd/integration_test.go
```
olegbespalov, line 723: Here and in other places marked with such TODOs? Why should we keep it, then? :thinking:

olegbespalov (commented):
LGTM, but if we can clean up the test, then we no longer need it will be good since this PR is probably the best time for it.

core/engine_test.go
na--: I kind of wanted to delete the old `Engine` tests in the PR that removes the whole `Engine`, the one that closes https://github.com/grafana/k6/issues/1889. 

Since the integration tests and the ones they replace in `core/engine_test.go` are not _exactly_ perfectly aligned, I don't see the harm in keeping the old ones until then, for a tiny bit of extra assurance. I plan to submit at least 2 more PRs before I submit the one with the `Engine` removal, one for v0.42.0 with another bugfix, and another for the start of v0.43.0, and https://github.com/grafana/k6/pull/2787 already exists for v0.43.0 merging, so having these extra tests for the current behavior until then seems like a good idea.

olegbespalov: Okay, then I believe we need to improve the comment again because what you're saying makes sense, but it's not clear from the current comment:
 
```
// TODO: delete, functionality duplicated in cmd/integration_test.go
```

From my point of view, even the references to the PR/issues could improve it.

```
// TODO: delete (#1889), functionality duplicated in cmd/integration_test.go
```

na--: Resolved by https://github.com/grafana/k6/pull/2795/commits/3e73e8783b13e7f82e3ce23a3c7b907f2e536e72

olegbespalov (approved):

codebien (approved):
LGTM

PR #2796, opened 2022-12-05
Simplify tests by moving some boilerplate init code to a helper

na--:


cmd/integration_test.go
```diff
@@ -461,6 +461,25 @@ func getCloudTestEndChecker(t *testing.T, expRunStatus lib.RunStatus, expResultS
 	return srv
 }
 
+func getSimpleCloudOutputTestState(
+	t *testing.T, script []byte, cliFlags []string,
+	expRunStatus lib.RunStatus, expResultStatus cloudapi.ResultStatus, expExitCode int,
+) *globalTestState {
+	srv := getCloudTestEndChecker(t, expRunStatus, expResultStatus)
```
olegbespalov, line 468: Maybe also marks this as a helper, WDYT? 

```suggestion
	t.Helper()
	
	srv := getCloudTestEndChecker(t, expRunStatus, expResultStatus)
```

olegbespalov (approved):

cmd/integration_test.go
na--: I find `t.Helper()` to be unhelpful when debugging test errors, it just hides error information that makes it harder to diagnose the root cause. Especially when "helpers" themselves make assertions and can fail the test, it is generally more harmful than helpful to use it.

cmd/integration_test.go
```diff
@@ -927,13 +912,10 @@ func TestMetricTagAndSetupDataIsolation(t *testing.T) {
...
-	srv := getCloudTestEndChecker(t, lib.RunStatusFinished, cloudapi.ResultStatusPassed)
-
-	ts := newGlobalTestState(t)
-	require.NoError(t, afero.WriteFile(ts.fs, filepath.Join(ts.cwd, "test.js"), script, 0o644))
-	ts.envVars = map[string]string{"K6_CLOUD_HOST": srv.URL}
-	ts.args = []string{"k6", "run", "--quiet", "--log-output=stdout", "--out", "cloud", "test.js"}
-
+	ts := getSimpleCloudOutputTestState(
+		t, script, []string{"--quiet", "--log-output", "stdout"},
+		lib.RunStatusFinished, cloudapi.ResultStatusPassed, 0,
```
codebien, line 917:
```suggestion
		lib.RunStatusFinished, cloudapi.ResultStatusPassed, exitcodes.Succeed,
```
Probably this isn't the PR for adding it, but WDYT to open an issue?

codebien (approved):

cmd/integration_test.go
na--: :thinking: maybe, but yeah, definitely in a different PR

```suggestion
		lib.RunStatusFinished, cloudapi.ResultStatusPassed, int(exitcodes.Success),
```

And, if we add an alias to the `0` exit code in our constants, then we should probably also fix this TODO I recently added here? https://github.com/grafana/k6/blob/65a75982ad85e0e0f8a30e05c614446b14e59f81/cmd/root.go#L255-L261

PR #1390, opened 2020-04-09
Refactor the Engine to process metrics independently from the test Run

na--:
This allows metric processing to still occur, even after the script run has completed. In practice, this allows you to run k6 with the `--linger --no-teardown` flags, and execute the `teardown()` function manually through the REST API, after the main test run has completed.

This also fixes https://github.com/loadimpact/k6/issues/1369 - now a second Ctrl+C will _always_ abort k6, albeit in a very abrupt way. On the other hand, the first Ctrl+C would be more graceful than it currently is, since it will now allow `teardown()` to be executed.

PR #2798, opened 2022-12-05
Wait for metrics handling to finish even when there is an error

na--:
Before these changes, `setup()` and `teardown()` exceptions or `test.abort()` calls would have made k6 exit prematurely, without waiting for the metrics processing to finish. **This bug was probably a cause for some of the "timed out" `k6 run --out cloud` tests we have seen, e.g. if they had some sort an error in `setup()` or `teardown()`.** It also caused a data race in the recent integration tests, this is how I caught it. So, as a nice bonus, this fix allowed me to remove one of the `FIXME`s for unnecessary locking in there earlier than expected :tada: 

This is another thing that should be fixed by the new `Engine`-less architecture (https://github.com/grafana/k6/issues/1889). However, once again, the fix was easy enough to do even now, and the bug severe enough that it's probably worth it to include it in the upcoming k6 v0.42.0. 

I also added and expanded a bunch of integration tests, particularly when it comes to validating aborted tests and `--linger`, and this PR is built on top of https://github.com/grafana/k6/pull/2795, so hopefully its last minute merging won't cause any regressions :crossed_fingers: :pray: :prayer_beads: .

It seemed like the bug might have been introduced in k6 v0.36.0, via https://github.com/grafana/k6/pull/2093 (the PR that added `test.abort()`), but it was actually probably introduced way back in k6 v0.27.0, with https://github.com/grafana/k6/pull/1390, or even older... :disappointed: Still, the added tests for `test.abort()` and setup/teardown exceptions should provide some assurance now and in the future.

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -185,35 +185,27 @@ func (c *cmdRun) run(cmd *cobra.Command, args []string) error {
 
 	// Start the test run
 	initBar.Modify(pb.WithConstProgress(0, "Starting test..."))
-	var interrupt error
 	err = engineRun()
 	if err != nil {
-		err = common.UnwrapGojaInterruptedError(err)
-		if errext.IsInterruptError(err) {
-			// Don't return here since we need to work with --linger,
-			// show the end-of-test summary and exit cleanly.
-			interrupt = err
-		}
-		if !conf.Linger.Bool && interrupt == nil {
-			return errext.WithExitCodeIfNone(err, exitcodes.GenericEngine)
-		}
+		logger.WithError(err).Debug("Engine terminated with an error")
+	} else {
+		logger.Debug("Engine run terminated cleanly")
 	}
 	runCancel()
-	logger.Debug("Engine run terminated cleanly")
 
 	progressCancel()
 	progressBarWG.Wait()
 
 	executionState := execScheduler.GetState()
 	// Warn if no iterations could be completed.
-	if executionState.GetFullIterationCount() == 0 {
+	if err == nil && executionState.GetFullIterationCount() == 0 {
 		logger.Warn("No script iterations finished, consider making the test duration longer")
 	}
 
 	// Handle the end-of-test summary.
 	if !testRunState.RuntimeOptions.NoSummary.Bool {
 		engine.MetricsEngine.MetricsLock.Lock() // TODO: refactor so this is not needed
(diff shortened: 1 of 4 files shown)


oleiade (dismissed):
Looks good to me 👍🏻

codebien (approved):

olegbespalov (approved):

cmd/run.go
```diff
@@ -185,35 +185,27 @@ func (c *cmdRun) run(cmd *cobra.Command, args []string) error {
...
 	executionState := execScheduler.GetState()
 	// Warn if no iterations could be completed.
-	if executionState.GetFullIterationCount() == 0 {
+	if err == nil && executionState.GetFullIterationCount() == 0 {
 		logger.Warn("No script iterations finished, consider making the test duration longer")
 	}
 
 	// Handle the end-of-test summary.
-	if !testRunState.RuntimeOptions.NoSummary.Bool {
+	if err == nil && !testRunState.RuntimeOptions.NoSummary.Bool {
```
imiric, line 206: Do we not want to output the summary even if `test.abort()` was called? We currently do on `master` and it would be useful to output whatever data we got before the test was aborted.

na--: Great catch! This is a regression from this PR, I'll fix it and add it to the test.

That said, I'd say that the current `master` code also has a bug. With it, this example also won't produce a summary, which is probably not what we want either:
```js
import http from 'k6/http';

export default function () {
    http.get('https://test.k6.io');
};

export function teardown() {
    console.log('wonky teardown');
    //throw new Error('foo');
}
``` 

So, I'll adjust the code so that k6 tries to print the `handleSummary()` even if there was an error during the test run. The only time `handleSummary()` won't be called is when it was explicitly disabled with `--no-summary`, or when the init code wasn't successful (because that probably means there is a script error in the init context and k6 won't be able to initialize a transient VU to run `handleSummary()` anyway)

WDYT?

codebien: A bit bad that no test failed for this. Could we move the logic as a method of `testRunState` so we could centralize it and cover it with a unit test?

imiric (approved):
LGTM, nice job with the tests :+1:

CI didn't run for some reason after your latest push :confused:

na--:
Weird :confused: force-pushed again, with no changes besides the commit message, and CI seems to have started fine this time :crossed_fingers:

imiric (approved):

PR #2800, opened 2022-12-05
Pass and respect a `Context` when initializing VUs

na--:
This closes https://github.com/grafana/k6/issues/2790. It is not very urgent, it doesn't need to be released in v0.42.0, instead it is meant to be merged some time early in the v0.43.0 cycle.

Most of the LoC changes are just boilerplate churn, mostly test and API signature adjustments. They have been grouped together into the first 2 commits of the PR and should contain no functional or test changes.

The important changes are focused in the next 2 commits, which should be the reviewed carefully. I may add some more tests there, though I am not sure what they should be :thinking: If anyone has any ideas, please share :pray:


imiric (commented):
LGTM from a first look, will make another pass later.

Nice job splitting it into atomic commits, makes reviewing much easier :thumbsup:

As for additional tests, and maybe I'm missing it, but I think there should be a test replicating the user pressing `^C` during VU initialization. I see a couple of tests sending `os.Interrupt`, but it happens while the test is running. The tests aborted in init should cover the same code paths, but receiving `SIGINT` should trigger it as well.

mstoykov (approved):

js/initcontext.go
```diff
@@ -70,7 +70,7 @@ func NewInitContext(
 		moduleRegistry:    getJSModules(),
 		exportsCache:      make(map[string]goja.Value),
 		moduleVUImpl: &moduleVUImpl{
-			ctx:     context.Background(),
+			ctx:     context.Background(), // TODO: pass a real context here as well?
```
codebien, line 73:
```suggestion
			ctx:     context.Background(), // TODO: pass a real context as we did for https://github.com/grafana/k6/pull/2800
```
It think it should be even better if we open a dedicated issue. Or does one of the next PR  resolve it?

js/runner.go
```diff
@@ -354,7 +358,7 @@ func (r *Runner) IsExecutable(name string) bool {
 }
 
 // HandleSummary calls the specified summary callback, if supplied.
-func (r *Runner) HandleSummary(ctx context.Context, summary *lib.Summary) (map[string]io.Reader, error) {
+func (r *Runner) HandleSummary(parentCtx context.Context, summary *lib.Summary) (map[string]io.Reader, error) {
```
codebien, line 361:
```suggestion
func (r *Runner) HandleSummary(ctx context.Context, summary *lib.Summary) (map[string]io.Reader, error) {
```

Why do you think is it better to rename it in this way? From a caller's point of view is it relevant? If yes, then we should document it. Otherwise, I think we should reverse the pattern and rename `s/ctx/summaryCtx/`.

na--:
@imiric, I was super surprised that a test didn't exist, and when I started to add it, I realized why I hadn't done it originally :sweat_smile:

While this PR makes it so VU initializations are going to be gracefully aborted with Ctrl+C, true. However, k6 won't return the right exit code or send the `run_status` value yet... :disappointed: IIRC, this is when I started writing https://github.com/grafana/k6/issues/2804 :sweat_smile: 

Still, I decided to go ahead and add a test anyway, even if it needs to check the current values of these... I just added a couple of `TODO`s, so we know the test doesn't verify the desired behavior, but the current one. 

So, I cherry-picked one commit (https://github.com/grafana/k6/pull/2800/commits/d5124d12bbefc028f8e8ff5481abfc492eb05dbc) from https://github.com/grafana/k6/pull/2807, to simplify the integration tests that use Ctrl+C or REST API to stop k6, and then added https://github.com/grafana/k6/pull/2800/commits/ed3949592f6dbfc2786681d211f3996743340f65 with the actual test to check what happens when an interrupt signal is given while VUs are initializing.

js/bundle.go
```diff
@@ -326,11 +332,24 @@ func (b *Bundle) instantiate(logger logrus.FieldLogger, rt *goja.Runtime, init *
...
 
+	// TODO: make something cleaner for interrupting scripts, and more unified
+	// (e.g. as a part of the event loop or RunWithPanicCatching()?
+	initCtxDone := init.moduleVUImpl.ctx.Done()
+	initDone := make(chan struct{})
+	watchDone := make(chan struct{})
+	go func() {
+		select {
+		case <-initCtxDone:
+			rt.Interrupt(context.Canceled)
```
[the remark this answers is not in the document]
na--: Resolved by the [latest force-push](https://github.com/grafana/k6/compare/dbde3c58530c7bc3b437dc665c725a859ceb6c9a..ed3949592f6dbfc2786681d211f3996743340f65#diff-663d689366c7fca665dcbb0e54a8401fbcd724f3ec4bce700c306a82926da9b9R346-R348).

js/bundle.go
```diff
@@ -335,6 +336,20 @@ func (b *Bundle) instantiate(init *InitContext, vuID uint64) (err error) {
...
 
+	// TODO: make something cleaner for interrupting scripts, and more unified
+	// (e.g. as a part of the event loop or RunWithPanicCatching()?
+	initCtxDone := init.moduleVUImpl.ctx.Done()
+	initDone := make(chan struct{})
+	watchDone := make(chan struct{})
+	go func() {
+		select {
+		case <-initCtxDone:
+			rt.Interrupt(init.moduleVUImpl.ctx.Err())
```
codebien, line 347: I need to check what the logic is in the other cases (not init) but if it is the same maybe could we delegate this listener directly in moduleVUImpl? :thinking: 

If the logic is different, we should use this opportunity to wrap the error so we know it is generated by the init process.

codebien (commented):
Thanks for doing it, I saw very nice improvements. :bow: I was reviewing when you force-pushed so something could be lost on my side.

js/initcontext.go
na--: None of my chain of PRs resolve this, I didn't have time and it's not super straightforward to do so, there are some chicken-and-egg problems and some tricky things to consider when it comes to test suites (https://github.com/grafana/k6/issues/1342). 

Still, it is necessary to solve that before https://github.com/grafana/k6/issues/2804 is implemented, so I'd rather add a link to both that issue and the #2800 PR as well :+1:

js/runner.go
na--: It is not better, but we need to add a sub-context inside of the method, so I chose to rename the parameter to `parentCtx`, to be able to create a `ctx` inside without overwriting this value.

js/bundle.go
na--: Unfortunately it wasn't straightforward to unify this init interrupt logic with the rest of the test abort logic during test execution :disappointed: At least I didn't see how to do that without a much, much  bigger refactoring of the whole `js/` package... :disappointed: 

So I chose to make only minor improvements in this PR (see the [`Simplify the js/Bundle.instantiate() API`](https://github.com/grafana/k6/pull/2800/commits/e076d4001dcd65d4b0473b98f07a7662eee9a47b) commit) and leave the rest for a future PR (see the TODO in the code snippet you've selected).

js/initcontext.go
na--:
```suggestion
			// TODO: pass a real context as we did for https://github.com/grafana/k6/pull/2800,
			// also see https://github.com/grafana/k6/pull/2800
			ctx:     context.Background(), 
```

js/initcontext.go
```diff
@@ -70,7 +70,9 @@ func NewInitContext(
 		moduleRegistry:    getJSModules(),
 		exportsCache:      make(map[string]goja.Value),
 		moduleVUImpl: &moduleVUImpl{
-			ctx:     context.Background(), // TODO: pass a real context here as well?
+			// TODO: pass a real context as we did for https://github.com/grafana/k6/pull/2800,
+			// also see https://github.com/grafana/k6/pull/2800
```
codebien, line 74: also see https://github.com/grafana/k6/pull/2804

js/runner.go
codebien: >but we need to add a sub-context inside of the method

I got that, I'm suggesting to reverse it selecting a specific name for the sub-context instead

js/bundle.go
codebien: >If the logic is different, we should use this opportunity to wrap the error so we know it is generated by the init process.

What about this, is it doable?

codebien (changes requested):

js/initcontext.go
na--: :man_facepalming: :sweat_smile: :+1:

cmd/integration_test.go
```diff
@@ -669,46 +666,80 @@ func TestAbortedByUserWithGoodThresholds(t *testing.T) {
...
 
-		reachedIteration = true
+	asyncWaitForStdoutAndRun(t, ts, attempts, interval, expText, func() {
+		t.Log("expected stdout text was found, sending interrupt signal...")
+		sendSignal <- struct{}{}
+		<-sendSignal
+	})
+}
+
+func asyncWaitForStdoutAndStopTestFromRESTAPI(
```
imiric, line 739: This function is only called from `TestAbortedByUserWithRestAPI`, and just wraps another helper function. So I would inline the `asyncWaitForStdoutAndRun()` call, and remove this.

`asyncWaitForStdoutAndRun()` is also called with the same arguments, except for the callback func. So I would drop the unneded arguments, and add them once they're needed.

In general, abstractions add a layer of indirection that complicates code comprehension, so I wouldn't use them so liberally, and would prefer even some duplication between tests. Or, if tests really are similar, turn them into table tests and avoid the helper functions.

imiric (approved):
LGTM :+1:

Just left a minor comment about test helpers.

js/bundle.go
na--: Hmm not sure what additional information we can add here, as the rest of the code is now :thinking: I think it will be much easier to fix this after we have the `NewTestRunContext()` in a "higher" place, i.e. after https://github.com/grafana/k6/pull/2813 :thinking:

cmd/integration_test.go
na--: I agree and I was about to fix this, but I think the fix here is to just add more tests that abort the test run via the k6 REST API :sweat_smile: See https://github.com/grafana/k6/issues/2804, we need more tests about this behavior... :disappointed: 

In https://github.com/grafana/k6/pull/2815 or soon after it, we should be able to nicely abort the test with the REST even during VU init or `setup()`, so I'd prefer to keep this helper method here and add more tests in that PR.

js/initcontext.go
na--: Resolved by the [latest force-push](https://github.com/grafana/k6/compare/9c9e35f63ef955140aff6ac122867832a249c7d8..a7ad53fef43a3a7885df96e08afab88c83d24436)

js/runner.go
na--: Resolved by the [latest force-push](https://github.com/grafana/k6/compare/9c9e35f63ef955140aff6ac122867832a249c7d8..a7ad53fef43a3a7885df96e08afab88c83d24436)

codebien (approved):
LGTM

mstoykov (approved):

PR #2803, opened 2022-12-05
Ensure REST API server is shut down after test ends and check for goroutine leaks

na--:
This depends on https://github.com/grafana/k6/pull/2800

### DIFF
--- a/cmd/root_test.go
+++ b/cmd/root_test.go
@@ -19,6 +19,7 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"go.k6.io/k6/lib/testutils"
+	"go.uber.org/goleak"
 )
 
 type blockingTransport struct {
@@ -59,7 +60,15 @@ func TestMain(m *testing.M) {
 		}
 	}()
 
-	// TODO: add https://github.com/uber-go/goleak
+	defer func() {
+		// TODO: figure out why logrus' `Entry.WriterLevel` goroutine sticks
+		// around and remove this exception.
+		opt := goleak.IgnoreTopFunction("io.(*pipe).read")
+		if err := goleak.Find(opt); err != nil {
+			fmt.Println(err) //nolint:forbidigo
+			exitCode = 3
+		}
+	}()
 
 	exitCode = m.Run()
 }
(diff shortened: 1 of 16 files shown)


cmd/run.go
```diff
@@ -130,6 +140,13 @@ func (c *cmdRun) run(cmd *cobra.Command, args []string) error {
 				}
 			}
 		}()
+		go func() {
+			defer apiWG.Done()
+			<-srvCtx.Done()
+			if aerr := srv.Close(); aerr != nil {
```
codebien, line 146: :thinking: Why directly Close and not a graceful Shutdown?

cmd/root_test.go
```diff
@@ -59,7 +60,15 @@ func TestMain(m *testing.M) {
...
-	// TODO: add https://github.com/uber-go/goleak
+	defer func() {
+		// TODO: figure out why logrus' `Entry.WriterLevel` goroutine sticks
+		// around and remove this exception.
+		opt := goleak.IgnoreTopFunction("io.(*pipe).read")
+		if err := goleak.Find(opt); err != nil {
+			fmt.Println(err) //nolint:forbidigo
+			exitCode = 3
+		}
+	}()
```
codebien, line 71: I help here, I will take a look into it.

cmd/run.go
```diff
@@ -117,10 +117,20 @@ func (c *cmdRun) run(cmd *cobra.Command, args []string) error {
...
 		initBar.Modify(pb.WithConstProgress(0, "Init API server"))
+
+		apiWG := &sync.WaitGroup{}
+		apiWG.Add(2)
+		defer apiWG.Wait()
+
+		srvCtx, srvCancel := context.WithCancel(globalCtx)
+		defer srvCancel()
+
+		// TODO: send the ExecutionState and MetricsEngine instead of the Engine
```
codebien, line 128: I see this is not removed from the latest PR, can we open an issue for it?

na--: Ah, I forgot to delete the comment in https://github.com/grafana/k6/pull/2813, added this as a reminder: https://github.com/grafana/k6/pull/2813/files#r1065929155

cmd/root_test.go
na--: For future reference, this will be resolved in https://github.com/grafana/k6/pull/2833, which I think we should keep as a separate PR, rather than trying to merge it in this series of other PRs?

cmd/run.go
na--: Resolved by the [latest force-push](https://github.com/grafana/k6/compare/2a9644af7538430ede422fdfdf44ebd2da74c4d4..eea76b15afa365ec8e95369cec3ccbd801ed7f2a)?

codebien: :+1:

codebien (approved):
LGTM

mstoykov (approved):

PR #2805, opened 2022-12-06
Show bundled extension versions in `k6 run` and `k6 version` outputs

### LINKED ISSUE #1741: Display any extensions bundled by xk6 in the pre-execution output

imiric:
This is based on the work done in #2754 (thanks @HarrisChu!), and the suggested changes there.

It evolved into something slightly more elaborate, where the bulk of the changes are to unify the extension registry, so that registration and retrieval of extension information is done in a single place. See commit cf5fa6102 for details. We still need to keep the existing registration functions in `js/modules/` and `output/` to avoid breaking extensions, but internally they're both now handled in the new root `ext` package. This DRYs the handling of extensions, and simplifies adding new types of extensions. It also simplifies the retrieval of extension versions, which is why it was done on top of #2754.

Here's how it currently looks:

```bash
$ xk6 build 4df8d46169ecc3a0417fd902bdcb28800e8098f0 \
  --with github.com/vesoft-inc/k6-plugin \
  --with github.com/grafana/xk6-output-influxdb
[...]

$ ./k6 version
k6 v0.41.0 ((devel), go1.19.1, linux/amd64)
Extensions:
  github.com/grafana/xk6-output-influxdb v0.2.2, xk6-influxdb [output]
  github.com/vesoft-inc/k6-plugin v1.0.1, aggcsv [output]
  github.com/vesoft-inc/k6-plugin v1.0.1, k6/x/nebulagraph [js]

$ ./k6 run ./samples/thresholds.js

          /\      |‾‾| /‾‾/   /‾‾/
     /\  /  \     |  |/  /   /  /

  execution: local
     script: ./samples/thresholds.js
 extensions: github.com/grafana/xk6-output-influxdb v0.2.2, xk6-influxdb [output]
             github.com/vesoft-inc/k6-plugin v1.0.1, aggcsv [output]
             github.com/vesoft-inc/k6-plugin v1.0.1, k6/x/nebulagraph [js]
     output: -

           * default: 1 iterations for each of 1 VUs (maxDuration: 10m0s, gracefulStop: 30s)
[...]
```

Closes #1741

### DIFF
--- a/cmd/ui.go
+++ b/cmd/ui.go
@@ -17,6 +17,7 @@ import (
 
 	"gopkg.in/yaml.v3"
 
+	"go.k6.io/k6/js/modules"
 	"go.k6.io/k6/lib"
 	"go.k6.io/k6/lib/consts"
 	"go.k6.io/k6/output"
@@ -81,7 +82,13 @@ func getColor(noColor bool, attributes ...color.Attribute) *color.Color {
 
 func getBanner(noColor bool) string {
 	c := getColor(noColor, color.FgCyan)
-	return c.Sprint(consts.Banner())
+	banner := make([]string, 0, len(modules.GetJSModuleVersions())+1)
+
+	banner = append(banner, consts.Banner())
+	for pkg, version := range modules.GetJSModuleVersions() {
+		banner = append(banner, fmt.Sprintf("extension  %s %s", pkg, version))
+	}
+	return c.Sprint(strings.Join(banner, "\n"))
 }
 
 func printBanner(gs *globalState) {
--- a/cmd/ui.go
+++ b/cmd/ui.go
@@ -17,7 +17,6 @@ import (
 
 	"gopkg.in/yaml.v3"
 
-	"go.k6.io/k6/js/modules"
 	"go.k6.io/k6/lib"
 	"go.k6.io/k6/lib/consts"
 	"go.k6.io/k6/output"
@@ -82,13 +81,7 @@ func getColor(noColor bool, attributes ...color.Attribute) *color.Color {
 
 func getBanner(noColor bool) string {
 	c := getColor(noColor, color.FgCyan)
-	banner := make([]string, 0, len(modules.GetJSModuleVersions())+1)
-
-	banner = append(banner, consts.Banner())
-	for pkg, version := range modules.GetJSModuleVersions() {
-		banner = append(banner, fmt.Sprintf("extension  %s %s", pkg, version))
-	}
-	return c.Sprint(strings.Join(banner, "\n"))
+	return c.Sprint(consts.Banner())
 }
 
 func printBanner(gs *globalState) {
--- a/cmd/ui.go
+++ b/cmd/ui.go
@@ -17,6 +17,7 @@ import (
 
 	"gopkg.in/yaml.v3"
 
+	"go.k6.io/k6/ext"
 	"go.k6.io/k6/lib"
 	"go.k6.io/k6/lib/consts"
 	"go.k6.io/k6/output"
@@ -136,6 +137,12 @@ func printExecutionDescription(
 	buf := &strings.Builder{}
 	fmt.Fprintf(buf, "  execution: %s\n", valueColor.Sprint(execution))
 	fmt.Fprintf(buf, "     script: %s\n", valueColor.Sprint(filename))
+
+	if exts := ext.GetAll(); len(exts) > 0 {
+		extsDesc := getExtensionsDescription(exts, 13, 1)
+		fmt.Fprintf(buf, " extensions: %s\n",
+			valueColor.Sprint(strings.Join(extsDesc, "\n")))
+	}
 
 	var outputDescriptions []string
 	switch {
@@ -400,4 +407,18 @@ func yamlPrint(w io.Writer, v interface{}) error {
 		return fmt.Errorf("could flush the data to the output: %w", err)
 	}
 	return nil
+}
+
+func getExtensionsDescription(exts []*ext.Extension, leftPad, padStart int) []string {
+	extsDesc := make([]string, 0, len(exts))
+	for i, e := range exts {
+		desc := e.String()
+		extFmt := "%s"
+		if i >= padStart {
+			extFmt = fmt.Sprintf("%%%ds", len(desc)+leftPad)
+		}
+		extsDesc = append(extsDesc, fmt.Sprintf(extFmt, desc))
+	}
+
+	return extsDesc
 }
--- a/cmd/ui.go
+++ b/cmd/ui.go
@@ -139,7 +139,7 @@ func printExecutionDescription(
 	fmt.Fprintf(buf, "     script: %s\n", valueColor.Sprint(filename))
 
 	if exts := ext.GetAll(); len(exts) > 0 {
-		extsDesc := getExtensionsDescription(exts, 13, 1)
+		extsDesc := formatExtensionsDescription(exts, 13, 1)
 		fmt.Fprintf(buf, " extensions: %s\n",
 			valueColor.Sprint(strings.Join(extsDesc, "\n")))
 	}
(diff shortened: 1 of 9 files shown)


cmd/ui.go
```diff
@@ -401,3 +408,17 @@ func yamlPrint(w io.Writer, v interface{}) error {
 	}
 	return nil
 }
+
+func getExtensionsDescription(exts []*ext.Extension, leftPad, padStart int) []string {
```
olegbespalov, line 412: Maybe `formatExtensionsDescription` ?

ext/ext.go
```diff
@@ -0,0 +1,160 @@
...
+		s = "out"
+	}
+	return s
+}
+
+// Extension is a generic container for any k6 extension.
+type Extension struct {
+	Name, Path, Version string
+	Type                ExtensionType
+	Mod                 interface{}
```
olegbespalov, line 45: Let's name it `Module` It's not longer than `Version` and doesn't harm but helps.

ext/ext.go
```diff
@@ -0,0 +1,160 @@
...
+		}
+		return result[i].Path < result[j].Path
+	})
+
+	return result
+}
+
+// getModuleInfo attempts to return the package path and version of the Go
+// module that created the given value.
+func getModuleInfo(mod interface{}) (path, version string) {
```
olegbespalov, line 125: Maybe `extractModuleInfo` ?

ext/ext.go
```diff
@@ -0,0 +1,160 @@
...
+
+	for name, ext := range exts {
+		result[name] = ext
+	}
+
+	return result
+}
+
+// GetAll returns all extensions, sorted by their import path and name.
+func GetAll() []*Extension {
```
olegbespalov, line 99: Maybe `List` ?

olegbespalov (commented):
I love that PR :heart_eyes: 

I left a couple of comments in PR and have one more general.

Although an extension output looks excellent, there are no questions for the `k6 version`. 

But I feel it's not always relevant to have the same level of details (precisely version and repository) for the `k6 run` :thinking: I had an impression that we tend to keep it minimal.

ext/ext.go
imiric: Sure. Done in 6de4f4eb7.

I'm slightly against "module" (or "mod") though, since in the case of outputs, it's actually a function, but :man_shrugging:

cmd/ui.go
imiric: :thinking: This is a bit more involved than what formatting would imply, so I think I prefer keeping it as `get*`, if you don't mind.

Otherwise, please state your case. :slightly_smiling_face:

ext/ext.go
imiric: What's wrong with `get*`? You really don't seem to like that word :smile: 

Sure, "extract" could be used here as well, but "get" is a standard name, and is clear enough IMO. :man_shrugging:

ext/ext.go
```diff
@@ -0,0 +1,160 @@
...
+	mx         sync.RWMutex
+	extensions = make(map[ExtensionType]map[string]*Extension)
+)
+
+// ExtensionType is the type of all supported k6 extensions.
+type ExtensionType uint8
+
+// All supported k6 extension types.
+const (
+	JSExtension ExtensionType = iota + 1
```
oleiade, line 26: That looks like an opportunity to learn something: why the `+1` on the `iota` here? 😇

ext/ext.go
```diff
@@ -0,0 +1,160 @@
...
+	OutputExtension
+)
+
+func (e ExtensionType) String() string {
+	var s string
+	switch e {
+	case JSExtension:
+		s = "js"
+	case OutputExtension:
+		s = "out"
```
oleiade, line 36: UX nit: I would find "output" to be more explicit, especially if we show this to the user. "out" is quite a polysemic word, used in a ton of different contexts, and as a user, I think I could be confused/unclear by what it refers to.

ext/ext.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+	return s
+}
+
+// Extension is a generic container for any k6 extension.
+type Extension struct {
+	Name, Path, Version string
+	Type                ExtensionType
+	Module              interface{}
```
oleiade, line 45: This comment might be caused by the fact that I'm only partially through the PR, but I must admit that looking at this field's definition and how it's used in `Register` as a user, I would be unclear as to what kind of value I should use to set the `Module` field.

oleiade (approved):
Really great work @imiric 👍🏻 🎉 🦖 

I'm also super happy to see this happen! I left a couple of nitpicks for your consideration, but I wouldn't consider any of them blocking in any way. Personally, I don't  necessarily mind the `k6 run` output as it is now 🙇🏻

ext/ext.go
olegbespalov: No, I'm not against the `get*`, but the truth is that it's too generic, and if it's possible to provide a more precise name, it will be better since it saves time for a code reader.

I have a feeling that extract can tell more to the end user than the `get`, since, from the method name user already can understand that we're extracting details from the first argument without jumping to the method body.

imiric: I see, that makes sense. I renamed it to `extractModuleInfo` in b1f3c94b6.

cmd/ui.go
imiric: Following your argument about `get` being too generic below, I also renamed this to `formatExtensionsDescription` in 37b0e1e95.

ext/ext.go
imiric: Just a convention to avoid constants with a value of 0. That way you can identify when a value was set or not.

ext/ext.go
```diff
@@ -0,0 +1,160 @@
...
+	case OutputExtension:
+		s = "out"
+	}
+	return s
+}
+
+// Extension is a generic container for any k6 extension.
+type Extension struct {
+	Name, Path, Version string
+	Type                ExtensionType
```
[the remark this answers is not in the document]
imiric: :thinking: I think we've always referred to these as extension "types", so I wouldn't want to change our vocabulary just because it's a reserved keyword. Though in this case `type` is the keyword, and `Type` works just fine, so I'd rather keep calling it this, if you don't mind.

ext/ext.go
imiric: Sure, it is confusing. :disappointed: We initially called this "module" when JS extensions were introduced, as JS extensions would register a struct of their JS module. But then we changed the API for output extensions and had them register a constructor function that would return a module, for initialization purposes. In retrospect, I think this was the wrong API to introduce, but since I doubt we'll change this, every extension type is now free to choose what works best.

Since I can't think of a more generic name that would fit better, I'm inclined to leave it as `Module`.

na--:
My vote would be to show extension information only in `k6 version`, but not in `k6 run` :thinking_face: Maybe we can add a `debug`-level log message with that information in all other commands (`k6 run`, `k6 archive`, `k6 inspect`), but that's it 

`k6 run` is already quite noisy, so we shouldn't add more things in there that might be completely unconnected to the actual test run :thinking:  Whatever extensions were compiled into the current k6 binary, there is no guarantee that they are being used at the moment. Besides, we don't show the k6 version, why should we show the extensions that were bundled in it?

Issue #2804, opened 2022-12-06
Unify and standardize behavior and exit codes when k6 is stopped


na--:
k6 has multiple different ways to stop a test:
1. Single Ctrl+C for semi-graceful shutdown
2. Second Ctrl+C for immediate abort
3. The `test.abort()` JS API from `k6/execution`
4. JS script exception
5. Via the REST API, e.g. with `curl -i -X PATCH -d '{"data":{"type":"status","id":"default","attributes":{"stopped":true}}}' 'http://localhost:6565/v1/status'`
6. A threshold with `abortOnFail: true`

And, for reference, we currently have these exit codes predefined: https://github.com/grafana/k6/blob/1b9c5fa55598363501a3db7c44cccddc8430af8f/errext/exitcodes/codes.go#L10-L21

The problem is that stopping the test in different ways and at different times behaves very inconsistently and unpredictably, even when you account for the unavoidable peculiarities of k6... :sob: Here is what I've determined:
- While k6 is [initializing for the first time](https://github.com/grafana/k6/blob/1b9c5fa55598363501a3db7c44cccddc8430af8f/cmd/run.go#L42) (to get the exported `options`):
  - :negative_squared_cross_mark: 1. a single Ctrl+C will not be graceful at all, it will behave like 2.; it will also cause the process to abort with an exit code of `1`, which is apparently the default Go interrupt signal handling behavior when it has not yet been handled by k6
  - :negative_squared_cross_mark: 2. will not even be reached (see above)
  - :heavy_check_mark: 3. `test.abort()` will surprisingly work as expected and cause k6 to exit with `ScriptAborted`
  - :heavy_check_mark: 4. script exceptions work as expected, k6 exits with `ScriptException`
  - :black_square_button: 5. doesn't apply, since the REST API has not yet been started at this point (and probably can't be started so early), so it can't be used to stop the test
  - :black_square_button: 6. also doesn't apply, the test hasn't been started
- While k6 is [initializing one of the actual test VUs](https://github.com/grafana/k6/blob/1b9c5fa55598363501a3db7c44cccddc8430af8f/cmd/run.go#L160-L162):
  - :negative_squared_cross_mark: 1. is now handled by the k6 signal handler, but again it won't be graceful and the process will exit with code `GenericEngine`, the catch-all exit code for when k6 doesn't "know" what caused the failure... :disappointed:
[...]

PR #2807, opened 2022-12-07
Use non-zero exit codes for some aborted by Ctrl+C or REST API tests

na--:
This is built on top of https://github.com/grafana/k6/pull/2803 and tackles a small part of https://github.com/grafana/k6/issues/2804, though definitely not the whole issue.

It was necessary to be able to split apart https://github.com/grafana/k6/issues/1889 into a few smaller PRs and it was also necessary for https://github.com/grafana/k6/issues/1342 in its own right.

### DIFF
--- a/cmd/integration_test.go
+++ b/cmd/integration_test.go
@@ -167,6 +167,19 @@ func TestWrongEnvVarIterations(t *testing.T) {
 	assert.Empty(t, ts.loggerHook.Drain())
 }
 
+func getSingleFileTestState(t *testing.T, script string, cliFlags []string, expExitCode exitcodes.ExitCode) *globalTestState {
+	if cliFlags == nil {
+		cliFlags = []string{"-v", "--log-output=stdout"}
+	}
+
+	ts := newGlobalTestState(t)
+	require.NoError(t, afero.WriteFile(ts.fs, filepath.Join(ts.cwd, "test.js"), []byte(script), 0o644))
+	ts.args = append(append([]string{"k6", "run"}, cliFlags...), "test.js")
+	ts.expectedExitCode = int(expExitCode)
+
+	return ts
+}
+
 func TestMetricsAndThresholds(t *testing.T) {
 	t.Parallel()
 	script := `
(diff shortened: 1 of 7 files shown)


cmd/run.go
```diff
@@ -263,13 +264,16 @@ func (c *cmdRun) run(cmd *cobra.Command, args []string) error {
...
-	if err != nil {
-		return errext.WithExitCodeIfNone(common.UnwrapGojaInterruptedError(err), exitcodes.GenericEngine)
-	}
+
 	if engine.IsTainted() {
-		return errext.WithExitCodeIfNone(errors.New("some thresholds have failed"), exitcodes.ThresholdsHaveFailed)
+		if err == nil {
+			err = errors.New("some thresholds have failed")
+		} else {
+			logger.Error("some thresholds have failed") // log this, even if there was already a previous error
```
olegbespalov, line 272: Just out of the curiosity, why does it look like this and not like:

```golang
err = fmt.Errorf("%w; some thresholds have failed", err)
```

na--: No strong reason, just 2 weak ones :sweat_smile: First, because the two errors are not actually connected, I don't think it makes sense to wrap one in the other and make a composite error :thinking: 

Second reason is that it is probably going to be a bit more visible to the user if `some thresholds have failed` is a separate `error` log message to the error message that is stored in `err` and will actually abort the test. It could be quite long, so `some thresholds have failed` might get lost at its end

olegbespalov: I'll buy it

cmd/integration_test.go
```diff
@@ -669,46 +668,80 @@ func TestAbortedByUserWithGoodThresholds(t *testing.T) {
...
+			ts.outMutex.Unlock()
+
+			if strings.Contains(stdOut, expText) {
+				t.Logf("found '%s' in the process stdout on try %d at t=%s", expText, i, time.Now())
+				reachedCondition = true
+				break
+			}
+
+			time.Sleep(interval)
+			t.Logf("did not find the text '%s' in the process stdout on try %d at t=%s", expText, i, time.Now())
```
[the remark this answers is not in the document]
na--: I will fix this, but I'll hold off a bit, until more people review this and the preceding PRs, so I minimize the number of rebases I have to do to the whole chain of PRs :sweat_smile:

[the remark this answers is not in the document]
na--: Actually, I ended up fixing moving this specific commit to an earlier PR and fixed this issue there, see https://github.com/grafana/k6/pull/2800#issuecomment-1351119868

olegbespalov (dismissed):

oleiade (dismissed):

olegbespalov (approved):

PR #2809, opened 2022-12-07
Remove dependency on the `js/` package (and goja) from `lib/*` packages

na--:
This doesn't depend on anything else, it should be safe to merge as is.

This should close https://github.com/grafana/k6/issues/2535, I think :thinking: Between it and https://github.com/grafana/k6/pull/2536, `go.k6.io/lib` and its sub-packages shouldn't depend on `go.k6.io/js` or `github.com/dop251/goja`, right?

`go list -f '{{ join .Imports "\n" }}'  ./... | sort -u | grep -E 'goja|go.k6.io/k6/js'` returns no results in the `lib/` package, at least :crossed_fingers: 

Unfortunately, I couldn't make depbuard work correctly, and I have no time to debug that further, so I'll leave that to someone else if they want to. From what I understand by skimming https://golangci-lint.run/usage/linters/#depguard, adding something like this in the `linters-settings:` section of `.golangci.yml` should have worked, but it didn't:
```yml
  depguard:
    list-type: denylist
    packages:
      - github.com/dop251/goja
      - go.k6.io/k6/js
      - go.k6.io/k6/js/**
    ignore-file-rules:
      - "js/**/*.go"
      - "cmd/*.go"
```

The goal is to ensure we don't import goja or the js package anywhere except `js/` and `cmd/`, and maybe we can even restrict that further :thinking: If someone can make the linter work, that is :sweat_smile:

### DIFF
--- a/lib/netext/grpcext/conn.go
+++ b/lib/netext/grpcext/conn.go
@@ -10,7 +10,7 @@ import (
 	"strings"
 
 	"github.com/sirupsen/logrus"
-	"go.k6.io/k6/js/modules"
+	"go.k6.io/k6/lib"
 	"go.k6.io/k6/metrics"
 
 	protov1 "github.com/golang/protobuf/proto" //nolint:staticcheck,nolintlint // this is the old v1 version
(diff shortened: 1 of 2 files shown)


lib/netext/grpcext/conn.go
```diff
@@ -174,7 +174,7 @@ func (c *Conn) Close() error {
 }
 
 type statsHandler struct {
-	vu modules.VU
+	getState func() *lib.State
```
mstoykov, line 177: nit: Do we need this to be a function :thinking: 

I guess for future proofing in case we start chaning what lib.State is returned. But IIRC we basically return the same struct eitherway.

We do need it to be a function on the VU as it is an interface. But I wonder if we can skip the inderection here :thinking:

na--: I'd slightly prefer to keep the function, just so we keep the same sort of API structure and in case it might matter in the future, but I can make the change if others feel the same as you.

FWIW, I even considered adding another interface with just a `State() *lib.State` function in `grpext/` and continuing to pass the `VU` directly, or even moving the whole `DefaultOptions` function back in the `js/` package. The current approach seemed like a nice middle ground :sweat_smile:

oleiade (approved):
👍🏻

mstoykov (approved):

Issue #2430, opened 2022-03-10
Refactor the `Output` interface


na--:
I've recently been going back to basics and making some substantial changes and refactors in how base components of k6 work. One of these has been the current [`core.Engine`](https://github.com/grafana/k6/blob/55c7ccd53d6c0775740e0880538bd4c8b231bc7a/core/engine.go#L48). Through https://github.com/grafana/k6/issues/1889, https://github.com/grafana/k6/pull/2426 and another WIP PR I've been slowly trying to split it apart and obviate it, since that will unlock a lot of flexibility and remove tons of cruft.

During that effort, I've noticed another suboptimal component we have in k6, our outputs. This is the current `output.Output` API: https://github.com/grafana/k6/blob/55c7ccd53d6c0775740e0880538bd4c8b231bc7a/output/types.go#L58-L76

And even though I was the one who kind of wrote it, I still think it's quite bad :sweat_smile: I think the biggest reason it's so bad is because it kind of had to be, due to the dysfunctional nature of our current Rube-Goldberg `core.Engine` :sweat_smile: But I'm really close to getting rid of the Engine completely, and it has been making obvious how bad the Outputs are because of it...

Here's what I actually propose the new Output API should look like:
```go
type Output interface {
	// The output stops when `samples` is closed, and k6 will wait() for it to finish.
	Start(samples chan stats.SampleContainer) (wait func() error, err error)
}

type WithDescription interface {
	Output
	Description() string
}
// ... same single-method optional interfaces: WithThresholds, WithTestRunStop, WithRunStatusUpdates, WithBuiltinMetrics
```

Besides the fact that then new API is much simpler and more idiomatic Go code, the old API actually has a problem for the use case of k6. Specifically, buffering is important when we have network IO, but counter-productive when we're passing metrics internally in k6. It's probably somewhat jittery to have these buffers from `AddMetricSamples()` internally, since their batched processing will cause memory churn, but also small CPU usage spikes at intermittent intervals when they are processed. If there are a CPU bursts every 50ms, that might affect the actual test execution and metric measurements... :disappointed:

I also think there is a way to do this transition gracefully and support both old and new Output extensions for a few releases (the old ones with an adapter to the new API that emits a warning). We can move the outputs to the new `/metrics/` top-level package and make the current `/output` package a adapter proxy for the new API for a few k6 version. That proxy can emit a deprecation warning, so users can switch their extensions to the new API.

MStoykov:
Without looking into the proposition too much: 

[...]

PR #2810, opened 2022-12-07
Add new `errext.AbortReason` type, move and use old `RunStatus` type only for k6 cloud code

na--:
This is built on top of https://github.com/grafana/k6/pull/2807 and is a part of https://github.com/grafana/k6/issues/1889 and part of https://github.com/grafana/k6/issues/2430

`RunStatus` is very specific to the k6 cloud and doesn't map perfectly to how k6 OSS sees the causes of prematurely stopped tests. So, this PR restricts the usage of `RunStatus` only to the `cloudapi/` package and to the `k6 cloud` command handling in `cmd/cloud.go`.

It does that by introducing a new type, `errext.AbortReason`, and a way to attach this type to test run errors. This allows us a cleaner error propagation to the outputs, and every output can map these generic values to its own internal ones however it wants. It is not necessary for that mapping to be exactly 1:1 and, indeed, as you can see, the `errext.AbortReason`:`cloudapi.RunStatus` mapping in `cloudapi/` is not 1:1.

### DIFF
--- a/core/engine.go
+++ b/core/engine.go
@@ -8,7 +8,6 @@ import (
 
 	"github.com/sirupsen/logrus"
 
-	"go.k6.io/k6/cloudapi"
 	"go.k6.io/k6/errext"
 	"go.k6.io/k6/errext/exitcodes"
 	"go.k6.io/k6/lib"
@@ -108,7 +107,6 @@ func (e *Engine) Init(globalCtx, runCtx context.Context) (run func() error, wait
 	// TODO: if we ever need metrics processing in the init context, we can move
 	// this below the other components... or even start them concurrently?
 	if err := e.ExecutionScheduler.Init(runCtx, e.Samples); err != nil {
-		e.setRunStatusFromError(err)
 		return nil, nil, err
 	}
 
@@ -148,18 +146,6 @@ func (e *Engine) Init(globalCtx, runCtx context.Context) (run func() error, wait
 	return runFn, waitFn, nil
 }
 
-func (e *Engine) setRunStatusFromError(err error) {
-	var serr errext.Exception
-	switch {
-	case errors.As(err, &serr):
-		e.OutputManager.SetRunStatus(cloudapi.RunStatusAbortedScriptError)
-	case errext.IsInterruptError(err):
-		e.OutputManager.SetRunStatus(cloudapi.RunStatusAbortedUser)
-	default:
-		e.OutputManager.SetRunStatus(cloudapi.RunStatusAbortedSystem)
-	}
-}
-
 // This starts a bunch of goroutines to process metrics, thresholds, and set the
 // test run status when it ends. It returns a function that can be used after
 // the provided context is called, to wait for the complete winding down of all
@@ -196,27 +182,31 @@ func (e *Engine) startBackgroundProcesses(
 		case err = <-execRunResult:
 			if err != nil {
 				e.logger.WithError(err).Debug("run: execution scheduler returned an error")
-				e.setRunStatusFromError(err)
 			} else {
 				e.logger.Debug("run: execution scheduler finished nominally")
-				e.OutputManager.SetRunStatus(cloudapi.RunStatusFinished)
 			}
 			// do nothing, return the same err value we got from the Run()
 			// ExecutionScheduler result, we just set the run_status based on it
 		case <-runCtx.Done():
 			e.logger.Debug("run: context expired; exiting...")
-			e.OutputManager.SetRunStatus(cloudapi.RunStatusAbortedUser)
-			err = errext.WithExitCodeIfNone(errors.New("test run aborted by signal"), exitcodes.ExternalAbort)
+			err = errext.WithAbortReasonIfNone(
+				errext.WithExitCodeIfNone(errors.New("test run aborted by signal"), exitcodes.ExternalAbort),
+				errext.AbortedByUser,
+			)
 		case <-e.stopChan:
 			runSubCancel()
 			e.logger.Debug("run: stopped by user via REST API; exiting...")
-			e.OutputManager.SetRunStatus(cloudapi.RunStatusAbortedUser)
-			err = errext.WithExitCodeIfNone(errors.New("test run stopped from the REST API"), exitcodes.ScriptStoppedFromRESTAPI)
+			err = errext.WithAbortReasonIfNone(
+				errext.WithExitCodeIfNone(errors.New("test run stopped from the REST API"), exitcodes.ScriptStoppedFromRESTAPI),
+				errext.AbortedByUser,
+			)
 		case <-thresholdAbortChan:
 			e.logger.Debug("run: stopped by thresholds; exiting...")
 			runSubCancel()
-			e.OutputManager.SetRunStatus(cloudapi.RunStatusAbortedThreshold)
-			err = errext.WithExitCodeIfNone(errors.New("test run aborted by failed thresholds"), exitcodes.ThresholdsHaveFailed)
+			err = errext.WithAbortReasonIfNone(
+				errext.WithExitCodeIfNone(errors.New("test run aborted by failed thresholds"), exitcodes.ThresholdsHaveFailed),
+				errext.AbortedByThreshold,
+			)
 		}
 	}()
 
--- a/core/engine.go
+++ b/core/engine.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/sirupsen/logrus"
 
+	"go.k6.io/k6/cloudapi"
 	"go.k6.io/k6/errext"
 	"go.k6.io/k6/errext/exitcodes"
 	"go.k6.io/k6/lib"
@@ -151,11 +152,11 @@ func (e *Engine) setRunStatusFromError(err error) {
 	var serr errext.Exception
 	switch {
 	case errors.As(err, &serr):
-		e.OutputManager.SetRunStatus(lib.RunStatusAbortedScriptError)
+		e.OutputManager.SetRunStatus(cloudapi.RunStatusAbortedScriptError)
 	case errext.IsInterruptError(err):
-		e.OutputManager.SetRunStatus(lib.RunStatusAbortedUser)
+		e.OutputManager.SetRunStatus(cloudapi.RunStatusAbortedUser)
 	default:
-		e.OutputManager.SetRunStatus(lib.RunStatusAbortedSystem)
+		e.OutputManager.SetRunStatus(cloudapi.RunStatusAbortedSystem)
 	}
 }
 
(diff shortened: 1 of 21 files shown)


output/cloud/output.go
```diff
@@ -259,17 +258,23 @@ func (out *Output) startBackgroundProcesses() {
 	}()
 }
 
-// Stop gracefully stops all metric emission from the output and when all metric
-// samples are emitted, it sends an API to the cloud to finish the test run.
+// Stop is deprecated, see the StopWithTestError() method.
```
olegbespalov, line 261: Just to give a chance for godoc (and :crossed_fingers: at some point vscode) to parse and mark it properly

```suggestion
// Stop gracefully stops all metric emission from the output and when all metric
// samples are emitted, it sends an API to the cloud to finish the test run.
//
// Deprecated: use StopWithTestError() instead.
```

olegbespalov (commented):
LGTM :+1: left minor comment

output/cloud/output.go
na--: Resolved by the [latest force-push](https://github.com/grafana/k6/compare/62eb4f967a8ac2bb492be65904e12b5d8f312de8..b456ed79672d120a585dda6f066deaa8f8bd5429#diff-f70f77b77caf0bb396844dacff629677ceb27f5e7cd183bca0d7d08b8672e4beR261-R266).

olegbespalov (dismissed):

core/engine.go
```diff
@@ -195,27 +182,31 @@ func (e *Engine) startBackgroundProcesses(
...
 			// do nothing, return the same err value we got from the Run()
 			// ExecutionScheduler result, we just set the run_status based on it
 		case <-runCtx.Done():
 			e.logger.Debug("run: context expired; exiting...")
-			e.OutputManager.SetRunStatus(lib.RunStatusAbortedUser)
-			err = errext.WithExitCodeIfNone(errors.New("test run aborted by signal"), exitcodes.ExternalAbort)
+			err = errext.WithAbortReasonIfNone(
+				errext.WithExitCodeIfNone(errors.New("test run aborted by signal"), exitcodes.ExternalAbort),
+				errext.AbortedByUser,
+			)
```
imiric, line 195: This wrapping is a bit convoluted. It seems we're only wrapping exit code errors in aborted errors because it's convenient to return a single error here. But conceptually, these errors aren't related in this way. If anything, maybe there should be a mapping from `AbortReason` to `ExitCode`, which can happen outside of `core`. After all, why should the `Engine` be concerned with process exit codes?

I'm not sure I have a concrete improvement suggestion here, but this seems fishy in several ways. :thinking:

output/types.go
```diff
@@ -75,10 +75,17 @@ type WithTestRunStop interface {
...
-type WithRunStatusUpdates interface {
+// WithStopWithTestError allows output to receive the error value that the test
+// finished with. It could be nil, if the test finished nominally.
+//
+// If this interface is implemented by the output, StopWithError() will be
+// called instead of Stop().
+//
+// TODO: refactor the main interface to use this method instead of Stop()? Or
+// something else along the lines of https://github.com/grafana/k6/issues/2430 ?
+type WithStopWithTestError interface {
```
imiric, line 86: I think the proposal in #2430 that outputs should stop when the `samples` channel is closed would be cleaner, but we would still need a way for them to query the test error, as suggested [here](https://github.com/grafana/k6/issues/2430#issuecomment-1318540637).

So WDYT if we add a method to `output.Params` for this purpose? This would avoid this deprecation of `Stop()` and eventual breaking change, and only outputs that need it can call it. Eventually, we can deprecate and remove `Stop()` in favor of the proposal in #2430, but that would be a single breaking change, instead of the two in this case.

An alternative to changing `output.Params` could be for outputs to subscribe to error events, using the event system suggested for JS extensions in #2432, but I think that's still a long ways off.

imiric (commented):
Moving `RunStatus` to `cloudapi` makes sense, but I have some suggestions and questions about the other changes.

codebien (changes requested):
`test-tip` is not happy with the latest push

olegbespalov:
I guess the go-tip is unhappy with the `master` branch https://github.com/golang/go/commit/62a994837a57a7d0c58bb364b580a389488446c9 :cry:

core/engine.go
na--: Great point! Now that this PR adds the new `AbortReason` type that can accurately represent the internal reason for why a k6 test run failed, we can use it for external mappings. The PR does the mapping from `AbortReason` to `cloudapi.RunStatus`, but I don't see a reason why we couldn't also have a mapping to `ExitCode`. We'd probably need to add more `AbortReason` values, but that should be fine :thinking: In any case, I'd prefer to handle this in a separate PR, so I added this https://github.com/grafana/k6/issues/2804#issuecomment-1399978135

output/cloud/output.go
```diff
@@ -283,9 +291,37 @@ func (out *Output) Description() string {
...
+			return cloudapi.RunStatusAbortedScriptError
+		case errext.AbortedByScriptAbort:
+			return cloudapi.RunStatusAbortedUser // TODO: have a better value than this?
+		case errext.AbortedByTimeout:
+			return cloudapi.RunStatusAbortedLimit
+		case errext.AbortedByThresholdsAfterTestEnd:
+			// The test finished normally, it wasn't prematurely aborted. Such
+			// failures are tracked by the restult_status, not the run_status
+			// (so called "tainted" in some places of the API here).
+			return cloudapi.RunStatusFinished
```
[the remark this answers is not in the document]
na--: I improved the comments in https://github.com/grafana/k6/pull/2810/commits/be3aa5f89c854dece753287783eb07c7d46af084, which hopefully clears up some things.

The TLDR versions is that we have a mismatch between the k6 OSS and cloud. k6 OSS needs to have an error here, so it can exit with a non-0 exit code, but the k6 cloud tracks that error differently and needs the normal `RunStatusFinished` here :disappointed: Though I am also not super with the `AbortedByThresholdsAfterTestEnd` name, so if anyone has any better suggestions, please comment here: https://github.com/grafana/k6/blob/be3aa5f89c854dece753287783eb07c7d46af084/errext/abort_reason.go#L13

output/types.go
na--: We don't have a breaking change here, this is why I added an extra optional interface. All of the existing outputs and output extensions that use the old `Stop()` method will continue to work. They will just continue not knowing if the test finished with an error or not, which most outputs don't care about anyway.

I am softly against adding a control surface to `output.Params`, it is counter-intuitive. But I am open to discussion and willing to be convinced otherwise, when we decide to tackle https://github.com/grafana/k6/issues/2430. For now, this solution allows us to delay any breaking changes until we do so, exactly so we can do them only once and not constantly break extensions.

output/types.go
```diff
@@ -75,10 +75,17 @@ type WithTestRunStop interface {
...
 	Output
-	SetRunStatus(latestStatus lib.RunStatus)
+	StopWithTestError(testRunErr error) error // nil testRunErr means error-free test run
```
codebien, line 88:
```suggestion
	StopWithTestError(ctx context.Context, testRunErr error) error // nil testRunErr means error-free test run
```

I would use this opportunity to introduce a graceful stop. I'm ok with just changing the API here and addressing the graceful stop from the output manager in a dedicated PR after the removal.

codebien (changes requested):

output/types.go
na--: hmm, the convention so far has been to always wait for outputs to finish stopping :thinking: that is, unless a user hits Ctrl+C twice to trigger the hard stop with `os.Exit()`, k6 will always wait for outputs to finish... which makes some sense to me and I am not sure we should change it, since we then will need some way to configure this timeout

so, I can add a context here, but we probably won't ever use it... similar to what I said in https://github.com/grafana/k6/pull/2810#discussion_r1083779161, my vote would be to add a note in https://github.com/grafana/k6/issues/2430 and we can discuss it when we completely refactor the outputs :thinking:

codebien: >hmm, the convention so far has been to always wait for outputs to finish stopping

It doesn't sound to be aligned with the comment here: https://github.com/grafana/k6/issues/2430#issuecomment-1073803076

>This will allow us to set a soft limit for how long outputs are allowed to stop.

I'm ok with moving this as a discussion for 2430

output/cloud/output.go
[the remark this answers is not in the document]
imiric: Thanks, but the comment was the least of my concerns here :sweat_smile: While a long explanation of this mismatch helps, I'd prefer if we could avoid it.

My main objection is that if thresholds fail after the test ends, the test wasn't _aborted_. So it's not so much about the naming of `AbortedByThresholdsAfterTestEnd`, but that it's wrong and confusing, and that it shouldn't be an `AbortReason` at all. The only reason we use it, is so we can easily catch it here, but we can avoid this if we detect this scenario in a different way.

Maybe use a different error altogether? I'm not sure what the best approach would be, but the current situation is confusing even with the long explanation.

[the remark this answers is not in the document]
na--: > but we can avoid this if we detect this scenario in a different way.

The problem is that this will require a very, very substantial refactoring, just for this one single minor mismatch... :disappointed: And we'd essentially end up needing two different mechanisms for reporting the error to the output, or something like that :thinking: 

But overall it is a very fair point :thinking: `AbortReason` is not that great of a name for other reasons, now that I think about it some more, especially if we want to use it for https://github.com/grafana/k6/pull/2810#discussion_r1083770450 :thinking:  What do you think about `TestRunError` as the type name and these as the values:
```go

// TestRunError is used to attach extra information for the cause of a test run failure or interruption.
type TestRunError uint8

const (
	TestRunAbortedByUser TestRunError = iota + 1
	TestRunAbortedByBreachedThresholds
	TestRunThresholdsWereBreached // at the end of the test
	TestRunAbortedByScriptError
	TestRunAbortedByScriptAbort
	TestRunAbortedByTimeout
)

// IsTestRunError is a wrapper around an error with an attached TestRunError.
type IsTestRunError interface {
	error
	TestRunError() TestRunError
}
```

output/types.go
imiric: We're avoiding a breaking change here by doing this optional expansion of the `Output` interface, which is IMO a code smell. If we decide to remove `Stop()` in #2430, then we'll also need to remove `StopWithTestError()`, hence two eventual breaking changes. I'd rather we avoided adding future work, and reconsider how outputs can query test errors. Essentially, doing this in a way that brings us closer to #2430, not farther.

Adding a method to `output.Params` was a suggestion, but I also agree not a very intuitive one. Maybe via some other public method?

na--: > which is IMO a code smell

Why? I'd consider it a fairly idiomatic Go thing, the stdlib does it all the time :confused: 

> If we decide to remove Stop() in https://github.com/grafana/k6/issues/2430, then we'll also need to remove StopWithTestError(), hence two eventual breaking changes.

It is likely that when we tackle https://github.com/grafana/k6/issues/2430, we will refactor [all of the existing interfaces](https://github.com/grafana/k6/blob/41472d1ba41737be699fc756ad0cdec4b2f0540f/output/types.go#L18-L88), and that won't be 6 different breaking changes, it will be one big change :sweat_smile: And even so, it's likely it will be somewhat graceful, since we will likely have a Go shim to the old `Output` APIs that will work (with a warning) for a few k6 versions.

In any case, we still don't have a clear idea of how exactly we will do https://github.com/grafana/k6/issues/2430, and when exactly we will do it, or even *if* we will actually do it... So I don't want to block this PR and its follow-ups until we have that figured out. And, FWIW, figuring out how the new Outputs will look like should come after we have removed the `Engine` in https://github.com/grafana/k6/pull/2815, since it changes things in this area a lot, but it also depends on this PR first.

output/cloud/output.go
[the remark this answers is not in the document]
imiric: :thinking: Yeah, `TestRunError` is a better name than `AbortReason`, but it seems like we're just prefixing `TestRun` to make it more generic so we can fit `TestRunThresholdsWereBreached` :smile:

The `cloud` output already knows if thresholds were breached with [this "tainted" check](https://github.com/grafana/k6/blob/be3aa5f89c854dece753287783eb07c7d46af084/output/cloud/output.go#L675-L685). It could arguably even avoid this check, since the `Engine` (or whatever we end up moving this to in this series of PRs) already knows this from [`engine.IsTainted()`](https://github.com/grafana/k6/blob/be3aa5f89c854dece753287783eb07c7d46af084/cmd/run.go#L273), and the output could just query it. In the meantime, could it infer the `RunStatus` from this?

I feel like we're abusing errors as a way to pass information to outputs, and making global changes that are only useful for the `cloud` output. This could be avoided if we had an API for outputs to query this and other information instead, and it's related to the discussion below about `StopWithTestError()`.

If you think this would be a much larger refactor, then let's use `TestRunError` for now, but we should discuss and improve this soon-ish.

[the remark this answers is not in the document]
na--: Yeah, it will be quite a big refactor, and it will be difficult to make an API that doesn't have timing issues. This API will only work once the test has finished and the thresholds have been crunched. But the output will only know that this has happened once its `Stop` method has been called. So my thinking was - why not pass that information to taht method anyway? :sweat_smile: 

Even my original idea from https://github.com/grafana/k6/issues/2430, to have outputs know that they need to stop when their `Samples` channel is closed, is not very workable... :disappointed: After we finish the test and run out of samples, we first need to crunch the thresholds and only then stop the rest of the outputs... So yeah, this is complicated and I'd very much like to leave solving it for after we have gotten rid of the `Engine`, since that will simplify things tremendously

output/types.go
imiric: OK, let's leave this for later then.

> Why? I'd consider it a fairly idiomatic Go thing, the stdlib does it all the time

I don't mean in the general sense. Composing interfaces is sometimes clean and convenient. But in this case it means that outputs don't just implement `output.Output`, but some outputs also implement `WithStopWithTestError`, which besides being awkwardly named, also requires runtime type checks wherever this additional functionality is required. And we're doing all this just to avoid a breaking change, which could be prevented by rethinking how this is done, and getting us closer to #2430.

imiric (dismissed):

codebien (commented):
LGTM, @na-- I wait for the rebase

na--:
Rebased

imiric (approved):

olegbespalov (approved):

codebien (approved):

PR #2812, opened 2022-12-07
Move `core/local/ExecutionScheduler` to `execution/Scheduler`

na--:
...and also move and expand the special context wrapper that `test.abort()` uses to interrupt the test and store the error internally. Its improved capabilities will be required for https://github.com/grafana/k6/issues/1889

This is built on top of https://github.com/grafana/k6/pull/2810 and is also the first step towards addressing https://github.com/grafana/k6/issues/1302. There is still a long way to go before we can move all of the execution-related things out of `lib/`, mostly because it will be quite tricky to both have a nice module structure and avoid import loops :disappointed: Even now, parts of `execution/` (esp. the tests) still import `lib/executor/`, while `lib/executor` imports the parts of `lib/` we want to move to `execution/`... :sob: I worked around the issue this time, though that might not always work as well as it did this time.

In any case, I just made this change to get the process of cleaning up `lib/` started. Between this PR and the next one to address https://github.com/grafana/k6/issues/1889, it's much easier to remove the `core/` package than it is to cleanly split apart `lib/`. Now, we have at least a place for these things, even if we can't move everything quite yet.


olegbespalov (dismissed):

execution/scheduler.go
```diff
@@ -16,8 +16,10 @@ import (
 	"go.k6.io/k6/ui/pb"
 )
 
-// ExecutionScheduler is the local implementation of lib.ExecutionScheduler
-type ExecutionScheduler struct {
+// A Scheduler is in charge of most of the test execution - initializing VUs and
+// executors, running setup() and teardown(), and actually starting the
+// executors for the different scenarios at the appropriate times.
+type Scheduler struct {
```
oleiade, line 21: As I worked on putting together some architecture documentation a while ago, this is how I described the `ExecutionScheduler`'s role in k6:

```
The Execution Scheduler is essentially the test run execution’s orchestrator. It is the component in charge of initializing the VUs, running the script’s setup, starting all the Executors necessary to perform the test run execution, and performing the script’s teardown once the execution is over.

Another essential responsibility of the Execution Scheduler is forwarding the collected samples back to the Engine. Because it is such a central piece in charge of orchestrating Executors and VUs, and as it has access to both, it is its role to pass those with the channel they should forward the samples they collect.

Finally, throughout the test run, the Execution Scheduler oversees maintaining and publishing the execution status. At which step of the execution we’re currently at and whether we’re in a healthy state.
```

I just wanted to post it here in case you find any of this relevant and valuable to add to the documentation of the struct 🙇🏻 

Also, as we're moving and renaming the `ExecutionScheduler` around: my understanding of the role of the ExecutionScheduler is that it acts more as an orchestrator rather than a scheduler. In the sense that it is in charge of starting components, ensuring other methods are called, and acting as a data forwarder rather than scheduling some tasks execution, for instance (this is more the role of the `Executor` in my understanding).

All of this is to say that I would find it better suited if it was named `Orchestrator` rather than `Scheduler`, although I'm neither hung up on the name we pick nor on the renaming itself. Let me know what you think 😸

oleiade (dismissed):
Very good work 👍🏻 I'm glad to see the move to a dedicated `execution` package happen 🎉 🦖 I've left a suggestion to maybe find a more suiting name for the Scheduler as we're at it, but I've really no strong opinion on this, and you should feel free to walk past it 🙇🏻

execution/scheduler.go
na--: Hmm :thinking: 

To me, an `Orchestrator` somewhat implies that the component will make decisions **where** (and maybe even **how**) its constituent jobs will be executed. While the code here, besides VU and executor initialization (and error handling), is almost entirely concerned with **when** the jobs will be executed. For now that means just respecting the scenarios' `startTime` options, but in the future with test suites and distributed execution (see the PoC in https://github.com/grafana/k6/pull/2816), it will also deal with synchronizing some things between instances.

Honestly, neither `execution.Orchestrator` nor `execution.Scheduler` seem like a good fit for this component's name, but between the two, `Scheduler` seems slightly less bad... :sweat_smile:  `execution.Controller` might be a slightly better name, but if we use it, we'll have to figure out what to name this component we need for distributed execution and test suites :sweat_smile: 
https://github.com/grafana/k6/blob/e8d93ecf7527d1206ee3ff2844da52e3bf826731/execution/controller.go#L1-L11

So, yeah, slightly against  `execution.Orchestrator`, but definitely open for better

oleiade: Thanks for the detailed insights. In hindsight, you're right; Orchestrator's not either an accurate or good enough name 🤝

I like Controller; it gets the point across nicely: "I control the overall execution of a test."  In the meantime, I also remembered that I found the `Runner` abstraction slightly misnamed, or "over"-named 😄. Now that I look at it from this perspective, `Runner` could be an excellent name for the currently named `Scheduler` abstraction, especially now that the Engine brick is gone: "I ensure the test is being run". But once again not an especially strong opinions on this 👍🏻

na--: Yeah, `Runner` is a terrible name for the thing that currently uses it, and would be a better fit here (though maybe not as good as `Controller`... I'll think on this a little bit more, it's not urgent that we fix this in this PR. 

`k6/execution.Scheduler` is still a better name than `k6/core/local.ExecutionScheduler` and we still got rid of the unnecessary interface, so even this PR in its current state is an improvement of the status quo :sweat_smile:

oleiade: Makes sense 👍🏻

execution/scheduler_int_test.go
```diff
@@ -0,0 +1,91 @@
...
+		RuntimeOptions: lib.RuntimeOptions{},
+		Registry:       reg,
+		BuiltinMetrics: metrics.RegisterBuiltinMetrics(reg),
+	}
+
+	return &lib.TestRunState{
+		TestPreInitState: piState,
+		Options:          lib.Options{},
+		Runner:           &minirunner.MiniRunner{},
+		RunTags:          piState.Registry.RootTagSet().WithTagsFromMap(nil),
```
codebien, line 27: Is it expected?

codebien (commented):
LGTM, just a minor nitpick

oleiade (dismissed):

execution/scheduler_int_test.go
na--: good catch - it won't hurt, but I have no idea how that got there :sweat_smile:

na--: Fixed with the [latest force-push](https://github.com/grafana/k6/compare/a42bdc7a4890271f333d24dc070459bff35c6e30..9f050f7298f802e394a2b08d24e7bd40f8a4df94)

codebien (approved):

oleiade (approved):

PR #2813, opened 2022-12-08
Move most capabilities out of the `core.Engine` in preparation for removal

na--:
This built on top of https://github.com/grafana/k6/pull/2812 and is the penultimate PR before https://github.com/grafana/k6/issues/1889 :tada: :crossed_fingers: 

It splits off as much of the capabilities of the `Engine` as it's possible to do and keep most of its tests (_mostly_) intact :sweat_smile: Commits should be atomic and it might be easier to review the PR commit by commit. I can even split this into smaller PRs, if needs be, though I think grouping these commits together in one PR and completely removing the `Engine` in another PR makes the most sense.

Issue #140, opened 2017-03-07
Clustering and distributed execution


uppfinnarn:
This is something that isn't top priority at the moment, but it's going to take a lot of design work, so I'd like to get the ball rolling on the actual planning.

My current idea is to have a `k6 hive` command (name subject to change), which hooks up to [etcd](https://github.com/coreos/etcd) - a lovely piece of software that can handle most of the heavy lifting around clustering, and is also the backbone that makes among other things Kubernetes tick.

Each instance registers itself, along with how many VUs it can handle, exposes an API to talk to the cluster, and triggers a leader election. The information registered in etcd might have a structure like:

* `/loadimpact/k6/nodes/nodeA` - node data (JSON)
* `/loadimpact/k6/nodes/nodeB` - node data (JSON)

Running a test on the cluster is a matter of calling `k6 run --remote=https://a.node.address/ script.js`. This causes it to, instead of running the test locally, roll up all the data and files needed and push them to the cluster, where they're stored in etcd - available from each node.

* `/loadimpact/k6/test` - test data (JSON)
* `/loadimpact/k6/test/src/...` - root for the pushed filesystem

When test data is loaded, each VU instantiates an engine and its maximum number of VUs right away, and watches its own registry information for changes. The elected leader then takes care of patching other nodes to distribute VUs evenly across the cluster.

neilstuartcraig:
WRT #202 / request for comments:

I don't have direct experience with etcd so i don't have an opinion there but the notion of using existing, proven software is clearly sound so i'd welcome that.

My initial thought is, how would someone know how many VU's a machine can handle? If there's an automated estimation based on system resources (though obv. e.g. a CPU core on AWS is definitely not === a CPU core on dedicated hardware) then that'd help to at least provide consistency which would be good. I'm aware though that it'd quite easy to overload a load generator with work and thus skew its output as it would lack the system resources to measure accurately. However it's implemented, i would imagine it's going to be necessary to allow users to set/amend the VU capability and a good user guide would help a lot - i.e. defining a way in which users can calculate/estimate - that might be the best way to start actually, keeping it simple and iterating/adding from there.

liclac:
Honestly, your best shot is probably trial and error. The limiting factor is not typically CPU power, but rather local socket usage, and to some lesser extent, RAM usage, both of which vary slightly between scripts. A good start would be just split your desired number of VUs across as many hosts as you want and seeing if it flies or not.

carlwright4:
(#140)

[...]

PR #2815, opened 2022-12-08
Remove the `core.Engine` :tada: 

na--:
This PR builds on top of https://github.com/grafana/k6/pull/2813 and closes https://github.com/grafana/k6/issues/1889, finally :tada: :confetti_ball: It also leaves us in a much better position to implement https://github.com/grafana/k6/issues/1342, https://github.com/grafana/k6/issues/140, https://github.com/grafana/k6/issues/2804, https://github.com/grafana/k6/issues/2430, and a myriad of other issues :tada: 

Some of the less important tests still need to be ported, so it it still technically a draft PR that's not quite ready to merge as it is. Still, given the fact that the new integration tests all work fine, I am fairly confident there aren't any major problems :crossed_fingers: I also don't intend to make any big changes to the code here, so it's more than ready for review.


cmd/run.go
```diff
@@ -37,12 +41,36 @@ type cmdRun struct {
 //
 //nolint:funlen,gocognit,gocyclo,cyclop
 func (c *cmdRun) run(cmd *cobra.Command, args []string) (err error) {
+	var logger logrus.FieldLogger = c.gs.logger
+	defer func() {
+		logger.Debugf("Everything has finished, exiting k6 with error '%s'!", err)
```
codebien, line 46: >DEBU[0004] Everything has finished, exiting k6 with error '%!s(<nil>)'!

Do we want this output format? I think it could generate confusion. Maybe add the error only if this is not nil.

cmd/run.go
```diff
@@ -105,17 +113,95 @@ func (c *cmdRun) run(cmd *cobra.Command, args []string) (err error) {
...
 	if err != nil {
 		return err
 	}
-	engine.AbortFn = runSubAbort
+	defer func() {
+		logger.Debug("Stopping outputs...")
+		// We call waitOutputsFlushed() below because the threshold calculations
+		// need all of the metrics to be sent to the MetricsEngine before we can
+		// calculate them one last time. We need the threshold calculated here,
+		// since they may change the run status for the outputs.
```
codebien, line 170: Instead, to do this by the output manager, could we do an explicit call to the ingester? I think it would simplify a bit the logic.

output/manager.go
```diff
@@ -1,10 +1,16 @@
...
+	"sync"
+	"time"
+
 	"github.com/sirupsen/logrus"
 	"go.k6.io/k6/metrics"
 )
 
+// TODO: completely get rid of this, see https://github.com/grafana/k6/issues/2430
+const sendBatchToOutputsRate = 50 * time.Millisecond
+
```
codebien, line 13: What do we plan instead of this? Would we replace it with a pull-based model?

metrics/engine/engine.go
```diff
@@ -245,7 +246,9 @@ func (me *MetricsEngine) evaluateThresholds(ignoreEmptySinks bool) (breachedTher
 			shouldAbort = true
 		}
 	}
-
+	if len(breachedThersholds) > 0 {
+		sort.Strings(breachedThersholds)
+	}
 	me.logger.Debugf("Thresholds on %d metrics breached: %v", len(breachedThersholds), breachedThersholds)
```
codebien, line 252: Should we move the other operations below in the if block?

errext/exitcodes/codes.go
```diff
@@ -14,11 +14,10 @@ const (
 	SetupTimeout             ExitCode = 100
 	TeardownTimeout          ExitCode = 101
 	GenericTimeout           ExitCode = 102 // TODO: remove?
-	GenericEngine            ExitCode = 103 // TODO: remove after https://github.com/grafana/k6/issues/2804
+	CannotStartRESTAPI       ExitCode = 103
 	InvalidConfig            ExitCode = 104
 	ExternalAbort            ExitCode = 105
-	CannotStartRESTAPI       ExitCode = 106
+	ScriptStoppedFromRESTAPI ExitCode = 106
```
codebien, line 20: Should it be `ScriptAbortedByRESTAPI`?

cmd/run.go
```diff
@@ -105,17 +113,95 @@ func (c *cmdRun) run(cmd *cobra.Command, args []string) (err error) {
...
+	// flushed their samples and when they have stopped in the defer statements.
+	initBar.Modify(pb.WithConstProgress(0, "Starting outputs"))
+	outputManager := output.NewManager(outputs, logger, func(err error) {
+		if err != nil {
+			logger.WithError(err).Error("Received error to stop from output")
+		}
+		// TODO: attach run status and exit code?
+		runAbort(err)
+	})
+	samples := make(chan metrics.SampleContainer, test.derivedConfig.MetricSamplesBufferSize.Int64)
```
codebien, line 160: It impacts a lot of parts, so I'm not sure. Could it now be directly a field of the OutputManager, instead of having it injected? Maybe it helps also to centralize control over it (the stop operation), and we can just pass it to the execution scheduler.

cmd/run.go
```diff
@@ -37,12 +41,36 @@ type cmdRun struct {
...
+
+	// lingerCtx is cancelled by Ctrl+C, and is used to wait for that event when
+	// k6 was started with the --linger option.
+	lingerCtx, lingerCancel := context.WithCancel(globalCtx)
+	defer lingerCancel()
+
+	// runCtx is used for the test run execution and is created with the special
+	// execution.NewTestRunContext() function so that it can be aborted even
+	// from sub-contexts while also attaching a reason for the abort.
+	runCtx, runAbort := execution.NewTestRunContext(lingerCtx, logger)
```
codebien, line 61: I think we are not cancelling the `runCtx` so the Report's goroutine is stuck waiting for it.

codebien (changes requested):

metrics/engine/engine_test.go
```diff
@@ -1,131 +1,10 @@
...
-	"go.k6.io/k6/loader"
-	"go.k6.io/k6/metrics"
-	"go.k6.io/k6/output"
-)
+package engine
 
-const isWindows = runtime.GOOS == "windows"
-
-// TODO: completely rewrite all of these tests
+// TODO: refactor and move the tests that still make sense in other packages
```
codebien, line 3: I've just seen this, add a comment as a reminder

cmd/run.go
na--: hmm, maybe, not sure it would be simpler though :thinking: here in the `defer` statements, we just have the following sequence after the test run finishes:
1. close the `samples` channel
2. wait for all outputs to finish flushing their metrics
3. finalize the thresholds (unless `--no-thresholds` was used)
4. stop all outputs

to me, this makes sense even if we completely ignore thresholds :thinking: and I am not sure how we can simplify it, when the piping of the metrics also happens by the `output.Manager` :thinking:

output/manager.go
na--: not sure, maybe :thinking: we'll certainly need to do some benchmarking before, maybe the current solution is the most efficient one, who knows... :thinking:

metrics/engine/engine.go
na--: hmm maybe the logging, with a separate log if no thresholds were breached :thinking: but the `atomic.StoreUint32()` and return should stay outside of the `if`, right?

errext/exitcodes/codes.go
na--: maybe, `ScriptStoppedFromRESTAPI` doesn't "sound" quite right, though `ScriptAbortedByRESTAPI` also doesn't seem correct - it is the user that stops the script, the REST API is just the means to do it :thinking: any other ideas?

metrics/engine/engine.go
codebien: Yeah, this comment is for logline.

errext/exitcodes/codes.go
codebien: > it is the user that stops the script, the REST API is just the means to do it
At this point, is it required to have a dedicated exit code? Could we use ExternalAbort?

na--: It's not required, but I don't see why we shouldn't have a dedicated exit cod. Since it's so easy to make the differentiation, it doesn't hurt to have it - maybe it matters for some user, or for us in `k6 cloud`, maybe it helps us with debugging some complex issue :man_shrugging:

cmd/run.go
na--: hmm I think the bug is that it actually tries to wait for `<-runCtx.Done()` at all :thinking: 

https://github.com/grafana/k6/blob/3c00d584bd1e87c8350d75fc168cfdbe6f9a3104/cmd/run.go#L285-L302

We call it after the test run has already finished, so there is no point in waiting for `<-runCtx.Done()`, we should just send the report :thinking: This bug is just a remnant from the old way convoluted way...

And we can wait for it to be done with a separate `defer`, "earlier" in the process. That is, k6 can continue crunching metrics, showing the summary, stopping outputs, etc., while the report is being sent.

na--: I think this should now be fixed by the [latest force-push](https://github.com/grafana/k6/compare/3c00d584bd1e87c8350d75fc168cfdbe6f9a3104..6f6b105ff041b2c317fa909d959bb77ea0946773)?

cmd/run.go
na--: Resolved by the [latest force-push](https://github.com/grafana/k6/compare/3c00d584bd1e87c8350d75fc168cfdbe6f9a3104..6f6b105ff041b2c317fa909d959bb77ea0946773)

cmd/run.go
na--: Hmm maybe, though I'd probably prefer this to be done in a separate PR :thinking: it's not very straightforward because of the thresholds, which need the sequence from https://github.com/grafana/k6/pull/2815#discussion_r1085273225, but the `OutputManager` doesn't know that the `MetricsEngine`'s `Ingester` is in any way a special output... 

With this PR, `samples` is created here, then it is passed around wherever it is needed, and then finally it is closed in a `defer` further down this same function. That is a relatively clean and Go-idiomatic bit of code and already a massive improvement to the previous state :sweat_smile: So any further improvements can probably wait for things to settle down a bit.

cmd/run.go
codebien: It is yes.

codebien (commented):
LGTM :tada:

na--:
I split off the first commit into a separate PR (https://github.com/grafana/k6/pull/2885) to fix a recent regression introduced by https://github.com/grafana/k6/commit/19dd50593b90a79fa1883e248c63d385dbe7a8b9 from https://github.com/grafana/k6/pull/2813. See the new PR for more details, and I'll rebase this PR on top of it.

cmd/run.go
```diff
@@ -177,9 +177,12 @@ func (c *cmdRun) run(cmd *cobra.Command, args []string) (err error) {
 		stopOutputs(err)
 	}()
 
-	if !testRunState.RuntimeOptions.NoThresholds.Bool {
+	if !testRunState.RuntimeOptions.NoThresholds.Bool { //nolint:nestif
 		finalizeThresholds := metricsEngine.StartThresholdCalculations(runAbort)
 		defer func() {
+			if finalizeThresholds == nil {
```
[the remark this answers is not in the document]
na--: Resolved by [this commit](https://github.com/grafana/k6/pull/2815/commits/e73e80455d9070a66c86cf62f408f48730e67297) (which is easiest to review [with whitespace differences off](https://github.com/grafana/k6/pull/2815/commits/e73e80455d9070a66c86cf62f408f48730e67297?diff=unified&w=1))

output/manager.go
codebien: I linked the code where we mention this in the issue's description

metrics/engine/engine_test.go
codebien: https://github.com/grafana/k6/pull/2878

cmd/run.go
codebien: I think it could be helpful to add these 4 points to the comment so we have one central place where we can read the flow without going across the entire code of the Run method. I think it could be helpful when we will work on #2430.

codebien (approved):
LGTM

mstoykov (approved):

cmd/run.go
na--: :+1:, but let's add this in the PR that converts the remaining tests, I'll merge this one shortly

Issue #1996, opened 2021-04-28
Reproducible builds


imiric:
As mentioned in #1247 and https://github.com/loadimpact/k6/pull/1245#discussion_r346252753, we'd like our binary builds to be reproducible/deterministic, so that the same set of inputs produces the same bit-for-bit binary output.

## Feature Description
This would give the assurance that building from a specific commit produces the same binary as the one released publicly. In case of the Dockerfile build, it would guarantee that the binary in the Docker image is exactly the same as the one built by `build-release.sh`.

## Suggested Solution (optional)
Right now we use the `go build -trimpath` flag which helps:
https://github.com/k6io/k6/blob/14c6cd9f0a8e1923a1dc8aebf304bbeffd58581e/build-release.sh#L19
https://github.com/k6io/k6/blob/14c6cd9f0a8e1923a1dc8aebf304bbeffd58581e/Dockerfile#L5

... but it likely doesn't work since we inject a date in the version string:
https://github.com/k6io/k6/blob/14c6cd9f0a8e1923a1dc8aebf304bbeffd58581e/build-release.sh#L12

So maybe this date should be set beforehand and shared for both the Dockerfile and `build-release.sh`? I doubt we want to get rid of it entirely since it's very convenient to have in the version string.

na--:
Instead of the build date, we can use the date of the git tag: https://stackoverflow.com/questions/13208734/get-the-time-and-date-of-git-tags/13208830

joanlopez:
As discussed in #1997, the `k6 version` output no longer includes the date, and the `VERSION_DETAILS` pointed out in the issue's description is no longer used. So, considering that we already have `-trimpath` in-place, as stated above, and that, as far as I know, there's no other temporal reference, I think we can consider this as solved.

If you find any other aspect of the resulting binary that varies over builds, and that would be nice to get rid of, for the sake of consistency/reproducibility, please open another issue.

Thanks! 🙇

PR #2821, opened 2022-12-16
Refactor integration tests

### LINKED ISSUE #2459: Move integration tests to a dedicated package

imiric:
This does the changes discussed in #2459, see the commits for details. I'm adding everyone as a reviewer because it's a large refactor (splitting it into more PRs wouldn't help much), but @na-- should give the final approval.

Closes #2459


main.go
```diff
@@ -1,9 +1,12 @@
...
 import (
+	"context"
+
 	"go.k6.io/k6/cmd"
+	"go.k6.io/k6/cmd/state"
 )
 
 func main() {
-	cmd.Execute()
+	cmd.Execute(state.NewGlobalState(context.Background()))
```
imiric, line 11: So this change breaks xk6. It's a simple fix, and I'll push the change later today.

I needed to do this because tests in `cmd/tests` must call an exported interface from `cmd` to actually execute the command. The way they did it while they were in `cmd` is with `newRootCommand(ts.GlobalState).execute()`, but exporting that involved exporting a whole lot of related types and methods. So it was easier to just change the signature of this already exported `cmd.Execute()` function.

cmd/tests/cmd_test.go
```diff
@@ -1038,23 +1065,24 @@ func TestAbortedByScriptAbortInTeardown(t *testing.T) {
 }
 
 func testAbortedByScriptTestAbort(
-	t *testing.T, shouldHaveMetrics bool, script []byte, runTest func(*testing.T, *globalTestState),
-) *globalTestState {
+	//nolint: unparam  // will be fixed by https://github.com/grafana/k6/pull/2800
+	t *testing.T, shouldHaveMetrics bool, script []byte, runTest func(*testing.T, *state.GlobalTestState),
+) *state.GlobalTestState {
```
imiric, line 1070: @na-- There were two linter issues here.

- `unparam  testAbortedByScriptTestAbort - shouldHaveMetrics always receives true`
   Which I decided to silence until #2800 is merged, which does pass `false`.

- And the one that currently fails:
  `unparam  testAbortedByScriptTestAbort - result 0 (*go.k6.io/k6/cmd/state.GlobalTestState) is never used`

  Which I think is pointing out that `GlobalTestState.ExpectedExitCode` is always always `0` and never checked?

  I may be wrong, and in any case, I'm not sure how to fix it, or if we should silence it as well. Do you have an idea?

BTW, this is partly why I'm not a fan of overusing helper functions in tests, which I pointed out in [this PR](https://github.com/grafana/k6/pull/2800#discussion_r1048641706). They make tests more difficult to follow, and result in cryptic issues like this.

oleiade (approved):
LGTM 👍🏻 🎉 🙇🏻

main.go
imiric: This is fixed by https://github.com/grafana/xk6/commit/52e45894f2d4509d8c877fe4af9289d52b5ed17d and and3ebefcebc, but merging these changes would still break building k6 binaries with versions prior to this change.

So I think we should either avoid breaking this, or handle the failure in xk6 in a more sophisticated way. E.g. try building with both `cmd.Execute()` and `cmd.Execute(state.NewGlobalState(context.Background()))`?

WDYT?

mstoykov: 1. add `ExecuteWIthGlobalState` which takes the state and leave the other one without it but calling the new one.
2. (the more hacky IMO) have the GlobalState be "not required" by being variadic(`func Execute(...GlobalState)`)


In both cases sprinkled with some amount of comments on why this is needed.

imiric: Thanks. Yeah, I guess the first is the way to go. :+1:

ui/console/writer.go
```diff
@@ -0,0 +1,41 @@
...
+import (
+	"bytes"
+	"io"
+	"os"
+	"sync"
+)
+
+// Writer syncs writes with a mutex and, if the output is a TTY, clears before
+// newlines.
+type Writer struct {
```
imiric, line 12: This is the same `consoleWriter`, but since it needed to be exported so that it could be used in `cmd/tests/` and elsewhere, it's the only thing in this new `ui/console` package. Otherwise the functionality is identical, and we can bring `console.Console` back later.

cmd/tests/cmd_test.go
imiric: I silenced the second error as well :man_shrugging:

main.go
imiric: Resolved by 96c1cca7.

mstoykov (approved):
LGTM! I tried to check that the previously found issues are now fixed but I would prefer that @na-- goes through it as well.

My only actionable feedback is -  I would've amended the original commit to add `ExecuteWithGlobalState`. IMO this would've been cleaner and I already think it is required to go through the PR commit by commit, so this makes it easier to not come back to the same change later.

imiric:
Sorry about the force-push again, but I squashed the latest two commits in previous ones, as suggested by Mihail, and resolved a merge conflict that GitHub couldn't, but somehow Git did automatically :man_shrugging:

lib/env.go
```diff
@@ -0,0 +1,22 @@
...
+// BuildEnvMap returns a map from raw environment values, such as returned from
+// os.Environ().
+func BuildEnvMap(environ []string) map[string]string {
+	env := make(map[string]string, len(environ))
+	for _, kv := range environ {
+		k, v := ParseEnvKeyValue(kv)
+		env[k] = v
+	}
+	return env
+}
```
na--, line 22: Moving these here doesn't seem like an improvement. 

First, we are trying to reduce the size of `lib/`, since it [itself is an antipattern](https://go.dev/blog/package-names), not increase it.

But, more importantly, these functions are not widely used enough to move them here. `ParseEnvKeyValue()` is only used here, in `BuildEnvMap()`, and in a single other place in `cmd/`. `BuildEnvMap()` itself is only used in `cmd/state` and in a test (where it's a convenience. not a necessity).

cmd/tests/cmd_test.go
```diff
@@ -22,160 +22,187 @@ import (
...
 	t.Parallel()
 
-	ts := newGlobalTestState(t)
-	ts.args = []string{"k6", "version"}
-	newRootCommand(ts.globalState).execute()
+	ts := state.NewGlobalTestState(t)
+	ts.CmdArgs = []string{"k6", "version"}
+	cmd.ExecuteWithGlobalState(ts.GlobalState)
 
-	stdOut := ts.stdOut.String()
```
na--, line 39: Is there a reason to rename `stdOut` to `stdout` everywhere? If this was new code, I won't care too much one way or the other, but in this PR, a big chunk of huge diff is just this one change and it's distracting, so my vote would be to keep `stdOut` unchanged

cmd/state/test_state.go
```diff
@@ -0,0 +1,155 @@
...
+	LoggerHook     *testutils.SimpleLogrusHook
+
+	Cwd string
+
+	ExpectedExitCode int
+}
+
+// NewGlobalTestState returns an initialized GlobalTestState, mocking all
+// GlobalState fields for use in tests.
+func NewGlobalTestState(t *testing.T) *GlobalTestState {
```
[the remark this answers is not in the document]
na--: However, since we import this from other packages, it can't be in a `something_test.go` file, right? So we probably need a separate package for it, similar to what we have done with `testutils` and `modulestest`, see https://github.com/grafana/k6/tree/master/lib/testutils and https://github.com/grafana/k6/tree/master/js/modulestest

cmd/tests/cmd_test.go
mstoykov: Yeah I highly prefer to not mass rename variable/names at the same time you are doing actual refactoring where you move stuff around. Especially in this big diffs.

Renaming the thing that is moved is usually okay, but other stuff usually just make the changes harder to follow.

The thing *I* had problems was the rename of `testState` to `ts` - yes it is more consistent, but it does introduce a change *everywhere* it is used even in the places nothing else changes. And it is used quite a lot ;)

imiric: This PR standardizes the capitalization of `GlobalState.Std{out,err}` to match `os.Std{out,err}`, so making local vars match that just made sense. I didn't do an exhaustive job changing it everywhere, but I would still prefer to keep it as is, instead of doing even more work now to revert it back to an IMO awkward capitalization.

Same with `s/testState/ts/`, and `s/globalState/gs/`. Previously we used both, and this was done to standardize things.

It's not unbearably distracting, right? After all, refactoring is mostly renaming things to begin with... :smile:

But I'll keep your comments in mind for future PRs, and do these in separate commits.

na--: But these are not `GlobalState.Stdout`, this is a local variable in a whole bunch of tests :confused:

lib/env.go
imiric: I just moved them to `cmd/state` in 504f134a. Let me know if you have a better idea.

cmd/tests/cmd_test.go
imiric: Right. I made the local variable match the lowercase change of `{out,err}` of the `GlobalState` fields. `stdOut := ts.Stdout` would look weird otherwise.

I feel like we're bikeshedding at this point. :sweat_smile: If you feel strongly about this, I can change it back, but like I said, it's more work for something that doesn't make sense.

cmd/state/test_state.go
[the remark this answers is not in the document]
na--: > Do you know how we could confirm this?

`go list -f '{{range $imp := .Imports}}{{printf "%s\n" $imp}}{{end}}' | sort` showed the `testing` package was imported and now it's not, but I didn't notice the binary size actually decrease after pulling your latest changes :thinking: `go version -m /path/to/k6/binary` doesn't actually show stdlib sub-packages were included :disappointed:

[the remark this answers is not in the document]
na--: not sure how the magic of https://github.com/jondot/goweight works and how accurate it is, but running it in the repo before and after your change actually shows a huge difference:
```
% diff pre.txt post.txt 
28d27
<   1.6 MB github.com/stretchr/testify/assert
32d30
<   1.3 MB testing
54d51
<   1.0 MB github.com/stretchr/testify/require
115d111
<   483 kB github.com/davecgh/go-spew/spew
158d153
<   319 kB net/http/httptest
194d188
<   222 kB github.com/pmezard/go-difflib/difflib
211d204
<   183 kB go.k6.io/k6/cmd/state
221d213
<   174 kB go.k6.io/k6/lib/testutils
267a260
>   115 kB go.k6.io/k6/cmd/state
310d302
<    75 kB runtime/trace
407d398
<    14 kB internal/sysinfo
414d404
<    10 kB net/http/internal/testcert
```

So, yeah, it seems like it was both a real problem and that it is now fixed :tada:

lib/env.go
na--: Still think they should probably be unexported and copied, but the bigger issue was having them in `lib/`, so I am reasonably happy with this compromise :sweat_smile:

cmd/tests/cmd_test.go
na--: Fair enough, it's not that big of a big deal, but it makes the PR review more difficult for no apparent benefit. I am usually on the other side of these discussions, since I am the one who usually made huge PRs (though I am trying to be better!), so I'll shut up this time :sweat_smile:

cmd/state/test_state.go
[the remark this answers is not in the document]
imiric: Thanks, I was looking for something that could be run from the repo root to list imports of all sub-packages, as they would be assembled by `go build`.

The size does decrease by a few kBs:

```
42029102 k6-03f8cee3
42004497 k6-24717844
```

So maybe `testing` is not that heavy(?).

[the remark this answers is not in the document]
imiric: Ah, cheers for mentioning `goweight`. :bookmark:

But it seems that `go build` strips out things like `testing` then?

cmd/tests/cmd_test.go
imiric: Alright, I'll keep this in mind for future PRs, and do the noisy renames in separate commits.

cmd/state/test_state.go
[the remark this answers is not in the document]
na--: there is actually a tiny difference in binary size
```
41915384  k6-post
41941793  k6-pre
```

I have no idea why it's only 25kb though, maybe Go is really, really go at tree shaking :sweat_smile: though given the things you can do with reflection, I am actually leaning towards a more uncharitable interpretation - that we are already somehow including the `testing` package and `goweight` just doesn't detect it :thinking:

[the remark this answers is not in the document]
imiric: I actually find it more interesting that our binaries differ by ~1MB! :thinking:

I had added #1996 to discuss for the upcoming cycle. It will never be a good time to tackle it, but the sooner we do it, the better.

cmd/tests/test_state.go
```diff
@@ -0,0 +1,156 @@
...
+
+// TestOSFileR is the mock implementation of stdin.
+type TestOSFileR struct {
+	io.Reader
+}
+
+// Fd returns a mock file descriptor ID.
+func (f *TestOSFileR) Fd() uintptr {
+	return 0
+}
```
na--, line 135: I think these are not used anywhere in this PR, right?

cmd/tests/doc.go
```diff
@@ -0,0 +1,8 @@
+// Package tests contains integration tests that run k6 commands, and interact
+// with standard I/O streams. They're the highest level tests we have, just
+// below E2E tests that execute the k6 binary. Since they initialize all
+// internal k6 components similarly to how a user would, they're very useful,
+// but also very expensive to run. They're also brittle, as they depend on large
+// parts of the codebase. When in doubt, prefer adding lower-level unit tests
+// first, and an integration test only if necessary.
```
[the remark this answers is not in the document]
imiric: The APIs certainly won't change much, but since they test a large portions of the system, integration tests are brittle by definition. Any issue in the underlying packages might trigger a failure in _all_ integration tests that depend on it. The failure matrix is much larger with integration tests than for unit tests.

In addition to this, integration tests are _much_ more expensive to run. I really don't want us to rely solely on them for any new features, since I would prefer to run them only occasionally locally (say, before `git push`). Whereas I can run a batch of unit tests much more quickly, and rely on them as a more frequent check while developing.

Related to this, prefering integration tests to unit tests goes against the [traditional test pyramid](https://martinfowler.com/articles/practical-test-pyramid.html). I actually don't think we need to prefer one over the other at all. All levels of testing complement each other, but it's clear that we should have _mostly_ unit tests, _many_ integration tests, and _some_ E2E tests.

The fact some of our unit tests are unreliable and flaky is not a reason to avoid them, but a reason to make them more reliable. Having to refactor them more frequently is the price of admission, and not a reason to prefer higher-level tests.

cmd/tests/tests_test.go
```diff
@@ -0,0 +1,53 @@
...
+
+import (
+	"fmt"
+	"net/http"
+	"os"
+	"sync/atomic"
+	"testing"
+)
+
+type blockingTransport struct {
```
na--, line 12: hmm currently this thing and `TestMain` are duplicated in `cmd/root_test.go` 

It is probably a good thing to have them in both places, though maybe they can be exported from this package and imported in `cmd/root_test.go`? :thinking:

cmd/tests/doc.go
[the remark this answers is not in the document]
na--: > The APIs certainly won't change much, but since they test a large portions of the system, integration tests are brittle by definition. Any issue in the underlying packages might trigger a failure in all integration tests that depend on it. The failure matrix is much larger with integration tests than for unit tests.

No, it won't, these integration tests don't actually use most of the underlying packages directly :sweat_smile: If you can compile the `k6` binary, you should pretty much be able to run the integration tests...

Again, these tests basically rely only on the user-facing APIs, which we try very, very hard to keep somewhat stable. They are the very opposite of brittle.

> I actually don't think we need to prefer one over the other at all. All levels of testing complement each other, but it's clear that we should have mostly unit tests, many integration tests, and some E2E tests.
> 
> The fact some of our unit tests are unreliable and flaky is not a reason to avoid them, but a reason to make them more reliable. Having to refactor them more frequently is the price of admission, and not a reason to prefer higher-level tests.

Mostly agree with this, though with a clarification. Low-level **unit** tests that verify a single API are great. High-level integration tests like these that verify the whole application (basically from the user's perspective) are also great. The more you go towards the middle, the worse things get. 

The most horrible and flaky tests we have are these mid-level tests that first have to do a whole lot of setup and initialization of a whole bunch of components and mocks. And then they have to inject and verify very complicated things. Basically, the old `Engine` and `ExecutionScheduler` tests, a bunch of the tests around the executors and `js/`. Most of them would be better off as integration tests, since they try to set up the whole world before the test can be executed, and they do it by somewhat replicating what happens during a normal k6 run, but not quite exactly the same. So, any API change breaks them, and they often are not very realistic or good tests, i.e. they are the brittle ones.

cmd/tests/cmd_test.go
```diff
@@ -22,160 +22,186 @@ import (
...
 	"go.k6.io/k6/cloudapi"
+	"go.k6.io/k6/cmd"
 	"go.k6.io/k6/errext/exitcodes"
 	"go.k6.io/k6/lib"
 	"go.k6.io/k6/lib/consts"
 	"go.k6.io/k6/lib/testutils"
 	"go.k6.io/k6/lib/testutils/httpmultibin"
 )
 
+func TestDeprecatedOptionWarning(t *testing.T) {
```
na--, line 33: This test also seems duplicated in `cmd/root_test.go`, in `TestDeprecatedOptionWarning`. I don't have a strong opinion where it should be, but it shouldn't be in both places. And given that it _can_ be in `cmd/` and was there originally, my vote would be to delete this from here.

imiric: :facepalm: I thought I removed all the duplication. Sorry, bad conflict resolution, will fix.

cmd/tests/doc.go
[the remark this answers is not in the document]
imiric: > No, it won't, these integration tests don't actually use most of the underlying packages directly :sweat_smile: If you can compile the k6 binary, you should pretty much be able to run the integration tests...

> Again, these tests basically rely only on the user-facing APIs, which we try very, very hard to keep somewhat stable. They are the very opposite of brittle.

I think we have different definitions of "brittle" :smile:

Having stable APIs doesn't matter if, e.g. `cmdRun.run()` returns an error, which can happen _for any reason_ in _any of the underlying packages_. The `k6` binary will build just fine, but the error will happen at runtime. You can try it yourself, just return a fake error from `cmdRun.run()` and see all integration tests fail. This is a contrived example, but a failure could occur deep within some dependent module, and trigger similar broad failures.

This is the opposite of a unit test where the only reason it will fail (or _should_ fail, assuming it's not flaky) is if the unit under test fails. It's not dependent on any other part of the system, and so, by definition, unit test are much more stable than higher level tests.

The reason you want to have both is because the integration test failure often doesn't tell you anything about the root cause. Preferably you'd want to have at least one unit test failure that would pinpoint exactly the root cause, and then you wouldn't really spend time debugging integration test failures at all. This is why it's important to have both. Lower level tests to debug and fix things quickly. And higher level tests to ensure components _integrate_ well, and E2E tests that it all works from the user's perspective.

BTW, as you mentioned, what we're currently calling integration tests are closer to E2E tests, just without actually running the k6 binary. So it's even a worse idea to suggest we should have a lot of them, or worse, prefer them over lower level tests.

I see your point about middle-layer tests (i.e. integration tests) being troublesome, but just as frequent refactoring is a pain-point for unit tests, mocking and test setup are a chore for integration tests. So I'm against trying to rewrite these as even higher-level almost-E2E tests to avoid the mocking, but try to make mockable APIs to begin with, and test helpers as we do now to assist with the setup boilerplate.

I think we fundamentally disagree on testing strategies, but a) we won't make much progress discussing it here, and b) I'd like to involve more people in this discussion. So let's schedule it for tomorrow's sync?

cmd/tests/test_state.go
imiric: Right. Removed in 737d26a0.

cmd/tests/tests_test.go
imiric: Right, I overlooked this.

Instead of exporting only `blockingTransport` and duplicating `TestMain` in both packages, I added a new function `tests.Main`, and called it from each package's `TestMain()`. See 895a657a.

This way we don't have to redefine all the hosts logic and goleak check, which is more significant.

cmd/tests/cmd_test.go
imiric: Fixed in 7216a745.

cmd/tests/doc.go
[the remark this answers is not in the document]
na--: Yes, we apparently have different preferences here, and :+1: for discussing it more with the rest of the team. 

Until then, can you remove the opinion parts of this we disagree on, so only the more factual bits on which we have consensus remain? This will allow us to merge the PR more quickly, which is important, since it's blocking a few things.

```suggestion
// but also very expensive to run. 
```

na-- (approved):
Everything except https://github.com/grafana/k6/pull/2821#discussion_r1072424210 LGTM now :heart:

imiric:
OK, one last force-push to squash as much as possible. Hopefully didn't screw anything up, so give it one last glance. :pray:

na-- (approved):
:heart:

PR #2828, opened 2022-12-21
go.mod: merge indirect require section

zchee:
go.mod: merge indirect require section.


na-- (approved):
Thanks!

imiric (approved):
Thanks! We don't really modify `go.mod` manually, and for some reason `go mod tidy` didn't make this change.

PR #2830, opened 2022-12-21
Goja async await

mstoykov:


olegbespalov (dismissed):

js/tc39/tc39_test.go
```diff
@@ -503,9 +508,10 @@ func (ctx *tc39TestCtx) runTC39File(name string, t testing.TB) {
 		// t.Logf("Running strict test: %s", name)
 		ctx.runTC39Test(t, name, src, meta, true)
 	} else { // Run test in non strict mode only if we won't run them in strict
+		// NOTE: in practice k6 does not run non strict code at all so this tests make no sense
 		// TODO uncomment the if above and delete this else so we run both parts when the tests
 		// don't take forever
-		ctx.runTC39Test(t, name, src, meta, false)
+		// ctx.runTC39Test(t, name, src, meta, false)
```
na--, line 514: hmm ok, but does running these result in a ton more failures? or does it just take longer? 

I don't have a firm opinion on whether we should run or should not run these, but if we aren't going to ever run them, my vote would be to delete the whole `else` clause and just comment about strict mode above

na-- (commented):
Mostly LGTM, but can we please add a new test or edit an existing test, either in `js/` or in the integration test, with even a single `async` function? :sweat_smile: 

The TC39 tests are awesome, but they are skipped by default. It seems useful to have some minor k6-specific test that will fail in normal `go test ./...` in our local machines.

js/tc39/tc39_test.go
mstoykov: reenabled it but dropped the generator ones with [e524b0e](https://github.com/grafana/k6/pull/2830/commits/e524b0e8e926bc0a7a6d181ed9ce513a84c561ed)

mstoykov:
> It seems useful to have some minor k6-specific test that will fail in normal go test ./... in our local machines. 

I would argue those will come when we start adding specific stuff for k6. But have added a small event loop test

na-- (approved):

olegbespalov (approved):

PR #2833, opened 2022-12-22
Remove detected goroutine leaks

codebien:


na-- (approved):
:heart:

lib/testutils/httpmultibin/httpmultibin.go
```diff
@@ -94,6 +94,12 @@ func getWebsocketHandler(echo bool, closePrematurely bool) http.Handler {
 		if err != nil {
 			return
 		}
+		defer func() {
+			err := conn.Close()
+			if err != nil {
+				panic(fmt.Errorf("close WS conn in httpmultibin failed: %w", err))
```
na--, line 100: Instead of a `panic()`, I think it will be better to slightly refactor `getWebsocketHandler()` to accept a `t testing.TB` parameter, and then you can simply use `require.NoError(t, err)` or something like that.

codebien: This a good idea :+1:

codebien: I reverted because we have some cases where we test abnormal closing, so it is expected to get errors there.

cmd/stdlog_integration_test.go
```diff
@@ -0,0 +1,48 @@
...
+func TestStdLogOutputIsSet(t *testing.T) {
+	tb := httpmultibin.NewHTTPMultiBin(t)
+	ts := newGlobalTestState(t)
+
+	// Sometimes the Go runtime uses the standard log output to
+	// log some messages directly.
+	// It does when an invalid char is found in a Cookie.
+	// Check for details https://github.com/grafana/k6/issues/711#issue-341414887
+	ts.stdIn = bytes.NewReader([]byte(tb.Replacer.Replace(`
+import http from 'k6/http';
```
na--, line 24: The indentation looks wrong here

cmd/root.go
```diff
@@ -241,7 +241,18 @@ func (c *rootCommand) persistentPreRunE(cmd *cobra.Command, args []string) error
...
+	// log some messages directly.
+	// It does when an invalid char is found in a Cookie.
+	// Check for details https://github.com/grafana/k6/issues/711#issue-341414887
+	stdlog.SetOutput(func() io.Writer {
+		w := c.globalState.logger.Writer()
+		go func() {
+			<-c.globalState.ctx.Done()
+			_ = w.Close()
+		}()
+		return w
```
na--, line 254: I am not sure this is sufficient and won't cause problems :thinking: I can think of 2 potential problems. :thinking: 

First, if something somehow logged to the stdlib logger after this writer was closed, I imagine this will panic, which is not ideal. It should not happen, but if it happens, it's better to not panic.

Second, I am not sure this internal goroutine will never be detected by the goroutine detector :thinking:  After all, it waits for the `globalState.ctx` to be done and closes something, but nothing actually waits for the goroutine to be finished before `execute()` is done.

na--: Off the top of my head, better approach would be a `io.Writer` that either discards incoming logs or falls back to the `fallbackLogger` once `globalState.ctx` is done, this way it won't panic. And something similar to `waitRemoteLogger()` that waits for this goroutine to end before `execute()` finishes, to not leave it potentially dangling.

codebien: >something similar to waitRemoteLogger() that waits for this goroutine to end before execute() finishes, to not leave it potentially dangling.

@na-- I refactored a bit as you asked. Check the latest version, please.

>First, if something somehow logged to the stdlib logger after this writer was closed, I imagine this will panic, which is not ideal. It should not happen, but if it happens, it's better to not panic.

The `io.Write` API defines error as returned type, so I expect it will not generate any panic in that case, and log is a stdlib package, so I expect some consistency respecting the design.

>Off the top of my head, better approach would be a io.Writer that either discards incoming logs or falls back to the fallbackLogger once globalState.ctx is done

Does it make sense to have this dedicated to this part of the code? Why does not add a global component that automatically handles it?

codebien:
Rebased after the e2e tests refactor

cmd/root.go
```diff
@@ -67,32 +67,27 @@ func newRootCommand(gs *state.GlobalState) *rootCommand {
 
 func (c *rootCommand) persistentPreRunE(cmd *cobra.Command, args []string) error {
 	var err error
-
-	c.loggerStopped, err = c.setupLoggers()
+	c.loggerStopped, err = c.setupLoggers(c.globalState.Ctx)
```
codebien, line 70: I think is relevant to mention that moving `setupLoggers` directly to the execute method could simplify the current synchronization process, but it would impact how it is integrated with Cobra. Something good for the next PR.

cmd/root.go
```diff
@@ -218,33 +204,32 @@ func (c *rootCommand) setupLoggers() (<-chan struct{}, error) {
...
-
+		c.globalState.Logger.SetOutput(io.Discard)
 	case strings.HasPrefix(line, "loki"):
+		c.loggerIsRemote = true
 		ch = make(chan struct{}) // TODO: refactor, get it from the constructor
-		hook, err := log.LokiFromConfigLine(c.globalState.Ctx, c.globalState.FallbackLogger, line, ch)
+		hook, err := log.LokiFromConfigLine(ctx, c.globalState.FallbackLogger, line, ch)
 		if err != nil {
-			return nil, err
+			close(ch)
```
imiric, line 213: Can we do the refactor suggested by the TODO in this PR?

We currently [close the channel from the receiver side](https://github.com/grafana/k6/blob/0b563d4428391f6170d65bd30704b48b03a3f80f/log/loki.go#L165), which violates the [Channel Closing Principle](https://go101.org/article/channel-closing.html), and is generally a code smell.

With the additional closes here, we're relying on the channel not being closed already by the receiver to avoid the panic, which doesn't seem like it happens now, but it's best if we can avoid the uncertainty altogether.

imiric (commented):
It mostly LGTM, just a few non-blocking comments and questions.

@codebien How did you see the goleak report? I couldn't see anything from `go test -v -race -count=1 ./cmd/tests` on `master`, and I can't figure it out from their README :confused: I would like to see leaks detected on `master` and then fixed on this branch.

codebien:
Hi @imiric, thanks for your review.

On master, the leak is not reported because we are skipping the check. You should see the report removing this option. https://github.com/grafana/k6/blob/0b563d4428391f6170d65bd30704b48b03a3f80f/cmd/tests/tests.go#L55-L57

js/modules/k6/ws/ws_test.go
```diff
@@ -732,9 +732,6 @@ func TestErrors(t *testing.T) {
 		if (!closed) {
 			throw new Error ("conn no closed");
 		}
-		if (!onerror) {
-			throw new Error ("error event has not triggered");
-		}
```
codebien: This following test doesn't seem to do what the name says. I don't know how we could refactor it to simulate this case. All the cases I can think are not directly generated from the `socket.close` method. If we don't have better suggestions then I think we should drop it.

https://github.com/grafana/k6/blob/45dde486be59f751a5888f96cc5707ab153ec7cc/js/modules/k6/ws/ws_test.go#L694-L703

codebien:
Hey @imiric,
I tried to apply some fixes for your request changes, the current code is not so clean as I would like but I think to do something better we need to start creating a dedicated component for logging then it should be moved to a dedicated PR.

Regarding Httpmultibin I've moved the `ws-close` as a dedicated handler with the expectation that the server closes the connection. Instead, the echo handler waits for the client to start the closing ops.
Some `ws` tests weren't aligned with this strict principle so I had to adjust how we are using them.

.github/workflows/test.yml
```diff
@@ -77,7 +77,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        go-version: [1.19.x]
+        go-version: [1.20.x]
```
mstoykov, line 80: This looks like something went wrong on a rebase

codebien: Nop, it is expected. I tried to apply it to see if we would have any issues with 1.20, but I will remove soon.

cmd/root.go
```diff
@@ -260,5 +250,41 @@ func (c *rootCommand) setupLoggers() (<-chan struct{}, error) {
...
+		_ = w.Close()
+		c.loggersWg.Done()
+	}()
+	return nil
+}
+
+type asyncHook interface {
+	logrus.Hook
+	Listen(ctx context.Context)
+}
```
imiric, line 280: Could you move this to the `log` package, and export it there?

That way you don't have to export `LokiHook`, and it would be less changes overall, I think.

Plus, this is really a logging interface, so it belongs more in `log` than in `cmd`.

imiric (changes requested):
Almost LGTM :smile:

I'd only like to move `asyncHook` to the `log` package.

There are many subtle changes to the logging mechanism here, and while it visually seems safe, I hope we're not breaking some obscure behavior. The testing for this isn't great, so hopefully we can catch any issues in manual test runs. :crossed_fingers:

cmd/stdlog_integration_test.go
```diff
@@ -0,0 +1,45 @@
...
+	"bytes"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"go.k6.io/k6/cmd/tests"
+	"go.k6.io/k6/lib/testutils/httpmultibin"
+)
+
+// SetOutput sets the global log so it is racy with other tests
```
na--, line 13: Hmm maybe move it to a separate package then? After all, it currently might not be executed in parallel with the other tests, but it will still affect their `log` defaults :thinking: 

Though, now that I think about it, won't all e2e/integration tests have the same problem, since we always do that? :confused:

na--:
Generally LGTM, with a few nitpicks I commented on or :+1: already existing comments

cmd/stdlog_integration_test.go
codebien: >but it will still affect their log defaults

Sorry, I'm not getting it. What do you mean?

>Though, now that I think about it, won't all e2e/integration tests have the same problem, since we always do that? 

We have everything mocked in the GlobalState and we create it once per test, so we shouldn't have that kind of problem. Right?

na--: Yes, but when every test runs with that mocked `GlobalTestState`, it will still execute that line: https://github.com/grafana/k6/blob/367163be338f0d23b679585996aa2b13e3ec9cae/cmd/root.go#L81

Which will call stdlib's [`log.SetOutput`](https://pkg.go.dev/log#SetOutput) and potentially affect other tests... :disappointed: Not a lot we can do about it besides what you have already done here in making the test not parallel :+1: [It is thankfully synchronized](https://cs.opensource.google/go/go/+/refs/tags/go1.20.2:src/log/log.go;drc=99f811ecca5674a6e6f423b97ea79f0a6eedf4b6;l=79), so at least that shouldn't be a big problem :thinking: 

It is just annoying that it is some global state we need to change for k6 to work correctly, and that we probably can't mock in any way :disappointed: Anyway, I will resolve this comment, your PR is handling this fine and other tests shouldn't be too affected by this :crossed_fingers:

imiric (approved):

na-- (approved):
:crossed_fingers:

PR #2838, opened 2023-01-04
Add experimental tracing js module

oleiade:
Hi folks 👋🏻 

## What
This PR adds a `k6/experimental/tracing` module to k6. It implements the public API specified and agreed upon during our internal design process. As we talked about internally, it is embedded directly in the repository rather than imported from an xk6 extension repository.

Using this module, users can transparently add tracing (trace context) headers to their HTTP calls and have k6 emit the used traced ids as part of the output's metadata. Calling the `instrumentHTTP` function exposed by the module will automatically wrap the imported `http` module's function to include tracing information without involving any changes from the user to their scripts. Each request will use a different trace id, and have a random span id.

A lot of the code in this PR is, in fact, an adaptation of what already existed in [xk6-distributed-tracing](https://github.com/grafana/xk6-distributed-tracing); so kudos @Blinkuu and @dgzlopes for that 🙇🏻 

Design note: this PR implements the logic in Go rather than executing JS directly. The main reason for that is convenience regarding debugging and the ability to set and delete metadata, which I believe is not yet exposed to our JS runtime.

## Scope

In the spirit of keeping PRs small, this one only implements support for:
- propagators: w3c, b3, and jaeger
- `instrumentHTTP` wraps the `delete`, `get`, `head`, `options`, `post`, `patch`, `put`, `request` functions.

Support for the `batch` functions shall be added in later iterations. As well as support for sampling control and the baggage headers specification.

## Support required ✋🏻 

In an ideal world, I'd like to add some tests for the `instrumentHTTP` function. I want to start a test HTTP server and assert that when running a test script using the `instrumentHTTP` function, the expected headers are received by the server. I'd also like to do the same for the output's metadata.

The main blocking challenge I've encountered has been to make the init context available in the `modulestest.Runtime`, so that the `require` function is available in the context of the test script. I think my attempts so far could have been more fruitful. If you have ideas or guidance to help me achieve that, I'd be grateful 🙇🏻 

## Demo

```javascript
// we import the HTTP module in a standard manner
import http from "k6/http";
import { check } from "k6";

// This is the only change users need to make to include tracing context to their requests
// instrumentHTTP will ensure that all requests made by the http module
// will be traced. The first argument is a configuration object that
// can be used to configure the tracer.
//
// Currently supported HTTP methods are: get, post, put, patch, head, del,
// and options.
tracing.instrumentHTTP({
	propagator: "w3c",
});

export default () => {
	const params = {
		headers: {
			"X-My-Header": "something",
		},
	};

        // this http.get call will automatically include a traceparent header
        // and the used trace id will be included in the related data points in
        // the output's metadata.
	let res = http.get("http://httpbin.org/get", params);
	check(res, {
	});

	let data = { name: "Bert" };

         // this http.get call will automatically include a traceparent header
         // and the used trace id will be included in the related data points in
        // the output's metadata.
	res = http.post("http://httpbin.org/post", JSON.stringify(data), params);
	check(res, {
	});
};

```

Produces the the following HTTP requests:
```
INFO[0000] Request:
GET /get HTTP/1.1
Host: httpbin.org
User-Agent: k6/0.42.0 (https://k6.io/)
Traceparent: 00-dc0718b190d7e2d730a6219538e16e47-28cf4f6cd985afbb-01
X-My-Header: something
Accept-Encoding: gzip

INFO[0003] Request:
POST /post HTTP/1.1
Host: httpbin.org
User-Agent: k6/0.42.0 (https://k6.io/)
Content-Length: 15
Traceparent: 00-dc0718b1a5d7e2d7300e957d6ec9fb8b-ff9f9edcbfa41b91-01
X-My-Header: something
Accept-Encoding: gzip
```

## Edit
- 5th of January 2023: added support for `http.request`, and added it to the description

### DIFF
--- a/js/modules/k6/experimental/tracing/trace_id.go
+++ b/js/modules/k6/experimental/tracing/trace_id.go
@@ -1,0 +1,79 @@
+package tracing
+
+import (
+	"crypto/rand"
+	"encoding/binary"
+	"encoding/hex"
+	"fmt"
+)
+
+const (
+	// Being 075 the ASCII code for 'K' :)
+	k6Prefix = 0o756
+
+	// To ingest and process the related spans in k6 Cloud.
+	k6CloudCode = 12
+
+	// To not ingest and process the related spans, b/c they are part of a non-cloud run.
+	k6LocalCode = 33
+
+	// metadataTraceIDKeyName is the key name of the traceID in the output metadata.
+	metadataTraceIDKeyName = "trace_id"
+)
+
+// TraceID represents a trace-id as defined by the [W3c specification], and
+// used by w3c, b3 and jaeger propagators. See [Considerations for trace-id field generation]
+// for more information.
+//
+// [W3c specification]: https://www.w3.org/TR/trace-context/#trace-id
+// [Considerations for trace-id field generation]: https://www.w3.org/TR/trace-context/#considerations-for-trace-id-field-generation
+//
+//nolint:lll
+type TraceID struct {
+	Prefix            int16
+	Code              int8
+	UnixTimestampNano uint64
+}
+
+// NewTraceID returns a new TraceID with the given prefix, code and unix timestamp in nanoseconds.
+func NewTraceID(prefix int16, code int8, unixTimestampNano uint64) TraceID {
+	return TraceID{
+		Prefix:            prefix,
+		Code:              code,
+		UnixTimestampNano: unixTimestampNano,
+	}
+}
+
+// Encode encodes the TraceID into a hex string and a byte slice.
+func (t TraceID) Encode() (string, []byte, error) {
+	if !t.isValid() {
+		return "", nil, fmt.Errorf("failed to encode traceID: %v", t)
+	}
+
+	buf := make([]byte, 16)
+
+	n := binary.PutVarint(buf, int64(t.Prefix))
+	n += binary.PutVarint(buf[n:], int64(t.Code))
+	n += binary.PutUvarint(buf[n:], t.UnixTimestampNano)
+
+	randomness := make([]byte, 16-n)
+	err := binary.Read(rand.Reader, binary.BigEndian, randomness)
+	if err != nil {
+		return "", nil, fmt.Errorf("failed to read random bytes from os; reason: %w", err)
+	}
+
+	buf = append(buf[:n], randomness...)
+	hx := hex.EncodeToString(buf)
+
+	return hx, buf, nil
+}
+
+func (t TraceID) isValid() bool {
+	var (
+		isk6Prefix = t.Prefix == k6Prefix
+		isk6Cloud  = t.Code == k6CloudCode
+		isk6Local  = t.Code == k6LocalCode
+	)
+
+	return isk6Prefix && (isk6Cloud || isk6Local)
+}
--- a/js/modules/k6/experimental/tracing/tracing.go
+++ b/js/modules/k6/experimental/tracing/tracing.go
@@ -1,0 +1,35 @@
+package tracing
+
+import (
+	"errors"
+	"fmt"
+
+	"github.com/dop251/goja"
+	"go.k6.io/k6/js/common"
+	"go.k6.io/k6/js/modules"
+)
+
+// Tracing is the JS module instance that will be created for each VU.
+type Tracing struct {
+	vu modules.VU
+}
+
+// NewClient is the JS constructor for the tracing.Client
+//
+// It expects a single configuration object as argument, which
+// will be used to instantiate an `Object` instance internally,
+// and will be used by the client to configure itself.
+func (t *Tracing) NewClient(cc goja.ConstructorCall) *goja.Object {
+	rt := t.vu.Runtime()
+
+	if len(cc.Arguments) < 1 {
+		common.Throw(rt, errors.New("Client constructor expects a single configuration object as argument; none given"))
+	}
+
+	var opts options
+	if err := rt.ExportTo(cc.Arguments[0], &opts); err != nil {
+		common.Throw(rt, fmt.Errorf("unable to parse options object; reason: %w", err))
+	}
+
+	return rt.ToValue(NewClient(t.vu, opts)).ToObject(rt)
+}
--- a/js/modules/k6/experimental/tracing/tracing.go
+++ b/js/modules/k6/experimental/tracing/tracing.go
@@ -12,6 +12,8 @@ import (
 // Tracing is the JS module instance that will be created for each VU.
 type Tracing struct {
 	vu modules.VU
+
+	defaultClient *Client
 }
 
 // NewClient is the JS constructor for the tracing.Client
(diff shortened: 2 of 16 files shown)


oleiade:
The reviewer selection mechanism added you @na-- as a reviewer; feel free to opt-out if you do not wish/have no time to review it 😸 🦖 🙇🏻

js/modules/k6/experimental/tracing/trace_id.go
```diff
@@ -0,0 +1,68 @@
...
+
+// Encode encodes the TraceID into a hex string and a byte slice.
+func (t *TraceID) Encode() (string, []byte, error) {
+	var (
+		isk6Prefix    = t.Prefix == k6Prefix
+		isk6CloudCode = t.Code == k6CloudCode
+		isk6LocalCode = t.Code == k6LocalCode
+	)
+
+	if !isk6Prefix && (!isk6CloudCode || !isk6LocalCode) {
```
Blinkuu, line 48: I believe this logic is invalid according to [De Morgan's Law](https://www.whitman.edu/mathematics/higher_math_online/section01.03.html). 

The original logic was as follows:
```
if !(t.Prefix == K6Prefix && (t.Code == K6Code_Cloud || t.Code == K6Code_Local)) {
    return "", nil, fmt.Errorf("failed to encode traceID: %v", t)
}
```

To simplify:
```
!(P && (Q || V)) -> !P || (!Q && !V)
```

In the end, this should be:
```
if !isk6Prefix || (!isk6CloudCode && !isk6LocalCode) {
  return "", nil, fmt.Errorf("failed to encode traceID: %v", t)
}
```

Anyhow, the current `if` statement is a bit hard to read IMO. I'd prefer to have it the same way we had it defined before, namely:
```
// We define IsValid method on TraceID struct
func (t *TraceID) IsValid() bool {
	return t.Prefix == K6Prefix && (t.Code == K6Code_Cloud || t.Code == K6Code_Local)
}

// The if logic is more readable then.
if !t.IsValid() {
	return "", nil, fmt.Errorf("failed to encode traceID: %v", t)
}
```

PS. Someone, please double-check this. It's been a few years since I've done boolean algebra.

js/modules/k6/experimental/tracing/trace_id.go
```diff
@@ -0,0 +1,68 @@
...
+	buf := make([]byte, 16)
+
+	n := binary.PutVarint(buf, int64(t.Prefix))
+	n += binary.PutVarint(buf[n:], int64(t.Code))
+	n += binary.PutUvarint(buf[n:], t.UnixTimestampNano)
+
+	randomness := make([]byte, 16-len(buf[:n]))
+	err := binary.Read(rand.Reader, binary.BigEndian, randomness)
+	if err != nil {
+		return "", nil, err
```
Blinkuu, line 61: nit: wrap the returned error to make searching easier, like so

```
if err != nil {
		return "", nil, fmt.Errorf("failed to read: %w", err)
}
```

js/modules/k6/experimental/tracing/trace_id.go
```diff
@@ -0,0 +1,68 @@
...
+	k6CloudCode = 12
+
+	// To not ingest and process the related spans, b/c they are part of a non-cloud run.
+	k6LocalCode = 33
+
+	// metadataTraceIDKeyName is the key name of the traceID in the output metadata.
+	metadataTraceIDKeyName = "trace_id"
+)
+
+// TraceID of 16 bytes (128 bits) are supported by w3c, b3 and jaeger
```
Blinkuu, line 24: This struct is 88 bits. I know we are encoding it as into a buffer of size 16 bytes, but this comment can be confusing.

js/modules/k6/experimental/tracing/trace_id.go
```diff
@@ -0,0 +1,68 @@
...
+		return "", nil, fmt.Errorf("failed to encode traceID: %v", t)
+	}
+
+	buf := make([]byte, 16)
+
+	n := binary.PutVarint(buf, int64(t.Prefix))
+	n += binary.PutVarint(buf[n:], int64(t.Code))
+	n += binary.PutUvarint(buf[n:], t.UnixTimestampNano)
+
+	randomness := make([]byte, 16-len(buf[:n]))
```
Blinkuu, line 58: `16-len(buf[:n])` -> `16 - n`

js/modules/k6/experimental/tracing/trace_id.go
```diff
@@ -0,0 +1,68 @@
...
+
+// TraceID of 16 bytes (128 bits) are supported by w3c, b3 and jaeger
+type TraceID struct {
+	Prefix            int16
+	Code              int8
+	UnixTimestampNano uint64
+}
+
+// NewTraceID returns a new TraceID with the given prefix, code and unix timestamp in nanoseconds.
+func NewTraceID(prefix int16, code int8, unixTimestampNano uint64) *TraceID {
```
Blinkuu, line 32: Why do we return a pointer? No need to allocate this data on the heap.

js/modules/k6/experimental/tracing/tracing.go
```diff
@@ -0,0 +1,286 @@
...
+			common.Throw(t.vu.Runtime(), err)
+		}
+	}
+
+	// Inject the updated HTTP module object in the runtime,
+	// overriding any previously imported one in the process.
+	err = t.vu.Runtime().Set("http", httpModuleObj)
+	if err != nil {
+		common.Throw(t.vu.Runtime(), err)
+	}
```
Blinkuu, line 71: I have 0 knowledge of how k6 actually works internally, but anytime I see something like this, a 💡 appears in my head. Can this be overridden by another k6 module that overrides HTTP? I understand that this module does not wrap the HTTP module serving as a proxy but rather replaces it with a modified HTTP client. If another module does the same thing, tracing will suddenly stop working, correct?

oleiade:
I figured out that I could add the `http.request` method support almost for free, and added it to this PR 👏🏻

js/modules/k6/experimental/tracing/tracing.go
```diff
@@ -0,0 +1,286 @@
...
+		if err != nil {
+			common.Throw(rt, fmt.Errorf("failed to normalize the params argument; reason: %w", err))
+		}
+
+		// Ensure that the params object contains a headers object.
+		// Create it if it doesn't.
+		headers, err := t.getOrCreateHeaders(params)
+		if err != nil {
+			common.Throw(rt, fmt.Errorf("failed to normalize params argument headers property; reason: %w", err))
+		}
```
Blinkuu, line 152: This `headers` value is first used in line `168`. Perhaps it makes sense to move this call a bit closer to that line. You can read more about this in Clean Code, Chapter 5 for example.

Blinkuu (changes requested):

js/modules/k6/experimental/tracing/tracing.go
```diff
@@ -0,0 +1,278 @@
...
+
+		traceID := NewTraceID(k6Prefix, k6CloudCode, uint64(time.Now().UnixNano())/uint64(time.Millisecond))
```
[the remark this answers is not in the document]
oleiade: You're right 🤔 

This effectively operates a conversion from nanoseconds to milliseconds. Was there an explicit reason for that @dgzlopes (this this statement is taken from xk6-distributed-tracing)? Maybe some precision requirements or considerations? Also, I've reused the `UnixTimestampNano` attribute definition, is there a reason for it to be a `uint64`, rather than a `int64`, like the result of `UnixNano()`?

Cheers 🙇🏻

js/modules/k6/experimental/tracing/trace_id.go
oleiade: Indeed 👍🏻

js/modules/k6/experimental/tracing/tracing.go
oleiade: I see your concern, but I don't believe that would be the case (there might be some corner case or JS devilry escaping me, though).

The tracing module imports the `http` module, and wraps its method to insert its logic before and after the execution of the underlying `http.{method}`. The call to the HTTP method itself is left untouched; all we do in the case of the tracing module is essentially modify the params argument the user gave to the call to the HTTP method so that it includes the tracing context, set metadata, calls the HTTP method with the updated args, delete the metadata, and return the result of the method. Thus, even if another method used the same technique, it would still work as long as they don't modify the function's prototype.

js/modules/k6/experimental/tracing/tracing.go
oleiade: In other scenarios, I would agree, and my C++ years tend to lead me to prefer code locality indeed. However, in this example, the choice I made was to prefer failing early. The trace id generation issues syscalls (reading from the os random source) to produce randomness which has a cost, thus I wanted to make sure that I would produce a trace id only if I was sure I'd have somewhere to put it.

js/modules/k6/experimental/tracing/tracing.go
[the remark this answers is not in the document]
dgzlopes: I'll be honest: I don't remember :smile: 

So yeah, probably no explicit reason.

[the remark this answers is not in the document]
oleiade: Hehe, alrighty, I'll just store nanoseconds, and as `int64` then 👍🏻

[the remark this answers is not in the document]
Blinkuu: @oleiade Please keep UnixTimestampNano as `uint64`. Otherwise, this will not work, as we will not be able to decode it on the Crocospans side.

js/modules/k6/experimental/tracing/tracing.go
Blinkuu: Got it 👍

js/modules/k6/experimental/tracing/tracing.go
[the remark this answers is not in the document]
oleiade: @Blinkuu, that's good to know, thanks for the info. I will keep it as a `uint64` then 👍🏻 🙇🏻

oleiade:
✋🏻 Switching the PR to draft, as I'm currently waiting on integration testing to be unblocked, and as I realized that I might as well "just" put together a `Client` in this PR anyway while waiting it ✋🏻

oleiade:
Closing this in favor of #2853 🚋 #2854 🚆 #2855 🚄 🚀

Issue #2840, opened 2023-01-09
Integration tests for CLI and other `cmd/` edge cases


na--:
We noticed a few subtle issues that the refactoring in https://github.com/grafana/k6/pull/2787 introduced and which were apparently uncovered by any existing integration tests :disappointed: So we might need to temporarily revert that PR and add some tests for them before we can safely refactor that part of the codebase again. I'll describe the issues we found below:

First, it's possible for a user to redirect only `stderr` (e.g. to a file or `/dev/null`), but leave `stdout` alone (or vice-versa), so it's their fully-featured terminal. Previously, if someone ran `k6 run github.com/grafana/k6/samples/http_get.js 2>/some/file`, they would still see the nice dynamic progress-bar, only their logs will be sent to a non-TTY file. Whereas now, because of [this line specifically](https://github.com/grafana/k6/blob/613fe8b56cf504ed85516b6a2584f8d9870f2725/ui/console/console.go#L43), they would get the somewhat nasty "print static progress every second" behavior we have in CI, even though their terminal is perfectly capable of showing the dynamic progress bars. Not to mention that [the parameters we pass to `handleSummary()`](https://github.com/grafana/k6/blob/613fe8b56cf504ed85516b6a2584f8d9870f2725/cmd/run.go#L221-L222) might now be wrong with such partial stream redirection.

Another thing is the `fallbackLogger`, which was wrongly removed/masked, so [this](https://github.com/grafana/k6/blob/613fe8b56cf504ed85516b6a2584f8d9870f2725/cmd/root.go#L263) and [this](https://github.com/grafana/k6/blob/613fe8b56cf504ed85516b6a2584f8d9870f2725/cmd/root.go#L284) lines are now problematic. When the console logger is replaced with something remote that has a buffer and can fail (like `--log-output=loki`), we need the fallback logger for logging and displaying errors if that thing failed, or if we need to log something after we have flushed the bufffer and closed the connection (i.e. right at the end of the test).

These things were apparently untested, so we didn't notice the problems. Moreover, another subtle change that masked other potential problems was the removal of manual locking capabilities in the integration tests. We intentionally didn't lock the mock-`stdout` and mock-`stderr` buffers, since we wanted to see data races in the tests when they were unexpectedly accessed. See this discussion for more details: https://github.com/grafana/k6/pull/2787#discussion_r1054364822

So, yeah, these things are tricky to test well, but with the recent integration test improvements, it should be possible to have at least a few tests for them.

PR #2841, opened 2023-01-09
Revert `cmd` UI refactor (#2787)

imiric:
We decided to revert #2787 due to some issues discovered recently (see #2840). If we decide to bring it back, it will need to take into account those issues.

This reverts it partially, only keeping 074b53214 since it's irrelevant, and adds some changes to appease the linter. See the commits for details.


na-- (approved):

na--:
Should we merge this with squashing? :thinking:

PR #2846, opened 2023-01-11
js: Refactor initializing a bundle

mstoykov:
DRYing the initialization of a bundle and adding a test for previously not tested archiving of an archive

### DIFF
--- a/js/bundle.go
+++ b/js/bundle.go
@@ -64,6 +64,13 @@ func (bi *BundleInstance) getExported(name string) goja.Value {
 func NewBundle(
 	piState *lib.TestPreInitState, src *loader.SourceData, filesystems map[string]afero.Fs,
 ) (*Bundle, error) {
+	return newBundle(piState, src, filesystems, lib.Options{}, true)
+}
+
+func newBundle(
+	piState *lib.TestPreInitState, src *loader.SourceData, filesystems map[string]afero.Fs,
+	options lib.Options, updateOptions bool, // TODO: try to figure out a way to not need both
+) (*Bundle, error) {
 	compatMode, err := lib.ValidateCompatibilityMode(piState.RuntimeOptions.CompatibilityMode.String)
 	if err != nil {
 		return nil, err
@@ -89,6 +96,7 @@ func NewBundle(
 		Program:           pgm,
 		BaseInitContext:   NewInitContext(piState.Logger, rt, c, compatMode, filesystems, loader.Dir(src.URL)),
 		RuntimeOptions:    piState.RuntimeOptions,
+		Options:           options,
 		CompatibilityMode: compatMode,
 		exports:           make(map[string]goja.Callable),
 		registry:          piState.Registry,
@@ -97,7 +105,7 @@ func NewBundle(
 		return nil, err
 	}
 
-	err = bundle.getExports(piState.Logger, rt, true)
+	err = bundle.getExports(piState.Logger, rt, updateOptions)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


mstoykov:
This is an extract from a much larger js refactor.

It just happens to be mostly self contained and a lot less ... controversial then the remaining changes so I decided to pull it out.

The test is because I *did* break this originally, but currently can't remember how or what I did wrong. But still a good idea to have a small test for this.

js/bundle.go
```diff
@@ -63,6 +63,13 @@ func (bi *BundleInstance) getExported(name string) goja.Value {
 // NewBundle creates a new bundle from a source file and a filesystem.
 func NewBundle(
 	piState *lib.TestPreInitState, src *loader.SourceData, filesystems map[string]afero.Fs,
+) (*Bundle, error) {
+	return newBundle(piState, src, filesystems, lib.Options{}, true)
+}
+
+func newBundle(
+	piState *lib.TestPreInitState, src *loader.SourceData, filesystems map[string]afero.Fs,
+	options lib.Options, updateOptions bool, // TODO: try to figure out a way to not need both
```
codebien, line 72: Out of curiosity, does the larger refactor help here?

codebien (approved):
LGTM

js/bundle.go
mstoykov: Not really, but IMO this is on the "small problems" list of this codebase :sweat_smile:

mstoykov: Arguably I can drop the boolean if I extract 

```golang
    err = bundle.getExports(piState.Logger, rt, updateOptions)
	if err != nil {
		return nil, err
	}
```

in each caller - but I am less happy with this idea then with providing the options and a boolean. And unfortunately we *need* the options for other reasons so not providing them is not possible :( .

js/runner_test.go
```diff
@@ -2745,3 +2745,35 @@ exports.default = () => {
 		})
 	}
 }
+
+func TestArchivingAnArchiveWorks(t *testing.T) {
```
imiric, line 2749: This test passes on `master` as well.

Doesn't it need to assert that the options _weren't_ updated after the second `NewFromArchive()` call?

imiric (commented):
LGTM, just a test comment.

But I would like to see the reason explaining _why_ this is done; both in the commit message and the PR description. GH will populate the PR description with the commit message if there's only one commit, so you only need to type it once.

If this is avoiding a second archive from updating options, then it sounds more like a fix than a refactor. But where is this an issue? All of this should be explained. :)

js/runner_test.go
mstoykov: It is passing on master as it isn't fixing anything it is just a test that I broke on the original writing of this.  If this commit was fixing somethign it wouldn't be a refactor ;) 

And the point was that I broke the whole *making* of the archive from the archive. Which is also what the names suggests

mstoykov:
> why this is done 

it is done as the logic for how the bundle was initilizied was basically repeated in two places and it is now in a single one. I don't think the commit message can be any more consice without going over the 60 character limit and I really don't think that for a refactor of this kind I need to write a small novel on why we are doing it.

js/runner_test.go
imiric: OK, makes sense.

imiric (approved):
> it is done as the logic for how the bundle was initilizied was basically repeated in two places and it is now in a single one.

I actually didn't notice that, so a more detailed commit message would've helped.

> I don't think the commit message can be any more consice without going over the 60 character limit and I really don't think that for a refactor of this kind I need to write a small novel on why we are doing it.

It's not about writing a small novel. Just mentioning that we're DRYing things and adding a test for a previously broken feature would answer the "why". All commit messages should answer that; a "refactor" doesn't say anything.

Issue #425, opened 2017-12-18
URL parsing of --out influxdb= not optimal


StianOvrevage:
First off, thanks for a great piece of awesome-sauce. So far I love working with k6!

Now to what is causing me some headache. I have several InfluxDBs hosted behind a reverse proxy that uses paths to send traffic to the correct InfluxDB instance. However, the behaviour of `--out influxdb` only allows me to specify the hostname, and not the sub-path of the InfluxDB instance. k6 automatically interprets the path-portion of the URL to be the InfluxDB database. So using `https://my.site.io/influxdbA/k6db` results in k6 trying to connect to InfluxDB hosted on `my.site.io` and using database `influxdbA/k6db`. Instead of my intended behaviour of connecting to `https://my.site.io/influxdbA/` and use database `k6db`.

Here is a few suggestions:
  - Only the last portion of the URL should be interpreted as the database name. Will be a problem if people are using database names with `/`, even though I'm not sure if that is allowed in InfluxDB.
  - Specify the database as an argument, like `site.io/influxA?db=k6db`.
  - Specify the database as a separate argument from the InfluxDB URL itself.

ragnarlonn:
It seems Influxdb is very permissive with what characters it allows in names (of databases and other things) but that especially in the case of db names this is considered more or less a bug, so not unlikely it will be fixed in the future. I'd say we go with your first suggestion and interpret only the part after the last slash as the database name.

liclac:
I agree, I'd appreciate a PR. If someone turns up and says they're using DB names with slashes, we can add an alternate means to specify it for those people.

ragnarlonn:
@StianOvrevage @robingustafsson @liclac I thought I'd take a stab at this one - it had to be really simple to fix, right?  I extracted the db name as the string following the last "/" in the URL path, then tried setting the address field wanted by the influxdb client to be everything before the db name, but no matter what I did, it seemed to connect to "http(s)://hostname:port/write".

Turns out the stupid influxdb client library ignores any URL path you use in the address string, it hardcodes the path to be "write" for a write operation. See https://github.com/influxdata/influxdb/blob/master/client/v2/client.go#L370

Argh. So this seems unsolvable unless we use another client library (if there is one), come up with another solution, or manage to get influxdata to fix their library (provided they think it *should* be fixed, that is).

StianOvrevage:
Damn. That is soo depressing. What is the point of using HTTP for transport if you're just gonna ignore even the basics of the protocol?

[...]

PR #2847, opened 2023-01-12
fix for https://github.com/grafana/k6/issues/425

YogeshSagarCB:


na-- (changes requested):
Unfortunately we can't merge the PR as it is. You have modified the code of the [external InfluxDB client library](https://github.com/influxdata/influxdb1-client/blob/master/v2/client.go) that we have [vendored](https://go.dev/ref/mod#vendoring) in our git repo. This issue is not something that would require such changes, which is basically equivalent to forking the library.

na--:
Closing due to lack of response, will reopen if PR author returns

PR #2851, opened 2023-01-13
Update goja

### LINKED ISSUE #2849: Inconsistent behavior of exec.test.abort within a group
### Brief summary

If I use `exec.test.abort` inside a `group` call, the k6 script exits immediately, that's OK.

But I got the exit code 0, while [the documentation says it should be 108](https://k6.io/docs/javascript-api/k6-execution/#test):

> abort([String])
> function
> It aborts the test run with the exit code 108, and an optional string parameter can provide an error message. Aborting the test will not prevent the teardown() execution.

When I move the `exec.test.abort` call outside of the group, the exit code is `108` (what I expect).

### k6 version

0.42.0

### OS

Ubuntu 20.04

docker.io/grafana/k6:0.42.0

Having 2 files:

## abort-main.js 

```javascript

export default function () {
    exec.test.abort("failed message from main");
    console.log("will not be visible");
}
```

## abort-group.js

```javascript
import { group } from "k6";

export default function () {
    group("some group", () => {
        exec.test.abort("failed message from group");
        console.log("will not be visible");
    });

    console.log("will not be visible either");
}
```

And then, run k6 with both file, using `docker.io/grafana/k6:0.42.0` Docker image, and finally get the exit code (`$?`), like this:

`docker run --rm -v $(pwd)/:/usr/local/me docker.io/grafana/k6:0.42.0 run /usr/local/me/my-file.js ; echo "exit code is:" $?`

### Expected behaviour

I expect following results:

1. both test failed
2. the exit code is 108
3. the log messages `will not be visible` and `will not be visible either` are not present
4. the error messages `failed message from main` and `failed message from group` are visible

## abort-main.js

```
$ docker run --rm -v $(pwd)/:/usr/local/me docker.io/grafana/k6:0.42.0 run /usr/local/me/abort-main.js ; echo "exit code is:" $?

          /\      |‾‾| /‾‾/   /‾‾/  
     /\  /  \     |  |/  /   /  /   

  execution: local
     script: /usr/local/me/abort-main.js
     output: -

           * default: 1 iterations for each of 1 VUs (maxDuration: 10m0s, gracefulStop: 30s)


running (00m00.0s), 0/1 VUs, 0 complete and 1 interrupted iterations
default ✗ [ 100% ] 1 VUs  00m00.0s/10m0s  1/1 iters, 1 per VU

     data_received........: 0 B 0 B/s
     data_sent............: 0 B 0 B/s
     iteration_duration...: avg=36.8µs min=36.8µs med=36.8µs max=36.8µs p(90)=36.8µs p(95)=36.8µs
     iterations...........: 1   10183.506792/s

exit code is: 108
```

### results

1. ✅ test failed (it's OK because we get `default ✗` message)
2. ✅ the exit code is 108
3. ✅ the log messages `will not be visible` and `will not be visible either` are not present
4. ❌ the error message `failed message from main` is visible (text is missing in the output)

## abort-group.js

```
$ docker run --rm -v $(pwd)/:/usr/local/me docker.io/grafana/k6:0.42.0 run /usr/local/me/abort-group.js ; echo "exit code is:" $?

          /\      |‾‾| /‾‾/   /‾‾/  
     /\  /  \     |  |/  /   /  /    

  execution: local
     script: /usr/local/me/abort-group.js
     output: -

           * default: 1 iterations for each of 1 VUs (maxDuration: 10m0s, gracefulStop: 30s)

time="2023-01-12T16:50:42Z" level=error msg="GoError: test aborted: failed message from group at file:///usr/local/me/abort-group.js:6:24(7)\n\tat go.k6.io/k6/js/modules/k6.(*K6).Group-fm (native)\n\tat file:///usr/local/me/abort-group.js:5:4(5)\n\tat native\n" executor=per-vu-iterations scenario=default source=stacktrace

running (00m00.0s), 0/1 VUs, 1 complete and 0 interrupted iterations
default ✓ [ 100% ] 1 VUs  00m00.0s/10m0s  1/1 iters, 1 per VU

     █ some group

     data_received........: 0 B 0 B/s
     data_sent............: 0 B 0 B/s
     group_duration.......: avg=18.37µs min=18.37µs med=18.37µs max=18.37µs p(90)=18.37µs p(95)=18.37µs
     iteration_duration...: avg=72.52µs min=72.52µs med=72.52µs max=72.52µs p(90)=72.52µs p(95)=72.52µs
     iterations...........: 1   5404.032489/s

exit code is: 0
```

### results

1. ❌ test failed (it's KO because we get `default ✓` message)
2. ❌ the exit code is 0
3. ✅ the log messages `will not be visible` and `will not be visible either` are not present
4. ✅ the error message `failed message from group` is visible

mstoykov:
fixes #2849

### DIFF
--- a/cmd/integration_test.go
+++ b/cmd/integration_test.go
@@ -1004,6 +1004,30 @@ func TestAbortedByScriptAbortInVUCode(t *testing.T) {
 	})
 }
 
+func TestAbortedByScriptAbortInVUCodeInGroup(t *testing.T) {
+	t.Parallel()
+	script := `
+		import exec from 'k6/execution';
+        import { group } from 'k6';
+		export default function () {
+            group("here", () => {
+                exec.test.abort('foo');
+            });
+		};
+		export function handleSummary() { return {stdout: '\n\n\nbogus summary\n\n\n'};}
+	`
+
+	t.Run("noLinger", func(t *testing.T) {
+		t.Parallel()
+		testAbortedByScriptTestAbort(t, true, script, runTestWithNoLinger)
+	})
+
+	t.Run("withLinger", func(t *testing.T) {
+		t.Parallel()
+		testAbortedByScriptTestAbort(t, true, script, runTestWithLinger)
+	})
+}
+
 func TestAbortedByScriptAbortInSetup(t *testing.T) {
 	t.Parallel()
 	script := `
--- a/go.mod
+++ b/go.mod
@@ -8,7 +8,7 @@ require (
 	github.com/PuerkitoBio/goquery v1.8.0
 	github.com/Soontao/goHttpDigestClient v0.0.0-20170320082612-6d28bb1415c5
 	github.com/andybalholm/brotli v1.0.4
-	github.com/dop251/goja v0.0.0-20230110123741-af048f025cbb
+	github.com/dop251/goja v0.0.0-20230113101057-bf9ef13dd7b1
 	github.com/fatih/color v1.13.0
 	github.com/go-sourcemap/sourcemap v2.1.4-0.20211119122758-180fcef48034+incompatible
 	github.com/golang/protobuf v1.5.2
--- a/go.sum
+++ b/go.sum
@@ -85,8 +85,8 @@ github.com/dop251/goja v0.0.0-20211022113120-dc8c55024d06/go.mod h1:R9ET47fwRVRP
 github.com/dop251/goja v0.0.0-20220405120441-9037c2b61cbf/go.mod h1:R9ET47fwRVRPZnOGvHxxhuZcbrMCuiqOz3Rlrh4KSnk=
 github.com/dop251/goja v0.0.0-20221003171542-5ea1285e6c91/go.mod h1:yRkwfj0CBpOGre+TwBsqPV0IH0Pk73e4PXJOeNDboGs=
 github.com/dop251/goja v0.0.0-20221106173738-3b8a68ca89b4/go.mod h1:yRkwfj0CBpOGre+TwBsqPV0IH0Pk73e4PXJOeNDboGs=
-github.com/dop251/goja v0.0.0-20230110123741-af048f025cbb h1:35N2yi2cM6XjKILDhSDWJuCg7ANcjUxRZw0TXJH4kH0=
-github.com/dop251/goja v0.0.0-20230110123741-af048f025cbb/go.mod h1:yRkwfj0CBpOGre+TwBsqPV0IH0Pk73e4PXJOeNDboGs=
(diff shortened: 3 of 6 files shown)


na--:
Hmm, looking at the goja diff in `vendor/`, it doesn't actually include the fix for the issue. 

Also, a small k6 test on our side is probably a good idea, to make sure this doesn't repeat. We plan to potentially make changes to `group()` soon anyway, so more tests about it won't hurt :sweat_smile: 

Finally, this comment https://github.com/dop251/goja/pull/471#discussion_r1069182207 makes it seem like there might be other follow-up changes in goja in the near future, should we wait?

codebien (approved):
:bow:

na-- (changes requested):
https://github.com/grafana/k6/pull/2851#issuecomment-1381621707

mstoykov:
>  Hmm, looking at the goja diff in vendor/, it doesn't actually include the fix for the issue.

modcache tricked me :facepalm: . I did previous to that test it with a local replace - I have no bumped it. 

> Also, a small k6 test on our side is probably a good idea, to make sure this doesn't repeat. We plan to potentially make changes to group() soon anyway, so more tests about it won't hurt sweat_smile 

This IMO is blocked on the integration test refactor so I am going to wait on that. 

> Finally, this comment https://github.com/dop251/goja/pull/471#discussion_r1069182207 makes it seem like there might be other follow-up changes in goja in the near future, should we wait?

Given that this fixes this exact problem and the other might fix other problems that we haven't noticed I see no point in waiting - especially as we don't have a particular thing that we expect will be fixed.

codebien:
>Also, a small k6 test on our side is probably a good idea

>This IMO is blocked on the integration test refactor so I am going to wait on that.

What should be the case for the test? Getting the correct error when the test is aborted from a nested function? Is an integration test required, can we do it with a unit test?

>Given that this fixes this exact problem and the other might fix other problems that we haven't noticed I see no point in waiting - especially as we don't have a particular thing that we expect will be fixed.

I agree

mstoykov:
> Getting the correct error when the test is aborted from a nested function? 

That is arguably already tested in goja. I am much more interested in an integration test at this point that even if inside a group the error will go through to the end and will be handled correctly.

na--:
It is somewhat blocked by the integration tests refactor, though you can just edit this current integration test to just abort from a `group()`, or, better yet, copy the whole test and add `group()`: https://github.com/grafana/k6/blob/60b8b150c3de13ce76fb80173e384bcebf50a504/cmd/integration_test.go#L986-L1005

Since most of the logic is actually in `testAbortedByScriptTestAbort()`, it shouldn't affect anything

na-- (approved):
:heart:

codebien (approved):

PR #2853, opened 2023-01-13
Experimental tracing module (1/3): implement trace ids generation and propagation

oleiade:
## About
To keep Pull Requests as small as possible and reduce the review burden, this PR is the first in a chain of 3 implementing the decided API design for the experimental tracing module.

This first PR exposes the `k6/experimental/tracing` module to k6 scripts but does not include any public APIs. Instead, it implements the trace id generation and propagation logic according to the [W3C trace contexts](https://www.w3.org/TR/trace-context) specification.

The code introduced in this PR will be used in further PRs by the `Client` implementation to produce a trace id to attach to a given request and emit as part of the output's data points metadata, as well as producing the trace context headers in the supported formats (W3C, B3, and Jaeger). 

Much of this PR's code and logic is essentially inherited and refactored from the [xk6-distributed-tracing](https://github.com/grafana/xk6-distributed-tracing) project.

Looking forward to your review 🙇🏻 

## References

Next: #2854


js/modules/k6/experimental/tracing/module.go
```diff
@@ -0,0 +1,48 @@
...
+	return &RootModule{}
+}
+
+// NewModuleInstance implements the modules.Module interface and returns
+// a new instance for each VU.
+func (*RootModule) NewModuleInstance(vu modules.VU) modules.Instance {
+	// Any goja.Value exported to a Go struct will be mapped
+	// to the matching fields using the `js:""` tags.
+	vu.Runtime().SetFieldNameMapper(goja.TagFieldNameMapper("js", true))
+
```
mstoykov, line 38: Is this just from that extension? As touching any of hte `Set*` method on the runtime from an extension is likely a big no-no. And in this particular case it is a breaking other APIs as k6 has a particular implementation of this.

js/modules/k6/experimental/tracing/trace_id.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+// NewTraceID returns a new TraceID with the given prefix, code and unix timestamp in nanoseconds.
+func NewTraceID(prefix int16, code int8, unixTimestampNano uint64) TraceID {
+	return TraceID{
+		Prefix:            prefix,
+		Code:              code,
+		UnixTimestampNano: unixTimestampNano,
+	}
+}
```
codebien, line 42: Has this constructor a defined reason or is it just syntax? Especially without the guarantee to generate a valid trace id, it sounds wrong to me.

js/modules/k6/experimental/tracing/trace_id.go
```diff
@@ -0,0 +1,76 @@
...
+)
+
+// TraceID represents a trace-id as defined by the [W3c specification], and
+// used by w3c, b3 and jaeger propagators. See [Considerations for trace-id field generation]
+// for more information.
+//
+// [W3c specification]: https://www.w3.org/TR/trace-context/#trace-id
+// [Considerations for trace-id field generation]: https://www.w3.org/TR/trace-context/#considerations-for-trace-id-field-generation
+//
+//nolint:lll
```
codebien, line 28: >//nolint:lll

Is it required?

js/modules/k6/experimental/tracing/trace_id.go
```diff
@@ -0,0 +1,76 @@
...
+type TraceID struct {
+	Prefix            int16
+	Code              int8
+	UnixTimestampNano uint64
```
codebien, line 32: Comments for each field could be useful, or a direct link to the w3c doc.

js/modules/k6/experimental/tracing/trace_id.go
```diff
@@ -0,0 +1,76 @@
...
+// NewTraceID returns a new TraceID with the given prefix, code and unix timestamp in nanoseconds.
+func NewTraceID(prefix int16, code int8, unixTimestampNano uint64) TraceID {
+	return TraceID{
+		Prefix:            prefix,
+		Code:              code,
+		UnixTimestampNano: unixTimestampNano,
+	}
+}
+
+// Encode encodes the TraceID into a hex string and a byte slice.
```
codebien, line 44: Why does it return both? If it has a reason good to have in the comment.

js/modules/k6/experimental/tracing/trace_id.go
```diff
@@ -0,0 +1,76 @@
...
+func (t TraceID) Encode() (string, []byte, error) {
+	if !t.isValid() {
+		return "", nil, fmt.Errorf("failed to encode traceID: %v", t)
+	}
+
+	buf := make([]byte, 16)
+
+	n := binary.PutVarint(buf, int64(t.Prefix))
```
codebien, line 52: You could use https://pkg.go.dev/encoding/binary#ByteOrder

codebien (changes requested):
Hey,
thanks for it :bow: I left a few comments.

js/modules/k6/experimental/tracing/module.go
oleiade: Good catch indeed! 👀 🙇🏻  Gonna look into removing this!

js/modules/k6/experimental/tracing/trace_id.go
oleiade: You're right, I did it automatically, but it's not even useful 👍🏻 Removing it.

js/modules/k6/experimental/tracing/trace_id.go
oleiade: None indeed, I blindly reproduced the behavior from xk6-distributed-tracing, but it's not necessary indeed 👍🏻 Removing it ✂️

js/modules/k6/experimental/tracing/trace_id.go
oleiade: Was caused by the link definition being really long in the godoc, shortening it 👍🏻

js/modules/k6/experimental/tracing/trace_id.go
oleiade: I must admit I'm not too familiar with the binary package, so I don't necessarily see neither how to do the same with the `ByteOrder` interface nor how it would benefit the implementation over `PutVarint`/`PutUvarint`. Would you be able to illustrate it with an example? 😃

js/modules/k6/experimental/tracing/module.go
oleiade: Dropped it ✅

js/modules/k6/experimental/tracing/trace_id.go
codebien: There is an example in the linked page. Let me know if it is not enough.

oleiade: Hehe, I've seen 👀 I even experimented with it, but it was somewhat inconclusive, and I had a hard time reproducing the current behavior. Hence me asking for your input as it sounded like you might have something specific in mind. 👍🏻

codebien: Re-checking better this, I've just realized that Prefix is signed int16. Does it have a specific reason?

Anyway, this is not a blocker, it is mostly a minor optimization for the feature.

codebien (dismissed):
LGTM :+1:

codebien (commented):
I'm not sure we should merge to master directly here, should we merge to a long-lived intermediate branch for aggregating the entire feature?

mstoykov (dismissed):
LGTM!

I am glad we have smaller PRs.

In general it seems to me like 90% of this PR can be *not* under the js package. This likely will be useful in the future so that other js modules or parts of the system can use the tracing code. 

This though likely should be done in a separate PR(s) as I don't want to add more things to be discussed for this PR chain. 

Also I would've likely *not* introduced and registered the module in this PR but done it later. 

Given that it seems like we agree to merge the other 2 PRs in this one and then merge this one - this is really not a problem that needs to be fixed. But is something I would like to note.

oleiade:
Heads-up @codebien and @mstoykov I've addressed the comments made so far, and also replaced the trace id's `UnixTimestampNano` with a `time.Time` as discussed in the second PR #2854 🙇🏻

codebien (commented):
The current code LGTM

mstoykov (approved):

samples/experimental/tracing-client.js
```diff
@@ -0,0 +1,46 @@
+import http from "k6/http";
```
codebien, line 1: I think we should move this file in `samples/experimental/tracing/` folder

codebien (changes requested):

samples/experimental/tracing-client.js
oleiade: Absolutely, completely forgot to do it indeed 🤦🏻

codebien:
@oleiade can we get https://github.com/grafana/k6/pull/2855#discussion_r1087664171 before merging this PR?

oleiade:
@codebien considering it was brought up as a non-blocking request, I was planning to implement it in the PR implementing sampling (as I will have to modify the tests in the process anyways). If that's okay with you, I'd prefer to merge as is now 🙇🏻

codebien (approved):
LGTM

PR #2854, opened 2023-01-13
Experimental tracing module (2/3): define and expose a `tracing.Client`

oleiade:
## About

To keep Pull Requests as small as possible and reduce the review burden, this PR is the second in a chain of three implementing the decided API design for the experimental tracing module.

⚠️ ✋🏻 This PR is part of a PR chain, and is based on #2853  

This second PR builds upon #2853 and adds a `Client` type to `k6/experimental/tracing`. This type mimics the relevant part of the HTTP module to allow users to perform HTTP requests embedding a trace context while also attaching the produced trace id as output metadata on the HTTP-related samples. In the larger context of the module's goal, the `Client` should hold the underlying logic for the `instrumentHTTP` function that's to come in the third PR. Still, it also allows users to selectively trace requests and reconfigure their tracing configuration on the fly if they wish.

To achieve that, it is built upon the HTTP module's request function, which wraps with *hooking* logic to ensure the expected headers are produced and present and emit the output metadata. 

The `Client` is configured using an options object that lets the user select the propagation format they use. Other options, such as sampling rate, and baggage, have been added for forward compatibility and convenience but will lead to "not implemented" errors if used.

## Example of it in action

```javascript
import http from "k6/http";
import { check } from "k6";

// Explicitly instantiating a tracing client allows to distringuish
// instrumented from non-instrumented HTTP calls, by keeping APIs separate.
// It also allows for finer-grained configuration control, by letting
// users changing the tracing configuration on the fly during their
// script's execution.
let instrumentedHTTP = new tracing.Client({
	propagator: "w3c",
});

const testData = { name: "Bert" };

export default () => {
	// Using the tracing client instance, HTTP calls will have
	// their trace context headers set.
	let res = instrumentedHTTP.request("GET", "http://httpbin.org/get", null, {
		headers: {
			"X-Example-Header": "instrumented/request",
		},
	});
	check(res, {
	});

	// The tracing client offers more flexibility over
	// the `instrumentHTTP` function, as it leaves the
	// imported standard http module untouched. Thus,
	// one can still perform non-instrumented HTTP calls
	// using it.
	res = http.post("http://httpbin.org/post", JSON.stringify(testData), {
		headers: { "X-Example-Header": "noninstrumented/post" },
	});
	check(res, {
	});

	res = instrumentedHTTP.del("http://httpbin.org/delete", null, {
		headers: { "X-Example-Header": "instrumented/delete" },
	});
	check(res, {
	});
};
```

In order to observe this example in action, I recommend to run it with the `--http-debug` too. That way you'll be able to observe the `Traceparent` header being added (note that when using the `b3` or `jaeger` propagators, the header is named differently.

```
go run go.k6.io/k6 run --http-debug samples/experimental/tracing-client.js
```

## Review

⚠️ Although this PR has >1000 lines addition, a considerable chunk of it is testing. I believe the amount of actual logic is somewhat limited ⚠️ 

The commits have been organized to be progressive, and make sense as a sequence; thus, you should review them in order. 

## References

Previous PR: #2853

### DIFF
--- a/js/modules/k6/experimental/tracing/client.go
+++ b/js/modules/k6/experimental/tracing/client.go
@@ -1,0 +1,237 @@
+package tracing
+
+import (
+	"fmt"
+	"net/http"
+	"time"
+
+	"github.com/dop251/goja"
+	"go.k6.io/k6/js/common"
+	"go.k6.io/k6/js/modules"
+	httpmodule "go.k6.io/k6/js/modules/k6/http"
+	"go.k6.io/k6/metrics"
+)
+
+// Client represents a HTTP Client instrumenting the requests
+// it performs with tracing information.
+type Client struct {
+	vu modules.VU
+
+	// opts holds the client's configuration options.
+	opts options
+
+	// propagator holds the client's trace propagator, used
+	// to produce trace context headers for each supported
+	// formats: w3c, b3, jaeger.
+	propagator Propagator
+
+	// requestFunc holds the http module's request function
+	// used to emit HTTP requests in k6 script. The client
+	// uses it under the hood to emit the requests it
+	// instruments.
+	requestFunc HTTPRequestFunc
+}
+
+// HTTPRequestFunc is a type alias representing the prototype of
+// k6's http module's request function
+type HTTPRequestFunc func(method string, url goja.Value, args ...goja.Value) (*httpmodule.Response, error)
+
+// NewClient instantiates a new tracing Client
+func NewClient(vu modules.VU, opts options) *Client {
+	rt := vu.Runtime()
+
+	// Get the http module's request function
+	httpModuleRequest, err := rt.RunString("require('k6/http').request")
+	if err != nil {
+		common.Throw(
+			rt,
+			fmt.Errorf(
+				"failed initializing tracing client, "+
+					"unable to require http.request method; reason: %w",
+				err,
+			),
+		)
+	}
+
+	// Export the http module's request function goja.Callable as a Go function
+	var requestFunc HTTPRequestFunc
+	if err := rt.ExportTo(httpModuleRequest, &requestFunc); err != nil {
+		common.Throw(
+			rt,
+			fmt.Errorf("failed initializing tracing client, unable to export http.request method; reason: %w", err),
+		)
+	}
+
+	client := &Client{vu: vu, requestFunc: requestFunc}
+	if err := client.Configure(opts); err != nil {
+		common.Throw(
+			rt,
+			fmt.Errorf("failed initializing tracing client, invalid configuration; reason: %w", err),
+		)
+	}
+
+	return client
+}
+
+// Configure configures the tracing client with the given options.
+func (c *Client) Configure(opts options) error {
+	if err := opts.validate(); err != nil {
+		return fmt.Errorf("invalid options: %w", err)
+	}
+
+	switch opts.Propagator {
+	case "w3c":
+		c.propagator = &W3CPropagator{}
+	case "jaeger":
+		c.propagator = &JaegerPropagator{}
+	default:
+		return fmt.Errorf("unknown propagator: %s", opts.Propagator)
+	}
+
+	c.opts = opts
+
+	return nil
+}
+
+// Request instruments the http module's request function with tracing headers,
+// and ensures the trace_id is emitted as part of the output's data points metadata.
+func (c *Client) Request(method string, url goja.Value, args ...goja.Value) (*httpmodule.Response, error) {
+	rt := c.vu.Runtime()
+
+	// The http module's request function expects the first argument to be the
+	// request body. If no body is provided, we need to pass null to the function.
+	if len(args) == 0 {
+		args = []goja.Value{goja.Null()}
+	}
+
+	traceID := TraceID{
+		Prefix: k6Prefix,
+		Code:   k6CloudCode,
+		Time:   time.Now(),
+	}
+
+	encodedTraceID, err := traceID.Encode()
+	if err != nil {
+		common.Throw(rt, fmt.Errorf("failed to encode the generated trace ID; reason: %w", err))
+	}
+
+	// Produce a trace header in the format defined by the
+	// configured propagator.
+	traceContextHeader, err := c.propagator.Propagate(encodedTraceID)
+	if err != nil {
+		common.Throw(rt, fmt.Errorf("failed to propagate trace ID; reason: %w", err))
+	}
+
+	// update the `params` argument with the trace context header
+	// so that it can be used by the http module's request function.
+	args, err = c.instrumentArguments(traceContextHeader, args...)
+	if err != nil {
+		common.Throw(rt, fmt.Errorf("failed to instrument request arguments; reason: %w", err))
+	}
+
+	// Add the trace ID to the VU's state, so that it can be
+	// used in the metrics emitted by the HTTP module.
+	c.vu.State().Tags.Modify(func(t *metrics.TagsAndMeta) {
+		t.SetMetadata(metadataTraceIDKeyName, encodedTraceID)
+	})
+
+	response, err := c.requestFunc(method, url, args...)
+	if err != nil {
+		common.Throw(rt, err)
+	}
+
+	// Remove the trace ID from the VU's state, so that it doesn't
+	// leak into other requests.
+	c.vu.State().Tags.Modify(func(t *metrics.TagsAndMeta) {
+		t.DeleteMetadata(metadataTraceIDKeyName)
+	})
+
+	return response, nil
+}
+
+// Del instruments the http module's delete method.
+func (c *Client) Del(url goja.Value, args ...goja.Value) (*httpmodule.Response, error) {
+	return c.Request(http.MethodDelete, url, args...)
+}
+
+// Get instruments the http module's get method.
+func (c *Client) Get(url goja.Value, args ...goja.Value) (*httpmodule.Response, error) {
+	// Here we prepend a null value that stands for the body parameter,
+	// that the request function expects as a first argument implicitly
+	args = append([]goja.Value{goja.Null()}, args...)
+	return c.Request(http.MethodGet, url, args...)
+}
+
+// Head instruments the http module's head method.
+func (c *Client) Head(url goja.Value, args ...goja.Value) (*httpmodule.Response, error) {
+	// NB: here we prepend a null value that stands for the body parameter,
+	// that the request function expects as a first argument implicitly
+	args = append([]goja.Value{goja.Null()}, args...)
+	return c.Request(http.MethodHead, url, args...)
+}
+
+// Options instruments the http module's options method.
+func (c *Client) Options(url goja.Value, args ...goja.Value) (*httpmodule.Response, error) {
+	return c.Request(http.MethodOptions, url, args...)
+}
+
+// Patch instruments the http module's patch method.
+func (c *Client) Patch(url goja.Value, args ...goja.Value) (*httpmodule.Response, error) {
+	return c.Request(http.MethodPatch, url, args...)
+}
+
+// Post instruments the http module's post method.
+func (c *Client) Post(url goja.Value, args ...goja.Value) (*httpmodule.Response, error) {
+	return c.Request(http.MethodPost, url, args...)
+}
+
+// Put instruments the http module's put method.
+func (c *Client) Put(url goja.Value, args ...goja.Value) (*httpmodule.Response, error) {
+	return c.Request(http.MethodPut, url, args...)
+}
+
+// instrumentArguments: expects args to be in the format expected by the
+// request method (body, params)
+func (c *Client) instrumentArguments(traceContext http.Header, args ...goja.Value) ([]goja.Value, error) {
+	rt := c.vu.Runtime()
+
+	var paramsObj *goja.Object
+
+	switch len(args) {
+	case 2:
+		// We received both a body and a params argument. In the
+		// event params would be nullish, we'll instantiate
+		// a new object.
+		if isNullish(args[1]) {
+			paramsObj = rt.NewObject()
+			args[1] = paramsObj
+		} else {
+			paramsObj = args[1].ToObject(rt)
+		}
+	case 1:
+		// We only received a body argument
+		paramsObj = rt.NewObject()
+		args = append(args, paramsObj)
+	default:
+		return nil, fmt.Errorf("invalid number of arguments; expected 1 or 2, got %d", len(args))
+	}
+
+	headersObj := rt.NewObject()
+
+	headersValue := paramsObj.Get("headers")
+	if !isNullish(headersValue) {
+		headersObj = headersValue.ToObject(rt)
+	}
+
+	if err := paramsObj.Set("headers", headersObj); err != nil {
+		return args, err
+	}
+
+	for key, value := range traceContext {
+		if err := headersObj.Set(key, value); err != nil {
+			return args, fmt.Errorf("failed to set the trace header; reason: %w", err)
+		}
+	}
+
+	return args, nil
+}
--- a/js/modules/k6/experimental/tracing/client.go
+++ b/js/modules/k6/experimental/tracing/client.go
@@ -96,8 +96,6 @@ func (c *Client) Configure(opts options) error {
 // Request instruments the http module's request function with tracing headers,
 // and ensures the trace_id is emitted as part of the output's data points metadata.
 func (c *Client) Request(method string, url goja.Value, args ...goja.Value) (*httpmodule.Response, error) {
-	rt := c.vu.Runtime()
-
 	// The http module's request function expects the first argument to be the
 	// request body. If no body is provided, we need to pass null to the function.
 	if len(args) == 0 {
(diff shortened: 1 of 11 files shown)


oleiade:
Heads-up @olegbespalov, I removed you from reviewers to ensure consistency in the PR-chain. This PR is indeed part of a chain, and the bot has automatically selected @imiric and @codebien for the first and last one. I aligned the reviewers for this one as I believe it will make the review easier for reviewers if they review the three of them.

If you wish to review this module, and have time for it, feel free to add yourself back or ping me to do; I'll gladly do so 🙇🏻 👍🏻

codebien:
@oleiade can you change the base branch to `experimental/tracing-module-propagators`, please? Otherwise, we see the changes from 1/3 here also.

oleiade:
@codebien done ✅ thanks for pointing it out 🙇🏻

js/modules/k6/experimental/tracing/client.go
```diff
@@ -0,0 +1,243 @@
...
+	// uses it under the hood to emit the requests it
+	// instruments.
+	requestFunc HTTPRequestFunc
+}
+
+type (
+	// HTTPRequestFunc is a type alias representing the prototype of
+	// k6's http module's request function
+	HTTPRequestFunc func(method string, url goja.Value, args ...goja.Value) (*httpmodule.Response, error)
+)
```
mstoykov, line 40: nit: I would have done the type definition on one line with the comment on top

js/modules/k6/experimental/tracing/client.go
```diff
@@ -0,0 +1,243 @@
...
+
+// NewClient instantiates a new tracing Client
+func NewClient(vu modules.VU, opts options) *Client {
+	rt := vu.Runtime()
+
+	// Instantiate the http module in our runtime and get its default exported object
+	httpModuleObj, ok := httpmodule.New().NewModuleInstance(vu).Exports().Default.(*goja.Object)
```
mstoykov, line 47: Have you tried executing `require('k6/http')` instead. That will use the whole machinery with modules and will give you arguably a more stable code (as you won't depend on the internal API of k6, but of the external one that is used by k6 scripts).

The major downside is that at this time it means you will need to polyfill `require` for your tests, but I would expect it will work perfectly in the real case.

js/modules/k6/experimental/tracing/client.go
```diff
@@ -0,0 +1,243 @@
...
+	if !ok {
+		common.Throw(rt, errors.New("failed to initialize tracing client, unable to load http module"))
+	}
+
+	// Get the http module's request function
+	httpModuleRequest := httpModuleObj.Get("request")
```
mstoykov, line 53: You can even do `require('k6/http').request`

js/modules/k6/experimental/tracing/client.go
```diff
@@ -0,0 +1,243 @@
...
+			rt,
+			fmt.Errorf("failed initializing tracing client, invalid configuration; reason: %w", err),
+		)
+	}
+
+	return client
+}
+
+// Configure configures the tracing client with the given options.
+func (c *Client) Configure(opts options) error {
```
codebien, line 76: Is it required to have it exported?

js/modules/k6/experimental/tracing/client.go
```diff
@@ -0,0 +1,243 @@
...
+	// The http module's request function expects the first argument to be the
+	// request body. If no body is provided, we need to pass null to the function.
+	if len(args) == 0 {
+		args = []goja.Value{goja.Null()}
+	}
+
+	traceID := TraceID{
+		Prefix:            k6Prefix,
+		Code:              k6CloudCode,
+		UnixTimestampNano: uint64(time.Now().UnixNano()),
```
codebien, line 111: Does it make sense to have it in unit64 if the stdlib uses int64?

js/modules/k6/experimental/tracing/client.go
```diff
@@ -0,0 +1,243 @@
...
+	for key, value := range traceContext {
+		if err := headersObj.Set(key, value); err != nil {
+			return args, fmt.Errorf("failed to set the trace header; reason: %w", err)
+		}
+	}
+
+	if err := paramsObj.Set("headers", headersObj); err != nil {
+		return args, err
+	}
+
```
mstoykov, line 236: I would move this set in the `if` above where we check if it nullish

js/modules/k6/experimental/tracing/client.go
```diff
@@ -0,0 +1,243 @@
...
+	// At this point we know that the params object has a headers field,
+	// with tracing information, and that it sits in position 2 in the slice
+	// so we modify the slice in place.
+	args[1] = paramsObj
```
mstoykov, line 240: I would also only do this above where I create new ones.

js/modules/k6/experimental/tracing/options.go
```diff
@@ -0,0 +1,40 @@
...
+	Baggage map[string]string `js:"baggage"`
+}
+
+func (i *options) validate() error {
+	var (
+		isW3C    = i.Propagator == W3CPropagatorName
+		isB3     = i.Propagator == B3PropagatorName
+		isJaeger = i.Propagator == JaegerPropagatorName
+	)
+	if !isW3C && !isB3 && !isJaeger {
```
mstoykov, line 27: Is there a reason we just don't do the equal comparision directly in the `if` sttatement?

js/modules/k6/experimental/tracing/client_test.go
```diff
@@ -0,0 +1,159 @@
...
+)
+
+func TestClientInstrumentArguments(t *testing.T) {
+	t.Parallel()
+
+	// Although the traceparent header is case insensitive, the
+	// Go http.Header sets it capitalized.
+	traceparentHeaderName := "Traceparent"
+	testTraceID := "00-4bf92f3577b34da6a3ce929d0e0e4736-00f067aa0ba902b7-00"
+
```
mstoykov, line 20: nit: can you make those constants

js/modules/k6/experimental/tracing/client_test.go
```diff
@@ -0,0 +1,159 @@
...
+	t.Run("1 arg should initialize params successfully", func(t *testing.T) {
+		t.Parallel()
+
+		testSetup := modulestest.NewRuntime(t)
+		rt := testSetup.VU.Runtime()
+		client := Client{vu: testSetup.VU}
+		traceContextHeader := http.Header{}
+		traceContextHeader.Add(traceparentHeaderName, testTraceID)
+
+		gotArgs, gotErr := client.instrumentArguments(traceContextHeader, goja.Null())
```
mstoykov, line 42: This seems to repeaat in almost all of the cases so I would  get it out. Maybe in a struct that has the fields that I am using instead of returning multiple results, most of which are not used. See https://github.com/grafana/k6/pull/2845/commits/3aedeedfb42ade8994a286d7756723c4b75b3f78 for a variant of what I am talking about

js/modules/k6/experimental/tracing/goja.go
```diff
@@ -0,0 +1,11 @@
...
+
+import "github.com/dop251/goja"
+
+// isNullish checks if the given value is nullish, i.e. nil, undefined or null.
+//
+// This helper function emulates the behavior of Javascript's nullish coalescing
+// operator (??).
+func isNullish(value goja.Value) bool {
+	return value == nil || goja.IsUndefined(value) || goja.IsNull(value)
+}
```
codebien, line 11: Should we move to `js/common.go`?

cmd/integration_test.go
```diff
@@ -1534,3 +1534,122 @@ func TestPrometheusRemoteWriteOutput(t *testing.T) {
...
+		require.NoError(t, json.Unmarshal(jsonLine, &line))
+
+		if line.Type != "Point" {
+			continue
+		}
+
+		// Filter metric samples which are not related to http
+		if !strings.HasPrefix(line.Metric, "http_") {
+			continue
+		}
```
mstoykov, line 1594: because of those continues (and hte above one) there is a chance that a test will just never go to the below lines. 

I would recommend a boolean that you set to see that at least once you got through the whole check

js/modules/k6/experimental/tracing/options.go
```diff
@@ -0,0 +1,40 @@
...
+		return fmt.Errorf("unknown propagator: %s", i.Propagator)
+	}
+
+	if i.Sampling != nil {
+		return errors.New("sampling is not yet supported")
```
codebien, line 32: s/not yet supported/not supported/

I don't know the plan, but it sounds like a promise. Are we going to add it soon?

mstoykov (commented):
Looks good in general but I have a bunch of comments

js/modules/k6/experimental/tracing/options.go
oleiade: It's a style recommendation I discovered in the Google style guide, which, I find, helps with readability, as the `if` statement ends up much shorter and reads almost like a sentence. I'm not hung onto this though, if you have a strong opinion against that, I'm happy to drop it 🙇🏻

js/modules/k6/experimental/tracing/client.go
oleiade: It is intentional indeed so that users can reconfigure their client on the fly. It will especially prove useful at a later point to give the ability to change the sampling rate and baggage attached to requests 👍🏻 🙇🏻

js/modules/k6/experimental/tracing/client.go
oleiade: I had the same question, and it is indeed a requirement for our internal projects. See [discussion](https://github.com/grafana/k6/pull/2838#discussion_r1064394243) 👍🏻 🙇🏻

js/modules/k6/experimental/tracing/options.go
oleiade: Yes, I have a branch ready, and will open a PR as soon as the first chain is merged 👍🏻

js/modules/k6/experimental/tracing/goja.go
oleiade: That's a good idea 🤔 What do you think @mstoykov ?

mstoykov: I am okay with that, but not certain if that shouldn't be done in a separate PR and also use it across the codebase - which will make it bigger. 

IMO we can do that as a refactoring PR in the cooldown period.

oleiade: Let's do it this way indeed then 👍🏻 I took note to do that during the cooldown 🙇🏻

js/modules/k6/experimental/tracing/client.go
codebien: I'm not getting what's the issue for them, there is an API in between from our data type and their. So why should it be an issue for them?

oleiade: Summoning @Blinkuu to answer those questions 🪄 🧙🏻 

PS: be mindful of not leaking private information here 🙏🏻

cmd/integration_test.go
oleiade: Good point, I modified the test to do just that indeed 👍🏻

js/modules/k6/experimental/tracing/client_test.go
oleiade: Initially, I wanted to make table tests, but it appeared hard, if possible to do it in a comprehensible/maintainable way. Your solution is a good tradeoff indeed, and I applied it 👍🏻

js/modules/k6/experimental/tracing/client.go
oleiade: Removed it and switched `require('k6/http').request` as recommended in another comment. 

I had troubles testing the module when it uses `require` under the hood, as `require` is only available from an initcontext, which, of course `goja.Runtime` doesn't have. That's why I resorted to integration tests. If you have an example of how to polyfill `require` available somewhere, that would be super valuable to me though 👍🏻 🙇🏻

js/modules/k6/experimental/tracing/client.go
oleiade: Resolved internally. I'll be changing this field to an `int64` 🤝

js/modules/k6/experimental/tracing/options.go
codebien: Can you add a to do for it, please?

codebien:
@oleiade you should rebase the chain with the master because we updated our e2e tests strategy.

oleiade:
@codebien I've rebased the PR chain on the latest state of `master`, and updated the integration tests accordingly 👍🏻

js/modules/k6/experimental/tracing/tracing.go
```diff
@@ -0,0 +1,35 @@
...
+		common.Throw(rt, errors.New("Client constructor expects a single configuration object as argument; none given"))
+	}
+
+	var opts options
+	if err := rt.ExportTo(cc.Arguments[0], &opts); err != nil {
+		common.Throw(rt, fmt.Errorf("unable to parse options object; reason: %w", err))
+	}
+
+	return rt.ToValue(NewClient(t.vu, opts)).ToObject(rt)
+}
```
codebien, line 35: I think this should be a method of ModuleInstance, it doesn't seem to be required as a dedicated struct.

codebien (changes requested):

js/modules/k6/experimental/tracing/tracing.go
oleiade: Fair enough 👍🏻

js/modules/k6/experimental/tracing/module.go
```diff
@@ -38,5 +43,29 @@ func (*RootModule) NewModuleInstance(vu modules.VU) modules.Instance {
...
+		},
+	}
+}
+
+// NewClient is the JS constructor for the tracing.Client
+//
+// It expects a single configuration object as argument, which
+// will be used to instantiate an `Object` instance internally,
+// and will be used by the client to configure itself.
+func (mi *ModuleInstance) NewClient(cc goja.ConstructorCall) *goja.Object {
```
codebien, line 58: It sounds it can be unexported

codebien (changes requested):

codebien (approved):
LGTM :clap:

js/modules/k6/experimental/tracing/client.go
```diff
@@ -0,0 +1,237 @@
...
+
+	traceID := TraceID{
+		Prefix: k6Prefix,
+		Code:   k6CloudCode,
+		Time:   time.Now(),
+	}
+
+	encodedTraceID, err := traceID.Encode()
+	if err != nil {
+		common.Throw(rt, fmt.Errorf("failed to encode the generated trace ID; reason: %w", err))
```
mstoykov, line 115: Given that you here can return an error is there a reason why:
1. you are not returning it
2. don't just not have the err.


I prefer not `common.Throw` where possible as that works better with go code IMO and it is automatically translated. But there are cases where that is either too much work for very little benefit  - not interopping with go code in a wayt that matter.

mstoykov (approved):
Left a nitpick about style with common.Throw, but LGTM even with it

js/modules/k6/experimental/tracing/client.go
oleiade: Great catch! This is a leftover from the previous implementation 🤦🏻 I'll fix it in a fixup commit 🙇🏻

js/modules/k6/experimental/tracing/client_test.go
```diff
@@ -0,0 +1,166 @@
...
+		assert.Len(t, gotArgs, 2)
+		assert.Equal(t, goja.Null(), gotArgs[0])
+
+		gotParams := gotArgs[1].ToObject(rt)
+		assert.NotNil(t, gotParams)
+		gotHeaders := gotParams.Get("headers").ToObject(rt)
+		assert.NotNil(t, gotHeaders)
+		gotTraceParent := gotHeaders.Get(traceparentHeaderName)
+		assert.NotNil(t, gotTraceParent)
+		assert.Equal(t, testTraceID, gotTraceParent.String())
```
mstoykov, line 72: nit: In general I prefer to user `require` as it will stop the execution directly which helps in two ways:
1. I won't get a wall of text when i break somethign listing 20 things that went wrong, when the first thing that went wrong is likely to tell me what I neeed to know to fix it.
2. In this particular case `assert.NotNil` failing means that hte test *will* panic with nil exception. Arguably you can mix assert and require, but I have rarely been in a case taht this makes sense :shrug:

mstoykov (approved):

codebien (changes requested):
Blocking so we remember to rebase the fixups before merging

mstoykov:
> Blocking so we remember to rebase the fixups before merging


Won't this just be "squashing" it in the previours PR ?

oleiade:
Indeed, I would expect to squash this in the 1/3 PR, which would obfuscate the fixups. If you have a strong opinion on this I'm happy to rebase them anyways though 👍🏻

js/modules/k6/experimental/tracing/client_test.go
oleiade: In general, I try to use `require` when it's a requirement for the test, as in: "this is required even to perform the test in the first place", and use `assert` for stuff that's tested. I guess I approach it in a rather semantic way, "this is required to perform the test", and "this asserts the behavior of the tested code is correct".

It's also how I understood it to be used in the tests I've been exposed.

Do you have a  strong opinion on this, or would it be fine if I keep it like this? I'm okay with either way.

codebien (approved):
>Indeed, I would expect to squash this in the 1/3 PR
I thought we would keep the history, if we are going to squash then okay for me.

PR #2855, opened 2023-01-13
Experimental tracing module (3/3): define and expose a `instrumentHTTP` function

oleiade:
## About

To keep Pull Requests as small as possible and reduce the review burden, this PR is the third one in a chain of three implementing the decided API design for the experimental tracing module.

warning ✋🏻 This PR is part of a PR chain, and is based on https://github.com/grafana/k6/pull/2854

This second PR builds upon #2853 and #2854  and adds an `instrumentHTTP` function to k6/experimental/tracing.

Using this function in their k6 scripts, once, users can automatically have all their calls to the `http` module's methods (`http.{del,get,head,options,patch,post,put,request}`) automatically instrumented with tracing information, in the same fashion as described in #2854. Indeed, this function under the hood remaps the `http` module's method to use the `Client`'s methods instead.

The `instrumentHTTP` takes the same option set as the `Client` as input but does not allow for reconfiguration, nor does it allow to perform non instrumented calls. As it is the intended behavior, once called, the `http` module's methods are replaced with the instrumented ones. 

## In action

```javascript
import http from "k6/http";
import { check } from "k6";

// instrumentHTTP will ensure that all requests made by the http module
// will be traced. The first argument is a configuration object that
// can be used to configure the tracer.
//
// Currently supported HTTP methods are: get, post, put, patch, head,
// del, options, and request.
tracing.instrumentHTTP({
	propagator: "w3c",
});

export default () => {
	let res = http.get("http://httpbin.org/get", {
		headers: {
			"X-Example-Header": "instrumented/get",
		},
	});
	check(res, {
	});
};
```

To observe this example in action, I recommend running it with the --http-debug too. That way, you'll be able to observe the Traceparent header being added (note that when using the b3 or jaeger propagators, the header is named differently.

```
go run go.k6.io/k6 run --http-debug samples/experimental/tracing.js
```

## Review

⚠️ Although this PR has >1000 lines addition, a considerable chunk of it is testing. I believe the amount of actual logic is somewhat limited ⚠️ 

The commits have been organized to be progressive and make sense as a sequence; thus, you should review them in order.

## References

Previous: #2853 #2854

### DIFF
--- a/cmd/tests/tracing_module_test.go
+++ b/cmd/tests/tracing_module_test.go
@@ -112,7 +112,6 @@ func TestTracingInstrumentHTTP_W3C(t *testing.T) {
 
 	script := tb.Replacer.Replace(`
 		import http from "k6/http";
-		import { check } from "k6";
 		import tracing from "k6/experimental/tracing";
 
 		tracing.instrumentHTTP({
--- a/cmd/tests/tracing_module_test.go
+++ b/cmd/tests/tracing_module_test.go
@@ -56,37 +56,7 @@ func TestTracingModuleClient(t *testing.T) {
 	jsonResults, err := afero.ReadFile(ts.FS, "results.json")
 	require.NoError(t, err)
 
-	gotHTTPDataPoints := false
-
-	for _, jsonLine := range bytes.Split(jsonResults, []byte("\n")) {
-		if len(jsonLine) == 0 {
-			continue
-		}
-
-		var line sampleEnvelope
-		require.NoError(t, json.Unmarshal(jsonLine, &line))
-
-		if line.Type != "Point" {
-			continue
-		}
-
-		// Filter metric samples which are not related to http
-		if !strings.HasPrefix(line.Metric, "http_") {
-			continue
-		}
-
-		gotHTTPDataPoints = true
-
-		anyTraceID, hasTraceID := line.Data.Metadata["trace_id"]
-		require.True(t, hasTraceID)
-
-		traceID, gotTraceID := anyTraceID.(string)
-		require.True(t, gotTraceID)
-
-		assert.Len(t, traceID, 32)
-	}
-
-	assert.True(t, gotHTTPDataPoints)
+	assertHasTraceIDMetadata(t, jsonResults)
 }
 
 func TestTracingClient_DoesNotInterfereWithHTTPModule(t *testing.T) {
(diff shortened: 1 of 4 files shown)


js/modules/k6/experimental/tracing/tracing.go
```diff
@@ -33,3 +35,61 @@ func (t *Tracing) NewClient(cc goja.ConstructorCall) *goja.Object {
...
+// InstrumentHTTP instruments the HTTP module with tracing headers.
+//
+// When used in the context of a k6 script, it will automatically replace
+// the imported http module's methods with instrumented ones.
+func (t *Tracing) InstrumentHTTP(options options) {
+	rt := t.vu.Runtime()
+
+	// Initialize the tracing module's instance default client,
+	// and configure it using the user-supplied set of options.
+	t.defaultClient = NewClient(t.vu, options)
```
mstoykov, line 48: Does this need to be a field ?

js/modules/k6/experimental/tracing/tracing.go
```diff
@@ -33,3 +35,61 @@ func (t *Tracing) NewClient(cc goja.ConstructorCall) *goja.Object {
...
+	mustSetHTTPMethod("post", httpModuleObj, t.defaultClient.Patch)
+	mustSetHTTPMethod("put", httpModuleObj, t.defaultClient.Patch)
+	mustSetHTTPMethod("request", httpModuleObj, t.defaultClient.Request)
+
+	// Inject the updated HTTP module object in the runtime,
+	// overriding any previously imported one in the process.
+	err = rt.Set("http", httpModuleObj)
+	if err != nil {
+		common.Throw(rt, err)
+	}
```
mstoykov, line 94: This shouldn't be needed as you are directly changign the export. And it won't work if it is imported as anythign else.
```suggestion
```

cmd/integration_test.go
```diff
@@ -1642,6 +1643,286 @@ func TestTracingClient_DoesNotInterfereWithHTTPModule(t *testing.T) {
...
+		require.NoError(t, json.Unmarshal(jsonLine, &line))
+
+		if line.Type != "Point" {
+			continue
+		}
+
+		// Filter metric samples which are not related to http
+		if !strings.HasPrefix(line.Metric, "http_") {
+			continue
+		}
```
mstoykov, line 1914: Same comment as in the previous PR, I guess you can even make this check into a function that returns a true if it ever got a single metric that matches all the conditions.

js/modules/k6/experimental/tracing/tracing.go
oleiade: Probably not indeed 👍🏻

js/modules/k6/experimental/tracing/tracing.go
oleiade: Good catch indeed! 🙇🏻

cmd/tests/tracing_module_test.go
```diff
@@ -127,6 +128,236 @@ func TestTracingClient_DoesNotInterfereWithHTTPModule(t *testing.T) {
...
+func TestTracingInstrumentHTTP_Jaeger(t *testing.T) {
+	t.Parallel()
+	tb := httpmultibin.NewHTTPMultiBin(t)
+
+	gotRequests := 0
+
+	tb.Mux.HandleFunc("/tracing", func(w http.ResponseWriter, r *http.Request) {
+		gotRequests++
+		assert.NotEmpty(t, r.Header.Get(tracing.JaegerHeaderName))
+		assert.Len(t, r.Header.Get(tracing.JaegerHeaderName), 45)
```
mstoykov, line 214: I would argue those are likely better as just inline constants instead of importing the whole package(not that you don't indirectly). This also means that if you somehow change it in the tracing API it won't suddenly break while the test is still passing ;)

cmd/tests/tracing_module_test.go
```diff
@@ -127,6 +128,236 @@ func TestTracingClient_DoesNotInterfereWithHTTPModule(t *testing.T) {
...
+		gotHTTPDataPoints = true
+
+		anyTraceID, hasTraceID := line.Data.Metadata["trace_id"]
+		require.True(t, hasTraceID)
+
+		traceID, gotTraceID := anyTraceID.(string)
+		require.True(t, gotTraceID)
+
+		assert.Len(t, traceID, 32)
+	}
```
mstoykov, line 274: Can we have this as a helper function as it seems to be done multiple times.

cmd/tests/tracing_module_test.go
```diff
@@ -127,6 +128,236 @@ func TestTracingClient_DoesNotInterfereWithHTTPModule(t *testing.T) {
...
 }
 
+func TestTracingInstrumentHTTP_W3C(t *testing.T) {
+	t.Parallel()
+	tb := httpmultibin.NewHTTPMultiBin(t)
+
+	gotRequests := 0
+
+	tb.Mux.HandleFunc("/tracing", func(w http.ResponseWriter, r *http.Request) {
+		gotRequests++
```
codebien, line 138: an atomic should be better for avoiding any race

js/modules/k6/experimental/tracing/module.go
```diff
@@ -18,6 +18,9 @@ type (
 	// ModuleInstance represents an instance of the JS module.
 	ModuleInstance struct {
 		vu modules.VU
+
+		// Client holds the module's default tracing client.
+		*Client
```
mstoykov, line 23: Is there a need for this field? It seems to me it is only used in one place so :shrug:

js/modules/k6/experimental/tracing/module.go
```diff
@@ -69,3 +76,54 @@ func (mi *ModuleInstance) newClient(cc goja.ConstructorCall) *goja.Object {
...
+// the imported http module's methods with instrumented ones.
+func (mi *ModuleInstance) instrumentHTTP(options options) {
+	rt := mi.vu.Runtime()
+
+	// Initialize the tracing module's instance default client,
+	// and configure it using the user-supplied set of options.
+	mi.Client = NewClient(mi.vu, options)
+
+	// Explicitly inject the http module in the VU's runtime.
+	// This allows us to later on override the http module's methods
```
mstoykov, line 92: nitpick: this doesn't prevent someone from calling `instrumentHTTP` multiple times with different settings.

I am not certain what the solution should be here. I would argue changing the client somehow instead of stacking them. But it is probably okay if we just return error for now :shrug:

codebien (changes requested):

js/modules/k6/experimental/tracing/module.go
oleiade: I agree with you, I would find it more convenient and explicit if we threw an error if calling `instrumentHTTP` a second time. We could even make the error explicit along the lines of "it looks like you're attempting to reconfigure the HTTP module's instrumentation on the fly. This is not supported. Consider using the tracing.Client instead".

js/modules/k6/experimental/tracing/module.go
oleiade: Following your comment on throwing errors when `instrumentHTTP` is called twice, I now rely on this field to check if we have already initialized a module-wide client or not 👍🏻

codebien (changes requested):
The CI is failing

oleiade:
@codebien indeed, sorry for the noise... 🙇🏻  It should be fixed now 😸 (once CI completes)

js/modules/k6/experimental/tracing/module.go
```diff
@@ -69,3 +73,63 @@ func (mi *ModuleInstance) newClient(cc goja.ConstructorCall) *goja.Object {
...
 }
+
+// InstrumentHTTP instruments the HTTP module with tracing headers.
+//
+// When used in the context of a k6 script, it will automatically replace
+// the imported http module's methods with instrumented ones.
+func (mi *ModuleInstance) instrumentHTTP(options options) {
+	rt := mi.vu.Runtime()
+
+	if mi.Client != nil {
```
codebien, line 84: Should it be protected from data race? Could it be called concurrently? Apparently, nothing blocks the user from using it in the iteration function.

oleiade: I'm not sure about the potential data race; probably @mstoykov might be more able to give a certain answer on this? My current reasoning is the module is instantiated once for each VU; thus this check will access the `Client` instance only once per-vu at a time, hence, no concurrent access? I might be completely off on this though, as I don't have the VU execution model freshly loaded in my brain 🤔 

Your second point actually brings me to think that considering `instrumentHTTP` is a one-way street at the moment, we probably want to limit it to the init context indeed 🤔

oleiade: Indeed, thanks for the pointer 🙏🏻 I had already done it in xk6-redis too, though 😸 I've added a check, and a couple tests to cover that 🤝

mstoykov: > the module is instantiated once for each VU

Yes this is true . 

There is no possible race condition as it will require that this is called twice *from* js at the same time.

But by specification js is single threaded so there is no way for that to happen or it will be a bug in something else - like using a js runtime concurrently.

codebien (approved):
LGTM

cmd/tests/tracing_module_test.go
```diff
@@ -128,6 +98,229 @@ func TestTracingClient_DoesNotInterfereWithHTTPModule(t *testing.T) {
...
+
+	tb.Mux.HandleFunc("/tracing", func(w http.ResponseWriter, r *http.Request) {
+		atomic.AddInt64(&gotRequests, 1)
+		assert.NotEmpty(t, r.Header.Get("traceparent"))
+		assert.Len(t, r.Header.Get("traceparent"), 55)
+	})
+
+	script := tb.Replacer.Replace(`
+		import http from "k6/http";
+		import { check } from "k6";
```
codebien, line 115:
```suggestion

```
It does seem to be required

cmd/tests/tracing_module_test.go
```diff
@@ -128,6 +98,229 @@ func TestTracingClient_DoesNotInterfereWithHTTPModule(t *testing.T) {
...
+		export default function() {
+		}
+	`)
+
+	ts := getSingleFileTestState(t, script, []string{}, 0)
+	ts.ExpectedExitCode = 107
+	cmd.ExecuteWithGlobalState(ts.GlobalState)
+
+	assert.Contains(t, ts.Stderr.String(), "instrumentHTTP can only be called once")
+}
```
mstoykov, line 285: I kind of prefer if those are not integration tests as they are more expensive and in this case there is no need for them IMO.

The functionality tested is pretty common :
1. exception in an iteration does not abort the whole test
2. an exception in the init context - does.

mstoykov (approved):
LGTM, but if possible I would like to move some of the integration tests to not integration ones :bow:

cmd/tests/tracing_module_test.go
```diff
@@ -128,6 +98,228 @@ func TestTracingClient_DoesNotInterfereWithHTTPModule(t *testing.T) {
...
+		assert.NotEmpty(t, r.Header.Get("traceparent"))
+		assert.Len(t, r.Header.Get("traceparent"), 55)
+
+		assert.NotEmpty(t, r.Header.Get("X-Test-Header"))
+		assert.Equal(t, "test", r.Header.Get("X-Test-Header"))
+	})
+
+	script := tb.Replacer.Replace(`
+		import http from "k6/http";
+		import { check } from "k6";
```
codebien, line 206:
```suggestion
```
@oleiade there are multiple lines of it

cmd/tests/tracing_module_test.go
oleiade: Let me see if I can easily turn them into standard unit tests indeed. I've struggled with `instrumentHTTP` depending on require in the past, but with the changes with made, it might not be such an issue anymore 👍🏻

cmd/tests/tracing_module_test.go
oleiade: Good catch! :eyes: I guess this is the downside of writing JS in a Go string, no eslint, no prettier, no checks whatsoever.

mstoykov (commented):
I guess we are do not have an integration test that instrumenting the API in one file will instrtument it across teh whole k6.

This likely is the most interesting integration test, so I think it is worth doing before merging

oleiade:
@mstoykov thanks a lot for your feedback 🙇🏻 I've replaced the relevant integration tests with unit tests. I've also added an integration test demonstrating `instrumentHTTP` instruments across the whole of k6.

js/modules/k6/experimental/tracing/module_test.go
```diff
@@ -0,0 +1,86 @@
...
+	rt := ts.VU.Runtime()
+	require.NoError(t, rt.Set("instrumentHTTP", m.Exports().Named["instrumentHTTP"]))
+
+	require.NoError(t, rt.Set("require", func(module string) *goja.Object {
+		require.Equal(t, "k6/http", module)
+		export := http.New().NewModuleInstance(ts.VU).Exports().Default
+
+		return rt.ToValue(export).ToObject(rt)
+	}))
+
```
mstoykov, line 79: nit: I feel like we can pull this out in a helper function like in other modules. One that returns a struct with fields of the necessary thing for the test like the modulestest.Runtime, the moduleinstance if necessary and so on.

cmd/tests/tracing_module_test.go
```diff
@@ -236,6 +237,59 @@ func TestTracingInstrumentHTTP_FillsParams(t *testing.T) {
...
+	t.Parallel()
+
+	var gotRequests int64
+
+	tb := httpmultibin.NewHTTPMultiBin(t)
+	tb.Mux.HandleFunc("/tracing", func(w http.ResponseWriter, r *http.Request) {
+		atomic.AddInt64(&gotRequests, 1)
+
+		assert.NotEmpty(t, r.Header.Get("traceparent"))
+		assert.Len(t, r.Header.Get("traceparent"), 55)
```
mstoykov, line 250: nit: Having asserts like this is in general an anti-pattern and the testing package will give you warnings if you ever actually fail here. 

Although maybe with assert instead of require it is fine as it won't call [FailNow](https://pkg.go.dev/testing#T.FailNow) which is the usual problem :thinking:

cmd/tests/tracing_module_test.go
```diff
@@ -236,6 +237,59 @@ func TestTracingInstrumentHTTP_FillsParams(t *testing.T) {
...
+	ts.CmdArgs = []string{"k6", "run", "--out", "json=results.json", "main.js"}
+	ts.ExpectedExitCode = 0
+
+	cmd.ExecuteWithGlobalState(ts.GlobalState)
+
+	jsonResults, err := afero.ReadFile(ts.FS, "results.json")
+	require.NoError(t, err)
+
+	assert.Equal(t, int64(2), atomic.LoadInt64(&gotRequests))
+	assertHasTraceIDMetadata(t, jsonResults)
```
mstoykov, line 290: None of those check that *both* requests have the traceid. Given that we already check for the one in other tests I will recommend just having 1 requests that is made *not* from the module calling `instrumentHTTP`

js/modules/k6/experimental/tracing/module_test.go
oleiade: I see what you mean, and you're right I'll move it in a dedicated `testSetup` struct like in other places 🤝

mstoykov (changes requested):

cmd/tests/tracing_module_test.go
oleiade: It works as I expected, though. I've just forced the test to fail at this statement, and I see no such warnings 🤔 

```
/home/oleiade/Dev/Grafana/k6/cmd/tests/tracing_module_test.go:250:
        	Error Trace:	/home/oleiade/Dev/Grafana/k6/cmd/tests/tracing_module_test.go:250
        	            				/home/oleiade/Dev/Grafana/k6/cmd/tests/server.go:2109
        	            				/home/oleiade/Dev/Grafana/k6/cmd/tests/server.go:2487
        	            				/home/oleiade/Dev/Grafana/k6/cmd/tests/server.go:2947
        	            				/home/oleiade/Dev/Grafana/k6/cmd/tests/server.go:1991
        	            				/home/oleiade/Dev/Grafana/k6/cmd/tests/asm_amd64.s:1594
        	Error:      	"00-1234567890-bf358b23435cbdd7-01" should have 55 item(s), but has 33
        	Test:       	TestTracingInstrummentHTTP_SupportsMultipleTestScripts
```

cmd/tests/tracing_module_test.go
oleiade: You're absolutely right, that's a really good point 👍🏻

cmd/tests/tracing_module_test.go
mstoykov: Do you get a warning if you use `require` :thinking:

oleiade: Nope, same output

```
/home/oleiade/Dev/Grafana/k6/cmd/tests/tracing_module_test.go:250:
        	Error Trace:	/home/oleiade/Dev/Grafana/k6/cmd/tests/tracing_module_test.go:250
        	            				/home/oleiade/Dev/Grafana/k6/cmd/tests/server.go:2109
        	            				/home/oleiade/Dev/Grafana/k6/cmd/tests/server.go:2487
        	            				/home/oleiade/Dev/Grafana/k6/cmd/tests/server.go:2947
        	            				/home/oleiade/Dev/Grafana/k6/cmd/tests/server.go:1991
        	            				/home/oleiade/Dev/Grafana/k6/cmd/tests/asm_amd64.s:1594
        	Error:      	"00-1234567890-3d5a73e6627d9b8e-01" should have 55 item(s), but has 33
        	Test:       	TestTracingInstrummentHTTP_SupportsMultipleTestScripts
```

cmd/tests/tracing_module_test.go
```diff
@@ -128,6 +99,232 @@ func TestTracingClient_DoesNotInterfereWithHTTPModule(t *testing.T) {
...
+
+	ts := getSingleFileTestState(t, script, []string{"--out", "json=results.json"}, 0)
+	cmd.ExecuteWithGlobalState(ts.GlobalState)
+
+	assert.Equal(t, int64(8), atomic.LoadInt64(&gotRequests))
+
+	jsonResults, err := afero.ReadFile(ts.FS, "results.json")
+	require.NoError(t, err)
+
+	assertHasTraceIDMetadata(t, jsonResults)
```
mstoykov, line 237: Kind of the same comment about this not checking that all of those have tracing ids.

Maybe `assertHasTraceIDMetadata` can have a number argument that is how many it expects :shrug:. 

Or to take the expected url(s) (as vararg not required argument)  and check that all of those have trace id metadata sample.

cmd/tests/tracing_module_test.go
mstoykov: Maybe I am mistaken and it only matters if it happens after the [test functions has returned](https://go.dev/play/p/cAtYXKMIOwBc) :thinking:  or if you have sub test and use the parent one :shrug: 

Whichever - it doesn't really matter. Sorry for the noise :bow:

mstoykov (approved):

mstoykov:
I have left one [comment](https://github.com/grafana/k6/pull/2855#discussion_r1087664171) that is not blocking IMO given that it is very unlikely breakage in that case (IMO)

oleiade:
Thanks a lot for the reviews folks 🙇🏻 👏🏻 much appreciated!

For the sake of getting a sense of achievement, I'll go ahead and merge this. I think your point on the `assertHasTraceIDMetadata` is valid @mstoykov and I will try to address it in a further PR 🦕 

🚀 🚀 🚀

Issue #2763, opened 2022-11-10
Migrate from deprecated features in Github Actions


olegbespalov:
# What?

We should update the GitHub Action pipelines since several features that we're using are deprecated and will be removed:

* The `set-output` and `save-state` commands are deprecated and will be disabled soon. We should use Environment Files. (sunset date 31st May 2023).
* Node.js 12 actions are deprecated (sunset ~September 2023). 

# Why?

* https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
* https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/.

PR #2861, opened 2023-01-19
Maintenance/GitHub actions

olegbespalov:
# What?

This change removes deprecated things that will be out of support later this year.

# Why?

Connected to the #2763

### DIFF
--- a/.github/workflows/arm-tests.yml
+++ b/.github/workflows/arm-tests.yml
@@ -9,7 +9,7 @@ jobs:
     runs-on: [self-hosted, Linux, ARM64]
     steps:
       - name: Checkout code
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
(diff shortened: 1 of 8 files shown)


mstoykov (approved):
LGTM, but we still have 2 warnings due to https://github.com/crazy-max/ghaction-chocolatey and https://github.com/kentaro-m/auto-assign using node v12.

Both seem to have bumped them over time, so maybe it is time we bump the versions we use to the latest ones? I am fine with this happening in a separate PR. It seems like it will be deprecated later then `set-output` so we have more time :shrug:

olegbespalov:
@mstoykov, not a big deal, I'll update them too :+1:

oleiade (approved):

olegbespalov:
Actually, I'll open another PR to also prove that auto-assign works as expected

Issue #2728, opened 2022-10-13
`group` doesn't work with async calls well 


mstoykov:
The following script: 
```javascript
import { Counter } from "k6/metrics";
import { group } from "k6";
const delay = () => { return Promise.resolve(); } // this is just a placeholder for something useful.
let l = new Counter("my coutner")

export default () => {
    group("coolgroup", () => {
        l.add(1) // nice

        delay(1).then(() => {
                l.add(1)// not so nice
        })
    })
}
```

Will *not* emit "my counter" with "group=coolgroup". 

To that the user will need to wrap the inner parts of the `then` in a group as in 
```javascript
import { Counter } from "k6/metrics";
import { group } from "k6";
const delay = () => { return Promise.resolve(); }
let l = new Counter("my coutner")

export default () => {
    group("coolgroup", () => {
        l.add(1) // nice

        delay(1).then(() => {
            group("coolgroup", () => {
                l.add(1)// not so nice
            });
        })
    })
}
```

This is because `group` is kind of 

```javascript
import exec from "k6/execution"
function group (name, fn) {
  let oldGroup = exec.vu.tags["group"];
  exec.vu.tags["group"] = oldGroup + ":" + name;
  try {
    fn() // the actual inner function
  } finally {
    exec.vu.tags["group"] = oldGroup;
  }
}
```

As the `then` "callbacks" get called only after the stack is empty the whole `group` code would have been executed, resetting the group back to the root name (which is empty). And when the `then` get's called it will have an empty group. 

The "workaround" works as it just calls `group` again. 

A more hacking but ... more useful workaround is 
```javascript
import { Counter } from "k6/metrics";
import { group } from "k6";
const delay = () => { return Promise.resolve(); }
let l = new Counter("my coutner")

let originalThen = Promise.prototype.then;
Promise.prototype.then = function(onResolved, onRejected) {
    let o = onResolved
    onResolved = function(res) {
        group("coolgroup", () => { // this in practice will need to get hte group name
            o(res);
        })
    }
    // https://stackoverflow.com/questions/55413715/cant-i-overwrite-the-then-function-of-promise
    return originalThen.call(originalThen.call(this, onResolved, onRejected), undefined, error => {
        console.log(`upload the error to server: ${error}`);
        throw error;
    });
}

export default () => {
    group("coolgroup", () => {
        l.add(1) // nice

        delay(1).then(() => {
            l.add(1)// not so nice
        })
    })
}
```

This ... likely has *other* problems. 

[...]

PR #2863, opened 2023-01-19
k6: group special async function treatment

### LINKED ISSUE #2848: Change how `group()` calls async functions
As of #2830 k6 has support `async` functions. 

A such the following code is now valid
```javascript
group("somename", async () => {
    console.log("A");
    await 5;
    console.log("B");
})
console.log("C")
```
And will print `A, C, B` (with new lines in between). 

This is okay but after an internal poll most people expect `C, A, B` - running the remaining parts before calling the `async` function. 

Given that irregardless of what people expect the returned value of `group` will be a promise so in order to get it the stack will need to be empty. 

It seems more reasonable to *not* have some parts of the `group` finish before continuing and some to finish.

## Proposed solution:

Have `group` recognise that it got an async function. And queue it's calling (with the appropriate setting of `group` tag) on the event loop. Or create another promise that immediately is resolved and do the actual `group` call in the resolution callback.

The difference between the two is that using a promise this will be called before anything else that will be called from the event loop. And as such also before any other immediately resolved promise created after the call to `group`.

So the difference in example is:
```javascript
group("somename", async () => {
    console.log("A");
    await 5;
    console.log("B");
})
console.log("C")
Promise.resolve("D").then((s) => {console.log(s)}); // (Promise.resolve returns a resolved promise with the provided value)
```
Wil in the one case write `C, D, A, B` in the other `C, A, D, B`.
`C, D, A, B` as the resolved promise will be executed *before* anything from the k6 event loop can be executed. 
`C, A, D, B` as the promise resolution of the `group` will be queued before the `D` one but then the await will once again 

Recognising it seems to be possible if we take the callback argument as `goja.Value` and get its Prototype. 

### Why not `A, B, C`?

This is currently possible if the code is 

```javascript
await group("somename", async () => {
    console.log("A");
    await 5;
    console.log("B");
})
console.log("C")
```
Which also requires that the code is within an async function in order for `await` to not be a syntax error.

This will still be possible after that as well.

### Is this a breaking change? 

Given that `async function` was not supported up until this point - I would argue not. AFAIK the ways to polyfill this does not return the same Prototype so that will still work the same way as before.

mstoykov:
A group provided with an async function will now:
1. Treat the whole time it take for the async function promise to finisher the duration of the group.
2. It will also use goja's AsyncContextTracker to make it so that the code using `await` within the group will still continue to be tagged with the group after `await` returns.
3. Instead of directly calling the async function it schedules it to be called the next time a promise job will work.

The current AsyncContextTracker is only used for this and as such is directly changed in the `k6` module. In the future there likely will be API so that multiple modules can use it simultaneously, but that seems way too involved to be included in this change and also currently only `group` needs this.

fixes #2848 #2728

### DIFF
--- a/js/modules/k6/asyncGroup.go
+++ b/js/modules/k6/asyncGroup.go
@@ -1,0 +1,75 @@
+package k6
+
+import (
+	"context"
+	"time"
+
+	"github.com/dop251/goja"
+	"go.k6.io/k6/lib"
+	"go.k6.io/k6/metrics"
+)
+
+type groupInstance struct {
+	*lib.Group
+	startTime time.Time
+}
+
+func (g *groupInstance) start() {
+	g.startTime = time.Now()
+}
+
+func (g *groupInstance) finalize(ctx context.Context, state *lib.State) {
+	t := time.Now()
+	ctm := state.Tags.GetCurrentValues()
+	metrics.PushIfNotDone(ctx, state.Samples, metrics.Sample{
+		TimeSeries: metrics.TimeSeries{
+			Metric: state.BuiltinMetrics.GroupDuration,
+			Tags:   ctm.Tags,
+		},
+		Time:     t,
+		Value:    metrics.D(t.Sub(g.startTime)),
+		Metadata: ctm.Metadata,
+	})
+}
+
+func (mi *K6) asyncGroup(name string, fn goja.Callable) (goja.Value, error) {
+	rt := mi.vu.Runtime()
+	state := mi.vu.State()
+	p, res, _ := rt.NewPromise()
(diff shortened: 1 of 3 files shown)


mstoykov:
Moving this to a draft in favor of https://github.com/grafana/k6/pull/2875

aakash3771:
How to get notified when this is fixed?

mstoykov:
For more information you likely should follow https://github.com/grafana/k6/issues/2728 as this PR at this time is very unlikely to be merged as is.

mstoykov:
I am closing this PR as it is really unlikely we will go down this route

PR #2870, opened 2023-01-23
Update the GitHub Actions to the latest versions

### LINKED ISSUE #2763: Migrate from deprecated features in Github Actions

olegbespalov:
# What?

These changes finalize the removalof the deprecated Node 12.X from our CI

Related to the:
* https://github.com/kentaro-m/auto-assign-action/issues/99
* https://github.com/kentaro-m/auto-assign-action/pull/98

# Why?

Resolves #2763

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -223,7 +223,7 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v3
       - name: Install pandoc
-        uses: crazy-max/ghaction-chocolatey@b6061d587628735be315d74358228b83a7dba9a7
+        uses: crazy-max/ghaction-chocolatey@90deb87d9fbf0bb2f022b91e3bf11b4441cddda5 # v2.1.0
         with:
           args: install -y pandoc
       - name: Install wix tools
--- a/.github/workflows/pr-auto-assign.yml
+++ b/.github/workflows/pr-auto-assign.yml
@@ -8,6 +8,6 @@ jobs:
     if: ${{ join(github.event.pull_request.requested_reviewers.*.login, ',') == '' }}
     runs-on: ubuntu-latest
     steps:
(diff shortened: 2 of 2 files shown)


mstoykov (approved):

olegbespalov:
`kentaro-m/auto-assign-action@v1.2.4` is still not at Node 16 :cry:

olegbespalov:
:thinking: but it should https://github.com/kentaro-m/auto-assign-action/pull/98

oleiade (approved):

codebien (dismissed):

na-- (changes requested):
I :+1: the update, but :-1: using tags instead of specific commits. 

Using tags doesn't offer any security benefits if someone hacks the repos (or hacks an account that can push tags to these repos). They will be able to delete the old tag, push a new tag, and exfiltrate all of our secrets. Doing so with a commit is going to be quite difficult, they'd need a SHA1 collision with the same hash, given git's content-addressable nature...

codebien:
I'm not sure if it works as I expect  but it sounds like we should also limit with https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

olegbespalov:
@codebien sounds good/reasanoble, but should we maybe create a dedicated issue to limit tokens across all CI, WDYT :thinking:

na-- (approved):
Thanks for changing it, and yeah, permissions for different jobs sound awesome, I didn't know that was a thing!

codebien (approved):

PR #2874, opened 2023-01-24
Experimental tracing module (4/3): support for sampling

oleiade:
## About

This Pull Request adds probabilistic sampling support to the tracing module. Using it, users can specify the percentage of their instrumented requests which should have their trace context header's sampled field set to true.

✋🏻 This Pull Request is a nice-to-have built on top of the initial PR chain meant to add a `k6/experimental/tracing` module to k6 (#2853, #2854, #2855) and should be treated separately and merged only once the initial PR chain is merged to master ✋🏻 

From a user-interface perspective, this feature is introduced as an option meant to be set during either the instantiation of a `tracing.Client`, or during a call to `instrumentHTTP`. It consists of a single integer field of the tracing options, whose value, within the `0 <= n <= 100` bounds, is meant to represent a percentage.

```javascript
// The sampling option is available when instantiating a Client
const sampledHTTP = new tracing.Client({
	propagator: "w3c",
	
	// Up to 55% of the requests emitted using the client instance's
	// methods will have their trace context header's sampled field
	// set to true.
	sampling: 55,
})

// It is also available when using the instrumentHTTP function
tracing.InstrumentHTTP({
	propagator: "w3c",
	
	// Up to 27% of the requests made with the HTTP module's functions
	// will have their trace context header's sampled field set to true
	// from this point onward 
	sampling: 27,
})
```

This feature is implemented using probabilistic sampling. Meaning that over a large series of iterations, the number of requests with their sampled flag set to true will tend to converge to the selected `sampling` rate value. That was a design choice without explicit constraints, preferring randomness over accuracy: each trace context generation evaluates its chance of being sampled independently. If you find an implementation emphasizing more accuracy, I'm open to going that route; just let me know.

Note that regardless of the sampled header flag's value, the `trace_id` metadata will be attached to each output sample.

### DIFF
--- a/cmd/tests/tracing_module_test.go
+++ b/cmd/tests/tracing_module_test.go
@@ -1,0 +1,140 @@
+package tests
+
+import (
+	"bytes"
+	"encoding/json"
+	"net/http"
+	"strings"
+	"testing"
+
+	"github.com/spf13/afero"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"go.k6.io/k6/cmd"
+	"go.k6.io/k6/lib/testutils/httpmultibin"
+)
+
+func TestTracingModuleClient(t *testing.T) {
+	t.Parallel()
+	tb := httpmultibin.NewHTTPMultiBin(t)
+
+	gotRequests := 0
+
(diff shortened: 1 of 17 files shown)


oleiade:
We'd like to address https://github.com/grafana/k6/pull/2855 in the context of this Pull Request 🤝

PR #2875, opened 2023-01-25
Disable async function as group and check callback

mstoykov:
The general inconsistencies between group as an idea and asynchronous code makes it not great fit and such we will try to discourage users to mix them.

Updates #2728

### DIFF
--- a/js/modules/k6/k6.go
+++ b/js/modules/k6/k6.go
@@ -95,7 +95,7 @@ func (mi *K6) Group(name string, val goja.Value) (goja.Value, error) {
 		return nil, ErrGroupInInitContext
 	}
 
-	if val == nil || goja.IsNull(val) {
+	if isNullish(val) {
 		return nil, errors.New("group() requires a callback as a second argument")
 	}
 	fn, ok := goja.AssertFunction(val)
--- a/js/modules/k6/k6.go
+++ b/js/modules/k6/k6.go
@@ -24,7 +24,7 @@ var (
 )
 
 const asyncFunctionNotSupportedMsg = "%s: async functions are not supported as arguments, " +
-	"please see https://github.com/grafana/k6/issues/2728#issuecomment-1404747660 for more info"
+	"please see https://k6.io/docs/javascript-api/k6/group/#working-with-async-functions for more info"
 
 type (
 	// RootModule is the global module instance that will create module
--- a/js/modules/k6/k6.go
+++ b/js/modules/k6/k6.go
@@ -3,6 +3,7 @@ package k6
 
 import (
 	"errors"
+	"fmt"
 	"math/rand"
 	"sync/atomic"
 	"time"
@@ -21,6 +22,9 @@ var (
 	// ErrCheckInInitContext is returned when check() are using in the init context.
 	ErrCheckInInitContext = common.NewInitContextError("Using check() in the init context is not supported")
 )
+
+const asyncFunctionNotSupportedMsg = "%s: async functions are not supported as arguments, " +
+	"please see https://github.com/grafana/k6/issues/2728#issuecomment-1404747660 for more info"
 
 type (
 	// RootModule is the global module instance that will create module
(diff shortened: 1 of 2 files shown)


js/modules/k6/k6.go
```diff
@@ -85,16 +89,24 @@ func (mi *K6) RandomSeed(seed int64) {
...
 // Group wraps a function call and executes it within the provided group name.
-func (mi *K6) Group(name string, fn goja.Callable) (goja.Value, error) {
+func (mi *K6) Group(name string, val goja.Value) (goja.Value, error) {
 	state := mi.vu.State()
 	if state == nil {
 		return nil, ErrGroupInInitContext
 	}
 
-	if fn == nil {
+	if val == nil || goja.IsNull(val) {
```
na--, line 98:
```suggestion
	if isNullish(val) {
```

js/modules/k6/k6.go
```diff
@@ -22,6 +23,9 @@ var (
 	ErrCheckInInitContext = common.NewInitContextError("Using check() in the init context is not supported")
 )
 
+const asyncFunctionNotSupportedMsg = "%s: async functions are not supported as arguments, " +
+	"please see https://github.com/grafana/k6/issues/2728#issuecomment-1404747660 for more info"
```
na--, line 27:
```suggestion
const asyncFunctionNotSupportedMsg = "async functions are currently unsupported as %s() arguments"
```

I don't think we need to link to an issue comment in an exception message, that would just add noise. And if we really think we should attach a longer explanation, then we should link to a page in the docs or a standalone issue with the information, https://github.com/grafana/k6/issues/2728 has way too many comments and unrelated discussion.

mstoykov: > , that would just add noise

Given that this error means your code doesn't work at all - adding noise seems like a good idea. I even wondered if I shouldn't abort the whole test but decided against that based on complexity. 

> should link to a page in the docs or a standalone issue with the information

That is probably better, but likely will not be any less noisy  and will mean that we will need to make a page specifically for that.

Linking the specific comment IMO is best -if somebody wants to use async function the specifically linked comment goes over why this is currently not supported and the issue has more information if somebody needs it.

If this is a problem arguably us being told that users are hitting it is a good idea.

na--: I am not convinced :thinking: 

> Given that this error means your code doesn't work at all

How is that any different than any other error? :sweat_smile: Though as I wrote in https://github.com/grafana/k6/issues/1681, I do want identifiable and easily searchable error IDs, so maybe we can start here :thinking: How about
```suggestion
const asyncFunctionNotSupportedMsg = "async functions are currently unsupported as %s() arguments " +
	" (k6AsyncCallbacksForbidden)";
```

> Linking the specific comment IMO is best -if somebody wants to use async function the specifically linked comment goes over why this is currently not supported and the issue has more information if somebody needs it.

I would be fine with providing more information in the error, if it contained some useful workaround for users. But sending them on a quest to read a long issue, only to indeed discover that there is no way to use `async` and `group`, that seems a bit rude :sweat_smile: 

Even if we ignore UX and the main goal is to collect more feedback, this is probably not the right way to go about it and opening a new issue with that summary a link to the old ones is better :thinking:

andrewslotin: > Even if we ignore UX and the main goal is to collect more feedback, this is probably not the right way to go about it and opening a new issue with that summary a link to the old ones is better

A humble suggestion to update the doc for `group()` describing the reason for this limitation there and reference it in the error message.

codebien: >Even if we ignore UX and the main goal is to collect more feedback,

I think it is important in this case, because the API is under discussion and maybe a user that hits this path has a  use case that we would be good to know. If we can have a cleaner issue I'm happy with it, but I think the current is better than nothing.

```suggestion
const asyncFunctionNotSupportedMsg = "%s: async functions are not supported as arguments, " +
	"check https://github.com/grafana/k6/issues/2728#issuecomment-1404747660 for sharing your experience and read about more details"
```

mstoykov: WDYT https://github.com/grafana/k6-docs/pull/1011

mstoykov: WDYT about the new URL. We can make the same message for the `check` API as well and redirect to the correct one.

Or leave it like that - I doubt anyone will try to use `check` with async functions :crossed_fingers:

js/modules/k6/k6.go
```diff
@@ -22,6 +23,9 @@ var (
 	ErrCheckInInitContext = common.NewInitContextError("Using check() in the init context is not supported")
 )
 
+const asyncFunctionNotSupportedMsg = "%s: async functions are not supported as arguments, " +
```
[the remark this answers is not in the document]
na--: or this
```suggestion
const asyncFunctionNotSupportedMsg = "%s() does not support async functions as arguments, " +
```

na-- (approved):
LGTM except the minor error message nitpick, to make this error match the other ones.

js/modules/k6/k6.go
na--: I think it's fine to leave it as it is, I very much doubt that someone will ever use an `async` function in a `check()`, but if they do, linking to the group warning is good enough

codebien (approved):

na-- (approved):

codebien (approved):

codebien (approved):

PR #2877, opened 2023-01-26
http: add asyncRequest

### LINKED ISSUE #2825: Add `http.asyncRequest`
### Feature Description

With the addition of event loop months ago, k6 did get the ability to have asynchronous operations, like all the other js tools ;). 

While we have added some *experimental* APIs and will likely add more that use this there is a small problem.

Using asynchronous APIs in combination with synchronous ones that take long time will *block* the event loop. Which will prevent the asynchronous calls from completing.

```javascript
export default () => {
  const ws = new WebSocket(url);
  ws.addEventListener("message", (e) => {
    console.log(e);
  });
  http.get(url2);
  http.get(url3);
  // more http requests
  http.get(url10);
}
```
This will effectively not log a single message until all the http requests are finished. This might take a few minutes. If the websocket was supposed to return something back there is a very good chance the server would've timeouted this connection. 

In order for this to not be the case we need asynchronous http call. 

While getting the [new-http](https://github.com/grafana/k6/issues?q=is%3Aopen+is%3Aissue+label%3Anew-http) API will be best this is unlikely to happen soon so a stop gap solution seems like a good middle ground.

Implement `asyncRequest` with the same signature as `request` but instead of returning the response directly, it will return a Promise that will be resolved with the response or rejected with an error in the same cases as `http.request` would do it.

No other changes are at this point proposed including:
1. changes to any of the arguments - as that will likely be more confusing
2. no `http.asyncGet`, `http.asyncPost` and friends. While they might be slightly easier to write, there are already confusion about which have bodies and which not and arguably are saving very small amount of writing. 
3. no `http.asyncBatch` - given that `http.batch` is mostly used to do multiple requests at the same time it seems less useful to have this in the case that you can just make a bunch of `asyncRequest`s and `Promise.all` on them for example.

The above might change, although the hope is that this will reduce the maintenance burden, while still giving enough functionality to users until the new http API lands. 

_No response_

mstoykov:
This is implementation of http.request but it return a promise and does all the networking off the event loop.

The code is pretty simple and seems fairly safe given that most possible race conditions would also happen in the case of
`http.batch`.

Even with that there is at least one test that previously couldn't have happened with http.batch.

closes #2825

### DIFF
--- a/js/modules/k6/experimental/tracing/client.go
+++ b/js/modules/k6/experimental/tracing/client.go
@@ -6,7 +6,6 @@ import (
 	"time"
 
 	"github.com/dop251/goja"
-	"go.k6.io/k6/js/common"
 	"go.k6.io/k6/js/modules"
 	httpmodule "go.k6.io/k6/js/modules/k6/http"
 	"go.k6.io/k6/metrics"
@@ -30,47 +29,54 @@ type Client struct {
 	// uses it under the hood to emit the requests it
 	// instruments.
 	requestFunc HTTPRequestFunc
+
+	// asyncRequestFunc holds the http module's asyncRequest function
+	// used to emit HTTP requests in k6 script. The client
+	// uses it under the hood to emit the requests it
+	// instruments.
+	asyncRequestFunc HTTPAsyncRequestFunc
 }
 
 // HTTPRequestFunc is a type alias representing the prototype of
 // k6's http module's request function
-type HTTPRequestFunc func(method string, url goja.Value, args ...goja.Value) (*httpmodule.Response, error)
+type (
+	HTTPRequestFunc      func(method string, url goja.Value, args ...goja.Value) (*httpmodule.Response, error)
+	HTTPAsyncRequestFunc func(method string, url goja.Value, args ...goja.Value) (*goja.Promise, error)
+)
 
 // NewClient instantiates a new tracing Client
-func NewClient(vu modules.VU, opts options) *Client {
+func NewClient(vu modules.VU, opts options) (*Client, error) {
 	rt := vu.Runtime()
 
-	// Get the http module's request function
-	httpModuleRequest, err := rt.RunString("require('k6/http').request")
+	// Get the http module
+	httpModule, err := rt.RunString("require('k6/http')")
 	if err != nil {
-		common.Throw(
-			rt,
-			fmt.Errorf(
-				"failed initializing tracing client, "+
-					"unable to require http.request method; reason: %w",
-				err,
-			),
-		)
+		return nil,
+			fmt.Errorf("failed initializing tracing client, unable to require k6/http module; reason: %w", err)
 	}
+	httpModuleObject := httpModule.ToObject(rt)
 
 	// Export the http module's request function goja.Callable as a Go function
 	var requestFunc HTTPRequestFunc
-	if err := rt.ExportTo(httpModuleRequest, &requestFunc); err != nil {
-		common.Throw(
-			rt,
-			fmt.Errorf("failed initializing tracing client, unable to export http.request method; reason: %w", err),
-		)
+	if err := rt.ExportTo(httpModuleObject.Get("request"), &requestFunc); err != nil {
+		return nil,
+			fmt.Errorf("failed initializing tracing client, unable to require http.request method; reason: %w", err)
+	}
+	// Export the http module's syncRequest function goja.Callable as a Go function
+	var asyncRequestFunc HTTPAsyncRequestFunc
+	if err := rt.ExportTo(httpModuleObject.Get("asyncRequest"), &asyncRequestFunc); err != nil {
+		return nil,
+			fmt.Errorf("failed initializing tracing client, unable to require http.asyncRequest method; reason: %w",
+				err)
 	}
 
-	client := &Client{vu: vu, requestFunc: requestFunc}
+	client := &Client{vu: vu, requestFunc: requestFunc, asyncRequestFunc: asyncRequestFunc}
 	if err := client.Configure(opts); err != nil {
-		common.Throw(
-			rt,
-			fmt.Errorf("failed initializing tracing client, invalid configuration; reason: %w", err),
-		)
+		return nil,
+			fmt.Errorf("failed initializing tracing client, invalid configuration; reason: %w", err)
 	}
 
-	return client
+	return client, nil
 }
 
 // Configure configures the tracing client with the given options.
@@ -93,15 +99,7 @@ func (c *Client) Configure(opts options) error {
 	return nil
 }
 
-// Request instruments the http module's request function with tracing headers,
-// and ensures the trace_id is emitted as part of the output's data points metadata.
-func (c *Client) Request(method string, url goja.Value, args ...goja.Value) (*httpmodule.Response, error) {
-	// The http module's request function expects the first argument to be the
-	// request body. If no body is provided, we need to pass null to the function.
-	if len(args) == 0 {
-		args = []goja.Value{goja.Null()}
-	}
-
+func (c *Client) generateTraceContext() (http.Header, string, error) {
 	traceID := TraceID{
 		Prefix: k6Prefix,
 		Code:   k6CloudCode,
@@ -110,21 +108,64 @@ func (c *Client) Request(method string, url goja.Value, args ...goja.Value) (*ht
 
 	encodedTraceID, err := traceID.Encode()
 	if err != nil {
-		return nil, fmt.Errorf("failed to encode the generated trace ID; reason: %w", err)
+		return http.Header{}, "", fmt.Errorf("failed to encode the generated trace ID; reason: %w", err)
 	}
 
-	// Produce a trace header in the format defined by the
-	// configured propagator.
+	// Produce a trace header in the format defined by the configured propagator.
 	traceContextHeader, err := c.propagator.Propagate(encodedTraceID)
 	if err != nil {
-		return nil, fmt.Errorf("failed to propagate trace ID; reason: %w", err)
+		return http.Header{}, "", fmt.Errorf("failed to propagate trace ID; reason: %w", err)
 	}
 
+	return traceContextHeader, encodedTraceID, nil
+}
+
+// Request instruments the http module's request function with tracing headers,
+// and ensures the trace_id is emitted as part of the output's data points metadata.
+func (c *Client) Request(method string, url goja.Value, args ...goja.Value) (*httpmodule.Response, error) {
+	var result *httpmodule.Response
+	var err error
+	err = c.instrumentedCall(func(args ...goja.Value) error {
+		result, err = c.requestFunc(method, url, args...)
+		return err
+	}, args...)
+
+	if err != nil {
+		return nil, err
+	}
+	return result, nil
+}
+
+// AsyncRequest instruments the http module's asyncRequest function with tracing headers,
+// and ensures the trace_id is emitted as part of the output's data points metadata.
+func (c *Client) AsyncRequest(method string, url goja.Value, args ...goja.Value) (*goja.Promise, error) {
+	var result *goja.Promise
+	var err error
+	err = c.instrumentedCall(func(args ...goja.Value) error {
+		result, err = c.asyncRequestFunc(method, url, args...)
+		return err
+	}, args...)
+
+	if err != nil {
+		return nil, err
+	}
+	return result, nil
+}
+
+func (c *Client) instrumentedCall(call func(args ...goja.Value) error, args ...goja.Value) error {
+	if len(args) == 0 {
+		args = []goja.Value{goja.Null()}
+	}
+
+	traceContextHeader, encodedTraceID, err := c.generateTraceContext()
+	if err != nil {
+		return err
+	}
 	// update the `params` argument with the trace context header
 	// so that it can be used by the http module's request function.
 	args, err = c.instrumentArguments(traceContextHeader, args...)
 	if err != nil {
-		return nil, fmt.Errorf("failed to instrument request arguments; reason: %w", err)
+		return fmt.Errorf("failed to instrument request arguments; reason: %w", err)
 	}
 
 	// Add the trace ID to the VU's state, so that it can be
@@ -132,19 +173,14 @@ func (c *Client) Request(method string, url goja.Value, args ...goja.Value) (*ht
 	c.vu.State().Tags.Modify(func(t *metrics.TagsAndMeta) {
 		t.SetMetadata(metadataTraceIDKeyName, encodedTraceID)
 	})
+	// Remove the trace ID from the VU's state, so that it doesn't leak into other requests.
+	defer func() {
+		c.vu.State().Tags.Modify(func(t *metrics.TagsAndMeta) {
+			t.DeleteMetadata(metadataTraceIDKeyName)
+		})
+	}()
 
-	response, err := c.requestFunc(method, url, args...)
-	if err != nil {
-		return nil, err
-	}
-
-	// Remove the trace ID from the VU's state, so that it doesn't
-	// leak into other requests.
-	c.vu.State().Tags.Modify(func(t *metrics.TagsAndMeta) {
-		t.DeleteMetadata(metadataTraceIDKeyName)
-	})
-
-	return response, nil
+	return call(args...)
 }
 
 // Del instruments the http module's delete method.
--- a/js/modules/k6/http/async_request_test.go
+++ b/js/modules/k6/http/async_request_test.go
@@ -1,0 +1,290 @@
+package http
+
+import (
+	"testing"
+	"time"
+
+	"github.com/sirupsen/logrus"
+	logtest "github.com/sirupsen/logrus/hooks/test"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"go.k6.io/k6/js/modulestest"
+)
+
+func wrapInAsyncLambda(input string) string {
+	// This makes it possible to use `await` freely on the "top" level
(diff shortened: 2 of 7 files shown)


js/modules/k6/http/request.go
```diff
@@ -73,6 +73,61 @@ func (c *Client) Request(method string, url goja.Value, args ...goja.Value) (*Re
...
+	}
+
+	p, resolve, reject := rt.NewPromise()
+	callback := c.moduleInstance.vu.RegisterCallback()
+
+	go func() {
+		resp, err := httpext.MakeRequest(c.moduleInstance.vu.Context(), state, req)
+		callback(func() error {
+			if err != nil {
+				_ = reject
```
oleiade, line 119: I'm unclear as to what this statement does. Could you elaborate? Also, how does it relate to returning the error? 🙇🏻

js/modules/k6/http/request.go
```diff
@@ -73,6 +73,61 @@ func (c *Client) Request(method string, url goja.Value, args ...goja.Value) (*Re
...
+	}
+	if len(args) > 1 {
+		params = args[1]
+	}
+
+	rt := c.moduleInstance.vu.Runtime()
+	req, err := c.parseRequest(method, url, body, params)
+	if err != nil {
+		if state.Options.Throw.Bool {
+			return nil, err
```
oleiade, line 98: I must admit I'm not sure what happens at the Runtime level when you return a promise **and** an error to it. Does throwing the error take precedence over processing the promise?

This question came from me seeing anything else than a plain `goja.Promise` being returned and realizing that the code does create a promise just in time (later) and that before that, there's nothing to reject yet.

js/modules/k6/http/async_request_test.go
```diff
@@ -0,0 +1,259 @@
...
+}
+
+func TestAsyncRequest(t *testing.T) {
+	t.Parallel()
+	t.Run("EmptyBody", func(t *testing.T) {
+		t.Parallel()
+		ts := newTestCase(t)
+
+		sr := ts.tb.Replacer.Replace
+		err := runOnEventLoop(ts.runtime, sr(`
```
oleiade, line 36: non-blocking nit: the helpers to run on the event loop inside of an async lambda make total sense. However, I would have gained a few minutes wrapping my head around it if the statement were not wrapped. I'm suggesting we unroll the statement as:
```javascript
runOnEventLoop(ts.Runtime, wrapInAsyncLambda(sr(`
```

js/modules/k6/experimental/tracing/client.go
```diff
@@ -93,9 +99,7 @@ func (c *Client) Configure(opts options) error {
 	return nil
 }
 
-// Request instruments the http module's request function with tracing headers,
-// and ensures the trace_id is emitted as part of the output's data points metadata.
-func (c *Client) Request(method string, url goja.Value, args ...goja.Value) (*httpmodule.Response, error) {
+func (c *Client) addTraceIDToArgs(args []goja.Value) ([]goja.Value, error) {
```
oleiade, line 102: Good call on attempting to extract the logic!

However, I think the function now does too much, also some things unrelated to modifying the arguments in place (we already have the `instrumentArguments` function to do that specifically. 

In the spirit of factorizing and making this logic reusable across the two request functions, I'd prefer it if instead of this specific function, we would move the trace id generation logic to a dedicated method: `Client.generateTraceContext`, and keep the setting of metadata external to that. Something along the lines of:
```javascript
func (c *Client) generateTraceContext() (header http.Header, traceID string, err error) {
	traceID = TraceID{
		Prefix: k6Prefix,
		Code:   k6CloudCode,
		Time:   time.Now(),
	}

	encodedTraceID, err = traceID.Encode()
	if err != nil {
		return http.Header{}, TraceID{}, fmt.Errorf("failed to encode the generated trace ID; reason: %w", err)
	}

	// Produce a trace header in the format defined by the
	// configured propagator.
	traceContextHeader, err = c.propagator.Propagate(traceID)
	if err != nil {
		return http.Header{}, TraceID{}, fmt.Errorf("failed to propagate trace ID; reason: %w", err)
	}
	
	return traceContextHeader, traceID, nil
}

func (c *Client) Request(method string, url goja.Value, args ...goja.Value) (*httpmodule.Response, error) {
    traceHeader, traceID, err = c.generateTraceContext()
    if err != nil {
        return nil, err
    }

    args, err = c.instrumentArguments(traceHeader, args)
	if err != nil {
		return nil, err
	}
	
	// Add the trace ID to the VU's state, so that it can be
	// used in the metrics emitted by the HTTP module.
	c.vu.State().Tags.Modify(func(t *metrics.TagsAndMeta) {
		t.SetMetadata(metadataTraceIDKeyName, encodedTraceID)
	})
	
	// Remove the trace ID from the VU's state, so that it doesn't
	// leak into other requests.
	defer func() {
		c.vu.State().Tags.Modify(func(t *metrics.TagsAndMeta) {
			t.DeleteMetadata(metadataTraceIDKeyName)
		})
	}()
	
	return c.requestFunc(method, url, args...) 
}

// same structure for asyncRequest
```

🙇🏻

oleiade (changes requested):
Really great work! 👏🏻 🎉 

I've left a bunch of comments that I don't consider blocking, but that I care enough about that I'd like them to be considered. 

Overall I have tried to find test cases that could be missing, but after more careful consideration, I believe the scenarios I thought of are already covered by the current test suite, or are very tied to the usage of the `async/await` syntax itself, and are very likely already covered by goja's test suite. I ran a few test scripts on my side too and couldn't break it, so kudos! 👏🏻 

🙇🏻

js/modules/k6/http/request.go
mstoykov: The same as with anything else that is returned along an error - the  thing is ignored the error is thrown as exception. 

In this case this needs more tests - or more accurately needs test which differentiate between whether an exception is thrown or a promise is rejected. As *unhandled* promise rejection ends up stopping the execution either way it can be caught by the current tests as well. 

So in practice is a thing that needs changes to test that I will be doing next week.

js/modules/k6/http/request.go
mstoykov: Ah... good catch. 

The line just tells go to not care about the fact that I do not use `reject`. Which is as I wasn't sure whether I should be rejecting this promise or just returning an error. But thinking about it ... this will in practice always stop the event loop which is definitely not what should happen so I will have to make a test for this and actually check that I don't break the event loop ;) 

Connected to https://github.com/grafana/k6/pull/2877/files#r1089241147

good catch again :bow:

js/modules/k6/http/async_request_test.go
mstoykov: Hm ... I am not really certain I agree. If anything I am more for this going on top of the testCase type here and actually calling both `wrapInAsyncLambda` and `sr` automatically so you don't need to forget about it.

And then use that over the whole tests in `k6/http` . A version with `sr` should probably be extracted in `moduletest`

What in particular was not clear or confusing to you about this statement?

js/modules/k6/experimental/tracing/client.go
oleiade: I'm happy with that 😄 🚀 🙇🏻

js/modules/k6/http/request.go
oleiade: Hehe, I'm glad I helped catch a potential issue... somehow 😄 🙇🏻

js/modules/k6/http/async_request_test.go
oleiade: Fair enough, that was really just an idea I had, but I see your point, and I'm happy to go your way 🤝 🙇🏻

js/modules/k6/http/request.go
oleiade: I see, thanks a lot for the clarification 🙇🏻 

My take away is: errors have higher precedence over promise rejections, and an exception will be thrown, before rejecting the promise is even considered 🤝

mstoykov: I am not certain this is *correct* but I guess it kind of works.

As usual goja tries to map go code (and idioms) to javascript and it's idioms. Unfortunately there is no 1-to-1 mapping.

Go can have multiple returns, javascript can't. Javascript error handling is exception based, go is error/value based. 

So in the case of error being returned goja has 2 options:
1. tread it specially and convert it to exception so that it is more natural
2. return it as normal, in the case of multiple returns that is an array

So you can argue that returning an error has precedence but it is over everything as goja choose 1 which arguably is more ergonomic 99% of the cases. Even within the go ecosystem" do something wit hteh value even if you get an error" is *super* rare and I think only applise to io.Writer/Reader.  

And promises are ... just values, they are cool CS values that represent the result of an operation that can be chained with other values and operations, but still values ;).

js/modules/k6/http/request.go
mstoykov: fixed in https://github.com/grafana/k6/pull/2877/commits/9d6afbc5a2a9420d1cfacc63b953a21ebe23989a

oleiade (approved):
LGTM 👏🏻 🎉

js/modules/k6/http/request.go
```diff
@@ -73,6 +73,61 @@ func (c *Client) Request(method string, url goja.Value, args ...goja.Value) (*Re
 	return c.responseFromHTTPext(resp), nil
 }
 
+// Request makes an http request of the provided `method` and returns a corresponding response by
+// taking goja.Values as arguments
+func (c *Client) asyncRequest(method string, url goja.Value, args ...goja.Value) (*goja.Promise, error) {
```
imiric, line 78: Let's update the doc comment.

js/modules/k6/http/request.go
```diff
@@ -73,6 +73,61 @@ func (c *Client) Request(method string, url goja.Value, args ...goja.Value) (*Re
...
+		}
+		state.Logger.WithField("error", err).Warn("Request Failed")
+		r := httpext.NewResponse()
+		r.Error = err.Error()
+		var k6e httpext.K6Error
+		if errors.As(err, &k6e) {
+			r.ErrorCode = int(k6e.Code)
+		}
+		resolve(&Response{Response: r, client: c})
+		return p, nil
```
imiric, line 110: Most of this method is repeated in `Request()`. Can we DRY it a bit?

Calling `MakeRequest()` below might be tricky, but everything above seems like it could go in a reusable function, so that the only difference between `Request()` and `asyncRequest()` is the async part of it. Similar to what you did in `tracing.Client`.

js/modules/k6/http/async_request_test.go
```diff
@@ -0,0 +1,290 @@
...
+	})
+}
+
+func TestAsyncRequestResponseCallbackRace(t *testing.T) {
+	// This test is here only to tease out race conditions
+	t.Parallel()
+	ts := newTestCase(t)
+	err := ts.runtime.VU.Runtime().Set("q", func(f func()) {
+		rg := ts.runtime.EventLoop.RegisterCallback()
+		time.AfterFunc(time.Millisecond*5, func() {
```
imiric, line 94: This is a very confusing test, and given the very small duration here and the unreliable state of CI, it might lead to a lot of false positives, and be more trouble than it's worth.

Have you run into these race conditions locally and fixed them, and do you really think this test would be helpful?

js/modules/k6/http/async_request_test.go
```diff
@@ -0,0 +1,290 @@
...
+	t.Run("Unroutable", func(t *testing.T) {
+		t.Parallel()
+		ts := newTestCase(t)
+		err := runOnEventLoop(ts.runtime, `
+            try {
+                http.asyncRequest("GET", "http://sdafsgdhfjg/").catch((e) => globalThis.promiseRejected = e )
+            } catch (e) {
+                globalThis.exceptionThrown = e
+            }`)
+		expErr := "lookup sdafsgdhfjg"
```
imiric, line 281: I got a failure from this test the first time I ran it:

```
=== CONT  TestAsyncRequestErrors/Unroutable
    async_request_test.go:288:
                Error Trace:    /home/ivan/Projects/grafana/k6/js/modules/k6/http/async_request_test.go:288
                Error:          "Get \"http://sdafsgdhfjg/\": request timeout" does not contain "lookup sdafsgdhfjg"
                Test:           TestAsyncRequestErrors/Unroutable
--- FAIL: TestAsyncRequestErrors (0.00s)
```

But then couldn't reproduce it after that. :confused:

Do you have an idea why it might've happened that one time?

imiric (commented):
Nice work! :clap:

Mostly LGTM, just left some questions and minor suggestions.

js/modules/k6/experimental/tracing/client.go
```diff
@@ -110,41 +108,79 @@ func (c *Client) Request(method string, url goja.Value, args ...goja.Value) (*ht
...
+		return err
+	}, args...)
+
+	if err != nil {
+		return nil, err
+	}
+	return result, nil
+}
+
+func (c *Client) insrumentedCall(call func(args ...goja.Value) error, args ...goja.Value) error {
```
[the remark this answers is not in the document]
mstoykov: [04cb762](https://github.com/grafana/k6/pull/2877/commits/04cb76286fd021c120fb5c115cb1c9ed7d997ea8)

js/modules/k6/http/request.go
mstoykov: [0834272](https://github.com/grafana/k6/pull/2877/commits/0834272dec6963d5f4e6432d0250ccc75722936b)

js/modules/k6/http/async_request_test.go
mstoykov: I see absolutely no way in which this is possible :thinking: :( 

More importantly I don't see anyway in which the original not async test will not be hitting it as well. If anything it likely can hit it easier as it is reusing a bunch of stuff from other tests while this one gets a clean slate each time. 

I can't even imagine how you can get "request timeout", have a 0s test run and only this test to fail. Even if you go change something or have not checked out cleanly I would expect a bunch of other tests to get basically the same error. 

The only explanation is that something within net/http does something wrong and mistakes the error which seems ... very unlikely, but I have nothing else. 

WDYT? Is there anything that you can think off that you did or could've happened

js/modules/k6/http/request.go
mstoykov: In the tracing.Client the only different part was what was returned. 

I kind of don't want to touch `http.request` if possible as I prefer to keep it as unbroken as possible :sweat_smile: 

[2f40547](https://github.com/grafana/k6/pull/2877/commits/2f405474b192e9e13fb3b9a3caee74eb278f17d1) does a lot of the work IMO, but DRYing the rest IMO will just make it way more unreadable than anything else

js/modules/k6/http/async_request_test.go
imiric: Well, sure, but hopefully a test would replicate a scenario that could possibly fail, or that failed previously and was fixed, and not just model a hypothetical scenario that has little to no chance of happening in real world usage.

But if this failed at some point for you before and now it's fixed, then it's good to keep around. And if it ends up being flaky in the long run, we can consider removing or fixing it then.

js/modules/k6/http/async_request_test.go
imiric: Yeah, no idea either :confused:

I just ran `go test -v -race -failfast -count=1 -run 'TestAsyncRequest' ./js/modules/k6/http`, and then ran it again and it passed, and then with `-count=1000` and it kept passing until I stopped it.

I did clean my go mod cache recently, so maybe something was "warming up" that first time, but have no real explanation for it. Ah well, no need to dig into it if we can't reproduce it.

imiric (approved):
:+1:

Let's just squash+rebase this.

imiric (dismissed):

js/modules/k6/experimental/tracing/client.go
```diff
@@ -30,47 +29,54 @@ type Client struct {
...
-		common.Throw(
-			rt,
-			fmt.Errorf(
-				"failed initializing tracing client, "+
-					"unable to require http.request method; reason: %w",
-				err,
-			),
-		)
+		return nil,
+			fmt.Errorf("failed initializing tracing client, unable to require http.request method; reason: %w", err)
```
codebien, line 55:
```suggestion
			fmt.Errorf("failed initializing tracing client, unable to require k6/http module; reason: %w", err)
```
nit

oleiade (dismissed):

js/modules/k6/experimental/tracing/client.go
```diff
@@ -110,41 +108,79 @@ func (c *Client) Request(method string, url goja.Value, args ...goja.Value) (*ht
...
-	// leak into other requests.
-	c.vu.State().Tags.Modify(func(t *metrics.TagsAndMeta) {
-		t.DeleteMetadata(metadataTraceIDKeyName)
-	})
-
-	return response, nil
+	// Remove the trace ID from the VU's state, so that it doesn't leak into other requests.
+	defer func() {
+		c.vu.State().Tags.Modify(func(t *metrics.TagsAndMeta) {
+			t.DeleteMetadata(metadataTraceIDKeyName)
```
codebien, line 179: If I remove this line nothing fails between the HTTP and tracing module tests

mstoykov: I do think this should be addressed separately. After taking a look this is not new I just refactor the code - if anything my code is less likely to break :P. But writing a test for this seems a bit more involved than I would prefer for this PR

mstoykov: I did open an issue with it https://github.com/grafana/k6/issues/2889

codebien (approved):

imiric (approved):

PR #2816, opened 2022-12-09
[WIP] Proof of concept changes (test suites, HDR histograms)

na--:
This is a repository for unpolished PoC changes before they can be split off in their own PRs. The commits here won't be merged without significant refactoring and changes and will probably be larger than average, to simplify rebases. It is the second iteration of such a PoC PR, the original PoC for distributed execution was in https://github.com/grafana/k6/pull/2438, see https://github.com/grafana/k6/pull/2438#issuecomment-1344034591 for details.

As of right now, this PR contains rudimentary, hackish and very janky implementations of test suites (https://github.com/grafana/k6/issues/1342) and HDR histograms (https://github.com/grafana/k6/issues/763) :tada: It used to contain a distributed execution (https://github.com/grafana/k6/issues/140) PoC too, but that got split off into separate PRs, see https://github.com/grafana/k6/pull/2816#issuecomment-1642271288.

PR #2878, opened 2023-01-26
Metrics engine tests

codebien:
https://github.com/grafana/k6/pull/2815#discussion_r1056459921

>// TODO: refactor and move the tests that still make sense in other packages

It resolves the unit tests parts

### DIFF
--- a/metrics/engine/engine.go
+++ b/metrics/engine/engine.go
@@ -25,9 +25,8 @@ const thresholdsRate = 2 * time.Second
 // aggregated metric sample values. They are used to generate the end-of-test
 // summary and to evaluate the test thresholds.
 type MetricsEngine struct {
-	es     *lib.ExecutionState
-	logger logrus.FieldLogger
-
+	logger         logrus.FieldLogger
+	test           *lib.TestRunState
 	outputIngester *outputIngester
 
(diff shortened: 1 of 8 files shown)


metrics/engine/engine.go
```diff
@@ -21,13 +21,21 @@ import (
 
 const thresholdsRate = 2 * time.Second
 
+type testRunState struct {
```
na--, line 24: Hmm why do we need a custom type for this? Can't we just add these 2 fields to the `MetricsEngine` directly? :thinking:

na--: btw I remembered that I did a somewhat similar refactoring to the `MetricsEngine` as a part of the PoCs in https://github.com/grafana/k6/pull/2816 :thinking: There I pass `getCurrentTestRunDuration func() time.Duration` as an argument to `evaluateThresholds()`. Maybe it's worth to backport some of these changes here, take a look at [this diff](https://github.com/grafana/k6/pull/2816/files#diff-802af5adfe429d78b1b72096ac7b586c28622dffa6a34024a7e91f5d75c1d077)

na-- (approved):

na-- (approved):
:shipit:

Issue #794, opened 2018-10-02
Remove/improve the per-VU buffer pool?


na--:
Currently each VU has [its own buffer pool](https://github.com/loadimpact/k6/blob/c665a73e561bcef243b40207640d8099d392f488/js/runner.go#L183) that's used for [storing HTTP response bodies](https://github.com/loadimpact/k6/blob/c665a73e561bcef243b40207640d8099d392f488/js/modules/k6/http/http_request.go#L537-L539). The current approach has a few limitations:
- Each buffer has a fixed size of 100 elements, which seems very arbitrary...
- Each VU has its own buffer, but each buffer has to be concurrent because of things like `http.batch()`, so why not simply have a global one?
- We use the [github.com/oxtoacart/bpool](https://github.com/oxtoacart/bpool) library for the buffer pools, which doesn't seem very alive. Though it looks simple and self-contained, so it's unlikely that any maintenance is needed... The bigger issue though is that these static buffer pools don't get de-allocated when the Go runtime needs space and the buffers live for the entire lifetime of a k6 script, which may not be what we want when we have hundreds of VUs.

Like I mention in [this TODO](https://github.com/loadimpact/k6/commit/72246ab2ecb69d43b75fe335c208a797afb15276#diff-c035263c8331ec178fe2d3fdd9082e70R60), using something like `sync.Pool` may be a better idea, pending some benchmarks of course. `sync.Pool` has dynamic resizes and allows objects to be automatically garbage collected when needed, which seems useful in our case. More details [here](https://blog.questionable.services/article/using-buffer-pools-with-go/#postscript) and [here](https://github.com/oxtoacart/bpool/issues/5)

na--:
This needs more investigation, but as part of my current fixes of https://github.com/loadimpact/k6/issues/1044 and https://github.com/loadimpact/k6/issues/1041, I decided to benchmark a few things. Specifically, how the removal of the current BPools or their replacement with `sync.Pool` (either a global one or a per-VU one) would affect the k6 performance in a script that's biased towards big response bodies. 

Here's the k6 script:
```js
import http from "k6/http";

export let options = {
    vus: 50,
    iterations: 10000,
    batchPerHost: 25,
};

var testServer = "http://127.0.0.1:8080"

export default function () {
    http.get(`${testServer}/`);
    http.post(`${testServer}/post`);
    http.batch([
        ["GET", `${testServer}/gzip`],
        ["GET", `${testServer}/gzip`],
[...]

PR #2879, opened 2023-01-27
Improve the per-VU buffer pool

davidpst:
This is an implementation of a shared buffer pool for all the VUs using `sync.Pool`. This implementation aims to improve memory consumption that has been commented here #794


na--:
The `deps` CI check probably fails because we no longer use `github.com/oxtoacart/bpool`, so you need to run something like `go mod tidy` and `go mod vendor` and commit the changes

davidpst:
You are right, done!

js/runner.go
```diff
@@ -54,8 +53,9 @@ type Runner struct {
 	RPSLimit       *rate.Limiter
 	RunTags        *metrics.TagSet
 
-	console   *console
-	setupData []byte
+	console    *console
+	setupData  []byte
+	BufferPool *lib.BufferPool
```
na--, line 58: Why did you add a `BufferPool` to the `js.Runner`?

davidpst: I added it to reuse same pool for all VUs. I suppose ´js.Runner´ instance is unique for a load test execution.

na--: Yes, but is that actually better for performance than a separate `BufferPool` per VU? In the original simplistic checks I did for https://github.com/grafana/k6/issues/794, a global `sync.Pool` actually performed a lot worse than a per-VU `sync.Pool`. 

Maybe that has changed in recent Go and k6 versions, but we'd need to see a new benchmark comparison between both to justify the decision to go with a global pool. And, as a small nitpick, even if the the global pool is performing better now, it should still probably be unexported.

davidpst: I ran the test ([BenchmarkReadResponseBody](https://github.com/grafana/k6/blob/163ceb55a05ee9e8878cec81249bdf775cb21e97/cmd/tests/cmd_run_test.go#L1648)) with both configurations and to have a global pool improves memory and time execution. First execution (with global pool) allocate less memory and can iterate more times.

I think the result make sense because we reuse more the pool objects, so we don't have to allocate memory neither spend time creating new objects.

`
Global pool for all VUs

BenchmarkReadResponseBody-10               1    30455673750 ns/op   138908381552 B/op   69924811 allocs/op


     checks.........................: 100.00% ✓ 222192      ✗ 0     
     data_received..................: 15 GB   507 MB/s
     data_sent......................: 21 MB   689 kB/s
     http_req_blocked...............: avg=16.18µs  min=0s       med=1µs      max=75.67ms  p(90)=2µs      p(95)=3µs    
     http_req_connecting............: avg=5.29µs   min=0s       med=0s       max=31.02ms  p(90)=0s       p(95)=0s     
     http_req_duration..............: avg=2.31ms   min=74µs     med=1.14ms   max=93.67ms  p(90)=5.53ms   p(95)=8.32ms 
       { expected_response:true }...: avg=2.31ms   min=74µs     med=1.14ms   max=93.67ms  p(90)=5.53ms   p(95)=8.32ms 
     http_req_failed................: 0.00%   ✓ 0           ✗ 222192
     http_req_receiving.............: avg=420.75µs min=9µs      med=34µs     max=56.95ms  p(90)=938µs    p(95)=1.88ms 
     http_req_sending...............: avg=27.8µs   min=1µs      med=3µs      max=69.87ms  p(90)=15µs     p(95)=43µs   
     http_req_tls_handshaking.......: avg=0s       min=0s       med=0s       max=0s       p(90)=0s       p(95)=0s     
     http_req_waiting...............: avg=1.86ms   min=60µs     med=860µs    max=93.54ms  p(90)=4.62ms   p(95)=6.87ms 
     http_reqs......................: 222192  7378.159847/s
     iteration_duration.............: avg=108.23ms min=101.33ms med=105.72ms max=204.63ms p(90)=117.13ms p(95)=121.4ms
     iterations.....................: 27774   922.269981/s
     vus............................: 100     min=100       max=100 
     vus_max........................: 100     min=100       max=100 
     
Individual pool for VUs

BenchmarkReadResponseBody-10               1    30450333375 ns/op   174401310944 B/op   73585506 allocs/op

     checks.........................: 100.00% ✓ 217008      ✗ 0     
     data_received..................: 15 GB   495 MB/s
     data_sent......................: 20 MB   673 kB/s
     http_req_blocked...............: avg=27µs     min=0s       med=1µs      max=44.97ms  p(90)=2µs     p(95)=3µs     
     http_req_connecting............: avg=14.71µs  min=0s       med=0s       max=42.62ms  p(90)=0s      p(95)=0s      
     http_req_duration..............: avg=2.99ms   min=75µs     med=1.36ms   max=106.88ms p(90)=7.15ms  p(95)=10.76ms 
       { expected_response:true }...: avg=2.99ms   min=75µs     med=1.36ms   max=106.88ms p(90)=7.15ms  p(95)=10.76ms 
     http_req_failed................: 0.00%   ✓ 0           ✗ 217008
     http_req_receiving.............: avg=590.9µs  min=10µs     med=59µs     max=71.81ms  p(90)=1.14ms  p(95)=2.56ms  
     http_req_sending...............: avg=34.37µs  min=1µs      med=3µs      max=64.95ms  p(90)=15µs    p(95)=45µs    
     http_req_tls_handshaking.......: avg=0s       min=0s       med=0s       max=0s       p(90)=0s      p(95)=0s      
     http_req_waiting...............: avg=2.37ms   min=61µs     med=1.02ms   max=92.94ms  p(90)=5.8ms   p(95)=8.92ms  
     http_reqs......................: 217008  7209.308715/s
     iteration_duration.............: avg=110.64ms min=101.37ms med=107.07ms max=227.74ms p(90)=121.7ms p(95)=129.85ms
     iterations.....................: 27126   901.163589/s
     vus............................: 100     min=100       max=100 
     vus_max........................: 100     min=100       max=100 
`

olegbespalov (dismissed):
Changes look good to me, thanks :+1: 

I am postponing the final decision about where the buffer sync pool should be to Ned since I feel that he has a better context.

na--:
OK, so I finally had time to re-run my original benchmarks from https://github.com/grafana/k6/issues/794#issuecomment-500331683, with these 4 configurations:
1. [[results]](https://gist.github.com/na--/3f2b663bedaab47b2fd4b2fbbb76c7e8#file-results-k6-master-txt) the current k6 `master`, as of https://github.com/grafana/k6/commit/0b1fe61c5762c83c52a2378f6cdf9c0c3ff1f53d
2. [[results]](https://gist.github.com/na--/3f2b663bedaab47b2fd4b2fbbb76c7e8#file-results-global-sync-pools-txt) this PR with the same `master` branch from 1. merged into it
3. [[results]](https://gist.github.com/na--/3f2b663bedaab47b2fd4b2fbbb76c7e8#file-results-local-sync-pools-txt) the code from 2. but with [this patch](https://gist.github.com/na--/3f2b663bedaab47b2fd4b2fbbb76c7e8#file-per-vu-sync-pools-patch) to make the `sync.Pool` values per-VU
4. [[results]](https://gist.github.com/na--/3f2b663bedaab47b2fd4b2fbbb76c7e8#file-results-no-buffer-pools-txt) no buffer pools (i.e. just `buf := &bytes.Buffer{}` in `readResponseBody()` :scream: ), for a control

Some details about how I ran all of the tests:
- [this is the test script](https://gist.github.com/na--/3f2b663bedaab47b2fd4b2fbbb76c7e8#file-test-js) I used
- I ran each test 3 times with `/usr/bin/time`, so we have both the external process CPU and memory usage metrics and the k6 end-of-test summary
- all of the tests were running against the [same external server process running this simple code](https://gist.github.com/na--/3f2b663bedaab47b2fd4b2fbbb76c7e8#file-server-go)
- both k6 and the server were compiled with Go 1.20.1
- I've tried to include anything relevant for reproducing my results in this gist: https://gist.github.com/na--/3f2b663bedaab47b2fd4b2fbbb76c7e8

As you can see, the test has a few significant differences with the one you added to the benchmarks. It is much, much larger in terms of VUs and iterations, but it also expresses the work in terms of test iterations, not in terms of test duration. So, assuming the server is roughly constant, a more performant k6 implementation will be done with the script sooner. And having the server outside of the measured k6 process helps us to (mostly) isolate the benchmark test to just the k6 performance.

So, it seems like the local per-VU `sync.Pool` variant (the 3rd test case) performs the worst in all categories except the "no buffer" control (4th test case) - most memory and CPU usage for the least performance. This is somewhat surprising to me, but it is consistent with both your benchmarks and my original ones from a few years ago that I [put in the issue](https://github.com/grafana/k6/issues/794#issuecomment-500331683).

However, things are more complicated when we compare 1. and 2 :confused: A global `sync.Pool` indeed has a significantly lower memory usage! :tada:  However, it also has a noticeably higher CPU usage and lower throughput/iteration rate :disappointed:  To be fair, this benchmark is not a very realistic real-life k6 use case. It just shows that the changes in this PR are not only positive, they also have some performance costs... :disappointed: 

Personally, I would be fine with merging the PR as it is, since having ~2.5x memory reduction is probably worth the ~10% CPU increase/throughput decrease (depending on what we measure and compare). We can just leave https://github.com/grafana/k6/issues/794 open and investigate some of the other potential improvements from [this comment](https://github.com/grafana/k6/issues/794#issuecomment-1406338293) (or others) at some later point? Alternatively, we can try to see if some further simple improvement won't push things enough so there are no downsides to merging the PR?

@grafana/k6-core, what do you think? @imiric, you recently did https://github.com/grafana/k6-docs/pull/992 and we now have https://github.com/grafana/k6-benchmarks, maybe it's worth it to run the benchmarks with some of the same variants I tested here? FWIW, I probably won't be able to dedicate time for a more thorough investigation here further benchmarks for at least the next 2 weeks :disappointed:

davidpst:
> Personally, I would be fine with merging the PR as it is, since having ~2.5x memory reduction is probably worth the ~10% CPU increase/throughput decrease (depending on what we measure and compare). We can just leave https://github.com/grafana/k6/issues/794 open and investigate some of the other potential improvements from https://github.com/grafana/k6/issues/794#issuecomment-1406338293 (or others) at some later point? Alternatively, we can try to see if some further simple improvement won't push things enough so there are no downsides to merging the PR?

Yes, now GC has to run to remove those buffers so it consume some cpu when run, you can see it using pprof. In any case the memory optimization is bigger than the cpu consumption.

> As you can see, the test has a few significant differences with the one you added to the benchmarks. It is much, much larger in terms of VUs and iterations, but it also expresses the work in terms of test iterations, not in terms of test duration. So, assuming the server is roughly constant, a more performant k6 implementation will be done with the script sooner. And having the server outside of the measured k6 process helps us to (mostly) isolate the benchmark test to just the k6 performance.

Sure, I committed the test with a small number of VUs and Iterations to avoid largest builds, but I ran it with a bigger number of course and the result was the same, memory optimization and the increase of GC related with this kind of pool.

imiric (commented):
Nice work @davidpst! :clap: I just have one minor comment.

I ran the script from [k6-benchmarks](https://github.com/grafana/k6-benchmarks) on `master` (367163be3) and this branch (b4a0314fb) on my local machine, and noticed the same behavior you guys did. This branch has much lower memory usage, at the slight cost of higher CPU usage and lower RPS.

See the results [here](https://gist.github.com/imiric/f3aa24a33956a880cda2462099a4431b). The graphs are a bit wonky, since the RAM line is squashed due to the higher order of RPS, but the effect is still visible. Otherwise, GitHub renders CSVs nicely in tables. :)

Question: would it be a good idea to make the new `BufferPool` the default implementation, but give the choice to the user to fallback to the previous behavior (via an env var), if they decide they'd rather not sacrifice CPU/RPS? I know we'd like to abandon the `bpool` dependency, but this approach might be safer in some circumstances.

This still requires more thorough testing, like @na-- says, but I don't think these additional benchmarks should block this PR from merging. We've seen that the memory improvement is substantial, and we can continue testing this on `master`.

lib/buffer_pool.go
```diff
@@ -0,0 +1,40 @@
...
+		},
+	}
+}
+
+// Get return a bytes.Buffer from the pool
+func (bp BufferPool) Get() *bytes.Buffer {
+	b, ok := bp.pool.Get().(*bytes.Buffer)
+	if !ok {
+		return bytes.NewBuffer([]byte{})
+	}
```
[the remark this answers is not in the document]
imiric: Yeah, I think it's safe, so we can just return the type asserted result directly.

This kind of makes having our own `sync.Pool` wrapper unnecessary, though I suppose there's still some value in avoiding type assertions, and making sure we call `b.Reset()` wherever it's used.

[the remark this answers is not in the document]
davidpst: Done! yes I think the wrapper is important to ensure `b.Reset()` is call before return the buffer.

imiric:
[Here are the benchmark results](https://github.com/imiric/k6/commit/7218c61fce2c9fc4274e3ce251f0bd2414d5f850) when using https://github.com/valyala/bytebufferpool.

It's a negligible improvement in memory usage, practically within the margin of error, but has an even bigger CPU and RPS penalty than plain `sync.Pool`.

I also tried https://github.com/libp2p/go-buffer-pool, which does something similar to what is mentioned in [this comment](https://github.com/grafana/k6/issues/794#issuecomment-1406338293). It manages a pool of `sync.Pool`s, to separate buffers into buckets of similar size. The issue is that it requires knowing the buffer size upfront, which doesn't work reliably for us, since we have to depend on the `Content-Length` header, which may be wrong or missing. See the [benchmark results](https://github.com/imiric/k6/commit/146ff04b9bc67486c187a2ca5199693184336848).

It's slightly better than bytebufferpool and a single `sync.Pool`, but I would also chalk it up to being within the margin of error.

So I don't see any conclusive improvements with these minor changes. Maybe we'll need a custom buffer implementation, like @na-- suggested, but that will have to wait for another time, or someone else to pick it up.

na-- (approved):
Let's :shipit: then :man_shrugging: we can keep the issue open, for now, since we can probably claw back some of that CPU inefficiency with a bit of effort

imiric (approved):
Sorry, forgot to re-approve this.

Thanks for your work @davidpst! :bow:

PR #2881, opened 2023-01-27
js: refactor how modules are loaded

mstoykov:
This refactor tries to simplify the implementation of `require` and
connected code by splitting it heavily into different parts.

These changes are similar to what will be needed for native ESM support
as shown in https://github.com/grafana/k6/pull/2563 , but without any of
the native parts and without doing anything that isn't currently needed.
This will hopefully make ESM PR much smaller and less intrusive.

This includes still keeping the wrong relativity of `require` as
explained in https://github.com/grafana/k6/issues/2674.

It also tries to simplify connected code, but due to this being very
sensitive code and the changes already being quite big, this is done
only to an extent.

The lack of new tests is mostly due to there not being really any new
code and the tests that were created along these changes already being
merged months ago with https://github.com/grafana/k6/pull/2782.

Future changes will try to address the above as well as potentially
moving the whole module types and logic in separate package to be reused
in tests.

### DIFF
--- a/js/bundle.go
+++ b/js/bundle.go
@@ -29,7 +29,6 @@ type Bundle struct {
 	Filename *url.URL
 	Source   []byte
 	Options  lib.Options
-	logger   *logrus.Logger
 
 	CompatibilityMode lib.CompatibilityMode // parsed value
 	preInitState      *lib.TestPreInitState
@@ -87,7 +86,6 @@ func newBundle(
 		callableExports:   make(map[string]struct{}),
 		filesystems:       filesystems,
 		pwd:               loader.Dir(src.URL),
-		logger:            piState.Logger,
 		preInitState:      piState,
 	}
 	c := bundle.newCompiler(piState.Logger)
@@ -261,13 +259,13 @@ func (b *Bundle) newCompiler(logger logrus.FieldLogger) *compiler.Compiler {
 
 func (b *Bundle) instantiate(vuImpl *moduleVUImpl, vuID uint64) (moduleInstance, error) {
 	rt := vuImpl.runtime
-	err := b.setupJSRuntime(rt, int64(vuID), b.logger)
+	err := b.setupJSRuntime(rt, int64(vuID), b.preInitState.Logger)
 	if err != nil {
 		return nil, err
 	}
 
 	initenv := &common.InitEnvironment{
-		Logger:      b.logger,
+		Logger:      b.preInitState.Logger,
 		FileSystems: b.filesystems,
 		CWD:         b.pwd,
 		Registry:    b.preInitState.Registry,
@@ -295,7 +293,7 @@ func (b *Bundle) instantiate(vuImpl *moduleVUImpl, vuID uint64) (moduleInstance,
 	}()
 
 	var instance moduleInstance
-	err = common.RunWithPanicCatching(b.logger, rt, func() error {
+	err = common.RunWithPanicCatching(b.preInitState.Logger, rt, func() error {
 		return vuImpl.eventLoop.Start(func() error {
 			//nolint:shadow,govet // here we shadow err on purpose
 			mod, err := b.moduleResolver.resolve(b.pwd, b.Filename.String())
@@ -394,12 +392,12 @@ func (b *Bundle) setInitGlobals(rt *goja.Runtime, modSys *moduleSystem) (unset f
 func generateCJSLoad(b *Bundle, c *compiler.Compiler) cjsModuleLoader {
 	return func(specifier *url.URL, name string) (*cjsModule, error) {
 		if filepath.IsAbs(name) && runtime.GOOS == "windows" {
-			b.logger.Warnf("'%s' was imported with an absolute path - this won't be cross-platform and won't work if"+
-				" you move the script between machines or run it with `k6 cloud`; if absolute paths are required,"+
-				" import them with the `file://` schema for slightly better compatibility",
+			b.preInitState.Logger.Warnf("'%s' was imported with an absolute path - this won't be cross-platform and "+
+				"won't work if you move the script between machines or run it with `k6 cloud`; if absolute paths are "+
+				"required, import them with the `file://` schema for slightly better compatibility",
 				name)
 		}
-		d, err := loader.Load(b.logger, b.filesystems, specifier, name)
+		d, err := loader.Load(b.preInitState.Logger, b.filesystems, specifier, name)
 		if err != nil {
 			return nil, err
 		}
--- a/js/bundle.go
+++ b/js/bundle.go
@@ -101,7 +101,7 @@ func newBundle(
 	// TODO use a real context
 	vuImpl := &moduleVUImpl{ctx: context.Background(), runtime: goja.New()}
 	vuImpl.eventLoop = eventloop.New(vuImpl)
-	instance, err := bundle.instantiate(vuImpl, 0, c)
+	instance, err := bundle.instantiate(vuImpl, 0)
 	if err != nil {
 		return nil, err
 	}
@@ -211,7 +211,7 @@ func (b *Bundle) Instantiate(ctx context.Context, vuID uint64) (*BundleInstance,
 	// runtime, but no state, to allow module-provided types to function within the init context.
 	vuImpl := &moduleVUImpl{ctx: ctx, runtime: goja.New()}
 	vuImpl.eventLoop = eventloop.New(vuImpl)
-	instance, err := b.instantiate(vuImpl, vuID, b.newCompiler(b.logger))
+	instance, err := b.instantiate(vuImpl, vuID)
 	if err != nil {
 		return nil, err
 	}
@@ -259,7 +259,7 @@ func (b *Bundle) newCompiler(logger logrus.FieldLogger) *compiler.Compiler {
 	return c
 }
 
-func (b *Bundle) instantiate(vuImpl *moduleVUImpl, vuID uint64, c *compiler.Compiler) (moduleInstance, error) {
+func (b *Bundle) instantiate(vuImpl *moduleVUImpl, vuID uint64) (moduleInstance, error) {
 	rt := vuImpl.runtime
 	err := b.setupJSRuntime(rt, int64(vuID), b.logger)
 	if err != nil {
--- a/js/bundle.go
+++ b/js/bundle.go
@@ -38,7 +38,7 @@ type Bundle struct {
 	pwd         *url.URL
 
 	callableExports map[string]struct{}
-	modResolution   *modulesResolution
+	moduleResolver  *moduleResolver
 }
 
 // A BundleInstance is a self-contained instance of a Bundle.
@@ -85,14 +85,14 @@ func newBundle(
 		Options:           options,
 		CompatibilityMode: compatMode,
 		callableExports:   make(map[string]struct{}),
-		modResolution:     newModuleResolution(getJSModules()),
+		moduleResolver:    newModuleResolution(getJSModules()),
 		filesystems:       filesystems,
 		pwd:               loader.Dir(src.URL),
 		logger:            piState.Logger,
 		preInitState:      piState,
 	}
 	c := bundle.newCompiler(piState.Logger)
-	if err = bundle.modResolution.setMain(src, c); err != nil {
+	if err = bundle.moduleResolver.setMain(src, c); err != nil {
 		return nil, err
 	}
 	// Instantiate the bundle into a new VM using a bound init context. This uses a context with a
@@ -274,7 +274,7 @@ func (b *Bundle) instantiate(vuImpl *moduleVUImpl, vuID uint64, c *compiler.Comp
 	}
 
 	cjsLoad := generateCJSLoad(b, c)
-	modSys := newModuleSystem(b.modResolution, vuImpl, cjsLoad)
+	modSys := newModuleSystem(b.moduleResolver, vuImpl, cjsLoad)
 	unbindInit := b.setInitGlobals(rt, modSys)
 	vuImpl.initEnv = initenv
 	defer func() {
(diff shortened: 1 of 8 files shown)


js/initcontext.go
```diff
@@ -288,13 +71,68 @@ func readFile(fileSystem afero.Fs, filename string) (data []byte, err error) {
...
+	// This is *not* the actual require behaviour defined in commonJS as it is actually always relative
+	// to the file it is in. This is unlikely to be an issue but this code is here to keep backwards
+	// compatibility *for now*.
+	// With native ESM this won't even be possible as `require` might not be called - instead an import
+	// might be used in which case we won't be able to be doing this hack. In that case we either will
+	// need some goja specific helper or to use stack traces as goja_nodejs does.
+	currentPWD := r.pwd
+	if specifier != "k6" && !strings.HasPrefix(specifier, "k6/") {
+		defer func() {
+			r.pwd = currentPWD
```
imiric, line 103: `currentPWD` is redundant, so maybe just `currWD`. I would also rename all uses of `pwd` to `cwd`, but that would touch many places (`loader`, tests, etc.), so let's leave that as is.

js/initcontext.go
```diff
@@ -288,13 +71,68 @@ func readFile(fileSystem afero.Fs, filename string) (data []byte, err error) {
...
+		}()
+		// In theory we can give that downwards, but this makes the code more tightly coupled
+		// plus as explained above this will be removed in the future so the code reflects more
+		// closely what will be needed then
+		fileURL, err := loader.Resolve(r.pwd, specifier)
+		if err != nil {
+			return nil, err
+		}
+		r.pwd = loader.Dir(fileURL)
```
imiric, line 112: Help me understand this hack... :sweat_smile: I read through and understand #2674, just not why we're doing _this_.

So you store the current `requireImpl.pwd` so you can revert it in the `defer`, but why do we need to change it here? The `return` below calls `r.modules.Require()` with `currentPWD`, so it doesn't seem the new WD is used(?).

In general, I'm not clear why we have a global reference to a working directory at all that needs to be changed. Couldn't each imported module have its own `pwd`, and we can resolve the paths to modules they import using the parent module's `pwd`? :confused: This way we could avoid keeping and reverting global state.

But I don't have the full picture here, so would appreciate some explanation.

js/initcontext.go
```diff
@@ -1,254 +1,37 @@
 package js
```
imiric, line 1: There's not much left in this file anymore, besides some helpers. Can we move those to where they make sense and remove this file? Same for `initcontext_test.go`.

We still refer to `InitContext` throughout the codebase, and we might want to rename some of those as well. Though I suppose we can still refer to the "init context" without having a direct implementation of it.

imiric (commented):
I like the removal of `InitContext` and separating it into dedicated and smaller implementations. :+1:

But I'm not clear about many things here yet, and will make another pass of this next week. I left some questions and minor suggestions in the meantime.

js/initcontext.go
mstoykov: > Couldn't each imported module have its own pwd

No because that will be the breaking change that I describe in #2674 ;) 

Or at least we could, and I would argue we *should*, but not in this PR.

This code as well as practically all the module resolving and loading is just broken down the previous code. 

You can see more or less this exact code in the previous version:
https://github.com/grafana/k6/blob/b750e206859a495b8a1fe12aa016dc8ed0e1c9db/js/initcontext.go#L186-L187

This code specifically does one thing and that is to make the `open` and `require` relative to the file that is currently being evaluated within `require`. Which is *different* from the file it is in which is basically what you are proposing and how `require` is supposed to work under commonJS.

How this currently works depend on where it is called from and when and not on just where it is written, Example:

file `/path/to/B/B.js`: 
```javascript
exports.load = func(s) {
  //do somethign cool 
  require(s); // by commonjs standard this is *always* relative to /path/to/B/B.js
  // but in our case it is relative to whatever 
  // do somethign else cool
}
```
we also have file `/path/to/C/C.js` which contains are not relevant

file /path/to/A.js has
```javascript
let s = require("./B/B.js"); // I will be using `require` instead of ESM for the purpose of being simpler ;) 
s.load("./C/C.js"); // by the standard this will require `/path/to/B/C/C.js` but k6 has always inturpretted this as `/path/to/C/C.js`
```

This  is likely more heavily used with `open` where you might want to do something in particular. 

So in order for this to work `require` (and `open`) need to  figure out which is the currently being "required" module, and is there could be cycles, and 1 module can be required multiple times I don't think there is another way to do it that isn't stack based which is how I would propose we do it in a future PR. 

If you want to try at it or have a specific proposal - I am all ears. But this is kept this way specifically for this to not be  a breaking change and so it does have minimal chance of breaking anything - as it works more or less exactly as before.

I do expect another problem is that ... the way this works in k6 is *confusing* and almost the exact way you expect it to work or should work by how the specification is written. 

Functions similar to `open` for example seem to always work relative to the *root* script/module and not based on the file it is called in.

Also with connected to https://github.com/grafana/k6/pull/2881#discussion_r1095753228 : this is the code of the call of `require` but this change to the pwd is for the `require` or `open` calls *after* this one. This `require` call uses the `currentPWD` *specifically* so that it can know what *it* is relative to.

js/initcontext.go
mstoykov: `currentPWD` is what is used for the current require the one that is being called at the moment.
`r.pwd` is set to the value it will need when `require` (or `open`) is called *within* the thing that we will require with this one.

js/initcontext.go
mstoykov: I kind of prefer not to make this bigger.

I will likely do more PRs (soon :tm:) around this code. Some of it will *hopefully* rip all the modules related parts and put it in a single place to be used in this package as well as in the tc39 test and even for the other *actual* module tests ;) 

I also *specifically* left the `open` and `require` implementation in it as ... that is what is currently in the initcontext . I could probably move some of the other helpers around but IMO that will just make the PR harder to review :shrug:

mstoykov:
I have also opened https://github.com/grafana/k6/pull/2900 for the stack based implementation of 
1. the current hacky way pwd is calculated
2. the way it should be if we actually adhere to commonjs

That PR is WIP and is only for ... illustrative purposes at this point :)

js/initcontext.go
imiric: I get what it's used for :)

My point is that `pwd` really means `cwd`, and "current current working directory" is redundant. But it's not a big deal.

js/initcontext.go
imiric: I see, thanks for explaining.

I see now that this is what we were doing before, so agree with not changing the behavior in this PR.

I don't have a specific proposal in mind, so let's keep this as is.

js/initcontext.go
imiric: I don't think this change would make things much harder to review, but alright, it's fine as is.

js/modules.go
```diff
@@ -0,0 +1,122 @@
...
+		mod, err := loadCJS(specifier, arg)
+		mr.cache[specifier.String()] = moduleCacheElement{mod: mod, err: err}
+		return mod, err
+	}
+}
+
+type moduleSystem struct {
+	vu            modules.VU
+	instanceCache map[module]moduleInstance
+	resolution    *modulesResolution
```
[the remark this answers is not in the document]
mstoykov: fixed in [d8fc49b](https://github.com/grafana/k6/pull/2881/commits/d8fc49bed400bfabed14307e83dc169f42fc9e45)

js/modules.go
```diff
@@ -0,0 +1,122 @@
...
+type moduleInstance interface {
+	execute() error
+	exports() *goja.Object
+}
+type moduleCacheElement struct {
+	mod module
+	err error
+}
+
+type modulesResolution struct {
```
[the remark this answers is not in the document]
mstoykov: fixed in [d8fc49b](https://github.com/grafana/k6/pull/2881/commits/d8fc49bed400bfabed14307e83dc169f42fc9e45)

imiric (dismissed):

js/gomodule.go
```diff
@@ -0,0 +1,93 @@
...
+var _ module = &baseGoModule{}
+
+func (b *baseGoModule) Instantiate(vu modules.VU) moduleInstance {
+	return &baseGoModuleInstance{mod: b.mod, vu: vu}
+}
+
+type baseGoModuleInstance struct {
+	mod      interface{}
+	vu       modules.VU
+	exportsO *goja.Object // this is so we only initialize the exports once per instance
```
na--, line 23: You have a `moduleObj` in the other `module` implementation, so this should probably be `exportsObj`. Or, better yet, remove the `O` and `Obj` from both and just have `module` and `exports`?

mstoykov: It can't be `exports` as `exports` is a method on the interface. 

I guess I can make the interface method upperCase as the intention is that this will be ripped out in a separate package in the future :thinking:

na--: :thinking: ignore this then, it's a minor nitpick all things considered

js/modules.go
```diff
@@ -0,0 +1,122 @@
...
+}
+
+type moduleSystem struct {
+	vu            modules.VU
+	instanceCache map[module]moduleInstance
+	resolver      *moduleResolver
+	cjsLoad       cjsModuleLoader
+}
+
+func newModuleSystem(resolution *moduleResolver, vu modules.VU, cjsLoad cjsModuleLoader) *moduleSystem {
```
na--, line 96: Shouldn't the `cjsModuleLoader` be part of the `moduleResolver` and not the `moduleSystem`? :thinking: 

You are not using it here, you are just saving it in the object and passing it as an argument to `ms.resolver.resolve()` on every call. It will be simpler to just pass it once to `newModuleResolution()`, right?

js/modules.go
```diff
@@ -0,0 +1,122 @@
...
+	"strings"
+
+	"github.com/dop251/goja"
+	"go.k6.io/k6/js/compiler"
+	"go.k6.io/k6/js/modules"
+	"go.k6.io/k6/loader"
+)
+
+type module interface {
+	Instantiate(vu modules.VU) moduleInstance
```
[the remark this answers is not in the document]
mstoykov: I don't remember - I did write 99% of this code in December ;) and then needed to fix 1 thing.

But in general all of this `module` and `moduleInstance` interfaces will be replaced with [`goja` ones](https://github.com/mstoykov/goja/blob/df602e6aca88705755b3954fb539f3fe5a39ef7a/modules.go#L9-L47), once we go to use ESM, so they will need to be uppercase - they also will have kind of different names - but will be exported. (For the record the goja API is very likely to change as some of the later added features made it way too cumbersome, but there are more ... urgent fixes I need to do like, finalizing merging the current goja master)

Also, as mentioned in a different comment, I want to rip the whole modules system out at some point (preferably before ESM) in a separate package and make it usable for tests including tc39. So if anything I can make the other methods uppercase as well. This likely will be needed to be done later either way :shrug:

js/modules.go
mstoykov: Fixed in [f020530](https://github.com/grafana/k6/pull/2881/commits/f0205307e0135b48e84bcb08b497776c3fbd738a) 

I am pretty sure this was the way easier thing to do before some other refactor, but now this is definitely an improvement :tada: 
good catch :clap:

js/modules.go
```diff
@@ -0,0 +1,122 @@
...
+	exports() *goja.Object
+}
+type moduleCacheElement struct {
+	mod module
+	err error
+}
+
+type moduleResolver struct {
+	cache     map[string]moduleCacheElement
+	goModules map[string]interface{}
```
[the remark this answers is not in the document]
na--: Added an issue about this: https://github.com/grafana/k6/issues/2949

imiric (dismissed):
While this looks good from reading the code, it's difficult to consider any functional impact. So let's merge it, and test it thoroughly before the next release.

js/bundle.go
```diff
@@ -74,40 +78,40 @@ func newBundle(
...
-		BaseInitContext:   NewInitContext(piState.Logger, rt, c, compatMode, filesystems, loader.Dir(src.URL)),
+		Source:            src.Data,
 		Options:           options,
 		CompatibilityMode: compatMode,
-		exports:           make(map[string]goja.Callable),
+		callableExports:   make(map[string]struct{}),
+		filesystems:       filesystems,
+		pwd:               loader.Dir(src.URL),
+		logger:            piState.Logger,
 		preInitState:      piState,
```
na--, line 91: why are you adding the `piState.logger` here, when you literally have the `piState` right below? :sweat_smile:

na-- (commented):
LGTM except the 2 nitpicks I left as inline comments

js/bundle.go
mstoykov: fixed in [0c8c08c](https://github.com/grafana/k6/pull/2881/commits/0c8c08cf48197b776867da35cf67523ee554f84b)

na-- (approved):

js/bundle.go
```diff
@@ -74,40 +78,40 @@ func newBundle(
...
-	}
 	// Make a bundle, instantiate it into a throwaway VM to populate caches.
-	rt := goja.New()
-	bundle := Bundle{
+	bundle := &Bundle{
 		Filename:          src.URL,
-		Source:            code,
-		Program:           pgm,
-		BaseInitContext:   NewInitContext(piState.Logger, rt, c, compatMode, filesystems, loader.Dir(src.URL)),
+		Source:            src.Data,
```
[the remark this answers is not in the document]
mstoykov: Fixed in [aa208f8](https://github.com/grafana/k6/pull/2881/commits/aa208f8f92f3eedc0774ba0044c5074526bad510)

Looked at the same thing in lib.Archive ... but there were so many usages of it in hte test that I decided not to go and try to fix all of them :grimacing:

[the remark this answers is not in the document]
na--: Yeah, we can probably simplify a lot of things in that whole code path by using `TestPreInitState` and `TestRunState` more widely and expanding them... Or by grouping other similar "states" into their own structs instead of repeating a bunch of variables in multiple function argument lists and struct properties... :disappointed:

Definitely in separate PRs though, this PR already simplifies quite a few things :tada:

imiric (approved):

Issue #743, opened 2018-08-16
Properly document output (i.e. collector) options


na--:
Specifically, the different environment variables available. Also, after https://github.com/loadimpact/k6/issues/587 is done, the JS/JSON options as well.

na--:
We probably should split up the [results output](https://docs.k6.io/docs/results-output) docs page into multiple pages, one summary page that has an overview and explains how metric outputs work, and for each different output. The current single page is getting huge and it's proving insufficient to properly document the end-of-test summary and all of the supported external outputs and their options.

As is evident from [this issue](https://github.com/loadimpact/k6/issues/1013) and others, we need to better document the InfluxDB output and its options. For it, we might also document common issues that people run into, since there are a few issues people run into fairly frequently.

https://github.com/loadimpact/k6/issues/628 is another connected issue to improve the docs about the cloud collector options. And depending on when and how we implement https://github.com/loadimpact/k6/issues/587, we may need to have a fairly comprehensive section with deprecation information about the environment variable options....

na--:
Closing this, since this was mostly implemented in the new docs at https://k6.io/docs/, and the remaining things are separate issues in the new docs repo (e.g. https://github.com/loadimpact/k6-docs/issues/179, https://github.com/loadimpact/k6-docs/issues/180)

PR #2884, opened 2023-01-30
Add xk6-browser as an experimental module

ankur22:
xk6-browser merge into k6 as an experimental module.

Closes: https://github.com/grafana/xk6-browser/issues/710


js/jsmodules.go
```diff
@@ -15,28 +15,30 @@ import (
...
+		"k6/crypto":                   crypto.New(),
+		"k6/crypto/x509":              x509.New(),
+		"k6/data":                     data.New(),
+		"k6/encoding":                 encoding.New(),
+		"k6/execution":                execution.New(),
+		"k6/experimental/redis":       redis.New(),
+		"k6/experimental/websockets":  &expws.RootModule{},
+		"k6/experimental/timers":      timers.New(),
+		"k6/experimental/tracing":     tracing.New(),
+		"k6/experimental/xk6-browser": browser.New(),
```
olegbespalov, line 36: Should it be just a `k6/experimental/browser`? :thinking:

go.mod
```diff
@@ -13,6 +13,7 @@ require (
 	github.com/go-sourcemap/sourcemap v2.1.4-0.20211119122758-180fcef48034+incompatible
 	github.com/golang/protobuf v1.5.2
 	github.com/gorilla/websocket v1.5.0
+	github.com/grafana/xk6-browser v0.7.0
```
na--, line 16: Run `go mod tidy` and `go mod vendor` and the CI issues should disappear

go.mod
```diff
@@ -24,40 +25,44 @@ require (
 	github.com/mattn/go-colorable v0.1.13
 	github.com/mattn/go-isatty v0.0.16
 	github.com/mccutchen/go-httpbin v1.1.2-0.20190116014521-c5cb2f4802fa
-	github.com/mstoykov/atlas v0.0.0-20220808085829-90340e9998bd
+	github.com/mstoykov/atlas v0.0.0-20220811071828-388f114305dd
```
na--, line 28: hmm :thinking: @mstoykov, I think we forgot to update this in k6 after some optimizations :sweat_smile:

js/jsmodules.go
ankur22: Thanks, resolved in c0a5be61f10deb54c756e638907b353cf0cd492e

go.mod
ankur22: Resolved in 85becc294d1975e90f8c3a480fe1155b59c916c1

go.mod
mstoykov: I guess so - the optimization likely matters very little currently in k6.

If we want we can update it outside of this PR?

go.mod
```diff
@@ -69,8 +74,9 @@ require (
 	github.com/tidwall/match v1.1.1 // indirect
 	go.buf.build/grpc/go/gogo/protobuf v1.4.9 // indirect
 	go.buf.build/grpc/go/prometheus/prometheus v1.4.4 // indirect
+	golang.org/x/exp v0.0.0-20221106115401-f9659909a136 // indirect
```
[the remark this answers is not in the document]
ankur22: Correct, this is no longer in the go.mod file after updating to the release version of xk6-browser.

go.mod
```diff
@@ -24,34 +25,36 @@ require (
...
-	github.com/mstoykov/atlas v0.0.0-20220808085829-90340e9998bd
+	github.com/mstoykov/atlas v0.0.0-20220811071828-388f114305dd
 	github.com/mstoykov/envconfig v1.4.1-0.20220114105314-765c6d8c76f1
 	github.com/nu7hatch/gouuid v0.0.0-20131221200532-179d4d0c4d8d
 	github.com/oxtoacart/bpool v0.0.0-20190530202638-03653db5a59c
 	github.com/pmezard/go-difflib v1.0.0
 	github.com/serenize/snaker v0.0.0-20201027110005-a7ad2135616e
 	github.com/sirupsen/logrus v1.9.0
-	github.com/spf13/afero v1.1.2
+	github.com/spf13/afero v1.9.2
```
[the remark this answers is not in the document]
ankur22: I just used `go get github.com/grafana/xk6-browser@v0.8.0`, and these automatically updated. Is there a way to avoid that? I thought the reason these updated was because of some other dependency requiring that version.

I'll try to undo some of the dep updates.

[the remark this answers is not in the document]
mstoykov: That dependency is likely xk6-browser.  If you can pin the versions for afero and the `x/*` dependencies to the old ones it will likely be better. I would even recommend doing it for atlas as I don't think it is good idea to bump dependencies as close to the end of the cycle as now. 

And AFAIK none of the updates are necessary or urgent for neither project. If that turns out to not be the case we can think about it then IMO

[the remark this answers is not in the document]
na--: :+1: I'd also request what @mstoykov suggests. If these dependency updates are indeed necessary, I'd prefer to handle them in a separate k6 PR, so the diff here is smaller and focused only on xk6-browser.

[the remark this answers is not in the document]
ankur22: I understand why this is being asked. Is there a way to avoid doing it for the `x/*`? Is there a good reason to pin them to an older version?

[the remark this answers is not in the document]
na--: I think we just forgot to update them on the k6 side at the start of the cycle :disappointed: Personally, I'd only slightly prefer to keep the old versions, but I'd also be fine with making a separate PR to update them in k6 as well. 

We still have nearly 3 weeks before the v0.43.0 release, so it should be relatively safe.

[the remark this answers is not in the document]
ankur22: As a team we've agreed to revert the update of the dependencies. I've created the PR [#743](https://github.com/grafana/xk6-browser/pull/743). Would be good if you could also take a quick look to make sure I've not missed anything.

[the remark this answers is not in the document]
na--: it's very difficult to check that in the PR, the best way would be to update this PR with xk6-browser from that branch and see how the diff changes, i.e. something like 
```
go get github.com/grafana/xk6-browser@revert/dependency-versions
go mod tidy
go mod vendor
```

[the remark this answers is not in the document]
mstoykov: Already did that and it works
Notes;
1. you will need to checkout `go.mod` from k6's master. As otherwise you will just be changing on top of the already updated dependancies
2. `go get` does not like "commitish" stuff that have `/` as it has separate meaning for modules ;) . SO you need to use the commit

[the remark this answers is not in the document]
ankur22: Resolved in 317633ab97cd775d501b6a8f3a13b20b3ef60575

js/jsmodules.go
```diff
@@ -32,6 +33,7 @@ func getInternalJSModules() map[string]interface{} {
 		"k6/experimental/websockets": &expws.RootModule{},
 		"k6/experimental/timers":     timers.New(),
 		"k6/experimental/tracing":    tracing.New(),
+		"k6/experimental/browser":    browser.New(),
```
na--, line 36: To require that the browser module is explicitly enabled with an environment variable, you'd basically need a bogus `RootModule` implementation here on the k6 side. I think we can add a `browser` subfolder in [here](https://github.com/grafana/k6/tree/master/js/modules/k6/experimental) and add the module there. That module's `NewModuleInstance()` method can make the env var check and, if the env var is present, call xk6-browser's actual `RootModule.NewModuleInstance()`.

Also, a e2e test or two in https://github.com/grafana/k6/blob/master/cmd/tests/cmd_run_test.go for that check would be ideal :pray:

ankur22: I think the team would be happy to do that, which would mean that we're not going to create a generic permission like the one in the example [here](https://deno.land/manual@v1.15.2/getting_started/permissions)?

ankur22: Resolved in: 04a9daf157d477edd43c8790b7e43e74bb582c38 & 989169120543a6a4d43912864782f3e2a76b6889.

cmd/tests/cmd_run_test.go
```diff
@@ -1644,3 +1644,75 @@ func TestPrometheusRemoteWriteOutput(t *testing.T) {
...
+		t.Run(tt.name, func(t *testing.T) {
+			script := `
+			import { chromium } from 'k6/experimental/browser';
+
+			export default function() {};
+			`
+
+			ts := getSingleFileTestState(t, script, []string{}, tt.expectedExitCode)
+			if tt.envVarValue != "" {
+				t.Setenv("K6_BROWSER_ENABLE_RUN", tt.envVarValue)
```
ankur22, line 1701: @na-- @mstoykov do you know of a better way to work with env vars that are also exposed to the `RootModule`?

na--: I added https://github.com/grafana/k6/pull/2890 to address this problem

na--: https://github.com/grafana/k6/pull/2890 was just merged, now you should be able to get the environment from `VU`'s `InitEnv`: 
https://github.com/grafana/k6/blob/134ca65a64f758c19f40fd67b0041b23733ef413/js/modules/modules.go#L44-L45
https://github.com/grafana/k6/blob/134ca65a64f758c19f40fd67b0041b23733ef413/js/common/initenv.go#L19

ankur22: Refactored in 919450d3e632110dd5bff9b3b8d9a508d4f52a11

cmd/tests/cmd_run_test.go
```diff
@@ -1644,3 +1644,75 @@ func TestPrometheusRemoteWriteOutput(t *testing.T) {
...
+		{
+			name:             "env var set and set to true",
+			envVarValue:      "true",
+			expectedExitCode: 0,
+			expectedError:    "",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
```
na--, line 1692:
```suggestion
	for _, tt := range tests {
		tt := tt
		t.Run(tt.name, func(t *testing.T) {
			t.Parallel()
```

js/modules/k6/experimental/browser/rootmodule.go
```diff
@@ -0,0 +1,49 @@
...
+	throwError := func() {
+		msg := "To run browser tests set env var K6_BROWSER_ENABLE_RUN=true"
+		if m, ok := env.LookupEnv("K6_BROWSER_ENABLE_RUN_MSG"); ok && m != "" {
+			msg = m
+		}
+
+		common.Throw(vu.Runtime(), errors.New(msg))
+	}
+
+	vs, ok := env.LookupEnv("K6_BROWSER_ENABLE_RUN")
```
na--, line 38:
```suggestion
	vs, ok := env.LookupEnv("K6_BROWSER_ENABLED")
```

This sounds a bit better? :thinking:

na-- (commented):
LGTM except the minor nitpicks I noted inline :tada:

na-- (dismissed):
:heart: :tada:

js/modules/k6/experimental/browser/rootmodule.go
```diff
@@ -0,0 +1,58 @@
...
+	"strconv"
+
+	xk6browser "github.com/grafana/xk6-browser/browser"
+	"go.k6.io/k6/js/common"
+	"go.k6.io/k6/js/modules"
+)
+
+type (
+	// RootModule is a wrapper around the experimental
+	// browser RootModule. It will prevent browser tests
```
imiric, line 17:
```suggestion
	// browser RootModule. It will prevent browser test
```

js/modules/k6/experimental/browser/rootmodule.go
```diff
@@ -0,0 +1,58 @@
...
+	// runs unless K6_BROWSER_ENABLED env var is set.
+	RootModule struct {
+		rm *xk6browser.RootModule
+	}
+)
+
+// New creates a experimental browser RootModule
```
imiric, line 24:
```suggestion
// New creates an experimental browser RootModule
```

imiric (approved):
I didn't review the vendored code :sweat_smile:, but the wrapper module and tests LGTM. Just spotted a couple of typos.

Nice work! :rocket:

imiric (approved):

na-- (approved):
LGTM, just please remember to squash + merge the PR, don't rebase or use a merge commit

codebien (approved):
:tada:

PR #2885, opened 2023-01-30
Remove the unnecessary `execution.Scheduler.Init()`

na--:
This is the first commit from https://github.com/grafana/k6/pull/2815, plus an extra test that checks the exit code when aborting a `k6 run --paused` test with Ctrl+C. 

Unfortunately, it turned out that https://github.com/grafana/k6/commit/19dd50593b90a79fa1883e248c63d385dbe7a8b9 from https://github.com/grafana/k6/pull/2813 introduced a regression. Interrupting the k6 process with a signal should result in a `105` exit code (`ExternalAbort`), but with that commit k6 exited with `103` if k6 was `--paused`. The good news is that the fix was already done in the first commit of https://github.com/grafana/k6/pull/2815, which was entirely self-sufficient and can be moved to this separate PR. As a bonus, this makes reviewing everything even easier... :sweat_smile: 

So, some details about the commit itself... :sweat_smile: In short, there is no need to have 2 separate `Run()` and `Init()` methods in the `execution.Scheduler`. Instead, Run() can start the VU initialization itself. As a bonus, this immediately makes the error handling around init errors much more in line with other error handling, allowing us to resolve the bug, but also to respect `--linger` and to try and execute `handleSummary()` if there were problems. Except the first init that is used to get the exported options, of course, that one is still special.

### DIFF
--- a/execution/scheduler.go
+++ b/execution/scheduler.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"runtime"
+	"sync"
 	"sync/atomic"
 	"time"
 
@@ -26,10 +27,6 @@ type Scheduler struct {
 	maxDuration     time.Duration // cached value derived from the execution plan
 	maxPossibleVUs  uint64        // cached value derived from the execution plan
 	state           *lib.ExecutionState
-
-	// TODO: remove these when we don't have separate Init() and Run() methods
-	// and can use a context + a WaitGroup (or something like that)
-	stopVUsEmission, vusEmissionStopped chan struct{}
 }
 
 // NewScheduler creates and returns a new Scheduler instance, without
@@ -84,9 +81,6 @@ func NewScheduler(trs *lib.TestRunState) (*Scheduler, error) {
 		maxDuration:     maxDuration,
 		maxPossibleVUs:  maxPossibleVUs,
 		state:           executionState,
-
-		stopVUsEmission:    make(chan struct{}),
-		vusEmissionStopped: make(chan struct{}),
 	}, nil
 }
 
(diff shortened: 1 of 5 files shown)


execution/scheduler.go
```diff
@@ -243,23 +239,17 @@ func (e *Scheduler) emitVUsAndVUsMax(ctx context.Context, out chan<- metrics.Sam
...
-	e.emitVUsAndVUsMax(ctx, samplesOut)
-	defer func() {
-		if err != nil {
-			close(e.stopVUsEmission)
-			<-e.vusEmissionStopped
-		}
-	}()
+// initVusAndExecutors concurrently initializes all of the planned VUs and then
+// sequentially initializes all of the configured executors.
+func (e *Scheduler) initVusAndExecutors(ctx context.Context, samplesOut chan<- metrics.SampleContainer) (err error) {
```
mstoykov, line 251: nit: 
```suggestion
func (e *Scheduler) initVUsAndExecutors(ctx context.Context, samplesOut chan<- metrics.SampleContainer) (err error) {
```

mstoykov (approved):
LGTM apart from the naming nitpick. 

But to be honest I will likely need a day or two to remember all the intricacies, *I* knew, of this code to be certain that I don't forget about a corner case.

 So I am hopeful that you working with it all the time means you would've caught it :P

execution/scheduler.go
na--: Resolved by [this force-push](https://github.com/grafana/k6/compare/60c1026af3fd2691487d52f313d7edea99c1b8bd..10700cd485373867c65d1e2e5ea654d3d7fa6fce).

na--:
> But to be honest I will likely need a day or two to remember all the intricacies, I knew, of this code to be certain that I don't forget about a corner case.
> So I am hopeful that you working with it all the time means you would've caught it :P

This whole sequence of 10+ PRs and tons of e2e tests has been mostly an attempt to simplify the old mess, so nobody has to remember all of its corner cases :sweat_smile: :crossed_fingers: :pray: 

There is some unavoidable inherent complexity here, true, but most of the complexity before was accidental... :disappointed:  That, coupled with all of the previously undefined behavior regarding exit codes and error handing (https://github.com/grafana/k6/issues/2804), was a millstone around our neck - a huge piece of technical debt that we had to constantly work around... :disappointed:

mstoykov (approved):

codebien (approved):

PR #2886, opened 2023-01-31
Add sampling capabilities to the tracing module

oleiade:
## About

This Pull Request adds probabilistic sampling support to the tracing module. Using it, users can specify the percentage of their instrumented requests which should have their trace context header's sampled field set to true.

From a user-interface perspective, this feature is introduced as an option meant to be set during either the instantiation of a `tracing.Client`, or during a call to `instrumentHTTP`. It consists of a single integer field of the tracing options, whose value, within the `0 <= n <= 100` bounds, is meant to represent a percentage.

```javascript
// The sampling option is available when instantiating a Client
const sampledHTTP = new tracing.Client({
	propagator: "w3c",
	
	// Up to 55% of the requests emitted using the client instance's
	// methods will have their trace context header's sampled field
	// set to true.
	sampling: 55,
})

// It is also available when using the instrumentHTTP function
tracing.InstrumentHTTP({
	propagator: "w3c",
	
	// Up to 27% of the requests made with the HTTP module's functions
	// will have their trace context header's sampled field set to true
	// from this point onward 
	sampling: 27,
})
```

This feature is implemented using probabilistic sampling. Meaning that over an extensive series of iterations, the number of requests with their sampled flag set to true will tend to converge to the selected `sampling` rate value. It was a design choice without explicit constraints, preferring randomness over accuracy: each trace context generation evaluates its chance of being sampled independently. If you find an implementation emphasizing accuracy more suitable, I'm open to exploring that route; just let me know.

Note that regardless of the sampled header flag's value, the `trace_id` metadata will be attached to each output sample.

### DIFF
--- a/cmd/tests/tracing_module_test.go
+++ b/cmd/tests/tracing_module_test.go
@@ -14,6 +14,7 @@ import (
 	"github.com/stretchr/testify/require"
 	"go.k6.io/k6/cmd"
 	"go.k6.io/k6/lib/testutils/httpmultibin"
+	"go.k6.io/k6/metrics"
 )
 
 func TestTracingModuleClient(t *testing.T) {
@@ -58,7 +59,7 @@ func TestTracingModuleClient(t *testing.T) {
 	jsonResults, err := afero.ReadFile(ts.FS, "results.json")
 	require.NoError(t, err)
 
-	assertHasTraceIDMetadata(t, jsonResults)
+	assertHasTraceIDMetadata(t, jsonResults, 9, tb.Replacer.Replace("HTTPBIN_IP_URL/tracing"))
 }
 
 func TestTracingClient_DoesNotInterfereWithHTTPModule(t *testing.T) {
@@ -98,6 +99,107 @@ func TestTracingClient_DoesNotInterfereWithHTTPModule(t *testing.T) {
 
 	assert.Equal(t, int64(3), atomic.LoadInt64(&gotRequests))
 	assert.Equal(t, int64(2), atomic.LoadInt64(&gotInstrumentedRequests))
+}
+
+func TestTracingModuleClient_HundredPercentSampling(t *testing.T) {
+	t.Parallel()
+	tb := httpmultibin.NewHTTPMultiBin(t)
+
+	var gotRequests int64
+	var gotSampleFlags int64
+
+	tb.Mux.HandleFunc("/tracing", func(w http.ResponseWriter, r *http.Request) {
+		atomic.AddInt64(&gotRequests, 1)
+
+		traceparent := r.Header.Get("traceparent")
+		require.NotEmpty(t, traceparent)
+		require.Len(t, traceparent, 55)
+
+		if traceparent[54] == '1' {
+			atomic.AddInt64(&gotSampleFlags, 1)
+		}
+	})
+
+	script := tb.Replacer.Replace(`
+		import http from "k6/http";
+		import { check } from "k6";
+		import tracing from "k6/experimental/tracing";
+
+		export const options = {
+			// 100 iterations to make sure we get 100% sampling
+			iterations: 100,
+		}
+
+		const instrumentedHTTP = new tracing.Client({
+			propagator: "w3c",
+
+			// 100% sampling
+			sampling: 1.0,
+		})
+
+		export default function () {
+			instrumentedHTTP.get("HTTPBIN_IP_URL/tracing");
+		};
+	`)
+
+	ts := getSingleFileTestState(t, script, []string{"--out", "json=results.json"}, 0)
+	cmd.ExecuteWithGlobalState(ts.GlobalState)
+
+	assert.Equal(t, int64(100), atomic.LoadInt64(&gotSampleFlags))
+	assert.Equal(t, int64(100), atomic.LoadInt64(&gotRequests))
+
+	jsonResults, err := afero.ReadFile(ts.FS, "results.json")
+	require.NoError(t, err)
+
+	assertHasTraceIDMetadata(t, jsonResults, 100, tb.Replacer.Replace("HTTPBIN_IP_URL/tracing"))
+}
+
+func TestTracingModuleClient_ZeroPercentSampling(t *testing.T) {
+	t.Parallel()
+	tb := httpmultibin.NewHTTPMultiBin(t)
+
+	var gotRequests int64
+	var gotSampleFlags int64
+
+	tb.Mux.HandleFunc("/tracing", func(w http.ResponseWriter, r *http.Request) {
+		atomic.AddInt64(&gotRequests, 1)
+
+		traceparent := r.Header.Get("traceparent")
+		require.NotEmpty(t, traceparent)
+		require.Len(t, traceparent, 55)
+
+		if traceparent[54] == '1' {
+			atomic.AddInt64(&gotSampleFlags, 1)
+		}
+	})
+
+	script := tb.Replacer.Replace(`
+		import http from "k6/http";
+		import { check } from "k6";
+		import tracing from "k6/experimental/tracing";
+
+		export const options = {
+			// 100 iterations to make sure we get 100% sampling
+			iterations: 100,
+		}
+
+		const instrumentedHTTP = new tracing.Client({
+			propagator: "w3c",
+
+			// 0% sampling
+			sampling: 0.0,
+		})
+
+		export default function () {
+			instrumentedHTTP.get("HTTPBIN_IP_URL/tracing");
+		};
+	`)
+
+	ts := getSingleFileTestState(t, script, []string{}, 0)
+	cmd.ExecuteWithGlobalState(ts.GlobalState)
+
+	assert.Equal(t, int64(0), atomic.LoadInt64(&gotSampleFlags))
+	assert.Equal(t, int64(100), atomic.LoadInt64(&gotRequests))
 }
 
 func TestTracingInstrumentHTTP_W3C(t *testing.T) {
@@ -141,7 +243,7 @@ func TestTracingInstrumentHTTP_W3C(t *testing.T) {
 	jsonResults, err := afero.ReadFile(ts.FS, "results.json")
 	require.NoError(t, err)
 
-	assertHasTraceIDMetadata(t, jsonResults)
+	assertHasTraceIDMetadata(t, jsonResults, 9, tb.Replacer.Replace("HTTPBIN_IP_URL/tracing"))
 }
 
 func TestTracingInstrumentHTTP_Jaeger(t *testing.T) {
@@ -185,7 +287,7 @@ func TestTracingInstrumentHTTP_Jaeger(t *testing.T) {
 	jsonResults, err := afero.ReadFile(ts.FS, "results.json")
 	require.NoError(t, err)
 
-	assertHasTraceIDMetadata(t, jsonResults)
+	assertHasTraceIDMetadata(t, jsonResults, 8, tb.Replacer.Replace("HTTPBIN_IP_URL/tracing"))
 }
 
 func TestTracingInstrumentHTTP_FillsParams(t *testing.T) {
@@ -236,7 +338,7 @@ func TestTracingInstrumentHTTP_FillsParams(t *testing.T) {
 	jsonResults, err := afero.ReadFile(ts.FS, "results.json")
 	require.NoError(t, err)
 
-	assertHasTraceIDMetadata(t, jsonResults)
+	assertHasTraceIDMetadata(t, jsonResults, 8, tb.Replacer.Replace("HTTPBIN_IP_URL/tracing"))
 }
 
 func TestTracingInstrummentHTTP_SupportsMultipleTestScripts(t *testing.T) {
@@ -288,13 +390,22 @@ func TestTracingInstrummentHTTP_SupportsMultipleTestScripts(t *testing.T) {
 	require.NoError(t, err)
 
 	assert.Equal(t, int64(1), atomic.LoadInt64(&gotRequests))
-	assertHasTraceIDMetadata(t, jsonResults)
+	assertHasTraceIDMetadata(t, jsonResults, 1, tb.Replacer.Replace("HTTPBIN_IP_URL/tracing"))
 }
 
 // assertHasTraceIDMetadata checks that the trace_id metadata is present and has the correct format
 // for all http metrics in the json results file.
-func assertHasTraceIDMetadata(t *testing.T, jsonResults []byte) {
+//
+// The `expectOccurences` parameter is used to check that the trace_id metadata is present for the
+// expected number of http metrics. For instance, in a script with 2 http requests, the
+// `expectOccurences` parameter should be 2.
+//
+// The onUrls parameter is used to check that the trace_id metadata is present for the data points
+// with the expected URLs. Its value should reflect the URLs used in the script.
+func assertHasTraceIDMetadata(t *testing.T, jsonResults []byte, expectOccurences int, onUrls ...string) {
 	gotHTTPDataPoints := false
+
+	urlHTTPTraceIDs := make(map[string]map[string]int)
 
 	for _, jsonLine := range bytes.Split(jsonResults, []byte("\n")) {
 		if len(jsonLine) == 0 {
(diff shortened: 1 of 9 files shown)


oleiade:
For your information, https://github.com/grafana/k6/pull/2886/commits/377abf9518133fc223b564fa717df4b925faef04 incorporates improvements to the `assertHasTraceIDMetadata` function as suggested by @mstoykov in previous PRs 👍🏻

codebien:
It could be good to address https://github.com/grafana/k6/issues/2889 here

oleiade:
@codebien I'll give it shot, if it takes more than an hour or two, I would address it during cooldown in a separate PR 👍🏻

oleiade:
@codebien #2889 was addressed in a separate PR https://github.com/grafana/k6/pull/2894 🚀

cmd/tests/tracing_module_test.go
```diff
@@ -99,6 +100,107 @@ func TestTracingClient_DoesNotInterfereWithHTTPModule(t *testing.T) {
...
+		export const options = {
+			// 100 iterations to make sure we get 100% sampling
+			iterations: 100,
+		}
+
+		const instrumentedHTTP = new tracing.Client({
+			propagator: "w3c",
+
+			// 100% sampling
+			sampling: 0,
```
mstoykov, line 189: Can we maybe have a test with sampling of like `10` as that is more likely wht users will do - sample some non-0 non-100 percent.

js/modules/k6/experimental/tracing/encoding.go
```diff
@@ -17,3 +17,15 @@ func randHexString(n int) string {
...
+// chance returns true with a `percentage` chance, otherwise false.
+// the `percentage` argument is expected to be
+// within 0 <= percentage <= 100 range.
+//
+// The chance function works under the assumption that the
+// go rand module has been seeded with a non-deterministic
+// value.
+func chance(percentage int) bool {
+	//nolint:gosec
+	return rand.Intn(100) < percentage
```
mstoykov, line 30: Using the global `rand` means that we lock across all CPU cores to get access to the same rand.

That actually is a problem in some cases and maybe in thos one - so preferably we should use a per VU :thinking: rand source. There is one in the goja VM, but maybe for this we can have one in the tracing module's module instance?

js/modules/k6/experimental/tracing/module.go
```diff
@@ -38,6 +40,11 @@ func New() *RootModule {
 // NewModuleInstance implements the modules.Module interface and returns
 // a new instance for each VU.
 func (*RootModule) NewModuleInstance(vu modules.VU) modules.Instance {
+	// Seed the random number generator with the current time.
+	// This ensures that any call to rand.Intn() will return
+	// less-deterministic results.
+	rand.Seed(time.Now().UTC().UnixNano())
```
mstoykov, line 46: `rand` is a global so this likely should be a field in the `ModuleInstance` see my other commetn

js/modules/k6/experimental/tracing/options_test.go
```diff
@@ -5,11 +5,19 @@ import "testing"
...
+		lowerBoundSampling       = 0
+		upperBoundSampling       = 100
+		lowerOutOfBoundsSampling = -1
+		upperOutOfBoundsSampling = 101
+	)
 
 	type fields struct {
 		Propagator string
-		Sampling   *float64
+		Sampling   *int
```
mstoykov, line 20: Is this change in type what we want? 

DO we not want to support `12.5%` sampling? Or should that be left for later?

js/modules/k6/experimental/tracing/propagator.go
```diff
@@ -9,6 +9,22 @@ type Propagator interface {
...
+
+	// Sampler ensures the implementer is able to perform sampling decisions.
+	Sampler
+
+	// SamplingFlag should return the sampling flag to use
+	// as part of the trace context header, according to
+	// the sampler's decision of whether or not a given
+	// trace should be sampled.
+	SamplingFlag() string
+}
```
mstoykov, line 26: I feel like we won't be having any *other* kind of propagator so maybe just have one interface ?

js/modules/k6/experimental/tracing/propagator.go
```diff
@@ -52,18 +106,59 @@ const (
...
+// set the flag to either W3CSampledTraceFlag or W3CUnsampledTraceFlag,
+// with a percent of chance based on the propagator's sampling
+// rate.
+func (p *JaegerPropagator) SamplingFlag() string {
+	if p.Sampled() {
+		return JaegerSampledTraceFlag
+	}
+
+	return JaegerUnsampledTraceFlag
+}
```
mstoykov, line 157: I kind of feel like this can be refactored to be a function that based on `Sampled` returns one of two strings that are given to it.

```golang
   flags := ternaryOperator(p.Sampled(), JaegerSampledTraceFlag, JaegerUnsampledTraceFlag)
```

Given that his is basically what happens in both implementations. I also kind of don't see what else could happen ... maybe there is only *1* option :shrug: 

Maybe the header isn't sent at all :shrug: 

Both of those can be handled with the "ternaryOperator" 

I am not really happy with how my code works as well but it does drop > 20 lines of code and a interface method so seems better to me

js/modules/k6/experimental/tracing/client.go
```diff
@@ -79,11 +79,16 @@ func (c *Client) Configure(opts options) error {
 		return fmt.Errorf("invalid options: %w", err)
 	}
 
+	var sampler Sampler = NewAlwaysSampledSampler()
+	if opts.Sampling != nil {
```
mstoykov, line 83:
```suggestion
	if opts.Sampling != nil && *opts.Sampling != 100 {
```

I feel that if we went through the trouble of actually having 2 implementations - let's use them more ;)

cmd/tests/tracing_module_test.go
oleiade: The reason I did not have such a test is that using the current probabilistic sampling technique the only insurance we have is that on a large enough number of samples, we will be converging towards 10% of sampled operations.

Maybe that's a good enough reason to consider switching to a more deterministic sampling strategy 🤔 I was thinking an alternative might be filling a boolean buffer of a fixed size with P% `true` values, and shuffling it every time we observed all of its values. That way we can assert that for a given buffer/window size we indeed have P% sampled operations.

What do you think?

js/modules/k6/experimental/tracing/encoding.go
oleiade: TIL 🤓 I genuinely didn't know. That's a super valuable insight, thanks for that 🙏🏻 
I'll add a rand source to the module's instances indeed 👍🏻

js/modules/k6/experimental/tracing/module.go
oleiade: Indeed 👍🏻

js/modules/k6/experimental/tracing/options_test.go
oleiade: This is based on an assumption of mine. I switched to `int` because I had the assumption `float` would lead to a usage of the form `sampling: 0.4` to express 40%, which as a user always confuses me. But I didn't consider the possibility of providing decimals to a percent 🤔

Any ideas, position, or feedback on this @dgzlopes and @Blinkuu? Are you fine with a `sampling` option in the form of a `int` bounded as : `0 <= n <= 100`, or would you prefer a `float64` with the same bounds but supporting decimals? 🙇🏻

dgzlopes: > would you prefer a float64 with the same bounds but supporting decimals? 🙇🏻

I think I would go with this one :+1:

Blinkuu: In Grafana Agent, sampling is configured using the following:

```json
 {
    "default_strategy": {
      "param": 1.0,
      "type": "probabilistic"
    }
}
```

In the above, `0.0` is the lower bound and `1.0` is the upper bound.

If we are implementing _probabilistic_ sampling, then I think we should go with the `float64` and `[0.0, 1.0]` range. I've never seen a probability higher than `1.0`.

cmd/tests/tracing_module_test.go
oleiade: As I'm at it, can I get your input on this @Blinkuu, I imagine you might have valuable input on this ? cheers 🤓

js/modules/k6/experimental/tracing/propagator.go
oleiade: As mentioned in another comment, I dropped the interface indeed.

js/modules/k6/experimental/tracing/sampling.go
```diff
@@ -0,0 +1,64 @@
+package tracing
+
+import "math/rand"
+
+// Sampler is an interface defining probabilistic sampling strategies.
+type Sampler interface {
```
Blinkuu, line 6: I suggest you look at the `Sampler` interface definition the Open Telemetry Go SDK provides. I think it's a bit more powerful:

https://github.com/open-telemetry/opentelemetry-go/blob/3015c86cfd1e1868b9c57a998d44ebe25acffedc/sdk/trace/sampling.go#L27-L54.

cmd/tests/tracing_module_test.go
Blinkuu: Sorry, I'm not entirely sure what you mean here. How would it differ if you used a slice of `bool` and traversed it with a pointer instead of making a decision on-the-fly in terms of determinicity? Could you elaborate on both ideas?

oleiade: It would differ in that with the slice of `bool` having a fixed size (that I would refer to as "window" size), I would have some capacity to test the outcome of sampling. As I would know the window size ("the slice contains 256 booleans, 10% of which should be `false` before we re-shuffle it"), I can, in a test, check that it contains the expected percentage of positives before we reshuffle it.

Whereas with the current probabilistic sampling strategy, as far as I found out through some research, it wouldn't be relevant to try to test that P% of the samples are picked, as every "pick" has a chance of P% but is random. For a 10% sampling rater, I could test that over 100K iterations, the number of picked samples would converge towards 10K, but there is still a (small) chance to miss that by a large shot and be whatever (7K); leading to a flaky test.

js/modules/k6/experimental/tracing/sampling.go
oleiade: Thanks for the pointer 🙇🏻 It looks thorough indeed, but I don't necessarily see anything that would need to introduce at the moment though. Did you have anything specific in mind? 🤓

Blinkuu: The `Sampler` interface you proposed seems to be coupled with "probabilistic" sampling. For example, you define the `SetSamplingRate(percent float64)` method, which only makes sense for "probabilistic" sampling.

The interface that Open Telemetry has is much more flexible. With `ShouldSample` method accepting the `SamplingParameters`, it's much easier to implement a struct implementing this interface that makes decisions based on more parameters.

cmd/tests/tracing_module_test.go
Blinkuu: > I can, in a test, check that it contains the expected percentage of positives before we reshuffle it.

I believe there are better ways to test code that depends on randomness. You can simply accept an interface that generates random values and mock it in the tests for determinism. You can look at how we do it in the insights project [here](https://github.com/grafana/k6-cloud-insights/blob/4bf1ac687a489568848b7e7541e549ec07379aa4/pkg/util/traceid/traceid_test.go#L141).

Regarding testing you have "exactly" 10% of requests being sampled - I see little value in such a test, especially if it were to be a "unit test".

js/modules/k6/experimental/tracing/options.go
```diff
@@ -9,13 +9,14 @@ import (
 // tracing.instrumentHTTP() method.
 type options struct {
 	// Propagation is the propagation format to use for the tracer.
-	Propagator string `js:"propagator"`
+	Propagator string `json:"propagator"`
 
-	// Sampling is the sampling rate to use for the tracer.
-	Sampling *float64 `js:"sampling"`
+	// Sampling is the sampling rate to use for the
+	// tracer, expressed in percents: 0 <= n <= 100.
```
mstoykov, line 15: This comment is no longer valid as it is now between 0.0 and 1.0

js/modules/k6/experimental/tracing/options.go
```diff
@@ -9,13 +9,14 @@ import (
...
+	Sampling *float64 `json:"sampling"`
```
mstoykov, line 16: Is this a pointer so that you can just take this from goja and know if it was set?

If so I kind of prefer if this is parsed from the goja value into a none pointer and then not actually need to access it as a pointer. But this can be done later as well

js/modules/k6/experimental/tracing/options_test.go
```diff
@@ -39,16 +47,58 @@ func TestOptionsValidate(t *testing.T) {
...
+		{
+			name: "sampling rate < 0 is invalid",
+			fields: fields{
+				Propagator: "w3c",
+				Sampling:   &lowerOutOfBoundsSampling,
+			},
+			wantErr: true,
+		},
+		{
+			name: "sampling rater > 100 is invalid",
```
mstoykov, line 90:
```suggestion
			name: "sampling rater > 1.0 is invalid",
```

js/modules/k6/experimental/tracing/options_test.go
```diff
@@ -39,16 +47,58 @@ func TestOptionsValidate(t *testing.T) {
...
+		{
+			name: "sampling rate = 0 is valid",
+			fields: fields{
+				Propagator: "w3c",
+				Sampling:   &lowerBoundSampling,
+			},
+			wantErr: false,
+		},
+		{
+			name: "sampling rate = 100 is valid",
```
mstoykov, line 74:
```suggestion
			name: "sampling rate = 1.0 is valid",
```

cmd/tests/tracing_module_test.go
oleiade: Coincidentally, I added support for custom randomness generation in another PR #2894 already. I guess that's a good data point indicating we want to actually do it then. I'll implement it in this PR too then. Thanks a lot for your input 🙇🏻

js/modules/k6/experimental/tracing/sampling.go
oleiade: I see your point then. You're right in that sense it's more flexible indeed 👍🏻 gonna give a shot at taking inspiration from it in this PR then. Thanks for your input 🙇🏻

js/modules/k6/experimental/tracing/options.go
oleiade: It's a pointer to differentiate between unset field and zero value field. That's the only sane way of doing it in Go that I know 🤓 

Do I understand your suggestion correctly: switch back to a value field instead of a pointer field, have an options constructor that takes a Goja value (holding options) as a default, and manage whether that specific field is bullish explicitly there?

mstoykov: This also can enable warnign users on providing unsupported options so that if they misspell something it is easier to find out.

But this can be done in a separte PR, no need to do this now.

js/modules/k6/experimental/tracing/sampling.go
oleiade: Adopted the same approach 👍🏻 Thanks for the pointer. I did not add a Parameters type, as we really don't need it yet at the moment, but it shouldn't be a problem to add it at a later point in time if the requirement changes 🙇🏻

js/modules/k6/experimental/tracing/options.go
oleiade: I addressed it by having a `newOptions` constructor which builds options from a `goja.Value` and defaults to value `1.0` (100%) if the field is absent/nullish. The field is not a pointer anymore 👍🏻

cmd/tests/tracing_module_test.go
oleiade: As testing a random behavior makes little sense, and I believe testing a mocked randomness behavior wouldn't bring much value either. I'll close this. Feel free to reopen if you have a concrete use case and approach for testing this reliably 👍🏻

mstoykov (dismissed):
LGTM!

A nitpick was that it seems to be we can just inline the `chance` function, but I am okay either way

js/modules/k6/experimental/tracing/sampling.go
```diff
@@ -0,0 +1,59 @@
+package tracing
+
+import (
+	"math/rand"
+)
+
+// Sampler is an interface defining probabilistic sampling strategies.
```
codebien, line 7:
```suggestion
// Sampler is an interface defining a sampling strategy.
```

js/modules/k6/experimental/tracing/propagator.go
```diff
@@ -52,18 +75,54 @@ const (
...
+type JaegerPropagator struct {
+	// Sampler is used to determine whether or not a trace should be sampled.
+	Sampler
+}
+
+// NewJaegerPropagator returns a new JaegerPropagator with the given sampler.
+//
+// Note that we allocate the propagator on the heap to ensure we conform
+// to the Sampler interface, as the [Sampler.SetSamplingRate]
+// method has a pointer receiver.
```
codebien, line 96: Is it still valid note? I don't see a SetSamplingRate method

js/modules/k6/experimental/tracing/propagator.go
```diff
@@ -52,18 +75,54 @@ const (
...
+
+	return rhs
+}
+
+// Ensure the JaegerPropagator implements the Propagator and
+// SampledPropagator interface
+var (
+	_ Propagator = &JaegerPropagator{}
+	_ Sampler    = &JaegerPropagator{}
+)
```
codebien, line 128: nit: we could move them together with W3C

js/modules/k6/experimental/tracing/sampling.go
```diff
@@ -0,0 +1,59 @@
...
+	random *rand.Rand
+
+	// samplingRate is a chance value defined as a percentage
+	// value within 0.0 <= samplingRate <= 1.0 bounds.
+	samplingRate float64
+}
+
+// NewProbabilisticSampler returns a new ProbablisticSampler with the provided sampling rate.
+func NewProbabilisticSampler(samplingRate float64) *ProbabilisticSampler {
+	return &ProbabilisticSampler{samplingRate: samplingRate}
```
codebien, line 27: Should we validate or discard negative numbers here?

codebien (changes requested):
Overall LGTM, has some minor comments. Really well done, I'm out from tracing domain but it was easy to understand :clap:

js/modules/k6/experimental/tracing/sampling.go
oleiade: Good call, they keep the number withing bounds in the opentelemetry implementation too. I've added bounds check, and will set any number `< 0.0` to `0.0`, and any number bigger than `1.0` to `1.0`.

oleiade:
Addressed all comments, looking forward to your 👍🏻 👏🏻 🙇🏻

mstoykov (approved):

codebien (approved):
LGTM

I restarted the failed checks

PR #2890, opened 2023-02-01
Pass `LookupEnv()` to `js/` modules

na--:
This PR threads a way for JS modules to lookup environment variables without directly accessing the `os` package. That ended up quite simple to do, with some additional simplification as a bonus, and it would be nice to have it before https://github.com/grafana/k6/pull/2884

### DIFF
--- a/js/runner_test.go
+++ b/js/runner_test.go
@@ -2501,8 +2501,8 @@ func TestMinIterationDurationIsCancellable(t *testing.T) {
 	}
 }
 
-//nolint:paralleltest
 func TestForceHTTP1Feature(t *testing.T) {
+	t.Parallel()
 	cases := map[string]struct {
 		godebug               string
 		expectedForceH1Result bool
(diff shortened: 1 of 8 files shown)


na--:
Will split this PR into two separate ones, so I'll only leave the `LookupEnv` changes here.

js/runner_test.go
```diff
@@ -2503,6 +2503,7 @@ func TestMinIterationDurationIsCancellable(t *testing.T) {
 
 //nolint:paralleltest
```
mstoykov, line 2504:
```suggestion
```

mstoykov (dismissed):
LGTM! 

left a nit about a nolint comment which isnt' blocking but if we can drop it safely - let's do it

js/runner_test.go
na--: Resolved by the [latest force-push](https://github.com/grafana/k6/compare/325d738133025da672eae4e41a094b72e4e60542..046c3feebcb66f453b5a9d350a36fc40a03a0ebb). I noticed we have similar no longer needed `//nolint` commands in a bunch of other places, I'll try to clean them up in https://github.com/grafana/k6/pull/2891 later

mstoykov (approved):

imiric (approved):
:+1:

PR #2891, opened 2023-02-01
Add lint check for os and syscall usage and fix some of the issues

na--:
Built on top of, https://github.com/grafana/k6/pull/2890, this PR adds linter rules that forbid the usage of most of the `os` and `syscall` packages. It also has the mechanical fixes to some of that previously improper usage in a bunch of places in the k6 codebase. I could safely replace a bunch of `os` usages with [the `fs` package introduced in Go 1.16](https://go.dev/doc/go1.16#fs) and the equivalent `syscall` constants, since the things in `os` are now [type aliases](https://yourbasic.org/golang/type-alias/) (see [this](https://github.com/golang/go/blob/56a14ad4bc19d5ee9d4257f370a570377e81e544/src/os/error.go#L21-L24), [this](https://github.com/golang/go/blob/56a14ad4bc19d5ee9d4257f370a570377e81e544/src/os/types.go#L20-L56) and [this](https://github.com/golang/go/blob/56a14ad4bc19d5ee9d4257f370a570377e81e544/src/os/file.go#L70-L83))

### DIFF
--- a/cmd/archive_test.go
+++ b/cmd/archive_test.go
@@ -6,9 +6,10 @@ import (
 	"encoding/json"
 	"errors"
 	"io"
+	"io/fs"
 	"io/ioutil"
-	"os"
 	"path/filepath"
+	"syscall"
 	"testing"
 
 	"github.com/spf13/afero"
@@ -155,10 +156,10 @@ func TestArchiveNotContainsEnv(t *testing.T) {
 }
 
 // untar untars a `fileName` file to a `destination` path
-func untar(t *testing.T, fs afero.Fs, fileName string, destination string) error {
+func untar(t *testing.T, fileSystem afero.Fs, fileName string, destination string) error {
 	t.Helper()
 
-	archiveFile, err := afero.ReadFile(fs, fileName)
+	archiveFile, err := afero.ReadFile(fileSystem, fileName)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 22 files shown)


cmd/archive_test.go
```diff
@@ -155,10 +156,10 @@ func TestArchiveNotContainsEnv(t *testing.T) {
 }
 
 // untar untars a `fileName` file to a `destination` path
-func untar(t *testing.T, fs afero.Fs, fileName string, destination string) error {
+func untar(t *testing.T, filesys afero.Fs, fileName string, destination string) error {
```
oleiade, line 159: absolutely non-blocking nit: `filesys` feels neither more accurate than `fs`, and not as explicit as `fileSystem`, the first word is not shortened.

 I don't remember where I learned this practice from, but in general, I would suggest keeping the parameter name as short as possible, or an acronym when the type itself is already super explicit: `fileSystem afero.FS` would feel kinda redundant here, and could be replaced by `f afero.FS` (although the previous `fs` would also work).

oleiade (dismissed):
LGTM 👏🏻 

I'm happy to see this change in the linter, thanks for doing it 🙇🏻 I left a naming nit pick, but feel free to disregard it if you disagree 👍🏻 🚀

mstoykov:
Maybe we can also fix https://github.com/grafana/k6/issues/2565 if it isn't too much work

cmd/archive_test.go
na--: The reason I had to rename this is that we now have a package with the `fs` name, https://pkg.go.dev/io/fs, so `fs afero.Fs` is no longer an option. Using just `f` will be too ambiguous, since I'd expect it'd represent just a single "file" or something like that. However, while `fileSystem` is a bit too long, I am fine with using it instead of `filesys` if other people also prefer it.

codebien: Generally speaking I would prefer `filesys`, but it sounds a very good candidate for a package name, so in this case `fileSystem` could improve the readability removing ambiguity. :+1: from me for renaming.

codebien (approved):
LGTM, it is very useful to me :bow:

mstoykov (approved):

cmd/archive_test.go
mstoykov: Maybe `fileSys` so it is still shorter but it makes it more obvious it is two words.

But I am fine with `fileSystem` as well

mstoykov (approved):

na--:
Now it's ready for the final reviews :sweat_smile:

mstoykov (approved):

codebien (approved):

PR #2892, opened 2023-02-01
Update ui_unix.go with z/OS build flags

### LINKED ISSUE #2888: Compatibility with z/OS Unix
### Feature Description

Golang is now supported on IBM z/OS, which allows Golang code to be compiled and run on a Unix System Services shell (i.e. z/OS's version of Unix). This also means that existing Golang programs like k6 can be ported to z/OS and run natively on the platform. k6 would be an especially valuable tool for load testing APIs on real z/OS hardware. 

I am interested in providing a compatibility update through updating specific build flags so that k6 can be built on z/OS "out of the box". 

An example: https://github.com/spf13/afero/pull/384
Here, I have submitted a PR to add `zos` build flags to afero, allowing it to be installed on z/OS. 

I would also be glad to follow up on other updates needed so that k6 could be considered 'officially' supported on z/OS Unix.

I have already tested a simple compatibility fix by updating `cmd/ui_unix.go`
https://github.com/msradam/k6/blob/feature/zos-port/cmd/ui_unix.go

Along with the afero changes, this update allows k6 to run on z/OS. I would be glad to open a pull request, or make additional changes if necessary.

#2892

msradam:
Added z/OS build flags for IBM z/OS compatibility.

Enabling these build flags (along with the changes to afero mentioned here https://github.com/spf13/afero/pull/384) allows k6 to be compiled on z/OS Unix System Services. 

Closes #2888

### DIFF
--- a/cmd/ui_unix.go
+++ b/cmd/ui_unix.go
@@ -1,5 +1,5 @@
-//go:build darwin || dragonfly || freebsd || linux || netbsd || openbsd
-// +build darwin dragonfly freebsd linux netbsd openbsd
+//go:build darwin || dragonfly || freebsd || linux || netbsd || openbsd || zos
+// +build darwin dragonfly freebsd linux netbsd openbsd zos
 
 package cmd
 


mstoykov (approved):
LGTM!

This seems pretty okay change, but the afero one will be more problematic as we have *deliberately* not updated it in quite a while and intend on actually deprecating it with https://github.com/grafana/k6/issues/1079 .

So while we should likely still merge this - it might not be enough and even if afero merges the PR to their repo, we are unlikely to update it just to be z/OS compatibile :(.

codebien (approved):
LGTM

PR #2894, opened 2023-02-02
Add tests ensuring the tracing module sets and cleans up metadata properly

### LINKED ISSUE #2889: Write a test that the metadata gets reset with the tracing API
See https://github.com/grafana/k6/pull/2877#discussion_r1092943793 for more info

oleiade:
This Pull Request addresses #2889, and adds tests ensuring that the tracing module does add a `trace_id` metadata entry when performing a request and cleans it up afterward.

To make it possible, I had to wrap a bunch of existing behaviors throughout the module in interfaces. That way I could easily swap them with mocked/controllable behaviors: trace id generation and randomness generation.

Let me know what you think 🤝 

closes #2889

### DIFF
--- a/js/modules/k6/experimental/tracing/client.go
+++ b/js/modules/k6/experimental/tracing/client.go
@@ -2,6 +2,7 @@ package tracing
 
 import (
 	"fmt"
+	"math/rand"
 	"net/http"
 	"time"
 
@@ -36,13 +37,19 @@ type Client struct {
 	// uses it under the hood to emit the requests it
 	// instruments.
 	asyncRequestFunc HTTPAsyncRequestFunc
+
+	// randSource holds the client's random source, used
+	// to generate random values for the trace ID.
+	randSource *rand.Rand
 }
 
 type (
-	// HTTPRequestFunc is a type alias representing the prototype of k6's http module's request function
+	// HTTPRequestFunc is a type alias representing the prototype of
+	// k6's http module's request function
 	HTTPRequestFunc func(method string, url goja.Value, args ...goja.Value) (*httpmodule.Response, error)
 
-	// HTTPAsyncRequestFunc is a type alias representing the prototype of k6's http module's asyncRequest function
+	// HTTPAsyncRequestFunc is a type alias representing the prototype of
+	// k6's http module's asyncRequest function
 	HTTPAsyncRequestFunc func(method string, url goja.Value, args ...goja.Value) (*goja.Promise, error)
 )
 
@@ -64,6 +71,7 @@ func NewClient(vu modules.VU, opts options) (*Client, error) {
 		return nil,
 			fmt.Errorf("failed initializing tracing client, unable to require http.request method; reason: %w", err)
 	}
+
 	// Export the http module's syncRequest function goja.Callable as a Go function
 	var asyncRequestFunc HTTPAsyncRequestFunc
 	if err := rt.ExportTo(httpModuleObject.Get("asyncRequest"), &asyncRequestFunc); err != nil {
@@ -72,7 +80,13 @@ func NewClient(vu modules.VU, opts options) (*Client, error) {
 				err)
 	}
 
-	client := &Client{vu: vu, requestFunc: requestFunc, asyncRequestFunc: asyncRequestFunc}
+	client := &Client{
+		vu:               vu,
+		requestFunc:      requestFunc,
+		asyncRequestFunc: asyncRequestFunc,
+		randSource:       rand.New(rand.NewSource(time.Now().UnixNano())), //nolint:gosec
+	}
+
 	if err := client.Configure(opts); err != nil {
 		return nil,
 			fmt.Errorf("failed initializing tracing client, invalid configuration; reason: %w", err)
@@ -106,31 +120,11 @@ func (c *Client) Configure(opts options) error {
 	return nil
 }
 
-func (c *Client) generateTraceContext() (http.Header, string, error) {
-	traceID := TraceID{
-		Prefix: k6Prefix,
-		Code:   k6CloudCode,
-		Time:   time.Now(),
-	}
-
-	encodedTraceID, err := traceID.Encode()
-	if err != nil {
-		return http.Header{}, "", fmt.Errorf("failed to encode the generated trace ID; reason: %w", err)
-	}
-
-	// Produce a trace header in the format defined by the configured propagator.
-	traceContextHeader, err := c.propagator.Propagate(encodedTraceID)
-	if err != nil {
-		return http.Header{}, "", fmt.Errorf("failed to propagate trace ID; reason: %w", err)
-	}
-
-	return traceContextHeader, encodedTraceID, nil
-}
-
 // Request instruments the http module's request function with tracing headers,
 // and ensures the trace_id is emitted as part of the output's data points metadata.
 func (c *Client) Request(method string, url goja.Value, args ...goja.Value) (*httpmodule.Response, error) {
 	var result *httpmodule.Response
+
 	var err error
 	err = c.instrumentedCall(func(args ...goja.Value) error {
 		result, err = c.requestFunc(method, url, args...)
@@ -157,37 +151,6 @@ func (c *Client) AsyncRequest(method string, url goja.Value, args ...goja.Value)
 		return nil, err
 	}
 	return result, nil
-}
-
-func (c *Client) instrumentedCall(call func(args ...goja.Value) error, args ...goja.Value) error {
-	if len(args) == 0 {
-		args = []goja.Value{goja.Null()}
-	}
-
-	traceContextHeader, encodedTraceID, err := c.generateTraceContext()
-	if err != nil {
-		return err
-	}
-	// update the `params` argument with the trace context header
-	// so that it can be used by the http module's request function.
-	args, err = c.instrumentArguments(traceContextHeader, args...)
-	if err != nil {
-		return fmt.Errorf("failed to instrument request arguments; reason: %w", err)
-	}
-
-	// Add the trace ID to the VU's state, so that it can be
-	// used in the metrics emitted by the HTTP module.
-	c.vu.State().Tags.Modify(func(t *metrics.TagsAndMeta) {
-		t.SetMetadata(metadataTraceIDKeyName, encodedTraceID)
-	})
-	// Remove the trace ID from the VU's state, so that it doesn't leak into other requests.
-	defer func() {
-		c.vu.State().Tags.Modify(func(t *metrics.TagsAndMeta) {
-			t.DeleteMetadata(metadataTraceIDKeyName)
-		})
-	}()
-
-	return call(args...)
 }
 
 // Del instruments the http module's delete method.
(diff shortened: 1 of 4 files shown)


js/modules/k6/experimental/tracing/trace_id.go
```diff
@@ -45,6 +46,23 @@ type TraceID struct {
...
+
+// NewTraceID produces a new TraceID with the given prefix, code and time.
+//
+// It sets the traceID randomness source to the `rand.Reader` as a default.
+func NewTraceID(prefix int16, code int8, t time.Time, randSource io.Reader) TraceID {
+	return TraceID{
+		Prefix:     prefix,
+		Code:       code,
+		Time:       t,
+		randSource: rand.Reader,
```
mstoykov, line 64: off-topic: Does the source need to be crypto secure?

mstoykov (dismissed):
LGTM!

I was thinking more of js code test that has 
```javascript
callToTestMetadata()
http.post(...)
callToTestMetadata() 
```

maybe something like what I do in the async group [code](https://github.com/grafana/k6/blob/8101130bf73b285f5305c450b64ad9e017002a3c/js/modules/k6/asyncGroup_test.go#L126) but with a bertter name, The script that is run is in test case sabove the line I linked.

Goja's author also does something similar in https://github.com/dop251/goja/blob/8a0df7296ebdd6b66ab2bb8230f9744dc8824825/func_test.go#L212

oleiade:
@mstoykov that's another clever way of doing it indeed 🤝 When looking at how to access the metadata before the callback is called, I didn't think of doing it from the JS indeed.

In this case, I find the unit test ends up being "lower-context", forcing me to make the code a bit more flexible by using interfaces. If you have a strong opinion towards making it in JS, I could give it a shot too. Otherwise, I would keep it as is as unit tests and the JS approach in mind for the next time. Let me know 😸

js/modules/k6/experimental/tracing/producer.go
```diff
@@ -0,0 +1,27 @@
...
+	"time"
+)
+
+// Producer is an interface for producing trace IDs.
+type Producer interface {
+	Produce() TraceID
+}
+
+// DefaultProducer is a default implementation of the Producer interface.
+type DefaultProducer struct{}
```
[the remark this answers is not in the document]
oleiade: Nope, you're right 👍🏻

js/modules/k6/experimental/tracing/trace_id.go
oleiade: As far as I know, it doesn't need to. I am switching to `math/rand.Rand` 👍🏻

codebien (approved):
LGTM

codebien:
I have re-run the test-tip check to verify it is just one of the flaky.

mstoykov (changes requested):
Having now gone through the code one more time and through https://github.com/grafana/k6/pull/2886 it seems like the new Producer only purpose is to make the current test possible.

There is also no "pruducer" abstraction or terminology within the tracing vocabulary as well. 

Given just how much code this adds and that I feel like it makes everything *harder* and the fact this all can be tested with just a single defintion of a helper js function as I said before - I would prefer we go that route and this  PR to only add tests instead of also refactor.

Additionally this will hopefully make this PR have less conflicts with the sampling one

oleiade:
@mstoykov Although I disagree with the perceived complexity introduced by the approach in this PR, I see your point, and I'm indeed open to doing it in JS.

After experimenting with your examples and experimenting with your approach, I want to make one of my findings explicit. Unless I missed something, as opposed to my current approach, testing in JS won't let me assert that an instrumented function follows the underlying steps of:
1. set the `trace_id` metadata
2. perform the instrumented operation
3. unset `trace_id` metadata
Simply because I don't have access to what happens under the hood (which I can do with the current relatively standard interfaces/mocks testing strategy).

I end up, however, asserting the higher level behavior of "performing instrumented operation X emits a `trace_id` metadata", and "performing instrumented operation Y right after emits a different `trace_id` value", which is more of an integration/e2e test, but with which I'm okay. Let me know if we have similar understandings and if that's satisfying from your perspective (or if I missed something 🤞🏻) 🙇🏻

mstoykov:
tl;dr I checked out the branch to test something and got carried away :( 


> perceived complexity introduced

The problem for me is that all the new changes exist to make it possible to test this (slightly) better. 

As you can see from my [try](https://github.com/grafana/k6/commit/64dd12e58432f7cc5bc899ffa74b2593c418f87a) at this the only thing I am not checking is the *concrete* value of the trace_id. 

I could be testing its format, but decided to not spend more time on it.

While I also worked on this I realized I also have an issue with the fact you are basically having a unit test, while I would argue that we need more of an integration test. One in which you do what the user will do and check that we do set the metadata correctly and then unset it after the operation. But doing this with actual JS code as the user will use it instead of what seems like a reimplementation of the actual code the user will call. 

Both have their place, but arguably we care more about the user experience instead of the underlying implementation details.

The "integration" test [here](https://github.com/grafana/k6/commit/5e5f4863d2424de191386419f346bb09a3203ae0) did take more fiddling in some parts due to the existence of two functions generating test setups, but mostly due to the requirements of the k6/http package so that it is *actually* functional. 

But even with that [both commits](https://github.com/grafana/k6/compare/experimental/tracing-test-metadata-wip?expand=1) are on par with just the code that is to add the Producers and add the mock ones. 

This also now lends itself to be expanded so we can add more tests directly in the package instead of adding it to actual integration tests https://github.com/grafana/k6/tree/master/cmd/tests

oleiade:
@mstoykov Thanks for the thorough examples, I see what you meant much better now indeed. I'm grateful for it 🙇🏻 

I hadn't consider the `instrumentCall` route to build an integration test, that's a good idea 👀 I'll integrate your approach in an integration test then indeed, and drop most of the interfaces/mocks I've added. 👍🏻

oleiade:
I've updated the PR according to your recommendations. I've integrated the tests you kindly put together Mihail, thanks again for that 🤝 I've also dropped the `Producer` abstraction.

codebien:
@oleiade the CI reported a panic

oleiade:
👍🏻 @codebien looking into it

codebien (approved):
LGTM. It seems you moved some functions, the next time would be helpful not to do it in the same commit you are doing the changes so it will be easier to review/skip.

js/modules/k6/experimental/tracing/trace_id.go
```diff
@@ -76,7 +94,7 @@ func (t TraceID) Encode() (string, error) {
 	// of available bytes left after writing the prefix, code and timestamp (index n)
 	// is filled with random bytes.
 	randomness := make([]byte, 16-n)
-	err := binary.Read(rand.Reader, binary.BigEndian, randomness)
+	err := binary.Read(t.randSource, binary.BigEndian, randomness)
```
mstoykov, line 97: This is more of a general comment. But it seems strange that calling `TraceID.Encode` multiple times will get you *different* values. 

Like it seems to me this should just basically be a single function instead of having code that, creates a struct and then calls `Encode` on it as the only thing it does.

js/modules/k6/experimental/tracing/client_test.go
```diff
@@ -144,6 +149,105 @@ func TestClientInstrumentArguments(t *testing.T) {
...
+	testCase := newTestCase(t)
+	testCase.testSetup.MoveToVUContext(&lib.State{
+		Tags: lib.NewVUStateTags(&metrics.TagSet{}),
+	})
+	testCase.client.propagator = NewW3CPropagator(NewAlwaysOnSampler())
+
+	callFn := func(args ...goja.Value) error {
+		gotMetadataTraceID, gotTraceIDKey := testCase.client.vu.State().Tags.GetCurrentValues().Metadata["trace_id"]
+		assert.True(t, gotTraceIDKey)
+		assert.NotEmpty(t, gotMetadataTraceID)
```
mstoykov, line 164: You went through the trouble of having this to be deterministic by setting a specific `randSource` and ... then you didn't test it is the specific value we expect :(

js/modules/k6/experimental/tracing/client_test.go
```diff
@@ -144,6 +149,105 @@ func TestClientInstrumentArguments(t *testing.T) {
...
+
+		return err
+	})
+	require.NoError(t, err)
+	close(samples)
+
+	var sampleRead bool
+	for sampleContainer := range samples {
+		for _, sample := range sampleContainer.GetSamples() {
+			require.NotEmpty(t, sample.Metadata["trace_id"])
```
mstoykov, line 244: Here again you could check what the value is, given that it is deterministc, but you don't

mstoykov (changes requested):
I do like htat we now do not use the global rand source - that was definitel a problem I didn't notice before, but it is a concurrency concern as I mentioned in another PR.

But you then go and do not actually check the value :(.

js/modules/k6/experimental/tracing/trace_id.go
oleiade: I've experimented around it. I have an idea of how to decouple trace id initialisation from encoding, but because the random part doesn't have a fixed size, it's gonna make things a bit more involved than I anticipated.

I'm already working on improving this in a separate PR. Resolving this for now, and stay tuned for the upcoming refactor PR 👍

js/modules/k6/experimental/tracing/trace_id.go
```diff
@@ -45,6 +46,23 @@ type TraceID struct {
...
+	// randSource holds the randomness source to use when encoding the
+	// trace-id. The `rand.Reader` should be your default pick. But
+	// you can replace it with a different source for testing purposes.
+	randSource io.Reader
+}
+
+// NewTraceID produces a new TraceID with the given prefix, code and time.
+//
+// It sets the traceID randomness source to the `rand.Reader` as a default.
+func NewTraceID(prefix int16, code int8, t time.Time, randSource io.Reader) TraceID {
```
codebien, line 59:
```suggestion
func NewTraceID(prefix int16, code int8, t time.Time) TraceID {
```
It sounds not used

codebien (changes requested):

js/modules/k6/experimental/tracing/trace_id.go
codebien: I see, this is part of https://github.com/grafana/k6/pull/3019

oleiade:
@mstoykov done 👍🏻

js/modules/k6/experimental/tracing/client.go
```diff
@@ -231,6 +194,61 @@ func (c *Client) Put(url goja.Value, args ...goja.Value) (*httpmodule.Response,
...
 }
 
+func (c *Client) instrumentedCall(call func(args ...goja.Value) error, args ...goja.Value) error {
+	if len(args) == 0 {
+		args = []goja.Value{goja.Null()}
+	}
+
+	traceContextHeader, encodedTraceID, err := c.generateTraceContext()
+	if err != nil {
+		return err
```
mstoykov, line 204: nit: this Pr would've been easier to code review if you didn't move this methods around in the codebase

oleiade: Fair enough 🙇‍♂️Sorry for that. I'm trying to get better at isolating these kind of changes, but still having a hard time.

PR #2896, opened 2023-02-02
Fix panic with setInterval in k6/ws

mstoykov:
The panic happens when interval between 0 and 1 is given.

The code that tested whether the interval was > 0 correctly converted to
floats first and then back to time. And then checked it in order to
prevent this very panic.

Unfortunately it then did the same calculation again but this time by
converting the interval directly to int before multiplying by
milliseconds. Which ends up rounding it to 0. Making the whole result
zero and leading to time.NewTicker to panic.

This was already done correctly for setTimeout when the same panic was
found there and the wrong conversation was added for setInterval.

Previous commit https://github.com/grafana/k6/commit/df3ad1bd371ac6d0bd2d5a4c10887b08f4374eba

### DIFF
--- a/js/modules/k6/ws/ws_test.go
+++ b/js/modules/k6/ws/ws_test.go
@@ -217,7 +217,7 @@ func TestSessionInterval(t *testing.T) {
 	assertSessionMetricsEmitted(t, metrics.GetBufferedSamples(test.samples), "", sr("WSBIN_URL/ws-echo"), statusProtocolSwitch, "")
 }
 
-func TestSessionBadInterval(t *testing.T) {
+func TestSessionNegativeInterval(t *testing.T) {
 	t.Parallel()
 	tb := httpmultibin.NewHTTPMultiBin(t)
 	sr := tb.Replacer.Replace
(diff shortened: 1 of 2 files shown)


js/modules/k6/ws/ws_test.go
```diff
@@ -236,6 +236,24 @@ func TestSessionBadInterval(t *testing.T) {
 	require.Contains(t, err.Error(), "setInterval requires a >0 timeout parameter, received -1.23 ")
 }
 
+func TestSessionIntervalSub1(t *testing.T) {
```
codebien, line 239:
```suggestion
func TestSocketSetInterval(t *testing.T) {
```
If the current name is meaningful then I'm not getting it.

codebien (changes requested):

oleiade (approved):

js/modules/k6/ws/ws_test.go
mstoykov: I kind of prefer to have it somehow mention that it is testing a very small interval as that was what was breaking. But am okay with a different name - I did not like this one either way, I just needed something to have a test to confirm what I was suspecting

codebien (approved):

PR #2898, opened 2023-02-03
Drop Format method

### LINKED ISSUE #2320: Remove the `Format` method of `stats.Sink`
## Proposal

During our discussions about [parsing and executing thresholds directly in Go](https://github.com/grafana/k6/pull/2251), @MStoykov [mentioned](https://github.com/grafana/k6/pull/2251#discussion_r769807203) that we could get rid of the `stats.Sink.Format` method all in all.

## Context

> I would also, looking at this code and the other places where sink.Format is used, I kind of want to remove this method. It's in most cases used as a way to get info about values for the given sink, but then we also cast the sink to its concrete type in order to get even more info from it.
Given that there are 4 sink types, I feel like a switch in all cases (as the fields might not be the same) might be better 🤔.
Especially as it turns out that the `CounterSink` does actually return a rate, but the summary code overwrites it, possibly with the same value 🤦
And in the REST API where it is used "correctly" we have a minor bug about `Trend`s and also doesn't show the same stuff as in the summary 🤦.

## Status

I only report this issue as it was mentioned during a somewhat related discussion. There must be some missing context, questions, and related issues that probably need mention and being addressed before moving on. Thus, this issue could use more input from historical maintainers. Don't hesitate to provide any information or feedback you judge necessary to the resolution of this, I will add them as edits to this description.

## Related 
- #2251 
- #1443 
- [comment in context](https://github.com/grafana/k6/pull/2251)

codebien:
Closes #2320

### DIFF
--- a/js/summary.go
+++ b/js/summary.go
@@ -31,20 +31,26 @@ func metricValueGetter(summaryTrendStats []string) func(metrics.Sink, time.Durat
 	return func(sink metrics.Sink, t time.Duration) (result map[string]float64) {
 		switch sink := sink.(type) {
 		case *metrics.CounterSink:
-			result = sink.Format(t)
+			result = map[string]float64{
+				"count": sink.Value,
+			}
 			rate := 0.0
 			if t > 0 {
-				rate = sink.Value / (float64(t) / float64(time.Second))
+				rate = sink.Rate(t)
 			}
 			result["rate"] = rate
 		case *metrics.GaugeSink:
-			result = sink.Format(t)
-			result["min"] = sink.Min
-			result["max"] = sink.Max
+			result = map[string]float64{
+				"value": sink.Value,
+				"min":   sink.Min,
+				"max":   sink.Max,
+			}
 		case *metrics.RateSink:
-			result = sink.Format(t)
-			result["passes"] = float64(sink.Trues)
-			result["fails"] = float64(sink.Total - sink.Trues)
+			result = map[string]float64{
+				"rate":   sink.Rate(),
+				"passes": float64(sink.Trues),
+				"fails":  float64(sink.Total - sink.Trues),
+			}
 		case *metrics.TrendSink:
 			result = make(map[string]float64, len(summaryTrendStats))
 			for _, col := range summaryTrendStats {
--- a/js/summary.go
+++ b/js/summary.go
@@ -14,6 +14,7 @@ import (
 )
 
 // Copied from https://github.com/k6io/jslib.k6.io/tree/master/lib/k6-summary
+//
 //go:embed summary.js
 var jslibSummaryCode string //nolint:gochecknoglobals
 
@@ -32,24 +33,20 @@ func metricValueGetter(summaryTrendStats []string) func(metrics.Sink, time.Durat
 		switch sink := sink.(type) {
 		case *metrics.CounterSink:
 			result = map[string]float64{
-				"count": sink.Value,
+				"count": sink.LastValue(),
 			}
-			rate := 0.0
-			if t > 0 {
-				rate = sink.Rate(t)
-			}
-			result["rate"] = rate
+			result["rate"] = sink.Rate(t)
 		case *metrics.GaugeSink:
 			result = map[string]float64{
-				"value": sink.Value,
-				"min":   sink.Min,
-				"max":   sink.Max,
+				"value": sink.LastValue(),
+				"min":   sink.Min(),
+				"max":   sink.Max(),
 			}
 		case *metrics.RateSink:
 			result = map[string]float64{
 				"rate":   sink.Rate(),
-				"passes": float64(sink.Trues),
-				"fails":  float64(sink.Total - sink.Trues),
+				"passes": float64(sink.NotZero()),
+				"fails":  float64(sink.Count() - sink.NotZero()),
 			}
 		case *metrics.TrendSink:
 			result = make(map[string]float64, len(summaryTrendStats))
--- a/metrics/sink.go
+++ b/metrics/sink.go
@@ -15,10 +15,13 @@ var (
 	_ Sink = &DummySink{}
 )
 
+// Sink collects Sample's values.
 type Sink interface {
-	Add(s Sample)                              // Add a sample to the sink.
-	Format(t time.Duration) map[string]float64 // Data for thresholds.
-	IsEmpty() bool                             // Check if the Sink is empty.
+	// Add adds a sample to the sink.
+	Add(s Sample)
+
+	// IsEmpty checks if the Sink has never received a value.
+	IsEmpty() bool
 }
 
 type CounterSink struct {
@@ -36,11 +39,12 @@ func (c *CounterSink) Add(s Sample) {
 // IsEmpty indicates whether the CounterSink is empty.
 func (c *CounterSink) IsEmpty() bool { return c.First.IsZero() }
 
-func (c *CounterSink) Format(t time.Duration) map[string]float64 {
-	return map[string]float64{
-		"count": c.Value,
-		"rate":  c.Value / (float64(t) / float64(time.Second)),
+// Rate computes the rate per second of the aggregated values.
+func (c *CounterSink) Rate(t time.Duration) float64 {
+	if t == 0 {
+		return 0
 	}
+	return c.Value / (float64(t) / float64(time.Second))
 }
 
 type GaugeSink struct {
@@ -61,10 +65,6 @@ func (g *GaugeSink) Add(s Sample) {
 		g.Min = s.Value
 		g.minSet = true
 	}
-}
-
-func (g *GaugeSink) Format(t time.Duration) map[string]float64 {
-	return map[string]float64{"value": g.Value}
 }
 
 type TrendSink struct {
@@ -118,18 +118,6 @@ func (t *TrendSink) P(pct float64) float64 {
 	}
 }
 
-func (t *TrendSink) Format(tt time.Duration) map[string]float64 {
-	// TODO: respect the summaryTrendStats for REST API
-	return map[string]float64{
-		"min":   t.Min,
-		"max":   t.Max,
-		"avg":   t.Avg,
-		"med":   t.P(0.5),
-		"p(90)": t.P(0.90),
-		"p(95)": t.P(0.95),
-	}
-}
-
 type RateSink struct {
 	Trues int64
 	Total int64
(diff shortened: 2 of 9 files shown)


vendor/github.com/grafana/xk6-output-prometheus-remote/pkg/remotewrite/remotewrite.go
```diff
@@ -340,8 +340,7 @@ func (swm seriesWithMeasure) MapPrompb() []*prompb.TimeSeries {
 	case metrics.Rate:
 		ts := mapMonoSeries(swm.TimeSeries, "rate", swm.Latest)
 		// pass zero duration here because time is useless for formatting rate
-		rateVals := swm.Measure.(*metrics.RateSink).Format(time.Duration(0))
-		ts.Samples[0].Value = rateVals["rate"]
+		ts.Samples[0].Value = swm.Measure.(*metrics.RateSink).Rate()
```
codebien, line 343: It requires a change in the extension, I will do after the v0.43.0 release

metrics/sink.go
```diff
@@ -36,11 +35,9 @@ func (c *CounterSink) Add(s Sample) {
...
 func (c *CounterSink) IsEmpty() bool { return c.First.IsZero() }
 
-func (c *CounterSink) Format(t time.Duration) map[string]float64 {
-	return map[string]float64{
-		"count": c.Value,
-		"rate":  c.Value / (float64(t) / float64(time.Second)),
-	}
+// Rate computes the rate per second of the aggregated values.
+func (c *CounterSink) Rate(t time.Duration) float64 {
+	return c.Value / (float64(t) / float64(time.Second))
```
oleiade, line 40: Should we be careful with division by zero here? If `t == 0`, you get `c.Value / (0 / time.Second)`, we'll get a `panic: runtime error: integer divide by zero`. We could check for `t == 0`, and if that's the case, return `t`?

I know that we check it one level higher (I think I did introduce that after it introduced a bug in a previous version). But seeing it isolated now, I do think it would be worth having that check locally too.

 What do you think?

metrics/sink_test.go
```diff
@@ -5,198 +5,188 @@ import (
...
 			sink.Add(Sample{TimeSeries: TimeSeries{Metric: &Metric{}}, Value: s, Time: now})
 		}
-		assert.Equal(t, map[string]float64{"count": 145, "rate": 145.0}, sink.Format(1*time.Second))
+		assert.Equal(t, 145.0, sink.Value)
+		assert.Equal(t, now, sink.First)
 	})
 }
 
-func TestGaugeSink(t *testing.T) {
+func TestCounterSinkRate(t *testing.T) {
```
oleiade, line 33: If you do think it's worth avoiding division by zero, as per my previous comment, could we add a test dedicated to that scenario here?

metrics/sink.go
```diff
@@ -16,9 +16,8 @@ var (
 )
 
 type Sink interface {
-	Add(s Sample)                              // Add a sample to the sink.
-	Format(t time.Duration) map[string]float64 // Data for thresholds.
-	IsEmpty() bool                             // Check if the Sink is empty.
+	Add(s Sample)  // Add a sample to the sink.
```
oleiade, line 19: Can we use this opportunity to have standard godoc comments for those methods? 😸

oleiade (commented):
Great job, I'm glad we picked this up 👍🏻 🙇🏻 

Looks good to me, still I've left a few comments about potential division by zero that I'd like to at least see considered 😄

oleiade (approved):
🚀 🚀

olegbespalov (approved):

na-- (changes requested):
In principle, I am fine with dropping the `Sink.Format()` method. However, being able to access the internal thread-**un**safe `Sink` state directly anywhere in the codebase may be an even bigger problem than `Format()` :disappointed: That requires [a global lock on the whole `MetricsEngine` data crunching](https://github.com/grafana/k6/blob/a78916c644a4ab741723db5d2050725163a569fe/metrics/engine/engine.go#L37-L42) and prevents us from making performance improvements there or being able to change how sinks are implemented, e.g. for https://github.com/grafana/k6/issues/763 or https://github.com/grafana/k6/issues/140

This way of dropping `Format()` seems to make the problem even worse than it currently is :disappointed: I don't have a better alternative solution to offer at the moment, and this PR may be a necessary intermediate step to such a solution, but we need to discuss and consider this and https://github.com/grafana/k6/pull/2905 more carefully before we merge either.

codebien:
@na-- I pushed the changes for moving the access from fields to getter methods.

Something like the following pseudo-code should work but considering the usage we have at the moment, a more complex architecture is overengineering except if someone else has a good but simpler solution.

```go
type QueryArguments struct {
	StaticQueryFunction enum (max, min, count, ...)
	DynamicQueryFunction struct {
	    percentile float64
            rate time.Duration
        }
}

func Percentile(p float64) QueryArguments {
	return QueryArguments{
		Dynamic{percentile: p}
	}
}

func Max() QueryArguments {
        return QueryArguments{
		Static: Max, // enum item
	}
}

// Sink collects Sample's values.
type Sink interface {
	// Add adds a sample to the sink.
	Add(s Sample)

	// Query resolves the request function
	Query(QueryArguments) (float64, error)

	// IsEmpty checks if the Sink has never received a value.
	IsEmpty() bool
}

func (trendSink) Query(qargs QueryArguments) (float64, error) {
	if StaticQueryFunction != Undefined {
		switch StaticQuery {
		case Max:
			// return max
		case Avg:
			// return avg
               default:
                       // unsupported, return error
		}
		return
	}

	if qargs.DynamicQueryFunction != nil {
		return trendSink.P(qargs.percentile)
	}
}

--- main.go

maxValue, err := sink.Query(metrics.Max())
p99, err := sink.Query(metrics.Percentile(0.99))
```

metrics/sink.go
```diff
@@ -15,143 +15,186 @@ var (
...
-	if c.First.IsZero() {
-		c.First = s.Time
+	c.value += s.Value
+	if c.first.IsZero() {
+		c.first = s.Time
 	}
 }
 
+// LastValue returns the counter value.
+func (c *CounterSink) LastValue() float64 {
```
na--, line 40: This shouldn't be called `LastValue`, that only makes sense for the `Gauge` metrics. Either `Value()` or `Total()` makes more sense.

js/summary.go
```diff
@@ -31,20 +32,22 @@ func metricValueGetter(summaryTrendStats []string) func(metrics.Sink, time.Durat
 	return func(sink metrics.Sink, t time.Duration) (result map[string]float64) {
 		switch sink := sink.(type) {
 		case *metrics.CounterSink:
-			result = sink.Format(t)
-			rate := 0.0
-			if t > 0 {
-				rate = sink.Value / (float64(t) / float64(time.Second))
+			result = map[string]float64{
+				"count": sink.LastValue(),
```
na--, line 36: why not also assign `rate` here, instead of manually assigning `result["rate"]` on the line below?

metrics/sink.go
```diff
@@ -15,143 +15,186 @@ var (
...
 // IsEmpty indicates whether the CounterSink is empty.
-func (c *CounterSink) IsEmpty() bool { return c.First.IsZero() }
+func (c *CounterSink) IsEmpty() bool { return c.first.IsZero() }
 
-func (c *CounterSink) Format(t time.Duration) map[string]float64 {
-	return map[string]float64{
-		"count": c.Value,
-		"rate":  c.Value / (float64(t) / float64(time.Second)),
+// Rate computes the rate per second of the aggregated values.
+func (c *CounterSink) Rate(t time.Duration) float64 {
```
na--, line 48: It seems like it should be a helper function, not a method of the `Counter` type :thinking: 

Most of the complexity here is in dealing with the external parameter `t`, which has nothing to do with the `value`. Everything it needs from the `CounterSink` it can get from its public methods.

metrics/sink.go
```diff
@@ -15,143 +15,186 @@ var (
...
+	return uint64(r.total)
+}
 
-	return map[string]float64{"rate": rate}
+// NotZero returns the non zero values observed.
+func (r *RateSink) NotZero() uint64 {
+	return uint64(r.trues)
 }
 
 type DummySink map[string]float64
```
na--, line 200: Do we even need this? I don't think it provides any value...

na-- (commented):
LGTM in general, though there are some nitpicks and a few things to further clean up that I left as inline comments.

metrics/sink.go
```diff
@@ -15,143 +15,186 @@ var (
 	_ Sink = &DummySink{}
 )
 
+// Sink collects Sample's values.
 type Sink interface {
```
na--, line 19: @codebien, I'll start a thread here to discuss [your proposal](https://github.com/grafana/k6/pull/2898#issuecomment-1484811792) for a more generic `Query` and `QueryArguments`, since I can't reply to it without a new top-level PR comment.

I don't particularly like the current `Sink` APIs and I am open to improvement suggestions. However, I am afraid that what you are proposing will result in even more verbose code :disappointed: Particularly having to deal with error handling everywhere instead of type assertions seems like a worse trade-off, e.g. if you `Query()` for things that a `Sink` doesn't have. Or maybe I am misunderstanding something?

mstoykov: At one point I remember as discussing basically dropping the Sink API?

The current api stats call, the summary and the thresholds are basically three "outputs" (for a very generaous definition of outputs) that put stuff in Sink and then use it.

There really isn't a good reason why they should use the same abstraction (or to use different ones if they want to). And even less reason for that abstraction to be in the `metric` package and being part of the metric. 

It might still be beneficial that they have a shared store that the metric samples get "sinked" to. But IMO having that be less generic solution will help a lot.

Arguably the current problem is that the 4 different types of metric sinks are combined under one interface and we try to keep them like that. While if the Store just stores Trends, Coutners, etc. as separate things we might completely circumvent this problem

na--: Interesting ideas... :thinking: First, a small correction:
> The current api stats call, the summary and the thresholds are basically three "outputs" (for a very generaous definition of outputs) that put stuff in Sink and then use it.

They are in fact a single output. And, after some of my recent refactoring, the current `MetricsIngester` for all three pretends to be an actual `output.Output`: https://github.com/grafana/k6/blob/124fc7919049803cbf404b345f3b67fed0a57ad8/cmd/run.go#L128-L132

> There really isn't a good reason why they should use the same abstraction (or to use different ones if they want to). 

I am not sure, it seems to me like the [simple `Sink` interface](https://github.com/grafana/k6/blob/83963aec4aaa39cf5c26d0f24f42b64c16ab5ca9/metrics/sink.go#L18-L25) from this PR is should simplify Adding values, since that part is universal. Querying the sinks is the different part, which I am not sure how can be made simpler than this :thinking: Can you elaborate?

> And even less reason for that abstraction to be in the metric package and being part of the metric.

This I fully agree with. A follow-up PR after this one should split away the `metrics.Metric` along the lines of the `TODO`: https://github.com/grafana/k6/blob/124fc7919049803cbf404b345f3b67fed0a57ad8/metrics/metric.go#L11-L26

Everything after it should be moved out of the `Metric` type and into various data structures in the `MetricsEngine`.

codebien:
Closing, as this is not happening at the moment and requires a new iteration. We will come up with a new re-iterated proposal.

PR #2900, opened 2023-02-03
WIP

mstoykov:
This is WIP 

TODO:
- [ ] tests
- [ ] give single warnings when the correct `require` implementation differs from the current one
- [ ] possibly do the same for some "sane" `open` implementation


mstoykov:
Already implemented in #3681

PR #2901, opened 2023-02-06
Too many time series warning

### LINKED ISSUE #2765: Add warning if there are too many time series
### Feature Description

Now that k6 v0.41.0 has been released with support for time series, k6 can run out of memory if there are many metric samples with high-cardinality tags. For example, millions of unique URLs (https://github.com/grafana/k6/issues/2762).

RAM usage and availability will vary between instances, but we can choose some fairly high time series count (e.g. 100k, 1m?) that should rarely be reached and emit a warning when it is. In the majority of cases, this will be a problem that the user might not realize.

Ideally, it should be configurable, so the few users that want to intentionally exceed it, can do so without that warning always triggering. However, because of #883, that part might be left for the future. Having the warning sooner rather than later is probably better than a few people seeing an unnecessary warning :thinking: 

- https://github.com/grafana/k6/issues/2762
- https://github.com/grafana/k6/issues/1889
- https://github.com/grafana/k6/issues/572
- https://github.com/grafana/k6/issues/2735

codebien:
Closes #2765

At the moment the implementation doesn't support turning it off, we may consider using 0 for it in the future.

### DIFF
--- a/lib/testutils/logrus_hook.go
+++ b/lib/testutils/logrus_hook.go
@@ -1,6 +1,7 @@
 package testutils
 
 import (
+	"io"
 	"strings"
 	"sync"
 
@@ -58,4 +59,19 @@ func LogContains(logEntries []logrus.Entry, expLevel logrus.Level, expContents s
 		}
 	}
 	return false
+}
+
+// NewMemLogger creates a Logrus logger mocked by SimpleLogrusHook.
+func NewMemLogger(levels ...logrus.Level) (*logrus.Logger, *SimpleLogrusHook) {
+	if len(levels) == 0 {
+		levels = logrus.AllLevels
+	}
+	logger := logrus.New()
+	logger.SetLevel(logrus.InfoLevel)
+	logger.Out = io.Discard
+	hook := &SimpleLogrusHook{
+		HookedLevels: levels,
+	}
+	logger.AddHook(hook)
+	return logger, hook
 }
--- a/metrics/engine/ingester.go
+++ b/metrics/engine/ingester.go
@@ -4,10 +4,14 @@ import (
 	"time"
 
 	"github.com/sirupsen/logrus"
+	"go.k6.io/k6/metrics"
 	"go.k6.io/k6/output"
 )
 
-const collectRate = 50 * time.Millisecond
+const (
+	collectRate          = 50 * time.Millisecond
+	timeSeriesFirstLimit = 100_000
+)
 
 var _ output.Output = &outputIngester{}
 
@@ -24,6 +28,7 @@ type outputIngester struct {
 
 	metricsEngine   *MetricsEngine
 	periodicFlusher *output.PeriodicFlusher
+	cardinality     *cardinalityControl
 }
 
 // Description returns a human-readable description of the output.
@@ -92,6 +97,60 @@ func (oi *outputIngester) flushMetrics() {
 				oi.metricsEngine.markObserved(sm.Metric)
 				sm.Metric.Sink.Add(sample)
 			}
+
+			oi.cardinality.Add(sample.TimeSeries)
 		}
 	}
+
+	if oi.cardinality.LimitHit() {
+		// TODO: suggest using the Metadata API as an alternative, once it's
+		// available (e.g. move high-cardinality tags as Metadata)
+		// https://github.com/grafana/k6/issues/2766
+
+		oi.logger.Warnf(
+			"The test has generated metrics with %d unique time series, "+
+				"which is higher than the suggested limit of %d "+
+				"and could cause high memory usage. "+
+				"Consider not using high-cardinality values like unique IDs as metric tags "+
+				"or, if you need them in the URL, use the name metric tag or URL grouping. "+
+				"See https://k6.io/docs/using-k6/tags-and-groups for details.", oi.cardinality.Count(), timeSeriesFirstLimit)
+	}
+}
+
+type cardinalityControl struct {
+	seen            map[metrics.TimeSeries]struct{}
+	timeSeriesLimit int
+}
+
+func newCardinalityControl() *cardinalityControl {
+	return &cardinalityControl{
+		timeSeriesLimit: timeSeriesFirstLimit,
+		seen:            make(map[metrics.TimeSeries]struct{}),
+	}
+}
+
+// Add adds the passed time series to the list of seen items.
+func (cc *cardinalityControl) Add(ts metrics.TimeSeries) {
+	if _, ok := cc.seen[ts]; ok {
+		return
+	}
+	cc.seen[ts] = struct{}{}
+}
+
+// LimitHit checks if the cardinality limit has been hit.
+func (cc *cardinalityControl) LimitHit() bool {
+	if len(cc.seen) <= cc.timeSeriesLimit {
+		return false
+	}
+
+	// we don't care about overflow
+	// the process should be already OOM
+	// if the number of generated time series goes higher than N-hundred-million(s).
+	cc.timeSeriesLimit *= 2
+	return true
+}
+
+// Count returns the number of distinct seen time series.
+func (cc *cardinalityControl) Count() int {
+	return len(cc.seen)
 }
--- a/metrics/engine/ingester_test.go
+++ b/metrics/engine/ingester_test.go
@@ -1,6 +1,7 @@
 package engine
 
 import (
+	"strconv"
 	"testing"
 
 	"github.com/sirupsen/logrus"
@@ -23,6 +24,7 @@ func TestIngesterOutputFlushMetrics(t *testing.T) {
 		metricsEngine: &MetricsEngine{
 			ObservedMetrics: make(map[string]*metrics.Metric),
 		},
+		cardinality: newCardinalityControl(),
 	}
 	require.NoError(t, ingester.Start())
 	ingester.AddMetricSamples([]metrics.SampleContainer{metrics.Sample{
@@ -67,6 +69,7 @@ func TestIngesterOutputFlushSubmetrics(t *testing.T) {
 	ingester := outputIngester{
 		logger:        piState.Logger,
 		metricsEngine: me,
+		cardinality:   newCardinalityControl(),
 	}
 	require.NoError(t, ingester.Start())
 	ingester.AddMetricSamples([]metrics.SampleContainer{metrics.Sample{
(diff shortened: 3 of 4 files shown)


metrics/engine/ingester.go
```diff
@@ -87,6 +91,31 @@ func (oi *outputIngester) flushMetrics() {
...
 		}
 	}
+
+	oi.warnTimeSeriesLimit()
+}
+
+func (oi *outputIngester) warnTimeSeriesLimit() {
+	if len(oi.seenTimeSeries) <= oi.timeSeriesLimit {
+		return
+	}
```
imiric, line 107: Can you move this check outside of the method?

I.e. this method should only be called if the limit was exceeded. This pattern of immediately returning avoids repeating the same check if the function is called in several places, but it has the bigger drawback of not being clear when reading the calling code. The reader needs to be aware that this is _maybe_ logging a warning.

metrics/engine/ingester.go
```diff
@@ -87,6 +91,31 @@ func (oi *outputIngester) flushMetrics() {
...
+
+	// TODO: suggest to use as an alternative option the Metadata API
+	// when it will be available (e.g. move high-cardinality tags as Metadata)
```
imiric, line 110:
```suggestion
	// TODO: suggest using the Metadata API as an alternative, once it's
	// available (e.g. move high-cardinality tags as Metadata)
```

metrics/engine/ingester.go
```diff
@@ -87,6 +91,31 @@ func (oi *outputIngester) flushMetrics() {
...
+	oi.logger.Warnf("The test has generated %d time series, it is higher than suggested limit value, "+
+		"it could generate a high memory usage. "+
+		"You may consider to not use high-cardinality values like IDs as tags "+
+		"or if you need them in the URL then use the URL-grouping feature. "+
+		"Check out https://k6.io/docs/using-k6/tags-and-groups documentation.", len(oi.seenTimeSeries))
```
imiric, line 115:
```suggestion
	oi.logger.Warnf("The test has generated %d time series, which is higher than the suggested limit (%d), "+
		"and could cause high memory usage. "+
		"Consider not using high-cardinality values like IDs as tags "+
		"or if you need them in the URL then use the URL-grouping feature. "+
		"See https://k6.io/docs/using-k6/tags-and-groups for details.", oi.timeSeriesLimit, len(oi.seenTimeSeries))
```

It's not properly formatted, so don't just accept the suggestion as is.

metrics/engine/ingester_test.go
```diff
@@ -96,6 +98,36 @@ func TestIngesterOutputFlushSubmetrics(t *testing.T) {
...
+	ingester := outputIngester{
+		logger:          logger,
+		seenTimeSeries:  make(map[metrics.TimeSeries]struct{}),
+		timeSeriesLimit: 1,
+	}
+	ingester.seenTimeSeries[metrics.TimeSeries{
+		Metric: testMetric,
+		Tags:   piState.Registry.RootTagSet().With("a", "2"),
+	}] = struct{}{}
+	ingester.warnTimeSeriesLimit()
```
imiric, line 119: I think this test is OK as a `warnTimeSeriesLimit()` unit test, but you're not testing that `flushMetrics()` will actually run this check and emit the warning.

So I'd rather see a higher-level test at the output level, but feel free to keep this one as well.

imiric (changes requested):
Nice work :+1:

I have some suggestions, particularly around testing.

metrics/engine/ingester.go
```diff
@@ -87,6 +91,31 @@ func (oi *outputIngester) flushMetrics() {
 				oi.metricsEngine.markObserved(sm.Metric)
 				sm.Metric.Sink.Add(sample)
 			}
+
+			if _, ok := oi.seenTimeSeries[sample.TimeSeries]; !ok {
+				oi.seenTimeSeries[sample.TimeSeries] = struct{}{}
```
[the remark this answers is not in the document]
na--: I wonder if we can substitute this with a [HyperLogLog](https://en.wikipedia.org/wiki/HyperLogLog) and get good enough results :thinking: 

We probably don't care about counting to exactly 100k time series, right? Just about emitting a warning when some sufficiently high number of time series is reached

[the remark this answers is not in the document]
codebien: 100k should be in the range 1-5 MB of memory, I thought it was fine for a first implementation. We can try an HyperLogLog implementation, it should decrease at least one order of magnitude.

codebien:
@imiric @na-- I pushed an HyperLogLog implementation, the last commit contains a comparison between a map and a potential HyperLogLog. The map is doing ~4x in B/op.

This is good but it has a cost, the first is the obvious complexity of the implementation, but also the CPU effort is different.

1. Complexity: The current hashing mechanism for the TimeSeries struct uses a hacky way to reuse the Go's hash function used by the map type which is very efficient. On the other side, this is fragile because it depends on Go's internals. In the case, we would adopt this solution we should probably reproduce directly in k6 the current hack that the external library is doing, so we can drop the new dependency. I haven't found another way to do this without relying on slow things.
In terms of HLL implementation I think we can have our internal implementation but for a first release should be good to rely on the external library.

3. CPUtime: The map solution is mostly invisible, instead `HLL.Add` makes an addition of ~150ns/op.

metrics/engine/ingester_test.go
imiric: [This](https://github.com/grafana/k6/blob/5c7058f0ddde0b3fab8dacca62fc68aa98acbd27/metrics/engine/ingester.go#L108-L117) is still not covered.

Can you add another test that calls `flushMetrics()` and confirms the warning is logged?

imiric (changes requested):
Nicely done! Almost LGTM, but I think we're missing a `flushMetrics()` test.

The HLL lib is a large dependency, but I agree that it's worth it in this case.

lib/testutils/logrus_hook.go
```diff
@@ -59,3 +60,18 @@ func LogContains(logEntries []logrus.Entry, expLevel logrus.Level, expContents s
 	}
 	return false
 }
+
+// NewMemLogger creates a Logrus logger mocked by SimpleLogrusHook.
+func NewMemLogger() (*logrus.Logger, *SimpleLogrusHook) {
```
imiric, line 65: This is not used anymore. Do we still want to keep it?

metrics/engine/ingester_test.go
```diff
@@ -107,3 +222,43 @@ func newTestPreInitState(tb testing.TB) *lib.TestPreInitState {
 		BuiltinMetrics: metrics.RegisterBuiltinMetrics(reg),
 	}
 }
+
+func cardinalityMap() {
+	m := make(map[uint64]struct{})
+	for i := 1; i <= 100000; i++ {
+		m[uint64(i)] = struct{}{}
```
na--, line 229: hmm I am not sure this benchmark accurately represents the reality :thinking: it creates a new `uint64` allocation here, but in our case, we would just be reusing the already existing atlas pointer, right? :thinking:

codebien: I'm using those benchmarks only for the B/op comparison. For ns/op or allocs/op these tests are not representative.

>hmm I am not sure this benchmark accurately represents the reality

In the case this is for B/op, why it shouldn't be realistic? I expect that a key of a pointer has mostly the same size as a uint64, because under the hood it should be a hash of the struct pointer. Am I missing something?

na--:
Sorry for the late review :disappointed: 

Honestly, I think I am leaning towards the map solution, despite being the one to propose HLL... :thinking:  For 100k elements, we get 4x the memory usage, but that is still only 3MB, and it's faster. For 1 million elements, we get a slightly better result, 50MB for a map vs 8MB for HLL (and HLL is a bit faster). But 50MB is still basically nothing, especially compared to the actual memory usage for the strings for these 1M time series, which will probably be several GB :sweat_smile: 

I don't see a situation where the memory usage of this counter would even remotely matter. If we were already using a HLL library for something else in the k6 codebase, it might be worth it to use it here as well. But I don't think it's worth the huge dependency just for this. Besides, the map solution provides us with an exact count of the time series, while the HLL variant will only have an approximate count, so its UX would be a bit worse.

TLDR: I vote for the map

metrics/engine/ingester_test.go
codebien: Added

lib/testutils/logrus_hook.go
codebien: The latest version uses it for checking the printed logline

metrics/engine/ingester.go
```diff
@@ -4,10 +4,14 @@ import (
...
 
 	"github.com/sirupsen/logrus"
+	"go.k6.io/k6/metrics"
 	"go.k6.io/k6/output"
 )
 
-const collectRate = 50 * time.Millisecond
+const (
+	collectRate          = 50 * time.Millisecond
+	timeSeriesFirstLimit = 100e3 // 100k
```
na--, line 13:
```suggestion
	timeSeriesFirstLimit = 100_000
```

this is a bit more readable

metrics/engine/ingester.go
```diff
@@ -92,6 +97,74 @@ func (oi *outputIngester) flushMetrics() {
...
+	if hit, limit := oi.cardinality.LimitHit(); hit {
+		// TODO: suggest using the Metadata API as an alternative, once it's
+		// available (e.g. move high-cardinality tags as Metadata)
+		// https://github.com/grafana/k6/issues/2766
+
+		oi.logger.Warnf("The test has generated %d time series, which is higher than the suggested limit (%d), "+
+			"and could cause high memory usage. "+
+			"Consider not using high-cardinality values like IDs as tags "+
+			"or if you need them in the URL then use the URL-grouping feature. "+
+			"See https://k6.io/docs/using-k6/tags-and-groups for details.", oi.cardinality.Count(), limit)
```
na--, line 114:
```suggestion
		oi.logger.Warnf(
			"The test has generated metrics with %d unique time series, "+
			"which is higher than the suggested limit of %d "+
			"and could cause high memory usage. "+
			"Consider not using high-cardinality values like unique IDs as metric tags "+
			"or, if you need them in the URL, use the name metric tag or URL grouping. "+
			"See https://k6.io/docs/using-k6/tags-and-groups for details.", oi.cardinality.Count(), timeSeriesFirstLimit)
```

Slightly edited the message and specified the limit to be `timeSeriesFirstLimit` - if we hit this message multiple times (e.g. at 100k, 200k, 400k), it doesn't make sense to me to have multiple warnings where we give a different value for the "suggested limit".

metrics/engine/ingester.go
```diff
@@ -92,6 +97,74 @@ func (oi *outputIngester) flushMetrics() {
...
+		timeSeriesLimit: timeSeriesFirstLimit,
+		addsCount:       0,
+		seen:            make(map[metrics.TimeSeries]struct{}),
+	}
+}
+
+// Add adds the passed time series to the list of seen items.
+func (cc *cardinalityControl) Add(ts metrics.TimeSeries) {
+	cc.addsCount++
+	cc.seen[ts] = struct{}{}
```
na--, line 135: hmm, shouldn't this be something like this:
```suggestion
	if _, ok := cc.seen[ts]; !ok {
		cc.addsCount++
		cc.seen[ts] = struct{}{}
	}
```

a test that we don't count the same time series twice would be nice too

metrics/engine/ingester.go
```diff
@@ -92,6 +97,74 @@ func (oi *outputIngester) flushMetrics() {
...
+			"Consider not using high-cardinality values like IDs as tags "+
+			"or if you need them in the URL then use the URL-grouping feature. "+
+			"See https://k6.io/docs/using-k6/tags-and-groups for details.", oi.cardinality.Count(), limit)
+	}
+}
+
+type cardinalityControl struct {
+	seen            map[metrics.TimeSeries]struct{}
+	timeSeriesLimit int
+	addsCount       int
```
na--, line 121: I was initially confused by `addsCount`, but I guess you are only using it for optimization purposes? Is `len(map)` a an operation that actually requires some calculation? I thought it just reads an internal counter? So this additional counter here might be a slight overhead, not an optimization

In any case, if we decide to keep it, it should probably be a `uint64` or something like that, considering `int` might be just 32 bits

metrics/engine/ingester.go
na--, line 135: hmm, I know Go has some optimizations for maps that have `struct{}` values, but we need to double-check that this is not doing any allocations on the heap for every call to `Add()`, or something weird like that :thinking: 

I hope this empty struct initialization and map assignment is optimized away by the compiler, but if it's not, then something like this might end up more efficient:
```suggestion
	if _, ok := cc.seen[ts]; !ok {
		cc.seen[ts] = struct{}{}
	}
```

metrics/engine/ingester.go
```diff
@@ -4,10 +4,14 @@ import (
 	"time"
 
 	"github.com/sirupsen/logrus"
+	"go.k6.io/k6/metrics"
 	"go.k6.io/k6/output"
 )
 
-const collectRate = 50 * time.Millisecond
+const (
+	collectRate          = 50 * time.Millisecond
```
mstoykov, line 12: This seems to not be used as well

metrics/engine/ingester.go
```diff
@@ -92,6 +97,60 @@ func (oi *outputIngester) flushMetrics() {
...
+// LimitHit checks if the cardinality limit has been hit.
+func (cc *cardinalityControl) LimitHit() bool {
+	if len(cc.seen) <= cc.timeSeriesLimit {
+		return false
+	}
+
+	// we don't care about overflow
+	// the process should be already OOM
+	// if the number of generated time series goes higher than N-hundred-million(s).
+	cc.timeSeriesLimit *= 2
```
mstoykov, line 149: Is the idea that we want them to get the message multiple times as they continue to make more time series?

I am not really against that, just in the case that we forgo this, we can actually stop counting once the limit is hit and even throw away the `seen` map. 

Not really certain if users will be better off seeing this on 100k, 200k, 400k, 800k, 1.6m and so on time series.

metrics/engine/ingester.go
codebien: https://github.com/grafana/k6/blob/d2db8522f0b300e720215002bd31221799448119/metrics/engine/ingester.go#L38

metrics/engine/ingester.go
na--: :thinking: yeah, maybe we can show that warning 5 times and then stop tracking time series :thinking: not that such an optimization will help very much at 3.2million time series, but still :sweat_smile:

metrics/engine/ingester.go
na--: Sorry for unresolving this, but my point was that I am not sure which is better, it depends on how Go optimizes `map[whatever]struct{}` operations. If it does optimize them properly, it's likely that your original solution is better.

metrics/engine/ingester.go
mstoykov: uff, sorry was looking it is new and then searched it in the PR code :facepalm: ... but it is not new :facepalm:

metrics/engine/ingester.go
na--: I'd argue that it would be useful to see it a few times, since the time between warnings would give information about how quickly the number of time series grows, which is impossible to see or guess otherwise with local k6 runs, and not always easy even with an external output.

imiric (approved):

metrics/engine/ingester.go
na--: I went in a bit of a rabbit hole, but this can probably only be settled by a realistic benchmark, and it doesn't really matter too much either way :sweat_smile: 

FWIW, if https://godbolt.org/ is accurate, then I don't think there are any special optimizations for `struct{}` map values on the map side specifically, though there are optimizations for allocating memory. I was playing with this code, which should be fairly representative, I think:
```go
package main

import "fmt"

var Data map[*int]struct{} = make(map[*int]struct{})
//var Data map[*int]any = make(map[*int]any)
//var Data map[*int]bool = make(map[*int]bool)

func set(p *int) {
    //if _, ok := Data[p]; !ok {
        Data[p] = struct{}{}
        //Data[p] = nil
        //Data[p] = true
    //}
}

func main() {
	data := [...]int{1, 2, 3, 4, 5}
	set(&data[0])
	set(&data[1])
	set(&data[0])
	fmt.Printf("%#q", Data)
}
```

With the `if` commented out, Go will only call [`runtime.mapassign_fast64ptr()`](https://github.com/golang/go/blob/f57f02fcd519a86683c47a444e0e24a086fea8e0/src/runtime/map_fast64.go#L183-L271). If we uncomment the `if` check, it first calls [`mapaccess2_fast64()`](https://github.com/golang/go/blob/f57f02fcd519a86683c47a444e0e24a086fea8e0/src/runtime/map_fast64.go#L53-L91) and then `runtime.mapassign_fast64ptr()`, as we'd expect if there are no special optimizations. And this holds true regardless of whether we use `struct{}`, `interface{}` / `any`, or `bool`. The `any` code has extra code with `runtime.writeBarrier()`, which I didn't get to, but the only difference between `struct{}` and `bool` seems to be a `TESTB` vs `MOVB` instruction at the end.

Looking at the code of these Go runtime functions, my gut feeling is that your original solution of just "assigning" the `struct{}{}` value without the check would be somewhat faster in most cases, though I haven't actually tested it.

na-- (approved):

metrics/engine/ingester.go
codebien: I think it is also helpful for us, in the case of reports, to see what it is the common order of magnitude. We should act differently if they often hit around millions instead of thousands.

PR #2903, opened 2023-02-06
js: Cache scripts by absolute not relative paths

### LINKED ISSUE #2902: Current file path is not taken into account when caching duplicate relative import paths
### Brief summary

k6 appears to be re-using a cached import for a relative path `../fragments/index.js` wherever you use it in the code that's being used in a testrun. 

### k6 version

0.42

### OS

macOS11 / Ubuntu Linux

_No response_

Using the relative path from the brief summary, if imported first in `src/graphql/a/operations/index.js` (to indicate `src/graphql/a/fragments/index.js`) and then later in `src/graphql/b/operations/index.js` (to indicate `src/graphql/b/fragments/index.js`) it will provide me with the former import (`src/graphql/a/fragments/index.js`).

We recently updated to v0.42.0 from v0.40 (where this issue did not occur) and after some digging into any recent changes, it appears to me (though bear in mind I have next to no knowledge in Go) that is what introduced the issue: https://github.com/grafana/k6/blob/b750e206859a495b8a1fe12aa016dc8ed0e1c9db/js/initcontext.go#L109

At this point it seems to me that `../fragments/index.js` is what will serve as key in the cache, instead of the full URL (or anything that includes the file that is doing the importing).

### Expected behaviour

For the relative import to provide me with the correct file relative to whatever file it is being imported from.

### Actual behaviour

A cached import for the exact same relative path name is being returned regardless of where it is being imported from.

mstoykov:
This fixes a regression introduced in 2753196ab where in all imports are cached on the specifier that is provided for them. This is fine for k6 modules which happen to always have the same absolute names.

But source modules can have relative paths that resolve to different absolute ones depending on where they are imported from.

Fixes #2902

### DIFF
--- a/js/initcontext.go
+++ b/js/initcontext.go
@@ -46,7 +46,7 @@ type InitContext struct {
 
 	// Cache of loaded programs and files.
 	programs map[string]programWithSource
-	// mege this and the above
+	// merge this and the above
 	k6ModulesCache map[string]goja.Value
 
 	compatibilityMode lib.CompatibilityMode
--- a/js/initcontext.go
+++ b/js/initcontext.go
@@ -45,8 +45,9 @@ type InitContext struct {
 	pwd         *url.URL
 
 	// Cache of loaded programs and files.
-	programs     map[string]programWithSource
-	exportsCache map[string]goja.Value
+	programs map[string]programWithSource
+	// mege this and the above
+	k6ModulesCache map[string]goja.Value
 
 	compatibilityMode lib.CompatibilityMode
 
@@ -68,7 +69,7 @@ func NewInitContext(
 		compatibilityMode: compatMode,
 		logger:            logger,
 		moduleRegistry:    getJSModules(),
-		exportsCache:      make(map[string]goja.Value),
+		k6ModulesCache:    make(map[string]goja.Value),
 		moduleVUImpl: &moduleVUImpl{
 			// TODO: pass a real context as we did for https://github.com/grafana/k6/pull/2800,
 			// also see https://github.com/grafana/k6/issues/2804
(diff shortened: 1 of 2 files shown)


js/initcontext.go
```diff
@@ -45,8 +45,9 @@ type InitContext struct {
 	pwd         *url.URL
 
 	// Cache of loaded programs and files.
-	programs     map[string]programWithSource
-	exportsCache map[string]goja.Value
+	programs map[string]programWithSource
+	// mege this and the above
```
codebien, line 49:
```suggestion
	// merge this and the above
```
Can you document when it is expected to be done and what is the issue at the moment that blocks it, please?

js/initcontext.go
```diff
@@ -105,13 +106,13 @@ func newBoundInitContext(base *InitContext, vuImpl *moduleVUImpl) *InitContext {
...
-		return export
-	}
-	defer func() { i.exportsCache[arg] = export }()
 	switch {
 	case arg == "k6", strings.HasPrefix(arg, "k6/"):
+		var ok bool
+		if export, ok = i.k6ModulesCache[arg]; ok {
+			return export
+		}
+		defer func() { i.k6ModulesCache[arg] = export }()
```
codebien, line 115: I need clarification on why this is required, does it expected to be _not nil_ in some cases?

codebien (changes requested):

olegbespalov (dismissed):

js/initcontext.go
mstoykov: The blockign part is mostly that thsi will be quite a big refactor which in practice is done in https://github.com/grafana/k6/pull/2881

js/initcontext.go
codebien: Nop, fine for me then

codebien (approved):
LGTM, can you check that the failed test is just one more flaky, please?

olegbespalov (approved):

Issue #2730, opened 2022-10-14
Relax k6's enviroment variables constrains


olegbespalov:
# What?

During the implementation of the [cloud environment variables](https://k6.io/docs/cloud/manage/environment-variables/), we found an inconsistent treating of the variable names validation.

A flag `--env` uses the pattern:
https://github.com/grafana/k6/blob/c6f57c6e22e6966e526643d2d0944e48df0ea36c/cmd/runtime_options.go#L19

And a flag `--include-system-env-vars` doesn't check the environment variables names for that pattern.

That's how task #2650 appears, a PR #2652 was implemented and merged.

Later we found that it leads to not the best UX, because, on windows machines, every k6 run will produce warnings, and the plan was to fix these warnings in #2720. But during the PR and inside the internal team discussion, we decided to re-think the approach there and maybe just relax constraints for the variable names.

That's why the original code from #2652 is reverted in #2723

## Potential issues

The main risk is that the k6 cloud has also implemented such contains; if we change the behavior, the cloud app should also be aligned.

# Why?

https://github.com/grafana/k6/pull/2720#issuecomment-1274834942

na--:
FWIW, as the person who implemented the "environment variables" feature ages ago (https://github.com/grafana/k6/pull/495), I think it was a mistake to use the words "environment variables" and the `-e` / `--env` flag for it... :disappointed: 

It's much more accurate to say that they are script parameters, since that is what they are. And we could have said that, by default, we include the system environment variables as script parameters for `k6 run` (but not for `k6 cloud`), e.g. have `--use-system-env-vars-as-params` instead of `--include-system-env-vars`. And I am not sure if we shouldn't have also made that opt-in even for `k6 run`... :sweat_smile: 

In any case, the current regular expression that validates the environment variable names are POSIX compliant doesn't really have a reason to exist. k6 should't have any issues with "environment variables" that are valid JS/JSON strings, even unicode ones.

So, given the fact that there are no technical limitations on the k6 side and the fact that [on Windows it seems like](https://stackoverflow.com/questions/20635695/what-are-valid-characters-for-windows-environment-variable-names-and-values) almost anything can be an environment variable name, I personally think we should remove this validation on the k6 side. 

Can anyone think of any reasons to keep this validation around?

imiric:
I agree with @na--.

Given that k6 is cross-platform, there's no reason we should enforce that env vars must be POSIX compliant.

[...]

PR #2904, opened 2023-02-07
Change env RegEx : support Cortex Tenant

lanore78:
When using Cortex as a prometheus remote storage, we have to add some http headers like,

* X-Scope-OrgID : for tenant

Cortex Guide Link : https://cortexmetrics.io/docs/guides/auth/

Current regEx for env variable does not contains hypen(-) so, cannot set above header using environment variable , as [this guide](https://k6.io/docs/results-output/real-time/prometheus-remote-write/#options) describe.

The env name for setting above header is 'K6_PROMETHEUS_RW_HEADERS_X-Scope-OrgID', and this cause the following error
```
time="2023-02-07T06:10:41Z" level=error msg="invalid environment variable name 'K6_PROMETHEUS_RW_HEADERS_X-Scope-OrgId'"

<!--


  (ﾉ◕ヮ◕)ﾉ*:・ﾟ✧
  
  Thank you for your interest in contributing to the k6 project!
  
  Before you get started, we'd kindly like to ask you to read our:
    - Contribution guidelines at https://github.com/grafana/k6/blob/master/CONTRIBUTING.md
    - Code of Conduct at https://github.com/grafana/k6/blob/master/CODE_OF_CONDUCT.md
    
  Out of respect for your time, please start a discussion regarding any bigger contributions either
  in a GitHub Issue, in the community forums or in the #contributors channel of the k6 slack before you
  get started on the implementation.
  
  If you've already done all of that, you're more than welcome to proceed with your pull request.
  Thank you again for your contribution! 🙏🏼
  
  
-->

### DIFF
--- a/cmd/runtime_options.go
+++ b/cmd/runtime_options.go
@@ -16,7 +16,7 @@ import (
 // https://github.com/k6io/k6/issues/883, since this code is fairly
 // self-contained and easily testable now, without any global dependencies...
 
-var userEnvVarName = regexp.MustCompile(`^[a-zA-Z_][a-zA-Z0-9_]*$`)
+var userEnvVarName = regexp.MustCompile(`^[a-zA-Z_][a-zA-Z0-9_-]*$`)
 
 func runtimeOptionFlagSet(includeSysEnv bool) *pflag.FlagSet {
 	flags := pflag.NewFlagSet("", 0)


codebien (changes requested):
Hi @lanore78,
thanks for your contribution. :bow: 

I think it could be better to move this contribution directly to the [output's code](https://github.com/grafana/xk6-output-prometheus-remote) addressing the new dedicated issue https://github.com/grafana/xk6-output-prometheus-remote/issues/101. It would be easier to be fixed considering the smaller scope and it could be useable without waiting for the k6 release cycle, using it as an extension. 

Relax this constraint for the environment variable is already tracked in #2730, and we prefer to address it with a final and complete solution.

PR #2905, opened 2023-02-07
Metric type refactor

codebien:

### DIFF
--- a/api/v1/metric.go
+++ b/api/v1/metric.go
@@ -64,11 +64,33 @@ type Metric struct {
 
 // NewMetric constructs a new Metric
 func NewMetric(m *metrics.Metric, t time.Duration) Metric {
-	return Metric{
+	data := Metric{
 		Name:     m.Name,
 		Type:     NullMetricType{m.Type, true},
 		Contains: NullValueType{m.Contains, true},
 		Tainted:  m.Tainted,
-		Sample:   m.Sink.Format(t),
 	}
+
+	switch sink := m.Sink.(type) {
+	case *metrics.CounterSink:
+		data.Sample = map[string]float64{
+			"count": sink.Value,
+			"rate":  sink.Rate(t),
+		}
+	case *metrics.GaugeSink:
+		data.Sample = map[string]float64{"value": sink.Value}
+	case *metrics.RateSink:
+		data.Sample = map[string]float64{"rate": sink.Rate()}
+	case *metrics.TrendSink:
+		data.Sample = map[string]float64{
+			"min":   sink.Min,
+			"max":   sink.Max,
+			"avg":   sink.Avg,
+			"med":   sink.P(0.5),
+			"p(90)": sink.P(0.90),
+			"p(95)": sink.P(0.95),
+		}
+	}
+
+	return data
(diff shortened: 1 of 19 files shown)


metrics/metric.go
```diff
@@ -144,3 +140,12 @@ func ParseMetricName(name string) (string, []string, error) {
 
 	return name[0:openingTokenPos], tags, nil
 }
+
+// ObservedMetric is a static representation
+// of a tracked metric.
+type ObservedMetric struct {
+	*Metric
+	Sink       Sink
+	Thresholds []Threshold
```
na--, line 149: Do we need to tie thresholds to a particular metric? :thinking: This will prevent us from having thresholds that reference multiple metrics in the future (https://github.com/grafana/k6/issues/1441), and maybe even from improvements like https://github.com/grafana/k6/issues/1253

api/v1/metric.go
```diff
@@ -63,12 +63,34 @@ type Metric struct {
...
+		m.Sample = map[string]float64{"value": sink.Value}
+	case *metrics.RateSink:
+		m.Sample = map[string]float64{"rate": sink.Rate()}
+	case *metrics.TrendSink:
+		m.Sample = map[string]float64{
+			"min":   sink.Min,
+			"max":   sink.Max,
+			"avg":   sink.Avg,
+			"med":   sink.P(0.5),
+			"p(90)": sink.P(0.90),
```
na--, line 90: Good opportunity to fix https://github.com/grafana/k6/issues/1253?

metrics/engine/ingester.go
```diff
@@ -55,18 +56,17 @@ func (oi *outputIngester) flushMetrics() {
...
 	// allow us to have a per-bucket lock, instead of one global one, and it
 	// will allow us to split apart the metric Name and Type from its Sink and
 	// Observed fields...
 	//
-	// And, to further optimize things, if every metric (and sub-metric) had a
+	// TODO: And, to further optimize things, if every metric (and sub-metric) had a
 	// sequential integer ID, we would be able to use a slice for these buckets
 	// and eliminate the map loopkups altogether!
 
+	samplesByMetric := make(map[*metrics.Metric][]metrics.Sample)
```
na--, line 68: This is inefficient, maybe we can avoid using `output.SampleBuffer` and do this in a custom `AddMetricSamples()`

metrics/engine/engine.go
```diff
@@ -29,35 +51,42 @@ type MetricsEngine struct {
 	test           *lib.TestRunState
 	outputIngester *outputIngester
 
-	// These can be both top-level metrics or sub-metrics
-	metricsWithThresholds []*metrics.Metric
+	// they can be both top-level metrics or sub-metrics
+	//
+	// TODO: remove the tracked map using the sequence number
+	metricsWithThresholds map[string]metrics.Thresholds
+	trackedMetrics        map[string]*trackedMetric
```
codebien, line 58: We need to investigate as a better and long-term solution the `metric.SequenceID` option

codebien:
This is outdated now. Better to re-work it and re-open after.

PR #2907, opened 2023-02-09
Fix the exit code and run_status when the cloud output aborts a test

na--:
This is a fix for https://github.com/grafana/k6/issues/2804#issuecomment-1415639681


olegbespalov (dismissed):
:crossed_fingers:

output/cloud/output.go
```diff
@@ -309,7 +310,7 @@ func (out *Output) getRunStatus(testErr error) cloudapi.RunStatus {
 			return cloudapi.RunStatusAbortedScriptError
 		case errext.AbortedByScriptAbort:
 			return cloudapi.RunStatusAbortedUser // TODO: have a better value than this?
-		case errext.AbortedByTimeout:
+		case errext.AbortedByTimeout, errext.AbortedByOutput:
```
mstoykov, line 313: Should this even be reported?

AFAIK this will also happen if a user aborted the script through the UI?

It seems like the output likely already set this `status` so us setting it again might not be a good idea? 

Or at least we should try to set it to "abortedbyOutput" or something like that.

na--: Resolved by the [latest force-push](https://github.com/grafana/k6/compare/058efe5bc5e570fc459194976be48e142f7f0bba..dfeca18d21a50de9936b685be94b977f17dc2b47)

mstoykov (approved):

olegbespalov (approved):

Issue #2869, opened 2023-01-23
Refactor `group()` and `check()` tight coupling with the `VU.State`


na--:
This tight coupling is preventing us from implementing `group()` and `check()` purely in JavaScript. It is also preventing us from implementing similar functions, but with better UX (e.g. https://github.com/grafana/k6/issues/1066, https://github.com/grafana/k6/issues/2865) or that work better with `async` / `await` (https://github.com/grafana/k6/issues/2848, https://github.com/grafana/k6/issues/2728)

The tight coupling can be removed with only a very small loss of efficiency, by just having the `MetricEngine` process `check`  and `group_duration` metrics as they appear and group them based on their `group` tag. Here is what basically needs to be removed:
https://github.com/grafana/k6/blob/1d99b0be9194bf9aaf86c070dd0d34604f266abe/lib/vu_state.go#L41-L42
https://github.com/grafana/k6/blob/1d99b0be9194bf9aaf86c070dd0d34604f266abe/lib/runner.go#L71-L72
https://github.com/grafana/k6/blob/1d99b0be9194bf9aaf86c070dd0d34604f266abe/js/modules/k6/k6.go#L98-L104
https://github.com/grafana/k6/blob/1d99b0be9194bf9aaf86c070dd0d34604f266abe/js/modules/k6/k6.go#L169

olegbespalov:
For the record. After an internal discussion, we decided to clean a milestone since the issue was jumping between milestones without completion.

Once we determine which milestone it lands, we set the right one.

PR #2915, opened 2023-02-14
k6.check: Fix returned values when iteration is ending

### LINKED ISSUE #2912: `check` randomly returning true unexpectedly
### Brief summary

During local tests with a large number of preallocated VUs, I have encountered situations
where `check` should have returned `false`, but it returned `true`.

This is difficult to reproduce, since it happens randomly. It seems to be caused by a race condition.

### k6 version

0.42.0.

### OS

WSL2 (Ubuntu 20.04) in Windows 10.

Docker version: 23.0.0; image: `loadimpact/k6:0.42.0`.

Unfortunately, this problem is difficult to reproduce. It seems to be caused by a race condition,
and I haven't been able to find a scenario under which it is guaranteed to happen.

You may need to run the below steps multiple times before you see the issue come up.

I was able to see this problem happen several times when running the following minimal example in Docker, using the following command:

```bash
cat script.js | docker run --cpus=2 --memory=4G -i loadimpact/k6:0.42.0 run -
```

The `script.js` file has the following content:

```js
import { check, sleep } from 'k6';

export const options = {
    scenarios: {
        test: {
            executor: "constant-arrival-rate",
            preAllocatedVUs: 20000,
            rate: 10000,
            timeUnit: "1s",
            duration: "60s",
            exec: "default",
        },
    }
};

function randomNumber() {
    return (Math.random() * 5) + 2
}

export default function() {
    sleep(randomNumber())
    const checkOutput1 = check(null, {
        "my check 1": (res) => false,
    })
    if (checkOutput1) {
        console.log("checkOutput1:", checkOutput1)
    }
    sleep(randomNumber())
    const checkOutput2 = check(null, {
        "my check 2": (res) => false,
    })
    if (checkOutput2) {
        console.log("checkOutput1:", checkOutput1, "checkOutput2:", checkOutput2)
    }
}
```

### Expected behaviour

The `console.log` lines should never be printed, since both checks always return `false`.

### Actual behaviour

Sometimes, randomly, one (or both) of the `console.log` lines is printed.

mstoykov:
Before this check was only returning `false` if it also could emit it, and true otherwise.

This should rarely be a problem as the same context that is being checked here is the one that interrupts the VM - so no code will be able to run after this.

Unfortunately the code checking if it should be emitted is racing with the one that interrupts the VM. So it is possible for the VM to still not be interrupted when a `check` returns a wrong value. It is even possible for more code to be run before the interrupt is actually called.

The code still checks the context as this also updates the internal check structure and we don't want to that if the context is done. The above should be changed with #2869

Fixes #2912

### DIFF
--- a/js/modules/k6/k6_test.go
+++ b/js/modules/k6/k6_test.go
@@ -323,6 +323,30 @@ func TestCheckArray(t *testing.T) {
 	}
 }
 
+func TestCheckContextDone(t *testing.T) {
+	t.Parallel()
+	t.Run("true", func(t *testing.T) {
+		t.Parallel()
+		tc := testCaseRuntime(t)
+
+		tc.testRuntime.CancelContext()
+		err := tc.run(`globalThis.result = k6.check(null, {"name": ()=>{ return true }})`)
+		assert.NoError(t, err)
+		assert.Len(t, metrics.GetBufferedSamples(tc.samples), 0)
+		assert.True(t, tc.testRuntime.VU.Runtime().Get("result").ToBoolean())
+	})
+	t.Run("false", func(t *testing.T) {
+		t.Parallel()
+		tc := testCaseRuntime(t)
+
+		tc.testRuntime.CancelContext()
+		err := tc.run(`globalThis.result = k6.check(null, {"name": ()=>{ return false }})`)
+		assert.NoError(t, err)
+		assert.Len(t, metrics.GetBufferedSamples(tc.samples), 0)
+		assert.False(t, tc.testRuntime.VU.Runtime().Get("result").ToBoolean())
+	})
+}
+
 func TestCheckLiteral(t *testing.T) {
 	t.Parallel()
 	rt, samples, _ := checkTestRuntime(t)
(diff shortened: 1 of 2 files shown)


imiric (commented):
Nice work tracking this issue! :clap:

In hindsight, it seems obvious when reading the code, and it's surprising we haven't run into it until now.

Let's merge this after the v0.43.0 relase, so we can test it properly.

Can you share the script and run options you used to reproduce this? Maybe post it in the issue.

mstoykov:
@imiric I have added tests (I needed more stuff than the current tests needed and didn't want to do a whole test refactor for this )

It is the same as the one in [the issue](https://github.com/grafana/k6/issues/2912) I just decreased the length and added `gracefulStop` as that makes it more likely (given the fix especially). GOMAXPROCS=256 also seems to increase the chances of it being hit.

js/modules/k6/k6_test.go
```diff
@@ -501,3 +525,43 @@ func TestCheckTags(t *testing.T) {
 		}, sample.Tags.Map())
 	}
 }
+
+type testCase struct {
+	samples     chan metrics.SampleContainer
+	testRuntime *modulestest.Runtime
+}
+
+func testCaseRuntime(t testing.TB) *testCase {
```
imiric, line 534: This seems useful, but don't we already have similar helpers? `checkTestRuntime()` is very similar.

Can we reuse an existing helper, and even better, move it to somewhere where it's usable by other packages? This is a generic helper that I'm sure we're duplicating elsewhere, as this setup boilerplate is very common.

mstoykov: I need more stuff than what is available in `checkTestRuntime`  - specifically cancelling the context.

I really didn't want to touch `checkTestRuntime` as this change "ripples" through all the tests using it. And I also didn't want this bugfix PR to be 90% me refactoring test code.

> This is a generic helper that I'm sure we're duplicating elsewhere, as this setup boilerplate is very commo

We are but in every case we need something additional for each module separately. 

As part of the ESM saga I would like to be able to test the module loading and resolution in tc39 and also to be usable from other k6 modules (and extensions as consequence). As part of that I will likely try to refactor all of the helpers through k6 modules at least a little bit, hopefully most will be pretty straightforward :crossed_fingers: 

But that will be a *huge* PR or a bunch of pretty big ones as in most cases this helpers are used a *lot* and then their result(s) are used even more.

mstoykov: https://github.com/grafana/k6/pull/2946 for just how many line changes the refactor of just this, *really* small module,  are.

imiric: Thanks! Yeah, I imagined it wouldn't be a short diff, but still, it DRYs things up considerably.

imiric (approved):

na-- (approved):

PR #2916, opened 2023-02-15
Show actual binary name in the usage help (#2850)

### LINKED ISSUE #2850: Show the actual binary name in the usage help
### Feature Description

When running a custom k6 binary  (for example, the binaries that are distributed with the xk6-disruptor and xk6-browser extensions) the usage help output makes reference to `k6` command and not to the actual binary name. 

For example, when running `xk6-disruptor` the following message is shown
```
$ xk6-disruptor

          /\      |‾‾| /‾‾/   /‾‾/   
     /\  /  \     |  |/  /   /  /    

Usage:
  k6 [command]

... some output omitted for brevity


Use "k6 [command] --help" for more information about a command
```
This may cause confusion to the user. 

Note: the K6.io banner is correct, because it makes reference to the company brand, not to the name of the command executed by the user.

The usage help message(s) should show the name of the binary that the user invoked.

_No response_

ariasmn:
This should close #2850 

I have added some e2e tests in the `cmd_run_test` file, although, as discussed earlier in the issue, perhaps it should be moved to another file since the `run` command is not actually being checked.

### DIFF
--- a/cmd/login_cloud.go
+++ b/cmd/login_cloud.go
@@ -1,10 +1,12 @@
 package cmd
 
 import (
+	"bytes"
 	"encoding/json"
 	"errors"
 	"fmt"
 	"syscall"
+	"text/template"
 
 	"github.com/fatih/color"
 	"github.com/spf13/cobra"
@@ -20,22 +22,29 @@ import (
 //nolint:funlen,gocognit
 func getCmdLoginCloud(gs *state.GlobalState) *cobra.Command {
 	// loginCloudCommand represents the 'login cloud' command
+	var exampleText bytes.Buffer
+	exampleTemplate := template.Must(template.New("").Parse(`
+  # Show the stored token.
+  {{.}} login cloud -s
+  
+  # Store a token.
+  {{.}} login cloud -t YOUR_TOKEN
+  
+  # Log in with an email/password.
+  {{.}} login cloud`[1:]))
+
+	if err := exampleTemplate.Execute(&exampleText, gs.BinaryName); err != nil {
+		gs.Logger.WithError(err).Error("Error during help example generation")
+	}
+
 	loginCloudCommand := &cobra.Command{
 		Use:   "cloud",
 		Short: "Authenticate with Load Impact",
 		Long: `Authenticate with Load Impact.
 
 This will set the default token used when just "k6 run -o cloud" is passed.`,
-		Example: `
-  # Show the stored token.
-  k6 login cloud -s
-
-  # Store a token.
-  k6 login cloud -t YOUR_TOKEN
-
-  # Log in with an email/password.
-  k6 login cloud`[1:],
-		Args: cobra.NoArgs,
+		Example: exampleText.String(),
+		Args:    cobra.NoArgs,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			currentDiskConf, err := readDiskConfig(gs)
 			if err != nil {
--- a/cmd/login_cloud.go
+++ b/cmd/login_cloud.go
@@ -1,12 +1,10 @@
 package cmd
 
 import (
-	"bytes"
 	"encoding/json"
 	"errors"
 	"fmt"
 	"syscall"
-	"text/template"
 
 	"github.com/fatih/color"
 	"github.com/spf13/cobra"
@@ -22,8 +20,7 @@ import (
 //nolint:funlen,gocognit
 func getCmdLoginCloud(gs *state.GlobalState) *cobra.Command {
 	// loginCloudCommand represents the 'login cloud' command
-	var exampleText bytes.Buffer
-	exampleTemplate := template.Must(template.New("").Parse(`
+	exampleText := getExampleText(gs, `
   # Show the stored token.
   {{.}} login cloud -s
   
@@ -31,11 +28,7 @@ func getCmdLoginCloud(gs *state.GlobalState) *cobra.Command {
   {{.}} login cloud -t YOUR_TOKEN
   
   # Log in with an email/password.
-  {{.}} login cloud`[1:]))
-
-	if err := exampleTemplate.Execute(&exampleText, gs.BinaryName); err != nil {
-		gs.Logger.WithError(err).Error("Error during help example generation")
-	}
+  {{.}} login cloud`[1:])
 
 	loginCloudCommand := &cobra.Command{
 		Use:   "cloud",
@@ -43,7 +36,7 @@ func getCmdLoginCloud(gs *state.GlobalState) *cobra.Command {
 		Long: `Authenticate with Load Impact.
 
 This will set the default token used when just "k6 run -o cloud" is passed.`,
-		Example: exampleText.String(),
+		Example: exampleText,
 		Args:    cobra.NoArgs,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			currentDiskConf, err := readDiskConfig(gs)
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -12,7 +12,6 @@ import (
 	"runtime"
 	"strings"
 	"sync"
-	"text/template"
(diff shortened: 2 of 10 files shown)


cmd/run.go
```diff
@@ -356,33 +357,40 @@ func getCmdRun(gs *state.GlobalState) *cobra.Command {
...
+  {{.}} run -u 5 -d 10s script.js
+  
+  # Ramp VUs from 0 to 100 over 10s, stay there for 60s, then 10s down to 0.
+  {{.}} run -u 0 -s 10s:100 -s 60s -s 10s:0
+  
+  # Send metrics to an influxdb server
+  {{.}} run -o influxdb=http://1.2.3.4:8086/k6`))
+
+	if err := exampleTemplate.Execute(&exampleText, gs.BinaryName); err != nil {
+		gs.Logger.WithError(err).Error("Could not parse binary name")
```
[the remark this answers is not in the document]
ariasmn: It shouldn't fail, you are right. 

Actually this error is related to the `template.Execute()` function, since, as you said, the `os.Executable()` error is already handled. (Did that because I was on autopilot, my bad.)

What is the best option here? My main problem is that I don't when the `template.Execute()` is throwing an error :thinking: 

I can log the error like it's done here:
https://github.com/grafana/k6/blob/bbfe357a39c406e5d47e65fae454fda3847afc50/js/modules/k6/html/gen/gen_elements.go#L323-L341

Or I can just execute the template and "ignore" the error

Let me know what you think!

cmd/run.go
```diff
@@ -356,33 +357,40 @@ func getCmdRun(gs *state.GlobalState) *cobra.Command {
 		gs: gs,
 	}
 
+	var exampleText bytes.Buffer
+	exampleTemplate := template.Must(template.New("").Parse(`
+  # Run a single VU, once.
+  {{.}} run script.js
```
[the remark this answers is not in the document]
ariasmn: You are right, this is way better. :+1: 

I will update the PR once I know what is the best way to handle the template errors. (other conversation)

cmd/run.go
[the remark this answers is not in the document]
imiric: From what I can see in the `template/exec.go` implementation, `Execute()` will return an error if the template is incomplete or empty, or if it fails to write to the passed `io.Writer`. I don't think either can happen in this case, so it's probably fine if we just ignore the error.

@mstoykov WDYT?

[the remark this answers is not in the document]
mstoykov: I feel like if the impossible happens it should not be swept under the rug, so if nothing else this seems like a thing we should log.

I would argue that if this suddenly becomes less impossible it will nigh impossible (:wink:) to debug it if we don't at least log it.

I don't know what the message should be though.

Maybe change that to something like "error during help example generation" :shrug: .

cmd/run.go
```diff
@@ -356,33 +357,40 @@ func getCmdRun(gs *state.GlobalState) *cobra.Command {
...
-  k6 run -u 5 -d 10s script.js
-
-  # Ramp VUs from 0 to 100 over 10s, stay there for 60s, then 10s down to 0.
-  k6 run -u 0 -s 10s:100 -s 60s -s 10s:0
-
-  # Send metrics to an influxdb server
-  k6 run -o influxdb=http://1.2.3.4:8086/k6`[1:],
-		Args: exactArgsWithMsg(1, "arg should either be \"-\", if reading script from stdin, or a path to a script file"),
-		RunE: c.run,
+		Example: exampleText.String()[1:],
```
mstoykov, line 391: nit: can you move the `[1:]` on the actual template string above

ariasmn: Done!

cmd/run.go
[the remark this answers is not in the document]
ariasmn: Yeah, better safe than sorry I guess!

I just updated the PR. I changed the error for something generic across all the commands.
If you feel like a specific error should be logged in each case, or if you find any other thing that could be done better, just let me know! :smiley:

cmd/login_cloud.go
```diff
@@ -20,22 +22,29 @@ import (
...
+  # Store a token.
+  {{.}} login cloud -t YOUR_TOKEN
+  
+  # Log in with an email/password.
+  {{.}} login cloud`[1:]))
+
+	if err := exampleTemplate.Execute(&exampleText, gs.BinaryName); err != nil {
+		gs.Logger.WithError(err).Error("Error during help example generation")
+	}
+
```
mstoykov, line 39: I feel like this can be extracted in a function that takes the template string and `gs` and returns a string that is the final example after parsing and Executing it.

ariasmn: Yeah, it makes perfect sense to be honest.

Is there a file that already exists where I can create such a function? If not, where should I place that file?
On another note, should I make a test for this function? I feel like if the command output is already tested it doesn't make much sense, but YMMV, so I'd better ask.

mstoykov: `cmd/common.go` seems like the correct place 

> On another note, should I make a test for this function?

I guess you can at least have 1 test, but I wouldn't go overboard on this to be honest

ariasmn: Done! The naming might be a bit odd, so let me know if I should change anything.

Since I wanted to keep the function in the `cmd/common.go` as unexported, I didn't create a test for it. However, as I said, a `help` command test already exists:

https://github.com/grafana/k6/blob/9aeebf8b8e7b1017fae0f575385e03fcbf5c3394/cmd/tests/cmd_run_test.go#L80-L93

So I guess that's good enough.

cmd/login_cloud.go
```diff
@@ -20,22 +20,24 @@ import (
...
+  # Store a token.
+  {{.}} login cloud -t YOUR_TOKEN
+  
+  # Log in with an email/password.
+  {{.}} login cloud`[1:])
+
 	loginCloudCommand := &cobra.Command{
 		Use:   "cloud",
 		Short: "Authenticate with Load Impact",
 		Long: `Authenticate with Load Impact.
```
imiric, line 36: Unrelated to this PR: we should update this, since we haven't been Load Impact for a few years now. Though I think we might deprecate this command soon, so it maybe doesn't matter.

cmd/tests/cmd_run_test.go
```diff
@@ -63,6 +63,35 @@ func TestSimpleTestStdin(t *testing.T) {
...
+	t.Parallel()
+
+	ts := NewGlobalTestState(t)
+	ts.BinaryName = "customBinaryName"
+	ts.CmdArgs = []string{ts.BinaryName, "help", "run"}
+	cmd.ExecuteWithGlobalState(ts.GlobalState)
+
+	stdout := ts.Stdout.String()
+	assert.Contains(t, stdout, fmt.Sprintf("%s run [flags]", ts.BinaryName))
+	assert.Contains(t, stdout, fmt.Sprintf("%s run -i 10 script.js", ts.BinaryName))
```
imiric, line 90: Would you mind adding tests for the other commands as well? It's not a big deal, but just to be thorough.

Remember to change the test name if you do.

cmd/run.go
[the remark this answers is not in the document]
imiric: Fair enough. The latest change LGTM. :+1:

imiric (dismissed):
LGTM, thanks @ariasmn! :clap:

That `TestStreamLogsToLogger` failure seems to be a flaky test, so we can ignore it.

cmd/tests/cmd_run_test.go
ariasmn: No problem, done!

I did change the parent test name and did it, as you said, using table tests (didn't know about that, so thanks!).

Let me know if anything else should be done.

imiric (approved):
Thanks again! :+1:

mstoykov (approved):

PR #2920, opened 2023-02-21
Bump Go version to 1.20

codebien:

### DIFF
--- a/.github/workflows/arm-tests.yml
+++ b/.github/workflows/arm-tests.yml
@@ -13,7 +13,7 @@ jobs:
       - name: Install Go
         uses: actions/setup-go@v3
         with:
-          go-version: 1.19.x
+          go-version: 1.20.x
           check-latest: true
       - name: Run tests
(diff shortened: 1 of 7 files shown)


imiric (dismissed):

olegbespalov (approved):

olegbespalov:
Probably it makes sense to change the version in `Dockerfile`, but it could a separated PR

mstoykov (commented):
We should bump 1.18.x in the `test.yml` to 1.19.x as well

And Maybe even look into bumping golangci-lint as we are stuck at 1.18.x for it as well

mstoykov (approved):

codebien:
>And Maybe even look into bumping golangci-lint as we are stuck at 1.18.x for it as well

I'm already on it but I would prefer to open a dedicated PR

mstoykov (approved):

na-- (approved):

Issue #2924, opened 2023-02-22
`k6/execution` could panic in `setup`


olegbespalov:
### Brief summary

A given script:

```js
import exec from "k6/execution";

export function setup() {
   if (exec.vu.iterationInScenario == 0) {
      console.log("Hey!");
    }
}

export default function () {}
```

 could produce a panic
```
 k6 run panic-1.js

          /\      |‾‾| /‾‾/   /‾‾/
     /\  /  \     |  |/  /   /  /
    /  \/    \    |     (   /   ‾‾\
   /          \   |  |\  \ |  (‾)  |
  / __________ \  |__| \__\ \_____/ .io

  execution: local
     script: panic-1.js
     output: -

  scenarios: (100.00%) 1 scenario, 1 max VUs, 10m30s max duration (incl. graceful stop):
           * default: 1 iterations for each of 1 VUs (maxDuration: 10m0s, gracefulStop: 30s)

ERRO[0000] panic: runtime error: invalid memory address or nil pointer dereference
goroutine 1 [running]:
runtime/debug.Stack()
        /home/olegbespalov/.go/src/runtime/debug/stack.go:24 +0x65
go.k6.io/k6/js.(*VU).runFn.func1()
        /home/olegbespalov/go/pkg/mod/go.k6.io/k6@v0.34.0/js/runner.go:737 +0x134
panic({0xf2b3e0, 0x1c5f7c0})
        /home/olegbespalov/.go/src/runtime/panic.go:1038 +0x215
github.com/dop251/goja.AssertFunction.func1.1()
        /home/olegbespalov/go/pkg/mod/github.com/dop251/goja@v0.0.0-20210817151038-07a7fd9355b4/runtime.go:2154 +0x74
panic({0xf2b3e0, 0x1c5f7c0})
        /home/olegbespalov/.go/src/runtime/panic.go:1038 +0x215
github.com/dop251/goja.(*vm).try.func1()
        /home/olegbespalov/go/pkg/mod/github.com/dop251/goja@v0.0.0-20210817151038-07a7fd9355b4/vm.go:511 +0x8eb
panic({0xf2b3e0, 0x1c5f7c0})
        /home/olegbespalov/.go/src/runtime/panic.go:1038 +0x215
go.k6.io/k6/js/modules/k6/execution.(*ModuleInstance).newVUInfo.func4()
[...]

PR #2926, opened 2023-02-22
Fix panic in `k6/execution`

olegbespalov:
# What?

Fix a minor panic that could happen if try to get insides about the `exec.vu.iterationInScenario` in `setup` method

In that case, we will return `0` as all other `vu.*` properties do.

# Why?

Panics are bad.

Resolves: #2924


imiric (approved):
LGTM :+1:

codebien (approved):

Issue #2927, opened 2023-02-22
Run windows test with go tip


mstoykov:
This would've caught #2925 ... way earlier

Issue #2925, opened 2023-02-22
`k6 archive` doesn't work on windows with go1.20


codebien:
After the bump of the Go version to 1.20, the Windows tests for the archive command are failing because an error is returned, so the exit code is not zero as expected.

https://github.com/grafana/k6/actions/runs/4232890600/jobs/7353760998

From a quick debug in a Windows VM, k6 or the tests generate a wrong path `\\c:test`.
![Screenshot from 2023-02-22 09-42-51](https://user-images.githubusercontent.com/2103732/220580143-f81576b5-3749-4e6d-a38d-217b0ac6d75c.png)

mstoykov:
Seems likely this was related to the [changes](https://go.dev/doc/go1.20#minor_library_changes) to `archive/tar`: 
> When the GODEBUG=tarinsecurepath=0 environment variable is set, [Reader.Next](https://go.dev/pkg/archive/tar/#Reader.Next) method will now return the error [ErrInsecurePath](https://go.dev/pkg/archive/tar/#ErrInsecurePath) for an entry with a file name that is an absolute path, refers to a location outside the current directory, contains invalid characters, or (on Windows) is a reserved name such as NUL. A future version of Go may disable insecure paths by default.

Can we get a dump of the whole archive contents and maybe to see if other tests fail or just this one?

Also it is a bit strange that this only fails on Windows ... 

Also we technically do not have *absolute* paths inside we have them under `file` which is relative :shrug:

imiric:
:thinking: Could this be related to https://github.com/golang/go/issues/58348 ?

olegbespalov:
I guess @mstoykov is right. Check this out :cry: 

![VirtualBox_WinDev2301Eval_22_02_2023_13_13_43](https://user-images.githubusercontent.com/5425600/220617080-1ad24943-9fbf-4764-980a-a915734e9e13.png)

It seems like the `k6 archive` doesn't work with a binary that was built using go 1.20

The good news, tests are good :+1:

PR #2928, opened 2023-02-22
Handle a new behaviour of  `filepath.Join` on windows go.120

olegbespalov:
# What?

In #2925 we found that our test on windows + go1.20 started to fail. After localizing the issue, we spotted a different behavior of the `filepath.Join` which uses underlining the Clean (and  https://github.com/golang/go/issues/56336#issuecomment-1416214885) and is platform dependent. 

For example:

```
// go 1.19 (win)

filepath.Join("\\c:", "test")  // \c:\test

// go 1.20 (win)

filepath.Join("\\c:", "test")  // \c:test
```

Adding the path separator helps to restore the original behavior:

```
// go 1.19, go 1.20 (win)

filepath.Join("\\c:", path.Clean("/"+"test"))  // \c:\test
filepath.Join("\\c:",path.Clean("/"+ "test"))  // \c:\test
```

Also, this PR contains a minor fix for the tests, switches the coverage to 1.20, and introduces go-tip testing on windows, to spot the issues like that earlier.

# Why?

Resolves: #2925 
Resolves: #2927


js/runner_test.go
```diff
@@ -2312,7 +2312,7 @@ type multiFileTestCase struct {
 
 func runMultiFileTestCase(t *testing.T, tc multiFileTestCase, tb *httpmultibin.HTTPMultiBin) {
 	t.Helper()
-	runner, err := getSimpleRunner(t, tc.cwd+"/script.js", tc.script, tc.rtOpts, tc.fses)
+	runner, err := getSimpleRunner(t, tc.cwd+"script.js", tc.script, tc.rtOpts, tc.fses)
```
mstoykov, line 2315: Shouldn't we use `fsext,JoinFilePath` here as well?

lib/fsext/filepath_windows_test.go
```diff
@@ -0,0 +1,50 @@
+//go:build windows
```
mstoykov, line 1: Unfortunately `k6 archive` can generate an archive on one OS that is run on another. So we *probably* need to add more tests about paths generated on Windows for the archive and then check if another OS can actually work with them and vice versa. 

This actually is big part of what the current test do, but it seems it might be a good idea to extend those and double check that this is enough. 

Maybe creating a full integration test where we run an archive generated from a windows machine (both per 1.20 and post if there are end result differences) and ones generated on a linux machine. Both should run on all supported OSes

js/runner_test.go
imiric, line 2315: I saw why this was done in db8dab64, but [not all `tc.cwd`s have a trailing slash](https://github.com/grafana/k6/blob/1993ac1b6d47d3504631648d3cf02aebcd5adfbf/js/runner_test.go#L2441-L2443). So in those cases this will evaluate to, e.g. `/whateverscript.js` or `/pathscript.js`. It apparently has no effect on the tests, as they still pass, so I'm not sure if we should do anything about this, just wanted to point it out.

Maybe change all `tc.cwd` to have a trailing slash?

imiric (commented):
Nice job tracking this down, @olegbespalov and @codebien! :clap:

The fix LGTM. It's unintuitive, but the explanation helps.

I tested this manually on Windows and Linux by creating archives on one and running them on the other, and the behavior seems unchanged for Go 1.20.1, besides the fix actually working. :+1:

Just left a minor comment about the GRPC test.

lib/fsext/filepath_windows_test.go
imiric: I tested this manually by creating archives on Windows with Go 1.19.6 and 1.20.1 (on this branch), and they both ran fine on Linux.

But I agree, it would be good to have integration tests that do that.

olegbespalov: Generally, I like the idea since it sounds reasonable. However, it should be implemented separately. WDYT?

js/runner_test.go
olegbespalov: You're right here, I've answered Mihail in another tread https://github.com/grafana/k6/pull/2928#discussion_r1120543725 Hope that works :crossed_fingers:

lib/fsext/filepath_windows_test.go
olegbespalov: Anyway, the last commit contains the tests. Please let me know if I need to extend something there.

mstoykov (approved):
LGTM! 

great job and takes for taking care of this :bow:

.github/workflows/test.yml
```diff
@@ -46,7 +46,11 @@ jobs:
           go test "${args[@]}" -timeout 800s ./...
 
   test-tip:
-    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        platform: [ubuntu-latest, windows-2019]
```
imiric, line 52: Somewhat unrelated to this PR, but should we also pin the version of Ubuntu, to avoid any issues when GH upgrades it? Same for the the other jobs.

cmd/tests/cmd_run_test.go
```diff
@@ -1837,3 +1837,47 @@ func TestUIRenderOutput(t *testing.T) {
...
+// └── script.js
+// archive was made using binary & platform from the test name
+func TestRunStaticArchives(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		archive string
+	}{
+		{archive: "archive_v0.42.0_linux.tar"},
+		{archive: "archive_v0.42.0_windows.tar"},
```
imiric, line 1857: Ah, I was picturing more of a somewhat complex CI workflow where we actually run `k6 archive` on one OS, and then run the test on the other. That way we wouldn't need to commit the packages. But I guess this is fine as well.

imiric (dismissed):
:+1:

lib/fsext/filepath.go
```diff
@@ -0,0 +1,17 @@
...
+	"path"
+	"path/filepath"
+)
+
+// JoinFilePath is a wrapper around filepath.Join
+// starting go 1.20 on Windows, Clean does not modify the
+// volume name other than to replace occurrences of "/" with `\`.
+// that's why we need to add a leading slash to the path
+// go.1.19: filepath.Join("\\c:", "test")  // \c:\test
+// go.1.20: filepath.Join("\\c:", "test")  // \c:test
```
codebien, line 14: @olegbespalov I think in this comment, we're missing why we're using `path` instead of `filepath`.

olegbespalov: @codebien sorry, I'm missing the point :cry: The comment is about the behavior change behavior of the Clean, which is used underneath the `filepath.Join`

But for sure, feel free to suggest a comment that you think will be more informative.

cmd/tests/cmd_run_test.go
olegbespalov: >I was picturing more of a somewhat complex CI workflow

In my mind, I have the same picture, but that picture also sounds costly for the scope of this PR, so this test is a tread-off. Maybe in the future, we could improve it  (should we create the task then :thinking: ).

.github/workflows/test.yml
na--: Hmm, wouldn't it be better to surface any issue earlier rather than later? We can always pin it if it stars to fail, but we'd at least know there is a problem.

cmd/tests/cmd_run_test.go
na--: Also, this has the benefit of being able to test archives generated from old k6 version on new k6 versions, which we won't easily be able to do with `k6 archive`. So, both are good to have eventually, but this is the more important one and the one easier to implement.

na-- (approved):

lib/fsext/filepath.go
codebien: I mean why `path.Clean` and not `filepath.Clean`

.github/workflows/test.yml
imiric: We've had issues with this before in some repo, I forget where now. And it's why Windows is pinned. I'd rather avoid surprises and possibly difficult to troubleshoot failures by pinning everything, just as we do with all dependencies, and updating manually on our schedule.

We could also add a `next` job, like we do with `test-tip`, and make any failures optional, so that they don't block PRs, but we would notice if they fail.

lib/fsext/filepath.go
olegbespalov: `path.Clean` and `filepath.Clean` are pretty similar in their behavior, and the thing is that in that particular case, it doesn't matter which one we use. The outcome remains. 

We kind of force to treat the second argument of  `filepath.Join` as the one that "has slash" regardless it has it or not. And later filepath.Join applies cross-platform manipulations.

If you feel that the `filepath.Join` is more appropriate here, I could change it :+1:

codebien: >The filepath package uses **either forward slashes or backslashes**, depending on the operating system. To process paths such as URLs that **always use forward slashes** regardless of the operating system, see the path package. 

We are using both, so my interpretation from the docs is that `filepath` could be more consistent with it.

olegbespalov: yeah, but as I said we use the `filepath.Join` later, but okay I'll change it for the sake of the consistency

imiric (approved):

mstoykov (approved):

codebien (approved):

PR #2934, opened 2023-02-24
Print the expected char with the default output

codebien:
It fixes the ui print `output: engine` restoring the expected text `output: -` when there isn't an explicit set from the user.


olegbespalov (approved):

imiric (approved):
:+1:

na-- (changes requested):
What happens if you have `k6 run --out json` or `--out cloud`? you'd see something like `output: -, cloud`, right?

imiric (approved):
It _should_ be fine now :sweat_smile:

Thanks for the test :+1:

na-- (approved):
Thanks! :bow: :heart:

Issue #2937, opened 2023-02-24
K6 Process Hangs After Summary Report


ichasepucks:
### Brief summary

Starting with 0.43.0 we're noticing the K6 process does not terminate after the summary report.  This does not happen with 0.42.0 and earlier versions.

We are building our own custom docker image with the k6-sql plugin.  However, I've tried with the [loadimpact/k6](https://hub.docker.com/r/loadimpact/k6) image as well and can reproduce, so I don't think it's actually plugin related.

Other configuration details:
- Using experimental-prometheus-rw (tried without this as well)
- K6_PROMETHEUS_RW_STALE_MARKERS=true (tried both)
- K6_PROMETHEUS_RW_TREND_STATS=p(90),p(95),p(99)
- options --log-format=json --no-color

The execution environment is on a Kubernetes cluster.

Tail of logs from 0.43.0 simplified
```
{"executor":"shared-iterations","gracefulStop":30000000000,"level":"debug","msg":"Regular duration is done, waiting for iterations to gracefully finish","scenario":"xxx","time":"2023-02-24T15:41:22Z"}
{"executor":"xxx","level":"debug","msg":"Executor finished successfully","startTime":0,"time":"2023-02-24T15:41:22Z","type":"shared-iterations"}
{"level":"debug","msg":"Running teardown()...","time":"2023-02-24T15:41:22Z"}
{"level":"debug","msg":"Test finished cleanly","time":"2023-02-24T15:41:22Z"}
{"level":"debug","msg":"Stopping vus and vux_max metrics emission...","phase":"execution-scheduler-init","time":"2023-02-24T15:41:22Z"}
{"level":"debug","msg":"Metrics emission of VUs and VUsMax metrics stopped","time":"2023-02-24T15:41:22Z"}
{"level":"debug","msg":"Releasing signal trap...","time":"2023-02-24T15:41:22Z"}
{"level":"debug","msg":"Sending usage report...","time":"2023-02-24T15:41:22Z"}
{"level":"debug","msg":"Waiting for metric processing to finish...","time":"2023-02-24T15:41:22Z"}
{"level":"debug","msg":"Metrics processing finished!","time":"2023-02-24T15:41:22Z"}
{"level":"debug","msg":"Finalizing thresholds...","time":"2023-02-24T15:41:22Z"}
{"component":"metrics-engine-ingester","level":"debug","msg":"Stopping...","time":"2023-02-24T15:41:22Z"}
[...]

PR #2938, opened 2023-02-25
Fix usage report block of k6 execution if it takes longer than 3s

na--:
This resolves https://github.com/grafana/k6/issues/2937

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -307,22 +307,17 @@ func (c *cmdRun) run(cmd *cobra.Command, args []string) (err error) {
 	// Init has passed successfully, so unless disabled, make sure we send a
 	// usage report after the context is done.
 	if !conf.NoUsageReport.Bool {
-		backgroundProcesses.Add(2)
-		reportCtx, reportCancel := context.WithCancel(globalCtx)
-		reportDone := make(chan error)
+		backgroundProcesses.Add(1)
 		go func() {
+			defer backgroundProcesses.Done()
+			reportCtx, reportCancel := context.WithTimeout(globalCtx, 3*time.Second)
+			defer reportCancel()
 			logger.Debug("Sending usage report...")
-			reportDone <- reportUsage(reportCtx, execScheduler)
-			close(reportDone)
-			backgroundProcesses.Done()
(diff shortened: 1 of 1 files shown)


cmd/run.go
```diff
@@ -307,22 +307,13 @@ func (c *cmdRun) run(cmd *cobra.Command, args []string) (err error) {
 	// Init has passed successfully, so unless disabled, make sure we send a
 	// usage report after the context is done.
 	if !conf.NoUsageReport.Bool {
-		backgroundProcesses.Add(2)
-		reportCtx, reportCancel := context.WithCancel(globalCtx)
-		reportDone := make(chan error)
+		backgroundProcesses.Add(1)
+		reportCtx, reportCancel := context.WithTimeout(globalCtx, 3*time.Second)
```
mstoykov, line 311: Can you move this line to below the first `defer` 

We do not need the new context outside of that funciton and I kind of prefer code where if there is somethign that it will be `defer`ed it is *right* after it needs to be.

mstoykov (dismissed):
LGTM, just a nitpick

cmd/run.go
na--: Resolved by [this force-push](https://github.com/grafana/k6/compare/1776af241bc048772c2042bea4e1a3e284f66114..635d0815c9b117aa79f213fccb12876fff79f134).

cmd/run.go
```diff
@@ -307,22 +307,13 @@ func (c *cmdRun) run(cmd *cobra.Command, args []string) (err error) {
...
-			backgroundProcesses.Done()
-		}()
-		go func() {
-			select {
-			case <-reportDone:
-			case <-time.After(3 * time.Second):
-			}
-			reportCancel()
-			backgroundProcesses.Done()
+			_ = reportUsage(reportCtx, execScheduler)
```
mstoykov, line 316: Hm, maybe log it after all?

na--: Resolved by the latest force-push

mstoykov (approved):

codebien (approved):

PR #2942, opened 2023-02-27
Reference 'k6 cloud' instead of 'Load Impact' in docs and errors

mstoykov:
triggered by https://github.com/grafana/k6/pull/2916/files#r1117146117

PR #2943, opened 2023-02-27
Drop a sample that uses a removed REST API

mstoykov:
This sample is 6 years old and uses a version of k6 cloud (at the time Load Impact) that predates k6 itself and has been removed for several years now.

The sample while more on the complex side is not very useful as it can't be run.

There are also now a lot more example in the wild of complex scripts then 6 years ago.


mstoykov:
This is the remaining places where "Load Impact" referenced apart from https://github.com/grafana/k6/pull/2942 and release notes which I don't think we should update.

na-- (approved):

imiric (approved):

Issue #1977, opened 2021-04-19
Why does the built-in metric “vus” in .json output not reflect the console output?


Dieviete:
In the console output of K6 you can easily see how many VUs are currently active. See picture k6 console output

![image](https://user-images.githubusercontent.com/43817715/115234662-94f0de80-a119-11eb-958c-21e9bfd3ce86.png)

In the .json output there is a metric "vus" and a metric "vus_max". I would expect "vus" to reflect the number of currently active virtual users. However, in my .json output, except for the entries where "vus" is 0, "vus" and "vus_max" always have the same value. See picture k6 .json output

![image](https://user-images.githubusercontent.com/43817715/115234695-9de1b000-a119-11eb-874b-3b8a45af9c12.png)

Is this a bug? If not, how to generate the correct number of active VUs in the .json output?

na--:
Thanks for reporting this! We've known about for a while, but we lost sight of it, so I'it was a compromise for expedience 

 it is something halfway between a bug and a missing feature... :sweat_smile: Its time to get fixed might have finally come, however, 

It was done like this for expedience and performance, basically it's relatively easy to adjust the con

na--:
Thanks for reporting this! We've known about for a while, but we lost sight of it... :disappointed: It is something halfway between a bug and a missing feature, one of the compromises for the sake of expedience to finally get #1007 out of the door last year... :sweat_smile: Its time to get fixed might have finally come, however, so I'll edit this issue a bit and keep it open. We probably should take care of it together with the rest of the items in https://github.com/k6io/k6/issues/1386 and https://github.com/k6io/k6/issues/1944

To give some details about the issue: currently, arrival-rate executors can easily show in the progress bars precisely how many active VUs they are using. However, since multiple scenarios can be executed simultaneously, and the `vus` metric is reported by k6 every second and encompasses all scenarios (though that _might_ also change, see https://github.com/k6io/k6/issues/1833), arrival-rate executors report as active all initialized VUs, even when that might not be the case.

Dieviete:
Thank you for the quick response ☺️

jsolv:
Hello!
How about status of this issue?
I can't get current actual VU's in metrics (from influxDB ) like progress bar show, only values after slash. But i think it's good to write value before slash also.

na--:
Sorry @jsolv, nobody has worked on this yet. I added it to the next milestone, but I can't make any promises when we'll get to it... :disappointed: If someone wants to make a PR before we can, we can review it though!

PR #2953, opened 2023-03-03
Fix active VU reporting by arrival-rate executors

imiric:
This is a somewhat quick and dirty fix of #1977, where instead of changing the global active VUs when the VU is activated, we change it when the VU is actually running. This synchronizes it with the running count of the internal `activeVUPool`, and reports the correct value everywhere the `ExecutionState` is used (the real-time total active VU count while the test is running, in all outputs, etc.).

### DIFF
--- a/lib/executor/constant_arrival_rate.go
+++ b/lib/executor/constant_arrival_rate.go
@@ -264,6 +264,10 @@ func (car ConstantArrivalRate) Run(parentCtx context.Context, out chan<- metrics
 	}()
 
 	returnVU := func(u lib.InitializedVU) {
+		// Return the VU without decreasing the global active VU counter, which
+		// is done in the goroutine started by activeVUPool.AddVU, whenever the
+		// VU finishes running an iteration. This results in a more accurate
+		// report of VUs that are _actually_ active.
 		car.executionState.ReturnVU(u, false)
 		activeVUsWg.Done()
 	}
--- a/lib/executor/constant_arrival_rate.go
+++ b/lib/executor/constant_arrival_rate.go
@@ -217,7 +217,7 @@ func (car ConstantArrivalRate) Run(parentCtx context.Context, out chan<- metrics
 		<-waitOnProgressChannel
 	}()
 
-	vusPool := newActiveVUPool()
+	vusPool := newActiveVUPool(car.executionState)
 	defer func() {
 		// Make sure all VUs aren't executing iterations anymore, for the cancel()
 		// below to deactivate them.
@@ -264,7 +264,7 @@ func (car ConstantArrivalRate) Run(parentCtx context.Context, out chan<- metrics
 	}()
 
 	returnVU := func(u lib.InitializedVU) {
-		car.executionState.ReturnVU(u, true)
+		car.executionState.ReturnVU(u, false)
 		activeVUsWg.Done()
 	}
 
@@ -275,7 +275,6 @@ func (car ConstantArrivalRate) Run(parentCtx context.Context, out chan<- metrics
 			maxDurationCtx, car.config.BaseConfig, returnVU,
 			car.nextIterationCounters,
 		))
-		car.executionState.ModCurrentlyActiveVUsCount(+1)
 		atomic.AddUint64(&activeVUsCount, 1)
 		vusPool.AddVU(maxDurationCtx, activeVU, runIterationBasic)
 		return activeVU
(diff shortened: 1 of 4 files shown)


lib/executor/ramping_arrival_rate_test.go
```diff
@@ -734,3 +734,45 @@ func TestRampingArrivalRateCornerCase(t *testing.T) {
...
+	})
+	test := setupExecutorTest(t, "", "", lib.Options{}, runner, config)
+	defer test.cancel()
+
+	getCurrActiveVUs = test.state.GetCurrentlyActiveVUsCount
+
+	engineOut := make(chan metrics.SampleContainer, 1000)
+	require.NoError(t, test.executor.Run(test.ctx, engineOut))
+
+	assert.GreaterOrEqual(t, running, int64(5))
```
codebien, line 777: nit: can we check that it isn't bigger than maxVUs?

lib/executor/constant_arrival_rate.go
```diff
@@ -264,7 +264,7 @@ func (car ConstantArrivalRate) Run(parentCtx context.Context, out chan<- metrics
 	}()
 
 	returnVU := func(u lib.InitializedVU) {
-		car.executionState.ReturnVU(u, true)
+		car.executionState.ReturnVU(u, false)
 		activeVUsWg.Done()
```
codebien, line 268: This could generate confusion in the future, we set ReturnVU with _false_ but are setting **activeVUs**Wg as Done. Should we add a small comment to remind us why are we doing it? e.g. `We set activeVUs as false because the counter is already updated by the pool`.

codebien (changes requested):
>where instead of changing the global active VUs when the VU is activated, we change it when the VU is actually running

Is this logic consistent across all executors now? :thinking:

lib/executor/constant_arrival_rate.go
imiric: Good point. See 8d43a834.

imiric:
@codebien

>> where instead of changing the global active VUs when the VU is activated, we change it when the VU is actually running
>
> Is this logic consistent across all executors now? 

This behavior is unique to the arrival rate executors, which are the only ones that have an internal VU pool. They introduce an additional meaning to "active" VUs, which can actually remain dormant until they're "running" an iteration. This change simply synchronizes the "running" counter to the global "active" one, which makes the value reported globally more accurate.

codebien (dismissed):
LGTM :bow:

lib/executor/constant_arrival_rate_test.go
```diff
@@ -354,3 +354,41 @@ func TestConstantArrivalRateGlobalIters(t *testing.T) {
...
+
+	var (
+		running          int64
+		getCurrActiveVUs func() int64
+		runMx            sync.Mutex
+	)
+
+	runner := simpleRunner(func(ctx context.Context, _ *lib.State) error {
+		runMx.Lock()
+		assert.Equal(t, atomic.AddInt64(&running, 1), getCurrActiveVUs())
```
mstoykov, line 378: nit: either the atomic here is pointless (we already have mutex) or you should also use it t o read the value in the checks at the end of the test

mstoykov (approved):
LGTM! That seems way too easy :rofl: but I expect it was hard to make it this easy :clap: 


I did some benchmarking and it seems to have some performance impact in one of my scripts lying around, but it seemed really small. After some more tests - specifically with empty default functions, I am pretty sure that was due to some randomness in my original script, so it seems to work really well.

lib/executor/constant_arrival_rate_test.go
imiric: Good point. I removed the atomic in 7162f778.

The reason I didn't use it at the end is because the test is already done at that point, so it should be safe to read it directly.

imiric:
> but I expect it was hard to make it this easy

Not really, I was surprised as well, and thought there might be a catch I'm missing. :sweat_smile: It does have the side effect that the global VU counter is now "bursty" as the arrival rate VU counter is, so we might want to cache it slightly at least in the UI, since currently it's not very legible on terminals with high refresh rate.

Thanks for confirming it doesn't impact performance. I was slightly concerned about that as well. :+1:

mstoykov (approved):

codebien (approved):
@imiric FYI, I triggered the stuck CLA check. Remember to squash and merge so we can aggregate the fixups. Thanks for doing this :bow:

agilob:
Awesome 
image

Issue #1751, opened 2020-12-02
New DNS resolver


imiric:
In #1612 we exposed some configuration options to control DNS behavior (and dropped the dnscache dependency :tada:), which provided a workaround for issues like #726 and #738. However, k6 still uses the native Go DNS resolver or the OS resolver (depending on platform, `GODEBUG=netdns`, etc.), which is limited by design. [The Go team has rejected proposals for adding a TTL-based cache to stdlib](https://github.com/golang/go/issues/24796), but such a feature is crucial to a web testing tool.

## Feature Description

The proper fix for both of these issues, and likely others that haven't been reported, is for k6 to use a more advanced DNS resolver that supports TTL and possibly other features (DoH, DNSSEC, ...?).

So far there have been two attempts at moving forward with this, but both have halted because of the complexity involved when making sure it all works cross-platform, with support on Windows being the more difficult part.

- https://github.com/imiric/k6/commits/fix/726-738-dns-caching-1085

  This is a continuation of previous work by bookmoons, replacing the DNS libraries used and attempting to sidestep the OS resolvers in favor of a single resolver in Go. The current branch works fine on Linux, but there were some complications on Windows that weren't unsurmountable, but caused us to abandon this approach for the v0.28.0 release window and prioritize the simpler configuration changes. There are some more details in [this *internal* Google doc](https://docs.google.com/document/d/1D9TMI2StZPBVKGWK40BifIP521O4BQ3dT4fErEgn3_4/edit?usp=sharing).

- https://github.com/loadimpact/k6/commits/fix/726-738-dns-cache-refresh-impr / #1637

  This is a simpler approach that replaces `net.Resolver`'s `Dial()` method, but IIRC also had issues on Windows, and it uses an old `Message` implementation instead of [the one currently in stdlib](https://github.com/golang/net/blob/master/dns/dnsmessage/message.go). @MStoykov can expand on this.

## Suggested Solution

We should explore the simpler approach of #1637, probably plucking the required parts from [benburkert/dns](https://github.com/benburkert/dns) and updating it to use the new `Message` implementation, resolving the issues mentioned in the PR, and ensuring it works well cross-platform. Yes, even on Windows... 😈

As discussed, this is not high priority as #1612 addressed the biggest pain points and there hasn't been much user requests for this, but we should consider picking it back up sometime in H1 2021.

Issue #1011, opened 2019-04-26
http_req_looking_up metric missing from output


jonboydell-massive:
I don't see the http_req_looking_up metric in the std output and if I try to create a Trend recording `myTrend.add(res.timings.looking_up);` I get a series of 0s in the output of that Trend.

I'm using v0.24.0

na--:
Thanks for reporting this and sorry for the confusion - the documentation was wrong, it seems like it hadn't been updated after the tracking of DNS lookup times was removed in k6 v0.13 by [this commit](https://github.com/loadimpact/k6/commit/21be5917036e59f09187aa14fbd67fc5da5f3159). 

I don't know the full details about what prompted the removal of the metric, though https://github.com/loadimpact/k6/issues/41 points to the possibility that it was also part of the `http_req_duration` meta-metric and skewed it too much. In any case, we plan to bring it back, and improve DNS support in general. I'll leave this issue open as a tracking point for the DNS metric measurements specifically, and here are the other DNS-related issues that we want to fix:
- https://github.com/loadimpact/k6/issues/726 for DNS caching issues - fixing this should happen before the current metrics issue, since the unexpected DNS caching will skew any metrics
- https://github.com/loadimpact/k6/issues/738 for ignoring DNS round-robin
- https://github.com/loadimpact/k6/issues/851 for actually exposing DNS resolving as a k6 JS API and measuring all DNS-related metrics

MattDodsonEnglish:
This response metric was still documented in the API. It said "currently unavailable", but nevertheless at least one user tried to use it anyway (that user was me). 

I've removed the documentation. If this issue is ever fixed, we can redocument it.

https://github.com/grafana/k6-docs/commit/26e510966efedbcfe199037a6d3a90a9cae3f955

Issue #41, opened 2016-12-22
DNS lookup produces duration spike


ragnarlonn:
Initial response duration spikes are collected at least when testing loadimpact.com.

*Imported from https://trello.com/c/mSdpLjRd/117-dns-lookup-produces-duration-spike*

https://zapier.com/engine/hydrate/875802/.eJwVjFsOgyAQAO-y334AyiMcolcg67qthEob2cQY491Lf2cmc0GuTbASp7xA1JNzY_B-gGfm95IqbgwRUARp3bhKgwFoZSqp8AkxeBuU6ehTpdsk5_ffP3pVDtxfDeIFhHtfgw12xpm8dgYna0eljGfWDu77B2ucKck:1cK13e:HlTPrSiruzbYTOBnBQH2i1BYxr8/

PR #2961, opened 2023-03-08
Add support for reporting DNS resolution times

mem:
The time it takes to resolve hostnames is missing from the reported metrcs. Add support for recording that information thru the httptrace DNSStart and DNSDone hooks. This requires changing the signature of the netext.MultiResolver type because nettrace depends on having the hooks passed in a structure in the request context (they are wired up by httptrace).

### DIFF
--- a/lib/netext/httpext/tracer.go
+++ b/lib/netext/httpext/tracer.go
@@ -22,6 +22,7 @@ type Trail struct {
 	// Total request duration, excluding DNS lookup and connect time.
 	Duration time.Duration
 
+	Resolve        time.Duration // DNS lookup.
 	Blocked        time.Duration // Waiting to acquire a connection.
 	Connecting     time.Duration // Connecting to remote host.
 	TLSHandshaking time.Duration // Executing TLS handshake.
@@ -63,6 +64,15 @@ func (tr *Trail) SaveSamples(builtinMetrics *metrics.BuiltinMetrics, ctm *metric
 			Time:     tr.EndTime,
 			Metadata: ctm.Metadata,
 			Value:    metrics.D(tr.Duration),
+		},
+		{
+			TimeSeries: metrics.TimeSeries{
+				Metric: builtinMetrics.HTTPReqResolve,
+				Tags:   ctm.Tags,
+			},
+			Time:     tr.EndTime,
+			Metadata: ctm.Metadata,
+			Value:    metrics.D(tr.Resolve),
 		},
 		{
 			TimeSeries: metrics.TimeSeries{
@@ -146,6 +156,8 @@ var _ metrics.ConnectedSampleContainer = &Trail{}
 // Cheers, love, the cavalry's here.
 type Tracer struct {
 	getConn              int64
+	dnsStart             int64
+	dnsDone              int64
 	connectStart         int64
 	connectDone          int64
 	tlsHandshakeStart    int64
@@ -162,6 +174,8 @@ type Tracer struct {
 func (t *Tracer) Trace() *httptrace.ClientTrace {
 	return &httptrace.ClientTrace{
 		GetConn:              t.GetConn,
+		DNSStart:             t.DnsStart,
+		DNSDone:              t.DnsDone,
 		ConnectStart:         t.ConnectStart,
 		ConnectDone:          t.ConnectDone,
 		TLSHandshakeStart:    t.TLSHandshakeStart,
@@ -170,6 +184,14 @@ func (t *Tracer) Trace() *httptrace.ClientTrace {
 		WroteRequest:         t.WroteRequest,
 		GotFirstResponseByte: t.GotFirstResponseByte,
 	}
+}
+
+func (t *Tracer) Start() {
+	t.start = now()
+}
+
+func (t *Tracer) End() {
+	t.end = now()
 }
 
 func now() int64 {
@@ -186,6 +208,16 @@ func now() int64 {
 // If it's called, it will be called before all other hooks.
 func (t *Tracer) GetConn(hostPort string) {
 	t.getConn = now()
+}
+
+// DnsStart is called when a DNS lookup begins.
+func (t *Tracer) DnsStart(_ httptrace.DNSStartInfo) {
+	t.dnsStart = now()
+}
+
+// DnsDone is called when a DNS lookup ends.
+func (t *Tracer) DnsDone(_ httptrace.DNSDoneInfo) {
+	t.dnsDone = now()
 }
 
 // ConnectStart is called when a new connection's Dial begins.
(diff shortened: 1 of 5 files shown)


mem:
Tests are broken because I want some feedback before I go figuring out how to fix those.

lib/netext/httpext/tracer.go
```diff
@@ -172,6 +186,14 @@ func (t *Tracer) Trace() *httptrace.ClientTrace {
 	}
 }
 
+func (t *Tracer) Start() {
+	t.start = now()
+}
+
+func (t *Tracer) End() {
```
mem, line 193: please ignore this, it's not part of this PR

lib/netext/httpext/tracer.go
```diff
@@ -172,6 +186,14 @@ func (t *Tracer) Trace() *httptrace.ClientTrace {
 	}
 }
 
+func (t *Tracer) Start() {
+	t.start = now()
```
mem, line 190: please ignore this, this is not part of this PR

na--:
Thanks, @mem! I removed myself as a reviewer, since I won't have any free time in the upcoming week and a half, but I can share some historical context here. 

First, we have an open issue about this, https://github.com/grafana/k6/issues/1011. And, as you can see in it, k6 used to measure DNS lookup times a long time ago, but then it was removed in https://github.com/grafana/k6/commit/21be5917036e59f09187aa14fbd67fc5da5f3159. The removal happened before my time, but from what I can see in the commit and the issue (https://github.com/grafana/k6/issues/41), it was done because we added some pretty heavy DNS caching. 

Since then, we have overhauled that caching pretty significantly, making it more configurable and usable. @imiric did the vast majority of the work (https://github.com/grafana/k6/pull/1612) here, that is why I added him to the PR. However, we didn't bring back the dns resolution metric at the time, so https://github.com/grafana/k6/issues/1011 remained open. 

There seemingly wasn't a ton of demand for that feature, and there is some complexity in doing this correctly, so we probably decided to cut the scope at the time. Can you explain your use specific case and why you need this in the issue?

Finally, a few words regarding the proposed implementation. `httptrace` may look like an easy place to measure this, but it will only work for HTTP requests and maybe WebSockets. It won't work for gRPC, Redis, or any other protocols we will want to introduce in the future, unless we figure out how to add it for every one of them too. Also, `httptrace` probably won't "understand" when k6 returns a cached result according to its [`--dns` resolution options](https://k6.io/docs/using-k6/k6-options/reference/#dns) and when it returns a fresh one, i.e. you are probably going to see a lot of `0` values measured due to instantaneous results.

So, it seems to me that the best place to add this metric would be the [`lib/netext/resolver.go`](https://github.com/grafana/k6/blob/master/lib/netext/resolver.go#L69-L70), though I am not sure what complications that would have, @imiric would know better than me. The one I can think of is that, while we know if k6 has cached the DNS result there, we don't actually know whether Go or the OS have cached it beyond us, so we will probably still get a lot of near-0 values for this new metric :disappointed: To fully prevent that and only measure non-cached lookups, we probably need https://github.com/grafana/k6/issues/1751, which is a much bigger project. Though I've noticed that the Go team are constantly making improvements to the native Go DNS resolver (e.g. see [Go 1.19](https://tip.golang.org/doc/go1.19#net) and [1.20](https://tip.golang.org/doc/go1.20#net) changelogs), so maybe something has changed and we wouldn't need to implement it from scratch :thinking:

mem:
Hi @na-- 

thanks for taking the time to look at this. And thanks for digging up the the PRs and issues.

> Finally, a few words regarding the proposed implementation. `httptrace` may look like an easy place to measure this, but it will only work for HTTP requests and maybe WebSockets. It won't work for gRPC, Redis, or any other protocols we will want to introduce in the future, unless we figure out how to add it for every one of them too. Also, `httptrace` probably won't "understand" when k6 returns a cached result according to its [`--dns` resolution options](https://k6.io/docs/using-k6/k6-options/reference/#dns) and when it returns a fresh one, i.e. you are probably going to see a lot of `0` values measured due to instantaneous results.

Yeah, I probably worked at this backwards: I added the missing hook to the http tracer, and it didn't work (it was never called), and I started to look at why it didn't work. Since I'm familiar with that implementation, it was relatively straightforward to figure out what was going on: name resolution is started by the HTTP client, and the implementation is such that the context that the resolver ends up using doesn't have the hooks in place. I did think about modifying the resolver so that it performs the measurement itself, but I couldn't quite spot how to hook it to the correct location to actually add the observations. I probably need to spend more time with this code.

Regarding the aggressive caching that is in place... I'm probably missing something here, but it seems to me that the caching layer is ignoring the TTL in the DNS responses, and in particular it seems to be ignoring the case where the DNS server is specifying that the answer should not be cached... because it's not talking to a DNS server in the first place (if it doesn't have a cached response it's falling back to the system's name resolution mechanism)?

So you are right, there will be cases where the timings for DNS resolution are very low. As long as the caching layer does as asked, my opinion is that that's fine. That possibly mimicks the normal operation of a system correctly (some responses are locally cached, some responses are cached at the system level, etc). For example, my work computer has a caching forwarder in place that is in turn talking to another local caching forwarder.

(As a side note, the reason why I ended up here is because I'm interested in reporting end-to-end timings, i.e. the time `http.Do(req)` takes to complete, and I noticed a large discrepancy between that duration and the duration reported by `http_req_duration` -- I think that's the name -- and I was trying to figure out where that is coming from)

> So, it seems to me that the best place to add this metric would be the [`lib/netext/resolver.go`](https://github.com/grafana/k6/blob/master/lib/netext/resolver.go#L69-L70), though I am not sure what complications that would have, @imiric would know better than me.

This part makes sense. I would just need to figure out the mechanics of getting the observations to the correct place.

> The one I can think of is that, while we know if k6 has cached the DNS result there, we don't actually know whether Go or the OS have cached it beyond us, so we will probably still get a lot of near-0 values for this new metric disappointed To fully prevent that and only measure non-cached lookups, we probably need #1751, which is a much bigger project. Though I've noticed that the Go team are constantly making improvements to the native Go DNS resolver (e.g. see [Go 1.19](https://tip.golang.org/doc/go1.19#net) and [1.20](https://tip.golang.org/doc/go1.20#net) changelogs), so maybe something has changed and we wouldn't need to implement it from scratch thinking

This part I'm not fully following. I'm probably missing the intention.

If you want to avoid caches, you have to go ask the DNS server and honor the TTL in the response. I'm pretty sure that would break in many situations (e.g. netadmins not wanting people to talk directly to DNS servers and enforcing the use of local _caching_ forwarders; that's solved by specifying the DNS server to use).

So the part I'm confused about is where Microsoft Windows comes into play...

mem:
Thank you for the discussion, this requires more planning to get the issue fixed.

Issue #2954, opened 2023-03-06
cloud: Binary-based ingestion


codebien:
## What

The current Cloud ingestion service receives the metrics on the `CLOUD_URL/v1/metrics/<TEST_REF_ID>` endpoint. Eeach HTTP request contains a JSON payload with an array of Sample at the root and each Sample contains a `data` field where the type is one of the types defined below.

We want replace it implementing a new HTTP body format using a binary encoding.

### Current JSON payload

```json
[
  {
    "type": "<TYPE>",
    "metric": "<NAME>",
    "data": {
      ...
    },
  },{
  ...
  }
]
```

### Single point

<details>
  <summary>Show me the JSON</summary>
  
  ```js
  {
  "type": "Point",
  "metric": "vus",
  "data": {
    "time": "%d",
    "type": "gauge",
    "tags": {
      "aaa": "bbb",
      "ccc": "123"
    },
    "value": 999
  }
  }
  ```
</details>

### Multi points

<details>
  <summary>Show me the JSON</summary>
  
  ```js
  {
  "type": "Points",
  "metric": "iter_li_all",
  "data": {
    "time": "%d",
    "type": "counter",
    "tags": {
      "test": "mest"
    },
    "values": {
      "data_received": 6789.1,
      "data_sent": 1234.5,
      "iteration_duration": 10000
    }
  }
}
  ```
</details>

### Aggregated points

<details>
  <summary>Show me the JSON</summary>
  
  ```js
  {
  "type": "AggregatedPoints",
  "metric": "http_req_li_all",
  "data": {
    "time": "%d",
    "type": "aggregated_trend",
    "count": 2,
    "tags": {
      "test": "mest"
    },
    "values": {
      "http_req_duration": {
        "min": 0.013,
        "max": 0.123,
        "avg": 0.068
      },
      "http_req_blocked": {
        "min": 0.001,
        "max": 0.003,
        "avg": 0.002
      },
      "http_req_connecting": {
        "min": 0.001,
        "max": 0.002,
        "avg": 0.0015
      },
      "http_req_tls_handshaking": {
        "min": 0.003,
        "max": 0.004,
        "avg": 0.0035
      },
      "http_req_sending": {
        "min": 0.004,
        "max": 0.005,
        "avg": 0.0045
      },
      "http_req_waiting": {
        "min": 0.005,
        "max": 0.008,
        "avg": 0.0065
      },
      "http_req_receiving": {
        "min": 0.006,
        "max": 0.008,
[...]

PR #2963, opened 2023-03-08
output/cloudv2: Binary-based payload

codebien:
It adds the HTTP client for flushing the metric to the remote cloud service and the new Protobuf-based required payload.

Updates #2954


cloudapi/config.go
```diff
@@ -28,6 +28,7 @@ type Config struct {
 	TestRunDetails null.String `json:"testRunDetails" envconfig:"K6_CLOUD_TEST_RUN_DETAILS"`
 	NoCompress     null.Bool   `json:"noCompress" envconfig:"K6_CLOUD_NO_COMPRESS"`
 	StopOnError    null.Bool   `json:"stopOnError" envconfig:"K6_CLOUD_STOP_ON_ERROR"`
+	UseVersion2    null.Bool   `json:"useVersion2" envconfig:"K6_CLOUD_USE_VERSION_2"`
```
na--, line 31: Maybe this can be something a bit more flexible?

```suggestion
	UsedAPIVersion    null.Int   `json:"usedVersion" envconfig:"K6_CLOUD_USED_VERSION"`
```

And, if the used version is higher than the supported one, k6 can exit with an error? :thinking:

output/cloud/output.go
```diff
@@ -45,6 +46,8 @@ type Output struct {
 	logger logrus.FieldLogger
 	opts   lib.Options
 
+	outv2 *expv2.Output
```
na--, line 49:
```suggestion
	SomeCloudOutputSubInterface
```

Instead of this embedded `*expv2.Output` and all of the `if !out.config.UseVersion2.Bool` and `if out.outv2` statements sprinkled in all of the methods, it seems to me that it would be easier to add a new small `interface` type here with a few methods. Potentially a sub-set of the `output.Output` interface.

Then embed that type in the struct and have 2 implementations of that interface - one for the old json-based cloud output and one for the new protobuf one. `newOutput()` can instantiate the correct one and save it in the struct instance it returns.

cloudapi/config.go
codebien: Addressed in https://github.com/grafana/k6/pull/2963/commits/ea2c1ef47959ab3ae81474440aa89009a34f86ae

codebien:
@na-- @olegbespalov I added the aggregation proposal also, it shouldn't be optimal but it should be a good first step to achieve a reasonable rate. We can iterate on it after we will get some benchmarks (I don't plan them for this PR).

I would get the reviews for it before I do the refactor asked in https://github.com/grafana/k6/pull/2963#discussion_r1157118508 so I can address it all at once.

output/cloud/expv2/output.go
```diff
@@ -0,0 +1,265 @@
...
+		metricSet = append(metricSet, o.mapMetricProto(m, aggr))
+		aggr.Clean()
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), o.config.MetricPushInterval.TimeDuration())
+	defer cancel()
+
+	err := o.client.Push(ctx, o.referenceID, &pbcloud.MetricSet{Metrics: metricSet})
+	if err != nil {
+		o.logger.Error(err)
```
olegbespalov, line 113: Let's be more precise here (instead of prefixing some of the errors inside the `Push` method with `failed to push metrics:`). 

That way we avoid any bare/unwrapped error from inside.

olegbespalov (commented):
I've done a first round, will do the other one a bit later

output/cloud/expv2/metrics_client_test.go
```diff
@@ -0,0 +1,96 @@
...
+
+		assert.Equal(t, "/v2/metrics/test-ref-id", r.URL.Path)
+		assert.Equal(t, http.MethodPost, r.Method)
+		assert.Equal(t, "Token fake-token", r.Header.Get("Authorization"))
+		assert.Contains(t, r.Header.Get("User-Agent"), "k6cloud/v0.4")
+		assert.Equal(t, "application/x-protobuf", r.Header.Get("Content-Type"))
+		assert.Equal(t, "snappy", r.Header.Get("Content-Encoding"))
+		assert.Equal(t, "2.0", r.Header.Get("K6-Metrics-Protocol-Version"))
+		assert.Equal(t, "2.0", r.Header.Get("K6-Metrics-Protocol-Version"))
+		b, err := io.ReadAll(r.Body)
```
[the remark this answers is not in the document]
mstoykov: [ioutil.ReadAll](https://pkg.go.dev/io/ioutil#ReadAll) is the deprecated one ;)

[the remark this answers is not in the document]
olegbespalov: My bad :facepalm:

output/cloud/expv2/pbcloud/metric.proto
```diff
@@ -0,0 +1,123 @@
...
+
+package metrics;
+
+option go_package = "go.k6.io/k6/otput/cloud/expv2/pbcloud";
+
+import "google/protobuf/timestamp.proto";
+
+// The type of a Metric.
+enum MetricType {
+  METRICTYPE_UNSPECIFIED = 0;
```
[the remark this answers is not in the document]
codebien: You can have multiple enums without the prefix I guess it would conflict.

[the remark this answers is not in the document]
codebien: The error here is not using the prefix also for the other items that is a suggested style guide https://protobuf.dev/programming-guides/style/#enums

output/cloud/expv2/output.go
```diff
@@ -0,0 +1,265 @@
...
+// StopWithTestError stops the output.
+func (o *Output) StopWithTestError(testErr error) error {
+	o.logger.Debug("Stopping...")
+	defer o.logger.Debug("Stopped!")
+	o.periodicFlusher.Stop()
+	return nil
+}
+
+// SetReferenceID sets the Cloud's test ID.
+func (o *Output) SetReferenceID(refID string) {
```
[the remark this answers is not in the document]
codebien: Because the reference ID is generated when we _Start_ the output and not when we instantiate the output. The reference id is generated from the cloud backend.

output/cloud/expv2/pbcloud/metric.proto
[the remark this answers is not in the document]
codebien: [019b27a](https://github.com/grafana/k6/pull/2963/commits/019b27a29df6c7ef8b8620ac37d729b4f9e0bd31)

output/cloud/expv2/metrics_client_test.go
```diff
@@ -0,0 +1,96 @@
...
+		reqs++
+
+		assert.Equal(t, "/v2/metrics/test-ref-id", r.URL.Path)
+		assert.Equal(t, http.MethodPost, r.Method)
+		assert.Equal(t, "Token fake-token", r.Header.Get("Authorization"))
+		assert.Contains(t, r.Header.Get("User-Agent"), "k6cloud/v0.4")
+		assert.Equal(t, "application/x-protobuf", r.Header.Get("Content-Type"))
+		assert.Equal(t, "snappy", r.Header.Get("Content-Encoding"))
+		assert.Equal(t, "2.0", r.Header.Get("K6-Metrics-Protocol-Version"))
+		assert.Equal(t, "2.0", r.Header.Get("K6-Metrics-Protocol-Version"))
```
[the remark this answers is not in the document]
olegbespalov: I still can see that line 41 duplicates line 40

```
assert.Equal(t, "2.0", r.Header.Get("K6-Metrics-Protocol-Version"))
```

output/cloud/output.go
```diff
@@ -200,18 +202,36 @@ func (out *Output) Start() error {
...
 		"duration":    out.duration,
 		"referenceId": out.referenceID,
+		"apiVersion":  out.config.APIVersion.Int64,
 	}).Debug("Started!")
 	return nil
 }
 
-func (out *Output) startBackgroundProcesses() {
+func (out *Output) startBackgroundProcesses() error {
+	if out.config.APIVersion.Int64 == 2 {
```
olegbespalov, line 221: I know that is obvious, and we probably don't have many versions, but I suggest avoiding using the "magic" number 2 and creating a constant. 

That way, we keep code more maintainable in the future since any other maintainer/contributor will easily find all such conditional places and won't spend (or miss) the places where we compare with some number.

output/cloud/expv2/output.go
```diff
@@ -0,0 +1,262 @@
...
+	o.referenceID = refID
+}
+
+// AddMetricSamples receives a set of metric samples.
```
olegbespalov, line 81: Probably it's an old comment :thinking:

output/cloud/output.go
codebien: @na-- I made an attempt in the last commits, they are just the final result of the process that can be fine in the history here: https://github.com/grafana/k6/tree/cloudv2

I've created a dedicated PR for the versioning gateway so it should be clear and easier to get the architecture and merge with a different timing https://github.com/grafana/k6/pull/3041

output/cloud/expv2/output.go
```diff
@@ -0,0 +1,324 @@
...
+	pbseries := make([]*pbcloud.TimeSeries, 0, len(as.Samples))
+	for ts, samples := range as.Samples {
+		if ts.Metric.Type == metrics.Trend {
+			// skip trend metrics
+			continue
+		}
+		pb := pbcloud.TimeSeries{}
+		// TODO: optimize removing Map
+		// and using https://github.com/grafana/k6/issues/2764
+		pb.Labels = make([]*pbcloud.Label, 0, ((*atlas.Node)(ts.Tags)).Len())
```
mstoykov, line 267: nit:
```suggestion
		pb.Labels = make([]*pbcloud.Label, 0, ((*atlas.Node)(ts.Tags)).Len()+2)
```

output/cloud/output.go
codebien: This is resolved now on the [root PR](https://github.com/grafana/k6/pull/3072) in the https://github.com/grafana/k6/pull/3072/commits/448da4c3217ccff1b21c07462724e13e628bc075 commit

output/cloud/expv2/metrics_client_test.go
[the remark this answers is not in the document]
codebien: The last push resolved it

output/cloud/expv2/metrics_client.go
```diff
@@ -0,0 +1,125 @@
...
+
+	resp, err := mc.httpClient.Do(req)
+	if err != nil {
+		return err
+	}
+	defer func() {
+		_ = resp.Body.Close()
+	}()
+	if resp.StatusCode != http.StatusOK {
+		return fmt.Errorf("response got an unexpected status code: %s", resp.Status)
```
olegbespalov, line 102: Is there any convention in BE, on how they structure error responses? I mean could we maybe try to extract something from there?

output/cloud/expv2/metrics_client.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+	mc.logger.WithField("t", time.Since(start)).WithField("size", len(b)).
+		Debug("Pushed the collected metrics to the Cloud service")
+	return nil
+}
+
+const b100KiB = 100 * 1024
```
olegbespalov, line 109: Let's maybe name this more precisely, like `cloudProcessingLimit` ?

codebien: https://github.com/grafana/k6/pull/2963/commits/3d401fe96daca1144d06ff0ecb89747a5c914e68

output/cloud/expv2/metrics_client.go
olegbespalov: We can do that in a separated PR

olegbespalov (dismissed):

output/cloud/expv2/metrics_client.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+	if token == "" {
+		return nil, errors.New("token is required")
+	}
+	return &MetricsClient{
+		httpClient: &http.Client{Timeout: 5 * time.Second},
+		logger:     logger,
+		baseURL:    host + "/v2/metrics/",
+		token:      token,
+		userAgent:  "k6cloud/v" + consts.Version,
```
olegbespalov, line 51: Isn't the `consts.Version` already has a `v` prefix? :thinking:

na--: Nope: https://github.com/grafana/k6/blob/4e8347ee4b12c65069e55e7c7fe82ec9aa57d777/lib/consts/consts.go#L11

na-- (dismissed):

output/cloud/expv2/metrics_client.go
olegbespalov: Right :facepalm: :+1:

output/cloud/expv2/metrics_client.go
```diff
@@ -0,0 +1,127 @@
...
+	if len(b) > payloadSizeLimit {
+		return nil, fmt.Errorf("the Protobuf message is too large to be handled from the Cloud processor; "+
+			"size: %d, limit: 100 KB", len(b))
+	}
+	if snappy.MaxEncodedLen(len(b)) < 0 {
+		return nil, fmt.Errorf("the Protobuf message is too large to be handled by Snappy encoder; "+
+			"size: %d, limit: %d", len(b), 0xffffffff)
+	}
+	return snappy.Encode(nil, b), nil
+}
```
mstoykov, line 127: Here we have:
1. []byte that is the marshaled data
2. second []byte that is the snappied data

I would expect both won't be *that* big, but if possible I would prefer if we can user writers and readers and preferably have only 1 final copy of the `[]byte` or even the writing to happen in stream fashion.

This *does* depend on whether the backend support streaming snappy format as there are two different formats.

olegbespalov (approved):

output/cloud/expv2/metrics_client.go
```diff
@@ -0,0 +1,127 @@
...
+
+	if err := cloudapi.CheckResponse(resp); err != nil {
+		return err
+	}
+	mc.logger.WithField("t", time.Since(start)).WithField("size", len(b)).
+		Debug("Pushed the collected metrics to the Cloud service")
+	return nil
+}
+
+const payloadSizeLimit = 100 * 1024 // 100 KiB
```
mstoykov, line 111: Shouldn't we just break them in parts then? 100kb doesn't soudn *that* big. Especially before compression.

output/cloud/expv2/metrics_client.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+
+	buf, _ := mc.pushBufferPool.Get().(*bytes.Buffer)
+	buf.Reset()
+	defer mc.pushBufferPool.Put(buf)
+
+	_, err = buf.Write(b)
+	if err != nil {
+		return err
+	}
```
mstoykov, line 79: Unless we are writing directly to the `buf` this is pointless and actually harmful. 

You already created the `[]byte` slice and now you are going to put it in a pool so it will survive longer. Only to get it out and not use it but just write on top of it.

If we are not going to be writing to it (seem my comments on `newRequestBody`) it is better to just use `bytes.Reader` and skip the pool.

output/cloud/expv2/metrics_client.go
```diff
@@ -0,0 +1,127 @@
...
+type MetricsClient struct {
+	httpClient httpDoer
+	logger     logrus.FieldLogger
+	token      string
+	userAgent  string
+
+	pushBufferPool sync.Pool
+	baseURL        string
+}
+
```
mstoykov, line 37: Does this need to be public?

go.mod
```diff
@@ -12,6 +12,7 @@ require (
 	github.com/fatih/color v1.15.0
 	github.com/go-sourcemap/sourcemap v2.1.4-0.20211119122758-180fcef48034+incompatible
 	github.com/golang/protobuf v1.5.3
+	github.com/golang/snappy v0.0.4
```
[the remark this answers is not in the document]
codebien: I didn't know about it, going to change it

output/cloud/expv2/metrics_client.go
codebien: How would you do that? Here the limit is related to the protobuf, how do you chunk it without breaking the metric set message? I'm not aware of Protobuf supporting a streaming/chunked version. Did I miss it?

This should be an extreme check, if we need to chunk, then I think we should do that before calling the metric's client, so in its caller. WDYT?

go.mod
```diff
@@ -62,6 +61,7 @@ require (
 	github.com/dlclark/regexp2 v1.9.0 // indirect
 	github.com/fsnotify/fsnotify v1.5.4 // indirect
 	github.com/go-redis/redis/v8 v8.11.5 // indirect
+	github.com/golang/snappy v0.0.4 // indirect
```
codebien: Now, we should do the same with the Prometheus remote write so we can drop it.

go.mod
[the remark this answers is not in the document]
codebien: Done

output/cloud/expv2/metrics_client.go
codebien: I added a TODO for doing and testing in a specific PR so we don't block this.

output/cloud/expv2/metrics_client.go
codebien: You're right, this is some remaining stuff from the old client. Done.

output/cloud/expv2/metrics_client.go
codebien: Done

go.mod
mstoykov: Just make certain to test it as it might turn out there are incompatibilities ... although I hope not :crossed_fingers:

codebien: Added a issue there

output/cloud/expv2/metrics_client.go
mstoykov: You can use https://pkg.go.dev/google.golang.org/protobuf/proto#Size to find out how big it will be *before* serializing and then splitting if necessary.

And I guess a method like Push can know that protobuf will be used and cut it in pieces when necessary.

We can probably leave that for after we merge everything. But I would like to know how many metrics are 100kb before we actually make this the default. As dropping your metrics because they were 101kb sounds like terrible experience.

mstoykov (approved):
LGTM! But we should test how big is 100kb encoded metrics before we actually make this the default way to push metrics to the cloud.

codebien:
>But we should test how big is 100kb encoded metrics

Yes, I already have it on my list. I add it also to the worklog in https://github.com/grafana/k6/issues/2954 as a reminder.

olegbespalov (approved):

Issue #2667, opened 2022-08-26
Request modification


TheAsda:
### Feature Description

We have specific authentication flow that requires to attach custom header to every request. I wrote extension that can generate header but attaching it to request is manual process. Sample:
```js
import http from 'k6/http';
import auth from 'k6/x/auth';

export function setup() {
  auth.setup();
}

export default function () {
  const header = auth.getHeader();
  http.get('<some url>', {
    headers: {
      Auth: header,
    },
  });
}
```
Header has to be generated for every request so i cannot initialize it once in _setup_. It would be great to be able to somehow modify request before execution.

### Suggested Solution (optional)

I see a couple of options here:
1. Add ability to attach some function for modification request like
```js
import http from 'k6/http';
import auth from 'k6/x/auth';

http.applyMiddleware((req) => {
  req.headers.Auth = auth.getHeader();
  return req;
});
```

2. Add ability to affect RoundTripper on Go level:
```go
func init() {
	custom := Custom{
		original: ctx.Transport,
	}
	ctx.Transport = custom
}

type Custom struct {
	original http.RoundTripper
}

func (c Custom) RoundTrip(req *http.Request) (*http.Response, error) {
	header := getHeader()
	req.Header.Add("Auth", header)
	return c.original.RoundTrip(req)
}

// or

func init() {
	modules.AddRoundTripper(func(original http.RoundTripper) http.RoundTripper {
		return Custom{
			original: original,
		}
	})
}
```

Personally i'd prefer second option as it can give more flexibility.

TheAsda:
Another option might be registration of custom auth method like
```go
modules.RegisterAuth("custom", func(req *http.Request) *http.Request {
	//...
})
```

```js
http.get('<some url>', { auth: 'custom' });
```

na--:
Thanks for opening this issue and sorry for the slow reply! It's a known problem that was previously mentioned in https://github.com/grafana/k6/issues/761 (and probably in other [`new-http` issues](https://github.com/grafana/k6/issues?page=1&q=is%3Aopen+is%3Aissue+label%3Anew-http)) that we'd like to address with a new k6 HTTP API soon (https://github.com/grafana/k6/issues/2461).

For now, the `httpx` JavaScript library that we host on [jslib.k6.io](https://jslib.k6.io/) might be sufficient for your needs. It's a wrapper around the existing `k6/http` API, take a look:
- https://k6.io/docs/javascript-api/jslib/httpx
[...]

PR #2971, opened 2023-03-10
Design document for the new HTTP API

imiric:
This is a proposal/design document for a new implementation of the JavaScript HTTP API in k6.

The team has been discussing it internally for a few weeks, but the new API has been in discussion for many years prior to this. It's finally time to start working on it, so we want to share our ideas with the k6 community, and invite opinions from existing k6 users. Please be mindful to keep discussion thoughtful and relevant, in order to minimize noise.

### DIFF
--- a/docs/design/018-new-http-api.md
+++ b/docs/design/018-new-http-api.md
@@ -88,7 +88,6 @@ The Socket state can either be _active_—meaning connected for a TCP socket, bo
 - TCP:
 ```javascript
 import { dialTCP } from 'k6/x/net';
-import { Client } from 'k6/x/net/http';
 
 export default async function () {
   const socket = await dialTCP('192.168.1.1:80', {
@@ -117,7 +116,6 @@ export default async function () {
 - UDP:
 ```javascript
 import { dialUDP } from 'k6/x/net';
-import { Client } from 'k6/x/net/http';
 
 export default async function () {
   const socket = new dialUDP('192.168.1.1:9090');
@@ -128,7 +126,7 @@ export default async function () {
 
 - IPC:
 ```javascript
-import { dialIPC } from 'k6/x/net;
+import { dialIPC } from 'k6/x/net';
 import { Client } from 'k6/x/net/http';
 
 export default async function () {
@@ -168,7 +166,7 @@ import { dialTCP } from 'k6/x/net';
 import { Client } from 'k6/x/net/http';
 
 export default async function () {
-  const socket = await dialTCP('10.0.0.10:80, { keepAlive: true });
+  const socket = await dialTCP('10.0.0.10:80', { keepAlive: true });
   const client = new Client({
     socket: socket,
     proxy: 'https://myproxy',
@@ -287,7 +285,7 @@ The [Fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) is a
 The implementation in k6 differs slightly from the web API, but we've tried to make it familiar to use wherever possible.
 
 Example:
-```
+```javascript
 import { fetch } from 'k6/x/net/http';
 
 export default async function () {
--- a/docs/design/018-new-http-api.md
+++ b/docs/design/018-new-http-api.md
@@ -1,0 +1,248 @@
+# New HTTP API
+
+## Authors
+The k6 core team
+
+## Why is this needed?
+
+The HTTP API (in k6 <=v0.43.0) used by k6 scripts has many limitations, inconsistencies and performance issues, that lead to a poor user experience. Considering that it's the most commonly used JS API, improving it would benefit most k6 users.
+
+The list of issues with the current API is too long to mention in this document, but you can see a detailed list of [GitHub issues labeled `new-http`](https://github.com/grafana/k6/issues?q=is%3Aopen+is%3Aissue+label%3Anew-http) that should be fixed by this proposal, as well as the [epic issue #2461](https://github.com/grafana/k6/issues/2461). Here we'll only mention the relatively more significant ones:
+
+* [#2311](https://github.com/grafana/k6/issues/2311): files being uploaded are copied several times in memory, causing more memory usage than necessary. Related issue: [#1931](https://github.com/grafana/k6/issues/1931)
+* [#857](https://github.com/grafana/k6/issues/857), [#1045](https://github.com/grafana/k6/issues/1045): it's not possible to configure transport options, such as proxies or DNS, per VU or group of requests.
+* [#761](https://github.com/grafana/k6/issues/761): specifying configuration options globally is not supported out-of-the-box, and workarounds like the [httpx library](https://k6.io/docs/javascript-api/jslib/httpx/) are required.
+* [#746](https://github.com/grafana/k6/issues/746): async functionality like Server-sent Events is not supported.
+* Related to the previous point, all (except asyncRequest) current methods are synchronous, which is inflexible, and doesn't align with modern APIs from other JS runtimes.
+* [#436](https://github.com/grafana/k6/issues/436): the current API is not very friendly or ergonomic. Different methods also have parameters that change places, e.g. `params` is the second argument in `http.get()`, but the third one in `http.post()`.
+
+
+## Proposed solution(s)
+
+### Design
+
+In general, the design of the API should follow these guidelines:
+
+- It should be familiar to users of HTTP APIs from other JS runtimes, and easy for new users to pick up.
+
+  As such, it would serve us well to draw inspiration from existing runtimes and frameworks. Particularly:
+
+  - The [Fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API), a [WHATWG standard](https://fetch.spec.whatwg.org/) supported by most modern browsers.
+    [Deno's implementation](https://deno.land/manual/examples/fetch_data) and [GitHub's polyfill](https://github.com/github/fetch) are good references to follow.
+
+    This was already suggested in [issue #2424](https://github.com/grafana/k6/issues/2424).
+
+  - The [Streams API](https://developer.mozilla.org/en-US/docs/Web/API/Streams_API), a [WHATWG standard](https://streams.spec.whatwg.org/) supported by most modern browsers.
+    [Deno's implementation](https://deno.land/manual@v1.30.3/examples/fetch_data#files-and-streams) is a good reference to follow.
+
+    There's a related, but very old [proposal](https://github.com/grafana/k6/issues/592) before the Streams API was standardized, so we shouldn't use it, but it's clear there's community interest in such an API.
+
+    Streaming files both from disk to RAM to the network, and from network to RAM and possibly disk, would also partly solve our [performance and memory issues with loading large files](https://github.com/grafana/k6/issues/2311).
+
+  - Native support for the [`FormData` API](https://developer.mozilla.org/en-US/docs/Web/API/FormData).
+
+    Currently this is supported with a [JS polyfill](https://k6.io/docs/examples/data-uploads/#advanced-multipart-request), which should be deprecated.
+
+  - Aborting requests or any other async process with the [`AbortSignal`/`AbortController` API](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal), part of the [WHATWG DOM standard](https://dom.spec.whatwg.org/#aborting-ongoing-activities).
+
+    This is slightly out of scope for the initial phases of implementation, but aborting async processes like `fetch()` is an important feature.
+
+- The Fetch API alone would not address all our requirements (e.g. specifying global and transport options), so we still need more flexible and composable interfaces.
+
+  One source of inspiration is the Go `net/http` package, which the k6 team is already familiar with. Based on this, our JS API could have similar entities:
+
+  - `Dialer`: a low-level interface for configuring TCP/IP options, such as TCP timeouts, keep-alive duration, DNS, and IP version preferences.
+
+  - `Transport`: interface for configuring HTTP connection options, such as proxies, TLS, HTTP version preferences, etc.
+
+    It enables advanced behaviors like intercepting requests before they're sent to the server.
+
+  - `Client`: the main entrypoint for making requests, it encompasses all of the above options. A k6 script should be able to initialize more than one `Client`, each with their separate configuration.
+
+    In order to simplify the API, the creation of a `Client` should use sane defaults for `Dialer` and `Transport`.
+
+  There should be some research into existing JS APIs that offer similar features (e.g. Node/Deno), as we want to offer an API familiar to JS developers, not necessarily Go developers.
+
+  - `Request`/`Response`: represent objects sent by the client, and received from the server. In contrast to the current API, the k6 script should be able to construct `Request` objects declaratively, and then reuse them to make multiple requests with the same (or similar) data.
+
+- All methods that perform I/O calls must be asynchronous. Now that we have `Promise`, event loop and `async`/`await` support natively in k6, there's no reason for these to be synchronous anymore.
+
+- The API should avoid any automagic behavior. That is, it should not attempt to infer desired behavior or options based on some implicit value.
+
+  We've historically had many issues with this ([#878](https://github.com/grafana/k6/issues/878), [#1185](https://github.com/grafana/k6/issues/1185)), resulting in confusion for users, and we want to avoid it in the new API. Even though we want to have sane defaults for most behavior, instead of guessing what the user wants, all behavior should be explicitly configured by the user. In cases where some behavior is ambiguous, the API should raise an error indicating so.
+
+
+### Implementation
+
+Trying to solve all `new-http` issues with a single large and glorious change wouldn't be reasonable, so improvements will undoubtedly need to be done gradually, in several phases, and over several k6 development cycles.
+
+With this in mind, we propose the following phases:
+
+#### Phase 1: create initial k6 extension
+
+**Goals**:
+
+- Implement a barebones async API that serves as a proof-of-concept for what the final developer experience will look and feel like.
+  The code should be in a state that allows it to be easily extended.
+
+  By barebones, we mean:
+
+  - The `Client` interface with only one method: `request()`, which will work similarly to the current `http.asyncRequest()`.
+
+    For the initial PoC, it's fine if only `GET` and `POST` methods are supported.
+
+    It's not required to make `Dialer` and `Transport` fully configurable at this point, but they should use sane defaults, and it should be clear how the configuration will be done.
+
+- This initial API should solve a minor, but concrete, issue of current API. It should fix something that's currently not possible and doesn't have a good workaround.
+
+  Good candidates: [#936](https://github.com/grafana/k6/issues/936), [#970](https://github.com/grafana/k6/issues/970).
+
+- Features like configuring options globally, or per VU or request, should be implemented.
+  Deprecating the `httpx` library should be possible after this phase.
+
+
+**Non-goals**:
+
+- We won't yet try to solve performance/memory issues of the current API, or implement major new features like data streaming.
+
+
+#### Phase 2: work on major issues
+
+**Goals**:
+
+- Work should be started on some of the most impactful issues from the current API.
+  Issues like high memory usage when uploading files ([#2311](https://github.com/grafana/k6/issues/2311)), and data streaming ([#592](https://github.com/grafana/k6/issues/592)), are good candidates to focus on first.
+
+
+#### Phase 3: work on leftover issues
+
+**Goals**:
+
+- All leftover `new-http` issues should be worked on in this phase.
+  **TODO**: Specify which issues and in what order should be worked on here.
+
+- The extension should be thoroughly tested, by both internal and external users.
+
+
+#### Phase 4: expand, polish and stabilize the API
+
+**Goals**:
+
+- The API should be expanded to include all HTTP methods supported by the current API.
+  For the most part, it should reach feature parity with the current API.
+
+- A standalone `fetch()` function should be added that resembles the web Fetch API. There will be some differences in the options compared to the web API, as we want to make parts of the transport/client configurable.
+
+    Internally, this function will create a new client (or reuse a global one?), and will simply act as a convenience wrapper over the underlying `Client`/`Dialer`/`Transport` implementations, which will be initialized with sane default values.
+
+- Towards the end of this phase, the API should be mostly stable, based on community feedback.
+  Small changes will be inevitable, but there should be no discussion about the overall design.
+
+
+#### Phase 5: merge into k6-core, more testing
+
+At this point the extension should be relatively featureful and stable to be useful to all k6 users.
+
+**Goals**:
+
+- Merge the extension into k6 core, and make it available to k6 Cloud users.
+
+- Continue to gather and address feedback from users, thorough testing and polishing.
+
+
+#### Phase 6: deprecate `k6/http`
+
+As the final phase, we should add deprecation warnings when `k6/http` is used, and point users to the new API.
+Eventually, months down the line, we can consider replacing `k6/http` altogether with the new module.
+
+
+### Examples
+
+- `basic-get.js`:
+  ```javascript
+  import { fetch } from 'k6/x/net/http';
+
+  export default async function () {
+    // Creates a new default Client internally.
+    const response = await fetch('https://httpbin.test.k6.io/get');
+    const jsonData = await response.json();
+    console.log(jsonData);
+  }
+  ```
+
+- `basic-post.js`:
+  ```javascript
+  import { fetch } from 'k6/x/net/http';
+
+  export default async function () {
+    await fetch('https://httpbin.test.k6.io/post', {
+      method: 'POST',
+      json: { name: 'k6' }, // automatically adds 'Content-Type: application/json' header
+    });
+  }
+  ```
+
+- `basic-get-request.js`:
+  ```javascript
+  import { fetch, Request } from 'k6/x/net/http';
+
+  export default async function () {
+    const request = new Request('https://httpbin.test.k6.io/get', {
+      headers: { 'Case-Sensitive-Header': 'somevalue' },
+      // Other options that can also be passed directly to fetch()
+    });
+    const response = await fetch(request, {
+      // Some options that will merge with or override the options specified
+      // in the Request.
+    });
+    const jsonData = await response.json();
+    console.log(jsonData);
+  }
+  ```
+
+- `advanced-get.js`:
+  ```javascript
+  import { Client } from 'k6/x/net/http';
+
+  const client = new Client({
+    proxy: 'https://myproxy',
+    forceHTTP1: true,
+    dns: { ... },
+    // other transport options: h2c, TLS, etc.
+  });
+
+  export default async function () {
+    const response = await client.get('https://httpbin.test.k6.io/get');
+    const jsonData = await response.json();
+    console.log(jsonData);
+  }
+  ```
+
+- `advanced-post-streaming.js`:
+  ```javascript
+  import { File } from 'k6/x/file';
+  import { fetch } from 'k6/x/net/http';
+
+  // Will need supporting await in init context
+  const file = await File.open('./logo.svg');  // by default assumes 'read'
+
+  export default async function () {
+    await fetch('https://httpbin.test.k6.io/post', {
+      method: 'POST',
+      body: file.readable,
+    });
+  }
+  ```
+
+
+## Potential risks
+
+* Long implementation time.
+
+  Not so much of a risk, but more of a necessary side-effect of spreading the work in phases, and over several development cycles. We need this approach in order to have ample time for community feedback, to implement any unplanned features, and to make sure the new API fixes all existing issues.
+  Given this, it's likely that the entire process will take many months, possibly more than a year to finalize.
+
+
+## Technical decisions
+
+TBD after team discussion. In the meantime, see the "Proposed solution(s)" section.
--- a/docs/design/018-new-http-api.md
+++ b/docs/design/018-new-http-api.md
@@ -341,12 +341,15 @@ With this in mind, we propose the following phases:
 - We won't yet try to solve performance/memory issues of the current API, or implement major new features like data streaming.
 
 
-#### Phase 2: work on major issues
+#### Phase 2: work on major issues, merge into k6 core as experimental module
 
 **Goals**:
 
 - Work should be started on some of the most impactful issues from the current API.
   Issues like high memory usage when uploading files ([#2311](https://github.com/grafana/k6/issues/2311)), and data streaming ([#592](https://github.com/grafana/k6/issues/592)), are good candidates to focus on first.
+
+- At the end of this phase the API should resolve major limitations of `k6/http`, and it would be a good time to merge it into k6 core as an experimental module (`k6/experimental/net/http`).
+  This would make it available to more users, including in the k6 Cloud.
 
 
 #### Phase 3: work on leftover issues
@@ -374,21 +377,16 @@ With this in mind, we propose the following phases:
   Small changes will be inevitable, but there should be no discussion about the overall design.
 
 
-#### Phase 5: merge into k6-core, more testing
+#### Phase 5: more testing, deprecating old API
 
 At this point the extension should be relatively featureful and stable to be useful to all k6 users.
 
 **Goals**:
 
-- Merge the extension into k6 core, and make it available to k6 Cloud users.
-
 - Continue to gather and address feedback from users, thorough testing and polishing.
 
-
-#### Phase 6: deprecate `k6/http`
-
-As the final phase, we should add deprecation warnings when `k6/http` is used, and point users to the new API.
-We'll have to maintain both `k6/http` and `k6/net/http` for likely years to come, though any new development will happen in `k6/net/http`, and `k6/http` would only receive bug and security fixes.
+- As the final step, we should add deprecation warnings when `k6/http` is used, and point users to the new API. We can also consider promoting the API from experimental to a main module under `k6/net/http`.
+  We'll have to maintain both `k6/http` and `k6/net/http` for likely years to come, though any new development will happen in `k6/net/http`, and `k6/http` would only receive bug and security fixes.
 
 
 ## Potential risks
--- a/docs/design/018-new-http-api.md
+++ b/docs/design/018-new-http-api.md
@@ -306,6 +306,90 @@ export default async function () {
 }
 ```
 
+#### Events
+
+The new HTTP API will emit events which scripts can subscribe to, in order to implement advanced functionality.
+
+For example, a `requestToBeSent` event is emitted when the request was processed by k6, and just before it is sent to the server. This allows changing the request body or headers, or introducing artificial delays.
+
+```javascript
+import { sleep } from 'k6';
+import { Client } from 'k6/x/net/http';
+
+export default async function () {
+  const client = new Client();
+  client.on('requestToBeSent', event => {
+    console.log(event.type);    // 'requestToBeSent'
+    const request = event.data;
+    request.headers['Cookie'] = 'somecookie=somevalue;'  // overwrites all previously set cookies
+    request.body += ' world!';  // the final body will be 'Hello world!'
+    sleep(Math.random() * 5);   // random delay up to 5s
+  });
+
+  await client.post('https://httpbin.test.k6.io/post', { body: 'Hello' });
+}
+```
+
+Similarly, a `responseReceived` event is emitted when a response is received from the server, but before it's been fully processed by k6. This can be useful to alter the response in some way, or edit the metrics emitted by k6.
+
+For example:
+
+```javascript
+import { Client } from 'k6/x/net/http';
+
+export default async function () {
+  const client = new Client();
+  let requestID;  // used to correlate a specific response with the request that initiated it
+
+  client.on('requestToBeSent', event => {
+    const request = event.data;
+    if (!requestID && request.url == 'https://httpbin.test.k6.io/get?name=k6'
+        && request.method == 'GET') {
+      // The request ID is a UUIDv4 string that uniquely identifies a single request.
+      // This is a contrived check and example, but you can imagine that in a complex
+      // script there would be many similar requests.
+      requestID = request.id;
+    }
+  });
+
+  client.on('responseReceived', event => {
+    const response = event.data;
+    if (requestID && response.request.id == requestID) {
+      // Change the request duration metric to any value
+      response.metrics['http_req_duration'].value = 3.1415;
+      // Consider the request successful regardless of its response
+      response.metrics['http_req_failed'].value = false;
+      // Or drop a single metric
+      delete response.metrics['http_req_duration'];
+      // Or drop all metrics
+      response.metrics = {};
+    }
+  });
+
+  await client.get('https://httpbin.test.k6.io/get', { query: { name: 'k6' } });
+}
+```
+
+Event handlers can also be attached directly to a single request/response cycle. This avoids having to correlate responses with requests as done above.
+
+```javascript
+import { Client } from 'k6/x/net/http';
+
+export default async function () {
+  const client = new Client();
+  await client.get('https://httpbin.test.k6.io/get', {
+    eventHandlers: {
+      'responseReceived': event => {
+        const response = event.data;
+        // ...
+      }
+    }
+  });
+}
+```
+
+**TODO**: List other possible event types, and their use cases.
+
 
 ### Implementation
 
--- a/docs/design/018-new-http-api.md
+++ b/docs/design/018-new-http-api.md
@@ -111,7 +111,7 @@ export default async function () {
   await socket.write(new TextEncoder().encode('GET / HTTP/1.1\r\n\r\n'));
 
   // And reading...
-  socket.on('data', (data) => {
+  socket.on('data', data => {
     console.log(`received ${data}`);
     socket.close();
   });
@@ -125,9 +125,9 @@ export default async function () {
 import { UDP } from 'k6/x/net';
 
 export default async function () {
-  const socket = new UDP.open('192.168.1.1:9090');
-
+  const socket = await UDP.open('192.168.1.1:9090');
   await socket.write(new TextEncoder().encode('GET / HTTP/1.1\r\n\r\n'));
+  socket.close();
 }
 ```
 
@@ -137,15 +137,17 @@ import { IPC } from 'k6/x/net';
 import { Client } from 'k6/x/net/http';
 
 export default async function () {
-  const socket = await IPC.open('/tmp/unix.sock');
-
-  console.log(socket.file.path); // /tmp/unix.sock
-
   // The HTTP client supports communicating over a Unix socket.
   const client = new Client({
-    socket: socket,
+    dial: async () => {
+      return await IPC.open('/tmp/unix.sock');
+    },
   });
   await client.get('http://unix/get');
+
+  console.log(client.socket.file.path); // /tmp/unix.sock
+
+  client.socket.close();
 }
 ```
 
@@ -167,17 +169,17 @@ export default async function () {
 }
 ```
 
-- Passing a socket with custom transport settings, some HTTP options, and making a POST request:
+- Creating a client with custom transport settings, some HTTP options, and making a POST request:
 ```javascript
 import { TCP } from 'k6/x/net';
 import { Client } from 'k6/x/net/http';
 
 export default async function () {
-  const socket = await TCP.open('10.0.0.10:80', { keepAlive: true });
   const client = new Client({
-    socket: socket,
+    dial: async address => {
+      return await TCP.open(address, { keepAlive: true });
+    },
     proxy: 'https://myproxy',
-    version: 1.1,                     // force a specific HTTP version
     headers: { 'User-Agent': 'k6' },  // set some global headers
   });
   await client.post('http://10.0.0.10/post', {
@@ -186,20 +188,42 @@ export default async function () {
 }
 ```
 
-- A tentative HTTP/3 example:
+- Configuring TLS with a custom CA certificate and forcing HTTP/2:
 ```javascript
-import { UDP } from 'k6/x/net';
+import { TCP } from 'k6/x/net';
+import { Client } from 'k6/x/net/http';
+import { open } from 'k6/x/file';
+
+const caCert = await open('./custom_cacert.pem');
+
+export default async function () {
+  const client = new Client({
+    dial: async address => {
+      return await TCP.open(address, {
+        tls: {
+          alpn: ['h2'],
+          caCerts: [caCert],
+        }
+      });
+    },
+  });
+  await client.get('https://10.0.0.10/');
+}
+```
+
+- Forcing unencrypted HTTP/2 (h2c):
+```javascript
+import { TCP } from 'k6/x/net';
 import { Client } from 'k6/x/net/http';
 
 export default async function () {
-  const socket = new UDP.open('192.168.1.1:9090');
-
   const client = new Client({
-    socket: socket,
-    version: 3,      // A UDP socket would imply HTTP/3, but this makes it explicit.
+    dial: async address => {
+      return await TCP.open(address, { tls: false });
+    },
+    version: [2],
   });
-  await client.get('https://httpbin.test.k6.io/get');
-}
+  await client.get('http://10.0.0.10/');
 ```
 
 
--- a/docs/design/018-new-http-api.md
+++ b/docs/design/018-new-http-api.md
@@ -72,6 +72,233 @@ In general, the design of the API should follow these guidelines:
   We've historically had many issues with this ([#878](https://github.com/grafana/k6/issues/878), [#1185](https://github.com/grafana/k6/issues/1185)), resulting in confusion for users, and we want to avoid it in the new API. Even though we want to have sane defaults for most behavior, instead of guessing what the user wants, all behavior should be explicitly configured by the user. In cases where some behavior is ambiguous, the API should raise an error indicating so.
 
 
+#### Sockets
+
+A Socket represents the file or network socket over which client/server or peer communication happens.
+
+It can be of three types:
+- `tcp`: a stream-oriented network socket using the Transmission Control Protocol.
+- `udp`: a message-oriented network socket using the User Datagram Protocol.
+- `ipc`: a mechanism for communicating between processes on the same machine, typically using files.
+
+The Socket state can either be _active_—meaning connected for a TCP socket, bound for a UDP socket, or open for an IPC socket—, or _inactive_—meaning disconnected, unbound, or closed, respectively.
+
+##### Example
+
+- TCP:
+```javascript
+import { dialTCP } from 'k6/x/net';
+import { Client } from 'k6/x/net/http';
+
+export default async function () {
+  const socket = await dialTCP('192.168.1.1:80', {
+                            // default      | possible values
+    ipVersion: 0,           // 0            | 4 (IPv4), 6 (IPv6), 0 (both)
+    keepAlive: true,        // false        |
+    lookup: null,           // dns.lookup() |
+    proxy: 'myproxy:3030',  // ''           |
+  });
+  console.log(socket.active); // true
+
+  // Writing directly to the socket.
+  // Requires TextEncoder implementation, otherwise typed arrays can be used as well.
+  await socket.write(new TextEncoder().encode('GET / HTTP/1.1\r\n\r\n'));
+
+  // And reading...
+  socket.on('data', (data) => {
+    console.log(`received ${data}`);
+    socket.close();
+  });
+}
+```
+
+- UDP:
+```javascript
+import { dialUDP } from 'k6/x/net';
+import { Client } from 'k6/x/net/http';
+
+export default async function () {
+  const socket = new dialUDP('192.168.1.1:9090');
+
+  await socket.write(new TextEncoder().encode('GET / HTTP/1.1\r\n\r\n'));
+}
+```
+
+- IPC:
+```javascript
+import { open } from 'k6/x/file';
+import { Client } from 'k6/x/net/http';
+
+export default async function () {
+  const file = await open('/tmp/unix.sock');
+
+  // The HTTP client supports communicating over a Unix socket.
+  // Otherwise it can also be read from and written to directly.
+  const client = new Client({
+    socket: file,
+  });
+  await client.get('http://127.0.0.1/get');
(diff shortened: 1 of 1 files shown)


docs/design/018-new-http-api.md
```diff
@@ -0,0 +1,248 @@
...
+    console.log(jsonData);
+  }
+  ```
+
+- `advanced-post-streaming.js`:
+  ```javascript
+  import { File } from 'k6/x/file';
+  import { fetch } from 'k6/x/net/http';
+
+  // Will need supporting await in init context
```
codebien, line 226: Do we have an issue with it? It sounds like a prerequisite for the 3rd phase.

docs/design/018-new-http-api.md
```diff
@@ -0,0 +1,248 @@
...
+
+As the final phase, we should add deprecation warnings when `k6/http` is used, and point users to the new API.
+Eventually, months down the line, we can consider replacing `k6/http` altogether with the new module.
+
+
+### Examples
+
+- `basic-get.js`:
+  ```javascript
+  import { fetch } from 'k6/x/net/http';
```
codebien, line 163: If we implement fetch as a JS polyfill can we combine it in a native module? I guess we could have a wrapper on top of the new-k6-http-client+the polyfill. Am I missing something?

I think this is a nice UX and if we would not able to do it then we should insert it as a risk.

docs/design/018-new-http-api.md
imiric: This was my attempt at coming up with a File API, and is more of a thought experiment than anything else. #2977 goes into much more detail, and these kinds of issues should be addressed there.

This is inspired by [Deno's API](https://deno.land/manual/examples/fetch_data#files-and-streams). While they do have sync versions of most APIs, it's clear that if we want to adopt streams, we'll need to defer actual reading of files to whatever process needs it. I.e. we can't read the whole file into memory here in the init context, so it's likely that this will open a file handle only, and `file.readable` will be the stream reference. Whether that can be done without `await`, or if we'll need to support this in the init context, is an open question.

docs/design/018-new-http-api.md
imiric: My thinking so far is that the Fetch API would be part of the native Go module. There's really no reason it needs to be a JS library, and it would be a small addition anyway, so it shouldn't matter.

docs/design/018-new-http-api.md
mstoykov: I thought I commented this ... but :( 

(top level await](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await#top_level_await) is a ES2022 feature that made all the module resolutions in ESM asynchrnous. Which is partly what makes them a bit more involved and is what blocked them on async/await support in goja.

As soon as ESM is merged this will work as well. While I am in practice blocked specifically on fixing basically this functionality - I do need it for the basic implementation, so ... yeah - once ESM is native it will work.

2 workarounds are possible until then:
1. we can make the already existing wrapper async - this likely will have ... strange behaviours, but we already need to fix those around #2674 
2. The workaround for before this was a thing was to make an inline async lambda and call it. This has some usability problems, but will likely be okay for a release or two, given that I expect this functionality will not be released before ESM is on the final stretch.

docs/design/018-new-http-api.md
```diff
@@ -0,0 +1,399 @@
...
+    [Deno's implementation](https://deno.land/manual/examples/fetch_data) and [GitHub's polyfill](https://github.com/github/fetch) are good references to follow.
+
+    This was already suggested in [issue #2424](https://github.com/grafana/k6/issues/2424).
+
+  - The [Streams API](https://developer.mozilla.org/en-US/docs/Web/API/Streams_API), a [WHATWG standard](https://streams.spec.whatwg.org/) supported by most modern browsers.
+    [Deno's implementation](https://deno.land/manual@v1.30.3/examples/fetch_data#files-and-streams) is a good reference to follow.
+
+    There's a related, but very old [proposal](https://github.com/grafana/k6/issues/592) before the Streams API was standardized, so we shouldn't use it, but it's clear there's community interest in such an API.
+
+    Streaming files both from disk to RAM to the network, and from network to RAM and possibly disk, would also partly solve our [performance and memory issues with loading large files](https://github.com/grafana/k6/issues/2311).
```
mstoykov, line 40: We should definitely be linking to https://github.com/grafana/k6/issues/2978 given that it is basically this part

docs/design/018-new-http-api.md
```diff
@@ -0,0 +1,399 @@
...
+- `ipc`: a mechanism for communicating between processes on the same machine, typically using files.
+
+The Socket state can either be _active_—meaning connected for a TCP socket, bound for a UDP socket, or open for an IPC socket—, or _inactive_—meaning disconnected, unbound, or closed, respectively.
+
+##### Example
+
+- TCP:
+```javascript
+import { dialTCP } from 'k6/x/net';
+import { Client } from 'k6/x/net/http';
```
mstoykov, line 91:
```suggestion
```
I can't see this used in this example

docs/design/018-new-http-api.md
```diff
@@ -0,0 +1,399 @@
...
+export default async function () {
+  const socket = new dialUDP('192.168.1.1:9090');
+
+  await socket.write(new TextEncoder().encode('GET / HTTP/1.1\r\n\r\n'));
+}
+```
+
+- IPC:
+```javascript
+import { dialIPC } from 'k6/x/net;
```
mstoykov, line 131:
```suggestion
import { dialIPC } from 'k6/x/net';
```

docs/design/018-new-http-api.md
```diff
@@ -0,0 +1,399 @@
...
+}
+```
+
+- Passing a socket with custom transport settings, some HTTP options, and making a POST request:
+```javascript
+import { dialTCP } from 'k6/x/net';
+import { Client } from 'k6/x/net/http';
+
+export default async function () {
+  const socket = await dialTCP('10.0.0.10:80, { keepAlive: true });
```
mstoykov, line 171:
```suggestion
  const socket = await dialTCP('10.0.0.10:80', { keepAlive: true });
```

docs/design/018-new-http-api.md
```diff
@@ -0,0 +1,399 @@
...
+
+
+#### Fetch API
+
+The [Fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) is a convenience wrapper over existing Client, Socket and other low-level interfaces, with the benefit of being easy to use, and having sane defaults. It's a quick way to fire off some HTTP requests and get some responses, without worrying about advanced configuration.
+
+The implementation in k6 differs slightly from the web API, but we've tried to make it familiar to use wherever possible.
+
+Example:
+```
```
mstoykov, line 290:
```suggestion
```javascript
```

docs/design/018-new-http-api.md
```diff
@@ -0,0 +1,399 @@
...
+- Work should be started on some of the most impactful issues from the current API.
+  Issues like high memory usage when uploading files ([#2311](https://github.com/grafana/k6/issues/2311)), and data streaming ([#592](https://github.com/grafana/k6/issues/592)), are good candidates to focus on first.
+
+
+#### Phase 3: work on leftover issues
+
+**Goals**:
+
+- All leftover `new-http` issues should be worked on in this phase.
+  **TODO**: Specify which issues and in what order should be worked on here.
```
mstoykov, line 352: I guess this is the part with all the socket and dns stuff?

docs/design/018-new-http-api.md
```diff
@@ -0,0 +1,399 @@
...
+
+- Merge the extension into k6 core, and make it available to k6 Cloud users.
+
+- Continue to gather and address feedback from users, thorough testing and polishing.
+
+
+#### Phase 6: deprecate `k6/http`
+
+As the final phase, we should add deprecation warnings when `k6/http` is used, and point users to the new API.
+Eventually, months down the line, we can consider replacing `k6/http` altogether with the new module.
```
mstoykov, line 386: 1. I don't think we will ever replace `k6/http` the new module IIRC is supposed to be under `k6/net/http` the same way `k6/net/grpc` was.
2. Even if we ever remove `k6/http` (which is under consideration, given that this likely has way more implications then we imagine) - making `k6/http` import one thing in one version and a different in another seems like the worst possible UX. 
 
So :-1: on this particular line all together

mstoykov (commented):
Over all LGTM - especially as I expect anything we decide today will hit problems down the line and we will reconsider or redesign as we are crossing bridges.

I have a bunch of "typo" fixes which I hope won't be controversial. 

My main "negative" feedback is - there is a lot of promises for expanding the API beyond what is needed to be implemented with stuff such as native `fetch` API. While this might happen one day - I am against it being discussed in this proposal or being part of it's goals.

Structurally i feel like all the *Design* subsections are tireing to read and think about and then go to the *Implemetnation* and see that 2/3 of them are basically going to be in the later parts of the implementation or not mentioned at all. 

I am not certain that reversing the two will help. 

On that same *Design*, `k6/x/net` and `k6/x/net/dns` seem like really nice ideas ... but also stuff that expand the scope of this proposal quite a bit. And at the same time are not really well defined especially as in *when* they should be implemented.

docs/design/018-new-http-api.md
imiric: Done in d5ffcd280.

docs/design/018-new-http-api.md
imiric: Makes sense. Removed and clarified in 336abcd11.

docs/design/018-new-http-api.md
```diff
@@ -0,0 +1,399 @@
...
+
+- The API should be expanded to include all HTTP methods supported by the current API.
+  For the most part, it should reach feature parity with the current API.
+
+- A standalone `fetch()` function should be added that resembles the web Fetch API. There will be some differences in the options compared to the web API, as we want to make parts of the transport/client configurable.
+
+    Internally, this function will create a new client (or reuse a global one?), and will simply act as a convenience wrapper over the underlying `Client`/`Dialer`/`Transport` implementations, which will be initialized with sane default values.
+
+- Towards the end of this phase, the API should be mostly stable, based on community feedback.
+  Small changes will be inevitable, but there should be no discussion about the overall design.
```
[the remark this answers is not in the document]
imiric: Which part are you specifically referring to? All 3 points of the Goals section, or just the Fetch point?

I'm surprised you object to the Fetch API, since you opened #2424. :wink: 

I don't understand why this is such a hot topic...

First of all, this is not a blocker for making the API generally available. It will be available to anyone who wants to use the extension starting from phase 1. And we now agreed that the extension should be merged into k6 as an experimental module at the end of phase 2.

Secondly, once the main API is implemented, adding a `fetch()` wrapper on top of it is such a minuscule and trivial part of it, that's it's not even worth debating at this point.

The reason I think it's worth mentioning as a general design goal, even if it's not as important as other aspects of the API (which is why it's in phase 4) is because it would be very convenient to use. Since the overall goal of k6 is to deliver the best DX possible, having familiar and friendly wrappers like this aligns well with that mission.

So please suggest which parts of this you would take out, or how you would change it, but I don't think we should remove the Fetch API as a design goal.

docs/design/018-new-http-api.md
```diff
@@ -0,0 +1,402 @@
...
+The Socket state can either be _active_—meaning connected for a TCP socket, bound for a UDP socket, or open for an IPC socket—, or _inactive_—meaning disconnected, unbound, or closed, respectively.
+
+##### Example
+
+- TCP:
+```javascript
+import { dialTCP } from 'k6/x/net';
+
+export default async function () {
+  const socket = await dialTCP('192.168.1.1:80', {
```
imiric, line 100: I'm not a fan of these `dial*` functions. I added them inspired by Go's `net` package, but this doesn't provide the ability to implement custom dialers. So maybe we should expose some `Dialer` type? I wasn't sure how to structure it, though.

And I'm slightly against calling these "dialers" to begin with, since it's mostly a Go concept. At the same time, I didn't want to call this `connect()` either, since it wouldn't make sense for UDP or IPC. `dialIPC` still doesn't make sense to me, since it's essentially opening a file, but this too is following Go's `net` package. :thinking:

WDYT?

imiric:
@mstoykov:

> My main "negative" feedback is - there is a lot of promises for expanding the API beyond what is needed to be implemented with stuff such as native fetch API. While this might happen one day - I am against it being discussed in this proposal or being part of it's goals.

As I mentioned in the other comment, the Fetch API will be a very minor addition in later phases of development. I still think it's worth mentioning as a general design goal that people can look forward to. On its own, I don't think it deserves to have a separate design document, but it can live as a bullet point here.

If you're referring to the Sockets and DNS APIs, I agree that it expands the scope outside of what a purely HTTP API should be concerned about, but I think it's important to think about them early on, even if we don't expose them to JS. This will help us structure the implementation and Go API in such a way that makes it easy to eventually expose them to JS. I think this is the case for the Sockets API, at least, while the DNS API can be implemented later.

We can always extract them to standalone design documents, and reference them here, but if we don't think about them early on then we run the risk of creating an API similar to the current `k6/http`, which does many things internally and makes it difficult to later refactor and expose.

> Structurally i feel like all the Design subsections are tireing to read and think about and then go to the Implemetnation and see that 2/3 of them are basically going to be in the later parts of the implementation or not mentioned at all.

Is it "tiring to read" because of the amount of information and the content itself, or because you don't see them mentioned in the Implementation section? Please suggest what would make the content less tiring to read, as that's not desirable. The reason some of it is not mentioned in the Implementation section is because we haven't decided how/when/if to approach it, which is what the TODO in phase 3 is partly for. If we decide to work on the Sockets API early on, then we should mention it in phase 1 or 2, and DNS API in phase 3, etc. So let's make a decision.

Personally, I think we should start from the bottom up, so implement the Sockets API first, and then build the HTTP API on top of it. We might decide to not expose the Sockets API to JS yet, but as I said above, I think it's important if we think about it from the start.

> On that same Design, k6/x/net and k6/x/net/dns seem like really nice ideas ... but also stuff that expand the scope of this proposal quite a bit. And at the same time are not really well defined especially as in when they should be implemented.

Yup, so let's decide what to do with them before merging this. :) I can theoretically see the DNS lib be part of a separate proposal, but I'm not so sure about the Sockets API.

docs/design/018-new-http-api.md
```diff
@@ -0,0 +1,486 @@
...
+    json: { name: 'k6' },
+  });
+}
+```
+
+#### Events
+
+The new HTTP API will emit events which scripts can subscribe to, in order to implement advanced functionality.
+
+For example, a `requestToBeSent` event is emitted when the request was processed by k6, and just before it is sent to the server. This allows changing the request body or headers, or introducing artificial delays.
```
imiric, line 313: I think this might be a way to implement features like #2667 and #1716. Let me know what you think.

It could also serve as a starting point for a global event system, which could be useful for #2432.

docs/design/018-new-http-api.md
codebien: I agree, what about trying to be consistent with the HTTP Client API and doing something similar?

We have a class, a constructor, and methods.

```js
import { TCPSocket, <Other Protocols>... } from 'k6/x/net'; // TCP, TCPSocket, TCPConn (We should use a name that makes sense for the context and is aligned with the RFC

  const socket = new TCPSocket(''192.168.1.1:80'")
  await socket.open()
```

If it doesn't make sense for UDP then we can have `new UDP()` and that's all. This should also be very similar to the Linux API.

imiric: I had something like that initially, but decided against it, since you shouldn't be able to create a socket object, without it being actually connected/bound/open. Otherwise, you might be tempted to set some of its properties, which would be meaningless unless it's actually bound to an OS socket. In fact, most/all of its properties should be read-only. This is why I opted to have a constructor function instead.

An HTTP client is a bit different, since it's only used to make requests, and doesn't correspond to some system state. So you could theoretically have a Client instance that hasn't established a socket connection yet, and making the first request would establish the connection.

I also considered something like:

```javascript
import { TCP, UDP, IPC } from 'k6/x/net';
const socket = TCP.dial('192.168.1.1:80');
```

So in a sense you have some abstract protocol object, and calling `dial()` on it returns you a `TCPSocket`, `UDPSocket` or `IPCSocket` instance. I'm slightly leaning towards this now, even though `dial` is still in the name. WDYT?

imiric: I went with `open` in 8c97fe07e.

`connect` doesn't quite make sense for UDP or IPC, whereas `open` is generic enough to apply for all 3.

I think `dial` is not just Go's invention. It does have some networking roots, though I'm not able to find definitive references outside of Go, and "dial-up modem" :smile:, right now.

imiric: One more thing: the proposal doesn't currently specify a way to define custom dialers.

I wonder if we could handle this with an event system as well. E.g. have `before`/`after` events scripts can subscribe to, similar to the HTTP events, emitted before/after the socket is opened.

This could be a way to handle DNS lookups, instead of the currently proposed `lookup` hook function, but it would be generic enough for any other use cases (TCP proxies, custom metrics, etc.).

So you would have something like:

```javascript
const socket = await TCP.open('myhost', {
  eventHandlers: {
    before: event => {
      event.data.address = await dns.resolve(event.data.hostname, {
        rrtype: 'A',
        servers: ['1.1.1.1:53', '8.8.8.8:53'],
      });
    },
  },
});
```

I'm not entirely sold on the syntax to change `event.data` like this, as it's unclear which properties will be used after the event handler executes. Also, event handlers typically shouldn't return any values either, as that would make them too specific for a single purpose, so returning is not an option.

I'm not really sure we should allow creating custom dialer implementations from scratch in JS. I.e. nobody would choose to implement custom protocols from scratch in JS, and if they need to do that, then our Go API should allow it by being easily extensible. So for JS, it might just be good enough to have hooks, or event handlers, so scripts can implement some custom logic around the socket implementation.

docs/design/018-new-http-api.md
```diff
@@ -0,0 +1,486 @@
...
+import { Client } from 'k6/x/net/http';
+
+export default async function () {
+  const socket = await TCP.open('10.0.0.10:80', { keepAlive: true });
+  const client = new Client({
+    socket: socket,
+    proxy: 'https://myproxy',
+    version: 1.1,                     // force a specific HTTP version
+    headers: { 'User-Agent': 'k6' },  // set some global headers
+  });
```
na--, line 182: This and the tentative HTTP/3 example below don't look very realistic, or even very user-friendly :thinking:  There are a couple of missing links here, we generally don't go from an TCP/UDP connection directly to an HTTP client. 

We want _something_ (i.e. another component) to make network connections, not to have to make them raw. And we want _something_ (also another component or a configurable part of one) to be able to negotiate which HTTP version to be used, or to be able to force that a specific HTTP version is used. Controlling these things by requiring users to make raw TCP/UDP connections doesn't seem like the way to go, especially given how complex HTTP protocol version negotiation and how it's intermingled with DNS and TLS...

docs/design/018-new-http-api.md
```diff
@@ -0,0 +1,486 @@
...
+##### Examples
+
+- Using a client with default transport settings, and making a GET request:
+```javascript
+import { Client } from 'k6/x/net/http';
+
+export default async function () {
+  const client = new Client();
+  const response = await client.get('https://httpbin.test.k6.io/get');
+  const jsonData = await response.json();
```
na--, line 165:
```suggestion
  const jsonData = await response.json();
```

This part of the current HTTP API is probably not a good design to copy. Why should a generic HTTP response have a `.json()` method? 

It's fine to have something that wraps a regular `Client` and works only with JSON, but a generic `http.Client` should probably not need to know what JSON is.

docs/design/018-new-http-api.md
```diff
@@ -0,0 +1,486 @@
...
+  const socket = await TCP.open('10.0.0.10:80', { keepAlive: true });
+  const client = new Client({
+    socket: socket,
+    proxy: 'https://myproxy',
+    version: 1.1,                     // force a specific HTTP version
+    headers: { 'User-Agent': 'k6' },  // set some global headers
+  });
+  await client.post('http://10.0.0.10/post', {
+    json: { name: 'k6' }, // automatically adds 'Content-Type: application/json' header
+  });
```
na--, line 185: Same logic as https://github.com/grafana/k6/pull/2971/files#r1161730930 - we shouldn't add any special handling of `json` options in the default Client. This is not composable, it's the opposite of that - trying to make a `Client` that satisfies all use cases. And it adds "automagic behavior", as your `// automatically adds 'Content-Type: application/json' header` comment directly shows.

The request bodies that the normal HTTP Client accepts should probably be limited to `string`, `ArrayBuffer` or some sort of a Stream. Anything else needs to be an error. We can have a separate `RESTClient` or `JSONClient` that has additional automated handling of JSON requests and responses.

docs/design/018-new-http-api.md
```diff
@@ -0,0 +1,486 @@
...
+    headers: { 'User-Agent': 'k6' },  // set some global headers
+  });
+  const request = new Request('https://httpbin.test.k6.io/get', {
+    // These will be merged with the Client options.
+    headers: { 'Case-Sensitive-Header': 'somevalue' },
+  });
+  const response = await client.get(request, {
+    // These will override any options for this specific submission.
+    headers: { 'Case-Sensitive-Header': 'anothervalue' },
+  });
```
na--, line 260: This API proposal seems a bit messy. The HTTP request method should probably be part of the `Request` object, right? The whole point of a `Request` object is to fully contain everything you need to make a request.

But then, what `client.get()` probably shouldn't accept `Request` parameters, since that `Request` could be a `POST`. I am not sure if methods like `.get()` and `.post()` are even necessary in the default HTTP `Client` API :thinking: Maybe a single generic `Client.request()` or `Client.do()` is enough. But if we choose to have helper methods, they probably shouldn't accept a `Request`, just a `method, params, body` or something like that :thinking:

na-- (commented):
:+1: for most of the text here, e.g. I agree with the "Why is this needed?" section, the rough "Implementation" plan, and the text at the start of the "Proposed solution(s) -> Design" section. 

However, looking over the JS API examples, I see a lot of potential issues. I noted some of the bigger ones as inline comments. We probably shouldn't (and realistically can't) figure out all of these details in this design document, those would probably be better off to leave for when we are iterating on the implementation, so we don't necessarily even need to fully fix them here. But at least we should agree on some rough guidelines (e.g. when it comes to automatic JSON handling or how to negotiate/fix HTTP versions).

docs/design/018-new-http-api.md
```diff
@@ -0,0 +1,486 @@
...
+
+  - The `Client` interface with only one method: `request()`, which will work similarly to the current `http.asyncRequest()`.
+
+    For the initial PoC, it's fine if only `GET` and `POST` methods are supported.
+
+    It's not required to make `Dialer` and `Transport` fully configurable at this point, but they should use sane defaults, and it should be clear how the configuration will be done.
+
+- This initial API should solve a minor, but concrete, issue of current API. It should fix something that's currently not possible and doesn't have a good workaround.
+
+  Good candidates: [#936](https://github.com/grafana/k6/issues/936), [#970](https://github.com/grafana/k6/issues/970).
```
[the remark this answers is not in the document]
na--: From what I know, HTTP version negotiation happens at a lower level, so I am not sure if the `Client` is the place where we should have a `version` property :thinking: On the network or transport level is probably more appropriate :thinking: Though I am not sure if we should even specify these things in this design doc, a PoC seems like a better place to figure them out

docs/design/018-new-http-api.md
imiric: `http.Client` wouldn't know what JSON is, but I don't see why `http.Response` couldn't have helper methods to parse the response body and return it as various objects.

The [web Fetch API has a `Response` with such methods](https://developer.mozilla.org/en-US/docs/Web/API/Response#instance_methods), and [so does Deno's implementation](https://deno.com/deploy/docs/runtime-response#methods). Regardless if we end up implementing Fetch or not, this is very convenient, and it would be a UX regression if we decide to remove them.

docs/design/018-new-http-api.md
imiric: True, conceptually there's a missing "transport" abstraction between raw sockets and the high-level HTTP client. But this is an example of how both sides can be integrated when the user needs to configure socket/transport-level options. The friendlier example is the one above this, where you simply instantiate a `Client`, and make requests with it, using the default transport settings.

> we generally don't go from an TCP/UDP connection directly to an HTTP client

You can certainly talk HTTP/1 over a raw TCP socket. HTTP/2+ and TLS complicate this substantially, but ideally we would like to keep a friendly API for all versions, without introducing too many concepts.

The reason I'm hesitant to expose yet another abstraction is because the API would be even more confusing. A "transport" is uniquely a Go concept; no other JS runtime that I could find exposes it in the way Go does. For example, Deno has [`Deno.connect()`](https://deno.land/api?s=Deno.connect) and [`Deno.connectTls()`](https://deno.land/api?s=Deno.connectTls). Node has entirely separate [`net`](https://nodejs.org/api/net.html) and [`tls`](https://nodejs.org/api/tls.html) modules. That's not to say that we can't do something entirely different, but it would go against our intention of building something familiar to JS users.

That said, the following makes sense to me:

- A raw Sockets API to read/write TCP/UDP/IPC directly. Since host name resolution must happen before a connection is established, or higher level protocols are negotiated, a `lookup` hook as shown in the current examples is a good way to configure the DNS logic. This is inspired by [Node](https://nodejs.org/api/net.html#socketconnectoptions-connectlistener).

- An optional Transport API that wraps the raw socket, and implements whatever protocol or security on top. These implementations must implement the same Socket interface, so that the `http.Client` can be unaware whether it's using a raw socket connection, or e.g. a TLS connection. This would mimic Node, where [`TLSSocket` extends the base `Socket`](https://nodejs.org/api/tls.html#class-tlstlssocket), so that it can be used transparently.

  So, for example, TLS options could be specified like this:

  ```javascript
  import { TLS } from 'k6/x/net';
  const tlsSocket = await TLS.connect('10.0.0.10:443',
    { version: { min: 'ssl3.0', max: 'tls1.3'} });
  const client = new Client({
    socket: tlsSocket,
  });
  ```

- The high-level Client API would do only HTTP-specific negotiation. So if HTTP version 2 is requested, but a non-TLS connection is passed to it, it would throw an error. For that to work, an `insecure: true` option must be passed, which would negotiate `h2c`.

Let me know if you disagree with anything above.

docs/design/018-new-http-api.md
na--: Fetch and Deno might have this, but [Go](https://pkg.go.dev/net/http#Response), [axios](https://github.com/axios/axios#response-schema) and [got](https://github.com/sindresorhus/got/blob/main/documentation/3-streams.md#response-2) don't. We should consider what makes sense for us. 

These are my thoughts on the topic, in roughly a descending order of certainty:
1. We definitely shouldn't have a [`.json([selector])` method like the current one](https://k6.io/docs/javascript-api/k6-http/response/response-json/), where we have a `selector` as the argument. That should be a separate JSONPath API that is not tied directly together to the core HTTP API, it should probably work with any string or buffer.
2.  It's fine to have a `Fetch` polyfill/wrapper that has a `.json()` method, built on top of our generic HTTP `Client` (either in Go or in JS!). This solves some of the UX problem of a generic API not having one.
3. It's probably a good idea to have a dedicated custom `JSONClient` type built on top of the generic `Client`, which will be well suited for easily working with JSON REST APIs. Including potentially automatically marshaling request bodies that are objects and maybe even automatically universalizing responses. Maybe even with some JSONPath integration :thinking: This provides all of the UX benefits, but without anything magical and with a clear separation of concerns (i.e. the composable approach).
4. The default generic HTTP `Client` should not make any guesses about the content of the HTTP requests and responses it is handling. This gives us more room for optimization, a clear separation of concerns, and a more consistent UX.
5. We probably shouldn't have a `.json()` method (without any arguments) on the generic HTTP `Client`. If someone _really_ wants to use the generic k6 http `Client` for one-off request with a JSON response, then `JSON.parse(resp.body)` is not that much worse than `resp.json()`...
6. If we decide to add `.json() `, are we also going to add to port the current [`.html()`](https://k6.io/docs/javascript-api/k6-http/response/response-html/)? If not, then why not? And are also going to have `.xml()` or `.protobuf()`?

na--: Fetch and Deno might have this, but [Go](https://pkg.go.dev/net/http#Response), [axios](https://github.com/axios/axios#response-schema) and [got](https://github.com/sindresorhus/got/blob/main/documentation/3-streams.md#response-2) don't. We should consider what makes sense for us. 

These are my thoughts on the topic, in roughly a descending order of certainty:
1. We definitely shouldn't have a [`.json([selector])` method like the current one](https://k6.io/docs/javascript-api/k6-http/response/response-json/), where we have a `selector` as the argument. That should be a separate JSONPath API that is not tied directly together to the core HTTP API, it should probably work with any string or buffer.
2.  It's fine to have a `Fetch` polyfill/wrapper that has a `.json()` method, built on top of our generic HTTP `Client` (either in Go or in JS!). This solves some of the UX problem of a generic API not having one.
3. It's probably a good idea to have a dedicated custom `JSONClient` type built on top of the generic `Client`, which will be well suited for easily working with JSON REST APIs. Including potentially automatically marshaling request bodies that are objects and maybe even automatically universalizing responses. Maybe even with some JSONPath integration :thinking: This provides all of the UX benefits, but without anything magical and with a clear separation of concerns (i.e. the composable approach).
4. The default generic HTTP `Client` should not make any guesses about the content of the HTTP requests and responses it is handling. This gives us more room for optimization, a clear separation of concerns, and a more consistent UX.
5. We probably shouldn't have a `.json()` method (without any arguments) on the generic HTTP `Client`. If someone _really_ wants to use the generic k6 http `Client` for one-off request with a JSON response, then `JSON.parse(resp.body)` is not that much worse than `resp.json()`...
6. If we decide to add `.json() `, are we also going to add to port the current [`.html()`](https://k6.io/docs/javascript-api/k6-http/response/response-html/)? If not, then why not? And are also going to have `.xml()` or `.protobuf()`?

docs/design/018-new-http-api.md
na--: I fundamentally disagree with a lot of it, mostly because I don't think it will work at all, and even the parts that might work won't work well. Even if they worked, raw sockets are probably going to be the wrong abstraction 99% of the time, for anything the user may want to do with custom HTTP settings. I don't dismiss we should have support for them (though I am not sure if we should start with that), I just think there should be a layer or two between them and the HTTP `Client`.

This is why I don't think they will even work for most things. Consider even the example you provided:
```js
const tlsSocket = await TLS.connect('10.0.0.10:443',
  { version: { min: 'ssl3.0', max: 'tls1.3'} });
const client = new Client({
  socket: tlsSocket,
});
```

I see at least 4 separate issues with this small piece of code:
1. Most of the time, you need DNS to establish TLS, right? Because the certificates are for the domain names. 
2. You need to negotiate the HTTP/2 protocol version during the TLS connection establishment, so we can't easily split the TLS connection and the HTTP requests... See https://en.wikipedia.org/wiki/Application-Layer_Protocol_Negotiation
3. What happens to connection reuse? What happens if we don't want that and want a new connection for every request. It's perfectly reasonable for a single HTTP client to make multiple (potentially simultaneous) network requests to the same or to different hosts.
4. What happens if we have a redirect and we want the `Client` to follow it? :sweat_smile: 

In short, an HTTP `Client` is something that needs to be able to make its own network requests through some mechanism. You can't give it a single Socket and expect it could work just on that

docs/design/018-new-http-api.md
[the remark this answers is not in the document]
imiric: > HTTP version negotiation happens at a lower level

Right, it can happen [during TLS negotiation, as part of ALPN](https://matthewparrilla.com/post/negotiation-http2-alpn-tls-handshake/). But [HTTP/1.1 connections can also be upgraded via a header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Protocol_upgrade_mechanism). I'm conflicted about it as well, though it definitely should be somewhere lower level as well.

Maybe part of `TLS.connect()`?

```javascript
import { TLS } from 'k6/x/net';
const tlsSocket = await TLS.connect('10.0.0.10:443',
  { alpn: [ 'h2', 'http/1.1' ] });
const client = new Client({
  socket: tlsSocket,
});
```

This way you could force HTTP/1.1 over TLS, even if the server supports HTTP/2. This could be simplified to a `versions` array, instead of `alpn`.

Since `h2c` must be negotiated via an `Upgrade` header, and can only be done without TLS, then something like the `insecure` flag above on the `Client` itself would be the way the go, in which case we should also keep the `version` property. It wouldn't make sense to specify that as a, say, `TCP.open()` option.

But I agree that we don't need to agree on every single detail to start working on the PoC. We can iron out these details as we make progress.

docs/design/018-new-http-api.md
```diff
@@ -306,6 +306,90 @@ export default async function () {
...
+import { Client } from 'k6/x/net/http';
+
+export default async function () {
+  const client = new Client();
+  client.on('requestToBeSent', event => {
+    console.log(event.type);    // 'requestToBeSent'
+    const request = event.data;
+    request.headers['Cookie'] = 'somecookie=somevalue;'  // overwrites all previously set cookies
+    request.body += ' world!';  // the final body will be 'Hello world!'
+    sleep(Math.random() * 5);   // random delay up to 5s
```
mstoykov, line 326: nit: let's not use `sleep` in examples. You can use `await delay` and import [this gist](https://gist.githubusercontent.com/mstoykov/7684bf54a1e2f35d048868ba6117e2cb/raw/d44e15a1a942c8aafb9e7a4f7c0d623214b3a651/delay.js) for example. 

This way this examples won't be showing how to block the whole event loop.

docs/design/018-new-http-api.md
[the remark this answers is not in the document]
mstoykov: I would argue the proposal here is about making things *possible*. 

Both we and every user can (and will) extend this API. 

But for me adding a bunch of UX improvements and that being not a small part of the specification does not help with the discussion around - it hinders it. I now or whoever has to read a bunch more text and then decide that whoever wrote it and the people who have agreed, meant those as inspirational things instead of as goals that need to be reached by this proposal to be called fulfilled.

docs/design/018-new-http-api.md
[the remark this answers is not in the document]
mstoykov: Just to note that the `UPGRADE` header is only used for (in practice) websockets and `h2c` without "prior knowledge". 

The usuall http2 upgrade happens in practice in the tls handshake and h2c with prior knowledge just starts talking http2 directly.

http3 being over UDP means that the upgrade is basically a completely new connection. Except again if "prior knowlege" is used in which case it is still on a new connection, it just skips the first one ;). 

I am not certain where this should be configured ... or even if it should be one place or if it should be solved with some more involved setup.

mstoykov:
> As I mentioned in the other comment, the Fetch API will be a very minor addition in later phases of development. I still think it's worth mentioning as a general design goal that people can look forward to. On its own, I don't think it deserves to have a separate design document, but it can live as a bullet point here.

That is great, but if it is in this proposal we need to discuss it and agree that we want it as part of the proposal. And I am not certain I would like to have the `fetch` API implemented. This also goes for all the other things that you argue are "general deisgn goals that people can look forward to". If its in the proposal, we need to agree on it with the proposal. If its in an issue that references the proposal as a requirement - it doesn't need to be agreed with the proposal.

I am not saying that you should open an issues or making new proposals. But the idea of the proposal is that we agree on what we want to do and how we are going to try to do it. And I (and maybe others) aren't certain that fetch or per HTTP verb methods are a good thing. I certainly don't think they should be part of this proposal. 

> If you're referring to the Sockets and DNS APIs, I agree that it expands the scope outside of what a purely HTTP API should be concerned about, but I think it's important to think about them early on, even if we don't expose them to JS. This will help us structure the implementation and Go API in such a way that makes it easy to eventually expose them to JS. I think this is the case for the Sockets API, at least, while the DNS API can be implemented later.


I am not against discussing them. The oppossite, but I would like to have some clarity on what will be implemetned as part of this proposal and what not. As it seems that to be part of the proposal idea - stating what we will do.

> We can always extract them to standalone design documents, and reference them here, but if we don't think about them early on then we run the risk of creating an API similar to the current `k6/http`, which does many things internally and makes it difficult to later refactor and expose.

Again I do agree we should think about them :+1: 

> Is it "tiring to read" because of the amount of information and the content itself, or because you don't see them mentioned in the Implementation section? Please suggest what would make the content less tiring to read, as that's not desirable. The reason some of it is not mentioned in the Implementation section is because we haven't decided how/when/if to approach it, which is what the TODO in phase 3 is partly for. If we decide to work on the Sockets API early on, then we should mention it in phase 1 or 2, and DNS API in phase 3, etc. So let's make a decision.

The tiring part is that I don't know why I am reading it and where it goes. And then when I get to the end, I am "why did I read this" as I never got to see when they will be implemented.

So I guess I am proposing that part of the proposals are more accurately described as ... ideas (?) on how to implement the API on top. 

> Personally, I think we should start from the bottom up, so implement the Sockets API first, and then build the HTTP API on top of it. We might decide to not expose the Sockets API to JS yet, but as I said above, I think it's important if we think about it from the start.

I am okay with that but arguably that should be stated in the proposal. When I last read it in full it very much sounded like you intend on implementing `Client`. And then (at some point) expose the underlying mechanisms (Dialer, Socket, Transport ... or w/e else) to JS for more control over different parts.

docs/design/018-new-http-api.md
na--: The problem is that it makes the API have weird edge cases. For example, what happens if I provide both `body` and `json` and `formData`? :sweat_smile: Now the library needs to have some extra internal logic for the order of precedence, and we have do document and test that, etc. Having more than one way to do the same thing in the same object is generally not great API design.

na--: Actually, from looking at the request library API you linked to, its `json` property is a boolean one, you still have to supply to `body`, right? Which may be even worse than this suggestion, since you now have extra parameters that just control the behavior of how the `body` is automagically processed.

docs/design/018-new-http-api.md
na--: Let me focus just on one thing then:

> This passing of an already established connection is done to configure the initial connection, which the client will use instead of creating a new one.

To me this seems like a fundamentally flawed approach which can't work. It also unfortunately seems to be somewhat at the core of your proposal for handling some of the lower-level connection details. 

I already explained some of the reasons why I don't think it will work. I also don't remember **ever** seeing an approach like that, in any of the languages or libraries I've used to make HTTP requests. I don't see such an approach in any of the links you provided either:
- [node's `http.request()` has a `createConnection()` callback](https://nodejs.org/api/http.html#httprequestoptions-callback)
- [deno seems to use `Fetch`, but you can create a custom HTTP client](https://deno.land/api@v1.32.3?s=Deno.CreateHttpClientOptions&unstable=)
- [axios seems to allow you to define the `transport` and `proxy` in every request](https://github.com/axios/axios#request-config)

I may be missing something, so please share at least *one* example from *any* language or HTTP library where something like this is done, to convince me that it's a viable path forward.

It's good to have low-level network connection APIs. It's fine to somehow even allow their usage from the HTTP `Client` (e.g. by specifying a callback function that can make and return custom connections). But an HTTP client doesn't need a single network connection, it needs a mechanism to make connections, a connection factory :sweat_smile:

docs/design/018-new-http-api.md
imiric: > Fetch and Deno might have this, but Go, axios and got don't.

I wasn't familiar with got, looks nice :thumbsup: It doesn't implement `json()` directly on `Response`, but [does on the `Promise`](https://github.com/sindresorhus/got/blob/main/documentation/1-promise.md#goturl-string--url-options-optionsinit-defaults-options). It even supports [the `json` property for `POST` requests](https://github.com/sindresorhus/got#json-mode).

BTW, there's another HTTP lib from the same team called [ky](https://github.com/sindresorhus/ky). :sweat_smile: This one apparently targets Deno, but it follows similar design principles as got.

[Axios actually parses JSON bodies by default](https://masteringjs.io/tutorials/axios/json) :smile: But I don't think we should look at it for inspiration, considering it's quite old at this point, and has been superseeded by other libs. It's still very popular and featureful, but also has [many limitations](https://github.com/sindresorhus/got#comparison).

Go shouldn't be an inspiration for the JS API. As stated in the design goals, we want to make things familiar to JS devs, not Go devs. We can borrow certain concepts to make the API composable and whatnot, but refering to it directly would only pollute our way of thinking about idiomatic JS.

> We should consider what makes sense for us.

If by "us" you mean "k6 users", then agreed. These convenience methods are all part of offering a good experience for JS developers. I don't see why that's so controversial.


To address some of your points:

- Sure, the Selector API makes sense as a separate JSONPath API. `json()` should just return the body as an object, as it does in the other JS libs.

- Having purpose-built `Client` implementations just for sending different headers, interpreting requests and responses, seems like overkill, and might be too limiting in some cases. What if you want to send JSON but receive binary data, or any such weird combination we didn't predict? Having a single `Client` that offers some helper mechanisms for the most common use cases adds a fairly low overhead, which can be ignored for those who don't need it. I.e. everyone can just ignore the `json` property, and use `body` instead.

- I'm not sure why you keep mentioning this, but `Client` itself wouldn't have a `json()` method. It would return a `Promise<Response>` which would have it.

- I don't see why we couldn't have `html()` either. The point of these methods are _convenience_ for the most common use cases. If there are good reasons to include `xml()` or `protobuf()`, then we should consider them as well.

  Are you suggesting that we have an `HTMLClient`, a `ProtobufClient` or an `XMLClient` instead?

  Separate client implementations make sense for truly different protocols. So it probably makes sense to have a `GRPCClient` that extends the base `Client`, or maybe a `SOAPClient` :smile:, but not just to avoid these simple payload handling helpers.

na--: > Go shouldn't be an inspiration for the JS API. As stated in the design goals, we want to make things familiar to JS devs, not Go devs. We can borrow certain concepts to make the API composable and whatnot, but refering to it directly would only pollute our way of thinking about idiomatic JS.

I mostly agree, but I will push back slightly on this. Go **can** and probably should be an inspiration, as long as the end result is also an idiomatic JavaScript API. If Go has an elegant solution to a problem, and if that solution doesn'tt look strange for JS APIs, is easy to understand, and cleanly solves a bunch of problems, why not adopt it? :confused: Sure, we don't need to copy the Go API directly, but we don't have to avoid its good ideas just be cause they are not JS. Good API design is _somewhat_ universal after all...

> Having purpose-built Client implementations just for sending different headers, interpreting requests and responses, seems like overkill, and might be too limiting in some cases. 

I didn't meed completely separate `Client` implementations. Just wrappers around a generic `Client` with maybe a few extra methods and some pre-defined Event handlers for the same events that you mention in the design doc.

> What if you want to send JSON but receive binary data, or any such weird combination we didn't predict? 

Use the generic `Client`, if it's a one-off, or build your own wrapper (because you should be able to easily do that, since everything is composable).

>  Having a single Client that offers some helper mechanisms for the most common use cases adds a fairly low overhead, which can be ignored for those who don't need it. I.e. everyone can just ignore the json property, and use body instead.

Sure, but having more than one way to do the same thing, especially when it's a simple thing, is generally somewhat of an API design smell. I don't object too strongly to the simple `.json()` method, it's only at point 5 in my list, but I don't see the need. It's also one of these extra things that we can easily add at any future point, but once we add it we can never remove, so I don't see the need to start with it.

> I'm not sure why you keep mentioning this, but Client itself wouldn't have a json() method. It would return a Promise<Response> which would have it.

Sorry, I meant the Response object returned by the generic Client (or, I guess, the Promise that resolves to it).

> I don't see why we couldn't have html() either. The point of these methods are convenience for the most common use cases. If there are good reasons to include xml() or protobuf(), then we should consider them as well.

See the problems that `.html()` causes for the current `k6/http` API... Tightly coupling these is a mistake. 

> Are you suggesting that we have an HTMLClient, a ProtobufClient or an XMLClient instead?

No, that was the point. At best, we may have some helpers and wrappers, but we probably shouldn't. We should make it super easy and efficient for users to build whatever helpers and wrappers. We should provide the flexible and generic toolbox, not try to build a dedicated tool to solve every possible weird use case that exists out there.

na--: Hmm actually, I am slightly walking back some of my claims... :sweat_smile:  Thinking about idiomatic `Promise`-based JS APIs, a `.json()` helper on a `Promise` that also returns another `Promise` actually makes some sense :sweat_smile: I still don't like it, but it _probably_ provides enough value to pass muster :thinking:

docs/design/018-new-http-api.md
imiric: I'm not claiming that what's currently here is a viable path forward. I'm expecting it will need many changes, and this is why we're discussing it. I'd like this document to be a team collaboration, rather than a traditional PR review. If you notice, the proposal author is "The k6 core team". :slightly_smiling_face:

I'm also not disagreeing with your feedback, for the most part. Some of the points you brought up are also unclear to me. So I can try to come up with a design that addresses those issues as well, and then wait more time for other concerns to pop up, or you can just suggest concrete improvements here, or—even better—push changes directly to this branch. This is not "my proposal".

This Socket <-> HTTP integration is probably my idea, so I don't have an example from similar APIs. What I was trying to avoid is the complexity of Node (notice that they also have the concept of an Agent that manages connections), and Deno doesn't really have the flexibility we need, since [its `HttpClient` is very limited](https://deno.land/api?s=Deno.HttpClient&unstable=), and is mostly a bag of options that are passed to `fetch()`. I haven't looked at Axios yet, but from a glance, that `transport` option is very unclear. :confused:

I think we can either handle this via some kind of factory hook, similar to Node's `createConnection()`, or via an event that can be emitted before creating the connection. I'm leaning towards events, as those can be applied for any advanced scenario, and are idiomatic to JS.

What I wouldn't want is to have a bunch of new abstractions, where we have to explain concepts like transport, dialer, agent, or whatever else.

imiric:
Reproducing [this comment](https://github.com/grafana/k6/pull/2971#discussion_r1163782771) here for visibility:

Would it help if we split this large proposal into several smaller ones? The scope is quite large already, and we haven't even fleshed out the details of the Sockets or DNS APIs. Splitting this into separate proposals would make each one more manageable, and allow us to prioritize them as a related group. Then the UX improvements, the Fetch API and any such convenience wrappers, could go into one.

I'm not sure where to start with this, so if we agree to do this, any suggestions to move it forward would be appreciated.

docs/design/018-new-http-api.md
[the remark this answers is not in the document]
na--: At this point, I think we would be better served by proofs of concept. Iterating on code instead of on more lengthy text seems like it would be the more productive way forward.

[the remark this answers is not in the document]
codebien: >Would it help if we split this large proposal into several smaller ones?

@imiric If possible, I would prefer to have one single source of truth and not fragment the information. In the end, the doc sounds still manageable to me.

I think the unique very detailed phase should be the first. All the rest should give us a guideline in terms of vision and roadmap for the long-term. We should re-iterate this doc at the end of each phase and before starting a new one.

I suggest changing phase 4 with something like this:

```suggestion
#### Phase 4: expand, polish and stabilize the API

**Goals**:

- The API should be expanded to include all features supported by the current API.
  For the most part, it should reach feature parity with the current API.
  
- Include all the major required features: [we can drop here the list of all the issues not included in the previous phases - I expect we will update it after the ].

- Towards the end of this phase, the API should be mostly stable, based on community feedback.
  Small changes will be inevitable, but there should be no discussion about the overall design.
```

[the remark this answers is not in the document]
imiric: > At this point, I think we would be better served by proofs of concept. Iterating on code instead of on more lengthy text seems like it would be the more productive way forward.

Agreed. I started working on a PoC weeks ago, but Ned then suggested it was better to discuss the design first, so I abandoned it. At this point, it doesn't even feel like we agree on phase 1 of the current proposal, so I don't think we're anymore ready to work on the PoC than we were back then.

The reason to split this proposal into several more focused ones is to address [Mihail's feedback](https://github.com/grafana/k6/pull/2971#pullrequestreview-1360816646) that the scope of this proposal has expanded beyond just an HTTP API. And particularly to get rid of any mentions of UX improvements and the Fetch API, which seems to be controversial.

If we don't want to split it, then I guess everyone is fine with most of the sections here being incomplete, and the proposal "tiring to read"? It's frustrating trying to address some feedback, and then getting mixed signals about the way to proceed. :sweat:

In order to work on the PoC again, **does everyone agree with the current phase 1**?

That is: we won't be exposing a Sockets API, and network/transport options won't be configurable. The only goal is to expose a `Client.request()` that fixes a minor issue like #936 or #970.

If you agree, please approve the PR and let's merge it as is. If not, please suggest improvements to **phase 1 only**. We can iterate on and flesh out the other phases later.

docs/design/018-new-http-api.md
```diff
@@ -0,0 +1,484 @@
...
+- Implement a barebones async API that serves as a proof-of-concept for what the final developer experience will look and feel like.
+  The code should be in a state that allows it to be easily extended.
+
+  By barebones, we mean:
+
+  - The `Client` interface with only one method: `request()`, which will work similarly to the current `http.asyncRequest()`.
+
+    For the initial PoC, it's fine if only `GET` and `POST` methods are supported.
+
+    It's not required to make `Dialer` and `Transport` fully configurable at this point, but they should use sane defaults, and it should be clear how the configuration will be done.
```
codebien, line 411: quote from https://github.com/grafana/k6/pull/2971#discussion_r1165513267
>and network/transport options won't be configurable

It sounds like a conflict to me. Whether they should be configurable or not will depend on the concrete API required to resolve #936 and #970. If you are confident they will not be required, can you amend this point, please? And maybe add them to non-goals.

docs/design/018-new-http-api.md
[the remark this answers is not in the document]
codebien:
```suggestion
This initial API must solve a minor, but concrete, issues of the current API. It fixes something that's currently not possible and doesn't have a good workaround as [#936](https://github.com/grafana/k6/issues/936) and [#970](https://github.com/grafana/k6/issues/970).
```
I think we can promote them.

docs/design/018-new-http-api.md
```diff
@@ -0,0 +1,484 @@
...
+  Good candidates: [#936](https://github.com/grafana/k6/issues/936), [#970](https://github.com/grafana/k6/issues/970).
+
+- Features like configuring options globally, or per VU or request, should be implemented.
+  Deprecating the `httpx` library should be possible after this phase.
+
+
+**Non-goals**:
+
+- We won't yet try to solve performance/memory issues of the current API, or implement major new features like data streaming.
+
```
codebien, line 424:
```suggestion
> The following phases should be intended as a broad vision and roadmap of the API. It is expected to re-work and better define them phase by phase.
```
In this way, future readers will know that they shouldn't interpret this as a finalized doc. Move it to another section if you think there are better places than this one.

docs/design/018-new-http-api.md
imiric: Agreed, both #936 and #970 will probably require configuration at the transport level.

Since we're not clear how this will be done yet, WDYT if we work on #761 instead? `httpx` is a workaround for that, but just setting reusable options at the client level, which can be overridden per request, would be a less involved first step. Then we can always expand phase 1 to include #936 and #970, once we agree on how the transport should be configured (either in this or a follow-up PR).

docs/design/018-new-http-api.md
imiric: I added a note to the beginning of the Implementation section. See ba3383ed1.

docs/design/018-new-http-api.md
imiric: I made #761 the highest priority of phase 1, and #936 and #970 stretch goals. See 81034aab4.

docs/design/018-new-http-api.md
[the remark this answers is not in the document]
imiric: @codebien See 81034aab4.

codebien (dismissed):
I'm fine with the current version and start iterating on the first phase. I think after it we will have a lot of acquired knowledge to report back and to have a more accurate and final version of this document.

docs/design/018-new-http-api.md
imiric: I added a `dial` hook to the `Client` constructor in 0fde82468, along with TLS options to `TCP.open()`.

I think this addresses what we discussed above, otherwise please suggest alternatives.

docs/design/018-new-http-api.md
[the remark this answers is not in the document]
imiric: In 0fde82468, I added TLS options to `TCP.open()`, instead of having `TLS` be a separate `k6/x/net` class.

docs/design/018-new-http-api.md
```diff
@@ -186,22 +188,44 @@ export default async function () {
...
+```javascript
+import { TCP } from 'k6/x/net';
+import { Client } from 'k6/x/net/http';
+
+export default async function () {
+  const client = new Client({
+    dial: async address => {
+      return await TCP.open(address, { tls: false });
+    },
+    version: [2],
```
codebien, line 224: Should we have different implementations?

```js
import { Client } from 'k6/x/net/http';  // this is v1
import { Clientv1, Clientv2, Clientv3 } from 'k6/x/net/http';
```
Ofc, they can share the common parts under the hood but still give us flexibility in handling different logic.

mstoykov (approved):

mstoykov:
I approve this with the idea that while we are very likely to not do most of the stuff in this proposal as is.

But I also do expect that somewhere along the way it will turn out a bunch of the stuff proposed will not be enough.

I also have to again point out that at least some of the (later) functionality is based on functionality we currently don't have good support such as streams and events (which arguably can be implemetned on case by case bases). 

So I do expect this will hit a bunch of issues a long the way and likely will need to be heavily refactored/rewritten as things arise. And that is okay with me.

codebien (approved):

imiric:
> I approve this with the idea that while we are very likely to not do most of the stuff in this proposal as is.
>
> But I also do expect that somewhere along the way it will turn out a bunch of the stuff proposed will not be enough.

I also expect many changes to this document as we make progress, but I'm not sure how you can be so certain that we're very likely to not do most of this as is. The design here is abstract at this point, but most of it is grounded in APIs from other JS runtimes. If you have strong objections to the practicality of any of it, then please propose alternatives we can discuss (in a new PR :smile:), but it's not helpful to be generally dismissive about the work we've done so far.

Thanks for reviewing! :bow:

PR #2981, opened 2023-03-16
[gRPC] Set authority header as host

ryanbrigden:
* Description *

This PR contains a small fix to set the `authority` request header to the `host` rather than `host:port` for unary gRPC requests in k6. I believe this is a _strict_ improvement as an `authority` value of `host:port` is not meaningful at the point of TLS termination and some reverse proxies / load balancers that do TLS termination fail to do so if the `authority` header doesn't match the certificate.

* Testing *

Previously my reverse proxy (Ambassador Emissary Ingress 3.5) would fail match the certificate to the authority and fail to terminate TLS for a gRPC request from k6, and forward the request through, still encrypted:
````
':method', 'POST'
':scheme', 'https'
':path', '/foo/bar'
':authority', 'foo.bar.com:80'
'content-type', 'application/grpc'
'user-agent', 'grpc-go/1.53.0'
```

Now with this PR in place TLS termination and routing is successful:

```
':method', 'POST'
':scheme', 'http'
':path', '/foo/bar'
':authority', 'foo.bar.com'
'content-type', 'application/grpc'
'user-agent', 'grpc-go/1.53.0'
```

### DIFF
--- a/js/modules/k6/grpc/client.go
+++ b/js/modules/k6/grpc/client.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"net"
 	"strings"
 	"time"
 
@@ -133,6 +134,12 @@ func (c *Client) Connect(addr string, params map[string]interface{}) (bool, erro
 		tcred = insecure.NewCredentials()
 	}
 	opts = append(opts, grpc.WithTransportCredentials(tcred))
+
+	host, _, err := net.SplitHostPort(addr)
+	if err != nil {
+		return false, err
+	}
+	opts = append(opts, grpc.WithAuthority(host))
 
 	if ua := state.Options.UserAgent; ua.Valid {
 		opts = append(opts, grpc.WithUserAgent(ua.ValueOrZero()))


ryanbrigden:
@olegbespalov @oleiade Do either of you have bandwidth to provide a review here? Appreciate it!

olegbespalov:
Hey @ryanbrigden !

Sorry for the delay in the response :cry: 

I've checked the changes, and they generally look good. However, I'm slightly concerned that we overwrite the default GRPC client behavior and start using `grpc.WithAuthority()` all the time, which potentially could affect _other_  targets that could expect default behavior. Also, looking at [the specification](https://www.rfc-editor.org/rfc/rfc7540#section-8.1.2.3) of having the port, there shouldn't be an issue.

What I'm thinking is to provide flexibility with the configuration and explicitly define new parameters like we already have https://k6.io/docs/javascript-api/k6-net-grpc/client/client-connect/#connectparams

I believe it will reflect the rationale that we [could find in the GRPC sources](https://github.com/grpc/grpc-go/blob/master/clientconn.go#L889-L898):
```
// getServerName determines the serverName to be used in the connection
// handshake. The default value for the serverName is the authority on the
// ClientConn, which either comes from the user's dial target or through an
// authority override specified using the WithAuthority dial option. Name
// resolvers can specify a per-address override for the serverName through the
// resolver.Address.ServerName field which is used only if the WithAuthority
// dial option was not used. The rationale is that per-address authority
// overrides specified by the name resolver can represent a security risk, while
// an override specified by the user is more dependable since they probably know
// what they are doing.
```

What do you think?

olegbespalov (commented):
@ryanbrigden, what do you think about the https://github.com/grafana/k6/pull/2981#issuecomment-1479068272 ?

ryanbrigden:
@olegbespalov apologies for my delay in response!

Thank you for your detailed look into this change and my use case. I agree that we should follow the specification and that this change as it stands does in fact violate the specification.

I think specifying this as an optional, explicit connection parameter does make sense and I'm open to that solution. On the other hand, what do you think about allowing users to specify server addresses without a port, indicating a default of port `80`.

olegbespalov:
Hey @ryanbrigden 

>What do you think about allowing users to specify server addresses without a port, assuming a default of port 80, as an alternative approach here?

I think that's fine :+1: It's customer input (custom authority) and in that case, we'll follow the _since they probably know
what they are doing._ So if the customer provided that option without the port that's fine.

olegbespalov:
Hey @ryanbrigden 

I want to check with you. Is there any plan to add more commits to that PR? I'm asking because we plan to close it soon and re-open it if needed once more activities happen.

Issue #2959, opened 2023-03-07
handleSummary() accept promise return OR synchronous redisClient


michael-sw-id:
### Feature Description

Provide blocking/sync RedisClient or change handleSummary() to accept promise.

### Already existing or connected issues / PRs (optional)

We would like to use redis and handleSummary to export created customers to json during "test customer creation" run so that we can feed customers to actual load test runs.

Unfortunately:
- RedisClient is non-blocking and requires promises
- HandleSummary() cannot return a promise so we can't use async:await

Example using async:await for handleSummary (because redisclient uses promise api)
```
export const  handleSummary = async (data) => {
  let parsedCustomers = [];

  sleep(1) // for some reason need to wait for redis 
  let redisCustomers = await redisClient.smembers('created_customers')

  console.log(redisCustomers.length + " customers found in redis..")
  redisCustomers.forEach(c => {
    parsedCustomers.push(JSON.parse(c))
  });

  redisClient.del("created_customers")
  return {
    "created-customers.json": JSON.stringify(parsedCustomers),
    "customer-creation-summary.html": htmlReport(data),
    // 'stdout': textSummary(data, { indent: ' ', enableColors: true }),
  };
}
```

The above fails with
```
ERRO[0002] failed to handle the end-of-test summary      error="handleSummary() should return a map with string keys"
```

Using promises it's not possible to pass result to return:

```
export function handleSummary(data) {
  let created_customers = [];

  sleep(1) // for some reason need to wait for redis 
  redisClient.smembers('created_customers')
    .then((customers) => {
      console.log(customers.length + " customers found in redis..")
      customers.forEach(c => {
        created_customers.push(JSON.parse(c))
      });
    })
    .then((_) => {
      redisClient.del("created_customers")
    })
    return {
      'stdout' : textSummary(data, { indent: ' ', enableColors: true }),
      'customer-creation-summary.html': htmlReport(data),
      'created-customers.json': JSON.stringify(created_customers)
    };
}
```

michael-sw-id:
it worked, thanks

```
export function handleSummary(data) {
  let result = {
    'stdout': textSummary(data, { indent: ' ', enableColors: true }),
    "customer-creation-summary.html": htmlReport(data),
  };
  getCreatedCustomersAsync(result).then(
    (customers) => (result["created-customers.json"] = JSON.stringify(customers))
  );
  return result;
}


export const  getCreatedCustomersAsync = async () => {
  let parsedCustomers = [];

[...]

PR #2985, opened 2023-03-23
Support async functions for setup and handleSummary

### LINKED ISSUE #2935: async setup does not create the data object
### Brief summary

When using the async version of setup() the return value of the function call does not create the data object that should be passed in to the default function.

If you change back to a non-async setup call then the data object is set correctly.

### k6 version

0.44

### OS

Mac OS 13.1

_No response_

export async function setup() {
    return {data: "Arbitrary value here"}
}

export default function (data) {
    console.log(data)
}

### Expected behaviour

The data object is created and the console log would display an object { data: "Arbitrary value here" }

### Actual behaviour

The console log will display an empty object.

mstoykov:
This in practice just checks that a promise was not return and get's it result.

Async functions happen to just always return promises.

Fixes #2935 and #2959

### DIFF
--- a/js/runner.go
+++ b/js/runner.go
@@ -10,6 +10,7 @@ import (
 	"net"
 	"net/http"
 	"net/http/cookiejar"
+	"reflect"
 	"strconv"
 	"strings"
 	"sync"
@@ -561,10 +562,17 @@ func (r *Runner) runPart(
 	return v, err
 }
 
+//nolint:gochecknoglobals
+var gojaPromiseType = reflect.TypeOf((*goja.Promise)(nil))
+
 // unPromisify gets the result of v if it is a promise, otherwise returns v
 func unPromisify(v goja.Value) goja.Value {
 	if !isNullish(v) {
-		if p, ok := v.Export().(*goja.Promise); ok {
+		if v.ExportType() == gojaPromiseType {
+			p, ok := v.Export().(*goja.Promise)
+			if !ok {
+				panic("Something that was promise did not export to a promise; this shouldn't happen")
+			}
 			return p.Result()
 		}
 	}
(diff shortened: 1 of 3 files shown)


js/runner.go
```diff
@@ -557,9 +557,25 @@ func (r *Runner) runPart(
...
+// unPromisify gets the result of v if it is a promise, otherwise returns v
+func unPromisify(v goja.Value) goja.Value {
+	if !isNullish(v) {
+		if p, ok := v.Export().(*goja.Promise); ok {
+			return p.Result()
+		}
+	}
+
+	return v
+}
```
na--, line 573: Hmm, `goja.Value.Export()` could be a potentially quite costly operation, right? So, ideally, i think that we should either only do it once and then return the result (if it's not a `Promise`), or have some other way of detecting of the `goja.Value` is a `goja.Promise` underneath :thinking: 

Option one could look somewhat like this:

```suggestion
// unPromisifyAndExport gets the result of v if it is a promise, otherwise returns v
func unPromisifyAndExport(v goja.Value) interface{} {
	if isNullish(v) {
		return v
	}
	exp := v.Export()
	if p, ok := exp.(*goja.Promise); !ok {
		return exp
	}
	return p.Result().Export()
}
```

The downside to this is that we'd need to change the `runFn()` signature or call this on every result of `runFn()`, which doesn't seem too bad.

mstoykov: While `Export` might be expensive(for some deffinition):
1.  the change you propose needs to ripple through each usafe of `runFn` and `runPart` to matter - which quite a bit bigger change
2. The only 2 functions that usually return are `setup` and `handleSummary`. In the common case of a `default` function not returning anything it will be caught by the `isNullish` check continue. I don't think making `setup` or `handleSummary` a bit more expensive is enough reason to make all the changes `1` entails.

mstoykov: [9398f37](https://github.com/grafana/k6/pull/2985/commits/9398f37addf3036c720c0d0489d6e162ed45415b) uses ExportType and reflect to the same check and it seems to work.

This should hopefully also fix any case where `setup` returns too big of an object and `Export` would actually be very expensive.

codebien (approved):

PR #2986, opened 2023-03-23
Refactor isNullish from different package to js.common

mstoykov:


js/modules/k6/experimental/tracing/options.go
```diff
@@ -37,7 +38,7 @@ func newOptions(rt *goja.Runtime, from goja.Value) (options, error) {
 	}
 
 	fromSamplingValue := from.ToObject(rt).Get("sampling")
-	if fromSamplingValue == nil || isNullish(fromSamplingValue) {
+	if fromSamplingValue == nil || common.IsNullish(fromSamplingValue) {
```
na--, line 41:
```suggestion
	if common.IsNullish(fromSamplingValue) {
```

na-- (approved):

js/runner.go
```diff
@@ -572,10 +572,6 @@ func unPromisify(v goja.Value) goja.Value {
 	return v
 }
```
codebien, line 573: I think we could move unpromisify in the common package also

mstoykov: We could ... but nothing else uses it - and I prefer to keep `common` and `util` packages smaller then bigger. 

Arguably a lot of changes will happen around the ESM changes and some other module work - if it turns out this is used a lot we can move it then.

codebien (approved):

na-- (approved):

PR #2990, opened 2023-03-28
Add comment for the modules package

mstoykov:


js/modules/modules.go
```diff
@@ -1,3 +1,4 @@
+// Package modules defines interfaces to be implemented by js modules to be imported by k6.
```
mstoykov, line 1: Not really happy with the wording, but I am working on some PoC stuff and I keep getting linter errors and decided to just get this out of the way :grimacing:

olegbespalov, line 1: Maybe something like this?
```suggestion
// Package modules defines interfaces used for developing k6's JavaScript modules.
```

olegbespalov (approved):

codebien (approved):

PR #2991, opened 2023-03-29
Refactor js modules system in js/modules package and use it

mstoykov:
This is mostly a stepping stone to let it be used in tests but also to make the js package smaller.

This also has a separate commit that uses it in `k6/experimental/tracing` where this actually *matters*.


na-- (dismissed):
I don't know all of the subtle details here, nor do I fully grok the desired end state yet, but this change LGTM on its own.

js/modules/resolution.go
```diff
@@ -0,0 +1,167 @@
...
+)
+
+// FileLoader is a type alias for a function that returns new the contents of the referenced file
+type FileLoader func(specifier *url.URL, name string) ([]byte, error)
+
+type module interface {
+	instantiate(vu VU) moduleInstance
+}
+
+type moduleInstance interface {
```
olegbespalov, line 20: What do you think about removing `module` prefix from the interfaces & structs of that (`modules`) package?

js/bundle.go
```diff
@@ -310,7 +299,12 @@ func (b *Bundle) instantiate(vuImpl *moduleVUImpl, vuID uint64) (moduleInstance,
 		}
 		return nil, err
 	}
-	if exports := instance.exports(); exports == nil {
+	if common.IsNullish(exportsV) {
+		return nil, errors.New("exports must be not nil")
```
olegbespalov, line 303: Is it an end-user message? 

If so, I'd suggest making it something like `exports must be defined` or avoiding using go specific's nil.

js/modules/require_impl.go
```diff
@@ -0,0 +1,68 @@
...
+	"strings"
+
+	"github.com/dop251/goja"
+	"go.k6.io/k6/loader"
+)
+
+// OldRequireImpl is an implementation of require that is the old k6 behaviours
+// It happens to not be commonJS compatible as it loads relative to the currentl required file,
+// instead of relative to the file the `require()` is written in.
+type OldRequireImpl struct {
```
imiric, line 15: "Legacy" is a bit more descriptive than "old". Plus `NewOldRequireImpl()` sounds wrong :wink:

```suggestion
// LegacyRequireImpl is a legacy implementation of `require()` that is not compatible with
// CommonJS as it loads modules relative to the currently required file,
// instead of relative to the file the `require()` is written in.
// See https://github.com/grafana/k6/issues/2674
type LegacyRequireImpl struct {
```

js/modules/resolution.go
```diff
@@ -0,0 +1,167 @@
...
+	"fmt"
+	"net/url"
+	"strings"
+
+	"github.com/dop251/goja"
+	"go.k6.io/k6/js/compiler"
+	"go.k6.io/k6/loader"
+)
+
+// FileLoader is a type alias for a function that returns new the contents of the referenced file
```
imiric, line 13:
```suggestion
// FileLoader is a type alias for a function that returns the contents of the referenced file.
```

js/modules/resolution.go
```diff
@@ -0,0 +1,167 @@
...
+//
+// TODO: this API will likely change as native ESM support will likely not let us have the exports
+// as one big goja.Value that we can manipulate
+func (ms *ModuleSystem) RunSourceData(source *loader.SourceData) (goja.Value, error) {
+	specifier := source.URL.String()
+	pwd := source.URL.ResolveReference(&url.URL{Path: "../"}) // TODO: fix
+	_, err := ms.resolver.resolveLoaded(pwd, specifier, source.Data)
+	if err != nil {
+		return nil, err // TODO wrap as this should never happen
+	}
```
imiric, line 165:
```suggestion
	if _, err := ms.resolver.resolveLoaded(pwd, specifier, source.Data); err != nil {
		return nil, err // TODO wrap as this should never happen
	}
```

js/modules/resolution.go
```diff
@@ -0,0 +1,167 @@
...
+
+// RunSourceData runs the provided sourceData and adds it to the cache.
+// If a module with the same specifier as the source is already cached
+// it will be used instead of reevaluating the source from the provided SourceData.
+//
+// TODO: this API will likely change as native ESM support will likely not let us have the exports
+// as one big goja.Value that we can manipulate
+func (ms *ModuleSystem) RunSourceData(source *loader.SourceData) (goja.Value, error) {
+	specifier := source.URL.String()
+	pwd := source.URL.ResolveReference(&url.URL{Path: "../"}) // TODO: fix
```
imiric, line 161: Can you elaborate the `TODO` here? What should be fixed?

js/modules/resolution.go
mstoykov: I was wondering about this, but in practice all of this will be ... kind of replaced with what goja's ESM will provide later.

I am currently trying to redo the ESM PoC PR on top of this to see how it will look and am (as you probably noticed) backporting some changes.

My predominant problem is that removing `module` from some doesn't work great if I still have to have some called `module` - but I can do it if you want to.

I do expect  you also mean the ModuleSystem, NewModuleSystem and co. and such?

js/modulestest/runtime.go
```diff
@@ -55,3 +58,14 @@ func (r *Runtime) MoveToVUContext(state *lib.State) {
 	r.VU.InitEnvField = nil
 	r.VU.StateField = state
 }
+
+// SetupModuleSystem setups the modules system for the current Runtime
+// see [modules.NewModuleResolver] for the meaining of the parameters
```
imiric, line 63:
```suggestion
// SetupModuleSystem sets up the modules system for the current Runtime.
// See [modules.NewModuleResolver] for the meaning of the parameters.
```

js/modules/k6/experimental/tracing/module_test.go
```diff
@@ -59,16 +59,14 @@ type testSetup struct {
...
-		return rt.ToValue(export).ToObject(rt)
-	}))
+	err := ts.SetupModuleSystem(map[string]interface{}{
+		"k6/http":                 http.New(),
+		"k6/experimental/tracing": new(RootModule),
+	}, nil, compiler.New(ts.VU.InitEnvField.Logger))
+	require.NoError(t, err)
+
+	_, err = ts.VU.Runtime().RunString("var instrumentHTTP = require('k6/experimental/tracing').instrumentHTTP")
+	require.NoError(t, err)
```
imiric, line 69: Nice, much cleaner :+1:

imiric (changes requested):
I like the refactor, but don't have the full context of why this is needed, besides being neater (which is not a reason to not do it).

I just left minor suggestions and questions, and mostly typo fixes.

js/initcontext.go
```diff
@@ -79,49 +78,7 @@ func allowOnlyOpenedFiles(fs fsext.Fs) {
...
-	}
-	if specifier == "" {
-		return nil, errors.New("require() can't be used with an empty specifier")
-	}
-
-	return r.modules.Require(currentPWD, specifier)
-}
-
-func generateSourceMapLoader(logger logrus.FieldLogger, filesystems map[string]fsext.Fs,
+func generateSourceMapLoader(logger logrus.FieldLogger, filesystems map[string]afero.Fs,
```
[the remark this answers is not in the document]
mstoykov: definitely a rebase mistake - good catch :+1:

js/modules.go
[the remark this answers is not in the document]
mstoykov: I think this adds a cycle as the module implementations (k6/http) import `js/modules`. 

We can probably do some more in depth refactoring but this isn't really necessary here

js/modules/resolution.go
mstoykov: just fixed it in e9a1e2f4e

js/modules/require_impl.go
mstoykov: fixed in 5396f22bb

js/initcontext.go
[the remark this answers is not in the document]
mstoykov: fixed in 33c60414d

js/bundle.go
mstoykov: Slightly different message in 039553ea4

olegbespalov: very javascriptish to me :+1:

js/modules/resolution.go
olegbespalov: >I was wondering about this, but in practice all of this will be ... kind of replaced with what goja's ESM will provide later.

It explains, no need to action :+1:

olegbespalov (approved):

mstoykov:
Sorry @olegbespalov it take me a while to rebase and squash the fixes in the correct commits as you can see in https://github.com/grafana/k6/compare/e9a1e2f4ec5f0150b7908d637d1eb24f3d4f61e7..51ee0214bbfe7d8b00664ce776ed917aa804eeca

olegbespalov (approved):
:rocket:

imiric (approved):

PR #2997, opened 2023-03-31
Fix panic handling and add a new exit code and a test for panics

na--:
In https://github.com/grafana/k6/pull/2833 we started calling `os.Exit()` in a `defer` statement, so we inadvertently circumvented the regular handling of panics that bubble up to the top by the Go runtime. 

I could have removed the `os.Exit()` call from the `defer`, but I think this is better. Handling panics ourselves in this way is explicit, simpler to understand, and it allows us to handle them exactly how we want - including assigning them their own unique exit code, and test the whole thing with an e2e test.

### DIFF
--- a/cmd/root_test.go
+++ b/cmd/root_test.go
@@ -5,8 +5,10 @@ import (
 	"testing"
 
 	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
 	"github.com/stretchr/testify/assert"
 	"go.k6.io/k6/cmd/tests"
+	"go.k6.io/k6/errext/exitcodes"
 	"go.k6.io/k6/lib/testutils"
 )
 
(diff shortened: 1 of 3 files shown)


cmd/root_test.go
```diff
@@ -37,3 +39,25 @@ func TestDeprecatedOptionWarning(t *testing.T) {
...
+
+	ts := tests.NewGlobalTestState(t)
+	ts.CmdArgs = []string{"k6", "panic"}
+	ts.ExpectedExitCode = int(exitcodes.Panic)
+
+	rootCmd := newRootCommand(ts.GlobalState)
+	rootCmd.cmd.AddCommand(&cobra.Command{
+		Use: "panic",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			panic("oh no, oh no, oh no,no,no,no,no")
```
imiric, line 54: PTSD dog

errext/exitcodes/codes.go
```diff
@@ -20,4 +20,5 @@ const (
 	CannotStartRESTAPI       ExitCode = 106
 	ScriptException          ExitCode = 107
 	ScriptAborted            ExitCode = 108
+	Panic                    ExitCode = 109
```
imiric, line 23: Just to be clear, since the others are very high-level and functional reasons.

```suggestion
	GoPanic                    ExitCode = 109
```

imiric (approved):
I like it :+1:

I don't see any drawbacks, and this gives us more flexibility like you said, while also logging the panic in external backends.

codebien (approved):

errext/exitcodes/codes.go
na--: Resolved by the [latest force-push](https://github.com/grafana/k6/compare/cf4fd49f36b32dabf34f5474ffcfeec953758c32..e4265bbdf8598a4887a5414b9f62a1719fb17a99)

codebien (approved):

imiric (approved):

Issue #3001, opened 2023-04-03
Investigate profiling with goja


mstoykov:
https://github.com/dop251/goja/commit/6964a11213a2976c329464a341255e0d588b275e added some profiling functionality.

From the API it seems it might not be great for usage across k6, but it is something we should investigate and probably inquire for better per VM API.

PR #3002, opened 2023-04-03
Update goja

mstoykov:
- cpu profiler see #3001
- improved source location of dotted assignment


codebien (approved):
LGTM, did you check the pprof dep impact on the binary size?

mstoykov:
> , did you check the pprof dep impact on the binary size?

36kb on my machine with `CGO_ENABLED=0` which is how we compile either way

na-- (approved):

PR #3004, opened 2023-04-04
Change eventloop.WaitOnRegistered to execute callbacks

mstoykov:
This is a behaviour change which will make it possible for users of the eventloop to use it as the only way to stop stuff, without needing to do additional stuff if the event loop has errored out, detect it and have a separate way to stop without using it.

The above is in practice very error prone and neigh impossible.

In the future when we have additional mechanism for notifying users of an iteration ending, this behaviour will stay the same as using the event loop is still likely the best way to synchronise around stuff such as the runtime.

### DIFF
--- a/js/eventloop/eventloop.go
+++ b/js/eventloop/eventloop.go
@@ -150,6 +150,12 @@ func (e *EventLoop) popAll() (queue []func() error, awaiting bool) {
 	return
 }
 
+func (e *EventLoop) putInfront(queue []func() error) {
+	e.lock.Lock()
+	e.queue = append(queue, e.queue...)
+	e.lock.Unlock()
+}
+
 // Start will run the event loop until it's empty and there are no uninvoked registered callbacks
 // or a queued function returns an error. The provided firstCallback will be the first thing executed.
 // After Start returns the event loop can be reused as long as waitOnRegistered is called.
@@ -167,8 +173,10 @@ func (e *EventLoop) Start(firstCallback func() error) error {
 			continue
 		}
 
-		for _, f := range queue {
+		for i, f := range queue {
 			if err := f(); err != nil {
+				e.putInfront(queue[i+1:])
+
 				return err
 			}
 		}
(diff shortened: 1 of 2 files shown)


mstoykov:
This is used in https://github.com/grafana/xk6-websockets/pull/42 to fix a race which could lead to the whole k6 process waiting indefinitely on a websocket connection.

olegbespalov (approved):

js/eventloop/eventloop.go
```diff
@@ -191,12 +199,25 @@ func (e *EventLoop) Start(firstCallback func() error) error {
...
+			if !awaiting {
+				return
+			}
+			<-e.wakeupCh
+			continue
+		}
+
+		for _, f := range queue {
+			if err := f(); err != nil {
+				// TODO figure out if we should buffer all errors happening or send them on a channel
```
codebien, line 218: I think a channel would be better here, so the caller can decide if wait or not for all of them. Maybe we could also stop to dequeue if the channel has been closed.

codebien (approved):

js/eventloop/eventloop.go
mstoykov: The comment is about returning the errors to the caller.

If the caller *closes* the channel this will panic ;).

One of the reasons for this change is that there really isn't a reason to want to not run them, unless you want to kill the *whole* process. In which case the caller can just not call this or not wait for it to finish.

If the need ever arises for something additional we will add it.

This code definitely will need a lot more work around #2432 for example. But that issue has a lot of other changes that it will need.

mstoykov: Did I understand you correctly @codebien ?

codebien: >The comment is about returning the errors to the caller.

I got it, my proposal is around defining the API for listening about returned errors:
```
WaitOnRegistered(errs <-chan error) {
```

>If the caller closes the channel this will panic ;).

We can do a `select` on this eventual channel as usual.

>If the need ever arises for something additional we will add it.

Anyway, I mainly approved for this reason, at the moment, the demand is to have #2432 more than just an extension here. I was just commenting with my preference on the proposal in the TODO. We should not add this before a concrete vision on #2432.

PR #3005, opened 2023-04-05
Rename samples to examples

mstoykov:


olegbespalov:
@mstoykov it seems like the [e9e84f4](https://github.com/grafana/k6/pull/3005/commits/e9e84f4e28ad4ea770f053acb92f1a2264ab501e) shouldn't be here, right?

olegbespalov:
I think the docker-compose should also be adjusted:

https://github.com/grafana/k6/blob/master/docker-compose.yml#L40

mstoykov:
Nice catch! :clap: :bow: 

I did search for `samples/` and not `/samples` and `samples` ... just returns so many stuff :facepalm: 

I also decided against editing some tests that are still running without a problem as those are just example *strings* and not actually loading the files.

We can update those as well if you want, cc @olegbespalov @oleiade

olegbespalov (approved):
As long as they continue to work, I'm fine :+1:

na-- (approved):

PR #3006, opened 2023-04-05
Update to xk6-websockets v0.2.0

mstoykov:
This "fixes" a potential lock up, but that was already fixed in k6 in #3004.

So this is mostly a somewhat of refactor and simplying PR. 

As well as make `ping` throw an exception if used with a closed connection.

### DIFF
--- a/vendor/modules.txt
+++ b/vendor/modules.txt
@@ -161,8 +161,8 @@ github.com/grafana/xk6-redis/redis
 # github.com/grafana/xk6-timers v0.1.2
 ## explicit; go 1.17
 github.com/grafana/xk6-timers/timers
-# github.com/grafana/xk6-websockets v0.1.6
-## explicit; go 1.17
+# github.com/grafana/xk6-websockets v0.2.0
+## explicit; go 1.18
 github.com/grafana/xk6-websockets/websockets
 github.com/grafana/xk6-websockets/websockets/events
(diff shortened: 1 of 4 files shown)


vendor/modules.txt
```diff
@@ -161,8 +161,8 @@ github.com/grafana/xk6-redis/redis
 # github.com/grafana/xk6-timers v0.1.2
 ## explicit; go 1.17
 github.com/grafana/xk6-timers/timers
-# github.com/grafana/xk6-websockets v0.1.6
-## explicit; go 1.17
+# github.com/grafana/xk6-websockets v0.2.0
+## explicit; go 1.18
```
codebien, line 165: Should we align with k6?

mstoykov: I am not even certain why we have bumpbed to 1.19 - I don't think we use anything from it.

The only reason why I did it for xk6-websockets was to stop having to give `-compat 1.17`, so I don't really care. But will prefer to not have to try to keep this number in sync, given that I don't see a reason to do it.

na--: [I bumped go.mod to `go 1.19`](https://github.com/grafana/k6/commit/52ee1228b75397c20e663378240b8fc4fe612d9d) to align it with the officially supported Go versions, and to also resolve some issues with `go mod tidy` not working well when it was `go 1.17`.

codebien (approved):

PR #3007, opened 2023-04-06
Add a `k6/experimental/webcrypto` module

oleiade:
This PR adds the [xk6-webcrypto](https://github.com/grafana/xk6-webcrypto) extension as an official, experimental k6 module.


js/jsmodules.go
```diff
@@ -18,6 +18,7 @@ import (
 
 	"github.com/grafana/xk6-redis/redis"
 	"github.com/grafana/xk6-timers/timers"
+	webcrypto "github.com/grafana/xk6-webcrypto/webcrypto"
```
mstoykov, line 21:
```suggestion
	"github.com/grafana/xk6-webcrypto/webcrypto"
```

go.mod
```diff
@@ -62,6 +63,7 @@ require (
 	github.com/go-redis/redis/v8 v8.11.5 // indirect
 	github.com/golang/snappy v0.0.4 // indirect
 	github.com/google/pprof v0.0.0-20230207041349-798e818bf904 // indirect
+	github.com/google/uuid v1.3.0 // indirect
```
mstoykov, line 66: we use github.com/nu7hatch/gouuid already - can we use that instead?

mstoykov (commented):
I am not certain if we should have 7 examples ... or examples in the codebase at all to be honest.

We do have them in the docs and in the other repo and we likely will update and add to them. Which means that at some point this `examples` folder will be just a huge amount of random scripts which likely will not be updated or maintained.

This is already a problem for a bunch of the other examples which are 5+ years old and IIRC some of them don't really work anymore as end points dont' return what is expected anymor.

codebien:
>I am not certain if we should have 7 examples ... or examples in the codebase at all to be honest.

I'm fine with adding 1 as `examples/experimental/webcrypto.js` file (maybe with the content of encrypt-decrypt).

oleiade:
I'll drop the examples and keep only a single one 👍🏻 

Regarding the UUID library, considering this is a cryptographic library, I preferred using the Google UUID library, which is more up-to-date (the last update was in 2021 vs. 2013), much more popular, and coming from the company that maintains Go. Also, switching now would mean operating the change/PR on the WebCrypto repo itself. I understand the rationale behind your comment, but considering the existing import is [used only once in the whole k6 codebase](https://github.com/grafana/k6/blob/master/lib/netext/httpext/httpdebug_transport.go#L25), I'd advocate switching to the newly introduced dependency instead. What do you think @mstoykov @codebien ?

codebien:
>Regarding the UUID library, considering this is a cryptographic library, I preferred using the Google UUID library, which is more up-to-date (the last update was in 2021 vs. 2013), much more popular, and coming from the company that maintains Go.

@oleiade I agree with you.

mstoykov (approved):

js/jsmodules.go
codebien: Did we change this?

oleiade: rebase fuck up 🤦🏻

mstoykov (approved):

codebien:
@oleiade apparently go-tip is failing

mstoykov:
This seems like a go tip issue @codebien

codebien (approved):

Issue #2799, opened 2022-12-05
`TestVUIntegrationMetrics` is flaky, make it into an integration test


na--:
This test: https://github.com/grafana/k6/blob/4e4e5b1ff730ce4fcce8f8296b8fbb767ec429d9/js/runner_test.go#L824-L885

Has been quite flaky recently, with failures like [this](https://github.com/grafana/k6/actions/runs/3619814761/jobs/6101303115):
```
2022-12-05T12:08:57.8321834Z time="2022-12-05T12:08:18Z" level=warning msg="Request Failed" error="Get \"http://httpbin.local:41655/html\": lookup httpbin.local: Temporary failure in name resolution"
2022-12-05T12:08:57.8322437Z time="2022-12-05T12:08:18Z" level=info msg=1 source=console
2022-12-05T12:08:57.8323053Z time="2022-12-05T12:08:19Z" level=warning msg="There were unknown fields in the options exported in the script" error="json: unknown field \"myOption\""
2022-12-05T12:08:57.8323733Z time="2022-12-05T12:08:21Z" level=warning msg="There were unknown fields in the options exported in the script" error="json: unknown field \"someField\""
2022-12-05T12:08:57.8325351Z time="2022-12-05T12:08:21Z" level=warning msg="Couldn't load source map for file:///script.js" error="The moduleSpecifier \"file:///test.min.js.map\" couldn't be found on local disk. Make sure that you've specified the right path to the file. If you're running k6 using the Docker image make sure you have mounted the local directory (-v /local/path/:/inside/docker/path) containing your script and modules so that they're accessible by k6 from inside of the container, see https://k6.io/docs/using-k6/modules#using-local-modules-with-docker."
2022-12-05T12:08:57.8327422Z time="2022-12-05T12:08:21Z" level=warning msg="Couldn't load source map for file:///script.js" error="The moduleSpecifier \"file:///test.min.js.map\" couldn't be found on local disk. Make sure that you've specified the right path to the file. If you're running k6 using the Docker image make sure you have mounted the local directory (-v /local/path/:/inside/docker/path) containing your script and modules so that they're accessible by k6 from inside of the container, see https://k6.io/docs/using-k6/modules#using-local-modules-with-docker."
[...]

PR #3012, opened 2023-04-06
Fix TestVUIntegrationMetrics not checking metric exactly

mstoykov:
turns out the test has been *slightly* broken as it checks the metric against a specific builtinMetrics instance.

Unfortunately that was the instance for the "archive" test and the "source" one was never used.

This works as Equals will compare recursively and both the registry and the builtinMetrics should be the same.

I only noticed it as a test failed with this, but it likely was supposed to fail with something differently. For example a sink having a different value :shrug:

### DIFF
--- a/js/runner_test.go
+++ b/js/runner_test.go
@@ -859,25 +859,29 @@ func TestVUIntegrationGroups(t *testing.T) {
 
 func TestVUIntegrationMetrics(t *testing.T) {
 	t.Parallel()
-	r1, err := getSimpleRunner(t, "/script.js", `
+	testdata := make(map[string]*Runner, 2)
+	{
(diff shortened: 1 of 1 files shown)


na-- (approved):
LGTM, as far as this change goes, but the whole test probably needs to be rewritten (https://github.com/grafana/k6/issues/2799). It test stupid things (why do we care if `data_sent` is before `data_received`? :sweat_smile:) in stupid ways...

imiric (approved):
Nice catch :+1:

I do agree that the test is needlessly strict wrt checking the order of received metrics, but don't think we need to convert it to an integration test. It can just be rewritten and simplified as is.

PR #3016, opened 2023-04-10
Drop most ioutil usage across k6

mstoykov:


mstoykov:
There are still some usages which conflict with not using `os` and likely will be better fixed around some work on how we handle files in general. 

So I decided to skip them for now.

na-- (approved):
I was about to push the same changes, I was using them locally to test some other CI experiments :sweat_smile: :rofl:

imiric (approved):

PR #3019, opened 2023-04-13
Decouple the trace id struct generation from its encoding

oleiade:
This PR was triggered by [this comment](https://github.com/grafana/k6/pull/2894#discussion_r1137476666) from @mstoykov. It intends to refactor the trace id structure, and its encoding method in a way that decouples them from each other.

Trace ids are serialised as 16 byte slices, which are in turn encoded into hexadecimal strings that are attached to requests HTTP headers. Some parts of the trace id have a fixed size, some others, a varying one (time, and random suffix). Because some of its parts have a varying size, we need to start serializing it to find out how much space is the random suffix gonna take.

Thus, in order to decouple the trace id creation from its encoding, we have made the trace id struct private (`traceId`) private, and exposed a trace id constructor (`NewTraceID`) which returns an interface object implementing an interface that offers the `Encode() (string, error)` method. That way, when the constructor is called, the serialization happens on the fly, and its result is stored in an internal field.

🦖

### DIFF
--- a/js/modules/k6/experimental/tracing/client.go
+++ b/js/modules/k6/experimental/tracing/client.go
@@ -228,25 +228,18 @@ func (c *Client) instrumentedCall(call func(args ...goja.Value) error, args ...g
 }
 
 func (c *Client) generateTraceContext() (http.Header, string, error) {
-	traceID := TraceID{
-		Prefix:     k6Prefix,
-		Code:       k6CloudCode,
-		Time:       time.Now(),
-		randSource: c.randSource,
-	}
-
-	encodedTraceID, err := traceID.Encode()
+	traceID, err := newTraceID(k6Prefix, k6CloudCode, time.Now(), c.randSource)
 	if err != nil {
-		return http.Header{}, "", fmt.Errorf("failed to encode the generated trace ID; reason: %w", err)
+		return http.Header{}, "", fmt.Errorf("failed to generate trace ID; reason: %w", err)
 	}
 
 	// Produce a trace header in the format defined by the configured propagator.
-	traceContextHeader, err := c.propagator.Propagate(encodedTraceID)
+	traceContextHeader, err := c.propagator.Propagate(traceID)
 	if err != nil {
 		return http.Header{}, "", fmt.Errorf("failed to propagate trace ID; reason: %w", err)
 	}
 
-	return traceContextHeader, encodedTraceID, nil
+	return traceContextHeader, traceID, nil
 }
 
 // instrumentArguments: expects args to be in the format expected by the
--- a/js/modules/k6/experimental/tracing/trace_id.go
+++ b/js/modules/k6/experimental/tracing/trace_id.go
@@ -20,84 +20,47 @@ const (
 
 	// metadataTraceIDKeyName is the key name of the traceID in the output metadata.
 	metadataTraceIDKeyName = "trace_id"
+
+	// traceIDEncodedSize is the size of the encoded traceID.
+	traceIDEncodedSize = 16
 )
 
-// TraceID represents a trace-id as defined by the [W3c specification], and
-// used by w3c, b3 and jaeger propagators. See Considerations for trace-id field [generation]
-// for more information.
+// newTraceID generates a new hexadecimal-encoded trace ID as defined by the [W3C specification].
 //
-// [W3c specification]: https://www.w3.org/TR/trace-context/#trace-id
-// [generation]: https://www.w3.org/TR/trace-context/#considerations-for-trace-id-field-generation
-type TraceID struct {
-	// Prefix is the first 2 bytes of the trace-id, and is used to identify the
-	// vendor of the trace-id.
-	Prefix int16
-
-	// Code is the third byte of the trace-id, and is used to identify the
-	// vendor's specific trace-id format.
-	Code int8
-
-	// Time is the time at which the trace-id was generated.
-	//
-	// The time component is used as a source of randomness, and to ensure
-	// uniqueness of the trace-id.
-	//
-	// When encoded, it should be in a format occupying the last 8 bytes of
-	// the trace-id, and should ideally be encoded as nanoseconds.
-	Time time.Time
-
-	// randSource holds the randomness source to use when encoding the
-	// trace-id. The `rand.Reader` should be your default pick. But
-	// you can replace it with a different source for testing purposes.
-	randSource io.Reader
-}
-
-// Encode encodes the TraceID into a hex string.
+// `prefix` is the first 2 bytes of the trace ID, and is used to identify the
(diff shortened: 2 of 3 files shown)


js/modules/k6/experimental/tracing/trace_id.go
```diff
@@ -22,29 +22,78 @@ const (
...
-// TraceID represents a trace-id as defined by the [W3c specification], and
+// NewTraceID generates a new trace ID.
+//
+// Because the trace ID relies on randomness, and its underlying serialization
+// depends on the time, it is mandatory to use this function to generate
+// trace IDs, instead of creating them manually.
+//
+// As a result this constructor returns an instantiated TraceIDEncoder, which
+// can be used to encode the trace ID into a string.
+func NewTraceID(prefix int16, code int8, t time.Time, randSource io.Reader) (TraceIDEncoder, error) {
```
codebien, line 33: I would discourage this approach. I think it's not really necessary to limit potential future extensions. We can just use the concrete type as an implementation of TraceIDEncoder.

This is still a very valuable guideline to me. https://github.com/golang/go/wiki/CodeReviewComments#interfaces
>Go interfaces generally **belong in the package that uses values** of the interface type, not the package that implements those values. **The implementing package should return concrete** (usually pointer or struct) types: that way, **new methods can be added to implementations without requiring** extensive **refactoring**.



```suggestion
func NewTraceID(prefix int16, code int8, t time.Time, randSource io.Reader) (TraceID, error) {
```

codebien (changes requested):

js/modules/k6/experimental/tracing/trace_id.go
codebien: >The problem this tries to solve is that instantiating a traceId directly

This is not a problem generally we should care in Go. We drive this by documentation if it is really required. The standard library has a lot of these cases. (e.g. [File](https://pkg.go.dev/os@go1.20.3#File), you don't expect to use it by doing `File{}`, [ServeMux](https://pkg.go.dev/net/http#ServeMux)).

>This is why the initial implementation did that in the Encode method.

I'm not against it, it think it is good.

>as a public constructor returning a private struct would also feel wrong

It seems that it is used from the same package so in the end, we don't need to export this function. We can have `newTraceID() traceID`.
Instead, if we need TraceID exported for any reason, then we can just have `NewTraceID() TraceID`. All the fields of TraceID are not exported so they are "protected" from external usage.

oleiade: > The standard library has a lot of these cases. (e.g. [File](https://pkg.go.dev/os@go1.20.3#File), you don't expect to use it by doing File{}, [ServeMux](https://pkg.go.dev/net/http#ServeMux)).

This is a fair point indeed. I did not consider it 👍🏻 

> I'm not against it, I think it is good.
@mstoykov disagrees, hence this PR. I'm somewhere in the middle, I must admit, I'm not necessarily horrified by it, but I prefer to have methods that do a single thing, whereas `Encode` before that PR does both serialisation and encoding.

> Instead, if we need TraceID exported for any reason, then we can just have NewTraceID() TraceID. All the fields of TraceID are not exported so they are "protected" from external usage.

I like the middle-ground there indeed 👍🏻 

Consider me convinced, thanks for your input @codebien 🙇🏻 much appreciated!

So in summary, and based on my assumption that @mstoykov would agree, I would keep make everything private, and have the constructor return a `traceId` instance instead of an interface (which I would drop).

js/modules/k6/experimental/tracing/trace_id.go
```diff
@@ -61,43 +102,22 @@ type TraceID struct {
 // 4. The remaining bytes are filled with random bytes.
 //
 // The resulting 16 bytes sequence is then encoded as a hex string.
-func (t TraceID) Encode() (string, error) {
+func (t traceID) Encode() (string, error) {
 	if !t.isValid() {
 		return "", fmt.Errorf("failed to encode traceID: %v", t)
 	}
```
codebien, line 108: Should we consider it _generation_?

codebien (commented):
LGTM, just one comment

js/modules/k6/experimental/tracing/trace_id.go
```diff
@@ -20,36 +20,77 @@ const (
...
 
-// TraceID represents a trace-id as defined by the [W3c specification], and
+// newTraceID generates a new trace ID.
+//
+// Because the trace ID relies on randomness, and its underlying serialization
+// depends on the time, it is mandatory to use this function to generate
+// trace IDs, instead of creating them manually.
+//
+// As a result this constructor returns an instantiated TraceIDEncoder, which
+// can be used to encode the trace ID into a string.
```
codebien, line 35:
```suggestion
// As a result this constructor returns an instantiated traceID, which
// can be used to encode the trace ID into a string.
```

js/modules/k6/experimental/tracing/trace_id.go
oleiade: I'm sorry but I don't understand what you mean, could you elaborate? Do you mean we should consider renaming the method, or the error message?

codebien: The essence of the API is decoupling the _generation_ from _encoding_. I meant that maybe the validation should be part of the generation so it should stay in the constructor. TBH, not sure so mostly asking for your vision on it.

oleiade: Thanks for the clarification 🙇‍♂️That's a fair point, having moved most of the logic in the constructor, the validation should probably happen there indeed considering the `Encode` only runs hexadecimal encoding at this point 🤝

codebien (approved):

oleiade:
You were misunderstood indeed @mstoykov 😄 
Those are fair points 👍🏻 Dropping the `Encode` method indeed.

js/modules/k6/experimental/tracing/client.go
```diff
@@ -228,11 +228,9 @@ func (c *Client) instrumentedCall(call func(args ...goja.Value) error, args ...g
...
-		Prefix:     k6Prefix,
-		Code:       k6CloudCode,
-		Time:       time.Now(),
-		randSource: c.randSource,
+	traceID, err := newTraceID(k6Prefix, k6CloudCode, time.Now(), c.randSource)
+	if err != nil {
+		return http.Header{}, "", fmt.Errorf("failed to generate trace ID; reason: %w", err)
 	}
 
 	encodedTraceID, err := traceID.Encode()
```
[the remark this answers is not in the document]
oleiade: Gone are the `TraceID` and its `Encode` method indeed. We shall reintroduce them if we need them in the future 👍🏻

mstoykov (approved):
LGTM!

codebien (approved):

PR #3023, opened 2023-04-18
metrics: Support Trend.Max with negative values

codebien:
Max is not computed correctly when negative numbers are used.

### DIFF
--- a/metrics/sink.go
+++ b/metrics/sink.go
@@ -80,18 +80,22 @@ type TrendSink struct {
 func (t *TrendSink) IsEmpty() bool { return t.Count == 0 }
 
 func (t *TrendSink) Add(s Sample) {
+	if t.Count == 0 {
+		t.Max, t.Min = s.Value, s.Value
+	} else {
+		if s.Value > t.Max {
+			t.Max = s.Value
+		}
+		if s.Value < t.Min {
+			t.Min = s.Value
+		}
+	}
+
 	t.Values = append(t.Values, s.Value)
 	t.sorted = false
 	t.Count += 1
 	t.Sum += s.Value
 	t.Avg = t.Sum / float64(t.Count)
-
-	if s.Value > t.Max {
-		t.Max = s.Value
-	}
-	if s.Value < t.Min || t.Count == 1 {
-		t.Min = s.Value
-	}
 }
 
 // P calculates the given percentile from sink values.
--- a/metrics/sink_test.go
+++ b/metrics/sink_test.go
@@ -1,6 +1,7 @@
 package metrics
 
 import (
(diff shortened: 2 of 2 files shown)


metrics/sink.go
```diff
@@ -86,10 +86,14 @@ func (t *TrendSink) Add(s Sample) {
 	t.Sum += s.Value
 	t.Avg = t.Sum / float64(t.Count)
 
+	if t.Count == 1 {
+		t.Max, t.Min = s.Value, s.Value
+		return
+	}
```
mstoykov, line 92: nit: I feel this would be kind of nicer if it is `if t.Count ==0` at the very beginning with everything that happens on the *first* `Add` .

And then have the rest as is.

*Alternatively* we could have TrendSink be created with `NewTrendSink` which sets Max,Min to `-math.MaxFloat64` :thinking: this will make this logic not needed on each `Add`

mstoykov (approved):
LGTM!

Left a comment with some alternatives to the current solution

metrics/sink.go
olegbespalov: The idea of having a proper constructur `NewTrendSink` sounds good :+1:

metrics/sink_test.go
```diff
@@ -93,24 +99,43 @@ func TestTrendSink(t *testing.T) {
 			assert.Equal(t, 100.0, sink.Max)
 			assert.Equal(t, 54.0, sink.Avg)
 		})
+		t.Run("negative", func(t *testing.T) {
```
olegbespalov, line 102: Is it possible to have a "mixed" Trend, e.g., `-1,0,2`? If so, is itng a test case for it is good?

metrics/sink.go
codebien: I also like this proposal, but I'm against doing the change in this PR (essentially in this release) for the following reasons:
1. We already have the Sink API under discussion at https://github.com/grafana/k6/pull/2898 and I would like to get a consensus there before. I need to restart that work.
2. I would like to have consistent Sink implementations so it would require adding the constructors also for the other sinks and it seems to me that it is too much for a last-minute change.

I add it as a suggestion to #2898 and will address it there.

olegbespalov: Make sense! Thanks for the explanation :+1:

olegbespalov (approved):

codebien:
Moved the condition at the beginning so `t.Count == 0 {` can be used. Added the checks for Values and Sum in the tests and removed sorted that is an internal field.

For Sum/Sub operations we should define a precision we would like to have to have better numbers. (Sum is not used in k6 so it is out of scope here).

PR #3025, opened 2023-04-19
Hide afero behind fsext package

mstoykov:
This is mostly so that no other part by `fsext` knows anything about afero.

In the future we can replace the actual implementation.

Updates #1079

### DIFF
--- a/cmd/archive_test.go
+++ b/cmd/archive_test.go
@@ -12,10 +12,10 @@ import (
 	"syscall"
 	"testing"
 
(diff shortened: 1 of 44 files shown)


codebien (approved):
I'm for not including it in the current release.

mstoykov:
>  I'm for not including it in the current release and merging it at the beginning of the next cycle.

This is why it is milestoned for 45 ;)

oleiade (approved):

PR #3027, opened 2023-04-20
output/cloudv2: Trend as Histogram

codebien:
Custom Histogram representation of the Trend metric type. It is the porting of the Histogram generation on the client side.

### DIFF
--- a/output/cloud/expv2/hdr.go
+++ b/output/cloud/expv2/hdr.go
@@ -1,0 +1,270 @@
+package expv2
+
+import (
+	"math"
+	"math/bits"
+	"time"
+
+	"go.k6.io/k6/metrics"
+	"go.k6.io/k6/output/cloud/expv2/pbcloud"
+	"google.golang.org/protobuf/types/known/timestamppb"
+)
+
+const (
+	// lowestTrackable represents the minimum value that the histogram tracks.
+	// Essentially, it excludes negative numbers.
+	// Most of metrics tracked by histograms are durations
+	// where we don't expect negative numbers.
+	//
+	// In the future, we may expand and include them,
+	// probably after https://github.com/grafana/k6/issues/763.
+	lowestTrackable = 0
+
+	// highestTrackable represents the maximum
+	// value that the histogram is able to track with high accuracy (0.1% of error).
+	// It should be a high enough
+	// and rationale value for the k6 context; 2^30 = 1_073_741_824
+	highestTrackable = 1 << 30
+)
+
+// histogram represents a distribution
+// of metrics samples' values as histogram.
+//
+// The histogram is the representation of base-2 exponential Histogram with two layers.
+// The first layer has primary buckets in the form of a power of two, and a second layer of buckets
+// for each primary bucket with an equally distributed amount of buckets inside.
+//
+// The histogram has a series of (N * 2^m) buckets, where:
+// N = a power of 2 that defines the number of primary buckets
+// m = a power of 2 that defines the number of the secondary buckets
+// The current version is: f(N = 25, m = 7) = 3200.
+type histogram struct {
+	// Buckets stores the counters for each bin of the histogram.
+	// It does not include the first and the last absolute bucket,
+	// because they contain exception cases
+	// and they requires to be tracked in a dedicated way.
+	//
+	// It is expected to start and end with a non-zero bucket,
+	// in this way we can avoid extra allocation for not significant buckets.
+	// All the zero buckets in between are preserved.
+	Buckets []uint32
+
+	// ExtraLowBucket counts occurrences of observed values smaller
+	// than the minimum trackable value.
+	ExtraLowBucket uint32
+
+	// ExtraLowBucket counts occurrences of observed values bigger
+	// than the maximum trackable value.
+	ExtraHighBucket uint32
+
+	// FirstNotZeroBucket represents the index of the first bucket
+	// with a significant counter in the Buckets slice (a not zero value).
+	// In this way, all the buckets before can be omitted.
+	FirstNotZeroBucket uint32
+
+	// LastNotZeroBucket represents the index of the last bucket
+	// with a significant counter in the Buckets slice (a not zero value).
+	// In this way, all the buckets after can be omitted.
+	LastNotZeroBucket uint32
+
+	// Max is the absolute maximum observed value.
+	Max float64
+
+	// Min is the absolute minimum observed value.
+	Min float64
+
+	// Sum is the sum of all observed values.
+	Sum float64
+
+	// Count is counts the amount of observed values.
+	Count uint32
+}
+
+// newHistogram creates an histogram of the provided values.
+//
+// TODO: after the aggregation layer probably this constructor
+// doesn't make any sense in this way.
+// It aggregates calling addToBucket time-to-time so
+// trimzeros has to be called at the end of the process.
+func newHistogram(values []float64) histogram {
+	h := histogram{}
+	if len(values) < 1 {
+		return h
+	}
+
+	for i := 0; i < len(values); i++ {
+		h.addToBucket(values[i])
+	}
+
+	h.trimzeros()
+	return h
+}
+
+// addToBucket increments the counter of the bucket
+// releated to the provided value.
+// If the value is lower or higher than the trackable limits
+// then it is counted into specific buckets.
+// All the stats are also updated accordingly.
+//
+// TODO: add the test case in units doing
+// addToBucket + trimzeros + addToBucket
+// so calling addToBucket after trimzeros was called.
+// We don't expect to have this case but the current API
+// would support it so we need to make sure it works or refactor.
+func (h *histogram) addToBucket(v float64) {
+	if h.Count == 0 {
+		h.Max, h.Min = v, v
+	} else {
+		if v > h.Max {
+			h.Max = v
+		}
+		if v < h.Min {
+			h.Min = v
+		}
+	}
+
+	h.Count++
+	h.Sum += v
+
+	if v > highestTrackable {
+		h.ExtraHighBucket++
+		return
+	}
+	if v < lowestTrackable {
+		h.ExtraLowBucket++
+		return
+	}
+
+	index := resolveBucketIndex(v)
+	blen := uint32(len(h.Buckets))
+	if blen == 0 {
+		h.FirstNotZeroBucket = index
+		h.LastNotZeroBucket = index
+	} else {
+		if index < h.FirstNotZeroBucket {
+			h.FirstNotZeroBucket = index
+		}
+		if index > h.LastNotZeroBucket {
+			h.LastNotZeroBucket = index
+		}
+	}
+
+	if index >= blen {
+		h.grow(index)
+	}
+	h.Buckets[index]++
+}
+
+// grow expands the buckets slice
+// with zeros up to the required index
+func (h *histogram) grow(index uint32) {
+	i := int(index)
+	if len(h.Buckets)-1 > i {
+		panic("buckets is already bigger than requested index")
+	}
+	if cap(h.Buckets) > i {
+		// See https://go.dev/ref/spec#Slice_expressions
+		// "For slices, the upper index bound is
+		// the slice capacity cap(a) rather than the length"
+		h.Buckets = h.Buckets[:index+1]
+	} else {
+		length := i + 1
+		// let's make two times larger of
+		// the current request
+		newBuckets := make([]uint32, length, (len(h.Buckets)+length)*2)
+		copy(newBuckets, h.Buckets)
+		h.Buckets = newBuckets
+	}
+}
+
+// trimzeros removes all buckets that have a zero value
+// from the begin and from the end until
+// the first not zero bucket.
+func (h *histogram) trimzeros() {
+	if h.Count < 1 || len(h.Buckets) < 1 {
+		return
+	}
+
+	// all the counters are set to zero, we can remove all
+	if h.FirstNotZeroBucket == 0 && h.LastNotZeroBucket == 0 {
+		h.Buckets = []uint32{}
+		return
+	}
+
+	h.Buckets = h.Buckets[h.FirstNotZeroBucket : h.LastNotZeroBucket+1]
+}
+
+// histogramAsProto converts the histogram into the equivalent Protobuf version.
+func histogramAsProto(h *histogram, time time.Time) *pbcloud.TrendHdrValue {
+	hval := &pbcloud.TrendHdrValue{
+		Time:              timestamppb.New(time),
+		MinResolution:     1.0,
+		SignificantDigits: 2,
+		LowerCounterIndex: h.FirstNotZeroBucket,
+		MinValue:          h.Min,
+		MaxValue:          h.Max,
+		Sum:               h.Sum,
+		Count:             h.Count,
+		Counters:          h.Buckets,
+	}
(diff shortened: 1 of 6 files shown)


output/cloud/expv2/hdr.go
```diff
@@ -0,0 +1,226 @@
...
+// resolveBucketIndex returns the index
+// of the bucket in the histogram for the provided value.
+func resolveBucketIndex(val float64) uint32 {
+	// the lowest trackable value is zero
+	// negative number are not expected
+	if val < 0 {
+		return 0
+	}
+
+	upscaled := int32(math.Ceil(val))
```
na--, line 196: Considering we don't accept negative numbers, this should probably be a `uint32` :thinking:

na--: If it's `uint32`, we won't need to convert it in the `return` below, and we can use something like [`bits.LeadingZeros32`](https://pkg.go.dev/math/bits#LeadingZeros32) to calculate the most significant digit position below much more efficiently than it currently is.

na--: Or, rather, [`bits.Len32`](https://pkg.go.dev/math/bits#Len32) :sweat_smile: Unless I am mistaken, if `upscaled` is uint32, then `nkdiff` can be calculated like this:
```go
nkdiff := bits.Len32(upscaled >> k)
```

output/cloud/expv2/hdr.go
```diff
@@ -0,0 +1,226 @@
...
+		if index > h.LastNotZeroBucket {
+			h.LastNotZeroBucket = index
+		}
+	}
+
+	if index >= blen {
+		// TODO: evaluate if a pool can improve
+
+		// expand with zeros up to the required index
+		h.Buckets = append(h.Buckets, make([]uint32, index-blen+1)...)
```
na--, line 143: :thinking: I think this way of growing a slice's capacity is probably not optimal. This `make()` call will just create a new slice that will immediately be discarded. And if `h.Buckets` doesn't have any space capacity, then `append()` will need to create a new slice and copy both of the slices you pass it anyway, with the second one empty...

So, this will result in either 1 or 2 immediately garbage collected slices and unneeded copying. Something like this should be more efficient:
```suggestion
		// TODO: evaluate if a pool can improve

		// expand with zeros up to the required index
		if cap(h.Buckets) > index {
			// See https://go.dev/ref/spec#Slice_expressions
			// "For slices, the upper index bound is the slice capacity cap(a) rather than the length"
			h.Buckets = h.Buckets[:index+1]
		} else {
			newBuckets := make([]uint32, index+1) // TODO: maybe add extra cap?
			copy(newBuckets, h.Buckets)
			h.Buckets = newBuckets
		}
```

I haven't tested this, so it's quite likely it has at least off-by-one errors somewhere, but I was just illustrating the principle :sweat_smile: In any case, it might make more sense to move this into a separate `grow()` helper method or function.

output/cloud/expv2/output.go
```diff
@@ -0,0 +1,266 @@
...
+				})
+			}
+			pb.Samples = &pbcloud.TimeSeries_RateSamples{
+				RateSamples: rateSamples,
+			}
+		case metrics.Trend:
+			trendSamples := &pbcloud.TrendHdrSamples{}
+			for _, trendSample := range samples {
+				hdrValue := histogramAsProto(
+					newHistogram([]float64{trendSample.Value}),
```
na--, line 253: Something is wonky here, we call `newHistogram()` with only a single sample value :confused: And we don't call `addToBucket()` anywhere.

output/cloud/expv2/hdr.go
codebien: https://github.com/grafana/k6/pull/3027/commits/3efff370bf2dd0bf6aa6121c26c90c74270bb499

output/cloud/expv2/hdr.go
codebien: https://github.com/grafana/k6/pull/3027/commits/40ad58a495881d00d2748dd5df4af62c9b2617a5

output/cloud/expv2/hdr.go
```diff
@@ -0,0 +1,246 @@
...
+	// let u = upscaled
+	// let n = msb(u) - most significant digit position
+	// i.e. n = floor(log(u, 2))
+	//   major_bucket_index = n - k + 1
+	//   sub_bucket_index = u>>(n - k) - (1<<k)
+	//   bucket = major_bucket_index << k + sub_bucket_index =
+	//          = (n-k+1)<<k + u>>(n-k) - (1<<k) =
+	//          = (n-k)<<k + u>>(n-k)
+	//
+	nkdiff := uint32(bits.Len32(upscaled>>7) - 1) // msb index
```
na--, line 244:
```suggestion
	nkdiff := uint32(bits.Len32(upscaled>>k) - 1) // msb index
```

output/cloud/expv2/hdr.go
```diff
@@ -0,0 +1,246 @@
...
+	}
+
+	//
+	// Here we use some math to get simple formula
+	// derivation:
+	// let u = upscaled
+	// let n = msb(u) - most significant digit position
+	// i.e. n = floor(log(u, 2))
+	//   major_bucket_index = n - k + 1
+	//   sub_bucket_index = u>>(n - k) - (1<<k)
```
na--, line 239: Adding this mostly as a reminder to myself to double-check next week, since I don't immediately understand this line - why can `sub_bucket_index` be calculated like this? :thinking:

The `n`th bucket contains `2^(n-1)` items, equally divided in `2^k` sub-buckets, right? So, I'd expect something like "zero out up to the `n`th bit to get the part of the number that needs to be divided into sub-buckets, divide it by `2^k` to get the sub-bucket index". 

But I'm probably missing something stupid, so if you have an explanation handy, please add it to the comment

output/cloud/expv2/integration/testdata/metricset.json
```diff
@@ -119,7 +119,23 @@
...
+                "time": "2023-05-01T04:00:01.500Z",
+                "count": 1,
+                "counters": [
+                  0,
+                  0,
+                  0,
+                  0,
+                  0,
+                  0,
+                  1
```
codebien, line 131: auto spot as a reminder, now with the aggregation we need to cover the trimzeros from `addToBucket`

output/cloud/expv2/hdr.go
```diff
@@ -0,0 +1,270 @@
...
+		newBuckets := make([]uint32, length, (len(h.Buckets)+length)*2)
+		copy(newBuckets, h.Buckets)
+		h.Buckets = newBuckets
+	}
+}
+
+// trimzeros removes all buckets that have a zero value
+// from the begin and from the end until
+// the first not zero bucket.
+func (h *histogram) trimzeros() {
```
oleiade, line 183: I've noticed in other PRs using this that this method had to be explicitly called here and there.

Just sharing a thought I had and I don't have an alternative proposal to offer, yet it felt like the potential for future bugs "we had this bug because we didn't call `trimzeroes` as we should have", and I wondered if there might be opportunities to avoid having to call it explicitly.

Is that relevant?

output/cloud/expv2/hdr.go
```diff
@@ -0,0 +1,270 @@
...
+// resolveBucketIndex returns the index
+// of the bucket in the histogram for the provided value.
+func resolveBucketIndex(val float64) uint32 {
+	// the lowest trackable value is zero
+	// negative number are not expected
+	if val < 0 {
+		return 0
+	}
+
+	upscaled := uint32(math.Ceil(val))
```
oleiade, line 228:
```suggestion
 	// We upscale to the next integer to ensure that each sample falls 
 	// within a specific bucket, even when the value is fractional. 
 	// This avoids under-representing the distribution in the histogram.
	upscaled := uint32(math.Ceil(val))
```

output/cloud/expv2/hdr.go
```diff
@@ -0,0 +1,270 @@
...
+
+	// k is a power of 2 closest to 10^precision_points
```
oleiade, line 230:
```suggestion
    // In histograms, bucket boundaries are usually defined as multiples of powers of 2, 
 	// allowing for efficient computation of bucket indexes.
 	//
 	// We define k=7 in our case, because it allows for sufficient granularity in the
 	// distribution (2^7=128 primary buckets of which each can be further 
 	// subdivided if needed). 
 	//
 	// k is the constant balancing factor between granularity and 
 	// computational efficiency.
 	//
 	// In our case:
 	// i.e 2^7  = 128  ~  100 = 10^2
 	//     2^10 = 1024 ~ 1000 = 10^3
 	// f(x) = 3*x + 1 - empiric formula that works for us
 	// since f(2)=7 and f(3)=10
```

output/cloud/expv2/hdr.go
```diff
@@ -0,0 +1,270 @@
...
+	// f(x) = 3*x + 1 - empiric formula that works for us
+	// since f(2)=7 and f(3)=10
+	const k = uint32(7)
+
+	// 256 = 1 << (k+1)
+	if upscaled < 256 {
+		return upscaled
+	}
+
+	//
```
oleiade, line 244:
```suggestion
	// `nkdiff` helps us find the right bucket for `upscaled`. It does so by determining the 
 	// index for the "major" bucket (a set of values within a power of two range) and then
 	// the "sub" bucket within that major bucket. This system provides us with a fine level
 	// of granularity within a computationally efficient bucketing system. The result is a 
 	// histogram that provides a detailed representation of the distribution of values.
 	//
```

oleiade (commented):
Some proposals to provide a bit more context to the documentation of `resolveBucketIndex` function 🙇🏻

output/cloud/expv2/hdr.go
codebien: Done

output/cloud/expv2/hdr.go

output/cloud/expv2/hdr.go
codebien: Done

output/cloud/expv2/hdr.go
```diff
@@ -0,0 +1,273 @@
...
+	}
+
+	newLen := index - h.FirstNotZeroBucket + 1
+
+	if uint32(cap(h.Buckets)) > newLen {
+		// See https://go.dev/ref/spec#Slice_expressions
+		// "For slices, the upper index bound is
+		// the slice capacity cap(a) rather than the length"
+		h.Buckets = h.Buckets[:newLen]
+	} else {
```
mstoykov, line 177: I don't see how this slice will ever have different cap and len as you never use `append` with it in a way that will introduce that.

So IMO these checks(and the one in growLeft) are pointless - you will always need to allocate a new one as you never actually let go allocate more for you.

output/cloud/expv2/hdr.go
```diff
@@ -0,0 +1,273 @@
...
+	}
+	return hval
+}
+
+// resolveBucketIndex returns the index
+// of the bucket in the histogram for the provided value.
+func resolveBucketIndex(val float64) uint32 {
+	// the lowest trackable value is zero
+	// negative number are not expected
+	if val < 0 {
```
mstoykov, line 211:
```suggestion
	if val < lowestTrackable {
```

output/cloud/expv2/hdr.go
```diff
@@ -0,0 +1,273 @@
...
+		hval.ExtraHighValuesCounter = &h.ExtraHighBucket
+	}
+	return hval
+}
+
+// resolveBucketIndex returns the index
+// of the bucket in the histogram for the provided value.
+func resolveBucketIndex(val float64) uint32 {
+	// the lowest trackable value is zero
+	// negative number are not expected
```
mstoykov, line 210:
```suggestion
```

No need for the comment if we have the self documenting code ;)

output/cloud/expv2/hdr.go
codebien: [At the begin](https://github.com/grafana/k6/blob/128e55a042c258d238786b9ba887093884299316/output/cloud/expv2/hdr.go#L120-L123) we allocate an empty slice with some cap (my expectation is: it saves from all that aggregation where we have very close/similar values). It should be most of the case for successful tests.

mstoykov: I did miss that. Why don't we create this in the constructor? 

I would argue that a lot of the code will be much more clean if creating a new histogram takes the first value. This way a bunch of the `if`s just dissappaer. I guess there the usage will be slightly different but arguably not that much :thinking:

codebien: Take a look at https://github.com/grafana/k6/pull/3027/commits/e01da6c2ca4522dd0041c68b1f187943e8b2385d, is it better? Note that it allocates without being useful in the case of un trackable values. But probably we don't care considering it shouldn't be a high percentage of the cases.

output/cloud/expv2/hdr.go
```diff
@@ -0,0 +1,266 @@
...
+		h.LastNotZeroBucket = index
+		h.Buckets = append(h.Buckets, 1)
+		return
+	}
+
+	if index < h.FirstNotZeroBucket {
+		h.growLeft(index)
+		h.FirstNotZeroBucket = index
+	}
+	if index > h.LastNotZeroBucket {
```
mstoykov, line 131:
```suggestion
	} else if index > h.LastNotZeroBucket {
```

output/cloud/expv2/hdr.go
```diff
@@ -0,0 +1,265 @@
...
+	}
+
+	if index < h.FirstNotZeroBucket {
+		h.growLeft(index)
+		h.FirstNotZeroBucket = index
+	} else if index > h.LastNotZeroBucket {
+		h.growRight(index)
+		h.LastNotZeroBucket = index
+	}
+	h.Buckets[index-h.FirstNotZeroBucket]++
```
mstoykov, line 134:
```suggestion
	if index < h.FirstNotZeroBucket {
		h.growLeft(index)
	} else if index > h.LastNotZeroBucket {
		h.growRight(index)
	} else {
		h.Buckets[index-h.FirstNotZeroBucket]++
	}
```

it seems to me that grow* should:
1. set teh last* as well.
2. add the `1` to the bucket that was added as the index.

output/cloud/expv2/hdr.go
```diff
@@ -0,0 +1,278 @@
...
+
+	// FirstNotZeroBucket represents the index of the first bucket
+	// with a significant counter in the Buckets slice (a not zero value).
+	// In this way, all the buckets before can be omitted.
+	FirstNotZeroBucket uint32
+
+	// LastNotZeroBucket represents the index of the last bucket
+	// with a significant counter in the Buckets slice (a not zero value).
+	// In this way, all the buckets after can be omitted.
+	LastNotZeroBucket uint32
```
mstoykov, line 68: nit: you use `first`/`last` here but `left`/`right` in the grow function ... it might be slightly better if we have a consistent language around this.

Maybe `growHead`/`growEnd` or `growStart`/`growEnd`, `prepend`/`append`

mstoykov (dismissed):
LGTM! 

I have left a naming nitpick which we can address later or never for that matter :sweat_smile:

mstoykov (dismissed):

output/cloud/expv2/hdr.go
codebien: I think `prepend / append` is the best considering the context. @oleiade WDYT?

PR #3030, opened 2023-04-24
cloud: only upload test archive without running it 

mstoykov:
This will let users update the archive for a test without needing to
running it at the exact time of updating it.


mstoykov:
The final CLI UI looks like 
image
Which is in line with `--exit-on-running` and how it looks with  just `k6 cloud` so it seems like the best option

olegbespalov (approved):

oleiade (approved):
Great work 👏🏻 🎉 🦖

mstoykov:
Documented in https://github.com/grafana/k6-docs/pull/1206

PR #3033, opened 2023-04-25
Make the linter CI action reusable by other repos

na--:
This allows other repositories to reuse the k6 linter checks, somewhat like this:
```yml
name: Lint
on:
  push:
    branches:
      - main
  pull_request:

jobs:
  lint:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v3
      - name: Run k6 golangci-lint CI
        uses: grafana/k6/.github/actions/lint@master
```

### DIFF
--- a/.github/actions/lint/action.yml
+++ b/.github/actions/lint/action.yml
@@ -1,0 +1,28 @@
+name: golangci-lint-k6
+description: Run golangci-lint with the k6 ruleset
+
+runs:
+  using: composite
+  steps:
+    - name: Install Go
+      uses: actions/setup-go@v4
+      with:
+        go-version: 1.20.x
+        check-latest: true
+    - name: Get golangci-lint version and download rules
+      shell: bash
+      env: #TODO: remove after https://github.com/actions/runner/issues/2473 is fixed
+        GITHUB_ACTION_REF: ${{ github.action_ref || github.head_ref }}
+      run: |
+        RulesURL="https://raw.githubusercontent.com/grafana/k6/${{ env.GITHUB_ACTION_REF }}/.golangci.yml"
+        echo "Downloading ${RulesURL}..."
+        curl --silent --show-error --fail --no-location "${RulesURL}" --output "${{ github.action_path }}/.golangci.yml"
+
+        echo "GolangCIVersion=$(head -n 1 "${{ github.action_path }}/.golangci.yml" | tr -d '# ')" >> "${GITHUB_OUTPUT}"
+      id: getenv
+    - name: golangci-lint
+      uses: golangci/golangci-lint-action@v3
+      with:
+        version: ${{ steps.getenv.outputs.GolangCIVersion }}
+        only-new-issues: true
+        args: "--config=${{ github.action_path }}/.golangci.yml"
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -39,18 +39,5 @@ jobs:
         uses: actions/checkout@v3
         with:
           fetch-depth: 0
-      - name: Install Go
-        uses: actions/setup-go@v3
-        with:
-          go-version: 1.20.x
-          check-latest: true
-      - name: Retrieve golangci-lint version
-        run: |
-          echo "Version=$(head -n 1 "${GITHUB_WORKSPACE}/.golangci.yml" | tr -d '# ')" >> $GITHUB_OUTPUT
-        id: version
-      - name: golangci-lint
-        uses: golangci/golangci-lint-action@v3
-        with:
-          # Required: the version of golangci-lint is required and must be specified without patch version: we always use the latest patch version.
-          version: ${{ steps.version.outputs.Version }}
-          only-new-issues: true
+      - name: Run linters
+        uses: ./.github/actions/lint/


.github/actions/lint/action.yml
```diff
@@ -0,0 +1,23 @@
...
+      shell: bash
+      run: |
+        echo "PWD=$(pwd)" >> $GITHUB_OUTPUT
+        echo "GolangCIVersion=$(head -n 1 "./.golangci.yml" | tr -d '# ')" >> $GITHUB_OUTPUT
+      id: getenv
+    - name: golangci-lint
+      uses: golangci/golangci-lint-action@v3
+      with:
+        version: ${{ steps.getenv.outputs.GolangCIVersion }}
+        only-new-issues: true
```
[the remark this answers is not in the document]
imiric: Yes, I've also ran into weirdness with this action's `only-new-issues` not working exactly as `--new-from-rev` does on the CLI. :face_exhaling:

Could we make this value configurable then, so that each repo can decide whether they want to enable it or not?

That said, I think the fact that the action actually makes GH comments on the lines with linter issues is a major UX win. Nobody wants to waste time going to the job output to see what failed :sweat_smile: That was a major reason we decided to switch to the action, after all, since we used the binary for a long time before that.

So I think the shortcomings of `only-new-issues` not working always should be a motivator to fix the lint issues, not to abandon the action. :)

imiric (commented):
Making the linter into a reusable action is a good idea, but I'm not so sure about the symlink.

For one, how does this work on Windows? [This SO answer](https://superuser.com/a/1736577) implies it would need a Git config change (or maybe this is for Windows symlinks only?). :-/ I don't have a Windows VM handy right now. Could you confirm?

I know we don't have Windows developers on the team, but we shouldn't make development difficult on other platforms. We might have external Windows developers, after all.

Could we maybe avoid the symlink by making the action download k6's file from `master`?

.github/actions/lint/action.yml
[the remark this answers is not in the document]
na--: > Could we make this value configurable then, so that each repo can decide whether they want to enable it or not?

Probably no point, since even if we make it configurable, it simply won't work in some cases... :disappointed: 

At the moment, it will work correctly with a `pull_request` dispatch. And it will work... if not correctly, than at least how you'd expect it, when we run this action on `master` / `main` branch - it will show all linter errors. 

However, if some project wants to reuse the action in another branch without a PR, and only show the new errors that are not present in `master`, it won't work now or even if we make `only-new-issues` configurable. Thankfully, we don't have cases like this yet, and I don't see a reason why we'd want that. But if we ever do, maybe we can make the `args:` value configurable and keep the `golangci/golangci-lint-action` action :thinking: 

Anyway, having played with this a bit now, I am fine to keep the current action, and I agree that the inline comments it provides are nice.

.golangci.yml
```diff
@@ -1,129 +0,0 @@
-# v1.51.2
```
na--, line 1: @imiric, to answer https://github.com/grafana/k6/pull/3033#pullrequestreview-1403575421 in a separate thread, I haven't tested how the symlink will work on Windows, but they should just work, NTFS supports them: https://blogs.windows.com/windowsdeveloper/2016/12/02/symlinks-windows-10/

Though, now that I think about this, `make ci-like-lint` won't work at the moment :disappointed: https://github.com/grafana/k6/blob/9fa50b2d1f259cdccff5cc7bc18a236d31c345ac/Makefile#L12-L17

> Could we maybe avoid the symlink by making the action download k6's file from master?

Hmm :thinking: If the `.golangci.yml` file is in the root of k6 repo and we download it on every CI run, then `make ci-like-lint` will continue working in this repo, though not in other repos that we want to switch to this approach :thinking: 

However, we won't be able to change the `.golangci.yml` file in a PR and observe how it affects the linter check results :thinking: They will always be downloaded from `master`.

na--: Actually, my bad, `make ci-like-lint` works in the k6 repo, because the whole folder is mounted in `/app`, including both the symlink and the actual `.golangci.yml`. However, it might not work in other repos that use these linting rules

imiric: I'm familiar with Windows symlinks, but they're an NTFS feature. I'm also concerned with how Git handles links across platforms, and [it's not straightforward](https://stackoverflow.com/a/11664406).

My point was that by using symlinks we complicate things substantially, so we should try to avoid them. If someone wants to develop k6 on Android on a FAT32-formatted SD card, we shouldn't make this more difficult for them.

Apparently [Git will probe if symlinks are supported, and set the value of `core.symlinks` appropriately](https://git-scm.com/docs/git-config#Documentation/git-config.txt-coresymlinks), but do we want to depend on this functionality? Does it work only on fresh clones, or on branch pulls as well?

Docker was also historically weird with symlink support, so I'm not quite sure if `make ci-like-lint` would work on macOS or Windows with this change. I personally don't find that make target useful, so I wouldn't mind removing it, actually.

So I'd strongly suggest we avoid using symlinks, and find an alternative way to make this work.

> However, we won't be able to change the `.golangci.yml` file in a PR and observe how it affects the linter check results :thinking_face: They will always be downloaded from `master`.

I think this is a minor issue compared to using a symlink. But it could be worked around by downloading from `$GITHUB_REF` if we're running in this repository, or from `master` otherwise.

.github/actions/lint/action.yml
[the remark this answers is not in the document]
olegbespalov: >We might want to get rid of this action and use the golangci-lint binary directly, for more flexibility and fewer bugs.

If it is possible, I'd prefer to keep it as action since it also has a nice feature of commenting on the pull request. So UX is better there. 

>It also doesn't always work correctly, e.g. on push events in branches or when running with https://github.com/nektos/act, even with a pull_request event, probably because of https://github.com/golangci/golangci-lint-action/issues/531.

But do we have a concrete example of misbehavior on our codebase? :thinking:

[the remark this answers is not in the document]
na--: As I said, it doesn't work as we'd expect when testing with `act` or when running in a naked branch. But it works as we want it when running it in a pull request or `master` / `main`, which is what we actually use, and the UX is indeed better as you both point out. So I'll resolve this thread and keep the action.

.golangci.yml
na--: > If someone wants to develop k6 on Android on a FAT32-formatted SD card, we shouldn't make this more difficult for them.

This is a bit of a stretch... :sweat_smile: In the very unlikely event that someone wants to do that, or develop it on a Windows XP machine, or something like that, I'd argue that they can rely on the linting we have in our CI :sweat_smile: Being able to run `golangci-lint` locally is not strictly required for anything.

> I think this is a minor issue compared to using a symlink. But it could be worked around by downloading from `$GITHUB_REF` if we're running in this repository, or from `master` otherwise.

It is not that simple, since the custom `composite` action is checked out differently when you reuse in another repo with `uses:`, but there actually might be a way to have this working the way you want. I'll play around with `{{ github.action_ref }}` and see how things work out.

olegbespalov: I am sharing my opinion. 

I'm okay with symlinks if they continue working for 80% of the cases. Yes, they are unpleasant, but it sorts the source-of-true issue for the k6 repository without additional actions from contributors/maintainers, which is good.

>If someone wants to develop k6 on Android on a FAT32-formatted SD card, we shouldn't make this more difficult for them.

But having a symlink doesn't block them from developing k6. They can still do that if they have a git and some text editor. And they still could use the `golangci-lint --config .github/actions/lint/.golangci.lint`, right? We could even stress it via makefile. Sure, it's overcomplicating the `inline` warnings for LSP, but other workarounds are possible there.

Regarding the `make ci-like-lint`, we could eliminate it if the team agrees and nobody uses them. I've constantly used it when we have "reversed" config file, and each update of the `goalngci-lint` causes many issues. But nowadays, it's almost not the case.

However, maybe having a customer `k6-golangci-linter` docker image could simplify a bit the possibility of using the rules for the external repositories.

olegbespalov (approved):
Generally, we probably need to test LGTM on various OSes to ensure that it keeps everything simple.

.golangci.yml
na--: OK, after stepping on a few rakes (including the quite annoying to pin down https://github.com/actions/runner/issues/2473 / https://github.com/actions/runner/issues/2525 :man_facepalming:), I think I've resolved this problem :crossed_fingers: :pray: 

`.golangci.yml` can continue living in the k6 repo's root folder, while the reusable GitHub action downloads the correct version of the file on every run.

@imiric, please resolve this thread if you like the current version.

olegbespalov (dismissed):
Awesome, great job @na-- ! :+1:

.github/actions/lint/action.yml
```diff
@@ -0,0 +1,28 @@
...
+        go-version: 1.20.x
+        check-latest: true
+    - name: Get golangci-lint version and download rules
+      shell: bash
+      env: #TODO: remove after https://github.com/actions/runner/issues/2473 is fixed
+        GITHUB_ACTION_REF: ${{ github.action_ref || github.head_ref }}
+      run: |
+        RulesURL="https://raw.githubusercontent.com/grafana/k6/${{ env.GITHUB_ACTION_REF }}/.golangci.yml"
+        echo "Downloading ${RulesURL}..."
+        curl "${RulesURL}" --output "${{ github.action_path }}/.golangci.yml"
```
imiric, line 19:
```suggestion
        curl -fsSL "${RulesURL}" --output "${{ github.action_path }}/.golangci.yml"
```

I instinctively use these flags in scripts. It's silent and safer.

.golangci.yml
imiric: Thanks for taking the time @na-- :bow:

imiric (approved):
LGTM now, thanks!

.github/actions/lint/action.yml
na--: Resolved by the [latest force-push](https://github.com/grafana/k6/compare/09b47c9c6e9dda7c323ed20a5389d4860df2025e..6de4f1c4c2666717f116b047759f4ee47bee0d9a)?

Though I used the long-form options, and I changed `-L` to its opposite, `--no-location`, since I don't want curl to follow redirects in this case.

imiric (approved):

olegbespalov (approved):

imiric:
@na-- Can you update the PR description to remove the mention of a symlink, for posterity? :pray:

Issue #3022, opened 2023-04-18
Expose a new `browser` specific option in `scenarios`


ankur22:
## How we currently work with the `browserType` and `browser`
With the `k6-browser` module we currently have to be very explicit in how we define and create the `browser` object (which starts and connects to the browser instance), which we now feel isn't a great DX. We also allow any number of browsers to be launched, which we think is a feature that isn't particularly useful. Each iteration will start and stop a browser process, which we think is wasteful since running browser instances is an expensive process.

In general we want to be able to abstract away the lifecycle of running the browser, and at the same time reduce the number of browser instances a test run needs to work with.

### Suggested Solution (optional)

We think the solution is for the k6-browser module (or an external process) to take ownership of the browser instances lifecycle depending on a heuristic that it sees fit for the test requirements. For the module/process to determine how many browser instances it requires, we want to leverage `scenarios` which will help us determine how many vus a test run will need. The k6-browser module will restrict the number of `browserContexts` to one per vu per iteration. So in the following example where we have defined one `scenario`, we know that the test will require 10 vus, and the module/process can work with this number to start up a suitable number of browser instances that can be shared between all the iterations.

```js
export const options = {
  scenarios: {
    ui: {
      executor: 'constant-vus',
      exec: 'ui',
      vus: 10,
      duration: '10s',
      options: {
        browser: {
          type: 'chromium'
        },
      },
    },
  },
};

export async function ui() {
  // test code here
}
```

To clarify, the requested change is specifically to add this:

```js
      options: {
        browser: {
          type: 'chromium'
        },
      },
```

Note that the validation of the properties should be done by the k6-browser module.

imiric:
@ankur22 I have a few questions about this feature.

1. > The k6-browser module will restrict the number of browserContexts to one per vu per iteration.

    This doesn't seem different from the current implementation if the script launches and closes one browser instance in the `default` function. I was under the impression that one k6 VU would map exactly to one `BrowserContext`, regardless of the number of iterations. I.e. k6-browser would launch a browser (context) when it's initialized (considering there's only one instance of the browser module per VU to begin with), and then this browser context would remain active for the lifetime of that VU, and would be reused by every iteration that VU executes.

    The above "one BrowserContext per VU _per iteration_" sounds confusing to me, so please let me know if my assumption is correct. Your later note that "browser instances [...] can be shared between all the iterations" hints towards this, but I wanted to confirm we're on the same page.


[...]

PR #3036, opened 2023-04-26
Add JS scenario options

imiric:
This is a quick and possibly naive change to support browser options per scenario (#3022). These options will be parsed and validated by the browser module.

It doesn't try to enable generic options for any JS module as #3000 did, but it seems Ned abandoned this idea. We can open this discussion again once #883 is resolved.

I tested it with `k6 archive`, and it does serialize and deserialize from `metadata.json` properly. Should I add a test for this? An archive unit test would be a bit convoluted to add because of the `ExecutorConfig` interface, but maybe an integration test would be simpler.


js/modules/k6/execution/execution_test.go
```diff
@@ -190,7 +190,7 @@ func TestAbortTest(t *testing.T) { //nolint:tparallel
...
...tionSegment":"0:1/4","executionSegmentSequence":"0,1/4,1/2,1","noSetup":true,"setupTimeout":"1m0s","noTeardown":true,"teardownTimeout":"5m0s","rps":100,"dns":{"ttl":"1m","select":"roundRobin","policy":"any"},"maxRedirects":3,"userAgent":"k6-user-agent","batch":15,"batchPerHost":5,"httpDebug":"full","insecureSkipTLSVerify":true,"tlsCipherSuites":["TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256"],"tlsVersion":{"min":"tls1.2","max":"tls1.3"},"tlsAuth":[{"domains":["example.com"],"cert":"mycert.pem","key":"mycert-key.pem","password":"mypwd"}],"throw":true,"thresholds":{"http_req_duration":[{"threshold":"rate>0.01","abortOnFail":true,"delayAbortEval":"10s"}]},"blacklistIPs":["192.0.2.0/24"],"blockHostnames":["test.k6.io","*.example.com"],"hosts":{"test.k6.io":"1.2.3.4:8443"},"noConnectionReuse":true,"noVUConnectionReuse":true,"minIterationDuration":"10s","ext":{"ext-one":{"rawkey":"rawvalue"}},"summaryTrendStats":["avg","min","max"],"summaryTimeUnit":"ms","systemTags":["iter","vu"],"tags":null,"metricSamplesBufferSize":8,"noCookiesReset":true,"discardResponseBodies":true,"consoleOutput":"loadtest.log","tags":{"runtag-key":"runtag-value"},"localIPs":"192.168.20.12-192.168.20.15,192.168.10.0/27"}`
```
codebien, line 193:
```suggestion
	expected := `{"paused":true,"scenarios":{"const-vus":{"executor":"constant-vus","options":{"browser": test},"startTime":"10s","gracefulStop":"30s","env":{"FOO":"bar"},"exec":"default","tags":{"tagkey":"tagvalue"},"vus":50,"duration":"10m0s"}},"executionSegment":"0:1/4","executionSegmentSequence":"0,1/4,1/2,1","noSetup":true,"setupTimeout":"1m0s","noTeardown":true,"teardownTimeout":"5m0s","rps":100,"dns":{"ttl":"1m","select":"roundRobin","policy":"any"},"maxRedirects":3,"userAgent":"k6-user-agent","batch":15,"batchPerHost":5,"httpDebug":"full","insecureSkipTLSVerify":true,"tlsCipherSuites":["TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256"],"tlsVersion":{"min":"tls1.2","max":"tls1.3"},"tlsAuth":[{"domains":["example.com"],"cert":"mycert.pem","key":"mycert-key.pem","password":"mypwd"}],"throw":true,"thresholds":{"http_req_duration":[{"threshold":"rate>0.01","abortOnFail":true,"delayAbortEval":"10s"}]},"blacklistIPs":["192.0.2.0/24"],"blockHostnames":["test.k6.io","*.example.com"],"hosts":{"test.k6.io":"1.2.3.4:8443"},"noConnectionReuse":true,"noVUConnectionReuse":true,"minIterationDuration":"10s","ext":{"ext-one":{"rawkey":"rawvalue"}},"summaryTrendStats":["avg","min","max"],"summaryTimeUnit":"ms","systemTags":["iter","vu"],"tags":null,"metricSamplesBufferSize":8,"noCookiesReset":true,"discardResponseBodies":true,"consoleOutput":"loadtest.log","tags":{"runtag-key":"runtag-value"},"localIPs":"192.168.20.12-192.168.20.15,192.168.10.0/27"}`
```
We should have a basic set here and not `null`.

codebien (changes requested):

mstoykov (approved):
LGTM! but I would prefer to hear that this will actually be sufficient for the @grafana/k6-browser team.

js/modules/k6/execution/execution_test.go
imiric: Sure, fixed in latest push.

inancgumus (approved):
LGTM 👍 Thanks for this 🙇

codebien (approved):

PR #3037, opened 2023-04-26
k6/execution: Add metadata JS API

### LINKED ISSUE #2766: JS API for user-specified high-cardinality metrics metadata
### Feature Description

k6 v0.41.0 added a `Metadata` property for storing high-cardinality metadata to all `metric.Sample` values that is not part of the `Sample`'s `TimeSeries`: https://github.com/grafana/k6/blob/e14f14d416b0af4e495e6f565a2a9787a73ade54/metrics/sample.go#L19-L33

However, we decided (https://github.com/grafana/k6/issues/2584#issuecomment-1269775631) to not merge the originally proposed JS APIs from https://github.com/grafana/k6/pull/2654 that allowed users to set whatever metadata values they wanted from their scripts, since they presented a somewhat confusing UX. The `Sample.Metadata` property is currently only accessible from Go code, either the built-in k6 modules and outputs or from [xk6](https://github.com/grafana/xk6) extensions.

I am adding this issue as a placeholder, so it can be referenced by other issues. We still haven't reached a consensus how (and even if) the `Metadata` should be made accessible from user scripts. 

A major use case for this feature is distributed tracing (https://github.com/grafana/k6/issues/2128), though it doesn't necessarily need JS code to be able to access and set custom metrics metadata. If anyone has another use cases that will benefit from this feature, please share them here.

No consensus and firm proposals yet. There are various ideas with different pros and cons, and this deserves a broader discussion. Some of them:
- at this point the original API with non-indexed tags (https://github.com/grafana/k6/pull/2654) is probably out of the running, though it still has its benefits
- a promising idea is to have a `metrics.withMetadata(metaObj, callback)` API similar to `group()`, though that has its own problems with async APIs (https://github.com/grafana/k6/issues/2728)
- a `metadata` property in all of the calls where we currently support `tags` has the problem that it's yet another custom thing that needs to be implemented in every protocol, it is ambiguous (is the "metadata" property about the call and not its metrics?) and it conflicts with the already existing `metadata` property in gRPC calls.

We are discussing this internally, but if someone has API proposals or ideas, please share them here.

- https://github.com/grafana/k6/issues/2584
- https://github.com/grafana/k6/issues/2580
- https://github.com/grafana/k6/pull/2726
- https://github.com/grafana/k6/issues/2128
- https://github.com/grafana/k6/issues/2762

mstoykov:
A bare minimum API to set and get metadata set for the whole VU.

It is added on top of the vu property and on top of a new `metrics` one so. If `k6/execution` is imported as `exec` -
`exec.vu.metrics.metadata["foo"]` will get you the currently set metadata for `foo` the same way `exec.vu.tags["bar"]` will get you the value of the tag `bar`.

Setting metadata is the same.

It also adds `exec.vu.metrics.tags` which is the same as `exec.vu.tags` for consistency. While leaving the old one for backwards compatibility.

This also includes dropping the ability to list metadata keys and values from `exec.vu.tags`. This was likely left in by mistake.

This is a breaking change due to the fact that before that `iter` and `vu` could've been accessed this way but this will now need to be done through the new `exec.vu.metrics.metadata`.

Closes #2766

### DIFF
--- a/js/common/tags.go
+++ b/js/common/tags.go
@@ -56,7 +56,7 @@ func ApplyCustomUserTag(tagsAndMeta *metrics.TagsAndMeta, key string, val goja.V
 }
 
 // ApplyCustomUserMetadata modifies the given metrics.TagsAndMeta object with the
-// given custom metadata and theirs value.
+// given custom metadata and their value.
 func ApplyCustomUserMetadata(tagsAndMeta *metrics.TagsAndMeta, key string, val goja.Value) error {
 	kind := reflect.Invalid
 	if typ := val.ExportType(); typ != nil {
@@ -76,7 +76,7 @@ func ApplyCustomUserMetadata(tagsAndMeta *metrics.TagsAndMeta, key string, val g
 	default:
 		return fmt.Errorf(
 			"invalid value for metric metadata '%s': "+
-				"only String, Boolean and Number types are accepted as a metric metadata values",
+				"only String, Boolean and Number types are accepted as a metric metadata value",
 			key,
 		)
 	}
--- a/js/common/tags.go
+++ b/js/common/tags.go
@@ -20,7 +20,7 @@ func ApplyCustomUserTags(rt *goja.Runtime, tagsAndMeta *metrics.TagsAndMeta, key
 	keyValuesObj := keyValues.ToObject(rt)
 
 	for _, key := range keyValuesObj.Keys() {
-		if err := ApplyCustomUserTag(rt, tagsAndMeta, key, keyValuesObj.Get(key)); err != nil {
+		if err := ApplyCustomUserTag(tagsAndMeta, key, keyValuesObj.Get(key)); err != nil {
 			return err
 		}
 	}
@@ -29,8 +29,8 @@ func ApplyCustomUserTags(rt *goja.Runtime, tagsAndMeta *metrics.TagsAndMeta, key
 }
 
 // ApplyCustomUserTag modifies the given metrics.TagsAndMeta object with the
-// given custom tag or metadata and theirs value.
-func ApplyCustomUserTag(rt *goja.Runtime, tagsAndMeta *metrics.TagsAndMeta, key string, val goja.Value) error {
+// given custom tag and theirs value.
+func ApplyCustomUserTag(tagsAndMeta *metrics.TagsAndMeta, key string, val goja.Value) error {
 	kind := reflect.Invalid
 	if typ := val.ExportType(); typ != nil {
 		kind = typ.Kind()
@@ -50,6 +50,33 @@ func ApplyCustomUserTag(rt *goja.Runtime, tagsAndMeta *metrics.TagsAndMeta, key 
 		return fmt.Errorf(
 			"invalid value for metric tag '%s': "+
 				"only String, Boolean and Number types are accepted as a metric tag values",
+			key,
+		)
+	}
+}
+
+// ApplyCustomUserMetadata modifies the given metrics.TagsAndMeta object with the
+// given custom metadata and theirs value.
+func ApplyCustomUserMetadata(tagsAndMeta *metrics.TagsAndMeta, key string, val goja.Value) error {
+	kind := reflect.Invalid
+	if typ := val.ExportType(); typ != nil {
+		kind = typ.Kind()
+	}
+
+	switch kind {
+	case
+		reflect.String,
+		reflect.Bool,
+		reflect.Int64,
+		reflect.Float64:
+
+		tagsAndMeta.SetMetadata(key, val.String())
+		return nil
+
+	default:
+		return fmt.Errorf(
+			"invalid value for metric metadata '%s': "+
+				"only String, Boolean and Number types are accepted as a metric metadata values",
 			key,
 		)
 	}
--- a/js/modules/k6/execution/execution.go
+++ b/js/modules/k6/execution/execution.go
@@ -216,11 +216,34 @@ func (mi *ModuleInstance) newVUInfo() (*goja.Object, error) {
 	if err != nil {
 		return o, err
 	}
-
-	err = o.Set("tags", rt.NewDynamicObject(&tagsDynamicObject{
+	tagsDynamicObject := rt.NewDynamicObject(&tagsDynamicObject{
 		runtime: rt,
 		state:   vuState,
-	}))
+	})
+	err = o.Set("tags", tagsDynamicObject)
+
+	if err != nil {
+		return o, err
+	}
+	metrics, err := newInfoObj(rt, map[string]func() interface{}{
+		"tags": func() interface{} { return tagsDynamicObject },
+		"metadata": func() interface{} {
+			return rt.NewDynamicObject(&metadataDynamicObject{
+				runtime: rt,
+				state:   vuState,
+			})
+		},
+	})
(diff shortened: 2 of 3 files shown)


js/modules/k6/execution/execution_test.go
```diff
@@ -84,24 +109,29 @@ func TestVUTagsSetSuccessAccetedTypes(t *testing.T) {
...
+				require.NoError(t, err)
+
+				assert.Equal(t, tc.exp, val.String())
+			}
+		})
 	}
 }
 
-func TestVUTagsSuccessOverwriteSystemTag(t *testing.T) {
+func TestVUTagMetadatasSuccessOverwriteSystemTag(t *testing.T) {
```
codebien, line 134:
```suggestion
func TestVUTagsMetadatasSuccessOverwriteSystemTag(t *testing.T) {
```
nit

js/modules/k6/execution/execution_test.go
```diff
@@ -34,43 +41,61 @@ func setupTagsExecEnv(t *testing.T) *modulestest.Runtime {
...
+			// not found
+			tag, err = tenv.VU.Runtime().RunString(fmt.Sprintf(`exec.vu.%s["not-existing-%s"]`, prop, propType))
+			require.NoError(t, err)
+			assert.Equal(t, "undefined", tag.String())
+		})
+	}
 }
 
-func TestVUTagsJSONEncoding(t *testing.T) {
+func TestVUTagMetadasJSONEncoding(t *testing.T) {
```
codebien, line 70:
```suggestion
func TestVUTagsMetadatasJSONEncoding(t *testing.T) {
```

js/modules/k6/execution/execution_test.go
```diff
@@ -34,43 +41,61 @@ func setupTagsExecEnv(t *testing.T) *modulestest.Runtime {
 	return testRuntime
 }
 
-func TestVUTagsGet(t *testing.T) {
+func TestVUTagMetadatasGet(t *testing.T) {
```
codebien, line 44:
```suggestion
func TestVUTagsMetadatasGet(t *testing.T) {
```

codebien (commented):
LGTM

mstoykov:
note: I will just squash at the end

js/common/tags.go
```diff
@@ -54,3 +54,30 @@ func ApplyCustomUserTag(rt *goja.Runtime, tagsAndMeta *metrics.TagsAndMeta, key
 		)
 	}
 }
+
+// ApplyCustomUserMetadata modifies the given metrics.TagsAndMeta object with the
+// given custom metadata and theirs value.
```
imiric, line 59:
```suggestion
// given custom metadata and their value.
```

js/common/tags.go
```diff
@@ -54,3 +54,30 @@ func ApplyCustomUserTag(rt *goja.Runtime, tagsAndMeta *metrics.TagsAndMeta, key
...
+		reflect.Int64,
+		reflect.Float64:
+
+		tagsAndMeta.SetMetadata(key, val.String())
+		return nil
+
+	default:
+		return fmt.Errorf(
+			"invalid value for metric metadata '%s': "+
+				"only String, Boolean and Number types are accepted as a metric metadata values",
```
imiric, line 79:
```suggestion
				"only String, Boolean and Number types are accepted as a metric metadata value",
```

js/modules/k6/execution/execution.go
```diff
@@ -362,6 +379,61 @@ func (o *tagsDynamicObject) Keys() []string {
...
+// Get a property value for the key. May return nil if the property does not exist.
+func (o *metadataDynamicObject) Get(key string) goja.Value {
+	tcv := o.state.Tags.GetCurrentValues()
+	if metadatum, ok := tcv.Metadata[key]; ok {
+		return o.runtime.ToValue(metadatum)
+	}
+	return nil
+}
+
+// Set a property value for the key. It returns true if succeed. String, Boolean
```
imiric, line 399:
```suggestion
// Set a property value for the key. It returns true if successful. String, Boolean
```

js/modules/k6/execution/execution.go
```diff
@@ -362,6 +379,61 @@ func (o *tagsDynamicObject) Keys() []string {
...
+func (o *metadataDynamicObject) Set(key string, val goja.Value) bool {
+	var err error
+	o.state.Tags.Modify(func(tagsAndMeta *metrics.TagsAndMeta) {
+		err = common.ApplyCustomUserMetadata(tagsAndMeta, key, val)
+		if err != nil {
+			panic(o.runtime.NewTypeError(err.Error()))
+		}
+	})
+	return true
+}
```
imiric, line 411: Is there a reason the `err` is declared outside of the closure?

I suppose this method used to return it before, and now we always return `true`, but it can just be scoped within the closure. And the `err != nil` check can be done on the same line.

Same goes for `tagsDynamicObject.Set()`, since this was copied from there, so we might as well update it there too.

js/modules/k6/execution/execution_test.go
```diff
@@ -34,43 +41,61 @@ func setupTagsExecEnv(t *testing.T) *modulestest.Runtime {
...
+
+			encoded, err := tenv.VU.Runtime().RunString(fmt.Sprintf(`JSON.stringify(exec.vu.%s)`, prop))
+			require.NoError(t, err)
+			assert.JSONEq(t, fmt.Sprintf(`{"%[1]s-vu":"42","custom-%[1]s":"my%[1]s1"}`, propType), encoded.String())
+		})
+	}
 }
 
-func TestVUTagsSetSuccessAccetedTypes(t *testing.T) {
+func TestVUTagMetadatasSetSuccessAccetedTypes(t *testing.T) {
```
imiric, line 98:
```suggestion
func TestVUTagMetadatasSetSuccessAcceptedTypes(t *testing.T) {
```

imiric (commented):
LGTM, just minor suggestions.

js/modules/k6/execution/execution.go
```diff
@@ -216,11 +216,34 @@ func (mi *ModuleInstance) newVUInfo() (*goja.Object, error) {
...
-
-	err = o.Set("tags", rt.NewDynamicObject(&tagsDynamicObject{
+	tagsDynamicObject := rt.NewDynamicObject(&tagsDynamicObject{
 		runtime: rt,
 		state:   vuState,
-	}))
+	})
+	err = o.Set("tags", tagsDynamicObject)
+
+	if err != nil {
```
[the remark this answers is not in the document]
mstoykov: I would prefer to add an issue to discuss if we even want to deprecate it :shrug:

js/modules/k6/execution/execution.go
mstoykov: I guess that is true - I just copied and edited the code from the tags. I did try to figure out if I can make it more generic - but unfortunately that seems like *way* more work then it would save code.

mstoykov: fixed in [f011c23](https://github.com/grafana/k6/pull/3037/commits/f011c23571315c37102a709ad00cda42a9a6ecc9)

imiric (approved):

codebien (approved):

mstoykov:
documented in https://github.com/grafana/k6-docs/pull/1207

PR #3040, opened 2023-04-27
tc39: enable modules tests

mstoykov:
A bunch are failing here as there is just no way to be implemented by a transpiler or because they are checking stuff babel doesn't know off.


oleiade (approved):
LGTM 👍🏻 as far as I can tell 😅

olegbespalov (approved):

js/tc39/tc39_test.go
```diff
@@ -647,6 +658,59 @@ func (ctx *tc39TestCtx) runTC39Script(name, src string, includes []string, vm *g
...
+	comp.Options = compiler.Options{Strict: false, CompatibilityMode: lib.CompatibilityModeExtended}
+
+	mr := modules.NewModuleResolver(nil,
+		func(specifier *url.URL, name string) ([]byte, error) {
+			return fs.ReadFile(currentFS, specifier.Path[1:])
+		},
+		comp)
+	u, err := url.Parse(path.Join(ctx.base, name))
+	if err != nil {
+		panic(err) // TODO fix
```
olegbespalov, line 697: Non-blocking: Out of curiosity, can we give more details for future fixing?

olegbespalov (approved):

js/tc39/tc39_test.go
mstoykov: Like fb139a484

olegbespalov (approved):

PR #3041, opened 2023-04-27
output/cloud: Versioning

codebien:
It implements the ability to have multiple cloud versions and it addresses a specific request change https://github.com/grafana/k6/pull/2963#discussion_r1157118508.

### DIFF
--- a/output/cloud/output_test.go
+++ b/output/cloud/output_test.go
@@ -127,3 +127,23 @@ func TestOutputCreateTestWithConfigOverwrite(t *testing.T) {
 
 	require.NoError(t, out.StopWithTestError(nil))
 }
+
+func TestOutputStartVersionError(t *testing.T) {
+	t.Parallel()
+	o, err := newOutput(output.Params{
+		Logger: testutils.NewLogger(t),
+		ScriptOptions: lib.Options{
+			Duration:   types.NullDurationFrom(1 * time.Second),
+			SystemTags: &metrics.DefaultSystemTagSet,
+		},
+		Environment: map[string]string{
+			"K6_CLOUD_API_VERSION": "3",
+		},
+		ScriptPath: &url.URL{Path: "/script.js"},
+	})
+	require.NoError(t, err)
+
+	o.referenceID = "123"
+	err = o.startVersionedOutput()
+	require.ErrorContains(t, err, "v3 is an unexpected version")
+}
--- a/output/cloud/output_test.go
+++ b/output/cloud/output_test.go
@@ -1,7 +1,9 @@
 package cloud
 
 import (
+	"errors"
 	"fmt"
+	"io"
 	"net/http"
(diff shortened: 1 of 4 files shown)


output/cloud/output_test.go
```diff
@@ -127,3 +132,177 @@ func TestOutputCreateTestWithConfigOverwrite(t *testing.T) {
...
+	called := false
+	o := &Output{
+		versionedOutput: versionedOutputMock{
+			callback: func() {
+				called = true
+			},
+		},
+	}
+	o.AddMetricSamples([]metrics.SampleContainer{})
+	assert.True(t, called)
```
olegbespalov, line 289: Looking into the `versionedOutputMock`, it's hard to say that this test case is valid. 

`callback` is calling from every method of the `versionedOutputMock`, so as the reader of the test, I could not be sure that the callback was called from the `AddMetricSamples`. Maybe at some point, the `StopWithTestError` was also called.

So it's better to explicitly register which method was executed (like passing the method name to the callback and keeping the state, whatever) instead of a generic callback.

output/cloud/output_test.go
```diff
@@ -127,3 +132,177 @@ func TestOutputCreateTestWithConfigOverwrite(t *testing.T) {
...
+		o := Output{referenceID: "74"}
+		o.config.WebAppURL = null.StringFrom("mywebappurl.com")
+		assert.Equal(t, "cloud (mywebappurl.com/runs/74)", o.Description())
+	})
+}
+
+func TestOutputStopWithTestError(t *testing.T) {
+	t.Parallel()
+
+	done := make(chan struct{}, 1)
```
olegbespalov, line 203: Why is the chan buffered? Could we produce more than one call? :thinking:

codebien: Otherwise, it deadlocks on `StopWithTestError` waiting for the server to send the response.

My original idea was to have this test with the cloudapi.Client mocked so we don't need to have the test server and concurrency here but unfortunately, the v1 requires it so we can't just add an interface to mock it. When we will drop the v1 we may consider to do it to restrict the scope. Ideally, splitting between two interfaces one for the test life cycle management and one for metrics flush.

codebien:
@olegbespalov I also added one more test, this should now be ready for a new review.

output/cloud/output_test.go
olegbespalov: Not sure if I'm getting why it could deadlock :thinking: 

Like I assume that the HTTP calling code is async. That's why you used a `chan` to ensure that call happened. If so, you don't need to send something to the chan, you could just close the chan once the handler responds.

If like you're saying, the StopWithTestError could lead to a deadlock. The code seems to me synchronous, and you probably don't need any chan at all :thinking:

na-- (approved):

PR #3046, opened 2023-05-01
fix: use default buildx builder

nrxr:
This is a fix for the PR #3015. Using an external builder will create issues for docker loading images locally and then won't be able to push the multi-platform images correctly.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -141,11 +141,7 @@ jobs:
           fetch-depth: 0
       - name: Build
         run: |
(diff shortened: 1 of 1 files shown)


nrxr:
This seems to be [only offered on Docker Desktop so far and as a beta feature](https://docs.docker.com/desktop/containerd/), hence the right solution would be the pragmatic way or the easy way.

Error:

```
ERROR: multiple platforms feature is currently not supported for docker driver. Please switch to a different driver (eg. "docker buildx create --use")
```

PR #3047, opened 2023-05-01
fix: load multi-platform build

nrxr:
This loads the just built docker image with --load and makes it available to push later on.

Fixes output from #3015


nrxr:
ERROR: docker exporter does not currently support exporting manifest lists

PR #3053, opened 2023-05-03
Add `GetScenarioOptions` method to `ExecutorConfig` interface

imiric:
This simplifies retrieval of `ScenarioOptions` (added in #3036), since it avoids type assertions of all possible `ExecutorConfig` implementations. See [this comment](https://github.com/grafana/xk6-browser/issues/858#issuecomment-1532652663).

I also added an archive test to serialize the scenario options and read them back, which was arguably missing from #3036.


ka3de (approved):
Thanks for this! LGTM.

lib/executor/executors_test.go
```diff
@@ -444,3 +449,68 @@ func TestConfigMapParsingAndValidation(t *testing.T) {
...
+				"/path/to/a.js": []byte(`// a contents`),
+			}),
+		},
+	}
+
+	buf := bytes.NewBuffer(nil)
+	require.NoError(t, arc1.Write(buf))
+
+	arc2, err := lib.ReadArchive(buf)
+	require.NoError(t, err)
```
codebien, line 500: We are doing more of an integration test here, if I'm getting the code right, why not just test the marshal operation as we did before?

I agree that an archive integration test could be helpful but it should assert the entire scenario config and not only one value.

codebien (changes requested):

lib/executor/executors_test.go
imiric: :confused: I don't see it as an integration test, but as an archive unit test, similar to the ones we have [here](https://github.com/grafana/k6/blob/698e894a36bebb60a9895d3abe8e94702b929a3b/lib/archive_test.go#L117). The only reason this is in `lib/executor/` instead is to avoid an import cycle (`lib` -> `lib/executor` -> `lib`), because we need to import a `lib.ExecutorConfig` implementation.

> why not just test the marshal operation as we did before?

As we did before, where?

> I agree that an archive integration test could be helpful but it should assert the entire scenario config and not only one value.

Sure, I can add more scenario assertions, if you want. I didn't consider anything else important, since I only wanted to confirm that the new `options` field is (de)serialized properly.

imiric: @codebien See 59d17daff.

codebien (commented):
LGTM, waiting for the green CI

lib/executor/executors_test.go
imiric: I don't see how the JSON API can be the minimal unit, given that this test is written from the perspective of `lib.Archive`. It just does a round-trip read/write to confirm all options are intact. The JSON API is not involved here at all, and the only "components" are `lib.Archive` and `lib.executor`... :confused: But let's not get into the semantics of an integration/unit test... :sweat_smile: It's really not important in this case, anyway.

I added a comment in 2d82729b2. The import cycle was also mentioned before in the commit message of 07762a22d.

`TestConfigMapParsingAndValidation` only does the unmarshalling process, and not the round-trip. I also thought involving `lib.Archive` was important, as these options need to be present in the archive `metadata.json`, so didn't just want to do `json.Marshal()` on them, but to test it with the actual archive.

imiric:
Apparently [there's an issue building gotip](https://github.com/grafana/k6/actions/runs/4894850279/jobs/8740008329) :face_exhaling:

```
go: downloading golang.org/dl v0.0.0-20230502172222-5216546bad51
Cloning into '/home/runner/sdk/gotip'...
Updating the go development tree...
From https://go.googlesource.com/go
 * branch            master     -> FETCH_HEAD
HEAD is now at 761e813 cmd/compile: work with new bisect command
Building Go cmd/dist using /opt/hostedtoolcache/go/1.20.4/x64. (go1.20.4 linux/amd64)
Building Go toolchain1 using /opt/hostedtoolcache/go/1.20.4/x64.
# bootstrap/internal/abi
/home/runner/sdk/gotip/src/internal/abi/unsafestring_go120.go:13: unsafe.String requires go1.20 or later (-lang was set to go1.16; check go.mod)
/home/runner/sdk/gotip/src/internal/abi/unsafestring_go120.go:17: unsafe.Slice requires go1.17 or later (-lang was set to go1.16; check go.mod)
go tool dist: FAILED: /opt/hostedtoolcache/go/1.20.4/x64/bin/go install -tags=math_big_pure_go compiler_bootstrap purego bootstrap/cmd/...: exit status 1
gotip: failed to build go: exit status 2
```

codebien (approved):

imiric:
@codebien Can you also review #3057, since you mentioned you want to merge that one before this?

PR #3056, opened 2023-05-05
Switch to fasthttp websocket

joeky888:
Related to https://github.com/grafana/xk6-websockets/pull/47


imiric (commented):
Hi there, thanks for creating this PR.

That's quite a large vendor diff, and I wonder about a couple of things.

Is the fasthttp fork an officially endorsed fork by the community? There are 3.3k forks, and any one of them could be better maintained.

I know fasthttp is well trusted within the Go ecosystem, but it doesn't mean this project will have the same success. So far they've only [updated some dependencies](https://github.com/fasthttp/websocket/commits/master), which is good, but doesn't necessarily inspire confidence.

That said, all of this might not be an issue for us, since:

a) this is a low effort change from our side, with not much impact (it only adds ~400kB to the binary),

b) the current library is quite stable, so major changes shouldn't be needed,

c) it's a good stop-gap measure until we actually replace it with something more performant and modern, as mentioned in https://github.com/grafana/xk6-websockets/issues/9#issue-1312826595.

So I'm OK with this change, but let's wait for other opinions. I'll add @olegbespalov as well, since @mstoykov is currently on PTO.

BTW, please rebase on latest `master` to resolve that `build-docker` failure. We'll need to look into the the `test-prev` and `lint` failures, but they're not related to this PR.

joeky888:
> Is the fasthttp fork an officially endorsed fork by the community? 

It is used by [fiber websocket](https://github.com/gofiber/websocket/blob/master/go.mod) and maintained by the fasthttp/fiber community.

> as mentioned in https://github.com/grafana/xk6-websockets/issues/9#issue-1312826595.

All of the mentioned are okay, but not a drop-in replacement for gorilla/websocket at this point.

> BTW, please rebase on latest master to resolve that build-docker failure. We'll need to look into the test-prev and lint failures, but they're not related to this PR.

I'll do it once https://github.com/grafana/xk6-websockets/pull/47 is merged. Because after that I can remove more gorilla related dependencies in the go.sum.

mstoykov:
Thanks for the PR!

The idea of that issue is to change the library to something that not only is not `gorilla/websocket` but actually performs better and or is easier to use. Given that the websocket specification hasn't really changed and we do not have any *particular* issue with the current library I wouldn't just change it to change it.

Given fasthttp track record on being specification compliant - I am personally am not for goign with something that seems to use it.

We are very likely to be performance bottlenecked way before the code for websocket networking being the problem. So even if fasthttp version is somehow faster - it wouldn't really matter IMO.

I would also recommend that this is only done for the experimental websocket and then if there are some noticeable improvements we can backport it (maybe).

So ... uh I am for closing this - and you(@joeky888) can try to prove that this makes enough of a difference while not breaking more stuff from the [compliance testing](https://github.com/grafana/xk6-websockets/tree/main/examples/autobahn) in the [extension repo](https://github.com/grafana/xk6-websockets).

imiric:
Hey @joeky888, the team has discussed this PR internally, and like @mstoykov mentioned, we don't see much benefit of changing this here. It would make more sense for the xk6-websockets repository, since it's still an experimental module in k6, but we might try an entirely different WS library first there instead.

So I'll close this PR, and leave @mstoykov and @olegbespalov to respond in https://github.com/grafana/xk6-websockets/pull/47. Thanks for proposing this anyway! :bow:

PR #3057, opened 2023-05-05
Make scenario options JSON field optional

imiric:
Previously, the new scenario `options` field added in #3036 was being serialized to JSON by default, which made it an incompatible change with some of our internal Cloud services that validate the existing configuration structure.

This change makes the object optional, and it will only be serialized if it's included in the data.

This is based on #3053, so let's merge that first.


codebien:
>This is based on https://github.com/grafana/k6/pull/3053, so let's merge that first.

LGTM, just to make it clear I would like to merge this PR first, because it only overwrites code added in 3053 so it doesn't make sense to me to have it on master without this.

imiric:
This PR doesn't overwrite #3053. It's just based on it, because it would create a conflict otherwise.

But sure, the end result will be the same whichever is merged first.

codebien (approved):

PR #3060, opened 2023-05-08
docker: Do not pin the ca-certs pkg

codebien:
Hadolint advises to pin a version, but in this specific case isn't very useful because:
- We always want the latest version for the ca-certificates package
- It has a date-based versioning so we couldn't use the common `=~` selector for patch updates. It mostly breaks every time the Alpine repository releases a new version because they don't keep the previous one.

See also: https://github.com/hadolint/hadolint/issues/464#issuecomment-1386948671

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,8 @@ RUN apk --no-cache add git=~2
 RUN CGO_ENABLED=0 go install -a -trimpath -ldflags "-s -w -X go.k6.io/k6/lib/consts.VersionDetails=$(date -u +"%FT%T%z")/$(git describe --tags --always --long --dirty)"
 
 FROM alpine:3.17
-RUN apk add --no-cache ca-certificates=~20220614 && \
+# hadolint ignore=DL3018
+RUN apk add --no-cache ca-certificates && \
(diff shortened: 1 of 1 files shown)


olegbespalov (dismissed):

Dockerfile
```diff
@@ -15,7 +15,7 @@ RUN go build -a -trimpath -ldflags "-s -w -X go.k6.io/k6/lib/consts.VersionDetai
 
 # Runtime stage
 FROM alpine:3.17
-RUN apk add --no-cache ca-certificates=~20220614 && \
+RUN apk add --no-cache ca-certificates && \
```
imiric, line 18: Maybe let's add an [ignore rule for Hadolint](https://github.com/hadolint/hadolint/blob/dc66e0996474b324f10649618f424fc76ca60e28/README.md#inline-ignores)?

codebien: I researched for it and I missed it, thanks for sharing

olegbespalov (approved):

imiric (approved):
Thanks! Please change the base to `release/v0.44.1` before merging.

PR #1130, opened 2019-08-29
stats/cloud: stop sending cloud metrics when limit reached

cuonglm:
Once the cloud test gets aborted by limit, we should not send metrics
anymore. The test and collector will still running, though.

Close #1074

PR #3061, opened 2023-05-08
cloud: Use static reference ID in tests

codebien:
The master's version mocks the test creation for all the tests doing them dependent on the resolution logic. The current PR removes this requirement and centralizes this assertion in a single test. Instead, all the other tests use a static-injected reference ID.

It should simplify the migration to something like the following architecture that I would do in https://github.com/grafana/k6/pull/3041:

output/cloud/output.go: has the logic for the test creation and cleaning up (common logic)
- output/cloud/v1/output.go has the **old** logic for metric collect/aggregate/flush
- output/cloud/expv2/output.go has the **new** logic for metric collect/aggregate/flush

Applying the current PR would be easier to eventually divide the unit tests of the test creation from the unit tests for metrics collection/aggregation/flushing logic across different components in the next PRs.

### DIFF
--- a/cloudapi/api_test.go
+++ b/cloudapi/api_test.go
@@ -24,18 +24,23 @@ func fprintf(t *testing.T, w io.Writer, format string, a ...interface{}) int {
 func TestCreateTestRun(t *testing.T) {
 	t.Parallel()
 	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		b, err := io.ReadAll(r.Body)
+		require.NoError(t, err)
+
+		exp := `{"name":"test","vus":0,"thresholds":null,"duration":0}`
+		assert.JSONEq(t, exp, string(b))
+
 		fprintf(t, w, `{"reference_id": "1", "config": {"aggregationPeriod": "2s"}}`)
 	}))
 	defer server.Close()
 
 	client := NewClient(testutils.NewLogger(t), "token", server.URL, "1.0", 1*time.Second)
 
-	tr := &TestRun{
+	resp, err := client.CreateTestRun(&TestRun{
 		Name: "test",
-	}
-	resp, err := client.CreateTestRun(tr)
+	})
+	assert.NoError(t, err)
 
-	assert.Nil(t, err)
 	assert.Equal(t, resp.ReferenceID, "1")
 	assert.NotNil(t, resp.ConfigOverride)
 	assert.True(t, resp.ConfigOverride.AggregationPeriod.Valid)
--- a/output/cloud/metrics_client.go
+++ b/output/cloud/metrics_client.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"compress/gzip"
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"net/http"
@@ -64,7 +65,11 @@ func (mc *MetricsClient) PushMetric(referenceID string, s []*Sample) error {
 	var additionalFields logrus.Fields
 
 	if !mc.noCompress {
-		buf := mc.pushBufferPool.Get().(*bytes.Buffer)
+		buf, ok := mc.pushBufferPool.Get().(*bytes.Buffer)
+		if !ok {
+			return errors.New("failed to convert a buffer pool item " +
+				"into the expected type bytes Buffer for gzip compression operation")
+		}
 		buf.Reset()
 		defer mc.pushBufferPool.Put(buf)
 		unzippedSize := len(b)
--- a/output/cloud/output_test.go
+++ b/output/cloud/output_test.go
@@ -139,7 +139,11 @@ func TestCloudOutput(t *testing.T) {
 	}
 
 	for tcNum, minSamples := range []int{60, 75, 100} {
-		t.Run(fmt.Sprintf("tc%d_minSamples%d", tcNum, minSamples), getTestRunner(minSamples))
+		tcNum, minSamples := tcNum, minSamples
+		t.Run(fmt.Sprintf("tc%d_minSamples%d", tcNum, minSamples), func(t *testing.T) {
+			t.Parallel()
+			getTestRunner(minSamples)
+		})
 	}
 }
 
@@ -223,7 +227,7 @@ func runCloudOutputTestCase(t *testing.T, minSamples int) {
 	smallSkew := 0.02
 
 	trails := []metrics.SampleContainer{}
-	durations := make([]time.Duration, len(trails))
+	durations := make([]time.Duration, 0, len(trails))
 	for i := int64(0); i < out.config.AggregationMinSamples.Int64; i++ {
 		similarTrail := skewTrail(r, simpleTrail, 1.0, 1.0+smallSkew)
 		trails = append(trails, &similarTrail)
@@ -322,6 +326,7 @@ func TestCloudOutputMaxPerPacket(t *testing.T) {
 	for j := time.Duration(1); j <= 200; j++ {
 		container := make([]metrics.SampleContainer, 0, 500)
 		for i := time.Duration(1); i <= 50; i++ {
+			//nolint:durationcheck
 			container = append(container, &httpext.Trail{
 				Blocked:        i % 200 * 100 * time.Millisecond,
 				Connecting:     i % 200 * 200 * time.Millisecond,
@@ -429,6 +434,7 @@ func testCloudOutputStopSendingMetric(t *testing.T, stopOnError bool) {
 	for j := time.Duration(1); j <= 200; j++ {
 		container := make([]metrics.SampleContainer, 0, 500)
 		for i := time.Duration(1); i <= 50; i++ {
+			//nolint:durationcheck
 			container = append(container, &httpext.Trail{
 				Blocked:        i % 200 * 100 * time.Millisecond,
 				Connecting:     i % 200 * 200 * time.Millisecond,
@@ -681,6 +687,7 @@ func TestCloudOutputRecvIterLIAllIterations(t *testing.T) {
 
 func TestNewName(t *testing.T) {
 	t.Parallel()
+
 	mustParse := func(u string) *url.URL {
 		result, err := url.Parse(u)
 		require.NoError(t, err)
@@ -715,6 +722,7 @@ func TestNewName(t *testing.T) {
 		testCase := testCase
 
 		t.Run(testCase.url.String(), func(t *testing.T) {
+			t.Parallel()
 			out, err := newOutput(output.Params{
 				Logger: testutils.NewLogger(t),
 				ScriptOptions: lib.Options{
@@ -730,6 +738,7 @@ func TestNewName(t *testing.T) {
 }
 
 func TestPublishMetric(t *testing.T) {
+	t.Parallel()
 	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		g, err := gzip.NewReader(r.Body)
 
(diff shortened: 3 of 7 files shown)


output/cloud/output_test.go
```diff
@@ -396,28 +361,6 @@ func testCloudOutputStopSendingMetric(t *testing.T, stopOnError bool) {
 	builtinMetrics := metrics.RegisterBuiltinMetrics(metrics.NewRegistry())
 
 	tb := httpmultibin.NewHTTPMultiBin(t)
-	tb.Mux.HandleFunc("/v1/tests", http.HandlerFunc(func(resp http.ResponseWriter, req *http.Request) {
-		body, err := io.ReadAll(req.Body)
-		require.NoError(t, err)
-		data := &cloudapi.TestRun{}
-		err = json.Unmarshal(body, &data)
-		require.NoError(t, err)
-		assert.Equal(t, "my-custom-name", data.Name)
```
na--, line 405: I don't think this test name logic or the validity of the JSON is checked anywhere now, you should probably add it to `TestOutputStartStopTestCreationWithConfigOverwrite`.

output/cloud/output_test.go
```diff
@@ -148,21 +149,8 @@ func runCloudOutputTestCase(t *testing.T, minSamples int) {
 	t.Logf("Random source seeded with %d\n", seed)
 
 	tb := httpmultibin.NewHTTPMultiBin(t)
-	tb.Mux.HandleFunc("/v1/tests", http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
na--, line 151: Considering this test seems to very precisely test just `http_req_li_all`, which is something very specific to the v1 cloud output, do you actually need to make any changes to it? It won't be useful at all for v2, so can't it just directly be moved to v1 as it is?

output/cloud/output_test.go
```diff
@@ -545,21 +488,12 @@ func TestCloudOutputRequireScriptName(t *testing.T) {
 
 func TestCloudOutputAggregationPeriodZeroNoBlock(t *testing.T) {
```
na--, line 489: I am actually struggling to understand how this test was useful and what it checked even before your changes... :sweat_smile: It originated from https://github.com/grafana/k6/pull/1130, but I think it got progressively mangled with subsequent refactorings :confused: With your latest changes, I am not sure it test anything meaningful beyond the fact that we can start and stop the output, which should already be covered by other tests :confused:

na-- (commented):
LGTM in general, though I think we can probably just move a lot of the current tests to v1/ as they are. A bunch of the tests are either not very high quality or are very specific to v1/ logic, so they can just be moved together with it and they can get discarded together with the v1/ logic.

output/cloud/output_test.go
codebien: The idea is to move in v1 and v2 only the logic for aggregate/flush (the logic not common) and to keep centralized the start/stop operations. To do it all the tests asserting specific logic will be moved to the appropriate package. So `v1`/`v2` will be not able to resolve the reference ID, they need it injected.

For this reason, we need to not invoke the test creation from all the tests that we plan to move to the specific versioned package.

output/cloud/output_test.go
codebien: The integration is aready covered by e2e tests, instead the unit I think it is more appropriate to be done in the `cloudapi` pkg. Done in https://github.com/grafana/k6/pull/3061/commits/97f22c76cc81d61dc9db0a0d5bd27fa62070fbb6

codebien:
I also applied the fixes for the linter, otherwise, we must fix them when we move the tests. It should be easier to review them here.

codebien:
I also removed `opts lib.Options` field because it seems not used at all. Let me know if it is unexpected and needs to cover with a dedicated unit test.

cloudapi/api_test.go
```diff
@@ -24,18 +24,23 @@ func fprintf(t *testing.T, w io.Writer, format string, a ...interface{}) int {
 func TestCreateTestRun(t *testing.T) {
 	t.Parallel()
 	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		exp := `{"name":"test","vus":0,"thresholds":null,"duration":0}`
+		b, err := io.ReadAll(r.Body)
+		if err != nil || string(b) != exp {
+			http.Error(w, "unexpected request body", http.StatusInternalServerError)
+			return
+		}
```
na--, line 32: hmm I think the order of marshaling fields in JSON it's not guaranteed, so something like [this](https://pkg.go.dev/github.com/stretchr/testify/assert#JSONEq) is probably better here:
```suggestion
		require.NoError(t, err)
		assert.JSONEq(t, exp, string(b))
```

output/cloud/output_test.go
na--: Hmm I don't see where in the e2e tests we are checking if setting `options.ext.loadimpact.name` to something like `my-custom-name` will result in the backend getting that info in the `POST /v1/tests` API request :confused: I see these 2 places and the default handler:
https://github.com/grafana/k6/blob/9e3359b7dd4b6dec7440d9e2cac6948b3ed64e4a/cmd/tests/cmd_run_test.go#L1687-L1702C6
https://github.com/grafana/k6/blob/9e3359b7dd4b6dec7440d9e2cac6948b3ed64e4a/cmd/tests/cmd_run_test.go#L1662-L1667
https://github.com/grafana/k6/blob/9e3359b7dd4b6dec7440d9e2cac6948b3ed64e4a/cmd/tests/cmd_cloud_test.go#L18-L24

But we are not even checking if we are getting valid JSON there :disappointed: If you add that check to an e2e test or two (or point me to an existing one I missed), then :+1: :+1: for getting it out of here. It will be much better suited to an e2e test indeed.

output/cloud/output_test.go
na--: Yeah, but AggregationPeriod is not set by default, so presumably this is already covered by a bunch of tests, that's why I thought we can just get rid of it :sweat_smile: anyway, it will go away soon, so it doesn't really matter.

output/cloud/output_test.go
codebien: >Hmm I don't see where in the e2e tests we are checking if setting options.ext.loadimpact.name

Sorry, I misunderstood this part from the previous comment. Should be done now.

Edited the current e2e for adding the coverage https://github.com/grafana/k6/pull/3061/commits/2cffc10c88ef093166c0b9978d437b7257a7c1de

na-- (approved):

output/cloud/output_test.go
```diff
@@ -449,30 +398,30 @@ func testCloudOutputStopSendingMetric(t *testing.T, stopOnError bool) {
...
+			type payload struct {
+				Error cloudapi.ErrorResponse `json:"error"`
 			}
-			body, err := io.ReadAll(r.Body)
-			assert.NoError(t, err)
-			receivedSamples := []Sample{}
-			assert.NoError(t, json.Unmarshal(body, &receivedSamples))
-		})
+			res := &payload{}
+			res.Error = cloudapi.ErrorResponse{Code: 4}
```
olegbespalov, line 408: If it's possible, let's maybe change this magic number `4` to anything more meaningful, if not it's okay since it's just formating

output/cloud/output_test.go
olegbespalov: I am joining the @na-- with missing the point of the original test and new additions that this PR is bringing to it.

output/cloud/output_test.go
```diff
@@ -868,28 +819,76 @@ func TestPublishMetric(t *testing.T) {
...
 
 	err = out.client.PushMetric("testmetric", nil)
-	assert.True(t, os.IsTimeout(err))
+
+	var nerr net.Error
+	require.True(t, errors.As(err, &nerr))
+	assert.True(t, nerr.Timeout())
+}
+
+func TestOutputStartStopTestCreationWithConfigOverwrite(t *testing.T) {
```
olegbespalov, line 850: I'm unsure why the test name contains start/stop.

Is it just something like: `TestOutputConfigOverwriting`?

output/cloud/output_test.go
```diff
@@ -868,28 +819,76 @@ func TestPublishMetric(t *testing.T) {
...
 			SystemTags: &metrics.DefaultSystemTagSet,
 		},
 		ScriptPath: &url.URL{Path: "script.js"},
 	})
 	require.NoError(t, err)
 
 	err = out.client.PushMetric("testmetric", nil)
-	assert.True(t, os.IsTimeout(err))
+
+	var nerr net.Error
```
olegbespalov, line 845: To be honest, just `os.IsTimeout` was much nicer

olegbespalov (commented):
Generally looks good, left a few comments :+1:

output/cloud/metrics_client.go
```diff
@@ -64,7 +65,10 @@ func (mc *MetricsClient) PushMetric(referenceID string, s []*Sample) error {
 	var additionalFields logrus.Fields
 
 	if !mc.noCompress {
-		buf := mc.pushBufferPool.Get().(*bytes.Buffer)
+		buf, ok := mc.pushBufferPool.Get().(*bytes.Buffer)
+		if !ok {
+			return errors.New("failed to convert a buffer pool item into the expected type bytes Buffer")
```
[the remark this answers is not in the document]
codebien: It can't happen but more important we are not going to use this metric_client in the v2, so I would not spend time on improving this.

I did it only to fix the linter otherwise we will get an error in the new PR moving it to a new file under `v1` pkg.

output/cloud/output_test.go
codebien: Good catch, it is a typo. The correct name should be:
```suggestion
func TestOutputTestCreationWithConfigOverwrite(t *testing.T) {
```

output/cloud/output_test.go
codebien: I don't know, I see we have this specific logic here: https://github.com/grafana/k6/blob/9e3359b7dd4b6dec7440d9e2cac6948b3ed64e4a/output/cloud/output.go#L582-L584

@na-- do you now more about it?

output/cloud/metrics_client.go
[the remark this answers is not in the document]
codebien: Anyway, I applied a small change.

output/cloud/output_test.go
na--: IIRC, this is a specific error that the backend return when they won't accept any more metrics. So, when k6 sees that, the cloud output just stops prematurely. The actual test execution continues, since for local `k6 run` tests the end-of-test summary (or any other outputs) will still work, but the `cloud` output doesn't send any more metrics. And, from what I can see, `K6_CLOUD_STOP_ON_ERROR` is enabled, the cloud output will stop not just itself, but the whole test run too.

output/cloud/output_test.go
codebien: Removed 24dfc4d64839a16785d92d53207e14968040187b

output/cloud/metrics_client.go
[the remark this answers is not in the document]
codebien: Done, check 8dba987439d8bac8312d17a0f561cb566234399b

output/cloud/output_test.go
codebien: Added a comment with mostly what Ned said. Check c2efd34771314ec1235c18ec8be163f8eae9d8b1

olegbespalov (approved):

na-- (approved):

PR #3063, opened 2023-05-10
cloud/v1: Aggregation in a dedicated pkg

codebien:
It splits the current output in a subpackage `v1`, it moves there the specific parts for version 1 and that are not expected to be used from version 2.
It uses part of the work done in #3041, and it is a foundation for adding, on top of it, the versioning config option (https://github.com/grafana/k6/pull/3041) and the `expv2` package (https://github.com/grafana/k6/pull/2963).

The proposed architecture should resolve the requested change in https://github.com/grafana/k6/pull/2963#discussion_r1157118508.

I think you should review the code going through the commits one by one to check the evolution, otherwise, the final code could be too chaotic.

I think the test coverage is not high enough for the new `output/cloud/output.go` to have good confidence. If the architecture is confirmed I will add a bunch of them in a dedicated PR (we can keep this open till it is approved).


codebien:
Rebased to resolve the conflicts, green now

na-- (dismissed):

output/cloud/output.go
```diff
@@ -269,25 +227,48 @@ func (out *Output) Stop() error {
...
-	close(out.stopAggregation)
-	out.aggregationDone.Wait() // could be a no-op, if we have never started the aggregation
-	out.logger.Debug("Aggregation stopped, stopping metric emission...")
-	close(out.stopOutput)
-	out.outputDone.Wait()
+	err := out.versionedOutput.StopWithTestError(testErr)
+	if err != nil {
+		out.logger.WithError(err).Error("An error occurred stopping the output")
+		// it doesn't return here because
+		// it is required to notify the Cloud backend before return
```
olegbespalov, line 234:
```suggestion
		// to notify the cloud backend we have no return here
```

output/cloud/output.go
```diff
@@ -339,373 +320,18 @@ func (out *Output) getRunStatus(testErr error) cloudapi.RunStatus {
...
-	select {
-	case <-out.stopSendingMetrics:
-		return
-	default:
-	}
-
+func (out *Output) startVersionedOutput() error {
 	if out.referenceID == "" {
-		return
+		return fmt.Errorf("ReferenceID is required")
```
olegbespalov, line 325:
```suggestion
		return errors.New("ReferenceID is required")
```

olegbespalov (dismissed):
LGTM, with a few minor non-blocking comments :+1:

olegbespalov (approved):

PR #3064, opened 2023-05-11
Update deps

mstoykov:


mstoykov:
grpc is not updated as there are problems - issue pending
prometheus is not updated - to be updated with the extension
browser defependancy also to be updated with browser update

vendor/github.com/dlclark/regexp2/README.md
```diff
@@ -92,6 +92,66 @@ if isMatch, _ := re.MatchString(`Something to match`); isMatch {
...
+
+`regexp2` supports features that can lead to catastrophic backtracking.
+`Regexp.MatchTimeout` can be set to to limit the impact of such behavior; the
+match will fail with an error after approximately MatchTimeout. No timeout
+checks are done by default.
+
+Timeout checking is not free. The current timeout checking implementation starts
+a background worker that updates a clock value approximately once every 100
+milliseconds. The matching code compares this value against the precomputed
+deadline for the match. The performance impact is as follows.
```
imiric, line 105: This is an indirect dependency used by goja, and `MatchTimeout` is not used there. I guess this is up to Dmitry to decide whether it makes sense, but it could have an impact for us if he decides to enable it.

vendor/github.com/dlclark/regexp2/README.md
```diff
@@ -92,6 +92,66 @@ if isMatch, _ := re.MatchString(`Something to match`); isMatch {
...
+    until the longest deadline (match timeout + the time when the match started)
+    is reached. E.g., if you set a timeout of one minute the load will persist
+    for approximately a minute even if the match finishes quickly.
+
+See [PR #58](https://github.com/dlclark/regexp2/pull/58) for more details and 
+alternatives considered.
+
+## Goroutine leak error
+If you're using a library during unit tests (e.g. https://github.com/uber-go/goleak) that validates all goroutines are exited then you'll likely get an error if you or any of your dependencies use regex's with a MatchTimeout. 
+To remedy the problem you'll need to tell the unit test to wait until the backgroup timeout goroutine is exited.
```
imiric, line 121: Also relevant for us if `MatchTimeout` is enabled in goja, since we do use `goleak`.

vendor/golang.org/x/net/http2/transport.go
```diff
@@ -1265,6 +1266,27 @@ func (cc *ClientConn) RoundTrip(req *http.Request) (*http.Response, error) {
...
+			// the request without waiting for a response.) Mark the connection as
+			// not reusable, since trying to reuse a dead connection is worse than
+			// unnecessarily creating a new one.
+			//
+			// If cs.ID is 0, then the request was never allocated a stream ID and
+			// whatever went wrong was unrelated to the connection. We might have
+			// timed out waiting for a stream slot when StrictMaxConcurrentStreams
+			// is set, for example, in which case retrying on a different connection
+			// will not help.
+			cs.cc.doNotReuse = true
```
imiric, line 1285: This seems like a good thing, and probably safe for us(?) :crossed_fingers:

imiric (approved):
Seems safe for the most part :crossed_fingers:

vendor/github.com/fatih/color/color_windows.go
```diff
@@ -0,0 +1,19 @@
...
+func init() {
+	// Opt-in for ansi color support for current process.
+	// https://learn.microsoft.com/en-us/windows/console/console-virtual-terminal-sequences#output-sequences
+	var outMode uint32
+	out := windows.Handle(os.Stdout.Fd())
+	if err := windows.GetConsoleMode(out, &outMode); err != nil {
+		return
+	}
+	outMode |= windows.ENABLE_PROCESSED_OUTPUT | windows.ENABLE_VIRTUAL_TERMINAL_PROCESSING
+	_ = windows.SetConsoleMode(out, outMode)
```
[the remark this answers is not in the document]
mstoykov: > Even if this seems like a safe change, we should still test the behavior on Windows to ensure our UI hacks continue to work.

I did test both powershell and cmd and it looks exactly the same as before :shrug: .

cmd still does not show unicode and powershell seems as okay as I can expect.

The only discrepancy is that with both old and new version hitting `enter` while a test is running - does nothing :(. 

But that seems like completely unrelated issue and also not really a problem as we do not care about anytnig but `^C` :crossed_fingers:

vendor/github.com/dlclark/regexp2/README.md
mstoykov: Yes - it is, but goja does not update it as regularly as I would prefer and we already have updated in the past to get some *unlikely* fixes.

This current update does not matter to use but removes the need to review this change in the future.

vendor/github.com/dlclark/regexp2/README.md
mstoykov: it isn't

vendor/golang.org/x/net/http2/transport.go
mstoykov: I am hopeful that this will fix some EOF errors for some users. Also this is very likely to already have landed in stdlib as it is just x/net vendored.

vendor/github.com/fatih/color/color_windows.go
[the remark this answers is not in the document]
mstoykov: From https://github.com/fatih/color/issues/169#issuecomment-1335995045 - it is apparently about other terminals that do the more ... accurate handling of colors - where they are not supported unless the application tells them they want them. Which apparently is what this code does. 

I unfortunately do not have a terminal that has his problem :(.

oleiade (approved):

go.mod
```diff
@@ -8,23 +8,23 @@ require (
...
-	github.com/dop251/goja v0.0.0-20230402114112-623f9dda9079
-	github.com/fatih/color v1.14.1
+	github.com/dop251/goja v0.0.0-20230427124612-428fc442ff5f
+	github.com/fatih/color v1.15.0
 	github.com/go-sourcemap/sourcemap v2.1.4-0.20211119122758-180fcef48034+incompatible
 	github.com/golang/protobuf v1.5.2
 	github.com/gorilla/websocket v1.5.0
 	github.com/grafana/xk6-browser v0.9.0
-	github.com/grafana/xk6-output-prometheus-remote v0.1.0
+	github.com/grafana/xk6-output-prometheus-remote v0.2.0
```
codebien, line 17: https://github.com/grafana/xk6-output-prometheus-remote/releases/tag/v0.2.1

Issue #3065, opened 2023-05-11
Restrict metric names to what is supported by Prometheus


mstoykov:
### What?

k6 currently supports very liberal metric names 
https://github.com/grafana/k6/blob/master/metrics/registry.go#L29
prometheus does not https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels

Open telemetry also has way more restricted list as well https://opentelemetry.io/docs/specification/otel/metrics/api/#instrument-name-syntax

### Proposed solution:

Limit to `[a-zA-Z_][a-zA-Z0-9_]*` which is Prometheus without `:` which have special meaning. We can always expand them again later.

In order to have as little disturbance as possible we will have some release (3) with warning, and then we will make this into an error.

### Caveats:

Depending on community feedback we might pivot more in one or another direction.

pkouzmineESS:
Suggestion: instead of completely eliminating ability to create metric names with a wider character set, why not put it behind a flag? By default, K6 could check that metrics are created only using [a-zA-Z_][a-zA-Z0-9_]*. But if launched with flag like --expanded-metric-name-charset, current behavior could be retained. This has a benefit of making it less of a breaking change.

mstoykov:
Hi @pkouzmineESS, we could add option to make it or not, but I would prefer not to have functionality that people have not requested. 

This option basically will make it impossible for us to optimize anything based on these restrictions. And will make any functionality (outputs to otel/prometheus compatible engines) to either not work or have strange name transformations.

So the question is whether anyone actually uses this more extended names and if yes - was that on purpose or did they just write a name and as it didn't tell them anything they went with it?

Do you have a case where you want to more expanded metric name charset?

pkouzmineESS:
In our case we test a little over 150 endpoints in over 50 scenarios. Some of our endpoints trigger differently optimized stored procedures on the database side depending on what query parameters are passed. In other cases, we have endpoints that do different things depending on the payload passed in and have different performance profiles based on the same (not the best design, I know, but what we have to live with for now). Further, different scenarios might call the same endpoint and sometimes we want to keep track of the originating scenario. For example, we might need to call search function as if it was used by individual user and also call it in different scenario as if it was called by user from a support center.

[...]

PR #3066, opened 2023-05-11
k6/metrics: Warn on metric names not compatible with Prometheus

mstoykov:
This adds a warning for custom metrics that are not Prometheus compatible.

The warning is in the k6/metrics as the alternative was threading a logger in the registry specifically for this and then removing it.

The major downside of this is that extension can still register incompatible metric names without any warnings.

Updates #3065

### DIFF
--- a/js/bundle.go
+++ b/js/bundle.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"net/url"
 	"path/filepath"
+	"regexp"
 	"runtime"
 
 	"github.com/dop251/goja"
@@ -37,6 +38,24 @@ type Bundle struct {
 
 	callableExports map[string]struct{}
 	moduleResolver  *moduleResolver
+}
+
+// TODO: this is to be removed once this is not a warning and it can be moved to the registry
+// https://github.com/grafana/k6/issues/3065
+func (b *Bundle) checkMetricNamesForPrometheusCompatibility() {
+	const (
+		nameRegexString = "^[a-zA-Z_][a-zA-Z0-9_]{1,63}$"
+		badNameWarning  = "Metric name should only include ASCII letters, numbers and underscores. " +
+			"This name will stop working in k6 v0.48.0 (around December 2023)."
+	)
+
+	compileNameRegex := regexp.MustCompile(nameRegexString)
+
+	for _, metric := range b.preInitState.Registry.All() {
+		if !compileNameRegex.MatchString(metric.Name) {
+			b.preInitState.Logger.WithField("name", metric.Name).Warn(badNameWarning)
+		}
+	}
 }
 
 // A BundleInstance is a self-contained instance of a Bundle.
@@ -106,6 +125,8 @@ func newBundle(
 	if err != nil {
 		return nil, err
 	}
+
+	bundle.checkMetricNamesForPrometheusCompatibility()
 
 	return bundle, nil
 }
--- a/cmd/tests/cmd_run_test.go
+++ b/cmd/tests/cmd_run_test.go
@@ -1543,6 +1543,44 @@ func TestMinIterationDuration(t *testing.T) {
 	assert.Contains(t, stdout, "✓ test_counter.........: 3")
 }
 
+func TestMetricNameWarning(t *testing.T) {
+	t.Parallel()
+	script := `
+		import { Counter } from 'k6/metrics';
+
+		export let options = {
+			vus: 2,
+			iterations: 2,
+			thresholds: {
+				'test counter': ['count == 4'],
+			},
+		};
+
+		var c = new Counter('test counter');
+		new Counter('test_counter_#');
+
+		export function setup() { c.add(1); };
+		export default function () { c.add(1); };
+		export function teardown() { c.add(1); };
+	`
+
+	ts := getSimpleCloudOutputTestState(t, script, nil, cloudapi.RunStatusFinished, cloudapi.ResultStatusPassed, 0)
+
+	cmd.ExecuteWithGlobalState(ts.GlobalState)
+
+	stdout := ts.Stdout.String()
+	t.Log(stdout)
+
+	logEntries := ts.LoggerHook.Drain()
+	expectedMsg := `Metric name should only include ASCII letters, numbers and underscores. This name will stop working in `
+	filteredEntries := testutils.FilterEntries(logEntries, logrus.WarnLevel, expectedMsg)
+	require.Len(t, filteredEntries, 2)
+	// we do it this way as ordering is not guaranteed
+	names := []interface{}{filteredEntries[0].Data["name"], filteredEntries[1].Data["name"]}
+	require.Contains(t, names, "test counter")
+	require.Contains(t, names, "test_counter_#")
+}
+
 func TestRunTags(t *testing.T) {
 	t.Parallel()
 
--- a/lib/testutils/logrus_hook.go
+++ b/lib/testutils/logrus_hook.go
@@ -87,3 +87,15 @@ func LogContains(logEntries []logrus.Entry, expLevel logrus.Level, expContents s
 	}
 	return false
 }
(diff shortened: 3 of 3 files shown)


mstoykov:
While I was writing this I started wondering if an alternative approach of checking this after the initial init context run might not be more beneficial. 

The major problem is the registry can't currently return all the metric names, but it will let us catch extensions registering "bad" names

codebien:
>but it will let us catch extensions registering "bad" names.

Yeah, I see value into this. I prefer an approach that includes extensions so we don't create shadow zones.

>the registry can't currently return all the metric names

Do you mean something like this? https://github.com/grafana/k6/pull/2905/commits/b485b2296e4fa5277524230467e7c83556ee8d01

mstoykov:
@codebien yeah that seems like a thing I can use.

Is that PR likely to be merged soon :tm: . If not can we pull this function out and then rebase both of our PRs on top of it?

js/bundle.go
```diff
@@ -39,6 +40,23 @@ type Bundle struct {
 	moduleResolver  *moduleResolver
 }
 
+// TODO: this is to be removed once this is not a warning and it can be moved to the registry
+func (b *Bundle) checkMetricNamesForPrometheusCompatibility() {
+	const (
+		nameRegexString = "^[a-zA-Z_][a-zA-Z0-9_]{1,128}$"
+		badNameWarning  = "Metric name should only include ASCII letters, numbers and underscores. " +
+			"This name will stop working in k6 v0.49.0."
```
codebien, line 48:
```suggestion
			"This name will stop working in k6 v0.48.0 (around December 2023)."
```
It should clarify expectations. TBH, I think it is still too relaxed.

codebien (commented):
LGTM, just a minor comment

codebien (approved):
LGTM

js/bundle.go
```diff
@@ -39,6 +40,23 @@ type Bundle struct {
 	moduleResolver  *moduleResolver
 }
 
+// TODO: this is to be removed once this is not a warning and it can be moved to the registry
+func (b *Bundle) checkMetricNamesForPrometheusCompatibility() {
+	const (
+		nameRegexString = "^[a-zA-Z_][a-zA-Z0-9_]{1,128}$"
```
imiric, line 46: I didn't see a length limit on Prometheus' metric names in their documentation, but [OTel limits it to 63 chars](https://opentelemetry.io/docs/specs/otel/metrics/api/#instrument-name-syntax).

I'm not sure if we should follow whatever Prometheus' limit is, or OTel's, but 63 chars is way too restricting. So... neither?

js/bundle.go
```diff
@@ -39,6 +40,23 @@ type Bundle struct {
...
+		badNameWarning  = "Metric name should only include ASCII letters, numbers and underscores. " +
+			"This name will stop working in k6 v0.48.0 (around December 2023)."
+	)
+
+	compileNameRegex := regexp.MustCompile(nameRegexString)
```
imiric, line 51: We should compile this regex only once. Currently it's done for every new bundle instance.

imiric (changes requested):
Compiling the regex outside of `newBundle` is the only blocker for me.

But I have a general concern about this change. Shouldn't we only limit metric names if the Prometheus or OTel outputs are used? That is, wouldn't it make more sense to have this warning and eventual failure in the outputs themselves? It seems unnecessary to break metric names for all k6 users, even if they're not using Prometheus/OTel.

If we want to suggest people stick to this restriction anyway, then just a warning would be enough, without the breaking change in v0.48.0. But even a general warning would be annoying if someone is not using Prometheus/OTel.

js/bundle.go
mstoykov: I can move it - but in a real k6 instance this happens *once*. Which is also why I chose to move it here specifically as I know this is only called once, so the logs will happen only once and I don't need to do anykind of additional work.

I am okay with moving outside, it will slightly speed up *some* tests :shrug: . My main reason is that in 6 months we will be able to just delete this, whcih seemed slightly cleaner to me

mstoykov:
> Compiling the regex outside of newBundle is the only blocker for me.

I will change that if my comment there is not enough - I really have no problem with that.

> But I have a general concern about this change. Shouldn't we only limit metric names if the Prometheus or OTel outputs are used? That is, wouldn't it make more sense to have this warning and eventual failure in the outputs themselves? It seems unnecessary to break metric names for all k6 users, even if they're not using Prometheus/OTel.

> If we want to suggest people stick to this restriction anyway, then just a warning would be enough, without the breaking change in v0.48.0. But even a general warning would be annoying if someone is not using Prometheus/OTel.

I feel like I have not explained the change very well. Most systems working with metrics have some metric name restrictions. Of all of the Prometheus and Otel seem to be the one that get the most traction. 

Anyone who uses k6 and custom metrics and output to one of those systems will have some problems if k6 allows characters that are not supported by the other system.

AFAIK the current k6 name restriction is mostly what it is because that is what we came up with. 

So this PR tries to make it so that k6 metric names will be supported by any output system irregardless of whether the users use them or not.

As the issues says we will pivot in case there is a lot of user feedback that they want spaces (for example, also the character that started this).

In practice this warning is a way to ask users for feedback in case they really want to keep something, until we actually deprecate it. 

cc @dgzlopes if you want to add more.

js/bundle.go
imiric: Ah, my bad. I had forgotten `newBundle` is indeed only called once. Nevermind, this is fine then.

js/bundle.go
imiric: True, 63 chars ought to be enough for anybody.:tm: :smile:

It's probably a safe limit to impose. Even the browser's long Web Vital metrics like `webvital_first_contentful_paint_good` is 36 chars, and going much above that would be messy to show in our CLI and Cloud UIs.

So if we want to be strict with OTel compatibility, we might also limit it to 63.

dgzlopes:
> cc @dgzlopes if you want to add more.
I think you covered all 👍

imiric:
@mstoykov Yeah, I see the benefit of restricting this to some common set of characters that would be compatible with most metric backends. I was just wondering whether it made sense to force the change on users who don't use outputs that would be impacted by this.

But OK, no blockers from me. Let's just agree on the length.

mstoykov:
@codebien I think the message is already long enough. And if users see a message that tells them that something that they need will not work in 2-3 versions, I would expect that:
1. they will change it if it is easy for them, and they do not care - WIN
2. actually come to tell us as in 2-3 versions they will not be able to use us.

We have been getting reports of this kind for a while without a big message, so I prefer to not make the message even longer.

sniku:
Eventually, we will store k6-cloud time-series data in Mimir, so we should enforce the metric name restrictions for all k6 tests (not only for those that use Prometheus output).

olegbespalov (approved):

imiric (approved):

js/bundle.go
```diff
@@ -39,6 +40,23 @@ type Bundle struct {
 	moduleResolver  *moduleResolver
 }
 
+// TODO: this is to be removed once this is not a warning and it can be moved to the registry
```
codebien, line 43: Can we open the issue and the milestone and link it here, please? So we will have the proper removal as planned, otherwise, we will forget about it.

mstoykov: The issue is https://github.com/grafana/k6/issues/3065 - I did create teh milestone though. 

I don't really see the point in linking it though

codebien: >I don't really see the point in linking it though

The point is if someone hits the TODO and they do not remember about the issue (like me in this case) or is not aware of it at all, it can quickly get the context if they need it. Just better UX for us (the devs).

olegbespalov (approved):

codebien (approved):

PR #3071, opened 2023-05-15
output/cloudv2: Aggregation

codebien:
It mainly implements the same logic as for `v1` but it extends the logic to all the generated time series instead to just limit it to HTTP metrics.

It applies basic separation of concerns, so it should be easier to improve independently the various components in the future.

The flusher is not implemented but it should be mostly a wrap around the client implemented in #2963 draining the `bucketQ` buffer. I would like to get a consensus on this PR before implementing it.

The placeholder for the Trend sink will be implemented by https://github.com/grafana/k6/pull/3027

### DIFF
--- a/output/cloud/expv2/collect.go
+++ b/output/cloud/expv2/collect.go
@@ -1,0 +1,168 @@
+package expv2
+
+import (
+	"errors"
+	"sync"
+	"time"
+
+	"go.k6.io/k6/metrics"
+)
+
+type timeBucket struct {
+	Time  time.Time
+	Sinks map[metrics.TimeSeries]metrics.Sink
+}
+
+// bucketQ is a queue for buffering the aggregated metrics
+// that have to be flushed. It is expected to be used concurrently.
+type bucketQ struct {
+	m       sync.Mutex
+	buckets []timeBucket
+}
+
+// PopAll returns a slice with all the pushed buckets.
+// It returns a nil slice if the queue is empty.
+func (q *bucketQ) PopAll() []timeBucket {
+	q.m.Lock()
+	defer q.m.Unlock()
+
+	if len(q.buckets) < 1 {
+		return nil
+	}
+
+	// return the enqueued slice and its relative array and allocate a new one
+	// using the same capacity.
+	b := q.buckets
+	q.buckets = make([]timeBucket, 0, len(b))
+	return b
+}
+
+// Push enqueues a new item in the queue.
+func (q *bucketQ) Push(b []timeBucket) {
+	if len(b) < 1 {
+		return
+	}
+	q.m.Lock()
+	q.buckets = append(q.buckets, b...)
+	q.m.Unlock()
+}
+
+type collector struct {
+	bq      bucketQ
+	nowFunc func() time.Time
+
+	aggregationPeriod time.Duration
+	waitPeriod        time.Duration
+
+	// We should no longer have to handle metrics that have times long in the past.
+	// So, instead of a map, we can probably use a simple slice (or even an array!)
+	// as a ring buffer to store the aggregation buckets.
+	// This should save us a some time, since it would make the lookups and
+	// WaitPeriod checks basically O(1).
+	// And even if for some reason there are occasional metrics with past times
+	// that don't fit in the chosen ring buffer size, we could just send them
+	// along to the buffer unaggregated.
+	timeBuckets map[int64]map[metrics.TimeSeries]metrics.Sink
+}
+
+func newCollector(aggrPeriod, waitPeriod time.Duration) (*collector, error) {
+	if aggrPeriod == 0 {
+		return nil, errors.New("aggregation period is not allowed to be zero")
+	}
+	if waitPeriod == 0 {
+		// TODO: we could simplify the expiring logic
+		// just having an internal static logic.
+		// Like skip only not closed buckets bucketEnd > now.
+		return nil, errors.New("aggregation wait period is not allowed to be zero")
+	}
+	return &collector{
+		bq:                bucketQ{},
+		nowFunc:           time.Now,
+		timeBuckets:       make(map[int64]map[metrics.TimeSeries]metrics.Sink),
+		aggregationPeriod: aggrPeriod,
+		waitPeriod:        waitPeriod,
+	}, nil
+}
+
+// CollectSamples drain the buffer and collect all the samples.
+func (c *collector) CollectSamples(containers []metrics.SampleContainer) {
+	// Distribute all newly buffered samples into related buckets
+	for _, sampleContainer := range containers {
+		samples := sampleContainer.GetSamples()
+
+		for i := 0; i < len(samples); i++ {
+			c.collectSample(samples[i])
+		}
+	}
+	c.bq.Push(c.expiredBuckets())
+}
+
+// DropExpiringDelay drops the waiting time for buckets
+// for the expiring checks.
+func (c *collector) DropExpiringDelay() {
+	c.waitPeriod = 0
+}
+
+func (c *collector) collectSample(s metrics.Sample) {
+	bucketID := c.bucketID(s.Time)
+
+	// Get or create a time bucket
+	bucket, ok := c.timeBuckets[bucketID]
+	if !ok {
+		bucket = make(map[metrics.TimeSeries]metrics.Sink)
+		c.timeBuckets[bucketID] = bucket
+	}
+
+	// Get or create the bucket's sinks map per time series
+	sink, ok := bucket[s.TimeSeries]
+	if !ok {
+		sink = newSink(s.Metric.Type)
+		bucket[s.TimeSeries] = sink
+	}
+
+	// TODO: we may consider to just pass
+	// the single value instead of the entire
+	// sample and save some memory
+	sink.Add(s)
+}
+
+func (c *collector) expiredBuckets() []timeBucket {
+	// Still too recent buckets
+	// where we prefer to wait a bit more
+	// then, hopefully, we can aggregate more samples before flushing.
+	bucketCutoffID := c.bucketCutoffID()
+
+	// Here, it avoids pre-allocation
+	// because it expects to be zero for most of the time
+	var expired []timeBucket //nolint:prealloc
+
+	// Mark as expired all aggregation buckets older than bucketCutoffID
+	for bucketID, seriesSinks := range c.timeBuckets {
+		if bucketID > bucketCutoffID {
+			continue
+		}
+
+		expired = append(expired, timeBucket{
+			Time:  c.timeFromBucketID(bucketID),
+			Sinks: seriesSinks,
+		})
+		delete(c.timeBuckets, bucketID)
+	}
+
+	return expired
+}
+
+func (c *collector) bucketID(t time.Time) int64 {
+	return t.UnixNano() / int64(c.aggregationPeriod)
+}
+
+func (c *collector) timeFromBucketID(id int64) time.Time {
+	return time.Unix(0,
+		// it uses the center of the bucket as time
+		(id*int64(c.aggregationPeriod))+int64(c.aggregationPeriod/2),
+	).Truncate(time.Microsecond).UTC()
+}
+
+func (c *collector) bucketCutoffID() int64 {
+	return c.nowFunc().Add(-c.waitPeriod).UnixNano() / int64(c.aggregationPeriod)
+}
--- a/output/cloud/expv2/collect.go
+++ b/output/cloud/expv2/collect.go
@@ -9,6 +9,8 @@ import (
 )
 
 type timeBucket struct {
+	// TODO: for performance reasons, use directly a unix time here
+	// so we can avoid time->unix->time
 	Time  time.Time
 	Sinks map[metrics.TimeSeries]metrics.Sink
 }
@@ -157,10 +159,8 @@ func (c *collector) bucketID(t time.Time) int64 {
 }
 
 func (c *collector) timeFromBucketID(id int64) time.Time {
-	return time.Unix(0,
-		// it uses the center of the bucket as time
-		(id*int64(c.aggregationPeriod))+int64(c.aggregationPeriod/2),
-	).Truncate(time.Microsecond).UTC()
+	return time.Unix(0, id*int64(c.aggregationPeriod)).
+		Truncate(time.Microsecond).UTC()
 }
 
 func (c *collector) bucketCutoffID() int64 {
--- a/output/cloud/expv2/output.go
+++ b/output/cloud/expv2/output.go
@@ -3,25 +3,43 @@
 package expv2
 
 import (
+	"context"
+	"net/http"
+	"time"
+
 	"go.k6.io/k6/cloudapi"
+	"go.k6.io/k6/errext"
+	"go.k6.io/k6/errext/exitcodes"
 	"go.k6.io/k6/metrics"
+	"go.k6.io/k6/output"
 
 	"github.com/sirupsen/logrus"
 )
 
 // Output sends result data to the k6 Cloud service.
 type Output struct {
+	output.SampleBuffer
+
 	logger       logrus.FieldLogger
 	config       cloudapi.Config
 	referenceID  string
 	testStopFunc func(error)
+
+	// TODO: replace with the real impl
+	metricsFlusher  noopFlusher
+	periodicFlusher *output.PeriodicFlusher
+
+	collector             *collector
+	periodicCollector     *output.PeriodicFlusher
+	stopMetricsCollection chan struct{}
 }
 
 // New creates a new cloud output.
 func New(logger logrus.FieldLogger, conf cloudapi.Config) (*Output, error) {
 	return &Output{
-		config: conf,
-		logger: logger.WithFields(logrus.Fields{"output": "cloudv2"}),
+		config:                conf,
+		logger:                logger.WithFields(logrus.Fields{"output": "cloudv2"}),
+		stopMetricsCollection: make(chan struct{}),
 	}, nil
 }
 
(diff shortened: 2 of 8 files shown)


output/cloud/expv2/collect.go
```diff
@@ -0,0 +1,153 @@
...
+type bucketQ struct {
+	m       sync.Mutex
+	buckets []timeBucket
+}
+
+// Buckets pop all the queued buckets.
+func (q *bucketQ) Buckets() []timeBucket {
+	q.m.Lock()
+	b := q.buckets
+	q.buckets = q.buckets[:0]
```
na--, line 21: I haven't grokked the reminder of this code yet, but this part seems problematic, since `b` and `q.buckets` will reference the same underlying array, and I can't think of a situation where this would be intended and we'd still need locking here.

codebien: https://github.com/grafana/k6/pull/3071/commits/5db6157071e6bcea7e7c3f822a4cbe9dbab50294 applied an improvement for it

output/cloud/expv2/collect.go
```diff
@@ -0,0 +1,158 @@
...
+// Insert adds an item in the queue.
+func (q *bucketQ) Push(b []timeBucket) {
+	q.m.Lock()
+	q.buckets = append(q.buckets, b...)
+	q.m.Unlock()
+}
+
+type timeBucket struct {
+	TimeSeries metrics.TimeSeries
+	Time       time.Time
```
mstoykov, line 40: I don't think we need those as this is information that already exists and this way it will be copied a bunch of times.

output/cloud/expv2/collect.go
```diff
@@ -0,0 +1,158 @@
...
+			continue
+		}
+
+		for _, bucket := range seriesBuckets {
+			// TODO: is there a valid math reason to have a minimum
+			// number of samples aggregated as we do in v1?
+			// I mean o.config.AggregationMinSamples.Int64
+
+			expired = append(expired, bucket)
+		}
```
mstoykov, line 150:
```suggestion
		expired = append(expired, seriesBuckets...)
```

output/cloud/expv2/collect.go
```diff
@@ -0,0 +1,158 @@
...
+func (c *collector) expiredBuckets() []timeBucket {
+	// Still new buckets where we have to wait to accumulate
+	// more samples before flushing
+	//
+	// TODO: Do we need it?
+	// if the Cloud handles the merge is not really required
+	bucketCutoffID := c.nowfn().Add(-c.waitPeriod).UnixNano() / int64(c.aggregationPeriod)
+
+	// TODO: multiply it at least for the number of metrics?
+	expired := make([]timeBucket, 0, len(c.aggrBuckets))
```
mstoykov, line 136: If you want you can first count them, but if you are not going to count them, I would recommend just leaving it as `nil`

The aggrBuckets will be some small number such usually `waitPeriod/aggregationPeriod` so this preallocation is likely not very useful.

My problem with precounting is that I expect 90% of the time there will be 0 buckets and every once in a while there will be 1 bucket and with the other proposed change below it will just allocate as much as it needs this one time.

If it ever happens that it goes through 2 buckets that will be the case where preallocating here will help.

output/cloud/expv2/collect.go
```diff
@@ -0,0 +1,158 @@
...
+	}
+}
+
+func (c *collector) expiredBuckets() []timeBucket {
+	// Still new buckets where we have to wait to accumulate
+	// more samples before flushing
+	//
+	// TODO: Do we need it?
+	// if the Cloud handles the merge is not really required
+	bucketCutoffID := c.nowfn().Add(-c.waitPeriod).UnixNano() / int64(c.aggregationPeriod)
```
mstoykov, line 133: I don't understand the question

output/cloud/expv2/output.go
```diff
@@ -52,13 +94,75 @@ func (o *Output) Start() error {
...
+		return false
+	}
+	if errResp, ok := err.(cloudapi.ErrorResponse); ok && errResp.Response != nil { //nolint:errorlint
+		// The Cloud service returns the error code 4 when it doesn't accept any more metrics.
+		// So, when k6 sees that, the cloud output just stops prematurely.
+		return errResp.Response.StatusCode == http.StatusForbidden && errResp.Code == 4
+	}
+
+	return false
 }
```
mstoykov, line 168: Maybe move that to cloudapi package?

output/cloud/expv2/collect.go
```diff
@@ -0,0 +1,158 @@
...
+	for bucketID, seriesBuckets := range c.aggrBuckets {
+		if bucketID > bucketCutoffID {
+			continue
+		}
+
+		for _, bucket := range seriesBuckets {
+			// TODO: is there a valid math reason to have a minimum
+			// number of samples aggregated as we do in v1?
+			// I mean o.config.AggregationMinSamples.Int64
+
```
mstoykov, line 148: Given that the aggregation happens as you add them - I don't think it is even possible :).

mstoykov (approved):
a :+1: from me with the caveat that I am not certain how well it will perform.

I have a bunch of comments, none of them are particularly blocking but also should be easy to fix :crossed_fingers: 

My main worry is taht adding samples now also aggregates, so adding samples is actually fairly expensive (or could be). But I would prefer to get it done and then wecan try to optimise it.

output/cloud/expv2/output.go
na--: Considering we want to remove the v1 cloud output implementation in the near future, this will only be used in v2, so no need to move it to a separate library IMO

output/cloud/expv2/collect.go
codebien: How do we pass them to the flush operation? I can drop them during the _aggregation_ phase because they are already in the _perSeriesBuckets_ so we can build this struct only before enqueueing. But we can't skip doing it before calling `bucketQ.Push`.
Am I missing something?

output/cloud/expv2/collect.go
```diff
@@ -0,0 +1,158 @@
...
+	"go.k6.io/k6/output"
+)
+
+type bucketQ struct {
+	m       sync.Mutex
+	buckets []timeBucket
+}
+
+// Buckets pop all the queued buckets.
+func (q *bucketQ) Buckets() []timeBucket {
```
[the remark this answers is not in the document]
codebien: What about `PopAll` as we do somewhere else (I guess in the eventloop :thinking: )?

output/cloud/expv2/collect.go
codebien: My comment is related to the `Add(-c.waitPeriod)` operation, the collector runs every X seconds so it already gives us some delay. Do we need more delay? Are samples flushed with some long delay with a relevant percentage that should justify this special handling?

output/cloud/expv2/collect.go
na--: Ah, I'd missed @codebien's comment when writing my own, so it's not a response to it but to @mstoykov's.

output/cloud/expv2/collect.go
codebien: I think we could count the number of samples per bucket. In any case, I don't think we have a valid use case to justify this overhead.

output/cloud/expv2/sink.go
```diff
@@ -0,0 +1,22 @@
...
+)
+
+func newSink(mt metrics.MetricType) metrics.Sink {
+	if mt == metrics.Trend {
+		return &histogram{}
+	}
+
+	return metrics.NewSink(mt)
+}
+
```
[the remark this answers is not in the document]
codebien: >IMO all but the rate sinks have more stuff in them then is necessary for this output.

The RateSink type has [two fields](https://github.com/grafana/k6/blob/f43cc83f82fe36ae75db018f4f935dd387fe675e/metrics/sink.go#L160-L163) the same fields required from the [remote service](https://github.com/grafana/k6/blob/2fb9e2da35a8081d2ee27d768943e8cdfd9d30f2/output/cloud/expv2/pbcloud/metric.proto#L110-L116).

output/cloud/expv2/collect.go
codebien: To clarify, of course, we should cut off not _closed_ buckets so that the obvious buckets where we will get more samples with high probability.

output/cloud/expv2/sink.go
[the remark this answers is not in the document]
mstoykov: Yes - rate is the only one for which the Sink implementation makes sense to be used. All the other have additional fields which arguably are not neede or even defined usually for those types of metrics. But we do have them to make the summary possible

output/cloud/expv2/output.go
```diff
@@ -42,7 +59,32 @@ func (o *Output) SetTestRunStopCallback(stopFunc func(error)) {
...
+		return err
+	}
+	o.periodicFlusher = pf
+
+	timeframe = o.config.AggregationPeriod.TimeDuration()
+	pfc, err := output.NewPeriodicFlusher(timeframe, o.collector.CollectSamples)
+	if err != nil {
+		return err
+	}
+	o.periodicCollector = pfc
```
mstoykov, line 87:
```suggestion
	o.periodicFlusher, err = output.NewPeriodicFlusher(timeframe, o.flushMetrics)
	if err != nil {
		return err
	}

	timeframe = o.config.AggregationPeriod.TimeDuration()
	o.periodicCollector , err = output.NewPeriodicFlusher(timeframe, o.collector.CollectSamples)
	if err != nil {
		return err
	}
```

To be honest I would even drop the `timeframe` variable jsut so I don't have to wonder if we are using it anywhere else.

output/cloud/expv2/collect.go
mstoykov: oops this actually is a map not a slice :facepalm:

output/cloud/expv2/sink.go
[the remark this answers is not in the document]
codebien: Oh, I got it now. Yes, you're right. The counter can be smaller, but the Gauge requires more fields instead.

https://github.com/grafana/k6/blob/2fb9e2da35a8081d2ee27d768943e8cdfd9d30f2/output/cloud/expv2/pbcloud/metric.proto#L96-L108

Gauges are used less so we should be able to support them.

output/cloud/expv2/collect.go
```diff
@@ -0,0 +1,158 @@
...
+
+		for i := 0; i < len(samples); i++ {
+			bucketID := c.bucketID(samples[i].Time)
+
+			// Get or create a time bucket
+			buckets, ok := c.aggrBuckets[bucketID]
+			if !ok {
+				buckets = newPerSeriesBuckets()
+				c.aggrBuckets[bucketID] = buckets
+			}
```
mstoykov, line 120: future optimization that we can do here that I came up while writing [this comment](https://github.com/grafana/k6/pull/3071#discussion_r1197945597) 

It is very likely that `bucketID` will haver the same value for sequential samples so will `buckets` but will do map access for each of those.

So we can potentially save the last `buecketID` + `bueckts` values and check against the new ones first.

I would levea that for when we have the rest of the things figured out and when there are some good benchmarks to run against.

codebien:
The last push should cover all request changes. It is ready for a new review.

output/cloud/expv2/sink.go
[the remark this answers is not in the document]
codebien: This is now a dedicated PR https://github.com/grafana/k6/pull/3085

output/cloud/expv2/output.go
codebien: This is now a dedicated PR https://github.com/grafana/k6/pull/3082 let's move the discussion there if you still want it.

oleiade (commented):
LGTM :+1: Left a couple of comments and questions. Let me know what you think :bow:

output/cloud/expv2/collect.go
```diff
@@ -0,0 +1,156 @@
...
+func (q *bucketQ) PopAll() []timeBucket {
+	q.m.Lock()
+	defer q.m.Unlock()
+
+	if len(q.buckets) < 1 {
+		return nil
+	}
+
+	b := q.buckets
+	q.buckets = make([]timeBucket, 0, len(b))
```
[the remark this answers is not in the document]
mstoykov: We return the existing slice so reusing it will race with whatever is getting the result

output/cloud/expv2/collect.go
```diff
@@ -0,0 +1,156 @@
...
+	var expired []timeBucket //nolint:prealloc
+
+	// Mark as expired all aggregation buckets older than bucketCutoffID
+	for bucketID, seriesSinks := range c.timeBuckets {
+		if bucketID > bucketCutoffID {
+			continue
+		}
+
+		expired = append(expired, timeBucket{
+			Time:  c.timeFromBucketID(bucketID),
```
mstoykov, line 134: I would probably just leave it as the `bucketID`. I am not certain but it seems to me that this is what we end up writing out later on?

output/cloud/expv2/output.go
```diff
@@ -42,7 +60,31 @@ func (o *Output) SetTestRunStopCallback(stopFunc func(error)) {
...
+	}
+	o.metricsFlusher = noopFlusher{
+		referenceID: o.referenceID,
+		bq:          &o.collector.bq,
+	}
+
+	pf, err := output.NewPeriodicFlusher(
+		o.config.MetricPushInterval.TimeDuration(), o.flushMetrics)
+	if err != nil {
+		return err
```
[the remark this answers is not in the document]
mstoykov: > I can't remember how k6 behaves if we fail starting an Output (do we retry?).

we stop the process

output/cloud/expv2/collect.go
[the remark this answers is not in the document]
oleiade: Fair enough, thanks for the clarification 🙇🏻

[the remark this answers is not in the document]
codebien: @oleiade I unresolved it because the main suggestion about the commenting makes sense to me.

output/cloud/expv2/collect.go
```diff
@@ -0,0 +1,162 @@
...
+}
+
+func (c *collector) bucketID(t time.Time) int64 {
+	return t.UnixNano() / int64(c.aggregationPeriod)
+}
+
+func (c *collector) timeFromBucketID(id int64) time.Time {
+	return time.Unix(0,
+		// it uses the center of the bucket as time
+		(id*int64(c.aggregationPeriod))+int64(c.aggregationPeriod/2),
```
codebien, line 156:
```suggestion
		// it uses the start point of the bucket as time
		id*int64(c.aggregationPeriod)
```
We don't need it anymore so we can simplify

output/cloud/expv2/collect.go
codebien: I will do this at the end of the chain, I prefer to have the support of the entire test suite so I'm sure I'm doing everything correctly, especially on the Protobuf side. Is it fine for you?

output/cloud/expv2/output.go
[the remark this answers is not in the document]
codebien: As said the expectation here is that k6 stops if we fail to start an output so no samples are added. Feel free to unresolve if I miss anything.

output/cloud/expv2/collect.go
```diff
@@ -0,0 +1,156 @@
...
+	// we should no longer have to handle metrics that have times long in the past. So instead of a
+	// map, we can probably use a simple slice (or even an array!) as a ring buffer to store the
+	// aggregation buckets. This should save us a some time, since it would make the lookups and WaitPeriod
+	// checks basically O(1). And even if for some reason there are occasional metrics with past times that
+	// don't fit in the chosen ring buffer size, we could just send them along to the buffer unaggregated
+	timeBuckets map[int64]map[metrics.TimeSeries]metrics.Sink
+}
+
+func newCollector(aggrPeriod, waitPeriod time.Duration) (*collector, error) {
+	if aggrPeriod == 0 {
```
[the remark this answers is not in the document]
codebien: I made it mandatory in the last push.

oleiade (approved):

output/cloud/expv2/output.go
mstoykov: This still seems to be a problem @codebien maybe you rebased without it?

output/cloud/expv2/collect.go
```diff
@@ -0,0 +1,168 @@
...
+		bq:                bucketQ{},
+		nowFunc:           time.Now,
+		timeBuckets:       make(map[int64]map[metrics.TimeSeries]metrics.Sink),
+		aggregationPeriod: aggrPeriod,
+		waitPeriod:        waitPeriod,
+	}, nil
+}
+
+// CollectSamples drain the buffer and collect all the samples.
+func (c *collector) CollectSamples(containers []metrics.SampleContainer) {
```
na--, line 90: Minor non-blocking nit, but is there a reason for some of the methods of `collector` and `bucketQ` to be exported? They are only used in the `expv2.Output`, right? So it seems like everything could be private for just this package?

output/cloud/expv2/output.go
```diff
@@ -51,14 +93,100 @@ func (o *Output) Start() error {
...
+
+// flushMetrics receives a set of metric samples.
+func (o *Output) flushMetrics() {
+	start := time.Now()
+
+	ctx, cancel := context.WithTimeout(context.Background(), o.config.MetricPushInterval.TimeDuration())
+	defer cancel()
+
+	err := o.metricsFlusher.Flush(ctx)
+	if err != nil {
```
na--, line 153: Minor non-blocking nit: you can probably flip this `if` around and have something like this:
```go
if err == nil {
	o.logger.WithField("t", time.Since(start)).Debug("Successfully flushed buffered samples to the cloud")
	return
}
// ...long nested logic...
```

na-- (approved):
LGTM in general (though I skipped looking over the tests this time)

mstoykov (approved):

output/cloud/expv2/output.go
codebien: This is one line after the refactor in this commit in the next PR https://github.com/grafana/k6/pull/3082/commits/8a9acd673f7191474db008924dc4d958d40a0311

oleiade (approved):

output/cloud/expv2/collect.go
codebien: I set a `TODO:` for this to check all the export/unexport states and add it to one of the last PRs.

Issue #3069, opened 2023-05-15
Update grpc lib to 1.55.0+


mstoykov:
## What?

Update grpc dependancy to v1.55.0+

## What is the problem

https://github.com/grpc/grpc-go/pull/6164 basically breaks all our grpc tests. While replacign with the new import works for big percentage of them some also require the proto file which is now not in the repo.

## Appendum:
github.com/golang/protobuf
google.golang.org/protobuf
github.com/jhump/protoreflect

Also should be updated but I prefer to do it with the big gRPC lib just in case there is something funky.

PR #3075, opened 2023-05-17
GRPC dependencies updates

olegbespalov:
# What?

This PR brings updates of the dependencies with the fixes (see commits breakdown):
* the first fix is to frizes the grpc testing proto as the k6 package, which removes the hard coupling to the dependency version and unlocks updating of the dependencies
* second fix is also around the logic in tests that could cause the issue with newer versions of the `github.com/jhump/protoreflect`

# Why?

Closes: #3069


lib/netext/grpcext/conn_test.go
```diff
@@ -259,9 +259,17 @@ func (m *getServiceFileDescriptorMock) Recv() (*reflectpb.ServerReflectionRespon
...
+	path := "any-path"
 	parser := protoparse.Parser{
 		InferImportPaths: false,
 		Accessor: protoparse.FileAccessor(func(filename string) (io.ReadCloser, error) {
+			// a small hack to make sure we are parsing the right file
+			// otherwise the parser will try to parse "google/protobuf/descriptor.proto"
+			// with exactly the same name as the one we are trying to parse for testing
+			if filename != path {
+				return nil, nil
+			}
```
mstoykov, line 271: Why is it trying to load a different file?

mstoykov (approved):
LGTM! although I feel like we now have 3 different packages with protosets and am not certain if we should have 3 different ones :shrug:

imiric (approved):
I didn't go through the vendored changes, but the ones in our repo look good. :+1: Besides that hack that Mihail mentioned, which is not very clear.

Do you know what happened to the `test.proto` file after it was deleted in https://github.com/grpc/grpc-go/pull/6164? Is it now embedded in the generated `pb.go` files? I couldn't figure out from the PR how they're testing with the new package.

It sucks that we now have to maintain `grpc_testing` ourselves, but it's better to have control over it, and we're forced to anyway :-/

lib/netext/grpcext/conn_test.go
olegbespalov: It's coming from the library update https://github.com/bufbuild/protocompile/commit/7c5114ee37b44b274e401c07b53f3fbfadf81dbb

olegbespalov:
>Do you know what happened to the test.proto file after it was deleted in https://github.com/grpc/grpc-go/pull/6164? Is it now embedded in the generated pb.go files? I couldn't figure out from the PR how they're testing with the new package.

Well, yeah, it seems like the source proto file was just deleted, we could maybe check with authors (or maybe just open the PR to them)

https://github.com/grpc/grpc-go/blob/master/interop/grpc_testing/test.pb.go#L22

PR #3078, opened 2023-05-19
pinning builder stage（alpine 3.17）

arukiidou:
- ~Upgrade to alpine 3.18, and pin alpine version~
- pinning builder stage（use alpine 3.17）


imiric (approved):
Thanks! :bow:

Pinning the builder stage makes sense, but do we have a reason to upgrade to 3.18? The only relevant change for k6 would be [the addition of TCP fallback in the DNS resolver](https://alpinelinux.org/posts/Alpine-3.18.0-released.html), but this is a niche scenario, and not very important for the Docker image, since users could just copy the k6 binary into their own base images and use any distro. They're not required to use the distro we ship our Docker image with.

So I'm not against this, but would like us to have a good reason for any upgrade. Otherwise we risk breaking something that worked with the previous version.

arukiidou:
> Thanks! 🙇
> 
> Pinning the builder stage makes sense, but do we have a reason to upgrade to 3.18? The only relevant change for k6 would be [the addition of TCP fallback in the DNS resolver](https://alpinelinux.org/posts/Alpine-3.18.0-released.html), but this is a niche scenario, and not very important for the Docker image, since users could just copy the k6 binary into their own base images and use any distro. They're not required to use the distro we ship our Docker image with.
> 
> So I'm not against this, but would like us to have a good reason for any upgrade. Otherwise we risk breaking something that worked with the previous version.

OK, I will remove the bumps, leaving only pinning the builder stage.
The reason I wanted to bump alpine was to use other new packages, but now does not seem like the time to bump😅.

arukiidou:
It may not be working?🤔

https://github.com/grafana/k6/blob/f43cc83f82fe36ae75db018f4f935dd387fe675e/.github/actions/lint/action.yml#L17

imiric (approved):

imiric:
Yeah, don't worry about the `lint` job failure. We have to fix it to work for PRs from forked repos.

codebien (approved):

PR #3082, opened 2023-05-21
output/cloudv2: Error handling for flush

codebien:
It is a small refactor for the error handling on the flush operation.

### DIFF
--- a/output/cloud/expv2/output.go
+++ b/output/cloud/expv2/output.go
@@ -183,7 +183,7 @@ func (o *Output) flushMetrics() {
 }
 
 // handleFlushError handles errors generated from the flushing operation.
-// It may interrupts the metric collection or invokes abort of the test.
+// It may interrupt the metric collection or invoke aborting of the test.
 //
 // note: The actual test execution should continue, since for local k6 run tests
 // the end-of-test summary (or any other outputs) will still work,
--- a/output/cloud/expv2/output.go
+++ b/output/cloud/expv2/output.go
@@ -31,19 +31,23 @@ type Output struct {
 	metricsFlusher noopFlusher
 	collector      *collector
 
-	stopSamplesCollection chan struct{}
+	// wg tracks background goroutines
+	wg sync.WaitGroup
 
-	wg   sync.WaitGroup
+	// stop signal to graceful stop
 	stop chan struct{}
+
+	// abort signal to interrupt immediately all background goroutines
+	abort chan struct{}
 }
 
 // New creates a new cloud output.
 func New(logger logrus.FieldLogger, conf cloudapi.Config) (*Output, error) {
 	return &Output{
-		config:                conf,
-		logger:                logger.WithFields(logrus.Fields{"output": "cloudv2"}),
-		stopSamplesCollection: make(chan struct{}),
-		stop:                  make(chan struct{}),
+		config: conf,
+		logger: logger.WithFields(logrus.Fields{"output": "cloudv2"}),
+		abort:  make(chan struct{}),
+		stop:   make(chan struct{}),
 	}, nil
 }
 
@@ -92,7 +96,7 @@ func (o *Output) StopWithTestError(testErr error) error {
 	o.wg.Wait()
 
 	select {
-	case <-o.stopSamplesCollection:
+	case <-o.abort:
 		return nil
 	default:
 	}
@@ -113,7 +117,7 @@ func (o *Output) AddMetricSamples(s []metrics.SampleContainer) {
 	// evaluate to do something smarter, maybe having a lock-free
 	// queue.
 	select {
-	case <-o.stopSamplesCollection:
+	case <-o.abort:
 		return
 	default:
 	}
@@ -143,7 +147,7 @@ func (o *Output) periodicInvoke(d time.Duration, callback func()) {
 			case <-t.C:
 				callback()
 			case <-o.stop:
-			case <-o.stopSamplesCollection:
+			case <-o.abort:
 				return
 			}
 		}
@@ -204,10 +208,10 @@ func (o *Output) handleFlushError(err error) {
 	// Do not close multiple times (that would panic) in the case
 	// we hit this multiple times and/or concurrently
 	select {
-	case <-o.stopSamplesCollection:
+	case <-o.abort:
 		return
 	default:
-		close(o.stopSamplesCollection)
+		close(o.abort)
 	}
 
 	if o.config.StopOnError.Bool {
--- a/output/cloud/expv2/output.go
+++ b/output/cloud/expv2/output.go
@@ -87,8 +87,16 @@ func (o *Output) Start() error {
 // StopWithTestError gracefully stops all metric emission from the output.
 func (o *Output) StopWithTestError(testErr error) error {
 	o.logger.Debug("Stopping...")
+	defer o.logger.Debug("Stopped!")
+
 	close(o.stop)
 	o.wg.Wait()
+
+	select {
+	case <-o.stopSamplesCollection:
+		return nil
+	default:
+	}
 
 	// Drain the SampleBuffer and force the aggregation for flushing
 	// all the queued samples even if they haven't yet passed the
@@ -97,7 +105,6 @@ func (o *Output) StopWithTestError(testErr error) error {
 	o.collector.CollectSamples(nil)
 	o.flushMetrics()
 
-	o.logger.Debug("Stopped!")
 	return nil
 }
 
@@ -128,7 +135,7 @@ func (o *Output) AddMetricSamples(s []metrics.SampleContainer) {
 func (o *Output) periodicInvoke(d time.Duration, callback func()) {
 	defer o.wg.Done()
 
-	t := time.NewTimer(d)
+	t := time.NewTicker(d)
 	defer t.Stop()
 	for {
 		select {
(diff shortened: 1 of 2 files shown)


output/cloud/expv2/output.go
```diff
@@ -144,42 +152,51 @@ func (o *Output) flushMetrics() {
...
 	}
 
 	o.logger.WithField("t", time.Since(start)).Debug("Successfully flushed buffered samples to the cloud")
 }
 
-// shouldStopSendingMetrics returns true if the output should interrupt the metric flush.
+// handleFlushError handles errors generated from
+// the flushing operation.
+// It may interrupts the metric collection
+// or invokes abort of the test.
```
imiric, line 165:
```suggestion
// handleFlushError handles errors generated from the flushing operation.
// It may interrupt the metric collection or invoke aborting of the test.
```

Why do you use such short lines? It's OK to wrap at ~80 or even ~120.

output/cloud/expv2/output.go
```diff
@@ -144,42 +152,51 @@ func (o *Output) flushMetrics() {
...
 //
 // note: The actual test execution should continues,
```
imiric, line 167:
```suggestion
// note: The actual test execution should continue,
```

output/cloud/expv2/output.go
```diff
@@ -144,42 +152,51 @@ func (o *Output) flushMetrics() {
...
+		return
 	}
-	if errResp, ok := err.(cloudapi.ErrorResponse); ok && errResp.Response != nil { //nolint:errorlint
-		// The Cloud service returns the error code 4 when it doesn't accept any more metrics.
-		// So, when k6 sees that, the cloud output just stops prematurely.
-		return errResp.Response.StatusCode == http.StatusForbidden && errResp.Code == 4
+
+	if errResp.Code != 4 {
+		return
 	}
```
imiric, line 189: Join this with the above check?

```go
if errResp.Response.StatusCode != http.StatusForbidden || errResp.Code != 4 {
	return
}
```

imiric (approved):
LGTM, just minor suggestions.

output/cloud/expv2/output.go
```diff
@@ -29,8 +30,14 @@ type Output struct {
...
-	periodicCollector  *output.PeriodicFlusher
+	collector         *collector
+	periodicCollector *output.PeriodicFlusher
+
+	// TODO: rename; this is an old name from v1
+	// it now stops the entry point to buffering samples
+	//
+	// What about stopMetricCollection
+	// Does it sound better?
 	stopSendingMetrics chan struct{}
```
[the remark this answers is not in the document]
codebien: It is renamed to `stopSamplesCollection` as suggested here: https://github.com/grafana/k6/pull/3085#discussion_r1201933374 because it respects what it does better.

output/cloud/expv2/output.go
codebien: Done

output/cloud/expv2/output.go
codebien: Done. And added `setcolumn` in my vim config :pray:

output/cloud/expv2/output.go
```diff
@@ -124,13 +118,30 @@ func (o *Output) AddMetricSamples(s []metrics.SampleContainer) {
...
 	// run it and save some memory across the e2e operation.
 	//
 	// It requires very specific benchmark.
 	o.SampleBuffer.AddMetricSamples(s)
 }
 
+func (o *Output) periodicInvoke(d time.Duration, callback func()) {
+	defer o.wg.Done()
+
+	t := time.NewTimer(d)
```
mstoykov, line 131:
```suggestion
	t := time.NewTicker(d)
```

output/cloud/expv2/output.go
```diff
@@ -151,42 +162,44 @@ func (o *Output) flushMetrics() {
...
+
+	// The Cloud service returns the error code 4 when it doesn't accept any more metrics.
+	// So, when k6 sees that, the cloud output just stops prematurely.
+	if errResp.Response.StatusCode != http.StatusForbidden || errResp.Code != 4 {
+		return
 	}
 
-	return false
+	o.logger.WithError(err).Warn("Interrupt sending metrics to cloud due to an error")
+	close(o.stopSamplesCollection)
```
mstoykov, line 195: I think this can panic as `handleFlushError` is called in `flushMetrics` which is also called in `Stop`. 

So it seems to me that when the output is being stopped it will go through this again and try to close this again.

codebien: Great catch, it should be protected now.

output/cloud/expv2/output.go
```diff
@@ -151,42 +171,52 @@ func (o *Output) flushMetrics() {
...
+	o.logger.WithError(err).Warn("Interrupt sending metrics to cloud due to an error")
+
+	// Do not close multiple times (that would panic) in the case
+	// we hit this multiple times and/or concurrently
+	select {
+	case <-o.stopSamplesCollection:
+		return
+	default:
+		close(o.stopSamplesCollection)
+	}
```
mstoykov, line 212: While this will work as `handleFlushError` is only called in flush and that is called only in one placea at a time.

I kind of feel we are doing this in the wrong place - if there was an error that will abort teh flushing of metrics we should stop *flushing* metrics. Not only collecting them.

output/cloud/expv2/output.go
```diff
@@ -72,19 +76,8 @@ func (o *Output) Start() error {
...
-	o.periodicFlusher = pf
-
-	pfc, err := output.NewPeriodicFlusher(
-		o.config.AggregationPeriod.TimeDuration(), o.collectSamples)
-	if err != nil {
-		return err
-	}
-	o.periodicCollector = pfc
+	o.periodicInvoke(o.config.MetricPushInterval.TimeDuration(), o.flushMetrics)
+	o.periodicInvoke(o.config.AggregationPeriod.TimeDuration(), o.collectSamples)
```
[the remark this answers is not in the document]
codebien: The unique alternative I have in mind is to have it directly as a function in the Start method.

```go
func (o Out) Start() {
periodic := func() {
  ...
}
periodic(d, collect)
periodic(d, flush)
}
```

output/cloud/expv2/output.go
codebien: Do we want to check this all the time when the expectation for this is to get it very few times? Why should we continuously sync over this select statement for a successful test where we will never hit it?

mstoykov: Why should we make a network call trhat will fail ?

mstoykov: Also in practice you can just have it in the `Stop` method I guess - no need to go through all the different parts if we are not going to flush anything.

codebien: >Also in practice you can just have it in the Stop method I guess 

The problem with this is when we will move to have multiple-concurrent flushers then we could get this function called multiple times.

codebien: I think we may try to have this handler as a goroutine and get the error on a channel in the classic go way. In this way, we have a centralized place for doing the close.

mstoykov: We do something like this in [the v1](https://github.com/grafana/k6/blob/master/output/cloud/v1/output.go) maybe more of the structure should be copied from there

codebien: I tried but I don't think it is cleaner. How does it achieve the goal?! Invoking the `flush` operation from the aggregation goroutine that it sounds not optimal.

I'm not against doing the refactor and I am happy to improve it, but it doesn't sound to me like a priority right now. Performances should be fine and the logic has been fixed not panicking. I included it in the reminders list.

I would like to merge the chain before and the retry + concurrent flushers then with a better picture we can do the refactor.

Also, the metadata flush will be added soon, so the risk is to refactor this multiple times in a few days.

codebien:
I renamed `stopSamplesCollection` to `abort` because it sounds more generic considering that it stops multiple operations.

mstoykov (dismissed):

output/cloud/expv2/output.go
```diff
@@ -151,42 +175,54 @@ func (o *Output) flushMetrics() {
...
+		o.handleFlushError(err)
 		return
 	}
 
 	o.logger.WithField("t", time.Since(start)).Debug("Successfully flushed buffered samples to the cloud")
 }
 
-// shouldStopSendingMetrics returns true if the output should interrupt the metric flush.
+// handleFlushError handles errors generated from the flushing operation.
+// It may interrupts the metric collection or invokes abort of the test.
```
imiric, line 186:
```suggestion
// It may interrupt the metric collection or invoke aborting of the test.
```

output/cloud/expv2/output_test.go
```diff
@@ -114,3 +115,163 @@ func TestOutputCollectSamples(t *testing.T) {
...
+	cb := func() {
+		updated := atomic.AddUint64(&called, 1)
+		if updated == 2 {
+			close(stop)
+		}
+	}
+	o := Output{stop: stop}
+	o.periodicInvoke(time.Duration(1), cb) // loop
+	<-stop
+	assert.Greater(t, atomic.LoadUint64(&called), uint64(1))
```
imiric, line 262: Why not `assert.Equal()` here?

output/cloud/expv2/output.go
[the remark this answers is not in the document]
imiric: Can we just modify `PeriodicFlusher` to not call the callback if configured so? We're essentially duplicating what it does to avoid that final call.

Otherwise, if we're fine with the duplication, then this being a method doesn't bother me, since it likely won't be needed anywhere else.

imiric (approved):
LGTM :+1:

output/cloud/expv2/output.go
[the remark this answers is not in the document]
codebien: We may evaluate extending the current periodic helper but we need to be sure this is the final architecture. As commented here https://github.com/grafana/k6/pull/3082#discussion_r1207875810, I would like to stabilize the rest then go back here and make some attempts for refactoring it in a new PR.

output/cloud/expv2/output_test.go
codebien: @imiric Because there isn't a guarantee that the close chan is picked before doing a new iteration with a so small time frame.

I don't know if there is a time frame that would guarantee us a sync flow but it would mean relying on time.

imiric (approved):

mstoykov (approved):

PR #3083, opened 2023-05-21
output/cloudv2: Flush the aggregated metrics

codebien:
It implements the drain of the queue of the aggregated metrics and the conversion to Protobuf.

PR #3085, opened 2023-05-22
output/cloudv2: Optimized metric sinks

codebien:
It address this request change https://github.com/grafana/k6/pull/3071#discussion_r1197893018


output/cloud/expv2/sink.go
```diff
@@ -1,13 +1,78 @@
...
 import (
+	"fmt"
+
 	"go.k6.io/k6/metrics"
 )
 
-func newSink(mt metrics.MetricType) metrics.Sink {
-	if mt == metrics.Trend {
-		return &histogram{}
+type aggregatedMetric interface {
```
oleiade, line 9: While I appreciate the intention to rename `Sink`, which got me confused more than once as I look at it, I find that `aggregatedMetric` is somewhat [tautological](https://dictionary.cambridge.org/dictionary/english/tautological) from my point of view. I think of metrics as aggregated views over specific classes of samples over time.

What would you think of `sampleAggregator`, `sampleAccumulator`, or `metricRunningValue` instead?

PS: absolutely not a blocker though 👍🏻

output/cloud/expv2/output.go
```diff
@@ -120,7 +117,7 @@ func (o *Output) AddMetricSamples(s []metrics.SampleContainer) {
 	// evaluate to do something smarter, maybe having a lock-free
 	// queue.
 	select {
-	case <-o.stopSendingMetrics:
+	case <-o.stopMetricsCollection:
```
oleiade, line 120: Wasn't this somewhat misnamed in the first place? From the context, I tend to think that it should be named `stopSamplesCollection`, or `stopSamplesAggregation` instead. What do you think? Am I missing something?

output/cloud/expv2/mapping.go
```diff
@@ -51,43 +51,42 @@ func addBucketToTimeSeriesProto(
...
-			// TODO: implement the custom rate for safe track them
-			NonzeroCount: uint32(r.Trues),
-			TotalCount:   uint32(r.Total),
+			Time:         timestamppb.New(time),
+			NonzeroCount: r.Trues,
+			TotalCount:   r.Total,
 		})
 	case metrics.Trend:
 		h := sink.(*histogram) //nolint: forcetypeassert
+		h.trimzeros()
```
oleiade, line 89: I posted a question on the [histogram PR directly ](https://github.com/grafana/k6/pull/3027/files#r1201939608), and I wonder if not calling `trimzeros` would cause potential bugs. In which case, I wonder if we should find ways to have that done automatically somehow, if possible, to not expose ourselves to potential bugs in the future because we forgot to call it?

oleiade (commented):
LGTM 👍🏻 🎉 

I left a comment on naming and a question about `trimzeros`, that I don't consider blockers. I would approve as is, but I wanted to make sure those were addressed (as in "I got an answer", not as in "they made the change I wanted" 😉) first 🙇🏻

output/cloud/expv2/sink.go
codebien: I appreciate this suggestion, I picked `metricValue`. I removed _running_ because I think we don't really have a no-running metric value so it doesn't add any value, IMHO.

Let's reopen if you disagree.

output/cloud/expv2/output.go
codebien: Moved this to https://github.com/grafana/k6/pull/3082, let's discuss there

output/cloud/expv2/sink.go
```diff
@@ -1,13 +1,78 @@
...
+
+type gauge struct {
+	Last     float64
+	Sum      float64
+	Min, Max float64
+	Avg      float64
+	Count    uint32
+
+	minSet bool
+}
```
codebien, line 51: Moving here the discussion from https://github.com/grafana/k6/pull/3083#discussion_r1205527232

>Shouldn't this be just Last? Which is what Gauge basically is?

codebien: Hey @esquonk, apparently other protocols (OTel, Prometheus remote write) don't support gauge in the same way. We may have issues if we want to migrate to them in the future. Considering this are not used at the moment, can we just drop them?

esquonk: We have those in the backend historically, and they are used. For example, if we have a "max" query for gauge metric, the code will look at "max" column.

We could just fill all the columns with "last" value at the cost losing a bit of data, but if we can collect some more statistical values, it make sense to do so. Unless it is too expensive to maintain.

As for compatibility with other protocols, I don't see how having more data internally can cause issues with that. If you need to produce an output in some protocol that only has `last`, you can just put last in there and ignore other values in your output.

output/cloud/expv2/mapping.go
```diff
@@ -52,39 +52,38 @@ func addBucketToTimeSeriesProto(
...
+	sink metricValue,
 ) {
 	if timeSeries.Samples == nil {
 		initTimeSeriesSamples(timeSeries, mt)
 	}
 
 	switch mt {
 	case metrics.Counter:
-		c := sink.(*metrics.CounterSink).Value //nolint: forcetypeassert
+		c := sink.(*counter).Sum //nolint: forcetypeassert
```
codebien, line 63: https://github.com/grafana/k6/pull/3083#discussion_r1209300347

output/cloud/expv2/mapping.go
```diff
@@ -52,44 +52,39 @@ func addBucketToTimeSeriesProto(
...
+	sink metricValue,
 ) {
 	if timeSeries.Samples == nil {
 		initTimeSeriesSamples(timeSeries, mt)
 	}
 
-	switch mt {
-	case metrics.Counter:
-		c := sink.(*metrics.CounterSink).Value //nolint: forcetypeassert
+	switch typedSink := sink.(type) {
```
mstoykov, line 61: the type is now `metricValue` instead of `Sink` so a rename of both is likely in order.

mstoykov (dismissed):
LGTM! in general with the understandign that Gauge is not really a Gauge because the k6 cloud backend wants it like this.

I have left some small comments, and one big one which is mostly an optimizaiton idea for after we have merged everything and this output works.

oleiade (approved):
LGTM 👍🏻

output/cloud/expv2/sink.go
```diff
@@ -1,13 +1,78 @@
...
+		g.Max = v
+	}
+	if !g.minSet || v < g.Min {
+		g.Min = v
+		g.minSet = true
+	}
+}
+
+type rate struct {
+	NonZeroCount uint32
```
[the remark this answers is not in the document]
codebien: They are not only positive, negative are accepted. The condition:
https://github.com/grafana/k6/blob/986db71ee65c1891ef0a5672e22e6d0ca84353c1/output/cloud/expv2/sink.go#L75-L77

Issue #1974, opened 2021-04-16
TestDNSResolver/cache is flaky on windows


mstoykov:
```
--- FAIL: TestDNSResolver (24.22s)
    --- FAIL: TestDNSResolver/cache (24.22s)
        --- FAIL: TestDNSResolver/cache/1000 (5.94s)
            local_test.go:1070: 
                	Error Trace:	local_test.go:1070
                	Error:      	Error message not equal:
                	            	expected: "context deadline exceeded"
                	            	actual  : "dial tcp 127.0.0.254:50679: i/o timeout"
                	Test:       	TestDNSResolver/cache/1000
```

https://github.com/k6io/k6/runs/2362979683?check_suite_focus=true

PR #3088, opened 2023-05-23
Maybe fix `TestDNSResolver` flakiness on Windows

imiric:
This should hopefully be more reliable, and avoid [the flakiness we've seen on Windows](https://github.com/grafana/k6/actions/runs/5047828671/jobs/9071063454), even after #1974:

```
scheduler_ext_test.go:1130:
        Error Trace:    D:/a/k6/k6/execution/scheduler_ext_test.go:1130
        Error:          Should be true
        Test:           TestDNSResolver/cache/3s
        Messages:       expected error to contain one of the list of messages
```

It doesn't get rid of the reliance on timings, but since the way we check DNS TTL also relies on time, we can't easily avoid it. Maybe by switching to the `MiniRunner` instead of `js.Runner` we could avoid the `sleep()` in the script, but the value of this test is also calling the `k6/http` module directly, so it's closer to a real world script.


imiric:
Nope, checking samples seems to be even flakier than checking logs... :face_exhaling: I guess we need to remove the reliance on timing altogether. :thinking:

execution/scheduler_ext_test.go
```diff
@@ -1045,96 +1048,92 @@ func TestDNSResolver(t *testing.T) {
...
-							strings.Contains(errMsg, "request timeout") ||
-							strings.Contains(errMsg, "dial: i/o timeout")
-						assert.Truef(t, contains, "expected error to contain one of the list of messages")
+			var failedRequests uint32
+			samplesHandler := func(sc metrics.SampleContainer) {
+				if sc == nil {
+					return
+				}
+				for _, s := range sc.GetSamples() {
+					if s.Metric.Name == metrics.HTTPReqFailedName && s.Value == 1 {
```
codebien, line 1110: I was thinking about something like:
```suggestion
					if s.Metric.Name == metrics.HTTPReqsName && s.Value == 1 {
					    changeMockResolver <- struct{}{}
```
then

```go
go func() {
<-changeMockResolver
mr.Set("myhost", "127.0.0.254")
}()
```
I think it should remove the timer.After requirement

imiric: I guess we could do that, but I don't like that we'd be relying on metric emission for the functionality of the test. That is, if `http_reqs` is never emitted, `mr.Set()` wouldn't be called, and the failure would be confusing.

And it doesn't solve the timing issue, since we still use `sleep()` in the script. But I'll give it a try, and also consider using `MiniRunner` instead.

imiric:
Hey @codebien @olegbespalov, please see the latest force push. I abandoned checking samples, since I didn't want to depend on metric emission for the functionality of the test, and using `MiniRunner` would make the test less useful, since it wouldn't be checking the DNS resolution behavior within `js.Runner`.

Adding a small hook to `MockResolver` avoids the time-sensitive behavior, so it should hopefully be more reliable now. Let me know.

execution/scheduler_ext_test.go
imiric: I removed checking samples in the latest force-push, so I'm not doing this anymore.

lib/testutils/mockresolver/resolver.go
```diff
@@ -9,17 +9,17 @@ import (
 // MockResolver implements netext.Resolver, and allows changing the host
 // mapping at runtime.
 type MockResolver struct {
-	m        sync.RWMutex
-	hosts    map[string][]net.IP
-	fallback func(host string) ([]net.IP, error)
+	m           sync.RWMutex
+	hosts       map[string][]net.IP
+	ResolveHook func(host string, result []net.IP)
```
olegbespalov, line 14: nit: I'd suggest naming this `LookupIPAllHock` since `MockResolver` has also `LookupIP` and there we don't execute the hock after the name resolving.

olegbespalov (approved):
Left a small non-blocking comment, but generally looks good :+1:

Let's merge it and make our tests trusty again :muscle: :sweat_smile:

execution/scheduler_ext_test.go
```diff
@@ -1045,96 +1044,97 @@ func TestDNSResolver(t *testing.T) {
...
-		}
-	})
+			errCh := make(chan error, 1)
+			go func() { errCh <- execScheduler.Run(ctx, ctx, samples) }()
+
+			select {
+			case err := <-errCh:
+				require.NoError(t, err)
+				assert.Equal(t, tc.expNewResolutions, atomic.LoadUint32(&newResolutions))
+			case <-time.After(10 * time.Second):
```
codebien, line 1133:
```suggestion
			case <-time.After(2 * time.Second):
```
nit: I think at this point we can cut the waiting time in case we will have a failure.

codebien (approved):
Well done! This is for sure a better solution. :clap:

lib/testutils/mockresolver/resolver.go
imiric: But `LookupIP` calls `LookupIPAll`, so the hook would run either way.

PR #3094, opened 2023-05-25
Fix example in run command

rockwotj:
The second stage was missing the number of VU, and actually running that command caused complaints.


imiric (approved):
Thanks for catching it! This example has been wrong since the beginning of k6 in 2017 :sweat_smile:

olegbespalov (approved):
Good catch, thanks! :+1:

PR #3099, opened 2023-05-29
Change thresholds-error verb from breach to cross

MattDodsonEnglish:
The error message "thresholds have been breached" sounded harshly to my ear. Normally I associate "breaches" with illegal acts, data exfiltrations, and military sieges.

I proposed changing the verb to "cross," which is the verb used by the docs. True, I believe I wrote those docs, so I'm biased. But I also checked the [Corpus of Contemporary American English](https://www.english-corpora.org/coca/), and "cross" is by far the most common collocation (word that co-occurs) with "threshold".

image


imiric (approved):
Sure, sounds good to me. :+1:

I don't find "breach" unclear, but you're right that "cross" is less harsh.

oleiade (approved):

PR #3100, opened 2023-05-29
output/cloudv2: Implement RequestMetadata output

Blinkuu:
This PR adds support for sending request metadata to the k6 Insights cloud backend.


output/cloud/expv2/output.go
```diff
@@ -233,3 +307,24 @@ func (o *Output) handleFlushError(err error) {
...
+
+func (o *Output) tracingEnabled() bool {
+	// TODO(lukasz): Check if the `tracing` module is imported.
+	//
+	// We want to check if the `tracing` module was imported. If
+	// it wasn't, we don't need to collect any data related to tracing
+	// as there won't be any.
+	//
+	// We currently don't have a way to check if the tracing module
+	// was imported by the user.
```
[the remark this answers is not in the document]
mstoykov: My initial thought is that `(2)` is not actually a requirement at all.

FIrst and foremost just because someone imports this doesn't mean that they are not or they are using it to do tracing. Especially on the `not` side.

In this version of k6 #3037 will be released which means that practically the whole `k6/experimental/tracing` can now be written entirely in js code. 

It also means that users can now create and tag metrics with metadata that looks like the one from k6/experimental/tracing. And they might want to do it for other protocols/APIS (including not HTTP ones, which I guess we are not going to support initially).

Given that and the fact that `K6_CLOUD_TRACES_ENABLED` flag I feel like this is needless optimization to try to not do some more processing that will just bites us later on.

output/cloud/expv2/collect.go
```diff
@@ -2,12 +2,24 @@ package expv2
...
+	"go.k6.io/k6/cloudapi/insights"
+	"go.k6.io/k6/js/modules/k6/experimental/tracing"
+	"go.k6.io/k6/lib/netext/httpext"
 	"go.k6.io/k6/metrics"
 )
 
+const (
+	scenarioTag = "scenario"
+	groupTag    = "group"
+	nameTag     = "name"
```
Blinkuu, line 18: Is this `name` tag by default set to `http_url` if users don't specify a value for it? My current understanding is that this tag is used to deal with cardinality issues ([source](https://k6.io/docs/using-k6/http-requests/#url-grouping)).

output/cloud/expv2/output.go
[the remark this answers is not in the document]
mstoykov: I am not really certain going through all the metrics and checking whether they have metadata will be all that much "compute" compared to everything else the output(s) need to do.

And we likely can do it in a single place, so that will be even less of an issue. 

By a single place I mean that output v2 will aggregate and on the side check that a given metric sample has metadata - if it does - send it to the insights output.

Even without that I expect this to be an insignificant amount of work.

Also, again - we will only to this for the users that have this enabled, and I expect they will likely want to use that.

cloudapi/insights/proto/gen.go
```diff
@@ -0,0 +1,5 @@
+// Package proto contains the Protobuf definitions used
+// by the k6 Insights.
+package proto
+
+//go:generate sh gen.sh
```
mstoykov, line 5: I would prefer if you just embed the whole thing in here and *not* use `find`.

It will be nice if this actually has good chances of working on a windows machine, withouhg needing fixes.

js/modules/k6/experimental/tracing/trace_id.go
```diff
@@ -18,8 +18,8 @@ const (
 	// To not ingest and process the related spans, b/c they are part of a non-cloud run.
 	k6LocalCode = 33
 
-	// metadataTraceIDKeyName is the key name of the traceID in the output metadata.
-	metadataTraceIDKeyName = "trace_id"
+	// MetadataTraceIDKeyName is the key name of the traceID in the output metadata.
+	MetadataTraceIDKeyName = "trace_id"
```
mstoykov, line 22: nit: I am kind of against this being public and you using it.

Again this package is 99% likely to never get out of experimental and users will be able to add `trace_id` even without it.

It seems a lot more reasonable for the insights output to just document that certain metrics with the `trace_id` metadata will have their samples processed. 

Having this exported to get this string seems completely unnecessary dependancy. 

see https://go-proverbs.github.io/

output/cloud/expv2/collect.go
mstoykov: It is set by default if it is enabled - which it must be for the usage of the cloud output. 

Other required tags for tis output include `method`, `status`, `error`, `check` and `group`

https://github.com/grafana/k6/blob/3ab0085ded5b85b54b7147faaa85e1520ecff7b4/output/cloud/output.go#L134-L154

cloudapi/config.go
```diff
@@ -143,11 +153,16 @@ type Config struct {
...
-		MetricPushInterval:         types.NewNullDuration(1*time.Second, false),
-		MetricPushConcurrency:      null.NewInt(1, false),
+		Host:                  null.NewString("https://ingest.k6.io", false),
+		LogsTailURL:           null.NewString("wss://cloudlogs.k6.io/api/v1/tail", false),
+		WebAppURL:             null.NewString("https://app.k6.io", false),
+		MetricPushInterval:    types.NewNullDuration(1*time.Second, false),
+		MetricPushConcurrency: null.NewInt(1, false),
+
+		TracesEnabled:      null.NewBool(false, false),
+		TracesHost:         null.NewString("insights.k6.io", false),
```
mstoykov, line 163: All the other urls/hosts here include a schema, so at least to be consistent it is likely a good idea to keep it that way.

Blinkuu: This is a gRPC service. Should I add the `dns:` prefix, then?

gRPC naming: https://github.com/grpc/grpc/blob/master/doc/naming.md

mstoykov: Hmmm ... of course grpc does not have `grpc` schema :( 


Does this code support the other ones `ipv4`/`ipv6`  and or `unix` ?

output/cloud/expv2/output.go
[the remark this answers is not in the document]
codebien: @Blinkuu maybe it's my misunderstanding, but my interpretation of `(2)` was that it would drop the requirement for the flag so if a user does `k6 cloud script.js` and it is importing the `tracing` module then we should enable this proxy of the metadata. If this is not required, then I agree with @mstoykov if the user has to explicit the flag there is no reason for doing more.

@mstoykov I'm not sure I'm following, do you mean we should enable the check of the metadata for all scripts that the cloud run? If not, then how do we decide when to enable it or not for the cloud runs without `(3)`?

cloudapi/insights/client.go
```diff
@@ -0,0 +1,206 @@
...
+	}
+
+	req, err := newBatchCreateRequestMetadatasRequest(requestMetadatas)
+	if err != nil {
+		return fmt.Errorf("failed to create request from request metadatas: %w", err)
+	}
+
+	// TODO(lukasz, retry-support): Retry request with returned metadatas.
+	//
+	// Note: There is currently no backend support backing up this retry mechanism.
```
codebien, line 123: What does it happen without it? I guess it just loses the batch. Is it fine?

cloudapi/insights/proto/v1/common/common.proto
```diff
@@ -0,0 +1,50 @@
+syntax = "proto3";
+
+package k6.cloud.insights.proto.v1.common;
+
+option go_package = "go.k6.io/k6/cloudapi/insights/proto/v1/common";
+
+// Messages below are copied from https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/common/v1/common.proto
```
codebien, line 7: Is this copy legit? Are we respecting the license?

cloudapi/insights/proto/v1/k6/labels.proto
codebien, line 1: Is there a specific reason for this structure with tiny files? Having all of them in one `k6.proto` / `trace.proto` sounds more readable to me.

output/cloud/expv2/output.go
```diff
@@ -213,6 +286,7 @@ func (o *Output) handleFlushError(err error) {
 
 	o.logger.WithError(err).Warn("Interrupt sending metrics to cloud due to an error")
 
+	// TODO(lukasz): This is not thread-safe
```
codebien, line 289: Master rebase required. Setting mostly as a reminder.

vendor/modules.txt
```diff
@@ -398,6 +398,7 @@ google.golang.org/grpc/serviceconfig
 google.golang.org/grpc/stats
 google.golang.org/grpc/status
 google.golang.org/grpc/tap
+google.golang.org/grpc/test/bufconn
```
codebien, line 401: Internally we have [HTTPMultibin](https://github.com/grafana/k6/blob/master/lib/testutils/httpmultibin/httpmultibin.go), which should do a similar logic. Can you check if it could be a replacement so we can save from adding a new dependency?

codebien (changes requested):

output/cloud/expv2/output.go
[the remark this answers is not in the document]
mstoykov: > Ideally, we would like to enable the request metadata output logic if all 3 conditions are met:

Given this sentence my only opinion is taht (2) is not really necessary for us to not have this enabled most of the times. 

*BUT*:
1. It probably shouldn't be limitted based on 1 module given that users can generate their own `trace_id` metadata as well after this release
2. Even if it is enabled for a test that does not have metadata - I doubt there is any performance problem to be going through all the samples and checking if there is metadata on them.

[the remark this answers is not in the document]
Blinkuu: The idea behind `K6_CLOUD_TRACES_ENABLED` is to hide `k6 x Tempo` behind a feature flag until it's stable enough we can make it `true` by default. Hence I brought up (2). If you think this is unnecessary, let's skip this and revisit it later if we see big enough utilization spikes. I would like to keep the `TODO` for now tho.

output/cloud/expv2/collect.go
Blinkuu: What about the `scenario` tag?

cloudapi/insights/proto/gen.go
Blinkuu: Done in [369297e](https://github.com/grafana/k6/pull/3100/commits/369297e10e1cecea4516cf5a3a7809528c550ed2).

js/modules/k6/experimental/tracing/trace_id.go
Blinkuu: Fixed in [8860d2f](https://github.com/grafana/k6/pull/3100/commits/8860d2f4ca90adba808d7fdafeedf4165adea123).

cloudapi/insights/client.go
```diff
@@ -0,0 +1,206 @@
...
+		opts = append(opts, grpc.WithTransportCredentials(insecure.NewCredentials()))
+	} else {
+		if cfg.TLSConfig.CertFile != "" {
+			creds, err := credentials.NewClientTLSFromFile(cfg.TLSConfig.CertFile, "")
+			if err != nil {
+				return nil, fmt.Errorf("failed to load TLS credentials from file: %w", err)
+			}
+			opts = append(opts, grpc.WithTransportCredentials(creds))
+		} else {
+			opts = append(opts, grpc.WithTransportCredentials(credentials.NewTLS(&tls.Config{MinVersion: tls.VersionTLS12})))
```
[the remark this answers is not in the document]
Blinkuu: Updated in [8939be1](https://github.com/grafana/k6/pull/3100/commits/8939be10135ea4901ddab0fd5bbbfa5755683981).

cloudapi/insights/client.go
```diff
@@ -0,0 +1,206 @@
...
+
+	if cfg.TLSConfig.Insecure { //nolint: nestif
+		opts = append(opts, grpc.WithTransportCredentials(insecure.NewCredentials()))
+	} else {
+		if cfg.TLSConfig.CertFile != "" {
+			creds, err := credentials.NewClientTLSFromFile(cfg.TLSConfig.CertFile, "")
+			if err != nil {
+				return nil, fmt.Errorf("failed to load TLS credentials from file: %w", err)
+			}
+			opts = append(opts, grpc.WithTransportCredentials(creds))
```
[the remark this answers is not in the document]
Blinkuu: This is useful if we ever have to deal with certificates coming from untrusted CAs.

output/cloud/expv2/collect.go
```diff
@@ -163,3 +175,104 @@ func (c *collector) timeFromBucketID(id int64) int64 {
...
+func (c *requestMetadatasCollector) CollectRequestMetadatas(sampleContainers []metrics.SampleContainer) {
+	if len(sampleContainers) < 1 {
+		return
+	}
+
+	// TODO(lukasz, other-proto-support): Support grpc/websocket trails.
+	httpTrails := c.filterTrailsWithTraces(sampleContainers)
+	if len(httpTrails) > 0 {
+		c.collectHTTPTrails(httpTrails)
+	}
```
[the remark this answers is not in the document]
Blinkuu: Done in [6db7e33](https://github.com/grafana/k6/pull/3100/commits/6db7e338a5980a724216499f2619897ec2ea6ef7).

vendor/modules.txt
Blinkuu: IMO it's quite more complex compared to `bufconn`. I'd rather stick with what is officially supported.

output/cloud/expv2/output.go
Blinkuu: Done. Also, I removed the TODO now.

cloudapi/insights/proto/v1/common/common.proto
Blinkuu: Not sure. I'm not a lawyer. Do we have somebody with more expertise in licensing that could chip in here?

Blinkuu: I added a copyright notice in [27137a9](https://github.com/grafana/k6/pull/3100/commits/27137a9a909b9b670d2be0c62554ee8c3d3cc73a). AFAIK, this should be enough, but perhaps someone with more knowledge could double-check it.

cloudapi/insights/client.go
```diff
@@ -0,0 +1,206 @@
...
+	return nil
+}
+
+// IngestRequestMetadatasBatch ingests a batch of request metadatas.
+func (c *Client) IngestRequestMetadatasBatch(ctx context.Context, requestMetadatas RequestMetadatas) error {
+	c.connMu.RLock()
+	if c.conn == nil {
+		return ErrClientClosed
+	}
+	c.connMu.RUnlock()
```
codebien, line 110:
```suggestion
	closed := c.conn == nil
	c.connMu.RUnlock()
	if closed {
			return ErrClientClosed
	}
```

cloudapi/insights/client.go
```diff
@@ -0,0 +1,206 @@
...
+
+	if len(requestMetadatas) < 1 {
+		return nil
+	}
+
+	req, err := newBatchCreateRequestMetadatasRequest(requestMetadatas)
+	if err != nil {
+		return fmt.Errorf("failed to create request from request metadatas: %w", err)
```
codebien, line 118:
```suggestion
		return fmt.Errorf("failed to create request from request metadata: %w", err)
```
Metadata doesn't have the plural, am I missing something? I see it is used all over the code so if we change then keep it consistent.

cloudapi/insights/client.go
codebien, line 110: Is it the rest of the logic fault tolerant in the case the connection closes after it?

output/cloud/expv2/output.go
```diff
@@ -19,6 +21,18 @@ import (
...
+type RequestMetadatasCollector interface {
+	CollectRequestMetadatas([]metrics.SampleContainer)
+	PopAll() insights.RequestMetadatas
+}
+
+// Flusher is an interface for flushing data to the cloud.
+type Flusher interface {
+	Flush(ctx context.Context) error
+}
+
```
codebien, line 35:
```suggestion
// RequestMetadatasCollector is an interface for collecting request metadatas
// and retrieving them so they can be flushed using a Flusher.
type requestMetadatasCollector interface {
	CollectRequestMetadatas([]metrics.SampleContainer)
	PopAll() insights.RequestMetadatas
}

// Flusher is an interface for flushing data to the cloud.
type flusher interface {
	flush(ctx context.Context) error
}

```
I'm going to open a PR for unexporting all the methods/interfaces/fields not required to be exported. So let's not add more. It sounds they are used only internally to the current package.

output/cloud/expv2/output.go
```diff
@@ -112,13 +161,21 @@ func (o *Output) StopWithTestError(testErr error) error {
 	default:
 	}
 
-	// Drain the SampleBuffer and force the aggregation for flushing
+	// Drain the SampleBuffer
+	o.collectSamples()
+
+	// Force the aggregation for flushing
```
codebien, line 167: A bad rebase? This operation is already called two lines below

output/cloud/expv2/output.go
```diff
@@ -231,3 +304,24 @@ func (o *Output) handleFlushError(err error) {
...
+
+func (o *Output) tracingEnabled() bool {
+	// TODO(lukasz): Check if the `tracing` module is imported.
+	//
+	// We want to check if the `tracing` module was imported. If
+	// it wasn't, we don't need to collect any data related to tracing
+	// as there won't be any.
+	//
+	// We currently don't have a way to check if the tracing module
+	// was imported by the user.
```
codebien, line 316: Based on the previous discussion it is not required anymore, right?
```suggestion
```

output/cloud/expv2/output.go
Blinkuu: Fixed in [dd1741f](https://github.com/grafana/k6/pull/3100/commits/dd1741f49285edbdabd289356b360932dc862beb).

output/cloud/expv2/output.go
Blinkuu: Indeed. Fixed in [7c41939](https://github.com/grafana/k6/pull/3100/commits/7c4193982f295285ced4a2198a122eb74d299492).

output/cloud/expv2/output.go
Blinkuu: Removed in [19fa719](https://github.com/grafana/k6/pull/3100/commits/19fa7196d032785f5c5a52eb6c83755f98e8e090).

cloudapi/insights/client.go
Blinkuu: Yes, we are aware it doesn't have a plural form. However, together with @vortegatorres we decided to add the `s` suffix to match the conventions defined in [Google API Design Guide](https://cloud.google.com/apis/design). I'd like to keep this as "request metadatas".

cloudapi/insights/client.go
Blinkuu: Done in [3f7e461](https://github.com/grafana/k6/pull/3100/commits/3f7e46144848cb2ce43edc26d3d8aa6ac4e92d6d).

Blinkuu: I added logic to close the client in [d5ec11e](https://github.com/grafana/k6/pull/3100/commits/d5ec11ef7a1310811ef3219d63800965ed31e6cd).

In practice, we should never hit the `ErrClientClosed` if the client is used properly.

codebien:
Hey @Blinkuu, it seems there is a conflict, can you rebase it, please?

output/cloud/expv2/output.go
```diff
@@ -19,8 +21,16 @@ import (
...
+// and retrieving them, so they can be flushed using a flusher.
+type requestMetadatasCollector interface {
+	CollectRequestMetadatas([]metrics.SampleContainer)
+	PopAll() insights.RequestMetadatas
+}
+
+// flusher is an interface for flushing data to the cloud.
 type flusher interface {
-	flush(context.Context) error
+	flush(ctx context.Context) error
```
codebien, line 33:
```suggestion
	flush(context.Context) error
```
I don't think we need it the type already documents it clearly.

js/modules/k6/experimental/tracing/client.go
```diff
@@ -7,6 +7,7 @@ import (
 	"time"
 
 	"github.com/dop251/goja"
+
```
codebien, line 10:
```suggestion
```
I don't think it is necessary

output/cloud/expv2/output.go
```diff
@@ -100,12 +117,49 @@ func (o *Output) Start() error {
...
+
+		ctx, cancel := context.WithTimeout(context.Background(), 1*time.Second)
+		defer cancel()
+		if err := insightsClient.Dial(ctx); err != nil {
+			return err
+		}
+
+		o.shutdownHooks = append(o.shutdownHooks, func() error {
+			return insightsClient.Close()
+		})
```
codebien, line 149: We are storing a state here, why not use the classic way of storing it in a field and then use the Stop function following the defined lifecycle from the Output interface?

cloudapi/insights/proto/v1/trace/span.proto
```diff
@@ -0,0 +1,72 @@
...
+  string TraceID = 1;
+
+  // A unique identifier for a span within a trace.
+  string SpanID = 2;
+
+  // The start time of the span.
+  sfixed64 StartTimeUnixNano = 3;
+
+  // The end time of the span.
+  sfixed64 EndTimeUnixNano = 4;
```
codebien, line 20: Out of curiosity, why are we not using the Timestamp type?

output/cloud/expv2/output.go
```diff
@@ -218,6 +284,21 @@ func (o *Output) flushMetrics() {
...
+// flushRequestMetadatas periodically flushes traces collected in RequestMetadatasCollector using flusher.
+func (o *Output) flushRequestMetadatas() {
+	start := time.Now()
+
+	ctx, cancel := context.WithTimeout(context.Background(), o.config.TracesPushInterval.TimeDuration())
+	defer cancel()
+
+	err := o.requestMetadatasFlusher.flush(ctx)
+	if err != nil {
+		o.logger.WithError(err).Error("Failed to push trace samples to the cloud")
```
codebien, line 296:
```suggestion
		o.logger.WithError(err).WithField("t", time.Since(start)).Error("Failed to push trace samples to the cloud")
```
In case of error it could even more relevant

output/cloud/expv2/output.go
```diff
@@ -260,6 +341,18 @@ func (o *Output) handleFlushError(err error) {
 	})
 }
 
+func (o *Output) tracingEnabled() bool {
+	// TODO(lukasz): Check if k6 x Tempo is enabled
+	//
+	// We want to check whether a given organization is
+	// eligible for k6 x Tempo feature. If it isn't, we don't
+	// want to enable the traces output.
```
codebien, line 349:
```suggestion
	// eligible for k6 x Tempo feature. If it isn't, we may
	// consider to skip the traces output.
```
At this point, it could be not so relevant to skip them, so let's try it first then we will have a more accurate decision.

output/cloud/expv2/flush_test.go
```diff
@@ -96,3 +160,79 @@ func (pm *pusherMock) push(_ string, _ *pbcloud.MetricSet) error {
...
+	require.NoError(t, err)
+	require.False(t, cli.ingestRequestMetadatasBatchInvoked)
+	require.False(t, cli.dataSent)
+	require.Empty(t, cli.data)
+}
+
+func Test_tracesFlusher_Flush_ReturnsNoErrorWithWorkingInsightsClientAndNonCancelledContextAndData(t *testing.T) {
+	t.Parallel()
+
+	// Given
```
codebien, line 186: Out of curiosity, why do you think this practice is helpful? 

If they work for you, I'm not against keeping them, but in my opinion, a unit test should try to achieve some level of auto-explanation and I think a test like the current already does.

The constructors are grouped, the logic method is only one, and the assert section is also grouped.

codebien:
Hey @mstoykov, I don't think we have an issue with the timeout here because the flush doesn't sound concurrent, but you may have another opinion regarding the timeout lifespan.

output/cloud/expv2/output.go
Blinkuu: You're right. Simplified in [de2257a](https://github.com/grafana/k6/pull/3100/commits/de2257a6d0b07cfd2972a724b14354f9917d503c).

output/cloud/expv2/flush_test.go
Blinkuu: At this point, it's muscle memory. It helps me when I revisit some bigger unit tests.

codebien (dismissed):
LGTM :bow: 

EDIT: Of course, we need all tests green

output/cloud/expv2/output.go
```diff
@@ -19,6 +21,14 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+// requestMetadatasCollector is an interface for collecting request metadatas
+// and retrieving them, so they can be flushed using a flusher.
+type requestMetadatasCollector interface {
+	CollectRequestMetadatas([]metrics.SampleContainer)
+	PopAll() insights.RequestMetadatas
+}
```
mstoykov, line 29: nit: I don't particularly see why you need a interface here :shrug: 

I expect this is to facilitate testing only?

cloudapi/config_test.go
```diff
@@ -28,16 +28,19 @@ func TestConfigApply(t *testing.T) {
 		ProjectID:                       null.NewInt(1, true),
 		Name:                            null.NewString("Name", true),
 		Host:                            null.NewString("Host", true),
+		Timeout:                         types.NewNullDuration(1*time.Minute, true),
```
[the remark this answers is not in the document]
Blinkuu: Good catch; it was by accident. Fixed in [3d2b256](https://github.com/grafana/k6/pull/3100/commits/3d2b25624f05b45b0a09de5a239ced4b6fa834ba).

cloudapi/insights/domain.go
```diff
@@ -0,0 +1,103 @@
...
+		return errEmptyTestRunID
+	}
+
+	if rm.TestRunLabels.Scenario == "" {
+		return errEmptyTestRunScenario
+	}
+
+	if rm.TestRunLabels.Group == "" {
+		return errEmptyTestRunGroup
+	}
```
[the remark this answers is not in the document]
Blinkuu: All the values must be present, this is how our backend works. The other part of the code should fill these gaps with "unknown" if not set.

output/cloud/expv2/collect.go
```diff
@@ -163,3 +175,94 @@ func (c *collector) timeFromBucketID(id int64) int64 {
...
+	c.buffer = nil
+	return b
+}
+
+func (c *rmCollector) getStringTagFromTrail(trail *httpext.Trail, key string) string {
+	if tag, found := trail.Tags.Get(key); found {
+		return tag
+	}
+
+	return "unknown"
```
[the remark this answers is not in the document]
Blinkuu: We rely on this being non-empty, at least for the time being.

cloudapi/insights/mappers.go
```diff
@@ -0,0 +1,71 @@
...
+
+	return &ingester.BatchCreateRequestMetadatasRequest{
+		Requests: reqs,
+	}, nil
+}
+
+func newCreateRequestMetadataRequest(requestMetadata RequestMetadata) (*ingester.CreateRequestMetadataRequest, error) {
+	if err := requestMetadata.Valid(); err != nil {
+		return nil, fmt.Errorf("invalid request metadata: %w", err)
+	}
```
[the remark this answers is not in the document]
Blinkuu: You're right. I was overly secure with this one. We do this validation in the backend anyway. Removed in [6a63b82](https://github.com/grafana/k6/pull/3100/commits/6a63b82e2f20d6fc4a83d0c04608697e7b0d939a).

output/cloud/expv2/collect.go
```diff
@@ -163,3 +175,94 @@ func (c *collector) timeFromBucketID(id int64) int64 {
...
+				ID:       c.testRunID,
+				Scenario: c.getStringTagFromTrail(trail, scenarioTag),
+				Group:    c.getStringTagFromTrail(trail, groupTag),
+			},
+			ProtocolLabels: insights.ProtocolHTTPLabels{
+				URL:        c.getStringTagFromTrail(trail, nameTag),
+				Method:     c.getStringTagFromTrail(trail, methodTag),
+				StatusCode: c.getIntTagFromTrail(trail, statusTag),
+			},
+		}
```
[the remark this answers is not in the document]
Blinkuu: I tried a few things, and IMO, it makes the code less readable. If we introduce support for other protocols in the future, I will update this code anyway.

output/cloud/expv2/output.go
Blinkuu: Yeah, this is currently unused. Do you want me to remove this interface?

Blinkuu: Yeah, this is currently unused. Do you want me to remove this interface?

cloudapi/insights/domain.go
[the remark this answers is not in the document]
mstoykov: > The other part of the code should fill these gaps with "unknown" if not set.

But `group` can have (and does) an empty value for metrics emitted outside of a group. 

So this validation (which is now removed) would've dropped all traces for request done outside of a group.

And if you do that on backside you will also do it.

The function that sets it to `unknown` only does it if it is *missing* not if the value is empty.

Which ... seems just as well to in practice be done on the backend. If you see an empty `group` - set it to unknwon there. Instead of setting it to `unknown` from here.

Specifically here `unknown` kind of ... "not allowed" value for `group` ... but even if it was I would not make a "magical" string be what I mean when a field is missing. The empty value seems just as well, especially if you don't want to differntiate between it and the thing not being set.

mstoykov (dismissed):
LGTM in general!

 I can't really comment on the protocol/wire format and stuff and so have just ignored it :grimacing: 

I would still prefer if you just set the values directly from what we have and if needed either:
1. not add the request data at all to be send
2. drop it in the backend.

I would like to remind you that this will be a public endpoint, so you should definitely do some defensive programming as other people might send you whatever they want.

output/cloud/expv2/output.go
codebien: If it is unused, then we should remove it. I don't see the point in having more abstraction, if not required

output/cloud/expv2/pbcloud/metric.pb.go
```diff
@@ -1,6 +1,6 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
-// 	protoc-gen-go v1.28.1
+// 	protoc-gen-go v1.30.0
```
codebien, line 3: I don't think we need this change here

output/cloud/expv2/output.go
Blinkuu: As discussed in private, it's used in testing. Leaving as is.

output/cloud/expv2/pbcloud/metric.pb.go
Blinkuu: Done in [a95ba44](https://github.com/grafana/k6/pull/3100/commits/a95ba44757fc39ec406624dd8eb1ade7c1c43dff).

codebien (approved):

Blinkuu:
Thanks for the help @codebien and @mstoykov! Looking forwarding to having this merged 🚀

codebien:
@Blinkuu apparently there is a change conflict, can you fix it, please?

codebien (approved):

PR #3102, opened 2023-05-30
Forbid async shared array

mstoykov:


olegbespalov (dismissed):

js/modules/k6/data/data.go
```diff
@@ -61,6 +61,11 @@ func (d *Data) Exports() modules.Exports {
 	}
 }
 
+// TODO potentially try to DRY with this with a similar message in the `k6` js module.
+// This likely won't be needed as SharedArray should likely support async function at some point ... somehow.
+const asyncFunctionNotSupportedMsg = "SharedArray constructor does not support async functions as second argument, " +
+	"please see https://k6.io/docs/javascript-api/k6/group/#working-with-async-functions for more info"
```
imiric, line 67: This links to the `group()` documentation. Should we have a central place in the docs where we address async functions, and where they're supported? This could be part of a larger "JavaScript support in k6" section we've discussed in the past, but we could start with this first.

imiric (dismissed):

js/modules/k6/data/data.go
mstoykov: Hm, I just copie ths while trying to figure things below and then update the first part. 

But to be honest the reason for SharedArray is because there is no async API that can be (and make sense to)  be used in a SharedArray at this point. 

And supporting it actually doing it will be more complicated and in. practice will either requiring `SharedArray` returning a promise in some cases or having an `AsyncSharedArray` .

So I wonder if I shouldn't just drop the url at all. There is no real *relevant* info - it just doesn't make sense at this point and this error is better then the generic one.

js/modules/k6/data/data.go
```diff
@@ -61,6 +61,11 @@ func (d *Data) Exports() modules.Exports {
 	}
 }
 
+// TODO potentially try to DRY this with a similar message in the `k6` js module.
+// This likely won't be needed as SharedArray should likely support async function at some point ... somehow.
+const asyncFunctionNotSupportedMsg = "SharedArray constructor does not support async functions as second argument, " +
+	"please see https://k6.io/docs/javascript-api/k6/group/#working-with-async-functions for more info"
```
mstoykov, line 67:
```suggestion
const asyncFunctionNotSupportedMsg = "SharedArray constructor does not support async functions as second argument"
```

?

oleiade (approved):
👏🏻

js/modules/k6/data/data.go
imiric: OK, sure, let's drop the URL then.

imiric (approved):

olegbespalov (approved):

PR #3104, opened 2023-05-30
output/cloudv2: Retry and flushers pool

codebien:
It adds two features:

- HTTP requests retry
- Pool of flushers so that if a flush operation is slow, the output can continue and not get stuck.

### DIFF
--- a/output/cloud/expv2/flush.go
+++ b/output/cloud/expv2/flush.go
@@ -8,7 +8,7 @@ import (
 )
 
 type pusher interface {
-	push(ctx context.Context, referenceID string, samples *pbcloud.MetricSet) error
+	push(referenceID string, samples *pbcloud.MetricSet) error
 }
 
 type metricsFlusher struct {
@@ -19,10 +19,10 @@ type metricsFlusher struct {
 	maxSeriesInSingleBatch     int
 }
 
-// Flush flushes the queued buckets sending them to the remote Cloud service.
-// If the number of time series collected is bigger than maximum batch size than
-// it splits in chunks.
-func (f *metricsFlusher) Flush(ctx context.Context) error {
+// flush flushes the queued buckets sending them to the remote Cloud service.
+// If the number of time series collected is bigger than maximum batch size
+// then it splits in chunks.
+func (f *metricsFlusher) flush(_ context.Context) error {
 	// drain the buffer
 	buckets := f.bq.PopAll()
 	if len(buckets) < 1 {
@@ -46,7 +46,7 @@ func (f *metricsFlusher) Flush(ctx context.Context) error {
 		}
 
 		// we hit the chunk size, let's flush
-		err := f.client.push(ctx, f.referenceID, msb.MetricSet)
+		err := f.client.push(f.referenceID, msb.MetricSet)
 		if err != nil {
 			return err
 		}
@@ -58,7 +58,7 @@ func (f *metricsFlusher) Flush(ctx context.Context) error {
 	}
 
 	// send the last (or the unique) MetricSet chunk to the remote service
-	return f.client.push(ctx, f.referenceID, msb.MetricSet)
+	return f.client.push(f.referenceID, msb.MetricSet)
 }
 
 type metricSetBuilder struct {
--- a/output/cloud/expv2/metrics_client.go
+++ b/output/cloud/expv2/metrics_client.go
@@ -5,64 +5,45 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"io"
 	"net/http"
-	"sync"
-	"time"
+	"strings"
 
 	"github.com/klauspost/compress/snappy"
-	"github.com/sirupsen/logrus"
 	"google.golang.org/protobuf/proto"
 
 	"go.k6.io/k6/cloudapi"
-	"go.k6.io/k6/lib/consts"
 	"go.k6.io/k6/output/cloud/expv2/pbcloud"
 )
-
-type httpDoer interface {
-	Do(*http.Request) (*http.Response, error)
-}
 
 // metricsClient is a Protobuf over HTTP client for sending
 // the collected metrics from the Cloud output
 // to the remote service.
 type metricsClient struct {
-	httpClient httpDoer
-	logger     logrus.FieldLogger
-	token      string
-	userAgent  string
-
-	pushBufferPool sync.Pool
-	baseURL        string
+	httpClient *cloudapi.Client
+	baseURL    string
 }
 
 // newMetricsClient creates and initializes a new MetricsClient.
-func newMetricsClient(logger logrus.FieldLogger, host string, token string) (*metricsClient, error) {
-	if host == "" {
-		return nil, errors.New("host is required")
-	}
-	if token == "" {
-		return nil, errors.New("token is required")
+func newMetricsClient(c *cloudapi.Client) (*metricsClient, error) {
+	// Unfortunately, the cloudapi.Client works across different versions
+	// of the API, but it has the v1 harcoded so we need to trim the wrong path
+	// to be able to replace it with the correct one.
+	u := c.BaseURL()
+	if !strings.HasSuffix(u, "/v1") {
+		return nil, errors.New("a /v1 suffix is expected in the Cloud service's BaseURL path")
 	}
 	return &metricsClient{
-		httpClient: &http.Client{Timeout: 5 * time.Second},
-		logger:     logger,
-		baseURL:    host + "/v2/metrics/",
-		token:      token,
-		userAgent:  "k6cloud/v" + consts.Version,
-		pushBufferPool: sync.Pool{
-			New: func() interface{} {
-				return &bytes.Buffer{}
-			},
-		},
+		httpClient: c,
+		baseURL:    strings.TrimSuffix(u, "/v1") + "/v2/metrics/",
 	}, nil
 }
 
-// Push pushes the provided metrics the given test run.
-func (mc *metricsClient) push(ctx context.Context, referenceID string, samples *pbcloud.MetricSet) error {
+// Push the provided metrics for the given test run ID.
+func (mc *metricsClient) push(referenceID string, samples *pbcloud.MetricSet) error {
 	if referenceID == "" {
 		return errors.New("TestRunID of the test is required")
 	}
-	start := time.Now()
 
 	b, err := newRequestBody(samples)
 	if err != nil {
@@ -73,30 +54,25 @@ func (mc *metricsClient) push(ctx context.Context, referenceID string, samples *
 	// we don't expect to share this client across different refID
 	// with a bit of effort we can find a way to just allocate once
 	url := mc.baseURL + referenceID
-	req, err := http.NewRequestWithContext(ctx, http.MethodPost, url, bytes.NewReader(b))
+	req, err := http.NewRequestWithContext(
+		context.Background(), http.MethodPost, url, io.NopCloser(bytes.NewReader(b)))
 	if err != nil {
 		return err
 	}
 
-	req.Header.Set("User-Agent", mc.userAgent)
+	req.GetBody = func() (io.ReadCloser, error) {
+		return io.NopCloser(bytes.NewReader(b)), nil
+	}
+
 	req.Header.Set("Content-Type", "application/x-protobuf")
 	req.Header.Set("Content-Encoding", "snappy")
 	req.Header.Set("K6-Metrics-Protocol-Version", "2.0")
-	req.Header.Set("Authorization", "Token "+mc.token)
 
-	resp, err := mc.httpClient.Do(req)
+	err = mc.httpClient.Do(req, nil)
 	if err != nil {
 		return err
 	}
-	defer func() {
-		_ = resp.Body.Close()
-	}()
 
-	if err := cloudapi.CheckResponse(resp); err != nil {
-		return err
-	}
-	mc.logger.WithField("t", time.Since(start)).WithField("size", len(b)).
-		Debug("Pushed the collected metrics to the Cloud service")
 	return nil
 }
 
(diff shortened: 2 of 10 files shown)


output/cloud/expv2/metrics_client.go
```diff
@@ -49,11 +57,7 @@ func newMetricsClient(logger logrus.FieldLogger, host string, token string) (*me
 		baseURL:    host + "/v2/metrics/",
 		token:      token,
 		userAgent:  "k6cloud/v" + consts.Version,
-		pushBufferPool: sync.Pool{
-			New: func() interface{} {
-				return &bytes.Buffer{}
-			},
-		},
+		waitFor:    func(d time.Duration) { <-time.After(d) },
```
mstoykov, line 60:
```suggestion
		waitFor:    func(d time.Duration) { time.Sleep(d) },
```

output/cloud/expv2/metrics_client.go
```diff
@@ -73,33 +77,73 @@ func (mc *metricsClient) push(ctx context.Context, referenceID string, samples *
...
+
+	return err
+}
+
+func (mc *metricsClient) retryableDo(req *http.Request, attempt int) (retry bool, err error) {
+	resp, err := mc.httpClient.Do(req)
+
+	defer func() {
+		if resp != nil {
+			_ = resp.Body.Close()
```
mstoykov, line 128: You should also read it in full before you close it

output/cloud/expv2/output.go
```diff
@@ -122,6 +126,37 @@ func (o *Output) StopWithTestError(testErr error) error {
...
+		go func() {
+			defer func() {
+				t.Stop()
+				o.wg.Done()
+			}()
+
+			for {
+				select {
+				case <-t.C:
+					o.flushMetrics()
```
mstoykov, line 149: Because of how `flush` is implemented. This in most cases will lead to *one* worker doing the work either way.

`flush` gets *all* the metrics each time it is called and then splits them if necessary.
So with multiple workers they will all try to get *all* the metric series and only *one* will succeed.

If that worker manages to flush them in the Push Interval - it will just come back to wait for this ticker.

This will still help if flush takes longer then the push interval *IF* the actual context used for that is actually not the pushInterval - so in practice this will never use more then 1 worker.

I am very much for:
1. increasing the context and timeouts for this push *significantly*. There actually is a cloudAPI default timeout which is 1minute. 
2. The timeout must be reset every time we retry. Currently, it doesn't so in practice if it times outs it will timeout again and again. 
3. Preferably something will every push interval trigger, get all the time series, split them if necessary *and* then give it to the workers so they can "only" do the marshalling, encoding and sending.

(3) can likely be skipped as the first 2 will likely help with that a lot.

https://github.com/grafana/k6/blob/9254ce7ddafe856ea58dea46f54049a8248fcfe1/output/cloud/expv2/output.go#L177C1-L191 
for the flushMetric code and showcasing where it has a context with the timeout

https://github.com/grafana/k6/blob/9254ce7ddafe856ea58dea46f54049a8248fcfe1/output/cloud/expv2/flush.go#L26-L31 

the `flush` code that drains the whole buckets


As an additional note, because we are using the same `http.Client` underneath if this is actually `http2` there are chances that we will be hitting the same ingest server over and over again, which might not be what we want.

But this will likely need more investigation and is a problem even for cloud output v1, so probably a thing we can do much later

output/cloud/expv2/output_test.go
```diff
@@ -278,3 +280,87 @@ func TestOutputStopWithTestError(t *testing.T) {
...
+	flusherMock := func() {
+		// it asserts that flushers are set and the flush is invoked
+		once.Do(func() { close(o.stop) })
+	}
+
+	o.flushing = flusherFunc(flusherMock)
+	o.runFlushWorkers()
+
+	// it asserts that all flushers exit
+	o.wg.Wait()
```
imiric, line 336: Could we add a timeout to both this test and the one below, to ensure that if this doesn't work, the test doesn't hang indefinitely?

output/cloud/expv2/metrics_client.go
```diff
@@ -18,6 +18,14 @@ import (
 	"go.k6.io/k6/output/cloud/expv2/pbcloud"
 )
 
+const (
+	// linearRetryIntervalFactor is the factor for the cloud request retry interval
+	retryIntervalFactor = 500 * time.Millisecond
+
+	// maxRetries specifies max retry attempts
+	maxRetryAttempts = 3
```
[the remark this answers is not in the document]
codebien: Done

output/cloud/expv2/output.go
```diff
@@ -122,6 +126,37 @@ func (o *Output) StopWithTestError(testErr error) error {
...
 
+func (o *Output) runFlushWorkers() {
+	t := time.NewTicker(o.config.MetricPushInterval.TimeDuration())
+
+	// TODO: drop it when the migration from v1 is completed
+	workers := 4
+	if o.config.MetricPushConcurrency.Int64 > 1 &&
+		o.config.MetricPushConcurrency.Int64 < 10 {
+		workers = int(o.config.MetricPushConcurrency.Int64)
+	}
```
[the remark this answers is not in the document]
codebien: Done

output/cloud/expv2/metrics_client.go
codebien: Done

output/cloud/expv2/metrics_client.go
codebien: Done

output/cloud/expv2/output.go
```diff
@@ -73,6 +73,17 @@ func (o *Output) SetTestRunStopCallback(stopFunc func(error)) {
...
+		o.config.MetricPushConcurrency.Int64 > 10 {
+		// TODO: the remote service sets this option with a very high value
+		// using the config overwriting feature. It is required for v1 so for reusing the
+		// same option here, it is necessary to force a smaller default value
+		// or use another the provided  if it is in acceptable range.
+		//
+		// Drop the following hack when the migration from v1 is completed.
+		o.config.MetricPushConcurrency = null.IntFrom(4)
+	}
+
```
mstoykov, line 86: I am not certain I follow this. 

AFAIK there are no problems with this being 1 or with this being very large. 

goroutines are cheap and even with 1 concurrency this will likely work pretty okay.

So I prefer that we only do this if there is actually a reason to do this.

The backend can change what it returns ... probably based on the version (in the headers) in the original request if we want to have different values.

output/cloud/expv2/metrics_client.go
```diff
@@ -44,15 +41,17 @@ type metricsClient struct {
...
+) (*metricsClient, error) {
 	if host == "" {
 		return nil, errors.New("host is required")
 	}
 	if token == "" {
 		return nil, errors.New("token is required")
 	}
 	return &metricsClient{
-		httpClient: &http.Client{Timeout: 5 * time.Second},
+		httpClient: &http.Client{Timeout: timeout},
```
mstoykov, line 54: Is there a particular reason you can't use the clouadpi client?

It does all of the relevant parts and already sets this timeout. 

I am also pretty sure we have an instance in the wrapping `output` struct that is used to create the test run if needed.

output/cloud/expv2/output.go
codebien: I would like to test with values close to what should be the potential final value. And having `20` as final numbers sounds a magic number to me.

>The backend can change what it returns 

I would avoid it, the cost is not balanced, so let's use 20 for now.

output/cloud/expv2/metrics_client.go
codebien: The main reason was the JSON payload + headers but checking better I've found a way for work-around them.

output/cloud/expv2/flush.go
```diff
@@ -19,10 +19,10 @@ type metricsFlusher struct {
...
 }
 
-// Flush flushes the queued buckets sending them to the remote Cloud service.
-// If the number of time series collected is bigger than maximum batch size than
-// it splits in chunks.
-func (f *metricsFlusher) Flush(ctx context.Context) error {
+// flush flushes the queued buckets sending them to the remote Cloud service.
+// If the number of time series collected is bigger than maximum batch size
+// then it splits in chunks.
+func (f *metricsFlusher) flush(_ context.Context) error {
```
mstoykov, line 25: Why keep the context if we are not going to use it?

output/cloud/expv2/output.go
```diff
@@ -231,3 +258,29 @@ func (o *Output) handleFlushError(err error) {
...
+		"maxMetricSamplesPerPackage": c.MaxMetricSamplesPerPackage.Int64,
+		"metricPushConcurrency":      c.MetricPushConcurrency.Int64,
+		"metricPushInterval":         c.MetricPushInterval.String(),
+	}
+
+	tokenValue := "***"
+	if !c.Token.Valid {
+		tokenValue = ""
+	}
+	m["token"] = tokenValue
```
mstoykov, line 283: nitpick: 

```suggestion
		"metricPushInterval":         c.MetricPushInterval.String(),
		"token":                                 "",	
	}

	if c.Token.Valid {
		m["token"]  = "***"
	}
```

output/cloud/expv2/metrics_client.go
```diff
@@ -18,47 +18,39 @@ import (
...
+) (*metricsClient, error) {
 	if host == "" {
 		return nil, errors.New("host is required")
 	}
 	if token == "" {
 		return nil, errors.New("token is required")
 	}
 	return &metricsClient{
-		httpClient: &http.Client{Timeout: 5 * time.Second},
+		httpClient: cloudapi.NewClient(logger, token, host, "v"+consts.Version, timeout),
```
mstoykov, line 43: Can we just get this from the `output/cloud/output.go` ? And not need to create a new one.

output/cloud/expv2/flush.go
codebien: Because https://github.com/grafana/k6/pull/3100#discussion_r1228403404 is going to use it

mstoykov: It seems to me they will have exactly the same problem ... but let's see I guess :shrug:

output/cloud/expv2/output.go
mstoykov: Part of https://github.com/grafana/k6/issues/3117 

Under > "reevaluate the current periodic and abort signal architecture/interaction"

output/cloud/expv2/metrics_client.go
```diff
@@ -5,98 +5,69 @@ import (
...
 
 // newMetricsClient creates and initializes a new MetricsClient.
-func newMetricsClient(logger logrus.FieldLogger, host string, token string) (*metricsClient, error) {
-	if host == "" {
-		return nil, errors.New("host is required")
+func newMetricsClient(c *cloudapi.Client, testRunID string) (*metricsClient, error) {
+	// Unfortunately, the cloudapi.Client works across different versions
+	// of the API, but it has the v1 harcoded so we need to trim the wrong path
+	// to be able to replace it with the correct one.
+	u := c.BaseURL()
```
mstoykov, line 32: Maybe this should be done in the cloudapi package?

As a separate PR?

mstoykov (approved):
LGTM :clap: :tada:

output/cloud/expv2/metrics_client.go
codebien: I thought the same but I would avoid the load with a new PR to review for you folks. You're asking it so let's do it #3125

mstoykov (dismissed):

output/cloud/expv2/metrics_client.go
```diff
@@ -73,33 +77,73 @@ func (mc *metricsClient) push(ctx context.Context, referenceID string, samples *
...
 }
 
+func (mc *metricsClient) do(r *http.Request) error {
+	var (
+		err   error
+		retry = true
+	)
+
+	for attempts := 1; retry; attempts++ {
+		retry, err = mc.retryableDo(r, attempts)
```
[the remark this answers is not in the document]
imiric: I suppose the answer to this is "no", since the v2 output now uses the `cloudapi.Client`, which [also doesn't care about this](https://github.com/grafana/k6/blob/9254ce7ddafe856ea58dea46f54049a8248fcfe1/cloudapi/client.go#L100-L112), and it's been that way since it was written in 2017.

output/cloud/expv2/metrics_client.go
```diff
@@ -5,64 +5,46 @@ import (
...
+		httpClient: c,
+		baseURL:    strings.TrimSuffix(u, "/v1") + "/v2/metrics/",
 	}, nil
 }
 
-// Push pushes the provided metrics the given test run.
-func (mc *metricsClient) push(ctx context.Context, referenceID string, samples *pbcloud.MetricSet) error {
+// Push pushes the provided metrics the given test run. Inject a timeout context
+// for limiting the request duration.
+func (mc *metricsClient) push(referenceID string, samples *pbcloud.MetricSet) error {
```
imiric, line 44: There's no context argument anymore, so let's remove the comment.

```suggestion
// Push the provided metrics for the given test run ID.
func (mc *metricsClient) push(referenceID string, samples *pbcloud.MetricSet) error {
```

imiric (approved):
LGTM :+1:

Good that this is now using the `cloudapi.Client` :+1:

output/cloud/expv2/metrics_client.go
codebien: Done

imiric: There were some suggestions for the first sentence there as well, but OK.

imiric (dismissed):

output/cloud/expv2/metrics_client.go
codebien: It should be applied now

mstoykov (approved):

PR #3110, opened 2023-06-02
Update to v0.10.0 of xk6-browser

ankur22:
This will bring k6 up to date with the latest "release" of the browser module that we want to bring to the attention of our users.


ankur22:
Will start over again against an up to date master branch.

PR #3112, opened 2023-06-02
Add event system for JS modules

### LINKED ISSUE #2432: Hooks for extensions - e.g. graceful shutdown, per VU/iteration, etc.

imiric:
This is a simple event system that JS modules written in Go can use to subscribe to test execution events they're interested in, and notify the event emitter when the event processing is done. In practice this is required for some modules like k6-browser so that they can start and stop browser processes depending on the execution status, but it could also be useful to any module that needs to do some cleanup. We could also expose it to outputs or any other extension, and eventually to JS scripts as well.

Currently these events are emitted:
- `Init`: when k6 starts initializing outputs, VUs and executors.
- `TestStart`: when the execution scheduler starts running the test.
- `TestEnd`: when the test execution ends.
- `IterStart`: when a VU starts an iteration.
- `IterEnd`: when a VU ends an iteration.
- `Exit`: when the k6 process is about to exit.

We can discuss if more are needed, but from my discussion with the k6-browser team, these should fulfill their requirements.

Note that this is **not** meant to be part of v0.45.0, so we should merge it after the release.

Closes #2432


cmd/tests/events/events.go
```diff
@@ -0,0 +1,74 @@
...
+			for {
+				select {
+				case evt, ok := <-evtCh:
+					if !ok {
+						return
+					}
+					var testStatus string
+					if evt.Type != event.Exit {
+						status, _ := api.Status(vu.Context())
+						testStatus = fmt.Sprintf(", test status: %s", status.Status.String())
```
imiric, line 55: I've noticed that this is flaky. :disappointed:

It can happen that the test status has changed between the time the event is received, and this API call returns, so the current test assertions aren't stable.

Any alternative ideas with how to match up k6 execution status with the events we receive here?

cmd/tests/cmd_run_test.go
```diff
@@ -1998,3 +2000,99 @@ func TestBadLogOutput(t *testing.T) {
...
+	// addresses need to be different for separate tests, we register separate
+	// modules to workaround it. I experimented with passing the address via
+	// the script, but the sync.Once is still a problem.
+	// Note that this also forbids running these tests with more than -count=1,
+	// unless we dynamically increment the module name... /sigh
+	modules.Register("k6/x/testevents1", events.New(
+		ts.GlobalState.DefaultFlags.Address, []event.Type{
+			event.Init, event.TestStart, event.IterStart, event.IterEnd,
+			event.TestEnd, event.Exit,
+		}))
```
imiric, line 2021: Any ideas for improving this?

mstoykov: Since some of the last changes in `js` package the module system gets teh modules it should support 

https://github.com/grafana/k6/blob/cf90835d199e8f9c89e9621f33a6e655dab94b04/js/bundle.go#L110

This is not exposed so that you can create a runner wtih *specific* modules. 

At this point I feel like this is not a bad idea, but also that likely we should do some moving around of the current `modules` registration system ... once again.

I don't really like how one of the modules are in one place and the other in another and are being merged strangely.

A quicker fix might be for the Runner to get only the "non-standard"/extension modules. But again this likely will be too big of a change.

I am mostly writing to ask you to actually do teh hack with the ever increasing index so that `-count 2` can run as I always find it very strange and frustrating when tests suddenly stop working with it and I am using it for some bug/race teasing.

An example of the hack: (whcih I think I can remove in that place :sweat_smile: ) 

https://github.com/grafana/k6/blob/cf90835d199e8f9c89e9621f33a6e655dab94b04/js/init_and_modules_test.go#L38-L44

js/runner.go
```diff
@@ -765,6 +766,16 @@ func (u *ActiveVU) RunOnce() error {
 	ctx, cancel := context.WithCancel(u.RunContext)
 	defer cancel()
 	u.moduleVUImpl.ctx = ctx
+
+	eventIterData := event.IterData{
+		Iteration:    uint64(u.iteration),
+		VUID:         u.ID,
+		ScenarioName: u.scenarioName,
+	}
+	if u.Runner.preInitState.Events != nil {
```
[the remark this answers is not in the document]
imiric: > The biggest change seems to be to execution package where no such helper exists and there are >10 uses.

Not really. It's used much more in `js` tests:

```
$ rg -g '*_test.go' -c 'TestPreInitState\{'
execution/scheduler_int_test.go:1
execution/scheduler_ext_exec_test.go:4
execution/scheduler_ext_test.go:3
cmd/runtime_options_test.go:2
converter/har/converter_test.go:1
js/init_and_modules_test.go:2
js/modules/k6/marshalling_test.go:1
js/runner_test.go:28
js/console_test.go:1
js/module_loading_test.go:11
js/bundle_test.go:1
js/modules/k6/html/html_test.go:1
js/share_test.go:1
js/modules/k6/metrics/metrics_test.go:3
lib/executor/common_test.go:1
api/v1/group_routes_test.go:1
metrics/engine/ingester_test.go:1
metrics/engine/engine_test.go:3
```

Many of them pass it to `NewFromArchive`, which in some cases doesn't matter if they don't run `VU.RunOnce()`, but most do. Since they check the behavior of creating the archive from the first runner, calling `getSimpleRunner()` wouldn't help in these cases.

We could create a separate helper that all these tests call, and just do the change there, but... it's a lot of tedious work. :sweat: I started doing something like that, but gave up and just added the nil check, hoping we could settle on a better solution. If we can't think of anything, I'll consider refactoring the tests... soon-ish.

> Also FYI you should be able to do this check in the Emit call before you use the nil instance of the Events.

I guess so, but that's just moving it to a less relevant place. I'd rather see it here where we know it can be nil, then in `Emit` where it mostly shouldn't be.

event/type.go
```diff
@@ -0,0 +1,33 @@
+package event
+
+// Type represents the different event types emitted by k6.
+type Type string
```
olegbespalov, line 4: Could it be a non-string type?

event/system.go
```diff
@@ -0,0 +1,153 @@
...
+// unsubscribing.
+type Subscriber interface {
+	Subscribe(events ...Type) (subID uint64, eventsCh <-chan *Event)
+	Unsubscribe(subID uint64)
+}
+
+// System keeps track of subscribers, and allows subscribing to and emitting
+// events.
+type System struct {
+	ctx         context.Context
```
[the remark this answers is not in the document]
imiric: Yeah, not even `Emit()`, since it's only used in the function returned from `Emit()`, so we can pass it there. I wanted to keep the signatures simple, but sure, I'll remove it from the struct.

event/system.go
```diff
@@ -0,0 +1,153 @@
...
+// UnsubscribeAll closes all event channels and removes all subscriptions.
+func (s *System) UnsubscribeAll() {
+	s.subMx.Lock()
+	defer s.subMx.Unlock()
+
+	seenSubs := make(map[uint64]struct{})
+	for _, sub := range s.subscribers {
+		for subID, evtCh := range sub {
+			if _, ok := seenSubs[subID]; !ok {
+				close(evtCh)
```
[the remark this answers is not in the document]
imiric: But we do. That's the point of the `seenSubs` map and this check. We use a map instead of a bool here because we're dealing with more than one subscriber ID.

event/type.go
imiric: Why?

I considered making it an enum, but then we'd have to use enumer and `go generate` and that whole mess, and it really makes no difference in this case. We just need something unique and stringifiable, and plain string consts work fine IMO.

cmd/run.go
```diff
@@ -300,9 +319,21 @@ func (c *cmdRun) run(cmd *cobra.Command, args []string) (err error) {
...
 
+	// TODO: Make the timeout configurable?
+	if werr := waitInitDone(10 * time.Second); werr != nil {
+		logger.WithError(werr).Warn()
+	}
+	test.preInitState.Events.Emit(&event.Event{Type: event.TestStart})
 	// Start the test! However, we won't immediately return if there was an
 	// error, we still have things to do.
 	err = execScheduler.Run(globalCtx, runCtx, samples)
+	waitTestEndDone := test.preInitState.Events.Emit(&event.Event{Type: event.TestEnd})
```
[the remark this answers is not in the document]
imiric: Possibly, but this error would also be sent in the `Exit` event, so it could be duplicated. I think the final error we return from `run()` is more valuable to external modules, since that also includes the `AbortReason` and `ExitCode`. So instead of subscribers checking both errors and handling these cases, they can only check the `Exit` one instead.

`Exit` is also guaranteed to be sent, since it's done in the `defer`, whereas this event could be skipped if an error happens before and we don't get to actually run the test.

cmd/tests/events/events.go
```diff
@@ -0,0 +1,74 @@
...
+					}
+					logger.Infof("got event %s with data '%+v'%s", evt.Type, evt.Data, testStatus)
```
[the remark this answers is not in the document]
imiric: This is a test module, and not used by k6 users. This is logged as `Info` since this output is read by the test. I suppose it could be lower severity, but we'd need to change [the `LoggerHook` level in `NewGlobalTestState()`](https://github.com/grafana/k6/blob/8f6f73b17f441054090f6fb8957b94d8ef551bad/cmd/tests/test_state.go#L50) as well, and practically there wouldn't be a difference. So I'd rather keep it as is, if you don't mind.

cmd/tests/events/events.go
imiric: @olegbespalov It's not just for purposes of logging. This ensures that the event is sent at the right time. I.e. if we receive the `Init` event, that k6 is also in the `InitVUs` stage.

I think this is very valuable, as otherwise the events could be sent at any point, and we have no way of confirming that they represent the actual state in k6. Unfortunately, it's also flaky, and I don't think we have another way of confirming this, other than sending it in the event `Data` as I mentioned above, which wouldn't be ideal.

So if we can't find an alternative, I'll remove it, since we definitely shouldn't have more flaky tests. :sweat_smile:

event/type.go
olegbespalov: Just because something like `uint8` can hold all variations of the types and has a smaller memory footprint, remember that the primary use case of using this is inside the go code.

I fully agree with you that following the current approach with enums sometimes looks scary (e.g. https://github.com/grafana/k6/blob/master/lib/compatibility_mode_gen.go), and just implementing `Stringer` method is enough.

But for sure, not forcing, but just sharing my opinion :relaxed:

event/system.go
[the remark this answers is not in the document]
olegbespalov: Right :facepalm:

cmd/tests/events/events.go
olegbespalov: Yeah, you're right here. And since it's a testing module (which I missed), it's not about logging. 

Regarding:
>EDIT: Crazy idea: continuously query the API for the status in the background with somewhat aggressive frequency, and log whatever the last status we received was. I suspect the biggest reason for the flakiness is the time taken to do this network call. But then again, we could always get the event at the "wrong" time and miss the actual test status, so this might be flaky as well...

Still sounds more flaky than your previous idea of just checking the order

cmd/tests/cmd_run_test.go
mstoykov: > So we need to return a local map, similarly as in getInternalJSModules(), but first we should get rid of the registration in init() pattern all extensions use, which is at the crux of the issue

the `init()` pattern has nothing to do with this. 

Even if we were using something else I doubt the js.Runner would've been written to get it from outside back then to begin with. If it is not taking it from the outside - it will still get it in some "global" way that will mean that you won't control it. 

And that is the problem. Removign `init()` will just break all extensions and will (IMO) make it more likely we will break them again and again in the future.

The curretn solution that I see that won't be that bad is for `js.Runner#New` (and NewArchive) to take `map[string]interface{}` with the custom modules and propage it through the code. Instead of it being taken from `getInternalJSModules()`


The rest of the comment - I don't understand:
1. at a call to `New()` or `NewModulesInstnace()` the options are not completed - so getting access to them is pointless, it won't help you in most cases.
2. `since you were against changing New()` that is because it is not part of the API. It is just a convention. And at the time it is called there is nothing you can do with it. If you want to make it part of the API you will basically make a completely new API.

event/type.go
imiric: Fair point, but I'm not certain if memory usage is a concern. Even with 2 events emitted per iteration, events are either consumed, or dropped if there are no consumers or the event channel buffer is full. So at most the size of the channel buffer will be kept in memory, everything else will be discarded.

But sure, it's not much work to switch to an enum, and it would be safer, so I'll do that.

cmd/tests/events/events.go
imiric: > Still sounds more flaky than your previous idea of just checking the order

It is. But like I said, the order tells us nothing about _when_ the events were emitted, so we lose an important assertion. But if there's no deterministic alternative, I'll remove it.

event/system.go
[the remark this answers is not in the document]
imiric: I removed it in 8658f68b9.

Since it doesn't make sense to pass both a `time.Duration` and a context to the wait function, I also removed the `time.Duration`, as we can pass a context that times out, which serves the same purpose.

event/system.go
```diff
@@ -0,0 +1,145 @@
...
+		var doneCount int
+		for {
+			if doneCount == totalSubs {
+				close(doneCh)
+				return nil
+			}
+			select {
+			case <-doneCh:
+				doneCount++
+			case <-ctx.Done():
```
olegbespalov, line 105: :thinking: since we derive from the `lingerCtx` could it be a constant error when somebody uses `Ctrl + C` un case of test end, I remember that the previous version of the code just returns without error when `ctx.Done`

event/system.go
```diff
@@ -0,0 +1,145 @@
...
+// eventBuffer determines the size of the Event channel buffer. Events might be
+// dropped if this buffer is full and there are no event listeners, or if events
+// are emitted very quickly and the event handler goroutine is busy. It is
+// recommended to handle events in a separate goroutine to not block the
+// listener goroutine.
+func NewEventSystem(eventBuffer int, logger logrus.FieldLogger) *System {
+	return &System{
+		subscribers: make(map[Type]map[uint64]chan *Event),
+		eventBuffer: eventBuffer,
+		logger:      logger, // TODO: Log some debug events
```
olegbespalov, line 38: Sounds good to write a debug log when a public API method is using (`Subscribe`, `Emit`, `Unsubscribe`, etc)

event/system.go
imiric: Yeah, you're right.

I wasn't sure whether `lingerCtx` is the correct base in this case, and we can't use `runCtx` since that can be aborted before waiting for events to be processed. I changed the base to `globalCtx` in c8b17b83b, and that works fine from my tests with `--linger` (it doesn't return an error). I think this should depend on a higher-level context, and not an entirely separate one, to ensure it can exit cleanly if needed.

As for returning without an error, I think that should only be done if we receive the signal that all events were processed. Previously it would silently exit if the context is done, but it should've raised a warning instead.

Now we don't have that distinction since we reuse the context timeout itself, but I think this is simpler, and we don't to handle a separate timeout.

olegbespalov (approved):
LGTM :+1:

js/runner.go
```diff
@@ -765,6 +766,16 @@ func (u *ActiveVU) RunOnce() error {
...
+		u.Runner.preInitState.Events.Emit(&event.Event{Type: event.IterStart, Data: eventIterData})
```
imiric, line 776: @ka3de Is the plan still to initialize browsers on the `IterStart` event?

We currently don't block to wait for the event to be processed here before running the iteration, since that could add a delay not just to the first iteration, but all.

Would it be possible for you to initialize browsers on the `Init` event, on which we do wait for? It's meant for that purpose, but I'm not sure if this aligns with your current approach.

ka3de: Hey @imiric , apologies for the delay, we (me or someone else from the team) still have pending to review this PR properly.

> Is the plan still to initialize browsers on the IterStart event?

That is still the plan, as we are currently using this _"1 browser process per iteration"_ model. I think initializing everything on the `Init` event would not be ideal, as that, with the current model, would mean to start up _N_ browsers initially, when most of these will be "unused" until the corresponding iteration for which it was allocated starts.

Nevertheless I understand non blocking on the `IterStart` event processing and I think we would be able to work around this from our side anyway.

Any thoughts on this @ankur22 , @inancgumus ?

ankur22: If i remember correctly, there's no way to retrieve the `scenario.options.browser` options when any of these events are fired, since the only way currently to retrieve these options is from the `vu` when an API is called. If that's the case then @ka3de has outlined everything already 👍

ka3de: > there's no way to retrieve the `scenario.options.browser` options when any of these events are fired

Ok, that's something that I overlooked, because that means that in the future, when we support more than one browser type, we won't be able to start the browser on `IterStart` event as we don't have access to that setting at that time.
In any case, by now this is an improvement and the biggest dependency was with `IterEnd`/`Exit` as this will allow us to do the cleanup.

imiric: Alright, sounds good.

Ankur is right, but you _can_ access `vu.State.Options` on `IterStart` and `IterEnd`, since the `vu.State` is accessible at that point. Technically, the options are available to k6 when we emit `Init` as well, and a hack could be sending them as part of the `Init` event `Data`. This might not be such a hack after all, if we consider all `Init` subscribers to require options for initialization. :thinking:

This would literally be a one-line change to do, and would avoid any complicated refactoring to pass options in a dedicated API. We don't need to do this now, just pointing out that it's a possibility. FYI @mstoykov

ka3de (commented):
Hey @imiric , thanks a lot for this!
I've reviewed the PR focusing only on whether it fulfills the k6 browser use case requirements or not.
Overall it LGTM. I have only one consideration similar to the one discussed in [this thread](https://github.com/grafana/k6/pull/3112#discussion_r1221758386):

¿Would it be possible to extend the current `IterData` struct to also send the scenario options? In that case we would have direct access to the browser settings for the scenario, otherwise we'll have to keep a reference to the `k6modules.VU` passed as input to the `NewModuleInstance` call, is that right?

imiric:
@ka3de We can certainly send the options in `IterStart`, for example, but a few things concern me about that:

- We shouldn't _just_ send scenario options, as the event system isn't meant to be used only by k6-browser. In order to keep it generic, we should send all options, but then...

- Sending all options might be too resource intensive. The `Iter*` events are emitted per iteration, and there could be thousands of these events. I tried to limit the `IterData` struct to the absolute essentials.

  We could send this only on the first iteration, but...

- Using the `Iter*` events to send options feels more like a hack than sending them in `Init`. At least `Init` is meant for initialization, and it makes sense for options to be part of the data.


> otherwise we'll have to keep a reference to each of the k6modules.VU passed as input to the NewModuleInstance call, is that right?

Don't you already keep this reference? You store it in the context, and then retrieve it with [`GetVU()`](https://github.com/grafana/xk6-browser/blob/92ac6b70f71c29f7a16fd629c233c7b87f104515/k6ext/context.go#L25-L35), no?

I think you'll need to devise a system with a goroutine that listens for k6 events, dispatches whatever process you need, like starting or stopping browser processes, and stores the mapping of scenario name to WS URL. Then each VU will look up the scenario it's in, and retrieve the WS URL it needs to connect to from the map.

At least that's how I'm picturing it, but you guys would know better.

ka3de:
> We can certainly send the options in IterStart, for example, but a few things concern me about that:

Yep, I also thought about probably having to make this more generic, and therefore how it could potentially get out of control in terms of data sent. So yes, completely understand these concerns, they make sense.

> Using the `Iter*` events to send options feels more like a hack than sending them in `Init`. At least `Init` is meant for initialization, and it makes sense for options to be part of the data.

So we would get all options definition in `Init` including all scenario definitions and their options. Is that right? That would work for us as we could pre initialize an index for `scenario - browser type`, as a scenario is limited to a single browser type. If this makes sense for k6, it would simplify things for us.

> Don't you already keep this reference? You store it in the context, and then retrieve it with [GetVU()](https://github.com/grafana/xk6-browser/blob/92ac6b70f71c29f7a16fd629c233c7b87f104515/k6ext/context.go#L25-L35), no?

Yes, we do, but that is in the context of an already initialized browser instance, as it's in scope only for a single iteration. But in this case, ideally the events listening will be done through a single goroutine for the whole test run, that is the one in charge to initialize the browser instances.


On another note. Coming back to [this comment](https://github.com/grafana/k6/pull/3112#discussion_r1221758386), I think I said "too happily" that we could work around this, but, if there is no wait for the `Done()` function to be called in the `IterStart` event, there is indeed a race condition between the event processing and the iteration execution, which basically makes the event not very "dev-friendly" usable in order to setup something that is required for the iteration execution. I guess you can implement some sort of polling to actually wait for whichever action is triggered from the event processing, but that seems trickier.

I don't have any knowledge on the `js/runner.go` implementation, where the `IterStart` event seems to be generated, but are there any alternatives on which the iteration start would actually wait for the `IterStart` event `Done()` method to be called without blocking all other iterations?

js/runner.go
```diff
@@ -765,6 +766,16 @@ func (u *ActiveVU) RunOnce() error {
 	ctx, cancel := context.WithCancel(u.RunContext)
 	defer cancel()
 	u.moduleVUImpl.ctx = ctx
+
+	eventIterData := event.IterData{
+		Iteration:    uint64(u.iteration),
```
ka3de, line 771: Out of curiosity, why use `uint64` for `Iteration` field in `IterData` if the same field in `State` is represented as `int64`? For what I know, `Iteration` is an unsigned monotonic int, so I guess it makes sense to use `uint64`, but would it be better to be coherent with the previous implementation?

No big deal in any case, but when integrating with these events I find myself having to convert the `State` iteration from `int64` to `uint64` numerous times :sweat_smile:

imiric: Yeah, it's better to be consistent than correct in this case, so I'll change it.

imiric: Fixed in the last push.

ka3de:
As discussed, in our initial experimentation, from k6-browser perspective, trying to integrate with the event system implemented in this PR, we encountered an issue that I will try to explain:

As per our current model in k6-browser, we start and stop a browser instance per iteration. This is the reason why our main motivation was to listen to the `IterStart` and `IterEnd` events in order to perform these actions.

To initialize the browser we need the `VU` [instance](https://github.com/grafana/k6/blob/2e5631feff27a14ca721d40a4997361fcd859f11/js/modules/modules.go#L41-L65) because through this one we have to access and parse the scenario options and other elements that are used for the initialization and the browser instance implementation for the iteration.

This `VU` is only made accessible to JSModules in the `NewModuleInstance` [call](https://github.com/grafana/xk6-browser/blob/4fdce45873f1d872e7582ea8271b9f9bda42bbe0/browser/module.go#L57), which happens once initially (due to initial requirements analysis I believe) and then once per every `VU` in the test. At this point we have to store that `VU` somehow, so we can keep a reference to it for when an `IterStart` event for that `VU` comes, so we can use it to initialize the browser. But the problem is that, at the time of the `NewModuleInstance` call, the `VU` is in "init" state, which means that we can not access the `VU.State()` and therefor we can not know to which scenario it belongs to or which VUID it "represents". Which eventually makes it not possible to initialize browsers on `IterStart` events because, considering that the event system is "global", it sends iter events for all iterations, and we don't have a way to link these events with the input `VU`s in `NewModuleInstance`.

**For that reason, the proposal is to either:**
- Not have a global implementation for the event system, and tight this to a specific VU context. In the sense that, when subscribing through the `VU.Events`, this subscription would be valid only for the events for that VU.
  In this case we'll need to clarify how the `Init` and `End` events would be delivered, as this are global per se.
- Keep the current global implementation for the event system, but allow to supply a `Filter` parameter in the subscription API that will allow to subscribe only for that specific `VU` which "subscribe" action is called through. Notice that in this case, no `VUID` could be used as input for that filter, as we are in the same situation, the subscriber, in this case k6-browser implementation of JSModule, does not have access to the `VUID` when the module instance is initialized.

imiric:
Hey @ka3de, please take a look at the latest changes.

The event system is now split into a global one that emits the `Init`, `TestStart`, `TestEnd`, and `Exit` events, and a local per-VU one that emits the `IterStart` and `IterEnd` events.

I experimented with the filter approach, and making all events local to the VU, but they were more complex than just splitting it.

You would still need to subscribe to VU events in `NewModuleInstance()`, or wherever you do the module initialization, if you want to catch the first `IterStart` event, but `vu.Events().Local` will now only notify you of events for that specific VU.

We now also wait for `IterStart` and `IterEnd` to complete processing, which will give you time to start and stop the browser processes.

The split has made the test a bit flaky, which is why the CI is currently failing. I'm looking into fixing that, but this shouldn't impact the functionality, and you can start testing.

Let me know if you have any other issues with this.

cmd/tests/cmd_run_test.go
```diff
@@ -1534,7 +1537,7 @@ func TestMinIterationDuration(t *testing.T) {
 	elapsed := time.Since(start)
 	assert.Greater(t, elapsed, 7*time.Second, "expected more time to have passed because of minIterationDuration")
 	assert.Less(
-		t, elapsed, 14*time.Second,
+		t, elapsed, 15*time.Second,
```
imiric, line 1540: FYI waiting on `IterStart` and `IterEnd` does have a non-negligible impact on iteration duration, even if the events aren't being handled. I've seen this test fail at least once after this change with the duration being ~14.1s.

Maybe we can further optimize the waiting function, but ideally, we shouldn't introduce waits in such a hot path, since it this impacts the overall performance of k6. So I would suggest we start looking into ways to do the browser initialization on the `Init` event, and shutdown on `TestEnd`/`Exit` during this cycle.

js/runner.go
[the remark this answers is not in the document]
imiric: I removed the nil checks in f7588d616, since this now uses the local per-VU event system, which is guaranteed to be defined here.

js/runner.go
imiric: I added waiting for `IterStart` and `IterEnd` to be processed in a97bdd057, but as mentioned [here](https://github.com/grafana/k6/pull/3112#discussion_r1237004171), this has an impact on performance.

Let's continue the discussion in that thread.

ka3de:
> The event system is now split into a global one that emits the Init, TestStart, TestEnd, and Exit events, and a local per-VU one that emits the IterStart and IterEnd events.

Thanks for this @imiric ! I have taken a brief look and it seems to fulfill our needs. I will take a better look at it tomorrow.
One question though. What are the "guarantees" for `IterEnd` and `Exit` events in case of error? Is `IterEnd` guaranteed to be sent in case of unexpected error, panic etc? Or only `Exit` is on these cases? Thanks!

js/modules/modules.go
```diff
@@ -42,6 +42,10 @@ type VU interface {
 	// Context return the context.Context about the current VU
 	Context() context.Context
 
+	// Events allows subscribing to global k6 execution events, such as Init and
+	// Exit, and to local (per-VU) events, such as IterStart and IterEnd.
+	Events() common.Events
```
olegbespalov, line 47: One concern that I have in my mind (but unfortunately not a proposal to resolve it) is that we expose the events (both `global` and `local`) under the VU, where logically and from the context here should be only the `local` (VU-related), but access to the global (Test Run-related) seems to be somewhere else. Have you considered this? :thinking:

event/system.go
```diff
@@ -0,0 +1,170 @@
...
+		logger:      logger,
+	}
+}
+
+// Subscribe to one or more events. It returns a subscriber ID that can be
+// used to unsubscribe, and an Event channel to receive events.
+// It panics if events is empty.
+func (s *System) Subscribe(events ...Type) (subID uint64, eventsCh <-chan *Event) {
+	if len(events) == 0 {
+		panic("must subscribe to at least 1 event type")
```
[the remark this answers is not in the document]
imiric: Because this is a programming error. Calling `Subscribe()` with 0 events is invalid. I'm only making it fail explicitly to avoid unnecessary locking, and everything that happens below.

js/modules/modules.go
imiric: I had a comment somewhere about this, but I must've deleted it.

The thing is that there's no way to pass "global" objects to JS modules. The only access to k6 they have is via this `VU` interface, since effectively every module instance is tied to a single VU.

So I agree with you; global objects don't belong here. I think there should be another way for modules to receive things like this and consolidated options that's not tied to a single VU, but to the global module itself. I experimented with passing these in some `ModuleState` object in the module constructor (e.g. `browser.New()`), but options are not consolidated at that point yet. We could pass these static global objects that way, though, and reuse the `Init` event to pass options, for example, but it's just an idea, and we haven't discussed it yet. Or we could expose another global API modules could access from... somewhere.

So, yeah, this would need some design discussion. Currently I see no way around this, other than removing the global event system, and using only the local per-VU one, but that has other complications...

cmd/tests/cmd_run_test.go
ankur22: Just wondering whether it would be possible to negate the time that k6 is waiting on `IterStart` and `IterEnd` from the iteration duration or related metrics as a short term solution? Or is there a bigger impact that just increasing the iteration duration?

I like the long term goal with working with `Init`, and `TestEnd`/`Exit`.

imiric:
@ka3de:

> What are the "guarantees" for `IterEnd` and `Exit` events in case of error? Is `IterEnd` guaranteed to be sent in case of unexpected error, panic etc? Or only `Exit` is on these cases?

Great question! We mentioned this briefly during the last meeting, but I forgot to follow-up on it.

The previous `TestEventSystemAbort` checked the behavior when the script is aborted in the `default` function, but it didn't subscribe to VU events, and we didn't test the other failure scenarios.

So in d5897faac I renamed it to `TestEventSystemError`, and added a couple of more failure tests. Take a look. `TestEventSystemError/init` checks the events emitted when there's an error in the init context, meaning k6 never gets to run the `default` function. For this to work I moved the emission of the `Exit` event before we even parse the script for the first time, which ensures that `Exit` is always emitted.

So to answer your question: neither `IterEnd` nor `TestEnd` are guaranteed to be emitted in some failure scenarios, but `Exit` is. So make sure to always listen to it and handle any leftover cleanup there. If there's no error in the init context, then you would also get an error in `IterEnd`, which could be a signal to do any cleanup there as well.

(BTW, please comment on a line in the diff, so that we can resolve any discussions when they're done, to avoid cluttering the main PR.)

cmd/tests/cmd_run_test.go
imiric: @ankur22 That sounds way too hacky for my taste :sweat_smile: It would effectively be faking metrics to compensate for the fact that k6 is now taking longer to run, but it wouldn't solve the actual problem of slower execution.

Besides, as you can see from this test, the issue is not with the value of the iteration duration metric, but with the elapsed time of the test. The metrics are unaffected by this, since we measure the actual iteration time in the `VU.runFn` method. But overall, the iteration rate would be slower, which might add up to considerably lower throughput after this change. We should benchmark it to be sure, but I think this is a blocker for merging this, and we should prioritize that change in how k6-browser handles browser processes.

ankur22: > But overall, the iteration rate would be slower, which might add up to considerably lower throughput after this change.

:thumbsup:

> We should benchmark it to be sure, but I think this is a blocker for merging this, and we should prioritize that change in how k6-browser handles browser processes.

Is the benchmarking the blocker or this issue with the current implementation? If the current implementation is the blocker than that will affect the release, and we need to work on it straight away 😬

imiric: Yeah, I think this implementation is a blocker. But maybe @olegbespalov or @mstoykov have a different opinion.

ka3de: Could these test failures have been outliers? I was testing this in a silly way, basically running the test _N_ times and did not see significant differences with the `main` version.

olegbespalov: @imiric I have the same opinion as you :+1:

js/modules/modules.go
olegbespalov: Also, let me stress that I'm not considering this as a blocker as long as we explicitly (with the comment maybe) mention that this API could change since we see a "separation of concerns" issue and users know that they could use "global" events keeping in mind that this API could change.

event/system.go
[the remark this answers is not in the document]
olegbespalov: That's correct, but I still missing why this is a panic and not an error, which also could prevent everything that happens bellow.

TL;DR, I'm asking why not:
```diff
- func (s *System) Subscribe(events ...Type) (subID uint64, eventsCh <-chan *Event) {
- 	if len(events) == 0 {
-		panic("must subscribe to at least 1 event type")
+ func (s *System) Subscribe(events ...Type) (subID uint64, eventsCh <-chan *Event) error {
+	if len(events) == 0 {
+		return errors.New("must subscribe to at least 1 event type")
```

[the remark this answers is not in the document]
inancgumus: It might be similar to [this](https://cs.opensource.google/go/go/+/refs/tags/go1.20.5:src/strings/replace.go;l=26-37). To keep the API more straightforward to use.

I've been just following this PR from time to time. You can ignore me :)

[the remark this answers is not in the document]
imiric: Exactly, thank you Inanc :bow:

This is a programming, and not an application or user error. So returning `error` instead would just be annoying to use.

If Go allowed a build-time check for this case, I would prefer that instead, but alas. :disappointed:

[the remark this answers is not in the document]
olegbespalov: I'm still of the opinion that there could be cases where a user of this API is deciding if they should or not should process after catching of error from there. And catching the panic isn't the best pattern vs. just an error. But keeping in mind the risk of happening such cases prefer just to resolve the tread :sweat:

cmd/tests/cmd_run_test.go
imiric: I added some benchmarks in 89fa05664. See the results [here](https://gist.github.com/imiric/b2094c79f45cbdbdcb067671434fde4a).

In summary:
- There's a very minor difference in performance between `master` (97d40e939) and this branch (d5897faac) when a module that uses events isn't imported; `17.32ms ± 1%` vs `17.34ms ± 1%`. But according to `benchstat` it's not statistically significant, so it might not be concerning, and the `TestMinIterationDuration` failure I experienced was probably a fluke(?).

  There is a minor overhead in allocations and memory usage, but it's not too bad.

- Understandably, there is a substantial impact to performance when a module that does use events is imported. The benchmarks aren't directly comparable, but it's around +~68%.

I didn't benchmark importing the `events` module, but not subscribing to `Iter*` events. Would this be interesting?

The more realistic test would be running [k6-benchmarks](https://github.com/grafana/k6-benchmarks), but that would take more time, and I'm not sure if we need it.

So I'm not sure what to do with these results. Importing the browser module would have an impact on performance anyway, so whatever the overhead is from event processing is minor in comparison to everything else.

But at the same time, `vu.RunOnce()` is a very sensitive place to add any delays in, and I'm hesitant about adding even event emission without waiting. So I still lean towards k6-browser moving to doing initialization on the `Init` event and shutdown on `TestEnd`/`Exit` rather than on the `Iter*` events. This would be the desired end-goal anyway, and we wouldn't run risks of impacting k6 performance.

ka3de: > I added some benchmarks in https://github.com/grafana/k6/commit/89fa056643ef9468f193e4f5446abdb786ded956. See the results [here](https://gist.github.com/imiric/b2094c79f45cbdbdcb067671434fde4a).

Thanks for this @imiric !

> I didn't benchmark importing the events module, but not subscribing to Iter* events. Would this be interesting?

If I'm understanding this correctly, I would say no, as, as you said, if a module is using the events, the impact of the overhead of the events system implementation is probably negligible in comparison with the actual work that is done as a reaction to the events. I think the critical point was understanding the impact of the event system implementation for a test that does not use the event system at all. Still, the benchmarks are done with a void iteration, which makes sense to isolate and understand the impact of the event system, but if we consider that usually an iteration in k6 would include at least 1 HTTP request, which would imply ~100-200ms, the impact in that context is "reduced". I understand this is what benchstat indicates with "not statistically significant".

All in all, I think waiting or not for the `IterStart` and `IterEnd` events, or even supporting these events at all, is a k6-core decision which should consider also the usage of these events for other extensions or for the JS API also, and how should it behave in these cases.
From the k6 browser side we will try to analyze what would be the effort to move the browser initialization to the `Init` event, but this will require proper testing and analysis also on the performance when being shared by multiple VUs, on sharing the WS connection or not, etc.

mstoykov: Sorry for the slow reply - I didn't manage to get to writing this last week :(.

Let's start with some presumptions/ideas about this PR and change at this point in time.

We are doing this almost entirely to support browser in their need for functionality within k6. It happens that this somewhat aligns with a more generic feature we want that isn't (well) fleshed out - events.
The PR while based on the idea is not really expected to be usable for other extensions or internal features apart from browser. Having this expectations raises (IMO) the complexity quite a lot. 
This also means that IMO this change is supposed to test how this will work in particular case while supporting it for the time being.
That is to say that we don't need it to be perfect and we will likely change it.

Browser has stipulated they need the following things to be possible:
1. to be signalled when an iteration ends, so they can disconnect/stop browser.
Also, they would like to:
1. Know when to start/connect to a browser. 
2. When k6 ends execution, so they can make certain they disconnect/stop all the browser if any left. 

A point here is that starting as many browsers as needed for the test at the beginning depends on the understanding of "how many" are needed. Which requires some calculations as scenarios might not overlap and as such k6 will reuse VUs between them and in theory browser can be reused as well. For some cases this calculations and internal implementation might be easy (a scenario starting 5 minutes after another stops and bot using 10 browsers). But for others this will be a bit harder - 2 ramping VUs scenarios that have differing peaks - one going 0 to 5 every minute , 1 going 5 to 0 every minute. 

To be perfectly honest at this point I expect that the later example would not even work all that well now that k6 execution is a lot more complex. Even before event loop waiting we in practice could've ended up waiting on a websocket (for example) connection to send end message throwing the things out of order.

A browser implementation will also need to continuously "keep tabs" on whether a browser is in use in this case - effectively needing at least `iterEnd`.  As otherwise ran iteration can end and the browser module can still think that a browser instance is assigned to a VU, but it might never start.

IMO browser team should decide what they need and how it will work, and we should go with that, unless it is detrimental for tests *not using* browser. Again my expectation is that this will change at least once more trying to make stuff work and the internal implementations will likely change even more overtime.

> Understandably, there is a substantial impact to performance when a module that does use events is imported. The benchmarks aren't directly comparable, but it's around +~68%.

68% on 17.32ms is still around ~11ms which while not 0 is ... likely not noticeable.

Some potential workarounds that can be used:
If browser team is okay with `iterStart` not being used but instead starting/connect to browser on first calls. On this first calls the VU can subscribe to `iterEnd` to stop the browser. This way only in those case we will have the needed calls.

If we have `scenarioStart`/`scenarioEnd` - again only on those events different VUs can subscribe to `iterStart`/`iterEnd` for the duration of the scenario.

Both of the above can done after an initial implementation as part of an optimization round. 

cc @imiric @ka3de IMO it is better to continue as is with https://github.com/grafana/xk6-browser/pull/944 and to see what problems will arise from what could be a considered a PoC (but full) solution given the current plan. 

This way we have a point to fallback and something that works.  

After that we can iterate on it if needed/ we have time. 

I will write a different comment on some ... points that I feel haven't been touched on.

mstoykov:
This comment about a fairly removed/high-level thing that isn't directly connected to the specific implementation. And might not be relevant *at this point*. 

If we ever intend on there being a way for a user to say that they want to run js code before/after an iteration. That in practice necessitates that somehow that gets access to the js Runtime. Also for ... consistency it likely should just run on the event loop.

In a related thread of thought - if a module/extension wants to access the js.Runtime as part of it's response to `iterStart`/`iterEnd` it likely should also should get exclusive access to it. The exclusive access to the js.Runtime is what the event loop is all about. 

This likely means that `iterStart`/`iterEnd` events should likely just run on the eventloop, which will also make them sequential. If an implementation wants to free the event loop (so another `iterStart` handler can run for example) it can use the already provided mechanism to get a callback, run something off the event loop and then call the callback after that.

This will be a not small change from the current code and execution where currently `iterStart` and co handlers will run concurrently effectively making it impossible to safely use the js runtime in them. 

So if @ka3de needs to access the runtime in them - we will probably need to fix that. 

This as well is
1. not needed for a PoC - nobody else will use this API for a while
2. not needed at all at this point if browser team will not be accessing the js runtime.

imiric:
@mstoykov:

> If we ever intend on there being a way for a user to say that they want to run js code before/after an iteration. That in practice necessitates that somehow that gets access to the js Runtime.

I'm not understanding what's stopping them from getting it from `VU.Runtime()` once we expose this functionality. The Runtime is available when the `IterStart`/`IterEnd` events are emitted, and since we now wait for them to be processed, whatever runs there happens by definition before/after an iteration. So per-iteration setup/teardown functionality could be built around what's currently here. We just need to expose it via some JS API.

> Also for ... consistency it likely should just run on the event loop.

I deliberately decided to keep the event system decoupled from JS internals. This is a global event system that could be useful for any type of Go module, not just JS ones. The only reason it's currently focused on JS modules is because k6-browser is the first use case that requires it. So I don't think it's a good idea to change this and how events are emitted, just to have deeper integration with the JS packages. I don't see what we would gain from it, either.


> This will be a not small change from the current code and execution where currently `iterStart` and co handlers will run concurrently effectively making it impossible to safely use the js runtime in them.

There should be only one handler for `Iter*` events per VU. This ensures that not only events will be sequential (see the tests), but that there would be no concurrent access to the `goja.Runtime` of that VU. So I don't see why it would be an issue for k6-browser code to get access to the `VU.Runtime()`, in the same way that they will access `VU.State().Options`.

cmd/tests/cmd_run_test.go
imiric: @mstoykov:

> The PR while based on the idea is not really expected to be usable for other extensions or internal features apart from browser. Having this expectations raises (IMO) the complexity quite a lot.

There are no such expectations and k6-browser is our main focus, but at the same time we should a) ensure that whatever is done here doesn't negatively impact current k6 functionality, and b) leave an open-ended design that could be extended if needed, and doesn't couple us to specific k6 internals. This last part is why I don't want to rely on the JS event loop.

> That is to say that we don't need it to be perfect and we will likely change it.

Agreed. What we need to decide here is if the performance impact is a concern or not, and if k6-browser should change the way they manage browser processes in the short-term, i.e. before this PR is merged.

I agree with you and @ka3de that these benchmark results are not concerning in the grand scheme of things, so let's proceed with the current approach, and we can optimize it later.

mstoykov:
> I'm not understanding what's stopping them from getting it from VU.Runtime() once we expose this functionality. The Runtime is available when the IterStart/IterEnd events are emitted, and since we now wait for them to be processed, whatever runs there happens by definition before/after an iteration. So per-iteration setup/teardown functionality could be built around what's currently here. We just need to expose it via some JS API.

Are the `iterStart` handlers executed concurrently? If so them accessing vu.Runtime concurrently is a race condition. So you can get it, but not use it safely. My use of `access` was not a good choice for a word. Likely `use` will be the more correct one.

For the particular case of per iteration setup/teardown *in JS code* - if they are not running on the event loop they will not be able to use promise based APIs for example. 

>  I deliberately decided to keep the event system decoupled from JS internals 

That is good. I am not saying it should be coupled. I am just pointing out that with this decoupling and no additional changes - it can't access the JS runtime in a safe way. For the reason outlined.

>  There should be only one handler for Iter* events per VU.

I specifically prefaced my previous comment to say that this only matters in the more general case where we could have multiple handlers (including JS code ones) that will fire from the same `Iter` event. So in that case it is possible 2 handles try to use the Runtime at the same time.

But even with only with the current changes -  this API now allows [xk6-sql](https://github.com/grafana/xk6-sql) (the first result in my search bar for `xk6`)  to also register per `iter*` handler.  Then both it and browser to use the runtime to run some JS code for some reason or just to create an object or w/e. Both of which are considered use of the JS Runtime and are [not safe](https://github.com/dop251/goja#is-it-goroutine-safe).

And at least from comments and from reading the code - the current emit makes it, so handlers run concurrently and are potentially waited for.

Ultimately JS events are *not* concurrently handled. They are sequentially handled on the event loop. The current handlers for `iterStart` and `iterEnd` are emitted to a thing that is a JS module written in go (the browser module). With the idea of at some point being usable through pure JS code.

While we can have a global event system that can be used for other stuff - topic on which I will not start commenting as I don't want to derail the discussion further. But while we can have such a global event system, at least parts of the events are in practice much closer to what a JS event should be and part of the original issue is actually for them to be JS events.


Maybe the problem is that we use (the same) "event system" for both `init` and `exit`  and `iterStart`/`iterEnd` events. And separating them will ultimately be more ergonomic for use case. 

But I am *not* saying we should do that.

My comment was mostly to point out that the current code does not make using the Runtime in `iterStart`/`iterEnd` safe *or* the same as in any other code. So the question is mostly towards @ka3de if browser will need access to the Runtime or any goja object/struct during the handling of those events. 

For the record using the Runtime in `exit` is ... having even bigger problems. Same for `init` but likely at least more easily fixable in that case.

ka3de (dismissed):
To be clear, from the k6 browser side this implementation fulfills our current requirements, so independently of possible future modifications, this LGTM as it is.

olegbespalov (dismissed):

js/runner.go
```diff
@@ -765,6 +766,26 @@ func (u *ActiveVU) RunOnce() error {
...
+		waitDone := u.moduleVUImpl.events.local.Emit(evt)
+		return func() {
+			waitCtx, waitCancel := context.WithTimeout(u.RunContext, 30*time.Minute)
+			defer waitCancel()
+			if werr := waitDone(waitCtx); werr != nil {
+				u.state.Logger.WithError(werr).Warn()
+			}
+		}
+	}
+
```
mstoykov, line 786: nit: In this case this helper lambda can be `emitAndWaitEvent` which will remove the need to call the wait function.

js/modules/modules.go
mstoykov: Can we put a Note that this is experimental API.

Also we probably should drop it from `RegisterCallback` :thinking:

mstoykov (approved):
LGTM!

I mostly will want to add a note that the Events API is experimental. 

My other comments, might not be as valuable, depending on when and how we continue with this event system.

I feel like that for that we probably should discuss what we want from it and how we expect it will work. I personally would've done a lot of stuff in what would be considered "totally the opposite way". But I don't see us discussing this productively in comments in this PR. 

And I definitely don't expect that even if we agree on changing on how it works we will do it in this PR.

cmd/run.go
```diff
@@ -66,6 +73,25 @@ func (c *cmdRun) run(cmd *cobra.Command, args []string) (err error) {
...
+	emitEvent := func(evt *event.Event) func() {
+		waitDone := c.gs.Events.Emit(evt)
+		return func() {
+			waitCtx, waitCancel := context.WithTimeout(globalCtx, waitEventDoneTimeout)
+			defer waitCancel()
+			if werr := waitDone(waitCtx); werr != nil {
+				logger.WithError(werr).Warn()
+			}
+		}
+	}
```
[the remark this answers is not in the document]
imiric: It did previosly receive a `time.Duration`, but [we decided to replace it with a context](https://github.com/grafana/k6/pull/3112#discussion_r1218019725) because it avoided embedding one in the `System` struct, and passing one avoids using a separate timer (`time.After()`). A context is also safer since it can be based on a parent context, which ensures it can exit early. And passing both a context and a duration seems redundant.

So I see this as a helper function specific to each caller. It doesn't make sense to have it part of the `event` package, as the logic here could be different. It's only similar in both `cmd` and `js` because we essentially want to treat it in the same way. But what if the caller wants to log an `Error` instead, or fail a test, or whatever?

cmd/run.go
```diff
@@ -66,6 +73,25 @@ func (c *cmdRun) run(cmd *cobra.Command, args []string) (err error) {
...
+
+	defer func() {
+		emitEvent(&event.Event{
+			Type: event.Exit,
+			Data: &event.ExitData{Error: err},
+		})()
```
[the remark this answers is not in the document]
imiric: Sure, I can see that.

Would it help if we named the returned function instead of calling it inline?

cmd/run.go
```diff
@@ -153,6 +179,12 @@ func (c *cmdRun) run(cmd *cobra.Command, args []string) (err error) {
 		}()
 	}
 
+	// TODO: Subscribe all initialization processes (outputs, VUs and executors)
+	// to the Init event. This would allow running them concurrently, and they
+	// could be synchronized by waiting for the event processing to complete.
+	// This could later be expanded to also initialize browser processes.
```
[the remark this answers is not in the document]
imiric: > Mostly because I would not find it more readable, and I don't see a case where that will be beneficial.

The benefit is explained in the TODO: it would allow running all independent initialization processes concurrently, and waiting for them to be done before starting the test. Currently the outputs are initialized first, then the VUs and executors. If we plan to eventually also do browser initialization at this point, which we mostly agree on, then running these sequentially doesn't make sense.

I agree that events make the code more difficult to follow, but the implementation here is not any different from using a WaitGroup or a channel to synchronize concurrent tasks. It just formalizes it with event names.

But since it's too early to make this decision, and it would need more discussion, I'll remove the TODO.

imiric:
@mstoykov @olegbespalov I addressed most of your comments in 91fc97ee4, and rebased this on `master`.

It would be difficult to clean up the history at this point, so I'll just squash+merge this instead.

mstoykov (approved):

olegbespalov (approved):

ka3de (approved):

PR #3116, opened 2023-06-06
output/cloud: Hide the versioning flag

codebien:
The new cloud output's version isn't yet ready to be shipped, so we are removing the possibility of enabling so we are not forced to maintain an unstable version in the future.


cloudapi/config.go
```diff
@@ -28,7 +28,9 @@ type Config struct {
 	TestRunDetails null.String `json:"testRunDetails" envconfig:"K6_CLOUD_TEST_RUN_DETAILS"`
 	NoCompress     null.Bool   `json:"noCompress" envconfig:"K6_CLOUD_NO_COMPRESS"`
 	StopOnError    null.Bool   `json:"stopOnError" envconfig:"K6_CLOUD_STOP_ON_ERROR"`
-	APIVersion     null.Int    `json:"apiVersion" envconfig:"K6_CLOUD_API_VERSION"`
+
+	// TODO: Renable the cloud output versioning
+	APIVersion null.Int `json:"-"`
```
mstoykov, line 33: nit : I would keep the old one as a comment for easier revert later.

output/cloud/output_test.go
```diff
@@ -135,26 +135,6 @@ func TestOutputCreateTestWithConfigOverwrite(t *testing.T) {
 	require.NoError(t, out.StopWithTestError(nil))
 }
 
-func TestOutputStartVersionError(t *testing.T) {
-	t.Parallel()
```
mstoykov, line 139: nit : I will also just put `t.Skip` with a message here

mstoykov (dismissed):
LGTM

mstoykov (approved):

imiric (approved):

PR #3120, opened 2023-06-14
output/cloudv2: Unlimited size for body payload

codebien:
We have decided to drop the check for the body's payload. The remote service will take care of splitting the chunk more if it turns out they are too big to be handled.

### DIFF
--- a/output/cloud/expv2/metrics_client.go
+++ b/output/cloud/expv2/metrics_client.go
@@ -76,16 +76,10 @@ func (mc *metricsClient) push(referenceID string, samples *pbcloud.MetricSet) er
 	return nil
 }
 
-const payloadSizeLimit = 100 * 1024 // 100 KiB
-
 func newRequestBody(data *pbcloud.MetricSet) ([]byte, error) {
 	b, err := proto.Marshal(data)
 	if err != nil {
 		return nil, fmt.Errorf("encoding metrics as Protobuf write request failed: %w", err)
-	}
-	if len(b) > payloadSizeLimit {
-		return nil, fmt.Errorf("the Protobuf message is too large to be handled from the Cloud processor; "+
-			"size: %d, limit: 100 KB", len(b))
 	}
 	// TODO: use the framing format
 	// https://github.com/google/snappy/blob/main/framing_format.txt


output/cloud/expv2/metrics_client.go
```diff
@@ -89,17 +89,13 @@ func (mc *metricsClient) push(referenceID string, samples *pbcloud.MetricSet) er
 	return nil
 }
 
-const payloadSizeLimit = 100 * 1024 // 100 KiB
+const payloadSizeLimit = 500 * 1024 // 100 KiB
```
olegbespalov, line 92:
```suggestion
const payloadSizeLimit = 500 * 1024
```

output/cloud/expv2/metrics_client.go
```diff
@@ -108,5 +104,10 @@ func newRequestBody(data *pbcloud.MetricSet) ([]byte, error) {
 		return nil, fmt.Errorf("the Protobuf message is too large to be handled by Snappy encoder; "+
 			"size: %d, limit: %d", len(b), 0xffffffff)
 	}
-	return snappy.Encode(nil, b), nil
+	encoded := snappy.Encode(nil, b)
+	if len(b) > payloadSizeLimit {
```
olegbespalov, line 108: Just double-checking if we should check the original payload size, not the compressed (encoded) one right?

output/cloud/expv2/metrics_client.go
```diff
@@ -18,47 +18,39 @@ import (
...
-type httpDoer interface {
-	Do(*http.Request) (*http.Response, error)
-}
-
 // metricsClient is a Protobuf over HTTP client for sending
 // the collected metrics from the Cloud output
 // to the remote service.
 type metricsClient struct {
-	httpClient httpDoer
+	httpClient *cloudapi.Client
```
imiric, line 25: These are changes from #3104, so can you rebase on `cloud-v2-retries` so that we can see a clear diff here?

codebien:
We discussed it internally and decided to drop the check on the client side.

output/cloud/expv2/metrics_client.go
codebien: FYI, nop based on the fact that the compression making a lot of reduction it was better to do it after otherwise, we could have it this quite often. Anyway, this is now not relevant anymore.

olegbespalov (approved):

codebien:
>Couldn't this be exceeded in some cases?

It could but it would be something deterministic that we could fix. For example, if we have a lot of aggregates per time series and we set on the backend a value that needs to be higher to support it. But it means we detect it, adjust the config, the limit and the chunk limit then it should not happen again because we are aggregating using a finite variable (the time - 3s for the aggregation period and 8s for the waiting period). 

>I suppose the backend would return HTTP 413, and the request would fail anyway

Yeah, it is the main reason for dropping it, we are delegating the control over the backend so we can tweak the limit without a direct dependency here that would increase the complexity.

I expect we and the backend will monitor our systems and we adjust things in case the metrics don't sound correct.

olegbespalov:
Hey @codebien. Are there any blockers to merging this?

codebien:
Hey @olegbespalov, the release was the blocker, it's now ready to be merged.

PR #3121, opened 2023-06-14
Deprecate loadimpact docker image

andrewslotin:
With this PR we begin the deprecation process for `loadimpact/k6` Docker image. `k6` is a part of Grafana now, and we aim to continue releasing future versions under [Grafana organization](https://hub.docker.com/r/grafana/) on DockerHub.

Once merged, all future versions of `loadimpact/k6` Docker image will be displaying a warning banner that suggests migrating to `grafana/k6` before running the tool.

This PR splits the release workflow step that builds and publishes the image to Docker Hub in two stages. The first one builds `grafana/k6` image and publishes it, the second stage creates a new image that displays deprecation warning before running `k6`. This image is than published as `loadimpact/k6`.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -205,7 +205,7 @@ jobs:
           fi
       - name: Build loadimpact/k6
         run: docker build -t $LI_DOCKER_IMAGE_ID .
-      - name: Publish loadimpact/k6 image to Docker Hub
+      - name: Publish loadimpact/k6:master image to Docker Hub
         if: ${{ github.ref == 'refs/heads/master' }}
         run: |
           echo "Publish to Docker Hub as $LI_DOCKER_IMAGE_ID:master"
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -140,7 +140,7 @@ jobs:
         with:
           fetch-depth: 0
       - name: Build
-        run: docker build -t $DOCKER_IMAGE_ID .
+        run: docker build --target release -t $DOCKER_IMAGE_ID .
       - name: Check
         run: |
             docker run $DOCKER_IMAGE_ID version
@@ -187,27 +187,41 @@ jobs:
       - name: Publish k6:master image to Docker Hub
         if: ${{ github.ref == 'refs/heads/master' }}
         run: |
-          echo "Publish to Docker Hub as $DOCKER_IMAGE_ID:master and $LI_DOCKER_IMAGE_ID:master"
+          echo "Publish to Docker Hub as $DOCKER_IMAGE_ID:master"
           docker tag "$DOCKER_IMAGE_ID" "$DOCKER_IMAGE_ID:master"
-          docker tag "$DOCKER_IMAGE_ID" "$LI_DOCKER_IMAGE_ID:master"
           docker push "$DOCKER_IMAGE_ID:master"
-          docker push "$LI_DOCKER_IMAGE_ID:master"
       - name: Publish tagged version image to Docker Hub
         if: ${{ startsWith(github.ref, 'refs/tags/v') }}
         run: |
-          # We need to push the same image in both the loadimpact and the grafana docker hub orgs
           VERSION="${VERSION#v}"
-          echo "Publish to Docker Hub as $DOCKER_IMAGE_ID:$VERSION and $LI_DOCKER_IMAGE_ID:$VERSION"
+          echo "Publish to Docker Hub as $DOCKER_IMAGE_ID:$VERSION"
           docker tag "$DOCKER_IMAGE_ID" "$DOCKER_IMAGE_ID:$VERSION"
-          docker tag "$DOCKER_IMAGE_ID" "$LI_DOCKER_IMAGE_ID:$VERSION"
           docker push "$DOCKER_IMAGE_ID:$VERSION"
+          # We also want to tag the latest stable version as latest
+          if [[ ! "$VERSION" =~ (RC|rc) ]]; then
+            echo "Publish to Docker Hub as $DOCKER_IMAGE_ID:latest"
+            docker tag "$DOCKER_IMAGE_ID" "$DOCKER_IMAGE_ID:latest"
+            docker push "$DOCKER_IMAGE_ID:latest"
+          fi
+      - name: Build loadimpact/k6
+        run: docker build -t $LI_DOCKER_IMAGE_ID .
+      - name: Publish loadimpact/k6 image to Docker Hub
+        if: ${{ github.ref == 'refs/heads/master' }}
+        run: |
+          echo "Publish to Docker Hub as $LI_DOCKER_IMAGE_ID:master"
+          docker tag "$LI_DOCKER_IMAGE_ID" "$LI_DOCKER_IMAGE_ID:master"
+          docker push "$LI_DOCKER_IMAGE_ID:master"
+      - name: Publish loadimpact/k6 tagged version image to Docker Hub
+        if: ${{ startsWith(github.ref, 'refs/tags/v') }}
+        run: |
+          VERSION="${VERSION#v}"
+          echo "Publish to Docker Hub as $LI_DOCKER_IMAGE_ID:$VERSION"
+          docker tag "$LI_DOCKER_IMAGE_ID" "$LI_DOCKER_IMAGE_ID:$VERSION"
           docker push "$LI_DOCKER_IMAGE_ID:$VERSION"
           # We also want to tag the latest stable version as latest
           if [[ ! "$VERSION" =~ (RC|rc) ]]; then
-            echo "Publish to Docker Hub as $DOCKER_IMAGE_ID:latest and $LI_DOCKER_IMAGE_ID:latest"
-            docker tag "$DOCKER_IMAGE_ID" "$DOCKER_IMAGE_ID:latest"
-            docker tag "$DOCKER_IMAGE_ID" "$LI_DOCKER_IMAGE_ID:latest"
-            docker push "$DOCKER_IMAGE_ID:latest"
+            echo "Publish to Docker Hub as $LI_DOCKER_IMAGE_ID:latest"
+            docker tag "$LI_DOCKER_IMAGE_ID" "$LI_DOCKER_IMAGE_ID:latest"
             docker push "$LI_DOCKER_IMAGE_ID:latest"
           fi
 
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -23,3 +23,14 @@ COPY --from=builder /usr/bin/k6 /usr/bin/k6
...
 WORKDIR /home/k6
 ENTRYPOINT ["k6"]
+
+# Legacy loadimpact/k6 image
+FROM grafana/k6
+
+# Temporarily switch back to root to move files
+USER 0
+
+RUN mv /usr/bin/k6 /usr/bin/k6.original
```
olegbespalov, line 33: That line concerns me since we don't copy just built binary, but moving the one that was in `grafana/k6`  :thinking:

andrewslotin: That's the point, the idea is to replace `k6` binary with an entrypoint script, that displays the warning and then runs the original binary.

andrewslotin: Oh, I think I've missed your point. Right, I'm using `grafana/k6` as a base for this image to avoid re-building the binary for `loadimpact/k6`, but since we're building and tagging a new version `grafana/k6` in previous step, this is the one that will be used as a base.

olegbespalov: Yeah, but I'm not sure if it's actually using the local built image (https://github.com/grafana/k6/actions/runs/5268112485/jobs/9524337784?pr=3121): 

```
#9 [linux/arm64 stage-2 1/3] FROM docker.io/grafana/k6@sha256:acad1a06a5e32b5f04f9dbfad112075c8c71b6be8d664bd9d8f79ce6b9559b3d
#9 sha2[56](https://github.com/grafana/k6/actions/runs/5268112485/jobs/9524337784?pr=3121#step:11:57):166b2807661b8e8712a96c60d16bb08ec088eb6a725c6746252a5017a176329f 12.56MB / 12.56MB 0.2s done
#9 sha256:211f8214916f225d429bb667937f6c1bed8d1693e6c081158347b940e4f0f55c 286.20kB / 286.20kB 0.2s done
#9 extracting sha256:211f8214916f225d429bb667937f6c1bed8d1693e6c0811[58](https://github.com/grafana/k6/actions/runs/5268112485/jobs/9524337784?pr=3121#step:11:59)347b940e4f0f55c 0.1s done
#9 extracting sha256:166b2807661b8e8712a96c60d16bb08ec088eb6a725c6746252a5017a176329f
#9 extracting sha256:166b2807661b8e8712a96c[60](https://github.com/grafana/k6/actions/runs/5268112485/jobs/9524337784?pr=3121#step:11:61)d16bb08ec088eb6a725c6746252a5017a176329f 0.3s done
#9 DONE 0.5s
```

entrypoint-legacy.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+
+cat >&2 <<-EOF
++------------------------------------------------------------------------------+
+| WARNING: loadimpact/k6 Docker image has been replaced by grafana/k6.         |
```
imiric, line 5:
```suggestion
| WARNING: The loadimpact/k6 Docker image has been replaced by grafana/k6.     |
```

Dockerfile
imiric: We can avoid the user change, and the copying of the binary if we simply override the entrypoint for this stage. I.e.:

```dockerfile
# Legacy loadimpact/k6 image
FROM --platform=$BUILDPLATFORM release

COPY entrypoint-legacy.sh /usr/bin/
ENTRYPOINT ["/usr/bin/entrypoint-legacy.sh"]
```

Then the entrypoint script can just call the `k6` binary directly.

entrypoint-legacy.sh
```diff
@@ -0,0 +1,13 @@
...
+|          THIS IMAGE IS DEPRECATED and its support will be discontinued after |
+|          Dec 31, 2023. Please update your scripts to use grafana/k6 to       |
+|          continue using the latest version of k6.                            |
++------------------------------------------------------------------------------+
+
+EOF
+
+k6.original $@
```
imiric, line 13: This can be the same binary used in the `grafana/k6` image if we change the entrypoint, as I mentioned above.

But two more things here:

- Using the absolute path is safer, as otherwise this would run any `k6` binary on the `PATH`. Come to think of it, we should also change the entrypoint of the `grafana/k6` image to be `/usr/bin/k6`. Can you do that here or in another PR? :pray:
- Quoting `$@` is safer, as otherwise arguments with spaces would be interpreted as arguments to k6 itself.

  Consider `docker run --rm -i loadimpact/k6 run --tag 'some tag=some value' - <examples/thresholds.js`, which would fail with a cryptic `accepts 1 arg(s), received 3`.

So this should be:

```suggestion
/usr/bin/k6 "$@"
```

Dockerfile
```diff
@@ -23,3 +23,14 @@ COPY --from=builder /usr/bin/k6 /usr/bin/k6
 USER 12345
```
imiric, line 23: Not related to this PR, but seeing you've reproduced it below (which we can remove if we change the entrypoint), this would be more readable and maintainable as `USER k6`.

So if you're doing the entrypoint change to the `grafana/k6` image as I suggested above, then let's make this change as well.

In reality, we should update the `builder` stage to not rely on the `GOPATH` workdir, since that's a legacy requirement before Go modules were introduced, but that would be a larger change we don't need to do now.

.github/workflows/build.yml
```diff
@@ -194,27 +195,44 @@ jobs:
...
+        if: ${{ github.ref == 'refs/heads/master' }}
+        run: |
+          echo "Publish to Docker Hub as $LI_DOCKER_IMAGE_ID:master"
+          docker tag "$LI_DOCKER_IMAGE_ID" "$LI_DOCKER_IMAGE_ID:master"
+          docker push "$LI_DOCKER_IMAGE_ID:master"
+      - name: Publish loadimpact/k6 tagged version image to Docker Hub
+        if: ${{ startsWith(github.ref, 'refs/tags/v') }}
+        run: |
+          VERSION="${VERSION#v}"
+          echo "Publish to Docker Hub as $LI_DOCKER_IMAGE_ID:$VERSION"
```
imiric, line 229: I like the splitting of this into a separate step, but I think we can abstract this logic (what to do in case it's a `master` or tag push, etc.) in a GH action. This way we don't have to repeat the same logic 3 times.

But we can do that in a subsequent PR. I can create an issue if you agree.

imiric (changes requested):
Nicely done @andrewslotin! :+1:

Just pointed out a few improvements we could make, but overall it LGTM. I tested the image locally and it worked as expected.

entrypoint-legacy.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+
+cat >&2 <<-EOF
++------------------------------------------------------------------------------+
+| WARNING: loadimpact/k6 Docker image has been replaced by grafana/k6.         |
+|          THIS IMAGE IS DEPRECATED and its support will be discontinued after |
+|          Dec 31, 2023. Please update your scripts to use grafana/k6 to       |
```
imiric, line 7: That's a good timeframe :+1:

The idea is that by "discontinued" we mean that we will stop pushing new k6 releases to it, but it will still work with this warning. And some time after that (6 months?) we would push the `exit 1` change, and a bit more (6 more months?) we would delete the `loadimpact` namespace altogether.

Is this mostly the plan, pending timeframe agreements?

.github/workflows/build.yml
andrewslotin: Let's do this separately 👍

entrypoint-legacy.sh
andrewslotin: I believe we can decide on how much do we want to push users to upgrade later. Personally I become more and more reluctant to adding `exit 1`. Not publishing new releases to the `loadimpact` org might be good enough.

entrypoint-legacy.sh
andrewslotin: I'd rather prefer to replace the `k6` binary to keep the entrypoint from the `release` step. Well spotted on quoting the args list 👍

andrewslotin: Actually this would affect the `k6 help` output:
```
Usage:
  k6.original [command]
     ~~~~~~~~
```
Let's go with your suggestions then 👍

andrewslotin:
@imiric, @olegbespalov, I've rebased this branch, updated my changes to use `docker build` instead of `buildx` and also incorporated your suggestions.

olegbespalov (approved):
Looks good to me :muscle: 

However, I'd wait with merging this till the `v0.46.0` cycle.

olegbespalov:
Sure. My main concern is that we have a release on Monday, and keeping in mind that today is Friday I feel like we have no time for proper testing of this, even if the change looks simple at first look. 

Regarding: 

>Merging it now would give users more time to migrate to the new image.

As far as I know, we don't have any strict deadlines around the date. It's just something we set for ourselves. So if you feel that our customers need more time, let's adjust the date to `Marth 1, 2024` and our users still will have the same amount of time (or even more since we merge this at the beginning of the next cycle).

imiric (approved):
LGTM! :clap:

I don't see much risk in merging this before the v0.45.0 release, as long as Andrey has tested the workflow in his fork. There wouldn't be any impact to the `grafana/k6` image, so as long as the workflow passes and the `loadimpact/k6` image does what we need, it should be fine.

But I like to err on the side of caution, so if Oleg is concerned about it, we don't have to rush this, and it can wait for v0.46.0. We'll actually test it even before, since the `loadimpact/k6:master` image would be pushed as soon as this is merged.

entrypoint-legacy.sh
imiric: > Personally I become more and more reluctant to adding `exit 1`. Not publishing new releases to the loadimpact org might be good enough.

`exit 1` is part of making a gradual change. The first step is this deprecation warning; the next step is k6 tests failing with `exit 1` so that users will actually notice the warning; and the final step is removal of the `loadimpact` org, where CI builds would fail much earlier.

I suppose there's no harm in letting old images of k6 there forever, but we do want users to upgrade to new versions, and since Load Impact is not an entity anymore, there's no sense in keeping the org around. This way we avoid some Docker Inc. manager deciding again that maybe OSS projects should start paying for the data and traffic :sweat_smile:

andrewslotin:
Let’s merge and test it after the release. After all, @olegbespalov is right and we can push the EOL date if needed.

entrypoint-legacy.sh
andrewslotin: I believe once the EOL date has passed, we can retire `loadimpact` anytime should such decision be made.

PR #3125, opened 2023-06-15
output/cloudv2: Use a static remote service url

codebien:
It moves the sets of the test run id and the relative URL generation from per request to once per client.

### DIFF
--- a/output/cloud/expv2/metrics_client.go
+++ b/output/cloud/expv2/metrics_client.go
@@ -21,41 +21,40 @@ import (
 // to the remote service.
 type metricsClient struct {
 	httpClient *cloudapi.Client
-	baseURL    string
+	url        string
 }
 
 // newMetricsClient creates and initializes a new MetricsClient.
-func newMetricsClient(c *cloudapi.Client) (*metricsClient, error) {
-	// Unfortunately, the cloudapi.Client works across different versions
-	// of the API, but it has the v1 harcoded so we need to trim the wrong path
+func newMetricsClient(c *cloudapi.Client, testRunID string) (*metricsClient, error) {
+	// The cloudapi.Client works across different versions of the API, the test
+	// lifecycle management is under /v1 instead the metrics ingestion is /v2.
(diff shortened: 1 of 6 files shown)


imiric:
@codebien I'm waiting for this PR to also be rebased and the conflicts resolved before reviewing it.

output/cloud/expv2/metrics_client.go
```diff
@@ -21,41 +21,36 @@ import (
...
 // newMetricsClient creates and initializes a new MetricsClient.
-func newMetricsClient(c *cloudapi.Client) (*metricsClient, error) {
+func newMetricsClient(c *cloudapi.Client, testRunID string) (*metricsClient, error) {
 	// Unfortunately, the cloudapi.Client works across different versions
 	// of the API, but it has the v1 harcoded so we need to trim the wrong path
 	// to be able to replace it with the correct one.
 	u := c.BaseURL()
 	if !strings.HasSuffix(u, "/v1") {
 		return nil, errors.New("a /v1 suffix is expected in the Cloud service's BaseURL path")
 	}
```
imiric, line 35: Why not remove the hardcoded `v1`, and allow specifying the API version in `cloudapi.NewClient()`?

imiric (approved):

output/cloud/expv2/metrics_client.go
codebien: >works across different versions

Mainly for this reason. Because the logic for the test lifecycle is still over the /v1 path.

Adding this logic to the client will also break the current API that is used by other services like the `k6-operator`. If want it I think we should do it when we are sure that the API is stable.

If you have a suggestions for doing this transition smoother I would like to implement it.

In any case, the current docs seems to not be clear enough, going to amend it.

codebien: Improved the documentation

imiric (approved):

olegbespalov (approved):

PR #3127, opened 2023-06-16
Use docker buildx to publish images

andrewslotin:
This PR addresses an issue with the current build process not pushing the latest built image caused by https://github.com/grafana/k6/pull/3015. For example `grafana/k6:master` image built a few days ago is the same as `grafana/k6:0.44.1` that has been pushed a month ago.

Screenshot 2023-06-16 at 11 19 39

The reason for this is that `docker buildx` uses a different context to build images, which means that all subsequent docker run and docker push commands work with an image that is different from what has been built at the “Build” step. 

The reason why this went unnoticed is the “Check” step, that “fixes” default context state by pulling the latest grafana/k6 image from DockerHub and this is the image used by all subsequent stages.

Once this applied, `build-docker` job will start using `docker buildx --push` to publish the image from `multibuilder` context instead of `default`. Since we don't change any files in-between, this command always uses the cached image created during the "Build" step. To run smoke test at the "Check" stage, we're going to load it into the `default` context by executing `docker buildx --load` before running the image.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -150,6 +150,7 @@ jobs:
             -t $DOCKER_IMAGE_ID .
       - name: Check
         run: |
+            docker buildx build --load -t $DOCKER_IMAGE_ID .
             docker run $DOCKER_IMAGE_ID version
             docker run $DOCKER_IMAGE_ID --help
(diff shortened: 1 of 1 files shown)


andrewslotin:
Closing as not necessary, since #3015 has been reverted.

PR #3150, opened 2023-06-27
output/cloudv2: Add retry support to insights client

vortegatorres:
This PR adds retries for sending request metadata to the k6 Insights cloud backend.


cloudapi/insights/client.go
```diff
@@ -58,6 +63,21 @@ type ClientTLSConfig struct {
 	CertFile string
 }
 
+// RetryConfig is the configuration for the client retries.
+type RetryConfig struct {
```
Blinkuu, line 67: nit: Prefix with `Client` to match other struct names.

cloudapi/insights/client.go
```diff
@@ -58,6 +63,21 @@ type ClientTLSConfig struct {
...
+	RetryableStatusCodes string
+	MaxAttempts          uint
+	PerRetryTimeout      time.Duration
+	BackoffConfig        BackoffConfig
+}
+
+// BackoffConfig is the configuration for the client retries using the backoff exponential with jitter algorithm.
+type BackoffConfig struct {
```
Blinkuu, line 75: nit: Prefix with `Client` to match other struct names.

cloudapi/insights/client_test.go
```diff
@@ -297,6 +355,96 @@ func TestClient_IngestRequestMetadatasBatch_ReturnsErrorWithFailingServerAndNonC
 	require.ErrorContains(t, err, testErr.Error())
 }
 
+func TestClient_IngestRequestMetadatasBatch_ReturnsErrorWithNoErrorAfterRetrySeveralTimes(t *testing.T) {
```
Blinkuu, line 358: I think the test name here is a bit misleading.

cloudapi/insights/client_test.go
```diff
@@ -297,6 +355,96 @@ func TestClient_IngestRequestMetadatasBatch_ReturnsErrorWithFailingServerAndNonC
...
+	err := cli.IngestRequestMetadatasBatch(context.Background(), data)
+
+	// Then
+	require.NoError(t, err)
+	require.True(t, ser.batchCreateRequestMetadatasInvoked)
+	require.True(t, ser.dataUploaded)
+	require.Equal(t, uint(10), ser.attempts)
+}
+
+func TestClient_IngestRequestMetadatasBatch_ReturnsErrorWithErrorAfterRetrySeveralTimes(t *testing.T) {
```
Blinkuu, line 402: Same here. Not really sure what `ReturnsErrorWithErrorAfterRetrySeveralTimes` means. Shouldn't it be something more like `ReturnsErrorAfterExhaustingMaxRetryAttempts`, or something along these lines.

output/cloud/expv2/output.go
```diff
@@ -133,6 +133,16 @@ func (o *Output) Start() error {
...
+			RetryConfig: insights.RetryConfig{
+				RetryableStatusCodes: "UNAVAILABLE,INTERNAL",
+				MaxAttempts:          5,
+				PerRetryTimeout:      1 * time.Second,
+				BackoffConfig: insights.BackoffConfig{
+					Enabled:        true,
+					JitterFraction: 0.1,
+					WaitBetween:    1 * time.Second,
+				},
+			},
```
Blinkuu, line 145: We need to adjust this config. Currently, we send data using the insights client on its own goroutine, but the calls themselves are *blocking*. The [flushRequestMetadatas()](https://github.com/grafana/k6/blob/1df72a0a1b23d7b0827813f6b709c7ad7bb4193b/output/cloud/expv2/output.go#L279-L291) method is invoked periodically every [TracesPushInterval](https://github.com/grafana/k6/blob/1df72a0a1b23d7b0827813f6b709c7ad7bb4193b/output/cloud/expv2/output.go#L147). What's more, this interval is also used as the context deadline [here](https://github.com/grafana/k6/blob/1df72a0a1b23d7b0827813f6b709c7ad7bb4193b/output/cloud/expv2/output.go#L282). This means that we currently get just 1 second *total* to send a request using this client. Therefore, configuring `MaxAttemps` to `5` with `PerRetryTimeout` equal to `1s` makes little sense.

I see a few paths forward:
1. We make the [requestMetadataFlusher](https://github.com/grafana/k6/blob/1df72a0a1b23d7b0827813f6b709c7ad7bb4193b/output/cloud/expv2/flush.go#L154) non-blocking and increase the context deadline required to flush traces so that it encompasses what we defined in the retry config.
2. We keep the requestMetadataFlusher blocking, increasing the context deadline overall. This may be dangerous as we could theoretically block for a long time while, at the same time, the traces accumulate in memory. This means that after some unavailability, the next request might be huge (in terms of the number of metadatas to send) compared to the mean.

I'd like to hear @codebien opinion on this as well.

Blinkuu (changes requested):

cloudapi/insights/client.go
vortegatorres: Done, thanks!

cloudapi/insights/client.go
vortegatorres: Done, thanks!

cloudapi/insights/client_test.go
vortegatorres: I updated to `TestClient_IngestRequestMetadatasBatch_ReturnsNoErrorAfterRetrySeveralTimes`. Thanks!

cloudapi/insights/client_test.go
vortegatorres: Changed to `ReturnsErrorAfterExhaustingMaxRetryAttempts`; it is much better, thanks!

output/cloud/expv2/output.go
vortegatorres: Thanks for the feedback you are totally right. If we decide to implement 1, I understand that we should create a channel to receive if the execution was successful to handle the error properly [here](https://github.com/grafana/k6/blob/1df72a0a1b23d7b0827813f6b709c7ad7bb4193b/output/cloud/expv2/output.go#L271), right?

There would be another alternative that I don't know makes sense that would be to keep the retries within the [TracesPushInterval](https://github.com/grafana/k6/blob/1df72a0a1b23d7b0827813f6b709c7ad7bb4193b/output/cloud/expv2/output.go#L147) of 1 second. I would like to hear the thoughts of @codebien as well.

Blinkuu: > If we decide to implement 1, I understand that we should create a channel to receive if the execution was successful to handle the error properly [here](https://github.com/grafana/k6/blob/1df72a0a1b23d7b0827813f6b709c7ad7bb4193b/output/cloud/expv2/output.go#L271), right?

We don't need more sophisticated error handling at this point. We don't return any sophisticated errors from the backend at this point. In the case of (1), I thought more about "fire and forget". Maybe in the future we add more logic to it.

Blinkuu: Let's follow the pattern you already established. It is easy to implement.

So, we would have to:
1. Add default timeout to `insights.Client`. It can be 1 minute as well. It will make retries very robust.
2. Add `TracesPushConcurrency` option to `cloudapi.Config`
3. Call `requestMetadatasFlusher.flush(ctx)` with `context.Background()`.
4. Introduce similar logic to `runFlushWorkers()`, but for our `flushRequestMetadatas()` method.

@vortegatorres WDYT?

vortegatorres: I agree to follow the same strategy.

cloudapi/config.go
```diff
@@ -161,9 +164,10 @@ func NewConfig() Config {
 		MetricPushInterval:    types.NewNullDuration(1*time.Second, false),
 		MetricPushConcurrency: null.NewInt(1, false),
 
-		TracesEnabled:      null.NewBool(false, false),
-		TracesHost:         null.NewString("insights.k6.io:4443", false),
-		TracesPushInterval: types.NewNullDuration(1*time.Second, false),
+		TracesEnabled:         null.NewBool(false, false),
+		TracesHost:            null.NewString("insights.k6.io:4443", false),
+		TracesPushInterval:    types.NewNullDuration(10*time.Second, false),
+		TracesPushConcurrency: null.NewInt(6, false),
```
codebien, line 170: It sounds like a magic number, do we have some math behind it?

cloudapi/config.go
```diff
@@ -161,9 +164,10 @@ func NewConfig() Config {
 		MetricPushInterval:    types.NewNullDuration(1*time.Second, false),
 		MetricPushConcurrency: null.NewInt(1, false),
 
-		TracesEnabled:      null.NewBool(false, false),
-		TracesHost:         null.NewString("insights.k6.io:4443", false),
-		TracesPushInterval: types.NewNullDuration(1*time.Second, false),
+		TracesEnabled:         null.NewBool(false, false),
+		TracesHost:            null.NewString("insights.k6.io:4443", false),
+		TracesPushInterval:    types.NewNullDuration(10*time.Second, false),
```
codebien, line 169: Out of curiosity, is it intentional? What is the rational for expanding it?

Blinkuu: +1, I'd still do it every second

cloudapi/config.go
Blinkuu: yeah, we could still default to `1` and see if we run into any issues

vortegatorres: Ok, I restored 1, thanks

cloudapi/config.go
vortegatorres: As we have a longer timeout, I though about extending the interval. I restored the second; thanks!

cloudapi/config.go
codebien: Does `1` mean no concurrency? Do you expect to tune this value during tests or ideally you would like to run with it as the default?

vortegatorres: yes, we would like to test first this value and then decide if we change ir.

codebien (commented):
LGTM, @mstoykov has it in its reviews queue

cloudapi/insights/client.go
```diff
@@ -180,9 +202,75 @@ func dialOptionsFromClientConfig(cfg ClientConfig) ([]grpc.DialOption, error) {
...
+	}
+
+	if len(retryableStatusCodes) == 0 {
+		return nil, fmt.Errorf("no retryable status codes provided")
+	}
+
+	statusCodes := strings.Split(retryableStatusCodes, ",")
+	errorCodes := make([]codes.Code, 0, len(statusCodes))
+	for _, code := range statusCodes {
+		errorCode, ok := statusCodeMap[code]
```
mstoykov, line 264: I feel like we might be better of using `UnsmarshalJSON` https://github.com/grpc/grpc-go/blob/v1.56.1/codes/codes.go#L219 

This iwll make it so we don't have to figure out if a code was added or not. I for example was certain you have missed "Unauthenticated" ... but it is just not in the same place as it is in the codes' code ( :wink: )

output/cloud/expv2/output.go
```diff
@@ -275,14 +287,36 @@ func (o *Output) flushMetrics() {
...
+
 // flushRequestMetadatas periodically flushes traces collected in RequestMetadatasCollector using flusher.
 func (o *Output) flushRequestMetadatas() {
 	start := time.Now()
 
-	ctx, cancel := context.WithTimeout(context.Background(), o.config.TracesPushInterval.TimeDuration())
-	defer cancel()
-
-	err := o.requestMetadatasFlusher.flush(ctx)
+	err := o.requestMetadatasFlusher.flush(context.Background())
```
mstoykov, line 319: I think we left this context so that this output can use it - if we are not going to use it - can we just drop it?

output/cloud/expv2/output.go
```diff
@@ -133,6 +135,16 @@ func (o *Output) Start() error {
...
+			RetryConfig: insights.ClientRetryConfig{
+				RetryableStatusCodes: "UNKNOWN,INTERNAL,UNAVAILABLE,DEADLINE_EXCEEDED",
+				MaxAttempts:          3,
+				PerRetryTimeout:      30 * time.Second,
+				BackoffConfig: insights.ClientBackoffConfig{
+					Enabled:        true,
+					JitterFraction: 0.1,
+					WaitBetween:    1 * time.Second,
+				},
+			},
```
mstoykov, line 147: Are those supposed to be configurable by the user?

output/cloud/expv2/output.go
```diff
@@ -275,14 +287,36 @@ func (o *Output) flushMetrics() {
...
+				o.wg.Done()
+			}()
+
+			for {
+				select {
+				case <-t.C:
+					o.flushRequestMetadatas()
+				case <-o.stop:
+					return
+				case <-o.abort:
```
mstoykov, line 307:
```suggestion
				case <-o.stop, <-o.abort:
```

output/cloud/expv2/output.go
```diff
@@ -275,14 +287,36 @@ func (o *Output) flushMetrics() {
...
+func (o *Output) runFlushRequestMetadatas() {
+	t := time.NewTicker(o.config.TracesPushInterval.TimeDuration())
+
+	for i := int64(0); i < o.config.TracesPushConcurrency.Int64; i++ {
+		o.wg.Add(1)
+		go func() {
+			defer func() {
+				t.Stop()
+				o.wg.Done()
+			}()
```
mstoykov, line 299: nit: To be honest I would likely just do `defer o.wg.Done()` and co. (in reverse) without actually making another function.

output/cloud/expv2/output.go
codebien: Are you sure this is a valid syntax?

mstoykov: Hmm ... I guess it isn't as it needs to be statements instead of expressions as in the case of `switch` ... I think I have fallen for this more then once at this point 

Sorry :bow:

cloudapi/insights/client.go
```diff
@@ -180,9 +202,75 @@ func dialOptionsFromClientConfig(cfg ClientConfig) ([]grpc.DialOption, error) {
...
+
+	backoffConfig := retryConfig.BackoffConfig
+	if backoffConfig.Enabled {
+		backoff := grpcRetry.WithBackoff(
+			grpcRetry.BackoffExponentialWithJitter(backoffConfig.WaitBetween, backoffConfig.JitterFraction))
+		unaryInterceptor := grpcRetry.UnaryClientInterceptor(withCodes, withMax, withPerRetryTimeout, backoff)
+		return []grpc.UnaryClientInterceptor{unaryInterceptor}, nil
+	}
+
+	unaryInterceptor := grpcRetry.UnaryClientInterceptor(withCodes, withMax, withPerRetryTimeout)
```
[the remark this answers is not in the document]
vortegatorres: Thanks for the feedback. Good idea. I created a `[]CallOption` variable, and `grpc.UnaryClientInterceptor` is initialized/returned once. Also, I changed the function signature to return a single `grpc.UnaryClientInterceptor`.

output/cloud/expv2/output.go
vortegatorres: Do you mean remove it from the `flusher interface` and all its implementations?

output/cloud/expv2/output.go
vortegatorres: No, for them, these values won't make sense.

output/cloud/expv2/output.go
vortegatorres: Done, thanks!

output/cloud/expv2/output.go
mstoykov: I meant the context argument specifically.

But to be honest I am not really a fan of adding interfaces for the sake of adding interfaces or to make stuff .. mockable. Especially in cases like this where I feel the end result is a singel function call.

I do feel though that dropping the whole interface might be more invbolved and not to be part of this PR.

cloudapi/insights/client.go
mstoykov: It also takes the name 

https://github.com/grpc/grpc-go/blob/5b67e5ea449ef0686a0c0b6de48cd4cb63e3db2a/codes/codes.go#L239

output/cloud/expv2/output.go
vortegatorres: After reading this [comment](https://raintank-corp.slack.com/archives/C042WNQFZEZ/p1688460082544839?thread_ts=1687937713.085209&cid=C042WNQFZEZ) in slack I removed the context, please let me know if now looks better to you.

cloudapi/insights/client.go
vortegatorres: Yes, you are totally right; I updated `retryableStatusCodes` to use `UnmarshalJSON`. Thanks!

cloudapi/insights/client.go
```diff
@@ -180,9 +202,56 @@ func dialOptionsFromClientConfig(cfg ClientConfig) ([]grpc.DialOption, error) {
...
+
+	statusCodes := strings.Split(retryableStatusCodes, ",")
+	errorCodes := make([]codes.Code, 0, len(statusCodes))
+	for _, code := range statusCodes {
+		var errorCode codes.Code
+		err := errorCode.UnmarshalJSON([]byte(code))
+		if err != nil {
+			return nil, fmt.Errorf("invalid status code %s provided", code)
+		}
+		errorCodes = append(errorCodes, errorCode)
```
mstoykov, line 249:
```suggestion
	errorCodes := make([]codes.Code, len(statusCodes))
	for i, code := range statusCodes {
		err := errorCodes[i].UnmarshalJSON([]byte(code))
		if err != nil {
			return nil, fmt.Errorf("invalid status code %s provided", code)
		}
```

mstoykov (approved):

mstoykov (approved):

Blinkuu (approved):

codebien (approved):

PR #3157, opened 2023-06-30
Introducing the MaxTimeSeriesInBatch config option

### LINKED ISSUE #3156: Use a dedicated option instead of MaxMetricSamplesPerPackage
### What

As defined by the code's comment below, we should use a better name that matches the purpose for `MaxMetricSamplesPerPackage` in `cloudv2`.

`cloudv2` uses `MaxMetricSamplesPerPackage` for defining the max number of time series in a single flush batch. Instead, the current option defines the number of k6 samples.

https://github.com/grafana/k6/blob/1df72a0a1b23d7b0827813f6b709c7ad7bb4193b/cloudapi/config.go#L34-L36

https://github.com/grafana/k6/blob/1df72a0a1b23d7b0827813f6b709c7ad7bb4193b/output/cloud/expv2/output.go#L110-L112

### Why

`MaxMetricSamplesPerPackage `is inherited from v1 but we should use a cleaner and dedicated field.

### Suggestion

Add a dedicated new field for defining the max number of series in a single flush batch.

olegbespalov:
## What?

Moving cloud output v2 from MaxMetricSamplesPerPackage to MaxTimeSeriesInBatch.

## Why?

The new option reflects the actual purpose and usage of the config option.

## Checklist

## Related PR(s)/Issue(s)

Closes #3156

### DIFF
--- a/cloudapi/config.go
+++ b/cloudapi/config.go
@@ -31,9 +31,8 @@ type Config struct {
 	StopOnError    null.Bool   `json:"stopOnError" envconfig:"K6_CLOUD_STOP_ON_ERROR"`
 	APIVersion     null.Int    `json:"apiVersion" envconfig:"K6_CLOUD_API_VERSION"`
 
-	// TODO: rename the config field to align to the new logic by time series
-	// when the migration from the version 1 is completed.
-	MaxMetricSamplesPerPackage null.Int `json:"maxMetricSamplesPerPackage" envconfig:"K6_CLOUD_MAX_METRIC_SAMPLES_PER_PACKAGE"`
+	// Defines the max allowed number of time series in a single batch.
+	MaxTimeSeriesInBatch null.Int `json:"maxTimeSeriesInBatch" envconfig:"K6_CLOUD_MAX_TIME_SERIES_IN_BATCH"`
 
 	// The time interval between periodic API calls for sending samples to the cloud ingest service.
 	MetricPushInterval types.NullDuration `json:"metricPushInterval" envconfig:"K6_CLOUD_METRIC_PUSH_INTERVAL"`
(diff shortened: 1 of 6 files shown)


cloudapi/config.go
```diff
@@ -166,6 +168,7 @@ func NewConfig() Config {
 		TracesPushInterval: types.NewNullDuration(1*time.Second, false),
 
 		MaxMetricSamplesPerPackage: null.NewInt(100000, false),
+		MaxTimeSeriesInBatch:       null.NewInt(100000, false),
```
codebien, line 171:
```suggestion
		MaxTimeSeriesInBatch:       null.NewInt(10000, false),
```
Starting with one order of magnitude lower should be better.

cloudapi/config.go
```diff
@@ -31,10 +31,12 @@ type Config struct {
 	StopOnError    null.Bool   `json:"stopOnError" envconfig:"K6_CLOUD_STOP_ON_ERROR"`
 	APIVersion     null.Int    `json:"apiVersion" envconfig:"K6_CLOUD_API_VERSION"`
 
-	// TODO: rename the config field to align to the new logic by time series
-	// when the migration from the version 1 is completed.
+	// TODO: remove this when migration from the cloud output v1 will be completed
```
codebien, line 34: At this point, we can move this [at the end of the struct](https://github.com/grafana/k6/blob/ecbe21512901753390e0a12c025f1c9c2cb732f9/cloudapi/config.go#L154) and mark all the _aggregation_ options as deprecated and to be removed after the transition.

codebien (changes requested):
I added a few comments. Thanks for taking care.

Can you please add a validation rule like we have for the old option? https://github.com/grafana/k6/blob/3291b98cad0cea54292025d91889575f116f0fe7/output/cloud/output.go#L116-L119

cloudapi/config.go
olegbespalov: Resolved with the latest force push

cloudapi/config.go
olegbespalov: Resolved with the latest force push

codebien (approved):

mstoykov (approved):

Issue #2311, opened 2021-12-22
Problems and deficiencies around uploading big files with k6


MStoykov:
This will discuss deficiencies around `net/http`, but in reality some of those apply to other APIs as well. 

The idea is mostly to:
1. Discuss the problems with the current API
2. Explore what is actually wrong(or not) with it
3. Figure out which parts are important, and we would like "fixed"
4. And then *possibly* discuss solutions, although unless they are simple I would advise having them in separate issues linking to this one.

But let's first get a fairly simple script
```
import http from 'k6/http';
import { sleep } from 'k6';

const binFile = open('/path/to/file.bin', 'b'); // line A

export default function () {
  const data = {
    field: Math.random(), // dynamic field
    file: http.file(binFile, 'test.bin'), // Line B
  };

  const res = http.post('https://example.com/upload', data); // Line C
  sleep(3);
}
```

Given this very simple example currently(v0.35.0) we have
1. A copy of the `file.bin` created for each VU on line A. This can be addressed with [1931](https://github.com/grafana/k6/issues/1931) and [PR2303](https://github.com/grafana/k6/pull/2303) 
2. on Line B there is currently no copy, but as `binFile` is a mutatable ArrayBuffer it technically can lead to a possibility to edit the underlying data after `data` is created. And if `http.post` was asynchronous(and 3 is fixed) possibly even a data race. Even if it was copying and if there wasn't a dynamic data this could've been done *once* in the init context.
3. On line C `data` gets ["marshaled"](https://github.com/grafana/k6/blob/804203bcf574da57cd05d67b39f03d05c54523ab/js/modules/k6/http/request.go#L233-L236) into [multipart](https://github.com/grafana/k6/blob/804203bcf574da57cd05d67b39f03d05c54523ab/js/modules/k6/http/request.go#L172-L211) body effectively copying the body. 
4. On line C *again* `res` also has a field `request` which has a field `body` which is a [string dump](https://github.com/grafana/k6/blob/804203bcf574da57cd05d67b39f03d05c54523ab/lib/netext/httpext/request.go#L143-L145) of the body that was created above.

So for this particular script there are 1,3 and 4 make copies of potentially *big* amount of bytes.

Let's go with a similar but *slightly different* script
```
import http from 'k6/http';
import { check } from 'k6';
import { FormData } from 'https://jslib.k6.io/formdata/0.0.2/index.js';

const binFile = open('/path/to/file.bin', 'b'); // line A;

export default function () {
[...]

PR #3159, opened 2023-07-02
Support for per-vu TLS configuration (grpc.Client.connect only) and custom (self-signed) RootCAs

chrismoran-mica:
## What? and Why?

You can now use self-signed ca-certificates when authenticating with TLS. Using the `cacerts` property of an options `tlsAuth` object, you can now indicate the CA certificate(s) that k6 will use to verify the server certificates your tests will access. CA certificates must be in `pem` format and multiple CA certificates can be included.

```javascript
export const options = {
    tlsAuth: [
        {
            cacerts: [open("cacerts.pem")],
            cert: open("cert.pem"),
            key: open("cert-key.pem"),
            password: "cert-passphrase",
            domains: ["grpcbin.test.k6.io"], // Deprecated
        },
    ],
};
```

You can also now use per-vu TLS configuration on `grpc.Client` `connect` params. Previously, `grpc.Client` `connect` made use of a shared (immutable) `tls.Config` cloned into every VU. Now, the `connect` `params` argument can contain a `tlsconfig` key that maps to the `TLSAuth` struct. This will create a per-vu TLS configuration for the `grpc.Client` allowing you to connect to multiple mTLS endpoints in a single VU. Note: the global tls.Config will remain unmodified.

```javascript
const params = {
    plaintext: false,
    tlsconfig: {
        cacerts: ["cacerts.pem"],
        cert: "cert.pem",
        key: "key.pem",
        password: "cert-passphrase",
        domains: ["grpcbin.test.notk6.io"], // Deprecated
    }
};
const hostPort = "grpcbin.test.notk6.io:9001";
const client = new grpc.Client();

client.connect(hostPort, params);
```

## Checklist

- [x] ~I have added tests for my changes.~
- [x] I have *updated* tests to consider my changes.

## Related PR(s)/Issue(s)

This PR does not close an issue but it does reference at least one conversation about per-vu mTLS: https://community.k6.io/t/grpc-mtls-per-vu-iteration/1868

### DIFF
--- a/js/modules/k6/grpc/client.go
+++ b/js/modules/k6/grpc/client.go
@@ -2,6 +2,8 @@ package grpc
 
 import (
 	"context"
+	"crypto/tls"
+	"crypto/x509"
 	"errors"
 	"fmt"
 	"io"
@@ -108,6 +110,51 @@ func (c *Client) LoadProtoset(protosetPath string) ([]MethodInfo, error) {
 	return c.convertToMethodInfo(fdset)
 }
 
+func buildTLSConfig(certificate, key []byte, caCertificates [][]byte) (*tls.Config, error) {
+	var cp *x509.CertPool
+	var err error
+	if cp, err = x509.SystemCertPool(); err != nil {
+		return nil, err
+	}
+	if len(caCertificates) > 0 {
+		for i, caCert := range caCertificates {
+			if ok := cp.AppendCertsFromPEM(caCert); !ok {
+				return nil, fmt.Errorf("failed to append ca certificate [%d] from PEM", i)
+			}
+		}
+	}
+	cert, certerr := tls.X509KeyPair(certificate, key)
+	if certerr != nil {
+		return nil, fmt.Errorf("failed to append certificate from PEM")
+	}
+	return &tls.Config{
+		RootCAs:      cp,
+		Certificates: []tls.Certificate{cert},
+		MinVersion:   tls.VersionTLS12,
+	}, nil
+}
+
+func buildTLSConfigFromMap(tlsConfigMap map[string]interface{}) (*tls.Config, error) {
+	var cert, key []byte
+	var ca [][]byte
+	if certstr, ok := tlsConfigMap["cert"].(string); ok {
+		cert = []byte(certstr)
+	}
+	if keystr, ok := tlsConfigMap["key"].(string); ok {
+		key = []byte(keystr)
+	}
+	cas := tlsConfigMap["cacerts"]
+	ca = make([][]byte, 0)
+	if casarr, ok := cas.([]string); ok {
+		for _, casEntry := range casarr {
+			ca = append(ca, []byte(casEntry))
+		}
+	} else if casstr, casstrok := cas.(string); casstrok {
+		ca = append(ca, []byte(casstr))
+	}
+	return buildTLSConfig(cert, key, ca)
+}
+
 // Connect is a block dial to the gRPC server at the given address (host:port)
 func (c *Client) Connect(addr string, params map[string]interface{}) (bool, error) {
 	state := c.vu.State()
@@ -124,10 +171,16 @@ func (c *Client) Connect(addr string, params map[string]interface{}) (bool, erro
 
 	var tcred credentials.TransportCredentials
 	if !p.IsPlaintext {
-		tlsCfg := state.TLSConfig.Clone()
+		var tlsCfg *tls.Config
+		if len(p.TLSConfig) > 0 {
+			if tlsCfg, err = buildTLSConfigFromMap(p.TLSConfig); err != nil {
+				return false, err
+			}
+		} else {
+			tlsCfg = state.TLSConfig.Clone()
+		}
 		tlsCfg.NextProtos = []string{"h2"}
 
-		// TODO(rogchap): Would be good to add support for custom RootCAs (self signed)
 		tcred = credentials.NewTLS(tlsCfg)
 	} else {
 		tcred = insecure.NewCredentials()
@@ -373,6 +426,7 @@ type connectParams struct {
 	Timeout               time.Duration
 	MaxReceiveSize        int64
 	MaxSendSize           int64
+	TLSConfig             map[string]interface{}
 }
 
 func (c *Client) parseConnectParams(raw map[string]interface{}) (connectParams, error) {
@@ -382,6 +436,7 @@ func (c *Client) parseConnectParams(raw map[string]interface{}) (connectParams, 
 		Timeout:               time.Minute,
 		MaxReceiveSize:        0,
 		MaxSendSize:           0,
+		TLSConfig:             make(map[string]interface{}),
 	}
 	for k, v := range raw {
 		switch k {
@@ -421,7 +476,35 @@ func (c *Client) parseConnectParams(raw map[string]interface{}) (connectParams, 
 			if params.MaxSendSize < 0 {
 				return params, fmt.Errorf("invalid maxSendSize value: '%#v, it needs to be a positive integer", v)
 			}
+		case "tlsconfig":
+			var ok bool
+			params.TLSConfig, ok = v.(map[string]interface{})
+			err := fmt.Errorf("invalid tlsconfig value: '%#v', tlsconfig needs cert, key, and (optionally) cacerts", v)
 
+			if !ok {
+				return params, err
+			}
+			if cert, certok := params.TLSConfig["cert"]; certok {
+				if _, ok = cert.(string); !ok {
+					return params, err
+				}
+			} else {
+				return params, err
+			}
+			if key, keyok := params.TLSConfig["key"]; keyok {
+				if _, ok = key.(string); !ok {
+					return params, err
+				}
+			} else {
+				return params, err
+			}
+			if cacerts, cacertsok := params.TLSConfig["cacerts"]; cacertsok {
+				if _, ok = cacerts.([]string); !ok {
+					if _, ok = cacerts.(string); !ok {
+						return params, err
+					}
+				}
+			}
 		default:
 			return params, fmt.Errorf("unknown connect param: %q", k)
 		}
--- a/js/modules/k6/execution/execution_test.go
+++ b/js/modules/k6/execution/execution_test.go
@@ -223,7 +223,7 @@ func TestAbortTest(t *testing.T) { //nolint:tparallel
 func TestOptionsTestFull(t *testing.T) {
 	t.Parallel()
 
-	expected := `{"paused":true,"scenarios":{"const-vus":{"executor":"constant-vus","options":{"browser":{"someOption":true}},"startTime":"10s","gracefulStop":"30s","env":{"FOO":"bar"},"exec":"default","tags":{"tagkey":"tagvalue"},"vus":50,"duration":"10m0s"}},"executionSegment":"0:1/4","executionSegmentSequence":"0,1/4,1/2,1","noSetup":true,"setupTimeout":"1m0s","noTeardown":true,"teardownTimeout":"5m0s","rps":100,"dns":{"ttl":"1m","select":"roundRobin","policy":"any"},"maxRedirects":3,"userAgent":"k6-user-agent","batch":15,"batchPerHost":5,"httpDebug":"full","insecureSkipTLSVerify":true,"tlsCipherSuites":["TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256"],"tlsVersion":{"min":"tls1.2","max":"tls1.3"},"tlsAuth":[{"domains":["example.com"],"cert":"mycert.pem","key":"mycert-key.pem","password":"mypwd"}],"throw":true,"thresholds":{"http_req_duration":[{"threshold":"rate>0.01","abortOnFail":true,"delayAbortEval":"10s"}]},"blacklistIPs":["192.0.2.0/24"],"blockHostnames":["test.k6.io","*.example.com"],"hosts":{"test.k6.io":"1.2.3.4:8443"},"noConnectionReuse":true,"noVUConnectionReuse":true,"minIterationDuration":"10s","ext":{"ext-one":{"rawkey":"rawvalue"}},"summaryTrendStats":["avg","min","max"],"summaryTimeUnit":"ms","systemTags":["iter","vu"],"tags":null,"metricSamplesBufferSize":8,"noCookiesReset":true,"discardResponseBodies":true,"consoleOutput":"loadtest.log","tags":{"runtag-key":"runtag-value"},"localIPs":"192.168.20.12-192.168.20.15,192.168.10.0/27"}`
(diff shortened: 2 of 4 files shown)


js/modules/k6/grpc/client.go
```diff
@@ -373,6 +426,7 @@ type connectParams struct {
 	Timeout               time.Duration
 	MaxReceiveSize        int64
 	MaxSendSize           int64
+	TLSConfig             map[string]interface{}
```
imiric, line 429: Rename this to just `TLS`? `connectParams` is already a configuration object, so `Config` seems redundant.

js/modules/k6/grpc/client.go
```diff
@@ -421,7 +476,35 @@ func (c *Client) parseConnectParams(raw map[string]interface{}) (connectParams,
 			if params.MaxSendSize < 0 {
 				return params, fmt.Errorf("invalid maxSendSize value: '%#v, it needs to be a positive integer", v)
 			}
+		case "tlsconfig":
```
imiric, line 479: Can you add some test cases for this option?

imiric (changes requested):
Hi there, thanks a lot for this PR! :bow:

I think it makes sense to do this for the gRPC client, but I'm not so sure about the global `tlsAuth` change. And I'd like to see some gRPC tests for this.

js/modules/k6/grpc/client.go
```diff
@@ -108,6 +110,51 @@ func (c *Client) LoadProtoset(protosetPath string) ([]MethodInfo, error) {
 	return c.convertToMethodInfo(fdset)
 }
 
+func buildTLSConfig(certificate, key []byte, caCertificates [][]byte) (*tls.Config, error) {
+	var cp *x509.CertPool
+	var err error
+	if cp, err = x509.SystemCertPool(); err != nil {
+		return nil, err
+	}
+	if len(caCertificates) > 0 {
```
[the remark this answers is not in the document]
cmmoran: I think I got carried away with the `make lint` results and put pedantic checks everywhere to avoid `make lint` issues...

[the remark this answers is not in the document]
chrismoran-mica: I think I got carried away with responding to the results of `make lint` and went a bit too far with some of the pedantic checks...

chrismoran-mica:
Thank you for your feedback @imiric.

I've addressed all of the feedback requests and added some (admittedly not 100% coverage) test cases for the new connectParams `tls` option. Let me know what I've missed that must/should be included before merge.

lib/testutils/httpmultibin/httpmultibin.go
```diff
@@ -395,6 +481,9 @@ func NewHTTPMultiBin(t testing.TB) *HTTPMultiBin {
 			"HTTP2BIN_PORT", http2URL.Port(),
 
 			"GRPCBIN_ADDR", fmt.Sprintf("%s:%s", httpsDomain, http2URL.Port()),
+			"LOCALHOST_CERT", strings.ReplaceAll(localhostCert, "\n", "\\n"),
+			"LOCALHOST_KEY", strings.ReplaceAll(localhostKey, "\n", "\\n"),
+			"LOCALHOST_ENCRYPTED_KEY", strings.ReplaceAll(localhostEncryptedKey, "\n", "\\n"),
```
mstoykov, line 486: Isn't this only used for the gRPC test - it isn't even used anywhere in the actuall httpmultibin for anything?

js/modules/k6/grpc/client.go
```diff
@@ -421,7 +510,43 @@ func (c *Client) parseConnectParams(raw map[string]interface{}) (connectParams,
 			if params.MaxSendSize < 0 {
 				return params, fmt.Errorf("invalid maxSendSize value: '%#v, it needs to be a positive integer", v)
 			}
+		case "tls":
+			var ok bool
+			params.TLS, ok = v.(map[string]interface{})
+			err := fmt.Errorf("invalid tls value: '%#v', tls needs cert, key, (optional) password, and (optionally) cacerts", v)
```
mstoykov, line 516: I am not a fan of us doing an allocation of this error even for every time there is no error.

Looking at the code - this is just validating stuff so `buildTLSConfigFromMap` can do 80% of the work ... but not return those errors.

But I would argue we can just have `params.TLS` be the result of `buildTLSConfigFromMap`, call it here with `v` and do all the validation there. 

I am not certain if `params.TLS`/`params` goes to js in some way and that makes this not possible. But it doesn't seem like it.

js/modules/k6/grpc/client_test.go
```diff
@@ -261,6 +262,123 @@ func TestClient(t *testing.T) {
 				client.load([], "../../../../lib/testutils/httpmultibin/grpc_testing/test.proto");`},
 			vuString: codeBlock{code: `client.connect("GRPCBIN_ADDR", { timeout: 3456.3 });`},
 		},
+		{
```
mstoykov, line 265: None of this test seem to test that we do *actually* use the configured tlsconfig to do the connection and error out if we use the wrong certificate

mstoykov (changes requested):
LGTM in general! Thanks for the contribution :bow: 

I have left a bunch of change requests and some questions.

My two concerns (which are not inline) are:
1. Organizational - we now have an experimental grpc module where we have some changes. This change will now need to be transfered there. While we might've prefered it the other way around. cc @olegbespalov 
2. API one - k6 has problems with [big files](https://github.com/grafana/k6/issues/2311) arguably with the handling of any files to be honest, just big ones make it worse. This particular case *seems* okay to me. Especially if users figure out how to load the key/cert once . Whihc arguably is doable with a SharedArray with element per each cert+key+password. My other problem is that currently you went with the `key` and `cert` being *strings* instead of binary data. Which I guess is based on what formats golang supports? I do ask the same question inline.

And both of those questions could've been answered  if we had an issue to discuss this before hand. 

p.s. Your example in the description shows "cert" (and all the other big options) as what looks like file names. But they are actually the contents of those. I would prefer to have the example either just use `open("./cert.pem")` or be the string (with a lot of `...`) that is the actual contents. As in this way it is confusing for both me as reviewer and likely anyone else who in the future tries to figure out what was added here.

lib/testutils/httpmultibin/httpmultibin.go
chrismoran-mica: Correct.

That said, GRPCBIN_ADDR is also only used in runner_test.go and grpc/client_test.go. I thought those Replacer entries would fit there. I'm open to other suggestions.

I can modify the httpmultibin to support "appending" Replacer entries. I'm not certain what the additional impact would be.

js/modules/k6/grpc/client.go
```diff
@@ -108,6 +111,84 @@ func (c *Client) LoadProtoset(protosetPath string) ([]MethodInfo, error) {
...
+		return nil, fmt.Errorf("failed to decode PEM key")
+	}
+
+	blockType := block.Type
+	if blockType == "ENCRYPTED PRIVATE KEY" {
+		return nil, fmt.Errorf("encrypted pkcs8 formatted key is not supported")
+	}
+	/*
+	   Even though `DecryptPEMBlock` has been deprecated since 1.16.x it is still
+	   being used here because it is deprecated due to it not supporting *good* crypography
```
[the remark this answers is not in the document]
chrismoran-mica: :) As I copy/pasted I did, indeed, copy the spelling mistake. See: https://github.com/grafana/k6/blob/7ab37dd210c8a01400738125be15e0df80e422cb/lib/options.go#L158C1-L184C2

js/modules/k6/grpc/client.go
```diff
@@ -108,6 +111,84 @@ func (c *Client) LoadProtoset(protosetPath string) ([]MethodInfo, error) {
...
+	   ultimately though we want to support something so we will be using it for now.
+	*/
```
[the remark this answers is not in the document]
chrismoran-mica: As with above. It came with the copy/paste.
https://github.com/grafana/k6/blob/7ab37dd210c8a01400738125be15e0df80e422cb/lib/options.go#L158C1-L184C2

js/modules/k6/grpc/client.go
```diff
@@ -108,6 +111,84 @@ func (c *Client) LoadProtoset(protosetPath string) ([]MethodInfo, error) {
...
+	block, _ := pem.Decode(key)
+	if block == nil {
+		return nil, fmt.Errorf("failed to decode PEM key")
+	}
+
+	blockType := block.Type
+	if blockType == "ENCRYPTED PRIVATE KEY" {
+		return nil, fmt.Errorf("encrypted pkcs8 formatted key is not supported")
+	}
+	/*
```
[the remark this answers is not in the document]
chrismoran-mica: Same.
https://github.com/grafana/k6/blob/7ab37dd210c8a01400738125be15e0df80e422cb/lib/options.go#L158C1-L184C2

js/modules/k6/grpc/client.go
```diff
@@ -108,6 +111,84 @@ func (c *Client) LoadProtoset(protosetPath string) ([]MethodInfo, error) {
...
+		}
+	}
+	cert, certerr := tls.X509KeyPair(certificate, key)
+	if certerr != nil {
+		return nil, fmt.Errorf("failed to append certificate from PEM")
+	}
+	return &tls.Config{
+		RootCAs:      cp,
+		Certificates: []tls.Certificate{cert},
+		MinVersion:   tls.VersionTLS12,
```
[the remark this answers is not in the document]
chrismoran-mica: None; whatsoever. Nice catch, I believe this, again, would be a copy/paste issue (but this time from something of my own internally)

js/modules/k6/grpc/client_test.go
cmmoran: Solid point. I'll sort this out.

chrismoran-mica: Solid point. Working on this now.

lib/testutils/httpmultibin/httpmultibin.go
chrismoran-mica: I've reverted the changes I made to httpmultibin.go and inlined my test data for my changes to a separate TestClient function within client_test.go

js/modules/k6/grpc/client.go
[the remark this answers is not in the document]
chrismoran-mica: I've added a `//nolint:golint,gosec` comment for this particular line of code. I've also modified the tls.Config that is created so that it inherits `CipherSuites`, `InsecureSkipVerify`, `MinVersion`, `MaxVersion`, and `Renegotiation` from the `state.TLSConfig.Clone()`

js/modules/k6/grpc/client.go
chrismoran-mica: Agreed. Refactored.

chrismoran-mica:
@mstoykov

I've updated my trivial examples with `open` statements to address your concern mentioned above.

chrismoran-mica:
I believe I've addressed all of the feedback so far. Please let me know what else I can do to prepare this for v0.46.0.

chrismoran-mica:
> 1. Organizational - we now have an experimental grpc module where we have some changes. This change will now need to be transfered there. While we might've prefered it the other way around. cc @olegbespalov
> 2. API one - k6 has problems with https://github.com/grafana/k6/issues/2311 arguably with the handling of any files to be honest, just big ones make it worse. This particular case seems okay to me. Especially if users figure out how to load the key/cert once . Whihc arguably is doable with a SharedArray with element per each cert+key+password. My other problem is that currently you went with the key and cert being strings instead of binary data. Which I guess is based on what formats golang supports? I do ask the same question inline.

> And both of those questions could've been answered if we had an issue to discuss this before hand.


1. Is there anything I can do to facilitate this?
2. Loading keys/certs are fairly trivial using `SharedArray`. The uncertainty / confusion may come with the fact that each VU has an _individual_ instance of `grpc.Client()` - I'm quite confident this will trip folks up. It may be obvious to many, but to me, it was not; I figured it was possible to share a single `grpc.Client` across multiple VU instances.

I apologize for not raising an issue beforehand. The simple truth is that I was so focused on getting something working within my organization that it definitely did not occur to me to create an issue for this prior to "making it work".

Apologies.

js/modules/k6/grpc/client.go
```diff
@@ -124,10 +222,16 @@ func (c *Client) Connect(addr string, params map[string]interface{}) (bool, erro
...
 	if !p.IsPlaintext {
-		tlsCfg := state.TLSConfig.Clone()
+		var tlsCfg *tls.Config
+		if len(p.TLS) > 0 {
+			if tlsCfg, err = buildTLSConfigFromMap(state.TLSConfig.Clone(), p.TLS); err != nil {
+				return false, err
+			}
+		} else {
+			tlsCfg = state.TLSConfig.Clone()
+		}
```
mstoykov, line 232:
```suggestion
		tlsCfg := state.TLSConfig.Clone()
		if len(p.TLS) > 0 {
			if tlsCfg, err = buildTLSConfigFromMap(tlsCfg, p.TLS); err != nil {
				return false, err
			}
		}
```

mstoykov (approved):
LGTM! thanks for the contribution.

js/modules/k6/grpc/client.go
```diff
@@ -108,6 +111,101 @@ func (c *Client) LoadProtoset(protosetPath string) ([]MethodInfo, error) {
 	return c.convertToMethodInfo(fdset)
 }
 
+// Note: this function was lifted from `lib/options.go`
+func decryptPrivateKey(key, password []byte) ([]byte, error) {
+	block, _ := pem.Decode(key)
+	if block == nil {
+		return nil, fmt.Errorf("failed to decode PEM key")
```
olegbespalov, line 118:
```suggestion
		return nil, errors.New("failed to decode PEM key")
```

js/modules/k6/grpc/client.go
```diff
@@ -108,6 +111,101 @@ func (c *Client) LoadProtoset(protosetPath string) ([]MethodInfo, error) {
...
+	}
+
+	blockType := block.Type
+	if blockType == "ENCRYPTED PRIVATE KEY" {
+		return nil, fmt.Errorf("encrypted pkcs8 formatted key is not supported")
```
olegbespalov, line 123:
```suggestion
		return nil, errors.New("encrypted pkcs8 formatted key is not supported")
```

js/modules/k6/grpc/client.go
```diff
@@ -421,7 +526,43 @@ func (c *Client) parseConnectParams(raw map[string]interface{}) (connectParams,
...
+					return params, fmt.Errorf("invalid tls password value: '%#v', it needs to be a string", v)
+				}
+			}
+			if cacerts, cacertsok := params.TLS["cacerts"]; cacertsok {
+				var cacertsArray []interface{}
+				if cacertsArray, ok = cacerts.([]interface{}); ok {
+					for _, cacertsArrayEntry := range cacertsArray {
+						if _, ok = cacertsArrayEntry.(string); !ok {
+							return params, fmt.Errorf("invalid tls cacerts value: '%#v',"+
+								" it needs to be a string or string[] of PEM formatted strings", v)
```
olegbespalov, line 558: Just to make this more human-oriented :relaxed: feel free to ignore
```suggestion
								" it needs to be a string or an array of PEM formatted strings", v)
```

js/modules/k6/grpc/client.go
```diff
@@ -421,7 +526,43 @@ func (c *Client) parseConnectParams(raw map[string]interface{}) (connectParams,
...
+						}
+					}
+				} else if _, ok = cacerts.(string); !ok {
+					return params, fmt.Errorf("invalid tls cacerts value: '%#v',"+
+						" it needs to be a string or string[] of PEM formatted strings", v)
```
olegbespalov, line 563:
```suggestion
						" it needs to be a string or an array of PEM formatted strings", v)
```

olegbespalov (dismissed):
LGTM, left a few non-blocking suggestions :+1: 

Thanks for the contribution!

olegbespalov (dismissed):

mstoykov (approved):

chrismoran-mica:
I approved the requested changes but unfortunately one of them broke a couple tests :)

I've fixed them I believe. Sorry for the ping-pong!

olegbespalov (approved):

mstoykov (approved):

mstoykov:
For future reference while squashing the changes I did delete the mention of the global `tlsAuth`. 

But unfortunately did not fix the example to use the actual contents of a pem file and instead it still has the pem file name :facepalm: . 

This likely doesn't really matter as both the current PR body and any documentation will have this correct.

chrismoran-mica:
I updated the example to use “open”. 🤔On Jul 12, 2023, at 12:07 PM, Mihail Stoykov ***@***.***> wrote:﻿
For future reference while squashing the changes I did delete the mention of the global tlsAuth.
But unfortunately did not fix the example to use the actual contents of a pem file and instead it still has the pem file name 🤦 .
This likely doesn't really matter as both the current PR body and any documentation will have this correct.

mstoykov:
@chrismoran-mica I should've been more specific. This wasn't your fault. 
When you squash on github it presents you with the commit messages not the PR description. 

Arguably you could've changed the commit message and force pushed, but *I* would probably also forget to do it or not even think of it.

I should've just been more careful and arguably just written a completely knew message when squashing. It is my fault

PR #3162, opened 2023-07-04
feat: catch, discard & warn about the labels that have reserved names

### LINKED ISSUE #3155: Warn and discard any label equal to the reserved
# What?

Figure out what reserved labels are and discard their usage with a warning.

# Why?

Labels like `__name__` and `test_run_id` are labels used by the k6 cloud backend, so if a test defines a set of tags that contains one of them, it will generate a conflict.

olegbespalov:
## What?

Cloud Output V2 drops labels that are reserved, for now, we're sticking with the `test_run_id` and Prometheus system metrics (everything that prefixed with `__`) 

## Why?

These tags are reserved for internal use and can't be set with the client.

## Checklist

## Related PR(s)/Issue(s)

Closes #3155

### DIFF
--- a/output/cloud/expv2/flush.go
+++ b/output/cloud/expv2/flush.go
@@ -74,7 +74,7 @@ func (f *metricsFlusher) push(msb metricSetBuilder) error {
 		}
 
 		f.discardedLabels[key] = struct{}{}
-		f.logger.Warnf("Label %s was discarded since it can't be used with the cloud output", key)
+		f.logger.Warnf("Tag %s has been discarded since it is reserved for Cloud operations.", key)
 	}
 
 	return f.client.push(msb.MetricSet)
@@ -107,7 +107,7 @@ type metricSetBuilder struct {
 	seriesIndex map[metrics.TimeSeries]uint
 
 	// discardedLabels tracks the labels that have been discarded
-	// because they are not supported by the remote service.
+	// since they are reserved for internal usage by the Cloud service.
 	discardedLabels map[string]struct{}
 }
 
@@ -157,12 +157,16 @@ func (msb *metricSetBuilder) addTimeBucket(bucket timeBucket) {
 	}
 }
 
-func (msb *metricSetBuilder) recordDiscardedLabels(labels map[string]struct{}) {
+func (msb *metricSetBuilder) recordDiscardedLabels(labels []string) {
+	if len(labels) == 0 {
+		return
+	}
+
 	if msb.discardedLabels == nil {
 		msb.discardedLabels = make(map[string]struct{})
 	}
 
-	for key := range labels {
+	for _, key := range labels {
 		if _, ok := msb.discardedLabels[key]; ok {
 			continue
 		}
--- a/output/cloud/expv2/flush.go
+++ b/output/cloud/expv2/flush.go
@@ -118,7 +118,7 @@ func newMetricSetBuilder(testRunID string, aggrPeriodSec uint32) metricSetBuilde
 		// is a better trade-off
 		metrics:         make(map[*metrics.Metric]*pbcloud.Metric),
 		seriesIndex:     make(map[metrics.TimeSeries]uint),
-		discardedLabels: make(map[string]struct{}),
+		discardedLabels: nil,
 	}
 	builder.MetricSet.TestRunId = testRunID
 	builder.MetricSet.AggregationPeriod = aggrPeriodSec
@@ -158,6 +158,10 @@ func (msb *metricSetBuilder) addTimeBucket(bucket timeBucket) {
 }
 
 func (msb *metricSetBuilder) recordDiscardedLabels(labels map[string]struct{}) {
+	if msb.discardedLabels == nil {
+		msb.discardedLabels = make(map[string]struct{})
+	}
+
 	for key := range labels {
 		if _, ok := msb.discardedLabels[key]; ok {
 			continue
--- a/output/cloud/expv2/flush.go
+++ b/output/cloud/expv2/flush.go
@@ -3,6 +3,7 @@ package expv2
 import (
 	"context"
 
+	"github.com/sirupsen/logrus"
 	"go.k6.io/k6/cloudapi/insights"
 	"go.k6.io/k6/metrics"
 	"go.k6.io/k6/output/cloud/expv2/pbcloud"
@@ -16,6 +17,8 @@ type metricsFlusher struct {
 	referenceID                string
 	bq                         *bucketQ
 	client                     pusher
+	logger                     logrus.FieldLogger
+	discardedLabels            map[string]struct{}
 	aggregationPeriodInSeconds uint32
 	maxSeriesInBatch           int
 }
@@ -47,7 +50,7 @@ func (f *metricsFlusher) flush(_ context.Context) error {
 		}
 
 		// we hit the chunk size, let's flush
-		err := f.client.push(msb.MetricSet)
+		err := f.push(msb)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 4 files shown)


output/cloud/expv2/mapping.go
```diff
@@ -10,21 +11,37 @@ import (
 )
 
 // TODO: unit test
-func mapTimeSeriesLabelsProto(tags *metrics.TagSet) []*pbcloud.Label {
+func mapTimeSeriesLabelsProto(tags *metrics.TagSet) ([]*pbcloud.Label, map[string]struct{}) {
 	labels := make([]*pbcloud.Label, 0, ((*atlas.Node)(tags)).Len())
+	discardedLabels := make(map[string]struct{})
```
codebien, line 16: Considering that this is expected to be empty most of the time, I'm for not allocating it and allocate only if we find at least one item.

output/cloud/expv2/flush.go
```diff
@@ -87,15 +105,20 @@ type metricSetBuilder struct {
...
 func newMetricSetBuilder(testRunID string, aggrPeriodSec uint32) metricSetBuilder {
 	builder := metricSetBuilder{
 		MetricSet: &pbcloud.MetricSet{},
 		// TODO: evaluate if removing the pointer from pbcloud.Metric
 		// is a better trade-off
-		metrics:     make(map[*metrics.Metric]*pbcloud.Metric),
-		seriesIndex: make(map[metrics.TimeSeries]uint),
+		metrics:         make(map[*metrics.Metric]*pbcloud.Metric),
+		seriesIndex:     make(map[metrics.TimeSeries]uint),
+		discardedLabels: make(map[string]struct{}),
```
codebien, line 121: We will invoke this code with big frequency and we should hit this some 0.x percentage so better keep it nil most of the time and init only if we need.

codebien (changes requested):
It requires minimizing the allocation because we invoke this code with high frequency.

output/cloud/expv2/mapping.go
olegbespalov: :heavy_check_mark: with the latest commit

output/cloud/expv2/flush.go
olegbespalov: :heavy_check_mark: with the latest commit

output/cloud/expv2/flush.go
```diff
@@ -59,6 +62,21 @@ func (f *metricsFlusher) flush(_ context.Context) error {
...
+// push sends the metric set to the remote service.
+// it also checks if the labels are discarded and logs a warning if so.
+func (f *metricsFlusher) push(msb metricSetBuilder) error {
+	for key := range msb.discardedLabels {
+		if _, ok := f.discardedLabels[key]; ok {
+			continue
+		}
+
+		f.discardedLabels[key] = struct{}{}
+		f.logger.Warnf("Label %s was discarded since it can't be used with the cloud output", key)
```
codebien, line 77:
```suggestion
		f.logger.Warnf("Tag %s was discarded since it is reserved for the Cloud operations.", key)
```
1. The user doesn't know the Label concept for k6. They set a Tag.
2. It isn't an issue of the Cloud output but more of the Cloud platform (I don't know how 
3. I think the _why_ isn't clear and the user could care otherwise it could be confused.

codebien (commented):
LGTM :bow: Just a minor on the warn message.

output/cloud/expv2/mapping.go
```diff
@@ -10,21 +11,41 @@ import (
 )
 
 // TODO: unit test
-func mapTimeSeriesLabelsProto(tags *metrics.TagSet) []*pbcloud.Label {
+func mapTimeSeriesLabelsProto(tags *metrics.TagSet) ([]*pbcloud.Label, map[string]struct{}) {
```
codebien, line 14:
```suggestion
func mapTimeSeriesLabelsProto(tags *metrics.TagSet) ([]*pbcloud.Label, []string) {
```

output/cloud/expv2/flush.go
```diff
@@ -131,6 +157,20 @@ func (msb *metricSetBuilder) addTimeBucket(bucket timeBucket) {
 	}
 }
 
+func (msb *metricSetBuilder) recordDiscardedLabels(labels map[string]struct{}) {
```
codebien, line 160:
```suggestion
func (msb *metricSetBuilder) recordDiscardedTags(keys []string) {
```

output/cloud/expv2/flush.go
```diff
@@ -131,6 +157,20 @@ func (msb *metricSetBuilder) addTimeBucket(bucket timeBucket) {
...
+	if msb.discardedLabels == nil {
```
codebien, line 161:
```suggestion
    if keys == nil {
        return
    }
	if msb.discardedLabels == nil {
```

codebien (changes requested):
Just caught only now. Probably, we can just use a slice storing the tags' keys during the detection process. A TagSet shouldn't include multiple tags with the same key.

output/cloud/expv2/flush.go
olegbespalov: I'd prefer to keep the `lables` naming here since, in the proto we call them labels.

output/cloud/expv2/mapping.go
```diff
@@ -10,21 +11,41 @@ import (
...
+		}
+
+		discardedLabels = append(discardedLabels, key)
 	}
-	return labels
+	return labels, discardedLabels
+}
+
+func isReservedLabelName(name string) bool {
+	// this is a reserved label prefix for the prometheus
```
codebien, line 43:
```suggestion
	// this is a reserved label pattern for Prometheus, that is used in the Cloud's remote service.
```
Someone of us could ask.."ok, why do we care about it in k6 which doesn't use Prometheus?"

output/cloud/expv2/mapping.go
```diff
@@ -10,21 +11,41 @@ import (
...
+	if strings.HasPrefix(name, "__") {
```
codebien, line 44:
```suggestion
	if strings.HasPrefix(name, "__") && strings.HasSuffix("__") {
```
In theory, `__my_amazing_label_` should be a legit tag key. Should we care about it?

olegbespalov: [According convention](https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels:~:text=Label%20names%20beginning%20with%20__%20(two%20%22_%22)%20are%20reserved%20for%20internal%20use) it's not since it begins with `__`

codebien: Yeah, you're right. I should double-check one more the next time. :pray:

codebien (approved):

olegbespalov:
@esquonk, adding you also as the reviewer, probably the most interesting part for you is https://github.com/grafana/k6/pull/3162/files#diff-2edc8a9a0a0d4ec2a88767baae9d8ad5c035168eeee41b7b5c8b36ce31d9ce70R42-R49

output/cloud/expv2/mapping.go
```diff
@@ -10,21 +11,41 @@ import (
...
+		return true
+	}
+
+	return name == "test_run_id"
```
oleiade, line 48: Feel free to ignore, but I'd suggest putting the static string in a constant.

oleiade (approved):
👏🏻 🙇🏻

output/cloud/expv2/mapping.go
olegbespalov: I considered the constant but decided to keep the context tinier (limited by a method). If we use it somewhere else in the application logic, we should move it to the const.

Issue #3141, opened 2023-06-27
Implement File API


oleiade:
We have agreed on [a design](https://github.com/grafana/k6/blob/master/docs/design/019-file-api.md) to tackle #2977.

This issue aims at tackling the implementation of the agreed specification. A reference [proof of concept implementation](https://github.com/grafana/k6/tree/poc/experimental/fs-module/js/modules/k6/experimental/fs) can be used as a reference for the implementation.

```[tasklist]
### Implementation Steps
- [ ] https://github.com/grafana/k6/issues/3142
- [ ] https://github.com/grafana/k6/issues/3354
- [ ] https://github.com/grafana/k6/issues/3143
- [ ] https://github.com/grafana/k6/issues/3145
- [ ] https://github.com/grafana/k6/issues/3147
- [ ] https://github.com/grafana/k6/issues/3148
```

Issue #2974, opened 2023-03-13
Improving k6 large files handling


oleiade:
# Story

## Problem Statement

Handling large files in k6, whether binary or structured format files such as CSV, leads to high memory usage. As a result, our user’s experience, especially in the cloud, is degraded as soon as they need to handle large data sets; of user ids, for instance.

This issue is experienced by our users in various situations and is caused by many design and implementation decisions in the current state of the k6 open-source tool.

This workgroup’s main objective is to approach this issue from a specific scope to discover what user-experience changes, new APIs, and implementation changes might be necessary to cater to a better user experience when using large files in k6.

### Product-oriented

The product-oriented objective of this story, and definition of success, will consist in landing a new streaming CSV parser in k6, allowing users to parse and use big CSV files (> 500MB) that wouldn’t fit in memory and likely lead their k6 scripts to crash. We are keen to provide any technical improvements to k6 that make it possible along the way.

### Technology Oriented

From a technological standpoint, the objective of this story is to provide all the design and technical changes necessary to complete the product story. Our primary objective should be to support the ability of users to use large data set files in their k6 scripts without leading to out-of-memory errors. As we pursue this goal, we aim to pick the solutions with the most negligible overhead and are keen to take on technical debt if necessary.

## Resolution

Through the courses of [workshops with @sniku and @mstoykov](https://docs.google.com/document/d/1ubZsjxGuHp5X1faUOcy2JEd7kT3Mu5-I8pdIbEBEejg/edit?usp=sharing), we surfaced a collection of topics and issues that would need to be addressed to fulfill the objective.

### Must-have

The bare minimum must have items to be able to even start tackling the top-level product objective would be: 
1. Support for direct access to files from a tar archive stored on disk

### Nice to have

As we're at it, some other set of features and refactors would be beneficial to the larger story of handling large files in k6

1. Improved file access API and internal design
2. Streams API (readable) support in k6

na--:
While thinking about https://github.com/grafana/k6/issues/2975#issuecomment-1495761640, I realized that I probably disagree with something here. I think the first points of the "Must-have" and "Nice to have" sections are somewhat flipped and need to be exchanged :sweat_smile: That is, https://github.com/grafana/k6/issues/2975 is probably nice to have, while https://github.com/grafana/k6/issues/2977 seems like a must. 

[...]

Issue #3020, opened 2023-04-13
Let `open` and `require` be used outside of the init context, somewhat


mstoykov:
## Description

Both `open` and `require` are only allowed in the init context. 
This is in big parts so that `k6 archive script.js` (and consequently `k6 cloud` and any future `k6 distributed-run ...`) to be able to gather all the files it needs by runnign the init context a single time.

This still let users `open` files randomly in the init context and we made this more consistent with https://github.com/grafana/k6/pull/2314. Where only files opened by the first VU to initialize are openable in the rest of the VUs as well.

## Proposal

Let `open` be used wherever, but only files opened in the original init context can be opened.

Do the same for `require`. Probably a good idea to also limit which `k6/*` are imported as well ... at least in the beginning.

### Open
- Only open one of many binary files when the given iteration needs it
- Only open files in the scenarios that need them.

### Require
- Only require a big library in the scenarios that need it. 
- Only require a big library if the need arises.

## Caveats

This isn't meant to fix any of the Use case entirely.

But the droppage of the `open` restriction will in practice be simplifying the implementation.

`require` likely should get restrictions similar to #2314 and tests to confirm that those work for:
1. local files
2. remote files
3. internal modules

[Dynamic import](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import) in practice will have similar problems like `require`. 

We can always relax all of those APIs in the future if we decided to.

pablochacin:
> Let open be used wherever, but only files opened in the original init context can be opened.
The phrasing here sounds strange: how can a file opened in the init be opened elsewhere?

mstoykov:
> The phrasing here sounds strange: how can a file opened in the init be opened elsewhere?

I am not certain I get your question. 

It currently is *not* possible and this is basically issue to *lift* this limitation. 

Specifically currently 
```javascript
if (__VU == 0) { // we likely should have better way to detect this that only runs on the really first init context
  open("./test0.bin");
  open("./test1.bin");
  // more opens
  open("./test99.bin");
}

export default function() {
  open(`./test${__ITER%100}.bin`); // this will never work and will tell you `open` can't be used outside the init context.
}
```
Does that answer your question @pablochacin ?

na--:
I think we should probably not mess with `open()` and `require()` anymore :thinking: 

[...]

PR #3165, opened 2023-07-05
[fs] Expose an experimental `fs` module with an `open` function and a `File` abstraction (1/3)

### LINKED ISSUE #3142: Implement the File API's `open` method, base `File` abstraction and `File.stat` operation.
As part of #3141, implementing the [agreed-upon design for k6's File API](https://github.com/grafana/k6/blob/master/docs/design/019-file-api.md), we intend to:

1. declare a dedicated experimental k6 module
2. add an `open` function to it, as specified
3. add an almost empty vessel `File` abstraction so that the `open` method does return an object, albeit not functional as part of this iteration.
4. implement the `File.stat` method to retrieve file information about the file itself
5. implement the `FileInfo` abstraction to fulfill 👆 

### Details

In this implementation step, we should declare and implement the `open` function, as showcased in the dedicated design document. This function inputs a `path` string and returns a Promises resolving to a `File` instance. To give some minimal usefulness to the abstraction, we expect to have the `File.stat` method and its `FileInfo` construct counterparts exposed alongside it.

We expect the `open` function to perform all the underlying operations it should, regardless of whether we expose helpful methods on the resulting `File` object yet:
- resolve, clean, and secure the path to the resource
- ensure it exists and is accessible
- open it
- map the file descriptor and track it internally in the module, as described in the design doc, and in a fashion that allows us to fulfill the scope of the `stat` method.

The result would look something like this
```javascript
open(path: string): Promise<File>

interface File {
	stat(): Promise<FileInfo>
}

interface FileInfo {
	name: string
	size: number
}
```

oleiade:
## What?

This Pull Request aims to address #3142 and is the first step towards introducing a `fs` module to k6 as tracked by #3141.

In this Pull Request we:
- introduce a `k6/experimental/fs` module
- exposing:
  - an `open` function that ensures the file is loaded in memory once, and returns a `File` object to the user.
  - a `File` object abstraction to the JS runtime: which currently supports a single `Stat` method and exposes a `Name` property.
  - a `FileInfo` abstraction that is returned by the `File.stat` method and exposes the `name` (basename) of the file and its `size` as properties to the JS runtime.

## Checklist

## Related PR(s)/Issue(s)

closes #3142

ref #3141
ref #2974


oleiade:
🗣️  the "convert to draft" button doesn't seem to work at the moment...

js/modules/k6/experimental/fs/module.go
```diff
@@ -0,0 +1,156 @@
...
+			"FileInfo": fileInfo{},
+		},
+	}
+}
+
+// Open opens a file and returns a promise that will resolve to a [File] instance
+func (mi *ModuleInstance) Open(path goja.Value) *goja.Promise {
+	rt := mi.vu.Runtime()
+
+	promise, resolve, reject := makeHandledPromise(mi.vu)
```
imiric, line 68: As part of that goja abstraction layer we discussed, what I meant was that we could have these user-facing methods wrap internal methods that actually have the implementation. This way we can avoid interacting with goja in most of our tests, which would simplify things, and then have just a few goja-level tests to ensure that the abstraction layer is working properly.

So in this case we could have an internal `open()` method that takes `pathStr` and works with `FileSystems["file"]`, without any of the promises and async behavior. And in this user-facing `Open()` we can do the `goja.Value` extraction and async implementation that calls the internal `open()`. This can then possibly be reused for other user-facing methods as well.

js/modules/k6/experimental/fs/module.go
```diff
@@ -0,0 +1,156 @@
...
+}
+
+// Exports implements the modules.Module interface and returns the exports of
+// our module.
+func (mi *ModuleInstance) Exports() modules.Exports {
+	return modules.Exports{
+		Named: map[string]any{
+			"open":     mi.Open,
+			"File":     file{},
+			"FileInfo": fileInfo{},
```
imiric, line 59: We don't expect these to be directly usable in scripts, right? I.e. they're only exported in case the user wants to do some `instanceof` checks, or something like that?

imiric (commented):
Good first step @oleiade! :clap:

Just have some minor and nitpicky comments/questions.

js/modules/k6/experimental/fs/registry.go
```diff
@@ -0,0 +1,57 @@
...
+// open opens the named file for reading.
+//
+// If the file was already opened, it returns a pointer to the cached file data. Otherwise, it
+// opens the file, reads its content, caches it, and returns a pointer to it.
+//
+// The file is always opened in read-only mode. The provided file path is cleaned before being
+// used.
+func (fr *fileRegistry) open(filename string, fromFs afero.Fs) (*[]byte, error) {
+	fr.mutex.Lock()
+	defer fr.mutex.Unlock()
```
imiric, line 36: We might want to use a `RWMutex` here, and limit the scope of the lock to the absolute minimum. Since the most expensive operation here will be reading the file data, we don't need to hold the lock during that time. We also don't need to hold the write lock if we're just returning data for a file already in the registry, so `RLock()` can be used first, and `Lock()` only once we've read the data to protect writing to the `files` map.

js/modules/k6/experimental/fs/module.go
```diff
@@ -0,0 +1,156 @@
...
+	registry *fileRegistry
+}
+
+// Stat returns a promise that will resolve to a [FileInfo] instance describing
+// the file.
+func (f *File) Stat() *goja.Promise {
+	promise, resolve, _ := makeHandledPromise(f.vu)
+
+	go func() {
+		resolve(f.Stat())
```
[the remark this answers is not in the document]
oleiade: Great catch indeed 🙇🏻 
I'll address in my upcoming fixups + tests commit 👍🏻

js/modules/k6/experimental/fs/file.go
```diff
@@ -0,0 +1,29 @@
+package fs
+
+import (
+	"path"
+)
+
+// file is an abstraction for interacting with files.
+type file struct {
+	// name holds the name of the file, as presented to [Open].
+	Name string `json:"name"`
```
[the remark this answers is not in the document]
oleiade: I stuck with the convention followed in Go's os/fs modules. I find `Name` quite confusing too indeed, as it sometimes refers to the (inode name) path, and sometimes to the base name, depending on the context. I'll rename it to `path` then 👍🏻

js/modules/k6/experimental/fs/goja.go
```diff
@@ -0,0 +1,29 @@
...
+
+import (
+	"github.com/dop251/goja"
+	"go.k6.io/k6/js/modules"
+)
+
+// makeHandledPromise will create a promise and return its resolve and reject methods,
+// wrapped in such a way that it will block the eventloop from exiting before they are
+// called even if the promise isn't resolved by the time the current script ends executing.
+func makeHandledPromise(vu modules.VU) (*goja.Promise, func(interface{}), func(interface{})) {
```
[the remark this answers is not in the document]
oleiade: Indeed, I used a similar version of it in WebCrypto too. I'll ask the team where they would like this to live 👍🏻

js/modules/k6/experimental/fs/module.go
```diff
@@ -0,0 +1,156 @@
...
+		return promise
+	} else if isDir {
+		reject(fmt.Errorf("open() failed; reason: %q is a directory", pathStr))
+		return promise
+	}
+
+	go func() {
+		fs, ok := mi.vu.InitEnv().FileSystems["file"]
+		if !ok {
+			reject("open() failed; reason: no file configured")
```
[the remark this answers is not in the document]
oleiade: a) was a forgetful mistake on my part, fixing it 👍🏻 
b) I really like the idea to panic in that scenario, will do just that, thanks 🙇🏻

js/modules/k6/experimental/fs/module.go
oleiade: I don't expect users to use them directly indeed, but as you mentioned, it could be convinient to be able to `instanceof` etc on them. I'm open to removing though 👍🏻

js/modules/k6/experimental/fs/registry.go
oleiade: During the proof of concept, I had written some benchmarks to test that specific question "is it better to lock only a sub-part of the function, or the whole function", and the results at the time were leaning towards "there's little difference, and when there is it's actually slightly faster to lock the whole function.

I'll unbury them to try and confirm that my memory is not playing tricks on me though 👍🏻 

I'll switch to a `RWMutex` indeed 🙇🏻

js/modules/k6/experimental/fs/module.go
oleiade: I'll give it a shot 👍🏻

js/modules/k6/experimental/fs/registry.go
oleiade: I've run some quick benchmarks comparing both solutions.

Single, global, lock statement with defer at the top of the function:
```
Running tool: /usr/local/go/bin/go test -benchmem -run=^$ -bench ^BenchmarkFileRegistry_open$ go.k6.io/k6/js/modules/k6/experimental/fs

goos: darwin
goarch: arm64
pkg: go.k6.io/k6/js/modules/k6/experimental/fs
BenchmarkFileRegistry_open-10    	 6920181	       171.6 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	go.k6.io/k6/js/modules/k6/experimental/fs	1.533s 
```

Explicitly read-locking and write-locking only where necessary:
```
Running tool: /usr/local/go/bin/go test -benchmem -run=^$ -bench ^BenchmarkFileRegistry_open$ go.k6.io/k6/js/modules/k6/experimental/fs

goos: darwin
goarch: arm64
pkg: go.k6.io/k6/js/modules/k6/experimental/fs
BenchmarkFileRegistry_open-10    	 6844564	       170.9 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	go.k6.io/k6/js/modules/k6/experimental/fs	1.582s
```

As in my previous benchmarks, not a massive difference indeed (.5 ns/op).

The resulting code when explicitly locking and unlocking turns out more explicit though, so I'll go for that 👍🏻

imiric: I don't see how the benchmark is implemented, but this would only be noticeable in a test with many VUs, and when reading (maybe many) large files. So you're right that in most cases the difference would be negligible, but you can imagine that if one VU tries to load a 1GB file from a slow HDD, any other VU would need to wait until this is done to open other files. Even if the file was already opened, and we just want to return a `File` handle to it.

Since we have no reason to keep the write lock for this entire method, I think it would improve the performance in these edge cases, and it's a simple change.

It's not a big deal, of course, and we can leave it for later. Just something that popped up and thought it would be good to mention.

js/modules/k6/experimental/fs/module.go
imiric: Sure, no worries, we can keep them. Was just wondering if you had other plans for them.

js/modules/k6/experimental/fs/module.go
```diff
@@ -0,0 +1,156 @@
...
+var (
+	_ modules.Module   = &RootModule{}
+	_ modules.Instance = &ModuleInstance{}
+)
+
+// New returns a pointer to a new [RootModule] instance.
+func New() *RootModule {
+	return &RootModule{
+		fileRegistry: fileRegistry{
+			files: make(map[string]*[]byte),
```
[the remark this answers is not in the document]
oleiade: Indeed, good catch, I forgot to change it from the POC 🤝

js/modules/k6/experimental/fs/registry.go
oleiade: I didn't think of posting the benchmark code indeed. I did some very basic tests. I agree with you anyway, and the upcoming update commit will contain the change you suggested 👍🏻

js/modules/k6/experimental/fs/goja.go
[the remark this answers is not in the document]
oleiade: Will open a separate PR, and update the code accordingly once it's merged 🙇🏻

codebien (changes requested):
Hey @oleiade thanks for your contribution :wink: :sunglasses: 

Some comments for you. :handshake:

js/modules/k6/experimental/fs/module.go
```diff
@@ -0,0 +1,185 @@
...
+	"go.k6.io/k6/js/modules"
+	"go.k6.io/k6/js/promises"
+	"go.k6.io/k6/lib/fsext"
+)
+
+type (
+	// RootModule is the global module instance that will create instances of our
+	// module for each VU.
+	RootModule struct {
+		fileRegistry fileRegistry
```
mstoykov, line 20: I do wonder if this even makes sense from design/implementation perspective.

Technically we want to have as few copies of the file data as possible. This in practice adds 1, so uh not great.

All the methods that in practice will not copy the whole body (which will need to have a complete copy either way) are in practice also implemented on the current file system abstraction (afero.FS) - seek, read, stat, close.

io/fs is not guraunteed to implement `seek` it seems, but it likely will for all of *our* cases.

So I do wonder if we even need this to begin with instead of directly using the underlying fs abstraction and wrap them in js.

Am I missing a case?

js/modules/k6/experimental/fs/errors.go
```diff
@@ -0,0 +1,51 @@
...
+	// instance when attempting to open a directory, which is not supported.
+	InvalidResourceError ErrorKind = "InvalidResourceError"
+
+	// ForbiddenError is emitted when an operation is forbidden.
+	ForbiddenError ErrorKind = "ForbiddenError"
+
+	// TypeError is emitted when an incorrect type has been used.
+	TypeError ErrorKind = "TypeError"
+)
+
```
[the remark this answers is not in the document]
oleiade: Do you have an example of that in action, I don't see what you mean? 😄

js/modules/k6/experimental/fs/module.go
```diff
@@ -0,0 +1,185 @@
...
+	// will probably be need for archive execution under windows if always consider '/...' as an
+	// absolute path.
+	if path[0] != '/' && path[0] != '\\' && !filepath.IsAbs(path) {
+		path = filepath.Join(initEnv.CWD.Path, path)
+	}
+	path = filepath.Clean(path)
+
+	if path[0:1] != fsext.FilePathSeparator {
+		path = fsext.FilePathSeparator + path
+	}
```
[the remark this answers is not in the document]
oleiade: That's a good idea, I'll look into it 👍🏻

js/modules/k6/experimental/fs/registry.go
```diff
@@ -0,0 +1,66 @@
...
+}
+
+// open opens the named file for reading.
+//
+// If the file was already opened, it returns a pointer to the cached file data. Otherwise, it
+// opens the file, reads its content, caches it, and returns a pointer to it.
+//
+// The file is always opened in read-only mode. The provided file path is cleaned before being
+// used.
+func (fr *fileRegistry) open(filename string, fromFs afero.Fs) ([]byte, error) {
```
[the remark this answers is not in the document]
oleiade: That's because we instantiate a single file registry module-wise, and we do that at the `RootModule` level, where we don't have the VU available to us yet, hence no access to the init environment yet: https://github.com/grafana/k6/blob/d6b43b9f16530f73241c510354f25951ba7be38a/js/modules/k6/experimental/fs/module.go#L39

js/modules/k6/experimental/fs/errors.go
```diff
@@ -0,0 +1,51 @@
...
+	// Message contains the error message as presented to the user.
+	Message string `json:"message"`
+}
+
+// Ensure that the Error type implements the Go `error` interface.
+var _ error = (*Error)(nil)
+
+// Error implements the Go `error` interface.
+func (e *Error) Error() string {
+	return fmt.Sprintf(e.Name)
```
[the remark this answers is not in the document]
oleiade: Absolutely 🙇🏻

js/modules/k6/experimental/fs/module.go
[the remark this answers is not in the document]
oleiade: I didn't think it was necessary, as there will be a single instance of it, but I might as well make it a pointer 👍🏻

js/modules/k6/experimental/fs/module.go
```diff
@@ -0,0 +1,185 @@
...
+	return &ModuleInstance{vu: vu, fileRegistry: &rm.fileRegistry}
+}
+
+// Exports implements the modules.Module interface and returns the exports of
+// our module.
+func (mi *ModuleInstance) Exports() modules.Exports {
+	return modules.Exports{
+		Named: map[string]any{
+			"open":     mi.Open,
+			"File":     File{},
```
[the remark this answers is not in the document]
oleiade: So that users can `instanceof` over the result of `open` and `stat`, see https://github.com/grafana/k6/pull/3165#discussion_r1254231656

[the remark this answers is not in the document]
olegbespalov: I'm afraid that `instanceof` won't work here since, in that case, the export `File` is an object (go's `File{}`), not the class. Maybe I'm wrong here, but I believe it is worth checking.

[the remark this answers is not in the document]
oleiade: I looked into it, and you're right, it doesn't work as I expected. I dug a bit further into it, and making it work the way I would like: `fileInfo instanceof FileInfo` looks quite tricky at the moment; or at least I couldn't make sense of how to make it work 😞

js/modules/k6/experimental/fs/module.go
oleiade: I think I understand where you're coming from: why do we cache the data when we could access it through the fs (or afero) library directly? Correct?

The initial main reason was: because there are limited file descriptors available to the OS, and in the event, we have a number of VUs > number of available file descriptors, the OS is going to start refusing our requests to open files.

But now that you say it, as I write this, I also come to realize that we actually open a file descriptor once, store it in the registry, and then have various `File` instances with their own internal offsets as we do now, but instead, later call `ReadAt` on them for instance. I'm gonna give that a shot 👍🏻

js/modules/k6/experimental/fs/module.go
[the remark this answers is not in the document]
oleiade: I've added a [`fsext.Abs` helper ](https://github.com/grafana/k6/pull/3165/commits/6664f01805e5085a8fca28235d0c4f33f78ecbe7)👍🏻 

I only had some of the background knowledge as to precisely why we need it beyond what was in the initial comment, so if you folks have any suggestions to improve it, go ahead! 💯

js/modules/k6/experimental/fs/module.go
mstoykov: > The initial main reason was: because there are limited file descriptors available to the OS

That is *true*. In the last 10+ years there is usually no problems to have hundreds of thousands if not even millions of file descriptors. The actual problem AFAIK for a long time has been that `select` has a limit of 1024 so if you have a big limit globally and a program with `select` it can panic in a ... bad way.

But in practice k6 has to be run with more than 1024 fds as it also does network requests and it does quite a lot of those. We do not share those between VUs for some reason, and I don't think we should do it for files.

Most use cases and users are unlikely to be running so many VUs on a single machine and opening so many files that they will hit the limit. And if they *do* - I would argue raising the limit will work way better than us having a lock and basically synchronizing access to the fs. 

I would expect that even at 1k VUs trying to read 100b each sequentially it will practically spend most of it in locking and unlocking.

This is most based on

> but instead, later call ReadAt on them for instance

Where it seems you want to `ReadAt` which seems to be race safe from reading the code - although I would argue the documentation doesn't make it as explicit as I would like. 

So maybe we can use it ... maybe we should just directly "proxy" to `Read` and let the OS (in the future when afero.MemMapsfs isn't the actual backing) handle it.

oleiade: I don't understand what you're suggesting we should do then? Can you clarify that?

_some context I have_: 
* my understanding is that Go reuses underlying TCP connections (and thus file descriptors?) under the hood, under normal conditions (not forgetting to close the response body), thus limiting the number of file descriptors opened at the same time, even when firing tons of requests concurrently. But I guess you'll have more details on this.
* my (just now) empirical tests show that linux has a limit of 1024 open fd (on my ubuntu server) per process, macos seems to be somewhere around 10k, and windows is much much farther. 
* changing the limit on linux is somewhat easy, on mac it seems possible, on windows I couldn't find much conclusive. All in all, telling users "you can open files, but you're gonna have to fine tune your machine" is not really a good user experience.
* the time spent locking/unlocking is good point indeed
* I've made some tests with readAt up to 10k concurrent readers and no race detected

mstoykov: > my (just now) empirical tests show that linux has a limit of 1024 open fd (on my ubuntu server) per process, macos seems to be somewhere around 10k, and windows is much much farther.

> changing the limit on linux is somewhat easy, on mac it seems possible, on windows I couldn't find much conclusive. All in all, telling users "you can open files, but you're gonna have to fine tune your machine" is not really a good user experience.

Both of those are valid and will have the same problem if you do 2000 network requests and your limit is 1000. 

So making it so that opening 1000 files is any different than making 1000 network requests seems like not needed optimization.

> I've made some tests with readAt up to 10k concurrent readers and no race detected

As I mention above - it seems like it will be (at least on Posix) from my reading of the code as it uses a syscall that does ReadAt in one go instead of doing as Seek +  Read. 

In our current underlying fs(afero.fs) - this is even better. The whole file is in memory already so `ReadAt` is reading parts of a slice essentially. 

My concerns:
-  ~I am not certain ReadAt is actually supposed to be safe for concurrent use~ Reading the docs on [`io.ReaderAt`](https://pkg.go.dev/io#ReaderAt) it is defined it is supposed to be safe to call it in parallel.
-  `io/fs` (which we would like to move to) is not guaranteed to have it implemented and implementing for some case might be harder than expected.

So keeping with `Read` and multiple files seems like the better option. 

Also, from implementation perspective having a registry means that we have to keep count of it, lock it at least on `open`. I guess we might also might want to remove stuff from it on `close` if all VUs have closed it :shrug:.

Not having one just removes all of those questions and needed considerations.
 
tl;dr: IMO having a registry is not worth it in terms of what it adds as features, but requires us to make more complicated implementation. But as with most code (especially one in experimental modules) I doubt it will stay the same long term, so I don't think it is that big of a problem having it as well. I am just mostly not sold on the benefits from it.

js/modules/k6/experimental/fs/errors.go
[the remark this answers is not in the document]
oleiade: I applied the same strategy 👍🏻 Thanks for the pointers 🙇🏻

js/modules/k6/experimental/fs/module.go
oleiade: Thanks for your writeup 🙇🏻 

I've been giving some thoughts to it, and experimented with how I understood the alternative you propose. 

For this first round of implementation of the module, I'd rather stick to the current implementation, and would be very keen to revise it at a later point in time indeed (#3020 and #2978 for instance).

oleiade:
Too many fixup comments at this point. Making a round of squashing as I've addressed most feedback received so far 👍🏻

imiric (dismissed):
LGTM :+1: Great work, Théo! :clap:

js/modules/k6/experimental/fs/registry_test.go
```diff
@@ -0,0 +1,52 @@
...
+		_, gotFirstOk := registry.files.Load("bonjour.txt")
+		secondData, secondErr := registry.open("bonjour.txt", fs)
+		_, gotSecondOk := registry.files.Load("bonjour.txt")
+
+		assert.True(t, gotFirstOk)
+		assert.NoError(t, firstErr)
+		assert.Equal(t, []byte("Bonjour, le monde"), firstData)
+		assert.True(t, gotSecondOk)
+		assert.NoError(t, secondErr)
+		assert.Equal(t, firstData, secondData) // same pointer
```
codebien, line 49:
```suggestion
		assert.Same(t, firstData, secondData)
```
https://pkg.go.dev/github.com/stretchr/testify/assert#Same

codebien (changes requested):
LGTM :clap: A few minor comments.

js/modules/k6/experimental/fs/file.go
```diff
@@ -0,0 +1,32 @@
...
+// Stat returns a FileInfo describing the named file.
+func (f *file) Stat() *FileInfo {
+	filename := filepath.Base(f.path)
+	return &FileInfo{Name: filename, Size: len(f.data)}
+}
+
+// FileInfo holds information about a file.
+//
+// It is a wrapper around the [fileInfo] struct, which is meant to be directly
+// exposed to the JS runtime.
```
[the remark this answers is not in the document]
oleiade: Absolutely, I changed the way this is done. Removing that comment indeed 👍🏻

js/modules/k6/experimental/fs/registry_test.go
oleiade: Actually, this is interesting, for some reason, the test fails with `Same`:

```
            	Error:      	Not same:
            	            	expected: 0x1400009ea00 []byte{0x42, 0x6f, 0x6e, 0x6a, 0x6f, 0x75, 0x72, 0x2c, 0x20, 0x6c, 0x65, 0x20, 0x6d, 0x6f, 0x6e, 0x64, 0x65}
            	            	actual  : 0x1400009ea00 []byte{0x42, 0x6f, 0x6e, 0x6a, 0x6f, 0x75, 0x72, 0x2c, 0x20, 0x6c, 0x65, 0x20, 0x6d, 0x6f, 0x6e, 0x64, 0x65}
```

But I fail to see why 😓 they look to be identical to me: same address, same content?

oleiade: After re-reading https://go.dev/blog/slices-intro I have a better intuition as to why `assert.Same` doesn't effectively test the two byte slices are the same here.

That's because, they are 🥁 not.

When the file registry returns the byte slice corresponding to the file's content, it returns a _copy_ of its internal slice, which points to the same underlying array `*ptr`, and has the same `len` and `cap` properties. Only the "slice header" is copied. But the result is that in that test, although `firstData`, and `secondData`'s internal pointer, len, and capacity are equal, their "slice header" address is different.

Apparently, `bytes.Equal` can help with that. I also read that checking if both slices' first byte has the same address would work too (as in they both point to the same first byte in memory).

oleiade: I've replaced the check with `assert.True(t, bytes.Equal(firstData, secondData))` 👍🏻

js/modules/k6/experimental/fs/errors.go
```diff
@@ -0,0 +1,60 @@
...
+	// kind contains the kind of error that has occurred.
+	kind errorKind
+}
+
+// Ensure that the Error type implements the Go `error` interface.
+var _ error = (*fsError)(nil)
+
+// Error implements the Go `error` interface.
+func (e *fsError) Error() string {
+	return fmt.Sprintf(e.Name + ": " + e.Message)
```
olegbespalov, line 59: Seems like we could safely drop the `fmt` and just concatenate here

js/modules/k6/experimental/fs/file.go
```diff
@@ -0,0 +1,29 @@
+package fs
+
+import (
+	"path/filepath"
+)
+
+// file is an abstraction for interacting with files.
+type file struct {
+	// name holds the name of the file, as presented to [Open].
```
olegbespalov, line 9: Obsolete comment
```suggestion
```

js/modules/k6/experimental/fs/file.go
```diff
@@ -0,0 +1,29 @@
...
+	filename := filepath.Base(f.path)
+	return &FileInfo{Name: filename, Size: len(f.data)}
+}
+
+// FileInfo holds information about a file.
+type FileInfo struct {
+	// Name holds the base name of the file.
+	Name string `json:"name"`
+
+	// Name holds the length in bytes of the file.
```
olegbespalov, line 27:
```suggestion
	// Size holds the size of the file in bytes.
```

js/modules/k6/experimental/fs/module.go
```diff
@@ -0,0 +1,167 @@
...
+}
+
+func (mi *ModuleInstance) openImpl(path string) (*File, error) {
+	initEnv := mi.vu.InitEnv()
+
+	path = fsext.Abs(initEnv.CWD.Path, path)
+
+	fs, ok := initEnv.FileSystems["file"]
+	if !ok {
+		common.Throw(mi.vu.Runtime(), fmt.Errorf("open() failed; reason: unable to access the file system"))
```
olegbespalov, line 100:
```suggestion
		common.Throw(mi.vu.Runtime(), errrors.New("open() failed; reason: unable to access the file system"))
```

js/modules/k6/experimental/fs/registry.go
```diff
@@ -0,0 +1,60 @@
...
+	// That way, we can cache the file's content and avoid opening too many
+	// file descriptor, and re-reading its content every time the file is opened.
+	//
+	// Importantly, this also means that if the
+	// file is modified from outside of k6, the changes will not be reflected in the file's data.
+	// files map[string][]byte
+	files sync.Map
+}
+
+// open opens the named file for reading.
```
olegbespalov, line 28: Here I have a small confusion :thinking: 

Looking at the code I'd say that the current implementation of the `open`, does not exactly open the file, but reads and returns the file data with in-memory caching

js/modules/k6/experimental/fs/registry.go
```diff
@@ -0,0 +1,60 @@
...
+	// Keys are expected to be strings holding the files' path.
+	// Values are expected to be byte slices holding the files' data.
+	//
+	// That way, we can cache the file's content and avoid opening too many
+	// file descriptor, and re-reading its content every time the file is opened.
+	//
+	// Importantly, this also means that if the
+	// file is modified from outside of k6, the changes will not be reflected in the file's data.
+	// files map[string][]byte
+	files sync.Map
```
olegbespalov, line 25: What do you think about ching the naming of this to `cache` or `opennedFiles` ?

js/modules/k6/experimental/fs/errors.go
oleiade: Absolutely 👍🏻

js/modules/k6/experimental/fs/registry.go
oleiade: I like it, it's indeed more explicit 👍🏻 

I'll rename it to `openedFiles`, as I don't want to rely too much on the caching terminology as we might evolve the implementation in the future 🙇🏻

js/modules/k6/experimental/fs/registry.go
oleiade: You're right, I'll adjust the docs to make the behavior more explicit indeed 👍🏻

oleiade: I've updated the docs, let me know if you spot further improvements, and feel free to unresolved accordingly 🙇🏻

olegbespalov (approved):
Generally, I have no other comments about the PRs code :+1: For sure, we need here at least one more pair of eyes :eyes:  
 
One concern that I have in my mind is the caching. We do the cache in the new registry we the idea in mind that this keeps file data in memory only once, but if I remember correctly, there is another layer of caching in the afero's filesystem https://github.com/grafana/k6/blob/4ad625c1ce84822d71ed9819e2c372ec63fdca7b/lib/fsext/cacheonread.go#L17. So it could turn out that we double the memory usage :man_shrugging: So maybe it worth benchmark the memory and compare the legacy `open` with the new one's :thinking:  

Still, I admit that my concern could be because of my lacking context and missing the bigger picture (idea) around the new File API, but I feel it necessary to bring that at least.

js/modules/k6/experimental/fs/registry.go
```diff
@@ -0,0 +1,75 @@
...
+	if f, ok := fr.openedFiles.Load(filename); ok {
+		data, ok := f.([]byte)
+		if !ok {
+			panic(fmt.Errorf("registry's file %s is not stored as a byte slice", filename))
+		}
+
+		return data, nil
+	}
+
+	f, err := fromFs.Open(filename)
```
codebien, line 62:
```suggestion
        // TODO: explain our plans around afero and link the related issue
	f, err := fromFs.Open(filename)
```
The issue should be https://github.com/grafana/k6/issues/1079. Probably good to mention also the 2nd caching layer as @olegbespalov reported.

js/modules/k6/experimental/fs/registry_test.go
```diff
@@ -0,0 +1,53 @@
...
+		_, gotFirstOk := registry.openedFiles.Load("bonjour.txt")
+		secondData, secondErr := registry.open("bonjour.txt", fs)
+		_, gotSecondOk := registry.openedFiles.Load("bonjour.txt")
+
+		assert.True(t, gotFirstOk)
+		assert.NoError(t, firstErr)
+		assert.Equal(t, []byte("Bonjour, le monde"), firstData)
+		assert.True(t, gotSecondOk)
+		assert.NoError(t, secondErr)
+		assert.True(t, bytes.Equal(firstData, secondData))
```
codebien, line 50: This is not enough and it isn't protecting us from doing one more copy. As you can check from: https://go.dev/play/p/NkHxud8QsNn

js/modules/k6/experimental/fs/errors.go
```diff
@@ -0,0 +1,58 @@
...
+
+	// ForbiddenError is emitted when an operation is forbidden.
+	ForbiddenError
+
+	// TypeError is emitted when an incorrect type has been used.
+	TypeError
+)
+
+// fsError represents a custom error object emitted by the fs module.
+type fsError struct {
```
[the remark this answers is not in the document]
oleiade: The `fsError` type exists in order to return `Error`-like Javascript objects on failure with a predefined structure (in this case, an error name and a message). That also makes testing easier as I know exactly what an error would look like when I catch one.

I did this in WebCrypto to have clear error types with a predefined structure too. I had looked into alternatives to this way of doing and hadn't found a satisfying one, but if you have an alternative recommendation, I'm happy to give it a shot 👍🏻 

Deno, which this API is inspired from, also has a code property on the error objects returned by their File API, which might be something to add too.

I'll add tests asserting the errors are of the right type with `errors.As` indeed 🙇🏻

js/modules/k6/experimental/fs/module.go
```diff
@@ -0,0 +1,168 @@
...
+	promise, resolve, reject := promises.New(mi.vu)
+
+	// Files can only be opened in the init context.
+	if mi.vu.State() != nil {
+		reject(newFsError(ForbiddenError, "open() failed; reason: opening a file in the VU context is forbidden"))
+		return promise
+	}
+
+	if common.IsNullish(path) {
+		reject(newFsError(TypeError, "open() failed; reason: path cannot be null or undefined"))
```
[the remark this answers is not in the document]
oleiade: I believe this is also covered by https://github.com/grafana/k6/pull/3165#discussion_r1300205638, let me know if not though 😃

js/modules/k6/experimental/fs/registry_test.go
codebien:
```
func SamePoint(x, y []int) bool {
	return cap(x) > 0 && cap(y) > 0 && &x[0:cap(x)][cap(x)-1] == &y[0:cap(y)][cap(y)-1]
}
```

js/modules/k6/experimental/fs/module.go
```diff
@@ -0,0 +1,168 @@
...
+
+	path = fsext.Abs(initEnv.CWD.Path, path)
+
+	fs, ok := initEnv.FileSystems["file"]
+	if !ok {
+		common.Throw(mi.vu.Runtime(), errors.New("open() failed; reason: unable to access the file system"))
+	}
+
+	if exists, err := fsext.Exists(fs, path); err != nil {
+		return nil, fmt.Errorf("open() failed, unable to verify if %q exists; reason: %w", path, err)
```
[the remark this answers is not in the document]
oleiade: The reasoning was that this was an "internal" error, as opposed to an error likely caused by the user's use of the module. 

In the line you referenced, we couldn't verify the file exists because some code internal to k6 failed to perform the operation. Whereas in the other branch, we were successful in checking the existence of the file, in which case we indicate to the user that the fs couldn't find the file.

Let me know if you have another solution/approach to this in mind 👍🏻

js/modules/k6/experimental/fs/module.go
```diff
@@ -0,0 +1,168 @@
...
+
+	return promise
+}
+
+func (mi *ModuleInstance) openImpl(path string) (*File, error) {
+	initEnv := mi.vu.InitEnv()
+
+	path = fsext.Abs(initEnv.CWD.Path, path)
+
+	fs, ok := initEnv.FileSystems["file"]
```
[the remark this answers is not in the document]
oleiade: I had the same reaction as you while using it. "Okay, so this is the "file" filesystem, but are the other available ones, and how can I find them out?

I would not change it in this PR, but I agree we could open an issue about it. I don't really have a better option in mind at the moment, but it's worth looking into 👍🏻 

I'll create one, and please feel free to help improving it 🙇🏻

js/modules/k6/experimental/fs/registry_test.go
oleiade: You're right, as per [my previous comment](https://github.com/grafana/k6/pull/3165#discussion_r1283150095) the only way to really check is to verify that each []byte array's first byte points to the same address, but I found it a bit hacky and wanted to find a better alternative, but didn't get to it.

Here's an example of what I mean: https://go.dev/play/p/4V5gOGZIxek

js/modules/k6/experimental/fs/registry.go
```diff
@@ -0,0 +1,75 @@
...
+	"fmt"
+	"io"
+	"path/filepath"
+	"sync"
+
+	"github.com/spf13/afero"
+)
+
+// registry is a registry of opened files.
+type registry struct {
```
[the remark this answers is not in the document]
oleiade: I'm not opposed to the `Cache` name indeed. I'll look into https://github.com/grafana/k6/pull/3165#issuecomment-1686420904, and if I end up not getting rid of it, I'll rename it accordingly 👍🏻

js/modules/k6/experimental/fs/registry.go
oleiade: I'll add a mention of it indeed 👍🏻

js/modules/k6/experimental/fs/errors.go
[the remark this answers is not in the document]
codebien: >The fsError type exists in order to return Error-like Javascript objects on failure with a predefined structure

Are you aware if there is some sort of Deno-like or modern-JS defined somewhere about it that I could read?

>That also makes testing easier as I know exactly what an error would look like when I catch one.

I see you are doing
```
if (err.name !== 'TypeError')
```
Is this idiomatic JavaScript? I see [Deno is using `instanceof`](https://deno.land/api@v1.36.1?s=Deno.errors) for doing something similar. Maybe we should do the same.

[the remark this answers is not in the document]
codebien: >I did this in WebCrypto to have clear error types with a predefined structure too.

Reading and recalling it, it could be the right moment for creating our opinionated struct and doing something generic in `js` package and reusing it across the modules instead of redefining it.

js/modules/k6/experimental/fs/module.go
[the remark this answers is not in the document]
codebien: >I'll create one, and please feel free to help improving it 🙇🏻

:bow: I will comment there

js/modules/k6/experimental/fs/registry.go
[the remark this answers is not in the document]
codebien: >and if I end up not getting rid of it

If you mean dropping the added Registry concept then I think we should not do it. It would complicate the migration from `afero`. The added Registry is an abstraction from `afero` that should simplify the work for dropping it.
We said `afero` debt wouldn't be part of this work so I guess having one more copy is something we can have until we decide to drop completely `afero`. Except, if we can not use the afero layer only for this specific module, but it sounds unrealistic to me.

I hope I got right your sentence, otherwise, feel free to correct me.

js/modules/k6/experimental/fs/module.go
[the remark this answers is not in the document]
codebien: Let's discuss there directly. Resolving.

js/modules/k6/experimental/fs/errors.go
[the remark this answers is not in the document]
oleiade: > Is this idiomatic JavaScript? I see [Deno uses instanceof](https://deno.land/api@v1.36.1?s=Deno.errors) for doing something similar. Maybe we should do the same.

My understanding is that the alternative is tricky in Goja (at least from the last time I attempted it). Supporting `isinstanceof` implies setting the prototype of an Object to match the right-hand side of the expression. I looked into it in the context of a previous pull request. I remember we discussed it with @olegbespalov. Had looked into it at the time but couldn't find the way to do that in Goja during the time I allocated to it (it's probably doable, but I struggled to find a way to do it with Goja lacking examples on that front).

I'll give it another shot 🤞🏻 

> creating our opinionated struct, doing something generic in the js package, and reusing it across the modules instead of redefining it.

I've been considering doing it for a while now, but I don't have a clear path forward yet. I'll open an issue 👍🏻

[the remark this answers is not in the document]
oleiade: I gave it another shot, and in the two hours I spent on it, I came somewhat close, but I couldn't get `e instanceof FSError` to work 😞 

The main reason is that `instanceof` expects its right-hand side to be a function, and at the moment, Goja does not expose the ability to instantiate new native functions. If you check Goja's code, with the implementation of the `TypeError`, for instance, which is natively supported by Goja, you'll see that it depends on a lot of APIs that are not publicly exposed.

I have tried different approaches to try to work around that limitation, but I couldn't get any to work in the end:
- one approach could be to replicate the code that Goja uses internally in k6 (through my quick look, I didn't spot any blockers on that path), but it isn't necessarily a good idea.
- another approach would be to make a JS expression directly executed in the runtime.

I find both approaches unsatisfactory, and they would require some work that's out of the scope of this PR. 

I will open an issue to track this problematic, and keep the current implementation as-is.

Let me know if that works for you @codebien 🙇🏻

js/modules/k6/experimental/fs/module.go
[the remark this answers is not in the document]
oleiade: I have created [#3296](https://github.com/grafana/k6/issues/3296)

js/modules/k6/experimental/fs/errors.go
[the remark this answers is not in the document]
oleiade: I have opened an [issue with goja](https://github.com/dop251/goja/issues/529) 👍🏻 

Until there is a sensible solution to this, I will keep the implementation as is.

Resolving this conversation, feel free to re-open it if something changes 🙇🏻

js/modules/k6/experimental/fs/registry.go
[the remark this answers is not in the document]
oleiade: My reply was too short and needed to be more specific about the context.

I'm unkeen to remove the concept of a file "registry" (whatever we name it) because we have somewhat of an agreement that we want to reduce our dependency on Afero. 

Even though we currently [have the file content cached twice instead of once](https://github.com/grafana/k6/pull/3165#issuecomment-1686420904), once by ourselves, and a second time by the underlying afero fs `CacheOnRead`, it's still better than VU times, and it affords us to limit our dependency on afero itself, and keeps the path towards reducing our use of it easier.

I would rename `registry` as `Cache` for now 👍🏻

js/modules/k6/experimental/fs/registry_test.go
oleiade: Interesting, I was wondering why the code you posted would work, rather than checking the first byte's pointer. Found the explanation in a Google Group:

```
The two slices overlap if they both end at the same address. Once a
slice is created, you can move its starting address forward, but never
beyond start + capacity, which is a fixed value for any slice. That
syntax reslices x and y to their full capacity and takes the address
of the last element.
```

oleiade:
Throughout discussions with @codebien, we've established that because we have the intention to limit our usage of afero as much as possible (as per #1079), we would rather keep the current caching mechanism in place, even though the file would end up being cached twice. It avoids us introducing one more explicit dependency on afero (to `afero.File`, for instance), and we assume this cost would go away with afero, and it doesn't spoil the benefits of the current approach either.

js/modules/k6/experimental/fs/cache.go
```diff
@@ -0,0 +1,80 @@
...
+		}
+
+		return data, nil
+	}
+
+	// The underlying afero.Fs.Open method will cache the file content during this
+	// operation. Which will lead to effectively holding the content of the file in memory twice.
+	// However, as per #1079, we plan to eventually reduce our dependency on afero, and
+	// expect this issue to be resolved at that point.
+	// TODO @oleiade: re-evaluate this once #1079 is resolved.
```
codebien, line 66:
```suggestion
	// TODO: re-evaluate this once #1079 is resolved.
```
I guess is not an individual responsibility. It is for the entire team.

js/modules/k6/experimental/fs/errors.go
codebien, line 40: What do you think to add here a TODO linking the goja issue?
```suggestion
// TODO:
type fsError struct {
```

js/modules/k6/experimental/fs/module.go
```diff
@@ -0,0 +1,168 @@
...
+	// RootModule is the global module instance that will create instances of our
+	// module for each VU.
+	RootModule struct {
+		registry *cache
+	}
+
+	// ModuleInstance represents an instance of the fs module for a single VU.
+	ModuleInstance struct {
+		vu modules.VU
+
```
codebien, line 26:
```suggestion
```
I think we can drop it

codebien (commented):
LGTM, I don't have any important logic requests to add, mostly some polish.

mstoykov:
I don't think there are any plans to *not* have a caching mem map fs in place what currently afero provides. So even when afero is removed it - there is still going to be a cachign layer in exactly the same place.

SO if that is  the only reason the current caching is staying - then it is not a valid reason. Also you should only interact with `fsext` after #3025, so you shouldn't even think of afero. 

On the memory results- can you provide the scripts?

js/modules/k6/experimental/fs/cache.go
oleiade: This is a practice I've picked up from the Prometheus project. The `@` is not communicating who should do it, but who wrote it, so you can directly go to them to ask. It saves the need to git blame.

I don't mind removing it, but I wanted to make the meaning explicit 👍🏻

oleiade:
Thanks for the insights @mstoykov 👍🏻 

I need to experiment with the branches implementing further features such as `read` and `readAll`, but if it's conclusive, I'll attempt to drop the caching all in all then 🙇🏻 

For the testing, I created a file with 10MB of random binary data:
```bash
dd if=/dev/urandom of=random-10mb.bin bs=102400 count=100
```

Current `master` implementation: 
```bash
const file = open("random-10mb.bin", "b");

export const options = {
	vus: 100,
	duration: "3m",
};

export default function () {
	file.byteLength + 1;
}

```

Current PR implementation:
```bash
import { open } from "k6/experimental/fs";

export const options = {
	vus: 100,
	duration: "3m",
};

let file;
(async function () {
	file = await open("random-10mb.bin");
})();

	const stat = await file.stat();
	stat.size + 1;
}
```

js/modules/k6/experimental/fs/errors.go
oleiade: Sure 👍🏻

mstoykov:
Okay so the second example basically doesn't get the data to the user/script.

I would argue for better comparision they should do the same thing and the seconds `fs.open` should also use `ReadAll`.

If the idea is to show that `fs.open` doesn't read the data for each VU - great. But as explained in all the issues about working with files, the problem isn't *opening* the file in a nice way and resource efficient way. It is actually using it.

So while I am not against the comparision I would argue the current 

> The impact is quite neglectable though (210MB vs 10010MB), but still, it's worth looking into it indeed 👀

Is comparing not being able to use the file with actually using it.

And yes somewhere in the future a user can use [streams](https://github.com/grafana/k6/issues/2978) where the script won't see the whole body in one place. Which is why I prefered if we started with streams as that is the thing that will make it possible to actually use the data.

js/modules/k6/experimental/fs/module.go
```diff
@@ -0,0 +1,167 @@
...
+		resolve(file)
+	}()
+
+	return promise
+}
+
+func (mi *ModuleInstance) openImpl(path string) (*File, error) {
+	initEnv := mi.vu.InitEnv()
+
+	path = fsext.Abs(initEnv.CWD.Path, path)
```
mstoykov, line 96: This currently hits https://github.com/grafana/k6/issues/2781 which makes all the files open relative to the main module. 

But thinking about this it is probably better that we have *this* behaviour then the current `open` one explained in https://github.com/grafana/k6/issues/2674 - where `open` is relative to the currently initialzing module/script.

I don't think there is anything in particular you should do apart from adding tests for that and potentially documenting it at this point.

oleiade:
Squashing commits to make rebasing on this branch easier 🤞🏻

js/modules/k6/experimental/fs/module.go
oleiade: I align with you on this 🤝 

As a result, I have documented this point on this line of code specifically and added a dedicated test to illustrate it 🙇🏻

oleiade:
Ping reviewers 👀 @olegbespalov @mstoykov @codebien

examples/experimental/fs/fs.js
```diff
@@ -0,0 +1,52 @@
...
+	while (true) {
+		// Read into the buffer
+		const bytesRead = await file.read(buffer);
+		if (bytesRead == null) {
+			// EOF
+			break;
+		}
+
+		// Do something useful with the content of the buffer
+
```
codebien, line 36:
```suggestion
```

js/modules/k6/experimental/fs/module.go
```diff
@@ -0,0 +1,327 @@
...
+
+// Open opens a file and returns a promise that will resolve to a [File] instance
+func (mi *ModuleInstance) Open(path goja.Value) *goja.Promise {
+	promise, resolve, reject := promises.New(mi.vu)
+
+	// Files can only be opened in the init context.
+	if mi.vu.State() != nil {
+		reject(newFsError(ForbiddenError, "open() failed; reason: opening a file in the VU context is forbidden"))
+		return promise
+	}
```
codebien, line 71:
```suggestion
	if mi.vu.State() != nil {
		reject(newFsError(ForbiddenError, "open() failed; reason: opening a file is allowed only in the Init context"))
		return promise
	}
```
It sounds the error message is not clear enough if we need to add a dedicated comment for it. So I would like to merge it in the error message so it is clear for the code-redear but also for any eventual user that may hit it.
In any case, I expect if the user hits this probably they aren't expert user so they don't know what a Init or VU context is, so if we can find an alternative without mentioning them then it should be better. Unfortunately, I don't have a concrete proposal.

js/modules/k6/experimental/fs/module.go
```diff
@@ -0,0 +1,327 @@
...
+// Resolves to either the number of bytes read during the operation
+// or EOF (null) if there was nothing more to read.
+//
+// It is possible for a read to successfully return with 0 bytes.
+// This does not indicate EOF.
+func (f *File) Read(into goja.Value) *goja.Promise {
+	promise, resolve, reject := f.vu.Runtime().NewPromise()
+
+	if common.IsNullish(into) {
+		reject(newFsError(TypeError, "read() failed; reason: into cannot be null or undefined"))
```
codebien, line 193:
```suggestion
		reject(newFsError(TypeError, "read() failed; reason: into argument cannot be null or undefined"))
```

codebien (commented):
I expect we don't plan to squash the commit history. Instead, if we want to squash then I would prefer if we split the Abs method refactor in a dedicated PR.

js/modules/k6/experimental/fs/cache.go
```diff
@@ -0,0 +1,86 @@
+package fs
+
+import (
+	"fmt"
+	"io"
+	"path/filepath"
+	"sync"
+
+	"github.com/spf13/afero"
```
olegbespalov, line 9: Considering that afero [is planned to be replaced](https://github.com/grafana/k6/issues/1079) would you mind switching to the `fsext`? (https://github.com/grafana/k6/blob/master/lib/fsext/afero_links.go)

js/modules/k6/experimental/fs/cache.go
```diff
@@ -0,0 +1,86 @@
...
+	defer func() {
+		cerr := f.Close()
+		if cerr != nil {
+			err = fmt.Errorf("failed to close file %s: %w", filename, cerr)
+		}
+	}()
+
+	data, err = io.ReadAll(f)
+	if err != nil {
+		return nil, err
```
olegbespalov, line 80: Let's wrap this error to make it more user-friendly?

js/modules/k6/experimental/fs/errors.go
```diff
@@ -0,0 +1,71 @@
+package fs
+
+// newFsError creates a new Error object of the provided kind and with the
+// provided message.
+func newFsError(k errorKind, message string) *fsError {
+	return &fsError{
+		Name:    k.String(),
```
olegbespalov, line 7: Out of curiosity, why do we need two properties with the same data? If the name is calculated from the kind, why can't we obtain this dynamically?

js/modules/k6/experimental/fs/module.go
```diff
@@ -0,0 +1,326 @@
...
+	promise, resolve, reject := f.vu.Runtime().NewPromise()
+
+	if common.IsNullish(into) {
+		reject(newFsError(TypeError, "read() failed; reason: into argument cannot be null or undefined"))
+		return promise
+	}
+
+	// We expect the into argument to be a `Uint8Array` instance
+
+	// intoObj := into.ToObject(f.vu.Runtime())
```
olegbespalov, line 198: Are these commented lines still relevant?

js/modules/k6/experimental/fs/module.go
```diff
@@ -0,0 +1,326 @@
...
+// This does not indicate EOF.
+func (f *File) Read(into goja.Value) *goja.Promise {
+	promise, resolve, reject := f.vu.Runtime().NewPromise()
+
+	if common.IsNullish(into) {
+		reject(newFsError(TypeError, "read() failed; reason: into argument cannot be null or undefined"))
+		return promise
+	}
+
+	// We expect the into argument to be a `Uint8Array` instance
```
olegbespalov, line 196: This requirement is already a part of the method description. Do we still need to have this comment here?

js/modules/k6/experimental/fs/module.go
```diff
@@ -0,0 +1,326 @@
...
+			// bytes read.
+			if readErr == nil {
+				resolve(n)
+				return nil
+			}
+
+			// The [file.Read] method will return an EOFError as soon as it reached
+			// the end of the file.
+			//
+			// However, following deno's behavior, we express
```
olegbespalov, line 239: I have a hard time getting what is `following deno's behavior`. Is this is a relevant details here?

js/modules/k6/experimental/fs/module.go
oleiade: Good catch 👀 I forgot that behind from the previous PR it seems 👍🏻

lib/fsext/filepath_unix_test.go
```diff
@@ -72,3 +72,59 @@ func TestJoinFilePath(t *testing.T) {
...
+		{
+			name: "relative path with leading double dot",
+			args: args{
+				root: "/",
+				path: "../test",
+			},
+			want: "/test",
+		},
+	}
+	for _, tt := range tests {
```
olegbespalov, line 121: Do we need to test windows-like cases here?

js/modules/k6/experimental/fs/errors.go
oleiade: The reason for having an explicit `Name` field here is that it will be exposed to the Javascript runtime directly as the `.name` property of errors returned by `fs` operations. This is a convenience to enforce a certain "form" to errors exposed to users I replicated from the `webcrypto` project.

Whereas we use the `kind` attribute on the Go side.

olegbespalov: Okay, I see :+1: 

Just sharing that I might think that it's generally possible to achieve without the duplication by using https://pkg.go.dev/github.com/dop251/goja#Object.DefineAccessorProperty

But I guess in that case, it requires many changes for generating, and effort won't cost the result :+1:

js/modules/k6/experimental/fs/module_test.go
```diff
@@ -0,0 +1,695 @@
...
+
+			fileContent.set(buffer, 0);
+
+			bytesRead = await file.read(buffer)
+			if (bytesRead !== 2) {
+				throw 'expected read to return 2, got ' + bytesRead + ' instead';
+			}
+
+			// We expect the buffer to hold the two last bytes of the
+			// file, namely '34', and as we read only two bytes, its last
```
mstoykov, line 270: nit: "namely '34'" and then below there is no '34'

Not certain if this used to say '34' or not. Maybe we can leave only the part about the last one not being changed :shrug:

js/modules/k6/experimental/fs/module_test.go
```diff
@@ -0,0 +1,695 @@
+package fs
+
+import (
+	"fmt"
+	"net/url"
+	"path/filepath"
+	"testing"
+
+	"github.com/spf13/afero"
```
mstoykov, line 9: It will be preferable if `afero` is not directly used in tests as well. 

If there is a need more things can be added to fsext - but given that it is already possible for other modules, I expect you only need to use more of hte package.

See https://github.com/grafana/k6/pull/3025 for usage and changes across the codebase.

js/modules/k6/experimental/fs/module.go
```diff
@@ -0,0 +1,315 @@
...
+	// This is done on purpose, although it diverges in some respect with
+	// how files are handled in different k6 contexts, so that we cater to
+	// and intuitive user experience.
+	//
+	// See #2781 and #2674.
+	path = fsext.Abs(initEnv.CWD.Path, path)
+
+	fs, ok := initEnv.FileSystems["file"]
+	if !ok {
+		common.Throw(mi.vu.Runtime(), errors.New("open() failed; reason: unable to access the file system"))
```
mstoykov, line 112: Given that this is of the event loop using the runtime will lead to races.

Also, this function already returns an error so this likely should just use that.

```suggestion
		return nil,  errors.New("open() failed; reason: unable to access the file system")
```

js/modules/k6/experimental/fs/file.go
```diff
@@ -0,0 +1,142 @@
...
+	// TODO: using an atomic here does not guarantee ordering of reads and seeks, and leaves
+	// the behavior not strictly defined. This is something we might want to address in the future, and
+	// is tracked as part of #3433.
+	offset atomic.Int64
+}
+
+// Stat returns a FileInfo describing the named file.
+func (f *file) stat() *FileInfo {
+	filename := filepath.Base(f.path)
+	return &FileInfo{Name: filename, Size: len(f.data)}
```
mstoykov, line 29: nit: this is already impleemented: 
```suggestion
	return &FileInfo{Name: filename, Size: f.size()}
```

js/modules/k6/experimental/fs/cache.go
```diff
@@ -0,0 +1,86 @@
...
+		if !ok {
+			panic(fmt.Errorf("registry's file %s is not stored as a byte slice", filename))
+		}
+
+		return data, nil
+	}
+
+	// The underlying fsext.Fs.Open method will cache the file content during this
+	// operation. Which will lead to effectively holding the content of the file in memory twice.
+	// However, as per #1079, we plan to eventually reduce our dependency on fsext, and
```
mstoykov, line 64:
```suggestion
	// However, as per #1079, we plan to eventually reduce our dependency on afero, and
```

js/modules/k6/experimental/fs/module_test.go
oleiade: Fair point 👍

js/modules/k6/experimental/fs/module.go
```diff
@@ -0,0 +1,315 @@
...
+// the provided `whence`.
+func (f *File) Seek(offset goja.Value, whence goja.Value) *goja.Promise {
+	promise, resolve, reject := f.vu.Runtime().NewPromise()
+
+	if common.IsNullish(offset) {
+		reject(newFsError(TypeError, "seek() failed; reason: the offset argument cannot be null or undefined"))
+		return promise
+	}
+
+	var intOffset int64
```
olegbespalov, line 265: Nit: the validation has a repetitive pattern that could be converted into a small abstraction, improving the code readability by reducing the method size

Like:

```golang

intOffset, err := extractInt(offset, rt)
if err != nil {
   reject(newFsError(TypeError, "seek() failed; reason: the offset argument " + err.Error()))
}

intWhence, err := extractInt(whence, rt)
if err != nil {
   reject(newFsError(TypeError, "seek() failed; reason: the whence argument " + err.Error()))
}

func extractInt(v goja.Value, rt *goja.Runtime) (int64, error) {
   if common.IsNullish(v) {
       return 0, errors.New("is null or undefined")
   }
   
   var intV int64
   if err := f.vu.Runtime().ExportTo(offset, &intV); err != nil {
      return 0, errors.New("cannot be interpreted as an integer")
   }

   return intV, nil
}
```

olegbespalov (commented):
Generally LGTM :+1: :muscle: 

I left a minor nit comment. For approval, I'm waiting on splitting the things as I understood for the several PRs based on @codebien proposal https://github.com/grafana/k6/pull/3165#pullrequestreview-1717211290

js/modules/k6/experimental/fs/cache.go
```diff
@@ -0,0 +1,86 @@
...
+	// file is modified from outside of k6, the changes will not be reflected in the file's data.
+	openedFiles sync.Map
+}
+
+// open retrieves the content of a given file from the specified filesystem (fromFs) and
+// stores it in the registry's internal `openedFiles` map.
+//
+// The function cleans the provided filename using filepath.Clean before using it.
+//
+// If the file was previously "opened" (and thus cached) by the registry, it
```
codebien, line 33:
```suggestion
// If the file was previously "opened" (and thus cached), it
```

js/modules/k6/experimental/fs/cache.go
```diff
@@ -0,0 +1,86 @@
...
+		}
+
+		return data, nil
+	}
+
+	// The underlying fsext.Fs.Open method will cache the file content during this
+	// operation. Which will lead to effectively holding the content of the file in memory twice.
+	// However, as per #1079, we plan to eventually reduce our dependency on afero, and
+	// expect this issue to be resolved at that point.
+	// TODO: re-evaluate opening from the FS this once #1079 is resolved.
```
codebien, line 66: I think this comment is missing an explanation about _why_ we did the same thing if it was already possible in another way. And _why_ we ended up doing it twice.

Related to it, generally, I think now that we have implemented it, and in some details, it diverges from the design doc. For this reason, I would like to have somewhere in this code a comment explaining the goals of this module.
We have them here https://github.com/grafana/k6/blob/master/docs/design/019-file-api.md#proposed-solution but I think it would make sense to have them repeated in a comment in the code. Because the design doc is not referenced here, it could be difficult to discover and it could become outdated in the future. I think the code's comments are easier to maintain.

oleiade:
I have created #3454 to factor `fsext.Abs` out of this PR. I will remove the relevant commit and rebase on `master` once #3454 is merged 🤞🏻

js/modules/k6/experimental/fs/cache.go
oleiade: I've added a more explicit documentation to the `cache` type, explaining why it exists, referring to how it differs some of the constraints expressed in the design document, and pointing to #1079 as opening future opportunities.

The type comment superseeds the previous one made at the site of the `Open` call which I removed.

I'll mark this as resolved, but feel free to comment further on the newer version introduced by my latest commit 👍🏻

js/modules/k6/experimental/fs/module.go
oleiade: 👍🏻 

I've applied the pattern you suggested in https://github.com/grafana/k6/commit/ac8a5a7deb2f9e70b1fbc00a6dc3f679bdbd9494

js/modules/k6/experimental/fs/module.go
```diff
@@ -0,0 +1,321 @@
...
+func isUint8Array(rt *goja.Runtime, o *goja.Object) bool {
+	uint8ArrayConstructor := rt.Get("Uint8Array")
+	if isUint8Array := o.Get("constructor").SameAs(uint8ArrayConstructor); !isUint8Array {
+		return false
+	}
+
+	return true
+}
+
+//nolint:unparam
```
codebien, line 309: I'm not getting why there aren't failing tests. :thinking:

The passed target `i` is not a pointer so the linter seems to be right here, you will end with always return `0`. Am I missing something?

codebien (changes requested):

js/modules/k6/experimental/fs/module.go
oleiade: I have brought back the tests in https://github.com/grafana/k6/pull/3165/commits/d48fb673f8fa0bb4e4f8f36e8f1e2ae96373ab62

However I'm now back to my initial issue:
```
var i int64
if err := rt.ExportTo(v, &i); err != nil {
	return 0, errors.New("cannot be interpreted as integer")
}
```

Not returning an error when the function is called with 'abc' as a value, despite the goja documentation mentioning "Returns error if conversion is not possible.". Either a bug, or a surprising behavior (for me). Will look into it 👍🏻

oleiade: I've published a fix. I'll look into more details into the Goja behavior I've found surprising during cooldown 👍🏻

codebien (approved):
:tada: :clap:

olegbespalov (approved):
:+1:

oleiade:
🎉 I'll go ahead and squash and merge instead of making a merge commit as I realized that some of the changes applied as part of the PR review apply to content that was part of children's PRs, and that would make the whole branch cleanup quite nasty (if not impossible).

PR #3167, opened 2023-07-05
Use crypto.randomInt to make the assignment more random

mstoykov:
I have been personally assigned a lot more then any other contributor this is likely due to the usage of "random" sort which seems to be much more likely to not move elements around. Leading to the first few elements staying in their places. And my username is first :(.

This now uses node's crypto library to get "better" random.

My testing showcases that I at least get picked less :)

### DIFF
--- a/.github/workflows/pr-auto-assign.yml
+++ b/.github/workflows/pr-auto-assign.yml
@@ -1,4 +1,4 @@
-name: 'Auto assign PR reviewers'
+name: "Auto assign PR reviewers"
 on:
   pull_request_target:
     types: [opened, ready_for_review, reopened]
@@ -16,11 +16,21 @@ jobs:
           script: |
             const reviewers = ['mstoykov', 'imiric', 'codebien', 'olegbespalov', 'oleiade'];
             const reviewerCount = 2;
+            const crypto = require("node:crypto");
 
-            const reviewersWithoutAuthor = reviewers.filter(user => user !== context.payload.pull_request.user.login);
+            const getNRandom = (n, array) => {
+              let result = new Array();
+              for (;n > 0 && array.length > 0; n--) {
+                const chosen = array[crypto.randomInt(array.length)];
+                result.push(chosen);
+                array = array.filter(el => el != chosen);
+              }
+              return result;
+            }
+
             github.rest.pulls.requestReviewers({
               owner: context.repo.owner,
               repo: context.repo.repo,
               pull_number: context.payload.pull_request.number,
-              reviewers: reviewersWithoutAuthor.sort(() => 0.5 - Math.random()).slice(0, reviewerCount),
+              reviewers: getNRandom(reviewerCount, reviewers),
             });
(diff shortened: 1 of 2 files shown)


mstoykov:
You can see my testing here.

https://github.com/mstoykov/github-action-random-test/actions

olegbespalov (approved):

.github/workflows/pr-auto-assign.yml
```diff
@@ -16,11 +16,21 @@ jobs:
...
+              return result;
+            }
 
-            const reviewersWithoutAuthor = reviewers.filter(user => user !== context.payload.pull_request.user.login);
             github.rest.pulls.requestReviewers({
               owner: context.repo.owner,
               repo: context.repo.repo,
               pull_number: context.payload.pull_request.number,
-              reviewers: reviewersWithoutAuthor.sort(() => 0.5 - Math.random()).slice(0, reviewerCount),
+              reviewers: getNRandom(reviewerCount, reviewers),
```
imiric, line 35: Interesting. I wasn't aware that using `Array.sort()` for random shuffling would be biased towards certain elements, depending on the underlying implementation.

I ran some tests and confirmed it:

```javascript
// auto-assign-test.js
const assignees = ['mstoykov', 'imiric', 'codebien', 'olegbespalov', 'oleiade'];
const assigneeCount = 1;
console.log(assignees.sort(() => 0.5 - Math.random()).slice(0, assigneeCount)[0]);
```
```shell
$ for i in {0..1000}; do node auto-assign-test.js; done > auto-assign-test-results.txt
$ sort auto-assign-test-results.txt | uniq -c | sort
    120 imiric
    155 codebien
    165 oleiade
    229 olegbespalov
    332 mstoykov
```

:smile:

This is explained in more detail [here](https://stackoverflow.com/a/962829) and [here](https://javascript.info/array-methods#shuffle-an-array) (click on "solution").

.github/workflows/pr-auto-assign.yml
```diff
@@ -16,11 +16,21 @@ jobs:
           script: |
             const reviewers = ['mstoykov', 'imiric', 'codebien', 'olegbespalov', 'oleiade'];
             const reviewerCount = 2;
+            const crypto = require("node:crypto");
+
+            const getNRandom = (n, array) => {
+              let result = new Array();
+              for (;n > 0 && array.length > 0; n--) {
+                const chosen = array[crypto.randomInt(array.length)];
```
imiric, line 24: BTW, we don't need to use `node:crypto` here. The PRNG would work just as well, and probably have less overhead.

So this works too:
```suggestion
                const chosen = array[Math.floor(Math.random()*array.length)];
```

We could also use the Fisher-Yates algorithm described [here](https://javascript.info/array-methods#shuffle-an-array) (click on "solution"), which is likely even more efficient.

Not that efficiency matters at all for this purpose, and this is fine too. Just wanted to mention it in case someone else finds it interesting as well. :slightly_smiling_face:

imiric (approved):
Good to know, thanks for the fix! :bow:

.github/workflows/pr-auto-assign.yml
na--: Well, I guess this shows what copy-pasting a random StackOverflow answer after a cursory "that seems about right" check will get you  :sweat_smile:  Sorry about that! :bow: 

And now, with ChatGPT/Co-pilot/etc, the future is looking very bright indeed :sob:

Issue #3117, opened 2023-06-06
Cloud output v2


codebien:
### Context

https://github.com/grafana/k6/issues/2954 introduces the new experimental Coud output with a Protobuf-based protocol.

#### Memory usage

After the first iteration, the memory usage is higher than required. Especially for the Trend metrics is very easy to saturate the bandwidth in a range from tons of KiloBytes up to the remote limit (1 MB).

We also decided to denormalize some fields to reduce the workload and keep the implementation simple on the remote server but the load generated on the client is high, we should revisit this decision.

#### Fault tolerance

The current flush process could be more fault tolerant, it doesn't retry on failures.

#### Validation

 `__name__` and `test_run_id` are reserved labels for the remote service and if a test also sets them then there are conflicts generating unexpected behavior for the user. A more dev-friendly UX should be implemented.

### Proposal

We identified some actions that should drive us to the goal:

- A more compact Protobuf representation for Histogram.
- Split in multiple requests when the flush process gets a number of time series higher than the `MaxMetricSamplesPerPackage` variable.
- Normalize as MetricSet's fields the common fields across time series.
- Fault-tolerant flush operation.
- Exclude `__name__` and `test_run_id` from the allowed tag names.

### Acceptance criteria

Change the [Cloud output default version](https://github.com/grafana/k6/blob/1df72a0a1b23d7b0827813f6b709c7ad7bb4193b/cloudapi/config.go#L170) to `2`.

### Worklog

```[tasklist]
### Must have
- [ ] https://github.com/grafana/k6/pull/3108
- [ ] https://github.com/grafana/k6/pull/3104
- [ ] https://github.com/grafana/k6/pull/3144
- [ ] https://github.com/grafana/k6/pull/3169
- [ ] https://github.com/grafana/k6/issues/3155
- [ ] https://github.com/grafana/k6/issues/3156
- [ ] https://github.com/grafana/k6/pull/3172
- [ ] https://github.com/grafana/k6/issues/3192
- [ ] https://github.com/grafana/k6/issues/3258 (postponed)
- [ ] https://github.com/grafana/k6/pull/3161
```

```[tasklist]
### Nice to have (in case we need to reduce the scope)
- [ ] https://github.com/grafana/k6/pull/3120
- [ ] https://github.com/grafana/k6/pull/3125
- [ ] https://github.com/grafana/k6/pull/3146
- [ ] https://github.com/grafana/k6/pull/3137
- [ ] https://github.com/grafana/k6/issues/3122
[...]

PR #3169, opened 2023-07-05
output/cloudv2: Compact histogram

codebien:
## What?

Implementation of a more compact Protobuf representation for the Histogram type. This new version only stores and pushes the histogram's significant buckets (non-zero).

## Why?

The current implementation only trims the zero at the limit of the histogram, doing a massive waste of memory if the histogram has a sparse representation. The new version instead allocates only for non-zero buckets.

## Related PR(s)/Issue(s)

Updates #3117


output/cloud/expv2/hdr.go
```diff
@@ -37,14 +38,13 @@ const (
...
-	// in this way we can avoid extra allocation for not significant buckets.
-	// All the zero buckets in between are preserved.
-	Buckets []uint32
+	// It does not include counters for the untrackable values,
+	// because they contain exception cases and require to be tracked in a dedicated way.
+	Buckets map[uint32]uint32
+
+	// Indexes keeps an ordered slice of unique-seen buckets' indexes.
+	// It allows to iterate the buckets in order. It uses an ascendent order.
+	Indexes []uint32
```
codebien, line 47: We may consider using a Btree as an optimization. But I would like to achieve stability in the protocol before.

output/cloud/expv2/hdr.go
```diff
@@ -104,92 +94,72 @@ func (h *histogram) addToBucket(v float64) {
...
-	if h.LastNotZeroBucket >= index {
-		panic("buckets is already bigger than requested index")
+// histogramAsProto converts the histogram into the equivalent Protobuf version.
+func histogramAsProto(h *histogram, time int64) *pbcloud.TrendHdrValue {
+	var (
+		counters []uint32
+		spans    []*pbcloud.BucketSpan
+	)
+
+	// allocate only if at least one item is available
```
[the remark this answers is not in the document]
mstoykov: As we only send observed metrics in the aggregation time period - it should not be possible. The first time we see a metric we create the histogram structure and add the first sample.

output/cloud/expv2/hdr.go
```diff
@@ -104,92 +94,72 @@ func (h *histogram) addToBucket(v float64) {
...
-	case len(h.Buckets) == 0:
-		h.init(index)
-	case index < h.FirstNotZeroBucket:
-		h.prependBuckets(index)
-	case index > h.LastNotZeroBucket:
-		h.appendBuckets(index)
-	default:
-		h.Buckets[index-h.FirstNotZeroBucket]++
+	if h.Buckets == nil {
+		h.Buckets = make(map[uint32]uint32)
```
mstoykov, line 103: I think this should be done just when we make the histogram struct.

output/cloud/expv2/hdr.go
```diff
@@ -104,92 +94,72 @@ func (h *histogram) addToBucket(v float64) {
...
-	// It requires a benchmark if it is better than just copy it.
-
-	newBuckets := make([]uint32, newLen)
-	copy(newBuckets[h.FirstNotZeroBucket-index:], h.Buckets)
-	h.Buckets = newBuckets
-
-	// Update the stats
-	h.Buckets[0] = 1
-	h.FirstNotZeroBucket = index
+	// insert in the middle
```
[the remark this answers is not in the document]
mstoykov: I am also for dropping the comment below

output/cloud/expv2/hdr.go
```diff
@@ -104,92 +94,72 @@ func (h *histogram) addToBucket(v float64) {
...
-		// the slice capacity cap(a) rather than the length"
-		h.Buckets = h.Buckets[:newLen]
-	} else {
-		newBuckets := make([]uint32, newLen)
-		copy(newBuckets, h.Buckets)
-		h.Buckets = newBuckets
-	}
+		// if the current and the previous indexes are not consecutive
+		// consider as closed the current on-going span and start a new one.
+		if diff := h.Indexes[i] - h.Indexes[i-1]; diff > 1 {
```
mstoykov, line 147: For more optimal space use this likely should be 3+ as adding 1 span is at least 2 counters (offset and length) .

So we can add at least 2 zeros before we became even. "at least" here is as I don't know if there is some more overhead for this in the protobuf protocol

mstoykov (dismissed):
LGTM! But we can probably fix some of the small fixes before merging.

The optimizations can likely wait a while and maybe benchmarks ;)

output/cloud/expv2/hdr.go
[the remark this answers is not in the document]
codebien: Without values not, but it is possible to have Indexes and Buckets empty when there are only _untrackable_ values that are tracked into the special extreme buckets.

output/cloud/expv2/hdr.go
codebien: I move it to a constructor

output/cloud/expv2/sink_test.go
```diff
@@ -16,7 +17,14 @@ func TestNewSink(t *testing.T) {
 		{metrics.Counter, &counter{}},
 		{metrics.Gauge, &gauge{}},
 		{metrics.Rate, &rate{}},
-		{metrics.Trend, &histogram{}},
+		{
+			metrics.Trend,
+			&histogram{
```
olegbespalov, line 22: newHistogram maybe :sweat_smile:

olegbespalov (approved):

output/cloud/expv2/sink_test.go
codebien: I was a bit unsure about it, but thinking more about it, it makes sense.

olegbespalov (approved):
Great job! :rocket:

output/cloud/expv2/hdr.go
mstoykov: I wonder if we can't just order the buckets indexes at the end :shrug:. There is no need to constantly keep track of it -if we only want it to be in order :shrug:

mstoykov (approved):

output/cloud/expv2/hdr.go
codebien: Yeah, it simplifies the code a lot, good suggestion.

mstoykov (approved):

olegbespalov (approved):

PR #3171, opened 2023-07-06
Don't try to assign the author of a PR to review it

mstoykov:


olegbespalov (approved):

olegbespalov (approved):

olegbespalov:
It seems like it hasn't taken into account the assigned reviewers :thinking: 

image

mstoykov:
> It seems like it hasn't taken into account the assigned reviewers

It wasn't doing that even before that I think

mstoykov:
> I think it makes sense to assign only the diff between required and already assigned.

Ah... this will be intrinsically racy - I did assign reviewers after making the PR. The action couldn't know it has to wait for 10 seconds for this. 

I also am very much for first *fixing* the action not actually working in some cases at all then it assigning more people.

olegbespalov:
>Ah... this will be intrinsically racy - I did assign reviewers after making the PR. The action couldn't know it has to wait for 10 seconds for this.

Okay, then it's not critical I belive :+1:

olegbespalov (approved):

codebien (approved):

PR #3176, opened 2023-07-07
Add a `promises` package with a `New` helper function

oleiade:
## What?

This Pull Request introduces a `promises` package (under `js`) with a single `New` helper function facilitating the creation of a promise and obtaining its `resolve` and `reject` callback from Go code.

## Why?

We noticed in the context of writing extensions and experimental modules that we would consistently rewrite the same function again and again. Thus, this specific PR aims at declaring this function in a reusable place within the k6 project. 

None of this is definitive, and really is a proposal, if you have improvement ideas, shoot 🚀 

## Checklist

## Related PR(s)/Issue(s)

ref #3165

### DIFF
--- a/js/promises/promises.go
+++ b/js/promises/promises.go
@@ -9,7 +9,7 @@ import (
 // MakeHandledPromise can be used to create promises that will be dispatched to k6's event loop.
 //
 // Calling the function will create a goja promise and return its `resolve` and `reject` callbacks, wrapped
-// in such a way that it will block the k6 JS runtime's event loop from exiting before they are
+// in such a way that it will block the k6 JavaScript runtime's event loop from exiting before they are
 // called, even if the promise isn't resolved by the time the current script ends executing.
 //
 // A typical usage would be:
--- a/js/promises/promises.go
+++ b/js/promises/promises.go
@@ -1,0 +1,46 @@
+// Package promises provides helpers for working with promises in k6.
+package promises
+
+import (
+	"github.com/dop251/goja"
+	"go.k6.io/k6/js/modules"
+)
+
+// MakeHandledPromise can be used to create promises that will be dispatched to k6's event loop.
+//
+// Calling the function will create a goja promise and return its `resolve` and `reject` callbacks, wrapped
+// in such a way that it will block the k6 JS runtime's event loop from exiting before they are
+// called, even if the promise isn't revoled by the time the current script ends executing.
+//
+// A typical usage would be:
+//
+//	   func myAsynchronousFunc(vu modules.VU) *(goja.Promise) {
+//		    promise, resolve, reject := promises.MakeHandledPromise(vu)
+//		    go func() {
+//		        v, err := someAsyncFunc()
+//				   if err != nil {
+//		            reject(err)
+//		            return
+//		        }
+//
+//		        resolve(v)
+//		    }()
+//		    return promise
+//		  }
+func MakeHandledPromise(vu modules.VU) (*goja.Promise, func(interface{}), func(interface{})) {
+	runtime := vu.Runtime()
+	callback := vu.RegisterCallback()
+	promise, resolve, reject := runtime.NewPromise()
+
+	return promise, func(i interface{}) {
+			callback(func() error {
+				resolve(i)
+				return nil
+			})
+		}, func(i interface{}) {
+			callback(func() error {
+				reject(i)
+				return nil
+			})
+		}
+}
--- a/js/promises/promises.go
+++ b/js/promises/promises.go
@@ -27,7 +27,7 @@ import (
 //		    }()
 //		    return promise
 //		  }
-func MakeHandledPromise(vu modules.VU) (promise *goja.Promise, resolve func(result interface{}), reject func(reason interface{})) {
+func New(vu modules.VU) (p *goja.Promise, resolve func(result any), reject func(reason any)) {
 	runtime := vu.Runtime()
 	promise, resolve, reject = runtime.NewPromise()
 	callback := vu.RegisterCallback()
--- a/js/promises/promises.go
+++ b/js/promises/promises.go
@@ -6,7 +6,7 @@ import (
 	"go.k6.io/k6/js/modules"
 )
 
-// MakeHandledPromise can be used to create promises that will be dispatched to k6's event loop.
+// New can be used to create promises that will be dispatched to k6's event loop.
 //
 // Calling the function will create a goja promise and return its `resolve` and `reject` callbacks, wrapped
 // in such a way that it will block the k6 JavaScript runtime's event loop from exiting before they are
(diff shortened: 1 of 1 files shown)


js/promises/promises.go
```diff
@@ -0,0 +1,46 @@
...
+//				   if err != nil {
+//		            reject(err)
+//		            return
+//		        }
+//
+//		        resolve(v)
+//		    }()
+//		    return promise
+//		  }
+func MakeHandledPromise(vu modules.VU) (*goja.Promise, func(interface{}), func(interface{})) {
```
mstoykov, line 30: This is the rare case where I am *for* named results.

```suggestion
func MakeHandledPromise(vu modules.VU) (promise *Promise, resolve func(result interface{}), reject func(reason interface{})) {
```

Copied from [upstream](https://pkg.go.dev/github.com/dop251/goja#Runtime.NewPromise)

js/promises/promises.go
```diff
@@ -0,0 +1,46 @@
...
+import (
+	"github.com/dop251/goja"
+	"go.k6.io/k6/js/modules"
+)
+
+// MakeHandledPromise can be used to create promises that will be dispatched to k6's event loop.
+//
+// Calling the function will create a goja promise and return its `resolve` and `reject` callbacks, wrapped
+// in such a way that it will block the k6 JS runtime's event loop from exiting before they are
+// called, even if the promise isn't revoled by the time the current script ends executing.
```
mstoykov, line 13:
```suggestion
// called, even if the promise isn't resolved by the time the current script ends executing.
```

mstoykov, line 13: The current documentation:
1. does not explain why you would use that instead of `goja.NewRuntime()` 
2. Explains things that arguably are better left explain in the eventloop package - and they are.


I would propose that we do actually document *why* you will want to use this and simple say it uses the underlying eventloop mechanism.

js/promises/promises.go
```diff
@@ -0,0 +1,46 @@
...
+//		        }
+//
+//		        resolve(v)
+//		    }()
+//		    return promise
+//		  }
+func MakeHandledPromise(vu modules.VU) (*goja.Promise, func(interface{}), func(interface{})) {
+	runtime := vu.Runtime()
+	callback := vu.RegisterCallback()
+	promise, resolve, reject := runtime.NewPromise()
```
mstoykov, line 33: Nit: in the unlikely event `NewRuntime` panics, this will not have made a callback that will never be called

```suggestion
	promise, resolve, reject := runtime.NewPromise()
	callback := vu.RegisterCallback()
```

mstoykov (approved):
LGTM in general. 

I have some worries about the documentation. I also find the name ... strange. And I do know big parts of this come from my original code, years ago.

But `promises.NewHandledPromise(vu)` doesn't really tell you what "handled" means. 

I don't have a particular proposition on that part though. My only thought is `promises.NewEasyPromise` as I guess that is the main part people care about - it is easy(ier) to use.

js/promises/promises.go
oleiade: Abolutely agree 👍🏻

js/promises/promises.go
oleiade: Good point 👍🏻 Will adapt it indeed 🚀

oleiade:
I totally agree on the naming, and I'm happy you bring it up, what would you folks think of `CreatePromise` or `PreparePromise`, `RegisterPromise` (to stay along the lines of `RegisterCallback), @codebien @mstoykov ? Any other ideas?

js/promises/promises.go
oleiade: Here's an attempt at improving the documentation accordingly:

```
// MakeHandledPromise facilitates the creation of promises within k6 modules.
//
// Whenever a k6 module function or methods returns a promise, it very likely
// should be created using this function as it ensures the promise is correctly
// managed by k6's event loop.
//
// The function returns a promise, a resolve function and a reject function. The
// promise is the one that should be returned by the function, and the `resolve`
// and `reject` functions should be used to resolve or reject the promise.
//
// Usage example:
//
//	   func myAsynchronousFunction(vu modules.VU) *goja.Promise {
//		    promise, resolve, reject := CreateAsyncFunctionPromise(vu)
//		    go func() {
//		        value, err := someAsynchronousOperation()
//				   if err != nil {
//		            reject(err)
//		            return
//		        }
//
//		        resolve(value)
//		    }()
//		    return promise
//		  }
//
// In the above example, `someAsynchronousOperation` represents an operation that might take a while to finish,
// and k6 script execution shouldn't be halted during its execution. This function ensures such an operation
// is correctly managed within k6's event loop.
```

WDYT? Suggestions to improve?

js/promises/promises.go
```diff
@@ -0,0 +1,46 @@
...
+
+import (
+	"github.com/dop251/goja"
+	"go.k6.io/k6/js/modules"
+)
+
+// MakeHandledPromise can be used to create promises that will be dispatched to k6's event loop.
+//
+// Calling the function will create a goja promise and return its `resolve` and `reject` callbacks, wrapped
+// in such a way that it will block the k6 JS runtime's event loop from exiting before they are
```
codebien, line 12:
```suggestion
// in such a way that it will block the k6 JavaScript runtime's event loop from exiting before they are
```

js/promises/promises.go
```diff
@@ -0,0 +1,46 @@
...
+//				   if err != nil {
+//		            reject(err)
+//		            return
+//		        }
+//
+//		        resolve(v)
+//		    }()
+//		    return promise
+//		  }
+func MakeHandledPromise(vu modules.VU) (promise *goja.Promise, resolve func(result interface{}), reject func(reason interface{})) {
```
codebien, line 30:
```suggestion
func New(vu modules.VU) (p *goja.Promise, resolve func(result any), reject func(reason any)) {
```
This should be already clear

oleiade: I love it ❤️

js/promises/promises.go
```diff
@@ -0,0 +1,46 @@
...
+//		            return
+//		        }
+//
+//		        resolve(v)
+//		    }()
+//		    return promise
+//		  }
+func New(vu modules.VU) (p *goja.Promise, resolve func(result any), reject func(reason any)) {
+	runtime := vu.Runtime()
+	p, resolve, reject = runtime.NewPromise()
```
codebien, line 32:
```suggestion
	p, resolve, reject = vu.Runtime().NewPromise()
```
nit

codebien (commented):
LGTM

js/promises/promises.go
oleiade: Fair enough 😄

codebien (approved):
@oleiade a simple test should be added, ofc you can do it in a different low-priority PR.

PR #3177, opened 2023-07-07
golangci-lint version update & minor fixes

olegbespalov:
## What?

Update the golangci lint to the latest version and fix some `nolint` (linter is not used) issues.

## Why?

The reason behind this is a minor inconvenience after the changes for the `deepguard` linter that caused inline reporting of the blocked package.

## Checklist

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,4 +1,4 @@
-# v1.51.2
+# v1.53.3
 # Please don't remove the first line. It uses in CI to determine the golangci version
 run:
   deadline: 5m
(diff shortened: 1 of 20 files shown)


mstoykov (dismissed):

.golangci.yml
```diff
@@ -59,6 +59,12 @@ linters-settings:
       # Forbid everything in syscall except the uppercase constants
       - '^syscall\.[^A-Z_]+$(# Using anything except constants from the syscall package is forbidden )?'
       - '^logrus\.Logger$'
+  depguard:
+     rules:
+      main:
+        deny:
+          - pkg: io/ioutil
+            desc: ioutil is deprecated, see https://go.dev/doc/go1.16#ioutil
```
codebien, line 67: Why are we adding it? I checked this better the last time I suggested it and it should not be necessary. The current linters already catch it.

codebien: Below the message from `staticcheck`, it sounds acceptable to me, so I'm for just dropping `depguard` (with the assumption that we are not benefiting from any default configuration).

```
SA1019: "io/ioutil" has been deprecated since Go 1.19: As of Go 1.16, the same functionality is now provided by package io or package os, and those implementations should be preferred in new code. See the specific function documentation for details.
```

olegbespalov: Dropped :+1:

mstoykov (approved):

codebien (approved):

PR #3179, opened 2023-07-10
Cross-VU shared grpc.ClientConn.

chrismoran-mica:
## What?

Add a parameter to `connectParams` that would allow `grpc.ClientConn`s to be shared across VUs. In order for connections to be shared (within a VU or across VUs) the following must be true:

1. A connection is eligible to be shared if a `Client` is `connect`ed with `{ connectionSharing: true }`
2. A connection will only be shared with another `Client` (within a VU or across VUs) if the `addr` (address) parameter to `connect` is the same (case-insensitive)
3. The decision to permit a `Client`s connection to be shared is made when `connect` is called

Logically, it would only make sense to share a connection (within a VU or across VUs) if the two `Client`s in question target an endpoint that either:

1. Serves the same gRPC service
2. Multiplexes the gRPC services needed by both `Client`s

Aside from comparing `protoreflect.MethodDescriptor`s, I cannot think of a way to ensure the `Client`s should share a connection prior to making the connection and calling `invoke` (to generate the `protoreflect.MethodDescriptor` mapping). So it would be left to the test writer to decide if the `Client`s should share connections.

## Why?

This feature has the potential to substantially reduce the resources required for multiple VUs connecting to the same service(s) or endpoint(s). This gives greater flexibility for testing scenarios. 

On the other hand, this does _blur the line_ among VUs with `connectionSharing` enabled on `grpc.Client`s. There could potentially be unforeseen consequences to sharing connections among VUs. This situation, however, is mitigated by the fact that this feature is opt-in and disabled by default.

TODO / In-process:

- Limit the amount of sharing allowed by a single connection (via Latch or other such mechanism) to ensure a single gRPC connection does not exhaust the maximum permitted open channels on the server side
  - This would require the client to know at connection time what that maximum is and if not set, use a sane default: (typically 100)
  - This could also potentially introduce complicated connection management dealing with multiple shared, open connections across many VUs

### Connection sharing limiting

- The `ConnectionSharing` parameter can be either a boolean value OR an integer > 1
  - A `true` value will mean ConnectionSharing is enabled for the Client and the maximum sharing is set to a sane default (100)
    - This further means that the client's underlying connection will be shared with at most (100 - 1) clients 
  - A `false` value will mean ConnectionSharing is disabled entirely _for that Client_
  - An integer > 1 will mean ConnectionSharing is enabled for the Client and the underlying connection will be shared at most `connectionSharing - 1` number of times before a new connection is made.
  - It is important to note that the number of _shares_ a client's connection has depends solely on the value of the `connectionSharing` parameter exactly when the actual connection was dialed.

Limiting the number of connections can be a complicated concept. Since each `client.connect` can have a `connectionSharing` parameter with any arbitrary value for `connectionSharing` it can quickly become confusing which connections are possibly being shared.

## Checklist

- [x] Limit the connection sharing amount using a maximum connection sharing amount parameter and/or a sane default
- [x] Limit complexity of the connection sharing logic

- [ ] ~~I have run linter locally (`make ci-like-lint`) and all checks pass.~~


mstoykov:
> I wanted to open a discussion on this topic before putting more polish on it.

As recommended in the [Contributing guide](https://github.com/grafana/k6/blob/master/CONTRIBUTING.md), it is better to open issues to discuss what and how you will want to change. 

> Find an issue you'd like to fix. If there is none already, or you'd like to add a feature, please open one, and we can talk about how to do it. Out of respect for your time, please start a discussion regarding any bigger contributions either in a GitHub Issue, in the community forums before you get started on the implementation.

> Remember, there's more to software development than code; if it's not properly planned, stuff gets messy real fast.

mstoykov:
Thanks for the PR.

As we have agreed this will first go in the extension and there is already open [PR](https://github.com/grafana/xk6-grpc/pull/31). 

So I will close it and when we decide to back  port it a new PR will be made.

Thanks again :bow:

PR #3180, opened 2023-07-10
Address eventloop's double registerCallback call when using promises.New

oleiade:
## What?

This Pull Request addresses the misuse of named arguments introduced by #3176, and fixes the resulting `registerCallback called twice` issue that followed from it.

It also adds two minimal tests covering ensuring the `resolve` and `reject` callbacks are effective. I wasn't entirely sure how to properly test that function; this is the simplest thing I came up with. I'm all ears 👂🏻 for alternatives and suggestions to improve it 🙇🏻 

## Why?

As I started using the `promises.New` function in another stream of work, I noticed that as soon as I used the `promises.New` function, I would end up having `registerCallback called twice` panic all over. I tracked down the issue, and it turned out that it was because we needed to use Go named return arguments properly.

## Checklist

## Related PR(s)/Issue(s)

ref #3176
ref https://github.com/grafana/k6/pull/3176#pullrequestreview-1521556345

### DIFF
--- a/js/promises/promises.go
+++ b/js/promises/promises.go
@@ -28,18 +28,22 @@ import (
 //		    return promise
 //		  }
 func New(vu modules.VU) (p *goja.Promise, resolve func(result any), reject func(reason any)) {
-	p, resolve, reject = vu.Runtime().NewPromise()
+	p, resolveFunc, rejectFunc := vu.Runtime().NewPromise()
 	callback := vu.RegisterCallback()
 
-	return p, func(i interface{}) {
-			callback(func() error {
-				resolve(i)
-				return nil
-			})
-		}, func(i interface{}) {
-			callback(func() error {
-				reject(i)
-				return nil
-			})
-		}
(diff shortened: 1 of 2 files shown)


codebien (approved):

js/promises/promises.go
```diff
@@ -28,18 +28,22 @@ import (
 //		    return promise
 //		  }
 func New(vu modules.VU) (p *goja.Promise, resolve func(result any), reject func(reason any)) {
-	p, resolve, reject = vu.Runtime().NewPromise()
+	p, resolveFunc, rejectFunc := vu.Runtime().NewPromise()
 	callback := vu.RegisterCallback()
 
-	return p, func(i interface{}) {
-			callback(func() error {
-				resolve(i)
```
mstoykov, line 36: So for future reference - the problem here was that `resolve` here  is the same resolve we return - which is the function we in. So this just recurse inside itself.

mstoykov (approved):

PR #3181, opened 2023-07-10
Add a `RunOnEventLoop` method to `modulestest.Runtime`

oleiade:
## What?

This Pull Request adds a `RunOnEventLoop` method to the `modulestest.Runtime`, to facilitate testing code that should be dispatched to the event loop, such as code that returns promises.

## Why?

While writing tests for the #3142, I stumbled upon this code from @mstoykov, with a `TODO` attached. I thought that as I was about to copy-paste that snippet anyway, I might as well do as instructed in `TODO` and move the code to the `modulestest.Runtime` class. 

The `TODO` hinted at potential improvements without detailing that I'd be nonetheless happy to include if necessary 🤞🏻 

## Checklist

### DIFF
--- a/js/modulestest/runtime.go
+++ b/js/modulestest/runtime.go
@@ -74,6 +74,34 @@ func (r *Runtime) SetupModuleSystemFromAnother(another *Runtime) error {
 	return r.innerSetupModuleSystem()
 }
 
+// RunOnEventLoop will run the given code on the event loop.
+//
+// It is meant as a helper to test code that is expected to be run on the event loop, such
+// as code that returns a promise.
+//
+// A typical usage is to facilitate writing tests for asynchrounous code:
+//
+//	func TestSomething(t *testing.T) {
+//	    runtime := modulestest.NewRuntime(t)
+//
+//	    err := runtime.RunOnEventLoop(`
+//	        doSomethingAsync().then(() => {
+//	            // do some assertions
+//	        });
+//	    `)
+//	    require.NoError(t, err)
+//	}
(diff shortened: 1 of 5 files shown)


js/modulestest/runtime.go
```diff
@@ -74,6 +74,34 @@ func (r *Runtime) SetupModuleSystemFromAnother(another *Runtime) error {
...
+//	    runtime := modulestest.NewRuntime(t)
+//
+//	    err := runtime.RunOnEventLoop(`
+//	        doSomethingAsync().then(() => {
+//	            // do some assertions
+//	        });
+//	    `)
+//	    require.NoError(t, err)
+//	}
+func (r *Runtime) RunOnEventLoop(code string) error {
```
olegbespalov, line 94: It seems like we already have something similar:
https://github.com/grafana/k6/blob/master/js/modules/k6/k6_test.go#L432-L439

So if we want to introduce that exposed method, we should keep only one implementation and usage. What do you think?

oleiade: We have  a bunch of similar functions scattered across the codebase indeed 🤝 Another occurence I'm aware of is the http package which has one to test the `asyncRequest`.

If you're suggesting we should also address the know use case in this PR, I rather agree 👍🏻 I'll give it a shot, let me know if you find more similar cases that we should also address 🙇🏻

oleiade: I've added tests accordingly for all the code I could find in the codebase that did some `EventLoop.Start` in one way or another 👍🏻

mstoykov (approved):

olegbespalov (approved):
:rocket:

PR #3182, opened 2023-07-10
cloudv2/hdr: Fix off-by-one error for spans

codebien:
## What?

Fix the histogram's Spans encoding.
 
i.e., when encoding a sequence like the following: `[1, 1, 1, 0, 2, 2]`

The second span must be:
`(offset=1, length=2)`

And not:
`(offset=2, length=2)`

## Why

`histogram.Spans.Offset` is expected to be the counter of the empty buckets and not the diff between the indexes.

## Related PR(s)/Issue(s)

Follow-up on https://github.com/grafana/k6/pull/3169

### DIFF
--- a/output/cloud/expv2/hdr.go
+++ b/output/cloud/expv2/hdr.go
@@ -128,8 +128,8 @@ func histogramAsProto(h *histogram, time int64) *pbcloud.TrendHdrValue {
 
 		// if the current and the previous indexes are not consecutive
 		// consider as closed the current on-going span and start a new one.
-		if diff := indexes[i] - indexes[i-1]; diff > 1 {
-			spans = append(spans, &pbcloud.BucketSpan{Offset: diff, Length: 1})
+		if zerosBetween := indexes[i] - indexes[i-1] - 1; zerosBetween > 0 {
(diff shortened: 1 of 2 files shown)


output/cloud/expv2/hdr.go
```diff
@@ -129,7 +129,12 @@ func histogramAsProto(h *histogram, time int64) *pbcloud.TrendHdrValue {
 		// if the current and the previous indexes are not consecutive
 		// consider as closed the current on-going span and start a new one.
 		if diff := indexes[i] - indexes[i-1]; diff > 1 {
-			spans = append(spans, &pbcloud.BucketSpan{Offset: diff, Length: 1})
+			spans = append(spans, &pbcloud.BucketSpan{
+				// Offset counts the empty buckets between two non-zero buckets.
+				Offset: diff - 1,
```
mstoykov, line 134: nitPick: I would actually renaame `diff` to `zerosBetween` and then check it against 0, but keep the single line `append`

mstoykov (approved):

mstoykov (approved):

olegbespalov (approved):

PR #3183, opened 2023-07-10
Fix race in js/promises package's tests

mstoykov:
The original tests were interacting with the runtime and the event loop from outside the eventloop.


olegbespalov (approved):

oleiade (approved):
Good catch 👀

PR #3193, opened 2023-07-13
cloudv2: Downscale the batch size

codebien:
The backend has the limit set to 1k so it doesn't make sense to use a bigger value forcing the system to split a single batch in multiple jobs.

We are already overwriting the default value for most of the cases.  This PR removes the requirement for setting the env var for doing the overwrite simplifing internal operations.

### DIFF
--- a/cloudapi/config.go
+++ b/cloudapi/config.go
@@ -175,9 +175,13 @@ func NewConfig() Config {
 		TracesPushConcurrency: null.NewInt(1, false),
 
 		MaxMetricSamplesPerPackage: null.NewInt(100000, false),
-		MaxTimeSeriesInBatch:       null.NewInt(10000, false),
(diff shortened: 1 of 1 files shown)


mstoykov (approved):

oleiade (dismissed):
Might be worth adding a comment explaining why we set it too 1000 directly in the code, as you explained here, but not blocking 👍🏻

codebien:
@oleiade good idea

mstoykov (approved):

codebien:
@olegbespalov added you to make sure you're aware of the change. We don't require anymore to set this option in our operations because the default already matches the requirement.

olegbespalov (approved):

mstoykov (approved):

PR #3194, opened 2023-07-13
Fix loki log output exiting before push is finished

mstoykov:
After #2833 log outputs were stopped but loki didn't wait until it push was finished to signal it was done.

Which leads to the actual k6 process exiting before loki could flush its messages.

### DIFF
--- a/log/loki.go
+++ b/log/loki.go
@@ -150,10 +150,13 @@ func (h *lokiHook) Listen(ctx context.Context) {
 		pushCh     = make(chan chan int64)
 	)
 
+	pushDone := make(chan struct{})
+	defer func() { <-pushDone }()
 	defer ticker.Stop()
 	defer close(pushCh)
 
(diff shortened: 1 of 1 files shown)


log/loki.go
```diff
@@ -150,10 +150,13 @@ func (h *lokiHook) Listen(ctx context.Context) {
 		pushCh     = make(chan chan int64)
 	)
 
+	pushDone := make(chan struct{})
+	defer func() { <-pushDone }()
```
mstoykov, line 154: I did use a sync.WaitGroup in thePoC, but decided there is no need to import the `sync` package and this code is even shorter so :shrug:

codebien (approved):
I know this code is complex and maybe not so easy, but it would be good to have a fixed test that fails without this patch.

Up to you if you have time for it.

olegbespalov (approved):

PR #3196, opened 2023-07-13
log packages lint fixes and test for loki flushing at end

mstoykov:


mstoykov:
Those are the tests for #3194

log/loki_test.go
```diff
@@ -139,3 +150,44 @@ func TestFilterLabels(t *testing.T) {
...
+			receivedData <- string(b)
+			close(receivedData) // this is mostly as this should never be called twice, so panicking here in that case
+		}),
+	)
+	configLine := fmt.Sprintf("loki=%s,pushPeriod=1h", srv.URL)
+	h, err := LokiFromConfigLine(nil, configLine)
+	require.NoError(t, err)
+
+	ctx, cancel := context.WithCancel(context.Background())
+	require.NoError(t, err)
```
olegbespalov, line 172:
```suggestion
```

log/loki_test.go
```diff
@@ -139,3 +150,44 @@ func TestFilterLabels(t *testing.T) {
...
+	wg := new(sync.WaitGroup)
+	now := time.Now()
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		err = h.Fire(&logrus.Entry{Time: now, Level: logrus.InfoLevel, Message: "test message"})
+		time.Sleep(time.Millisecond * 10)
```
olegbespalov, line 179: Could this be flaky?

mstoykov: In my local experience - no.

The only *really* flaky part is that `h.Listen` should be running before we call the `cancel` (below). 

And there realistically isn't a way to make that certain, unless `h.Listen` does something additional to signal that.
The easiest way will be to let it push logs *once* and then push this new entry and cancel. But IMO this will be:
1. just as flaky if not as we now will also race with it not sending logs once more of its own timer instead of because we cancel.
2. will make the test even longer as we will need to wait for one push but not have them fast enough, so we do not get the above race.

In my local experiment k6 shutting didn't once send the logs in time before this fix. And neither did this test on older versions.  

I would propose to just bump this timer in case it is flaky and :crossed_fingers:.

olegbespalov (approved):

oleiade (approved):

PR #3199, opened 2023-07-14
Build and publish browser-enabled Docker image

### LINKED ISSUE #2914: Publish Docker image that includes a browser
As per this [conversation](https://github.com/grafana/k6-docs/pull/996#discussion_r1102454704), [The k6 Docker](https://github.com/grafana/k6/blob/a78916c644a4ab741723db5d2050725163a569fe/Dockerfile) doesn't include Chromium so if someone wants to use grafana/k6 directly, they won't be able to run browser tests. 

There is an example [Dockerfile](https://github.com/grafana/xk6-browser/blob/main/Dockerfile) from xk6-browser so I think we also need a similar one for k6.

I'm not sure what the best way forward so raising this as an issue as we believe this needs more discussion with the k6-core side.

andrewslotin:
## What?

This PR adds a stage to the Dockerfile that adds `chromium` to the Docker image and enables k6 Browser API. The new stage is used to build and publish images with `-with-browser` label suffix.

The resulting image is fairly large (~220MB compressed compared to ~16MB without `chromium` installed), hence a separate label for those who want to opt-in to use Browser API.

## Why?

Currently there is no easy way to use Browser API when running k6 as a Docker container. This makes it difficult to run browser-based tests in CI, since users need to first launch browser process in a separate container, provide the CDP URL to the k6 container and conditionally use `browser.connect()` instead of `browser.launch()` in their test scripts.

## Checklist

- [x] I have tested the build pipeline changes in [andrewslotin/k6](https://github.com/andrewslotin/k6) fork (see the [Docker Hub image](https://hub.docker.com/r/andrewslotin/k6/tags))

## Related PR(s)/Issue(s)

Closes #2914

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,3 +23,21 @@ FROM release as legacy
 COPY entrypoint-legacy.sh /usr/bin/
 
 ENTRYPOINT ["/usr/bin/entrypoint-legacy.sh"]
+
+# Browser-enabled bundle
+FROM release as with-browser
+
+USER root
+
+COPY --from=release /usr/bin/k6 /usr/bin/k6
+RUN apk --no-cache add chromium-swiftshader
+
+USER k6
+
+ENV CHROME_BIN=/usr/bin/chromium-browser
+ENV CHROME_PATH=/usr/lib/chromium/
+
+ENV K6_BROWSER_ENABLED=true
+ENV K6_BROWSER_HEADLESS=true
+
+ENTRYPOINT ["k6"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,7 +18,7 @@ WORKDIR /home/k6
 ENTRYPOINT ["k6"]
 
 # Legacy loadimpact/k6 image
-FROM release
+FROM release as legacy
 
 COPY entrypoint-legacy.sh /usr/bin/
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -204,7 +204,7 @@ jobs:
             docker push "$DOCKER_IMAGE_ID:latest"
           fi
       - name: Build loadimpact/k6
-        run: docker build -t $LI_DOCKER_IMAGE_ID .
+        run: docker build -t $LI_DOCKER_IMAGE_ID --target legacy .
       - name: Publish loadimpact/k6:master image to Docker Hub
         if: ${{ github.ref == 'refs/heads/master' }}
         run: |
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -154,6 +154,8 @@ jobs:
             docker run $DOCKER_IMAGE_ID pause --help
             docker run $DOCKER_IMAGE_ID resume --help
 
+      - name: Build browser-enabled image
+        run: docker build -t $DOCKER_IMAGE_ID:with-browser --target with-browser .
       - name: Log into ghcr.io
         if: ${{ github.ref == 'refs/heads/master' || startsWith(github.ref, 'refs/tags/v') }}
         run: |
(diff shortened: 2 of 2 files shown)


Dockerfile
```diff
@@ -23,3 +23,20 @@ FROM release as legacy
...
+
+COPY --from=release /usr/bin/k6 /usr/bin/k6
+RUN apk --no-cache add chromium-swiftshader
+
+USER k6
+
+ENV K6_BROWSER_ENABLED=true
+ENV CHROME_BIN=/usr/bin/chromium-browser
+ENV CHROME_PATH=/usr/lib/chromium/
+ENV CHROMIUM_FLAGS="--disable-software-rasterizer --disable-dev-shm-usage"
```
ankur22, line 40: What's the reason for setting these flags?

Can we use `ENV K6_BROWSER_HEADLESS=true` and rely on the default flags that the browser module works with?

ankur22 (commented):
Thanks for this! I just have that one question around the flags and running in headless, but otherwise it looks good.

Dockerfile
andrewslotin: Good catch on the headless flag!

`--disable-dev-shm-usage` prevents Chrome from using `/dev/shm` that is [too small by default in Docker](https://github.com/puppeteer/puppeteer/blob/main/docs/troubleshooting.md#tips), and makes it use `/tmp` instead.
`--disable-software-rasterizer` turns off RAM-greedy GPU rasterizer, again due to limited resources available to the container.

andrewslotin: I've added a comment that explains what are these flags for.

ankur22: The browser module sets `disable-dev-shm-usage` when it launches a chrome instance. I've not witnessed any issues so far in an running container that would require `disable-software-rasterizer`, and the browser module doesn't use that flag.

My concern is that this flag could alter the behaviour of chrome somehow that could make it difficult to debug issues compared to when we run it inside/outside of a container without that flag.

If you still think it's worth adding, then could you work with `K6_BROWSER_ARGS` instead of `CHROMIUM_FLAGS`? I'm 100% sure that any new args are added to the launch process when working with `K6_BROWSER_ARGS`, but i'm not sure if flags defined in `CHROMIUM_FLAGS` are ignored or not when we launch chrome from the browser module.

andrewslotin: Makes sense, I'll move the args to the `K6_BROWSER_ARGS` to make sure we have control over them.

Dockerfile
```diff
@@ -18,8 +18,29 @@ WORKDIR /home/k6
...
+USER k6
+
+ENV CHROME_BIN=/usr/bin/chromium-browser
+ENV CHROME_PATH=/usr/lib/chromium/
+
+ENV K6_BROWSER_ENABLED=true
+ENV K6_BROWSER_HEADLESS=true
+# Disable RAM-heavy GPU rasterizer and make chromium use /tmp instead of /dev/shm that is typically
+# too small in Docker
+ENV K6_BROWSER_ARGS="--disable-software-rasterizer --disable-dev-shm-usage"
```
ka3de, line 44: I believe this should not include the starting `--`, and should be a comma separated list in our current [implementation](https://github.com/grafana/xk6-browser/blob/1d75cd5b32feba6135a615e6f86bcd1d06bf0fa0/common/browser_options.go#L96). See launch options args [doc](https://k6.io/docs/javascript-api/k6-experimental/browser/browsertype/launch/).
```suggestion
ENV K6_BROWSER_ARGS="disable-software-rasterizer,disable-dev-shm-usage"
```

Dockerfile
andrewslotin: Interesting, running it with `--` and space as a delimiter did not make any difference and my tests has passed successfully. I'm now leaning towards the initial suggestion from @ankur22 and remove these flags.

ankur22 (approved):
LGTM

mstoykov (approved):
LGTM!

Are you going to squash this PR if not I would recommend force pushing to fix the `fixup`s

andrewslotin:
@mstoykov, my plan is to rebase and squash the fixup commits before merging.

PR #3201, opened 2023-07-17
k6 Insights (1/2): Refactor request metadata output to a separate package

Blinkuu:
Due to cloud output v2 being at risk for v0.46.0, we decided to incorporate the new request metadata output into cloud output v1. This change is done to mitigate the risk of failing to deliver the new insights features.


output/cloud/insights/flush.go
```diff
@@ -0,0 +1,44 @@
...
+}
+
+// Flush retrieves data from the collector and sends it to the insights backend.
+func (f *Flusher) Flush() error {
+	requestMetadatas := f.collector.PopAll()
+	if len(requestMetadatas) < 1 {
+		return nil
+	}
+
+	return f.client.IngestRequestMetadatasBatch(context.Background(), requestMetadatas)
```
olegbespalov, line 43: Since we are passing here the `context.Background()` and this is a net operation, is the timeout inside the client configured?

output/cloud/expv2/output.go
```diff
@@ -156,14 +150,14 @@ func (o *Output) Start() error {
 		}
 		insightsClient := insights.NewClient(insightsClientConfig)
 
-		ctx, cancel := context.WithTimeout(context.Background(), 1*time.Second)
+		ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
```
olegbespalov, line 153: Why has this increased ten times? :thinking: 

If we are not sure about the value, it seems better to make this configurable

olegbespalov (commented):
It generally looks good. Left two minor comments :+1:

Also, adding the @oleiade as the second reviewer from the core.

output/cloud/insights/flush.go
Blinkuu: Yes, you can find this [here](https://github.com/grafana/k6/blob/47c78a5a00881f5a10ebb9e11e94e2ba05b05ddb/output/cloud/expv2/output.go#L128-L151).

output/cloud/expv2/output.go
Blinkuu: I have increased this since 1 second to establish a connection is not that long, considering that our backend is in a single region, but the cloud tests can be distributed geographically. This can be inferred from RTT times across the globe.

olegbespalov (approved):

output/cloud/expv2/output.go
Blinkuu: Ideally, yeah. However, these are already tuned based on a few experiments we did in the past. Let's leave this for the future.

PR #3202, opened 2023-07-17
k6 Insights (2/2): Integrate request metadata output to cloud output v1

Blinkuu:
## What?

This PR incorporates the request metadata output into cloud output v1.

## Why?

Due to cloud output v2 being at risk for v0.46.0, we decided to incorporate the new request metadata output into cloud output v1. This change is done to mitigate the risk of failing to deliver the new insights features.

## Related PR(s)/Issue(s)

This PR is part of a PR chain. Previous PR - https://github.com/grafana/k6/pull/3201.

Previous PRs:
- https://github.com/grafana/k6/pull/3100
- https://github.com/grafana/k6/pull/3158
- https://github.com/grafana/k6/pull/3184

### DIFF
--- a/output/cloud/v1/output.go
+++ b/output/cloud/v1/output.go
@@ -123,9 +123,7 @@ func (out *Output) Start() error {
 		)
 		insightsClient := insights.NewClient(insightsClientConfig)
 
-		ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
-		defer cancel()
-		if err := insightsClient.Dial(ctx); err != nil {
+		if err := insightsClient.Dial(context.Background()); err != nil {
 			return err
 		}
 
--- a/output/cloud/v1/output.go
+++ b/output/cloud/v1/output.go
@@ -16,7 +16,6 @@ import (
 	"go.k6.io/k6/cloudapi/insights"
 	"go.k6.io/k6/errext"
 	"go.k6.io/k6/errext/exitcodes"
-	"go.k6.io/k6/lib/types"
 	insightsOutput "go.k6.io/k6/output/cloud/insights"
 
 	"go.k6.io/k6/lib/netext"
@@ -117,29 +116,11 @@ func (out *Output) Start() error {
 		}
 		out.requestMetadatasCollector = insightsOutput.NewCollector(testRunID)
 
-		insightsClientConfig := insights.ClientConfig{
-			IngesterHost: out.config.TracesHost.String,
-			Timeout:      types.NewNullDuration(90*time.Second, false),
-			AuthConfig: insights.ClientAuthConfig{
-				Enabled:                  true,
-				TestRunID:                testRunID,
-				Token:                    out.config.Token.String,
-				RequireTransportSecurity: true,
-			},
-			TLSConfig: insights.ClientTLSConfig{
-				Insecure: false,
-			},
-			RetryConfig: insights.ClientRetryConfig{
-				RetryableStatusCodes: `"UNKNOWN","INTERNAL","UNAVAILABLE","DEADLINE_EXCEEDED"`,
-				MaxAttempts:          3,
-				PerRetryTimeout:      30 * time.Second,
-				BackoffConfig: insights.ClientBackoffConfig{
-					Enabled:        true,
-					JitterFraction: 0.1,
-					WaitBetween:    1 * time.Second,
-				},
-			},
-		}
+		insightsClientConfig := insights.NewDefaultClientConfigForTestRun(
+			out.config.TracesHost.String,
+			out.config.Token.String,
+			testRunID,
+		)
 		insightsClient := insights.NewClient(insightsClientConfig)
 
 		ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
--- a/output/cloud/v1/output.go
+++ b/output/cloud/v1/output.go
@@ -3,16 +3,21 @@
 package cloud
 
 import (
+	"context"
 	"net/http"
+	"strconv"
 	"sync"
 	"time"
 
-	easyjson "github.com/mailru/easyjson"
+	"github.com/mailru/easyjson"
 	"github.com/sirupsen/logrus"
 
 	"go.k6.io/k6/cloudapi"
+	"go.k6.io/k6/cloudapi/insights"
 	"go.k6.io/k6/errext"
 	"go.k6.io/k6/errext/exitcodes"
+	"go.k6.io/k6/lib/types"
+	insightsOutput "go.k6.io/k6/output/cloud/insights"
 
 	"go.k6.io/k6/lib/netext"
 	"go.k6.io/k6/lib/netext/httpext"
@@ -33,6 +38,10 @@ type Output struct {
 	bufferMutex      sync.Mutex
 	bufferHTTPTrails []*httpext.Trail
 	bufferSamples    []*Sample
+
+	insightsClient            insightsOutput.Client
+	requestMetadatasCollector insightsOutput.RequestMetadatasCollector
+	requestMetadatasFlusher   insightsOutput.RequestMetadatasFlusher
 
 	// TODO: optimize this
 	//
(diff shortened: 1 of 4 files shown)


output/cloud/v1/output.go
```diff
@@ -472,4 +533,42 @@ func (out *Output) pushMetrics() {
...
+
+	for i := int64(0); i < out.config.TracesPushConcurrency.Int64; i++ {
+		out.outputDone.Add(1)
+		go func() {
+			defer out.outputDone.Done()
+			defer t.Stop()
+
+			for {
+				select {
+				case <-out.stopSendingMetrics:
```
olegbespalov, line 547: It seems like this is related more to the metrics, not the traces. Out of curiosity, why is this here? :thinking: 

I'm also asking since generally code of the `runFlushRequestMetadatas` isn't so different from the v2 and probably could be extracted to the insights' package

output/cloud/v1/output.go
```diff
@@ -101,6 +110,49 @@ func (out *Output) Start() error {
...
+				BackoffConfig: insights.ClientBackoffConfig{
+					Enabled:        true,
+					JitterFraction: 0.1,
+					WaitBetween:    1 * time.Second,
+				},
+			},
+		}
+		insightsClient := insights.NewClient(insightsClientConfig)
+
+		ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
```
olegbespalov, line 145: Same comment about the 10 seconds https://github.com/grafana/k6/pull/3201#discussion_r1265632026

olegbespalov (changes requested):
Let's keep config under a single source of truth.

And question about the `runFlushRequestMetadatas`

output/cloud/v1/output.go
```diff
@@ -101,6 +110,49 @@ func (out *Output) Start() error {
...
 	}
 
+	if insightsOutput.Enabled(out.config) {
+		testRunID, err := strconv.ParseInt(out.referenceID, 10, 64)
+		if err != nil {
+			return err
+		}
+		out.requestMetadatasCollector = insightsOutput.NewCollector(testRunID)
+
+		insightsClientConfig := insights.ClientConfig{
```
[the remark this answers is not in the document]
Blinkuu: Great idea, done in [a65ec43](https://github.com/grafana/k6/pull/3202/commits/a65ec438167e3e2fe3ad7b3644d2af1d0d91a03b).

output/cloud/v1/output.go
Blinkuu: Our backend does not support aborting test runs. I think it makes sense to stop sending data to our backend as well once the core backend figures out that the test `shouldStopSendingMetrics(...)`.

output/cloud/v1/output.go
```diff
@@ -472,4 +514,42 @@ func (out *Output) pushMetrics() {
...
+
+func (out *Output) flushRequestMetadatas() {
+	start := time.Now()
+
+	err := out.requestMetadatasFlusher.Flush()
+	if err != nil {
+		out.logger.WithError(err).WithField("t", time.Since(start)).Error("Failed to push trace samples to the cloud")
+	}
+
+	out.logger.WithField("t", time.Since(start)).Debug("Successfully flushed buffered trace samples to the cloud")
```
olegbespalov, line 552: In case of an error, this log entry still will be written. This applies to the v2 output too.

output/cloud/v1/output.go
olegbespalov: Even if we agreed that making this configurable isn't a priority, it's still better to keep this along with the other configuration (part of the `ClientConfig`). That way, we will be sure that now the configuration is distributed (:crossed_fingers:), and the next dev who should adjust it thank us since it's pretty clear where it should be adjusted.

Also, do you think about making the connection logic internal thing of the sight client, by the way? I mean, introduce something like:

```golang
func (c *Client) Connect() error {
   ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
   defer cancel()
   if err := insightsClient.Dial(ctx); err != nil {
		return err
	}

    return nil
}
```

output/cloud/v1/output.go
Blinkuu: A semantic difference exists between `<-out.stopSendingMetrics` and `<-out.stopOutput`. Stop output is for a graceful exit - we still flush what's remaining. In case of "stop sending metrics", we exit and stop sending data. We are not coupling two domains here. The tracing metrics are highly correlated with HTTP metrics. There is no reason to send data to the insights backend if we don't send data to the metrics backend.

output/cloud/v1/output.go
Blinkuu: >Even if we agreed that making this configurable isn't a priority, it's still better to keep this along with the other configuration (part of the ClientConfig).

I completely disagree with this one. I think abstracting away `context.Context` in functions that are indeed blocking is a bad design. Also, the pattern of having a "global default timeout" and cancellation semantics via `context.Context` is well established in Golang - you can take a look at `http.Client` for example.

output/cloud/v1/output.go
Blinkuu: Great catch, fixed in [25ec9fa](https://github.com/grafana/k6/pull/3202/commits/25ec9fa9489f1f319044bce7aeac8a94a5776ff0).

output/cloud/v1/output.go
olegbespalov: Well, I agree that abstracting away from the context is a bad idea, but basically, the proposal is based on the existing code where you use top-level `context.Background()`. Ideally, it should accept the context, and the signature for the `Connect` should look like `Connect(ctx context.Context)`.

Still, my main point was about encapsulating the client's timeout logic into the package code & consolidating the config. Since duplicating this in the v1 & v2 again outputs a pretty strong signal.

olegbespalov: >However, I see no value in doing this as there is no use case.

The value here is not having distributed configuration across code. You already have the structure that keeps the config, so why the other configuration thing should be separated?

Blinkuu: I think I'm lost.

> Well, I agree that abstracting away from the context is a bad idea, but basically, the proposal is based on the existing code where you use top-level context.Background().

Why would this be better than explicitly using a context with a set timeout? If anything, that's worse.

> Ideally, it should accept the context, and the signature for the Connect should look like Connect(ctx context.Context). That way it will respect any another timeout that will come with the context.

Yeah, but again - what problem does it solve? Setting up this in two places is relatively easy. A little copy is better than over-engineering (it will require adding tests, etc.). It's not going to be used in another place. It will soon be removed from the v1 output once it gets deprecated and removed altogether.

olegbespalov: > Yeah, but again - what problem does it solve? 

Again, it **consolidates** the config and makes things predictable and explicit. 

In your current case, you have to spend some time to learn that line
```
ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
```
is dialing timeout. So having instead of [magic number](https://en.wikipedia.org/wiki/Magic_number_(programming)) `10*time.Second` is something like `cfg.DialTimeout` fixes explicit issues.

The other is predictable. When I'm looking into the struct like https://github.com/grafana/k6/pull/3202/files#diff-fb2855cb8726c46cb031a9973fb066a19fc0b37cf60faf9a631cee8ab6a5997cR94-R117 I see a lot of configurations there, along with multiply boolean. Why, for instance, are they there?

>A little copy is better than over-engineering (it will require adding tests, etc.). I

I'm not following you on the `over-engineering`. Is adding a one-more field over-engineering? :thinking: Why do you think that it immediately requires the tests? Why then right now we have no tests?

oleiade: It seems to me the discussion drifted off a bit from its starting point to touch on what each of us value and consider bad or good engineering. From my perspective, both positions stand on their feet, and unfortunately, we're not going to solve it here and now (personally, I think both your positions make sense).

As it looks like we reached kind of a deadlock, I would offer that we base our decision on "practices" that the k6 codebase has been following so far to keep it consistent 🤝

A quick [GitHub search](https://github.com/search?q=repo%3Agrafana%2Fk6+context.WithTimeout&type=code) outlined to me that in most Go files (that are not testing files) using the `context.WithTimeout` function, we are setting the timeout from _some_ configuration as opposed to hardcoded values.

I suggest we follow that pattern for now and thus apply @olegbespalov's recommendation and that we take this discussion someplace else if we deem it worth discussing further 🙇🏻

Blinkuu: Done in [0844bcb](https://github.com/grafana/k6/pull/3202/commits/0844bcb429b091c230a4a14c6ad886148da71e7c).

olegbespalov (dismissed):

output/cloud/v1/output.go
oleiade: Thanks @Blinkuu, appreciate it 🙇🏻

oleiade (dismissed):

olegbespalov (approved):

oleiade (approved):

PR #3203, opened 2023-07-17
[fs] Add a `fs.opensync` function to the `k6/experimental/fs` module

### LINKED ISSUE #3143: Implement the File API's `openSync` method as a temporary helper/workaround for await operations in the init context
As part of #3141, implementing the [agreed-upon design for k6's File API](https://github.com/grafana/k6/blob/master/docs/design/019-file-api.md), we intend to implement a temporary `openSync` function for the File API module.

### Why

Using asynchronous code or the `await` keyword in the init context is currently impossible. For instance, the following is not valid within a k6 script and will throw an error:
```javascript 
await someAsyncFunc()

    // ...
}
```

We anticipate this capability will sooner or later land in k6, with the ESM (#2258) support. But in the meantime, that would make it harder for users to effectively use the `open` function in the init context.

### What

As a result, we intend to implement the `openSync` as a synchronous counterpart to the async `open` function #3142, to allow users to open files in the init context. As pointed out by @mstoykov, there is [an existing workaround in Javascript](https://github.com/grafana/k6/pull/3101/files#r1209915190), and we should probably have the function implementation use that workaround under the hood.

```
/*
 * openSync opens a file and returns an instance of a
 * `File`.
 */
openSync(path: string): File
```

### ⚠ Disclaimer

We expect this method to be removed when we support asynchronous code in the init context. Therefore, we should clearly, and consistently document that this method is a convenience for users and is expected to be deprecated in the future.

oleiade:
## What?

This Pull Request adds an `fs.openSync` function to the `k6/experimental/fs` module, as per #3143.

It allows to open a file as such:

```javascript
const f = fs.openSync('myfile.txt')
```

As opposed to the current workaround necessary to go around the limitation of not being able to use await in the init context:

```javascript
let file
(() => {
  await file = await fs.open('myfile.txt')
)();
```

## Why?

This function acts as a workaround for the lack of support for the `await` keyword in the init context and is meant to go away as soon as we do. It aims at ensuring a good UX that's more intuitive than the existing workaround ☝🏻 

We should document it as such and make it explicit that it won't be supported any further once we address the core issue and that the following syntax is possible:

```javascript
let f = await fs.open('myfile.txt')
```

## Checklist

## Related PR(s)/Issue(s)

closes #3143

### DIFF
--- a/examples/experimental/fs/opensync.js
+++ b/examples/experimental/fs/opensync.js
@@ -1,0 +1,17 @@
+import { openSync } from "k6/experimental/fs";
+
+export const options = {
+	vus: 100,
+	iterations: 1000,
+};
+
+// As k6 does not support asynchronous code in the init context, yet, we need to
+// use a top-level async function to be able to use the `await` keyword.
+const file = openSync("bonjour.txt")
+
+export default async function () {
+	const fileinfo = await file.stat();
+	if (fileinfo.name != "bonjour.txt") {
+		throw new Error("Unexpected file name");
+	}
+}
--- a/js/modules/k6/experimental/fs/module.go
+++ b/js/modules/k6/experimental/fs/module.go
@@ -50,7 +50,8 @@ func (rm *RootModule) NewModuleInstance(vu modules.VU) modules.Instance {
 func (mi *ModuleInstance) Exports() modules.Exports {
 	return modules.Exports{
 		Named: map[string]any{
-			"open": mi.Open,
+			"open":     mi.Open,
+			"openSync": mi.OpenSync,
 		},
 	}
 }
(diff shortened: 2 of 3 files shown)


codebien (approved):
>We should document it as such and make it explicit that it won't be supported any further once we address the core issue and that the following syntax is possible

I'm not very happy with it. We are adding something that in a cycle could be removed.

If it is useful a lot of people will use it then removing it will make them a lot unhappy.
If it isn't useful then we could just not add.

I would wait for one more cycle before adding it to see the evolution around JS ES modules.

olegbespalov (approved):
I'm sharing concerns with @codebien regarding something adding something temporary.

On the other side, it's an experimental module, and by our definition, the API isn't stable, so it's OK to have this for a few releases.

oleiade:
There seems to be a consensus that we should not add `openSync` if we know we won't need it in the foreseeable future.

I'm happy to oblige and will close this Pull Request as a result 🙇🏻 

I would consider support for `await` in the init stage a must-have for taking the `fs` module out of experimental though, and will add that condition to the issue 👍🏻

PR #3206, opened 2023-07-18
Implement  concurrent pushes across batches

### LINKED ISSUE #3192: Concurrent flush across batches
### What

Move the concurrency for flushing metrics from [per-flush](https://github.com/grafana/k6/blob/f86938d7c7a4c9e19487fb54b8cd475c81366976/output/cloud/expv2/output.go#L199-L222) to [per-batch](https://github.com/grafana/k6/blob/f86938d7c7a4c9e19487fb54b8cd475c81366976/output/cloud/expv2/flush.go#L69-L85).

The expected architecture is one goroutine doing the following operations:
- Fetch the buckets from the buckets queue
- Split time series into batches
  - Encoding as protobuf
  - Enqueue the batch as a job to be pushed to the remote service

And a series of concurrent goroutines doing the following operations:
- Fetch a job
- Invoke the `metricsClient.push` operation

### Why

We have seen not optimal handling when we hit tests with lot of _active time series_ (> 100k). The flush operation will split them in batches and then pushes them sequentially, doing some math like the following, it is to see why we could hit some >10s per single flush operation.

##### Example

100k time series
1k time series as batch limit

that generates 100 batches

in the case, we don't have perfect networking (e.g 100ms per request) then we will end with a total of 10 seconds for flushing a single iteration of 100k active series (100 batches * 100 ms), and it can even grow with worst cases.

olegbespalov:
## What?

This change moves the concurrency from flushes to batches. That fixes issue #3192 and generally improves processing.

## Why?

Even if the single network request to ingester is fast enough, executing them sequentially increases the time.

## Checklist

## Related PR(s)/Issue(s)

Closes #3192

### DIFF
--- a/output/cloud/expv2/flush.go
+++ b/output/cloud/expv2/flush.go
@@ -94,42 +94,36 @@ func (f *metricsFlusher) flushBatches(batches []*pbcloud.MetricSet) error {
 	}
 
 	var (
-		wg   = sync.WaitGroup{}
-		errs = make(chan error)
-		done = make(chan struct{})
-		stop = make(chan struct{})
-		feed = make(chan *pbcloud.MetricSet)
-
+		wg      = sync.WaitGroup{}
 		workers = min(len(batches), f.batchPushConcurrency)
+		errs    = make(chan error, workers)
+		feed    = make(chan *pbcloud.MetricSet)
+		done    = make(chan struct{})
 	)
 
 	wg.Add(workers)
-	for workersIndex := 0; workersIndex < workers; workersIndex++ {
+	for i := 0; i < workers; i++ {
 		go func() {
 			defer wg.Done()
 
 			for chunk := range feed {
-				select {
-				case <-stop:
-					return
-				default:
-				}
 				if err := f.client.push(chunk); err != nil {
-					select {
-					case errs <- err:
-					case <-stop:
-						return
-					}
+					errs <- err
+					return
 				}
 			}
 		}()
 	}
+	go func() {
+		defer close(done)
+		wg.Wait()
+	}()
 
 	for i := 0; i < len(batches); i++ {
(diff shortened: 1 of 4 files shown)


output/cloud/expv2/flush.go
```diff
@@ -43,60 +46,110 @@ func (f *metricsFlusher) flush() error {
...
+		workers   = int(math.Min(float64(len(batches)), float64(f.batchPushConcurrency)))
+		chunkSize = len(batches) / workers
+	)
+
+	wg.Add(workers)
+	for workersIndex := 0; workersIndex < workers; workersIndex++ {
+		offset := (workersIndex * chunkSize)
+		end := offset + chunkSize
+		if workersIndex == workers-1 {
+			end = len(batches)
```
mstoykov, line 104: I would not do it like that ... and I m not certain this math works out perfectly :grimacing: 

I would've just given each worker a channel and written to it from the main goroutine adding all the pieces and let the workers get them on at a time.

This way they will balance out if one piece/worker takes longer and not hard math needed.

mstoykov: Also at least in theory this can be done while the pieces are being broken down.

mstoykov (commented):
LGTM, but I would prefer if we can do some changes to some of the math ... mostly removing some.

oleiade (commented):
LGTM too 🤝 

Non-blocking on my part, but I share the same concern regarding the work dispatching to multiple workers as @mstoykov. I think the code could be simplified indeed, to have one input channel, and multiple workers reading from it concurrently, I find it might also make the logic easier to follow.

oleiade (approved):
🚀 🚀 🚀

mstoykov (approved):
LGTM !

Issue #2951, opened 2023-03-01
`ramping-arrival-rate` executor shows wrong progressbar 


na--:
### Brief summary

The [`ramping-arrival-rate` scenario executor](https://k6.io/docs/using-k6/scenarios/executors/ramping-arrival-rate/) doesn't handle situations in which the iteration rate doesn't match the duration exactly and iterations finish up very quickly. The progress bar appears unfinished and has a red cross (✗), suggesting that there was a problem, when in fact the executor finished normally.

### k6 version

any since k6 v0.27.0, i.e. this has been a problem since the initial introduction of the `ramping-arrival-rate` executor :sweat_smile:

### OS

Any

### Steps to reproduce the problem

Have a script like this, where the iterations finish quickly and the scenario duration doesn't match the `timeUnit * iteration rate` exactly.
```js
export const options = {
    scenarios: {
        demo: {
            executor: 'ramping-arrival-rate',
            startRate: 1,
            preAllocatedVUs: 5,
            stages: [
                { duration: '2.5s', target: 1 },
            ],
        },
    },
};

export default function () {
    console.log(`iter`);
}

export function handleSummary(data) { }
```

### Expected behaviour

Expect the test run above to finish normally after running exactly 2 iterations.

### Actual behaviour

The test runs exactly 2 iterations, but the progress bar makes it seem like the scenario didn't fully finish or there and was instead prematurely interrupted:
```
...
INFO[0001] iter                                          source=console
INFO[0002] iter                                          source=console

running (02.0s), 0/1 VUs, 2 complete and 0 interrupted iterations
demo ✗ [=============================>--------] 0/1 VUs  2.0s/2.5s  1.00 iters/s
...
```

Basically, it seems like the executor just aborts early, since it has finished all of the 2 iterations it was supposed to execute and it "knows" it's not going to have time to start that third iteration. That is normal and good and desired. However, it doesn't seem to update its progressbar correctly :disappointed: 

One way to "fix" the progress bars is to force the executor it to wait, i.e. slow down the iterations. If I just add `sleep(2)` in the iteration, and then the progress bar is displayed correctly. But so far this seems like just a visualization issue, the execution seems to be working as expected :relieved:

PR #3220, opened 2023-07-21
Addressed an issue where the ramping-arrival-rate executor was showing the wrong progress bar

Shimizu1111:
## What?
This PR attempts to address https://github.com/grafana/k6/issues/2951.

## Why?

The problem was caused by the fact that when the number of seconds in `duration` was a decimal point, the number of loops was also a decimal point.

## Checklist


lib/executor/ramping_arrival_rate.go
```diff
@@ -263,7 +263,7 @@ func (varc RampingArrivalRateConfig) cal(et *lib.ExecutionTuple, ch chan<- time.
 			}
 		} else {
 			endCount += dur * to
-			for ; i <= endCount; i += float64(next()) {
+			for ; i <= math.Ceil(endCount); i += float64(next()) {
```
mstoykov, line 266: I feel like this should be the fix. 

The reported bug is mostly a visual one, where we tell an executor did not finish, but it ... did.

This is confirmed by the test time below changing, which should not happen. 

Maybe I am missing some very small detail which makes the below times previously wrong. But the current numbers seem to progress with the correct steps between them while with your change you move them around a bit.

IMO the fix should be in the thing reporting to the progressbar or the progressbar itself to not think of this as an interruption

mstoykov (commented):
Sorry for the slow reply :bow: A lot of people are on vacation and this fell through :( 

Thanks for the PR :bow: 

Unfortunately I did take a look at it and this doesn't seem like it is fixing the correct thing. Instead, it is changing how we calculate the steps, which might be needed :shrug: . 
 
I definitely do not want to change this code so close to v0.46.0 release date especially for a visual bug. 

So I moved it to v0.47.0. Can you please also add an integration test for this.

You can use the script from the issue and do something like https://github.com/grafana/k6/blob/6c310759e78aca046b85a59083e99b94cc0fc004/cmd/tests/cmd_run_test.go#L629-L669

mstoykov:
@Shimizu1111 are you still interested on working on this? If not we will likely close this as inactive at this point

PR #3221, opened 2023-07-21
Remove `K6_BROWSER_ENABLED` flag requirement

### LINKED ISSUE #3197: `K6_BROWSER_ENABLED` flag is redundant with browser type scenario option
Currently, as of commit [a8ebe8c](https://github.com/grafana/xk6-browser/tree/a8ebe8cc24f639e0553b1d969df97368db7b5ec5), the `K6_BROWSER_ENABLED` environment variable is redundant with the "browser type" option defined in the scenarios specification, as both of these are required in order to run a test that uses the browser module, being the latter one more restrictive as it must be set per scenario.
This is due to the changes applied in https://github.com/grafana/k6/pull/3036 to support browser options inside scenarios definition in k6 core, and the modifications done in https://github.com/grafana/xk6-browser/pull/944 in order to automatically initialize browsers based on this option.

From the k6 browser team we think that the usage of this flag is no longer required, and removing it will provide a better UX.

ka3de:
## What?

Removes the requirement for the `K6_BROWSER_ENABLED` environment variable in order to execute tests that use the experimental browser module.

## Why?

Currently (as of grafana/xk6-browser@a8ebe8c) the browser extension requires the definition of the browser type parameter inside the options element for every scenario that wants to use the browser module. Therefore the usage of `K6_BROWSER_ENABLED` flag as an explicit approval to use the browser module has become redundant, as both parameters have to be set, being the scenario one more restrictive.

## Checklist

## Related PR(s)/Issue(s)

Closes #3197


ankur22 (commented):
Thanks for this 👍 

I think we can remove the whole of the `browser/rootmodule.go` directory and file. We should be able to directly work with the root module that we define in the browser module -- replace `"github.com/grafana/xk6-browser/browser"` in [jsmodules.go](https://github.com/grafana/k6/blob/2390e225191509493b0e2d485df5a9a1cc838641/js/jsmodules.go#L11).

ka3de:
> We should be able to directly work with the root module that we define in the browser module -- replace "github.com/grafana/xk6-browser/browser" in [jsmodules.go](https://github.com/grafana/k6/blob/2390e225191509493b0e2d485df5a9a1cc838641/js/jsmodules.go#L11).

Right, thanks for the heads up, I think that makes sense, as I see there are other examples of using the external implementation directly. Let me address this.

mstoykov:
Shouldn't this be merged *after* we update xk6-browser in k6?

ka3de:
> Shouldn't this be merged after we update xk6-browser in k6?

Right, I was assuming we will upgrade the current _xk6-browser_ version, which release is planned for this Friday, at some point before the next _k6_ release. But yes, in practice we can wait for this to happen before merging this PR.

ka3de:
Closing as superseded by https://github.com/grafana/k6/pull/3235.

Issue #3222, opened 2023-07-21
GRPC reflection will crash when there is no package specifier defined


tmulkern:
### Brief summary

If the proto files used in a GRPC omits the package specifier and the k6 GRPC client uses reflection, it will result in a error

`ERRO[0002] panic: runtime error: invalid memory address or nil pointer dereference`

The package specifier is *optional* as per the proto [spec](https://protobuf.dev/programming-guides/proto3/#packages) so the reflection should not result in an error

### k6 version

0.45

### OS

All

### Steps to reproduce the problem

1. Use the following proto specfication (please take note that there is no package specifier)

```proto
// based on https://grpc.io/docs/guides/concepts.html

syntax = "proto2";

service HelloService {
  rpc SayHello(HelloRequest) returns (HelloResponse);
  rpc LotsOfReplies(HelloRequest) returns (stream HelloResponse);
  rpc LotsOfGreetings(stream HelloRequest) returns (HelloResponse);
  rpc BidiHello(stream HelloRequest) returns (stream HelloResponse);
}

message HelloRequest {
  optional string greeting = 1;
}

message HelloResponse {
  required string reply = 1;
}
```

2. Use this proto to host a GRPC service
3. Connect to the GRPC service using the GRPC client in K6 with relection enabled

### Expected behaviour

No Error

### Actual behaviour

Error produced

```
ERRO[0002] panic: runtime error: invalid memory address or nil pointer dereference
goroutine 108 [running]:
runtime/debug.Stack()
        runtime/debug/stack.go:24 +0x65
go.k6.io/k6/js/common.RunWithPanicCatching.func1()
        go.k6.io/k6@v0.45.0/js/common/util.go:82 +0x1f3
panic({0x1976e80, 0x2d12a50})
        runtime/panic.go:884 +0x213
github.com/dop251/goja.(*Runtime).runWrapped.func1()
        github.com/dop251/goja@v0.0.0-20230531210528-d7324b2d74f7/runtime.go:2516 +0x185
panic({0x1976e80, 0x2d12a50})
        runtime/panic.go:884 +0x213
github.com/dop251/goja.(*vm).handleThrow(0xc000c9c480, {0x1976e80, 0x2d12a50})
        github.com/dop251/goja@v0.0.0-20230531210528-d7324b2d74f7/vm.go:788 +0x497
github.com/dop251/goja.(*vm).try.func1()
[...]

PR #3223, opened 2023-07-21
Added Nil Check for Package in File Descriptor

tmulkern:
## What?

This code is to guard against a `runtime error: invalid memory address or nil pointer dereference` occuring

## Why?

The package specifier is optional in protobuf spec https://protobuf.dev/programming-guides/proto3/#packages

## Checklist

- [ ] I have added tests for my changes. N/A
- [x] I have run linter locally (`make ci-like-lint`) and all checks pass. N/A
- [x] I have run tests locally (`make tests`) and all tests pass. N/A
- [ ] I have commented on my code, particularly in hard-to-understand areas. N/A

## Related PR(s)/Issue(s)

Fixes issue #3222

### DIFF
--- a/lib/netext/grpcext/conn_test.go
+++ b/lib/netext/grpcext/conn_test.go
@@ -189,6 +189,28 @@ func TestResolveFileDescriptors(t *testing.T) {
 			services:            []string{},
 			expectedDescriptors: 0,
 		},
+		{
+			name:                "NoPackage",
+			services:            []string{"Service1", "Service2"},
+			expectedDescriptors: 2,
+		},
+		{
+			name:                "NoPackageDeduplicateServices",
+			services:            []string{"Service1", "Service2", "Service1"},
+			expectedDescriptors: 2,
+		},
+		{
+			name:                "MixPackage",
+			pkgs:                []string{"mypkg1"},
+			services:            []string{"Service1", "Service2"},
+			expectedDescriptors: 2,
+		},
+		{
+			name:                "MixPackageDeduplicateServices",
+			pkgs:                []string{"mypkg1"},
+			services:            []string{"Service1", "Service2", "Service1"},
+			expectedDescriptors: 2,
+		},
 	}
 
 	for _, tt := range tests {
@@ -199,10 +221,16 @@ func TestResolveFileDescriptors(t *testing.T) {
 				lsr  = &reflectpb.ListServiceResponse{}
 				mock = &getServiceFileDescriptorMock{}
 			)
+
+			var pkg string
+			if len(tt.pkgs) == 1 {
+				pkg = tt.pkgs[0]
+			}
+
 			for i, service := range tt.services {
 				// if only one package is defined then
 				// the package is the same for every service
-				pkg := tt.pkgs[0]
+
 				if len(tt.pkgs) > 1 {
 					pkg = tt.pkgs[i]
 				}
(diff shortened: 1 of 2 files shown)


mstoykov (approved):
LGTM!

Sorry for the slow review :bow: . 

Due to vacations across the team and the automatic assigning not taking this into account, sometimes things get assigned to people who are on vacation :(.

Unfortunately we are also at the end of the development cycle and have some higher priorities things we want to make certain do work for v0.46.0.

The fix seems pretty safe and I can't see any problems with it, but I still prefer if @olegbespalov gives it a :+1: specifically. This might mean that we do merge this in the codefreeze, but given the size and the fact it is a bug that doesn't seem like that much of a problem.

@tmulkern Can you please add a small test that panics without this change though.

Also we will need forward/backport this to [xk6-grpc](https://github.com/grafana/xk6-grpc/). 

I will move this back to v0.46.0 milestone for now, but I can't promise we will actually make it :(

Thanks for the bug report and fix :bow: And once again sorry for the slow reply

tmulkern:
@mstoykov no problem on the response, I am in the same situation myself in my day job, it is the time for Vacations!!

tmulkern:
Regarding the test, no problem in creating, but do we have a GRPC Test Service/Mock Service that I can reference. I couldn't find any. I am not *that* familiar with the code base, so I could be blind to it ;-)

codebien (changes requested):
Hey @tmulkern,
thanks for your contribution. :bow: 

>Regarding the test, no problem in creating, but do we have a GRPC Test Service/Mock Service that I can reference

I guess it should be enough to add a new test case to this test table. I think we should cover also the case with clashing names now that we are skipping the package.

https://github.com/grafana/k6/blob/3c006a5e7b4a44adbc85aefc1933759d22b99286/lib/netext/grpcext/conn_test.go#L160

tmulkern:
Hey @codebien 

Added tests for following scenarios

* Services with no package - "NoPackage"
* Services with no package and duplicated - "NoPackageDeduplicateServices"
* One services with a package, one without - "MixPackage"
* Two services with a package (Duplicate), one without - "MixPackageDeduplicateServices"

Also updated the TestResolveFileDescriptors to handle cases with no packages

codebien (changes requested):

lib/netext/grpcext/conn_test.go
```diff
@@ -199,10 +221,16 @@ func TestResolveFileDescriptors(t *testing.T) {
...
 			)
+
+			var pkg string
+			if len(tt.pkgs) == 1 {
+				pkg = tt.pkgs[0]
+			}
+
 			for i, service := range tt.services {
 				// if only one package is defined then
 				// the package is the same for every service
```
olegbespalov, line 232:
```suggestion
```

olegbespalov (approved):
To be honest, in that particular case, I'm more up for keeping the logic (with the comment), like

```golang
pkg := ""

// if only one package is defined then
// the package is the same for every service
if len(tt.pkgs) == 1 {
   pkg = tt.pkgs[0]
}
```

The rationale is that having the empty strings confuses me more than the logic of the comment.

tmulkern:
Applogies people for the late reply, as mstoykov alluded to it is vacation season and I was away on vacation

olegbespalov:
Hi @tmulkern !

I am sharing this here for the sake of transparency. We have another PR related to the [support of gRPC's reflection v1](https://github.com/grafana/k6/pull/3338). 

I believe that PR also fixes the issue that you reported. So we will probably close this PR in favor of https://github.com/grafana/k6/pull/3338.

Anyway, thank you for reporting and contribution! :bow:

tmulkern:
Hey [olegbespalov](https://github.com/olegbespalov), thank's for the update. Sorry, life got in the way the 2 months and I hadn't got round to finishing this PR.

Issue #3216, opened 2023-07-20
Add Authentication information for Loki logs tail+push


federicotdn:
### Feature Description

Add additional authentication information for Cloudlogs-related requests done to k6 Cloud. This will enable users to also upload logs when running k6 with `-o cloud`.

PR #3224, opened 2023-07-24
Add new auth headers required for tailing k6 Cloud logs

federicotdn:
## What?

Adds two new HTTP headers required to tail logs from k6 Cloud in the near future.

## Why?

See issue https://github.com/grafana/k6/issues/3216 and backlinked issue from the Grafana private repository.

## Checklist

## Related PR(s)/Issue(s)

https://github.com/grafana/k6/issues/3216

### DIFF
--- a/cloudapi/logs.go
+++ b/cloudapi/logs.go
@@ -100,6 +100,8 @@ func (c *Config) logtailConn(ctx context.Context, referenceID string, since time
 
 	headers := make(http.Header)
 	headers.Add("Sec-WebSocket-Protocol", "token="+c.Token.String)
+	headers.Add("Authorization", "token "+c.Token.String)
+	headers.Add("X-K6TestRun-Id", referenceID)
 
 	var conn *websocket.Conn
 	err = retry(sleeperFunc(time.Sleep), 3, 5*time.Second, 2*time.Minute, func() (err error) {


cloudapi/logs.go
```diff
@@ -100,6 +100,8 @@ func (c *Config) logtailConn(ctx context.Context, referenceID string, since time
 
 	headers := make(http.Header)
 	headers.Add("Sec-WebSocket-Protocol", "token="+c.Token.String)
+	headers.Add("Authorization", "token "+c.Token.String)
```
mstoykov, line 103: This has the same token as the above? 

Can't we just keep the old one?

cloudapi/logs.go
```diff
@@ -100,6 +100,8 @@ func (c *Config) logtailConn(ctx context.Context, referenceID string, since time
...
+	headers.Add("X-K6TestRun-Id", referenceID)
```
mstoykov, line 104: Does it mean that if currently I run two test I will get *both* of their logs?

cloudapi/logs.go
mstoykov: Does that mean that this will stop working with just `Sec-WebSocket-Protocol` ?

cloudapi/logs.go
federicotdn: Nope! The `test_run_id` label is already making sure that doesn't happen. Without going to much into details, the new headers added are part of the more generic auth layer. In the end though, all information received is used to authenticate+authorize.

cloudapi/logs.go
federicotdn: I will extend the auth layer specifically for Cloudlogs to support old clients as well, until we are OK with removing support for them (whenever that happens).

mstoykov: Shall we then drop the other ones if they will no longer be used?

federicotdn: We could remove `Sec-WebSocket-Protocol` now, but to make the transition easier it is better if we leave it in. If we send all three headers, then k6 will be compatible with both old and new auth layers.

mstoykov (approved):

oleiade (approved):

PR #3231, opened 2023-07-27
Fix tracing module sampling option should default to 1.0 when not set

oleiade:
## What?

This Pull Request ensures that if the `sampling` option of the tracing module's client is not provided by the user, the Client defaults to sample 100% of the request.

```javascript
		import http from "k6/http";

		// We do not set the sampling option, thus the default
		// behavior should be to always sample.
		const instrumentedHTTP = new tracing.Client({
			propagator: "w3c",
		})
```

## Why?

As noted by @Blinkuu, the current behavior of the tracing module client is to use a 0% sampling when the option is not set. It leads to effectively no HTTP request being sampled. Whereas the desired behavior would be to have all requests sampled unless instructed otherwise.

## Checklist

### DIFF
--- a/cmd/tests/tracing_module_test.go
+++ b/cmd/tests/tracing_module_test.go
@@ -154,6 +154,58 @@ func TestTracingModuleClient_HundredPercentSampling(t *testing.T) {
 	assertHasTraceIDMetadata(t, jsonResults, 100, tb.Replacer.Replace("HTTPBIN_IP_URL/tracing"))
 }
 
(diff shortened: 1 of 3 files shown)


mstoykov (approved):

Blinkuu (approved):
LGTM, thanks!

oleiade:
It turns out this was caused by a difference in how options are handled between the client and the options. I addressed the issue differently and ensured all the tests passed now.

mstoykov (approved):

codebien (approved):

Issue #2579, opened 2022-06-28
Warn users on unsupported parts of `setup` returned data


mstoykov:
### Feature Description

Currently `setup` can return whatever and it will be send to `default` and `teardown`. 

A lot of users are confused what actually happens because for the most part it works perfectly and as long as you:
1. don't put unsupported types in the data
2. don't expect that edititng it will be observable between VUs and stages
you can 100% think that it is just that exact same object.

But it isn't it is a copy made by marshalling and then unmarshalling from JSON. This definitely makes it so
1. If you have added a function it will be `nil` ... which is very confusing. Doubly so [if it's a method on a class](https://community.k6.io/t/exported-method-from-a-class-is-not-available-in-the-test-teardown-function/4035) and the instance of that class is till there.
2. You try to edit it and expect that `teardown` will see the changes. 

Both of those are *kind of* [documented](https://k6.io/docs/using-k6/test-life-cycle/#using-data-from-setup-in-default-and-init) but I would expect that a lot of people wouldn't find them. 

### Suggested Solution (optional)

Write some, hopefully not very complicated code, to go through the whole object and check for known problems such as `functions` and print a `warning` about it. 

We can probably try to write the name of the property or even the "path" in it as that will improve the message a lot.

For the second one we discuss it a lot in https://github.com/grafana/k6/issues/2043, but I wonder if we can just `freeze` that object after it's unmarshalled :shrug:  

### Already existing or connected issues / PRs (optional)

https://github.com/grafana/k6/issues/2043

Aksh-Bansal-dev:
I would like to work on this issue. I'm new to the codebase but with some help, I will be able to make a PR.

Aksh-Bansal-dev:
Currently, we get this error if we return a function as setupData.

![Screenshot from 2022-09-30 17-27-57](https://user-images.githubusercontent.com/63552235/193264896-7c3d38c4-0847-48df-a6cc-5d507e0f5b67.png)

Should we print some different error/warning in this case(functions as return value)?

na--:
This is the place of the code where we marshal the data returned from `setup()`: https://github.com/grafana/k6/blob/33056563520820fee98c70458cef7a485757508e/js/runner.go#L289

[...]

PR #3234, opened 2023-07-28
Support binary passed in GRPC metadata

sapphire-janrain:
Use Uint8Array type in metadata declaration to avoid any reencoding issues in goja.

## What?

Changes k6's GRPC client to accept Uint8Array types passed as metadata values.

## Why?

Currently k6 does not support adding binary data in the metadata. The suggestion seems to have been to convert it to a string of some sort but there are issues therein:

* Attempting to just cast bytes into a string via String.fromCharCode or String.fromCodePoint seems to have issues on the goja side. Whenever a byte is >= 0x80 (presumably) another byte is inserted before it by the time it makes it to k6 and protobuf.
* We cannot base64 the data on the JS side because protobuf will re-encode any `-bin` metadata fields itself.

As such, it would be best for k6 to properly accept Uint8Arrays in metadata.

## Checklist

## Caveats

You cannot store Uint8Array in the returned data object from setup (possibly related to #2579 ) as it's converted to a base64'd string before it's passed to the scenario. I'm not sure why exactly and didn't investigate, I assume that's expected behavior (though it was surprising to me). I'm not inclined to fix it here as it's simple to work around for my use case, but happy to create a ticket if it does happen to be unexpected.

### DIFF
--- a/js/modules/k6/grpc/client.go
+++ b/js/modules/k6/grpc/client.go
@@ -461,7 +461,11 @@ func (c *Client) parseInvokeParams(paramsVal goja.Value) (*invokeParams, error) 
 				// TODO(rogchap): Should we manage a string slice?
 				strval, ok := kv.(string)
 				if !ok {
-					return result, fmt.Errorf("metadata %q value must be a string", hk)
+					binval, ok := kv.([]byte)
+					if !ok {
+						return result, fmt.Errorf("metadata %q value must be a string or []byte", hk)
+					}
+					strval = string(binval)
 				}
 				result.Metadata[hk] = strval
 			}
--- a/js/modules/k6/grpc/client.go
+++ b/js/modules/k6/grpc/client.go
@@ -461,6 +461,8 @@ func (c *Client) parseInvokeParams(paramsVal goja.Value) (*invokeParams, error) 
 				// TODO(rogchap): Should we manage a string slice?
 				strval, ok := kv.(string)
 				if !ok {
+					// The spec defines that Binary-valued keys end in -bin
+					//  https://grpc.io/docs/what-is-grpc/core-concepts/#metadata
 					if strings.HasSuffix(hk, "-bin") {
 						binval, ok := kv.([]byte)
 						if !ok {
--- a/js/modules/k6/grpc/client.go
+++ b/js/modules/k6/grpc/client.go
@@ -463,7 +463,8 @@ func (c *Client) parseInvokeParams(paramsVal goja.Value) (*invokeParams, error) 
 				//  https://grpc.io/docs/what-is-grpc/core-concepts/#metadata
 				var strval string
 				if strings.HasSuffix(hk, "-bin") {
-					binval, ok := kv.([]byte)
+					var binval []byte
+					binval, ok = kv.([]byte)
 					if !ok {
 						return result, fmt.Errorf("metadata %q value must be binary", hk)
 					}
--- a/js/modules/k6/grpc/client.go
+++ b/js/modules/k6/grpc/client.go
@@ -468,6 +468,7 @@ func (c *Client) parseInvokeParams(paramsVal goja.Value) (*invokeParams, error) 
 					if !ok {
 						return result, fmt.Errorf("metadata %q value must be binary", hk)
 					}
+					// https://github.com/grpc/grpc-go/blob/v1.57.0/Documentation/grpc-metadata.md#storing-binary-data-in-metadata
 					strval = string(binval)
 				} else if strval, ok = kv.(string); !ok {
 					return result, fmt.Errorf("metadata %q value must be a string", hk)
--- a/js/modules/k6/grpc/client.go
+++ b/js/modules/k6/grpc/client.go
@@ -459,19 +459,17 @@ func (c *Client) parseInvokeParams(paramsVal goja.Value) (*invokeParams, error) 
 			}
 			for hk, kv := range rawHeaders {
 				// TODO(rogchap): Should we manage a string slice?
-				strval, ok := kv.(string)
-				if !ok {
-					// The spec defines that Binary-valued keys end in -bin
-					//  https://grpc.io/docs/what-is-grpc/core-concepts/#metadata
-					if strings.HasSuffix(hk, "-bin") {
-						binval, ok := kv.([]byte)
-						if !ok {
-							return result, fmt.Errorf("metadata %q value must be a string or binary", hk)
-						}
-						strval = string(binval)
-					} else {
-						return result, fmt.Errorf("metadata %q value must be a string", hk)
+				// The spec defines that Binary-valued keys end in -bin
+				//  https://grpc.io/docs/what-is-grpc/core-concepts/#metadata
+				var strval string
+				if strings.HasSuffix(hk, "-bin") {
+					binval, ok := kv.([]byte)
+					if !ok {
+						return result, fmt.Errorf("metadata %q value must be binary", hk)
 					}
+					strval = string(binval)
+				} else if strval, ok = kv.(string); !ok {
+					return result, fmt.Errorf("metadata %q value must be a string", hk)
 				}
 				result.Metadata[hk] = strval
 			}
--- a/js/modules/k6/grpc/client.go
+++ b/js/modules/k6/grpc/client.go
@@ -461,11 +461,15 @@ func (c *Client) parseInvokeParams(paramsVal goja.Value) (*invokeParams, error) 
 				// TODO(rogchap): Should we manage a string slice?
 				strval, ok := kv.(string)
 				if !ok {
-					binval, ok := kv.([]byte)
-					if !ok {
-						return result, fmt.Errorf("metadata %q value must be a string or []byte", hk)
+					if strings.HasSuffix(hk, "-bin") {
+						binval, ok := kv.([]byte)
+						if !ok {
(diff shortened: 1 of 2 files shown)


js/modules/k6/grpc/client.go
```diff
@@ -461,7 +461,11 @@ func (c *Client) parseInvokeParams(paramsVal goja.Value) (*invokeParams, error)
 				// TODO(rogchap): Should we manage a string slice?
 				strval, ok := kv.(string)
 				if !ok {
-					return result, fmt.Errorf("metadata %q value must be a string", hk)
+					binval, ok := kv.([]byte)
```
olegbespalov, line 464: Checking the https://grpc.io/docs/what-is-grpc/core-concepts/#metadata, it seems like we should only treat the value as binary only for the cases, where keys have a postfix `-bin`.

What do you think about considering this?

js/modules/k6/grpc/client.go
```diff
@@ -461,7 +461,11 @@ func (c *Client) parseInvokeParams(paramsVal goja.Value) (*invokeParams, error)
...
+					if !ok {
+						return result, fmt.Errorf("metadata %q value must be a string or []byte", hk)
```
olegbespalov, line 466: It seems like this is a user-faced error, so it's better to explicitly refer to the `Uint8Array` instead of the golang's (`[]byte`)

olegbespalov:
@sapphire-janrain Thanks for the contribution :bow: left a few comments

js/modules/k6/grpc/client.go
```diff
@@ -461,7 +461,15 @@ func (c *Client) parseInvokeParams(paramsVal goja.Value) (*invokeParams, error)
 				// TODO(rogchap): Should we manage a string slice?
 				strval, ok := kv.(string)
 				if !ok {
-					return result, fmt.Errorf("metadata %q value must be a string", hk)
+					if strings.HasSuffix(hk, "-bin") {
```
codebien, line 464:
```suggestion
                    // The spec defines that Binary-valued keys end in -bin
                    //  https://grpc.io/docs/what-is-grpc/core-concepts/#metadata
					if strings.HasSuffix(hk, "-bin") {
```


I would prefer if we mention directly the spec in a comment so for the next contributor is clear that k6 is just following the spec and we haven't applied an arbitrary decision.

codebien (changes requested):
Hey @sapphire-janrain, thanks for your contribution :bow:

LGTM, just a minor comment. Note: You have to rebase the branch on top of `master` for fixing the `xk6` steps of the CI.

codebien (approved):
LGTM, thanks @sapphire-janrain!

js/modules/k6/grpc/client.go
```diff
@@ -461,7 +461,17 @@ func (c *Client) parseInvokeParams(paramsVal goja.Value) (*invokeParams, error)
...
+					//  https://grpc.io/docs/what-is-grpc/core-concepts/#metadata
+					if strings.HasSuffix(hk, "-bin") {
+						binval, ok := kv.([]byte)
+						if !ok {
+							return result, fmt.Errorf("metadata %q value must be a string or binary", hk)
+						}
+						strval = string(binval)
+					} else {
+						return result, fmt.Errorf("metadata %q value must be a string", hk)
+					}
```
mstoykov, line 474: A passing comment without much knowledge about gRPC:
1. I find it strange that we do not check the suffix *before* we try to make it into a string.
2. I find it strange that after we get something that *isn't* as string as per it's definition we just cast it into a string.

js/modules/k6/grpc/client.go
```diff
@@ -461,7 +461,17 @@ func (c *Client) parseInvokeParams(paramsVal goja.Value) (*invokeParams, error)
 				// TODO(rogchap): Should we manage a string slice?
 				strval, ok := kv.(string)
 				if !ok {
-					return result, fmt.Errorf("metadata %q value must be a string", hk)
+					// The spec defines that Binary-valued keys end in -bin
+					//  https://grpc.io/docs/what-is-grpc/core-concepts/#metadata
+					if strings.HasSuffix(hk, "-bin") {
```
olegbespalov, line 466: @sapphire-janrain, thanks for the change. Generally looks good :+1: 

However, I think we might have an issue with the logical levels. We have the binary-dependent logic, under the `"string"` case. 

How about moving the change outer of the current if/else block, which belongs to the non-binary keys? I believe it also should address the TODO `// TODO(rogchap): Should we manage a string slice?`

In a pseudo-code, I meant modifying the body of the for `hk, kv := range rawHeaders`'s loop like:
```
for key, value {
   if key is binary {
      // try to cast value to a binary and store strval
      // if can't cast binary return an error
   } else {
      // try to cast to a string and store it
      // if can't cast to a string, return an error
   }

   // put the string representation into map
}
```

In that way, it has fewer nested layers, and we could make the error message even more precise since we know that the specific key is either binary or string.

How does that all sound to you?

js/modules/k6/grpc/client.go
codebien: >I find it strange that we do not check the suffix before we try to make it into a string.

I agree, good suggestion

codebien (changes requested):
I agree w @mstoykov suggestion, so marking as `Request changes` again.

js/modules/k6/grpc/client.go
mstoykov: This is part `1` - yes

but `2` is more about the fact we still actually make the binary metadata into a string one - which likely will lose "preceision".

olegbespalov: :thinking:  right, `2` sounds like a valid concern

js/modules/k6/grpc/client.go
sapphire-janrain: > I find it strange that we do not check the suffix before we try to make it into a string.
My main motivation for this is that more often than not it's going to be a string, so I was avoiding this change causing any impact. Happy to change if that's preferred.

> I find it strange that after we get something that isn't as string as per it's definition we just cast it into a string.
This is all the GRPC go library does too. If you're expecting some kind of encoding that wouldn't be desirable. It's just an unfortunate side-effect of the GRPC library having no means of accepting binary data directly.

js/modules/k6/grpc/client.go
sapphire-janrain: I don't think we should *require* `-bin` keys to be binary as that's ultimately a breaking change. People likely have existing work-arounds in their JS which work for their use cases, and this is supposed to be a simple change.

mstoykov: Seems to me like `-bin` means that it wil be binary. So us threating it as anything different is not great.

>  ultimately a breaking change

Okay, but it is a breaking change that fixes a thing we do not support to something that we do support in a good way. 

Which ultimately makes it a bugfix.

> People likely have existing work-arounds in their JS which work for their use cases

People *can't* have workarounds for all cases - so there will be users who will need this to be binary and they will nto have a way to fix this. 

Also ... I guess this never worked until this point, so it won't even be a thing users could've actually did.

sapphire-janrain: It currently works for any data in the low end, so if someone's existing binary data only used < 0x80 it could have passed a string none the wiser. I'm not against the change exactly it's just something to consider. But yeah there certainly aren't any workarounds for *all* cases, just some.

olegbespalov:
@sapphire-janrain looks good to me :+1: let's resolve the linter issue: 

```
shadow: declaration of "ok" shadows declaration at line 456 (govet)
```

and I'll approve that. Thanks for the contribution! :bow:

js/modules/k6/grpc/client.go
```diff
@@ -459,8 +459,17 @@ func (c *Client) parseInvokeParams(paramsVal goja.Value) (*invokeParams, error)
...
+				// The spec defines that Binary-valued keys end in -bin
+				//  https://grpc.io/docs/what-is-grpc/core-concepts/#metadata
+				var strval string
+				if strings.HasSuffix(hk, "-bin") {
+					var binval []byte
+					binval, ok = kv.([]byte)
+					if !ok {
+						return result, fmt.Errorf("metadata %q value must be binary", hk)
+					}
+					strval = string(binval)
```
mstoykov, line 471: Can we link to https://github.com/grpc/grpc-go/blob/master/Documentation/grpc-metadata.md#storing-binary-data-in-metadata so that whoever comes after us knows why this is how it is done instead of keeping it in binary and giving it to the grpc library as binary

mstoykov (dismissed):
LGTM! Thanks for the work :bow:  

I will still like a comment as I had to go on a wide goose chase to figure out that the grpc library actually wants `string` and that it is supposed to work like that. 
Which I still find strange, but at least is what is expect so :shrug: 

Additionally this brings the question of how do we handle it if we get binary metadata back :thinking:

sapphire-janrain:
I made it a more permanent link in case future docs changes on their end rename or reorganize it.

mstoykov (approved):

olegbespalov (approved):

olegbespalov (approved):
LGTM! Thanks for the contribution! :bow: 

@codebien will postpone merging action to you :sweat_smile:

codebien (approved):
LGTM, thanks again for your contribution @sapphire-janrain :tada: I will merge as soon as the CI will be green.

PR #3235, opened 2023-07-28
Upgrade xk6-browser to v1.0.0

### LINKED ISSUE #3197: `K6_BROWSER_ENABLED` flag is redundant with browser type scenario option

ka3de:
## What?

This PR upgrades the [xk6-browser](https://github.com/grafana/xk6-browser) extension version to [v1.0.0](https://github.com/grafana/xk6-browser/releases/tag/v1.0.0). Being this one the latest release for the extension.

Additionally, it also removes the requirement for the `K6_BROWSER_ENABLED` environment variable in order to execute tests that use the experimental browser module.

## Why?

**Upgrade**
Keep xk6-browser extension version up to date in k6.

**Removal of `K6_BROWSER_ENABLED` flag**
Currently (as of [xk6-browser v1.0.0](https://github.com/grafana/xk6-browser/releases/tag/v1.0.0)) the browser extension requires the definition of the browser type parameter inside the options element for every scenario that wants to use the browser module. Therefore the usage of `K6_BROWSER_ENABLED` flag as an explicit approval to use the browser module has become redundant, as both parameters have to be set, being the scenario one more restrictive.

## Checklist

## Related PR(s)/Issue(s)

Closes #3197


cmd/tests/cmd_run_test.go
```diff
@@ -1827,80 +1827,6 @@ func BenchmarkReadResponseBody(b *testing.B) {
 	cmd.ExecuteWithGlobalState(ts.GlobalState)
 }
 
-func TestBrowserPermissions(t *testing.T) {
```
mstoykov, line 1830: Can we actually have a test with the scenarios options instead?

mstoykov (approved):
I would still prefer to have a test that the options enable the browser and without them it won't work.

inancgumus (approved):

oleiade (approved):

cmd/tests/cmd_run_test.go
codebien: +1 for having an e2e coverage

js/jsmodules.go
```diff
@@ -37,7 +37,7 @@ func getInternalJSModules() map[string]interface{} {
 		"k6/experimental/grpc":       expGrpc.New(),
 		"k6/experimental/timers":     timers.New(),
 		"k6/experimental/tracing":    tracing.New(),
-		"k6/experimental/browser":    browser.New(),
+		"k6/experimental/browser":    expBrowser.New(),
```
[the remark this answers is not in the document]
codebien: @ankur22 Yes, I meant it.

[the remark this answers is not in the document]
ankur22: Resolved in 8e31247639c2b9b79abc0979c467fcdfce66deba

cmd/tests/cmd_run_test.go
codebien: > isn't being closed when the test exits, am i correct in thinking that?

Yes, it seems you're right. Do you have a guess why this is happening? I see the code for unsubscribing is implemented on the core and invoked from the browser.

>I'd prefer to suppress this error for now and work on the fix in the next cycle.

How do you intend to suppress it? Do you want to remove the e2e test? I would prefer if at least we understand the root of the problem and we open an issue so we can classify it as non-critical with confidence.

ankur22: Ok, I'll take a look.

ankur22: I've created a new [issue](https://github.com/grafana/xk6-browser/issues/989). I believe the issue actually lies within the browser module scope and so we will work on it asap.

When the test run completes, does the k6 processes exit, or can it ever be reused? The reason I'm asking is because these uncleared goroutines will leak if the k6 process is reused. If it isn't then this is non-critical issue, but we will not be able to add the e2e test in this release.

ankur22: We have a fix for the leaky goroutines in https://github.com/grafana/xk6-browser/pull/990. As soon as that has been merged into main, i'll upgrade this branch to work with that commit from main, and add the e2e test in.

ankur22: I've updated this branch to work with v1.0.1 of xk6-browser which contains the fix for the goroutine leaks. I've also added an e2e test.

andrewslotin (commented):
@ka3de, could you also drop [K6_BROWSER_ENABLED](https://github.com/grafana/k6/blob/5b5c5ccbb869d45a962b22b2f8f343cc0a01b906/Dockerfile#L40) from the Dockerfile?

andrewslotin (approved):

ka3de:
> @ka3de, could you also drop [K6_BROWSER_ENABLED](https://github.com/grafana/k6/blob/5b5c5ccbb869d45a962b22b2f8f343cc0a01b906/Dockerfile#L40) from the Dockerfile?

Done in [85af866](https://github.com/grafana/k6/pull/3235/commits/85af866134c39dcf7fbc561feac48c463c22ce99).

inancgumus (approved):

cmd/tests/cmd_run_test.go
ankur22: [This](https://github.com/grafana/k6/actions/runs/5751162829/job/15589353654?pr=3235) has now been resolved in a32cc6fda7f2f57f8e26a1e0140bb54aa4ceba67. Could you please take a look @mstoykov and @codebien?

mstoykov (approved):

Issue #3232, opened 2023-07-27
Google/protobuf/wrappers.proto support is not implemented in K6


zibul444:
### Brief summary

I wrote my service, and decided to test it:

service TestService {

  rpc Foo(Request) returns (google.protobuf.BoolValue) {
  }

  rpc Bar(Request) returns (Response) {
  }

}

message Request {
}

message Response {
  bool value = 1;
}

console.info("GRPC response code: " + response.status + ", response body: " + JSON.stringify(response.message));
console.info(`${subMethod} Response: ${JSON.stringify(response)}`)

INFO[0000] GRPC response code: 0, response body: {}      source=console
INFO[0000] Foo Response: {"message":{},"error":null,"headers":{"content-type":["application/grpc"]},"trailers":{},"status":0}  source=console
INFO[0000] GRPC response code: 0, response body: {"value":false}  source=console
INFO[0000] Bar Response: {"message":{"value":false},"error":null,"headers":{"content-type":["application/grpc"]},"trailers":{},"status":0}  source=console
Actually, this is the problem. Google/protobuf/wrappers.proto support is not implemented in K6

### k6 version

v0.44.1

### OS

all

### Steps to reproduce the problem

```
rpc Foo(Request) returns (google.protobuf.BoolValue) {
  }
```

```
client.connect(api_url, {
    plaintext: true,
    reflect: true,
    timeout: params.timeout,
    maxReceiveSize: 300000
});
response = client.invoke(method, body, params)

console.info(`${subMethod} Response: ${JSON.stringify(response)}`);
```

### Expected behaviour

INFO[0000] Bar Response: {"message":"false","error":null,"headers":{"content-type":["application/grpc"]},"trailers":{},"status":0}  source=console

### Actual behaviour

INFO[0000] Bar Response: {"message":"","error":null,"headers":{"content-type":["application/grpc"]},"trailers":{},"status":0}  source=console

oleiade:
Original community support request: https://github.com/grafana/k6/issues/3232

PR #3238, opened 2023-07-28
Google/protobuf/wrappers.proto support is not implemented in K6

zibul444:
## What?
The GRPC response.message is empty

## Why?

The response from the GRPC service, which uses google/protobuf/wrappers, is not processed correctly

## Checklist

## Related PR(s)/Issue(s)

<https://github.com/grafana/k6/issues/3232>

Yes, this closes the problem

<https://github.com/grafana/k6/issues/3232>

### DIFF
--- a/lib/netext/grpcext/conn.go
+++ b/lib/netext/grpcext/conn.go
@@ -162,8 +162,11 @@ func (c *Conn) Invoke(
 		// {"x":6,"y":4,"z":0}
 		raw, _ := marshaler.Marshal(resp)
 		msg := make(map[string]interface{})
-		_ = json.Unmarshal(raw, &msg)
+		er := json.Unmarshal(raw, &msg)
 		response.Message = msg
+		if er != nil {
+			response.Message = string(raw)
+		}
 	}
 	return &response, nil
 }


zibul444:
I could not write a test because protoparce does not give an error when importing the import package "google/protobuf/wrappers.proto";. I do not know how to solve this, I cannot send the code of the test itself, I rewrote the test of course. But I could not overcome the protoparce error

zibul444:
test-current-cov 
GoError: context deadline exceeded: connection error: desc = \"transport: failed to write client preface: write tcp 127.0.0.1:53004->127.0.0.1:34205: write: connection reset by peer\" at reflect.methodValueCall (native)" does not contain "remote error: tls: bad certificate
a you help my problem?

fabiogoshippo:
I got a similar error
image

No gRPC responce.

I am binding the service on kubernetes cluster port forward 
image

Any chances to get this sorted? No problems calling the service with postman

lib/netext/grpcext/conn.go
```diff
@@ -162,8 +162,11 @@ func (c *Conn) Invoke(
 		// {"x":6,"y":4,"z":0}
 		raw, _ := marshaler.Marshal(resp)
 		msg := make(map[string]interface{})
-		_ = json.Unmarshal(raw, &msg)
+		er := json.Unmarshal(raw, &msg)
```
olegbespalov, line 165: I'm wondering if we could somehow identify that raw is a wrapped :thinking: 

Or, more precisely, I'm concerned that in case of Unmarshaling error, we safely assume that we can use raw

olegbespalov:
@fabiogoshippo, from the error, it seems your problem is either authentication or a different one, and anyway, the pull request isn't the best place to discuss it. 

Could you please open the topic (with all details and script without any business specific) on [our community forums](https://community.grafana.com/c/grafana-k6/70)? We (or the community) will try to help you there.

Thanks.

zibul444:
@olegbespalov
> Unfortunately this is a flaky tests that is known, you could ignore this failure
yes, without this error, all other tests work. 

> What kind of error? Could you maybe commit the test (even if it's failing)
The test crashes, because server emulation has not been performed, due to the lack of dependencies, maybe you can tell where exactly they are and how you can put them?
Error:
--- FAIL: TestInvoke (0.00s)
panic: panic handling "google/protobuf/wrappers.proto": runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: panic handling "google/protobuf/wrappers.proto": runtime error: invalid memory address or nil pointer dereference

lib/netext/grpcext/conn.go
zibul444: unfortunately, I can't tell you exactly how this can be done, but I will accept any help. I will assume that it is possible, but if you complicate the checks, it can greatly affect performance - this is also bad.

zibul444:
@olegbespalov

> Unfortunately this is a flaky tests that is known, you could ignore this failure

yes, without this error, all other tests work.

> What kind of error? Could you maybe commit the test (even if it's failing)

The test crashes, because server emulation has not been performed, due to the lack of dependencies, maybe you can tell where exactly they are and how you can put them?
Error:
--- FAIL: TestInvoke (0.00s)
panic: panic handling "google/protobuf/wrappers.proto": runtime error: invalid memory address or nil pointer dereference [recovered]
panic: panic handling "google/protobuf/wrappers.proto": runtime error: invalid memory address or nil pointer dereference

zibul444:
@olegbespalov
I can send my test here, in the comments, I can push it - but then the check will get another failed unit test

olegbespalov:
@zibul444 I mean, the end-goal is to have the test for the functionality, so even if it's failing, I guess it's OK to push it, I'll have a look at the code and maybe come up with the advice on how to resolve failure

lib/netext/grpcext/conn.go
mstoykov, line 165: nit: you should be able to assign to `err` without `:=`  and without getting a complain about shadowing

mstoykov: I find this fix ... strange at best and mostly hacky at this point.

I can't imagine that this is how it is supposed to work and also this will practically return you broken responses in case of real errors.

Additionally the lack of unit test for this means that I can't even easily check how this change at this point.

Can we please at least get a unit test for this.

@olegbespalov I guess you are trying to figure out how this is supposed to work actually ?

zibul444: I don't know what you're talking about . Changing the variable name solves the shading problem. The `err` variable already exists. What exactly did I do wrong?

mstoykov: You should just assign this to `err` instead of addng a new variable.

I don't se ay reason not to.

zibul444:
@olegbespalov
> @zibul444 I mean, the end-goal is to have the test for the functionality, so even if it's failing, I guess it's OK to push it, I'll have a look at the code and maybe come up with the advice on how to resolve failure

I have added a test, but it is always with an error

```
--- FAIL: Test Invoke (0.00)
panic: panic handling "google/protobuf/wrappers.proto": runtime error: invalid memory address or null pointer dereference [recovered]
panic: panic handling "google/protobuf/wrappers.proto": runtime error: invalid memory address or nil pointer dereference

```
I do not know how to add `import "google/protobuf/wrappers.proto";`, I assume that files are required

lib/netext/grpcext/conn.go
```diff
@@ -162,8 +162,11 @@ func (c *Conn) Invoke(
 		// {"x":6,"y":4,"z":0}
 		raw, _ := marshaler.Marshal(resp)
 		msg := make(map[string]interface{})
```
olegbespalov, line 164: Okay, I've looked into this and believe that this line is an issue, switching to this should help
```suggestion
		var msg interface{}
```

lib/netext/grpcext/conn_test.go
```diff
@@ -24,26 +24,70 @@ import (
 )
 
 func TestInvoke(t *testing.T) {
-	t.Parallel()
+	ctx := context.Background()
```
olegbespalov, line 27: I believe the https://github.com/grafana/k6/blob/master/js/modules/k6/grpc/client_test.go is the better test for testing this, and we might need to introduce one more test service, which could contain a different method to test the wrappers.

olegbespalov (changes requested):
Hi @zibul444, I might belive I found the root cause and as mentioned inline the fix is switching from the `map[string]interface` to just an `interface`.

But we need a good test coverage for that.

I'm going to open the PR to the [xk6-grpc](https://github.com/grafana/xk6-grpc) and depending on the pace we could process this PR.

olegbespalov:
There is an example of he fix + tests for the unary parts of the xk6-grpc module. I still need to do a couple of more commits related to the streaming API https://github.com/grafana/xk6-grpc/pull/49

olegbespalov:
Hey @zibul444 

The changes I mentioned earlier were merged into the xk6-grpc module, and the `k6/experimental/grpc` module should now support google wrappers. So, in theory, if you build the k6 from a master branch, the issue you've reported should be fixed.

Since we want to foster community contributions, I'd like to check if you plan to finalize this PR? If not, we could backport the changes by ourselves since we're close to the moment when we're planning what should the following k6 v0.47 release includes.

In any case, thank you for reporting the issue :bow:

olegbespalov:
@zibul444  since we're wrapping the `v0.47` we decided to close this PR in favor of #3344. Hope that works, thanks for contributing!

PR #3261, opened 2023-08-09
Refactor browser module test to run a fake cmd

ankur22:
## What?

Instead of relying on a chrome instance being present on the dev/ci machine to verify that the correct browser options do indeed start a browser process and run the browser tests, we will instead rely on the browser module trying to start a fake command and ensure that the error message verifies that it attempted to start the fake command.

We also think that this is a useful feature to help avoid unnecessary issues for users who do not work with chrome on their dev machines or aren't working with the browser module.

## Why?

We're not able to run the browser module test on the arm test runner since it does not have chrome installed.

## Checklist

## Related PR(s)/Issue(s)

Related to the ARM runnier error here: https://github.com/grafana/k6/actions/runs/5806262818/job/15739071012

### DIFF
--- a/cmd/tests/cmd_run_test.go
+++ b/cmd/tests/cmd_run_test.go
@@ -2164,12 +2164,27 @@ func TestBrowserPermissions(t *testing.T) {
 		expectedExitCode exitcodes.ExitCode
 		expectedError    string
 	}{
+		// When we do not supply the correct browser options,
+		// we expect that the test iteration will stop and not
+		// attempt to start a chrome instance.
 		{
 			name:             "browser option not set",
 			options:          "",
 			expectedExitCode: 0,
 			expectedError:    "GoError: browser not found in registry. make sure to set browser type option in scenario definition in order to use the browser module",
 		},
+		// When we do supply the correct browser options,
+		// we expect that the browser module will start
+		// the a chrome instance and work with the browser
+		// APIs defined in the test script.
+		//
+		// We do not actually want to rely on having chrome
+		// installed on dev/ci machines, and all we need to
+		// verify is that the browser module does try to exec
+		// a command when the correct browser options are supplied.
+		// This test will try to run a "fake" command and we
+		// expect that the test will fail when attempting to
+		// run that command.
 		{
 			name: "browser option set",
 			options: `export const options = {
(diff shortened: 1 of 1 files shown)


oleiade (approved):

cmd/tests/cmd_run_test.go
```diff
@@ -2158,6 +2159,13 @@ func BenchmarkRunEvents(b *testing.B) {
 func TestBrowserPermissions(t *testing.T) {
 	t.Parallel()
 
+	// This will skip running this test if chrome or chromium is not installed
+	// which is required to run the test to ensure that the browser module test
+	// only runs when the prerequisite browser options are supplied.
+	if !foundChrome(t) {
+		t.Skip()
+	}
+
```
mstoykov, line 2168: Two problems with this:
1. if we stop having chrome in any CI it will stop actually runnign this test at all
2. this still runs chromium for this test.

I would kind of recommend trying to run a command that *won't* work and then expect that:
1. if no borwser optiosn - we get one error
2. if browser option is available - we get the error that you can't run the "fake" command we chose.

ankur22: > if browser option is available - we get the error that you can't run the "fake" command we chose.
That's a good option. I'll give that a go.

mstoykov (dismissed):

cmd/tests/cmd_run_test.go
ankur22: Resolved in 04f2fb9baee3411003704aea80cbccc70e46600d

cmd/tests/cmd_run_test.go
```diff
@@ -2184,8 +2199,8 @@ func TestBrowserPermissions(t *testing.T) {
 					},
 				},
 			}`,
-			expectedExitCode: 0,
-			expectedError:    "",
+			expectedExitCode: 108,
+			expectedError:    "error building browser on IterStart: launching browser: exec: \"k6-browser-fake-cmd\": executable file not found in $PATH",
```
olegbespalov, line 2203: I'm curious if it's the same on windows machines:thinking:

ankur22: Very good spot, i've resolved it in 59055e787215d99637aa9656d55c69df906bbe64

olegbespalov (approved):

mstoykov (approved):

Issue #3273, opened 2023-08-14
SBOM github action still not triggering


mstoykov:
# What?
Initially added in #3089 and tried to be fixed in #3254  - the sbom github action did not trigger for the v0.46.0 release.

# Fix:
I have no idea what is going on ... and I guess we can just make it part of the release action instead and that will work :shrug:

tungbq:
Hello, I have a look at the issue and here is what's my finding so far:

There has a similar issue discussed at https://github.com/orgs/community/discussions/27028
`There is a limitation of workflow: An action in a workflow run can’t trigger a new workflow run.
`
> When you use GITHUB_TOKEN in your actions, all of the interactions with the repository are on behalf of the Github-actions bot. The operations act by Github-actions bot cannot trigger a new workflow run.

That's what's going on in our release:
- Looking at the release we could see that it was published by the `github-action bot`
![image](https://github.com/grafana/k6/assets/85242618/f957e0c3-b878-4747-93c8-f4a3093a1131)
- Then that release event generated by `github-action bot` cannot trigger a new workflow run
- That's why the #3254 change did not work

I think we possibly have 3 approaches to resolve this issue:

1. Same as @mstoykov idea - we can just make it part of the release action instead
2. Use our own PAT when creating release. We can store your PAT in secrets and use ${{ secrets.PATNAME } in our actions.
    ```
    env:
       GITHUB_TOKEN: ${{ secrets. PATNAME }}
    ```
    Downside: I personally think adding a PAT to every repo in our organization is not a good option. Also, we'd need the admin permission to add the credentials

3. Update https://github.com/grafana/k6/blob/master/.github/workflows/sbom-report.yml workflow to run on tag created event instead

Not sure which is the best approach for our repository, I personally prefer the approach 1.

Please check and give advice if any. Then if we come with the approach the I could help, I will be willing to open the PR to resolve the issue. Thanks

mstoykov:
Thanks for the excellent investigation @tungbq :bow: 

Looking at the three option I think we can't do `3` as part of the idea here is to attach this to the release (automatically). 

And this is something that won't work as at the creation of the tag the release is not created yet.

Given that `2` seems like a lot more work, I expect that we will go with putting the sbom generation as part of the release process.

If you want I think we can have a PR made by you :)

tungbq:
> Thanks for the excellent investigation @tungbq 🙇
> 
> Looking at the three option I think we can't do `3` as part of the idea here is to attach this to the release (automatically).
[...]

PR #3089, opened 2023-05-23
SBOM

SadFaceSmith:
Quick PR to add SBOM report generation on release. We (Security Engineering) have this running in grafana/grafana and grafana/loki. We are planning to expand this to all our OSS Projects.

PR #3279, opened 2023-08-15
ci: fix issue SBOM not being triggered 🐛 

tungbq:
## What?
Follow analysis provided in #3273 

Putting the SBOM generation as part of the release process:
- Add  `Anchore SBOM Action` as part of publish-github stage
- Remove `.github/workflows/sbom-report.yml`

## Why?
Initially SOM was added in https://github.com/grafana/k6/pull/3089 and tried to be fixed in https://github.com/grafana/k6/pull/3254 - the sbom github action did not trigger for the v0.46.0 release.

## Checklist

## Related PR(s)/Issue(s)
Closes: #3273


.github/workflows/build.yml
```diff
@@ -353,6 +353,10 @@ jobs:
           path: dist
       - name: Generate checksum file
         run: cd dist && sha256sum * > "k6-${VERSION}-checksums.txt"
+      - name: Anchore SBOM Action
+        uses: anchore/sbom-action@v0.12.0
+        with:
+          artifact-name: ${{ github.event.repository.name }}-spdx.json
```
mstoykov, line 359: I wonder if the asset detection wilkl "work" as such I do think it might be better if we actually output it to the `dist` directory and let teh `create release` below add it to the assets :thinking: 

Maybe also setting a bunch of the other options to false as action should not try to upload anything on it 's own now that we have it as part of the release

mstoykov (changes requested):

.github/workflows/build.yml
tungbq: Hi @mstoykov, I've just updated the **Anchore SBOM Action** to add the SBOM output to `dist` directory in my latest commit. It then be processed by the `create release` step to add to the release assets.

mstoykov (approved):

.github/workflows/build.yml
```diff
@@ -353,6 +353,12 @@ jobs:
           path: dist
       - name: Generate checksum file
         run: cd dist && sha256sum * > "k6-${VERSION}-checksums.txt"
+      - name: Anchore SBOM Action
```
olegbespalov, line 356: I'm wondering if we should block the release if this action fails :thinking: Like should this part has something like continue-on-error https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error

@tungbq @mstoykov what do you think?

tungbq: @olegbespalov Yep, I understand the concern as this is part of the release process, and this is the first time we integrate it.
From my side, I'd prefer adding `continue-on-error:true` too.
Not sure how important this SBOM file is (from the release perspetive), but I think we can first add `continue-on-error:true` flag, then monitor some releases and once everything stable, we can disable this flag in the future.

mstoykov: I think this is fine for now.

Preferably we will get notified if this doesn't run, but to be honest I prefer to get this silently broken then having to redo a release because of it.

I guess if we can make it a step *after* the release that just uploads it :thinking: ... that might work :shrug:

olegbespalov: Okay, since the concern is really hypothetical let's leave it for now :+1:

olegbespalov (approved):

.github/workflows/build.yml
mstoykov: I think my comment was not clear - I am fine with what @olegbespalov proposes.

`continue-on-error:true` will also not send email to the person who triggered the workflow if it fails. So it will "silently" fail, which was what I referred to.

But breaking the whole release for this seems ... bad to me

mstoykov (changes requested):
Can we get the "continue-on-error:true" set.This seems way better for this step, it was added 2 releases ago and the only reason why we are actually fixing it is because we noticed. Nobody actually came to tell us it isn't working :)

tungbq:
> Can we get the "continue-on-error:true" set.This seems way better for this step, it was added 2 releases ago and the only reason why we are actually fixing it is because we noticed. Nobody actually came to tell us it isn't working :)

Sure! I added that flag on the latest commit. Please review again. Thanks!

mstoykov (approved):

olegbespalov (approved):

PR #3289, opened 2023-08-18
tc39: Don't run some unicode tests as they are flaky between go versions

mstoykov:
Go 1.21 adds support for more unicode, which is great, but this also means that now some tests are passing that previously weren't on 1.20.

It is preferable that our tests don't fail or not fail between different versions of go - especially ones that we support.

This can be dropped once 1.21 is the lowest go version we support.


js/tc39/tc39_test.go
```diff
@@ -224,6 +224,11 @@ var (
 		"test/built-ins/Object/prototype/__define", // AnnexB defineGetter defineSetter
 
 		"test/language/expressions/dynamic-import", // not supported
+
+		// Go 1.16 only supports unicode 13; Whiule go 1.21 has support for this - we disable them
```
olegbespalov, line 228:
```suggestion
		// Go 1.16 only supports unicode 13; While go 1.21 has support for this - we disable them
```

olegbespalov (dismissed):

olegbespalov (approved):

oleiade (approved):

PR #3302, opened 2023-08-24
cloudv2: Higher resolution for Histograms

codebien:
## What?

It expands the resolution for Histograms, setting `0.001` as the minimum resolution.

## Why?

It allows us to support decimal numbers up to 3 digits, increasing the detail of the values that we can store.

### DIFF
--- a/output/cloud/expv2/hdr.go
+++ b/output/cloud/expv2/hdr.go
@@ -28,7 +28,8 @@ const (
 // of metrics samples' values as histogram.
 //
 // The histogram is the representation of base-2 exponential Histogram with two layers.
-// The first layer has primary buckets in the form of a power of two, and a second layer of buckets for each primary bucket with an equally distributed amount of buckets inside.
+// The first layer has primary buckets in the form of a power of two, and a second layer of buckets
+// for each primary bucket with an equally distributed amount of buckets inside.
 //
 // The histogram has a series of (N * 2^m) buckets, where:
 // N = a power of 2 that defines the number of primary buckets
--- a/output/cloud/expv2/hdr.go
+++ b/output/cloud/expv2/hdr.go
@@ -9,6 +9,11 @@ import (
 )
 
 const (
+	// defaultMinimumResolution is the default resolution used by histogram.
+	// It allows to have a higher granularity compared to the basic 1.0 value,
+	// supporting floating points up to 3 digits.
+	defaultMinimumResolution = .001
+
 	// lowestTrackable represents the minimum value that the histogram tracks.
 	// Essentially, it excludes negative numbers.
 	// Most of metrics tracked by histograms are durations
@@ -17,12 +22,6 @@ const (
 	// In the future, we may expand and include them,
 	// probably after https://github.com/grafana/k6/issues/763.
 	lowestTrackable = 0
-
-	// highestTrackable represents the maximum
-	// value that the histogram is able to track with high accuracy (0.1% of error).
-	// It should be a high enough
-	// and rationale value for the k6 context; 2^30 = 1_073_741_824
-	highestTrackable = 1 << 30
 )
 
 // histogram represents a distribution
@@ -61,13 +60,18 @@ type histogram struct {
 
 	// Count is counts the amount of observed values.
 	Count uint32
+
+	// MinimumResolution represents resolution used by Histogram.
+	// In principle, it is a multiplier factor for the tracked values.
+	MinimumResolution float64
 }
 
 func newHistogram() *histogram {
 	return &histogram{
-		Buckets: make(map[uint32]uint32),
-		Max:     -math.MaxFloat64,
-		Min:     math.MaxFloat64,
+		MinimumResolution: defaultMinimumResolution,
+		Buckets:           make(map[uint32]uint32),
+		Max:               -math.MaxFloat64,
+		Min:               math.MaxFloat64,
 	}
 }
 
@@ -85,7 +89,9 @@ func (h *histogram) addToBucket(v float64) {
 	h.Count++
 	h.Sum += v
 
-	if v > highestTrackable {
+	v /= h.MinimumResolution
+
+	if v > math.MaxInt64 {
 		h.ExtraHighBucket++
 		return
 	}
@@ -151,6 +157,9 @@ func histogramAsProto(h *histogram, time int64) *pbcloud.TrendHdrValue {
 	if h.ExtraHighBucket > 0 {
 		hval.ExtraHighValuesCounter = &h.ExtraHighBucket
 	}
+	// We don't expect to change the minimum resolution at runtime
+	// so it is safe use directly a pointer without creating a copy
+	hval.MinResolution = &h.MinimumResolution
 	return hval
 }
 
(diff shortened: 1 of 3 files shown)


output/cloud/expv2/hdr.go
```diff
@@ -151,6 +164,9 @@ func histogramAsProto(h *histogram, time int64) *pbcloud.TrendHdrValue {
 	if h.ExtraHighBucket > 0 {
 		hval.ExtraHighValuesCounter = &h.ExtraHighBucket
 	}
+	// We don't expect to change the minimum resolution at runtime
+	// so a pointer is safe here
+	hval.MinResolution = &h.MinimumResolution
```
mstoykov, line 169: Is there are reason to use a pointer though?

It won't save memory and will add dereferencing the point when we have to actually write it.

mstoykov (approved):
LGTM in general :+1: 

I left some not blocking comments.

But I would prefer if we get some tests with values between 0 and 1 which is what this was about. Maybe even with some values that the browser team think are representable.

output/cloud/expv2/hdr.go
mstoykov: I guess the problem is that this is because in the protobuf definition this is not required :facepalm: 

Not certain we can do anythign about this :shrug:

codebien: Yeah, the reason is how Protobuf defines optional values. We haven't released yet the new cloud output v2 as the default so we could think of migrating directly to a new Protobuf version making it required. But my feeling is that it would require a maintenance window.

olegbespalov (approved):
LGTM, but joining to Mihail on his comments :+1:

output/cloud/expv2/hdr.go
```diff
@@ -20,9 +25,10 @@ const (
 
 	// highestTrackable represents the maximum
 	// value that the histogram is able to track with high accuracy (0.1% of error).
-	// It should be a high enough
-	// and rationale value for the k6 context; 2^30 = 1_073_741_824
-	highestTrackable = 1 << 30
+	// It represents the maximum number for int64 using the
+	// default minimum resolution, where it multiplies the value ingested from the histogram
+	// of a specific factor. Essentially, having higher resolution reduces the highest limit.
+	highestTrackable = 9223372036854775 // math.MaxInt64*defaultMinimumResolution
```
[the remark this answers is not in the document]

output/cloud/expv2/hdr_test.go
```diff
@@ -161,43 +179,49 @@ func TestHistogramAddWithMultipleOccurances(t *testing.T) {
 		Sum:             466.20000000000005,
 		Count:           5,
 	}
+	exp.MinimumResolution = 1.0
 	assert.Equal(t, exp, h)
 }
 
 func TestHistogramAddWithNegativeNum(t *testing.T) {
```
codebien, line 186: Autospot: add a test with a 0-1 range values as the lined comment

>Maybe even with some values that the browser team think are representable.

olegbespalov (dismissed):

mstoykov (dismissed):
LGTM!

Do we have any statistics on how much this change makes changes over 1 less accurate?

I guess it will be fairly good to have some kind of graph showing how accurate the samples are between the values 0 and ... something Less say whatever 1 minute is in ms so 1 *60 * 100 = 6000?

mstoykov:
That graph is mostly interesting for the release notes, but it will still be good to have it ealier IMO.

codebien:
We should not lose any precision because what we are doing is just using a factor for moving the values across the histogram. The relative error rate is generated from the number of buckets that we have in the range, and it is a constant value. In our case the error rate is ~1%.

```
Formula Precision =  100% / (number of sub-buckets in a major bucket) = 100 / 2^7 = 0.78
```

So the boundaries of each value are just upscaled for storage and downscaled for query but they will be the same.

| Resolution | Seconds (v/min_res) | In Millis (v*1e3) | Lower limit [ millis - (millis * 1%) ]+ | Upper limit [ millis + (millis * 1%) ] |
|------------|---------------------|-------------------|-----------------------------------------|----------------------------------------|
| 1.0        | 60                  | 60k               | 59.4k                                   | 60.6k                                  |
| 0.001      | 60k                 | 60M               | 59.4M                                   | 60.6M                                  |

@mstoykov does it make sense for you?

output/cloud/expv2/hdr.go
```diff
@@ -20,9 +25,10 @@ const (
 
 	// highestTrackable represents the maximum
 	// value that the histogram is able to track with high accuracy (0.1% of error).
-	// It should be a high enough
-	// and rationale value for the k6 context; 2^30 = 1_073_741_824
-	highestTrackable = 1 << 30
+	// It represents the maximum number for int64 using the
+	// default minimum resolution, where it multiplies the value ingested from the histogram
+	// of a specific factor. Essentially, having higher resolution reduces the highest limit.
+	highestTrackable = math.MaxInt64 * defaultMinimumResolution
```
esquonk, line 31: Perhaps I am missing something but highest trackable value is checked after dividing by resolution `v /= h.MinimumResolution`, and so resolution should not be mentioned here.

```suggestion
	highestTrackable = math.MaxInt64
```

output/cloud/expv2/hdr.go
esquonk: I believe that if we un-optional this field, this doesn't really change anything for the backend?

codebien: @esquonk if the suggestion is to change it only on the client then it would require having different proto file versions between the backend and client, that doesn't sound optimal to me.
Instead, if you mean for both of them, then In that case, the problem is how to handle the versioning. If we set the field on the backend as required then the already k6 v0.46.0 released version will stop working if we force the `K6_CLOUD_API_VERSION=2` overwriting it from the backend. We could consider it as a no-problem as we are pretty confident that we will set it directly from the code releasing v0.47.0, but I'm not sure. Thoughts?

output/cloud/expv2/hdr.go
codebien: This is now expanded to the highest, good catch.

codebien:
Hey @olegbespalov @mstoykov I've pushed a new fixup for fixing https://github.com/grafana/k6/pull/3302#discussion_r1310488965. It was a mistake because we are checking the max value after the resolution is applied, so the value will be checked already as the final value.

Can you take another look, please?

mstoykov (approved):

output/cloud/expv2/hdr_test.go
```diff
@@ -116,9 +130,11 @@ func TestHistogramAddWithSimpleValues(t *testing.T) {
 		t.Run(strconv.Itoa(i), func(t *testing.T) {
 			t.Parallel()
 			h := newHistogram()
+			h.MinimumResolution = 1.0
 			for _, v := range tc.vals {
 				h.Add(v)
 			}
+			tc.exp.MinimumResolution = 1.0
```
olegbespalov, line 137: nit: it's better to keep this along with the other expectations

olegbespalov (approved):
Generally LGTM :+1: 

But I prefer to do a clean-up of the tests (mentioned in the comments) in the current PR to avoid future confusion.

However, I don't know how urgent is merging this, so approving it.

codebien:
Hey @olegbespalov, we decided to make the `min_resolution` required instead of `optional`, so I need to rework those tests anyway. I will address your valuable suggestions directly in the new PR (that I will push today, maximum tomorrow). I would prefer to not block this PR so I can unlock the browser for testing the full picture.

olegbespalov:
@codebien Sure, go ahead and merge it :+1:

PR #3303, opened 2023-08-24
Add `triage` label to both of the current issue templates

mstoykov:


olegbespalov:
How about adding the logic to the auto-assigner instead of in the template? The rational is that we want to use the same logic that we use in the auto-assigner, like if the not maintainer opens the issue, then assigns `triage`, how does that sound, and having this only in temples forces us to have the temples in other repos?  labels we probably should create in any case :thinking:

mstoykov:
I was left with the impression we will also use the templates in most places?

>  like if the not maintainer opens the issue,

That is a good point , but AFAIK you can just remove it while making the issue and I personally rarely use the template as it request 10 fields and usually it is much easier and IMO useful to make a custom template if you know what you are doing.

The template is mostly there because half the time the bug reported were "k6 run my-script.js throws exception" and thats it. At whihc point you start a game of 20 questions. While this is not a problem when a maintainer reports an issue.

To be honest if we are going to do in the action, we might as well try to add it to a kanban board of some kind so that there is a single place where you can see all triage issues across all projects. As that is the other problem currently - you need to go to each project to see all not triaged issues. (unless they are assigned to you in which case you can see that in https://github.com/issues/assigned). 

But it will be nice to have a more central place for this k6 OSS issues. 

https://docs.github.com/en/enterprise-cloud@latest/issues/planning-and-tracking-with-projects/automating-your-project/adding-items-automatically this seems relevant :thinking:

olegbespalov:
Yeah, some board could be good, but it could raise some permission questions and so on. I was thinking that we could have a simple search with a query like:

```
is:open is:issue label:triage user:grafana sort:comments-asc  
```

Or a link
https://github.com/issues?q=is%3Aopen+is%3Aissue+label%3Atriage+user%3Agrafana+sort%3Acomments-asc

That we check from time to time

olegbespalov (approved):

mstoykov:
This looks okay to me :thinking: . I am worried that it will be terrible once more then just us start using `triage` as a lable as we will see it from all repos under grafana.

But at least currently this is not a problem.

Have you looked into setting a tag with the action - if that is can be done, I guess it is just as okay to hae it that way. 

I did go look at it and projects can have a way to add stuff to them, but we are limited to 5 or 20 depending on what our subscription is. And each auto-add is per *repo* so best case scenario we are limited to 20 repos. 

Whcih seems enough, but likely a limit will hit.

I do think that it probably will be okay to have a public project and using the authentication that the action automatically gets currently. IF it fails - it means someone is running it from a repo not under grafana ....  I think. So that is likely fine.


But let's first see if the action can just add `triage` for now, and then if that works - let's go with that for now. WDYT @olegbespalov @oleiade

olegbespalov:
>Have you looked into setting a tag with the action - if that is can be done, I guess it is just as okay to hae it that way.

Yeah, it is listed as the example :sweat_smile: there https://docs.github.com/en/actions/managing-issues-and-pull-requests/adding-labels-to-issues

```
github.rest.issues.addLabels({
  issue_number: context.issue.number,
  owner: context.repo.owner,
   repo: context.repo.repo,
  labels: ["triage"]})
```

I believe we still have to create the `triage` label by ourselves in each repo (including k6), but even for the template, it's also the case.

mstoykov:
Closing in favor of #3304

PR #3309, opened 2023-08-29
[fs] Add `file.read` and `file.seek` methods to the `fs` module (2/3)

### LINKED ISSUE #3145: Implement the File API's `File.read` method
As part of #3141, implementing the [agreed-upon design for k6's File API](https://github.com/grafana/k6/blob/master/docs/design/019-file-api.md), we intend to implement a `File.read` method.

The `read` method reads the file into an `ArrayBuffer`, and resolves to either the number of bytes read during the operation or `null` (EOF) if there is nothing more to read. Note that it should be possible for a read to be successful and return `0` bytes read while not indicating an EOF.

### prototype
```javascript
interface File {
    read(p ArrayBuffer | TypedArray | DataView): Promise<number>
}
```

### example

```javascript
const file = fs.openSync("fou.txt")
const buf = new Uint8Array(100);
const numberOfBytesRead = file.readSync(buf)
```

### LINKED ISSUE #3148: Implement the File API's `File.seek` method
As part of #3141, implementing the [agreed-upon design for k6's File API](https://github.com/grafana/k6/blob/master/docs/design/019-file-api.md), we intend to implement a `File.seek` method.

The `seek` method seeks the file to handle to the given `offset` under the mode provided by `whence`. The call resolves with the new position within the resource (offset from the start, in bytes).

### prototype
```typescript
interface File {
    seek(offset: number, whence: SeekMode): Promise<number>
}

enum SeekMode {
    Current = 1
    End = 2
    Start = 0
}
```

### example

```typescript
const file = fs.openSync("fou.txt")
const cursorPosition = await file.seek(6, fs.SeekMode.Start)

```

oleiade:
## What?

This Pull Request builds (and is based) upon #3142 and adds a `read` and a `seek` method, the the `File` object exposed by the `k6/experimental/fs` module. 

### `read`

The `File.read` method allows to read a file into a `Uint8Array`. The amount of bytes read is equal to the size of the provided buffer.

The method resolves to either the number of bytes read during the operation or EOF (represented as `null`) if there is nothing more to read.

### `seek`

The `File.seek` method allows to seek to a specified `offset` within a file under the given `whence`.

The offset is expressed in bytes. Whence allows you to seek from the start of the file (the offset needs to be strictly negative), the current position within the file (offset can be either positive or negative), or from the end of the file (offset needs to be strictly negative). A dedicated `SeekMode` type/enum is exposed to help with it.

The method resolves to the new offset in bytes after the seek operation.

### In action

```javascript
import { open } from "k6/experimental/fs";

export const options = {
	vus: 100,
	iterations: 1000,
};

// As k6 does not support asynchronous code in the init context, yet, we need to
// use a top-level async function to be able to use the `await` keyword.
let file;
(async function () {
	file = await open("bonjour.txt");
})();

	// About information about the file
	const fileinfo = await file.stat();
	if (fileinfo.name != "bonjour.txt") {
		throw new Error("Unexpected file name");
	}

	// Define a buffer of the same size as the file
	// to read the file content into.
	const buffer = new Uint8Array(fileinfo.size);

	// Read the file's content into the buffer
	const bytesRead = await file.read(buffer);

	// Check that we read the expected number of bytes
	if (bytesRead != fileinfo.size) {
		throw new Error("Unexpected number of bytes read");
	}

	// Seek back to the beginning of the file
	await file.seek(0);
}

```

## Checklist

## Related PR(s)/Issue(s)

ref #3142
ref #2974
fixes #3145 
fixes #3148

### DIFF
--- a/examples/experimental/fs/fs.js
+++ b/examples/experimental/fs/fs.js
@@ -1,20 +1,52 @@
-import { open } from "k6/experimental/fs";
+import { open, SeekMode } from "k6/experimental/fs";
 
 export const options = {
 	vus: 100,
 	iterations: 1000,
 };
 
-// As k6 does not support asynchronous code in the init context, yet, we need to
-// use a top-level async function to be able to use the `await` keyword.
+// k6 doesn't support async in the init context. We use a top-level async function for `await`.
+//
+// Each Virtual User gets its own `file` copy.
+// So, operations like `seek` or `read` won't impact other VUs.
 let file;
 (async function () {
 	file = await open("bonjour.txt");
 })();
 
 export default async function () {
+	// About information about the file
 	const fileinfo = await file.stat();
 	if (fileinfo.name != "bonjour.txt") {
 		throw new Error("Unexpected file name");
 	}
+
+	const buffer = new Uint8Array(4);
+
+	let totalBytesRead = 0;
+	while (true) {
+		// Read into the buffer
+		const bytesRead = await file.read(buffer);
+		if (bytesRead == null) {
+			// EOF
+			break;
+		}
+
+		// Do something useful with the content of the buffer
+
+		totalBytesRead += bytesRead;
+
+		// If bytesRead is less than the buffer size, we've read the whole file
+		if (bytesRead < buffer.byteLength) {
+			break;
+		}
+	}
+
+	// Check that we read the expected number of bytes
+	if (totalBytesRead != fileinfo.size) {
+		throw new Error("Unexpected number of bytes read");
+	}
+
+	// Seek back to the beginning of the file
+	await file.seek(0, SeekMode.Start);
 }
--- a/js/modules/k6/experimental/fs/file.go
+++ b/js/modules/k6/experimental/fs/file.go
@@ -62,3 +62,66 @@ func (f *file) Read(into []byte) (n int, err error) {
 
 // Ensure that `file` implements the io.Reader interface.
 var _ io.Reader = (*file)(nil)
+
+// Seek sets the offset for the next operation on the file, under the mode given by `whence`.
+//
+// `offset` indicates the number of bytes to move the offset. Based on
+// the `whence` parameter, the offset is set relative to the start,
+// current offset or end of the file.
+//
+// When using SeekModeStart, the offset must be positive.
+// Negative offsets are allowed when using `SeekModeCurrent` or `SeekModeEnd`.
+func (f *file) Seek(offset int, whence SeekMode) (int, error) {
+	newOffset := f.offset
+
+	switch whence {
+	case SeekModeStart:
+		if offset < 0 {
+			return 0, newFsError(TypeError, "offset cannot be negative when using SeekModeStart")
+		}
+
+		newOffset = offset
+	case SeekModeCurrent:
+		newOffset += offset
+	case SeekModeEnd:
+		if offset > 0 {
+			return 0, newFsError(TypeError, "offset cannot be positive when using SeekModeEnd")
+		}
+
+		newOffset = len(f.data) + offset
+	default:
+		return 0, newFsError(TypeError, "invalid seek mode")
+	}
+
+	if newOffset < 0 {
+		return 0, newFsError(TypeError, "seeking before start of file")
+	}
+
+	if newOffset > len(f.data) {
+		return 0, newFsError(TypeError, "seeking beyond end of file")
+	}
+
+	// Note that the implementation assumes one `file` instance per file/vu.
+	// If that assumption was invalidated, we would need to atomically update
+	// the offset instead.
+	f.offset = newOffset
+
+	return newOffset, nil
+}
+
+// SeekMode is used to specify the seek mode when seeking in a file.
+type SeekMode = int
+
+const (
+	// SeekModeStart sets the offset relative to the start of the file.
+	SeekModeStart SeekMode = iota + 1
+
+	// SeekModeCurrent seeks relative to the current offset.
+	SeekModeCurrent
+
+	// SeekModeEnd seeks relative to the end of the file.
+	//
+	// When using this mode the seek operation will move backwards from
+	// the end of the file.
+	SeekModeEnd
+)
--- a/js/modules/k6/experimental/fs/file.go
+++ b/js/modules/k6/experimental/fs/file.go
@@ -3,6 +3,9 @@ package fs
 import (
 	"io"
 	"path/filepath"
+	"sync/atomic"
+
+	"go.k6.io/k6/lib"
 )
 
 // file is an abstraction for interacting with files.
@@ -13,7 +16,11 @@ type file struct {
 	data []byte
 
 	// offset holds the current offset in the file
-	offset int
+	//
+	// TODO: using an atomic here does not guarantee ordering of reads and seeks, and leaves
+	// the behavior not strictly defined. This is something we might want to address in the future, and
+	// is tracked as part of #3433.
+	offset atomic.Int64
 }
 
 // Stat returns a FileInfo describing the named file.
@@ -38,24 +45,27 @@ type FileInfo struct {
 //
 // If the end of the file has been reached, it returns EOFError.
 func (f *file) Read(into []byte) (n int, err error) {
-	start := f.offset
-	if start == len(f.data) {
+	currentOffset := f.offset.Load()
+	fileSize := f.size()
+
+	// Check if we have reached the end of the file
+	if currentOffset == fileSize {
 		return 0, newFsError(EOFError, "EOF")
 	}
 
-	end := f.offset + len(into)
-	if end > len(f.data) {
-		end = len(f.data)
-		// We align with the [io.Reader.Read] method's behavior
-		// and return EOFError when we reach the end of the
-		// file, regardless of how much data we were able to
-		// read.
+	// Calculate the effective new offset
+	targetOffset := currentOffset + int64(len(into))
+	newOffset := lib.Min(targetOffset, fileSize)
+
+	// Read the data into the provided slice, and update
+	// the offset accordingly
+	n = copy(into, f.data[currentOffset:newOffset])
+	f.offset.Store(newOffset)
+
+	// If we've reached or surpassed the end, set the error to EOF
+	if targetOffset > fileSize {
 		err = newFsError(EOFError, "EOF")
 	}
-
-	n = copy(into, f.data[start:end])
-
-	f.offset += n
 
 	return n, err
 }
@@ -71,9 +81,10 @@ var _ io.Reader = (*file)(nil)
 //
 // When using SeekModeStart, the offset must be positive.
 // Negative offsets are allowed when using `SeekModeCurrent` or `SeekModeEnd`.
-func (f *file) Seek(offset int, whence SeekMode) (int, error) {
-	newOffset := f.offset
+func (f *file) Seek(offset int64, whence SeekMode) (int64, error) {
+	startingOffset := f.offset.Load()
 
+	newOffset := startingOffset
 	switch whence {
 	case SeekModeStart:
 		if offset < 0 {
@@ -88,7 +99,7 @@ func (f *file) Seek(offset int, whence SeekMode) (int, error) {
 			return 0, newFsError(TypeError, "offset cannot be positive when using SeekModeEnd")
 		}
 
-		newOffset = len(f.data) + offset
+		newOffset = f.size() + offset
 	default:
 		return 0, newFsError(TypeError, "invalid seek mode")
 	}
@@ -97,17 +108,17 @@ func (f *file) Seek(offset int, whence SeekMode) (int, error) {
 		return 0, newFsError(TypeError, "seeking before start of file")
 	}
 
-	if newOffset > len(f.data) {
+	if newOffset > f.size() {
 		return 0, newFsError(TypeError, "seeking beyond end of file")
 	}
 
-	// Note that the implementation assumes one `file` instance per file/vu.
-	// If that assumption was invalidated, we would need to atomically update
-	// the offset instead.
-	f.offset = newOffset
+	// Update the file instance's offset to the new selected position
+	f.offset.Store(newOffset)
 
 	return newOffset, nil
 }
+
+var _ io.Seeker = (*file)(nil)
 
 // SeekMode is used to specify the seek mode when seeking in a file.
 type SeekMode = int
@@ -125,3 +136,7 @@ const (
 	// the end of the file.
 	SeekModeEnd
 )
+
+func (f *file) size() int64 {
+	return int64(len(f.data))
+}
--- a/js/modules/k6/experimental/fs/file.go
+++ b/js/modules/k6/experimental/fs/file.go
@@ -125,16 +125,16 @@ type SeekMode = int
 
 const (
 	// SeekModeStart sets the offset relative to the start of the file.
-	SeekModeStart SeekMode = iota + 1
+	SeekModeStart SeekMode = 0
 
 	// SeekModeCurrent seeks relative to the current offset.
-	SeekModeCurrent
+	SeekModeCurrent = 1
 
 	// SeekModeEnd seeks relative to the end of the file.
 	//
 	// When using this mode the seek operation will move backwards from
 	// the end of the file.
-	SeekModeEnd
+	SeekModeEnd = 2
 )
 
 func (f *file) size() int64 {
--- a/js/modules/k6/experimental/fs/file.go
+++ b/js/modules/k6/experimental/fs/file.go
@@ -1,6 +1,7 @@
 package fs
 
 import (
+	"io"
 	"path/filepath"
 )
 
@@ -10,6 +11,9 @@ type file struct {
 
 	// data holds a pointer to the file's data
 	data []byte
+
+	// offset holds the current offset in the file
+	offset int
 }
 
 // Stat returns a FileInfo describing the named file.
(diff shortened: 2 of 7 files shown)


js/modules/k6/experimental/fs/module.go
```diff
@@ -165,3 +170,102 @@ func (f *File) Stat() *goja.Promise {
...
+	}
+
+	// Obtain the underlying byte slice from the ArrayBuffer.
+	// Note that this is not a copy, and will be modified by the Read operation
+	// in place.
+	buffer := ab.Bytes()
+
+	go func() {
+		n, err := f.file.Read(buffer)
+		if err != nil {
```
mstoykov, line 211: nit: I think this is relevant here as this `Read` is not the same as [os.File#Read](https://pkg.go.dev/os#File.Read) but it can both have read stuff *and*  returned data. 

[`io.Reader`](https://pkg.go.dev/io#Reader) does define this as a behaviour that can be expected.

I am not particularly certain how to handle this in js as  if we did read some amount and then got EOF and `Read` returns both we will hit the bleow and resolve with `nil`. Which likely isn't great.

At this point this is less of a concern as the below implementation is specifically implemented by us. But in a refactor this might become a problem. 

A possible solution could be to retrurn this buffer and remember that there was an error and return it the next time :shrug:

examples/experimental/fs/fs.js
```diff
@@ -13,8 +13,24 @@ let file;
...
 	}
+
+	// Define a buffer of the same size as the file
+	// to read the file content into.
+	const buffer = new Uint8Array(fileinfo.size);
+
+	// Read the file's content into the buffer
+	const bytesRead = await file.read(buffer);
+
+	// Check that we read the expected number of bytes
```
mstoykov, line 29: Does this mean that we want to make `read` always filling the full buffer? 

This is not how go io.Reader is usually defined to work and is not how deno's [`read`](https://deno.land/api@v1.36.3?s=Deno.FsFile#method_read_0) is. 

I do expect with the current implementation that to be true, but even so, maybe we should not say that this is how it is expected to work. And the current inline documentation suggests that this was the intention, but then this script depends on it actually reading the full bugger everytime.

In that case it might be better to have it retry reading until it reads the whole `fileinfo.size`?

js/modules/k6/experimental/fs/module_test.go
```diff
@@ -212,6 +212,133 @@ func TestFile(t *testing.T) {
...
+			return afero.WriteFile(fs, testFilePath, []byte("Bonjour, le monde"), 0o644)
+		})
+		runtime.VU.InitEnvField.FileSystems["file"] = fs
+
+		_, err = runtime.RunOnEventLoop(fmt.Sprintf(`
+			fs.open(%q).then(file => {
+				// No buffer argument provided should fail with TypeError.
+				file.read().then(
+					res => { throw 'unexpected promise resolution with result: ' + res },
+					err => { if (err.name !== 'TypeError') { throw 'unexpected error: ' + err } },
```
mstoykov, line 233: tip: You can use `await` if you wrap in `async` function as is done in https://github.com/grafana/k6/blob/4615f0a1ed1b45fa16b83f7157349a4943dc1a05/js/modules/k6/http/async_request_test.go#L12-L15

Or did you have a different reason to not using `async`/`await`

examples/experimental/fs/fs.js
```diff
@@ -13,8 +13,24 @@ let file;
...
+	if (bytesRead != fileinfo.size) {
+		throw new Error("Unexpected number of bytes read");
+	}
+
+	// Seek back to the beginning of the file
+	await file.seek(0);
```
olegbespalov, line 35: I've tried to run this example and got a bunch of errors, like:
```
ERRO[0000] Uncaught (in promise) Error: Unexpected number of bytes read
        at file:///home/olegbespalov/projects/grafana/k6/k6/examples/experimental/fs/fs.js:31:8(33)  executor=shared-iterations scenario=default
ERRO[0000] Uncaught (in promise) Error: Unexpected number of bytes read
```

Which is expected since we have 100 VUs

But even running this with just one VU and one iteration gives us an error:

```
ERRO[0000] Uncaught (in promise) TypeError: seek() failed; reason: whence cannot be null or undefined  executor=shared-iterations scenario=default
```

js/modules/k6/experimental/fs/module_test.go
oleiade: Good catch. I used the promise syntax out of convenience and because it helped avoid `try/catch` block. But as I've found out, you can use `.catch` on the result of an `await` call, which helps keep the code nice and tidy. Thus, I've ported the `wrapInAsyncLambda` helper, and used `async/await` instead 🤝

examples/experimental/fs/fs.js
oleiade: Turns out I forgot to pass the `whence` attribute 🤦🏻 
I've added it so the call looks like `await file.seek(0, SeekMode.Start)`, and it should pass on your side too now 👍🏻

examples/experimental/fs/fs.js
oleiade: > Does this mean that we want to make read always fill the full buffer?

No, `read` behaves how I understand you suggest it works. It reads as many bytes as possible into the buffer, but that amount could be smaller than the buffer indeed. Also, the resulting bytes count can be `null`, like in Deno, to indicate EOF.

I agree that the example might have been confusing in that regards, and I adapted it accordingly, to reflect reading a file in a buffered way indeed 👍🏻

js/modules/k6/experimental/fs/module.go
oleiade: Hey, I'm not sure I understand what you mean here. Especially when it comes to "it can both have read stuff and returned data.".

The reason I don't understand is likely that Read modifies the buffer in place (`buffer` is a pointer and not a copy, as I learned the hard way in WebCrypto), and returns either a number or `null` to indicate EOF.

Could you elaborate, or even better provide a JS example to illustrate what you meant? That would be super helpful 🙇🏻

mstoykov: > returns either a number or null to indicate EOF

This is true, but `io.Reader` and `os.File#Read` *can* return both `5` and `EOF` at the same time - which means that you read 5 bytes in the array and got `EOF`.

The current code will though resolve with `null` which means that a user of the js code will have no idea taht actually there were 5 bytes read. Yes they are in the ArrayBuffer - true, but you don't know that there are 5 new bytes, just that you got EOF. 

Even if for EOF you return `5`, this can still happen with some other error. In which case it won't work so well.

> or even better provide a JS example to illustrate what you meant? 

The problem here is that JS does not have multiple returns as go does. 

Again as I mention currently the underlying implementation would not present this problem as .... we can't have anyu errors really apart from EOF and it happens only when it returns 0 bytes . Which is why this is a  `nit`.

But in the future if this changes it might become a problem.

Hope this explained it more?

js/modules/k6/experimental/fs/module_test.go
```diff
@@ -212,6 +212,196 @@ func TestFile(t *testing.T) {
...
+				})
+
+			// undefined buffer argument should fail with TypeError.
+			bytesRead = await file
+				.read(undefined)
+				.catch(err => {
+					if (err.name !== 'TypeError') {
+						throw 'unexpected error: ' + err;
+					}
+				})
```
mstoykov, line 300: nit: just for a comparision 

```suggestion
			try { 
				bytesRead = await file.read(undefined)
			} catch (err)  {
				if (err.name !== 'TypeError') {
					throw 'unexpected error: ' + err;
				}
			}
```

For all my dislike of exceptions, this IMO looks just as good if not better and IMO a lot more what most people will write. I actually am interested why you didn't write it like that? You don't like how it works? You thought it won't work?

js/modules/k6/experimental/fs/module.go
oleiade: > Hope this explained it more?

Absolutely, thank you very much 🙇🏻 

Makes sense, I had completely overlooked that case, indeed! I am going to look into how to cover that scenario indeed, I find it worth it 👀

js/modules/k6/experimental/fs/module_test.go
oleiade: I see what you mean. I wrote it this way out of personal preference and curiosity for this notation way I didn't know yet.

Seeing it compared like this in hindsight, I think I agree that the `try/catch` block might indeed be easier to read and decipher 👍🏻

js/modules/k6/experimental/fs/module.go
oleiade: I aligned the behavior of `file.Read` with `io.Reader.Read`'s, adjusted the module code, and tested accordingly in [`3000972` (#3309)](https://github.com/grafana/k6/pull/3309/commits/3000972bdc4f085567e023128c894ce56a00ff1e). Let me know if you think that addresses. fully your point 🙇🏻

js/modules/k6/experimental/fs/module_test.go
oleiade: I've replaced the relevant test code with `try/catch` blocks indeed 👍🏻

js/modules/k6/experimental/fs/module.go
```diff
@@ -165,3 +170,114 @@ func (f *File) Stat() *goja.Promise {
...
+	ab, ok := intoObj.Get("buffer").Export().(goja.ArrayBuffer)
+	if !ok {
+		reject(newFsError(TypeError, "read() failed; reason: into argument cannot be interpreted as ArrayBuffer"))
+		return promise
+	}
+
+	// Obtain the underlying byte slice from the ArrayBuffer.
+	// Note that this is not a copy, and will be modified by the Read operation
+	// in place.
+	buffer := ab.Bytes()
```
mstoykov, line 207: Unfortunately I think this makes the whole operation really easy to race and I will show two examples that I have tested.

Both of them are not likely code, but this code racing IMO is pretty bad and will need to be fixed before this becomes not experimental:

```javascript
let file = fs.open(name);
let buffer = new Uint8Array(5)
let p1 = file.read(buffer);
let p2 = file.read(buffer);
await Promise.all(p1, p2);
```

```javascript
let file = fs.open(name);
let buffer = new Uint8Array(5)
let p1 = file.read(buffer);
buffer[2] = 3;
console.log(buffer); // this here is not certain whether it will have teh data read from file
// or will have just the third value 3, or combination of the two
await p1;
```

in both cases this can could modify the underlying bytes array multiple times.

Arguably the observability of a change in `buffer` before awaiting the promises is the bigger conceptual problem as that basically breaks promises behaviour - where you can not observe (and consequently race) your state around promises in any other way apart from callbacks - which make the whole thing easier.

Some experimentation(probably more required, and no documentation confirms this) with deno shows that:
1. if you have sequential calls to `read` (without await) the moment you await one of the promises - its data is populated
2. awaiting a promise from a read before one htat was already done - does not change the array buffer
3. the array buffer does not change if you do not await a promise on the read.

So it seems like the deno's implementation orders calls to `read` and skips previous ones :shrug: no idea - more investigation needed, and I would like to see their rational on this. Maybe node has some more info on this as well :shrug:  

I have the feeling we have a similar issue with `seek`, but less ... dramatic


Again I am not against this staying this way for the `experimental` module, but IMO having a way to make race conditions from js code is big no-no for any stable code.

olegbespalov:
I did a round of reviewing, mostly looks good :+1: waiting for Mihail's comment to be addressed.

js/modules/k6/experimental/fs/module_test.go
```diff
@@ -212,6 +212,196 @@ func TestFile(t *testing.T) {
...
+		require.NoError(t, err)
+
+		testFilePath := fsext.FilePathSeparator + "bonjour.txt"
+		fs := newTestFs(t, func(fs afero.Fs) error {
+			return afero.WriteFile(fs, testFilePath, []byte("hello"), 0o644)
+		})
+		runtime.VU.InitEnvField.FileSystems["file"] = fs
+
+		_, err = runtime.RunOnEventLoop(fmt.Sprintf(`
+			fs.open(%q).then(file => {
```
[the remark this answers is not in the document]
oleiade: I'll update it in the base `experimental/fs` branch indeed 👍🏻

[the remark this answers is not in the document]
oleiade: I have updated the base branch, squashed the changes, and they should be reflected in this one too now 👍🏻

js/modules/k6/experimental/fs/module.go
oleiade: That's very insightful and a great catch, indeed. I hadn't considered that concurrency issue.

I align with you and would rather not block the feature on the resolution of this issue, but I would consider it a blocker for a stable release of the module 👍🏻 

No apparent solutions come to my mind at the moment, and I will look at Deno's and Node's implementation to find out how they handled it.

In the meantime, I have created a [dedicated issue to keep track of it](https://github.com/grafana/k6/issues/3354) 🙇🏻

codebien: >from js code is big no-no for any stable code.

If we evaluate this use case so important then I disagree with the following sentence.

>I align with you and would rather not block the feature on the resolution of this issue, but I would consider it a blocker for a stable release of the module 👍🏻

We can avoid blocking this specific PR but it should have some answer before we launch the experimental module. The reason is that if we evaluate this use case so important then in the eventual case that we can't fix it with the current solution then we will have to waste a bunch of stuff.

We should have at least a rough idea of what could be the solution.

oleiade: I have a potential solution in mind that I'll post in the dedicated issue 👍🏻

olegbespalov (approved):

olegbespalov:
Just for the record.

I'm OK with merging this and working on the fix in a separate PR. The tiny commit that could be done is to write the test case, which should catch the bug and `t.Skip` it for now.

My rationale is that merging this will put the focus on fixing the bug. And we still have plenty of time until the release; in the worst case, we could be transparent that a known issue exists. However, some of the users of this experiment already could try to use it and provide feedback.

js/modules/k6/experimental/fs/module.go
oleiade: This has now been addressed by #3420 🙇🏻

js/modules/k6/experimental/fs/module.go
```diff
@@ -173,3 +178,158 @@ func (f *File) Stat() *goja.Promise {
...
+			//   2. This approach ensures that while the file read operation can proceed
+			//   asynchronously, any side effects that might interfere with the JS runtime
+			//   are executed in a controlled and sequential manner on the main thread.
+			callback(func() error {
+				_ = copy(intoBytes, buffer)
+				resolveFunc(n)
+				return nil
+			})
+			return
+		}
```
mstoykov, line 257: Currently, if there is a read with big enough buffer to read the whole file +1 byte - the buffer will not be updated.

```javascript
import fs from "k6/experimental/fs"
let f;
fs.open("./numbers.txt").then((s) => f = s);
	let b = new Uint8Array(12); // works with 11 
	let n = await f.read(b)
	console.log(n)
	console.log(b)
}
```

With a numbers.txt

```
0123456789
```

Looking at the code and what the remaining one do I think that the easiest solution is to just
1. check `n == 0` here
2. not resolve here, but go to the err code and let it does it. (checking for nil as well)
3. Actually wrap the whole code after the `Read` in callback and use the original promise's resolveFunc and rejectFUnc directly.
4. due to the above -you no longer need the wrappers around resolve and rejectFunc as all calls are now either in a callback or on the eventloop already (before we return from the parent function)

Also obviously a test for this will be nice

mstoykov (changes requested):

js/modules/k6/experimental/fs/module.go
oleiade: Good catch indeed!

I have addressed the issue, and added a dedicated test in the following commit: f5b47a21dc72a88a5862639c949af0e66cf6fad7

Let me know if that works for you 🙇🏻

js/modules/k6/experimental/fs/module.go
```diff
@@ -198,37 +198,32 @@ func (f *File) Read(into goja.Value) *goja.Promise {
...
 
 	// We expect the into argument to be a `Uint8Array` instance
 	intoObj := into.ToObject(f.vu.Runtime())
 	uint8ArrayConstructor := f.vu.Runtime().Get("Uint8Array")
 	if isUint8Array := intoObj.Get("constructor").SameAs(uint8ArrayConstructor); !isUint8Array {
-		reject(newFsError(TypeError, "read() failed; reason: into argument must be a Uint8Array"))
+		callback(func() error {
+			rejectFunc(newFsError(TypeError, "read() failed; reason: into argument must be a Uint8Array"))
+			return nil
+		})
```
mstoykov, line 216: You don't need to use `callback` here you are already on the eventloop.

But you should move the `RegisterCallback` down to just before the `go func()` so it doesn't block

mstoykov: nit: Also given the other changes I would rename `resolveFunc` to `resolve` and same for `rejectFunc`

oleiade: TIL! Wilco 👍🏻

js/modules/k6/experimental/fs/file.go
```diff
@@ -26,3 +37,106 @@ type FileInfo struct {
...
+
+	// SeekModeCurrent seeks relative to the current offset.
+	SeekModeCurrent
+
+	// SeekModeEnd seeks relative to the end of the file.
+	//
+	// When using this mode the seek operation will move backwards from
+	// the end of the file.
+	SeekModeEnd
+)
```
mstoykov, line 138: nit: deno has Start at `0` and I do expect at least some code be just written with `0` .

js/modules/k6/experimental/fs/module.go
```diff
@@ -173,3 +178,150 @@ func (f *File) Stat() *goja.Promise {
...
+	buffer := make([]byte, len(intoBytes))
+
+	// We register a callback to be executed by the VU's runtime.
+	// This ensures that the modification of the JS runtime's `into` buffer
+	// occurs on the main thread, during the promise's resolution.
+	callback := f.vu.RegisterCallback()
+	go func() {
+		n, readErr := f.file.Read(buffer)
+		callback(func() error {
+			_ = copy(intoBytes[0:n], buffer)
```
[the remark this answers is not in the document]
oleiade: This is to follow Deno's behavior, which leaves the buffer in the state it found it in. Thus if you read only two bytes into a non-empty/non-zeroed out buffer of size 4 (which was already prefilled with some bytes from a previous read), the two last bytes should remain identical to before the `read`.

This was the initial reason for copying the content of the `into` into the buffer we read into, but as we have removed it, this was the best I could think of achieving that 🤝 Can you think of/prefer another way of achieving the same result?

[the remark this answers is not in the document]
oleiade: I'll merge this now to end my day on a positive note, and feel free to bring this up again in the `open` PR if you have more thoughts on this 🤝

Issue #3149, opened 2023-06-27
Implement the File API's `File.close` method


oleiade:

oleiade:
We have discussed this topic heavily with @codebien and concluded that in the current state of k6, introducing this API wouldn't bring outstanding benefits and probably wouldn't lead to a better user experience either.

The current implementation of the `fs` module does not rely on file descriptors under the hood but rather behaves like a filesystem cache, and the benefit of closing files, for that matter, is not immediately apparent.

We have looked into introducing reference-counted file opening/closing. Still, because we are constrained to opening them in the init context at the moment, the performance benefits we wanted to introduce by doing so are unfortunately canceled. 

As a result we decided to put the introduction of this API on hold until we have a good enough reason to do so, both from the technical and/or the user-experience perspective. A good technical reason to introduce it in the future might come as a result of #1079.

PR #3314, opened 2023-08-30
[fs] Add the `file.close` method to the `fs` module (3/3)

oleiade:
## What?

This Pull Request builds (and is based) upon https://github.com/grafana/k6/issues/3142 and adds a `close` method to the `File` object exposed by the `k6/experimental/fs` module.

The close method currently does not perform any operations. This is intentional.

### Rationale Behind No-Op Close Method:

- **User Expectations**: During our [design discussions](https://github.com/grafana/k6/blob/master/docs/design/019-file-api.md), we identified that users would likely anticipate the presence of a `close` method for consistency. Thus, we're aligning with these expectations.
- **Futureproofing**: Although the method is a no-op now, we might consider enriching its functionality in the future.

### Why Not Implement Functionality Now?

Certainly, functionality can be added. As mentioned in [#3142](https://github.com/grafana/k6/issues/3142), we maintain a file cache that users access through `File` instance pointers. This ensures a more efficient memory utilization across various VUs, compared to our existing `open` method.

Envision a scenario where only one out of several user scenarios opens a file. In such a case, tracking how many VUs haven't closed the file (via reference counting) becomes feasible. When the count hits zero, the file content can be purged from memory.

I've crafted a [proof of concept for this approach](https://github.com/grafana/k6/tree/experimental/fs-ref-counted-close). However, a current k6 limitation hinders its realization. Presently, k6 doesn't support opening files within the VU context (refer to [#3020](https://github.com/grafana/k6/issues/3020)). This implies that even with reference counting via `file.close`, we'd always have an uncollected reference, preventing the count from reaching zero.

## Checklist

## Related PR(s)/Issue(s)

ref #3142 
ref #3149
ref #3165

### DIFF
--- a/examples/experimental/fs/fs.js
+++ b/examples/experimental/fs/fs.js
@@ -50,3 +50,7 @@ export default async function () {
 	// Seek back to the beginning of the file
 	await file.seek(0, SeekMode.Start);
 }
+
+export async function teardown() {
+	file.close();
+}
--- a/js/modules/k6/experimental/fs/file.go
+++ b/js/modules/k6/experimental/fs/file.go
@@ -3,6 +3,7 @@ package fs
 import (
 	"io"
 	"path/filepath"
+	"sync/atomic"
 )
 
 // file is an abstraction for interacting with files.
@@ -14,6 +15,9 @@ type file struct {
 
 	// offset holds the current offset in the file
 	offset int
+
+	// closed indicates whether the file has been closed
+	closed atomic.Bool
 }
 
 // Stat returns a FileInfo describing the named file.
@@ -125,3 +129,14 @@ const (
 	// the end of the file.
 	SeekModeEnd
 )
+
+// Close closes the file instance, and marks it as closed.
+func (f *file) Close() error {
+	if f.closed.Load() {
+		return newFsError(BadResourceError, "file already closed")
+	}
+
+	f.closed.Store(true)
+
+	return nil
+}
--- a/js/modules/k6/experimental/fs/file.go
+++ b/js/modules/k6/experimental/fs/file.go
@@ -21,9 +21,13 @@ type file struct {
 }
 
 // Stat returns a FileInfo describing the named file.
-func (f *file) stat() *FileInfo {
+func (f *file) stat() (*FileInfo, error) {
+	if f.closed.Load() {
+		return nil, newFsError(BadResourceError, "cannot stat closed file")
+	}
+
 	filename := filepath.Base(f.path)
-	return &FileInfo{Name: filename, Size: len(f.data)}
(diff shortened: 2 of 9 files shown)


codebien:
>This implies that even with reference counting via file.close, we'd always have an uncollected reference, preventing the count from reaching zero.

@oleiade We are able to have `k6/execution.vu.idInTest` so we should already have a function for getting the accurate counter. Or Can you explain more, please?

oleiade:
Hey @codebien 👋🏻 

I've spent a couple of hours more on my ref-counting POC to see if there was an alternative, as you suggested, to make sure we would release the memory once we don't need it anymore.

I made some progress and was able to release the memory as soon as all VUs referencing it were done indeed. 

**However**, I encountered another issue, which I'm not sure can be addressed, nor if it should: `handleSummary` evaluates the init context, which in the current context will make the module reload the file in memory if it's not present yet. Effectively defeating the initial goal of reference counting.

Eventually, once we have the ability to open files in the VU stage as per the design proposal I submitted internally, it would be worth attempting to resume this approach (if the file is opened in the VU stage, the `handleSummary` wouldn't expect to have access to it, and with a VU id based reference counting, we would probably be able to clear the file from memory safely). But I think it's not worth it in the current state of things.

What do you think? Do you maybe have other approaches that might be worth considering?

codebien (commented):
Just as a reminder note. We have discussed it internally and we've set a potential solution. @oleiade will write a proposal for integrating the current design doc.

examples/experimental/fs/fs.js
```diff
@@ -50,3 +50,7 @@ export default async function () {
 	// Seek back to the beginning of the file
 	await file.seek(0, SeekMode.Start);
 }
+
+export async function teardown() {
+	file.close();
+}
```
codebien, line 56: In a distributed mode, what should the expectation be? Teardown runs only once per Test instead of the init context that tuns once per VU.

js/modules/k6/experimental/fs/file.go
```diff
@@ -125,3 +141,14 @@ const (
...
 )
+
+// Close closes the file instance, and marks it as closed.
+func (f *file) Close() error {
+	if f.closed.Load() {
+		return newFsError(BadResourceError, "file already closed")
+	}
+
+	f.closed.Store(true)
+
```
codebien, line 152:
```suggestion
	f.closed.Store(true)
	f.data = nil
```
I think here we may release the pointer to the data. We don't expect to access them anymore, and in the bad case there is an attempt we can just check the closed variable and return an error saying _the file has been closed_.

oleiade:
We have discussed this feature (`fs.close`) in depth with @codebien, and concluded that in the current state of things, we prefer not to merge this and delay introducing the `fs.close` feature to a later point.

PR #3317, opened 2023-09-01
cloud: min_resolution field is required

codebien:
## What?

The min_resolution has been set as mandatory after we have changed the default min resolution to increase the accuracy. It also address request changes from the parent PR https://github.com/grafana/k6/pull/3302#pullrequestreview-1604230323

## Why?

In this way, we can use just a value instead of a pointer and hopefully reduce some GC pressure.

## Related PR(s)/Issue(s)

https://github.com/grafana/k6/pull/3302


output/cloud/expv2/hdr.go
```diff
@@ -159,7 +159,7 @@ func histogramAsProto(h *histogram, time int64) *pbcloud.TrendHdrValue {
 	}
 	// We don't expect to change the minimum resolution at runtime
 	// so it is safe use directly a pointer without creating a copy
-	hval.MinResolution = &h.MinimumResolution
+	hval.MinResolution = h.MinimumResolution
```
olegbespalov, line 162: :heart: but please also clean up the comment which became obsolete

output/cloud/expv2/pbcloud/metric.proto
```diff
@@ -140,13 +140,14 @@ message TrendHdrValue {
   double min_value = 7;
   double max_value = 8;
 
+  // histogram parameter - value multiplier aka smallest value
```
olegbespalov, line 143: I don't think that it's critical since, in any case, the number is essential, but what's the reason for moving it?

mstoykov (approved):

output/cloud/expv2/pbcloud/metric.proto
codebien: Good catch, I forgot the order

output/cloud/expv2/hdr.go

olegbespalov (approved):

oleiade (approved):

PR #3320, opened 2023-09-04
docker: Build multi platform images

### LINKED ISSUE #3301: publish arm64 images to Docker hub
### Brief summary

so I found #3015, which seems to imply that the arm64 image is buildable, but when I goto https://hub.docker.com/r/grafana/k6/tags there's only a `linux/amd64` image.

### k6 version

any

### OS

OSX, Docker Desktop, Ubuntu, 

_No response_

go to https://hub.docker.com/r/grafana/k6/tags - at this point 0.46.0 is the current release, and unlike (for eg https://hub.docker.com/r/grafana/grafana/tags ), the k6 image is not multi-arch

### Expected behaviour

the desired nirvana would be that all the architectures work - in my case, it'd be nice for my arm64 systems to be happy without needing the cross-arch support :)

### Actual behaviour

multi-arch images, or a nice sentence of why it doesn't make sense to do it atm, and who could help make it happen / change that

codebien:
## What?

Support docker builds for arm64

## Why?

Users are requesting it more and more times 

## Related PR(s)/Issue(s)

Closes #3301

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -140,9 +140,18 @@ jobs:
         with:
           fetch-depth: 0
       - name: Build
-        run: docker build --target release -t $DOCKER_IMAGE_ID .
+        run: |
+          docker buildx create \
+            --name multibuilder \
+            --platform linux/amd64,linux/arm64 \
+            --bootstrap --use
+          docker buildx build \
+            --platform linux/amd64,linux/arm64 \
+            -t $DOCKER_IMAGE_ID .
       - name: Check
         run: |
+            docker buildx build --load -t $DOCKER_IMAGE_ID .
+            # Assert that simple cases works for the new built image
             docker run $DOCKER_IMAGE_ID version
             docker run $DOCKER_IMAGE_ID --help
             docker run $DOCKER_IMAGE_ID help
@@ -154,8 +163,6 @@ jobs:
             docker run $DOCKER_IMAGE_ID pause --help
             docker run $DOCKER_IMAGE_ID resume --help
 
-      - name: Build browser-enabled image
-        run: docker build -t $DOCKER_IMAGE_ID:with-browser --target with-browser .
       - name: Log into ghcr.io
         if: ${{ github.ref == 'refs/heads/master' || startsWith(github.ref, 'refs/tags/v') }}
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -140,103 +140,122 @@ jobs:
...
         if: ${{ github.ref == 'refs/heads/master' }}
         run: |
           echo "Publish as ghcr.io/$GHCR_IMAGE_ID:$VERSION"
-          docker tag "$DOCKER_IMAGE_ID" "ghcr.io/$GHCR_IMAGE_ID:master"
-          docker push "ghcr.io/$GHCR_IMAGE_ID:master"
-          docker tag "$DOCKER_IMAGE_ID:with-browser" "ghcr.io/$GHCR_IMAGE_ID:master-with-browser"
-          docker push "ghcr.io/$GHCR_IMAGE_ID:master-with-browser"
+          docker buildx build --push \
+            --platform linux/amd64,linux/arm64 \
+            -t ghcr.io/$GHCR_IMAGE_ID:master .
```
mstoykov, line 174: I guess this - using `docker buildx build --push` to push the images is the difference with what #3128 reverted ? 

Can we have this tested with some tag and have someone on a macos + arm test it? Or have we done this already?

.github/workflows/build.yml
```diff
@@ -140,103 +140,122 @@ jobs:
...
       - name: Build
-        run: docker build --target release -t $DOCKER_IMAGE_ID .
+        run: |
+          docker buildx create \
+            --name multibuilder \
+            --platform linux/amd64,linux/arm64 \
+            --bootstrap --use
+          docker buildx build \
+            --platform linux/amd64,linux/arm64 \
+            -t $DOCKER_IMAGE_ID .
```
mstoykov, line 150: I am not a docker person and definitely not a buildx person, but the previous attempt at this also had changse to the docker file - https://github.com/grafana/k6/pull/3015

Are they not needed now? Were they never needed?

.github/workflows/build.yml
codebien: I could split it in a PR adding only ARM64 for master-with-browser, and then they could test it after we've merged. WDYT? We have tested the build not the push.

.github/workflows/build.yml
```diff
@@ -140,103 +140,122 @@ jobs:
...
             docker run $DOCKER_IMAGE_ID version
             docker run $DOCKER_IMAGE_ID --help
-            docker run $DOCKER_IMAGE_ID help
-            docker run $DOCKER_IMAGE_ID run --help
-            docker run $DOCKER_IMAGE_ID inspect --help
-            docker run $DOCKER_IMAGE_ID status --help
-            docker run $DOCKER_IMAGE_ID stats --help
-            docker run $DOCKER_IMAGE_ID scale --help
-            docker run $DOCKER_IMAGE_ID pause --help
-            docker run $DOCKER_IMAGE_ID resume --help
```
mstoykov, line 155: I am not against dropping some of those, but is there any reason it is done in this PR?

.github/workflows/build.yml
```diff
@@ -140,103 +140,122 @@ jobs:
...
-      - name: Build browser-enabled image
-        run: docker build -t $DOCKER_IMAGE_ID:with-browser --target with-browser .
       - name: Log into ghcr.io
         if: ${{ github.ref == 'refs/heads/master' || startsWith(github.ref, 'refs/tags/v') }}
         run: |
+          if [[ `git status --porcelain` ]]; then
+            echo "The repository has unexpected changes identified from the git history."\
+            "Force the fail as the job can't guarantee that the code is not changed between the previous build and the next build+push."
+            exit 1
+          fi
```
mstoykov, line 165: does `docker buildx build` touches local directory?

This seems strange. I am not against us checking, just seems strange for us to need it given that there is nothing really that we do that is expected to touch local directory here.

.github/workflows/build.yml
```diff
@@ -140,103 +140,122 @@ jobs:
...
-            docker push "$LI_DOCKER_IMAGE_ID:latest"
+            echo "Publish to Docker Hub as $DOCKER_IMAGE_ID:latest and $LI_DOCKER_IMAGE_ID:latest"
+            docker buildx build --push \
+              --platform linux/amd64,linux/arm64 \
+              -t $DOCKER_IMAGE_ID:latest .
+            docker buildx build --push \
+              --target with-browser \
+              --platform linux/amd64,linux/arm64 \
+              -t $DOCKER_IMAGE_ID:latest-with-browser .
+            docker buildx build --push \
```
mstoykov, line 256: nit: can we have the same comment about loadimpact images being deprecated here as well.

.github/workflows/build.yml
mstoykov: You can also copy one of the sections below and make it push a build for just any commit - that should build and push it when you push to commit and we can test that.

.github/workflows/build.yml
codebien: They are the support for [cross-compilation](https://docs.docker.com/build/building/multi-platform/#cross-compilation), if I'm understanding buildx in the right way without using it we are instead emulating the platform. It has the cons of being slower. I would prefer if we could keep the PR separated.

>Using an emulator this way is very easy. We don’t need to modify our Dockerfile at all and can build for multiple platforms automatically. But it doesn’t come without downsides. The binaries running this way need to constantly convert their instructions between architectures and therefore don’t run with native speed. Occasionally you might also find a case that [triggers a bug](https://github.com/docker/buildx/issues?q=is%3Aopen+is%3Aissue+label%3Aqemu) in the emulation layer.

So, I'm for merging this PR and check that it builds fine then I will open a new PR specifically for the dockerfile and enabling the cross-compilation.

.github/workflows/build.yml
codebien: It was my plan, I forgot to split it.

codebien:
@mstoykov @olegbespalov Just for sharing other plans. I would like to apply in a dedicated PR after this how we are doing the push. We now are processing sequentially by registry. I would prefer if we do it instead by kind of build. I take the assumption that the risk of failure is higher between the different versions than the same build pushed across different registries.
So, if one build fails, all the steps before are already fine across all the registries, so we could skip them and rebuild only the failed+missing steps with the guarantee that the same version is the same code across different registries.

Essentially, I'm suggesting doing the equivalent solution: https://docs.docker.com/build/ci/github-actions/push-multi-registries/

.github/workflows/build.yml
codebien: Tested, the browser told me that is works

.github/workflows/build.yml
codebien: I removed, this was mostly a request by Andrey. But I think we can remove and add in the future when he is back.

olegbespalov (approved):

PR #3322, opened 2023-09-05
Use the commit date for build metadata

codebien:
## What?

Use the date of the latest commit in the working tree during the build process.

Note: GitHub uses the COMMIT DATE and not the AUTHOR DATE as you see in `git log` classic dump.

## Why?

Previously, the date of the build was generated at runtime blocking the possibility of reproducible builds.

## Related PR(s)/Issue(s)

#1996

### DIFF
--- a/build-release.sh
+++ b/build-release.sh
@@ -10,7 +10,7 @@ export OUT_DIR="${1-dist}"
 export VERSION=${2:-$(git describe --tags --always --dirty)}
 
 # To overwrite the version details, pass something as the third arg. Empty string disables it.
-export COMMIT_DATE=$(git log -1 --format=%at $VERSION | { read d ; date -u +"%FT%T%z" -d @"$d" ; })
+export COMMIT_DATE=$(git log -1 --format=%ct $VERSION | { read d ; date -u +"%FT%T%z" -d @"$d" ; })
 export VERSION_DETAILS=${3-"${COMMIT_DATE}/$(git describe --tags --always --long --dirty)"}
 set +x
 
--- a/build-release.sh
+++ b/build-release.sh
@@ -10,7 +10,8 @@ export OUT_DIR="${1-dist}"
 export VERSION=${2:-$(git describe --tags --always --dirty)}
 
 # To overwrite the version details, pass something as the third arg. Empty string disables it.
-export VERSION_DETAILS=${3-"$(date -u +"%FT%T%z")/$(git describe --tags --always --long --dirty)"}
+export COMMIT_DATE=$(git log -1 --format=%at $VERSION | { read d ; date -u +"%FT%T%z" -d @"$d" ; })
+export VERSION_DETAILS=${3-"${COMMIT_DATE}/$(git describe --tags --always --long --dirty)"}
 set +x
 
 build() {


codebien:
It seems to works as you can see from the Build check:
```
Building platform: linux-amd64 ( GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go build -o dist/k6-v0.46.0-31-gc68b379b-linux-amd64/k6 -trimpath -ldflags \
-X go.k6.io/k6/lib/consts.VersionDetails=2023-09-05T17:42:51+0000/v0.46.0-31-gc68b379b )
```
I've rebuilt it again today to confirm that the date doesn't change. The 2nd run: https://github.com/grafana/k6/actions/runs/6087972451/job/16536142053

build-release.sh
```diff
@@ -10,7 +10,8 @@ export OUT_DIR="${1-dist}"
 export VERSION=${2:-$(git describe --tags --always --dirty)}
 
 # To overwrite the version details, pass something as the third arg. Empty string disables it.
-export VERSION_DETAILS=${3-"$(date -u +"%FT%T%z")/$(git describe --tags --always --long --dirty)"}
+export COMMIT_DATE=$(git log -1 --format=%at $VERSION | { read d ; date -u +"%FT%T%z" -d @"$d" ; })
```
olegbespalov, line 13: How about using `--format=%aI`, it could help us have no conversion magic?

```bash
git log -1 --format=%aI 48dddc3005c9fbc2fd9a52990560790535a29f69
2023-09-05T15:27:34+02:00
```

codebien: I tried but it uses local time, so in this way, we would have a dependency on the enviroment's timezone. And, I didn't find a way to use `git log` for dumping an ISO8601 date time in UTC.

olegbespalov: I see. Then yeah makes sense :+1: 

And the last one juts double-checking if we want to use the author datetime instead of the commuter datetime https://docs.github.com/en/account-and-profile/setting-up-and-managing-your-github-profile/managing-contribution-settings-on-your-profile/troubleshooting-commits-on-your-timeline#how-github-uses-the-git-author-date-and-commit-date ?

olegbespalov (approved):

build-release.sh
olegbespalov: Actually :arrow_up: make nos sense since our main use case where version is the tag :thinking:

codebien:
```suggestion
export COMMIT_DATE=$(git log -1 --format=%ct $VERSION | { read d ; date -u +"%FT%T%z" -d @"$d" ; })
```
Instead, I think you're right it matters. If GitHub builds in one way and on the local build we print in another way then we are again environment-dependent. Let's force to use the COMMIT_DATE all the time.

codebien:
The version printed by `git describe --tags --always --long --dirty` on my machine is different from what I see printend in the action. :thinking:

@olegbespalov If you fetch this branch what do you get running the command?

olegbespalov:
```
git describe --tags --always --long --dirty
v0.46.0-31-g732549d7
```

codebien:
The build stage says https://github.com/grafana/k6/actions/runs/6096099795/job/16540972908?pr=3322  `+ export VERSION=v0.46.0-32-g07b68e19`

Any idea why? :thinking:

olegbespalov:
I belive it's because it's pull request https://stackoverflow.com/questions/68061051/get-commit-sha-in-github-actions

build-release.sh
mstoykov: To be honest ... I am kind of for dropping this :shrug: 

We have the commit - that is all we need. Or we have a really well tagged build - so again the date of the build doesn't really matter.

To be honest I really was hoping that we can get rid of this all together and just rely on ReadBuildInfo, unfortunately `go build` in practice never populates this (see [here](https://github.com/golang/go/issues/51279#issuecomment-1048196896) for reasons). 

So in practice that is not ... possible unless we figure out how to make `go install` compile. Which works *except* that:
1. `-o` does not [work](https://github.com/golang/go/issues/44469#issuecomment-1656803740) :( 
2. `GOBIN` setting does [not work](https://github.com/golang/go/issues/57485) with `GOOS` and friends 


So ... in theory we can have `go install ...; mv $GOBIN/k6 ... somehwere` 

But this seems ... bad - I currently hope that the 57485 will be fixed :crossed_fingers: or `go build` will be allowed to buildinfo.

Currently if you do `go install` vs what we have as binary the output is different:

```
[mstoykov@fedora k6]$ go install go.k6.io/k6@v0.46.0
[mstoykov@fedora k6]$ k6 version
k6 v0.46.0 (v0.46.0, go1.21.1, linux/amd64)
[mstoykov@fedora k6]$ k6.v0.46.0 version
k6 v0.46.0 (2023-08-14T13:23:26+0000/v0.46.0-0-gcbd9e9ad, go1.20.7, linux/amd64)
```

And to be honest I find *our* build to be worse :shrug: 

so .. to reiterate ... can we drop the date at least?

build-release.sh
```diff
@@ -10,7 +10,8 @@ export OUT_DIR="${1-dist}"
 export VERSION=${2:-$(git describe --tags --always --dirty)}
 
 # To overwrite the version details, pass something as the third arg. Empty string disables it.
-export VERSION_DETAILS=${3-"$(date -u +"%FT%T%z")/$(git describe --tags --always --long --dirty)"}
+export COMMIT_DATE=$(git log -1 --format=%ct $VERSION | { read d ; date -u +"%FT%T%z" -d @"$d" ; })
```
mstoykov, line 13:
```suggestion
export COMMIT_DATE=$(git log -1 --format=%ct HEAD | { read d ; date -u +"%FT%T%z" -d @"$d" ; })
```

I feel like this is what you meant ?

You also can just drop `HEAD` I guess.

codebien:
Closing in favor of https://github.com/grafana/k6/pull/3327

PR #3323, opened 2023-09-06
docker: Multiplatform cross-compilation

codebien:
## What?

Move to use the cross compilation for docker multi-arch builds.

## Why?

https://www.docker.com/blog/faster-multi-platform-builds-dockerfile-cross-compilation-guide/

## Related PR(s)/Issue(s)

#3301

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@ WORKDIR $GOPATH/src/go.k6.io/k6
 COPY . .
 ARG TARGETOS TARGETARCH
 RUN apk --no-cache add git=~2
-RUN CGO_ENABLED=0 GOOS=$TARGETOS GOARCH=$TARGETARCH go build -a -trimpath -o /usr/bin/k6
+RUN CGO_ENABLED=0 GOOS=$TARGETOS GOARCH=$TARGETARCH go build -trimpath -o /usr/bin/k6
 
 # Runtime stage
 FROM alpine:3.18 as release
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,9 @@
-FROM golang:1.20-alpine3.18 as builder
+FROM --platform=$BUILDPLATFORM golang:1.20-alpine3.18 as builder
 WORKDIR $GOPATH/src/go.k6.io/k6
 COPY . .
+ARG TARGETOS TARGETARCH
 RUN apk --no-cache add git=~2
-RUN CGO_ENABLED=0 go install -a -trimpath -ldflags "-s -w -X go.k6.io/k6/lib/consts.VersionDetails=$(date -u +"%FT%T%z")/$(git describe --tags --always --long --dirty)"
+RUN CGO_ENABLED=0 GOOS=$TARGETOS GOARCH=$TARGETARCH go build -a -trimpath -o /usr/bin/k6
 
 # Runtime stage
 FROM alpine:3.18 as release
@@ -10,7 +11,7 @@ FROM alpine:3.18 as release
 # hadolint ignore=DL3018
 RUN apk add --no-cache ca-certificates && \
     adduser -D -u 12345 -g 12345 k6
-COPY --from=builder /go/bin/k6 /usr/bin/k6
+COPY --from=builder /usr/bin/k6 /usr/bin/k6
 
 USER k6
 WORKDIR /home/k6
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,4 @@ tests :
 
 check : ci-like-lint tests
 
-container:
-	docker build --rm --pull --no-cache -t grafana/k6 .
-
-.PHONY: build format ci-like-lint lint tests check container
+.PHONY: build format ci-like-lint lint tests check


Dockerfile
```diff
@@ -1,16 +1,19 @@
-FROM golang:1.20-alpine3.18 as builder
+FROM --platform=$BUILDPLATFORM golang:1.20-alpine3.18 as builder
 WORKDIR $GOPATH/src/go.k6.io/k6
-COPY . .
+ARG TARGETOS TARGETARCH
 RUN apk --no-cache add git=~2
-RUN CGO_ENABLED=0 go install -a -trimpath -ldflags "-s -w -X go.k6.io/k6/lib/consts.VersionDetails=$(date -u +"%FT%T%z")/$(git describe --tags --always --long --dirty)"
+RUN --mount=target=. \
+    --mount=type=cache,target=/root/.cache/go-build \
+    --mount=type=cache,target=/go/pkg \
```
codebien, line 7:
```suggestion
```
We don't need it, we already include the vendor folder

Dockerfile
```diff
@@ -1,16 +1,18 @@
-FROM golang:1.20-alpine3.18 as builder
+FROM --platform=$BUILDPLATFORM golang:1.20-alpine3.18 as builder
 WORKDIR $GOPATH/src/go.k6.io/k6
-COPY . .
+ARG TARGETOS TARGETARCH
```
olegbespalov, line 3: Not sure if this is a big issue, but just stressing that this change makes the Dockerfile unbuildable without providing these args and think like ` docker build . -t k6-test` won't work anymore :cry:

codebien: >docker build . -t k6-test

Do we need to support it? I guess these PRs are setting the requirement for `buildx`. Do you see issues in doing it?

olegbespalov: It was just nice to have the ability to build an image for local testing without digging into what you should pass.  On the other side, it's a rare case. 

However, maybe a comment with the command or better fixing the Makefile's target (which also) resolves my concern.

Dockerfile
```diff
@@ -1,16 +1,18 @@
...
 RUN apk --no-cache add git=~2
-RUN CGO_ENABLED=0 go install -a -trimpath -ldflags "-s -w -X go.k6.io/k6/lib/consts.VersionDetails=$(date -u +"%FT%T%z")/$(git describe --tags --always --long --dirty)"
+RUN --mount=target=.,readonly \
```
mstoykov, line 5: On systems with selinux (fedora, redhat, other redhat derivatives :P), you need `,Z` after this.

I guess this makes it way faster then `COPY` from my test? But I am not certain if the `,Z` works on not selinux, so it might be better if we keep `COPY` if it doesn't.

Dockerfile
mstoykov: This actually works for me ... with `podman`. 

My problem was https://github.com/grafana/k6/pull/3323/files#r1321478143 

What error did you get ?

codebien: I get the same error as @olegbespalov 

```
docker build --rm -t testbuildclassic  .
Sending build context to Docker daemon  93.62MB
Step 1/23 : FROM --platform=$BUILDPLATFORM golang:1.20-alpine3.18 as builder
failed to parse platform : "" is an invalid component of "": platform specifier component must match "^[A-Za-z0-9_-]+$": invalid argument
```

Dockerfile
codebien: Do you get the same error if you build via `buildx`?

Dockerfile
mstoykov: From the google searches I did ... it seems like this should work for you too :thinking: although maybe `moby` and what `docker` is for most people is now very different :shrug: . 

Does this work if you run `DOCKER_BUILDKIT=1 docker build ...` ?

codebien: I guess we are getting the error because we have a version installed of Docker Engine `< 23.0`. @olegbespalov do you confirm that it is the case for you as well?

olegbespalov: That's correct :+1:

Dockerfile
codebien: >The --mount flag does not support z or Z options for modifying selinux labels.

Also, set `Z` doesn't seem to be a pattern to encourage so considering this is not really a requirement and the negative impact it could have I'm going to roll back this change.

codebien: @mstoykov I expect you have the same issue with the go build cache folder. Right?

mstoykov: I removed both .. to be honest - but it shouldn't matter in that case as the problem is that I am mounting filesystem that is selinux protected and don't do anything to make it work. While I think the cache is just a container volume. 

Testing it - it didn't error out so it worked :+1: 

No idea how useful it is to add it though. Most people will not be building this constantly and k6 compilation times are already pretty okay. I also expect this will mostly be run by github actions. And I will be plesently surprised if this helps there at all if we constantly bounce between runners.

Dockerfile
```diff
@@ -1,16 +1,18 @@
...
+    --mount=type=cache,target=/root/.cache/go-build \
+	CGO_ENABLED=0 GOOS=$TARGETOS GOARCH=$TARGETARCH go build -a -trimpath -o /usr/bin/k6
```
mstoykov, line 7: Additionalyl I think `-a` compeltely negates the fact that we are havging a cache of the builds as `-a` basically saays to ignore that and rebuild

olegbespalov (dismissed):

Dockerfile
```diff
@@ -1,16 +1,17 @@
-FROM golang:1.20-alpine3.18 as builder
+FROM --platform=$BUILDPLATFORM golang:1.20-alpine3.18 as builder
 WORKDIR $GOPATH/src/go.k6.io/k6
 COPY . .
+ARG TARGETOS TARGETARCH
 RUN apk --no-cache add git=~2
-RUN CGO_ENABLED=0 go install -a -trimpath -ldflags "-s -w -X go.k6.io/k6/lib/consts.VersionDetails=$(date -u +"%FT%T%z")/$(git describe --tags --always --long --dirty)"
+RUN CGO_ENABLED=0 GOOS=$TARGETOS GOARCH=$TARGETARCH go build -a -trimpath -o /usr/bin/k6
```
mstoykov, line 6: nit: Is there a reason we want the `-a`?  I think it rebuilds the stdlib which seems .. pointless :shrug:

mstoykov (dismissed):
LGTM! Just one nit

Dockerfile
codebien: I was commenting the same, but you were faster!!

I researched about it and I didn't find any reason, I guess there was some historical reason for it considering how old this command is.
I'm going to remove it as it isn't needed at the moment.

olegbespalov (approved):

PR #3325, opened 2023-09-06
docker: Use go build command

codebien:
## What?

- Uses `go build`
- https://tip.golang.org/doc/go1.20#cgo  It drops CGO_ENABLED=0 as it is now the default (TODO: do the same for the build-release.sh). EDIT: I decided to postpone this change after v0.47.0 release. We already applied many changes to our release process for v0.47.0 release.

## Why?

Keep build processes consistent

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,7 @@ FROM golang:1.20-alpine3.18 as builder
 WORKDIR $GOPATH/src/go.k6.io/k6
 COPY . .
 RUN apk --no-cache add git=~2
-RUN CGO_ENABLED=0 go install -a -trimpath -ldflags "-s -w -X go.k6.io/k6/lib/consts.VersionDetails=$(date -u +"%FT%T%z")/$(git describe --tags --always --long --dirty)"
+RUN CGO_ENABLED=0 go build -trimpath -ldflags "-s -w -X go.k6.io/k6/lib/consts.VersionDetails=$(date -u +"%FT%T%z")/$(git describe --tags --always --long --dirty)" -o /go/bin/k6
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -2,7 +2,7 @@ FROM golang:1.20-alpine3.18 as builder
 WORKDIR $GOPATH/src/go.k6.io/k6
 COPY . .
 RUN apk --no-cache add git=~2
-RUN CGO_ENABLED=0 go install -a -trimpath -ldflags "-s -w -X go.k6.io/k6/lib/consts.VersionDetails=$(date -u +"%FT%T%z")/$(git describe --tags --always --long --dirty)"
+RUN go build -a -trimpath -ldflags "-s -w -X go.k6.io/k6/lib/consts.VersionDetails=$(date -u +"%FT%T%z")/$(git describe --tags --always --long --dirty)" -o /go/bin/k6
```
codebien, line 5:
```suggestion
RUN go build -trimpath -ldflags "-s -w -X go.k6.io/k6/lib/consts.VersionDetails=$(date -u +"%FT%T%z")/$(git describe --tags --always --long --dirty)" -o /go/bin/k6
```
autospot

olegbespalov (commented):
@codebien I remember that there was a reason why we started using the `go install` mentioned in https://github.com/grafana/k6/pull/3015. It's not the case anymore, right?

codebien:
This is not really required, we can keep using the same command as users do when installing from the Go toolchain so it is better to keep `go install`.

So we will be more confident that `go install go.k6.io/k6@latest` will produce mostly the same build as the k6 build process.

PR #3327, opened 2023-09-07
cmd/version: Commit hash

codebien:
## What?

Before we were injecting the version details via ldflags. go1.18 introduced the ability to detect the commit directly from the code.

An example of the new format:
```
k6 v0.46.0 (commit/a7aa07c26b, go1.21.0, linux/amd64)
```

## Why?

We benefit from it for simplifying our build process and making it more reproducible. It now prints the commit hash instead of the timestamp, in this way should be easier to use in case of troubleshooting and reports.

## Related PR(s)/Issue(s)

The original discussion where it started https://github.com/grafana/k6/pull/3322#discussion_r1318510114

### DIFF
--- a/lib/consts/consts.go
+++ b/lib/consts/consts.go
@@ -10,9 +10,6 @@ import (
 // Version contains the current semantic version of k6.
 const Version = "0.46.0"
 
-// VersionDetails can be set externally as part of the build process
-var VersionDetails = "" //nolint:gochecknoglobals
-
 // FullVersion returns the maximally full version and build information for
 // the currently running k6 executable.
 func FullVersion() string {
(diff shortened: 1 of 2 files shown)


lib/consts/consts.go
```diff
@@ -10,22 +10,45 @@ import (
...
+		return fmt.Sprintf("%s (%s)", Version, goVersionArch)
+	}
+
+	var (
+		commit string
+		dirty  bool
+	)
+	for _, s := range buildInfo.Settings {
+		switch s.Key {
+		case "vcs.revision":
```
mstoykov, line 29: We even have `vcs.time` so ... I guess we can even add that if somebody wants :shrug:

mstoykov (approved):

oleiade (approved):
This is awesome 👍🏻

PR #3328, opened 2023-09-08
Amend File API design document to address `file.Close` constraints

oleiade:
## What?

This Pull Request updates the File API design document, focusing on the limitations of the `file.Close` operation, especially when files are opened in the init context.

## Why?

The current design treats the `file.Close` operation as a non-operation, as detailed in #3314. This choice was by design, as we anticipated changing this once k6 can [open files during the VU stage](https://github.com/grafana/k6/issues/3020).

We've explored an alternative where k6 tracks how many times a file has been opened. When this count reaches zero, k6 would free up the memory holding the file content. More details can be found in the [proof-of-concept for this approach](https://github.com/grafana/k6/tree/experimental/fs-ref-counted-close/js/modules/k6/experimental/fs).

A challenge arose: opening files in the init context means the count might never reach zero. This is because the file keeps a reference until the test concludes.

After discussions with @codebien, we agreed that due to k6's lifecycle stages and the init context's frequent calls, the best approach would be to:
- Adopt the reference count approach.
- Keeping data in memory as long as the init context code hasn't closed the file to a point where zero references are left.
- Close the file and free memory only at the test's end, possibly using [the newly added event system](https://github.com/grafana/k6/pull/3112).

When #3020 is in place, our plan is to adjust this so that files opened exclusively in the VU stage can have their memory released earlier, given they are consistently closed in that stage.

## Refs

#3149
#3101 
#3314
#3020


codebien (changes requested):
@oleiade We should move this change to the `possible future improvements` section and update it by integrating https://github.com/grafana/k6/issues/3149#issuecomment-1779181153.

oleiade:
Agreed 👍🏻 

What did you refer to as the `possible future improvements` section?

codebien:
https://github.com/grafana/k6/blob/master/docs/design/019-file-api.md#possible-future-improvements

oleiade:
@codebien I see, you meant the one in the file directly 🙇🏻 Fair enough, I'll amend the document 👍🏻

oleiade:
✅

PR #3335, opened 2023-09-18
Bump future metric name length to 128

mstoykov:
This is inline with the current restriction.

The previous 63 character one came from Otel, but with https://github.com/open-telemetry/opentelemetry-specification/pull/3648 This has been bumped to 255.

Updates #3065

### DIFF
--- a/js/bundle.go
+++ b/js/bundle.go
@@ -46,8 +46,8 @@ type Bundle struct {
 // https://github.com/grafana/k6/issues/3065
 func (b *Bundle) checkMetricNamesForPrometheusCompatibility() {
 	const (
-		nameRegexString = "^[a-zA-Z_][a-zA-Z0-9_]{1,63}$"
-		badNameWarning  = "Metric name should only include ASCII letters, numbers and underscores. " +
+		nameRegexString = "^[a-zA-Z_][a-zA-Z0-9_]{1,128}$"
+		badNameWarning  = "Metric name should only include up to 128 ASCII letters, numbers and/or underscores. " +
 			"This name will stop working in k6 v0.48.0 (around December 2023)."
 	)
 
--- a/js/bundle.go
+++ b/js/bundle.go
@@ -46,6 +46,8 @@ type Bundle struct {
 // https://github.com/grafana/k6/issues/3065
 func (b *Bundle) checkMetricNamesForPrometheusCompatibility() {
 	const (
+		// The name restrictions are the union of Otel and Prometheus naming restrictions, with the length restrictions of 128
+		// coming from old k6 restrictions where character set was way bigger though.
 		nameRegexString = "^[a-zA-Z_][a-zA-Z0-9_]{1,128}$"
 		badNameWarning  = "Metric name should only include up to 128 ASCII letters, numbers and/or underscores. " +
 			"This name will stop working in k6 v0.48.0 (around December 2023)."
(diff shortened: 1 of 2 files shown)


codebien:
Apparently, there is also:

>Add "/" to valid characters for instrument names (https://github.com/open-telemetry/opentelemetry-specification/pull/3684)

Are we going to include it?

mstoykov:
That is not allowed in prometheus - so no.

js/bundle.go
```diff
@@ -46,8 +46,8 @@ type Bundle struct {
 // https://github.com/grafana/k6/issues/3065
 func (b *Bundle) checkMetricNamesForPrometheusCompatibility() {
 	const (
-		nameRegexString = "^[a-zA-Z_][a-zA-Z0-9_]{1,63}$"
-		badNameWarning  = "Metric name should only include ASCII letters, numbers and underscores. " +
+		nameRegexString = "^[a-zA-Z_][a-zA-Z0-9_]{1,128}$"
```
codebien, line 49:
```suggestion
		nameRegexString = "^[a-zA-Z_][a-zA-Z0-9_]{1,255}$"
```
Is there a specific reason behind it? If not then I would prefer if we could keep it aligned with OTel.

js/bundle.go
```diff
@@ -46,8 +46,8 @@ type Bundle struct {
...
+		badNameWarning  = "Metric name should only include upto 128 ASCII letters, numbers and/or underscores. " +
```
codebien, line 50:
```suggestion
		badNameWarning  = "Metric name should only include up to 128 ASCII characters between letters, numbers and underscores. " +
```

codebien (changes requested):

js/bundle.go
mstoykov: Yes that is the current limit in the registry. The place where this limit will go after this becomes an error instead of a warning.

codebien: My suggestion was about the size limit `s/128/255/`

mstoykov: Yes this is hte current limit - 128

https://github.com/grafana/k6/blob/981b770946f5df5d772554e5906f888ef1ca8a94/metrics/registry.go#L29

olegbespalov (approved):

js/bundle.go
mstoykov: "Characters between" seems off to me :thinking: I also prefer if it is shorter, and I don't think the usage of `or` here complicates the understanding.

mstoykov, line 50:
```suggestion
		badNameWarning  = "Metric name should only include up to 128 ASCII letters, numbers and/or underscores. " +
```

olegbespalov (approved):

js/bundle.go
codebien: Sorry, but I'm not still getting why we are not just bumping everything to 255. What is the reason to keep as 128? If there is one technical then I would like us adding a comment on the registry's regex explaining it. Prometheus doesn't have any limit and the OpenTelemetry limit is set to 255 so it sounds to me we should just match the OTel limit.

mstoykov:
Docs updates in https://github.com/grafana/k6-docs/pull/1331

olegbespalov (approved):

codebien (approved):

PR #3340, opened 2023-09-19
Add `no-sandbox` arg to browser enabled Docker image

ka3de:
## What?

Sets the `no-sandbox` chrome argument by default through the `K6_BROWSER_ARGS` ENV variable in the Dockerfile definition for the browser enabled docker image.

## Why?

In order to allow chrome to execute in the browser enabled image, either `SYS_ADMIN` Docker capability or `no-sandbox` browser argument had to be set. This sets the `no-sandbox` option in the Dockerfile definition itself so this action is no longer necessary to run the Docker container. This is arguably also the better option, as we are running the chrome browser inside a container, versus using the Docker capability.

## Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -39,5 +39,8 @@ ENV CHROME_BIN=/usr/bin/chromium-browser
 ENV CHROME_PATH=/usr/lib/chromium/
 
 ENV K6_BROWSER_HEADLESS=true
+# no-sandbox chrome arg is required to run chrome browser in
+# alpine and avoids the usage of SYS_ADMIN Docker capability
+ENV K6_BROWSER_ARGS=no-sandbox
 
 ENTRYPOINT ["k6"]


ankur22 (approved):
Thanks for this! Makes things simpler for users. LGTM

inancgumus (approved):
Nice 👍 Documentation updates await 😅 I can help.

olegbespalov (approved):
Not familiar with this arg, but the change looks good :+1:

codebien:
It seems to be not really safe and we should research for alternatives. https://www.google.com/googlebooks/chrome/med_26.html

We should open a dedicated issue for researching alternatives. Here is an example: https://github.com/Zenika/alpine-chrome#-the-best-with-seccomp

I'm fine with it if we make it temporary, we document it and we are clear on the risks.

ka3de:
> It seems to be not really safe and we should research for alternatives. https://www.google.com/googlebooks/chrome/med_26.html
> 
> We should open a dedicated issue for researching alternatives. Here is an example: https://github.com/Zenika/alpine-chrome#-the-best-with-seccomp

We are aware of this, but considering that in our case the browser runs in a container, the `no-sandbox` usage was more reasonable, even if it's not as running it in a virtual machine. We use a SECCOMP in the cloud, but that is not much user friendly either.

But I agree, if we stick to the `no-sandbox` option, we should document it as a note/warning in the docs so users are aware that it's set by default.

codebien (approved):

Issue #3241, opened 2023-08-01
add metadata to reflection in grpc


gubtos:
### Feature Description

Reflection call on grpc Connect don`t support metadata.

https://github.com/grafana/k6/blob/master/js/modules/k6/grpc/client.go#L263

Current params parsing

https://github.com/grafana/k6/blob/master/js/modules/k6/grpc/client.go#L495-L583

### Suggested Solution (optional)

Add to reflection context the metadata from params argument

olegbespalov:
Would it be helpful to introduce the GRPC connection's default metadata, which could be automatically passed by default to the places that support metadata (unary's invoke, stream's, reflection calls)? In case of the presence of in-line metadata, they should have priority.

@grafana/k6-core, what do you think?

PR #3343, opened 2023-09-20
Refactoring of gRPC internals & tests and introducing the `reflectMetadata`

olegbespalov:
## What?

This PR contains several parts:
* First is refactoring of calling and connection params to make code more re-usable
* Second is the significant (but is mostly moving  & cleaning) refactoring of the gRPC client's test
* last but not least is an introduction to the `reflectMetadata` 

## Why?

The refactoring was done to make the following changes tinier and pay some technical dept of removing copy-pasting.

The `reflectMetadata` was requested in #3241 

## Checklist

## Related PR(s)/Issue(s)

https://github.com/grafana/xk6-grpc/pull/51

Closes: #3241

### DIFF
--- a/js/modules/k6/grpc/client_test.go
+++ b/js/modules/k6/grpc/client_test.go
@@ -1,16 +1,11 @@
-package grpc
+package grpc_test
 
 import (
 	"bytes"
 	"context"
 	"crypto/tls"
 	"crypto/x509"
-	"errors"
 	"fmt"
-	"io"
-	"net/url"
-	"os"
-	"runtime"
 	"strings"
 	"testing"
 
@@ -19,7 +14,6 @@ import (
 	"google.golang.org/protobuf/reflect/protoregistry"
 	"google.golang.org/protobuf/types/known/wrapperspb"
 
-	"github.com/dop251/goja"
 	"github.com/golang/protobuf/ptypes/any"
 	"github.com/golang/protobuf/ptypes/wrappers"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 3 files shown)


mstoykov (approved):

olegbespalov:
I rebased, since there was a minor conflict with `master` branch.

js/modules/k6/grpc/client_test.go
```diff
@@ -1,16 +1,11 @@
-package grpc
+package grpc_test
```
codebien, line 1: Out of curiosity, why are we moving them to a specific package? Do we have cycle dependencies or other issues? I'm not against it, but I've seen some ongoing discussions around gRPC and extensions and I have the feeling I lost something.

js/modules/k6/grpc/client_test.go
```diff
@@ -1325,74 +1131,44 @@ func TestDebugStat(t *testing.T) {
 func TestClientInvokeHeadersDeprecated(t *testing.T) {
```
codebien, line 1132: It sounds like it could be the time to drop it completly.

codebien (approved):

js/modules/k6/grpc/client_test.go
olegbespalov: I'd say it's just a better way of structuring the tests (best practice), basically test package becomes a user of the main package, which blocks access to the main package's internals, and as a result, we should write more robust tests, since even something internally in the package changed the tests continue to works since they rely on the public API

js/modules/k6/grpc/client_test.go
olegbespalov: The test is still relevant for the existing functionality. However, I like the idea of dropping the `headers` property, but how about opening a separate PR  where we start erroring (right now, we just `fallthrough` to the `metadata`) and keep this for a cycle/too till we altogether remove it?

mstoykov (approved):

js/modules/k6/grpc/client_test.go
codebien: I'm fine with it :+1:

PR #3346, opened 2023-09-20
Fix race in http.batch after latest goja update

mstoykov:
Previous http.batch code did call `runtime.NewArrayBuffer` concurrently.

Which never has been guaranteed to be safe, but with the latest goja changes to having prototypes created on demand it races.

The fix *might* have some performance implications, but they are likely to be very small as `NewArrayBuffer` mostly wraps stuff so ... hopefully not a big deal.

### DIFF
--- a/js/modules/k6/http/request.go
+++ b/js/modules/k6/http/request.go
@@ -503,6 +503,12 @@ func (c *Client) Batch(reqsV ...goja.Value) (interface{}, error) {
 			err = e
 		}
 	}
+	for i, req := range batchReqs {
+		if req.Response != nil {
+			c.processResponse(req.Response, req.ParsedHTTPRequest.ResponseType)
+		}
+		batchReqs[i] = req
+	}
 	return results, err
 }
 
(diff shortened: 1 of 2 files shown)


js/modules/k6/http/request.go
```diff
@@ -503,6 +503,12 @@ func (c *Client) Batch(reqsV ...goja.Value) (interface{}, error) {
 			err = e
 		}
 	}
+	for i, req := range batchReqs {
+		if req.Response != nil {
+			c.processResponse(req.Response, req.ParsedHTTPRequest.ResponseType)
+		}
+		batchReqs[i] = req
```
olegbespalov, line 510: I believe there is no need to re-assign the variable since in any case, `processResponse` works with the pointers
```suggestion
	for _, req := range batchReqs {
		if req.Response != nil {
			c.processResponse(req.Response, req.ParsedHTTPRequest.ResponseType)
		}
```

olegbespalov (commented):
Looks good, left a minor comment :+1:

olegbespalov (approved):

codebien (approved):

Issue #2982, opened 2023-03-21
Future of StatsD outputs


imiric:
The k6-core team recently had an internal meeting to discuss the current status of StatsD outputs in k6, the several issues reported by users, and the best way to move forward that benefits both users, and maintenance of k6.

## Background

The [`statsd` output](https://k6.io/docs/results-output/real-time/statsd/) is a generic implementation of the [StatsD line protocol](https://github.com/statsd/statsd#usage). It allows pushing metrics to any StatsD server, and, in theory, any StatsD backend.

Under the hood, it uses the [DogStatsD client library](https://github.com/DataDog/datadog-go), but it doesn't use any Datadog-specific functionality. This is why the same output can be used to send metrics to both [Datadog](https://k6.io/docs/results-output/real-time/datadog/) and [New Relic](https://k6.io/docs/results-output/real-time/new-relic/).

## Issues

The problem is that, on its own, this generic implementation is very basic, and only supports [3 metric types](https://github.com/statsd/statsd/blob/master/docs/metric_types.md) (Counters, Gauges, and Sets). Datadog and New Relic have extended the protocol to add support for more metric types, though these changes, are not standard, and likely wouldn't work across different servers and backends. So we can't use this functionality in the `statsd` output without risking breaking generic StatsD support, or specifically supporting individual backends, which we want to avoid.

There are several issues reported by users on the forum:

- https://community.k6.io/t/trend-metrics-on-datadog-percentile/3995
- https://community.k6.io/t/http-req-duration-wrong-datadog/440
- https://community.k6.io/t/why-do-k6-http-req-duration-max-and-k6-http-req-duration-95percentile-have-the-same-value-in-datadog/5487
- https://community.k6.io/t/percentiles-calculation-on-new-relic/3690
- https://community.k6.io/t/how-to-get-the-correct-values-for-p-90-on-k6-newrelic/5192

Some of these, particularly those related to percentile values, and summarized in #2819, could be fixed by using the non-standard metric types, such as [Distribution](https://docs.datadoghq.com/metrics/distributions/). But, as mentioned above, we can't use it directly in the output.

## Proposed solutions

We agreed that the way forward consists of several steps:

1. Research the status of alternative protocol support for Datadog and New Relic.

[...]

PR #3347, opened 2023-09-21
Deprecate statsd output

mstoykov:
Updates #2982

### DIFF
--- a/cmd/outputs.go
+++ b/cmd/outputs.go
@@ -30,7 +30,13 @@ func getAllOutputConstructors() (map[string]output.Constructor, error) {
 			return nil, errors.New("the kafka output was deprecated in k6 v0.32.0 and removed in k6 v0.34.0, " +
 				"please use the new xk6 kafka output extension instead - https://github.com/k6io/xk6-output-kafka")
 		},
-		"statsd": statsd.New,
+		"statsd": func(params output.Params) (output.Output, error) {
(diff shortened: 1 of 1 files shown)


codebien (approved):

cmd/outputs.go
```diff
@@ -30,7 +30,13 @@ func getAllOutputConstructors() (map[string]output.Constructor, error) {
 			return nil, errors.New("the kafka output was deprecated in k6 v0.32.0 and removed in k6 v0.34.0, " +
 				"please use the new xk6 kafka output extension instead - https://github.com/k6io/xk6-output-kafka")
 		},
-		"statsd": statsd.New,
+		"statsd": func(params output.Params) (output.Output, error) {
+			params.Logger.Warn("The statsd output is deprecated, and will be removed in a future k6 version. " +
```
olegbespalov, line 34: We probably should update the `datadog` case too since it recommends using this `statsd` :sweat_smile:

mstoykov: I thought about this but prefer to leave it for when we remove this - to update both to the extension. Or to be honest to just drop kafka and datadog as they have not been a thing for over 2 years at this point :sweat_smile:

olegbespalov: :+1: for dropping (but OK with doing this in a separate PR)

olegbespalov (approved):

javaducky (approved):
LGTM! Thank you!

Issue #2508, opened 2022-04-28
Change config directory name from "loadimpact" to "k6"


MattDodsonEnglish:
### Feature Description

Looking in my `.config` , I happened to notice the `loadimpact` directory. Inside that directory, there was a `k6` directory, which had a `config.json` file with my API key.

Can the `loadimpact` part be removed? Users looking for their configs will be much more likely to look for k6.  Besides, not everyone will associate k6 with loadimpact. As time goes on, this association will become weaker and weaker, too, as the official business name is now k6.

### Suggested Solution (optional)

`$XDG_CONFIG_HOME/k6/config.json`

For example,
`~/.config/loadimpact/k6/config.json ` 
↓
`~/.config/k6/config.json`

Appropriate default directory locations might be different on Mac and Windows.

MStoykov:
This will likely also mean that we need code to:
1. copy the old config in the place of the new one if it doesn't exist. But print warning if it does
2. delete the old, so we don't print warnings. 
 
Hopefully with the latest changes to the `cmd` package this will need less hacks to implement :crossed_fingers:

na--:
Yeah, not sure how we should handle old configs, but this should definitely be easier to implement **and test** nowadays :crossed_fingers: 

We should probably look into how other grafana projects handle this, if they have similar config files in `.config/grafana`, so we can move ours to `.config/grafana/k6`? :thinking: 

Personally, I am slightly against automatically moving `.config/loadimpact` during a `k6 run` command. Instead, I think we should:
- support both `.config/loadimpact` and `.config/grafana`, with `.config/grafana` being checked for and used first, and `.config/loadimpact` emitting a deprecation warning if it ends up being loaded
- move the config files when `k6 login` is executed (and explain that in the deprecation warning :arrow_up: )

MattDodsonEnglish:
I realized that the docs will need to be updated in a few places if this change happens:
- [ ] https://k6.io/docs/cloud/integrations/token/#authentication-with-a-config-file
- [ ] https://k6.io/docs/using-k6/options/#config ( or `/docs/using-k6/k6-options/reference#config` )

dgzlopes:
I would go with `.config/k6`

As far as I know, in the Grafana world, this pattern doesn't exist. Also, it looks easier to find for k6 users.

markjmeier:
+1 to I would go with `.config/k6` for the same reasons Daniel mentions.

[...]

PR #3348, opened 2023-09-21
Introducing options.cloud

### LINKED ISSUE #1155: Refactor how cloud options are specified in the script file
This is a proposal for renaming `options.ext.loadimpact` to `options.cloud`. 

```JavaScript
// current structure
export let options = {
    ext: {
        loadimpact: {
            name: "My test name",
            projectID: 12345,
            staticIPs: true,
        }
    }
};

// proposed structure
export let options = {
    cloud: {
        name: "My test name",
        projectID: 12345,
        staticIPs: true,
    }
};
```

The command used to run cloud tests is `k6 cloud`. It's logical to expect that options for cloud would be specified in `options.cloud` rather than `options.ext.loadimpact`.

Connected issues: https://github.com/loadimpact/k6/issues/587, https://github.com/loadimpact/k6/issues/883

olegbespalov:
## What?

This is an attempt to introduce the new option `cloud`, which should start the process of sunsetting the `options.ext.loadimpact`

## Why?

The `options.ext.loadimpact` is the legacy option, and we should migrate from it to something more neutral like `cloud`

## Checklist

## Related PR(s)/Issue(s)

Closes #1155

### DIFF
--- a/cloudapi/config.go
+++ b/cloudapi/config.go
@@ -1,6 +1,7 @@
 package cloudapi
 
 import (
+	"bytes"
 	"encoding/json"
 	"time"
 
@@ -8,6 +9,9 @@ import (
 	"go.k6.io/k6/lib/types"
 	"gopkg.in/guregu/null.v3"
 )
+
+// LegacyCloudConfigKey is the key used in the JSON config for the cloud output.
+const LegacyCloudConfigKey = "loadimpact"
 
 // Config holds all the necessary data and options for sending metrics to the k6 Cloud.
 //
@@ -165,44 +169,40 @@ func (c Config) Apply(cfg Config) Config {
 	return c
 }
 
-// MergeFromExternal merges three fields from the JSON in a loadimpact key of
-// the provided external map. Used for options.ext.loadimpact settings.
-func MergeFromExternal(external map[string]json.RawMessage, conf *Config) error {
-	if val, ok := external["loadimpact"]; ok {
-		// TODO: Important! Separate configs and fix the whole 2 configs mess!
-		tmpConfig := Config{}
-		if err := json.Unmarshal(val, &tmpConfig); err != nil {
-			return err
-		}
-		// Only take out the ProjectID, Name and Token from the options.ext.loadimpact map:
-		if tmpConfig.ProjectID.Valid {
-			conf.ProjectID = tmpConfig.ProjectID
-		}
-		if tmpConfig.Name.Valid {
-			conf.Name = tmpConfig.Name
-		}
-		if tmpConfig.Token.Valid {
-			conf.Token = tmpConfig.Token
-		}
-	}
-	return nil
-}
-
 // GetConsolidatedConfig combines the default config values with the JSON config
 // values and environment variables and returns the final result.
+// it also returns a warning message that could be shown to the user.
+// to bring some attention to the fact that the user.
 func GetConsolidatedConfig(
-	jsonRawConf json.RawMessage, env map[string]string, configArg string, external map[string]json.RawMessage,
-) (Config, error) {
+	jsonRawConf json.RawMessage,
+	env map[string]string,
+	configArg string,
+	cloudConfig json.RawMessage,
+	external map[string]json.RawMessage,
+) (Config, string, error) {
+	warn := ""
+
 	result := NewConfig()
 	if jsonRawConf != nil {
 		jsonConf := Config{}
 		if err := json.Unmarshal(jsonRawConf, &jsonConf); err != nil {
-			return result, err
+			return result, warn, err
 		}
 		result = result.Apply(jsonConf)
 	}
-	if err := MergeFromExternal(external, &result); err != nil {
-		return result, err
+
+	if err := mergeFromCloudOptionAndExternal(cloudConfig, external, &result); err != nil {
+		return result, warn, err
+	}
+
+	// We want to show a warning if the user is using the only old way of defining the config.
+	// Note: Since the migration to the options.cloud is a long process, this warning is planned
+	// to be emitted for a long time (1-2 years), after some point, and depending on the state
+	// of migration we could re-evaluate this warning.
+	if cloudConfig == nil && external != nil {
+		if _, ok := external[LegacyCloudConfigKey]; ok {
+			warn = "The options.ext.loadimpact option is deprecated, please use options.cloud instead"
+		}
 	}
 
 	envConfig := Config{}
@@ -211,7 +211,7 @@ func GetConsolidatedConfig(
 		return v, ok
 	}); err != nil {
 		// TODO: get rid of envconfig and actually use the env parameter...
-		return result, err
+		return result, warn, err
 	}
 	result = result.Apply(envConfig)
 
@@ -219,5 +219,85 @@ func GetConsolidatedConfig(
 		result.Name = null.StringFrom(configArg)
 	}
 
-	return result, nil
+	return result, warn, nil
+}
+
+// mergeFromCloudOptionAndExternal merges three fields from the JSON in a cloud key of
+// the provided external map. Used for options.cloud settings.
+func mergeFromCloudOptionAndExternal(
+	cloudConfig json.RawMessage,
+	external map[string]json.RawMessage,
+	conf *Config,
+) error {
+	source := pickSource(cloudConfig, external)
+	if source == nil {
+		return nil
+	}
+
+	// Original comment
+	// TODO: Important! Separate configs and fix the whole 2 configs mess!
+	tmpConfig := Config{}
+	if err := json.Unmarshal(source, &tmpConfig); err != nil {
+		return err
+	}
+	// Only take out the ProjectID, Name and Token from the options.cloud (or legacy loadimpact struct) map:
+	if tmpConfig.ProjectID.Valid {
+		conf.ProjectID = tmpConfig.ProjectID
+	}
+	if tmpConfig.Name.Valid {
+		conf.Name = tmpConfig.Name
+	}
+	if tmpConfig.Token.Valid {
+		conf.Token = tmpConfig.Token
+	}
+
+	return nil
+}
+
+// GetTemporaryCloudConfig returns a temporary cloud config.
+// Original comment
+// TODO: Fix this
+// We reuse cloud.Config for parsing options.cloud (or legacy loadimpact struct), but this probably shouldn't be
+// done, as the idea of options.ext is that they are extensible without touching k6. But in
+// order for this to happen, we shouldn't actually marshal cloud.Config on top of it, because
+// it will be missing some fields that aren't actually mentioned in the struct.
+// So in order for use to copy the fields that we need for k6 cloud's api we unmarshal in
+// map[string]interface{} and copy what we need if it isn't set already
+func GetTemporaryCloudConfig(
+	cloudConfig json.RawMessage,
+	external map[string]json.RawMessage,
+) (map[string]interface{}, error) {
+	tmpCloudConfig := make(map[string]interface{}, 3)
+
+	source := pickSource(cloudConfig, external)
+	if source == nil {
+		return tmpCloudConfig, nil
+	}
+
+	dec := json.NewDecoder(bytes.NewReader(source))
+	dec.UseNumber() // otherwise float64 are used
+	if err := dec.Decode(&tmpCloudConfig); err != nil {
+		return nil, err
+	}
+
+	return tmpCloudConfig, nil
+}
+
+// pickSource returns the config source to use.
+func pickSource(
+	cloudConfig json.RawMessage,
+	external map[string]json.RawMessage,
+) json.RawMessage {
+	// priority is the new way of defining the config
+	// via options.cloud
+	if cloudConfig != nil {
+		return cloudConfig
+	}
+
+	// fallback to the old way of defining the config
+	if val, ok := external[LegacyCloudConfigKey]; ok {
+		return val
+	}
+
+	return nil
 }
--- a/cloudapi/config_test.go
+++ b/cloudapi/config_test.go
@@ -55,17 +55,78 @@ func TestConfigApply(t *testing.T) {
 
 func TestGetConsolidatedConfig(t *testing.T) {
 	t.Parallel()
-	config, err := GetConsolidatedConfig(json.RawMessage(`{"token":"jsonraw"}`), nil, "", nil)
+	config, warn, err := GetConsolidatedConfig(json.RawMessage(`{"token":"jsonraw"}`), nil, "", nil, nil)
 	require.NoError(t, err)
 	require.Equal(t, config.Token.String, "jsonraw")
+	require.Empty(t, warn)
 
-	config, err = GetConsolidatedConfig(json.RawMessage(`{"token":"jsonraw"}`), nil, "",
+	config, warn, err = GetConsolidatedConfig(
+		json.RawMessage(`{"token":"jsonraw"}`),
+		nil,
+		"",
+		json.RawMessage(`{"token":"ext"}`),
+		nil,
+	)
+	require.NoError(t, err)
+	require.Equal(t, config.Token.String, "ext")
+	require.Empty(t, warn)
+
+	config, warn, err = GetConsolidatedConfig(
+		json.RawMessage(`{"token":"jsonraw"}`),
+		map[string]string{"K6_CLOUD_TOKEN": "envvalue"},
+		"",
+		json.RawMessage(`{"token":"ext"}`),
+		nil,
+	)
+	require.NoError(t, err)
+	require.Equal(t, config.Token.String, "envvalue")
+	require.Empty(t, warn)
+}
+
+func TestGetConsolidatedConfig_WithLegacyOnly(t *testing.T) {
+	t.Parallel()
+
+	config, warn, err := GetConsolidatedConfig(json.RawMessage(`{"token":"jsonraw"}`), nil, "", nil,
 		map[string]json.RawMessage{"loadimpact": json.RawMessage(`{"token":"ext"}`)})
 	require.NoError(t, err)
 	require.Equal(t, config.Token.String, "ext")
+	require.NotEmpty(t, warn)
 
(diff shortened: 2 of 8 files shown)


cloudapi/config.go
```diff
@@ -346,3 +341,83 @@ func GetConsolidatedConfig(
...
+	return tmpCloudConfig, nil
+}
+
+// pickSource returns the config source to use.
+func pickSource(
+	cloudConfig json.RawMessage,
+	external map[string]json.RawMessage,
+) json.RawMessage {
+	// priority is the new way of defining the config
+	// via opt.config
```
codebien, line 412: I have the feeling this comment should be rearranged. It seems to me `priority` is an option and could be set somewhere, but isn't, right?

cmd/archive_test.go
```diff
@@ -128,6 +128,110 @@ func TestArchiveContainsEnv(t *testing.T) {
...
+		export default function () {}
+	`)
+	ts := tests.NewGlobalTestState(t)
+	require.NoError(t, fsext.WriteFile(ts.FS, filepath.Join(ts.Cwd, fileName), testScript, 0o644))
+
+	// when we do archiving
+	ts.CmdArgs = []string{"k6", "archive", fileName}
+
+	newRootCommand(ts.GlobalState).execute()
+	require.NoError(t, untar(t, ts.FS, "archive.tar", "tmp/"))
```
codebien, line 156: Can we also add a test with a static archive, please? In this way, we don't directly depend on the _current_ logic of the archive command. So, in the event, we will change the logic (for whatever reason) then we are still sure that an old archive is compatible.

cloudapi/config_test.go
codebien, line 1: I think we are missing here a specific test for the parsing function asserting that the entire set of available options is correctly parsed. So if we drop or add an option for the Cloud I would prefer if at least a test fails.

Because all the tests here are focusing on the _consolidation_ logic.

cmd/cloud.go
```diff
@@ -148,11 +138,15 @@ func (c *cmdCloud) run(cmd *cobra.Command, args []string) error {
...
 	}
-	arc.Options.External["loadimpact"], err = json.Marshal(tmpCloudConfig)
+
+	b, err := json.Marshal(tmpCloudConfig)
 	if err != nil {
 		return err
 	}
 
+	arc.Options.Cloud = b
+	arc.Options.External[cloudapi.LegacyCloudConfigKey] = b
```
codebien, line 148:
```suggestion
	arc.Options.SetCloud(b)
```
Can we add a method directly a method on the Options type? In this way, we guarantee the _atomicity_ of the operation in the future preventing bugs, in case we change the code here.

cmd/login_cloud.go
codebien, line 1: I think this is not yet available but if I'm wrong then discard it.

We should add a set of test cases for asserting the written config. (e.g. Having the matrix of starting from the legacy and redoing the login or starting from the new one.)

codebien (changes requested):
@olegbespalov did a first cycle, thanks for the great work :bow:

cloudapi/config.go
```diff
@@ -346,3 +341,83 @@ func GetConsolidatedConfig(
...
+	external map[string]json.RawMessage,
+	conf *Config,
+) error {
+	source := pickSource(cloudConfig, external)
+	if source == nil {
+		return nil
+	}
+
+	// Original comment
+	// TODO: Important! Separate configs and fix the whole 2 configs mess!
```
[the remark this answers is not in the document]
olegbespalov: The issue that I'm not getting this comment to, it's an original comment that was presented in the place around.

TBH for the case of refactoring that I did it wasn't so relevant, but on the other side I'm not comfortable removing it.

cloudapi/config.go
```diff
@@ -346,3 +341,83 @@ func GetConsolidatedConfig(
...
+	tmpConfig := Config{}
+	if err := json.Unmarshal(source, &tmpConfig); err != nil {
+		return err
+	}
+	// Only take out the ProjectID, Name and Token from the options.cloud (or legacy loadimpact struct) map:
```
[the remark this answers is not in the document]
olegbespalov: This is again the comment from the past https://github.com/grafana/k6/pull/3348/files#diff-55ae88ea92051c07942fe7b2477d0b5ee5c7889eb78825e8dd8db7fbf0f96e69L302

cmd/login_cloud.go
olegbespalov: Not sure if I'm getting what is your suggestions :cry:

cloudapi/config.go
```diff
@@ -346,3 +341,83 @@ func GetConsolidatedConfig(
...
+	if tmpConfig.Token.Valid {
+		conf.Token = tmpConfig.Token
+	}
+
+	return nil
+}
+
+// GetTemporaryCloudConfig returns a temporary cloud config.
+// Original comment
+// TODO: Fix this
```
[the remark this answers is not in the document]
olegbespalov: I don't see a way of improving it, and to be honest, I prefer to keep the scope smaller to have a chance of delivering it.

cloudapi/config.go
```diff
@@ -158,6 +162,9 @@ type Config struct {
 
 	// Deprecated: Remove this when migration from the cloud output v1 will be completed
 	MaxMetricSamplesPerPackage null.Int `json:"maxMetricSamplesPerPackage" envconfig:"K6_CLOUD_MAX_METRIC_SAMPLES_PER_PACKAGE"`
+
+	// WarningMessage is a string that will be shown to the user if the config has some invalid values.
+	WarningMessage string `json:"-"`
```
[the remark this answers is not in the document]
olegbespalov: In my opinion, we could start small, and if we see the need to extend it, but that particular place I don't mind if you think this serve some future needs I could convert to the array

cloudapi/config.go
```diff
@@ -346,3 +341,83 @@ func GetConsolidatedConfig(
...
+	return tmpCloudConfig, nil
+}
+
+// pickSource returns the config source to use.
+func pickSource(
+	cloudConfig json.RawMessage,
+	external map[string]json.RawMessage,
+) json.RawMessage {
+	// priority is the new way of defining the config
+	// via opt.config
```
olegbespalov, line 412:
```suggestion
	// via options.cloud
```

cloudapi/config.go
olegbespalov: Good catch :facepalm:

cmd/cloud.go
olegbespalov: On the one side, I'm with you. On the other side, I don't think that this should be publically exposed with the API :thinking:

cloudapi/config_test.go
olegbespalov: That's correct since the subject of the tests of this is `GetConsolidatedConfig`

I think we discussed this before, but my understanding is that the cloud options are more under the control of the backend and that's why they more "flexible", that's why for instance they have a type `json.RawMessage`

cloudapi/config.go
```diff
@@ -326,9 +314,16 @@ func GetConsolidatedConfig(
...
 	}
-	if err := MergeFromExternal(external, &result); err != nil {
+
+	if err := mergeFromCloudOptionAndExternal(cloudConfig, external, &result); err != nil {
 		return result, err
 	}
+	// we want to show a warning if the user is using the only old way of defining the config
+	if cloudConfig == nil && external != nil {
+		if _, ok := external[LegacyCloudConfigKey]; ok {
+			result.WarningMessage = "The options.ext.loadimpact option is deprecated, please use the a options.cloud instead"
```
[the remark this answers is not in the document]
olegbespalov: :thinking: we already have a comment 
```
// we want to show a warning if the user is using the only old way of defining the config
```
Do you think more is needed? And I'm not sure what you suggest to emphasize with 

```
// comment here about the deprecation expectations
```

cmd/cloud.go
olegbespalov: Also important to say that this is more for safety, and the future version should have only `arc.Options.Cloud = b`

cloudapi/config.go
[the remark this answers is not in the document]
codebien: I meant to comment about our plans for the depreciation period. Should we remove them at some point or should we keep forever?

>// We plan to keep the `ext.loadimpact` support even if it is deprecated for at least 1-2 years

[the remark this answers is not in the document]
olegbespalov: Ah, okay :+1: makes sense, I'll update a comment that stays before the whole block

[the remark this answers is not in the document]
olegbespalov: Updated the comment with the latest force-push

cmd/archive_test.go
codebien: Consider a script that includes the following options:

```js
  ext: {
    loadimpact: {
      name: 'Hello k6 cloud!',
      distribution: {
        distributionLabel1: { loadZone: 'amazon:us:ashburn', percent: 50 },
        distributionLabel2: { loadZone: 'amazon:ie:dublin', percent: 50 },
      },
      note: 'Anything that may be worth noting about your test.',
   },
```

If we archive and we reuse it for running a script in the Cloud then the options are used for defining the distribution.

```
k6 archive script.js
k6 cloud archive.tar
```
So, if we break the archive regarding the Cloud part I'm not sure it will fail somewhere if we don't add one here.

cmd/cloud.go
codebien: >On the other side, I don't think that this should be publically exposed with the API 🤔

I'm not getting it. `arc.Options.Cloud = b` is going to be exported, so what will be the difference?

>Also important to say that this is more for safety, and the future version should have only arc.Options.Cloud = b

I got the feeling we would like to keep the 2-way forever. Have we changed the plan around it?

lib/options.go
```diff
@@ -299,6 +299,10 @@ type Options struct {
 	// iteration is shorter than the specified value.
 	MinIterationDuration types.NullDuration `json:"minIterationDuration" envconfig:"K6_MIN_ITERATION_DURATION"`
 
+	// Cloud is the config for the cloud
+	// formally known as ext.loadimpact
+	Cloud json.RawMessage `json:"cloud,omitempty"`
```
codebien, line 304:
```suggestion
	Cloud json.RawMessage `json:"cloud,omitempty"`
```
Why not use the opportunity of this change for defining a strict struct instead of using `json.RawMessage`? Sorry, for asking only now.

cmd/login_cloud.go
codebien: I meant `k6 login cloud` but I guess there isn't any way for that command to be impacted by the change in this PR.

olegbespalov: Yes, I think it's more related to the https://github.com/grafana/k6/issues/2508

cmd/cloud.go
olegbespalov: > I'm not getting it. arc.Options.Cloud = b is going to be exported, so what will be the difference?

The difference is that the `Cloud` and `Ext` are the fields that will stay almost forever, but this two-times setting is a temporary thing. So simply the method like `Options.SetCloudOptions` is also a temporary thing.

>I got the feeling we would like to keep the 2-way forever. Have we changed the plan around it?

Not, that was probably a wrong impression. We keep the support of the `options.ext.loadimpact` for a long time, but this place is writing. There is no need to continue filling out the `options.ext.loadimpact` if all backend components support the new way of `options.cloud`. So once we rich that state, there is no more need to write the ``options.ext.loadimpact` and this place will become just `arc.Options.Cloud = b`

cmd/archive_test.go
olegbespalov: Okay, sounds like a  not common, but still valid case, so I will add the test :+1:

lib/options.go
olegbespalov: Agreed internally that this could increase the scope of the issue, but doesn't bring the value for original PRs purpose (introducing the `options.cloud`). So if we need to switch to the struct we can do that later.

cmd/archive_test.go
olegbespalov: Extracted as separated PR https://github.com/grafana/k6/pull/3402

cloudapi/config.go
codebien, line 167: Another comment about it, I have the feeling it could be better to have it as a returned value of the _consolidation_ method. I know it will be a breaking change and it will require more work but it adds some better feedback for the code depending on it, making this change safer. Most of the code that uses this use this method should be internal (right?) so we aren't creating breaking changes for the users.
I think you originally thought about it, so probably you have a wider analysis of it.

olegbespalov: It's not a breaking change. I've considered this and discarded it since I found no value in having a warning message as a peace.

My point about having this as the property is that I don't consider this warning as something you need immediately after the config consolidation. So, it's just a config property. It's so different from the error since we should stop the process if the `GetConsolidatedConfig` errors. But a warning isn't like that. Next, that change requires streaming down this warning along with the config anyway, and this likely wants to improve the code from the reader's perspective.

From your standpoint, you are having this separated. What do you think will be an improvement? And what problem will this separation solve?

codebien: >And what problem will this separation solve?

I see the following code block implemented in multiple places:
```go
<configVar>.WarningMessage != "" {
```
so if the next time we implement a new part of code where consolidation would be required, or refactoring one of the already existing, it is a piece of logic that could be forgotten and not notified, where instead we could have the compiler doing the work for us and benefit from it.

codebien: And the reason, behind so much pickiness, is the attempt to reduce adding more things to be aware of, in a part of the code where we already have a bunch of unknown/unwritten logic. As the discussions before for the unknown comments show.

olegbespalov: >I see the following code block implemented in multiple places:

Maybe I'm missing something, but the only change will be that instead of the 

```
<configVar>.WarningMessage != "" {
```
We will end up with the 
```
warningMessage != "" {
```

And the predominant issue here is that you likely won't use the `warn` in the same method where you execute the `GetConsolidatedConfig`. 

So still, to be honest, I don't see it as a valuable improvement that deserves a decrease in the readability of the code. 

However, I might be wrong. So let's wait for the third opinion here. It's not a big deal to implement this and as I said it's not a breaking change, but rather an internal thing.

codebien: >Maybe I'm missing something, but the only change will be that instead of the

The main change to me is that when you call `GetConsolidatedConfig` instead of getting `c, err :=` you would get `c, warning/s, err :=`.

>And the predominant issue here is that you likely won't use the warn in the same method where you execute the GetConsolidatedConfig.

I agree that it would be the predominant issue in that case, but It seems to me that the current code does the check exactly after invoking `GetConsolidatedConfig`.

olegbespalov: Okay, you're right here :+1: Let me adjust

olegbespalov: Done with the latest force push

codebien (approved):
Thanks for the effort :bow:

olegbespalov:
@codebien I've extracted warnings (or referings) of the `options.cloud` to the #3407. The reason is that it could be so that we could be in a state where we can merge this PR, but not in the state where we should actively tell customers to use the `options.cloud` . Feel free to bring your opinion.

codebien (approved):
>I've extracted warnings (or referings) of the options.cloud to the https://github.com/grafana/k6/pull/3407

Yep, I like the idea. LGTM

mstoykov (dismissed):
LGTM! :clap: 

but I have not tried to break it extensively locally. 

My main feedback is that I still am of the opinion that if `options.cloud` and `options.ext.loadimpact` exist - this should lead to an early error instead of one of them taking precedence. 

There is IMO no reason someone will have both and at best it will be confusing to users which one they need to change.

olegbespalov:
@mstoykov 

>My main feedback is that I still am of the opinion that if options.cloud and options.ext.loadimpact exist - this should lead to an early error instead of one of them taking precedence.

I feel you, but my main concern is that we can't start warn users till the backend supports the `options.cloud`, which requires more time and work on the backend side, so this PR is aiming to open possibilities for silent support of the new way. 

@codebien @mstoykov I've re-based (with start using `testutils.Untar` over the `untar`). Could you please give a final approve I'm going to merge this.

codebien:
>Could you please give a final approve I'm going to merge this.

@olegbespalov What is the plan? Are we going to merge but not announce it, so not mention it in the release notes?

olegbespalov:
We could announce that we started this migration, I mean it fits into our future plans section, but yes this PR I'm going to merge since it's quite independent

mstoykov:
@olegbespalov if someone tomorrow uses `options.cloud` and `options.ext.loadimpact` will it work in the k6 cloud?

olegbespalov:
@mstoykov no it won't work

mstoykov:
Then I am very much against merging something that when used will break UX

olegbespalov:
> Then I am very much against merging something that, when used will break UX

From my point of view if we don't offer to use `options.cloud` the only way of breaking it is intentionally trying to use it, which will not break UX :thinking: 

The changes in this PR is just a refactoring :shrug: 

That's the whole reason I've decided to extract parts of this PR (https://github.com/grafana/k6/pull/3348#issuecomment-1770662180).

codebien (changes requested):
Setting it as blocked, as we're waiting for changes on the impacted services.

olegbespalov:
Okay, @codebien, here is an explicit request for a review of this (and #3407) PRs :relaxed:

codebien (dismissed):

mstoykov (dismissed):

olegbespalov:
@codebien @mstoykov please re-approve. The PR was rebased because of the deletion of the cloud v1

mstoykov (approved):

codebien (approved):
Approving ~~with the assumption we have a green CI~~ we have :tada:

yorugac:
Hello :wave: Just for understanding of this change: can a user specify both the `loadimpact` and `cloud` at the same time? E.g. put the name in `loadimpact` and distribution and project ID in `cloud`?

olegbespalov:
hey @yorugac . Nope, mixing isn't acceptable. If the `options.cloud` exists, we'll use it as the source; otherwise, we fall back to the `options.ext.loadimpact`.

PR #3358, opened 2023-09-28
Add a design document proposal to support open in VU context

oleiade:
This draft aims to address [#3020](https://github.com/grafana/k6/issues/3020), focusing on enabling k6 scripts to access files directly from the VU stage, not just the init stage.

While the primary focus is on file access, I've currently left out considerations regarding `require` from [#3020](https://github.com/grafana/k6/issues/3020). If require should be part of this discussion, based on similar requirements or challenges, I'm more than open to updating the proposal to encompass it.

If you've got alternative solutions or insights, please contribute with a separate commit.

I'm eager to hear your thoughts and feedback in the comments below. Thanks for taking the time to review! 🙇🏻

### DIFF
--- a/docs/design/020-opening-files-in-vu-context.md
+++ b/docs/design/020-opening-files-in-vu-context.md
@@ -1,0 +1,101 @@
+# Opening files within the VU stage
+
+
+| Date | Change |
+| --- | --- |
+| Authors | @oleiade |
+| Status | Draft |
+| Issue | [#3020](https://github.com/grafana/k6/issues/3020)
+
+## Problem statement
+
+In its present state, k6's approach to handling external files, both in local/OSS and cloud contexts, restricts the ability to open files within the VU (Virtual User) stage.
+
+Question: **How might we enable k6 to access external resources directly from the VU stage?**
+
+
+### Rationale
+
+One of k6 main features is the ability to create reproducible, idempotent archives. These archives can be run consistently across various environments, permitting users to draft scripts locally and then either upload or execute them in the cloud. 
+
+To accomplish this, k6 assembles user scripts and their associated assets (which can range from other scripts, as well as text (JSON, CSV, etc.) and binary files (protobuf, images, etc.)) into a specific tar archive format. Furthermore, k6 can initiate a test by unpacking and executing such an archive.
+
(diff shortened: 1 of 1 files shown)


codebien (commented):
We should add a comparison in terms of pros&cons for the solutions proposed and we should add a section for eventual risks.

I usually prefer coding soltuions but this time I would prefer the option. I think it enables the concept of injecting the infrastructure where the file system could be part of.

Having the option could allow us to control this injection from the CLI, which sounds closer to what ops are already doing, mounting volumes.

I have the feeling that the Option's implementation will be simpler as we don't have to go through the code and scan for this special function.

docs/design/020-opening-files-in-vu-context.md
```diff
@@ -0,0 +1,101 @@
...
+}
+```
+
+**Key Points**:
+- The rules from the explicit API still apply: if a file isn't referenced in the options and is opened within the VU stage, an error occurs.
+- Attempting to open an un-referenced file within the init stage may not result in an error, ensuring backward compatibility.
+
+### Solution B: Awaiting Proposals
+
+Have a novel idea? Please share it with us in a dedicated commit :bow:
```
mstoykov, line 97: I would argue we do not need a *new* API at all.

There are two problems:
1. you can't *at* all `open` files outside of init code
2. it might be beneficial to have an API that tells k6 to allow a given file to be openable outside of the init code.

There is also the sligth twist that *actually* `open` will return you an error even in init code if you try to open a file in a VU after the initial one that was *not* opened by the initial one.

https://github.com/grafana/k6/pull/2314

```javascript
open("./someotherfile.txt") // will not error out for either the initial nor the VU with id `1`
if (__VU == 1) {
  open("./somefile.txt") // will always error out as the initial VU did not open it.
}
```

Again the first `open` is called both in VU with id `0` but also for any VU that will execute a default function (for example).  See https://github.com/grafana/k6/issues/1771 for the rational and discussion.

the tl;dr is that if we want to allow `open` to at all execute in vu code we can have the same limitation - the file needs to be opened in the VU==0. This is one decision and change.

Adding API/configuration that adds files (again in the initial VU as discussed in #1771) is IMO a separate discussion.

oleiade:
Heads-up that @codebien volunteered himself to take over this topic, and will be handling the discussions and work on that front from now on 🙇🏻

codebien:
We never prioritized to work on it. I'm removing myself as it is unrealistic that I will work soon on it.

PR #3370, opened 2023-10-05
Expose profiling endpoints

### LINKED ISSUE #1692: Add profiling endpoint
> Package pprof serves via its HTTP server runtime profiling data in the format expected by the pprof visualization tool. The package is typically only imported for the side effect of registering its HTTP handlers. The handled paths all begin with /debug/pprof/ [0]

I would suggest adding an environment variable. 

If set, k6 would export debugging information under the `/debug` path. In this case, the pprof information (`/debug/pprof`).

This is interesting in case someone wants to continuously profile k6 (with something like [Conprof](https://github.com/conprof/conprof) [1]) or manually use the [pprof visualization](https://github.com/google/pprof) tool [2].

[0] https://golang.org/pkg/net/http/pprof/
[1] https://github.com/conprof/conprof
[2] https://github.com/google/pprof

olegbespalov:
## What?

This change exposes profiling endpoints for the k6.

## Why?

This won't have significant performance implications (if the endpoints are not used), so we could expose them by default.

But having this endpoint allows efficient development of k6 (and extension) 

## Checklist

## Related PR(s)/Issue(s)

Closes #1692

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -270,10 +270,20 @@ func (c *cmdRun) run(cmd *cobra.Command, args []string) (err error) {
 		srvCtx, srvCancel := context.WithCancel(globalCtx)
 		defer srvCancel()
 
-		srv := api.GetServer(runCtx, c.gs.Flags.Address, testRunState, samples, metricsEngine, execScheduler)
+		srv := api.GetServer(
+			runCtx,
+			c.gs.Flags.Address, c.gs.Flags.ProfilingEnabled,
+			testRunState,
+			samples,
+			metricsEngine,
+			execScheduler,
+		)
 		go func() {
 			defer apiWG.Done()
 			logger.Debugf("Starting the REST API server on %s", c.gs.Flags.Address)
+			if c.gs.Flags.ProfilingEnabled {
+				logger.Debugf("Profiling exposed on http://%s/debug/pprof/", c.gs.Flags.Address)
+			}
 			if aerr := srv.ListenAndServe(); aerr != nil && !errors.Is(aerr, http.ErrServerClosed) {
 				// Only exit k6 if the user has explicitly set the REST API address
 				if cmd.Flags().Lookup("address").Changed {
--- a/api/server.go
+++ b/api/server.go
@@ -1,9 +1,12 @@
+// Package api contains the REST API implementation for k6.
+// It also registers the services endpoints like pprof
 package api
 
 import (
 	"context"
 	"fmt"
 	"net/http"
+	_ "net/http/pprof" //nolint:gosec // Register pprof handlers
 	"time"
 
 	"github.com/sirupsen/logrus"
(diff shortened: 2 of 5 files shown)


mstoykov (dismissed):
Not certain if we shouldn't make this configruable but otherwise LGTM!

My predominant problem wiht lack of configuration isn't even it being "expensive" to  have it around - it isn't.

But the fact that we add it on top of the `/api/v1` endpoint which we might drop at some future point and then anything using this needs to change. But I feel like it will probably be better at taht *future* point to leave this endpoint where it is and build v2 around it :shrug:

codebien (dismissed):
LGTM

oleiade (dismissed):
👏🏻

olegbespalov:
I am also unsure if we need to expose the profiling endpoints by default even if it doesn't harm (at least, I can't name the cases). So, if the team decides, I'll happily produce one more commit.

@oleiade @codebien, anybody else who sees that? 

Please vote on this comment:
* :+1: let's add the config option
* :-1: no need to option

oleiade:
I would align with the Grafana ecosystem and have config for it 👍🏻

codebien:
>grafana's projects where enabling profiles are usually controlled by setting

>even if it doesn't harm (at least, I can't name the cases).

Maybe we can find more rationale about _why_ in their original issues.

olegbespalov:
Hey there! I've implemented a flag `--profile-enabled` (it's pretty close to the one I saw in other grafana products). 

Unfortunately, it's not so easy to add the configuration via environment variables; for example, REST API  has no way to configure it through environment variables. But for the starter, that should be enough.

oleiade (dismissed):
LGTM 🚀 

As I tried the feature locally, I had to go and read the code to find out what `port` to use in the URL. 

I wonder if we could somehow display the profile URL as part of k6 output somehow. The dashboard project displays it as part of the `output` line at the moment. Feel free to merge without, and don't consider my comment blocking 👍🏻 
 
Screenshot 2023-11-07 at 17 58 27

cmd/run.go
```diff
@@ -270,10 +270,18 @@ func (c *cmdRun) run(cmd *cobra.Command, args []string) (err error) {
...
+			c.gs.Flags.Address, c.gs.Flags.ProfilingEnabled,
+			testRunState,
+			samples,
+			metricsEngine,
+			execScheduler,
+		)
 		go func() {
 			defer apiWG.Done()
 			logger.Debugf("Starting the REST API server on %s", c.gs.Flags.Address)
+			logger.Debugf("Profiling endpoints exposed: %v", c.gs.Flags.ProfilingEnabled)
```
mstoykov, line 284: I would prefer this to be behind an if instead of printing  a `false`/`true`

mstoykov (approved):

olegbespalov:
@oleiade @mstoykov I've tried to combine both of your suggestions under the [2089d5b](https://github.com/grafana/k6/commit/2089d5b10d88f369e3fc52a9c7c873b88f850171)

image


Would this make sense?

mstoykov (dismissed):
LGTM! 

I still kind of prefer to have a debug log as well, but I am okay either way.

oleiade (approved):
🚀 🚀 🚀

mstoykov (approved):

Issue #2968, opened 2023-03-09
Refactor `js.newFileConsole()` and e2e test `--console-output`


na--:
This code has a whole bunch of issues: https://github.com/grafana/k6/blob/333a3d1ed383b5e7cc931386155882c76a4f5b65/js/console.go#L12-L89

- it directly uses `os.OpenFile()`, so we can't test it with integration/e2e tests
- it initializes its own logger, and requires a `logrus.Formatter` as a parameter, making https://github.com/grafana/k6/issues/2958 much more difficult and requiring hacks like [this](https://github.com/grafana/k6/pull/2960/files#diff-0f675fa3e61b737c161f4231ff48d85831738d5ec7aea1223315c16c994b2e44R442-R447).
- [it is called in `js.Runner.SetOptions()`](https://github.com/grafana/k6/blob/333a3d1ed383b5e7cc931386155882c76a4f5b65/js/runner.go#L431-L441), which is probably a pretty terrible place to initialize such things...
- it seems like nothing actually closes the file...

PR #3380, opened 2023-10-11
fix: Set console log level to the application one when writing to file

mstoykov:
This fixes `console.debug()` messages printing with `--console-output file.txt` and `-v` given to k6 cli.

This already works when it isn't redirected to a file as it goes through the application wide logger.

### DIFF
--- a/js/console.go
+++ b/js/console.go
@@ -20,13 +20,14 @@ func newConsole(logger logrus.FieldLogger) *console {
 }
 
 // Creates a console logger with its output set to the file at the provided `filepath`.
-func newFileConsole(filepath string, formatter logrus.Formatter) (*console, error) {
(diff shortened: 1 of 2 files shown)


codebien (commented):
IMHO, we should address https://github.com/grafana/k6/issues/2968 and apply the fix at the same time. I can do an attempt as this is marked as v0.48.0.

mstoykov:
I only made this fix as investigating the issue directly pointed out a 3 lines fix.

The other issue has multiple things in it and even just the `os.OpenFile` one still seems like a thing that will take quite a big refactor not to speak adding and e2e test will be a lot more work./

So I still prefer to fix this now and leave the other issue for when we have more time or more of the problems have been resolved for other reasons.

codebien (approved):

oleiade (approved):

PR #3389, opened 2023-10-12
Remove the --logformat flag

codebien:
## What?

Removed the deprecated --logformat flag, the --log-format alternative should be used.

## Why?

Deprecation process.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -174,11 +174,6 @@ func rootCmdPersistentFlagSet(gs *state.GlobalState) *pflag.FlagSet {
 		"change the output for k6 logs, possible values are stderr,stdout,none,loki[=host:port],file[=./path.fileformat]")
 	flags.Lookup("log-output").DefValue = gs.DefaultFlags.LogOutput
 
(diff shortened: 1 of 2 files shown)


olegbespalov (approved):

olegbespalov (changes requested):
We need to migrate first: https://github.com/search?q=org%3Agrafana%20--logformat&type=code

mstoykov:
Given that this is still so widely used it might be worth postponing this a release or three after we remove it everywhere. 

This isn't really a problem we need to fix, but breaking user experiecne is not great if until yesterday we were telling them to a thing .

olegbespalov:
In my opinion, I don't think that we need to postpone this shrug:  We need to fix usage, and we have plenty of time.

However, I might think that following v0.48 could be one of the biggest from the perspective of breaking changes :sweat_smile:

codebien:
@olegbespalov the other parts have been changed, we should be now ready to merge this.

olegbespalov (approved):
🚀

codebien:
@oleiade as @mstoykov is on PTO, I add you as a reviewer.

oleiade (approved):

PR #3390, opened 2023-10-12
output/csv: Remove the support for snake_case options

codebien:
Removed the options using the snake_case format, instead use the alternative camelCase versions.

## What?

Removed the options using the snake_case format (e.g.`file_name`), instead use the alternative camelCase versions (e.g. `fileName`).

## Why?

Deprecation process.

### DIFF
--- a/output/csv/config.go
+++ b/output/csv/config.go
@@ -15,10 +15,9 @@ import (
 
 // Config is the config for the csv output
 type Config struct {
-	// Samples.
-	FileName     null.String        `json:"file_name" envconfig:"K6_CSV_FILENAME"`
-	SaveInterval types.NullDuration `json:"save_interval" envconfig:"K6_CSV_SAVE_INTERVAL"`
-	TimeFormat   null.String        `json:"time_format" envconfig:"K6_CSV_TIME_FORMAT"`
+	FileName     null.String        `json:"fileName" envconfig:"K6_CSV_FILENAME"`
+	SaveInterval types.NullDuration `json:"saveInterval" envconfig:"K6_CSV_SAVE_INTERVAL"`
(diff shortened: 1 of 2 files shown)


output/csv/config.go
```diff
@@ -15,10 +15,9 @@ import (
 
 // Config is the config for the csv output
 type Config struct {
-	// Samples.
-	FileName     null.String        `json:"file_name" envconfig:"K6_CSV_FILENAME"`
-	SaveInterval types.NullDuration `json:"save_interval" envconfig:"K6_CSV_SAVE_INTERVAL"`
-	TimeFormat   null.String        `json:"time_format" envconfig:"K6_CSV_TIME_FORMAT"`
+	FileName     null.String        `json:"fileName" envconfig:"K6_CSV_FILENAME"`
+	SaveInterval types.NullDuration `json:"saveInterval" envconfig:"K6_CSV_SAVE_INTERVAL"`
+	TimeFormat   null.String        `json:"timeFormat" envconfig:"K6_CSV_TIME_FORMAT"`
```
mstoykov, line 20: Do we use the json encoding on this anywhere?

olegbespalov (approved):

output/csv/config.go
codebien: I'm not aware of it, I changed it mostly for consistency

mstoykov: Hmm this is used in the config file ... Can we revert this as this is another breaking change.

I have no idea if somebody is *using* this, but breaking it seems bad to me.

mstoykov: Specifically here https://github.com/grafana/k6/blob/8457c91728bcdb3710628d27b8110be14fa73cfa/cmd/config.go#L46

PR #3394, opened 2023-10-13
Add k6 new subcommand

### LINKED ISSUE #3294: A subcommand to initialize a new test script file
### Feature Description

A new k6 script requires writing some boilerplate code. For new users this requires reading the documentation before being able to execute their first `k6 run`, which creates an additional entry barrier, while for experienced users it's a tedious task that could have been automated.

I propose to add a new `init` subcommand to `k6` that creates a JS file containing a minimal k6 script that can be executed with `k6 run`. As a content of this file I'd propose the [simple local script](https://k6.io/docs/get-started/running-k6/#running-local-tests) from the docs:

```javascript
import http from 'k6/http';
import { sleep } from 'k6';

export default function () {
  http.get('https://test.k6.io');
  sleep(1);
}
```

I'd also consider adding the code that enables running browser tests, i.e. the [`options.scenarios`](https://k6.io/docs/javascript-api/k6-experimental/browser/#example) section that enables Browser API, and have it commented out similar to what [`vagrant init`](https://developer.hashicorp.com/vagrant/docs/cli/init) does.

_No response_

andrewslotin:
## What?

This PR adds a new subcommand to `k6` that creates a new test script file and populates it with a sample script similar to the one in Grafana Cloud k6 script editor.

## Why?

This subcommand reduces the amount of boilerplate code required to author a new script. Sample code provides new k6 users with instructions on how to run this test in Grafana Cloud and/or enable Browser API by pointing them to appropriate places in documentation.

## Checklist

## Related PR(s)/Issue(s)

Closes #3294


cmd/init.go
```diff
@@ -0,0 +1,160 @@
...
+  // A number specifying the number of VUs to run concurrently.
+  vus: 10,
+  // A string specifying the total duration of the test run.
+  duration: '30s',
+
+  // The following section contains configuration options for execution of this
+  // test script in Grafana Cloud.
+  //
+  // See https://grafana.com/docs/grafana-cloud/k6/get-started/run-cloud-tests-from-the-cli/
+  // to learn about authoring and running k6 test scripts in Grafana Cloud.
```
codebien, line 33:
```suggestion
  // to learn about authoring and running k6 test scripts in Grafana k6 Cloud.
```

cmd/init.go
```diff
@@ -0,0 +1,160 @@
...
+  // },
+
+  // Uncomment this section to enable the use of Browser API in your tests.
+  //
+  // See https://k6.io/docs/using-k6-browser/running-browser-tests/ to learn more
+  // about using Browser API in your test scripts.
+  //
+  // scenarios: {
+  //   // The scenario name appears in the result summary, tags, and so on.
+  //   SCENARIO NAME: {
```
codebien, line 51:
```suggestion
  //   "browser example scenario": {
```
I would prefer if we provide only _ready to use_ code.

cmd/root.go
```diff
@@ -59,7 +59,7 @@ func newRootCommand(gs *state.GlobalState) *rootCommand {
 	subCommands := []func(*state.GlobalState) *cobra.Command{
 		getCmdArchive, getCmdCloud, getCmdInspect,
 		getCmdLogin, getCmdPause, getCmdResume, getCmdScale, getCmdRun,
-		getCmdStats, getCmdStatus, getCmdVersion,
+		getCmdStats, getCmdStatus, getCmdVersion, getCmdInit,
```
codebien, line 62: Can you add it following the A-z order, please?

cmd/init.go
```diff
@@ -0,0 +1,160 @@
...
+
+  # Create minimal k6 script in the current directory and store it in test.js
+  {{.}} init test.js
+
+  # Overwrite existing test.js with a minimal k6 script
+  {{.}} init -f test.js`[1:])
+
+	initCmd := &cobra.Command{
+		Use:   "init",
+		Short: "Initialize a new k6 script.",
```
codebien, line 145:
```suggestion
		Short: "Initialize a new k6 script",
```
It seems to me other commands are not setting a period, even if I agree with you, probably I would prefer consistency.

cmd/init.go
```diff
@@ -0,0 +1,160 @@
...
+
+	fileExists, err := fsext.Exists(c.gs.FS, target)
+	if err != nil {
+		return err
+	}
+
+	if fileExists && !c.overwriteFiles {
+		c.gs.Logger.Errorf("%s already exists", target)
+		return errFileExists
+	}
```
codebien, line 110: It prints the following log
```
ERRO[0000] sls.xt already exists
ERRO[0000] file already exists
exit status 255
```
maybe we can remove one of them
```suggestion
		return fmt.Errrof("The requested %s file already exists, please use the `--force` flag if you want overwrite it.", target)
	}
```
I think we could improve the message also. Probably you will have some better suggestions.

cmd/init.go
```diff
@@ -0,0 +1,160 @@
...
+	defer fd.Close() //nolint:errcheck
+
+	if err := defaultNewScriptTemplate.Execute(fd, initScriptTemplateArgs{
+		ScriptName: path.Base(target),
+	}); err != nil {
+		return err
+	}
+
+	printToStdout(c.gs, fmt.Sprintf("Initialized a new k6 test script in %s.\n", target))
+	printToStdout(c.gs, fmt.Sprintf("You can now execute it by running `%s run %s`.\n", c.gs.BinaryName, target))
```
codebien, line 125:
```suggestion
	printToStdout(c.gs, fmt.Sprintf("Initialized a new k6 test script in %s. You can now execute it by running `%s run %s`.\n", target, c.gs.BinaryName, target))
```
Do we really need to have it multiline?

cmd/init.go
```diff
@@ -0,0 +1,160 @@
...
+	if fileExists && !c.overwriteFiles {
+		c.gs.Logger.Errorf("%s already exists", target)
+		return errFileExists
+	}
+
+	fd, err := c.gs.FS.Create(target)
+	if err != nil {
+		return err
+	}
+	defer fd.Close() //nolint:errcheck
```
codebien, line 116:
```suggestion
	defer func() {
	    _ = fd.Close() // we may think to check the error and log
	}
```

cmd/init.go
```diff
@@ -0,0 +1,160 @@
...
+
+	if err := defaultNewScriptTemplate.Execute(fd, initScriptTemplateArgs{
+		ScriptName: path.Base(target),
+	}); err != nil {
+		return err
+	}
+
+	printToStdout(c.gs, fmt.Sprintf("Initialized a new k6 test script in %s.\n", target))
```
codebien, line 124: Should we consider printing a highlighted `target`? Maybe using the k6's purple?

cmd/init.go
```diff
@@ -0,0 +1,160 @@
...
+  // }
+};
+
+// The function that defines VU logic.
+//
+// See https://k6.io/docs/examples/tutorials/get-started-with-k6/ to learn more
+// about authoring k6 scripts.
+//
+export default function() {
+  http.get('http://test.k6.io');
```
codebien, line 72:
```suggestion
  http.get('https://test.k6.io');
```

Let's be good Web's citizens :)

codebien (changes requested):
Hey @andrewslotin,
thanks for addressing it :bow:

cmd/init.go
andrewslotin: I don't have a preference here 👍

cmd/init.go
andrewslotin: I've been thinking about this too and ended up with making this JSON incorrect intentionally. I want to draw user attention to this section and encourage them to read it through if they decide to opt-in for the Browser API. My concern is that if I use something like `ui` or `browser` there, this might give an impression that this is a mandatory name.

Another option I had is to have `exec` there and add another commented block with a browser-based test. WDYT?

cmd/init.go
andrewslotin: Yeah, I've added sentinel error later in the code to make `k6` exit with a non-zero code. The log line is redundant now, will remove 👍

cmd/init.go
```diff
@@ -0,0 +1,160 @@
...
+	"go.k6.io/k6/cmd/state"
+	"go.k6.io/k6/lib/fsext"
+)
+
+const defaultNewScriptName = "script.js"
+
+//nolint:gochecknoglobals
+var (
+	errFileExists            = errors.New("file already exists")
+	defaultNewScriptTemplate = template.Must(template.New("init").Parse(`import http from 'k6/http';
```
codebien, line 20:
```suggestion

const (
	defaultNewScriptTemplate = `import http from 'k6/http';
```
The error is not used anymore so we can remove it. I think we can move the string as a constant and create directly the template on the fly without allocating it for every command that has the benefit of not requiring the `nolint` directive.

codebien (changes requested):

cmd/init.go
andrewslotin: I usually opt for parsing templates at the very top because it has pretty small footprint — parsed templates tend to be singletones — while providing the benefit of validating templates without running the command, which makes testing easier.

cmd/init.go
andrewslotin: > I think you may expand this comment on the previous line making clear that the name is configurable and it is a good practice to pick a meaningful name for the customer's business context.

That's a way to solve it, but I'm quite pessimistic about how often people read comments unless things go wrong 😅

codebien: Let's see if the other reviewers have some other suggestions otherwise, I'm fine with your current proposal.

cmd/init.go
codebien: >it has pretty small footprint

It is secondary to me, I care more about the fact we are defining a global var. It sounds like we don't have a real need - as we are not using it. It seems there isn't any test using it.

andrewslotin: The one that asserts on the content of generated file does :) It just checks the major parts, not the full text, since it’s going to be a pain to maintain it otherwise.

.gitignore
```diff
@@ -10,6 +10,7 @@
 *.syso
 .idea
 *.DS_Store
+script.js
```
codebien, line 13:
```suggestion
# Excluding as it is the file generated by the k6 init command
script.js
```
or something similar so if someone isn't aware of the init command then they can understand why it is there

cmd/init.go
oleiade: I tend to agree with the point raised by @codebien, especially considering that this comment section starts with the statement "uncomment this section to enable". Technically, and to be nitpicky (for once), if I uncomment this section, the script won't work (right?).

> I'm quite pessimistic about how often people read comments

I see where you're coming from, but my own personal experience is that I do really appreciate comments in an automatically generated file to guide me forward. I don't immediately read them, but I do as soon as I'm blocked. My assumption is that the init file is there to get started quickly, and I anticipate that as users introduce complexity in their script, they might rely on comments more and more if they're present 😄

All of this to say I'm rather for having a valid name and expanding on the comment a bit more too 👍🏻

cmd/init.go
```diff
@@ -0,0 +1,163 @@
...
+
+  // Uncomment this section to enable the use of Browser API in your tests.
+  //
+  // See https://k6.io/docs/using-k6-browser/running-browser-tests/ to learn more
+  // about using Browser API in your test scripts.
+  //
+  // scenarios: {
+  //   // The scenario name appears in the result summary, tags, and so on.
+  //   SCENARIO NAME: {
+  //     // Mandatory parameter for browser-based tests.
```
oleiade, line 50: It might be worth clarifying whether the presence of the `executor` is necessary or whether the `shared-iterations` executor is mandatory 🥽

oleiade (dismissed):

cmd/init.go
```diff
@@ -0,0 +1,163 @@
...
+  // A string specifying the total duration of the test run.
+  duration: '30s',
+
+  // The following section contains configuration options for execution of this
+  // test script in Grafana Cloud.
+  //
+  // See https://grafana.com/docs/grafana-cloud/k6/get-started/run-cloud-tests-from-the-cli/
+  // to learn about authoring and running k6 test scripts in Grafana k6 Cloud.
+  //
+  // cloud: {
```
olegbespalov, line 33: Unfortunately, even if I'd like to have here a `cloud` it's still safer to have this PR with the `ext.loadimpact` :cry:

cmd/init.go
```diff
@@ -0,0 +1,163 @@
...
+	))
+
+	return nil
+}
+
+func getCmdInit(gs *state.GlobalState) *cobra.Command {
+	c := &initCmd{gs: gs}
+
+	exampleText := getExampleText(gs, `
+  # Create minimal k6 script.js in the current directory
```
olegbespalov, line 137: Just aligning with the other help, where mostly `k6 script` mentioned

```suggestion
  # Create a minimal k6 script in the current directory. By default, k6 creates script.js.
```

cmd/init.go
```diff
@@ -0,0 +1,163 @@
...
+  {{.}} init
+
+  # Create minimal k6 script in the current directory and store it in test.js
```
olegbespalov, line 140:
```suggestion
  # Create a minimal k6 script in the current directory and store it in test.js
```

cmd/init.go
```diff
@@ -0,0 +1,163 @@
...
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+	"go.k6.io/k6/cmd/state"
+	"go.k6.io/k6/lib/fsext"
+)
+
+const defaultNewScriptName = "script.js"
+
+//nolint:gochecknoglobals
+var defaultNewScriptTemplate = template.Must(template.New("init").Parse(`import http from 'k6/http';
```
olegbespalov, line 18: Have you considered moving this into a separate file(s)?

It also could benefit us by introducing one more flag that can control the template and just running the 

```
k6 init
```
will produce a simple example
```javascript
import http from 'k6/http';
import { sleep } from 'k6';

export default function() {
  http.get('https://test.k6.io');

  sleep(1);
}

```

by running the

```
k6 init --template cloud
```
Produces the script with the cloud options set up, browser one produces the script with some browser usage, and so on...

My rationale is that this could produce a simpler script that could be more readable shrug:

olegbespalov (commented):
Looks good. I left a few minor comments and one suggestion :+1:

cmd/init.go
andrewslotin: @olegbespalov, could you clarify? My assumption was that this subcommand will land in the same release where we add support for `options.cloud`. I'm fine with changing it to be `options.ext.loadimpact` though and let you change it again in your PRs.

olegbespalov: Your code will probably be merged first, so producing a working example based on the current state is better. The current state is that there is no such thing `options.cloud`. 

It's not a big deal to rebase the branch that introduces the `options.cloud`  on top of the master after merging this branch (or vice-versa). But the change itself should be atomic.

cmd/init.go
andrewslotin: We can definitely iterate on the script content, but I'd like to do it outside of the scope of this PR. I did consider adding a flag to skip the comments, similar to `vagrant init --minimal`, but then decided to not to include to keep the PR focused on introducing the command. My main target was to provide users who are new to k6 with some directions and structure to author their first script. I believe this also simplifies initializing a new script file for experienced users, since the cognitive load required to remove a comment is lower than to write a few lines of valid JS code :) But, again, I think this is a good idea that is fairly simple to implement, so I'm up for contributing in separately.

cmd/init.go
codebien: I see the risk that for whatever reason we will end up suggesting the old option. For example, for whatever reason we need to roll back the Cloud option.
What about dividing the current PR in two? One adds the command and one adds the specific Cloud part and it will be merged only when we have the new option.
I'm really against with us having `k6 init` with loadimpact inside.

andrewslotin: @codebien, not sure what's the risk there. My suggestion was to introduce `k6 init` with `options.ext.loadimpact` and change it to `options.cloud` as a part of @olegbespalov's PR. This way if we decide to roll back the `options.cloud`, it will also revert script sample to use the old version.

codebien: >it will also revert script sample to use the old version.

That's my point. In that case, we will end by boosting the use of the old option.

andrewslotin: Right, but if we revert the support for the new one, we should not be suggesting it, right? Anyway, I don't think we will be able to safely revert the config changes anyway, but I see your concern.

cmd/init.go
```diff
@@ -0,0 +1,169 @@
...
+  //   loadimpact: {
+  //     // The ID of the project to which the test is assigned in the k6 Cloud UI.
+  //     // By default tests are executed in default project.
+  //     projectID: "",
+  //     // The name of the test in the k6 Cloud UI.
+  //     // Test runs with the same name will be grouped.
+  //     name: "{{ .ScriptName }}"
+  //   }
+  // },
+
```
codebien, line 43:
```suggestion
```

codebien (dismissed):
I'm fine with most of the changes as I commented before I would prefer if we not include here the Cloud suggestion with the old option.

cmd/init.go
codebien: I meant rollback during the dev cycle.

- We merge here
- We merge the cloud option
- We need to rollback the cloud option
- We release

With the current version, we will end up suggesting the old way. With my suggestion, we will not provide the Cloud block.

andrewslotin: I see, but even without `options.cloud` we still want to enable users to run their scripts in the cloud.

olegbespalov: I think promoting the `options.cloud` is a different topic. It's not dependent on the #3348, but rather on the state of our backend, which is the totally opposite thing. So I'd suggest keeping it simple and just accepting in the scope of this PR that only `options.ext.loadimact` exists. 

I see Ivan's point that if we introduce the `k6 init`, it could bring "launch" attention which could produce some scripts that later should be adjusted, but honestly still believe that the amount is less than the ones that are produced from the examples (docs, grafana, etc).

codebien: It seems you have a consensus on it so I'm fine with you :+1:

olegbespalov (dismissed):

na--:
Hey, spotted this by chance when I was skimming notification emails, and I was immediately confused by the PR title, specifically about what is being initialized by `k6 init` :sweat_smile: Considering we have an ["init context"](https://k6.io/docs/javascript-api/init-context/) in the [test lifecycle](https://k6.io/docs/using-k6/test-lifecycle/), I thought it had something to do with that or even https://github.com/grafana/k6/issues/785 :sweat_smile: 

My 2c: having such a sub-command to is probably a good idea. Though I very rarely make use of such commands in other tools, I know other people find them useful. However, the `k6 init` sub-command might be somewhat confusing because we already have other things that we call "init". `k6 create` or `k6 new` or something like that seems less likely to be confusing.

andrewslotin:
I like the idea of `k6 new`! Indeed, while `init` seems to be widely adopted by other tools, it may convey false assumption that it can be used only once per project, while the idea is to have it a standard way to create a new script. 

@olegbespalov, @codebien, what are your thoughts on this? I’d be up for making necessary renamings.

codebien:
>what are your thoughts on this?

I don't have a strong opinion here, `k6 new` has the benefit of being shorter so I am fine with it.

andrewslotin:
Renamed `k6 init` to `k6 new` as per discussion above.

andrewslotin:
OK, there is one more edit: I've replaced the link to k6.io/docs with Grafana docs ones, since we're migrating our documentation there.

olegbespalov (approved):
I don't have a strong opinion about the `init` vs. `new`.

codebien (approved):
I'm quite happy with the final solution. Thanks for the effort. :bow: @andrewslotin

oleiade (approved):
🚀 🚀 🚀

Issue #3258, opened 2023-08-08
cloud: Set v2 as the default output


codebien:
### Feature Description

#3117 added a new experimental output for the cloud.

When the new experimental output is evaluated as stable based on the following conditions we should set it as the default output for the cloud.

Acceptance criteria

1. Any critical errors have been observed
2. Any negative feedback has been provided

### Suggested Solution (optional)

Set `2` in the default config https://github.com/grafana/k6/blob/fcb19fbd8c82dc12f725c0d4be9a033b7c5b8968/cloudapi/config.go#L179

### Already existing or connected issues / PRs (optional)

#3117
#2954

PR #3400, opened 2023-10-16
output/cloud: Set v2 as the default

codebien:
## What?

It sets the default version for the Cloud output as `v2`.

### Related PR(s)/Issue(s)

Closes https://github.com/grafana/k6/issues/3258

### DIFF
--- a/cloudapi/config.go
+++ b/cloudapi/config.go
@@ -162,7 +162,7 @@ type Config struct {
 
 // NewConfig creates a new Config instance with default values for some fields.
 func NewConfig() Config {
-	return Config{
+	c := Config{
 		Host:                  null.NewString("https://ingest.k6.io", false),
 		LogsTailURL:           null.NewString("wss://cloudlogs.k6.io/api/v1/tail", false),
 		WebAppURL:             null.NewString("https://app.k6.io", false),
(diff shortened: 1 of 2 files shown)


mstoykov (dismissed):

cloudapi/config.go
```diff
@@ -196,6 +196,14 @@ func NewConfig() Config {
 		AggregationOutlierIqrCoefLower: null.NewFloat(1.5, false),
 		AggregationOutlierIqrCoefUpper: null.NewFloat(1.3, false),
 	}
+
+	// v2 enables aggregation by default
+	if c.APIVersion.Int64 == 2 {
```
olegbespalov, line 201: A non-blocking suggestion is doing the opposite way. 

As we know, the v2 is the current default, so `AggregationWaitPeriod` and `AggregationWaitPeriod`  could be aligned with the v2 (enabled aggregation) in the initializing of config above, but the previous default values could be set if the APIVersion is 1 (error-prone) in a similar to that logic block.

Later when we'd like to remove the v1, it's easier and clearer that only the logical block `if c.APIVersion.Int64 == 1` could be removed.

olegbespalov (dismissed):
LGTM, with a minor non-blocking comment.

output/cloud/output_test.go
```diff
@@ -126,7 +126,7 @@ func TestOutputCreateTestWithConfigOverwrite(t *testing.T) {
 	require.NoError(t, out.Start())
 
 	assert.Equal(t, types.NullDurationFrom(10*time.Millisecond), out.config.MetricPushInterval)
-	assert.Equal(t, types.NullDurationFrom(30*time.Millisecond), out.config.AggregationPeriod)
+	assert.Equal(t, types.NullDurationFrom(1*time.Second), out.config.AggregationPeriod)
```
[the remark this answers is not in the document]
codebien: There is, the aggregation in v2 can't be sub-second precision https://github.com/grafana/k6/blob/44e1e63aadb66784ff0a12b8d9821a0fdc9e7467/output/cloud/expv2/collect.go#L69-L71

cloudapi/config.go
codebien: Yeah, I liked the suggestion. I submitted a change to address it.

olegbespalov (approved):

mstoykov (approved):

PR #3407, opened 2023-10-19
Warnings about options.cloud

olegbespalov:
## What?

These changes aim to start promoting the `options.cloud`. It's the state when it's safe to say to customers that it's safe to use. 

These changes were originally part of #3348, but there is a chance that we could merge #3348, but we won't be ready to start offering usage of the `options.cloud`.

## Why?

Reduce the scope and try to make #3348 mergable.

## Checklist

- [ ] **Check the status** #3394 that could bring additional usage of the `options.ext.loadimapct` and adjust this PR if need
- [ ] Ensure that we can promote the `options.cloud` (backend ready)

## Related PR(s)/Issue(s)

* #3348 
* #3394

### DIFF
--- a/cmd/cloud.go
+++ b/cmd/cloud.go
@@ -111,13 +111,18 @@ func (c *cmdCloud) run(cmd *cobra.Command, args []string) error {
 	}
 
 	// Cloud config
-	cloudConfig, _, err := cloudapi.GetConsolidatedConfig(
+	cloudConfig, warn, err := cloudapi.GetConsolidatedConfig(
 		test.derivedConfig.Collectors["cloud"], c.gs.Env, "", arc.Options.Cloud, arc.Options.External)
 	if err != nil {
 		return err
 	}
 	if !cloudConfig.Token.Valid {
 		return errors.New("Not logged in, please use `k6 login cloud`.") //nolint:golint,revive,stylecheck
+	}
+
+	// Display config warning if needed
+	if warn != "" {
+		modifyAndPrintBar(c.gs, progressBar, pb.WithConstProgress(0, "Warning: "+warn))
 	}
 
 	if cloudConfig.Token.Valid {
--- a/cmd/login_cloud.go
+++ b/cmd/login_cloud.go
@@ -55,10 +55,14 @@ This will set the default token used when just "k6 run -o cloud" is passed.`,
 
 			// We want to use this fully consolidated config for things like
 			// host addresses, so users can overwrite them with env vars.
-			consolidatedCurrentConfig, _, err := cloudapi.GetConsolidatedConfig(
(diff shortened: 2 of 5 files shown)


codebien (changes requested):
Blocking as we will have to change the part related to https://github.com/grafana/k6/pull/3394 when it is merged.

But, at this point, I would like to make another attempt in considering to directly introduce the block for the cloud here.

Now we have this PR, @olegbespalov @andrewslotin do you still want to introduce the block there? This PR has a limited scope that fits the case.

olegbespalov:
If the question is "Do you still want to use `options.ext.loadimact` in the #3394?" then I do belive that using a legacy block is still the right thing since it guarantees atomically of the changes.

All parties are also aware of this problem of syncing between these two PRs, and we will do the right things depending on the pace.

I'm not sure this PR should be considered blocked, but I'm not against keeping it like that to stress the importance of some actions that could happen once the #3394 is merged. Also, I'm putting the note to the PR's first message to emphasize the importance of checking.

codebien:
>These changes were originally part of https://github.com/grafana/k6/pull/3348, but there is a chance that we could merge https://github.com/grafana/k6/pull/3348, but we won't be ready to start offering usage of the options.cloud.

I took the assumption that this PR was already blocked from the quoted condition to be verified. Isn't that? Do you intend to merge it without being sure that we are going to offer the usage of the option? Then eventually revert only this PR?

olegbespalov:
>I took the assumption that this PR was already blocked from the quoted condition to be verified. Isn't that? Do you intend to merge it without being sure that we are going to offer the usage of the option? Then eventually revert only this PR?

It's blocked because of that. That's true. Let me be explicit in the PR body.

But you wrote above:

>Blocking as we will have to change the part related to https://github.com/grafana/k6/pull/3394 when it is merged.

I was answering when I referring that I don't agree that it's blocked.

olegbespalov:
Also, I converted it to a daft since it waits for some internal decisions.

codebien (dismissed):

mstoykov (approved):
I do wonder if we should do this in the same version we introduce it.

Maybe have a version with it and let us see if there aren't any bugs will be better.

This will also let us go update it ... everywhere - which will be a major problem IMO with all of our documentation and examples

codebien:
>Maybe have a version with it and let us see if there aren't any bugs will be better.

Any bug on this specific feature will require us to hotfix. Plus, on the other side, I also see the risk that if we don't promote it then we may not get exposure and the users will not hit any eventual bugs.
So, considering that it isn't a blocker, and in any case, we have to fix it fast, I'm for just merging it.

oleiade (dismissed):
I align with @codebien ☝🏻 and think we should probably just merge it to get it some visibility indeed 🙇🏻

mstoykov (approved):

codebien (approved):

PR #3410, opened 2023-10-21
Fixed the tests reference in the all rule of the Makefile

flyck:
## What?

Tiny fix for the makefile.

Current behavior:
```bash
❯ make all
cleaning
make: *** No rule to make target `test', needed by `all'.  Stop.
```

New behavior: Runs all commands

## Why?

Make all should succeed

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 MAKEFLAGS += --silent
 GOLANGCI_LINT_VERSION = $(shell head -n 1 .golangci.yml | tr -d '\# ')
 
-all: clean format test build
+all: clean format tests build
 
 ## build: Builds the 'k6' binary.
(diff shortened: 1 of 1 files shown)


mstoykov:
Thanks for this catch and fix @flyck :bow: 

Unfortunately we won't be able to merge this without you signing the cla :(.

I will wait for you for a few days, but if it is not possible I guess we will do this change ourself, sorry for the inconvenience :(

codebien (approved):
LGTM, if the CLA will be signed

flyck:
Thanks guys, turns out I had the wrong email in the commit which is why the CLA was not showing as signed. Just force-pushed that author change ☝🏻

PR #3413, opened 2023-10-23
Add waitForEvent tasks to release notes

ankur22:
## What?

Update the next release notes with links to the PRs that have fixed and refactored code to get `waitForEvent` to work.

## Why?

To update the release notes for the next release.

## Checklist

### DIFF
--- a/release notes/v0.48.0.md
+++ b/release notes/v0.48.0.md
@@ -30,12 +30,14 @@ _what, why, and what this means for the user_
 ## Bug fixes
 
 - [browser#1077](https://github.com/grafana/xk6-browser/pull/1077) Fixes `browserContext.clearPermissions` to clear permissions without panic.
+- [browser#1042](https://github.com/grafana/xk6-browser/pull/1042) Fixes `browserContext.waitForEvent` which involved promisifying the `waitForEvent` API.
 
 ## Maintenance and internal improvements
 
 - [browser#1075](https://github.com/grafana/xk6-browser/pull/1075), [browser#1076](https://github.com/grafana/xk6-browser/pull/1076) Refactors `clearPermissions` and `grantPermissions`.
 - [browser#1043](https://github.com/grafana/xk6-browser/pull/1043) Refine tests.
 - [browser#1047](https://github.com/grafana/xk6-browser/pull/1047) Update dependencies.
+- [browser#1069](https://github.com/grafana/xk6-browser/pull/1069) Internal refactors.
 
 ## _Optional_ Roadmap
 


release notes/v0.48.0.md
```diff
@@ -33,8 +33,7 @@ _Format as `<number> <present_verb> <object>. <credit>`_:
 
 ## Bug fixes
 
-_Format as `<number> <present_verb> <object>. <credit>`_:
-- _`#111` fixes race condition in runtime_
+- [browser#1042](https://github.com/grafana/xk6-browser/pull/1042), [browser#1069](https://github.com/grafana/xk6-browser/pull/1069) Fixes `browserContext.waitForEvent`. This fix involved promisifying the `waitForEvent` API as well as various internal refactors.
```
inancgumus, line 36: Is this a breaking change? 🤔 Should we move it to breaking changes if so?

ankur22: Good point, i guess it could go in breaking changes since we're making it async now. I think though, that this feature never really worked even when it was synchronous (which is likely a sign that maybe it's not that important), so I still think it's better left where it is. WDYT?

inancgumus: Yep, I'm in between, too 😒

inancgumus (approved):

release notes/v0.48.0.md
ka3de, line 36: My understanding is that https://github.com/grafana/xk6-browser/pull/1069 is not specifically related with `waitForEvent`, as it also refactors the usage of `TaskQueue` for `Page.on`. Should we remove it from here and instead add an entry for it under `Maintenance and internal improvements` section?

ankur22: Yeah, good point, it does include quite a bit of refactoring to move the taskqueue out of the common package and into the mapping layer. The main motivation though was due to the issue of not working with the `taskqueue` when using the predicate function for `waitForEvent`, i.e. without this change `waitForEvent` wouldn't work. Happy to move it to `Maintenance and internal improvements` though if you still think it should, WDYT?

ankur22: Ok, happy with that change added it to a fixup: ef9213368a9a5b7010584fd532d97c329a40b791

ka3de: How specific should we be with these internal refactors? To be honest I'm never sure of what should be the level of detail in these cases. Considering that this is a completely internal change, I don't disagree with the current text, but I wonder if it should be more specific such as _"Internal refactor to move Goja's TaskQueue away from core implementation"_ (or similar).

Since you are a reviewer for this also, any thoughts @oleiade ?

ankur22: Mind if i merge this and we can open a new issue for this if we need?

ka3de (approved):
LGTM.

Issue #3257, opened 2023-08-08
Exception in `handleSummary()` do not have a stack


mstoykov:
### Brief summary

Exceptions in handleSummary do not print stack, just a message that there was an exception. 

This makes debugging it way harder than it should

### k6 version

v0.45.1

### OS

linux

### Steps to reproduce the problem

```javascript
export default () => { };
export function handleSummary() {
        throw "here";
}
```

### Expected behaviour

```
ERRO[0000] here
        at handleSummary (file:///home/mstoykov/work/k6io/k6/handlesummary.js:3:1(2))  hint="script exception"
```

### Actual behaviour

```
ERRO[0000] handleSummary() failed with error "here", falling back to the default summary  source=console
```

Azanul:
May I take this up?

mstoykov:
@Azanul you are welcome to do so :+1: and sorry for the slow reply

Issue #2958, opened 2023-03-07
Replace `github.com/sirupsen/logrus`


na--:
[github.com/sirupsen/logrus](https://github.com/sirupsen/logrus) is the structured logging library that has been used in k6 since literally the second commit in the repo, https://github.com/grafana/k6/commit/9d2b96c89b0f425112962325213419764e3f8baa :exploding_head: 

However, it has its fair share of API issues that we have had to work around, it doesn't have the greatest performance or feature set, and it has been in maintenance mode for the last 3+ years (https://github.com/sirupsen/logrus/commit/4ecd9a62bd57b6533e0d87ae23a6b10adc32b5ce). So it has been a low-key goal to replace it with something else for a while...

There are [a bunch of other structured logging Go libraries](https://github.com/avelino/awesome-go#logging) out there that are better in various ways. Previous discussions and issues like https://github.com/grafana/k6/issues/216 advocated for switching to one of them. However, with the proposal from https://github.com/golang/go/issues/56345, it seems likely that Go will have structured logging support in the standard library at some future point :tada: I haven't looked at the proposal in detail, but it seemed reasonable at a very quick glance and it already has an implementation in https://pkg.go.dev/golang.org/x/exp/slog that people can play around with! So, barring any show-stopping issues, I'd probably prefer to switch k6 to use it instead of another external dependency...

Regardless of which library we choose to go with, the initial steps we need to do seem fairly clear to me. I think we should first do a few small refactors of the k6 code and get rid of any non-interface logrus types we rely one, especially the [`*logrus.Logger`](https://pkg.go.dev/github.com/sirupsen/logrus#Logger) variables we have in a few central places like these: 
https://github.com/grafana/k6/blob/0b1fe61c5762c83c52a2378f6cdf9c0c3ff1f53d/cmd/state/state.go#L53
https://github.com/grafana/k6/blob/0b1fe61c5762c83c52a2378f6cdf9c0c3ff1f53d/lib/test_state.go#L19-L20
https://github.com/grafana/k6/blob/0b1fe61c5762c83c52a2378f6cdf9c0c3ff1f53d/lib/vu_state.go#L38-L39

[...]

PR #3416, opened 2023-10-24
js: Print stack trace when exception in handleSummary()

joanlopez:
## What?

It splits the execution of the `handleSummary()` wrapper, so if there's an exception on the script, it prints the stack trace first, then falls back to the default summary.

It also moves the logic to print `errext.HasHint` and `errext.Exception` details for a given error into a shared function within the `errext` package, following *"the fprint pattern"* (that accepts a "writer", in this case a "logger"), so it can be reused where it was used until now ( `rootCommand`) and in the aforementioned exception handling.

## Why?

Currently, exceptions in `handleSummary()` do not print stack, just a message that there was an exception, and that makes debugging it way harder than it should.

## Checklist

## Related PR(s)/Issue(s)

Closes https://github.com/grafana/k6/issues/3257

## Other notes

I haven't added any new test because if I'm not wrong, I think all the logic changed is already covered by tests present in `summary_test.go`, like `TestExceptionInHandleSummaryFallsBackToTextSummary` that precisely asserts that the default summary is used as summary when the `handleSummary()` throws an exception.

In fact, if I'm not wrong, the main behavior should not change, it's just a matter of bringing feedback to the user (logs) in a slightly different way (more clear/explicit).

Thanks!


errext/print.go
```diff
@@ -0,0 +1,30 @@
+package errext
+
+import (
+	"errors"
+
+	"github.com/sirupsen/logrus"
```
joanlopez, line 6: It might be good idea to keep this package free of external dependencies, in such case I'm completely open to suggestions for a new location for this function.

errext/print.go
```diff
@@ -0,0 +1,30 @@
...
+)
+
+// Fprint formats the given error and writes it to l.
+// In case of Exception, it uses the stack trace as the error message.
+// In case of HasHint, it also adds the hint as a field.
+func Fprint(l logrus.FieldLogger, err error) {
```
joanlopez, line 12: I could add more fields as a variadic argument (might be useful for future uses), so in this case I could add a `msg` telling the user that despite of the exception, it's falling back to the default summary.

js/runner.go
```diff
@@ -388,20 +388,25 @@ func (r *Runner) HandleSummary(ctx context.Context, summary *lib.Summary) (map[s
...
 	}()
 	vu.moduleVUImpl.ctx = summaryCtx
 
+	callbackResult := goja.Undefined()
+	fn := vu.getExported(consts.HandleSummaryFn)
+	if fn != nil {
+		if handleSummaryFn, ok := goja.AssertFunction(fn); ok {
+			callbackResult, _, _, err = vu.runFn(summaryCtx, false, handleSummaryFn, nil, vu.Runtime.ToValue(summaryDataForJS))
+			if err != nil {
+				errext.Fprint(r.preInitState.Logger, err)
```
joanlopez, line 403: Alternatively to what I suggested at https://github.com/grafana/k6/pull/3416/files#r1370043306, I could add another log line here, giving that feedback to the user in a different line (and maybe as `debug`?).

js/runner.go
```diff
@@ -388,20 +388,25 @@ func (r *Runner) HandleSummary(ctx context.Context, summary *lib.Summary) (map[s
...
+			}
+		} else if fn != nil {
+			return nil, fmt.Errorf("exported identifier %s must be a function", consts.HandleSummaryFn)
+		}
```
mstoykov, line 407: nit: we usually prefer to nest less. In this case you can check that `!ok` first and error out with this error.  

If we are in this if `fn != nil` is already true ;)

errext/print.go
mstoykov: Yeah but as explained  in https://github.com/grafana/k6/issues/2958 it might not be that easy todo.

I feel like we either:
1. move this to some other package (`lib` is the usual junk drawer)
2. make the helper here return `hint`/`fields` (`map[string]string`) and the `errText`

mstoykov (commented):
LGTM in general :clap: 

Can we maybe try option 2 from https://github.com/grafana/k6/pull/3416/files#r1370141890

mstoykov:
> In fact, if I'm not wrong, the main behavior should not change, it's just a matter of bringing feedback to the user (logs) in a slightly different way (more clear/explicit). 

That is true, but maybe we can extend https://github.com/grafana/k6/blob/036f8a041429f690fb5caf07c34093bc2c55915d/js/summary_test.go#L669 to test that we do actually get the stacktrace

errext/print.go
joanlopez: Good idea, let's go with opt. 2, [as you suggested](https://github.com/grafana/k6/pull/3416#discussion_r1370142485).

js/runner.go
joanlopez: I also prefer it! It's just that as the code was there more less like that, I simply forgot to simplify it, good catch, thanks!

joanlopez:
> > In fact, if I'm not wrong, the main behavior should not change, it's just a matter of bringing feedback to the user (logs) in a slightly different way (more clear/explicit).
> 
> That is true, but maybe we can extend
> 
> https://github.com/grafana/k6/blob/036f8a041429f690fb5caf07c34093bc2c55915d/js/summary_test.go#L669
> 
> to test that we do actually get the stacktrace

Changed, as per what discussed "offline", thanks!

mstoykov (dismissed):

errext/format.go
```diff
@@ -0,0 +1,28 @@
+package errext
+
+import (
+	"errors"
+)
+
+// Format formats the given error and writes it to l.
```
oleiade, line 7: Could it be that this description of what the function does is somewhat outdated? I don't see any `l` variable in this function 😸

oleiade (approved):
LGTM 👍🏻 👏🏻 Great job @joanlopez 🙇🏻

errext/format.go
joanlopez: Great catch! Fixed! 🙏🏻

errext/format.go
```diff
@@ -0,0 +1,28 @@
+package errext
+
+import (
+	"errors"
+)
+
+// Format formats the given error as a message (string) and a map of fields.
+// In case of Exception, it uses the stack trace as the error message.
+// In case of HasHint, it also adds the hint as a field.
```
mstoykov, line 9:
```suggestion
// In case of [Exception], it uses the stack trace as the error message.
// In case of [HasHint], it also adds the hint as a field.
```

This should make this link within godoc

mstoykov (dismissed):

oleiade (dismissed):

mstoykov (approved):

oleiade (approved):

PR #3420, opened 2023-10-25
Address potential races in `fs.read` and `fs.seek` operations

### LINKED ISSUE #3354: `fs.read` race condition when using `Promise.all`
## Description

The current implementation of the **experimental/fs** module `read` (and potentially `seek`) method seems vulnerable to race conditions under specific circumstances. Below are the issues identified:

1. Concurrent reads on the same buffer can lead to unpredictable behavior

```javascript
let file = fs.open(name);
let buffer = new Uint8Array(5);
let p1 = file.read(buffer);
let p2 = file.read(buffer);
await Promise.all([p1, p2]);
```

2. A buffer modification prior to the resolution of a promise can lead to uncertainty about the final state of the buffer:
```javascript
let file = fs.open(name);
let buffer = new Uint8Array(5);
let p1 = file.read(buffer);
buffer[2] = 3;
console.log(buffer); 
await p1;
```

In this case, it's unclear if the buffer will hold the data read from the file, just the third value `3`, or a combination of both.

This behavior seems inconsistent with the promise paradigm, where state changes should only be observable once the promise resolves.

## Additional Observations

Preliminary tests with Deno suggest that:
- Sequential read calls populate data once one of the promises is awaited.
- Awaiting a promise from an already completed read doesn't affect the buffer.
- The buffer remains unchanged if a read promise isn't awaited.

Reviewing how Deno and Node handle such situations might be beneficial for insight.

## Concerns

While the current behavior might be acceptable for an experimental module, allowing race conditions in stable code could lead to unpredictable behavior and potential bugs. **We expect this issue to be addressed as a prerequisite for the `experimental/fs` module to become stable.**

## References

#2977
#3141 
#3309
#3145
#3148

oleiade:
## What?

This Pull Request aims to address #3354, and avoid data races involved in using the [currently proposed implementation of `fs.read` and `fs.seek`](#3309).

## Why?

Namely, it modifies the behavior of the proposed `fs.read` and `fs.seek` methods to avoid data races due to:
1. accessing the `fs.read` buffer argument concurrently
2. updating the file offset concurrently when calling seek

In a nutshell, we addressed _1_ by ensuring that the provided buffer was only modified from the JS runtime's main thread, by performing it as a side-effect of the promise resolution's. We addressed _2_ by making sure the internal file structure's offset is an atomic int that cannot, as a result, be modified by two concurrent operations at the same time.

## Checklist

## Related PR(s)/Issue(s)

ref #3309
ref #2974 
closes #3354

### DIFF
--- a/js/modules/k6/experimental/fs/file.go
+++ b/js/modules/k6/experimental/fs/file.go
@@ -3,6 +3,7 @@ package fs
 import (
 	"io"
 	"path/filepath"
+	"sync/atomic"
 )
 
 // file is an abstraction for interacting with files.
@@ -13,7 +14,7 @@ type file struct {
 	data []byte
 
 	// offset holds the current offset in the file
-	offset int
+	offset atomic.Int64
 }
 
 // Stat returns a FileInfo describing the named file.
@@ -38,14 +39,16 @@ type FileInfo struct {
 //
 // If the end of the file has been reached, it returns EOFError.
 func (f *file) Read(into []byte) (n int, err error) {
-	start := f.offset
-	if start == len(f.data) {
+	offset := f.offset.Load()
+
+	start := offset
+	if start == f.size() {
 		return 0, newFsError(EOFError, "EOF")
 	}
 
-	end := f.offset + len(into)
-	if end > len(f.data) {
-		end = len(f.data)
+	end := offset + int64(len(into))
+	if end > f.size() {
+		end = f.size()
 		// We align with the [io.Reader.Read] method's behavior
 		// and return EOFError when we reach the end of the
 		// file, regardless of how much data we were able to
@@ -55,7 +58,7 @@ func (f *file) Read(into []byte) (n int, err error) {
 
 	n = copy(into, f.data[start:end])
 
-	f.offset += n
+	f.offset.Store(int64(n))
 
 	return n, err
 }
@@ -71,9 +74,10 @@ var _ io.Reader = (*file)(nil)
 //
 // When using SeekModeStart, the offset must be positive.
 // Negative offsets are allowed when using `SeekModeCurrent` or `SeekModeEnd`.
-func (f *file) Seek(offset int, whence SeekMode) (int, error) {
-	newOffset := f.offset
+func (f *file) Seek(offset int64, whence SeekMode) (int64, error) {
+	startingOffset := f.offset.Load()
 
+	newOffset := startingOffset
 	switch whence {
 	case SeekModeStart:
 		if offset < 0 {
@@ -88,7 +92,7 @@ func (f *file) Seek(offset int, whence SeekMode) (int, error) {
 			return 0, newFsError(TypeError, "offset cannot be positive when using SeekModeEnd")
 		}
 
-		newOffset = len(f.data) + offset
+		newOffset = f.size() + offset
 	default:
 		return 0, newFsError(TypeError, "invalid seek mode")
 	}
@@ -97,17 +101,19 @@ func (f *file) Seek(offset int, whence SeekMode) (int, error) {
 		return 0, newFsError(TypeError, "seeking before start of file")
 	}
 
-	if newOffset > len(f.data) {
+	if newOffset > f.size() {
 		return 0, newFsError(TypeError, "seeking beyond end of file")
 	}
 
 	// Note that the implementation assumes one `file` instance per file/vu.
 	// If that assumption was invalidated, we would need to atomically update
 	// the offset instead.
-	f.offset = newOffset
+	f.offset.Store(newOffset)
 
 	return newOffset, nil
 }
+
+var _ io.Seeker = (*file)(nil)
 
 // SeekMode is used to specify the seek mode when seeking in a file.
 type SeekMode = int
@@ -125,3 +131,7 @@ const (
 	// the end of the file.
 	SeekModeEnd
 )
+
+func (f *file) size() int64 {
+	return int64(len(f.data))
+}
--- a/js/modules/k6/experimental/fs/file.go
+++ b/js/modules/k6/experimental/fs/file.go
@@ -4,6 +4,8 @@ import (
 	"io"
 	"path/filepath"
 	"sync/atomic"
+
+	"go.k6.io/k6/lib"
 )
 
 // file is an abstraction for interacting with files.
@@ -39,26 +41,27 @@ type FileInfo struct {
 //
 // If the end of the file has been reached, it returns EOFError.
 func (f *file) Read(into []byte) (n int, err error) {
-	offset := f.offset.Load()
+	currentOffset := f.offset.Load()
+	fileSize := f.size()
 
-	start := offset
-	if start == f.size() {
+	// Check if we have reached the end of the file
+	if currentOffset == fileSize {
 		return 0, newFsError(EOFError, "EOF")
 	}
 
-	end := offset + int64(len(into))
-	if end > f.size() {
-		end = f.size()
-		// We align with the [io.Reader.Read] method's behavior
-		// and return EOFError when we reach the end of the
-		// file, regardless of how much data we were able to
-		// read.
+	// Calculate the effective new offset
+	targetOffset := currentOffset + int64(len(into))
+	newOffset := lib.Min(targetOffset, fileSize)
+
+	// Read the data into the provided slice, and update
+	// the offset accordingly
+	n = copy(into, f.data[currentOffset:newOffset])
+	f.offset.Store(newOffset)
+
+	// If we've reached or surpassed the end, set the error to EOF
+	if targetOffset > fileSize {
 		err = newFsError(EOFError, "EOF")
 	}
-
-	n = copy(into, f.data[start:end])
-
-	f.offset.Store(offset + int64(n))
 
 	return n, err
 }
--- a/js/modules/k6/experimental/fs/file.go
+++ b/js/modules/k6/experimental/fs/file.go
@@ -16,6 +16,10 @@ type file struct {
 	data []byte
 
 	// offset holds the current offset in the file
+	//
+	// TODO: using an atomic here does not guarantee ordering of reads and seeks, and leaves
+	// the behavior not strictly defined. This is something we might want to address in the future, and
+	// is tracked as part of #3433.
 	offset atomic.Int64
 }
 
@@ -108,9 +112,7 @@ func (f *file) Seek(offset int64, whence SeekMode) (int64, error) {
 		return 0, newFsError(TypeError, "seeking beyond end of file")
 	}
 
-	// Note that the implementation assumes one `file` instance per file/vu.
-	// If that assumption was invalidated, we would need to atomically update
-	// the offset instead.
+	// Update the file instance's offset to the new selected position
 	f.offset.Store(newOffset)
 
 	return newOffset, nil
--- a/js/modules/k6/experimental/fs/file.go
+++ b/js/modules/k6/experimental/fs/file.go
@@ -58,7 +58,7 @@ func (f *file) Read(into []byte) (n int, err error) {
 
 	n = copy(into, f.data[start:end])
 
-	f.offset.Store(int64(n))
+	f.offset.Store(offset + int64(n))
 
 	return n, err
 }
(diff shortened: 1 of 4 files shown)


js/modules/k6/experimental/fs/file.go
```diff
@@ -38,14 +39,16 @@ type FileInfo struct {
...
+	if start == f.size() {
 		return 0, newFsError(EOFError, "EOF")
 	}
 
-	end := f.offset + len(into)
-	if end > len(f.data) {
-		end = len(f.data)
+	end := offset + int64(len(into))
+	if end > f.size() {
+		end = f.size()
```
codebien, line 51:
```suggestion
	if end < start + len(into) {
		end = start + int64(len(into))
```

js/modules/k6/experimental/fs/file.go
```diff
@@ -55,7 +58,7 @@ func (f *file) Read(into []byte) (n int, err error) {
 
 	n = copy(into, f.data[start:end])
 
-	f.offset += n
+	f.offset.Store(offset + int64(n))
```
codebien, line 61:
```suggestion
	f.offset.Store(end)
```
or do we expect to be different?

js/modules/k6/experimental/fs/file.go
```diff
@@ -13,7 +14,7 @@ type file struct {
 	data []byte
 
 	// offset holds the current offset in the file
-	offset int
+	offset atomic.Int64
```
codebien, line 17:
```suggestion
    // we should explain here that this is a partial solution.
	offset atomic.Int64
```
This is a partial solution and the logic could still have race condition if we run Read+Seek. We should explicit that this is delegated to the user of caring of it and not writing racy logic.
We also should add a suggestion that if this becomes an issue then synchronizing the offset movements with a channel to keep them in order could be a better solution.

codebien (changes requested):
LGTM, added a comment regarding f.size usage polish, and to make explicit the assumption we are making regarding the concurrent use of Read and Seek.

js/modules/k6/experimental/fs/file.go
mstoykov: Yeah, I do feel like the currently proposed solution is okay *but* from my experiments deno [^1] - it actually sequences the reads and seeks (and likely writes) so that if you have 

```javascript
let s1 = file.operation1(...)
let s2 = file.operation1(...)
let s3 = file.operation1(...)
await s3;
```
Actually both s1 and s2 also execute in that order instead of any other.

This likely has no practical benefit, but might be better to just have it this way from the very beginning. This likely can be made through having some kind of list of operations or just each operation referencing the last operation added and waiting for them :shrug: .

In the current implementation for example 
```javascript
let s1 = file.seek(10, 1)
let s2 = file.seek(10, 1)
await s2;
await file.read(...);
// there doesn't seem to be any gurantee whether the offset here will be 10 or 20
// but with deno it is *always* 20 in my experiments.
```

[^1]: This is purely on my experiments with it. No part of the documentation makes this claim or anything against it.

js/modules/k6/experimental/fs/file.go
```diff
@@ -38,14 +39,16 @@ type FileInfo struct {
 //
 // If the end of the file has been reached, it returns EOFError.
 func (f *file) Read(into []byte) (n int, err error) {
-	start := f.offset
-	if start == len(f.data) {
+	offset := f.offset.Load()
+
+	start := offset
+	if start == f.size() {
```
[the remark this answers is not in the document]
oleiade: I have revamped the function completely to make it more readable and cater to your comment 🙇🏻

js/modules/k6/experimental/fs/file.go
oleiade: I have revamped the function completely to make it more readable and cater to your comment 🙇🏻

js/modules/k6/experimental/fs/file.go
oleiade: I seem to have missed that Deno behavior specifically indeed. It does appear that in this scenario the operations are indeed kept in the input sequence.

Building on what @codebien mentioned in his previous comments, I had explored the idea of passing "operations/commands" messages around, as opposed to immediately executing operations, but as the benefits were not immediately clear, I left it at that. I'm gonna look into it again 👍🏻

js/modules/k6/experimental/fs/file.go
oleiade: I have revamped the function completely to make it more readable and cater to your comment 🙇🏻

js/modules/k6/experimental/fs/file.go
mstoykov: Again - I am not certain we should or must do the same as Deno - I am just pointing out that they seem to have a specific behaviour. While the currently implemented one not only deviates from theirs, which is not really a problem IMO, and also has arguably a harder to reason behaviour - which IMO is the bigger problem. 

Both of those IMO are less blocking and we can still have the more "racy" behaviour or something different from Deno that still is more "stable". I am not certain what that would be though :shrug: .

>  I had explored the idea of passing "operations/commands" messages around

I have not tried implementing this so no idea how well that will work or not, but it seems a bit complicated with moving parts. Arguably it is aligned with the "share by communicating" slogan of golang.

But to expand on my comment 
> or just each operation referencing the last operation added and waiting for them 🤷 .

Here I mean to have the ... file struct ... for example (:thinking:) have a "last currently not finished operation". All operations when created look at that field and replace it with "themselves". Then if the field had previously value "await" it to finish before they start to execute.

This seems like way easier to implement (from the current perspective of having not tried it). If you do all the manipulation of that field on the eventloop - also drops on the communication perspectice significantly.

I don't think we have a use case for knowing there are N operations or reordering them so having a single-linked-list of sort seems fine. 

But again - I have not implemented something like this in the context of k6 and goja + eventloop, so maybe something else will be nicer :shrug: 

This whole thing is slightly related to https://github.com/grafana/xk6-timers/issues/3 although maybe just in theory as that is about timers and waiting for them.

codebien: >I'm gonna look into it again 👍🏻

We resolved the main data race, and now this new issue is not a must-have. As @mstoykov reported, there isn't a spec defining it, so developers shouldn't have expectations on this behavior as it can differ between implementations.
However, it's still nice to have for us to align with the rest of the JavaScript ecosystem, so I'm for just opening an issue and setting a TODO comment in the code.

js/modules/k6/experimental/fs/module.go
```diff
@@ -209,15 +232,28 @@ func (f *File) Read(into goja.Value) *goja.Promise {
...
-	// Obtain the underlying byte slice from the ArrayBuffer.
-	// Note that this is not a copy, and will be modified by the Read operation
-	// in place.
-	buffer := ab.Bytes()
+	// Copy the ArrayBuffer into a byte slice, so that we can pass it to the
+	// [file.Read] method without risking to modify the original ArrayBuffer, and
+	// running into concurrency issues (data race).
+	intoBytes := ab.Bytes()
+	buffer := make([]byte, len(intoBytes))
+	_ = copy(buffer, intoBytes)
```
mstoykov, line 240: Is there a reason for this `copy` ?

It seems to me like there is no guarantee that `Read` will keep the bytes in the buffer it doesn't read. 

> Even if Read returns n < len(p), it may use all of p as scratch space during the call

From [Reader](https://pkg.go.dev/io#Reader)

And even if it was - you already copy on the way back.

So it seems like we here only want to make a buffer with the correct length

js/modules/k6/experimental/fs/module.go
```diff
@@ -186,8 +186,31 @@ func (f *File) Stat() *goja.Promise {
...
+			return nil
+		})
+	}
+
+	reject := func(reason any) {
+		callback(func() error {
+			rejectFunc(reason)
+			return nil
+		})
+	}
```
mstoykov, line 213: You can use the helper you create [here](https://github.com/grafana/k6/blob/61db2fea4b3bdaf73e0505aa7b6fd5c07d650117/js/promises/promises.go#L30C7-L30C7)

js/modules/k6/experimental/fs/file.go
oleiade: Thanks for your input @mstoykov, you actually started answering the main question I had about the implementation of your idea, so cheers for that 🙇🏻 

I will open a dedicated issue and a TODO in the code then 👍🏻

js/modules/k6/experimental/fs/file.go
```diff
@@ -97,18 +104,20 @@ func (f *file) Seek(offset int, whence SeekMode) (int, error) {
...
 	}
 
-	if newOffset > len(f.data) {
+	if newOffset > f.size() {
 		return 0, newFsError(TypeError, "seeking beyond end of file")
 	}
 
 	// Note that the implementation assumes one `file` instance per file/vu.
 	// If that assumption was invalidated, we would need to atomically update
 	// the offset instead.
```
codebien, line 113: Is this comment obsolete?

codebien (commented):
LGTM, mostly waiting for the TODO comment as expressed on https://github.com/grafana/k6/pull/3420#discussion_r1377256904

js/modules/k6/experimental/fs/file.go
```diff
@@ -71,9 +77,10 @@ var _ io.Reader = (*file)(nil)
 //
 // When using SeekModeStart, the offset must be positive.
 // Negative offsets are allowed when using `SeekModeCurrent` or `SeekModeEnd`.
-func (f *file) Seek(offset int, whence SeekMode) (int, error) {
-	newOffset := f.offset
+func (f *file) Seek(offset int64, whence SeekMode) (int64, error) {
+	startingOffset := f.offset.Load()
```
mstoykov, line 81: As mentioned in other places - this definitely still have a logical race even if the race detector will not complain as it is not a data one.

Here specifically two calls to `Seek` can race on reading the same value and then both changing that instead of being sequential.

We should probably fix this later along with the rest of the sequencing of the operations.

mstoykov (approved):
I have left some none blocking comments

js/modules/k6/experimental/fs/module.go
oleiade: The idea there was that we save the state of the input buffer, as given by the runtime, before starting the async operation of reading into it, so that we read (and return) into a buffer that's in a consistent state with the one we receive as input.

In light our internal discussions, I'm not sure that's the correct approach indeed, and this `copy` can probably go 👍🏻

js/modules/k6/experimental/fs/module.go
oleiade: I considered it as of the writing of the code, but I was not sure how it would be possible to do https://github.com/grafana/k6/pull/3420/files#diff-879feac1c36a4a6578f2509b47bf5dab13a5a93d89a05e9dcf4c71613758d19dR254 then.

Namely, the helper doesn't give me access to the underlying goja resolving function. Also, would it be okay to declare a "side" callback manually, side-by-side with the one created under the hood for the `resolve` and `reject` functions returned by `promises.New`? 😃 ❓

js/modules/k6/experimental/fs/file.go
oleiade: Good catch 👍🏻

oleiade:
Created issue #3433 to track the issue related to operations not being treated sequentially (cc @codebien ) 👍🏻

js/modules/k6/experimental/fs/file.go
oleiade: See issue #3433

codebien (approved):

js/modules/k6/experimental/fs/module.go
mstoykov: You are correct, and having now gone through the code I saw other problems. 

But let's merge this now and I will point them out and we will fix them in the PRs that aren't about the races.

mstoykov (approved):

js/modules/k6/experimental/fs/module.go
oleiade: Sounds good 🙇🏻

PR #3423, opened 2023-10-26
Adds the K6_INFLUXDB_PROXY environment variable support, closes #3418

### LINKED ISSUE #3418: Separate proxy configuration for outputs
### Feature Description

Hello,
In my organization InfluxDB cluster is not available for direct connections from certain locations, but it is possible using a proxy.
I tried to use a command like this:
```
HTTP_PROXY=proxy:3128 HTTPS_PROXY=proxy:3128 k6 run /tests/generic_endpoint.js --out influxdb=$K6_INFLUXDB_URL:$K6_INFLUXDB_PORT/$K6_INFLUXDB_DATABASE
```
but it looks like k6 uses those vars only for the testing requests. I also tried to export those vars, but got the same results. But even if it would work like that, it is still not a good solution since I don't want to send my requests through the same proxy.

Add the K6_INFLUXDB_PROXY env var

_No response_

IvanovOleg:
## What?

This PR closes #3418 by adding the K6_INFLUXDB_PROXY environment variable support.

## Why?

This PR fixes the problem when your InfluxDB server is not opened for direct connections and requires a connection through proxy.

## Checklist

## Related PR(s)/Issue(s)

Closes #3418

### DIFF
--- a/output/influxdb/config.go
+++ b/output/influxdb/config.go
@@ -17,6 +17,7 @@ import (
 type Config struct {
 	// Connection.
 	Addr             null.String        `json:"addr" envconfig:"K6_INFLUXDB_ADDR"`
+	Proxy            null.String        `json:"proxy,omitempty" envconfig:"K6_INFLUXDB_PROXY"`
 	Username         null.String        `json:"username,omitempty" envconfig:"K6_INFLUXDB_USERNAME"`
 	Password         null.String        `json:"password,omitempty" envconfig:"K6_INFLUXDB_PASSWORD"`
 	Insecure         null.Bool          `json:"insecure,omitempty" envconfig:"K6_INFLUXDB_INSECURE"`
@@ -56,6 +57,9 @@ func NewConfig() Config {
 func (c Config) Apply(cfg Config) Config {
 	if cfg.Addr.Valid {
 		c.Addr = cfg.Addr
+	}
+	if cfg.Proxy.Valid {
+		c.Proxy = cfg.Proxy
 	}
 	if cfg.Username.Valid {
 		c.Username = cfg.Username
--- a/output/influxdb/config_test.go
+++ b/output/influxdb/config_test.go
@@ -51,3 +51,26 @@ func TestParseURL(t *testing.T) {
 		})
 	}
 }
+
+func TestGetConsolidatedConfigHTTPProxy(t *testing.T) {
+	t.Parallel()
+	t.Run("Valid Proxy URL", func(t *testing.T) {
+		t.Parallel()
+		testdata := map[string]string{
+			"K6_INFLUXDB_PROXY": "http://localhost:3128",
+		}
+		config, err := GetConsolidatedConfig(nil, testdata, "")
+		assert.NoError(t, err)
+		assert.Equal(t, "http://localhost:3128", config.Proxy.String)
+	})
+	t.Run("Invalid Proxy URL", func(t *testing.T) {
+		t.Parallel()
+		testdata := map[string]string{
+			"K6_INFLUXDB_PROXY": "http://foo\x7f.com/",
+		}
+		config, err := GetConsolidatedConfig(nil, testdata, "")
+		assert.NoError(t, err)
+		_, err = MakeClient(config)
+		assert.Error(t, err)
+	})
+}
--- a/output/influxdb/util.go
+++ b/output/influxdb/util.go
@@ -2,6 +2,8 @@ package influxdb
 
 import (
 	"fmt"
+	"net/http"
+	"net/url"
 	"strings"
 
 	client "github.com/influxdata/influxdb1-client/v2"
@@ -18,13 +20,21 @@ func MakeClient(conf Config) (client.Client, error) {
 	if conf.Addr.String == "" {
 		conf.Addr = null.StringFrom("http://localhost:8086")
 	}
-	return client.NewHTTPClient(client.HTTPConfig{
+	clientHTTPConfig := client.HTTPConfig{
 		Addr:               conf.Addr.String,
 		Username:           conf.Username.String,
 		Password:           conf.Password.String,
 		UserAgent:          "k6",
 		InsecureSkipVerify: conf.Insecure.Bool,
-	})
+	}
+	if conf.Proxy.Valid {
+		parsedProxyURL, err := url.Parse(conf.Proxy.String)
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse the http proxy URL: %w", err)
+		}
+		clientHTTPConfig.Proxy = http.ProxyURL(parsedProxyURL)
+	}
+	return client.NewHTTPClient(clientHTTPConfig)
 }
 
 func MakeBatchConfig(conf Config) client.BatchPointsConfig {


output/influxdb/util.go
```diff
@@ -18,12 +20,21 @@ func MakeClient(conf Config) (client.Client, error) {
...
+		}
+		ProxyURL = parsedProxyURL
+	}
 	return client.NewHTTPClient(client.HTTPConfig{
 		Addr:               conf.Addr.String,
 		Username:           conf.Username.String,
 		Password:           conf.Password.String,
 		UserAgent:          "k6",
 		InsecureSkipVerify: conf.Insecure.Bool,
+		Proxy:              http.ProxyURL(ProxyURL),
```
mstoykov, line 37: This is seems strange -  I don't know if `http.ProxyURL` works with an empty URL or if the client handles it - but it seems strange ot set it if it is *not* valid.

I would prefer that the `client.HTTPConfig` is made and tehn the proxyURL is set only if ti is configured and then use that config at the end again to make the new client.

mstoykov (changes requested):
LGTM in general, thanks :bow:

But can you change the way it is set currently, and I would prefer if we have at least one config test.  And if possible actually test the proxying?

I also saw that you didn't change it to be parsable from the cli directly (see `ParseURL`  in the config.go). 

I am not certain how good of an idea is to support this ... given that I think it will need to be escaped quite a lot.  So I kind of don't think you need to do it , but let's see what @codebien thinks as well

output/influxdb/util.go
IvanovOleg: I tested this code with K6_INFLUXDB_PROXY set and unset, worked both ways without problems.

IvanovOleg: I am going to check that parsing

IvanovOleg: @mstoykov saparatetd client.HTTPConfig

IvanovOleg: I am not sure how to parse the Proxy field since it goes to the transport, not to the URL.

codebien: I share the same idea as @mstoykov, I would prefer not to rely on InfluxDB-client internal logic. I prefer if we stay conservative here. In this way, we make clear to the next contributor what the logic is.

codebien:
>I also saw that you didn't change it to be parsable from the cli directly (see ParseURL in the config.go).

@mstoykov Yeah, I would prefer if we avoid it. An URL in another URL doesn't sound like the best UX.

output/influxdb/util.go
```diff
@@ -18,13 +20,21 @@ func MakeClient(conf Config) (client.Client, error) {
...
 		Username:           conf.Username.String,
 		Password:           conf.Password.String,
 		UserAgent:          "k6",
 		InsecureSkipVerify: conf.Insecure.Bool,
-	})
+	}
+	if conf.Proxy.Valid && conf.Proxy.String != "" {
+		parsedProxyURL, err := url.Parse(conf.Proxy.String)
+		if err != nil {
+			return nil, err
```
codebien, line 33:
```suggestion
			return nil, fmt.Errorf("failed to parse the http proxy URL: %w", err)
```
I expect we need to wrap this error. The user may have issues in debugging where the error comes from.

output/influxdb/config.go
```diff
@@ -17,6 +17,7 @@ import (
 type Config struct {
 	// Connection.
 	Addr             null.String        `json:"addr" envconfig:"K6_INFLUXDB_ADDR"`
+	Proxy            null.String        `json:"proxy,omitempty" envconfig:"K6_INFLUXDB_PROXY"`
```
codebien, line 20:
```suggestion
	Proxy            null.String        `json:"proxy,omitempty" envconfig:"K6_INFLUXDB_HTTP_PROXY"`
```
@mstoykov WDYT? Should we do it? I like the consistency with the original HTTP_PROXY variable, but on the other side, I'm not sure because it could introduce confusion around the expected HTTP/S value.

codebien (changes requested):
As @mstoykov reported, we should add a test for it.

output/influxdb/util.go
mstoykov: Maybe also some indicatio nit is `influxdb` output as otherwise it will still be hard to guess where it comes from

output/influxdb/config.go
IvanovOleg: it's because when I set HTTP_PROXY and HTTPS_PROXY vars, k6 sends all request through proxy, not just output results.

output/influxdb/util.go
IvanovOleg: I will add this

output/influxdb/config.go
IvanovOleg: Added a config test for the K6_INFLUXDB_PROXY variable. I will add a test for the proxy connectivity as well.

output/influxdb/util.go
```diff
@@ -18,13 +20,21 @@ func MakeClient(conf Config) (client.Client, error) {
...
-	return client.NewHTTPClient(client.HTTPConfig{
+	clientHTTPConfig := client.HTTPConfig{
 		Addr:               conf.Addr.String,
 		Username:           conf.Username.String,
 		Password:           conf.Password.String,
 		UserAgent:          "k6",
 		InsecureSkipVerify: conf.Insecure.Bool,
-	})
+	}
+	if conf.Proxy.Valid && conf.Proxy.String != "" {
```
mstoykov, line 30:
```suggestion
	if conf.Proxy.Valid {
```

`Valid` Should be enough - did you have problems with only it?

IvanovOleg: removed

output/influxdb/config_test.go
```diff
@@ -51,3 +52,13 @@ func TestParseURL(t *testing.T) {
...
 }
+
+func TestGetConsolidatedConfig(t *testing.T) {
+	t.Parallel()
+	testdata := map[string]string{
+		"K6_INFLUXDB_PROXY": "http://localhost:3128",
+	}
+	config, err := GetConsolidatedConfig(nil, testdata, "")
+	require.NoError(t, err)
+	require.Equal(t, config.Proxy, null.StringFrom("http://localhost:3128"))
```
codebien, line 63:
```suggestion
	assert.Equal(t, "http://localhost:3128", config.Proxy.String)
```
The order is wrong here, the expected value is the first. Also I think it's better to assert over a simpler type like `string` instead of the more complex `null.String`.

output/influxdb/config_test.go
```diff
@@ -51,3 +52,13 @@ func TestParseURL(t *testing.T) {
 		})
 	}
 }
+
+func TestGetConsolidatedConfig(t *testing.T) {
```
codebien, line 56:
```suggestion
func TestGetConsolidatedConfigHTTPProxy(t *testing.T) {
```
The current name is too generic for the current implementation which is very limited to the HTTP proxy.

output/influxdb/config_test.go
```diff
@@ -51,3 +52,13 @@ func TestParseURL(t *testing.T) {
...
+	t.Parallel()
+	testdata := map[string]string{
+		"K6_INFLUXDB_PROXY": "http://localhost:3128",
+	}
+	config, err := GetConsolidatedConfig(nil, testdata, "")
+	require.NoError(t, err)
```
codebien, line 62:
```suggestion
	assert.Error(t, err)
```
We should also add a test case for the opposite case, checking that an error is returned as expected if an unparsable URL is passed.

codebien (changes requested):
@mstoykov I checked the returned UX in case of error and it seems fine to me.

```js
ERRO[0000] could not create the 'influxdb' output: failed to parse the http proxy URL: parse "http://foo\x7f.com/": net/url: invalid control character in URL
```

codebien (approved):

mstoykov (approved):
Thanks for this change @IvanovOleg

PR #3429, opened 2023-10-30
fix TestSetupTimeout by increasing the time range it should finish in

mstoykov:


codebien (approved):

olegbespalov (approved):
TBH sounds really strange to have this big difference even if the github runner isn't so performant. But :+1:

oleiade (approved):

PR #3439, opened 2023-11-03
Update xk6-redis dependency to v0.2.0

oleiade:
## What?

This Pull Request updates the [xk6-redis](https://github.com/grafana/xk6-redis) dependency backing the `k6/experimental/redis` module to version `v0.2.0`.

## Why?

The latest version of [xk6-redis](https://github.com/grafana/xk6-redis) addresses the [support for (m)TLS](https://github.com/grafana/xk6-redis/pull/17), and updates the `redis.Client` constructor options to reflect that change. This is a module API **breaking change**, that should be documented accordingly 🤝 

## Related PR(s)/Issue(s)

ref https://github.com/grafana/xk6-redis/pull/17
ref https://github.com/grafana/xk6-redis/issues/14


go.mod
```diff
@@ -33,7 +33,7 @@ require (
 	github.com/nu7hatch/gouuid v0.0.0-20131221200532-179d4d0c4d8d
 	github.com/serenize/snaker v0.0.0-20201027110005-a7ad2135616e
 	github.com/sirupsen/logrus v1.9.3
-	github.com/spf13/afero v1.1.2
+	github.com/spf13/afero v1.9.5
```
mstoykov, line 36: This and null should not be updated

mstoykov (approved):

go.mod
olegbespalov: @oleiade, you could include the https://github.com/grafana/k6/blob/master/modtools_frozen.yml to the xk6-redis (and any other extension) to prevent such cases in the future

olegbespalov (approved):

go.mod
mstoykov: This will only work if you also use [modtools](https://github.com/dop251/modtools) which arguably most people do not. So arguably just not doing `go get` with `-u` is the more important part :P.

Or us prioritizing dropping `afero` :shrug:. 

Actually having have written this - xk6-redis should likely just not use `afero` but [fsext](https://github.com/grafana/k6/tree/master/lib/fsext)

olegbespalov: I've been using `modtools` since I learned that it existed and we have some dependencies bumped. I think we maybe need to advertise to folks about using it, via Slack/documentation since it's not only about removing the afero but more about having the proper process.

Regarding the `fsext,` also I like the idea; I am also going to check if I can switch to it in the projects that I've actively maintaine :+1:

PR #3440, opened 2023-11-03
Add built-in root certificate fallbacks

mem:
Blank-importing golang.org/x/crypto/x509roots/fallback bundles a set of root fallback certificates from Mozilla into the resulting binary. This allows the program to run in environments where the system root certificates are not available, for example inside a minimal container. These are _fallbacks_, meaning that if the system _does have_ a set of root certificates, those will be given priority. The binary size will increase a little (~ 220 kB).

It's added to main.go instead of somewhere else because the recommendation is for the package to be imported from binaries, not from libraries. Calling x509.SetFallbackRoots (what the imported package does in its init function) more than once will cause the program to panic. In principle, the Go import system will prevent the package from being imported twice, so that shouldn't be a problem, but it's probably better to keep this very visible, therefore main.go.

## Checklist


vendor/modules.txt
```diff
@@ -312,6 +312,9 @@ go.uber.org/goleak/internal/stack
 golang.org/x/crypto/md4
 golang.org/x/crypto/ocsp
 golang.org/x/crypto/ripemd160
+# golang.org/x/crypto/x509roots/fallback v0.0.0-20231030152948-74c2ba9521f1
+## explicit; go 1.20
+golang.org/x/crypto/x509roots/fallback
```
mem, line 317: This is because the SetFallbackRoots was added in 1.20.

I might be able to work around this if building with < 1.20 is required.

main.go
```diff
@@ -2,6 +2,15 @@ package main
...
 	"go.k6.io/k6/cmd"
+
+	// Blank-importing golang.org/x/crypto/x509roots/fallback bundles a set of
+	// root fallback certificates from Mozilla into the resulting binary. This
+	// allows the program to run in environments where the system root
+	// certificates are not available, for example inside a minimal container.
+	// These are _fallbacks_, meaning that if the system _does have_ a set of
+	// root certificates, those will be given priority. The binary size will
+	// increase a little (~ 220 kB).
+	_ "golang.org/x/crypto/x509roots/fallback"
```
olegbespalov, line 13: I might be wrong, but there could be an issue of having an import in `main.go` any custom binary built with the https://github.com/grafana/xk6 won't have it :cry: 

```sh
$ xk6 build 12a0441084ce27d7235b01fd4aab3a0c7d0987aa
2023/11/07 09:34:47 [INFO] Temporary folder: /tmp/buildenv_2023-11-07-0934.3891399373
2023/11/07 09:34:47 [INFO] Initializing Go module
2023/11/07 09:34:47 [INFO] exec (timeout=10s): /home/olegbespalov/go/bin/go mod init k6
go: creating new go.mod: module k6
2023/11/07 09:34:47 [INFO] Pinning versions
2023/11/07 09:34:47 [INFO] exec (timeout=0s): /home/olegbespalov/go/bin/go mod tidy -compat=1.17
go: warning: "all" matched no packages
2023/11/07 09:34:47 [INFO] Writing main module: /tmp/buildenv_2023-11-07-0934.3891399373/main.go
2023/11/07 09:34:47 [INFO] exec (timeout=0s): /home/olegbespalov/go/bin/go mod edit -require go.k6.io/k6@12a0441084ce27d7235b01fd4aab3a0c7d0987aa
2023/11/07 09:34:47 [INFO] exec (timeout=0s): /home/olegbespalov/go/bin/go mod tidy -compat=1.17
2023/11/07 09:34:47 [INFO] exec (timeout=0s): /home/olegbespalov/go/bin/go mod tidy -compat=1.17
2023/11/07 09:34:47 [INFO] Build environment ready
2023/11/07 09:34:47 [INFO] Building k6
2023/11/07 09:34:47 [INFO] exec (timeout=0s): /home/olegbespalov/go/bin/go mod tidy -compat=1.17
2023/11/07 09:34:47 [INFO] exec (timeout=0s): /home/olegbespalov/go/bin/go build -o /home/olegbespalov/k6 -ldflags=-w -s -trimpath
2023/11/07 09:34:48 [INFO] Build complete: ./k6
2023/11/07 09:34:48 [INFO] Cleaning up temporary folder: /tmp/buildenv_2023-11-07-0934.3891399373

xk6 has now produced a new k6 binary which may be different than the command on your system path!
Be sure to run './k6 run <SCRIPT_NAME>' from the '/home/olegbespalov' directory.
$ go version -m k6 | grep crypto
        dep     github.com/grafana/xk6-webcrypto        v0.1.0  h1:StrQZkUi4vo3bAMmBUHvIQ8P+zNKCH3AwN22TZdDwHs=
        dep     golang.org/x/crypto     v0.14.0 h1:wBqGXzWJW6m1XrIKlAH0Hs1JJ7+9KBwnIO8v66Q9cHc=
```

vendor/modules.txt
olegbespalov: There is no need for a workaround :+1: So far we do support the two last golang's versions, which aligns the go team

main.go
```diff
@@ -2,6 +2,15 @@ package main
...
 import (
 	"go.k6.io/k6/cmd"
+
+	// Blank-importing golang.org/x/crypto/x509roots/fallback bundles a set of
+	// root fallback certificates from Mozilla into the resulting binary. This
+	// allows the program to run in environments where the system root
+	// certificates are not available, for example inside a minimal container.
+	// These are _fallbacks_, meaning that if the system _does have_ a set of
+	// root certificates, those will be given priority. The binary size will
+	// increase a little (~ 220 kB).
```
olegbespalov, line 12: Would you mind removing the size details? We have no guarantee that this will be always the case and this isn't really important here

```suggestion
	// root certificates, those will be given priority.
```

codebien (commented):
Hey @mem, thanks for your contribution! :bow:

What is the expected logic if I copy a CA in one of the file system`s directories? https://go.dev/src/crypto/x509/root_linux.go What is the first, the CA in the folder or the fallback?

> for example inside a minimal container.

It sounds like an infrastructure problem to me, and the issue should be resolved by copying or installing the certificate there. Can you clarify why in your use case it doesn't work? :pray:

main.go
mem: I moved the import to "cmd"

mem:
> What is the expected logic if I copy a CA in one of the file system`s directories? https://go.dev/src/crypto/x509/root_linux.go What is the first, the CA in the folder or the fallback?

It works as before: this is just a fallback meant to be used when the system does not provide its own certificate store.

> > for example inside a minimal container.
> 
> It sounds like an infrastructure problem to me, and the issue should be resolved by copying or installing the certificate there. Can you clarify why in your use case it doesn't work? 🙏

While I might agree that you should set up your containers correctly, we do not control what other people do. Even if you create a container for k6 that contains a certificate store, it's perfectly possible to reuse that container (e.g. Docker's COPY) and _not_ have a certificate store there.

The fallback package is continuously and automatically updated from Mozilla's store. This means the certificates are handled are any other dependency (e.g. you get dependabot notifications / updates). That means there's a security advantage to using this package vs trying to keep containers up-to-date. If you look at most containers in common use, they have out-of-date certificate stores.

olegbespalov (dismissed):
LGTM :+1: Thanks for the contribution :bow:

codebien:
>This means the certificates are handled are any other dependency (e.g. you get dependabot notifications / updates).

Good point, I like it.

@mem Adding the fallbacks does mean we can remove the `ca-certificates` from the Dockerfile, right? https://github.com/grafana/k6/blob/5e83d389fde71d3d7bf96fbe98d9a2938e903b82/Dockerfile#L12

Now, I'm curious why no one of the other Grafana's main projects is not doing this, and it probably makes even more sense for projects like Grafana or Mimir. @mem Do you know why? Or are you going to open the same PRs there? I would like to have a shared vision of this across the Grafana org.

mem:
> @mem Adding the fallbacks does mean we can remove the `ca-certificates` from the Dockerfile, right?
> 
> https://github.com/grafana/k6/blob/5e83d389fde71d3d7bf96fbe98d9a2938e903b82/Dockerfile#L12

That is correct. I didn't want to go overboard and remove that straight away. I have done that now. 

> Now, I'm curious why the other Grafana's main projects are not doing this, and it probably makes even more sense for projects like Grafana or Mimir. @mem Do you know why? Or are you going to open the same PRs there? I would like to have a shared vision of this across the Grafana org.

This is relatively new ("this year" new, not "yesterday" new), and it hasn't received a lot of "press".

mem:
```
--- FAIL: TestClient_TlsParameters (0.00s)
    --- FAIL: TestClient_TlsParameters/ConnectTlsClientCertWithPasswordNoClientAuth (60.01s)
        teststate_test.go:155: 
            	Error Trace:	/home/runner/work/k6/k6/js/modules/k6/grpc/teststate_test.go:155
            	            				/home/runner/work/k6/k6/js/modules/k6/grpc/client_test.go:1053
            	Error:      	"GoError: context deadline exceeded: write tcp 127.0.0.1:51926->127.0.0.1:46399: write: connection reset by peer at reflect.methodValueCall (native)" does not contain "remote error: tls: bad certificate"
            	Test:       	TestClient_TlsParameters/ConnectTlsClientCertWithPasswordNoClientAuth
```

Any idea how the clientAuthCA contents was generated in that file?

I don't understand why this specific test is failing if it's supposed to be using an empty cert pool. That shouldn't have changed.

Dockerfile
```diff
@@ -9,8 +9,7 @@ RUN CGO_ENABLED=0 GOOS=$TARGETOS GOARCH=$TARGETARCH go build -trimpath -o /usr/b
 FROM alpine:3.18 as release
 
 # hadolint ignore=DL3018
-RUN apk add --no-cache ca-certificates && \
-    adduser -D -u 12345 -g 12345 k6
+RUN adduser -D -u 12345 -g 12345 k6
```
olegbespalov, line 12: I'm not sure if it's required to remove the `ca-certificates` :man_shrugging:  From my point of view, `Dockerfile` is a different "layer" and could stay as it's untouched. There is no cost of maintaining it.

However, if we decide to remove it, let's also clear the comment `hadolint ignore=DL3018` since it relates to that package.

olegbespalov:
>I don't understand why this specific test is failing if it's supposed to be using an empty cert pool. That shouldn't have changed.

Unfortunately, you hit our flaky test. The issue itself somehow goes down to the grpc library https://github.com/grpc/grpc-go/issues/6593. I was starting to work on that, but unfortunately, I didn't succeed since it requires more changes in the grpc-go library. It is still somehow on my list, but in a low-priority.

> Any idea how the clientAuthCA contents was generated in that file?

This certificate is supposed to work only for email signing. We have a low-priority task to "script" the process to make it more transparent and redundant https://github.com/grafana/xk6-grpc/issues/52.

olegbespalov (dismissed):

Dockerfile
codebien: >Dockerfile is a different "layer"

It is but they have the same end goal.

>There is no cost of maintaining it.

From my point of view, it has a cost in case we need to troubleshoot an issue. It could be difficult to be sure what was the used certificate from the system.

@mem I share @olegbespalov's request. Can you drop the `hadolint` line, please?

codebien (changes requested):
LGTM, mostly waiting for @olegbespalov's suggestion to be addressed.

olegbespalov (approved):

codebien (approved):

Issue #2352, opened 2022-01-25
Revisit `k6 version` command and related configuration


yorugac:
### Feature Description

Currently version of k6 can be looked up only with `k6 version` while `k6 -v` switches on verbose logging. Many people are more used to `--version` and `-v` flags instead. OTOH, `k6 version` is very likely required for backwards compatibility.

### Suggested Solution (optional)

Add support for `--version` flag.

### Already existing or connected issues / PRs (optional)

Original source: discussion in https://github.com/grafana/k6/pull/2334#discussion_r789781094

na--:
Yeah, `k6 version` can't be removed at this point, but I don't see a problem with also having `--version` for users that are used to other tools. I think that convention is mostly adopted by CLI tools that don't have sub-commands, and `k6 version` feels more appropriate for us and should still be the default though.

MStoykov:
I am mostly against adding more flags and depending on cobra functionality. Yes it's unlikely we will ever ditch cobra, and even if we do this likely will be the least of our problems. But also I don't remember anyone ever complaining about the lack of `--version` so :shrug:

codebien:
>k6 version feels more appropriate for us and should still be the default though.

Based on the quote, same for me, one more flag without a concrete end goal doesn't sound a good thing to me

MattDodsonEnglish:
I'm chiming in to say that this gets me every time:

```
 $ k6 --version
ERRO[0000] unknown flag: --version 
```

It's just so against my normal expectations. I would bet a significant amount of money that I'm not the only one who has seen this error many times.[^1]
[^1]: That error message could be better, but that's maybe a different issue.

> But also I don't remember anyone ever complaining about the lack of --version so shrug

I think the annoyance is minor enough that most people won't complain. A person would have to be really nitpicky to complain about it. Anyway, I just complained about it internally and was directed to this issue. 

> Based on the quote, same for me, one more flag without a concrete end goal doesn't sound a good thing

Generally I am for exposing only one interface to do one thing. But conventions are strong here, and this is about as simple as a command can be, so I don't expect multiple ways would cause much confusion here. There's already `k6 help` and `k6 help`.

ffapitalle:
Looking for a good first issue to start contributing I've done this [PR](#3442). Just my 2 cents.

PR #3442, opened 2023-11-06
Add support for --version flag

ffapitalle:
## What?

Adds support for `--version` global flag that matches the `version` subcommand.

## Why?

Running the usual Unix-like flags returns an error.

## Checklist

## Related PR(s)/Issue(s)

<https://github.com/grafana/k6/issues/2352>

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -48,6 +48,7 @@ func newRootCommand(gs *state.GlobalState) *rootCommand {
 		SilenceUsage:      true,
 		SilenceErrors:     true,
 		PersistentPreRunE: c.persistentPreRunE,
+		Version:           versionString(),
 	}
 
 	rootCmd.PersistentFlags().AddFlagSet(rootCmdPersistentFlagSet(gs))
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -52,7 +52,7 @@ func newRootCommand(gs *state.GlobalState) *rootCommand {
 	}
 
 	rootCmd.SetVersionTemplate(
-		`{{with .Name}}{{printf "%s " .}}{{end}}{{printf "v%s" .Version}}\n`,
+		`{{with .Name}}{{printf "%s " .}}{{end}}{{printf "v%s\n" .Version}}`,
 	)
 	rootCmd.PersistentFlags().AddFlagSet(rootCmdPersistentFlagSet(gs))
 	rootCmd.SetArgs(gs.CmdArgs[1:])
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -51,6 +51,9 @@ func newRootCommand(gs *state.GlobalState) *rootCommand {
 		Version:           versionString(),
 	}
 
+	rootCmd.SetVersionTemplate(
+		`{{with .Name}}{{printf "%s " .}}{{end}}{{printf "v%s" .Version}}\n`,
+	)
 	rootCmd.PersistentFlags().AddFlagSet(rootCmdPersistentFlagSet(gs))
 	rootCmd.SetArgs(gs.CmdArgs[1:])
 	rootCmd.SetOut(gs.Stdout)
--- a/cmd/tests/cmd_run_test.go
+++ b/cmd/tests/cmd_run_test.go
@@ -42,7 +42,7 @@ func TestVersion(t *testing.T) {
 	cmd.ExecuteWithGlobalState(ts.GlobalState)
 
 	stdout := ts.Stdout.String()
-	assert.Contains(t, stdout, "k6 version "+consts.Version)
+	assert.Contains(t, stdout, "k6 v"+consts.Version)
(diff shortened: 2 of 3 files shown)


cmd/tests/cmd_run_test.go
```diff
@@ -42,7 +42,7 @@ func TestVersion(t *testing.T) {
 	cmd.ExecuteWithGlobalState(ts.GlobalState)
 
 	stdout := ts.Stdout.String()
-	assert.Contains(t, stdout, "k6 v"+consts.Version)
+	assert.Contains(t, stdout, "k6 version "+consts.Version)
```
codebien, line 45: It prints
```
k6 version 0.47.0 (go1.21.1, linux/amd64)
```

That means it would be a breaking change for someone relying on it on automated tools. And, it doesn't match really the version we usually announce, check https://github.com/grafana/k6/releases/tag/v0.47.0 for an example.

Can you roll back it, please? I expect you're doing it because it is the default template for Cobra but if you use [SetVersionTemplate](https://pkg.go.dev/github.com/spf13/cobra#Command.SetVersionTemplate) then you should be able to change it.

codebien (changes requested):
Hey @ffapitalle,
thanks for your contribution! :bow:

cmd/tests/cmd_run_test.go
ffapitalle: Hi @codebien !

Thank you for your feedback. I did change that and the corresponding test to match Cobra's default. 
I'll rollback that and change the default.

cmd/root.go
```diff
@@ -48,8 +48,12 @@ func newRootCommand(gs *state.GlobalState) *rootCommand {
 		SilenceUsage:      true,
 		SilenceErrors:     true,
 		PersistentPreRunE: c.persistentPreRunE,
+		Version:           versionString(),
 	}
 
+	rootCmd.SetVersionTemplate(
+		`{{with .Name}}{{printf "%s " .}}{{end}}{{printf "v%s" .Version}}\n`,
```
codebien, line 55:
```suggestion
		`{{with .Name}}{{printf "%s " .}}{{end}}{{printf "v%s" .Version}}`,
```
Is it required? On Alacritty I get the following output:
```
k6 v0.47.0 (go1.21.1, linux/amd64)\n% 
```

In any case, even after removing the newline the `%` char still remains. It requires a bit of investigation about why.

codebien (changes requested):
@ffapitalle Thanks for the changes.

I think we're now missing two more things:

Must to have: add to the related test here https://github.com/grafana/k6/blob/e44a08fa5d642ba148c08d4a1eb291de576009c5/cmd/tests/cmd_run_test.go#L37C1-L52C2 the case with the new flag. It requires a bit of refactoring to have a table test.

Nice to have: having the `k6 version` command calling the root's command or vice versa. In this way, we are 100% sure that one change will be also applied to the other.

cmd/root.go
ffapitalle: My fault on this. It was the wrong place to include the newline character.

ffapitalle:
> Must to have: add to the related test here https://github.com/grafana/k6/blob/e44a08fa5d642ba148c08d4a1eb291de576009c5/cmd/tests/cmd_run_test.go#L37C1-L52C2 the case with the new flag. It requires a bit of refactoring to have a table test. 

I went for the table test schema. 

> Nice to have: having the `k6 version` command calling the root's command or vice versa. In this way, we are 100% sure that one change will be also applied to the other.

I gave it a few shots but with no luck. Both (flag and subcommand) rely on the `versionString` [function](https://github.com/ffapitalle/k6/blob/ISSUE-2352/add-support--version-flag/cmd/version.go#L13), so the actual version comes from this function.

cmd/version.go
```diff
@@ -10,23 +10,28 @@ import (
...
-
-			if exts := ext.GetAll(); len(exts) > 0 {
-				extsDesc := make([]string, 0, len(exts))
-				for _, e := range exts {
-					extsDesc = append(extsDesc, fmt.Sprintf("  %s", e.String()))
-				}
-				printToStdout(gs, fmt.Sprintf("Extensions:\n%s\n",
-					strings.Join(extsDesc, "\n")))
-			}
+			printToStdout(gs, fmt.Sprintf("k6 v%s\n", versionString()))
```
codebien, line 34:
```suggestion
	root := cmd.Root()
    root.SetArgs([]string{"--version"})
	root.Execute()
```
This should do the trick for running a `k6 version` as `k6 --version`

codebien (changes requested):

codebien (changes requested):
You have to apply `go fmt` otherwise the linter will complain about indentation.

olegbespalov (approved):
Thanks for the contribution! :bow:

codebien (approved):

PR #3445, opened 2023-11-08
Add tracing instrumentation

ka3de:
## What?

This PR adds a base tracing instrumentation around [opentelemetry-go](https://github.com/open-telemetry/opentelemetry-go) library. Specifically initializes a `TracerProvider` based on a new traces ouput configuration option (`--traces-output` / `K6_TRACES_OUTPUT`). The `TracerProvider` is exposed _partially_ to modules through the _VU.State()_  so these can request _[Tracers](https://pkg.go.dev/go.opentelemetry.io/otel/trace@v1.19.0#Tracer)_ initialization that can be used to generate traces.

## Why?

There has been various requests to add more tracing functionality to k6, both from k6 users and from extension developers. This PR represents a first step towards promoting tracing as a first class citizen in k6.

## Checklist


js/common/trace.go
```diff
@@ -0,0 +1,8 @@
+package common
+
+import "go.opentelemetry.io/otel/trace"
+
+// TracerProvider provides methods for OTEL tracers initialization.
+type TracerProvider interface {
```
ka3de, line 6: I was doubtful whether to define this in _[js/modules/modules.go](https://github.com/grafana/k6/blob/9d1d028392094a8de6897f752784e3dd22a8371d/js/modules/modules.go#L39)_ along with the _VU_ interface, but decided to define it in `common` package mimicking the `Events` implementation and also thinking that more tracing related functionalities could be added here in the future. Feedback is welcomed.

js/modules/modules.go
```diff
@@ -50,6 +50,11 @@ type VU interface {
 	// interface.
 	Events() common.Events
 
+	// TracerProvider returns the VU TracerProvider.
+	// NOTE: This API is EXPERIMENTAL and may be changed, renamed or
+	// completely removed in a later k6 release.
+	TracerProvider() common.TracerProvider
```
ka3de, line 56: Initially I added the `TracerProvider` to `VU.State` as suggested by @mstoykov , but the problem that I was facing with this, at least from the _xk6-browser_ side, is that `State` is `nil` when instantiating a new module instance (actual call to `NewModuleInstance`), where we expect to have access to the tracer provider. Considering also that `State` seems to represent the `VU`'s volatile state, and that the tracer provider is not tight to a `VU`, but is a "more global entity" I decided to expose it directly through `VU` interface. Let me know what do you think.

mstoykov:
a quick feedback: the first two commits should be one commit. Not vendoring the dependencies breaks the build and no commit should break the build if possible.

js/modules/modules.go
mstoykov: From my understanding just like with metrics you will only emit traces within the actual execution of the test code.

As such I will prefer if keep all the tracing similar to metrics within the bounds of the execution of the test.

mstoykov:
I would also recommend to pull the `parseutils` change in a separate PR so we can review and merge it faster and without all the noise of this one.

You can rebase this one after that one is merged - which hopefully will be really fast.

ka3de:
> a quick feedback: the first two commits should be one commit. Not vendoring the dependencies breaks the build and no commit should break the build if possible.

> I would also recommend to pull the parseutils change in a separate PR

I will address these two points. Treat this PR as a draft currently.

js/common/trace.go
oleiade: Having given this a quick thought, I do think it makes sense to have it in common, considering it was explicitly designed to be reusable by different parts/modules of k6. 

I also remember we already discussed doing something along those lines as we implemented the `k6/experimental/tracing` module, so 👍🏻 on my side.

oleiade:
@ka3de could you also fill the PR checklist accordingly so as a reviewer, I have an idea of where we're at? 🙇🏻

cmd/root.go
```diff
@@ -86,7 +96,21 @@ func (c *rootCommand) execute() {
...
+
+		// Wait for loggers and tracer provider to stop.
+		// As both methods already use a timeout, just wait.
+		var wg sync.WaitGroup
+		ff := [...]func(){c.stopLoggers, c.stopTracerProvider}
+		wg.Add(len(ff))
+		for _, f := range ff {
+			f := f
+			go func() {
+				f()
```
Blinkuu, line 108: nit: I would make `stopLoggers(...)` and `stopTracerProvider(...)` methods accept a `context.Context` parameter, as you seem to have a more "global context" available. If the global context gets canceled, it won't impact the other goroutines.

ka3de: If I'm understanding you correctly, the idea of making both `stopLoggers()` and `stopTracerProvider()` accept a `context` LGTM. So it would be similar to:
<details>
<summary>code</summary>

```go
func (c *rootCommand) stopLoggers(ctx context.Context) {
	done := make(chan struct{})
	go func() {
		c.loggersWg.Wait()
		close(done)
	}()
	close(c.stopLoggersCh)
	select {
	case <-done:
	case <-ctx.Done():
		c.globalState.FallbackLogger.Error("The logger didn't stop in time")
	}
}

func (c *rootCommand) stopTracerProvider(ctx context.Context) {
	if err := c.globalState.TracerProvider.Shutdown(ctx); err != nil {
		c.globalState.FallbackLogger.Errorf(
			"The tracer provider didn't stop gracefully: %v", err,
		)
	}
}
```
</details>

And then from `execute()`'s defer we could use a single context with timeout such as:
<details>
<summary>code</summary>

```go
var wg sync.WaitGroup
ctx, cancel := context.WithTimeout(context.Background(), waitLoggerTracerProviderStopTimeout)
defer cancel()
ff := [...]func(context.Context){c.stopLoggers, c.stopTracerProvider}
wg.Add(len(ff))
for _, f := range ff {
	f := f
	go func() {
		f(ctx)
		wg.Done()
	}()
}
wg.Wait()
```
</details>

Can you clarify what do you  mean with _"you seem to have a more 'global context' available"_?
Are you referring to the `c.globalState.Ctx`? For what I understand we are in control of that context from `execute()` and [canceling](https://github.com/grafana/k6/blob/bbc461bb4d598493d7df16c01c95f965ad2daed0/cmd/root.go#L98) it just before stopping loggers and tracer provider, so I think it should be OK to build the context with timeout off of `Background()`?

ka3de:
Once #3446 is merged, I will force push rebasing the changes along the modifications discussed in https://github.com/grafana/k6/pull/3445#discussion_r1386243267, and then set the PR up for review.

js/common/trace.go
ka3de: Due to the changes discussed in https://github.com/grafana/k6/pull/3445#discussion_r1386243267, the `TracerProvider` interface that defines the methods to be exposed to modules has been [moved](https://github.com/grafana/k6/blob/6225e08ce8e956d534da839f1116f3c8d346f088/lib/vu_state.go#L24) to `lib` package, as otherwise there was an import cycle between `common` and `lib` packages.

cmd/root.go
```diff
@@ -80,6 +84,12 @@ func (c *rootCommand) persistentPreRunE(cmd *cobra.Command, args []string) error
 		return err
 	}
 	c.globalState.Logger.Debugf("k6 version: v%s", consts.FullVersion())
+
+	err = c.setupTracerProvider()
+	if err != nil {
+		return err
+	}
+
```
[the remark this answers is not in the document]
mstoykov: This also means that it should not be part of the globalState ... probably

lib/trace/otel.go
```diff
@@ -0,0 +1,250 @@
...
+
+	if params.proto == "grpc" && params.urlPath != "" {
+		return params, ErrInvalidGRPCWithURLPath
+	}
+
+	return params, nil
+}
+
+func (p *tracerProviderParams) parseURL(url string) error {
+	u, err := stdurl.Parse(url)
```
mstoykov, line 220: nit: I would prefer to keep the package name isntead of the variable name

cmd/root.go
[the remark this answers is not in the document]
ka3de: Right, I will move the initialization/stopping of the `TracerProvider` to the _run_ command.
Can you provide more details on what would be the best approach to "propagate" the tracer provider from the run command to the VU? Does it make sense to modify the `Runner` [interface](https://github.com/grafana/k6/blob/641af0a002dedc756fae6814d0b4dc739a4dd529/lib/runner.go#L54) _NewVU_, and possibly _Setup_ and _Teardown_ methods to accept a `TracerProvider` as input and propagate the `TracerProvider` through the `Scheduler` implementation? Although this might imply quite a few changes.

Otherwise, we might keep it in `GlobalState` so it's propagated to `TestPreInitState` when [loading](https://github.com/grafana/k6/blob/641af0a002dedc756fae6814d0b4dc739a4dd529/cmd/run.go#L96) the test.

Differently from the metrics, where we control the full implementation, the underlying OTEL's `TracerProvider` already supports configuring multiple exporters, so we might not need a "manager" even in the long run?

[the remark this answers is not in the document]
mstoykov: Yeah, I looked at the code, and it will be nice if we can just use the same mechanism as the Samples channel, but it seems like a lot of work.

I am okay with you keeping at the `TestPreInitState` for this PR, but I think we should refactor this later.

This will currently work, but having this somehow get the credentials from `--out cloud` will be a nightmare.

cmd/root.go
Blinkuu: Regarding code snippets - yeah, this is what I had in mind. In general, it's a good practise to accept `context.Context` in shutdown functions that can be blocked. This pattern is widely used throughout stdlib ([example](https://cs.opensource.google/go/go/+/master:src/net/http/server.go;l=2979))

> Can you clarify what do you mean with "you seem to have a more 'global context' available"?
Are you referring to the c.globalState.Ctx? For what I understand we are in control of that context from execute() and [canceling](https://github.com/grafana/k6/blob/bbc461bb4d598493d7df16c01c95f965ad2daed0/cmd/root.go#L98) it just before stopping loggers and tracer provider, so I think it should be OK to build the context with timeout off of Background()?

Yes, I was referring to that context. The question is - Is there a use case where that global context gets canceled, but you don't want to shut down loggers and tracers? If not, I would probably derive the context you inject to shutdown functions from that global context and ensure that the "green path" executes your shutdown logic before the global context gets canceled. This would ensure that no matter what goes wrong, you can always shutdown these goroutines by canceling the global context.

cmd/root.go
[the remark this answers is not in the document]
ka3de: In relation with moving this to _run_ command and its exiting process. How should we treat traces as of now in this regard? Should we wait for all traces to be flushed before exiting, as it's done for metrics, or should we set a reasonable TO for it?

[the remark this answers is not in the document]
mstoykov: I mean - we do wait metric outputs, so why not wait on trace outputs? 

It makes little sense in the general case to not wait to output the traces and spans we've collected. 

Maybe adding some timeout is okay, but given that we don't really have one for metric outputs I would argue this falls outside of this PR.

Just please start stopping it *concurrently* with the metric outputs so we don't wait for one and then for the other.

lib/trace/otel.go
ka3de: Fixed in [88d4f84](https://github.com/grafana/k6/pull/3445/commits/88d4f84329ce50d96a0f576b826156b21a480e72).

cmd/root.go
ka3de: This discussion is superseded by the discussion held in https://github.com/grafana/k6/pull/3445#discussion_r1389525664, and particularly by commit [3fbc0b8](https://github.com/grafana/k6/pull/3445/commits/3fbc0b89391478708a53973701f64cfec7e509de).

cmd/root.go
```diff
@@ -196,6 +196,10 @@ func rootCmdPersistentFlagSet(gs *state.GlobalState) *pflag.FlagSet {
 		"enable profiling (pprof) endpoints, k6's REST API should be enabled as well",
 	)
 
+	flags.StringVar(&gs.Flags.TracesOutput, "traces-output", gs.Flags.TracesOutput,
+		"set the output for k6 traces, possible values are none,otel[=host:port]")
+	flags.Lookup("traces-output").DefValue = gs.DefaultFlags.TracesOutput
+
```
mstoykov, line 202: This should be in the [runtime options](https://github.com/grafana/k6/blob/master/cmd/runtime_options.go) and only be parsed and applied on `k6 run` which I think it does automatically if it is there :thinking:

mstoykov (commented):
LGTM in general - just move the flag configuration to the runtime options.

I have not really looked at the tracing parts - I expect @Blinkuu  will be way more useful there.

I would also prefer that with/after the merge from k6-browser where this is used to get an integration test that mocks a tracing endpoint and checks that we get something useful.

cmd/root.go
ka3de: I believe there is an issue related to moving the _traces output_ flag to the runtime options. The runtime options are parsed in the call to [`loadAndConfigureTest`](https://github.com/grafana/k6/blob/85fdca1c857bae7609b7efc248eab98b6ad0a888/cmd/run.go#L114) from `run` command execution, which is the same function that sets the `TracerProvider` in `TestPreInitState` from global state, which means that at that time the `TracerProvider` must have been initialized.

So what do you think about moving the _traces output_ option to the runtime options and then adding a new method to just parse this option, such as:
<details>
<summary>code</summary>

```go
// TODO: Merge this into getRuntimeOptions when ...
func getTracesOutputOption(flags *pflag.FlagSet, environment map[string]string) null.String {
	to := getNullString(flags, "traces-output")
	if envVar, ok := environment["K6_TRACES_OUTPUT"]; ok {
		if !to.Valid {
			to = null.StringFrom(envVar)
		}
	}

	return to
}
```
</details>

Then this will be called from `setupTracerProvider` method. We can still add the `TracesOutput` field to `lib.RuntimeOptions` and [call](https://github.com/grafana/k6/blob/85fdca1c857bae7609b7efc248eab98b6ad0a888/cmd/runtime_options.go#L70) this `getTracesOutputOption` method from `getRuntimeOptions` to parse it, so we can add the corresponding tests for it, and in the future just remove this method call and merge the implementation in `getRuntimeOptions`.

Not the nicest. but might be the best option until the `Runner` interface is modified to accept the `TracerProvider` as input so this can be initialized later and use the parsed configuration from `loadAndConfigureTest` call.

mstoykov: I am not certain why you can't just move the tracer initalization *after* the call to `loadAndConfigureTest` at this point ?

cmd/root.go
mstoykov: I will resolve this at this point

cmd/root.go
mstoykov: By the looks of it you can:
1. move the whole configuration in the runtime options.
2. remove anything about it from the global state
3. use the TestPreinitstate to both get the runtime optiosn and set it's tracer output in the setupTracerProvider

ka3de: That seems a little bit "obscure" to me, as we are modifying the `TracerProvider` by reference, when this one has already been propagated along with the `TestPreInitState` to the [`Runner`](https://github.com/grafana/k6/blob/ef550a555115fe8397a0a099055a2b750f4fef4a/js/runner.go#L95) in the call to `loadAndConfigureTest`.
Done in [5a24571](https://github.com/grafana/k6/pull/3445/commits/5a24571f66f9dadefb901a36d1c029bee6bf2534)..[ef550a5](https://github.com/grafana/k6/pull/3445/commits/ef550a555115fe8397a0a099055a2b750f4fef4a).

Blinkuu (commented):
LGTM

Blinkuu (approved):

cmd/run.go
```diff
@@ -407,6 +440,22 @@ func (c *cmdRun) flagSet() *pflag.FlagSet {
...
 }
 
+func (c *cmdRun) setupTracerProvider(test *loadedAndConfiguredTest) error {
+	ro := test.preInitState.RuntimeOptions
+	if ro.TracesOutput.String == "none" {
+		test.preInitState.TracerProvider = trace.NewNoopTracerProvider()
+		return nil
+	}
+
+	tp, err := trace.TracerProviderFromConfigLine(c.gs.Ctx, ro.TracesOutput.String)
```
mstoykov, line 450: nit: you can provide the `globalcCtx` as is done in the cancelation timeout

mstoykov (approved):
LGTM! 

left a small nitpick

mstoykov:
Also please either squash the whole thing on merge or make it into 2(?) commits instead of 15

cmd/run.go
ka3de: Right, that makes total sense. Fixed in [663c6c2](https://github.com/grafana/k6/pull/3445/commits/663c6c2e9d2ad597f4ca419107e4783dc465507d).

ka3de:
> Also please either squash the whole thing on merge or make it into 2(?) commits instead of 15

I don't have permissions to merge. I would say squash at this point if that works for you. Otherwise let me know and I'll try to organize it in a few commits that make sense.

mstoykov (approved):

oleiade (approved):

oleiade (approved):

PR #3446, opened 2023-11-08
Move log tokenizer to `lib/strvals` package

ka3de:
## What?

Moves the tokenizer implementation, currently only used to parse the log output option configuration, to a more generic package.

## Why?

Up until now the tokenizer implementation was only used to parse the log output option configuration. With our current intention of using this format for other options, such as traces output, it becomes evident that this implementation should be moved away from the log package and into a more generic one.

## Checklist

## Related PR(s)/Issue(s)

Related #3445


mstoykov (approved):
LGTM!

My only nit is that I kind of don't like `parseutils` in part due to the `utils` part.

But `tokenizer.Tokenize` is no better as well as `tokens.Parse` or `tokens.New`

So unless @oleiade comes up wiht something better I guess this is it.

ka3de:
> My only nit is that I kind of don't like parseutils in part due to the utils part

Yep, to be honest I don't quite like it either. I considered `parse` or `parseutil`, but because there was already a `testutils` package, decided to be coherent with that. But open to any suggestion for another name.

oleiade:
I like `strvals` 👍🏻 Let's go for that if you're okay with it @ka3de 🙇🏻

ka3de:
> My reasoning is that the parsing mechanism described in the package is quite specific to the key=value,... format, which is meant to be used in logs, and I don't think that calling parseutils.Tokenize is very clear on what to expect from it

I guess this depends also on whether this package will hold more parsing implementations in the future. But as of now I agree that the implementation is pretty specific.

> I would vote for making the name of the package as specific and explicit as possible, rather than broad and generic. Thus, I would suggest something along the lines of lib/logparser, or lib/kvparser package or something along those lines instead.

My understanding is that the format implemented by `Tokenize` is not strictly related with log/s. We are actually now, with this PR, using it to parse the _traces output_ configuration. So to me `kvparser` is a better fit as of now :+1:

lib/strvals/parser.go
```diff
@@ -1,9 +1,10 @@
-package log
+package strvals
 
 import "fmt"
 
-type token struct {
-	key, value string
+// Token represents a key-value token.
+type Token struct {
```
ka3de, line 6: After the renaming to `strvals` package and `Parse` method, I left this as `Token` as the definition has not changed, but I could also see this and `tokenizer` changing to:
```
s/Token/KV
s/tokenizer/parser
```

mstoykov (approved):
LGMT!

Just rename the PR title before merge ;)

lib/strvals/parser.go
mstoykov: The latter can definitely be changed later as it is not exported.

I am don't think `Token` is not specific enough - so it is okay IMO.

oleiade (approved):

ka3de:
I have no permission to merge the PR. Could you merge this please?

PR #3447, opened 2023-11-09
Add request interception fix to release notes

inancgumus:
## What?

Add grafana/xk6-browser#1078 to release notes.

## Why?

This adds a note about a network interception deadlock fix.

## Related PR(s)/Issue(s)

grafana/xk6-browser#1072

### DIFF
--- a/release notes/v0.48.0.md
+++ b/release notes/v0.48.0.md
@@ -33,6 +33,7 @@ If no filename is provided, the default filename `script.js` will be used. The s
 
 - [browser#1077](https://github.com/grafana/xk6-browser/pull/1077) Fixes `browserContext.clearPermissions` to clear permissions without panic.
 - [browser#1042](https://github.com/grafana/xk6-browser/pull/1042) Fixes `browserContext.waitForEvent` which involved promisifying the `waitForEvent` API.
+- [browser#1078](https://github.com/grafana/xk6-browser/pull/1078) Fixes request interception deadlock to improve stability.
 
 ## Maintenance and internal improvements
 


ankur22 (approved):
LGTM

olegbespalov:
@inancgumus, the changes look minimal. Do you think they still need to be done in the separate PR instead of just adding the commit to the [release-notes-v0.48.0](https://github.com/grafana/k6/tree/release-notes-v0.48.0) branch?

My concern is that, in any case, we're doing an iteration of the reviewing once we create the PR from `release-notes-v0.48.0` to master, so having fewer PRs simplifies the process without harming the quality and producing additional noise. What are your thoughts here?

inancgumus:
@olegbespalov, we found that reviewing changes one by one helps us tackle issues early. If it's a small tweak, merging it directly _should_ be fine. But what counts as 'small' can be subjective. How about we outline what 'small' looks like in our case? I'm open to hear other ideas.

olegbespalov:
@inancgumus

Here are my thoughts.

Having the PRs for every line could also harm the process since they create so much noise. For example, I am I wonder if k6-core should review such PRs at that stage.

Also, As the release driver at the end of the cycle, I should check that all merged PRs will be mentioned in the release notes appropriately since we have many inputs, like the changes done by team members and changes that could come from the community. Imagine that PRs like that should be filtered because they will be part of the main release notes PR.

I'm not suggesting you thoroughly eliminate the practice of reviewing the per-issue change. We can do the review directly in the release notes PR before the last weeks. Opening a separate pull request for more extensive changes is OK, but there should be some rationale. I agree with you that `small` is pretty subjective. But IMO, this PR change couldn't be an example of what could be quickly done with commit. And for sure, we should define the char limit for measuring. 

In the future, we could solve this with automation, but till that point, let's try to find some balance that will allow you to do reviews and doesn't overcomplicate the final process of shaping the release notes.

@grafana/k6-core, if you have an opinion, please feel free to chime in.

olegbespalov:
Okay, an update that I found a way to solve _my pain_ I'll just use the search like: https://github.com/grafana/k6/pulls?q=is%3Apr+is%3Aclosed+base%3Amaster+merged%3A%3E2023-10-10

So @inancgumus feel free to ignore my need.

olegbespalov:
@mstoykov I see that you've added a milestone. 

Do you think it's needed for this PR, considering that it targets the branch that will later be merged with the master? 

Honestly, I have a feeling the final PR (release notes) should only be labeled with milestones to keep in milestones the most valuable items and fewer dublications.

mstoykov:
I think I just added to a bunch of prs while searching for something else.

I still prefer if all prs are milestones , but is definitely not a requirement

PR #3448, opened 2023-11-09
metrics: enforce more otel/prometheus aligned names

### LINKED ISSUE #3065: Restrict metric names to what is supported by Prometheus

mstoykov:
Closes #3065

### DIFF
--- a/metrics/registry.go
+++ b/metrics/registry.go
@@ -28,7 +28,8 @@ func NewRegistry() *Registry {
 
 const (
 	nameRegexString = "^[a-zA-Z_][a-zA-Z0-9_]{1,128}$"
-	badNameWarning  = "Metric names should start with a letter or an underscore and only include up to 128 ASCII letters, numbers, or underscores."
+	badNameWarning  = "Metric names must only include up to 128 ASCII letters, numbers, or underscores " +
+		"and start with a letter or an underscore."
 )
 
 var compileNameRegex = regexp.MustCompile(nameRegexString)
--- a/metrics/registry.go
+++ b/metrics/registry.go
@@ -26,7 +26,10 @@ func NewRegistry() *Registry {
 	}
 }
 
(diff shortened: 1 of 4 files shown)


metrics/registry.go
```diff
@@ -26,7 +26,10 @@ func NewRegistry() *Registry {
 	}
 }
 
-const nameRegexString = "^[\\p{L}\\p{N}\\._ !\\?/&#\\(\\)<>%-]{1,128}$"
+const (
+	nameRegexString = "^[a-zA-Z_][a-zA-Z0-9_]{1,128}$"
+	badNameWarning  = "Metric name should only include up to 128 ASCII letters, numbers and/or underscores."
```
olegbespalov, line 31:
```suggestion
	badNameWarning  = "Metric names should start with a letter or an underscore and only include up to 128 ASCII letters, numbers, or underscores."
```

olegbespalov (dismissed):

mstoykov:
I reedited the message while fixing the test as it seemed strange to not start with what it should include and `letter` needed to become `ASCII letter`

But we can edit it more or revert it.

olegbespalov (approved):

oleiade (approved):

PR #3449, opened 2023-11-09
Erroring on modules with no scheme

### LINKED ISSUE #3287: Remove trying to load remote modules by prepending `https://`
## What?

if currently we get module name "something" k6 will try to load it as `https://` + `something` in case it is a URL without a schema, which used to be the only way to load urls more than 4 years ago.

This is mostly confusing users when they make mistake with importing something instead of actually helping.

## History

This functionality dates back to before my involvement with the project and in one of my first big rewrite #1046 it was left as backwards compatibility. Especially as that PR also let users actually prepend `https://` infront of modules specifiers which made them more consistent and obvious. 

## Solution
Given that it has been 4 years - I am for dropping this functionality. I very much doubt anyone still uses that, and if they do they have been seeing the message all of this time.

We can also change the message to us actually deprecating it one release and removing it a few releases later, to be more consistent and to not break someone script. Although again I really doubt anyone has been using this with this message constantly logging.

olegbespalov:
## What?

As was planned in the https://github.com/grafana/k6/issues/3287 we removed the support of adding `https://` for the module resolution.

## Why?

https://github.com/grafana/k6/issues/3287

## Checklist

## Related PR(s)/Issue(s)

Closes #3287

### DIFF
--- a/loader/loader.go
+++ b/loader/loader.go
@@ -198,11 +198,8 @@ func Load(
 				return nil, err
 			}
 		case moduleSpecifier.Scheme == "":
-			logger.Warningf(`The moduleSpecifier "%s" has no scheme but we will try to resolve it as remote module. `+
-				`This is deprecated and will be removed in v0.48.0 - all remote modules will `+
-				`need to explicitly be prepended with "https://".`, originalModuleSpecifier)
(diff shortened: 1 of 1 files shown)


olegbespalov:
I am closing in favor of the solution with more code cleaning involved.

PR #3456, opened 2023-11-15
Native ESM support

### LINKED ISSUE #3265: Native ESM support
## History 

k6 uses a [goja](https://github.com/dop251/goja) as JS VM. Historically goja does not have support for every JS feature, and as such [babel](https://babeljs.io/) has been used for years to supplement that. The way this works is that if goja returns an error while parsing, k6 will try run it through babel (see [compatibility mode](https://k6.io/docs/using-k6/javascript-compatibility-mode/) for more info)

Over the years though all the things babel has been used have been added to goja. Actually goja does now have support for features that the version of babel used by k6 does not have support for:

- Object spread [#824](https://github.com/grafana/k6/issues/824) has been supported for years at this point
- Optional chaining [#2168](https://github.com/grafana/k6/issues/2168) has also been supported for years in goja
- Class private fields [#2887](https://github.com/grafana/k6/issues/2887) was added along all the class support

All of those are currently not *really* support in k6 as they are not supported by the internal babel which needs to parse them even if it won't transform them. As such using this features with ESM lead to parsing errors.

## Solution: 

The solution here is to have ESM support in goja and used by k6. This will:
1. drop babel from the k6 binary
2. drop babel from needing support for whatever k6 needs support. Letting us only worry about how goja supports features.
3. Due to babel being JS that needs to be run by goja 
	1. speed up starting times by 200-500ms (in my small level testing).
	2. speed up starting times for big scripts by quite a lot more. Big scripts are > 10k lines and they start to take multiple seconds (sometimes minutes) to be parsed by babel, but goja parses them much faster (<1s). 

## Status: 

There is [draft PR](https://github.com/dop251/goja/pull/430) to add the functionality to goja, with the following high level problems:
1. ~It currently doesn't work with `eval` - which is definitely blocking~
2. ~The current way some of the evaluation happens predates async/await support and is not based on it - it is actually 100% a hack. This seems to also be a reason for somewhat more complicated scripts to *not* work.~
3. ~Connected to the above the current code does predate async/await and does not have support for top-level-await. This is likely blocking due to goja usually adding support for the full set of features around a given concept when that one is introduced. My experience is that this will be super simple once `2` is done~
4. It has only been lightly reviewed by the goja author and maintainer as it predating async/await was a blocking at the time. 
5. The API exposed to users of goja (k6 for example) - evolved naturally while I was making progress. This again predates async/await - it even predated modules being asynchronous in the specification, which was added in ES2022. As such at that point the API suddenly bubbled up and is currently - "not great™". As such it will likely be a good idea to try to fix that once the first 3 parts are fixed and some discussion with the maintainer of goja. 

There is an even more WIP/PoC [PR](https://github.com/grafana/k6/pull/2563) to k6 core that is very dated at this point. This should be updated along/after the goja PR is in better shape. 

Big parts of the original k6 PR have been backported to the current k6 codebase making this a lot easier, but still some work is required.

The major blocking part is the goja PR which also happens to be more complicated than expected and all the issues there are very interconnected and connected to the internals of goja runtime.

### Other connected issues with some context on them

- [#2674](https://github.com/grafana/k6/issues/2674) - currently `require` is not relative to the file it is *written* in, but to the one we are evaluating. This is not how commonjs defines it and will be somewhat harder (impossible in some cases?!?) to be implemented once ESM default support is added. Especially if users are mixing ESM and commonjs.
- [#2949](https://github.com/grafana/k6/issues/2949) - my current WIP code does not require breaking the modules API k6 exposes to extensions and internally. But in order for some additional features - for example one JS [module accessing another](https://github.com/grafana/k6/issues/2293) we will probably need changes. 
- [#2258](https://github.com/grafana/k6/issues/2258) - is mostly closed, the current implementation support differentiating between default and named exports. But the exact API might change with #2949 being taken into consideration as well.

mstoykov:
## What?

Native support for ESM in k6

## Why?

Having native ESM support both allows us to use more goja features and drops our internal babel. 

This also closes a number of js-compat issues as we no longer are blocked by babel not supporting something. As well as freeing up the ability for Sobek to add support for even newer things. 

The `breaking change` label is there as this both breaks old `require` behavior as part of #3534, but also now makes combining CommonJS and ESM syntax in one file a syntax error. The latter has been found to be a lot more common than the first, unfortunately. But there isn't any sane way to support that, and it also makes no sense to be mixed. 

## Checklist

## Related PR(s)/Issue(s)

Closes #3265


js/bundle.go
```diff
@@ -409,9 +430,25 @@ func (b *Bundle) setInitGlobals(rt *sobek.Runtime, vu *moduleVUImpl, modSys *mod
...
+		pwd, err := modSys.CurrentlyRequiredModule()
+		if err != nil {
+			return nil, err
+		}
+		return openImpl(rt, b.filesystems["file"], pwd, filename, args...)
 	})
+	warnAboutModuleMixing := func(name string) {
+		warnFunc := rt.ToValue(func() error {
+			return fmt.Errorf(
+				"you are trying to access identifier %q, this likely is due to mixing ESM and CommonJS syntax",
```
mstoykov, line 442:
```suggestion
				"you are trying to access identifier %q, this likely is due to mixing ESM and CommonJS syntax." + 
				"This isn't supported in the JavaScript standard, please use only one or the other",
```

cc @sniku @grafana/k6-core

js/bundle.go
```diff
@@ -436,6 +436,20 @@ func (b *Bundle) setInitGlobals(rt *sobek.Runtime, vu *moduleVUImpl, modSys *mod
 		}
 		return openImpl(rt, b.filesystems["file"], pwd, filename, args...)
 	})
+	warnAboutModuleMixing := func(name string) {
+		warnFunc := rt.ToValue(func() error {
+			return fmt.Errorf(
+				"you are trying to access identifier %q, this likely is due to mixing ESM and CommonJS syntax."+
+					"This isn't supported in the JavaScript standard, please use only one or the other",
```
codebien, line 443: > It tried to access the _module_ identifier, this operation is not supported by JavaScript ECMAScript standard. It happens if the script mixes CommonJS and ECMAScript Modules syntaxes.

Suggesting an alternative, feel free to discard or change

mstoykov: It is possible that it was coding mistake though, which is why I don't want to say it is definitely mixing.

And the standard lets you have idenfitiers that are `module` or `exports` - just in this case they weren't defined until the global object

mstoykov: To be honest I am not particularly happy with the current wording either, just pointing out why it was choosen and worded in the particular way.

codebien: What about this new suggestion?

>There was an attempt to access the _module_ identifier. It might be an issue generated by the script's code if it is mixing CommonJS and ECMAScript Modules syntaxes, which is not supported by JavaScript ECMAScript standard.

I'm using _module_ just as a placeholder, of course, we will have there a `%q``.

mstoykov: JavaScript ECMAScript standard sounds ... bad to me, and "generated by the script's code":
1. generated is not the use I will use. Arguably this makes the whole sentence longer than "It might be an issue with code that mixes ...." 
2. script's code - technically here in one of the cases it isn't a script it is a module - that is what ESM denotes.

Other than that I am okay with both the current and your wording, and do not have a strong preference, even the my nitpicks are not very serious IMO.

I just would prefer for us to discuss it with the other reviewers before I make one more change so I stop changing the message :)

codebien: @joanlopez @olegbespalov we need your input here :pray:

joanlopez: I think the current version is good enough, except that I don't know how clear are the specific terms ESM, CommonJS, and JavaScript standard.

olegbespalov: I don't have a strong opinion, however I do believe that just ESM isn't enough, IMO we should use and refer to it as `ECMAScript Modules (ESM)`

js/modules/require_impl.go
```diff
@@ -2,125 +2,139 @@ package modules
...
+			return nil, err
+		}
+		mi = rt.GetModuleInstance(wm)
 	}
-
-	if specifier == "" {
-		return nil, errors.New("require() can't be used with an empty specifier")
+	gmi, ok := mi.(*goModuleInstance)
+	if !ok {
+		panic("a goModule instance is of not goModuleInstance type. This is a k6 bug please report!")
```
olegbespalov, line 84:
```suggestion
		panic(fmt.Sprintf("a goModule instance is of not goModuleInstance type (got %T). This is a k6 bug please report it (https://github.com/grafana/k6/issues)", mi))
```

js/modules/require_impl.go
```diff
@@ -2,125 +2,139 @@ package modules
...
+		}
+	case strings.HasPrefix(fileStr, "https://"):
+		var err error
+		u, err = url.Parse(fileStr)
 		if err != nil {
 			return nil, err
 		}
-		return loader.Dir(u), nil
+	default:
+		return nil, fmt.Errorf("couldn't parse %q as module url - this is a k6 bug, please report it", fileStr)
```
olegbespalov, line 115:
```suggestion
		return nil, fmt.Errorf("couldn't parse %q as module url - this is a k6 bug, please report it (https://github.com/grafana/k6/issues)", fileStr)
```

js/modules/resolution.go
```diff
@@ -95,14 +128,19 @@ func (mr *ModuleResolver) Lock() {
 	mr.locked = true
 }
 
-func (mr *ModuleResolver) resolve(basePWD *url.URL, arg string) (module, error) {
-	if cached, ok := mr.cache[arg]; ok {
-		return cached.mod, cached.err
-	}
+// fix
```
olegbespalov, line 131: What should we fix here? :thinking:

olegbespalov (commented):
I did an initial round of review and it looks good :rocket: I left mostly minors. Great work! :muscle:

js/modules/cjsmodule.go
```diff
@@ -1,78 +1,118 @@
...
+func (cm *cjsModule) Instantiate(_ *sobek.Runtime) (sobek.CyclicModuleInstance, error) {
+	return &cjsModuleInstance{w: cm}, nil
 }
 
-func (c *cjsModule) instantiate(vu VU) moduleInstance {
-	return &cjsModuleInstance{vu: vu, mod: c}
+func (cm *cjsModule) RequestedModules() []string { return nil }
+
+func (cm *cjsModule) Evaluate(_ *sobek.Runtime) *sobek.Promise {
+	panic("this shouldn't be called in the current implementation")
```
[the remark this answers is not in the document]
mstoykov: No - the implementation should never call this - as in like this is a design problem with the API Sobek exposes and given what happens this method will never be called, but we still need to implement to satisfy the interface.

js/modules/gomodule.go
```diff
@@ -4,89 +4,99 @@ import (
...
 	}
-	return b.exportsO
+	return &goModuleInstance{rt: rt, mi: mi}, nil
 }
 
-// goModule is a go module which implements Module
-type goModule struct {
-	Module
-}
+func (gm *goModule) Evaluate(_ *sobek.Runtime) *sobek.Promise { panic("this shouldn't happen") }
```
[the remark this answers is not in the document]
mstoykov: Same as the other one - this should just never be called by Sobek, we just need to have it for interface satisfaction.

js/bundle.go
mstoykov: I  can change it ... but IMO if people don't know what ESM is in this context with CommonJS ... ECMAScript Modules will be just as hard to understand.

I am not against it, just please make a code suggestion.

olegbespalov: > if people don't know what ESM is in this context with CommonJS

That's the trick. If they don't know what it is ESM, they probably don't know ECMAScript Modules either, but the `ECMAScript Modules` are more searchable IMO

So my minimal requested change looks like this:
```suggestion
				"you are trying to access identifier %q, this likely is due to mixing ECMAScript Modules (ESM) and CommonJS syntax."+
					"This isn't supported in the JavaScript standard, please use only one or the other",
```

For the other changes in the sentence I don't have strong opinion :shrug:

js/modules/gomodule.go
```diff
@@ -4,89 +4,99 @@ import (
...
 
-	for k, v := range exp.Named {
-		result[k] = v
+	// if there are only named exports we make a default object out of them
+	// this allows scripts to modify this acting similar to how it would act
+	// if the default export was an object to begin with.
+	o := gmi.rt.NewObject()
+	gmi.defaultExport = o
+	for name, value := range exports.Named {
+		// TODO:maybe do something slightly smarter
```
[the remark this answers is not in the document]
mstoykov: I think I was hoping I can do something where we only import and set them if/when they are used. But that has pretty big problems in the general case where that might happen *after* init context, so I am just going to drop it.

Using field accessors doesn't seem like it will add any functionality here ... maybe for counting what is used, but that is a separate problem.

js/modules/require_impl.go
```diff
@@ -2,125 +2,139 @@ package modules
...
+	if exp.Default == nil {
+		return exp.Named
 	}
+
+	result := make(map[string]interface{}, len(exp.Named)+2)
+
+	for k, v := range exp.Named {
+		result[k] = v
+	}
+	result["default"] = exp.Default
```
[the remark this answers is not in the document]
mstoykov: Did that, but the `__esModule` is here only for fallback - nothing in k6 is using it now when Babel is not here.

js/modules/resolution.go
mstoykov: Drop this - there are other comments where it used, but I am not doing this now as it requires Sobek work.

codebien (approved):

js/compiler/compiler.go
```diff
@@ -1,94 +1,29 @@
...
-	errGlobalBabelCode error          //nolint:gochecknoglobals
-	onceBabel          sync.Once      //nolint:gochecknoglobals
-	globalBabel        *babel         //nolint:gochecknoglobals
-)
-
-const (
-	maxSrcLenForBabelSourceMapVarName = "K6_DEBUG_SOURCEMAP_FILESIZE_LIMIT"
-	sourceMapURLFromBabel             = "k6://internal-should-not-leak/file.map"
-)
+// A Compiler compiles JavaScript source code (ES5.1 or ES6) into a sobek.Program
```
joanlopez, line 21:
```suggestion
```

Looks like this comment is duplicated from the one below.

mstoykov: I already have a branch that goes and fixes this and other comments, I would prefer to not block this PR on this.

There is going to quite a lot of cleanup either way.

olegbespalov (approved):
:rocket:

oleiade (approved):

Issue #769, opened 2018-09-14
Fix inter issues in `master` branch


na--:
# What?

We should fix the linter issues that exist in our master branch.

```[tasklist]
### Tasks
- [ ] https://github.com/grafana/k6/pull/3701
- [ ] https://github.com/grafana/k6/pull/3566
- [x] [Fix 51 issues in lib](https://github.com/grafana/k6/pull/3561)
- [x] Fix 22 issues in cloudapi
- [x] Fix 39 issues in metrics
- [x] Fix rest of the packages
- [ ] https://github.com/grafana/k6/pull/3495
- [ ] https://github.com/grafana/k6/pull/3484
- [ ] https://github.com/grafana/k6/pull/3482
- [ ] https://github.com/grafana/k6/pull/3460
- [ ] https://github.com/grafana/k6/pull/3463
- [ ] https://github.com/grafana/k6/pull/3479
- [ ] https://github.com/grafana/k6/pull/3481
- [ ] https://github.com/grafana/k6/pull/3462
- [ ] https://github.com/grafana/k6/pull/3480
- [ ] https://github.com/grafana/k6/pull/3485
- [ ] https://github.com/grafana/k6/pull/3483
- [ ] https://github.com/grafana/k6/pull/3478
```

Right now we use the `golangci-lint` as the main tool for linting. The latest report:

```bash
golangci-lint run --out-format=tab | awk 'NF { print $1 }' | sed -E 's|/[^/]*(:[0-9]+){0,2}$||' | uniq -c | sort -rn
 233 js/modules/k6/html
    112 js/modules/k6/http
     48 lib
     38 metrics
     32 lib/executor
     28 lib/netext/httpext
     26 lib/netext
     22 cloudapi
      5 lib/fsext
      5 lib
      3 output/statsd
      3 output/json
      3 js/modules/k6/grpc
      2 lib/types
      2 lib/netext/grpcext
      2 lib/consts
      2 js/modules/k6/metrics
      2 js/common
      2 cmd
```

# Why?

Most of those are about undocumented methods or using underscores or ALL_CAPS in variables, but it's still worth it to spend some time and eliminate, or at least substantially reduce, the number of linter complaints. 

Fixing linter issues would significantly help any potential contributors because:
- it forces best practices
- it would force us to improve the in-code documentation, tremendously helping new people who are unfamiliar with the codebase
- it would make the code more idiomatic
[...]

PR #3460, opened 2023-11-20
Linter fixes for the crypto and encoding k6 modules

olegbespalov:
## What?

This PR fixes ~105 linter issues in `js/k6/crypto` and `js/k6/encoding` packages.

It's recommended to start looking by commits.

## Why?

As we agreed, we aim to fix most of the linter issues.

Part of #769

## Checklist

- [x] I have run linter locally (`make lint`), and all checks pass.
- [x] I have run tests locally (`make tests`), and all tests pass.

### DIFF
--- a/js/modules/k6/crypto/crypto.go
+++ b/js/modules/k6/crypto/crypto.go
@@ -30,7 +30,8 @@ type (
 
 	// Crypto represents an instance of the crypto module.
 	Crypto struct {
-		vu modules.VU
+		randReader func(b []byte) (n int, err error)
+		vu         modules.VU
 	}
 )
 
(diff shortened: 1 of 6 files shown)


olegbespalov:
Converted to a draft for data race investigation :thinking:

olegbespalov:
The latest commit does the trick and fixes the data race. However, I prefer to drop a `RandomBytesFailure` test case since, in my opinion, it's not a big value to have it :shrug: What do you think?

oleiade (approved):
🚀 

I agree with you, the `RandomBytesFailure` does not look obviously useful to me, and I'd be happy with dropping it indeed 👍🏻

codebien (approved):
>I prefer to drop a RandomBytesFailure test case since, in my opinion, it's not a big value to have it

I don't have a strong opinion but I found it useful, it is testing that in case of an error returned from the reader then the `randomBytes` function returns an error. Error handling, in the end, is business logic for the `randomBytes` func.

PR #3464, opened 2023-11-21
fix handleSummary do not check if directory exists

### LINKED ISSUE #3084: handleSummary do not check if directory exists
### Brief summary

If I forget to create my output directory, runs will fails in the after-stage when creating summary.

Nice to have a way to verify this upfront that output dir is missing if you have set it.

Just add some simple example how to verify this.

### k6 version

0.44.1

### OS

windows 11

_No response_

Use a handleSummary like this:

```
export function handleSummary(data) {
  const extras = Object.assign(
    {
      testenv: testenv,
      tenant: tenant
    },
    data
  );
  return {
    "./report/result.html": htmlReport(data),
    stdout: textSummary(data, { indent: " ", enableColors: true }),
    "./report/result.xml": jUnit(data),
    "./report/summary.json": JSON.stringify(extras, null, 2)
  };
}
```

If I forget to create `results` before I run I loose the summary:

```
:
    vus_max........................: 50     min=50      max=50
ERRO[0011] failed to handle the end-of-test summary      error="Could not save some summary information:\n\t- could not open './report/summary.json': open ./report/summary.json: The system cannot find the path specified.\n\t- could not open './report/result.html': open ./report/result.html: The system cannot find the path specified.\n\t- could not open './report/result.xml': open ./report/result.xml: The system cannot find the path specified."
```

### Expected behaviour

Create the summary files

### Actual behaviour

crash and no summary files

jsolana:
## What?

Fix create the parent directories if they don't exist yet described in #3084  creating the parent directories if they don't exist yet, like `mkdir -p` would do.

## Why?

k6 run will fails in the after-stage when creating summary if  the output directories don't exist previously

## Checklist

## Related PR(s)/Issue(s)

Closes #3084

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 /dist
 /pkg-build
 /js/tc39/TestTC39
+/data
 
 .vscode
 *.sublime-workspace
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -9,6 +9,7 @@ import (
 	"io"
 	"net/http"
 	"os"
+	"path/filepath"
 	"runtime"
 	"strings"
 	"sync"
@@ -541,6 +542,9 @@ func handleSummaryResult(fs fsext.Fs, stdOut, stdErr io.Writer, result map[strin
 		case "stderr":
 			return stdErr, nil
 		default:
+			if err := fs.MkdirAll(filepath.Dir(path), 0o700); err != nil {
+				return nil, err
+			}
 			return fs.OpenFile(path, syscall.O_WRONLY|syscall.O_CREAT|syscall.O_TRUNC, 0o666)
 		}
 	}
--- a/cmd/run_test.go
+++ b/cmd/run_test.go
@@ -86,6 +86,24 @@ func TestHandleSummaryResultSimple(t *testing.T) {
 	require.Empty(t, stderr.Bytes())
 }
 
+func TestHandleSummaryResultFilePathWithParentDirectories(t *testing.T) {
+	t.Parallel()
+	content, stdout, stderr, fs := initVars()
+
+	filePath1 := "./path/file1"
+	filePath2 := "./path/file2"
+	if runtime.GOOS == "windows" {
+		filePath1 = ".\\path\\file1"
+		filePath2 = ".\\path\\file2"
+	}
+
+	content["stdout"] = bytes.NewBufferString("some stdout summary")
+	content["stderr"] = bytes.NewBufferString("some stderr summary")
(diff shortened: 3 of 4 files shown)


oleiade:
Hey @jsolana 👋🏻 

Thanks for your contribution 🤝 

To be fair and transparent with you, I'm torn about doing this. My reasoning is the following:
1. On one hand, I don't think it's k6's responsibility to manage the filesystem and ensure a directory exists. Most common tools I can think of will produce an error if you configure them to write their artifacts in a custom folder that doesn't exist (instead of creating it themselves).
2. On the other hand, I want to acknowledge that the summary we're producing is a byproduct of a k6 run, and it could be argued that if we allow to specify a destination that contains a path, it could make sense to allow k6 to make sure the destination exists. Also, k6 itself doesn't let you create directories from your scripts which adds to the argument.

Overall, I'm rather against merging this as per ☝🏻 and as there is an easy workaround: either manually ensuring your destination folder exists, or scripting it as part of the testing environment.

jsolana:
Ey @oleiade, sorry I was out cause a little break (and celebrating my birthday!)
Im fine with your point of view, my main concern is, somethings you are running costly tests and with no previous control, we are failing saving the information.
When you say:
>either manually ensuring your destination folder exists, or scripting it as part of the testing environment.

oleiade:
Happy birthday @jsolana 🍰 🎂 🎁 🍾 🎉 

> What do you mean?

What I had in mind was that you would chain the `k6 run` command with an `mkdir -p` statement to make sure the destination exists:

```bash
mkdir -p /some/folder/holding/data && k6 run script.js
```

Which could also be integrated as part of a CI job, and so on.

Please keep in mind that my understanding of the issue you're trying to solve might be limited, though, and that without more thorough and concrete examples, I might also be missing the point 👍🏻 

Please feel free to go ahead with `first good issue` issues. We're very keen on receiving contributions, and we'd be happy to provide the support you need. My recommendation would be that you start by looking at the issue, and the related code, and start by asking any questions you might have as a result, before opening a PR 🙇🏻 

I hope that's helpful, and happy birthday again!

jsolana:
Thanks! and thanks a lot for the clarifications and tips (make total sense start talking about instead to do the PR directly :D )

I assume, more or less the solution was defined cause [this comment](https://github.com/grafana/k6/issues/3084#issuecomment-1557197806) from @imiric 

In the case of this:

```
mkdir -p /some/folder/holding/data && k6 run script.js
```


Do you think make sense to modify the documentation to provide any kind of awareness?

Thanks!

PR #3470, opened 2023-11-25
handling large vu shifts

### LINKED ISSUE #1693: Huge VU numbers in ramping-vus result in a panic
The following example:
```js
export let options = {
  stages: [
    {
      target: 40000000000000,
      duration: "1m",
    },
  ],
};

export default function () {
}
```
And some smaller numbers than that will result in a `fatal error: runtime: out of memory`. We should add some validation in here: https://github.com/loadimpact/k6/blob/bf826ba614ded605c93302d04464ab0252c63a6b/lib/executor/ramping_vus.go#L99-L106

That checks if the total number of VU up and down shifts exceeds some ridiculous number, say 100 million or something like that, and treat it as a config error.

VincentSchmid:
## What?

It validates the VU values. If a VU is larger than the chosen constant, an error is appended.
This includes the startVU.

## Why?

The conclusion was reached to handle it that way during the discussions of this pull request.

## Checklist

## Related PR(s)/Issue(s)

Closes #1693

### DIFF
--- a/lib/executor/helpers.go
+++ b/lib/executor/helpers.go
@@ -76,8 +76,6 @@ func validateStages(stages []Stage) []error {
 			errors = append(errors, fmt.Errorf("stage %d doesn't have a target", stageNum))
 		} else if s.Target.Int64 < 0 {
 			errors = append(errors, fmt.Errorf("the target for stage %d can't be negative", stageNum))
-		} else if s.Target.Int64 > 100000000 {
-			errors = append(errors, fmt.Errorf("the target for stage %d is too large", stageNum))
 		}
 	}
 	return errors
--- a/lib/executor/helpers.go
+++ b/lib/executor/helpers.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"math"
 	"math/big"
 	"time"
 
@@ -21,6 +22,29 @@ func sumStagesDuration(stages []Stage) (result time.Duration) {
 		result += s.Duration.TimeDuration()
 	}
 	return
+}
+
+func sumTotalNumberOfUpAndDonwShifts(startVus int64, stages []Stage) int64 {
+	totalShifts := big.NewInt(startVus)
+	prevNumVus := totalShifts
+
+	for _, s := range stages {
+		current := big.NewInt(s.Target.Int64)
+		shiftDelta := new(big.Int).Sub(current, prevNumVus)
+		shiftDelta.Abs(shiftDelta)
+
+		totalShifts.Add(totalShifts, shiftDelta)
+
+		prevNumVus = current
+	}
+
+	totalShifts.Add(totalShifts, prevNumVus)
+
+	if totalShifts.IsInt64() {
+		return totalShifts.Int64()
+	}
+
+	return math.MaxInt64
 }
 
 func getStagesUnscaledMaxTarget(unscaledStartValue int64, stages []Stage) int64 {
@@ -52,6 +76,8 @@ func validateStages(stages []Stage) []error {
 			errors = append(errors, fmt.Errorf("stage %d doesn't have a target", stageNum))
 		} else if s.Target.Int64 < 0 {
 			errors = append(errors, fmt.Errorf("the target for stage %d can't be negative", stageNum))
+		} else if s.Target.Int64 > 100000000 {
+			errors = append(errors, fmt.Errorf("the target for stage %d is too large", stageNum))
 		}
 	}
 	return errors
--- a/lib/executor/helpers.go
+++ b/lib/executor/helpers.go
@@ -17,8 +17,8 @@ import (
 )
 
 const (
-	// this value is an arbitrary limit. It pervenets VU shifts from being too large.
-	maxTargetShift int = 100_000_000
+	// maxConcurrentVus is an arbitrary limit for sanity checks. It prevents running an exaggeratedly large number of concurrent VUs which may lead to an out-of-memory.
+	maxConcurrentVus int = 100_000_000
 )
 
 func sumStagesDuration(stages []Stage) (result time.Duration) {
@@ -38,26 +38,21 @@ func getStagesUnscaledMaxTarget(unscaledStartValue int64, stages []Stage) int64 
 	return max
 }
 
-// Validates target up and down shifts. For any shift that is larger than maxTargetShift, it will append an error.
+// validateTargetShifts validates the target up shifts. It will append an error for any shift that is larger than the maximum value allowed.
 // This includes the shift from 0 to the startTarget and the shift from the last stage to 0.
-// Targets can be Vus or iteration rates.
 // Each Stage needs a Target value. The stages array can be empty. The Targes could be negative.
-func validateTargetShifts(startTarget int64, stages []Stage) []error {
+func validateTargetShifts(startVus int64, stages []Stage) []error {
 	var errors []error
-	targets := []int64{0, startTarget}
 
-	for _, s := range stages {
-		targets = append(targets, s.Target.Int64)
-	}
-	targets = append(targets, 0)
+	if startVus > int64(maxConcurrentVus) {
+		errors = append(errors, fmt.Errorf(
+			"the startVUs exceed max limit of %d", maxConcurrentVus))
+	 }
 
-	for i := 0; i < len(targets)-1; i++ {
-		currentTarget := targets[i]
-		nextTarget := targets[i+1]
-
-		if absInt64(currentTarget-nextTarget) > int64(maxTargetShift) {
+	for i := 0; i < len(stages); i++ {
+		if stages[i].Target.Int64 > int64(maxConcurrentVus) {
 			errors = append(errors, fmt.Errorf(
-				"the target shifts from %d to %d is larger than the maximum allowed %d", currentTarget, nextTarget, maxTargetShift))
+				"target for stage %d exceeds max limit of %d", i+1, maxConcurrentVus))
 		}
 	}
 
--- a/lib/executor/helpers.go
+++ b/lib/executor/helpers.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"math"
 	"math/big"
 	"time"
 
@@ -24,7 +23,20 @@ func sumStagesDuration(stages []Stage) (result time.Duration) {
 	return
 }
 
-func sumTotalNumberOfUpAndDonwShifts(startVus int64, stages []Stage) int64 {
+func getStagesUnscaledMaxTarget(unscaledStartValue int64, stages []Stage) int64 {
+	max := unscaledStartValue
+	for _, s := range stages {
+		if s.Target.Int64 > max {
+			max = s.Target.Int64
+		}
+	}
+	return max
+}
+
+func validateNumberOfVuShifts(startVus int64, stages []Stage) []error {
+	const MaxRampingVUShift = 100000000
+	
+	var errors []error
 	totalShifts := big.NewInt(startVus)
 	prevNumVus := totalShifts
 
@@ -40,21 +52,11 @@ func sumTotalNumberOfUpAndDonwShifts(startVus int64, stages []Stage) int64 {
 
 	totalShifts.Add(totalShifts, prevNumVus)
 
-	if totalShifts.IsInt64() {
-		return totalShifts.Int64()
+	if totalShifts.Cmp(big.NewInt(MaxRampingVUShift)) > 0 {
+		errors = append(errors, fmt.Errorf("total number of VU shifts is too large"))
 	}
 
-	return math.MaxInt64
-}
-
-func getStagesUnscaledMaxTarget(unscaledStartValue int64, stages []Stage) int64 {
-	max := unscaledStartValue
-	for _, s := range stages {
-		if s.Target.Int64 > max {
-			max = s.Target.Int64
-		}
-	}
-	return max
+	return errors
 }
 
 // A helper function to avoid code duplication
--- a/lib/executor/helpers.go
+++ b/lib/executor/helpers.go
@@ -17,8 +17,8 @@ import (
 )
 
 const (
-	// maxConcurrentVus is an arbitrary limit for sanity checks. It prevents running an exaggeratedly large number of concurrent VUs which may lead to an out-of-memory.
-	maxConcurrentVus int = 100_000_000
+	// maxConcurrentVUs is an arbitrary limit for sanity checks. It prevents running an exaggeratedly large number of concurrent VUs which may lead to an out-of-memory.
+	maxConcurrentVUs int = 100_000_000
 )
 
 func sumStagesDuration(stages []Stage) (result time.Duration) {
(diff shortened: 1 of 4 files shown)


oleiade (approved):

lib/executor/ramping_vus_test.go
```diff
@@ -36,6 +37,26 @@ func TestRampingVUsConfigValidation(t *testing.T) {
 	errs = c.Validate()
 	require.NotEmpty(t, errs)
 	assert.Contains(t, errs[0].Error(), "greater than 0")
+
+	c.StartVUs = null.IntFrom(100000001)
```
oleiade, line 41: I find that this test already fulfills the job of multiple test cases by itself. Changing that is probably out of the scope of this specific PR, but would you mind revamping your addition to the test in a way that it acts as a dedicated test instead?

Something along the lines of

```golang
	t.Run("validate large VU shifts should fail", func(t *testing.T) {
	    t.Parallel()
	    
	    // ... your test code
	})
```

lib/executor/ramping_vus.go
```diff
@@ -86,6 +86,10 @@ func (vlvc RampingVUsConfig) Validate() []error {
 		errors = append(errors, fmt.Errorf("either startVUs or one of the stages' target values must be greater than 0"))
 	}
 
+	if sumTotalNumberOfUpAndDonwShifts(vlvc.StartVUs.Int64, vlvc.Stages) > 100000000 {
```
oleiade, line 89: Without any external context (when we look back at it 6 months from now), this value would feel somewhat arbitrary. The fact that the value is arbitrary is acceptable, but I would prefer if it were stored in a constant with an explicit name instead so that the code states what it does, and if I need to know what the value is or change it, I can consult a single place for that.

I would suggest some constant name along the lines of `MaxRampingVUShift` or something along those lines.

lib/executor/ramping_vus_test.go
```diff
@@ -36,6 +37,26 @@ func TestRampingVUsConfigValidation(t *testing.T) {
...
+	errs = c.Validate()
+	require.NotEmpty(t, errs)
+	assert.Contains(t, errs[0].Error(), "number of VU shifts is too large")
+
+	c.StartVUs = null.IntFrom(60000000)
+	c.Stages = []Stage{
+		{Target: null.IntFrom(50000000), Duration: types.NullDurationFrom(12 * time.Second)},
+		{Target: null.IntFrom(50000000), Duration: types.NullDurationFrom(12 * time.Second)},
+	}
```
oleiade, line 50: As I reviewed the test, I wondered, "why those specific values?". Thus, I would suggest to either rely on the constant I suggested before, in a way that makes the reason why we pick a number explicit: `MaxRampingVUShift + x`, or to at least add a comment explaining why we use that value 🙇🏻

oleiade (changes requested):

codebien (changes requested):
Hey @VincentSchmid,
thanks for your contribution :tada: 

The current approach is working but I would prefer if we use a different one. We should have a validating function instead of one returning the sum. In this way, we can directly return an error from the function. It would be similar to the validateStages function that already exists.

Also, note, that the same logic has to be applied to the ramping arrival rate executor.

lib/executor/helpers.go
```diff
@@ -33,6 +33,32 @@ func getStagesUnscaledMaxTarget(unscaledStartValue int64, stages []Stage) int64
 	return max
 }
 
+func validateNumberOfVuShifts(startVus int64, stages []Stage) []error {
```
oleiade, line 36: Could we add documentation for this function please, especially regarding what the arguments to the function do, and what their constraints are 🙇🏻

lib/executor/helpers.go
```diff
@@ -33,6 +33,32 @@ func getStagesUnscaledMaxTarget(unscaledStartValue int64, stages []Stage) int64
...
+	const MaxRampingVUShift = 100000000
```
oleiade, line 37: Is there a specific reason for having this constant defined in the function body? Otherwise, I would suggest moving this constant out of the function scope and probably keeping it unexported (name starting with a lowercase letter). What do you think? Opinions on this @codebien ?

lib/executor/ramping_vus_test.go
```diff
@@ -36,6 +37,34 @@ func TestRampingVUsConfigValidation(t *testing.T) {
...
+		const HalfMaxInt = math.MaxInt64 / 2
+
+		c.StartVUs = null.IntFrom(MaxRampingVUShift)
+		errs = c.Validate()
+		require.NotEmpty(t, errs)
+		assert.Contains(t, errs[0].Error(), "number of VU shifts is too large")
+
+		c.StartVUs = null.IntFrom(0)
+		c.Stages = []Stage{
+			{Target: null.IntFrom(HalfMaxRampingVUShift + 1), Duration: types.NullDurationFrom(12 * time.Second)},
```
oleiade, line 55: I'm pretty sure I misunderstood or missed something, but I'm a bit confused as to why this specific test case should fail, could you please clarify? 🙇🏻

lib/executor/helpers.go
```diff
@@ -33,6 +33,32 @@ func getStagesUnscaledMaxTarget(unscaledStartValue int64, stages []Stage) int64
...
+		current := big.NewInt(s.Target.Int64)
+		vuDelta := new(big.Int).Sub(current, prevNumVus)
+		vuDelta.Abs(vuDelta)
+
+		totalShifts.Add(totalShifts, vuDelta)
+
+		prevNumVus = current
+	}
+
+	totalShifts.Add(totalShifts, prevNumVus)
```
oleiade, line 53: I'm not sure I grasp why we're adding to `totalShifts` again here. Could you clarify, and if relevant, could you add a comment detailing why we do it?

lib/executor/ramping_vus_test.go
```diff
@@ -36,6 +37,34 @@ func TestRampingVUsConfigValidation(t *testing.T) {
...
+		}
+		errs = c.Validate()
+		assert.Contains(t, errs[0].Error(), "number of VU shifts is too large")
+
+		// Integer overflow is correctly handled
+		c.StartVUs = null.IntFrom(0)
+		c.Stages = []Stage{
+			{Target: null.IntFrom(HalfMaxInt + 1), Duration: types.NullDurationFrom(12 * time.Second)},
```
oleiade, line 63: Same comment as before, I was a bit confused and surprised by this test. Could you clarify why this is expected to fail (probably related to my other `totalShifts` comment), and add relevant comments to simplify the maintenance of it?

lib/executor/ramping_vus_test.go
```diff
@@ -36,6 +37,34 @@ func TestRampingVUsConfigValidation(t *testing.T) {
 	errs = c.Validate()
 	require.NotEmpty(t, errs)
 	assert.Contains(t, errs[0].Error(), "greater than 0")
+
+	t.Run("validate large VU shifts should fail", func(t *testing.T) {
```
oleiade, line 41: I would prefer if we would split this test case in three different smaller ones, with comments detailing what and why we're testing explicitly 🙏🏻

oleiade, line 41: I would also suggest that we add test(s) for the happy scenarios 🙇🏻

oleiade (changes requested):

lib/executor/ramping_vus_test.go
codebien: Yes, I also need to get this clarified. Why is it expected to fail?

codebien (commented):
Now, checking this again it doesn't sound like the logic for the validation based on the sum is required.

I may miss some obvious reason, but can you clarify why is it not enough to just validate the `maxTarget` returned by [this code](https://github.com/grafana/k6/blob/65b2f27d1a0612846467e48418f469984724e007/lib/executor/ramping_vus.go#L85-L87)?

lib/executor/ramping_vus_test.go
VincentSchmid: I think I might have misunderstood what the problem is we're addressing.

The issue for this MR mentioned: _total number of VU up and down shifts_. So I understood, a 'VU Shift' occurs each time there's a change in the number of VUs, either an increase or a decrease. So that is why I assumed maxTarget would not be enough.

For instance, consider a test scenario that starts at 0 VUs, increases to 20, drops to 5, rises to 10, and finally returns to 0. Here's how I would count the VU shifts in this scenario:

Increase from 0 to 20: 20 VU shift up
Decrease from 20 to 5: 15 VU shift down
Increase from 5 to 10: 5 VU shift up
Decrease from 10 to 0: 10 VU shift down

Sum Total VU Shifts I count: 20 (up) + 15 (down) + 5 (up) + 10 (down) = 50 shifts.
In the above example, the maxTarget is 20.

This is also why the test fails. Even in the most basic example (one VU value) The number of VU Shifts I calculate will be 2x maxTarget (since VUs start at 0 and return to 0, up- and downshift are equal to maxTarget). So in the above example that would be `(HalfMaxInt + 1) * 2` which would overflow.

The question is, whether the memory issue arises because of the number of VU shifts or because of the number of VUs at one time (or if maxTarget is a good enough approximation).

codebien: As you can see running this test https://github.com/grafana/k6/blob/master/examples/stages.js, `target` is an absolute number, not a relative increase/decrease.

Sorry if the issue description is creating confusion. And, let me know if you think I'm missing something.

oleiade: @VincentSchmid, thanks for your detailed explanation. To be fair, the issue itself doesn't do a good enough job at defining the problem, in my opinion, and it's easy to misinterpret it.

My understanding from your writeup and the code is that in this PR, you went with a cumulative approach to what is referred to as VU shifts. 

My own understanding of the problem differs, as I understood the initial issue as validating that between one stage to another, we do not have a shift that's too high "locally". As in: "the target value difference between `stage[n]` and `stage[n+1]` in absolute value does not exceed a constant value set ahead of time:
```
abs(stage[n+1] - stage[n]) >= maxRampingVUTargetShif
```

Do you share the same understanding @codebien ?

oleiade: @codebien I agree that there might be a misunderstanding regarding our approach. To clarify:
* **Current Proposed Solution understanding**: My understanding is that the current solution aims to guard against the scenario where the total sum of the differences between consecutive stages is less than a certain predefined value. 
In other words, we're calculating the differences between each pair of stages (stage[1] and stage[0], stage[2] and stage[1], etc.) and ensuring the sum of these differences remains below a predefined threshold. Somewhat formally, it looks like this: `sum(difference(stage[1], stage[0]), difference(stage[2], stage[1]), ...) < predefinedValue`.
* **Intended Approach**: However, my understanding of what we initially had in mind was slightly different. We intended to protect against the situation where the difference between any two consecutive stages in the definition of options is greater than a predefined value. 
This means checking if any pair of consecutive stages has a difference exceeding our predefined threshold. This can be represented as: `any([difference(stage[1], stage[0]), difference(stage[2], stage[1]), ...], function d: d > predefinedValue)`.

I apologize for the confusion caused by my use of pseudoalgorithmic notation. I hope this explanation is clearer.


Regarding the solution space, and how to concretely move forward, I think the **intended approach** I described above would be the relevant way to move forward.

Regarding how this would behave in a context where k6 is running distributed, I don't know on top of my head either how this would behave. My assumption is that when using the `--execution-segment` argument, the actual stage target VU would be divided by the value passed to the flag. But we should verify this indeed 👍🏻

oleiade: I did a quick verification with the following script:
```javascript
import http from "k6/http";
import { sleep } from "k6";

export const options = {
	discardResponseBodies: true,
	scenarios: {
		contacts: {
			executor: "ramping-vus",
			startVUs: 0,
			stages: [
				{ duration: "20s", target: 10 },
				{ duration: "10s", target: 0 },
				{ duration: "10s", target: 100000000 },
			],
			gracefulRampDown: "0s",
		},
	},
};

export default function () {
	http.get("https://test.k6.io/contacts.php");
	sleep(0.5);
}
```

Ran with the following command: `k6 run --execution-segment 3%`

Debugger breakpoints in the `func (vlvc RampingVUsConfig) Validate() error` function and the `validateStages` helper, that regardless of the `execution-segment`, the stages that are being validated are the ones received from the options: 10, 0, 100000000 (as opposed to fractioned ones as I assumed (~3, 0, 300000)).

VincentSchmid: Ok, I understand how the intended approach differs from mine. 

What we want to validate for is that the maximum difference in VUs between any two stages (including down shifts) does not exceed our predefined constant. 

Additionally, I'll account for the initial and final stages, that increase from 0 and a decrease back to 0.

I don't yet completely understand the execution-segment part, but I will look into it and make sure I make that work as well, or reach out for clarification.

lib/executor/ramping_arrival_rate.go
```diff
@@ -96,6 +96,7 @@ func (varc *RampingArrivalRateConfig) Validate() []error {
 	}
 
 	errors = append(errors, validateStages(varc.Stages)...)
+	errors = append(errors, validateVUShifts(varc.StartRate.Int64, varc.Stages)...)
```
oleiade, line 99: I've noticed that we call `validateVUShifts` here, but in practice, the `Stages` we receive here, in the context of the ramping arrival rate are "iterations per `timeUnit`" instead of VUs.

Thus, I would suggest  we consider a more fitting name for that function? Maybe `validateTargetShifts` considering in both cases (ramping vus, and ramping arrival rate, the parameter the users rely on is called "target").

What do you folks think?

lib/executor/helpers.go
```diff
@@ -33,6 +38,32 @@ func getStagesUnscaledMaxTarget(unscaledStartValue int64, stages []Stage) int64
 	return max
 }
 
+// Validates all the VU up and down shifts. For any shift that is larger than maxVUShift it will append an error.
```
oleiade, line 41: As mentioned in another comment, we're not quite checking the VUs shift, but also iteration rate in other contexts. Once we have established a better name, we should probably rephrase this comment to reflect the change too 🙇‍♂️

oleiade (changes requested):
Great job @VincentSchmid 👏🏻 🙇🏻 

The current implementation matches my initial assumptions and expectations. I've left a bunch of comments more on the minor side that I think should still be addressed before we merge this, but besides that, it starts looking good 🎉 

I couldn't help but notice that although we use the mechanism in the context of the ramping arrival rate executor, we haven't added tests for it (which might have led to the issue of naming somehow?). I would prefer to see tests for that usage too 🙇🏻

oleiade (changes requested):
Great job @VincentSchmid 👏🏻 🙇🏻 

The current implementation matches my initial assumptions and expectations. I've left a bunch of comments more on the minor side that I think should still be addressed before we merge this, but besides that, it starts looking good 🎉 

I couldn't help but notice that although we use the mechanism in the context of the ramping arrival rate executor, we haven't added tests for it (which might have led to the issue of naming somehow?). I would prefer to see tests for that usage too 🙇🏻

oleiade:
As I'm at it, in case that proves useful to any of you folks, here's the script I used to try the feature locally:

```javascript
import http from 'k6/http';

export const options = {
	scenarios: {
		vus: {
			executor: 'ramping-vus',
			startVUs: 0,
			stages: [
				{duration: '20s', target: 10},
				{duration: '10s', target: 100000100},
			],
		},
		rate: {
			startRate: 0,
			preAllocatedVUs: 50,
			stages: [
				{duration: '20s', target: 10},
				{duration: '10s', target: 100000100},
			]
		}
	}
}

export default function () {
	http.get('http://test.k6.io')
}
```

Which made me think that it could also be worth adding end 2 end tests demonstrating the behavior?

lib/executor/ramping_arrival_rate.go
```diff
@@ -96,6 +96,7 @@ func (varc *RampingArrivalRateConfig) Validate() []error {
 	}
 
 	errors = append(errors, validateStages(varc.Stages)...)
+	errors = append(errors, validateTargetShifts(varc.StartRate.Int64, varc.Stages)...)
```
codebien, line 99: Thought better on it and now I don't think we need it. The target in this case is related to iterations and not to VUs so having a high number here will not guarantee an insane number of VUs. In any case, for sure, the number of max iterations is in order of magnitude higher than the max allowed for VUs. Each VU can handle a rate of thousands of iterations.

At this point, it makes sense to just drop it from here. Having a static-constant value doesn't seem to be the right approach.
I'm sorry, I know I requested it in the previous review.

Instead, all the other executors that directly set the VUs number may also benefit from the same validation (shared-iterations, constant-vus, per-vu-iterations). But it is out-of-scope for this pull request, we will open a dedicated issue for them.

codebien (approved):

oleiade (approved):

VincentSchmid:
Thank you guys for your patience, I had fun doing this with you and I learned some things!

codebien:
Hey @VincentSchmid, thank you so much for your effort here. :tada:

PR #3492, opened 2023-12-06
Revert "Pin golang version to 1.21.5 in github actions and docker"

mstoykov:
## What?

This reverts commit 3808747f1629937010629adb212ccc9fcaa16918.

## Why?

We need that to release v0.48.0 with the latest golang version. But we really don't want to keep having to bump this every time so it is better to just use the wildcard version.

## Checklist


codebien:
In my experience, it should be enough if we force the cache once then the next build should pick up the right version. We have `workflow_dispatch` for the build workflow so we might manually trigger a new job to force a build with a newly released Go version.

codebien (approved):

PR #3505, opened 2023-12-15
feat: Integrating the web dashboard feature

szkiba:
As per our product DNA, we aim, by integrating the web dashboard (xk6-dashboard) directly into k6, to enhance and streamline the user experience, making real-time and end-of-test visualizations immediately accessible as part of the k6 command-line experience.

This is not yet the final version of the integration. The goal is to integrate xk6-dashboard v0.7.0, but that release does not exist yet (waiting for dependencies).

I created the PR to ask for a review and to get feedback on whether the integration would be good this way, or if the team had something else in mind.


go.mod
```diff
@@ -14,6 +14,7 @@ require (
 	github.com/golang/protobuf v1.5.3
 	github.com/gorilla/websocket v1.5.1
 	github.com/grafana/xk6-browser v1.2.1
+	github.com/grafana/xk6-dashboard v0.6.1
```
mstoykov, line 17: You here use the latest tagged version, but you also dropped a bunch of dependancies and other changes after that.

Can we please either:
1. tag a new release and use that
2. just use the commit for now

szkiba: as I wrote
> This is not yet the final version of the integration. The goal is to integrate xk6-dashboard v0.7.0, but that release does not exist yet (waiting for dependencies).

from integration point of view there is no difference (expect dependencies), so this is why I just used latest tagge version.... I'll update to v0.7.0 when it will ready

cmd/ui.go
```diff
@@ -380,3 +385,13 @@ func yamlPrint(w io.Writer, v interface{}) error {
...
+
+// checkWebDashboardDescription returns true if desc contains web dashboard description.
+// The returned string contains info string (URL).
+func checkWebDashboardDescription(desc string) (bool, string) {
+	const webDashboardDescPrefix = webDashboardName + " "
+	if strings.HasPrefix(desc, webDashboardDescPrefix) {
+		return true, strings.TrimPrefix(desc, webDashboardDescPrefix)
+	}
+	return false, ""
+}
```
codebien, line 397: As the extension is going to be included in the core, can we just not include the name in it so we can simplify the code here? IIRC, the description is not used somewhere else.

cmd/ui.go
```diff
@@ -114,18 +114,23 @@ func printExecutionDescription(
...
 			desc := out.Description()
 			if desc == engine.IngesterDescription {
-				if len(outputs) != 1 {
-					continue
-				}
-				desc = "-"
+				continue
+			}
+			if ok, v := checkWebDashboardDescription(desc); ok {
+				fmt.Fprintf(buf, "web dashboard: %s\n", valueColor.Sprint(v))
```
codebien, line 120: It does print the following string on my side.
```
web dashboard: (:5665) http://127.0.0.1:5665
```
Do we need to repeat the URL twice?

On my side, it also prints the line without any padding from the left border of the terminal which doesn't seem optimal.

codebien (changes requested):
Hey @szkiba,
thanks for your contribution. Congrats for the progress with the dashboard :clap: 

I've left some comments.

cmd/ui.go
szkiba: yes, this appears for me too. In xk6-dashboard v0.7.0, there will no longer be redundant information, only the URL.
Otherwise, I increased the indentation by 2 characters (the "web dashboard" string is so much longer than the previous field size). Please let me know if I should increase the indentation by one more character for the other fields as well?
​

cmd/outputs.go
```diff
@@ -16,9 +16,12 @@ import (
 	"go.k6.io/k6/output/json"
 	"go.k6.io/k6/output/statsd"
 
+	"github.com/grafana/xk6-dashboard/dashboard"
 	"github.com/grafana/xk6-output-prometheus-remote/pkg/remotewrite"
 )
 
+const webDashboardName = "web-dashboard"
+
```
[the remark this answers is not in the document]
szkiba: The string "web-dashboard" occurs at least twice (or three times) in the code. In such cases, I usually define it as a constant. I understand the consistency aspect, please let me know if this is the stronger point or not to include the same string more than once in the code, and then I will eliminate the constant definition.

cmd/outputs.go
```diff
@@ -45,6 +48,7 @@ func getAllOutputConstructors() (map[string]output.Constructor, error) {
 		"experimental-prometheus-rw": func(params output.Params) (output.Output, error) {
 			return remotewrite.New(params)
 		},
+		webDashboardName: dashboard.New,
```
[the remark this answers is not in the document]
szkiba: The string "web-dashboard" occurs at least twice (or three times) in the code. In such cases, I usually define it as a constant. I understand the consistency aspect, please let me know if this is the stronger point or not to include the same string more than once in the code, and then I will eliminate the constant definition.

cmd/outputs.go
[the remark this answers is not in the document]
szkiba: see bellow

cmd/config.go
```diff
@@ -67,6 +69,9 @@ func (c Config) Apply(cfg Config) Config {
 	if cfg.NoUsageReport.Valid {
 		c.NoUsageReport = cfg.NoUsageReport
 	}
+	if cfg.NoWebDashboard.Valid {
+		c.NoWebDashboard = cfg.NoWebDashboard
+	}
```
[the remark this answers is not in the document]
szkiba: I followed the "usage report" example here, so I aligned the name of the NoWebDashboard property with the name of the NoUsageReport property.
Let me know if avoiding the "no" pattern or consistency with the "usage report" is more important and then I'll change it.

cmd/ui.go
szkiba: I followed the `desc == engine.IngesterDescription` example here. I agree that the use of description is unfortunate, the name would be better. By the way, the name is not hard-coded into the xk6-dashboard, it is named "web-dashboard" in the k6 source code. Its original name is "dashboard" in the xk6-dashboard source code. Thus, the name included in the extension is not included in the core source code, but the name defined in the core source code is used in several places in the core source code.
Let me know if it would be good if I check the name here instead of the description.

szkiba:
> Hey @szkiba, thanks for your contribution. Congrats for the progress with the dashboard 👏
> 
> I've left some comments.

Hi @codebien , thanks for the feedback!

go.mod
szkiba: Done, I created a pre-release.

cmd/ui.go
szkiba: I simplified and eliminated the check in a separate function

cmd/outputs.go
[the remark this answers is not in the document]
szkiba: Done

cmd/ui.go
codebien: The blue line is where the terminal's tile finishes.

cmd/config.go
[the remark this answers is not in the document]
codebien: Sorry, this is my fault. I thought it was lib/options.go, instead, it is cmd/config.go. It makes sense to me to have the `no-` version. Can you rollback, please?

cmd/ui.go
codebien: >By the way, the name is not hard-coded into the xk6-dashboard, it is named "web-dashboard" in the k6 source code. Its original name is "dashboard" in the xk6-dashboard source code.

It sounds like we should align them. Otherwise, we will have this difference, and in the long term it could create issues. I think could be a good idea to define an exported variable directly in the dashboard package as `const dashboard.OutputName = "web dashboard"`. For example, here https://github.com/grafana/xk6-dashboard/blob/6a4cbea58a4fd2a5984c66f6893411e3736c1eae/register.go#L18.

In this way, we are sure both use the same name consistently.

cmd/ui.go
szkiba: yes. I can indent each lines with one more space, but I'm worry about indent became too deep and will cause line breaks on longer lines. So, should I add one more space?

szkiba: or can we reconsider using "web dashboard" name? I'd prefer simply "dashboard" (yes it is general, but probably we will have not other dashboard in near future)

codebien: >but I'm worry about indent became too deep and will cause line breaks on longer lines

The metrics summary is already longer so I don't think it should be a problem having them a bit more indented. Do you have any specific line you're thinking it could break?

>or can we reconsider using "web dashboard" name? 

As we are part of a company that builds tools for graphing data, everything is a dashboard, so I prefer `web dashboard` as it is a bit clearer. We already have dashboards for other outputs, and they are all Grafana-based so during communications, verbal or written, having `web` prefix should clarify the context. In any case, if you want to re-discuss it open the question to the working group.

szkiba: >  Do you have any specific line you're thinking it could break?
No
I added one more space, so "web dashboard" no indented by one space from left.

cmd/config.go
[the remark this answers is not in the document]
codebien: Ok, I didn't know about the change, then LGTM.

cmd/ui.go
szkiba: I would prefer this constant to remain in the k6 repo. The xk6-dashboard output extension defines output as "dashboard". For the xk6-dashboard built into k6, k6 defines an output as "internal-web-dashboard" (I renamed it), and we defined this name here. That is, during the integration into k6, we define a name for the output that we only use here. It is logical that this constant should be in the k6 repo.

.gitignore
```diff
@@ -28,6 +28,7 @@ script.js
 /vendor/**/*.y*ml
 /vendor/**/.*.y*ml
 /vendor/github.com/dlclark/regexp2/testoutput1
+!/vendor/github.com/grafana/xk6-dashboard/dashboard/assets/packages/config/dist/config.json
```
codebien, line 31: Should we move it close to the rule it is overwriting? It would be also helpful to have a comment about why it is relevant.

cmd/ui.go
codebien: >I would prefer this constant to remain in the k6 repo.

Fine for me.

>The xk6-dashboard output extension defines output as "dashboard". For the xk6-dashboard built into k6, k6 defines an output as "internal-web-dashboard"

I still think this is an error and will generate confusion. @oleiade @mstoykov it would be good to get your opinion here.

codebien (changes requested):

.gitignore
szkiba: done

cmd/ui.go
szkiba: done, xk6-dashboard now use same name, "web-dashboard" and exports a constant for it

cmd/outputs.go
```diff
@@ -92,9 +94,15 @@ func createOutputs(
...
 
-	for _, outputFullArg := range test.derivedConfig.Out {
+	outputs := test.derivedConfig.Out
+	if test.derivedConfig.WebDashboard.Bool {
+		outputs = append(outputs, dashboard.OutputName)
+	}
+
+	result := make([]output.Output, 0, len(outputs))
+
+	for _, outputFullArg := range outputs {
```
mstoykov, line 105: Looking at this and the changes to the way we format stuff ... I wonder ... why?

Like why do you want ot have an additional `--web-dashboard` cli flag when `-o dashboard` ... works.

Why do you want to move it on separate line when outputs already have a way to show a url for users to use as in 
image
^ arguably the above can be improved, but I see no reason to move it in a different one.

And this will practically limit the integration to the above "registration" of the output.

szkiba: > Like why do you want ot have an additional `--web-dashboard` cli flag when `-o dashboard` ... works.

Because this is what is written in the design document, if you want to re-discuss it open the question to the working group. It's perfectly fine for me if it has to be activated with the -o flag.

oleiade: As @szkiba mentioned, this was a UX/product decision made by the workgroup that is reflected in the design document. 

Part of the reasoning behind that decision was a desire to have the dashboard become a first-class citizen of the k6 UX down the road, the fact that it acts as an output under the hood being judged an implementation detail.

If there is a strong **technical** argument to be made to keep exposing/displaying it as an output as part of our command-line UI, we'd be happy to discuss it at the beginning of next year when everyone is back 🤝

oleiade (commented):
LGTM 🚀 

Will wait until the rest of the environment variables are implemented, and we resolve some of the discussions we're having internally to approve. But looks good to me, great job @szkiba 👏🏻

cmd/outputs.go
mstoykov: > first-class citizen of the k6 UX 

Outputs are first class citizen.

> down the road

Kind of implies to me this shouldn't be in the initial release.

The technical and none technical argument is the same - users already have the concept of output. Adding a thing that is clearly and output and then treating it differently isn't cohesive. 

At best people will be confused why there are two ways to the same thing at worse they will think they are two different things. And anyone who works on it will need to wonder the same thing.

On connected note - yes the current summary fits this same bill, except that it predates most of that, and if possible I will certainly try to push it towards being an output through and through.

But adding more special cases IMO doesn't help anyone.

szkiba:
> LGTM 🚀
> 
> Will wait until the rest of the environment variables are implemented, and we resolve some of the discussions we're having internally to approve. But looks good to me, great job @szkiba 👏🏻

Thank you @oleiade 

What environmental variables are you referring to?
Here is the list of environment variables managed by xk6-dashboard:

- `K6_WEB_DASHBOARD_HOST`
- `K6_WEB_DASHBOARD_PORT`
- `K6_WEB_DASHBOARD_PERIOD`
- `K6_WEB_DASHBOARD_OPEN`
- `K6_WEB_DASHBOARD_EXPORT`
- `K6_WEB_DASHBOARD_RECORD`
- `K6_WEB_DASHBOARD_TAG`

see: https://github.com/grafana/xk6-dashboard#environment

In addition, the `K6_WEB_DASHBOARD` variable can be set to true to activate the dashboard in k6.
Correct me if I'm wrong, but I think all environment variables in the design document are implemented....

see: https://github.com/grafana/k6/blob/feat/3504-dashboard-integration/cmd/config.go#L45

.gitignore
```diff
@@ -29,5 +31,6 @@ script.js
 /vendor/**/.*.y*ml
 /vendor/github.com/dlclark/regexp2/testoutput1
 
+
```
codebien, line 34:
```suggestion
```

go.mod
```diff
@@ -78,11 +79,13 @@ require (
 	github.com/inconshreveable/mousetrap v1.0.0 // indirect
 	github.com/josharian/intern v1.0.0 // indirect
 	github.com/matttproud/golang_protobuf_extensions v1.0.4 // indirect
+	github.com/pkg/browser v0.0.0-20210911075715-681adbf594b8 // indirect
 	github.com/pmezard/go-difflib v1.0.0 // indirect
 	github.com/prometheus/client_golang v1.16.0 // indirect
 	github.com/prometheus/client_model v0.4.0 // indirect
 	github.com/prometheus/common v0.42.0 // indirect
 	github.com/prometheus/procfs v0.10.1 // indirect
+	github.com/r3labs/sse/v2 v2.10.0 // indirect
```
codebien, line 88: @mstoykov I remember you asked about it, can you point me there, please?

codebien (changes requested):

cmd/outputs.go
oleiade: I see your point 👍🏻 I personally don't have a strong opinion on this.

Will add it to the agenda of our internal discussions when everyone is back from 🎄 holidays, and I'll invite you to it so we can decide on a way forward that makes sense for everyone 🙇🏻

go.mod
mstoykov: I guess you mean https://github.com/grafana/xk6-dashboard/issues/108#issuecomment-1855397461 ?

codebien: @szkiba It would be good to have a quick sketch about an overview of the extension's architecture and add it to the `docs` folder in the `xk6-dashboard`. So we can quickly realize what are the connections in place and the flow of data.

szkiba: @codebien  FYI https://github.com/grafana/xk6-dashboard#how-it-works
Is it what you mean? Or something more detailed?

cmd/outputs.go
oleiade: We've discussed this internally, and our preference goes with proceeding with the current UX provided by the PR:
* The feature needs to be activated through an environment variable
* when activated, it has a dedicated line in the output to point users to the URL
* the `--output` argument is supported for consistency and to align with a larger idea to mark outputs (including the end-of-test summary), more explicitly as such in the future.

```
K6_WEB_DASHBOARD=true ./k6 run ../extensions/xk6-dashboard/script.js

          /\      |‾‾| /‾‾/   /‾‾/
     /\  /  \     |  |/  /   /  /

     execution: local
        script: ../extensions/xk6-dashboard/script.js
 web dashboard: http://127.0.0.1:5665
        output: -

     scenarios: (100.00%) 2 scenarios, 9 max VUs, 12m30s max duration (incl. graceful stop):
              * camel: Up to 5 looping VUs for 12m0s over 6 stages (gracefulRampDown: 0s, gracefulStop: 30s)
              * snake: Up to 4 looping VUs for 12m0s over 12 stages (gracefulRampDown: 0s, gracefulStop: 30s)
```

olegbespalov (dismissed):

cmd/config.go
```diff
@@ -31,6 +31,9 @@ func configFlagSet() *pflag.FlagSet {
 	flags.StringArrayP("out", "o", []string{}, "`uri` for an external metrics database")
 	flags.BoolP("linger", "l", false, "keep the API server alive past test end")
 	flags.Bool("no-usage-report", false, "don't send anonymous stats to the developers")
+	flags.Bool("web-dashboard", false, "enable web dashboard")
+	// we don't want to introduce it, in future releases logic may change
+	flags.MarkHidden("web-dashboard") //nolint:errcheck,gosec
```
codebien, line 36:
```suggestion
```
What's the point here? If we don't want to introduce it then postpone the addition in a dedicated PR in the future.

go.mod
codebien: Yes, it's good enough thanks! :tada: Maybe, just as a suggestion it would be good to split the Dashboard readme and have in the main page just a short version with the index as we do here in the k6 repository.

codebien (changes requested):

go.mod
szkiba: Yes, you areright, the readme a bit long now. I'm going to reorganize it, but it will be independent from the release (I'm not planning a release after it)

codebien: >but it will be independent from the release
Yes, definitely it shouldn't be a blocker for this PR.

cmd/config.go
szkiba: Done, removed.

cmd/tests/cmd_run_test.go
```diff
@@ -1865,6 +1865,50 @@ func TestUIRenderOutput(t *testing.T) {
 	}
 }
 
+func TestUIRenderWebDashboard(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		flag      string
```
codebien, line 1872:
```suggestion
```
It shouldn't work anymore. Or am I missing something?

codebien (changes requested):

cmd/tests/cmd_run_test.go
szkiba: You are right, sorry. Fixed now.

go.mod
szkiba: >  Maybe, just as a suggestion it would be good to split the Dashboard readme

I made the readme a bit more readable...

codebien (commented):
The current code looks good to me.

But, unfortunately, there is something in the UX that sounds unexpected.
If I send an interrupt while k6 is running and the dashboard is opened in the browser then k6 gets stuck.

@szkiba is it expected?

szkiba:
> The current code looks good to me.
> 
> But, unfortunately, there is something in the UX that sounds unexpected. If I send an interrupt (CTRL+c) while k6 is running and the dashboard is opened in the browser then k6 gets stuck.
> 
> @szkiba is it expected?
> 
> I tested running the following command:
> 
> ```go
> K6_WEB_DASHBOARD=true go run . run -d 1m ./examples/http_get.js
> ```

@codebien  Actually, this is a feature (as long as the browser is open, k6 does not stop after the test run), but it really does not handle the abort situation well. I will fix it and make a patch release soon.

oleiade (approved):
🚀

olegbespalov (approved):

Issue #238, opened 2017-06-06
Support for QUIC protocol


neilstuartcraig:
Hi

As discussed briefly in Slack, it'd be great to get support for the [QUIC](https://www.chromium.org/quic) transport protocol (currently an [IETF draft](https://tools.ietf.org/html/draft-hamilton-early-deployment-quic-00) though has a dedicated working group and very wide support so is extremely likely to be ratified before too long): 

There's a potentially helpful QUIC lib for Go which is used by [Caddy](https://caddyserver.com/) - [quic-go](https://github.com/lucas-clemente/quic-go) (MIT license).

QUIC is UDP-based so this might cause some issues/complexity and is worth calling out in docs as folks will need to ensure they allow udp:443 from load gen(s) to server(s)/LBs.

Please let me know if you need any more info from me.

Cheers

liclac:
Okay, some super newbie questions from someone with only a surface-level understanding of the protocol:

* How exactly does dialing QUIC work? Does the client need to know that a server supports it, or is it a TLS handshake over TCP + a next-protocol marker (similar to HTTP/2)?
* Are the semantics different from HTTP/HTTP2 over TCP, or are there any gotchas that may not be immediately obvious?

What this basically adds up to is: can we add transparent QUIC support to `k6/http`, or would we need a separate `k6/quic` module?

neilstuartcraig:
Sorry it took so long to get back rto you @liclac. i'm not massively familiar with the internals of quic (though i know a couple of people who are if we need more info) but:

> How exactly does dialing QUIC work? Does the client need to know that a server supports it, or is it a TLS handshake over TCP + a next-protocol marker (similar to HTTP/2)?

QUIC works via an alt-svc advertisement (HTTP response header). I am just tinkering with a new website for myself at the moment so you can check that out if you want to see it in action: https://caddy.thedotproduct.org
You'll see the behaviour there, probably easiest to use Chrome/Canary or you can enable QUIC for all sites in Chrome/Stable or Beta via chrome://flags#enable-quic -> "enabled"

[...]

PR #3507, opened 2023-12-16
HTTP/3 support for k6

bandorko:
## What?

HTTP/3 support for k6

## Why?

This is a proposal for implementing HTTP/3 support for K6. I know that it is incomplete (e.g.: there are no test), but I'm really interested in your thoughts on this approach. It uses a ConnectionTracker to collect the metrics about the communication.
I also created it as an extension (https://github.com/bandorko/xk6-http3), but because its code is mostly copied from the k6 source, I thought that HTTP/3 implementation would have a better place in the k6 repository.

## Checklist

## Related PR(s)/Issue(s)
https://github.com/grafana/k6/issues/238


lib/netext/httpext/http3.go
```diff
@@ -0,0 +1,82 @@
...
+	// HTTP3ReqWaitingName is the name of the HTTP3ReqWaiting metric.
+	HTTP3ReqWaitingName = "http3_req_waiting"
+	// HTTP3ReqReceivingName is the name of the HTTP3ReqReceiving metric.
+	HTTP3ReqReceivingName = "http3_req_receiving"
+	// HTTP3ReqTLSHandshakingName is the name of the HTTP3ReqTLSHandshaking metric.
+	HTTP3ReqTLSHandshakingName = "http3_req_tls_handshaking"
+	// HTTP3ReqConnectingName is the name of the HTTP3ReqConnecting metric.
+	HTTP3ReqConnectingName = "http3_req_connecting"
+	// HTTP3ReqsName is the name of the HTTP3Reqs metric.
+	HTTP3ReqsName = "http3_reqs"
```
codebien, line 43: I don't think we need to define new metrics. We could reuse the already defined for the current HTTP module, of course, the proto tag should be different.

codebien (changes requested):
Hey @bandorko,
thanks for your contribution :tada: I appreciate the fact you want to discuss with us the vision of the new API.

As we always suggest and similar to the process we use for building built-in modules, the process for including a new API is to have an extension where we start gathering feedback and usage--you already did it :handshake: 
When it gets enough traction to justify being included in the core then we start the process.

However, regarding the HTTP proposal. We already have internally the plan to create a new HTTP module, and we added a new design doc explaining our vision around it. So, it sounds like a better approach if an eventual new protocol is implemented following the design we described there.

Regarding the HTTP3 version, the Go team is working on supporting native in Go the 3rd version of the HTTP protocol. Considering it, I think it makes sense to wait for this release before including it directly into the core. In the meantime, the Core team and the community can continue to experiment with other libraries in the extension.

oleiade:
I fully align with @codebien points 🙇🏻

bandorko:
@codebien, @oleiade: Thank you for the feedback. Then I close this PR and fix the metric names in the extension until http3 will be supported in go and k6.
Do you have a link for the plan of the new HTTP module, or it is still internal?

codebien:
@bandorko you can find the design doc [here](https://github.com/grafana/k6/blob/master/docs/design/018-new-http-api.md). If you would like to align and you need support, feel free to open an issue directly on the extension and ping me.

PR #3512, opened 2023-12-19
Fix/3510 questionable gitignore entries

### LINKED ISSUE #3510: Questionable .gitignore entries
During the process of vendoring the xk6-dashboard, I ran into an error, which, as it turned out, was caused by some entries in the `.gitignore` related to the vendoring directory:

```
/vendor/**/.editorconfig
/vendor/**/*.jpg
/vendor/**/*.json
/vendor/**/.*.json
/vendor/**/Makefile
/vendor/**/*.png
/vendor/**/*.tmpl
/vendor/**/*.txt
!/vendor/**/*LICENSE*
!/vendor/modules.txt
/vendor/**/*.y*ml
/vendor/**/.*.y*ml
/vendor/github.com/dlclark/regexp2/testoutput1
```

Together with the `//go:embed` directive, these entries can cause a rather difficult to detect error. You should consider deleting them.
It was probably the author's goal to reduce the size of the repo, or a workaround for an early go vendoring problem.

Correct me if I'm wrong, but these seem unnecessary, rather a source of error.

szkiba:
## What?

The vendoring directory entries in .gitignore cause a potential error when using files embedded with the `//go:embed` directive. This change fixes a potential problem.

## Why?

The use of the current .gitignore results in a vendor directory that is about 800k smaller, but it also causes an error in the integration of the xk6-dashboard (and potentially also in the case of other dependencies in the future). The 800k increase (about 1.6%) is not worth the occasional correction of errors in the .gitignore file.

## Checklist

Closes #3510


.gitignore
```diff
@@ -15,19 +15,7 @@ script.js
 
 /versioninfo.json
 
-/vendor/**/.editorconfig
-/vendor/**/*.jpg
-/vendor/**/*.json
-/vendor/**/.*.json
-/vendor/**/Makefile
-/vendor/**/*.png
```
mstoykov, line 23: I guess this is the only line that actually was problematic?

szkiba: Actually the first one (*.json)

mstoykov: Given that we did not have (any known to me) problems for 3 years, I would kind of prefer to just fix the case that this doesn't work for you. 

There is not a single `.json` file here so I think just dropping that is way easier?

The alternative will be to actually readd all the known CI ones, MakeFile and the currently 2 big testdata files.

I am also okay with adding the line you did alongside these.

szkiba: > There is not a single .json file here so I think just dropping that is way easier?

I did this on the dashboard integration branch.
This is fine for me, and in this case let's reject this PR. I just thought potential future problems could be avoided.

joanlopez: I also agree that if this has been like this for a long time, and now we only have a very specific need, then I'd just add that as an exception to the rule, instead of reverting wider condition 👍🏻

szkiba: OK, I close it.

Issue #2683, opened 2022-09-13
k6/html: Have a way to get the selection out of an element


mstoykov:
### Feature Description

As reported in https://community.k6.io/t/create-selection-from-element/4643 it is comment in jquery to do [`selection.each`](https://k6.io/docs/javascript-api/k6-html/selection/selection-each/) and then get the selection of the element that `each` will give you and continue selecting.

This is currently not possible in k6, but we do have the selection of the element as part of the element so it shouldn't be that hard.

This likely though needs some though on the API

### Suggested Solution (optional)

We can just have new [Element](https://k6.io/docs/javascript-api/k6-html/element/) method `selection()` and be done with it. But it likely won't work a lot like jquery, how important is that is different question.

We can defined `$` to mean something in k6 - maybe by calling a method on `k6/html`. And then use that :shrug:.

But it will be confusing as `$('li')` will not really mean anything as we don't know which page to query.

Maybe it can only be difined in `each` :shrug:

Azhovan:
@mstoykov i would like to work on this issue. can you please assign it to me?

allansson:
> The $ symbol would be a property of the Element type, not a global variable or a function, and it would take a selector string as an argument and return a Selection object that contains the element itself and any other elements that match the selector. e.g. if element is an Element that represents a `div` element, then element.$('p') would return a Selection object that contains all the `p` elements inside the `div` element.

I've never seen a `$` being used in a method/property name unless to mark it as an observable or something internal/hidden. And, in my opinion, it reads horribly. It's hard to for the eye to parse and the non-descriptive name means I would have to look it up in the docs in order to understand what it does.

> We can defined $ to mean something in k6 - maybe by calling a method on k6/html. And then use that 🤷.
>
> But it will be confusing as $('li') will not really mean anything as we don't know which page to query.

I see no need to try and be jQuery. We have the Selection API. Reading the forum post, I don't think the author even expects that the API would be using `$`. They're just using jQuery as a reference.

Why not just a method `Selection.fromElement(element: Element)`?:

```ts
doc.find("li")
  .each((_, element) => {
    const listEl = Selection.fromElement(element)
      .closest('ul')
  })
```

We're not married to WHATWG or W3C either, so maybe `Element extends Selection`?

```ts
doc.find("li")
  .each((_, element) => {
    const listEl = element.closest("ul")
  })
```

 @mstoykov's alternative is also good:

```ts
element.selection()
element.toSelection()
element.asSelection()
```

[...]

PR #3519, opened 2023-12-28
feat: enabling the extraction of a specific selection from an element

Azhovan:
## What?
This pull request introduces enhancements to the k6 HTML module, including the addition of the `Selection` function. Its is designed to improve HTML parsing and manipulation capabilities within k6:

- `Selection` Method: Added to the `Element` type, it facilitates operations on the HTML content represented within an `Element`.

issue: https://github.com/grafana/k6/issues/2683 

## Why?
- Addresses [the need](https://github.com/grafana/k6/issues/2683 ) for a more intuitive API for DOM element selection and manipulation in performance testing scripts.

## Checklist

### DIFF
--- a/js/modules/k6/html/element.go
+++ b/js/modules/k6/html/element.go
@@ -255,7 +255,7 @@ func (e Element) IsSameNode(v goja.Value) bool {
 // Example:
 // sel := element.Selection()
 func (e Element) Selection() Selection {
-	return Selection{rt: e.sel.rt, sel: e.sel.sel, URL: e.sel.URL}
+	return *e.sel
 }
 
 func (e Element) GetElementsByClassName(name string) []goja.Value {
--- a/js/modules/k6/html/element.go
+++ b/js/modules/k6/html/element.go
@@ -246,6 +246,18 @@ func (e Element) IsSameNode(v goja.Value) bool {
 	return false
 }
 
+// Selection returns a Selection object based on the current Element.
+//
+// This function is used to create a Selection object that represents the same HTML
+// content as the Element. It is useful for performing operations or manipulations
+// on the HTML content within the scope of this Element.
+//
+// Example:
+// sel := element.Selection()
+func (e Element) Selection() Selection {
+	return Selection{rt: e.sel.rt, sel: e.sel.sel, URL: e.sel.URL}
+}
+
 func (e Element) GetElementsByClassName(name string) []goja.Value {
 	return elemList(Selection{e.sel.rt, e.sel.sel.Find("." + name), e.sel.URL})
 }
--- a/js/modules/k6/html/html.go
+++ b/js/modules/k6/html/html.go
@@ -280,38 +280,6 @@ func (s Selection) Last() Selection {
 	return Selection{s.rt, s.sel.Last(), s.URL}
 }
 
-// FromElement creates a new Selection based on a specified Element.
-//
-// This function is useful for creating a new Selection object that
-// encapsulates the same HTML structure as the provided Element. It
-// is particularly helpful when you need to perform further operations
-// or manipulations on the HTML content within the scope of this Element.
-// If the provided value is not an Element, an empty Selection is returned.
-//
-// Example:
-// sel := existingSelection.FromElement(jsElement)
-func (s Selection) FromElement(v goja.Value) Selection {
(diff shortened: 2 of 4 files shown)


js/modules/k6/html/element.go
```diff
@@ -246,6 +246,18 @@ func (e Element) IsSameNode(v goja.Value) bool {
...
+//
+// This function is used to create a Selection object that represents the same HTML
+// content as the Element. It is useful for performing operations or manipulations
+// on the HTML content within the scope of this Element.
+//
+// Example:
+// sel := element.Selection()
+func (e Element) Selection() Selection {
+	return Selection{rt: e.sel.rt, sel: e.sel.sel, URL: e.sel.URL}
+}
```
joanlopez, line 259:
```suggestion
func (e Element) Selection() Selection {
	return *e.sel
}
```

I think this would be slightly better in terms of maintainability and potential compiler optimizations, and the only drawback I see is in case `e.sel` is `nil, but that would need special handling in the current code as well.

js/modules/k6/html/html.go
```diff
@@ -280,6 +280,38 @@ func (s Selection) Last() Selection {
...
+// Example:
+// sel := existingSelection.FromElement(jsElement)
+func (s Selection) FromElement(v goja.Value) Selection {
+	elm, ok := v.Export().(Element)
+	if !ok {
+		// Return an empty Selection if the value is not an Element
+		return s.emptySelection()
+	}
+
+	return Selection{rt: s.rt, sel: elm.sel.sel, URL: s.URL}
```
joanlopez, line 300: What's the reasoning to re-use `rt` and `URL` from `s` instead of `elm.sel` ones? 🤔

js/modules/k6/html/html.go
```diff
@@ -280,6 +280,38 @@ func (s Selection) Last() Selection {
...
+// Locator selects HTML elements based on a CSS selector or XPath.
+//
+// This method is designed to provide a consistent way of selecting elements,
+// similar to how 'page.locator' works in the 'k6-browser'. It uses goquery
+// to perform the selection based on the given selector string.
+//
+// Example:
+// sel := mySelection.Locator(".myClass")
+func (s Selection) Locator(selector string) Selection {
+	return Selection{s.rt, s.sel.Find(selector), s.URL}
```
joanlopez, line 312: Isn't it equivalent to `s.Find()`? If so, why not re-using such method, so we reduce chances of divergency in the future?

joanlopez:
It looks generally good, despite of the minor suggestions I left. However, after reading the discussion on the issue, I am not super confident about whether or not we want to expose/add (all) these methods. So, adding Mihail as reviewer, cause I've seen he's been involved in the discussion and so might have some feedback/thoughts to share around that.

Thanks!

Azhovan:
what do you think @mstoykov ?

mstoykov:
I think `Locator` was just something @w1kman talked about, but shouldn't really be part of this fix at all.

the `FromElement` was based on the idea tha `Selection` is the prototype that is implemented by each selection and it will have a method (like static/moduled method in most other cases) and htat will be used.

As we do not have currently a `Selection` prototype (AFAIK) - this isn't really possible.

as such I feel like we should just go with the `selection` method and drop everythign else ?

js/modules/k6/html/element.go
Azhovan: thanks!

mstoykov (approved):
LGTM!

I kind of would prefer if the test was with teh [original example](https://community.grafana.com/t/create-selection-from-element/95923?u=mstoykov), but not a big problem as long as it works.

Azhovan:
> I think `Locator` was just something @w1kman talked about, but shouldn't really be part of this fix at all.
> 
> The `FromElement` was based on the idea that `Selection` is the prototype that is implemented by each selection, and it will have a method (like static/moduled method in most other cases) and that will be used.
> 
> As we do not have currently a `Selection` prototype (AFAIK) - this isn't really possible.
> 
> As such I feel like we should just go with the `selection` method and drop everything else?
> 
> p.s. sorry I was away for the holidays and this then fell through the cracks 🙇

Thank you! The modifications have been implemented.

joanlopez (approved):

PR #3522, opened 2024-01-08
Add browser tracing release notes

ka3de:
## What?

Adds release notes for tracing instrumentation implemented in k6 browser module.

## Why?

Required step towards `v0.49.0` release.

## Checklist

## Related PR(s)/Issue(s)

grafana/xk6-browser#1100
grafana/xk6-browser#1104

### DIFF
--- a/release notes/v0.49.0.md
+++ b/release notes/v0.49.0.md
@@ -17,9 +17,13 @@ k6 `v0.49.0` is here 🎉! This release includes:
 
 _optional intro here_
 
-### `<big_feature_1>` `#pr`
+### Add tracing to browser module [browser#1100](https://github.com/grafana/xk6-browser/pull/1100)
 
-_what, why, and what this means for the user_
+The browser module now generates traces that provide a representation of its inner workings, such as API methods executed (for example `browser.newPage` and `page.goto`), page navigations, and [Web Vitals](https://grafana.com/docs/k6/latest/using-k6-browser/metrics/#googles-core-web-vitals) measurements.
+
+Take into account that currently the instrumented methods are a subset of all the methods exposed by the browser module API, but this will be extended in the future.
+
+The traces generation for the browser module depends on the overall `k6` traces option introduced in [v0.48.0](https://github.com/grafana/k6/releases/tag/v0.48.0). Check out the [documentation](https://grafana.com/docs/k6/latest/using-k6/k6-options/reference/#traces-output) to learn more about it.
 
 ### `<big_feature_n>` `#pr`
 
--- a/release notes/v0.49.0.md
+++ b/release notes/v0.49.0.md
@@ -21,7 +21,7 @@ _optional intro here_
 
 The browser module now generates traces that provide a representation of its inner workings, such as API methods executed (for example `browser.newPage` and `page.goto`), page navigations, and [Web Vitals](https://grafana.com/docs/k6/latest/using-k6-browser/metrics/#googles-core-web-vitals) measurements.
 
-Take into account that currently the instrumented methods are a subset of all the methods exposed by the browser module API, but this will be extended in the future.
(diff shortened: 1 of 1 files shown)


release notes/v0.49.0.md
```diff
@@ -17,9 +17,13 @@ k6 `v0.49.0` is here 🎉! This release includes:
 
 _optional intro here_
 
-### `<big_feature_1>` `#pr`
+### Add tracing to browser module [browser#1100](https://github.com/grafana/xk6-browser/pull/1100)
 
-_what, why, and what this means for the user_
+The browser module now generates traces that provide a representation of its inner workings, such as API methods executed (for example `browser.newPage` and `page.goto`), page navigations, and [Web Vitals](https://grafana.com/docs/k6/latest/using-k6-browser/metrics/#googles-core-web-vitals) measurements.
+
```
ka3de, line 23: I was thinking on adding an image here of how a browser trace can is visualized in Tempo, but I see it is not a common practice in release notes to add images. Should we add a folder under `release notes/` for resources that are referenced in the release notes? Or is the text description good enough?

ankur22: Agreed, a visual representation of this would be useful. Maybe best for the release email that goes out after the release instead of here?

release notes/v0.49.0.md
```diff
@@ -17,9 +17,13 @@ k6 `v0.49.0` is here 🎉! This release includes:
...
+Take into account that currently the instrumented methods are a subset of all the methods exposed by the browser module API, but this will be extended in the future.
```
ankur22, line 24: It felt like the word `currently` was enough to help highlight the point. Happy either way though.

```suggestion
Currently the instrumented methods are a subset of all the methods exposed by the browser module API, but this will be extended in the future.
```

release notes/v0.49.0.md
```diff
@@ -17,9 +17,13 @@ k6 `v0.49.0` is here 🎉! This release includes:
...
+
+The traces generation for the browser module depends on the overall `k6` traces option introduced in [v0.48.0](https://github.com/grafana/k6/releases/tag/v0.48.0). Check out the [documentation](https://grafana.com/docs/k6/latest/using-k6/k6-options/reference/#traces-output) to learn more about it.
```
ankur22, line 26: Should the link to the traces output doc point to the one in `v0.48.0`? I can't remember how many versions we keep of docs though, so latest might be the best option if we only keep the last 3 (as an example).

ankur22 (approved):
LGTM. I left some minor things to ponder over, but happy as it is too.

release notes/v0.49.0.md
ka3de: Taking a look at the [Grafana docs](https://grafana.com/docs/versions/?project=/docs/k6/), seems like it only shows docs for the latest two versions, and for the "next" one, which is WIP.

Considering that we are referencing a feature that was released in a previous version, it might be OK to reference the docs for that specific version, but personally I think it's best to refer to the latest doc, as that will be correct once we release this version (which is most probably the time where most users would read these release notes), and will also not lead to a 404 in a couple of releases.

So both options will be correct once we release `v0.49.0`, the question is, what do we prefer "long term"?
- A link that might lead to a 404 in a couple of releases
- A link that might lead to not completely correct docs for `v0.49.0` (which can happen if the option implementation changes).

The thing is, if we link to the `v0.48.0` version here, in a couple of releases the user will get a 404 and will end up most probably querying the latest docs for the option. That's why I'm thinking it might be better to keep the reference to `latest`.

Still both cases produce small inconvenience that can be worked out by users.

release notes/v0.49.0.md
ka3de: What are your thoughts on this @mstoykov ?

PR #3523, opened 2024-01-08
Automate experimental module updates

andrewslotin:
## What?

This PR adds Dependabot config to check for experimental module updates daily and create PRs to upgrade them. 

## Why?

The process of releasing a new version of an experimental module, such as xk6-browser or xk6-redis, requires a manual step to upgrade the dependency in k6's `go.mod`, vendor the code and open a PR. This is a routine task that can be automated to avoid human error.

## Checklist

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,7 +8,6 @@ updates:
       # Keep the experimental modules up-to-date
       - dependency-name: "github.com/grafana/xk6-*"
         dependency-type: "all"
-    target-branch: "master"
     commit-message:
       prefix: "Bump "
       include: "scope"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,18 @@
+version: 2
+updates:
+  - package-ecosystem: "gomod"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    allow:
+      # Keep the experimental modules up-to-date
+      - dependency-name: "github.com/grafana/xk6-*"
+        dependency-type: "all"
+    target-branch: "master"
+    commit-message:
+      prefix: "Bump "
+      include: "scope"
+    assignees:
+      - "octocat"
+    reviewers:
+      - "grafana/k6-core"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -11,7 +11,5 @@ updates:
     commit-message:
       prefix: "Upgrade experimental module "
       include: "scope"
-    assignees:
-      - "octocat"
     reviewers:
       - "grafana/k6-core"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,7 +9,7 @@ updates:
       - dependency-name: "github.com/grafana/xk6-*"
         dependency-type: "all"
     commit-message:
-      prefix: "Bump "
+      prefix: "Upgrade experimental module "
       include: "scope"
     assignees:
       - "octocat"


.github/dependabot.yml
```diff
@@ -0,0 +1,18 @@
...
+    allow:
+      # Keep the experimental modules up-to-date
+      - dependency-name: "github.com/grafana/xk6-*"
+        dependency-type: "all"
+    target-branch: "master"
+    commit-message:
+      prefix: "Bump "
+      include: "scope"
+    assignees:
+      - "octocat"
```
olegbespalov, line 16: Is it required to pass something? I mean I wondering if we should use an `octocat` handle here

andrewslotin: No, it's not required. My intention was to highlight these PRs as automatically created, but I'm OK with removing this option.

.github/dependabot.yml
```diff
@@ -0,0 +1,18 @@
...
+updates:
+  - package-ecosystem: "gomod"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    allow:
+      # Keep the experimental modules up-to-date
+      - dependency-name: "github.com/grafana/xk6-*"
+        dependency-type: "all"
+    target-branch: "master"
```
joanlopez, line 11: Nit; I think this is not strictly required because it automatically looks for the default branch, and so if we don't add it, we don't need to change it if anytime we in the future we change the default branch. But it's also true that in such case, this will be the least worrying thing, so up to you!

.github/dependabot.yml
```diff
@@ -0,0 +1,18 @@
+version: 2
+updates:
+  - package-ecosystem: "gomod"
```
joanlopez, line 3: Nit; As I had the doubt and had to double-check the docs, do you think it worths adding a comment here stating that `gomod` manages vendor by default, and thus `vendor: true` is not needed?

.github/dependabot.yml
```diff
@@ -0,0 +1,18 @@
...
+    directory: "/"
+    schedule:
+      interval: "daily"
+    allow:
+      # Keep the experimental modules up-to-date
+      - dependency-name: "github.com/grafana/xk6-*"
+        dependency-type: "all"
+    target-branch: "master"
+    commit-message:
+      prefix: "Bump "
```
joanlopez, line 13: I think it already adds `Bump` as prefix by default 🤔 

Looking at an example of Dependabot commit, without `commit-message` configured, it looks like:
`Bump github.com/spf13/afero from 1.6.0 to 1.11.0`

joanlopez (dismissed):
It looks good! I left some comments, that these are mostly nits, and non-blockers, so 👍🏻

.github/dependabot.yml
andrewslotin: I did check this indeed :)

> Don't use this option if you're using `gomod` as Dependabot automatically detects vendoring for this tool.

.github/dependabot.yml
andrewslotin: Would it make sense to make it a bit more descriptive instead? Something like `Update experimental module x/y/z from a.b.c to d.e.f`?

.github/dependabot.yml
andrewslotin: Good point, will drop this one.

.github/dependabot.yml
olegbespalov: > My intention was to highlight these PRs as automatically created, but I'm OK with removing this option.

Probably, my comment shows that the intention could be easily misinterpreted or at least not clear :sweat_smile: 

If I'm correct, here is an example of the PR created by Dependabot: https://github.com/grafana/jslib.k6.io/pull/104. The first comment already shows that this PR is automatically created, so your concerns are already covered.

olegbespalov (approved):

.github/dependabot.yml
joanlopez: Exactly, that's why I was wondering whether or not would make sense to leave a comment, for future readers.

joanlopez (approved):

Issue #2952, opened 2023-03-02
Expand usage collection to include extensions in use


javaducky:
### Feature Description

Currently, k6 reports anonymous usage data as described in the [docs](https://k6.io/docs/misc/usage-collection/). Thanks to #1741 in v0.43.0, we can now access version info for bundled extensions with the `k6 version` output. To help determine priority for certain extensions, having this version information included in the payload of the usage reports would be nice. I want to discuss bringing this additional metadata to the usage collection payload.

Of course, anonymity and the ability to opt-out are paramount. The only scenario I can think of which could break anonymity would be if someone is creating/using an extension within a private repository. Their Github organization would be "leaked," but we would not have any insight into the extension other than what could be implied by its name.

imiric:
I think adding this would a far greater privacy concern than the current opt-out usage report. Remember that with the current usage report, even though the submitted data is minimal, the IP address of the submitter is still potentially leaked and logged. IP addresses are treated as personal data under GDPR, so the current opt-out telemetry we use is non-compliant in that sense.

Before even considering this, I think we should step back and have a discussion about making the report opt-in instead. Recently, there has been [a lot of discussion](https://news.ycombinator.com/item?id=34707583) about [the proposal to add telemetry to Go](https://research.swtch.com/telemetry-opt-in). Thankfully, feedback from the community was taken into account, and this will be opt-in.

Once we make that change, then in principle, I wouldn't be opposed with expanding the data to include the used k6 extensions.

But I'll ask this: for what purpose do we realistically want to use that data, other than it "would be nice"? We should have a clear purpose for it, otherwise it will just sit there, and we'll glance at it every few months out of curiosity, like I think we do with the current usage report data.

na--:
I am on the fence about whether the benefits of adding more telemetry would outweigh the costs... I am heavily leaning towards "no", at least if we take the simple approach to just send what we have in `k6 version` :thinking: 

> The only scenario I can think of which could break anonymity would be if someone is creating/using an extension within a private repository. Their Github organization would be "leaked," but we would not have any insight into the extension other than what could be implied by its name.

To increase the privacy, we can strip everything besides the last part of the module name and version :thinking: e.g. instead of `github.com/grafana/xk6-browser@v0.8.1`, we can just send `xk6-browser@v0.8.1`?

[...]

PR #1343, opened 2020-03-04
Re-enable the usage report

imiric:
Closes #1301

This re-enables the usage report, and updates its payload to include executor types and number of them used.

Example of payload:

```json
{
  "duration": "1m0.029405562s",
  "executors": {
    "constant-arrival-rate": 1,
    "constant-looping-vus": 2,
    "externally-controlled": 1,
    "per-vu-iterations": 1,
    "shared-iterations": 1,
    "variable-arrival-rate": 1,
    "variable-looping-vus": 2
  },
  "goarch": "amd64",
  "goos": "linux",
  "iterations": 2175,
  "k6_version": "0.26.0",
  "vus_max": 130
}
```

Note that currently this POST to https://reports.k6.io/ fails with a `502 Bad Gateway`, which I'm assuming is because of the change in schema and the backend will have to be updated to match. The differences are minor: this drops `duration_st` and adds `executors`.

PR #3525, opened 2024-01-08
Collect usage about used js modules and outputs

codebien:
## What?

It collects usage stats about the js modules and outputs used in the test. It collects only the names of the used modules or outputs that are built-in in k6 binary.  

## Why?

It will help to refine our prioritization method based on more objective and data-driven decisions regarding a feature or a breaking change to introduce.

Updates https://github.com/grafana/k6/issues/2952


cmd/report.go
```diff
@@ -0,0 +1,100 @@
...
+	for _, ec := range execScheduler.GetExecutorConfigs() {
+		executors[ec.GetType()]++
+	}
+
+	// collect the report only with k6 public modules
+	publicModules := make([]string, 0, len(importedModules))
+	for _, module := range importedModules {
+		// Exclude JS modules extensions to prevent to leak
+		// any user's custom extensions
+		if strings.HasPrefix(module, "k6/x") {
```
joanlopez, line 38: Would make sense to keep a list, even if manually, here to also get usage stats for the extensions that we do own? At least until we find a better way to do so.

From the product perspective, I guess it could make lot of sense to know what extensions worth some attention and which ones not.

cmd/report.go
```diff
@@ -0,0 +1,100 @@
...
+		publicModules = append(publicModules, module)
+	}
+
+	builtinOutputs := map[string]bool{
+		"cloud": true, "csv": true, "experimental-prometheus-rw": true,
+		"influxdb": true, "json": true, "statsd": true,
+	}
+
+	// collect only the used outputs that are builtin
+	publicOutputs := make([]string, 0, len(outputs))
```
joanlopez, line 56:
```suggestion
	publicOutputs := make([]string, 0, len(builtinOutputs))
```

Nit; I think `publicOutputs` could only be as big as `builtinOutputs`, despite of the length of `outputs`.

cmd/report.go
```diff
@@ -0,0 +1,100 @@
...
+			continue
+		}
+		publicOutputs = append(publicOutputs, o)
+	}
+
+	execState := execScheduler.GetState()
+	return report{
+		Version:    consts.Version,
+		Executors:  executors,
+		VUsMax:     execState.GetInitializedVUsCount(),
```
joanlopez, line 68: Is this accurate? I'm not expert enough yet, but looks like the value returned under the hood `initializedVUs` can be decreased, so I'm wondering whether this is accurate or not depending on the scenarios configured, and whether this represents a "max" value.

cmd/report.go
```diff
@@ -0,0 +1,100 @@
...
+	body, err := json.Marshal(r)
+	if err != nil {
+		return err
+	}
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, "https://reports.k6.io", bytes.NewBuffer(body))
+	if err != nil {
+		return err
+	}
+	req.Header.Set("Content-Type", "application/json")
+	res, err := http.DefaultClient.Do(req)
```
joanlopez, line 95: I know the use of the `DefaultClient` is generally discouraged, because it lacks the definition of timeouts and other proper settings for production environments. Is it safe to use it here?

cmd/report_test.go
```diff
@@ -0,0 +1,61 @@
...
+	})
+	require.NoError(t, err)
+	s.GetState().ModInitializedVUsCount(6)
+	s.GetState().AddFullIterations(170)
+	s.GetState().MarkStarted()
+	time.Sleep(10 * time.Millisecond)
+	s.GetState().MarkEnded()
+
+	r := createReport(s, importedModules, outputs)
+	assert.NotEmpty(t, r.Version)
```
joanlopez, line 54: Would make sense to switch to  (or add) something like?

```suggestion
	assert.Equal(t, consts.Version, r.Version)
```

cmd/report_test.go
```diff
@@ -0,0 +1,61 @@
...
+
+	outputs := []string{
+		"json",
+		"xk6-output-influxdb",
+	}
+
+	logger := testutils.NewLogger(t)
+	opts, err := executor.DeriveScenariosFromShortcuts(lib.Options{
+		VUs:      null.IntFrom(10),
+		Duration: types.NullDurationFrom(5 * time.Minute),
```
joanlopez, line 36: Nit; If I'm not wrong, this is only needed to have `constant-vus` as executor. However, imho this is a bit confusing, because it's used nowhere, and differs from the actual execution (~10ms) and the actual amount of VUs (6).

So, I'd suggest to either make it match with the values below (perhaps extracting these values as a shared variable?), or just removing these lines and switching the executor type in the assertion, which is even simpler/shorter, thus easier to understand.

cmd/report.go
codebien:
```
xk6-client-tracing
xk6-notification
xk6-disruptor
xk6-sql
xk6-exec
xk6-kubernetes
xk6-ssh
xk6-loki
xk6-dashboard
xk6-output-influxdb
xk6-output-kafka
xk6-output-timescaledb
```
This is the list of extensions marked as official here: https://k6.io/docs/extensions/get-started/explore. None of them sounds so relevant at the moment to justify the risk of having this logic. Let me know if you have a different opinion.

joanlopez: Sounds fair! Thanks!

cmd/report.go
codebien: Yes, at least in terms of _planned_ max VUs, it should be accurate. The unique case for decreasing is in the externally controlled executor which is a special case.

As an alternative, we may use the latest generated `VUID` but it has the same issue of being an indirect logic, so if the VUID logic changes then the report would be impacted without any guarantee to not introduce a bug.

You can read more about the original decision here https://github.com/grafana/k6/pull/1343#discussion_r388141537

cmd/report.go
codebien: >it lacks the definition of timeouts
>Is it safe to use it here?

Specifically, regarding the timeout, we should be safe as we are passing the timeout-limited context.

In general, it would be a good potential improvement. However, not so far in the future, we have to migrate to the new Grafana's service API. Then it may be convenient to rewrite and refactor this client all at once.

cmd/report.go
joanlopez: Thanks so much, all clear now! :)

cmd/report.go
joanlopez: Fair enough, let's move on! Also, looks like this code already existed before, so 🚀

joanlopez (dismissed):
I left a few comments, but none of them is blocking, so it looks good enough to be merged to me! 👍🏻 Thanks!

cmd/test_load.go
```diff
@@ -39,6 +40,7 @@ type loadedTest struct {
 	preInitState   *lib.TestPreInitState
 	initRunner     lib.Runner // TODO: rename to something more appropriate
 	keyLogger      io.Closer
+	moduleResolver *modules.ModuleResolver
```
olegbespalov, line 43: Looking into the usage of this property in all cases, it sets the value using the value of the `runner,` which already exists as a `test.initRunner`. Am I missing something, or could we avoid introducing the duplication?

codebien: `test.initRunner` is a lib.Runner interface https://github.com/grafana/k6/blob/65b2f27d1a0612846467e48418f469984724e007/lib/runner.go#L54 and it doesn't provide direct access to the ModuleResolver type that is a specific of the [js.Runner](https://github.com/grafana/k6/blob/65b2f27d1a0612846467e48418f469984724e007/js/runner.go#L47-L48) implementation.

It sounds like this is not clear enough from the code, do you think a comment somewhere could help?

olegbespalov: Ah, I see 👍

No I prefer to avoid commenting on this. It's just a lack of the PR context. In Code it should be fine.

codebien:
@mstoykov as you are working on refactoring some closer areas, it would be good to get a review from you as well.

cmd/report.go
```diff
@@ -0,0 +1,100 @@
...
+		// that identifies a k6 built-in stable or experimental module.
+		// For example, it doesn't include any modules imported from the file system.
+		if !strings.HasPrefix(module, "k6") {
+			continue
+		}
+		publicModules = append(publicModules, module)
+	}
+
+	builtinOutputs := map[string]bool{
+		"cloud": true, "csv": true, "experimental-prometheus-rw": true,
```
olegbespalov, line 51: Nit: These output keys (cloud, csv, etc.) come from the map https://github.com/grafana/k6/blob/master/cmd/outputs.go#L26-L47. It's better to extract this as the variable/method and use it here. That way, we protect the future maintainers/contributors from forgetting to update this place. 

Some alternatives (e.g., using `const`) could also be considered, but the main idea is to couple the places through the code, not the string values.

olegbespalov (dismissed):
LGTM, with a minor non-blocking nit.

mstoykov (approved):
LGTM! I did check commit by commit and did some :+1: on some comments that I agreed more strongly with :sweat_smile: 

My only concern around pushing every imported module has been adressed so I am happy.

I guess (as future improvement) we can also send whether and how many `k6/x/` modules were imported? Maybe even how many source modules were imported :shrug: . Just as a number.

Also, I guess at some point we might want to have a "whitelist" of `k6/x/' modules that are still reported, but that definitely should be done later and with more discussion.

cmd/report.go
codebien: Added a new proposal for addressing this in the last commits

mstoykov (approved):

cmd/outputs.go
```diff
@@ -19,30 +19,46 @@ import (
 	"github.com/grafana/xk6-output-prometheus-remote/pkg/remotewrite"
 )
 
+// builtinOutput marks the available builtin outputs.
+//
+//go:generate enumer -type=builtinOutput -trimprefix builtinOutput -transform=kebab -output builtin_output_gen.go
+type builtinOutput uint32
+
+const (
```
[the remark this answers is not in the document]
codebien: >enums decrease code reading/maintaining abilities

The reason is to reduce the probability of introducing bugs in the future by changing the list of builtin outputs (adding or removing them). In this way, we can rely on a single source of truth across the code more than directly on strings.

>and it seems like the builtin prefix could be omitted

Do you mean only from the items' names or also from the type? Because having an `output` type when we already have the `output` package and `output.Output` type could create confusion if there isn't an element to differentiate them.

joanlopez (approved):

cmd/outputs.go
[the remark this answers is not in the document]
joanlopez: Honestly, as someone who has never used `enumer` before, I have to admit that the generated code looks a bit overwhelming. But as suggested before, I'm _generally_ in favor or reducing chances to introduce errors by reducing duplicity, and as `enumer` seems to be already used in the project, I'd say it looks good enough to be merged to me.

olegbespalov (approved):

PR #3526, opened 2024-01-09
js: path based tests refactor

mstoykov:
Highly recommended to go commit by commit as it makes the changes way more managable.

## What?

A bunch of refactors for some of the js path based tests.

## Why?

 Mostly to make them easier to understand and extend.

## Checklist

### DIFF
--- a/js/path_resolution_test.go
+++ b/js/path_resolution_test.go
@@ -17,16 +17,16 @@ func TestOpenPathResolution(t *testing.T) {
 		fs := fsext.NewMemMapFs()
 		err := writeToFs(fs, map[string]any{
 			"/path/to/data.txt": "data file",
-		})
-		require.NoError(t, err)
-		data := `
+			"/path/scripts/script.js": `
 		export let data = open("../to/data.txt");
 		if (data != "data file") {
 			throw new Error("wrong content " + data);
 		}
 		export default function() {}
-	`
-		b, err := getSimpleBundle(t, "/path/scripts/script.js", data, fs)
+	`,
+		})
+		require.NoError(t, err)
+		b, err := getSimpleBundle(t, "/main.js", `export { default } from "/path/scripts/script.js"`, fs)
 		require.NoError(t, err)
 
 		_, err = b.Instantiate(context.Background(), 0)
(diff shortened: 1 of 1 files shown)


js/path_resolution_test.go
```diff
@@ -17,9 +17,9 @@ func TestOpenPathResolution(t *testing.T) {
 	}{
 		"simple": {
 			fsMap: map[string]any{
-				"/path/to/data.txt": "data file",
-				"/path/totally/different/directory/script.js": `
-					export let data = open("../../../to/data.txt");
+				"/A/B/data.txt": "data file",
```
codebien, line 20: I find something like `/u/myuser/projectb/script1/test` (or as it was in the previous form) easier to navigate visually with eyes.

I didn't find it to be easier to check the correctness of the path in something like the following:
```js
"/A/A/A/A/script.js": `
  export let data = open("../../../B/data.txt");
```
Just sharing my opinion, it's not required you change this current PR. Maybe, it will be useful to you in the next PR.

codebien (approved):
"/path/totally/different/directory/script.js": 
"/A/A/A/A/script.js": `

js/path_resolution_test.go
mstoykov: I definitely had problems with the previous names ... especially with the cases I needed to come up with.

The original try was with words such as "orangutan" and "pikachu" to be more distinctive, but then I figured out ... the letters A, B and C shall be sufficient, and it also makes the paths *way* shorter, which was also a readability problem for me.

But it will be nice if more people give their opinions as we can change that to something everybody is mostly happy with.

I don't really want `/home/username/path/to/somewhere/` as this IMO makes it way too long. Being more realistic here isn't really a key, arguably readability and understandability of the tests is more important than them looking like more closely to what a user might hit. At least IMO

olegbespalov (approved):
Generally LGTM :+1:

js/path_resolution_test.go
mstoykov: > it was an apparent attempt to distinguish between

do you mean that the paths were different between the cases? Because it was mostly a case of ... them building on top of each other - actually at least one of the paths was always the same. I did have to normalize additionally when I was making the table tests, but even before that the idea was that the paths were supposed to be the same between tests. Just in the more complex ones it needed more files - so more paths.

Not certain if that is what you meant though.

> a brief explanation of what the case is checking will serve that need better.

As in like what simple, intermediate and complex are? That might be a good idea, given that the major comment is the link to the issue where it talks about what the current behavior is.

olegbespalov: > do you mean that the paths were different between the cases? Because it was mostly a case of ... them building on top of each other - actually at least one of the paths was always the same. 

Yes, but if it's a copy-paste, then the current version is better.

 > As in like what simple, intermediate and complex are? That might be a good idea, given that the major comment is the link to the issue where it talks about what the current behavior is.

:+1:

mstoykov: I think I will prefer to do the documentation in a separate PR as I already have changes on top of this and don't want to retrigger reviews.

So I will merge this and try to get both the remaining changes and the documentation in PRs.

Issue #2578, opened 2022-06-27
Tests fail on macos with go 1.18


Llandy3d:
### Brief summary

Trying to run tests with `make tests` on macos with `golang 1.18` results in errors while the expected output is for them to complete.

The problem seems to not be present by using `golang 1.17` and it seems related to his bug:
https://github.com/golang/go/issues/52010

### k6 version

3e21e678fa13f1cdd442877a105a12f0f255c505 <- commit hash, but it doesn't matter since the go version it's the problem

### OS

MacOs Monterey 12.1

### Steps to reproduce the problem

with `go 1.18`:

`make tests`

### Expected behaviour

tests complete without errors.

### Actual behaviour

```
❯ make tests
go test -race -timeout 210s ./...
?   	go.k6.io/k6	[no test files]
ok  	go.k6.io/k6/api	(cached)
?   	go.k6.io/k6/api/common	[no test files]
ok  	go.k6.io/k6/api/v1	(cached)
?   	go.k6.io/k6/api/v1/client	[no test files]
ok  	go.k6.io/k6/cloudapi	(cached)
ok  	go.k6.io/k6/cmd	19.669s
ok  	go.k6.io/k6/cmd/integration_tests/eventloop	(cached)
ok  	go.k6.io/k6/cmd/integration_tests/testmodules/events	(cached)
ok  	go.k6.io/k6/converter/har	(cached)
ok  	go.k6.io/k6/core	(cached)
ok  	go.k6.io/k6/core/local	(cached)
ok  	go.k6.io/k6/errext	(cached)
?   	go.k6.io/k6/errext/exitcodes	[no test files]
time="2022-06-27T12:54:58+02:00" level=info msg="{\"text\":\"nativeObject\"}" source=console
time="2022-06-27T12:54:58+02:00" level=warning msg="There were unknown fields in the options exported in the script" error="json: unknown field \"myOption\""
time="2022-06-27T12:54:58+02:00" level=warning msg="The moduleSpecifier \"buffer\" has no scheme but we will try to resolve it as remote module. This will be deprecated in the future and all remote modules will need to explicitly use \"https\" as scheme."
time="2022-06-27T12:54:58+02:00" level=info msg=1 source=console
time="2022-06-27T12:54:58+02:00" level=warning msg="The moduleSpecifier \"buffer\" has no scheme but we will try to resolve it as remote module. This will be deprecated in the future and all remote modules will need to explicitly use \"https\" as scheme."
[...]

PR #3527, opened 2024-01-10
Generate certificates at runtime for runner test

joanlopez:
## What?

It modifies the `TestVUIntegrationClientCerts` test on `runner_test.go` in order to generate certificates at runtime, instead of using pre-generated ones.

## Why?

This test is currently failing when `runtime.GOOS == "darwin"`, because it expects the `"certificate signed by unknown authority"` error but it gets `"certificate is not standards compliant"` instead, because the pre-generated certificate has a validity of 1000 years since 2022, and that's not considered standards compliant by Apple.

So, instead of having a pre-generated certificate that expires every year (so we'd need to update it every year), this makes the test to generate the required certificates at runtime (so always up to date), which also seems to be fine for MacOS.

## Checklist

## Related PR(s)/Issue(s)

I think it fixes https://github.com/grafana/k6/issues/2578, cause with this I no longer see errors on MacOS.

### DIFF
--- a/js/runner_test.go
+++ b/js/runner_test.go
@@ -910,15 +910,11 @@ func TestVUIntegrationMetrics(t *testing.T) {
 	}
 }
 
-func GenerateTLSCertificateWithCA(t *testing.T, host string, notBefore time.Time, validFor time.Duration, caCert *x509.Certificate, caKey *rsa.PrivateKey) ([]byte, []byte) {
-	return generateTLSCertificate(t, host, notBefore, validFor, caCert, caKey)
+func generateTLSCertificate(t *testing.T, host string, notBefore time.Time, validFor time.Duration) ([]byte, []byte) {
+	return generateTLSCertificateWithCA(t, host, notBefore, validFor, nil, nil)
 }
 
-func GenerateTLSCertificate(t *testing.T, host string, notBefore time.Time, validFor time.Duration) ([]byte, []byte) {
-	return generateTLSCertificate(t, host, notBefore, validFor, nil, nil)
-}
-
-func generateTLSCertificate(t *testing.T, host string, notBefore time.Time, validFor time.Duration, parent *x509.Certificate, ppriv *rsa.PrivateKey) ([]byte, []byte) {
+func generateTLSCertificateWithCA(t *testing.T, host string, notBefore time.Time, validFor time.Duration, parent *x509.Certificate, ppriv *rsa.PrivateKey) ([]byte, []byte) {
 	priv, err := rsa.GenerateKey(rand.Reader, 2048)
 	require.NoError(t, err)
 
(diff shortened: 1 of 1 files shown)


js/runner_test.go
```diff
@@ -937,7 +945,6 @@ func GenerateTLSCertificate(t *testing.T, host string, notBefore time.Time, vali
 		NotAfter:  notAfter,
 
 		KeyUsage:              keyUsage,
-		ExtKeyUsage:           []x509.ExtKeyUsage{x509.ExtKeyUsageServerAuth},
```
joanlopez, line 940: Beyond the new logic to have support for CA cert & key (`if parent == nil { ... }`), this is the only line from the original method that I have modified (removed), because with this line the test doesn't work (because the generated cert cannot be used to sign later) and without it everything works as expected, so 👌🏻

js/runner_test.go
```diff
@@ -910,7 +910,15 @@ func TestVUIntegrationMetrics(t *testing.T) {
 	}
 }
 
+func GenerateTLSCertificateWithCA(t *testing.T, host string, notBefore time.Time, validFor time.Duration, caCert *x509.Certificate, caKey *rsa.PrivateKey) ([]byte, []byte) {
```
codebien, line 913:
```suggestion
func generateTLSCertificateWithCA(t *testing.T, host string, caCert *x509.Certificate, caKey *rsa.PrivateKey) ([]byte, []byte) {
```
- Do we need to have it exported? (In alternative, if we find a way to make it a bit simpler API and we have other tests that could benefit from it then we may consider moving to the `testutils` package.)
- It sounds like we always use the same time config so we may skip it from the API
- Do we need to define a CA? Could we use a pre-defined CA for all our internal tests? Like `k6ca.test.example`

joanlopez: > Do we need to have it exported? 

I don't think so, as it's only used within this file.
I just kept it exporter because it already was, similar to `GetTestServerWithCertificate` function also present in that file, but I'm happy to make all these unexported.

> It sounds like we always use the same time config so we may skip it from the API
> Do we need to define a CA? Could we use a pre-defined CA for all our internal tests? Like k6ca.test.example

I'm not sure what you mean here. Do you mean the CA args?
Could you elaborate a bit more how would you have a pre-defined CA? 
Storing the CA cert/key as a global variable, and having a `sync.Once` to initialize it once? Or now?
Because I think in this case we need to generate two certificates (client & server) that are signed with the same CA cert & key.

Thanks! 🙏🏻

codebien: >I'd feel like prematurely generalizing.

Yes, I feel the same if we don't have other occurrences. Then we can unexport them.

>I'm not sure what you mean here.

Sorry, I meant the host.

joanlopez: > Sorry, I meant the host.

Okay! So, technically it's true that all the current calls to `GenerateTLSCertificateWithCA` use the `host` = `127.0.0.1` (in fact, I think it only really matters for the server one), but at the same time I feel like hardcoding that value would make that method less usable, as it's exactly the same as the former `GenerateTLSCertificate` but with custom CA cert & key, so `host` could be any value (the consumer needs, like with the former method).

In fact, the host for the CA is defined on the first call to `GenerateTLSCertificate`, and that method is used with different `host` along the code, so that function needs to remain with a parameterized host.

So, that being said, unless you want to suggest something different specifically (please do, if you want!), I'd prefer to leave it as is, because I don't see any good way to do it.

joanlopez: And yes, I'm going to turn these methods into unexported ones! 🙏🏻

mstoykov (approved):

Issue #1230, opened 2019-11-07
ui test fails on Mac OSX


cuonglm:
```
$ cd ui
$ go1.13.4 test
--- FAIL: TestForm (0.00s)
    --- FAIL: TestForm/Banner (0.00s)
        Error Trace:    form_test.go:42
        Error:      	Not equal:
                    	expected: "Hi!\n\n"
                    	actual  : "\x1b[34mHi!\x1b[0m\n\n"
        Test:       	TestForm/Banner
    --- FAIL: TestForm/Defaults (0.00s)
        Error Trace:    form_test.go:83
        Error:      	Not equal:
                    	expected: "  label a [default a]:   label b [default b]: "
                    	actual  : "  label a \x1b[2;36m[default a]\x1b[0m:   label b \x1b[2;36m[default b]\x1b[0m: "
        Test:       	TestForm/Defaults
    --- FAIL: TestForm/Errors (0.00s)
        Error Trace:    form_test.go:96
        Error:      	Not equal:
                    	expected: "  label: - invalid input, min length is 6\n  label: - invalid input, max length is 10\n  label: "
                    	actual  : "  label: \x1b[31m- invalid input, min length is 6\x1b[0m\n  label: \x1b[31m- invalid input, max length is 10\x1b[0m\n  label: "
        Test:       	TestForm/Errors
--- FAIL: TestSummary (0.00s)
    --- FAIL: TestSummary/SummarizeMetrics (0.00s)
        --- FAIL: TestSummary/SummarizeMetrics/[avg_min_med_max_p(90)_p(95)_p(99.9)] (0.00s)
        Error Trace:    summary_test.go:75
        Error:      	Not equal:
                    	expected: "     █ child\n\n       ✗ check1\n        ↳  33% — ✓ 5 / ✗ 10\n\n   ✓ checks......: 100.00% ✓ 3   ✗ 0  \n   ✗ http_reqs...: 3       3/s\n     my_trend....: avg=15ms min=10ms med=15ms max=20ms p(90)=19ms p(95)=19.5ms p(99.9)=19.99ms\n     vus.........: 1       min=1 max=1\n"
[...]

Issue #2797, opened 2022-12-05
Fix codecov covergae reports


olegbespalov:
# What?

It seems like our code coverage reports are going to the following:
https://app.codecov.io/github/k6io/k6/commits

instead of the

https://app.codecov.io/gh/grafana/k6/commits

And we also don't use an existing GitHub Action https://github.com/codecov/codecov-action

So, we should fix that.

# Why?

We no longer use the https://app.codecov.io/github/k6io/k6/commits

codebien:
Codecov has been removed in https://github.com/grafana/k6/pull/5008

Issue #2144, opened 2021-09-27
Cloud logs tests are flaky 


codebien:
We are getting some failing CI runs caused by the logs tailing tests. They are generated from the [time.Sleep use](https://github.com/grafana/k6/blob/75d30e6aa18455ce31fd8ae0fba2b92df30e421e/cloudapi/logs_test.go#L327) that is not deterministic.

Historically, we decided to not add a mock library just to support those tests. Instead, the code is using [HTTPMultiBin](https://github.com/grafana/k6/blob/75d30e6aa18455ce31fd8ae0fba2b92df30e421e/lib/testutils/httpmultibin/httpmultibin.go#L90) that has the network dependency, so Sleep is waiting for an approximate amount of time necessary for flushing messages across the WebSocket connection, but not always the used duration value is enough.

The suggestion here is to find a more deterministic approach for detecting when the sent message from the HTTPMultiBin server has been received from the client.

inancgumus:
@imiric:

```
We have a couple of these: https://github.com/grafana/k6/search?q=flaky&type=issues

We used to have a single issue for all (https://github.com/grafana/k6/issues/1357), which I think would still make sense to consolidate all of them.

Mocking time would help (https://github.com/grafana/k6/issues/1986), but for a lot of these, we need a non-time-based way of making assertions, which is tricky in most cases. :confused:
```

@inanc:
<img width="1195" alt="screen_shot_2021-09-30_at_11 01 40_am" src="https://user-images.githubusercontent.com/621232/135420417-8e3fe703-53ed-4f60-a8e4-2b2a6135cc43.png">

<img width="1208" alt="screen_shot_2021-09-30_at_11 01 28_am" src="https://user-images.githubusercontent.com/621232/135420440-954c7107-488c-4998-93b5-af68d43387f9.png">

na--:
TestStreamLogsToLogger [seems to also be flaky](https://github.com/grafana/k6/runs/3872217805?check_suite_focus=true):
```
--- FAIL: TestStreamLogsToLogger (0.00s)
    --- FAIL: TestStreamLogsToLogger/RestoreConnFromLatestMessage (5.00s)
        logs_test.go:333: 
            	Error Trace:	logs_test.go:333
            	            				logs_test.go:225
            	            				server.go:2056
            	            				server.go:2434
            	            				server.go:2888
            	            				server.go:1938
            	            				asm_amd64.s:1446
            	Error:      	Not equal: 
[...]

PR #3528, opened 2024-01-11
Run CI (tests) on MacOS

joanlopez:
## What?

It enables `macos-latest` for test-related (`test.yml`) CI workflows.

## Why?

Ha, why not? Jokes aside, imho now that MacOS-related issues (https://github.com/grafana/k6/issues/2578 and https://github.com/grafana/k6/issues/1230) have been closed (fixed), and that it looks like MacOS runners have at the very least a performance equivalent to Windows runners' (in fact, [test jobs seems to be taking less time in MacOS runners](https://github.com/grafana/k6/actions/runs/7493056548)), I guess it's good idea to make the check of the good functioning of k6 on MacOS part of the CI checks.

Also, in the worst case we can disable it again, but why not giving it a try?

## Checklist

## Related PR(s)/Issue(s)

I haven't found any.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -37,7 +37,7 @@ jobs:
           go version
           export GOMAXPROCS=2
           args=("-p" "2" "-race")
-          # Run with less concurrency on Windows to minimize flakiness.
+          # Run with less concurrency on Windows/MacOS to minimize flakiness.
           if [[ "${{ matrix.platform }}" == windows* || "${{ matrix.platform }}" == macos* ]]; then
             unset args[2]
             args[1]="1"
@@ -75,7 +75,7 @@ jobs:
           go version
           export GOMAXPROCS=2
           args=("-p" "2" "-race")
-          # Run with less concurrency on Windows to minimize flakiness.
+          # Run with less concurrency on Windows/MacOS to minimize flakiness.
           if [[ "${{ matrix.platform }}" == windows* || "${{ matrix.platform }}" == macos* ]]; then
             unset args[2]
             args[1]="1"
@@ -103,7 +103,7 @@ jobs:
           go version
           export GOMAXPROCS=2
           args=("-p" "2" "-race")
-          # Run with less concurrency on Windows to minimize flakiness.
+          # Run with less concurrency on Windows/MacOS to minimize flakiness.
           if [[ "${{ matrix.platform }}" == windows* || "${{ matrix.platform }}" == macos* ]]; then
             unset args[2]
             args[1]="1"
@@ -126,6 +126,9 @@ jobs:
           CODECOV_BASH_SHA512SUM: d075b412a362a9a2b7aedfec3b8b9a9a927b3b99e98c7c15a2b76ef09862aeb005e91d76a5fd71b511141496d0fd23d1b42095f722ebcd509d768fba030f159e
           CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
         run: |
+          if [ "$(uname)" = "Darwin" ]; then
+            alias sha256sum='shasum -a 256'
+          fi
           curl -fsSLO "https://raw.githubusercontent.com/codecov/codecov-bash/${CODECOV_BASH_VERSION}/codecov"
           echo "$CODECOV_BASH_SHA512SUM codecov" | sha512sum -c -
           platform="${{ matrix.platform }}"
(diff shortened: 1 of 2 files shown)


js/runner_test.go
```diff
@@ -765,7 +765,7 @@ func TestVURunInterruptDoesntPanic(t *testing.T) {
 					assert.Contains(t, vuErr.Error(), "context canceled")
 				}()
 				<-ch
-				time.Sleep(time.Microsecond * 1) // NOTE: increase this in case of problems ;)
+				time.Sleep(time.Millisecond * 1) // NOTE: increase this in case of problems ;)
```
mstoykov, line 768: nit: Could we increase it by less than 1000x :)

mstoykov: Looking at the test - this doesn't really matter, so it is probably okay. I was thinking it is one of those cases where we are trying to tease out a race, but we literally want for `RunOnce` to actually start.

joanlopez: but yeah, I agree would be good to keep it as close to its original value as possible, ofc! 
although.. I also wonder, do we really care about 1 ms? haha

.github/workflows/test.yml
```diff
@@ -126,8 +126,12 @@ jobs:
           CODECOV_BASH_SHA512SUM: d075b412a362a9a2b7aedfec3b8b9a9a927b3b99e98c7c15a2b76ef09862aeb005e91d76a5fd71b511141496d0fd23d1b42095f722ebcd509d768fba030f159e
           CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
         run: |
+          if [[ "${{ matrix.platform }}" == macos* ]]; then
+            shopt -s expand_aliases
+            alias sha512sum='shasum -a 512'
+          fi
           curl -fsSLO "https://raw.githubusercontent.com/codecov/codecov-bash/${CODECOV_BASH_VERSION}/codecov"
-          echo "$CODECOV_BASH_SHA512SUM codecov" | sha512sum -c -
+          echo "$CODECOV_BASH_SHA512SUM  codecov" | sha512sum -c -
```
joanlopez, line 134: For some reason (I didn't fully get), without the extra space it doesn't work on MacOS, but with it, it works correctly in all platforms 🤷🏻

js/runner_test.go
```diff
@@ -765,7 +765,7 @@ func TestVURunInterruptDoesntPanic(t *testing.T) {
 					assert.Contains(t, vuErr.Error(), "context canceled")
 				}()
 				<-ch
-				time.Sleep(time.Millisecond * 1) // NOTE: increase this in case of problems ;)
+				time.Sleep(time.Microsecond * 1) // NOTE: increase this in case of problems ;)
```
joanlopez, line 768: Why not reducing it if it works correctly? In fact, I've seen the flakiness induced by this test is lower with a lower value 🤷🏻

.github/workflows/test.yml
olegbespalov: I remember that we had issues with `codecov` (which later were resolved) and created [the task](https://github.com/grafana/k6/issues/2797), but one of the highlights was that `codecov` has the GH action. So maybe it's worth migrating to it

.github/workflows/test.yml
```diff
@@ -37,8 +37,8 @@ jobs:
           go version
           export GOMAXPROCS=2
           args=("-p" "2" "-race")
-          # Run with less concurrency on Windows to minimize flakiness.
-          if [[ "${{ matrix.platform }}" == windows* ]]; then
+          # Run with less concurrency on Windows/MacOS to minimize flakiness.
```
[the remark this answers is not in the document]
olegbespalov: The issue is that tests are flaky on all platforms, primarily GRPC-TLS related https://github.com/grafana/xk6-grpc/issues/39. Which ones do you constantly see?

.github/workflows/test.yml
joanlopez: Oh, nice! Thanks for letting me know, I could consider working on that right after, but I'd prefer to keep it for a separate PR, if that sounds good to you as well 🙏🏻

.github/workflows/test.yml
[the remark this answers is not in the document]
joanlopez: Yeah, there's nothing new, as I think all (or most) of them are documented at https://github.com/grafana/k6/issues/2144.

It's just that it seems that Windows and MacOS runners are way more likely to fail (it's probably a matter of their performance), so I decided to add MacOS into that exception we already had for Windows, to avoid adding more flakiness.

olegbespalov (approved):

mstoykov (approved):

PR #3535, opened 2024-01-11
Expose runtime Go metrics

### LINKED ISSUE #2955: Expose runtime Go metrics
### Feature Description

While running a k6 test, there is no easy way to understand how the garbage collector behaves, how many goroutines exist, or how many file descriptors the k6 process uses. 

These metrics can be critical to spot problems and further optimize scripts, extensions, and k6 core.

Given that k6 already has a web server, one option would be to initialize the Go Prometheus instrumentation and expose the default metrics (that cover all these core runtime metrics) on `/metrics`.

If we don't want to enable this behavior by default, we could require an environment variable to be passed.

https://github.com/grafana/k6/issues/1692 also tries to make k6 more observable.

joanlopez:
## What?

It exposes the `/metrics` endpoint when `--profiling-enabled`, with the Go runtime metrics from https://github.com/prometheus/client_golang.

## Why?

As stated in #2955, these metrics can be critical to spot problems and further optimize scripts, extensions, and k6 core. It comes almost by free (as part of `github.com/prometheus/client_golang` is already an -_indirect_- dependency of the project), and in the future we could even introduce our own custom metrics, like the amount of VUs, ongoing HTTP requests, etc.

## Checklist

## Related PR(s)/Issue(s)

Closes #2955


api/server.go
```diff
@@ -42,6 +43,7 @@ func injectProfilerHandler(mux *http.ServeMux, profilingEnabled bool) {
 	}
 
 	mux.Handle("/debug/pprof/", handler)
+	mux.Handle("/metrics", promhttp.Handler())
```
joanlopez, line 46: I have reused the `--profiling-enabled` flag, but I'm also happy with any other, separate, suggestion.

My rationale is that the potential user or use case for both probably overlaps most of the cases, like trying to understand how k6 behaves in terms of resource utilization, etc.

oleiade (dismissed):
Looks good to me, thanks a lot for acting on this, much appreciated 🙇🏻 🚀

mstoykov:
> as part of github.com/prometheus/client_golang is already an -indirect- dependency of the project

This is currently through due to us having an experimental prometheus remote-write output. But it is not certain it will stay or will we move fully to oTEL as such I would argue at some point it won't be free at all.

I also really do prefer to just use stdlib options where possible and golang does have expvar and (after some unsuccesful tries)

```
       mux.Handle("/debug/vars/", expvar.Handler())
```

Seems to does the job.

joanlopez:
TBH, I'm not sure about the usability of _expvar_ compared to _promhttp_, nor what will be "the new standard" in the near/mid future, but I agree that, considering this has been there for a year without much claim, it's better to start simple and avoid adding new dependencies, if we have an alternative way to expose such data in case we need them for profiling/debugging.

mstoykov (approved):

PR #3544, opened 2024-01-16
Add Dashboard information to v0.49.0 release notes

oleiade:
## What?

This PR adds a web-dashboard dedicated section to the release notes.

⚠️ Currently the report image is far to big, but we have opened https://github.com/grafana/xk6-dashboard/pull/144 to add a smaller screenshot, and will make sure to replace the current screenshot with that one before merging 👍🏻


release notes/v0.49.0.md
```diff
@@ -1,5 +1,6 @@
 k6 `v0.49.0` is here 🎉! This release includes:
 
+- a web-dashboard enabling users to track their load test results in real-time, directly in their web browser
```
olegbespalov, line 3: I'd suggest making this shorter, keeping in mind the context of the list
```suggestion
- a built-in web-dashboard for test run
```

release notes/v0.49.0.md
```diff
@@ -17,6 +18,33 @@ k6 `v0.49.0` is here 🎉! This release includes:
...
+![k6 Web Dashboard Overview](https://github.com/grafana/xk6-dashboard/blob/master/screenshot/k6-dashboard-overview-light.png?raw=true)
+
+#### Test report
+
+The web dashboard also offers an HTML test report for detailed analysis, enabling easy sharing and downloading capabilities for
+collaboration.
+
+To access and download the report, click on the `Report` button located in the top right corner of the dashboard.
+
+![HTML Report Generated by k6 Dashboard](https://github.com/grafana/xk6-dashboard/blob/master/screenshot/k6-dashboard-html-report.png?raw=true)
```
olegbespalov, line 46: TBH, I'm wondering if both images should be embedded. For instance, the second image could remain a link, and customers could open and see a complete example of the report in a separate tab/window.

oleiade: I was wondering the same, I don't think I have a strong take on either way, although I find that this feature is very visual, and giving a visual clue is a good call to action. Maybe we keep the first image and keep the second as link instead?

olegbespalov: > Maybe we keep the first image and keep the second as link instead?

Yes, that's exactly what I was wondering, too :+1: It's essential to attract the attention of our customers, but I believe if both images are embedded, it could overload attention. 

So :+1: for keeping the first image as it is and moving the second to a link, something like _an example of the report you could find here_ or a better sugegstion

oleiade: Perfect wilco 👍🏻

release notes/v0.49.0.md
oleiade: I think you're right, it's probably too long, I've given a shot to rephrase it to a shorter form indeed: "- a built-in web-dashboard displaying test results in real-time".

olegbespalov: It also works too :+1:

olegbespalov (approved):

mstoykov (approved):

PR #3545, opened 2024-01-16
Experimental

joanlopez:
## Checklist


olegbespalov:
I don't mind increasing the timeout of the server, but unfortunately, I bet it doesn't fix the issue of the flaky [test that you've mentioned](https://github.com/grafana/k6/actions/runs/7539030179/job/20520691770).

Long story short, the cause of the issue is not in our code base but in the GRPC package, which has no guarantee that `connect` will always return the same error since many concurrent things are happening inside. 

For instance, if we look closer at the logs of the CI job, we see that the expected message is in CI output, but it wasn't propagated by grpc, and as a result, the test fails with context ended since it simply can't wait any longer.

image

I even [opened an issue to the grpc-go](https://github.com/grpc/grpc-go/issues/6593) and planned to work on it but failed since it has no capacity and a good understanding of the internals of GRPC.

olegbespalov (approved):

joanlopez:
> I don't mind increasing the timeout of the server, but unfortunately, I bet it doesn't fix the issue of the flaky [test that you've mentioned](https://github.com/grafana/k6/actions/runs/7539030179/job/20520691770).
> ...
> ... 

Aha! I think that unfortunately you're right (see https://github.com/grafana/k6/actions/runs/7543208013/job/20546713199).

Thanks for the context and explanations!

olegbespalov:
> PS: Although I have to admit that I'm curious about why it's flaky, or even more, why it's almost impossible to make it fail locally?

I guess it's possible that we have to instruct tests with something like `go test -count=1000 ...` :see_no_evil:

olegbespalov:
> Indeed, I feel like if we would increase both timeouts enough (client & server), it would never fail, because it seems to remain waiting [here](https://github.com/grpc/grpc-go/blob/ddd377f19841eae70862559c854d957d61b3b692/clientconn.go#L281) and then follow [this path](https://github.com/grpc/grpc-go/blob/master/clientconn.go#L283), which ends up returning the connection error.

I think the predominant problem is that the connection error is stored in the structure and [could be overwritten](https://github.com/grpc/grpc-go/blob/ddd377f19841eae70862559c854d957d61b3b692/clientconn.go#L1706), and keeping in mind that internally GRPC manages things concurrently we could meet not what we expected there.

GRPC team says that we shouldn't rely on these connection errors; it's even [an anti-pattern there](https://github.com/grpc/grpc-go/blob/master/Documentation/anti-patterns.md#using-failonnontempdialerror-withblock-and-withreturnconnectionerror). I believe they are considering them from the customer perspective, but from our testing/load testing perspective having a chance to catch this connection errors is useful.

joanlopez:
> I think the predominant problem is that the connection error is stored in the structure and [could be overwritten](https://github.com/grpc/grpc-go/blob/ddd377f19841eae70862559c854d957d61b3b692/clientconn.go#L1706), and keeping in mind that internally GRPC manages things concurrently we could meet not what we expected there. This theory is proven by the fact that we see the `bad certificate` error in the logs at the beginning.

Okay, I see now what you mean. So, `lastConnectionError` could be overwritten by `context deadline exceeded`, even if there was a `bad certificate` there before. In such case, I guess it makes no sense to try to merge this PR, as increasing the timeout just makes the test slower (cause the clients waits till the context deadline is exceeded) without increasing the chances to success.

> GRPC team says that we shouldn't rely on these connection errors; it's even [an anti-pattern there](https://github.com/grpc/grpc-go/blob/master/Documentation/anti-patterns.md#using-failonnontempdialerror-withblock-and-withreturnconnectionerror). I believe they are considering them from the regular customer perspective, but from our testing/load testing perspective having a chance to catch this connection errors is useful.

Yeah, I get that error generalization is tricky, and for instance the reason why `Temporary` and `Timeout` have been largely discussed (and deprecated/removed) in Go. However, I think that ideally, in this case, the `tls` package should return a [`tls.CertificateVerificationError`](https://pkg.go.dev/crypto/tls#CertificateVerificationError), so we could do something similar to what you suggested for `grpc-go` but specifically for certificate errors (so we fail fast in such case, what I think would make sense).

However, it's not as easy as that because although the returned error models a [`badCertificate`](https://cs.opensource.google/go/go/+/refs/tags/go1.21.6:src/crypto/tls/alert.go;l=35), there's no proper way to assert that (like with `errors.As/Is`) and act consequently, and achieving that is either impossible or would take too much (way too out of the scope of our project).

olegbespalov:
My issue is that this test case is a bit artificial. When I was tried to investigate I found that that kind of the error we're getting when we try to use a valid certificate, but that certificate issued for the email signing. 

I also believe that in that case, we are actually trying to test not k6's functionality but rather TLS or grpc, with a notice that we are "proxying" the errors back to the client. Keeping in mind the comments from above, maybe it's better just to remove that test case and make a comment in `connect` documentation that, for some cases, we can't guarantee that an error should be returned, and it's better to handle errors coming from Invoke.

PR #3547, opened 2024-01-17
github: Run the triage process for comments

codebien:
## What?

Run the Triage workflow if a comment has been added to an issue without an assignee. It adds an event as described here https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#issue_comment

## Why?

It makes sure comments are not lost and we can react in case needed.

### DIFF
--- a/.github/workflows/issue-auto-assign.yml
+++ b/.github/workflows/issue-auto-assign.yml
@@ -13,6 +13,8 @@ permissions:
 jobs:
   assign-user:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


mstoykov (approved):

oleiade (approved):

codebien:
Sorry for dismissing the reviews but I noticed the condition could be not so clear for the next code reader so I added a comment.

mstoykov (approved):

oleiade (approved):

PR #3564, opened 2024-01-24
cloudapi: fix lint issues

mstoykov:
This has a breaking change in the API as it renames an error type.

This should not really be used by anything, but tests :crossed-fingers:

works on #769

### DIFF
--- a/cloudapi/client.go
+++ b/cloudapi/client.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"encoding/hex"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"math/rand"
@@ -70,7 +71,7 @@ func (c *Client) NewRequest(method, url string, data interface{}) (*http.Request
 		buf = bytes.NewBuffer(b)
 	}
 
-	req, err := http.NewRequest(method, url, buf)
+	req, err := http.NewRequest(method, url, buf) //nolint:noctx // the user can add this
 	if err != nil {
 		return nil, err
 	}
@@ -78,6 +79,7 @@ func (c *Client) NewRequest(method, url string, data interface{}) (*http.Request
 	return req, nil
 }
 
+// Do is simpler to http.Do but also unmarshals the response in hte provided v
 func (c *Client) Do(req *http.Request, v interface{}) error {
 	if req.Body != nil && req.GetBody == nil {
 		originalBody, err := io.ReadAll(req.Body)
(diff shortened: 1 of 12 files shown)


cloudapi/client.go
```diff
@@ -70,14 +71,15 @@ func (c *Client) NewRequest(method, url string, data interface{}) (*http.Request
...
-	req, err := http.NewRequest(method, url, buf)
+	req, err := http.NewRequest(method, url, buf) //nolint:noctx // the user can add this
 	if err != nil {
 		return nil, err
 	}
 
 	return req, nil
 }
 
+// Do is simpler to http.Do but also unmarshals the response in hte provided v
```
codebien, line 82:
```suggestion
// Do is simpler to http.Do but also unmarshals the response in the provided v
```

cloudapi/client.go
```diff
@@ -70,14 +71,15 @@ func (c *Client) NewRequest(method, url string, data interface{}) (*http.Request
 		buf = bytes.NewBuffer(b)
 	}
 
-	req, err := http.NewRequest(method, url, buf)
+	req, err := http.NewRequest(method, url, buf) //nolint:noctx // the user can add this
```
codebien, line 74:
```suggestion
	req, err := http.NewRequest(method, url, buf) //nolint:noctx // the user can add this
```
What do you mean by _the user_? It means in the sense of the caller of the function?

cloudapi/errors.go
```diff
@@ -13,8 +13,8 @@ var (
 	errUnknown          = errors.New("an error occurred communicating with k6 Cloud")
 )
 
-// ErrorResponse represents an error cause by talking to the API
-type ErrorResponse struct {
+// ResponseError represents an error cause by talking to the API
+type ResponseError struct {
```
codebien, line 17: @yorugac If I'm not wrong, the operator relies on this pkg. Does this change break anything? And if it does, is it easy to fix?

codebien (changes requested):

cloudapi/client.go
mstoykov: Yeah -this is basically a helping method on top of http.NewRequest.

cloudapi/client.go
mstoykov: hmmm ... usually I get those reported - somethign isn't working again :facepalm:

cloudapi/errors.go
mstoykov: It doesn't look as it is used https://github.com/search?q=repo%3Agrafana%2Fk6-operator%20ErrorResponse&type=code

And neither is it in anything else by the looks of it.

I am okay with reverthing this if we want? I can just put in a nolint

codebien (approved):

olegbespalov (approved):

PR #3586, opened 2024-02-09
fix zip-slip noise in tests

olegbespalov:
## What?

This attempts to fix the "noise" from the security tab (CodeQL reported).

## Why?

Fixing such cases helps us focus on the right things.

## Checklist

### DIFF
--- a/lib/testutils/untar.go
+++ b/lib/testutils/untar.go
@@ -38,9 +38,12 @@ func Untar(t *testing.T, fileSystem fsext.Fs, fileName string, destination strin
 			continue
 		}
 
-		// as long as this code in a test helper, we can safely
-		// omit G305: File traversal when extracting zip/tar archive
-		target := filepath.Join(destination, header.Name) //nolint:gosec
+		fileName := header.Name
+		if !filepath.IsLocal(fileName) {
+			return errors.New("tar file contains non-local file names")
+		}
+
+		target := filepath.Join(destination, filepath.Clean(fileName))
 
 		switch header.Typeflag {
 		case tar.TypeDir:


lib/testutils/untar.go
```diff
@@ -38,9 +38,12 @@ func Untar(t *testing.T, fileSystem fsext.Fs, fileName string, destination strin
...
 
-		// as long as this code in a test helper, we can safely
-		// omit G305: File traversal when extracting zip/tar archive
-		target := filepath.Join(destination, header.Name) //nolint:gosec
+		fileName := header.Name
+		if !filepath.IsLocal(fileName) {
+			return errors.New("tar file contains non-local file names")
+		}
+
+		target := filepath.Join(destination, filepath.Clean(fileName))
```
mstoykov, line 46: nit: Pretty sure there is no need to call Clean before Join

mstoykov (approved):
LGTM in general.

I don't think we have ever been fulnerable to this as this more or less only goes in memory, but I guess some shenanigans were still possible with `../https/hostname.com/path` .. but I still have no idea why that will be useful :shrug: 

I personally do not want to touch code that plays with `filepath` as unfortunately it is quite ... funky with us having paths from windows but running on linux and vice-versa.

This seems to not break any tests, and given that `filepath.Join` already does `Clean` (which is by far the thing that IIRC breaks most stuff) - hopefully nothing new will break :crossed_fingers:

lib/testutils/untar.go
olegbespalov: Having no `filepath.Clean` here will trigger golang-ci lint (see the omitting lines).

olegbespalov:
First, I share with you the opinion that the code is vulnerable for the reasons you mention (like in memory).

>but I still have no idea why that will be useful 🤷

Still, the primary purpose of doing this for me is to fix the `Security` tab noise. I'd like to have these 0 issues as the standard state. And if something appears there, it's a signal for checking. And for sure, I don't want to make calculations like there are two known, so when it will be 3, I need to check. Using the simple flow, 0, or anything else is better.

Now, we have two options to reach this state: omit it inline or try fixing it. In choosing between these two options, I feel that it's worth trying to go with the fixing first instead of just omitting inline.

joanlopez (approved):

PR #3624, opened 2024-03-06
Add/browser 1112 to release notes

inancgumus:
## What?

Add/browser 1112 to release notes

### DIFF
--- a/release notes/v0.50.0.md
+++ b/release notes/v0.50.0.md
@@ -28,6 +28,10 @@ _what, why, and what this means for the user_
 - [browser#1197](https://github.com/grafana/xk6-browser/pull/1197), [browser#1202](https://github.com/grafana/xk6-browser/pull/1202), [browser#1203](https://github.com/grafana/xk6-browser/pull/1203), [browser#1221](https://github.com/grafana/xk6-browser/pull/1221) add the ability to upload screenshots to a remote location.
 - [browser#1209](https://github.com/grafana/xk6-browser/pull/1209) add a shadow DOM usage example.
 
+### Browser Context Isolation [browser#1112](https://github.com/grafana/xk6-browser/issues/1112)
+
+With this release, we have overhauled and improved the performance and stability of the browser module. It's now possible to run tests with unlimited VUs concurrently without any performance issues. Previously, when running tests with multiple VUs concurrently, each VU shared the pages opened by other VU's browser contexts. This was leading to unexpected behavior and performance issues.
+
 ## Bug fixes
 
 - [browser#1215](https://github.com/grafana/xk6-browser/pull/1215) fixes a data race during logging that panics.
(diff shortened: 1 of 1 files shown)


release notes/v0.50.0.md
```diff
@@ -28,6 +28,10 @@ _what, why, and what this means for the user_
 - [browser#1197](https://github.com/grafana/xk6-browser/pull/1197), [browser#1202](https://github.com/grafana/xk6-browser/pull/1202), [browser#1203](https://github.com/grafana/xk6-browser/pull/1203), [browser#1221](https://github.com/grafana/xk6-browser/pull/1221) add the ability to upload screenshots to a remote location.
 - [browser#1209](https://github.com/grafana/xk6-browser/pull/1209) add a shadow DOM usage example.
 
+### Browser Context Isolation [browser#1112](https://github.com/grafana/xk6-browser/issues/1112)
+
+With this release, we have overhauled and improved the performance and stability of the browser module. It's now possible to run tests with unlimited VUs concurrently without any performance issues. Previously, when running tests with multiple VUs concurrently, each VU shared the pages opened by other VU's browser contexts. This was leading to unexpected behavior and performance issues.
```
ankur22, line 33: I feel that the use of unlimited isn't quite right. This was also the case before the fix but was bound by the hardware and the implementation, which is still the same scenario after the fix. I think a "larger number" (or something along those lines) sounds more accurate.

```suggestion
With this release, we have overhauled and improved the performance and stability of the browser module. It's now possible to run tests with a larger number of VUs concurrently without any performance issues. Previously when running tests with multiple VUs concurrently, each VU's browser context would attach to the pages from other VU's browser context. This was leading to unexpected behavior and performance issues.
```

ankur22 (approved):

release notes/v0.50.0.md
inancgumus: > I feel that the use of unlimited isn't quite right. This was also the case before the fix but was bound by the hardware and the implementation

Yes, but it wasn't possible to run multi-VU tests beyond some point (like 10-20 VUs), even with beefy hardware.

So, WDYT?

```suggestion
With this release, we have overhauled and tremendously improved the performance and stability of the browser module. It's now possible to run tests with a larger number of VUs concurrently without any performance issues. Previously, when running tests with multiple VUs concurrently, each VU's browser context would attach to the pages from the other VUS' browser contexts. This led to unexpected behavior and performance issues, preventing multi-VU tests from running after some point.
```

ankur22: I'm not sure about the last addition "preventing to run multi-VU tests to an extent". Could we add some numbers instead e.g.:

_On a machine with X CPU and Y Ram we used to be able to run n VUs using the `fillform.js` test, but now we can run m VUs on the same hardware._

```suggestion
With this release, we have overhauled and (tremendously) improved the performance and stability of the browser module. It's now possible to run tests with a larger number of VUs concurrently without any performance issues. Previously, when running tests with multiple VUs concurrently, each VU's browser context would attach to the pages from the other VUS' browser contexts. This led to unexpected behavior and performance issues, preventing to run multi-VU tests to an extent.
```

inancgumus: That makes sense. Since I'll come up with some numbers soon, let's merge this in, then I might amend this.

PR #3639, opened 2024-03-13
sse: support Server Sent Event

### LINKED ISSUE #746: HTTP SSE (Server-Sent Event) support for k6
Hi, 

Server-Sent Events (SSE) : a lightweight and standardized technology for pushing notifications from a HTTP server to a HTTP client (One way direction notification).

https://fr.wikipedia.org/wiki/Server-sent_events

I'd like to add SSE support in k6 and I have questions about this task:

- Anyone already working on this feature ? 
- Any directions for the implementation ? 

Thanks 

Olivier

phymbert:
## What?

Server-Sent event http client support as an experimental module.

## Why?

[Server sent events ](https://fr.wikipedia.org/wiki/Server-sent_events) are more and more used nowadays and I would love k6 to support it.

## Checklist

## Related PR(s)/Issue(s)

- #746

Close #746

### DIFF
--- a/docs/design/021-sse-api.md
+++ b/docs/design/021-sse-api.md
@@ -1,0 +1,59 @@
+# Introduce an SSE API module for k6
+
+|                      |                                                              |
+|:---------------------|:-------------------------------------------------------------|
+| **author**           | @phymbert                                                    |
+| **status**           | 🔧 proposal                                                  |
+| **revisions**        | [initial](https://github.com/grafana/k6/pull/3639)           |
+| **Proof of concept** | [branch](https://github.com/phymbert/k6/tree/hp/feature/sse) |
+| **references**       | [#746](https://github.com/grafana/k6/issues/746)             |
+
+## Problem definition
+
+The current version of k6 reads the full http response body before returning to the client,
+which make impossible testing [Server-Sent Event](https://fr.wikipedia.org/wiki/Server-sent_events). 
+
+We propose to introduce a new `sse` module.
+This module is intended to offer an intuitive and user-friendly API for SSE interactions within k6 scripts.
+
+## Proposed solution
+
+We suggest to implement a minimalist, experimental (`sse`) module based on the go http client.
+The new module will allow users to interact with SSE.
+The module will provide an `open` which will allow the user to pass a setup function to configure an `event` callback as it is done in the `ws` module with `message`.
+
+### Limitation
+The module will not support async io and the javascript main loop will be blocked during the http request duration.
+
+### Example usage
+
+```javascript
+import sse from 'k6/experimental/sse';
+
+var url = "https://echo.websocket.org/.sse";
+var params = {"tags": {"my_tag": "hello"}};
+
+var response = sse.open(url, params, function (client) {
+    client.on('open', function open() {
+        console.log('connected');
+    });
+
+    client.on('event', function (event) {
+        console.log(`event id=${event.id}, name=${event.name}, data=${event.data}`);
+    });
+
+    client.on('close', function close() {
+        console.log('disconnected');
+    });
+
+    client.on('error', function (e) {
+        console.log('An unexpected error occurred: ', e.error());
+    });
+});
+
+check(response, {"status is 200": (r) => r && r.status === 200});
+```
+
+### Conclusion
+
+We believe the [proof of concept](https://github.com/grafana/k6/blob/d5cd1010ecb2381376188c8a47ab861cf8b5dc3d/js/modules/k6/experimental/sse/sse.go) developed with this proposal illustrates the feasibility and benefits of developing such an API.
--- a/docs/design/021-sse-api.md
+++ b/docs/design/021-sse-api.md
@@ -42,10 +42,6 @@ var response = sse.open(url, params, function (client) {
         console.log(`event id=${event.id}, name=${event.name}, data=${event.data}`);
     });
 
-    client.on('close', function close() {
-        console.log('disconnected');
-    });
-
     client.on('error', function (e) {
         console.log('An unexpected error occurred: ', e.error());
     });
--- a/examples/sse.js
+++ b/examples/sse.js
@@ -14,10 +14,6 @@ export default function () {
 			console.log(`event id=${event.id}, name=${event.name}, data=${event.data}`);
 		});
 
-		client.on('close', function close() {
-			console.log('disconnected');
-		});
-
 		client.on('error', function (e) {
 			console.log('An unexpected error occurred: ', e.error());
 		});
--- a/examples/sse.js
+++ b/examples/sse.js
@@ -1,0 +1,27 @@
+import sse from "k6/experimental/sse";
+import {check} from "k6";
+
+export default function () {
+	var url = "https://echo.websocket.org/.sse";
+	var params = {"tags": {"my_tag": "hello"}};
+
+	var response = sse.open(url, params, function (client) {
+		client.on('open', function open() {
+			console.log('connected');
+		});
+
+		client.on('event', function (event) {
+			console.log(`event id=${event.id}, name=${event.name}, data=${event.data}`);
+		});
+
+		client.on('close', function close() {
+			console.log('disconnected');
+		});
+
+		client.on('error', function (e) {
+			console.log('An unexpected error occurred: ', e.error());
+		});
+	});
+
+	check(response, {"status is 200": (r) => r && r.status === 200});
+};
--- a/js/modules/k6/experimental/sse/sse.go
+++ b/js/modules/k6/experimental/sse/sse.go
@@ -95,9 +95,10 @@ type HTTPResponse struct {
 
 // Event represents a Server-Sent Event
 type Event struct {
-	ID   string
-	Name string
-	Data string
+	ID      string
+	Comment string
+	Name    string
+	Data    string
 }
 
 type sseOpenArgs struct {
@@ -364,13 +365,13 @@ func (c *Client) pushSSEMetrics(connStart, connEnd time.Time) func() {
 	}
 }
 
-// Wraps SSE in a channel
+// Wraps SSE in a channel, follow the SSE format described in:
+// https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events
 func (c *Client) readEvents(readChan chan Event, errorChan chan error, closeChan chan int) {
 	reader := bufio.NewReader(c.resp.Body)
 	ev := Event{}
 	var buf bytes.Buffer
 
-	sendEvent := false
 	for {
 		line, err := reader.ReadBytes('\n')
 		if err != nil {
@@ -392,17 +393,17 @@ func (c *Client) readEvents(readChan chan Event, errorChan chan error, closeChan
 		}
 
 		switch {
-		case hasPrefix(line, ":"):
-			// Comment, do nothing
-
-		case hasPrefix(line, "retry:"):
-			// Retry, do nothing for now
-
 		// id of event
 		case hasPrefix(line, "id: "):
 			ev.ID = stripPrefix(line, 4)
 		case hasPrefix(line, "id:"):
 			ev.ID = stripPrefix(line, 3)
+
+		// Comment
+		case hasPrefix(line, ": "):
+			ev.Comment = stripPrefix(line, 2)
+		case hasPrefix(line, ":"):
+			ev.Comment = stripPrefix(line, 1)
 
 		// name of event
 		case hasPrefix(line, "event: "):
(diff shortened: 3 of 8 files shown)


phymbert:
@mstoykov @oleiade Hello, this is a first version, but I would be happy to receive your early feedback to guide me to get this feature implemented. Thanks

oleiade:
Bringing @szkiba to this discussion, too, as I know [the dashboard feature](https://github.com/grafana/xk6-dashboard) is already built on top of SSE, and some code exists for supporting it there, too 🙇🏻

oleiade:
Hey @phymbert 👋🏻 

To clarify, I was bringing @szkiba to the discussion as he might have good insights into the topic, as he's built a whole extension that relies on SSE to provide the dashboard feature. Not to question neither the rationale nor the quality of this PR 🙇🏻 

The dashboard collects and aggregates metrics data from k6, and communicates them to the frontend part using SSE, I think the relevant files to look at for context are [`sse.go`](https://github.com/grafana/xk6-dashboard/blob/master/dashboard/sse.go),  [`aggregate.go`](https://github.com/grafana/xk6-dashboard/blob/master/dashboard/aggregate.go), and [`web.go`](https://github.com/grafana/xk6-dashboard/blob/master/dashboard/web.go).

I've approved the CI to run 👍🏻

phymbert:
Thanks @oleiade for the CI run and clarification 🙇🏻. I understand failing checks are not related to these changes ? Please tell me if I have to adapt something.

szkiba (commented):
As a rule, similar modules are first created as xk6 extensions. This way, anyone can use it and receive feedback. After that, if the topic is popular enough, the extension will become an experimental module. I would recommend this path here as well.

I would also consider using an SSE client library. There must be a compelling reason for the protocol implementation if there is a usable library (if it does not exist, then it should be justified why none of the existing libraries are suitable).

js/modules/k6/experimental/sse/sse.go
```diff
@@ -0,0 +1,540 @@
...
+					return
+				case <-c.done:
+					return
+				}
+			}
+		}
+
+		switch {
+		case hasPrefix(line, ":"):
+			// Comment, do nothing
```
szkiba, line 396: If we want a general solution, I think we can't ignore the comment. The user will test the SSE stream, which includes the comment...

js/modules/k6/experimental/sse/sse.go
```diff
@@ -0,0 +1,540 @@
...
+					Tags:   client.tagsAndMeta.Tags,
+				},
+				Time:     time.Now(),
+				Metadata: client.tagsAndMeta.Metadata,
+				Value:    1,
+			})
+
+			client.handleEvent("event", rt.ToValue(event))
+
+		case readErr := <-readErrChan:
```
szkiba, line 177: Errors should also be measured with metrics, as in the http module. Instead of handling errors at the JavaScript level, it is much more convenient to set a threshold for the error metric...

docs/design/021-sse-api.md
```diff
@@ -0,0 +1,55 @@
...
+This module is intended to offer an intuitive and user-friendly API for SSE interactions within k6 scripts.
+
+## Proposed solution
+
+We suggest to implement a minimalist, experimental (`sse`) module based on the go http client.
+The new module will allow users to interact with SSE.
+The module will provide an `open` which will allow the user to pass a setup function to configure an `event` callback as it is done in the `ws` module with `message`.
+
+### Limitation
+The module will not support async io and the javascript main loop will be blocked during the http request duration.
```
szkiba, line 26: Async support is becoming more and more important, so integrating a new module without async support is questionable. This is just my personal opinion, @mstoykov  is more competent in this...

examples/sse.js
```diff
@@ -0,0 +1,23 @@
+import sse from "k6/experimental/sse";
+import {check} from "k6";
+
+export default function () {
+	var url = "https://echo.websocket.org/.sse";
+	var params = {"tags": {"my_tag": "hello"}};
+
+	var response = sse.open(url, params, function (client) {
```
szkiba, line 8: An SSE connection is usually long-lived. We do not test normal operation by opening a new connection in each iteration.

phymbert: Not at all, see: https://platform.openai.com/docs/api-reference/streaming

I mainly use K6 at the moment to test LLM inference server:
 - https://github.com/ggerganov/llama.cpp/pull/5941

szkiba:
> Rules like this are also a blocker for the adoption of the tool.

I rather see such rules as helping to keep the code base stable.

> SSE is just a custom lines break on top of HTTP 1.1, I always prefer to stick on the go sdk.

go sdk is great i like it too. At the same time, I don't tend to reimplement every protocol, even the simplest ones, if there is a usable implementation.

> I do not see any understandable reason not to resolve the issue, adding a few lines of code, and letting people use SSE with k6.

The opinion of the maintainers of the k6 code base really matters here, that's why there is a code review.

> If you have relevant feedback on the code, I will integrate them, or else I keep the PR opened if people need it in the future.

I wrote a few comments, I don't know if they are relevant enough: metrics should be made about errors and not simply delegated to the JavaScript level, SSE comments should also be made available and not simply ignored, and async support should be considered.

phymbert:
@szkiba Thanks for your review, I will integrate the proposed changes.

> go sdk is great i like it too. At the same time, I don't tend to reimplement every protocol, even the simplest ones, if there is a usable implementation.

Do you have some ? I see one here:

https://github.com/r3labs/sse/blob/c6d5381ee3ca63828b321c16baa008fd6c0b4564/client.go#L322-L335

IMHO we do not need to depend on a whole library for a small function, although the proposed version here must be improved (trailing new lines, empty data).

@mstoykov regarding async io, could you please assist me with a code to look at it ?

oleiade:
Hello, everyone 👋🏻 

Firstly, I'd like to express our gratitude for the time and effort you've dedicated to implementing a missing and desirable feature for k6 with your proposal. We value contributions as such!

However, after discussions among the k6 maintainers' team, we've decided not to merge this PR in the core of k6 for several reasons outlined below, which aim to clarify our reasoning and rationale behind that decision.

### 1. HTTP module is implicitly frozen

We're currently focusing on developing a new `http` module. During this transitional period, we're limiting changes to the existing HTTP stack and feature-set to bug fixes. We acknowledge that we could have communicated this more effectively and publicly.

### 2. Extensions-first

As highlighted by @szkiba, this module is a perfect candidate for an xk6 extension: it allows for innovation, experimentation, and adding valuable APIs and features to k6 without impacting the stability of the core of k6, and relying on k6 code and constructs that are not accessible publicly (as far as I can tell). 

It probably needs to be communicated better at the moment, but as @szkiba outlined, our position is that anything that can start or remain an extension should probably be. We likely want to support SSE indeed, but for this reason, as well as those depicted in `1.` we would prefer to start having it as an extension. I'd be happy to provide support to get you started on that front should you want to do it too 🙇🏻 (Furthermore, we are also actively working on reducing the hurdles currently involved in using extensions in the future).

### 3. It's not just a few lines of code

Finally, the conversation around this PR has raised important points about the perceived simplicity of integrating new code versus the long-term commitments it entails.

>  I do not see any understandable reason not to resolve the issue, adding a few lines of code and letting people use SSE with k6.

And later:

> I don't plan to support any k6 extension for 1,5k lines of code.

And the latter is a good illustration of the reason for the former. It highlights the perceived simplicity of integrating new code versus the long-term commitments it entails. While it may appear as a trivial change to do, and for us to "just" merge these 1.5k lines of code, doing so would delegate the support and maintenance of these lines of code to us. 

We understand that this decision might be disappointing, but it's made with the best interest of the k6 open-source project and users in mind. k6 is a widely popular and adopted open-source project. Our goal and responsibility as maintainers is to ensure our software is easily maintainable and provides the best user experience, quality, and performance.

## Next steps

Regarding the next steps, I sincerely think this is valuable work. I recommend making an extension out of it so that users who wish to use SSE can do so in the meantime (as we make the HTTP module ready to receive further updates). So we can learn more about its design, implementation, and performance aspects. 

If extensions are not an option, another option would be to start a fork of k6 in the meantime. We make releases roughly every three months,  rarely break APIs, and it would be a low-effort task.

As I mentioned earlier, I'd be eager to support you if you were to start an extension and pave the way for improvements to it if that's something you'd find useful and valuable.

phymbert:
Thanks for the clear explanation @oleiade , I understand all the point you raised.

I suggest the following approach - please tell me if I miss something:

1. I finish the changes identified by @szkiba 
3. someone at GrafanaLabs creates a `grafana/xk6-sse` github repo
4. I am pleased to submit the first PR with your help, you can PM me if needed
2. close this pull request for archive
5. following the interest of the community and with your support, you can count on me to maintain the `xk6-sse` in the future

Highly appreciated your kind feedback here 🙇🏻

mstoykov:
Thanks for the change, but I do agree with @szkiba and @oleiade on this being an extension. I will try to give some code feedback either way. 

On top of the points by @oleiade, I will try to give more explanation on the topic of extensions and experimental modules.  


### ~Plugins~ Extensions

The whole idea of [extensions](https://github.com/grafana/k6/issues/1353) (for the core team specifically) from the get go was allowing users of k6 to create modules (and later outputs) that aren't supposed to be maintained and scrutinized by the k6 core team. 

This was particularly important when the team was smaller, but I would argue has become even more important as the scope has broadened. 

This both allows users (you) to develop and use their extensions, without needing to care about whether anyone on the k6 core team cares whether their style is nice, the protocol should be implemented this way or async is important or not(async wasn't a thing back then - but you get the point). It allows you to make a single (potentially, not very working) things that fixes your immediate need.

And just as importantly lets you continue to iterate on it outside and without regards for the k6 release cycle or anything related with it. Potentially letting you break and improve the extension over time. 

Also makes it possible (although not really relevant here) to implement 

### Experimental ~extensions~ modules

Experimental modules on the other hand are meant for things that the core team wants to be part of core more or less. But needs time to figure out what the API is and how it works with everything else. 

We used to make extensions first and then import them - but this ... isn't ideal from development perspective. So for some new experimental modules *we* are skipping it.

But one of the important requirements is that the core team will want this in the core - usually without seeing big problems with it. Just because a module is experimental doesn't mean that we want to introduce breaking changes all the time or that it doesn't need to match the same quality that is in k6 core already. 

As @oleiade said though - `k6/http` is in maintenance mode - we do make changes, but they usually need to be really small *and* (not *or*) really necessary. 

Point in case the last 2 "features" that have been added are: [asyncRequest](https://github.com/grafana/k6/pull/2877) which arguably makes asynchronous code useful and is a fairly small change, and support for [metadata](https://github.com/grafana/k6/pull/2726) which arguably is not even `k6/http` specific at all.

js/modules/k6/experimental/sse/sse.go
mstoykov: I kind of disagree on this - we do more or less the same for websockets. Especially as a lot of the cases it is just an error to close. 

This still allows users to create custom metrics for errors, but doesn't force them.

We do have error metrics when teh connection couldn't be started though.

Also not SSE expert - so maybe not as similar :shrug:

js/modules/k6/experimental/sse/sse.go
mstoykov: I do think that if there is a library that already parses this - it might be better to use it and if not - maybe split this as a library ;)

mstoykov:
> someone at GrafanaLabs creates a grafana/xk6-sse github repo

The actual idea is that you will create the repo and continue to maintain and own it.

Again big part of the extensions vs in core is about who *owns* and *maintains* the code. If it is a grafana repository we will now need to give you permission to commit and/or even worse review your commit and merge it without giving you any rights.

As pointed out - we are very unlikely to put this in core in the future. Any SSE solution will likely need to be based on http v2 api to be considered. And unfortunately that is blocked on ReadableStreams API.

> @mstoykov regarding async io, could you please assist me with a code to look at it ?

This arguably is a lot to explain, but a good example of something similar is [gRPC stream implementation](https://github.com/grafana/k6/blob/master/js/modules/k6/grpc/stream.go). And in general the documentation on [`RegisterCallback`](https://pkg.go.dev/go.k6.io/k6@v0.49.0/js/eventloop#EventLoop.RegisterCallback). (I technically should write docs on this - it just always is not prioritized ... :grimacing: )

Also general knowledge about extensions creating can be found in the [docs](https://grafana.com/docs/k6/latest/extensions/create/).

phymbert:
> The actual idea is that you will create the repo and continue to maintain and own it.

This is not what I understood from @oleiade, having an extension on my own will ridicously reduced its visibility. And again, users will just go to another solution like Gatling which supports [SSE very well](https://docs.gatling.io/reference/script/protocols/sse/) out of the box.

> If it is a grafana repository we will now need to give you permission to commit and/or even worse review your commit and merge it without giving you any rights.

Is it a problem to open a repo to other open-source contributor outside of GrafanaLabs ? is there any technical blocking point ? What are the difference with other extension like [https://github.com/grafana/xk6-websockets](https://github.com/grafana/xk6-websockets) ? You can as all open source project later on give `Collaborator` access to external users based on their contribution.

Thanks for the doc, I will read it out.

js/modules/k6/experimental/sse/sse.go
phymbert: Yes I follow the approach from the websockets module as it was proposed in:
- https://github.com/grafana/k6/issues/746#issuecomment-414150215

docs/design/021-sse-api.md
phymbert: What do you mean by asynchronous ? it's one way server to client data flow on top of HTTP 1.1 w/ single connection. The event loop is blocked while waiting for next event, but on each new event the client callback is triggered.

js/modules/k6/experimental/sse/sse.go
phymbert: @mstoykov the format is well described here:
https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events#event_stream_format

Pretty simple

oleiade:
Hi @phymbert 👋🏻 

Thank you for your patience. We've taken some further time to discuss and evaluate your proposal internally.

Unfortunately, we're unable to offer hosting the extension as part of the Grafana GitHub organization. Therefore, our suggestion remains unchanged, and we recommend you create a dedicated xk6 extension repository.

Regarding visibility, if that is something you'd be able, allowed, and willing to do, we often feature user and community stories on the Grafana Labs blog. We'd be interested in exploring the possibility of collaborating on a blog post to highlight your work with k6 and how the development of this extension served your requirements. Would that be something of interest to you?

🙇🏻

PR #3656, opened 2024-03-20
Add 'negative' value type for negative events

### LINKED ISSUE #2306: [Console output] http_req_failed : "succMark" & "failMark" seems inversed?
### Brief summary

```
     data_received..................: 7.2 MB 240 kB/s
     data_sent......................: 498 kB 17 kB/s
[...]
     http_req_failed................: 56.68% ✓ 403      ✗ 308
[...]
     http_reqs......................: 711    23.61168/s
```

I think ✓ and ✗ are inversed here. There should be more failed than succeed.

### k6 version

0.35.0

### OS

Windows 10

_No response_

Do a simple run of k6. Nothing fancy.

### Expected behaviour

```
http_req_failed................: 56.68% ✓ 308      ✗ 403
```

### Actual behaviour

```
http_req_failed................: 56.68% ✓ 403      ✗ 308
```

joanlopez:
## What?

It adds a new value type `negative` for those metrics that represent negative events, and uses it in the default `handleSummary` to display either ✓ or ✗ for the amount of passes/fails based on that.

## Why?

Now the `handleSummary` always displays ✓ for passes and ✗ for fails, but that's not accurate (and even confusing) for those metrics that represent negative events (like `http_req_failed`).

So, I decided to give a try to [@na--'s proposal](https://github.com/grafana/k6/issues/2306#issuecomment-997728913) on the reported issue, as a solution that can be used by any other `Rate` metric, not only by `http_req_failed`.

Also, because I feel that using a label instead of ✓ or ✗ might be more difficult to generalize. For instance, what would `fails` or `passes` (to give an example) mean in the context of `http_req_failed`? Or having to parameterize that (the labels) would start making the metric model a little bit more complex with no clear value.

## Checklist

## Related PR(s)/Issue(s)

Closes #2306

### DIFF
--- a/js/summary.js
+++ b/js/summary.js
@@ -230,12 +230,14 @@ function nonTrendMetricValueForSum(metric, timeUnit) {
         'min=' + humanizeValue(metric.values.min, metric, timeUnit),
         'max=' + humanizeValue(metric.values.max, metric, timeUnit),
       ]
-    case 'rate':
-      return [
-        humanizeValue(metric.values.rate, metric, timeUnit),
-        succMark + ' ' + metric.values.passes,
-        failMark + ' ' + metric.values.fails,
-      ]
+	  case 'rate':
+		  const passesMark = metric.contains === 'negative' ? failMark : succMark;
+		  const failsMark = metric.contains === 'negative' ? succMark : failMark;
+		  return [
+			  humanizeValue(metric.values.rate, metric, timeUnit),
+			  passesMark + ' ' + metric.values.passes,
+			  failsMark + ' ' + metric.values.fails,
+		  ]
     default:
       return ['[no data]']
   }
--- a/metrics/builtin.go
+++ b/metrics/builtin.go
@@ -87,7 +87,7 @@ func RegisterBuiltinMetrics(registry *Registry) *BuiltinMetrics {
 		GroupDuration: registry.MustNewMetric(GroupDurationName, Trend, Time),
 
 		HTTPReqs:              registry.MustNewMetric(HTTPReqsName, Counter),
-		HTTPReqFailed:         registry.MustNewMetric(HTTPReqFailedName, Rate),
+		HTTPReqFailed:         registry.MustNewMetric(HTTPReqFailedName, Rate, Negative),
 		HTTPReqDuration:       registry.MustNewMetric(HTTPReqDurationName, Trend, Time),
 		HTTPReqBlocked:        registry.MustNewMetric(HTTPReqBlockedName, Trend, Time),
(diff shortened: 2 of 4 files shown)


mstoykov:
Hi sorry for the slow reply :( 

I have 2 problems with this solution:
1. it makes metrics even more complicated adding stuff to them that are not really present in other metric presentations
2. makes the whole thing even more confusing if you actually use it for thresholds. And especially if you are used to this (which is the least problematic part). 

To me it seems like the best solution was for us to reverse this ... years ago after this was reported and have the old metric for a while.

~A bunch of people also seem to agree that having not checks/crosses will fix this as well and has no problems with the UI now changing depending on what the UI is.~

But that also depends on the above negative changes by the looks of it.


It is great someone tried to fix so we can see how it looks, but to me this doesn't seem like the correct solution.

I also would recommend showcasing how the UI changes when making changes to it. 

old:
image
this PR:
image

As an alternative solution making it obvious that is `X out of N` stuff seems to look okay to me.

image


Also opentelemtry has units as part of the specificaiton which seem like at thing we can (ab)use to get this and hte other "value type" under control, but arguably that should be part of a bigger refactor if we go down this route. One that likely should wait for some opentelemtry work to be done (cc @olegbespalov )

joanlopez:
> Hi sorry for the slow reply :(

Don't worry at all! :) 

<br/>
<br/>

To be honest, I don't have any strong preference, I opted for this approach because it seemed to me to be the one with greater acceptance, but it's true that this is difficult to measure. Also, because I felt that having this mechanism for defining whether a metric is positive or negative (or none) despite the unit, could be useful for other purposes. But it's also true that we don't have any immediate plans that justify that.

> As an alternative solution making it obvious that is `X out of N` stuff seems to look okay to me.

I think the question is.. do we really want that? 
Or do we really want to postpone this until a major refactor happens (as you said)?

If so, I can easily do it. But this looks like one of these situations where deciding _what_ is way more challenging than deciding _how_, and actually doing it.

mstoykov:
> I think the question is.. do we really want that?
Or do we really want to postpone this until a major refactor happens (as you said)?

I guess this is more a question for the issue? 

I will also paste this there. 

Arguably this solution has 2 possitives:
1. no code changes apart from visualization (potentially the same needs to happen for checks btw)
2. IMO it removes even more of ambiguoty as previously what one culture will take as a :heavy_check_mark: and one as :x: when we are talking about a mteric type such as Rate which counts falsy and truthy stuff is a bit questinable. Arguably this is the reason why it wasn't found to be so bad when first implemented.

codebien:
>I think the question is.. do we really want that?

I would prefer if we do it now. The text proposal sounds already a good solution.

PR #3660, opened 2024-03-21
Add NullSystemTagSet to correctly read K6_SYSTEM_TAGS from env

### LINKED ISSUE #3638: K6_SYSTEM_TAGS environment variable not being read
### Brief summary

The [system tags option](https://k6.io/docs/using-k6/k6-options/reference/#system-tags) is not being picked up when specified by the environment variable.

When using the CLI flag `--system-tags` or code `options.systemTags`, the system tags are being parsed and processed correctly. When using the `K6_SYSTEM_TAGS` environment variable option it reverts back to the default setting.

### k6 version

v0.49.0

### OS

macOS 14

_No response_

Run the following test script
```
import exec from 'k6/execution'

export default function () {
  console.log(exec.test.options.systemTags)
}
```
with the following commands
```
export K6_SYSTEM_TAGS=proto
k6 run test.js
```

### Expected behaviour

Console log is
```
["check","error","error_code","expected_response","group","method","name","proto","scenario","service","status","subproto","tls_version","url"]
```

### Actual behaviour

Console log should be 
```
["proto"]
```

(Running `k6 run test.js --system-tags=proto` produces this correct output)

joanlopez:
## What?

It adds `metrics.NullSystemTagSet` as a _nullable_ wrapper of `metrics.SystemTagSet`, following the same pattern we use from [`guregu/null`](https://github.com/guregu/null), and with some custom types like `types.NullHosts` in order to correctly read `K6_SYSTEM_TAGS` from env.

## Why?

Because now `options.SystemTags` is typed as `*metrics.SystemTagSet`, so as the default value for pointers is `nil`, it cannot ever be re-assigned, causing the call to `envconfig.Process(...)` to read the value set to `K6_SYSTEM_TAGS`, but without actually making any change to the options.

## Checklist

## Related PR(s)/Issue(s)

Closes #3638


metrics/system_tag.go
```diff
@@ -153,3 +153,91 @@ func (i *SystemTagSet) UnmarshalText(data []byte) error {
...
+
+// Map returns the EnabledTags with current value from SystemTagSet
+func (n NullSystemTagSet) Map() EnabledTags {
+	return n.Set.Map()
+}
+
+// SetString returns comma separated list of the string representation of all values in the set
+func (n NullSystemTagSet) SetString() string {
+	return n.Set.SetString()
+}
```
mstoykov, line 204: I take it that without this additional methods the change is much bigger ?

joanlopez: I might be missing something, but some of these methods are used widely (for instance, a quick search gives ~30 occurrences of `SystemTags.Has(...)`, so the only alternative I see is replacing each of them by: `SystemTags.Set.Has(...)`.

Do you have any other suggestion? 🤔

metrics/system_tag.go
```diff
@@ -153,3 +153,91 @@ func (i *SystemTagSet) UnmarshalText(data []byte) error {
 	}
 	return nil
 }
+
+// NullSystemTagSet is a wrapper around SystemTagSet like guregu/null
+type NullSystemTagSet struct {
+	Set   *SystemTagSet
+	Valid bool
+}
```
codebien, line 161: In principle:
Do we need the nullable type at this level? It resolves a config/option's issue so I would prefer if this new type remains a solution for those spaces and let the rest of the code only use SystemTagSet.
In this way, the rest of the code is independent of what we need on the config side. If in the future we change the way we configure k6, and use other types for the Null problem then all the code will be impacted. Instead, if it only relies on pure SystemTagSet then we don't have to change it.

In concrete, I may miss something marking my suggestion as not feasible. @joanlopez What do you think?

metrics/system_tag.go
joanlopez: Any news? @mstoykov 🙏🏻

mstoykov: Sorry, I started writing and then apparently never published this.


Given how big this change is and that it now mixes Null types with the actual implementation of how it is used - which we never do, I am kind of for not going forward with this. 

I kind of prefer if we just use Null type and then fix it everywhere, but it might be *a bit* much. Arguably we will have to do this huge breaking change when we move to https://github.com/grafana/croconf or for any other fix for #883 . But not certain if we should do this before.

So at this point I kind of thing that your proposal to just patch `envconfig` might be better :thinking: 

Can you open a PR so we can at least look at it in parallel. 

cc @codebien I guess this also is your concern

joanlopez: Here's the PR: https://github.com/mstoykov/envconfig/pull/1
cc/ @codebien

joanlopez:
Closing in favor of https://github.com/grafana/k6/pull/3672

Issue #1408, opened 2020-04-24
Remove magic resolution of modules imports from CDNJS/GitHub


imiric:
We've seen [intermittent failures of `TestCDNJS`](https://circleci.com/gh/loadimpact/k6/9689) because their API sometimes returns empty results (`files` is empty). Right now you can see this if you request https://api.cdnjs.com/libraries/Faker a few times.

To avoid test failures we should skip this test for now, but this issue can also serve as a starting point for discussing eventually removing the "magic" URL loading from CDNJS or maybe even GitHub, to avoid confusing users (e.g. https://github.com/loadimpact/k6-docs/pull/8).
```[tasklist]
### Tasks
- [x] Warning about usage/future removal
- [ ] Remove in v0.53
```

MStoykov:
as I mentioned in the internal discussion, I am for disabling both github and cdnjs magic urls and the schemeless remote ones as well, but add a flag to enable them (I would argue we don't need separate flags). Have it this way for ... a year and drop them altogether.
Obviously we will need to also remove them from the documentation and make certain all our *-to-k6 tools generate only scripts not using both. 
We will probably also have some logic that if you run an archive created from before the version we release that the flag is ON instead of OFF but we still print a warning ...

mstoykov:
I have update the issue to reflect the current consensus of going forward with removing this magical resolution.

For v0.51.0 we will be adding a warning. Anc currently plan to drop it in 0.53.0

PR #3671, opened 2024-04-02
Deprecate magic resolving of cdnjs and github "URLs"

mstoykov:
## What?

Deprecate magic resolving of cdnjs and github "URLs"

## Why?

As part of #1408 we are first going to deprecate it for a couple of versions and then remove it.

## Checklist

## Related PR(s)/Issue(s)

#1408

### DIFF
--- a/loader/github.go
+++ b/loader/github.go
@@ -11,5 +11,5 @@ func github(logger logrus.FieldLogger, specifier string, parts []string) (string
 	return realURL, nil
 }
 
-const magicURLsDeprecationWarning = "Specifier %q was resolved to use a none conventional %s loader. " +
+const magicURLsDeprecationWarning = "Specifier %q resolved to use a non-conventional %s loader. " +
 	"That loader is deprecated and will be removed in v0.53.0. Please use the real URL %q instead."
--- a/loader/github.go
+++ b/loader/github.go
@@ -2,14 +2,9 @@ package loader
 
 import "github.com/sirupsen/logrus"
 
-func github(logger logrus.FieldLogger, specifier string, parts []string) (string, error) {
+func github(_ logrus.FieldLogger, _ string, parts []string) (string, error) {
 	username := parts[0]
 	repo := parts[1]
 	filepath := parts[2]
-	realURL := "https://raw.githubusercontent.com/" + username + "/" + repo + "/master/" + filepath
-	logger.Warnf(magicURLsDeprecationWarning, specifier, "github", realURL)
-	return realURL, nil
+	return "https://raw.githubusercontent.com/" + username + "/" + repo + "/master/" + filepath, nil
 }
-
-const magicURLsDeprecationWarning = "Specifier %q resolved to use a non-conventional %s loader. " +
-	"That loader is deprecated and will be removed in v0.53.0. Please use the real URL %q instead."
--- a/loader/github.go
+++ b/loader/github.go
@@ -2,9 +2,14 @@ package loader
 
 import "github.com/sirupsen/logrus"
 
-func github(_ logrus.FieldLogger, _ string, parts []string) (string, error) {
+func github(logger logrus.FieldLogger, specifier string, parts []string) (string, error) {
 	username := parts[0]
 	repo := parts[1]
 	filepath := parts[2]
-	return "https://raw.githubusercontent.com/" + username + "/" + repo + "/master/" + filepath, nil
+	realURL := "https://raw.githubusercontent.com/" + username + "/" + repo + "/master/" + filepath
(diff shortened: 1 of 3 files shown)


mstoykov:
Unfortunately adding a test in cmd/tests (where it seemed most relevant) hits the check that no goroutines are still alive at the end of the test. There are alive goroutines as the test actually hits the network and uses http2 to get the code. Unfortunately it also uses the default client and calling `CloseIdleConnections` on it doesn't do anything ;(.

Rewriting this so it can be more easily tested seems like quite a lot of work that also will mostly be thrown away once we have it removed so ... I decided not to go down this path.


Due to the way this is called internally the message is *only* printed once. It also doesn't get printed when running the archive, which might require additional code changes. But does get printed when creating the archive or running `k6 cloud`

loader/github.go
```diff
@@ -2,9 +2,14 @@ package loader
...
 	repo := parts[1]
 	filepath := parts[2]
-	return "https://raw.githubusercontent.com/" + username + "/" + repo + "/master/" + filepath, nil
+	realURL := "https://raw.githubusercontent.com/" + username + "/" + repo + "/master/" + filepath
+	logger.Warnf(magicURLsDeprecationWarning, specifier, "github", realURL)
+	return realURL, nil
 }
+
+const magicURLsDeprecationWarning = "Specifier %q was resolved to use a none conventional %s loader. " +
+	"That loader is deprecated and will be removed in v0.53.0. Please use the real URL %q instead."
```
olegbespalov, line 15: Maybe something like this?

```suggestion
const magicURLsDeprecationWarning = "Specifier %q resolved to use a non-conventional %s loader. " +
	"That loader is deprecated and will be removed in v0.53.0. Please use the real URL %q instead."
```

olegbespalov (dismissed):

loader/github.go
```diff
@@ -2,9 +2,14 @@ package loader
...
 	repo := parts[1]
 	filepath := parts[2]
-	return "https://raw.githubusercontent.com/" + username + "/" + repo + "/master/" + filepath, nil
+	realURL := "https://raw.githubusercontent.com/" + username + "/" + repo + "/master/" + filepath
+	logger.Warnf(magicURLsDeprecationWarning, specifier, "github", realURL)
+	return realURL, nil
 }
+
+const magicURLsDeprecationWarning = "Specifier %q resolved to use a non-conventional %s loader. " +
+	"That loader is deprecated and will be removed in v0.53.0. Please use the real URL %q instead."
```
codebien, line 15:
```suggestion
const magicURLsDeprecationWarning = "Specifier %q was resolved by a non-conventional %s loader. " +
	"The used %q loader is deprecated and will be removed in k6 v0.53.0. Please, use the real URL %q instead."
```
In any case, my feeling is that `non-conventional` isn't clear enough information. Is there a de-facto list of _conventional loaders_ for JavaScript that I'm not aware of?
I don't think we define somewhere the k6 _conventional_ (supported?) loaders. We should explicit them here or in the documentation.

I don't know if this is

codebien (changes requested):

loader/github.go
mstoykov: There isn't - and there is no documentation about cdnjs and github loaders for years at this point - we did drop it from every documentation like 3+ years ago.

I do agree - but IMO us making this message be more specific likely will have no benefit. I have not seen anyone use the github ones *ever* and the cdnjs ones seems to be used *very* rarely. 

So my expectation is most people will never see this message. 

If you have a concrete simple fix - I am all for it, but my only other proposal was "magic" (with quotes) and I think that is even less obvious. 

I would also argue that most people will agree that pure paths and URLs are the "conventional" specifiers even if the specification has left this completely unspecified :wink: :wink:

mstoykov:
I did redo practically the whole code so:
1. some parts of @codebien message changes were incorporated - the remaining ones were already kind of reverted from by @olegbespalov so I would prefer to not just flip-flop
2. This now works on running an archive, but doesn't print the real URL as in the case of cdnjs that will require making http requests, which goes against the whole idea of running from an archive. I don't really think this is such a big problem, but we can extend the message I guess.
3. The whole code changes is way more concentrated in one place.

I do think it migth be better to also link the original issue in the message, maybe as a way to also expand on the "non-conventional loader" term?

codebien (approved):

olegbespalov (approved):

codebien:
As of now, the loader name is returned as part of the message, I find the current version enough. As we defined it as a rare case, I don't think we need more. What is wrong and what should be done to fix it should be clear to the user.

PR #3676, opened 2024-04-05
Set working dir to the current one if test is read from stdin

### LINKED ISSUE #1462: k6 path resolution doesn't work from CWD when running from stdin
## Environment
- k6 version: 0.26.2
- OS and version: Linux (though this probably happens on other OSes as well)

## Expected Behavior
If I have the following as `test.js`:
```js
export var test = 123;
```

and if I have this as `script.js`:
```js
import { test } from "./test.js"

export default function (data) {
    console.log(test);
}
```
And both of these files are in the same folder, I should be able to execute them both by running `k6 run ./script.js` and `k6 run - <./script.js`, and get `123` printed either way.

## Actual Behavior

Instead, only the first command works. The second gives the following error:
```
GoError: The moduleSpecifier "file:///test.js" couldn't be found on local disk. Make sure that you've specified the right path to the file. If you're running k6 using the Docker image make sure you have mounted the local directory (-v /local/path/:/inside/docker/path) containing your script and modules so that they're accessible by k6 from inside of the container, see https://docs.k6.io/v1.0/docs/modules#section-using-local-modules-with-docker. 
```

For some reason, it's looking at the import file in the root folder.

mstoykov:
## What?

Usually the working dir for a given script is set based on it's path.

For test from stdin that path is file:///- so it is basically the root folder. With this change this changes to the directory k6 was ran from.

This means when you do `cat script.js | k6 run -` the working dir will be the same as if you did `k6 run script.js`.

But there is still no way to fix `cat path/to/somewhere/script.js | k6 run -` as k6 can't know where it got it's stdin from.

fixes #1462

## Why?

Fixing the issue and potentially having less people report it 
## Checklist

## Related PR(s)/Issue(s)

This is marked as breaking change as it changes the behaviour - arguably most people expect the new behaviour, so I expect no users actually depend on the current one.

### DIFF
--- a/loader/readsource.go
+++ b/loader/readsource.go
@@ -16,6 +16,7 @@ import (
 func ReadSource(
 	logger logrus.FieldLogger, src, pwd string, filesystems map[string]fsext.Fs, stdin io.Reader,
 ) (*SourceData, error) {
+	pwdURL := &url.URL{Scheme: "file", Path: filepath.ToSlash(filepath.Clean(pwd)) + "/"}
 	if src == "-" {
 		data, err := io.ReadAll(stdin)
 		if err != nil {
@@ -27,7 +28,7 @@ func ReadSource(
 		if err != nil {
 			return nil, fmt.Errorf("caching data read from -: %w", err)
 		}
-		return &SourceData{URL: &url.URL{Path: "/-", Scheme: "file"}, Data: data}, err
+		return &SourceData{URL: &url.URL{Path: "/-", Scheme: "file"}, Data: data, PWD: pwdURL}, err
 	}
 	var srcLocalPath string
 	if filepath.IsAbs(src) {
@@ -43,7 +44,6 @@ func ReadSource(
 		return Load(logger, filesystems, &url.URL{Scheme: "file", Path: filepath.ToSlash(srcLocalPath)}, src)
 	}
 
-	pwdURL := &url.URL{Scheme: "file", Path: filepath.ToSlash(filepath.Clean(pwd)) + "/"}
 	srcURL, err := Resolve(pwdURL, filepath.ToSlash(src))
 	if err != nil {
 		var unresolvedError unresolvableURLError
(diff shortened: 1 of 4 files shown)


codebien:
>But there is still no way to fix cat path/to/somewhere/script.js | k6 run -

It sounds like a nice to have from your point of view. Why? The new behavior sounds like the final solution to me.

mstoykov:
@codebien this was mostly just pointing out that this will still not work.

Not that I do think that there is a way to do it or that we should do something more. As you see I am closing the issue with the commit and the PR.

loader/readsource.go
```diff
@@ -16,6 +16,7 @@ import (
 func ReadSource(
 	logger logrus.FieldLogger, src, pwd string, filesystems map[string]fsext.Fs, stdin io.Reader,
 ) (*SourceData, error) {
+	pwdURL := &url.URL{Scheme: "file", Path: filepath.ToSlash(filepath.Clean(pwd)) + "/"}
```
codebien, line 19:
```suggestion
	pwdURL := &url.URL{Scheme: "file", Path: filepath.Clean(pwd) + "/"}
```
If I remove `ToSlash` I don't get any test failing. Why is it required? Isn't Clean enough?

mstoykov: ToSlash is about windows support. To be honest it might not be required, but as this is just a copy from a few lines below and the path related issues with windows are a nightmare tofix ... I prefer not to touch it :sweat_smile:

mstoykov: Aaah ... actually it is required because on windows Clean returns paths with `\` but that isn't a separator in URLs

codebien: Can you add a comment on it, please? So, the next time I read it I have at least an idea of what to research for. :pray:

codebien: Fine to me

codebien (approved):

joanlopez (approved):

PR #3680, opened 2024-04-08
Lock module resolution after initialization

mstoykov:
## What?

Make it an error to try to resolve not previously resolved modules, after initialization

## Why?

Partly because it should be given how we make k6 work and want `k6 archive` and `k6 cloud` to "just work".

Partly because this ended up being the only not completely hackish way to know that we are past the initialization state which was needed in order to make a warning only in the init context. 

See https://github.com/grafana/k6/pull/3681 

## Checklist

### DIFF
--- a/js/modules/resolution.go
+++ b/js/modules/resolution.go
@@ -89,7 +89,7 @@ func (mr *ModuleResolver) resolveLoaded(basePWD *url.URL, arg string, data []byt
 
 // Lock locks the module's resolution from any further new resolving operation.
 // It means that it relays only its internal cache and on the fact that it has already
-// seen previously the module during the initialization. 
+// seen previously the module during the initialization.
 // It is the same approach used for opening file operations.
 func (mr *ModuleResolver) Lock() {
 	mr.locked = true
--- a/js/modules/resolution.go
+++ b/js/modules/resolution.go
@@ -87,8 +87,8 @@ func (mr *ModuleResolver) resolveLoaded(basePWD *url.URL, arg string, data []byt
 	return mod, err
 }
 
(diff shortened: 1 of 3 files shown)


js/modules/resolution.go
```diff
@@ -81,6 +87,12 @@ func (mr *ModuleResolver) resolveLoaded(basePWD *url.URL, arg string, data []byt
 	return mod, err
 }
 
+// Lock locks the resolved from making any further new resolving and now depends only on it's cache
+// This to facilitate loading all the modules a test needs in it's initilization the same as with openning files.
```
codebien, line 91:
```suggestion
// Lock locks the module's resolution from any further new resolving operation. It means that it relays only its internal cache and on the fact that it has already seen previously the module during the initialization. 
// It is the same approach used for opening file operations.
```
I had to read several times to get the context. I prefer if we try to clarify a bit.

codebien (changes requested):
The logic LGTM, just a request for improving the comment.

codebien (approved):

js/modules/resolution.go
```diff
@@ -81,6 +87,12 @@ func (mr *ModuleResolver) resolveLoaded(basePWD *url.URL, arg string, data []byt
 	return mod, err
 }
 
+// Lock locks the module's resolution from any further new resolving operation. It means that it relays only its internal cache and on the fact that it has already seen previously the module during the initialization.
```
mstoykov, line 90:
```suggestion
// Lock locks the module's resolution from any further new resolving operation.
// It means that it relays only its internal cache and on the fact that it has already
// seen previously the module during the initialization. 
```

joanlopez (approved):

Issue #3534, opened 2024-01-11
Fix `require` to be specification compliant in respect to what it resolves relative to


mstoykov:
### Brief summary

k6's `require` has always resolved specifiers relative to the current "root of execution" module/script instead of to relative to the script/module the file is in.

An example below will help illustrate what "root of execution" is.

This behavior has been around forever and is against the [specification](https://wiki.commonjs.org/wiki/Website:Index/specs/modules/1.0) for commonJS and is not how dynamic `import` works or to be honest anything else I have come around.

The same issue is true for `open`, but:
1. this is issues is `require` specific to limit it scope
2. `open` is not based on any specification

For the rest of the issue I might add "Also relevant to `open`" as reminder for this, but `open` is not what we discuss here.

This is extracted from #2674 within a more actionable issue 

### k6 version

all

### OS

all

### Steps to reproduce the problem

Have 3 files:
main.js
```javascript
const s = require("./A/a.js")
if (s() != 5) {
	throw "Bad"
}
module.exports.default = () =>{} // just to not error
```

/A/a.js:
```javascript
module.exports =  function () {
  return require("./B.js");
}
```
/A/b.js
```javascript
module.exports = 5
```

This should not throw an exception - but it does. As at the time `s()` is executed the "root of the execution" is `main.js` not `/A/a.js` and consequently k6  resolves `./B.js` based on that and then loading it from disk doesn't find it leading to

```
ERRO[0000] GoError: The moduleSpecifier "./B.js" couldn't be found on local disk. Make sure that you've specified the right path to the file. If you're running k6 using the Docker image make sure you have mounted the local directory (-v /local/path/:/inside/docker/path) containing your script and modules so that they're accessible by k6 from inside of the container, see https://k6.io/docs/using-k6/modules#using-local-modules-with-docker.
        at go.k6.io/k6/js.(*requireImpl).require-fm (native)
        at file:///home/mstoykov/work/k6io/k6/path-based-tests-playground/issue-example/A/a.js:3:16(3)
        at file:///home/mstoykov/work/k6io/k6/path-based-tests-playground/issue-example/main.js:3:6(7)  hint="script exception"
```

### Expected behaviour

Script runs as expected and there is no exception.

### Actual behaviour

```
[...]

PR #3681, opened 2024-04-08
Emit warning if require will not comply with specification

mstoykov:
## What?

Add warning on require invocations that do not follow the specification.

Depends on https://github.com/grafana/k6/pull/3680

## Why?

As this isn't how it works we would like to fix it.

Also happens to be hard to keep it working once we move to ESM.

part of #3534 

## Checklist

### DIFF
--- a/js/modules/require_impl.go
+++ b/js/modules/require_impl.go
@@ -28,15 +28,19 @@ func NewLegacyRequireImpl(vu VU, ms *ModuleSystem, pwd url.URL) *LegacyRequireIm
 	}
 }
 
+const issueLink = "https://github.com/grafana/k6/issues/3534"
+
 // Require is the actual call that implements require
 func (r *LegacyRequireImpl) Require(specifier string) (*goja.Object, error) {
 	// TODO remove this in the future when we address https://github.com/grafana/k6/issues/2674
 	// This is currently needed as each time require is called we need to record it's new pwd
 	// to be used if a require *or* open is used within the file as they are relative to the
 	// latest call to require.
+	//
 	// This is *not* the actual require behaviour defined in commonJS as it is actually always relative
 	// to the file it is in. This is unlikely to be an issue but this code is here to keep backwards
 	// compatibility *for now*.
+	//
 	// With native ESM this won't even be possible as `require` might not be called - instead an import
 	// might be used in which case we won't be able to be doing this hack. In that case we either will
 	// need some goja specific helper or to use stack traces as goja_nodejs does.
@@ -48,6 +52,9 @@ func (r *LegacyRequireImpl) Require(specifier string) (*goja.Object, error) {
 		// In theory we can give that downwards, but this makes the code more tightly coupled
 		// plus as explained above this will be removed in the future so the code reflects more
 		// closely what will be needed then
+		if !r.modules.resolver.locked {
+			r.warnUserOnPathResolutionDifferences(specifier)
+		}
 		fileURL, err := loader.Resolve(r.currentlyRequiredModule, specifier)
 		if err != nil {
 			return nil, err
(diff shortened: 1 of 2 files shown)


js/modules/require_impl.go
```diff
@@ -67,3 +74,81 @@ func (r *LegacyRequireImpl) Require(specifier string) (*goja.Object, error) {
...
+			specifier, err, issueLink)
+		return
+	}
+	if r.currentlyRequiredModule.String() != k6behaviour.String() {
+		// this should always be equal, but check anyway to be certain we won't break something
+		logger.Warningf("The \"wrong\" path (%q) and the actually used by k6 path (%q) to resolve %q are different. "+
+			"Please report to issue %s. "+
+			"This will not currently break your script but might mean the future it won't work",
+			k6behaviour, r.currentlyRequiredModule, specifier, issueLink)
+	}
```
mstoykov, line 124: I meant to redo this text before I actually push this ...  :facepalm: 

But likely it will still need editorial changes.

codebien:
@mstoykov do you intend to open the pull request dropping the support directly now? It would be useful to be sure we are not forgetting to act on this.

mstoykov:
Finally managed to finish it in https://github.com/grafana/k6/pull/3694

codebien (dismissed):

oleiade (approved):

js/modules/require_impl.go
```diff
@@ -67,3 +74,81 @@ func (r *LegacyRequireImpl) Require(specifier string) (*goja.Object, error) {
...
+		return loader.Dir(u), nil
+	}
+	// Warn users on their require depending on the none standard k6 behaviour.
+	rt := r.vu.Runtime()
+	logger := r.vu.InitEnv().Logger
+	correct, err := normalizePathToURL(getCurrentModuleScript(rt))
+	if err != nil {
+		logger.Warningf("Couldn't get the \"correct\" path to resolve specifier %q against: %q"+
+			"Please report to issue %s. "+
+			"This will not currently break your script but might mean the future it won't work",
```
joanlopez, line 93:
```suggestion
			"This will not currently break your script, but it might mean that in the future it won't work",
```

js/modules/require_impl.go
```diff
@@ -67,3 +74,81 @@ func (r *LegacyRequireImpl) Require(specifier string) (*goja.Object, error) {
...
+			specifier, err, issueLink)
+	} else if r.currentlyRequiredModule.String() != correct.String() {
+		logger.Warningf("The \"wrong\" path (%q) and the actually used by k6 path (%q) to resolve %q are different. "+
```
joanlopez, line 96:
```suggestion
		logger.Warningf("The \"wrong\" path (%q) and the path actually used by k6 (%q) to resolve %q are different. "+
```

js/modules/require_impl.go
```diff
@@ -67,3 +74,81 @@ func (r *LegacyRequireImpl) Require(specifier string) (*goja.Object, error) {
...
+			"Please report to issue %s. "+
+			"This will not currently break your script but *WILL* in the future, please report this!!!",
+			correct, r.currentlyRequiredModule, specifier, issueLink)
+	}
+
+	k6behaviourString, err := getPreviousRequiringFile(rt)
+	if err != nil {
+		logger.Warningf("Couldn't get the \"wrong\" path to resolve specifier %q against: %q"+
+			"Please report to issue %s. "+
+			"This will not currently break your script but might mean the future it won't work",
```
joanlopez, line 106:
```suggestion
			"This will not currently break your script but it might mean that in the future it won't work",
```

js/modules/require_impl.go
```diff
@@ -67,3 +74,81 @@ func (r *LegacyRequireImpl) Require(specifier string) (*goja.Object, error) {
...
+			specifier, err, issueLink)
+		return
+	}
+	k6behaviour, err := normalizePathToURL(k6behaviourString)
+	if err != nil {
+		logger.Warningf("Couldn't get the \"wrong\" path to resolve specifier %q against: %q"+
+			"Please report to issue %s. "+
+			"This will not currently break your script but might mean the future it won't work",
```
joanlopez, line 114:
```suggestion
			"This will not currently break your script but it might mean that in the future it won't work",
```

js/modules/require_impl.go
```diff
@@ -67,3 +74,81 @@ func (r *LegacyRequireImpl) Require(specifier string) (*goja.Object, error) {
...
+			specifier, err, issueLink)
+		return
+	}
+	if r.currentlyRequiredModule.String() != k6behaviour.String() {
+		// this should always be equal, but check anyway to be certain we won't break something
+		logger.Warningf("The \"wrong\" path (%q) and the actually used by k6 path (%q) to resolve %q are different. "+
+			"Please report to issue %s. "+
+			"This will not currently break your script but might mean the future it won't work",
```
joanlopez, line 122:
```suggestion
			"This will not currently break your script but it might mean that in the future it won't work",
```

js/modules/require_impl.go
```diff
@@ -67,3 +74,81 @@ func (r *LegacyRequireImpl) Require(specifier string) (*goja.Object, error) {
...
+	if err != nil {
+		logger.Warningf("Couldn't get the \"wrong\" path to resolve specifier %q against: %q"+
+			"Please report to issue %s. "+
+			"This will not currently break your script but might mean the future it won't work",
+			specifier, err, issueLink)
+		return
+	}
+	if r.currentlyRequiredModule.String() != k6behaviour.String() {
+		// this should always be equal, but check anyway to be certain we won't break something
+		logger.Warningf("The \"wrong\" path (%q) and the actually used by k6 path (%q) to resolve %q are different. "+
```
joanlopez, line 120:
```suggestion
		logger.Warningf("The \"wrong\" path (%q) and the path actually used by k6 (%q) to resolve %q are different. "+
```

js/modules/require_impl.go
```diff
@@ -67,3 +74,81 @@ func (r *LegacyRequireImpl) Require(specifier string) (*goja.Object, error) {
...
+			"This will not currently break your script but might mean the future it won't work",
+			k6behaviour, r.currentlyRequiredModule, specifier, issueLink)
+	}
+}
+
+func getCurrentModuleScript(rt *goja.Runtime) string {
+	var parent string
+	var buf [2]goja.StackFrame
+	frames := rt.CaptureCallStack(2, buf[:0])
+	parent = frames[1].SrcName()
```
joanlopez, line 131: Is this always safe? Is there any change to get `frames` with only one or no items at all?🤔

mstoykov: In all of the experiments I have ran in the last 4 months or so (this particular code is very old) I have had 1 panic when I called it when there was nothing running on the runtime. 

As I would argue that is very much the wrong way to use it ... I am particularly certain it makes sense to capture this it seems.

olegbespalov (commented):
Generally, it looks good to me. 

One thing that could be improved is to migrate from the terms "correct" and "wrong" words. 

Maybe instead, the wrong use of CommonJS specification is non-compliant? I know that's longer, but I'm trying to think from the user perspective how they understand what's the issue faster. WDYT?

oleiade (approved):

olegbespalov (approved):

mstoykov:
@olegbespalov I used to call it "compliant" earlier .. but then I had to explain to what it is "compliant" and it becomes longer and longer.

Arguably the linked issue is the much better place to find why this is happening and what needs to be done. 

Also as I explain in the issue - you are *really* unlikely to hit this issue at all. If you just use ESM syntax - babel transpile it correctly. If not ... you literally need to be doing dynamic ~import~ require   which AFAIK nobody is doing. 

So I expect this will have practically 0 comments. But we will see :crossed_fingers:

PR #3687, opened 2024-04-11
Fix line numbers in case of commonJS file with no sourcemap

mstoykov:
## What?

Fix line numbers in case of commonJS file with no sourcemap:

```javascript
exports.something = "something"
throw "b"; // this was line 3 
```

## Why?

Previously if someone imports or runs a commonJS file (one that won't
need to go through babel) we still need to wrap it in some code to run
it. But we don't have a sourcemap to update, unless it already had
one(internal babel adds it). This leads to all lines being off by 1.

This is fixed by updating the sourcemap in the other case.

Unfortunately short of generating an "identity" sourcemap and then
shifting it by 1 with adding `;` to the first mapping, this can't be
done.

The sourcemap implementation specifically wants to find each mapping, so
unless it can do that it will not do anything.

The remaining hack (adjusting any available sourcemap) still stands as
it also fixes the columns on the *first* line of the file. As well as
actually being better.

Also found that while fixing linter issue I broke the tc39 error generation :(. 
## Checklist


js/compiler/compiler.go
```diff
@@ -207,8 +208,14 @@ func (c *Compiler) compileImpl(
 ) (*goja.Program, string, error) {
 	code := src
 	state := compilationState{srcMap: srcMap, compiler: c, wrapped: wrap}
-	if wrap { // the lines in the sourcemap (if available) will be fixed by increaseMappingsByOne
-		code = "(function(module, exports){\n" + code + "\n})\n"
+	if wrap {
+		if strings.Contains(code, "//# sourceMappingURL=") {
+			// the lines in the sourcemap (if available) will be fixed by increaseMappingsByOne
+			code = "(function(module, exports){\n" + code + "\n})\n"
```
oleiade, line 214: Really just a thought, and not blocking, but I had to squint to figure out this line was different from the one in the `else` branch (it adds a `\n` before we append the code value in the function body.

Maybe setting a variable to `""` as a default, updating it in this `if` statement, and using it in both cases after the condition could avoid such squinting in the future?

```go
var conditionalNewLine string = ""

if strings.Contains(code, "//# sourceMappingURL=") {
			// the lines in the sourcemap (if available) will be fixed by increaseMappingsByOne
			conditionalNewLine = "\n"
}

code = "(function(module, exports){" + conditionalNewLine + code + "\n})\n"
```

Or something along those lines, as I'm not necessarily super happy with the above as is 🙂

oleiade (approved):
🚀 🚀

olegbespalov (approved):

oleiade (approved):

olegbespalov (approved):

Issue #2978, opened 2023-03-13
Add Streams API support to k6


oleiade:
## Problem Statement

The [Streams API](https://developer.mozilla.org/en-US/docs/Web/API/Streams_API) is a specification for JavaScript which aims to provide users with programmatically access streams of data and process them as desired by the developer. It was initially designed for networking and with the Fetch API in mind, however, in light of recent work made on both [improving the handling of large files in k6](https://github.com/grafana/k6/issues/2974) and our future HTTP API, it appears that it might be a good candidate to improve and streamline data consumption and manipulation in k6 too.

## Proposal

We propose to provide a subset of the streams API as a k6 module, and expose it to both k6 internals as well as our public API. Judging by the scope of the Streams API itself, we propose that only `ReadableStream` would be implemented at first. If necessary at a later time, we could then implement support for writable and transformable streams too.

One of the main use case for streams as of this day would be around data processing, either from a file, or from a source in memory. It would likely allow us to consume data from various sources, without requiring to load it as a whole beforehand as it often is the case now.

Beyond the optimized resources consumption it would offer, we could also extend our existing APIs to have the ability to consume Streams. We believe beyond implementing and exposing streams to JavaScript user scripts, we would need to make sure our underlying API remains usable and intuitive from the Go side too to facilitate adoption from modules and extensions.

## Problem space

- https://github.com/grafana/k6/issues/2977 would probably be a good first candidate to return a stream to underlying file data, as opposed to reading it as a whole.
- https://github.com/grafana/k6/issues/592

## Execution

```[tasklist]
### Tasks
- [ ] https://github.com/grafana/k6/issues/3666
- [ ] https://github.com/grafana/k6/issues/3659
- [x] Expose a `streams` module to k6 scripts
```

andrewslotin:
Closing as done.

PR #3696, opened 2024-04-19
Add ReadableStream support to k6

### LINKED ISSUE #3659: Make ReadableStream (WIP) implementation pass WPT test suite
As we have a starting ReadableStream implementation, we aim to pass the relevant [Web Platform Test](https://github.com/web-platform-tests/wpt) suite's relevant tests.

```[tasklist]
### Tasks
- [x] Pass [bad-strategies.any.js](https://github.com/web-platform-tests/wpt/blob/master/streams/readable-streams/bad-strategies.any.js)
- [x] Pass [bad-underlying-sources.any.js](https://github.com/web-platform-tests/wpt/blob/master/streams/readable-streams/bad-underlying-sources.any.js)
- [x] Pass [cancel.any.js](https://github.com/web-platform-tests/wpt/blob/master/streams/readable-streams/cancel.any.js)
- [x] Pass [constructor.any.js](https://github.com/web-platform-tests/wpt/blob/master/streams/readable-streams/constructor.any.js)
- [x] Pass [count-queuing-strategy-integration.any.js](https://github.com/web-platform-tests/wpt/blob/master/streams/readable-streams/count-queuing-strategy-integration.any.js)
- [x] Pass [default-reader.any.js](https://github.com/web-platform-tests/wpt/blob/master/streams/readable-streams/default-reader.any.js)
- [x] Pass [floating-point-total-queue-size.any.js](https://github.com/web-platform-tests/wpt/blob/master/streams/readable-streams/floating-point-total-queue-size.any.js)
- [x] Pass [general.any.js](https://github.com/web-platform-tests/wpt/blob/master/streams/readable-streams/general.any.js)
- [ ] Pass [patched-global.any.js](https://github.com/web-platform-tests/wpt/blob/master/streams/readable-streams/patched-global.any.js)
- [x] Pass [reentrant-strategies.any.js](https://github.com/web-platform-tests/wpt/blob/master/streams/readable-streams/reentrant-strategies.any.js)
- [ ] Pass [tee.any.js](https://github.com/web-platform-tests/wpt/blob/master/streams/readable-streams/tee.any.js)
- [x] Pass [templated.any.js](https://github.com/web-platform-tests/wpt/blob/master/streams/readable-streams/templated.any.js)
```

joanlopez:
## What?

It adds support for [ReadableStream](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream) from [Streams API](https://streams.spec.whatwg.org/) through experimental `k6/experimental/streams` module. 

## Why?

This is a prerequisite to unlock the development of other features in k6. It is also a nice-to-have feature for k6 users in general, that may want to use `ReadableStream` on their own. 

## Note for the reviewer

You can find the rationale behind some technical decisions (around goja, promises, etc) present on this changeset, in [this
Google Doc](https://docs.google.com/document/d/1sWf2-Ypm0sVcHjSF9K4P0SvHMvKDnR5vy2TBEQFJmYQ/edit) that I used to write some personal notes, as takeaways during the development process.

On top of that, @oleiade and I decided to keep the code comments for each step of those operations defined in the spec because it makes it easier to:
- correlate the code and the spec.
- identify bugs or non-compliant behaviors.
- identify patterns that are repeated over the spec, and its correspondence in Go/goja.

So, although I know these could be arguable for some people, I'd personally push to keep them as long as I need to maintain and take care of this code.

## Checklist

## Related PR(s)/Issue(s)

Closes #3659
Contributes to #2978


js/modulestest/runtime.go
```diff
@@ -107,3 +112,45 @@ func (r *Runtime) innerSetupModuleSystem() error {
...
+
+		program, err := CompileFileFromFS(wptutils, fmt.Sprintf("wptutils/%s", file.Name()))
+		require.NoError(t, err)
+
+		_, err = runtime.VU.Runtime().RunProgram(program)
+		require.NoError(t, err)
+	}
+
+	return runtime
+}
```
mstoykov, line 156: I would really prefer if this isn't part of this module but part of the one that is being tested. This goes for practically all teh changes in `modulestest` although arguably the `console` one might be a good candidate to be extracted in a separate module and used both in the runner and `modulestest`

js/modules/k6/experimental/streams/tests/readable-streams/bad-strategies.js
```diff
@@ -0,0 +1,198 @@
+// Original source file: https://github.com/web-platform-tests/wpt/blob/f0d6f241e3a16ddd6e0a3ebcd9cb814b408bc5a6/streams/readable-streams/bad-strategies.any.js
```
mstoykov, line 1: I prefer to have this be downloaded similar to how `tc39` tests are that way I won't need to wonder if this file is edited or not. And also make it more obvious that we get it from another repo.

THe licensing around just copying some files in your repo are bit iffy and likely we will need to put some LICENSE file in this folder otherwise.


https://github.com/grafana/k6/blob/master/js/tc39/checkout.sh

https://github.com/grafana/k6/blob/e6e1016137667c3089e6f8b10b62e4bbefe1f2ea/js/tc39/tc39_test.go#L753-L759

js/modulestest/runtime.go
joanlopez: I can do that if you agree that's a good idea. I'd say I'm even surprised by to see that we have no other needs to expose `console`.

examples/experimental/streams.js
```diff
@@ -0,0 +1,34 @@
...
+	const stream = numbersStream()
+	const reader = stream.getReader()
+
+	while (true) {
+		const { done, value } = await reader.read()
+		if (done) break
+		console.log(`received number ${value} from stream`)
+	}
+
+	console.log('we are done')
```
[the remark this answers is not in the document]
joanlopez: Well.. yeah.. in practice, but while debugging WPTs is extremely useful, not to say that I'd even consider it a requirement, as it's the only way to debug the JS side (I don't think it's fair to consider goja debugger a real solution here, afaik).

So, do you really want me (or the person who needs to maintain that code) to copy and paste `console` code back & forth every time is needed? In the end, it's tests code...

js/modulestest/runtime.go
joanlopez: For the rest, I'm happy to move it as well, although that's a function that arguably could be used by other modules relying on WPTs. But lets copy & paste for now,

js/modules/k6/experimental/streams/tests/readable-streams/bad-strategies.js
joanlopez: I'd love to be able to do so, as it would make the whole process much easier, but as I stated in the aforementioned document, WPTs aren't applicable as-is, as they have parts that aren't fully compatible with k6/goja. Not to mention that we don't fully support all the test suites/tests.

If copy-pasting is not a solution, we'll need to find another one, but I feel like we most likely will need to have our own version of that thing. Any suggestion? Adding a LICENSE and/or an explanation file sounds good to me, but I'm not a legal expert either.

mstoykov: > s they have parts that aren't fully compatible with k6/goja

which parts are those?

> Not to mention that we don't fully support all the test suites/tests.

You can always run *some* of the tests. the tc39 test also do not run all the tests. Or you can (as our version of the tc39 tests) - run the tests but expect that they will error out. The later also helps with showcasing improvements and makes it a lot easier when adding functionality as you don't need   to also go hunt down a bunch of tests that you need to enable. Arguably you still need to go find out if some of the tests that should pass are ... well passing :)

examples/experimental/streams.js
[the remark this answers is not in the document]
mstoykov: I usually create a helper such as `print` which can even call `spew.Dump` or just `fmt.Println` - which also lets you do custom stuff in case where you need it.

Even in some tests where we've needed it for the test itself ot function it makes it easier to have `print` that appends to a slice ... instead of having to do all the logrus plumbing to get the logs in the end.

Doing the plumbing makes sense when you are testing the whole of k6 running something ... but in htat case you  also already have the logger.

As I said - I am not against console being moved so that it can be imported by both, but I would argue this now is a liability someone will make tests using console.log and think it is the real thing and then it will have different behaviour in a real execution.

js/modules/k6/experimental/streams/tests/readable-streams/bad-strategies.js
joanlopez: > which parts are those?

On the one hand, there are JS constructions that I think we don't support (yet?), like:
- [`for await...of`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for-await...of)
- [`async function*`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function*)

On the other hand, Web Platform Tests are mainly intended to be executed in a browser, and the ~5k lines of "basic" harness plus the surrounding code assumes that, most of the times. Of course, I can try to dig as deep as possible to that harness and to the Python code (which mainly glues a browser instance with the code) used to run Web Platform Tests in their standard way, and try to see if I can find a way to run them without modifying them at all, but overriding some parts. However, I'm not sure it really worths the effort, especially considering that we're only using that for ReadableStream tests, and that apparently we don't plan to expose such functionality as something available for extensions or more widely/generally).

> You can always run some of the tests

I'm not super convinced it's that easy to just run certain tests, especially considering that we don't want to modify the `wpt` repository contents at all, and considering what I mentioned above. In the worst case I guess we could just disable the test suites that we cannot fully support, and just enable them when modifications are required, or some workaround like that 😓

examples/experimental/streams.js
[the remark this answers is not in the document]
oleiade: I concur with @joanlopez on this one. I'd be interested in understanding if you see a concrete risk or downsides to adding out-of-the-box support for `console.log` in the test runtime.

Quite often, when implementing modules and writing tests for them, I need to be able to log values as I debug. The stream's implementation was just yet another occurrence.

While it's fair to argue that one could do some plumbing themselves every time they have this need, it doesn't feel like a good developer experience (for myself 😇 at the very least) to have to stop what I'm doing and switch to figuring out how to do it every single time. As a user, my expectation is: "My JS code is failing, and I need to see the value that's failing; my initial instinct is to resort to `console.log`." 
I've been using `throw` to do that, but in the streams API, where everything happens concurrently, it didn't scale that much. If I have to write a helper to do what I expect every time, I'd much rather have the tool I'd expect to find already present in the runtime to remove that overhead for myself.

This is a pretty common need when writing modules. Unless there's a downside to it, I think it would be a good quality-of-life improvement to have it as a default in the test runtime or at least to have a dedicated "rich" test runtime (side-by-side with the current one) or whatever that contains additional tooling such as this one.

js/modules/k6/experimental/streams/tests/readable-streams/bad-strategies.js
joanlopez: For further reference, you can see that from [`testharness.js API`](https://web-platform-tests.org/writing-tests/testharness-api.html), it states:
 _"The test harness script can be used from HTML or SVG documents and workers."_
 
 Also, for instance:
_"By default tests must be created before the load event fires."_
referring to the [`window.load` event](https://developer.mozilla.org/en-US/docs/Web/API/Window/load_event), which I'm not sure it makes lot of sense in the context of k6/goja.

and more, and more...


...which doesn't necessarily mean it's impossible to run them as they are with some tooling in place, but I guess the purpose of this issue is to bring `ReadableStream` support to k6, not building our own custom runner of WPTs based on goja, which sounds like it could be a completely separated project on its own.

So, unless I'm missing something and going the wrong way, I'd prefer to back to the beginning and re-evaluate how feasible would be to merge this as-is now, and see what would be required (adding a README/LICENSE? what else?).

> unless I'm missing something and going the wrong way 
if that's the case, I apologize in advance 🙏🏻  and would like to ask you, please, to shed some light, as I've been diving through swampy waters so far 😓 

Thanks! 🙏🏻

examples/experimental/streams.js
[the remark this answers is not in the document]
mstoykov: >  be interested in understanding if you see a concrete risk or downsides to adding out-of-the-box support for console.log in the test runtime.

>  As I said - I am not against console being moved so that it can be imported by both, but I would argue this now is a liability someone will make tests using console.log and think it is the real thing and then it will have different behaviour in a real execution.


So again - I am not against it. I am against it being reimplemented, and given that this PR does *not* use this - to be done in *this* PR.

The remaining post is reaffirming that *I* do not find it to be useful most of the time and showcases what I do. 

So if you want this - move this in another PR. But again given that this is the first time someone wanted it I would expect it won't be used a lot.

js/modules/k6/experimental/streams/tests/readable-streams/bad-strategies.js
mstoykov: > On the one hand, there are JS constructions that I think we don't support (yet?), like:

Both of those are https://github.com/dop251/goja/issues/555 essentially, so yeah - not supported and no idea when they will be,but likely after ESM (given that babel will block their usage either way). 

I don't really feel the need for a dedicated repo as the usage of them for extension have shown IME that they add more overhead than actual benefit. They are used because this is how we started - separate repo for an extension that we then wanted in core.

My whole take here is - I would prefer to get the test and harness from the original form in some repeatable way. For tc39 tests that is doable through cloning the repo. If here that will mean that we can get 90% of the test we run by just cloning and the rest 10% will need to apply "cool.patch" to them, I think that is also fine. Percentages are examples. 

In another month or two (or worse a year) someone will try to add something else and will spend days if not weeks figuring out how to get this tests and what was changed to make them work.

If that is too complicated to automate I would argue it also needs a considerable *in repo* explanation on how to do it. 

I managed to automate the test retrieval and patching in apparently an hour. Of which 40 minutes was me trying to figure what was changed in the harness, and remembering that the harness isn't in the module lost me a bunch of time trying to find out `step_timeout`.

https://gist.github.com/mstoykov/284e8ef15a07ed82521b52d389c2c354

You need to run `clone-tests.sh` (which should be automated in the CI) and everything  else seems to work. 

I really would prefer if the harness was also not modified or at least was *less* modified. But I don't really want to do that as well.

Again this is similar to how we do stuff for tc39 tests. There it just so happens that you come at the time most of the tests and the harness *work* in k6. This didn't used to be the case, and re-enabling stuff was hard even when we didn't rewrite and copy-paste with changes source code.

joanlopez: > I managed to automate the test retrieval and patching in apparently an hour.

That's not the complex part, but the harness, which is the what's mostly under the assumption of being executed in a browser. 

> My whole take here is - I would prefer to get the test and harness from the original form in some repeatable way. For tc39 tests that is doable through cloning the repo. If here that will mean that we can get 90% of the test we run by just cloning and the rest 10% will need to apply "cool.patch" to them, I think that is also fine. Percentages are examples.

But yeah, if having a `local.patch` for that, despite the conflicts it may cause in the future when retrieving the remote source, is acceptable, then I guess I can get something that works, much faster.

<br/>

Let me work on a proposal and see how acceptable would be what I get.

mstoykov: >  that, despite the conflicts it may cause in the future when retrieving the remote source

I would prefer to have to figure out how to update  a patch once a year when we update the tests. Than having to figure out what changed across all the tests and harnessess in this time and what were our changes.

As you saw we have pinned both the tc39 tests and we will be pinning this to a version. 

> Let me work on a proposal and see how acceptable would be what I get.

:+1: Again I prefer if we changed their sources *less* and try to make them work as much as possible. Deleting parts of them as we do not support syntax or something is way more acceptable and easy to figure out in a year, than it is to rewrite the whole thing.

js/modules/k6/experimental/streams/module.go
```diff
@@ -0,0 +1,285 @@
...
+
+	proto := call.This.Prototype()
+	if proto.Get("locked") == nil {
+		err = proto.DefineAccessorProperty("locked", rt.ToValue(func() goja.Value {
+			return rt.ToValue(stream.Locked)
+		}), nil, goja.FLAG_FALSE, goja.FLAG_TRUE)
+		if err != nil {
+			common.Throw(rt, newError(RuntimeError, err.Error()))
+		}
+	}
```
mstoykov, line 121: I guess you have done this to past some tests using `instanceof` or comparing prototypes? 
But at least for me dropping it still passes the tests :thinking: 

Also in general this is done by creating a prototype object and then using that as goja does through out. Looking at the specification for the streams I can't find where this is defined as well as actually anything on prototypes. So I guess it might be in some of the connected specifications :grimacing:

js/modules/k6/experimental/streams/module.go
```diff
@@ -0,0 +1,285 @@
...
+
+	// If the stream type is 'bytes', we don't want the size function.
+	// Except, when it is manually specified.
+	size := runtime.ToValue(defaultSizeFunc)
+	if len(call.Arguments) > 0 && !common.IsNullish(call.Arguments[0]) {
+		srcArg := call.Arguments[0].ToObject(runtime)
+		if !common.IsNullish(srcArg.Get("type")) && srcArg.Get("type").String() == ReadableStreamTypeBytes {
+			size = nil
```
mstoykov, line 154: in general `.Get` isn't particularly ... performant. So we might prefer to not call it multiple times with the same arguments unless the specification requires it.

edit: I will not be making the same comment on each place this happens

js/modules/k6/experimental/streams/readable_stream_default_controller.go
```diff
@@ -0,0 +1,485 @@
...
+}
+
+// releaseSteps performs the controller’s steps that run when a reader is
+// released, used to clean up reader-specific resources stored in the controller.
+//
+// It implements the ReadableStreamDefaultControllerReleaseSteps [specification]
+// algorithm.
+//
+// [specification]: https://streams.spec.whatwg.org/#readable-stream-default-controller-release-steps
+func (controller *ReadableStreamDefaultController) releaseSteps() {
```
mstoykov, line 218: it is a bit strange that this is empty, and the link about doesn't work for me :thinking:

js/modules/k6/experimental/streams/readable_streams.go
```diff
@@ -0,0 +1,586 @@
...
+		defaultReader.readRequests = []ReadRequest{}
+
+		// 6.3. For each readRequest of readRequests,
+		for _, readRequest := range readRequests {
+			// 6.3.1. Perform readRequest’s close steps.
+			stream.vu.RegisterCallback()(func() error {
+				readRequest.closeSteps()
+				return nil
+			})
+		}
```
mstoykov, line 451: I am a bit confused - why are we doing this in a callback? 

Also you are hitting the usual range gotcha that is fixed in go1.22+ (with the appropriate go.mod setting).

js/modules/k6/experimental/streams/readable_stream_default_reader.go
```diff
@@ -0,0 +1,211 @@
...
+	readRequests := reader.readRequests
+
+	// 2. Set reader.[[readRequests]] to a new empty list.
+	reader.readRequests = []ReadRequest{}
+
+	// 3. For each readRequest of readRequests,
+	for _, request := range readRequests {
+		// 3.1. Perform readRequest’s error steps, given e.
+		reader.vu.RegisterCallback()(func() error {
+			request.errorSteps(e)
```
mstoykov, line 173: Same as the case with closesteps

js/modules/k6/experimental/streams/readable_streams.go
```diff
@@ -0,0 +1,586 @@
...
+		common.Throw(stream.vu.Runtime(), newError(AssertionError, "reader.[[readRequests]] is empty"))
+	}
+
+	// 4. Let readRequest be reader.[[readRequests]][0].
+	readRequest := reader.readRequests[0]
+
+	// 5. Remove readRequest from reader.[[readRequests]].
+	reader.readRequests = reader.readRequests[1:]
+
+	callback := reader.vu.RegisterCallback()
```
mstoykov, line 537: I am also not certain why this is used.

In all of the places where RegisterCallback is used currently it is on the event loop and it queues otehr stuff on the event loop.

none of the tests fail if I remove any of htem.

The only place where you go off the event loop is in ReadableStream#Cancel and ther you use `promises.New` :+1: 

So I am not certain why this and the other RegisterCallback exist.

examples/experimental/streams.js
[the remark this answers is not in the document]
mstoykov: I will resolve this - I guess we will be making an issue and/or a PR at some future point

js/modules/k6/experimental/streams/tests/readable-streams/bad-strategies.js
mstoykov: Resolving this as well as the currently proposed solution in the PR works great IMO

js/modulestest/runtime.go
```diff
@@ -79,7 +79,7 @@ func (r *Runtime) SetupModuleSystemFromAnother(another *Runtime) error {
 // It is meant as a helper to test code that is expected to be run on the event loop, such
 // as code that returns a promise.
 //
-// A typical usage is to facilitate writing tests for asynchrounous code:
+// A typical usage is to facilitate writing tests for asynchrounous code:
```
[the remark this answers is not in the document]
joanlopez: 🤷🏻 (I don't fully get it, I think it was "fixed" by the linter itself)

js/modules/k6/experimental/streams/errors.go
```diff
@@ -0,0 +1,64 @@
...
+	// AssertionError is thrown when an assertion fails
+	AssertionError
+)
+
+type streamError struct {
+	// Name contains the name of the error
+	Name string `json:"name"`
+
+	// Message contains the error message
+	Message string `json:"message"`
```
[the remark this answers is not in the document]
joanlopez: That's a really good idea! I may give it a try, I fully agree that, despite not blocking (most likely), it will probably make our lives easier in the long term

js/modules/k6/experimental/streams/goja.go
```diff
@@ -0,0 +1,107 @@
...
+// setReadOnlyPropertyOf sets a read-only property on the given [goja.Object].
+func setReadOnlyPropertyOf(obj *goja.Object, name string, value goja.Value) error {
+	err := obj.DefineDataProperty(name,
+		value,
+		goja.FLAG_FALSE,
+		goja.FLAG_FALSE,
+		goja.FLAG_TRUE,
+	)
+	if err != nil {
+		return fmt.Errorf("unable to define %s read-only property on TextEncoder object; reason: %w", name, err)
```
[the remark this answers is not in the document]
joanlopez: I think they basically make the the code a bit less verbose, and probably a bit less error prone (as I think it's easy to make a mistake when specifying the flags), and these lines are repeated over and over. But nothing else.

Regarding parameterization, do you know if there's any safe way to get it from `*goja.Object`? 
Or just make it as a `string` argument, as with the property name? 🤔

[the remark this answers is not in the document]
oleiade: For context, yes, apologies, this is a snippet I took from my TextEncoder/TextDecoder implementation in an extension, and should be adjusted 👍🏻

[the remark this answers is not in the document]
mstoykov: > Regarding parameterization, do you know if there's any safe way to get it from *goja.Object?

Not really

>  Or just make it as a string argument, as with the property name? 🤔

this was my proposal

js/modules/k6/experimental/streams/readable_stream_default_controller.go
```diff
@@ -0,0 +1,485 @@
...
+
+// Error signals that the stream has been errored, and performs the necessary cleanup
+// steps.
+//
+// It implements the ReadableStreamDefaultController.error(e) [specification] algorithm.
+//
+// [specification]: https://streams.spec.whatwg.org/#rs-default-controller-error
+func (controller *ReadableStreamDefaultController) Error(err goja.Value) {
+	if err == nil {
+		err = goja.Undefined()
```
olegbespalov, line 138: This logic looks a bit weird to me; why is it so? I mean, why, in case of nil error, do we assign it to undefined and perform some `controller.error` ?

js/modules/k6/experimental/streams/readable_streams.go
```diff
@@ -0,0 +1,586 @@
...
+	call, ok := goja.AssertFunction(underlyingSource.Get("start"))
+	if !ok {
+		common.Throw(stream.runtime, errors.New("underlyingSource.[[start]] must be a function"))
+	}
+
+	return func(obj *goja.Object) (v goja.Value) {
+		var err error
+		v, err = call(underlyingSource, obj)
+		if err != nil {
+			panic(err)
```
olegbespalov, line 209: Out of curiosity, why is here (and in the following methods) a vanilla `panic`? :thinking:

js/modules/k6/experimental/streams/readable_stream_default_controller.go
oleiade: Good catch! Fixing the link, the correct one: https://streams.spec.whatwg.org/#readable-stream-default-controller-release-steps

Turns out, the specification contains a single instruction which is to return in the case of the default controller. I'm modifying the code to reflect that.

The specification mentions that this is to implement the [[ReleaseSteps]] contract, and the ByteStream implementation which we haven't implemented yet has much more steps for it: https://streams.spec.whatwg.org/#abstract-opdef-readablebytestreamcontroller-releasesteps

js/modules/k6/experimental/streams/module.go
```diff
@@ -0,0 +1,285 @@
...
+	// 3. Perform ! InitializeReadableStream(this).
+	stream := &ReadableStream{
+		runtime: mi.vu.Runtime(),
+		vu:      mi.vu,
+	}
+	stream.initialize()
+
+	// 4. If underlyingSourceDict["type"] is "bytes":
+	if underlyingSourceDict.Type == "bytes" {
+		common.Throw(stream.runtime, newError(RuntimeError, "'bytes' stream is not supported yet"))
```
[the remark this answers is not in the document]
oleiade: Good call 👍🏻 On it!

js/modulestest/compile.go
```diff
@@ -0,0 +1,43 @@
...
+}
+
+func compile(name string, b []byte) (*goja.Program, error) {
+	program, err := goja.Compile(name, string(b), false)
+	if err != nil {
+		return nil, err
+	}
+
+	return program, nil
+}
```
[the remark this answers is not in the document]
oleiade: I reused this snippet from what I did already in the past in webcrypto. If I remember, at the time I copied this pattern from the k6 codebase, it might have been linked to tc39 indeed, I can't remember. I think I did that at the time to make evaluating a couple of scripts involved in testing more convenient.

I'll look into it again 👀 and switch to ReadFile 👍🏻

js/modules/k6/experimental/streams/readable_stream_default_controller.go
joanlopez: I discovered that thanks to WPTs. I think it's because the `Error` method is a publicly exposed one, and it can be called with any argument. So, it turns that some values (probably `null`, or `undefined`, but I don't remember to be honest), is translated to `nil` in Go. However, when that error surfaces back to JS code, we need it to be a valid `goja.Value`, thus the need to sanity check it, and use a valid `goja.Value` instead.

joanlopez: Quickly going over all the WPTs that call `.Error()`, I see a few calls with no arguments, so probably that was one of the cases where the Go value received is `nil`. As stated above, then JS code expects `undefined`.

js/modules/k6/experimental/streams/tests/checkout.sh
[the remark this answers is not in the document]
joanlopez: Yeah! In the future we could even explore how to make that a reusable piece within k6, at least for internal modules, but eventually could be useful also for extensions.

js/modules/k6/experimental/streams/readable_streams.go
joanlopez: What `common.Throw` and our custom `throw` (see the description for further details) methods do is some manipulation over the error, and then `panic`. In these cases, we can just _throw_ whatever error we get, so we can just `panic` straightway.

Later, I think it's all the Goja wrapping who's responsible to adequately propagate that _throw_ to the JS surface.

js/modulestest/compile.go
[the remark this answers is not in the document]
oleiade: I remembered as I modifed it what made me create this helper at the time. The goal was really to fill the gap for the lack of `goja.Runtime.RunFile` or something like that. I think it's reasonably useful in the context of testing extensions like here or webcrypto, although it could probably use a different name to align with the Program terminology.

I've rewritten the function to use `os.ReadFile`, simplify it, and document it 👍🏻

js/modules/k6/experimental/streams/readable_streams.go
```diff
@@ -0,0 +1,586 @@
...
+// ReadableStream is a concrete instance of the general [readable stream] concept.
+//
+// It is adaptable to any chunk type, and maintains an internal queue to keep track of
+// data supplied by the underlying source but not yet read by any consumer.
+//
+// [readable stream]: https://streams.spec.whatwg.org/#rs-class
+type ReadableStream struct {
+	// Locked indicate whether the readable stream is locked to a reader
+	Locked bool
+
```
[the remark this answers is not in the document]
joanlopez: I think the explanation is partly similar to https://github.com/grafana/k6/pull/3696#discussion_r1579132598.

To satisfy some of the edge cases verified by WPTs, like asserting the property exists, it's read-only, it's a boolean, etc, etc; plus because at least for now we don't _manually_ "convert" or "transform" the Go object to a `*goja.Object`, as we do with `ReadableStreamDefaultController`.

I guess in the future that might be the way to go, but unless there's something that's really blocking, I'd leave to that for a later refactor, especially considering that this as-is is (_or seems to be_) WPTs compliant.

js/modules/k6/experimental/streams/readable_streams.go
oleiade: For additional context (correct me if I'm wrong joan), we `panic` directly here to avoid the wrapping involved in `common.Throw` because it keeps the WPT tests to pass, as they expect to find JS errors and receive GoError (goja type if I'm not mistaken) instead.

js/modules/k6/experimental/streams/readable_stream_default_controller.go
olegbespalov: I see, okay thanks for explanation :+1:

js/modules/k6/experimental/streams/tests/testharness.js.patch
```diff
@@ -0,0 +1,34 @@
...
+-            }
+             var status = e instanceof OptionalFeatureUnsupportedError ? this.PRECONDITION_FAILED : this.FAIL;
+             var message = String((typeof e === "object" && e !== null) ? e.message : e);
+             var stack = e.stack ? e.stack : null;
+@@ -2631,6 +2624,8 @@
+             this.set_status(status, message, stack);
+             this.phase = this.phases.HAS_RESULT;
+             this.done();
++
++            throw `${this.name} failed - ${e}`;
```
mstoykov, line 31: I am not certain this changes make sense, can you explain them?

joanlopez: To be honest, I haven't fully got how it works yet, but the management of the whole tests suite, its phases and all that, messes it up and prevents tests from being executed by running a single file, as we do. Again, all because of the assumptions it has on running on a browser, how the feedback is returned, etc.

joanlopez: I think I finally managed to get it working, _coming soon_!

js/modules/k6/experimental/streams/errors.go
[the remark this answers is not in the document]
joanlopez: It has been trickier than I initially thought, because sometimes we need to return these errors as Go `error`s, or keep them stored in-memory and `return`/`throw` them later. So, we need to either convert every signature into `any` instead of `error`, which would be messy, or un/wrap these errors, which is... meh, but feasible.

That said, I feel like like I'm having low inspiration these days, so if you have any other better approach, suggestions will be more than welcome.

js/modules/k6/experimental/streams/readable_streams.go
joanlopez: Good catch! I think the reason why I left this here is because at some point, `closeSteps` involved a goroutine call, so I decided to use callbacks to ensure their execution as part of the event loop. But yeah, it's no longer needed, I think.

js/modules/k6/experimental/streams/readable_stream_default_reader.go
joanlopez: Same explanation, thanks!

js/modules/k6/experimental/streams/readable_streams.go
joanlopez: Same, same as with `closeSteps` and `errorSteps`. Thanks!

js/modules/k6/experimental/streams/errors.go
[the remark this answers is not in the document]
oleiade: I concur with the points brought up by @joanlopez, refactoring the error mechanism here would be quite herculean. I would also much rather do that as a separate task, and even use this as an opportunity to return more standard/conventional JS errors to the runtime :)

I'll open an issue 👍🏻

js/modules/k6/experimental/streams/tests/testharness.js.patch
```diff
@@ -0,0 +1,46 @@
...
++
++						// We don't want to rely on the DOM and other browser-based
++            // mechanisms for reporting test failures. Instead, we just
++            // throw the error and make it fail fast, to be aware of it.
++            //
++            // In the future, we might want to add a way to report these
++            // that doesn't rely on a browser, but let all test end before
++            // actually reporting failures.
++            // But, as a first iteration, this approach should suffice.
++            throw `${this.name} failed - ${e}`;
```
mstoykov, line 31: I guess here `set_status` above should be the thing that marks tests as failed?

joanlopez: Yeah, but that status isn't directly reflected in any way through how we do run the tests. That's why I suggested there to explore a mechanism to do so, like "building" our own test environment, but I'd prefer to leave that for the next iteration as for now it's more than enough (it reports the failures) and the major drawback is that it fail fast and doesn't run all the tests when there's an error, which shouldn't be that bad for an initial release, as long as the behavior is correct.

(cc/ @oleiade as you'll most likely take over 🙇🏻)

js/modules/k6/experimental/streams/readable_streams_test.go
```diff
@@ -0,0 +1,123 @@
...
+func compileAndRun(t testing.TB, runtime *modulestest.Runtime, base, file string) {
+	program, err := modulestest.CompileFile(base, file)
+	require.NoError(t, err)
+
+	_, err = runtime.VU.Runtime().RunProgram(program)
+	require.NoError(t, err)
+}
+
+func executeTestScripts(vu modules.VU, base string, scripts ...string) error {
+	for _, script := range scripts {
```
mstoykov, line 88: I was very confused by this taking of multiple scripts , and why we are runnign them in the same runtime .. until I figured out we never run multiple scrtipts.

Can this be removed ?

just so somebody doesn't try to use it and then to try to figureo out why it isn't working.

js/modules/k6/experimental/streams/tests/testharness.js.patch
```diff
@@ -0,0 +1,46 @@
...
++            // this.all_loaded = true
+             if (this.on_loaded_callback) {
+                 this.on_loaded_callback();
+             }
+@@ -2631,6 +2635,16 @@
+             this.set_status(status, message, stack);
+             this.phase = this.phases.HAS_RESULT;
+             this.done();
++
++						// We don't want to rely on the DOM and other browser-based
```
mstoykov, line 23: nit; this are tabs, the rest of the file is with spaces.

Same for hte other part of the patch.

mstoykov (approved):
LGTM in general.

I have left some more nits.

I am still not certain about a bunch of the stuff around the running of the tests.

Also if `locked` is supposed to be read only - the test definitely do not test for this as having as an exported field on a struct let it be set from goja.

So maybe we should have more test than just hte ones in the test suite

oleiade (approved):

oleiade (dismissed):

olegbespalov (approved):
LGTM

mstoykov (approved):

mstoykov:
Just please squash on merge :grimacing:

oleiade:
👍🏻 😄 🦖

PR #3712, opened 2024-04-25
golangci-lint: enable it for all issues not just new

mstoykov:
## What?

Just call golangci-lint over the whole project not just the "new" parts.

## Why?

As now all lint issues are fixed we can enable without that option.

Unfortunately the option only looks at code at and around the places that was changed and if any of it has new lint errors it will trigger.

That does mean that if you update golangci-lint - it won't showcase new lint errors it triggers.

Also won't work if you add a bunch of code to the end an already somewhat long function. As the lint error will be at the beginning of the function.

## Checklist

## Related PR(s)/Issue(s)

#769

### DIFF
--- a/.github/actions/lint/action.yml
+++ b/.github/actions/lint/action.yml
@@ -28,5 +28,4 @@ runs:
       uses: golangci/golangci-lint-action@349d20632dbaed38f0a492cc991152e3d351e854 # latest commit at the time that uses node20
       with:
         version: ${{ steps.getenv.outputs.GolangCIVersion }}
-        only-new-issues: true
         args: "--config=${{ github.action_path }}/.golangci.yml"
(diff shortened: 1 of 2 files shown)


olegbespalov (approved):
:crossed_fingers:

codebien (approved):

olegbespalov:
As part of this PR, we might consider adding (increasing default) timeout  option, see https://github.com/grafana/k6/actions/runs/8831719518/job/24247507781?pr=3713

```
run golangci-lint
  Running [/home/runner/golangci-lint-1.57.2-linux-amd64/golangci-lint run --out-format=github-actions --config=/home/runner/work/k6/k6/./.github/actions/lint//.golangci.yml] in [] ...
  level=error msg="Timeout exceeded: try increasing it by passing --timeout option"
  
  Error: golangci-lint exit with code 4
```

I believe this could happen more often if we increase the scope of scanning :thinking:

Issue #3689, opened 2024-04-15
The line number from the sourcemap is one value higher


szkiba:
### Brief summary

In the attached example, the error reported line number is 14 instead of 13.
[k6smesbuild.zip](https://github.com/grafana/k6/files/14979788/k6smesbuild.zip)

### k6 version

k6 v0.50.0 (commit/011cccf9f1, go1.21.2, linux/amd64)

### OS

Ununtu Linux

### Steps to reproduce the problem

run script.js

### Expected behaviour

report line number 13 instead of 14

### Actual behaviour

line number 14 reported as source of error

PR #3717, opened 2024-04-26
fix line numbers when inline sourcemap is used

mstoykov:
## What?

Fix sourcemaps when they are inlined.

Before that if there was an inline sourcemap we won't load it and go through the already used method of updating it. 

This isn't a problem if babel actually transpiles it as it will load it, but in case the sourcecode was already transpiled with inline maps this will now update the lines correctly.

## Why?

To have correct lines when people user sourcemaps . 

## Checklist

## Related PR(s)/Issue(s)

#3689

### DIFF
--- a/js/compiler/compiler.go
+++ b/js/compiler/compiler.go
@@ -216,7 +216,7 @@ func (c *Compiler) compileImpl(
 			conditionalNewLine = "\n"
 			newCode, err := state.updateInlineSourceMap(code, index)
 			if err != nil {
-				c.logger.Warnf("while compiling %q we couldn't update it's inline sourcemap which might lead "+
+				c.logger.Warnf("while compiling %q, couldn't update its inline sourcemap which might lead "+
 					"to some line numbers being off: %s", filename, err)
 			} else {
 				code = newCode
--- a/js/compiler/compiler.go
+++ b/js/compiler/compiler.go
@@ -216,8 +216,8 @@ func (c *Compiler) compileImpl(
 			conditionalNewLine = "\n"
 			newCode, err := state.updateInlineSourceMap(code, index)
 			if err != nil {
-				c.logger.Warn("while compiling %q we couldn't update it's inline sourcemap which might lead "+
-					"to some line numbers being off", filename)
+				c.logger.Warnf("while compiling %q we couldn't update it's inline sourcemap which might lead "+
+					"to some line numbers being off: %s", filename, err)
 			} else {
 				code = newCode
 			}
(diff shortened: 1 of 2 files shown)


js/compiler/compiler.go
```diff
@@ -210,9 +211,17 @@ func (c *Compiler) compileImpl(
 	state := compilationState{srcMap: srcMap, compiler: c, wrapped: wrap}
 	if wrap {
 		conditionalNewLine := ""
-		if strings.Contains(code, "//# sourceMappingURL=") {
+		if index := strings.LastIndex(code, "//# sourceMappingURL="); index != -1 {
 			// the lines in the sourcemap (if available) will be fixed by increaseMappingsByOne
 			conditionalNewLine = "\n"
+			newCode, err := state.updateInlineSourceMap(code, index)
+			if err != nil {
+				c.logger.Warn("while compiling %q we couldn't update it's inline sourcemap which might lead "+
```
codebien, line 219: It is not clear to me as a user, if I can fix it in some way. Do we have a suggestion? Or the answer is `there isn't much you can do`.

szkiba (commented):
In the case of the originally attached script, the line number has been corrected. At the same time, if I transfer the throw expression to another function, it no longer works.

```ts
interface User {
  name: string;
  id: number;
}

class UserAccount implements User {
  name: string;
  id: number;

  constructor(name: string) {
    this.name = name;
    this.id = Math.floor(Math.random() * Number.MAX_SAFE_INTEGER);
  }
}

function newUser(name: string): User {
  throw "other";
  return new UserAccount(name);
}

export { User, newUser };
```

js/compiler/compiler.go
mstoykov: I would argue this warning is mostly so that when they report an issue - they report it with this warning

szkiba (dismissed):
It works now.

js/compiler/compiler.go
```diff
@@ -210,9 +211,17 @@ func (c *Compiler) compileImpl(
 	state := compilationState{srcMap: srcMap, compiler: c, wrapped: wrap}
 	if wrap {
 		conditionalNewLine := ""
-		if strings.Contains(code, "//# sourceMappingURL=") {
+		if index := strings.LastIndex(code, "//# sourceMappingURL="); index != -1 {
 			// the lines in the sourcemap (if available) will be fixed by increaseMappingsByOne
 			conditionalNewLine = "\n"
+			newCode, err := state.updateInlineSourceMap(code, index)
+			if err != nil {
+				c.logger.Warnf("while compiling %q we couldn't update it's inline sourcemap which might lead "+
```
codebien, line 219:
```suggestion
				c.logger.Warnf("while compiling %q, couldn't update its inline sourcemap which might lead "+
```

codebien (commented):
LGTM, just a small text suggestion

codebien (approved):

szkiba (approved):
LGTM

olegbespalov (approved):

PR #3682, opened 2024-04-10
Added summary generated event

ameetpal:
## What?

Emitting Summary Generated event

## Why?

https://github.com/grafana/k6/issues/3683

## Checklist

- [x] I have performed a self-review of my code.
- [x] I have added tests for my changes.
- [x] I have run linter locally (`make lint`) and all checks pass.
- [x] I have run tests locally (`make tests`) and all tests pass.
- [x] I have commented on my code, particularly in hard-to-understand areas.

PR #3730, opened 2024-05-02
Fixed Summary Generated Events

ameetpal:
## What?

This PR fixes the shortcomings of TestSummaryGenerated Event  :- https://github.com/grafana/k6/pull/3682

## Why?

The previously added TestSummaryGenerated had data in the form of a map with keys of type string and values of type io.Reader, however, the stream had already been read by stdout. Additionally, the data was formatted for console output, which was not consistent with what was expected by the handleSummary function of the test. Therefore, I returned the exact summary passed into the function and sent it as an event.

## Checklist


olegbespalov (dismissed):
LGTM

@codebien and maybe @grafana/k6-core we might consider adding this into upcoming release

olegbespalov:
Hey @ameetpal. After an internal discussion, we decided to revert to the original changes to reduce it from the scope of the upcoming release #3736 (since they will change). In the next v0.51, we could continue working on that PR. In any case, thank you for your contributions!

olegbespalov:
Hey @ameetpal 

We did release the k6 v0.51. Could you please rebase this PR using the latest `master` branch? Also, out of curiosity, this version of the changes satisfies your needs, right?

ameetpal:
Hey, i have updated the branch, some random tests are failing, but are not related to my changes
Yes these changes are correct

olegbespalov (approved):
Looks good to me :+1: Thanks for the contribution!

ameetpal:
@codebien can you please provide your review

ameetpal:
@codebien I have added a test for jsSummary. 
Also i have documented HandleSummary function but i have not added detailed keys and values as it may change based on a test and also boader key can change in future in runner function which will result in stale comment. 
If you still want, i can add comment to current keys that are being returned.

olegbespalov:
@ameetpal, out of curiosity, do the changes from this PR cover your case from #3732? I mean, have you tested everything assembled? I'm asking since we'd like to be sure that this PR serves the real needs

codebien:
@ameetpal can you also clarify if this pull request is going to solve https://github.com/grafana/k6/issues/3732 for you? Can you explain a bit more about how you intend to use it, please? So, in the case we decide to document this feature, we could use your use case as an example in the documentation. Or, we can have it here just as historical information in case someone in the future needs to investigate more on why we decided to introduce this event.

ameetpal:
Hey @olegbespalov , @codebien , as of now i will consume this event in the new extension that i am working on. I was thinking of some way to include extension without explicitly importing this in test script but since this is not an option now, we will ask user to import our extension in the test script. I have tested this in local, and that is why i have raised this PR, since i was getting empty data in extension from the changes in my previous PR. Currently this is the only feasible method we find to solve our problem.

ameetpal:
Also, if we are able to use this successfuly, in future we might flow out other events like TestStarted. For auditing purposes.

js/summary_test.go
```diff
@@ -593,6 +593,33 @@ func TestRawHandleSummaryData(t *testing.T) {
...
+	summary := createTestSummary(t)
+	result, jsData, err := runner.HandleSummary(context.Background(), summary)
+	delete(jsData, "setup_data")
+	require.NoError(t, err)
+	jsonString, jerr := json.Marshal(jsData)
+	require.NoError(t, jerr)
+	require.NotNil(t, result["rawdata.json"])
+	newRawData, err := io.ReadAll(result["rawdata.json"])
+	require.NoError(t, err)
+	assert.JSONEq(t, string(newRawData), string(jsonString))
```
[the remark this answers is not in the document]
ameetpal: Both are same in test as i want to get the data supplied to handleSummary function in test script file.

map[string]io.Reader, contains stdout related data which can only be read once. Also we can use this read also, but we need to write a converter function that will return the json data from which it was created. 

If you want i can look into that approach also

ameetpal:
@codebien any updates here ?

codebien:
Hi @ameetpal,
I and @olegbespalov met and synced on this pull request.

The current approach doesn't seem to be ideal, because it creates a dependency with an internal API. The current submitted changes rely upon a `map[]` structure [designed internally](https://github.com/grafana/k6/blob/32b6c1e6ead5e785338a4bb63e12dc20c06f3012/js/summary.go#L62) to the `js` package. It mostly prepares the structure to serialize the Summary as a JSON. It is used from `--summary-export` feature, which is already a deprecated part of k6 that we might change in the future.

Instead, a more solid option could be to export `lib.Summary` type as `Data` field for the _event_ emitted. In that way, you can create directly in your extension the best structure for your use case. `lib.Summary` is the type currently passed through `HandleSummary` and it is already under a shared package (`lib`).
https://github.com/grafana/k6/blob/f57dc2f76826d3601e1c4b28039be373849fc4f3/cmd/run.go#L195

Does this proposal still solve your problem?

joanlopez:
Moving to https://github.com/grafana/k6/milestone/42 since there's no news since a while.

codebien:
Hey @ameetpal, do you have any news on the suggested approach, and would it work for you?

ameetpal:
Hey @codebien , approch seems good, i hot busy with some other task, will push changes in 2 or 3 days

ameetpal:
Hey @codebien, i have incorporated the comment, please review the PR

cmd/run.go
```diff
@@ -201,9 +201,15 @@ func (c *cmdRun) run(cmd *cobra.Command, args []string) (err error) {
...
 					IsStdErrTTY: c.gs.Stderr.IsTTY,
 				},
-			})
+			}
+			summaryResult, hsErr := test.initRunner.HandleSummary(globalCtx, summary)
 			if hsErr == nil {
 				hsErr = handleSummaryResult(c.gs.FS, c.gs.Stdout, c.gs.Stderr, summaryResult)
+				waitForSummaryGeneratedEvent := emitEvent(&event.Event{
+					Type: event.TestSummaryGenerated,
+					Data: summary,
```
codebien, line 210: It should be better to copy ObservedMetrics and RootGroup instead of directly passing them. They are pointers and they might generate data races if there is a read/write across extensions and/or from k6core.

lib/runner.go
```diff
@@ -99,3 +99,8 @@ type Summary struct {
 	NoColor         bool          // TODO: drop this when noColor is part of the (runtime) options
 	UIState         UIState
 }
+
+// String Overiding for tests
+func (s *Summary) String() string {
+	return "&{Summary:map[...]...}"
+}
```
codebien, line 106: I guess when we agree on the generation part then we should update this with some additional information. We might have `Summary{Metrics: number, RootGroup: name/(or nil), and the other fields}`.

codebien (changes requested):

ameetpal:
Closing this PR

PR #3738, opened 2024-05-13
TypeScript and ES6+ support using esbuild.

### LINKED ISSUE #3703: Native TypeScript support
### Feature Description

TypeScript has become increasingly popular for its strong typing system, which enhances code quality and reliability—key aspects in performance testing scenarios. Currently, k6 does not support TypeScript out-of-the-box, requiring users to rely on external tools to transpile TypeScript code into JavaScript before it can be run by the tool. This additional step not only complicates the setup process but also impacts the overall developer experience, especially for those who primarily work with TypeScript. This setup also discourages TypeScript developers from adopting k6, despite its powerful testing capabilities.

To address the current gap, I suggest integrating [k6pack](https://github.com/grafana/k6pack), a solution based on [esbuild](https://esbuild.github.io/), into the core functionality of k6. esbuild is a TypeScript transpiler and bundler written in Go, which aligns well with k6’s Go-based architecture. The integration would involve automatically invoking k6pack during the k6 launch process to transpile TypeScript scripts into executable JavaScript.

This integration can be structured as follows:

* Automatic Detection: k6 should automatically detect TypeScript files (.ts/.tsx) and invoke k6pack.
* Seamless Transpilation: Behind the scenes, k6pack uses esbuild to transpile TypeScript to JavaScript, handling any necessary bundling and minification. This process should be transparent to the user, requiring no manual steps.
* Configuration Options: Users can provide custom configuration for the transpilation process via k6's existing configuration files or command-line options, allowing flexibility in how TypeScript is handled (e.g., specifying custom tsconfig.json settings).

_No response_

szkiba:
## What?

With the introduction of the "enhanced" compatibility mode, the test source code is transformed using esbuild instead of Babel.

## Why?

Source files with the extension ".ts" are loaded by esbuild's TypeScript loader, which results in partial TypeScript support. This esbuild removes exactly the type information, but does not provide type safety.
    
Source files other than ".ts" are loaded by esbuild's JavaScript loader, which results in the support of a more modern JavaScript dialect than goja.

## Checklist

## Related PR(s)/Issue(s)

https://github.com/grafana/k6/issues/3703

Closes #3703


codebien (commented):
>Source files with the extension ".ts" are loaded by esbuild's TypeScript loader, which results in partial TypeScript support. This esbuild removes exactly the type information, but does not provide type safety.

@szkiba can you expand on this topic? Isn't the main advantage of having TypeScript? Does it mean we are delegating it to another tool and k6 makes the assumption to receive a valid script?

>With the introduction of the "enhanced" compatibility mode

Can you explain the reason behind the _enhanced_ term, please? Isn't better for clarity to use something related to `typescript` or `esbuild`?

Last point, I see we are not running the `js/tc39` suite. I expect it is the first requirement for being able to merge this pull request. Can you please make the changes for running it with the new mode? It is probably the fastest way to catch any potential issue without manually checking all the bits.

szkiba:
Hi @codebien , thank you for your review.

> @szkiba can you expand on this topic? Isn't the main advantage of having TypeScript? Does it mean we are delegating it to another tool and k6 makes the assumption to receive a valid script?

Sorry if I was unclear: we do not delegate the loading to a tool, but to the esbuild library. Within the esbuild library, there are so-called loaders for each type of content. The quoted sentence indicates that files with the .ts extension will be treated as TypeScript files, in the case of other file extensions, the input will be assumed to be JavaScript.

> Can you explain the reason behind the _enhanced_ term, please? Isn't better for clarity to use something related to `typescript` or `esbuild`?

The name of the mode is not t`ypescript`, because it can also be used with JavaScript files. It's not `esbuild` because I followed the naming convention, the `extended` compatibility mode isn't called `babel` either. In `enhanced` compatibility mode, esbuild replaces babel.

> Last point, I see we are not running the `js/tc39` suite. I expect it is the first requirement for being able to merge this pull request. Can you please make the changes for running it with the new mode? It is probably the fastest way to catch any potential issue without manually checking all the bits.

I think the tc39 test is running, I modified the tc39_test.go file:
https://github.com/grafana/k6/pull/3738/files#diff-a1d828027ee2f0cea82c600f5a52b13f91091474fcfc7f2d381db0994363ed62R768

codebien:
>The name of the mode is not typescript, because it can also be used with JavaScript files. It's not esbuild because I followed the naming convention, the extended compatibility mode isn't called babel either. In enhanced compatibility mode, esbuild replaces babel.

If this is the case, what is missing for attempting a Babel replacement? So, instead of creating a new mode, we do an expansion of the _extended_ mode.

>I think the tc39 test is running, I modified the tc39_test.go file:

Oh, thanks. I missed it.

szkiba:
> If this is the case, what is missing for attempting a Babel replacement? So, instead of creating a new mode, we do an expansion of the _extended_ mode.

That's exactly what I think, esbuild will replace babel (one day). I created a new compatibility mode so that esbuild can be completely disabled (and disabled by default). With such a major, potential incompatibility change, I think this is important. Later, esbuild can take over the role of babel and then the two compatibility modes (extended and enhanced) can mean the same thing.

codebien:
If this is the case, should we go with something like `experimental-extended-v2`?

szkiba:
> If this is the case, should we go with something like `experimental-extended-v2`?

How about `experimental-enhanced` ? And we will simply remove this if we think that esbuild can replace babel. Otherwise, it can renamed to `enhanced` if we want to keep both.

pablochacin:
@szkiba  I tried running a tests in [grafana-api-test library](https://github.com/grafana/grafana-api-tests/blob/main/tests/dashboards/dashboard_read.ts) but I get this error:

```
time="2024-05-21T13:09:45+02:00" level=error msg="GoError: The moduleSpecifier \"../../lib/session\" couldn't be found on local disk. Make sure that you've specified the right path to the file. If you're running k6 using the Docker image make sure you have mounted the local directory (-v /local/path/:/inside/docker/path) containing your script and modules so that they're accessible by k6 from inside of the container, see https://grafana.com/docs/k6/latest/using-k6/modules/#using-local-modules-with-docker.\n\tat go.k6.io/k6/js.(*requireImpl).require-fm (native)\n\tat file:///home/pablo/go/src/github.com/grafana/grafana-api-tests/tests/dashboards/dashboard_read.ts:1:27(61)\n" hint="script exception"
```

szkiba:
> @szkiba I tried running a tests in [grafana-api-test library](https://github.com/grafana/grafana-api-tests/blob/main/tests/dashboards/dashboard_read.ts) but I get this error:
> 
> ```
> time="2024-05-21T13:09:45+02:00" level=error msg="GoError: The moduleSpecifier \"../../lib/session\" couldn't be found on local disk. Make sure that you've specified the right path to the file. If you're running k6 using the Docker image make sure you have mounted the local directory (-v /local/path/:/inside/docker/path) containing your script and modules so that they're accessible by k6 from inside of the container, see https://grafana.com/docs/k6/latest/using-k6/modules/#using-local-modules-with-docker.\n\tat go.k6.io/k6/js.(*requireImpl).require-fm (native)\n\tat file:///home/pablo/go/src/github.com/grafana/grafana-api-tests/tests/dashboards/dashboard_read.ts:1:27(61)\n" hint="script exception"
> ```

This is a consequence of the fact that esbuild does not run as a bundler, but transforms per module. The module resolving (require function) is implemented by the k6 runtime. Specifying the file extension is mandatory in this implementation.
Currently, the workaround is to specify the file extension in the import.
In the future, it should be considered to modify the implementation of the require function so that the file name extension is not mandatory.

js/compiler/enhanced.go
```diff
@@ -0,0 +1,55 @@
...
+	"github.com/dop251/goja/file"
+	"github.com/dop251/goja/parser"
+	"github.com/evanw/esbuild/pkg/api"
+)
+
+func esbuildTransform(src, filename string) (code string, srcMap []byte, err error) {
+	opts := api.TransformOptions{
+		Sourcefile:     filename,
+		Loader:         api.LoaderJS,
+		Target:         api.ES2017,
```
mstoykov, line 15: This makes esbuild unneedingly transform a whole bunch of code that goja supports.

In practice from the list of features on https://esbuild.github.io/content-types/#javascript 

The only ones not supported are async generators, async iteration and  `??=`. And in my test the async generator and async iteration doesn't work with the example from [mdn](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function*).

This also adds a bunch of esnext stuff which I haven't tested, but will argue are probably even less desired by even our less average users given that they aren't released yet.

If you will likely to enable any of those it will be nice to also enable their tc39 tests respectfully. I would recommend that this is done *after* the initial release of this instead of now.

So my strong recommendation is this to be set to ESNext - which in practice disables esbuild doing any transformation for features and leave just the TypeScript + commonJS transofmration. Some tc39 tests get slightly different results witht that.

You can use the `Supported` field to set which transformation you want to add. Again I will recommend doing this after this is merged, maybe even after we have had some feedback.

js/tc39/breaking_test_errors-extended.json
```diff
@@ -29,10 +29,10 @@
...
   "test/built-ins/TypedArrayConstructors/BigUint64Array/is-a-constructor.js-strict:true": "test/built-ins/TypedArrayConstructors/BigUint64Array/is-a-constructor.js: ReferenceError: BigUint64Array is not defined <at omitted>",
-  "test/language/comments/hashbang/function-constructor.js-strict:true": "test/language/comments/hashbang/function-constructor.js: SyntaxError: test/language/comments/hashbang/function-constructor.js: Unexpected token (17:47)\n  15 | const AsyncFunction = (async function (){}).constructor;\n  16 | const GeneratorFunction = (function *(){}).constructor;\n> 17 | const AsyncGeneratorFunction = (async function *(){}).constructor;\n     |                                                ^\n  18 | for (const ctor of [\n  19 |   Function,\n  20 |   AsyncFunction, <at omitted>",
+  "test/language/comments/hashbang/function-constructor.js-strict:true": "test/language/comments/hashbang/function-constructor.js: SyntaxError: Async generators are not supported yet <at omitted>",
```
mstoykov, line 32: I guess this is due to some change in tc39_test.go - most likely because this is the error from babel before and the newer one is from goja.

I don't think this matters in this case - just mentioning it for others.

mstoykov:
> But the convention is not to include the extension, right?

This hasn't been the convetion with ESM ... at all. 

[Node docs](https://nodejs.org/api/esm.html#import-specifiers)
>  Relative specifiers like './startup.js' or '../config.mjs'. They refer to a path relative to the location of the importing file. The file extension is always necessary for these.

[Deno docs}(https://docs.deno.com/runtime/manual/node/migrate#replace-require-with-import)
> Also, note that we include the full file extension when importing modules, much as you would in the browser. There is also no special handling of files named index.js.

Browsers have always done this.

`require` in node (and anythign compatible with it ) follows and algorithm to figure out what the user actually meant by probing the filesystem.

This was also one of Ryan Dahl [regrets](https://youtu.be/M3BM9TB-8yA?si=ys0yAquuHB_Gyi0v&t=836) with node. The video is from the middle of a bunch of regrets aroudn how node does modules and resolution and such.

szkiba:
@mstoykov , @codebien, Do I understand correctly that if I rename the "enhanced" compatibility mode to "experimental-enhanced", then this PR will be mergeable?

mstoykov:
I would argue https://github.com/grafana/k6/pull/3738#discussion_r1616178683 is also a fairly good idea. But otherwise I don't see anything else

szkiba:
> I would argue [#3738 (comment)](https://github.com/grafana/k6/pull/3738#discussion_r1616178683) is also a fairly good idea. 

Sure, but you suggested to do it after the merge (or after releasing and get some feedback)

> But otherwise I don't see anything else

Thank you, I'll rename the enhanced mode to experimental-enhanced

mstoykov:
The recommended after merging and feedback was about enabling back some of the transformation, not moving to ESNext

szkiba:
> The recommended after merging and feedback was about enabling back some of the transformation, not moving to ESNext

Oh, I see, sorry

codebien (approved):
LGTM.

@szkiba can you report what is the new binary size with `esbuild` dependency included, please?

szkiba:
> LGTM.
> 
> @szkiba can you report what is the new binary size with `esbuild` dependency included, please?

On master, the k6 binary size is 39612416 bytes, on branch with esbuild the k6 binary size is 43503616 bytes (x86 linux).
The increment is 3891200 bytes ~ 3.5M

mstoykov (approved):

PR #3740, opened 2024-05-14
Support to init a ReadableStream from a io.Reader

joanlopez:
## What?

It exposes a mechanism from `streams` (Go) package to initialize a new `ReadableStream` from a `io.Reader`:
- `NewReadableStreamForReader(vu modules.VU, reader io.Reader) *goja.Object`

## Why?

Because this would makes it possible to reuse the `ReadableStream` implementation added in `v0.51.0` from Go code, so other pieces in k6 (or any k6 extension) that have a Go `io.Reader` and want to translate to a `ReadableStream`, usable from JS code, could it.

## Checklist

## Related PR(s)/Issue(s)

N/A


oleiade:
I think this is great 🚀 🚀 🚀 

This is an idea I wanted to bring up here, but I'm not suggesting we should do it, consider it blocking, nor even arguing this is a good idea: I often find myself creating `NewSomething() *goja.Object`, and `NewSomethingFrom(goja.Value)` functions when working in extensions. A while ago, when working in WebCrypto, I experimented with the idea of leveraging Go generics to create `From` and `To` generic traits. 

However, we ended up not needing generics at the time. I wonder if this idea would make sense in such a context, where for a type `T`, we can implement a trait `To[T] *goja.Object` and `From(goja.Value) *T`.

This is somewhat blurry in my head, but I keep thinking about it. Maybe this is something we could try in the future, although I'm not yet certain if it would have benefits beyond creating a contract for creating from and exporting as object given types (and I might be biased by Rust, where this would likely be the way to go in a lot of situations). 

I don't know, but if you have thoughts, I'd love to hear them 😄

olegbespalov (approved):
LGTM! :rocket: 

It's not related to the PR, but it's worth using branches in the `grafana/k6` repository. That way, if needed, another maintainer could pick up the work and finish it.

js/modules/k6/experimental/streams/module.go
```diff
@@ -285,3 +290,49 @@ func (mi *ModuleInstance) NewReadableStreamDefaultReader(call goja.ConstructorCa
 
 	return object
 }
+
+// NewReadableStreamForReader is the equivalent of [NewReadableStreamDefaultReader] but to initialize
+// a new [ReadableStream] from a given [io.Reader] in Go code.
+// It is useful for those situations when a [io.Reader] needs to be surfaced up to the JS runtime.
+func NewReadableStreamForReader(vu modules.VU, reader io.Reader) *goja.Object {
```
codebien, line 297:
```suggestion
func NewReadableStreamFromReader(vu modules.VU, reader io.Reader) *goja.Object {
```

Is `For` intentional in this case?

joanlopez: Yeah, it was intentional, but I don't have a strong opinion.

mstoykov (approved):

olegbespalov (approved):

js/modules/k6/experimental/streams/module.go
```diff
@@ -285,3 +290,49 @@ func (mi *ModuleInstance) NewReadableStreamDefaultReader(call sobek.ConstructorC
...
+// It is useful for those situations when a [io.Reader] needs to be surfaced up to the JS runtime.
+func NewReadableStreamFromReader(vu modules.VU, reader io.Reader) *sobek.Object {
+	rt := vu.Runtime()
+	return newReadableStream(vu, sobek.ConstructorCall{
+		Arguments: []sobek.Value{rt.ToValue(underlyingSourceForReader(vu, reader))},
+		This:      rt.NewObject(),
+	})
+}
+
+func underlyingSourceForReader(vu modules.VU, reader io.Reader) *sobek.Object {
```
codebien, line 305: It seems we miss a replacement here `s/For/From/`

js/modules/k6/experimental/streams/module_test.go
```diff
@@ -0,0 +1,41 @@
...
+	"bytes"
+	"testing"
+
+	"github.com/grafana/sobek"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"go.k6.io/k6/js/modulestest"
+)
+
+func TestNewReadableStreamForReader(t *testing.T) {
```
codebien, line 13: Same

Issue #2976, opened 2023-03-13
Add a streaming-based CSV parser to k6


oleiade:
Users with big CSV files with a size superior to, say, 500MB, and using a large numbers of VU directly experience our [issue with handling large files](https://github.com/grafana/k6/issues/2974).

As a result, we would like for k6 to offer an alternative way to handle CSV files. Ideally, we would like it to be streaming-based, and to hold only a subset of the data at a time in memory. That way k6 memory footprint would remain sustainable for such users.

## Non-final target API

```javascript
import http from 'k6/http';
import { csv } from 'k6/files'

let filename = '10M-rows.csv';

// username, password, email
// pawel, test123, pawel@k6.io
// ...

// not using the old open() api.
// let fileContents = open(filename);

let fileHandle = streamingOpenFileHandler(filename);

const csvHandler = csv.objectReader(fileHandle.stream, {
  delimiter: ',',
  consumptionStrategy: 'uniqueSequential', // VU-safe, non-repeating.
  endOfFileHandling: 'startFromBeginning', // what to do when we run out of rows
})

export default function () {
  let object = csvHandler.next() // unique row across all VUs
  object.username

  const res = http.post('http://test.k6.io/login', {
    user: object.username,
    pass: object.password
  });
}
```

## Prerequisites

However, being able to provide such an alternative implementation of a CSV parser that would work both for open-source and cloud users is currently blocked by issues listed in "[improving the handling of large files in k6](https://github.com/grafana/k6/issues/2974)".

Namely, we would first need the ability to access such files, seek through them, and stream their content without having to first decompress them on disk, and without having to load their whole content in memory first. Another prerequisites would also be the presence of an API that allows to open and read files separately too, as opposed to storing their content in memory.

mstoykov:
I would like to argue about pivoting the initial implementation to an API and technics that will be applicable for the current architecture of k6 and it's capabilities. 

That is I would argue that for users it will be more beneficial if the new csv parser can return an array ... even maybe directly a SharedArray instead of having to also get #3014 to work.

This means that users will be able to use this as a more of a drop in replacement to get the [data parametization](https://grafana.com/docs/k6/latest/examples/data-parameterization/) code they use today to be faster and lighter.

Given that due to afero we in practice will load the file twice in memory before we even start reading it and that I doubt uploading more than 100mb of an archive to the cloud will work reliably either way. And it will likely need changing the archive type of k6 at the minimum.

[...]

PR #3743, opened 2024-05-15
Add an experimental csv module exposing a streaming csv parser 

oleiade:
## What?

This PR is a cleaned-up version of the CSV streaming parser we hacked during Crococon.

It aims to address #2976, and adds the capability to parse a whole CSV file as a `SharedArray` natively (without having to resort to parse).

#### Parse function

The `parse` function takes a `fs.File` instance as input, as well as options, parses the whole file as csv, and returns a `SharedArray` instance containing the parsed records.

It aims to offer a similar experience as to what is currently possible with the `open` function and `papaparse` with the added benefits to:
- consume less memory as it uses the new `fs.open` function: the file content and the parsed records will be shared across VUs, too (albeit a copy in itself).
- be faster, especially for larger files, as it is designed to bypass most of the JS runtime, and directly parse the file in, and store the results in a `SharedArray` in Go. Through our pairing sessions with @joanlopez we profiled the execution extensively with Pyroscope. We made some comparisons, and found out that most of the CPU time spent parsing using papaparse into a `SharedArray` was spent in the JS runtime. The approach picked in this PR mitigates that.

This API allows the trade of memory for performance. The whole file content will still be held in memory a couple of times, and we'll also hold a copy of all the file's parsed rows, however, in our benchmark, this approach was significantly faster than using papaparse.

```javascript
import { open } from 'k6/experimental/fs'
import csv from 'k6/experimental/csv'
import { scenario } from 'k6/execution'

export const options = {
	iterations: 10,
}

// Open the csv file, and parse it ahead of time.
let file;
let csvRecords;
(async function () {
	file = await open('data.csv');

	// The `csv.parse` function consumes the entire file at once, and returns
	// the parsed records as a SharedArray object.
	csvRecords = await csv.parse(file, { delimiter: ',', skipFirstLine: true, fromLine: 10, toLine: 1000})
})();


	console.log(csvRecords[scenario.iterationInTest])
}
``` 

#### Parser

The parser results from our initial CSV parsing workshop at Crococon. Its API is specifically designed to address #2976. It exposes a Parser object and its constructor which behave similarly to a JS iterator, on which the `next` method can be called and returns the next set of records as well as a `done` marker indicating whether there is more to consume.

The parser relies exclusively on the `fs.File` constructs and parses rows as they go, instead of storing them all in memory. As such, it consumes less memory but is also somewhat slower (comparable to paparse) to parse as each call to `next()` needs to go through the whole JS Runtime and event loop (observed during our profiling sessions in Pyroscope); making the cost of creating/await the `next` promise significantly bigger than the actual parsing operation.

The parser effectively trades performance for memory but offers some flexibility in parsing and interpreting the results.  

```javascript
import { open } from 'k6/experimental/fs'
import csv from 'k6/experimental/csv'

export const options = {
	iterations: 10,
}

let file;
let parser;
(async function () {
	file = await open('data.csv');
	parser = new csv.Parser(file, { delimiter: ',', skipFirstLine: true, fromLine: 10, toLine: 1000});
})();

	const {done, value} = await parser.next();
	if (done) {
		throw new Error("No more rows to read");
	}

	console.log(value);
}
```

### Implementation details & Open Questions

- In order to support the module, we had to significantly reshape the internals of the `fs` module in order to facilitate opening and manipulating files using it from another module. The biggest part of the change was to introduce an interface specific to the `fs.File` behavior that we needed to rely on from the outside, namely read, seek and stat: `ReadSeekStater`. See commit https://github.com/grafana/k6/pull/3743/commits/8e782c1867d7a3900149e65f0d09b2c1c3b06e92 for more details.
- As we found per our profiling investigation, instantiating shared arrays that we would fill with the results of parsing the csv file in Go showed little benefit, and displayed that most of the execution was spent in the JS runtime. In order to make things faster, we added a Go `SharedArray` constructor to the Go data module that allows to replicate the behavior of the JS constructor in Go, and effectively bypass most of the runtime overhead. We were not sure this was the best approach, let us know if you think of something better. See commit https://github.com/grafana/k6/pull/3743/commits/d5e6ebc650a2a9138380ffaaa5e7abc288e59bec for more details.

### What's not there yet

* Part of the initial design described in #2976 included two concepts I haven't yet included here, as I'm not sure what the best API or performance-oriented solution would be (ideas welcome 🤝):
* The ability to describe a strategy for the parser to select which lines should be picked for parsing or ignored (say, you want a file's lines parsing to be spread evenly across all your VUs, for instance).
* The ability to **instruct the parser to cycle through the file**: once it reaches the end, it restarts from the top. My main question mark is that as it would probably be possible using the existing APIs (`csv.Parser.next` returns an iterator-like object with a `done` property, seeking through the file is possible, and re-instantiating the parser once the end is reached is an option), would we want indeed to have a dedicated method/API for that? 

## Why?

Using CSV files in k6 tests is a very common pattern, and until recently, doing it efficiently could prove tricky. One common issue users encounter is that JS tends to be rather slow when performing parsing operations. Hence, we are leveraging the `fs` module constructs and asynchronous APIs introduced in Goja over the last year to implement a Go-based CSV "high-performance" streaming parser. 

## Checklist

## Related PR(s)/Issue(s)

#2976


joanlopez:
> One common issue encountered by users is that JS tends to be rather slow when performing parsing operations.

Take this just a simple idea rather than something that's really a requirement for this pull request to move forward, but considering that you explicitly mentioned that, would be nice to have a small benchmark for comparison.

js/modules/k6/experimental/csv/data.csv
```diff
@@ -0,0 +1,21 @@
+firstname,email
```
joanlopez, line 1: Nit; but do we really need to have non-ASCII characters, IDNs and... are there email addresses with whitespaces? 🤔 😅

js/modules/k6/experimental/csv/module.go
```diff
@@ -0,0 +1,238 @@
...
+	"go.k6.io/k6/js/promises"
+
+	"go.k6.io/k6/js/modules/k6/experimental/fs"
+
+	"github.com/dop251/goja"
+	"go.k6.io/k6/js/common"
+	"go.k6.io/k6/js/modules"
+)
+
+// TODO: Should we have an option to cycle through the file (restart from start when EOF is reached, or done is true)?
```
joanlopez, line 22: My personal opinion; if it doesn't pollute a lot the code/API, then I'd go for it, as in theory it doesn't sound like something very complex while it might be very handy for certain use cases.

js/modules/k6/experimental/csv/module.go
```diff
@@ -0,0 +1,238 @@
...
+// TODO: Should we have an option to skip empty lines (lines with no fields?)
```
joanlopez, line 23: My first reaction is: why not doing so by default? But I guess we should/could take a look at similar parser libraries/tools, and get inspiration from there (what looks to be more common).

PS: Do you mean lines with at least one value missing, or those lines completely empty? 🤔

js/modules/k6/experimental/csv/module.go
```diff
@@ -0,0 +1,238 @@
...
+// TODO: Should we have an option to skip lines based on a specific predicate func(linenum, fields): bool ?:w
```
joanlopez, line 24: Again my opinion; I'd say that'd be a nice-to-have, but not critical for a first iteration, so I'd keep that idea for a second iteration.

js/modules/k6/experimental/csv/module.go
```diff
@@ -0,0 +1,238 @@
...
+}
+
+// parseResult holds the result of a CSV parser's parsing operation such
+// as when calling the [Next].
+type parseResult struct {
+	// Done indicates whether the parser has finished reading the file.
+	Done bool `js:"done"`
+
+	// Value holds the line's records value.
+	Value []string `js:"value"`
```
joanlopez, line 180: Nit; why not using the plural form, as I guess it will generally contain multiple values, thus it'd be more clear that it's an array, especially considering how little explicit is JS.

js/modules/k6/experimental/fs/module.go
```diff
@@ -154,8 +157,14 @@ type File struct {
...
-	// file contains the actual implementation for the file system.
-	file
+	// Impl contains the actual implementation of the file logic, and
+	// interacts with the underlying file system.
+	//
+	// Note that we explicitly omit exposing this to JS to avoid leaking
+	// implementation details, but keep it public so that we can access it
+	// from other modules that would want to leverage its implementation of
+	// io.Reader and io.Seeker.
+	Impl file `js:"-"`
```
joanlopez, line 167: Nit; I think in general it's discouraged to expose an attribute as part of the public surface of the package (`fs` in this case) which type is private.

joanlopez (commented):
Thanks for giving form to what we started during the Crococon 💟 

I left multiple comments as some form of initial feedback, but generally speaking I think this approach is more than okay, and from my side I'd suggest to move forward (with tests and all that) 🚀

I'm not sure how far are we from being able to publish this as an _experimental_ module, but I guess it's part of the its _experimental_ stage, the feedback and usage we will collect from users, what will help us answer some of the open questions that you left, and to actually confirm whether the current approach is good enough or not.

js/modules/k6/experimental/csv/data.csv
oleiade: Hehe I used one of my tools to produce "fake" csv, I'm completely open to using another file 😅
This should probably live in a test case anyway 👍🏻 I'll remove the file ^^

js/modules/k6/experimental/csv/module.go
oleiade: Talking with Pawel, he agreed with you, and mentioned that our current scope is good enough for 90% of the use cases, and we could do it in the future if we still want to 👍🏻

js/modules/k6/experimental/csv/data.csv
joanlopez: To be honest, I'm fine with having this file with fake data, I was just captured by some of details of the data shape haha 😄

js/modules/k6/experimental/csv/csv.js
codebien, line 1: I would expect to use the parser mostly within two scenarios:

1. Getting one line per iteration
2. Getting one line per VU

The current example iterating over the batch during the single iteration doesn't sound representative of the most common use case.

Am I missing something?

codebien (changes requested):

js/modules/k6/experimental/csv/csv.js
oleiade: You're both right 🙂

If I summarize:
- "As a user, I want to create a parser, and every executed iteration gets the next line, and advances the line "head" to the next one for the next iteration (regardless of the VU executing the iteration)"
- "As a user, I want to create a parser, and when the VU code calls the `next` method, the next `line_number % vu_number` is returned to it"

Do you agree with these statements? 🙂

I think both call for a rather different design and on top of my head, it might be not easy to support both in an efficient manner. I'm gonna look into it, and see what I come up with 🙇‍♂️

joanlopez: Yeah! I guess we could leave the second one for the future, if we manage to let different VUs use different parsers, and leave the work for the test author. I know it's not ideal, and for certain situations where the amount of VUs varies over time it might be more difficult, tho.

In general I'd say, let's get an MVP and ship it! And, imho, that MVP should have at least one mechanism that doesn't imply having a new parser for each iteration but shared across iterations/vus somehow (the one we think have a better balance between popularity and feasibility). Once we have that, I'd vote for merging it, then iterate with more strategies.

js/modules/k6/experimental/fs/module.go
oleiade: I agree, this is an abstraction leak. The reason I went with that as a start was because I found the `Read`/`ExportedRead` approach we had initially taken somewhat inelegant (at least it tickled my sense of "nice" code).

But in hindsight it's probably the best compromise indeed.

The `Impl` approach that's explicitly not exposed to users, but to other k6 libs and modules, started from the rationale that the underlying `file` struct already implements `ReadSeeker`, and I might as well leverage that. But because I couldn't find a way to retrieve the underlying `file` implementation struct using `ExportTo` or by doing it manually, I went with a public `Impl` instead.

I'll give another shot at our initial approach and try to make it look okay 👍🏻

js/modules/k6/experimental/csv/data.csv
oleiade: My little tool tries to be inclusive ^^

js/modules/k6/experimental/csv/module.go
joanlopez: Agreed! That would be nice 👍🏻

js/modules/k6/experimental/fs/module.go
joanlopez: Would it be feasible to define it as a `io.ReadSeeker`? 🤔  

If possible, even if that requires a couple of tricks internally (because of `stat()`), maybe that's a decent way to expose it for other k6 libs and modules.

oleiade: Do you mean?

```go
type File struct {
    // ...
    
	ReadSeeker io.ReadSeeker `js:"-"`
	
	// ...
}
```

If so, it's something I considered, and I think it could work (would have to double-check) 🙇‍♂️

js/modules/k6/experimental/csv/csv.js
codebien: Per-vu:
```
vuID := exec.vu.idInTest

let parser = new csv.Parser(file, {
  fromLine: vuID, // potentially with the modulo
  toLine: vuID, // potentially with the modulo
})

const {done, value} = await parser.next();

		console.log(value)
}
```

Per-iteration:
```
let parser = new csv.Parser(file)

  const {done, value} = await parser.next();
  console.log(value)
}
```

They should already work, we have to just add a unit test with them

js/modules/k6/experimental/fs/module.go
joanlopez: Something like that, yeah. I think it won't work as-is, because of `stat()`, as I said, but perhaps we can find some workaround that only lives internally/privately in the package/implementation, with the benefit of keeping the public surface cleaner.

oleiade: Let me see what I can do 🙇‍♂️

js/modules/k6/experimental/csv/csv.js
joanlopez: Yeah, for the "Per-vu", I guess there could be multiple strategies:
- Enabling an option to skip rows, like just process the rows _multiple of..._
- Suggesting the user, if they now the amount of rows in advance, to split the readers accordingly.

js/modules/k6/experimental/csv/csv.js
```diff
@@ -0,0 +1,25 @@
+import { open } from 'k6/experimental/fs'
+import csv from 'k6/experimental/csv'
+
+let file;
+(async function () {
+	file = await open('data.csv');
+})();
+
+export default async function() {
+	let parser = new csv.Parser(file, {
```
oleiade, line 10: For context @joanlopez and @codebien I just found out that currently, because we don't have support for await in the init context, the only way (with the current design) to instantiate the parser from the init context is the following:

```javascript
let file;
let parser;
(async function () {
	file = await open('data.csv');
	parser = new csv.Parser(file, {
		delimiter: ',',
		skipFirstLine: true,
		fromLine: 3,
		toLine: 13,
	})
})();
```

Which works, but adds a workaround to the workaround :-/

An alternative I would consider would be to have the parser support for receiving a file path, and opening the file under the hood (which might involve exposing the currently private `openImpl` function from the `fs` module to other modules such as this one). 

What do you think? 🦖

js/modules/k6/experimental/csv/csv.js
joanlopez: > Re per-vu: It's interesting. I think we had completely different assumptions/understandings regarding the per-vu strategy 🤓 From your example, it looks like you wish to select a chunk of the whole file based on a vu number, whereas I had in mind that for 16 VU, vu 1 would receive line 1, 17, 33, etc.

Aren't these two the strategies I mentioned above? 🤔 Or did I miss any?

js/modules/k6/experimental/csv/csv.js
codebien: >Which works, but adds a workaround to the workaround :-/

It doesn't sound like a problem of this scope/pull-request. The problem is async in Init context and at some point we have to resolve it (we are already on it with ESM).

oleiade: You're probably right indeed 👍🏻 🙇🏻

js/modules/k6/experimental/csv/csv.js
oleiade: Indeed. I think we posted our comments at the same time, not seeing each other's 🤓 

CleanShot 2024-05-23 at 11 31 54

js/modules/k6/experimental/fs/module.go
oleiade: Alright, I've pushed a [commit](https://github.com/grafana/k6/pull/3743/commits/b8def4c0bbcd539782dc63a65acb83f23159ff8f), which exposes the behavior as an interface instead. I like it much more, indeed.

I've created a `Stater` interface, made `file` implement it, and incorporated it in a `ReadSeekStater` interface, which I use to expose the underlying behavior instead of `file` itself. Let me know what you think 👍🏻

js/modules/k6/experimental/csv/module.go
oleiade: Regarding `skipEmptyLines`, it turns out that skipping them is the default behavior of the Go CSV parser, and it looks like it can't be overridden, so it sorts it out 👍🏻

js/modules/k6/experimental/fs/module.go
joanlopez: It looks much cleaner now, at the very least! 👌🏻

js/modules/k6/experimental/csv/csv.js
joanlopez: Glad to see we're aligned, then! Also at when to write comments hahahaha :)

oleiade:
Posting here a summary of the use-cases we discussed privately, and that we'd like the module to tackle:

1. As a user, I want to read a CSV file containing 1000 credentials, and have each credential being processed by a single iteration.
  - no credential should be processed more than once
  - unless the parser is explicitly to restart from the begining? In that scenario, the same credential can be processed multiple times.
  - if the option is not set, and the user calls parser.next() after all credentials are consumed, they keep getting a `{ done: true, value: undefined }` response.
2. As a user, I want to read a CSV file containing 1000 credentials, and have each subset of those credentials reserved to be processed by a single VU.
- the subset of credentials could be for instance a chunk: 0-100 credentials go to VU 1, 101-200 credentials go to VU 2, etc.
- the subset of credentials could be every Nth credential: 0, 10, 20, 30, etc. go to VU 1, 1, 11, 21, 31, etc. go to VU 2, etc.
- This is possible with the existing SharedArray approach, but it needs a faster way of processing the rows.
3. As a user, I want each iteration to stream through my CSV file, and have the ability to act upon each returned records.
- The user has the ability to skip a record, or to stop the iteration, based on the content of the record, or the line number.
- This is assuming that each iteration needs the whole content of the file to perform its test.

joanlopez:
@oleiade is it fine to move this to https://github.com/grafana/k6/milestone/42? 
Or do you still plan to try to get this part of https://github.com/grafana/k6/milestone/41?

examples/experimental/csv-parse.js
```diff
@@ -0,0 +1,24 @@
...
+
+// Open the csv file, and parse it ahead of time.
+let file;
+let csvRecords;
+(async function () {
+	file = await open('data.csv');
+
+	// The `csv.parse` function consumes the entire file at once, and returns
+	// the parsed records as a SharedArray object.
+	csvRecords = await csv.parse(file, {delimiter: ','})
```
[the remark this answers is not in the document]
joanlopez: My initial reaction to your comment is positive, but to be honest I don't have a strong opinion. I see pros & cons for both approaches, either hiding the implementation details or being specific to leave the more generic methods of the API for the future (stabilization).

js/modules/k6/data/data.go
```diff
@@ -90,7 +93,55 @@ func (d *Data) sharedArray(call sobek.ConstructorCall) *sobek.Object {
...
 	array := d.shared.get(rt, name, fn)
-	return array.wrap(rt).ToObject(rt)
+	return array.Wrap(rt).ToObject(rt)
+}
+
+// Reader is the interface that wraps the basic Read method.
+//
+// The data module Reader interface is implemented by types that can read data from a source, such
+// as a CSV file, etc.
+type Reader interface {
```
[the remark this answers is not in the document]
joanlopez: The interface in stdlib looks like:

```
type Reader interface {
	Read(p [][byte](https://pkg.go.dev/builtin#byte)) (n [int](https://pkg.go.dev/builtin#int), err [error](https://pkg.go.dev/builtin#error))
}
```

which is slightly different, as I guess the expectations are.

Just for context, this was designed like this to avoid extra allocations, following a similar pattern. However, there's probably an alternative approach that could consist on somehow "wrapping" one into the other one, and making this package to receive `io.Reader`.

For me, as-is is fine, but I'd be happy to see suggestions with `io.Reader`.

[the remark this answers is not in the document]
codebien: No, I don't want to move to `io.Reader`. I hoped there was an interface avilable for it in `bufio`/`scanner`.

Btw, what about the renaming part `s/Reader/LineReader`?

js/modules/k6/experimental/csv/module.go
```diff
@@ -0,0 +1,289 @@
...
+
+	"go.k6.io/k6/js/common"
+	"go.k6.io/k6/js/modules"
+)
+
+type (
+	// RootModule is the global module instance that will create instances of our
+	// module for each VU.
+	RootModule struct {
+		// FIXME: we end up instantiating the data module ourselves, which would live side by side with other
```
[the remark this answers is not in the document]
joanlopez: To be honest, I'm not really sure what's preferable in the tradeoff of global state vs having two instances of the data module 🤔

js/modules/k6/experimental/csv/module.go
```diff
@@ -0,0 +1,289 @@
...
+		var err error
+
+		records, err = p.reader.Read()
+		if err != nil {
+			if errors.Is(err, io.EOF) {
+				done = true
+			} else {
+				reject(err)
+				return
+			}
```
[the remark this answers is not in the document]
joanlopez: Doesn't that suggestion modify the behavior? 🤔 

Like, when `errors.Is(err, io.EOF) == true`, the promise will be rejected with the error, instead of being resolved with done.

If we want to simplify it, I think it should be something like:

```
if err != nil {
	if !errors.Is(err, io.EOF) {
		reject(err)
		return
	}
	done = true
}
```

js/modules/k6/experimental/csv/module.go
```diff
@@ -0,0 +1,289 @@
...
+
+	if obj == nil {
+		return options, nil
+	}
+
+	if v := obj.Get("delimiter"); v != nil {
+		delimiter := v.String()
+
+		// A delimiter is gonna be treated as a rune in the Go code, so we need to make sure it's a single character.
+		if len(delimiter) > 1 {
```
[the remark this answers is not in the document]
joanlopez: Fair point! Perhaps we should add a test with that scenario, and evaluate how could be handled?

Unless it's super tricky, because in such case I'd suggest to ship an initial release explicitly stating that limitation. But I hope that's not the case.

js/modules/k6/experimental/csv/module.go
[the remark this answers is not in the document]
olegbespalov: My bad! You're right, so then I'm resolving the tread, but @oleiade feel free to incorporate the @joanlopez 's proposal

js/modules/k6/experimental/csv/module.go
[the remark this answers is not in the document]
olegbespalov: I also don't see an issue in having two instances of data. But the current version relies on the global root's module state but initializes it in the module instance, the suggestion to move is relevant. Otherwise, we could just move the data to the instance :shrug:

examples/experimental/csv-parse.js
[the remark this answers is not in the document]
oleiade: I'm unsure I fully understood the initial point about shared arrays and their constructor not being async. Can you clarify what the underlying concern is, especially from the user's perspective? 🙇‍♂️

Under those premice, I have a somewhat firm opinion that we should keep the `CSV.parse` name as it does not emphasize the return type and leaves us free to change it in the future if we wish. Also, from a user perspective, `csv.parse` says what it does without any further context and ensures sensible defaults., whereas `csv.sharedArray` or alternative implies I have to have or acquire the context of what that is and why I would want to use it.

js/modules/k6/experimental/csv/module.go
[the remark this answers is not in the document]
oleiade: To make sure I don't misunderstand, the alternative would be to run the following (or equivalent) in the `RootModule` constructor?
```
	globalThis.data = require("k6/data");
```

js/modules/k6/experimental/csv/module.go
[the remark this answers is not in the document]
oleiade: To be sure I don't misunderstand, a matching delimiter would be something along the lines of `\t,` or `\t,\t`?

js/modules/k6/data/data.go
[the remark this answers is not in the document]
oleiade: Renamed to `RecordReader` for explicitness (if we implemented support for retrieving data from a JSON array for instance, "records" would make more sense than "line")

js/modules/k6/experimental/csv/module.go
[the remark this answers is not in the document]
olegbespalov: Okay, after re-checking the code, I think I missed that `Data` requires the VU for initialization. TBH, I felt that it was just the shared memory, then yeah, probably for a single initialization, the current way of initialization, even if it looks hacky, is the only option. So closing the tread :+1:

js/modules/k6/experimental/csv/module.go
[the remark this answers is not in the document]
olegbespalov: @oleiade I mean, the delimiter can have a length of more than one rune like in the case of tab-separated CSV, it's `\t`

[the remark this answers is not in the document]
oleiade: Then I'm not sure this is an actual problem, as `\t` is a single character and fits in a single run (ascii decimal code 9) 👀 Or am I missing something?

```go
fmt.Println(rune('\t'))  // is valid
```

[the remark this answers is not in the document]
olegbespalov: @oleiade yes, you're right, resolving the tread :+1:

js/modules/k6/experimental/csv/module.go
```diff
@@ -0,0 +1,285 @@
...
+		Named: map[string]any{
+			"parse":  mi.Parse,
+			"Parser": mi.NewParser,
+		},
+	}
+}
+
+// Parser is a CSV parser.
+type Parser struct {
+	CurrentLine atomic.Int64
```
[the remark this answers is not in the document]
oleiade: I think the original intention was to allow users to check which line of the file the parser is currently parsing, but in hindsight, we should probably wait for a concrete use-case to make this public 👍🏻 Will make it private.

js/modules/k6/experimental/csv/module.go
```diff
@@ -0,0 +1,285 @@
...
+		reject(fmt.Errorf("first arg doesn't appear to be a *file.File instance, it's %T", file))
+		return promise
+	}
+
+	parserOptions := newDefaultParserOptions()
+	if options != nil {
+		var err error
+		parserOptions, err = newParserOptionsFrom(options.ToObject(rt))
+		if err != nil {
+			reject(fmt.Errorf("encountered an error while interpreting Parser options; reason: %w", err))
```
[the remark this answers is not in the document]
oleiade: "I am the alpha and the omega, the generator of bugs" 🤖

js/modules/k6/experimental/fs/file.go
```diff
@@ -8,25 +8,28 @@ import (
 
 // file is an abstraction for interacting with files.
 type file struct {
-	path string
+	path string `js:"path"`
```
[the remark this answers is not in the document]
oleiade: We made this change a long time ago, but if my memory is correct, this is to ensure that although the struct and those fields are private, they are correctly serialized in the object at the JS runtime level; so that when the public object `Stat()` method needs access to it, it does 🙇🏻

js/modules/k6/data/data.go
```diff
@@ -90,7 +93,55 @@ func (d *Data) sharedArray(call sobek.ConstructorCall) *sobek.Object {
...
+			common.Throw(rt, fmt.Errorf("failed to marshal record; reason: %w", err))
+		}
+
+		arr = append(arr, string(marshaled))
+	}
+
+	d.shared.mu.Lock()
+	defer d.shared.mu.Unlock()
+	array := sharedArray{arr: arr}
+	d.shared.data[name] = array
```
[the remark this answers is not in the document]
oleiade: I've create a `set` method on `sharedArrays` which we use instead 👍🏻

olegbespalov (approved):
The module is good enough to start using as an experimental :+1:

examples/experimental/csv-parse.js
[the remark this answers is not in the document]
joanlopez: To be honest, I don't fully see the problems mentioned above, because:

> The main point here is that this is not as users used SharedArray till now. We never had a function returning a SharedArray directly. Instead, we required them to use an array and generate it from the constructor.

I only see it as a different (new) way to initialize a `SharedArray`, but nothing changes with respect on how it is used later in the code, I guess. And, although it's true that would be better to have a way to use the `csv` module with the "standard" way of constructing a `SharedArray`, I don't see it like a problem.

> In addition, it has the downside that the user can't change/modify the array before storing it in the SharedArray. I don't know if it is a real case and if we should care about it, just raising the question.

Imho, this is likely something to work on in the future, as a second iteration of this package. Like, in general I don't see the _immutability_ of the parsed records as a problem, instead I'd just like to be able to give some further options to the users, or build some abstraction on top. Moreover, unless I'm wrong, `SharedArray`s are read-only, so it I see it as another sign that it fits the use case.

<br/>
All that said, I think it is fine enough to keep it as-is, also considering this is going to be _experimental_ for a while.

examples/experimental/csv-parse.js
```diff
@@ -0,0 +1,24 @@
...
+})();
+
+
+export default async function() {
+	console.log(csvRecords[scenario.iterationInTest])
```
joanlopez, line 22: Nit; I think it would be nice to add a comment here that gives a clue on how `csvRecords` looks like. Like, what does `csvRecords[scenario.iterationInTest]` returns? An array of strings? 🤔

examples/experimental/csv-parser.js
```diff
@@ -0,0 +1,22 @@
...
+	parser = new csv.Parser(file);
+})();
+
+export default async function() {
+	const {done, value} = await parser.next();
+	if (done) {
+		throw new Error("No more rows to read");
+	}
+
+	console.log(value);
```
joanlopez, line 21: Same as https://github.com/grafana/k6/pull/3743/files#r1749933652

js/jsmodules.go
```diff
@@ -4,6 +4,8 @@ import (
 	"errors"
 	"sync"
 
+	"go.k6.io/k6/js/modules/k6/experimental/csv"
```
joanlopez, line 7: Nit; Doesn't this needs to be part of the imports block below? 🤔

js/modules/k6/data/data.go
```diff
@@ -90,7 +93,61 @@ func (d *Data) sharedArray(call sobek.ConstructorCall) *sobek.Object {
...
+
+		arr = append(arr, string(marshaled))
+	}
+
+	return d.shared.set(name, arr).Wrap(rt).ToObject(rt)
+}
+
+// set is a helper method to set a shared array in the underlying shared arrays map.
+func (s *sharedArrays) set(name string, arr []string) sharedArray {
+	// FIXME (@oleiade): we should probably return an error if the name is empty?
```
joanlopez, line 144: Can't we do:

```
if name == "" {
	common.Throw(rt, errors.New("empty name provided to SharedArray's constructor"))
}
``` 

as in `NewSharedArrayFrom`? 🤔

js/modules/k6/data/share.go
```diff
@@ -20,7 +20,7 @@ type wrappedSharedArray struct {
 	parse    sobek.Callable
 }
 
-func (s sharedArray) wrap(rt *sobek.Runtime) sobek.Value {
+func (s sharedArray) Wrap(rt *sobek.Runtime) sobek.Value {
```
joanlopez, line 23: Nit; after a quick look, I haven't been able to identify the reason why need to expose this. If there's a reason, I guess it's but, but if there isn't, I'd prefer to keep it private so we reduce the surface. Especially considering the type is also a private one.

js/modules/k6/experimental/csv/module.go
```diff
@@ -0,0 +1,286 @@
...
+		reader:  r,
+		options: options,
+		vu:      mi.vu,
+	}
+
+	return rt.ToValue(&parser).ToObject(rt)
+}
+
+// Next returns the next row in the CSV file.
+func (p *Parser) Next() *sobek.Promise {
```
joanlopez, line 186: Nit; what happens if `Next` is called after done? Will it resolve to `{ "done": true } again? And `records`? Will be `nil`? 🤔

js/modules/k6/experimental/csv/module_test.go
```diff
@@ -0,0 +1,551 @@
...
+
+		_, err = r.RunOnEventLoop(wrapInAsyncLambda(fmt.Sprintf(`
+			  const file = await fs.open(%q);
+			  const parser = new csv.Parser(file);
+		`, testFilePath)))
+
+		require.NoError(t, err)
+	})
+
+	t.Run("constructing a parser with options valid options should succeed", func(t *testing.T) {
```
joanlopez, line 63:
```suggestion
	t.Run("constructing a parser with valid options should succeed", func(t *testing.T) {
```

js/modules/k6/experimental/csv/module_test.go
```diff
@@ -0,0 +1,551 @@
...
+		_, err = r.RunOnEventLoop(wrapInAsyncLambda(`
+			// Regardless of whether a file is passed, the parser should not be constructed in the VU context.
+			const parser = new csv.Parser(null);
+		`))
+
+		require.Error(t, err)
+		require.Contains(t, err.Error(), "csv Parser constructor must be called in the init context")
+	})
+
+	t.Run("constructing a parser in VU context should fail", func(t *testing.T) {
```
joanlopez, line 100: Looks like this test, and the one above are exactly the same, while I think both are testing different scenarios. By looking at the returned error, I think this is correct, and the one above should be corrected.

js/modules/k6/experimental/csv/reader.go
```diff
@@ -0,0 +1,113 @@
...
+	csv *csv.Reader
+
+	// currentLine tracks the current line number.
+	currentLine atomic.Int64
+
+	// options holds the reader's options.
+	options options
+}
+
+// NewReaderFrom creates a new CSV reader from the provided csv.Reader.
```
joanlopez, line 23:
```suggestion
// NewReaderFrom creates a new CSV reader from the provided io.Reader.
```

js/modules/k6/experimental/csv/reader.go
```diff
@@ -0,0 +1,113 @@
...
+		fromLineSet      = options.FromLine.Valid
+		toLineSet        = options.ToLine.Valid
+		skipFirstLineSet = options.SkipFirstLine
+	)
+
+	// If set, the fromLine option should be greater than 0.
+	if fromLineSet && options.FromLine.Int64 < 0 {
+		return nil, fmt.Errorf("the 'fromLine' option must be greater than 0; got %d", options.FromLine.Int64)
+	}
+
```
joanlopez, line 59: Should it be "greater" (`>`), or "greater or equal" (`>=`), because the condition doesn't seem to be matching the comment/error message.

js/modules/k6/experimental/csv/module.go
```diff
@@ -0,0 +1,286 @@
...
+
+	if v := obj.Get("fromLine"); v != nil {
+		options.FromLine = null.IntFrom(v.ToInteger())
+	}
+
+	if v := obj.Get("toLine"); v != nil {
+		options.ToLine = null.IntFrom(v.ToInteger())
+	}
+
+	if options.FromLine.Valid && options.ToLine.Valid && options.FromLine.Int64 > options.ToLine.Int64 {
```
joanlopez, line 281: This doesn't seem to match the equivalent in `csv/reader.go:66`, as there it seems to be checking `>=` while here it just checks `>`.

Personally, I think this is the correct one, because as `ToLine` is inclusive, if you just want to parse one line, you would set `FromLine` and `ToLine` to the line you want to read, but either this one or the other needs to be adjusted. I leave it up to you!

js/modules/k6/experimental/csv/reader_test.go
```diff
@@ -0,0 +1,174 @@
...
+func TestNewReaderFrom(t *testing.T) {
+	t.Parallel()
+
+	t.Run("instantiating a new reader with a nil io.Reader should fail", func(t *testing.T) {
+		t.Parallel()
+
+		_, err := NewReaderFrom(nil, options{})
+		if err == nil {
+			t.Error("NewReaderFrom() error = nil, wantErr true")
+		}
```
joanlopez, line 23: Shouldn't we use `require.Error` here, as we do for the rest of test cases? 🤔

js/modules/k6/experimental/fs/module.go
```diff
@@ -154,8 +178,14 @@ type File struct {
 	// Path holds the name of the file, as presented to [Open].
 	Path string `json:"path"`
 
-	// file contains the actual implementation for the file system.
-	file
+	// Impl contains the actual implementation of the file logic, and
```
joanlopez, line 181:
```suggestion
	// ReadSeekStater contains the actual implementation of the file logic, and
```

examples/experimental/csv-parse.js
oleiade: Good call 👍🏻 Adding a comment indeed 🙇🏻

js/modules/k6/data/data.go
oleiade: I was unsure how shared array would behave if provided an empty name under the hood, but after verification an empty name raises an error indeed 👍 I'll add the check 😃

js/modules/k6/data/share.go
oleiade: Good catch 👀 I'll give a shot at making it private then 👍🏻

js/modules/k6/experimental/csv/module.go
oleiade: Good question, let me look into it 👀

js/modules/k6/experimental/csv/reader.go
oleiade: Good catch indeed, switching to `> 0`, because I think `toLine == 0` doesn't make much sense from a user perspective 🙇🏻

oleiade: I think the logic is almost correct, but we should invert the logic to make the if statement and the error message easier to comprehend without turning our brains to spaghetti indeed.

I'm switching to add:
```
		fromLineIsPositive = fromLineSet && options.FromLine.Int64 > 0
		toLineIsPositive   = toLineSet && options.ToLine.Int64 > 0
```

In the var declaration above, and changing the if statements to `if !fromLineIsPositive` and `if !toLineIsPositive`.

js/modules/k6/experimental/csv/module.go
oleiade: I've added a test for this. When `{done: true}`, the `value` field will be an empty array 🙇🏻

joanlopez (approved):
Awesome! 🚀 💜

js/modules/k6/experimental/fs/file.go
```diff
@@ -48,7 +51,7 @@ func (f *file) Read(into []byte) (n int, err error) {
 
 	// Check if we have reached the end of the file
 	if currentOffset == fileSize {
-		return 0, newFsError(EOFError, "EOF")
+		return 0, io.EOF
```
olegbespalov, line 54: I might be wrong since I am missing a module context, but this seems like a breaking change. Since later on [there is logic that depends on this handling](https://github.com/grafana/k6/blob/master/js/modules/k6/experimental/fs/module.go#L242-L246) this error type.

oleiade: For this specific case, it isn't: this is the "private" file construct and, as far as I can tell, is never exposed directly to the JS runtime. By the looks of it, we have also updated all the places calling this code to expect `io.EOF` instead of an fsError too 🙇🏻

olegbespalov: Sorry, I wasn't clear; I meant that the breaking change is that previously, the read method in case of the EOF [resolves with null](https://github.com/grafana/k6/blob/master/js/modules/k6/experimental/fs/module.go#L243), whenever after the changes, it's probably resolved with [the EOF error](https://github.com/grafana/k6/blob/master/js/modules/k6/experimental/fs/module.go#L238).

olegbespalov: Anyway, keeping in mind that both modules are experimental I'm postponing for you @oleiade decision whenever it should be fixed, or when it should be fixed. The rest of PR looks good to me :+1:

olegbespalov (approved):

js/modules/k6/experimental/fs/file.go
oleiade: Oh, I see what you mean. Thanks for clarifying, indeed 👍🏻 

I reverified, and we're safe —no breaking changes— but I admit the code does not convey what happens clearly enough. 

The module's `Read` method callback checks if the returned error from `File.Read` is `io.EOF`, and if it is, it rewraps it as a `FsError` (which we do expose to the runtime) with its kind set to `EOFEror`. At a later point in the code, if the error is an `FsError` with its kind set to `EOFError`, then it returns null. So the behavior is unchanged.

We have a test to assert that already, but I'm taking note for the cooldown period and see if I can improve it and make it easier to read and understand 🙇🏻

PR #3746, opened 2024-05-17
Fix metrics for GRPC streams do not correctly include tags when stream ends

### LINKED ISSUE #3745: Metrics for GRPC streams do not correctly include tags when stream ends.
### Brief summary

When using GRPC streams in k6 when a stream is closed because of an error or normal execution a `grpc_req_duration` is emitted but it contains no information about the stream and it does not include custom tags provided when the stream was started.

### k6 version

0.51.0

### OS

Linux, MacOS

_No response_

1. Create a grpc stream, with custom tags
2. Close the stream on the server
3. Observe the `grpc_req_duration` metric does not include the url or method name, and does not include custom tags.

### Expected behaviour

Custom tags and the method name would be included in the metric.

### Actual behaviour

No custom tags or method name are included in the metric.

cchamplin:
## What?

Update the grpc stream being call to pass tags and params that include custom tags and method information

## Why?

Addresses https://github.com/grafana/k6/issues/3745

## Checklist

## Related PR(s)/Issue(s)

https://github.com/grafana/k6/issues/3745

Closes #3745

### DIFF
--- a/js/modules/k6/grpc/stream_test.go
+++ b/js/modules/k6/grpc/stream_test.go
@@ -10,6 +10,7 @@ import (
 
 	"go.k6.io/k6/lib/testutils/grpcservice"
 	"go.k6.io/k6/lib/testutils/httpmultibin/grpc_wrappers_testing"
+	"go.k6.io/k6/metrics"
 
 	"github.com/golang/protobuf/ptypes/wrappers"
 	"github.com/grafana/sobek"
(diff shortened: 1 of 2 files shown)


olegbespalov (commented):
Hi @cchamplin 

Good catch! Thank you for your contribution!

The change looks good to me, but it is also essential to introduce the test for it. It could be similar to something from the [cmd_run_grpc_test.go](https://github.com/grafana/k6/blob/master/cmd/tests/cmd_run_grpc_test.go), and the [getSampleValues](https://github.com/grafana/k6/blob/f947acb0bc57957d04d8e2a151c573fdc978e65e/cmd/tests/cmd_run_test.go#L1665-L1669) could be used to extract metrics/tags.

Do you think you could address this, or should we consider splitting a test into a separate issue or pull request?

olegbespalov:
@cchamplin just double-checking if you don't have the capacity for the tests? I could take over; one thing is, could you please maybe name the case? I mean which metrics missed the custom tags? Thanks!

joanlopez:
Pushing to https://github.com/grafana/k6/milestone/42 while waiting for news.

cchamplin:
@olegbespalov @joanlopez I have added the test as requested, let me know if that works.

js/modules/k6/grpc/stream_test.go
```diff
@@ -406,3 +407,107 @@ func TestStream_UndefinedHandler(t *testing.T) {
 
 	require.ErrorContains(t, err, "handler for \"data\" event isn't a callable function")
 }
+
+// TestStream_MetricsTagsMetadata tests that the metrics tags are correctly
+// added to samples.
+func TestStream_MetricsTagsMetadata(t *testing.T) {
```
olegbespalov, line 413: Seems that this test case passes even without changes [in stream.go](https://github.com/grafana/k6/pull/3746/commits/75be2cadd52216ffc89b397538c2b7a84709c10b#diff-d9e7a5b109fbc0ebc6d7bfa11c793707f250f4009d7bf24e7f545113c91aa64aR81) :thinking:

cchamplin: oops should be fixed now.

js/modules/k6/grpc/stream_test.go
```diff
@@ -406,3 +407,108 @@ func TestStream_UndefinedHandler(t *testing.T) {
...
+			// adding a delay to make server response "slower"
+			time.Sleep(200 * time.Millisecond)
+
+			if err := stream.Send(feature); err != nil {
+				return err
+			}
+		}
+
+		return nil
+	}
```
olegbespalov, line 448: What do you think about simplifying the test case? We test only tags, so it doesn't actually matter how big the response will be. Also, we don't need to emulate any delays (`time.Sleep`) here.

I mean to cut the logic to something like that:
```suggestion
	stub.listFeatures = func(_ *grpcservice.Rectangle, stream grpcservice.FeatureExplorer_ListFeaturesServer) error       {
		return stream.Send(&grpcservice.Feature{
			Name: "foo",
			Location: &grpcservice.Point{
				Latitude:  1,
				Longitude: 2,
			},
		})
	}
```

olegbespalov (commented):
@cchamplin great, thanks! I left the last comment about simplifying the logic. Apart from that, it seems green from my side :+1:

joanlopez:
Hey @cchamplin, 
Could you try to address what @olegbespalov suggested, please? 
Once done, I think we'll be ready to merge it. Thanks! 🙇🏻

olegbespalov (approved):
Okay, we could merge this as it's and update the test in a separate PR. Thanks for the contribution!

Issue #3748, opened 2024-05-20
Stop capturing panics and trying to continue execution after them.


mstoykov:
### What?

Drop the following code and it uses:

https://github.com/grafana/k6/blob/153e0e505807da9e330fda816ce6d63a4bbf4691/js/common/util.go#L67-L87

### Why?

The code was added many years ago when k6 had a bunch of really hard to fix race conditions. And was left as a safeguard.

Unfortunately it has two problems - it is not enough for most cases with multiple goroutines/async. And probably even worse - makes a bunch of potential bugs seem like less of an issue.

We have had at least one user come and ask us to have this be optional as it was problem during extension development.

Also, as far as I am aware it hasn't caught a single issue where k6 continuing has been a good idea for years. Some of the early things were nice, but now it seems like it just gets in the way.

### Suggestion:

Just remove the code and it's called site to be rewritten with no defers.

PR #3749, opened 2024-05-20
Don't print empty goja stacktraces on captured panics

mstoykov:
## What?

Stop printing an empty stacktrace of goja when capturing panics.

## Why?

This has been broken for years after some goja refactoring.

There doesn't seem to be a way for this to be fixed without a lot of boilerplate all over the code. And this functionality has not been used in the last few years in my experience.

## Checklist

### DIFF
--- a/js/common/util.go
+++ b/js/common/util.go
@@ -65,21 +65,15 @@ func ToString(data interface{}) (string, error) {
 }
 
 // RunWithPanicCatching catches panic and converts into an InterruptError error that should abort a script
-func RunWithPanicCatching(logger logrus.FieldLogger, rt *goja.Runtime, fn func() error) (err error) {
+func RunWithPanicCatching(logger logrus.FieldLogger, _ *goja.Runtime, fn func() error) (err error) {
 	defer func() {
 		if r := recover(); r != nil {
-			gojaStack := rt.CaptureCallStack(20, nil)
-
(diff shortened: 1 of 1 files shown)


js/common/util.go
```diff
@@ -65,21 +65,15 @@ func ToString(data interface{}) (string, error) {
 }
 
 // RunWithPanicCatching catches panic and converts into an InterruptError error that should abort a script
-func RunWithPanicCatching(logger logrus.FieldLogger, rt *goja.Runtime, fn func() error) (err error) {
+func RunWithPanicCatching(logger logrus.FieldLogger, _ *goja.Runtime, fn func() error) (err error) {
```
mstoykov, line 68: I specifically skipped on cleaning this up as it is currently in `js/common` package and I didn't want to introduce a potential breaking change.

Depending on the decision from #3748 we might delete it. 

If we decided against it - I will probably want to move this off `common` and in hte `js` package as I don't think this is functionality anyone but k6 should have.

joanlopez:
> This has been broken for years after some goja refactoring.

Do you mean that `gojaStack` is always empty? 🤔

mstoykov:
yes @joanlopez - have you noticed something different when there is a panic ?

joanlopez (approved):

oleiade (approved):

PR #3750, opened 2024-05-21
Refactor group and check to not be tightly coupled

mstoykov:
## What?

Group and check were tightly coupled between them and also to the end of test summary and the REST API.

This change the tight coupling is removed. Group and check no longer interact with each other while being run.

In order to keep the same end of test summary and REST API behavior the code makes the same aggregation it used to be but *only* based on the metrics emitted by `check` and `group`.

There are several breaking changes still in it as a bunch of things are no longer useful, and technically not implementable if we want to remove the tight coupling.

The only one that is relevant though is that
`lib.State` no longer has `Group`.

There are 2 extensions using that and both of them use it to use the "magic" that tight `group` and `check` together.

## Why?

The previous code apart from having problems code smell wise, meant that you can't implement this functionality in JS.

Which also means that they act very strangely especially as if you do not look closely it just seems like they set a few tags and emit metrics. 

While working on #2728 it became apparent that it is unlikely we will get a perfect solution and one of the big problems is that you can't even experiment without having to write a fairly complicated async go code.

So this also makes that issue a bit more actionable. And definitely a thing that now people can have not perfect or even 80%, but still workable for them solutions.

## Checklist

## Related PR(s)/Issue(s)

#2869

### DIFF
--- a/api/v1/group_routes_test.go
+++ b/api/v1/group_routes_test.go
@@ -66,7 +66,7 @@ func TestGetGroups(t *testing.T) {
 	t.Parallel()
 
 	cs := getControlSurface(t, getTestRunState(t, lib.Options{}, &minirunner.MiniRunner{}))
-	_ = cs.RunState.GroupSummary.Start()
+    require.NoError(t, cs.RunState.GroupSummary.Start())
 	cs.RunState.GroupSummary.AddMetricSamples([]metrics.SampleContainer{
 		metrics.Sample{
 			TimeSeries: metrics.TimeSeries{
@@ -87,7 +87,7 @@ func TestGetGroups(t *testing.T) {
 			},
 		},
 	})
-	_ = cs.RunState.GroupSummary.Stop()
+	require.NoError(t, cs.RunState.GroupSummary.Stop())
 
 	g0, err := lib.NewGroup("", nil)
 	assert.NoError(t, err)
--- a/api/v1/group_routes_test.go
+++ b/api/v1/group_routes_test.go
@@ -66,7 +66,7 @@ func TestGetGroups(t *testing.T) {
 	t.Parallel()
 
 	cs := getControlSurface(t, getTestRunState(t, lib.Options{}, &minirunner.MiniRunner{}))
-    require.NoError(t, cs.RunState.GroupSummary.Start())
+	require.NoError(t, cs.RunState.GroupSummary.Start())
 	cs.RunState.GroupSummary.AddMetricSamples([]metrics.SampleContainer{
 		metrics.Sample{
 			TimeSeries: metrics.TimeSeries{
--- a/api/v1/group_routes_test.go
+++ b/api/v1/group_routes_test.go
@@ -37,6 +37,7 @@ func getTestRunState(tb testing.TB, options lib.Options, runner lib.Runner) *lib
 		TestPreInitState: piState,
 		Options:          options,
 		Runner:           runner,
+		GroupSummary:     lib.NewGroupSummary(piState.Logger),
 		RunTags:          piState.Registry.RootTagSet().WithTagsFromMap(options.RunTags),
 	}
 }
@@ -64,14 +65,36 @@ func getControlSurface(tb testing.TB, testState *lib.TestRunState) *ControlSurfa
 func TestGetGroups(t *testing.T) {
 	t.Parallel()
 
+	cs := getControlSurface(t, getTestRunState(t, lib.Options{}, &minirunner.MiniRunner{}))
+	_ = cs.RunState.GroupSummary.Start()
+	cs.RunState.GroupSummary.AddMetricSamples([]metrics.SampleContainer{
+		metrics.Sample{
+			TimeSeries: metrics.TimeSeries{
+				Metric: cs.RunState.BuiltinMetrics.GroupDuration,
+				Tags:   cs.RunState.Registry.RootTagSet().With("group", "::group 1"),
+			},
+		},
+		metrics.Sample{
+			TimeSeries: metrics.TimeSeries{
+				Metric: cs.RunState.BuiltinMetrics.GroupDuration,
+				Tags:   cs.RunState.Registry.RootTagSet().With("group", ""),
+			},
+		},
+		metrics.Sample{
+			TimeSeries: metrics.TimeSeries{
+				Metric: cs.RunState.BuiltinMetrics.GroupDuration,
+				Tags:   cs.RunState.Registry.RootTagSet().With("group", "::group 1::group 2"),
+			},
+		},
+	})
+	_ = cs.RunState.GroupSummary.Stop()
+
 	g0, err := lib.NewGroup("", nil)
 	assert.NoError(t, err)
 	g1, err := g0.Group("group 1")
 	assert.NoError(t, err)
 	g2, err := g1.Group("group 2")
 	assert.NoError(t, err)
-
-	cs := getControlSurface(t, getTestRunState(t, lib.Options{}, &minirunner.MiniRunner{Group: g0}))
 
 	t.Run("list", func(t *testing.T) {
 		t.Parallel()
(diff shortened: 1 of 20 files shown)


lib/test_state.go
```diff
@@ -31,6 +33,118 @@ type TestRunState struct {
...
+// GroupSummary is an internal output implementation that facilitates the aggregation of
+// group and check metrics for the purposes of the end of test summary and the REST API
+type GroupSummary struct {
+	group   *Group // TODO(@mstoykov): move the whole type outside of lib later
+	ch      chan []metrics.SampleContainer
+	stopped chan struct{}
+	logger  logrus.FieldLogger
+}
+
+// NewGroupSummary returns new GroupSummary ready to be started.
```
mstoykov, line 55: I kind of wanted to move the whole Group, Check and GroupSummary somewhere else, but then I decided that is quite a big change and will make the review even harder.

It also turned out to not be easy.

So if we have any good places to move them I would prefer to do them in a separate PR.

lib/testutils/minirunner/minirunner.go
```diff
@@ -28,7 +28,6 @@ type MiniRunner struct {
 
 	SetupData []byte
 
-	Group        *lib.Group
```
mstoykov, line 31: This will break 2 extensions:

```
xk6-fasthttp/checks.go
41:     check, err := state.Group.Check("check status is " + strconv.FormatInt(int64(wantStatus), 10))

xk6-g0/g0/addon/assert.go
67:     check, err := state.Group.Check(name)
```

I don't really think that keeping this compatible is possible, and even more importantly warranted.

As you can see both extensions basically hit exactly the same issue with this being the way to create a new check instead of just emitting the metrics.

lib/vu_state.go
```diff
@@ -42,9 +42,6 @@ type State struct {
 	// Logger instance for every VU.
 	Logger logrus.FieldLogger
 
-	// Current group; all emitted metrics are tagged with this.
-	Group *Group
-
```
mstoykov, line 47: This will break 2 extensions:

```
xk6-fasthttp/checks.go
41:     check, err := state.Group.Check("check status is " + strconv.FormatInt(int64(wantStatus), 10))

xk6-g0/g0/addon/assert.go
67:     check, err := state.Group.Check(name)
```

I don't really think that keeping this compatible is possible, and even more importantly warranted.

As you can see both extensions basically hit exactly the same issue with this being the way to create a new check instead of just emitting the metrics.

lib/test_state.go
codebien: Ideally, I would use this opportunity to move them to an `internal` package. This is my unique request, reflecting what the comment is mentioning.

lib/test_state.go
```diff
@@ -31,6 +33,118 @@ type TestRunState struct {
...
+		groupName, ok := sample.Tags.Get(metrics.TagGroup.String())
+		if !ok {
+			return nil
+		}
+		group, err := buildGroup(gs.group, groupName)
+		if err != nil {
+			return err
+		}
+
+		if sample.Metric.Name != "checks" {
```
mstoykov, line 101:
```suggestion
		// In case of checks we need to also generate the check and add it values. Otherwise we are done.
		if sample.Metric.Name != "checks" {
```

lib/test_state.go
```diff
@@ -31,6 +33,118 @@ type TestRunState struct {
...
+			return nil
+		}
```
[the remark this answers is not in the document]
mstoykov: https://github.com/grafana/k6/pull/3750/files#r1611293199

lib/test_state.go
mstoykov: I am not against this - but that will create a much bigger change that will make following what happens ever harder. So I will prefer to keep it in a separate PR.

Not certain `internal` is needed as well so :shrug: on that particular part. But it is probably a good usage as well.

api/v1/group_routes_test.go
```diff
@@ -64,15 +65,37 @@ func getControlSurface(tb testing.TB, testState *lib.TestRunState) *ControlSurfa
 func TestGetGroups(t *testing.T) {
 	t.Parallel()
 
+	cs := getControlSurface(t, getTestRunState(t, lib.Options{}, &minirunner.MiniRunner{}))
+	_ = cs.RunState.GroupSummary.Start()
```
joanlopez, line 69:
```suggestion
require.NoError(t, cs.RunState.GroupSummary.Start())
```

Even if it cannot fail, it's easier to follow, as you don't need to check the shape of what's being ignored, and why it will never fail.

api/v1/group_routes_test.go
```diff
@@ -64,15 +65,37 @@ func getControlSurface(tb testing.TB, testState *lib.TestRunState) *ControlSurfa
...
+			},
+		},
+		metrics.Sample{
+			TimeSeries: metrics.TimeSeries{
+				Metric: cs.RunState.BuiltinMetrics.GroupDuration,
+				Tags:   cs.RunState.Registry.RootTagSet().With("group", "::group 1::group 2"),
+			},
+		},
+	})
+	_ = cs.RunState.GroupSummary.Stop()
```
joanlopez, line 90:
```suggestion
	require.NoError(t, cs.RunState.GroupSummary.Stop())
```

joanlopez: Same as above

lib/test_state.go
joanlopez: Yeah, especially with the v1.0 on the road, would be great to start considering hiding internal details behind an `internal` package, to reduce the public surface ( 👋🏻 say hello to [Hyrum's Law](https://www.hyrumslaw.com/)).

That said, I fully agree on doing so on a separate PR, to avoid making this even harder to read, follow and review.

joanlopez (approved):
Looks good despite of the minor comments!

js/modules/k6/k6_test.go
```diff
@@ -92,20 +92,16 @@ func TestGroup(t *testing.T) {
...
-			assert.Equal(t, state.Group.Name, "my group")
-			assert.Equal(t, state.Group.Parent, root)
 		}))
 		_, err := tc.testRuntime.RunOnEventLoop(`k6.group("my group", fn)`)
 		assert.NoError(t, err)
-		assert.Equal(t, state.Group, root)
 		groupTag, ok := state.Tags.GetCurrentValues().Tags.Get("group")
 		require.True(t, ok)
-		assert.Equal(t, groupTag, root.Name)
+		assert.Equal(t, groupTag, "")
```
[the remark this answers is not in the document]
joanlopez: Yes, you're right, sometimes you used `""`, sometimes the constant.

> but as changing this constant will be a breaking change I kind of prefer if changing it breaks a few tests for sure

I fully agree with end goal, it was me just probably confused about where and for which tests it makes more sense to use the constant and for which not. Let's keep it as-is.

That said, as I had previous negative experiences with this kind of tradeoff: "de-coupling" the test from the constant, to notice any change, would make sense to also leave a comment together with the constant, stating that explicitly, for the ourselves of the future, who might forget/omit it by just updating the tests anyway?

joanlopez (approved):
Approved! ✅ Although I guess we need a to reach a consensus around the breaking changes to move forward. Nevertheless, I don't see many alternatives at this point.

oleiade (approved):
LGTM 🚀 👏🏻 

I concur with what has been said about starting an `internal` package too, as it's inline with our ongoing plans for v1.

lib/models.go
```diff
@@ -18,8 +18,11 @@ import (
 // GroupSeparator for group IDs.
 const GroupSeparator = "::"
 
+// RootGroupPath is the id of the root group
```
mstoykov, line 21:
```suggestion
// RootGroupPath is the id of the root group
//
// Note(@mstoykov): the constant shouldn't be used in all tests in order to not couple the tests too much with it.
```

cc @joanlopez

joanlopez: What I thought initially was more something like telling the future reader that changing this value would represent a breaking change, from which what you suggested now can be deduced from, I guess.

PR #3765, opened 2024-06-04
Get rid of succ & fail marks on rate summary

joanlopez:
## What?

It replaces the use of succ & fail marks (✓/✗) for rate metrics on summary, in favor of the text: `{success} out of {total}`.

**Note:** the check summary remains the same, as for checks there's no ambiguity on the use of ✓ and ✗.

## Why?

Because as discussed in https://github.com/grafana/k6/issues/2306 and https://github.com/grafana/k6/pull/3656, the current form is _sometimes_ confusing, because the metric may have either a "positive" meaning (like successful requests) or a "negative" meaning (like failed requests), for which the use of ✓ may result in confusions, especially for the latter.

## Checklist

## Related PR(s)/Issue(s)

Closes https://github.com/grafana/k6/issues/2306

Second iteration of https://github.com/grafana/k6/pull/3656, based on what was discussed there.


js/summary_test.go
```diff
@@ -23,9 +23,9 @@ const (
 		"       ✓ check1\n" +
 		"       ✗ check3\n        ↳  66% — ✓ 10 / ✗ 5\n" +
 		"       ✗ check2\n        ↳  33% — ✓ 5 / ✗ 10\n\n" +
-		"   ✓ checks......: 75.00% ✓ 45  ✗ 15 \n"
+		"   ✓ checks......: 75.00% 45 out of 60\n"
```
olegbespalov, line 26: It seems inconsistent with the `check1`, `check2`, and `check3` since they still use icons. Do you think so?

joanlopez: I think the reason is because one is stored as a `Rate` metric, so it is handled as such, while the other falls under the `summarizeCheck` path.

That said, as I mentioned in the description, I didn't modify `summarizeCheck` because I feel that the use of icons for that may still be useful.

However, it's true that once all together, it looks a bit inconsistent.
So, would you suggest to use `out of` everywhere?
cc/ @oleiade @codebien @mstoykov  (as you already shared some feedback before around this discussion) 🙇🏻

codebien: Isn't still dependent from what is the meaning of the check? For example if `check3` would be something like `whatever_metric_failed` wouldn't be still confusing?

joanlopez: Well... yes and no 😅

Technically, a check that succeeds is always ✓, while a rate metric is just a number.
Think analogously in programming terms, if the code execution enters into an `if` statement, it's always because the condition is satisfied, even if the condition reflects a negative event (e.g. `if failed { ... }`), but a rate is just a variable that holds a number, where the number might reflect succeeds or fails. So, the ambiguity is different/smaller.

That said, I agree that could make sense to change it everywhere. So, let's open the discussion widely and see what other people prefers, honestly I don't have ~strong~ preferences.

oleiade:
This PR aligns with our goal and designs to refactor the end-of-test summary. As a matter of fact, we integrated its approach in our designs. From my perspective, this is good to go and merge, in principle 👍🏻

mstoykov (approved):

oleiade (approved):

mstoykov (approved):

olegbespalov (approved):

joanlopez:
We (@oleiade and I, among others) have been discussing details like what's mentioned in https://github.com/grafana/k6/pull/3765#discussion_r1627148359, and others covered in the design doc, like grouping metrics (http, network, execution, etc) and we agreed to merge this, as it already brings some value to those users that still nowadays are getting confused with the current format using ✓ and ✗ for _rate_ metrics (see [this] example), and is compatible with changes that will come in the near future. 

So, we ship some value in the next/current release, and iterate later, instead of waiting till the end result to ship the value.

Besides, I'm not adding the `breaking-change` label, as we don't consider this summary format to be part of the expectations associated with version compatibility, while others could (e.g. machine readable ones, like JSON).

PR #3768, opened 2024-06-04
Update OTel deps

joanlopez:
## What?

Updates OTel dependencies.

## Checklist


go.mod
```diff
@@ -1,6 +1,8 @@
 module go.k6.io/k6
 
-go 1.20
```
joanlopez, line 3: I can separate this on another PR, but apparently it seems to be required now?
I guess it's related with https://go.dev/doc/toolchain

joanlopez:
After updating the `go-grpc` package, some linter warnings (deprecated) appeared: 
```
Error: SA1019: grpc.DialContext is deprecated: use NewClient instead.  Will be supported throughout 1.x.  (staticcheck)
Error: SA1019: grpc.WithBlock is deprecated: this DialOption is not supported by NewClient. Will be supported throughout 1.x.  (staticcheck)
Error: SA1019: grpc.FailOnNonTempDialError is deprecated: this DialOption is not supported by NewClient. This API may be changed or removed in a later release.  (staticcheck)
Error: SA1019: grpc.WithBlock is deprecated: this DialOption is not supported by NewClient. Will be supported throughout 1.x.  (staticcheck)
Error: SA1019: grpc.FailOnNonTempDialError is deprecated: this DialOption is not supported by NewClient. This API may be changed or removed in a later release.  (staticcheck)
Error: SA1019: grpc.WithReturnConnectionError is deprecated: this DialOption is not supported by NewClient. Will be supported throughout 1.x.  (staticcheck)
Error: SA1019: grpc.DialContext is deprecated: use NewClient instead.  Will be supported throughout 1.x.  (staticcheck)
```

See more details [here](https://github.com/grpc/grpc-go/blob/master/Documentation/anti-patterns.md).

Is it okay for you if I disable the linter for those (`//nolint`), and create a issue to track that, and address it in a separate PR?

mstoykov:
Given that browser is the only oTel user at the moment I would argue this kind of falls on them to update it - I am pretty sure we did it the last time.

joanlopez:
> Given that browser is the only oTel user at the moment I would argue this kind of falls on them to update it - I am pretty sure we did it the last time.

Do you mean that stuff under `lib/trace` (which apparently uses OTel deps) is owned by @grafana/k6-browser?

If so, and we want to leave that update to them;do you want me to create another PR updating all except OTel? 
Note that some _interconnected_ dependencies (like `x/net` and `go-grpc` are being updated as well).

mstoykov:
Basically yes - there is no other usage of otel in core but to facilitate xk6-browser. As such it will be nice for us to not bump it and then to turn out we broke something that they would've caught testing it.

I also do prefer to update dependencies separately, so it is more obvious what the changes are between dependencies. 

This sometimes mean trying to update them in order of having most dependents not from having most dependencies `x/*` usually is the good start.

ankur22:
Thanks for the nudge. Do we usually update dependencies after the release or just before it?

joanlopez:
> Thanks for the nudge. Do we usually update dependencies after the release or just before it?

Based on what's stated in https://github.com/grafana/k6/blob/master/Dependencies.md, it's after the release.
But I think we haven't updated them since v0.51.0 was released, that's why I'm doing so.
But please, correct me if I'm wrong 🙏🏻

PR #3774, opened 2024-06-06
Fix require warning with tests from stdin

mstoykov:
## Checklist


mstoykov:
This specifically fixes https://github.com/grafana/k6/issues/3534#issuecomment-2135045438 and related problems

oleiade (approved):

js/bundle_test.go
```diff
@@ -73,6 +73,34 @@ func getSimpleBundle(tb testing.TB, filename, data string, opts ...interface{})
...
 
+func getSimpleBundleStdin(tb testing.TB, pwd *url.URL, data string, opts ...interface{}) (*Bundle, error) {
+	fs := fsext.NewMemMapFs()
+	var rtOpts *lib.RuntimeOptions
+	var logger logrus.FieldLogger
+	for _, o := range opts {
+		switch opt := o.(type) {
+		case fsext.Fs:
+			fs = opt
+		case lib.RuntimeOptions:
```
joanlopez, line 84: Nit; out of curiosity - if we normally use `*lib.RuntimeOptions`, is there any reason why not using it here instead of the value? 🤔

joanlopez (approved):

js/bundle_test.go
mstoykov: Do you mean instead of using `opts`?

We can - but this is a copy of the helper function just above here that does a *very* similar thing, but doesn't go through stdin. 

I kind of prefered to not rewrite the whole funciton, and me trying to add it to the above was pretty painful and ended up being way bigger change.

So in the interest of not having to have a huge change (across everything using the above helper) - I decided to copy it and leave everything but what I need as it is.
If we add more tests for executing from stdin - this will likely need most of the other options and stuff.

joanlopez: No, no, I was asking value vs ref.

mstoykov: oh. Looking at it we basically have a pointer ... for no reason. It is used to check if it was set - but then we just set it to the empty ones, so IMO we probably can move to a not pointer variant.

Issue #3773, opened 2024-06-06
Moving to a goja fork


mstoykov:
## What?

This is an issue for the actual work that needs to be done around #3772. 

## Problem

Due to the fact we are doing a fork and how golang modules work we need to somehow tell the go toolchain that it now needs to use the goja fork from *this* place.

### fork and use `replace`

First this means we need to keep it called `goja` and it means anyone else using this fork will have to forever use `replace`, unless we break again and move to one of the below methods.

Due to how `xk6` works it also means that we will need to teach it to add `replace` for goja and figure out the correct version. Which also means that in order for people to continue using xk6 they need to know they have to upgrade it. 

The only upside of this option is that we don't need to change much code.

At least until we do any visible changes to the API (new functions, types, changing some of them) then every extension also needs to put `replace` clause in its repo *as well*.

Given the above while it seems like the "easiest" option it will continue to haunt us forever. So I personally think this is a bad idea.

### fork and switch

The obvious thing is to name the new module something and then go over the code and change it to use that new module.

This would break every extension the moment it has to be compiled with this new k6 - and they will need to the exact same thing.

The change is arguably fairly easy and can be done with `regex` maybe [gopatch](https://github.com/uber-go/gopatch) for some parts.

We can also make PRs for every extension as we have done [some times before](https://github.com/mstoykov/xk6-extensions/blob/main/build-with-v0.45.1/update.sh) 

Unfortunately due to how experimental modules are implemented the original PR for this within k6 will be quite the nightmare and will need to happen for all extensions. This potentially might block people working on other work and likely will be quite hard to revert if we end up needing to pause the whole thing.

As such if we are going to be doing this I would recommend doing it early in the release.

### alias, switch and then fork

The alternative is to make a new module with the new name and instead of putting the goja code in - we [alias](https://go.dev/ref/spec#Alias_declarations) every type and copy-paste the consts and variables. 

This way you can now point k6 code to this new module - it will work. But at the same time every extension will be able to build as for go those are the same types and the functions and variables and consts get matched by values. 

This than lets us having a much smaller initial PR for k6, go through all the experimental modules within the next days/weeks, update them to the new module and update them in k6.

We can do the same for extensions, and they will work with both old k6 but also the version of k6 we release with this.

After 1 release of k6 we make new version of the module *now* with the actual goja code - actually forking it. At this point updating it in k6 will just work and as long as extensions have actually updated - it will also work for them. 

[...]

Issue #3772, opened 2024-06-05
Fork the goja project


sniku:
### Feature Description

Create a fork of the [goja](https://github.com/dop251/goja/) project under the Grafana GitHub organization to accelerate the development speed and unblock the ESM work.

Goja, the JS runtime is the biggest k6 dependency. While the project is stable and mostly feature-complete, it has only one time-constrained maintainer. As goja is the cornerstone of k6, and a dependency of hundreds of xk6 extensions, the k6 core team needs an ability to develop changes at a higher speed than is possible today. 

This is especially important now that we are preparing for the v1.0.0 stable release. 

The fork will unblock merging of the [Modules support](https://github.com/dop251/goja/pull/430) PR, the possible future work on a debugger, and other changes.
We should do our best not to diverge from master/goja more than necessary. We should upstream enhancements whenever possible and work with @dop251, the maintainer, to keep the project in a healthy state.

PR #3775, opened 2024-06-06
Switch to our fork of goja - sobek

mstoykov:
## What?

Move to sobek from goja - currently only aliasing types over from goja.

## Why?

This is part of #3772 and #3773.

The PR also doesn't try to fix every other reference of goja - a bunch of them are docs that will break at this point, and will likely need to be updated over time. Likely after sobek is an actual fork of goja.

This is marked as breaking change mostly because I do expect it will still break something, even if not for most users.

## Checklist


js/modules/k6/experimental/streams/readable_streams.go
```diff
@@ -62,9 +62,10 @@ func (stream *ReadableStream) Cancel(reason goja.Value) *goja.Promise {
 // GetReader implements the [getReader] operation.
 //
 // [getReader]: https://streams.spec.whatwg.org/#rs-get-reader
-func (stream *ReadableStream) GetReader(options *goja.Object) goja.Value {
+func (stream *ReadableStream) GetReader(options *sobek.Object) sobek.Value {
 	// 1. If options["mode"] does not exist, return ? AcquireReadableStreamDefaultReader(this).
-	if options == nil || common.IsNullish(options) || options.Get("mode") == nil || goja.IsUndefined(options.Get("mode")) {
+	if options == nil || common.IsNullish(options) ||
+		options.Get("mode") == nil || sobek.IsUndefined(options.Get("mode")) {
```
mstoykov, line 68: This unfortunately becomes 121 symbols and also both cases hit the "is interface nil?" trap so ... yeah. This is the best I can do without having to add a lot mroe code.

codebien (commented):
Am I getting it right? Isn't anything else than `s/goja/sobek/`, right?

mstoykov:
Technically it is ... as I am leaving the parser and ast packages from goja in js/compiler. Otherwise it needs a lot more aliases, which won't matter for any other codebase - no extension ever uses those.

And also I replace `goja.` with `sobek.` as to not break a bunch of comments referncing issues and docs for goja.

Both of those will be updated in the future - one of them when we move actually to a fork, the others when we have time as part of cleaning. But I didn't want to figure out all the cases for each comment individually.

joanlopez (approved):
All the replacements looks good, and I have also confirmed that the only remaining uses of `github.com/dop251/goja` are for `goja/parser` and `goja/file`, expected as described in the issue. 

So, I guess I could say it looks good, as far as I can tell 😄

codebien (approved):

joanlopez:
Hey! While working on another PR/branch, I realized that for ongoing changes, it's pretty easy to forget about goja imports, as we'll keep the dependency for a while. 

Do you think worths adding a temporary linter that forbids those imports?
cc/ @mstoykov

PR #3776, opened 2024-06-06
Handling undefined function in timers

olegbespalov:
## What?

This PR fixes a panic that could be caused by a script like:

```js
export default function() {
	setInterval(undefined, 100);
}
```

## Why?

We shouldn't panic

## Checklist

### DIFF
--- a/js/modules/k6/timers/timers.go
+++ b/js/modules/k6/timers/timers.go
@@ -2,12 +2,14 @@
 package timers
 
 import (
+	"fmt"
 	"time"
 
 	"github.com/mstoykov/k6-taskqueue-lib/taskqueue"
 	"github.com/sirupsen/logrus"
 
 	"github.com/grafana/sobek"
+	"go.k6.io/k6/js/common"
 	"go.k6.io/k6/js/modules"
 )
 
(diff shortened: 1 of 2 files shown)


js/modules/k6/timers/timers.go
```diff
@@ -126,7 +127,16 @@ func (e *Timers) timerInitialization(
...
 	}
 
+	name := setTimeoutName
+	if repeat {
+		name = setIntervalName
+	}
+
 	task := func() error {
+		if callback == nil {
+			return fmt.Errorf("%s's callback isn't a callable function", name)
```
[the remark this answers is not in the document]
joanlopez: To be honest, I had the same question when looking at this code 🤔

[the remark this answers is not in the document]
olegbespalov: Good point, I 'll adjust accordingly 👍

mstoykov (approved):

js/modules/k6/timers/timers.go
```diff
@@ -126,6 +128,15 @@ func (e *Timers) timerInitialization(
 		timeout = 0
 	}
 
+	name := setTimeoutName
+	if repeat {
+		name = setIntervalName
+	}
+
+	if callback == nil {
```
joanlopez, line 136: Should we be even more defensive and assert `callback` is actually "callable"?
What if the value is not callable? Would it be handled as a JS error/exception, or as an _uncaptured_ Go panic? 🤔

olegbespalov: My understanding is that it's either nil or callable since we already use here sobek.Callable type. For instance, sobek.AssertCallable works with Dobel.Value only.

joanlopez: Ah yeah, right, my fault! Apologies! 🙏🏻

joanlopez (approved):

PR #3777, opened 2024-06-06
Stop catching panics

### LINKED ISSUE #3748: Stop capturing panics and trying to continue execution after them.

olegbespalov:
## What?

We stop catching the panic in k6. 

## Why?

Catching the panic could prevent us/customers from identifying them early. Instead, we should deliver the fixes for the panic cases.

## Checklist

## Related PR(s)/Issue(s)

Closes #3748


mstoykov (dismissed):
LGTM!

Have you checked if some extension doesn't use it - maybe we can tell them.

codebien (approved):
LGTM

We should add the breaking change label

joanlopez (dismissed):

olegbespalov:
>Have you checked if some extension doesn't use it - maybe we can tell them.

A quick search shows that only forks, but I'll check better later on :+1:

mstoykov (approved):
LGTM!

I did check through my https://github.com/mstoykov/xk6-extensions/tree/main/checkout-all-extensions as I was working around it either way - no usage were found.

codebien (approved):

PR #3779, opened 2024-06-07
Fixes `undefined` handler passing to the gRPC stream

olegbespalov:
## What?

This PR fixes a panic that happens if pass undefined (null) handler to the `stream.on`

E.g.
```js
stream.on("data", undefined)
```

## Why?

We shouldn't panic

## Checklist

### DIFF
--- a/js/modules/k6/grpc/stream.go
+++ b/js/modules/k6/grpc/stream.go
@@ -293,9 +293,13 @@ func (s *stream) processSendError(err error) {
 	})
 }
 
-// on registers a listener for a certain event type
-func (s *stream) on(event string, listener func(sobek.Value) (sobek.Value, error)) {
-	if err := s.eventListeners.add(event, listener); err != nil {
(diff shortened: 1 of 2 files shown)


mstoykov (approved):

joanlopez:
Similarly to what I left [here](https://github.com/grafana/k6/pull/3776#discussion_r1630248979), this time I think we do actually need an extra check for the sake of correctness. Right? 🤔

joanlopez:
> Similarly to what I left [here](https://github.com/grafana/k6/pull/3776#discussion_r1630248979), this time I think we do actually need an extra check for the sake of correctness. Right? 🤔

I quickly checked that, and it returns a `TypeError`. And, although I still think the error should look pretty similar to the one returned when the handler is `undefined`, I'm gonna approve the PR because that corner case doesn't produce a `panic`, so we can address it later, in another PR.

joanlopez (approved):

PR #3792, opened 2024-06-11
Move to the forked Sobek code

mstoykov:
## What?

Move to Grafana fork of goja called Sobek - now with the actual code not just aliases

## Why?

See #3772 

## Checklist


codebien (dismissed):

oleiade (dismissed):
🎉 🚀

codebien (dismissed):

cmd/runtime_options.go
```diff
@@ -24,7 +24,7 @@ func runtimeOptionFlagSet(includeSysEnv bool) *pflag.FlagSet {
 	flags.Bool("include-system-env-vars", includeSysEnv, "pass the real system environment variables to the runtime")
 	flags.String("compatibility-mode", "extended",
 		`JavaScript compiler compatibility mode, "extended" or "base" or "experimental_enhanced"
-base: pure goja - Golang JS VM supporting ES5.1+
+base: pure sobek - Golang JS VM supporting ES5.1+
```
codebien, line 27:
```suggestion
base: pure Sobek - Golang JS VM supporting ES5.1+
```
Can we have it with a capitalized letter? I see this across multiple lines across this pull request. Consider this suggestion for all of them.

mstoykov: I am not strictly against this, but we do not capitalize it anywhere else as well as far as I can see - and we do not really do it for our other projects.

codebien: You did it for the first file `Dependencies` :sweat_smile: 

Btw, I don't think we have other projects with a name independent from `k6`. All repositories are prefixed by `k6` or some combination of it like `xk6` where doesn't make sense to capitalize. But in this case, it is a proper name, so I don't see why we shouldn't as it is the correct grammar.

Of course, I expect this only for text and not for code (e.g repo, package names, etc..)

mstoykov: I made that lower case as well :sweat_smile:

codebien (approved):

oleiade (approved):

Issue #3778, opened 2024-06-07
Browser: Register sync and async JS APIs (k6-core side)


inancgumus:
# What

Expose two JS import paths for the browser module:
* `k6/browser` -> The new async API
* `k6/experimental/browser` -> The old sync API

# Why

The browser team has been working on migrating the existing API to [Async](https://github.com/grafana/xk6-browser/issues/428).

With v0.52.0's release date two weeks away, we're confident that the new Async API can be delivered in one shot. Since these changes are the last piece of the API design and we don't anticipate any breaking changes in the near future, it's worth discussing _the graduation of the experimental Browser API to a stable module_.

Here are a few reasons that brought us to this idea:
* With the async API, we're aligned with Playwright, and we don't expect any further breaking changes to bridge this gap.
* Releasing the new API may make future changes in the import path more challenging.
* Releasing the async version of the API under a separate import path, `k6/browser`, while keeping `k6/experimental/browser` for a while during the deprecation window is a non-breaking change for existing user scripts.

PR #3793, opened 2024-06-11
Graduate to `k6/browser` from an experimental module

inancgumus:
## What?

Provides two JS import paths:
* `k6/experimental/browser`: The k6 browser sync API shipped in k6 0.51 (which we're planning to remove in k6 0.54).
* `k6/browser`: The k6 browser async API.

Closes: #3778

## Why?

* Large breaking changes are no longer expected.
* [Async API](https://github.com/grafana/xk6-browser/issues/428) w/Playwright compatibility and API stability.
* Major runtime stability issues are fixed (although there are still some minor issues).
* Tested on the cloud, and no regressions.

## Checklist

## Related PR(s)/Issue(s)

- grafana/xk6-browser#428
- grafana/xk6-browser#1117

### DIFF
--- a/cmd/tests/cmd_run_test.go
+++ b/cmd/tests/cmd_run_test.go
@@ -2262,7 +2262,7 @@ func TestBrowserPermissions(t *testing.T) {
 			name:             "browser option not set",
 			options:          "",
 			expectedExitCode: 0,
-			expectedError:    "GoError: browser not found in registry. make sure to set browser type option in scenario definition in order to use the browser module",
+			expectedError:    "browser not found in registry. make sure to set browser type option in scenario definition in order to use the browser module",
 		},
 		// When we do supply the correct browser options,
 		// we expect that the browser module will start
@@ -2300,12 +2300,12 @@ func TestBrowserPermissions(t *testing.T) {
 		t.Run(tt.name, func(t *testing.T) {
 			t.Parallel()
 			script := fmt.Sprintf(`
-			import { browser } from 'k6/experimental/browser';
+			import { browser } from 'k6/browser';
 
 			%s
 
 			export default function() {
-			  browser.isConnected();
+			  browser.isConnected()
 			};
 			`, tt.options)
 
--- a/cmd/tests/cmd_run_test.go
+++ b/cmd/tests/cmd_run_test.go
@@ -2319,6 +2319,39 @@ func TestBrowserPermissions(t *testing.T) {
 	}
 }
 
+func TestBrowserExperimentalImport(t *testing.T) {
+	t.Parallel()
+
+	const script = `
+		import { browser } from 'k6/experimental/browser';
+
+		export const options = {
+			scenarios: {
+					browser: {
+					executor: 'shared-iterations',
+					options: {
+						browser: {
+							type: 'chromium',
+						},
+					},
+				},
+			},
+		}
+
+		export default function() {
+			browser.isConnected()
+		};
+	`
+
+	const wantExitCode = 108
+	ts := getSingleFileTestState(t, script, []string{}, wantExitCode)
+	ts.Env["K6_BROWSER_EXECUTABLE_PATH"] = "k6-browser-fake-cmd"
+	cmd.ExecuteWithGlobalState(ts.GlobalState)
+	loglines := ts.LoggerHook.Drain()
+
+	assert.Contains(t, loglines[0].Message, "use k6/browser instead of k6/experimental/browser")
+}
+
 func TestSetupTimeout(t *testing.T) {
 	t.Parallel()
 	ts := NewGlobalTestState(t)
--- a/js/jsmodules.go
+++ b/js/jsmodules.go
@@ -49,7 +49,7 @@ func getInternalJSModules() map[string]interface{} {
 		"k6/experimental/browser": newWarnExperimentalModule(browser.NewSync(),
 			"Please update your imports to use k6/browser instead of k6/experimental/browser,"+
 				" which will be removed after September 23rd. Ensure your scripts are migrated by then."+
-				"For more information, see the migration guide at the link:"+
+				" For more information, see the migration guide at the link:"+
 				" https://grafana.com/docs/k6/latest/using-k6-browser/migrating-to-k6-v0-52/"),
 		"k6/browser":         browser.New(),
 		"k6/experimental/fs": fs.New(),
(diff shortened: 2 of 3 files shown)


cmd/tests/cmd_run_test.go
```diff
@@ -2300,12 +2300,12 @@ func TestBrowserPermissions(t *testing.T) {
 		t.Run(tt.name, func(t *testing.T) {
 			t.Parallel()
 			script := fmt.Sprintf(`
-			import { browser } from 'k6/experimental/browser';
+			import { browser } from 'k6/browser';
```
ankur22, line 2303: I think we should have a test for both imports until we remove the experimental import.

js/jsmodules.go
```diff
@@ -46,13 +46,16 @@ func getInternalJSModules() map[string]interface{} {
...
-		"k6/experimental/browser": browser.NewSync(),
-		"k6/experimental/fs":      fs.New(),
-		"k6/net/grpc":             grpc.New(),
-		"k6/html":                 html.New(),
-		"k6/http":                 http.New(),
-		"k6/metrics":              metrics.New(),
-		"k6/ws":                   ws.New(),
+		"k6/experimental/browser": newWarnExperimentalModule(browser.NewSync(),
+			"k6/experimental/browser is now part of the k6 core, please change your imports to use k6/browser instead."+
+				" The k6/experimental/browser will be removed in k6 v0.54.0"),
```
ankur22, line 51: Is this official? I don't think we've properly discussed when it will be removed. Could this be changed to something like "...The k6/experimental/browser is planned to be removed in k6 v0.54.0, but may be delayed"?

js/jsmodules.go
```diff
@@ -46,13 +46,16 @@ func getInternalJSModules() map[string]interface{} {
...
+		"k6/browser":         browser.New(),
```
ankur22, line 52: 🎉 🚀

ankur22 (dismissed):
LGTM 🙂

js/jsmodules.go
joanlopez: Isn't the migration from `k6/experimental/browser` to `k6/browser` breaking? 🤔 

I think we usually leave a message like that when we graduate a module, but in such cases there are no changes other than the import path, but in this case I don't think that just changing the import is enough, right?

inancgumus: @andrewslotin I'd like to hear your thoughts about this as well. Thanks!

inancgumus: @joanlopez Yes, it's breaking change. We want to give users some room to adjust their scripts. I'm open to your ideas.

joanlopez: I don't have any particular opinion because I don't know if that happened before and how that was handled in such case, if so. But I guess that a short message with a reference for context/instructions on how to migrate, at least, would be nice to have.

codebien: Yeah, I would second @joanlopez. You have to mention the fact they have to await or handle the returned promise. Potentially, you can directly link the documentation where you are explaining and guiding the to-do steps.

andrewslotin: I agree with @joanlopez on providing a link to a migration doc here. As for the version, I'd suggest that we rather use a date here, since this provides a clearer timeframe when users need to take action. For example this is what we used as a warning while deprecating the legacy k6 Docker image:

> **THIS IMAGE IS DEPRECATED and its support will be discontinued after Dec 31, 2023. This image has been replaced by [`grafana/k6`](https://hub.docker.com/r/grafana/k6).<br>
Please update your scripts to use [`grafana/k6`](https://hub.docker.com/r/grafana/k6) to continue using the latest version of k6.**

inancgumus: To give users enough time to migrate, we want to go with the date September 23rd (Cycle 23 k6 release date).

cmd/tests/cmd_run_test.go
inancgumus: This test is about k6 options and is not relevant to the module path.

In the commit, I changed the import path just to accommodate the module's current state. So, I don't think we should add another test here.

js/jsmodules.go
inancgumus: WDYT about this warning message?

> k6/experimental/browser is now part of the k6 core, please change your imports to use k6/browser instead. The k6/experimental/browser will be removed on or after 23rd of September, please ensure your scripts have been migrated to k6/browser by that date. Please also see the migration guide at https://grafana.com/docs/k6/latest/using-k6-browser/migrating-to-k6-v0-52/ for more information on how to migrate your scripts.

[8b59034](https://github.com/grafana/k6/pull/3793/commits/8b59034a30f2ea975880b738542b7faf17b23a83)

cmd/tests/cmd_run_test.go
ankur22: Ok, got it, so it's nothing to do with the import itself.

I still think it's worth duplicating this test and the browser.js test, just to be extra cautious and ensure that both imports still work with what we're testing here.

inancgumus: Duplicating them won't be worth it because we already tested the `browser` option. Instead, I've added a test that checks whether the `k6/experimental/browser` path is there and gives us a warning.

See: [48683b0](https://github.com/grafana/k6/pull/3793/commits/48683b09fdfcd60c5e2fe3cc2b04143e83993ffd)

ankur22 (approved):

js/jsmodules.go
joanlopez: Much more clear @inancgumus, looks good to me! 👍🏻

joanlopez (approved):

js/jsmodules.go
```diff
@@ -46,13 +46,18 @@ func getInternalJSModules() map[string]interface{} {
...
-		"k6/experimental/browser": browser.NewSync(),
-		"k6/experimental/fs":      fs.New(),
-		"k6/net/grpc":             grpc.New(),
-		"k6/html":                 html.New(),
-		"k6/http":                 http.New(),
-		"k6/metrics":              metrics.New(),
-		"k6/ws":                   ws.New(),
+		"k6/experimental/browser": newWarnExperimentalModule(browser.NewSync(),
+			"Please update your imports to use k6/browser instead of k6/experimental/browser,"+
+				" which will be removed after September 23rd. Ensure your scripts are migrated by then."+
```
olegbespalov, line 51: I'd suggest to be explicit about that date

```suggestion
				" which will be removed after September 23rd, 2024. Ensure your scripts are migrated by then."+
```

olegbespalov (approved):

joanlopez (approved):

PR #3800, opened 2024-06-18
 cmd/archive: additionally test for non-existence of file named `-`

roobre:
## What?

A follow-up to https://github.com/grafana/k6/pull/3797, which I didn't want to push over as it already has approvals.

## Why?

Additionally to check for non-existence of `archive.tar`, also check for non-existence of a file named `-`, which could happen if the special case of `-` = stdout is mishandled.

## Checklist

## Related PR(s)/Issue(s)

https://github.com/grafana/k6/pull/3797

### DIFF
--- a/cmd/archive_test.go
+++ b/cmd/archive_test.go
@@ -268,8 +268,10 @@ func TestArchiveStdout(t *testing.T) {
 
 	newRootCommand(ts.GlobalState).execute()
 
-	_, err := ts.FS.Stat("archive.tar")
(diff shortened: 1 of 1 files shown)


codebien (changes requested):
As `archive.tar` is the default value, it makes sense to check that it doesn't exist. Maybe, we can check both, as they are the obvious catch in case of bugs.

codebien (approved):

joanlopez (approved):

codebien (approved):

PR #3807, opened 2024-06-24
Warn on mixing ESM and commonJS

mstoykov:
## What?

As part of testing for #3265 it was noticed this isn't uncommon.

## Why?

As there is no way for this to be supported and it will be against how the specification works we prefer to warn users on this at least for a little while.

## Checklist

### DIFF
--- a/js/compiler/compiler.go
+++ b/js/compiler/compiler.go
@@ -251,7 +251,15 @@ func (c *Compiler) compileImpl(
 			return nil, code, err
 		}
 		// the compatibility mode "decreases" here as we shouldn't transform twice
-		return c.compileImpl(code, filename, wrap, lib.CompatibilityModeBase, state.srcMap)
+		var prg *sobek.Program
+		prg, code, err = c.compileImpl(code, filename, wrap, lib.CompatibilityModeBase, state.srcMap)
+		if err == nil && strings.Contains(src, "module.exports") {
+			c.logger.Warningf(
+				"While compiling %q it was noticed that it mixes import/export syntax (ESM) and commonJS module.exports. "+
+					"This isn't standard behaviour and will soon not work. Please use one or the other.",
+				filename)
+		}
+		return prg, code, err
 	}
 
 	if compatibilityMode == lib.CompatibilityModeExperimentalEnhanced {
--- a/js/compiler/compiler.go
+++ b/js/compiler/compiler.go
@@ -255,8 +255,10 @@ func (c *Compiler) compileImpl(
 		prg, code, err = c.compileImpl(code, filename, wrap, lib.CompatibilityModeBase, state.srcMap)
 		if err == nil && strings.Contains(src, "module.exports") {
 			c.logger.Warningf(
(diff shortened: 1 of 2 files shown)


js/compiler/compiler.go
```diff
@@ -251,7 +251,15 @@ func (c *Compiler) compileImpl(
 			return nil, code, err
 		}
 		// the compatibility mode "decreases" here as we shouldn't transform twice
-		return c.compileImpl(code, filename, wrap, lib.CompatibilityModeBase, state.srcMap)
+		var prg *sobek.Program
+		prg, code, err = c.compileImpl(code, filename, wrap, lib.CompatibilityModeBase, state.srcMap)
+		if err == nil && strings.Contains(src, "module.exports") {
+			c.logger.Warningf(
+				"While compiling %q it was noticed that it mixes import/export syntax (ESM) and commonJS module.exports. "+
+					"This isn't standard behaviour and will soon not work. Please use one or the other.",
```
oleiade, line 259: Some minor wording nit 🙇🏻 

```suggestion
				"During the compilation of %q, it has been detected that the file combines ECMAScript modules (ESM)  import/export syntax with commonJS module.exports. "+
					"Mixing these two module systems is non-standard and will not be supported anymore in future releases. Please ensure to solely one or the other syntax.",
```

oleiade (approved):

js/compiler/compiler.go
```diff
@@ -251,7 +251,17 @@ func (c *Compiler) compileImpl(
 			return nil, code, err
 		}
 		// the compatibility mode "decreases" here as we shouldn't transform twice
-		return c.compileImpl(code, filename, wrap, lib.CompatibilityModeBase, state.srcMap)
+		var prg *sobek.Program
+		prg, code, err = c.compileImpl(code, filename, wrap, lib.CompatibilityModeBase, state.srcMap)
+		if err == nil && strings.Contains(src, "module.exports") {
+			c.logger.Warningf(
+				"During the compilation of %q, it has been detected that the file combines ECMAScript modules (ESM) "+
```
codebien, line 258:
```suggestion
				"During the compilation of %q, it has been detected that the file %q combines ECMAScript modules (ESM) "+
```
Could help if we write the filename there?

js/compiler/compiler.go
```diff
@@ -251,7 +251,17 @@ func (c *Compiler) compileImpl(
...
+					"import/export syntax with commonJS module.exports. "+
+					"Mixing these two module systems is non-standard and will not be supported anymore in future releases. "+
+					"Please ensure to solely one or the other syntax.",
```
codebien, line 261:
```suggestion
					"Please ensure to use solely one or the other syntax.",
```
It sounds a verb is missing here

js/compiler/compiler.go
mstoykov: It is already mentioned just before the comma, so I don't see why it should be repeated. Especially if the path is long it will make the message huge. And will probably be confusing that we repeat it.

codebien (approved):

joanlopez (approved):

PR #3813, opened 2024-06-25
Move cloud run and login functionalities under cloud subcommands

oleiade:
## What?

This PR modifies the k6 command-line interface and introduces the cloud run and login functionalities under specific subcommands: `k6 cloud run` and `k6 cloud login`.

These two new subcommands supersede the existing `k6 cloud` and `k6 login cloud` functionalities, but as planned, they will remain in place until we reach k6 version 1.0.0. However, they now display a deprecation warning inviting users to move to the new commands.

Future plans include to eventually, on k6 v1.0.0 release, we expect to modify the `k6 cloud` command behavior so that it doesn't run scripts in the cloud anymore but only displays the help text documenting its subcommands. 

### Technicalities

From a technical perspective, this PR "only" duplicates the existing logic under new subcommands and registers those. Some outstanding refactors are mentioned throughout their code, but we adopted the approach of delegating those refactors and improvements to a future PR.

## Why?

This PR is part of the effort to improve the UX of our command-line interface towards k6 v1.0.0. It aligns with our goals to unify and expand the k6 CLI's capabilities to interact with the cloud. In my absence for UX-specific concerns and discussions, please get in touch with @dgzlopes, who have as much context as we do.

## TaskList

- [ ] As requested by @dgzlopes, drop `k6 cloud login` support for email and password login: this includes the interactive cli triggered when running `k6 cloud login`, it should instead request your API token. 
- [x] As we're at it, we might as well remove the `k6 cloud run --upload-only` flag, and replace it with `k6 cloud upload`. If this is a bigger task that I anticipate, we should do it in a separate PR.
- [ ] Add more tests for the `k6 cloud run` command (asserting the arguments management especially)
- [ ] Add more tests for the `k6 cloud login` command: I wonder what we can improve on that front, but it would be worth looking into it.

## Checklist


cmd/cloud.go
```diff
@@ -343,20 +344,69 @@ func getCmdCloud(gs *state.GlobalState) *cobra.Command {
...
+
 	cloudCmd.Flags().SortFlags = false
 	cloudCmd.Flags().AddFlagSet(c.flagSet())
+
 	return cloudCmd
 }
+
+func exactCloudArgs() cobra.PositionalArgs {
+	return func(_ *cobra.Command, args []string) error {
+		if len(args) < 1 || len(args) > 2 {
```
joanlopez, line 391: In such case, now the message is something like:

```
ERRO[0000] accepts 1 arg(s), received 0: arg should either be "-", if reading script from stdin, or a path to a script file 
```

So, I'm wondering, should we try to be somehow more explicit? Or just delegate to `-h`?

cmd/cloud.go
```diff
@@ -343,20 +344,69 @@ func getCmdCloud(gs *state.GlobalState) *cobra.Command {
...
 		Short: "Run a test on the cloud",
-		Long: `Run a test on the cloud.
+		Long: `[deprecation notice]
+The k6 team is in the process of modifying and deprecating the cloud command behavior. In the future, the "cloud"
+command will only display a help text, instead of running tests in the cloud.
+To run tests in the cloud, users are now invited to migrate to the "k6 cloud run" command instead.
+
+Run a test on the cloud.
 
 This will execute the test on the k6 cloud service. Use "k6 login cloud" to authenticate.`,
```
joanlopez, line 372:
```suggestion
This will execute the test on the k6 cloud service. Use "k6 cloud login" to authenticate.`,
```

joanlopez: At this point, I think we can start suggesting the new commands in help messages, right?

cmd/cloud.go
```diff
@@ -343,20 +344,69 @@ func getCmdCloud(gs *state.GlobalState) *cobra.Command {
...
+	return func(_ *cobra.Command, args []string) error {
+		if len(args) < 1 || len(args) > 2 {
+			return fmt.Errorf("accepts 1 or 2 arg(s), received %d", len(args))
+		}
+
+		var (
+			isRunSubcommand   = args[0] == "run"
+			isLoginSubcommand = args[0] == "login"
+			isScript          = filepath.Ext(args[0]) == ".js"
+			isArchive         = filepath.Ext(args[0]) == ".tar"
```
joanlopez, line 399: Do we really need such differentiation? 
Even further, do we really want this limitation (up to `.js` and `.tar`)? 

For instance, what if the user wants to run a test with the new `experimental_enhanced` compatibility mode, and submit a TS file?

cmd/cloud.go
```diff
@@ -343,20 +344,69 @@ func getCmdCloud(gs *state.GlobalState) *cobra.Command {
...
+		)
+
+		if len(args) == 1 && !isScript && !isArchive {
+			return fmt.Errorf("unexpected argument: %s", args[0])
+		}
+
+		if len(args) == 2 && !isRunSubcommand && !isLoginSubcommand {
+			return fmt.Errorf("unexpected argument: %s", args[0])
```
joanlopez, line 407: Close to https://github.com/grafana/k6/pull/3813/files#r1668585673

cmd/cloud_run.go
```diff
@@ -0,0 +1,376 @@
...
+		exitOnRunning: false,
+		uploadOnly:    false,
+	}
+
+	exampleText := getExampleText(gs, `
+  # Run a test in the Grafana k6 cloud
+  $ {{.}} cloud run script.js
+
+  # Run a test in the Grafana k6 cloud with a specific token
+  $ {{.}} cloud run  --token <YOUR_API_TOKEN> script.js`[1:])
```
joanlopez, line 49:
```suggestion
  $ {{.}} cloud run --token <YOUR_API_TOKEN> script.js`[1:])
```

cmd/cloud_login.go
```diff
@@ -0,0 +1,175 @@
...
+
+  # Display the stored token.
+  {{.}} cloud login -s`[1:])
+
+	loginCloudCommand := &cobra.Command{
+		Use:   cloudLoginCommandName,
+		Short: "Authenticate with Grafana k6 Cloud",
+		Long: `Authenticate with Grafana Cloud k6.
+
+This command will authenticate you with Grafana Cloud k6. Once authenticated,
```
joanlopez, line 47:
```suggestion
This command will authenticate you with Grafana Cloud k6.
```
Duplicated with the line below

cmd/cloud_run.go
```diff
@@ -0,0 +1,376 @@
...
+	"go.k6.io/k6/cloudapi"
+	"go.k6.io/k6/cmd/state"
+	"go.k6.io/k6/errext"
+	"go.k6.io/k6/errext/exitcodes"
+	"go.k6.io/k6/lib"
+	"go.k6.io/k6/lib/consts"
+	"go.k6.io/k6/ui/pb"
+)
+
+// cmdCloudRun handles the `k6 cloud` sub-command
```
joanlopez, line 27:
```suggestion
// cmdCloudRun handles the `k6 cloud run` sub-command
```

cmd/cloud.go
joanlopez: Also, not sure if 2 args is technically correct, or it should be one command plus one arg, if so.

cmd/cloud_run.go
```diff
@@ -0,0 +1,376 @@
...
+	}
+
+	exampleText := getExampleText(gs, `
+  # Run a test in the Grafana k6 cloud
+  $ {{.}} cloud run script.js
+
+  # Run a test in the Grafana k6 cloud with a specific token
+  $ {{.}} cloud run  --token <YOUR_API_TOKEN> script.js`[1:])
+
+	// FIXME: when the command is "k6 cloud run" without an script/archive, we should display an error and the help
```
joanlopez, line 51: Does this mean that the call to `exactArgsWithMsg` on line 60 isn't enough?
Or can we just remove that line?

cmd/cloud_login.go
```diff
@@ -0,0 +1,175 @@
...
+  # Store a token in k6's persistent configuration.
+  {{.}} cloud login -t <YOUR_TOKEN>
+
+  # Display the stored token.
+  {{.}} cloud login -s`[1:])
+
+	loginCloudCommand := &cobra.Command{
+		Use:   cloudLoginCommandName,
+		Short: "Authenticate with Grafana k6 Cloud",
+		Long: `Authenticate with Grafana Cloud k6.
```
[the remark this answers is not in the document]
dgzlopes: Yeah, let's use Grafana Cloud k6.

cmd/cloud_run.go
joanlopez: I don't think the `--token` flag is supported for now, do we want to add support for it? 🤔

cmd/cloud.go
oleiade: What about inverting the logic, and instead of looking for script/archive directly, based our check on whether the argument is **not** a subcommand.

cmd/cloud.go
oleiade: I always found it a bit counterintuitive that `k6 run` and `k6 cloud` return an error when missing an argument, as opposed to displaying "usage help text" instead for instance.

However, I also notice that so far we don't seem to have explicitly documented the use of the `-` argument, besides that specific error message, to allow for piping the script/archive as content either. 

So I would be keen to modify the help text to mention the argument structure, and support for '-' explicitly, and show the help text indeed in that case 👍🏻

cmd/cloud_run.go
oleiade: No, I think it's definetly a mistak on my side, I did both cloud run and cloud login at the same time and I must have updated `cloud login` without realizing I did 👍🏻 

Dropping it 👍🏻

cmd/cloud.go
oleiade: I've excluded this check 👍🏻

cmd/cloud.go
oleiade: For now I've simplified the logic, and made it so that `k6 cloud` (without argument) displays

```
ERRO[0000] the k6 cloud command accepts 1 argument consisting in either in a subcommand such as `run` or `cloud`, or the path to a script/archive, or the `-` symbol, received: 0 arguments instead
```

And so that `k6 cloud run` (without script argument) displays

```
 accepts 1 arg(s), received 0: the k6 cloud run command expects a single argument consisting in either a path to a script or archive file, or the "-" symbol indicating the script or archive should be read from stdin
 ```
 
 Let me know if that's good enough, I'll resolve in the meantime, but feel free to unresolve 🙇‍♂️

File: cmd/cloud_run.go (hunk 8 above)
Comment (oleiade) [reply cmd/cloud_run.go]: Dropped 👍🏻

**Comment by oleiade** [comment]
Cleaning up the PR history in preparation for leaving draft state 👀

File: cmd/cloud.go [hunk 10]
```diff
@@ -343,20 +344,57 @@ func getCmdCloud(gs *state.GlobalState) *cobra.Command {
...
-		Long: `Run a test on the cloud.
+		Long: `[deprecation notice]
+The k6 team is in the process of modifying and deprecating the cloud command behavior. In the future, the "cloud"
+command will only display a help text, instead of running tests in the cloud.
+To run tests in the cloud, users are now invited to migrate to the "k6 cloud run" command instead.
 
-This will execute the test on the k6 cloud service. Use "k6 login cloud" to authenticate.`,
-		Example: exampleText,
-		Args:    exactArgsWithMsg(1, "arg should either be \"-\", if reading script from stdin, or a path to a script file"),
+Run a test on the cloud.
```
joanlopez, line 370: Nit;

```suggestion
Run a test in the cloud.
```

cmd/cloud.go
```diff
@@ -343,20 +344,57 @@ func getCmdCloud(gs *state.GlobalState) *cobra.Command {
...
+
+This will execute the test on the k6 cloud service. Use "k6 cloud login" to authenticate.`,
```
joanlopez, line 372:
```suggestion
This will execute the test in the Grafana Cloud k6 service. Use "k6 cloud login" to authenticate.`,
```

cmd/cloud.go
```diff
@@ -343,20 +344,57 @@ func getCmdCloud(gs *state.GlobalState) *cobra.Command {
...
 	return cloudCmd
 }
+
+func exactCloudArgs() cobra.PositionalArgs {
+	return func(_ *cobra.Command, args []string) error {
+		if len(args) == 0 {
+			return fmt.Errorf(
+				"the k6 cloud command accepts 1 argument consisting in either in "+
+					"a subcommand such as `run` or `cloud`, or the path to a script/archive, or "+
+					"the `-` symbol, received: %d arguments instead", len(args))
```
joanlopez, line 395:
```suggestion
				"the k6 cloud command accepts 1 argument consisting in either "+
					"a subcommand such as `run` or `cloud`, or the path to a script/archive, or "+
					"the `-` symbol, received: %d arguments instead", len(args))
```

joanlopez: What if the amount of arguments is larger than expected? 🤔

cmd/cloud_login.go
```diff
@@ -0,0 +1,175 @@
...
+  # Display the stored token.
+  {{.}} cloud login -s`[1:])
+
+	loginCloudCommand := &cobra.Command{
+		Use:   cloudLoginCommandName,
+		Short: "Authenticate with Grafana Cloud k6",
+		Long: `Authenticate with Grafana Cloud k6.
+
+This command will authenticate you with Grafana Cloud k6.
+Once authenticated you can start running tests in the cloud by using the "k6 cloud"
```
joanlopez, line 48:
```suggestion
Once authenticated you can start running tests in the cloud by using the "k6 cloud run"
```

cmd/cloud_login.go
```diff
@@ -0,0 +1,175 @@
...
+	exampleText := getExampleText(gs, `
+  # Log in with an email/password.
+  {{.}} cloud login
+
+  # Store a token in k6's persistent configuration.
+  {{.}} cloud login -t <YOUR_TOKEN>
+
+  # Display the stored token.
+  {{.}} cloud login -s`[1:])
+
```
joanlopez, line 41:
```suggestion
  # Display the stored token.
  {{.}} cloud login -s
  
  # Reset the stored token.
  {{.}} cloud login -r`[1:])

```

cmd/cloud.go
```diff
@@ -343,20 +344,57 @@ func getCmdCloud(gs *state.GlobalState) *cobra.Command {
...
+  $ {{.}} cloud script.js
+
+  # [deprecated] Run a k6 archive in the Grafana Cloud k6
+  $ {{.}} cloud archive.tar`[1:])
 
 	cloudCmd := &cobra.Command{
 		Use:   "cloud",
 		Short: "Run a test on the cloud",
-		Long: `Run a test on the cloud.
+		Long: `[deprecation notice]
```
olegbespalov, line 365: Here in for other subcommands. There is a `Deprecated` field of the `cobra.Command`, which probably could serve better for the marking command as deprecated.

cmd/cloud_login.go
```diff
@@ -0,0 +1,175 @@
...
+the "k6 run -o cloud" command.
+`,
+		Example: exampleText,
+		Args:    cobra.NoArgs,
+		RunE:    c.run,
+	}
+
+	loginCloudCommand.Flags().StringP("token", "t", "", "specify `token` to use")
+	loginCloudCommand.Flags().BoolP("show", "s", false, "display saved token and exit")
+	loginCloudCommand.Flags().BoolP("reset", "r", false, "reset token")
```
olegbespalov, line 59:
```suggestion
	loginCloudCommand.Flags().BoolP("reset", "r", false, "reset stored token")
```

cmd/tests/cmd_cloud_run_test.go
```diff
@@ -0,0 +1,248 @@
...
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"go.k6.io/k6/cloudapi"
+	"go.k6.io/k6/cmd"
+	"go.k6.io/k6/lib/fsext"
+	"go.k6.io/k6/lib/testutils"
+)
+
+func TestCloudRunNotLoggedIn(t *testing.T) {
```
olegbespalov, line 20: The test cases below are copies of the ones that `cmd_cloud_test.go` with the running argument difference. What if we turn them into the constructors for the period of deprecation and just run the old/new commands accordingly? 

There are a couple of rationales behind that suggestion. The most obvious is avoiding duplication, keeping the git history more conventional, and consolidating the test case.

The alternative to that is the wish to drop the comment at least near the old tests with referring that at some point it will be dropped in favor of the new one, which should help for future maintenance of these parts of the code

cmd/cloud.go
```diff
@@ -33,6 +33,7 @@ type cmdCloud struct {
 	uploadOnly    bool
 }
 
+//nolint:dupl // remove this statement once the migration from the `k6 cloud` to the `k6 cloud run` is complete.
```
codebien, line 36: The _why_ is missing. I don't get why we have this nolint rule, it requires me to check the code in detail.

cmd/cloud.go
```diff
@@ -343,20 +344,57 @@ func getCmdCloud(gs *state.GlobalState) *cobra.Command {
...
+
+  # [deprecated] Run a k6 archive in the Grafana Cloud k6
+  $ {{.}} cloud archive.tar`[1:])
 
 	cloudCmd := &cobra.Command{
 		Use:   "cloud",
 		Short: "Run a test on the cloud",
-		Long: `Run a test on the cloud.
+		Long: `[deprecation notice]
+The k6 team is in the process of modifying and deprecating the cloud command behavior. In the future, the "cloud"
```
codebien, line 366:
```suggestion
The k6 team is in the process of modifying and deprecating the `k6 cloud` command behavior. In the future, the "cloud"
```

cmd/cloud.go
codebien, line 370:
```suggestion
Run a k6 script in the Grafana Cloud k6.
```
Should it be consistent with the text used in the help?

cmd/cloud.go
codebien, line 372:
```suggestion
This will execute the test on the Grafana Cloud k6 service. Use before "k6 cloud login" to authenticate.`,
```

cmd/cloud_login.go
```diff
@@ -0,0 +1,175 @@
...
+	// loginCloudCommand represents the 'cloud login' command
+	exampleText := getExampleText(gs, `
+  # Log in with an email/password.
+  {{.}} cloud login
+
+  # Store a token in k6's persistent configuration.
+  {{.}} cloud login -t <YOUR_TOKEN>
+
+  # Display the stored token.
+  {{.}} cloud login -s`[1:])
```
codebien, line 40:
```suggestion
  {{.}} cloud login -s`[1:])
```
Can we store it partially (masked) by default? Something like `AF&s*****kS8`, and add one more flag to show it fully?

cmd/cloud_run.go
```diff
@@ -0,0 +1,381 @@
...
+		RunE:    c.run,
+	}
+
+	cloudRunCmd.Flags().SortFlags = false
+	cloudRunCmd.Flags().AddFlagSet(c.flagSet())
+
+	return cloudRunCmd
+}
+
+//nolint:dupl // remove this statement once the migration from the `k6 cloud` to the `k6 cloud run` is complete.
```
codebien, line 76: Same, why do we have this rule?

cmd/login_cloud.go
```diff
@@ -33,7 +33,11 @@ func getCmdLoginCloud(gs *state.GlobalState) *cobra.Command {
 	loginCloudCommand := &cobra.Command{
 		Use:   "cloud",
 		Short: "Authenticate with k6 Cloud",
-		Long: `Authenticate with k6 Cloud",
+		Long: `[deprecation notice]
```
codebien, line 36: Same as Oleg's comment, there is a Deprecated field for it.

cmd/tests/cmd_cloud_run_test.go
codebien: Yeah, I agree, I suggested the same about the command itself, if possible we should wrap and do toggles as much as we can. Otherwise, there is also the risk of having a bug somewhere and fixing only one part of the code.

codebien (commented):
If I get the code correctly, there aren't direct dependencies between the `cloud run` and `cloud login` commands. So, instead of having them in the same pull request can we split them? They are independent, so we can split reviews and consequently mergeability and any eventual flow from them (a revent for example).

cmd/cloud.go
oleiade: Oh I didn't know that! Thanks for the pointer, will switch to that indeed 🙇🏻

cmd/tests/cmd_cloud_run_test.go
oleiade: Agree with you folks, let's do it 🙇🏻

cmd/cloud_run.go
oleiade: I've modified the messages to make it clearer that it's currently duplicated code. Expect this to go away as we revamp the implementation to avoid duplication.

cmd/cloud_login.go
```diff
@@ -0,0 +1,178 @@
...
+  {{.}} cloud login
+
+  # Store a token in k6's persistent configuration.
+  {{.}} cloud login -t <YOUR_TOKEN>
+
+  # Display the stored token.
+  {{.}} cloud login -s
+  
+  # Reset the stored token.
+  {{.}} cloud login -r`[1:])
```
joanlopez, line 43: Nit; but I think we don't use dots at the end of the example titles for the rest of the commands.

```suggestion
  # Log in with an email/password
  {{.}} cloud login

  # Store a token in k6's persistent configuration
  {{.}} cloud login -t <YOUR_TOKEN>

  # Display the stored token
  {{.}} cloud login -s
  
  # Reset the stored token
  {{.}} cloud login -r`[1:])
```

cmd/login.go
```diff
@@ -11,11 +11,17 @@ func getCmdLogin(gs *state.GlobalState) *cobra.Command {
 	loginCmd := &cobra.Command{
 		Use:   "login",
 		Short: "Authenticate with a service",
-		Long: `Authenticate with a service.
+		Long: `[deprecation notice]
```
joanlopez, line 14: As you're already using the `Deprecated` field, I think this change isn't longer needed. Otherwise, it looks a bit redundant:

```
❯ go run main.go login -h
Command "login" is deprecated, This command is deprecated and will be removed in a future release.
Please use the "k6 cloud login" command instead.

[deprecation notice]
This command is deprecated and will be removed in a future release. Please use the "k6 cloud login" command instead.

Authenticate with a service.
```

cmd/login.go
```diff
@@ -11,11 +11,17 @@ func getCmdLogin(gs *state.GlobalState) *cobra.Command {
...
+This command is deprecated and will be removed in a future release. Please use the "k6 cloud login" command instead.
+
+Authenticate with a service.
 
 Logging into a service changes the default when just "-o [type]" is passed with
 no parameters, you can always override the stored credentials by passing some
 on the commandline.`,
+		Deprecated: `This command is deprecated and will be removed in a future release.
+Please use the "k6 cloud login" command instead.
+`,
```
joanlopez, line 24: By default, Cobra already prepends: `Command "login" is deprecated, `, see:

```
❯ go run main.go login -h
Command "login" is deprecated, This command is deprecated and will be removed in a future release.
Please use the "k6 cloud login" command instead.
```

so I'd suggest:

```suggestion
		Deprecated: `and will be removed in a future release.
Please use the "k6 cloud login" command instead.
`,
```

cmd/login_cloud.go
```diff
@@ -33,11 +33,14 @@ func getCmdLoginCloud(gs *state.GlobalState) *cobra.Command {
...
 		Short: "Authenticate with k6 Cloud",
-		Long: `Authenticate with k6 Cloud",
+		Long: `Authenticate with Grafana Cloud k6.
 
 This will set the default token used when just "k6 run -o cloud" is passed.`,
 		Example: exampleText,
 		Args:    cobra.NoArgs,
+		Deprecated: `This command is deprecated and will be removed in a future release.
+Please use the "k6 cloud login" command instead.
+`,
```
joanlopez, line 43: Same reasoning as above:

```suggestion
		Deprecated: `and will be removed in a future release.
Please use the "k6 cloud login" command instead.
`,
```

cmd/cloud.go
```diff
@@ -343,20 +345,67 @@ func getCmdCloud(gs *state.GlobalState) *cobra.Command {
...
+
 	cloudCmd.Flags().SortFlags = false
 	cloudCmd.Flags().AddFlagSet(c.flagSet())
+
 	return cloudCmd
 }
+
+func exactCloudArgs() cobra.PositionalArgs {
+	return func(_ *cobra.Command, args []string) error {
+		const baseErrMsg = "the k6 cloud command expects either a subcommand such as `run` or `login`, or " +
```
joanlopez, line 393:
```suggestion
		const baseErrMsg = `the "k6 cloud" command expects either a subcommand such as "run" or "login", or ` +
```

cmd/cloud.go
```diff
@@ -343,20 +345,67 @@ func getCmdCloud(gs *state.GlobalState) *cobra.Command {
...
+			"a single argument consisting in a path to a script/archive, or the `-` symbol instructing " +
+			"the command to read the test content from stdin"
+
+		if len(args) == 0 {
+			return fmt.Errorf(baseErrMsg + "; " + "received no arguments")
+		}
+
+		hasSubcommand := len(args) >= 1 && (args[0] == "run" || args[0] == "login")
+		if !hasSubcommand {
```
joanlopez, line 402: As it is now, when you provide a single argument (legacy behavior), it displays the error, while it shouldn't, I think:

```suggestion
		hasSubcommand := args[0] == "run" || args[0] == "login"
		if len(args) > 1 && !hasSubcommand {
```

cmd/cloud.go
```diff
@@ -343,20 +345,67 @@ func getCmdCloud(gs *state.GlobalState) *cobra.Command {
...
+			return fmt.Errorf(
+				baseErrMsg+"; "+"received %d arguments %q, and none of them is a subcommand",
+				len(args), strings.Join(args, " "),
```
joanlopez, line 405: I think the way below is a bit more accurate, as we only care about the first argument (to be a valid subcommand):

```suggestion
				baseErrMsg+"; "+"received %d arguments %q, and %s is not a valid subcommand",
				len(args), strings.Join(args, " "), args[0],
```

cmd/cloud.go
joanlopez: Indeed, I think that's the reason why tests are failing: CI is 🔴

cmd/login.go
oleiade: Forgot this one behind 👍

cmd/cloud_login.go
joanlopez: I'd leave such things for a separate pull request, if so. Not sure tho, how critical/sensitive it really is.

oleiade: Agreed 👍🏻

joanlopez (approved):
Great job! 🚀

oleiade:
I have squashed the PR suggestion commits.

I've also applied a pretty significant refactor to avoid code duplication: the cloud run command relies on the cloud command's code directly, and I've refactored the tests to share the same test code between both the `k6 cloud` and `k6 cloud run` commands and parameterize the tests. Please check it out 2b0bc1f7b17712a77e0951dc41094d001dcf51af and let me know what you think @joanlopez and @olegbespalov 🙇🏻 

If we're happy with this approach, I'll also apply it the `k6 cloud login` command 👍🏻

cmd/cloud.go
```diff
@@ -117,7 +119,10 @@ func (c *cmdCloud) run(cmd *cobra.Command, args []string) error {
 		return err
 	}
 	if !cloudConfig.Token.Valid {
-		return errors.New("Not logged in, please use `k6 login cloud`.") //nolint:golint,revive,stylecheck
+		return errors.New( //nolint:golint
+			"not logged in, please login to the Grafana Cloud k6 " +
```
olegbespalov, line 123:
```suggestion
			"not logged in, please login first to the Grafana Cloud k6 " +
```

cmd/cloud.go
```diff
@@ -343,20 +348,67 @@ func getCmdCloud(gs *state.GlobalState) *cobra.Command {
...
+  $ {{.}} cloud run script.js
+
+  # Run a k6 archive in the Grafana Cloud k6
+  $ {{.}} cloud run archive.tar`[1:])
 
 	cloudCmd := &cobra.Command{
 		Use:   "cloud",
 		Short: "Run a test on the cloud",
-		Long: `Run a test on the cloud.
+		Long: `Run a test archive in the Grafana Cloud k6.
```
olegbespalov, line 369: Archive details details seem not so important here

```suggestion
		Long: `Run a test in the Grafana Cloud k6.
```

cmd/cloud.go
```diff
@@ -343,20 +348,67 @@ func getCmdCloud(gs *state.GlobalState) *cobra.Command {
...
 
-This will execute the test on the k6 cloud service. Use "k6 login cloud" to authenticate.`,
-		Example: exampleText,
-		Args:    exactArgsWithMsg(1, "arg should either be \"-\", if reading script from stdin, or a path to a script file"),
+This will execute the test in the Grafana Cloud k6 service. Be sure to run the "k6 cloud login" command prior to
```
olegbespalov, line 371: How about expanding a bit to give more explanations of what is happening? Feel free to ignore suggestion

```suggestion
This will archive test script(s), including all necessary resources, and execute the test in the Grafana Cloud k6 service. Be sure to run the "k6 cloud login" command prior to
```

cmd/cloud_run.go
```diff
@@ -0,0 +1,48 @@
...
+
+  # Read a test script or archive from stdin and run it in Grafana Cloud k6
+  $ {{.}} cloud run - < script.js`[1:])
+
+	cloudRunCmd := &cobra.Command{
+		Use:   cloudRunCommandName,
+		Short: "Run a test in Grafana Cloud k6",
+		Long: `Run a test in Grafana Cloud k6.
+
+This will execute the test in the Grafana Cloud k6 service. Using this command requires to be authenticated
```
olegbespalov, line 33: Same as previous comment

```suggestion
This will archive test script(s), including all necessary resources, and execute the test in the Grafana Cloud k6 service. Using this command requires to be authenticated
```

cmd/cloud_login.go
```diff
@@ -0,0 +1,178 @@
...
+	switch {
+	case reset.Valid:
+		newCloudConf.Token = null.StringFromPtr(nil)
+		printToStdout(c.globalState, "  token reset\n")
+	case show.Bool:
+	case token.Valid:
+		newCloudConf.Token = token
+	default:
+		form := ui.Form{
+			Fields: []ui.Field{
```
olegbespalov, line 113: Let's add some banner to the form to make UI a bit friendly

```suggestion
		form := ui.Form{
			Banner: "Please enter your Grafana Cloud k6 credentials",
			Fields: []ui.Field{
```

olegbespalov (approved):
Logs good to me, left a few minor non-blocking suggestions :+1:

olegbespalov (approved):

joanlopez (approved):

PR #3818, opened 2024-06-27
WIP Add a `k6 cloud run --local-execution` mode of execution

oleiade:
## What?

This PR adds support for a new `--local-execution` flag to the `k6 cloud run` command. Using it, users can run their test script or archive on their local machine, while streaming the results to the cloud.

It is built upon the latter, and ⚠️ cannot be merged before #3813 ⚠️.

In the context of this PR, the goal is to reproduce the behavior of the `k6 run -o cloud` behavior, and depending on the advancement of internal work, to also add support for adding archive upload. In a future PR, at a later point in time, we will also add support for logs and traces streaming to the cloud but are dependent on internal work to be able to do so.

⚠️ This is currently a work-in-progress, and non-functional (yet) PR ⚠️ 

## Why?

We aim to integrate the experience of running tests locally, whilst streaming results to the cloud more intuitive, and integrated in the k6 experience. We also aim to take this opportunity to align some of the logic and behavior involved in running tests in the cloud and running tests locally better.

## Task list

- [ ] We're gonna introduce a couple of options specific to `--local-execution`. One of the ideas we've been having in the k6 v1 UX work stream revolved around the idea of scoping options with a dot separator like in Prometheus: `--local.execution`, `--local.send-archive`, `--local.send-logs`, etc. cc @dgzlopes 
- [ ] Once we're at that point we'd like sending the archive to be **opt-in** (via an option that users need to explicitly pass). Same goes for logs. The reason being that they might contain sensitive info. cc @dgzlopes 
- [ ] When using `k6 cloud run --local-execution`, we should make sure the `outputs` field of the terminal output of the command displays a sensible value (like the URL of the job in the grafana cloud app?). cc @dgzlopes 
- [ ] Should the `k6 cloud run --local-execution` show the same terminal output as `k6 run` or `k6 cloud run`?

## Checklist

### DIFF
--- a/cmd/cloud_run.go
+++ b/cmd/cloud_run.go
@@ -8,8 +8,21 @@ import (
 	"os"
 	"path/filepath"
 	"strconv"
+	"strings"
 	"sync"
 	"time"
+
+	"gopkg.in/guregu/null.v3"
+
+	"github.com/sirupsen/logrus"
+	"go.k6.io/k6/event"
+	"go.k6.io/k6/execution"
+	"go.k6.io/k6/execution/local"
+	"go.k6.io/k6/js/common"
+	"go.k6.io/k6/lib/trace"
+	"go.k6.io/k6/metrics"
+	"go.k6.io/k6/metrics/engine"
+	"go.k6.io/k6/output"
 
 	"github.com/fatih/color"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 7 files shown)


joanlopez:
> One of the ideas we've been having in the k6 v1 UX work stream revolved around the idea of scoping options with a dot separator like in Prometheus: --local.execution, --local.send-archive, --local.send-logs, etc

I don't how sensible that would be, but as far as I know, using the dot as separator is not very Windows friendly (it requires quoting, see https://github.com/PowerShell/PowerShell/issues/19640).

cmd/cloud_run.go
```diff
@@ -68,6 +93,33 @@ against the Grafana k6 cloud. Use the "k6 cloud login" command to authenticate.`
...
+	// for local-execution.
+	flags.AddFlagSet(optionFlagSet())
+
+	// FIXME: this is brought directly from the `k6 run` command, and we might want to use a limited set of
+	// those flags here?
+	flags.AddFlagSet(runtimeOptionFlagSet(false))
+
+	// TODO: Figure out a better way to handle the CLI flags
+	flags.BoolVar(&c.exitOnRunning, "exit-on-running", c.exitOnRunning,
+		"exits when test reaches the running status")
```
joanlopez, line 110: I think this one can be simply removed, as I think it doesn't make sense for `--local-execution`.

cmd/cloud_run.go
```diff
@@ -68,6 +93,33 @@ against the Grafana k6 cloud. Use the "k6 cloud login" command to authenticate.`
...
+	flags.BoolVar(&c.showCloudLogs, "show-logs", c.showCloudLogs,
+		"enable showing of logs when a test is executed in the cloud")
```
joanlopez, line 112: Same as above? Not sure if it does really make sense.

cmd/cloud_run.go
```diff
@@ -68,6 +93,33 @@ against the Grafana k6 cloud. Use the "k6 cloud login" command to authenticate.`
...
+	flags.BoolVar(&c.uploadOnly, "upload-only", c.uploadOnly,
+		"only upload the test to the cloud without actually starting a test run")
```
joanlopez, line 114: See https://github.com/grafana/k6/pull/3850

cmd/cloud_run.go
oleiade: Great idea, let's look into it 👍🏻

cmd/cloud_run.go
oleiade: Dropping this :+1:

codebien:
>the idea of scoping options with a dot separator like in Prometheus

If we already envision the need for it, isn't better to have a dedicated subcommand? I mean something like `k6 cloud local-run --send-logs script.js`

oleiade:
Closing in favor of #3904

Issue #2497, opened 2022-04-14
grpc: Remove the strict JSON dependency


codebien:
It could be useful to remove the tight-coupled dependency that the gRPC module has from `protojson`.
https://github.com/grafana/k6/blob/57e994242f10122bccfbda229c4b29499a42457b/js/modules/k6/grpc/client.go#L232-L241

The `protojson` dependency is a requirement for the `js/k6/grpc` module for creating requests when the JavaScript objects are used for passing data.

However, in the case of extensions, this requirement could be removed where the generated Go structs could be directly used without the requirement for marshalling/unmarshalling from JSON.

We could add a field into the new `grpcext.Request` for providing the required marshaller/unmarshaller.

DanielPerfana:
Hi, any progress on this? We are running into issues with an system under test that returns 5Mb json responses, resulting in high cpu and memory usage of k6 and a response times overhead of approximately 100ms. It would be nice to have an option equivalent to `discardResponseBodies` for grpc.

olegbespalov:
Hey @DanielPerfana . 

I'm sorry for not getting back to you sooner. Unfortunately, there are no updates on this.

However, your request for an equivalent to the HTTP module `discardResponseBodies` sounds interesting. I'll bring it to the upcoming team sync and reply. Maybe we could extract this into separate minor issues and implement it sooner than removing the  `strict JSON dependency.`

lzakharov:
Hi, @olegbespalov!

Any updates about the `discardResponseBodies` equivalent? Has the team decided to decline this option or not?

PR #3820, opened 2024-07-01
Add `discardResponseMessage` option for gRPC client

lzakharov:
## What?

Adds a new gRPC client option `discardResponseMessage`, which is equivalent to `discardResponseBodies`. 

## Why?

Parsing heavy responses can result in high CPU and memory usage.

## Checklist

## Related PR(s)/Issue(s)

https://github.com/grafana/k6/issues/2497


codebien (commented):
Hi @lzakharov,
sorry for the late review. Thanks for your contribution :bow:

did you run the profiler to make sure it reduces the memory as expected?

After checking a bit the gRPC library, I assume that we are not discarding the response body as we do in HTTP.
With the suggested code, we are mostly only skipping the JSON parsing. That might lead to not being very useful in reducing memory usage. It might be more impactful on the CPU.

lzakharov:
Hi, thanks for your review comment!

By reducing memory consumption, I meant the absence of the need to allocate objects when parsing into JSON. I don't yet know how to ignore a message at the gRPC client side, need to do some research. My main focus was on CPU consumption, since this is the problem we encountered in our tests.

lzakharov:
After some research, I discovered that [emptypb.Empty](https://pkg.go.dev/google.golang.org/protobuf@v1.34.2/types/known/emptypb#Empty) can be used as a response message type to reduce memory consumption. It still requires reading all response data into the bytes buffer [here](https://github.com/grpc/grpc-go/blob/d27ddb5eb5940c949f88bc2cb21eed9254f8be75/rpc_util.go#L631), but I haven't found how to optimize this. Anyway, the current solution shows a noticeable improvement in performance.

lib/netext/grpcext/conn.go
```diff
@@ -26,16 +26,20 @@ import (
 	"google.golang.org/protobuf/reflect/protoreflect"
 	"google.golang.org/protobuf/types/descriptorpb"
 	"google.golang.org/protobuf/types/dynamicpb"
+	"google.golang.org/protobuf/types/known/emptypb"
 )
 
+var emptyDescriptor = (&emptypb.Empty{}).ProtoReflect().Descriptor() //nolint:gochecknoglobals
```
codebien, line 32: Having this global might lead to a data race. I don't know how `Descriptor` is used by gRPC internally. But if they write some field then when k6 runs with multiple VUs it will generate a data race.

And, even if they are not doing it now, it might happen in the future if they apply changes. I don't think the API makes the guarantee to be concurrent-safe.

So we need to have a dedicated copy for it. Make sure to re-run the benchmark to see if the changes are not dependent on this specific global approach.

codebien (changes requested):
Hey @lzakharov,
the results are quite interesting! Thanks for taking care of it.

If we can find an alternative solution to the current global variable then we are fine with merging your proposal. :+1:

lib/netext/grpcext/conn.go
lzakharov: Yes, intuitively it looked safe, but I agree with the comment. Benchmarks didn't show much of a difference.

lzakharov:
> It would be helpful if you could do something similar also for gRPC streaming. You can split it in a second pull request.

I'll try to add it in the next PR.

PR #3824, opened 2024-07-03
Update to xk6-websockets@v0.6.0

joanlopez:
## What?

Updates to the latest version of xk6-websockets, which includes support for [Blob](https://developer.mozilla.org/en-US/docs/Web/API/Blob).

See https://github.com/grafana/xk6-websockets/pull/74 for further details.

## Checklist

## Related PR(s)/Issue(s)

Contributes to https://github.com/grafana/xk6-websockets/issues/35

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -19,7 +19,7 @@ require (
 	github.com/grafana/xk6-output-prometheus-remote v0.4.0
 	github.com/grafana/xk6-redis v0.3.0
 	github.com/grafana/xk6-webcrypto v0.4.0
-	github.com/grafana/xk6-websockets v0.5.1
+	github.com/grafana/xk6-websockets v0.6.0
 	github.com/grpc-ecosystem/go-grpc-middleware v1.4.0
 	github.com/influxdata/influxdb1-client v0.0.0-20190402204710-8ff2fc3824fc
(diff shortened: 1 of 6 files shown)


joanlopez:
Open question: should we tag it as `breaking-change`? In my opinion, even if it's kinda expected/permitted/accepted (because the module is _experimental_), it's nice to list it as part of the breaking changes within Release Notes, so users can easily notice it, but I'm not sure what we usually do in such cases.

cc/ @mstoykov @codebien

codebien:
>In my opinion, even if it's kinda expected/permitted/accepted (because the module is experimental)

It remains a breaking change, and it still makes sense to announce it. IMHO, the experimental concept is something on top. In the event we were strictly following semver, the fact that the module is experimental just allows us to not bump a major version.

joanlopez:
> > In my opinion, even if it's kinda expected/permitted/accepted (because the module is experimental)
> 
> It remains a breaking change, and it still makes sense to announce it. IMHO, the experimental concept is something on top. In the event we were strictly following semver, the fact that the module is experimental just allows us to not bump a major version.

I agree!

codebien (approved):

mstoykov (approved):

PR #3835, opened 2024-07-09
Drop support for user & pass login

joanlopez:
## What?

Drop k6 Cloud login support for email and password login: this includes the interactive cli triggered when running k6 cloud login, it should instead request your API token.

- [ ] Add warning about when it will be removed (v0.55)

## Checklist


cmd/cloud_login.go
```diff
@@ -30,7 +28,7 @@ func getCmdCloudLogin(gs *state.GlobalState) *cobra.Command {
 
 	// loginCloudCommand represents the 'cloud login' command
 	exampleText := getExampleText(gs, `
-  # Log in with an email/password.
+  # Prompt for a k6 Cloud token.
```
joanlopez, line 31: Should we keep both: prompting for a token when running `k6 cloud login` and `-t [TOKEN]`?

It sounds a bit redundant, and as we're planning to ship it in v1.0, I think it deserves some care in terms of carefully thinking of its implications.

Some thoughts:
- In favor of `k6 cloud login`: 
  - Default behavior, user friendly walkthrough (although it could be replaced with a decent help)
  - Doesn't fully (only partially) break the currently existing, interactive, behavior.
- In favor of `-t [TOKEN]`:
  - Friendlier for CLI environments (less interactivity is required)
  - More flexible for future auth mechanisms (extensibility after v1.0)

cmd/cloud_login.go
```diff
@@ -79,18 +77,6 @@ func (c *cmdCloudLogin) run(cmd *cobra.Command, _ []string) error {
 		}
 	}
 
-	// We want to use this fully consolidated config for things like
-	// host addresses, so users can overwrite them with env vars.
-	consolidatedCurrentConfig, warn, err := cloudapi.GetConsolidatedConfig(
```
joanlopez, line 84: The consolidated config was used to initialize the `cloudapi.Client`, no longer needed.

cmd/cloud_login.go
```diff
@@ -126,29 +108,7 @@ func (c *cmdCloudLogin) run(cmd *cobra.Command, _ []string) error {
...
-
-		client := cloudapi.NewClient(
-			c.globalState.Logger,
-			"",
-			consolidatedCurrentConfig.Host.String,
-			consts.Version,
-			consolidatedCurrentConfig.Timeout.TimeDuration())
-
-		var res *cloudapi.LoginResponse
-		res, err = client.Login(email, password)
```
joanlopez, line 140: I think that would be nice to have a similar method but to check the correctness of the given token. 
Like a health endpoint, but with authentication. Wdyt? cc/ @oleiade @dgzlopes

PS: I think that's not supported yet, but would require some changes in Backend.

dgzlopes: I agree! It would be nice to validate the token. 

Can we do this by just trying to call one of the authenticated APIs? But yeah, a dedicated endpoint maybe would be better. I think you should involve someone from backend :eyes:

cmd/cloud_login.go
dgzlopes: I would like to invest time in improving the interactive flow in the future (picking default stack, magic link to the browser to authenticate, etc). But yeah, no strong opinions on what we do with it as of today. It is true that if it is only to pass the token it feels a bit redundant.

cmd/cloud_login.go
joanlopez: We discussed internally with the Backend team, and they'll add support for the `/login` endpoint to also accept a `token` for auth (as an alternative for now, for backwards compatibility, and as the unique method support in the future).

However, as that will take some time, I guess we can move forward with this PR and add that in the next iteration (I'll create an issue to do not forget).

joanlopez: Sure @codebien! The purpose of such check would be to make the resulting user experience better.

Now, when the user logs in with email & password, they get immediate feedback whether those credentials were correct or not, and the token saved is always valid, as it is returned by the API.

However, after these changes, the user won't log in, but would just set a token. So, if we don't add an extra check to verify whether the token is valid or not, the user won't receive such feedback until they try to run another command. While, in my opinion, would be much better to give them the feedback of whether the token was valid or not at time of running `k6 cloud login`.

Does that sound good to you? Do you have any other suggestion?

Thanks! 🙇🏻

codebien: Gotcha, that sounds good to me and valuable! :tada:

cmd/cloud_login.go
joanlopez: Then, I'd probably vote to keep both but generally recommend the use of `-t`, I guess 🤔

cmd/cloud_login.go
mstoykov: I would argue that this was the behaviour before for `-t` on the cli.

I am not against both that and this to do some checks, but I would not block this PR on endpoint to check this.

joanlopez: > I would argue that this was the behaviour before for `-t` on the cli.

Yeah, that was my initial thought, but having being like this doesn't mean we cannot improve the user experience, imo. Plus, the use of token was arguably an alternative for the main login flow with standard credentials (although I know that in practice that's no longer the case), but now it will become the main (and unique?), and so I think there's more reason to keep the main login flow with a good user experience, as it used to be the case for email and password.

> I am not against both that and this to do some checks, but I would not block this PR on endpoint to check this.

Yeah, I agree! That's why said above:

```
However, as that will take some time, I guess we can move forward with this PR and add that in the next iteration (I'll create an issue to do not forget).
```

cmd/cloud_login.go
oleiade: Sorry for the delay, I agree, we should keep both: prompt for the token, and support `-t` 👍🏻

cmd/cloud_login.go
```diff
@@ -0,0 +1,138 @@
...
+		newCloudConf.Token = null.StringFromPtr(nil)
+		printToStdout(c.globalState, "  token reset\n")
+	case show.Bool:
+	case token.Valid:
+		newCloudConf.Token = token
+	default:
+		form := ui.Form{
+			Banner: "Enter your token to authenticate with k6 Cloud.\n" +
+				"Please, consult the k6 Cloud documentation for instructions on how to generate one:\n" +
+				"https://grafana.com/docs/grafana-cloud/testing/k6/author-run/tokens-and-cli-authentication",
```
codebien, line 98: Should we link directly to the Grafana app page where the token is?

joanlopez:
Closing in favor of https://github.com/grafana/k6/pull/3886

Issue #3856, opened 2024-07-18
Implement `import.meta.resolve`


mstoykov:
### Feature Description

Implement `import.meta.resolve` as in https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import.meta/resolve 

There are already problems with what `require` , `open`, `k6/experimental/fs.open` and `k6/net/grpc.Client#load` are relative to or have been in the past (https://github.com/grafana/k6/issues/3534)

`import.meta.resolve` will allow users to get path relative in a consitent way and even get a path relative to one file but use a helper function from another. 

```
import { helperOpen } from "./some/path/that/has/nothing/to/do.js"

helperOpen(import.meta.resolve("./path/relative/to/this/one.csv"));
```

Here `helperOpen` will be able to be implemented with one of the `open` functions without users needing to figure out what is relative to what.

### Suggested Solution (optional)

Implement it as part of k6 and have it globally available.

PR #3873, opened 2024-07-26
Implement import.meta.resolve()

mstoykov:
## What?

This implements and tests `import.meta.resolve()` a way to get the path to a module the same way ESM and/or `require` will do.

This doesn't read or load the file in question from the file.

This also showed that we currently always parse code as ECMAScript modules. Even when the parsed/compiled file was to be wrapped as CommonJS.

This likely had no other side effects as all other ESM specific syntax is also either not implemented in k6 or forces CommonJS wrapping disabled.

## Why?

This lets people use `open`, `require` , `experimental/fs.open`, etc. and have a way to always be relative to the module it is written in. 

The fixes to compiler were as otherwise CommonJS modules don't error on `import.meta` and doesn't just isn't supported in the standard.
## Checklist

## Related PR(s)/Issue(s)

#3856


codebien (dismissed):

js/bundle.go
```diff
@@ -451,6 +451,16 @@ func (b *Bundle) setInitGlobals(rt *sobek.Runtime, vu *moduleVUImpl, modSys *mod
 	}
 	warnAboutModuleMixing("module")
 	warnAboutModuleMixing("exports")
+
+	rt.SetFinalImportMeta(func(o *sobek.Object, mr sobek.ModuleRecord) {
+		_ = o.Set("resolve", func(specifier string) (string, error) {
```
joanlopez, line 456: Shouldn't we do something with the error in case there's any? If it isn't supposed to happen, at all, I'm generally in favor of panicking.

js/modules/require_impl.go
```diff
@@ -87,6 +87,17 @@ func (ms *ModuleSystem) getModuleInstanceFromGoModule(wm *goModule) (wmi *goModu
 	return gmi, nil
 }
 
+// Resolve returns what the provided speicifer will get resolved to if it was to be imported
```
joanlopez, line 90:
```suggestion
// Resolve returns what the provided specifier will get resolved to if it was to be imported
```

joanlopez (approved):
I left a couple of minor comments, but it looks generally good!

joanlopez (approved):

codebien (approved):

PR #3879, opened 2024-08-01
Deprecation notice for email & password login (k6 cloud login)

joanlopez:
## What?

Displays a warning, telling the user that support for email and password authentication will be dropped by v0.55.
Look at the screenshots below:

image

Captura de pantalla 2024-08-01 a las 11 06 04

## Why?

Because we're planning to drop that support (see #3835), and we would like to follow our procedure of warning the users two versions ahead.

## Checklist

### DIFF
--- a/cmd/cloud_login.go
+++ b/cmd/cloud_login.go
@@ -23,6 +23,13 @@ type cmdCloudLogin struct {
 	globalState *state.GlobalState
 }
 
+const warning = `
+[Warning]
+Support for authentication with email and password will be removed after November 11th, 2024 (v0.55.0). 
+Ensure you migrate to token-based authentication by then.
+
+`
+
 func getCmdCloudLogin(gs *state.GlobalState) *cobra.Command {
 	c := &cmdCloudLogin{
 		globalState: gs,
@@ -45,13 +52,14 @@ func getCmdCloudLogin(gs *state.GlobalState) *cobra.Command {
 	loginCloudCommand := &cobra.Command{
 		Use:   cloudLoginCommandName,
 		Short: "Authenticate with Grafana Cloud k6",
-		Long: `Authenticate with Grafana Cloud k6.
+		Long: fmt.Sprintf(`Authenticate with Grafana Cloud k6.
 
 This command will authenticate you with Grafana Cloud k6.
 Once authenticated you can start running tests in the cloud by using the "k6 cloud run"
 command, or by executing a test locally and outputting samples to the cloud using
 the "k6 run -o cloud" command.
-`,
+%s
+`, warning),
 		Example: exampleText,
 		Args:    cobra.NoArgs,
 		RunE:    c.run,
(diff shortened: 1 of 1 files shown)


cmd/cloud_login.go
```diff
@@ -68,6 +76,9 @@ the "k6 run -o cloud" command.
 //
 //nolint:funlen
 func (c *cmdCloudLogin) run(cmd *cobra.Command, _ []string) error {
+	warnColor := getColor(c.globalState.Flags.NoColor || !c.globalState.Stdout.IsTTY, color.FgRed)
```
codebien, line 79:
```suggestion
	warnColor := getColor(c.globalState.Flags.NoColor || !c.globalState.Stdout.IsTTY, color.FgYellow)
```
Aren't warnings generally yellow/orange? :thinking:

joanlopez: I'm fine with any, what do you suggest @olegbespalov @oleiade? 
Let's go with `color.FgYellow` as suggested by @codebien?

oleiade: Indeed 👍🏻

cmd/cloud_login.go
```diff
@@ -45,13 +52,14 @@ func getCmdCloudLogin(gs *state.GlobalState) *cobra.Command {
...
-		Long: `Authenticate with Grafana Cloud k6.
+		Long: fmt.Sprintf(`Authenticate with Grafana Cloud k6.
 
 This command will authenticate you with Grafana Cloud k6.
 Once authenticated you can start running tests in the cloud by using the "k6 cloud run"
 command, or by executing a test locally and outputting samples to the cloud using
 the "k6 run -o cloud" command.
-`,
+%s
+`, warning),
```
olegbespalov, line 62: There is a `Deprecated` in `cobra.Command ` by setting it so we can see the deprecated message for both info and usage. However, I am not sure if it's colored, so maybe you way is good to go :shrug:

olegbespalov (approved):

cmd/cloud_login.go
codebien: I'm in favor of both. I hope the deprecation is useful for the help command.

joanlopez: Yeah, I thought about the `Deprecated` field, as we discussed in the other PR, but in this case I decided to go this way because the command per se it's not deprecated, indeed a new one, added in the upcoming release. So, once we drop the support we can just remove the warning, and keep the command as-is.

Wdyt?

codebien: >this way because the command per se it's not deprecated

That's true. Fair enough. Then, it sounds fine to me to move as-is.

olegbespalov: I feel most important is the end goal, notifying the customers that the command is deprecated, and will be removed is reached. Keeping in mind the timeline, I'm happy with the current changes :+1: Resolving the tread from my side.

codebien (approved):

joanlopez:
Closing in favor of https://github.com/grafana/k6/pull/3886, which basically removes this support from the recently introduced `k6 cloud login`, so we directly ship the new command without that support (and so no warning is needed).

PR #3883, opened 2024-08-01
Collect the Cloud test run ID

codebien:
## What?

When it is available, add the Cloud test run id to the usage report.

## Why?

In our telemetry system, we need to differentiate non-Cloud from Cloud tests.
Just a flag is not enough, we need the ID to aggregate reports when they come from distributed test runs.

### DIFF
--- a/cmd/report.go
+++ b/cmd/report.go
@@ -13,15 +13,24 @@ import (
 )
 
 type report struct {
-	Version    string         `json:"k6_version"`
-	Executors  map[string]int `json:"executors"`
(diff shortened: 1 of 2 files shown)


joanlopez (approved):

mstoykov:
Finally found time to finish https://github.com/grafana/k6/pull/3917 - a proposal to do this in a more generic way. 

There are definitely some additional things that need to be figure out in there, but after some tries I decided to just publish something that we can discuss over.

codebien:
Closing in favor of #3917,

PR #3886, opened 2024-08-05
Drop support for log in with email & password in the new `k6 cloud login`

joanlopez:
**Supersedes #3835** 

## What?

It drops the support for the interactive login with email and password from the [recently introduced](https://github.com/grafana/k6/pull/3813) `k6 cloud login` subcommand.

## Why?

Because that authentication mechanism is considered deprecated, and so we only leave the support for it through the deprecated `k6 login cloud` subcommand, but not in the new `k6 cloud login` command.

## Checklist

### DIFF
--- a/cmd/cloud_login.go
+++ b/cmd/cloud_login.go
@@ -2,7 +2,6 @@ package cmd
 
 import (
 	"encoding/json"
-	"errors"
 	"fmt"
 	"syscall"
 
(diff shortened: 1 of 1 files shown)


codebien (approved):
I see some potential improvements we should make to polish a bit the UX. We can do that during the next cycle.

The important part has been addressed. Thanks @joanlopez :bow:

mstoykov (approved):

Issue #1174, opened 2019-09-24
setup/teardownTimeout behavior when set to 0s


cuonglm:
Currently, setup/teaddownTimeout duration if set to `0s` means we can't run setup/teardown functions anymore. With script:

```
import http from 'k6/http';
import { k6 } from 'k6';

export function setup() {
  sleep(1);
}

export default function() {
  const response = http.get("http://test.loadimpact.com");
};
```

Run:

```
$ K6_SETUP_TIMEOUT=0s k6 run t.js

          /\      |‾‾|  /‾‾/  /‾/
     /\  /  \     |  |_/  /  / /
    /  \/    \    |      |  /  ‾‾\
   /          \   |  |‾\  \ | (_) |
  / __________ \  |__|  \__\ \___/ .io

  execution: local
     output: -
     script: t.js

    duration: -,  iterations: 1
         vus: 1, max: 1

ERRO[0000] Setup timeout                                 error="setup: Timeout during setup"
ERRO[0000] Setup timeout
exit 100
```

In most libraries, setting timeout duration to `0` often means disable the timeout. Should we adopt that behavior to `k6` as well?

cuonglm:
We do have `minIterationDuration` which will be disable when set to 0.

JustJordanT:
Is this something that still need to be worked on?

mstoykov:
Hu @JustJordanT - yes it appears nobody has worked on it and it still a problem.

tsukasaI:
Hi, can I take this one?

tsukasaI:
Hi, I opened [PR](https://github.com/grafana/k6/pull/3898).
I would appreciate your reply. Thank you!

mstoykov:
I don't think we should have a way to disable the timeout in more or less the same vein as #2701 - there is no need for that.

The current proposed PR disables the setup/teardown execution, but I am not particularly certain that is a good idea either. There already is an option for that, and I will expect that ... yes most people will expect this to mean it is infinite, and then it just doesn't run.

I am kind of for this just being a validation error similar to if you define it as negative number.


cc @grafana/k6-core

tsukasaI:
@mstoykov 

Thank you for comment.

Oh Yeah, I would think we should have a validation for that this property is positive number.

Can I introduce the validation?

tsukasaI:
Hi @mstoykov 
I updated my PR. #3898 

I would appreciate your review. Thank you!

mstoykov:
[...]

PR #3898, opened 2024-08-15
validate setupTimeout option

tsukasaI:
## What?

Validate setuptimeout option

## Checklist

## Related PR(s)/Issue(s)

https://github.com/grafana/k6/issues/1174

## Manual test

Make t.js.

```js
import http from 'k6/http';

export function setup() {
  console.log("setup called")
}

export default function() {
  const response = http.get("http://test.loadimpact.com");
};
```

Run this command with setupTimeone is 0s, and obtain error.

```sh
$ K6_SETUP_TIMEOUT=0s go run . run t.js

          /\      |‾‾| /‾‾/   /‾‾/   
     /\  /  \     |  |/  /   /  /    

ERRO[0000] setupTimeout must be positive                
exit status 104
```


mstoykov:
:wave: @tsukasaI this PR is now empty.  I guessed you tried to revert them before making the changes discussed in the issue, but those have not been pushed.

Just pointing this out in case you think that was done and are waiting on us to review this - there is currently no code to review.

tsukasaI:
Oh, sorry.
Just converted Draft.

lib/options.go
```diff
@@ -406,7 +407,9 @@ func (o Options) Apply(opts Options) Options {
 	if opts.NoSetup.Valid {
 		o.NoSetup = opts.NoSetup
 	}
-	if opts.SetupTimeout.Valid {
+	if opts.SetupTimeout.Valid && opts.SetupTimeout.Duration <= 0 {
+		return o, errors.New("setupTimeout must be positive")
+	} else if opts.SetupTimeout.Valid {
```
mstoykov, line 412: I really don't think changing Apply signature is warranted. 

We do validation checks for most of the options, and they all happen in the respected places instead of in Apply - so please do not change the signature of Apply - that will also make the change a lot smaller

mstoykov (changes requested):

lib/options.go
tsukasaI: @mstoykov 
Thank you for comment.

I updated to move in Validate method.
Is it reasonable?

mstoykov (approved):

lib/options.go
```diff
@@ -519,7 +520,7 @@ func (o Options) Apply(opts Options) Options {
 func (o Options) Validate() []error {
 	// TODO: validate all of the other options... that we should have already been validating...
 	// TODO: maybe integrate an external validation lib: https://github.com/avelino/awesome-go#validation
-	var errors []error
+	var errorsSlice []error
```
olegbespalov, line 523: nit: I'd suggest using something more explicit like `validationErrors`, rather than bringing the type into the variable name, which doesn't improve the readability of the code

tsukasaI: Thank you for the suggestion!
Fixed.

mstoykov (approved):

olegbespalov (approved):
@tsukasaI LGTM! Thank you for your contribution!

Issue #3282, opened 2023-08-16
cloudRunLocally command


mstoykov:
This is issue is about making a `k6 run -o cloud` like command that does know and does more than just outputing metrics to the cloud. But still runs locally.

Even now the cloud output:
1. outputs metrics to the cloud - which is exactly what a metric output should do
2. since v0.46.0 now it can also output trace data ... which is a bit more "iffy" IMO but at least it uses the same input as a metric output, so this seems pretty okay.
3. It also can tell the cloud why the test stopped ... which is more of QoL thing than anything to do with metrics
4. It also can stop the test (if a particular env variable is set) if the output tells it.
5. Probably something else I forget about.

And with [#3259](https://github.com/grafana/k6/issues/3259) we now intend it to be able to configure log outputs. 

Not only does this seem like too many thing a metric output should do ... it makes a bunch of things harder and harder to do. 

As in this case for example log outputs are configured earlier than practically anything else. Moving this to a command might not remove the need for some log output refactoring, but at least means that we don't make metric outputs "all powerful". 

Also issues such as [adding log outputs as extensions](https://github.com/grafana/k6/issues/1939) and having [multiple log outputs](https://github.com/grafana/k6/issues/2287) might cross this as well. 

## Solution 0/Current situation

We keep adding [Output additions](https://github.com/grafana/k6/blob/647db132c8f4ab7828510812bd9f1718127be655/output/types.go#L62-L95) they keep getting more and more complicated. This also means that other parts of the system need to know about outputs and how they handle stuff and that they can actually do something to them. 

## Solution 1

Make a new command cloudRunLocally(variant 1) or make `k6 cloud` take an argument `--run-locally` (variant 2) 🤷‍♂ - naming IMO will be the hardest here.

But that command can do all of those things without keep adding stuff to one output and features that every other output will practically *never* use.

## Solution 2

While writing this I did notice that the argument is called "output" instead of "metric-output" which is likely an oversight, because .. well that was the only thing k6 cared about years ago.

As such *if* we decide that we want to rename a bunch of stuff we can have a new type `Output` that is combination of many outputs. There a few problems here the major one of which is naming - we already have cli arg `--output` which is for metric outputs, so breaking this will be a problem.

The benefit of this is mostly that users will not change the command but add cli args - which IMO isn't that big of a benefit.

From implementation standpoint we will likely want to make this extendable through xk6 which will add complexity.

So Solution 1 is kind of easier in that regard.

### Implementation note:
[...]

PR #3904, opened 2024-08-21
Implement the --local-execution mode for k6 cloud run

oleiade:
## What?

This PR adds support for a new `--local-execution` flag to the k6 cloud run command. Using it, users can run their test script or archive on their local machine while streaming the results to the cloud.

In the context of this PR, the goal is to reproduce the behavior of the `k6 run -o cloud` behavior; we will add support for archive upload, logs, and trace forwarding in future PRs.

## Considerations and open questions

* During the implementation, we realized with @joanlopez that having a dedicated subcommand, as opposed to a flag, could lead to a more compartimatenlized flag structure, and clearer inter-dependencies between them: with this PR some `k6 cloud run` flags are now only allowed/relevant in conjunction with `--local-execution`. However, after brainstorming, we couldn't find a satisfactory name, that wouldn't feel weird and confusing (I'm looking at you `k6 cloud run-locally`.
* Related to ☝🏻 we imported the `--linger` and `--no-usage-report` options from the `k6 run` command and made them available/functional only when `--local-execution` is passed. Do we consider this approach satisfactory?
* We have overridden the `getConfig` function to enforce the output to be set to `cloud`. This is the most straightforward way we found to do this. If you have a better idea, please feel free to suggest 🙇🏻 

## Checklist

## Related PR(s)/Issue(s)

ref #3818
ref #3282

### DIFF
--- a/cmd/cloud_run.go
+++ b/cmd/cloud_run.go
@@ -1,21 +1,62 @@
 package cmd
 
 import (
+	"fmt"
+
 	"github.com/spf13/cobra"
-	"go.k6.io/k6/cmd/state"
+	"github.com/spf13/pflag"
+	"go.k6.io/k6/execution"
+	"go.k6.io/k6/execution/local"
 )
 
 const cloudRunCommandName string = "run"
 
-func getCmdCloudRun(gs *state.GlobalState) *cobra.Command {
-	deprecatedCloudCmd := &cmdCloud{
-		gs:            gs,
-		showCloudLogs: true,
-		exitOnRunning: false,
-		uploadOnly:    false,
+type cmdCloudRun struct {
+	// localExecution stores the state of the --local-execution flag.
+	localExecution bool
+
+	// linger stores the state of the --linger flag.
+	linger bool
+
+	// noUsageReport stores the state of the --no-usage-report flag.
+	noUsageReport bool
+
+	// runCmd holds an instance of the k6 run command that we store
+	// in order to be able to call its run method to support
+	// the --local-execution flag mode.
+	runCmd *cmdRun
+
+	// deprecatedCloudCmd holds an instance of the k6 cloud command that we store
+	// in order to be able to call its run method to support the cloud execution
+	// feature, and to have access to its flagSet if necessary.
+	deprecatedCloudCmd *cmdCloud
+}
+
+func getCmdCloudRun(cloudCmd *cmdCloud) *cobra.Command {
+	// We instantiate the run command here to be able to call its run method
+	// when the --local-execution flag is set.
+	runCmd := &cmdRun{
+		gs: cloudCmd.gs,
+
+		// We override the loadConfiguredTest func to use the local execution
+		// configuration which enforces the use of the cloud output among other
+		// side effects.
+		loadConfiguredTest: func(cmd *cobra.Command, args []string) (
+			*loadedAndConfiguredTest,
+			execution.Controller,
+			error,
+		) {
+			test, err := loadAndConfigureLocalTest(cloudCmd.gs, cmd, args, getCloudRunLocalExecutionConfig)
+			return test, local.NewController(), err
+		},
 	}
 
-	exampleText := getExampleText(gs, `
+	cloudRunCmd := &cmdCloudRun{
+		deprecatedCloudCmd: cloudCmd,
+		runCmd:             runCmd,
+	}
+
+	exampleText := getExampleText(cloudCmd.gs, `
   # Run a test script in Grafana Cloud k6
   $ {{.}} cloud run script.js
 
@@ -25,7 +66,7 @@ func getCmdCloudRun(gs *state.GlobalState) *cobra.Command {
   # Read a test script or archive from stdin and run it in Grafana Cloud k6
   $ {{.}} cloud run - < script.js`[1:])
 
-	cloudRunCmd := &cobra.Command{
+	thisCmd := &cobra.Command{
 		Use:   cloudRunCommandName,
 		Short: "Run a test in Grafana Cloud k6",
 		Long: `Run a test in Grafana Cloud k6.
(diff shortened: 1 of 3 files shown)


cmd/cloud_run.go
```diff
@@ -38,12 +79,87 @@ Use the "k6 cloud login" command to authenticate.`,
...
+
+func (c *cmdCloudRun) run(cmd *cobra.Command, args []string) error {
+	if c.localExecution {
+		if cmd.Flags().Changed("exit-on-running") {
+			return fmt.Errorf("the --local-execution flag is not compatible with the --exit-on-running flag")
+		}
+
+		if cmd.Flags().Changed("show-logs") {
+			return fmt.Errorf("the --local-execution flag is not compatible with the --show-logs flag")
+		}
```
joanlopez, line 117: I just realized it 🙏🏻  shouldn't this be within the `preRun` function instead of here, as the rest of flags validations? 🤔

cmd/cloud_run.go
```diff
@@ -38,12 +79,87 @@ Use the "k6 cloud login" command to authenticate.`,
...
+
+		return c.runCmd.run(cmd, args)
+	}
+
+	// When executing in the cloud, we enforce the usage report to be deactivated.
+	c.noUsageReport = true
```
joanlopez, line 123: I guess in the future, we'll most likely change this for the 2nd iteration of https://github.com/grafana/k6-cloud/issues/2613 cc/ @codebien

codebien: No, this isn't a control we enable/disable here, we do it directly using the env var. Do we need this change?

oleiade: I'm sorry I don't understand. Should we keep the no usage report option, and should we force it to be set to true in this context?

codebien: We should have the same concept as we have right now with `k6 run -o cloud script.js`. In that case, the report is enabled.

If the report is enabled or not in the Cloud is not controlled by k6 but it is injected from the Cloud at runtime.

oleiade: So my understanding is that the actionables step for us here would be:
1. keep the option to support `k6 cloud run --local-execution --no-usage-report`
2. remove forcing it to true, as in this snippet, in the case of `k6 cloud run`, as it won't be useful anyway.

Is that correct?

codebien: >as it won't be useful anyway.

No, the risk is quite the opposite. It has an impact and it might generate unexpected conditions.

But let me restart because maybe I'm missing some context.

Why did we write this line change (`c.noUsageReport = true`)? What are the problem and the solution proposal?

oleiade: We added the `--linger` and `--no-usage-report` options to `k6 cloud run`, which do something when the `--local-execution` option is used. We want to reproduce the same behavior as if people used the `k6 run` command with the `-o cloud` option.

This specific line of code has been added to simulate that the `--no-usage-report` option we've added is always passed (set to true) if the user does not use the `--local-execution`; meaning they're using the `k6 cloud run` command and execute their test in the cloud, as opposed to locally.

cmd/cloud_run.go
oleiade: Good catch, indeed! 👀

cmd/cloud_run.go
oleiade: I have changed the behavior of the command according to our internal discussions:
* `k6 cloud run --no-usage-report` should ignore the flag, and make sure we don't send the report, simulating the behavior.
* `k6 cloud run --local-execution --no-usage-report`  should take the flag in consideration, and act accordingly.

cmd/cloud_run.go
```diff
@@ -38,12 +79,87 @@ Use the "k6 cloud login" command to authenticate.`,
...
+func (c *cmdCloudRun) run(cmd *cobra.Command, args []string) error {
+	if c.localExecution {
+		// Note that when running the k6 cloud run command with the --local-execution
+		// flag, we handle the no-usage-report flag here as we would do in the k6 run
+		// command.
+		return c.runCmd.run(cmd, args)
+	}
+
+	// When running the k6 cloud run command and executing in the cloud however, we
+	// explicitly disable the usage report.
```
codebien, line 122:
```suggestion
	// When running the `k6 cloud run` command explicitly disable the usage report.
```
I don't get what you mean with `and executing in the cloud however`.

cmd/tests/cmd_cloud_run_test.go
```diff
@@ -10,3 +15,44 @@ func TestK6CloudRun(t *testing.T) {
...
+	}
+
+	for _, tc := range testCases {
+		tc := tc
+
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+
+			ts := getSimpleCloudTestState(t, nil, setupK6CloudRunCmd, tc.cliArgs, nil, nil)
+			ts.ExpectedExitCode = -1
```
codebien, line 51:
```suggestion
			ts.ExpectedExitCode = -1
```
Shouldn't be an InvalidConfig type instead of generic `-1`?

cmd/cloud_run.go
```diff
@@ -38,12 +79,86 @@ Use the "k6 cloud login" command to authenticate.`,
...
+	if c.linger {
+		return fmt.Errorf("the --linger flag can only be used in conjunction with the --local-execution flag")
+	}
+
+	return c.deprecatedCloudCmd.preRun(cmd, args)
+}
+
+func (c *cmdCloudRun) run(cmd *cobra.Command, args []string) error {
+	if c.localExecution {
+		// Note that when running the k6 cloud run command with the --local-execution
```
olegbespalov, line 115: I'm having a hard time understanding the meaning of this comment since we don't handle the flag here. Is this comment relevant?

cmd/tests/cmd_cloud_run_test.go
olegbespalov: Double on that, it's probably just switching to `errext.WithExitCodeIfNone` for errors

oleiade: Good catch, done 👍🏻

cmd/cloud_run.go
oleiade: Good catch, this is a left over indeed 👍🏻

cmd/cloud_run.go
```diff
@@ -38,12 +83,92 @@ Use the "k6 cloud login" command to authenticate.`,
...
+		&c.linger,
+		"linger",
+		c.linger,
+		"only when using the local-execution mode, keeps the API server alive past the test end",
+	)
+	flags.BoolVar(
+		&c.noUsageReport,
+		"no-usage-report",
+		c.noUsageReport,
+		"only when using the local-execution mode, don't send anonymous stats to the developers",
```
olegbespalov, line 153: It is not in the scope of this PR, but we should adjust texting here since `send anonymous stats to the developers` sounds a bit incomplete and not clear. 

It's probably better to provide the link to documentation, like:
```suggestion
		"only when using the local-execution mode, don't send anonymous usage stats (https://grafana.com/docs/k6/latest/set-up/usage-collection/)",
```

olegbespalov (approved):

cmd/cloud_run.go
joanlopez: I think we just re-used what already existed there https://github.com/grafana/k6/blob/master/cmd/config.go#L33

In theory I'm not against your proposal, but then, we change it in both places, right?

joanlopez (dismissed):

cmd/cloud_run.go
olegbespalov: Yes, the proposal is relevant for the other place, too, and as I said, it's not the scope of the PR. I'm totally fine with adjusting texting in a separate PR, that's why I've approved this one.

olegbespalov (approved):

cmd/cloud_run.go
```diff
@@ -150,7 +150,7 @@ func (c *cmdCloudRun) flagSet() *pflag.FlagSet {
 		&c.noUsageReport,
 		"no-usage-report",
 		c.noUsageReport,
-		"only when using the local-execution mode, don't send anonymous stats to the developers",
+		"only when using the local-execution mode, don't send anonymous usage stats (https://grafana.com/docs/k6/latest/set-up/usage-collection/)",
```
joanlopez: I know the other place it's out of the scope of this PR; but to avoid discrepancies, could you please update the same text in other place there's `--no-usage-report` flag, please?

https://github.com/grafana/k6/blob/master/cmd/config.go#L33 🙏🏻

oleiade: Arf, cheers for pointing it out, I looked for the text pattern but couldn't find it... Addressing it 🙏🏻

olegbespalov (approved):

joanlopez (approved):

PR #3906, opened 2024-08-21
Introduce k6 cloud upload to replace --upload-only

joanlopez:
## What?

It introduces a new cloud subcommand: `k6 cloud upload` to replace the old `k6 cloud --upload-only`.

It also marks the old flag as deprecated, so it is no longer displayed in help, and a deprecation message is displayed when used, as follows: *`Flag --upload-only has been deprecated, use "k6 cloud upload" instead`*.

## Why?

As part of the task of revamping the k6 cloud commands.

## Checklist

## Related PR(s)/Issue(s)

N/A

### DIFF
--- a/cmd/cloud.go
+++ b/cmd/cloud.go
@@ -335,7 +335,7 @@ func (c *cmdCloud) flagSet() *pflag.FlagSet {
 	flags.BoolVar(&c.uploadOnly, "upload-only", c.uploadOnly,
 		"only upload the test to the cloud without actually starting a test run")
 	if err := flags.MarkDeprecated("upload-only", "use \"k6 cloud upload\" instead"); err != nil {
-		panic(err)
+		panic(err) // Should never happen
 	}
 
 	return flags
--- a/cmd/cloud.go
+++ b/cmd/cloud.go
@@ -12,18 +12,17 @@ import (
 	"sync"
 	"time"
 
-	"github.com/fatih/color"
 	"go.k6.io/k6/cloudapi"
+	"go.k6.io/k6/cmd/state"
 	"go.k6.io/k6/errext"
 	"go.k6.io/k6/errext/exitcodes"
 	"go.k6.io/k6/lib"
 	"go.k6.io/k6/lib/consts"
 	"go.k6.io/k6/ui/pb"
 
+	"github.com/fatih/color"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
-
-	"go.k6.io/k6/cmd/state"
 )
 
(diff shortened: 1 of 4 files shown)


cmd/cloud.go
```diff
@@ -335,6 +334,9 @@ func (c *cmdCloud) flagSet() *pflag.FlagSet {
 		"enable showing of logs when a test is executed in the cloud")
 	flags.BoolVar(&c.uploadOnly, "upload-only", c.uploadOnly,
 		"only upload the test to the cloud without actually starting a test run")
+	if err := flags.MarkDeprecated("upload-only", "use \"k6 cloud upload\" instead"); err != nil {
```
joanlopez, line 337: This must never error out; and the message it displays is displayed in the description.

cmd/cloud_login.go
```diff
@@ -29,16 +29,16 @@ func getCmdCloudLogin(gs *state.GlobalState) *cobra.Command {
 	// loginCloudCommand represents the 'cloud login' command
 	exampleText := getExampleText(gs, `
   # Prompt for a Grafana Cloud k6 token
-  {{.}} cloud login
+  $ {{.}} cloud login
```
joanlopez, line 32: Leftover from another PR, just trying to keep all commands coherent in style. I think it's small and scoped enough to be added here instead of creating a PR specifically for that.

olegbespalov (dismissed):
Nit: maybe you could consider switching the `upload the test` (in all places) to `upload the test data` or even `upload the test scripts and resources.` Perhaps it's just me, but `upload the test` sounds incomplete. Shrug: Anyway, for sure, this is a non-blocking thing.

codebien (changes requested):
LGTM

cmd/tests/cmd_cloud_upload_test.go
```diff
@@ -0,0 +1,157 @@
...
+
+func TestK6CloudUpload(t *testing.T) {
+	t.Parallel()
+
+	t.Run("TestCloudUploadNotLoggedIn", func(t *testing.T) {
+		t.Parallel()
+
+		ts := getSimpleCloudTestState(t, nil, setupK6CloudUploadCmd, nil, nil, nil)
+		delete(ts.Env, "K6_CLOUD_TOKEN")
+		ts.ExpectedExitCode = -1 // TODO: use a more specific exit code?
```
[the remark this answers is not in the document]
joanlopez: Nah, I think this is just a copy-paste from another test, that also asserts the same: the cloud command fails because the user is not logged in. I agree!

olegbespalov (approved):

Issue #1250, opened 2019-11-19
Emit the end-of-iteration metrics outside of the networking and JS code


na--:
After https://github.com/loadimpact/k6/pull/1203, the `iterations` metric is emitted as a [part of the `netext.NetTrail` struct](https://github.com/loadimpact/k6/blob/c5f6881d37c44b90b999ac2db4081d5614ab7aaa/lib/netext/dialer.go#L103-L150), together with the `iteration_duration`, `data_received`, and `data_sent` metrics. This somewhat makes sense, however it doesn't make sense to call that metric sample container `netext.NetTrail` (since only 50% of its metrics are network-related). It also doesn't make sense to [emit these metrics from the JS code](https://github.com/loadimpact/k6/blob/c5f6881d37c44b90b999ac2db4081d5614ab7aaa/js/runner.go#L497) - since they relate to iterations, they should be emitted by the thing that controls iterations... 

So, after #1007 is merged, we should refactor the code, probably like this:
- call the `SampleContainer` something else, `IterationMetrics` or something like that
- the `netext.Dialer` would still be needed, but it will only be concerned with the `data_sent` and `data_received` metrics, which would then be re-packaged in that new struct
- that struct should be assembled and emitted somewhere outside of the JS code, probably [here](https://github.com/loadimpact/k6/blob/3421078d34e9eb2dce507c2687eeb3981d1afc87/lib/executor/helpers.go#L78-L81) (that would also allow us to have more accurate non-JS unit tests, for example [this one](https://github.com/loadimpact/k6/blob/3421078d34e9eb2dce507c2687eeb3981d1afc87/core/local/local_test.go#L323))

na--:
@Shehab7osny, I've assigned you, let us know if you have any questions or problems with this.

mstoykov:
This was fixed in https://github.com/grafana/k6/pull/3908

PR #3908, opened 2024-08-22
Drop NetTrail - emit iterations metrics in runner instead of dialer

### LINKED ISSUE #1605: setup() function creates an iteration metric
`setup()` creates an iteration metric.

## Environment

- k6 version: 0.27.0
- OS and version: Linux / cloud execution

## Expected Behavior

I expected to see a single iteration.

## Actual Behavior
`setup()` function created a separate iteration. Interestingly the `teardown()` function doesn't create a new iteration.

## script
```javascript
import http from 'k6/http'
import { sleep } from 'k6';

export let options = {
  iterations: 1,
  vus: 1
}

export function setup() { // creates 1 iteration
  let res = http.get('http://test.k6.io?where=setup')
}
export default function () { // created 1 iteration
  sleep(10);
  let res = http.get('http://test.k6.io?where=defaultfunction')
}

export function teardown() { // created 0 iterations
  let res = http.get('http://test.k6.io?where=teardown')
}
```

Cloud link: https://app.k6.io/runs/public/7c0d02656de748608dfd68eed2f701c2

image

Unrelated side note: It's interesting that the total count of requests is 2. The request from the `teardown()` function isn't counted.

mstoykov:
## What?

Drop the NetTrail container and emit iterations metric in runner instead of in Dialer.

Also fix #1605 

## Why?

As part of looking over issues I found #1250 which has some additional work required. But dropping the specific NetTrail and moving the code around was really simple. 

FIxing some of the test took a bit more time :(

## Checklist

## Related PR(s)/Issue(s)
#1250 
#1605

### DIFF
--- a/execution/scheduler_ext_test.go
+++ b/execution/scheduler_ext_test.go
@@ -363,8 +363,8 @@ func TestSchedulerSystemTags(t *testing.T) {
 
 	expTrailPVUTags := expCommonTrailTags.With("scenario", "per_vu_test")
 	expTrailSITags := expCommonTrailTags.With("scenario", "shared_test")
-	expNetTrailPVUTags := testRunState.RunTags.With("group", "").With("scenario", "per_vu_test")
-	expNetTrailSITags := testRunState.RunTags.With("group", "").With("scenario", "shared_test")
+	expDataSentPVUTags := testRunState.RunTags.With("group", "").With("scenario", "per_vu_test")
+	expDataSentSITags := testRunState.RunTags.With("group", "").With("scenario", "shared_test")
 
 	var gotCorrectTags int
 	for {
@@ -378,7 +378,7 @@ func TestSchedulerSystemTags(t *testing.T) {
 			default:
 				for _, sample := range s.GetSamples() {
 					if sample.Metric.Name == metrics.DataSentName {
-						if sample.Tags == expNetTrailPVUTags || sample.Tags == expNetTrailSITags {
+						if sample.Tags == expDataSentPVUTags || sample.Tags == expDataSentSITags {
 							gotCorrectTags++
 						}
 					}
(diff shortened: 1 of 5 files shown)


execution/scheduler_ext_test.go
```diff
@@ -497,10 +501,12 @@ func TestSchedulerRunCustomTags(t *testing.T) {
 							gotTrailTag = true
 						}
 					}
-					if netTrail, ok := sample.(*netext.NetTrail); ok && !gotNetTrailTag {
-						tags := netTrail.Tags.Map()
-						if v, ok := tags["customTag"]; ok && v == "value" {
-							gotNetTrailTag = true
+					for _, s := range sample.GetSamples() {
+						if s.Metric.Name == metrics.DataSentName && !gotNetTrailTag {
```
olegbespalov, line 505: nit: after the change, this `gotNetTrailTag` could confuse

lib/netext/dialer.go
```diff
@@ -69,17 +69,14 @@ func (d *Dialer) DialContext(ctx context.Context, proto, addr string) (net.Conn,
 	return conn, err
 }
 
-// GetTrail creates a new NetTrail instance with the Dialer
+// Sample creates a new NetTrail instance with the Dialer
```
olegbespalov, line 72: How about naming this something like `ReadsWritesSamples` or `IOSamples`, something more precise? Also, a method comment not longer valid :cry:

olegbespalov (commented):
Looks good, left two minors

lib/netext/dialer.go
mstoykov: I meant it more like hte verb "sample" as in dialer samples whatever it has to sample and returns it. But it isn't a thing we usually do so I had it renamed

olegbespalov (approved):

joanlopez (approved):

PR #3912, opened 2024-08-23
Upgraded google.golang.org/grpc v1.64.1 => v1.65.0

olegbespalov:
## What?

This updates `google.golang.org/grpc`, which is nothing significant for the k6, but keeping the change diff tinier is better. 

## Why?

keeping the change diff tinier is better

## Checklist


mstoykov (approved):

mstoykov:
Maybe run `go mod tidy`? I guess some lines should be dropped given the CI failure.

olegbespalov:
> Maybe run go mod tidy? 

The local `go mod tidy` using go1.20 showed nothing. If I run this using the go 1.22.6, it produces the diff like [4806b2f](https://github.com/grafana/k6/pull/3912/commits/4806b2fb923544ef4e40617fe72f5041cb55ef2e). Maybe we should make a separate PR for bringing this toolchain which is now defaults behavior, WDYT?

go.mod
```diff
@@ -1,6 +1,8 @@
 module go.k6.io/k6
 
-go 1.20
+go 1.21
+
+toolchain go1.22.6
```
mstoykov, line 5: This effectively also removes 1.21 from the versions that will be used. 

I am not against this as we in practice likely will upgrade to 1.23 for v0.54 (?) but still probably relevant to be added to the description.

mstoykov:
> Maybe we should make a separate PR for bringing this toolchain which is now defaults behavior, WDYT

Yeah, that probably is a good idea as I would prefer to have it separate. 

Strangely grpc doesn't really specify 1.22 , so I guess we can specifiy 1.21.11 in the toolchain ... but let's discuss this in the separate PR. As I said in hte comment - we likely will upgrade to 1.23 for v0.54.0 either way.

mstoykov (approved):

joanlopez (approved):

PR #3930, opened 2024-09-04
Perform token validation as part of 'k6 cloud login'

joanlopez:
## What?

It performs token validation after as part of `k6 cloud login` process.

## Why?

As discussed in https://github.com/grafana/k6/pull/3886, it's nice to give immediate feedback to users when "logging-in", instead of waiting until they use any other cloud command.

## Checklist


cmd/cloud_login.go
```diff
@@ -90,6 +104,9 @@ func (c *cmdCloudLogin) run(cmd *cobra.Command, _ []string) error {
 		newCloudConf.Token = null.StringFromPtr(nil)
 		printToStdout(c.globalState, "  token reset\n")
 	case show.Bool:
+		valueColor := getColor(c.globalState.Flags.NoColor || !c.globalState.Stdout.IsTTY, color.FgCyan)
```
joanlopez, line 107: I moved these lines here, from the bottom of the file, as I think it only makes sense to print the token when providing the `-s` flag, as also discussed at https://github.com/grafana/k6/pull/3886.

Otherwise, it doesn't, because:
- When resetting, with `-r`, there's no token.
- When providing it with `-t`, it's already there.
- When using the form, we define it as a _sensitive_ field (so it's not displayed while being written).

If we want to show it in the last scenario, I'd just switch the form input type to regular text, otherwise (as it was until know) it is a bit incoherent.

cloudapi/api.go
```diff
@@ -258,3 +265,27 @@ func (c *Client) Login(email string, password string) (*LoginResponse, error) {
 
 	return &lr, nil
 }
+
+// ValidateToken calls the endpoint to validate the Client's token and returns the result.
+func (c *Client) ValidateToken() (*ValidateTokenResponse, error) {
+	url := fmt.Sprintf("%s/validate-token", c.baseURL)
```
mstoykov, line 271: nit: this can be a string concatination 
```suggestion
	url := c.baseURL + "/validate-token"
```

I am pretty sure a bunch of the other such operations here can be concatinations as well.

cloudapi/api.go
```diff
@@ -258,3 +265,27 @@ func (c *Client) Login(email string, password string) (*LoginResponse, error) {
...
+
+	data := struct {
+		Token string `json:"token"`
+	}{
+		c.token,
+	}
```
mstoykov, line 277: Maybe better if this is named struct `validateTokenRequest`

cmd/cloud_login.go
```diff
@@ -115,6 +134,26 @@ func (c *cmdCloudLogin) run(cmd *cobra.Command, _ []string) error {
...
+		)
+
+		var res *cloudapi.ValidateTokenResponse
+		res, err = client.ValidateToken()
+		if err != nil {
+			return err
+		}
+
+		if !res.IsValid {
+			return errors.New(`your API token is invalid, please generate a new one at https://app.k6.io/account/api-token`)
```
mstoykov, line 153: Don't we have a grafana url? as we are probably going to have to drop this either way soon.

cmd/cloud_login.go
```diff
@@ -115,6 +134,26 @@ func (c *cmdCloudLogin) run(cmd *cobra.Command, _ []string) error {
...
+			newCloudConf.Token.String,
+			consolidatedCurrentConfig.Host.String,
+			consts.Version,
+			consolidatedCurrentConfig.Timeout.TimeDuration(),
+		)
+
+		var res *cloudapi.ValidateTokenResponse
+		res, err = client.ValidateToken()
+		if err != nil {
+			return err
```
olegbespalov, line 149: We do some raw error returns inside the `client.ValidateToken()`. Let's wrap error here to make it more human-friendly, like:
```suggestion
			return fmt.Errorf("can't validate the API token: %s", err)
```

or a better version

cloudapi/api.go
oleiade: Agreed 👍

oleiade (dismissed):
I share most of the comments brought up so far, but consider those potential improvements rather than blockers.

olegbespalov (dismissed):

oleiade (approved):

cmd/cloud_login.go
```diff
@@ -115,6 +134,28 @@ func (c *cmdCloudLogin) run(cmd *cobra.Command, _ []string) error {
...
+			return fmt.Errorf("can't validate the API token: %s", err.Error())
+		}
+
+		if !res.IsValid {
+			return errors.New("your API token is invalid - " +
+				"please, consult the Grafana Cloud k6 documentation for instructions on how to generate a new one:\n" +
+				"https://grafana.com/docs/grafana-cloud/testing/k6/author-run/tokens-and-cli-authentication")
+		}
+	}
+
```
mstoykov, line 158: nit: I would probably move this to a function `validateToken` to not have to keep adding to the function and add nolint:funlen. 

`consoliudatedCurrentConfig` also seems like a thing only used for this.

mstoykov (dismissed):

joanlopez:
Hey folks! Sorry for the noise once approved, but I applied the changes suggested by Mihail [here](https://github.com/grafana/k6/pull/3930#discussion_r1749799272) at https://github.com/grafana/k6/pull/3930/commits/e0de560de85835dbc704302e9b9334fa0fbc8b08 🙏🏻

mstoykov (approved):

PR #3931, opened 2024-09-04
[2/2] Make k6 cloud run --local-execution upload test archive

oleiade:
## What?

This Pull request modifies the behavior of the `k6 cloud run` command when using the `--local-execution` mode. It makes uploading the test archive to the cloud a default. This is meant to support our internal efforts to improve the UX of our cloud offering.

This Pull Request also adds support for an additional `--no-archive-upload` flag to the `k6 cloud run` command, which enables users to prohibit uploading their test archive to our cloud.

### Archive upload

Following a recently published change in our platform, creating test runs now supports providing the test archive along the test run parameters. To provide an archive when creating a test run, the request is expected to be a multipart form instead of the traditional JSON body. Creating test runs without archives does not change and works as before. To support this change you will find that we had to modify some of the behavior of the cloud API client `CreateTestRun` method.

### Configuration aspects

We developed this PR hand-in-hand with @joanlopez and particularly struggled to set up its configuration aspects. We took the approach that felt the most convenient and practical to us, in lack of a better approach. If you have ideas or insights on how to handle that particular part specifically, please let us know 👍🏻 

## Checklist

## Related PR(s)/Issue(s)

#3904


oleiade:
The tests are now green. There is a test in the `cloudapi` package that's very flaky on windows because it expects log lines in a given order, and is time based.

I created a dedicated PR to try and cater to that flakiness. In the meantime, please still consider this ready for review 🦾

cmd/cloud_run.go
```diff
@@ -150,8 +156,13 @@ func (c *cmdCloudRun) flagSet() *pflag.FlagSet {
 		&c.noUsageReport,
 		"no-usage-report",
 		c.noUsageReport,
-		"only when using the local-execution mode, don't send anonymous usage "+
-			"stats (https://grafana.com/docs/k6/latest/set-up/usage-collection/)",
+		"only when using the local-execution mode, don't send anonymous stats to the developers",
```
mstoykov, line 159: can we keep the link to the docs here for the usage?

oleiade: Indeed, lost in rebase 👍🏻 Addressing it 🙇🏻

output/types.go
```diff
@@ -68,6 +68,13 @@ type WithThresholds interface {
 	SetThresholds(map[string]metrics.Thresholds)
 }
 
+// WithArchive is an output that can receive the archive object in order
+// to upload it to the cloud.
+type WithArchive interface {
+	Output
+	SetArchive(archive *lib.Archive)
+}
+
```
mstoykov, line 77: This is the exact thing https://github.com/grafana/k6/issues/3282 is trying to prevent - us keep adding "abstracitons" to the Output type that only makes sense for the cloud output and don't really make it better just make it do all the work. 

IMO in this case `k6 cloud run --local-execution` probahly shouldn't just the output *interface* at all and use the cloud output directly. 

Preferably it will make the test before cloud output `Start` gets called so it won't actually call `CreateTestRun`, but the `k6 cloud run --local-execution` does.

oleiade: I see your point, and I agree that it's a good idea. We actually considered a similar approach with @joanlopez. However, we encountered some challenges when trying to access the cloud output from the run entry point to handle the archiving and upload process as you described (before `Start` is called). Based on my understanding of your point, I think most of our implementation matches what you describe, besides indeed how we pass down "configuration" to the cloud output.

One of the reasons for this difficulty is that the cloud run command currently relies on the k6 run command internally but forces its output to cloud by overriding the loadConfiguredTest function. This setup makes it hard to achieve what you're suggesting without the workaround we've implemented here—adding an interface that the run command can detect and use to configure the cloud output.

Do you have any suggestions on how we could accomplish this (passing down a test archive to the cloud output)  without the `WithArchive` interface and `run` command specific logic we've introduced?

Or, are you proposing that we should define the logic for local execution directly in the cloud run command, rather than relying on the run command? This would give us more flexibility and control, potentially avoiding changes to the output interface and the run command, as discussed in #3282, but would result in a much bigger and riskier change. 

Please let me know if I’ve interpreted your suggestion correctly 🙇🏻

output/types.go
```diff
@@ -68,6 +68,13 @@ type WithThresholds interface {
 	SetThresholds(map[string]metrics.Thresholds)
 }
 
+// WithArchive is an output that can receive the archive object in order
+// to upload it to the cloud.
```
mstoykov, line 72:
```suggestion
// to upload it to the cloud.
// NOTE: This API is EXPERIMENTAL and may be changed, renamed or
// completely removed in a later k6 release.
```

output/types.go
mstoykov: Unfortunately I don't have a 10 line fix that will make this possible.

As explained in the linked issue, this will only get worse and for logs AFAIK practically impossible to be done without refactoring how we run k6. 

At this point it seems that the code in the "run" command will need to be refactored to be more composable and then reused in `k6 cloud run --local-execution`. 

Looking at the `cmd` package it seems like it might be a good idea if the `cmd` packages deals more wiht the configuraiton and parsing of cli flags and less wiht actually running the k6 app. Currently both of those are very intertwined but that likely is not a good idea.

As mentioned privately I am okay with us merging this for the purposes of getting this released in 0.54.0, but will definitely prefer if work on the actual refactoring and working towards the mentioned issue goals of stop making the output interface just one big pile of one off editions and the cloud output in an omnipotent part of the code.

mstoykov (approved):

olegbespalov (approved):

PR #3933, opened 2024-09-05
Add testutils.MakeMemMapFs test helper

oleiade:
## What?

This PR introduces a `MakeMemMapFs` test helper to the testutils package.

This helper function creates a new in-memory filesystem with the given files. It is particularly useful for testing code that interacts with the filesystem, as it allows the creation of a filesystem with a known state without having to create temporary directories and files on disk.

## Why?

While working on another PR, I was looking for a way to initialize a test Fs quickly. I found in the codebase occurences of a `makeMemMapFs` function, which was already duplicated twice in the codebase.

In the spirit of continuous improvement, instead of adding a third copy, I thought I would make a `testutil` function out of it.

## Checklist

### DIFF
--- a/lib/archive_test.go
+++ b/lib/archive_test.go
@@ -10,6 +10,8 @@ import (
 	"runtime"
 	"testing"
 
+	"go.k6.io/k6/lib/testutils"
+
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"gopkg.in/guregu/null.v3"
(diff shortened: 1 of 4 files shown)


joanlopez (approved):
Looks good to me, the only question I have is: now that we're planning v1.0, do we really want to expose it under: `/lib/...`, or should we (at least first) add it `internal`ly? 🤔  Although, it's also true that it might be useful for writing some extensions tests.

Let's see what others say.

mstoykov (approved):

PR #3935, opened 2024-09-06
js: DRY creation of a runner from archive in the tests

mstoykov:
## What?

Refactor js tests so we do not repeat the same pattern to create a runner from an archive.

## Why?

This was copy-pasted around and similar to how getSimpleRunner makes it hard to refactor if you add new fields that need to be around always.

We could also make a helper to create a runner + archive and run the provided test as function argument. But that would work in a lot less of the cases and isn't as necessary for #3917 to be a bit less noisy. 

## Checklist

### DIFF
--- a/js/console_test.go
+++ b/js/console_test.go
@@ -82,6 +82,37 @@ func getSimpleRunner(tb testing.TB, filename, data string, opts ...interface{}) 
 	)
 }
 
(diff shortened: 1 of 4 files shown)


olegbespalov (approved):

joanlopez (approved):
Nice idea, thanks! 😄

PR #3950, opened 2024-09-12
Add homebrew forumlar for k6 version v0.39.0

cheuk0324:
## Checklist

### DIFF
--- a/Formula/k6.rb
+++ b/Formula/k6.rb
@@ -1,0 +1,15 @@
+class K6 < Formula
+	desc "A modern load testing tool for developers and testers"
+	homepage "https://k6.io/"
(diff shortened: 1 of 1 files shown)


mstoykov:
I understand that this was unintentionally opened PR, but I would just like to ask why you are trying to get v0.39.0 which at this point is over 2 years old?

Thank you for your time @cheuk0324 :bow:

cheuk0324:
@mstoykov Apology for the Inconvenience. So we have tests thats only run with older version of the k6 but for some of the new engineers, they are not able to install the older through brew, because the k6 from homebrew is always the latest. So we tried to created the tap with the tag version. 

I hope there is an easier way that we can install older version of k6 from homebrew. Or is there better way to do that? Thanks!

PR #3968, opened 2024-09-24
Fix buffered chan event system

ankur22:
## What?

This removes the `default` case for when reading done events from the extension or when sending events to the extension.

## Why?

If we drop events due to the channel being full and there being a `default` case, then we can end up in a situation where either the event system indefinitely waits since incoming done events were not handled by the event system or the extension waits indefinitely if the event was dropped when sending it the event.

## Checklist

## Related PR(s)/Issue(s)

https://github.com/grafana/xk6-browser/issues/1440

### DIFF
--- a/event/system.go
+++ b/event/system.go
@@ -86,18 +86,16 @@ func (s *System) Emit(event *Event) (wait func(context.Context) error) {
 	doneCh := make(chan struct{}, s.eventBuffer)
 	doneFn := func() {
 		origDoneFn()
-		select {
-		case doneCh <- struct{}{}:
-		default:
-		}
+		// The done must be read by the reading side to prevent
+		// a goroutine that waits indefinitely.
+		doneCh <- struct{}{}
 	}
 	event.Done = doneFn
 
 	for _, evtCh := range s.subscribers[event.Type] {
-		select {
-		case evtCh <- event:
-		default:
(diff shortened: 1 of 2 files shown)


inancgumus (approved):
LGTM 👍 One point to keep in mind: More calls to `Done` will again block :)

ankur22:
> One point to keep in mind: More calls to `Done` than the number of subscribers will again block :)

Yeah, that's a good point. I hadn't considered the reason why there was a `default` in the first place.

Having looked at this again, I still feel that this is the correct approach. Dropping done events or the events going to the extension will mean that one of the two will misbehave, either the extension will hang indefinitely or k6 will.

In the extension side (i.e. in browser) we could call [e.Done()](https://github.com/grafana/xk6-browser/blob/a761731de4cba66caccadbe0ae30536147c2306e/browser/registry.go#L337) on a goroutine for the exit event. We don't want the browser module to hang around waiting for `e.Done()` to complete, so feels like a safe option. In my tests so far i've not needed to do this though, so i'd avoid doing that for now.

There is the chance that the subscriber could have called subscribe twice and handle the same subscription in the same goroutine, this could cause a deadlock (I think). This feels like user error though.

oleiade (approved):

Issue #2823, opened 2022-12-19
Display a warning when using `http.get` and `http.head` with more than 2 arguments


oleiade:
### Brief summary

The current prototype of the `http.get` and `http.head` methods can prove to be somewhat error-prone, as they take only two arguments, whereas the other HTTP verb-related methods do take 3. 

Although not a bug per-se, this UX can lead to confusion in scenarios like the one I ran into recently: trying to set headers on an HTTP request. Following the docs examples, I intuitively wrote `http.get('myurl', null, params)` with `params` containing my headers. But the `http.get` method would ignore my params completely, as it would treat the second parameter as `null` params instead of a `null` body.  

### k6 version

v0.40.0

### OS

*

### Steps to reproduce the problem

import http from "k6/http";
import { check } from "k6";
// import tracing from "k6/x/tracing";

const options = {
  vus: 1,
  iterations: 1,
};

tracing.instrument(http, {
  sampling: 12,
  propagator: "w3c",
  baggage: { "X-My-baggage": "something" },
});

// let client = new tracing.Client();

export default () => {
  const params = {
    headers: {
      "X-My-Header": "something",
    },
  };

  const res = http.get("http://localhost:3434/latency/50ms", null, params);
  check(res, {
    "status is 200": (r) => r.status === 200,
  });
};

### Expected behaviour

It's been discussed internally that to let the user know that this is a misuse of the `http.get` and `http.head` methods, we should emit a warning. Something indicating to the user that they're probably doing something wrong: "beware: the http.get and http.head methods do not take a payload argument" or something along those lines would probably do.

### Actual behaviour

⚔️

codebien:
>we should emit a warning.

@oleiade Why not directly an error?

codebien:
Hi @juliaokmenezes, yes, it still makes sense. Feel free to open a new PR for it or ask for details if something is not clear.

juliaokmenezes:
Hi @codebien, i will try it! Could you please provide, if possible, indications where in the project the HTTP validations are treated?

codebien:
Hey @juliaokmenezes,

You can find the `get` and `head` methods defined here https://github.com/grafana/k6/blob/001aff55869523cafc66a7abd6e50435cc2b0347/js/modules/k6/http/http.go#L71-L82.

It sounds like a viable way to define a small validation function for the logic required for this issue and then call it from those methods before invoking the `defaultClient.Request`.

NawafSwe:
[...]

PR #3974, opened 2024-10-02
fix/validate-args-length #2823

ArjitGopher:
## What?

This PR aims to solve the issue #2823

## Why?

1. The changes made will warn the user if more than two arguments are passed in http.get and http.head method
2. A helper function is added.


js/modules/k6/http/http.go
```diff
@@ -60,6 +60,13 @@ func (r *RootModule) NewModuleInstance(vu modules.VU) modules.Instance {
 		}
 	}
 
+	
+	validateArgCount := func(methodName string,args ...sobek.Value){
+	if len(args)>2 {
+		vu.State().Logger.Warningf("%s method has more than two arguments",methodName)
+		}
+	}
```
joanlopez, line 68:
```suggestion
	validateArgCount := func(methodName string,args ...sobek.Value){
	if len(args)>2 {
		vu.State().Logger.Warnf("http.%s method has more than two arguments",methodName)
		}
	}
```

Small suggestion

joanlopez (commented):
Looks like the new code is not correctly formatted, @ArjitGopher could you run the linter and fix the related issues, please?

Beyond that, I only have a small suggestion because I think these functions are usually called like `http.get(...)`, as suggested in the official docs, and in any case, it's always good idea to disambiguate. 

Other than that, it looks good! 👍🏻  
cc/ @olegbespalov

js/modules/k6/http/http.go
olegbespalov: I do agree with @joanlopez that it's better to add `http` prefix. 

Also, how about changing the message to something more actionable, like `http.%s requires two arguments, but %d were provided. Please adjust the input.`

olegbespalov (commented):
Thanks for the contribution!

I do agree with what @joanlopez said, also have a minor suggestion for the actual message

olegbespalov:
Hi @ArjitGopher. Out of curiosity, are you going to apply feedback, or you have no capacity of doing that?

ArjitGopher:
Hi @olegbespalov I agree with the suggestions and will add them today, Thanks

js/modules/k6/http/http.go
```diff
@@ -60,6 +60,14 @@ func (r *RootModule) NewModuleInstance(vu modules.VU) modules.Instance {
 		}
 	}
 
+	validateArgCount := func(methodName string, args ...sobek.Value) {
+		numberOfArgs := len(args)
+		if numberOfArgs > 2 {
+			vu.State().Logger.Warningf("http.%s method has more than two arguments, but %d were provided",
```
joanlopez, line 66: Nit; I'd prefer if we can use the "actionable" version of the message suggested by @olegbespalov 👉🏻 [here](https://github.com/grafana/k6/pull/3974#discussion_r1796491116).

js/modules/k6/http/http.go
```diff
@@ -71,12 +79,18 @@ func (r *RootModule) NewModuleInstance(vu modules.VU) modules.Instance {
 	mustExport("get", func(url sobek.Value, args ...sobek.Value) (*Response, error) {
 		// http.get(url, params) doesn't have a body argument, so we add undefined
 		// as the third argument to http.request(method, url, body, params)
+
+		// get method should not have more than two arguments
+		validateArgCount("get", args...)
 		args = append([]sobek.Value{sobek.Undefined()}, args...)
 		return mi.defaultClient.Request(http.MethodGet, url, args...)
```
joanlopez, line 86:
```suggestion
		// get method should not have more than two arguments
		validateArgCount("get", args...)
		// http.get(url, params) doesn't have a body argument, so we add undefined
		// as the third argument to http.request(method, url, body, params)
		args = append([]sobek.Value{sobek.Undefined()}, args...)
		return mi.defaultClient.Request(http.MethodGet, url, args...)
```

Nit; I'd prefer if we can keep the order as suggested here, as I think the already existing comment is tied to the `args` slice we build below, so it kinda makes more sense to keep both together. Same for the `"head"` method.

Could you confirm it, please @olegbespalov?

joanlopez (approved):

js/modules/k6/http/http.go
```diff
@@ -69,14 +78,24 @@ func (r *RootModule) NewModuleInstance(vu modules.VU) modules.Instance {
 	// wrappers (facades) that convert the old k6 idiosyncratic APIs to the new
 	// proper Client ones that accept Request objects and don't suck
 	mustExport("get", func(url sobek.Value, args ...sobek.Value) (*Response, error) {
+		// get method should not have more than two arguments
```
olegbespalov, line 81:
```suggestion
		// get method should not have more than one additional arguments
```

olegbespalov (changes requested):
I've tested and, checked the UI and actually spotted a bug :cry: 

So my script was:

```js
import http from "k6/http";

export default function () {
	http.get("https://test-api.k6.io/", {}, "lorem")
}
```

and it didn't produce any warning

When I re-looked into the code I found that `validateArgCount` doesn't get all arguments, it just gets all arguments, except `url` which is separated in the signatures.

So to make it work as expected, we need to adjust the logical expression inside validate, like `len(args) > 1`, but I'd also ask adjusting both comments (made inline suggestion)

olegbespalov:
As the team, we decided to close this PR for now. Happy to return to it if feedback will be addressed.

PR #3977, opened 2024-10-03
Make handleSummary func timeout configurable

### LINKED ISSUE #3976: Ability to configure the 120 second timeout when generating end of test statistics
### Feature Description

Hello,

At the moment, when we run longer tests (4-5 hours+), the end of test summary generation is failing with
`"handleSummary() execution timed out after 120 seconds"`

(tracking workarounds as part of the post on https://community.grafana.com/t/execution-timed-out-after-120-seconds/132323)

Would it be possible to add a way to configure this timeout, if we're running longer tests we don't really mind if it takes 10-20 minutes to process the stats at the end?

Thanks.

_No response_

_No response_

alex-kattathra-johnson:
## What?

It makes the timeout for the `handleSummary` function configurable

## Why?

To solve issue #3976. To allow for generating summary for long running tests

## Checklist

## Related PR(s)/Issue(s)
Closes #3976


taylorflatt:
I'd make sure to update the docs as well if this change does move forward: https://grafana.com/docs/k6/latest/using-k6/k6-options/reference/

See: https://github.com/grafana/k6-docs

js/summary_test.go
```diff
@@ -55,7 +55,10 @@ func TestTextSummary(t *testing.T) {
 			runner, err := getSimpleRunner(
 				t, "/script.js",
 				fmt.Sprintf(`
-					exports.options = {summaryTrendStats: %s};
+					exports.options = {
+						summaryTrendStats: %s,
+						handleSummaryTimeout: '120s',
+					};
```
mstoykov, line 61: Do you really need to update all of those tests ?

mstoykov (changes requested):
Thank you for the PR :bow: 

It seems to me like you are changing way too many tests and I expect there might be some problems with the propagation of the default.

Can you try to rever each of the test changes and see if you can reduce them? I haven't commented in each place.

It also will be nice to get a test case in https://github.com/grafana/k6/blob/master/cmd/config_consolidation_test.go#L146 as that is where consolidation is usually tested as well as defaults.

js/summary_test.go
alex-kattathra-johnson: I think the defaults are set at the `cmd` level. But, these unit tests are testing the functionality of these respective modules independent of `cmd` execution.

alex-kattathra-johnson: @mstoykov : Friendly ping on this! 😄

PR #3989, opened 2024-10-16
dependabot: add a bunch more dependencies to it

mstoykov:
## What?

Add more dependancies to be checked by dependabot

## Why?

While k6 still doesn't update all its dependencies, we do update quite a lot of them.

Currently there is a manual step of not only updating them and making a PR but also trying to gather what has changed.

While having a PR for every small change is likely going to be a lot noisier and we will have to ignore some of them. It seems like a better idea than having someone go through 20 repositories trying to figure out what commits have changed, given that dependabot does this automatically.

## Checklist

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,8 +8,20 @@ updates:
       # Keep the experimental modules up-to-date
       - dependency-name: "github.com/grafana/xk6-*"
         dependency-type: "all"
-    commit-message:
-      prefix: "Upgrade experimental module "
-      include: "scope"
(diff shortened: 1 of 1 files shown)


mstoykov:
I specifically didn't try to get everything in as a bunch of our dependencies have very rare updates. 

So this still requires someone to check that some other dependencies have not slipped by.  Over time, I expect we will keep adding to the list.

olegbespalov (approved):

joanlopez (approved):

PR #4007, opened 2024-10-22
Top level await

mstoykov:
## What?

top-level-await support

## Why?

It is part of the specification and it is hepful if you have async APIs ... which k6 starts to get more and more off.

## Checklist

### DIFF
--- a/js/bundle.go
+++ b/js/bundle.go
@@ -331,9 +331,10 @@ func (b *Bundle) instantiate(vuImpl *moduleVUImpl, vuID uint64) (*BundleInstance
 		env:          b.preInitState.RuntimeOptions.Env,
 		moduleVUImpl: vuImpl,
 	}
+	var result *modules.RunSourceDataResult
 	callback := func() error { // this exists so that Sobek catches uncatchable panics such as Interrupt
 		var err error
-		bi.mainModule, err = modSys.RunSourceData(b.sourceData)
+		result, err = modSys.RunSourceData(b.sourceData)
 		return err
 	}
 
@@ -345,6 +346,14 @@ func (b *Bundle) instantiate(vuImpl *moduleVUImpl, vuID uint64) (*BundleInstance
 	})
 
 	<-initDone
+
+	if err == nil {
+		var finished bool
+		bi.mainModule, finished, err = result.Result()
+		if !finished {
+			return nil, errors.New("initializing the main module hasn't finished, this is a bug in k6 please report it")
+		}
+	}
 
 	if err != nil {
 		var exception *sobek.Exception
--- a/js/modules/require_impl.go
+++ b/js/modules/require_impl.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 
 	"github.com/grafana/sobek"
+
 	"go.k6.io/k6/loader"
 )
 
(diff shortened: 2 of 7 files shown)


js/bundle.go
```diff
@@ -346,6 +347,14 @@ func (b *Bundle) instantiate(vuImpl *moduleVUImpl, vuID uint64) (*BundleInstance
 
 	<-initDone
 
+	if err == nil {
+		var finished bool
+		bi.mainModule, finished, err = result.Result()
+		if !finished {
+			return nil, errors.New("initializing the main module hasn't finished, this is a bug in k6 please report it")
```
olegbespalov, line 354: Are there any other essential details we could include in the message to help us identify the case when needed? Like error, something else, maybe even panic to have a stack trace? What do you think?

mstoykov: I would expect in all cases we will need the test to actually figure this out. It will likely be some very strange case as it will need to have the runnign the things on thevent loop to:
1. not have errored out
2. not have finished everything

The stacktrace here is ... very likely pointless as it wil just show us which executors is used, as the thing that hasn't finished is alredy past.

We can panic :shrug: I am just not into panicking all that much and again - I couldn't figure out a case where this can happen. This is here mostly just in case the unthinkable happens.

But I am okay with panicking if that seems better.

olegbespalov: If panic doesn't help, we shouldn't. 

My point is that, if possible, we can print out some details about the state that could be included in future bug reports, which could be handy for our future investigations.

oleiade (approved):
Took it for a spin locally, and you have no idea how having the ability to do this:
```javascript
import { open, SeekMode } from 'k6/experimental/fs'

const file = await open('package.json')

  // ...
}
```

Makes me happy! Great job @mstoykov, thanks a lot for making this happen, much appreciated! 🙇🏻 🎉 🚀

js/bundle.go
mstoykov: There isn't really any info at this point - we started to load modules, after the event loop had nothing else to run, the modules were still not loaded.

This will require us to get the test and see how it runs to figure it out.

olegbespalov (dismissed):

js/modules/require_impl.go
```diff
@@ -195,3 +197,11 @@ func getPreviousRequiringFile(vu VU) (string, error) {
...
 	return result, nil
 }
+
+// sets the provided promise in such way as to ignore falures
+// this is mostly needed as failures are handled separately and we do not want those to lead to stopping the event loop
+func promisesThenIgnore(rt *sobek.Runtime, promise *sobek.Promise) {
+	cal, _ := sobek.AssertFunction(rt.ToValue(promise).ToObject(rt).Get("then"))
+	handler := rt.ToValue(func(_ sobek.Value) {})
+	_, _ = cal(rt.ToValue(promise), handler, handler)
+}
```
joanlopez, line 207: _Nit; I suspect that's what you really wanted to write 😅_ 

```suggestion
func promisesThenIgnore(rt *sobek.Runtime, promise *sobek.Promise) {
	call, _ := sobek.AssertFunction(rt.ToValue(promise).ToObject(rt).Get("then"))
	handler := rt.ToValue(func(_ sobek.Value) {})
	_, _ = call(rt.ToValue(promise), handler, handler)
}
```

joanlopez: Is it fine to ignore the boolean return? I guess it's because we now it's a Promise, and `then` must always be a function, right? 🤔

joanlopez (dismissed):

oleiade (approved):

olegbespalov (approved):

joanlopez (approved):

olegbespalov:
It's worth updating also examples (like `experimental/fs`) and documentation to reflect the change :+1:

oleiade:
@olegbespalov 👍🏻 I'm on it.

PR #4009, opened 2024-10-23
Update Sobek and fix http.File to not use []byte

mstoykov:
## What?

Update Sobek

## Why?

Sobek updates bring fixes from goja for:
- nesting of destructuring with arrays
- using slices from go code in js code

Also fix for async errors in modules triggering the promise rejection tracker as well, which will be needed for top-level-await

The changes to handling of slices did break some tests and showcased that k6 was handling `http.File().data` badly. Previously it was `[]byte` which it really shouldn't be.

This fix makes certain that the provided input is `string` or `ArrayBuffer` and keeps it that way until it needs to be written.

This aligns with the documentation as well as removes a source of pure `[]byte` in js code which there shouldn't be any.

Separate PR should be made to drop being able to use `common.ToBytes` with `[]byte` as that should never happen and should be 100% a bug.

Also made the go code return a pointer as that just makes sense instead of a struct that will get copied.

## Checklist


oleiade (approved):
Cheers for the detailled explanation 🙇🏻

olegbespalov (approved):

manjunadh5369:
import { FormData } from 'https://jslib.k6.io/formdata/0.0.2/index.js'

Let  documentFile = http.get( <first api>)

const fd = new FormData();
 fd.append('file', http.file(documentFile.body, `${<file Name>}`, 'application/vnd.openxmlformats-officedocument.wordprocessingml.document'));

let response = http.put(<second api>, fd.body(), {headers: {Content-Type': 'multipart/form-data; boundary=' + fd.boundary}})

My code used to work earlier now after this release its working in the opensource k6 but failing in the grafana cloud with following error: {"statusCode":406,"error":"Not Acceptable","message":"The document appears to be corrupted and cannot be loaded."}

It appears that there may be a discrepancy in how file uploads are handled between the open-source version of k6 and Grafana Cloud, particularly after the recent release. Any guidance or updates regarding this issue would be greatly appreciated.

mstoykov:
My expectation is that if you set the response to be binary on the request getting the document it will work again. 

Can you try it? 

You need to set the request [params](https://grafana.com/docs/k6/latest/javascript-api/k6-http/params/) to have `responseType` equal to `binary`

manjunadh5369:
Thank you very much @mstoykov! It's working now.

mstoykov:
Glad I could help. 

To be honest I expect that before that thsi request was jsut a touch corrupted as going to string and back to binary as you did before this last change likely broke the file underneed.

I guess your API either doesn't check the send API *that* carefully, you were *very* lucky for the transition to be lossless, or the response is actually text and your API just expects it as binary data for some reason.

PR #4022, opened 2024-10-28
Replace strings.SplitN(2) with strings.Cut

joanlopez:
## What?

It updates the calls to `strings.SplitN`, with `n=2` with `strings.Cut`, because most of the times that's why `strings.Cut` was added into the Go stdlib, as it makes the code involved more readable.

## Why?

I started with a single `// TODO` I found while working on another task, but then realized that the codebase is plenty of calls to `strings.SplitN` with `n=2`, where ~most~ all the cases what we really want is to "cut" (split) a string containing key-values (e.g. `key:value`, `key=value`,  etc).

## Checklist


mstoykov (approved):
nice, great job!

olegbespalov (approved):

Issue #4050, opened 2024-11-11
Interop from ClojureScript doesn't seem to work since v0.53


trueneu:
### Brief summary

Starting from k6 v0.53, it is not possible to call any `http` methods from ClojureScript for no obvious reason. It seems that imports stopped working correctly.

### k6 version

v0.53, v0.54

### OS

Ubuntu 22.04

### Steps to reproduce the problem

I am using shadow-cljs to compile ClojureScript with the following relevant configuration:

```
            :target           :esm
            :js-options       {:keep-as-import #{"k6" "k6/http" "k6/crypto" "k6/encoding"}}
            :modules          {:main {:exports {default k6-test.core/default
                                                options k6-test.core/options
                                                setup   k6-test.core/setup}}}
            :compiler-options {:optimizations      :advanced
                               :output-feature-set :es5}}
```

Then, I import `k6/http` (the results are the same):
```
(:require ["k6/http" :as http])
_or_
(:require ["k6/http$default" :as http])
```

k6 v0.52 will make an http call if I do
```
(http/get url)
```
However, starting from v0.53, it complains there's no such method.

If I do `(js/Object.keys http)`, with v0.52 everything's fine:
```
["TLS_1_0","TLS_1_1","TLS_1_2","TLS_1_3","OCSP_STATUS_GOOD","OCSP_STATUS_REVOKED","OCSP_STATUS_SERVER_FAILED","OCSP_STATUS_UNKNOWN","OCSP_REASON_UNSPECIFIED","OCSP_REASON_KEY_COMPROMISE","OCSP_REASON_CA_COMPROMISE","OCSP_REASON_AFFILIATION_CHANGED","OCSP_REASON_SUPERSEDED","OCSP_REASON_CESSATION_OF_OPERATION","OCSP_REASON_CERTIFICATE_HOLD","OCSP_REASON_REMOVE_FROM_CRL","OCSP_REASON_PRIVILEGE_WITHDRAWN","OCSP_REASON_AA_COMPROMISE","url","CookieJar","cookieJar","file","get","head","post","put","patch","del","options","request","asyncRequest","batch","setResponseCallback","expectedStatuses","default"]
```
With v0.53 onwards, it's either an empty array (regular require), or this ($default require):
```
TypeError: Cannot convert undefined or null to object
        at keys (native)
```
If I perform the same from TypeScript code, everything's fine with the new version.

[...]

PR #4058, opened 2024-11-14
Fix namespaced object export when default is the only one available

ankur22:
## What?

This fixes an issue where a module only expose a default object, but we also want to allow the the module to be imported as a namespaced object. 

## Why?

This is how node works, and this is how we have chosen to work even after the changes to work with ESM from commonJS. It prevents breaking changes with older scripts that were not intended.

## Checklist

## Related PR(s)/Issue(s)

Related: https://github.com/grafana/k6/issues/4050

### DIFF
--- a/js/module_loading_test.go
+++ b/js/module_loading_test.go
@@ -626,6 +626,9 @@ func TestStarImport(t *testing.T) {
 	r1, err := getSimpleRunner(t, "/script.js", `
 		import * as cjs from "./commonjs_file.js"; // commonjs
 		import * as k6 from "k6"; // "new" go module
+		// go module that only exports default object, but we want it to
+		// export as a namespace object.
+		import * as http from "k6/http";
 		// TODO: test with basic go module maybe
 
 		if (cjs.something != 5) {
@@ -633,6 +636,9 @@ func TestStarImport(t *testing.T) {
 		}
 		if (typeof k6.sleep != "function") {
 			throw "k6.sleep has wrong type" + typeof k6.sleep;
+		}
+		if (typeof http.get != "function") {
+			throw "http.get has wrong type" + typeof http.get;
 		}
 		export default () => {}
 	`, fs, lib.RuntimeOptions{CompatibilityMode: null.StringFrom("extended")})
--- a/js/module_loading_test.go
+++ b/js/module_loading_test.go
@@ -617,36 +617,40 @@ func TestDefaultNamedExports(t *testing.T) {
 
 func TestStarImport(t *testing.T) {
 	t.Parallel()
-	fs := fsext.NewMemMapFs()
-	err := writeToFs(fs, map[string]any{
-		"/commonjs_file.js": `exports.something = 5;`,
+
+	t.Run("esm_spec", func(t *testing.T) {
+		t.Parallel()
+		fs := fsext.NewMemMapFs()
+		err := writeToFs(fs, map[string]any{
+			"/commonjs_file.js": `exports.something = 5;`,
+		})
+		require.NoError(t, err)
+
+		r1, err := getSimpleRunner(t, "/script.js", `
+			import * as cjs from "./commonjs_file.js"; // commonjs
+			import * as k6 from "k6"; // "new" go module
+			// go module that only exports default object, but we want it to
+			// export as a namespace object.
+			import * as http from "k6/http";
+			// TODO: test with basic go module maybe
+	
+			if (cjs.something != 5) {
+				throw "cjs.something has wrong value" + cjs.something;
+			}
+			if (typeof k6.sleep != "function") {
+				throw "k6.sleep has wrong type" + typeof k6.sleep;
+			}
+			if (typeof http.get != "function") {
+				throw "http.get has wrong type" + typeof http.get;
+			}
+			export default () => {}
+		`, fs, lib.RuntimeOptions{CompatibilityMode: null.StringFrom("extended")})
+		require.NoError(t, err)
+
+		arc := r1.MakeArchive()
+		_, err = getSimpleArchiveRunner(t, arc)
+		require.NoError(t, err)
 	})
-	require.NoError(t, err)
-
-	r1, err := getSimpleRunner(t, "/script.js", `
-		import * as cjs from "./commonjs_file.js"; // commonjs
-		import * as k6 from "k6"; // "new" go module
-		// go module that only exports default object, but we want it to
-		// export as a namespace object.
-		import * as http from "k6/http";
-		// TODO: test with basic go module maybe
-
-		if (cjs.something != 5) {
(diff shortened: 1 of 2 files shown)


js/modules/gomodule.go
```diff
@@ -25,10 +25,22 @@ func (gm *goModule) Instantiate(rt *sobek.Runtime) (sobek.CyclicModuleInstance,
 	if gm.exportedNames == nil {
 		named := mi.Exports().Named
 
-		gm.exportedNames = make([]string, len(named))
-		for name := range named {
-			gm.exportedNames = append(gm.exportedNames, name)
+		switch {
+		case len(named) == 0 && mi.Exports().Default != nil:
```
mstoykov, line 29: I really do prefer using `len` as well instead of checking for `nil`. But in the other place we have this logic `toESModuleExports` it is specifically checking for `nil` so that if you do not want to use this functionality you can have `Named` be an empty list.

I very much doubt this is used, but will prefer to keep it consistent

js/module_loading_test.go
```diff
@@ -634,6 +637,9 @@ func TestStarImport(t *testing.T) {
 		if (typeof k6.sleep != "function") {
 			throw "k6.sleep has wrong type" + typeof k6.sleep;
 		}
+		if (typeof http.get != "function") {
+			throw "http.get has wrong type" + typeof http.get;
+		}
```
mstoykov, line 642: Maybe import the default export as well and check that they have the same keys?

That will also help with checking if we do not export too much.

js/modules/gomodule.go
```diff
@@ -25,10 +25,22 @@ func (gm *goModule) Instantiate(rt *sobek.Runtime) (sobek.CyclicModuleInstance,
...
+			// If named length is 0 but default is defined, then try to work with
+			// default and extract the names of the object's properties. This
+			// behavior isn't ESM compatible, but we do want to allow defaults to
+			// be imported as namespaced object, which is also how node works.
+			if obj, ok := mi.Exports().Default.(*sobek.Object); ok {
+				gm.exportedNames = obj.GetOwnPropertyNames()
```
mstoykov, line 35: I would use `ToObject` given that a user can just make `ToValue(anygoobject)` and IIRC this will not lead to a `sobek.Object` underneed on its own.

js/modules/gomodule.go
```diff
@@ -25,10 +25,22 @@ func (gm *goModule) Instantiate(rt *sobek.Runtime) (sobek.CyclicModuleInstance,
...
+			}
+		default:
+			gm.exportedNames = make([]string, len(named))
```
mstoykov, line 38: I guess this is from the original code, but it is ... wrong :sweat_smile: As it first make a slice with the required size and then add more elements to the end

https://go.dev/play/p/W66MinRKLc3

This is doubtful to be an issue, as ... well empty strings aren't going to be useful but still
```suggestion
			gm.exportedNames = make([]string, 0, len(named))
```

js/modules/gomodule.go
```diff
@@ -25,10 +25,22 @@ func (gm *goModule) Instantiate(rt *sobek.Runtime) (sobek.CyclicModuleInstance,
 	if gm.exportedNames == nil {
 		named := mi.Exports().Named
 
-		gm.exportedNames = make([]string, len(named))
-		for name := range named {
-			gm.exportedNames = append(gm.exportedNames, name)
+		switch {
```
mstoykov, line 28: I expect you used `switch` as there were more than one cases earlier?

mstoykov (changes requested):
LGTM in general - i have a couple of comments mostly on old issues

js/modules/gomodule.go
ankur22: Sure, i've made the change in 241af938bc40e49c94787e05d10dd41958a7ecb0

js/modules/gomodule.go
ankur22: Fixed in 134399b4273ecd631c6fa41bc0a91cec1623c827

js/modules/gomodule.go
ankur22: I refactored it since i think it's easier to follow when there are more than one condition int he same function, instead of a if/else/else if etc. I also believe it's more go like to work with switch instead of if/else/else if.

js/modules/gomodule.go
ankur22: I don't follow you.

So a user could have assigned `Default` with `ToValue(anygoobject)`, therefore the cast to `*sobek.Object` wouldn't work. So if we convert `Default` to a Sobek object first, then it would be able to work?

mstoykov: ToValue is not required to return a *sobek.Object - it returns a Value. It likely does for practically all cases we have or extensions will use and I am pretty sure we just set `*sobek.Object` for most cases within k6 either way.

But still the correct way to get an *Object form Value is to use the [ToObject](https://pkg.go.dev/github.com/grafana/sobek#Value) funciton on the Value not to do direct cast.

ankur22: Ok, i understand the part about sobek `Value` 👍 

So since `Default` is an `interface{}`, I'd first need to convert it to a value with `rt.ToValue(Default)`, and then convert that to an object with `ToObject`.

I made the change in 296f8c286baa940621685b0a606536b955fc3792

js/module_loading_test.go
ankur22: Sure, what do you think of this change: f492b4033a07d65ea3d910498a4f7b948d5273c7

js/modules/gomodule.go
mstoykov: I do consider it go like in the cases where:
1. it can't be rewritten as a signel if/else pair - this can
2. or when using a case can be used to check a single variable against multiple values and that is the only thing happening:

```go
s := "apple";
switch s{
case "cucumber", "pepper", "leek":
  return "vegetable";
default:
  return "not vegetable"
}
```

As in both of this case it makes it easier to read. I would argue that using a switch in other cases makes me think there is something strange going on.

I will let other people also look at this and see what they think on it - so this is not a request for a change at this point, but a nit.

mstoykov (dismissed):

js/modules/gomodule.go
ankur22: Yeah, happy to hear more feedback, before merging/reverting the switch change.

inancgumus (dismissed):
LGTM 👍 Nice fix.

js/modules/gomodule.go
ankur22: I've reverted it back to if/else in 37a78c6cd139d3945244f8377e9278c4f33db8df

mstoykov (approved):

inancgumus (approved):

PR #4072, opened 2024-11-18
Use the precompiled 'gotip' on the xk6 workflow

joanlopez:
## What?

It updates the `xk6` workflow to use the [precompiled gotip](https://github.com/grafana/gotip), as we already do for the `Test` workflow.

## Why?

Last week I realized that we aren't using it because the CI workflow failed due to *'https://go.googlesource.com/go/'* returning `503 Service Unavailable` (see [here](https://github.com/grafana/k6/actions/runs/11842018792/job/32999653106)).

This isn't a "perfect" solution for that situation, because I guess the GitHub APIs (Assets, in this case) could eventually go down, as the site where Google hosts code was in that case, but it speeds up the workflow and reuses what we're already using in other workflows as `Test` or [k6-ci ](https://github.com/grafana/k6-ci), for the sake of consistency.

## Checklist


.github/workflows/xk6.yml
```diff
@@ -17,20 +17,26 @@ jobs:
       matrix:
         go: [stable, tip]
         platform: [ubuntu-latest, windows-2019, macos-latest]
-    runs-on: ubuntu-latest
```
joanlopez, line 20: Looks like we weren't using `matrix.platform`, so actually all the jobs were being executed on the same platform 😅 

I realized it, because when I tried to use the precompiled Go tip, it complained because the binary wasn't in the expected platform (e.g. a binary built for Windows, trying to be executed on Ubuntu, in this case haha)

.github/workflows/xk6.yml
```diff
@@ -54,6 +60,13 @@ jobs:
                "${{ github.event.pull_request.head.repo.full_name }}" != "${{ github.repository }}" ]; then
             export XK6_K6_REPO="github.com/${{ github.event.pull_request.head.repo.full_name }}"
           fi
+          # The following is a workaround for Windows, cause when using 'shell: bash', the PATH is expressed
```
joanlopez, line 63: Something I learned along the way, while working on this PR 💡

oleiade (approved):
🚀

.github/workflows/xk6.yml
olegbespalov: Good catch!

olegbespalov (approved):

PR #4082, opened 2024-11-29
Add test for race in PushIfNotDone

mstoykov:
## Checklist

### DIFF
--- a/metrics/sample.go
+++ b/metrics/sample.go
@@ -132,8 +132,12 @@ func PushIfNotDone(ctx context.Context, output chan<- SampleContainer, sample Sa
 	if ctx.Err() != nil {
 		return false
 	}
-	output <- sample
-	return true
+	select {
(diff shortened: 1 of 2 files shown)


oleiade (approved):
Nice catch! 🚀

olegbespalov (changes requested):
Waiting for actual fix, or closing the PR

ankur22 (approved):
LGTM 🚀

mstoykov:
This was a PR to illustrate that there is no way for the current PushIfNotDOne to not race with the context done and channel close.

The thing is that in the current code there is no place where the channel gets closed before every VU has returned so nothing should be using this - which is signaled with the context close.

The "real" fix for this will be to not use close the channel, but that is already used in some places, so it is way more involved.

Again in practice this should never happen - if it does it is a bug in the other code.

PR #4030, opened 2024-10-30
Warn on usage of open that will break in the future

mstoykov:
## What?

Warn on usage of open that will break in the future

TODO: do it for `grpc.Client#load`  and `fs.open()` as well.

This is kind of a draft at this point to get some of the code inline.

## Why?

Align open like functions and align them with how module imports work 

see #3857 

## Checklist

- [ ] I have performed a self-review of my code.
- [ ] I have added tests for my changes.
- [ ] I have run linter locally (`make lint`) and all checks pass.
- [ ] I have run tests locally (`make tests`) and all tests pass.
- [ ] I have commented on my code, particularly in hard-to-understand areas.

PR #4084, opened 2024-12-02
Report usages of the deprecated 'ext.loadimpact' option

joanlopez:
## What?

It adds a new `deprecated_options` field into the usage reports, and accounts if `ext.loadimpact` is still being used, and how often.

## Why?

Because we likely will remove that option parameter before v1, so we need to make some plans.

## Checklist

## Related PR(s)/Issue(s)

Related with https://github.com/grafana/k6-cloud/issues/2210

### DIFF
--- a/cmd/test_load.go
+++ b/cmd/test_load.go
@@ -273,7 +273,7 @@ func (lct *loadedAndConfiguredTest) buildTestRunState(
 	// Here, where we get the consolidated options, is where we check if any
 	// of the deprecated options is being used, and we report it.
 	if _, isPresent := configToReinject.External["loadimpact"]; isPresent {
-		if err := lct.preInitState.Usage.Strings("deprecated_options", "ext.loadimpact"); err != nil {
+		if err := lct.preInitState.Usage.Uint64("deprecations/options.ext.loadimpact", 1); err != nil {
 			return nil, err
 		}
 	}
--- a/cmd/test_load.go
+++ b/cmd/test_load.go
@@ -265,9 +265,17 @@ func loadSystemCertPool(logger logrus.FieldLogger) {
 func (lct *loadedAndConfiguredTest) buildTestRunState(
 	configToReinject lib.Options,
 ) (*lib.TestRunState, error) {
-	// This might be the full derived or just the consodlidated options
+	// This might be the full derived or just the consolidated options
 	if err := lct.initRunner.SetOptions(configToReinject); err != nil {
 		return nil, err
+	}
+
+	// Here, where we get the consolidated options, is where we check if any
+	// of the deprecated options is being used, and we report it.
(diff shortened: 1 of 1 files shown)


cmd/test_load.go
```diff
@@ -265,11 +265,19 @@ func loadSystemCertPool(logger logrus.FieldLogger) {
 func (lct *loadedAndConfiguredTest) buildTestRunState(
 	configToReinject lib.Options,
 ) (*lib.TestRunState, error) {
-	// This might be the full derived or just the consodlidated options
+	// This might be the full derived or just the consolidated options
 	if err := lct.initRunner.SetOptions(configToReinject); err != nil {
 		return nil, err
 	}
 
+	// Here, where we get the consolidated options, is where we check if any
```
joanlopez, line 273: I added it here because it looks like a considerably safe place among all the places configuration is read/written and manipulated in general, but any suggestion will be more than welcome.

joanlopez: Also, note that this will be reported despite it's a local (`k6 run`) or a remote execution (`k6 cloud`), but I guess it's fine because a script that holds `ext.loadimpact` is still a candidate of experiencing issues if we remove such support.

In any case, if we want to get real metrics, we can always account reports with this and a test run id.

olegbespalov (dismissed):

cmd/test_load.go
```diff
@@ -265,11 +265,19 @@ func loadSystemCertPool(logger logrus.FieldLogger) {
...
+	// of the deprecated options is being used, and we report it.
+	if _, isPresent := configToReinject.External["loadimpact"]; isPresent {
+		if err := lct.preInitState.Usage.Strings("deprecated_options", "ext.loadimpact"); err != nil {
```
mstoykov, line 276: In https://github.com/grafana/k6/pull/4030/files the key is `deprecations/<what is deprecated>`. I would prefer if we keep the keys more generic and `deprecated_options` is very specific to options by the looks of it and I do not know of any other option that will make it relevant to have a separate key.

mstoykov (dismissed):
LGTM, but would prefer if we change the key to be consistent

cmd/test_load.go
joanlopez: So, is your suggestion: `deprecations/ext.loadimpact`? 
I'm happy with that one as well, sure!

mstoykov: I guess that looks okay - yes. Don't know if we want `options` somewhere in there, but as a one of case this seems good enough.

olegbespalov: I've applied the feedback to get it mergable

mstoykov (approved):

ankur22 (approved):

Issue #3907, opened 2024-08-22
Make metrics.Metric be immutable


mstoykov:
### What?

Basically the TODO here:

https://github.com/grafana/k6/blob/55eb8bca07286ccf4f14e417e07de491f5912acc/metrics/metric.go#L11-L26

The current metric has a bunch of fields used by the ingester, summary and engine to:
1. aggregate metric samples for the summary and thresholds
2. evelaute thresholds and it subthresholds

None of those are essentials to hte metrics and the current users of the code need to know to not touch those values.

Arguably also those fields shouldn't be exported at all

### How?

Move all of those fields in the places where they are useful as well as the type definitions. Probably make them internal.

In the past users (and core developers) have used those to implement metric outputs among other things, and while it *might* work for some cases, it has also introduced complex bugs and have always fell short. As well that isn't the intended purpose.

PR #4089, opened 2024-12-04
Revamp the end-of-test summary

joanlopez:
## Overview

This pull request changes the current end-of-test summary by a new design, with two available formats: 
- a) compact *(default)*
- b) full
aiming to bring clearer a more valuable results to users. _Find a screenshot below._

### User-facing details

- 💹  The appearance of the end-of-test summary is now different, with thresholds at the beginning (instead on being inlined with metrics), checks slightly modified and metrics grouped by category (http, ws, network, etc). 
- 🔠 The user can choose between:
  - the `compact` summary with `--with-summary=compact`, or with no argument, as it is the default choice.
  - the `full` summary with `--with-summary=full`.
  - the `legacy` summary with `--with-summary=legacy`.
- ⚠️  The data model passed into the custom `handleSummary` function is now different. So, those users relying on it must migrate their implementation or use `--with-summary=legacy` meanwhile.
  - 🙏🏻  (_Dear reviewer_) If you definitely think we should keep the old format, I can give it a try and see if I can accommodate the new to it before calling the function, without many extra allocations (I guess that since we don't propagate sinks here it should be fine in general). But please, comment it explicitly, exposing the reasons and what's your concrete proposal.
 
### Technical details (_for review purposes_)

- The core logic of the new end of test summary, and how it collects metrics, etc is based on a new `output.Output` named `summary`.
- There's a new test script as example under `internal/cmd/testdata/summary/...` with different scenarios, groups, thresholds, custom metrics and what not... that can be used for both automated and manual testing. If you think anything is missing, just suggest it.
- The JS code responsible for rendering the summary have been largely refactored and type-documented, 👏🏻  big shot-out to @oleiade, aiming to make that code way more maintainable that it has been until now. I guess that, once we merge this PR, we may need to copy-past it back to https://github.com/grafana/k6-jslib-summary.
- I left two data structures for the summary representation (`lib.Summary` vs `lib.LegacySummary`), to keep support for the legacy summary for some time, in an easy way, so things aren't complexity mixed and the the clean up in the future is simpler, just by removing that type, all the references to it, and simplifying the few conditionals that behave depending on which summary type is provided.
  - Similarly, I left the old JS code for the summary as the `summary-legacy.js`, for simpler cleanup whenever we remove that support, which I guess might be for v2 (once we ship the formalized JSON output format within v1).

#### Internal Checklist

**Before review readiness**

- [x] Add support for a compact summary mode (likely enabled by default, so perhaps a "extended" flag, to avoid memory allocation issues), that only "relays" metrics but not stores metrics for groups and scenarios.
- [x] Revisit sorting: is there anything else we can sort for this first iteration _(note that some ideas have been moved to a second iteration, see below)_
- [x] Take a decision on what do we want to do with the existing summary, whether we want to use the new `lib.Report` as the new API for custom `handleSummary` (note this would be a breaking change), or if we want to ship this progressively.
  - [x] Review and modify (if needed) the code in `js/summary.go` according to that decision.
- [x] Review and refactor the code in `js/summary.js`:
  - [x] All the pieces from the old summary can probably be removed, if not removed yet.
  - [x] De-duplicate the functions `summarizeMetrics` and `summarizeMetricsWithThresholds`, or just replace the first with the second one (as we may no longer need the first one if we remove the old summary code).
- [x] Review the structure and code of the `output/summary` package:
- [x] Verify all scenarios (different test scripts) work as expected (test with no groups nor scenarios, test with only groups, test with only scenarios, test with the combination of both, all of them with compact mode enabled or not).
- [x] Make the CI look 🟢 as an 🍏  again!
- [x] _(Optional)_ Define the JSDoc for the newly introduced functions in JS code.
- [x] Remove the `playground/full-summary` files, or define a proper location for them.
--- Ideas left for a second iteration---
- Re-write the JS code in TS.
- Sorting the thresholds by metric name, sorting the tags within the metric name and perhaps sorting the sources.
- Implementing a more optimized data structure (as explored in https://github.com/joanlopez/xk6-custosummary/tree/main/timeseries) to use less memory allocations while keeping scenarios & groups samples.

**General**


internal/js/runner.go
```diff
@@ -360,7 +362,7 @@ func (r *Runner) HandleSummary(ctx context.Context, summary *lib.Summary) (map[s
 		}
 	}()
 
-	summaryCtx, cancel := context.WithTimeout(ctx, r.getTimeoutFor(consts.HandleSummaryFn))
+	summaryCtx, cancel := context.WithTimeout(ctx, 10*time.Minute)
```
olegbespalov, line 365: Out of curriocity, why do we switch to a 10 mins timeout? :thinking:

lib/models.go
```diff
@@ -207,7 +207,7 @@ type Check struct {
 	Fails  int64 `json:"fails"`
 }
 
-// NewCheck creates a new check with the given name and parent group. The group may not be nil.
+// NewCheck creates a new check with the given name and parent group. The group must not be nil.
```
olegbespalov, line 210: Good catch! Keeping in mind comment, maybe add a validation to constructor with explicit error? :thinking:

olegbespalov (commented):
I did a pass on that, but need a couple of more for sure, commenting just signaling that I'm on it

internal/js/runner.go
joanlopez: Likely something I modified while debugging, that I forgot to set back, so great catch! 👍

joanlopez:
> I did a pass on that, but need a couple of more for sure, commenting just signaling that I'm on it

Sure, thanks! Take your time!

olegbespalov:
>full aiming to bring clearer a more valuable results to users. Find a screenshot below.

Where can I find the screenshot for that? :thinking: 

I've tried to run in two modes and not sure if I see any difference

image

internal/js/summary.go
```diff
@@ -159,3 +160,10 @@ func getSummaryResult(rawResult sobek.Value) (map[string]io.Reader, error) {
 
 	return results, nil
 }
+
+func calculateCounterRate(count float64, duration time.Duration) float64 {
```
olegbespalov, line 164: Sounds like it should be one of the methods of the `*metrics.CounterSink`, if I'm not mistaken, we duplicate it two more times

internal/lib/testutils/minirunner/minirunner.go
```diff
@@ -24,7 +24,7 @@ type MiniRunner struct {
 	Fn              func(ctx context.Context, state *lib.State, out chan<- metrics.SampleContainer) error
 	SetupFn         func(ctx context.Context, out chan<- metrics.SampleContainer) ([]byte, error)
 	TeardownFn      func(ctx context.Context, out chan<- metrics.SampleContainer) error
-	HandleSummaryFn func(context.Context, *lib.Summary) (map[string]io.Reader, error)
+	HandleSummaryFn func(context.Context, *lib.LegacySummary) (map[string]io.Reader, error)
```
olegbespalov, line 27: Shouldn't we also use the new signature here? :thinking:

lib/summary.go
```diff
@@ -0,0 +1,298 @@
...
+	SummaryGroup
+	Scenarios map[string]SummaryGroup
+
+	TestRunDuration time.Duration // TODO: use lib.ExecutionState-based interface instead?
+	NoColor         bool          // TODO: drop this when noColor is part of the (runtime) options
+	UIState         UIState
+}
+
+// NewSummary instantiates a new empty Summary.
+func NewSummary() *Summary {
```
olegbespalov, line 106: Out of the curiosity, why have we decided to use the empty summary constructor? I mean, why don't we ask require the mandatory values throw the constructor (and maybe even apply validation)?

output/summary/data.go
```diff
@@ -0,0 +1,355 @@
...
+		if err != nil {
+			panic(err) // This should never happen
+		}
+		a.checks[name] = check
+		a.orderedChecks = append(a.orderedChecks, check)
+	}
+	return check
+}
+
+func populateSummaryGroup(
```
olegbespalov, line 158: Even it's an internal, unexported method, I feel that it also maybe good to add some unit tests since the beginning, which maintain the classification and will help perhaps to cover some future cases.

internal/cmd/run.go
```diff
@@ -189,26 +190,76 @@ func (c *cmdRun) run(cmd *cobra.Command, args []string) (err error) {
...
-				TestRunDuration: executionState.GetCurrentTestRunDuration(),
-				NoColor:         c.gs.Flags.NoColor,
-				UIState: lib.UIState{
-					IsStdOutTTY: c.gs.Stdout.IsTTY,
-					IsStdErrTTY: c.gs.Stderr.IsTTY,
-				},
+	if !testRunState.RuntimeOptions.NoSummary.Bool { //nolint:nestif
+		sm, err := lib.ValidateSummaryMode(testRunState.RuntimeOptions.SummaryMode.String)
+		if err != nil {
+			logger.WithError(err).Error("invalid summary mode, falling back to \"compact\" (default)")
```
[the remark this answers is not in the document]
joanlopez: Yes, great idea! 👍🏻

internal/cmd/testdata/summary/browser.js
```diff
@@ -0,0 +1,12 @@
+import {browser} from 'k6/browser'
+
+export async function browserTest() {
+	const page = await browser.newPage()
+
+	try {
+		await page.goto('https://test.k6.io/')
```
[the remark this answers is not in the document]
joanlopez: Sure, good catch!

internal/js/summary.go
joanlopez: Yeah, good recommendation! Thanks!

lib/summary.go
joanlopez: I don't think I have a concrete answer, to be honest. I think the reason why I followed this approach is because it's mostly a DTO, and concretely a recursive one, so it felt easier to initialize it empty, like when you initialize a `map`,  and populate it on the go, instead of asking for the inner data in the constructor.

Most of the logic is on the `summary.Output` side, but I preferred not to couple both, even if that's the main use, at least for now.

output/summary/data.go
joanlopez: What do you think if I add a couple of tests for the:

```
func (o *Output) Summary(
	executionState *lib.ExecutionState,
	observedMetrics map[string]*metrics.Metric,
	options lib.Options,
) *lib.Summary
```

function that's public and relies on this one under the hood?

Perhaps even better? 🤔

internal/lib/testutils/minirunner/minirunner.go
joanlopez: I think the reason why I didn't do it is because I think it is used nowhere, but yeah, it's a good idea overall.

output/summary/data.go
olegbespalov: yep, testing of the Public API if that is possible for sure better, great idea :+1:

joanlopez: Added a couple of tests in https://github.com/grafana/k6/pull/4089/commits/58e3d021ea7509bfa7aba35b0babefb7c54e04cc

internal/lib/testutils/minirunner/minirunner.go
joanlopez: Done @ https://github.com/grafana/k6/pull/4089/commits/58e3d021ea7509bfa7aba35b0babefb7c54e04cc

internal/js/summary.go
joanlopez: Done @ https://github.com/grafana/k6/pull/4089/commits/58e3d021ea7509bfa7aba35b0babefb7c54e04cc

internal/cmd/testdata/summary/browser.js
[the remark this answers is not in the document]
joanlopez: Done @ https://github.com/grafana/k6/pull/4089/commits/58e3d021ea7509bfa7aba35b0babefb7c54e04cc

internal/cmd/run.go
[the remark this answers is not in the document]
joanlopez: Done @ https://github.com/grafana/k6/pull/4089/commits/58e3d021ea7509bfa7aba35b0babefb7c54e04cc

output/summary/summary.go
```diff
@@ -0,0 +1,195 @@
...
+	output.SampleBuffer
+
+	periodicFlusher *output.PeriodicFlusher
+	logger          logrus.FieldLogger
+
+	dataModel   dataModel
+	summaryMode lib.SummaryMode
+}
+
+// New returns a new JSON output.
```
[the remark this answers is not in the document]
joanlopez: Done @ https://github.com/grafana/k6/pull/4089/commits/58e3d021ea7509bfa7aba35b0babefb7c54e04cc

joanlopez:
> > full aiming to bring clearer a more valuable results to users. Find a screenshot below.
> 
> Where can I find the screenshot for that? 🤔
> 
> I've tried to run in two modes and not sure if I see any difference
> 
> image

The key difference between `full` and `compact` modes is that the former also displays partial results for groups and scenarios, while the latter only displays total results. However, if there are no groups nor scenarios, then their appearance is the same.

Do you have any other suggestion to differentiate among them? 
cc/ @oleiade do you have any other idea? Perhaps hiding some data? 🤔 

To be fully transparent, I don't have a strong opinion here, but I'd advocate to either make any change w'all fully agree, or move forward as-is, to avoid looping in cycles. As far as I know, we offer no guarantee on the text summary format, so it should be fine to iterate it in the near future if needed. The one shipped as part of this PR doesn't need to be the definitive one before and along the v1.

lib/models.go
joanlopez: Done @ https://github.com/grafana/k6/pull/4089/commits/58e3d021ea7509bfa7aba35b0babefb7c54e04cc

internal/js/runner.go
joanlopez: Done @ https://github.com/grafana/k6/pull/4089/commits/58e3d021ea7509bfa7aba35b0babefb7c54e04cc

olegbespalov:
@joanlopez like I said internally, I'm totally fine with processing as it's with one exception that it's worth adjusting texting when it lands documentation

>full - aiming to bring clearer a more valuable results to users. Find a screenshot below.

IMO too generic and vague and in my case it was the source of confusion since I expected to see these more valuable results since the beginning

joanlopez:
> @joanlopez like I said internally, I'm totally fine with processing as it's with one exception that it's worth adjusting texting when it lands documentation
> 
> > full - aiming to bring clearer a more valuable results to users. Find a screenshot below.
> 
> IMO too generic and vague and in my case it was the source of confusion since I expected to see these more valuable results since the beginning

Nice, thanks for your input @olegbespalov!
Let's wait for @oleiade's input, but yeah, I'll definitely take that into consideration when writing the corresponding docs.

oleiade:
Hey @joanlopez @olegbespalov 👋🏻 

Apologies for the delay and being blocking here. I overlooked the compact vs full list of metrics during the last phases of the design, but doing a bit of digging into some of our initial design docs, I found back that we had come up with a candidate list of metrics to exclude in compact mode (and include in full/extended mode):

> We exclude the following metrics from default results:
> http_req_blocked
> http_req_connecting
> http_req_receiving
> http_req_sending
> http_req_tls_handshaking
> http_req_waiting

The rationale there, being that we wanted to only show what is relevant to the vast majority of users in compact mode, and only bring them back when users explicit request the full mode. In general I remember that we discussed focusing on removing things that would only be relevant to a very small portion of users by default, or to very specific use-cases.

At the time we outlined only HTTP metrics, because that's where we thought the most of those cases were, but I think we should feel free to expand that list if we can consider other metrics as "not absolutely mandatory".

I also agree with @olegbespalov in that we are explicitly excluding the end of test summary from the v1.Y.Z support policy, and we should feel free to iterate on it in the future. So we don't have to block on this if there are diverging ideas about what the list of included/excluded metrics would be for instance: even if we kept the list at what it is now, that would be 👍🏻 for me 🙇🏻 

Hope that's helpful, and again, great work @joanlopez I love it ❤️ 🚀

olegbespalov (dismissed):
LGTM! :rocket:

js/.eslintignore
```diff
@@ -0,0 +1,5 @@
+# Ignore everything in this folder
+*
+
+# But the end-of-test summary file
+!summary.js
```
mstoykov, line 5: I am not certain if this was intended to be commited or not , but all of those configurations for eslint and prettier IMO either should
1. not exist
2. be project wide

This currently seems a bit ad hoc and I am not cetain how common it will be for those to be updated.

The original idea with the js summary code is that it lives in jslib and k6 just gets the artefact and puts it here.

This allows us to skip having this checks in k6 repo and have it one place where it can be used or combined with others by users in scripts.

joanlopez: @oleiade I think it's you who added it. Any preference? 

I guess the reason here is that DX is a bit worse, because you normally use both in combination (k6+summary), even if we distribute the summary code as a jslib. But it's also true that in theory we shouldn't be changing it very often, maybe it's just fine to keep everything in the jslib repo and just copy files back and forth when needed.

internal/js/summary.js
```diff
@@ -1,413 +1,1392 @@
-var forEach = function (obj, callback) {
-  for (var key in obj) {
-    if (obj.hasOwnProperty(key)) {
-      if (callback(key, obj[key])) {
-        break
-      }
-    }
-  }
+/**
```
mstoykov, line 1: for code reviews reason it would've been easier if the files weren't switched yet but done in a separate PR, as now I do not know if the other code was changed and how much and this diff here is pointless

output/summary/data.go
```diff
@@ -0,0 +1,367 @@
...
+		case isBrowserMetric(info.Name):
+			dest.Browser[info.Name] = summaryMetric
+		case isGrpcMetric(info.Name):
+			dest.Grpc[info.Name] = summaryMetric
+		case isWebSocketsMetric(info.Name):
+			dest.WebSocket[info.Name] = summaryMetric
+		case isWebVitalsMetric(info.Name):
+			dest.WebVitals[info.Name] = summaryMetric
+		default:
+			dest.Custom[info.Name] = summaryMetric
```
mstoykov, line 196: This makes the summary very much coupled with what modules are in core and what are their metrics in a way that will mean that this will need to be updated on us adding stuff and anything outside of the core mdoules will go to custom. 

So in the future when/if we have more and more modules as extensions or even move some out of here this will either have to stop working or will need some different way to make the same thing.

Have there been other approaches that have been considered?

output/summary/data.go
```diff
@@ -0,0 +1,367 @@
...
+
+	for _, metricData := range groupData.aggregatedMetrics {
+		storeMetric(
+			summaryGroup.Metrics,
+			lib.SummaryMetricInfo{
+				Name:     metricData.Metric.Name,
+				Type:     metricData.Metric.Type.String(),
+				Contains: metricData.Metric.Contains.String(),
+			},
+			metricData.Sink,
```
mstoykov, line 208: I am not particularly certain this should actually get the whole `Metric` instead of just some parts of it

output/summary/data.go
```diff
@@ -0,0 +1,367 @@
...
+	// FIXME (@joan): Drop this and replace it with a concrete copy of the metric data we want to track
+	// to avoid any potential confusion.
+	Metric *metrics.Metric
+
+	// FIXME (@joan): Introduce our own way of tracking thresholds, and whether they're crossed or not.
+	// Without relying on the internal submetrics the engine maintains specifically for thresholds.
+	// Thresholds []OurThreshold // { crossed: boolean }
+
+	Sink metrics.Sink
+}
```
mstoykov, line 123: @I would really prefer if you do not use metrics.Sink as it is to be removed hopefully before v1.0.0  https://github.com/grafana/k6/issues/3907

In the past it has been a case of many races as people do not realize there is no locking on this. They are also mostly designed to serve the current summary, thresholds and `k6 stats` all of which have slightly different requirements and make the code very clunky ... and arguably not great. https://github.com/grafana/k6/issues/2320

This also means that this summary still makes Trends collect each and every sample value that has been emitted which is a thing we could fix now, instead of adding another breaking change later. Maybe in a separate PR still if that is a problem. https://github.com/grafana/k6/issues/763 https://github.com/grafana/k6/pull/3213 for a possible fix for summary and https://github.com/grafana/k6/blob/master/output/cloud/expv2/hdr.go for the current used one in cloud output.

mstoykov (commented):
Still not done with the review, but would like to try to make it in parts so, posting this for now. 

Very happy with this being an output and still supporting handleSummary. 

I do have a bunch of comments that are around the reviewing of this, not certain whether we should work on them or just for future reference:
1. it would've been nice for some of the refactor stuff that are not related to have been separate as they now bump up the line numbers and github is not great at actually figureing out what changed.
2. I would've prefered if the old summary stayed in the same file and the new one had a new name. I did the same thing a few days ago with k6/timers PR and it definitely isn't great to have to figure out if changes were made or just renamed. Given the code changes here it is even harder.

My main problems are:
1. reusing metric.Sink - I am even okay with just copy pasting it in the output and usign it, but given this likely is to be (re)moved soon to make metrics immutable, I really would prefer not to have to make changes to this later on. This also is probably good chance to use hdr for Trend summary, doing now isntead of later and making another breaking change.
2. The current code depends on the summary knowing the exact names of metrics that it wants to group. This makes it beign coupled to the names and means we will have to keep editing it and probvably worse it will not work very well with extensions

internal/cmd/runtime_options.go
```diff
@@ -31,6 +31,8 @@ extended: base + sets "global" as alias for "globalThis"
 	flags.StringArrayP("env", "e", nil, "add/override environment variable with `VAR=value`")
 	flags.Bool("no-thresholds", false, "don't run thresholds")
 	flags.Bool("no-summary", false, "don't show the summary at the end of the test")
+	flags.String("with-summary", lib.SummaryModeCompact.String(), "determine the summary mode,"+
+		" \"compact\", \"full\" or \"legacy\"")
```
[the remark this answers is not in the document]
joanlopez: I think the initial suggestion was from @oleiade, but as far as I know it's not set in stone. 
I'm happy with `--summary-mode`, wdyt? 🤔

output/summary/data.go
joanlopez: We haven't considered any other approach for this iteration. And I'm not sure there's any without extending the current metrics API, to register new metrics with some sort of common/group identifier.

Do you have any particular idea? cc/ @oleiade

output/summary/data.go
joanlopez: Do you mean that `aggregatedMetric` may already hold the `lib.SummaryMetricInfo`, which is a subset of the data hold by the `*metrics.Metric`, and replace that instead? 

If so, I agree, and I'm happy to make that change. 
But, could you confirm it, please? If not, please can you clarify what do you mean? 🙏🏻

output/summary/doc.go
```diff
@@ -0,0 +1,4 @@
+/*
+Package summary implements an output that collects metrics to be displayed in the end-of-test summary
+*/
+package summary
```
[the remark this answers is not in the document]
joanlopez: Yeah, I don't see any reason why not. I think I didn't even consider it because when we did most of the work in this PR, `internal` wasn't really there. But now, it's definitely its appropriate place.

internal/cmd/runtime_options.go
```diff
@@ -41,94 +43,119 @@ extended: base + sets "global" as alias for "globalThis"
...
+		// some early validation
+		return opts, err
 	}
 
-	if envVar, ok := environment["K6_TRACES_OUTPUT"]; ok {
-		if !opts.TracesOutput.Valid {
-			opts.TracesOutput = null.StringFrom(envVar)
-		}
+	if envVar, ok := environment["K6_SUMMARY_EXPORT"]; !opts.SummaryExport.Valid && ok {
+		opts.SummaryExport = null.StringFrom(envVar)
```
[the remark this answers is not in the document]
joanlopez: I can try that now, yeah!

output/summary/doc.go
```diff
@@ -0,0 +1,4 @@
+/*
+Package summary implements an output that collects metrics to be displayed in the end-of-test summary
+*/
```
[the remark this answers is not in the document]
joanlopez: I think it's because unluckily I copied an example `doc.go` from another output that has the same style (`csv`). 
I can fix this one here, and create a small PR to fix a couple for that are wrong in a separate PR.

output/summary/doc.go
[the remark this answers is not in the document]
joanlopez: Done @ https://github.com/grafana/k6/pull/4089/commits/b4c0509040d0c1fdaaaa1dcfed38f0bfe906596e!

output/summary/data.go
joanlopez: Sorry, I don't fully get what's the suggestion here.

To "duplicate" the `metrics.Sink` values locally (in the summary) and use the HDR in that case to avoid that duplication being painful with `Trend` metrics? 

If so, I'm happy doing such change, as long as we're fine with the loss of precision or whatever implication HDR histograms may have in the summary results. 

And in such case, why don't we change `metrics.Sink` implementation at all? It's a part of the public API, but I guess that as struct values are private, we could keep compatibility on all the public methods while using HDR under the hood? 🤔

joanlopez: Re-reading your comment here:

> 1. reusing metric.Sink - I am even okay with just copy pasting it in the output and usign it, but given this likely is to be (re)moved soon to make metrics immutable, I really would prefer not to have to make changes to this later on. This also is probably good chance to use hdr for Trend summary, doing now isntead of later and making another breaking change.

It looks like you're suggesting precisely to re-implement `TrendSink` with HDR histograms? Do I read that correctly? I can do that in a separate PR, if that's what we want, yeah.

olegbespalov:
> The current code depends on the summary knowing the exact names of metrics that it wants to group. This makes it beign coupled to the names and means we will have to keep editing it and probvably worse it will not work very well with extensions

Honestly, now since Mihail brought it, I must share that I have a same feeling. That's why I was going to suggest putting the classification methods under the metric package, to at least show the future contributors that this classification exists and should be considered. However, not sure where _else_ it could be relevant :shrug: , so that's why I bought an existing approach.

joanlopez:
> That's why I was going to suggest putting the classification methods under the metric package, to at least show the future contributors that this classification exists and should be considered.

Can you expand a bit what do you mean with that, please? 🤔 

> However, not sure where _else_ it could be relevant 🤷 , so that's why I bought an existing approach.

I think that's the main reason why I haven't explored any other option (no one really convinced me).

olegbespalov:
> Can you expand a bit what do you mean with that, please? 🤔

I meant that these functions which I currently internal
https://github.com/grafana/k6/pull/4089/files#diff-f691646a2c46424ae9b5070111ec337287aa57f553d295587bc85697101e7c71R295-R336

could potentially be a part of the metric package (along with the metric categories) and if future contributors want to introduce new metrics, my assumption that they also could think that they need to care about metric category, but the summary logic will remain unchanged/or less affected to changes

>I think that's the main reason why I haven't explored any other option (no one really convinced me).

:+1: yes, I totally accept it, just wanted to share my perspective

js/.eslintignore
oleiade: Hey folks, this is indeed a legacy artifact left behind when I rewrote the `summary.js` file completely. During the refactor, I wanted to be sure to address most linting issues to catch potential pre-existing issues before I modify it. And I wanted to ensure the style was consistent across the whole file at the very least.

If we move the code in a JS code in a jslib, then I'd be completely fine with moving the configuration there indeed 👍🏻 

If we keep it in the repository, I would then prefer for us to have an eslint/prettier setup in some fashion still, project wide as suggested would be ideal for me, but I acknowledge there might be a lot of special cases/ignores to handle.

I have no strong opinion either way, but I would prefer to not have to copy-paste eslint/prettier configurations around whenever I need to touch the code ^^

internal/cmd/runtime_options.go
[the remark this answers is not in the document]
oleiade: It was the original suggestion indeed, but it is not set in stone indeed. I'm not completely hung upon `--with-summary`, but I proposed it because I like how it is close to natural language and somewhat guessable: "run k6 with its summary in full mode".

But like I said, I really don't have a strong opinion on this one, I'd be fine with `--sumary-mode` too. What I care about is that summary related options are consistent; thus if we go for `--summary-mode`, when we revamp the machine readable summary, we should make sure the option also keeps the `--summary-*` prefix 🙇🏻

output/summary/data.go
oleiade: I think at the time we looked into various options. But the only way that allowed to us to move forward and ship this in reasonable time was to do it this way, as just like you mention it @joanlopez, at least currently, we couldn't really think of another way to do it that wouldn't involve a big, scary, change.

To be fair, I also think the tradeoff there is limited, in the sense that everything being equal otherwise, we quite rarely add official metrics. Moving forward, I suggest we open an issue, and tackle it separately, in the future. But if there's another somewhat quick and achievable solution that we can get it in time for v1, let's go for it 👍🏻

mstoykov: Can you open the issue and please list the options that were big and scary. I also do not have a particular proposal, but it will be nice to have them out there. 

Arguably we do not add a lot of official metrics in an attempt to reduce the things we have to support forever. One of the ideas with v1 is that in v2, v3 and so on, we will be making breaking changes more easily without necessarily having a transition period for every breaking change.

So there is good chance that we will add a lot more metrics.

Also, even if we do not add new official metrics, this means that everybody using extensions more (another v1 goal) will get custom metrics for everything from those extensions.

I do not think we will be fixing this perfectly either way, but would still prefer to have at least an artifact of what the reasons were and what other things we considered.

output/summary/data.go
mstoykov: Yeah - basically `metrics.Metric` will get thinned out, so a lot of the stuff you currently don't use from it will be removed. So no real reason to copy stuff around IMO.

output/summary/data.go
mstoykov: Metrics.Sink is very likely going to stay in some capacity as internal implementation for the current summary and the thresholds and so on. 

The idea though is that it will not be part of the `metrics.Metric` struct and likely not part of the `metrics` package. 

Ideally it will be replaced with something more ergonimic for each use case it has, instead of trying to be useful in all contexts.

As such it is preferable if this new summary just has a new struct to actually keep the aggregated data that doesn't add to those cases. 

Moving to HDR for Trend will also be nice as we keep talking about how that will happen, but it will be a breaking change, so we keep postponing it. Now that we are doing one breaking change we can piggy back this in.

js/.eslintignore
joanlopez: Okay, so with the aim of moving forward, let's just remove these files from this PR. Once merged, I'll copy back all the changes (configurations included) to the specific repository. In the future, we can explore how to make this work more smoothly, especially if we really plan to do it more often 👍🏻 

Thanks for your inputs folks! 🙇🏻

joanlopez: Done @ https://github.com/grafana/k6/pull/4089/commits/f0b9306ec8acc98ad9bbc3c070a18e1c7a42a769

internal/cmd/runtime_options.go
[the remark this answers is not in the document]
joanlopez: Then, let's go with `--summary-mode`, which isn't only consistent with the `--summary-*` prefix but also with the `*-mode` suffix we use for compatibility, for instance. Not a deal breaker, but it's something more 🤷🏻 

Also, if we talk about grouping flags, I also prefer to group them by "feature" or "area", like "summary" in this case, than by the type of flag, for instance, like: all flags to customize something start with the `--with-*` prefix. 

The former I think it's a bit cleaner and favors discoverability.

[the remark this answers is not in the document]
joanlopez: Done @ https://github.com/grafana/k6/pull/4089/commits/5349b6535d224da348b6b2ce4f8b141653222b05

output/summary/data.go
joanlopez: @mstoykov is it okay for you if I move this discussion into an issue as you suggested, and consider this resolved in regards to this pull request? 🤔

Aiming to unblock things, but I'd be happy to consider that other issue, for instance, as blocking for v1.0 (not for the v1.0-rc tho). Wdyt?

joanlopez: I have created #4617, feel free to mark this conversation as resolved, and to move there! 🙇🏻

lib/summary.go
```diff
@@ -0,0 +1,291 @@
+package lib
```
mstoykov, line 1: Cn this be moved to internal package?

mstoykov (dismissed):
I have done one more pass, and given the time restraint (and the size of a bunch of the code) and that it is mostly internal - I am okay with merging this as is.

Please  do move `lib/summary.go` under  internal package  - maybe `internal/lib` just to get it out of the way - although having a `summary` package is probably a good idea.

For the record this was not used last I checked during the move packages that are not used in `internal` - just it is part of `lib` package which is imported a lot.

mstoykov (approved):
Do not forget to squash this :sweat_smile:

joanlopez:
> Do not forget to squash this 😅

Since we don't have a convention (or a strong one? 🤔), and personally I feel like what's atomic and what not is blurry, ambiguous and subjective, I generally squash all my PRs, so it would be very weird that specially for this one I misclick on "Merge",  but thanks for the reminder.

olegbespalov (approved):
LGTM! Great work! :rocket: 

Thinking loudly: it seems like the summary mode could also take responsibility of displaying no summary, which makes the UX more straightforward, but it's probably not a big deal

Issue #4041, opened 2024-11-07
Expose GCk6 test run ID via a JS API to the test script


andrewslotin:
Whenever a k6 test is executed in Grafana Cloud using `k6 cloud run` or if Grafana Cloud is set as an output with `k6 cloud run --local-execution`, a test run ID is generated for that execution. This identifier is helpful for distinguishing between local runs, such as debugging a load test script, and cloud executions. It also aids in correlating requests made during the test run.

Currently, there is no reliable way to retrieve this value from within the test script code. To address this, I propose modifying the `options.cloud` object returned by `k6/execution`, adding a field that will be populated with the test run ID for the current execution.

joanlopez:
Side-note; I'm not sure if now there's _"a reliable way to retrieve this value from within the"_ (_paraphrasing the sentence above_) Go code (e.g. an extension or output), but I'd really love if we make that possible as well as part of this task, as that will unlock/make possible other ideas we have on the backlog.

cc/ @Blinkuu this might be of your relevance, as this would be a blocker/requirement for the "Insights output" idea we discussed, as you need the Test Run ID to fetch the insights from the API.

cc/ @oleiade I see you assigned here, so ping me if you need further context.

mstoykov:
@joanlopez the original proposal is about putting it in `options.cloud` and practically every part of k6 has access to the options. Which is why I proposed it like that internally, instead of coming up with another specific API.

PR #4092, opened 2024-12-05
Populate __ENV.K6_CLOUDRUN_TEST_RUN_ID on local executions of k6 cloud run

joanlopez:
## What?

It populates the `__ENV.K6_CLOUDRUN_TEST_RUN_ID`, which is [the environment variable that contains the test run id on Cloud executions](https://grafana.com/docs/grafana-cloud/testing/k6/author-run/cloud-scripting-extras/cloud-execution-context-variables/), with the corresponding value for local executions streaming results to the Cloud: `k6 cloud run --local-execution`.

## Why?

Because this unlocks different projects we have ongoing, currently blocked by the lack of a mechanism to read that variable on local executions streaming results to the Cloud.

## Checklist

## Related PR(s)/Issue(s)

Contributes to #4041


joanlopez:
Note that the approach followed here won't make `testRunId` available for those executions still using the deprecated command to run a local execution: `k6 run -o cloud`, mainly because I cannot think of any feasible way to do that other than one that looks along the lines of what I did in [`k6/cloud-v2`](https://github.com/grafana/k6/compare/master...k6/cloud-v2), and for which I know @mstoykov have objections to.

But I'd be very happy to see any suggestions or ideas for possible workarounds. 
Although, I'm also happy with using that as mechanism to push those still using the deprecated command.

js/modules/k6/cloud/cloud.go
```diff
@@ -0,0 +1,110 @@
...
+
+// NewModuleInstance implements the modules.Module interface to return
+// a new instance for each VU.
+func (*RootModule) NewModuleInstance(vu modules.VU) modules.Instance {
+	mi := &ModuleInstance{vu: vu}
+
+	rt := vu.Runtime()
+
+	mi.obj = rt.NewObject()
+	defProp := func(name string, getter func() (sobek.Value, error)) {
```
oleiade, line 48: Genuine curiosity, is there an advantage to defining a closure for this operation considering it's used only once? Or is it to scope the logic in the code and make it easier to maintain: which I'm completely okay with.

js/modules/k6/cloud/cloud.go
```diff
@@ -0,0 +1,110 @@
...
+// testRunId returns a sobek.Value(string) with the Cloud test run id.
+//
+// This code can be executed in two situations, either when the k6 binary is running in the Cloud, in which case
+// the value of the test run id would be available in the environment, and we would have loaded at module initialization
+// time; or when the k6 binary is running locally and test run id is present in the options, which we try to read at
+// time of running this method, but only once for the whole execution as options won't change anymore.
+func (mi *ModuleInstance) testRunId() (sobek.Value, error) {
+	// In case we have a value (e.g. loaded from env), we return it.
+	// If we're in the init context (where we cannot read the options), we return undefined (the default value).
+	if !sobek.IsUndefined(mi.testRunID) || mi.vu.State() == nil {
```
oleiade, line 91: Would it make sens to use common.IsNullish here? Or do we want to specifically look only for undefined here?

js/modules/k6/cloud/cloud.go
```diff
@@ -0,0 +1,110 @@
...
+		return mi.testRunID, nil
+	}
+
+	// Otherwise, we try to read the test run id from options.
+	// We only try it once for the whole execution, as options won't change.
+	vuState := mi.vu.State()
```
oleiade, line 97: nit: as you already check whether mi.vu.State() is nil before, would it make sense moving this assignment at the top of the function? What do you think?

oleiade (dismissed):
I left a couple of nit comments behind, but none of them blocking, I'd be happy with this merged as-is. Great work!

cmd/cloud_run.go
```diff
@@ -131,6 +131,25 @@ func (c *cmdCloudRun) preRun(cmd *cobra.Command, args []string) error {
...
+		// As we've already loaded the test, we can modify the init function to
+		// reuse the initialized one.
+		c.runCmd.loadConfiguredTest = func(*cobra.Command, []string) (*loadedAndConfiguredTest, execution.Controller, error) {
+			return test, local.NewController(), nil
+		}
+
+		// After that, we can create the remote test run.
+		if err := createCloudTest(c.runCmd.gs, test); err != nil {
+			return fmt.Errorf("could not create the cloud test run: %w", err)
+		}
```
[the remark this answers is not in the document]
mstoykov: Also In order to get the env varible set in the script you need to set 

```
test.preInitState.RuntimeOptions.Env["something"] = value
```

In this case maybe this can be moved to createCloudTest?

cmd/outputs_cloud.go
```diff
@@ -0,0 +1,174 @@
...
+	conf.TestRunID = null.NewString(response.ReferenceID, true)
+
+	raw, err := cloudConfToRawMessage(conf)
+	if err != nil {
+		return fmt.Errorf("could not serialize cloud configuration: %w", err)
+	}
+
+	test.derivedConfig.Collectors[builtinOutputCloud.String()] = raw
+
+	return nil
```
mstoykov, line 141: Is this actually needed ? I guess maybe if we save it to disk, but not otherwise?

cmd/outputs_cloud.go
```diff
@@ -0,0 +1,174 @@
...
+	// If this is true, then it means that this code is being executed in the k6 Cloud.
+	// Therefore, we don't need to continue with the test run creation,
+	// as we don't need to create any test run.
+	//
+	// Precisely, the identifier of the test run is conf.TestRunID.
+	if conf.TestRunID.Valid {
+		return nil
+	}
+
+	// If not, we continue with some validations and the creation of the test run.
```
mstoykov, line 54: nit: from here onwards it seems like a thing that could be in cloudapi and be reused.

cloudapi/config.go
```diff
@@ -32,12 +32,17 @@ type Config struct {
...
+	// belong to the same test run/execution. Currently, it is equivalent to the TestRunID.
+	// But, in the future, or in future solutions (e.g. Synthetic Monitoring), there might be
+	// no TestRunID and we may still need an identifier to correlate all the things.
 	PushRefID null.String `json:"pushRefID" envconfig:"K6_CLOUD_PUSH_REF_ID"`
 
+	// TestRunID is the test run id, a unique identifier across all the test runs.
+	// It might be used to correlate all the things that belong to the same test run/execution,
+	// see PushRefID for more details.
+	TestRunID null.String `json:"testRunID" envconfig:"K6_CLOUD_TEST_RUN_ID"`
+
```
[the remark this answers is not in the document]
mstoykov: For the record due to how PushRefID is used in the cloud output, in order for the output to actually *stop* the test at the end you will need to set it's value, but have the PushRefID.Valid be false. 

See https://github.com/grafana/k6/blob/90bb9415d0724355e93eb276624d25394751d54d/output/cloud/output.go#L284

mstoykov:
@joanlopez maybe first rebase though as there have been a few changes including to the CI that mean theat you will have more lint failures.

cloudapi/config.go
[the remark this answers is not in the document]
joanlopez: Thanks for the information! But yeah, as you suggested, I'm going to remove it. If we're going to move forward with the `K6_CLOUDRUN_TEST_RUN_ID` environment variable, I think there's no other reason to add it.

cmd/outputs_cloud.go
```diff
@@ -0,0 +1,171 @@
...
+	if err != nil {
+		return err
+	}
+
+	if response.ConfigOverride != nil {
+		logger.WithFields(logrus.Fields{"override": response.ConfigOverride}).Debug("overriding config options")
+		conf = conf.Apply(*response.ConfigOverride)
+	}
+
+	test.preInitState.RuntimeOptions.Env[testRunIDKey] = response.ReferenceID
```
mstoykov, line 136: Please add an integration tests for this

cmd/outputs_cloud.go
```diff
@@ -0,0 +1,171 @@
...
+		logger, conf.Token.String, conf.Host.String, consts.Version, conf.Timeout.TimeDuration())
+
+	response, err := apiClient.CreateTestRun(testRun)
+	if err != nil {
+		return err
+	}
+
+	if response.ConfigOverride != nil {
+		logger.WithFields(logrus.Fields{"override": response.ConfigOverride}).Debug("overriding config options")
+		conf = conf.Apply(*response.ConfigOverride)
```
mstoykov, line 133: This lint error likely catches an issue.

I guess we might need to set this options specifically for the cloud output when it actually starts

joanlopez: Yeah, I think in this case (when there's cloud config overrides) we still need to previous code you suggested to remove, at least for now, as it is what the Cloud output will use 🤔

cmd/outputs_cloud.go
joanlopez: Not really sure about how to do this refactor in a way that it remains clearer for the future, I wrote you directly to discuss it in detail.

cmd/outputs_cloud.go
joanlopez: Done!

cmd/cloud_run.go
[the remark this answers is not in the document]
joanlopez: Done!

joanlopez:
Please @andrewslotin, @mstoykov, ~@oleiade~, @olegbespalov, note that these changes will make `__ENV.K6_CLOUDRUN_TEST_RUN_ID` to be available when using the new `k6 cloud run --local-execution`, and **NOT** when using the legacy (deprecated) `k6 run -o cloud`.

That's mainly because that would require a larger refactor, to completely move the creation of the test run outside of the Cloud output, as I did in https://github.com/grafana/k6/compare/master...k6/cloud, but that would mean "making the Cloud output even a more special output", and @mstoykov discouraged that. 

Otherwise, I don't think there's any way we can modify the environment from within the Cloud output particularly, and reflect that to the test's environment. But I'm open for suggestions/ideas, if any.

PS: @olegbespalov adding you as a reviewer, cause I think @oleiade won't be able to follow up on this PR for now.

andrewslotin:
@joanlopez, since `k6 run -o cloud` has been deprecated, I believe it's fair to not add new functionality there 👍

joanlopez:
> @joanlopez, since `k6 run -o cloud` has been deprecated, I believe it's fair to not add new functionality there 👍

I agree, I just wanted to make it clear so w'all set proper expectations.

cmd/outputs_cloud.go
```diff
@@ -0,0 +1,178 @@
...
+	apiClient := cloudapi.NewClient(
+		logger, conf.Token.String, conf.Host.String, consts.Version, conf.Timeout.TimeDuration())
+
+	response, err := apiClient.CreateTestRun(testRun)
+	if err != nil {
+		return err
+	}
+
+	// We store the test run id in the environment, so it can be used later.
+	test.preInitState.RuntimeOptions.Env[testRunIDKey] = response.ReferenceID
```
joanlopez, line 127: @mstoykov While checking if this would be enough, I realized that both:

- `output.Params.Environment`
- and `lib.TestPreInitState.LookupEnv`

rely on `gs.Env`. 

So, If you think it's better, for safety, just in case anybody tries to read that value from any of these, I'll be happy to also populate that environment variable onto `gs.Env`. 

But, for now, I decided to store it only where needed, so we don't start populating everything just because we have different mechanisms to query the environment for legacy reasons, and thus making a potential refactor/deprecation harder to reason about.

cmd/outputs_cloud.go
joanlopez: I'm lacking historical context here (about when and how those overrides are used), but in case there's overrides from Cloud, I guess we might want those to be applied and "sent" to the Cloud output 🤔 

If not, I'm happy with removing the entire `if` block and just ignore overrides from Cloud here.

cmd/outputs_cloud.go
mstoykov: Lets leave this for another PR

cmd/cloud_run.go
```diff
@@ -131,6 +131,27 @@ func (c *cmdCloudRun) preRun(cmd *cobra.Command, args []string) error {
...
+			}
+
+			// If the "K6_CLOUDRUN_TEST_RUN_ID" is set, then it means that this code is being executed in the k6 Cloud.
+			// Therefore, we don't need to continue with the test run creation, as we don't need to create any test run.
+			// This should technically never happen, as k6, when executed in the Cloud, it uses the standard "run"
+			// command "locally", but we add this early return just in case, for safety.
+			//
+			// If not, we know this execution requires a test run to be created in the Cloud.
+			// So, we create it as part of the process of loading and configuring the test.
+			if _, isSet := c.runCmd.gs.Env[testRunIDKey]; !isSet {
```
mstoykov, line 147: I think we should just drop this part for now and maybe add it later.

mstoykov (changes requested):
LGTM, just lets remoe the one if

cmd/cloud_run.go
olegbespalov: I do agree

output/cloud/output.go
```diff
@@ -17,11 +19,12 @@ import (
 	"go.k6.io/k6/output"
 	cloudv2 "go.k6.io/k6/output/cloud/expv2"
 	"go.k6.io/k6/usage"
-	"gopkg.in/guregu/null.v3"
 )
 
-// TestName is the default k6 Cloud test name
-const TestName = "k6 test"
```
olegbespalov, line 24: I believe it used in [k6-operator](https://github.com/grafana/k6-operator/blob/b5505372ca66fb028b6f1ac3c81d03316cfe504d/pkg/cloud/types.go#L41) so we need to coordinate with @yorugac

olegbespalov: Ah, no I am wrong here, please excuse the noise

cmd/outputs_cloud.go
joanlopez: Created https://github.com/grafana/k6/issues/4145

mstoykov (dismissed):

mstoykov (approved):

olegbespalov (approved):

PR #4093, opened 2024-12-05
Introduces --json flag for k6 version sub-command

olegbespalov:
## What?

This PR introduces the new `--json` flag for `k6 version` sub-command, which switches the output format to a JSON. 

Here are some examples:

Basic
```bash
./k6 version --json | jq
{
  "commit": "ec1f601e4f",
  "commit_dirty": true,
  "go_arch": "amd64",
  "go_os": "linux",
  "go_version": "go1.23.2",
  "version": "v0.55.0"
}
```
```bash
./k6 version
k6 v0.55.0 (commit/ec1f601e4f-dirty, go1.23.2, linux/amd64)
```
```bash
./k6 --version
k6 v0.55.0 (commit/ec1f601e4f-dirty, go1.23.2, linux/amd64)
```

With extensions:
```bash
./k6 version --json | jq
{
  "extensions": [
    {
      "imports": [
        "k6/x/sql"
      ],
      "module": "github.com/grafana/xk6-sql",
      "version": "v0.0.1"
    }
  ],
  "go_arch": "amd64",
  "go_os": "linux",
  "go_version": "go1.23.2",
  "version": "v0.55.0"
}
```

```bash
./k6 --version
k6 v0.55.0 (go1.23.2, linux/amd64)
Extensions:
  github.com/grafana/xk6-sql v0.0.1, k6/x/sql [js]
```
```bash
./k6 version
k6 v0.55.0 (go1.23.2, linux/amd64)
Extensions:
  github.com/grafana/xk6-sql v0.0.1, k6/x/sql [js]
```

## Why?

Having the ability to produce JSON structured output is essential if k6 used as part of some automation.

## Checklist

### DIFF
--- a/cmd/version.go
+++ b/cmd/version.go
@@ -154,13 +154,11 @@ func (c *versionCmd) run(cmd *cobra.Command, _ []string) error {
 		details["extensions"] = list
 	}
 
-	jsonDetails, err := json.Marshal(details)
-	if err != nil {
-		return fmt.Errorf("failed produce a JSON version details: %w", err)
+	if err := json.NewEncoder(c.gs.Stdout).Encode(details); err != nil {
+		return fmt.Errorf("failed to encode/output version details: %w", err)
 	}
 
-	_, err = fmt.Fprintln(c.gs.Stdout, string(jsonDetails))
-	return err
+	return nil
 }
 
 func getCmdVersion(gs *state.GlobalState) *cobra.Command {
--- a/cmd/version.go
+++ b/cmd/version.go
@@ -40,17 +40,33 @@ func (c *versionCmd) run(cmd *cobra.Command, _ []string) error {
 
 	details := consts.VersionDetails()
 	if exts := ext.GetAll(); len(exts) > 0 {
-		ext := make([]map[string]string, 0, len(exts))
-		for _, e := range exts {
-			ext = append(ext, map[string]string{
-				"name":    e.Name,
-				"type":    e.Type.String(),
-				"version": e.Version,
-				"path":    e.Path,
-			})
+		type extInfo struct {
+			Module  string   `json:"module"`
+			Version string   `json:"version"`
+			Imports []string `json:"imports"`
 		}
 
-		details["extensions"] = ext
+		ext := make(map[string]extInfo)
+		for _, e := range exts {
+			if v, ok := ext[e.Path]; ok {
+				v.Imports = append(v.Imports, e.Name)
+				ext[e.Path] = v
+				continue
+			}
+
+			ext[e.Path] = extInfo{
+				Module:  e.Path,
+				Version: e.Version,
+				Imports: []string{e.Name},
+			}
+		}
+
+		list := make([]extInfo, 0, len(ext))
+		for _, v := range ext {
+			list = append(list, v)
+		}
+
+		details["extensions"] = list
 	}
 
 	jsonDetails, err := json.Marshal(details)
--- a/cmd/version.go
+++ b/cmd/version.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"encoding/json"
 	"fmt"
 	"strings"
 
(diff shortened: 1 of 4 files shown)


szkiba:
I suggest using the extension registry terminology.

First of all, there is no such thing as a JavaScript or Output extension. An extension can register one or more import paths and one or more output names. That is why the registry has array properties of type "imports" and "outputs" for extensions.
Also, the extension does not have a name. There is a go module path, and there is (if there is one) a repository for the extension, and the repository has an owner and a name. However, the repository name is not unique. What is unique is the go module path.

I suggest taking the following properties from the registry:
- module
- imports
- outputs
I also suggest deleting the following properties:
- name
- type

olegbespalov:
>I suggest taking the following properties from the registry:

@szkiba do you mean to perform a network call to registry? :thinking:

szkiba:
> > I suggest taking the following properties from the registry:
> 
> @szkiba do you mean to perform a network call to registry? 🤔

No, sorry.  I mean use that property names and values

olegbespalov:
@szkiba wouldn't be hard for you modifying the output to reflect your request, in general I do agree that it make sense to follow registry terminology :handshake: 

```
"extensions": [
    {
      "name": "k6/x/sql",
      "path": "github.com/grafana/xk6-sql",
      "type": "js",
      "version": "v0.0.1"
    }
  ]
```

szkiba:
```json
"extensions": [
    {
      "module": "github.com/grafana/xk6-sql",
      "imports": ["k6/x/sql"]
      "version": "v0.0.1"
    }
  ]
```

like this

cmd/version.go
```diff
@@ -24,16 +25,70 @@ func versionString() string {
...
+		type extInfo struct {
+			Module  string   `json:"module"`
+			Version string   `json:"version"`
+			Imports []string `json:"imports"`
+		}
+
+		ext := make(map[string]extInfo)
+		for _, e := range exts {
+			if v, ok := ext[e.Path]; ok {
+				v.Imports = append(v.Imports, e.Name)
```
mstoykov, line 52: nitpick: I guess this is impossible, but still might be good idea to check th e version also matches

cmd/version.go
```diff
@@ -24,16 +25,70 @@ func versionString() string {
...
+		details["extensions"] = list
+	}
+
+	jsonDetails, err := json.Marshal(details)
+	if err != nil {
+		return fmt.Errorf("failed produce a JSON version details: %w", err)
+	}
+
+	_, err = fmt.Fprintln(c.gs.Stdout, string(jsonDetails))
+	return err
```
mstoykov, line 78:
```suggestion
  return json.NewEncoder(c.gs.Stdout).Encode(details)
```

cmd/version.go
```diff
@@ -24,16 +25,70 @@ func versionString() string {
...
+	cmd := &cobra.Command{
 		Use:   "version",
 		Short: "Show application version",
 		Long:  `Show the application version and exit.`,
-		Run: func(cmd *cobra.Command, _ []string) {
-			root := cmd.Root()
-			root.SetArgs([]string{"--version"})
-			_ = root.Execute()
-		},
+		RunE:  versionCmd.run,
```
mstoykov, line 88: I guess this means that now `k6 --version` and `k6 version` will output different things.

mstoykov (dismissed):
LGTM in general, but would really prefer if we keep making `lib/consts` not actually about constants and especialyl as this is only used in `cmd` it seems like a good idea to just move it there and unexport it

cmd/version.go
olegbespalov: > I guess this means that now k6 --version and k6 version will output different things.

That [behavior remains](https://github.com/grafana/k6/pull/4093/files#diff-7cb26babce207ed9c17e9dd3ead0b66d204da3cf4d11649412c197defbb3029cR34-R39) for a non JSON output

cmd/version.go
olegbespalov: We discussed it a bit with @szkiba, currently they all have the same version, but I could implement this check in advance

lib/consts/consts.go
```diff
@@ -11,14 +11,53 @@ import (
...
+const (
+	commitKey      = "commit"
+	commitDirtyKey = "commit_dirty"
+)
+
 // FullVersion returns the maximally full version and build information for
 // the currently running k6 executable.
 func FullVersion() string {
-	goVersionArch := fmt.Sprintf("%s, %s/%s", runtime.Version(), runtime.GOOS, runtime.GOARCH)
+	details := VersionDetails()
```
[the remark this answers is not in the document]
olegbespalov: Okay, I do agree that the current `consts` took too many responsibilities against their general purpose. Let me iterate on that.

cmd/version.go
olegbespalov: Even if I like the simplicity of the approach that you're suggesting, I can't share the same feeling about returning plane errors from `Encode`, so I'd like to keep a human friendly wrapping `failed produce a JSON version details: ` here.

lib/consts/consts.go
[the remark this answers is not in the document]
olegbespalov: Done in 37bfccb26c51b7be4ced0a0e1f32ee44b7f0453a

cmd/version.go
olegbespalov: Done in 1854d82aa7c15db3f212d59912630a209b042181

mstoykov (approved):

cmd/version.go
olegbespalov: On the other side it's still can be applied, done in e2535adc1f30926d98c37f25801b185c19797645

mstoykov (approved):

oleiade (approved):
This looks really good, great work! 🚀

PR #4095, opened 2024-12-10
Update golangci-lint to latest v1.62.2

mstoykov:
## What?

update golangci-lint to latest version 

Quite a lot of gosec ones about integers under/overflows - I tried to fix the ones that weren't problematic, but there were quite a lot.

## Why?

Keeping up to date makes it easier to update and it is harder for IC to keep the exact version we use locally and unfortuantely there are always differences.

## Checklist

### DIFF
--- a/lib/executor/ramping_arrival_rate_test.go
+++ b/lib/executor/ramping_arrival_rate_test.go
@@ -281,7 +281,7 @@ func BenchmarkRampingArrivalRateRun(b *testing.B) {
 			testRunState := getTestRunState(b, lib.Options{}, runner)
 			es := lib.NewExecutionState(
 				testRunState, mustNewExecutionTuple(nil, nil),
-				uint64(tc.prealloc.Int64), uint64(tc.prealloc.Int64),
+				uint64(tc.prealloc.Int64), uint64(tc.prealloc.Int64), //nolint:gosec
 			)
 
 			// an high target to get the highest rate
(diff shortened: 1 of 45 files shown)


lib/executor/ramping_arrival_rate_test.go
```diff
@@ -578,8 +578,8 @@ func sqrtRat(x *big.Rat) *big.Rat {
 	var z, a, b big.Rat
 	var ns, ds big.Int
 	ni, di := x.Num(), x.Denom()
-	z.SetFrac(ns.Rsh(ni, uint(ni.BitLen())/2), ds.Rsh(di, uint(di.BitLen())/2))
-	for i := 10; i > 0; i-- { // TODO: better termination
+	z.SetFrac(ns.Rsh(ni, uint(ni.BitLen())/2), ds.Rsh(di, uint(di.BitLen())/2)) //nolint:gosec
+	for i := 10; i > 0; i-- {                                                   // TODO: better termination
```
olegbespalov, line 582: Why indent is so weird? :thinking:

olegbespalov (approved):
Keeping in mind number of the type conversions, `int`/`uint`/etc with omitting of the `gosec` it sounds that it might be so that we probably need some refactoring, on the other side it's for sure out of scope of this PR, so LGTM :+1:

lib/executor/ramping_arrival_rate_test.go
mstoykov: Do you want me to change this @olegbespalov ?

olegbespalov: Whatever you prefer, I don't mind having this even if it looks weird.

mstoykov: I will leave it for now then :sweat_smile:

ankur22 (approved):
LGTM 🚀

PR #4102, opened 2024-12-12
Fix windows packaging

mstoykov:
## What?

Fix windows-packaging workflow on windows-latest

## Why?

So we have signed executables for windows.

## Checklist


ankur22 (approved):

.github/workflows/build.yml
```diff
@@ -244,7 +244,7 @@ jobs:
           move .\packaging\k6-$env:VERSION-windows-amd64\k6.exe .\packaging\
           rmdir .\packaging\k6-$env:VERSION-windows-amd64\
       - name: Add signtool to PATH
-        run: echo "${env:ProgramFiles(x86)}\Windows Kits\10\bin\x64" | Out-File -FilePath $env:GITHUB_PATH -Append
+        run: echo "${env:ProgramFiles(x86)}\Windows Kits\10\bin\10.0.17763.0\x64" | Out-File -FilePath $env:GITHUB_PATH -Append
```
olegbespalov, line 247: This looks really brittle :thinking:

mstoykov: It is, but unfortunately in my search on the internet M$ hasn't made this anyway that this isn;'t like this that I can find. So luckily they only realease windows rarely.

olegbespalov: If you don't mind let's open the issue for fixing that in the long term 👍 but right now yeah, it's probably better to merge it as it's if it works

olegbespalov (approved):

PR #4140, opened 2024-12-19
Add resource_type tag

ankur22:
## What?

A new `resource_type` tag is being added to allow users to view and filter metrics based on the chrome resource types, which are currently defined as:

- "Document"
- "Stylesheet"
- "Image"
- "Media"
- "Font"
- "Script"
- "TextTrack"
- "XHR"
- "Fetch"
- "Prefetch"
- "EventSource"
- "WebSocket"
- "Manifest"
- "SignedExchange"
- "Ping"
- "CSPViolationReport"
- "Preflight"
- "Other"

## Checklist

## Related PR(s)/Issue(s)

Linked: https://github.com/grafana/xk6-browser/issues/1551

### DIFF
--- a/js/modules/k6/browser/common/network_manager.go
+++ b/js/modules/k6/browser/common/network_manager.go
@@ -491,7 +491,7 @@ func (m *NetworkManager) onRequest(event *network.EventRequestWillBeSent, interc
 			event.Request.URL, event.Request.Method, event.Initiator.Type, event.FrameID)
 	}
 
-	req, err := NewRequest(m.ctx, NewRequestParams{
+	req, err := NewRequest(m.ctx, m.logger, NewRequestParams{
 		event:             event,
 		frame:             frame,
 		redirectChain:     redirectChain,
--- a/js/modules/k6/browser/common/network_manager.go
+++ b/js/modules/k6/browser/common/network_manager.go
@@ -193,6 +193,7 @@ func (m *NetworkManager) emitRequestMetrics(req *Request) {
 	})
 }
 
+//nolint:funlen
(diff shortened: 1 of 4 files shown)


js/modules/k6/browser/common/network_manager.go
```diff
@@ -180,6 +180,7 @@ func (m *NetworkManager) emitRequestMetrics(req *Request) {
 	if state.Options.SystemTags.Has(k6metrics.TagURL) {
 		tags = handleURLTag(m.mi, req.URL(), req.method, tags)
 	}
+	tags = tags.With("resource_type", req.ResourceType())
```
ankur22, line 183: There are two things that I'm not sure of:

1. This isn't a system tag, but should it be?
2. This list can change at a moments notice by the chrome team, should we map them from the [chrome ResourceType](https://github.com/grafana/k6/blob/4847cef857bb4ac3a5f104a276496b18602f182e/vendor/github.com/chromedp/cdproto/network/types.go#L32-L49) to a k6 version of it. We would be able to detect if we stop matching on any of the types and log a warning, plus create a unit test to catch this as early as possible.

joanlopez: > We would be able to detect if we stop matching on any of the types and log a warning, plus create a unit test to catch this as early as possible.

This sounds like the way to go! 👍🏻

ankur22: I've added a way to validate the ResourceTypes.

mstoykov (approved):

joanlopez (approved):

oleiade (approved):

PR #4153, opened 2024-12-27
Refactor `new` command

dgzlopes:
Early draft PR. Issue with more context: https://github.com/grafana/k6/issues/4154 

## Checklist


dgzlopes:
Happy to fix the linter things if we feel this approach makes sense 👍

cmd/newtemplates/protocol.js
```diff
@@ -0,0 +1,52 @@
...
+    excludedIngredients: ["pepperoni"],
+    excludedTools: ["knife"],
+    maxNumberOfToppings: 6,
+    minNumberOfToppings: 2
+  };
+
+  let res = http.post(BASE_URL + "/api/pizza", JSON.stringify(restrictions), {
+    headers: {
+      'Content-Type': 'application/json',
+      'X-User-ID': 23423,
```
dgzlopes, line 45: The new versions of QuickPizza no longer use this approach to Auth. 

This needs to be fixed, and the public QuickPizza service needs to be redeployed. cc @federicotdn for visibility.

dgzlopes: Fixed.

oleiade (approved):
Looks good to me 👍🏻 🚀 
Honestly, I'd be happy for us to move forward with this. Have you considered moving back to an engineering career Daniel ? 🙉

cmd/new.go
```diff
@@ -104,64 +38,72 @@ func (c *newScriptCmd) run(cmd *cobra.Command, args []string) error { //nolint:r
...
 }
 
 func getCmdNewScript(gs *state.GlobalState) *cobra.Command {
 	c := &newScriptCmd{gs: gs}
 
-	exampleText := getExampleText(gs, `
-  # Create a minimal k6 script in the current directory. By default, k6 creates script.js.
-  {{.}} new
+	exampleText := getExampleText(c.gs, `
+	# Create a minimal k6 script
```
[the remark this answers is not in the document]
dgzlopes: Makes sense! I changed the help text a bit. Wdyt?

cmd/newtemplates/browser.js
```diff
@@ -0,0 +1,53 @@
...
+export function setup() {
+  let res = http.get(BASE_URL);
+  if (res.status !== 200) {
+    throw new Error(
+      `Got unexpected status code ${res.status} when trying to setup. Exiting.`
+    );
+  }
+}
+
+export default async function() {
```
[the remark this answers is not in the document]
dgzlopes: I copy pasted it from the QuickPizza repo :see_no_evil: 

Better now?

cmd/newtemplates/browser.js
```diff
@@ -0,0 +1,60 @@
+import { browser } from "k6/browser";
+import http from "k6/http";
+import { sleep, check } from 'k6';
+
+const BASE_URL = __ENV.BASE_URL || "https://quickpizza.grafana.com";
+
+export const options = {
+  scenarios: {
+    ui: {
+      executor: "shared-iterations",
```
joanlopez, line 10: I see in other examples we set duration/vus, should we set some "scope" here as well? 
Otherwise, does it bring any value to define the executor?

cmd/newtemplates/browser.js
```diff
@@ -0,0 +1,60 @@
...
+      `Got unexpected status code ${res.status} when trying to setup. Exiting.`
+    );
+  }
+}
+
+export default async function() {
+  let checkData;
+  const page = await browser.newPage();
+
+  try {
```
joanlopez, line 37: Would adding a `catch` here a recommended practice? 🤔 
If so, I'd prefer to have it in place, as this not being a simple test, but likely an entrypoint for k6 newcomers.

cmd/newtemplates/templates.go
```diff
@@ -0,0 +1,74 @@
+// Package templates provides the templates for the k6 new command
+package templates
```
joanlopez, line 2: I have no very strong opinion here, but I think I'd prefer if we can rename the folder to match the package name (i.e. just `templates`) or to a bit more concise/generic name, because I'm not sure if adding the command prefix to the folder really brings any value.

Let's see what others think - cc/ @oleiade

cmd/newtemplates/templates.go
```diff
@@ -0,0 +1,74 @@
...
+	return &TemplateManager{
+		minimalTemplate:  minimalTmpl,
+		protocolTemplate: protocolTmpl,
+		browserTemplate:  browserTmpl,
+	}, nil
+}
+
+// GetTemplate selects the appropriate template based on the type
+func (tm *TemplateManager) GetTemplate(templateType string) (*template.Template, error) {
+	switch templateType {
```
joanlopez, line 53: Nit; I guess we could define the available "types" as exposed constants from the package, so it's a bit more explicit. But feel free to disregard this comment, cause for now it's only used by `new`, and very concretely.

cmd/new.go
```diff
@@ -104,64 +38,78 @@ func (c *newScriptCmd) run(cmd *cobra.Command, args []string) error { //nolint:r
 	if err != nil {
 		return err
 	}
-
 	if fileExists && !c.overwriteFiles {
-		return fmt.Errorf("%s already exists, please use the `--force` flag if you want overwrite it", target)
+		return fmt.Errorf("%s already exists. Use the `--force` flag to overwrite", target)
```
joanlopez, line 42:
```suggestion
		return fmt.Errorf("%s already exists. Use the `--force` flag to overwrite it", target)
```

cmd/new.go
```diff
@@ -104,64 +38,78 @@ func (c *newScriptCmd) run(cmd *cobra.Command, args []string) error { //nolint:r
...
 	if err != nil {
 		return err
 	}
 	defer func() {
-		_ = fd.Close() // we may think to check the error and log
+		if cerr := fd.Close(); cerr != nil {
+			if _, err := fmt.Fprintf(c.gs.Stderr, "error closing file: %v\n", cerr); err != nil {
+				panic(fmt.Sprintf("error writing error message to stderr: %v", err))
+			}
+		}
```
joanlopez, line 54: I'd try to avoid the use of `panic` here, but return whatever error happened (or the result of joining both if more than one happened).

cmd/newtemplates/templates.go
oleiade: I agree with you @joanlopez 👍🏻  renaming the folder to just `templates` would make sense here, considering it's the only command using templates so far 🙇🏻

cmd/newtemplates/browser.js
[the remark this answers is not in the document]
oleiade: Yes 👍🏻 Cheers 🙇🏻

cmd/newtemplates/browser.js
oleiade: I believe in this case, if the error is not caught, the iteration will be aborted with an error anyway. But it would make sense indeed, if possible to catch the error in order to maybe explicitly call `fail` or `exec.test.abort` with a custom message guiding the user through what failed, why, and what they can do about it 👍🏻

cmd/newtemplates/templates.go
oleiade: I'd pretty much like that too 👍🏻 Having them as constants, would help defining them as keys we explicitly own (and also help with later refactorings).

cmd/newtemplates/browser.js
oleiade: 👍🏻 Furthermore, my understanding is that in general users tend to run browser tests with 1 VU/1 iteration, at least when they get started. I know it's the default anyway, but as we're all addressing beginners with this command, it could be helpful to remove as much implicit as possible?

dgzlopes: Makes sense! I have added VUs/Iterations. 

> Otherwise, does it bring any value to define the executor?

It is needed to be able to run a browser test. You must use scenarios, and scenarios must have an executor.

Btw... we do this same thing in lots of other places; maybe it is something we should change?  cc @ankur22
e.g. https://grafana.com/docs/k6/latest/using-k6-browser/#a-simple-browser-test

cmd/newtemplates/browser.js
dgzlopes: I have added catch + fail :eyes: 

Also, another thing we can maybe use in all examples cc @ankur22

cmd/newtemplates/templates.go
dgzlopes: My text editor kills highlighting/IntelliSense for all files inside the folder if it is called `templates` 

I'll try to fix that! But yeah, I have renamed the folder.

cmd/newtemplates/templates.go
dgzlopes: Makes sense. Done!

cmd/new.go
dgzlopes: Fixed!

cmd/new.go
dgzlopes: Gotcha! I removed the panic :eyes:

dgzlopes:
> What do you think about adding code comments on every relevant line present on each template? I know that's very common nowadays across frameworks or similar tools that offer features for setting up a skeleton of your project. In fact, that's how our static template looks like now, and I think it's a nice-to-keep.
> 
> Especially considering that, I guess, the main target of this feature is new users of k6 - I suspect experienced users have their own scripts they duplicate and adjust.

@joanlopez My feeling is that the scripts are small/simple enough to be understandable out of the box. Also, after checking our docs, I saw that we have a weird mix of commented/un-commented examples. 

With that said, I have no strong opinions on the topic, so yeah, happy to add comments if you think it would provide a better experience for newcomers :muscle:

joanlopez:
> @joanlopez My feeling is that the scripts are small/simple enough to be understandable out of the box. Also, after checking our docs, I saw that we have a weird mix of commented/un-commented examples.
> 
> With that said, I have no strong opinions on the topic, so yeah, happy to add comments if you think it would provide a better experience for newcomers 💪

If you ask me, I prefer it with comments, as we have now. But I don't think it's (only) me who should take that decision.
So, approving for now - we can always add the comments later if that's what we decided as a group consensus.

joanlopez (dismissed):

cmd/newtemplates/browser.js
ankur22: You mean adding the VU/iteration to all the other examples too? The situation with scenarios and browser tests isn't great tbh, and being forced to add an executor. I think we wanted to avoid making a very large options block, and the minimal is:

```js
export const options = {
  scenarios: {
    ui: {
      options: {
        browser: {
          type: 'chromium',
        },
      },
    },
  },
};
```

Adding iteration and vus adds a couple more lines to it.

If the general consensus is that it makes is clearer for new users then it's something to consider. CC @inancgumus

inancgumus: I like being explicit in this k6 new command’s output and including VUs and iterations as outputted examples. Sure, we can also mention VUs and iterations in k6-docs here and there, but in general (all k6-docs, if that’s what we mean), I don’t believe we should add more clutter to that already crowded browser options block.

oleiade (dismissed):

dgzlopes:
I added a new commit to fix the linter @oleiade @joanlopez

oleiade (approved):

joanlopez (approved):

inancgumus (approved):

dgzlopes:
Can someone merge this if it is ready to go? I don't have enough permissions :disappointed:

Issue #4124, opened 2023-10-18
nil pointer: page.getFrameElement


inancgumus:
### Brief summary

Test Run ID: 2015119

There is a nil pointer exception in:

```go
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0x11af156]

goroutine 2213 [running]:
github.com/grafana/xk6-browser/common.(*Page).getFrameElement(0xc00225d1e0, 0xc002d50540)
github.com/grafana/xk6-browser@v1.1.0/common/page.go:274 +0x2d6
github.com/grafana/xk6-browser/common.(*Frame).FrameElement(0xc002d50540)
github.com/grafana/xk6-browser@v1.1.0/common/frame.go:834 +0x106
github.com/grafana/xk6-browser/common.(*ElementHandle).checkHitTargetAt(0xc0002d1a90, {0x19d8f28?, 0xc002a7ab60?}, {0x407eff2000000000, 0x407da00000000000})
github.com/grafana/xk6-browser@v1.1.0/common/element_handle.go:70 +0x6a
github.com/grafana/xk6-browser/common.(*ElementHandle).newPointerAction.func1({0x19d8f28, 0xc002a7ab60}, 0x0?)
github.com/grafana/xk6-browser@v1.1.0/common/element_handle.go:1449 +0x52b
github.com/grafana/xk6-browser/common.retryPointerAction({0x19d8f28, 0xc002a7ab60}, 0xc000ea60a0, 0xc000b3cf00)
github.com/grafana/xk6-browser@v1.1.0/common/element_handle.go:1488 +0x3c
github.com/grafana/xk6-browser/common.(*ElementHandle).newPointerAction.func2({0x19d8f28?, 0xc002a7ab60?}, 0xc000b3cf00?, 0x44f120?)
github.com/grafana/xk6-browser@v1.1.0/common/element_handle.go:1476 +0x35
github.com/grafana/xk6-browser/common.(*Frame).click.(*Frame).newPointerAction.func2({0x19d8f28, 0xc002a7ab60}, 0xc0009d3e00, 0xc0009d3e60)
github.com/grafana/xk6-browser@v1.1.0/common/frame.go:1991 +0x116
created by github.com/grafana/xk6-browser/common.call in goroutine 2212
github.com/grafana/xk6-browser@v1.1.0/common/helpers.go:102 +0x15c
```

### xk6-browser version

1.1.0

### OS

[...]

PR #4174, opened 2025-01-13
Fix/4123 click panic

ankur22:
## What?

This makes changes to the codebase so that we alert the user of an error that they could work with to try and fix their environment.

## Why?

When either the load generator and/or chrome are under heavy load (CPU is nearly maxed out), a user might sometimes see an NPD error. This isn't useful and the user cannot do anything with that information. We protect against an NPD by returning an error which should give the user some idea of what steps to take to get their test to work (increase compute resources).

More details can be found in https://github.com/grafana/k6/issues/4124#issuecomment-2577675900.

## Checklist

## Related PR(s)/Issue(s)

https://github.com/grafana/k6/issues/4124

### DIFF
--- a/js/modules/k6/browser/common/execution_context.go
+++ b/js/modules/k6/browser/common/execution_context.go
@@ -113,6 +113,10 @@ func (e *ExecutionContext) adoptBackendNodeID(backendNodeID cdp.BackendNodeID) (
 		return nil, fmt.Errorf("resolving DOM node: %w", err)
 	}
 
+	if remoteObj == nil {
+		return nil, fmt.Errorf("the page may have navigated away or the element is now missing: %w", err)
+	}
+
 	return NewJSHandle(e.ctx, e.session, e, e.frame, remoteObj, e.logger).AsElement(), nil
 }
 
--- a/js/modules/k6/browser/common/execution_context.go
+++ b/js/modules/k6/browser/common/execution_context.go
@@ -113,6 +113,8 @@ func (e *ExecutionContext) adoptBackendNodeID(backendNodeID cdp.BackendNodeID) (
 		return nil, fmt.Errorf("resolving DOM node: %w", err)
 	}
 
+	// This can occur due to race conditions between trying to click on an element
+	// and chrome moving on (e.g. navigating).
 	if remoteObj == nil {
 		return nil, fmt.Errorf("the page may have navigated away or the element is now missing: %w", err)
 	}
--- a/js/modules/k6/browser/common/execution_context.go
+++ b/js/modules/k6/browser/common/execution_context.go
@@ -118,7 +118,7 @@ func (e *ExecutionContext) adoptBackendNodeID(backendNodeID cdp.BackendNodeID) (
 	if remoteObj == nil {
 		return nil, fmt.Errorf(`the page may have navigated away or the element is
 			now missing. It might happen when k6 and/or Chrome are overloaded. You
-			might need to increase the compute resources: %w`, err)
+			might need to increase the compute resources`)
 	}
 
 	return NewJSHandle(e.ctx, e.session, e, e.frame, remoteObj, e.logger).AsElement(), nil
--- a/js/modules/k6/browser/common/execution_context.go
+++ b/js/modules/k6/browser/common/execution_context.go
@@ -116,7 +116,9 @@ func (e *ExecutionContext) adoptBackendNodeID(backendNodeID cdp.BackendNodeID) (
 	// This can occur due to race conditions between trying to click on an element
 	// and chrome moving on (e.g. navigating).
 	if remoteObj == nil {
-		return nil, fmt.Errorf("the page may have navigated away or the element is now missing: %w", err)
+		return nil, fmt.Errorf(`the page may have navigated away or the element is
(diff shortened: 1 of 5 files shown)


js/modules/k6/browser/common/page.go
```diff
@@ -685,11 +689,14 @@ func (p *Page) attachFrameSession(fid cdp.FrameID, fs *FrameSession) {
...
 }
 
-func (p *Page) getFrameSession(frameID cdp.FrameID) *FrameSession {
+func (p *Page) getFrameSession(frameID cdp.FrameID) (*FrameSession, bool) {
 	p.logger.Debugf("Page:getFrameSession", "sid:%v fid:%v", p.sessionID(), frameID)
 	p.frameSessionsMu.RLock()
 	defer p.frameSessionsMu.RUnlock()
-	return p.frameSessions[frameID]
+
+	v, ok := p.frameSessions[frameID]
```
olegbespalov, line 697: Should we also add some logic that will check that `*FrameSession` isn't nil before writing into the `p.frameSessions` map, to be safe against cases like:

```go
p.attachFrameSession(fid, nil)

fs, ok := p. getFrameSession(fid)

// fs = nil
// ok = true
```

What do you think?

js/modules/k6/browser/common/execution_context.go
```diff
@@ -113,6 +113,12 @@ func (e *ExecutionContext) adoptBackendNodeID(backendNodeID cdp.BackendNodeID) (
 		return nil, fmt.Errorf("resolving DOM node: %w", err)
 	}
 
+	// This can occur due to race conditions between trying to click on an element
+	// and chrome moving on (e.g. navigating).
+	if remoteObj == nil {
+		return nil, fmt.Errorf("the page may have navigated away or the element is now missing: %w", err)
```
olegbespalov, line 119: Are there any useful details we could let users know to help them prevent such errors, or at least give a better picture of why that error happens?

js/modules/k6/browser/common/page.go
```diff
@@ -603,7 +603,11 @@ func (p *Page) getFrameElement(f *Frame) (handle *ElementHandle, _ error) {
 	for ; rootFrame.parentFrame != nil; rootFrame = rootFrame.parentFrame {
 	}
 
-	parentSession := p.getFrameSession(cdp.FrameID(rootFrame.ID()))
+	parentSession, ok := p.getFrameSession(cdp.FrameID(rootFrame.ID()))
+	if !ok {
+		return nil, errors.New("parent frame has been detached")
```
olegbespalov, line 608: Same question about the error text and any useful details that we could provide to the user?

olegbespalov (dismissed):
Looks good to me, since still not fully have context :+1:  Left a few non-blocking comments :+1:

js/modules/k6/browser/common/execution_context.go
ankur22: There can be multiple reasons for this occurring, so it's not easy to add into an error -- load generator overloaded; chrome overloaded; use of elementHandle instead of locators.

I think for now we can return a new error, and wait for feedback to help us understand where exactly users are hitting this issue. At the moment I've seen this occur when chrome is overloaded, but the load generator is fine.

js/modules/k6/browser/common/page.go
ankur22: I don't see anything wrong with being defensive. In our case `attachFrameSession` is called from [here](https://github.com/grafana/k6/blob/88f765b9dafa2749192f415d983960a66708bc47/js/modules/k6/browser/common/frame_session.go#L1027), where the `frameSession` is created on the line above.

Added in 45e8b6a0fd3f13565e6551bed7a0361e9ade6987

js/modules/k6/browser/common/execution_context.go
codebien: If we don't add this information, then we might not fully respect what the description in What/Why is reporting.

What about the following form?

```suggestion
		return nil, fmt.Errorf("the page may have navigated away or the element is now missing. It might happen when k6 and/or Chrome are overloaded. You might need to increase the compute resources: %w", err)
```

js/modules/k6/browser/common/execution_context.go
```diff
@@ -113,6 +113,14 @@ func (e *ExecutionContext) adoptBackendNodeID(backendNodeID cdp.BackendNodeID) (
 		return nil, fmt.Errorf("resolving DOM node: %w", err)
 	}
 
+	// This can occur due to race conditions between trying to click on an element
+	// and chrome moving on (e.g. navigating).
+	if remoteObj == nil {
+		return nil, fmt.Errorf(`the page may have navigated away or the element is
+			now missing. It might happen when k6 and/or Chrome are overloaded. You
+			might need to increase the compute resources: %w`, err)
```
olegbespalov, line 121: Isn't `err` always nil in that case?

ankur22: Good spot, removed in 7d4fa08e59274c40af949037688a351ee676e346

olegbespalov (approved):

codebien (approved):

inancgumus (approved):
👍

PR #4176, opened 2025-01-14
Warn on shorthand options overriding scenarios

### LINKED ISSUE #2529: Warn the user that using `-i`/`-u`/`-d` will overwrite everything about their scenarios
### Feature Description

Users are sometimes confused why using `-i`, `-u` or/and `-d` works the way it does. This includes both why does it suddenly not use their [`exec` function](https://community.k6.io/t/harness-docker-k6-error-function-default-not-found-in-exports/4965) or why they don't [use the `env` they have defined](https://community.k6.io/t/scenario-env-variables-are-undefined-when-iterations-or-duration-is-provided/3674). 

While there are some indications that we no longer run the same scenarios, it likely will be nice to be more explicit.  

p.s. forgot about `-s` for stages, as I never see it used :facepalm: 

On usage of any of the shortcut options - print a warning stating that this will *completely* overwrite the whole `options.scenarios` possibly somewhere in https://github.com/grafana/k6/blob/c055402af97ee7e403edeb13396a49185e304365/lib/executor/execution_config_shortcuts.go#L72 

_No response_

mstoykov:
## What?

Warn on shorthand options(vus, duration, stages, iterations) overriding scenarios from previously defined scenarios. Mostly applicable when using cli flags to override them from within the script.

This is currently implemented only for scenarios as that is the case that is most confusing. Implementing this for duration overriding stages (for example) is possible, but it seems to not be nearly hit enough.

Unfortunately due to too many of things of #883 and the way shorthand options are handled in particular, the code can't be in any Apply function with enough information to have a nice message.

## Why?

This is probably one of the things that confuses the most users as explained in the original #2529 issue.

## Checklist

## Related PR(s)/Issue(s)

Fixes #2529

### DIFF
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -198,9 +198,15 @@ func getConsolidatedConfig(gs *state.GlobalState, cliConf Config, runnerOpts lib
 
 	conf = cliConf.Apply(fileConf)
 
+	warnOnShortHandOverride(conf.Options, runnerOpts, "script", gs.Logger)
 	conf = conf.Apply(Config{Options: runnerOpts})
 
-	conf = conf.Apply(envConf).Apply(cliConf)
+	warnOnShortHandOverride(conf.Options, envConf.Options, "env", gs.Logger)
+	conf = conf.Apply(envConf)
+
+	warnOnShortHandOverride(conf.Options, cliConf.Options, "cli", gs.Logger)
+	conf = conf.Apply(cliConf)
+
 	conf = applyDefault(conf)
 
 	// TODO(imiric): Move this validation where it makes sense in the configuration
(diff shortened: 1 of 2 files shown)


cmd/config.go
```diff
@@ -215,6 +221,15 @@ func getConsolidatedConfig(gs *state.GlobalState, cliConf Config, runnerOpts lib
 	return conf, nil
 }
 
+func warnOnShortHandOverride(a, b lib.Options, bName string, logger logrus.FieldLogger) {
+	if a.Scenarios != nil &&
+		(b.Duration.Valid || b.Iterations.Valid || b.Stages != nil || b.Scenarios != nil) {
+		logger.Warnf(
+			"%q level configuration overrode scenarios configuration entirely",
```
olegbespalov, line 228: Maybe explicitly mention implications, what do you think?

```suggestion
			"%q level configuration overrode scenarios configuration entirely. It's highly likely, that the test load result will be different from what you expect.",
```

olegbespalov (approved):

cmd/config.go
mstoykov: This usually isn't the problem - the problem is people expect:
1. to run their scenarios but with a different duration, iterations or VUs, but with all other options the same
2. having some other option stay the same - for example `exec`, `env`, and  also `options` for browser

But making the message a bunch of sentences IMO isn't a great idea. The one we have with docker and stuff about not finding files - it still gets users to post it and ask questions. I only expect this is happening less because of google actually now fidning the results from previous asks.

joanlopez (approved):

PR #4191, opened 2025-01-17
Remove BaseEventEmitter from components

ankur22:
## What?

Removing the BaseEventEmitter from `Touchscreen`, `Page` and `Worker`, as well as the events which are emitted.

## Why?

While debugging another issue i found that these components had instances of their own event emitter, which were emitting events, but nothing was consuming them. It's confusing to see these events being emitted and can cause confusion while debugging issues with the event emitter. This is why they are being removed.

## Checklist

### DIFF
--- a/js/modules/k6/browser/common/frame_manager.go
+++ b/js/modules/k6/browser/common/frame_manager.go
@@ -470,12 +470,6 @@ func (m *FrameManager) removeFramesRecursively(frame *Frame) error {
 	delete(m.frames, cdp.FrameID(frame.ID()))
 	m.framesMu.Unlock()
 
-	if !m.page.IsClosed() {
-		m.logger.Debugf("FrameManager:removeFramesRecursively:emit:EventPageFrameDetached",
-			"fmid:%d fid:%v fname:%s furl:%s",
-			m.ID(), frame.ID(), frame.Name(), frame.URL())
-	}
-
 	return nil
 }
 
--- a/js/modules/k6/browser/common/frame_manager.go
+++ b/js/modules/k6/browser/common/frame_manager.go
@@ -157,8 +157,6 @@ func (m *FrameManager) frameAttached(frameID cdp.FrameID, parentFrameID cdp.Fram
 
 		m.logger.Debugf("FrameManager:frameAttached:emit:EventPageFrameAttached",
 			"fmid:%d fid:%v pfid:%v", m.ID(), frameID, parentFrameID)
-
-		m.page.emit(EventPageFrameAttached, frame)
 	}
 }
 
@@ -476,8 +474,6 @@ func (m *FrameManager) removeFramesRecursively(frame *Frame) error {
 		m.logger.Debugf("FrameManager:removeFramesRecursively:emit:EventPageFrameDetached",
 			"fmid:%d fid:%v fname:%s furl:%s",
 			m.ID(), frame.ID(), frame.Name(), frame.URL())
-
-		m.page.emit(EventPageFrameDetached, frame)
 	}
 
 	return nil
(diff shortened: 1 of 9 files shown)


inancgumus (approved):
They are not needed 👍

js/modules/k6/browser/common/frame_session.go
```diff
@@ -661,7 +661,7 @@ func (fs *FrameSession) onConsoleAPICalled(event *cdpruntime.EventConsoleAPICall
 }
 
 func (fs *FrameSession) onExceptionThrown(event *cdpruntime.EventExceptionThrown) {
-	fs.page.emit(EventPageError, event.ExceptionDetails)
+	// TODO: Test and handle this
```
olegbespalov, line 664: I am not sure if I'm getting this todo and also, does that make sense to have this empty body method at all?

js/modules/k6/browser/common/frame_manager.go
```diff
@@ -537,16 +529,13 @@ func (m *FrameManager) requestFinished(req *Request) {
 
 func (m *FrameManager) requestReceivedResponse(res *Response) {
 	m.logger.Debugf("FrameManager:requestReceivedResponse", "fmid:%d rurl:%s", m.ID(), res.URL())
-
-	m.page.emit(EventPageResponse, res)
```
olegbespalov, line 541: Do we still need this method if it only does logs?

js/modules/k6/browser/common/frame_manager.go
```diff
@@ -476,8 +474,6 @@ func (m *FrameManager) removeFramesRecursively(frame *Frame) error {
 		m.logger.Debugf("FrameManager:removeFramesRecursively:emit:EventPageFrameDetached",
 			"fmid:%d fid:%v fname:%s furl:%s",
 			m.ID(), frame.ID(), frame.Name(), frame.URL())
-
-		m.page.emit(EventPageFrameDetached, frame)
```
olegbespalov, line 480: Do we need also to keep the whole `m.page.IsClosed() ` only for logging?

joanlopez (commented):
Is that somehow "publicly" exposed? Like, can it be somehow used by k6 users? Or those events are only being consumed internally by other components?

js/modules/k6/browser/common/page.go
```diff
@@ -546,10 +543,7 @@ func (p *Page) consoleMsgFromConsoleEvent(e *runtime.EventConsoleAPICalled) (*Co
 func (p *Page) closeWorker(sessionID target.SessionID) {
 	p.logger.Debugf("Page:closeWorker", "sid:%v", sessionID)
 
-	if worker, ok := p.workers[sessionID]; ok {
-		worker.didClose()
```
olegbespalov, line 550: Maybe the method name `closeWorker` also should be re-visited, since now, it's more about cleaning ? :thinking:

ankur22:
> Is that somehow "publicly" exposed? Like, can it be somehow used by k6 users?
> Or those events are only being consumed internally by other components?

@joanlopez These are all internal events, nothing exposed to the k6 user.

js/modules/k6/browser/common/frame_session.go
ankur22: Right, i think we need an issue for this as it may be something useful to build an API around. When the website throws an exception then the user might want to track this.

I don't want to delete it. Instead i might add some debug logging when this is called.

js/modules/k6/browser/common/frame_manager.go
ankur22: Agreed, removed in d94958f816a165533fa62631f114f3e23147c6bc

js/modules/k6/browser/common/page.go
ankur22: Renamed it to `removeWorker` in 38d7d79a756fdea45db8ed2a89e56d0007eab988

js/modules/k6/browser/common/frame_session.go
olegbespalov: I see, yeah logging that sounds reasonable, feel free to do this later, but it's worth maybe updating TODO within this PR to put more context for the future implementation :+1:

olegbespalov (dismissed):
LGTM, like I said in-line it's worth providing more context for TODO, like `Why?` and optionally `What?`, but consider this non-blocking for merge

inancgumus (approved):

olegbespalov (approved):

PR #4192, opened 2025-01-17
Fix memory leak in queue

ankur22:
## What?

Overwrite the empty `queue.read` with a new slice.

## Why?

When we read off the `queue.read` slice, it essentially pops the first available element by doing this:

```go
case eh.ch <- eh.queue.read[0]:
        eh.queue.read[0] = Event{}
	eh.queue.read = eh.queue.read[1:]
```
([link to code](https://github.com/grafana/k6/blob/3f1536e258273e0df5ff347ee63e39fbf4a3c798/js/modules/k6/browser/common/event_emitter.go#L146-L148))

So the entry where the event used to exist has been overwritten with a fresh instance of `Event` which will help avoid a memory leak (which was recently [fixed](https://github.com/grafana/xk6-browser/pull/1488)). However the issue is that the underlying slice keeps growing. During a long running test the slice could grow to an unnecessarily length.

To avoid this issue, when `queue.read` is [empty](https://github.com/grafana/k6/blob/3f1536e258273e0df5ff347ee63e39fbf4a3c798/js/modules/k6/browser/common/event_emitter.go#L137) and is swapped with `queue.write`, just before swapping them around the existing `queue.read` is overwritten with a new slice, freeing up the old potentially long slice. This should help avoid further memory leaks.

## Checklist

### DIFF
--- a/internal/js/modules/k6/browser/common/event_emitter.go
+++ b/internal/js/modules/k6/browser/common/event_emitter.go
@@ -136,6 +136,8 @@ func (e *BaseEventEmitter) emit(event string, data any) {
 		// the read queue until that is again depleted.
 		if len(eh.queue.read) == 0 {
 			eh.queue.writeMutex.Lock()
+			// Clear the read slice before swapping to prevent keeping references
+			eh.queue.read = make([]Event, 0)
 			eh.queue.read, eh.queue.write = eh.queue.write, eh.queue.read
 			eh.queue.writeMutex.Unlock()
 		}


inancgumus (approved):
Interesting! 👍 Btw, the older commits from that other PR are still showing up in this PR.

joanlopez:
> However the issue is that the underlying slice keeps growing.

Does that mean that the Go GC isn't able to collect the memory allocations that are no longer pointed by any slice? 🤔 

I guess that after:

```go
eh.queue.read = eh.queue.read[1:]
```

The memory address where `eh.queue.read[0]` is stored remains, but since after that line there's no slice pointing to that position (`eh.queue.read` now points to, since the next position `eh.queue.read[1]` till the end of the underlying array), I'd expect the GC to be smart enough to collect these addresses.

Instead, I guess that (based on the change on this PR), it is only able to collect a range of memory addresses that's not pointed at all by any slice. Is that reasoning right? 🤔

joanlopez (approved):

ankur22:
@joanlopez the GC will only cleanup the element that the 0th entry is pointing to if nothing is referencing it. So the underlying array remains in place i believe.

In https://github.com/grafana/xk6-browser/pull/1488, we found this the hard way when the reference was being kept even though there was no way to get back to the 0th element after performing

```go
eh.queue.read = eh.queue.read[1:]
```

Therefore by making the change in this PR, we are sure that the underlying array is empty and the previous array is cleaned up by the GC.

joanlopez (dismissed):

joanlopez (approved):

inancgumus (approved):

Issue #4194, opened 2025-01-17
Enable TypeScript support by default


andrewslotin:
### Feature Description

k6 v0.52.0 introduced experimental support for TypeScript as an opt-in feature. To enable this, users must specify `experimental_enhanced` as a [compatibility mode](https://grafana.com/docs/k6/latest/using-k6/javascript-typescript-compatibility-mode/) when running the `k6 run` command.

After six months, this feature can now be regarded as stable and is set as the default for executing scripts. In contrast, the current default `extended` mode is available as an opt-out option.

### Suggested Solution (optional)

Drop the `experimental_` prefix from the mode name and use it as the new default for `--compatibility-mode`/`K6_COMPATIBILITY_MODE` option, i.e. `--comaptibility-mode=enhanced`.

### Already existing or connected issues / PRs (optional)

* https://github.com/grafana/k6/issues/3703
* https://github.com/grafana/k6/pull/3738

Issue #3864, opened 2024-07-19
Drop aliasing `global` to `globalThis` in `extented` compatibility-mode


mstoykov:
### What?

As of #3456 the only difference between `base` and `extended` compatibility-mode is that `extended` adds an alias for `global` that points to `globalThis`.

This only exists as this was what core-js did and when it was dropped it was readded in https://github.com/grafana/k6/pull/1845. 

I don't remember if we had a particular user that complaint about this or not, but it has been a few years and `globalThis` is pretty much everywhere.

This also has been known to be used as a test if you are running in a browser or NodeJS like environment - which doesn't usually work great with k6.

### Solution

Just stop aliasing it - making `base` and `extented` the same.

#### The support for this is literally 3 lines
https://github.com/grafana/k6/blob/111985bfe258f1b68cc8d19c8b8a21737a5b4f89/js/bundle.go#L387-L392
it doesn't really add much

Arguably removing this will require us to explain why extended and base are the same thing :sweat_smile: 

#### This being used

The only thing *I* know about that this is being used for is k6chaijs which is due to dependancy.

https://github.com/chaijs/type-detect which is part of chaijs is using it instead of `globalThis` a patch for this has been done years ago, but no new version has been released in 7 years.

https://github.com/chaijs/type-detect/issues/146 asked for it to be bumped few years ago, and it might be good idea for us to go ahead and start this rolling

I would argue that this is blocking this specifically from going forward, but once this is done we can think about it more.

### Related issues

#3860

PR #4200, opened 2025-01-20
Enable typescript support by default

szkiba:
## What?

Enable typescript support by default.

## Why?

k6 v0.52.0 introduced experimental support for TypeScript as an opt-in feature. To enable this, users must specify `experimental_enhanced` as a [compatibility mode](https://grafana.com/docs/k6/latest/using-k6/javascript-typescript-compatibility-mode/) when running the `k6 run` command.

After six months, this feature can now be regarded as stable and is set as the default for executing scripts. In contrast, the current default `extended` mode is available as an opt-out option.

## Checklist

## Related PR(s)/Issue(s)

* https://github.com/grafana/k6/issues/3703
* https://github.com/grafana/k6/pull/3738

 Closes https://github.com/grafana/k6/issues/4194


mstoykov (changes requested):
The request changes is just so it doesn't get merged by accident.

This introduces 2 breaking changes:
1. with removign experimental_enhanced as compatiblity mode it will break anyone who is using it, including any archives made with it. 
2. as https://github.com/grafana/k6/issues/3860 explained *enhanced doesn't alias `global` to `globalThis` and this is likely going to break even more scripts given the usage statistics we have.

Additionally as discussed in https://github.com/grafana/k6/issues/3864 we kind of want to remove compatibility-mode now, arguably this should have never been part of it but separate option, but alas.

I would prefer if I take over this and remove the dependancy on compatibility-mode for this and just enable it by default. The question is do we add an option a kin to node to disable/enable this or is that not relevant as it only is enabled on files with `.ts` file extension?

szkiba:
Please go ahead. I would be happy if you would take it.

mstoykov:
Closing in favor of #4292

PR #4212, opened 2025-01-21
Grpc get localhost certificate programically

mstoykov:
## What?

Get the localhost certificate programatically instead of having a hard copy of it from within stdlib

## Why?
This localhost certificate was update in https://github.com/golang/go/commit/6783377295e0878aa3ad821eefe3d7879064df6d and this broke this tests in a none obvious way. 

Big part of the problem is that the `httptest` stdlib package always use the in stdlib certificate. 

But this test had a copy of this, but it is not obvious that the problem was this Certificate instead of something else.

## Checklist

### DIFF
--- a/js/modules/k6/grpc/client_test.go
+++ b/js/modules/k6/grpc/client_test.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"crypto/tls"
 	"crypto/x509"
+	"encoding/pem"
 	"fmt"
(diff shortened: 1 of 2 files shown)


mstoykov:
This also reverts #4211 as this now fixes the specific issue.

inancgumus (approved):

ankur22 (approved):

PR #4274, opened 2025-01-23
refactor crypto_test.go againts web platform tests

### LINKED ISSUE #4268: Refactor `crypto_test.go`
# What?

The original implementation of the WebCrypto API kept static Deplatform tests in the repository. That's approach was fine for the beginning, but less maintainable since we weren't able easily to answer the question of what's the difference between our version and original version of the test case from the web platform test suite.

Likely in grafana/xk6-webcrypto#87 and grafana/xk6-webcrypto#90 we adjusted that and started to use original suite of tests, with the patches applied on top of it, but that was done only for `subtle_crypto_test.go`. `crypto_test.go` still have static versions of the tests.

It's probably good to do align `crypto_test.go` and run the tests against the web platform tests, rather the static versions.

# Why?

Running tests against WebPlatfrom test suite guaranties that our implementation compliant with the WebCrypto API specs, and if there are some differences we could easily store them as patches which makes it more transparent for our customers.

aniqaqill:
## What?

This PR adds Web Platform Tests (WPT) for the `Crypto` interface in the k6 WebCrypto API implementation. 

The tests are integrated into the existing WPT framework, and any necessary patches are applied to ensure compatibility with the k6 runtime.

---

## Why?

These changes ensure that the k6 implementation of the WebCrypto API is compliant with the WebCrypto specification. By running the WPT suite, we can:

1. **Improve Maintainability**:
   - Use the WPT suite to catch regressions and ensure consistency with the specification.

2. **Increase Transparency**:
   - Make it easier for contributors and users to understand how the k6 implementation aligns with the WebCrypto API.

---

## Checklist

---

## Related PR(s)/Issue(s)

---

## Changes Made

1. **Updated `crypto_test.go`**:
   - Integrated the WPT test files into the test suite.

2. **Applied Patches**:
   - Generated and applied patches to ensure the WPT tests are compatible with the k6 runtime.

3. **Verified Tests**:
   - Ran the tests locally to ensure they pass and cover the implemented functionality.

---

## Thanks for your contribution! 🙏🏼

### DIFF
--- a/internal/js/modules/k6/webcrypto/tests/crypto_test.go
+++ b/internal/js/modules/k6/webcrypto/tests/crypto_test.go
@@ -3,267 +3,177 @@
 package tests
 
 import (
-	"fmt"
+	"os"
+	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
 )
 
-func TestGetRandomValues(t *testing.T) {
+// TestCryptoWebPlatformTestSuite runs the Web Platform Tests for the Crypto interface.
+// It ensures that the k6 implementation of the WebCrypto API is compliant with the specification.
+//
+// The tests are loaded from the Web Platform Tests repository and executed in the k6 runtime.
+// Any differences between the k6 implementation and the specification are documented in patches.
+func TestCryptoWebPlatformTestSuite(t *testing.T) {
 	t.Parallel()
 
-	ts := newConfiguredRuntime(t)
-
-	_, gotErr := ts.RunOnEventLoop(`
-		var input = new Uint8Array(10);
-		var output = crypto.getRandomValues(input);
-
-		if (output.length != 10) {
-			throw new Error("output.length != 10");
-		}
-
-		// Note that we're comparing references here, not values.
-		// Thus we're testing that the same typed array is returned.
-		if (input !== output) {
-			throw new Error("input !== output");
-		}
-	`)
-
-	assert.NoError(t, gotErr)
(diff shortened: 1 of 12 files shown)


olegbespalov (changes requested):
Hi @aniqaqill !

Thanks for the opening the pull-request, however I'm not sure if it addresses the need of #4268.

So the predominant point of the #4268 is to migrate from the hardcoded tests (JavaScript code) to running the tests that are part of the web platform test suite. 

For instance, `TestGetRandomValuesSupportedTypedArrays` has a comment that it was based on the https://github.com/web-platform-tests/wpt/blob/75db68b05454e7f3a81e063373defd094c176a16/WebCryptoAPI/getRandomValues.any.js (not I've replaced the `master` reference to a precise commit that we use for checking out the test). 

So the desired state is that instead of current hardcoded logic:

```golang
...
_, gotErr := ts.RunOnEventLoop(fmt.Sprintf(`
				var buf = new %s(10);
				crypto.getRandomValues(buf);
				if (buf.length != 10) {
					throw new Error("buf.length != 10");
				}
			`, tc.typedArray))
...
```

We run the test against the files of web platform tests.

Furthermore, it seems like `crypto_test.go` just duplicates what is presented in the `subtle_crypto_test.go`.

Please let me know if you have more questions.

olegbespalov:
For now, we decided to close this PR, feel free to re-open once feedback will be applied.

PR #4276, opened 2025-01-24
Do not stop REST API before outputs are flushed

mstoykov:
## What?

Do not stop REST API until flushing outputs is done

## Why?

Flushing outputs may take some time and it will be good if the REST API is up for this - there is no need for it not to be.

## Checklist


joanlopez (approved):

ankur22 (approved):
What i understand is that the REST API will close after the outputs are flushed due to the change in ordering of the defer.

LGTM

Issue #4279, opened 2025-01-24
NPD when calling deferred handle.Dispose


ankur22:
### Brief summary

Here's the stack trace:

```bash
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x13957a1]

goroutine 1517439 [running]:
go.k6.io/k6/js/modules/k6/browser/common.(*Frame).waitForSelector.func1()
go.k6.io/k6/js/modules/k6/browser/common/frame.go:505 +0x21
go.k6.io/k6/js/modules/k6/browser/common.(*Frame).waitForSelector(0xc004638fc0, {0xc00041e2d0, 0x2a}, 0xc00005d3e0)
go.k6.io/k6/js/modules/k6/browser/common/frame.go:511 +0x434
go.k6.io/k6/js/modules/k6/browser/common.(*Frame).typ.(*Frame).newAction.func2({0x1f9f158, 0xc001bee770}, 0xc004f21340, 0xc004f21c70)
go.k6.io/k6/js/modules/k6/browser/common/frame.go:2111 +0xfb
created by go.k6.io/k6/js/modules/k6/browser/common.call in goroutine 1517438
go.k6.io/k6/js/modules/k6/browser/common/helpers.go:104 +0x14c
```

Looking at the code in question:

```go
		defer func() {
			if err := handle.Dispose(); err != nil {
				err = fmt.Errorf("disposing element handle: %w", err)
				rerr = errors.Join(err, rerr)
			}
		}()
		if handle, err = ec.adoptElementHandle(handle); err != nil {
			return nil, fmt.Errorf("waiting for selector %q: adopting element handle: %w", selector, err)
		}
```

If an error is returned from `ec.adoptElementHandle` it overwrites the `handle` with a `nil`. The deferred function tries to use the `handle` but fails with an NPD since it is now `nil`.

### k6 version

v0.56.0

### OS

na

### Steps to reproduce the problem

NA

### Expected behaviour

Don't NPD

### Actual behaviour

NPD

PR #4280, opened 2025-01-24
Fix an NPD when handle is overwritten with nil

ankur22:
## What?

Call the defer function with the non nil handle, and allow the handle to be overwritten later. Now the handle in the defer will still be non-nil.

## Why?

Fixes an NPD.

## Checklist

## Related PR(s)/Issue(s)

Linked to: https://github.com/grafana/k6/issues/4279

### DIFF
--- a/internal/js/modules/k6/browser/common/frame.go
+++ b/internal/js/modules/k6/browser/common/frame.go
@@ -500,12 +500,12 @@ func (f *Frame) waitForSelector(selector string, opts *FrameWaitForSelectorOptio
 	// an element should belong to the current execution context.
 	// otherwise, we should adopt it to this execution context.
 	if ec != handle.execCtx {
-		defer func() {
+		defer func(handle *ElementHandle) {
 			if err := handle.Dispose(); err != nil {
 				err = fmt.Errorf("disposing element handle: %w", err)
 				rerr = errors.Join(err, rerr)
 			}
-		}()
+		}(handle)
 		if handle, err = ec.adoptElementHandle(handle); err != nil {
 			return nil, fmt.Errorf("waiting for selector %q: adopting element handle: %w", selector, err)
 		}


internal/js/modules/k6/browser/common/frame.go
```diff
@@ -500,12 +500,12 @@ func (f *Frame) waitForSelector(selector string, opts *FrameWaitForSelectorOptio
...
 	// otherwise, we should adopt it to this execution context.
 	if ec != handle.execCtx {
-		defer func() {
+		defer func(handle *ElementHandle) {
 			if err := handle.Dispose(); err != nil {
 				err = fmt.Errorf("disposing element handle: %w", err)
 				rerr = errors.Join(err, rerr)
 			}
-		}()
+		}(handle)
```
mstoykov, line 508: I find this fix *strange*.  IMO either this fixes an additional bug, it creates a new bug ... or is unnecessary :sweat_smile: 

So previously we got a handle, and then we adopted, and then we Disposed of it - which seems like ... not what we want?

We are now no longer going to do this - we will Dispose of the not adopted handle each time. Regardless of whether the adoption error-ed or not.

So are we fixing an additonal bug? Should we have always been disposing the previous handle? 
Or should we dispose the newly created handle if it was actually created *or* the old one if it error-ed?

ankur22: Yep, good spot. It doesn't make sense to dispose of the new handle. Instead the previous handle needs to be disposed of when we have the adopted handle.

I'll change it so that:
1. The old handle is disposed **only** when;
2. the adoption was successful.

It looks like we currently [error on when the handle is already part of the correct execution context](https://github.com/grafana/k6/blob/7129195a85e57533564e46086018e43915d27945/internal/js/modules/k6/browser/common/execution_context.go#L146), which doesn't seem right. We should probably catch that error [here](https://github.com/grafana/k6/blob/7129195a85e57533564e46086018e43915d27945/internal/js/modules/k6/browser/common/frame.go#L509); not dispose (since it is the original handle) and carry on with the process as normal. However, we also [perform the same check](https://github.com/grafana/k6/blob/7129195a85e57533564e46086018e43915d27945/internal/js/modules/k6/browser/common/frame.go#L503) before trying to adopt, so i will leave it as it is.

Tbh, i'm not entirely sure of the logic here, why we need to adopt and why we need to be explicit in disposing of the older handle. @inancgumus do you have an idea of why we need to do all this in the first place?

ankur22: Fixed in 8100f041339472b2ae1acc8e6a32ef338db4b2cb

ankur22: Fixed in 3f48b127455ea8a9972ad12ac6ae4e6678e29bd0

mstoykov (commented):
LGTM!

Can we check if this fixes some other issues? The previous code really looked suspiciously like a thing that will break stuff.

oleiade (approved):

internal/js/modules/k6/browser/common/frame.go
inancgumus: @ankur22 We adopt a new handle because the older one's execution context is no longer valid (e.g., its frame might have been navigated away). We dispose of the older handle because we now know that it's no longer valid. If the element belongs to the same execution context, we won't adopt it. Erroring out if the handle is in the same execution context is just a safety check.

PR #4056, opened 2024-11-13
Merge browser in k6

mstoykov:
## What?

Move xk6-browser inside k6 codebase

## Why?

Browser is no longer experimental and it makes more sense to have a single repository

## List of commands I've ran:

1. `git subtree add -P  js/modules/k6/browser git@github.com:grafana/xk6-browser.git main` - requires git-subtree - in practice can be done with a bunch of additional commands, just this is faser and easier
2. cd js/modules/k6/browser/
3. `git rm -r  CODE_OF_CONDUCT.md CODEOWNERS CONTRIBUTING.md docker-compose.yaml Dockerfile go.mod go.sum LICENSE Makefile README.md register.go SECURITY.md SUPPORT.md sync_register.go TROUBLESHOOTING.md release\ notes/ assets .golangci.yml  .gitignore .github/`
4. `git commit -m "browser: cleanup unnsecessary files" -n`
5. `find . -iname '*.go'   | xargs sed -i -r  's|github.com/grafana/xk6-browser|go.k6.io/k6/js/modules/k6/browser|g'`
6. edit js/jsmodules.go to import the new import
7. go mod tidy && go mod vendor
8. `git commit -m "Move to using the in k6 browser code"`
9. Update .golangci.yml to disable revive and contextcheck and git commit it
10. `git mv js/modules/k6/browser/examples/ examples/browser &&  git rm examples/browser.js examples/experimental/browser.js &&  git commit -m "Move browser examples and clean old ones" -a`
11. `cd examples/browser/ && sed -i -r  's!/x/browser/async!/browser!g' * && git commit -m "Fix imports in browser examples"`
12. `cd ../../js/modules/k6/browser && git rm -r  sync-examples/ &&  git commit -m "browser: remove sync examples"`
13. Copy an updates e2e.yml from the browser repo to make the browser examples execute  

## TODO:
- [x] general workflow on how this PR will be made
- [x] moving examples
- [x] running example tests
- [x] fix lint issues

## Checklist

- [ ] I have performed a self-review of my code.
- [ ] I have added tests for my changes.
- [ ] I have run linter locally (`make lint`) and all checks pass.
- [ ] I have run tests locally (`make tests`) and all tests pass.
- [ ] I have commented on my code, particularly in hard-to-understand areas.

PR #4282, opened 2025-01-24
Move Prometheus remote write to k6 repository

codebien:
## What?

It migrates the codebase of Prometheus remote-write metrics output from the external dedicated repository to k6 repository.

## Why?

It is the first step to graduate Prometheus remote-write as a stable metrics output.


joanlopez:
> How about preserving the git history, by using `git subtree` like we did for browser (#4056) or webcrypto?

I agree this would be nice, yeah! 👍🏻

olegbespalov (approved):

codebien:
Closes in favor of https://github.com/grafana/k6/pull/4519

PR #4288, opened 2025-01-27
Fix to remove browser downloads path

inancgumus:
## What?

Each `BrowserContext` has a download path. We now clean them up when the `Browser` they are associated with closes.

## Why?

To avoid inflating the user's temporary directory folder.

## Checklist


internal/js/modules/k6/browser/common/browser_context.go
```diff
@@ -110,6 +110,16 @@ func (b *BrowserContext) setDownloadsPath(path string) error {
 	return nil
 }
 
+// cleanup cleans up the resources associated with the browser context.
+func (b *BrowserContext) cleanup() error {
```
ankur22, line 114: Do we want to cleanup a remote browser's browserContext's directory too? At the moment we only cleanup the data directory when we [start](https://github.com/grafana/k6/blob/7129195a85e57533564e46086018e43915d27945/internal/js/modules/k6/browser/common/browser_process.go#L186) a local chrome process.

ankur22 (approved):

mstoykov (approved):
LGTM and thank you :bow: 

 This does get my `*browser*` directories in /tmp from 866 to 33. 

The remaining ones are 
- xk6-browser-data* 
- k6browser-artifacts*

Which seem to be completely different parts, although we likely should at least remove the `xk6-` part :sweat_smile: 

But we can do in a separate PR.

internal/js/modules/k6/browser/common/browser_context.go
inancgumus: In this PR, we only clean up what we have directly created that is local to a k6 process. For remote browsers, `cleanup` is similar to a no-op:

> // If the path does not exist, RemoveAll
> // returns nil (no error).
https://pkg.go.dev/os#RemoveAll

Do you mean something else?

ankur22: Ah, right! It's downloaded files 👍

Issue #4281, opened 2025-01-24
Implement `page.on('request')`


ankur22:
## What

We want to implement `page.on('request')`. More details of the API can be found [here](https://playwright.dev/docs/api/class-page#page-event-request). The request object that is returned to the handler is detailed [here](https://playwright.dev/docs/api/class-request). This is a read only API, and no modifications can be done on the request object. k6 browser should not wait for a response from the handler, it fire and forgets.

## Why

Users have asked to be able to read the requests that are being sent out from chromium to the website under test. This can be helpful in validating/asserting certain aspects of the request which could be important for the test to be deemed successful.

## Comments

While testing the Playwright implementation i noticed that some of the methods on the request object are async. An example of this is `data.response()`:

```js
    page.on('request', async (data) => {
      console.log("url: " + data.url());
      if (data.url() === "https://quickpizza.grafana.com/api/quotes") {
        console.log(await data.response());
      }
    })
```

What's interesting about this is that the call to `data.response()` will in fact wait until a response is received. The current implementation of [response](https://github.com/grafana/k6/blob/7129195a85e57533564e46086018e43915d27945/internal/js/modules/k6/browser/common/http.go#L332-L335) on the request object does not fulfil this behaviour and we will need to change that too.

Async methods on request that will need to be checked to see if they behave the same as in PW:

- [allHeaders](https://playwright.dev/docs/api/class-request#request-all-headers)
- [headerValue](https://playwright.dev/docs/api/class-request#request-header-value)
- [headersArray](https://playwright.dev/docs/api/class-request#request-headers-array)
- [response](https://playwright.dev/docs/api/class-request#request-response)
- [sizes](https://playwright.dev/docs/api/class-request#request-sizes)

ankur22:
I've started work on implementing this API here: https://github.com/grafana/k6/pull/4290

I have tested the change with:

```js
import { browser } from 'k6/browser'

export const options = {
  scenarios: {
    ui: {
      executor: 'shared-iterations',
      options: {
        browser: {
          type: 'chromium',
        },
      },
    },
  },
}

export default async function () {
  const page = await browser.newPage()

  const requestCounter = (function() {
    let count = 0;
    return () => ++count;
  })();

  page.on('request', async (request) => {
    const currentCount = requestCounter();
    console.log(`${currentCount}: ${await request.allHeaders()}`);
    console.log(`${currentCount}: ${request.frame()}`);
[...]

PR #4290, opened 2025-01-28
Add `page.on('request')`

ankur22:
## What?

This implements the `page.on('request')` API from Playwright. When the handler is created, all subsequent requests will be directed to the handler, where the user can interact with the read only request. The request is intercepted just before it is sent from the browser to the website under test.

An example usage is:

```js
import { browser } from 'k6/browser'

export const options = {
  scenarios: {
    ui: {
      options: {
        browser: {
          type: 'chromium',
        },
      },
    },
  },
}

  const page = await browser.newPage()

  // Logs the url of the request
  page.on('request', async (request) => {
    console.log(request.url());
  })

  await page.goto('https://quickpizza.grafana.com/', { waitUntil: 'networkidle' })

  await page.close();
}
```

## Why?

This can be useful to:

* retrieve data that is to be used later;
* assert certain conditions on an expected request;

## Checklist

## Related PR(s)/Issue(s)

Updates: https://github.com/grafana/k6/issues/4281

### DIFF
--- a/internal/js/modules/k6/browser/common/network_manager.go
+++ b/internal/js/modules/k6/browser/common/network_manager.go
@@ -45,6 +45,7 @@ func (c Credentials) IsEmpty() bool {
 
 type metricInterceptor interface {
 	urlTagName(urlTag string, method string) (string, bool)
+	onRequest(request *Request)
 }
 
 // NetworkManager manages all frames in HTML document.
@@ -511,6 +512,8 @@ func (m *NetworkManager) onRequest(event *network.EventRequestWillBeSent, interc
 	m.reqsMu.Unlock()
 	m.emitRequestMetrics(req)
 	m.frameManager.requestStarted(req)
+
+	m.mi.onRequest(req)
 }
 
 func (m *NetworkManager) onRequestPaused(event *fetch.EventRequestPaused) {
--- a/internal/js/modules/k6/browser/common/network_manager.go
+++ b/internal/js/modules/k6/browser/common/network_manager.go
@@ -43,7 +43,7 @@ func (c Credentials) IsEmpty() bool {
 	return c == (Credentials{})
 }
 
-type metricInterceptor interface {
+type eventInterceptor interface {
 	urlTagName(urlTag string, method string) (string, bool)
 	onRequest(request *Request)
 }
@@ -52,16 +52,16 @@ type metricInterceptor interface {
 type NetworkManager struct {
 	BaseEventEmitter
 
-	ctx           context.Context
-	logger        *log.Logger
-	session       session
-	parent        *NetworkManager
-	frameManager  *FrameManager
-	credentials   Credentials
-	resolver      k6netext.Resolver
-	vu            k6modules.VU
-	customMetrics *k6ext.CustomMetrics
-	mi            metricInterceptor
+	ctx              context.Context
+	logger           *log.Logger
+	session          session
+	parent           *NetworkManager
+	frameManager     *FrameManager
+	credentials      Credentials
+	resolver         k6netext.Resolver
+	vu               k6modules.VU
+	customMetrics    *k6ext.CustomMetrics
+	eventInterceptor eventInterceptor
 
 	// TODO: manage inflight requests separately (move them between the two maps
 	// as they transition from inflight -> completed)
@@ -85,7 +85,7 @@ func NewNetworkManager(
 	s session,
 	fm *FrameManager,
 	parent *NetworkManager,
-	mi metricInterceptor,
+	ei eventInterceptor,
 ) (*NetworkManager, error) {
 	vu := k6ext.GetVU(ctx)
 	state := vu.State()
@@ -111,7 +111,7 @@ func NewNetworkManager(
 		attemptedAuth:    make(map[fetch.RequestID]bool),
 		extraHTTPHeaders: make(map[string]string),
 		networkProfile:   NewNetworkProfile(),
-		mi:               mi,
+		eventInterceptor: ei,
 	}
 	m.initEvents()
 	if err := m.initDomains(); err != nil {
@@ -179,7 +179,7 @@ func (m *NetworkManager) emitRequestMetrics(req *Request) {
 		tags = tags.With("method", req.method)
 	}
 	if state.Options.SystemTags.Has(k6metrics.TagURL) {
-		tags = handleURLTag(m.mi, req.URL(), req.method, tags)
+		tags = handleURLTag(m.eventInterceptor, req.URL(), req.method, tags)
 	}
 	tags = tags.With("resource_type", req.ResourceType())
 
@@ -233,7 +233,7 @@ func (m *NetworkManager) emitResponseMetrics(resp *Response, req *Request) {
 		tags = tags.With("method", req.method)
 	}
 	if state.Options.SystemTags.Has(k6metrics.TagURL) {
-		tags = handleURLTag(m.mi, url, req.method, tags)
+		tags = handleURLTag(m.eventInterceptor, url, req.method, tags)
 	}
 	if state.Options.SystemTags.Has(k6metrics.TagIP) {
 		tags = tags.With("ip", ipAddress)
@@ -281,7 +281,7 @@ func (m *NetworkManager) emitResponseMetrics(resp *Response, req *Request) {
 // handleURLTag will check if the url tag needs to be grouped by testing
 // against user supplied regex. If there's a match a user supplied name will
 // be used instead of the url for the url tag, otherwise the url will be used.
-func handleURLTag(mi metricInterceptor, url string, method string, tags *k6metrics.TagSet) *k6metrics.TagSet {
+func handleURLTag(mi eventInterceptor, url string, method string, tags *k6metrics.TagSet) *k6metrics.TagSet {
 	if newTagName, urlMatched := mi.urlTagName(url, method); urlMatched {
 		tags = tags.With("url", newTagName)
 		tags = tags.With("name", newTagName)
@@ -513,7 +513,7 @@ func (m *NetworkManager) onRequest(event *network.EventRequestWillBeSent, interc
 	m.emitRequestMetrics(req)
 	m.frameManager.requestStarted(req)
 
-	m.mi.onRequest(req)
+	m.eventInterceptor.onRequest(req)
 }
 
 func (m *NetworkManager) onRequestPaused(event *fetch.EventRequestPaused) {
--- a/internal/js/modules/k6/browser/common/page.go
+++ b/internal/js/modules/k6/browser/common/page.go
@@ -488,6 +488,32 @@ func (p *Page) urlTagName(url string, method string) (string, bool) {
 	return newTagName, urlMatched
 }
 
+func (p *Page) onRequest(request *Request) {
+	if !hasPageOnHandler(p, EventPageRequestCalled) {
+		return
+	}
+
+	p.eventHandlersMu.RLock()
+	defer p.eventHandlersMu.RUnlock()
+	for _, h := range p.eventHandlers[EventPageRequestCalled] {
+		err := func() error {
+			// Handlers can register other handlers, so we need to
+			// unlock the mutex before calling the next handler.
+			p.eventHandlersMu.RUnlock()
+			defer p.eventHandlersMu.RLock()
+
+			// Call and wait for the handler to complete.
+			return h(PageOnEvent{
+				Request: request,
+			})
+		}()
+		if err != nil {
+			p.logger.Warnf("onRequest", "handler returned an error: %v", err)
+			return
+		}
+	}
+}
+
 func (p *Page) onConsoleAPICalled(event *runtime.EventConsoleAPICalled) {
 	if !hasPageOnHandler(p, EventPageConsoleAPICalled) {
 		return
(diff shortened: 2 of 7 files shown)


internal/js/modules/k6/browser/tests/page_test.go
```diff
@@ -2182,3 +2183,295 @@ func TestPageOnMetric(t *testing.T) {
...
+		assert.Equal(t, expected[i].FrameURL, req.FrameURL, "FrameUrl mismatch")
+		assert.Equal(t, expected[i].AcceptLanguageHeader, req.AcceptLanguageHeader, "AcceptLanguageHeader mismatch")
+		assert.Equal(t, expected[i].Headers, req.Headers, "Headers mismatch")
+		assert.Equal(t, expected[i].IsNavigationRequest, req.IsNavigationRequest, "IsNavigationRequest mismatch")
+		assert.Equal(t, expected[i].Method, req.Method, "Method mismatch")
+		assert.Equal(t, expected[i].PostData, req.PostData, "PostData mismatch")
+		assert.Equal(t, expected[i].PostDataBuffer, req.PostDataBuffer, "PostDataBuffer mismatch")
+		assert.Equal(t, expected[i].ResourceType, req.ResourceType, "ResourceType mismatch")
+		assert.Equal(t, expected[i].Size, req.Size, "Size mismatch")
+		assert.Equal(t, expected[i].URL, req.URL, "URL mismatch")
```
inancgumus, line 2447: `assert.Equal` produces a nice failure diff that already shows the differing fields. Maybe we can compare `expected[i]` with `req` directly?

```diff
Diff:
--- Expected
+++ Actual
@@ -19,8 +19,8 @@
   (map[string]string) (len=2) {
+   (string) (len=4) "name": (string) (len=10) "User-Agent",
+   (string) (len=5) "value": (string) (len=15) "some-user-agent"
+  },
+  (map[string]string) (len=2) {
   (string) (len=4) "name": (string) (len=7) "Referer",
   (string) (len=5) "value": (string) (len=26) "http://127.0.0.1:8080/home"
-  },
-  (map[string]string) (len=2) {
-   (string) (len=4) "name": (string) (len=10) "User-Agent",
-   (string) (len=5) "value": (string) (len=15) "some-user-agent"
   }
```

Maybe we can compare `expected[i]` with `req` directly. The benefit is that: less code to read (the test is already large), and if we ever add/remove fields, we won't have to update the code except for the test cases. It's also easier to find the missing fields that we may never test.

internal/js/modules/k6/browser/tests/page_test.go
```diff
@@ -2182,3 +2183,295 @@ func TestPageOnMetric(t *testing.T) {
...
+			IsNavigationRequest: true,
+			Method:              "GET",
+			PostData:            "",
+			PostDataBuffer:      "",
+			ResourceType:        "Document",
+			Size: map[string]int{
+				"body":    0,
+				"headers": 103,
+			},
+			URL: "http://127.0.0.1:8080/home",
```
inancgumus, line 2332: Instead of a static URL that we need to transform (right above the `expected` variable), we can also use `tb.url` to get the server's URL:

```suggestion
			URL: tb.url("/home"),
```

inancgumus (commented):
Nice bit of work 👏 Some suggestions.

internal/js/modules/k6/browser/common/network_manager.go
```diff
@@ -43,24 +43,25 @@ func (c Credentials) IsEmpty() bool {
 	return c == (Credentials{})
 }
 
-type metricInterceptor interface {
+type eventInterceptor interface {
 	urlTagName(urlTag string, method string) (string, bool)
+	onRequest(request *Request)
```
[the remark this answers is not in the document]
ankur22: > Do we need to change this interface?

Using this interface makes it simpler to test `network_manager`. By working with the interface i was able to [mock](https://github.com/grafana/k6/blob/db87e57388ddd8b044b651b460563630cd1f1175/internal/js/modules/k6/browser/common/network_manager_test.go#L218) the calls for the network_manager.go tests, so i didn't have to take into account the calls to `onRequest` in `page`.

> I noticed that directly calling m.frameManager.page.onRequest(req) in NetworkManager.onRequest also works. What do you think?

Yeah, that's true, it just makes testing simpler and i don't need to update the network_manager tests.

Let me know if i misunderstood something.

internal/js/modules/k6/browser/common/network_manager.go
```diff
@@ -512,6 +513,8 @@ func (m *NetworkManager) onRequest(event *network.EventRequestWillBeSent, interc
 	m.reqsMu.Unlock()
 	m.emitRequestMetrics(req)
 	m.frameManager.requestStarted(req)
+
+	m.eventInterceptor.onRequest(req)
```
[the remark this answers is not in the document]
ankur22: Sure, i get where you are coming from. I think working with the interface and mock in the network_manager test is easier than having to ensure that the network_manager tests take into account the call to `onRequest` in `page`.

internal/js/modules/k6/browser/tests/page_test.go
ankur22: Yeah, I tried working with `expected[i]` and `req` directly when i first wrote the test. The `assert` library (and `DeepEqual`) didn't work when comparing the `HeadersArray` field. This is why i ended up comparing each field one by one. I may have missed something though.

internal/js/modules/k6/browser/tests/page_test.go
ankur22: Oh, nice suggestion!

internal/js/modules/k6/browser/tests/page_test.go
```diff
@@ -2182,3 +2183,286 @@ func TestPageOnMetric(t *testing.T) {
...
+      body: JSON.stringify({name: 'tester'})
+    })</script>
+</body>
+</html>`)
+		require.NoError(t, err)
+	})
+	tb.withHandler("/api", func(w http.ResponseWriter, r *http.Request) {
+		body, err := io.ReadAll(r.Body)
+		require.NoError(t, err)
+		defer require.NoError(t, r.Body.Close())
```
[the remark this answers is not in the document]
ankur22: Not sure i understand what you mean here. Why would `defer` call `r.Body.Close()` immediately?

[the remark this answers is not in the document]
ankur22: I made a change in the ordering of the call to `r.Body.Close()` in 0a519cc9038237bdd033e16a020c275c06043ccb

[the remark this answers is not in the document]
ankur22: Aha, right, i read the docs, thanks for the suggestion, fixed in 522a41135805237605bb3a62345537a96facd3f0

internal/js/modules/k6/browser/tests/page_test.go
```diff
@@ -2182,3 +2183,286 @@ func TestPageOnMetric(t *testing.T) {
...
+
+	// Convert the result to a string and then to a slice of requests.
+	var requests []request
+	err = json.Unmarshal([]byte(got.Result().String()), &requests)
+	require.NoError(t, err)
+
+	for i := range requests {
+		// Normalize any port numbers in the string values to :8080
+		if requests[i].URL != "" {
+			requests[i].URL = regexp.MustCompile(`:\d+`).ReplaceAllString(requests[i].URL, ":8080")
```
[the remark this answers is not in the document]
ankur22: Refactored in 6acd0320a05182a372a2c0d575ec30e71e16c5a4

internal/js/modules/k6/browser/common/network_manager.go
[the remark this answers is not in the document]
inancgumus: Yep, I see why you wanted to do it. Thanks for explaining. But, still, I'm wondering, all tests pass even if we remove the extra `onRequest` method from the `eventInterceptor` interface. Just having `urlTagName` seems enough. I've run the tests using this diff below. Am I missing something? 🤔

<details>
<summary>Without extra onRequest method</summary>

```diff
diff --git a/internal/js/modules/k6/browser/common/network_manager.go b/internal/js/modules/k6/browser/common/network_manager.go
index a1c336377..8c05bfbb4 100644
--- a/internal/js/modules/k6/browser/common/network_manager.go
+++ b/internal/js/modules/k6/browser/common/network_manager.go
@@ -45,7 +45,6 @@ func (c Credentials) IsEmpty() bool {
 
 type eventInterceptor interface {
 	urlTagName(urlTag string, method string) (string, bool)
-	onRequest(request *Request)
 }
 
 // NetworkManager manages all frames in HTML document.
@@ -513,8 +512,7 @@ func (m *NetworkManager) onRequest(event *network.EventRequestWillBeSent, interc
 	m.reqsMu.Unlock()
 	m.emitRequestMetrics(req)
 	m.frameManager.requestStarted(req)
-
-	m.eventInterceptor.onRequest(req)
+	m.frameManager.page.onRequest(req)
 }
 
 func (m *NetworkManager) onRequestPaused(event *fetch.EventRequestPaused) {
diff --git a/internal/js/modules/k6/browser/common/network_manager_test.go b/internal/js/modules/k6/browser/common/network_manager_test.go
index 7f0eef1bd..484c083ee 100644
--- a/internal/js/modules/k6/browser/common/network_manager_test.go
+++ b/internal/js/modules/k6/browser/common/network_manager_test.go
@@ -215,8 +215,6 @@ func (m *EventInterceptorMock) urlTagName(_ string, _ string) (string, bool) {
 	return "", false
 }
 
-func (m *EventInterceptorMock) onRequest(request *Request) {}
-
 func TestNetworkManagerEmitRequestResponseMetricsTimingSkew(t *testing.T) {
 	t.Parallel()
 
```

</details>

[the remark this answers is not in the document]
ankur22: Right, i thought there was a test for `NetworkManager.onRequest` but didn't actually check in `common/network_manager_test.go`. The change you present does work.

However, I like that this "side affect" is decoupled from everything (`NetworkManager` isn't aware that the `onRequest` can be called from `m.frameManager.page`), which feels like a rarity in k6 browser and one of the reason it's so difficult to write unit tests for it. Maybe we need a sit down to discuss a better long term strategy on decoupling types.

I still think this is a benefit as it is, but if you have a strong opinion i'll change it.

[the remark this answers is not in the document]
inancgumus: TBH, I'm not a fan of introducing a layer (interface) of abstraction when there are no multiple implementations of an interface and/or if we do it only for testing (in this case, it doesn't even benefit testing) since it makes it harder to understand the code (we need to jump on multiple hops to see how it works). I'd personally remove the interface method, but the decision to keep it as it is yours, of course. I'm fine with that. I'm just trying to point out that we don't currently need it.

internal/js/modules/k6/browser/tests/page_test.go
```diff
@@ -2182,3 +2182,273 @@ func TestPageOnMetric(t *testing.T) {
...
+
+	// Compare each request one by one for better test failure visibility
+	for _, req := range requests {
+		i := -1
+		for j, e := range expected {
+			if req.URL == e.URL {
+				i = j
+				break
+			}
+		}
```
inancgumus, line 2411: We can shorten this to:

```suggestion
		i := slices.IndexFunc(expected, func(r request) bool { return req.URL == r.URL })
		assert.NotEqual(t, -1, i, "failed to find expected request with URL %s", req.URL)
```

inancgumus (commented):
One more non-critical thing I noticed.

inancgumus (dismissed):

internal/js/modules/k6/browser/tests/page_test.go
ankur22: Very nice! Refactored in 42481333556f747b8c54efc0910330f0fcb72771

internal/js/modules/k6/browser/tests/page_test.go
ankur22: TIL! Thanks 😄  refactored in 8d5b9d11217c403e3f2f4d5ef8ffbc078e51af21

internal/js/modules/k6/browser/common/page.go
```diff
@@ -485,6 +488,32 @@ func (p *Page) urlTagName(url string, method string) (string, bool) {
...
+	}
+
+	p.eventHandlersMu.RLock()
+	defer p.eventHandlersMu.RUnlock()
+	for _, h := range p.eventHandlers[EventPageRequestCalled] {
+		err := func() error {
+			// Handlers can register other handlers, so we need to
+			// unlock the mutex before calling the next handler.
+			p.eventHandlersMu.RUnlock()
+			defer p.eventHandlersMu.RLock()
```
codebien, line 503: Why do we need to lock again? Isn't the lock before already sage-guarding for it?

ankur22: Yeah, this looks a bit odd. The reason behind this is to allow the handler to be able to add more handlers. This is a behaviour that Playwright exhibits and also documents, so it's something we're replicating.

```js
page.on('request', async () => {
    page.on('response', async () => {
        // Do something with the request and response data...
    })
})
```

inancgumus: @codebien I asked the same question before for another related `page.on` method. Here's a little bit more detail @ankur22 previously answered: https://github.com/grafana/xk6-browser/pull/1456#discussion_r1786305452

codebien:
@inancgumus can you rebase and resolve the conflicts here, please? :pray:

codebien (approved):

inancgumus (approved):

Issue #4284, opened 2025-01-27
Option to parsing CSV into array of objects keyed by the field name


going-confetti:
### Background 
We are adding support for data files to k6 Studio (https://github.com/grafana/k6-studio/issues/363) and we would like to discuss a feature request.

In short, the user will be able to import a data file (CSV/JSON array) and use values from this file in Parameterization rules (e.g. `Pick random row from users.csv and replace id in the request with the value of user_id from that row`). From the UX perspective, the file type won't matter much and the UI of the parameterization rule editor will look something like this: 

<img width="530" alt="Image" src="https://github.com/user-attachments/assets/617ecf6e-02ea-4b72-977d-49be069d5740" />

What will be different is the actual file parsing code in the generated script. For CSVs, we plan on using `k6/experimental/csv` in the actual generated script, but the current return format of the `parse` function (array of string arrays) makes extraction of the specified value harder and means that we probably need different snippets for JSON and CSV.

### What we're asking
Ideally, we'd like an option similar to `header: true` in `papaparse`:

```csv
id,name
1,hdb
```

```js
const csvRecordsArr = await csv.parse(file, { delimiter: ',' }) // [['id', 'name'], ['1', 'hdb']]
const csvRecordsObj = await csv.parse(file, {delimiter: ',', header: true}) // [{ id: '1', name: 'hdb' }]
```

### Already existing or connected issues / PRs (optional)

https://github.com/grafana/k6-studio/issues/363

going-confetti:
@oleiade we had a brief discussion about this the other day

oleiade:
Thanks for opening this @going-confetti 🙇🏻

PR #4295, opened 2025-01-29
Add `header` option support to the CSV module

oleiade:
## What?

Adds support for a `header` option to the csv module's option. When set to `true`, the `header` option leads the module's `parse` function as well as the `Parser.next` method to return records as objects with the header corresping row name as the property key, and value as the value.

```javascript
const file = await open('data.csv');
const csvRecords = await csv.parse(file, { header: true })

	console.log(csvRecords[scenario.iterationInTest])
}
```

**Would print:**
```
INFO[0000] {"age":"72","firstname":"fariha","lastname":"ehlenfeldt"}  source=console
INFO[0000] {"age":"50","firstname":"qudratullah","lastname":"gillfillan"}  source=console
INFO[0000] {"age":"41","firstname":"jeleah","lastname":"rodina"}  source=console
INFO[0000] {"age":"99","firstname":"thaisia","lastname":"nowalk"}  source=console
INFO[0000] {"age":"75","firstname":"joey-lynn","lastname":"wilsford"}  source=console
INFO[0000] {"age":"81","firstname":"tudur","lastname":"granville"}  source=console
INFO[0000] {"age":"56","firstname":"aytek","lastname":"umber"}  source=console
INFO[0000] {"age":"30","firstname":"aynoor","lastname":"hisaw"}  source=console
INFO[0000] {"age":"31","firstname":"fiadh-rose","lastname":"iravani"}  source=console
INFO[0000] {"age":"70","firstname":"annely","lastname":"ooley"}  source=console
```

As opposed to, for reference, treating records as array without the header option set:
```
INFO[0000] ["firstname","lastname","age"]                source=console
INFO[0000] ["fariha","ehlenfeldt","72"]                  source=console
INFO[0000] ["qudratullah","gillfillan","50"]             source=console
INFO[0000] ["jeleah","rodina","41"]                      source=console
INFO[0000] ["thaisia","nowalk","99"]                     source=console
INFO[0000] ["joey-lynn","wilsford","75"]                 source=console
INFO[0000] ["tudur","granville","81"]                    source=console
INFO[0000] ["aytek","umber","56"]                        source=console
INFO[0000] ["aynoor","hisaw","30"]                       source=console
INFO[0000] ["fiadh-rose","iravani","31"]                 source=console
```

**Note**: that the same behavior can be observed when using a `Parser`'s `next` method with the `header` option set to `true`.

## Why?

This PR tackles #4284, and directly addresses a request from the team behind k6 studio, and one of the project's issues [k6-studio/#363](https://github.com/grafana/k6-studio/issues/363).

It also nicely complements the feature set of the csv parser in general, and aligns its options-set with [papaparse's](https://www.papaparse.com/docs).

## Checklist

## Related PR(s)/Issue(s)

[k6/#4284](https://github.com/grafana/issues/4284)
[k6-studio/#363](https://github.com/grafana/k6-studio/issues/363)

### DIFF
--- a/internal/js/modules/k6/experimental/csv/module_test.go
+++ b/internal/js/modules/k6/experimental/csv/module_test.go
@@ -621,6 +621,7 @@ func TestParse(t *testing.T) {
 
 			if (csvRecords.length !== 10) {
 				throw new Error("Expected 10 records, but got " + csvRecords.length);
+			}
 
 			for (const record of csvRecords) {
 				if (typeof record !== 'object' || record === null || Array.isArray(record)) {
--- a/internal/js/modules/k6/experimental/csv/module_test.go
+++ b/internal/js/modules/k6/experimental/csv/module_test.go
@@ -79,6 +79,44 @@ func TestParserConstructor(t *testing.T) {
 		require.NoError(t, err)
 	})
 
+	t.Run("constructing a parser with both header and skipFirstLine options should fail", func(t *testing.T) {
+		t.Parallel()
+
+		r, err := newConfiguredRuntime(t)
+		require.NoError(t, err)
+
+		// Ensure the testdata.csv file is present on the test filesystem.
+		r.VU.InitEnvField.FileSystems["file"] = newTestFs(t, func(fs fsext.Fs) error {
+			return fsext.WriteFile(fs, testFilePath, []byte(csvTestData), 0o644)
+		})
+
+		_, err = r.RunOnEventLoop(wrapInAsyncLambda(fmt.Sprintf(`
+			  const file = await fs.open(%q);
+			  const parser = new csv.Parser(file, { delimiter: ';', skipFirstLine: true, header: true });
+		`, testFilePath)))
+
+		require.Error(t, err)
+	})
+
+	t.Run("constructing a parser with both the header option and fromLine option greater than 0 should fail", func(t *testing.T) {
+		t.Parallel()
+
+		r, err := newConfiguredRuntime(t)
+		require.NoError(t, err)
+
+		// Ensure the testdata.csv file is present on the test filesystem.
+		r.VU.InitEnvField.FileSystems["file"] = newTestFs(t, func(fs fsext.Fs) error {
+			return fsext.WriteFile(fs, testFilePath, []byte(csvTestData), 0o644)
+		})
+
+		_, err = r.RunOnEventLoop(wrapInAsyncLambda(fmt.Sprintf(`
+			  const file = await fs.open(%q);
+			  const parser = new csv.Parser(file, { delimiter: ';', fromLine: 1, header: true });
+		`, testFilePath)))
+
+		require.Error(t, err)
+	})
+
 	t.Run("constructing a parser without providing a file instance should fail", func(t *testing.T) {
 		t.Parallel()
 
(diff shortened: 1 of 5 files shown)


codebien:
>Adds support for a header option to the csv module's option.

I find the DX introduced here not optimal. To be honest, without reading the docs, my intuition would be that `headers: true` just includes the headers row. 

I would prefer a clearer name, something like `headerAsFields`. Or, overall, I would prefer a different approach by using a different method. For example, `parseAsObject()`.

oleiade:
I see where you're coming from @codebien, and I think I agree, I reused the name paparse uses, but in hindsight, I also find it a bit confusing 👍🏻 

Giving this a thought, I propose we rename it to `asObjects` instead, and document explicitly that the header will be used as the "mapping" reference. I would prefer it to `headerAsFields` that doesn't express enough that the output type changes 🙇🏻

oleiade:
Also heads-up @going-confetti 👋🏻

codebien:
>Giving this a thought, I propose we rename it to asObjects instead

Yes, this is better. I like it.

Now, another question as a devil's advocate. Are we sure that it shouldn't be the default option? Why do we think array of arrays is better?

allansson:
> > Giving this a thought, I propose we rename it to asObjects instead

```ts
// users.csv
// Username,First name,Last name
// jdoe,john,doe
// immicky,mickey,mouse

const arrayOfArrays = csv.parse("./users.csv", { rowsAs: "array" })
const arrayOfObjects = csv.parse("./users.csv", { rowsAs: "object" })

// Future possibility:
const customMapping = csv.parse("./users.csv", {
  // Might seem similar to just doing `array.map` but this would avoid having to
  // iterate the list twice.
  rowsAs: ([userName, firstName, lastName]) => {
    return {
      userName,
      fullName: `${firstName} ${lastName}`
    }
  }
})
```

> @oleiade Now, another question as a devil's advocate. Are we sure that it shouldn't be the default option? Why do we think an array of arrays is better for the default?

"If only this was an array of arrays", said no one ever.

internal/js/modules/k6/experimental/csv/reader.go
```diff
@@ -96,20 +84,80 @@ func NewReaderFrom(r io.Reader, options options) (*Reader, error) {
...
+		}
+
+		if len(record) != len(r.header) {
+			return nil, fmt.Errorf("record length (%d) doesn't match header length (%d)", len(record), len(r.header))
+		}
+
+		recordMap := make(map[string]string)
+		for i, value := range record {
+			recordMap[r.header[i]] = value
+		}
```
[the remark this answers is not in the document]
oleiade: Initially the whole design idea was to effectively separate the layer that reads the data (pure-Go-centric) from the layer that produces the values we provide and expose to Sobek. Hence, the current way this is done, and hence, I would rather not be keen to return a `sobek.Object` here directly: to preserve the separation of concern.

In terms of optimization, in case this is called by `Parser.Next`, this most likely will be a single allocation every time, and in the even of a `parse` call, where we parse the whole file and then produce a new shared array using the `NewSharedArrayFrom` we introduced, every row is gonna end up going through `json.Marshal`, in which case I don't believe it changes much if it's a `map` or a `sobek.Object` (which I guess, would then need to be potentially transformed back to be serializable?).

All of this to say unless you have a strong objection, I'd like to stick to this approach for now, and reevaluate if we observe issues in the future 🙇🏻

oleiade:
@codebien @allansson 

I think the approach you folks suggest makes a ton of sense, looking around about this, I actually noticed that Deno's 2.0 csv module took a similar approach:

```typescript
import { parse, stringify } from "jsr:@std/csv";
let text = `
url,views,likes
https://deno.land,10,7
https://deno.land/x,20,15
https://deno.dev,30,23
`;
let data = parse(text, {
  skipFirstRow: true,
  strip: true,
});
console.log(data[0].url); // https://deno.land
console.log(data[0].views); // 10
console.log(data[0].likes); // 7
text = `
https://deno.land,10,7
https://deno.land/x,20,15
https://deno.dev,30,23
`;
data = parse(text, {
  columns: ["url", "views", "likes"],
});
console.log(data[0].url); // https://deno.land
console.log(data[0].views); // 10
console.log(data[0].likes); // 7
const obj = [
  { mascot: "dino", fans: { old: 100, new: 200 } },
  { mascot: "bread", fans: { old: 5, new: 2 } },
];
const csv = stringify(obj, {
  columns: [
    "mascot",
    ["fans", "new"],
  ],
});
console.log(csv);
// mascot,new
// dino,200
// bread,2
```

Note that by default rows are treated as maps (arrays are also objects in js after all), with the option to provide the column names yourself if you want. I find it aligns pretty well with your example @allansson too. 

I really like this direction, but I suggest we open an issue, and address such a redesign separately, because that would be a big chunk of work, and as to not block this PR which would bring immediate value already. 

What's your take, do you folks agree, or would you say we should completely change the design itself?

(bringing @Llandy3d in the loop to give him visibility on this)

codebien:
>What's your take, do you folks agree, or would you say we should completely change the design itself?

@oleiade I'm fine with it. Please, remember to open an issue so we can link it into the epic issue to stabilize the csv module.

oleiade:
I have created #4507 as a result @codebien 🙇🏻 (with a couple of todos in the issue that are design aspects we should look into during the implementation)

codebien (approved):

internal/js/modules/k6/experimental/csv/module_test.go
```diff
@@ -527,6 +603,38 @@ func TestParse(t *testing.T) {
...
+			return fsext.WriteFile(fs, testFilePath, []byte(csvTestData), 0o644)
+		})
+
+		_, err = r.RunOnEventLoop(wrapInAsyncLambda(fmt.Sprintf(`
+			const file = await fs.open(%q);
+			const csvRecords = await csv.parse(file, { asObjects: true });
+
+			if (csvRecords.length !== 10) {
+				throw new Error("Expected 11 records, but got " + csvRecords.length);
+			}
```
inancgumus, line 624: WDYT?

```suggestion
			if (csvRecords.length !== 10) {
				throw new Error("Expected 10 records, but got " + csvRecords.length);
```

inancgumus (dismissed):
LGTM 👍 Just one suggestion.

internal/js/modules/k6/experimental/csv/module_test.go
oleiade: Good catch 🎉 Cheers!

codebien (approved):

inancgumus (approved):

PR #4297, opened 2025-01-29
fix: race condition when registering grpc types

### LINKED ISSUE #3961: gRPC using reflect race on registering types
### Brief summary

We have tests that use `constant-arrival-rate` executor against a number of grpc applications.  We are using reflection.  Most of these run fine.  Against some applications, we occasionally receive a panic:

```
panic: proto: message proto.com.XXXXX.XXXXX.XXXXX.MyService1 is already registered
```
This is not completely deterministic.  However, for applications where it occurs, the probability increases as we increase `rate`.

As a workaround we are switching these tests to `ramping-arrival-rate`, but ideally we would like to be able to use `constant-arrival-rate`.

(I've obfuscated actual host and service names in all output)

### k6 version

v0.53.0 

### OS

Mariner 2.0.20240628

n/a

version:

```
$ k6 --version
k6 v0.53.0 (go1.23.1, linux/amd64)
Extensions:
  github.com/grafana/xk6-sql v0.4.1, k6/x/sql [js]
  github.com/mostafa/xk6-kafka v0.27.0, k6/x/kafka [js]
  github.com/thotasrinath/xk6-couchbase v0.0.8, k6/x/couchbase [js]
```

OS:
```
$ cat lsb-release 
DISTRIB_ID="Mariner"
DISTRIB_RELEASE="2.0.20240628"
DISTRIB_CODENAME=Mariner
DISTRIB_DESCRIPTION="CBL-Mariner 2.0.20240628"

$ cat mariner-release 
CBL-Mariner 2.0.20240628
MARINER_BUILD_NUMBER=baf1849

$ uname -a
Linux XXXX-XX999.XXX.XXXX.com 5.15.160.1-1.cm2 #1 SMP Fri Jun 28 10:31:37 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux
```

Simple test script.   See attached for script and output from several runs.  Note the panic line may show different grpc endpoints, it doesn't always show the same one (I gather this is related to reflection).

```
$ /path/to/k6/intel/k6 run   /tmp/XXXXX/k6.js  

          /\      |‾‾| /‾‾/   /‾‾/   
     /\  /  \     |  |/  /   /  /    

     execution: local
        script: /tmp/XXXXX/k6.js
        output: -

     scenarios: (100.00%) 1 scenario, 50 max VUs, 35s max duration (incl. graceful stop):
              * constant-qps: 20000.00 iterations/s for 5s (maxVUs: 50, gracefulStop: 30s)

WARN[0000] Insufficient VUs, reached 50 active VUs and cannot initialize more  executor=constant-arrival-rate scenario=constant-qps
panic: proto: message proto.com.XXXX.XXXX.XXXX.SomeServiceName is already registered
```

### Expected behaviour

Test executes.

### Actual behaviour

See attached output files.
 [k6.js.txt](https://github.com/user-attachments/files/17048244/k6.js.txt)
[k6.out.1.txt](https://github.com/user-attachments/files/17048247/k6.out.1.txt)
[k6.out.2.txt](https://github.com/user-attachments/files/17048246/k6.out.2.txt)
[k6.out.3.txt](https://github.com/user-attachments/files/17048245/k6.out.3.txt)

abemedia:
## What?

Adds a mutex to avoid race condition when registering proto types. 

## Why?

When registering proto types we first check if they exist. With multiple tests running concurrently the check could result in false for multiple tests which then simultaneously try to register the type, resulting in the following error:

```
panic: proto: message XXX is already registered
```
See the following issue:
- #3961

## Checklist

## Related PR(s)/Issue(s)

Closes #3961


oleiade:
Hi @abemedia 👋🏻 

Apologies for coming back to this real late 🙇🏻 

We appreciate your PR, however, I'm not super satisfied with the current solution this PR brings to the issue #3961. The main reason for it is because it relies on a global variable mutex to address it; which, I find is not a great practice. The mutex is a good idea, the locking/unlocking too, but my main issue with the current approach is that the mutex should live somewhere else if possible.

It turns out this is a somewhat common pattern in k6, as we spawn concurrent VUs, and often need to provide guarantees of concurrent access. Thus, if we keep the mutext approach, I would prefer to try and make the mutex  live in the `RootModule` as defined in the `grpc.go` instead, and ensure that we cascade the locking logic down to where it's meant to be used: that might imply the grpc Client to keep a reference to the module somehow too.

Otherwise, just an intuition, but I wonder if maybe `sync.Once` might turn out to be a somewhat functional approach in this situation too?

Finally, if possible, it would be great to be able to test this, if possible, and I acknowledge that might be tricky. But I'm definetly not terribly well versed in gRPC; so this might be a stab in the dark.

@inancgumus any other ideas/approaches/suggestions? 😸 

Hope this is helpful, and let us know if you have any further questions, or need further support with this PR 🙇🏻

inancgumus:
> @inancgumus any other ideas/approaches/suggestions? 😸

@oleiade I agree with everything you mentioned 🤝 For the `RootModule` approach, the k6 browser module follows a similar method. @abemedia, you can find details about this approach [here](https://github.com/grafana/k6/blob/e8e93e5a004b89747e620a63f229d7fda63a693b/internal/js/modules/k6/browser/browser/module.go#L33-L43). Thanks for your contribution 🙇

PR #4298, opened 2025-01-30
Unit tests for config file operations

codebien:
## What?

It adds a series of unit tests to cover operations on the config file on disk storage.   

## Why?

It creates a safer context for applying changes for #2508


internal/cmd/config_test.go
```diff
@@ -202,3 +206,199 @@ func TestDeriveAndValidateConfig(t *testing.T) {
...
+	conf := []byte(`bad json format`)
+	defaultConfigPath := ".config/loadimpact/k6/config.json"
+	require.NoError(t, fsext.WriteFile(memfs, defaultConfigPath, conf, 0o644))
+
+	gs := &state.GlobalState{
+		FS:    memfs,
+		Flags: state.GetDefaultFlags(".config"),
+	}
+	_, err := readDiskConfig(gs)
+	require.ErrorContains(t, err, "couldn't parse")
```
inancgumus, line 308: We can also compare the error type instead of string matching (since we wrap the error in the code), but it's also fine as it is:

```suggestion
	var serr *json.SyntaxError
	assert.ErrorAs(t, err, &jerr)
```

internal/cmd/config_test.go
```diff
@@ -202,3 +206,199 @@ func TestDeriveAndValidateConfig(t *testing.T) {
...
+	}
+
+	c := Config{
+		WebDashboard: null.BoolFrom(true),
+		Options: lib.Options{
+			Cloud: []byte(`invalid-json`),
+		},
+	}
+	err := writeDiskConfig(gs, c)
+	assert.ErrorContains(t, err, "json: error")
```
inancgumus, line 403: Same suggestion from before:

```suggestion
	var serr *json.SyntaxError
	assert.ErrorAs(t, err, &jerr)
```

inancgumus (approved):
LGTM 👍 I'd group some tests into table tests, but it's a matter of style.

codebien:
>I'd group some tests into table tests, but it's a matter of style.

@inancgumus my feeling is that there are differences between tests that will make readability worst if we are going to do that as table tests. I can do a refactor attempt in a dedicated pull request, but I'll look into it after we merged https://github.com/grafana/k6/pull/4301 :+1:

mstoykov (approved):

oleiade (approved):

PR #4301, opened 2025-01-30
New default path for config file

codebien:
## What?

It changes the default path for config file to `.config/k6/config.json` from the legacy `loadimpact` version.

## Why?

Check https://github.com/grafana/k6/issues/2508

## Related PR(s)/Issue(s)

Closes https://github.com/grafana/k6/issues/2508


internal/cmd/config.go
```diff
@@ -303,3 +335,34 @@ func validateScenarioConfig(conf lib.ExecutorConfig, isExecutable func(string) b
...
+
+// migrateLegacyConfigFileIfAny moves the configuration file from
+// the old default `~/.config/loadimpact/...` folder
+// to the new `~/.config/k6/...` default folder.
+func migrateLegacyConfigFileIfAny(gs *state.GlobalState) error {
+	fn := func() error {
+		legacyFpath := legacyConfigFilePath(gs)
+		if _, err := gs.FS.Stat(legacyFpath); errors.Is(err, fs.ErrNotExist) {
+			return nil
+		} else if err != nil {
```
inancgumus, line 347: No need for `else if` here.

internal/cmd/config.go
```diff
@@ -151,6 +151,28 @@ func readDiskConfig(gs *state.GlobalState) (Config, error) {
 	return conf, nil
 }
 
+func legacyConfigFilePath(gs *state.GlobalState) string {
+	return filepath.Join(gs.UserOSConfigDir, "loadimpact", "k6", "config.json")
+}
```
inancgumus, line 156: Why is this a function? It seems it was used only once.

internal/cmd/config.go
```diff
@@ -303,3 +335,34 @@ func validateScenarioConfig(conf lib.ExecutorConfig, isExecutable func(string) b
 	}
 	return nil
 }
+
+// migrateLegacyConfigFileIfAny moves the configuration file from
+// the old default `~/.config/loadimpact/...` folder
+// to the new `~/.config/k6/...` default folder.
+func migrateLegacyConfigFileIfAny(gs *state.GlobalState) error {
```
inancgumus, line 342: Can this be a part of `readDiskConfig` (e.g., calling this function from readDiskConfig)? Do we need to know the details of how we're getting the config file?

inancgumus (dismissed):
LGTM 👍 Feel free to skip my suggestions if they don't make sense.

internal/cmd/config.go
```diff
@@ -187,10 +209,20 @@ func readEnvConfig(envMap map[string]string) (Config, error) {
...
+			return conf, errext.WithExitCodeIfNone(err, exitcodes.InvalidConfig)
+		}
+	} else if err != nil {
 		return conf, errext.WithExitCodeIfNone(err, exitcodes.InvalidConfig)
+	} else {
+		gs.Logger.Warn("The configuration file has been found on the old path. " +
+			"Please, run again `k6 cloud login` or `k6 login` commands to migrate to the new path. " +
+			"If you already migrated it manually, then remove the file from the old path.\n\n")
 	}
+
```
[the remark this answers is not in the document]
codebien: >gocritic is right here.

I don't see how. We have to dump the warning line only if the file has been found AND it is legacy. I could reduce indentation by using `if err == nil` but I think it creates more complicated code instead of simplifying it.

>I'd also return Config{} instead of conf although we know it's an empty Config. The code currently reads like conf is not empty.

Done

internal/cmd/config.go
codebien: Good catch, it is now reused from the other part of the code where it is needed.

internal/cmd/config.go
codebien: Done

internal/cmd/config.go
```diff
@@ -186,17 +210,27 @@ func readEnvConfig(envMap map[string]string) (Config, error) {
...
+	if errors.Is(err, fs.ErrNotExist) { //nolint:gocritic
+		fileConf, err = readDiskConfig(gs)
+		if err != nil {
+			return Config{}, errext.WithExitCodeIfNone(err, exitcodes.InvalidConfig)
+		}
+	} else if err != nil {
+		return Config{}, errext.WithExitCodeIfNone(err, exitcodes.InvalidConfig)
+	} else {
+		gs.Logger.Warn("The configuration file has been found on the old path. " +
+			"Please, run again `k6 cloud login` or `k6 login` commands to migrate to the new path. " +
```
joanlopez, line 224:
```suggestion
			"Please, run again `k6 cloud login` commands to migrate to the new path. " +
```

As `k6 login` has been deprecated, I think it's better to not mention it. Right? 🤔

internal/cmd/config.go
```diff
@@ -303,3 +337,36 @@ func validateScenarioConfig(conf lib.ExecutorConfig, isExecutable func(string) b
...
+			return err
+		}
+
+		err = gs.FS.Rename(legacyFpath, gs.Flags.ConfigFilePath)
+		if err != nil {
+			return err
+		}
+
+		gs.Logger.Infof("Note, the configuration file has been migrated "+
+			"from old default path (%q) to the new version (%q).\n\n", legacyFpath, gs.Flags.ConfigFilePath)
```
joanlopez, line 365:
```suggestion
			"from the old default path (%q) to the new one (%q).\n\n", legacyFpath, gs.Flags.ConfigFilePath)
```

joanlopez (commented):
Overall it looks mostly fine, except for a few minor things I pointed out during the review.

However, I have the general doubt around when should we warn users and when should we perform the configuration migration. Isn't file disk configuration also used for other commands like `run`, `archive` or `cloud run`?

Why do we only consider it for `k6 cloud login`? 
Do users normally don't store anything beyond cloud configuration? 🤔

codebien:
Hey @joanlopez, thanks for reviewing :wave: 

>However, I have the general doubt around when should we warn users and when should we perform the configuration migration. Isn't file disk configuration also used for other commands like run, archive or cloud run?

I applied mostly what was suggested here https://github.com/grafana/k6/issues/2508#issuecomment-1111988506. It makes sense for me because it keeps the same operations' flows. k6 **writes** the config on `k6 login` operations, and it **reads** on all the others. I wouldn't break this principle, so it migrates (write) on `k6 login` and it warns (reads) on all the others.

In addition, re-reading Ned's comment, I realized I'm applying a reversed logic on the reading path, where we read before the legacy config and only after the new. But there isn't really a requirement for doing it, we can just directly read the new path and only if it doesn't exist then do an additional attempt on the legacy.

Does it make sense for you? The goal here is to provide the best UX so if you have any suggestion to improve it, I'm happy to evaluate them.

internal/cmd/config.go
codebien: If a user has a config file because it used `k6 login influxdb` but they don't have the Cloud they won't be able to migrate anymore. Do you think we should not support them?

internal/cmd/config.go
```diff
@@ -186,17 +210,27 @@ func readEnvConfig(envMap map[string]string) (Config, error) {
...
+		fileConf, err = readDiskConfig(gs)
+		if err != nil {
+			return Config{}, errext.WithExitCodeIfNone(err, exitcodes.InvalidConfig)
+		}
+	} else if err != nil {
+		return Config{}, errext.WithExitCodeIfNone(err, exitcodes.InvalidConfig)
+	} else {
+		gs.Logger.Warn("The configuration file has been found on the old path. " +
+			"Please, run again `k6 cloud login` or `k6 login` commands to migrate to the new path. " +
+			"If you already migrated it manually, then remove the file from the old path.\n\n")
```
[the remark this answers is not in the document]
codebien: I'm considering to reverse the flow, where we always check the new default before, and only eventually we check the old one. I didn't do that before because the code is a bit tricky in the current version because we are silencing the Not Found error. But I realized it's better to stretch a bit the code to fit it instead of yielding this cognitive load on the user.

internal/cmd/config.go
inancgumus: Nice 🎉

internal/cmd/config.go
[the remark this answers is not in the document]
inancgumus: The last change you made looks clear to me. Thanks.

codebien:
@inancgumus @joanlopez I have a doubt. During the migration, do you think should we delete the old directories (`loadimpact/k6`) or should we just leave them there to prevent any additional error?

internal/cmd/config.go
joanlopez: Mmm.. not sure. Let's leave it for now, then. Although I guess that `k6 login influxdb` will eventually disappear.

codebien:
Hey @inancgumus @joanlopez I've done most of the discussed changes, take another look, please. :pray: 
Make sure to test directly building the binary on your machines, as I'm not on OSX then we might discover bugs not visible to me on Linux.

inancgumus (approved):
LGTM 👍 It works on OS X 15.3, too.

joanlopez:
> LGTM 👍 It works on OS X 15.3, too.

Also tested on MacOS 15.3 and it also worked for me! 👍🏻

Issue #4139, opened 2024-12-19
Secrets Source extensions


mstoykov:
## What?

Implement a way to get secrets from extensible sources and hide them in logs.

This is implementation proposal for https://github.com/grafana/k6/issues/3207 

#### Configuration

new cli flag `--secret-source` to add a secret source to k6. 

Which takes a secret source name, and identifier and additional configuration.
`--secret-source=sourcename=identifier,option1=value1,option2=value2`

### JS API

(the exist import is under consideration)

```javascript
import secrets from "k6/secrets";  
  
export default async () => {  
    const my_secret = await secrets.get("my_secret"); // get from 1 secret source if only 1, exception if more than 1 secret source or if no secret sources
    console.log(my_secret);  
}
```

```javascript
import secrets from "k6/secrets";  
  
export default async () => {  
    const my_secret = await secrets.source("identifier").get("my_secret"); // get secret from a source with the provided identifier
    console.log(my_secret);
}
```

No other functionality is currently planned.

### Go implementation details

Pluggable similar to output and js extensions. With minimal configuration.

Centralized log sanitization - k6 will see each secret passing and will start sanitizing logs for those values it has seen. This does mean that secret sources have less responsibilities and it is less likely for bugs.

##### File based on
This is mostly for ease of use for development, but it just reads a text file with key=value pairs

##### REST API based on
For now very simple one with a url + headers, potentially with templating for the *key* of the secret requests. Potentially other options later

PR #4514, opened 2025-02-05
Secret Source Implementation

mstoykov:
## What?

Adds an extensible secret sources. Basically #4139 but without http secret source as that turned out to have too many configuration problems. But it has a mock secret source which basically gets them form the cli arguments - probably should be renamed.

Usage:

```javascript
import secrets from "k6/secrets";

export default async () => {
	const my_secret = await secrets.get("cool"); // get secret from a source with the provided identifier
	console.log(my_secret);
	await secrets.get("else"); // get secret from a source with the provided identifier
	console.log(my_secret);
}
```

```
go run . run --secret-source=mock=cool=some,else=source script.js
```
image

or with a secret file:

```
cool=some
else=source
```

```
go run . run --secret-source=file=file.secret script.js
```

The extension points are under ./secretsource and mostly around the `Get` method which gets a secret and returns it.

The rest of k6 makes certain to not request it again and redact it from logs.

TODO:
- [x] Should we throw an exception or return `undefined` on errors, or at least on not found secrets. I currently think having an exception lets sources have specific errors potentially helping with debugging. But maybe logging the error is good enough :shrug: 
- [X] Multiple secret sources is almost done but not quite
- [X] Unit Tests and likely some integration ones will be nice - although I do plan to add some more.
- [ ] ~Have HTTP secret source? Maybe for later version~ this seems to have way too many moving pieces to be a good idea for first versions
- [ ] Configuration currently only happens from CLI flags which might not be ideal ~~and is definitely problematic as `,` is special, so you need ot use some other character for separation like `:` in the mock configuration.~~
- [x] Having `Name` be parsed by the internal parts *before* we actually make the secret source
- [x] Having a way to define a Default secret source when named ones are also possible.
- [ ] Some Benchmarks will be nice
- [X] Docs of any kind  
- [x] Moving to `k6/secrets` seems like a better idea - both easier to track if it used, but also will stop adding stuff to `"k6"` a module that mostly has stuff that ... are not great.
- [ ] usage statistics on used secret source
- [ ] Use Description and log which secret sources are used a kin to outputs

## Checklist

### DIFF
--- a/cmd/state/state.go
+++ b/cmd/state/state.go
@@ -14,7 +14,9 @@ import (
 
 	"go.k6.io/k6/internal/event"
 	"go.k6.io/k6/internal/ui/console"
+	"go.k6.io/k6/internal/usage"
 	"go.k6.io/k6/lib/fsext"
+	"go.k6.io/k6/secretsource"
 )
 
 const defaultConfigFileName = "config.json"
@@ -55,6 +57,9 @@ type GlobalState struct {
 
 	Logger         *logrus.Logger //nolint:forbidigo //TODO:change to FieldLogger
 	FallbackLogger logrus.FieldLogger
+
+	SecretsManager *secretsource.SecretsManager
+	Usage          *usage.Usage
 }
 
 // NewGlobalState returns a new GlobalState with the given ctx.
@@ -131,6 +136,7 @@ func NewGlobalState(ctx context.Context) *GlobalState {
 			Hooks:     make(logrus.LevelHooks),
 			Level:     logrus.InfoLevel,
 		},
+		Usage: usage.New(),
 	}
 }
 
@@ -142,6 +148,7 @@ type GlobalFlags struct {
 	Address          string
 	ProfilingEnabled bool
 	LogOutput        string
+	SecretSource     []string
 	LogFormat        string
 	Verbose          bool
 }
--- a/cmd/state/state.go
+++ b/cmd/state/state.go
@@ -58,7 +58,7 @@ type GlobalState struct {
 	Logger         *logrus.Logger //nolint:forbidigo //TODO:change to FieldLogger
 	FallbackLogger logrus.FieldLogger
 
-	SecretsManager *secretsource.SecretsManager
+	SecretsManager *secretsource.Manager
 	Usage          *usage.Usage
 }
 
--- a/internal/cmd/root.go
+++ b/internal/cmd/root.go
@@ -19,7 +19,11 @@ import (
 	"go.k6.io/k6/cmd/state"
 	"go.k6.io/k6/errext"
 	"go.k6.io/k6/errext/exitcodes"
+	"go.k6.io/k6/ext"
 	"go.k6.io/k6/internal/log"
+	"go.k6.io/k6/secretsource"
+
+	_ "go.k6.io/k6/internal/secretsource" // import it to register internal secret sources
 )
 
 const waitLoggerCloseTimeout = time.Second * 5
@@ -162,6 +166,10 @@ func rootCmdPersistentFlagSet(gs *state.GlobalState) *pflag.FlagSet {
 	// `gs.DefaultFlags.<value>`, so that the `k6 --help` message is
 	// not messed up...
 
+	// TODO(@mstoykov): likely needs work - no env variables and such. No config.json.
+	flags.StringArrayVar(&gs.Flags.SecretSource, "secret-source", gs.Flags.SecretSource,
+		"setting secret sources for k6 file[=./path.fileformat],")
+
 	flags.StringVar(&gs.Flags.LogOutput, "log-output", gs.Flags.LogOutput,
 		"change the output for k6 logs, possible values are stderr,stdout,none,loki[=host:port],file[=./path.fileformat]")
 	flags.Lookup("log-output").DefValue = gs.DefaultFlags.LogOutput
@@ -257,6 +265,22 @@ func (c *rootCommand) setupLoggers(stop <-chan struct{}) error {
 		c.globalState.Logger.Debug("Logger format: TEXT")
 	}
 
+	secretsources, err := createSecretSources(c.globalState)
+	if err != nil {
+		return err
+	}
+	// it is important that we add this hook first as hooks are executed in order of addition
+	// and this means no other hook will get secrets
+	var secretsHook logrus.Hook
+	c.globalState.SecretsManager, secretsHook, err = secretsource.NewSecretsManager(secretsources)
+	if err != nil {
+		return err
+	}
+	if len(secretsources) != 0 {
+		// don't actually filter anything if there will be no secrets
+		c.globalState.Logger.AddHook(secretsHook)
+	}
+
 	cancel := func() {} // noop as default
 	if hook != nil {
 		ctx := context.Background()
@@ -288,4 +312,50 @@ func (c *rootCommand) setLoggerHook(ctx context.Context, h log.AsyncHook) {
 	}()
 	c.globalState.Logger.AddHook(h)
 	c.globalState.Logger.SetOutput(io.Discard) // don't output to anywhere else
+}
+
+func createSecretSources(gs *state.GlobalState) (map[string]secretsource.SecretSource, error) {
+	baseParams := secretsource.Params{
+		Logger:      gs.Logger,
+		Environment: gs.Env,
+		FS:          gs.FS,
+		Usage:       gs.Usage,
+	}
+
+	result := make(map[string]secretsource.SecretSource)
+	for _, line := range gs.Flags.SecretSource {
+		t, config, ok := strings.Cut(line, "=")
+		if !ok {
+			return nil, fmt.Errorf("couldn't parse secret source configuration %q", line)
+		}
+		secretSources := ext.Get(ext.SecretSourceExtension)
+		found, ok := secretSources[t]
+		if !ok {
+			return nil, fmt.Errorf("no secret source for type %q for configuration %q", t, line)
+		}
+		c := found.Module.(secretsource.Constructor) //nolint:forcetypeassert
+		params := baseParams
+		params.ConfigArgument = config
+		// TODO(@mstoykov): make it not configurable just from cmd line
+		// params.JSONConfig = test.derivedConfig.Collectors[outputType]
+
+		secretSource, err := c(params)
+		if err != nil {
+			return nil, err
+		}
+		name := secretSource.Name()
+		_, alreadRegistered := result[name]
+		if alreadRegistered {
+			return nil, fmt.Errorf("secret source for name %q already registered before configuration %q", t, line)
+		}
+		result[name] = secretSource
+	}
+
+	if len(result) == 1 {
+		for _, l := range result {
+			result["default"] = l
+		}
+	}
+
+	return result, nil
 }
--- a/internal/cmd/root.go
+++ b/internal/cmd/root.go
@@ -335,18 +335,24 @@ func createSecretSources(gs *state.GlobalState) (map[string]secretsource.Source,
 		}
 		c := found.Module.(secretsource.Constructor) //nolint:forcetypeassert
 		params := baseParams
+		name, isDefault, config := extractNameAndDefault(config)
 		params.ConfigArgument = config
 
 		secretSource, err := c(params)
 		if err != nil {
 			return nil, err
 		}
-		name := secretSource.Name()
 		_, alreadRegistered := result[name]
 		if alreadRegistered {
 			return nil, fmt.Errorf("secret source for name %q already registered before configuration %q", t, line)
 		}
 		result[name] = secretSource
+		if isDefault {
+			if _, ok := result["default"]; ok {
+				return nil, fmt.Errorf("can't have two secret sources that are default ones, second one was %q", config)
+			}
+			result["default"] = secretSource
+		}
 	}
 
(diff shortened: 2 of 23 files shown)


internal/log/secrets.go
```diff
@@ -0,0 +1,72 @@
...
+func recursiveReplace(v any, replacer *strings.Replacer) any {
+	switch s := v.(type) {
+	case string:
+		return replacer.Replace(s)
+	case int, uint, int64, int32, int16, int8, uint64, uint32, uint16, uint8:
+		// if the secret is encodable in 64 bits ... it is probably not a great secret
+		return v
+	}
+	// replace this with a log after more testing
+	panic(fmt.Sprintf("Had a logrus.fields value with type %T, please report that this is unsupported", v))
```
d0ugal, line 71: I got this:

> panic: Had a logrus.fields value with type time.Duration, please report that this is unsupported

mstoykov: [f738100](https://github.com/grafana/k6/pull/4514/commits/f738100962d833bde31bb6c1f4b8de28f99cc7bb) should fix this case

internal/cmd/root.go
```diff
@@ -19,7 +19,11 @@ import (
 	"go.k6.io/k6/cmd/state"
 	"go.k6.io/k6/errext"
 	"go.k6.io/k6/errext/exitcodes"
+	"go.k6.io/k6/ext"
 	"go.k6.io/k6/internal/log"
+	"go.k6.io/k6/secretsource"
+
+	_ "go.k6.io/k6/internal/secretsource" // import it to register internal secret sources
```
mstoykov, line 26: This is contrast with how it is done for outputs or js modules. 

I kind of prefer this to having a function with all of them in one place which gets constant conflicts.

But I can redo it to be more aligned with other parts of k6.

internal/cmd/root.go
```diff
@@ -162,6 +166,10 @@ func rootCmdPersistentFlagSet(gs *state.GlobalState) *pflag.FlagSet {
 	// `gs.DefaultFlags.<value>`, so that the `k6 --help` message is
 	// not messed up...
 
+	// TODO(@mstoykov): likely needs work - no env variables and such. No config.json.
+	flags.StringArrayVar(&gs.Flags.SecretSource, "secret-source", gs.Flags.SecretSource,
+		"setting secret sources for k6 file[=./path.fileformat],")
+
```
mstoykov, line 172: I would prefer to give this a go in a separate PR, maybe in the next version.

internal/cmd/root.go
```diff
@@ -289,3 +313,49 @@ func (c *rootCommand) setLoggerHook(ctx context.Context, h log.AsyncHook) {
...
+		secretSources := ext.Get(ext.SecretSourceExtension)
+		found, ok := secretSources[t]
+		if !ok {
+			return nil, fmt.Errorf("no secret source for type %q for configuration %q", t, line)
+		}
+		c := found.Module.(secretsource.Constructor) //nolint:forcetypeassert
+		params := baseParams
+		params.ConfigArgument = config
+		// TODO(@mstoykov): make it not configurable just from cmd line
+		// params.JSONConfig = test.derivedConfig.Collectors[outputType]
```
mstoykov, line 340:
```suggestion
```

already mentioned in a different place

secretsource/doc.go
```diff
@@ -0,0 +1,4 @@
+// Package secretsource is a package to provide secret source interface and common functionality
+// This functionality is to be used to provide k6 with a way to get secrets and help it handl them correctly.
+// Predominantly by not redact them from logs.
```
mstoykov, line 3:
```suggestion
// Package secretsource is a package to provide secret source interface and common functionality
// This functionality is to be used to provide k6 with a way to get secrets and help it handle them correctly.
// Predominantly by redacting them from logs.
```

secretsource/interface.go
```diff
@@ -0,0 +1,8 @@
+package secretsource
+
+// SecretSource is the interface a secret source needs to implement
+type SecretSource interface {
+	Name() string
```
mstoykov, line 5: Ideally this and whether a secret source is the default will be parsed from the configuration from the manager before they are provided to the constructor.

secretsource/hook.go
```diff
@@ -0,0 +1,69 @@
...
+	case string:
+		return replacer.Replace(s)
+	case int, uint, int64, int32, int16, int8, uint64, uint32, uint16, uint8:
+		// if the secret is encodable in 64 bits ... it is probably not a great secret
+		return v
+	case time.Duration:
+		return v
+	}
+	// replace this with a log after more testing
+	panic(fmt.Sprintf("Had a logrus.fields value with type %T, please report that this is unsupported", v))
```
mstoykov, line 68: Reminder for me

cmd/state/state.go
```diff
@@ -55,6 +57,9 @@ type GlobalState struct {
 
 	Logger         *logrus.Logger //nolint:forbidigo //TODO:change to FieldLogger
 	FallbackLogger logrus.FieldLogger
+
+	SecretsManager *secretsource.SecretsManager
```
inancgumus, line 61: I'll continue to review this PR, but for now, I find these are a bit of a mouthful and repetitive:

- `secretsource.SecretsManager`
- `secretsource.SecretSource`

What do you think about:

- `secret.LogRedacter` (more descriptive)
- `secret.Source` (easy to type, read, and remember)

Also, does the `secretsource.SecretSource` interface need all these methods? Why not?

```go
package secret

type Provider struct {
    Name        string
    Description string
    // more extensible without breaking the interface
}

type Source interface {
	Provider() Provider
	Get(key string) (value string, err error)
}
```

Sidenote: I'm also not sure if we need `Name()` in the original interface because we use it to look up if the secret source provider is registered. We could also check through its concrete typename (through the interface) to see if it's registered.

mstoykov: I am not against renaming some of the stuff, but:
1. I think LogRedactor isn't more descriptive. The idea is more that it manages the secrets and is later used to get them back. It also facilitating redaction of the logs is not it's main function.
2. I kind of do not like really small names such as `secret`. So I prefer to keep that as `secretsource`. 

So `secretsource.Manager` and `secretsource.Source` seems good enough.


On the struct part ... I am not against it, but not certain it is really better. 

Also of note is that :
1. Name is actually the name of the source to be used in case of multiple secret sources. This likely will be easier to get from the manager or something else preparsing the configurations. This seems to not be what people think it is as I have another datapoint where people think it is the name of the secret source as type.
2. Description isn't currently used, and I guess I should add adding the secrets to the logs around `outputs` and such to the TODO.

inancgumus: > I think LogRedactor isn't more descriptive. The idea is more that it manages the secrets and is later used to get them back. It also facilitating redaction of the logs is not it's main function.

`Manager` sounds more like `Registry` then? `Manager` is also fine, but it's generic. It also sounds like it has multiple responsibilities: registering secret sources and attaching a logger. Splitting these might make responsibilities clearer.

> I kind of do not like really small names such as secret. So I prefer to keep that as secretsource.

I see that we already have a package called `secrets` (a module). Since you already have `k6/secrets`, I believe using `secret` here might be clearer, shorter, and more idiomatic. But `secretsource` is also fine since you prefer to keep it that way.

> So secretsource.Manager and secretsource.Source seems good enough.

I see that you changed `SecretSource` to `Source`; that seems better 👍 If you're open, I still think `Provider` might communicate intent clearer. `Source` is fine, but there's still slight stuttering (`secretsource.Source`). Not critical, just mentioning.

> On the struct part ... I am not against it, but not certain it is really better.
> Description isn't currently used, and I guess I should add adding the secrets to the logs around outputs and such to the TODO.

I totally get your point about descriptions for logs/CLI. But interfaces usually work best when minimal—only essential behavior. One small idea could be to store descriptions separately at registration (rather than directly in the interface). But if that feels overly complex, sticking to your approach is okay, too. Maybe like:

```go
type ... struct {
    Source      Source
    Description string
}
```

Anyway, these are all just suggestions. I don't want to push you into anything, and I'm fine with whatever you decide. Totally up to you, of course 🙂

oleiade (dismissed):
Great work 👏🏻  I really like the shape this takes, and I had a good time experimenting with it.

I'm happy with the current state of things, and although I have just a couple of comments, I approve the PR, as I don't consider any of those blocking whatsoever.

* I really like the secret source module approach 👍🏻
* I've found that the introduction of the `Usage` wiring was a bit unnecessary. You actually left comment around that. No strong opinion, but it could make sense to me if it was in a separate PR indeed.
* The first test run I've made with PR, I forgot to pass the secret source argument, and I noticed that k6 produced an error message which I found could be more explicit:
CleanShot 2025-03-05 at 12 31 20
In this case, as a user I would have expected a message along the lines of "The test script tries to access secrets, but no source where defined, make sure to use the command-line argument --etc..."
* on the checklist item above "Should we throw an exception or return undefined on errors, or at least on not found secrets. I currently think having an exception lets sources have specific errors potentially helping with debugging. But maybe logging the error is good enough 🤷": I tend to think exceptions would make sense here 🙇🏻

secretsource/hook.go
```diff
@@ -0,0 +1,69 @@
...
+	// replace both keys and values with
+	for k, v := range entry.Data {
+		newk := replacer.Replace(k)
+		if newk != k {
+			entry.Data[newk] = v
+			delete(entry.Data, k)
+			k = newk
+		}
+		entry.Data[k] = recursiveReplace(v, replacer)
+	}
```
d0ugal, line 52: I'm not sure about this in Go, so just asking. Is it safe to modify the keys of a map while iterating it or should a copy/new map be created?

mstoykov: Yes it is fine here is some [SO answer with relevant links to the  specification](https://stackoverflow.com/a/68639401/5427244)

In this case as we are removing the key we are on - that is fine, we already iterated over it so nothing will be missed.

We also add a key if we remove one, in that (likely very rare case) we might or might not iterate over the new key in this going over. If we do not - great, if  we do, I guess there is a case where if the secret matches some part of  `SECRET_REDACTED` we can go in some kind of strange loop. 

All of this though seems very unlikely and ... I would expect nothing like `SECRET_REDACTED` or `***` to be a secret in a real world case. 

Making copy of the map for each log line logged though seems like quite a considerable increase in processing and memory usage.

So IMO there is an extremely unlikely scenario something might go wrong here, but it will not be a real world case and the real world case of us copying the maps will likely have way worse performance.

d0ugal: That makes sense. Thanks for the detailed answer!

internal/js/modules/k6/secrets/secrets.go
```diff
@@ -0,0 +1,89 @@
...
+
+// Exports returns the exports of the k6 module.
+func (mi *Secrets) Exports() modules.Exports {
+	s, err := mi.secrets()
+	if err != nil {
+		common.Throw(mi.vu.Runtime(), err)
+	}
+	return modules.Exports{
+		Default: s,
+		Named:   make(map[string]any), // this is intentially not nil so it doesn't export anything as named expeorts
```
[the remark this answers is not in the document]
mstoykov: fixed in [38e7d06](https://github.com/grafana/k6/pull/4514/commits/38e7d0678696111a59681bfec51d0d09c9187433)

mstoykov:
@oleiade I've changed the original message and added an additional one in case there is nothing configured.

I do not really want to be too specific on stuff given k6 direction in a more generic solution. So I do not want to return errors that have stuff such as `script` in there, given that secrets can be used by not just from js and so on. 

Arguably we can have better errors if we add even more types for those and handle them in the runner or something like this.

oleiade:
@mstoykov Cheers! I find the new error message is much more explicit, and I agree with your point on keeping it general/generic 🙇🏻

oleiade (approved):

cmd/state/state.go
mstoykov: I would like @oleiade opinion on this as well, but also please review the remaining part of the PR. 

From my perspective:
1. The manager is not a registry - the registry is in different place. It is structure that manages secret sources and returns secrets from them. We use the same name for (metric) outputs
2. `k6/secrets` is named that as that is what the user uses it for - it is to get secrets. They do not explicitly work with secret sources. Maybe we will add other functionality around secrets there one day.
3. `secretsource` is named that as this is the only thing the package does - it defines secretsources and their interface and how to work with them. The people working with the two packages are for different use cases, even if they are connected. Having a package `secrets` that than has `Source` interface inside sounds great, but I would argue that the package name will not be telling what is inside it - there are not secrets there, just an interface for Sources.

inancgumus: I believe we're looking at this from different perspectives because I see packages as providers of functionality rather than containers of functionality. So, I consider the `secret` (or `secrets`) package to be a provider of `secret.Source` functionality or its protocol.

I find "manager" very generic and can (and probably will) have many responsibilities beyond its purposes. That's what usually happens to manager types. It's like a "utility" name. Even now, it seems like it has two responsibilities, a registry of secret sources and a hook adder. Maybe this hooking behavior should not be its responsibility.

Anyway, like I said, I don't want to block this PR just because of semantic discussions. I will approve it since I don't see any issues with the code 🤝 ❤️

inancgumus (approved):
There are some issues I pointed out in [this comment](https://github.com/grafana/k6/pull/4514#discussion_r1974047746). Other than that it's LGTM. Good work! 👍

PR #4519, opened 2025-02-07
Merge Prometheus remote write output

codebien:
## What?

It migrates the codebase of Prometheus remote-write metrics output from the external dedicated xk6 repository to k6 repository.

## Why?

It is the first step to graduate Prometheus remote-write as a stable metrics output.


internal/output/prometheusrw/remote/client_test.go
```diff
@@ -0,0 +1,229 @@
...
+	"net/http"
+	"net/http/httptest"
+	"net/url"
+	"testing"
+	"time"
+
+	"go.k6.io/k6/internal/output/prometheusrw/stale"
+
+	prompb "buf.build/gen/go/prometheus/prometheus/protocolbuffers/go"
+	"github.com/golang/snappy"
```
mstoykov, line 16: This likely should be the klaus snappy instead - which is already used through the project. This PR now adds this as a dependancy while it didn't before. Specifically because of this test.

codebien: Good catch :handshake:

oleiade (approved):

Issue #4219, opened 2023-10-12
Move Sobek entirely out of the business logic


inancgumus:
### What?

Move Sobek entirely out of the business logic.

### Why?

* Prevent Sobek data races.
* Better type-safety using native Go constructs.
* Developer productivity increase.
* Separation of concerns.
* Moving Sobek handling to the mapping layer will make the code easy to reason about, reduce the business logic's Sobek dependency, and simplify the business logic code and tests.
* Having Sobek in the business logic requires constant encoding and decoding.

### Desobekifying Sobek transformation

Besides moving the calling of Sobek option parsing in the business logic, we also plan to move the Sobek-dependent option parsing part out of the business logic. Once we finish doing the same with the rest of the option parsing, we can entirely remove the Sobek dependency from the business logic.

We currently parse Sobek values into our business logic types, but the parsing dependency is in an unrelated package (common). To enhance maintainability, we move the parsing process and keep related components together (in the mapping layer). Additionally, we could implement a `Validate` method for the business types (in the common package) to check their validity separately from the Sobek transformation.

Moreover, we use pointers quite liberally, which is not good in terms of reduced performance and also because of many nil-pointer panics. Also, since our module is highly current, it's dangerous to have so many pointers around. So, another goal would be to get rid of as many pointer-ed options as possible to boost stability even more.

See grafana/xk6-browser#1510 for an example.

### Tasks

```[tasklist]
### Tasks
- [ ] https://github.com/grafana/xk6-browser/issues/1272
- [ ] https://github.com/grafana/xk6-browser/issues/1271
- [ ] https://github.com/grafana/xk6-browser/issues/1273
- [ ] https://github.com/grafana/xk6-browser/issues/1274
- [ ] https://github.com/grafana/xk6-browser/issues/1525
- [ ] https://github.com/grafana/xk6-browser/issues/1270
- [ ] https://github.com/grafana/xk6-browser/issues/1282
- [ ] https://github.com/grafana/xk6-browser/issues/1522
- [ ] grafana/xk6-browser#1393
- [ ] https://github.com/grafana/xk6-browser/issues/1269
- [ ] https://github.com/grafana/xk6-browser/issues/1275
- [ ] https://github.com/grafana/xk6-browser/issues/1277
- [ ] https://github.com/grafana/xk6-browser/issues/1281
- [ ] https://github.com/grafana/xk6-browser/pull/1228
- [ ] https://github.com/grafana/xk6-browser/pull/1198
```

[...]

PR #4521, opened 2025-02-10
browser: Mark a bunch of places where races happen

mstoykov:
## What?

In a lot of places in the k6ext.Promise will create a separate goroutine and then access the arguments in it. This isn't safe and will race with anything running on the event loop.

It is very likely that this happens very rarely especially as in a bunch of the cases this is only for `opts` - which seems to be less used in most cases.

Additionally in a bunch of cases mapping back to the sobek.Values is done in that same goroutine instead of on the event loop.

## Why?

I am marking all of those in one commit and not fixing them all as that will take a lot longer and likely a bunch of them will not be the same as the rest.

## Checklist


codebien (approved):
/cc @inancgumus

joanlopez (approved):

inancgumus:
Just to document/link issues and PRs.

#4219

Issue #4168, opened 2025-01-09
Issue Loading Local Modules when using `k6 cloud run –local-execution`


bijopunnoose:
### Brief summary

This bug created from the community page: https://community.grafana.com/t/issue-loading-local-modules-when-using-k6-cloud-run-local-execution/136919
The issue is affecting me as well as many other users. Kindly prioritise accordingly. 

I have created a series of K6 test that are working fine when I run them with k6 run MY_TEST.js, and these test are importing a couple local modules including one named config.js

But when I try to run the same test with k6 cloud run MY_TEST.js --local-execution I am getting:

error="GoError: the module \"./config.js\" was not previously resolved during initialization (__VU==0)\n"
I am unclear if there is something I am missing here or if this is running in some other directory when you use cloud run but my assumption would be the local execution flag would make it so it should run in the exact some way as k6 run would.

Not sure if it matters but I am running this on windows and any help would be appreciated.

### k6 version

v0.55.2

### OS

windows 10

### Steps to reproduce the problem

This bug created from the community page: https://community.grafana.com/t/issue-loading-local-modules-when-using-k6-cloud-run-local-execution/136919
The issue is affecting me as well as many other users. Kindly prioritise accordingly. 

I have created a series of K6 test that are working fine when I run them with k6 run MY_TEST.js, and these test are importing a couple local modules including one named config.js

But when I try to run the same test with k6 cloud run MY_TEST.js --local-execution I am getting:

error="GoError: the module \"./config.js\" was not previously resolved during initialization (__VU==0)\n"
I am unclear if there is something I am missing here or if this is running in some other directory when you use cloud run but my assumption would be the local execution flag would make it so it should run in the exact some way as k6 run would.

Not sure if it matters but I am running this on windows and any help would be appreciated.

### Expected behaviour

Should run same as k6 cloud run MY_TEST.js

### Actual behaviour

error="GoError: the module \"./config.js\" was not previously resolved during initialization (__VU==0)\n"
I am unclear if there is something I am missing here or if this is running in some other directory when you use cloud run but my assumption would be the local execution flag would make it so it should run in the exact some way as k6 run would.

oleiade:
Apologies for the late reply 🙇🏻 

I can confirm I was able to reproduce. 

**Test entrypoint script:**
```javascript
import { config } from "./config.js";

export const options = {
    cloud: {
        projectID: 123456,
    }
}

export default function main() {
  console.log(config.foo);
}
```

**Test config script**:
```javascript
[...]

PR #4532, opened 2025-02-10
Clone Bundle's URLs into lib.Archive

joanlopez:
## What?

It modifies the `Bundle.makeArchive` function so `Bundle`'s URLs (pointers) aren't directly reused by the `lib.Archive` it constructs, but cloned ("deep copy") and set.

## Why?

Because the `Archive.Write` method normalizes and anonymizes these URLs to prevent certain details (like OS's user, HOME dir, etc) to be leaked to the GCk6, but we not always want these modifications to be propagated to the `Bundle`'s URLs - e.g. when running the test script locally: `k6 cloud ... --local-execution`.

## Review notes

A better explanation of why this (#4168) happens is because the process goes as follows:
1. The test is loaded and configured, which includes the initialization of the first runner, and thus the cache of the module resolutions.
2. The process of creating the test remotely also involves the creation of the archive, which modifies the URLs, as described above.
3. Later, when the test is executed, the module resolution fails because it points to a different, non-cached, import path, due to the step 2.

Note this started to fail since we introduced this change because the old way of doing the same was `k6 run -o cloud`, and in that case we were not sending any archive at all. Now, the archive is an opt-out "feature", enabled by default.

Also, note that another approach to fix this issue (#4168) could consist on something like "reinstantiating" the first runner, so the "new paths" are cached again, and/or to somehow make it all work with the normalized and anonymized URLs.

However, that sounded a bit counterintuitive to me, plus I suspected that may cause other issue later (like k6 errors/warning not pointing to the correct paths locally, etc). 

While doing a "deep copy" of these URLs to "isolate" changes on `Archive`s sounded good to me, because I could not  think of any scenario where that could be a problem, and logically it kinda makes sense to make the `Archive` "standalone".

### Checklist

## Related PR(s)/Issue(s)

Fixes https://github.com/grafana/k6/issues/4168

### DIFF
--- a/internal/js/bundle.go
+++ b/internal/js/bundle.go
@@ -164,13 +164,16 @@ func NewBundleFromArchive(piState *lib.TestPreInitState, arc *lib.Archive) (*Bun
 }
 
 func (b *Bundle) makeArchive() *lib.Archive {
+	clonedSourceDataURL, _ := url.Parse(b.sourceData.URL.String())
+	clonedPwdURL, _ := url.Parse(b.pwd.String())
+
 	arc := &lib.Archive{
 		Type:              "js",
 		Filesystems:       b.filesystems,
 		Options:           b.Options,
-		FilenameURL:       b.sourceData.URL,
+		FilenameURL:       clonedSourceDataURL,
 		Data:              b.sourceData.Data,
-		PwdURL:            b.pwd,
+		PwdURL:            clonedPwdURL,
 		Env:               make(map[string]string, len(b.preInitState.RuntimeOptions.Env)),
 		CompatibilityMode: b.CompatibilityMode.String(),
 		K6Version:         consts.Version,


internal/js/bundle.go
```diff
@@ -164,13 +164,16 @@ func NewBundleFromArchive(piState *lib.TestPreInitState, arc *lib.Archive) (*Bun
 }
 
 func (b *Bundle) makeArchive() *lib.Archive {
+	clonedSourceDataURL, _ := url.Parse(b.sourceData.URL.String())
+	clonedPwdURL, _ := url.Parse(b.pwd.String())
+
 	arc := &lib.Archive{
 		Type:              "js",
 		Filesystems:       b.filesystems,
```
joanlopez, line 172: Other fields here, like this `Filesystems` are also passed by reference (note it's a `map`), and thus are likely to be modified. However, that doesn't seem to cause any issue so far.

mstoykov (approved):

mstoykov:
Nice catch @joanlopez :tada:

olegbespalov (approved):

olegbespalov:
Looks good! Would it be possible to introduce a test case for that (maybe in separated PR)?

oleiade:
Just passing by to send a huge kudos for fixing this @joanlopez 🚀

PR #4524, opened 2025-02-10
browser: Fix options races

codebien:
## What?

It expands https://github.com/grafana/k6/pull/4522.

It fixes the following funcs:
- `getAttribute`
- `hover`
- `innerHTML`
- `innerText`
- `inputValue`

## Why?

Updates https://github.com/grafana/k6/issues/4085

PR #4533, opened 2025-02-10
 browser: Fix options races p5

codebien:
## What?

Continue https://github.com/grafana/k6/pull/4524

### DIFF
--- a/internal/js/modules/k6/browser/common/frame.go
+++ b/internal/js/modules/k6/browser/common/frame.go
@@ -1474,15 +1474,8 @@ func (f *Frame) selectOption(selector string, values sobek.Value, opts *FrameSel
 }
 
 // SetContent replaces the entire HTML document content.
-func (f *Frame) SetContent(html string, opts sobek.Value) error {
+func (f *Frame) SetContent(html string, _ *FrameSetContentOptions) error {
 	f.log.Debugf("Frame:SetContent", "fid:%s furl:%q", f.ID(), f.URL())
-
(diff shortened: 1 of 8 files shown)


mstoykov (approved):

internal/js/modules/k6/browser/common/frame.go
```diff
@@ -1478,16 +1474,9 @@ func (f *Frame) selectOption(selector string, values sobek.Value, opts *FrameSel
 }
 
 // SetContent replaces the entire HTML document content.
-func (f *Frame) SetContent(html string, opts sobek.Value) error {
+func (f *Frame) SetContent(html string, _ *FrameSetContentOptions) error {
```
olegbespalov, line 1477: Sounds like here (and maybe upstream) we can clean up the opts completely? :thinking:

codebien: I don't know, for this reason, I would prefer if we defer this to the next release. In my opinion, it's better to coordinate with @inancgumus on similar refactors.

codebien: @inancgumus do you have any plan in mind for it? Can we remove it or do we expect to add features where it is needed?

inancgumus: It's better to keep it and add a TODO comment instead because here, we (the early-browser team) forgot to respect the options.

inancgumus: https://github.com/grafana/k6/pull/4538

PR #4534, opened 2025-02-10
browser: fix some mappings and remove overzealous TODOs

mstoykov:
## What?
More fixes for browser mappings and removing some overzealous TODOs
## Why?

While trying to fix some mappings that originally I recognized as places that races are possible, I found that is not the case.

As mappings are just a map that is methods/fields to their values, as long as no usage the runtime happens outside of any mapping it is fine.

This is the case for most cases, this fixes the one that isn't.

## Checklist


inancgumus (approved):
> As mappings are just a map that is methods/fields to their values, as long as no usage the runtime happens outside of any mapping it is fine.

Exactly. Thanks 👍

joanlopez (approved):

PR #4535, opened 2025-02-11
Avoid hardcoding signtool PATH in package-windows build step

### LINKED ISSUE #4105: Make the `signtool` usage in windows-packaging less brittle
This looks really brittle :thinking:

_Originally posted by @olegbespalov in https://github.com/grafana/k6/pull/4102#discussion_r1881815314_
           

The current way to sign windows executables uses `signtool` which is part of windows kit. Unfortuantely it isn't in the path by default and the path to the exact location is dependant on the version of the windows kit.

https://github.com/grafana/k6/blob/ba6a71ea63aca6b2976a34631c5b26d3c52272a5/.github/workflows/build.yml#L246-L247

Moving from windows-2019 to windows-2022 github action image meant that the path chagned and we had to fix it.

Given the information I found there doesn't seem to be a fix for this. If we ran from the terminal inside visual studio that sets the path correctly ... which doesn't seem like a good thing for a CI ;). 

Luckily there aren't that many windows releases so hopefully we need to fix this once every 3-4 years, but it will still be nice if it isn't as brittle as it is.

McMastS:
## What?
Searches the `Windows Kit\bin` directory for the most recent `signtool` executable, instead of hardcoding the exact version.

This PR also moves the `Add signtool to PATH` build step to right above the signing step, and only runs in the same conditions used for the signing step. The condition is added because the search adds around 20 seconds, so it shouldn't be run for all builds.

[Successful manual test showing signtool was called](https://github.com/McMastS/k6/actions/runs/13255968023/job/37002839657)
ManualTest

## Why?

The path to the `signtool` executable recently had to be [changed ](https://github.com/grafana/k6/pull/4102) because Windows moved its directory. This change means the `signtool` path won't need to be manually updated every time the Windows version is upgraded.

## Checklist

- [x] I have added tests for my changes. # tested manually in CI on my fork

## Related PR(s)/Issue(s)
Closes #4105

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -268,10 +268,10 @@ jobs:
           $SignTool = Get-ChildItem -Path "${env:ProgramFiles(x86)}\Windows Kits\10\bin" -Recurse -Filter signtool.exe | Where-Object { $_.DirectoryName -like "*\x64" } | Sort-Object -Descending | Select-Object -First 1
           
           # Sign the Windows binary
-          $SignTool sign /f k6.pfx /p "${{ secrets.WIN_SIGN_PASS }}" /tr "http://timestamp.digicert.com" /td sha256 /fd sha256 "packaging\k6.exe"
+          & $SignTool sign /f k6.pfx /p "${{ secrets.WIN_SIGN_PASS }}" /tr "http://timestamp.digicert.com" /td sha256 /fd sha256 "packaging\k6.exe"
 
           # Sign the MSI package
-          $SignTool sign /f k6.pfx /p "${{ secrets.WIN_SIGN_PASS }}" /tr "http://timestamp.digicert.com" /td sha256 /fd sha256 "packaging\k6.msi"
+          & $SignTool sign /f k6.pfx /p "${{ secrets.WIN_SIGN_PASS }}" /tr "http://timestamp.digicert.com" /td sha256 /fd sha256 "packaging\k6.msi"
 
           # Cleanup signing artifacts
           del k6.pfx
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -243,8 +243,6 @@ jobs:
           Expand-Archive -Path ".\dist\k6-$env:VERSION-windows-amd64.zip" -DestinationPath .\packaging\
           move .\packaging\k6-$env:VERSION-windows-amd64\k6.exe .\packaging\
           rmdir .\packaging\k6-$env:VERSION-windows-amd64\
-      - name: Add signtool to PATH
-        run: echo "${env:ProgramFiles(x86)}\Windows Kits\10\bin\10.0.17763.0\x64" | Out-File -FilePath $env:GITHUB_PATH -Append
 
       - name: Create the MSI package
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -254,8 +252,18 @@ jobs:
           candle.exe -arch x64 "-dVERSION=$env:VERSION" k6.wxs
           light.exe -ext WixUIExtension k6.wixobj
 
+      - name: Add signtool to PATH
+        # Searching for the signtool executable adds around 10 seconds to the
+        # job, but it avoids hardcoding the path to the signtool executable.
+        # We only run this step if we are building from master or a version tag,
+        # or if the workflow was manually triggered, to avoid the extra time.
+        if: ${{ github.ref == 'refs/heads/master' || startsWith(github.ref, 'refs/tags/v') || github.event_name == 'workflow_dispatch' }}
```
codebien, line 260: Instead of checking it on the single step, we can probably do it on the entire job and skip it fully.

codebien (approved):
Hey @McMastS, thanks for your contribution :clap: :tada:

codebien (changes requested):
@mstoykov Swapping @oleiade with you as he is on PTO. You worked recently on it, so it makes sense to me if you review it.

.github/workflows/build.yml
McMastS: I attempted moving the PATH addition into the same step [here](https://github.com/McMastS/k6/actions/runs/13255839293/job/37002511783), but `signtool` wasn't added to the PATH, I suppose the shell environment needs to be reloaded for the change to take place. I'll look into it some more.

What do you think of something like:
```
$SignToolPath = Get-ChildItem -Path "${env:ProgramFiles(x86)}\Windows Kits\10\bin" -Recurse -Filter signtool.exe | Where-Object { $_.DirectoryName -like "*\x64" } | Sort-Object -Descending | Select-Object -First 1

# Sign the Windows binary
& $SignToolPath\signtool.exe sign /f k6.pfx /p "${{ secrets.WIN_SIGN_PASS }}" /tr "http://timestamp.digicert.com" /td sha256 /fd sha256 "packaging\k6.exe"
```
I'm forgetting the exact syntax to call the exe from the item returned from `Get-ChildItem` but if that looks good to your team I'll figure it out when I'm at my Windows machine. It's a little less readable (to me) but avoids the multiple steps.

McMastS: I ended up moving the PATH step into the `package-windows` job, and just calling the executable directly as I mentioned above.

Here's the "successful" build where I kicked it off:
https://github.com/McMastS/k6/actions/runs/13275739721/job/37064965620
image

.github/workflows/build.yml
```diff
@@ -266,11 +264,14 @@ jobs:
           $bytes = [Convert]::FromBase64String("${{ secrets.WIN_SIGN_CERT }}")
           [IO.File]::WriteAllBytes("k6.pfx", $bytes)
 
+          # Get the latest signtool executable
+          $SignTool = Get-ChildItem -Path "${env:ProgramFiles(x86)}\Windows Kits\10\bin" -Recurse -Filter signtool.exe | Where-Object { $_.DirectoryName -like "*\x64" } | Sort-Object -Descending | Select-Object -First 1
+
```
mstoykov, line 269: I am not familiar with this windows syntax so no idea how well that is working.

cc @oleiade as our windows person on that.

On the remaining parts - I am also not certain what the directory structure of this is - I rememer trying to figure it out when I last updated it and .... the docs weren't useful.

is the `10` here the windows version? And we need to change it to 11 when we go to windows 11? 

Is it the version of the Windows kits?!?!

I guess this works somewhat better as at least we do not need the `10.0.17763.0` part :shrug: 

But IMO M$ should just have the thing in the PATH ... or have a way to set it without having to stuff like this.

mstoykov (approved):
LGTM in general, thanks for taking on this @McMastS :bow: 

Again I am not certain this improves stuff enough, I was really hoping I just have skipped some standard M$ blessed way of getting/calling singtool.

I have not tested this and do not understand enough of windows/pwsh syntax to know if this is how you are suppsoed to do it. 

Maybe @oleiade knows more about this.

.github/workflows/build.yml
McMastS: Yea, I 100% agree this should just be on the PATH. Everything I found in my search was people hardcoding the path or doing a search like this. As you said the official docs were very unhelpful.

I'm not sure what the `10` is in the signtool path. My Windows 11 machine also has the latest `signtool` in the same `Windows Kits\10\bin` directory, so it's not tied to the OS version as far as I can tell.

McMastS:
Regarding the failed `test-xk6` checks here, is that because this PR is coming from a fork or should I look into that failure a little deeper? The build step that I changed isn't actually run in this case

codebien (approved):

codebien:
Hey @McMastS thanks for the contribution! :tada: Much appreciated :bow:

McMastS:
> > should I look into that failure a little deeper
> 
> @McMastS No, you shouldn't. This is our fault, we are going to look into it, and we'll let you know as soon as it is fixed. Sorry for the inconvenience.

Oh, no problem! I wanted to make sure I hadn't broken anything. Happy to help!

PR #4537, opened 2025-02-11
Refactor lib/consts/consts.go

codebien:
## What?

It creates a dedicated package for version management. It moves the Semantic Versioning constant string to the new internal pkg.

## Why?

It reduces the public surface of k6 API.


inancgumus (dismissed):
👍

version/version.go
```diff
@@ -0,0 +1,6 @@
+// Package version handles k6 versioning
+package version
```
olegbespalov, line 2: I'm curious if we really need the package for that? :thinking: 

How about moving this to the root (main package) and just access it as `go.k6.io/k6.Version` ?

codebien: Unfortunately, If I understand right your question, AFAIK, it is not possible to import a `main` package from another library.

lib/consts/consts.go
```diff
@@ -1,22 +1,10 @@
...
-		`  /          \   |   (  |  (‾)  |`,
-		` / __________ \  |_|\_\  \_____/ `,
-	}, "\n")
-
-	return banner
-}
+//
+// Deprecated: alias to support the legacy versioning API. Use the new version package,
+// it will be removed as soon as the external services stop to depend on it.
+const Version = version.SemVer
```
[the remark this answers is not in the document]
codebien: The problem is the OTel output, which is part of the vendor folder. If I drop the alias, then k6 is not buildable.

version/version.go
olegbespalov: I kind of think that there should be a way without referring to the `main` package, but using `go.k6.io/k6` on the other side quickly checking there immediately cycle imports appears so maybe not this time

lib/consts/consts.go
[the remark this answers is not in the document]
olegbespalov: I see, but if the OTEL is the only issue, it's not so challenging to iterate on that either do to step migration of the OTEL or relaxing it from usage of the package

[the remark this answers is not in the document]
codebien: The migration is the plan, but for doing it, we need the new value available, this is the reason for the alias.

We merge this pull request then:
1. We can open the pull request on OTel output to fix to migrate to the new version
2. We release a new version
3. We open a new pull request on k6 for the new OTel version
4. We open a new pull request on k6 for dropping the alias

>or relaxing it from usage of the package

I didn't check into it, I assumed the current form is the best because it fetches automatically from k6 without us updating the value. But maybe there is a better way.

What is your preference?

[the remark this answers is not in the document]
olegbespalov: I'm fine with both strategies :+1: So we can keep the PR as it's and use as an intermediate step for switching usage in OTEL output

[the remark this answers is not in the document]
olegbespalov: Another maybe alternative could be backmerging the OTEL to the k6, which also should solve OTEL dependency issues that happening from time to time. The output remains experimental, but its management and maintenance will be simpler and straightforward.

I'm also curious to hear @mstoykov opinion, since I feel he has one in that topic in his mind :+1:

version/version.go
```diff
@@ -0,0 +1,6 @@
...
+
+// SemVer contains the current version of k6
+// represented using Semantic Versioning expression.
+const SemVer = "0.56.0"
```
joanlopez, line 6: I'm not super convinced `SemVer` is really the best choice here, is there any reason why not to keep `Version`?
Do we really expect this package to contain multiple "formats" of the k6 version?

I'd like to see what other colleagues think.

codebien: >Do we really expect this package to contain multiple "formats" of the k6 version?

Not really, but it still makes sense for me because it's what it contains. A version expressed in semantic versioning format.

joanlopez: Well, I guess one could argue that `version.SemVer` may also express that there are other representations of the k6 version, like `version.Full` (referred to as `vX.Y.Z` for instance), which isn't the case. 

Anyway, I don't see major problems going with this one, and we can always add a `Version` if needed, preferred. So, I'm going to give it a 👍🏻

lib/consts/consts.go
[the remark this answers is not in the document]
mstoykov: tl;dr :+1: on backmerging [^1], not certain if the version we send should be

Part of the idea of the code changes was to move experimental extensions in k6. 

As I've said before experimental modules and outputs are in separate repos because we started that way. Therewere extensions that we decided to add as experimental. They were already in separate repos and we just kept them there.

Arguably in some cases this lets us have faster development, but to be honest this can also be done by not exposing the not extension by default.

By diagrace - in all cases most of the code is already in the repo in the vendor folder.

Looking at the *usage* of this, I have two thoughts:
1. probably there should be a way to get k6 version, both in js and go
2. not certain opentelemtry should be sending k6 version. I do not know what this is used for, but 
2.1. k6 version on an extension doesn't mean much abotu the extension version, so if it something about finding that you ran otel output with X version that did that, k6's version, isn't waht you want - you want the extension one
2.2. if it is is just a formality of otel that we need to have a version, I guess in that case it is okay



[^1]: not certain on the term, givne that it was never in k6 and it isn't an upstream, but :shrug:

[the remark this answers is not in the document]
olegbespalov: >not certain on the term, givne that it was never in k6 and it isn't an upstream, but 🤷 [↩](https://github.com/grafana/k6/pull/4537#user-content-fnref-1-a63f1a20d7fd68cbe1b9a7c81271115d)

:facepalm: Right, it's entirely wrong term here, but I meant merging code of the OTEL output to the k6, same way how we have done with browser, xk6-webcrypto and recently Prometheus

version/version.go
olegbespalov: I share the same feeling as @joanlopez, it was also one of the reasons why I raised the thread https://github.com/grafana/k6/pull/4537#discussion_r1954796281

version/version.go
codebien: [Packages and commands in the same repository](https://go.dev/doc/modules/layout#packages-and-commands-in-the-same-repository) section of this `godoc` page explains what we should do for supporting a root package as `k6`.

TL;DR we should move the `main.go` file under some other package. Here a draft pull request experimenting with the idea. https://github.com/grafana/k6/pull/4560.

The issue is that it would break anyone building from master. Our CI included.

joanlopez: Passing through, just to say that whatever decision we make in regards to this, it should be now or "never" (till v2).
Perhaps it worth a larger discussion with the entire group.

codebien: > Perhaps it worth a larger discussion with the entire group.

@joanlopez do we really have an option here? I think the cost-benefit here is really unbalanced on the costs.

Because it would introduce a lot of breaking changes on critical commands like `go build` / `go install`, but also the UX introduced for the `go install` command is going to be less effective. `go install go.k6.io/k6@latest` is easier than `go install go.k6.io/k6/cmd/k6@latest`.

In terms of benefit, we don't have a real case except the one we are discussing, which has already a fine work-around with the `version` package.

version/version.go
codebien: @joanlopez @olegbespalov Just to clarify, now that we clarified the inconvenience of `k6.Version` what is the request here? To have `version.Version`?

version/version.go
joanlopez: Ah, no, no. I just commented out what I think should be next steps in case we want to explore that path, which is what I got from your draft PR.

But not something I really want to push for.


Indeed, my first reaction with the `version` package was similar to @olegbespalov's one, but I didn't ever commented it because I suspected a bunch of difficulties.

version/version.go
joanlopez: Yeah, that would have probably been my first choice if anybody asked me.

olegbespalov: `Version` also works for me :+1:

inancgumus: Does `k6` have to be on the root directory? Although not ideal, for the complications we confronted, why not create another `k6` package and put `Version` in it? `version.Version` stutters. I believe `version.SemVer` is more descriptive than that. I don't have a strong opinion, though. Just my two cents.

codebien: >Does a k6 package have to be on the root directory?

@inancgumus Not, but what's a good package to not overcomplicate it too much? `go.k6.io/k6/k6.Version`? I don't have any rationale to be against it but I'm definitely not excited about it?
What's your proposal?

olegbespalov: _Technically_ it could be `go.k6.io/k6/cmd.Version`, not sure thought about imports issues :shrug:

codebien: >Technically it could be go.k6.io/k6/cmd.Version,

I like the idea! @olegbespalov :tada: It works for me, and it makes sense about the current program structure. @inancgumus @joanlopez How does it sound for you?

inancgumus: That sounds like a good idea.

mstoykov: if this will not be needed to be public one as we move otel output inside k6 , can we just ove it to internal/lib/consts ?

codebien: >what about meta.Version (go.k6.io/k6/meta.Version)?

I'm fine with that, but I would find difficult to explain what it should contain. What is the difference with the current `lib` package? There is the risk to add another junk drawer similar to `lib` or `common`, where I would find it difficult to understand if the code I might want to add in the future fits it or not.

For example, to stay concrete, can you provide a description that you would put as documentation for the `meta` package?

By the way, I have three additional suggestions:

1. `semver.Version`
2. `buildinfo.Version`
3. `appversion.Version`

Do you prefer one of them?

codebien: @mstoykov Yes, we can do it, and it's a good idea if we merge OTel. But one of my original goals were also to use this pull request to find a better place for those constants instead to stay in a generic `lib` package.

So, I think it still makes sense to have `go.k6.io./k6/internal/buildinfo.Version`.

inancgumus: I believe `meta` is more descriptive than a utility-like package `lib`. Another one could be `metadata`. But, still, they can turn out to be grab-all. So, I also suggested using a `build` package, which "provides" k6-build-related information, like version.

```go
// Package build provides information about the k6 binary at build time.
package build

// Version is the current semantic version of k6.
const Version = "0.57.0"
```

`semver` and `appversion` are too narrow. I'm also fine with `buildinfo`, but I'm a fan of more concise names.

codebien: Ok, I've incorporated all the suggestions and I think we've reached a good balance.

internal/build/version.go
```diff
@@ -0,0 +1,6 @@
+// Package build handles information and processes about the k6 binary's build.
+package build
+
+// Version contains the current version of k6
+// represented using Semantic Versioning expression.
+const Version = "0.56.0"
```
olegbespalov, line 6: I believe it's not up-to-date
```suggestion
const Version = "0.57.0"
```

olegbespalov (changes requested):

olegbespalov (approved):

joanlopez (approved):

inancgumus (approved):
🚀

PR #4538, opened 2025-02-11
Add TODO to Frame.SetContent

inancgumus:
As per https://github.com/grafana/k6/pull/4533#discussion_r1951060681.

### DIFF
--- a/internal/js/modules/k6/browser/common/frame.go
+++ b/internal/js/modules/k6/browser/common/frame.go
@@ -1474,6 +1474,7 @@ func (f *Frame) SetContent(html string, _ *FrameSetContentOptions) error {
 	f.log.Debugf("Frame:SetContent", "fid:%s furl:%q", f.ID(), f.URL())
 
 	// TODO(@inancgumus): Respect the FrameSetContentOptions before executing the action.
+	// A solution is similar to the logic in `WaitForLoadState`.
 
 	js := `(html) => {
 		window.stop();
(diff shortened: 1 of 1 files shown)


internal/js/modules/k6/browser/common/frame.go
```diff
@@ -1473,6 +1473,8 @@ func (f *Frame) selectOption(selector string, values []any, opts *FrameSelectOpt
 func (f *Frame) SetContent(html string, _ *FrameSetContentOptions) error {
 	f.log.Debugf("Frame:SetContent", "fid:%s furl:%q", f.ID(), f.URL())
 
+	// TODO(@inancgumus): Respect the FrameSetContentOptions before executing the action.
```
codebien, line 1476: Out of curiosity, would it be enough by wrapping the context with the timeout or do we need something else? It would be good to have a suggestion regarding how to fix it so whoever will look into it then they might have a better a starting point.

codebien (approved):

internal/js/modules/k6/browser/common/frame.go
inancgumus: Yes, a suggestion would be good. But it's complicated to explain since the option not only involves a timeout, it also involves waiting for a "lifecycle" event. I've added a small suggestion: [41af72a](https://github.com/grafana/k6/pull/4538/commits/41af72a196c7141b4aced1acb7a8f39b0f81ef7a)

codebien (approved):
@inancgumus LGTM :tada: Remember to squash, please :pray:

olegbespalov (approved):

PR #4541, opened 2025-02-12
Support comma in tag values through cli flag

mstoykov:
## What?

Support setting tags which values contain commas through cli flags

## Why?

It seems very strange that this isn't supported at all on the cli, but you can do it through options. 

This is a breaking change as before you can do `--tag one=value,two=value`  which will produce 2 flags one with key `one`, the other with `two`. 

This will now produce only 1 tag. 

I would argue though that the alternative is to forbid commas in the values. And that seems like a bigger breaking change. 

Given our documentation and help we have never documented that you can put two tags with the same flag, so :shrug: 

## Checklist


inancgumus (approved):
Producing one tag sounds more intuitive.

codebien (approved):

PR #4560, opened 2025-02-19
Move `main` package under `cmd`

codebien:
## What?

k6 is used both as library and as command. 

## Why?

This change aligns better with the Go's recommendation https://go.dev/doc/modules/layout#packages-and-commands-in-the-same-repository.

Doing in this way, the root package is now available and importable and we can use it for some specific APIs.


olegbespalov:
My main concern of moving package that we brake `go install go.k6.io/k6@master` command which is widely presented in many places :(

codebien:
>My main concern of moving package that we brake go install go.k6.io/k6@master command which is widely presented in many places :(

Yes, this is my same fear and this is the main reason why I think that it isn't worth to apply this change.

mstoykov:
I do not think the recommendations is valid for our use case:
1. we do not have multiple commands
2. we do not have top level package that is useful - I also do not know what that will be
3. arguably k6 being used as library is not
3.1. the main use for k6
3.2. likely a thing to change once we decide how to move stuff such as `js/modules` out and actually make it not break everything on each major update

PR #4278, opened 2025-01-24
Clean up after merging webcrypto & making module globally avaliable

olegbespalov:
## What?

This PR does [clean-ups after merge of module](https://github.com/grafana/k6/pull/4188) and also makes crypto (and other methods) globally available

## Why?

This finalizes graduation of the web crypto

## Checklist

- [x] I have performed a self-review of my code.
- [ ] I have added tests for my changes.
- [x] I have run linter locally (`make lint`) and all checks pass.
- [x] I have run tests locally (`make tests`) and all tests pass.
- [ ] I have commented on my code, particularly in hard-to-understand areas.

## Related PR(s)/Issue(s)

https://github.com/grafana/k6/pull/4188

Closes: #3154
Closes: #4031

PR #4563, opened 2025-02-19
timers: refactor how setTimeout and co are made available globally

### LINKED ISSUE #3814: Fix k6/timers always being reported as used
### Brief summary

Due to how #3589 was implemented it is currently reported that *every* script is using `k6/timers` even if they do not.

### k6 version

v0.51.0

### OS

all

_No response_

Run a script without using `k6/timers`

### Expected behaviour

It isn't reported it uses k6/timers

### Actual behaviour

it is reported k6/timers are used.

mstoykov:
## What?

This makes the setTimeout and co. timers package into a not js module package. The current one is tc55, the newly made common API to make technical committee alongside tc39. This might be changed later.

This also allows us to make them globally available without importing js module "k6/timers" allowing us to fix #3814.

## Why?

Reporting k6/timers as always used is not good.

Also, setTimeout and co shouldn't; be dependent on being in a module.

Likely more changes will happen in the future and tc55 will either grow or disappear as a package. 

breaking change is set only because previously in modulestest `k6/timers` was automatically made available - I doubt anyone used that, but it is a thing.

## Checklist

### DIFF
--- a/internal/js/tc55/timers/timers.go
+++ b/internal/js/tc55/timers/timers.go
@@ -3,6 +3,7 @@ package timers
 
 import (
 	"fmt"
+	"slices"
 	"time"
 
 	"github.com/mstoykov/k6-taskqueue-lib/taskqueue"
@@ -247,8 +248,7 @@ func (e *timers) setupTaskQueueCloserOnIterationEnd() {
 						timer.name, timer.id)
 				}
 
-				// TODO: use `clear` when we only support go 1.21 and above
-				e.timers = make(map[uint64]time.Time)
+				clear(e.timers)
 				e.queue.stopTimer()
 				e.queue = new(timerQueue)
 				e.taskQueue = nil
@@ -282,17 +282,13 @@ type timerQueue struct {
 }
 
 func (tq *timerQueue) add(t *timer) int {
-	var i int
-	// don't use range as we want to index to go over one if it needs to go to the end
-	for ; i < len(tq.queue); i++ {
-		if tq.queue[i].nextTrigger.After(t.nextTrigger) {
-			break
-		}
+	i := slices.IndexFunc(tq.queue, func(tt *timer) bool {
+		return tt.nextTrigger.After(t.nextTrigger)
+	})
+	if i < 0 {
+		i = len(tq.queue)
 	}
-
-	tq.queue = append(tq.queue, nil)
-	copy(tq.queue[i+1:], tq.queue[i:])
-	tq.queue[i] = t
+	tq.queue = slices.Insert(tq.queue, i, t)
 	return i
 }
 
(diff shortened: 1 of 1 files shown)


internal/js/tc55/timers/timers.go
```diff
@@ -0,0 +1,346 @@
...
+		if tq.queue[i].nextTrigger.After(t.nextTrigger) {
+			break
+		}
+	}
+
+	tq.queue = append(tq.queue, nil)
+	copy(tq.queue[i+1:], tq.queue[i:])
+	tq.queue[i] = t
+	return i
+}
```
inancgumus, line 297: I find this easier to grok, but it's up to you 🙇 

```suggestion
func (tq *timerQueue) add(t *timer) int {
	i := slices.IndexFunc(tq.queue, func(tt *timer) bool {
		return tt.nextTrigger.After(t.nextTrigger)
	})
	if i < 0 {
		i = len(tq.queue)
	}
	tq.queue = slices.Insert(tq.queue, i, t)
	return i
}
```

inancgumus (dismissed):
LGTM 👍 With mild suggestions.

olegbespalov (dismissed):
In my opinion, the current timers implementation still hardly coupled to the k6's modules and can't be really named as the tc55 thing, so it's just increasing of the complexity.

But if you had a plan how it should evolve, I'm fine with it

mstoykov:
How is it coupled ? Do you want me to not have `modules.VU` in it at all as well? I can refactor it more to not care about anything from k6 as well, it just doesn't make that much sense until we put more work into moving stuff such as the event loop outside of k6

olegbespalov:
>How is it coupled ? Do you want me to not have modules.VU in it at all as well?

Yes, that's what I meant. Like I said, if you have a plan, it's good, for now. I don't want changes right now. Frankly, I mostly care about wrapping the work on the #4278 which I'm going to align, once this will be merged

olegbespalov (approved):

oleiade (approved):
LGTM as-is 👍🏻 🚀 

Absolutely non-blocking comment from me—just sharing my perspective (feel free to proceed as-is). The only real downside I see here is that the way the timers module integrates feels somewhat implicit. Specifically, it relies on certain functions being injected into the global object as a side effect of calling a function from the tc55/timers module in the bundle code.

It took me about five minutes to connect the dots in the context of this PR, but I’m not confident it will be as easy six months from now. If I were to revisit this code in isolation—outside the context of this PR—I’d likely be a bit confused about the expectation that the timers functions should be set in the global object. Tracing back where and how this setup happens would take some effort.

I don’t have a concrete alternative in mind, just an intuition: since we’ll likely be adding more to the global object over time, maybe it would be worth having a single, explicit place in the code dedicated to managing its state? Something like a centralized abstraction or entry point where all GlobalObject().Set calls related to module function injection live, making it clear where to look.

Once again, I'm happy to merge as-is, this is really me dumping my brain 🧠 🙇🏻

mstoykov:
I wrote the above comment before I saw the approvals or your comment @oleiade .

> Specifically, it relies on certain functions being injected into the global object as a side effect of calling a functio

I did originally set them globally from the bundle and modulestest and had to have `SetTimeout` and co. exported, but :
1. it has code repeated - arguably as I've mentioned in other places, modulestest and bundle should share a lot more code. But this will be even bigger problem IMO 
2. timers don't really make much of a sense if they are not set globally - that is practically the idea, so it being part of the package made sense. I specifically also chose a function name that made it more obvious this is happening. But to be honest bundle in particular does so many things setuping stuff, that I too constantly wonder why is this done here, and hten I remember the interconnections, but a lot of the stuff depend on each other. 

I will try to get through bundle and maybe reorder stuff and see if I can make most of them in `setGlobal` and move `setInitGlobals` there as well.

PR #4604, opened 2025-03-06
Update golangci-lint to 1.64.6 and update code to conform

mstoykov:
## What?

Bump golangci-lint to latest version

## Why?

Mostly so it actually works with a version that works with go1.24, but also as to get some bugfixes ... and new stuff.

- usetesting seems okay and we already had tenv - I have done some t.TempDir() fixes around it as well
- exportloopref - is not relevant after using go 1.22
## Checklist

- [ ] I have updated the [k6-documentation](https://github.com/grafana/k6-docs): grafana/k6-docs#PR-NUMBER
- [ ] I have updated the [TypeScript definitions](https://github.com/grafana/k6-DefinitelyTyped/tree/master/types/k6): grafana/k6-DefinitelyTyped#PR-NUMBER

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,4 +1,4 @@
-# v1.63.4
+# v1.64.6
 # Please don't remove the first line. It uses in CI to determine the golangci version
 run:
   timeout: 5m
@@ -65,6 +65,8 @@ linters-settings:
       # Forbid everything in syscall except the uppercase constants
       - '^syscall\.[^A-Z_]+$(# Using anything except constants from the syscall package is forbidden )?'
       - '^logrus\.Logger$'
+  usetesting:
+    os-setenv: true
 
 linters:
   disable-all: true
@@ -84,7 +86,6 @@ linters:
     - errname
     - errorlint
     - exhaustive
-    - exportloopref
     - fatcontext
     - forbidigo
     - forcetypeassert
@@ -125,13 +126,13 @@ linters:
     - sqlclosecheck
     - staticcheck
     - stylecheck
-    - tenv
     - tparallel
     - typecheck
     - unconvert
     - unparam
     - unused
     - usestdlibvars
+    - usetesting
     - wastedassign
     - whitespace
   fast: false
--- a/internal/js/bundle_test.go
+++ b/internal/js/bundle_test.go
@@ -687,8 +687,7 @@ func TestOpen(t *testing.T) {
 			return fs, "", func() {}
 		},
 		"OsFS": func() (fsext.Fs, string, func()) {
-			prefix, err := os.MkdirTemp("", "k6_open_test") //nolint:forbidigo
-			require.NoError(t, err)
+			prefix := t.TempDir()
 			fs := fsext.NewOsFs()
 			filePath := filepath.Join(prefix, "/path/to/file.txt")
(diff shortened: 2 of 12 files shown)


.golangci.yml
```diff
@@ -65,6 +65,8 @@ linters-settings:
       # Forbid everything in syscall except the uppercase constants
       - '^syscall\.[^A-Z_]+$(# Using anything except constants from the syscall package is forbidden )?'
       - '^logrus\.Logger$'
+  usetesting:
+    os-setenv: true
```
joanlopez, line 69: Is there any reason to enable this explicitly, which is enabled by default, and while we don't explicitly set others. 

I'm even more curious considering that we rarely (never till now?) use `Setenv()`, cause it is incompatible with `t.Parallel`, and we have a decently good environment variables "injection".

joanlopez: Also related, why not enabling for instance:
```
os-temp-dir: true
```
which is disabled by default, and we have one occurrence.

Wdyt? 🤔

joanlopez (commented):
Apart from the small comments I made around the `usetesting` configuration, the rest looks pretty well. Nice improvement!

mstoykov:
Hmmm ... I really just didn't want to remove some linting we are doing. But it seems that this doesn't work great with windows, so I am currently for just dropping usetesting entirely and opening an issue around it?

WDYT @joanlopez

joanlopez:
> Hmmm ... I really just didn't want to remove some linting we are doing. But it seems that this doesn't work great with windows, so I am currently for just dropping usetesting entirely and opening an issue around it?
> 
> WDYT @joanlopez

I think that's because Unix and Windows handle open files differently: Unix permits removing an open file while Windows does not, but overall I think we should try to always close temporary files in tests as good practice.

Indeed, in this case I suspect that the line `defer os.Remove(logFilename)` on the failing test (`TestFileConsole`) is hiding a similar error, and we're likely "leaking" files not actually being removed on Windows when running tests. Not 100% sure, but I'd not be surprised if that's the case.

That said, I think the concrete problem here is that:
- The file is opened by the console logger, and never closed.
- The file is opened at the end of the test, to read its contents, and never closed.

So, I'd suggest doing something like https://github.com/grafana/k6/compare/bumpGolangci...bumpGolangci-2. There's likely a bit cleaner way of doing so, but just to give you an example of the idea. That seems to be working fine.

mstoykov:
OOPS I read that as it is way too hard to make it work and went ahead and reverted the changes. 

But if the needed changes are what you provided in that PR do you want to fix the lint problem there and merge it here ?

.golangci.yml
mstoykov: > which is enabled by default,

I actually read their example in the repo as that it is `false` by default :facepalm: and didn't read the exampel above saying it is actually true by default.

I mostly was trying to keep the same as before just transition to newer linters.

I would prefer to open a separate issue to update this and maybe enable more stuff than something else. After all I do this mostly to be able to update to go 1.24 without golangci-lint OOMing.

Does that sound okay?

joanlopez: Yup! Let's go!

joanlopez (approved):

olegbespalov (approved):

PR #4607, opened 2025-03-06
Chore: Close FDs on TestFileConsole to avoid issues on Win

joanlopez:
## What?

It closes the open file descriptors on the logger's output file on `TestFileConsole`. 

## Why?

Because, as discussed [here](https://github.com/grafana/k6/pull/4604#issuecomment-2703809796), that's what's causing that test to fail on Windows.

### DIFF
--- a/internal/js/console_test.go
+++ b/internal/js/console_test.go
@@ -378,7 +378,6 @@ func TestFileConsole(t *testing.T) {
 							f, err := os.CreateTemp(t.TempDir(), "") //nolint:forbidigo // fix with https://github.com/grafana/k6/issues/2565
 							require.NoError(t, err)
 							logFilename := f.Name()
-							defer os.Remove(logFilename) //nolint:errcheck,forbidigo // fix with https://github.com/grafana/k6/issues/2565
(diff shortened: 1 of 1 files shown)


internal/js/console_test.go
```diff
@@ -378,7 +378,6 @@ func TestFileConsole(t *testing.T) {
 							f, err := os.CreateTemp(t.TempDir(), "") //nolint:forbidigo // fix with https://github.com/grafana/k6/issues/2565
 							require.NoError(t, err)
 							logFilename := f.Name()
-							defer os.Remove(logFilename) //nolint:errcheck,forbidigo // fix with https://github.com/grafana/k6/issues/2565
```
joanlopez, line 381: I think this line is no longer required, as `t.TempDir()` is responsible, thanks to `t.Cleanup()` behavior, to completely remove the temporary directory created for this test, once it finishes. Right?

mstoykov: Seems reasonable - I did remove some lines like this from some places, but likely missed some as well

mstoykov (approved):

Issue #4154, opened 2024-12-27
Improve UX of the `new` command


dgzlopes:
### Feature Description

The `k6 new` command was introduced to:
- Simplify the process of creating new k6 scripts.
- Teach users about k6's capabilities (incl. Cloud integration and browser support).

After seeing it used in many places, I have detected some problems:

- It doesn't work well with the Cloud flows.
   - On our Cloud onboarding, in some cases, we tell the user to generate the script, open it, and change things manually.
      - <img width="1099" alt="Image" src="https://github.com/user-attachments/assets/e3284164-da94-4efc-b847-4a2e7b703b11" />
   - In others, we ask the user to copy-paste a script directly instead of using this command (e.g., new Test Run page).
      - <img width="1099" alt="Image" src="https://github.com/user-attachments/assets/1c49b228-5a0e-45a1-a734-aeb0d808ba1c" />
- The current example script tries to do too much without being able to be minimal enough or complete enough.  
   - It is a protocol test with a massive commented config block that always includes Cloud, Browser, etc.  
   - Even if you have Browser in the options block... there is no browser code in the test logic. 
 

### Suggested Solution (optional)

- Having two 'new' commands
   - `k6 new`
   - `k6 cloud new`
      - The Cloud block will appear in the config block of the generated test when you use this one.
      - Also, optionally, users can pass `--project-id` to override the dummy one. 
         - This would solve the "manual changes" required step of Cloud onboarding. We can tell users to copy-paste:
              - `k6 cloud new --project-id 123145`
- Having templates instead of one massive example (note: I was inspired by how Svelte handles templates).
    - Usage: `k6 cloud new --project-id 123145 --template=browser`  

Minimal (default) - Small and concise script. It is ideal for folks who need a starting point for a new project
```js
import http from 'k6/http';
import { sleep, check } from 'k6';

export const options = {
  vus: 10,
  duration: '30s',
};

export default function() {
  let res = http.get('https://quickpizza.grafana.com');
  check(res, { "status is 200": (res) => res.status === 200 });
  sleep(1);
}
```

Protocol - More real script. It is ideal for folks learning, as a reference, or for demos.
```js
import http from "k6/http";
import { check, sleep } from "k6";

const BASE_URL = __ENV.BASE_URL || 'https://quickpizza.grafana.com';

export const options = {
  vus: 10,
  stages: [
    { duration: "10s", target: 5 },
    { duration: "20s", target: 10 },
    { duration: "1s", target: 0 },
[...]

PR #4618, opened 2025-03-10
Local templates support in `k6 new`

andrewslotin:
## What?

This PR adds support for local templates and addresses a bug when an empty script file is created even if the template name is invalid.

## Why?

The current script templating feature is limited to three pre-defined templates that are bundled with `k6`: minimal, protocol and browser. While this already reduces the amount of boilerplate code to write when starting a new project, being able to use custom user-defined templates would help to simplify the first steps even further, by allowing users to extract common parts, such as authentication, custom sets of options, into a template file that can be used to initialize tests.

## Checklist

- [x] I have updated the [k6-documentation](https://github.com/grafana/k6-docs): https://github.com/grafana/k6-docs/pull/1889
- [x] I have updated the release notes

## Related PR(s)/Issue(s)

Relevant issue https://github.com/grafana/k6/issues/4154

### DIFF
--- a/internal/cmd/new.go
+++ b/internal/cmd/new.go
@@ -25,7 +25,7 @@ func (c *newScriptCmd) flagSet() *pflag.FlagSet {
 	flags := pflag.NewFlagSet("", pflag.ContinueOnError)
 	flags.SortFlags = false
 	flags.BoolVarP(&c.overwriteFiles, "force", "f", false, "overwrite existing files")
-	flags.StringVar(&c.templateType, "template", "minimal", "template type (choices: minimal, protocol, browser)")
+	flags.StringVar(&c.templateType, "template", "minimal", "template type (choices: minimal, protocol, browser) or relative/absolute path to a custom template file") //nolint:lll
 	flags.StringVar(&c.projectID, "project-id", "", "specify the Grafana Cloud project ID for the test")
 	return flags
 }
--- a/internal/cmd/new.go
+++ b/internal/cmd/new.go
@@ -2,6 +2,8 @@ package cmd
 
 import (
 	"fmt"
+	"io"
+	"strings"
 
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
@@ -28,7 +30,7 @@ func (c *newScriptCmd) flagSet() *pflag.FlagSet {
 	return flags
 }
 
-func (c *newScriptCmd) run(_ *cobra.Command, args []string) error {
+func (c *newScriptCmd) run(_ *cobra.Command, args []string) (err error) {
 	target := defaultNewScriptName
 	if len(args) > 0 {
 		target = args[0]
@@ -42,26 +44,7 @@ func (c *newScriptCmd) run(_ *cobra.Command, args []string) error {
 		return fmt.Errorf("%s already exists. Use the `--force` flag to overwrite it", target)
 	}
 
-	fd, err := c.gs.FS.Create(target)
-	if err != nil {
-		return err
-	}
-
-	var closeErr error
-	defer func() {
-		if cerr := fd.Close(); cerr != nil {
-			if _, err := fmt.Fprintf(c.gs.Stderr, "error closing file: %v\n", cerr); err != nil {
-				closeErr = fmt.Errorf("error writing error message to stderr: %w", err)
-			} else {
-				closeErr = cerr
-			}
-		}
-	}()
-
-	if closeErr != nil {
-		return closeErr
-	}
-
+	// Initialize template manager and validate template before creating any files
 	tm, err := templates.NewTemplateManager()
 	if err != nil {
 		return fmt.Errorf("error initializing template manager: %w", err)
@@ -72,12 +55,36 @@ func (c *newScriptCmd) run(_ *cobra.Command, args []string) error {
 		return fmt.Errorf("error retrieving template: %w", err)
 	}
 
+	// Prepare template arguments
 	argsStruct := templates.TemplateArgs{
 		ScriptName: target,
 		ProjectID:  c.projectID,
 	}
 
-	if err := templates.ExecuteTemplate(fd, tmpl, argsStruct); err != nil {
+	// First render the template to a buffer to validate it
+	var buf strings.Builder
+	if err := templates.ExecuteTemplate(&buf, tmpl, argsStruct); err != nil {
+		return fmt.Errorf("failed to execute template %s: %w", c.templateType, err)
+	}
+
+	// Only create the file after template rendering succeeds
+	fd, err := c.gs.FS.Create(target)
+	if err != nil {
+		return err
+	}
+
+	defer func() {
+		if cerr := fd.Close(); cerr != nil {
+			if _, err := fmt.Fprintf(c.gs.Stderr, "error closing file: %v\n", cerr); err != nil {
+				err = fmt.Errorf("error writing error message to stderr: %w", err) //nolint:staticcheck,ineffassign,wastedassign
+			} else {
+				err = cerr //nolint:staticcheck,ineffassign,wastedassign
+			}
+		}
+	}()
+
+	// Write the rendered content to the file
+	if _, err := io.WriteString(fd, buf.String()); err != nil {
 		return err
 	}
 
--- a/internal/cmd/new.go
+++ b/internal/cmd/new.go
@@ -45,7 +45,7 @@ func (c *newScriptCmd) run(_ *cobra.Command, args []string) (err error) {
 	}
 
 	// Initialize template manager and validate template before creating any files
-	tm, err := templates.NewTemplateManager()
+	tm, err := templates.NewTemplateManager(c.gs.FS)
 	if err != nil {
 		return fmt.Errorf("error initializing template manager: %w", err)
 	}
--- a/internal/cmd/new.go
+++ b/internal/cmd/new.go
@@ -75,10 +75,10 @@ func (c *newScriptCmd) run(_ *cobra.Command, args []string) (err error) {
 
 	defer func() {
 		if cerr := fd.Close(); cerr != nil {
-			if _, err := fmt.Fprintf(c.gs.Stderr, "error closing file: %v\n", cerr); err != nil {
-				err = fmt.Errorf("error writing error message to stderr: %w", err) //nolint:staticcheck,ineffassign,wastedassign
+			if _, werr := fmt.Fprintf(c.gs.Stderr, "error closing file: %v\n", cerr); werr != nil {
+				err = fmt.Errorf("error writing error message to stderr: %w", werr)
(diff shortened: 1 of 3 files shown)


internal/cmd/templates/templates.go
```diff
@@ -52,21 +57,59 @@ func NewTemplateManager() (*TemplateManager, error) {
...
+	if strings.HasPrefix(path, ".\\") ||
+		strings.HasPrefix(path, "..\\") ||
+		strings.HasPrefix(path, "\\") ||
+		strings.HasPrefix(path, "\\\\") || // UNC paths
+		(len(path) >= 2 && path[1] == ':') { // Drive letter paths like C:
+		return true
+	}
+
+	return false
 }
```
inancgumus, line 113: This heuristic check can fail when the file's name is the same as a valid template type (e.g., browser). Another case it might fail when the file is provided with a relative path (e.g., --template=myFile). A better way might be to separate the template type and the file name. We can then use `fs.Exists` instead of a heuristic check for reliability.

internal/cmd/new.go
```diff
@@ -28,7 +30,7 @@ func (c *newScriptCmd) flagSet() *pflag.FlagSet {
 	return flags
 }
 
-func (c *newScriptCmd) run(_ *cobra.Command, args []string) error {
+func (c *newScriptCmd) run(_ *cobra.Command, args []string) (err error) {
```
inancgumus, line 33: This can remove `nolint`s (`rerr` (return error) instead of `err`):

```suggestion
func (c *newScriptCmd) run(_ *cobra.Command, args []string) (rerr error) {
```

inancgumus (approved):
Nice feature! LGTM 👍

I have some suggestions.

The `--template` option specifies the template type for the output. The current implementation of the `--template` option also reads the file name. Could this be confusing for users? What do you think about adding a `--template-file` option to explicitly specify the file name? Doing so can also simplify the implementation of [this](https://github.com/grafana/k6/pull/4618#discussion_r1987621331).

```bash
$ ... --template=file --template-file=templateFile
# Or, just:
$ ... --template-file=templateFile
```

internal/cmd/new_test.go
```diff
@@ -119,3 +124,101 @@ func TestNewScriptCmd_ProjectID(t *testing.T) {
...
+	require.NoError(t, fsext.WriteFile(ts.FS, templatePath, []byte(templateContent), 0o600))
+
+	ts.CmdArgs = []string{"k6", "new", "--template", templatePath}
+
+	newRootCommand(ts.GlobalState).execute()
+
+	data, err := fsext.ReadFile(ts.FS, defaultNewScriptName)
+	require.NoError(t, err)
+
+	assert.Equal(t, templateContent, string(data), "generated file matches the template content")
```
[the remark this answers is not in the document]
andrewslotin: Makes sense 👍

internal/cmd/templates/templates.go
andrewslotin: The implementation expects the file path to be either absolute or relative, so it always starts with `/`, `./` or `../`. I believe it's safe to assume that none of the built-in templates will have these prefixes. To use a template stored in `browser` file, user would need to specify it either as `./browser` or `/foo/bar/browser`.

internal/cmd/new.go
andrewslotin: Is this the case? I believe this is rather a linter issue. It claims that inside the `defer` block we assign a value that is never user afterwards, which is not true, because `err` will be returned upon the exit from `run()`. Or am I getting your suggestion wrong?

andrewslotin: My bad, the linter was complaining because `err` is being shadowed inside of the `if` block in the deferred function. I'll push a fix

internal/cmd/templates/templates.go
inancgumus: Yep, I see that users need to provide a file path like `./browser` for files and `browser` for a built-in template. However, distinguishing between a file and a built-in template just by checking a prefix like `./` can be easy to overlook and error-prone for users. It feels like we're mixing two concepts: template type and template source.

internal/cmd/new.go
inancgumus: Shadowing always hunts us 😄 👻

inancgumus:
@oleiade 

> A pattern we currently use in k6 is to parse the value as a = separated key value pair: `k6 new --template file=template.js`.

This makes sense to be consistent 👍

> Otherwise, we could indeed go for --template-file, but the concern I have there is that we might want to support more than files down the road (folders?) in which case the "file" terminology might turn out confusing.

We could call it `--template-source`. However, consistency matters more.

internal/cmd/templates/templates.go
andrewslotin: I've added an explicit error message, that checks if a file with this name exists in current working directory and suggests the correct value.

inancgumus: Thanks! Better 👍 I liked the way it says Google-like _"invalid template type "foo", did you mean ./foo?"_ 😄 But it still feels a little bit odd because the template type is a file path and users might have typed `browse` (without the `r` suffix), which would be confusing to tell them to type `./browser` :(

andrewslotin:
@oleiade, good catch on unclear error and missing changes to the help message. Both should be fixed now.

@inancgumus, introducing a separate command-line flag for template files would allow users to do something like this `--template minimal --template-file ./custom-template.js`, which is an undefined behavior. To generate a script `k6 new` needs exactly one template, and using one flag to handle the value enforces this. I see your point that this may lead to a confusion, if someone tries to store and use template with exactly the same file name, as one of the built-in ones, but this seems an unlikely case and we can address it with user-friendly errors and documentation. For the context, I'm following [this proposal](https://github.com/grafana/k6/issues/4154#issuecomment-2582905637) and plan to add support for remote templates (at least ones that can be downloaded via HTTP/HTTPS). I'd rather avoid introducing yet another flag for URLs.

internal/cmd/templates/templates.go
andrewslotin: I think it's would be too user-friendly to implement fuzzy matching with all files in the current directory 😃

inancgumus:
Thanks for the clarification, @andrewslotin 🙇 However, [this suggestion](https://github.com/grafana/k6/pull/4618#issuecomment-2714900938) avoids introducing a new flag and remains consistent with the rest of the `k6` command: `k6 new --template=file=foo`.

andrewslotin:
@inancgumus, what's the benefit of requiring users to type `file=` instead of providing a path if we can reliably infer what they meant? From UX perspective the latter is more convenient, because it works with <kbd>TAB</kbd> autocompletion and drag-and-drop of files supported by modern terminals. The `key=value` syntax is commonly used to provide key-value maps as arguments, like we do in `--env`. We require a single scalar value, while `key=value` format allows things like `builtin=minimal,file=./custom-template.js`.

andrewslotin:
@inancgumus, that would simplify the implementation for sure. Can you write down the full proposal you have in mind? Let's say we aim to support a small number of built-in templates, local files and remote ones. How would the arguments to `--template` look like in all these cases?

inancgumus:
*Given the constraints we discussed:*
- Avoid confusion for users.
- Avoid introducing another flag since `--template` enforces the one template idea.
- Avoid specifying external templates like files as key=value pairs (e.g., `file=..`).
- Avoid undefined behavior like `--template minimal --template-file=foo`
- Provide reliable functionality and clear error messages.
- Support easy drag and drop and tab completion.
- Support for built-in templates, files, directories, and remote protocols like HTTP.

I believe a better approach might be:

```bash
# built-in templates
k6 new --template builtin:minimal    # default
k6 new --template builtin:protocol
k6 new --template builtin:browser

# http etc.
k6 new --template http://foo.com

# files and directories
k6 new --template ~/foo
```

So:
- We can first check the `"builtin:"` prefix.
- If not, we can check if it's a valid remote path (`http://` prefix), which is reliably easy.
- Finally, check if the file exists instead of a heuristic string match.

Downsides:
- Users might type foo.com/foo as a remote URL.
- However, due to the constraints we have (avoiding separate flags), this is what we have.

The `builtin:` prefix removes any ambiguity with filenames. If we just type `minimal`, the tool has to figure out if it's a built‑in or a file with that name. By requiring `builtin:minimal`, we make it explicit. It's highly unlikely to be interpreted as a valid file path on most operating systems, so it cleanly distinguishes built‑in templates from real files and other resources.

-----

If we didn't have the "multiple flags" constraint, we could have:

```bash
# built-in templates
k6 new --builtin-template minimal    # default
k6 new --builtin-template protocol
k6 new --builtin-template browser

# http etc.
k6 new --remote-template http://foo.com

# files and directories
k6 new --local-template ~/foo
```

andrewslotin:
@inancgumus, I believe we're reconsidering the previously discussed and agreed design now. Instead, I'd suggest that we focus on the implementation.

I see your point that requiring users to prepend the file name with `./` if a template is stored in the same folder as a new project might cause confusion; however, I believe this scenario is rather unlikely to happen. If someone has already copied the template file, there is not much benefit in running the `k6 new` command. However, to address this use case, a better error message would make sense – thanks for finding this.

The point about edge cases like using `~/`, etc. is a good one. Although I've tried this locally and the `~` seems to be expanded to my home directory before passing this string to `k6`, I'm not 100% confident that it works the same way in other terminals/shells, so I've simplified the `isFilePath()` check to only look for a path separator in the string. This puts limitations on the built-in template names, as we won't be able to add one with a `/` or `\\`, but I believe it's a fair tradeoff. Besides, if we ever need to add such a built-in template, we can iterate on the file path check and make it more sophisticated, but, again, we control built-ins, so I don't think this will cause us trouble anytime soon.

internal/cmd/templates/templates.go
```diff
@@ -52,21 +58,49 @@ func NewTemplateManager() (*TemplateManager, error) {
...
 		return tm.minimalTemplate, nil
 	case ProtocolTemplate:
 		return tm.protocolTemplate, nil
 	case BrowserTemplate:
 		return tm.browserTemplate, nil
-	default:
-		return nil, fmt.Errorf("invalid template type: %s", templateType)
 	}
+
+	// Then check if it's a file path
```
oleiade, line 77: Non-blocking, feel free to ignore.

I don't think it's absolutely worth it for this specific feature, but more as a best practice, it could be worth cleaning up and transforming the path to an absolute path here, at least as we work with it internally: `tplPath, err := filepath.Abs(filepath.Clean(tpl))`.

oleiade (approved):
🚀 👏🏻 🍾 Looks good to me!

inancgumus (approved):
🤝 🚀

inancgumus (approved):

oleiade (approved):

PR #4629, opened 2025-03-13
Fix options parsing error printing the most of the options

mstoykov:
## What?

Make the error message not print practically the whole options on error.

## Why?

The original idea of #4602 was to report at least the key with potentially the option in question. Due to the lack of support from the error - it can't parse it. The commit was trying to get everything between the previous and next new line.

But in practice was getting anything until the end, both because of error in the coed to get the index, but also because there were no newlines due to how we marshal the data before that.

This should fix both and make it less likely we just get everything until the end.

## Checklist

- [ ] I have updated the [k6-documentation](https://github.com/grafana/k6-docs): grafana/k6-docs#PR-NUMBER
- [ ] I have updated the [TypeScript definitions](https://github.com/grafana/k6-DefinitelyTyped/tree/master/types/k6): grafana/k6-DefinitelyTyped#PR-NUMBER

### DIFF
--- a/internal/js/bundle.go
+++ b/internal/js/bundle.go
@@ -213,7 +213,6 @@ func (b *Bundle) populateExports(updateOptions bool, bi *BundleInstance) error {
 					err = fmt.Errorf("error parsing script options: %w", err)
 					return
 				}
-				fmt.Println(string(data))
 				dec := json.NewDecoder(bytes.NewReader(data))
 				dec.DisallowUnknownFields()
 				if err = dec.Decode(&b.Options); err != nil {
--- a/internal/js/bundle.go
+++ b/internal/js/bundle.go
@@ -217,8 +217,7 @@ func (b *Bundle) populateExports(updateOptions bool, bi *BundleInstance) error {
 				dec.DisallowUnknownFields()
 				if err = dec.Decode(&b.Options); err != nil {
 					if uerr := json.Unmarshal(data, &b.Options); uerr != nil {
-// comment about the intentions here like
-// Beautify the error because we want to return only the failing key and not the entire options' body as it would be without it. 
+						// Beautify the error so we can try to show the user the key and potential value that is failing to parse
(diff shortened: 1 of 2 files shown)


internal/js/bundle.go
```diff
@@ -208,16 +208,17 @@ func (b *Bundle) populateExports(updateOptions bool, bi *BundleInstance) error {
 					continue
 				}
 				var data []byte
-				data, err = json.Marshal(v.Export())
+				data, err = json.MarshalIndent(v.Export(), "", "  ")
 				if err != nil {
 					err = fmt.Errorf("error parsing script options: %w", err)
 					return
 				}
+				fmt.Println(string(data))
```
olegbespalov, line 216:
```suggestion
```

olegbespalov (approved):

internal/js/bundle.go
```diff
@@ -217,7 +217,7 @@ func (b *Bundle) populateExports(updateOptions bool, bi *BundleInstance) error {
 				dec.DisallowUnknownFields()
 				if err = dec.Decode(&b.Options); err != nil {
 					if uerr := json.Unmarshal(data, &b.Options); uerr != nil {
-						uerr = beautifyJSONUnmarshalError(data, uerr)
+						uerr = beautifyOptionsJSONUnmarshalError(data, uerr)
```
codebien, line 220: >The original idea of https://github.com/grafana/k6/pull/4602 was to report at least the key with potentially the option in question.

I think this expectation is not really visible from the code, nor testing. I would add a comment explaining it, so I don't have to recall the pull requests, if I hit it for some reason the next time.  

```suggestion
// comment about the intentions here
						uerr = beautifyOptionsJSONUnmarshalError(data, uerr)
```

codebien (changes requested):

mstoykov:
I will squash this at the end

olegbespalov (approved):

codebien (approved):

PR #4630, opened 2025-03-13
browser: fix web vitals test flakiness on Windows

andrewslotin:
## What?

Increase timeout and add a 1s delay after page load to ensure metrics are collected.

## Why?

The `TestWebVitalMetricNoInteraction` test is [failing on Windows](https://github.com/grafana/k6/actions/runs/13816953723/job/38652540550#step:4:69) due to timing issues when collecting web vital metrics.

## Checklist

### DIFF
--- a/internal/js/modules/k6/browser/tests/webvital_test.go
+++ b/internal/js/modules/k6/browser/tests/webvital_test.go
@@ -94,6 +94,12 @@ func TestWebVitalMetric(t *testing.T) {
 func TestWebVitalMetricNoInteraction(t *testing.T) {
 	t.Parallel()
 
+	// Increase timeout for Windows
+	timeout := 5 * time.Second
+	if runtime.GOOS == "windows" {
+		timeout = 10 * time.Second
+	}
+
 	var (
 		samples  = make(chan k6metrics.SampleContainer)
 		browser  = newTestBrowser(t, withFileServer(), withSamples(samples))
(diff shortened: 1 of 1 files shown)


andrewslotin:
**_Narrator:_** now _another_ test is failing, but we will come back to in the next episode.

internal/js/modules/k6/browser/tests/webvital_test.go
```diff
@@ -94,6 +94,12 @@ func TestWebVitalMetric(t *testing.T) {
 func TestWebVitalMetricNoInteraction(t *testing.T) {
 	t.Parallel()
 
+	// Increase timeout for Windows
+	timeout := 5 * time.Second
+	if runtime.GOOS == "windows" {
+		timeout = 10 * time.Second
+	}
```
codebien, line 101: Do we really need to differentiate? In the end, it is a timeout, so it doesn't count too much on the real execution for the average run.

It counts only on failures but they shouldn't be the average daily run.

internal/js/modules/k6/browser/tests/webvital_test.go
```diff
@@ -140,6 +146,9 @@ func TestWebVitalMetricNoInteraction(t *testing.T) {
 	require.NoError(t, err)
 	require.NotNil(t, resp)
 
+	// Add a small delay to ensure all metrics are collected
+	time.Sleep(1 * time.Second)
```
codebien, line 150: Do you mean samples collection here? If so, can we do it without the use of time so we can have a more deterministic test, please? We can close a `samplesDone` chan when the dedicated goroutine exits.

codebien (changes requested):
@andrewslotin thanks for taking care of it :heart:

internal/js/modules/k6/browser/tests/webvital_test.go
codebien: >I don't see a good reason not to use common.DefaultTimeout for example, which is 30s.

This is another good observation :+1: I'm for it.

codebien (approved):

oleiade (approved):

PR #4631, opened 2025-03-15
fix: handle NaN and Infinity float values in gRPC

ariasmn:
TBD

### DIFF
--- a/internal/js/modules/k6/grpc/client.go
+++ b/internal/js/modules/k6/grpc/client.go
@@ -4,10 +4,12 @@ import (
 	"context"
 	"crypto/tls"
 	"crypto/x509"
+	"encoding/json"
 	"encoding/pem"
 	"errors"
 	"fmt"
 	"io"
+	"math"
 	"strings"
 	"time"
 
@@ -360,7 +362,8 @@ func (c *Client) buildInvokeRequest(
 	if req == nil {
 		return grpcReq, errors.New("request cannot be nil")
 	}
-	b, err := req.ToObject(c.vu.Runtime()).MarshalJSON()
+
+	b, err := json.Marshal(encodeFloatsForJSON(req.ToObject(c.vu.Runtime()).Export()))
 	if err != nil {
 		return grpcReq, fmt.Errorf("unable to serialise request object: %w", err)
 	}
@@ -376,6 +379,35 @@ func (c *Client) buildInvokeRequest(
 		TagsAndMeta:            &p.TagsAndMeta,
 		Metadata:               p.Metadata,
 	}, nil
+}
+
+// encodeFloatsForJSON recursively processes the object and replaces NaN and Infinity with their string equivalents.
+// - math.NaN() is converted to the string "NaN"
+// - math.Inf(1) is converted to the string "Infinity"
+// - math.Inf(-1) is converted to the string "-Infinity"
+// This is necessary because NaN and Infinity are not valid in ProtoJSON, unless represented as strings.
+func encodeFloatsForJSON(data interface{}) interface{} {
+	switch v := data.(type) {
+	case map[string]interface{}:
+		for key, val := range v {
+			v[key] = encodeFloatsForJSON(val)
+		}
+		return v
+	case []interface{}:
+		for i, val := range v {
+			v[i] = encodeFloatsForJSON(val)
+		}
+		return v
+	case float64:
+		if math.IsNaN(v) {
+			return "NaN"
+		} else if math.IsInf(v, 1) {
+			return "Infinity"
+		} else if math.IsInf(v, -1) {
+			return "-Infinity"
+		}
+	}
+	return data
 }
 
 // Close will close the client gRPC connection
--- a/internal/js/modules/k6/grpc/client.go
+++ b/internal/js/modules/k6/grpc/client.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"crypto/tls"
 	"crypto/x509"
-	"encoding/json"
 	"encoding/pem"
 	"errors"
 	"fmt"
@@ -363,7 +362,10 @@ func (c *Client) buildInvokeRequest(
 		return grpcReq, errors.New("request cannot be nil")
 	}
 
-	b, err := json.Marshal(encodeFloatsForJSON(req.ToObject(c.vu.Runtime()).Export()))
+	object := req.ToObject(c.vu.Runtime())
+	normalizeNumberStrings(object, c.vu.Runtime())
+
+	b, err := object.MarshalJSON()
 	if err != nil {
 		return grpcReq, fmt.Errorf("unable to serialise request object: %w", err)
 	}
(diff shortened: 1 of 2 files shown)


ariasmn:
Couple of things:

- Having a hard time trying to understand how to build the tests. I just checked that current tests pass, but I'd need a bit of help to build a test case for this, sorry about that.
- I still don't think that this approach is good enough... I feel like maybe the best bet would be to create the proto message using `protoreflect` and then marshaling normally, but I haven't tried if that would break wrappers and reflection implementations.

mimani68 (approved):

mstoykov (commented):
Thank you for the PR and sorry for the slow reply :bow: 

I have left some comments around the code. 

On the whole idea I am still not particularly certain this is ... what should be done, if anything. 

Trying to figure out how this works in other cases I found https://github.com/protocolbuffers/protobuf-javascript/issues/49 which seems to suggest that this isn't really supported. But maybe I am getting something wrong.

From the test it seems that this just works on the way back which seems awfully convient.

internal/js/modules/k6/grpc/client.go
```diff
@@ -360,7 +361,14 @@ func (c *Client) buildInvokeRequest(
 	if req == nil {
 		return grpcReq, errors.New("request cannot be nil")
 	}
-	b, err := req.ToObject(c.vu.Runtime()).MarshalJSON()
+
+	object := req.ToObject(c.vu.Runtime())
+	normalized, err := normalizeNumberStrings(object, c.vu.Runtime())
```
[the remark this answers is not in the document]
ariasmn: But the object is built based on the `req`, which is a `sobek.Value` passed by value, so we shouldn't worry about it, no? Or am I misunderstanding you? (Probably this, my brain is fried today, sorry :disappointed: )

ariasmn:
Hey @mstoykov, thanks a lot for the review!

I’ve made some changes based on your comments. There’s one part I didn't change yet because I’m not fully sure I understand it correctly, but we can keep discussing it in the open conversation.

Regarding the issue you mentioned: it looks like an older open issue. From what I saw, the code has since changed and now properly asserts both `Infinity` and `NaN`:

https://github.com/protocolbuffers/protobuf-javascript/blob/0768cc9f18bec192c393d7c2962ee8ab5ad1931b/binary/encoder.js#L411-L418

I also tested it locally using the dummy.proto from the issue, for both client and server (wire format). Additionally, the [ProtoJSON documentation](https://protobuf.dev/programming-guides/json/) confirms that both values are officially supported.

Just a heads-up: I might be a bit slower next week if any further changes are needed — apologies in advance.

Thanks again for the great feedback. 🙏

internal/js/modules/k6/grpc/client.go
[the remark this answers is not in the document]
ariasmn: Forget about my last comment, I took some time today (it's been a while, sorry) and completely understood what you meant.

Don't know if the approach I did is good enough or if there's a better way... We can discuss it in another conversation.

ariasmn:
Hi there,

sorry for the delay on this! It's been a while since I could get back to this PR (I haven't had much time recently).

I've made the requested corrections. Please take a look and let me know if everything looks good (I feel like there should be a better way of handling the copy for sure) or if there's anything else that needs to be addressed.

Thank you! 🙇

joanlopez:
Hi @ariasmn,

Sorry for the late review, but finally I've been able to properly test it.

I think it's close to be ready to be merged, but there's one extra thing that we should fix, which is that if `normalizeNumberStrings` receives a cyclic reference to an object (_see below_), it will end up in an infinite loop of recursive calls that eventually will lead to a panic.

For instance:

```
client.connect("GRPCBIN_ADDR");

let obj1 = {};
let obj2 = {};
obj1.a = obj2;
obj2.b = obj1;  // Now both objects reference each other

let respNaN = client.invoke("grpc.wrappers.testing.Service/TestDouble", obj1); // <-- this causes the panic
```

Can you take a look at this and try to fix it, please? Thanks!

ariasmn:
Hi @joanlopez,

Thank you for catching that one! I implemented a fix by following what's done in the `stringify` method in Sobek, and added a test case as well:

https://github.com/grafana/k6/blob/602467dee735ba74fc9664b23ca53220f1212f0c/vendor/github.com/grafana/sobek/builtin_json.go#L366-L372

Let me know if my choice of wording for variables/comments is odd, I can fix them no problem.

Thanks!

internal/js/modules/k6/grpc/client.go
```diff
@@ -387,6 +397,71 @@ func (c *Client) buildInvokeRequest(
...
+// objects or arrays, and converts special float values (NaN, Infinity) to their
+// string representations for proper JSON serialization.
+func normalizeNumberStrings(v sobek.Value, runtime *sobek.Runtime, stack []*sobek.Object) (sobek.Value, error) {
+	// If v is a primitive (not an object or array), handle it directly.
+	if v == nil || sobek.IsNull(v) || sobek.IsUndefined(v) {
+		return v, nil
+	}
+
+	// Handle special float values.
+	if f, ok := v.Export().(float64); ok {
```
joanlopez: As @mstoykov suggested before, could we use `ExportType()` here instead of `Export()`, as the latter is a much more expensive operation.

In case it's a `float64`, I guess you'll need to call `v.Export()` anyway, but in all the other cases, it won't be called that early.

joanlopez: Or maybe even something like:

```suggestion
	if v.ExportType().Kind() == reflect.Float64 {
		f := v.ToFloat()
		switch {
		case math.IsNaN(f):
			return runtime.ToValue("NaN"), nil
		case math.IsInf(f, 1):
			return runtime.ToValue("Infinity"), nil
		case math.IsInf(f, -1):
			return runtime.ToValue("-Infinity"), nil
		}

		return v, nil
	}
```

So, you don't need `Export()` at all.
Wdyt? @mstoykov

mstoykov: This seems better.

For the record - the idea is that if you are using `Export` it will do a lot of work in some of the cases. If the end type is flaot64 - that probably is fine. But it also might be a big object that you just fully transformed from JS to go type. And then you just throw it away because it isn't a float64, just to repeat the operation each time.

ariasmn: Sorry about that, I resolved the previous conversation thinking I had already implemented the change, totally my bad.
Thanks for the explanation and sorry for the inconvenience!

Let me know if there's anything else I can do on my end.

joanlopez: Sure, no problem @ariasmn, that's why we have a review process, we're all humans and sometimes it happens!

Looks good now! 👌🏻 Thanks! 🙌🏻

mstoykov (approved):
LGTM! Thanks for the changes and going through figuring it out @ariasmn :bow:

PR #4641, opened 2025-03-20
Fix: tagset json buffer content

dvordrova:
## What?

Fix easyjson behaviour when buffer grows

## Why?

Tried to output samples as json and it failed

## Checklist

### DIFF
--- a/metrics/tags_test.go
+++ b/metrics/tags_test.go
@@ -97,10 +97,7 @@ func TestBigTagSetMarshalJSON(t *testing.T) {
 	expextedJSON := `{"tag0":"value0","tag1":"value1","tag2":"value2","tag3":"value3","tag4":"value4","tag5":"value5","tag6":"value6","tag7":"value7"}`
 	assert.Greater(t, len(expextedJSON), 128)
 
-	r := NewRegistry()
-	root := r.RootTagSet()
-
-	tags := root
+	tags := NewRegistry().RootTagSet()
 	for i := range 8 {
 		tags = tags.With(fmt.Sprintf("tag%d", i), fmt.Sprintf("value%d", i))
 	}
--- a/metrics/tags_test.go
+++ b/metrics/tags_test.go
@@ -94,8 +94,8 @@ func TestTagSets(t *testing.T) {
 func TestBigTagSetMarshalJSON(t *testing.T) {
 	t.Parallel()
 
-	expextedJSON := `{"tag0":"value0","tag1":"value1","tag2":"value2","tag3":"value3","tag4":"value4","tag5":"value5","tag6":"value6","tag7":"value7"}`
-	assert.Greater(t, len(expextedJSON), 128)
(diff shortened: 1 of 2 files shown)


mstoykov (dismissed):
Thank you for the contribution :bow: 

Looking at the code I see what the problem is, but can't figure out how have we not hit this earlier for the last 3 years. I also can not reproduce it locally :( 

Do I understand correctly that you produced this, just by using `-o json` ?

metrics/tags_test.go
```diff
@@ -90,6 +91,25 @@ func TestTagSets(t *testing.T) {
...
+func TestBigTagSetMarshalJSON(t *testing.T) {
+	t.Parallel()
+
+	expextedJSON := `{"tag0":"value0","tag1":"value1","tag2":"value2","tag3":"value3","tag4":"value4","tag5":"value5","tag6":"value6","tag7":"value7"}`
+	assert.Greater(t, len(expextedJSON), 128)
+
+	r := NewRegistry()
+	root := r.RootTagSet()
+
+	tags := root
```
mstoykov, line 103:
```suggestion
	tags := NewRegistry().RootTagSet()
```
No need for the extra variable if we won't use it. :)

metrics/tags_test.go
```diff
@@ -90,6 +91,25 @@ func TestTagSets(t *testing.T) {
 	assert.True(t, tags4.Contains(tags4))
 }
 
+func TestBigTagSetMarshalJSON(t *testing.T) {
+	t.Parallel()
+
+	expextedJSON := `{"tag0":"value0","tag1":"value1","tag2":"value2","tag3":"value3","tag4":"value4","tag5":"value5","tag6":"value6","tag7":"value7"}`
+	assert.Greater(t, len(expextedJSON), 128)
```
mstoykov, line 98: Hm is this the requirement? For the end tags to be over 128 bytes? 

Pretty sure I have way bigger ones with `-o json` on my terminal and that still doesn't break it ;(

dvordrova: Yes it happens when initial buffer size is not enough
https://github.com/mailru/easyjson/blob/master/buffer/pool.go#L19

dvordrova:
> Looking at the code I see what the problem is, but can't figure out how have we not hit this earlier for the last 3 years. I also can not reproduce it locally :(
> 
> Do I understand correctly that you produced this, just by using `-o json` ?

I reproduced it when wrote plugin for xk6 and needed some debug output for samples, not sure that is the thing with `-o json`

mstoykov:
Looked into it more and the json output doesn't use this at all. 

I guess this was only used for the old cloud output, and I guess it had bugs then as well :facepalm: 

Given the current lack of usage, I wonder if we should be going through the trouble of using easyjson for htis at all :thinking: 

cc @grafana/k6-core

olegbespalov (approved):

mstoykov (approved):

Issue #4640, opened 2025-03-20
Error when using `--summary-export` on latest k6 master docker image.


brendenehlers:
### Brief summary

Unable to get K6 results when the `--summary-export` option is provided on the latest master docker image.

### k6 version

v0.57.0

### OS

linux/arm64

### Docker version and image (if applicable)

master:latest

### Steps to reproduce the problem

On the `grafana/k6:master` docker image, running this test:

```
import http from ‘k6/http’;

export default function () {
    http.get(‘https://test.k6.io/’);
}
```

with this command:

```
k6 run test.js  --summary-export results.json
```

results in this error message:
```
time=“2025-03-20T15:51:51Z” level=error msg=“failed to handle the end-of-test summary” error=“unexpected error while generating the summary: TypeError: jannot read property ‘groups’ of undefined\n\tat transformGroup (<eval>:1411:33(4))\n\tat oldJSONSummary (<eval>:1450:44(27))\n\tat <eval>:1467:53(27)\n”
```

This issue is blocking my team from running our K6 tests before going live.

Running `k6 --version` shows this message:
```
k6 v0.57.0 (commit/6e82967f3d, go1.24.1, linux/arm64)
```

### Expected behaviour

Last week, the results would be saved to the results.json file and there would not be an error.

### Actual behaviour

This error is logged:
time=“2025-03-20T15:51:51Z” level=error msg=“failed to handle the end-of-test summary” error=“unexpected error while generating the summary: TypeError: jannot read property ‘groups’ of undefined\n\tat transformGroup (<eval>:1411:33(4))\n\tat oldJSONSummary (<eval>:1450:44(27))\n\tat <eval>:1467:53(27)\n”

mstoykov:
Thank you for reporting  @brendenehlers 🙇 

I have opened https://github.com/grafana/k6/pull/4642 , but I think this might need more work.

PR #4647, opened 2025-03-21
Summary: Adjust JSON output to the new data structure

joanlopez:
## What?

It adjusts the code responsible for the JSON format of the end-of-test summary to be compatible with the new data structure introduced in https://github.com/grafana/k6/pull/4089.

## Why?

As pointed out in https://github.com/grafana/k6/issues/4640, the new data structure for summary data introduced in the aforementioned PR isn't compatible with the existing code, _completely_ breaking that feature.

We quickly fixed the feature with https://github.com/grafana/k6/pull/4642, which was nice.

However, I think we can do better, and keep the JSON summary structure much closer to the existing one. 
Complete compatibility isn't possible (or isn't worth the effort), but there's room for improvement.

In summary, the new code produces an JSON document that looks like:

```
{
  "thresholds": {...},
  "metrics": {...},
  "root_group": {
    "name": "...",
    "metrics": {...},
    "groups": {...},
    "scenarios": {...}
  }
}
```

So, the main remarkable changes are:
- _Checks_ and _groups_ now don't have `path` and `id` attributes.
- Now it includes _thresholds_ as a separate -root-level- section, instead of being part of the _metrics_ object.
- _Groups_ also contain _metrics_ inside, so "partial" metrics are included for groups other than the "root group".
- _Scenarios_ are also included, following the same structure as _groups_.
- Some metrics previously added (e.g. `group_duration`), now aren't unless you use `--summary-mode=full`.
- Sub-groups checks aren't included unless you use `--summary-mode=full`. 

If you want to compare a concrete example, based on the same test case used for the revamped summary, you can have a look at these two files attached:
- [results.json](https://github.com/user-attachments/files/19427110/results.json)
- [results-old.json](https://github.com/user-attachments/files/19427112/results-old.json)

You can also see the comparison with https://nosmileface.dev/jsondiff/ in the following screenshot:

screencapture-nosmileface-dev-jsondiff-2025-03-24-10_48_07

However, certain diffs aren't very easy, because the order of most of the information displayed in the legacy (old) format is not predictable (every execution it has different order). I tried to make the new output a bit more predictable by sorting certain keys.

## Checklist

### DIFF
--- a/internal/js/runner.go
+++ b/internal/js/runner.go
@@ -394,7 +394,8 @@ func (r *Runner) HandleSummary(
 		return nil, fmt.Errorf("unexpected error did not get a callable summary wrapper")
 	}
 
-	wrapperArgs := prepareHandleWrapperArgs(vu, noColor, enableColors, callbackResult, handleSummaryDataAsValue)
+	wrapperArgs := prepareHandleWrapperArgs(
+		vu, noColor, enableColors, callbackResult, handleSummaryDataAsValue, legacy != nil)
 	rawResult, _, _, err := vu.runFn(summaryCtx, false, handleSummaryWrapper, nil, wrapperArgs...)
 
 	if deadlineError := r.checkDeadline(summaryCtx, consts.HandleSummaryFn, rawResult, err); deadlineError != nil {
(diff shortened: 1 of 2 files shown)


joanlopez:
Discarded, in favor of #4649

Issue #4265, opened 2024-10-30
Warn about usage `saltLength: 0` for RSA PSS sign/verify


olegbespalov:
# What?

While implementing support of RSA grafana/xk6-webcrypto#85 we've found that standard Golang's SDK [doesn't provide support of the saltLength: 0](https://github.com/golang/go/blob/master/src/crypto/rsa/pss.go#L293-L297), instead it always tries to generate [salt with the maximum length](https://github.com/golang/go/blob/master/src/crypto/rsa/pss.go#L248-L250). 

This led us for the patch to WebPlatfrom tests:
https://github.com/grafana/xk6-webcrypto/blob/main/webcrypto/tests/wpt-patches/WebCryptoAPI__sign_verify__rsa.js.patch

Since some use cases could actually try to use the lengths of salt 0 we need to issue a log warning user that the actual behavior of our implementation is different.

# Why?

It improves UX and makes implementation predictable.

vincentandreas:
hi @olegbespalov I'm interested to contribute this, but, I need time to understanding and implement this. is it okay? maybe around 2 weeks

olegbespalov:
Hey @vincentandreas ! Sure, it's totally fine 👍

mstoykov:
Looking at the PR I wonder if we shouldn't actually throw an exception - we are *not* doing what is expected, so us just logging a warning seems strange. Especially with stuff such as crypto this will have completely different results - right? 

So it seems like we either have to fix *it* or throw an exception and maybe fix it later. Given that fixing it likely will be quite involved, I propose we just throw an exception.

Do we have any idea how common this behaviour is? Is this like a thing that is there just because it can be or is something that is used? 


cc @grafana/k6-core

oleiade:
@mstoykov I have no hard facts around it, but my main intuition as to why one would want 0-length salt is for testing purposes, and keeping the output as deterministic as possible. But I have no certainties that's the reason.

Not a strong opinion: my preference would go to throwing an exception as it clearly indicates that's both something we're unable to achieve, and it's also something that users should probably be kept from doing, until there's a hard requirement for it.

inancgumus:
+1 for throwing an exception. It's also worth documenting this fact in our API docs.

mstoykov:
There seems to be aggreement on this, so please @vincentandreas change the code to throw an exception in those cases and please add a test that this is what happens.

vincentandreas:
[...]

PR #4659, opened 2025-03-26
Fix : add warning in RSA PSS webcrypto

vincentandreas:
## What?

Implement warning in RSA PSS, when given salt length = 0, in Sign / Verify function 

## Why?
It improves UX and makes implementation predictable.

## Checklist

## Related PR(s)/Issue(s)

https://github.com/grafana/k6/issues/4265 

Yes

### DIFF
--- a/internal/js/modules/k6/webcrypto/rsa.go
+++ b/internal/js/modules/k6/webcrypto/rsa.go
@@ -9,6 +9,8 @@ import (
 	"fmt"
 	"math/big"
 
+	"github.com/sirupsen/logrus"
+
 	"github.com/grafana/sobek"
 )
 
@@ -386,6 +388,11 @@ func (rsasv *RSAPssParams) Sign(key CryptoKey, data []byte) ([]byte, error) {
 
 	hashedData := hash.New()
 	hashedData.Write(data)
+	if rsasv.SaltLength == 0 {
+		logger := logrus.New()
+		warnMsg := "K6 RSA-PSS use standard Golang SDK, doesn't support salt length=0. Sign result might be different!"
+		logger.Warn(warnMsg)
+	}
 
 	signature, err := rsa.SignPSS(rand.Reader, rsaKey, hash, hashedData.Sum(nil), &rsa.PSSOptions{
 		SaltLength: rsasv.SaltLength,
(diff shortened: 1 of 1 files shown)


mstoykov (changes requested):
Thank you for the PR :bow: 

Unfortunately I can not approve it as is.

First - we definitely should create a new logger and use it , but use the one that is already created in k6. I am not certain if at this part of the code you have easy access to it though, so that might be a bit of a problem.

Given that this basically does something completely different from what is expected I do wonder if it won't actually be better for this to actually be an exception, which is alos way easier to implement . You need to just return an error.

I will raise this in the original issue

vincentandreas:
Sure @mstoykov I'll wait other discussion related to this, thx for your suggestion

mstoykov:
After over a week it seems like the agreement is to have this as an exception instead of warning. 

You can do that by returning an error - Sobek (the JS VM) will take this as an exception. I would also prefer if there are some tests added to check this.

mstoykov:
:wave: @vincentandreas 

Please rebase on latest master as https://github.com/grafana/k6/pull/4701 is needed to see if the test are working.

This was the problem we were having but now it should fail on test failure.

Sorry for the inconvience and the time it took, but it was kind of hard to figure out why was thsi working before :(

inancgumus:
Hi @vincentandreas, will you be able to work on the changes? Thanks!

~_PS: test-current-cov fails, and I'm working on a fix: #4837_~ Fixed.

mstoykov:
Hi @vincentandreas , first - thank you for the work you have done :bow: 

This PR has been open for a while now. We would like to get a fix for the issue, but as discussed before, instead of logging we would like to throw an exception. And we would also would like a test about it.

If you are not able to continue, that is fine. We can take over, but prefer to let you finish the work. There is no rush on getting this done instantly, we just want to make certain that we do not take over while you are working it or expected some feedback from us.

mstoykov:
I will close this PR, due to inactivity :(

Thank you for bringing this up and helping figureing out the correct approach.

Given the low priority of the issue and that it seems like a good first issue we will be leaving it open instead of taking over. At least for now.

If you @vincentandreas or somebody else wants to work on, you are welcome to open a PR.

PR #4666, opened 2025-04-01
Log metrics & traces flush to cloud as 'trace'

joanlopez:
## What?

It creates log lines that announce that metrics and traces have been flushed to the cloud at the `trace` level instead of `debug`.

## Why?

Because these operations are executed many times per test execution, these lines are repeated repeatedly, generating noise rather than bringing value. We generally use `debug` as the "default" log level when monitoring our systems and/or in integration tests, but these log lines are generally filtered out.

In case of error while flushing, the error will be handled and reported as such (with `error` log level).

## Checklist

## Related PR(s)/Issue(s)

N/A

### DIFF
--- a/output/cloud/expv2/output.go
+++ b/output/cloud/expv2/output.go
@@ -267,7 +267,7 @@ func (o *Output) flushMetrics() {
 		return
 	}
 
-	o.logger.WithField("t", time.Since(start)).Debug("Successfully flushed buffered samples to the cloud")
+	o.logger.WithField("t", time.Since(start)).Trace("Successfully flushed buffered samples to the cloud")
 }
 
 func (o *Output) runFlushRequestMetadatas() {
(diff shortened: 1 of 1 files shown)


oleiade (approved):

mstoykov:
Given that the lack of this lines was how we diagnosed some recent problems. Or at least noticed that we stopped gettign them, I am not really certain removing them is a good idea.

If I remember correctly we do not actually log trace logs at all so that will mean that moving them to `trace` is also likely exactly the same as removing them. 

To be honest I do not think that the usage of this log in the way we had to diagnose it is ... good. But it happened to be the only indication that this was a problem when getting to the error logs was taking too long and the whole process was killed.

mstoykov:
It meant that due to the error being timeout and this is being retries + the timeout on our side being 1 minute, it pracitcally never logged the errors as we would have killed k6 before all of this goes through.

So if anything logging more for when we are going to be retrying a call might be what is needed. And then maybe this lines can be removed.

mstoykov (approved):

PR #4671, opened 2025-04-03
Integrate Binary provisioning

pablochacin:
## What?

Integrate into k6 the logic for dynamically provision a k6 custom binary that includes the dependencies (extensions) required by a test.

This change set introduces the launcher as the new starting point for executing k6. The launcher checks if a binary provisioning is enabled and if it is required to execute the command. If not, it delegates the executing to the regular execution entry point.

Binary provisioning is required if the command processes a script or archive (`run`, `archive`, ...) and the input has a dependency that cannot be resolved with the current k6 binary. This is resolved by the `k6deps` library that scans the input for imports of k6 extensions.

If a new binary is required, it is downloaded from the build service using the `k6provider` library. This requires a cloud token for authentication. Binaries are stored in a local cache, so subsequent requests will be served from the local file system instead of downloading it.

Note: This is a work in progress. It is fully functional, but a few tasks are still pending:

TODO:
- [x] Minimize the dependency diff, precisely cobra and Prometheus dependencies, which should be defined by k6

Also, some features are intended for follow-up PRs:

FOLLOW-UP:
- [ ] Reuse existing logic for reading k6 config
- [ ] Handle stdin as input (this requires creating a tmp file, and it's not clear how to do this early in the execution path)
- [ ] handle URLs as input (same as above and also has some added complexity of identifying the URL of the input as other arguments may also be URLs)
- [ ] When checking if binary provisioning is required, consider case when binary has built in extensions
- [ ] Better visibility for cache usage & management

## Why?

We'd like k6 to be able to run scripts with extensions without recompiling the binary

## Checklist


internal/cmd/root.go
```diff
@@ -127,16 +127,7 @@ func (c *rootCommand) execute() {
...
-func Execute() {
-	gs := state.NewGlobalState(context.Background())
-	newRootCommand(gs).execute()
-}
-
-// ExecuteWithGlobalState runs the root command with an existing GlobalState.
-// This is needed by integration tests, and we don't want to modify the
-// Execute() signature to avoid breaking k6 extensions.
-func ExecuteWithGlobalState(gs *state.GlobalState) {
+func Execute(gs *state.GlobalState) {
```
codebien, line 130: Can we isolate this change in a separated pull request, please? It doesn't sound really a requirement to have it here. Or am I missing something?

codebien (commented):
@pablochacin I didn't catch big blockers, except the UX for errors, as you already know.

internal/launcher/launcher.go
```diff
@@ -0,0 +1,156 @@
...
+	"go.k6.io/k6/internal/build"
+	k6Cmd "go.k6.io/k6/internal/cmd"
+)
+
+// Execute runs the k6 command.
+func Execute() {
+	gs := state.NewGlobalState(context.Background())
+
+	gs.OSExit(newLauncher(gs).launch())
+}
```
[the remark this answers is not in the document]
pablochacin: Not sure what you refer to here. The `Execute` function? Can you please clarify?

cmd/state/state.go
```diff
@@ -189,5 +200,17 @@ func getFlags(defaultFlags GlobalFlags, env map[string]string) GlobalFlags {
...
+		result.BinaryProvisioning = true
+	}
+	if val, ok := env["K6_BUILD_SERVICE_URL"]; ok {
+		result.BuildServiceURL = val
+	}
+
+	// check if verbose flag is set
+	if slices.Contains(args, "-v") || slices.Contains(args, "--verbose") {
+		result.Verbose = true
+	}
```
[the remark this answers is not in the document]
pablochacin: My guess is that @olegbespalov  tried to minimize the blast radius of this chane set and I keep it. I'm not sure of how `rootCmdPersistentFlagSet` is used, but can do the change if you consider this is more convenient.

cmd/state/state.go
```diff
@@ -88,16 +89,6 @@ func NewGlobalState(ctx context.Context) *GlobalState {
 	}
 
 	env := BuildEnvMap(os.Environ())
-	_, noColorsSet := env["NO_COLOR"] // even empty values disable colors
```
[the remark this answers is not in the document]
pablochacin: No, it it not. Fixing.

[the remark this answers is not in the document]
pablochacin: It was moved to the `getFlags` function

internal/launcher/launcher.go
```diff
@@ -0,0 +1,156 @@
...
+		}
+	}
+
+	return 0, nil
+}
+
+// isCustomBuildRequired checks if the build is required
+// it's required if there is one or more non k6 dependencies
+// or if the required k6 version is not satisfied by the current binary's version
+// TODO: get the version of any built-in extension and check if they satisfy the dependencies
```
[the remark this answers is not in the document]
pablochacin: We are only checking if the version of `k6` currently running satisfies the constraints in the test (that is, k6 version pinning such as `use k6 > v0.56.0`). But in case the k6 binary is a custom k6 build with extensions, we are not checking those extensions and can be downloading a binary unnecessarily.

internal/launcher/launcher.go
```diff
@@ -0,0 +1,156 @@
...
+		if errors.As(err, &eerr) {
+			return eerr.ExitCode(), err
+		}
+	}
+
+	return 0, nil
+}
+
+// isCustomBuildRequired checks if the build is required
+// it's required if there is one or more non k6 dependencies
```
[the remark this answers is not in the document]
pablochacin: `k6` itself is a dependency as we allow version pinning. I will rephrase the comment.

internal/launcher/options.go
```diff
@@ -0,0 +1,80 @@
...
+	Collectors struct {
+		Cloud struct {
+			Token string `json:"token"`
+		} `json:"cloud"`
+	} `json:"collectors"`
+}
+
+// loadConfig loads the k6 config file from the given path or the default location.
+// if using the default location and the file does not exist, it returns an empty config.
+func loadConfig(gs *k6State.GlobalState) (k6configFile, error) {
```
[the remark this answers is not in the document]
pablochacin: Agree. This is the intention and is listed as a follow-up action.

oleiade:
Quick heads-up that I'm on it, and the week is a bit crazy and will likely have more time for a more thorough review beginning of next week 🙇🏻

internal/launcher/options.go
```diff
@@ -0,0 +1,80 @@
...
+import (
+	"encoding/json"
+	"fmt"
+
+	k6State "go.k6.io/k6/cmd/state"
+	"go.k6.io/k6/lib/fsext"
+)
+
+// Options contains the optional parameters for provisioning.
+type Options struct {
```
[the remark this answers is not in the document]
pablochacin: No. I changed and also renamed `NewOptions` as `newOptions`

oleiade:
You've been closer to the topic and involved in the hand-over @mstoykov, and I'm in a bit of a rush this week and the next one, if you're not too busy yourself, would you have capacity to take over the review from me?

mstoykov:
Unfortunately I am still not certain if the wpt test are not failing due to a bug in how we run and check them or if there is a bug in Sobek/k6

internal/cmd/root.go
pablochacin: This is done

internal/launcher/launcher.go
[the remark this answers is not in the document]
codebien: I would appreciate then if you can integrate the comment with explanation you provided me here. In addition, it seems this case is not tested on the `launcher_test.go`, can we add it? It would be good to see the test failing if we resolve the TODO. It means we have to change the test if we want to change to address the TODO.

cmd/state/state.go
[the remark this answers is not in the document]
codebien: So, if I'm understanding right, this was already redundant already before. I will look into creating a dedicated pull request for it, so we don't have to bother about it here.

internal/launcher/analyze.go
```diff
@@ -0,0 +1,76 @@
...
+		return "", false
+	}
+
+	// search for a command that requires binary provisioning and then get the target script or archive
+	for i, arg := range args {
+		if slices.Contains([]string{"run", "archive", "inspect", "cloud"}, arg) {
+			// Look for script files (non-flag arguments with .js, or .tar extension) in the reminder args
+			for _, arg = range args[i+1:] {
+				if strings.HasPrefix(arg, "-") {
+					// TODO: it may be we are using stdin as a source. Handle this case
```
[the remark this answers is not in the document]
pablochacin: I was hoping to be able to include this use case here but I realized it is possibly too big, so I will add the log

internal/launcher/launcher.go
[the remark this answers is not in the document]
pablochacin: Not sure I follow your suggestion regarding adding a test case.

The test case for testing this TODO would be having a binary that includes an extension and checking we don't call the binary provisioning. Until then, any dependency to an extension will require the binary provisioning (and this is tested)

But we presently are testing just k6 binary, so we can't simulate this behavior. We can include a way to mock this in the test somehow, but I'm not sure this makes sense here given that it is not implemented.

internal/launcher/launcher.go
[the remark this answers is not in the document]
pablochacin: This is addressed. Please let me know if the comment is now more clear.

internal/launcher/launcher.go
```diff
@@ -0,0 +1,156 @@
...
+		fallback:  k6Cmd.ExecuteWithGlobalState,
+		provision: k6buildProvision,
+		run:       runK6Cmd,
+	}
+}
+
+// launch executes k6 either by launching a provisioned binary or defaulting to the
+// current binary it this is not necessary.
+// Returns an int to be used as exit code.
+// If the fhe fallback is called, it can exit the process so don't assume it will return
```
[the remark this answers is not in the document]
pablochacin: This is done.

cmd/state/state.go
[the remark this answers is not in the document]
codebien: https://github.com/grafana/k6/pull/4686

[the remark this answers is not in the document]
codebien: I resolve this as it will be handled in the dedicated pull request

internal/launcher/analyze.go
```diff
@@ -0,0 +1,80 @@
...
+		gs.Logger.
+			Warn("binary provisioning is not supported when using input from stdin")
+		return dopts
+	}
+
+	if _, err := gs.FS.Stat(scriptname); err != nil {
+		gs.Logger.
+			WithField("scriptname", scriptname).
+			WithError(err).
+			Error("failed to stat script")
```
codebien, line 35:
```suggestion
			Error("The requested test script's file is not available on the file system. Make sure that the correct name has been passed or that the file still exists on the file system.")
```

Something similar is required to have a better user experience.

internal/launcher/analyze.go
```diff
@@ -0,0 +1,80 @@
...
+
+	if scriptname == "-" {
+		gs.Logger.
+			Warn("binary provisioning is not supported when using input from stdin")
+		return dopts
+	}
+
+	if _, err := gs.FS.Stat(scriptname); err != nil {
+		gs.Logger.
+			WithField("scriptname", scriptname).
```
codebien, line 33:
```suggestion
			WithField("path", scriptname).
```
or `location`

internal/launcher/launcher.go
```diff
@@ -0,0 +1,158 @@
...
+	}
+}
+
+// launch executes k6 either by launching a provisioned binary or defaulting to the
+// current binary it this is not necessary.
+// If the fhe fallback is called, it can exit the process so don't assume it will return
+func (l *launcher) launch() {
+	// if binary provisioning not enabled, continue with regular k6 execution path
+	if !l.gs.Flags.BinaryProvisioning {
+		l.gs.Logger.Debug("binary provisioning disabled")
```
codebien, line 49:
```suggestion
		l.gs.Logger.Debug("Binary provisioning feature is disabled")
```

internal/launcher/launcher.go
```diff
@@ -0,0 +1,158 @@
...
+		l.fallback(l.gs)
+		return
+	}
+
+	l.gs.Logger.Info("trying to provision binary")
```
codebien, line 54:
```suggestion
	l.gs.Logger.Info("Binary provisioning feature is enabled. If it's required, k6 will try to provision a new binary.")
```

internal/launcher/launcher.go
```diff
@@ -0,0 +1,158 @@
...
+
+	deps, err := k6deps.Analyze(newDepsOptions(l.gs, l.gs.CmdArgs[1:]))
+	if err != nil {
+		l.gs.Logger.
+			WithError(err).
+			Error("failed to analyze dependencies, can't try binary provisioning, please report this issue")
```
codebien, line 60:
```suggestion
			Error("Failed to analyze the required dependencies. Please, make sure to report this issue by opening a bug report.")
```

internal/launcher/launcher.go
```diff
@@ -0,0 +1,158 @@
...
+		l.gs.OSExit(1)
+	}
+
+	// binary provisioning enabled but not required by this command
+	// continue with regular k6 execution path
+	if !isCustomBuildRequired(build.Version, deps) {
+		l.gs.Logger.
+			Debug("binary provisioning not required")
```
codebien, line 68:
```suggestion
			Debug("The current k6 binary already satisfies all the required dependencies, it isn't required to provision a new binary.")
```

internal/launcher/launcher.go
```diff
@@ -0,0 +1,158 @@
...
+		l.fallback(l.gs)
+		return
+	}
+
+	l.gs.Logger.
+		WithField("deps", deps).
```
codebien, line 74: Are we sure of the print here? Is it safe? Is it a nice and understandable print?

internal/launcher/launcher.go
```diff
@@ -0,0 +1,158 @@
...
+	l.launchCustomBuild(deps)
+}
+
+func (l *launcher) launchCustomBuild(deps k6deps.Dependencies) {
+	// get the k6 binary from the build service
+	binPath, versions, err := l.provision(l.gs, deps)
+	if err != nil {
+		l.gs.Logger.
+			WithError(err).
+			Error("failed to fetch a binary with required dependencies, please report this issue")
```
codebien, line 86:
```suggestion
			Error("Failed to provision a new k6 binary with required dependencies. Please, make sure to report this issue by opening a bug report.")
```

internal/launcher/launcher.go
```diff
@@ -0,0 +1,158 @@
...
+		l.gs.OSExit(1)
+		// in tests calling l.gs.OSExit does not ends execution so we have to return
+		return
+	}
+
+	l.gs.Logger.
+		Info("k6 has been provisioned with version(s) ", versions)
```
codebien, line 93:
```suggestion
		Info("A new k6 binary has been provisioned with version(s): ", versions)
```

internal/launcher/launcher.go
```diff
@@ -0,0 +1,158 @@
...
+
+	l.gs.Logger.Debug("launching provisioned k6 binary")
```
codebien, line 95:
```suggestion
	l.gs.Logger.Debug("Launching the new provisioned k6 binary")
```

internal/launcher/launcher.go
```diff
@@ -0,0 +1,158 @@
...
+
+	// execute provisioned binary
+	if rc, err := l.run(l.gs, binPath); err != nil {
+		l.gs.Logger.Error(err)
```
codebien, line 99:
```suggestion
		l.gs.Logger.WithErr(err).Error("Failed to run the new provisioned k6 binary")
```

internal/launcher/launcher_test.go
```diff
@@ -0,0 +1,284 @@
...
+
+export default function() {
+  let res = http.get('https://quickpizza.grafana.com');
+  check(res, { "status is 200": (res) => res.status === 200 });
+  sleep(1);
+}
+`
+)
+
+func Test_Launcher(t *testing.T) {
```
codebien, line 96:
```suggestion
func TestLauncherLaunch(t *testing.T) {
```

internal/launcher/provision.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/grafana/k6provider"
+
+	"go.k6.io/k6/cmd/state"
+)
+
+// k6buildProvision returns the path to a k6 binary that satisfies the dependencies and the list of versions it provides
+func k6buildProvision(gs *state.GlobalState, deps k6deps.Dependencies) (string, string, error) {
+	opt := newOptions(gs)
+	if opt.BuildServiceToken == "" {
+		return "", "", errors.New("Need a k6 cloud token for binary provisioning. " + //nolint:stylecheck //false positive
```
codebien, line 19:
```suggestion
		return "", "", errors.New("k6 cloud token is required when Binary provisioning feature is enabled. " +
```

internal/launcher/provision.go
```diff
@@ -0,0 +1,47 @@
...
+			"Setting K6_CLOUD_TOKEN environment variable or executing k6 cloud login is required.")
```
codebien, line 20:
```suggestion
			"Set K6_CLOUD_TOKEN environment variable or execute the k6 cloud login command.")
```

internal/launcher/analyze.go
```diff
@@ -0,0 +1,80 @@
...
+)
+
+// newDepsOptions returns the options for dependency resolution.
+// Presently, only the k6 input script or archive (if any) is passed to k6deps for scanning.
+// TODO: if k6 receives the input from stdin, it is not used for scanning because we don't know
+// if it is a script or an archive
+func newDepsOptions(gs *state.GlobalState, args []string) *k6deps.Options {
+	dopts := &k6deps.Options{
+		LookupEnv: func(key string) (string, bool) { v, ok := gs.Env[key]; return v, ok },
+		// TODO: figure out if we need to set FindManifest
```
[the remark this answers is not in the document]
pablochacin: I think the question here (this is a comment left by Oleg) is if we want to support manifests in k6. That is, defining defaults for extensions if not specified in the script. 

If so, maybe we need to provide `k6deps` with a function that searches for that manifest in the context of the script.

My opinion is that this should bot be part of this PR.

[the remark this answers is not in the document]
codebien: Ok, if this is undefined, then it might be better to just drop the comment.

internal/launcher/launcher.go
```diff
@@ -0,0 +1,158 @@
...
+// TODO: get the version of any built-in extension and check if they satisfy the dependencies
+func isCustomBuildRequired(baseK6Version string, deps k6deps.Dependencies) bool {
+	if len(deps) == 0 {
+		return false
+	}
+
+	// Early return if there are multiple dependencies
+	if len(deps) > 1 {
+		return true
+	}
```
joanlopez, line 128: I might be missing some context (around `k6deps`), but what's considered a dependency here? How do you know what extensions are available from the current binary?

Let's say this code is being executed by a binary compiled with `xk6` with support for `xk6-my-custom-extension`. Will it have `len(deps) > 1` anyway? 🤔

joanlopez: If that's what's stated by the `TODO` in the code comment at the very beginning (together with the function definition), then I think this is likely a must before actually being ready to merge this PR? 🤔 

Otherwise, we may require a custom build for dozens of cases where it's not really needed. And I think this is quite critical because custom builds require network -downloading the new binary-, setting up cloud token, etc. Not just a bit of extra work.

internal/launcher/launcher.go
```diff
@@ -0,0 +1,158 @@
...
+}
+
+// runK6Cmd runs the k6 binary passing the original arguments
+func runK6Cmd(gs *state.GlobalState, binPath string) (int, error) {
+	cmd := exec.CommandContext(gs.Ctx, binPath, gs.CmdArgs[1:]...) //nolint:gosec
+	cmd.Stderr = gs.Stderr
+	cmd.Stdout = gs.Stdout
+	cmd.Stdin = gs.Stdin
+
+	// disable binary provisioning to avoid a provisioning loop
```
joanlopez, line 103: Why do we need this? It looks like a "patch", other than how it should really behave. Is it because so far we're not able to correctly detect if the dependencies (extensions used) are really support by the current binary? 🤔 

Any other reason?

internal/launcher/launcher_test.go
```diff
@@ -0,0 +1,284 @@
...
+	"github.com/grafana/k6deps"
+	"github.com/spf13/afero"
+	"github.com/stretchr/testify/assert"
+	"go.k6.io/k6/cmd/state"
+	"go.k6.io/k6/internal/build"
+	"go.k6.io/k6/internal/cmd/tests"
+)
+
+// launcherFixture mocks the functions used by the launcher and tracks their execution
+type luncherFixture struct {
```
joanlopez, line 19:
```suggestion
type launcherFixture struct {
```

internal/launcher/launcher_test.go
```diff
@@ -0,0 +1,284 @@
...
+
+export default function() {
+  let res = http.get('https://quickpizza.grafana.com');
+  check(res, { "status is 200": (res) => res.status === 200 });
+  sleep(1);
+}
+`
+	// FIXME: when the build version is a prerelease (e.g v1.0.0-rc1), k6deps fails to parse this pragma
+	// and creates an invalid constrain that is ignored by the test.
+	// see https://github.com/grafana/k6deps/issues/91
```
joanlopez, line 83: Would be possible to fix this before moving forward with this PR? 🤔 Although, I guess we could go without it, tho..

internal/launcher/launcher.go
```diff
@@ -0,0 +1,158 @@
...
+	gs := state.NewGlobalState(context.Background())
+
+	newLauncher(gs).launch()
+}
+
+// launcher is a k6 launcher
+type launcher struct {
+	gs *state.GlobalState
+	// function to fall back if binary provisioning is not required
+	fallback func(gs *state.GlobalState)
```
joanlopez, line 19: nit; I'm not really convinced that `fallback` is a good name for this, as I'd expect not to have to provision the binary the "happy" and most common path.

internal/launcher/options.go
```diff
@@ -0,0 +1,81 @@
...
+)
+
+// options contains the optional parameters for provisioning.
+type options struct {
+	// BuildServiceURL contains the URL of the k6 build service to be used.
+	// If the value is not nil, the k6 binary is built using the build service instead of the local build.
+	BuildServiceURL string
+
+	// BuildServiceToken contains the token to be used to authenticate with the build service.
+	// Defaults to K6_CLOUD_TOKEN environment variable is set, or the value stored in the k6 config file.
```
joanlopez, line 18:
```suggestion
	// Defaults to K6_CLOUD_TOKEN environment variable if set, or the value stored in the k6 config file.
```

internal/launcher/options.go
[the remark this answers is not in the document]
joanlopez: I'd like to see this, yeah!

joanlopez (commented):
Some of the things I mention as "likely a blocker" are likely acceptable considering for now this will be behind a feature toggle (enabled through environment variables) 👍🏻

internal/cmd/root.go
```diff
@@ -134,8 +134,7 @@ func Execute() {
 }
 
 // ExecuteWithGlobalState runs the root command with an existing GlobalState.
-// This is needed by integration tests, and we don't want to modify the
-// Execute() signature to avoid breaking k6 extensions.
+// We keep it don't want to modify the Execute() signature to avoid breaking k6 extensions.
```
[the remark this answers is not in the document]
pablochacin: The reason for the change in the comment is that before this function existed only for integration tests but now it is use for the actual execution of k6. I actually suggested removing the `Execute` function but @codebien suggested doing this in another PR. I'm open to bring that change back.

cmd/execute.go
```diff
@@ -1,10 +1,10 @@
 // Package cmd is here to provide a way for xk6 to build a binary with added extensions
 package cmd
 
-import "go.k6.io/k6/internal/cmd"
+import "go.k6.io/k6/internal/launcher"
 
 // Execute exectues the k6 command
 // It only is exported here for backwards compatibility and the ability to use xk6 to build extended k6
 func Execute() {
-	cmd.Execute()
```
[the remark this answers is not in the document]
pablochacin: I suggested removing the `Execute` function but @codebien suggested doing this in another PR. I'm open to bringing that change back.

internal/launcher/launcher_test.go
pablochacin: That requires changes in the implementation of `k6deps` (https://github.com/grafana/k6deps/issues/91) Not sure if this is severe enough to block this PR.

internal/launcher/analyze.go
```diff
@@ -0,0 +1,80 @@
...
+		dopts.Archive.Name = scriptname
+	} else {
+		dopts.Script.Name = scriptname
+	}
+
+	return dopts
+}
+
+// scriptNameFromArgs returns the file name passed as input and true if it's a valid script name
+func scriptNameFromArgs(args []string) (string, bool) {
```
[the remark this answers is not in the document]
pablochacin: This is a limitation of the approach in this PR. We can't reliably parse the arguments and know which is the script. For doing so we should have to integrate the binary provisioning into the root command execution. There's a PoC for that approach.

internal/launcher/launcher.go
```diff
@@ -0,0 +1,158 @@
...
+	}
+
+	l.gs.Logger.Info("trying to provision binary")
+
+	deps, err := k6deps.Analyze(newDepsOptions(l.gs, l.gs.CmdArgs[1:]))
+	if err != nil {
+		l.gs.Logger.
+			WithError(err).
+			Error("failed to analyze dependencies, can't try binary provisioning, please report this issue")
+		l.gs.OSExit(1)
```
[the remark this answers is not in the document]
pablochacin: I don't have a strong opinion. @codebien?

internal/launcher/launcher.go
pablochacin: The binary we are providing is a `k6` binary that is also capable of binary provisioning. If we don't disable the option, it will execute the same path as the original one: scan the input script, detect the dependencies, call binary provisioning, execute the binary ... which will repeat the loop.

internal/launcher/launcher.go
pablochacin: Even when I agree we should prevent this unnecessary binary provisioning (hence the TODO). However, it is important to consider that the binary provisioning is not enabled by default. If someone builds a custom binary with extensions, they have to enable the binary provisioning, what looks more like a misconfiguration.

internal/launcher/analyze.go
```diff
@@ -0,0 +1,80 @@
...
+	"strings"
+
+	"github.com/grafana/k6deps"
+	"go.k6.io/k6/cmd/state"
+)
+
+// newDepsOptions returns the options for dependency resolution.
+// Presently, only the k6 input script or archive (if any) is passed to k6deps for scanning.
+// TODO: if k6 receives the input from stdin, it is not used for scanning because we don't know
+// if it is a script or an archive
```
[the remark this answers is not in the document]
pablochacin: Yes, I did. But it seems to have a considerable context around it. The loader expects a map of `Fs`. I'm not sure if this approach can be implemented at this early stage. Besides, `k6deps` works on files using the `os` package, so passing a `Fs` won't work. That will require a change in the implementation of `k6deps`. Therefore, it seems a lager enough change to consider in a follow-up PR.

cmd/execute.go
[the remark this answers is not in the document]
codebien: I have the feeling @joanlopez was reviewing a previous version and not the latest where we have the launcher directly called. My guess is related to the fact that I see a few of comments on the code that we don't have anymore on the latest version.

@joanlopez what about the latest version here? Do you expect something different?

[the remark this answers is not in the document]
joanlopez: > I suggested removing the Execute function but @codebien suggested doing this in another PR. I'm open to bringing that change back.

Sounds good! 👍🏻

internal/launcher/launcher_test.go
pablochacin: https://github.com/grafana/k6/issues/4694

internal/launcher/analyze.go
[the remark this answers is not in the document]
pablochacin: There's an open issue for this and I'm already working on a PR to address it https://github.com/grafana/k6deps/issues/97

internal/launcher/analyze.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+
+	if scriptname == "-" {
+		gs.Logger.
+			Warn("binary provisioning is not supported when using input from stdin")
+		return dopts
+	}
+
+	if _, err := gs.FS.Stat(scriptname); err != nil {
+		gs.Logger.
```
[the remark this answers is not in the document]
pablochacin: I changed the structure of the code to attend to other comments, and this has changed, so in case of this error, we abort the execution, but the message is printed in the launcher.

internal/launcher/launcher.go
```diff
@@ -0,0 +1,155 @@
...
+// Launcher is a k6 Launcher
+type Launcher struct {
+	gs *state.GlobalState
+	// function to fall back if binary provisioning is not required
+	fallback func(gs *state.GlobalState)
+	// function to provision a k6 binary that satisfies the dependencies
+	provision func(*state.GlobalState, k6deps.Dependencies) (string, string, error)
+	// function to execute k6 binary
+	run func(*state.GlobalState, string) (int, error)
+}
```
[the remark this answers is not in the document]
pablochacin: I like this. I implemented the changes.

internal/launcher/launcher.go
```diff
@@ -0,0 +1,158 @@
...
+	// if binary provisioning not enabled, continue with regular k6 execution path
+	if !l.gs.Flags.BinaryProvisioning {
+		l.gs.Logger.Debug("binary provisioning disabled")
+		l.fallback(l.gs)
+		return
+	}
+
+	l.gs.Logger.Info("trying to provision binary")
+
+	deps, err := k6deps.Analyze(newDepsOptions(l.gs, l.gs.CmdArgs[1:]))
```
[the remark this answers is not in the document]
pablochacin: I implemented this. For now, any error will abort execution because I think this gives users of this experimental feature feedback early in the execution. However, we can decide to continue in some cases.

internal/launcher/options.go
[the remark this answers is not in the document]
codebien: I'll try to work on this, I leave the discussion open as a reminder for myself.

go.mod
```diff
@@ -17,7 +18,9 @@ require (
 	github.com/golang/protobuf v1.5.4
 	github.com/google/uuid v1.6.0
 	github.com/gorilla/websocket v1.5.3
-	github.com/grafana/sobek v0.0.0-20250320150027-203dc85b6d98
+	github.com/grafana/k6deps v0.2.5
+	github.com/grafana/k6provider v0.1.13
+	github.com/grafana/sobek v0.0.0-20250219104821-ed22af7a8d6c
```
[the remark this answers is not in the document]
codebien: It seems fixed, I resolve this.

internal/launcher/launcher.go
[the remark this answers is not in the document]
codebien: The current log is more verbose and informative. This seems resolved.

internal/launcher/launcher.go
codebien: >If someone builds a custom binary with extensions, they have to enable the binary provisioning for this to happen, which looks more like a misconfiguration.

@pablochacin extensions are modules but also outputs. If I interpret the code right, we aren't addressing any case for extension outputs. So there is still a valid case for this to happen.

If I need both module and output extensions, does it mean it won't work?

internal/launcher/launcher_test.go
codebien: @joanlopez does the epic address your request?

internal/launcher/launcher.go
pablochacin: There's a way using pragmas as explained in `k6exec` documentation:

```
The pragma syntax can also be used to specify an extension dependency that is not referenced in an import expression. A typical example of this is the Output type extension such as [xk6-top](https://github.com/szkiba/xk6-top):

"use k6 with top >= 0.1";
```

We could also scan the CLI for `--out` flags, but there's no way to provide information about the version required for the extension.

In any case, at this moment the extension catalog used by the binary provisioning does not include output extensions. I suggest we handle this case in a follow-up PR.

go.mod
```diff
@@ -7,6 +7,7 @@ toolchain go1.23.7
 require (
 	buf.build/gen/go/prometheus/prometheus/protocolbuffers/go v1.36.5-20240802094132-5b212ab78fb7.1
 	github.com/Azure/go-ntlmssp v0.0.0-20221128193559-754e69321358
+	github.com/Masterminds/semver/v3 v3.3.1
```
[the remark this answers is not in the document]
pablochacin: We are using in `customBuildRequired`  to check if current version satisfies the constrains (if any) for `k6` version:

```
	k6Ver, err := semver.NewVersion(baseK6Version)
	if err != nil {
		// ignore if baseK6Version is not a valid sem ver (e.g. a development version)
		return true
	}

	// if the current version satisfies the constrains, binary provisioning is not required
	return !k6Dependency.Constraints.Check(k6Ver)
	```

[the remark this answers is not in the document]
Comment (codebien) [reply go.mod:10]: >Maybe we could move this logic to k6deps.

Yeah, in particular, if we re-use the same logic across several services.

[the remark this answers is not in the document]
Comment (pablochacin) [reply go.mod:10]: No, we don't. That's why I implemented it here. Is this a blocking requirement for this PR?

File: internal/launcher/launcher_test.go [hunk 42]
```diff
@@ -0,0 +1,253 @@
...
+		script          string
+		k6Env           map[string]string
+		k6Cmd           string
+		k6Args          []string
+		expectProvision bool
+		provisionError  error
+		expectK6Run     bool
+		expectDefault   bool
+		k6ReturnCode    int
+		expectOsExit    int
```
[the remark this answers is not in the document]
pablochacin: With the exception of the missing script input, all tests share the same setup and validations. I think it makes sense to keep it in one test instead of duplicating almost exactly the same logic in multiple tests.

internal/launcher/runner.go
```diff
@@ -0,0 +1,67 @@
...
+	cmd.Stderr = gs.Stderr
+	cmd.Stdout = gs.Stdout
+	cmd.Stdin = gs.Stdin
+
+	// disable binary provisioning to avoid a provisioning loop.
+	// if we keep it enabled the k6 binary executed here will receive the same input script
+	// will analyze it and detect the dependencies, triggering the binary provisioning again
+	gs.Env["K6_BINARY_PROVISIONING"] = "false"
+
+	gs.Logger.Debug("Launching the new provisioned k6 binary")
```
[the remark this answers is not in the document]
pablochacin: Not sure what value it has. It is a path to the cache directory and includes the binary id; it can be quite noisy. I wouldn't add it.

internal/launcher/launcher.go
pablochacin: https://github.com/grafana/k6/issues/4720

internal/launcher/runner.go
```diff
@@ -0,0 +1,67 @@
...
+func (r *execRunner) run(gs *state.GlobalState) {
+	cmd := exec.CommandContext(gs.Ctx, r.binPath, gs.CmdArgs[1:]...) //nolint:gosec
+	cmd.Stderr = gs.Stderr
+	cmd.Stdout = gs.Stdout
+	cmd.Stdin = gs.Stdin
+
+	// disable binary provisioning to avoid a provisioning loop.
+	// if we keep it enabled the k6 binary executed here will receive the same input script
+	// will analyze it and detect the dependencies, triggering the binary provisioning again
+	gs.Env["K6_BINARY_PROVISIONING"] = "false"
```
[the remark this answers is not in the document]
pablochacin: This is actuallt a bug! We are not passing the environment variables to the subprocess, just changing in the current process's global state. I fixed this.

internal/launcher/analyze.go
```diff
@@ -0,0 +1,86 @@
...
+func scriptNameFromArgs(args []string) (string, bool) {
+	// return early if no arguments passed
+	if len(args) == 0 {
+		return "", false
+	}
+
+	// search for a command that requires binary provisioning and then get the target script or archive
+	for i, arg := range args {
+		switch arg {
+		case "run", "archive", "inspect", "cloud":
```
pablochacin, line 65: @codebien I realized this logic has a false positive `cloud login` that fails because there's no script associated with the command. I need to rework this logic. The more "corner" cases I find, the more I think we should explore the alternative integration of this logic into the root command as exemplified in this (now outdated) PR https://github.com/grafana/k6/pull/4689

internal/cmd/root.go
```diff
@@ -134,8 +134,7 @@ func Execute() {
...
 func ExecuteWithGlobalState(gs *state.GlobalState) {
 	newRootCommand(gs).execute()
```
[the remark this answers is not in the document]
codebien: I agree, we should call the launcher from here and rollback the change on `cmd.Execute`. I'm opening pull requests for it.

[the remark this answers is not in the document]
codebien: No, I was wrong. It is used from the Launcher to call the current binary's execution path. I guess we have to add a comment mentioning it.

internal/launcher/options.go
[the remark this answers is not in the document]
codebien: #4712 fixed it

cmd/state/state.go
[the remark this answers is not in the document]
codebien: @pablochacin we don't have the time to address it, we should add an issue to the epic for researching and exploring the alternative solution by using pre-run as you have in the other pull request. Can you open it, plaese?

codebien:
Hey @joanlopez, this should be now in a good state for one more review. Can you check it again, plaese?

cmd/state/state.go
```diff
@@ -193,5 +203,17 @@ func getFlags(defaultFlags GlobalFlags, env map[string]string) GlobalFlags {
 	if _, ok := env["K6_PROFILING_ENABLED"]; ok {
 		result.ProfilingEnabled = true
 	}
+	if env["K6_BINARY_PROVISIONING"] == "true" {
```
joanlopez, line 206: For consistency with other "flags", like `K6_PROFILING_ENABLED` right above, I'd suggest to use the presence of that environment variable as the condition to consider this feature to be enabled.

Or, in case we consider that risky, I'd at least accept any "truthy" value there. Not that it's not only that `K6_BINARY_PROVISIONING=1` won't enable it, but nor `K6_BINARY_PROVISIONING=True` or `K6_BINARY_PROVISIONING=TRUE` would, which in my humble opinion isn't very good user experience.

joanlopez: You can use `strconv.ParseBool` as reference/inspiration:
https://pkg.go.dev/strconv#ParseBool

internal/cmd/launcher.go
```diff
@@ -0,0 +1,348 @@
...
+	buffer := &bytes.Buffer{}
+	for dep, version := range deps {
+		buffer.WriteString(fmt.Sprintf("%s:%s ", dep, version))
+	}
+	return strings.Trim(buffer.String(), " ")
+}
+
+// extractToken gets the cloud token required to access the build service
+// from the environment or from the config file
+func extractToken(gs *state.GlobalState) (string, error) {
```
joanlopez, line 238: Is it intentional that while most (if not all) of the other Cloud commands can still read old configuration, this one can't, and just tells you that you don't have a valid token?

If this is a way to force users to move to the new configuration if they want to test the feature, then it's alright. 

But, if not, then we should fix it, because in my opinion is a bit confusing: I can successfully run a Cloud command, like `k6 cloud run`, but when I run it with binary provisioning it tells me that I don't have a valid token, while both rely on the same token.

codebien: >If this is a way to force users to move to the new configuration if they want to test the feature, then it's alright. It may cause some friction and confusion, but it's the "cost" of forcing users to move forward.

@joanlopez Yes, this was my intention. I wanted to reduce the already complex world around this feature. I guess it's an acceptable scenario. Let me know if you disagree.

joanlopez: If we all agree about the implications, it's fine! I see no major problems beyond what's obvious and aforementioned.

joanlopez:
cc/ @pablochacin @codebien 

I haven't explored yet the reason why it happens, but I realized that when I run `k6 cloud run` with binary provisioning, and I abort the execution (with `CMD+C`/`CTRL+C`), the inner process (the one that launches the downloaded binary) isn't properly stopped.

For instance, if I run:

> K6_BINARY_PROVISIONING=true ./dist/k6 cloud run playground/faker.js

(where `./dist/k6` is the binary I compiled with `go build` from this branch)

and while running, I run `ps aux | grep k6`, it shows two processes:

```
... ... <USER> <PID>, etc   /Users/<user>/Library/Caches/k6provider/c1dc905cee0631cf924c3eb98d96df3cbe90992f/k6 cloud run playground/faker.js
... ... <USER> <PID>, etc  ./dist/k6 cloud run playground/faker.js
```

Then, I cancel the main process, which is cancelled immediately, despite it shows:

```
INFO[0014] Stopping cloud test run in response to signal...  sig=interrupt
INFO[0014] Successfully sent signal to stop the cloud test, now waiting for it to actually stop... 
```

but when running `ps aux | grep k6` I still see the inner process:

```
... ... <USER> <PID>, etc   /Users/<user>/Library/Caches/k6provider/c1dc905cee0631cf924c3eb98d96df3cbe90992f/k6 cloud run playground/faker.js
```

cmd/state/state.go
```diff
@@ -164,10 +173,11 @@ func GetDefaultFlags(homeDir string) GlobalFlags {
 		ProfilingEnabled: false,
 		ConfigFilePath:   filepath.Join(homeDir, "k6", defaultConfigFileName),
 		LogOutput:        "stderr",
+		BuildServiceURL:  "https://ingest.k6.io/builder/api/v1",
```
joanlopez, line 176: Nit; Can we define this as a `const`, like `defaultConfigFileName`, please? So, this way we have the consts that belong to the default flags together.

internal/cmd/launcher.go
```diff
@@ -0,0 +1,348 @@
...
+	return &Launcher{
+		gs:              gs,
+		provision:       k6buildProvision,
+		commandExecutor: defaultExecutor,
+	}
+}
+
+// Launch executes k6 either by launching a provisioned binary or defaulting to the
+// current binary if this is not necessary.
+// If the fhe fallback is called, it can exit the process so don't assume it will return
```
joanlopez, line 56:
```suggestion
// If the fallback is called, it can exit the process so don't assume it will return
```

internal/cmd/launcher.go
```diff
@@ -0,0 +1,348 @@
...
+	if err != nil {
+		l.gs.Logger.
+			WithError(err).
+			Error("Binary provisioning is enabled but it failed to analyze the dependencies." +
+				" Please, make sure to report this issue by opening a bug report.")
+		l.gs.OSExit(1)
+		return // this is required for testing
+	}
+
+	// if the command does not have dependencies or a custom build
```
joanlopez, line 77:
```suggestion
	// if the command does not have dependencies or a custom build is required
```

internal/cmd/launcher.go
pablochacin: I don't have a strong opinion here. `k6exec` supported old configurations. I was expecting that by reusing `k6` logic we would inherit this ability. But I'm ok with not supporting it. Should we add an item to follow-up this or we consider it settled?

internal/cmd/launcher.go
```diff
@@ -0,0 +1,348 @@
...
+	cmd.Stderr = os.Stderr
+	cmd.Stdout = os.Stdout
+	cmd.Stdin = os.Stdin
+
+	// Copy environment variables to the k6 process and skip binary provisioning feature flag to disable it.
+	// If not disabled, then the executed k6 binary would enter an infinite loop, where it continuously
+	// process the input script, detect dependencies, and retrigger provisioning.
+	env := []string{}
+	for k, v := range gs.Env {
+		if k == "K6_BINARY_PROVISIONING" {
```
joanlopez, line 128: I'd say I have seen this literal string at least a couple of times along this PR, so I'd prefer if we can move this into a `const`, so prevent mistakes in the future, in case it changes, possible typos, etc.

internal/cmd/launcher.go
```diff
@@ -0,0 +1,348 @@
...
+		dopts.Archive.Name = scriptname
+	} else {
+		dopts.Script.Name = scriptname
+	}
+
+	return k6deps.Analyze(dopts)
+}
+
+// isScriptRequired searches for the command and returns a boolean indicating if it is required to pass a script or not
+func isScriptRequired(args []string) bool {
```
joanlopez, line 299: In case we run `k6 cloud run` with `--local-execution`, I think this should return `false`, as I guess it's equivalent to `k6 run`, isn't it? 🤔 
Or, do we intentionally want to run a custom binary locally in such case?

In any case, I'd suggest to add that case to `TestIsScriptRequired`, so we leave it explicit. Perhaps even accompanied with a code comment, as in my opinion isn't obvious.

pablochacin: `could run --local-execution`  is expected to run with binary provisioning. We are not making a difference with `cloud run`.

internal/cmd/launcher.go
```diff
@@ -0,0 +1,348 @@
...
+			if arg == "-" { // we are running a script from stdin
+				return arg
+			}
+			continue
+		}
+		if strings.HasSuffix(arg, ".js") ||
+			strings.HasSuffix(arg, ".tar") ||
+			strings.HasSuffix(arg, ".ts") {
+			return arg
+		}
```
joanlopez, line 343: I still think this isn't a good idea because:
1. k6 normally works despite of the file extension.
2. If I have binary provisioning enabled, it breaks the execution flow even when there's no dependency to analyze.
3. The error is confusing, because it says: `error="script not found"`, while arguably the problem is that the file extension isn't supported/recognized as a file script, or something along those lines.

In my opinion, we should try to fix at least 2 and 3 before merging PR. 1 is something we can work later, I guess - I'd appreciate if we can create and issue for that.

pablochacin:
@joanlopez Regarding the issue you mention with cancelling a cloud execution in the subprocess, the fact you see this message

```
INFO[0014] Stopping cloud test run in response to signal...  sig=interrupt
``` 

Means that the subprocess is actually being properly stopped by the parent. Otherwise, you woulnd' t see that message.

My guess is that you still see the process because as this message says, it is waiting for the cloud.:
```
INFO[0014] Successfully sent signal to stop the cloud test, now waiting for it to actually stop... 
```
I'm not sure what the expectation is in this case. 

Should the parent wait for the children? to finish. I'm not 100% sure, but I would expect this to be the case. So I'm not sure why the parent finishes first

cc @codebien

internal/cmd/launcher.go
pablochacin: My understanding is that we accepted this as a limitation for the preview. Addressing this issue in the current logic is complex and probably will still fail in some cases. It is documented here: https://github.com/grafana/k6/issues/4727

joanlopez:
> I'm not 100% sure, but I would expect this to be the case. 

Same here, that's why I asked.

internal/cmd/launcher.go
pablochacin: There's no fallback anymore. I fixed the command to make reference to the `commandExecutor`

internal/cmd/launcher.go
```diff
@@ -0,0 +1,348 @@
...
+	cmd := exec.CommandContext(gs.Ctx, b.path, gs.CmdArgs[1:]...) //nolint:gosec
+
+	// we pass os stdout, err, in because passing them from GlobalState changes how
+	// the subprocess detects the type of terminal
+	cmd.Stderr = os.Stderr
+	cmd.Stdout = os.Stdout
+	cmd.Stdin = os.Stdin
+
+	// Copy environment variables to the k6 process and skip binary provisioning feature flag to disable it.
+	// If not disabled, then the executed k6 binary would enter an infinite loop, where it continuously
```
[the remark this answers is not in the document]
pablochacin: I think it will always be required/desirable to signal the subprocess that the binary provisioning is not required to avoid the overhead of reanalyzing the dependencies.

codebien:
>Should the parent wait for the children? to finish. I'm not 100% sure, but I would expect this to be the case. So I'm not sure why the parent finishes first

It should be related to the signal trap enabled by the subprocess where the launcher doesn't have any because it doesn't reach that part of the code.

pablochacin:
> > I'm not 100% sure, but I would expect this to be the case.
> 
> Same here, that's why I asked.

@codebien commented that the cause is  we are not handling the signal in the parent, so it is ending prematurely. 
I will try to add the signal handling to prevent this.

internal/cmd/launcher.go
```diff
@@ -0,0 +1,348 @@
...
+
+// Launcher is a k6 launcher. It analyses the requirements of a k6 execution,
+// then if required, it provisions a binary executor to satisfy the requirements.
+type Launcher struct {
+	// gs is the global state of k6.
+	gs *state.GlobalState
+
+	// provision generates a custom binary from the received list of dependencies
+	// with their constrains, and it returns an executor that satisfies them.
+	provision func(*state.GlobalState, k6deps.Dependencies) (commandExecutor, error)
```
oleiade, line 38: Nit (feel free to ignore/dismiss): although it is not strictly necessary here, I tend to prefer, when possible, defining and implementing an interface (`Provisioner`?), instead of a callback. The reason is ease of debugging and codebase navigation in IDEs. It's much easier to go to definition of an interface and ask for implementations, than to track down where the concrete callback instance is defined and passed down. Sometimes it also makes testing more straightforward.

No arguing on my end, really just a suggestion and expressing a preference, I'm okay with it as-is 👍

oleiade (approved):
LGMT 👍🏻 🚀 

Once the signals handling PR is merged, I think this is good to go. I share some of the observations already made, but considering we already have most of the points raised on the radar documented as issues to be tackled in further iterations 👍🏻 on my end.

Really happy to see this (tentatively) passing the finish line 🏁 I'm grateful for the work you put in @pablochacin 👏🏻

codebien:
@pablochacin additional reason for #4737

Help for subcommands seems an additional issue with the current solution:
```sh
K6_BINARY_PROVISIONING=1 k6 cloud run --help
ERRO[0000] Binary provisioning is enabled but it failed to analyze the dependencies. Please, make sure to report this issue by opening a bug report.  error="script not found"
```

inancgumus (dismissed):
LGTM 👍 Nice work.

pablochacin:
@codebien I implemented a workaround fo the issue with the help commands here https://github.com/grafana/k6/pull/4753 
But agree this approach is not sustainable as we are pre-parsing the CLI instead of relying on k6 command parsing logic.

internal/cmd/launcher.go
codebien: @pablochacin I agree with @oleiade, at least it should be worth to have a reflection/attempt on refactor it. Can you open and add an issue to the stability epic, please?

codebien (approved):

andrewslotin (approved):

oleiade (approved):

PR #4686, opened 2025-04-11
Drop the redundant NO_COLOR detection

codebien:
## What?

It drops a redundant logic about `NO_COLOR`.

## Why?

Because we are already checking here https://github.com/grafana/k6/blob/b61ca160537a021d26695b2b36d2b63259b7016f/cmd/state/state.go#L188-L192

## Checklist


oleiade (approved):
🚀

joanlopez (approved):

PR #4690, opened 2025-04-14
Add fix to reduce memory and cpu consumption

ankur22:
## What?

Using a single `bytes.Buffer` across the lifespan of the connection when reading from the websocket.

## Why?

By reducing how much memory it uses as well as how often it is allocated, we should also reduce the cpu usage since GC will have an easier time cleaning up.

## Checklist

### DIFF
--- a/internal/js/modules/k6/browser/common/connection.go
+++ b/internal/js/modules/k6/browser/common/connection.go
@@ -328,6 +328,7 @@ func (c *Connection) findTargetIDForLog(id target.SessionID) target.ID {
 
 //nolint:funlen,gocognit,cyclop
 func (c *Connection) recvLoop() {
+	decoder := jlexer.Lexer{}
 	var buffer bytes.Buffer
 	c.logger.Debugf("Connection:recvLoop", "wsURL:%q", c.wsURL)
 	for {
@@ -352,7 +353,8 @@ func (c *Connection) recvLoop() {
 		c.logger.Tracef("cdp:recv", "<- %s", buf)
 
 		var msg cdproto.Message
-		c.decoder = jlexer.Lexer{Data: buf}
+		decoder.Data = buf
+		c.decoder = decoder
 		msg.UnmarshalEasyJSON(&c.decoder)
 		if err := c.decoder.Error(); err != nil {
 			select {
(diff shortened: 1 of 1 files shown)


oleiade (approved):
🚀

codebien:
@ankur22 I see this pull request being marked as draft, but there was a review request at the same time? Can you clarify your expectation, please?

In addition, I don't see a pool used here, it sounds you probably want to introduce one, to maximize the effort. Did you consider it?

ankur22:
> @ankur22 I see this pull request being marked as draft, but there was a review request at the same time? Can you clarify your expectation, please?

@codebien and @oleiade sorry for the noise! This is indeed a draft PR... i'm not sure how to open one without it auto assigning reviewers.

> In addition, I don't see a pool used here, it sounds you probably want to introduce one, to maximize the effort. Did you consider it?

Good shout, do you have any material you can point me to so that I can understand the implementation of a pool of memory?

codebien:
>not sure how to open one without it auto assigning reviewers

@ankur22 If you open it directly as a draft, then reviewers are not assigned. They will be when you turn it into a `Ready for review` pull request.

>EDIT: I think you're referring to sync.Pool. Sounds promising, i'll give it a go 👍

Yes, my proposal was around using `sync.Pool` for the buffers to limit the allocation and reduce GC.

internal/js/modules/k6/browser/common/connection.go
```diff
@@ -327,18 +328,33 @@ func (c *Connection) findTargetIDForLog(id target.SessionID) target.ID {
...
 
+		if _, err := buffer.ReadFrom(reader); err != nil {
+			c.handleIOError(err)
+			return
+		}
+
+		// Clone the bytes from the recvBuffer to buf before unmarshaling
+		// the message. Without this, the unmarshaling will fail as it
+		// reuses the buf as the underlying bytes buffer.
+		buf := bytes.Clone(buffer.Bytes())
```
ankur22, line 349: @codebien thanks for the `sync.Pool` suggestion. I think there's some promise in going down that root. However, it's not going to be a straight forward change. Since the browser module works with events, it's not straight forward to keep track of all the events and have them put the buffer back in the pool. We'd need to refactor a lot of the codebase to be able to put the underlying buffer back in the pool, either by keeping track of the data or by allowing the handlers of the event to put the buffer back into the pool 😭

PR #4698, opened 2025-04-16
Summary: Display thresholds values even when not in summaryTrendStats

### LINKED ISSUE #4695: A percentile for not default values is undefined in the report
### Brief summary

Setting thresholds for `http_reuqest_duration` with NOT the default (p(90), p(95)) will produce `undefined` in the output.

```
thresholds: {
	// 99.5% of requests should be below 50ms
	// 99.99% of requests should be below 100ms
	http_req_duration: ["p(99.5)<50", "p(99.99)<100"],
}
```

However, the threshold works completely fine.

### k6 version

k6 v1.0.0-rc1

### OS

macOS 15.2 (24C101)

-

Setting thresholds for `http_reuqest_duration` with NOT the default (p(90), p(95)) will produce `undefined` in the output.

```js
import http from "k6/http";
import { check, sleep } from "k6";

export const options = {
	thresholds: {
		// 99.5% of requests should be below 50ms
		// 99.99% of requests should be below 100ms
		http_req_duration: ["p(99.5)<50", "p(99.99)<100", "p(95)<200"],
	},
	scenarios: {
		ping: {
			executor: "ramping-arrival-rate",
			preAllocatedVUs: 100,
			exec: "ping",
			stages: [
				{ target: 10, duration: "5s" },
				{ target: 5, duration: "2s" },
			],
		},
	},
};

export function ping() {
	// let res = http.get("http://host.docker.internal:8888/ping");
	// let res = http.get("http://localhost:8888/ping");
	const res = http.get("https://test.k6.io");

	check(res, { kek: (r) => r.status === 202 });

	sleep(0.3);
}
```

However, the thresholds work completely fine., the issue lies in output.

Image

### Expected behaviour

The expected behaviour would be printing the actual result for provided percentiles.  

F.e. 
```
✗ 'p(99.5)<50' p(99.5)=107s
✗ 'p(99.99)<100' p(99.99)=3141692ms
```

### Actual behaviour

The results are undefined:

```
  http_req_duration
    ✗ 'p(99.5)<50' p(99.5)=undefined
    ✗ 'p(99.99)<100' p(99.99)=undefined
```

joanlopez:
## What?

Makes it possible to display threshold values (since #4089) even when the aggregation on the thresholds expression (e.g. `p(99.5)`) isn't included as part of `options.summaryTrendStats`.

## Why?

Now those values aren't displayed (instead, it prints `undefined`, which I would consider a bug), because we're not calculating values for aggregations other than  `options.summaryTrendStats`.

Alternatively, we could just omit these lines from the summary, but I think these are valuable information for the user, and it's aligned with the goal of the revamped summary.

## Checklist

## Related PR(s)/Issue(s)

Closes #4695


oleiade (dismissed):
🚀 LGTM on my end 🚀 

(I believe removing the logs 👇🏻 would fix the tests 🙇🏻 )

internal/js/summary.js
```diff
@@ -941,13 +941,19 @@ function renderMetricValueForThresholds(
 	info,
 	formatter,
 ) {
-	const {trendStats, trendCols, nonTrendValues, nonTrendExtras} = info;
+	const {trendStats, trendCols, trendKeys, nonTrendValues, nonTrendExtras} = info;
 	const thresholdAgg = threshold.source.split(/[=><]/)[0].trim();
 
+	console.log(info);
+	console.log(metric.values);
+
```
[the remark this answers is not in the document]
joanlopez: Yeah, sorry, I forgot one extra force-push before opening the PR haha 😅

internal/js/summary.js
```diff
@@ -839,7 +839,7 @@ function renderThresholdResults(
 			: formatter.decorate(failMark, 'red');
 
 		const sourceText = formatter.decorate(
-			`'${threshold.source}'`,
```
joanlopez, line 842: This isn't directly related, but it's so small that I prefer to add it as part of this PR. It's not a huge deal, but I think it's nice if we trim the threshold's source before displaying it in the summary.

internal/output/summary/data.go
```diff
@@ -393,3 +404,28 @@ func metricValueGetter(summaryTrendStats []string) func(metrics.Sink, time.Durat
 		return result
 	}
 }
+
+var percentileThresholdSourceRe = regexp.MustCompile(`^p\((\d+(?:\.\d+)?)\)\s*([<>=])`)
+
+func extractPercentileThresholdSource(source string) (agg string, percentile float64, isPercentile bool) {
```
joanlopez, line 410: I know this isn't ideal, but I cannot see any alternative to this that doesn't imply exposing more things from the `metrics` package and type, and I think that isn't ideal as well because we do plan to do some refactors there soon-ish. 

So, that's why I suggest to go with this, which should be fine as I don't expect the thresholds format to change anytime soon, especially now with v1.0 and the versioning strategy and guarantees.

codebien (changes requested):
Just an observation: the user experience might be tricky here. Because now we might have the potential scenarios where disabled stats appear in the summary.

I know that isn't a new limitation introduced here, instead it should highlight the need to track/untrack specific metrics as described in #1321, where we should extend the same concept to stats.

joanlopez:
> Just an observation: the user experience might be tricky here. Because now we might have the potential scenarios where disabled stats appear in the summary.
> 
> I know that isn't a new limitation introduced here, instead it should highlight the need to track/untrack specific metrics as described in #1321, where we should extend the same concept to stats as well.

I'm sorry @codebien, but I don't fully get your point here. As far as I can see, #1321  is all about metrics and sub-metrics, but here we only care about different "calculations" (percentiles, basically") of already existing metrics/sub-metrics, so I cannot see how it related 🤔  Conceptually I see those two as different, unrelated things.

internal/js/summary.js
joanlopez: > If the source includes spaces why we don't interpret it as intentional?

I'm a bit confused now; is there any reason for doing so? 

To me, it feels like when you mistakenly add a leading or trailing whitespace in an input designed to write your name or similar.

Looking at the threshold syntax expression ([ref](https://grafana.com/docs/k6/latest/using-k6/thresholds/#threshold-expression-syntax)), I can see for instance users adding (or not) whitespaces before/after the `<operator>`, for readability, but not why doing so at the beginning/end of the threshold definition.

internal/js/summary.js
```diff
@@ -1061,6 +1065,10 @@ function computeSummaryInfo(metrics, renderContext, options) {
 	const nonTrendExtras = {};
 	const trendCols = {};
 
+	// While "trendCols" contain the values for each "trendStats" aggregation (e.g. p(90) as a sorted array,
+	// "trendKeys" is used to store specific aggregation values that aren't part of "trendStats"; mainly for thresholds.
+	const trendKeys = {};
```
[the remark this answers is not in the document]
joanlopez: I'm not sure 🤷🏻 The code/data structure is quite generic, and while it's true that this is now used for threshold, it could technically be used for anything else, so I tried to keep it consistent with the existing "conventions" (if any) - one data structure holds stats identified by "columns", and this one by "keys" 😅 

Overall, that code has many aspects that could be improved (there's already some `// TODOs`), and @oleiade already did a great job fixing some as part of #4089, but I'd prefer if we can keep focus on fixing the issue that's been reported, and re-evaluate namings and possible improvements over this code in another PR/ as part of another issue/work.

codebien:
@joanlopez, my interpretation of the issue is that the user wants to add a threshold for a specific metric at a particular percentile without adding that percentile to all the other metrics.

If that's not the case, then I guess we could simply suggest using `summaryTrendStats: [/* other stats */..., 'p(99.99)']` or on our side, skip visualizing the non-computed stats to avoid `undefined` values.

However, neither of these solutions addresses the user's specific requirement.

Instead, I think 1321 might address it, if we integrate the proposed API with an option for defining the stats. Something similar to the example below:
```js
export const options = { 
  summaryTrendStats: ['max', 'p(95)'], 
};

track("http_req_duration", {percentile: 99.9})
```
Where without the `track` line, the threshold will never fail because the stats for the metric doesn't exist. Adding this feature then will lead to a breaking change, but that should be acceptable since I don't expect this being real before `v2`.

joanlopez:
> @joanlopez, my interpretation of the issue is that the user wants to add a threshold for a specific metric at a particular percentile without adding that percentile to all the other metrics.

Let me slightly correct you, because I think it's key here:

"_...at a particular percentile without_ **displaying** _that percentile_ **in summary** _to all the_ ~other~ _metrics_."

What I mean here is basically two things (in my humble opinion):
1. The `summaryTrendStats` option is, as its name hints, an option to configure the "end-of-test summary", that let users choose what's displayed there (which percentiles values are displayed in the summary for trend metrics). NOT an option to configure what metrics are tracked/untracked.
2. Percentiles aren't a form of "(sub)metrics tracked", but just a calculation you do over a metric, or more correctly over a sink. It's like when we refer to "min" or "max" for a given "gauge sink". What I think could make sense in the future, as you mentioned, is a mechanism to let the user choose what (sub)metrics are tracked, but less likely what calculations.

> If that's not the case, then I guess we could simply suggest using `summaryTrendStats: [/* other stats */..., 'p(99.99)']` or on our side, skip visualizing the non-computed stats to avoid `undefined` values.

I disagree with both, because:
- I still see value of having the opportunity to define a threshold for a concrete percentile, even if you don't want to see the value of that percentile for ALL the metrics displayed in the end-of-test summary. In other words, why do I need to add to the summary the value of `p(99.99)` for all the trend metrics, if I only care about the `p(99.99)` of this particular metric X?
- I think it's better to display the value in the thresholds section that hiding it because:
  - It keeps that behavior consistent with the whole "thresholds" section, where the value of the aggregation the threshold is defined on is always displayed. It also keeps it consistent with the rest of the summary,as the value won't be displayed in the list below, where all the metrics are displayed, according to `summaryTrendStats`.
  - Accompanying thresholds results with the value of the particular aggregator used by the threshold was precisely one of the main ideas of the revamped end-of-test summary, because it really makes it very easy for the user to understand why that threshold succeed/failed.
 
> However, neither of these solutions addresses the user's specific requirement.
> 
> Instead, I think 1321 might address it, if we integrate the proposed API with an option for defining the stats. Something similar to the example below:
> 
> ```js
> export const options = { 
>   summaryTrendStats: ['max', 'p(95)'], 
> };
> 
> track("http_req_duration", {percentile: 99.9})
> ```
> 
> Where without the `track` line, the threshold will never fail because the stats for the metric doesn't exist. Adding this feature then will lead to a breaking change, but that should be acceptable since I don't expect this being real before `v2`.

That's definitely out of the scope of this PR, so I'd postpone that discussion for whenever we design that feature, but I'm not super convinced that having to specify what calculations do we want to track is really a good idea, because calculations aren't that associated with noise/high memory consumption, which I think are the underneath goal of that feature, and doing so would make both the design and implementation of such feature way more complex.

cc/ @oleiade any take here?

codebien (approved):
I want to unblock the merge for the pull request, so it mostly looks good to me. I want to check what is the context to provide an answer for https://github.com/grafana/k6/pull/4698#discussion_r2054336804, but I don't have time today, I will do it at the end of the week. :bow:

codebien:
@joanlopez However, the CI seems to be red, can you check that is not related, please?

joanlopez:
> @joanlopez However, the CI seems to be red, can you check that is not related, please?

It isn't, it's just one of the flaky Browser E2E tests that define a thresholds that isn't always satisfied.

oleiade (approved):
🚀 🍾

PR #4708, opened 2025-04-18
[Binary provisioning] Rename runners

codebien:
## What?

It refactors by renaming a few components.

## Why?

The current naming doesn't sound consistent with the name commonly used. It tries to apply the principle of ubiquitous language.  

## Checklist

### DIFF
--- a/internal/launcher/launcher.go
+++ b/internal/launcher/launcher.go
@@ -8,24 +8,33 @@ import (
 	"go.k6.io/k6/internal/build"
 )
 
-// Launcher is a k6 Launcher
-// It analyses the requirements of a k6 execution. If required, provisions a k6Runner to satisfy these
-// requirements.
+// binaryRunner executes the requested k6 command line command.
+// It abstract the execution path from the concrete binary.
+type binaryRunner interface {
+	run(*state.GlobalState)
+}
+
+// Launcher is a k6 launcher. It analyses the requirements of a k6 execution,
+// then if required, it provisions a binary executor to satisfy the requirements.
 type Launcher struct {
+	// gs is the global state of k6.
 	gs *state.GlobalState
-	// provision function receives a list of dependencies with their constrains and returns
-	// a k6runner than satisfies them
-	provision func(*state.GlobalState, k6deps.Dependencies) (k6Runner, error)
-	// k6Runner to execute k6 command
-	runner k6Runner
+
+	// provision generates a custom binary from the received list of dependencies
+	// with their constrains, and it returns an executor that satisfies them.
+	provision func(*state.GlobalState, k6deps.Dependencies) (binaryRunner, error)
+
+	// binaryRunner executes the requested k6 command line command
+	binaryRunner binaryRunner
 }
 
 // New creates a new Launcher from a GlobalState using the default fallback and provision functions
 func New(gs *state.GlobalState) *Launcher {
+	defaultRunner := &currentBinary{}
 	return &Launcher{
-		gs:        gs,
-		provision: k6buildProvision,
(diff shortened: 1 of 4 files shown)


internal/launcher/launcher.go
```diff
@@ -8,35 +8,44 @@ import (
...
-	provision func(*state.GlobalState, k6deps.Dependencies) (k6Runner, error)
-	// k6Runner to execute k6 command
-	runner k6Runner
+
+	// provision generates a custom binary from the received list of dependencies
+	// with their constrains, and it returns an executor that satisfies them.
+	provision func(*state.GlobalState, k6deps.Dependencies) (binaryRunner, error)
+
+	// binaryRunner executes the requested k6 command line command
+	binaryRunner binaryRunner
```
pablochacin, line 28: I'm not sure about this change. I prefer k6Runner as it reflects what it does and not how. Only the binary provisioning runner is using an external binary.

joanlopez (commented):
Although I'm not 100% sure whether "launcher" is a good name here, I'm not super convinced that naming it "runner" is a good idea, because "runner" is a core piece in k6, and despite the "binary" prefix (which is btw being discussed) it might cause some conceptual confusion.

Why not something along the lines of "execute" or "executor", which I think it's more common when speaking about "running processes", which is what I think the code does here. For instance, the corresponding Go package is named "exec".

PS: Of course, this is only my humble take on this topic, because I'm not a huge fan of renaming something if the new name isn't clearly better than the former. But overall, this shouldn't be considered a blocker for merging the binary provisioning feature at all, from my side.

internal/launcher/launcher.go
codebien: I don't like the fact we have `k6` as a prefix because it doesn't add any value. It isn't clear what `k6` means in this context.

But I guess I've found one term that should address your request about expressing what it does.

Question: What it does?
Answer: It executes a sub-command. 

This is the reason why I propose `subcommandExecutor`. It integrates @joanlopez's feedback as well.

```suggestion
	subCmdExec subcommandExecutor
```

codebien: I'm fine with `commandExecutor`. Going to submit the changes.

pablochacin (approved):
LGTM. Minor nitpick.

codebien:
>Minor nitpick.

@pablochacin I've merged this pull request as I don't see any additional comment. But let me know if you have any additional request and I will address them in a follow-up pull request.

internal/launcher/launcher.go
```diff
@@ -8,35 +8,44 @@ import (
...
+	// with their constrains, and it returns an executor that satisfies them.
+	provision func(*state.GlobalState, k6deps.Dependencies) (commandExecutor, error)
+
+	// commandExecutor executes the requested k6 command line command
+	commandExecutor commandExecutor
 }
 
 // New creates a new Launcher from a GlobalState using the default fallback and provision functions
 func New(gs *state.GlobalState) *Launcher {
+	defaultRunner := &currentBinary{}
```
joanlopez, line 33: I'm not sure what nit Pablo referred to @codebien, but I think this variable name is likely the only one missing the renaming. I'd advocate for "defaultExecutor" or "defaultBinary", or something along those lines as we no longer use "runner".

Btw, thanks for this renaming, now I like how it ended up, not only around the "executor" vs "runner", but also the other part of the rename, like "customBinary", etc. 🙇🏻

pablochacin: @joanlopez was the same you pointed out. It's Strange, I was sure I made the comment in the code.

Issue #4687, opened 2025-04-11
Injected script: Avoid depending on globals


inancgumus:
# What

The injected script should not depend on global JavaScript objects.

# Why

Websites under test can overwrite these objects, leading to unexpected behavior.

# Example

One website can overwrite `Set` with a simple function like this:

```javascript
function Set() {
    ...
}
```

The injected script would then use the function above while it was expecting a native [Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set) object, leading to unexpected issues.

inancgumus:
Fixed by #4709.

PR #4709, opened 2025-04-18
browser: Fix JS native objects override

inancgumus:
## What?

Avoids a page under the test to overwrite native JavaScript objects, like `Set` and `Map`, by injecting a temporary and invisible frame in execution contexts. We can extend this to support other native objects in the future. However, for now, this solves crucial issues for users.

Note: I will update the release notes after I get the approvals, but before the merge.

## Why?

See #4687 for details.

## Checklist

## Related PR(s)/Issue(s)

Updates: #4687

### DIFF
--- a/internal/js/modules/k6/browser/common/js/injected_script.js
+++ b/internal/js/modules/k6/browser/common/js/injected_script.js
@@ -14,7 +14,32 @@
  * limitations under the License.
  */
 
-const autoClosingTags = new Set([
+// k6BrowserNative allows accessing native browser objects
+// even if the page under test has overridden them.
+const k6BrowserNative = (() => {
+  const iframe = document.createElement('iframe');
+  // hide it offscreen with zero size
+  iframe.style.position = 'absolute';
+  iframe.style.width = '0';
+  iframe.style.height = '0';
+  iframe.style.border = '0';
+  iframe.style.top = '-9999px';
+  iframe.style.left = '-9999px';
+  iframe.style.display = 'none';
+
+  // grab the native browser object
+  document.documentElement.appendChild(iframe);
+  const win = iframe.contentWindow;
+  document.documentElement.removeChild(iframe);
+
+  return {
+    Set: win.Set,
+    Map: win.Map,
+    // Add other native browser objects as needed.
+  }
+})();
+
+const autoClosingTags = new k6BrowserNative.Set([
   "AREA",
   "BASE",
   "BR",
@@ -33,14 +58,14 @@ const autoClosingTags = new Set([
   "TRACK",
   "WBR",
 ]);
-const booleanAttributes = new Set([
+const booleanAttributes = new k6BrowserNative.Set([
   "checked",
   "selected",
   "disabled",
   "readonly",
   "multiple",
 ]);
-const eventType = new Map([
+const eventType = new k6BrowserNative.Map([
   ["auxclick", "mouse"],
   ["click", "mouse"],
   ["dblclick", "mouse"],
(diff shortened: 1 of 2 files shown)


internal/js/modules/k6/browser/common/js/injected_script.js
```diff
@@ -14,7 +14,22 @@
...
 
-const autoClosingTags = new Set([
+// k6BrowserNative allows accessing native browser objects
+// even if the page under test has overridden them.
+const k6BrowserNative = (() => {
+  const iframe = document.createElement('iframe');
+  iframe.style.display = 'none';
+  document.documentElement.appendChild(iframe);
+  const win = iframe.contentWindow;
+  document.documentElement.removeChild(iframe);
```
mstoykov, line 24: I am hopeful that adding and removing an iframe before it has any chance of actually redrawing is enough for this to not have side effects. 

But it seems kind of risky. Unfortunately my quick search hasn't shown another way to do this, so :shrug:

I guess another more involved option will be to not depend on map or set ... but to be honest someone breaking them seems pretty ... strange and likely a  much bigger problem for them than this script not working.

inancgumus: Yep. I also am not a fan of this solution (despite the fact that other JavaScript testing engines follow a similar approach). However, whatever we do, users can override them even if we use our potentially error-prone shims instead of native objects. So, I prefer fixing the immediate issue with an invisible iframe that never paints. I'm open to other suggestions, but the user needs this fix as soon as possible. Since we have a test, we can refactor it later if needed.

inancgumus: > I am hopeful that adding and removing an iframe before it has any chance of actually redrawing is enough for this to not have side effects. But it seems kind of risky. Unfortunately my quick search hasn't shown another way to do this, so 🤷

This user can test their webpage on other test engines without issues. This never-painting iframe method resolves the user's issue, and all our integration tests pass. Other test engines use a similar approach.

> I guess another more involved option will be to not depend on map or set ... but to be honest someone breaking them seems pretty ... strange and likely a much bigger problem for them than this script not working.

I tried that approach before concluding on this one. It can quickly become error-prone and complicated, since we already have a working (injected) script. Besides, users can also overwrite other native objects (e.g., not just `Set`), making it work for their webpages. We'll eventually end up using nothing native in our script :) Just because users shim native objects that shouldn't break k6. Still, I'm open to other solutions.

codebien (approved):
LGTM, but why only Set and Map? Can we open an issue for all the other missing functions, please? :bow:

codebien:
@inancgumus However, the CI seems to fail on `TestPageMustUseNativeJavaScriptObjects`. Is that expected?

inancgumus:
@codebien 

> The CI seems to fail on TestPageMustUseNativeJavaScriptObjects. Is that expected?

No, it's definitely not. I'm trying to fix the flakiness on CI. It runs fine locally.

> LGTM, but why only Set and Map? Can we open an issue for all the other missing functions, please? 🙇

Listing all the other functions would be cumbersome in an issue, as we'll need to update that single issue every time we use a native object. However, I can keep [the main issue](https://github.com/grafana/k6/issues/4687) open instead of closing it out. We used `Set` and `Map` here to fix the user's immediate issue.

inancgumus:
@codebien I've simplified the test and also eliminated the flakiness.

codebien (approved):

mstoykov (approved):

PR #4721, opened 2025-04-22
Fix argument scanning

pablochacin:
## What?

Fixes the logic for scanning command arguments to identify if the command requires an input script and if the script was provided.

## Why?

Separates the two concerns.

## Checklist


internal/launcher/analyze.go
```diff
@@ -52,35 +56,46 @@ func analyze(gs *state.GlobalState, args []string) (k6deps.Dependencies, error)
 	return k6deps.Analyze(dopts)
 }
 
-// scriptNameFromArgs returns the file name passed as input and true if it's a valid script name
-func scriptNameFromArgs(args []string) (string, bool) {
+// search for the command and returns a boolean indicating it takes a script or not
+func requiresScript(args []string) bool {
```
codebien, line 60:
```suggestion
// isScriptRequired search for the command and returns a boolean indicating if it is required to pass a script or not
func isScriptRequired(args []string) bool {
```

internal/launcher/analyze.go
```diff
@@ -52,35 +56,46 @@ func analyze(gs *state.GlobalState, args []string) (k6deps.Dependencies, error)
...
 	// return early if no arguments passed
 	if len(args) == 0 {
-		return "", false
+		return false
 	}
```
codebien, line 64: It confuses me, what does it happen if it is required, but the user didn't pass any for a mistake?

pablochacin: Yes. I don't think we should make this validation here.

internal/launcher/launcher_test.go
```diff
@@ -186,6 +186,19 @@ func TestLauncherLaunch(t *testing.T) {
 			expectK6Run:     true,
 			expectOsExit:    108,
 		},
+		{
+			name:   "missing input script",
+			k6Cmd:  "run",
+			k6Args: []string{},
+			k6Env: map[string]string{
+				"K6_BINARY_PROVISIONING": "true",
+			},
```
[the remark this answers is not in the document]
pablochacin: I removed the environment variables from the test setup. I think that adding a test for a single test case introduces an unnecessary duplication of the whole test logic. 

With the current logic I could add an option to the test case for enabling/disabling the binary provisioning. WDYT?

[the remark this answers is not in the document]
codebien: Do you mean by having something like this:

 ```go
bpEnabled := range []bool{true , false} {
t.Run(fmt.Spritnf("BinaryPro=%v", bpEnabled), func(t *testing.T
    for _, tc := range testCases {
        t.Run(tc.name, func(t *testing.T
```

I'm fine with it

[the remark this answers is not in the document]
Comment (pablochacin) [reply internal/launcher/launcher_test.go]: I don't think it worth testing every option with/without binary enabled. 
 
 What I had in mind was adding a attribute to the test case: 
 ```
testCases := []struct {
		name            string
		enabled        bool
```

and use it when setting the test
```
ts.GlobalState.Flags.BinaryProvisioning = tc.enabled
```

We have to repeat this attribute for each test case, but I think that's better than duplicating the test logic.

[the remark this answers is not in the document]
codebien: I'm fine with it. But do the opposite condition with `disableBinaryProvisioning` so we have to write it only for one case.

internal/launcher/analyze_test.go
```diff
@@ -6,138 +6,186 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-func TestScriptArg(t *testing.T) {
+func TestIsScriptRequired(t *testing.T) {
```
joanlopez, line 9: I think this should be pointing to "scriptNameFromArgs", as it's the method tested by this test suite.

pablochacin: Fixed

codebien (approved):

PR #4731, opened 2025-04-23
skip binary provisioning for run command

pablochacin:
## What?

skip binary provisioning for `k6 run` command

## Why?

Binary provisioning is currently supported only for users who have a cloud account, because it requires authenticating to GCK6.
Enabling binary provisioning only for cloud command (e.g. `k6 cloud run`) offers a more consistent user experience. The exception to this are the `k6 archive` and `k6 inspect` commands that are require for the users to upload tests to the cloud.

## Checklist

### DIFF
--- a/internal/cmd/launcher.go
+++ b/internal/cmd/launcher.go
@@ -251,7 +251,7 @@ func analyze(gs *state.GlobalState, args []string) (k6deps.Dependencies, error) 
 
 	if !isScriptRequired(args) {
 		gs.Logger.
-			Debug("command does not require Binary provisioning")
+			Debug("The command to execute does not require Binary provisioning")
 		return k6deps.Dependencies{}, nil
 	}
 
--- a/internal/cmd/launcher.go
+++ b/internal/cmd/launcher.go
@@ -300,7 +300,7 @@ func isScriptRequired(args []string) bool {
 				}
 			}
 			return true
-		case "run", "archive", "inspect":
+		case "archive", "inspect":
 			return true
 		}
 	}
--- a/internal/cmd/launcher.go
+++ b/internal/cmd/launcher.go
@@ -56,18 +56,19 @@ func NewLauncher(gs *state.GlobalState) *Launcher {
 func (l *Launcher) Launch() {
 	// If binary provisioning is not enabled, continue with the regular k6 execution path
 	if !l.gs.Flags.BinaryProvisioning {
-		l.gs.Logger.Debug("Binary provisioning feature is disabled")
+		l.gs.Logger.Debug("Binary provisioning feature is disabled.")
 		l.commandExecutor.run(l.gs)
 		return
 	}
 
-	l.gs.Logger.Info("Binary provisioning feature is enabled. If it's required, k6 will provision a new binary")
+	l.gs.Logger.
+		Debug("Binary provisioning feature is enabled.")
 
 	deps, err := analyze(l.gs, l.gs.CmdArgs[1:])
 	if err != nil {
 		l.gs.Logger.
 			WithError(err).
-			Error("Failed to analyze the required dependencies. Please, make sure to report this issue by" +
+			Error("Binary provisioning enabled but failed to analyze dependencies. Please, make sure to report this issue by" +
 				" opening a bug report.")
 		l.gs.OSExit(1)
 		return // this is required for testing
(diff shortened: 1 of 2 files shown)


codebien (changes requested):
```
Binary provisioning feature is enabled. If one of the commands between `cloud run`, `cloud run --local-exeuction`, `archive` or `inspect` is used then k6 will try to provision a new k6 binary.
```
I think we need a better UX, because if someone now tries `k6 run` it won't be able to get why it doesn't work as expected.

internal/cmd/launcher.go
```diff
@@ -56,18 +56,19 @@ func NewLauncher(gs *state.GlobalState) *Launcher {
...
-	l.gs.Logger.Info("Binary provisioning feature is enabled. If it's required, k6 will provision a new binary")
+	l.gs.Logger.
+		Debug("Binary provisioning feature is enabled.")
 
 	deps, err := analyze(l.gs, l.gs.CmdArgs[1:])
 	if err != nil {
 		l.gs.Logger.
 			WithError(err).
-			Error("Failed to analyze the required dependencies. Please, make sure to report this issue by" +
+			Error("Binary provisioning enabled but failed to analyze dependencies. Please, make sure to report this issue by" +
```
codebien, line 71:
```suggestion
			Error("Binary provisioning is enabled but it failed to analyze the dependencies. Please, make sure to report this issue by" +
```

internal/cmd/launcher.go
```diff
@@ -76,22 +77,21 @@ func (l *Launcher) Launch() {
 	// if the command does not have dependencies or a custom build
 	if !customBuildRequired(build.Version, deps) {
 		l.gs.Logger.
-			Debug("The current k6 binary already satisfies all the required dependencies," +
-				" it isn't required to provision a new binary.")
+			Debug("Binary provisioning is not required to execute this command.")
```
codebien, line 80: I find the previous message better because it explains _the why_. Can we restore it?

internal/cmd/launcher.go
```diff
@@ -76,22 +77,21 @@ func (l *Launcher) Launch() {
...
 		l.commandExecutor.run(l.gs)
 		return
 	}
 
 	l.gs.Logger.
 		WithField("deps", deps).
-		Info("The current k6 binary doesn't satisfy all the required dependencies, it is required to" +
-			" provision a new binary.")
+		Info("The current k6 binary doesn't satisfy all dependencies, it is required to" +
+			" provision a custom binary.")
```
codebien, line 88:
```suggestion
			" provision ad hoc binary.")
```
`custom` sounds too generic term

internal/cmd/launcher.go
```diff
@@ -194,7 +194,7 @@ func k6buildProvision(gs *state.GlobalState, deps k6deps.Dependencies) (commandE
 	}
 
 	if config.BuildServiceAuth == "" {
-		return nil, errors.New("k6 cloud token is required when Binary provisioning feature is enabled." +
+		return nil, errors.New("k6 cloud token is required when the binary provisioning feature is enabled." +
```
codebien, line 197:
```suggestion
		return nil, errors.New("k6 cloud token is required when the binary provisioning feature is enabled." +
```
I'm using Binary provisioning because it is the _name_ of the feature. So, any name requires to be capitalized.

internal/cmd/launcher.go
```diff
@@ -249,6 +249,8 @@ func analyze(gs *state.GlobalState, args []string) (k6deps.Dependencies, error)
 	}
 
 	if !isScriptRequired(args) {
+		gs.Logger.
+			Debug("command does not require binary provisioning")
```
codebien, line 253:
```suggestion
			Debug("The command to execute does not require binary provisioning")
```
A follow-up improvement might be to return the identified command from `isScriptRequired` function and log it here. However, it can be a low priority improvement for the future.

internal/cmd/launcher.go
pablochacin: I think this is the term we use in the documentation referring to custom k6 build.

internal/cmd/launcher.go
pablochacin: I think it should be "Binary Provisioning"

codebien: It is fine to me, go for it :+1:

codebien (approved):

Issue #4724, opened 2025-04-22
--summary-mode=full outputs results by group name in alphabetical order


wlahti-aderant:
Using version `0.58.0` with k6chaijs for my grouping/checks, the output from running my tests with `k6 run --summary-mode=full` returns in alphabetical order by GROUP and nested GROUP names. I'd expect the output to display in the order the code is written in tests and subsequently executed. 

The current full output is very confusing to look at considering things like my `teardown` function shows under a GROUP and isn't the final group in the output. 

```
  █ GROUP: Groups

  █ GROUP: teardown 

  █ GROUP: Users
```

Similarly, I have sub test cases chained together (and nested under top-level `describe` blocks): 
```describe
    describe('Users', () => {
        describe('CRUD APIs', () => {
            describe('Initial State', () => {
                // check for empty list of users
            }) &&
                describe('Create first user successfully', () => {
```

that result in output (details redacted to just show the GROUPs):
```
 █ GROUP: Users 

    ↳ GROUP: CRUD APIs 

    ↳ GROUP: Create a user with a duplicate email 

    ↳ GROUP: Create first user successfully 

    ↳ GROUP: Create the second user 

    ↳ GROUP: Delete the users successfully 

    ↳ GROUP: Initial State 

    ↳ GROUP: Retrieve the created user 

    ↳ GROUP: Update a user with an outdated resourceVersion 

    ↳ GROUP: Update the user successfully 
```

joanlopez:
Hi @wlahti-aderant,

Thanks for reporting this, I really appreciate you for taking the time for reporting your observations on this ~new~ revamped feature. From there, I see two possible actions:

- Filtering out `setup` and `teardown` from group results:
  - Which we will do for now, see https://github.com/grafana/k6/pull/4734, to keep this behavior consistent with the legacy summary.
  - But, arguably, this could be useful information for some users, thus a reason to perhaps hide in the default summary but to display it in the full summary. So, this might change in the future.
- Preserving the order of groups from code to the summary:
  - Which I think is a feature request, that we might consider for https://github.com/grafana/k6/milestone/49
  - For now, if this is critical, I'd suggest you use alphanumeric sortable identifiers, like:  "1. do X", "2. do Y", etc.

I hope that helps! 🙇🏻

joanlopez:
[...]

PR #4734, opened 2025-04-24
Summary: Filter out samples tagged with group: setup/teardown

joanlopez:
## What?

It filters out samples tagged with `group: ::setup` or `group: ::teardown` so they aren't collected as "group samples".

## Why?

Since #4089, we modified not only how the data in the summary is displayed but also how it is collected. So, since then, we started to collect samples with `group: ::setup` or `group: ::teardown`, while previously we did not because there are no groups named like `::setup` or `::teardown`, we only "artificially" tag samples with the `group` label (more details [here](https://grafana.com/docs/k6/latest/using-k6/workaround-iteration-duration/)).

So, for now, and until we decide the opposite (if any), we want to preserve the old behavior, where none of these two are displayed in the summary as "real" groups.

## Checklist

## Related PR(s)/Issue(s)

Contribute to #4724

### DIFF
--- a/internal/output/summary/summary.go
+++ b/internal/output/summary/summary.go
@@ -5,6 +5,7 @@ import (
 	"sync/atomic"
 	"time"
 
+	"go.k6.io/k6/internal/lib/consts"
 	"go.k6.io/k6/internal/lib/summary"
 	"go.k6.io/k6/lib"
 	"go.k6.io/k6/metrics"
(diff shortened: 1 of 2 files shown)


codebien (commented):
LGTM. However, we might see CPU spikes for tests with high usage of groups. So, it could require to rethink a bit the end-to-end flow for this specific case. Something to follow-up in the future.

joanlopez:
> LGTM. However, we might see CPU spikes for tests with high usage of groups. So, it could require to rethink a bit the end-to-end flow for this specific case. Something to follow-up in the future.

My (perhaps wrong) feeling tells me that it's improbable that this comparison will ever have a significant impact (in the form of CPU spikes), because I think the operations around the new code (e.g. the `strings.Split` right after) are significantly slower. But yeah, if that ever happens, we can explore ways to optimize this. For instance, we could "cache" the result of certain operations.

EDIT: In any case, I replaced the `strings.Contains` with straight equality comparison against constant values, which should be a bit faster, if the previous code wasn't already optimized by the compiler, which I would not be surprised by, if so.

oleiade (approved):
🚀

Issue #4735, opened 2025-04-24
Filter errors when execution of the sub-process fails


pablochacin:
When the provisioned binary is executed, any failure in execution is reported as an error. 

```
ERRO[0001] Failed to run the provisioned k6 binary       error="exit status 255"
```

This is done to detect any possible error derived from the binary provisioning.

However, in cases where the sub-process fails for valid reasons, this double error may be confusing. For example:

```
ERRO[0001] could not create the cloud test run: (403/E5) User does not have permission to perform this action in project 3542013.
ERRO[0001] Failed to run the provisioned k6 binary       error="exit status 255"
```

It may be convenient to report only unexpected errors.

pablochacin:
Fixed by #4740

PR #4740, opened 2025-04-24
handle signals

pablochacin:
## What?

Handle signals in the parent process when calling a subprocess.

## Why?

Prevent the parent from ending prematurely before the subprocess ends.

Also fixes https://github.com/grafana/k6/issues/4735. By separating the `Start` of the process from waiting for the child to end, we can handle separately errors like binary not found (and report them) from the non-zero return codes from the child.   

## Checklist

### DIFF
--- a/internal/cmd/launcher.go
+++ b/internal/cmd/launcher.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"os/exec"
 	"strings"
+	"syscall"
 
 	"github.com/Masterminds/semver/v3"
 	"github.com/grafana/k6deps"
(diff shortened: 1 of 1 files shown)


oleiade (approved):
**TL;DR**: Confirming it works as expected, even on Windows. I feel comfortable merging as-is in the spirit of getting v1 in time.

Can confirm this works as expected, even on Windows. I think we could test it more thoroughly, especially on Windows, and I agree with the `TODO` comment that we should probably have a timeout forcing the process group to exit in case it takes too long or never does (considering OOMs and 100% CPU tests are still a thing, it might be a potential issue with signals never arriving).

I'll open an issue to track those topics, and make sure it's on the map for the next release.

internal/cmd/launcher.go
```diff
@@ -132,21 +133,42 @@ func (b *customBinary) run(gs *state.GlobalState) {
...
+				rc = 1
+				var eerr *exec.ExitError
+				if errors.As(err, &eerr) {
+					rc = eerr.ExitCode()
+				}
+			}
+			gs.OSExit(rc)
+		case <-sigC:
+			// TODO: maybe we should set a timeout while waiting for the subprocess
+			gs.Logger.Debug("Signal received. Waiting for subprocess to end.")
```
codebien, line 169:
```suggestion
		case sig := <-sigC:
			gs.Logger.WithField("signal", sig.String()).Debug("Signal received, waiting for the subprocess to handle it and return.")
```

codebien (approved):
LGTM :tada:

PR #4741, opened 2025-04-24
[Binary Provisioning] Change the default cache dir

codebien:
## What?

It changes the default folder to `$HOME/.cache/k6/builds/*`.

## Why?

The previous `k6provider` might generate confusion for users, as it doesn't match any installed application.

## Checklist

- [x] I have performed a self-review of my code.
- [x] I have commented on my code, particularly in hard-to-understand areas.
- [x] I have added tests for my changes.
- [x] I have run linter and tests locally (`make check`) and all pass.

## Checklist: Documentation (only for k6 maintainers and if relevant)

**Please do not merge this PR until the following items are filled out.**

- [ ] I have added the correct milestone and labels to the PR.
- [ ] I have updated the release notes: _link_
- [ ] I have updated or added an issue to the [k6-documentation](https://github.com/grafana/k6-docs): grafana/k6-docs#NUMBER if applicable
- [ ] I have updated or added an issue to the [TypeScript definitions](https://github.com/grafana/k6-DefinitelyTyped/tree/master/types/k6): grafana/k6-DefinitelyTyped#NUMBER if applicable

PR #4743, opened 2025-04-24
Binary provisioning/add cache dir flag

pablochacin:
## What?

Adds a flag for configuring the directory used to cache provisioned binaries.

## Why?

The default value may result confusing as it makes reference to the `k6provider` library. 

Fixes #4741 

## Checklist

### DIFF
--- a/cmd/state/state.go
+++ b/cmd/state/state.go
@@ -227,9 +227,6 @@ func getFlags(defaultFlags GlobalFlags, env map[string]string, args []string) Gl
 	if val, ok := env["K6_BUILD_SERVICE_URL"]; ok {
 		result.BuildServiceURL = val
 	}
-	if val, ok := env["K6_BINARY_CACHE"]; ok {
-		result.BinaryCache = val
-	}
 
 	// check if verbose flag is set
 	if slices.Contains(args, "-v") || slices.Contains(args, "--verbose") {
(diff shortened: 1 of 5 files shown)


cmd/state/state.go
```diff
@@ -219,6 +227,9 @@ func getFlags(defaultFlags GlobalFlags, env map[string]string, args []string) Gl
 	if val, ok := env["K6_BUILD_SERVICE_URL"]; ok {
 		result.BuildServiceURL = val
 	}
+	if val, ok := env["K6_BINARY_CACHE"]; ok {
+		result.BinaryCache = val
+	}
```
codebien, line 232:
```suggestion
```
I don't think we have to introduce this option. Go already provides a way to override the OS cache dir, if they really need it. However, even if it would be required, I would prefer if we wait for a real feature-request instead of introducing it now.

codebien (commented):
LGTM, just one comment

cmd/state/state.go
pablochacin: Elaborating on my previous comment, the cache dir we are discussing is for storing the custom binaries downloaded from the build service, there's no go involved (users don't need to have it). Moreover, if in the future we support local builds, these two caches will be different

codebien (approved):

PR #4784, opened 2025-05-12
Golangci lint version detection

tbourrely:
## What?

Update golangci-lint version check & delete removed flag.

## Why?

Following [contributing](https://github.com/grafana/k6/blob/master/CONTRIBUTING.md) guide i ran into two issues.

First, `golangci-lint version` not properly detected as correct:

```
 ~/p/k6  master  make check-linter-version
Your installation of golangci-lint is different from the one that is specified in k6's linter config (there it's v2.1.6). Results could be different in the CI.
 ~/p/k6  master  golangci-lint version
golangci-lint has version 2.1.6 built with go1.24.2 from eabc2638 on 2025-05-04T15:41:19Z
```

Second, `golangci-lint run ...` fails because of an unrecognized flag : 

```
 ~/p/k6  master  make lint
Your installation of golangci-lint is different from the one that is specified in k6's linter config (there it's v2.1.6). Results could be different in the CI.
Running linters...
Error: unknown flag: --out-format
Failed executing command with error: unknown flag: --out-format
make: *** [Makefile:25: lint] Error 3
```

If I'm not mistaken, the argument `--out-format` has been removed in golangci-lint@v2 : https://golangci-lint.run/product/migration-guide/#command-line-flags.

## Checklist

### DIFF
--- a/.github/actions/lint/action.yml
+++ b/.github/actions/lint/action.yml
@@ -24,7 +24,7 @@ runs:
         echo "Downloading '${rules_url}' ..."
         curl --silent --show-error --fail --no-location "${rules_url}" --output "${{ github.action_path }}/.golangci.yml"
 
-        echo "GolangCIVersion=$(head -n 1 "${{ github.action_path }}/.golangci.yml" | tr -d '# ')" >> "${GITHUB_OUTPUT}"
+        echo "GolangCIVersion=v$(head -n 1 "${{ github.action_path }}/.golangci.yml" | tr -d '# ')" >> "${GITHUB_OUTPUT}"
       id: getenv
     - name: golangci-lint
       uses: golangci/golangci-lint-action@4afd733a84b1f43292c63897423277bb7f4313a9 # v8.0.0
--- a/.github/actions/lint/action.yml
+++ b/.github/actions/lint/action.yml
@@ -24,7 +24,7 @@ runs:
         echo "Downloading '${rules_url}' ..."
         curl --silent --show-error --fail --no-location "${rules_url}" --output "${{ github.action_path }}/.golangci.yml"
 
-        echo "GolangCIVersion=v$(head -n 1 "${{ github.action_path }}/.golangci.yml" | tr -d '# ')" >> "${GITHUB_OUTPUT}"
+        echo "GolangCIVersion=$(head -n 1 "${{ github.action_path }}/.golangci.yml" | tr -d '# ')" >> "${GITHUB_OUTPUT}"
       id: getenv
     - name: golangci-lint
       uses: golangci/golangci-lint-action@4afd733a84b1f43292c63897423277bb7f4313a9 # v8.0.0
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,4 +1,4 @@
-# 2.1.6
+# v2.1.6
 # Please don't remove the first line. It uses in CI to determine the golangci version
 version: "2"
 linters:
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,4 +1,4 @@
-# v2.1.6
+# 2.1.6
 # Please don't remove the first line. It uses in CI to determine the golangci version
 version: "2"
 linters:
--- a/Makefile
+++ b/Makefile
@@ -22,7 +22,7 @@ check-linter-version:
 ## lint: Runs the linters.
 lint: check-linter-version
 	echo "Running linters..."
-	golangci-lint run --out-format=tab ./...
+	golangci-lint run ./...
 
 ## tests: Executes any unit tests.
 tests:
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,7 @@ grpc-server-run:
 
 ## check-linter-version: Checks if the linter version is the same as the one specified in the linter config.
 check-linter-version:
-	(golangci-lint version | grep -E "version v?$(shell head -n 1 .golangci.yml | tr -d '\# ')") || echo "Your installation of golangci-lint is different from the one that is specified in k6's linter config (there it's $(shell head -n 1 .golangci.yml | tr -d '\# ')). Results could be different in the CI."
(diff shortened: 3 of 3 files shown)


tbourrely:
Regarding the version pattern matching, I have no idea if golangci-lint version output is platform / architecture dependent.

.golangci.yml
```diff
@@ -1,4 +1,4 @@
-# v2.1.6
+# 2.1.6
```
mstoykov, line 1: I do not think this is needed?

It was prefixed with `v` before the update to v2 and it seems you need to update another place to make it work.

I tried locally and the make works only with the other changes, so I am not certain why this is needed.

mstoykov (changes requested):
Thank you for the contribution :bow: 

The Makefile changes seem fine, and I probably should've done them originally :facepalm: 

The other changes though seem to not be necessary. I can't figure out what they are fixing, so can you please expand on why it was done?

.golangci.yml
tbourrely: Of course !
It is because of the `check-linter-version` target which extracts the golangci-lint version from the configuration file and "compares" it which the installed version using grep.

With `v2.1.6`, in the configuration file, the make target expects `v2.1.6` to be in the `golangci-lint version` output. Although on my system, `golangci-lint version` output does not contain `v2.1.6` __but__ `2.1.6`:

```
golangci-lint has version 2.1.6 built with go1.24.2 from eabc2638 on 2025-05-04T15:41:19Z
```

Therefore the make target considers the version to be invalid even when it is the right one.

Although I do not know if `golangci-lint version` output the version like `v<version>` on some builds and only `<version>` on others :thinking:

tbourrely: I can totally rollback this change if needed and keep only the removal of the deprecated flag :relaxed:

mstoykov: Hm, apparently I can not read, sorry about that :bow: 

Can you provide some data how you are installing this and what is your OS?

tbourrely: no worries.
Here is the os infos: 

```
 ~  cat /etc/os-release | grep PRETTY_NAME
PRETTY_NAME="Ubuntu 24.10"
```

I followed golangci-lint install guide (binaries section): https://golangci-lint.run/welcome/install/#binaries

Removing the binary and doing it again leads to the same result:

```
 ~  golangci-lint version
Command 'golangci-lint' not found, but can be installed with:
sudo snap install golangci-lint
 !  ~  # binary will be $(go env GOPATH)/bin/golangci-lint
       curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/HEAD/install.sh | sh -s -- -b $(go env GOPATH)/bin v2.1.6

       golangci-lint --version
golangci/golangci-lint info checking GitHub for tag 'v2.1.6'
golangci/golangci-lint info found version: 2.1.6 for v2.1.6/linux/amd64
golangci/golangci-lint info installed /home/keiko/go/bin/golangci-lint
golangci-lint has version 2.1.6 built with go1.24.2 from eabc2638 on 2025-05-04T15:41:19Z
```

tbourrely: If you wish, I can isolate the removal of the deprecated flag in another PR and we can keep digging about the version matching here.

If on your side `golangci-lint version` uses the `v<version>` format, maybe the proper solution would be to rework the `grep` call in order to match both `version <version>` __and__ `version v<version>` by using a regex, what do you think ?

tbourrely: I pushed this change as a third commit (i'll squash if we keep it :smile:)

mstoykov: Yeah - can we split thsi in two.

I think this might be a bug with the snap versioning. I can't find anything in the workflow of how it is built, but it seems strange. 

Can you potentially report it and ask why it is differing from if you do `go install` ? (which is how i have it installed)

.github/actions/lint/action.yml
```diff
@@ -24,7 +24,7 @@ runs:
         echo "Downloading '${rules_url}' ..."
         curl --silent --show-error --fail --no-location "${rules_url}" --output "${{ github.action_path }}/.golangci.yml"
 
-        echo "GolangCIVersion=$(head -n 1 "${{ github.action_path }}/.golangci.yml" | tr -d '# ')" >> "${GITHUB_OUTPUT}"
+        echo "GolangCIVersion=v$(head -n 1 "${{ github.action_path }}/.golangci.yml" | tr -d '# ')" >> "${GITHUB_OUTPUT}"
```
mstoykov, line 27:
```suggestion
        echo "GolangCIVersion=$(head -n 1 "${{ github.action_path }}/.golangci.yml" | tr -d '# ')" >> "${GITHUB_OUTPUT}"
```

mstoykov (dismissed):
It seems like the binary releases all have it without the `v` .

And after some more checking - it seems that it has always been like this. It seems like we just never did check that and nobody used this so far with binary release.

Makefile
```diff
@@ -17,12 +17,12 @@ grpc-server-run:
 
 ## check-linter-version: Checks if the linter version is the same as the one specified in the linter config.
 check-linter-version:
-	(golangci-lint version | grep "version $(shell head -n 1 .golangci.yml | tr -d '\# ')") || echo "Your installation of golangci-lint is different from the one that is specified in k6's linter config (there it's $(shell head -n 1 .golangci.yml | tr -d '\# ')). Results could be different in the CI."
+	(golangci-lint version | grep -E "version v?$(shell head -n 1 .golangci.yml | tr -d '\# ')") || echo "Your installation of golangci-lint is different from the one that is specified in k6's linter config (there it's $(shell head -n 1 .golangci.yml | tr -d '\# ')). Results could be different in the CI."
```
mstoykov, line 20:
```suggestion
	(golangci-lint version | grep -E "version v?$(shell head -n 1 .golangci.yml | tr -d '\# v')") || echo "Your installation of golangci-lint is different from the one that is specified in k6's linter config (there it's $(shell head -n 1 .golangci.yml | tr -d '\# ')). Results could be different in the CI."
```

mstoykov (approved):

inancgumus (approved):

tbourrely:
Hello sorry for not being so active the last few days on this PR. 
Thanks for the details, fixes, approves and merge 😄

PR #4787, opened 2025-05-15
Fix thresholds evaluated incorrectly

### LINKED ISSUE #4283: Thresholds evaluated incorrectly
### Brief summary

When test script has errors thresholds is evaluated incorrectly indicating success instead of failure.

### k6 version

0.49

### OS

macOS 14.6

_No response_

Run test with below script:
```
import { check } from "k6";
import http from "k6/http";
export const options = {
  stages: [
    { duration: "2s", target: 1 }
  ],
  thresholds: {
    "checks": [{ threshold: "rate>=0.9", abortOnFail: true }]
  }
}


export default function (setupData) {

  const result = http.get("https://test-api.k6.io/public/crocodiles/");
  throw new Error("This is an error");

  check(result, {
    "status:200": result.status === 200,
  });
}
```
Test is not failed even though `checks` metric value is 0. The summary json also indicates same:
```
"checks": {
      "type": "rate",
      "contains": "default",
      "values": {
        "passes": 0,
        "fails": 0,
        "rate": 0
      },
      "thresholds": {
        "rate>=0.9": {
          "ok": true
        }
      }
    },
```

### Expected behaviour

Test should fail when thresholds is breached even if test script has errors.

### Actual behaviour

Test is passed even though thresholds is breached.

Fernando-hub527:
## What?

This PR ensures that once thresholds are defined, they are still evaluated correctly even if check(...) is never called.

## Why?
Guarantee that if an error occurs before any checks are executed, the threshold evaluation will correctly register a failure.

## Checklist

## Related PR(s)/Issue(s)

Closes #4283

### DIFF
--- a/internal/output/summary/data.go
+++ b/internal/output/summary/data.go
@@ -296,7 +296,11 @@ func populateSummaryChecks(
 
 	summaryGroup.Checks.Metrics.Fail.Values["passes"] = totalChecks - successChecks
 	summaryGroup.Checks.Metrics.Fail.Values["fails"] = successChecks
-	summaryGroup.Checks.Metrics.Fail.Values["rate"] = (totalChecks - successChecks) / totalChecks
+	if totalChecks > 0 {
+		summaryGroup.Checks.Metrics.Fail.Values["rate"] = (totalChecks - successChecks) / totalChecks
+	} else {
+		summaryGroup.Checks.Metrics.Fail.Values["rate"] = 0
+	}
 
 	summaryGroup.Checks.OrderedChecks = groupData.checks.orderedChecks
 }
--- a/metrics/thresholds.go
+++ b/metrics/thresholds.go
@@ -205,6 +205,8 @@ func (ts *Thresholds) Run(sink Sink, duration time.Duration) (bool, error) {
 		// would lead to [#2520](https://github.com/grafana/k6/issues/2520)
 		if sinkImpl.Total > 0 {
 			ts.sinked["rate"] = float64(sinkImpl.Trues) / float64(sinkImpl.Total)
+		} else {
+			ts.sinked["rate"] = 0
 		}
 	default:
 		return false, fmt.Errorf("unable to run Thresholds; reason: unknown sink type")
--- a/metrics/thresholds_test.go
+++ b/metrics/thresholds_test.go
@@ -731,6 +731,16 @@ func TestThresholdsRun(t *testing.T) {
 			want:    false,
 			wantErr: false,
 		},
+		{
+			name: "when sink type is rate and sink.Total is 0, should set rate with 0",
+			args: args{
+				sink:                 &RateSink{Trues: 0, Total: 0},
+				thresholdExpressions: []string{"rate>=0.5"},
+				duration:             0,
+			},
+			want:    false,
+			wantErr: false,
+		},
 	}
 	for _, testCase := range tests {
 		testCase := testCase


Fernando-hub527:
As far as I could determine, the issue was happening because, to avoid a division by zero, the rate metric wasn’t populated when the total number of checks was zero ([see code here](https://github.com/grafana/k6/blob/master/metrics/thresholds.go#L206C3-L206C26)). That means if an error occurred before any checks were executed, the threshold failure was never triggered.

Here’s an example of the threshold evaluation after the fix:

**code:**
Captura de tela de 2025-05-14 22-57-57

**result:**
Captura de tela de 2025-05-14 22-58-48

ankur22 (approved):
Thanks for the PR @Fernando-hub527!

I took a look at the change and used the following script to test before and after the fix:

```js
import { check } from "k6";
import http from "k6/http";

export const options = {
 stages: [
  { duration: "2s", target: 1 },
 ],
 thresholds: {
  "checks": [{threshold: "rate>0.9", abortOnFail: true}],
 },
};

export default function () {
  const res = http.get("https://test-api.k6.io/public/crocodiles");
  throw new Error('This is an error');

  check(res, {
    "status was 200": (res) => res.status === 200,
  });
}
```

Before the fix:

```
  █ THRESHOLDS 

    checks
    ✓ 'rate>0.9' rate=0.00%


  █ TOTAL RESULTS 

    checks_total.......................: 0     0/s
    checks_succeeded...................: 0.00% 0 out of 0
    checks_failed......................: NaN%  0 out of 0
```

After the fix:

```
  █ THRESHOLDS 

    checks
    ✗ 'rate>0.9' rate=0.00%


  █ TOTAL RESULTS 

    checks_total.......................: 0     0/s
    checks_succeeded...................: 0.00% 0 out of 0
    checks_failed......................: 0.00% 0 out of 0
```

NOTE: The results are focusing on the areas that differ between the two test runs. Other metrics are presented such as HTTP, but they look correct in both test runs.

I think with the fix is a better outcome than before, so i'm approving it.

joanlopez:
Hi @Fernando-hub527,

We're discussing internally if really want to merge this new behavior for thresholds as-is, or perhaps offer another workaround to sort it out.

So, could you please split it into two different PRs, one for changing the threshold evaluation behavior and the other one to fix the `NaN%`? You can leave this one as-is, just extract the small bits into another one.

This way we can merge this part of your contribution and close https://github.com/grafana/k6/issues/4842. 
And later continue with the ongoing discussion around thresholds.
cc/ @mstoykov for visibility.

Thanks! 🙇🏻 🙌🏻

Fernando-hub527:
Hi @joanlopez, sorry for the delay.
Here's the PR containing only the fix for the NaN issue: #4878.
If you think it would be useful to share the discussion, I'd be happy to help.

joanlopez:
Hey @Fernando-hub527,

I'm going to close this pull request because after some internal discussions, I even stronger believe we won't want to change this behavior on thresholds, as it would introduce a breaking change, and it's unlikely that's always going to be the desired behavior by users, while it won't be configurable.

So, we'll likely agree on another workaround, like "exposing" the "total" value for Rate metrics, letting users to define another threshold that ensures "total > 0"; or adding a new option, like `abortOnFail`, but specifically for the case where there's no data at all, like `failWhenNoData`.

Anyway, thanks for the other half of the PR that you split and that has already been merged, as that was really a small bug to be fixed. Really appreciated! 🙌🏻

Issue #4758, opened 2025-04-30
refactor provision callback into an interface


pablochacin:
Currently the launcher receives a provision function as a parameter. This function returns a `commandExecutor` that satisfies the dependencies.

Transform  this function into an interface to ease of debugging and codebase navigation in IDEs

```
type Provider interface {
         provision func(*state.GlobalState, k6deps.Dependencies) (commandExecutor, error)
}
````

PR #4801, opened 2025-05-20
initial integration

pablochacin:
## Checklist


internal/cmd/root.go
```diff
@@ -43,12 +43,20 @@ type rootCommand struct {
...
 	loggerIsRemote bool
+	launcher       *launcher
 }
 
+// creates a root command with a default launcher
 func newRootCommand(gs *state.GlobalState) *rootCommand {
+	return newRootWithLauncher(gs, newLauncher(gs))
+}
+
+// creates a root command with a launcher. Used to facilitate testing
```
codebien:
```suggestion
// newRootWithLauncher creates a root command with a launcher. Used to facilitate testing
```

internal/cmd/root.go
```diff
@@ -43,12 +43,20 @@ type rootCommand struct {
 	stopLoggersCh  chan struct{}
 	loggersWg      sync.WaitGroup
 	loggerIsRemote bool
+	launcher       *launcher
 }
 
+// creates a root command with a default launcher
```
codebien:
```suggestion
// newRootCommand creates a root command with a default launcher
```

internal/cmd/launcher.go
```diff
@@ -319,56 +313,18 @@ func analyze(gs *state.GlobalState, args []string) (k6deps.Dependencies, error)
...
-					return false
-				}
-			}
-			return true
-		case "archive", "inspect":
+// isAnalysisRequired returns a boolean indicating if dependency analysis is required for the command
+func isAnalysisRequired(cmd *cobra.Command) bool {
+	switch cmd.Name() {
+	case "run":
+		if cmd.Parent() != nil && cmd.Parent().Name() == "cloud" {
```
codebien, line 315:
```suggestion
        // exclude `k6 cloud run` command
		if cmd.Parent() != nil && cmd.Parent().Name() == "cloud" {
```

internal/cmd/launcher.go
```diff
@@ -25,63 +26,66 @@ var (
...
-	// If binary provisioning is not enabled, continue with the regular k6 execution path
-	if !l.gs.Flags.BinaryProvisioning {
-		l.gs.Logger.Debug("Binary Provisioning feature is disabled.")
-		l.commandExecutor.run(l.gs)
-		return
+// launch analyzies the command to be executed and its input (e.g. script)
+// to identify if its dependencies. It it has dependencies tha cannot be satisfied by the
+// current binary, it obtains a custom binary usign the provision function and delegates
+// the execution of the command to this binary. if not, continues with the execution of the
+// command in the current binary.
```
codebien: This comment needs some adjustment.

internal/cmd/launcher.go
```diff
@@ -25,63 +26,66 @@ var (
...
+	// path represents the local file path
+	// on the file system of the binary
+	path string
+}
 
-	// provision generates a custom binary from the received list of dependencies
-	// with their constrains, and it returns an executor that satisfies them.
-	provision func(*state.GlobalState, k6deps.Dependencies) (commandExecutor, error)
+// provider defines the interface for provisioning a custom k6 binary for a set of dependencies
+type provider interface {
```
codebien:
```suggestion
type provisioner interface {
```
It sounds to me more idiomatic. If we want `provider` then I expect a `provide` function.

internal/cmd/launcher_test.go
```diff
@@ -222,133 +231,36 @@ func TestLauncherLaunch(t *testing.T) {
...
-			t.Parallel()
+			// replace command's the RunE function by a mock that indicates if the command was executed
+			runECalled := false
+			cmd.RunE = func(_ *cobra.Command, _ []string) error {
+				runECalled = true
+				return nil
+			}
 
-			script := scriptNameFromArgs(tc.args)
+			// TODO: check error
```
```suggestion
			// TODO: check error
```
Can we address this TODO, please?

internal/cmd/launcher.go
```diff
@@ -172,18 +171,11 @@ func (b *customBinary) run(gs *state.GlobalState) {
...
-
-func (b *currentBinary) run(gs *state.GlobalState) {
-	ExecuteWithGlobalState(gs)
-}
-
-// customBuildRequired checks if the build is required
+// isCustomBuildRequired checks if the build is required
 // it's required if there is one or more dependencies other than k6 itself
 // or if the required k6 version is not satisfied by the current binary's version
 // TODO: get the version of any built-in extension and check if they satisfy the dependencies
```
codebien, line 176: I have the feeling we have an issue for this TODO. Please, mention here directly the issue so they are bidirectionally linked.

pablochacin: Added. This isssue is addressed in this PR https://github.com/grafana/k6/pull/4812

internal/cmd/launcher_test.go
pablochacin: it is a leftover. execute does not return any error

internal/cmd/launcher.go
```diff
@@ -124,6 +128,8 @@ func (b *customBinary) run(gs *state.GlobalState) {
 	// Copy environment variables to the k6 process and skip binary provisioning feature flag to disable it.
 	// If not disabled, then the executed k6 binary would enter an infinite loop, where it continuously
 	// process the input script, detect dependencies, and retrigger provisioning.
+	// This can be avoided by checking if the current binary has already extensins that
+	// satisfies the dependencies. See comment in isCustomBuildRequired function.
```
mstoykov:
```suggestion
	// This can be avoided by checking if the current binary has already extensions that
	// satisfies the dependencies. See comment in isCustomBuildRequired function.
```

pablochacin:
@mstoykov 
 
> In the sense that it isn't being integrated, it does practically the same thing as if it is a separate program, but now internally

I think part of this impression may come from the fact that I split the integration into two PRs (this and https://github.com/grafana/k6/pull/4811)) to reduce the scope of review. 

The difference of this approach from having an external program (which is basically how we did the [initial "integration" ](https://github.com/grafana/k6/pull/4671)as a wrapper of `k6` command),  is that now we rely on `k6` command logic for parsing arguments and to read the source (see https://github.com/grafana/k6/pull/4811)  

> Is there a reason this was not considered or are there any known problems with something like this that I do not know about ?

Basically, to have a clear execution path: either with binary provisioning or without it. See my comment below about why I considered this an advantage.

> After this depending on the use case I guess you can error out with a specific error of "We need this modules" and then capturing it potentially in a similar way in the root command.

I'm not sure about this option. It depends on the command not having side-effects, and therefore stopping its execution  and then retrying with binary provisioning will not find these side-effects. I find it harder to reason about this execution logic that the one existing now.

Also, I'm not sure how k6 handles importing files. From what I saw in the `readSource` we are only reading the "main" test script. What happens if it imports other files? can we analyze all the import tree before starting execution?

> This IMO not only will reduce the code here, but will also be way more future prove. As you won't need to try to work around you not having parsed some configuration or us having anew command that reads the source code that needs to be wrapped.

> I looked at some of the other changes in the other PRs, but it doesn't seem like any of them stop trying to parse the configuration as k6, just to do it again.

Not sure what duplicated parsing you refer to? There's no additional parsing, unless you mean parsing the test code itself searching dependencies. Can you give an example of this duplicated parsing? (again, consider this in the context of https://github.com/grafana/k6/pull/4811. 

In general, I'm not against a deeper integration, but maybe only if we ensure this `readSource` function is executed earlier in the command execution to avoid side-effects and we can decide which way to follow.

internal/cmd/launcher.go
```diff
@@ -90,29 +93,30 @@ func (l *Launcher) Launch() {
...
+
+	return nil
 }
 
-// customBinary runs the requested commands
-// on a different binary on a subprocess passing the original arguments
-type customBinary struct {
-	// path represents the local file path
-	// on the file system of the binary
-	path string
```
codebien: Can we restore this struct definition to keep the same position as before? I don't see the reason to separate the struct and `run` method definitions.

internal/cmd/launcher.go
```diff
@@ -215,11 +207,19 @@ func customBuildRequired(baseK6Version string, deps k6deps.Dependencies) bool {
 	return !k6Dependency.Constraints.Check(k6Ver)
 }
 
-// k6buildProvision returns the path to a k6 binary that satisfies the dependencies and the list of versions it provides
-func k6buildProvision(gs *state.GlobalState, deps k6deps.Dependencies) (commandExecutor, error) {
-	token, err := extractToken(gs)
+// k6buildProvider provides a k6 binary that satisfies the dependencies using the k6build service
+type k6buildProvider struct {
```
codebien:
```suggestion
type k6buildProvisioning struct {
```
I think you missed the renaming here. However, what is the value to pass from a single function to a struct? We now have 3 things (a struct, a constructor and a method) to do the same thing we were doing with a single function before.

internal/cmd/launcher_test.go
```diff
@@ -2,26 +2,43 @@ package cmd
...
-	rc      int
+	err     error
+}
+
+func (m *mockExecutor) run(_ *state.GlobalState) error {
+	m.invoked = true
+	return m.err
+}
+
+type mockProvider struct {
```
```suggestion
type mockProvisioner struct {
```

internal/cmd/root.go
```diff
@@ -43,12 +43,20 @@ type rootCommand struct {
...
 	loggerIsRemote bool
+	launcher       *launcher
 }
 
+// newRootCommand creates a root command with a default launcher
 func newRootCommand(gs *state.GlobalState) *rootCommand {
+	return newRootWithLauncher(gs, newLauncher(gs))
+}
+
+// newRootWithLauncher creates a root command with a launcher. Used to facilitate testing
```
codebien:
```suggestion
// newRootWithLauncher creates a root command with a launcher.
// It facilitates unit testing scenarios.
```

codebien (approved):
LGTM in terms of approach so I've approved. There are a few comments I would like to see addressed before merging.

internal/cmd/launcher.go
```diff
@@ -283,19 +283,13 @@ func analyze(gs *state.GlobalState, args []string) (k6deps.Dependencies, error)
...
 
-	if !isAnalysisRequired(args) {
-		gs.Logger.
-			Debug("The command to execute does not require dependency analysis.")
-		return k6deps.Dependencies{}, nil
-	}
-
-	scriptname := scriptNameFromArgs(args)
-	if len(scriptname) == 0 {
+	if len(args) == 0 {
```
mstoykov: This can't realistically happen as the cli flag parsing step checks for all the commands that are analyziable that there is an argument.

But I actually started looking into this because IMO even if it didn't - this should just be handled by the analyzer saying "nothing is wrong and there is nothing to build" and let the rest of k6 actually error, instead of trying to copy the same behaviour from there.

mstoykov (dismissed):
I am approving this so we can continue merging the rest of the PRs. And because this seems like an improvement on the current implementation/

I still do think that this is more wrapping k6 than integrating with it. 

It is now parsing the config file as an example. It also checks and returns errors that arguably the "wrapped" commands should do. 

I do agree with the changes from @codebien and I do have a comment, although mine is currently kind of void due to it being deadcode AFAIK.

internal/cmd/launcher.go
pablochacin: I tend to agree, and also we use call-back functions extensively in the k6 code, but this was requested by @oleiade in https://github.com/grafana/k6/pull/4671 (tracked in https://github.com/grafana/k6/issues/4758). I assumed we were moving towards not having more call-back functions.

internal/cmd/launcher.go
pablochacin: @codebien can you elaborate on what adjustments?

codebien: This is old. The current version is fine.

codebien (approved):

Issue #464, opened 2018-01-17
Make use of a local http server for tests


cstyan:
current http tests rely on `httpbin.org`, we should make use of a local http server (perhaps written in go) so that we're not relying on an external resource, and to possibly support testing `CONNECT` and `TRACE` methods

https://github.com/loadimpact/k6/pull/451#issuecomment-357296743

na--:
I'll start to work on this today. I could find two go libraries ([1](https://github.com/ahmetb/go-httpbin), [2](https://github.com/mccutchen/go-httpbin)) that implement httpbin-like functionality. I chose the second one for now because of [this](https://github.com/mccutchen/go-httpbin#motivation--prior-art) and I've already used it in a test [here](https://github.com/loadimpact/k6/pull/523), it seems to be [very easy to integrate](https://github.com/loadimpact/k6/blob/a0246467ad76e312beb9e67a70f0f52b0f6ca5ef/core/engine_test.go#L495-L496). Since the [migration](https://github.com/loadimpact/k6/pull/526) to [`dep`](https://github.com/golang/dep) it's already vendored properly as well.

PR #4809, opened 2025-05-22
Fix fill method in injection script

ankur22:
## What?

Preventing an early exit in the fill method in the injection script to actually allow the keyboard to enter the value into the input element.

## Why?

This fill function was failing to fill the input text field. It would seem that it was returning to early, when it should have continued onto the end of the function and returned "needsinput".

## Test script

Without the fix this script was unable to enter the username in the input element. After the fix it does:

```js
import { browser } from 'k6/browser';

export const options = {
  scenarios: {
    ui: {
      options: {
        browser: {
          type: "chromium",
        },
      },
    },
  }
}

  const context = await browser.newContext();
  const page = await context.newPage();

  await page.goto('https://grafana.com/auth/sign-in');
  
  await page.locator('input[name="login"]').fill('my-username');

  await page.waitForTimeout(5000);
}
```

## Checklist


joanlopez:
Is this something that can be tested? Like, with a unit test, I mean 🤔

ankur22:
>Is this something that can be tested? Like, with a unit test, I mean 🤔

I did try, and i wasn't able to get it to not work before the fix! Very strange. I'll give it another go though 👍

ankur22:
@joanlopez I've added an integration test in 9a3567596e3278ffa5caa319ab546dfc63a31885

internal/js/modules/k6/browser/tests/static/react_input.html
```diff
@@ -0,0 +1,40 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="UTF-8" />
+  <title>Controlled Grafana-like Input Demo</title>
+  <!-- React and ReactDOM via UNPKG -->
+  <script crossorigin src="https://unpkg.com/react@18/umd/react.development.js"></script>
+  <script crossorigin src="https://unpkg.com/react-dom@18/umd/react-dom.development.js"></script>
```
mstoykov, line 8: In general it is not great when our test depends on other services being up and running and we have tried to remove those.

https://github.com/grafana/k6/issues/464 https://github.com/grafana/k6/issues/2302

mstoykov (dismissed):
LGTM, but it will be nice if we do not have to import 2 react dependancies from a third party to test this.

internal/js/modules/k6/browser/tests/static/react_input.html
joanlopez: I guess a "quick-win" could be copy-pasting those files locally, right? 🤔 So, this way we don't rely on an external server, to be available. Reproducing the issue (if React-specific) with a subset of the library might more challenging, tho.

ankur22: Yeah, that makes sense. I tried to isolate the changes from the react dependency files, but the work that needs to go into doing that outweighs the benefits IMO. I've just added the files that are needed as static files. The commit is 16a4550fd69fcb7f7e3927a16375f34815fd44e5

inancgumus: Instead of saving these scripts in the repository and bloating it, why don't we download them **only once**, right before the test starts, if the files don't exist?

inancgumus (approved):

internal/js/modules/k6/browser/tests/static/react_input.html
ankur22: That would still depend on a network connection, i guess we want to avoid that where possible. We're already vendoring our Go dependency, so i feel like this is good enough for now. Preferably it would be nice to only include the areas of the code that the test actually uses from the react library -- i tried manually (and using AI) to do this but it was taking too long so i abandoned it.

inancgumus: I wouldn't worry much about the network connection for one-time-download assets, especially if we put it onto one of our servers, like S3 or jslib.k6.io. Getting Go modules always involves a network anyway. That said, I'm fine with the current solution as well 🤗

Issue #4490, opened 2021-11-12
Locator API


robingustafsson:
We started the development of the first version on May 16th.

### Strict forwarders (milestone: [v0.7.0](https://github.com/grafana/xk6-browser/milestone/6))

- [x] grafana/xk6-browser#660

### Strict forwarders (milestone: [v0.4.0](https://github.com/grafana/xk6-browser/milestone/3))

The locator methods that forward calls to the `Frame` methods with the strict mode on (meaning: only select a single element).

- [x] grafana/xk6-browser#308
- [x] grafana/xk6-browser#334
- [x] grafana/xk6-browser#311
- [x] grafana/xk6-browser#331
- [x] grafana/xk6-browser#359
- [x] grafana/xk6-browser#346
- [x] grafana/xk6-browser#348
- [x] grafana/xk6-browser#349
- [x] grafana/xk6-browser#347
- [x] grafana/xk6-browser#351
- [x] grafana/xk6-browser#350
- [x] grafana/xk6-browser#338
- [x] grafana/xk6-browser#339
- [x] grafana/xk6-browser#341
- [x] grafana/xk6-browser#343
- [x] grafana/xk6-browser#344
- [x] grafana/xk6-browser#357
- [x] grafana/xk6-browser#345
- [x] grafana/xk6-browser#353
- [x] grafana/xk6-browser#352
- [x] grafana/xk6-browser#358
- [x] grafana/xk6-browser#356
- [x] grafana/xk6-browser#342
- [x] grafana/xk6-browser#335
- [x] grafana/xk6-browser#360

### Just-in-time element finder methods

The locator methods wrap around existing `ElementHandle` methods. For example:

```typescript
// locator method
async evaluate<R, Arg>(pageFunction: structs.PageFunctionOn, arg?: Arg, options?: TimeoutOptions): Promise<R> {
  return this._withElement(h => h.evaluate(pageFunction, arg), options?.timeout);
}

// wrapped method (jsHandle.ts)
async evaluate<R, Arg>(pageFunction: structs.PageFunctionOn, arg?: Arg): Promise < R > { ... }
```

`_withElement` queries the element using `waitForSelector` in the current frame and returns a disposed element handle:

```typescript
  private async _withElement<R>(task: (handle: ElementHandle<SVGElement | HTMLElement>, timeout?: number) => Promise<R>, timeout?: number): Promise<R> {
    timeout = this._frame.page()._timeoutSettings.timeout({ timeout });
[...]

PR #4825, opened 2025-06-04
Add `locator` APIs `nth`, `first`, and `last`

ankur22:
## What?

This PR adds three APIs for the `locator` class. These are:

* `nth` which can be used to return the nth element given a selector (e.g. if we have many rows, and we want to select the 5th row in a table, then we can do `locator.nth(4)`).
* `first` which will return the first element (0th) given a selector (e.g. if we have many rows, and we want to select the first row in a table, then we can do `locator.first()` instead of `locator.nth(0)`).
* `last` which will return the last element given a selector (e.g. if we have many rows, and we want to select the last row in a table, then we can do `locator.last()` instead of `locator.nth(-1)`).

## Why?

There are many benefits of working with `locator` based APIs over element based APIs such as `$` and `$$`. The main two are that we can declare it once and use it over an over again even when the page navigates or the underlying DOM changes, and it performs actionability checks such as auto waiting for the element to be visible before trying to work with it. I.e. it's less fragile and easier to work with.

If we wanted to select a row in a table, before we would have to do:

```js
    const links = await page.$$('a[class="thumbnail"]');
    if (links && links.length > 0) {
        const link = links[0];
        while (!link.isVisible()) {
            sleep(1);
        }
        link.click();
    }
```

Now we can do:

```js
    await page.locator('a[class="thumbnail"]').first().click();
```

## Checklist

## Related PR(s)/Issue(s)

https://github.com/grafana/k6/issues/4490

### DIFF
--- a/internal/js/modules/k6/browser/common/locator.go
+++ b/internal/js/modules/k6/browser/common/locator.go
@@ -429,6 +429,12 @@ func (l *Locator) innerText(opts *FrameInnerTextOptions) (string, error) {
 	return l.frame.innerText(l.selector, opts)
 }
 
+// Last will return the last child of the element matching the locator's
+// selector.
+func (l *Locator) Last() *Locator {
+	return NewLocator(l.ctx, fmt.Sprintf("%s >> nth=-1", l.selector), l.frame, l.log)
+}
+
 // Nth will return the nth child of the element matching the locator's
 // selector.
 func (l *Locator) Nth(nth int) *Locator {
--- a/internal/js/modules/k6/browser/common/locator.go
+++ b/internal/js/modules/k6/browser/common/locator.go
@@ -3,6 +3,7 @@ package common
 import (
 	"context"
 	"fmt"
+	"strconv"
 	"time"
 
 	"github.com/grafana/sobek"
@@ -335,7 +336,7 @@ func (l *Locator) fill(value string, opts *FrameFillOptions) error {
 // First will return the first child of the element matching the locator's
 // selector.
 func (l *Locator) First() *Locator {
-	return NewLocator(l.ctx, fmt.Sprintf("%s >> nth=0", l.selector), l.frame, l.log)
+	return NewLocator(l.ctx, l.selector+" >> nth=0", l.frame, l.log)
 }
 
 // Focus on the element using locator's selector with strict mode on.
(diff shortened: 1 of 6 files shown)


internal/js/modules/k6/browser/common/locator.go
```diff
@@ -325,6 +325,12 @@ func (l *Locator) fill(value string, opts *FrameFillOptions) error {
 	return l.frame.fill(l.selector, value, opts)
 }
 
+// First will return the first child of the element matching the locator's
+// selector.
+func (l *Locator) First() *Locator {
+	return NewLocator(l.ctx, fmt.Sprintf("%s >> nth=0", l.selector), l.frame, l.log)
```
inancgumus: I would usually not comment on such a thing, but since we want to optimize the module, I'd use an old-fashioned string concat here, instead of the less-performant `Sprintf`, which may allocate and run a lot of reflection code:

```suggestion
	return NewLocator(l.ctx, l.selector + " >> nth=0", l.frame, l.log)
```

internal/js/modules/k6/browser/common/locator.go
```diff
@@ -416,6 +422,18 @@ func (l *Locator) innerText(opts *FrameInnerTextOptions) (string, error) {
 	return l.frame.innerText(l.selector, opts)
 }
 
+// Last will return the last child of the element matching the locator's
+// selector.
+func (l *Locator) Last() *Locator {
+	return NewLocator(l.ctx, fmt.Sprintf("%s >> nth=-1", l.selector), l.frame, l.log)
```
inancgumus:
```suggestion
	return NewLocator(l.ctx, l.selector +" >> nth=-1", l.frame, l.log)
```

internal/js/modules/k6/browser/common/locator.go
```diff
@@ -416,6 +422,18 @@ func (l *Locator) innerText(opts *FrameInnerTextOptions) (string, error) {
...
+}
+
+// Nth will return the nth child of the element matching the locator's
+// selector.
+func (l *Locator) Nth(nth int) *Locator {
+	return NewLocator(l.ctx, fmt.Sprintf("%s >> nth=%d", l.selector, nth), l.frame, l.log)
```
inancgumus:
```suggestion
	return NewLocator(l.ctx, l.selector +" >> nth="+ strconv.Itoa(nth), l.frame, l.log)

inancgumus (approved):
Nice!

inancgumus (approved):

PR #4831, opened 2025-06-05
Fix user in dockerfile

ankur22:
## What?

Defaults to using user `12345` instead of `k6`.

## Why?

This fix helps those who are working with the image in k8s. Before this change the user had to work with runAsUser in the pod manifest file, which could lead to issues if the correct user wasn't used (which is only 12345). This will mitigate that risk since we default to working with 12345 and not k6.

## Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,10 +8,10 @@ RUN CGO_ENABLED=0 GOOS=$TARGETOS GOARCH=$TARGETARCH go build -trimpath -o /usr/b
 # Runtime stage
 FROM alpine:3.21 as release
 
-RUN adduser -D -u 12345 -g 12345 k6
+RUN adduser -D -u 12345 -g 12345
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -8,10 +8,10 @@ RUN CGO_ENABLED=0 GOOS=$TARGETOS GOARCH=$TARGETARCH go build -trimpath -o /usr/b
 # Runtime stage
 FROM alpine:3.21 as release
 
-RUN adduser -D -u 12345 -g 12345 k6
+RUN adduser -D -u 12345 -g 12345
```
nadiamoe: I think this line is fine as it was, the user needs a name regardless of whether we refer to it by ID or by name.

```suggestion
RUN adduser -D -u 12345 -g 12345 k6
```

joanlopez (approved):

PR #4833, opened 2025-06-05
Switch k6 packager to Vault

inancgumus:
## What?

Retrieves secrets from Vault and adjusts the k6packager deployment process.

Passing builds:
- https://github.com/grafana/k6/actions/runs/15641571150/job/44070316796
- https://github.com/grafana/k6/actions/runs/15641760295/job/44070908513

See grafana/k6-cloud#3483 for more details.

## Why?

For stronger security.

## Checklist

## Related PR(s)/Issue(s)

- grafana/k6-cloud#3483

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -370,6 +370,7 @@ jobs:
       actions: read
       contents: read
       packages: read
+      id-token: write # Required for Vault
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
@@ -391,6 +392,11 @@ jobs:
           mv "dist/k6-$VERSION-windows-amd64.msi" "dist/k6-$VERSION-amd64.msi"
           mv "dist/k6-$VERSION-linux-amd64.rpm" "dist/k6-$VERSION-amd64.rpm"
           mv "dist/k6-$VERSION-linux-amd64.deb" "dist/k6-$VERSION-amd64.deb"
+      - uses: grafana/shared-workflows/actions/get-vault-secrets@get-vault-secrets/v1.2.1
+        with:
+          repo_secrets: |
+            PGP_SIGN_KEY_PASSPHRASE=pgp:PGP_SIGN_KEY_PASSPHRASE
+            PGP_SIGN_KEY=pgp:PGP_SIGN_KEY
       - name: Setup docker compose environment
         run: |
           cat > packaging/.env <<EOF
@@ -398,9 +404,9 @@ jobs:
           AWS_SECRET_ACCESS_KEY=${{ secrets.AWS_SECRET_ACCESS_KEY }}
           AWS_DEFAULT_REGION=us-east-1
           AWS_CF_DISTRIBUTION="${{ secrets.AWS_CF_DISTRIBUTION }}"
-          PGP_SIGN_KEY_PASSPHRASE=${{ secrets.PGP_SIGN_KEY_PASSPHRASE }}
+          PGP_SIGN_KEY_PASSPHRASE=${{ env.PGP_SIGN_KEY_PASSPHRASE }}
           EOF
-          echo "${{ secrets.PGP_SIGN_KEY }}" > packaging/sign-key.gpg
+          echo "${{ env.PGP_SIGN_KEY }}" > packaging/sign-key.gpg
       - name: Publish packages
         env:
           GITHUB_ACTOR: ${{ github.actor }}
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -232,6 +232,7 @@ jobs:
     permissions:
       contents: read
       actions: read
+      id-token: write # Required for Vault
 
     runs-on: windows-latest
     defaults:
@@ -273,6 +274,12 @@ jobs:
           candle.exe -arch x64 "-dVERSION=$env:VERSION" k6.wxs
           light.exe -ext WixUIExtension k6.wixobj
 
+      - uses: grafana/shared-workflows/actions/get-vault-secrets@get-vault-secrets/v1.2.1
+        with:
+          repo_secrets: |
+            WIN_SIGN_CERT=winsign:WIN_SIGN_CERT
+            WIN_SIGN_PASS=winsign:WIN_SIGN_PASS
+
       - name: Sign Windows binary and .msi package
         # GH secrets are unavailable when building from project forks, so this
         # will fail for external PRs, even if we wanted to do it. And we don't.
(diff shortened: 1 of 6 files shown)


.github/workflows/build.yml
```diff
@@ -384,16 +392,30 @@ jobs:
...
           mv "dist/k6-$VERSION-linux-amd64.deb" "dist/k6-$VERSION-amd64.deb"
+      - uses: grafana/shared-workflows/actions/get-vault-secrets@get-vault-secrets-v1.2.0
+        with:
+          repo_secrets: |
+            IAM_ROLE_ARN=deploy:packager-iam-role
+            AWS_CF_DISTRIBUTION=cloudfront:AWS_CF_DISTRIBUTION
+            S3_BUCKET=s3:AWS_S3_BUCKET
+            PGP_SIGN_KEY_PASSPHRASE=pgp:PGP_SIGN_KEY_PASSPHRASE
+            PGP_SIGN_KEY=pgp:PGP_SIGN_KEY
+      - uses: grafana/shared-workflows/actions/aws-auth@aws-auth-v1.0.1
```
codebien:
```suggestion
      - uses: grafana/shared-workflows/actions/aws-auth@aws-auth/v1.0.2
```
this is the new recommended method

.github/workflows/build.yml
```diff
@@ -273,6 +274,12 @@ jobs:
           candle.exe -arch x64 "-dVERSION=$env:VERSION" k6.wxs
           light.exe -ext WixUIExtension k6.wixobj
 
+      - uses: grafana/shared-workflows/actions/get-vault-secrets@get-vault-secrets-v1.2.0
```
codebien:
```suggestion
      - uses: grafana/shared-workflows/actions/get-vault-secrets@get-vault-secrets/v1.2.1
```

.github/workflows/build.yml
```diff
@@ -384,16 +392,31 @@ jobs:
...
         run: |
           cat > packaging/.env <<EOF
-          AWS_ACCESS_KEY_ID=${{ secrets.AWS_ACCESS_KEY_ID }}
-          AWS_SECRET_ACCESS_KEY=${{ secrets.AWS_SECRET_ACCESS_KEY }}
-          AWS_DEFAULT_REGION=us-east-1
-          AWS_CF_DISTRIBUTION="${{ secrets.AWS_CF_DISTRIBUTION }}"
-          PGP_SIGN_KEY_PASSPHRASE=${{ secrets.PGP_SIGN_KEY_PASSPHRASE }}
+          AWS_ACCESS_KEY_ID=${{ env.AWS_ACCESS_KEY_ID }}
+          AWS_CF_DISTRIBUTION="${{ env.AWS_CF_DISTRIBUTION }}"
+          AWS_DEFAULT_REGION=us-east-2
```
codebien: Are you sure about the change around the default region?

.github/workflows/build.yml
```diff
@@ -384,16 +392,31 @@ jobs:
...
+          AWS_SECRET_ACCESS_KEY=${{ env.AWS_SECRET_ACCESS_KEY }}
+          AWS_SESSION_TOKEN=${{ env.AWS_SESSION_TOKEN }}
+          PGP_SIGN_KEY_PASSPHRASE=${{ env.PGP_SIGN_KEY_PASSPHRASE }}
+          S3_BUCKET=${{ env.S3_BUCKET }}
           EOF
-          echo "${{ secrets.PGP_SIGN_KEY }}" > packaging/sign-key.gpg
+          printf '%s' "${{ env.PGP_SIGN_KEY }}" | base64 -d > packaging/sign-key.gpg
```
codebien: Why do we need now the base64?

.github/workflows/build.yml
inancgumus: I'm not sure about which region to pick here. Do you have suggestions? I'll see what happens when we leave it as the default.

.github/workflows/build.yml
inancgumus: ~It must be stored in Vault as a base64 string (to avoid weird issues as it needs to be put in JSON and safely transmitted). That's the practice Vault [recommends](https://developer.hashicorp.com/vault/tutorials/encryption-as-a-service/eaas-transit#encrypt-secrets). So, we're decrypting it here.~ I confused this with the Windows signing key. I'll just use the PGP key as text with newlines embedded instead of encoding/decoding.

.github/workflows/build.yml
```diff
@@ -273,6 +274,12 @@ jobs:
           candle.exe -arch x64 "-dVERSION=$env:VERSION" k6.wxs
           light.exe -ext WixUIExtension k6.wixobj
 
+      - uses: grafana/shared-workflows/actions/get-vault-secrets@get-vault-secrets/v1.2.1
+        with:
+          repo_secrets: |
+            WIN_SIGN_CERT=winsign:WIN_SIGN_CERT
+            WIN_SIGN_PASS=winsign:WIN_SIGN_PASS
+
```
codebien: Can we move this part in a dedicated pull request, as it doesn't have any dependency from the parts after. So we can merge it without waiting for the rest.

packaging/docker-compose.yml
```diff
@@ -1,20 +1,19 @@
...
-
 services:
   packager:
     build:
       context: .
       args:
-        - AWSCLI_VERSION=${AWSCLI_VERSION:-2.1.36}
-        - S3CMD_VERSION=${S3CMD_VERSION:-2.1.0}
+        - AWSCLI_VERSION=${AWSCLI_VERSION:-2.26.5}
+        - S3CMD_VERSION=${S3CMD_VERSION:-2.4.0}
```
codebien, line 7: Can't we use relaxed format for versioning? It would be good to use at least the latest patched version.

.github/workflows/build.yml
codebien: >Do you have suggestions? I'll see what happens when we leave it as the default.

Is it used for the s3 bucket? If yes, then `us-east-2` is correct.

packaging/docker-compose.yml
inancgumus: Wouldn't using the exact versioning be better for pinning down possible issues and leaving us in control?

.github/workflows/build.yml
inancgumus: Sure, but out of curiosity, what would be the benefit of releasing this part early?

packaging/docker-compose.yml
codebien: I agree.

.github/workflows/build.yml
codebien: To reduce the scope of the not migrated keys. However, this is up to you. If you prefer to keep them all together then it is fine to me.

.github/workflows/build.yml
inancgumus: Done in the last push.

codebien (approved):

Issue #4764, opened 2025-05-02
Role based selector engine for k6 browser


ankur22:
### Feature Description

The current way of working with elements in a frontend/browser test is to use CSS or XPath selectors. Users are finding that these selectors are both cumbersome and not maintainable when a website changes. These selectors usually rely on the DOM structure which can quite easily change, especially since new frontend frameworks will usually "compile" down to HTML, while changing many of the attributes on elements in the process.

We need a more maintainable way of working with selectors.

### Suggested Solution (optional)

The proposal is to work with the role based selector engine, one that behaves exactly like in Playwright. This will involve two steps:

1. Implement the role based selector engine in the [injected script](https://github.com/grafana/k6/blob/master/internal/js/modules/k6/browser/common/js/injected_script.js), like we have already done for the [XPath](https://github.com/grafana/k6/blob/master/internal/js/modules/k6/browser/common/js/injected_script.js) and [CSS](https://github.com/grafana/k6/blob/master/internal/js/modules/k6/browser/common/js/injected_script.js) selector engines.
2. Implement the [FrameLocator](https://playwright.dev/docs/api/class-framelocator) class, which is essentially a wrapper around the various selector engines. The one to focus on is [getByRole](https://playwright.dev/docs/api/class-framelocator#frame-locator-get-by-role), but it's worth exploring all the other APIs in that class too.

### Already existing or connected issues / PRs (optional)

I have worked on a POC. I basically copied over the injected script from Playwright and pasted it into the k6 injected script. It does seem to work with the following test script:

```js
import { browser } from 'k6/browser'

export const options = {
  scenarios: {
    ui: {
      executor: 'shared-iterations',
      options: {
        browser: {
          type: 'chromium',
        },
      },
    },
  },
}

export default async function () {
  const page = await browser.newPage()

  await page.goto('https://quickpizza.grafana.com/')

  await page.locator(
    'role=button[name="Pizza, Please!"]'
  ).click();

  await page.waitForTimeout(2000);

  await page.close();
}
```

ankur22:
PW:
1. https://github.com/microsoft/playwright/blob/0535a2ee1112ae750af8e64a32440c2f81db3b10/packages/playwright-core/src/client/page.ts#L701
2. https://github.com/microsoft/playwright/blob/0535a2ee1112ae750af8e64a32440c2f81db3b10/packages/playwright-core/src/client/frame.ts#L337
[...]

Issue #4320, opened 2024-07-05
native playwright ARIA-ROLE API


arukiidou:
### Feature Description

I would like to see an ARIA-ROLE like this
DX will be greatly improved.
It will be advantage from Jmeter

```ts
await expect(page.getByRole('heading', { name: 'Sign up' })).toBeVisible();

await page.getByRole('checkbox', { name: 'Subscribe' }).check();

await page.getByRole('button', { name: /submit/i }).click();
```

https://playwright.dev/docs/locators#locate-by-role

ankur22:
Marking as complete since the PR has been merged to implement this feature. It should be in k6 `v1.2.0`.

arukiidou:
- releates to https://github.com/grafana/k6/issues/4248

Issue #4248, opened 2024-02-29
Implement page.getBy* APIs


ankur22:
### Feature Description

Working with `page.locator` is better than working with `page.$` since it allows us to work with elements on the page at any moment. These APIs both face the same issue which is that they work with `XPath` or `CSS` selectors which are cumbersome to use, hard to work with if you're not a frontend engineer, and depend on the DOM structure which can easily change. Having your tests depend on your DOM structure can lead to flakey/difficult to maintain failing tests ([PW link to best practices](https://playwright.dev/docs/best-practices#prefer-user-facing-attributes-to-xpath-or-css-selectors))

### Suggested Solution (optional)

Playwright has the following APIs on `page`:

`getByAltText`, `getByLabel`, `getByPlaceholder`, `getByRole`, `getByTestId`, `getByText`, and `getByTitle`.

They are less prone to DOM structure changes, instead working with use facing attributes. Take a look at these two and you will see that one is easier to understand than the other as well as less prone to DOM structures (which can be very flakey when id and class names are auto generated by the underlying frontend framework):

```js
page.locator('button.buttonIcon.episode-actions-later');
```
vs

```js
page.getByRole('button', { name: 'submit' });
```

### Already existing or connected issues / PRs (optional)

https://github.com/grafana/xk6-browser/issues/784

Issue #4458, opened 2022-03-10
Separate Sobek and k6 logic from the module logic


inancgumus:
This can also mitigate some of the known bugs caused by the concurrent Sobek usage throughout the extension 🤞 

`Runtime` is most needed because of Sobek conversions. If we can provide a translation layer between Sobek and the extension code, we can greatly simplify it. For example:

```go
func (b *BrowserContext) WaitForEvent(event string, optsOrPredicate sobek.Value) interface{} {
	rt := k6common.GetRuntime(b.ctx)

	if optsOrPredicate != nil && !sobek.IsUndefined(optsOrPredicate) && !sobek.IsNull(optsOrPredicate) {
		switch optsOrPredicate.ExportType() {
		case reflect.TypeOf(sobek.Object{}):
			opts := optsOrPredicate.ToObject(rt)
			for _, k := range opts.Keys() {
				switch k {
				case "predicate":
					predicateFn, isCallable = sobek.AssertFunction(opts.Get(k))
					if !isCallable {
						k6Throw(b.ctx, "expected callable predicate")
					}
				case "timeout":
					timeout = time.Duration(opts.Get(k).ToInteger()) * time.Millisecond
				}
			}
```

This kind of cluttering happens because of internally exposing Sobek, whereas we could do:

```go
// Extension mechanics package (actual logic here)

package browser

func (c *Context) WaitForEvent(event string, predicate func()) (Event, error) {
	...
	return predicate(c.page)
}

// Translation layer (Sobek -> extension) — It's now named as the mapping layer.

type BrowserContext struct {
	...
	bc browser.Context
	...
}

func (b *BrowserContext) WaitForEvent(event string, optsOrPredicate sobek.Value) interface{} {
	// sobek stuff here
	r, err := b.bc.WaitForEvent(event predicate)
	if err != nil {
		k6Throw(...)
	}   
}
```

---

```
                      Test script
                           │
                           │
┌──────────────────────────▼────────────────────────────┐
│            Exposed methods to the test script         │
│          Extension translation layer: JS -> Sobek     │
└──────────────────────────┬────────────────────────────┘
                           │
┌──────────────────────────┴────────────────────────────┐
│                          ▼                            │
│                                                       │
│               Extension logic (Sobek-free)            │
│                                                       │
│                          │                            │
└──────────────────────────┼────────────────────────────┘
                           │
                           │
             ┌─────────────┼─────────────┐
             │             ▼             │
             │                           │
             │          Browser          │
[...]

PR #4843, opened 2025-06-09
Add getByRole in the browser module

ankur22:
## What?

This PR adds a long awaited feature which is `page.getByRole`. It makes working with selectors slightly simpler. More `getBy*` APIs are in the works.

## Why?

It is a lot simpler to work with `getByRole` instead of XPath or CSS selectors in most cases.

## Checklist

## Related PR(s)/Issue(s)

https://github.com/grafana/k6/issues/4764
https://github.com/grafana/k6/issues/4320
https://github.com/grafana/k6/issues/4248

### DIFF
--- a/internal/js/modules/k6/browser/common/frame.go
+++ b/internal/js/modules/k6/browser/common/frame.go
@@ -1009,7 +1009,7 @@ func (f *Frame) GetByRole(role string, opts *GetByRoleOptions) *Locator {
 	properties := make(map[string]string)
 
 	if opts == nil {
-		return f.Locator(fmt.Sprintf("role=%s", role), nil)
+		return f.Locator(fmt.Sprintf("internal:role=%s", role), nil)
 	}
 
 	if opts.Checked != nil {
@@ -1046,7 +1046,7 @@ func (f *Frame) GetByRole(role string, opts *GetByRoleOptions) *Locator {
 	}
 
 	var builder strings.Builder
-	builder.WriteString(fmt.Sprintf("role=%s", role))
+	builder.WriteString(fmt.Sprintf("internal:role=%s", role))
 	for key, value := range properties {
 		builder.WriteString(fmt.Sprintf("[%s=%s]", key, value))
 	}
--- a/internal/js/modules/k6/browser/common/frame.go
+++ b/internal/js/modules/k6/browser/common/frame.go
@@ -1002,6 +1002,58 @@ func (f *Frame) getAttribute(selector, name string, opts *FrameBaseOptions) (str
 	return s, true, nil
 }
 
+// Locator creates and returns a new locator for this frame.
+func (f *Frame) GetByRole(role string, opts *GetByRoleOptions) *Locator {
+	f.log.Debugf("Frame:Locator", "fid:%s furl:%q role:%q opts:%+v", f.ID(), f.URL(), role, opts)
+
+	properties := make(map[string]string)
+
+	if opts == nil {
+		return f.Locator(fmt.Sprintf("role=%s", role), nil)
+	}
+
+	if opts.Checked != nil {
+		properties["checked"] = fmt.Sprintf("%v", *opts.Checked)
+	}
+	if opts.Disabled != nil {
+		properties["disabled"] = fmt.Sprintf("%v", *opts.Disabled)
+	}
+	if opts.Selected != nil {
+		properties["selected"] = fmt.Sprintf("%v", *opts.Selected)
+	}
+	if opts.Expanded != nil {
+		properties["expanded"] = fmt.Sprintf("%v", *opts.Expanded)
+	}
+	if opts.IncludeHidden != nil {
+		properties["include-hidden"] = fmt.Sprintf("%v", *opts.IncludeHidden)
+	}
+	if opts.Level != nil {
+		properties["level"] = fmt.Sprintf("%v", *opts.Level)
+	}
+	if opts.Name != nil && *opts.Name != "" {
+		// Exact option can only be applied to quoted strings.
+		if (*opts.Name)[0] == '\'' && (*opts.Name)[len(*opts.Name)-1] == '\'' {
+			if opts.Exact != nil && *opts.Exact {
+				*opts.Name = fmt.Sprintf("%vs", *opts.Name)
+			} else {
+				*opts.Name = fmt.Sprintf("%vi", *opts.Name)
+			}
+		}
+		properties["name"] = *opts.Name
+	}
+	if opts.Pressed != nil {
+		properties["pressed"] = fmt.Sprintf("%v", *opts.Pressed)
+	}
+
+	var builder strings.Builder
+	builder.WriteString(fmt.Sprintf("role=%s", role))
+	for key, value := range properties {
+		builder.WriteString(fmt.Sprintf("[%s=%s]", key, value))
+	}
+
+	return f.Locator(builder.String(), nil)
+}
+
 // Referrer returns the referrer of the frame from the network manager
 // of the frame's session.
 // It's an internal method not to be exposed as a JS API.
--- a/internal/js/modules/k6/browser/common/frame.go
+++ b/internal/js/modules/k6/browser/common/frame.go
@@ -1002,9 +1002,10 @@ func (f *Frame) getAttribute(selector, name string, opts *FrameBaseOptions) (str
 	return s, true, nil
 }
 
-// Locator creates and returns a new locator for this frame.
+// GetByRole creates and returns a new locator for this frame using the ARIA role
+// and any additional options.
 func (f *Frame) GetByRole(role string, opts *GetByRoleOptions) *Locator {
-	f.log.Debugf("Frame:Locator", "fid:%s furl:%q role:%q opts:%+v", f.ID(), f.URL(), role, opts)
+	f.log.Debugf("Frame:GetByRole", "fid:%s furl:%q role:%q opts:%+v", f.ID(), f.URL(), role, opts)
 
 	properties := make(map[string]string)
 
--- a/internal/js/modules/k6/browser/common/frame.go
+++ b/internal/js/modules/k6/browser/common/frame.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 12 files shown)


internal/js/modules/k6/browser/common/js/injected_script.js
```diff
@@ -14,6 +14,1259 @@
  * limitations under the License.
```
ankur22, line 14: The main implementation is in the `injected_script.js` file. The code was directly copied over from the compiled JS code from Playwright. I have added as many integration tests as possible to ensure that all the possible flows are accounted for. While i understand the jist of what is happening in the copied code, some of it is still a bit of a mystery as to why things are done this way. A lot of the code seems to be inherited behaviour from specifications, such as not being able to work with `header` or `footer` unless they're not encapsulated -- exact spec is unknown though.

Do we need to add the Playwright license and notice file in the browser module?

internal/js/modules/k6/browser/common/js/injected_script.js
```diff
@@ -225,6 +1478,7 @@ class InjectedScript {
       css: new CSSQueryEngine(),
       text: new TextQueryEngine(),
       xpath: new XPathQueryEngine(),
+      'internal:role': createRoleEngine(true),
```
ankur22, line 1481: Sticking with how Playwright has labelled it. There is also a selector without `internal` which we could consider adding. It is so that users can query without the `getByRole` API, and directly do something like: `page.locator('role=button');`.

internal/js/modules/k6/browser/tests/get_by_test.go
```diff
@@ -0,0 +1,775 @@
+package tests
```
ankur22, line 4: Tests are failing on Windows for an unknown reason. I can run the same test in a script file and it works on Windows. I'm looking into it 😕

ankur22: WINDOWS! WHY ART THOU SO ANNOYING! I WANT TO LOVE YOU BUT I CAN'T 😭

I tried to run some other tests which work with `page.locator` and they don't work! These tests are in `keyboard_test.go`, and at the top of the file there is [this](https://github.com/grafana/k6/blob/85b609c9d3c523bc6a32e18f6fe3649abb15daef/internal/js/modules/k6/browser/tests/keyboard_test.go#L1-L2) comment:

```go
// practically none of this work on windows
//go:build !windows
```

I will do the same for the new tests 😢

internal/js/modules/k6/browser/common/frame.go
```diff
@@ -1002,6 +1002,58 @@ func (f *Frame) getAttribute(selector, name string, opts *FrameBaseOptions) (str
 	return s, true, nil
 }
 
+// Locator creates and returns a new locator for this frame.
```
joanlopez: This GoDoc sentence looks like a leftover from copy-pasting 🤔

internal/js/modules/k6/browser/common/frame.go
```diff
@@ -1002,6 +1002,58 @@ func (f *Frame) getAttribute(selector, name string, opts *FrameBaseOptions) (str
...
+func (f *Frame) GetByRole(role string, opts *GetByRoleOptions) *Locator {
+	f.log.Debugf("Frame:Locator", "fid:%s furl:%q role:%q opts:%+v", f.ID(), f.URL(), role, opts)
```
joanlopez: Should we refer to the method name instead, to keep it consistent with the other methods? 🤔 

```suggestion
	f.log.Debugf("Frame:GetByRole", "fid:%s furl:%q role:%q opts:%+v", f.ID(), f.URL(), role, opts)
```

internal/js/modules/k6/browser/common/frame.go
```diff
@@ -1002,6 +1002,58 @@ func (f *Frame) getAttribute(selector, name string, opts *FrameBaseOptions) (str
...
+		properties["disabled"] = fmt.Sprintf("%v", *opts.Disabled)
+	}
+	if opts.Selected != nil {
+		properties["selected"] = fmt.Sprintf("%v", *opts.Selected)
+	}
+	if opts.Expanded != nil {
+		properties["expanded"] = fmt.Sprintf("%v", *opts.Expanded)
+	}
+	if opts.IncludeHidden != nil {
+		properties["include-hidden"] = fmt.Sprintf("%v", *opts.IncludeHidden)
```
joanlopez: Why not using `strconv.FormatBool` instead of string formatting here? It's slightly more idiomatic, and perhaps even a bit faster, I guess. But maybe there's a reason not to.

PS: Same, whenever it applies, with the rest of the opts parsed on this method.

internal/js/modules/k6/browser/tests/get_by_test.go
```diff
@@ -0,0 +1,778 @@
...
+	return &s
+}
+
+func boolPtr(b bool) *bool {
+	return &b
+}
+
+func int64Ptr(i int64) *int64 {
+	return &i
+}
```
joanlopez: I think nowadays, thanks to generics, this can be simplified into a single generic method, like:

```go
func ptr[T any](v T) *T {
  return &v
}
```

internal/js/modules/k6/browser/common/frame.go
ankur22: Updated in f490d706646975071b6378b8e0c92399ae1ac172

internal/js/modules/k6/browser/common/frame.go
ankur22: Updated in f490d706646975071b6378b8e0c92399ae1ac172

internal/js/modules/k6/browser/common/frame.go
ankur22: Yep, good spot. Changed in a098ac38b2e0d6033080563bb4063d9374fd81bf

internal/js/modules/k6/browser/tests/get_by_test.go
ankur22: Thanks! Great idea. Updated in dfcd80f9cbf454a961833130a04ae3c4f8857d57

internal/js/modules/k6/browser/common/js/injected_script.js
inancgumus: I don't think we need to add another license note, as we already have this one:

https://github.com/grafana/k6/blob/113abf5fdd7bc5ddc36c87a10aa7febdf7a0bb71/internal/js/modules/k6/browser/common/js/injected_script.js#L1-L15

internal/js/modules/k6/browser/common/locator_options.go
```diff
@@ -0,0 +1,81 @@
...
+			val := obj.Get(k).ToBoolean()
+			o.Pressed = &val
+		case "selected":
+			val := obj.Get(k).ToBoolean()
+			o.Selected = &val
+		}
+	}
+
+	return nil
+}
```
inancgumus: For #4219 #4458, can you move this part to the mapping layer?

Example: https://github.com/grafana/xk6-browser/pull/1512/files.

internal/js/modules/k6/browser/tests/get_by_test.go
```diff
@@ -0,0 +1,770 @@
...
+
+	"github.com/grafana/sobek"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"go.k6.io/k6/internal/js/modules/k6/browser/common"
+)
+
+func ptr[T any](v T) *T {
+	return &v
+}
```
inancgumus: For consistency, I'd move `ptr` next to [these functions](https://github.com/grafana/k6/blob/113abf5fdd7bc5ddc36c87a10aa7febdf7a0bb71/internal/js/modules/k6/browser/tests/test_browser.go#L352-L380). Instead of `ptr`, `toPtr` or `ptrOf` could be more readable and understandable about what it does, but I have a weak opinion

inancgumus (commented):
I focused on the easy parts of this PR. But I'll continue tomorrow.

internal/js/modules/k6/browser/common/locator_options.go
ankur22: Done in 0802c85340990b5e45d2dd201a633b84d394073a, e270179019fd9424912219f5f17787edb4a80865 and a5f13dbfc67ee913ff20bcd943d3470471f82ee9

internal/js/modules/k6/browser/tests/get_by_test.go
ankur22: Sure, done in 728e95aefa75b17704027e8e96f87e592c7dde00

internal/js/modules/k6/browser/common/js/injected_script.js
AgnesToulet, line 1: Could we split this file into multiple files and try to follow the Playwright structure as much as possible (same file names)? I think this would help with reviewing and also maintaining if any changes happen in Playwright in the future.

internal/js/modules/k6/browser/tests/get_by_test.go
```diff
@@ -0,0 +1,766 @@
...
+					assert.Equal(t, tt.expectedText, text)
+				}
+			})
+		}
+	})
+
+	// This tests all the options, and different attributes (such as explicit
+	// aria attributes vs the text value of an element) that can be used in
+	// the DOM with the same role.
+	t.Run("edge_cases", func(t *testing.T) {
```
AgnesToulet, line 587: We are not testing the `exact` attribute here, I think it would be nice to add one or two test for this use case.

AgnesToulet (commented):
I played a bit with it and everything seems to work great :+1:  (plus the tests are covering a lot of use cases :rocket:)

My main concern is about `injected_script.js` that is already big and will only grow bigger and bigger with the upcoming PRs.

internal/js/modules/k6/browser/common/selectors.go
```diff
@@ -91,6 +91,9 @@ func (s *Selector) parse() error {
 			// If selector starts with '..', consider xpath as well.
 			name = "xpath"
 			body = part
+		case strings.HasPrefix(part, "internal:role="):
+			name = "internal:role"
+			body = part
```
AgnesToulet, line 96: Question for my own understanding: is it what makes the link between the Go `GetByRole` function and the script in `injected_script.js`? Or something else?

internal/js/modules/k6/browser/common/js/injected_script.js
ankur22: This JS version of PW's injected script is what is compiled from the type script injected scripts. I can see the benefit of splitting it if the files were easy to compare against PW, but it's something that seems to be available when installing it locally using node. Essentially, i feel it might be extra work to split the code into separate files, without the ability to be able to compare them against PW's changes, and so no real benefit.

If we split them and store them as separate files, we would also need to perform multiple requests to chrome to inject them one at a time during runtime, unless we had a job during CI/CD which combined them into a single file -- which is just another moving part, that could go wrong.

I toyed with the idea of copying the whole injected script from PW into k6 (not just take the parts we need to get the functionality working for the features we're implementing), so that it would make importing changes easier, but that didn't seem correct. At least with the selective approach has meant that I've been forced to understand what I've copied to help me get a good enough idea of what it is doing instead of no idea at all.

internal/js/modules/k6/browser/common/selectors.go
ankur22: That's correct. A selector is split in this function. When the selector is passed to the injected script (that is running in chrome) it works with the parsed selector to pick the correct [selector engine](https://github.com/grafana/k6/blob/poc/role-based-selector-engine/internal/js/modules/k6/browser/common/js/injected_script.js#L1478-L1481).

internal/js/modules/k6/browser/tests/get_by_test.go
ankur22: Good spot! Added in a18bbd46fc60d4f3bc4a32db30a20efc9a063817

internal/js/modules/k6/browser/common/js/injected_script.js
AgnesToulet: Ah, I looked only at the file name and didn't notice the extension was different. Thanks for the detailed answer. 

I think it could still be worth it if we copy a lot more because JS and TS are fairly easy to compare (looking at the domUtils.ts file for example). But for now, it's true that, as we copy only what we need, it makes it harder to compare. Also, if we decide to do it, it should be its own PR so that's fine for this PR :+1:

AgnesToulet (approved):
LGTM!

internal/js/modules/k6/browser/common/js/injected_script.js
joanlopez: Yeah, I think that would be nice, specially because I think `internal:role` is particular to Playwright, while the other looks more like "the standard"?

Is there any reason why not doing so on this PR? 
Does it require any changes beyond also adding that pair of keyword and engine? 🤔

ankur22: > Is there any reason why not doing so on this PR?

Mainly because I wanted to narrow the focus to working with the dedicated APIs that we have implemented `getByRole` instead of allowing for the manual creation of role selectors. It can get quite complicated to manually create a role selector with options. So it's an ease of use thing mainly.

> Does it require any changes beyond also adding that pair of keyword and engine? 🤔

That's basically it, add another case for `role` without `internal:`.

I'd like to see how this is used and whether users ask us to open up to a manual way of writing role selectors.

PR #4848, opened 2025-06-12
run CI/CD tests also on arm

nadiamoe:
This PR enables CI/CD pipelines to additionally run on arm machines, more precisely on `ubuntu-24.04-arm`. This runner is provided by Github to the general public in beta quality, and is documented here: https://github.com/actions/partner-runner-images/blob/main/images/arm-ubuntu-24-image.md

Running CI on arm is a first step towards increasing our confidence on the reliability of k6 and browser in this platform.

Some considerations for reviewers:
- Unlike `ubuntu-24.04`, `ubuntu-24.04-arm` does not come with chrome or chromium preinstalled, thus the reason of the conditional "install chromium" step.
- Browser e2e tests force the use of `google-chrome` binary, as opposed to chromium: https://github.com/grafana/k6/blob/b60a08336f9724e2683eb22f2026eec2e0f5a428/.github/workflows/browser_e2e.yml#L64. For arm, a modern chromium is used instead, as that's what is available.

### DIFF
--- a/.github/workflows/browser_e2e.yml
+++ b/.github/workflows/browser_e2e.yml
@@ -61,7 +61,7 @@ jobs:
       - name: Run E2E tests
         run: |
           set -x
-          if [ "$RUNNER_OS" == "Linux" ]; then
+          if [ "$RUNNER_OS" == "Linux" ] && [ "$RUNNER_ARCH" == "X64" ]; then
             export K6_BROWSER_EXECUTABLE_PATH=/usr/bin/google-chrome
           fi
           export K6_BROWSER_HEADLESS=true


nadiamoe:
Tired to use crocochrome for this in 99eca809701323d245dff19cff7f7c7e52c566da, but unfortunately that did not work: The test suite includes tests for the lifecycle of the built-in browser process manager, so that's not an option.

.github/workflows/test.yml
```diff
@@ -53,6 +59,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        # TODO: Add ubuntu-22.04-arm to this list after figuring out how to run browser on arm runners.
         platform: [ubuntu-22.04, windows-latest]
```
mstoykov:
```suggestion
        platform: [ubuntu-22.04, ubuntu-22.04-arm, windows-latest]
```
I think this was left over or was done before the gotip repo was updated.

.github/workflows/test.yml
```diff
@@ -53,6 +59,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        # TODO: Add ubuntu-22.04-arm to this list after figuring out how to run browser on arm runners.
```
joanlopez: Can't this be resolved now with the progress you made so far? 🤔

mstoykov (approved):
Thank you for the PR :bow: 

It is strange that 24.04 works with arm but not x86 ... I do wonder if we want to run on 22.04 in both cases :thinking: or not.

But we definitely want to also run this on gotip

.github/workflows/test.yml
nadiamoe: Yep, this was a leftover and should work now. Taking a look!

mstoykov (approved):

joanlopez (approved):

Issue #4861, opened 2025-06-18
Browser module doesn't respect navigation error text


ankur22:
### Brief summary

If we run the following test:

```js
import { browser } from 'k6/browser';

export const options = {
  scenarios: {
    ui: {
      executor: 'shared-iterations',
      options: {
        browser: {
          type: 'chromium',
        },
      },
    },
  },
};

export default async function () {
  const page = await browser.newPage();

  try {
    await page.goto("https://blah.madeupurl.com");
    console.log("test")
  } finally {
    await page.close();
  }
}
```

It should fail on the `page.goto` since the DNS cannot be resolved, and therefore we shouldn't see `"test"` being logged to the console.

At the moment we're not capturing the error text correctly [here](https://github.com/grafana/k6/blob/ffda43895a9789ce4a2d07cd90b7078df7f3f04e/internal/js/modules/k6/browser/common/frame_session.go#L624-L631). We should be checking to see if it is not empty and fail the navigation if it is not empty.

### k6 version

NA

### OS

NA

### Steps to reproduce the problem

NA

### Expected behaviour

Test should fail on `page.goto`.

### Actual behaviour

Test continues on assuming that the page loaded correctly.

PR #4862, opened 2025-06-18
Fix navigation when errorText is not empty

ankur22:
## What?

We now check whether `errorText` is not empty during a navigation, to fail the test early.

## Why?

When `errorText` is not empty and `err` is `nil`, it means that the navigation failed; for example when a DNS lookup fails.

## Checklist

## Related PR(s)/Issue(s)

https://github.com/grafana/k6/issues/4861

### DIFF
--- a/internal/js/modules/k6/browser/tests/page_test.go
+++ b/internal/js/modules/k6/browser/tests/page_test.go
@@ -2788,7 +2788,7 @@ func TestPageMustUseNativeJavaScriptObjects(t *testing.T) {
 	require.NoErrorf(t, err, "page should not override the native objects, but it did")
 }
 
-func TestNameNotResolved(t *testing.T) {
+func TestDNSNameNotResolved(t *testing.T) {
 	t.Parallel()
 
 	b := newTestBrowser(t)


internal/js/modules/k6/browser/common/frame_session.go
```diff
@@ -629,6 +629,9 @@ func (fs *FrameSession) navigateFrame(frame *Frame, url, referrer string) (strin
 			err = fmt.Errorf("%q: %w", errorText, err)
 		}
 	}
+	if errorText != "" {
+		err = fmt.Errorf("navigation failed: %v", errorText)
+	}
```
ankur22: In k6 browser, when a navigation fails due to something like a DNS lookup, at the moment it continues on with the navigation and seems to assume that it succeeded. This is a problem down the line since future calls to APIs in the test script will eventually fail.

Running all the browser tests, i realised it seems that the old behaviour is valid? For example [TestBlockHostname](https://github.com/grafana/k6/blob/7129195a85e57533564e46086018e43915d27945/internal/js/modules/k6/browser/tests/network_manager_test.go#L43-L71) now fails on `page.goto`, but the test expects the `page.goto` call to succeed. What do you think the behaviour should be?

CC @inancgumus

internal/js/modules/k6/browser/tests/page_test.go
```diff
@@ -2787,3 +2787,20 @@ func TestPageMustUseNativeJavaScriptObjects(t *testing.T) {
 	_, err = page.QueryAll("#textField")
 	require.NoErrorf(t, err, "page should not override the native objects, but it did")
 }
+
+func TestNameNotResolved(t *testing.T) {
```
codebien:
```suggestion
func TestDNSNameNotResolved(t *testing.T) {
```
It might be my low knowledge with the product, but DNS would help me to define the scope.

internal/js/modules/k6/browser/common/frame_session.go
```diff
@@ -629,7 +629,7 @@ func (fs *FrameSession) navigateFrame(frame *Frame, url, referrer string) (strin
 			err = fmt.Errorf("%q: %w", errorText, err)
 		}
 	}
-	return documentID.String(), err
+	return documentID.String(), errorText, err
```
codebien, line 632:
```suggestion
	return documentID.String(), err
```

Why do we need to return this error? Can't we throw an error here if we identify that the `errorText` is not empty?

In addition, a DNS error is generally an error on any system, why isn't the case here? Are we sure to not silencing the error somewhere else?

ankur22: Right, i agree that this is a smell. Happy to have any suggestions on how to improve this.

### Context

The issue in this case where `err` is `nil` and `errorText` is not empty is that once we return an error after calling [navigateFrame](https://github.com/grafana/k6/blob/d404e718230fd4f35d3b5ae24591b8757de7ca9b/internal/js/modules/k6/browser/common/frame_manager.go#L651), it will return an error back to the caller, but the event loop will contain older events which could be consumed by the next event handler of the same type.

k6 browser uses event loops in the background. We register an event handler [here](https://github.com/grafana/k6/blob/d404e718230fd4f35d3b5ae24591b8757de7ca9b/internal/js/modules/k6/browser/common/frame_manager.go#L603-L614) to listen out for incoming `EventFrameNavigation` CDP events from chromium once a page navigation occurs -- it turns out that this event is fired even when a DNS issue occurs, which makes sense since if you type in https://blah.madeupurl.com into chromium you will see `This site can’t be reached` page, so a navigation does indeed occur.

If we don't consume the event then the next time we create the same event handler it might consume the older event, which is what was happening in `TestBlockHostnames` when i ran the test when returning `err` early.

After an event is consumed the event handler is [destroyed](https://github.com/grafana/k6/blob/7129195a85e57533564e46086018e43915d27945/internal/js/modules/k6/browser/common/helpers.go#L162) regardless of whether it successfully handled the event or not.

To avoid this, the handling of `errorText` needs to be done differently to `err` when calling `navigateFrame` ([code link](https://github.com/grafana/k6/blob/d404e718230fd4f35d3b5ae24591b8757de7ca9b/internal/js/modules/k6/browser/common/frame_manager.go#L651-L654)). The error case is handled straight away since we assume that we will never get a `EventFrameNavigation` event. For the case where `errorText` is not empty, we need to wait for the `EventFrameNavigation`, and then we can return an error ([code link](https://github.com/grafana/k6/blob/d404e718230fd4f35d3b5ae24591b8757de7ca9b/internal/js/modules/k6/browser/common/frame_manager.go#L703-L708)).

Changing the event loop itself will be super complex -- we have tried to make fixes and refactors in the past and it's super difficult which can cause other issues further down the line! I want to avoid that.

codebien:
@ankur22 the latest added tests seems to fail

ankur22:
> @ankur22 the latest added tests seems to fail

🤷 it's one of those cases where it works on my machine and not in CI. I've tried working with GOMAXPROCS and i can't get the test to fail locally. I am seeing an unusaly high number of CI failures this week, could it be a change in githubs actions?

@inancgumus any pointers on how I could try to cox the new `TestClickSkipRequestForBlobFile` and `TestURLSkipRequestForBlobFile` to fail locally?

codebien:
>it's one of those cases where it works on my machine and not in CI.

@ankur22 Did you try to run with `-race` enabled?

> could it be a change in githubs actions?

I don't think so, we should see it across all the modules. Instead, they seem localized on the browser.

Honestly, based on all the `Please call browser.close only once` reported, it smells like there is a data race somewhere.

ankur22:
> @ankur22 Did you try to run with -race enabled?

yeah i used that too.

> I don't think so, we should see it across all the modules. Instead, they seem localized on the browser.

The browser module is a little more sensitive to github action changes since the tests require significantly more resource than the other modules to run a single test with a browser.

> Honestly, based on all the Please call browser.close only once reported, it smells like there is a data race somewhere.

That's nothing new (unfortunately), neither are the context canceled issues which end the test early.

internal/js/modules/k6/browser/common/frame_session.go
inancgumus: Can you help me understand your concerns? How can I run and compare the "new" and "old" behavior? Do you mean the code in this PR by the "new" behavior? Thanks!

inancgumus:
> could it be a change in githubs actions

@ankur22 Do you mean the current (this PR's) failures? The Github actions should not affect the outcome of these runs unless you see a Vault communication error.

> @inancgumus any pointers on how I could try to cox the new TestClickSkipRequestForBlobFile and TestURLSkipRequestForBlobFile to fail locally?

It's probably a Windows issue because the preceding errors (before your tests) are often like this:

```
time="2025-06-24T16:38:47Z" level=error msg="process with PID 4716 unexpectedly ended: exit status 1" category=browser elapsed="0 ms" source=browser
time="2025-06-24T16:38:47Z" level=error msg="cleaning up the user data directory: unlinkat C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\k6browser-data-2454286471\\Default\\Account Web Data: The process cannot access the file because it is being used by another process." category="Browser:Close" elapsed="42 ms" source=browser
time="2025-06-24T16:38:52Z" level=error msg="error on evaluating window.k6SpanId: evaluating JS in global context: context canceled" category="FrameSession:onFrameNavigated" elapsed="0 ms" source=browser
```

I can get the same test error when I give it a file other than `blob.html`, like `foo.html`.

internal/js/modules/k6/browser/common/frame_session.go
joanlopez: Have you seen #4918? It seems to be related 🤔

ankur22:
Closing this PR in favour of https://github.com/grafana/k6/pull/4918

PR #4879, opened 2025-06-25
k6packager: Adjust versions of dependencies (s3cmd, aws)

joanlopez:
**After a quick chat with @mstoykov, and as he suggested, I tried to summarize the context of this pull request in the following text:**

*This PR addresses issues related to inconsistent version definitions used across the docker-compose, GitHub workflow, and Dockerfile during the release process.*

*Originally, fallback versions in docker-compose.yml were updated due to failures with outdated versions (thanks to @inancgumus). However, the corresponding versions in the GitHub workflow were not updated, causing the workflow to fail during the v1.1.0 release.*

*To resolve this, the fallback versions were updated in the workflow file, and the workflow was manually triggered from the branch to generate the k6packager release—successfully completing the release process. The resulting PR remained open and unmerged.*

*To prevent similar issues in the future, and as suggested by both @inancgumus and @codebien, the version definitions have now been centralized by hardcoding them in the Dockerfile. This eliminates the need to define versions in multiple places (e.g., docker-compose.yml), reducing the risk of inconsistencies and enabling potential use of Dependabot for automated version updates.*

*The removed versions in docker-compose.yml match those now defined in the Dockerfile. 
No new versions are being introduced—only relocated.*

## What?

It removes the versions from both, the GitHub Workflow definition and the docker-compose.yml, in favor of hard-coding them in the Dockerfile. It doesn't really update the versions. The versions referred in this pull request were already used for v1.1.0 release.

## Why?

~Because I think this should have been updated as part of https://github.com/grafana/k6/pull/4833/commits/3667631e691936f8223060e5ff5bbfb87832ac4d, because these refs are the actual source in most cases (unless you run `docker compose build`), so not updating them means that both manually triggered and schedule `k6packager` builds are still using the old version.~

Because having the versions defined in a couple of places makes it more prone to update them in a single place. So, this way, we have a SSOT, plus we can enable Dependabot for automated version updates.


inancgumus (dismissed):
LGTM 👍

Having two duplicate versions (e.g., s3cmd) is error-prone (that's likely why I forgot to update these versions in #4833).

So, I suggest:

- Having a single source of truth for these versions only in `packager.yml`.
- Removing the default versions (e.g., `S3CMD_VERSION`) from `packaging/docker-compose.yml`.

I can send a PR to do that, or would you like to do it in this PR? I'm fine both ways.

inancgumus:
@codebien As long as there is a single truth instead of scattered versions, any solution sounds good to me.

packaging/Dockerfile
```diff
@@ -7,18 +7,16 @@ ENV DEBIAN_FRONTEND=noninteractive
 RUN apt-get update -y && \
     apt-get install -y apt-utils createrepo curl git gnupg2 python3-pip unzip
 
-ARG S3CMD_VERSION
-RUN pip3 install "s3cmd${S3CMD_VERSION:+==$S3CMD_VERSION}"
+RUN pip3 install "s3cmd==2.4.0"
```
codebien, line 13:
```suggestion
RUN pip3 install "s3cmd==2.4.*"
```
Should we relax the condition here?

packaging/Dockerfile
```diff
@@ -4,21 +4,22 @@ LABEL maintainer="k6 Developers <developers@k6.io>"
...
 
-COPY ./awscli-key.gpg .
-ARG AWSCLI_VERSION
 # Download awscli, check GPG signature and install.
+COPY ./awscli-key.gpg .
 RUN export GNUPGHOME="$(mktemp -d)" && \
     gpg2 --import ./awscli-key.gpg && \
     fpr="$(gpg2 --with-colons --fingerprint aws-cli | grep '^fpr' | cut -d: -f10)" && \
```
codebien, line 19: Isn't better if we move this to a package manager (e.g. pip or apt) so we can set a version condition?

packaging/Dockerfile
joanlopez: We can do it, but then we would wave slightly less reproducible builds. But if you want, I can do, of course.

In this case I think it should be something like: `s3cmd~=2.4.0`, which I think means:
- _Any `s3cmd` version >=`2.4.0` and <`2.5.0`_

packaging/Dockerfile
joanlopez: I'm not sure, because [AWS CLI install and update instructions](https://docs.aws.amazon.com/cli/latest/userguide/getting-started-install.html) suggest to follow these instructions.

Also, I'd argue this is beyond the scope of this PR.

packaging/Dockerfile
```diff
@@ -4,21 +4,22 @@ LABEL maintainer="k6 Developers <developers@k6.io>"
 
 ENV DEBIAN_FRONTEND=noninteractive
 
-RUN apt-get update -y && \
+RUN sed -i 's|http://deb.debian.org/debian|http://archive.debian.org/debian|g' /etc/apt/sources.list && \
```
joanlopez, line 7: I had to add these lines to point to Debian Archive, as it's no longer available because EOL.
See it worked [here](https://github.com/grafana/k6/actions/runs/16315429361/job/46080104458), while the most recent builds from `master` failed because of that.

The next step would be to upgrade to a newer version of Debian, which I tried, but it isn't that easy because newer releases don't have one of the dependencies we use (`createrepo`), and so the upgrade process will require more research.

So, I'd suggest to leave it as-is, with these lines, so it keeps working and don't breaks/blocks the next release. And I can create an issue to research how to upgrade for the next cycle. Wdyt? 🤔

mstoykov (approved):
~Have you tested that all the options at least still work?~

Thsi was addressed in the PR summary

Issue #4788, opened 2025-05-15
Implement page.getByAltText(selectorText: string, options?):


ankur22:
### Feature Description

This feature would allow k6 Browser scripts to locate elements—typically images, input fields, or other media elements—that include a specific `alt` attribute value. By mirroring Playwright’s `getByAltText`, users could write more accessible, intention-revealing tests (for example, `const logo = page.getByAltText('Company logo')`). Under the hood, k6 would query `img[alt="…"], area[alt="…"]`, and any element with an `aria-label` matching the text, falling back to substring or regex matching based on options. Implementing this API will greatly simplify scripts that validate visual or accessibility-related aspects of web pages and ensure consistent behavior across different test suites.

Playwright docs: https://playwright.dev/docs/api/class-page#page-get-by-alt-text

Note: this is only to be implemented for the `page`.

This should work once the implementation is complete:

```js
  await page.setContent(`
    <html><body>
      <img alt="Company logo" src="logo.png" />
    </body></html>
  `);
  const logo = page.getByAltText('Company logo');
```

### Already existing or connected issues / PRs (optional)

https://github.com/grafana/k6/issues/4248

PR #4881, opened 2025-06-25
Add getByAltText in the browser module

ankur22:
## What?

This adds a convenience method on `page` to be able to select on elements with the `alt` attribute.

## Why?

This is part of the story of adding `getBy*`, which makes working with selectors a little easier.

## Checklist

## Related PR(s)/Issue(s)

Updates: https://github.com/grafana/k6/issues/4788, https://github.com/grafana/k6/issues/4248


internal/js/modules/k6/browser/common/js/injected_script.js
```diff
@@ -1266,31 +1291,502 @@ function createRoleEngine(internal) {
...
-  iframe.style.left = '-9999px';
-  iframe.style.display = 'none';
-
-  // grab the native browser object
-  document.documentElement.appendChild(iframe);
-  const win = iframe.contentWindow;
-  document.documentElement.removeChild(iframe);
-
+// packages/injected/src/selectorEvaluator.ts
+var SelectorEvaluatorImpl = class {
```
AgnesToulet, line 1298: Do you think we will be using more from this class in the future? Because, from what I understood from the current implementation, we only use `_queryCSS` and could remove most of this class' functions for the `getByAltText` feature

internal/js/modules/k6/browser/common/frame.go
```diff
@@ -1056,6 +1056,22 @@ func (f *Frame) GetByRole(role string, opts *GetByRoleOptions) *Locator {
 	return f.Locator(builder.String(), nil)
 }
 
+// Locator creates and returns a new locator for this frame.
+func (f *Frame) GetByAltText(alt string, opts *GetByAltTextOptions) *Locator {
+	f.log.Debugf("Frame:GetByAltText", "fid:%s furl:%q alt:%q opts:%+v", f.ID(), f.URL(), alt, opts)
+
+	a := "[alt=" + alt + "]"
+	if len(alt) > 0 && alt[0] == '\'' && alt[len(alt)-1] == '\'' {
```
AgnesToulet: Would be nice to put this line in a `isRegex` function. It would be clearer (because it's not obvious that strings are enclosed in `'` while regex are not) and I could also use it in my PR for `page.route` :smile:

internal/js/modules/k6/browser/common/js/injected_script.js
ankur22: Good catch! I forgot to clean this up. However, now I'll do that in a new PR so that I take into account the requirements for all the other `getBy*` APIs.

ankur22: I created an issue so that it's not forgotten again: https://github.com/grafana/k6/issues/4933

internal/js/modules/k6/browser/common/frame.go
ankur22: Sure, i've changed that in 7ea1dd4f5a6d7627f39e82a45d6a2f71aa858e35.

internal/js/modules/k6/browser/common/frame.go
```diff
@@ -1056,6 +1056,29 @@ func (f *Frame) GetByRole(role string, opts *GetByRoleOptions) *Locator {
 	return f.Locator(builder.String(), nil)
 }
 
+// isExactString returns true if the string is a quoted string.
+// This is used to determine if the string will be used to match
+// an exact string instead of as a regex in the getBy* APIs.
+func isExactString(s string) bool {
+	return len(s) > 0 && s[0] == '\'' && s[len(s)-1] == '\''
+}
```
AgnesToulet: Nit: `isExactString` is a bit confusing with the `exact` option that is used later, here are some alternatives:
```suggestion
func isStringMatcher(s string) bool {
	return len(s) > 0 && s[0] == '\'' && s[len(s)-1] == '\''
}
```
```suggestion
func isText(s string) bool {
	return len(s) > 0 && s[0] == '\'' && s[len(s)-1] == '\''
}
```

Another option would be to use `isRegex` and reverse the condition. The name of the function would be much more clear but the condition would be less readable:
```suggestion
func isRegex(s string) bool {
	return len(s) == 0 || s[0] != '\'' || s[len(s)-1] != '\''
}
```

What do you think?

ankur22: How about `isQuotedText`?

AgnesToulet: Sounds good :+1:

ankur22: Rebased and push to 964ff708e98a23cfd23a08297d2424b6a5b30a09

AgnesToulet (dismissed):

AgnesToulet (approved):

internal/js/modules/k6/browser/common/frame.go
```diff
@@ -1084,6 +1080,29 @@ func (f *Frame) GetByRole(role string, opts *GetByRoleOptions) *Locator {
 	return f.Locator(builder.String(), nil)
 }
 
+// isQuotedText returns true if the string is a quoted string.
+// This is used to determine if the string will be used to match
+// a string instead of as a regex in the getBy* APIs.
+func isQuotedText(s string) bool {
+	return len(s) > 1 && s[0] == '\'' && s[len(s)-1] == '\''
+}
```
inancgumus, line 1088: This might fail if Unicode characters are passed. See: https://go.dev/blog/strings.

internal/js/modules/k6/browser/common/js/injected_script.js
```diff
@@ -14,6 +14,31 @@
...
+  const win = iframe.contentWindow;
+  document.documentElement.removeChild(iframe);
+
+  return {
+    Set: win.Set,
+    Map: win.Map,
+    // Add other native browser objects as needed.
+  }
+})();
+
```
inancgumus, line 41: Not sure why this seems like a new code addition, as it was from a previous PR?

inancgumus (approved):
Nice work 👍

internal/js/modules/k6/browser/common/frame.go
inancgumus: I'd suggest: `isQuotedString`. But `isQuotedText` is also fine.

internal/js/modules/k6/browser/common/js/injected_script.js
ankur22: I had to move it up in the file so that the injected script code from Playwright could work with the `k6BrowserNative` `set` and `map`. It was purely a move up in the file, this code remains the same as it was before.

internal/js/modules/k6/browser/common/frame.go
ankur22: I will leave it as is for now 👍

internal/js/modules/k6/browser/common/frame.go
ankur22: I have added this to another [issue](https://github.com/grafana/k6/issues/4964) to look into once all the other `getBy*` APIs have been merged in.

internal/js/modules/k6/browser/common/frame.go
inancgumus: That's fine.

internal/js/modules/k6/browser/common/frame.go
ankur22: I took a look at this and realised that since we're really only concerned at looking for the `'` and `"` which are ASCII, then the indexing should be fine, since in UTF-8 these are also still one byte. Let me know if i missed anything though.

inancgumus: I think you must be right. What about smart quotes?

```
fmt.Println(isQuotedText(`“smart quotes”`)) // false
```

ankur22: That's correct, it's regarded as not valid quoted text. The input of an API (e.g. `getByAltText`) will [parse the input](https://github.com/grafana/k6/blob/7f46faac0923d779c058a6acb8db80cbc63ef9a4/internal/js/modules/k6/browser/browser/page_mapping.go#L835-L852). In some cases we need the string to be wrapped with `"` and in other cases we need it to be wrapped with `'` -- this inconsistency arises due to the injected scripts copied over from Playwright. So we will only regard a string to be quoted if it has `"` or `'`, and anything else is regarded as not quoted.

Issue #4789, opened 2025-05-15
getByLabel(labelText: string, options?)


ankur22:
### Feature Description

Implementing `getByLabel` in k6 Browser will let users select form controls by their associated `<label>` text, reducing brittle selectors and improving readability. For example, `page.getByLabel('Email address')` would internally locate the `<label>` whose text matches “Email address” (allowing `for` options like exact or substring matching) and then return its associated form control via the for attribute or by wrapping. This API aligns with how users think about forms—by label—rather than low-level CSS or XPath selectors, making scripts more maintainable and accessible-focused.

### Already existing or connected issues / PRs (optional)

https://github.com/grafana/k6/issues/4248

PR #4890, opened 2025-06-26
Add getByLabel in the browser module

ankur22:
## What?

This adds a convenience method on `page` to be able to select on `label` elements or elements with the `aria-label` attribute.

## Why?

This is part of the story of adding `getBy*`, which makes working with selectors a little easier. We used to have to work with the `page.locator` API providing either a `CSS` selector (when the `aria-label` attribute was used) or a `XPath` selector if we wanted to get the element by the `label`:

```html
  <label for="password-input">Password</label>
  <input id="password-input">
```

```js
// Not possible in CSS
const l = page.locator('//label[text()="Password"]'); // XPath
```

Now we can use:

```js
const l = page.getByLabel('Password');
```

## Checklist

## Related PR(s)/Issue(s)

Updates: https://github.com/grafana/k6/issues/4789, https://github.com/grafana/k6/issues/4248

### DIFF
--- a/internal/js/modules/k6/browser/browser/page_mapping.go
+++ b/internal/js/modules/k6/browser/browser/page_mapping.go
@@ -171,7 +171,7 @@ func mapPage(vu moduleVU, p *common.Page) mapping { //nolint:gocognit,cyclop
 			return rt.ToValue(ml).ToObject(rt), nil
 		},
 		"getByAltText": func(alt sobek.Value, opts sobek.Value) (*sobek.Object, error) {
-			palt, popts := parseGetByAltTextOptions(vu.Context(), alt, opts)
+			palt, popts := parseGetByBaseOptions(vu.Context(), alt, opts)
 
 			ml := mapLocator(vu, p.GetByAltText(palt, popts))
 			return rt.ToValue(ml).ToObject(rt), nil
@@ -808,20 +808,20 @@ func parseGetByRoleOptions(ctx context.Context, opts sobek.Value) *common.GetByR
 	return o
 }
 
-// parseGetByAltTextOptions parses the GetByAltText alt input value and the
-// options from the Sobek.Value.
-func parseGetByAltTextOptions(
+// parseGetByBaseOptions parses the options for the GetBy* APIs and the input
+// text/regex.
+func parseGetByBaseOptions(
 	ctx context.Context,
-	alt sobek.Value,
+	input sobek.Value,
 	opts sobek.Value,
-) (string, *common.GetByAltTextOptions) {
-	a := parseStringOrRegex(alt)
+) (string, *common.GetByBaseOptions) {
+	a := parseStringOrRegex(input)
 
 	if !sobekValueExists(opts) {
 		return a, nil
 	}
 
-	o := &common.GetByAltTextOptions{}
+	o := &common.GetByBaseOptions{}
 
 	rt := k6ext.Runtime(ctx)
 
--- a/internal/js/modules/k6/browser/browser/page_mapping.go
+++ b/internal/js/modules/k6/browser/browser/page_mapping.go
@@ -176,6 +176,14 @@ func mapPage(vu moduleVU, p *common.Page) mapping { //nolint:gocognit,cyclop
 			ml := mapLocator(vu, p.GetByAltText(palt, popts))
 			return rt.ToValue(ml).ToObject(rt), nil
 		},
+		"getByLabel": func(label sobek.Value, opts sobek.Value) (*sobek.Object, error) {
+			plabel, popts := parseGetByBaseOptions(vu.Context(), label, true, opts)
+
+			fmt.Println("plabel", plabel)
+
+			ml := mapLocator(vu, p.GetByLabel(plabel, popts))
+			return rt.ToValue(ml).ToObject(rt), nil
+		},
 		"goto": func(url string, opts sobek.Value) (*sobek.Promise, error) {
 			gopts := common.NewFrameGotoOptions(
 				p.Referrer(),
--- a/internal/js/modules/k6/browser/browser/page_mapping.go
+++ b/internal/js/modules/k6/browser/browser/page_mapping.go
@@ -179,8 +179,6 @@ func mapPage(vu moduleVU, p *common.Page) mapping { //nolint:gocognit,cyclop
 		"getByLabel": func(label sobek.Value, opts sobek.Value) (*sobek.Object, error) {
 			plabel, popts := parseGetByBaseOptions(vu.Context(), label, true, opts)
 
-			fmt.Println("plabel", plabel)
-
 			ml := mapLocator(vu, p.GetByLabel(plabel, popts))
 			return rt.ToValue(ml).ToObject(rt), nil
 		},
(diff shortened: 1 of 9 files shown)


internal/js/modules/k6/browser/browser/page_mapping.go
```diff
@@ -171,11 +171,19 @@ func mapPage(vu moduleVU, p *common.Page) mapping { //nolint:gocognit,cyclop
...
-			palt, popts := parseGetByAltTextOptions(vu.Context(), alt, opts)
+			palt, popts := parseGetByBaseOptions(vu.Context(), alt, false, opts)
 
 			ml := mapLocator(vu, p.GetByAltText(palt, popts))
 			return rt.ToValue(ml).ToObject(rt), nil
 		},
+		"getByLabel": func(label sobek.Value, opts sobek.Value) (*sobek.Object, error) {
+			plabel, popts := parseGetByBaseOptions(vu.Context(), label, true, opts)
+
+			fmt.Println("plabel", plabel)
```
AgnesToulet:
```suggestion
```
I guess this was forgotten

internal/js/modules/k6/browser/browser/page_mapping.go
```diff
@@ -749,11 +757,15 @@ func parseWaitForFunctionArgs(
 	return js, popts, exportArgs(gargs), nil
 }
 
-func parseStringOrRegex(v sobek.Value) string {
+func parseStringOrRegex(v sobek.Value, doubleQuote bool) string {
```
AgnesToulet, line 765: Could you add a comment to explain use cases for double quote and those for single quote?

internal/js/modules/k6/browser/browser/page_mapping.go
```diff
@@ -749,11 +757,15 @@ func parseWaitForFunctionArgs(
...
 	var a string
 	switch v.ExportType() {
 	case reflect.TypeOf(string("")):
-		a = fmt.Sprintf("'%s'", v.String()) // Strings require quotes
+		if doubleQuote {
+			a = fmt.Sprintf(`"%s"`, v.String()) // Strings require quotes
+		} else {
+			a = fmt.Sprintf("'%s'", v.String()) // Strings require quotes
+		}
```
AgnesToulet: Nit
```suggestion
		// Strings require quotes
		if doubleQuote {
			a = fmt.Sprintf(`"%s"`, v.String())
		} else {
			a = fmt.Sprintf("'%s'", v.String())
		}
```

internal/js/modules/k6/browser/tests/static/get_by_label.html
```diff
@@ -0,0 +1,11 @@
+<!DOCTYPE html>
+<head>
+  <title>Label Selector Engine Test Suite</title>
+</head>
+<body>
+  <input aria-label="Username">
+  <label for="password-input">Password</label>
+  <input id="password-input">
+  <input aria-label="abc123">
```
AgnesToulet, line 9:
```suggestion
  <input aria-label="abc123">
  <span
    role="checkbox"
    aria-checked="false"
    tabindex="0"
    aria-labelledby="tac"></span>
  <span id="tac">I agree to the Terms and Conditions.</span>
```
They mention the `aria-labelledby` attribute in [Playwright doc](https://playwright.dev/docs/api/class-locator#locator-get-by-label) so here's an example from [the mdn doc](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-labelledby) to add a test for it. I've tested it and it's working fine so I'm OK if you prefer to add it in a following PR.

internal/js/modules/k6/browser/browser/page_mapping.go
ankur22: Good spot, removed in 92d686ec922fb09639f039ee5d10d2a5d22b9c33

internal/js/modules/k6/browser/browser/page_mapping.go
ankur22: Sure, added in 92d686ec922fb09639f039ee5d10d2a5d22b9c33

internal/js/modules/k6/browser/browser/page_mapping.go
ankur22: Done in 92d686ec922fb09639f039ee5d10d2a5d22b9c33

internal/js/modules/k6/browser/tests/static/get_by_label.html
ankur22: I'll add this to a new PR, thanks for spotting this!

ankur22: Added an issue to track this: https://github.com/grafana/k6/issues/4956

internal/js/modules/k6/browser/browser/page_mapping.go
```diff
@@ -749,11 +755,22 @@ func parseWaitForFunctionArgs(
...
+//
+// I would prefer not to change the copied injected script code from Playwright
+// so that it is easier to copy over updates/fixes from Playwright when we need
+// to.
+func parseStringOrRegex(v sobek.Value, doubleQuote bool) string {
 	var a string
 	switch v.ExportType() {
-	case reflect.TypeOf(string("")):
-		a = fmt.Sprintf("'%s'", v.String()) // Strings require quotes
+	case reflect.TypeOf(string("")): // text values require quotes
```
codebien, line 768: Should we have a constant here? Or do we expect the compiler to optimize?

internal/js/modules/k6/browser/browser/page_mapping.go
```diff
@@ -749,11 +755,22 @@ func parseWaitForFunctionArgs(
...
+		if doubleQuote {
+			a = fmt.Sprintf(`"%s"`, v.String())
```
codebien, line 770:
```suggestion
			a = `"`+ v.String())+`"`
```
This should be more efficient

internal/js/modules/k6/browser/common/frame.go
```diff
@@ -1082,13 +1082,20 @@ func (f *Frame) GetByRole(role string, opts *GetByRoleOptions) *Locator {
 
 // isQuotedText returns true if the string is a quoted string.
 // This is used to determine if the string will be used to match
-// a string instead of as a regex in the getBy* APIs.
+// a string instead of as a regex in the getBy* APIs. It handles
+// both single and double quotes.
 func isQuotedText(s string) bool {
-	return len(s) > 1 && s[0] == '\'' && s[len(s)-1] == '\''
+	switch {
```
codebien, line 1088: It seems you could simplify by removing len(s) > 0 and have a `if` statement in line outside with `len(s) < return false`

codebien (approved):

internal/js/modules/k6/browser/browser/page_mapping.go
ankur22: I didn't think about that tbh. I would expect the compiler to optimize it. I can make it a const though. I've added it to the catch all [issue](https://github.com/grafana/k6/issues/4964) which I will work on after all the `getBy*` PRs have been merged.

internal/js/modules/k6/browser/browser/page_mapping.go
ankur22: I've added it to the catch all [issue](https://github.com/grafana/k6/issues/4964) which I will work on after all the getBy* PRs have been merged.

internal/js/modules/k6/browser/common/frame.go
ankur22: You mean like:

```go
	if len(s) < 1 {
		return false
	}

	switch {
	case s[0] == '\'' && s[len(s)-1] == '\'':
		return true
	case s[0] == '"' && s[len(s)-1] == '"':
		return true
	}
	return false
```

If so, i've added it to the catch all [issue](https://github.com/grafana/k6/issues/4964) to work on later 👍

codebien (approved):

Issue #4793, opened 2025-05-15
Implement page.getByTitle(titleText: string, options?)


ankur22:
### Feature Description

Implementing `getByTitle` enables selecting elements via their `title` attribute—helpful for tooltips, icons, or elements with semantic hints. For instance, `page.getByTitle('Settings')` would locate `<button title="Settings">`. The matcher should allow different matching strategies (exact, substring, regex) and consider elements that expose title via ARIA (`aria-label`). Adding this API completes the suite of attribute-based selectors, giving k6 Browser users a consistent, Playwright-inspired interface for DOM querying.

Playwright docs: https://playwright.dev/docs/api/class-page#page-get-by-title

Note: this is only to be implemented on the `page`.

This should work after the implementation work:

```js
  await page.setContent(`
    <html><body>
      <button title="Settings">⚙️</button>
      <div id="prefs" hidden>Preferences</div>
      <script>
        document.querySelector('button').addEventListener('click', () => {
          const prefs = document.getElementById('prefs');
          prefs.hidden = !prefs.hidden;
        });
      </script>
    </body></html>
  `);
  const settingsBtn = page.getByTitle('Settings');
  // click to toggle preferences
  await settingsBtn.click();
  const prefs = page.getByText('Preferences');
```

### Already existing or connected issues / PRs (optional)

https://github.com/grafana/k6/issues/4248

PR #4910, opened 2025-07-09
Add getByTitle in the browser module

ankur22:
## What?

This adds a convenience method on `page` to be able to select on elements with the `title` attribute.

## Why?

This is part of the story of adding `getBy*`, which makes working with selectors a little easier. We used to have to work with the `page.locator` API providing either a `CSS` selector or a `XPath` selector if we wanted to get the element by the `title` attribute on an element:

```html
<div title="Information box">This is some information</div>
```

```js
const l = page.locator('div[title="Information box"]'); // CSS
const l = page.locator('//div[@title="Information box"]'); // XPath
```

Now we can use:

```js
const l = page.getByTitle('Information box');
```

## Checklist

## Related PR(s)/Issue(s)

https://github.com/grafana/k6/issues/4793, https://github.com/grafana/k6/issues/4248

### DIFF
--- a/internal/js/modules/k6/browser/tests/get_by_test.go
+++ b/internal/js/modules/k6/browser/tests/get_by_test.go
@@ -1011,3 +1011,91 @@ func TestGetByPlaceholderSuccess(t *testing.T) {
 		})
 	}
 }
+
+func TestGetByTitleSuccess(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name     string
+		title    string
+		opts     *common.GetByBaseOptions
+		expected int
+	}{
+		{
+			"missing_title",
+			"",
+			nil,
+			0,
+		},
+		{
+			// matches on all the elements with a title attribute.
+			"empty_string",
+			"''",
+			nil,
+			5,
+		},
+		{
+			"no_options",
+			"'Click me'",
+			nil,
+			1,
+		},
+		{
+			"exact_match",
+			"'Link to somewhere'",
+			&common.GetByBaseOptions{Exact: toPtr(true)},
+			1,
+		},
+		{
+			"no_exact_match",
+			"'link to somewhere'",
+			&common.GetByBaseOptions{Exact: toPtr(true)},
+			0,
+		},
+		{
+			"case_insensitive_match",
+			"'link to somewhere'",
+			nil,
+			1,
+		},
+		{
+			"regex_match",
+			`/^[a-z0-9]+$/`,
+			nil,
+			1,
+		},
+		{
+			"image_title",
+			"'Placeholder image'",
+			nil,
+			1,
+		},
+		{
+			"div_title",
+			"'Information box'",
+			nil,
+			1,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+
+			tb := newTestBrowser(t, withFileServer())
+			p := tb.NewPage(nil)
+			opts := &common.FrameGotoOptions{
+				Timeout: common.DefaultTimeout,
+			}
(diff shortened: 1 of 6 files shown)


internal/js/modules/k6/browser/tests/static/get_by_title.html
codebien, line 1: Out of curiosity, is there a specific reason why we use `static` instead of the idiomatic `testdata` folder?

internal/js/modules/k6/browser/tests/get_by_test.go
```diff
@@ -1011,3 +1011,91 @@ func TestGetByPlaceholderSuccess(t *testing.T) {
...
+			}
+			_, err := p.Goto(
+				tb.staticURL("get_by_title.html"),
+				opts,
+			)
+			require.NoError(t, err)
+
+			l := p.GetByTitle(tt.title, tt.opts)
+			c, err := l.Count()
+			assert.NoError(t, err)
```
codebien, line 1097:
```suggestion
			require.NoError(t, err)
```

codebien (commented):
@ankur22 I don't see any concurrency primitive. Is this API synchronous?

ankur22:
> @ankur22 I don't see any concurrency primitive. Is this API synchronous?

That's correct, it is a synchronous method. When we call `page.waitForTitle` it builds a locator object and returns it to the caller. This is done without performing any long running IO (e.g. CDP requests to chromium) and is a very quick process.

However, when we call a method on the locator object (e.g. `l.click()`) that is when it performs the IO ops against chromium to perform the action. These methods are asynchronous.

internal/js/modules/k6/browser/tests/static/get_by_title.html
ankur22: No idea, it's been like that from the beginning as far as I can tell. Are there any advantages to working with `testData` over `static`?

internal/js/modules/k6/browser/tests/get_by_test.go
ankur22: Is this to force the test to stop if the `assert` fails?

internal/js/modules/k6/browser/tests/static/get_by_title.html
codebien: From `go help`:
>Directory and file names that begin with "." or "_" are ignored by the go tool, as are directories named "testdata".

AFAIK, this means `static` folder might be included in the built binary.

internal/js/modules/k6/browser/tests/get_by_test.go
codebien: It is, I don't see the reason to continue running tests if they fail in a place where we don't expect.

internal/js/modules/k6/browser/tests/static/get_by_title.html
ankur22: I will add this to the catch all [issue](https://github.com/grafana/k6/issues/4964) to explore once the `getBy*` PRs are all merged in 👍

internal/js/modules/k6/browser/tests/get_by_test.go
ankur22: I will review all the `assert`s and make sure they are `require` where needed in the new tests. I have added an action point in the catch all [issue](https://github.com/grafana/k6/issues/4964) to work on after all the `getBy*` APIs have been merged in.

codebien (approved):

internal/js/modules/k6/browser/tests/static/get_by_title.html
ankur22: I double checked this and the size of the binary doesn't change when we rename it from `static` to `testData`. I'll leave it as it for now.

Issue #4791, opened 2025-05-15
Implement page.getByTestId(testId: string, options?)


ankur22:
### Feature Description

To streamline selecting elements meant solely for tests, `getByTestId` would query by the `data-testid` attribute (e.g., `page.getByTestId('user-avatar')`). This approach cleanly separates test hooks from production selectors, preventing collateral breakage when styles or layouts change. The implementation should allow exact or regex matching and integrate seamlessly with k6’s existing locator API, giving users a first-class, intention-revealing way to target ephemeral or non-semantic DOM nodes.

Playwright docs: https://playwright.dev/docs/api/class-page#page-get-by-test-id

Note: this is only to be implemented on `page`.

This should work after the implementation work:

```js
  await page.setContent(`
    <html><body>
      <img data-testid="user-avatar" src="avatar.png" />
    </body></html>
  `);
  const avatar = page.getByTestId('user-avatar');
```

### Already existing or connected issues / PRs (optional)

https://github.com/grafana/k6/issues/4248

PR #4911, opened 2025-07-09
Add getByTestId in the browser module

ankur22:
## What?

This adds a convenience method on `page` to be able to select on elements with the `data-testid` attribute.

## Why?

This is part of the story of adding `getBy*`, which makes working with selectors a little easier. We used to have to work with the `page.locator` API providing either a `CSS` selector or a `XPath` selector if we wanted to get the element by the `data-testid` attribute on an element:

```html
<button data-testid="submit-button">Submit</button>
```

```js
const l = page.locator('button[data-testid="submit-button"]'); // CSS
const l = page.locator('//button[@data-testid="submit-button"]'); // XPath
```

Now we can use:

```js
const l = page.getByTestId('submit-button');
```

## Checklist

## Related PR(s)/Issue(s)

https://github.com/grafana/k6/issues/4791, https://github.com/grafana/k6/issues/4248


ankur22:
> Just had one usability issue while testing locally: I wrote only part of the string thinking that it would be non-exact matching like the other APIs but it wasn't. This is how Playwright does it and I guess it makes sense because this a "code" label and not something shown in the UI or used by accessibility tools but it doesn't feel super consistent with the other APIs (maybe also because I'm testing them one by one 😅)

Are you saying it's not consistent with playwright, or not consistent with the other `getBy*` APIs but the behaviour is consistent with playwright? Can you share your script? Could be a bug that we should fix on our side and maybe open an issue on Playwright's side.

AgnesToulet:
It's consistent with Playwright but not with the other `getBy*` APIs (but probably because it's a string used in the codebase and not seen by a user).
Here's a script:
```
	await page.goto("https://agnestoulet1.grafana.net/login", {
		timeout: 60000,
	});

	const options = await page.getByTestId("portal").all(); // Should use /.*portal.*/ to retrieve one element
	console.log(`Found ${options.length} options.`);

	for (const option of options) {
		console.log("Inner text: ", await option.innerText());
	}
	sleep(5);

	await page.close();
```

AgnesToulet (approved):

PR #4826, opened 2025-06-04
Fix web-dashboard output conflict

inancgumus:
## What?

Fix web-dashboard output conflict.

## Why?

Running:
```
K6_WEB_DASHBOARD=true k6 run --out web-dashboard script.js
```

Results:
```
...listen tcp :port: bind: address already in use
```

Also see #4707.

## Checklist

- [x] I have performed a self-review of my code.
- [x] I have commented on my code, particularly in hard-to-understand areas.
- [ ] I have added tests for my changes.
- [x] I have run linter and tests locally (`make check`) and all pass.

## Checklist: Documentation (only for k6 maintainers and if relevant)

**Please do not merge this PR until the following items are filled out.**

- [x] I have added the correct milestone and labels to the PR.
- [ ] I have updated the release notes: _link_

## Related PR(s)/Issue(s)

#4707

PR #4913, opened 2025-07-10
feat: Enable web dashboard by default

### LINKED ISSUE #4903: Enable Built-in Web Dashboard by Default
### **Feature Request: Enable Built-in Web Dashboard by Default**

**Is your feature request related to a problem? Please describe.**

Currently, the Grafana k6 built-in web dashboard is a powerful feature for real-time test visualization, but it is disabled by default. To activate it, users must set the `K6_WEB_DASHBOARD=true` environment variable. This requirement adds an extra step to the initial user experience and may not be immediately discoverable for newcomers. The benefits of real-time monitoring are significant, and making this feature readily available would enhance the out-of-the-box experience with k6.

**Describe the solution you'd like**

The k6 web dashboard should be **enabled by default** when running `k6 run`. This would automatically start the web server and provide users with a link to the dashboard in the console output upon test execution.

To maintain flexibility and accommodate environments where a web server is not desired (e.g., certain CI/CD pipelines), a clear and simple opt-out mechanism should be provided. A suggested environment variable for this is `K6_NO_WEB_DASHBOARD=true`.

The existing `K6_WEB_DASHBOARD_OPEN=true` could still be used to automatically open the dashboard in the browser, but the default behavior would be to simply start the server and display the URL.

**Describe alternatives you've considered**

The current method of using `K6_WEB_DASHBOARD=true` is a viable alternative but is not as user-friendly as having the feature enabled by default. Another alternative would be to prompt the user to enable the dashboard on the first run, but this could add unnecessary complexity to the command-line interface.

**Additional context**

Enabling the web dashboard by default would align with a more "batteries-included" approach, offering a richer initial experience. Many developers and testers, especially those new to k6, would benefit from seeing their test results visualized without having to consult the documentation to enable this feature. The console output could look something like this:

```bash
          /\      |‾‾| /‾‾/   /‾‾/
     /\  /  \     |  |/  /   /  /

  execution: local
     script: test.js
     output: web dashboard available at http://localhost:5665

  scenarios: (100.00%) 1 scenario, 1 max VUs, 10s max duration (executing)
           * default: 1 looping VUs for 10s (gracefulStop: 30s)

running (00m01s/10s), 1/1 VUs, 1 complete and 0 interrupted iterations
...
```

This change would make a powerful k6 feature more accessible and improve the overall developer experience. ✨

szkiba:
### What

This PR changes the behavior of the built-in web dashboard to be enabled by default.

  * The dashboard now starts automatically with `k6 run`.
  * The opt-in `K6_WEB_DASHBOARD` environment variable and `webDashboard` JSON property have been removed.
  * New opt-out mechanisms have been introduced to disable the dashboard:
      * Environment variable: `K6_NO_WEB_DASHBOARD=true`
      * JSON config property: `noWebDashboard: true`

**BREAKING CHANGE:**

The `webDashboard` property and `K6_WEB_DASHBOARD` environment variable are no longer supported. Users who explicitly set these to `true` will now have the dashboard enabled by default and should remove the old settings. To disable the dashboard, they must migrate to the new `noWebDashboard` or `K6_NO_WEB_DASHBOARD` options.

### Why

The primary motivation is to improve the out-of-the-box user experience. By enabling the dashboard by default, we provide immediate, real-time visual feedback for test runs without requiring manual configuration. This makes a powerful feature more discoverable and provides instant value to both new and existing users. The opt-out mechanism ensures that users who run k6 in environments where a web server is not desired (e.g., CI/CD) retain full control.

-----

Closes #4903

### DIFF
--- a/internal/cmd/tests/cmd_run_test.go
+++ b/internal/cmd/tests/cmd_run_test.go
@@ -1952,9 +1952,9 @@ func TestUIRenderWebDashboard(t *testing.T) {
 		active    bool
 		expRender string
 	}{
-		{expRender: "web dashboard:"},
-		{env: "false", expRender: "web dashboard:"},
-		{env: "true", active: true, expRender: "web dashboard: http://127.0.0.1:"},
+		{active: true, expRender: "web dashboard: http://127.0.0.1:"},
+		{env: "false", active: true, expRender: "web dashboard: http://127.0.0.1:"},
+		{env: "true", expRender: "web dashboard:"},
 	}
 
 	for _, tc := range tests {
@@ -1963,7 +1963,7 @@ func TestUIRenderWebDashboard(t *testing.T) {
 
 			ts := NewGlobalTestState(t)
 			if tc.env != "" {
-				ts.Env["K6_WEB_DASHBOARD"] = tc.env
+				ts.Env["K6_NO_WEB_DASHBOARD"] = tc.env
 			}
 			ts.Env["K6_WEB_DASHBOARD_PORT"] = "0"
 			ts.CmdArgs = []string{"k6", "run", "--log-output=stdout"}
--- a/internal/cmd/config.go
+++ b/internal/cmd/config.go
@@ -42,10 +42,10 @@ func configFlagSet() *pflag.FlagSet {
 type Config struct {
 	lib.Options
 
-	Out           []string  `json:"out" envconfig:"K6_OUT"`
-	Linger        null.Bool `json:"linger" envconfig:"K6_LINGER"`
-	NoUsageReport null.Bool `json:"noUsageReport" envconfig:"K6_NO_USAGE_REPORT"`
-	WebDashboard  null.Bool `json:"webDashboard" envconfig:"K6_WEB_DASHBOARD"`
+	Out            []string  `json:"out" envconfig:"K6_OUT"`
+	Linger         null.Bool `json:"linger" envconfig:"K6_LINGER"`
+	NoUsageReport  null.Bool `json:"noUsageReport" envconfig:"K6_NO_USAGE_REPORT"`
+	NoWebDashboard null.Bool `json:"noWebDashboard" envconfig:"K6_NO_WEB_DASHBOARD"`
 
 	// NoArchiveUpload is an option that is only used when running in local-execution mode with the cloud run
 	// command.
@@ -79,8 +79,8 @@ func (c Config) Apply(cfg Config) Config {
 	if cfg.NoUsageReport.Valid {
 		c.NoUsageReport = cfg.NoUsageReport
 	}
-	if cfg.WebDashboard.Valid {
-		c.WebDashboard = cfg.WebDashboard
+	if cfg.NoWebDashboard.Valid {
+		c.NoWebDashboard = cfg.NoWebDashboard
 	}
 	if cfg.NoArchiveUpload.Valid {
 		c.NoArchiveUpload = cfg.NoArchiveUpload
--- a/internal/cmd/config_test.go
+++ b/internal/cmd/config_test.go
@@ -91,10 +91,10 @@ func TestConfigEnv(t *testing.T) {
 			"true":  func(c Config) { assert.Equal(t, null.BoolFrom(true), c.NoUsageReport) },
 			"false": func(c Config) { assert.Equal(t, null.BoolFrom(false), c.NoUsageReport) },
 		},
-		{"WebDashboard", "K6_WEB_DASHBOARD"}: {
-			"":      func(c Config) { assert.Equal(t, null.Bool{}, c.WebDashboard) },
-			"true":  func(c Config) { assert.Equal(t, null.BoolFrom(true), c.WebDashboard) },
-			"false": func(c Config) { assert.Equal(t, null.BoolFrom(false), c.WebDashboard) },
+		{"NoWebDashboard", "K6_NO_WEB_DASHBOARD"}: {
+			"":      func(c Config) { assert.Equal(t, null.Bool{}, c.NoWebDashboard) },
+			"true":  func(c Config) { assert.Equal(t, null.BoolFrom(true), c.NoWebDashboard) },
+			"false": func(c Config) { assert.Equal(t, null.BoolFrom(false), c.NoWebDashboard) },
 		},
 		{"Out", "K6_OUT"}: {
 			"":         func(c Config) { assert.Equal(t, []string{}, c.Out) },
@@ -335,7 +335,7 @@ func TestWriteDiskConfigWithDefaultFlags(t *testing.T) {
 		DefaultFlags: defaultFlags,
 	}
 
-	c := Config{WebDashboard: null.BoolFrom(true)}
+	c := Config{NoWebDashboard: null.BoolFrom(true)}
 	err := writeDiskConfig(gs, c)
 	require.NoError(t, err)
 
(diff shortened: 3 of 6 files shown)


internal/cmd/tests/cmd_run_test.go
```diff
@@ -1952,9 +1952,9 @@ func TestUIRenderWebDashboard(t *testing.T) {
 		active    bool
 		expRender string
 	}{
-		{expRender: "web dashboard:"},
-		{env: "false", expRender: "web dashboard:"},
-		{env: "true", active: true, expRender: "web dashboard: http://127.0.0.1:"},
+		{active: true, expRender: "web dashboard: http://127.0.0.1:"},
```
ankur22, line 1955: > To support environments where a web server is not desired (e.g., CI/CD),
the logic has been inverted from opt-in to opt-out. The dashboard can
now be explicitly disabled.

What will happen in CI/CD pipelines? Could this change break those?

szkiba: In the CI/CD environment, where it is not disabled, the web dashboard is started on IP address 127.0.0.1 on port 5665. If multiple k6s are running on the same host (e.g. non-containerized environment), if there is a port conflict on the default port, the dashboard will only log it and not consider it a fatal error. This means there will be no breaking change in the CI/CD environment.

ankur22 (approved):
LGTM 🚀

szkiba:
@joanlopez 

> I'd suggest to not reverse the configuration attribute name, but just its default value. Otherwise, it would be a breaking change, and I don't see much value on doing so. Wdyt?

I reversed the name of the configuration attribute so that it would be consistent with other attributes with similar roles (like `K6_NO_USAGE_REPORT`). It doesn't cause a breaking change, because until now the `K6_WEB_DASHBOARD` environment variable was set in cases where they wanted to turn it on. Now it's turned on. I think it's important that the configuration attributes are coherent in the long term.

oleiade (changes requested):
Hey @szkiba 👋🏻 

Thanks a lot for putting this together! I think it would make sense to find a way to provide the web-dashboard experience as a default, so it gets the visibility it deserves.

However, I have some concerns about the current approach that I think we should make sure to address before we move forward with it:

### Breaking change impact
This is fundamentally a breaking change since we're modifying the options structure and flipping a previously opt-in behavior to default. While the user-experience improvement is valuable, introducing breaking changes in a minor release goes [against our new versioning/maintenance policy](https://grafana.com/docs/k6/latest/reference/versioning-and-stability-guarantees/), and could disrupt existing integrations and workflows.

### Performance Considerations

I don't remember if we had thoroughly tested the performance impact of enabling the web dashboard by default across different scenarios. I think we might have though, and if such, could you point us to the results? Given that this will now run for every k6 execution, we should ensure that we have a good idea of the overhead it would introduce.

### Ecosystem coordination

This change will most likely require some alignment and coordination across our internal ecosystem, as well as public one (k6-operator?), in order to explicitly disable the dashboard where appropriate. We should have a clear migration plan and timeline for these updates, before we proceed.

## Solutions

I'd like to offer alternative solutions:

1. **Delay for v2**: Plan this as part of our next major release where breaking changes are expected.
2. **Non-breaking transition**: Is there a way we could improve the visibility of this feature without having to switch to default enable in the meantime: better docs, cli hints, etc?
3. **Gradual rollout**: Perhaps we could also look into way to "decide" whether the dashboard can be enabled by default, or at least proposed to the user based on what kind of test we run, and in what environment.

To be clear, I'm definitely not against the end goal, but I want to make sure we're thoughtful about the implementation path. What do you folks think?

szkiba:
> Hey @szkiba 👋🏻
> 
> Thanks a lot for putting this together! I think it would make sense to find a way to provide the web-dashboard experience as a default, so it gets the visibility it deserves.
> 
> However, I have some concerns about the current approach that I think we should make sure to address before we move forward with it:
> ### Breaking change impact
> 
> This is fundamentally a breaking change since we're modifying the options structure and flipping a previously opt-in behavior to default. While the user-experience improvement is valuable, introducing breaking changes in a minor release goes [against our new versioning/maintenance policy](https://grafana.com/docs/k6/latest/reference/versioning-and-stability-guarantees/), and could disrupt existing integrations and workflows.
> ### Performance Considerations
> 
> I don't remember if we had thoroughly tested the performance impact of enabling the web dashboard by default across different scenarios. I think we might have though, and if such, could you point us to the results? Given that this will now run for every k6 execution, we should ensure that we have a good idea of the overhead it would introduce.
> ### Ecosystem coordination
> 
> This change will most likely require some alignment and coordination across our internal ecosystem, as well as public one (k6-operator?), in order to explicitly disable the dashboard where appropriate. We should have a clear migration plan and timeline for these updates, before we proceed.
> ## Solutions
> 
> I'd like to offer alternative solutions:
> 
>     1. **Delay for v2**: Plan this as part of our next major release where breaking changes are expected.
> 
>     2. **Non-breaking transition**: Is there a way we could improve the visibility of this feature without having to switch to default enable in the meantime: better docs, cli hints, etc?
> 
>     3. **Gradual rollout**: Perhaps we could also look into way to "decide" whether the dashboard can be enabled by default, or at least proposed to the user based on what kind of test we run, and in what environment.
> 
> 
> To be clear, I'm definitely not against the end goal, but I want to make sure we're thoughtful about the implementation path. What do you folks think?

Hi @oleiade 

Thank you for the thorough review and for outlining your concerns. I appreciate you taking the time to provide this detailed feedback.

Regarding the **breaking change impact**, I completely agree with the principle of avoiding breaking changes in minor releases. To help me better understand the full scope, could you elaborate on some of the use cases you believe would be disrupted? I'm trying to picture if this would affect CI/CD pipelines that parse logs, specific integrations, or other workflows I might not be aware of. Your perspective here would be very helpful.

On the topic of **performance considerations**, I agree that having solid performance data is essential before moving forward. I wanted to share some context from my previous attempt to run these tests about a year and a half ago, where I ran into some difficulties with the performance test suite itself. It seemed at the time that it wasn't in a state where I could easily get a clean measurement.

With that in mind, has the state of the performance test suite changed recently? If it's now more accessible, I would be happy to give it another shot. If the suite itself still requires maintenance, I agree that this is a blocker. In that case, I'd appreciate the core team's guidance on whether they have the capacity to investigate the performance impact, or if there are resources I could use to help get the tests running.

Thanks again for the crucial feedback. I'm keen to find the right path forward for this.

szkiba:
The performance impact could be reduced if the web dashboard was only enabled automatically when k6 is run from a terminal (stdio is a terminal). In this case, I would keep the original opt-in environment variable name (`K6_WEB_DASHBOARD`)

inancgumus:
@szkiba I believe this PR takes over #4826. I'll close #4826 if that's the case. Please let me know. Thanks!

szkiba:
> @szkiba I believe this PR takes over #4826. I'll close #4826 if that's the case. Please let me know. Thanks!

@inancgumus  Actually, xk6-dashboard [v0.7.9]( https://github.com/grafana/xk6-dashboard/releases/tag/v0.7.9) fixes the conflict mentioned in #4826 as a side effect. v0.7.9 has already been merged into master. So I think #4826 can be closed.

Honestly, I originally thought it wasn't worth fixing because the output name "web-dashboard" is unsupported and undocumented. The fact that the user found out that the built-in dashboard is an output name called "web-dashboard" doesn't justify fixing the conflict.

However, the fix is done, it might be worth merging (I thought the fix was simple and elegant).

oleiade:
In the spirit of keeping the k6 maintainers' backlog clean, we'll be closing this PR.

The steps forward will most likely be to bring the HTML report feature to the k6 core project, but leave the dashboard as an extension now that we have native extensions support.

Please feel free to reopen this PR in the future if there is a change of decision 🙇‍♂️

Issue #4895, opened 2025-07-02
k6 browser does not surface browser network errors when navigating to a page


nadiamoe:
### Brief summary

A script that navigates and waits for a page to load will report no error even if Chromium cannot load that page due to a network error.

### k6 version

k6 v1.1.0 (go1.24.4, linux/amd64)

### OS

Arch linux (btw)

### Docker version and image (if applicable)

N/A

### Steps to reproduce the problem

Run the following script, provided that nothing is running on `localhost:4444` (so connections to that port **will fail**):

```js
import { browser } from 'k6/browser';

export const options = {
  scenarios: {
    ui: {
      executor: 'shared-iterations',
      options: {
        browser: {
          type: 'chromium',
        },
      },
    },
  },
};

export default async function () {
  const context = await browser.newContext();
  const page = await context.newPage();
  const startUrl = 'http://localhost:4444';

  try {
    await page.goto(startUrl);
    await page.waitForLoadState('networkidle');
    const myLocator = await page.locator('#something');
    // myLocator.click();
  } catch (error) {
    console.error(error);
  } finally {
    await page.close();
  }
}
```

### Expected behaviour

k6 should error out complaining that the page it was asked to navigate to was unreachable.

### Actual behaviour

k6 will report somewhat sensical data and exit with success (0):
```
nadia@Nadiarch 󱃾 curry-admin@Curry
17:58:37 /tmp/nadia $> k6 run repro.js

         /\      Grafana   /‾‾/
    /\  /  \     |\  __   /  /
   /  \/    \    | |/ /  /   ‾‾\
  /          \   |   (  |  (‾)  |
 / __________ \  |_|\_\  \_____/


     execution: local
        script: repro.js
        output: -

     scenarios: (100.00%) 1 scenario, 1 max VUs, 10m30s max duration (incl. graceful stop):
              * ui: 1 iterations shared among 1 VUs (maxDuration: 10m0s, gracefulStop: 30s)



  █ TOTAL RESULTS

    EXECUTION
    iteration_duration.............................: avg=2.13s    min=2.13s    med=2.13s    max=2.13s    p(90)=2.13s    p(95)=2.13s
    iterations.....................................: 1     0.438437/s
    vus............................................: 1     min=1      max=1
    vus_max........................................: 1     min=1      max=1

    NETWORK
    data_received..................................: 0 B   0 B/s
    data_sent......................................: 0 B   0 B/s

    BROWSER
    browser_data_sent..............................: 185 B 81 B/s

[...]

PR #4916, opened 2025-07-10
Refine console.log to properly print errors

joanlopez:
## What?

It adds some new test cases to some of the `console_test.go` tests in order to make sure that different values are console logged in a meaningful way. It also adds the changes required in the `func (c console) valueString(v sobek.Value) string` method to adequately convert such values into a meaningful string (for human readability).

## Why?

Because while working on https://github.com/grafana/k6/issues/4895 (_I'll create a PR for that, right after this one_), I realized that in most cases, when we have the following pattern:

```js
try {
  ...
  await someAsyncOperation()
} catch(error) {
  console.error(error)
}
``` 

The error captured isn't being correctly printed, but k6 just prints: `{}`. Giving no clue to the user about what actually failed.

The workaround in such case is to use: `error.toString()`, but in my opinion that's not trivial to figure out, and it causes a bad user experience, while one of our main goals is to improve the ease of use of the tool.

Arguably, this could be fixed by finding a way where the _thrown_ value isn't represented as an object, and thus it doesn't implement the `json.Marshaler` interface. But, there's no trivial way to do so, because:
- There's no (feasible?) way to instantiate a new `sobek.Exception` from Go code, which doesn't satisfy the aforementioned  interface.
- The `Sobek.Runtime.ToValue()` method represents any non-explicitly recognized Go value as an object (`sobek.objectGoReflect` to be concrete), which also includes the standard Go `error`, and thus suffers the same issue.

Indeed, an example as simple as:

```js
console.log(Error("error"))
```

would, again, just print `{}`.

So, reached this point, even if it's not a perfect solution, I think it's good idea to try to slightly improve the current status quo, but letting k6 users have meaningful messages when they `console.log` the errors they got, especially considering these changes shouldn't have any drawback in the form of side-effect, as far as I can tell.

## Checklist


ankur22 (approved):
Seems to be in the correct direction.

internal/js/console.go
```diff
@@ -77,7 +77,25 @@ func (c console) Error(args ...sobek.Value) {
...
 func (c console) valueString(v sobek.Value) string {
+	if _, isFunction := sobek.AssertFunction(v); isFunction {
+		return functionLog
+	}
+
+	if exported := v.Export(); exported != nil {
+		if err, isError := exported.(error); isError {
+			return err.Error()
+		}
+	}
```
mstoykov: `.Export` is one of the more heavy operations in some cases. It will be nice if we can use [ExportType](https://pkg.go.dev/github.com/grafana/sobek#Symbol.ExportType), before doing an Export.

joanlopez: Yeah, right! Good observation! At some point, I mentally noted this as something to review later but I finally forgot it. Good that you spot it during the review, thanks! 🙌🏻 

Does an approach like https://github.com/grafana/k6/pull/4916/commits/a22df531deb70a9d895aad9ea8d1df275e996133 look good?

Thanks! 🙇🏻

ankur22 (approved):

ankur22 (approved):

Issue #4461, opened 2022-02-24
Support URL option for `Page.waitForNavigation`


imiric:
Currently the [URL option for `waitForNavigation`](https://github.com/grafana/xk6-browser/blob/ae264530b6fbe0285ac9df9b0f4a907b460bb805/common/frame_options.go#L151) is unused.

Previously it wasn't a major priority, but now that [waiting for navigation within the same document works](https://github.com/grafana/xk6-browser/pull/247), we need a way to distinguish same document and new document navigations.

## Suggested solution

[Playwright handles this with the URL option](https://playwright.dev/docs/navigations#multiple-navigations), and it makes sense for us to do the same.

Issue #4385, opened 2023-03-13
Add support for page.waitForURL


mdcruz:
### Feature Description

According to the Playwright documentation, [page.waitForNavigation()](https://playwright.dev/docs/api/class-page#page-wait-for-navigation) has been deprecated. We have been using this in our examples and recommendations so far. 

For example:

```
await Promise.all([
      page.waitForNavigation(),
      page.locator('input[type="submit"]').click(),
    ]);
```

Playwright has created [page.waitForURL()](https://playwright.dev/docs/api/class-page#page-wait-for-url) in favour of `page.waitForNavigation()`.

The new usage can look like this:

```
await Promise.all([
      page.locator('input[type="submit"]').click(),
      page.waitForURL('**/test'),
]);
```

We should also try and support this in future versions of k6 browser.

PR #4917, opened 2025-07-10
Add `url` option for `waitForNavigation`

ankur22:
## What?

This adds/uses the `url` option in `waitForNavigation`. This extends `waitForNavigation` so that k6 waits for a specific URL before continuing on.

## Why?

Many users have to work with websites which perform lots of redirects during an auth login process. They would currently have to either:

```js
    await Promise.all([
      page.waitForNavigation(),
      page.locator('a[href="/my_messages.php"]').click(),
    ]);

    // Wait for an arbitrary amount of time to make sure that any redirects during the navigation have completed entirely and we are now on the specific page.
    page.waitForTimeout(2000);
```

Or

```js
    await page.locator('a[href="/my_messages.php"]').click()

    await page.locator('//div[text()="You made it passed all the redirects!"]').waitFor()
```

Now we have another option for those who want a repeatable, consistent and deterministic way where the url doesn't change as often as the elements on the page and without arbitrary time outs, which is to do:

```js
    await Promise.all([
      page.waitForNavigation({ url:'https://quickpizza.grafana.com/my_messages.php' }),
      page.locator('a[href="/my_messages.php"]').click(),
    ]);
```

This also works for regex:

```js
    const [response3] = await Promise.all([
      page.waitForNavigation({ url: /.*\/contacts\.php.*/ }),
      page.locator('a[href^="/contacts.php"]').click()
    ]);
```

## Checklist

## Related PR(s)/Issue(s)

https://github.com/grafana/k6/issues/4461, https://github.com/grafana/k6/issues/4385

### DIFF
--- a/internal/js/modules/k6/browser/browser/frame_mapping.go
+++ b/internal/js/modules/k6/browser/browser/frame_mapping.go
@@ -388,7 +388,7 @@ func mapFrame(vu moduleVU, f *common.Frame) mapping {
 			}
 
 			return k6ext.Promise(vu.Context(), func() (result any, reason error) {
-				resp, err := f.WaitForNavigation(popts)
+				resp, err := f.WaitForNavigation(popts, nil)
 				if err != nil {
 					return nil, err //nolint:wrapcheck
 				}
--- a/internal/js/modules/k6/browser/browser/frame_mapping.go
+++ b/internal/js/modules/k6/browser/browser/frame_mapping.go
@@ -387,8 +387,15 @@ func mapFrame(vu moduleVU, f *common.Frame) mapping {
 				return nil, fmt.Errorf("parsing frame wait for navigation options: %w", err)
 			}
 
-			return k6ext.Promise(vu.Context(), func() (result any, reason error) {
-				resp, err := f.WaitForNavigation(popts, nil)
+			// Inject JS regex checker for URL regex pattern matching
+			ctx := vu.Context()
+			jsRegexChecker, err := injectRegexMatcherScript(ctx, vu, f.Page().TargetID())
+			if err != nil {
+				return nil, err
+			}
+
+			return k6ext.Promise(ctx, func() (result any, reason error) {
+				resp, err := f.WaitForNavigation(popts, jsRegexChecker)
 				if err != nil {
 					return nil, err //nolint:wrapcheck
 				}
--- a/internal/js/modules/k6/browser/common/frame.go
+++ b/internal/js/modules/k6/browser/common/frame.go
@@ -1902,7 +1902,7 @@ func (f *Frame) WaitForNavigation(
 	jsRegexChecker JSRegexChecker,
 ) (*Response, error) {
 	f.log.Debugf("Frame:WaitForNavigation",
-		"fid:%s furl:%s", f.ID(), f.URL())
+		"fid:%s furl:%s url:%s", f.ID(), f.URL(), opts.URL)
 	defer f.log.Debugf("Frame:WaitForNavigation:return",
 		"fid:%s furl:%s", f.ID(), f.URL())
 
@@ -1930,6 +1930,9 @@ func (f *Frame) WaitForNavigation(
 			e := &k6ext.UserFriendlyError{
 				Err:     err,
 				Timeout: opts.Timeout,
+			}
+			if opts.URL != "" {
+				return fmt.Errorf("waiting for navigation to URL matching %q: %w", opts.URL, e)
 			}
 			return fmt.Errorf("waiting for navigation: %w", e)
 		}
--- a/internal/js/modules/k6/browser/common/frame.go
+++ b/internal/js/modules/k6/browser/common/frame.go
@@ -1894,7 +1894,13 @@ func (f *Frame) WaitForLoadState(state string, popts *FrameWaitForLoadStateOptio
 }
 
 // WaitForNavigation waits for the given navigation lifecycle event to happen.
-func (f *Frame) WaitForNavigation(opts *FrameWaitForNavigationOptions) (*Response, error) {
+// jsRegexChecker should be non-nil to be able to test against a URL pattern in the options.
+//
+//nolint:funlen
+func (f *Frame) WaitForNavigation(
+	opts *FrameWaitForNavigationOptions,
+	jsRegexChecker JSRegexChecker,
+) (*Response, error) {
 	f.log.Debugf("Frame:WaitForNavigation",
 		"fid:%s furl:%s", f.ID(), f.URL())
 	defer f.log.Debugf("Frame:WaitForNavigation:return",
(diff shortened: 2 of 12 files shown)


internal/js/modules/k6/browser/browser/frame_mapping.go
```diff
@@ -387,8 +387,15 @@ func mapFrame(vu moduleVU, f *common.Frame) mapping {
 				return nil, fmt.Errorf("parsing frame wait for navigation options: %w", err)
 			}
 
-			return k6ext.Promise(vu.Context(), func() (result any, reason error) {
-				resp, err := f.WaitForNavigation(popts)
+			// Inject JS regex checker for URL pattern matching
+			ctx := vu.Context()
+			jsRegexChecker, err := injectURLMatcherScript(ctx, vu, f.Page().TargetID())
```
joanlopez: Out of curiosity, is this code executed every time there's a call to `waitForNavigation`? I guess it's a tradeoff, between this and injecting it only once but always, and I'm not familiar enough with the overhead of injecting scripts, but wondering whether there's another approach that could be slightly better 🤔 

PS: The same I guess would apply for `page_mapping`.

internal/js/modules/k6/browser/common/frame.go
```diff
@@ -1866,22 +1886,45 @@ func (f *Frame) WaitForLoadState(state string, popts *FrameWaitForLoadStateOptio
...
+	matcher, err := urlMatcher(opts.URL, jsRegexChecker)
+	if err != nil {
+		timeoutCancel()
+		return nil, fmt.Errorf("parsing URL pattern: %w", err)
+	}
+
+	navEvtCh, navEvtCancel := createWaitForEventPredicateHandler(timeoutCtx, f, []string{EventFrameNavigation},
 		func(data any) bool {
-			return true // Both successful and failed navigations are considered
+			if navEvt, ok := data.(*NavigationEvent); ok {
```
joanlopez, line 1921: As I did for https://github.com/grafana/k6/pull/4918, what if `navEvt.err != nil`? 🤔 
Should we early return `false` in such case?

internal/js/modules/k6/browser/tests/page_test.go
```diff
@@ -2787,3 +2786,132 @@ func TestPageMustUseNativeJavaScriptObjects(t *testing.T) {
 	_, err = page.QueryAll("#textField")
 	require.NoErrorf(t, err, "page should not override the native objects, but it did")
 }
+
+func TestWaitForNavigationWithURL(t *testing.T) {
```
joanlopez, line 2790: Can we add a test case where the user provides invalid input, like something that's not a valid regular expression?

internal/js/modules/k6/browser/browser/frame_mapping.go
ankur22: This is running the JS code into the current Sobek context during runtime. I was wondering if there was a way of doing it once, or if there was a best practices. With our existing APIs (such as `page.on`) we run the JS script every time `page.on` is called.

For the browser module, I think the bottleneck is the JSON marshalling/unmarshalling when it talks to chromium over the websocket, which this isn't using, so maybe we don't need to worry about it today?

ankur22:
> Do we plan to also work on waitForURL()? I guess that, even if full Playwright compatibility isn't the main focus, it'd be nice to be compatible on this aspect, especially if it's low effort (I guess most of the code would be the same and it would be mostly about exposing the new function).

Yep! here's the [PR](https://github.com/grafana/k6/pull/4920) for it 😄 It is a requested API from a user, and since it builds on `page.waitForNavigation` with the `url` option i split the two changes into two PRs.

> Not to be nitpicky, but to fully understand the context of this PR; I see the url option was already there, is it that it wasn't being taken into consideration? I see, for instance, it's not mention in docs.

Correct, the `url` option was already there but wasn't being used. I'm not entirely sure why `url` wasn't implemented at the same time as the rest of the API, i suspect other priorities got in the way, and was deemed to be less of a priority since `waitForNavigation` without `url` was the most important functionality that was needed back when it was first worked on.

internal/js/modules/k6/browser/common/frame.go
ankur22: Good point. Depending on which is merged first, we definitely need to take this into account.

In terms of early exit, I'm wondering if we need to wait for all the background go routines to complete before exiting. I'd need to test that first before I could say for sure. It would seem that as long as all the contexts and channels are closed then it might be ok. However, since in this method the channels are created and read off of (not written), i would think closing them could be an issue from here... so maybe best to wait for all the channels/goroutines to complete and then return an error at the end.

internal/js/modules/k6/browser/tests/page_test.go
ankur22: Tests simply do not run when regular expressions are invalid. I tried with something like:

```js
  await page.goto('https://quickpizza.grafana.com/test.k6.io/');

  await Promise.all([
    page.waitForURL(/^.*/my_messages.*$/),
    page.locator('a[href="/my_messages.php"]').click(),
  ]);
```

Error:

```
ERRO[0000] GoError: file:///Users/ankuragarwal/go/src/github.com/grafana/k6/e2e/invalid_regex.js: Line 25:38 Unexpected token * (and 5 more errors)  hint="script exception"
```

It just doesn't run at all. Do you have an example anywhere that I can take a look at?

ankur22:
I have removed the glob pattern matcher from this implementation. The reason for this is that the glob pattern matcher in Playwright is quite complex, and rather than copy it here, we will wait for user feedback on whether we need it or not.

internal/js/modules/k6/browser/tests/page_test.go
joanlopez: I think erroring out is fine, I just want to understand how this feature behave when the user inputs an invalid regular expression, especially from the user perspective.

internal/js/modules/k6/browser/common/frame.go
joanlopez: > Depending on which is merged first, we definitely need to take this into account.

This is new code, so I guess there's no need to sync both PRs. Right?

ankur22: The changes i've made here are in `waitForNavigation` in `frame`, which i think you've handled already [here](https://github.com/grafana/k6/pull/4918/files#r2208196572). So once i merge in your changes then i should get it for free 😄

internal/js/modules/k6/browser/tests/page_test.go
```diff
@@ -2787,3 +2786,102 @@ func TestPageMustUseNativeJavaScriptObjects(t *testing.T) {
...
+			await page.close();
+		}
+	`,
+		tb.staticURL("waitfornavigation_test.html"),
+		tb.staticURL("page1.html"),
+	)
+	assert.Equal(t, sobek.Undefined(), got.Result())
+}
+
+func TestWaitForNavigationWithURL_RegexFailure(t *testing.T) {
```
ankur22, line 2872: @joanlopez I created this test, but it's really only testing the Sobek side of things. Invalid regex don't manage to get passed to the `waitForURL` API itself.

internal/js/modules/k6/browser/tests/page_test.go
ankur22: I created 94a0bca671a4978fe668dd8d024432355962a76a and 80612d7e4d5d1dfd34d021fe2142ab0e7c7c91ac.

internal/js/modules/k6/browser/tests/page_test.go
joanlopez: So, I guess the error returned by Sobek (when calling `RunAsync`), would be possible to `catch` and `console.log`, right? 🤔  Thanks!

joanlopez (approved):

PR #4918, opened 2025-07-10
Browser: Surface navigation errors from navigation events

### LINKED ISSUE #4895: k6 browser does not surface browser network errors when navigating to a page

joanlopez:
## What?

It surfaces navigation errors from navigation events.

## Why?

Because as explained in #4895, the fact that the error is surfaced when an action is performed misleads users to think the problem lies in that action, and not in the page not being loaded in the first place.

## Checklist

## Related PR(s)/Issue(s)

Closes #4895


joanlopez:
@ankur22, do you think I should try to write a Go test to make it sure we don't break that behavior again? 

In such case, could you recommend a couple of Go tests to use as an example/reference to write this, as I see there are no tests for this particular file, and I guess considering it's part of the Browser module, there's probably some things to be initialized for the test, and to be considered. So, any pointer will really help a lot! 🙏🏻  Thanks!

internal/js/modules/k6/browser/common/frame.go
```diff
@@ -1915,6 +1915,10 @@ func (f *Frame) WaitForNavigation(opts *FrameWaitForNavigationOptions) (*Respons
 	select {
 	case evt := <-navEvtCh:
 		if e, ok := evt.(*NavigationEvent); ok {
+			if e.err != nil {
```
joanlopez, line 1918: Note that here I'm early returning the error, while in the `frame_manager.go` code I'm just "saving" the error, to return it later. But I'm not 100% about all possible cases here, and I'm wondering if this is just yet another case in Go where on a given error it's enough with early returning the error, or if the response might have something meaningful that could complement the error 🤔

ankur22: If it's a network error then i wouldn't expect a response object, so _I think_ it's ok to not worry about waiting for a response.

joanlopez: Gotcha, thanks! I have _force pushed_ the changes to return the error immediately in both cases.

Any suggestion about tests?

ankur22: Actually, having taken a better look at this and compared against another PR that was trying to solve this same issue (specifically this [comment](https://github.com/grafana/k6/pull/4862/commits/e9f2b6e57e84a313f67e28dcf73b51a02e206f1a#diff-2ba58823421b91d5955b4483d7ebb7fd3a96a7a45ec9d3081c74e26bc06a0b23R703-R705) in that PR), i think you were correct before (not returning early).

internal/js/modules/k6/browser/common/frame_manager.go
```diff
@@ -682,6 +682,10 @@ func (m *FrameManager) NavigateFrame(frame *Frame, url string, parsedOpts *Frame
 	select {
 	case evt := <-navEvtCh:
 		if e, ok := evt.(*NavigationEvent); ok {
+			if e.err != nil {
+				return nil, e.err
+			}
+
```
ankur22, line 697: Good spot, you've found that we need to handle this for `goto` and `waitForNavigation`. I think there's one more that we need to account for, which is ~`page.goto`~ `page.retry`: https://github.com/grafana/k6/blob/e422e35b3d642d24b20f11831aba177af0e714d9/internal/js/modules/k6/browser/common/page.go#L1361-L1363. WDYT?

internal/js/modules/k6/browser/common/frame.go
```diff
@@ -1915,6 +1915,10 @@ func (f *Frame) WaitForNavigation(opts *FrameWaitForNavigationOptions) (*Respons
...
+				return nil, e.err
+			}
+
 			if e.newDocument == nil {
 				sameDocNav = true
 				break
```
ankur22, line 1924: Re tests:

It looks like this change breaks the same tests that i broke with my PR ([TestBlockHostnames](https://github.com/grafana/k6/blob/7129195a85e57533564e46086018e43915d27945/internal/js/modules/k6/browser/tests/network_manager_test.go#L43) and [TestBlockIPs](https://github.com/grafana/k6/blob/7129195a85e57533564e46086018e43915d27945/internal/js/modules/k6/browser/tests/network_manager_test.go#L73)). So they will need to be fixed first. i really struggled to get my tests to pass in CI though :(

Having had all those random issues in CI, I would opt to work with a different approach of creating a test to see if it helps prevent odd behaviours in CI -- take a look at [TestSelectOption](https://github.com/grafana/k6/blob/1ce8053f9295c492659d922c24199493e9348a84/internal/js/modules/k6/browser/tests/locator_test.go#L688).

Then it should be a case of recreating the test that was in the original [issue](https://github.com/grafana/k6/issues/4895).

internal/js/modules/k6/browser/common/frame_manager.go
```diff
@@ -599,14 +599,21 @@ func (m *FrameManager) NavigateFrame(frame *Frame, url string, parsedOpts *Frame
 	timeoutCtx, timeoutCancelFn := context.WithTimeout(m.ctx, parsedOpts.Timeout)
 	defer timeoutCancelFn()
 
-	newDocIDCh := make(chan string, 1)
+	var newDocumentID string
+	newDocIDIsReadyCh := make(chan struct{})
```
joanlopez, line 603: The problem here was that there might be more than one `NavigationEvent`. For instance, in the case where you try to navigate to a blocked IP or hostname.

And, if there's more than one event, the line `newDocID := <-newDocIDCh` below is blocking for all the events except the first one. But the first one isn't necessarily the one we're really interested on (the one for which we return `true`).

So, instead, I only use the channel as a synchronization primitive (see I `close` it below, once we have the document identifier), while the value is stored in a standard variable, read by the closure, instead of using a channel for sending such value.

internal/js/modules/k6/browser/common/frame.go
```diff
@@ -1915,6 +1915,10 @@ func (f *Frame) WaitForNavigation(opts *FrameWaitForNavigationOptions) (*Respons
 	select {
 	case evt := <-navEvtCh:
 		if e, ok := evt.(*NavigationEvent); ok {
+			if e.err != nil {
+				return nil, e.err
```
joanlopez, line 1919: Here's the key to surface the errors, previously we were ignoring if the `NavigationEvent` was holding an `error`.

internal/js/modules/k6/browser/common/frame_manager.go
```diff
@@ -682,6 +691,10 @@ func (m *FrameManager) NavigateFrame(frame *Frame, url string, parsedOpts *Frame
 	select {
 	case evt := <-navEvtCh:
 		if e, ok := evt.(*NavigationEvent); ok {
+			if e.err != nil {
+				return nil, e.err
```
joanlopez, line 695: Here's the key to surface the errors, previously we were ignoring if the `NavigationEvent` was holding an `error`.

internal/js/modules/k6/browser/common/helpers.go
```diff
@@ -116,15 +117,6 @@ func call(
 	return result, err
 }
 
-func stringSliceContains(s []string, e string) bool {
```
joanlopez, line 119: Just trying to be a good boy scout, while I passed through this code. But if you really consider it critical, I can move this into a separate PR. I'd prefer not to, so we keep things simpler. Overall I think it's a pretty trivial change, and isolated on its own commit.

internal/js/modules/k6/browser/common/helpers.go
```diff
@@ -138,34 +130,27 @@ func createWaitForEventHandler(
...
 							return
 						}
 					}
-					close(ch)
-
-					// We wait for one matching event only,
-					// then remove event handler by cancelling context and stopping goroutine.
-					evCancelFn()
-
-					return
```
joanlopez: This is connected to what I wrote in https://github.com/grafana/k6/pull/4918/files#r2208193018.

We can receive more than one event of the same type, and the first one isn't necessarily the one we want.

internal/js/modules/k6/browser/tests/network_manager_test.go
```diff
@@ -22,21 +22,32 @@ func TestURLSkipRequest(t *testing.T) {
...
-	}
 	_, err := p.Goto(
 		"data:text/html,hello",
-		opts,
+		&common.FrameGotoOptions{Timeout: common.DefaultTimeout},
 	)
 	require.NoError(t, err)
 	tb.logCache.assertContains(t, "skipping request handling of data URL")
 
+	// In this test, we're checking that the network manager is skipping request handling
```
joanlopez, line 32: This wasn't needed before because we were not surfacing errors. So, we were just "ignoring" everything.
In this case, I think this is the expected behavior for an invalid Blob URL, but I'm open to suggestions.

In the future:
- If we ever support both `Blob` and `URL` WebAPIs, we can slightly modify this test to try to navigate to a valid Blob URL. In such case, we should still see the log, because we'll still skip any network operation, but the operation shouldn't timeout.
- If we ever find a way to surface `NavigationEvent`s that navigate to `chrome-error://`, we could also adjust this test case, to expect the error surfaced from such event. But this is out of the scope of #4895.

internal/js/modules/k6/browser/common/frame_manager.go
ankur22: Could you work with `createWaitForEventPredicateHandler` instead of `createWaitForEventHandler`? `createWaitForEventPredicateHandler` should wait until the predicate function returns `true` ~and then it will pass the data to the `navEvtCh`~.

EDIT: I now understand why you have made the change to work with `newDocIDIsReadyCh` as a synchronization primitive, and i think that makes sense 👍

internal/js/modules/k6/browser/common/helpers.go
ankur22: Nice refactor! I would be scared to make this change myself since it feels pretty fragile and a core piece to orchestrate various areas in the browser module, but I think it makes sense 😄

ankur22 (commented):
Thanks for all the comments and work on this! I think it's coming along really well. I've left a few things to (re)consider.

internal/js/modules/k6/browser/common/helpers.go
ankur22: Actually, this changes the behaviour of `createWaitForEventHandler`. It now doesn't exit until the predicate function passes when it returns `true`. I think we should revert this change and use `createWaitForEventPredicateHandler`.

internal/js/modules/k6/browser/common/frame_manager.go
joanlopez: Sure, applied in https://github.com/grafana/k6/pull/4918/commits/168755784d2c9aa00e089a151abe28361bfacfe5

internal/js/modules/k6/browser/common/frame.go
joanlopez: Well, I think these tests were broken because they weren't expecting the error to be surface, which is precisely the goal of this changeset. Now I think should be fine. Do you have any other suggestion?

ankur22 (approved):
LGTM 🚀

oleiade (approved):

PR #4920, opened 2025-07-10
Add waitForURL in frame and page

ankur22:
## What?

This adds support for a asked for API from Playwright -- `page.waitForURL`. This API does a few things more than the equivalent `page.waitForNavigation({ url: ... })`:

1. It will compare the current page's URL to see if it matches the input url;
2. If there's a match it will then wait for the load state of the navigation;
3. Otherwise it will use `page.waitForNavigation({ url: ... })`.

## Why?

This is a more robust API than `page.waitForNavigation` when a user wants to wait for a specific URL, since it builds on `page.waitForNavigation` to prevent race conditions where a page completes navigation before `page.waitForNavigation` is setup.

It's just another convenient way of waiting for something specific when a navigation occurs, which is especially important when many redirects occur.

Note: It's still recommended that the user wait for specific elements or just work with the locator API which auto waits when the page does navigate to the specified URL.

Example usage:

```js
  await Promise.all([
    page.waitForURL('https://quickpizza.grafana.com/login'),
    page.locator('a[href="Login/Profile"]').click()
  ]);
```

## Checklist

## Related PR(s)/Issue(s)

https://github.com/grafana/k6/issues/4385

### DIFF
--- a/internal/js/modules/k6/browser/tests/page_test.go
+++ b/internal/js/modules/k6/browser/tests/page_test.go
@@ -2891,3 +2891,109 @@ func TestWaitForNavigationWithURL_RegexFailure(t *testing.T) {
 	)
 	assert.ErrorContains(t, err, "Unexpected token *")
 }
+
+func TestWaitForURL(t *testing.T) {
+	if runtime.GOOS == "windows" {
+		t.Skip("Skipped due to https://github.com/grafana/k6/issues/4937")
+	}
+
+	t.Parallel()
+
+	tb := newTestBrowser(t, withFileServer())
+	tb.vu.ActivateVU()
+	tb.vu.StartIteration(t)
+
+	got := tb.vu.RunPromise(t, `
+		const page = await browser.newPage();
+		const testURL = '%s';
+
+		try {
+			// Test when already at matching URL (should just wait for load state)
+			await page.goto('%s');
+			await page.waitForURL(/.*page1\.html$/);
+			let currentURL = page.url();
+			if (!currentURL.endsWith('page1.html')) {
+				throw new Error('Expected to stay at page1.html but got ' + currentURL);
+			}
+
+			// Test exact URL match with navigation
+			await page.goto(testURL);
+			await Promise.all([
+				page.waitForURL('%s'),
+				page.locator('#page1').click()
+			]);
+			currentURL = page.url();
+			if (!currentURL.endsWith('page1.html')) {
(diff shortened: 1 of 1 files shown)


internal/js/modules/k6/browser/browser/frame_mapping.go
```diff
@@ -422,6 +423,31 @@ func mapFrame(vu moduleVU, f *common.Frame) mapping {
...
+				return nil, fmt.Errorf("parsing waitForURL options: %w", err)
+			}
+
+			var val string
+			switch url.ExportType() {
+			case reflect.TypeOf(string("")):
+				val = fmt.Sprintf("'%s'", url.String()) // Strings require quotes
+			default: // JS Regex, CSS, numbers or booleans
+				val = url.String() // No quotes
+			}
```
ankur22, line 438: I will refactor this into a helper method once this PR is merged in. There are several other PRs which work with similar/same logic to differentiate between regex and strings.

internal/js/modules/k6/browser/tests/page_test.go
```diff
@@ -2891,3 +2891,109 @@ func TestWaitForNavigationWithURL_RegexFailure(t *testing.T) {
 	)
 	assert.ErrorContains(t, err, "Unexpected token *")
 }
+
+func TestWaitForURL(t *testing.T) {
```
AgnesToulet, line 2895: Nit: It might be clearer to split this into several tests (the `%s` are not easy to map with the actual value and if one test fails here, it won't let the others to continue, which helps with debugging and fixing all the failing tests faster).

ankur22:
> Do we plan to deprecate WaitForNavigation (and maybe remove it in the next major version)? If so, do we have/should we have an issue to plan for that? I see it's deprecated in PW doc and I think it makes sense to remove it from our API as our goal is to have a simpler API.

I'm not 100% sure what the next steps are for `WaitForNavigation`, it's probably worth having a discussion first. I like the idea of streamlining the APIs in general though. I have added an issue to track this idea [here](https://github.com/grafana/k6/issues/4982) 👍

internal/js/modules/k6/browser/tests/page_test.go
ankur22: Yeah, i agree. I will refactor the test in a refactoring PR along with [this](https://github.com/grafana/k6/pull/4920#discussion_r2215746887) change i want to make.

Issue #4884, opened 2025-06-26
Add option for opt in to use community extensions


pablochacin:
Presently, the binary provisioning only allows building extensions tagged as part of the "cloud" catalog (soon to be renamed "official" extensions). This is mostly due to security reasons.

However, we want users (in particular, OSS users( to run tests using any extension in the [registry](https://github.com/grafana/k6-extension-registry) 

We need an option in `k6` CLI interface to allow users to opt in to using the community extensions using binary provisioning.

PR #4922, opened 2025-07-11
support community extensions

pablochacin:
## What?

Allow users to opt for using community extensions (defined in the `oss` catalog) with the  `K6_ENABLE_COMMUNITY_EXTENSIONS` environment variable. If not enabled, the default `cloud` catalog will be used.

## Why?

By default, binary provisioning only supports extensions defined in the `cloud` catalog.

We want to offer users the possibility of opting in for the complete catalog of extensions when running tests locally.

Note: if the user enables the community extensions and sends a test to the cloud using `cloud run`, the test will fail at execution time, as the cloud runners do not support these extensions.

## Checklist

## Related PR(s)/Issue(s)

Closes https://github.com/grafana/k6/issues/4884

### DIFF
--- a/internal/cmd/launcher_test.go
+++ b/internal/cmd/launcher_test.go
@@ -359,21 +359,21 @@ func TestGetBuildServiceURL(t *testing.T) {
 		buildSrvURL               string
 		enableCommunityExtensions bool
 		expectErr                 bool
-		expected                  string
+		expectedURL               string
 	}{
 		{
 			name:                      "default build service url",
 			buildSrvURL:               "https://build.srv",
 			enableCommunityExtensions: false,
 			expectErr:                 false,
-			expected:                  "https://build.srv/cloud",
+			expectedURL:               "https://build.srv/cloud",
 		},
 		{
 			name:                      "enable community extensions",
 			buildSrvURL:               "https://build.srv",
 			enableCommunityExtensions: true,
 			expectErr:                 false,
-			expected:                  "https://build.srv/oss",
+			expectedURL:               "https://build.srv/oss",
 		},
 		{
 			name:                      "invalid buildServiceURL",
@@ -402,7 +402,7 @@ func TestGetBuildServiceURL(t *testing.T) {
 				assert.Error(t, err)
 			} else {
 				assert.NoError(t, err)
-				assert.Equal(t, tc.expected, buildSrvURL)
+				assert.Equal(t, tc.expectedURL, buildSrvURL)
 			}
 		})
 	}


cmd/state/state.go
```diff
@@ -27,9 +27,15 @@ const (
...
 
-	defaultBuildServiceURL = "https://ingest.k6.io/builder/api/v1"
-	defaultConfigFileName  = "config.json"
-	defaultBinaryCacheDir  = "builds"
+	// DefaultBuildServiceURL defines the URL to the default (grafana hosted) build service
+	DefaultBuildServiceURL = "https://ingest.k6.io/builder/api/v1"
+	// CloudExtensionsCatalog defines the extensions catalog for cloud supported extensions
+	CloudExtensionsCatalog = "cloud"
+	// CommunityExtensionsCatalog defines the extensions catalog for community extensions
+	CommunityExtensionsCatalog = "oss"
```
codebien: Is it required to have them exported?

codebien (commented):
@pablochacin unit tests are missing here

cmd/state/state.go
pablochacin: They are referenced in `internal/cmd/launcher.go`

pablochacin:
@codebien 
>@pablochacin unit tests are missing here

What kind of tests are you thinking? The main impact of this changes are internal to the `provision` function (selecting the catalog to use), and this change is only visible to the `k6provider` library. I'm open to suggestions.

codebien:
>What kind of tests are you thinking?

We should assert the URL generated if the community extensions are enabled. So the new logic added.

internal/cmd/launcher.go
```diff
@@ -216,8 +217,23 @@ func newK6BuildProvisioner(gs *state.GlobalState) provisioner {
...
+	buildSrvURL, err := url.Parse(buildSrv)
+	if err != nil {
+		return nil, fmt.Errorf("invalid URL to binary provisioning build service: %w", err)
+	}
+
+	catalog := state.CloudExtensionsCatalog
+	if p.gs.Flags.EnableCommunityExtensions {
+		catalog = state.CommunityExtensionsCatalog
+	}
+	buildSrv = buildSrvURL.JoinPath(catalog).String()
```
codebien: We might consider moving this logic into a dedicated function that accepts as input a flagset and it returns the URL or an error.

In this way, it's easier to test.

internal/cmd/launcher_test.go
```diff
@@ -348,3 +350,64 @@ func TestIsAnalysisRequired(t *testing.T) {
...
+		enableCommunityExtensions bool
+		expectErr                 bool
+		expected                  string
+	}{
+		{
+			name:                      "default build service url",
+			buildSrvURL:               state.DefaultBuildServiceURL,
+			enableCommunityExtensions: false,
+			expectErr:                 false,
+			expected:                  fmt.Sprintf("%s/%s", state.DefaultBuildServiceURL, state.CloudExtensionsCatalog),
```
codebien: Let's use a static string here. For example, if we wrote a bad value in `state.CloudExtensionsCatalog` then we wouldn't be able to catch it. Because the assertion is depending on the past value.
```suggestion
			expected:                  "http://my-fake-exected-url/path-of-the-community,
```

cmd/state/state.go
joanlopez: But they aren't used here, no? 

Do we need to define and export them here? Just out of curiosity.

internal/cmd/launcher_test.go
joanlopez: Yeah, I'm generally in favor of the practice @codebien is suggesting here, so if anytime we change that value, it will be more explicit, and likely easier to catch potential breaking changes, etc 👍🏻

internal/cmd/launcher_test.go
```diff
@@ -348,3 +350,64 @@ func TestIsAnalysisRequired(t *testing.T) {
...
+				t.Fatalf("unexpected error %v", err)
+			}
+
+			if tc.expectErr && err == nil {
+				t.Fatalf("expected error got none")
+			}
+
+			if !tc.expectErr && buildSrvURL != tc.expected {
+				t.Fatalf("expected buildSrvURL %q got %q", tc.expected, buildSrvURL)
+			}
```
joanlopez: Nit; I think we generally use `assert` or `require` from [testify](https://github.com/stretchr/testify) 🤔

internal/cmd/launcher_test.go
```diff
@@ -348,3 +350,60 @@ func TestIsAnalysisRequired(t *testing.T) {
...
 }
+
+func TestGetBuildServiceURL(t *testing.T) {
+	t.Parallel()
+
+	testCases := []struct {
+		name                      string
+		buildSrvURL               string
+		enableCommunityExtensions bool
+		expectErr                 bool
```
codebien, line 361:
```suggestion
		expectErr                 error
```
Instead of using a boolean here, you can use directly an error. In this way you can remove the condition and directly assert via `require.Equal(t, tc.expectErr, err)`. When the error isn't expcted then it is just set to `nil` like `expectErr: nil`.

internal/cmd/launcher_test.go
```diff
@@ -348,3 +350,60 @@ func TestIsAnalysisRequired(t *testing.T) {
...
+		expected                  string
```
```suggestion
		expectURL                  string
```
To be consistent with the error field

internal/cmd/launcher_test.go
```diff
@@ -348,3 +350,60 @@ func TestIsAnalysisRequired(t *testing.T) {
...
+				EnableCommunityExtensions: tc.enableCommunityExtensions,
+			}
+
+			buildSrvURL, err := getBuildServiceURL(flags, logger)
+			if tc.expectErr {
+				assert.Error(t, err)
+			} else {
+				assert.NoError(t, err)
+				assert.Equal(t, tc.expected, buildSrvURL)
+			}
```
```suggestion
		require.Equal(t, tc.expectErr, err)
		assert.Equal(t, tc.expectURL, buildSrvURL)
```
This could be simplified like this

PR #4932, opened 2025-07-15
Update the logging around request abort

ankur22:
## What?

Making the logging around the request being aborted clearer.

## Why?

When a request is aborted due to either the hostname or ip address matching a block list, we should make it a little clearer what has happened. Using the word abort over interrupted lets the user know that the request will not be resumed.

## Checklist

### DIFF
--- a/internal/js/modules/k6/browser/tests/network_manager_test.go
+++ b/internal/js/modules/k6/browser/tests/network_manager_test.go
@@ -60,7 +60,7 @@ func TestBlockHostnames(t *testing.T) {
 	)
 	require.NoError(t, err)
 	require.Nil(t, res)
-	tb.logCache.assertContains(t, "was interrupted: hostname host.test is in a blocked pattern")
+	tb.logCache.assertContains(t, "was aborted: hostname host.test matches a blocked pattern")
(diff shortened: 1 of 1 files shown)


ankur22:
> I'm not sure the proposal is a concrete improvement. If we don't mention that we are going to resume a request, why someone should assume it?

I think this could be a subjective thing. When i read interrupt I think more from a runtime point of view, the execution has been interrupted, and it might be resumed later. I feel the word abort is more definite, it won't be started again. Happy to change it back or consider other ways of wording this messaging to better relay what is happening.

> Is the request started at that stage? If not, isn't blocked a better word for the case?

When we run the browser module in the cloud, we give it a black list of hosts and IPs. For the browser module to be able to actually block requests which match the black list, all requests that chromium makes are paused, and the browser module will need to either abort or continue the request depending on whether the request URL matches the black lists.

In this case we're aborting/interrupting the request because the hostname matches a pattern in the hostname blacklist, so we're blocking the request. There could be other reasons we're aborting/interrupting the request such as when we work with `page.route` and in that case we will be aborting/interrupting due to a user request.

PR #4940, opened 2025-07-17
Add support to set default Stack Cloud commands should target

dgzlopes:
## Checklist

### DIFF
--- a/cloudapi/api.go
+++ b/cloudapi/api.go
@@ -336,3 +336,33 @@ func (c *Client) ValidateToken() (*ValidateTokenResponse, error) {
 
 	return &vtr, nil
 }
+
+type accountMeRequest struct {
+}
+
+type Organization struct {
+	ID               int    `json:"id"`
+	GrafanaStackName string `json:"grafana_stack_name"`
+	GrafanaStackID   int    `json:"grafana_stack_id"`
+}
+
+type accountMeResponse struct {
+	Organizations []Organization `json:"organizations"`
+}
+
+// AccountMe retrieves the current user's account information.
+func (c *Client) AccountMe() (*accountMeResponse, error) {
+	// TODO: remove this hardcoded URL
+	req, err := c.NewRequest("GET", "https://api.k6.io/v3/account/me", accountMeRequest{})
+	if err != nil {
+		return nil, err
+	}
+
+	amr := accountMeResponse{}
+	err = c.Do(req, &amr)
+	if err != nil {
+		return nil, err
+	}
+
+	return &amr, nil
+}
--- a/cloudapi/config.go
+++ b/cloudapi/config.go
@@ -18,6 +18,8 @@ const LegacyCloudConfigKey = "loadimpact"
 //nolint:lll
 type Config struct {
 	// TODO: refactor common stuff between cloud execution and output
+	StackID   null.String `json:"stackID,omitempty" envconfig:"K6_CLOUD_STACK_ID"`
+	StackSlug null.String `json:"stackSlug,omitempty" envconfig:"K6_CLOUD_STACK_SLUG"`
 	Token     null.String `json:"token" envconfig:"K6_CLOUD_TOKEN"`
 	ProjectID null.Int    `json:"projectID" envconfig:"K6_CLOUD_PROJECT_ID"`
 	Name      null.String `json:"name" envconfig:"K6_CLOUD_NAME"`
@@ -165,6 +167,14 @@ func (c Config) Apply(cfg Config) Config {
 	}
 	if cfg.AggregationWaitPeriod.Valid {
 		c.AggregationWaitPeriod = cfg.AggregationWaitPeriod
+	}
+	if cfg.StackID.Valid {
+		c.StackID = cfg.StackID
+		c.StackSlug = null.StringFromPtr(nil)
+	}
+	if cfg.StackSlug.Valid {
+		c.StackSlug = cfg.StackSlug
+		c.StackID = null.StringFromPtr(nil)
 	}
 	return c
 }
--- a/internal/cmd/cloud_login.go
+++ b/internal/cmd/cloud_login.go
@@ -69,9 +69,11 @@ the "k6 run -o cloud" command.
 
 // run is the code that runs when the user executes `k6 cloud login`
 func (c *cmdCloudLogin) run(cmd *cobra.Command, _ []string) error {
-	err := migrateLegacyConfigFileIfAny(c.globalState)
-	if err != nil {
-		return err
+	if !checkIfMigrationCompleted(c.globalState) {
+		err := migrateLegacyConfigFileIfAny(c.globalState)
+		if err != nil {
+			return err
+		}
 	}
 
 	currentDiskConf, err := readDiskConfig(c.globalState)
--- a/internal/cmd/cloud_login.go
+++ b/internal/cmd/cloud_login.go
@@ -159,33 +159,41 @@ func (c *cmdCloudLogin) run(cmd *cobra.Command, _ []string) error {
 			}
 		}
 
-		defaultStack := func(token string) string {
-			return "slug"
-		}(tokenValue)
+		defaultStack, err := func(token string) (string, error) {
+			stacks, err := getStacks(c.globalState, currentJSONConfigRaw, token)
+			if err != nil {
+				return "", fmt.Errorf("failed to get default stack slug: %w", err)
+			}
 
-		slugForm := ui.Form{
+			// TODO: Can we make this better? Picking the first one is not ideal.
+			for slug := range stacks {
+				return slug, nil
+			}
+			return "", errors.New("no stacks found for the provided token, please create a stack in Grafana Cloud and initialize GCk6 app")
+		}(tokenValue)
+		if err != nil {
+			return fmt.Errorf("failed to get default stack: %w", err)
+		}
+
+		stackForm := ui.Form{
 			Banner: "\nConfigure the stack where your tests will run by default.\n" +
-				"Please, consult the Grafana Cloud k6 documentation for instructions on how to find your stack slug:\n" +
-				// TODO: Update the link when the documentation is ready
-				"FIXME",
+				"Please, use the slug from your Grafana Cloud URL, e.g. my-team from https://my-team.grafana.net.\n",
 			Fields: []ui.Field{
 				ui.StringField{
-					Key:     "StackSlug",
+					Key:     "Stack",
 					Label:   "Stack",
 					Default: defaultStack,
 				},
 			},
 		}
-		slugVals, err := slugForm.Run(c.globalState.Stdin, c.globalState.Stdout)
+		stackVals, err := stackForm.Run(c.globalState.Stdin, c.globalState.Stdout)
 		if err != nil {
 			return err
 		}
 
-		stackValue := slugVals["StackSlug"]
-		stackSlugInput := null.StringFrom(stackValue)
-
-		if stackSlugInput.Valid {
-			stackSlug = stackSlugInput
+		stackValue := null.StringFrom(stackVals["Stack"])
+		if stackValue.Valid {
+			stackSlug = stackValue
 		} else {
 			return errors.New("stack cannot be empty")
 		}
@@ -273,6 +281,39 @@ func validateToken(gs *state.GlobalState, jsonRawConf json.RawMessage, token str
 	}
 
 	return nil
+}
+
+func getStacks(gs *state.GlobalState, jsonRawConf json.RawMessage, token string) (map[string]int, error) {
+	// We want to use this fully consolidated config for things like
+	// host addresses, so users can overwrite them with env vars.
+	consolidatedCurrentConfig, warn, err := cloudapi.GetConsolidatedConfig(
+		jsonRawConf, gs.Env, "", nil, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	if warn != "" {
+		gs.Logger.Warn(warn)
+	}
+	client := cloudapi.NewClient(
+		gs.Logger,
+		token,
+		"",
+		build.Version,
+		consolidatedCurrentConfig.Timeout.TimeDuration(),
+	)
+
+	res, err := client.AccountMe()
+	if err != nil {
+		return nil, fmt.Errorf("can't get account info: %s", err.Error())
+	}
+
+	stacks := make(map[string]int)
+	for _, organization := range res.Organizations {
+		stacks[organization.GrafanaStackName] = organization.GrafanaStackID
+	}
+
+	return stacks, nil
 }
 
 func parseStackID(val string) (int64, error) {
--- a/internal/cmd/cloud_login.go
+++ b/internal/cmd/cloud_login.go
@@ -4,7 +4,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"strconv"
 	"syscall"
 
 	"github.com/fatih/color"
@@ -31,16 +30,19 @@ func getCmdCloudLogin(gs *state.GlobalState) *cobra.Command {
 
 	// loginCloudCommand represents the 'cloud login' command
 	exampleText := getExampleText(gs, `
-  # Prompt for a Grafana Cloud k6 token
+  # Authenticate interactively with Grafana Cloud k6
   $ {{.}} cloud login
 
   # Store a token in k6's persistent configuration
   $ {{.}} cloud login -t <YOUR_TOKEN>
 
-  # Display the stored token
+  # Store a token in k6's persistent configuration and set the stack slug
+  $ {{.}} cloud login -t <YOUR_TOKEN> -s <YOUR_STACK_SLUG>
+
+  # Display the stored token and stack info
   $ {{.}} cloud login -s
 
-  # Reset the stored token
+  # Reset the stored token and stack info
   $ {{.}} cloud login -r`[1:])
 
 	loginCloudCommand := &cobra.Command{
@@ -59,10 +61,9 @@ the "k6 run -o cloud" command.
 	}
 
 	loginCloudCommand.Flags().StringP("token", "t", "", "specify `token` to use")
-	loginCloudCommand.Flags().BoolP("show", "s", false, "display saved token and exit")
+	loginCloudCommand.Flags().BoolP("show", "s", false, "display saved token, stack info and exit")
 	loginCloudCommand.Flags().BoolP("reset", "r", false, "reset stored token and stack info")
-	loginCloudCommand.Flags().String("stack-id", "", "set stack ID (cannot be used with --stack-slug)")
-	loginCloudCommand.Flags().String("stack-slug", "", "set stack slug (cannot be used with --stack-id)")
+	loginCloudCommand.Flags().String("stack-slug", "", "specify the stack where commands will run by default")
 
 	return loginCloudCommand
 }
@@ -97,22 +98,7 @@ func (c *cmdCloudLogin) run(cmd *cobra.Command, _ []string) error {
 	show := getNullBool(cmd.Flags(), "show")
 	reset := getNullBool(cmd.Flags(), "reset")
 	token := getNullString(cmd.Flags(), "token")
-	stackIDStr := cmd.Flags().Lookup("stack-id").Value.String()
 	stackSlug := getNullString(cmd.Flags(), "stack-slug")
-
-	var stackIDInt int64
-	var stackIDValid bool
-	if stackIDStr != "" {
-		var parseErr error
-		stackIDInt, parseErr = parseStackID(stackIDStr)
-		if parseErr != nil {
-			return fmt.Errorf("invalid --stack-id: %w", parseErr)
-		}
-		stackIDValid = true
-	}
-	if stackIDValid && stackSlug.Valid {
-		return errors.New("only one of --stack-id or --stack-slug can be specified")
-	}
 
 	switch {
 	case reset.Valid:
@@ -124,15 +110,26 @@ func (c *cmdCloudLogin) run(cmd *cobra.Command, _ []string) error {
 	case show.Bool:
 		valueColor := getColor(c.globalState.Flags.NoColor || !c.globalState.Stdout.IsTTY, color.FgCyan)
 		printToStdout(c.globalState, fmt.Sprintf("  token: %s\n", valueColor.Sprint(newCloudConf.Token.String)))
-		if newCloudConf.StackID.Valid {
+		if !newCloudConf.StackID.Valid && !newCloudConf.StackSlug.Valid {
+			printToStdout(c.globalState, "  stack-id: <not set>\n")
+			printToStdout(c.globalState, "  stack-slug: <not set>\n")
+		} else {
 			printToStdout(c.globalState, fmt.Sprintf("  stack-id: %s\n", valueColor.Sprint(newCloudConf.StackID.String)))
-		}
-		if newCloudConf.StackSlug.Valid {
 			printToStdout(c.globalState, fmt.Sprintf("  stack-slug: %s\n", valueColor.Sprint(newCloudConf.StackSlug.String)))
 		}
 		return nil
 	case token.Valid:
 		newCloudConf.Token = token
+		if stackSlug.Valid {
+			normalizedSlug := stripGrafanaNetSuffix(stackSlug.String)
+			newCloudConf.StackSlug = null.StringFrom(normalizedSlug)
+			id, err := resolveStackSlugToID(c.globalState, currentJSONConfigRaw, token.String, normalizedSlug)
+			if err == nil {
+				newCloudConf.StackID = null.StringFrom(fmt.Sprintf("%d", id))
+			} else {
+				return fmt.Errorf("could not resolve stack slug. Are you sure the slug is correct? %w", err)
+			}
+		}
 	default:
 		tokenForm := ui.Form{
 			Banner: "Enter your token to authenticate with Grafana Cloud k6.\n" +
@@ -155,10 +152,11 @@ func (c *cmdCloudLogin) run(cmd *cobra.Command, _ []string) error {
 		tokenValue := tokenVals["Token"]
 		newCloudConf.Token = null.StringFrom(tokenValue)
 
-		if newCloudConf.Token.Valid {
-			if err := validateToken(c.globalState, currentJSONConfigRaw, newCloudConf.Token.String); err != nil {
-				return fmt.Errorf("token validation failed: %w", err)
-			}
+		if !newCloudConf.Token.Valid {
+			return errors.New("token cannot be empty")
+		}
+		if err := validateToken(c.globalState, currentJSONConfigRaw, newCloudConf.Token.String); err != nil {
+			return fmt.Errorf("token validation failed: %w", err)
 		}
 
 		defaultStack, err := func(token string) (string, error) {
@@ -166,8 +164,6 @@ func (c *cmdCloudLogin) run(cmd *cobra.Command, _ []string) error {
 			if err != nil {
 				return "", fmt.Errorf("failed to get default stack slug: %w", err)
 			}
-
-			// TODO: Can we make this better? Picking the first one is not ideal.
 			for slug := range stacks {
 				return slug, nil
 			}
@@ -178,11 +174,11 @@ func (c *cmdCloudLogin) run(cmd *cobra.Command, _ []string) error {
 		}
 
 		stackForm := ui.Form{
-			Banner: "\nConfigure the stack where your tests will run by default.\n" +
-				"Please, use the slug from your Grafana Cloud URL, e.g. my-team from https://my-team.grafana.net.\n",
+			Banner: "\nEnter the stack where you want to run k6's commands by default.\n" +
+				"Use the slug from your Grafana Cloud URL, e.g. my-team from https://my-team.grafana.net):",
 			Fields: []ui.Field{
 				ui.StringField{
-					Key:     "Stack",
+					Key:     "StackSlug",
 					Label:   "Stack",
 					Default: defaultStack,
 				},
@@ -193,23 +189,18 @@ func (c *cmdCloudLogin) run(cmd *cobra.Command, _ []string) error {
 			return err
 		}
 
-		stackValue := null.StringFrom(stackVals["Stack"])
-		if stackValue.Valid {
-			stackSlug = stackValue
-		} else {
+		stackSlugInput := null.StringFrom(stackVals["StackSlug"])
+		if !stackSlugInput.Valid {
 			return errors.New("stack cannot be empty")
 		}
-	}
+		normalizedSlug := stripGrafanaNetSuffix(stackSlugInput.String)
+		newCloudConf.StackSlug = null.StringFrom(normalizedSlug)
 
-	if stackIDValid {
-		newCloudConf.StackID = null.StringFrom(fmt.Sprintf("%d", stackIDInt))
-		newCloudConf.StackSlug = null.StringFromPtr(nil)
-	} else if stackSlug.Valid {
-		newCloudConf.StackSlug = stackSlug
-		newCloudConf.StackID = null.StringFromPtr(nil)
-	} else {
-		newCloudConf.StackID = null.StringFromPtr(nil)
-		newCloudConf.StackSlug = null.StringFromPtr(nil)
+		id, err := resolveStackSlugToID(c.globalState, currentJSONConfigRaw, tokenValue, normalizedSlug)
+		if err != nil {
+			return fmt.Errorf("could not resolve stack slug. Are you sure the slug is correct? %w", err)
+		}
+		newCloudConf.StackID = null.StringFrom(fmt.Sprintf("%d", id))
 	}
 
 	if newCloudConf.Token.Valid {
@@ -285,6 +276,14 @@ func validateToken(gs *state.GlobalState, jsonRawConf json.RawMessage, token str
 	return nil
 }
 
+func stripGrafanaNetSuffix(s string) string {
+	const suffix = ".grafana.net"
+	if len(s) > len(suffix) && s[len(s)-len(suffix):] == suffix {
+		return s[:len(s)-len(suffix)]
+	}
+	return s
+}
+
 func getStacks(gs *state.GlobalState, jsonRawConf json.RawMessage, token string) (map[string]int, error) {
 	// We want to use this fully consolidated config for things like
 	// host addresses, so users can overwrite them with env vars.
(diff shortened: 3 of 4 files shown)


cloudapi/api.go
```diff
@@ -336,3 +338,95 @@ func (c *Client) ValidateToken() (*ValidateTokenResponse, error) {
...
+	err = c.Do(req, &amr)
+	if err != nil {
+		return nil, err
+	}
+
+	return &amr, nil
+}
+
+// GetDefaultProject retrieves the default project for the given stack ID.
+func (c *Client) GetDefaultProject(stackID int64) (int64, string, error) {
```
dgzlopes, line 372: I don't know what the long-term plans are here.

The [new REST API](https://grafana.com/docs/grafana-cloud/testing/k6/reference/cloud-rest-api/) does a few things differently:
- StackID is mandatory
- Auth header is different (using Bearer instead of Token)

Should we change the existing Client, or add a new one? 

What do you think about using the [auto-generated Go SDK](https://github.com/grafana/k6-cloud-openapi-client-go)?

Moving to the new API completely is impossible until we enforce StackID (maybe in v2?). So yeah, we will be living in this mixed APIs state for a while.

internal/cmd/cloud_login.go
```diff
@@ -58,17 +61,22 @@ the "k6 run -o cloud" command.
...
+//nolint:funlen,gocognit,cyclop
 func (c *cmdCloudLogin) run(cmd *cobra.Command, _ []string) error {
-	err := migrateLegacyConfigFileIfAny(c.globalState)
-	if err != nil {
-		return err
+	if !checkIfMigrationCompleted(c.globalState) {
+		err := migrateLegacyConfigFileIfAny(c.globalState)
+		if err != nil {
+			return err
+		}
```
dgzlopes, line 80: I tackled this on this PR because it was messing up how stackID/Slug are persisted.

internal/cmd/common.go
```diff
@@ -127,3 +129,62 @@ func handleTestAbortSignals(gs *state.GlobalState, gracefulStopHandler, onHardSt
 		gs.SignalStop(sigC)
 	}
 }
+
```
dgzlopes, line 132: I'm not really sure if `common.go` is the right place for these shared functions. 

Happy to move it around.

internal/cmd/cloud_login.go
```diff
@@ -117,12 +147,63 @@ func (c *cmdCloudLogin) run(cmd *cobra.Command, _ []string) error {
...
+
+		defaultStack, err := func(token string) (string, error) {
+			stacks, err := getStacks(c.globalState, currentJSONConfigRaw, token)
+			if err != nil {
+				return "", fmt.Errorf("failed to get default stack slug: %w", err)
+			}
+			for slug := range stacks {
+				return slug, nil
+			}
```
dgzlopes: I'm not very happy with this approach for selecting the "default" stack...

But yeah, I would love to hear suggestions!

internal/cmd/cloud_login.go
```diff
@@ -92,23 +100,52 @@ func (c *cmdCloudLogin) run(cmd *cobra.Command, _ []string) error {
...
 	default:
-		form := ui.Form{
+		/* Token form */
+		tokenForm := ui.Form{
 			Banner: "Enter your token to authenticate with Grafana Cloud k6.\n" +
 				"Please, consult the Grafana Cloud k6 documentation for instructions on how to generate one:\n" +
 				"https://grafana.com/docs/grafana-cloud/testing/k6/author-run/tokens-and-cli-authentication",
 			Fields: []ui.Field{
-				ui.PasswordField{
+				ui.StringField{
```
dgzlopes: While testing the feature, I never knew if the token was pasted properly. 

I feel this is a saner default.

internal/cmd/cloud_login.go
```diff
@@ -186,3 +261,58 @@ func validateToken(gs *state.GlobalState, jsonRawConf json.RawMessage, token str
...
+	return stacks, nil
+}
+
+func getDefaultStack(gs *state.GlobalState, jsonRawConf json.RawMessage, token string) (int64, string, error) {
+	stacks, err := getStacks(gs, jsonRawConf, token)
+	if err != nil {
+		// TODO: Improve handling of this error. This happens when the user uses a Stack tocken instead of a Personal token.
+		if strings.Contains(err.Error(), "Authentication failed") {
+			return 0, "", nil
+		}
```
dgzlopes, line 306: Do we have an API to check if a token is personal or stack? If stack, from which stack?

It would be nice to be able to set it as a default and to improve this error handling.

cloudapi/config.go
```diff
@@ -166,6 +168,14 @@ func (c Config) Apply(cfg Config) Config {
...
 		c.AggregationWaitPeriod = cfg.AggregationWaitPeriod
 	}
+	if cfg.StackID.Valid {
+		c.StackID = cfg.StackID
+		c.StackSlug = null.StringFromPtr(nil)
+	}
+	if cfg.StackSlug.Valid {
+		c.StackSlug = cfg.StackSlug
+		c.StackID = null.StringFromPtr(nil)
+	}
```
ankur22: Shamefully, i don't know the difference between `StackID` and `StackSlug`. Looks like we can't work with both here. If someone did provide both, could k6 return an error (or log a warning)? The user might want to work with `StackID` and instead it gets overwritten by `StackSlug` as they forgot to undefine `StackSlug`.

internal/cmd/cloud_login.go
```diff
@@ -117,12 +146,60 @@ func (c *cmdCloudLogin) run(cmd *cobra.Command, _ []string) error {
...
 			c.globalState.Logger.Warn("Stdin is not a terminal, falling back to plain text input")
 		}
-		var vals map[string]string
-		vals, err = form.Run(c.globalState.Stdin, c.globalState.Stdout)
+		tokenVals, err := tokenForm.Run(c.globalState.Stdin, c.globalState.Stdout)
 		if err != nil {
 			return err
 		}
-		newCloudConf.Token = null.StringFrom(vals["Token"])
+		tokenValue := tokenVals["Token"]
```
ankur22, line 162: Random thought -- If we have the token, couldn't k6 fetch the slug for the user by hitting a backend endpoint?

internal/cmd/cloud_login.go
ankur22: Nit: I feel like this doesn't belong in this PR since it doesn't help the problem set out in the issue description.

internal/cmd/cloud_login.go
```diff
@@ -92,23 +95,49 @@ func (c *cmdCloudLogin) run(cmd *cobra.Command, _ []string) error {
...
 	reset := getNullBool(cmd.Flags(), "reset")
 	token := getNullString(cmd.Flags(), "token")
+	stackIDStr := cmd.Flags().Lookup("stack-id").Value.String()
+	stackSlug := getNullString(cmd.Flags(), "stack-slug")
+
+	var stackIDInt int64
+	var stackIDValid bool
+	if stackIDStr != "" {
+		var parseErr error
+		stackIDInt, parseErr = parseStackID(stackIDStr)
```
ankur22: Nit: I think this is an unnecessary abstraction (`parseStackID`).

internal/cmd/cloud_login.go
ankur22: Would be good if there was a default flag on the actual default stack.

Could we display them all, clearly displaying to them which one we will pick as default, and allow the user to pick a different one?

```
Pick a stack: (will default to stack1)

stack1        < 
stack2
coolStack
```

internal/cmd/cloud.go
```diff
@@ -183,6 +183,8 @@ func (c *cmdCloud) run(cmd *cobra.Command, args []string) error {
 		arc.Options.External[cloudapi.LegacyCloudConfigKey] = b
 
 		cloudConfig.ProjectID = null.IntFrom(projectID)
+
+		logger.Warn("Warning: no projectID specified, using default project of the stack: " + cloudConfig.StackSlug.String)
```
ankur22: At the moment i read is as "we're using the default project of the stack which is "blah""

WDYT of?:

```suggestion
		logger.Warn("Warning: no projectID specified, using default project of the '" + cloudConfig.StackSlug.String + "' stack")
```

Also, would it be possible to specify the name of the default project here?

internal/cmd/cloud.go
```diff
@@ -168,6 +168,11 @@ func (c *cmdCloud) run(cmd *cobra.Command, args []string) error {
 		logger, cloudConfig.Token.String, cloudConfig.Host.String, build.Version, cloudConfig.Timeout.TimeDuration())
 
 	if cloudConfig.ProjectID.Int64 == 0 {
+		if cloudConfig.StackID.Int64 == 0 {
+			logger.Error("please specify a projectID in your test or use `k6 cloud login` to set up a default stack")
```
ankur22: I'm confused here -- even if the user specifies the `stackID`, the `projectID` still needs to be specified, correct? If so, maybe we need to be clearer here, something along the lines of:

```suggestion
		logger.Error("please specify a projectID in your test or use `k6 cloud login` to set up a default stack with a default project")
```

This is based on the previous commit where there's a warning "no projectID is specified using the default project of the stack"

internal/cmd/common.go
```diff
@@ -127,3 +130,61 @@ func handleTestAbortSignals(gs *state.GlobalState, gracefulStopHandler, onHardSt
...
+func resolveStackSlugToID(gs *state.GlobalState, jsonRawConf json.RawMessage, token, slug string) (int64, error) {
+	slug = stripGrafanaNetSuffix(slug)
+	stacks, err := getStacks(gs, jsonRawConf, token)
+	if err != nil {
+		return 0, err
+	}
+	id, ok := stacks[slug]
+	if !ok {
+		return 0, fmt.Errorf("stack slug %q not found in your Grafana Cloud account", slug)
```
ankur22, line 141: Should we list all the available stacks in the error message so that the user knows which ones are available?

internal/cmd/cloud_login.go
```diff
@@ -223,20 +224,18 @@ func (c *cmdCloudLogin) run(cmd *cobra.Command, _ []string) error {
...
-			}
-			if newCloudConf.StackSlug.Valid {
-				printToStdout(c.globalState, fmt.Sprintf("  stack-slug: %s\n", valueColor.Sprint(newCloudConf.StackSlug.String)))
-			}
+	valueColor := getColor(c.globalState.Flags.NoColor || !c.globalState.Stdout.IsTTY, color.FgCyan)
+	printToStdout(c.globalState, fmt.Sprintf(
+		"\nLogged in successfully, token and stack info saved in %s\n", c.globalState.Flags.ConfigFilePath,
+	))
+	if !c.globalState.Flags.Quiet {
+		printToStdout(c.globalState, fmt.Sprintf("  token: %s\n", valueColor.Sprint(newCloudConf.Token.String)))
```
ankur22: You removed the check to see if `newCloudConf.Token` was valid first before printing this. Do we need to early exist when `newCloudConf.Token.Valid` is `false`?

internal/cmd/cloud_login.go
```diff
@@ -309,3 +293,20 @@ func getStacks(gs *state.GlobalState, jsonRawConf json.RawMessage, token string)
...
+
+	if len(stacks) == 0 {
+		return 0, "", errors.New("no stacks found for the provided token " +
+			"please create a stack in Grafana Cloud and initialize GCk6 app")
+	}
+
+	for slug, id := range stacks {
+		return int64(id), slug, nil
+	}
+	return 0, "", nil // This should never happen
```
ankur22: Probably should return an error, even if it should never happen, right?

ankur22 (approved):
Looks good. Some minor things to consider.

I'm not fully versed on this area of k6 though so someone else who has a better idea might be a better person to add their approval too.

internal/cmd/cloud.go
dgzlopes: > even if the user specifies the stackID, the projectID still needs to be specified, correct?

Each stack has a default project. So knowing the stackID is enough.

internal/cmd/cloud_login.go
dgzlopes: A user can be in multiple stacks. So for personal tokens... we can't.

We could do it for stack tokens, though; that's a 1-to-1 relationship.

internal/cmd/cloud_login.go
dgzlopes: This would be amazing. I agree. I decided not to do this initially to not complicate things even more (adding a new thing to the UI package, etc). Wdyt putting this in a follow-up UX PR? (with the change on how we handle the Token, and a couple more things).

internal/cmd/cloud_login.go
dgzlopes: This is gone in a future commit. Sorry for the messy history, I changed things around a few times..

internal/cmd/cloud_login.go
dgzlopes: Fair! I'll put this in a follow-up PR. Removed.

cloudapi/config.go
dgzlopes: StackID is a number. StackSlug is a string. You can spot it easily on the URL of your Grafana Cloud instance.

e.g. <slug>.grafana.com

StackIDs can't change. StackSlugs only change on edge cases. 

IDs are pretty hard to find, while Slugs are well-known. I wanted to focus the UX on Slugs because they seem way easier to understand, find, and reference. But yeah, this is quite confusing, I know :smile:

internal/cmd/cloud_login.go
dgzlopes: This is gone in a future commit. Sorry for the messy history. I changed things around a few times.

internal/cmd/cloud.go
dgzlopes: Makes sense. Fixed!

> Also, would it be possible to specify the name of the default project here?

I did that... but bc the name is usually "Default project" it was a bit weird :smile:

internal/cmd/common.go
dgzlopes: I tried this, but bc I'm in ~60 stacks... the output was terrible. I guess the average user is in a couple of stacks.

internal/cmd/cloud_login.go
dgzlopes: > You removed the check

I added it back. I think it is not needed bc of the logic inside the switch, but yeah, better to add the safeguard in case we modify that code in the future.

internal/cmd/cloud_login.go
dgzlopes: True! Done.

internal/cmd/common.go
dgzlopes: I think it is "OK" to keep it as it is. At least it is consistent with how we handle wrong projectIDs.

cloudapi/api.go
codebien: >until we enforce StackID (maybe in v2?).

Is it really a strict requirement? It seems we can get the StackID from the token just by calling `/account/me`, or is only available in `/v3`?

>What do you think about using the auto-generated Go SDK?

It sounds to me a great idea to explore.

cloudapi/api.go
```diff
@@ -336,3 +338,95 @@ func (c *Client) ValidateToken() (*ValidateTokenResponse, error) {
...
+
+// AccountMeResponse represents the response of the /account/me endpoint.
+type AccountMeResponse struct {
+	Organizations []Organization `json:"organizations"`
+}
+
+// AccountMe retrieves the current user's account information.
+func (c *Client) AccountMe() (*AccountMeResponse, error) {
+	// TODO: remove this hardcoded URL
+	req, err := c.NewRequest("GET", "https://api.k6.io/v3/account/me", nil)
```
codebien, line 357:
```suggestion
	req, err := c.NewRequest(http.GetMethod, "https://api.k6.io/v3/account/me", nil)
```
Same for the others

cloudapi/api.go
```diff
@@ -336,3 +338,95 @@ func (c *Client) ValidateToken() (*ValidateTokenResponse, error) {
...
+	}
+
+	q := req.URL.Query()
+	q.Add("$orderby", "created")
+	req.URL.RawQuery = q.Encode()
+
+	req.Header.Set("X-Stack-Id", fmt.Sprintf("%d", stackID))
+	// TODO: by default the client uses Token instead of Bearer
+	req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", c.token))
+	req.Header.Set("User-Agent", "Go-http-client")
```
codebien, line 386:
```suggestion
```
Is it required?

cloudapi/api.go
```diff
@@ -336,3 +338,95 @@ func (c *Client) ValidateToken() (*ValidateTokenResponse, error) {
...
+
+	// TODO: Can't use c.Do bc it messes up the headers
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
```
codebien, line 390: Then we need a dedicated client overwriting this different logic, otherwise testing will be quite difficult

cloudapi/api.go
```diff
@@ -336,3 +338,95 @@ func (c *Client) ValidateToken() (*ValidateTokenResponse, error) {
...
+			FolderUID string `json:"grafana_folder_uid"`
+		} `json:"value"`
+	}
+
+	if err := json.NewDecoder(resp.Body).Decode(&parsed); err != nil {
+		return 0, "", fmt.Errorf("failed to decode response: %w", err)
+	}
+
+	if len(parsed.Value) == 0 {
+		return 0, "", fmt.Errorf("no projects found for stack ID %d", stackID)
```
codebien, line 422:
```suggestion
		return 0, "", fmt.Errorf("no projects found for stack ID %d", stackSlug)
```
We should be consistent, if we prefer the Slug on input then we should return a slug on errors. Otherwise, they will experience the same difficulties we're carefully avoiding on inputs.

internal/cmd/cloud_login.go
```diff
@@ -30,16 +31,19 @@ func getCmdCloudLogin(gs *state.GlobalState) *cobra.Command {
...
 
-  # Display the stored token
+  # Store a token in k6's persistent configuration and set the stack slug
+  $ {{.}} cloud login -t <YOUR_TOKEN> -s <YOUR_STACK_SLUG>
+
+  # Display the stored token and stack info
   $ {{.}} cloud login -s
 
-  # Reset the stored token
+  # Reset the stored token and stack info
```
codebien, line 46:
```suggestion
  # Reset the account's stored info
```

internal/cmd/cloud_login.go
```diff
@@ -58,17 +62,22 @@ the "k6 run -o cloud" command.
 	}
 
 	loginCloudCommand.Flags().StringP("token", "t", "", "specify `token` to use")
-	loginCloudCommand.Flags().BoolP("show", "s", false, "display saved token and exit")
-	loginCloudCommand.Flags().BoolP("reset", "r", false, "reset stored token")
+	loginCloudCommand.Flags().BoolP("show", "s", false, "display saved token, stack info and exit")
```
codebien, line 65:
```suggestion
	loginCloudCommand.Flags().BoolP("show", "s", false, "display the saved account info and exit")
```

internal/cmd/outputs_cloud.go
```diff
@@ -118,6 +118,23 @@ func createCloudTest(gs *state.GlobalState, test *loadedAndConfiguredTest) error
...
 		logger, conf.Token.String, conf.Host.String, build.Version, conf.Timeout.TimeDuration())
 
+	if testRun.ProjectID == 0 {
+		projectID, err := resolveDefaultProjectID(
+			gs,
+			apiClient,
+			test.derivedConfig.Collectors["cloud"],
+			conf.Token.String,
+			conf.StackSlug,
+			&conf.StackID,
```
codebien, line 128:
```suggestion
			conf.StackSlug.String,
			&conf.StackID.Int64,
```
I guess it should be this the correct form. Isn't returning an error? 🤔

internal/cmd/common.go
```diff
@@ -127,3 +129,62 @@ func handleTestAbortSignals(gs *state.GlobalState, gracefulStopHandler, onHardSt
...
+	return s
+}
+
+func resolveDefaultProjectID(
+	gs *state.GlobalState,
+	apiClient *cloudapi.Client,
+	jsonRawConf json.RawMessage,
+	token string,
+	stackSlug null.String,
+	stackID *null.Int,
```
codebien, line 160:
```suggestion
	stackSlug string,
	stackID int,
```
Let's use simpler form without the `null` dependency

internal/cmd/config.go
```diff
@@ -404,3 +404,16 @@ func migrateLegacyConfigFileIfAny(gs *state.GlobalState) error {
 	}
 	return nil
 }
+
+// checkIfMigrationCompleted checks if the migration has been done by checking if the new file exists).
```
codebien, line 408:
```suggestion
// checkIfMigrationCompleted checks if the migration has been done by checking if the new file exists.
```

internal/cmd/cloud_login.go
codebien: Can you explain the issue that you were getting, please? I want to be sure we aren't hiding any bug in the migration's logic.

internal/cmd/common.go
```diff
@@ -127,3 +129,62 @@ func handleTestAbortSignals(gs *state.GlobalState, gracefulStopHandler, onHardSt
...
+		return 0, err
+	}
+	id, ok := stacks[slug]
+	if !ok {
+		return 0, fmt.Errorf("stack slug %q not found in your Grafana Cloud account", slug)
+	}
+	return int64(id), nil
+}
+
+func stripGrafanaNetSuffix(s string) string {
```
codebien, line 146:
```suggestion
func slugFromGrafanaURL(s string) string {
```

cloudapi/api.go
```diff
@@ -336,3 +338,95 @@ func (c *Client) ValidateToken() (*ValidateTokenResponse, error) {
...
+}
+
+// AccountMeResponse represents the response of the /account/me endpoint.
+type AccountMeResponse struct {
+	Organizations []Organization `json:"organizations"`
+}
+
+// AccountMe retrieves the current user's account information.
+func (c *Client) AccountMe() (*AccountMeResponse, error) {
+	// TODO: remove this hardcoded URL
```
codebien, line 356: What is the problem with this? Can we create a dedicated client?

cloudapi/api.go
```diff
@@ -336,3 +338,95 @@ func (c *Client) ValidateToken() (*ValidateTokenResponse, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	return &amr, nil
+}
+
+// GetDefaultProject retrieves the default project for the given stack ID.
+func (c *Client) GetDefaultProject(stackID int64) (int64, string, error) {
+	// TODO: remove this hardcoded URL
```
codebien, line 373: Same as before

cloudapi/api.go
```diff
@@ -336,3 +338,95 @@ func (c *Client) ValidateToken() (*ValidateTokenResponse, error) {
...
+		}
+	}()
+
+	if resp.StatusCode != http.StatusOK {
+		body, _ := io.ReadAll(resp.Body)
+		return 0, "", fmt.Errorf("unexpected status %d: %s", resp.StatusCode, string(body))
+	}
+
+	var parsed struct {
+		Count int64 `json:"@count"`
```
codebien, line 408:
```suggestion
```
Do we need to parse it?

cloudapi/config.go
```diff
@@ -240,7 +248,8 @@ func mergeFromCloudOptionAndExternal(
 	if err := json.Unmarshal(source, &tmpConfig); err != nil {
 		return err
 	}
-	// Only take out the ProjectID, Name and Token from the options.cloud (or legacy loadimpact struct) map:
+
+	// Only take out the ProjectID, Name, Token and StackID from the options.cloud (or legacy loadimpact struct) map:
```
codebien, line 252:
```suggestion
	// Only take out the ProjectID, Name, Token, StackID, StackSlug from the options.cloud (or legacy loadimpact struct) map:
```

internal/cmd/cloud_login.go
```diff
@@ -92,18 +101,47 @@ func (c *cmdCloudLogin) run(cmd *cobra.Command, _ []string) error {
...
+		newCloudConf.StackID = null.IntFromPtr(nil)
+		newCloudConf.StackSlug = null.StringFromPtr(nil)
+		printToStdout(c.globalState, "  token and stack info reset\n")
+		return nil
 	case show.Bool:
 		valueColor := getColor(c.globalState.Flags.NoColor || !c.globalState.Stdout.IsTTY, color.FgCyan)
 		printToStdout(c.globalState, fmt.Sprintf("  token: %s\n", valueColor.Sprint(newCloudConf.Token.String)))
+		if !newCloudConf.StackID.Valid && !newCloudConf.StackSlug.Valid {
+			printToStdout(c.globalState, "  stack-id: <not set>\n")
+			printToStdout(c.globalState, "  stack-slug: <not set>\n")
```
codebien, line 118: Can we find a better text for it? I feel it might create confusion, as it sounds a programmatic error.

If we can't find a better text to improve the UX then can we omit them for this case?

internal/cmd/cloud_login.go
```diff
@@ -92,18 +101,47 @@ func (c *cmdCloudLogin) run(cmd *cobra.Command, _ []string) error {
...
+		} else {
+			printToStdout(c.globalState, fmt.Sprintf("  stack-id: %s\n", valueColor.Sprint(newCloudConf.StackID.Int64)))
+			printToStdout(c.globalState, fmt.Sprintf("  stack-slug: %s\n", valueColor.Sprint(newCloudConf.StackSlug.String)))
```
codebien, line 121:
```suggestion
			printToStdout(c.globalState, fmt.Sprintf("  stack: %s\n", valueColor.Sprint(newCloudConf.StackSlug.String)))
```
Do we really need to print the ID?

internal/cmd/cloud_login.go
```diff
@@ -186,3 +263,61 @@ func validateToken(gs *state.GlobalState, jsonRawConf json.RawMessage, token str
...
+			"please create a stack in Grafana Cloud and initialize GCk6 app")
+	}
+
+	for slug, id := range stacks {
+		return int64(id), slug, nil
+	}
+
+	// This should never happen
+	return 0, "", errors.New("no stacks found for the provided token " +
+		"please create a stack in Grafana Cloud and initialize GCk6 app")
```
codebien, line 322:
```suggestion
return stacks[0].id, stacks[0].slug, nil
```
Am I missing something or we can simplify in this way?

internal/cmd/common.go
codebien: In big packages like `internal/cmd` I would prefer a specialized file. Can you move to a `stack.go` or `cloud_stack.go` file, please?

internal/cmd/common.go
```diff
@@ -127,3 +129,62 @@ func handleTestAbortSignals(gs *state.GlobalState, gracefulStopHandler, onHardSt
...
+	const suffix = ".grafana.net"
+	if len(s) > len(suffix) && s[len(s)-len(suffix):] == suffix {
+		return s[:len(s)-len(suffix)]
+	}
+	return s
```
codebien, line 151: It seems you want https://pkg.go.dev/strings#TrimSuffix

internal/cmd/cloud_login.go
```diff
@@ -186,3 +263,61 @@ func validateToken(gs *state.GlobalState, jsonRawConf json.RawMessage, token str
...
 }
+
+func getStacks(gs *state.GlobalState, jsonRawConf json.RawMessage, token string) (map[string]int, error) {
+	// We want to use this fully consolidated config for things like
+	// host addresses, so users can overwrite them with env vars.
+	consolidatedCurrentConfig, warn, err := cloudapi.GetConsolidatedConfig(
+		jsonRawConf, gs.Env, "", nil, nil)
+	if err != nil {
+		return nil, err
+	}
```
codebien, line 274: It seems to me this should be a dependency of this function instead of being part of the logic

internal/cmd/cloud_login.go
```diff
@@ -92,18 +101,47 @@ func (c *cmdCloudLogin) run(cmd *cobra.Command, _ []string) error {
 	show := getNullBool(cmd.Flags(), "show")
 	reset := getNullBool(cmd.Flags(), "reset")
 	token := getNullString(cmd.Flags(), "token")
+	stackSlug := getNullString(cmd.Flags(), "stack-slug")
```
codebien, line 104: Is the pull request's description outdated, or they are intentionally different?

internal/cmd/cloud_login.go
codebien: I guess you need someone from the backend to address this question.

internal/cmd/common.go
```diff
@@ -127,3 +129,62 @@ func handleTestAbortSignals(gs *state.GlobalState, gracefulStopHandler, onHardSt
...
+		if stackSlug.Valid && stackSlug.String != "" {
+			id, err := resolveStackSlugToID(gs, jsonRawConf, token, stackSlug.String)
+			if err != nil {
+				return 0, fmt.Errorf("could not resolve stack slug %q to stack ID: %w", stackSlug.String, err)
+			}
+			*stackID = null.IntFrom(id)
+		} else {
+			// TODO: do this in the future? But it is a breaking change...
+			// gs.Logger.Error("please specify a projectID in your test or use `k6 cloud login` to set up a default stack")
+			// return 0, fmt.Errorf("no projectID specified and no default stack set")
```
codebien, line 177: We can do two things here:

1. Add a warning and make the user aware of the future changes without breaking the current logic
2. We can start having feature flags with breaking changes, we might also have some `next` feature flag, where we enable all the feature in the shape of `v2`. So, the users that don't care about breaking changes can benefit of the new features.

codebien (commented):
Hey @dgzlopes, thanks for your contribution here 💌 

### Review of the user experience's definition

I feel this is complex enough that we have to pay attention to clearly document the user's scenarios. It should serve as a future reference, detailing our decisions and their rationale. I feel the current pull request's description would benefit if we could update, improve the format and create a better structured tree of the expected user cases.

I feel the new use cases and the technical decisions/limitations of the implementation are mixed together. Separating these into distinct paragraphs would enhance readability.

### Review of the code

Testing: It looks like we might be missing some tests for this logic. Perhaps you were aiming to get some initial feedback before diving deeper into testing, that's perfectly fine. We can definitely plan to incorporate them in a next step.
However, if the idea was to bypass tests altogether, I'd suggest we reconsider. The logic already has a good amount of history and complexity, and adding more on top of it without the safety net of tests would make future changes very error-prone.

HTTP clients: as soon as you will add tests then you will hit the issue with faking things with the http.DefaultClient. We should inject the clients as we are doing in other parts so we can mock their behavior for easily test the scenarios.

Sparse logic: I didn’t explore this assumption in detail but I have the impression that the source code might benefit if we add some sort of StackResolver component and we delegate to it all the logic for doing the resolution.

cloudapi/api.go
joanlopez: Yeah, I agree that as we usually rely on `Client.baseURL`, and in this case it seems to be a different host, would likely be better to split into a dedicated client, as otherwise we could easily mess things up in the future.

cloudapi/api.go
```diff
@@ -336,3 +338,95 @@ func (c *Client) ValidateToken() (*ValidateTokenResponse, error) {
 
 	return &vtr, nil
 }
+
+// Organization represents a Grafana Cloud k6 organization
+type Organization struct {
+	ID               int    `json:"id"`
+	GrafanaStackName string `json:"grafana_stack_name"`
+	GrafanaStackID   int    `json:"grafana_stack_id"`
```
joanlopez, line 346: I have tested the endpoint, and looks like it could return a _null_ `grafana_stack_name` and/or `grafana_stack_id`.

Is that only a particular case for my account? Or an edge case that we should handle?
Because, in such case, I think we should make this `nil`-able, and/or somehow explicitly document that it can be the case.

I'm wondering the wide variety of corner cases we can find once we release this to users 🤔

cloudapi/api.go
joanlopez: > Should we change the existing Client, or add a new one?

As discussed above, I'd suggest adding a new client, so we keep things isolated, especially considering we're speaking about different hosts or different versions of the API.

> What do you think about using the [auto-generated Go SDK](https://github.com/grafana/k6-cloud-openapi-client-go)?

I think we should use it when possible, but I suspect most of the operations that are supported now (by k6) aren't supported by the latest version of the Public API. But yeah, if we start adding support into k6 for the operations available through the auto-generated client, then I'd suggest using it for those, for sure.

cloudapi/api.go
joanlopez: > What is the problem with this?

I guess that, at the very least, we should parameterize the `baseURL` (ideally into a different client, as mentioned above), and only keep the specific path/endpoint hard-coded.

cloudapi/api.go
joanlopez: I guess Backend folks may want to know where the API is being consumed from, right @dgzlopes? 🤔 

If we ever replace this custom implementation in favor of the auto-generated Go SDK, the User-Agent is yet another configuration attribute that you can customize.

Wrt the value, I think we might want to use something that explicitly let us know this request is coming from the k6 binary.

cloudapi/api.go
```diff
@@ -336,3 +338,95 @@ func (c *Client) ValidateToken() (*ValidateTokenResponse, error) {
...
+	q := req.URL.Query()
+	q.Add("$orderby", "created")
+	req.URL.RawQuery = q.Encode()
+
+	req.Header.Set("X-Stack-Id", fmt.Sprintf("%d", stackID))
+	// TODO: by default the client uses Token instead of Bearer
+	req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", c.token))
+	req.Header.Set("User-Agent", "Go-http-client")
+
+	// TODO: Can't use c.Do bc it messes up the headers
```
joanlopez, line 388: I think this is yet another indicator that we should likely have a different client implementation for this.

internal/cmd/cloud_login.go
joanlopez: I guess the same applies to the flag defined right below this one.

cloudapi/api.go
dgzlopes: > Is it really a strict requirement? It seems we can get the StackID from the token just by calling /account/me, or is only available in /v3?

A user can be in many Stacks! Which one should we pick? :eyes: 

That's the current problem, IMO. 

> I'd suggest adding a new client

I'll try to do this.  I'll also ask the backend about the `/me` endpoint and if they plan to include it in the new API.

That said, I'll do the smallest thing we can in this area to move this PR forward. We can keep adding things to that new client in future PRs (e.g., when I add `k6 plan` I'll need a few more methods)

joanlopez:
@dgzlopes should we turn this into a draft PR? Or planning to make some progress soon? Thanks!

dgzlopes:
Closing in favor of https://github.com/grafana/k6/pull/5420

I implemented most things from scratch based on what we discussed in the design doc.

PR #4944, opened 2025-07-18
Add Prometheus metrics endpoint

gouthamve:
I have a friend who is using k6 for some massive database load-tests that run for hours. However, k6 is OOMing sometimes and they'd like to understand if they need to modify GOGC values.

The easiest way for them to do it is to leverage the Prometheus Golang runtime metrics. This was brought up before in #3535 however, it was rejected for two reasons:

1. `client_golang` dependency might go away in favor of OTel
2. The future of metrics exposition is OTel

However I'd like to reopen this for two reasons. `client_golang` is a dependency of `opentelemetry-go` as well. So that dependency is not going away.

But the more important reason is that there are already many high performance Golang systems exposing `/metrics` and the metrics and tooling around them is pretty mature and extensive.

The OTel Golang runtime metrics are new and not yet stable: https://opentelemetry.io/docs/specs/semconv/runtime/go-metrics/ Reimplementing everything for these metrics is not a trivial effort.

cc @joanlopez @oleiade @mstoykov @dgzlopes 

### DIFF
--- a/internal/api/server.go
+++ b/internal/api/server.go
@@ -10,6 +10,7 @@ import (
 	_ "net/http/pprof" //nolint:gosec // Register pprof handlers
 	"time"
 
+	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/sirupsen/logrus"
 
 	v1 "go.k6.io/k6/api/v1"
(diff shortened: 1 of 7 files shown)


ankur22 (commented):
I'm probably not the best person to review this, but I'd like to get some clarification still. With this change, we would be exposing the standard Go runtime and process metrics that the `client_golang` library auto‐registers, correct? I don't think k6 registers any other data using prometheus.

gouthamve:
> we would be exposing the standard Go runtime and process metrics that the client_golang library auto‐registers, correct?

Yes

szkiba:
FYI https://github.com/szkiba/xk6-prometheus
(an extension does similar or same thing)

Issue #4502, opened 2021-09-30
Implement `route(url, handler[, options])`


robingustafsson:
Add support for `Page.route(url, handler[, options])`, to enable request interception allowing any request initiated in the page to be modified. A request with a matching URL will stall until continued, fulfilled or aborted.

**Relevant links**:
- Playwright docs: https://playwright.dev/docs/api/class-page#page-route
- xk6-browser code starting point:
    1. https://github.com/grafana/k6/blob/0ded7206042c40a5452b525d077da0d82a8ca607/internal/js/modules/k6/browser/browser/page_mapping.go#L314
    2. https://github.com/grafana/k6/blob/7129195a85e57533564e46086018e43915d27945/internal/js/modules/k6/browser/common/frame_manager.go#L524
- Playwright code to add a route:
    1. https://github.com/microsoft/playwright/blob/main/packages/playwright-core/src/client/page.ts#L517
      - Adding a new RouteHandler onto _routes
    2. https://github.com/microsoft/playwright/blob/main/packages/playwright-core/src/client/network.ts#L584
      - Is it a string, regex or pattern match on all?
    3. https://github.com/microsoft/playwright/blob/main/packages/playwright-core/src/server/dispatchers/pageDispatcher.ts#L172
      - Setting up the handler on the page
      - Create a RouteDispatcher, which implementes the methods to continue, abort, fufill etc.
    4. https://github.com/microsoft/playwright/blob/a15e94aa3f411232e4edd05b4b11b8c2f09eadee/packages/playwright-core/src/client/page.ts#L148
      - Possibly handles the route that was dispatched in the previous step.
    5. https://github.com/microsoft/playwright/blob/a15e94aa3f411232e4edd05b4b11b8c2f09eadee/packages/playwright-core/src/client/page.ts#L190
      - ?
- Playwright code to work with the routes:
    1. https://github.com/microsoft/playwright/blob/main/packages/playwright-core/src/server/chromium/crNetworkManager.ts#L278
      - About to send the request to the server from chromium, playwright has intercepted it.
    2. https://github.com/microsoft/playwright/blob/main/packages/playwright-core/src/server/frames.ts#L311
      - If a page ourte handler has been defined, call it.
    3. https://github.com/microsoft/playwright/blob/main/packages/playwright-core/src/server/dispatchers/pageDispatcher.ts#L179
      - The call to the route handler.

## Why

[...]

PR #4953, opened 2025-07-21
Browser: add page.route

AgnesToulet:
## What?
This PR implements the `page.route(url, handler)` function. 
The handler has the following signature `function(route)` and allows to call `route.abort` (`continue` and `fulfill` function will be fully implemented in following PRs).

## Why?
This is a very useful function implemented in Playwright and allowing users to skip requests they don't want to include in the tests (currently by aborting them, later by fulfilling them with a test response). 

## Checklist

- [x] I have updated or added an issue to the [k6-documentation](https://github.com/grafana/k6-docs): grafana/k6-docs#NUMBER if applicable
- [x] I have updated or added an issue to the [TypeScript definitions](https://github.com/grafana/k6-DefinitelyTyped/tree/master/types/k6): grafana/k6-DefinitelyTyped#NUMBER if applicable

## Related PR(s)/Issue(s)
Part of https://github.com/grafana/k6/issues/4502

### DIFF
--- a/internal/js/modules/k6/browser/common/frame_manager.go
+++ b/internal/js/modules/k6/browser/common/frame_manager.go
@@ -546,8 +546,9 @@ func (m *FrameManager) requestStarted(req *Request) {
 	}
 
 	route := NewRoute(m.logger, m.page.mainFrameSession.networkManager, req)
-	routeHandlers := m.page.getRoutes()
-	for _, r := range routeHandlers {
+	m.page.routesMu.RLock()
+	defer m.page.routesMu.RUnlock()
+	for _, r := range m.page.routes {
 		matched, err := r.urlMatcher(req.URL())
 		if err != nil {
 			m.logger.Errorf("FrameManager:requestStarted",
@@ -559,11 +560,18 @@ func (m *FrameManager) requestStarted(req *Request) {
 			continue
 		}
 
-		err = r.handler(route)
-		if err != nil {
-			m.logger.Errorf("FrameManager:requestStarted",
-				"fmid:%d rurl:%s error handling request with route: %v", m.ID(), req.URL(), err)
-		}
+		func() {
+			// In case routes are updated in the handler
+			m.page.routesMu.RUnlock()
+			defer m.page.routesMu.RLock()
+
+			err := r.handler(route)
+			if err != nil {
+				m.logger.Errorf("FrameManager:requestStarted",
+					"fmid:%d rurl:%s error handling request with route: %v", m.ID(), req.URL(), err)
+			}
+		}()
+
 		return
 	}
 	route.Continue()
--- a/internal/js/modules/k6/browser/common/page.go
+++ b/internal/js/modules/k6/browser/common/page.go
@@ -792,13 +792,6 @@ func (p *Page) hasRoutes() bool {
 	return len(p.routes) > 0
 }
 
-func (p *Page) getRoutes() []*RouteHandler {
-	p.routesMu.RLock()
-	defer p.routesMu.RUnlock()
-
-	return p.routes
-}
-
 func (p *Page) resetViewport() error {
 	p.logger.Debugf("Page:resetViewport", "sid:%v", p.sessionID())
 


internal/js/modules/k6/browser/browser/page_mapping.go
```diff
@@ -312,6 +312,38 @@ func mapPage(vu moduleVU, p *common.Page) mapping { //nolint:gocognit,cyclop
...
+		"route": func(path string, handler sobek.Callable) (*sobek.Promise, error) {
+			return k6ext.Promise(vu.Context(), func() (any, error) {
+				ctx := vu.Context()
+
+				// TODO: Is this necessary? 
+				// Run the event handler in the task queue to
+				// ensure that the handler is executed on the event loop.
+				tq := vu.get(ctx, p.TargetID())
+				routeHandler := func(route *common.Route) error {
+					done := make(chan struct{})
```
ankur22: It doesn't look like we're waiting on anything. Do we need `done`?

internal/js/modules/k6/browser/common/http.go
```diff
@@ -706,3 +708,55 @@ func (r *Response) Text() (string, error) {
...
+		r.logger.Errorf("Route:Continue", "rurl:%s err:%s", r.request.URL(), err)
+		return
+	}
+
+	r.networkManager.ContinueRequest(r.request.interceptionID)
+}
+
+func (r *Route) startHandling() error {
+	if r.handled {
+		return fmt.Errorf("route is already handled")
```
ankur22: Nice catch!

internal/js/modules/k6/browser/browser/page_mapping.go
```diff
@@ -796,3 +797,65 @@ func parseGetByRoleOptions(ctx context.Context, opts sobek.Value) *common.GetByR
...
+					if err != nil {
+						rtnErr = fmt.Errorf("executing page.route('%s') handler: %w", path, err)
+						return nil
+					}
+
+					return nil
+				})
+
+				select {
+				case <-done:
```
ankur22: Yeah, this makes sense. We want to wait for the user provided function (`handler`) before proceeding 👍

internal/js/modules/k6/browser/common/http.go
```diff
@@ -706,3 +708,55 @@ func (r *Response) Text() (string, error) {
...
+		logger:         logger,
+		networkManager: networkManager,
+		request:        request,
+		handled:        false,
+	}
+}
+
+func (r *Route) Request() *Request { return r.request }
+
+func (r *Route) Abort(errorCode string) {
```
ankur22, line 734: I think we need to make `abort` (and all but `request`) return a `promise`. At the moment this is synchronous.

internal/js/modules/k6/browser/common/http.go
```diff
@@ -706,3 +708,55 @@ func (r *Response) Text() (string, error) {
...
+	err := r.startHandling()
+	if err != nil {
+		r.logger.Errorf("Route:Abort", "rurl:%s err:%s", r.request.URL(), err)
```
ankur22, line 737: Should we return an error so that Sobek throws the error and stops the iteration, to copy the Playwright behaviour? At the moment we log an error and continue on with the iteration.

internal/js/modules/k6/browser/common/network_manager.go
```diff
@@ -749,6 +815,77 @@ func (m *NetworkManager) Authenticate(credentials Credentials) error {
 	return nil
 }
 
+func (m *NetworkManager) AbortRequest(requestID fetch.RequestID, errorReason string) {
+	netErrorReason, ok := m.errorReasons[errorReason]
+	if !ok {
+		m.logger.Errorf("NetworkManager:AbortRequest", "unknown error code: %s", errorReason)
```
ankur22, line 860: Should we propagate the error back to the caller so that it can be thrown to stop the iteration, which will match the Playwright behaviour?

internal/js/modules/k6/browser/common/network_manager.go
```diff
@@ -749,6 +815,77 @@ func (m *NetworkManager) Authenticate(credentials Credentials) error {
...
+		return
+	}
+	m.logger.Debugf("NetworkManager:AbortRequest", "aborting request (id: %s, errorReason: %s)",
```
ankur22: It might be useful to have the debug log at the beginning of the method.

internal/js/modules/k6/browser/common/network_manager.go
```diff
@@ -749,6 +815,77 @@ func (m *NetworkManager) Authenticate(credentials Credentials) error {
...
+		// action on. It's also usually ok to ignore since it means that the page has navigated
+		// away or something has occurred which means that the request is no longer needed and
+		// isn't being tracked by chromium.
+		if strings.Contains(err.Error(), "Invalid InterceptionId") {
+			m.logger.Debugf("NetworkManager:ContinueRequest", "invalid interception ID (%s) continuing request: %s",
+				requestID, err)
+			return
+		}
+
+		m.logger.Errorf("NetworkManager:ContinueRequest", "fail to continue request (id: %s): %s",
```
ankur22, line 901: I agree with what you have done here. We don't rely on the user to input any arguments when calling `continue`, so we shouldn't return the error back to the caller here, and instead just log it and continue. I wonder if it's worth downgrading to a warning though, since an error might confuse the user and I don't think there's anything they can do.

internal/js/modules/k6/browser/common/page.go
```diff
@@ -1234,6 +1254,19 @@ func (p *Page) Referrer() string {
...
 
+// Route register a handler to be executed for a given request path
+func (p *Page) Route(
+	path string,
+	handlerCallback RouteHandlerCallback,
+	jsRegexChecker JSRegexChecker,
+) error {
+	p.logger.Debugf("Page:Route", "sid:%v path:%s", p.sessionID(), path)
+	routeHandler := NewRouteHandler(path, handlerCallback, jsRegexChecker)
+	p.routes = append([]*RouteHandler{routeHandler}, p.routes...)
```
ankur22, line 1320: Does this need to be protected with a mutex?

internal/js/modules/k6/browser/tests/frame_manager_test.go
```diff
@@ -118,3 +119,122 @@ func TestWaitForFrameNavigation(t *testing.T) {
 	require.NoError(t, err)
 	assert.Equal(t, "Second page", title)
 }
+
+func TestFrameManagerRequestStartedWithRoutes(t *testing.T) {
```
ankur22, line 123: I think we need tests for the other `route` methods e.g. `abort`.

internal/js/modules/k6/browser/common/page.go
```diff
@@ -1234,6 +1254,19 @@ func (p *Page) Referrer() string {
...
+
+	return p.mainFrameSession.updateRequestInterception(true)
```
ankur22: Does this need to be called on every call to `page.route`? Could it be called once on the first call to `page.route`?

internal/js/modules/k6/browser/common/frame_manager.go
```diff
@@ -540,7 +540,40 @@ func (m *FrameManager) requestStarted(req *Request) {
...
 	}
-	m.logger.Debugf("FrameManager:requestStarted", "fmid:%d rurl:%s pdoc:nil", m.ID(), req.URL())
+
+	if !m.page.hasRoutes() {
+		return
+	}
+
+	route := NewRoute(m.logger, m.page.mainFrameSession.networkManager, req)
+	for _, r := range m.page.routes {
+		matcher, err := urlMatcher(r.path, r.jsRegexChecker)
```
ankur22: Could we create this url matcher once in the `route` method in `page` where we create a `NewRoute` instance? At the moment a new instance is created every time a request is created.

internal/js/modules/k6/browser/common/frame_manager.go
```diff
@@ -540,7 +540,40 @@ func (m *FrameManager) requestStarted(req *Request) {
...
+			m.logger.Errorf("FrameManager:requestStarted",
+				"fmid:%d rurl:%s error matching url: %v", m.ID(), req.URL(), err)
+			continue
+		}
+
+		if !matched {
+			continue
+		}
+
+		err = r.handler(route)
```
ankur22: Could we create a new instance of `route` just before passing it to `handler`?

internal/js/modules/k6/browser/common/frame_manager.go
```diff
@@ -540,7 +540,40 @@ func (m *FrameManager) requestStarted(req *Request) {
...
+		if err != nil {
+			m.logger.Errorf("FrameManager:requestStarted",
+				"fmid:%d rurl:%s error handling request with route: %v", m.ID(), req.URL(), err)
+			continue
+		}
+		return
+	}
+	route.Continue()
```
ankur22, line 577: Could we create a new instance of `route` just before calling `continue`?

examples/browser/route.js
```diff
@@ -0,0 +1,44 @@
+import { sleep } from "k6";
```
ankur22: `sleep` isn't being used.

internal/js/modules/k6/browser/common/page.go
ankur22: I've not been able to cause any data races yet, so maybe it's not an issue 🤔

internal/js/modules/k6/browser/common/network_manager.go
```diff
@@ -749,6 +815,77 @@ func (m *NetworkManager) Authenticate(credentials Credentials) error {
...
+		WithBody(params.Body)
+
+	if err := action.Do(cdp.WithExecutor(m.ctx, m.session)); err != nil {
+		// Avoid logging as error when context is canceled.
+		// Most probably this happens when trying to fail a site's background request
+		// while the iteration is ending and therefore the browser context is being closed.
+		if errors.Is(err, context.Canceled) {
+			m.logger.Debug("NetworkManager:FulfillRequest", "context canceled fulfilling request")
+		} else {
+			m.logger.Errorf("NetworkManager:FulfillRequest", "fail to fulfill request (id: %s): %s",
```
ankur22, line 919: Not sure if we should return an error here or not. Since we're taking user args, could it cause an error that should be shown to the user and for the iteration to end so that they can fix any issues with the args?

internal/js/modules/k6/browser/browser/page_mapping.go
AgnesToulet: See https://github.com/grafana/k6/pull/4953#discussion_r2221890796

internal/js/modules/k6/browser/common/page.go
AgnesToulet: Users should write `await page.route(...)` so the function should finish before another call to the `route` function is made. 
I can imagine this failing if we don't `await` but do we want to support this use case? Forgetting the `await` will cause other weird behaviors.

internal/js/modules/k6/browser/common/http.go
AgnesToulet: Yes, I've started this work and returning errors in the following PR. I can commit part of the work in this PR if it's easier to review as you made a few comments about returning errors to the user.

internal/js/modules/k6/browser/common/network_manager.go
AgnesToulet: This line is not a change for now, I took it from [there](https://github.com/grafana/k6/pull/4953/files/a7266df4d347f62a0a5a5e0f5bf24b925493ddb6#diff-7f7619be15080b7697b5d630cd5c2a14d5953c8496eabae91b1908dfcb371f53L547-L564) and moved it to its own function to be able to use it in other places. 
I think an error makes sense in the longer term as users will be able to do `route.continue(...)` and could cause an error with the options they give to this function.

internal/js/modules/k6/browser/common/http.go
ankur22: I think you can link to the other PR in this and other comments so that anyone else reading this can follow along too. Once both PRs have been approved we can merge them both at the same time.

internal/js/modules/k6/browser/tests/frame_manager_test.go
AgnesToulet: Added in https://github.com/grafana/k6/pull/4953/commits/150c16ee214d4bd4d9a0a0aaf34fa214d89b8e91

internal/js/modules/k6/browser/common/network_manager.go
AgnesToulet: Fixed in https://github.com/grafana/k6/pull/4953/commits/caa3e356e42ef6533f8c0c4425ac3263ceb4d107

internal/js/modules/k6/browser/common/page.go
AgnesToulet: Fixed in https://github.com/grafana/k6/pull/4953/commits/a3f7932ff9ee8766375a98f7bde0b3845112f5f9

internal/js/modules/k6/browser/common/frame_manager.go
AgnesToulet: Great idea! Updated in https://github.com/grafana/k6/pull/4953/commits/48787ab2a9c41c510eed541f1fafba49167c4ecb

internal/js/modules/k6/browser/common/frame_manager.go
AgnesToulet: I'm not sure it makes sense. I would understand if it avoids creating it and not using it but here the route object will be used every time, either here or in the `route.Continue` line at the end. 
If you think the route creation and its usage is too far away, the update in https://github.com/grafana/k6/pull/4953/commits/48787ab2a9c41c510eed541f1fafba49167c4ecb improves this.

examples/browser/route.js
AgnesToulet: Removed in https://github.com/grafana/k6/pull/4953/commits/0de5ead0717fbdfa7d66b65d159cf9d960897f9e

internal/js/modules/k6/browser/common/http.go
AgnesToulet: Here's the PR to add the fulfill function: https://github.com/grafana/k6/pull/4961. The first two commits are refactoring without introducing new changes and could be added to this PR instead.

examples/browser/route.js
```diff
@@ -0,0 +1,43 @@
...
+		ui: {
+			executor: "shared-iterations",
+			options: {
+				browser: {
+					type: "chromium",
+				},
+			},
+		},
+	},
+};
```
codebien: Is the indentation correct here? 🤔 It sounds very wide.

internal/js/modules/k6/browser/browser/page_mapping.go
```diff
@@ -796,3 +797,65 @@ func parseGetByRoleOptions(ctx context.Context, opts sobek.Value) *common.GetByR
 
 	return o
 }
+
+func parseStringOrRegex(v sobek.Value) string {
```
codebien: As someone not very expert in this area, I would benefit from having a comment on this function to reduce the cognitive load.

internal/js/modules/k6/browser/common/http.go
```diff
@@ -706,3 +708,55 @@ func (r *Response) Text() (string, error) {
...
+func NewRoute(logger *log.Logger, networkManager *NetworkManager, request *Request) *Route {
+	return &Route{
+		logger:         logger,
+		networkManager: networkManager,
+		request:        request,
+		handled:        false,
+	}
+}
+
+func (r *Route) Request() *Request { return r.request }
```
codebien, line 731:
```suggestion
```
As we return the pointer, isn't better to simply export this field?

internal/js/modules/k6/browser/common/http.go
```diff
@@ -706,3 +708,55 @@ func (r *Response) Text() (string, error) {
...
+// Route allows to handle a request
+type Route struct {
+	logger         *log.Logger
+	networkManager *NetworkManager
+
+	request *Request
+	handled bool
+}
+
+func NewRoute(logger *log.Logger, networkManager *NetworkManager, request *Request) *Route {
```
codebien, line 722: I'm surprised, this and the following APIs aren't caught from the linter. The comment should be required here.

internal/js/modules/k6/browser/common/http.go
codebien:
```suggestion
		return fmt.Errorf("route %s is already handled")
```
Should we let them know about which route? Could it be useful?

internal/js/modules/k6/browser/common/network_manager.go
```diff
@@ -383,6 +409,7 @@ func (m *NetworkManager) handleEvents(in <-chan Event) bool {
 
 func (m *NetworkManager) onLoadingFailed(event *network.EventLoadingFailed) {
 	req, ok := m.requestFromID(event.RequestID)
+
```
codebien:
```suggestion
```
If it is expected, discard my suggestion

internal/js/modules/k6/browser/common/http.go
AgnesToulet: See https://github.com/grafana/k6/pull/4961

internal/js/modules/k6/browser/common/network_manager.go
AgnesToulet: See https://github.com/grafana/k6/pull/4961

internal/js/modules/k6/browser/common/network_manager.go
AgnesToulet: Done in the next PR: https://github.com/grafana/k6/pull/4961/files#diff-7f7619be15080b7697b5d630cd5c2a14d5953c8496eabae91b1908dfcb371f53R911

internal/js/modules/k6/browser/common/http.go
AgnesToulet: I'm not sure, I mostly followed what was done with the `Request` and `Response` objects (in the same file) because I thought it was better for the mapping files and sobek conversions. 

@ankur22 any opinion on this?

examples/browser/route.js
AgnesToulet: Good catch, fixed in https://github.com/grafana/k6/pull/4953/commits/af32d19900ed7930a79b0e787f9ca570551a52b3

internal/js/modules/k6/browser/common/network_manager.go
AgnesToulet: Not really expected, I added some changes here that I finally removed and left a new line behind. Fixed in https://github.com/grafana/k6/pull/4953/commits/0d38394af3881bc4eddc6702b0d7655451a979ab

internal/js/modules/k6/browser/common/http.go
AgnesToulet: Good idea! Updated in https://github.com/grafana/k6/pull/4953/commits/66a23ea8ad690ec08d1dc9ffc305ee4c502a0e60

internal/js/modules/k6/browser/common/http.go
AgnesToulet: Added for this function and the next ones in https://github.com/grafana/k6/pull/4953/commits/46131c5a97a5429116f7c627221900be669b0e1c

internal/js/modules/k6/browser/browser/page_mapping.go
AgnesToulet: Added in https://github.com/grafana/k6/pull/4953/commits/46131c5a97a5429116f7c627221900be669b0e1c

internal/js/modules/k6/browser/common/http.go
ankur22: I think you mean why is it a function, why not just export it as a field i.e.:

```js
const r = route.request;
```

We're copying [route.request()](https://playwright.dev/docs/api/class-route#route-request) exactly so that we match Playwright's implementation.

codebien: Gotcha!

codebien:
@AgnesToulet it seems a rebase/merge of master is required as we have conflicts

AgnesToulet:
Yes, but I also have failing tests since my latest merge that I'm looking at first.

internal/js/modules/k6/browser/common/network_manager.go
```diff
@@ -517,15 +566,40 @@ func (m *NetworkManager) onRequest(event *network.EventRequestWillBeSent, interc
...
 
 	defer func() {
+		requestID := event.NetworkID
+		if requestWillBeSentEvent, ok := m.reqIDToRequestWillBeSentEvent[requestID]; ok {
+			m.onRequest(requestWillBeSentEvent, event)
+			delete(m.reqIDToRequestWillBeSentEvent, requestID)
+		} else {
+			m.reqIDToRequestPausedEvent[requestID] = event
+		}
+
```
AgnesToulet: Just moved this from after to before sending the fail event. This will lead to the exact same behavior as before when both `onRequest` and `onRequestPaused` are called for all requests, whether they are aborted or not.
I think we could avoid calling to `onRequest` though if we know we'll abort the request but:
1. It's breaking the [the recent change from Joan](https://github.com/grafana/k6/pull/4918) to surface navigation errors because we receive less events, but we do receive navigation to `chrome-error://` that we could use to indicate the error (see https://github.com/grafana/k6/pull/4918/files#r2208252864).
2. This could cause a lot of other things that are hard to determine so I think it's better to do it later, when we have a stable implementation of `page.route`.

codebien: >i's breaking the https://github.com/grafana/k6/pull/4918 to surface navigation errors

@AgnesToulet it sounds an issue :warning: 

If I recall correctly from our sync, we decided to move this pull request to v1.3.0. Unless we roll back Joan's changes, we'll need to mitigate the breaking change introduced here. We can't introduce breaking changes to user APIs in minor versions, as that doesn't align with our [stability guarantees](https://grafana.com/docs/k6/latest/reference/versioning-and-stability-guarantees/).

AgnesToulet: Sorry for not being clear: with this change, we are not breaking anything. It's the exact same behavior as before.

I think it could be improved though because I don't see the point of calling `onRequest` for a request we know is aborted but I don't think this change should happen now for the two reasons I mentioned above. And if we decide to update this in a future release, we would need to improve how we handle navigation errors to not introduce a breaking change there.

examples/browser/route.js
```diff
@@ -0,0 +1,43 @@
...
+    console.log("on request", "url", request.url());
+  });
+
+  page.on("response", function (response) {
+    console.log("on response", "url", response.url());
+  });
+
+  await page.goto("https://quickpizza.grafana.com/", {
+    waitUntil: "networkidle",
+  });
```
ankur22: Just for consistency reasons, could you place this part of the code in a `try` and the `page.close` in a `finally` block?

Take a look at our other browser examples e.g. https://github.com/grafana/k6/blob/master/examples/browser/getattribute.js

internal/js/modules/k6/browser/common/frame_manager.go
```diff
@@ -540,7 +540,33 @@ func (m *FrameManager) requestStarted(req *Request) {
...
+
+		if !matched {
+			continue
+		}
+
+		err = r.handler(route)
+		if err != nil {
+			m.logger.Errorf("FrameManager:requestStarted",
+				"fmid:%d rurl:%s error handling request with route: %v", m.ID(), req.URL(), err)
+			continue
```
ankur22: Should we return here since the handler has been handled, but with an error.

internal/js/modules/k6/browser/common/page.go
```diff
@@ -1256,6 +1276,32 @@ func (p *Page) Referrer() string {
...
+			return err
+		}
+	}
+
+	matcher, err := urlMatcher(path, jsRegexChecker)
+	if err != nil {
+		return fmt.Errorf("creating url matcher for path %s: %w", path, err)
+	}
+
+	routeHandler := NewRouteHandler(path, handlerCallback, matcher)
```
ankur22, line 1317: Do we need to ensure that more than one route handler isn't registered for the same url pattern?

internal/js/modules/k6/browser/common/network_manager.go
```diff
@@ -517,15 +566,40 @@ func (m *NetworkManager) onRequest(event *network.EventRequestWillBeSent, interc
...
+		event.Request.URL, event.Request.Method, event.Initiator.Type, event.FrameID)
+
+	if m.protocolReqInterceptionEnabled {
+		requestID := event.RequestID
+		if requestPausedEvent, ok := m.reqIDToRequestPausedEvent[requestID]; ok {
+			m.onRequest(event, requestPausedEvent)
+			delete(m.reqIDToRequestPausedEvent, requestID)
+		} else {
+			m.reqIDToRequestWillBeSentEvent[requestID] = event
+		}
```
ankur22: Can you clarify (maybe with a comment somewhere) why we need this here in `onRequestWillBeSent`, and we need something similar in `onRequestPaused`?

I can't remember the ordering of the incoming CDP requests. Is the order?:

1. `requestStarted` 
2. `onRequestPaused`
3. `onRequestWillBeSent`

But the ordering can differ, and regardless of the ordering, `onRequest` must be called at the end of either call?

internal/js/modules/k6/browser/common/network_manager.go
```diff
@@ -487,11 +517,30 @@ func (m *NetworkManager) onRequest(event *network.EventRequestWillBeSent, interc
...
+	}
+
 	if !ok {
 		m.logger.Debugf("NetworkManager:onRequest", "url:%s method:%s type:%s fid:%s frame is nil",
 			event.Request.URL, event.Request.Method, event.Initiator.Type, event.FrameID)
 	}
 
+	var interceptionID fetch.RequestID
+	if requestPausedEvent != nil {
+		interceptionID = requestPausedEvent.RequestID
```
ankur22, line 545: Why does it matter that we use the `interceptionID` from `requestPausedEvent` in `request`?

internal/js/modules/k6/browser/common/page.go
ankur22: Tested with PW, only one handler is ran. We will stick with allowing this to match PW's behaviour 👍

internal/js/modules/k6/browser/common/network_manager.go
AgnesToulet: I don't think there is a `requestStarted`. There are `onRequestPaused` and `onRequestWillBeSent` that are sent at the beginning of a request but the order is not guaranteed and yes we should call `onRequest` either way but only once and with the event object from `onRequestPaused` if we want to get the correct interception ID. I'll try to explain that in a comment :+1:

internal/js/modules/k6/browser/common/network_manager.go
AgnesToulet: This is the only way to get the interception ID. Before, this was a parameter of the function but it was actually always empty (see https://github.com/grafana/k6/pull/4953/files#diff-7f7619be15080b7697b5d630cd5c2a14d5953c8496eabae91b1908dfcb371f53L370)

All my updates to the request workflow was done to get this interception ID here and be able to use it in the route handler.

internal/js/modules/k6/browser/common/network_manager.go
ankur22: Right, i see now, thanks!

internal/js/modules/k6/browser/common/frame_manager.go
AgnesToulet: I'm not sure when this can happen to be honest so no strong opinion on this :sweat_smile: Do you have a use case in mind? (Updated in https://github.com/grafana/k6/pull/4953/commits/1156299d912716129176cd726f446d2f945444dc)

examples/browser/route.js
AgnesToulet: Sure! I also updated the double quotes to single quotes. Do we need a prettier config file in the repo to keep the files consistent?

Fixed in https://github.com/grafana/k6/pull/4953/commits/2f8a0b0cf75572ab1112abc298e20784fcd8ee79

internal/js/modules/k6/browser/common/network_manager.go
AgnesToulet: Comments added in https://github.com/grafana/k6/pull/4953/commits/e06fbd9975f3c6d697af3c6b0d2ef0091b4cf19e

examples/browser/route.js
ankur22: > Do we need a prettier config file in the repo to keep the files consistent?

I think this could be something to bring to the team to see what they think. I am slightly against it since it feels like another hurdle during the PR process. As long as the basic structure is similar to the other examples, I think that's ok.

internal/js/modules/k6/browser/common/frame_manager.go
ankur22: Nit: sorry, i should've been clearer (although thanks for the change). You probably don't need a `return` just let it fall out of the if scope and the next line is already a `return`.

ankur22: > I'm not sure when this can happen to be honest so no strong opinion on this 😅

TBH, i'm not sure how to recreate that. I guess it's when Sobek errors. JS issues are usually picked up well before the script is executed, so we can't really check for those.

ankur22 (approved):

internal/js/modules/k6/browser/common/frame_manager.go
AgnesToulet: > Nit: sorry, i should've been clearer (although thanks for the change). You probably don't need a return just let it fall out of the if scope and the next line is already a return.

My bad, end of the week mistake :sweat_smile: Removed in https://github.com/grafana/k6/pull/4953/commits/4d74073ec03ddb97523aa89fbccfb79ee15a9ac3

ankur22 (approved):
🚀

examples/browser/route.js
AgnesToulet: It would not necessarily be a new check but just some rules to have the right auto-format. It's just that I had prettier auto-format enabled for the JS files and with these files, it was automatically replacing single quotes with double quotes and converting my indentation to tabs. I had to switch to the default "Typescript and Javascript" formatter in VSCode. I guess if we don't have other projects that rely on Prettier, it's fine, else it's a bit cumbersome to have to switch the formatter for different projects.

internal/js/modules/k6/browser/browser/page_mapping.go
```diff
@@ -871,3 +875,52 @@ func parseGetByBaseOptions(
...
+			// Inject JS regex checker for URL regex pattern matching
+			jsRegexChecker, err := injectRegexMatcherScript(ctx, vu, targetID)
+			if err != nil {
+				return nil, err
+			}
+			pathStr := parseStringOrRegex(path, false)
+
+			// Run the event handler in the task queue to
+			// ensure that the handler is executed on the event loop.
+			tq := vu.get(ctx, targetID)
```
mstoykov: I am not certain on the exact separation that needs to happen here, but:
1. currently everything in this function is ran off the event loop
2. the idea of the event loop (not only in k6, but in general) is to synchronyze so nothing is executing or interacting the JS VM (Which is single threaded by definition). In browsers this also is due to (g)UI usually being single threaded for the most part (especially 20+ years ago). 
3. for k6, the VU is also in general not thread safe
4. `injectRegexMatcherScript definitely touches the js runtime quite a bit
5. registering a taskqueue requires that you are on the eventloop as well.

I am pretty sure you should just move all of the lines outside of this return and then use them.

Again this is due to the JS VM and VU not being thread safe, so you need to make certain that you only touch them when running on the event loop. The callback of k6ext.Promise is called in a goroutine *off* the event loop, while the call back to the taskqueueu is called on the vent loop as well as the original call here.

internal/js/modules/k6/browser/browser/page_mapping.go
```diff
@@ -871,3 +875,52 @@ func parseGetByBaseOptions(
...
+				select {
+				case <-done:
+				case <-ctx.Done():
+					rtnErr = errors.New("iteration ended before route completed")
+				}
+
+				return rtnErr
+			}
+
+			return nil, p.Route(pathStr, routeHandler, jsRegexChecker)
```
mstoykov, line 923: This also is called off the event loop, so some of the mutex discussions that are now closed, in the `Route` implementation, are actually valid - this can race. Do `let manyroutres = [route, route, route]; await Promise.all(manyroutes)` as in register a bunch of routes and then await them (the second part isn't necessary technically)

internal/js/modules/k6/browser/browser/page_mapping.go
ankur22: Ok, i was under the impression that working with the js `runtime` in `promises` was safe to do since it was orchestrated in such a way to not cause a race condition even though it wasn't running on the event loop.

If that's not the case then we need refactor a lot of the code elsewhere in the browser module which follows a similar pattern.

If we want to interact with the JS `runtime` or the `VU`, it sounds like we need to do it:

1. on the "main thread" i.e. in this case outside the call to promise;
2. in a task queue which is registered and queue on the "main thread", which means we would need to move `tq.Queue` outside the promise too?

internal/js/modules/k6/browser/browser/page_mapping.go
AgnesToulet: I understand, this seems a weird use case but I'll make the fix :+1:

AgnesToulet: Fixed in https://github.com/grafana/k6/pull/4953/commits/f60ab7b39ca4a16a1bbf48b7bb8d4f41de4398b6

After thinking twice about race, I also added mutex in the network_manager.go layer in https://github.com/grafana/k6/pull/4953/commits/b1f6fa58287f5b8bcc8206955b04d00f76c7e734

internal/js/modules/k6/browser/browser/page_mapping.go
AgnesToulet: Makes sense, thanks for the detailed explanation! Fixed in https://github.com/grafana/k6/pull/4953/commits/9db0adad3ad3118db9f355d6ccb45f1625b22b61

internal/js/modules/k6/browser/common/frame_manager.go
```diff
@@ -540,7 +540,32 @@ func (m *FrameManager) requestStarted(req *Request) {
...
 		frame.pendingDocumentMu.Unlock()
 	}
-	m.logger.Debugf("FrameManager:requestStarted", "fmid:%d rurl:%s pdoc:nil", m.ID(), req.URL())
+
+	if !m.page.hasRoutes() {
+		return
+	}
+
+	route := NewRoute(m.logger, m.page.mainFrameSession.networkManager, req)
+	for _, r := range m.page.routes {
```
ankur22, line 551: Do you need to protect this read with the `routesMu`?

AgnesToulet: Good catch. I think route should not be added when we start navigating as it will lead to unpredictable behavior but it might be worth to protect it to avoid errors (especially when we'll work on `unroute` and start deleting route).

AgnesToulet: Fixed in https://github.com/grafana/k6/pull/4953/commits/ae4f370acb8fa6e131864aedae0ba9dae50f6491

internal/js/modules/k6/browser/common/frame_manager.go
```diff
@@ -540,7 +540,33 @@ func (m *FrameManager) requestStarted(req *Request) {
...
 		frame.pendingDocumentMu.Unlock()
 	}
-	m.logger.Debugf("FrameManager:requestStarted", "fmid:%d rurl:%s pdoc:nil", m.ID(), req.URL())
+
+	if !m.page.hasRoutes() {
+		return
+	}
+
+	route := NewRoute(m.logger, m.page.mainFrameSession.networkManager, req)
+	routeHandlers := m.page.getRoutes()
```
ankur22: I don't think this will work. We're returning the same reference to `p.routes` which could be amended by another thread.

Might need to bring the mutex into the caller i.e. in `FrameManager.requestStarted`. My concern there though is if the lock `routesMu` is being held in `requestStarted`, would that cause a deadlock? Maybe we need to something similar to in [urlTagName](https://github.com/grafana/k6/blob/80553b08ca52814a52ec4c970f464db622895d3d/internal/js/modules/k6/browser/common/page.go#L464-L482), where we unlock the lock before calling the handler.

internal/js/modules/k6/browser/browser/page_mapping.go
mstoykov: @ankur22 This was basically very similar to what triggered https://github.com/grafana/k6/pull/4521 and the following 5+ PRs fixing a bunch of those places. 

k6ext.Promise just runs callback in a goroutine and then *safely* returns the result or error to through the promise. But it doesn't make function callbakc on the event loop. It actually is meant to do exactly the opposite - let you run off the event loop, do some networking and return the results or errors to the event loop and JS.

ankur22: Yep, i understand now, seems obvious 🤦

codebien: @ankur22 @AgnesToulet I have the feeling there is some lack of documentation / tooling. Do you have any feedback regarding possible improvements for this API?

internal/js/modules/k6/browser/common/frame_manager.go
AgnesToulet: Ah right, fixed in https://github.com/grafana/k6/pull/4953/commits/86ed45f49f125ec61c153d0792620e74785a5edd

internal/js/modules/k6/browser/browser/page_mapping.go
AgnesToulet: To be honest, the fact that the function called by `k6ext.Promise` is ran in a go routine is written in the function documentation. 

There is nothing about the VU not being thread safe though, maybe this should be added to the main functions' documentation (`Context()` and `taskQueueRegistry.get(...)`)?

ankur22: > @ankur22 @AgnesToulet I have the feeling there is some lack of documentation / tooling. Do you have any feedback regarding possible improvements for this API?

I don't understand where the lack of documentation / tooling is. I think the issue was a misunderstanding of how VUs, the JS runtime and promises work, not the `page.route` api itself. Can you elaborate @codebien?

ankur22 (approved):
LGTM 🚀 I believe further changes are awaiting in two more PRs 👍

internal/js/modules/k6/browser/common/frame_manager.go
mstoykov: Given that the routes are update in a goroutine off the event loop and this events are also  handled off the event loop it needs some synchronization.

internal/js/modules/k6/browser/common/frame_manager.go
```diff
@@ -540,7 +540,41 @@ func (m *FrameManager) requestStarted(req *Request) {
...
+			m.page.routesMu.RUnlock()
+			defer m.page.routesMu.RLock()
+
+			err := r.handler(route)
+			if err != nil {
+				m.logger.Errorf("FrameManager:requestStarted",
+					"fmid:%d rurl:%s error handling request with route: %v", m.ID(), req.URL(), err)
+			}
+		}()
+
```
mstoykov, line 574: Should this be in a separate goroutine?

Looking at the code this goroutine will block until the defer before the loop is executed. 

I am also not certain what other changes can happen apart from the route being removed from the list, which is fine.

internal/js/modules/k6/browser/browser/page_mapping.go
```diff
@@ -871,3 +875,52 @@ func parseGetByBaseOptions(
...
+			return nil, err
+		}
+		pathStr := parseStringOrRegex(path, false)
+
+		// Run the event handler in the task queue to
+		// ensure that the handler is executed on the event loop.
+		tq := vu.get(ctx, targetID)
+		routeHandler := func(route *common.Route) error {
+			done := make(chan bool)
+			var rtnErr error
```
mstoykov, line 897:
```suggestion
			done := make(chan struct{})
			var rtnErr error
```

`chan struct{}` is usually way more optimized for signaling such as this.

But to be honest here I probably will make it an `make(chan error, 1)` and use it to return the error instead of having separate variable that is updated, and have a single slot for the one error we might need.

This also will get around potential races where the context is done, but we just got error back

mstoykov (approved):
LGTM!

I have left some more comments around some small problems that are not that important.

internal/js/modules/k6/browser/common/frame_manager.go
AgnesToulet: Would it be enough to write some best practices in the function documentation? Users are supposed to await for the `route` function to finish before navigating. Else, it would lead to unpredictable behavior as routes can be registered after the request event and not be taken into account.

internal/js/modules/k6/browser/common/frame_manager.go
AgnesToulet: > Looking at the code this goroutine will block until the defer before the loop is executed.

I don't understand, this function unlock the mutex so that `routes` can be updated in the handler function (unlikely but better to handle it properly rather than encountering weird errors), why would it block?

Currently the only thing that can happen is a new route being added in the handler.

internal/js/modules/k6/browser/common/page.go
```diff
@@ -1270,6 +1294,34 @@ func (p *Page) Referrer() string {
...
+	p.routesMu.Lock()
+	defer p.routesMu.Unlock()
+	p.routes = append([]*RouteHandler{routeHandler}, p.routes...)
```
mstoykov, line 1320: Why are we not appending to the end?

internal/js/modules/k6/browser/common/frame_manager.go
mstoykov: I meant that it will block until it returns. But I guess given that it also returns immediately instead of going through the rest of the routes to search for more, it will be pretty okay. 

> Currently the only thing that can happen is a new route being added in the handler.

I think that the only thing that happens is that we add to `routes` , check if it is empty and iterate over it (here).

We do not even remove routes from it. Going over the other PRs this seems to stay this way in the other 2.

internal/js/modules/k6/browser/common/frame_manager.go
mstoykov: I do not think that we should allow there to be a panic due to a script - irregardless of it will be undefined behaviour. Undefined behaviour is fine, panicking isn't.

The current code does have mutexes so it can just stay as it is.

internal/js/modules/k6/browser/browser/page_mapping.go
codebien: @ankur22 I meant the API on the Event loop / concurrency side (taskQueue, k6ext.Promise, etc..).

internal/js/modules/k6/browser/browser/page_mapping.go
AgnesToulet: As it's a minor update, I'll do it in a following PR :+1:

internal/js/modules/k6/browser/common/page.go
AgnesToulet: To follow Playwright's behavior:
> When several routes match the given pattern, they run in the order opposite to their registration.

I'll add a comment about it in a following PR.

mstoykov: But we only run *one* route?

AgnesToulet: Yes, we run the first route that matches the given pattern (unless the [`fallback` function](https://playwright.dev/docs/api/class-route#route-fallback) is used but this hasn't been implemented yet.

internal/js/modules/k6/browser/browser/page_mapping.go
AgnesToulet: Updated in https://github.com/grafana/k6/pull/4985

internal/js/modules/k6/browser/common/page.go
AgnesToulet: Comment added in https://github.com/grafana/k6/pull/4985

PR #4961, opened 2025-07-22
Browser: add route.fulfill

AgnesToulet:
## What?
This PR adds the `route.fulfill` function. Used in `page.route`, it allows to fulfill a request with a given response.

## Why?
This helps with Playwright compatibility and allow to mock requests while testing.

## Checklist

## Related PR(s)/Issue(s)
Part of https://github.com/grafana/k6/issues/4502

## Known limitation
As you can see in the test script below, if you want to fulfill a request to an external website you need to handle CORS yourself. Playwright had the same issue described [here](https://github.com/microsoft/playwright/issues/12929) with multiple proposed solution. They fixed it by adding CORS for external request if no overrides are provided: see [this function](https://github.com/microsoft/playwright/blob/main/packages/playwright-core/src/server/network.ts#L328). I tried to do the same but struggled with the request origin not being set on our side. I don't think it's blocking this PR though but I'd like to know your thoughts about this issue, if we should try to follow Playwright or choose another option (document it or allow it if an option is set).

## Test script
To test the full workflow, I'm using the following k6 script. It's not part of the PR because it depends on a change on quickpizza: https://github.com/grafana/quickpizza/pull/225 but you can play with it if you run QuickPizza locally.
```
import { check, sleep } from "k6";
import { browser } from "k6/browser";

export const options = {
	scenarios: {
		ui: {
			executor: "shared-iterations",
			options: {
				browser: {
					type: "chromium",
				},
			},
		},
	},
};

	await page.route(
		"https://jsonplaceholder.typicode.com/todos/1",
		function (route) {
			console.log("Intercepting request to external API");
			route.fulfill({
				status: 200,
				body: JSON.stringify({
					id: 1,
					title: "Test Todo",
					completed: false,
				}),
				contentType: "application/json",
				headers: {
					"Access-Control-Allow-Origin": "*",
					"Access-Control-Allow-Credentials": "true",
				},
			});
		}
	);

	await page.goto("http://localhost:3333/browser.php");

	await page.getByRole("button", { name: "external data" }).click();

	sleep(1);

	const checkData = await page.locator("#external-data-display").innerText();
	check(page, {
		externalData: checkData === "External data: Test Todo",
	});

	sleep(3);

	await page.close();
}
```


internal/js/modules/k6/browser/browser/route_mapping.go
```diff
@@ -0,0 +1,64 @@
...
+	return mapping{
+		"abort": func(reason string) *sobek.Promise {
+			return k6ext.Promise(vu.Context(), func() (any, error) {
+				return nil, route.Abort(reason)
+			})
+		},
+		"fulfill": func(opts sobek.Value) *sobek.Promise {
+			return k6ext.Promise(vu.Context(), func() (any, error) {
+				fopts := parseFulfillOptions(vu.Context(), opts)
+				fmt.Printf("Route:Fulfill: %+v\n", fopts)
```
inancgumus: Forgotten?

internal/js/modules/k6/browser/browser/route_mapping.go
```diff
@@ -0,0 +1,64 @@
...
+				return nil, route.Fulfill(fopts)
+			})
+		},
+		"request": route.Request,
```
inancgumus: Is this mapped with `mapRequest`?

internal/js/modules/k6/browser/common/route.go
```diff
@@ -0,0 +1,73 @@
+package common
```
inancgumus: I'm not sure about this refactoring. I'd keep all HTTP-related code in `http.go` as before, as I don't mind about the file line length. No strong opinion, though.

internal/js/modules/k6/browser/common/network_manager.go
```diff
@@ -815,13 +821,12 @@ func (m *NetworkManager) Authenticate(credentials Credentials) error {
 	return nil
 }
 
-func (m *NetworkManager) AbortRequest(requestID fetch.RequestID, errorReason string) {
+func (m *NetworkManager) AbortRequest(requestID fetch.RequestID, errorReason string) error {
```
inancgumus, line 861: I'm not sure whether we should return an error from these functions. There are many requests, and failing one would kill the whole test run. It's good to fail fast, but in practice, it might not be best for user experience. That said, we can always revert this change if it becomes a problem later. I believe it's fine for now.

inancgumus (commented):
Good work, despite it being your first PRs 🙇. I've posted some suggestions. We should also fix the linter errors.

internal/js/modules/k6/browser/common/route.go
AgnesToulet: I'll revert this change! Now that I think again about this, I think it makes sense to keep it in `http.go`. And if the file size becomes an issue, it would probably make more sense to split it into three to have `response.go`, `request.go` and `route.go`. Reverted in https://github.com/grafana/k6/pull/4961/commits/2e31eae7c3baf6a277e0c564725c54f296e57216

internal/js/modules/k6/browser/browser/route_mapping.go
AgnesToulet: Removed in https://github.com/grafana/k6/pull/4961/commits/c4e1731b187aa53e6926750a9d604dca8fc96a87

internal/js/modules/k6/browser/browser/route_mapping.go
AgnesToulet: Fixed in https://github.com/grafana/k6/pull/4961/commits/53526d3e2346716a30d17d5ca6156b1caa60ec7a

internal/js/modules/k6/browser/common/route.go
inancgumus: ❤️ In that case, I'd suggest `http_response.go`, `http_request.go`, etc.

inancgumus (approved):

oleiade (dismissed):

internal/js/modules/k6/browser/common/frame_manager.go
```diff
@@ -566,7 +566,11 @@ func (m *FrameManager) requestStarted(req *Request) {
 		}
 		return
 	}
-	route.Continue()
+
+	if err := route.Continue(); err != nil {
+		m.logger.Warnf("FrameManager:requestStarted",
```
ankur22: Should this also be an error log, like it is in [onRequestPaused](https://github.com/grafana/k6/pull/4961/files#diff-7f7619be15080b7697b5d630cd5c2a14d5953c8496eabae91b1908dfcb371f53R626-R630)?

internal/js/modules/k6/browser/common/network_manager.go
```diff
@@ -872,24 +874,60 @@ func (m *NetworkManager) ContinueRequest(requestID fetch.RequestID) error {
...
-		WithBody(params.Body)
+func (m *NetworkManager) FulfillRequest(request *Request, opts *FulfillOptions) error {
+	responseCode := int64(http.StatusOK)
+	if opts != nil && opts.Status != 0 {
+		responseCode = opts.Status
+	}
+
+	action := fetch.FulfillRequest(request.interceptionID, responseCode)
+
+	if opts.ContentType != "" {
```
ankur22, line 921: Still need to check for `nil` `opts`? Maybe we need to check for `nil` opts at the beginning and `fulfill` with no extra options? Otherwise adding the options fields that aren't `nil`.

ankur22 (commented):
Nice! Looks good. Just a couple of things to consider first before we can merge this in.

internal/js/modules/k6/browser/common/frame_manager.go
inancgumus: I believe the other one better be a warning to reduce log clutter.

internal/js/modules/k6/browser/common/network_manager.go
inancgumus: Is there a reason for `parseFulfillOptions` to return a pointer: `*common.FulfillOption`? If not, we could just return a zero-value `FulfillOption`.

AgnesToulet: Yes, you're right, I fixed it in the following PR but can do it here (I'm focusing on merging page.route first, then update this PR with your feedback). I'll look into removing the pointer as well.

internal/js/modules/k6/browser/common/frame_manager.go
AgnesToulet: I think this one is important because if it fails, it means than the request is stalling and navigation will probably never ends. But at the same time, there is no user option here so I don't know what the user can do about this error (same for the other one in `onRequestPaused`, pretty important but not much users can do about it).

internal/js/modules/k6/browser/common/network_manager.go
```diff
@@ -895,32 +902,69 @@ func (m *NetworkManager) ContinueRequest(requestID fetch.RequestID) {
...
-		WithBody(params.Body)
+func (m *NetworkManager) FulfillRequest(request *Request, opts FulfillOptions) error {
+	responseCode := int64(http.StatusOK)
+	if opts.Status != 0 {
+		responseCode = opts.Status
+	}
+
+	action := fetch.FulfillRequest(request.interceptionID, responseCode)
+
+	if opts.ContentType != "" {
```
AgnesToulet, line 921: I'm wondering here... Is it better to keep it like this or actually remove this field from the `FulfillOptions` struct and directly add it to the `Headers` option field in `parseFulfillOptions` method, WDYT?

internal/js/modules/k6/browser/common/frame_manager.go
AgnesToulet: Updated in https://github.com/grafana/k6/pull/4961/commits/55db13065729ccd81042d0be119aea913fc2d104 for consistency. I have no strong opinion on the log level but I think it should not happen often and is important enough to be an error.

internal/js/modules/k6/browser/common/network_manager.go
AgnesToulet: Updated in https://github.com/grafana/k6/pull/4961/commits/4452b9e7a6bd7196a429689bd91ca7365035d56e

internal/js/modules/k6/browser/common/frame_manager.go
inancgumus: I'm okay with using the "error" level for logging. However, if this routing issue is caused by a non-mainframe navigation, such as a CSS route, it could lead to an increase in log entries. This was a significant problem in the past, as we were logging numerous errors that were not related to the success or failure of test runs.

internal/js/modules/k6/browser/common/network_manager.go
inancgumus: I believe it's better to be as close to Playwright as possible in UX terms.

AgnesToulet: Yes I agree, this would only be "under the hood". The API spec would stay the same but we wouldn't bother with this field in the other layers of the codebase.

internal/js/modules/k6/browser/browser/route_mapping.go
```diff
@@ -0,0 +1,62 @@
...
+	if !sobekValueExists(opts) {
+		return fopts
+	}
+
+	rt := k6ext.Runtime(ctx)
+	obj := opts.ToObject(rt)
+	for _, k := range obj.Keys() {
+		switch k {
+		case "body":
+			fopts.Body = obj.Get(k).String()
```
mstoykov: Same comment as I gave in the continue PR - this should use `common.ToBytes`  and handle this as binary data by the most part.

internal/js/modules/k6/browser/browser/route_mapping.go
```diff
@@ -0,0 +1,62 @@
...
+			fopts.Headers = make([]common.HTTPHeader, len(headersKeys))
+			for i, hk := range headersKeys {
+				fopts.Headers[i] = common.HTTPHeader{
+					Name:  hk,
+					Value: headers.Get(hk).String(),
+				}
+			}
+		case "status":
+			fopts.Status = obj.Get(k).ToInteger()
+		}
```
mstoykov, line 70: Looking at https://playwright.dev/docs/api/class-route#route-fulfill - you are not handling `path`, `json` and `response` . Can we have a comment here. Also I feel like handling all but `response` is likely fairly quick :thinking: .

But if not maybe lets throw an exception instead of silently doing the wrong thing

ankur22: I think we should be ok to wait for feedback for `response` and `json`.

Instead of working with `path` we should probably encourage users to work with the [current idiomatic k6](https://grafana.com/docs/k6/latest/javascript-api/init-context/open/) way of reading a file and then passing it the bytes to `body`.

So, yeah, happy with throwing an error if a user tries to work with those fields that aren't implemented.

oleiade (approved):

internal/js/modules/k6/browser/browser/route_mapping.go
AgnesToulet: Updated in https://github.com/grafana/k6/pull/4961/commits/67f24be3c8cb74ac463af2cf62248fecba3024e0

I agree that `json` should be quick to implement and used by a lot of users but wanted to keep the scope of the PR as small as possible. We can definitely iterate on it in future releases.

internal/js/modules/k6/browser/browser/route_mapping.go
AgnesToulet: Updated in https://github.com/grafana/k6/pull/4961/commits/adc5394b6e735ad94abce20f47bbc92d62b29cbf

ankur22 (approved):

PR #4970, opened 2025-07-25
Use `crypto/rand` instead of insecure `math/rand`

ankur22:
## What?

This is using `crypto/rand` instead of `math/rand`.

## Why?

`math/rand` raises a gosec issue.

## Checklist

### DIFF
--- a/internal/js/modules/k6/browser/chromium/browser_type.go
+++ b/internal/js/modules/k6/browser/chromium/browser_type.go
@@ -86,8 +86,10 @@ func (b *BrowserType) initContext(ctx context.Context) context.Context {
 	ctx = k6ext.WithCustomMetrics(ctx, b.k6Metrics)
 	ctx = common.WithHooks(ctx, b.hooks)
 	var buf [8]byte
-	rand.Read(buf[:])
-	ctx = common.WithIterationID(ctx, fmt.Sprintf("%x", buf[:]))
+	_, err := rand.Read(buf[:])
+	if err == nil {
+		ctx = common.WithIterationID(ctx, fmt.Sprintf("%x", buf[:]))
+	}
 	return ctx
 }
 


codebien (dismissed):

joanlopez (dismissed):

ankur22:
> I'm fine with this change. Although we haven't used this feature for the past year, it was very useful in identifying scalability and race condition issues on remote runs. PS: We can turn off the gosec warning by switching to crypto/rand.

Yeah, i was happy with removing it, but then I wondered whether it might still be useful to track a single iteration in a load test. Could we work with something like this instead of rand? It gives us more information than just a random string too:

```go
itID := "v:" + strconv.FormatInt(b.vu.State().VUID, 10) + "i:" + strconv.FormatInt(b.vu.State().Iteration, 10)
```

inancgumus:
@ankur22 That solution works well if VU IDs are unique across instances. However, if they are not, using `rand` provides a better ID for distinguishing iterations across instances, which is particularly useful for Chromium-related issues, since we might share Chromium instances. Perhaps we could introduce a bit of randomness to the `itID`, such as adding some random letters at the beginning, to address this concern.

codebien (commented):
Is this generated for each iteration? If we are looking for ways for improving performance then this sounds low-hanging fruit to address. If we really need it, can't we reuse the same value that we use for `iterationInTest` https://grafana.com/docs/k6/latest/javascript-api/k6-execution/#getting-unique-data-once?

ankur22:
>Is this generated for each iteration? If we are looking for ways for improving performance this sounds low-hanging fruit for improvements. If we really need it, can't we reuse the same value that we use for iterationInTest https://grafana.com/docs/k6/latest/javascript-api/k6-execution/#getting-unique-data-once?

That's what [I was going to work with](https://github.com/grafana/k6/pull/4970#issuecomment-3119140686) but when we're debugging with multiple instances all running many VUs and iterations then this could be [difficult to correlate](https://github.com/grafana/k6/pull/4970#issuecomment-3120576105). This is why i opted to not change the behaviour.

mstoykov:
The original idea was to remove it if it isn't used or useful or to replace it with ItearationInTest - which is unique for each scenario across instances and VUs. 

The idea isn't to use crypto/rand for things that do not need to be cryptographically safe at all.

ankur22:
> The original idea was to remove it if it isn't used or useful or to replace it with ItearationInTest - which is unique for each scenario across instances and VUs.
> 
> The idea isn't to use crypto/rand for things that do not need to be cryptographically safe at all.

Yep, i see your point here and I was wondering whether this would be picked up.

Can we just leave this as is then since we're not doing anything security related with `math/rand`?

PR #4971, opened 2025-07-25
Browser: add route.continue

AgnesToulet:
## What?

This PR adds the `route.continue` function. Used in `page.route`, it allows to continue a request with modified: header, body, method or URL. 

## Why?

This helps with Playwright compatibility and allow to add headers to specific request (especially useful when dealing with CORS or authentication issues).

## Checklist

### DIFF
--- a/examples/browser/route_continue.js
+++ b/examples/browser/route_continue.js
@@ -1,13 +1,13 @@
-import { check, sleep } from "k6";
-import { browser } from "k6/browser";
+import { check, sleep } from 'k6';
+import { browser } from 'k6/browser';
 
 export const options = {
   scenarios: {
     ui: {
-      executor: "shared-iterations",
+      executor: 'shared-iterations',
       options: {
         browser: {
-          type: "chromium",
+          type: 'chromium',
         },
       },
     },
@@ -17,25 +17,27 @@ export const options = {
 export default async function () {
   const page = await browser.newPage();
 
-  await page.route(/.*\/api\/pizza/, function (route) {
-    route.continue({
-      postData: JSON.stringify({
-        customName: "My Pizza",
-      }),
+  try {
+    await page.route(/.*\/api\/pizza/, function (route) {
+      route.continue({
+        postData: JSON.stringify({
+          customName: 'My Pizza',
+        }),
+      });
     });
-  });
 
-  await page.goto("https://quickpizza.grafana.com/");
+    await page.goto('https://quickpizza.grafana.com/');
 
-  await page.getByRole("button", { name: "pizza, please" }).click();
-  sleep(1);
+    await page.getByRole('button', { name: 'pizza, please' }).click();
+    sleep(1);
 
-  const checkData = await page
-    .locator('//p[text()[contains(., "Name:")]]')
-    .innerText();
-  check(page, {
-    pizzaName: checkData === "Name: My Pizza",
-  });
-
-  await page.close();
+    const checkData = await page
+      .locator('//p[text()[contains(., "Name:")]]')
+      .innerText();
+    check(page, {
+      pizzaName: checkData === 'Name: My Pizza',
+    });
+  } finally {
+    await page.close();
+  }
 }
--- a/examples/browser/route_continue.js
+++ b/examples/browser/route_continue.js
@@ -1,0 +1,41 @@
+import { check, sleep } from "k6";
+import { browser } from "k6/browser";
+
+export const options = {
+  scenarios: {
+    ui: {
+      executor: "shared-iterations",
+      options: {
+        browser: {
+          type: "chromium",
+        },
+      },
+    },
+  },
+};
+
+export default async function () {
+  const page = await browser.newPage();
+
+  await page.route(/.*\/api\/pizza/, function (route) {
+    route.continue({
+      postData: JSON.stringify({
+        customName: "My Pizza",
+      }),
+    });
+  });
+
+  await page.goto("https://quickpizza.grafana.com/");
+
+  await page.getByRole("button", { name: "pizza, please" }).click();
+  sleep(1);
+
+  const checkData = await page
+    .locator('//p[text()[contains(., "Name:")]]')
+    .innerText();
+  check(page, {
+    pizzaName: checkData === "Name: My Pizza",
+  });
+
+  await page.close();
+}
--- a/examples/browser/route_continue.js
+++ b/examples/browser/route_continue.js
@@ -1,5 +1,7 @@
-import { check, sleep } from 'k6';
+import { sleep } from 'k6';
 import { browser } from 'k6/browser';
+import { check } from 'https://jslib.k6.io/k6-utils/1.5.0/index.js';
+
 
 export const options = {
   scenarios: {
@@ -31,11 +33,9 @@ export default async function () {
     await page.getByRole('button', { name: 'pizza, please' }).click();
     sleep(1);
 
-    const checkData = await page
-      .locator('//p[text()[contains(., "Name:")]]')
-      .innerText();
-    check(page, {
-      pizzaName: checkData === 'Name: My Pizza',
+    const e = page.getByText('Name:')
+    check(null, {
+      pizzaName: await e.innerText() === 'Name: My Pizza',
     });
   } finally {
     await page.close();
--- a/internal/js/modules/k6/browser/browser/route_mapping.go
+++ b/internal/js/modules/k6/browser/browser/route_mapping.go
@@ -20,8 +20,8 @@ func mapRoute(vu moduleVU, route *common.Route) mapping {
 			})
 		},
 		"continue": func(opts sobek.Value) *sobek.Promise {
+			copts := parseContinueOptions(vu.Context(), opts)
 			return k6ext.Promise(vu.Context(), func() (any, error) {
-				copts := parseContinueOptions(vu.Context(), opts)
 				return nil, route.Continue(copts)
 			})
 		},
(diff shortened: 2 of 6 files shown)


examples/browser/route_continue.js
```diff
@@ -0,0 +1,43 @@
...
+
+    await page.goto('https://quickpizza.grafana.com/');
+
+    await page.getByRole('button', { name: 'pizza, please' }).click();
+    sleep(1);
+
+    const checkData = await page
+      .locator('//p[text()[contains(., "Name:")]]')
+      .innerText();
+    check(page, {
```
joanlopez: Nit; why do we pass `page` as the `check` value, if it isn't really used? 🤔

joanlopez (dismissed):

examples/browser/route_continue.js
AgnesToulet: Improved in https://github.com/grafana/k6/pull/4971/commits/0c3b9716344834421c166d749c067e4dbf24de35 (also tried to align a bit more with the existing scripts and used the new `getByText` function).

internal/js/modules/k6/browser/common/http.go
```diff
@@ -718,6 +718,14 @@ type Route struct {
 	handled bool
 }
 
+// ContinueOptions are request fields that can be overridden when continuing a request.
+type ContinueOptions struct {
+	Headers  []HTTPHeader
+	Method   string
+	PostData string
```
mstoykov: This likely should be `[]byte` and get converted from string or extracted from an ArrayBuffer. 

See https://pkg.go.dev/go.k6.io/k6/js/common#ToBytes

internal/js/modules/k6/browser/common/network_manager.go
```diff
@@ -881,10 +881,29 @@ func (m *NetworkManager) AbortRequest(requestID fetch.RequestID, errorReason str
...
+	}
+	if opts.URL != "" {
+		action = action.WithURL(opts.URL)
+	}
+	if opts.Method != "" {
+		action = action.WithMethod(opts.Method)
+	}
+	if opts.PostData != "" {
+		b64PostData := base64.StdEncoding.EncodeToString([]byte(opts.PostData))
+		action = action.WithPostData(b64PostData)
```
mstoykov: I commented in the other place, but if this is the only place this is used, I guess we can directly base64 encode it and store it like that.

If that is a thing the user can access, IMO you might need more code around it being accessed correctly.

internal/js/modules/k6/browser/common/network_manager.go
```diff
@@ -881,10 +881,29 @@ func (m *NetworkManager) AbortRequest(requestID fetch.RequestID, errorReason str
...
+	opts ContinueOptions,
+	originalHeaders []HTTPHeader,
+) error {
 	m.logger.Debugf("NetworkManager:ContinueRequest", "continuing request (id: %s)", requestID)
-
 	action := fetch.ContinueRequest(requestID)
+
+	if len(opts.Headers) > 0 {
+		allHeaders := mergeHeaders(originalHeaders, opts.Headers)
+		action = action.WithHeaders(allHeaders)
```
mstoykov: Looking at the example in https://playwright.dev/docs/api/class-route#route-continue - if headers are provided it is overriding them. Otherwise it woulnd't be copying the headers ?!?!

There is also some comment at the end for Cookie ... which might actually just be that browsers will do this irregardless   :thinking:

mstoykov (changes requested):
In general looks fine, but this seems to expect string bodies only, which I prefer to fix now, before it becomes a problem.

Also looking at hte example in the playwright docs it seems like the handling of headers wouldn't match their behaviour? Or is there some docs I am missing?

internal/js/modules/k6/browser/common/http.go
ankur22: Why is that @mstoykov? I'd like to understand too incase i've made a mistake elsewhere.

mstoykov: From https://playwright.dev/docs/api/class-route#route-continue - it can take binary data as input as ArrayBuffer.

If you try to safe binary data as string .. it might work, or it might mangle it a little bit (or a lot). As such in general we prefer to keep it as a slice of bytes. Sobek/goja (used to(?))keeps some string as UTF16 for compatibility with the standard. So in a lot of the cases we will have copy/conversion either way.

internal/js/modules/k6/browser/common/network_manager.go
AgnesToulet: Do you mean "encoding it in the `parseContinueOptions` function"? I don't think this is something the user should access (except if we implement the `fallback` function but this will be accessed through the `request` object so... code will have to change anyway).

internal/js/modules/k6/browser/common/http.go
AgnesToulet: Updated in https://github.com/grafana/k6/pull/4971/commits/05f4c7e2edff106981338efdbfd84328ea2d9a15 :+1: If it looks good, I'll do the same for the `fulfill` PR.

internal/js/modules/k6/browser/common/network_manager.go
AgnesToulet: Right, I got confused because these are called "overrides" in PW codebase but you're right. Updated in https://github.com/grafana/k6/pull/4971/commits/591b9a84a691cfffb82e09e918724b7b13d863cc

Regarding cookie, they do have a specific filter in their codebase to remove it (https://github.com/microsoft/playwright/blob/main/packages/playwright-core/src/server/network.ts#L352-L353) I wasn't sure why to be honest.

internal/js/modules/k6/browser/common/http.go
joanlopez: Thanks for bringing this topic @mstoykov, as at some point while reviewing this pull request, I wondered if `string` would be good for all cases, but then I forgot to express my doubts.

Yeah, I think that if we speak about data in an HTTP POST request's body, it's safer to use raw/bytes data, as I'd expect the transformation from `string` to `bytes`(binary) be always safe but not the other way round, as you said.

internal/js/modules/k6/browser/common/network_manager.go
mstoykov: I guess - yes, but I do think it is probably fine not to do it as well. Just a random though I had.

 Arguably having the body as is and then encoding it as the protocol requires seems way more logical. Than haivng it in the exact format we will use it later, just because we do know this in advance.

Given that we have to encode this either way, I doubt there is any performance difference.

internal/js/modules/k6/browser/common/network_manager.go
```diff
@@ -881,10 +881,28 @@ func (m *NetworkManager) AbortRequest(requestID fetch.RequestID, errorReason str
...
+	requestID fetch.RequestID,
+	opts ContinueOptions,
+	originalHeaders []HTTPHeader,
+) error {
 	m.logger.Debugf("NetworkManager:ContinueRequest", "continuing request (id: %s)", requestID)
-
 	action := fetch.ContinueRequest(requestID)
+
+	if len(opts.Headers) > 0 {
+		action = action.WithHeaders(toFetchHeaders(opts.Headers))
```
mstoykov, line 893: `Cookie` is still not filtered. I do not know if want to do this as Playwright, whether we need to do something else as well. 

cc @ankur22 if you have more info on that

internal/js/modules/k6/browser/browser/route_mapping.go
```diff
@@ -72,3 +101,15 @@ func parseFulfillOptions(ctx context.Context, opts sobek.Value) (common.FulfillO
...
+	headersKeys := headers.Keys()
+	result := make([]common.HTTPHeader, len(headersKeys))
+	for i, hk := range headersKeys {
+		result[i] = common.HTTPHeader{
+			Name:  hk,
+			Value: headers.Get(hk).String(),
+		}
+	}
+	return result
+}
```
mstoykov, line 115: The example they have in the [docs](https://playwright.dev/docs/api/class-route#route-continue) specifically use `undefined` as a way to remove a key that is used. But that leaves the value *as* undefined.

You can run the following in your browser console:
```javascript
const origHeaders = {bar: "something", else: "here"}
  const headers = {
    ...origHeaders,
    foo: 'foo-value', // set "foo" header
    bar: undefined, // remove "bar" header
  };
  console.log(headers)
```
to get
```
Object { bar: undefined, else: "here", foo: "foo-value" }
```

Maybe we should filter here any value that is `undefined` here as well.

mstoykov (dismissed):
LGTM!

I have left two comments around headers, and there is compilation errors in tests due to types.

The latter needs to be fixed the other, we can probably kick down the road, but still nice to comment on.

internal/js/modules/k6/browser/common/network_manager.go
ankur22: To help this PR go through before the release, i'd be happy to make this change in another PR once we have done some more research around why we need to filter out the cookie header. I unfortunately haven't done enough research myself to know.

internal/js/modules/k6/browser/browser/route_mapping.go
ankur22: I think this a good point, and probably worth doing soon. However, to help this PR go through before the release, i'd be happy to make this change in another PR. Hope that's ok with everyone.

ankur22 (approved):

internal/js/modules/k6/browser/browser/route_mapping.go
AgnesToulet: I started to work on another branch for this, it makes sense but if it brings more discussion, I'd prefer to merge this PR without this update.

internal/js/modules/k6/browser/common/network_manager.go
AgnesToulet: Agree, here's the origin of the change in Playwright with more info: https://github.com/microsoft/playwright/pull/35168

tldr. It seems that it's to align behaviors between browsers. As we only support Chromium for now, it doesn't really impact us. The important part though is that, in Chromium, cookies are not overridden if there is a cookie in the cookie jar for given origin so users can think they are setting a cookie but it's actually not used. IMO this should be documented rather than just removed from the request.

internal/js/modules/k6/browser/browser/route_mapping.go
AgnesToulet: See https://github.com/grafana/k6/pull/5004

internal/js/modules/k6/browser/common/network_manager.go
mstoykov: Lets open an issue than

PR #4976, opened 2025-07-28
Fix metric results alignment in end of test summary

oleiade:
## What?

This Pull Request makes sure the metric results of the end of test summary follow a global, shortest possible, alignment to reduce the horizontal space consumed, and make reading the results easier.

CleanShot 2025-07-28 at 11 45 54

This was done by computing the alignment value (how many dots should be printed before the `: {results}`) globally, and cascading it down to each specific renderers. The minimum number of dot separators has been set to 3, which by my testing "feels" the best and most liable option.

Importantly, the human-readable end of test summary is explicitly excluded from verisioning policy, and **this is not considered a breaking change**. Tentatively putting this in the v1.2.0 roadmap, as it is a rather small change, but happy to drop it if reviews uncover more work 🙇🏻 

## Why?

It was reported to us recently, that the end of test summary would sometimes now keep the metric results properly aligned, and would use more horizontal spacing than necessary.

unaligned-eot

We noticed that in certain more advanced cases, including using `summary-mode=full`, this would lead to unexpected alignment patterns, when using groups and scenarios. Most of them caused by alignment being computed locally to each section rather than globally.

## Checklist

### DIFF
--- a/internal/js/summary.js
+++ b/internal/js/summary.js
@@ -171,6 +171,9 @@ class TestReportGenerator {
 			this.renderContext,
 			this.options,
 		);
+		// Compute global alignment before building sections
+		reportBuilder._computeGlobalMaxNameWidth(report);
+		
 		return reportBuilder
 			.addThresholds(report.thresholds)
 			.addTotalResults(report)
@@ -196,6 +199,7 @@ class ReportBuilder {
 		this.renderContext = renderContext;
 		this.options = options;
 		this.sections = [];
+		this.globalMaxNameWidth = null;
 	}
 
 	/**
@@ -225,7 +229,7 @@ class ReportBuilder {
 			title: 'TOTAL RESULTS',
 			content: [
 				...this._renderChecks(report.checks),
-				...this._renderMetrics(report.metrics),
+				...this._renderMetrics(report.metrics, this.renderContext, this.globalMaxNameWidth),
 			],
 		});
 		return this;
@@ -268,6 +272,64 @@ class ReportBuilder {
 				});
 			});
 		return this;
+	}
+
+	/**
+	 * Computes the global maximum name width across all metrics in all sections.
+	 * This ensures consistent colon alignment across the entire report.
+	 *
+	 * @param {Report} report - The complete report object
+	 * @returns {void}
+	 */
+	_computeGlobalMaxNameWidth(report) {
+		const allMetrics = {};
+		const renderContext = this.renderContext.indentedContext(1);
+
+		// Collect metrics from total results
+		if (report.metrics) {
+			Object.entries(report.metrics).forEach(([_, sectionMetrics]) => {
+				Object.assign(allMetrics, sectionMetrics);
+			});
+		}
+
+		// Collect metrics from groups
+		if (report.groups) {
+			this._collectMetricsFromGroups(report.groups, allMetrics);
+		}
+
+		// Collect metrics from scenarios
+		if (report.scenarios) {
+			this._collectMetricsFromGroups(report.scenarios, allMetrics);
+		}
+
+		// Compute the global max name width
+		let maxNameWidth = 0;
+		Object.keys(allMetrics).forEach((name) => {
+			const displayName = renderContext.indent(renderMetricDisplayName(name));
+			maxNameWidth = Math.max(maxNameWidth, strWidth(displayName));
+		});
+
+		this.globalMaxNameWidth = maxNameWidth;
+	}
+
+	/**
+	 * Recursively collects metrics from groups/scenarios and their nested groups.
+	 *
+	 * @param {Record<string, ReportGroup>} groups - The groups to collect metrics from
+	 * @param {Object} allMetrics - The accumulator object for all metrics
+	 * @returns {void}
+	 */
+	_collectMetricsFromGroups(groups, allMetrics) {
+		Object.values(groups).forEach((group) => {
+			if (group.metrics) {
+				Object.entries(group.metrics).forEach(([_, sectionMetrics]) => {
+					Object.assign(allMetrics, sectionMetrics);
+				});
+			}
+			if (group.groups) {
+				this._collectMetricsFromGroups(group.groups, allMetrics);
+			}
+		});
 	}
 
 	/**
@@ -326,10 +388,11 @@ class ReportBuilder {
 	/**
 	 * @param {ReportMetrics} metrics - The metrics to render.
 	 * @param {RenderContext} [renderContext] - The render context to use for text rendering.
+	 * @param {number} [globalMaxNameWidth] - Optional global max name width for cross-section alignment.
 	 * @returns {string[]}
 	 * @private
 	 */
-	_renderMetrics(metrics, renderContext) {
+	_renderMetrics(metrics, renderContext, globalMaxNameWidth) {
 		renderContext = renderContext || this.renderContext;
 		renderContext = renderContext.indentedContext(1);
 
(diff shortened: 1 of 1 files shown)


oleiade:
cc @dgzlopes this should be tackling your report 🙇🏻 
cc @joanlopez as we initially worked on this together.

internal/js/summary.js
```diff
@@ -196,6 +199,7 @@ class ReportBuilder {
 		this.renderContext = renderContext;
 		this.options = options;
 		this.sections = [];
+		this.globalMaxNameWidth = null;
```
inancgumus: Should this be `0`?

internal/js/summary.js
```diff
@@ -270,6 +274,64 @@ class ReportBuilder {
...
+	 * @returns {void}
+	 */
+	_computeGlobalMaxNameWidth(report) {
+		const allMetrics = {};
+		const renderContext = this.renderContext.indentedContext(1);
+
+		// Collect metrics from total results
+		if (report.metrics) {
+			Object.entries(report.metrics).forEach(([_, sectionMetrics]) => {
+				Object.assign(allMetrics, sectionMetrics);
```
inancgumus: Is the overwrite (by object keys) here intentional?

inancgumus (commented):
I have some mild questions. We should fix the tests.

internal/js/summary.js
oleiade: It probably should, indeed 👍🏻 Fixing it 🚀

internal/js/summary.js
oleiade: It is, the objective here is to collect all metrics info (uniquely) in a single object, so we can later 👇🏻 make a pass over it to find the longest name. Thus if there are duplicate entries, we're happy to have them overwritten to keep a single one.

Let me know if I misunderstood your question though 🙇🏻

mstoykov:
```suggestion
			allMetrics = Object.values(report.metrics)
```

Pretty sure this is the same ?

mstoykov: Actually as we want the values are arrays.

```suggestion
			Object.assign(allMetrics, ...Object.values(report.metrics))
```

internal/js/summary.js
```diff
@@ -270,6 +274,64 @@ class ReportBuilder {
...
+	 * @param {Record<string, ReportGroup>} groups - The groups to collect metrics from
+	 * @param {Object} allMetrics - The accumulator object for all metrics
+	 * @returns {void}
+	 */
+	_collectMetricsFromGroups(groups, allMetrics) {
+		Object.values(groups).forEach((group) => {
+			if (group.metrics) {
+				Object.entries(group.metrics).forEach(([_, sectionMetrics]) => {
+					Object.assign(allMetrics, sectionMetrics);
+				});
```
mstoykov:
```suggestion
				Object.assign(allMetrics, ...Object.values(group.metrics))
```

internal/js/summary.js
```diff
@@ -270,6 +274,64 @@ class ReportBuilder {
...
+	/**
+	 * Recursively collects metrics from groups/scenarios and their nested groups.
+	 *
+	 * @param {Record<string, ReportGroup>} groups - The groups to collect metrics from
+	 * @param {Object} allMetrics - The accumulator object for all metrics
+	 * @returns {void}
+	 */
+	_collectMetricsFromGroups(groups, allMetrics) {
+		Object.values(groups).forEach((group) => {
+			if (group.metrics) {
```
mstoykov, line 318: Pretty sure all of those ifs are unnecessary as it will just skip the operations. 

Maybe we can remove them and see how it goes instead of having additional lines and identation?

internal/js/summary.js
```diff
@@ -270,6 +274,64 @@ class ReportBuilder {
...
+		}
+
+		// Compute the global max name width
+		let maxNameWidth = 0;
+		Object.keys(allMetrics).forEach((name) => {
+			const displayName = renderContext.indent(renderMetricDisplayName(name));
+			maxNameWidth = Math.max(maxNameWidth, strWidth(displayName));
+		});
+
+		this.globalMaxNameWidth = maxNameWidth;
```
mstoykov:
```suggestion
		this.globalMaxNameWidth = Object.keys(allMetrics).reduce((max, name) => {
			return Math.max(max, strWidth(renderContext.indent(renderMetricDisplayName(name))));
		})
```

Not certain it won't be more readable with map to get the function and then a reduce or just using Math.max on the resulting array from the map. But this likely is the fastest :shrug:

internal/js/summary.js
```diff
@@ -270,6 +274,64 @@ class ReportBuilder {
...
 
+	/**
+	 * Computes the global maximum name width across all metrics in all sections.
+	 * This ensures consistent colon alignment across the entire report.
+	 *
+	 * @param {Report} report - The complete report object
+	 * @returns {void}
+	 */
+	_computeGlobalMaxNameWidth(report) {
+		const allMetrics = {};
```
mstoykov, line 285:
```suggestion
		const metricNames = {};
```

It is only the metric names after all

mstoykov (approved):
LGTM!

I have left some nit picks and golfing changes to reduce the size and arguably make it more functional... hopefully also easier to read.

internal/js/summary.js
joanlopez: > only the metric names

Are you suggesting to do it? Because, unless I'm wrong, I think the current code is collecting the entire metric objects, with their values, etc.

But yeah, I do agree that collecting names here is likely enough, as it's the only metric's attribute used after all.

internal/js/summary.js
```diff
@@ -270,6 +274,64 @@ class ReportBuilder {
...
+
+		// Collect metrics from scenarios
+		if (report.scenarios) {
+			this._collectMetricsFromGroups(report.scenarios, allMetrics);
+		}
+
+		// Compute the global max name width
+		let maxNameWidth = 0;
+		Object.keys(allMetrics).forEach((name) => {
+			const displayName = renderContext.indent(renderMetricDisplayName(name));
```
joanlopez: Do we really need to use `renderContext` here? In my opinion it's confusing because this code is treating all the metrics (total results, groups, scenarios, etc) as the same, but each level will be later indented differently.

As the purpose is to calculate the longest metric name, I suspect we could just omit the indentation for this calculus.

internal/js/summary.js
joanlopez: I think we would need something like: `Object.entries(group.metrics ?? {})`, as it would throw a `TypeError` if we pass an `undefined` to `Object.entries` 🤔

internal/js/summary.js
joanlopez: I did a quick test, and apparently one of the metrics from total results (and a couple of tags), doesn't get properly aligned. So, feel free to leave it. But anyway, I'm confused for what I mentioned above.

joanlopez: Indeed, after a few more tests, if you nest enough groups, the alignment breaks.

In practical terms, the current code is likely "good enough", but if you find an easy way to adequately take into account the identantion of each level, it would be great.

I guess you would something like slightly modifying the current heuristics, by just collecting the metric names from groups and scenarios, already indented at "collection time", so when you calculate the max you don't need to care about from what level of nesting a metric name comes from.

joanlopez: Another valid approach, likely much simpler, would be to only align each section (total results, scenario and group) individually. That would solve the issue described above as well, as I think the key difference here is that you're using the longest metric name plus three dots as the reference length, which is what really prevents "too many dots" and stuff overflowing the length of a single line.

Indeed, if you have multiple nested groups, the current approach adds unnecessary dots, and aligns total results more to the right only because of the fact that below there are some nested groups.

I think it's a tradeoff, where I'm happy with both, as long as the implementation is correct and looks good. Up to you!

joanlopez (dismissed):
Approving, because despite the minor suggestions I left, and what I pointed out in https://github.com/grafana/k6/pull/4976/files#r2239337215, I think this changeset is _almost always_ correct in practical terms 👍🏻

internal/js/summary.js
inancgumus: You got the question correctly. I wanted to point out if there was a mistake 👍

inancgumus (approved):

PR #4981, opened 2025-07-29
test: Add test for gRPC proto loading in separate process

### LINKED ISSUE #3552: Run tests in separate process so that we actually test loading proto definitions at runtime
Originally raised in this [comment](https://github.com/grafana/k6/issues/3160#issuecomment-1620990687) by @thiagodpf. 

### What?

The current unit tests run the server and clients in the same process.

gRPC seems to have a global registry for types, so once the server loads them, the client doesn't need to.

This means that all tests that test that clients actually manage to load types correctly only with `.proto` files are not actually testing that :(

### Solutions:

Preferably `go test ./...` should run the unit tests with the server and client as separate process - preferably even separate from the main `go test` one :thinking: 

This sounds involved.

A quicker solution will be to have another CI step that runs a server + `k6 run script.js` exercising the most common cases.

Both will be best.

paulnegz:
This test verifies that k6 can properly load proto definitions at runtime when the server and client are in separate processes. This ensures that we're not relying on the global registry being populated by the server process.

Fixes #3552

## What?

This PR adds a new test (`TestGRPCSeparateProcess`) that:
- Starts a gRPC server in a separate process.
- Runs a k6 script in another process, which loads the proto file at runtime and connects to the server.
- Verifies that the client can successfully invoke a gRPC method using only the proto definition, ensuring that type registration is not leaking between processes.

## Why?

Previously, gRPC tests ran both the server and client in the same process, which could mask issues with proto type registration due to Go's global registry. This test ensures that k6's gRPC client can independently load and use proto definitions at runtime, matching real-world usage and preventing false positives in tests.

## Checklist

## Related PR(s)/Issue(s)

- Fixes #3552


mstoykov:
Hi @paulnegz, Thank you for the PR :bow: 

I have not looked into this too closely, but as a first though I am wondering if we can reuse one of the already available grpc services like the one under testutils, instead of making a new one. There are arguably already quite a few of them, I would prefer if we keep them more at bay :)

paulnegz:
> Hi @paulnegz, Thank you for the PR 🙇
> 
> I have not looked into this too closely, but as a first though I am wondering if we can reuse one of the already available grpc services like the one under testutils, instead of making a new one. There are arguably already quite a few of them, I would prefer if we keep them more at bay :)

done. Also worked on a little refactor to remove redundant logic. Please let me know if you have more changes to more forward with this. Do I need to squash commits to a single one?

ankur22:
Hi @paulnegz, thanks for contributing with these tests, we appreciate the effort you have to already 🙇

We are going to close this PR for now as it's been a month since we last heard from you. Please do reopen it with the requested changes and any other relevant updates you think might be needed when you have some time.

PR #4990, opened 2025-07-31
chore: Fix race in timers test due to context reset

mstoykov:
## What?

Sometimes the test will race and reset the context, before it has fully finished. This test makes certain that is not the case.

## Why?
Races in tests fail CI. 

## Checklist

### DIFF
--- a/internal/js/tc55/timers/timers_test.go
+++ b/internal/js/tc55/timers/timers_test.go
@@ -2,7 +2,6 @@ package timers_test
 
 import (
 	"context"
-	"sync"
 	"testing"
 	"time"
 
@@ -162,12 +161,13 @@ func TestSetTimeoutContextCancel(t *testing.T) {
 		runtime.VU.CtxField = ctx //nolint:fatcontext
 		runtime.VU.RuntimeField.ClearInterrupt()
 		const interruptMsg = "definitely an interrupt"
-		wg := sync.WaitGroup{}
-		defer wg.Wait()
-		defer cancel()
-		wg.Add(1)
+		sync := make(chan struct{})
+		defer func() {
+			cancel()
+			<-sync
+		}()
 		go func() {
-			defer wg.Done()
+			defer close(sync)
 			select {
 			case <-interruptChannel:
 			case <-ctx.Done():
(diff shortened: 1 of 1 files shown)


AgnesToulet (approved):

internal/js/tc55/timers/timers_test.go
```diff
@@ -154,14 +155,25 @@ func TestSetTimeoutContextCancel(t *testing.T) {
 		}
 	}))
 
-	for i := 0; i < 2000; i++ {
+	for range 2000 {
```
joanlopez, line 158: I'm not sure I do get the purpose of every single line of this test's code, but looks like the data race is still there, based on https://github.com/grafana/k6/actions/runs/17580308829/job/49934957577?pr=5157 
cc/ @mstoykov

Some thoughts:
- Looks like all the lines around `log` and the `"print"` function defined in JS is dead code, isn't it?
- Wouldn't it be enough to call `cancel()` inside the goroutine, instead of `runtime.CancelContext()`?

The way I get it is that, when the system is slow, we have at least one goroutine calling that line (`runtime.CancelContext()`) when the main goroutine (the main test function) is still running iterations, and thus overwriting `runtime.CancelContext()` value. 

Overall, also considering that we use the same `interruptChannel` for all the goroutines/iterations, I feel like that it's completely unpredictable to know which one cancels/interrupts what. That's why I said above that I'm not quite sure I fully understand this, and also because if I remove some lines from this test, it still passes, so perhaps expectations aren't clear.

Going further, I've been tempted to add something like:

```go
const total = 2000
done := atomic.Int32{}
defer func() {
	require.Equal(t, total, int(done.Load()))
}()
```
where the atomic counter would be incremented as the very last thing on the deferred function, right after receiving the `sync` signal. 

But again, this would always be satisfied as the first line in the main deferred function cancels the context, so if not resolved yet, the goroutine is resolved then (through `<-ctx.Done()`), and then the `sync` signal is emitted, by closing the channel in the function deferred inside the goroutine.

----

All that said, I'd be happy to open a PR, but to be honest, I'm not quite sure what's the appropriate fix here.
What do you think?

Would you describe what's the purpose (and thus, the assertions) of this test in more or less natural language/function/user-facing words?

Thanks!

PR #5002, opened 2025-08-01
Add/wait for response

### LINKED ISSUE #4339: Implement page.waitForResponse
### Feature Description

This is similar to how `page.on('response')` works (check https://github.com/grafana/k6/issues/4234 for details). What this adds is the ability to wait for the response to be received before carrying on with the test. This helps ensure that the response must have been received before carrying on with the request, as well as being able to dynamically/programmatically being able to perform fine grained checks on the responses of dependencies.

https://playwright.dev/docs/api/class-page#page-wait-for-response

https://github.com/grafana/k6/issues/4234

HasithDeAlwis:
Based on the implementation and testing we've done, here's a complete PR description:

## What?

This PR adds the `waitForResponse` method to the k6 browser module, allowing users to wait for HTTP responses that match specific URL patterns during browser automation.

The implementation includes:
- Core `waitForResponse` functionality in Frame and Page classes
- JavaScript API bindings with sobek.Promise support
- Support for exact URL matching, regex patterns, and empty patterns (match any response)
- Proper timeout handling and event cleanup
- Integration tests following existing repository patterns

## Why?

Browser automation often requires waiting for specific network responses before proceeding with test actions. This is particularly important for:
- **API Testing**: Waiting for AJAX/fetch requests to complete
- **Performance Testing**: Measuring response times for critical resources
- **E2E Testing**: Ensuring data loading is complete before assertions
- **Playwright Compatibility**: Matching Playwright's `page.waitForResponse()` API for easier migration

This feature complements the existing `waitForURL` method by focusing on HTTP responses rather than navigation events.

## Implementation Details

**URL Pattern Matching:**
```javascript
// Exact URL match
await page.waitForResponse('https://api.example.com/data');

// Regex pattern
await page.waitForResponse(/\/api\/.*\.json$/);

// Match any response
await page.waitForResponse('');
```

**Note:** This PR implements URL pattern matching for `waitForResponse`. A future PR will add support for predicate functions (e.g., `await page.waitForResponse(response => response.status() === 200)`) as the current implementation is already quite substantial.

## Checklist

## Related PR(s)/Issue(s)

This implements similar functionality to Playwright's `page.waitForResponse()` method, enhancing k6's browser automation capabilities for modern web application testing.

---

This PR description highlights the key functionality while noting that predicate functions (JavaScript functions as response matchers) will come in a future PR, keeping the scope manageable for reviewers.

## Summary by Sourcery

Add waitForResponse API to k6 browser module to enable waiting for HTTP responses matching URL patterns (string, regex, or wildcard) with timeout support and event handler cleanup

New Features:
- Implement Frame.waitForResponse and Page.waitForResponse methods to await HTTP responses by URL pattern
- Support exact URL, regex, and empty-pattern matching with injected JS regex checker
- Add FrameWaitForResponseOptions for configurable timeouts and proper timeout error handling
- Expose JavaScript API bindings for page.waitForResponse and frame.waitForResponse

Tests:
- Add integration tests for waitForResponse covering exact match, regex, wildcard, and timeout scenarios
- Extend mapping tests to include waitForResponse API bindings on Page and Frame

## Related PR(s)/Issue(s)
Closes #4339

### DIFF
--- a/internal/js/modules/k6/browser/browser/frame_mapping.go
+++ b/internal/js/modules/k6/browser/browser/frame_mapping.go
@@ -448,6 +448,31 @@ func mapFrame(vu moduleVU, f *common.Frame) mapping {
 				return nil, f.WaitForURL(val, popts, jsRegexChecker)
 			}), nil
 		},
+		"waitForResponse": func(url sobek.Value, opts sobek.Value) (*sobek.Promise, error) {
+			popts := common.NewFrameWaitForResponseOptions(f.Timeout())
+			if err := popts.Parse(vu.Context(), opts); err != nil {
+				return nil, fmt.Errorf("parsing waitForResponse options: %w", err)
+			}
+
+			var val string
+			switch url.ExportType() {
+			case reflect.TypeOf(string("")):
+				val = fmt.Sprintf("'%s'", url.String()) // Strings require quotes
+			default: // JS Regex, CSS, numbers or booleans
+				val = url.String() // No quotes
+			}
+
+			// Inject JS regex checker for URL pattern matching
+			ctx := vu.Context()
+			jsRegexChecker, err := injectRegexMatcherScript(ctx, vu, f.Page().TargetID())
+			if err != nil {
+				return nil, err
+			}
+
+			return k6ext.Promise(ctx, func() (result any, reason error) {
+				return f.WaitForResponse(val, popts, jsRegexChecker)
+			}), nil
+		},
 	}
 	maps["$"] = func(selector string) *sobek.Promise {
 		return k6ext.Promise(vu.Context(), func() (any, error) {
--- a/internal/js/modules/k6/browser/browser/page_mapping.go
+++ b/internal/js/modules/k6/browser/browser/page_mapping.go
@@ -597,6 +597,31 @@ func mapPage(vu moduleVU, p *common.Page) mapping { //nolint:gocognit,cyclop
 				return nil, p.WaitForURL(val, popts, jsRegexChecker)
 			}), nil
 		},
+		"waitForResponse": func(url sobek.Value, opts sobek.Value) (*sobek.Promise, error) {
+			popts := common.NewFrameWaitForResponseOptions(p.Timeout())
+			if err := popts.Parse(vu.Context(), opts); err != nil {
+				return nil, fmt.Errorf("parsing waitForResponse options: %w", err)
+			}
+
+			var val string
+			switch url.ExportType() {
+			case reflect.TypeOf(string("")):
+				val = fmt.Sprintf("'%s'", url.String()) // Strings require quotes
+			default: // JS Regex, CSS, numbers or booleans
+				val = url.String() // No quotes
+			}
+
+			// Inject JS regex checker for URL pattern matching
+			ctx := vu.Context()
+			jsRegexChecker, err := injectRegexMatcherScript(ctx, vu, p.TargetID())
+			if err != nil {
+				return nil, err
+			}
+
+			return k6ext.Promise(ctx, func() (result any, reason error) {
+				return p.WaitForResponse(val, popts, jsRegexChecker)
+			}), nil
+		},
 		"workers": func() *sobek.Object {
 			var mws []mapping
 			for _, w := range p.Workers() {
--- a/internal/js/modules/k6/browser/common/frame.go
+++ b/internal/js/modules/k6/browser/common/frame.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"reflect"
 	"strconv"
 	"strings"
 	"sync"
@@ -2162,6 +2163,89 @@ func (f *Frame) WaitForURL(urlPattern string, opts *FrameWaitForURLOptions, jsRe
 	}
 	_, err = f.WaitForNavigation(navOpts, jsRegexChecker)
 	return err
+}
+
+// WaitForResponse waits for a response that matches the given URL pattern or predicate function.
+// urlOrPredicate can be a string URL pattern, a regex pattern, or a predicate function.
+func (f *Frame) WaitForResponse(urlOrPredicate string, opts *FrameWaitForResponseOptions, jsRegexChecker JSRegexChecker) (*Response, error) {
+	f.log.Debugf("Frame:WaitForResponse", "fid:%s furl:%q pattern:%s", f.ID(), f.URL(), urlOrPredicate)
+	defer f.log.Debugf("Frame:WaitForResponse:return", "fid:%s furl:%q pattern:%s", f.ID(), f.URL(), urlOrPredicate)
+
+	timeoutCtx, timeoutCancel := context.WithTimeout(f.ctx, opts.Timeout)
+	defer timeoutCancel()
+
+	// Create URL matcher based on the pattern
+	matcher, err := urlMatcher(urlOrPredicate, jsRegexChecker)
+	if err != nil {
+		return nil, fmt.Errorf("parsing URL pattern: %w", err)
+	}
+
+	// Create a channel to receive matching responses
+	responseCh := make(chan *Response, 1)
+	var matcherErr error
+
+	// Create a handler function that will be registered with the page
+	handler := func(event PageOnEvent) error {
+		if event.Response != nil {
+			matched, err := matcher(event.Response.URL())
+			if err != nil {
+				matcherErr = err
+				select {
+				case responseCh <- nil:
+				default:
+				}
+				return nil
+			}
+			if matched {
+				select {
+				case responseCh <- event.Response:
+				default:
+				}
+			}
+		}
+		return nil
+	}
+
+	// Register the handler with the page
+	f.page.eventHandlersMu.Lock()
+	if f.page.eventHandlers[EventPageResponseCalled] == nil {
+		f.page.eventHandlers[EventPageResponseCalled] = make([]PageOnHandler, 0)
+	}
+	f.page.eventHandlers[EventPageResponseCalled] = append(f.page.eventHandlers[EventPageResponseCalled], handler)
+	f.page.eventHandlersMu.Unlock()
+
+	// Clean up handler when done
+	defer func() {
+		f.page.eventHandlersMu.Lock()
+		handlers := f.page.eventHandlers[EventPageResponseCalled]
+		for i, h := range handlers {
+			// Compare function pointers to find our handler
+			if reflect.ValueOf(h).Pointer() == reflect.ValueOf(handler).Pointer() {
+				// Remove the handler by replacing it with the last element and shrinking slice
+				handlers[i] = handlers[len(handlers)-1]
+				f.page.eventHandlers[EventPageResponseCalled] = handlers[:len(handlers)-1]
+				break
+			}
+		}
+		f.page.eventHandlersMu.Unlock()
+	}()
+
+	select {
+	case resp := <-responseCh:
+		if matcherErr != nil {
+			return nil, fmt.Errorf("URL pattern matching error: %w", matcherErr)
+		}
+		if resp == nil {
+			return nil, fmt.Errorf("received nil response due to matcher error")
+		}
+		return resp, nil
+	case <-timeoutCtx.Done():
+		e := &k6ext.UserFriendlyError{
+			Err:     timeoutCtx.Err(),
+			Timeout: opts.Timeout,
+		}
+		return nil, fmt.Errorf("waiting for response %q: %w", urlOrPredicate, e)
+	}
 }
 
 func (f *Frame) adoptBackendNodeID(world executionWorld, id cdp.BackendNodeID) (*ElementHandle, error) {
--- a/internal/js/modules/k6/browser/common/frame_options.go
+++ b/internal/js/modules/k6/browser/common/frame_options.go
@@ -802,3 +802,30 @@ func (o *FrameWaitForURLOptions) Parse(ctx context.Context, opts sobek.Value) er
 	}
 	return nil
 }
+
+// FrameWaitForResponseOptions are options for Frame.waitForResponse and Page.waitForResponse.
+type FrameWaitForResponseOptions struct {
+	Timeout time.Duration
+}
+
+// NewFrameWaitForResponseOptions returns a new FrameWaitForResponseOptions.
+func NewFrameWaitForResponseOptions(defaultTimeout time.Duration) *FrameWaitForResponseOptions {
+	return &FrameWaitForResponseOptions{
+		Timeout: defaultTimeout,
+	}
+}
+
+// Parse parses the frame waitForResponse options.
+func (o *FrameWaitForResponseOptions) Parse(ctx context.Context, opts sobek.Value) error {
+	rt := k6ext.Runtime(ctx)
+	if opts != nil && !sobek.IsUndefined(opts) && !sobek.IsNull(opts) {
+		opts := opts.ToObject(rt)
+		for _, k := range opts.Keys() {
+			switch k {
+			case "timeout":
+				o.Timeout = time.Duration(opts.Get(k).ToInteger()) * time.Millisecond
+			}
+		}
+	}
+	return nil
+}
--- a/internal/js/modules/k6/browser/common/page.go
+++ b/internal/js/modules/k6/browser/common/page.go
@@ -1728,6 +1728,22 @@ func (p *Page) WaitForURL(urlPattern string, opts *FrameWaitForURLOptions, jsReg
 	return nil
 }
 
+// WaitForResponse waits for a response that matches the given URL pattern.
+// jsRegexChecker should be non-nil to be able to test against a URL pattern.
+func (p *Page) WaitForResponse(urlPattern string, opts *FrameWaitForResponseOptions, jsRegexChecker JSRegexChecker) (*Response, error) {
+	p.logger.Debugf("Page:WaitForResponse", "sid:%v pattern:%s", p.sessionID(), urlPattern)
+	_, span := TraceAPICall(p.ctx, p.targetID.String(), "page.waitForResponse")
+	defer span.End()
+
+	resp, err := p.frameManager.MainFrame().WaitForResponse(urlPattern, opts, jsRegexChecker)
+	if err != nil {
+		spanRecordError(span, err)
+		return nil, err
+	}
+
+	return resp, nil
+}
+
 // Workers returns all WebWorkers of page.
 func (p *Page) Workers() []*Worker {
 	p.workersMu.Lock()
--- a/internal/js/modules/k6/browser/tests/page_test.go
+++ b/internal/js/modules/k6/browser/tests/page_test.go
@@ -2997,3 +2997,80 @@ func TestWaitForURL(t *testing.T) {
 	)
 	assert.Equal(t, sobek.Undefined(), got.Result())
 }
+
+func TestWaitForResponse(t *testing.T) {
+	if runtime.GOOS == "windows" {
+		t.Skip("Skipped due to https://github.com/grafana/k6/issues/4937")
+	}
+
+	t.Parallel()
+
+	tb := newTestBrowser(t, withFileServer())
+	tb.vu.ActivateVU()
+	tb.vu.StartIteration(t)
+
+	got := tb.vu.RunPromise(t, `
+		const page = await browser.newPage();
+		const testURL = '%s';
+
+		try {
+			await page.goto(testURL);
+
+			// Test exact URL match for response
+			let response = await Promise.all([
+				page.waitForResponse('%s'),
+				page.locator('#page1').click()
+			]);
+			if (!response[0].url().endsWith('page1.html')) {
+				throw new Error('Expected response for page1.html but got ' + response[0].url());
+			}
+
+			// Test regex pattern for response
+			await page.goto(testURL);
+			response = await Promise.all([
+				page.waitForResponse(/.*\.html$/),
+				page.locator('#page2').click()
+			]);
+			if (!response[0].url().endsWith('.html')) {
+				throw new Error('Expected response URL to end with .html but got ' + response[0].url());
+			}
+
+			// Test timeout when response doesn't match
+			await page.goto(testURL);
+			let timedOut = false;
+			try {
+				response = await Promise.all([
+					page.waitForResponse(/.*nonexistent\.html$/, { timeout: 500 }),
+					page.locator('#page1').click()  // This loads page1.html, not nonexistent.html
+				]);
+			} catch (error) {
+				if (error.toString().includes('waiting for response')) {
+					timedOut = true;
+				} else {
+					throw error;
+				}
+			}
+			if (!timedOut) {
+				throw new Error('Expected timeout error when response URL does not match');
+			}
+
+			// Test empty pattern (matches any response)
+			await page.goto(testURL);
+			response = await Promise.all([
+				page.waitForResponse(''),
+				page.locator('#page2').click()
+			]);
+			if (!response[0].url().includes('page2.html') && !response[0].url().includes('waitfornavigation_test.html')) {
+				throw new Error('Expected empty pattern to match any response but got ' + response[0].url());
+			}
+
+		} finally {
+			// Must call close() which will clean up the taskqueue.
+			await page.close();
+		}
+	`,
+		tb.staticURL("waitfornavigation_test.html"),
+		tb.staticURL("page1.html"),
+	)
+	assert.Equal(t, sobek.Undefined(), got.Result())
+}
--- a/internal/js/modules/k6/browser/browser/mapping_test.go
+++ b/internal/js/modules/k6/browser/browser/mapping_test.go
@@ -373,6 +373,7 @@ type pageAPI interface { //nolint:interfacebloat
 	WaitForSelector(selector string, opts sobek.Value) (*common.ElementHandle, error)
 	WaitForTimeout(timeout int64)
 	WaitForURL(url string, opts sobek.Value) (*sobek.Promise, error)
+	WaitForResponse(url string, opts sobek.Value) (*sobek.Promise, error)
 	Workers() []*common.Worker
 }
 
@@ -442,6 +443,7 @@ type frameAPI interface { //nolint:interfacebloat
 	WaitForSelector(selector string, opts sobek.Value) (*common.ElementHandle, error)
 	WaitForTimeout(timeout int64)
 	WaitForURL(url string, opts sobek.Value) (*sobek.Promise, error)
+	WaitForResponse(url string, opts sobek.Value) (*sobek.Promise, error)
 }
 
 // elementHandleAPI is the interface of an in-page DOM element.


HasithDeAlwis:
Work is needed to have behaviour exactly the same as Playwright, which is not too much. 
Here is the official playwright implementation: https://github.com/microsoft/playwright/blob/main/packages/playwright-core/src/client/page.ts#L437

internal/js/modules/k6/browser/browser/frame_mapping.go
```diff
@@ -448,6 +448,31 @@ func mapFrame(vu moduleVU, f *common.Frame) mapping {
 				return nil, f.WaitForURL(val, popts, jsRegexChecker)
 			}), nil
 		},
+		"waitForResponse": func(url sobek.Value, opts sobek.Value) (*sobek.Promise, error) {
+			popts := common.NewFrameWaitForResponseOptions(f.Timeout())
+			if err := popts.Parse(vu.Context(), opts); err != nil {
+				return nil, fmt.Errorf("parsing waitForResponse options: %w", err)
+			}
```
inancgumus: Could you refactor this option parsing and do it in the `browser` package? Here's an [example](https://github.com/grafana/k6/blob/2226dfe560ca9f383393dc1b02b86d5d540dac01/internal/js/modules/k6/browser/browser/route_mapping.go#L22-L72). Let me know if you have questions.

inancgumus (commented):
Thanks for your contribution 🙇 I haven't looked at the entire PR, but I noticed something with option parsing.

internal/js/modules/k6/browser/browser/frame_mapping.go
HasithDeAlwis: Done, and thank you for the feedback!

internal/js/modules/k6/browser/common/frame.go
```diff
@@ -2164,6 +2165,89 @@ func (f *Frame) WaitForURL(urlPattern string, opts *FrameWaitForURLOptions, jsRe
 	return err
 }
 
+// WaitForResponse waits for a response that matches the given URL pattern or predicate function.
+// urlOrPredicate can be a string URL pattern, a regex pattern, or a predicate function.
+func (f *Frame) WaitForResponse(urlOrPredicate string, opts *FrameWaitForResponseOptions, jsRegexChecker JSRegexChecker) (*Response, error) {
```
ankur22: What is the reason for implementing the core logic of `waitForResponse` in `frame`? I think we're only interested in `page.waitForResponse`. It also looks like Playwright only supports `page.waitForResponse`. Please let me know if i missed something. I think we should move this to `page`, if i'm correct in what I'm seeing in Playwright.

internal/js/modules/k6/browser/common/frame.go
```diff
@@ -2164,6 +2165,89 @@ func (f *Frame) WaitForURL(urlPattern string, opts *FrameWaitForURLOptions, jsRe
...
+				default:
+				}
+			}
+		}
+		return nil
+	}
+
+	// Register the handler with the page
+	f.page.eventHandlersMu.Lock()
+	if f.page.eventHandlers[EventPageResponseCalled] == nil {
```
ankur22: I don't think `page. eventHandlers` has been set up to allow for an easy way to remove handlers, and this is why you're having to work with reflect. I'd prefer if a new `responseEventHandler` (or something similar) was created specifically for `page.waitForResponse` where it would be easy to remove the handler by matching against the input url.

internal/js/modules/k6/browser/common/frame.go
```diff
@@ -2164,6 +2165,89 @@ func (f *Frame) WaitForURL(urlPattern string, opts *FrameWaitForURLOptions, jsRe
...
+		f.page.eventHandlersMu.Unlock()
+	}()
+
+	select {
+	case resp := <-responseCh:
+		if matcherErr != nil {
+			return nil, fmt.Errorf("URL pattern matching error: %w", matcherErr)
+		}
+		if resp == nil {
+			return nil, fmt.Errorf("received nil response due to matcher error")
```
ankur22: This doesn't seem correct since the `matcherErr` should be non-nil when `response` is `nil`. Are there any other cases where we could end up with a `nil` response and a `nil` `matcherErr`?

internal/js/modules/k6/browser/common/frame.go
```diff
@@ -2164,6 +2165,89 @@ func (f *Frame) WaitForURL(urlPattern string, opts *FrameWaitForURLOptions, jsRe
 	return err
 }
 
+// WaitForResponse waits for a response that matches the given URL pattern or predicate function.
+// urlOrPredicate can be a string URL pattern, a regex pattern, or a predicate function.
+func (f *Frame) WaitForResponse(urlOrPredicate string, opts *FrameWaitForResponseOptions, jsRegexChecker JSRegexChecker) (*Response, error) {
+	f.log.Debugf("Frame:WaitForResponse", "fid:%s furl:%q pattern:%s", f.ID(), f.URL(), urlOrPredicate)
```
ankur22: Nit: `urlOrPredicate` maybe should just be `urlOrRegex`?

internal/js/modules/k6/browser/common/frame_options.go
```diff
@@ -802,3 +802,30 @@ func (o *FrameWaitForURLOptions) Parse(ctx context.Context, opts sobek.Value) er
 	}
 	return nil
 }
+
+// FrameWaitForResponseOptions are options for Frame.waitForResponse and Page.waitForResponse.
```
ankur22: Maybe best to place in page_options.go if my assumption is correct around `page.waitForResponse`.

internal/js/modules/k6/browser/common/page.go
```diff
@@ -1728,6 +1728,22 @@ func (p *Page) WaitForURL(urlPattern string, opts *FrameWaitForURLOptions, jsReg
 	return nil
 }
 
+// WaitForResponse waits for a response that matches the given URL pattern.
+// jsRegexChecker should be non-nil to be able to test against a URL pattern.
+func (p *Page) WaitForResponse(urlPattern string, opts *FrameWaitForResponseOptions, jsRegexChecker JSRegexChecker) (*Response, error) {
+	p.logger.Debugf("Page:WaitForResponse", "sid:%v pattern:%s", p.sessionID(), urlPattern)
+	_, span := TraceAPICall(p.ctx, p.targetID.String(), "page.waitForResponse")
+	defer span.End()
```
ankur22, line 1819: Nice 😄 !

internal/js/modules/k6/browser/browser/frame_mapping.go
```diff
@@ -448,6 +448,31 @@ func mapFrame(vu moduleVU, f *common.Frame) mapping {
 				return nil, f.WaitForURL(val, popts, jsRegexChecker)
 			}), nil
 		},
+		"waitForResponse": func(url sobek.Value, opts sobek.Value) (*sobek.Promise, error) {
```
ankur22: We don't want to support `frame.waitForResponse` just `page.waitForResponse`, to match Playwright. At least for now, until there is a definite use case for `frame.waitForResponse`.

internal/js/modules/k6/browser/browser/route_mapping.go
```diff
@@ -113,3 +114,23 @@ func parseHeaders(headers *sobek.Object) []common.HTTPHeader {
...
+	}
+
+	rt := k6ext.Runtime(ctx)
+	obj := opts.ToObject(rt)
+	for _, k := range obj.Keys() {
+		switch k {
+		case "timeout":
+			ropts.Timeout = time.Duration(obj.Get(k).ToInteger()) * time.Millisecond
+		default:
+			return ropts, fmt.Errorf("unsupported waitForResponse option: '%s'", k)
```
ankur22, line 133: Nice! 👏

internal/js/modules/k6/browser/common/frame.go
```diff
@@ -2164,6 +2165,91 @@ func (f *Frame) WaitForURL(urlPattern string, opts *FrameWaitForURLOptions, jsRe
...
+				case responseCh <- event.Response:
+				default:
+				}
+			}
+		}
+		return nil
+	}
+
+	// Register the handler with the page
+	f.page.eventHandlersMu.Lock()
```
ankur22: I haven't tested this yet, but I don't feel like it's a good idea to reuse the `page.on` data structures for `page.waitForResponse`. I'm worried that `page.on` might work with the handler that is destined for `page.waitForResponse` 🤔 I need to have another think and test it properly.

ankur22 (changes requested):
@HasithDeAlwis thank you so much for working on this API which we would definitely like to have in k6 browser 🙇 

I think we might need to move a few things around and also consider a dedicated data structure just for `page.waitForResponse`. The core logic however looks good.

internal/js/modules/k6/browser/common/frame.go
HasithDeAlwis: I just rechecked some playwright docs, I completly misremebered the implementation, your 100% right, let me fix this.

HasithDeAlwis: Resolved :+1:

internal/js/modules/k6/browser/common/frame.go
HasithDeAlwis: This is a great idea, let me try to work on it and I'll let you know if I have any questions

HasithDeAlwis:
@ankur22 thanks for the feedback 🙇 🚀 ! Let me have a think about how to implement the new data structure cleanly, but I'm sure that is the best way moving forward. I'll ping you again here once I have a clearer vision, and I'll consult you before making any changes.

HasithDeAlwis:
@ankur22, after thinking about this for a while, I put my architecture plan into copilot to synthesize all my ideas together, and it came up with this nice summary. Let me know your thoughts :) 
# WaitForResponse Fix: Implementation Plan

## Problem Summary
Current `WaitForResponse` incorrectly reuses `page.eventHandlers` infrastructure designed for `page.on()`:

```go
// Current problematic approach in WaitForResponse
func (p *Page) WaitForResponse(...) {
    // Manually adds handler to page.eventHandlers slice
    p.eventHandlers[EventPageResponseCalled] = append(p.eventHandlers[EventPageResponseCalled], handler)
    
    // Complex reflection-based cleanup to find and remove handler
    defer func() {
        for i, h := range handlers {
            if reflect.ValueOf(h).Pointer() == reflect.ValueOf(handler).Pointer() {
                // Remove handler by swapping with last element
                handlers[i] = handlers[len(handlers)-1]
                p.eventHandlers[EventPageResponseCalled] = handlers[:len(handlers)-1]
                break
            }
        }
    }()
}
```

**Key Issues:**
1. **Wrong abstraction**: `page.eventHandlers` designed for persistent `page.on()` handlers, not temporary waiters
2. **Difficult cleanup**: No easy way to remove handlers, requires complex reflection-based unsubscription  
3. **Interference risk**: `page.on('response')` handlers might interfere with `waitForResponse` handlers
4. **Architectural mixing**: Persistent user handlers and temporary internal waiters share the same system

## Solution: Dedicated ResponseEventHandler

Create a separate event handler specifically for `waitForResponse` that allows easy handler removal by URL pattern matching.

### Core Architecture

```go
type ResponseEventHandler struct {
    mu            sync.RWMutex
    activeWaiters map[string]*responseWaiter
    nextWaiterID  int64  // Atomic counter for unique waiter IDs
}

type responseWaiter struct {
    id           string
    matcher      URLMatcher     // Compiled matcher for efficient URL checking
    responseChan chan *Response
    ctx          context.Context
    cancel       context.CancelFunc
    createdAt    time.Time      // For debugging/monitoring
}
```

### Key Benefits
- **Purpose-built**: Designed specifically for temporary response waiting, not persistent handlers
- **Thread-safe**: All operations protected by sync.RWMutex with proper atomic ID generation
- **Complete isolation**: Zero interference with `page.on('response')` handlers  
- **Clean resource management**: Proper context cancellation and automatic cleanup
- **No reflection**: Uses unique IDs for waiter management instead of function pointer comparison

### Implementation Changes

**1 File Modified:**

1. **Updated: `page.go`**
   - Added `ResponseEventHandler` struct with thread-safe waiter management
   - Added `responseWaiter` struct with proper context and cancellation support
   - Added `responseEventHandler` field to Page struct  
   - Replaced WaitForResponse method to use dedicated handler (eliminated reflection cleanup)
   - Wired response events directly to new handler via `onResponse()` method
   - Added cleanup to `Page.Close()` to prevent resource leaks
   - **Reused existing URLMatcher**: Leveraged existing `urlMatcher()` function from frame.go

### ResponseEventHandler Implementation Details

**Core Methods:**
- `addWaiter(waiter *responseWaiter)` - Thread-safe waiter registration
- `removeWaiter(id string)` - Clean waiter removal with context cancellation  
- `processResponse(response *Response)` - Efficient response matching and delivery
- `waitForMatch(ctx context.Context, matcher URLMatcher)` - Main waiting logic with timeout handling
- `generateWaiterID()` - Atomic ID generation following k6 patterns
- `Cleanup()` - Resource cleanup for page shutdown

**Thread Safety Design:**
```go
func (reh *ResponseEventHandler) processResponse(response *Response) {
    reh.mu.RLock()
    waitersToNotify := make([]*responseWaiter, 0)
    
    // Find matching waiters under read lock
    for _, waiter := range reh.activeWaiters {
        matched, err := waiter.matcher(response.URL())
        if err == nil && matched {
            waitersToNotify = append(waitersToNotify, waiter)
        }
    }
    reh.mu.RUnlock()
    
    // Notify outside of lock to prevent deadlocks
    for _, waiter := range waitersToNotify {
        select {
        case waiter.responseChan <- response:
        default: // Non-blocking to prevent goroutine leaks
        }
    }
}
```

**Integration Points:**
- Initialized in `NewPage()` constructor
- Called from existing `onResponse()` method
- Uses existing `URLMatcher` type and `urlMatcher()` function

**Page.WaitForResponse Changes:**
```go
// Before: Complex reflection-based cleanup
func (p *Page) WaitForResponse(...) (*Response, error) {
    // 60+ lines of handler registration, reflection cleanup
    defer func() {
        for i, h := range handlers {
            if reflect.ValueOf(h).Pointer() == reflect.ValueOf(handler).Pointer() {
                handlers[i] = handlers[len(handlers)-1]
                p.eventHandlers[EventPageResponseCalled] = handlers[:len(handlers)-1]
                break
            }
        }
    }()
    // ... complex channel and timeout logic
}

// After: Clean delegation to ResponseEventHandler  
func (p *Page) WaitForResponse(...) (*Response, error) {
    timeoutCtx, timeoutCancel := context.WithTimeout(p.ctx, opts.Timeout)
    defer timeoutCancel()

    // Create URL matcher based on the pattern
    matcher, err := urlMatcher(urlOrRegex, jsRegexChecker)
    if err != nil {
        return nil, fmt.Errorf("parsing URL pattern: %w", err)
    }

    // Use the dedicated ResponseEventHandler for clean separation
    return p.responseEventHandler.waitForMatch(timeoutCtx, matcher)
}
```

**Page.Close() Integration:**
```go
func (p *Page) Close() error {
    // ... existing cleanup code ...
    
    // Clean up response event handler to prevent resource leaks
    p.responseEventHandler.Cleanup()
    
    // ... rest of close logic ...
}
```

### What's NOT changing
- Public API stays the same (`page.waitForResponse()` signature unchanged)
- Uses existing URL pattern matching (`urlMatcher` function from frame.go)  
- `page.on('response')` behavior completely unchanged
- Performance characteristics improved (no reflection, better concurrency)

internal/js/modules/k6/browser/common/frame.go
HasithDeAlwis: Agreed, this will be fixed with the new data structure, see the comment I left below

ankur22:
@HasithDeAlwis thanks for working on this API 🙂

I'm definitely the sort of person who prefers to see it in action rather than a plan from an AI assistant.

I think it is best that you take on board the feedback, come up with a new solution (with or without an AI assistant), implement it, and finally test it thoroughly making sure that any or all edge cases that you can think of are also implemented.

I will add that I would prefer a solution whereby there was no cleanup required in `page.close` (if that is at all possible).

HasithDeAlwis:
@ankur22 thanks again for the feedback! Feel free to take a look at this now, I added some more tests to check for edge cases and can add more if you feel it's necessary. 
I also got rid of any logic in `page.close`, I added a more elegant cleanup on the `waitForMatch` function that uses the `page` context as it's parent so on a `page.close()` there is already logic in place to clean up the new event listener!

internal/js/modules/k6/browser/tests/page_test.go
```diff
@@ -2997,3 +2997,131 @@ func TestWaitForURL(t *testing.T) {
...
+			]);
+			if (!response[0].url().includes('page2.html') && !response[0].url().includes('waitfornavigation_test.html')) {
+				throw new Error('Expected empty pattern to match any response but got ' + response[0].url());
+			}
+
+		} finally {
+			// Must call close() which will clean up the taskqueue.
+			await page.close();
+		}
+	`,
```
inancgumus: Is there any reason why we use JavaScript code instead of directly using the browser module's Go API here, and also in the other test ([example](https://github.com/grafana/k6/blob/d1bf378f04d5ddc7b7ebe66b90b7d7f20fdc30dd/internal/js/modules/k6/browser/tests/page_test.go#L1824-L1878))? JavaScript tests in the Go code make it harder to read, debug, and refactor later on.

HasithDeAlwis: Hey! I was just following what I thought was the convention since `waitForURL`, `waitForNavigation`, and `onResponse` were tested using JavaScript. Happy to change it though!

inancgumus: The convention is to use Go code in tests unless there's a need for a JavaScript test, such as using the mapped API. Thanks!

HasithDeAlwis: Perfect, I'll make those changes then. Should I remove these tests then? I think they have value by directly testing the mapped API, but either way, all good with any changes

inancgumus: We're not removing them, but converting to Go code. We automatically test mappings [here](https://github.com/grafana/k6/blob/master/internal/js/modules/k6/browser/browser/mapping_test.go).

internal/js/modules/k6/browser/common/page.go
```diff
@@ -210,6 +211,103 @@ func NewRouteHandler(
...
+	nextWaiterID  int64
+}
+
+type responseWaiter struct {
+	id           string
+	matcher      URLMatcher
+	responseChan chan *Response
+	ctx          context.Context
+	cancel       context.CancelFunc
+	createdAt    time.Time
```
ankur22: Why do we need `createdAt`?

internal/js/modules/k6/browser/common/page.go
```diff
@@ -210,6 +211,103 @@ func NewRouteHandler(
...
+		select {
+		case waiter.responseChan <- response:
+		case <-waiter.ctx.Done():
+		}
+	}
+}
+
+func (reh *ResponseEventHandler) generateWaiterID() string {
+	id := atomic.AddInt64(&reh.nextWaiterID, 1)
+	return fmt.Sprintf("waiter_%d", id)
```
ankur22: We're trying to work with concatenation when it comes to string formatting since it's generally more performant: `"waiter_" + strconv.Itoa(id)`

internal/js/modules/k6/browser/common/page.go
```diff
@@ -210,6 +211,103 @@ func NewRouteHandler(
...
+}
+
+// ResponseEventHandler handles the internal mechanics
```
ankur22: This comment doesn't seem to reference the correct method.

internal/js/modules/k6/browser/common/page.go
```diff
@@ -210,6 +211,103 @@ func NewRouteHandler(
...
+		matched, err := waiter.matcher(response.URL())
+		if err == nil && matched {
+			waitersToNotify = append(waitersToNotify, waiter)
+		}
+	}
+	reh.mu.RUnlock()
+
+	for _, waiter := range waitersToNotify {
+		select {
+		case waiter.responseChan <- response:
```
ankur22: This doesn't feel safe. In `waitForMatch` in the defer it closes the `responseChan`. To me there's a chance that the:

1. context could close after a `waiter` is added to `waitersToNotify` which would force the defer in `waitForMatch` to be called;
2. Here there's a chance that either case can be hit, even when the context is done. If it tries to write to `responseChan` when it's closed, it will cause a `panic`.

It's also considered safer for the sender to close the channel, so i think it should be ok to close the channel straight after sending something on it.

internal/js/modules/k6/browser/common/page.go
```diff
@@ -210,6 +211,103 @@ func NewRouteHandler(
...
+		responseChan: make(chan *Response, 1),
+		ctx:          waiterContext,
+		cancel:       waiterCancel,
+	}
+
+	reh.addWaiter(waiter)
+	defer func() {
+		waiterCancel()
+		reh.removeWaiter(waiterID)
+		close(waiter.responseChan)
```
ankur22: It's best if we don't close the `responseChan` channel here. It should be the sender that closes to avoid a situation where a sender tries to send on a closed channel.

internal/js/modules/k6/browser/common/page.go
```diff
@@ -1826,6 +1830,41 @@ func (p *Page) WaitForURL(urlPattern string, opts *FrameWaitForURLOptions, jsReg
...
+
+	// Use the dedicated ResponseEventHandler for clean separation
+	resp, err := p.responseEventHandler.waitForMatch(timeoutCtx, matcher)
+	if err != nil {
+		if errors.Is(err, context.DeadlineExceeded) {
+			e := &k6ext.UserFriendlyError{
+				Err:     err,
+				Timeout: opts.Timeout,
+			}
+			err = fmt.Errorf("timed out waiting for response - %q: %w", urlOrRegex, e)
```
ankur22: I'm not 100% sure about this. The context could close because the:

1. the iteration is ending;
2. the test is ending;
3. or when the timeout is hit.

internal/js/modules/k6/browser/tests/page_test.go
ankur22: It looks like you could benefit from working with the mapping layer. You could also consider writing your tests like `TestBrowserContextMappingBehavior`.

ankur22 (changes requested):
Nice work! I like the refactors and the new flow with its own data structure looks good 🙌 

Left some comments that I think need to be considered to ensure we don't cause panics.

Also, could you extend an existing test `examples/browser/route.js` to also work with `page.waitForResponse`? This will help make sure we're not causing any deadlocks when working with both `page.on('response')` and `page.waitForResponse`.

internal/js/modules/k6/browser/common/page.go
```diff
@@ -210,6 +211,103 @@ func NewRouteHandler(
...
+	return fmt.Sprintf("waiter_%d", id)
+}
+
+// ResponseEventHandler handles the internal mechanics
+func (reh *ResponseEventHandler) waitForMatch(ctx context.Context, matcher URLMatcher) (*Response, error) {
+	waiterID := reh.generateWaiterID()
+	waiterContext, waiterCancel := context.WithCancel(ctx)
+
+	waiter := &responseWaiter{
+		id:           waiterID,
```
ankur22: I don't think we need to store the `id` in the `responseWaiter`. I think it can be generated in `addWaiter`, used there, and returned as a return value from `addWaiter`. We can then reference that `id` in the `defer` function below.

internal/js/modules/k6/browser/common/page.go
ankur22: Actually, why do we need to create a `string` id like this at all, why not just return an `int64`?

internal/js/modules/k6/browser/common/page.go
HasithDeAlwis: Removed, it dead code from a previous implementation :+1:

internal/js/modules/k6/browser/tests/page_test.go
HasithDeAlwis: Sounds good, I made some changes to convert the tests to Go, I'll see what the `TestBrowserContextMappingBehavior` is about!

Edit: Just checked `TestBrowserContextMappingBehavior` and I believe I followed the same format, let me know what you think and if it's up to standard

internal/js/modules/k6/browser/common/page.go
HasithDeAlwis: Done thanks!

internal/js/modules/k6/browser/common/page.go
HasithDeAlwis: cleaned, thanks

internal/js/modules/k6/browser/common/page.go
HasithDeAlwis: Resolved!

internal/js/modules/k6/browser/common/page.go
HasithDeAlwis: Totally right! With my new changes, the current implementation doesn't close the channel at all, I removed the `close(waiter.responseChan)` call from the defer to follow the Go idiom that senders should close channels. Since the channel is buffered and never closed, it's completely safe for `processResponse` to send even after the context is cancelled, and since the channel is never closed a `panic` won't occur. If I'm correct, the worst case is one orphaned response enters the buffer that gets garbage collected with the channel.

ankur22: Aha! I hadn't spotted that `responseChan` is buffered. I think that generally regarded as a smell and usually hiding some underlying bad synchronisation that is hiding a deadlock. Could you please try to work with an unbuffered channel?

internal/js/modules/k6/browser/common/page.go
```diff
@@ -227,17 +226,20 @@ type responseWaiter struct {
...
 	}
 }
 
-func (reh *ResponseEventHandler) addWaiter(waiter *responseWaiter) {
+func (reh *ResponseEventHandler) addWaiter(waiter *responseWaiter) int64 {
 	reh.mu.Lock()
 	defer reh.mu.Unlock()
-	reh.activeWaiters[waiter.id] = waiter
+
+	id := reh.generateWaiterID()
```
ankur22: Can we move `id := reh.generateWaiterID()` to before `reh.mu.Lock()`? Avoids a situation where this is holding onto two locks.

examples/browser/route.js
```diff
@@ -29,16 +30,40 @@ export default async function () {
     });
 
     page.on('request', function (request) {
-      console.log('on request', 'url', request.url());
+      console.log('Request:', request.url());
```
ankur22: Can you please revert this change. If you used an AI can you tell it not to refactor existing parts of the script.

examples/browser/route.js
```diff
@@ -29,16 +30,40 @@ export default async function () {
...
     });
 
     page.on('response', function (response) {
-      console.log('on response', 'url', response.url());
+      console.log('Response:', response.url(), response.status());
```
ankur22: Can you please revert this change.

examples/browser/route.js
```diff
@@ -29,16 +30,40 @@ export default async function () {
...
+
+    await page.getByRole('button', { name: /pizza/i }).click();
+
+    const pizzaResponse = await pizzaResponsePromise;
+
+    // Check that the pizza API call was successful
+    check(pizzaResponse, {
+      'pizza API status is 200': (r) => r.status() === 200,
+      'pizza API URL is correct': (r) => r.url() === 'https://quickpizza.grafana.com/api/pizza',
+    });
```
ankur22, line 66: Nice! Thanks for this!

internal/js/modules/k6/browser/tests/page_test.go
ankur22: I have to admit i find the new tests very difficult to follow. I would prefer if you reverted it back to what it was originally, and did something like:

```go
got := tb.vu.RunPromise(t, `...`) // setup the test
require.Equals(...);

got := tb.vu.RunPromise(t, `...`) // Test exact URL match for response
require.Equals(...);

...
```

If you need you can use `t.Run("name", func(t *testing.T) {})` when needed to break up the tests for the same API.

internal/js/modules/k6/browser/common/page.go
```diff
@@ -543,6 +639,8 @@ func (p *Page) onRequest(request *Request) {
 func (p *Page) onResponse(resp *Response) {
 	p.logger.Debugf("Page:onResponse", "sid:%v url:%v", p.sessionID(), resp.URL())
 
+	p.responseEventHandler.processResponse(resp)
```
ankur22: I think it should be safe to put this on its own goroutine here since `page.waitForResponse` is a readonly operation. We also don't want to block the processing of responses while we loop through and execute each of the `page.waitForResponse` handlers.

internal/js/modules/k6/browser/common/page.go
HasithDeAlwis: Thanks again for the feedback, really appreciate all of it! 
~~I've been thinking about this for a while and I was curious, do you think using the [createWaitForEventPredicateHandler](https://github.com/grafana/k6/blob/master/internal/js/modules/k6/browser/common/helpers.go#L173) would be right approach to this problem? I initially didn't use it, because I saw from discussions that you were planning on deprecating [`WaitForNavigation`](https://github.com/grafana/k6/blob/master/internal/js/modules/k6/browser/common/frame.go#L1989) (which uses this) and were having problems with some race conditions, so I thought using a different implementation would be better, but I might have misjudged the source of the problem there. Would love your thoughts.~~

Sorry, scratch what I said here, I realize if I did this implementation I'd be reusing the `page.on` to make this work

internal/js/modules/k6/browser/tests/page_test.go
inancgumus: I don't think reverting them is [a better solution](https://github.com/grafana/k6/pull/5002#discussion_r2252062416). I agree that they need a bit of refactoring to be more readable, though. You might use [these helpers](https://github.com/grafana/k6/blob/0c5783cbb425470436bba15da793d2cc7c216955/internal/js/modules/k6/browser/tests/test_browser.go#L302-L350).

internal/js/modules/k6/browser/common/page.go
HasithDeAlwis: Just had another think about this, we don't need the buffer at all, it's unecessary. Fixed, let me know if this is better now.

internal/js/modules/k6/browser/tests/page_test.go
HasithDeAlwis: Let me know your thoughts on what I just pushed, I hope its a winner 😅, I followed what we had in [TestPageEvaluate](https://github.com/grafana/k6/blob/0c5783cbb425470436bba15da793d2cc7c216955/internal/js/modules/k6/browser/tests/page_test.go#L134) and [TestWaitForFrameNavigationWithinDocument](https://github.com/grafana/k6/blob/0c5783cbb425470436bba15da793d2cc7c216955/internal/js/modules/k6/browser/tests/frame_manager_test.go#L58). I made the `mockRegexChecker` similar to how it was done in [TestFrameManagerRequestStartedWithRoutes](https://github.com/grafana/k6/blob/0c5783cbb425470436bba15da793d2cc7c216955/internal/js/modules/k6/browser/tests/frame_manager_test.go#L58).


 Let me know if this is readable, I saw in other areas we had large `testCases` structures, which I found not work too well for my usecases, but I do have them in a separate commit, so if you'd like to see how that looks, please let me know ❤️

internal/js/modules/k6/browser/common/page.go
```diff
@@ -251,6 +251,7 @@ func (reh *ResponseEventHandler) processResponse(response *Response) {
 	}
 
 	reh.mu.RLock()
+	defer reh.mu.RUnlock()
```
ankur22, line 239: What was the reason to refactor `reh.mu.RUnlock()`?

HasithDeAlwis: Just to be more defensive. In the case that `Response A` and `Response B` came in one after another and they both match `waiter C`. I wanted to eliminate the possibility of building `waitersToNotify` with `waiter C` when `Response B` get's processed. Just removing any small race windows.

ankur22: Right, yeah, that make sense. I guess the `waiter.ctx.Done()` case would handle a situation that you're protecting against, but I think you are right to make the change. Let's go with it 👍 

Thanks for explaining the change!

internal/js/modules/k6/browser/tests/page_test.go
ankur22: Can you double check your test changes, please, looks like there's a redundant test function? See if you can make things a little simpler too. We're close to the end now IMO 😄 

The implementation side of the feature i'm happy with 🙇

HasithDeAlwis: Thanks! Can you let me know which parts could be made simple? I've been looking at these tests for some time now, so it's hard for me to judge. Fixed the redundant tests.

internal/js/modules/k6/browser/common/page.go
```diff
@@ -210,6 +211,99 @@ func NewRouteHandler(
...
+	if response == nil {
+		return
+	}
+
+	reh.mu.RLock()
+	defer reh.mu.RUnlock()
+	waitersToNotify := make([]*responseWaiter, 0)
+
+	for _, waiter := range reh.activeWaiters {
+		// Check if the context is still active before processing
```
inancgumus: The code is clear. Can you remove the comment?

internal/js/modules/k6/browser/common/page.go
```diff
@@ -210,6 +211,99 @@ func NewRouteHandler(
...
+
+	reh.activeWaiters[id] = waiter
+	return id
+}
+
+func (reh *ResponseEventHandler) removeWaiter(id int64) {
+	reh.mu.Lock()
+	defer reh.mu.Unlock()
+	delete(reh.activeWaiters, id)
+}
```
inancgumus: This code can directly go into `waitForMatch`, as that's the only method that uses these methods.

internal/js/modules/k6/browser/browser/page_mapping.go
```diff
@@ -597,6 +597,31 @@ func mapPage(vu moduleVU, p *common.Page) mapping { //nolint:gocognit,cyclop
...
+		"waitForResponse": func(url sobek.Value, opts sobek.Value) (*sobek.Promise, error) {
+			popts, err := parseWaitForResponseOptions(vu.Context(), opts, p.Timeout())
+			if err != nil {
+				return nil, fmt.Errorf("parsing waitForResponse options: %w", err)
+			}
+
+			var val string
+			switch url.ExportType() {
+			case reflect.TypeOf(string("")):
+				val = fmt.Sprintf("'%s'", url.String()) // Strings require quotes
```
inancgumus: Can you use the following to reduce allocations?

```suggestion
				val = "'"+ url.String() +"'" // Strings require quotes
```

internal/js/modules/k6/browser/common/page.go
```diff
@@ -210,6 +211,99 @@ func NewRouteHandler(
...
+		case <-waiter.ctx.Done():
+			continue
+		default:
+		}
+
+		matched, err := waiter.matcher(response.URL())
+		if err == nil && matched {
+			waitersToNotify = append(waitersToNotify, waiter)
+		}
+	}
```
inancgumus: Why do we need to put these waiters in a slice? Can you loop once without the slice?

internal/js/modules/k6/browser/common/page.go
```diff
@@ -210,6 +211,99 @@ func NewRouteHandler(
...
+		select {
+		case waiter.responseChan <- response:
+		case <-waiter.ctx.Done():
+		}
+	}
+}
+
+func (reh *ResponseEventHandler) generateWaiterID() int64 {
+	return atomic.AddInt64(&reh.nextWaiterID, 1)
+}
```
inancgumus: This code can also be inlined. There is only one call site.

inancgumus (commented):
Thank you for all the changes you've made 🙇 I have a few final suggestions to share.

internal/js/modules/k6/browser/common/page.go
HasithDeAlwis: Ah, thanks for this catch, we don't need it anymore, before there was a chance that two responses could be processed at the same time, but now we're sequentially processing responses, so this is no longer needed. 
Fixed :+1:

internal/js/modules/k6/browser/browser/page_mapping.go
HasithDeAlwis: Sounds good, I added the same change in `waitForURL` since I just copied the pattern from there.

internal/js/modules/k6/browser/tests/page_test.go
ankur22: I took another look with fresh eyes, lGTM 🚀 🙇

ankur22 (dismissed):
Great work! Thanks for taking on all the suggestions. LGTM 🚀

HasithDeAlwis:
Thanks @ankur22 @inancgumus for the thorough reviews and patience with all the iterations! Learned a lot about Go best practices.

inancgumus (approved):

HasithDeAlwis:
Would you like me to do the PRs for TypeScript and documentation updates before merging?

ankur22:
> Would you like me to do the PRs for TypeScript and documentation updates before merging?

We will wait to merge until later next week -- we're in the middle of a release.

Sure, if you want, happy for you to take a stab at the type defs. Can you make the changes in https://github.com/grafana/k6-DefinitelyTyped? When creating the PR, make sure to set the base branch to master on the k6-DefinitelyTyped repo fork. We will merge it into the appropriate branch once the current release is done.

ankur22:
@HasithDeAlwis hope it's ok if i take over and get this merged in 👍 

Currently battling conflicts 😅

PR #5003, opened 2025-08-01
fix logic for getting the build service URL

### LINKED ISSUE #4995: Build Service path should be changed only for the default build service URL
After the changes introduced in [support community extensions #4922](https://github.com/grafana/k6/pull/4922/files#diff-30ce7dbdc9b4a925daffaf36676c91b1771a48555dd4f1db879d48cf62b9b4b8R255-R271) k6 adds a path component to the URL for the builds service (`cloud` or `oss` depending on which catalog should be used)

As a result, the requests are sent to `K6_BUILD_SERVICE_URL/cloud` by default or to `K6_BUILD_SERVICE_URL/oss` if the `K6_ENABLE_COMMUNITY_EXTENSION` flag is enabled.

However, this change should apply ONLY if the URL for the build service is the default build service. That is, the one run by GCK6. 

If this URL is overridden with the `K6_BUILD_SERVICE_URL`, the path should not be changed in any case.

pablochacin:
## What?

Modifies the logic for using the community extensions catalog to prevent changing the URL for custom build service deployments.

## Why?

The changes introduced in [support community extensions #4922](https://github.com/grafana/k6/pull/4922) added the `/cloud` path to the build service URL by default and added `/oss` when community extensions are enabled.

This is correct for the default build services, but not for custom build services that don't implement either of these paths.

## Related PR(s)/Issue(s)

Closes #4995

### DIFF
--- a/cmd/state/state.go
+++ b/cmd/state/state.go
@@ -2,6 +2,7 @@ package state
 
 import (
 	"context"
+	"fmt"
 	"io"
 	"os"
 	"os/signal"
@@ -27,8 +28,11 @@ const (
 	// BinaryProvisioningFeatureFlag defines the environment variable that enables the binary provisioning
 	BinaryProvisioningFeatureFlag = "K6_BINARY_PROVISIONING"
 
-	// DefaultBuildServiceURL defines the URL to the default (grafana hosted) build service
-	DefaultBuildServiceURL = "https://ingest.k6.io/builder/api/v1"
+	// communityExtensionsCatalog defines the catalog for community extensions
+	communityExtensionsCatalog = "oss"
+
+	// defaultBuildServiceURL defines the URL to the default (grafana hosted) build service
+	defaultBuildServiceURL = "https://ingest.k6.io/builder/api/v1"
 
 	defaultConfigFileName = "config.json"
 	defaultBinaryCacheDir = "builds"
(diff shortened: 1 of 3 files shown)


codebien (changes requested):
I have an idea to centralize the logic currently split between cmd/state and internal/cmd. We could move the getServerBuild function directly into cmd/state. This would ensure that the value we're using is always the final, correct version. What do you think about making this change?

Unfortunately, I'm realizing it only now.

pablochacin:
@codebien If you mean moving `getBuildServiceURL` logic to `state`, I think that makes sense and I can do it in the PR. Or are you thinking about something more general? If so, I prefer to postpone to a follow-up PR and limit this PR to fix this particular issue. 

Regarding moving the logic to `state`, the only thing I see as a "drawback" is that we lose the log messages, but that's probably not critical. What do you think?

codebien:
>If you mean moving getBuildServiceURL logic to state, I think that makes sense and I can do it in the PR

@pablochacin Yes, this is what I meant. :+1:

pablochacin:
@codebien I removed the tests because the logic now is straightforward (just an if statement) and the state package does not have any tests. Do you consider this requires developing tests for this case?

codebien:
We can address them in a follow-up pull request

codebien (approved):

PR #5014, opened 2025-08-06
test for binary provisioning disabled

pablochacin:
## What?

Adds a smoke test for running k6 with the automatic extension resolution (formerly known as binary provisioning) disabled.

## Why?

This feature is now enabled by default, so all tests in the k6 test suite will go through the binary provisioning path. It is important to have a test that goes through the other path (binary provisioning disabled)

## Checklist

### DIFF
--- a/internal/cmd/tests/cmd_run_test.go
+++ b/internal/cmd/tests/cmd_run_test.go
@@ -2435,7 +2435,7 @@ func TestTypeScriptSupport(t *testing.T) {
 	assert.Contains(t, stderr, `something 42`)
 }
 
-func TestTypeScriptSupportWithoutBinaryProvisioning(t *testing.T) {
+func TestTypeScriptSupportWithoutAutomaticExtensionResolution(t *testing.T) {
 	t.Parallel()
 	depScript := `
 		export default function(): number {
--- a/internal/cmd/tests/cmd_run_test.go
+++ b/internal/cmd/tests/cmd_run_test.go
@@ -2423,6 +2423,35 @@ func TestTypeScriptSupport(t *testing.T) {
 	`
 
 	ts := NewGlobalTestState(t)
+	require.NoError(t, fsext.WriteFile(ts.FS, filepath.Join(ts.Cwd, "test.ts"), []byte(mainScript), 0o644))
+	require.NoError(t, fsext.WriteFile(ts.FS, filepath.Join(ts.Cwd, "bar.ts"), []byte(depScript), 0o644))
+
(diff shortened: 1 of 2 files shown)


internal/cmd/tests/cmd_run_test.go
```diff
@@ -2406,7 +2406,7 @@ func TestSetupTimeout(t *testing.T) {
 	assert.Contains(t, stderr, "setup() execution timed out after 1 seconds")
 }
 
-func TestTypeScriptSupportWithoutBinaryProvisioning(t *testing.T) {
+func TestTypeScriptSupport(t *testing.T) {
```
mstoykov: The point of the test is to have us test that we do not just break everything if binary provisioning isn't enabled. 

Is there a reason you want them removed?

pablochacin: I find it an arbitrary choice to pick only this test. 

I'm wondering if we should better relay on the integration tests (as we did with the Binary Provisioning when it was disable by default) 

But don't have a strong opinion agains introducing smoke test(s) for binary provisioning disable. Including re-introducing this one.

mstoykov: I would prefer for us to run all/most tests with both, but I do not see how to do this without us blocking this PR for ... undefinite amount of time. 

 So this at least give us some coverage, on a test that is more likely than a lot of the others to hit this problems

pablochacin: I opened this issue to follow up this https://github.com/grafana/k6/issues/5018

pablochacin:
@mstoykov yes, https://github.com/grafana/k6/pull/5019 (which was on top of this PR) was merged before an enabled the feature by default. Now this only adds a test for when the feature is not enabled.

mstoykov (approved):

internal/cmd/tests/cmd_run_test.go
```diff
@@ -2422,6 +2422,35 @@ func TestTypeScriptSupport(t *testing.T) {
...
+	ts.CmdArgs = []string{"k6", "run", "--quiet", "test.ts"}
+
+	cmd.ExecuteWithGlobalState(ts.GlobalState)
+
+	stderr := ts.Stderr.String()
+	t.Log(stderr)
+	assert.Contains(t, stderr, `something 42`)
+}
+
+func TestTypeScriptSupportWithoutBinaryProvisioning(t *testing.T) {
```
codebien:
```suggestion
func TestTypeScriptSupportWithoutAutomaticExtensionResolution(t *testing.T) {
```

mstoykov (approved):

PR #5015, opened 2025-08-07
Fix issue with `taskqueue` and `waitForNavigation`

ankur22:
## What?

We're only creating a `taskqueue` if the `url` option is used with a valid non empty url.

## Why?

This is a temporary fix for `waitForNavigation`. When the `taskqueue` is created. It seems that we need to perform a manual cleanup with `page.close` when we work with the `taskqueue` even with this [change](https://github.com/grafana/k6/commit/8fce16a75360a2317dfe2ff4c8ac4e5a2c292b5b). Since `page.waitForNavigation` is a common API that is used in a lot of scripts, the first step is to mitigate the deadlock for the scripts which don't already work with `page.close`. We will have to document the usage of `page.close` when working with any browser APIs that work with the `taskqueue`.

## Checklist


internal/js/modules/k6/browser/browser/page_mapping.go
```diff
@@ -537,11 +537,19 @@ func mapPage(vu moduleVU, p *common.Page) mapping { //nolint:gocognit,cyclop
...
-				return nil, err
+
+			// Avoid working with the taskqueue unless the URL option is used.
+			// At the moment the taskqueue needs to be cleaned up manually with
+			// page.close.
+			var jsRegexChecker common.JSRegexChecker
+			if popts.URL != "" {
+				// Inject JS regex checker for URL regex pattern matching
+				var err error
+				jsRegexChecker, err = injectRegexMatcherScript(ctx, vu, p.TargetID())
```
mstoykov: Looking at the code for this function, it probably shouldn't use a taskqueue.

The taskqueue is meant for the cases where you will need to repeatedly execute stuff on  the event loop - such as on `page.on("request", ...`. In this case there is a singel call, so it seems to me like it will be better to just change it to use RegisterCallback from https://pkg.go.dev/go.k6.io/k6/js/modules#VU

ankur22: Sounds promising!

I'm a bit confused though. The matcher might be called multiple times, it depends on the URL and the requests that are processed. e.g.:

```js
    const [response1] = await Promise.all([
      page.waitForNavigation({ url: /.*\/100/ }),
      page.locator('a[href="/count-to-100"]').click()
    ]);
```

When `count-to-100` is clicked on, it will fire 100 sequential requests, and the `waitForNavigation` url matcher will therefore be called 100 times until the last request matches the given url (in regex form).

mstoykov: Looking at the current code this isn't the case ... :thinking: 

Can you test it?

mstoykov: Oh okay, I did find the path - yes we do give the result of injectRegexMatcherScript down and that can be called multiple times.

mstoykov: I guess in this case there should be a way for you to get a separate taskqueue for this operation only and when WaitForNavigation finishes.

Which I guess begs the quesiton - will it exit on the first event, or what?

ankur22: Are you suggesting:

1. create a dedicated queue just for this method;
2. when WaitForNavigation completes (finds a match) it closes the queue at the end;
3. if WaitForNavigation times out it should also close the queue;
4. we should also ensure that page.close also tries to close any remaining queue incase WaitForNavigation was unable to close the queue.

mstoykov: yes

ankur22: I've started on the proposal: e85a1585ca504f2a92f99ed0c8f0b41189822e33. Still WIP. Now getting other errors which need to be looked into.

inancgumus (commented):
I believe running RegExes on a JavaScript engine and task queue stuff gets complicated, making the code harder to understand and follow. I know we follow this approach for consistency reasons.

What about?
- Proposal 1: Registering another Sobek VM in the module and using it only for running RegExes? A VM that is separate from VU's VM.
- Proposal 2: Using Go's [RE2 RegExes](https://github.com/google/re2/wiki/syntax) and be open to users about this.

mstoykov:
@inancgumus I also entertained the idea, but I would expect that both options will be way more changes and likely will still have other problems. 

IIRC one of the possibilities is that currently a user provides an already parsed regex object, which I do not know how we will support. 

The current changes seem fine ... if a little bit complciated. I have ran some test locally and it seems to work wihtout problems, but I definitely could be missing something

ankur22:
> IIRC one of the possibilities is that currently a user provides an already parsed regex object, which I do not know how we will support.

Yeah correct, and to be consistent with the JS regex engine instead of the golang one.

> Proposal 1: Using Go's [RE2 RegExes](https://github.com/google/re2/wiki/syntax) and be open to users about this.

Yeah, could be an option, maybe a breaking change for v2 if we really need to, but considering we already have to run user provided functions in JS i'm not sure we will escape this complication... although i do like the idea of using a channel and another goroutine dedicated to user provided JS scripts.

inancgumus (dismissed):
I've verified that this change resolves the currently problematic tests 🎉. There is only a linter issue.

codebien (approved):

PR #5017, opened 2025-08-07
gRPC: Stop using protoregistry.GlobalTypes

### LINKED ISSUE #3961: gRPC using reflect race on registering types

joanlopez:
## What?

Introduce the following changes:
- (a) stop using `protoregistry.GlobalTypes` for `grpc.Client.load`, in favor of client-specific ones.
- (b) combined with the use of a `sync.Mutex` per client.

## Why?

To:
- (a) reduce chances of racing on concurrent registrations of the same types - see #3961.
- (b) to prevent concurrent calls of `client.load` to the same client instance from competing/racing.

## Checklist

## Related PR(s)/Issue(s)

Closes #3961

### DIFF
--- a/internal/js/modules/k6/grpc/client_test.go
+++ b/internal/js/modules/k6/grpc/client_test.go
@@ -8,10 +8,8 @@ import (
 	"encoding/pem"
 	"fmt"
 	"math"
-	"reflect"
 	"strings"
 	"testing"
-	"unsafe"
 
 	k6grpc "go.k6.io/k6/internal/js/modules/k6/grpc"
 	"go.k6.io/k6/internal/lib/netext/grpcext"
@@ -20,8 +18,6 @@ import (
 	"go.k6.io/k6/internal/lib/testutils/httpmultibin/grpc_testing"
 	"go.k6.io/k6/internal/lib/testutils/httpmultibin/grpc_wrappers_testing"
 	"go.k6.io/k6/metrics"
-
-	"github.com/grafana/sobek"
 
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
@@ -37,7 +33,6 @@ import (
 	_struct "github.com/golang/protobuf/ptypes/struct"
 	"github.com/golang/protobuf/ptypes/wrappers"
 	"google.golang.org/protobuf/reflect/protoreflect"
-	"google.golang.org/protobuf/reflect/protoregistry"
 	"google.golang.org/protobuf/types/known/wrapperspb"
 
 	"github.com/sirupsen/logrus"
@@ -1510,6 +1505,9 @@ func TestClientLoadProto(t *testing.T) {
 	val, err := ts.Run(tt.initString.code)
 	assertResponse(t, tt.initString, err, val, ts)
 
+	client, ok := val.Export().(*k6grpc.Client)
+	require.True(t, ok, "got: %T, want *k6grpc.Client", client)
+
 	expectedTypes := []string{
 		"grpc.testdata.nested.types.Outer",
 		"grpc.testdata.nested.types.Outer.MiddleAA",
@@ -1519,9 +1517,8 @@ func TestClientLoadProto(t *testing.T) {
 		"grpc.testdata.nested.types.MeldOuter",
 	}
 
-	types := extractTypesFromClientAsSobekValue(t, val)
 	for _, expected := range expectedTypes {
-		found, err := types.FindMessageByName(protoreflect.FullName(expected))
+		found, err := client.Types().FindMessageByName(protoreflect.FullName(expected))
 
 		assert.NotNil(t, found, "Expected to find the message type %s, but an error occurred", expected)
 		assert.Nil(t, err, "It was not expected that there would be an error, but it got: %v", err)
@@ -1546,6 +1543,9 @@ func TestClientLoadProtoAbsoluteRootWithFile(t *testing.T) {
 	val, err := ts.Run(tt.initString.code)
 	assertResponse(t, tt.initString, err, val, ts)
 
+	client, ok := val.Export().(*k6grpc.Client)
+	require.True(t, ok, "got: %T, want *k6grpc.Client", client)
+
 	expectedTypes := []string{
 		"grpc.testdata.nested.types.Outer",
 		"grpc.testdata.nested.types.Outer.MiddleAA",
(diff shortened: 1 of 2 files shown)


internal/js/modules/k6/grpc/client_test.go
```diff
@@ -1597,3 +1606,16 @@ func TestClientConnectionReflectMetadata(t *testing.T) {
...
+	t.Helper()
+
+	client, ok := val.Export().(*k6grpc.Client)
+	require.True(t, ok, "got: %T, want *k6grpc.Client", client)
+
+	clientValue := reflect.ValueOf(client).Elem()
+	typesField := clientValue.FieldByName("types")
+
+	typesField = reflect.NewAt(typesField.Type(), unsafe.Pointer(typesField.UnsafeAddr())).Elem()
+	return typesField.Interface().(*protoregistry.Types)
```
joanlopez: I know this type of magic is generally discouraged, but I hope it's fine as long as it's used only for testing purposes.

The other alternatives I could think of were:
- Renaming the test package name, from `grpc_test` to `grpc` to have access to `Client` private attributes/methods, _but I think it's useful to leave it as-is as long as we can._ 
- Exposing the `types` attribute or some of its methods publicly, _which at the beginning I thought it could also be a valid approach as long as the package lives under `internal`, but then realized that public methods are automatically exposed to the JS layer, so likely a no-go._ 

So, finally I decided that this is the least painful approach.

internal/js/modules/k6/grpc/client.go
```diff
@@ -516,6 +520,13 @@ func (c *Client) convertToMethodInfo(fdset *descriptorpb.FileDescriptorSet) ([]M
 			FullMethod: name,
 		})
 	}
+
+	// We want to avoid concurrent calls to client.load competing
+	// for registering types, as they may end up panicking if they
+	// try to register the same types.
+	c.typesMtx.Lock()
+	defer c.typesMtx.Unlock()
```
joanlopez, line 528: Correct me if I'm wrong, but I think we could still have panics if we don't prevent two concurrent calls to `client.load` from competing, although arguably that's less likely, and most of the cases seen so far are likely due to the use of global types.

mstoykov: As load has to open files it needs to be called in hte first init context, before we stop loading anything from the filesystem.

As such I do think this is unlikely to have any problems, but I am not against keeping hte mutex here

mstoykov (dismissed):

inancgumus (commented):
[Here](https://github.com/grafana/k6/pull/5017#discussion_r2260035931), you don't have to change the package `grpc_test` to `grpc` or export `types` in the code. Instead, you can add a test-only exported `Types` method and call that method in tests (without exposing it to the non-test code).

To do that, create the following `client_export_test.go` file:

```go
package grpc // <- important

import "google.golang.org/protobuf/reflect/protoregistry"

// Types is exported only for testing purposes.
func (c *Client) Types() *protoregistry.Types {
	return c.types
}
```

You can then get `Client.Types` from tests:

```go
client, ok := val.Export().(*k6grpc.Client)
require.True(t, ok, "got: %T, want *k6grpc.Client", client)
return client.Types() // no reflection is needed
```

`Client.Types()` will be available only in tests.

joanlopez:
Right, thanks! @inancgumus

I'm pretty sure I have used that "trick" before, for similar purposes, but this time I wasn't that inspired.
Thanks for suggesting it, much cleaner! 🙌🏻

mstoykov (approved):

PR #5040, opened 2025-08-11
Use new expect() syntax in script templates

andrewslotin:
## What?
Update script templates used by `k6 new` to use the new `expect()` syntax introduced in [k6-jslib-testing](https://github.com/grafana/k6-jslib-testing)

## Why?
`k6-jslib-testing` aims to replace the old `check()` API with a flexible and user-friendly set o matchers. This PR updates the templates used by `k6 new` to use this API.

## Checklist

### DIFF
--- a/internal/cmd/templates/browser.js
+++ b/internal/cmd/templates/browser.js
@@ -1,7 +1,8 @@
 import http from "k6/http";
 import exec from 'k6/execution';
 import { browser } from "k6/browser";
-import { sleep, check, fail } from 'k6';
+import { sleep, fail } from 'k6';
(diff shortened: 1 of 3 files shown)


andrewslotin:
@ankur22, the plan is to move away from `check()` in long-term. In future `k6-testing` will become the go-to API to assert on values and needs to have feature parity with `check()`. I'm wondering what are the use-cases that currently cannot be covered by `expect()`?

oleiade:
@andrewslotin We have, at this point, explicitly excluded emitting assertion-related metrics, by design, because it wasn't part of the initial intent. That's the main difference. We could eventually address that difference, but I'm not entirely sure how much work would be needed.

PR #5024, opened 2025-08-08
Fix panics in waitForURL and waitBy*

mstoykov:
## What?

Fixes 2 panics in new APIs found during documentation creation by @ankur22 

## Why?

Panics are bad, even in new APIs. 

## Checklist

- [ ] I have performed a self-review of my code.
- [ ] I have commented on my code, particularly in hard-to-understand areas.
- [ ] I have added tests for my changes.
- [ ] I have run linter and tests locally (`make check`) and all pass.

## Checklist: Documentation (only for k6 maintainers and if relevant)

**Please do not merge this PR until the following items are filled out.**

- [ ] I have added the correct milestone and labels to the PR.
- [ ] I have updated the release notes: _link_
- [ ] I have updated or added an issue to the [k6-documentation](https://github.com/grafana/k6-docs): grafana/k6-docs#NUMBER if applicable
- [ ] I have updated or added an issue to the [TypeScript definitions](https://github.com/grafana/k6-DefinitelyTyped/tree/master/types/k6): grafana/k6-DefinitelyTyped#NUMBER if applicable

PR #5051, opened 2025-08-13
Add tests to ensure correct `null` checks in `page.getBy*` APIs

ankur22:
## What?

This does two things:

1. Adds a `nullish` check on `getByRole`.
2. Adds tests for correct `null` handling.

## Why?

To avoid incorrect behaviour and panics.

## Checklist

- [x] I have updated the release notes: 5aefc356fd2c8d2b56a3d9403721d6b759c57744

## Related PR(s)/Issue(s)

https://github.com/grafana/k6/pull/5024

### DIFF
--- a/internal/js/modules/k6/browser/tests/get_by_test.go
+++ b/internal/js/modules/k6/browser/tests/get_by_test.go
@@ -1296,3 +1296,53 @@ func TestGetByTextSuccess(t *testing.T) {
 		})
 	}
 }
+
+func TestGetByNullHandling(t *testing.T) {
+	t.Parallel()
+
+	tb := newTestBrowser(t, withFileServer())
+	tb.vu.ActivateVU()
+	tb.vu.StartIteration(t)
+
+	// Setup
+	tb.vu.SetVar(t, "page", &sobek.Object{})
+	_, err := tb.vu.RunAsync(t, `
+		page = await browser.newPage();
+	`)
+	require.NoError(t, err)
+
+	_, err = tb.vu.RunAsync(t, `
+		await page.getByRole().click();
+	`)
+	require.ErrorContains(t, err, "missing required argument 'role'")
+
+	_, err = tb.vu.RunAsync(t, `
+		await page.getByAltText().click();
+	`)
+	require.ErrorContains(t, err, "missing required argument 'altText'")
+
+	_, err = tb.vu.RunAsync(t, `
+		await page.getByLabel().click();
+	`)
+	require.ErrorContains(t, err, "missing required argument 'label'")
+
+	_, err = tb.vu.RunAsync(t, `
+		await page.getByPlaceholder().click();
+	`)
+	require.ErrorContains(t, err, "missing required argument 'placeholder'")
+
+	_, err = tb.vu.RunAsync(t, `
+		await page.getByTitle().click();
+	`)
+	require.ErrorContains(t, err, "missing required argument 'title'")
+
+	_, err = tb.vu.RunAsync(t, `
+		await page.getByTestId().click();
+	`)
+	require.ErrorContains(t, err, "missing required argument 'testId'")
+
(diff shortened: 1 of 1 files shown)


internal/js/modules/k6/browser/tests/get_by_test.go
```diff
@@ -1296,3 +1296,53 @@ func TestGetByTextSuccess(t *testing.T) {
...
+	tb.vu.SetVar(t, "page", &sobek.Object{})
+	_, err := tb.vu.RunAsync(t, `
+		page = await browser.newPage();
+	`)
+	require.NoError(t, err)
+
+	_, err = tb.vu.RunAsync(t, `
+		await page.getByRole().click();
+	`)
+	require.ErrorContains(t, err, "missing required argument 'role'")
```
ankur22, line 1317: There is a lot of logic in the mapping layer that should be tested, not just mocked. This is why I have used this approach to testing rather than pure Go tests. Let me know what you think.

inancgumus: I believe these tests are valuable, but they are not type-safe and becomes very difficult to adopt later on. For example, we can't easily find which tests use which Go API (unless we do text search rather than "find callers" in our IDEs). I currently don't have a nice idea for a solution. I usually write Go tests if the mapping logic is simple. Or separate the mapping logic in another function and write a test for it.

ankur22: Yeah, i see your point, especially around being able to easily identify where the usages of an API are from within the IDE.

Splitting the two bites of logic and testing them in isolation is also a good way forward, and at least that keeps it within Go.

My worry is that we wouldn't be able to create an integration test that takes the whole flow into account. We could move our integration tests into JS files, like the `examples` but more focused on testing the browser module rather than used as examples for users to copy.

mstoykov: I actually find it strange that so much of the test do not actually test the js API directly. 

Nobody but us use the internal API, so it is implementation detail, while this tests are had to be called unit tests either way. 

Also so much depends on how the JS runtime behaves that there ened up being  bunch of things to emulate it just to get test to be  able to run. 

Historically hte JS runtime had big overhead due to babel+core.js, but this has not been the case for a long time, as long as you do not use import/export. And ESM fixed that over a year ago.

Given the browser I am nto certain this overhead was a big problem to begin with. 

> where the usages of an API are from within the IDE.

I find this to not really be a problem, but I guess that depends on your style of development. 

I can definitely tell you that I had to do *way more* changes in the browser code where due to using go to call it instead of js I had to go over a everywehre to add a "not required" argument or whatever.

ankur22: Right, I do see value in both styles of testing. One leans towards more unit testing and the other is integration.

I think one of the issues I have with the browser module is that it's very difficult to properly unit test without needing a real browser. If we had the time to properly encapsulate different components/abstractions then we could create a lot more unit tests to test the internals in isolation, and have integration tests which test the full flow of all the APIs and any edge cases.

For now, when i see value in wanting to test the mapping layer, i'll stick with this approach of working with JS, otherwise fall back to trying to create "unit tests".

inancgumus: I feel both kinds of tests have their place. Since most of our code lives in Go, I naturally lean toward Go tests (business logic tests). I don’t see Go as some “internal detail” — it’s the foundation of what we build. Keeping things isolated and easy to test is just part of writing good software.

That said, JS (or JS API) tests are also helpful when they give us something Go tests can’t. They work best for checking the high-level API mapping, without trying to cover every detail. After all, the mapping layer’s main job (like `getByRole`) is to connect JS to Go, while the real logic sits in Go and can be tested directly.

PR #5052, opened 2025-08-13
Update, add and refactor `waitForURL` and `waitForNavigation` tests

ankur22:
## What?

This PR aims to tidy up the tests and add new ones where necessary including a refactor in the implementation for `waitForURL`:

1. Refactor to work with `parseStringOrRegex` in `waitForURL`;
2. Refactor tests so that they're easier to maintain and fix when/if they fail;
3. Refactor tests so that they're not flakey;
4. Add `waitForURL` tests to `frame`.

## Why?

Refactors based on comments made in this PR: https://github.com/grafana/k6/pull/4920
- https://github.com/grafana/k6/pull/4920#discussion_r2215746887
- https://github.com/grafana/k6/pull/4920#discussion_r2238980488

## Checklist

## Related PR(s)/Issue(s)

https://github.com/grafana/k6/pull/4920

### DIFF
--- a/internal/js/modules/k6/browser/tests/frame_test.go
+++ b/internal/js/modules/k6/browser/tests/frame_test.go
@@ -6,9 +6,11 @@ package tests
 import (
 	"context"
 	"net/http"
+	"strings"
 	"testing"
 	"time"
 
+	"github.com/grafana/sobek"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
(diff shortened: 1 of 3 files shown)


internal/js/modules/k6/browser/tests/frame_test.go
```diff
@@ -242,3 +244,119 @@ func TestFrameSetChecked(t *testing.T) {
...
+			const page = await browser.newPage();
+
+			await page.setContent('<iframe></iframe>');
+
+			const iframeElement = await page.$('iframe');
+			frame = await iframeElement.contentFrame();
+		`)
+	require.NoError(t, err)
+
+	// Test when already at matching URL (should just wait for load state)
```
ankur22: Should I make this into a table test?

codebien: It might make sense to have it

ankur22: Done in ~a3cf2b0c0f754d3409c5fa1df072d7e3ee32064e~ db6537e3fff74c554955434e3c29b01fad1f78d7

internal/js/modules/k6/browser/tests/frame_test.go
```diff
@@ -242,3 +244,159 @@ func TestFrameSetChecked(t *testing.T) {
 	require.NoError(t, err)
 	assert.False(t, checked)
 }
+
+//nolint:tparallel
```
codebien: This is unexpected. Do we really need? Can't we mitigate it? How can we guarantee that we are not hiding a race condition here? 

If this is needed, then the comment should document _The Why_.

ankur22: So here's my motivation for doing it:

1. Parallelising browser integration tests i think is quite expensive due to them needing their own chrome instance. So i personally think we should possibly reconsider this approach with browser tests, mainly in CI. I don't have numbers though, so i can't tell you whether it's a good assumption.
2. I've started to prefer trying to perform a single setup function at the top, with only the important test logic in the test block in `for _, tt := range tests`.

Since we've not discussed this before i'll change it to working with parallel tests for now.

Done in 58fcc25b1c77e699c1b759c133c22f8eff4de92a

codebien: Yep, I have the feeling the problem is quite complex and we should open an issue identifying the problem. I see various possible ways for addressing it. For example, did we consider to have a pool of browsers to use? It would automatically rate limit the parallelism.
However, any assumption if not supported by data could be difficult to justify.

> I don't have numbers though, so i can't tell you whether it's a good assumption.

So this sounds a requirement if we want to optimize in any direction.

However, if the decision is to stop the parallelism then I'd prefer if we add a condition for the entire package. So we don't have to add it for each test.

codebien (approved):

Issue #4980, opened 2025-07-29
`click` fails when clicking on button in `iframe`s


ankur22:
### Brief summary

NOTE: It's unclear exactly how to recreate a website that exhibits this behaviour at the moment.

It's possible to retrieve the `iframe`, the `frameContent` and even details of the button that we want to click on such as the `textContent`.

What fails is the `click` itself. It seems that the browser module isn't calculating the coordinates correctly relative to the main page, and instead it is creating them relative to the `iframe` only.

### k6 version

NA

### OS

NA

### Steps to reproduce the problem

TBD

### Expected behaviour

`click` clicks on the button in the `iframe`.

### Actual behaviour

Nothing happens, and actually an error is being suppressed somewhere when the click fails.

PR #5071, opened 2025-08-19
Fix `click` for CORS frames

ankur22:
## What?

This fixes `click` so that it works even when clicking on elements from within a frame that is from a different origin to the page.

## Why?

Before the fix, the click point that was calculated was relative to the frame the element that is being clicked on is in. Now the click point is calculated (and if need be) adjusted so that it is relative to the page.

Without the fix, some users would have to retrieve the iframe, and then work with `evaluate` to click on the element in the iframe e.g.:

```js
    iframeContent.evaluate(() => {
      document.querySelector('button').click();
    });
```

With the fix they can now just work with `click` as you'd expect:

```js
    iframeContent.locator('button').click();
```

## Test

The test that was written looks quite complicated. You can run the same test in `master` and it will fail, whereas it passes after the fix is applied. Basically the page renders like this, and the test clicks on each of the buttons ensuring that the count is equal to 1:

Screenshot 2025-08-19 at 14 03 56

## Checklist

## Related PR(s)/Issue(s)

https://github.com/grafana/k6/issues/4980

### DIFF
--- a/internal/js/modules/k6/browser/tests/page_test.go
+++ b/internal/js/modules/k6/browser/tests/page_test.go
@@ -9,6 +9,7 @@ import (
 	"io"
 	"net"
 	"net/http"
+	"net/http/httptest"
 	"runtime"
 	"slices"
 	"strconv"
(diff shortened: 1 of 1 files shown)


ankur22:
> I also do not know if some of this will be a big problem, but I guess we will have to try something and nothing here seems off.

Yep, hopefully we'll spot any issues once this is merged in.

> Maybe we can get some consultancy from some frontend people if you have some concrete questiones @ankur22 , but I do not have any .

Yeah, good shout, i'll ask them just in case.

internal/js/modules/k6/browser/common/page.go
```diff
@@ -823,7 +823,7 @@ func (p *Page) getOwnerFrame(apiCtx context.Context, h *ElementHandle) (cdp.Fram
 	}
 
 	action := dom.DescribeNode().WithObjectID(documentElement.remoteObject.ObjectID)
-	node, err := action.Do(cdp.WithExecutor(p.ctx, p.session))
+	node, err := action.Do(cdp.WithExecutor(p.ctx, h.session))
```
inancgumus, line 826: Do you think it would be helpful to include [this commit](https://github.com/grafana/k6/pull/5071/commits/38a49a8ac468118ff77ed82f9a249e66636fe908)'s comment here? It might clear things up for anyone wondering why we're using the `ElementHandle` session instead of the `Page` session.

internal/js/modules/k6/browser/common/element_handle.go
```diff
@@ -156,132 +156,12 @@ func (h *ElementHandle) click(p *Position, opts *MouseClickOptions) error {
...
-	if ry := p.Y - math.Floor(p.Y); ry > 0.49 && ry < 0.51 {
-		p.Y -= 0.02
-	}
-	return p
+// This will get the clickable point of the element relative to the page even
+// when the element is in an iframe. In some cases this isn't the case and we
+// need to translate the point to the page.
+func (h *ElementHandle) clickablePoint() *Position {
+	r := h.BoundingBox()
+	return &Position{X: r.X + r.Width/2, Y: r.Y + r.Height/2}
```
inancgumus, line 202: I believe this significant change requires more testing, as it seems to alter behavior that affects every pointer action. Do you think we have enough tests to make this change?

inancgumus (approved):
LGTM, but it feels like it needs more testing.

internal/js/modules/k6/browser/common/element_handle.go
ankur22: Right, i agree. With the limited testing i've done, it does work. I need to test it on a larger scale, so i'm hoping to deploy it next week (if we merge it in it soon) and then keep an eye on test runs to see what the impact is.

inancgumus: 🙇

internal/js/modules/k6/browser/common/page.go
ankur22: Added in https://github.com/grafana/k6/pull/5086

PR #5075, opened 2025-08-20
Add `frameLocator`, `frameLocator.locator` and `locator.contentFrame`

ankur22:
## What?

This adds a few new APIs. It would have been difficult to implement `frameLocator` on its own without the other two APIs:

1. It implements `frameLocator` which allows the user to interact with elements within `iframe`s with `locator` like APIs.
2. On top of that it implements `locator.contentFrame` which returns a `frameLocator`, allowing the user to actually retrieve a `frameLocator` to start working with elements from with an `iframe`.
3. It implements `frameLocator.locator` allowing users to actually select elements in an `iframe`.

## Why?

Working with `locator` like APIs is important for maintainable, less flakey tests. It's also a preferred way of working when interacting with `iframe`s in Playwright. We're currently on a drive to improve k6 browser's parity to make migrating from Playwright to k6 easier.

Essentially, `locator` and `frameLocator` APIs will retrieve the latest version of an element or perform an action on the latest version of an element when such APIs are called. Currently websites are very dynamic and we always want to try to work with the latest version of the DOM when trying to retrieve data or perform actions. Working with `page$` and `frame.contentFrame` are currently the only way to interact with elements in an `iframe`, but these are flakey since they work with old versions of the DOM.

The end result of this implementation also allows us to easily chain selectors. Before we used to have to do (imagine having to navigate to multiple nested `iframe`s):

```js
const frameB = await page.$('#frameB');
const frameBContent = await frameB.contentFrame(); // Takes a snapshot of the DOM

const frameC = await frameBContent.$('#frameB');
const frameCContent = await frameC.contentFrame(); // Takes a snapshot of the DOM

await frameCContent.locator("#increment", nil).click();

const count = await frameCContent.locator("#count").textContent();
```

and now we can do:

```js
const frameCContent = page.locator('#frameB').contentFrame().locator('#frameC').contentFrame();

await frameCContent.locator("#increment", nil).click(); // Works with the latest DOM and performs actionability checks.

const count = await frameCContent.locator("#count").textContent(); // Works with the latest DOM and performs actionability checks.
```

## Checklist

## Related PR(s)/Issue(s)

https://github.com/grafana/k6/issues/5032

### DIFF
--- a/internal/js/modules/k6/browser/common/element_handle.go
+++ b/internal/js/modules/k6/browser/common/element_handle.go
@@ -778,6 +778,42 @@ func (h *ElementHandle) count(apiCtx context.Context, selector string) (int, err
 		return 0, err
 	}
 
+	// Check for frame navigation in the selector
+	frameNavIndex := h.findFrameNavigationIndex(parsedSelector)
+	if frameNavIndex != -1 {
+		// Split selector at frame navigation boundary
+		beforeFrame, afterFrame := h.splitSelectorAtFrame(parsedSelector, frameNavIndex)
+
+		// Find the iframe element using the "before frame" selector
+		iframeSelector := h.reconstructSelector(beforeFrame)
+
+		// TODO: We need to define a new opts outside of count depending on who is calling it (frame vs locator).
+		iframeHandle, err := h.waitForSelector(apiCtx, iframeSelector, &FrameWaitForSelectorOptions{
+			State:   DOMElementStateAttached,
+			Timeout: h.frame.defaultTimeout(),
+			Strict:  true,
+		})
+		if err != nil {
+			return 0, fmt.Errorf("finding iframe with selector %q: %w", iframeSelector, err)
+		}
+
+		if iframeHandle == nil {
+			// TODO: Do we need to correctly handle this?
+			return 0, nil // No iframe found
+		}
+
+		frame, err := iframeHandle.ContentFrame()
+		if err != nil {
+			return 0, fmt.Errorf("getting iframe frame: %w", err)
+		}
+
+		// Count in the iframe document using the "after frame" selector
+		afterFrameSelector := h.reconstructSelector(afterFrame)
+
+		return frame.count(afterFrameSelector)
+	}
+
+	// No frame navigation - proceed with normal count logic
 	fn := `
 			(node, injected, selector) => {
 				return injected.count(selector, node);
--- a/internal/js/modules/k6/browser/common/element_handle.go
+++ b/internal/js/modules/k6/browser/common/element_handle.go
@@ -725,9 +725,10 @@ func (h *ElementHandle) stepIntoFrame(
 		return nil, "", fmt.Errorf("finding iframe with selector %q: %w", iframeSelector, err)
 	}
 
+	// This is a valid response from waitForSelector. It means that the element
+	// was either hidden or detached.
 	if iframeHandle == nil {
-		// TODO: Do we need to correctly handle this?
-		return nil, "", nil // No iframe found
+		return nil, "", errors.New("check if element is visible")
 	}
 
 	frame, err := iframeHandle.ContentFrame()
@@ -790,6 +791,8 @@ func (h *ElementHandle) waitForSelector(
 	case *ElementHandle:
 		return r, nil
 	default:
+		// This is a valid response, which means that the element was either
+		// hidden or detached.
 		return nil, nil //nolint:nilnil
 	}
 }
--- a/internal/js/modules/k6/browser/common/element_handle.go
+++ b/internal/js/modules/k6/browser/common/element_handle.go
@@ -708,6 +708,39 @@ func (h *ElementHandle) waitForElementState(
 		"waiting for states %v of element %q", states, reflect.TypeOf(result))
 }
 
+// stepIntoFrame steps into an iframe/frame. Due to CORS, we need to perform this
+// step outside of the browser (chromium). It returns the frame that it has stepped
+// into and the selector to use within that frame.
+func (h *ElementHandle) stepIntoFrame(
+	apiCtx context.Context, parsedSelector *Selector, frameNavIndex int, opts *FrameWaitForSelectorOptions,
+) (*Frame, string, error) {
+	// Split selector at frame navigation boundary
+	beforeFrame, afterFrame := h.splitSelectorAtFrame(parsedSelector, frameNavIndex)
+
+	// Find the iframe element using the "before frame" selector
+	iframeSelector := h.reconstructSelector(beforeFrame)
+
+	iframeHandle, err := h.waitForSelector(apiCtx, iframeSelector, opts)
+	if err != nil {
+		return nil, "", fmt.Errorf("finding iframe with selector %q: %w", iframeSelector, err)
+	}
+
+	if iframeHandle == nil {
+		// TODO: Do we need to correctly handle this?
+		return nil, "", nil // No iframe found
+	}
+
+	frame, err := iframeHandle.ContentFrame()
+	if err != nil {
+		return nil, "", fmt.Errorf("getting iframe frame: %w", err)
+	}
+
+	// Wait for selector in the iframe using the "after frame" selector
+	afterFrameSelector := h.reconstructSelector(afterFrame)
+
+	return frame, afterFrameSelector, nil
+}
+
 func (h *ElementHandle) waitForSelector(
 	apiCtx context.Context, selector string, opts *FrameWaitForSelectorOptions,
 ) (*ElementHandle, error) {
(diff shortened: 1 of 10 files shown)


internal/js/modules/k6/browser/common/element_handle.go
```diff
@@ -780,6 +780,38 @@ func (h *ElementHandle) waitForSelector(
 	if err != nil {
 		return nil, err
 	}
+
+	// Check for frame navigation in the selector
```
ankur22, line 784: TODO: Add test

internal/js/modules/k6/browser/common/element_handle.go
```diff
@@ -811,6 +843,42 @@ func (h *ElementHandle) count(apiCtx context.Context, selector string) (int, err
 		return 0, err
 	}
 
+	// Check for frame navigation in the selector
+	frameNavIndex := h.findFrameNavigationIndex(parsedSelector)
```
ankur22, line 807: TODO: Add test

internal/js/modules/k6/browser/common/element_handle.go
```diff
@@ -1213,6 +1326,41 @@ func (h *ElementHandle) Query(selector string, strict bool) (_ *ElementHandle, r
 	if err != nil {
 		return nil, fmt.Errorf("parsing selector %q: %w", selector, err)
 	}
+
+	// Check for frame navigation in the selector
+	frameNavIndex := h.findFrameNavigationIndex(parsedSelector)
```
ankur22, line 1278: TODO: Add test

internal/js/modules/k6/browser/common/frame.go
```diff
@@ -403,6 +403,26 @@ func (f *Frame) onLoadingStopped() {
 	//       website never stops performing network requests.
 }
 
+func (f *Frame) boundingBox(selector string, opts *FrameBaseOptions) (*Rect, error) {
```
ankur22: TODO: Moved to https://github.com/grafana/k6/pull/5076, so drop from here

internal/js/modules/k6/browser/common/element_handle.go
```diff
@@ -691,13 +691,68 @@ func (h *ElementHandle) waitForElementState(
...
+
+	iframeHandle, err := h.waitForSelector(apiCtx, iframeSelector, opts)
+	if err != nil {
+		return nil, "", fmt.Errorf("finding iframe with selector %q: %w", iframeSelector, err)
+	}
+
+	// This is a valid response from waitForSelector. It means that the element
+	// was either hidden or detached.
+	if iframeHandle == nil {
+		return nil, "", errors.New("check if element is visible")
```
ankur22, line 731: TODO: Use new visible error from https://github.com/grafana/k6/pull/5111

internal/js/modules/k6/browser/common/element_handle.go
```diff
@@ -691,13 +691,68 @@ func (h *ElementHandle) waitForElementState(
...
+		opts := &FrameWaitForSelectorOptions{
+			State:   DOMElementStateAttached,
+			Timeout: opts.Timeout,
+			Strict:  true,
+		}
+
+		frame, afterFrameSelector, err := h.stepIntoFrame(apiCtx, parsedSelector, frameNavIndex, opts)
+		if err != nil {
+			return nil, err
+		}
```
AgnesToulet, line 767: What do you think about extracting this to another function as it's used a lot throughout the file?

internal/js/modules/k6/browser/common/element_handle.go
```diff
@@ -1131,6 +1253,27 @@ func (h *ElementHandle) Query(selector string, strict bool) (_ *ElementHandle, r
 	if err != nil {
 		return nil, fmt.Errorf("parsing selector %q: %w", selector, err)
 	}
+
+	// Check for frame navigation in the selector
+	frameNavIndex := h.findFrameNavigationIndex(parsedSelector)
```
AgnesToulet, line 1278: Should this be added to the `QueryAll` function as well?

internal/js/modules/k6/browser/tests/locator_test.go
```diff
@@ -795,50 +797,136 @@ func TestSelectOption(t *testing.T) {
...
+		_, err := p.Goto(
+			tb.staticURL("locators.html"),
+			opts,
+		)
+		require.NoError(t, err)
+
+		return tb, p
+	}
+
+	setupCORS := func(t *testing.T) (*testBrowser, *common.Page) {
```
AgnesToulet, line 818: This would be nice to extract this in a utility function.

AgnesToulet (commented):
This looks great, I tested it with some basic tests and it seems to work fine.

Here's my first thoughts while I try to test it a bit more.

internal/js/modules/k6/browser/common/element_handle.go
ankur22: Right, I was hoping to do a better job at abstracting this. I managed to abstract most of the core logic into `stepIntoFrame`, but some of it still remains (call to `findFrameNavigationIndex`, creation of `FrameWaitForSelectorOptions` and a recursive call to the method it is in). I feel like what is left isn't easily generalised, there are ever so slight differences.

I won't completely disregard this, and think it through again. Let me know if you spot anything obvious though.

internal/js/modules/k6/browser/common/element_handle.go
ankur22: I did originally consider it but decided against it. The reason for not wanting to add to `queryAll` is that there's currently no way to chain such a directive to a selector for it to work in `queryAll`. This chaining and stepping into `iframes` from a different origin is really only there to support `locator` and `frameLocator` based APIs. `queryAll` isn't a `locator` based API, and I think it shouldn't be needed now that we have `locator.all` 🤔 Looking at the Playwright docs (and a quick google/AI search) it looks like there's no way to step into an iframe with `queryAll`.

I added it to `query` because `locator.isVisible` and `locator.isHidden` use it -- i do wonder if they should be refactored to not work with `query` 🤔

internal/js/modules/k6/browser/tests/locator_test.go
ankur22: You mean that ability to create several servers to mimic the CORS behaviour?

internal/js/modules/k6/browser/common/frame_locator.go
```diff
@@ -0,0 +1,34 @@
+package common
+
+import (
+	"context"
+
+	"go.k6.io/k6/internal/js/modules/k6/browser/log"
+)
+
+// FrameLocator represent a way to find element(s) in an iframe.
+type FrameLocator struct {
```
inancgumus, line 10: It's OK, but I'd expect to find this in the same `locator.go` file.

inancgumus (dismissed):

internal/js/modules/k6/browser/common/frame_locator.go
ankur22: Why is that? It's a new type, which does work like a locator, but it's use case is specific to when working with `iframe`s.

inancgumus: Sorry, I could've explained it better 😄 I like the idiomatic Go approach of collecting multiple types and functionalities in the same file for related functionality (e.g., `http`'s [client](https://cs.opensource.google/go/go/+/refs/tags/go1.25.0:src/net/http/client.go) and [server](https://cs.opensource.google/go/go/+/refs/tags/go1.25.0:src/net/http/server.go) _aspects_). Although `FrameLocator` is a new type, it's one aspect of a `Locator`.

internal/js/modules/k6/browser/common/element_handle.go
AgnesToulet: I had an idea but when trying to implement it, it's too complicated and doesn't help with readability.

internal/js/modules/k6/browser/tests/locator_test.go
AgnesToulet: For now, just the ability to create a page with an iframe. If I'm not missing anything, the difference with [this](https://github.com/grafana/k6/pull/5075/files#diff-e3251587a1e2960de20c10c1fb124d626fcdcc9976638e8e8c7d382232a7c8b8R1042-R1101) is only the `originBHTML` content that could be a parameter.

AgnesToulet (approved):
LGTM! Just replied to your question on the test improvement and there is also [your comment](https://github.com/grafana/k6/pull/5075/files#r2309976139) that still needs to be addressed but nothing blocking.

ankur22:
> LGTM! Just replied to your question on the test improvement and there is also [your comment](https://github.com/grafana/k6/pull/5075/files#r2309976139) that still needs to be addressed but nothing blocking.

Thanks! i'll open a new issue and PR for the refactoring work for the PR, but i'll try to get this merged in asap as is for now.

EDIT: Issue is https://github.com/grafana/k6/issues/5129

PR #5076, opened 2025-08-20
Add `locator.boundingbox`

ankur22:
## What?

This change adds `boundingBox` to `locator`.

## Why?

When working on another issue (https://github.com/grafana/k6/pull/5071) I found it useful to work with this API. It's also part of Playwright, so bringing this in will give us better parity with Playwright.

## Checklist

## Related PR(s)/Issue(s)

https://github.com/grafana/k6/issues/4490

### DIFF
--- a/internal/js/modules/k6/browser/browser/mapping_test.go
+++ b/internal/js/modules/k6/browser/browser/mapping_test.go
@@ -528,6 +528,7 @@ type responseAPI interface { //nolint:interfacebloat
 // locatorAPI represents a way to find element(s) on a page at any moment.
 type locatorAPI interface { //nolint:interfacebloat
 	All() ([]*common.Locator, error)
+	BoundingBox(opts *common.FrameBaseOptions) (*common.Rect, error)
 	Clear(opts *common.FrameFillOptions) error
 	Click(opts sobek.Value) error
 	Count() (int, error)
--- a/internal/js/modules/k6/browser/browser/locator_mapping.go
+++ b/internal/js/modules/k6/browser/browser/locator_mapping.go
@@ -29,6 +29,15 @@ func mapLocator(vu moduleVU, lo *common.Locator) mapping { //nolint:funlen
 				return res, nil
 			})
 		},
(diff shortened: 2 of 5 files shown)


inancgumus (dismissed):

internal/js/modules/k6/browser/browser/mapping_test.go
```diff
@@ -528,6 +528,7 @@ type responseAPI interface { //nolint:interfacebloat
 // locatorAPI represents a way to find element(s) on a page at any moment.
 type locatorAPI interface { //nolint:interfacebloat
 	All() ([]*common.Locator, error)
+	BoundingBox(opts *common.FrameBaseOptions) (*common.Rect, error)
```
joanlopez, line 531: I'm confused, why do we sometimes define methods in this interface where:
- `opts` is a `sobek.Value`
- and sometimes we define them as a concrete type.

What's even more confusing is that if I look for instance at the "Click" implementation, in the interface is defined like:

```go
Click(opts sobek.Value) error
```

while the actual implementation looks like:

```
func (l *Locator) Click(opts *FrameClickOptions) error {
```

🤔

ankur22: Aha, this probably does look a bit confusing. So this interface is only used in mapping_test.go, otherwise everything else is a concrete type. The interface is compared against what is implemented by the mapping layer for each type, e.g. for locator we compare the [interface](https://github.com/grafana/k6/blob/master/internal/js/modules/k6/browser/browser/mapping_test.go#L529-L562) with what was [mapped](https://github.com/grafana/k6/blob/master/internal/js/modules/k6/browser/browser/locator_mapping.go#L17-L214).


> What's even more confusing is that if I look for instance at the "Click" implementation, in the interface is defined like:

This inconsistency probably should be fixed, but at the moment the test doesn't care about the input arguments, just the [name of the methods](https://github.com/grafana/k6/blob/master/internal/js/modules/k6/browser/browser/mapping_test.go#L81-L116).

joanlopez (approved):

inancgumus (approved):

PR #5057, opened 2025-08-14
fix: do not panic on `k6 login cloud`

mstoykov:
## What?

This happened due to auto extension resolution wrongly guessing that `cloud` here means it needs to analyze a script. But it doesn't, and consequently panics.

## Why?
Panics are bad.
## Checklist

- [ ] I have performed a self-review of my code.
- [ ] I have commented on my code, particularly in hard-to-understand areas.
- [ ] I have added tests for my changes.
- [ ] I have run linter and tests locally (`make check`) and all pass.

## Checklist: Documentation (only for k6 maintainers and if relevant)

**Please do not merge this PR until the following items are filled out.**

- [ ] I have added the correct milestone and labels to the PR.
- [ ] I have updated the release notes: _link_
- [ ] I have updated or added an issue to the [k6-documentation](https://github.com/grafana/k6-docs): grafana/k6-docs#NUMBER if applicable
- [ ] I have updated or added an issue to the [TypeScript definitions](https://github.com/grafana/k6-DefinitelyTyped/tree/master/types/k6): grafana/k6-DefinitelyTyped#NUMBER if applicable

## Related PR(s)/Issue(s)
Fixes #5055

PR #4866, opened 2025-06-19
js: Error with a list of unknown `k6/*` modules

mstoykov:
This doesn't abort immediately the execution, but continues the resolution step and then errors if there are any unknown modules.

This potentially might be used by binary provisioning in the future.

## Checklist

- [ ] I have performed a self-review of my code.
- [ ] I have commented on my code, particularly in hard-to-understand areas.
- [ ] I have added tests for my changes.
- [ ] I have run linter and tests locally (`make check`) and all pass.

## Checklist: Documentation (only for k6 maintainers and if relevant)

**Please do not merge this PR until the following items are filled out.**

- [ ] I have added the correct milestone and labels to the PR.
- [ ] I have updated the release notes: _link_
- [ ] I have updated or added an issue to the [k6-documentation](https://github.com/grafana/k6-docs): grafana/k6-docs#NUMBER if applicable
- [ ] I have updated or added an issue to the [TypeScript definitions](https://github.com/grafana/k6-DefinitelyTyped/tree/master/types/k6): grafana/k6-DefinitelyTyped#NUMBER if applicable

PR #5081, opened 2025-08-21
fix: better detection of root command for auto extension resolution

mstoykov:
## What?

Better detect the top root command when trying to figure if auto extension resolution feature needs to scan scripts.

## Why?
After https://github.com/grafana/k6/pull/5057 the check tries to figure out the top command by checkign if it is equal to `k6`. Unfortuantely in order for the `--help` output to be copy-pastable, it does use the real binary name. 

This does mean that it will also not work if somebody has the binary as `k6-cool` or `k6.v1.2.2`. 
## Checklist

## Related PR(s)/Issue(s)

https://github.com/grafana/k6/pull/5057


mstoykov:
@grafana/k6-core this might require `.3` release as it might mean the auto extension resolution isn't working for some users using `v1.2.2`.

I guess this depends on how often k6 is ran like this.

internal/cmd/launcher.go
```diff
@@ -335,24 +335,27 @@ func isAnalysisRequired(cmd *cobra.Command) bool {
 	switch cmd.Name() {
 	case "run":
 		if parent := cmd.Parent(); parent != nil {
-			switch parent.Name() {
-			case "k6", "cloud":
-				return true
-			}
+			return isCommandName(cmd, "cloud") || isRootCommand(parent)
```
joanlopez: I think the condition is different now, as the previous code was comparing whether `cmd.Parent().Name() == "cloud"` while now it compares if `cmd.Name() == "cloud"`.

Indeed, considering that it is already under the case where `cmd.Name() == "run"`, I think the first condition, as-is: `isCommandName(cmd, "cloud")` will never be satisfied, and it should call `isCommandName()` with `parent` instead.

```suggestion
			return isCommandName(parent, "cloud") || isRootCommand(parent)
```

mstoykov: Ah yes, I did add the isCommandName to remove some of the ifs, but apperantly didn't do great at actually using it :facepalm:

internal/cmd/launcher.go
```diff
@@ -335,24 +335,27 @@ func isAnalysisRequired(cmd *cobra.Command) bool {
...
-			switch parent.Name() {
-			case "k6", "cloud":
-				return true
-			}
+			return isCommandName(parent, "cloud") || isRootCommand(parent)
 		}
 	case "upload":
 		if parent := cmd.Parent(); parent != nil {
-			return parent.Name() == "cloud"
+			return isCommandName(cmd, "cloud")
```
joanlopez: Same as above

```suggestion
			return isCommandName(parent, "cloud")
```

mstoykov: I have actually redone this part to be as I planned when I added the `nil` check for the `cmd` parameter

joanlopez (dismissed):

mstoykov:
> In addition, it would be good to have a test for it as it is a case difficult to remember and to eventually try manually.

I do not know how to make a test about what the final binary name will be, but will try to figure it out tomorrow.

Lets see if it will fix the problem for @pablochacin first I guess

oleiade (dismissed):

internal/cmd/launcher_test.go
```diff
@@ -386,6 +387,7 @@ func TestIsAnalysisRequired(t *testing.T) {
 
 			args := append([]string{"k6"}, tc.args...)
 			ts := tests.NewGlobalTestState(t)
+			ts.BinaryName = "somethingelse" // specifically set to not be the default k6
```
```suggestion
			// Specifically set to not be the default k6 name.
			// It asserts the scenario where a user has a custom name for the binary,
			// such as k6v1.2.2, which is useful for managing multiple installed versions.
			ts.BinaryName = "somethingelse"
```

internal/cmd/launcher.go
```diff
@@ -332,27 +332,27 @@ func analyze(gs *state.GlobalState, args []string) (k6deps.Dependencies, error)
...
+		"k6 cloud",
+		"k6 cloud run",
+		"k6 cloud upload",
+		"k6 upload",
+		"k6 archive",
+		"k6 inspect":
+		return true
 	}
 
 	return false
```
codebien, line 346: Now, I'm realizing that an improvement should be to list the cases where we wan to return false + the default section with a panic.

I'm concerned that if we add a new command, we might forget to include it in the `true` list. Since it would be automatically excluded, we wouldn't catch the issue until we ran a script that specifically required the command to resolve an extension. And I see the difficulty that we're going to remember to include the extension resolution for automated or manual tests, so there's a high chance we could release a version where this problem isn't discovered until a user encounters it.

mstoykov: This is the case in reverse as well. If we add a new command and it shouldn't be analysed. This is just one more reason why we shouldn't be doing it as a wrapper but as part of the usual k6 flow and when we load the script to analyze it.

codebien:
```suggestion
switch stringifyCommand(cmd) {
	case "k6 run",
		"k6 cloud",
		"k6 cloud run",
		"k6 cloud upload",
		"k6 upload",
		"k6 archive",
		"k6 inspect":
		return true
	case "k6 login cloud", "k6 login influxdb", others?:
	    return false
	default:
	 panic("unexpected command executed"+ stringifiedCommand)
	}
```

To be explicit, something like this.

mstoykov: Well we do have way more commands than that so it will just be constnatly growing list . 

Again IMO this is just a bandaid to fix the concrete issue, the real fix will be for this to not try to guess what the command is and whether it will have a script or not. But for the k6 logic to be that when we get to the point that k6 knows it has to touch a script

codebien: >This is the case in reverse as well.

Yes, it is.

>This is just one more reason why we shouldn't be doing it as a wrapper but as part of the usual k6 flow and when we load the script to analyze it.

You know, I'm not against it. And I think @pablochacin is going to prioritize it now that the previous part is done.

pablochacin: @codebien  Indeed. But there was also the proposal from @mstoykov sketched in https://github.com/grafana/k6/pull/4866.  Maybe we should create a design document to discuss alternatives.

mstoykov: I do not think that we should be listing every command here. There is much greater chance we will forget something and break it, than anything else IMO. 

We do not add commands ... commonly enough for this to be a case of "this will happen next release". Actually apart from the `k6 cloud run` and co. addition I do not know when we last did this.

codebien: >Maybe we should create a design document to discuss alternatives

@pablochacin Do you have the capacity to lead it? It would be nice if so.

joanlopez (approved):

PR #5082, opened 2025-08-22
Improve the communication for automatic extension resolution

codebien:
## What?

1. Now that automatic extension resolution is the default doesn't make sense to repeat it for each logline, this should be obvious as it is the default.
2. The current waiting time on a blinking prompt is not optimal to me, it doesn't seem obvious to me if I should wait or the program got stuck so I clarified this aspect in the message.

## Checklist

### DIFF
--- a/internal/cmd/launcher.go
+++ b/internal/cmd/launcher.go
@@ -89,7 +89,7 @@ func (l *launcher) launch(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		l.gs.Logger.
 			WithError(err).
-			Error("Automatic extension resolution is enabled but it failed to analyze the dependencies." +
+			Error("Failed to analyze the extensions required." +
 				" Please, make sure to report this issue by opening a bug report.")
 		return err
 	}
@@ -103,8 +103,8 @@ func (l *launcher) launch(cmd *cobra.Command, args []string) error {
 
 	l.gs.Logger.
 		WithField("deps", deps).
-		Info("Automatic extension resolution is enabled. The current k6 binary doesn't satisfy all dependencies," +
-			" it's required to provision a custom binary.")
+		Info("Provisioning a custom binary as the current doesn't satisfy all dependencies," +
+			" may take a few seconds as it needs to download the new one.")
 
 	customBinary, err := l.provisioner.provision(deps)
 	if err != nil {


internal/cmd/launcher.go
```diff
@@ -103,8 +103,8 @@ func (l *launcher) launch(cmd *cobra.Command, args []string) error {
 
 	l.gs.Logger.
 		WithField("deps", deps).
-		Info("Automatic extension resolution is enabled. The current k6 binary doesn't satisfy all dependencies," +
-			" it's required to provision a custom binary.")
+		Info("Provisioning a custom binary as the current doesn't satisfy all dependencies," +
```
joanlopez, line 106: Nit;

```suggestion
		Info("Provisioning a custom binary because the current one doesn't satisfy all the dependencies," +
```

I think it's slightly more precise and clear

joanlopez: Overall, I'm wondering whether we should use either `extensions` or `dependencies`, and if we should be consistent with that use in the two messages modified in this PR.

internal/cmd/launcher.go
```diff
@@ -103,8 +103,8 @@ func (l *launcher) launch(cmd *cobra.Command, args []string) error {
...
+			" may take a few seconds as it needs to download the new one.")
```
mstoykov, line 107:
```suggestion
		Info("Downloading a binary as the current isn't build with all required extensions," +
			" may take a few seconds.")
```

The current listing of `deps` also IMO needs help.

internal/cmd/launcher.go
mstoykov: Some thoughts:
1. I do think we should use `extensions` instead of `dependancies` as the secodn can and likely will be misunderstood as `./somelib.js` Or even worse something else that needs to be download such as chromium or a c library or whatever
2. I do not like the wordign of "Provisioning" - we will be downloading a binary. I think we should jsut use download instead of the internal term.
3. "custom binary" also makes it seems like we will make this specifically for the user, but it isn't.

I have made proposal below

internal/cmd/launcher.go
```diff
@@ -103,8 +103,8 @@ func (l *launcher) launch(cmd *cobra.Command, args []string) error {
 
 	l.gs.Logger.
 		WithField("deps", deps).
```
mstoykov, line 105:
```suggestion
		WithField("extensions", deps).
```

internal/cmd/launcher.go
mstoykov: I guess after the discussion now nad looking more closely at some of the code:
1. we do not always download, sometimes we use a local copy
2. we do not log anything in the above cases at any log level

I do reall think it will be nice to one differntiate between those two case and potentially list both the binary link and potentially the binary path on the fs. Maybe not at Info but debug level

codebien: >Overall, I'm wondering whether we should use either extensions or dependencies

I guess the word `dependencies` here has been used because there is k6 itself to be resolved which is not an extensions.

I will look deeper into a better proposal.

codebien: >I do not like the wordign of "Provisioning" 

@mstoykov same for provisioning. I guess it used to collect the two operations (build + download).

codebien:
After various iteration on this area, this pull request is now outdated. We should tackle https://github.com/grafana/k6/issues/5192 in a new pull request.

PR #5084, opened 2025-08-22
Fix exec context adoption

ankur22:
## What?

Having looked at how Playwright works through adoption, I've realised that it only adopts when the handle is from the current frame, and when the handle is in the utility exec context. I've made the change to ensure that adoption only occurs when the handle is part of the current frame's utility exec context.

## Why?

While working on https://github.com/grafana/k6/issues/5032, the current implementation is causing an issue whereby it is trying to adopt a handle from a nested frame to the page's frame (different origin), which isn't possible and is preventing APIs from working when using `locators` in `iframe`s.

## Checklist

### DIFF
--- a/internal/js/modules/k6/browser/common/frame.go
+++ b/internal/js/modules/k6/browser/common/frame.go
@@ -533,16 +533,20 @@ func (f *Frame) waitForSelector(selector string, opts *FrameWaitForSelectorOptio
 	f.executionContextMu.RLock()
 	defer f.executionContextMu.RUnlock()
 
-	ec := f.executionContexts[mainWorld]
-	if ec == nil {
-		return nil, fmt.Errorf("waiting for selector %q: execution context %q not found", selector, mainWorld)
-	}
+	uec := f.executionContexts[utilityWorld]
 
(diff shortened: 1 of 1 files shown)


inancgumus (approved):
🤞

codebien (changes requested):
>Adoption only works when adopting between mainWorld and utilityWorld from within the same frame. Each frame has these two execution contexts. We cannot adopt elements/handles from one frame to another, and there should be no reason to do so.

@ankur22 If I understand correctly, before it was allowed to mix elements and frames which is not the case anymore with the new change. We should guard this logic with a unit test. Can you add a one, please?

ankur22:
> @ankur22 If I understand correctly, before it was allowed to mix elements and frames which is not the case anymore with the new change. We should guard this logic with a unit test. Can you add a one, please?

I'd love to be able to unit test this, but I think it would require a lot of refactoring. The changes in another PR (https://github.com/grafana/k6/pull/5075) require the fix in this PR, and there is an integration test which will fail due to any regression.

codebien (approved):

PR #5103, opened 2025-08-26
h2c is here

JordanHoggart:
Ignore the Dockerfile, I'm not making use of next but I'm going to. 
xk6 is an easy build option even if it looks a bit weird there, but needed for plugins which we need (statsd output for DD).
I'm also likely going to bake in some other testing tools all in one img, just to have them.

This adds a condition to check the url, and if its http then we swap out 
```
state.Transport = sharedH2CTransport
```

Very simple and dumb way of doing this. It works. Breaks h1 only, which for our tests we dont really care about right now, we shouldn't be trying to hit at scale with h1.

Fiddles some metrics since I want to see when we use h2c vs h2 on the graphs

Fiddles in some log lines just to sanity check I wasn't init a bunch of extra 

image

image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,37 +1,31 @@
-FROM --platform=$BUILDPLATFORM golang:1.24-alpine3.22 as builder
-WORKDIR $GOPATH/src/go.k6.io/k6
-COPY . .
-ARG TARGETOS TARGETARCH
-RUN apk --no-cache add git=~2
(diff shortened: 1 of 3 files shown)


JordanHoggart:
Welp, meant to PR this into our fork. This is just us hacking in h2c to make use of it for our project, but is only some hacking

Ignore me here! 

P.S. Apologies for any ping from this

PR #5105, opened 2025-08-27
Implement the `frame.getBy*` methods

### LINKED ISSUE #5037: Implement `frame.getBy*`
# What

We want to implement `frame.getBy*` APIs. The main reason is to allow for easier usage for existing users who work with the `frame` class to be able to still work with the `getBy*` APIs. This scenario will come up when working with `iframes`. We want to implement:

* `frame.getByRole`
* `frame.getByAltText`
* `frame.getByLabel`
* `frame.getByPlaceholder`
* `frame.getByTitle`
* `frame.getByTestId`
* `frame.getByText`

# Why

This will help make migrating from PW to k6 a little easier. When working with PW's code gen tool in VSCode or working with MCP/AI, it will work with the latest PW APIs which k6 currently lacks. By implementing this, we will make the migration process simpler.

# How

Take a look at how we already share API code between `page` and `frame`. I'm hoping it will be the same when adding these APIs in.

Playwright API docs: (start here) https://playwright.dev/docs/api/class-frame#frame-get-by-alt-text
Playwright code: (start here) https://github.com/microsoft/playwright/blob/6626bba937af513cbcf05d5e164aa20b6d390ea2/packages/playwright-core/src/client/frame.ts#L328

joanlopez:
## What?

It adds the mapping for all the `getBy*` methods for the `frame` API.

## Why?

Because we have already implemented all those methods, and we want to expose them as a way to increase our compatibility with Playwright, as described in #5037.

## Checklist

## Related PR(s)/Issue(s)

Closes  #5037

### DIFF
--- a/internal/js/modules/k6/browser/tests/get_by_test.go
+++ b/internal/js/modules/k6/browser/tests/get_by_test.go
@@ -440,12 +440,9 @@ func TestGetByRoleSuccess(t *testing.T) {
 				)
 				require.NoError(t, err)
 
-				getByRoleImplementations := map[string]interface {
+				getByRoleImplementations := getByImplementationsOf[interface {
 					GetByRole(role string, opts *common.GetByRoleOptions) *common.Locator
-				}{
-					"page":  p,
-					"frame": p.MainFrame(),
-				}
+				}](p)
 
 				for implName, impl := range getByRoleImplementations {
 					t.Run(implName, func(t *testing.T) { //nolint:paralleltest
@@ -577,12 +574,9 @@ func TestGetByRoleSuccess(t *testing.T) {
 				)
 				require.NoError(t, err)
 
-				getByRoleImplementations := map[string]interface {
+				getByRoleImplementations := getByImplementationsOf[interface {
 					GetByRole(role string, opts *common.GetByRoleOptions) *common.Locator
-				}{
-					"page":  p,
-					"frame": p.MainFrame(),
-				}
+				}](p)
 
 				for implName, impl := range getByRoleImplementations {
 					t.Run(implName, func(t *testing.T) { //nolint:paralleltest
@@ -745,12 +739,9 @@ func TestGetByRoleSuccess(t *testing.T) {
 				)
 				require.NoError(t, err)
 
-				getByRoleImplementations := map[string]interface {
+				getByRoleImplementations := getByImplementationsOf[interface {
 					GetByRole(role string, opts *common.GetByRoleOptions) *common.Locator
-				}{
-					"page":  p,
-					"frame": p.MainFrame(),
-				}
+				}](p)
 
 				for implName, impl := range getByRoleImplementations {
 					t.Run(implName, func(t *testing.T) { //nolint:paralleltest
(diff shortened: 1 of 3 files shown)


ankur22 (dismissed):
Woohoo! Nice and easy 😄 

I think we should test the change still, even though most of the implementation is already covered in the [get_by_test.go](https://github.com/grafana/k6/blob/master/internal/js/modules/k6/browser/tests/get_by_test.go) file.

It could just be a subset of those tests translated to work with `frame` instead of `page`, so for each it could just be:

- a happy path test that works with the required fields;
- a happy path test that works with an option;
- a failing test to ensure we are testing `null` values for the required fields.

It might just be a case of copying the whole file and using `frame` instead of `page`.

Happy for you to merge this PR in and work on the test in another PR if you wish.

joanlopez:
> Woohoo! Nice and easy 😄
> 
> ...
> ...

Thanks @ankur22! 🫶 

-- 
What do you think about what I did in https://github.com/grafana/k6/pull/5105/commits/bcb22e6f8a571bb594ea113e21b9b0caebc4e21a?

What I like about this approach is that hopefully later it will be pretty easy to adjust it to https://github.com/grafana/k6/pull/5106, with an empty or equivalent locator that make existing test cases still applicable.

Beyond that, the only other approach I could think of is directly testing the JS surface, but I see we don't have tests like that neither for `page.getBy*`, so I'd consider it out of the scope of this but rather a nice to have, eventually.

internal/js/modules/k6/browser/tests/get_by_test.go
```diff
@@ -440,38 +429,49 @@ func TestGetByRoleSuccess(t *testing.T) {
...
-					require.Equal(t, tt.expectedText, text)
+				getByRoleImplementations := map[string]interface {
+					GetByRole(role string, opts *common.GetByRoleOptions) *common.Locator
+				}{
+					"page":  p,
+					"frame": p.MainFrame(),
+				}
+
+				for implName, impl := range getByRoleImplementations {
+					t.Run(implName, func(t *testing.T) { //nolint:paralleltest
```
joanlopez, line 448: I'm intentionally not calling `t.Parallel()`, as I think others suggested to not use the same browser instance for parallel tests (the reason why I refactored the `TestGetByRole*` to follow the same pattern as the other tests in this file).

Instead, I only use `t.Run` as a syntatic way of running the same test code that calls the function and does some assertions over both implementations, on the same instance, which must be fine, but not in parallel.

ankur22 (dismissed):
Thanks for the changes! I like how you incorporated the `frame` into the existing tests 👍  LGTM 🚀

internal/js/modules/k6/browser/tests/get_by_test.go
```diff
@@ -440,38 +429,49 @@ func TestGetByRoleSuccess(t *testing.T) {
...
-				if tt.expectedText != "" {
-					text, _, err := l.TextContent(sobek.Undefined())
-					require.NoError(t, err)
-					require.Equal(t, tt.expectedText, text)
+				getByRoleImplementations := map[string]interface {
+					GetByRole(role string, opts *common.GetByRoleOptions) *common.Locator
+				}{
+					"page":  p,
+					"frame": p.MainFrame(),
+				}
```
AgnesToulet: What about putting these lines in a function? I feel like it's repeated a lot in this test file (at least for the GetByRole function which has a lot of tests).

AgnesToulet (dismissed):
LGTM!

internal/js/modules/k6/browser/tests/get_by_test.go
joanlopez: Right, good idea! What do you think about https://github.com/grafana/k6/pull/5105/commits/353a1973d20c34d6e9ccfc2234ff565e2829c951?

Issue #5110, opened 2025-08-27
`click` panics after recent CORS fix


ankur22:
### Brief summary

When using `click` it now panics after working on the latest master (43ed87b251) after the CORS fix (https://github.com/grafana/k6/pull/5071). The stack trace is here:

```log
process_error: exit status 2, stderr_error: k6 process was unexpectedly aborted, process_exit_code: 2, process_state: exit status 2, session_id: uUbwPJqHSDa5Cve0qWkhN2up7cKguohn, panic_text:

panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0x12d5170]

goroutine 1207580 [running]:
go.k6.io/k6/internal/js/modules/k6/browser/common.(*ElementHandle).clickablePoint(0xc00e87bc20?)
go.k6.io/k6/internal/js/modules/k6/browser/common/element_handle.go:202 +0x30
go.k6.io/k6/internal/js/modules/k6/browser/common.(*ElementHandle).newPointerAction.func1({0x2074b38, 0xc0130a77a0}, 0x0)
go.k6.io/k6/internal/js/modules/k6/browser/common/element_handle.go:1619 +0x573
go.k6.io/k6/internal/js/modules/k6/browser/common.retryPointerAction({0x2074b38, 0xc0130a77a0}, 0xc013de3ae0, 0xc00f3a3080)
go.k6.io/k6/internal/js/modules/k6/browser/common/element_handle.go:1677 +0x3c
go.k6.io/k6/internal/js/modules/k6/browser/common.(*ElementHandle).newPointerAction.func2({0x2074b38, 0xc0130a77a0}, 0xc00e4d7490, 0xc00e4d7570)
go.k6.io/k6/internal/js/modules/k6/browser/common/element_handle.go:1659 +0x53
go.k6.io/k6/internal/js/modules/k6/browser/common.(*Frame).click.(*Frame).newPointerAction.func2({0x2074b38, 0xc0130a77a0}, 0xc00e4d7490, 0xc00e4d7570)
go.k6.io/k6/internal/js/modules/k6/browser/common/frame.go:2337 +0x114
created by go.k6.io/k6/internal/js/modules/k6/browser/common.call in goroutine 1207579
go.k6.io/k6/internal/js/modules/k6/browser/common/helpers.go:105 +0x14c
```

### k6 version

43ed87b251

### OS

NA

### Steps to reproduce the problem

[...]

PR #5111, opened 2025-08-27
Fix actionability retry

ankur22:
## What

This fixes an issue that was long standing but was made evident after a recent fix for CORS and `click` (https://github.com/grafana/k6/pull/5071). It does three things:

1. Calling `BoundingBox` in `ElementHandle` now returns an error to the caller. It wasn't obvious that the bounding box could be `nil` which caused the NPD in the first place. There is a valid reason for the bounding box being `nil` when returned back to the caller if it was called in a script, which is why we keep that behaviour but move it to the mapping layer.
2. We automatically retry when chrome returns us an error indicating that the element is not visible and therefore it can't retrieve data on it (such as the bounding box). The retry will keep retrying until it hits the default timeout or a timeout set by the user -- this now matches the expected behaviour of `locator` based APIs.
3. In the injected script, it now checks the `stable` state first before trying to continue onto trying to see if the element is `visible` and `enabled`. This ensures that the stability counter is reset back to 0 as soon as the stability check fails, which was being skipped previously -- this is also how Playwright handles the same situation.

## Why

This fixes an NPD. The NPD was due to a race condition:

1. A `locator.click` is called on an element;
2. The element is initially visible, and therefore we can [focus](https://github.com/grafana/k6/blob/47514286136f67e9c297bdddf92317e818677587/internal/js/modules/k6/browser/common/element_handle.go#L1574) on it and all the actionability checks [pass](https://github.com/grafana/k6/blob/47514286136f67e9c297bdddf92317e818677587/internal/js/modules/k6/browser/common/element_handle.go#L1585);
3. However, in between the previous checks and getting the rect for the element, the element is hidden or detached which then [returns `nil`](https://github.com/grafana/k6/blob/47514286136f67e9c297bdddf92317e818677587/internal/js/modules/k6/browser/common/element_handle.go#L60) when getting the bounding box;
4. The error is [ignored](https://github.com/grafana/k6/blob/47514286136f67e9c297bdddf92317e818677587/internal/js/modules/k6/browser/common/element_handle.go#L766-L769) since the behaviour is valid (only when `boundingBox` is called from the user script directly) since it means that the element is hidden or detached;
5. A [NPD panic](https://github.com/grafana/k6/blob/47514286136f67e9c297bdddf92317e818677587/internal/js/modules/k6/browser/common/element_handle.go#L202) occurs since I didn't expect that behaviour.

## Test

Reproducing the NPD panic is tricky. I tried with the following test which helped identify where the possible race could occur. This test is likely to cause an error to do with the visibility of the element being the issue. It is still helpful to validate that we do indeed correctly wait for stability and retry on visibility errors.

You can test the change with this:

```js
import { browser } from "k6/browser";

export const options = {
    scenarios: {
      ui: {
        options: {
          browser: {
            type: 'chromium',
          },
        },
      },
    },
  }

  // We shouldn't see any console messages from chrome being written to the terminal.
  page.on('console', (msg) => {
    // Should assert here and fail the iteration.
    console.log(`PAGE LOG: ${msg.text()}`);
  });

  page.setContent(`
<!doctype html>
<html>
<head>
  <meta charset="utf-8">
  <title>RAF-Toggling Counter Button</title>
</head>
<body>
  <p>Count: <span id="value">0</span></p>

  <button id="incBtn" type="button">+1</button>

  <script>
    let count = 0;
    const valueEl = document.getElementById('value');
    const incBtn = document.getElementById('incBtn');

    incBtn.addEventListener('click', function () {
      count += 1;
      valueEl.textContent = count;
      console.log('click', count);
    });

    function loop() {
      // Flip visibility every animation frame (~60 times per second)
      incBtn.hidden = !incBtn.hidden;
      requestAnimationFrame(loop);
    }

    requestAnimationFrame(loop);
  </script>
</body>
</html>
  `)

  // This should throw a timeout error.
  for (let i = 0; i < 10; i++) {
    await page.locator('button').click();
  }

  // Should assert here to make sure that the counter hasn't increased.

  await page.close();
}
```

## Checklist

## Related PR(s)/Issue(s)

https://github.com/grafana/k6/issues/5110

### DIFF
--- a/internal/js/modules/k6/browser/browser/element_handle_mapping.go
+++ b/internal/js/modules/k6/browser/browser/element_handle_mapping.go
@@ -2,6 +2,7 @@ package browser
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/grafana/sobek"
 
@@ -15,7 +16,13 @@ func mapElementHandle(vu moduleVU, eh *common.ElementHandle) mapping { //nolint:
 	maps := mapping{
 		"boundingBox": func() *sobek.Promise {
 			return k6ext.Promise(vu.Context(), func() (any, error) {
-				return eh.BoundingBox(), nil
+				box, err := eh.BoundingBox()
+				// We want to avoid errors when an element is not visible and instead
+				// opt to return a nil rectangle -- this matches Playwright's behaviour.
+				if err != nil && strings.Contains(err.Error(), "check if element is visible") {
+					return nil, nil
+				}
+				return box, err
 			})
 		},
 		"check": func(opts sobek.Value) (*sobek.Promise, error) {
--- a/internal/js/modules/k6/browser/browser/element_handle_mapping.go
+++ b/internal/js/modules/k6/browser/browser/element_handle_mapping.go
@@ -1,8 +1,8 @@
 package browser
 
 import (
+	"errors"
 	"fmt"
-	"strings"
 
 	"github.com/grafana/sobek"
 
@@ -19,7 +19,7 @@ func mapElementHandle(vu moduleVU, eh *common.ElementHandle) mapping { //nolint:
 				box, err := eh.BoundingBox()
 				// We want to avoid errors when an element is not visible and instead
 				// opt to return a nil rectangle -- this matches Playwright's behaviour.
-				if err != nil && strings.Contains(err.Error(), "check if element is visible") {
+				if errors.Is(err, common.ErrElementNotVisible) {
 					return nil, nil
 				}
 				return box, err
--- a/internal/js/modules/k6/browser/common/element_handle.go
+++ b/internal/js/modules/k6/browser/common/element_handle.go
@@ -68,7 +68,7 @@ func (h *ElementHandle) boundingBox() (*Rect, error) {
 
 	position, err := h.frame.position()
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("getting position of parent frame: %w", err)
 	}
 
 	return &Rect{X: x + position.X, Y: y + position.Y, Width: width, Height: height}, nil
@@ -94,7 +94,11 @@ func (h *ElementHandle) translatePointToPage(apiCtx context.Context, point Posit
 		return Position{}, err
 	}
 
-	box := el.BoundingBox()
+	box, err := el.BoundingBox()
+	if err != nil {
+		return Position{}, fmt.Errorf("getting bounding box of parent frame: %w", err)
+	}
+
 	if box.contains(point) {
 		h.logger.Debugf("ElementHandle:translatePointToPage", "point is already in the page")
 		return point, nil
@@ -197,9 +201,12 @@ func (h *ElementHandle) click(p *Position, opts *MouseClickOptions) error {
 // This will get the clickable point of the element relative to the page even
 // when the element is in an iframe. In some cases this isn't the case and we
 // need to translate the point to the page.
-func (h *ElementHandle) clickablePoint() *Position {
-	r := h.BoundingBox()
-	return &Position{X: r.X + r.Width/2, Y: r.Y + r.Height/2}
+func (h *ElementHandle) clickablePoint() (*Position, error) {
+	r, err := h.BoundingBox()
+	if err != nil {
+		return nil, err
+	}
+	return &Position{X: r.X + r.Width/2, Y: r.Y + r.Height/2}, nil
 }
 
 func (h *ElementHandle) dblclick(p *Position, opts *MouseClickOptions) error {
@@ -388,7 +395,11 @@ func (h *ElementHandle) offsetPosition(apiCtx context.Context, offset *Position)
 		return nil, fmt.Errorf("converting result (%v of type %t) to border: %w", result, result, err)
 	}
 
-	box := h.BoundingBox()
+	box, err := h.BoundingBox()
+	if err != nil {
+		return nil, err
+	}
+
 	if box == nil || (border.Left == 0 && border.Top == 0) {
 		return nil, errorFromDOMError("error:notvisible")
 	}
@@ -762,12 +773,15 @@ func (h *ElementHandle) AsElement() *ElementHandle {
 }
 
 // BoundingBox returns this element's bounding box.
-func (h *ElementHandle) BoundingBox() *Rect {
+func (h *ElementHandle) BoundingBox() (*Rect, error) {
 	bbox, err := h.boundingBox()
-	if err != nil {
-		return nil // Don't panic here, just return nil
+	if err != nil && strings.Contains(err.Error(), "Could not compute box model") {
+		return nil, fmt.Errorf("check if element is visible: %w", err)
 	}
-	return bbox
+	if err != nil {
+		return nil, fmt.Errorf("getting bounding box: %w", err)
+	}
+	return bbox, nil
 }
 
 // Click scrolls element into view and clicks in the center of the element
@@ -1616,7 +1630,10 @@ func (h *ElementHandle) newPointerAction(
 				return nil, fmt.Errorf("getting element position: %w", err)
 			}
 		} else {
-			p = h.clickablePoint()
+			p, err = h.clickablePoint()
+			if err != nil {
+				return nil, fmt.Errorf("getting clickable point: %w", err)
+			}
 		}
 
 		// Further translation of the point might be necessary if the point
--- a/internal/js/modules/k6/browser/common/element_handle.go
+++ b/internal/js/modules/k6/browser/common/element_handle.go
@@ -20,6 +20,12 @@ import (
 	k6common "go.k6.io/k6/js/common"
 )
 
+// Common error types for element visibility.
+var (
+	// ErrElementNotVisible is returned when an element is not visible for an operation.
+	ErrElementNotVisible = errors.New("element is not visible")
+)
+
 const (
 	resultDone       = "done"
 	resultNeedsInput = "needsinput"
@@ -776,7 +782,7 @@ func (h *ElementHandle) AsElement() *ElementHandle {
 func (h *ElementHandle) BoundingBox() (*Rect, error) {
 	bbox, err := h.boundingBox()
 	if err != nil && strings.Contains(err.Error(), "Could not compute box model") {
-		return nil, fmt.Errorf("check if element is visible: %w", err)
+		return nil, fmt.Errorf("%w: %w", ErrElementNotVisible, err)
 	}
 	if err != nil {
 		return nil, fmt.Errorf("getting bounding box: %w", err)
@@ -1709,7 +1715,7 @@ func retryPointerAction(
 			}
 		}
 
-		if !strings.Contains(err.Error(), "check if element is visible") &&
+		if !errors.Is(err, ErrElementNotVisible) &&
 			!strings.Contains(err.Error(), "error:notvisible") {
 			return res, err
 		}
--- a/internal/js/modules/k6/browser/common/element_handle.go
+++ b/internal/js/modules/k6/browser/common/element_handle.go
@@ -1690,24 +1690,38 @@ func (h *ElementHandle) newPointerAction(
 func retryPointerAction(
 	apiCtx context.Context, fn retryablePointerActionFunc, opts *ElementHandleBasePointerOptions,
 ) (res any, err error) {
-	// try the default scrolling
-	if res, err = fn(apiCtx, nil); opts.Force || err == nil {
-		return res, err
-	}
-	// try with different scrolling options
-	for _, p := range []ScrollPosition{
-		ScrollPositionStart,
-		ScrollPositionCenter,
-		ScrollPositionEnd,
-		ScrollPositionNearest,
-	} {
-		s := ScrollIntoViewOptions{Block: p, Inline: p}
-		if res, err = fn(apiCtx, &s); err == nil {
-			break
+	for {
+		res, err = fn(apiCtx, nil)
+		if opts.Force || err == nil {
+			return res, err
+		}
+
+		// try with different scrolling options
+		for _, p := range []ScrollPosition{
+			ScrollPositionStart,
+			ScrollPositionCenter,
+			ScrollPositionEnd,
+			ScrollPositionNearest,
+		} {
+			s := ScrollIntoViewOptions{Block: p, Inline: p}
+			if res, err = fn(apiCtx, &s); err == nil {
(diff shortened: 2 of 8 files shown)


internal/js/modules/k6/browser/browser/element_handle_mapping.go
```diff
@@ -15,7 +16,13 @@ func mapElementHandle(vu moduleVU, eh *common.ElementHandle) mapping { //nolint:
 	maps := mapping{
 		"boundingBox": func() *sobek.Promise {
 			return k6ext.Promise(vu.Context(), func() (any, error) {
-				return eh.BoundingBox(), nil
+				box, err := eh.BoundingBox()
+				// We want to avoid errors when an element is not visible and instead
+				// opt to return a nil rectangle -- this matches Playwright's behaviour.
+				if err != nil && strings.Contains(err.Error(), "check if element is visible") {
```
joanlopez: Nit; As this text is something we generate in the inner function, I'd likely prefer to use typed errors instead of string comparison.

ankur22: Yeah, i agree. What's a good way of doing this? I was thinking of creating a new error and then using `error.Is`. Is there any best practices used in k6 already that I can copy?

internal/js/modules/k6/browser/common/element_handle.go
```diff
@@ -1673,24 +1690,38 @@ func (h *ElementHandle) newPointerAction(
...
+			ScrollPositionEnd,
+			ScrollPositionNearest,
+		} {
+			s := ScrollIntoViewOptions{Block: p, Inline: p}
+			if res, err = fn(apiCtx, &s); err == nil {
+				return res, nil
+			}
+		}
+
+		if !strings.Contains(err.Error(), "check if element is visible") &&
```
joanlopez: Same as above

internal/js/modules/k6/browser/common/element_handle.go
```diff
@@ -1673,24 +1690,38 @@ func (h *ElementHandle) newPointerAction(
...
+			return res, err
+		}
+
+		// Wait with timeout or context cancellation
+		select {
+		case <-apiCtx.Done():
+			return nil, apiCtx.Err()
+		case <-time.After(20 * time.Millisecond):
+			// Continue retrying after delay
+		}
```
joanlopez, line 1729: Is `opts.Timeout` somehow considered here (e.g. implicitly as part of `apiCtx.Done() handling)? Otherwise, I think we might need to handle it explicitly 🤔

internal/js/modules/k6/browser/browser/element_handle_mapping.go
joanlopez: I have nothing k6-specific in particular in mind, but I think standard Go practices, as you suggested, should be enough and fine here 👍🏻

internal/js/modules/k6/browser/common/js/injected_script.js
```diff
@@ -2448,18 +2448,7 @@ class InjectedScript {
...
-          if (typeof result !== "boolean") {
-            return result;
-          }
-          if (!result) {
-            return continuePolling;
-          }
-          continue;
-        }
-
+      if (states.includes("stable")) {
```
joanlopez, line 2450: I'm not sure I can understand this, but as long as it's what Playwright does, I guess it's fine! 👍🏻

internal/js/modules/k6/browser/tests/element_handle_test.go
```diff
@@ -65,7 +66,32 @@ func TestElementHandleBoundingBoxInvisibleElement(t *testing.T) {
 	require.NoError(t, err)
 	element, err := p.Query("div")
 	require.NoError(t, err)
-	require.Nil(t, element.BoundingBox())
+
+	_, err = element.BoundingBox()
+	require.ErrorContains(t, err, "check if element is visible: getting bounding box model of DOM node: Could not compute box model.")
```
joanlopez: I think asserting the return value is `nil` is still a valid assertion, to preserve the same behavior in such case.

ankur22: Yep, agree. There's now `TestElementHandleBoundingBoxInvisibleElementWithMapper` that tests that, and preserves that behaviour.

joanlopez (commented):
I left a few comments, but no one is really a red flag/blocking.

What concerns me, a bit more, is that, as discussed in chat, I cannot get a reproducible behavior from the example script in the description, nor to reproduce the panic and see it fixed now.

Although it's true that the NPD was clearly caused because the box was `nil`, which is something we handle now 👍🏻

internal/js/modules/k6/browser/common/js/injected_script.js
```diff
@@ -2461,13 +2460,6 @@ class InjectedScript {
           return continuePolling;
         }
 
-        // Drop frames that are shorter than 16ms - WebKit Win bug.
-        const time = performance.now();
-        if (this._stableRafCount > 1 && time - lastTime < 15) {
-          return continuePolling;
-        }
-        lastTime = time;
-
```
ankur22, line 2481: This was causing a strange issue. With the test provided in the description of this PR, what was happening when skipping short frames was:

1. button is hidden
2. short frame skipped
3. button is visible
4. check stability -- rect is same
1. button is hidden
2. short frame skipped
3. button is visible
4. check stability -- rect is same
5. ...
6. element is stable! 🤯 

So the stability check is ONLY done after the button is visible.

After removing this short frame check we now have:

1. element is hidden
2. check stability -- rect is not the same
3. element is visible
4. check stability -- rect is not the same
5. element is hidden
6. check stability -- rect is not the same
8. element is visible
9. check stability -- rect is not the same
10. ...
11. timeout! ✅ 

So the stability check is done after the button is hidden/visible.

internal/js/modules/k6/browser/common/element_handle.go
ankur22: Yeah, the context with a timeout is used here. The timeout context is setup [here](https://github.com/grafana/k6/blob/dfedf4beeed4d344306598c8374e4c0641270c12/internal/js/modules/k6/browser/common/helpers.go#L100-L103).

internal/js/modules/k6/browser/common/js/injected_script.js
ankur22: Yeah 😅 the logic is a bit strange. So now we're:

1. Check that the rect is stable first. This is done by comparing the rect and previous rect 10 times. If it's the same then it's regarded as stable. If the rect changes since previous rect, then the counter is reset back to 0.
2. Once it's deemed stable, we do the other state checks.

Before it was checking the other states first, and then checking stability. This meant that when it came to checking stability it was always in the stable state for the example in this PR since element was always visible, which is incorrect behaviour.

internal/js/modules/k6/browser/browser/element_handle_mapping.go
ankur22: Done in e5ce6ae7cdf2affc08cffe68acf45d5cb14d4278

internal/js/modules/k6/browser/common/element_handle.go
ankur22: Done in e5ce6ae7cdf2affc08cffe68acf45d5cb14d4278

internal/js/modules/k6/browser/common/js/injected_script.js
joanlopez: I see there was a comment, pointing to an issue in _"WebKit Win"_. Does that mean Safari on Windows? 
If so, I guess we're fine for long time in future, right? So 👍🏻

internal/js/modules/k6/browser/common/element_handle.go
joanlopez: Ah, okay, so it's managed up to two abstraction layers up, managed by the caller, that's why I wasn't seeing it. I guess it's fine for now! 👍🏻  - Thanks!

joanlopez (approved):

internal/js/modules/k6/browser/browser/element_handle_mapping.go
```diff
@@ -15,7 +16,13 @@ func mapElementHandle(vu moduleVU, eh *common.ElementHandle) mapping { //nolint:
 	maps := mapping{
 		"boundingBox": func() *sobek.Promise {
 			return k6ext.Promise(vu.Context(), func() (any, error) {
-				return eh.BoundingBox(), nil
+				box, err := eh.BoundingBox()
+				// We want to avoid errors when an element is not visible and instead
+				// opt to return a nil rectangle -- this matches Playwright's behaviour.
+				if errors.Is(err, common.ErrElementNotVisible) {
+					return nil, nil
```
inancgumus, line 23: Should we return a `nil Rect`, or `&Rect{}` to reduce possible errors due to NPDs? Having NPDs can be useful for learning about the behavior, though.

internal/js/modules/k6/browser/common/element_handle.go
```diff
@@ -197,9 +207,12 @@ func (h *ElementHandle) click(p *Position, opts *MouseClickOptions) error {
 // This will get the clickable point of the element relative to the page even
 // when the element is in an iframe. In some cases this isn't the case and we
 // need to translate the point to the page.
-func (h *ElementHandle) clickablePoint() *Position {
-	r := h.BoundingBox()
-	return &Position{X: r.X + r.Width/2, Y: r.Y + r.Height/2}
+func (h *ElementHandle) clickablePoint() (*Position, error) {
+	r, err := h.BoundingBox()
+	if err != nil {
+		return nil, err
```
inancgumus, line 213: This could be useful for us to track errors while debugging or on incidents:

```suggestion
		return nil, fmt.Errorf("finding clickable point: %w", err)
```

internal/js/modules/k6/browser/common/element_handle.go
```diff
@@ -388,7 +401,11 @@ func (h *ElementHandle) offsetPosition(apiCtx context.Context, offset *Position)
 		return nil, fmt.Errorf("converting result (%v of type %t) to border: %w", result, result, err)
 	}
 
-	box := h.BoundingBox()
+	box, err := h.BoundingBox()
+	if err != nil {
+		return nil, err
```
inancgumus, line 406:
```suggestion
		return nil, fmt.Errorf("finding offset position: %w", err)
```

internal/js/modules/k6/browser/common/element_handle.go
```diff
@@ -1616,7 +1636,10 @@ func (h *ElementHandle) newPointerAction(
 				return nil, fmt.Errorf("getting element position: %w", err)
 			}
 		} else {
-			p = h.clickablePoint()
+			p, err = h.clickablePoint()
+			if err != nil {
+				return nil, fmt.Errorf("getting clickable point: %w", err)
```
inancgumus, line 1641: I'd make this as follows so that we can trace the error. `clickablePoint()` can determine on its own that it was a clickable point error (see my early comment).

```suggestion
				return nil, fmt.Errorf("pointer action: %w", err)
```

internal/js/modules/k6/browser/common/element_handle.go
```diff
@@ -1673,24 +1696,38 @@ func (h *ElementHandle) newPointerAction(
...
+			return res, err
 		}
-	}
 
-	return res, err
+		// Wait with timeout or context cancellation
+		select {
+		case <-apiCtx.Done():
+			return nil, apiCtx.Err()
+		case <-time.After(20 * time.Millisecond):
```
inancgumus, line 1727: Could 20ms be costly in terms of CPU use and the number of browser module<->Chrome interactions? Do you think we should bump it?

internal/js/modules/k6/browser/common/frame.go
```diff
@@ -416,7 +416,10 @@ func (f *Frame) position() (*Position, error) {
 		return nil, err
 	}
 
-	box := element.BoundingBox()
+	box, err := element.BoundingBox()
+	if err != nil {
+		return nil, err
```
inancgumus, line 421:
```suggestion
		return nil, fmt.Errorf("finding position in frame: %w", err)
```

inancgumus (approved):
Nice 👍 Some suggestions.

internal/js/modules/k6/browser/browser/element_handle_mapping.go
ankur22: Yep, that could be the next step. I didn't want to make this any bigger than it already is. I'll add this to a refactor issue as i'd like this merged in ASAP.

New issue: https://github.com/grafana/k6/issues/5123

internal/js/modules/k6/browser/common/element_handle.go
ankur22: For now i'll add it to a new issue: https://github.com/grafana/k6/issues/5123

internal/js/modules/k6/browser/common/element_handle.go
ankur22: For now i'll add it to a new issue: https://github.com/grafana/k6/issues/5123

internal/js/modules/k6/browser/common/element_handle.go
ankur22: For now i'll add it to a new issue: https://github.com/grafana/k6/issues/5123

internal/js/modules/k6/browser/common/element_handle.go
ankur22: In Playwright they use a backoff retry. It starts at 20ms, and then increases. I don't think we should be hitting this that often. When i ran tests, it would retry, and then usually wait on the actionability checks -- i.e. it wouldn't really hit that many retries. I think rather than increasing this a backoff retry implementation would be better. Not for this PR though.

internal/js/modules/k6/browser/common/frame.go
ankur22: For now i'll add it to a new issue: https://github.com/grafana/k6/issues/5123

internal/js/modules/k6/browser/common/element_handle.go
inancgumus: Yes, a backoff retry sounds better 👍

ankur22: Cool, added it to https://github.com/grafana/k6/issues/5123 to explore that option 😃

PR #5113, opened 2025-08-29
Add client-side request compression

tank-500m:
## What?

Enable client-side gRPC request compression in k6.
Support opt-in codecs: gzip, zstd, and snappy.

## Why?

Bandwidth & cost savings: large protobuf messages and high RPS scenarios benefit from request compression.

## Checklist


ankur22:
Thanks @tank-500m for the PR! We are taking a look and considering our options. We're a little unsure about the additional dependencies that are being added, and we need to think about how the new option impacts the overall API, and whether we have something else within k6 that does something similar already (enabling compression). We will get back to you with next steps soon.

tank-500m:
@ankur22 
Thanks for the thoughtful review!

If we drop `mostynb/go-grpc-compression`, we’ll need to add a small amount of internal code to support custom compressors, because `grpc.WithDefaultCallOptions(grpc.UseCompressor("..."))` only works for names that have been registered with gRPC’s encoding registry. The third-party package was just a thin adapter that registers zstd/snappy for us.

If remove `mostynb/go-grpc-compression`, the new direction is this:
zstd: use `github.com/klauspost/compress/zstd` to implement a tiny encoding.Compressor adapter (registering name zstd). It’s a well-maintained, battle-tested codec.

snappy (optional): similarly, we can wire up github.com/golang/snappy.

gzip: continues to use the built-in gRPC gzip support.

This keeps the external surface area small while still giving users a practical compression choice (zstd is the main ask for high-throughput gRPC). Let me know if you prefer zstd-only to start; I can push an update that adds an internal zstd compressor + registry and removes the mostynb dependency.

ankur22:
> Let me know if you prefer zstd-only to start; I can push an update that adds an internal zstd compressor + registry and removes the mostynb dependency.

@tank-500m I think this is a good step forward. If you could implement the changes to allow for this single compression (zstd) without adding any external dependencies, please 🙇

tank-500m:
@ankur22 
Sorry for the delay here. 
Per the review feedback about extra deps, I’ve removed mostynb/go-grpc-compression and switched to an in-tree compressor that uses github.com/klauspost/compress/zstd. The compressor is registered via gRPC’s encoding registry so grpc.UseCompressor("zstd") works without third-party adapters.

go.mod
AgnesToulet, line 1: I think you forgot to update this file to remove the extra deps.

tank-500m: sorry i will fix soon

internal/js/modules/k6/grpc/client.go
```diff
@@ -264,6 +265,14 @@ func (c *Client) Connect(addr string, params sobek.Value) (bool, error) {
 		opts = append(opts, grpc.WithAuthority(p.Authority))
 	}
 
+	if p.Compression.Name != "" {
```
AgnesToulet:
```suggestion
	if p.Compression != nil && p.Compression.Name != "" {
```
This is causing panic in the tests.

internal/js/modules/k6/grpc/compression/zstd.go
```diff
@@ -0,0 +1,77 @@
...
+
+const name = "zstd"
+
+type plugin struct{}
+
+//nolint:gochecknoglobals
+var (
+	registerOnce sync.Once
+	encOpts      atomic.Value
+)
```
AgnesToulet: I think this should be moved to the `plugin` structure.

internal/js/modules/k6/grpc/params.go
```diff
@@ -132,19 +134,23 @@ type connectParams struct {
...
-		IsPlaintext:           false,
-		UseReflectionProtocol: false,
-		Timeout:               time.Minute,
-		MaxReceiveSize:        0,
-		MaxSendSize:           0,
-		Authority:             "",
-		ReflectionMetadata:    metadata.New(nil),
-	}
+//nolint:gochecknoglobals
+var defaultConnectParams = &connectParams{
```
AgnesToulet: I think this was better in the function itself rather than in a variable.

internal/js/modules/k6/grpc/compression/registry.go
```diff
@@ -0,0 +1,56 @@
...
+	Name() string
+	EnsureRegistered() error
+	Configure(options map[string]any) error
+	CallOption() grpc.CallOption
+}
+
+var (
+	regMu    sync.RWMutex          //nolint:gochecknoglobals
+	registry = map[string]Plugin{} //nolint:gochecknoglobals
+)
```
AgnesToulet: Maybe this could be put in a `Registry` structure instead of global variables?

AgnesToulet (commented):
Thanks for the updates, left a few more suggestions.

internal/js/modules/k6/grpc/params.go
tank-500m: Added nolint funlen due to over 80.

internal/js/modules/k6/grpc/params.go
```diff
@@ -269,18 +267,36 @@ func parseConnectTLSParam(params *connectParams, v interface{}) error {
...
+		}
+		return &grpccompress.Spec{Name: name}, nil
+	case map[string]any:
+		rawName, ok := x["name"]
+		if !ok {
+			return nil, fmt.Errorf("compression.name is required")
+		}
+		name, ok := rawName.(string)
+		if !ok {
+			return nil, fmt.Errorf("compression.name must be a non-empty string")
```
ankur22, line 285: An empty string seems to be valid, which is the same as "none". Is that intentional? If so, probably worth amending this error to say something like "compression.name must be a string".

internal/js/modules/k6/grpc/compression/zstd.go
```diff
@@ -49,29 +49,35 @@ func (plugin) Configure(m map[string]any) error {
...
-type zstdCompressor struct{}
+type zstdCompressor struct{ p *plugin }
 
-func (zstdCompressor) Name() string { return name }
-func (zstdCompressor) Compress(w io.Writer) (io.WriteCloser, error) {
-	v, ok := encOpts.Load().([]kz.EOption)
-	if !ok || v == nil {
-		return kz.NewWriter(w)
+func (z zstdCompressor) Name() string { return name }
+func (z zstdCompressor) Compress(w io.Writer) (io.WriteCloser, error) {
```
ankur22, line 61: I just want to make sure that this was intentional -- using value type instead of pointer type here.

internal/js/modules/k6/grpc/compression/zstd.go
```diff
@@ -0,0 +1,83 @@
...
+func (z zstdCompressor) Decompress(r io.Reader) (io.Reader, error) {
+	dec, err := kz.NewReader(r, kz.WithDecoderConcurrency(1))
+	if err != nil {
+		return nil, err
+	}
+	return struct{ io.ReadCloser }{dec.IOReadCloser()}, nil // EOF 시 Close()
+}
+
+func init() {
+	_ = NewRegistry().Register(&plugin{})
```
ankur22, line 82: Can you explain why we need this here?

internal/js/modules/k6/grpc/compression/zstd.go
```diff
@@ -0,0 +1,83 @@
...
+func (z zstdCompressor) Name() string { return name }
+func (z zstdCompressor) Compress(w io.Writer) (io.WriteCloser, error) {
+	opts := z.p.opts.Load()
+	if opts == nil {
+		z.p.opts.Store(&[]kz.EOption{
+			kz.WithEncoderLevel(kz.EncoderLevelFromZstd(1)),
+			kz.WithEncoderConcurrency(1),
+		})
+		opts = z.p.opts.Load()
+	}
```
ankur22, line 69: Is this needed? Looks like it's also being done in `EnsureRegistered`. Can we set this up in a single place?

internal/js/modules/k6/grpc/params_test.go
```diff
@@ -206,6 +207,60 @@ func TestConnectParamsAuthority(t *testing.T) {
...
+			ExpectedCompression: &grpccompress.Spec{
+				Name: "zstd",
+			},
+		},
+		{
+			Name: "zstdCustomCompression",
+			JSON: `{compression: {name: "zstd", params: {"level": 1}}}`,
+			ExpectedCompression: &grpccompress.Spec{
+				Name:    "zstd",
+				Options: map[string]interface{}{"level": int64(1)},
```
ankur22, line 245: I believe Sobek will be giving us a `float64` (needs to be double checked). I wonder if we need to do proper type checking when working with the level value.

internal/js/modules/k6/grpc/client.go
```diff
@@ -264,6 +265,15 @@ func (c *Client) Connect(addr string, params sobek.Value) (bool, error) {
 		opts = append(opts, grpc.WithAuthority(p.Authority))
 	}
 
+	if p.Compression != nil && p.Compression.Name != "" {
+		registry := grpccompress.NewRegistry()
```
ankur22, line 269: This is a new registry, where a plugin hasn't been registered, why is that?

internal/js/modules/k6/grpc/compression/zstd.go
```diff
@@ -0,0 +1,83 @@
...
+	return kz.NewWriter(w, *opts...)
+}
+
+func (z zstdCompressor) Decompress(r io.Reader) (io.Reader, error) {
+	dec, err := kz.NewReader(r, kz.WithDecoderConcurrency(1))
+	if err != nil {
+		return nil, err
+	}
+	return struct{ io.ReadCloser }{dec.IOReadCloser()}, nil // EOF 시 Close()
```
ankur22, line 78: I took a look at the gzip encoder, and it got me wondering whether we need to handle io.EOF by closing the readerCloser? WDYT?

Source: https://github.com/grpc/grpc-go/blob/v1.76.0/encoding/gzip/gzip.go#L106-L112

ankur22:
@tank-500m thank you so far for the changes. We'd really like it if you can take it forward to completion and action on the comments 🙏 

We will have to close this if we don't see any action in a weeks time (by 2025-12-04). However, please free to reopen when you have some time 🙇

tank-500m:
@ankur22 
sorry i'm late. I don't have enough time to address this properly right now, so I'll come back to it and open a new PR once I can improve it.

PR #5114, opened 2025-08-29
Add `Locator.Filter`

inancgumus:
## What?

- Implements `Locator.Filter` to get a new locator with a selector that filters sub-elements from a set of elements described by another locator on which the `Filter` method is called.
- Adds a `locator_filter.js` example E2E test.

## Why?

- To give users more flexibility when working with locators.
- To make migrating from PW to k6-browser easier.

## Use

`Filter` matches [Playwright's](https://playwright.dev/docs/locators#filtering-locators) `Filter` method behavior. We currently only support `HasText` and `HasNotText` filters.

Given an HTML:

```html
<ul>
  <li>
    <h3>Product 1</h3>
    <button onclick="console.log('clicked 1')">Add to cart 1</button>
  </li>
  <li>
    <h3>Product 2</h3>
    <button onclick="console.log('clicked 2')">Add to cart 2</button>
  </li>
</ul>
```

Filter the 2nd item using `hasText`:

```js
await p.getByRole('listitem')
  .filter({ hasText: 'Product 2' })
  .first()
```

Filter the 1st item using `hasNotText` and a regex:

```js
await p.getByRole('listitem')
  .filter({ hasNotText: /Product 2/ })
  .first()
```

See [this document](https://playwright.dev/docs/locators#filtering-locators) to find more details.

## Checklist

- [x] Release notes, k6-docs, and TS are being managed by the #5033 issue.

## Related PR(s)/Issue(s)

Updates #5033.

### DIFF
--- a/internal/js/modules/k6/browser/common/locator.go
+++ b/internal/js/modules/k6/browser/common/locator.go
@@ -516,7 +516,7 @@ func (l *Locator) GetByTitle(title string, opts *GetByBaseOptions) *Locator {
 
 // Locator creates and returns a new locator chained/relative to the current locator.
 func (l *Locator) Locator(selector string) *Locator {
-	return NewLocator(l.ctx, l.opts, l.selector+" >> "+selector, l.frame, l.log)
+	return NewLocator(l.ctx, nil, l.selector+" >> "+selector, l.frame, l.log)
 }
 
 // InnerHTML returns the element's inner HTML that matches


internal/js/modules/k6/browser/common/locator.go
```diff
@@ -460,13 +495,13 @@ func (l *Locator) innerText(opts *FrameInnerTextOptions) (string, error) {
 // Last will return the last child of the element matching the locator's
 // selector.
 func (l *Locator) Last() *Locator {
-	return NewLocator(l.ctx, l.selector+" >> nth=-1", l.frame, l.log)
+	return NewLocator(l.ctx, nil, l.selector+" >> nth=-1", l.frame, l.log)
```
inancgumus, line 569: We don't pass the options here, otherwise, we'd duplicate options when iterating on elements. For example:

```javascript
page.getByRole('listitem').filter({ hasText: 'foo' }).last()
```

Would produce:

```bash
internal:role=listitem >> internal:has-text=foo >> nth=42 >> internal:has-text=foo
```

Now, it produces (matching the PW behavior):

```
internal:role=listitem >> internal:has-text=foo >> nth=42
```

internal/js/modules/k6/browser/common/locator.go
```diff
@@ -417,7 +425,7 @@ func (l *Locator) getAttribute(name string, opts *FrameBaseOptions) (string, boo
 
 // Locator creates and returns a new locator chained/relative to the current locator.
 func (l *Locator) Locator(selector string) *Locator {
-	return NewLocator(l.ctx, l.selector+" >> "+selector, l.frame, l.log)
+	return NewLocator(l.ctx, l.opts, l.selector+" >> "+selector, l.frame, l.log)
```
ankur22: Why are we passing `options` from the current locator?

Eventually when `options` can be passed to Locator (this method), then we should pass in the new option and not the current option, right?

e.g.:

At the moment the second `locator` would use the filter from the first `locator`, no?

```js
locator.('div', {hasText: 'hello'}).locator('ul')
```

Maybe option support for this method is coming in another PR?

internal/js/modules/k6/browser/common/locator.go
```diff
@@ -35,9 +36,16 @@ type LocatorOptions struct {
...
 
 // NewLocator creates and returns a new locator.
-func NewLocator(ctx context.Context, selector string, f *Frame, l *log.Logger) *Locator {
+func NewLocator(ctx context.Context, opts *LocatorOptions, selector string, f *Frame, l *log.Logger) *Locator {
+	// We can create a locator from another. We clone the options
+	// to avoid surprising aliasing effects between locators.
+	var copts LocatorOptions
+	if opts != nil {
+		copts = *opts
+	}
```
ankur22, line 48: I'm not sure I understand why this is needed. What is the cloning protecting us against? Is this something we should be doing for other options for other APIs (e.g. options for `page.goto`)?

internal/js/modules/k6/browser/browser/locator_mapping.go
```diff
@@ -222,3 +223,21 @@ func mapLocator(vu moduleVU, lo *common.Locator) mapping {
...
+		return nil
+	}
+
+	var popts common.LocatorOptions
+
+	obj := opts.ToObject(rt)
+	for _, k := range obj.Keys() {
+		switch k {
+		case "hasText":
+			popts.HasText = obj.Get(k).String()
```
ankur22, line 307: Do we need to work with `parseStringOrRegex` since the value of `hasText` can either be a `string` or a `regex`?

Maybe regex support is coming in another PR?

internal/js/modules/k6/browser/common/selectors.go
```diff
@@ -22,6 +22,12 @@ import (
 	"strings"
 )
 
+// TODO: Reimplement this by looking at:
+// https://github.com/microsoft/playwright/blob/
+// 		35cd6fb279ccf4c91eae4b8094697eceb61426cc/
+// 		packages/playwright-core/src/utils/isomorphic/selectorParser.ts#L154
+// Current implementation is a simplified version of the above and cannot handle all edge cases.
+
```
ankur22, line 30: I only think we should reimplement if there is a good use case for it.

I feel that in the browser module lots of things were implemented, but only now we're understanding and using changes made from the initial commit.

Once we understand the use case, then reimplementation will be easier, otherwise we will be doing a lot of work without understanding the problem.

internal/js/modules/k6/browser/common/selectors.go
```diff
@@ -22,6 +22,12 @@ import (
 	"strings"
 )
 
+// TODO: Reimplement this by looking at:
+// https://github.com/microsoft/playwright/blob/
+// 		35cd6fb279ccf4c91eae4b8094697eceb61426cc/
+// 		packages/playwright-core/src/utils/isomorphic/selectorParser.ts#L154
+// Current implementation is a simplified version of the above and cannot handle all edge cases.
```
AgnesToulet, line 29: Do we have an issue about this with the edge cases we don't support?

internal/js/modules/k6/browser/common/selectors.go
```diff
@@ -73,6 +79,12 @@ func (s *Selector) parse() error {
 		var name, body string
 
 		switch {
+		case strings.HasPrefix(part, "internal:has-text="):
+			name = "internal:has-text"
+			body = TrimQuotes(part[eqIndex+1:])
```
ankur22, line 84: I'm not sure why we need `Trim` here and below, since there are no quotes on the value here.

internal/js/modules/k6/browser/tests/locator_test.go
```diff
@@ -947,3 +947,139 @@ func TestActionabilityRetry(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, "0", text)
 }
+
+func TestLocatorFilter(t *testing.T) {
```
ankur22, line 951: Thanks for the test for `page.filter` 👍 

I feel that we should have a test for `frame.filter` too, even though the implementation is currently almost identical.

We should also have tests for `locator.locator` using filter (at least in the first `locator` should contain a `option`, but not the second since `options` aren't supported in `locator.locator`).

internal/js/modules/k6/browser/common/locator.go
```diff
@@ -20,17 +20,41 @@ import (
...
+	// Matches only elements that do not contain the specified text.
+	// String or RegExp. Optional.
+	HasNotText string
+}
+
 // NewLocator creates and returns a new locator.
-func NewLocator(ctx context.Context, selector string, f *Frame, l *log.Logger) *Locator {
+func NewLocator(ctx context.Context, opts *LocatorOptions, selector string, f *Frame, l *log.Logger) *Locator {
+	// We can create a locator from another. We clone the options
+	// to avoid surprising aliasing effects between locators.
```
AgnesToulet, line 44:
```suggestion
func NewLocator(ctx context.Context, opts LocatorOptions, selector string, f *Frame, l *log.Logger) *Locator {
	// We can create a locator from another. We clone the options
	// to avoid surprising aliasing effects between locators.
```
What about removing the pointer then so we are sure to not have these issues? We don't seem to rely on the nil value for anything anyway.

internal/js/modules/k6/browser/common/locator.go
```diff
@@ -355,10 +379,21 @@ func (l *Locator) fill(value string, opts *FrameFillOptions) error {
 	return l.frame.fill(l.selector, value, opts)
 }
 
+// LocatorFilterOptions allows filtering a [Locator] by various criteria.
+// It's similar to [LocatorOptions] but used for filtering existing locators.
+type LocatorFilterOptions struct {
+	*LocatorOptions
+}
```
AgnesToulet, line 386: What's the current benefit of having a new type instead of using `LocatorOptions` directly?

internal/js/modules/k6/browser/common/selectors.go
inancgumus: It prevents issues with selectors like:

```js
page.getByRole("listitem >> internal:has-text='Product 2'")
```

We'd receive the following error because of the quotes (from [here](https://github.com/grafana/k6/blob/a23ef68b17c8e74ef8d2b35b8db012eec32a2bb0/internal/js/modules/k6/browser/common/js/injected_script.js#L1433)):

```bash
Uncaught (in promise) getting text content of "internal:role=listitem >> internal:has-text='Product 2' >> nth=0": SyntaxError: Unexpected token ''', "'Product 2'" is not valid JSON
```

These selectors can be generated/used in another part of the module while chaining, and if we remove `TrimQuotes` from here, we'd be making the execution definitely more brittle down the line.

I've tried to solve this issue in alternative ways, but it proved more challenging than I thought, which is why I've added a note here: https://github.com/grafana/k6/pull/5114#discussion_r2321404068. Our implementation differs from PW's, and I believe that might be the cause of issues like this. For example, [here](https://github.com/grafana/k6/blob/a23ef68b17c8e74ef8d2b35b8db012eec32a2bb0/internal/js/modules/k6/browser/common/selectors.go#L82-L88) in `selectors.go`, a `case` location determines the parsing order, and if we move these two `case` down, it won't work. Over time, this will lead to a behavioral difference between us and PW, and reduce our productivity due to bugs, and user unhappiness, and so on.

internal/js/modules/k6/browser/common/selectors.go
inancgumus: Tried to explain why this is here, [here](https://github.com/grafana/k6/pull/5114#discussion_r2322220566).

internal/js/modules/k6/browser/common/selectors.go
inancgumus: I've tried to explain it [here](https://github.com/grafana/k6/pull/5114#discussion_r2322220566). Let me know if you have other questions 🙇

internal/js/modules/k6/browser/common/locator.go
inancgumus: I went on that route. However, it proved to be a significant change to a lot of existing code (in my trial, I applied these changes). I then thought that it could expose us to aliasing effects because `LocatorOptions` will include pointers very soon (e.g., to other `Locator`s). For example, if we create a new `Locator`, it would copy the pointer fields; even if we replace those pointers with something else later, it would cause unexpected effects that are difficult to detect, as the pointer will be stale. It's similar to slice aliasing effects in the wild.

internal/js/modules/k6/browser/common/locator.go
inancgumus: I explain it in [this commit comment](https://github.com/grafana/k6/pull/5114/commits/c93a793bdd5d690a87ff0b5b3cfafefa0ab31ae8):

`LocatorFilterOptions` embed `LocatorOptions`, as their options are the same except for the `visible` option, which we'll soon add. Their options are similar because `filter` is just a chain operation for locators.

References:
- https://playwright.dev/docs/api/class-locator#locator-filter
- https://playwright.dev/docs/api/class-page#page-locator

---

So, this is about reflecting the nature of these options; one is like a parent to the other. We avoid deviation in time to reduce mistakes (like forgetting to add an option in `LocatorOptions`, and forgetting to add it in `LocatorFilterOptions`) @AgnesToulet Let me know if you think removing embedding from `LocatorFilterOptions` is a better idea and why (❤️). I'm happy to add a task to work on it in the next PR.

internal/js/modules/k6/browser/browser/locator_mapping.go
inancgumus: I added that when I was first working on this; however, I experimented with it and saw that it was not required for regexes to work. For example, [here](https://github.com/grafana/k6/blob/a23ef68b17c8e74ef8d2b35b8db012eec32a2bb0/examples/browser/locator_filter.js#L53), users can still pass regexes. If you can explain the benefit of adding `parseStringOrRegex` here, I can change it 🙇

internal/js/modules/k6/browser/common/locator.go
inancgumus: I think we should consider other options as well, because we create a new "something" with the options of the other "thing". It's semantically incorrect. Since locators are chainable in nature, changing one of the previous locator options can create a cascading aliasing effects for others, which can lead to unexpected consequences. I think this should stay as it is, but I can remove it if we have a strong case for why it should not be here 🙇

internal/js/modules/k6/browser/tests/locator_test.go
inancgumus: Yep, good ideas. I'm collecting this work as tasks in #5033 to tackle them in the next PR 👍

internal/js/modules/k6/browser/common/locator.go
inancgumus: Good catch 🙇 I noticed [the others](https://github.com/grafana/k6/pull/5114#discussion_r2319057865), but missed this one!

I've tried the current solution with options passing:

```js
let l = await page.getByRole('listitem').filter({ hasText: 'Product 2' });
l = l.locator("nth=0");
```

Results in:

```bash
internal:role=listitem >> internal:has-text=Product 2 >> nth=0 >> internal:has-text=Product 2
```

Without passing options, it results in:

```bash
internal:role=listitem >> internal:has-text=Product 2 >> nth=0
```

I'm adding this as a task to the main issue.

internal/js/modules/k6/browser/common/locator.go
ankur22: Are options shared between one locator and another, i don't think that should be the case. Can we change options once they're set? I guess you mean when something like `locator('div' {hasText: 'foo'}).filter({hasText: 'bar'})`, in which i would expect the `filter` to not override the original filter, but to add another clause to the chain.

I think for consistency we should work with options in the same way as we have done, otherwise it's confusing. If there's a strong case to change this then i would suggest we attempt to change them all.

internal/js/modules/k6/browser/browser/locator_mapping.go
ankur22: Oh, i missed that. Nice, that's great 👍 No need for `parseStringOrRegex` here in that case.

internal/js/modules/k6/browser/common/selectors.go
ankur22: 👍 Fair enough. Let's track this in a new issue, but can we remove the TODO comment?

internal/js/modules/k6/browser/common/selectors.go
ankur22: We shouldn't be working with raw/naked selectors like:

```js
page.getByRole("listitem >> internal:has-text='Product 2'")
```

We should be pushing users to work with the APIs that we have now provided. Raw selectors are difficult to work with. Plus the use of `internal:has-text` isn't something we want to expose to the user, since it's an _internal_ detail.

Considering that, do we still need `TrimQuotes`?

internal/js/modules/k6/browser/common/locator.go
ankur22: Do you have an example of this aliasing issue showing up in a test script? I don't understand it 😅

internal/js/modules/k6/browser/common/locator.go
ankur22: Would you mind fixing that in this PR? If we forget then it'll be released with this incorrect functionality.

internal/js/modules/k6/browser/common/selectors.go
inancgumus: I'll create an issue and change the TODO to a link to that issue 👍

internal/js/modules/k6/browser/common/selectors.go
inancgumus: > We should be pushing users to work with the APIs that we have now provided. 

I agree that it's an internal detail, however, I don't think we'll push users since we don't document it. Besides, there are PW scripts that use this style in the wild (e.g., [PW docs](https://playwright.dev/docs/other-locators#chaining-selectors), [2](https://www.playwright-user-event.org/playwright-tips/playwright-tip-of-the-week-using-internal-has-text-in-locators)), and they'll have issues while adding their scripts to k6-browser. `internal:has-text` (and other `internal` use), by its nature, is already exposed to users, even if we don't document it. People who are coming from PW are already aware of this, and we use the same underlying engine. I don't think that's a good idea to remove `TrimQuotes` from here. Since you have a strong opinion on this, do you see a specific case where using it creates an issue?

internal/js/modules/k6/browser/common/locator.go
inancgumus: ~~I'll send the next PR today or tomorrow, so we won't forget. It's a quick fix.~~ Alright, let's do that in this PR.

internal/js/modules/k6/browser/common/selectors.go
ankur22: Ok, sure, leave it in. Can you add a comment why it's there. Thanks!

internal/js/modules/k6/browser/common/locator.go
ankur22: Btw, i meant replace `l.opts` to `nil`, not to add the `options` to `locator` -- that part can be done in another PR.

internal/js/modules/k6/browser/common/locator.go
inancgumus: I don't have an example to prove this, but I have a strong hunch that this will be problematic down the line, and it won't be easy to debug. It's because it's semantically incorrect. We use the same pointer for different things. Currently, I don't have the time to modify the remaining options, but I could add a new issue. But, other options are not chainable like `Locator`s. So, for now, I'll simply remove this copying since you have a strong opinion. I've updated the main issue.

internal/js/modules/k6/browser/common/locator.go
inancgumus: Yep, let's do the first in this PR, and I'll add a task to add `options` to `Locator.Locator` in another PR.

internal/js/modules/k6/browser/common/selectors.go
inancgumus: Sure thing. Adding it as a task.

inancgumus:
@ankur22, @AgnesToulet, asking for another round of reviews, thanks! I've added a new commit to fix [a possible issue](https://github.com/grafana/k6/pull/5114#discussion_r2321340163) with `Locator.Locator`.

internal/js/modules/k6/browser/browser/locator_mapping.go
inancgumus: 🙇

internal/js/modules/k6/browser/common/selectors.go
inancgumus: See #5131.

internal/js/modules/k6/browser/common/locator.go
inancgumus: See #5131.

internal/js/modules/k6/browser/common/selectors.go
inancgumus: See #5131.

ankur22 (approved):
Thanks for making the `locator` `opts` fix 🙇  LGTM 🚀

Issue #4624, opened 2025-03-12
Move `--no-summary` as a `--summary-mode` and deprecate it


joanlopez:
Since https://github.com/grafana/k6/pull/4089, there's a new flag (and option) named `--summary-mode`, which lets users choose between `compact`, `full` or `legacy`.

But, we still have `--no-summary` as separated flag, which arguably could be another mode: "disabled" or "none", to avoid having two ways of doing the same (i.e. configuring what to show as end-of-test summary).

However, as that flag is being widely used, we should follow a proper deprecation path:

1. Adding the new option to summary mode, and make it behave as no summary.
2. Mark no summary as deprecated (display a warning).
3. (_Likely on v2.0, create another issue once ready_) Remove support for no summary.

PR #5118, opened 2025-09-01
Move `--no-summary` as a `--summary-mode` and deprecate it

joanlopez:
👋🏻  _(I have already created https://github.com/grafana/k6/issues/5119, which somehow accompanies the `TODO`s I left in the code, and will represent the effective cleanup for `--no-summary`)_

## What?

It adds a new `disabled` summary mode and marks `--no-summary` as deprecated, with the idea of being removed by v2.0, as described in #5119.

## Why?

Because we want to unify both the codebase and the user-experience, by unifying the available options around the end-of-test summary and its different modes.

## Checklist

## Related PR(s)/Issue(s)

Closes https://github.com/grafana/k6/issues/4624

### DIFF
--- a/internal/cmd/run.go
+++ b/internal/cmd/run.go
@@ -174,13 +174,15 @@ func (c *cmdRun) run(cmd *cobra.Command, args []string) (err error) {
 		return err
 	}
 
+	summaryMode, summaryEnabled := getSummaryMode(testRunState.RuntimeOptions, logger)
+	thresholdsEnabled := !testRunState.RuntimeOptions.NoThresholds.Bool
+
 	// We'll need to pipe metrics to the MetricsEngine and process them if any
 	// of these are enabled: thresholds, end-of-test summary
-	shouldProcessMetrics := !testRunState.RuntimeOptions.NoSummary.Bool ||
-		!testRunState.RuntimeOptions.NoThresholds.Bool
+	shouldProcessMetrics := summaryEnabled || thresholdsEnabled
 	var metricsIngester *engine.OutputIngester
 	if shouldProcessMetrics {
-		err = metricsEngine.InitSubMetricsAndThresholds(conf.Options, testRunState.RuntimeOptions.NoThresholds.Bool)
+		err = metricsEngine.InitSubMetricsAndThresholds(conf.Options, !thresholdsEnabled)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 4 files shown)


internal/cmd/run.go
```diff
@@ -504,6 +498,22 @@ func (c *cmdRun) run(cmd *cobra.Command, args []string) (err error) {
...
+		return summary.ModeDisabled, false
+	}
+
+	sm, err := summary.ValidateMode(runtimeOptions.SummaryMode.String)
+	if err != nil {
+		logger.WithError(err).Warnf(
+			"invalid summary mode %q, falling back to \"compact\" (default)",
+			runtimeOptions.SummaryMode.String,
+		)
+	}
```
mstoykov: This seems like a thing that should abort the execution?

internal/cmd/tests/cmd_run_test.go
```diff
@@ -154,32 +154,71 @@ func TestStdoutAndStderrAreEmptyWithQuietAndHandleSummary(t *testing.T) {
...
-	logContents, err := fsext.ReadFile(ts.FS, logFilePath)
-	require.NoError(t, err)
-	assert.Equal(t, "init\ninit\nfoo\n", string(logContents)) //nolint:dupword
+		// Instead, it should be in the log file
+		logContents, err := fsext.ReadFile(ts.FS, logFilePath)
+		require.NoError(t, err)
+		assert.Equal(t, "init\ninit\nfoo\n", string(logContents)) //nolint:dupword
+	})
+
+	t.Run("--summary-mode=disabled", func(t *testing.T) {
```
mstoykov, line 192: Can we also get a test with unknown summary mode?

mstoykov (changes requested):
LGTM in general. My only request is to have a test for unknown summary mode option and for that to actually error out to the user.

internal/cmd/run.go
joanlopez: I have mixed feelings here, but yeah, maybe.

In any case this is not something new, see https://github.com/grafana/k6/blob/master/internal/cmd/run.go#L217.
So, if you suggest this, I could agree, but then I'd leave for a separate PR.

mstoykov: We do clearly error out 
image
 very likely in some other place. But maybe we can have it better and also for it to not look as we do not.

joanlopez: Turns out to be the case that my memory is horrible 😓 

Here it is: https://github.com/grafana/k6/blob/master/internal/cmd/runtime_options.go#L122-L125

joanlopez: So, the fact is that this should never happen, unless we eventually introduce a "bug" that drops the early validation.
In such case, what would happen here? Again, throw an error? Panic? I can leave a comment explaining why.

mstoykov (dismissed):

internal/cmd/tests/cmd_run_test.go
joanlopez: Done in https://github.com/grafana/k6/pull/5118/commits/d47caf654f180b74f919d36288eea8e6d8956a81! 👍🏻

internal/cmd/run.go
joanlopez: Resolved in https://github.com/grafana/k6/commit/d47caf654f180b74f919d36288eea8e6d8956a81! 👍🏻

PR #5121, opened 2025-09-02
Feature/3552 grpc tests separate process

### LINKED ISSUE #3552: Run tests in separate process so that we actually test loading proto definitions at runtime

paulnegz:
This test verifies that k6 can properly load proto definitions at runtime when the server and client are in separate processes. This ensures that we're not relying on the global registry being populated by the server process.

Fixes #3552

## What?

This PR adds a new test (`TestGRPCSeparateProcess`) that:
- Starts a gRPC server in a separate process.
- Runs a k6 script in another process, which loads the proto file at runtime and connects to the server.
- Verifies that the client can successfully invoke a gRPC method using only the proto definition, ensuring that type registration is not leaking between processes.

## Why?

Previously, gRPC tests ran both the server and client in the same process, which could mask issues with proto type registration due to Go's global registry. This test ensures that k6's gRPC client can independently load and use proto definitions at runtime, matching real-world usage and preventing false positives in tests.

## Checklist

## Related PR(s)/Issue(s)

- Fixes #3552


examples/grpc_server/main.go
```diff
@@ -35,7 +35,7 @@ import (
 	"google.golang.org/grpc/reflection"
 	"google.golang.org/grpc/testdata"
 
-	"go.k6.io/k6/internal/lib/testutils/grpcservice"
```
mstoykov, line 38: I checked and this works locally, so I am not certain why you are moving it, but I really would prefer not to have `testutils` be a public package

internal/cmd/tests/cmd_run_grpc_separate_process_test.go
```diff
@@ -0,0 +1,72 @@
...
+  
+  client.close();
+};
+`, grpcServer.Addr)
+
+	// Set up test state using the existing framework
+	ts := getSingleFileTestState(t, script, []string{"-v", "--log-output=stdout", "--no-usage-report"}, 0)
+
+	// Read the actual proto file from testutils (same as working gRPC test)
+	protoContent, err := os.ReadFile("../../../testutils/grpcservice/route_guide.proto") //nolint:forbidigo
```
mstoykov, line 59: Just changign this to be the correct internal path makes this pass, without any of the other changes.

internal/cmd/tests/cmd_run_grpc_separate_process_test.go
```diff
@@ -0,0 +1,72 @@
...
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// TestGRPCSeparateProcess tests that the gRPC client can properly load proto definitions
+// at runtime, which is the core functionality needed for separate process scenarios.
+// This validates the issue #3552 requirement without actually launching separate processes.
+func TestGRPCSeparateProcess(t *testing.T) {
+	t.Parallel()
```
mstoykov, line 20: I do not see how you can test this without separate process and actually be certain it is working. 

The whole point is that as long as the process are the same, there is no gurauntee that we are correctly loading them, or if it just piggyback on the fact the server has already loaded the same protobufs.

mstoykov (changes requested):
Hi @paulnegz, thank you for the PR :bow: 

Unfortunately there are a couple of problesm with it:
1. it does move an internal package to be not internal, and I am pretty sure this is not needed as I commented
2. The issue specifically wants to run in separate processes so we are *not* dependant on knowing how the current implementaiton works. As shown recently this does have problems and we have hit an issue that a test with separate process would've caught, but we didn't have such one.

mstoykov:
I am closing this due to inactivity. Please if you want it reopened write here or just open a new one. 

Thanks for all the work and ressearch @paulnegz

PR #5122, opened 2025-09-02
Use Azure Trusted Signing

martincostello:
## What?

Use Azure Trusted Signing to sign the Windows artifacts instead of the existing code-signing certificate.

The `k6.exe` file itself is now also signed.

~~I suggest this is merged as-is to complete the rotation of the certificate, then I can do a follow-up PR to refactor this to use a shared re-usable action for signing once that's implemented and merged (see https://github.com/grafana/shared-workflows/pull/1290).~~

### Before

image

### After

image

## Why?

The existing certificate expires on Thursday 2nd October 2025.

## Checklist

- [ ] ~~I have added tests for my changes.~~
- [ ] ~~I have run linter and tests locally (`make check`) and all pass.~~

- [ ] ~~I have updated or added an issue to the [TypeScript definitions](https://github.com/grafana/k6-DefinitelyTyped/tree/master/types/k6): grafana/k6-DefinitelyTyped#NUMBER if applicable~~

## Related PR(s)/Issue(s)

None.


martincostello:
Verified artifacts from [this workflow run](https://github.com/grafana/k6/actions/runs/17405442100?pr=5122):

- Have a `.zip` that contains the signed Windows binary
- Have a signed `.msi`
- The `.msi` contains the signed Windows binary

martincostello:
Verified artifacts from [this workflow run](https://github.com/grafana/k6/actions/runs/17405761926?pr=5122):

- Skip all signing-related steps
- `signed-windows-binary` contains an unsigned version of `k6.exe`
- `binaries-signed` contains an unsigned version of `k6.exe`
- `binaries-windows-signed` contains an unsigned `.msi`.

codebien (commented):
Hey @martincostello, thanks for addressing it.

I see now we're generating several artifacts, where before we were generating only two. Is it expected? I'm also thinking if it is really needed to have a zip output if we already have the msi.

martincostello:
> I see we are now generating multiple artifacts instead of just two. Is this expected?

Yes. GitHub Actions artifacts are immutable, so the contents cannot be overwritten to replace the unsigned files with the signed files. Artifacts are being used to transfer files between jobs so that the signing jobs (which are isolated from the jobs that access the code itself) can sign the unsigned files.

> I'm also wondering if a ZIP output is truly necessary if we already have the MSI.

It can be removed if it's actually redundant, but I was just replicating the current process, which also makes the `k6.exe` binary available for download without requiring it be installed.

.github/workflows/build.yml
```diff
@@ -75,6 +78,26 @@ jobs:
           echo "GO_VERSION=${GO_VERSION}"
           echo "go_version=${GO_VERSION}" >> $GITHUB_OUTPUT
 
+      # GH secrets are unavailable when building from project forks, so this
```
codebien:
```suggestion
      # Secrets are unavailable when building from project forks, so this
```

.github/workflows/build.yml
```diff
@@ -228,43 +251,149 @@ jobs:
             -t $DOCKER_IMAGE_ID:latest-with-browser \
             -t ghcr.io/$GHCR_IMAGE_ID:latest-with-browser .
 
-  package-windows:
+  # Forks, PRs etc. won't actually sign the binary, but the workflow will run most of the same steps as
+  # GitHub Actions workflows don't support conditional `needs` so we have to run the signing step unconditionally.
```
codebien, line 253: Can't we fully skip `sign-windows-binary` by adding `if: needs.configure.outputs.sign_windows_artifacts == 'true'` to all the steps?

.github/workflows/build.yml
```diff
@@ -273,53 +402,114 @@ jobs:
...
+  sign-windows-installer:
+    permissions:
+      actions: read
+      contents: read
+      id-token: write # Required for Vault
 
-          # Get the latest signtool executable
-          $SignTool = Get-ChildItem -Path "${env:ProgramFiles(x86)}\Windows Kits\10\bin" -Recurse -Filter signtool.exe | Where-Object { $_.DirectoryName -like "*\x64" } | Sort-Object -Descending | Select-Object -First 1
+    environment:
+      name: azure-trusted-signing
```
codebien, line 409: Is this environment required for the final version?

.github/workflows/build.yml
```diff
@@ -273,53 +402,114 @@ jobs:
...
+          # renovate: datasource=dotnet-version depName=dotnet-sdk
+          dotnet-version: '8.0.413'
 
-          # Cleanup signing artifacts
-          del k6.pfx
+      - name: Install Sign CLI tool
+        if: needs.configure.outputs.sign_windows_artifacts == 'true'
+        env:
+          # renovate: datasource=nuget depName=sign
+          DOTNET_SIGN_VERSION: '0.9.1-beta.25379.1'
```
codebien:
```suggestion
          DOTNET_SIGN_VERSION: '0.9.1.25379.1'
```
Isn't a stable version available?

.github/workflows/build.yml
martincostello: No, because we need the `publish` steps to wait until the artifacts exist. It's not possible to have conditional dependencies in GitHub Actions workflows, otherwise I'd have done that.

.github/workflows/build.yml
martincostello: Yes, because the environment is part of the OIDC trust relationship.

.github/workflows/build.yml
martincostello: No there isn't. [sign versions in NuGet](https://www.nuget.org/packages/sign#versions-body-tab).

.github/workflows/build.yml
```diff
@@ -35,19 +35,21 @@ jobs:
...
           persist-credentials: false
+
       - name: Get the k6 version
         id: get_k6_version
         env:
           INPUT_K6_VERSION: ${{ github.event.inputs.k6_version  }}
         run: |
           set -x # Show exactly what commands are executed
-          if [[ "${{ github.event_name }}" == "workflow_dispatch" ]] && [[ "${INPUT_K6_VERSION}" != "" ]]; then
+          if [[ "${GITHUB_EVENT_NAME}" == "workflow_dispatch" ]] && [[ "${INPUT_K6_VERSION}" != "" ]]; then
```
mstoykov, line 51: Is there a reason we are moving to the ENV variable? This isn't one of those that can be changed by someone and execute some shell code, but still, I do not see why we should move to it if the other one is working.

.github/workflows/build.yml
```diff
@@ -228,43 +251,149 @@ jobs:
...
+        uses: azure/login@a457da9ea143d694b1b9c7c869ebb04ebe844ef5 # v2.3.0
+        if: needs.configure.outputs.sign_windows_artifacts == 'true'
+        with:
+          client-id: ${{ fromJSON(steps.get-signing-secrets.outputs.secrets).client-id }}
+          subscription-id: ${{ fromJSON(steps.get-signing-secrets.outputs.secrets).subscription-id }}
+          tenant-id: ${{ fromJSON(steps.get-signing-secrets.outputs.secrets).tenant-id }}
+
+      - name: Sign Windows binary
+        if: needs.configure.outputs.sign_windows_artifacts == 'true'
+        env:
```
mstoykov: Given that we are doing all of thos steps just to sign it, and in a lot of cases we won't be - should we move the check as upwards as possible? or on each step? 

Or just have all of the stesp combined ?

.github/workflows/build.yml
```diff
@@ -331,15 +521,15 @@ jobs:
...
         with:
-          name: binaries
+          name: binaries-signed
           path: dist
-      - name: Download Windows binaries
+      - name: Download signed Windows binaries
         uses: actions/download-artifact@v4
         with:
-          name: binaries-windows
+          name: binaries-windows-signed
```
mstoykov: I find the naming of this to be ... strange and not actually useful.

I would recommend keeping the final actual zip with signed binaries to be `binaries` and the rest to be `binaries-unsigned` 

This is based on us not being able to just replace `binaries` with the new one, when/if we sign it in the new step. If that is possible, I think I will prefer that. And then for the step with the `msi` installer to just do the same adding to it. 

Currently the final `Artifacts` on this look pretty much impossible to guess what it means. 

So if not possible ,please default to skipping `signed` and just using `unsigned` and also make the msi zip ... something like `windows-msi-installer` :shrug: so it is actually able to figure out what it is.

mstoykov:
> I suggest this is merged as-is to complete the rotation of the certificate, then I can do a follow-up PR to refactor this to use a shared re-usable action for signing once that's implemented and merged (see https://github.com/grafana/shared-workflows/pull/1290).

> The existing certificate expires on Thursday 2nd October 2025.

What will happen with binaries signed with the previous certificate? Will they still be installable? Will they stop being installable on 3rd ? 

I am asking as we have a release on 23th of September. And if that workflow will be available by then we can wait. 

We likely can ... still wait if we do not have to make a `.1` release this time around. 

If they are not going to be installable ... that is a bigger problem, but I do not know what to do about that :shrug:

.github/workflows/build.yml
martincostello: It avoids any possibility of shell injection and makes it possible to copy-paste snippets into a terminal for testing, and allows linters to parse the code better.

.github/workflows/build.yml
martincostello: 1. A lot of the steps cannot be combined for functional reasons because they aren't all arbitrary shell scripts.
2. I've left some of them in so _some_ of the pipeline still gets exercised for test purposes in non-release runs (the existing workflow also did this)
3. Various parts of this will moved to a shared action in shared-workflows at a future date, which will simplify it

.github/workflows/build.yml
martincostello: I can rename the artifacts (`x` vs `x-unsigned`), but it isn't possible to have conditional dependencies in the jobs (i.e. have the release job only dependent on the sign job if it runs). I can possibly make the _names_ of the artifacts to use dynamic (I did consider this during working on this), but it increases the complexity so I left it out as I figured that would be harder to follow/reason about.

martincostello:
> What will happen with binaries signed with the previous certificate? Will they still be installable? Will they stop being installable on 3rd ?

As long as you timestamped them (which the existing code appears to, using `/tr "http://timestamp.digicert.com" /td sha256`), they will be valid indefinitely as long as the old certificate isn't revoked.

martincostello:
Confirmed with [this build](https://github.com/grafana/k6/actions/runs/17461508316?pr=5122) that the `*-signed` artifacts are signed and that the installer contains the signed `k6.exe`.

martincostello:
Addressed various review comments:

- Shared action now used for signing
- Refactored artifact naming to flow between jobs
- Renamed new jobs
- More steps are skipped when signing isn't being used
- Reduced the number of artifacts produced by unsigned builds

.github/workflows/build.yml
```diff
@@ -228,17 +249,112 @@ jobs:
...
+        id: get-signing-secrets
+        if: needs.configure.outputs.sign_windows_artifacts == 'true'
+        with:
+          export_env: false
+          repo_secrets: |
+            client-id=azure-trusted-signing:client-id
+            subscription-id=azure-trusted-signing:subscription-id
+            tenant-id=azure-trusted-signing:tenant-id
+
+      # TODO Add functionality to allow passing a directory to sign without needing an artifact first
```
mstoykov: So we are uploading this before so the workflow can download it?

If so can we ranem it to `windows-binary-to-sign` to again be something that is more obvious

mstoykov (dismissed):
LGTM! 

I have a comment around naming, that is not blocking, and would be nice if we can it run the signing again before we merge it. 

Thank you for all the work :bow:

.github/workflows/build.yml
martincostello: No, it's also needed for when the workflow runs when signing is disabled (e.g. in a PR) (see https://github.com/grafana/k6/pull/5122/commits/69ba0db7766d013aed526f67da67b96f07357e06).

Although that does mean as the artifact has to exist _anyway_ I can just delete the comment.

martincostello:
```suggestion
```

martincostello:
> would be nice if we can it run the signing again before we merge it.

I think we're good as-is. To run it again with signing, as this is a PR, I would have to re-edit the PR force it on and ensure there isn't an accidental real publish, then revert those changes again before it can be merged. The edits since the last run with signing are trivial enough that I think there's enough confidence it works for both unsigned and signed builds.

mstoykov (dismissed):

mstoykov (approved):

Issue #5036, opened 2025-08-11
Implement `frameLocator.getBy*`


ankur22:
# What

Implement `frameLocator.getBy*` APIs which are the preferred way of working with selectors. Since we now have `page.getBy*` we should hopefully be able to easily piggy back off that work. This works involves adding:

* `frameLocator.getByRole`
* `frameLocator.getByAltText`
* `frameLocator.getByLabel`
* `frameLocator.getByPlaceholder`
* `frameLocator.getByTitle`
* `frameLocator.getByTestId`
* `frameLocator.getByText`

# Why

Working with locator based APIs is a more robust way of working with browser based APIs since it works well during navigations and other scenarios, plus it comes with a whole host of checks which are done automatically under-the-hood.

This will help make migrating from PW to k6 a little easier. When working with PW's code gen tool in VSCode or working with MCP/AI, it will work with the latest PW APIs which k6 currently lacks. By implementing this, we will make the migration process simpler.

# How

Playwright APIs: (start here) https://playwright.dev/docs/api/class-framelocator#frame-locator-get-by-alt-text
Playwright code: (start here) https://github.com/microsoft/playwright/blob/6626bba937af513cbcf05d5e164aa20b6d390ea2/packages/playwright-core/src/client/locator.ts#L415

PR #5135, opened 2025-09-08
Implement the frameLocator.getBy* methods

AgnesToulet:
## What?

Implement the following `frameLocator.getBy*` APIs which are the preferred way of working with selectors:
- frameLocator.getByAltText
- frameLocator.getByLabel
- frameLocator.getByPlaceholder
- frameLocator.getByRole
- frameLocator.getByTestId
- frameLocator.getByText
- frameLocator.getByTitle

## Why?

Because we have already implemented all those methods for `page`, `frame`, and `locator`, and we want to implement them as well for `frameLocator` as a way to increase our compatibility with Playwright, as described in https://github.com/grafana/k6/issues/5036.

## Checklist

## Related PR(s)/Issue(s)

Closes https://github.com/grafana/k6/issues/5036

### DIFF
--- a/internal/js/modules/k6/browser/tests/get_by_test.go
+++ b/internal/js/modules/k6/browser/tests/get_by_test.go
@@ -11,6 +11,13 @@ import (
 	"go.k6.io/k6/internal/js/modules/k6/browser/common"
 )
 
+const (
+	frameLocatorImpl = "frameLocator"
+	locatorImpl      = "locator"
+	frameImpl        = "frame"
+	pageImpl         = "page"
+)
+
 func TestGetByRoleSuccess(t *testing.T) {
 	t.Parallel()
 
@@ -430,15 +437,9 @@ func TestGetByRoleSuccess(t *testing.T) {
 				t.Parallel()
 
 				tb := newTestBrowser(t, withFileServer())
+				staticURL := tb.staticURL("get_by_role_implicit.html")
+				tb.withIFrameURL(staticURL)
 				p := tb.NewPage(nil)
-				opts := &common.FrameGotoOptions{
-					Timeout: common.DefaultTimeout,
-				}
-				_, err := p.Goto(
-					tb.staticURL("get_by_role_implicit.html"),
-					opts,
-				)
-				require.NoError(t, err)
 
 				getByRoleImplementations := getByImplementationsOf[interface {
 					GetByRole(role string, opts *common.GetByRoleOptions) *common.Locator
@@ -446,6 +447,12 @@ func TestGetByRoleSuccess(t *testing.T) {
 
 				for implName, impl := range getByRoleImplementations {
 					t.Run(implName, func(t *testing.T) { //nolint:paralleltest
+						if implName == frameLocatorImpl {
+							tb.GoToPage(p, tb.url("/iframe"))
+						} else {
+							tb.GoToPage(p, staticURL)
+						}
+
 						l := impl.GetByRole(tt.role, tt.opts)
 						c, err := l.Count()
 						require.NoError(t, err)
@@ -562,15 +569,9 @@ func TestGetByRoleSuccess(t *testing.T) {
 				t.Parallel()
 
 				tb := newTestBrowser(t, withFileServer())
+				staticURL := tb.staticURL("get_by_role_explicit.html")
+				tb.withIFrameURL(staticURL)
 				p := tb.NewPage(nil)
-				opts := &common.FrameGotoOptions{
-					Timeout: common.DefaultTimeout,
-				}
-				_, err := p.Goto(
-					tb.staticURL("get_by_role_explicit.html"),
-					opts,
-				)
-				require.NoError(t, err)
 
 				getByRoleImplementations := getByImplementationsOf[interface {
 					GetByRole(role string, opts *common.GetByRoleOptions) *common.Locator
@@ -578,6 +579,12 @@ func TestGetByRoleSuccess(t *testing.T) {
 
 				for implName, impl := range getByRoleImplementations {
 					t.Run(implName, func(t *testing.T) { //nolint:paralleltest
+						if implName == frameLocatorImpl {
+							tb.GoToPage(p, tb.url("/iframe"))
+						} else {
+							tb.GoToPage(p, staticURL)
+						}
+
 						l := impl.GetByRole(tt.role, nil)
 						c, err := l.Count()
 						require.NoError(t, err)
@@ -599,28 +606,29 @@ func TestGetByRoleSuccess(t *testing.T) {
 			t.Parallel()
 
 			tb := newTestBrowser(t, withFileServer())
+			staticURL := tb.staticURL("get_by_role_explicit.html")
+			tb.withIFrameURL(staticURL)
 			p := tb.NewPage(nil)
-			opts := &common.FrameGotoOptions{
-				Timeout: common.DefaultTimeout,
-			}
-			_, err := p.Goto(
-				tb.staticURL("get_by_role_explicit.html"),
-				opts,
-			)
-			require.NoError(t, err)
 
 			getByRoleImplementations := getByImplementationsOf[interface {
 				GetByRole(role string, opts *common.GetByRoleOptions) *common.Locator
 			}](p)
 
 			expectedByImplementation := map[string]int{
-				"page":    2,
-				"frame":   2,
-				"locator": 1,
+				pageImpl:         2,
+				frameImpl:        2,
+				locatorImpl:      1,
+				frameLocatorImpl: 2,
 			}
 
 			for implName, impl := range getByRoleImplementations {
 				t.Run(implName, func(t *testing.T) { //nolint:paralleltest
+					if implName == frameLocatorImpl {
+						tb.GoToPage(p, tb.url("/iframe"))
+					} else {
+						tb.GoToPage(p, staticURL)
+					}
+
 					l := impl.GetByRole("document", nil)
 					c, err := l.Count()
 					require.NoError(t, err)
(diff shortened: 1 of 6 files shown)


internal/js/modules/k6/browser/tests/test_browser.go
```diff
@@ -219,6 +220,51 @@ func withHTTPServer() func(*testBrowser) {
 	}
 }
 
+// withIFrameContent sets up a handler for /iframe that serves a page embedding
+// an iframe with the given content.
+func (tb *testBrowser) withIframeContent(iframeHTML string) {
```
inancgumus: `IFrame` instead of `Iframe`?

Should this method and `withIFrameURL` be _options_ that are passed to `newTestBrowser`, as with other options, for consistency?


```go
tb := newTestBrowser(t, withHTTPServer(), withIFrameContent(...))
```

internal/js/modules/k6/browser/tests/test_browser.go
```diff
@@ -250,6 +296,39 @@ func withSkipClose() func(*testBrowser) {
...
+		opts,
+	)
+	require.NoError(b.t, err)
+
+	return p
+}
+
+// GoToPage is a wrapper around Page.Goto that fails the test if an error occurs.
+// Added this helper to avoid boilerplate code in tests.
+func (b *testBrowser) GoToPage(p *common.Page, url string) {
```
inancgumus: Minor cosmetic suggestion: Although `GoTo` is correct, in k6-browser and PW Lingua Franca, we refer to `GoTo` as `Goto` everywhere (e.g., `Page.Goto`, `FrameGotoOptions`, etc.). Should we change the name for consistency (i.e., `GotoNewPage` and `GotoPage`)?

internal/js/modules/k6/browser/tests/test_browser.go
```diff
@@ -250,6 +296,39 @@ func withSkipClose() func(*testBrowser) {
...
+	b.t.Helper()
+
+	p := b.NewPage(nil)
+	opts := &common.FrameGotoOptions{
+		Timeout: common.DefaultTimeout,
+	}
+	_, err := p.Goto(
+		url,
+		opts,
+	)
```
inancgumus: Minor suggestion: Should this on a single line, as it's not that long (same for the other one)?

```suggestion
_, err := p.Goto(url, opts)
```

internal/js/modules/k6/browser/tests/test_browser.go
```diff
@@ -219,6 +220,51 @@ func withHTTPServer() func(*testBrowser) {
...
+// an iframe with the given URL.
+func (tb *testBrowser) withIFrameURL(iframeURL string) {
+	tb.t.Helper()
+
+	if tb.http == nil {
+		tb.t.Fatalf("You should enable HTTP test server, see: withHTTPServer option")
+	}
+
+	// Origin A: main page embedding origin B and same-origin frame A (with dynamic B URL)
+	originAHTML := fmt.Sprintf(`<!DOCTYPE html>
```
inancgumus: Should this be just `doc` or something, because there is only one origin here?

internal/js/modules/k6/browser/tests/test_browser.go
```diff
@@ -219,6 +220,51 @@ func withHTTPServer() func(*testBrowser) {
...
+		<html>
+		<head></head>
+		<body>
+			<iframe id="frameB" src="%s"></iframe>
```
inancgumus: Do we need the `frameB` id?

internal/js/modules/k6/browser/tests/test_browser.go
```diff
@@ -219,6 +220,51 @@ func withHTTPServer() func(*testBrowser) {
...
+		w.Header().Set("Content-Type", "text/html; charset=utf-8")
+		_, err := w.Write([]byte(iframeHTML))
+		require.NoError(tb.t, err)
+	})
+	srvB := httptest.NewServer(muxB)
+	tb.t.Cleanup(func() {
+		srvB.Close()
+	})
+
+	tb.withIFrameURL(srvB.URL)
```
inancgumus: I believe these `muxB`, `srvB` come from the pasted code. Should they be better without a `B`, like `srv` instead of `srvB`, etc.?

internal/js/modules/k6/browser/tests/get_by_test.go
```diff
@@ -1472,8 +1481,9 @@ func TestGetByNullHandling(t *testing.T) {
 
 func getByImplementationsOf[T any](p *common.Page) map[string]T {
 	return map[string]T{
-		"page":    any(p).(T),
-		"frame":   any(p.MainFrame()).(T),
-		"locator": any(p.Locator(":root", nil)).(T),
+		pageImpl:         any(p).(T),
```
inancgumus, line 1485: Not related to this PR, but I wonder why we use generics here when we could do the same with interfaces, or use a much simpler approach. A task for another time, I guess.

internal/js/modules/k6/browser/tests/test_browser.go
AgnesToulet: Renamed in https://github.com/grafana/k6/pull/5135/commits/62090aa62f81de0ecfc632c6ae10412f5992da8e

internal/js/modules/k6/browser/tests/test_browser.go
AgnesToulet: Yes, it's used in the tests with the `Locator` function to retrieve the iframe.

internal/js/modules/k6/browser/tests/test_browser.go
AgnesToulet: Renamed in https://github.com/grafana/k6/pull/5135/commits/62090aa62f81de0ecfc632c6ae10412f5992da8e

internal/js/modules/k6/browser/tests/test_browser.go
AgnesToulet: Done in https://github.com/grafana/k6/pull/5135/commits/e27052736905c28548d9faeab42fe85e433c84e3

internal/js/modules/k6/browser/tests/test_browser.go
AgnesToulet: Done in https://github.com/grafana/k6/pull/5135/commits/e27052736905c28548d9faeab42fe85e433c84e3

internal/js/modules/k6/browser/tests/test_browser.go
AgnesToulet: Good idea! `withIFrameURL` can't be an option because the URL is sometimes known after the testBrowser gets initialized (see https://github.com/grafana/k6/pull/5135/files#diff-7ec25e41d5b272a1ab43d25bb03d48bdd68072ac9632c3295c27f14ce4370011R572-R573) but I refactored `withIFrameContent` to be an option as it simplifies a bit the locator tests (in https://github.com/grafana/k6/pull/5135/commits/62090aa62f81de0ecfc632c6ae10412f5992da8e).

AgnesToulet:
> Not related to this PR, but I wonder why we need to load IFrames using an HTTP server when we could do it with a simple iframe srcdoc

I've tested this approach and I like how it simplifies the tests but, I'm not sure why, it makes one test (`TestVisibilityWithCORS/visible`) in `locator_test.go` fails. I'll do some other tests with Playwright to see if it's a bug or an expected behavior but I don't think it should be handled in this PR.
@ankur22 Have you looked into this while working on those tests?

inancgumus:
@AgnesToulet 

> I've tested this approach and I like how it simplifies the tests but, I'm not sure why, it makes one test (`TestVisibilityWithCORS/visible`) in `locator_test.go` fails. I'll do some other tests with Playwright to see if it's a bug or an expected behavior but I don't think it should be handled in this PR.

Adding the `allow-same-origin` and `allow-scripts` attributes to the `sandbox` attribute of `iframe` can likely solve that issue (e.g., `<iframe srcdoc='..' sandbox='allow-same-origin'></iframe>`).

internal/js/modules/k6/browser/tests/test_browser.go
inancgumus: I'd explicitly pass the frame ID to that helper then, as it creates a temporal coupling issue. We'd have to look at that helper code to see why `frameB` is used in the test every time.

internal/js/modules/k6/browser/tests/test_browser.go
inancgumus: 👍 Hmm, don't we already initialize `testBrowser` before applying the options? The only issue is with `withIFrameURL`, we'd need to pass that after `withHTTPServer`.

https://github.com/grafana/k6/blob/62090aa62f81de0ecfc632c6ae10412f5992da8e/internal/js/modules/k6/browser/tests/test_browser.go#L76-L83

ankur22:
> > Not related to this PR, but I wonder why we need to load IFrames using an HTTP server when we could do it with a simple iframe srcdoc

> I've tested this approach and I like how it simplifies the tests but, I'm not sure why, it makes one test (TestVisibilityWithCORS/visible) in locator_test.go fails. I'll do some other tests with Playwright to see if it's a bug or an expected behavior but I don't think it should be handled in this PR.
@ankur22 Have you looked into this while working on those tests?

Might have misunderstood the question, but the reason why we're working with a HTTP server is to mimic a different origin i.e. CORS. Anything in injected_script.js can read from an `iframe` from the same origin, but if it is a different origin then chrome will prevent it due to CORS. So if we're working with the `iframe srcdoc` attribute, wouldn't the iframe be from the same origin as the main frame? The assumption in browser is that all content from an iframe is form a different origin. If there's a less complex way then great 👍

internal/js/modules/k6/browser/tests/test_browser.go
AgnesToulet: Updated in https://github.com/grafana/k6/pull/5135/commits/91affdd91bcfc6ac8fc8650d146ae6ede327173e

PR #5131, opened 2025-09-04
Changes for `Locator.Filter` reviews

inancgumus:
## What?

Changes for `Locator.Filter` reviews.

## Why?

See #5033 tasks.

## Checklist

- [x] I have performed a self-review of my code.
- [x] I have commented on my code, particularly in hard-to-understand areas.
- [ ] I have added tests for my changes.
- [x] I have run linter and tests locally (`make check`) and all pass.

## Checklist: Documentation (only for k6 maintainers and if relevant)

**Please do not merge this PR until the following items are filled out.**

- [x] I have added the correct milestone and labels to the PR.
- [ ] I have updated the release notes: _link_

## Related PR(s)/Issue(s)

#5114

PR #5150, opened 2025-09-08
Add `Locator|FrameLocator.Locator(..., LocatorOptions)`

inancgumus:
## What?

- Adds `LocatorOptions` to `Locator.Locator`.
- Adds `LocatorOptions` to `FramrLocator.Locator`.
- Tests the behavior with `LocatorOptions` and `Locator.Filter`.

## Why?

It's a part of #5033.

## Checklist

- [x] Release notes, k6-docs, and TS are being managed by the #5033 issue.

## Related PR(s)/Issue(s)

- Updates: #5033 
- Related: #5114
- Related: #5131

### DIFF
--- a/examples/browser/locator_filter.js
+++ b/examples/browser/locator_filter.js
@@ -35,7 +35,7 @@ export default async function() {
 
   // Filter the 2nd item by filtering items with the "Product 2" text.
   await check(page, {
-    'textContent': async p => {
+    'Use locator.filter the Add to cart 2 button': async p => {
       const txt = await p
         .getByRole('listitem')
         .filter({ hasText: 'Product 2' })
@@ -47,10 +47,23 @@ export default async function() {
 
   // Using a regex, filter the 1st item by filtering items without the "Product 2" text.
   await check(page, {
-    'textContent': async p => {
+    'Use locator.filter the Add to cart 1 button': async p => {
       const txt = await p
         .getByRole('listitem')
         .filter({ hasNotText: /Product 2/ })
+        .first()
+        .textContent();
+      return txt.includes(`Add to cart 1`);
+    }
+  });
+
+  // Filtering with the locator.locator options instead of the locator.filter method.
+  await check(page, {
+    'Use locator.options to filter the Add to cart 1 button': async p => {
+      const txt = await p
+        .getByRole('list')
+        .locator('li', { hasText: 'Product' })
+        .locator('button', { hasText: /Add to cart 1/ })
         .first()
         .textContent();
       return txt.includes(`Add to cart 1`);
--- a/examples/browser/locator_filter.js
+++ b/examples/browser/locator_filter.js
@@ -35,7 +35,7 @@ export default async function() {
 
   // Filter the 2nd item by filtering items with the "Product 2" text.
   await check(page, {
-    'Use locator.filter the Add to cart 2 button': async p => {
+    'Filter the Add to cart 2 button': async p => {
       const txt = await p
         .getByRole('listitem')
         .filter({ hasText: 'Product 2' })
@@ -47,7 +47,7 @@ export default async function() {
 
   // Using a regex, filter the 1st item by filtering items without the "Product 2" text.
   await check(page, {
-    'Use locator.filter the Add to cart 1 button': async p => {
+    'Filter the Add to cart 1 button': async p => {
       const txt = await p
         .getByRole('listitem')
         .filter({ hasNotText: /Product 2/ })
@@ -59,7 +59,7 @@ export default async function() {
 
   // Filtering with the locator.locator options instead of the locator.filter method.
   await check(page, {
-    'Use locator.options to filter the Add to cart 1 button': async p => {
+    'Filter the Add to cart 1 button with locator options': async p => {
       const txt = await p
         .getByRole('list')
         .locator('li', { hasText: 'Product' })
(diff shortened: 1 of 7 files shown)


examples/browser/locator_filter.js
```diff
@@ -35,7 +35,7 @@ export default async function() {
 
   // Filter the 2nd item by filtering items with the "Product 2" text.
   await check(page, {
-    'textContent': async p => {
+    'Use locator.filter the Add to cart 2 button': async p => {
```
AgnesToulet:
```suggestion
    'Use locator.filter to filter the Add to cart 2 button': async p => {
```
Nit: I know this is a bit redundant but the initial sentence confuses me a bit :sweat_smile:

internal/js/modules/k6/browser/browser/frame_locator_mapping.go
```diff
@@ -9,9 +9,8 @@ import (
 func mapFrameLocator(vu moduleVU, fl *common.FrameLocator) mapping {
 	rt := vu.Runtime()
 	return mapping{
-		"locator": func(selector string) *sobek.Object {
-			ml := mapLocator(vu, fl.Locator(selector))
-			return rt.ToValue(ml).ToObject(rt)
+		"locator": func(selector string, opts sobek.Value) mapping {
+			return mapLocator(vu, fl.Locator(selector, parseLocatorOptions(rt, opts)))
```
AgnesToulet, line 79: Can we really do this? There are a lot of places in the codebase where we return `rt.ToValue(ml).ToObject(rt)`, what does it do exactly and why can we remove it here?

internal/js/modules/k6/browser/tests/locator_test.go
```diff
@@ -1172,6 +1172,180 @@ func TestLocatorFilter(t *testing.T) {
...
+			Locator("div", &common.LocatorOptions{
+				HasText: "hello",
+			}).
+			Count()
+		require.NoError(t, err)
+		require.Equal(t, 1, n)
+	})
+}
+
+func TestLocatorLocatorOptions(t *testing.T) {
```
AgnesToulet, line 1183: This would be maybe a bit more concise using an array of test cases.

AgnesToulet (commented):
LGTM! Left a few minor comments, I'm mostly confused about the removal of `rt.ToValue(ml).ToObject(rt)` but this might be because I don't understand fully what it does.

internal/js/modules/k6/browser/browser/frame_locator_mapping.go
inancgumus: Sobek automatically converts the returned `mapping`. That's what we do when we return a `mapping` in other places of the mapping layer. Turning it into an object is unnecessary because the mapping layer does that here:

https://github.com/grafana/k6/blob/3dfe1259311f3e77fd20429a65150055a005e3c9/internal/js/modules/k6/browser/browser/mapping.go#L24-L37

internal/js/modules/k6/browser/tests/locator_test.go
inancgumus: I'd like to keep tests readable rather than concise (still, `setupPage` helps to keep it short). Before this, I tried table tests, but it became an unreadable mess, as each subtest is a different scenario.

examples/browser/locator_filter.js
inancgumus: Ah, right :)

internal/js/modules/k6/browser/browser/frame_locator_mapping.go
AgnesToulet: If I understand this code correctly, this turns the mapping to an object only for the set of mappings defined in `mapBrowser`, I don't see how it's related to our `mapLocator` function.

inancgumus: Returned mappings become a part of that returned object. Also: Maps and objects are similar. If you were to `console.log` this (returning a `mapping` or an `Object`) from the script side, both would return `object`. That's also what we do in every mapping function (i.e., `mapSomething(..) mapping`. Think of it like a tree of `mapping`s: `mapBrowser` returns a `mapping`, which in turn returns another `mapping` when one of `Browser` methods like `context` is called:

https://github.com/grafana/k6/blob/3dfe1259311f3e77fd20429a65150055a005e3c9/internal/js/modules/k6/browser/browser/browser_mapping.go#L17-L22

Returning a `mapping` from `mapLocator` is similar, as it returns a `mapping`.

ankur22: I do agree that we should just be letting Sobek convert from a Go [mapping](https://github.com/grafana/k6/blob/master/internal/js/modules/k6/browser/browser/mapping.go#L18) to a Sobek object. I think that's what is happening in the case of `return mapLocator(vu, fl.Locator(selector, parseLocatorOptions(rt, opts)))` as Inanc has mentioned.

I think `mapBrowserToSobek` was used as an example of what Sobek is doing under the hood when we return a `mapping` back to the Sobek runtime, internally it is converting the Go `map` to a `sobek.Object`.

Side note: I did a quick experiment with `mapBrowserToSobek`, i don't think we need to convert to a `sobek.Object`, we should just be able to work with the `mapping` i.e. `map[string]any`.

ankur22 (approved):
LGTM 🚀 Thanks! 😄

ankur22 (approved):

AgnesToulet (approved):

PR #5154, opened 2025-09-09
Migrate to Renovate

codebien:
## What?

We want to use Renovate instead of Dependabot.

## Why?

A more flexible and powerful configuration.

## Checklist


renovate.json
```diff
@@ -0,0 +1,72 @@
+{
+	"$schema": "https://docs.renovatebot.com/renovate-schema.json",
+	"extends": ["config:best-practices"],
```
codebien: @dsotirakis Should we explicitly mention any Grafana presets, or is the current form sufficient?

dsotirakis: @codebien 👋 
Grafana presets will be enabled by default (they are in the global config that our self-hosted solution uses).
You can explicitly disable them though if something doesn't work for you. 
https://github.com/grafana/grafana-renovate-config, which will soon be public 😄

codebien:
```suggestion
```
Hey @dsotirakis, thanks for the support. :bow: 

If I understand correctly, there isn't any need for defining this additional entry to the config. I'm dropping it.

renovate.json
```diff
@@ -0,0 +1,71 @@
...
+				"github.com/grafana/sobek",
+				"github.com/jhump/protoreflect",
+				"github.com/klauspost/compress",
+				"github.com/tidwall/gjson",
+				"github.com/chromedp/cdproto",
+				"github.com/mailru/easyjson",
+				"github.com/mattn/go-colorable",
+				"github.com/mstoykov/k6-taskqueue-lib",
+				"github.com/stretchr/testify"
+			]
```
mstoykov: Can we actually have a disable list? I wanted this with dependabot as well, it just didn't have the functionality, but it seems we can put `"enabled": false` and it will disable it for some depenancies?

renovate.json
```diff
@@ -0,0 +1,71 @@
...
+		{
+			"description": "Dockerfiles root",
+			"managers": ["docker"],
+			"excludePaths": ["packaging/**"],
+			"schedule": ["quarterly"]
+		},
+		{
+			"description": "Dockerfiles packaging",
+			"managers": ["docker"],
+			"matchPaths": ["packaging/**"]
```
mstoykov: Should this also have quaterly schedule >?

renovate.json
```diff
@@ -0,0 +1,71 @@
...
+			"groupName": "otel",
+			"description": "Group OpenTelemetry packages",
+			"managers": ["gomod"],
+			"matchPackagePatterns": ["^go\\.opentelemetry\\.io/"]
+		},
+		{
+			"description": "xk6 experimental modules",
+			"managers": ["gomod"],
+			"matchPackagePatterns": ["^github\\.com/grafana/xk6-"]
+		},
```
mstoykov: We can probably drop this if we have a disable list

codebien:
>I would prefer to have a disable list

I'll do it :+1:

codebien:
@mstoykov this now handles the disable list. I've also fixed a few of issues reported from the Renovate's validator.

renovate.json
codebien: This was a mistake, I intended to do the opposite. My intention was to slow down the packaging dockerfiles updates, as being on the latest version shouldn't be critical for it.

I set a faster (monthly) update schedule for the k6 Dockerfiles, because I expect we prefer regular updates on those.

codebien: Ideally, we should add the same rule for Alpine regarding update only when a patch is available. But let's see how it works and we can refine it after.

mstoykov (approved):

oleiade (approved):
Looks Good To Me 👀

joanlopez (approved):

Issue #4573, opened 2024-06-13
Better handling k6's `Rate` metric


olegbespalov:
# What?

[k6's `Rate` metric](https://grafana.com/docs/k6/latest/using-k6/metrics/) tracks how frequently a non-zero value occurs.

So from the k6's metrics engine, we get either the `0` or `1`, and inside the open telemetry output extension, we split this into two metrics:

* `metric_name.occurred`
* `metric_name.total`

where the `metric_name.occurred` registers only under the condition that the [value was non-zero](https://github.com/grafana/xk6-output-opentelemetry/blob/main/pkg/opentelemetry/output.go#L179)

This approach has an obvious downside: it could produce a case where `metric_name.occurred`  could not be produced at all.

For instance, if we run an example script with a demo dashboard after navigation to [the `Checks list`](https://github.com/grafana/xk6-output-opentelemetry/blob/main/grafana/dashboards/k6-prometheus.json#L1751) panel, we quickly spot that `protocol is HTTP/2` is missing.

So we need to think about how we can handle this more efficiently.

### Always start metric with the `0` value

One of the possible approaches suggested by @joanlopez  and was used [inside grafana](https://github.com/grafana/grafana/blob/main/pkg/infra/metrics/metricutil/utils.go#L46)

TBD

# Why?

Handling of the Rate metric should be transparent and cover all possible cases.

PR #5164, opened 2025-09-10
otel: Export rate metrics as a single counter 

codebien:
## What?

Rate metrics are now exported as a single counter with a label that can have two values: `zero` and `nonzero`.

## Why?

Most of the time a single metric should be enough to query the main stats for this kind of query. We are aware that sometimes it might require a bit more complexity in case of aggregation but it seems to be acceptable.

It automatically solves the issue with cold startup of the metric because we expect at least a value between nonzero and zero.

## Checklist

## Related PR(s)/Issue(s)

Closes https://github.com/grafana/k6/issues/4573

### DIFF
--- a/internal/output/opentelemetry/output.go
+++ b/internal/output/opentelemetry/output.go
@@ -183,7 +183,7 @@ func (o *Output) dispatch(entry metrics.Sample) error {
 		if o.config.SingleCounterForRate.Bool {
 			err = o.singleCounterForRate(ctx, name, attributeSetOpt, entry)
 		} else {
-			// Deprecated path, remove with https://github.com/grafana/k6/issues/5185 
+			// Deprecated path, remove with https://github.com/grafana/k6/issues/5185
 			err = o.pairOfCountersForRate(ctx, name, attributeSetOpt, entry)
 		}
 		if err != nil {


internal/output/opentelemetry/output.go
```diff
@@ -53,12 +54,12 @@ func (o *Output) StopWithTestError(_ error) error {
 	o.logger.Debug("Stopping...")
 	defer o.logger.Debug("Stopped!")
 
+	o.periodicFlusher.Stop()
```
codebien, line 57: We have to stop the ingestion before stopping the flushing.

internal/output/opentelemetry/output.go
```diff
@@ -173,19 +174,23 @@ func (o *Output) dispatch(entry metrics.Sample) error {
...
+		} else {
+			valueType = "zero"
 		}
-		total.Add(ctx, 1, attributeSetOpt)
+		valset := attribute.NewSet(attribute.String("condition", valueType))
+
+		rate.Add(ctx, 1, attributeSetOpt, otelMetric.WithAttributeSet(valset))
 	default:
-		o.logger.Warnf("metric %q has unsupported metric type", entry.Metric.Name)
+		return fmt.Errorf("metric %q has unsupported metric type", entry.Metric.Name)
```
codebien, line 193: If this happens is an error. There isn't a way to create an untyped metric.

codebien:
@oleiade @inancgumus I've decided to be less brutal and do this change a bit more gradual. We will have a feature flag for the next release, so if we overlooked at any topic or if users need more time then they can still use the previous method without breaking their workflow.

examples/docker-compose/opentelemetry/docker-compose.yml
```diff
@@ -5,7 +5,7 @@ networks:
 
 services:
   prometheus:
-    image: prom/prometheus:v3.2.0
+    image: prom/prometheus:v3.5.0
```
joanlopez, line 6: Thanks for keeping this file up-to-date. Do you think we can also get rid of the `version` attribute, as per https://docs.docker.com/reference/compose-file/version-and-name/?

codebien: Done

examples/docker-compose/opentelemetry/grafana/dashboards/k6-prometheus.json
```diff
@@ -264,7 +264,7 @@
             "uid": "${DS_PROMETHEUS}"
           },
           "editorMode": "code",
-          "expr": "avg(k6_http_req_failed_occurred{testid=~\"$testid\"}/k6_http_req_failed_total)*100)",
+          "expr": "avg(rate(k6_http_reqs{testid=~\"$testid\", expected_response=\"false\"}[$__rate_interval])/rate(k6_http_reqs{testid=~\"$testid\"}[$__rate_interval]))*100 or vector(0)",
```
codebien: reminder: unexpected

examples/docker-compose/opentelemetry/grafana/dashboards/k6-prometheus.json
```diff
@@ -1748,7 +1751,7 @@
           },
           "editorMode": "code",
           "exemplar": false,
-          "expr": "k6_checks_occurred{testid=~\"$testid\"} / k6_checks_total{testid=~\"$testid\"}",
+          "expr": "rate(k6_checks_total{testid=~\"$testid\", condition=\"nonzero\"}[$__rate_interval]) / rate(k6_checks_total{testid=~\"$testid\"}[$__rate_interval])",
```
joanlopez, line 1754: I think this is wrong, as I have tested it with successful and unsuccessful checks and it still display: `100%`.

joanlopez: It might be something along the lines of:

```
sum by (check) (k6_checks_total{testid=~"$testid", condition="nonzero"}) / sum by (check) (k6_checks_total{testid=~"$testid"})
```

internal/output/opentelemetry/config.go
```diff
@@ -68,6 +68,11 @@ type Config struct {
 	// GRPCExporterInsecure disables client transport security for the Exporter's gRPC
 	// connection.
 	GRPCExporterInsecure null.Bool `json:"grpcExporterInsecure" envconfig:"K6_OTEL_GRPC_EXPORTER_INSECURE"`
+
+	// SingleCounterForRate sets the feature flag defining how to export metrics defined as Rate type.
+	// When it is sets to true, metrics are exported as a single counter.
```
joanlopez: Nit;

```suggestion
	// When it is set to true, metrics are exported as a single counter, using an attribute as discriminator.
```

internal/output/opentelemetry/output.go
```diff
@@ -173,19 +179,57 @@ func (o *Output) dispatch(entry metrics.Sample) error {
 
 		trend.Record(ctx, entry.Value, attributeSetOpt)
 	case metrics.Rate:
-		nonZero, total, err := o.metricsRegistry.getOrCreateCountersForRate(name)
+		var err error
+		if o.config.SingleCounterForRate.Bool {
+			err = o.singleCounterForRate(ctx, name, attributeSetOpt, entry)
+		} else {
+			// Deprecated path, remove on k6 1.4.0
```
joanlopez: Kind reminder for creating an issue for this, and assigning it to the 1.4.0 milestone.

internal/output/opentelemetry/registry.go
```diff
@@ -77,6 +77,38 @@ func (r *registry) getOrCreateHistogram(name, unit string) (otelMetric.Float64Hi
 	return h, nil
 }
 
+func (r *registry) getOrCreateCounterForRate(name string) (otelMetric.Int64Counter, error) {
+	// k6's rate metric tracks how frequently a non-zero value occurs.
+	// To be accurate is a percentage, which is a ratio.
+	// To correctly calculate this metric in a metrics backend,
+	// we need to split the rate metric via a dedicated label.
```
joanlopez: Nit;

```suggestion
	// we need to split the rate metric via a dedicated attribute.
```

internal/output/opentelemetry/output.go
joanlopez: Good catch! 👏🏻

internal/output/opentelemetry/output.go
joanlopez: _(Note this IS NOT blocking! ⚠️ )_

I'm not sure whether it's related to this or not, but from time to time I see the following error:

```sh
INFO[0010] 2025/09/16 15:39:10 failed to upload metrics: context canceled: rpc error: code = Canceled desc = context canceled 
```

especially when using the old method, which I guess is more prone to fail because it needs to send two metrics instead of one.

Have you seen this before? Was that what made you change the order of these calls?

**This is definitely **NOT** an issue introduced by your changes, because it happens with `master` as well**, but I'm wondering whether there's something within this shutdown process we could to completely avoid that.

_Perhaps we should create an issue and investigate it later, and consider if we should include it as part of https://github.com/grafana/k6/issues/5173._

internal/output/opentelemetry/output.go
codebien: https://github.com/grafana/k6/issues/5185 done

internal/output/opentelemetry/output.go
codebien: Honestly, I'm not aware of it. Did you see it on local tests or across networks? I've ran multiples local tests and I never hit it.

I discovered the current bug mostly by writing the unit test. Because they are sensible if the last batch has been correctly flushed or not.

The error reported smells like a bug related a bad management of the context in the recurring flush operation. It's a good idea to open a bug report and we have to tackle it as part of the graduation process.

examples/docker-compose/opentelemetry/grafana/dashboards/k6-prometheus.json
inancgumus: Do we have a test that verifies this behavior (whether it's incorrect or not)?

codebien: @inancgumus Do you mean an automated or a manual tests? In terms of automation, we don't have it. If you want to manual test it, then you can go to `examples/docker-compose/opentelemetry/*` which is the path where this dashboard is in.

joanlopez:
> In general, I'm quite satisfied with the approach, and it really solves the issue reported 👍🏻 Although, I have to admit that condition="zero|nonzero" still resonates in my head as unlikely the best candidate/most intuitive one.

After an internal chat, we decided to stick with `condition="zero|nonzero"` as any other alternative evaluated so far is also confusing in certain conditions (_note here that signs of positivity like `positive|negative` and `success|failure` would suffer the same as discussed in https://github.com/grafana/k6/issues/2306, an in some other related issues/pull requests, so we stick to a boolean representation_). 

If we ever find a better candidate, we can always add the new attribute and eventually get rid of the old one 👍🏻

examples/docker-compose/opentelemetry/grafana/dashboards/k6-prometheus.json
joanlopez: Yeah, I basically tested this manually.

I'm not sure if there's an easy way to automate that, or if it does worth the effort of automating such thing (probably not?).

internal/output/opentelemetry/output.go
mstoykov:
```suggestion
			// Deprecated path, remove with https://github.com/grafana/k6/issues/5185 
```

I do not think it will be in v1.4.0 as we will be releasing this change there. I am also not certain if single version is good idea. Especially as maintaining this seems to not be particularly problematic. So I would probably leave it without specific version and potentially remove it in v1.5.0 if there is no feedback against it.

mstoykov (approved):
LGTM :+1:

internal/output/opentelemetry/output.go
codebien: I agree, I had too optimistic hopes here.

PR #5167, opened 2025-09-10
remove enable community extensions flag

pablochacin:
## What?

Removes the enable community extensions flag and the need to adjust the build service's URL to select the `oss` catalog when they were enabled.

## Why?

After alll extensions were integrated into a single catalog that includes both official and community extensions, and all these extensions are supported in GCK6, there's no need to explicitly enable community extensions.

## Checklist

### DIFF
--- a/cmd/state/state.go
+++ b/cmd/state/state.go
@@ -2,7 +2,6 @@ package state
 
 import (
 	"context"
-	"fmt"
 	"io"
 	"os"
 	"os/signal"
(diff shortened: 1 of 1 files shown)


codebien (approved):

codebien:
@pablochacin Is it really a breaking change? In the end the behavior is not changing. If this is really a breaking change then we have to work-around it because it isn't allowed in v1.x, or we have to wait for v2.x.

pablochacin:
@codebien I think it is a breaking change in the sense that users can't opt out of community extensions.
Also, we are removing an environment variable and in my understanding those are part of the API

codebien (changes requested):
@pablochacin I now agree with you, this is a breaking change and it isn't allowed so we have to discuss how to make it backward compatible, or wait for v2.

codebien:
We decided to just deprecate this flag for now and drop it for v2.

PR #5168, opened 2025-09-10
Browser: Don't skip handling local requests

joanlopez:
## Checklist

### DIFF
--- a/internal/js/modules/k6/browser/common/network_manager.go
+++ b/internal/js/modules/k6/browser/common/network_manager.go
@@ -558,15 +558,16 @@ func (m *NetworkManager) onRequest(event *network.EventRequestWillBeSent,
 		m.logger.Errorf("NetworkManager", "creating request: %s", err)
 		return
 	}
-	// Skip data and blob URLs, since they're internal to the browser.
-	if isInternalURL(req.url) {
-		m.logger.Debugf("NetworkManager", "skipping request handling of %s URL", req.url.Scheme)
-		return
-	}
+
 	m.reqsMu.Lock()
 	m.reqIDToRequest[event.RequestID] = req
 	m.reqsMu.Unlock()
-	m.emitRequestMetrics(req)
+
+	// Skip data and blob URLs, since they're internal to the browser.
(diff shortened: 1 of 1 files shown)


internal/js/modules/k6/browser/common/network_manager.go
```diff
@@ -558,15 +558,16 @@ func (m *NetworkManager) onRequest(event *network.EventRequestWillBeSent,
 		m.logger.Errorf("NetworkManager", "creating request: %s", err)
 		return
 	}
-	// Skip data and blob URLs, since they're internal to the browser.
```
joanlopez, line 561: Of course this would break `TestURLSkipRequest`, as that test is precisely preventing regressions on these lines.

joanlopez: Arguably this would be a breaking change, so we might need to consider carefully, and perhaps leave it up to k6 v2.0, if we really want to ship it.

internal/js/modules/k6/browser/common/network_manager.go
```diff
@@ -558,15 +558,16 @@ func (m *NetworkManager) onRequest(event *network.EventRequestWillBeSent,
...
+
 	m.reqsMu.Lock()
 	m.reqIDToRequest[event.RequestID] = req
 	m.reqsMu.Unlock()
-	m.emitRequestMetrics(req)
+
+	// Skip data and blob URLs, since they're internal to the browser.
+	if !isInternalURL(req.url) {
+		m.emitRequestMetrics(req)
+	}
```
joanlopez, line 569: What I don't know, because I lack the historical context from the original PR, is: do we want to skip handling local requests at all, or only to not emit metrics for those requests?

Because with the former, k6 differs from Playwright behavior, in the sense that if a request is supposed to fail, like:
`blob:something`, for which Chrome (and Playwright) throws ` net::ERR_FILE_NOT_FOUND at blob:something`, k6 doesn't, and the error wouldn't get any error.

While, with the changeset suggested here, the error would be surfaces, as with PW+Chrome, while we would still not emit request metrics for those URLs.

Issue #5175, opened 2025-09-12
With K6_PROMETHEUS_RW_TREND_STATS and some tags set the metrics are not writing __name__ label 


cuelebreorg:
### Brief summary

When using K6_PROMETHEUS_RW_TREND_STATS=p(95),p(99),min,max with "experimental-prometheus-rw" if a tag is added the metrics are not correctly expanded and the first label is written with the trend suffix

### k6 version

master

### OS

windows 10 , mac OS 

### Steps to reproduce the problem

Use this simple test 

```js
import http from 'k6/http';
import { check, sleep } from 'k6';
import exec from 'k6/execution';

export let options = {
    vus: 1,
    duration: '2s',
    iterations: 1
    
};

export default function () {
    exec.vu.metrics.tags.BU = 'AA';
    const res = http.get('https://www.google.com');
    check(res, {
        'status is 200': (r) => r.status === 200,
    });

```
You can also check same behavior using tags #https://grafana.com/docs/k6/latest/using-k6/tags-and-groups/#test-wide-tags

### Expected behaviour

__name=k6_http_req_receiving_min
BU=AA

### Actual behaviour

__name=k6_http_req_receiving
BU=AA_min

cuelebreorg:
After some extra investigation i think the problem is here 
https://github.com/grafana/k6/blob/4bdc1093a6ae6452f1893d1bddfd2cafcc534350/internal/output/prometheusrw/remotewrite/trend.go#L113

And should be 
` return tg.labels[i].Name >= namelbl `
As the docs in https://pkg.go.dev/sort#Search tell us

joanlopez:
Hey @cuelebreorg,

Thanks for reporting the issue and for doing such a detailed research 🙌🏻 
You're right, that's what's causing the issue, although I took the opportunity to move this code to the newer APIs, so the final  code changes with the fix are slightly different.

[Here](https://github.com/grafana/k6/pull/5178)'s the PR with the fix. The tentative target release is `v1.3.0`, but I cannot guarantee it. If not, it will definitely be in `v1.4.0`, or `v1.3.1` if there's any other more impactful bug or security issue that forces us to do such release.

In any case, I really want to express you my gratitude for your contribution! 🙇🏻

PR #5178, opened 2025-09-15
Fix Prometheus RW trend stats metric transformation

joanlopez:
## What?

It changes the way we look for the `__name__` label in order to transform trend stats metrics, according to our docs.

## Why?

Because the current code doesn't always find the label, as expected, and thus it doesn't produce the expected outcome.

## Checklist

## Related PR(s)/Issue(s)

Closes https://github.com/grafana/k6/issues/5175

Thanks for your contribution @cuelebreorg! 🙏🏼

### DIFF
--- a/internal/output/prometheusrw/remotewrite/prometheus.go
+++ b/internal/output/prometheusrw/remotewrite/prometheus.go
@@ -1,7 +1,8 @@
 package remotewrite
 
 import (
-	"sort"
+	"cmp"
+	"slices"
 
 	prompb "buf.build/gen/go/prometheus/prometheus/protocolbuffers/go"
 	"github.com/mstoykov/atlas"
@@ -45,8 +46,8 @@ func MapSeries(series metrics.TimeSeries, suffix string) []*prompb.Label {
 		Name:  namelbl,
 		Value: v,
 	})
-	sort.Slice(lbls, func(i int, j int) bool {
-		return lbls[i].Name < lbls[j].Name
+	slices.SortStableFunc(lbls, func(i, j *prompb.Label) int {
+		return cmp.Compare(i.Name, j.Name)
 	})
 	return lbls
 }
--- a/internal/output/prometheusrw/remotewrite/remotewrite_test.go
+++ b/internal/output/prometheusrw/remotewrite/remotewrite_test.go
@@ -67,15 +67,29 @@ func TestOutputConvertToPbSeries(t *testing.T) {
 			Time:  time.Date(2022, time.September, 1, 0, 0, 0, 0, time.UTC),
 			Value: 7,
 		},
+		metrics.Sample{
+			TimeSeries: metrics.TimeSeries{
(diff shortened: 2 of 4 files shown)


internal/output/prometheusrw/remotewrite/prometheus.go
```diff
@@ -45,8 +46,8 @@ func MapSeries(series metrics.TimeSeries, suffix string) []*prompb.Label {
 		Name:  namelbl,
 		Value: v,
 	})
-	sort.Slice(lbls, func(i int, j int) bool {
-		return lbls[i].Name < lbls[j].Name
+	slices.SortStableFunc(lbls, func(i, j *prompb.Label) int {
```
joanlopez, line 49: This isn't strictly required, but let's be a good citizen and move onto the modern APIs.

internal/output/prometheusrw/remotewrite/remotewrite_test.go
```diff
@@ -67,15 +67,29 @@ func TestOutputConvertToPbSeries(t *testing.T) {
...
 		},
+		metrics.Sample{
+			TimeSeries: metrics.TimeSeries{
+				Metric: registry.MustNewMetric("metric4", metrics.Trend),
+				Tags: registry.RootTagSet().WithTagsFromMap(tagset.Map()).
+					With("We_need", "value").
+					With("a_long", "value").
+					With("list_of", "value").
+					With("tags_to", "value").
+					With("reproduce_5175", "value"),
```
joanlopez, line 78: While [this](https://github.com/grafana/k6/issues/5175#issuecomment-3287788875) comment was super illuminating, it required me some extra time to figure out while the test was still passing with the old code.

Turns out to be the case that, the issue was not only reproducible when there was a label that starts with a capital letter, but only when the total amount of labels is also large enough. That's why adding a new tag starting with a capital letter wasn't enough.

Note that in the real-world scenario reported in #5175, the user is only adding one tag (starting with capital letter!), but k6 adds some more (enough!) when reporting the HTTP metrics (like `proto`,  `url`, etc).

internal/output/prometheusrw/remotewrite/trend.go
```diff
@@ -108,12 +109,10 @@ func (tg *trendAsGauges) CacheNameIndex() {
 
 	// in the case __name__ is not the first
 	// then search for its position
-
-	i := sort.Search(len(tg.labels), func(i int) bool {
-		return tg.labels[i].Name == namelbl
+	i, found := slices.BinarySearchFunc(tg.labels, namelbl, func(lbl *prompb.Label, key string) int {
```
joanlopez, line 112: The fix was as simple as replacing:

```go
return tg.labels[i].Name == namelbl
```

with:

```go
return tg.labels[i].Name >= namelbl
```

as per what's mentioned in the issue's comment, and as stated in the Go docs.

--

However, as said above as well, I preferred to be a good citizen and move onto the modern APIs, which are also less confusing and thus less error prone.

codebien (approved):

Issue #4501, opened 2021-10-14
Implement `unroute(url[, handler])`


robingustafsson:
Add support for `page.unroute(url[, handler])`, to remove route(s).

**Relevant links**:
- Playwright docs: https://playwright.dev/docs/api/class-page#page-unroute
- xk6-browser code starting point: https://github.com/grafana/xk6-browser/blob/main/common/page.go
- Playwright code: https://github.com/microsoft/playwright/blob/6626bba937af513cbcf05d5e164aa20b6d390ea2/packages/playwright-core/src/client/page.ts#L545

PR #5223, opened 2025-10-07
browser: Implement unroute(url) and unrouteAll

AgnesToulet:
## What?

This PR implements two functions to remove routes that were added using the `page.route` function:
- `page.unroute(url)`: removes the route(s) that match the same URL (url needs to be exactly what was used in the `route` function, it can't be a regex that matches the string used in `route` - this is the behavior of Playwright as well).
- `page.unrouteAll()`: removes all routes

Note: `unroute(url, handler)` as in PW will be way harder to implement in the k6 project because of the Sobek layer so this isn't included in this PR. We are waiting for strong interest from users before deciding to look into it.

## Why?

This allows users to reset the routes setup during a test or between tests if needed. 

## Checklist

- [x] I have updated or added an issue to the [k6-documentation](https://github.com/grafana/k6-docs): grafana/k6-docs#NUMBER if applicable
- [x] I have updated or added an issue to the [TypeScript definitions](https://github.com/grafana/k6-DefinitelyTyped/tree/master/types/k6): grafana/k6-DefinitelyTyped#NUMBER if applicable

## Related PR(s)/Issue(s)

Closes https://github.com/grafana/k6/issues/4501

### DIFF
--- a/internal/js/modules/k6/browser/common/page.go
+++ b/internal/js/modules/k6/browser/common/page.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"slices"
 	"strings"
 	"sync"
 	"sync/atomic"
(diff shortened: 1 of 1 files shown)


internal/js/modules/k6/browser/common/page.go
```diff
@@ -1418,6 +1418,44 @@ func (p *Page) Route(
...
+	defer p.routesMu.Unlock()
+
+	var filteredRoutes []*RouteHandler
+	for _, route := range p.routes {
+		if route.path != path {
+			filteredRoutes = append(filteredRoutes, route)
+		}
+	}
+
+	p.routes = filteredRoutes
```
joanlopez: Nit; I guess this could be simplified to:

```suggestion
p.routes = slices.DeleteFunc(p.routes, func(rh *RouteHandler) bool {
    return rh.path == path
})
```

PS: It would require adding the "slices" import.

AgnesToulet: Good idea :+1: Updated in https://github.com/grafana/k6/pull/5223/commits/3337f162093506c1c212d7e8eeb9a998ee50e46b

mstoykov (approved):

Issue #3846, opened 2024-07-15
Flakt tests due to api address collisions


mstoykov:
![image](https://github.com/user-attachments/assets/50638317-b699-48a1-a763-1fff670a3d97)

This seems to happen because 
https://github.com/grafana/k6/blob/cb3f41c50b37b1fddc1b5da4856b5098c8db37c7/cmd/tests/test_state.go#L121
is used to get the address, but it gets an address, then closes it. And after that k6 uses it.

There is a not small chance that another test will ask for an address between the close and k6 using it and then it will not be free. In practice even another process on the OS can get that address.

Unfortunately quite a lot of tests *do use* the address, so we can't just remove it.

It might be nice if we can get it *from* k6 running (maybe logs) and then use it to communicate?
Maybe we should not set it for tests that don't need it?

PR #5227, opened 2025-10-08
Fix API Address Collisions: expose listener for serving

rMaxiQp:
## What?

Expose listeners that `getFreeBindAddr()` creates

## Why?

Currently, `getFreeBindAddr()` finds a free-to-use address by try-binding. After it finds the valid port, it returns the API address, and closes the listener. In between the old listener closing and new listener opening, the address port can be stolen. One fix here would be exposing the old listener without closing.

One caveat, now the `GlobalState` has a `net.Listener` field hanging mainly because it's needed for testing... Maybe there can be a better approach here.

## Checklist

## Related PR(s)/Issue(s)

Relates to #3846

### DIFF
--- a/internal/cmd/run.go
+++ b/internal/cmd/run.go
@@ -57,68 +57,6 @@ const (
 	waitForTracerProviderStopTimeout = 3 * time.Minute
 )
 
-// serve starts the REST API server and ensures it is properly shut down when
-// the srvCtx is done. It returns a function that can be waited on to ensure
-// the server has fully stopped.
-func (c *cmdRun) serve(globalCtx context.Context, srv *http.Server, addrSetByUser bool) func() {
-	var logger logrus.FieldLogger = c.gs.Logger
-
-	srvCtx, srvCancel := context.WithCancel(globalCtx)
-
-	// We cannot use backgroundProcesses here, since we need the REST API to
-	// be down before we can close the samples channel above and finish the
-	// processing the metrics pipeline.
-	apiWG := &sync.WaitGroup{}
-	apiWG.Add(2)
-
-	go func() {
-		defer apiWG.Done()
-		logger.Debugf("Starting the REST API server on %s", c.gs.Flags.Address)
-		if c.gs.Flags.ProfilingEnabled {
-			logger.Debugf("Profiling exposed on http://%s/debug/pprof/", c.gs.Flags.Address)
-		}
-
-		// ServerListener is set up in tests
-		if c.gs.ServerListener != nil {
-			if err := srv.Serve(c.gs.ServerListener); err != nil && !errors.Is(err, http.ErrServerClosed) {
-				if addrSetByUser {
-					logger.WithError(err).Error("Error from API server")
-					c.gs.OSExit(int(exitcodes.CannotStartRESTAPI))
-				} else {
-					logger.WithError(err).Warn("Error from API server")
-				}
-			}
-			return
-		}
-
-		if err := srv.ListenAndServe(); err != nil && !errors.Is(err, http.ErrServerClosed) {
-			// Only exit k6 if the user has explicitly set the REST API address
-			if addrSetByUser {
-				logger.WithError(err).Error("Error from API server")
-				c.gs.OSExit(int(exitcodes.CannotStartRESTAPI))
-			} else {
-				logger.WithError(err).Warn("Error from API server")
-			}
-		}
-	}()
-
-	go func() {
-		defer apiWG.Done()
-		<-srvCtx.Done()
-		shutdCtx, shutdCancel := context.WithTimeout(globalCtx, 1*time.Second)
-		defer shutdCancel()
-		if aerr := srv.Shutdown(shutdCtx); aerr != nil {
-			logger.WithError(aerr).Debug("REST API server did not shut down correctly")
-		}
-	}()
-
-	return func() {
-		// Cancel the server context before shutdown to avoid the deadlocks
-		srvCancel()
-		apiWG.Wait()
-	}
-}
-
 // TODO: split apart some more
 //
 //nolint:funlen,gocognit,gocyclo,cyclop
@@ -363,8 +301,18 @@ func (c *cmdRun) run(cmd *cobra.Command, args []string) (err error) {
 	})
 	samples := make(chan metrics.SampleContainer, test.derivedConfig.MetricSamplesBufferSize.Int64)
 	// Spin up the REST API server, if not disabled.
-	if c.gs.Flags.Address != "" {
+	if c.gs.Flags.Address != "" { //nolint:nestif
 		initBar.Modify(pb.WithConstProgress(0, "Init API server"))
+
+		// We cannot use backgroundProcesses here, since we need the REST API to
+		// be down before we can close the samples channel above and finish the
+		// processing the metrics pipeline.
+		apiWG := &sync.WaitGroup{}
+		apiWG.Add(2)
+		defer apiWG.Wait()
+
+		srvCtx, srvCancel := context.WithCancel(globalCtx)
+		defer srvCancel()
 
 		srv := api.GetServer(
 			runCtx,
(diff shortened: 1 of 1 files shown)


AgnesToulet:
Hi! Thanks for your contribution, the approach seems OK to me. An alternative would be to move the code responsible for setting up the server in the `api` package (along the `GetServer` function) and mock it in the tests but it also seems a big refactor "just" for testing.

To make the code more readable, I would move [theses lines](https://github.com/grafana/k6/blob/9087c24e7343493570ca604128adb3e8bef06e20/internal/cmd/run.go#L294-L352) in a separate function though and I would completely separate the code used in tests (where `c.gs.ServerListener != nil`) and the code that existed before with comments to explain all this. Something like: 
```
// ServerListener is set up in tests 
if c.gs.ServerListener != nil {
	if aerr := srv.Serve(c.gs.ServerListener); aerr != nil {
		logger.WithError(aerr).Error("Error from API server")
		c.gs.OSExit(int(exitcodes.CannotStartRESTAPI))
	}
} else {
	if aerr := srv.ListenAndServe(); aerr != nil && !errors.Is(aerr, http.ErrServerClosed) {
		// Only exit k6 if the user has explicitly set the REST API address
		if cmd.Flags().Lookup("address").Changed {
			logger.WithError(aerr).Error("Error from API server")
			c.gs.OSExit(int(exitcodes.CannotStartRESTAPI))
		} else {
			logger.WithError(aerr).Warn("Error from API server")
		}
	}
}
```
This ensures the behavior at runtime doesn't change (as in your changes, where the error handling is a bit different than it was before).

Also, the `xk6` tests are failing on Windows and MacOS with your changes. I'm not sure why but I can reproduce on my Windows machine so I'll look into it deeper today.

rMaxiQp:
Hi @AgnesToulet , thank you for your input! I tried to refactor the code a bit to address your comment. Failing on Mac and Windows is strange... I was trying to reproduce the error on my Mac Air, and it seems to struggle with other issues atm (many browser instances are created). I'll try to reproduce on Mac a bit more later but also create a Windows setup.

internal/cmd/run.go
```diff
@@ -55,6 +55,54 @@ const (
...
+			if err := srv.ListenAndServe(); err != nil && !errors.Is(err, http.ErrServerClosed) {
+				// Only exit k6 if the user has explicitly set the REST API address
+				if cmd.Flags().Lookup("address").Changed {
+					logger.WithError(err).Error("Error from API server")
+					c.gs.OSExit(int(exitcodes.CannotStartRESTAPI))
+				} else {
+					logger.WithError(err).Warn("Error from API server")
+				}
+			}
+		}
```
AgnesToulet:
```suggestion
		if c.gs.ServerListener != nil {
			if err := srv.Serve(c.gs.ServerListener); err != nil && !errors.Is(err, http.ErrServerClosed) {
				logger.WithError(err).Error("Error from API server")
				c.gs.OSExit(int(exitcodes.CannotStartRESTAPI))
			}
			return
		}
		
	    if err := srv.ListenAndServe(); err != nil && !errors.Is(err, http.ErrServerClosed) {
			// Only exit k6 if the user has explicitly set the REST API address
			if cmd.Flags().Lookup("address").Changed {
				logger.WithError(err).Error("Error from API server")
				c.gs.OSExit(int(exitcodes.CannotStartRESTAPI))
			} else {
				logger.WithError(err).Warn("Error from API server")
			}
		}
	}
```
This would reduce the complexity and allow to get rid of the `nolint:nestif`.

internal/cmd/run.go
```diff
@@ -55,6 +55,54 @@ const (
 	waitForTracerProviderStopTimeout = 3 * time.Minute
 )
 
+// serve starts the REST API server and ensures it is properly shut down when
+// the srvCtx is done. It returns a function that can be waited on to ensure
+// the server has fully stopped.
+//
+// nolint:nestif
+func (c *cmdRun) serve(cmd *cobra.Command, srv *http.Server, shutdown func()) func() {
```
AgnesToulet: This doesn't seem necessary to pass the full `Command` object to the function, what about just having an `addressFlagChanged bool` param? (name could probably be improved :sweat_smile:)

AgnesToulet (commented):
Thanks for refactoring the code. It fixed the xk6 issue as well as you were changing the existing behavior a bit before (returning an error not only when the address was changed).

I still find it a bit hard to read as part of the code related to the server is still in the `run` function. What about moving more code in the `serve` function, something like:
```
func (c *cmdRun) serve(globalCtx context.Context, cmd *cobra.Command, srv *http.Server) func() {
	var logger logrus.FieldLogger = c.gs.Logger
	srvCtx, srvCancel := context.WithCancel(globalCtx)

	// We cannot use backgroundProcesses here, since we need the REST API to
	// be down before we can close the samples channel above and finish the
	// processing the metrics pipeline.
	apiWG := &sync.WaitGroup{}
	apiWG.Add(2)

	go func() {
		defer apiWG.Done()
		logger.Debugf("Starting the REST API server on %s", c.gs.Flags.Address)
		if c.gs.Flags.ProfilingEnabled {
			logger.Debugf("Profiling exposed on http://%s/debug/pprof/", c.gs.Flags.Address)
		}

		// ServerListener is set up in tests
		if c.gs.ServerListener != nil {
			if err := srv.Serve(c.gs.ServerListener); err != nil && !errors.Is(err, http.ErrServerClosed) {
				logger.WithError(err).Error("Error from API server")
				c.gs.OSExit(int(exitcodes.CannotStartRESTAPI))
			}

			return
		}

		if err := srv.ListenAndServe(); err != nil && !errors.Is(err, http.ErrServerClosed) {
			// Only exit k6 if the user has explicitly set the REST API address
			if cmd.Flags().Lookup("address").Changed {
				logger.WithError(err).Error("Error from API server")
				c.gs.OSExit(int(exitcodes.CannotStartRESTAPI))
			} else {
				logger.WithError(err).Warn("Error from API server")
			}
		}
	}()

	go func() {
		defer apiWG.Done()
		<-srvCtx.Done()
		shutdCtx, shutdCancel := context.WithTimeout(globalCtx, 1*time.Second)
		defer shutdCancel()
		if aerr := srv.Shutdown(shutdCtx); aerr != nil {
			logger.WithError(aerr).Debug("REST API server did not shut down correctly")
		}
	}()

	return func() {
		srvCancel()
		apiWG.Wait()
	}
}
```

rMaxiQp:
Hi @AgnesToulet , thank you for the reviews! 

I didn't realize that my change altered the default behavior earlier 🙈. Thank you for pointing it out. Though it's a bit verbose, now we should have same behavior between actual and test besides the `net.Listener` change.

internal/cmd/run.go
```diff
@@ -55,6 +55,68 @@ const (
 	waitForTracerProviderStopTimeout = 3 * time.Minute
 )
 
+// serve starts the REST API server and ensures it is properly shut down when
+// the srvCtx is done. It returns a function that can be waited on to ensure
+// the server has fully stopped.
+func (c *cmdRun) serve(globalCtx context.Context, srv *http.Server, addrSetByUser bool) func() {
```
codebien: Why are we doing this refactor here? It doesn't seem to be directly useful to the `getFreeBindAddr` issue. We should divide this in a different pull request.

internal/cmd/run.go
```diff
@@ -310,31 +362,8 @@ func (c *cmdRun) run(cmd *cobra.Command, args []string) (err error) {
 			metricsEngine,
 			execScheduler,
 		)
-		go func() {
```
codebien, line 325: It usually better to have the concurrency control on the caller instead of moving nested in a function.

We should do
```go
go c.serve(
// or

go func () {
  c.serve(
}()
```

internal/cmd/tests/test_state.go
```diff
@@ -118,28 +119,26 @@ func NewGlobalTestState(tb testing.TB) *GlobalTestState {
...
 		addr := net.JoinHostPort("localhost", strconv.FormatUint(port, 10))
 
 		listener, err := (&net.ListenConfig{}).Listen(tb.Context(), "tcp", addr)
 		if err != nil {
 			continue // port was busy for some reason
 		}
-		defer func() {
-			assert.NoError(tb, listener.Close())
-		}()
-		return addr
```
codebien, line 140: Why are we not closing anymore?

internal/cmd/tests/test_state.go
```diff
@@ -118,28 +119,26 @@ func NewGlobalTestState(tb testing.TB) *GlobalTestState {
 		FallbackLogger: testutils.NewLogger(tb).WithField("fallback", true),
 		Usage:          usage.New(),
 		TestStatus:     lib.NewTestStatus(),
+		ServerListener: listener,
```
codebien, line 122: If `ServerListener` is only used in tests then we should just deprecate it and use a local variable for the testing package.

codebien (changes requested):
Hey @rMaxiQp, thanks for your contribution. I've left a few comments.

internal/cmd/tests/test_state.go
rMaxiQp: The original issue was that the port is tested out then released. In between the port release and reclaim, this port can be taken by other processes, leading to flaky address collision.

The current proposal would be returning the opened connection instead. From a quick search, `getFreeBindAddr` can be exposed to gRPC usage or http server usage. The original gRPC usage creates the listener using returning address and no explicit `Close()`. For HTTP server, the original implementation is using `ListenAndServe` which create the listener then `Serve`. Here, we can use `Serve` instead which would also handle [Listener Closing](https://cs.opensource.google/go/go/+/master:src/net/http/server.go;l=3447;drc=0e1bd8b5f17e337df0ffb57af03419b96c695fe4?q=serve&ss=go%2Fgo).

internal/cmd/tests/test_state.go
rMaxiQp: Could you elaborate more on this topic? 

`ServerListener` is used in `/internal/cmd/run.go` where we want to span up a http server using specific address. Since the current address is still claimed by the listener, calling `ListenAndServe` would fail but `Serve` would work. If we use a local variable for the testing package, is there an alternative to alter how we spin up the server in `/internal/cmd/run.go` ?

rMaxiQp:
Hi @codebien, thank you for the comments! I can start working on migrating the refactor on top a separate PR. For other comments, I shared my current understanding and hopefully it helps to complete the picture. If there are some context that I'm missing out, please let me know!

internal/cmd/run.go
rMaxiQp: I'm trying out with #5370 to address your comment on the refactoring, and please let me know if I misunderstood your suggestion 🙏

codebien: Yes, it is what I meant.

rMaxiQp: Sounds good! We can get the refactor #5370 reviewed & merged first then clean up the mess in this PR then. It should have less diff afterwards.

codebien:
Apologies for the delayed response. I've been investigating this week intot this and unfortunately, it sounds we have a structural issue.
I’m exploring a different approach to address this. We have largely eliminated the core reasons why the server needed to start with every new execution, so now it might be easier if we could simply [skip starting the server](https://github.com/grafana/k6/issues/5648) altogether.

I’m closing the pull request for now because I believe this is too tight-coupled to be handled from an external contributor. I’m going to re-open the pull request as soon as we have a clear path forward and we find an option to integrate your changes. Sorry for the inconvenience.

PR #5237, opened 2025-10-09
Deprecate experimental/redis module

### LINKED ISSUE #5236: Move Redis extension to Automatic Resolution
### What?
Deprecate `k6/experimental/redis` module from core and migrate to [grafana/xk6-redis](https://github.com/grafana/xk6-redis) extension to benefit of automatic resolution.

### Migration Path

* Add deprecation warnings to k6/experimental/redis
* Update documentation with migration guide
  * In essence, use `k6/x/redis` instead of  `k6/experimental/redis`
* Remove from core on version `v1.6.0` (To be confirmed)
  * Add a sub-task to the removal collection

codebien:
## What?

It deprecates the `k6/experiemental/redis`, and it encourages to migrate to `k6/x/redis` resolved via automatic extension resolution.

The intention is to remove fully the module in one of the future releases. It has to be defined if it will be directly in v1.x or in v2.x.

## Why?

As part of the experimental process we decided to not keep redis client builtin in the k6 core binary. Instead, we want to keep it as an official extension and benefit of automatic extension resolution.

Closes #5236

### DIFF
--- a/internal/js/jsmodules.go
+++ b/internal/js/jsmodules.go
@@ -57,7 +57,7 @@ func getInternalJSModules() map[string]interface{} {
 		"k6/experimental/redis": newWarnExperimentalModule(redis.New(),
 			"k6/experimental/redis has been deprecated and will be removed in future versions."+
 				" Please migrate to the new version by changing your import to 'k6/x/redis'."+
-				" Read more here: https://grafana.com/docs/k6/latest/javascript-api/k6-x/redis"),
+				" Read more here: https://grafana.com/docs/k6/latest/javascript-api/k6-x-redis"),
 
 		// Removed modules
 		"k6/experimental/browser": newRemovedModule(
@@ -118,7 +118,6 @@ func newWarnExperimentalModule(base modules.Module, msg string) modules.Module {
 	return &warnExperimentalModule{
 		msg:  msg,
 		base: base,
-		once: sync.Once{},
 	}
 }
 


internal/js/jsmodules.go
```diff
@@ -49,10 +50,15 @@ func getInternalJSModules() map[string]interface{} {
...
 		"k6/experimental/fs":         fs.New(),
-		"k6/experimental/redis":      redis.New(),
 		"k6/experimental/streams":    streams.New(),
 		"k6/experimental/websockets": expws.New(),
 
+		// Deprecated modules
+		"k6/experimental/redis": newWarnExperimentalModule(redis.New(),
+			"k6/experimental/redis has been deprecated and will be removed in future versions."+
+				" Please migrate to the new version by changing your import to 'k6/x/redis'."+
+				" Read more here: https://grafana.com/docs/k6/latest/javascript-api/k6-x/redis"),
```
codebien:
```suggestion
				" Read more here: https://grafana.com/docs/k6/latest/javascript-api/redis"),
```

This is an invented path. For now, we aren't hosting any documentation of official extensions on grafana.com/docs. I guess we don't want to drop the documentation for redis from the official docs so we need to find a new home for it.


The current experiemental documentation is here https://grafana.com/docs/k6/latest/javascript-api/k6-experimental/redis/

internal/js/jsmodules.go
```diff
@@ -101,3 +107,22 @@ func (rm *removedModule) NewModuleInstance(vu modules.VU) modules.Instance {
 
 	return nil
 }
+
+type warnExperimentalModule struct {
+	once *sync.Once
```
inancgumus: We can get rid of the initialization code in `newWarnExperimentalModule()` if we make this a non-pointer:

```suggestion
	once sync.Once
```

internal/js/jsmodules.go
```diff
@@ -101,3 +107,22 @@ func (rm *removedModule) NewModuleInstance(vu modules.VU) modules.Instance {
...
+	once sync.Once
+	msg  string
+	base modules.Module
+}
+
+func newWarnExperimentalModule(base modules.Module, msg string) modules.Module {
+	return &warnExperimentalModule{
+		msg:  msg,
+		base: base,
+		once: sync.Once{},
```
inancgumus: We can remove this now (line 121) 🎉

internal/js/jsmodules.go
oleiade: This aligns with the approach I took with the dns extension's docs so far.

In xk6-dns, I used: `docs/sources/k6/next/javascript-api/k6-x-dns/_index.md` as a format, because my preferred approach was to treat extensions as first class citizens in the docs. Not distinguish them from modules. Simply adding a common "admonition" (I think that's how they're called, at the top mentioning that the following module is an extension (as users need to know, but it concretely doesn't change much to their usage in practice).

I wouldn't mind putting it in a separate folder like you're doing here though.

oleiade (approved):

internal/js/jsmodules.go
codebien: Done https://github.com/grafana/k6-docs/pull/2139

Issue #4311, opened 2024-10-07
Early `Context` cancelation issue


ankur22:
### What?

In CI we're getting a lot of test failures which requires a rerun of the CI job for them to pass. We usually see something like this in the output:

```bash
--- FAIL: TestLifecycleGoto (0.00s)
    --- FAIL: TestLifecycleGoto/load (0.39s)
        lifecycle_wait_test.go:582: testBrowser: canceled
```

Example CI run: https://github.com/grafana/xk6-browser/actions/runs/11222847117/job/31196156631?pr=1456

### Why?

This issue came about after we started working with the vu context to control the iteration's lifecycle.

### How?

We need to take a look at the tests that fail and fix why the context is closing earlier than expected.

### Tasks

```[tasklist]
### Tasks
- [ ] PR
- [ ] Update the k6 release notes and release tasks
```

PR #5238, opened 2025-10-10
Fix early context cancellation in browser tests

AgnesToulet:
## What?

Fix early context cancellation issue in browser tests, recent examples:
- https://github.com/grafana/k6/actions/runs/18380409959/job/52365216823?pr=5234
- https://github.com/grafana/k6/actions/runs/18351497019/job/52272403299

This is reproducible locally running `go test ./internal/js/modules/k6/browser/tests/ -race -count=2` (it almost always happen with count=2 but if you want to be sure to run into the issue, you can increase it).

## Why?

My understanding of the issue is that:
1. In `connection.Execute`, we [cancel the context](https://github.com/grafana/k6/blob/master/internal/js/modules/k6/browser/common/connection.go#L620) right after [sending the message](https://github.com/grafana/k6/blob/master/internal/js/modules/k6/browser/common/connection.go#L617).
2. In `connection.send` method, we have a select that wait on different event (like [receiving the message](https://github.com/grafana/k6/blob/master/internal/js/modules/k6/browser/common/connection.go#L499) or [receiving a context cancelled event](https://github.com/grafana/k6/blob/master/internal/js/modules/k6/browser/common/connection.go#L526)). In most cases, the message is received before and everything goes right. In some cases, the two events are received at the same time and in this case, the Golang doc states `It chooses one at random if multiple are ready.`.

My change ensures this doesn't happen by not immediately cancelling the context but making sure we return from the `send` function when we receive the message and then relying on the existing [`defer evCancelFn()`](https://github.com/grafana/k6/blob/master/internal/js/modules/k6/browser/common/connection.go#L628) to cancel the context when we're done. 

## Checklist

## Related PR(s)/Issue(s)

Closes https://github.com/grafana/k6/issues/4311

### DIFF
--- a/internal/js/modules/k6/browser/common/connection.go
+++ b/internal/js/modules/k6/browser/common/connection.go
@@ -616,7 +616,7 @@ func (c *Connection) Execute(
 					case <-evCancelCtx.Done():
 						c.logger.Debugf("connection:Execute:<-evCancelCtx.Done()#2", "wsURL:%q err:%v", c.wsURL, evCancelCtx.Err())
 					case ch <- msg:
-						// Stopping goroutine as we expect only one response with the matching message ID,
+						// Stopping goroutine as we expect only one response with the matching message ID
 						return
 					}
 				}


internal/js/modules/k6/browser/common/connection.go
```diff
@@ -513,6 +513,7 @@ func (c *Connection) send(
 		case res != nil:
 			return jsonv2.Unmarshal(msg.Result, res, defaultJSONV2Options)
 		}
+		return nil
```
mstoykov, line 516: This seems to be just for readability .. .right? As it will retunr nil just at the end of the outer switch?

joanlopez: That's my assumption as well 👍🏻

Issue #5166, opened 2025-09-10
Implement dependency analysis sub command


pablochacin:
Once the [dependency analysis logic is integrated into k6](https://github.com/grafana/k6/issues/5102) it will no longer be needed in [k6deps](https://github.com/grafana/k6deps), which likely will become an internal package of k6 with a limited functionality of resolving the version constraints for the dependencies. 

However, `k6deps` is still used in some GCK6 processes, as a library as a CLI tool.

It will therefore be convenient to offer this functionality as a `k6` subcommand returning the dependencies in the same format as `k6deps` does.

To replicate the functionality of k6deps CLI, this subcommand requires [support for the dependency manifest](https://github.com/grafana/k6/issues/4729) implemented in k6.

PR #5239, opened 2025-10-10
modules: refactor to be able to load modules without running

mstoykov:
## What?

Let us have a module resolver that just loads a main module and all of its dependencies without running an js code.

This is useful if we need to do more checks or something else and we need the modules that will be loaded but not actually run it.

Will be used for #5166 among other stuff including removing k6deps and esbuild for doing the auto extension resolution.

## Why?

This is basically split from #4866 so it isn't just one huge change. 

And this is one of those changes where the majority of the line changes are in test updates. 

## Checklist

### DIFF
--- a/internal/js/compiler/compiler.go
+++ b/internal/js/compiler/compiler.go
@@ -16,7 +16,6 @@ import (
 	"github.com/sirupsen/logrus"
 
 	"go.k6.io/k6/internal/usage"
-	"go.k6.io/k6/lib"
 )
 
 // A Compiler compiles JavaScript or TypeScript source code into a sobek.Program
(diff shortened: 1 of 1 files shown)


internal/js/compiler/compiler.go
```diff
@@ -70,11 +69,10 @@ func (c *Compiler) Parse(
 	src, filename string, commonJSWrap bool, esm bool,
 ) (prg *ast.Program, finalCode string, err error) {
 	state := &parsingState{
-		loader:            c.Options.SourceMapLoader,
-		compatibilityMode: c.Options.CompatibilityMode,
```
joanlopez, line 74: Nit; looks like this field is no longer used 🤔 so I guess we can remove it?

mstoykov: Unfortunately it is stull used for some stuff https://github.com/grafana/k6/issues/3864 - but it no longer is used by the parser at all.

joanlopez: Sorry, I don't fully get it.

Do we need to keep it there? If it isn't used by the parser, and it's a private attribute 🤔

I know the compatibility mode is used elsewhere, like in the `Bundle` for the global aliasing, but isn't that one read from `RuntimeOptions`? In other words, couldn't we remove the attribute from the `parsingState`?

Thanks!

mstoykov: Ahhh ... I though you meant in general, when I checked that I drop it from the compiler.Options :facepalm: 

yes - it should be remove it, good catch

joanlopez: Apologies, as I wasn't clear enough from the first comment! Thanks! 🙌🏻

mstoykov: Fixed in [b3ccf7a](https://github.com/grafana/k6/pull/5239/commits/b3ccf7a3ba67004d27ee6886cb3232800f1b6c28)

PR #5240, opened 2025-10-10
js: Error with a list of unknown `k6/*` modules

mstoykov:
## What?

This doesn't abort immediately the execution, but continues the resolution step and then errors if there are any unknown modules.

## Why?

This allows the resolution to continue finding unknown modules and potentially even let us then get a new binary that can run the script.

## Checklist

### DIFF
--- a/js/modules/resolution.go
+++ b/js/modules/resolution.go
@@ -2,7 +2,6 @@ package modules
 
 import (
 	"fmt"
-	"maps"
 	"net/url"
 	"slices"
 	"strconv"
@@ -31,16 +30,16 @@ type moduleCacheElement struct {
 
 // ModuleResolver knows how to get base Module that can be initialized
 type ModuleResolver struct {
-	cache           map[string]moduleCacheElement
-	goModules       map[string]any
-	loadCJS         FileLoader
-	compiler        *compiler.Compiler
-	locked          bool
-	reverse         map[any]*url.URL // maybe use sobek.ModuleRecord as key
-	base            *url.URL
-	usage           *usage.Usage
-	logger          logrus.FieldLogger
-	unknowndModules map[string]struct{}
+	cache          map[string]moduleCacheElement
+	goModules      map[string]any
+	loadCJS        FileLoader
+	compiler       *compiler.Compiler
+	locked         bool
+	reverse        map[any]*url.URL // maybe use sobek.ModuleRecord as key
+	base           *url.URL
+	usage          *usage.Usage
+	logger         logrus.FieldLogger
+	unknownModules []string
 }
 
 // NewModuleResolver returns a new module resolution instance that will resolve.
@@ -51,15 +50,14 @@ func NewModuleResolver(
 	u *usage.Usage, logger logrus.FieldLogger,
 ) *ModuleResolver {
 	return &ModuleResolver{
-		goModules:       goModules,
-		cache:           make(map[string]moduleCacheElement),
-		loadCJS:         loadCJS,
-		compiler:        c,
-		reverse:         make(map[any]*url.URL),
-		base:            base,
-		usage:           u,
-		logger:          logger,
-		unknowndModules: make(map[string]struct{}),
+		goModules: goModules,
+		cache:     make(map[string]moduleCacheElement),
+		loadCJS:   loadCJS,
+		compiler:  c,
+		reverse:   make(map[any]*url.URL),
+		base:      base,
+		usage:     u,
+		logger:    logger,
 	}
 }
 
@@ -71,20 +69,13 @@ func (mr *ModuleResolver) resolveSpecifier(basePWD *url.URL, arg string) (*url.U
 	return specifier, nil
 }
 
-func (mr *ModuleResolver) unknownModules() []string {
-	if len(mr.unknowndModules) == 0 {
-		return nil
-	}
-	return slices.Collect(maps.Keys(mr.unknowndModules))
-}
-
-func (mr *ModuleResolver) requireModule(name string) (sobek.ModuleRecord, error) {
+func (mr *ModuleResolver) initializeGoModule(name string) (sobek.ModuleRecord, error) {
 	if mr.locked {
 		return nil, fmt.Errorf(notPreviouslyResolvedModule, name)
 	}
 	mod, ok := mr.goModules[name]
 	if !ok {
-		mr.unknowndModules[name] = struct{}{}
+		mr.unknownModules = append(mr.unknownModules, name)
 		return &unknownModule{name: name, requested: make(map[string]struct{})}, nil
 	}
 	// we don't want to report extensions and we would have hit cache if this isn't the first time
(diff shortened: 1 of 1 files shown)


js/modules/resolution.go
```diff
@@ -328,3 +343,20 @@ func ExportGloballyModule(rt *sobek.Runtime, modSys *ModuleSystem, moduleName st
...
+// UnknownModulesError is returned when loading a module was not possbile due to one or more of dependencies
+// that couldn't be resolved.
+type UnknownModulesError struct {
+	unknownModules []string
+}
+
+func (u UnknownModulesError) Error() string {
+	return fmt.Sprintf("unknown modules %q were tried to be loaded, but couldn't - "+
+		"this likely means binary provisioning is required or a custom k6 binary with more extensions",
+		u.unknownModules)
```
ankur22: I think the wording should be along the lines of:

_couldn't load unknown modules %q -- likely due to either requiring binary provisioning or needing a custom k6 binary with the required extensions._

ankur22 (dismissed):

internal/cmd/tests/cmd_run_test.go
```diff
@@ -1032,6 +1032,35 @@ func TestAbortedByScriptSetupErrorWithDependency(t *testing.T) {
...
+	`
+	mainScript := `
+		import { something } from "k6/x/somethinghere"
+		import "./a.js"
+		export default function () { }
+	`
+
+	ts := NewGlobalTestState(t)
+	ts.DefaultFlags.AutoExtensionResolution = false
+	ts.Flags.AutoExtensionResolution = false
```
codebien: Is it really required to set both?

internal/cmd/tests/cmd_run_test.go
```diff
@@ -1032,6 +1032,35 @@ func TestAbortedByScriptSetupErrorWithDependency(t *testing.T) {
...
+
+	ts.CmdArgs = []string{"k6", "run", "-v", "--log-output=stdout", "test.js"}
+	ts.ExpectedExitCode = int(exitcodes.ScriptException)
+
+	cmd.ExecuteWithGlobalState(ts.GlobalState)
+
+	stdout := ts.Stdout.String()
+	t.Log(stdout)
+
+	assert.Contains(t, stdout, `unknown modules [\"k6/x/anotherone\" \"k6/x/somethinghere\"] were tried to be loaded,`)
```
codebien:
```suggestion
	assert.Contains(t, stdout, `unknown modules [\"k6/x/anotherone\", \"k6/x/somethinghere\"] were tried to be loaded,`)
```
Can we separate the items by commas?

js/modules/resolution.go
```diff
@@ -28,15 +30,16 @@ type moduleCacheElement struct {
...
+	cache           map[string]moduleCacheElement
+	goModules       map[string]any
+	loadCJS         FileLoader
+	compiler        *compiler.Compiler
+	locked          bool
+	reverse         map[any]*url.URL // maybe use sobek.ModuleRecord as key
+	base            *url.URL
+	usage           *usage.Usage
+	logger          logrus.FieldLogger
+	unknowndModules map[string]struct{}
```
codebien:
```suggestion
	unknownModules map[string]struct{}
```

js/modules/resolution.go
```diff
@@ -229,6 +238,10 @@ func (mr *ModuleResolver) LoadMainModule(pwd *url.URL, specifier string, data []
 		panic("somehow running source data for " + specifier + " didn't produce a cyclic module record")
 	}
 
+	unknownModules := mr.unknownModules()
```
codebien:
```suggestion
	if mr.unknownModules() > 0 {
		return newUnknownModulesError(mr.unknownModules())
	}
```
We can probably avoid to allocate it for the happy path

js/modules/resolution.go
codebien:
```suggestion
	return fmt.Sprintf("unknown modules %q were tried to be loaded, but couldn't - "+
		"this likely means automatic extension extension is required or a custom k6 binary with more extensions",
		u.unknownModules)
```
in addition to @ankur22's suggestion, binary provisioning is an outdated name

codebien (commented):
The main logic LGTM, added a few minor comments

js/modules/resolution.go
inancgumus: @codebien FYI: `automatic extension extension`

js/modules/resolution.go
mstoykov: I went with two different thigns and in the end remembered that where we add to the map, we already go through cachining so it will be called once. So I opted for just adding to a slice in the end.

js/modules/resolution.go
mstoykov: I've updated the message

js/modules/resolution.go
```diff
@@ -328,3 +336,33 @@ func ExportGloballyModule(rt *sobek.Runtime, modSys *ModuleSystem, moduleName st
...
+	unknownModules []string
+}
+
+func newUnknownModulesError(list []string) UnknownModulesError {
+	slices.Sort(list)
+	return UnknownModulesError{unknownModules: list}
+}
+
+func (u UnknownModulesError) Error() string {
+	return fmt.Sprintf("unknown modules [%s] were tried to be loaded, but couldn't - "+
```
codebien, line 352: I think there is still a bit of improvement to be done to have a better DX. A revisited version of original Ankur's proposal.

```suggestion
	return fmt.Sprintf("unknown module(s): [%s] k6 couldn't resolve the them because they are not built-in k6 modules. If these are extension modules, enable automatic extension resolution or build a custom binary with xk6. See the docs for more details: https://grafana.com/docs/k6/latest/extensions/explore/#use-extensions
```

mstoykov: I think that we are going overboard on this.

There are more changes to come, and this error is either never going to be seen.  Or it will be seen before it either loads a new binary or returns a next error about extension resolutions. 

I would prefer all the DX/UX work to be done after we move to the new implementation, which already fix bugs that are reported and people have problems with. I do not think that if someone has disabled auto extension resolution, they would not know about extensions.

js/modules/resolution.go
```diff
@@ -328,3 +336,33 @@ func ExportGloballyModule(rt *sobek.Runtime, modSys *ModuleSystem, moduleName st
 		}
 	}
 }
+
+// UnknownModulesError is returned when loading a module was not possbile due to one or more of dependencies
```
joanlopez, line 340:
```suggestion
// UnknownModulesError is returned when loading a module was not possible due to one or more of dependencies
```

joanlopez: Nit; ofc non-blocking.

mstoykov:
@joanlopez if the loading could find a module, the Sobek (ESM) resolution  mechanism will stop ... because well, it could find it. But what we want here is for it to continue until we have loaded everything that we *can*. And then we will check "did we try to load anything that was not found" if so ... abort and propagate back what that was.

This allows us to have k6 internally figure out that `k6/x/sql` and `k6/x/faker` are not found, but we need them for the script. Instead of it stopping on the first one. 

The panic is because an unknown module should never actually be evaluated, if it is that is a bug somewhere else. We only parse the code and link the modules together.

joanlopez:
@mstoykov Thanks for the detailed explanation! 🙌🏻 🙇🏻 

I think I got most of it (especially the high-level picture, and the main purpose), what I wasn't completely sure was if there could be a misuse (now or introduced in the future by mistake) somehow hidden because only `Evaluate` panics. But I guess it's fine for now. Again, thanks! 👍🏻

